PR #959, opened 2026-06-27
fix(server): bound decodeBase58 input length before decoding

jgyy:
## Problem
`decodeBase58` in `server/wallet_link.ts` has no input-length guard. `bs58.decode` runs an **O(n^2)** loop over the input, so a long attacker-controlled string fed through the wallet-link address/signature paths (`isSolanaAddress`, `verifySolanaSignature`) pins the event loop for the duration of the decode. The endpoints are rate-limited, so the impact is bounded, but the per-request work is still unnecessarily large and trivially avoidable.

The existing charset regex guaranteed the decode never *throws*, but said nothing about its *cost*.

## Fix
Reject any input longer than a 128-char cap before decoding:

```ts
if (s.length > MAX_BASE58_LEN) return null;
```

The longest input we ever legitimately decode is a 64-byte ed25519 signature, which base58-encodes to ~88 chars (a 32-byte address is ~44). A 128-char cap leaves comfortable headroom for every real address and signature while bounding the work. Note a tighter `> 64` cap would be **incorrect**: it would reject all valid 64-byte signatures and silently break `verifySolanaSignature`. Guarding inside `decodeBase58` protects all callers at once.

## Tests (test-first)
Added cases in `tests/wallet.test.ts`. The key assertion uses an all-`'1'` string, which is valid base58 (so it is rejected by length alone, not the charset), proving the guard is what does the work:

```
 Test Files  1 passed (1)
      Tests  11 passed (11)
```

Also re-ran `tests/malware_scan.test.ts` (the release-gate backstop that asserts the scanner stays quiet on the real wallet code): 72 passed. `tsc --noEmit` clean; Biome clean on both files.

## Screenshots
Not applicable: this is a pure, IO-free server helper with no UI surface. The green test run above is the behavioral evidence.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["wallet-link address/signature -> decodeBase58"] --> B{"Before"}
  B --> C["charset regex only, no length cap"]
  C --> D["bs58.decode runs O(n^2) over a long string -> pins the event loop"]
  A --> E{"After"}
  E --> F{"length > MAX_BASE58_LEN (128)?"}
  F -->|yes| G["return null - reject before decoding"]
  F -->|no| H["decode (88-char signature fits with headroom)"]
```

### DIFF
--- a/server/wallet_link.ts
+++ b/server/wallet_link.ts
@@ -2,14 +2,22 @@
 // validation, the challenge-message format, and ed25519 signature verification.
 // Kept separate from server/wallet.ts (which does DB + HTTP) so it can be unit
 // tested without a database.
-import bs58 from 'bs58';
+
 import { ed25519 } from '@noble/curves/ed25519';
+import bs58 from 'bs58';
 
 // Base58 alphabet (Bitcoin/Solana). Testing the charset first guarantees the
-// decode below never throws, so no input-shape guard is needed around it.
+// decode below never throws.
 const BASE58 = /^[1-9A-HJ-NP-Za-km-z]+$/;
 
+// bs58.decode is O(n^2) in the input length. The longest input we ever
+// legitimately decode is a 64-byte ed25519 signature (~88 base58 chars), so a
+// generous 128-char cap keeps a hostile caller from pinning the loop with a huge
+// string while leaving every real address/signature comfortably under it.
+const MAX_BASE58_LEN = 128;
+
 export function decodeBase58(s: string): Uint8Array | null {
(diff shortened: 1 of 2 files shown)


jgyy:
Ping for review (folks with push/merge access): @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk

Small, isolated, backend-only hardening of `decodeBase58` (adds a length cap before the O(n^2) decode). Test-first, all green, no behavior change for valid addresses/signatures. Thanks!

jgyy:

PR #961, opened 2026-06-27
fix(sim): persist parked mana for shapeshifted druids on save

jgyy:
## Problem
A druid saved while shapeshifted reloads with a wrong, very low resource. In bear/cat form `recalcPlayerStats` (`entity.ts`) parks the real mana in `e.savedMana` and repurposes `e.resource` to hold the form bar (bear rage / cat energy). `serializeCharacter` persisted that form bar verbatim, e.g. `35`. On reload `addPlayer` sees the class is a mana class and clamps `35` into the **mana** pool, so the druid loads with ~35 mana and has to self-heal/regen back up.

Forms are auras and are **not** persisted, so the reload always rebuilds the character in caster form. That means the persisted value must be the parked mana, not the form bar.

## Fix
New pure, host-agnostic leaf `src/sim/serialize_resource.ts`:

```ts
persistedResource(classResourceType, liveResourceType, liveResource, savedMana)
```

Returns the parked `savedMana` when a mana class is currently on a non-mana form bar, and the live resource otherwise. `serializeCharacter` consumes it; every non-shifted and non-mana class is byte-identical to before. Kept as a leaf so a Vitest drives it directly and the `sim.ts` coordinator stays a thin consumer (per `src/sim/CLAUDE.md`).

## Tests
`tests/shapeshift_resource_persist.test.ts`:
- shifted druid: serialize -> reload restores mana (not the 35 form bar), via the real `recalcPlayerStats` form path
- unshifted mana class and rogue energy round-trip unchanged
- pure-function unit cases (shifted mana class, matching bar, null-resource class)

Determinism unaffected (no rng draws in the serialize path); `tests/parity`, `tests/persistence_round_trip`, `tests/form_command`, `tests/architecture`, `tests/low_resource` all green. `tsc --noEmit` clean. No player-visible strings changed (S3 i18n guard not implicated). Backend-only persistence fix, so no screenshots.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["druid in bear/cat form"] --> B["recalcPlayerStats parks mana in savedMana; resource holds form bar (rage/energy)"]
  B --> C{"serializeCharacter"}
  C -->|before| D["persists the form bar value (e.g. 35)"]
  D --> E["reload clamps 35 into the MANA pool -> loads with ~35 mana"]
  C -->|after| F["persistedResource() returns the parked savedMana"]
  F --> G["reload rebuilds in caster form with real mana restored"]
```

### DIFF
--- a/src/sim/serialize_resource.ts
+++ b/src/sim/serialize_resource.ts
@@ -1,0 +1,35 @@
+// Which resource value to persist for a character at save time.
+//
+// Druid forms swap the live resource bar: bear runs on rage, cat on energy, and
+// the character's real mana is parked in `e.savedMana` (see recalcPlayerStats in
+// entity.ts). Forms are auras and are NOT persisted, so a reloaded character is
+// always rebuilt in caster form. That makes saving the live form bar a bug: a
+// mana class would persist e.g. 35 (rage) and reload clamped to 35 MANA.
+//
+// This leaf decides the value to write into CharacterState.resource. Kept pure
+// and host-agnostic so a Vitest drives it directly and serializeCharacter stays a
+// thin consumer.
+
+import type { ResourceType } from './types';
+
+/**
+ * The resource value to persist for a character.
+ *
+ * @param classResourceType the class's natural resource (CLASSES[cls].resourceType)
+ * @param liveResourceType   the entity's CURRENT bar (rage/energy while shifted)
+ * @param liveResource       the entity's current resource amount (e.resource)
+ * @param savedMana          mana parked while shifted (e.savedMana; 0 when unshifted)
+ */
+export function persistedResource(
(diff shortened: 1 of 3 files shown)


jgyy:
cc @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk for review.

Small, self-contained persistence correctness fix: shapeshifted druids were saving their form bar (rage/energy) into `resource`, which reloaded as near-empty mana. Fix normalizes to the parked `savedMana` on save via a pure tested leaf (`src/sim/serialize_resource.ts`); no behavior change for any non-shifted or non-mana class, determinism and the parity gate unaffected. Backend-only, no UI surface.

jgyy:

PR #963, opened 2026-06-27
fix(sim): don't persist warlock demons across logout

jgyy:
## Problem
Summoned warlock demons reuse the hunter-pet save/load path, so they were persisted in `CharacterState.pet` and re-granted on the next login. A relog returned the demon (Imp / Voidwalker / Succubus / Felhunter / Doomguard / Infernal) **instantly and for free**, and because summon cooldowns are not persisted, the 180s summon cooldown was reset in the process. Logging out and back in laundered the cooldown and skipped the summon cost.

In classic terms, warlocks re-summon their demon on login (paying its cost); only hunter pets persist across sessions.

## Fix
Drop the pet snapshot to `null` at the `serializeCharacter` persistence boundary when the pet's `MOBS` family is `demon` (new pure `isDemonPetState` predicate in `src/sim/pet/pet_commands.ts`). Every server save path (`saveCharacterOnLeave`, `saveCharacter`, autosave, shutdown `saveAll`) routes through `sim.serializeCharacter`, so this one boundary covers all logout paths. The only restore path (`addPlayer` -> `restorePet`, gated on `state.pet`) then spawns no demon and the player must re-summon.

### Why the persistence boundary and not `serializePet`
The issue suggested returning `null` from `serializePet` directly, but that function is also reused by the **in-session delve pet-park** (`stowPetForDelve`), which deliberately preserves a demon across a delve within the same session. Dropping demons inside `serializePet` would regress that. Placing the drop in `serializeCharacter` (the cross-logout boundary) fixes the laundering without touching in-session behavior. Hunter beast/spider pets continue to persist everywhere.

## Tests
Test-first. `tests/pet_command.test.ts`:
- Flips the prior "restores a saved warlock demon" case to assert the demon is dropped on save and absent after reload (forcing a fresh re-summon).
- Adds a case asserting a non-demon hunter beast pet still persists across save/reload.

Green locally: `pet_command`, `warlock_pets`, `persistence_round_trip`, `delves`, `delve_companion`, `architecture`, and the full `tests/parity` golden-trace + rng-draw-order gate (154 cases) all pass; `tsc --noEmit` clean. The fix is outside `tick()` and changes no rng draw order.

Note: this is a save/reload persistence change with no rendering change, so there is no meaningful screenshot; the deterministic test is the proof.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["logout -> serializeCharacter"] --> B{"pet MOBS family"}
  B -->|hunter beast/spider| C["persist pet (unchanged)"]
  B -->|"demon (before)"| D["persisted in CharacterState.pet"]
  D --> E["login re-grants demon free + launders the 180s summon cooldown"]
  B -->|"demon (after)"| F["isDemonPetState -> drop snapshot to null"]
  F --> G["login spawns no demon; player must re-summon, paying the cost"]
  C --> H["in-session delve pet-park still preserves the demon"]
```


jgyy:
Ping for review: @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk

Small, self-contained sim fix: warlock demons no longer persist across logout (they were re-granted free on relog, laundering the 180s summon cooldown). Drop is at the `serializeCharacter` boundary so in-session delve pet-parking and hunter pets are unaffected. Test-first; full parity gate + tsc green.

jgyy:

FernandoX7 (commented):
Solid fix, and the reasoning is the strong part. Putting the drop at the `serializeCharacter` persistence boundary instead of inside `serializePet` is the right call: it leaves the in-session delve pet-park (`stowPetForDelve`) intact while closing the cross-logout laundering. The predicate is pure, lives in the right module, and `MOBS` was already imported there. Verified against the head ref:

- All save paths go through `sim.serializeCharacter` (`game.ts:1299` leave, `game.ts:1521` autosave/saveAll, `main.ts:202` offline), and `serializePet` is only otherwise consumed by the in-session delve park, so the one boundary really does cover every logout path.
- `serializePet` is a side-effect-free read (no `Rng`, no entity mutation), so hoisting it into `const petSnapshot` ahead of the object literal is behavior-identical and does not touch rng draw order. The change is outside `tick()`.
- Restore stays gated on `savedState?.pet` (`sim.ts:1271`), so a `null` snapshot spawns nothing and forces a fresh summon.
- `family === 'demon'` catches all seven warlock pets (imp, voidwalker, succubus, felhunter, felguard, infernal, doomguard); hunter beast/spider pets are untouched. The mid-delve logout case also resolves correctly: `serializePet` falls back to the stashed snapshot, which the predicate then drops, so a warlock who logs out inside a delve also re-summons. That is consistent with the rule.

Test-first is done properly: the old "restores a saved demon" case is flipped to assert the drop, and the new hunter-persists case guards the other half. `pet_command` passes locally (8/8).

Nice catch on the last commit (`04b52b08`): the base merge had reintroduced a second unguarded `pet: this.serializePet(pid)` next to the guarded one, which would have silently reverted the fix. The head ref now has exactly one `pet:` key and it is the guarded one.

Nits, none blocking:
- The PR description says six demons; there are seven in `warlock_pets.ts` (felguard is the one omitted from the prose). No code impact, the family predicate covers all of them.
- Worth a one-line mention in the description that summon-cooldown persistence is orthogonal and out of scope: the fix removes the free re-grant, and re-summon cost/cooldown behavior on login follows from cooldowns not being laundered anymore. Not a code change request.

Looks good to merge.

FernandoX7 (approved):

PR #965, opened 2026-06-27
fix(sim): drain heal-absorb shields on HoT, food/drink, and potion ticks

jgyy:
## Problem

A heal-absorb shield (the `heal_absorb` aura, e.g. the Gravecaller Summoner's **"Grave Blight"**) is meant to devour the next chunk of incoming healing before any of it lands. But `consumeHealAbsorb` ran **only inside `applyHeal`**. Three other heal paths add HP directly and so slipped straight past the shield:

- **Heal-over-time ticks** (the `hot` aura branch) - `src/sim/combat/auras.ts`
- **Food / drink regen ticks** - `src/sim/combat/auras.ts`
- **Healing potions** - `src/sim/sim.ts`

Net effect: a player branded with Grave Blight could just HoT/eat/quaff through it, and the debuff expired unused.

## Fix

Route all three paths through `consumeHealAbsorb`, in the **same order of operations as `applyHeal`**: drain the shield against the pre-`maxHp`-cap rounded amount, then cap to missing HP. So overheal still drains the shield, exactly like a direct heal.

`consumeHealAbsorb` already existed as a `heal.ts` function and a private `Sim` method; it is now exposed on the **`SimContext` seam** (append-only, mirroring how `healingTakenMult` / `healingThreat` are already surfaced) so the aura/regen runner drains it through the seam instead of reaching into `Sim`.

## Determinism

`consumeHealAbsorb` draws **no** rng, so no draw is added or moved. The parity gate confirms the rng draw-order fingerprint is byte-identical (`draws: 10`, digest unchanged); only HP/shield state moved.

## Test harness honesty

The `multi_class_heal` parity scenario had **encoded the bug** in a comment:
> "the hot branch never calls consumeHealAbsorb, only applyHeal does."

That scenario is rewritten to demonstrate the **corrected** behavior - the HoT drains a fresh shield, depletes it, then heals - and the golden is regenerated via `UPDATE_PARITY=1` (diff scoped to that one file). Coverage of `healingTakenMult` / `healingThreat` via the HoT path is retained, plus a new assertion that the HoT drained + depleted its shield.

## Evidence

New `tests/heal_absorb_ticks.test.ts` (HoT, food, potion, and a control that heals normally once the shield is gone):

```
Test Files  1 passed (1)
     Tests  4 passed (4)
```

Parity golden delta for the HoT-under-shield window (this is a deterministic sim-logic fix with no visual surface, so the behavior change is shown via the pinned golden rather than a UI screenshot):

```
- "value": 3700      (shield rode along untouched - the bug)
- "hp": 3200         (HoT healed straight through the shield)
+ "hp": 2750         (HoT absorbed until the shield depletes, then heals; shield gone)
```

Verified green: `tsc --noEmit`, `tests/architecture.test.ts` (sim purity + determinism guard), `tests/localization_fixes.test.ts` (S3 i18n guard), the full `tests/parity` gate, and `npm run build`.

## Scope note

The bug report names HoT + food/drink + potion; those three are fixed. Other direct-HP heal paths (passive out-of-combat regen, delve-companion heal, mob/pet self-heals, polymorph self-heal) are intentionally left out of scope - heal-absorb is an anti-healer debuff and these are separate, lower-risk surfaces.

cc @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk - small deterministic sim-core bug fix, would appreciate a look.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["heal_absorb shield active (e.g. Grave Blight)"] --> B{"heal path"}
  B -->|applyHeal| C["consumeHealAbsorb - drains shield first"]
  B -->|"HoT tick (before)"| D["direct HP add - bypasses shield"]
  B -->|"food/drink (before)"| D
  B -->|"potion (before)"| D
  D --> E["shield never drains, debuff expires unused - bug"]
  B -->|"HoT / food / potion (after)"| F["consumeHealAbsorb via SimContext seam"]
  F --> G["shield drains first (same order as applyHeal), then heal"]
```


jgyy:

FernandoX7 (commented):
Nice catch and a clean fix. The diagnosis is right (only `applyHeal` ran `consumeHealAbsorb`, so the three direct-HP paths slipped past the shield), the three new call sites preserve `applyHeal`'s order exactly (round the `healingTakenMult` amount, drain the shield, then cap to missing HP, so overheal still drains), and surfacing it on the `SimContext` seam instead of reaching into `Sim` is the right shape. Determinism holds since `consumeHealAbsorb` draws no rng, and the seam addition is mirrored in both test mocks. Rewriting the `multi_class_heal` scenario that had literally encoded the bug in a comment is the honest move.

One real issue, plus two smaller notes.

**should-fix** `src/sim/combat/auras.ts:162` (with `src/sim/combat/heal.ts:72`): the HoT branch double-processes its own aura on the tick a shield depletes. `consumeHealAbsorb` drops a depleted shield with `target.auras = target.auras.filter(...)` (heal.ts:72), but `updateAuras` is walking `e.auras` with a reverse index loop (auras.ts:126) and calls it from the `hot` branch. When the depleting `heal_absorb` sits at a lower index than the HoT (the common case: the debuff was applied before the HoT), the array shrinks mid-loop, the HoT shifts from `i` to `i-1`, and the next iteration re-reads it, so it takes an extra `remaining -= DT` and `tickTimer -= DT`.

Repro, single `updateAuras` call with the absorb before the hot: shield survives gives `hot.remaining 9.95`; shield depletes gives `hot.remaining 9.90` (two DT). Control with the absorb after the hot gives `9.95`. That is exactly this PR's golden delta on `hot_tk` (`remaining 2.6 -> 2.55`, `tickTimer 0.1 -> 0.05`), so the artifact is currently baked into the parity baseline as if intended. Practically it only shortens the HoT by one tick today (no double-heal, since every HoT interval is >= ~1s), but it is a latent double-tick if a short-interval HoT is ever added, and locking it into the golden makes it harder to spot later. The `useItem` and `updateRegen` paths do not iterate `e.auras`, so only this one call site is affected.

Direction: do not mutate `e.auras` during that loop. For example, have `consumeHealAbsorb` zero the depleted shield and defer removal (the `a.value <= 0` guard already makes a 0-value shield a no-op, and it expires on its own `remaining`), or snapshot the array in `updateAuras`. The `applyHeal` site is unaffected. Heads up that `tests/parity/coverage.test.ts`'s new `hot_absorb`-absent assertion leans on the immediate filter, so the fix, that assertion, and the golden move together.

**nit** `src/sim/combat/auras.ts:76`: passive out-of-combat regen still adds HP directly without draining the shield, four lines above the food tick you did fix. The scope note flags this as intentional and lower-risk, which is fair, just noting it is the one heal-through left sitting right next to a fixed one.

**nit** `tests/heal_absorb_ticks.test.ts`: none of the unit tests exercise a depleting HoT tick (test 1 uses absorb 100 against a 60 heal, so the shield survives). Only the parity golden hits depletion, and it absorbed the artifact above. A unit assertion on `hot.remaining` / hp after a shield-depleting HoT tick would have caught it, and would pin the corrected behavior once the loop issue is fixed.

jgyy:
Fixed in 927ccf66, thanks for the catch.

consumeHealAbsorb (src/sim/combat/heal.ts) no longer splices the depleted shield out of target.auras synchronously. It now just zeroes the shield's value (the existing a.value <= 0 guard already treats a zeroed shield as a no-op on the next drain), and lets it expire naturally via its own remaining. That means updateAuras's reverse-index loop is never mutated mid-iteration by the HoT branch's call into consumeHealAbsorb, so the double-DT decrement you found is gone. applyHeal is unaffected since it never iterates target.auras.

Updated tests/parity/coverage.test.ts's multi_class_heal assertion: it now checks the hot_absorb shield is present-but-zeroed instead of absent, and I regenerated the multi_class_heal golden (via UPDATE_PARITY=1) so hot_tk's remaining/tickTimer reflect the corrected single-DT tick instead of the old double-tick artifact.

Also added the unit test you flagged in tests/heal_absorb_ticks.test.ts: it pushes a shield sized to deplete on the same tick as a HoT sitting at a higher array index, runs a single updateAuras call, and pins hot.remaining to 9.95 / tickTimer to 1.95 (the single-DT values) plus hp to 550 and the zeroed shield value to 0. It fails against the old double-splice behavior (9.90 / 1.90) and passes now.

npx vitest run tests/heal_absorb_ticks.test.ts tests/parity tests/architecture.test.ts and npx tsc --noEmit are all green.

jgyy:
Closeout check: FernandoX7's should-fix (HoT double-tick from mid-loop `e.auras` mutation on shield depletion) and both nits are already resolved on this branch (af90d7d2, 959c98a7, d3fed500): `consumeHealAbsorb` now zeroes a depleted shield instead of splicing it out mid-iteration, the parity golden and `multi_class_heal` coverage assertion were updated accordingly, and `tests/heal_absorb_ticks.test.ts` now pins the single-DT tick on a depleting HoT.

Verified on the current head (d3fed500): `npx tsc --noEmit` clean, `npx vitest run tests/heal_absorb_ticks.test.ts tests/parity tests/architecture.test.ts` all green (188 tests), no merge conflicts against `release/v0.18.0` (mergeable). No further changes needed.

jgyy:

PR #966, opened 2026-06-27
fix(talents): scale Consecration damage via Holy Shield (groundAoE in scaleEffect)

jgyy:
## Problem

The **Holy Shield** talent (`prot_holy_shield`, `talents_classic.ts`) is documented as *"Increases Consecration damage by 12% and threat by 8% per rank"*, but the **damage half is a no-op**. Only the threat bonus (`global.threatPct`) ever applies.

## Root cause

The talent contributes an `ability: [{ ability: 'consecration', dmgPct: 0.12 }]` modifier. `applyTalentMods` (`src/sim/content/classes.ts`) folds that into a damage multiplier and maps every effect through `scaleEffect`. But `scaleEffect`'s per-effect-type `switch` had **no `groundAoE` case** -- and Consecration's only effect is `{ type: 'groundAoE', ... }`. So the effect fell through to `default: return eff` and came back unscaled.

`groundAoE` is structurally a twin of `aoeDamage` (both carry `min`/`max`/`radius`), but it is a distinct union member, so it was simply missed.

## Fix

Add a `groundAoE` case to `scaleEffect`, mirroring `aoeDamage` (scale `min`/`max` by the damage multiplier, keeping the `flat` term so a future flat-damage talent on a ground AoE also lands).

```ts
case 'groundAoE': return { ...eff, min: Math.round(eff.min * dmgMult + flat), max: Math.round(eff.max * dmgMult + flat) };
```

## Test

Added test-first in `tests/talents.test.ts`: snapshot-locks Consecration's `groundAoE` magnitudes before/after Holy Shield r2 (+24%) and asserts the shared content data is not mutated by the modifier pass. It fails before the fix (28 stays 28) and passes after (28 -> 35 on min).

## Verification

- `npx vitest run tests/talents.test.ts` (47 passed)
- `npx vitest run tests/sim.test.ts tests/progression.test.ts` (101 passed)
- `npx vitest run tests/architecture.test.ts` (sim purity, passed)
- `npx tsc --noEmit` clean

No balance numbers invented; the talent's existing `dmgPct` now simply reaches the ground-AoE tick.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["Holy Shield contributes dmgPct 0.12 to Consecration"] --> B["applyTalentMods -> scaleEffect"]
  B --> C{"per-effect-type switch"}
  C -->|aoeDamage| D["scaled by dmgMult"]
  C -->|"groundAoE (before)"| E["no case -> default returns eff UNSCALED"]
  E --> F["Consecration damage bonus = NO-OP (only threat applied)"]
  C -->|"groundAoE (after)"| G["scale min/max by dmgMult + flat"]
  G --> H["Consecration damage actually increases (28 -> 35 at r2)"]
```

### DIFF
--- a/src/sim/content/classes.ts
+++ b/src/sim/content/classes.ts
@@ -3536,6 +3536,12 @@ function scaleEffect(
         min: Math.round(eff.min * dmgMult + flat),
         max: Math.round(eff.max * dmgMult + flat),
       };
+    case 'groundAoE':
+      return {
+        ...eff,
+        min: Math.round(eff.min * dmgMult + flat),
+        max: Math.round(eff.max * dmgMult + flat),
+      };
     case 'aoeRoot':
       return { ...eff, min: Math.round(eff.min * dmgMult), max: Math.round(eff.max * dmgMult) };
     case 'drainTick':
--- a/src/sim/content/classes.ts
+++ b/src/sim/content/classes.ts
@@ -3581,24 +3581,6 @@ function scaleEffect(
       return { ...eff, amount: Math.round(eff.amount * dmgMult + flat) };
     default:
       return eff;
-    case 'weaponDamage': return { ...eff, bonus: Math.round(eff.bonus * dmgMult + flat) };
-    case 'weaponStrike': return { ...eff, bonus: Math.round(eff.bonus * dmgMult + flat) };
-    case 'directDamage': return { ...eff, min: Math.round(eff.min * dmgMult + flat), max: Math.round(eff.max * dmgMult + flat) };
-    case 'dot': return { ...eff, total: Math.round(eff.total * dmgMult + flat) };
-    case 'aoeDamage': return { ...eff, min: Math.round(eff.min * dmgMult + flat), max: Math.round(eff.max * dmgMult + flat) };
-    case 'groundAoE': return { ...eff, min: Math.round(eff.min * dmgMult + flat), max: Math.round(eff.max * dmgMult + flat) };
-    case 'aoeRoot': return { ...eff, min: Math.round(eff.min * dmgMult), max: Math.round(eff.max * dmgMult) };
-    case 'drainTick': return { ...eff, min: Math.round(eff.min * dmgMult), max: Math.round(eff.max * dmgMult) };
-    case 'finisherDamage': return { ...eff, base: Math.round(eff.base * dmgMult + flat), perCombo: Math.round(eff.perCombo * dmgMult) };
-    case 'imbue': return { ...eff, bonus: Math.round(eff.bonus * dmgMult + flat), judgeMin: eff.judgeMin === undefined ? undefined : Math.round(eff.judgeMin * dmgMult + flat), judgeMax: eff.judgeMax === undefined ? undefined : Math.round(eff.judgeMax * dmgMult + flat) };
-    case 'heal': return { ...eff, min: Math.round(eff.min * healMult + flat), max: Math.round(eff.max * healMult + flat) };
-    case 'hot': return { ...eff, total: Math.round(eff.total * healMult + flat) };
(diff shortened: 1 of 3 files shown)


jgyy:
cc maintainers with merge access: @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk -- small, isolated `paladin`/`talents` correctness fix, would appreciate a look.

**Before/after evidence** (the new `tests/talents.test.ts` case, Consecration `groundAoE` `min` at level 20 with Holy Shield r2, +24%):

Before the fix (talent damage dropped):
```
AssertionError: expected 28 to be 35
- 35   (expected, +24%)
+ 28   (received, unscaled)
```

After the fix:
```
Test Files  1 passed (1)
     Tests  47 passed (47)
```

A note on screenshots: this is a damage-formula fix in `scaleEffect`, not a UI change -- the talent tooltip already read "+12% per rank" before this PR; only the underlying ground-AoE tick magnitude was wrong. So the meaningful before/after is the numeric one above rather than an in-game capture. Happy to add an in-client damage-log screenshot if a maintainer prefers it.

jgyy:

FernandoX7 (commented):
Clean, well-scoped fix with a correctly diagnosed root cause. `scaleEffect` had an `aoeDamage` arm but no `groundAoE` arm (`classes.ts:3539`), and Consecration's only effect is `groundAoE`, so the talent's `dmgPct` fell through `default` and came back unscaled while the threat half still applied. The new arm mirrors `aoeDamage` exactly.

I confirmed the scaling is load-bearing rather than cosmetic: the dispatcher builds the live `GroundAoE` straight from `eff.min`/`eff.max` (`src/sim/combat/effect_dispatch.ts:439`), so the resolved magnitudes feed the actual per-tick damage.

Verified at the PR head:

- `scaleEffect` returns a fresh object via spread, and the test's non-mutation assertion holds, so the shared content arrays are safe.
- Pure arithmetic, no Rng / wall-clock / DOM. `tests/architecture.test.ts` passes.
- `tests/talents.test.ts` (including the new case) and `tsc --noEmit` are clean. Consecration goes 28 to 35 / 34 to 42 at Holy Shield r2 as described.

One thing worth a line in your notes, not a change request:

- note: the fix is broader than the title. Any global or per-ability damage mod now scales a `groundAoE` effect, not just Holy Shield. Today Consecration is the only `groundAoE` ability in content, so it is the only behavioral change, but the seam is now correct for any future ground AoE. Good direction.

The test-first regression assertion (that min actually increases, not just matches the formula) is exactly what would have caught the original no-op. Nothing blocking.

jgyy:
Rebased onto release/v0.18.0.

Review from @FernandoX7 was an approval-style walkthrough with no change requests (just a helpful note that the fix is a correct, general-purpose groundAoE arm in scaleEffect, not Holy Shield-specific -- acknowledged, no code change needed there). No other reviewers left inline comments or change requests on this PR.

Rebase notes:
- Two conflicts surfaced re-applying this PR's intent onto the moved base: an import-order conflict in tests/talents.test.ts (kept both sets of imports, alphabetized) and a bad auto-merge in src/sim/content/classes.ts that left a duplicated scaleEffect switch body (two default clauses, a TS1113 compile error). Removed the stale duplicate, keeping the well-formatted arms including the new groundAoE case this PR adds.
- Also dropped an em dash from an unrelated hud.ts comment that came in through the rebase, to satisfy the repo's no-em/en-dash convention enforced by the pre-push hook.

Verified after rebase: npx tsc --noEmit clean, tests/talents.test.ts (59 tests) and tests/architecture.test.ts green, biome check clean on touched files. Force-pushed the rebased branch.

jgyy:

PR #967, opened 2026-06-27
fix(server): cap chat text before the hard-word filter scan

jgyy:
## Problem

`enforceChatPolicy` (server/game.ts) runs the hard-word filter
(`findHardWord` -> `text.match(TOKEN_RE)`, a Unicode tokenizing regex) over
the **entire** trimmed chat text. A `chat` message rides a WS frame that can
be up to the `maxPayload` of 16 KiB, but only the first
`MAX_CHAT_MESSAGE_LEN` (255) chars are ever routed to other players or stored.

So every chat send scans up to ~64-170x more text than will ever be
delivered. The work is bounded only by the chat token bucket, making it a
cheap per-message CPU-amplification vector against the single shared world
loop.

It is also a correctness bug: a hard word sitting **past** char 255 currently
triggers a punitive mute, even though it is sliced off before anyone sees the
message.

## Fix

Cap the text to `MAX_CHAT_MESSAGE_LEN` before scanning (and bound the
recorded-violation `message` to the same slice). The filter now never inspects
bytes that are not part of the routed/stored message.

## Tests

`tests/chat_filter_cap.test.ts` pins the behavior via `GameServer`:
- a hard word within the routed slice still flags (mute path intact),
- a hard word that only appears past the cap no longer flags,
- a hard word ending exactly at the cap boundary still flags.

Verified: new test red before / green after, `tests/chat_filter.test.ts`,
`tests/game_sessions.test.ts`, the S3 i18n guard
(`tests/localization_fixes.test.ts`), and `tsc --noEmit` all pass.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["chat WS frame, up to 16 KiB maxPayload"] --> B{"Before"}
  B --> C["findHardWord scans the ENTIRE trimmed text"]
  C --> D["~64-170x routed slice: CPU amplification + a slur past char 255 unfairly mutes"]
  A --> E{"After"}
  E --> F["slice to MAX_CHAT_MESSAGE_LEN (255) first"]
  F --> G["filter scans only the routed/stored slice"]
```

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -74,7 +74,7 @@ import { nextRaidResetMs } from './raid_reset';
 import { REALM, REALM_PUBLIC_ORIGIN, REALM_RESET_TIME_ZONE } from './realm';
 import { createSerialWriter } from './serial_writer';
 import type { Presence, PresenceStatus, SocialActor, SocialTransport } from './social';
-import { SocialService } from './social';
+import { GUILD_MESSAGE_MAX, SocialService } from './social';
 import { PgSocialDb } from './social_db';
 import { TickProfiler } from './tick_profiler';
 import { holderInfoForPubkey } from './woc_balance';
@@ -113,6 +113,10 @@ const LOCKPICK_ACTIONS = new Set<PickAction>(['hardSet', 'set', 'steady', 'ease'
 const LEAVE_SAVE_MAX_ATTEMPTS = 5;
 const LEAVE_SAVE_RETRY_BASE_MS = 250;
 const LEAVE_SAVE_RETRY_MAX_MS = 4000;
+// Guild/officer chat prefixes: shared between enforceChatPolicy (which must
+// scan the exact body guildChat/officerChat will deliver) and the chat router.
+const GUILD_CHAT_PREFIX = /^\/(?:g|gu|guild)\s+([\s\S]+)$/i;
+const OFFICER_CHAT_PREFIX = /^\/(?:o|officer)\s+([\s\S]+)$/i;
 const CHAT_RATE_BURST = 5;
 const CHAT_RATE_REFILL_PER_SECOND = 1 / 3; // sustained 20 messages/minute
 const CHAT_RATE_ERROR_COOLDOWN_SECONDS = 4;
@@ -2295,8 +2299,8 @@ export class GameServer {
         // guild and officer chat are persistent + cross-zone, so they live in
         // the server's SocialService rather than the sim (no guild concept).
         // MMO convention: /g is guild; /general remains world chat.
-        const gm = /^\/(?:g|gu|guild)\s+([\s\S]+)$/i.exec(text);
-        const om = gm ? null : /^\/(?:o|officer)\s+([\s\S]+)$/i.exec(text);
+        const gm = GUILD_CHAT_PREFIX.exec(text);
+        const om = gm ? null : OFFICER_CHAT_PREFIX.exec(text);
         if (gm || om) {
           const channel = gm ? 'guild' : 'officer';
           const match = gm ?? om;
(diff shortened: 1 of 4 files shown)


jgyy:

FernandoX7 (commented):
Good motivation on the CPU side: capping the scan to `MAX_CHAT_MESSAGE_LEN` closes the per-message amplification against the shared world loop, and the say-path tests pin it cleanly (within-cap flags, past-cap does not, boundary flags).

The problem is that "only the first 255 chars are ever routed or stored" does not hold for guild and officer chat, and the cap opens a hard-word bypass there.

**blocking** — guild/officer chat can deliver a slur the filter never scans (`server/game.ts:2442`). `enforceChatPolicy` runs on the raw frame text before any channel parsing (`server/game.ts:1556`) and now scans `text.slice(0, 255)`. But the explicit `/g` and `/o` path takes its body from the regex group *after* the prefix (`server/game.ts:1560-1568`, `body = (gm ?? om!)[1]`), and `guildChat`/`officerChat` deliver `body.trim().slice(0, 200)` (`server/social.ts:464`). Since `\s+` strips leading whitespace, the routed body can come from raw positions well past 255. So a frame like:

```
/g<253 spaces>zzslur
```

scans as `"/g" + spaces` (no hard word, passes the gate), while `guildChat` delivers exactly `zzslur` to the whole guild. Pre-PR, `findHardHit` scanned the full text and caught this, so it is a regression. Confirmed with a repro at the PR head:

```ts
const padded = `/g${' '.repeat(MAX_CHAT_MESSAGE_LEN + 10)}${HARD}`;
expect(enforce(server, session, padded)).toBe(false); // filter does not flag it
// ...while this is what guildChat would deliver:
const body = /^\/(?:g|gu|guild)\s+([\s\S]+)$/i.exec(padded)?.[1] ?? '';
expect(body.trim().slice(0, 200)).toBe(HARD); // the slur, delivered to the guild
```

The say/world/yell/party/whisper/general/lfg channels are fine: `sim.chat` also slices text-relative at 255 (`text.trim().slice(0, 255)`, then parses), so the filter's slice matches exactly what those channels parse. The mismatch is unique to `/g` and `/o`, which slice the body *after* stripping a variable-length prefix, so a raw-position cap cannot cover them (the stripped whitespace is unbounded). The cap needs to apply to the per-channel body the router will actually deliver, for example by normalizing the leading command and whitespace before slicing, rather than the first 255 chars of the raw frame.

**should-fix** — the new test only exercises the unprefixed say path, which is exactly the path where the cap stays aligned, so the suite reads green while the broken path (guild/officer) is uncovered. Worth adding a `/g` case so the bypass can't pass silently.

jgyy:
Fixed, thanks for the sharp catch and the repro, it made this trivial to verify.

The bug: enforceChatPolicy capped the raw WS frame at MAX_CHAT_MESSAGE_LEN before scanning, but guild/officer chat route a body taken from the regex group after the /g or /o prefix (and its leading whitespace), then cap that separately at GUILD_MESSAGE_MAX in guildChat/officerChat (server/social.ts). Since the stripped prefix/whitespace length is unbounded, padding it past the raw cap let a slur land outside the scanned slice while still being exactly what guildChat/officerChat delivered.

Fix (server/game.ts): enforceChatPolicy now recognizes the same /g and /o prefixes as the chat router (shared GUILD_CHAT_PREFIX/OFFICER_CHAT_PREFIX constants, also used at the existing router call site so there's one source of truth), and when the message matches, scans the trimmed routed body capped at GUILD_MESSAGE_MAX (now exported from social.ts) instead of a raw-position slice of the frame. That matches exactly what guildChat/officerChat will deliver. Every other channel keeps the existing raw-frame cap at MAX_CHAT_MESSAGE_LEN, since sim.chat already slices text-relative there before parsing, so the filter's slice still matches what those channels parse.

Added regression tests in tests/chat_filter_cap.test.ts:
- "flags a /g message whose slur is pushed past the raw cap by padding" and the /o equivalent, using your repro shape
- "does NOT flag a /g slur past the routed guild body cap" to confirm the filter still doesn't burn a scan on bytes that guildChat itself would slice off

All green: npx vitest run tests/chat_filter_cap.test.ts (6 passed) plus the social/snapshot suites, and npx tsc --noEmit clean.

jgyy:
Closeout check: confirmed the blocking review from @FernandoX7 (guild/officer /g and /o chat bypassing the hard-word scan cap) is fully resolved by 003d95ff / a013de15, which now scan the routed guild/officer body (capped at GUILD_MESSAGE_MAX via the shared GUILD_CHAT_PREFIX/OFFICER_CHAT_PREFIX constants) instead of a raw-frame slice. tests/chat_filter_cap.test.ts covers the padded /g and /o bypass repro plus the routed-cap boundary. Re-verified locally: tests/chat_filter_cap.test.ts and tests/chat_filter.test.ts (21/21 passed), tsc --noEmit clean, and gh pr checks all green (Biome changed-files pass, PR gate pass). No further changes needed.

jgyy:

PR #970, opened 2026-06-27
fix(sim): keep the taunt/forced-target window ticking while a mob is stunned

jgyy:
## The bug

`updateMob` (`src/sim/mob/locomotion.ts`) takes an early return when the mob is stunned:

```ts
if (ctx.isStunned(mob)) {
  if (ctx.updateFearMovement(mob)) return;
  // ...polymorph wander...
  return;
}
```

This return happens *before* the `aiState` switch, and the switch is the only place that calls `updateMobTarget`, where the taunt/growl window (`forcedTargetTimer`) is decremented:

```ts
if (mob.forcedTargetTimer > 0) {
  mob.forcedTargetTimer -= DT;
  // ...
}
```

So if a stun lands while a taunt is active, the forced-target window **freezes for the stun's full duration**. A 3s taunt against a mob that gets stunned for 2s effectively lasts 5s. The taunt window is a real-time effect and should keep counting down regardless of crowd control.

## The fix

Extract the timer-tick + expiry into a small pure helper in `mob/targeting.ts`:

```ts
export function tickForcedTarget(mob: Entity): void {
  if (mob.forcedTargetTimer > 0) mob.forcedTargetTimer -= DT;
  if (mob.forcedTargetTimer <= 0) mob.forcedTargetId = null;
}
```

and call it on the stun path. It does **not** touch aggro: a stunned mob cannot act, and when the stun ends the normal `updateMobTarget` path re-applies the forced target if the window is still running.

`updateMobTarget` itself is left byte-identical (it has its own intentional one-extra-tick honoring of the forced target before the clear, which the parity gate pins), so this is purely additive on the stun branch.

## Determinism / parity

Untaunted mobs carry `forcedTargetTimer === 0`, so `tickForcedTarget` is a no-op for them and draws no rng. The parity draw-order is unchanged: `tests/parity` passes 154/154.

## Tests

- Pure unit cases for `tickForcedTarget` (decrement, expiry, no-op) in `tests/mob_targeting.test.ts`.
- An integration case driving a stunned mob through `updateMob` and asserting the window ticks by `DT` and then expires, in `tests/mob_locomotion.test.ts`. Confirmed failing before the fix (timer stayed at 3), green after.

`npx tsc --noEmit`, `tests/architecture.test.ts`, `tests/parity`, and `npm run build` all clean.

## Fix flow (before / after)

```mermaid
flowchart TD
  A["updateMob tick"] --> B{"isStunned?"}
  B -->|"yes (before)"| C["early return, before the aiState switch"]
  C --> D["updateMobTarget skipped, forcedTargetTimer frozen"]
  D --> E["taunt window outlasts the stun (3s taunt + 2s stun = 5s)"]
  B -->|"yes (after)"| F["tickForcedTarget: decrement + expire on the stun path"]
  F --> G["early return"]
  G --> H["taunt window keeps counting down under crowd control"]
  B -->|no| I["aiState switch -> updateMobTarget (unchanged)"]
```


jgyy:
Pinging the maintainers with merge access for review: @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk

Small, parity-safe sim correctness fix: a stun landing mid-taunt froze the forced-target (taunt/growl) window for the stun's duration. Headless logic only (no renderer/UI surface, so no screenshots apply); covered by new unit + integration tests, and `tests/parity` stays 154/154 since untaunted mobs leave the new helper a no-op.

jgyy:

FernandoX7 (approved):

PR #971, opened 2026-06-27
fix(sim): fold debuff_ap into player attack power (PvP)

jgyy:
## Problem

Player **Demoralizing Shout** (and the druid's **Demoralizing Roar**) land a `debuff_ap` aura, but it was a **no-op versus enemy players** (PvP). It worked against mobs and was silently ignored on players.

Root cause is an asymmetry in how attack power folds auras:

- **Mobs:** `effectiveAttackPower` (`src/sim/sim.ts`) folds `buff_ap`/`debuff_ap` from auras live every swing, but the loop is gated `if (e.kind !== 'player')`.
- **Players:** attack power is baked once in `recalcPlayerStats` (`src/sim/entity.ts`), whose aura loop folded every `buff_*` kind but **never `debuff_ap`**. So the aura landed and the buff icon showed, yet the target player's AP never moved.

## Fix (three parts)

1. **Fold `debuff_ap`** in the player aura loop (`entity.ts`).
2. **Floor** the baked `attackPower`/`rangedPower` at 0, so a heavy debuff stack can never bake a negative value. This mirrors `effectiveAttackPower`'s existing mob floor and the agi/spi floors already in `recalcPlayerStats`.
3. **Mark stats dirty on `debuff_ap` expiry** (`combat/auras.ts`). The expiry path only re-ran `recalcPlayerStats` for `buff*`/`form*` kinds; without this the AP cut would persist forever after the aura faded. The mob demoralize affix dodges this by modelling its cut as a negative `buff_ap`, so player Shout is the only `debuff_ap`-on-player consumer and the gap was latent until this fix reached it.

Pure deterministic arithmetic, no new `rng` draw: the golden-trace + draw-order parity gate stays green.

## Screenshots

Level-20 warrior character sheet, before vs after landing the `debuff_ap` aura (with this fix applied). Attack Power **122 -> 92** (-30) and Damage/sec **10.5 -> 8.3**. Before this fix the "after" panel still read 122 (the debuff was inert).

**Before debuff**

before

**After debuff (-30 AP)**

after

## Fix flow

```mermaid
flowchart TD
  A["Demoralizing Shout lands a debuff_ap aura"] --> B{"Target kind?"}
  B -->|mob| C["effectiveAttackPower folds debuff_ap live - works"]
  B -->|"player (before)"| D["recalcPlayerStats folds buff_* only, debuff_ap skipped"]
  D --> E["Player AP unchanged in PvP - the NO-OP bug"]
  B -->|"player (after)"| F["recalcPlayerStats folds debuff_ap, AP floored at 0"]
  F --> G["AP drops in PvP"]
  G --> H["aura expiry marks stats dirty -> recalc"]
  H --> I["AP restored when the debuff fades"]
```

## Tests (test-first, each verified red before the fix)

`tests/demoralizing_shout.test.ts` gains three cases:
- enemy player effective AP drops by the debuff value (PvP),
- AP is restored after the debuff expires,
- AP floors at 0 when the debuff exceeds base AP.

```
npx vitest run tests/demoralizing_shout.test.ts   # 7 passed
npx tsc --noEmit                                   # clean
npx vitest run tests/parity tests/architecture.test.ts \
  tests/combat_auras.test.ts tests/character_sheet.test.ts   # 183 passed
```

### DIFF
--- a/src/sim/combat/auras.ts
+++ b/src/sim/combat/auras.ts
@@ -167,7 +167,10 @@ export function updateAuras(ctx: SimContext, e: Entity): void {
       e.auras.splice(i, 1);
       ctx.applyNonPlayerStatAura(e, a, -1);
       ctx.emit({ type: 'aura', targetId: e.id, name: a.name, gained: false });
-      if (a.kind.startsWith('buff') || a.kind.startsWith('form')) statsDirty = true;
+      // debuff_ap is the one non-buff kind recalcPlayerStats folds, so it must
+      // mark stats dirty on expiry or the AP cut would persist after the fade.
+      if (a.kind.startsWith('buff') || a.kind.startsWith('form') || a.kind === 'debuff_ap')
+        statsDirty = true;
     }
   }
   if (statsDirty && e.kind === 'player') {
--- a/src/sim/entity.ts
+++ b/src/sim/entity.ts
@@ -184,6 +184,10 @@ export function recalcPlayerStats(
   let scaleMul = 1; // Fiesta buff_scale: body-size multiplier (>1 also adds hp)
   for (const a of e.auras) {
     if (a.kind === 'buff_ap') bonusAp += a.value;
+    // Attack-power debuff (Demoralizing Shout/Roar). Mobs fold this live in
+    // effectiveAttackPower; players bake it here, so without this arm the debuff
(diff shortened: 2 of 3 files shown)


jgyy:

PR #973, opened 2026-06-27
fix(pet): notify owner when a stored pet's template no longer exists

jgyy:
## Problem

When a player's saved pet is restored on character load, `restorePet` rebuilds the live
mob from `MOBS[state.templateId]`. If a content update **removed or renamed** that mob
template, the lookup returns `undefined` and the function **silently `return`s**, leaving
the player's pet slot empty with no message, toast, or log line. The player logs in
pet-less and has no idea why.

`src/sim/pet/pet_commands.ts` (the old `sim.ts:3435-3437` guard, since extracted to the
pet command module):

```ts
export function restorePet(ctx: SimContext, owner: Entity, state: PetState): void {
  const template = MOBS[state.templateId];
  if (!template) return;   // <- silent data-loss, no notice
  ...
```

## In-game screenshots

Captured from the **real offline browser world** (the same `Sim` that ships), not a
mockup: `scripts/pet_restore_notice_shot.mjs` drives the actual `Sim.restorePet` with a
saved pet whose template id was removed, so the notice flows through the genuine
`SimEvent` -> `hud.handleEvents` -> `localizeSimText` path and renders in the live chat log.

English chat log:

pet-restore notice, English

German (booted with `?lang=de_DE`, proving the client re-localization works in-engine):

pet-restore notice, German

Full HUD:

full HUD with the notice

### Follow-up: name-free notice when the saved name is unclean

Review nit: when `cleanPetName` rejects the saved name there is no localizable
proper noun, so the old fallback spliced an English "Your pet" that the matcher
left untranslated in a non-English locale. That path now emits a distinct,
name-free sentence that localizes wholesale. German, named vs name-free:

pet-restore notice, German, named

pet-restore notice, German, name-free

The second line ("Dein Begleiter konnte nicht wiederhergestellt werden und ist
verloren.") is fully German, with no embedded English.

## Behavior before vs after

```mermaid
flowchart TD
    A[Character load: addPlayer] --> B[savedState.pet present?]
    B -- no --> Z[normal login]
    B -- yes --> C["restorePet: template = MOBS[state.templateId]"]
    C --> D{template exists?}
    D -- yes --> E[createMob, restore name/hp/mode, addEntity]
    D -- "no (removed/renamed)" --> F1["BEFORE: return silently<br/>empty pet slot, no message"]
    D -- "no (removed/renamed)" --> N{saved name clean?}
    N -- yes --> F2["AFTER: ctx.notice owner<br/>'NAME could not be restored and has been lost.'"]
    N -- "no (unclean)" --> F3["AFTER: ctx.notice owner<br/>'Your pet could not be restored and has been lost.'<br/>(name-free, localizes wholesale)"]
    style F1 fill:#3a1411,stroke:#7a2a22,color:#ffb4ac
    style F2 fill:#0f2a14,stroke:#2f5a2f,color:#9be59b
    style F3 fill:#0f2a14,stroke:#2f5a2f,color:#9be59b
```

## Fix

Replace the silent guard with an owner-facing `ctx.notice`. We still cannot rebuild a pet
whose creature template no longer exists, but the player is now told instead of finding a
silently empty slot. The pet's stored name survives in `PetState`, so we surface it
(falling back to `"Your pet"` when the saved name is unclean).

```ts
  if (!template) {
    const lost = cleanPetName(state.name) ?? 'Your pet';
    ctx.notice(owner.id, `${lost} could not be restored and has been lost.`);
    return;
  }
```

## i18n

The sim stays language-agnostic: it emits English and the client re-localizes. Added the
`log.petRestoreLost` key across **all 14 `sim_i18n` DICT locales** plus the matching
`localizeSimText` RULE (the S3 drift guard scans the extracted pet module, so an
unrecognized emit fails CI). The string is re-rendered per locale via
`tSim('log.petRestoreLost', { name })`:

| Locale | Rendered (pet named "Rex") |
|---|---|
| English | Rex could not be restored and has been lost. |
| Espanol | Rex no pudo ser restaurado y se ha perdido. |
| Francais | Rex n'a pas pu etre restaure et a ete perdu. |
| Deutsch | Rex konnte nicht wiederhergestellt werden und ist verloren. |
| Italiano | Rex non e stato possibile ripristinarlo ed e andato perso. |
| Portugues (BR) | Rex nao pode ser restaurado e foi perdido. |
| 简体中文 | 无法恢复Rex，它已永远消失。 |
| 繁體中文 | 無法恢復Rex，牠已永遠消失。 |
| 한국어 | Rex을(를) 되돌릴 수 없어 잃어버렸습니다. |
| 日本語 | Rexを復元できず、失われました。 |
| Русский | Rex не удалось восстановить, питомец потерян. |

## Tests

Test-first (reproduce, then fix). `tests/pet_commands_module.test.ts` asserts an unknown
template creates **no** pet **and** emits a `log` event to the owner containing the pet
name; added the emit to the S1 localized-sample list so every locale's wording is
exercised.

- `npx vitest run tests/pet_commands_module.test.ts` -> 5 passed
- `npx vitest run tests/localization_fixes.test.ts tests/architecture.test.ts tests/parity` -> green (S3 guard + sim purity + golden-trace parity)
- `npx tsc --noEmit`, `npm run build`, `npm run i18n:gen` -> clean (summary.json regenerated, pending stays 0)
- Full suite: 4321 passed (one unrelated delve-affix test flakes on a 5s timeout under full-suite load; passes in isolation)

Scope is a single guard clause plus its localization; no balance numbers, no rng draws,
no tick-order changes (parity gate stays green).


jgyy:

FernandoX7 (commented):
Solid fix. The silent `return` in `restorePet` was real data loss with zero feedback, and the replacement keeps the same outcome (no pet, since a removed template genuinely cannot be rebuilt) while telling the owner. Scope is tight: one guard clause plus its localization, no rng, no tick-order change, and the early return lands before `createMob`/`nextId++`, so the parity trace is untouched.

What I checked against the code, not just the description:

- The new emit is genuinely under the S3 drift guard. The scanner's `notice(` regex in `tests/localization_fixes.test.ts` captures `ctx.notice(owner.id, ...)`, and the guard reads `src/sim/pet/pet_commands.ts`, so the RULE you added to `localizeSimText` is load-bearing, not decorative: `s3_registered` would fail without it.
- i18n completeness is compiler-enforced. `Record<PetSimMessageKey, string>` makes `tsc` require the key in every DICT, and the regenerated `i18n.status.summary.json` shows `pending: 0` with `translated` up by 20, so no locale shipped English.
- Test-first reproduces the gap (no pet created, plus a `log` event to the owner), and the English string is in the S1 sample so every locale's wording gets exercised.
- The delve-stash restore (`restorePetFromDelveStash`) also routes through `restorePet`, but it stows a live pet whose template exists and `MOBS` is static for the session, so the new notice cannot spuriously fire on a delve exit.

One nit:

- nit, `src/sim/pet/pet_commands.ts:136`: when the saved name fails `cleanPetName`, the fallback `'Your pet'` is English and `locMob` leaves it untouched, so in a non-English locale it renders embedded untranslated ("Your pet konnte nicht wiederhergestellt werden ..."). It needs both a removed template and an unclean saved name, and the success path falls back to the localizable `template.name` instead, so this is borderline. Fine to ship, just flagging it.

Nice touch driving the screenshots through the real `Sim.restorePet` path rather than a mockup.

jgyy:
Thanks, addressed the locale nit you flagged (`server/character_sheet.ts` aside; this was the `'Your pet'` fallback one).

When `cleanPetName` rejects the saved name there is no localizable proper noun to splice, so the old fallback put an English "Your pet" into the templated emit, and the matcher (`locMob` on a non-mob string is a no-op) left it untranslated in a non-English locale. That path now emits a distinct, name-free key (`log.petRestoreLostNoName`), filled across every sim locale; `localizeSimText` resolves it wholesale via the EXACT map, so nothing English survives. The named path is byte-identical.

- New unit case in `pet_commands_module.test.ts`: unclean name -> name-free sentence, recognized by `localizeSimText`.
- Extended `scripts/pet_restore_notice_shot.mjs` with the en/de name-free shots. The German one renders "Dein Begleiter konnte nicht wiederhergestellt werden und ist verloren." (fully German), added to the PR body above.

tsc clean, S3 guard green, `tests/pet_commands_module.test.ts` 6/6.

FernandoX7 (approved):

PR #979, opened 2026-06-27
fix(sim): cap fleeing-mob speed at the player base run speed

jgyy:
## Enemies flee too quickly

A cowardly, low-HP mob panics at `FLEE_HP_THRESHOLD` and runs from its attacker for `FLEE_DURATION` at `FLEE_SPEED_MULT` (1.4x) its base move speed. The design intent is that this flee speed stays **well below the player's base run speed** (`RUN_SPEED`): a chasing player should always be able to close the gap and catch the fleeing mob. Per maintainer guidance on this PR, the ceiling is **65% of `RUN_SPEED`** so the player catches fleeing mobs easily rather than merely matching their speed. The only intended risk is that the flee pulls in social aggro before the kill lands. If the mob can outrun the player, the chase becomes an un-winnable, annoying mechanic.

### The bug

`fleeMoveSpeed` capped the **base projection** and then multiplied the result **again** by `moveSpeedMult(e)`, so the cap was applied in the wrong place. A fleeing mob carrying a haste aura (`buff_speed` / `form_travel`, multiplier > 1) escaped the cap and outran the player:

```ts
// before — the * moveSpeedMult(e) lands OUTSIDE the cap
Math.min(e.moveSpeed * FLEE_SPEED_MULT, FLEE_MAX_SPEED) * this.moveSpeedMult(e)
```

```mermaid
flowchart LR
  A["base moveSpeed"] --> B["x FLEE_SPEED_MULT (1.4)"]
  B --> C["min(_, RUN_SPEED)"]
  C --> D["x moveSpeedMult(e)"]
  D --> E{"haste aura?"}
  E -- "mult &gt; 1" --> F["speed &gt; RUN_SPEED<br/>mob outruns player (BUG)"]
  E -- "mult = 1" --> G["= RUN_SPEED (ok)"]
  style F fill:#d9534f,color:#fff
```

### The fix

Fold the multiplier **inside** the cap, so the final speed the mob actually travels at is bounded by the player base run speed. A slow aura (`mult < 1`) still drags it below the cap; a haste aura can no longer push it above:

```ts
// after — the whole intended speed is capped at 65% of RUN_SPEED
// FLEE_MAX_SPEED = RUN_SPEED * 0.65
Math.min(e.moveSpeed * FLEE_SPEED_MULT * this.moveSpeedMult(e), FLEE_MAX_SPEED)
```

```mermaid
flowchart LR
  A["base moveSpeed"] --> B["x FLEE_SPEED_MULT (1.4)"]
  B --> C["x moveSpeedMult(e)<br/>(slow or haste auras)"]
  C --> D["min(_, 0.65 x RUN_SPEED)"]
  D --> E["flee speed &le; 65% of RUN_SPEED<br/>player always outruns and catches it"]
  style E fill:#4ea36b,color:#fff
```

The flee-speed math is extracted into a pure, host-agnostic leaf `src/sim/flee_speed.ts` with a direct unit test, following the repo's pure-core + thin-consumer convention. `sim.ts` keeps only a thin delegate.

### Behaviour change surface

| case | before | after |
|---|---|---|
| normal fleeing mob (`mult = 1`) | `min(base*1.4, 7)` | `min(base*1.4, 4.55)` — lower ceiling |
| hasted fleeing mob (`mult > 1`) | can exceed `RUN_SPEED` (bug) | clamped to `0.65 * RUN_SPEED` |
| snared fleeing mob (`mult < 1`) | slowed | slowed (still below cap) |

The lower ceiling changes the locomotion of fleeing mobs, so the `mob_locomotion` and `c4b_effect_dispatch` parity goldens were regenerated deliberately; the rng draw order is unchanged.

### Screenshot

Flee speed of a hasted fleeing mob against the 7 yd/s player run speed (yellow) and the new 65% flee ceiling (blue, 4.55 yd/s): the **before** curve escapes above the player (peaks ~9.8 yd/s), the **after** curve clamps to the 65% ceiling so the player always catches it. Captured live from the offline build via `scripts/flee_speed_cap_shot.mjs`.

flee speed cap before vs after

### Tests
- `tests/flee_speed.test.ts` — new pure-unit coverage (cap, haste escape, snare-below-cap, determinism)
- `tsc --noEmit`, `tests/parity`, `tests/sim.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/sim_context.test.ts` all green
- `npm run build` green


ryan-foo:
Much needed, or I would even put it at 50% (wounded)

jgyy:

Rubsey:
Can we make this 65% of default player's movement speed. We want players to be able to catch the mobs easily. There's also a bug right now where mobs aren't socially aggroing upon flee. Possibly for another ticket but that needs to be addressed too

jgyy:
Thanks both! Pushed a follow-up that lowers the cap to **65% of the player base run speed** (4.55 yd/s) per @Rubsey, so players reliably catch fleeing mobs rather than just matching them. @ryan-foo I went with 65% over 50% to keep it readable as a flee while still very catchable; happy to drop it further if you'd prefer.

Regenerated the `mob_locomotion` / `c4b_effect_dispatch` parity goldens for the new ceiling and refreshed the before/after chart (now shows the 7 yd/s player line and the 4.55 yd/s flee ceiling).

On the social-aggro-on-flee note: that's tracked separately, and #991 ("stop fleeing mobs from socially aggroing nearby allies") is the related change there.

PR #926, opened 2026-06-24
feat(actionbar): optional second action bar + on-bar key bindings

nicadeddu:
## Summary

Adds an optional **second action bar**, doubling the available slots from 12 to 24, for classes/roles that run out of room (warrior tank above all). It is opt-in and off by default, so players who want a minimal combat UI are not affected. Works on desktop and mobile.

The design keeps two concepts separate: **Layout** (which spell sits in which slot) is shared across devices, while **Binding** (how a slot is triggered) is per platform (keys on desktop, tap on mobile).

What it does:

- **Second bar:** opt-in toggle, off by default, saved per character (localStorage). Bar 1 stays put; bar 2 stacks with it. Slots go from 12 to 24.
- **Adding spells:** drag from the spellbook on desktop; the `+` button (first free slot) on mobile, then drag to rearrange. Disabling the second bar dismisses its spells back to the spellbook (they show `+` again) instead of stranding them on a hidden bar; the empty-slot search is capped to the enabled bars.
- **Key bindings reworked:** the per-slot rows are removed from the Key Bindings menu. Slots are rebound directly on the bar via an on-bar "key bindings mode" (enter, click a slot, press the combo, repeat, then Done), with a **Reset** behind a confirm (bar 1 back to defaults, bar 2 unbound). The mode is desktop / external-keyboard only and its entry is hidden on touch. Bar 2 ships fully unbound.
- **Action bar size:** new "Action bar size" slider (Interface), 40% to 140%, default 100%, resizing both bars via CSS zoom independently of the global UI scale.
- **Layout/CSS:** bars hug their buttons and stay centered on desktop and mobile (no more full-width dark panel); empty slots stay visible so a bar keeps its width and never collapses when emptied; the Attack-slot hide is scoped to bar 1.

This is entirely client-side: the bar count is UI only (casting is by ability id), and the saved loadout already stores a variable-length bar, so there are **no server or schema changes**. Per-spec switching keeps working through the existing loadout flow.

**Fix included in this PR:** `saveLoadout` used to cap the stored bar at 16 entries (a legacy bound from the single 11-slot bar), which dropped the second bar's last 7 slots (positions 6 to 12) when a build was saved and restored on spec switch. It now caps at `MAX_LOADOUT_BAR_SLOTS` (24 = two 12-slot bars). Covered by a new `tests/talents.test.ts` case.

The pattern is validated by long-standing WoW addons (Bartender4, Dominos) and the native Edit Mode; drag-to-add from the spellbook is native behavior.

[...]

PR #983, opened 2026-06-27
feat(ui): add a second action bar

ryan-foo:
## What
Adds a **second action bar**, stacked just above the primary one.

Second action bar

The top row is the new secondary bar (defaulting to the numpad keys, shown as `Num1`..`Num.`); the bottom row is the unchanged primary bar (Attack + `1`..`=`).

## How
Both rows share a **single `hotbarActions` array** (slot 0 = Attack, 1..11 primary, 12..22 secondary), so every existing behaviour works across both rows with no per-bar bookkeeping: drag/drop (including dragging an ability between the two rows), shift-right-click to clear, mobile long-press drag, per-frame icon/cooldown repaint, keybind-label refresh, and per-class/per-form localStorage persistence.

- `keybinds.ts`: `ACTION_BAR_SLOTS` 12 -> 23; slots 12..22 default to the numpad (`Numpad1`..`Numpad0`, `NumpadDecimal`) so they never collide with the primary number row, and are **fully rebindable** like every other slot.
- `hud.ts`: `BAR_ABILITY_SLOTS` 11 -> 22 (`PRIMARY_BAR_ABILITY_SLOTS` = 11); `buildActionBar` renders slots 0..11 into `#actionbar` and 12..22 into `#actionbar2`; drag-cleanup selectors span both rows; the mobile `many-spells` toggle counts only the primary bar.
- `index.html`: `#actionbar2` row above the primary bar, hidden on `mobile-touch` (the secondary slots stay reachable via their keybinds).
- loadouts: bump the saved `bar` cap 16 -> 23 (`sim.ts` + `online.ts`) so both rows round-trip.

## Tests & verification
- `tests/keybinds.test.ts`: the Action Bar category now has 23 slots with the secondary numpad defaults.
- `scripts/second_action_bar_shot.mjs`: boots the offline world, levels a mage so the kit overflows onto the second bar, and captures the screenshot above (`#actionbar`: 12 buttons, `#actionbar2`: 11, visible). Existing hotbar/keybind/client-shell/progression/talents tests pass; `tsc` clean.


nicadeddu:
seriously? https://github.com/levy-street/world-of-claudecraft/pull/926

ryan-foo:
didn't see it, do yours instead (Y) @nicadeddu

nicadeddu:
> didn't see it, do yours instead (Y) @nicadeddu

It’s impossible to have missed it; I even flagged it on Trello before you did. Instead, I suspect there is some favoritism involved. Looking at your feature, it is less "advanced" than the one I built. I’d like to understand how the decision-makers handle merge choices, given the time spent planning, implementing, and testing, as well as the token usage involved. Yeah, it really frustrates me.

https://trello.com/c/cQ2rstRn

@Rubsey @FernandoX7

PR #985, opened 2026-06-27
feat(server): filter /who by a name substring

ryan-foo:
## What
`/who Mr` now lists only online players whose name contains "Mr" (case-insensitive). Bare `/who` is unchanged (full roster).

## How
- `game.ts`: the `/who` dispatch parses an optional argument (stripped of quotes/spaces/hyphens, capped at 24 chars) and passes it to `sendWhoRoster(session, filter?)`, which filters the roster by `name.toLowerCase().includes(query)`. A filtered run emits its own header — `Who: N players matching "q" on REALM.` — while the unfiltered header is untouched.
- i18n: a new `hudChrome.plurals.playersMatching` key plus a `localizeServerText` rule for the filtered header (server stays language-agnostic; the client re-localizes, same as the existing `/who` header).

## Tests
`tests/who_filter.test.ts`: unfiltered lists everyone; the filter is case-insensitive and matches a substring anywhere in the name; a single result singularizes the header; an unknown filter reports zero matches. The S3 localization guard recognizes the new emit.


ryan-foo:
Superseded: v0.18.0 already filters /who by a name substring (server/game.ts sendWhoRoster(filter), 'players whose name OR zone', plus tests/who_filter.test.ts) — it landed via the stacked #988. Closing as obsolete.

PR #986, opened 2026-06-28
feat(server): add configurable progression rates

hanclintoclaw-pixel:
## Summary
- Add RATE_XP and RATE_DROP_MONEY settings for friendly private servers
- Wire rates through SimConfig so the authoritative sim scales XP awards and mob copper drops
- Keep base quest reward data unchanged for display, applying XP rates only at award time
- Add focused tests for env parsing, XP scaling, copper scaling, and determinism

## Tests
- npx vitest run tests/sim_rates.test.ts tests/rested_xp.test.ts tests/loot_roll.test.ts tests/architecture.test.ts
- npx tsc --noEmit


HanClinto:
I implemented this so that I could add faster progression scaling for my home-network installation of the game (just playing with my family and don't want to grind so long to 20). Testing myself, will let y'all know how it goes.

HanClinto:
image

Testing locally with a value of "3" for both RATE_XP and RATE_DROP_MONEY, and it seems to give a nice feel to the game for casual / private-family servers. Unless people have other asks for this one, I think it's ready for review + merge.

HanClinto:
Example of what it looks like -- I chose to still display the original xp / copper reward amounts, but when it applies, it scales it:

Screenshot 2026-06-29 at 4 25 42 PM
Screenshot 2026-06-29 at 4 25 54 PM

Note that the copper scaling doesn't (currently) apply to quest rewards -- only mob drops.

EnriqueGF:
I think we will close this as we added a Housekeeping area in /admin to configure rates and much more things:
https://github.com/levy-street/world-of-claudecraft/pull/1340

PR #990, opened 2026-06-28
feat(sim): deal projectile damage on impact, not on cast

jgyy:
## What and why

Ranged attacks were **hitscan**: a spell cast, ranged auto-attack, channel tick, or pet bolt emitted its `fx:'projectile'` visual AND resolved its damage in the **same 20 Hz tick**, while the renderer (`src/render/vfx.ts`) flew the bolt at 26 yd/s toward the target. The damage number therefore popped before the bolt visually landed.

This change re-times it so **projectiles deal damage on impact**: the launch site emits the visual now and schedules the whole resolution (hit roll, crit/damage rolls, `dealDamage` / `runEffects`) to run on the tick the bolt reaches the target. Because every `Rng` draw is deferred to the landing tick, a projectile whose caster or target dies or despawns mid-flight **fizzles**: it draws nothing and deals nothing (no overkill credit, no loot/XP from a corpse that is already gone).

The sim bolt **homes on its live target**, exactly like the renderer: each tick it steps 26 yd/s toward the target's CURRENT position and resolves on the tick it comes within reach, so a target that moves during flight takes its damage in step with the visual impact (see "Moving targets" below).

before/after timeline

The chart drives the real offline `Sim`: a level-20 mage casts Fire Blast at a dummy 20 yd away. The damage (77) lands on tick 16 (0.80s, the real bolt flight), not tick 0.

## Before vs after

Before (hitscan, the bug):

```mermaid
flowchart LR
  A[cast / swing tick] --> B[emit projectile visual]
  A --> C[hit roll + damage NOW]
  C --> D[target HP drops on the cast tick]
  B -. bolt still in flight .-> E[bolt reaches target later, visual only]
```

After (resolve on impact, homing on the live target):

```mermaid
flowchart LR
  A[cast / swing tick] --> B[emit projectile visual]
  A --> C["scheduleProjectile: queue a bolt at the caster position"]
  C --> F{each tick: step the bolt 26 yd/s toward the target LIVE position}
  F -->|within reach this tick| G[hit roll + damage on impact]
  F -->|still in flight| F
  F -->|caster/target gone| H[fizzle: no draw, no damage]
  F -->|deadline reached, never caught| G
```

## Moving targets (review follow-up)

Thanks to @Rubsey for catching the moving-target case. The first cut scheduled the impact once from the launch distance (`at = time + dist / 26`), but the renderer homes the bolt on the target's live position every frame, so a target that moved during flight could take damage before or after the visual landed.

The sim now models the projectile as a moving thing, mirroring `vfx.ts`: `advancePendingProjectiles` steps each bolt `PROJECTILE_SPEED * DT` yards toward the target's current position every tick and resolves on the tick it lands (the `Math.max(PROJECTILE_REACH, step)` arrival test mirrors the renderer's `Math.max(0.7, step)`). A target running directly away lands its damage later; running in lands it earlier; a target kiting at or above projectile speed can never be physically caught, so the bolt lands by force at the `PROJECTILE_MAX_FLIGHT` deadline (a released projectile cannot be escaped, see the follow-up comment below).

## How

- Pure leaf `src/sim/projectile_travel.ts`:
  - `PROJECTILE_SPEED = 26` (kept in step with `render/vfx.ts`), `PROJECTILE_REACH`, and `PROJECTILE_MAX_FLIGHT`.
  - A `PendingProjectile` carries the bolt's live `(x, z)` position; `stepProjectile` (pure, unit-tested) advances it one tick toward the target and reports whether it arrived.
  - The queue is a Sim field exposed as an append-only `SimContext` primitive view, mirroring `delayedEvents`; it is transient (not serialized).
  - `advancePendingProjectiles` runs in the tick prologue (after `runDespawnDecay`) on a fixed, deterministic phase boundary, so a bolt always lands at least one tick after it is fired.
- All four hitscan sites defer through it: `applyAbility` (single-target spell), `applyChannelTick` (Arcane Missiles), `rangedSwing` (Auto Shot / Wand), `petRangedAttack` (imp bolt).
- Boss telegraphs (Nythraxis Soul Rend, wardstones) already time-separate their damage from the `fx:'projectile'` visual, so they are deliberately untouched.

## Determinism and parity

Deferring damage moves the damage event and its `Rng` draws to a later tick, so this intentionally forks the parity goldens that cast/shoot a projectile. The homing re-time also nudges a couple of ranged-bolt landing ticks by one; the affected goldens were regenerated via `UPDATE_PARITY=1` (the casting/effect/auto-attack/pet/warlock-pet scenarios from the first cut, plus `c5_auto_attack` and `pet_ai` for the homing re-time), and the diffs show exactly the expected shift (rolls move off the cast/swing tick to the landing tick). The gate re-records each golden twice, so determinism is still proven.

## Tests

- `tests/projectile_travel.test.ts`: pure `stepProjectile` homing math, scheduling + advance, fizzle-on-death/despawn, and the two moving-target cases @Rubsey asked for (target running away lands LATER, target running toward the caster lands EARLIER), a target that outruns the bolt landing by force at the flight deadline, plus an end-to-end Fire Blast that deals no damage the instant it is cast but lands it ~14 ticks later (and a same-seed determinism check).
- Existing combat/CC/duel/arena/threat tests updated to advance to the landing tick (the projectile now applies its effect a few ticks after the cast bar empties). Polymorph/Fear DR tests retry on a missed bolt, since a miss no longer consumes a DR stage.
- `npm test` green, `npx tsc --noEmit` clean, `tests/architecture.test.ts` (sim purity) green.


jgyy:

Rubsey:
Thanks for the detailed writeup. The direction makes sense for the stationary-target version of the issue, but I think there is still a moving-target mismatch.

The sim schedules the impact time once from the launch distance:

- `src/sim/projectile_travel.ts`: `at = ctx.time + projectileTravelTime(dist2d(source.pos, target.pos))`

The renderer, though, homes the projectile toward the target's live position every frame:

- `src/render/vfx.ts`: `const target = this.anchor(pr.targetId, 0.5)` inside the projectile update loop

So if a target moves during flight, damage can still land before or after the visual impact. Example: target starts 18 yd away and runs directly away at `RUN_SPEED = 7`. The sim schedules impact at `18 / 26 = 0.69s`, but the visual catches at roughly `18 / (26 - 7) = 0.95s`, around 5 ticks later. If the target runs toward the caster, the visual arrives earlier than the scheduled sim impact.

I would expect the sim-side pending projectile to track its own position and advance toward the live target each tick, resolving when it reaches the target, rather than storing only a fixed launch-time `at`. It would also be worth adding tests for a target moving away and moving toward the caster during flight, since the current projectile tests appear to cover static timing, fizzle on death/despawn, and determinism, but not target movement.

jgyy:
@Rubsey good catch, you were right about the moving-target mismatch. I pushed a follow-up that makes the sim bolt home on the target's live position instead of baking in a fixed launch-time impact.

Instead of storing `at = time + dist / 26`, a `PendingProjectile` now carries its own `(x, z)` and each tick `advancePendingProjectiles` steps it `PROJECTILE_SPEED * DT` yd toward the target's CURRENT position, resolving on the tick it comes within reach. The arrival test (`Math.max(PROJECTILE_REACH, step)`) mirrors `vfx.ts`'s `Math.max(0.7, step)`, and a bolt that can't catch a target kiting at or above projectile speed fizzles after `PROJECTILE_MAX_FLIGHT` seconds, matching the bolt's ttl in the renderer.

For your exact example (18 yd, target fleeing at 7 yd/s): the bolt now lands at ~18 / (26 - 7) = ~0.95s instead of the old fixed 18 / 26 = 0.69s, in step with the visual.

Added the two tests you asked for plus a runaway-fizzle case:
- target running directly away lands LATER than a static schedule,
- target running toward the caster lands EARLIER,
- target outrunning the bolt fizzles (no damage).

The pure `stepProjectile` helper is unit-tested directly, and the two ranged-bolt goldens whose landing tick shifted (`c5_auto_attack`, `pet_ai`) were regenerated. Full suite + tsc + the sim-purity guard are green. The before/after chart and the "after" mermaid in the description are updated for the homing model.

jgyy:
Updated before/after timeline for the homing model (driven by the real `Sim`): Fire Blast at 20 yd now lands on tick 16 (0.80s) as the bolt reaches the target.

before/after timeline

jgyy:
## Follow-up: a released projectile cannot be escaped

Per the request: **once a projectile is released there is no escaping it; the only way to avoid it is to be out of the player's cast range when it fires.**

The fire-time range gate was already there at every launch site (`applyAbility` / `applyChannelTick` in `casting_lifecycle.ts`, `rangedSwing` via the auto-attack range check in `auto_attack.ts`, and the pet path), so a target that leaves cast range before the bolt fires never gets a bolt. The change here removes the *after-launch* escape: previously a target kiting at or above `PROJECTILE_SPEED` (which the homing can never physically catch) made the bolt fizzle at the `PROJECTILE_MAX_FLIGHT` deadline. Now the bolt **lands by force** at that deadline instead. A bolt still fizzles only when its caster or target dies or despawns mid-flight, so no damage or kill credit ever lands on a corpse.

This supersedes the "a target kiting at or above projectile speed never gets hit and the bolt fizzles" line in the **Moving targets** section above.

```mermaid
flowchart LR
  A[cast / swing tick] --> R{target within cast range?}
  R -->|no| X[do not fire: no projectile]
  R -->|yes| B[emit projectile visual + scheduleProjectile]
  B --> F{each tick: step the bolt toward the target LIVE position}
  F -->|within reach| G[hit roll + damage on impact]
  F -->|caster/target dead or despawned| H[fizzle: no damage]
  F -->|deadline reached, never caught| G
  F -->|still in flight| F
```

Proof (drives the real `projectile_travel` homing logic in Node): a target kiting directly away at `PROJECTILE_SPEED + 5` yd/s grows the bolt-to-target gap the whole flight, then the bolt lands by force at the 3s deadline (tick 61) rather than escaping.

released projectile cannot be escaped

Tests: `tests/projectile_travel.test.ts` (the former "outruns the bolt -> fizzle" case now asserts it **lands**). `npm test` green, `npx tsc --noEmit` clean, `tests/architecture.test.ts` and `tests/parity/parity.test.ts` green (no golden regen: the deadline path is only hit by a kiting target, never the standing dummies in the goldens).

PR #991, opened 2026-06-28
fix(sim): stop fleeing mobs from socially aggroing nearby allies

jgyy:
## What and why

A cowardly mob that panics at low HP should not fight to the death. It runs, and the
question this PR settles is what it does as it runs.

The right behavior: it runs away from its attacker, looks each tick for the first LOCAL
same-family ally within 5yd, and the instant it reaches one it turns back to re-engage
WITH that ally. If it never finds anyone, it recovers its nerve and returns on its own.

The trap to avoid is a per-tick rally that never stops. Because wild mobs are camped
close together, a fleer threading its camp would chain in every same-family ally it ran
past, turning one pull into a whole-camp wipe. That is the **Incorrect** panel below. The
fix keeps the rally LOCAL by ending the flee on first contact: the fleer pulls one
cluster, then heads back the way it came, so it never reaches the rest of the pack. That
is the **Correct** panel.

Incorrect vs correct: a fleeing mob runs to its first local ally and returns with it, it does not chain across the pack

## The change

- `src/sim/mob/locomotion.ts`: the flee arm now runs a per-tick 5yd scan
  (`rallyFleeingAllies`). The instant it returns a non-empty rally, the fleer recovers
  (turns back to re-engage) WITH that local cluster.
- `src/sim/sim.ts`: `maybeFlee` no longer rallies at the panic spot (that was the old
  one-shot pull); the now-unused import is dropped.
- `src/sim/mob/social_aggro.ts`: documented as the per-tick, first-contact local rally;
  `FLEE_HELP_RADIUS` stays 5. It is called directly from locomotion (sibling import), not
  via the `SimContext` seam.
- Determinism: the rally only mutates entity state and seeds threat, drawing **zero rng**,
  so the shared draw order is unchanged and the `tests/parity` goldens need no
  regeneration.
- Pure sim AI: flee + aggro are server-authoritative entity state mirrored to clients via
  snapshots, so there are **no** `IWorld` / `ClientWorld` / wire / server changes.

### Incorrect (the rejected per-tick rally that never stops)

```mermaid
flowchart TD
  A[Mob HP drops below 20%] --> B[enter flee and run]
  B --> C[every tick: rally same-family allies in range]
  C --> D{new idle ally beside the fleer?}
  D -- yes --> E[ally chained into the fight]
  E --> C
  D -- no --> F[keep running]
  E -.cascades across the camp.-> G[whole pack engaged: too punishing]
```

### Correct (run out, return with the first local ally)

```mermaid
flowchart TD
  A[Mob HP drops below 20%] --> B[maybeFlee: enter flee, run away from the attacker]
  B --> C[each tick: scan 5yd for an idle same-family ally]
  C --> D{local ally within 5yd yet?}
  D -- not yet, flee window left --> E[keep running away]
  E --> C
  D -- yes --> F[rally that local cluster: they chase the attacker]
  F --> G[fleer recovers nerve, turns back to re-engage WITH them]
  G --> H[heads back the way it came: allies further down the lane are never reached, stay idle]
  D -- window expired or leash edge --> I[recover on its own and return, no ally found]
```

## Tests

- `tests/mob_flee.test.ts`: a fleer runs to the first local same-family ally, rallies it,
  and turns back to fight (no longer in `flee`). A second test pins the no-chain
  guarantee: an ally 6yd down the lane joins, one 40yd down stays idle because first
  contact ends the flee.
- `tests/social_aggro.test.ts`: local-cluster rally (inside 5yd joins, beyond stays idle),
  returns the pulled count, `FLEE_HELP_RADIUS` pinned at 5, deterministic.
- `scripts/flee_social_aggro_chart.mjs`: derives the out-and-back flee path and the
  first-contact rally from the real sim (correct 1/5 vs incorrect 5/5).
- `tests/parity` green (goldens unchanged, the rally draws no rng), `tests/architecture`
  green, `tsc` clean, and the S3 i18n guard green (the flee emit text is unchanged).

`docs/screenshots/flee-social-aggro.png` is the hand-drawn incorrect/correct comparison.
`scripts/flee_social_aggro_chart.mjs` is the runnable, source-of-truth derivation of it.


jgyy:

Rubsey:
So we actually want this @jgyy , the counter play is to make sure that mob run speed isn't too quick, which is a change you already have up

Rubsey:
For this one, the issue is that social aggro isn't actually working. So mobs can flee, but if they run through their allies, they don't pull.

What should happen here is that when a mob runs past an ally, it should socially aggro back in with that ally immediately.

jgyy:
updated the aggro logic as mentioned check updated diagram

jgyy:
Reworked this to the **local one-shot** model from the design diagram (Incorrect vs Correct).

The previous commit rallied allies **every tick as the fleer ran past**, which is the **Incorrect** panel: a fleer threading a tight camp chained the pull across the whole pack (8yd, re-scanned every tick), so one pull cascaded into a wipe.

This commit makes the rally tight and one-shot, matching the **Correct** panel:
- `FLEE_HELP_RADIUS` 8 -> **5yd**.
- The rally fires **once**, at the panic spot (in `maybeFlee`), not per-tick. The per-tick `ctx.rallyFleeingAllies` call is removed from the flee arm in `mob/locomotion.ts`, so allies down the escape lane stay idle ("no chain") as the fleer runs past them.
- With no remaining caller, `rallyFleeingAllies` comes off the `SimContext` seam; `maybeFlee` calls the module directly.

Still zero rng in the rally, so the parity goldens are untouched. Tests inverted to assert no chaining (`mob_flee`) plus a local-cluster + radius pin (`social_aggro`). The screenshot is the hand-drawn diagram; `scripts/flee_social_aggro_chart.mjs` is the runnable derivation (incorrect 5/5 vs correct 2/5).

PR #980, opened 2026-06-27
feat(sim): item set bonuses for caster, strength, and agility gear

ryan-foo:
## What

Adds a classic-style **item set** system: equipping enough pieces of a set grants stacking 2- and 3-piece bonuses. All bonuses resolve through the one true derivation function (`recalcPlayerStats`), so they apply identically offline, on the server, and on the character sheet.

### Sets and bonuses
| Set | 2-piece | 3-piece |
|---|---|---|
| Vestments of the Whispering Dark (cloth caster, rare/epic) | 50% cast-pushback reduction from damage | 100% (immune to cast pushback from damage) |
| Ironhide Battleplate (mail, strength) | +20 attack power | +8 Strength, +8 Stamina |
| Duskstalker Pelt (leather, agility) | +20 attack power | +8 Agility, +1% crit |

The caster bonus directly addresses "cast timer interrupt reduction": `Sim.pushbackCast` now scales the damage-driven pushback by the equipped reduction (`1.0` = fully immune).

## How
- `types.ts`: `ItemDef.set`, the `SetBonusEffect`/`SetBonusTier`/`ItemSet` shapes, and a derived `Entity.castPushbackReduction` (0..1).
- `content/item_sets.ts`: the three sets (three rare/epic pieces each) plus the pure `aggregateSetBonuses` resolver. Stat/AP/crit effects sum across met tiers; pushback reduction max-combines (never sums past 1). Balance numbers live here, not inline in engine code (per `sim/CLAUDE.md`).
- `entity.ts` (`recalcPlayerStats`): counts equipped set pieces in the existing gear loop and folds the aggregate in. Flat primary stats join the gear totals so they feed every derivation (AP/crit/HP); flat AP/crit add at their steps; pushback reduction is stashed on the entity.
- `sim.ts` (`pushbackCast`): scales pushback (and channel pushback) by `1 - castPushbackReduction`.
- `data.ts`: merges the set items into `ITEMS` and re-exports `ITEM_SETS`/`aggregateSetBonuses`.

## Tests
`tests/item_sets.test.ts`: pure-resolver thresholds (below 2pc = nothing; tiers stack; pushback max-combine), `recalcPlayerStats` integration for all three sets, and `pushbackCast` scaling at 0/50/100% for both casts and channels. Architecture purity guard, character-sheet, gear, stat-tooltip, snapshot, and i18n guards all still pass.

## Determinism & invariants
[...]

PR #994, opened 2026-06-28
feat(sim): haste (melee, ranged, spell) via item-set bonuses (T2 3pc + leveling haste kits)

ryan-foo:
Adds a haste stat through the existing ITEM_SETS set-bonus framework (reworked per review; the original parallel setId/HASTE_SETS system is deleted).

Mechanic
- SetBonusEffect gains one optional `haste` field (fraction; SET_HASTE_3PC = 0.15 in content/item_sets.ts). aggregateSetBonuses sums it and recalcPlayerStats folds the aggregated value into Entity.meleeHaste / rangedHaste / spellHaste (one stat drives all three channels).
- Spell haste divides cast time and channel duration + tick interval by (1 + spellHaste); the tick count is unchanged and the GCD is deliberately NOT hasted. Melee/ranged haste divide the respective auto-attack swing interval. Mobs and pets stay at 0.
- Every tier-2 3-piece bonus (Crownforged, Nighttalon, Soulflame, Stormcaller's) now also grants the 15% haste on top of its existing stats, so the Nythraxis-tier sets carry the speed reward. Tier-1 sets are unchanged.
- Three leveling haste kits assembled from existing world-drop items grant haste alone at 3 pieces: Vale Arcanist's Regalia (cloth), Boundstone Vanguard (mail), Greyjaw Stalker's Kit (leather). Members are tagged with `set`, so the item tooltip set block and piece count render for free.

UI
- Ability tooltips and the spellbook summary show the real, hasted cast/channel time (abilityCastLine divides by 1 + spellHaste). Works online too: the server self-frame sends `sh: p.spellHaste` and ClientWorld decodes it, mirroring `sp`.

i18n
- Changed tier-2 bonus lines and the new kit names/bonus lines are translated in all 17 locale overlays; the catalog and entity_i18n register only tiers a set actually has. Regenerated i18n artifacts committed.

Determinism: haste only scales already-computed durations and draws no RNG; with no set equipped it is 0 and the parity goldens are byte-identical (159 pass).

Tests: tests/haste_set_bonus.test.ts (kit definitions, resolver, recalc derivation, T2 haste vs T1 none, cast/channel/swing math), tests/item_sets.test.ts and i18n coverage counts updated. Verified in the live offline client via scripts/haste_set_shot.mjs.


Rubsey:
Hey, thanks for this. The haste mechanic itself is solid: the cast/channel/swing math is right, the GCD is correctly left un-hasted, channels keep their tick count while shortening, and it's deterministic (parity is byte-identical at haste 0). That part is a keeper.

The one structural thing: this PR builds its own item-set framework (`setId` + `HASTE_SETS` + `ItemSetDef`), but #980 (your item-set bonuses PR) merged into the release right after you opened this, so `release/v0.18.0` already has the framework: `ItemDef.set` + `ITEM_SETS` + `SetBonusEffect` + `aggregateSetBonuses`, already wired into the item tooltip and i18n. So we've ended up with two parallel set systems. Could you fold haste into the existing one instead?

Concretely:
- Add `meleeHaste?` / `rangedHaste?` / `spellHaste?` to `SetBonusEffect`, and sum them in `aggregateSetBonuses` (plus `AggregatedSetEffect` / `zeroEffect`).
- Define the three haste sets as `SetBonusTier` entries in `ITEM_SETS` (with a `text` line, same as your Soulflame set in #1065), and tag the member items with `set` instead of `setId`.
- Read the haste off the `setEff` that `recalcPlayerStats` already computes, then delete `haste_sets.ts`, `ItemSetDef` / `ItemSetBonus`, the `setId` field, and the second piece-count loop.

Nice side effect: the haste sets currently show no set-bonus block on their item tooltips (they bypass the `item.set` path), so players get no in-game signal the set exists. Folding into the framework fixes that and the i18n for free.

One more, separate from the above: the hasted-cast-time tooltip (second commit) only works offline. `spellHaste` isn't synced to the client the way `spellPower` is, so online casters see the un-hasted number even though the real cast bar is correct. Mirror it like `sp`: add `sh: p.spellHaste` in the server self-frame and `e.spellHaste = s.sh ?? 0` in the client decode. (Melee/ranged haste don't need wiring; swing timers already ride the snapshot.)

Minor content nits, optional: the "Boundstone Vanguard" pieces are `mail`, not plate (the comment says plate), and `mistveil_grips` is a cloth item sitting in the leather "Greyjaw Stalker" set.

Note I rebased this onto `release/v0.18.0` and retargeted the PR so it's mergeable and green (the `HASTE_SETS` name you'll see is just a rebase artifact to avoid colliding with `ITEM_SETS`, which is itself the tell that these are the same concept). Thanks again, the mechanic is good, this is just plumbing it into what we've got.

ryan-foo:
Reworked as requested, rebased onto release/v0.20.0 and retargeted (the old base had drifted and gone CONFLICTING).

Folded into the existing framework:
- haste_sets.ts, ItemSetDef / ItemSetBonus, the setId field, and the second piece-count loop are gone. SetBonusEffect gains ONE optional haste field (per Ryan's follow-up: a single haste stat, not the melee/ranged/spell triple), aggregateSetBonuses sums it (AggregatedSetEffect / zeroEffect extended), and recalcPlayerStats folds the one aggregated value into meleeHaste / rangedHaste / spellHaste. The cast/channel/swing application sites are unchanged from what you reviewed.
- The three haste sets are now plain ITEM_SETS entries (single 3-piece tier, text line included) and their members carry `set`, so the item tooltip set block and the i18n come along for free, as you noted.
- On top of that, every tier-2 3-piece bonus (Crownforged, Nighttalon, Soulflame, Stormcaller's) now also grants the 15% haste on top of its existing stats, so the Nythraxis drops carry the speed reward. Tier-1 sets stay haste-free.

Content nits:
- mistveil_grips (cloth) is out of Greyjaw Stalker's Kit; trail_leggings (leather, unrestricted) completes it. Each kit is 3 distinct slots, uniform armor type.
- The plate comment is gone with haste_sets.ts; Boundstone Vanguard is described as mail everywhere now.

Online tooltip: server self-frame carries sh: p.spellHaste and ClientWorld decodes e.spellHaste = s.sh ?? 0, mirroring sp, so online casters see the hasted number. Melee/ranged left unwired per your note.

i18n: the four changed tier-2 bonus3 lines and the three new set names + bonus lines are translated in all 17 overlays; the catalog and entity_i18n only register tiers a set actually has, so the single-tier kits do not emit an id-fallback bonus2 row. Regenerated status/resolved artifacts committed.

One deliberate call to flag: Slam is the only physical-school cast-time ability, and its cast rides p.spellHaste like every other cast. Under the single-stat design meleeHaste always equals spellHaste, so the classic melee-haste scaling falls out identically; there is a comment at the division site saying to switch physical casts to meleeHaste (and wire mh) if the channels ever split.

Gates: tsc clean, parity byte-identical (159), architecture, haste_set_bonus + item_sets, the S3 guard, i18n coverage/completeness/status all green; biome ci:changed clean.

ryan-foo:
Rebased onto the moved release/v0.20.0 tip (the conflicts came from the W1 IP-pivot rename sweep touching item_sets.ts, plus the generated i18n.status.summary.json).

Resolution: the LOCKED NAME-MAP display names win everywhere (Barrowlord Battlegear, Nightfang Vestments, Mournweave Raiment, Bonewrought Regalia, Direfang Pelt, Wraithfire Regalia, Galecall Vestments) on top of this PR's T1/T2 bonus split, so the four tier-2 3-piece bonuses keep their haste under the new names. Set ids are frozen, so the item `set` tags, the haste kit definitions, and the per-id overlay translations are unaffected; the kit member items were not renamed by the sweep, so the three kit names still match their pieces. The generated summary was regenerated, not hand-merged.

Re-verified on the new base: tsc clean, parity byte-identical, architecture, haste + item-set suites, S3 guard, i18n coverage/completeness/status registry, guide freshness, snapshots/bandwidth all green; biome ci:changed clean. Mergeable again.

PR #996, opened 2026-06-28
fix(sim): allow targeting your own dead pet so its menu stays usable

jgyy:
## Problem

When a hunter logs in with a pet that died before logout, the dead pet appears
beside the player but **cannot be selected**. `targetEntity` rejects every dead
entity that is not a lootable corpse, and a pet is dead-but-not-lootable. Because
the pet context menu (Rename / **Revive** / **Abandon**) lives on the target
portrait and only opens for the current target, a wiped pet could never be
revived or abandoned. It was stuck, permanently uninteractable.

## Root cause

`Targeting.targetEntity` (the authoritative `Sim`) and `ClientWorld.targetEntity`
(the online optimistic mirror) both gate on the same rule:

```ts
if (!e || (e.dead && !e.lootable)) return;   // dead + unlootable  ->  not selectable
```

A player's own dead pet falls into exactly that hole.

```mermaid
flowchart TD
    A[Click / right-click own dead pet] --> B[world.targetEntity id]
    B --> C{dead && !lootable?}
    C -- "before: yes -> return" --> X[targetId stays null]
    X --> Y[Right-click target frame finds no target]
    Y --> Z[Pet menu never opens: Revive/Abandon unreachable]
    C -- "after: own pet -> allow" --> S[targetId = pet]
    S --> M[Pet menu opens: Rename / Revive / Abandon]
```

## Fix

A small pure predicate decides whether a **dead** entity stays selectable, and
both worlds route through it so server and client never disagree:

```ts
// src/sim/dead_target.ts
export function deadTargetSelectable(e: Entity, viewerId: number): boolean {
  if (e.lootable) return true;                       // lootable corpse (unchanged)
  return e.kind === 'mob' && e.ownerId === viewerId; // your own pet
}
```

```mermaid
sequenceDiagram
    participant P as Player
    participant W as Sim / ClientWorld
    participant D as deadTargetSelectable
    participant H as HUD (target portrait)
    P->>W: targetEntity(deadPetId)
    W->>D: deadTargetSelectable(pet, viewerId)
    D-->>W: true (own pet)
    W-->>W: targetId = pet (autoAttack stays off, pet is non-hostile)
    P->>H: right-click target frame
    H-->>P: Rename / Revive / Abandon Pet
```

- `src/sim/dead_target.ts` - new host-agnostic leaf predicate (sim-pure: no DOM/Three/rng).
- `src/sim/targeting.ts` - authoritative `Sim` path routes through it.
- `src/net/online.ts` - `ClientWorld` optimistic mirror routes through it (parity).
- A non-hostile dead pet still never flips on auto-attack; lootable corpses and
  other players' pets behave exactly as before.

## Tests

`tests/dead_target.test.ts` (test-first - the integration case failed before the
change, for the right reason):

- predicate: own pet yes; another player's pet no; unlootable/unowned corpse no;
  player corpse no; lootable corpse yes.
- `Targeting.targetEntity`: selects your own dead pet (targetId set, autoAttack
  stays off); does **not** select another player's dead pet.

`npx tsc --noEmit`, `tests/architecture.test.ts` (sim purity), and the
targeting / pet / snapshot suites are green. No new player-facing strings, so no
i18n changes.

## Visual proof

The dead pet ("WOLF / Dead" on the target portrait) is now selectable, with the
pet context menu open. `scripts/target_dead_pet_shot.mjs` reproduces it.

Dead pet targeted with Revive/Abandon menu open

---

cc maintainers with merge access for review: @levy-street @FernandoX7 @Rubsey


jgyy:
Pinging maintainers with upstream push access for review: @levy-street @FernandoX7 @Rubsey

Small, isolated fix (one pure predicate, both worlds routed through it, test-first). Self-contained dead hunter pets are currently impossible to revive or abandon after login; this restores it. Happy to adjust if you'd prefer the predicate to live elsewhere or want the rule widened (e.g. friendly party members' dead pets).

jgyy:

PR #998, opened 2026-06-28
feat(items): item level system with stat-budget normalization

ryan-foo:
## What

Adds an **item level** system: a single number derived from *where an item drops*, the stat budget an item of that level should carry, and an opt-in tooltip readout. Off by default, so the classic look is unchanged unless a player turns it on.

## How it works

New pure, host-agnostic leaf module `src/sim/item_level.ts` (no rng, no Sim state, no DOM, so the HUD imports it directly the way the renderer imports `terrainHeight`). It exposes:

- `itemSourceLevel(id)` / `itemLevel(item)` - the drop level (top of the dropping mob's band, or the boss a quest reward is gated behind) plus a rarity bonus (rare +3, epic +6, ...).
- `primaryStatBudget(level, quality, slot)` - the total primary-stat points an item of that level/quality/slot should grant (quality x slot x level). Whites carry no stat budget; smaller slots weigh below chest/main-hand, matching the existing armor slot weighting in `items.ts`.
- `itemScore(item)` - a single comparable power number: primary stats + converted armor + converted weapon DPS.
- `normalizePrimaryStats(stats, budget)` - redistributes a budget across whichever attributes an item already uses, preserving its stat identity and keeping the integer sum exact (largest-remainder, deterministic).

## UI

An off-by-default **"Show Item Level"** toggle in Esc > Interface. When on, every item tooltip gains an `Item Level N` line plus a power `Score`. Read live, so toggling takes effect on the next hover; sourceless items (vendor/starter/junk) simply omit the line.

## Showcase: same-tier normalization

The Hollow Crypt rare trios are normalized to their item-level budget, demonstrating that items from the same place carry the same total power while keeping their own identity:

Chest trio (chapel level-7 elites -> item level 10 -> 6-point budget):
- Hollowbone Hauberk: +3 str, +3 sta (plate, armor 105)
- Gravewoven Raiment: +3 int, +3 spi (cloth, armor 38)
- Cryptstalker Jerkin: +4 agi, +2 sta (leather, armor 65)

Weapon trio (q_hollow reward for slaying Morthen, level 10 -> item level 13 -> 7-point budget):
- Gravecaller's Broadblade: +4 str, +3 sta
- Widowfang Dirk: +4 agi, +3 sta
- Staff of the Hollow: +5 int, +2 spi

`tests/item_level.test.ts` pins data == formula (each showcase item's stat sum equals its computed budget, and each trio shares one item level and one budget).

## i18n

New `hudChrome.options` keys (toggle + tooltip lines) with locale-overlay translations (all 5 non-Latin locales plus the major Latin bases); resolved tables and the `i18n.resolved.sha256` baseline regenerated.

## Tests

- New `tests/item_level.test.ts` (16 cases): source derivation, tier number, budget formula, showcase normalization, `normalizePrimaryStats`, `itemScore`, purity/determinism.
- `tests/architecture.test.ts` (sim purity/determinism), the i18n completeness + resolved-equivalence + status-registry freshness gates, and `tsc --noEmit` all pass. Full suite green (one unrelated `delves` test is flaky under full-suite parallel load; it passes in isolation).


ryan-foo:
## Follow-up: level-20 balancing + screenshots

Two additions since the initial commit (see commits `item-level balance the level-20 endgame` and `screenshot script + proof shots`):

### Level-20 endgame, item-level balanced
Extended the model so a 10-player raid encounter confers a +3 item-level bonus (derived from each dungeon's `suggestedPlayers`, no new mob flag). The Nythraxis raid set now reads a tier above the level-20 Korzul 5-player dungeon set instead of collapsing to the same item level. Then **all 43 level-20 gear pieces** (uncommon -> legendary) were normalized to the stat budget their item level earns, preserving each piece's stat identity and armor. The previously wild epic-helmet spread (16-28 primary-stat points) and over/under-statted slots now follow one slot-weighted curve; same (item level, quality, slot) groups carry an identical primary-stat sum. `tests/item_level.test.ts` pins "every level-20 item is on budget" (43 pieces) plus the raid-tier and same-tier-equality gates.

### Screenshots (real offline client, via `scripts/item_level_shot.mjs`)

Interface toggle (off by default), shown ON:

interface toggle

Tooltips across the level-20 tiers, with the readout on:

| Raid epic (ilvl 29) | Dungeon epic (ilvl 26) | Legendary (ilvl 33) |
|---|---|---|
| raid | dungeon | legendary |

The raid helmet reads Item Level 29, a tier above the dungeon chest's 26. Before / after the toggle on the same item:

| Toggle off | Toggle on |
|---|---|
| off | on |

PR #999, opened 2026-06-28
feat(sim): Wolf Form swings at the rogue cadence, not the carried staff

jgyy:
## What and why

A feral druid in **Wolf Form** (internally the `form_cat` shapeshift) fights with
its claws, but the auto-attack swing timer still keyed off the **equipped weapon's
speed**. A druid carrying a slow caster staff (2.9s) therefore swung far slower
than intended in melee form. Classic feral forms ignore the carried weapon and use
a fixed normalized cadence; this PR makes Wolf Form swing at the **rogue's base
speed** so a feral druid attacks as fast as a rogue.

## The fix

A new pure leaf, `src/sim/combat/form_swing.ts`, exposes `baseSwingSpeed(e)`:

- Wolf Form (`form_cat`) returns `ROGUE_BASE_SWING_SPEED`, sourced from
  `ITEMS[CLASSES.rogue.startWeapon]` so the two can never drift apart.
- Every other entity keeps its own `weapon.speed` (no behavior change).

It is wired into **both** sites in `meleeSwing` / the swing-timer reset:

1. the swing-timer reset (the cadence), and
2. the attack-power normalization term in the damage formula.

Wiring both matters: re-cadencing alone would let Wolf Form double-dip (fast swings
**and** the slow staff's heavy AP-per-swing weighting). Classic feral normalizes
both, so a faster swing correctly hits softer.

```mermaid
flowchart TD
    A[player auto-attack swing fires] --> B{form_cat aura present?}
    B -- "no (everyone else)" --> C[use equipped weapon.speed]
    B -- "yes (Wolf Form)" --> D[use ROGUE_BASE_SWING_SPEED = rogue dagger 1.8s]
    C --> E[swingTimer reset]
    C --> F[AP normalization in meleeSwing]
    D --> E
    D --> F
    E --> G[next swing eligible]
    F --> H[per-swing damage]
```

```mermaid
flowchart LR
    subgraph content [static content tables]
      R[CLASSES.rogue.startWeapon] --> RD[rusty_dagger speed 1.8]
    end
    RD --> K[ROGUE_BASE_SWING_SPEED]
    K --> BSS["baseSwingSpeed(e)"]
    BSS --> ST[auto_attack.ts swing timer]
    BSS --> DMG[auto_attack.ts AP normalization]
```

## Verification

Before/after, driving the **real offline Sim** in-browser: a level-20 druid
auto-attacks a god-mode dummy for 20s. Wolf Form **after the fix** lands 12 swings,
exactly overlapping the rogue reference, up from 7 on the slow 2.9s staff.

Wolf Form swing cadence: before vs after vs rogue

The druid shifted into Wolf Form in-game (reproduce with
`node scripts/wolf_form_swing_chart.mjs` while `npm run dev` runs):

Druid in Wolf Form attacking

## Tests and gates

- New `tests/form_swing.test.ts` (5 cases): cadence equals rogue base, ignores the
  staff, leaves out-of-form druids and rogues untouched, and an exact white-hit
  damage assertion proving the AP term uses the form speed (no double-dip).
- Determinism: `baseSwingSpeed` is a pure leaf (only static content tables, no rng /
  Date.now / DOM). Only `form_cat` entities change, so the **parity gate stays green**
  and `tests/architecture.test.ts` passes.
- Server-authoritative `swingTimer` is already mirrored over the wire, so online and
  offline stay in lockstep with no `IWorld` / `ClientWorld` change.
- Scope is strictly Wolf Form; Bear and Travel forms are intentionally untouched
  (`form_swing.ts` is the natural seam if tank-form normalization is ever wanted).

## Notes

Per CONTRIBUTING: no new player-visible strings, no balance numbers invented (the
cadence is derived from the existing rogue weapon, the more classic-correct value).

cc @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk for a look.


jgyy:

PR #1001, opened 2026-06-28
fix(content): unchain Korzul quest from Velkhar so the Sanctum runs once

jgyy:
## Problem

The **Gravewyrm Sanctum** is a single 5-player dungeon instance with three boss quests, all given at Highwatch:

- `q_korgath` (Korgath the Bound, at the threshold) -> requires `q_sanctum_gate`
- `q_velkhar` (Grand Necromancer Velkhar, in the Ritual Vault) -> requires `q_sanctum_gate`
- `q_gravewyrm` (**Korzul the Gravewyrm**, on the final dais) -> **requires `q_velkhar`**

Korgath and Velkhar both unlock the moment the gate is opened, so they can be picked up together and cleared in one pass. But Korzul chained off `q_velkhar`, whose kill objective is *inside* the same instance. You cannot turn in Velkhar until you have killed him in the dungeon, and you cannot pick up Korzul until you have turned in Velkhar, so the only way to get Korzul's quest is to leave and re-enter, **running the entire dungeon a second time** to reach the boss who stands on the dais just past Velkhar.

## Fix

Gate `q_gravewyrm` on `q_sanctum_gate`, exactly like its sibling boss quests. Now a party accepts all three Sanctum quests up front and clears the instance once. The narrative still holds: Velkhar falls in the vault earlier in the run, and Korzul is slain on the dais at the end.

One-line content change in `src/sim/content/zone3.ts`:

```diff
-    requiresQuest: 'q_velkhar',
+    requiresQuest: 'q_sanctum_gate',
```

### Quest graph: before

```mermaid
flowchart LR
  G["q_sanctum_gate"]
  K["q_korgath"]
  V["q_velkhar"]
  Z["q_gravewyrm (Korzul)"]
  G --> K
  G --> V
  V -->|requiresQuest| Z
```

before

### Quest graph: after

```mermaid
flowchart LR
  G["q_sanctum_gate"]
  K["q_korgath"]
  V["q_velkhar"]
  Z["q_gravewyrm (Korzul)"]
  G --> K
  G --> V
  G --> Z
```

after

## Tests

Added `tests/zone3_sanctum_quests.test.ts` (TDD: failing first, then green): asserts all three Sanctum boss quests share the `q_sanctum_gate` prerequisite and that Korzul never chains off Velkhar.

```
npx vitest run tests/zone3_sanctum_quests.test.ts tests/progression.test.ts \
  tests/quest_link_share.test.ts tests/localization_coverage.test.ts \
  tests/nythraxis_final_quest.test.ts tests/architecture.test.ts
# all green
```

No balance numbers, names, or player-facing strings changed, so the i18n catalogs and the generated `/wiki` content are untouched (referential-integrity and freshness gates stay green).

cc upstream maintainers with push access: @Rubsey @FernandoX7 @patrick261 @TrevCavill

### DIFF
--- a/docs/design/master-spec.md
+++ b/docs/design/master-spec.md
@@ -221,7 +221,7 @@ Ground sparkles: highwatch_summons ×2 at (1,654),(-2,657); ogre_war_totem ×7 a
 | q_sanctum_gate | The Sanctum Gate | brother_aldric_highwatch | collect 3 sanctum_key_shard (sparkles, gate plaza) | 4000 | 2000 | — | requires q_voice_below |
 | q_korgath | The Bound Guardian | scout_maren_highwatch | kill 1 korgath_the_bound | 4200 | 2500 | korgaths_chainwraps (all) | requires q_sanctum_gate, minLevel 18, suggestedPlayers 5 |
 | q_velkhar | The Grand Necromancer | brother_aldric_highwatch | kill 1 grand_necromancer_velkhar | 4500 | 3000 | boneguard_breastplate / staff_of_velkhar / shadowmeld_tunic | requires q_sanctum_gate, minLevel 18, suggestedPlayers 5 |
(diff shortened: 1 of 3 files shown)


ryan-foo:
agree with this one, no need for people to run this twice

jgyy:

jgyy:
Thanks @ryan-foo, agreed: this makes the Sanctum a one-time run so nobody is forced through it twice.

PR #1002, opened 2026-06-28
feat(web): add /press coming-soon page

jgyy:
## What

Adds a standalone **Press** landing page served at `/press` (and `/press/`). It is a coming-soon press-kit shell: it leads with a press-inquiries contact, an anti-impersonator notice, and a preview of what the full kit will contain, so journalists landing today get a real contact point and the URL is ready for the full kit later.

It clones the exact self-contained recipe of `merch.html` so the realm's standalone pages stay one consistent set: inline CSS design system (shared `:root` tokens, Cinzel/Alegreya fonts, CRT/torch chrome), an inline **14-locale** i18n dictionary, and the same `data-i18n*` language auto-loader (`?lang=` then `localStorage` then `navigator.language` then `en`).

## How it's wired

```mermaid
flowchart LR
  U[Visitor hits /press] --> A{STATIC_PAGE_ALIASES}
  A -->|dev / preview| V[vite.config.ts rewrite]
  A -->|production| S[server/main.ts serveStatic]
  V --> H[public/press.html]
  S --> H
  H --> L[inline 14-locale loader<br/>?lang then localStorage<br/>then navigator.language then en]
  L --> R[localized page]
  SM[public/sitemap.xml] -.->|/press url| Crawl[search + answer engines]
```

Changes are additive only, no `src/` modules and no new Vite input:

- `public/press.html` - the page (HTML + inline CSS + inline i18n + loader)
- `vite.config.ts` and `server/main.ts` - add `/press` + `/press/` to `STATIC_PAGE_ALIASES` (dev and prod parity)
- `public/sitemap.xml` - add the `/press` URL (preserved byte-for-byte by `build_sitemap.mjs`)

## i18n

All 14 locales (`en, es, es_ES, fr_FR, fr_CA, en_CA, it_IT, de_DE, zh_CN, zh_TW, ko_KR, ja_JP, pt_BR, ru_RU`) are filled in the same change. Per `public/CLAUDE.md` these standalone pages have **no** build-time English-fill or `pending` gate, so a missing locale would silently leak English; a key/locale-completeness check confirmed all 30 keys exist in every locale and every `data-i18n*` DOM reference resolves.

## Verification

- `npm run build` green; `dist/press.html` emitted and `/press` preserved in `dist/sitemap.xml`
- `/press` returns HTTP 200 in dev via the alias
- Rendered at desktop, mobile, and `?lang=ja_JP` (title swaps to `World of ClaudeCraft - プレス`)

### Screenshots

Desktop:

desktop

Mobile:

mobile

Japanese (`?lang=ja_JP`):

japanese

## Notes / follow-ups

- Press contact reuses the existing official address `woc@levystreet.com` (same one `/support` uses) rather than introducing a new mailbox.
- Discoverable via `sitemap.xml`; intentionally not linked from `links.html` yet.
- Framed as "press kit coming soon" so it can grow into a full presskit()-style kit (fact sheet, brand assets, screenshots/trailers) at the same URL.

cc upstream maintainers with push access: @Rubsey @FernandoX7 @patrick261 @TrevCavill

### DIFF
--- a/public/press.html
+++ b/public/press.html
@@ -1,0 +1,1383 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
+  <meta name="color-scheme" content="dark" />
+  <meta name="theme-color" content="#08080d" />
+  <title>World of ClaudeCraft - Press</title>
+  <meta name="description" data-i18n-content="metaDescription" content="The World of ClaudeCraft press kit is being assembled. For press inquiries, contact woc@levystreet.com. Any official statement not from our channels is not us." />
+
+  <link rel="canonical" href="https://worldofclaudecraft.com/press" />
+
+  <!-- Open Graph / Twitter -->
+  <meta property="og:type" content="website" />
+  <meta property="og:site_name" content="World of ClaudeCraft" />
+  <meta property="og:title" data-i18n-content="pageTitle" content="World of ClaudeCraft - Press" />
+  <meta property="og:description" data-i18n-content="metaDescription" content="The World of ClaudeCraft press kit is coming soon. For press inquiries, contact woc@levystreet.com." />
+  <meta property="og:url" content="https://worldofclaudecraft.com/press" />
+  <meta property="og:image" content="https://worldofclaudecraft.com/woc_logo_square.webp" />
+  <meta property="og:image:alt" content="World of ClaudeCraft" />
+  <meta property="og:locale" content="en_US" />
+  <meta name="twitter:card" content="summary_large_image" />
+  <meta name="twitter:title" data-i18n-content="pageTitle" content="World of ClaudeCraft - Press" />
+  <meta name="twitter:description" data-i18n-content="metaDescription" content="The World of ClaudeCraft press kit is coming soon. For press inquiries, contact woc@levystreet.com." />
+  <meta name="twitter:image" content="https://worldofclaudecraft.com/woc_logo_square.webp" />
+  <meta name="twitter:image:alt" content="World of ClaudeCraft" />
+
+  <!-- Structured data: declares the upcoming press page and ties it to the brand,
+       so search + answer engines associate official press contact with this realm only. -->
+  <script type="application/ld+json">
+  {
+    "@context": "https://schema.org",
+    "@type": "WebPage",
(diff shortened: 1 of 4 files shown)


jgyy:

FernandoX7 (commented):
This is a clean, faithful clone of the `merch.html` recipe and it shows. The dev/prod alias parity is correct (`/press` + `/press/` added to `STATIC_PAGE_ALIASES` in both `vite.config.ts:75` and `server/main.ts:170`), the sitemap entry is well formed and survives the build (`build_sitemap.mjs` only rewrites the `/guide` block, so the hand-added `/press` url is preserved), the loader is byte-for-byte identical to `merch.html`, and the inline copy map is internally consistent: 30 keys, every shipped locale matches `en` exactly, every `data-i18n*` reference resolves to a key, no empty values, and the non-Latin locales are real translations (ja `プレス`, ko `보도`, ru `Пресса`, not copied English). No em/en dashes or emojis, and every referenced asset exists. Nice work.

One real gap and one nit.

**should-fix** `public/press.html` (the `copy` map) ships 14 locales, but `SUPPORTED_LANGUAGES` (`src/ui/i18n.resolved.generated/loaders.ts:34`) has 21. Missing: `nl_NL, pl_PL, id_ID, tr_TR, sv_SE, vi_VN, da_DK`. Per `public/CLAUDE.md`, the inline set must match `supportedLanguages` exactly, and there is no English-fill or `pending` backstop here, so a visitor in those 7 languages silently gets the English defaults. To be clear this is inherited: `merch.html` and `links.html` have the identical 14-locale gap (the 7 were added repo-wide in v0.16.0 and never backfilled into the standalone pages), and no test gates it, so this is your call. Cleanest is one follow-up that backfills all 7 across press + merch + links together rather than holding only this new page to the stricter bar. I am not drafting the strings here.

**nit** `public/press.html:9,17,24` the static (no-JS) meta description fallbacks diverge: `name="description"` reads "...is being assembled... Any official statement not from our channels is not us." while `og:description` and `twitter:description` read the shorter "...is coming soon." (no anti-impersonator clause). At runtime the loader overwrites all three from the single `metaDescription` key so they unify, this only shows to crawlers that do not run JS. On an SEO-focused press page it is worth making the three static defaults match.

Wiring, parity, and i18n internals all check out. Happy to re-look once the locale question is settled.

FernandoX7 (approved):

PR #1003, opened 2026-06-28
feat(loot): open tapped corpse loot to all after a 1-minute owner lock

jgyy:
## What and why

Tap rights (the first player or pet to damage a mob owns its loot, XP, and quest credit) previously held a corpse hostage for its **entire** lifetime. If the tapper wandered off, died, or logged out without looting, no one else could take the loot before the corpse decayed, even with other players standing right on it.

Classic-era behavior is that the owner has a limited window, after which the corpse goes **free-for-all (FFA)**. This PR adds that: a corpse the tapper (and their party) have not cleared within **one minute** of it becoming lootable opens to everyone.

### Corpse loot lifecycle

Corpse loot lifecycle: owner lock to FFA

### Timeline: tapper vs stranger

Loot FFA timeout over time

## How it works

- When a corpse becomes lootable, `rollLoot` starts a `LOOT_FFA_DELAY` (60s) owner-lock countdown on `Entity.lootFfaTimer`. The dead-mob tick decrements it by `DT`.
- Once it lapses (`lootFfaTimer <= 0`), `Sim.lootCorpse` treats the corpse as untapped, so any player may take the shared loot pool.
- Personal quest drops (`personalFor`) and already-passed need/greed items (`openToAll`) are unaffected: this only governs the tap-owned shared pool.
- The window is genuinely reachable because corpse despawn is gated on `corpseTimer <= 0` **and** `respawnTimer <= 0`, so longer-respawn mobs (and `Infinity`-corpse world bosses) outlive the lock.

## Architecture / parity

- The FFA decision resolves **server-side in the sim** (`lootCorpse`); the client never independently decides loot permission, so **no `IWorld`/wire change** is needed and three-host parity holds.
- New logic lives in a pure leaf `src/sim/loot/loot_ffa.ts` (`LOOT_FFA_DELAY`, `lootHasGoneFfa`, `hasSharedLootRights`), unit-tested directly. The monolith edits are thin wiring.
- `Entity.lootFfaTimer` defaults to `Infinity` (no countdown running) so corpses only go FFA after a real loot roll.
- Determinism: only `DT` and the constant; no new `Rng` draws.

## Tests

- `tests/loot_ffa.test.ts`: pure coverage of the timeout predicate and the shared-loot-rights matrix (tapper / party / stranger / untapped, before and after FFA).
- `tests/loot_ffa_sim.test.ts`: end-to-end on a live `Sim`: a stranger is blocked during the lock, the tapper can loot during it, the stranger can loot after it lapses, plus a same-seed determinism check.
- Parity goldens regenerated: the diff is **only** the additive `lootFfaTimer` field and the resulting state hashes; `rng` draw-order and `events` hashes are untouched (verified), so no behavioral drift.
- `npm test` green (one unrelated pre-existing flaky timeout in `tests/delves.test.ts` under full-suite load, passes in isolation), `npx tsc --noEmit` clean, `npm run build` clean, S3 i18n guard and `tests/architecture.test.ts` green.


jgyy:

PR #1004, opened 2026-06-28
fix(pvp): add diminishing returns to stuns (break chain-stun lock)

jgyy:
## Summary
Player-cast **stuns had no diminishing returns**, so a duel/arena opponent could be chain-stunned indefinitely. Root, Polymorph, and Fear already diminish in PvP (the `diminishedCrowdControlDuration` path), but every stun apply site set its duration directly, leaving a near-permanent stun-lock as the strongest PvP control in the game.

This adds classic-style stun DR (**100% -> 50% -> 25% -> immune**, resetting **18s** after the last application). Following review feedback from @Rubsey, stuns are **not** one shared bucket: they split into three independent classic categories so a from-stealth opener never eats into a controlled stun's diminishing chain. PvE is unaffected: DR only applies when both source and target are players.

## Stun DR categories
A small pure module `src/sim/stun_dr.ts` classifies each stun ability into one bucket; same-category stuns diminish together, different categories are fully independent:
- **openerStun**: Cheap Shot, Pounce
- **controlledStun**: Kidney Shot, Hammer of Justice, Bash, Charge, Bear/Feral Charge
- **randomStun**: proc-style stuns (none represented yet; the safe default)

So Cheap Shot no longer diminishes a following Kidney Shot, keeping the core rogue opener flow intact, while repeated Kidney Shots (or repeated Hammer of Justice) still diminish within their own bucket.

## Root cause
`diminishedCrowdControlDuration(source, target, category, duration)` covered only `root | polymorph | fear`. The `stun` and `finisherStun` effects in `combat/effect_dispatch.ts` applied `eff.duration` straight onto the aura, never consulting the DR table, so repeated stuns always landed at full length.

## Fix
- `src/sim/stun_dr.ts` (new): `stunDrCategory(abilityId)` maps an ability to `openerStun | controlledStun | randomStun`, plus `isStunDrCategory()`.
- `types.ts`: `CrowdControlDrCategory` replaces `'stun'` with the three stun buckets.
- `sim.ts`: add `PVP_STUN_DR_RESET = 18`; all three stun buckets share it via `isStunDrCategory(category)`, then fall through the existing multiplier path (`[1, 0.5, 0.25]` then `null` = immune), identical to `root`.
- `combat/effect_dispatch.ts`: route both the `stun` and `finisherStun` effects through `ctx.diminishedCrowdControlDuration(p, target, stunDrCategory(ability.id), ...)`, and skip applying the aura when it returns `null` (immune).

PvE/mob/boss stuns (War Stomp, Crushing Blow, Concussive Blow, Nythraxis mechanics) are intentionally left untouched: the DR gate requires a player source AND a player target, so a mob stun always lands at full duration.

## How stun DR resolves
```mermaid
flowchart TD
    A["Player casts a stun on an enemy player"] --> B{Both source and target are players?}
    B -- "No (PvE)" --> P["Full duration applied (DR is PvP only)"]
    B -- "Yes (duel / arena)" --> K{"stunDrCategory(ability.id)"}
    K -- "Cheap Shot, Pounce" --> O["openerStun bucket"]
    K -- "Kidney Shot, Hammer of Justice,<br/>Bash, Charge, Bear Charge" --> C["controlledStun bucket"]
    K -- "proc stuns" --> R["randomStun bucket"]
    O --> O1["100% to 50% to 25% to immune<br/>(18s reset window)"]
    C --> C1["100% to 50% to 25% to immune<br/>(18s reset window)"]
    R --> R1["100% to 50% to 25% to immune<br/>(18s reset window)"]
    O1 -. "independent chain" .- C1
    C1 -. "independent chain" .- R1
```

Stun DR flow

## Tests
- `tests/stun_dr.test.ts` (new): unit coverage of the classifier, including that opener and controlled stuns land in independent buckets and unknown/proc stuns default to `randomStun`.
- `tests/pvp_safety.test.ts`:
  - **Cross-category independence**: a fully diminished opener chain does not reduce a following controlled stun, which still diminishes within its own bucket.
  - **Duel stun DR**: repeated Hammer of Justice (4s rank-2 stun) diminishes 4s -> 2s -> 1s -> immune (no aura), then restores to 4s after the 18s window resets.
  - **PvE no-DR guard**: a player stunning a mob keeps the full 4s on every repeat.

## Validation
- `npx vitest run tests/stun_dr.test.ts tests/pvp_safety.test.ts` (20 passed)
- `npx vitest run tests/architecture.test.ts`
- `npx tsc --noEmit`
- `npm test` (one unrelated flaky timeout in `tests/delves.test.ts`, passes in isolation)
- `npm run build`

No wire/persistence/IWorld surface changes: `ccDr` is server-internal sim state and is not serialized, so no parity golden regen is needed.

### DIFF
--- a/src/sim/combat/effect_dispatch.ts
+++ b/src/sim/combat/effect_dispatch.ts
@@ -19,6 +19,7 @@ import { recalcPlayerStats } from '../entity';
 import type { GroundAoE } from '../entity_roster';
 import type { PlayerMeta, ResolvedAbility } from '../sim';
 import type { SimContext } from '../sim_context';
+import { stunDrCategory } from '../stun_dr';
 import { addThreat } from '../threat';
 import type { AbilityDef, Entity } from '../types';
 import { armorReduction, meleeMissChance } from '../types';
@@ -132,7 +133,7 @@ export function runEffects(
         const dur = ctx.diminishedCrowdControlDuration(
           p,
           target,
-          'stun',
+          stunDrCategory(ability.id),
           eff.base + eff.perCombo * spentCombo,
         );
         if (dur === null) break;
@@ -306,7 +307,12 @@ export function runEffects(
       }
       case 'stun': {
         if (!target || target.dead) break;
-        const remaining = ctx.diminishedCrowdControlDuration(p, target, 'stun', eff.duration);
+        const remaining = ctx.diminishedCrowdControlDuration(
+          p,
+          target,
+          stunDrCategory(ability.id),
+          eff.duration,
+        );
         if (remaining === null) break;
         ctx.applyAura(target, {
           id: `${ability.id}_stun`,
--- a/src/sim/combat/effect_dispatch.ts
+++ b/src/sim/combat/effect_dispatch.ts
@@ -129,7 +129,13 @@ export function runEffects(
       }
       case 'finisherStun': {
         if (!target || target.dead || spentCombo <= 0) break;
-        const dur = eff.base + eff.perCombo * spentCombo;
+        const dur = ctx.diminishedCrowdControlDuration(
+          p,
+          target,
+          'stun',
+          eff.base + eff.perCombo * spentCombo,
+        );
+        if (dur === null) break;
         ctx.applyAura(target, {
           id: `${ability.id}_stun`,
           name: ability.name,
@@ -300,12 +306,14 @@ export function runEffects(
       }
       case 'stun': {
         if (!target || target.dead) break;
+        const remaining = ctx.diminishedCrowdControlDuration(p, target, 'stun', eff.duration);
+        if (remaining === null) break;
         ctx.applyAura(target, {
           id: `${ability.id}_stun`,
           name: ability.name,
           kind: 'stun',
-          remaining: eff.duration,
-          duration: eff.duration,
+          remaining,
+          duration: remaining,
           value: 0,
           sourceId: p.id,
           school: ability.school,
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -230,6 +230,7 @@ import {
 import * as fiestaBotsMod from './social/fiesta_bots';
 import { PartyMachine } from './social/party';
 import { SpatialGrid } from './spatial';
+import { isStunDrCategory } from './stun_dr';
 import { Targeting } from './targeting';
 import {
   addThreat,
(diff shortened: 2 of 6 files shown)


jgyy:
cc @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk for review. This closes a PvP balance hole: stuns were the only hard CC with no diminishing returns, so chain-stun lock was possible in duels and arena. The fix mirrors the existing root DR (100/50/25/immune, 18s reset) and leaves PvE stuns untouched. Tests and validation are in the PR body.

jgyy:

Rubsey:
Tagging @jgyy on the stun DR shape here.

I think the current implementation is too broad for a classic-style ruleset because it routes every player-cast `stun` and `finisherStun` through one shared `stun` DR category. That means a rogue Kidney Shot would reduce the next Cheap Shot on the same target, and Cheap Shot would reduce Kidney Shot, because both now hit the same target `ccDr.get("stun")` bucket.

Classic-era handling is more granular: same-category CC diminishes together, but stuns are not all one category. Cheap Shot is an opener stun, while Kidney Shot / Hammer of Justice / Bash are controlled stuns. So the safer shape is probably to split this into explicit stun DR categories, for example:

- `openerStun`: Cheap Shot, Pounce
- `controlledStun`: Kidney Shot, Hammer of Justice, Bash, and similar deliberate player stuns
- `randomStun`: proc-style stuns, if/when represented

That still fixes repeated same-bucket chain stuns, but avoids gutting core rogue opener flow by making Cheap Shot and Kidney Shot punish each other. The PR tests should probably include a rogue case proving Cheap Shot does not diminish Kidney Shot, while repeated Kidney Shots or repeated HoJ-style controlled stuns still diminish full -> half -> quarter -> immune.

jgyy:
Thanks @Rubsey, that's a fair call: one shared `stun` bucket was too broad for a classic-style ruleset. Pushed a follow-up that splits stun DR into three independent classic categories so a from-stealth opener never eats into a controlled stun's chain.

New shape (pure `src/sim/stun_dr.ts` classifier, keyed by ability):

- **openerStun**: Cheap Shot, Pounce
- **controlledStun**: Kidney Shot, Hammer of Justice, Bash, Charge, Bear/Feral Charge
- **randomStun**: proc-style stuns (none represented yet; the safe default)

Same-category stuns still diminish full to half to quarter to immune over the 18s window; different categories are fully independent, so Cheap Shot no longer diminishes Kidney Shot and the rogue opener flow stays intact. PvE stuns remain untouched, and `ccDr` is not serialized so there is no parity golden regen.

The updated flow diagram is in the PR description above.

Tests added per your suggestion: a unit suite over the classifier, plus a duel test proving a fully diminished **opener** chain does not reduce a following **controlled** stun, while repeated controlled stuns still diminish within their own bucket (full to half to quarter). `npm test` and `npm run build` green.

PR #1005, opened 2026-06-28
feat(ui): right-click a player buff to cancel it

jgyy:
## Right-click a buff to cancel it

Players can now right-click one of their own helpful auras on the buff bar to cancel it, the classic-MMO convention. Debuffs are never cancelable, so this is purely quality-of-life and not a CC-removal exploit.

### Behavior
- Right-click a buff on **your own** buff bar to remove it. The icon shows a `context-menu` cursor and a gold hover border to signal the affordance.
- A target's debuff strip stays read-only; only the local player's own non-debuff icons are wired.
- Canceling a stat buff (`buff_*`) or a shapeshift `form_*` recalculates derived stats so the bonus actually un-folds (e.g. a shapeshift cancel reverts to caster form). HoTs / absorbs / imbues need no recalc.

### Screenshots
Before (Stoneskin `buff_armor`, Renew `hot`, Crippling Poison `slow` debuff) and after right-click-canceling Stoneskin (and attempting Crippling Poison, which is refused):

| Before | After |
|---|---|
| before | after |

The Stoneskin shield is gone; the red-bordered Crippling Poison **debuff remains** (cancel refused), and the player's armor un-folded `310 -> 110`. Reproduce with `node scripts/shot_cancel_buff.mjs` (needs `npm run dev`).

### Flow (one sim, three hosts)
The action goes through the single `IWorld` seam, so the offline browser world, the authoritative server, and the headless env all behave identically. The client only *requests*; the `Sim` *decides*.

```mermaid
sequenceDiagram
    participant U as Player
    participant HUD as HUD (renderAuras)
    participant IW as IWorld
    participant CW as ClientWorld
    participant SRV as Server (dispatchMessage)
    participant SIM as Sim.cancelAura (authoritative)

    U->>HUD: right-click own buff icon
    HUD->>IW: cancelAura(auraId)
    alt offline
        IW->>SIM: (Sim implements IWorld) resolve primary player
    else online
        IW->>CW: cancelAura(auraId)
        CW->>SRV: cmd { cancel_aura, aura }
        SRV->>SIM: sim.cancelAura(aura, pid)
    end
    SIM->>SIM: removeCancelableAura (refuses debuffs)
    SIM->>SIM: emit aura-fade + recalcPlayerStats if buff_*/form_*
    SIM-->>HUD: next snapshot reflects dropped aura
```

### Design: one predicate, no drift
The buff-vs-debuff decision lives in a pure leaf, `src/sim/combat/aura_cancel.ts`. The HUD consumes the same `isDebuffAura` it uses for styling, so **"renders as a helpful buff" and "is right-click cancelable" are provably the same set**. `isDebuffAura` covers the full hard-CC / silence family (`stun, root, silence, disarm, blind, hex, polymorph, incapacitate, lockout, ...`) plus negative-value `buff_*` drains, so a player can never strip a silence or root off themselves.

### Wire / determinism
- No new wire field: the `auras` array already rides the self snapshot, so a dropped buff reflects on the next snap. Stat recalc stays server-owned (no optimistic client removal).
- `Sim.cancelAura` draws no RNG and runs outside the tick loop, so the golden-trace / draw-order parity gate is untouched.
- The server validates the `cancel_aura` command field exactly like its `cast` sibling; `pid` is the authenticated session's, never taken from the message.

### Tests
- `tests/aura_cancel.test.ts` - pure predicate + removal (debuff refusal, stat-relevance, first-match, no-op).
- `tests/cancel_aura.test.ts` - Sim-level: cancel a buff, stat un-fold via recalc, debuff refused, unknown-id no-op.

Verified: `npx tsc --noEmit`, `npm run build`, and `tests/architecture.test.ts` + `tests/localization_fixes.test.ts` + the i18n completeness guard all green. Biome reports zero errors on the changed files.

---
cc maintainers with merge access: @levy-street @Rubsey @FernandoX7 @patrick261 @TrevCavill


jgyy:

PR #1010, opened 2026-06-28
feat(ui): show armor type (cloth/leather/mail) on item tooltips

jgyy:
### Problem
Every armor item already has an armor type, either set explicitly on the item or inferred from its class restriction (mail = warrior/paladin/shaman, leather = druid/rogue/hunter, cloth = mage/priest/warlock). But the tooltip never displayed it, so a player looking at a chest piece could not tell which classes it was meant for. The data existed; it just was not surfaced.

### Change
Display the armor type on the tooltip's slot line, the classic way: slot on the left, armor class on the right. When the viewing player's class cannot wear that armor weight, the type renders in **red** (per @Rubsey's request) so ineligible gear reads at a glance. The shots below are captured as a **mage** (a cloth wearer): Mail and Leather show red, Cloth shows the normal light color.

| Mail | Leather | Cloth |
|---|---|---|
| mail | leather | cloth |

Note the currently-equipped Apprentice's Robe in the comparison block shows no armor type: it is generic starter gear with no class hint, so nothing is invented.

In context (bags open):

full

### How it resolves

flow

### Implementation notes
- New pure leaf `src/ui/item_armor_type.ts` (`itemArmorTypeLabelKey`) maps the sim's existing `armorTypeForItem` result to an i18n label key. DOM-free, runtime-i18n-free (type-only import), unit tested in `tests/item_armor_type.test.ts`.
- `hud.ts` `itemTooltip` renders the armor class on the slot row when one resolves; otherwise the plain slot line is unchanged. The red cue reuses the existing pure `canEquipItem(class, item)` predicate against the local player's class (`sim.cfg.playerClass`) and a new presentational `.tt-armor-bad` modifier; no sim/`IWorld`/parity change. Every item tooltip in the client (bags, character window, vendor, market, loot, the comparison block) flows through this one painter, so they are all covered with no per-surface duplication. No `IWorld`/parity change is needed: the data comes from the static content tables resolved identically in `Sim` and `ClientWorld`.
- All interpolated text passes through `esc()`.
- New English-only keys live in the `hud_chrome` catalog domain (the one domain without tsc-enforced per-locale blocks, per `src/ui/CLAUDE.md`); the 13 locale overlays are maintainer-filled at release (marked pending). Generated i18n tables regenerated via `npm run i18n:gen` + `npm run i18n:hash -- --write`.
- New `.tt-row` CSS lays out slot left / type right; the armor class reads brighter so the new cue stands out.

### Verification
- `npx tsc --noEmit` clean
- `npx vitest run tests/item_armor_type.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/i18n_status_registry.test.ts` green
- `npm run build` green
- Screenshots above captured with `node scripts/armor_type_tooltip_shot.mjs`

cc @Rubsey @FernandoX7 @patrick261 (upstream maintainers / push access) for review.


jgyy:

Rubsey:
@jgyy if a class is not eligible to use the gear, can we make the text red?

I.e a mage hovering a mail item would see the armor type in red text. So they'd know that they can't use it.

jgyy:
Great idea, done! @Rubsey When the viewing player's class can't wear that armor weight, the armor type now renders in **red** (e.g. a mage hovering a Mail piece sees "Mail" in red), so ineligible gear reads at a glance.

It reuses the existing pure `canEquipItem(class, item)` predicate, so it's correct for every class/weight combo, and it's purely presentational (new `.tt-armor-bad` modifier, no sim/parity change). Refreshed the screenshots as a mage so you can see Mail/Leather red and Cloth normal, and updated the flow diagram with the new branch.

PR #939, opened 2026-06-25
perf(render): kill open-world travel freezes (constant point-light count + prewarm)

EnriqueGF:
## Open-world travel freezes: investigation + fixes (WIP)

Hunting the in-play "tirones" (multi-hundred-ms hitches) players feel when they travel
across the world. This is the client/renderer follow-up to PR #898 (the boot prewarm).
Draft: the dominant cause is found and fixed; smaller residual freezes are next.

### TL;DR result (measured on a real GPU, RTX 3060 Ti, walking north out of Eastbrook)
- Travel-time shader-program compiles: **34 → 8**
- Worst hitch: **~1375ms → ~54-150ms**
- Post-prewarm program count: **364 → 154** (far fewer permutations)
- Loading screen back to fast (an interim "world tour" approach was reverted as redundant)

### The root cause (the big one): dynamic point-light-count permutation
Three.js compiles a separate shader program per `numPointLights`, and counts a light
iff it is `visible`. `budgetFireLights` toggled the nearest campfires' visibility by
distance, so as the player moved the visible point-light count swung `0 <-> maxPointLights`
and **every material in range recompiled** - ~20 programs linking in a single frame, the
0.6-1.5s hitch. The Mirefen impact-site `PointLight` made it worse: its cull group toggled
visibility within 280u of the crater, flipping the count crossing ~z14 in the Vale.

Confirmed by diffing program `cacheKey`s before/after each hitch: they differed in exactly
one field (`numPointLights`), e.g. `8->7`, `3->0`.

**Fix** (`perf(render): constant point-light count ...`): keep a CONSTANT number of point
lights `visible` so `numPointLights` - and thus every material's program - never changes
while travelling. The nearest `maxPointLights` stay visible (one stable program per
material); lights past the live budget or out of range contribute nothing (intensity 0)
but still count. The impact-site light now rides the same budget (moved out of its
cull-toggled group) instead of flipping the count on its own.

### The other fixes in this branch
- **Dungeon interiors + assets**: the kit + Halloween-bits GLBs were never boot-preloaded
  (a stale comment claimed otherwise), so they streamed in - and their shaders compiled -
  the moment the camera neared a dungeon door (the Fallen Chapel hitch). Now boot-preloaded,
  and a prewarm group compiles the interior materials up front.
- **Temple-water shader bug**: the Drowned Temple water vertex shader named its view-space
  position `mv` but `#include <fog_vertex>` reads `mvPosition`, so it failed to compile
[...]

PR #1012, opened 2026-06-28
perf(render): FPS-first auto-quality detection (on by default) + high-DPI caps

EnriqueGF:
## FPS-first auto-quality: a graphics detector that actually works, on by default for everyone

The graphics default was **ultra with the adaptive governor disabled on ultra**, and there was
no real hardware detection: a fresh or weak machine ran the full pipeline with zero runtime
adaptation. The "auto-adjust" players asked for never existed (the earlier attempt was reverted;
its i18n key was the only thing left). This restores it, biased toward frame rate.

Profiled and measured on a real GPU (RTX 3060 Ti, headed Chrome, vsync off) the whole way.

### What changes
- **Real first-run detection (`src/render/gfx_autodetect.ts`, pure + unit-tested).** Classifies
  the GPU renderer string into a perf class and picks the highest tier that should hold a
  comfortable frame rate, **FPS-first**: it never auto-picks `ultra`, steps a rung down for
  high-DPI panels / low memory / few cores, and forces `low` on phones and software GL.
- **`graphicsPreset 0 = Auto` is the new default**, so the adaptive governor (`render_budget.ts`)
  now runs out of the box instead of being disabled on the old ultra default. A cross-session
  memory (`woc_gfx_autotune`) feeds last session's sustained FPS back in, so a struggling machine
  starts a rung lower next launch.
- **Rolled out to everyone once, silently.** A one-time settings migration force-resets every
  existing player (including manually pinned tiers) to Auto, then sticks so a later manual choice
  is preserved. Ultra stays a deliberate manual escape hatch. The Esc menu gains an Auto option.
- **FPS-first device-pixel-ratio caps.** The post composer (N8AO + bloom) is the dominant GPU
  cost on high/ultra and scales with pixel count, so a high-DPI panel is what craters those tiers.
  Lower the caps (high 1.75 to 1.5, ultra 2.5 to 2.0, medium/low lower). No effect on the desktop
  1x majority; on a dpr=2 panel **high measured 102.7 to 121.5 fps (+18%)**.
- **Telemetry fix.** Three's `info.render` autoReset wiped the scene's draw-call/triangle counts
  on each composer post-pass, so high/ultra reported 1 call / 2 triangles. Reset once per frame
  instead, so the in-game perf overlay/reporter report real GPU load on the composer tiers.
- **Tooling:** `scripts/fps_tier_bench.mjs` (headed real-GPU FPS sweep across tiers) and
  `scripts/fps_profile.mjs` (steady-state main-loop / renderer-phase / foliage profiler).

### Measured (RTX 3060 Ti, 1920x1080, vsync off)
| tier | dpr=1 idle | dpr=2 idle |
|---|---|---|
| medium | ~129 | ~119 |
| high | ~117 | ~103 -> **~121** (after the cap) |
| ultra | ~116 | ~65 (manual-only) |

Auto resolves to `high` here with the governor on. The composer (N8AO + bloom) is the GPU cost
that scales with pixels; `medium` (no composer) is the most DPI-robust, which is why the FPS-first
heuristic leans toward it on weaker hardware.

### Notes / follow-ups
- Throughput axis only; the open-world streaming hitches (multi-hundred-ms first-visit spikes) are
  the separate freeze work.
- The `render_budget` submit-stall hold floors foliage/grass for ~8s after a single transient
  travel spike, even at idle, a visible quality dip with no FPS benefit on capable machines. Left
  untouched here (freeze-governor territory); flagged for a follow-up.

### Verification
`tsc --noEmit` clean, `npm run build` green, full Vitest green (one pre-existing `delves.test.ts`
timeout that is load-flaky and passes in isolation, unrelated to these files), biome formatted.

### DIFF
--- a/scripts/fps_profile.mjs
+++ b/scripts/fps_profile.mjs
@@ -1,0 +1,124 @@
+// Steady-state CPU/GPU frame profiler (headed, real GPU).
+//
+// Boots one tier, runs a sustained forward tour in the open world, then dumps
+// where the frame time goes: the main-loop buckets (sim.tick / renderer.sync /
+// hud.update), the renderer's internal phase split (entities / world /
+// nameplates / submit), and the foliage draw/triangle breakdown. Use it to pick
+// what to optimize, then re-run to confirm the win.
+//
+//   npm run dev   # other terminal
+//   BENCH_TIERS=high node scripts/fps_profile.mjs
+//
+// Env: BENCH_TIER (default auto), BENCH_W/H, BENCH_DPR (1|2), BENCH_RUN_MS.
+import puppeteer from 'puppeteer-core';
+import { BROWSER_PATH } from './browser_path.mjs';
+
+const BASE_URL = process.env.GAME_URL ?? 'http://localhost:5173';
+const TIER = process.env.BENCH_TIER ?? 'auto';
+const W = Number(process.env.BENCH_W ?? 1920);
+const H = Number(process.env.BENCH_H ?? 1080);
+const DPR = Number(process.env.BENCH_DPR ?? 1);
+const RUN_MS = Number(process.env.BENCH_RUN_MS ?? 6000);
+const KLASS = process.env.BENCH_CLASS ?? 'warrior';
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+async function main() {
+  console.log(`Profile: tier=${TIER} ${W}x${H} dpr=${DPR} run=${RUN_MS}ms (HEADED, vsync OFF)`);
+  const browser = await puppeteer.launch({
+    executablePath: BROWSER_PATH,
+    headless: false,
+    args: [
+      `--window-size=${W},${H + 120}`,
+      '--ignore-gpu-blocklist',
+      '--enable-gpu',
+      '--disable-gpu-vsync',
+      '--disable-frame-rate-limit',
+      '--autoplay-policy=no-user-gesture-required',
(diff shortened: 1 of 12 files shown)


EnriqueGF:
Closing as superseded by release/v0.16.0.

Lining this up against release/v0.16.0 for the QA pass, every part of this PR is already in the release, implemented independently with the same approach:

- FPS-first auto graphics detection -> `resolveDefaultGraphicsPreset` in gfx.ts reads `UNMASKED_RENDERER_WEBGL` and classifies by GPU family, with deviceMemory / hardwareConcurrency as tie-breakers (the same model this PR used), plus the runtime `shouldUseAutoGovernor`.
- Auto on by default for everyone -> on first boot with no persisted preset the release already resolves a DEVICE-AWARE default through `resolveDefaultGraphicsPreset`.
- FPS-first device-pixel-ratio caps -> already capped in gfx.ts.
- Accurate per-frame draw-call telemetry -> already in renderer.ts.
- The graphics options UI -> the release extracted options into `options_view.ts` / `options_window.ts`, which already carries the graphics-quality preset selector (including Auto).

There is no behavioral delta left to land, so forcing this version would only duplicate equivalent logic and revert the options-module refactor. The genuinely-new perf work is lined up against release/v0.16.0 separately: #1015 (async-compile gate), #1013 (crowd-adaptive LOD), #939 (prewarm coverage: foliage + dungeon prewarm, mob-model coverage, quest-light hiding).

PR #1016, opened 2026-06-28
perf(ui): idle-prewarm class ability/aura icons (no mid-game compose stall)

EnriqueGF:
## Idle-prewarm the class's ability/aura icons (kill the on-open compose stall)

Hunting **main-thread blocks that hit mid-game** (not at boot). Found by static analysis:
procedural icons are composed **synchronously on first use** - `iconDataUrl()` draws the
recipe to a canvas and PNG-encodes it (`toDataURL`), both on the main thread, then caches the
result. So the moment you **open the spellbook / action bar**, or the **first time each
buff/debuff lands in combat**, the game composes + encodes every *new* icon at once: a
main-thread stall proportional to the icon count, right in the middle of play.

### Fix
Warm them during idle. After entering the world, enumerate the class's known abilities
(`abilitiesKnownAt`) and compose each ability + aura icon during idle (`requestIdleCallback`,
sliced, with a `setTimeout` fallback) - mirroring the existing **map-bg idle prewarm**. By the
time a window opens or a buff lands the icon is already cached, so nothing composes on the
main thread mid-game.

- **Best-effort + safe**: a failure never blocks entering the world; a stale idle callback is
  a cheap no-op (the warm-list is exhausted); already-cached icons are skipped.
- The warm-list logic is a **pure, unit-tested module** (`src/ui/icon_prewarm.ts`); the HUD is
  the thin consumer.

### Verification
`tsc --noEmit` + `npm run build` + the unit test (3 cases) green. The stall reduction itself
was found statically and is **not yet live-profiled** - the authoring session exhausted the
machine's Chrome-launch capacity. Confirm with `node scripts/profile.mjs combat` (open windows /
fight) before merge; the mechanism (prewarm -> cache hit) is straightforward.

### Context
Part of the same main-thread-stall hunt as the async shader-compile gate (#1015): that covers
GPU program links; this covers the CPU-side canvas compose for icons. Same idea - move
first-use work off the hot path.

### DIFF
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -53,6 +53,7 @@ import {
 import type { ZoneDef } from '../sim/data';
 import {
   ABILITIES,
+  abilitiesKnownAt,
   CLASSES,
   COMPANION_UPGRADE_COSTS,
   DELVE_AFFIXES,
@@ -206,6 +207,7 @@ import {
   tOptional,
   tPlural,
 } from './i18n';
+import { abilityIconWarmKeys, type IconWarmKey } from './icon_prewarm';
 import { iconDataUrl, QUALITY_COLOR, raidMarkerDataUrl } from './icons';
 import { itemStatDeltas } from './item_compare';
 import { PICK_ACTION_HOTKEYS } from './lockpick_panel';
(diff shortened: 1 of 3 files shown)


EnriqueGF:
Closing: the idle icon prewarm causes stutter while moving/playing (maintainer confirmed twice). Removing the change.

PR #1020, opened 2026-06-29
feat(items): gear requires a minimum level to equip

jgyy:
## Gear needs level requirements

Higher-tier gear can now require a minimum character level to equip, the classic-MMO guard that stops a level-1 character equipping a lucky raid drop (or a twink hand-up) far above their level. The requirement is enforced in the deterministic sim core, so it holds identically offline, on the server, and headless.

### Live behaviour

A level-5 character inspecting a rare helmet: the tooltip shows a red **Requires Level 12**, and the equip is refused with `You must be level 12 to equip that.`

Requires Level tooltip

### How the requirement is decided

There is no per-item "item level" in this repo, so the requirement is **derived from `quality`** (the one tier signal every item already carries), mirroring the existing derive-don't-store approach in `armorTypeForItem`. An item may still pin an explicit `requiredLevel` to override the default.

The leveling greens (poor/common/uncommon) are the gear quests and vendors hand you AS you level, so they stay ungated, a flat band there would strand an early quest reward you just earned but are a level or two short of. The gate begins at rare and clamps to `[1, MAX_LEVEL]` so the rarest pieces stay reachable at the cap (a band above the cap would be unequippable forever).

Quality to required level

### The equip path

The gate sits after the existing class/armor proficiency check. Auto-equip mirrors its proficiency guard with a silent level guard, so looting gear above your level no longer fires the error toast (the explicit equip path keeps the message). Equipped/saved gear is never retroactively stripped, the gate only fires on a new equip action.

Equip gate flow

### What is in the change

- **`src/sim/item_level_req.ts`** (new pure leaf): `requiredLevelFor(item)` + `meetsLevelRequirement(level, item)`, host-agnostic and unit-tested directly.
- **`src/sim/items.ts`**: the equip-path gate, emitting a localized `You must be level N to equip that.`
- **`src/sim/sim.ts`**: `maybeAutoEquip` skips level-gated gear silently.
- **`src/sim/types.ts`**: optional `requiredLevel?: number` override on `ItemDef`.
- **HUD tooltip**: classic `Requires Level N` line for equippable gear gated above level 1, red when the viewer is below it (`hudChrome.itemTooltip.requiresLevel`, English-only add plus the five non-Latin overlay fills per M16).
- **i18n**: `error.equipLevel` added to `sim_i18n` across all 21 locales plus a matcher RULE, so the sim emit is recognized by the S3 guard.

### Testing

- New `tests/item_level_req.test.ts` (pure bands + clamp + override + determinism) and `tests/equip_level_requirement.test.ts` (gate blocks below / allows at-and-above, error text, ungated green, silent auto-equip skip, determinism).
- Full suite green (`5404` passing); the only red is the pre-existing `delves.test` full-suite ordering flake (passes in isolation, reproduces on a clean base) and the `tests/browser/*` suite which needs uninstalled optional deps.
- `tsc`, `npm run ci:changed` (Biome on changed files), the S3 i18n guard, the M16 completeness gate, the architecture/determinism guard, the parity golden gate, and `npm run build` all pass. The parity golden is untouched (the leveling-green path it exercises stays ungated).

cc @Rubsey @FernandoX7 @patrick261 for a look when you have a moment, happy to adjust the bands or wording.

Co-authored-by: Claude Opus 4.8 (1M context)


ryan-foo:
Hahaha I have been giving so many low level users high level gear to enjoy. Low key a little sad that this is getting patched, but prob needed :P

FernandoX7 (commented):
This is a clean, well-scoped change. The requirement is derived in a pure sim leaf with no rng or clock, the gate sits at the single `equipItem` chokepoint so every path inherits it (the `equip` command, `useItem` on a weapon/armor, and `maybeAutoEquip`), and the client stays a renderer that just forwards the command. Determinism holds: no new draws, and the parity golden is untouched because the leveling-green path stays ungated. Auto-equip mirroring its proficiency guard with a silent level guard (so looting above-level gear does not spam the error toast) is the right call, and it is tested. tsc, the targeted suites, and the S3 i18n guard all pass locally.

A few small things, none blocking:

- nit `src/ui/hud.ts:3032`: `requiredLevelFor(item)` is evaluated twice, in the `> 1` guard and then again for `req` on the next line. It is a pure, cheap call, but you can compute it once and reuse.
- design note `src/sim/item_level_req.ts:37`: `legendary` maps to `MAX_LEVEL`, so a legendary piece is equippable only at exactly the cap. Both current legendaries (`deathless_heartwood`, `kingsbane_last_oath`) are endgame raid weapons, so nothing is stranded today, but if a sub-cap legendary reward ever lands it would be unequippable until the cap. Worth keeping in mind since you asked about the bands.
- The only real merge conflicts are the standard stale-base i18n trio (`src/ui/i18n.resolved.generated/pending.ts`, `src/ui/i18n.resolved.sha256`, `src/ui/i18n.status.summary.json`). That is mechanical regen, not a design problem: merge the base and re-run the i18n build (`i18n:build` / `i18n:scan` / `i18n:hash --write`).

The English emit plus the matcher RULE for `error.equipLevel` is the sanctioned pattern, and the `Requires Level` tooltip key is correctly English-only in `hud_chrome` with the number going through the formatter and reusing the existing `tt-red` class (no play.html CSS parity gap). Nice work.

Rubsey (commented):
Nice groundwork here. `item_level_req.ts` is a clean pure leaf (no rng/clock/DOM, imports only `./types`), the gate sits in the shared `equipItem` so the server enforces it and the tooltip prediction cannot diverge from the server, and the determinism, architecture, and parity gates all stay green with the golden fixtures unchanged. The auto-equip silent-skip vs explicit-error split is the right call.

The one thing I would change before this goes in: quality is the wrong signal to gate on.

**blocking** `src/sim/item_level_req.ts:31-46`: deriving the requirement from `quality` misfires in both directions, and the header comment's premise ("this repo has no per-item item level") is no longer true. release/v0.18.0 landed `src/sim/item_level.ts`, which already computes `itemSourceLevel(itemId)`, the level of the content a piece drops from. Gating on flat quality bands instead:

- Over-gates earned loot. 21 rare-and-above pieces drop from content below their new required level. `moggers_shiv` drops from level-6 content but needs level 12; `deathlord_sabatons` (epic) drops at 16 but is gated to 18; the first dungeon's blues and the delve rewards are all stranded. A rare that drops at level 6 is at your level, not a twink item.
- Under-gates the endgame. Every rare is gated to 12 regardless of where it comes from, so a level-12 character can equip a level-20 rare, and uncommons are not gated at all. That is the exact twink case this PR is trying to stop.

Suggest deriving `requiredLevelFor` from `itemSourceLevel(item.id)` (with a sensible floor, and a fallback for items that have no source), and keeping the explicit `requiredLevel` override on top. That reuses the system the repo already has, closes the twink hole, stops stranding at-level drops, and keeps us honest on the "do not invent balance numbers" rule.

Smaller notes:

- **nit** `src/sim/item_level_req.ts` vs `src/sim/item_level.ts`: two near-identical names, both imported into `hud.ts`. If the requirement folds onto the source-level basis, this may collapse into a couple of functions on `item_level.ts` rather than its own module.
- **nit** `src/ui/hud.ts:3031-3038`: the tooltip can now show an "Item Level N" line and a "Requires Level N" line together (two different numbers on one card). Worth a look that it reads clearly.
- **nit** `src/sim/item_level_req.ts:44`: a non-finite explicit `requiredLevel` slips past the clamp (`Math.floor(NaN)` is `NaN`), which makes the piece unequippable by anyone and also hides the tooltip line (`NaN > 1` is false). A `Number.isFinite` guard covers it. Authored data only, so minor.
- **nit** tests exercise synthetic items only. Once the basis is source level, an assertion that no real content item is gated above where it drops is the natural regression test.

Heads up: I rebased the branch onto release/v0.18.0 to clear the conflicts, so the history moved and it is mergeable now once the derivation is sorted.

jgyy:
Addressed Rubsey's blocking review comment: `requiredLevelFor` now derives the requirement from `itemSourceLevel` (`src/sim/item_level.ts`, the level of the content an item actually drops or is quested from) for `rare` and above, instead of a flat quality band. This fixes both directions Rubsey called out:

- Over-gating: a level-6 rare (`moggers_shiv`) no longer requires level 12, it requires 6.
- Under-gating: a level-20 rare no longer caps out at 12, it requires 20.

Items with no derivable source (vendor stock, starter gear, synthetic items) still fall back to the old per-quality band. An explicit `requiredLevel` override still wins, now with a `Number.isFinite` guard so a non-finite value cannot slip past the clamp (Rubsey's smaller nit).

Also updated:
- `tests/equip_level_requirement.test.ts` to use a real content item (`knight_commanders_greaves`, derived required level 13) instead of a stale flat-band assumption.
- `tests/item_level_req.test.ts` with a new regression check that no real rare-and-above item in `ITEMS` is ever gated above its own source level, plus a pinned check on `moggers_shiv`.

Merged the branch onto the current `release/v0.18.0` tip: the only conflicts were the mechanical generated-artifact trio (`i18n.resolved.generated/pending.ts`, `i18n.status.summary.json`, `i18n.resolved.sha256`), resolved by regenerating (`npm run i18n:gen` + `npm run i18n:hash -- --write`) per the repo's standard playbook.

`tsc`, the targeted suites, the S3 i18n guard, the parity gate, and `npm run ci:changed` all pass. CI on the PR is green (`PR gate`, `Format + lint`).

Left as-is (non-blocking nits from the reviews):
- FernandoX7's nit about `requiredLevelFor` being called twice in the tooltip was already fixed on this branch before I picked it up.
- Rubsey's naming nit (`item_level_req.ts` vs `item_level.ts` collapsing into one module) is a reasonable follow-up but not required for this change; happy to fold it in if preferred.
- The two-numbers-on-one-tooltip-card nit: Item Level and Requires Level render as distinct, differently colored lines and read clearly; left as is.

Screenshot refresh: skipped. This is a sandboxed environment with no display and no way to run `npm run dev`/`npm run server` for the puppeteer-based screenshot scripts, so I did not touch the tooltip screenshots already linked in the PR description. The visual behavior itself (tooltip layout, colors, wording) is unchanged by this fix, only the derived number changes per item.

Rubsey (commented):
The gate lands cleanly. Deriving the requirement instead of hand-authoring it per item is the right call, `item_level_req.ts` is a genuinely pure, host-agnostic leaf with direct unit tests, and anchoring the derivation to `itemSourceLevel` (so a rare is gated at where it actually drops, not a flat quality band) is the correct model here: it gates gear at its power tier, which is what stops a below-level character wielding a twinked hand-up. Server stays authoritative, the client only reads `requiredLevelFor` for the tooltip, and the sim error ships with its matcher RULE in the same change. Determinism, sim purity, the SimContext seam, and cross-host/wire parity all check out: `requiredLevel` is static item content, so there is no new Entity or wire field. Coverage is good, including the regression test asserting no rare-and-above item is gated above its own source level.

A few small things:

**should-fix**
- `package.json:115-117` adds a top-level `allowScripts: { "sharp@0.34.5": true }` that is unrelated to this feature and has no consumer in the repo (no lavamoat/allow-scripts tooling, and npm ignores a top-level `allowScripts`). It reads like a stray artifact from a local tool. Drop it to keep the PR scoped, or note what reads it.

**nit**
- The PR description still frames the requirement as "derived from `quality`" (with the quality-bands screenshot), but the code now derives from `itemSourceLevel` for rare+ and only falls back to the bands when an item has no source. Worth refreshing the body so a reader is not misled.
- `tests/equip_level_requirement.test.ts:11` names the constant `HELM`, but `knight_commanders_greaves` is a legs piece (it equips into `equipment.legs`). Harmless, just a confusing name.

Nothing blocking. The auto-equip silent-skip versus explicit-equip toast split is a nice touch and is covered by a test.

PR #1022, opened 2026-06-29
feat(items): sell food and drink in stacks of 5 at classic per-unit pricing

jgyy:
## What

Vendors now sell **food and drink in stacks of 5 for the price they already cost**. Buying Freshly Baked Bread or Refreshing Spring Water from a merchant hands you a stack of 5 for the same listed `buyValue` (the price is unchanged, you just get the stack). Gear, reagents, and instant potions/elixirs still buy one at a time.

This matches classic vendor behavior, where staple food and drink are bought as a stack rather than clicking five times.

Vendor window showing food and drink with an x5 stack badge

The four food/drink goods carry the `x5` badge; the two potions do not.

## How

The "how many units per purchase" decision is a pure leaf, `src/sim/vendor_stack.ts`:

```ts
export const VENDOR_STACK_SIZE = 5;

export function vendorStackSize(def: ItemDef): number {
  return def.kind === 'food' || def.kind === 'drink' ? VENDOR_STACK_SIZE : 1;
}
```

One rule, two consumers:

- **`items.ts buyItem`** (server-authoritative grant): charges the listed `buyValue` once, then `addItem(item, vendorStackSize(def))`.
- **`ui/vendor_view.ts`** (the vendor window): tags each goods row with a `quantity`, and the painter shows the existing `x{count}` stack badge (the same one buyback rows already use, so no new i18n key).

Sharing the leaf means the `x5` the player sees and what the buy grants can never drift.

Flow of the shared stack rule

## Determinism / parity

No RNG and no new `Entity` field. The `inventory_vendor` parity golden is a pure state regen (bread/water `1` to `5`, potion stays `1`); every step still reports `rng.draws: 0`, so there is no draw-order change.

## Tests

- New `tests/vendor_stack.test.ts` pins the rule (food/drink to 5, everything else to 1).
- `tests/vendor_view.test.ts` asserts goods carry the right `quantity`.
- `tests/items.test.ts` and `tests/sim.test.ts` updated: buying bread/water yields 5 at the unchanged price; a potion stays single.
- `tests/parity` `inventory_vendor` golden regenerated (state-only).
- `tsc`, `tests/architecture.test.ts`, and `biome ci --changed` are green over the diff.


jgyy:
cc @Rubsey @FernandoX7 @TrevCavill @patrick261 @CharlieSaxton @maxpolaczuk for a look when you have a moment, thanks. Small, gameplay-neutral vendor tweak: food/drink now buy as a stack of 5 for the same price, behind one shared pure rule, with the inventory_vendor parity golden regenerated (state-only, no rng draws).

Rubsey:
Rebased this onto `release/v0.18.0` and repointed the base.

While reviewing I found the flat-price stack (5 units granted for the single-unit `buyValue`) had two problems:

1. **Not classic-faithful:** it cut the effective per-unit price of every vendor consumable to 1/5. Classic vendors charge per unit, so a stack of 5 costs 5x, not 1x.
2. **Money-printing loop:** on `baked_bread` and `spring_water` a stack cost 25c to buy but sold back for `sellValue * 5 = 30c`, netting +5c per cycle, unbounded, right at the starter vendor (Trader Wilkes). That inverts the invariant that per-unit sell value stays below per-unit buy price.

Pushed a fix (0829b4e3): `buyItem` now charges `buyValue * vendorStackSize` with a matching affordability guard, so the per-unit price is unchanged from classic and the arbitrage is gone. The vendor window shows the stack total and folds the `x5` into the buy `aria-label`. Added a buy-then-sell-back regression test (asserts a net copper loss) and regenerated the `inventory_vendor` parity golden (state-only; rng draws and digest unchanged).

Net effect: one click still buys a stack of 5 (the QoL you intended), but it costs 5x (125c for bread), matching classic per-unit pricing. `tsc`, the vendor/items/sim tests, the parity gate, and `biome ci` on the changed files are all green.

The rest of the implementation is clean: the pure `vendorStackSize` leaf, the shared rule between the sim buy path and the UI badge, determinism, server authority (quantity is server-derived, the client sends only the itemId), and the i18n key reuse are all good.

PR #1029, opened 2026-06-29
feat(ui): inspect another player's equipped gear

jgyy:
## What

Right-clicking another player offers **Inspect** (the existing "View Profile" action), and the profile window now shows that player's **equipped gear** as a read-only paperdoll, not just portrait / name / level / class.

| Right-click menu | Inspect window |
|---|---|
|  |  |

### Follow-up: wider window + mobile one-item-per-row list

Per review, the desktop window is wider (280px to 420px) and the read-only gear
names now wrap instead of ellipsizing, so the full item name is always readable.
On mobile the paperdoll stacks into a single, full-width, scrollable, one-item-per-row
list so each name reads one at a time.

| Desktop (wider, names wrap) | Mobile (single-column scrollable list) |
|---|---|
|  |  |

## Why it needed a wire change

Other players' equipment was not available client-side: only the mainhand weapon id rode the wire (for the held-weapon model). Item names, icons, and tooltips already resolve fully client-side from the static `ITEMS` table, so the only thing that has to cross the wire is the slot to item-id map.

I mirror the full worn set onto the `Entity` as a **render-only** `equippedItems` field, recomputed in `recalcPlayerStats` right next to `mainhandItemId`, and ride it over the **identity record** as the terse `eq` field (players only, first appearance + on change, exactly like `mh` / `hb`). The sim never reads it for gameplay.

### Data flow

data flow

sequence

## How

- `src/sim/types.ts` / `src/sim/entity.ts`: render-only `Entity.equippedItems`, defaulted in `baseEntity` and copied from `PlayerMeta.equipment` in `recalcPlayerStats` (a copy, never an alias).
- `server/game.ts`: `identityFields()` emits `eq` (gated `kind === 'player'` + non-empty, mirroring the `mh` gating).
- `src/net/online.ts`: `applyWire()` decodes `e.equippedItems = w.eq ?? {}` inside the full-identity branch, and the mirror-entity factory defaults it to `{}`.
- `src/ui/hud.ts`: `openInspect()` reuses the pure `buildPaperdollView` core + the existing item icon / name / tooltip + slot-name helpers to paint a read-only paperdoll. **No new i18n keys** (the section header reuses `classDetails.sections.equipment`).
- `src/styles/shell.css`: `.inspect-equip*` rules (token-based) for the two-column gear layout in the narrower inspect window.

## Tests

- `tests/inspect_equipment.test.ts` (9): the sim mirror (copy, determinism, mob-empty) and the full `eq` identity encode -> `applyWire` decode round-trip (encode, decode, default-on-omit, omit-when-empty, omit-for-non-player).
- `tests/parity/`: `equippedItems` added to `ENTITY_EXCLUDE` (render-only, like `mainhandItemId`) and its anti-loosening pin.
- `scripts/inspect_gear_shot.mjs`: the screenshot tour above.
- Green: `tsc`, `architecture`, `snapshots`, `bandwidth`, `world_api_parity`, `command_schema`, `parity`, `localization_fixes`, `build`.

Reviewed clean by the repo's `qa-checklist` (verdict: READY) and `cross-platform-sync` (parity clean) agents before opening.


FernandoX7:
This is awesome! I recommend making the window for a player's gear wider so we don't have to worry about not being able to read their entire gear name. Right now, the screenshots show ellipsis.

On mobile, we should make the layout scrollable and display gear in a list so it's one item at a time, allowing us to read the entire item name.

This is the UI/UX we are aiming for.

---

Not part of this PR, but should the right-click menu have an option to add friend/block?

jgyy:
Thanks! Both layout points addressed in the latest push (CSS-only; css_corpus/styles_extraction/css_value_validity/client_shell green):

- **Wider desktop window:** 280px to 420px, and the read-only gear name now WRAPS instead of ellipsizing (`.inspect-paperdoll .slot-item`), so the full item name is always readable (e.g. "Heartwood of the Deathless Crown" wraps to two lines). The editable character sheet keeps its nowrap/ellipsis.
- **Mobile list:** the two-column paperdoll stacks into a single, full-width, scrollable, one-item-per-row list (dropping the right column's mirrored row-reverse/right-align), so each name reads in full one at a time. Desktop + mobile screenshots added to the PR body.

On the other question (a right-click add-friend / block option): agreed it is a good addition and, as you noted, out of scope for this PR. The player context menu (`player_context_menu.ts`) already has whisper/invite/friend/ignore/report actions, so wiring those into the inspect window's menu path is a clean, separate follow-up I am happy to take.

PR #1031, opened 2026-06-29
feat(raid): standardize raid lockout to a daily us-east-1 reset

jgyy:
## What and why

Raid lockouts currently expire on a rolling "24 hours from kill" window, so the
unlock time drifts a little every clear and differs between raiders depending on
exactly when they got the kill. This standardizes the boundary to a single daily
reset at **midnight in the us-east-1 region's civil time (US Eastern,
America/New_York)**, so a guild can plan its raid night around one predictable
time and the whole raid unlocks together.

Same behavior as classic-era daily resets: a kill late in the evening unlocks at
the upcoming midnight rather than a full day later.

## How

- New pure leaf module `src/sim/instances/raid_reset.ts` exporting
  `nextRaidResetMs(nowMs)` (and `RAID_RESET_TIME_ZONE`). It computes the next
  US-Eastern midnight strictly after a given epoch-ms instant, resolving the zone
  offset twice so the boundary is correct across both DST transitions.
- `grantNythraxisLockout` now stores `nextRaidResetMs(ctx.lockoutNowMs())`
  instead of `ctx.lockoutNowMs() + 24h`. The wall-clock seam is unchanged: the
  host still supplies the instant via `SimContext.lockoutNowMs` (server:
  `Date.now()`; offline sim: the sim clock).

### Determinism / invariants

- The new module is a pure function of its input (no `Date.now`/`Math.random`/
  `performance.now`); it uses only `Intl.DateTimeFormat`, `new Date(ms)`, and
  `Date.UTC`, all pure. `tests/architecture.test.ts` (the sim purity + determinism
  guard) stays green.
- The lockout grant draws no RNG, so the global rng draw-order is unaffected. The
  only golden-trace change is the stored `until` value in
  `nythraxis_full_pull` (regenerated); the rng-draw-order log is identical.
- No new player-facing strings; the lockout tooltip already renders a generic
  time-remaining duration, so i18n is untouched.

## Diagram

lockout flow

## Screenshot

The affected raid-lockout tooltip, rendered from the real `raidLockoutPanelHtml`
view + `nextRaidResetMs` (kill at 21:45 ET shows time-to-reset, not 24h):

raid lockout tooltip

## Tests

- New `tests/raid_reset.test.ts` (determinism, EST/EDT, both DST transition days,
  the just-before-midnight and at-midnight boundaries).
- Updated the Nythraxis lockout test to assert the reset boundary and re-entry
  after it.
- `npm test` (sim/parity/lockout suites), `tsc --noEmit`, `npm run build`, and
  `biome ci --changed` all green locally. (`tests/admin/*` and a delve affix test
  fail only in my local worktree due to a symlinked node_modules; they pass in a
  clean `npm ci` and in isolation, and this change touches neither.)

Note: `tests/nythraxis_raid.test.ts` shows a large diff because Biome's
forward-ratchet reformats touched files; the only logic change there is the
lockout assertion.


EnriqueGF (commented):
This is a clean change and it follows the repo conventions well. The boundary logic lands as a pure leaf module behind the existing `SimContext.lockoutNowMs` seam instead of more logic on a coordinator, the two-pass offset resolution is the correct way to map a civil midnight to UTC across a DST transition, and the test coverage is genuinely thorough (both DST transition days, the year boundary, the just-before-midnight roll, and the at-boundary no-zero-length case).

Verified locally against the PR head:

- `tests/architecture.test.ts` green. The determinism guard bans only `Math.random` / `Date.now` / `performance.now`; the module uses `Intl.DateTimeFormat`, `new Date(ms)`, and `Date.UTC`, all pure functions of their args, so sim purity holds.
- `tests/parity` green. The golden `until` change (86431350 to 18000000) is the only sim-state delta and the rng draw-order log is unchanged, as the PR claims.
- `tests/raid_reset.test.ts` + `tests/nythraxis_raid.test.ts` green (63 tests). The updated lockout test drives a controllable `lockoutNowMs` and asserts both still-locked-before-reset and re-entry-after-reset, which is the right way to cover the real-time path.

Findings:

- nit (worth a conscious call), offline semantics, `src/sim/sim.ts:904`: in the offline Sim, `lockoutNowMs` defaults to the sim clock (`Math.floor(this.time * 1000)`), which is anchored to epoch 1970. So "next US-Eastern midnight" offline is computed relative to 1970-01-01, and the lockout duration now varies with the absolute sim clock rather than being a flat 24h. The new golden value 18000000 is exactly 1970-01-01 05:00 UTC, the first ET midnight after epoch, so a kill anywhere in the first five hours of sim-time unlocks at that same instant. On the server (`Date.now()`) the behavior is exactly what you want; this is purely an offline artifact and it stays deterministic, so it is not a bug. Just flagging that offline raid lockouts go from "always 24h" to "1970-anchored and uptime-dependent," in case you would rather offline keep wall-clock-independent semantics.

- note: the boundary correctness rests on the host ICU tz database for `America/New_York`, which the architecture regex cannot see. That is a safe assumption (full ICU in Node and browsers, stable ET rules), I only mention it because the determinism guard does not cover that surface.

Otherwise this looks good to me.

PR #1036, opened 2026-06-29
tooling(ci): non-blocking PR AI assist (screenshots + OpenRouter owl-alpha review)

EnriqueGF:
## What

Adds two informational, non-blocking GitHub Actions jobs in a new workflow (`.github/workflows/pr-ai.yml`), kept separate from the `ci.yml` gate so they never affect merge.

1. **Screenshots of changes** (`screenshots`): boots the Vite dev client headless (SwiftShader, no GPU), runs a fixed offline tour (character select, desktop HUD, mobile HUD) via the shared `enterOfflineGame` flow, uploads the PNGs as the `pr-screenshots` artifact, and links them from a sticky PR comment. Tour: `scripts/pr_screenshots.mjs`.
2. **AI review** (`ai-review`): sends the PR diff to an OpenRouter model and posts a short review as a sticky PR comment (Correctness / Invariants / Tests / Nits, severity-tagged). Reviewer: `scripts/ai_review.mjs`; sticky-comment helper: `scripts/gh_sticky_comment.mjs`. No new npm dependencies (Node `fetch` + the GitHub REST API).

## Setup

- The screenshots job needs no configuration.
- AI review is opt-in: add an `OPENROUTER_API_KEY` repo secret. Without it the job runs but no-ops and exits green, so this is safe to merge before the key exists.
- Optional `OPENROUTER_MODEL` repo variable to swap the model. Default is `openrouter/owl-alpha`, a free stealth model (1M context, tool use) that is free for now and can be pulled at any time, so it is a prototype default; switching is one variable, no workflow edit.

## Privacy

The free owl-alpha tier logs prompts to improve the model, so the diff sent is retained by a third party. Documented in `docs/ai-pr-bot.md`; point `OPENROUTER_MODEL` at a paid / non-logging or self-hosted model before using it on code you cannot disclose. The screenshots job sends nothing to a third party.

## Fork PRs

Fork PRs get a read-only token and no secrets, so the comment steps and the AI review degrade to a no-op there (handled in the scripts); the workflow never errors. Screenshots are still captured and uploaded as an artifact.

## Verification

- Ran the screenshot tour end to end against a local dev server: 3 frames captured, fully rendered HUD, exit 0.
- Confirmed the graceful paths: missing `OPENROUTER_API_KEY`, empty diff, and missing GitHub token all skip without failing.
- Biome (changed files), `node --check` on all scripts, and YAML validation all pass; no em/en dashes or emojis.

### DIFF
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -36,6 +36,9 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
+        with:
+          # Full history so the diff against the PR base drives change-aware capture.
+          fetch-depth: 0
 
       - name: Set up Node.js
         uses: actions/setup-node@v6
@@ -70,10 +73,21 @@ jobs:
             sleep 2
           done
 
+      # Change-aware capture: map the diff to the screens it implies (inventory, world
+      # map, ...). Only on pull_request; otherwise pr_screenshots falls back to the fixed
+      # tour. pr.diff is gitignored.
+      - name: Compute PR diff (for change-aware capture)
+        if: github.event_name == 'pull_request'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: git diff --no-color "$BASE_SHA" "$HEAD_SHA" > pr.diff || true
+
       - name: Capture screenshots
         env:
           GAME_URL: http://127.0.0.1:5173
           SHOTS_DIR: pr-shots
+          DIFF_FILE: pr.diff
         run: node scripts/pr_screenshots.mjs
 
       - name: Stop the dev client
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -1,0 +1,134 @@
+name: PR AI assist
+
+# Two INFORMATIONAL, NON-BLOCKING jobs that help review a pull request. Neither is a
+# required check; they assist review, they do not gate merge. They live in their own
+# workflow so they never interfere with the CI gate in ci.yml.
+#
+#   screenshots  boots the Vite dev client headless, runs a fixed offline tour
+#                (character select, desktop HUD, mobile HUD), uploads the PNGs as an
+#                artifact, and links them in a sticky PR comment.
+#   ai-review    sends the PR diff to an OpenRouter model (free openrouter/owl-alpha by
+#                default) and posts the review as a sticky PR comment. No-ops without the
+#                OPENROUTER_API_KEY secret (for example on a fork PR), so it never blocks.
+#
+# Setup (one time): add an OPENROUTER_API_KEY repo secret to enable ai-review. Optionally
+# set an OPENROUTER_MODEL repo variable to swap the model (owl-alpha is free but ephemeral).
+# See docs/ai-pr-bot.md, including the privacy note on the free owl-alpha tier logging prompts.
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+# Needed to post the sticky review/screenshot comments. Fork PRs get a read-only token,
+# so the comment steps degrade to a no-op there (handled in the scripts).
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: pr-ai-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  screenshots:
+    name: Screenshots of changes
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      # The client imports committed-but-derived i18n tables at boot; regenerate to be safe.
+      - name: Generate i18n artifacts
+        run: npm run i18n:gen
+
+      - name: Resolve a Chrome binary
+        run: |
+          CHROME="$(command -v google-chrome || command -v google-chrome-stable || command -v chromium-browser || command -v chromium || true)"
+          if [ -z "$CHROME" ]; then
+            echo "No system Chrome on PATH; installing a Chrome for Testing build."
+            CHROME="$(npx --yes puppeteer browsers install chrome | awk '{print $NF}')"
+          fi
+          echo "BROWSER_PATH=$CHROME" >> "$GITHUB_ENV"
+          echo "Using browser: $CHROME"
+
+      - name: Start the dev client
+        run: |
+          # strictPort: bind exactly 5173 or fail loudly, never silently bump the port.
+          npm run dev -- --host 127.0.0.1 --port 5173 --strictPort &
+          echo $! > .devpid
+          for i in $(seq 1 60); do
+            curl -sf http://127.0.0.1:5173 >/dev/null && { echo "dev client up"; break; }
+            sleep 2
+          done
+
+      - name: Capture screenshots
+        env:
+          GAME_URL: http://127.0.0.1:5173
+          SHOTS_DIR: pr-shots
+        run: node scripts/pr_screenshots.mjs
+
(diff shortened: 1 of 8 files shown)


EnriqueGF:
## AI review (`openrouter/owl-alpha`)

This change is a well-structured, purely additive PR: it introduces two informational, non-blocking GitHub Actions jobs (screenshots + AI review) and their supporting scripts, plus a `.gitignore` update and a design doc. None of it touches `src/`, so the sim-purity, server-authoritative, and `t()` invariants are not in scope here. Findings below.

## Correctness

- **`pr_screenshots.mjs` swallows the offline-select timeout and then immediately screenshots.** If `#offline-select .mini-class` never appears, `waitForSelector` is caught and the tour continues; the resulting `01-character-select` frame may be a blank or partial select screen with no class cards visible, and there is no note in the manifest distinguishing "selector never appeared" from "rendered fine." Medium.
- **`pr_screenshots.mjs` mobile branch wraps `goto` + `enterOfflineGame` in try/catch but the catch only records the error and falls through to `browser.close()`.** If the mobile `goto` itself fails (e.g. dev server down), the error is recorded but `captured` still contains the desktop frames, so the job still uploads a partial set — fine. But the `catch` also swallows a rejection from `page.emulate()` if it were ever added; not a real bug today. Low.
- **`ai_review.mjs` `listTouchedDirs` uses `git ls-files` with shell-escaped single quotes around each dir.** If a touched directory path contained a single quote, the command would break. Unlikely in this repo (no such dirs visible), but worth noting. Low.
- **`ai_review.mjs` reads `package.json` from cwd, not from the repo root.** In GitHub Actions the checkout step sets cwd to the repo root, so this is fine in CI; running locally from a subdirectory would return an empty deps list and the prompt would not include the dependency hint. Low.

## Invariants

- No `src/` files are touched, so no sim-purity, Rng, server-authority, or `t()` violations are possible from this diff. The new scripts live entirely under `scripts/` and `.github/`, which are explicitly exempt.

## Tests

- The diff adds no tests and changes no `src/` or server logic, so the "missing tests" rule does not apply. The new scripts are tooling and are not covered by the repo's existing test setup; this is consistent with how the repo treats other `scripts/*.mjs` files. No finding.

## Nits

- `pr_comment_shots.mjs` filters `l !== null` on a list that never contains `null` (every entry is a string or an empty string). Harmless but dead code. Low.
- `pr-ai.yml` `screenshots` job runs `npm run i18n:gen` but does not upload the regenerated i18n tables anywhere; if they differ from what's checked in, that drift is silently ignored. This is intentional (the job is informational), but worth being aware of. Low.
- `pr-ai.yml` `ai-review` job does `actions/checkout` with `fetch-depth: 0` (full history) only to run a single `git diff`. A `fetch-depth: 2` would be sufficient and faster on large histories. Low.
- `docs/ai-pr-bot.md` is a useful design doc and is referenced from the workflow comments; good. No finding.

Overall this is a clean, additive, well-documented change. The one medium finding (silent offline-select timeout producing a misleading screenshot) is worth addressing before relying on the tour for visual regression confidence.


<sub>Automated and non-blocking. May be wrong; a human review still decides. The free owl-alpha tier logs prompts, so the diff is retained by a third party.</sub>

EnriqueGF:
## Screenshots of changes (manual run)

The `screenshots` job does not run on a fork PR (read-only token), so these were captured locally by `scripts/pr_screenshots.mjs` against the offline client, exactly as the CI job would on an internal branch. Rendered headless via SwiftShader, no GPU.

**Character select**

character select

**Desktop HUD**

desktop HUD

**Mobile HUD**

mobile HUD

<sub>Demo of the screenshot capability for this PR. Frames hosted on the fork's pr-ai-bot-shots branch.</sub>

EnriqueGF:
## Change-aware screenshots (demo)

Proving the screenshot job can shoot **exactly what a diff touches**, not just a fixed tour. Driven by `scripts/pr_shot_targets.mjs` (a target registry: changed paths -> the screen they imply + how to open and crop it). Captured locally against the offline client (the CI job cannot post on a fork PR).

Simulated a diff touching `src/ui/bags_window.ts` (-> **inventory** target) and `src/sim/content/zones/eastbrook_vale.ts` (-> **world-map** target), then made a visible tweak to the bags title to prove the change shows up in the targeted shot.

**Inventory, before** (resolved from the bags path, cropped to `#bags`)

inventory before

**Inventory, after** the simulated change (red title bar is the tweak, captured automatically)

inventory after

**World map** (resolved from the zone path: teleports to Boar Meadow, opens and crops `#map-window`)

world map

<sub>Resolver: bags path -> inventory, zone path -> world-map, both -> both, unrelated -> fixed-tour fallback. Frames on the fork's pr-ai-bot-shots branch.</sub>

FernandoX7 (commented):
Really nicely done, and the security-sensitive parts are right where it counts: `on: pull_request` rather than `pull_request_target`, minimal `permissions:` (`contents: read` + `pull-requests: write`), and the secret-gated no-op so fork PRs and the key-less state both degrade green instead of erroring. Graceful degradation is consistent across all three scripts, the docs are honest about the owl-alpha prompt-logging tradeoff, no new deps, and it matches the `ci.yml` conventions (checkout@v6, setup-node@v6, node 22). Good change.

One thing to fix and a couple of small notes.

**should-fix** `scripts/ai_review.mjs:76`: ``execSync(`git ls-files -- ${dirs.map((x) => `'${x}'`).join(' ')}`)`` interpolates diff-derived file paths into a shell command with single-quote wrapping. Git does not quote a literal single quote in a path (it is printable ASCII), so a PR that adds a file like `a'$(...)'b/f.ts` breaks out of the quotes and runs an arbitrary command. I reproduced it: the path lands verbatim in the diff's `+++ b/` line, `listTouchedDirs` builds `git ls-files -- 'a'$(...)'b'`, and the substitution executes.

Today this is contained, not an escalation: on a fork PR there are no secrets and a read-only token, and the job already runs the PR's own copy of this script, so the injection grants nothing the attacker did not already have; on a same-repo PR the author is trusted. The reason to fix it anyway is that it is a latent landmine. It also silently breaks (the `catch` drops `repoFiles` to empty) on any legitimate path containing a quote, and it turns into a secret-stealing RCE the moment anyone switches this to `pull_request_target` to get fork-PR comments working, which is exactly the change people reach for. Drop the shell: `execFileSync('git', ['ls-files', '--', ...dirs])`. Same result, no quoting, immune to the path.

**nit** PR description: the auto-generated "Generated with Claude Code" footer carries a robot emoji, which the repo bans in PR text (the code, docs, and workflow themselves are all clean).

**nit** (optional) `.github/workflows/pr-ai.yml`: no `paths`/`types` filter, so the screenshots job runs `npm ci` + dev server + headless browser on every PR, including docs-only ones (which fall through to the fixed tour). Fine if intended; a `paths-ignore` would trim the cost.

Things I checked and am not raising, for the record: the fork-PR no-op paths hold (list-comments read works, then the POST 403s on a read-only token and returns null; the reviewer exits 0 with no key); line 76 is the only interpolated shell exec in the changed scripts (the screenshot scripts use diff paths only as JS substrings); and sim/wire/i18n are untouched, so the `scripts/` + `.github/` tooling is correctly exempt from t()/Rng/sim-purity.

EnriqueGF:
Thanks for the review, appreciate the thoroughness.

**should-fix (ai_review.mjs:76):** fixed in 26a771a8. `listTouchedDirs` now calls `execFileSync('git', ['ls-files', '--', ...dirs])` instead of shelling out through `execSync` with quoted string interpolation, so a diff-derived path can no longer break out of the quoting (verified locally with a `'$(...)'`-style path: it now lands as a literal argv entry, no shell parsing).

**nit (PR description emoji):** noted, leaving the description as is for now.

**nit (paths-ignore on the screenshots job):** leaving this one alone deliberately. As you called out yourself, it is fine as intended: the change-aware capture needs the diff on every PR to decide between a targeted shot and the fixed-tour fallback, so filtering by path would just make docs-only PRs silently skip that decision rather than save meaningful cost. Happy to revisit if the run cost becomes a real problem.

The other two things you checked and didn't raise (fork-PR no-op paths, no other interpolated shell exec in the changed scripts) still hold after this fix, nothing else in the diff touches a shell.

FernandoX7 (approved):

PR #1045, opened 2026-06-29
feat(admin): add live suspicious-player review dashboard

madmatah:
## Summary

Adds an admin investigation page for reviewing online players with active bot-detection evidence.

The new page:

- displays each player's name, IP address, evidence count, score, and evidence details;
- opens account details from the player name and IP exploration from the address;
- sorts by score or evidence count;
- supports optional 5-second auto-refresh, persisted in the browser;
- keeps evidence groups compact and visually separated for large datasets.

This also improves the Shared IPs page with server-side sorting by account count or last-seen time. Sorting is applied before pagination and works for both historical and online-only results.

The endpoint remains protected by the existing admin authentication layer.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:

`npm run check:types && npm run check:admin && npm test && npm run ci:changed &
& npm run build`

- Manual steps:
  - launch bot script and use private bot detector implementation
  - reviewed the Suspicious Players page in Chrome with representative mocked data;
  - verified account and IP navigation reuse the existing admin flows;
  - verified auto-refresh preference persistence;
  - verified Shared IP sorting in both directions.

## Screenshots / recordings

Screenshot from 2026-06-29 15-23-11

Screenshot from 2026-06-29 16-02-26

---

### Quality

- [x] **Typecheck passes.** `npx tsc --noEmit` and the admin Svelte checks are clean.

### Cross-platform

- [x] **Tested on desktop and mobile.**
- [x] **Accessible.** Sort controls and toggles are keyboard-operable, have visible focus states, and respect reduced-motion preferences.

### Localization (i18n)

- [ ] **New operator-visible strings are translated into every supported locale.**
- [x] Numbers and dates use locale-aware formatting.
- [x] N/A. No player-facing text is emitted from `src/sim/` or `server/`.

### Hygiene

- [x] No secrets, credentials, or production dev-command flags are included.
- [x] Generated admin i18n files were regenerated rather than edited manually.


server/admin_db.ts
```diff
@@ -588,7 +601,7 @@ export async function listSharedIps(page: number, limit: number): Promise<Pagina
      )
      SELECT *, count(*) OVER ()::int AS total
      FROM shared
-     ORDER BY account_count DESC, last_seen_at DESC, ip
+     ORDER BY ${orderBy}
      LIMIT $1 OFFSET $2`,
     [limit, offset],
```
AlanGreyjoy: SQL injection, use parameterized queries only.

AlanGreyjoy (changes requested):

FernandoX7 (commented):
Clean, self-contained admin feature, and the wiring is careful where it counts.

What's solid:

- The new endpoint sits behind the existing admin gate (`handleAdminApi` returns 401 before the route block at `server/admin.ts:174`), and the test asserts an authed admin is required.
- The shared-IP `ORDER BY` is built from a closed whitelist (`sort` and `dir` validated to enums in `sharedIpSortParams`) with the values still parameterized, so there is no injection surface, and the live-path `sortSharedIpRows` mirror matches the SQL tiebreak order exactly. `lastSeenAt` is an ISO string on the live path (`live_shared_ips.ts:37`), so the `localeCompare` ordering is chronological. Both comparators carry a unique final key (ip, accountId), so they are total orders.
- Stale responses are guarded by `requestId` in both pages, and the auto-refresh interval is torn down through the `$effect` cleanup.
- The new table uses real `<button>` headers with `aria-sort`, which is a step up from the clickable `<th>` in CharactersTable.
- Good test coverage: endpoint plus auth, the SQL order in both directions and the default, online vs historical paths, nav routing and the nav link, and the page sort and persistence behavior.

Findings:

- nit: `sharedIps.description` is now orphaned. The page switched to `sharedIps.allDescription` and nothing else references the old key, so it and its translations are dead weight. Safe to drop. `src/admin/i18n.en.ts:420`.
- nit: the Shared IPs sort headers are buttons inside a `<div>` pseudo-table, so the current sort state is not exposed through `aria-sort` the way the new table does it. The arrow glyph still conveys direction when read, so this is minor, but SuspiciousPlayers is the better model if you want to align them. `src/admin/pages/SharedIps.svelte:95`.

One note, not a change request: the evidence `kind` and `detail` render as raw server English from the bot detector, bypassing the admin `t()` matcher. That matches how the dashboard surfaces other server data and the values are dynamic, so keying them is not really feasible. Flagging only so it is a known gap.

madmatah:
@AlanGreyjoy @FernandoX7 Thanks for the review, your feedbacks have been addressed.
@Rubsey Rebased and ready to merge

FernandoX7 (approved):

PR #1046, opened 2026-06-29
fix(render+ui): close the "Could not start the renderer" entry crashes (tier-independent GLB preload + play.html #actionbar2)

EnriqueGF:
## P0: "Could not start the renderer" on world entry (v0.16.0)

Players intermittently hit the red `Could not start the renderer: try reloading.`
overlay and cannot enter the world. Root cause is a time-of-check / time-of-use bug
in the renderer's asset preloading.

`src/render/props.ts` and `src/render/characters/assets.ts` each freeze their GLB
**preload** set at module-import time from an early graphics-tier guess
(`GFX.standardMaterials`). But both **place** their assets later, against the **live**
tier the Renderer resolves in `initGfxTier()` (which reassigns the `GFX` global from
the real WebGL context, after import). When the import-time guess came in lower than
the render tier (a weak or hybrid-GPU probe guesses `low`, the high-performance
renderer then resolves `medium`+), an asset was placed that the lower import tier
never preloaded, and the synchronous accessor threw, surfacing as `rendererFailed`.

This started in v0.16.0: the device-aware first-run preset (`f275462f`) made the
import-time tier resolve to `low` for the first time, opening the divergence window.

## What this fixes

- **`prop asset not preloaded: farmCrate`** (and 12 other props placed only at
  `medium`+: `anvil`, `weaponStand`, `mushroomTan`, the `statue*`, the non-round
  gravestones, the tall/large rocks, `lanternWall`). `props.ts` now preloads **every**
  entry in `PROP_ASSET_DEFS` regardless of tier; the low subset stays only to scope
  the material prewarm.
- **`character asset not preloaded: .../rogue_hooded.glb`** (the `mob_bandit` body,
  which is the humanoid-family default **and** the global mob fallback, so it is placed
  on essentially every populated map). A dedicated audit found `characters/assets.ts`
  has the identical TOCTOU, so it would still crash entry on hybrid GPUs after the props
  fix alone. `characters/manifest.ts` + `assets.ts` now preload the **union** of both
  tiers' resolved URLs.

Both preload sets are now tier-independent supersets of any placement set, the way
`src/render/foliage.ts` has always been immune (one frozen list feeds both preload and
placement). Because every placement key is typed (`PropKey` / a `manifestUrls()` URL),
the full/union set is a provable superset of anything the renderer can place, on every
tier and device, so the fix is cross-platform by construction (it removes the
dependency on tier resolution entirely; it does not try to detect each GPU correctly).

**Deliberate cost:** weak devices now fetch the full prop catalog plus the character
union (the extra `medium`+ props and the high-tier bodies) instead of a tier subset.
That is the conscious price of dropping the tier dependency, and it matches `foliage.ts`.

Second commit (`perf`): the prop **material prewarm** also read the import-frozen tier.
It runs after `initGfxTier`, so on the same hybrid scenario it warmed only the low
subset while the renderer drew the full catalog, leaving the `medium`+ props to take a
first-frame shader-compile hitch. It now keys off the live tier. Perf only, not a crash.

## Second entry crash, on the `/play` page: missing `#actionbar2`

A reviewer caught that this PR's original "verification" section was wrong about the
`#actionbar2` `appendChild` crash. It said the second action bar "does not exist in any
branch." That was a **stale read**: this branch was cut from `13f5418f` (#1032), an
ancestor that predates `9ec87a59` (#983, the second action bar), so the re-check ran
against a tree that genuinely lacks the feature. On the real base (`release/v0.17.0`)
the bug is live: `index.html` carries `<div id="actionbar2">`, `hud.ts` `buildActionBar`
resolves `$('#actionbar2')` and `appendChild`s slots 12..22 onto it, but `play.html`'s
`#game-ui-template` shipped without that div, so the `/play` entry deterministically
throws `Cannot read properties of null (reading 'appendChild')` on world entry, the same
"Could not start the renderer" overlay, on that page only.

This PR now merges `release/v0.17.0` to drop the stale branch point and fixes it:

- `play.html`: add `<div id="actionbar2" class="panel"></div>`, mirroring `index.html`.
- `hud.ts`: guard the append (`container?.appendChild(btn)`) so any entry whose template
  omits the second bar leaves those buttons detached instead of crashing (keybind
  dispatch by slot still works).
- `tests/login_parity.test.ts`: extend the renderer-critical markup parity guard to
  cover `#actionbar2`, so a future entry that drops it fails at build time. This was
  written test-first: **red** on `play.html` before the div, **green** after.

The "Honor a stored preset as authoritative" handoff item is still **not** applied as
written: returning `null` from `firstRunGraphicsPreset` whenever a preset is stored would
re-introduce exactly the bug the `graphicsDefaultApplied` marker exists to avoid
(`graphicsPreset` is def-filled to `medium` the moment any unrelated setting saves, so
"a preset is stored" is true long before the player ever chooses). The preload fix makes
the crash impossible regardless of how the tier resolves, so that cleanup is a follow-up.

## Tests

- New `tests/render_asset_preload.test.ts`: asserts both preload sets cover placement at
  every import-time tier, in particular the lowest. Written test-first: **red** before
  the fix (`preloadPropKeys(false)` omitted `farmCrate`; `characterPreloadUrls(false)`
  omitted `rogue_hooded.glb`), **green** after.
- `tests/login_parity.test.ts`: now also pins `#actionbar2` to index/play parity.
- `tsc --noEmit` clean, `biome check` clean on the touched files (pre-existing
  `noNonNullAssertion` warnings in `hud.ts` are not on the changed lines and do not gate),
  the play.html / HUD parity suite green (102 tests).

## Follow-ups (not in this PR)

- Settings: redesign the "explicit player graphics choice is never overridden" guarantee
  for pre-marker / post-reset storage, using the `graphicsDefaultApplied` marker rather
  than preset presence.
- Perf nit: `MERGE_BAND_DEPTH` is still import-frozen (draw-call granularity only, no
  crash, visually identical).
- Test hardening: enumerate real character placement (iterate `VISUALS`) in the guard
  rather than trusting `manifestUrls()`, so a future eagerly-placed `lazyPreload` def is
  also caught (the mech is safe today via the renderer's `mechAssetsReady()` guards).

## Base branch

Targets `release/v0.17.0` (rebased from the earlier `release/v0.16.0` target): the
implicated commit `f275462f` and both crashes live on the release line; `main` does not
yet contain the device-aware preset code, so the P0 cannot be fixed there. Will need
forward-porting to `main` once it catches up. The branch now merges `release/v0.17.0`
directly, so the diff reflects the real base and the `/play` crash is no longer read
against a stale tree.


FernandoX7 (commented):
Nice fix. The TOCTOU diagnosis is right, and making both preload sets tier-independent supersets (the way `foliage.ts` already is) is the durable choice over trying to get per-device GPU tiering correct. I verified the core against the code on the base:

- Props: every placement key is statically `PropKey` (`addParts`/`addInstance`/`propAsset` all take `PropKey`, and the dynamic arrays `housePool`/`graveKinds`/`rockKinds`/`kind`/`asset` are `PropKey[]`), so `ALL_PROP_KEYS` is a provable superset of anything `buildProps` can place.
- Characters: for any non-lazy base url in `manifestUrls()`, the live tier resolves to either `url` or `LOW_URL_ALIAS[url]`, and the union of both `manifestUrlsForGraphics` tiers contains both, so it covers any tier. `rogue_hooded.glb` (the `mob_bandit` body) is the real diverging key. Mech is `lazyPreload` and guarded by `mechAssetsReady()`, so it stays out of scope correctly.
- Prewarm: `buildPropMaterialPrewarmGroup` runs after `initGfxTier` (renderer.ts:899 before 1135/2463), so keying it off the live `GFX.standardMaterials` is right.
- The new test passes (4/4), `tsc --noEmit` is clean, and the merge into `release/v0.17.0` is conflict-free.

Findings:

**should-fix (the bug #2 dismissal is wrong for this base): play.html + src/ui/hud.ts:3564.** The verification section states the `#actionbar2` `appendChild` crash "does not exist in any branch or the working tree" and that `9ec87a59` / #983 "is not present here." That is not true for this PR's base. On `release/v0.17.0`: `index.html:323` has `<div id="actionbar2" class="panel">`, `play.html` has no `actionbar2`, `hud.ts:608-609` is the 11/22 split, and `hud.ts:3564,3569` is `const bar2 = $('#actionbar2'); ... container = i <= 11 ? bar : bar2`. `9ec87a59` (#983) is an ancestor of `release/v0.17.0` but not of this PR head, which branched from `13f5418f` (#1032), a commit that predates the second action bar, so the re-check ran against a tree that genuinely lacks the feature. Scoping this PR to the preload P0 is fine, but bug #2 is a real deterministic crash on the `/play` entry that the merge leaves in place. Please do not record it as nonexistent: it still needs the `#actionbar2` div in `play.html`'s `#game-ui-template` plus a `bar2` null-guard in `buildActionBar`.

**nit, PR metadata.** The body says "Targets `release/v0.16.0`" but the base is `release/v0.17.0` (both carry the crash, so the fix lands either way). The branch point is `13f5418f`, an older common ancestor rather than either release tip, which is what made the bug #2 re-check read a stale tree.

**nit (already self-noted).** The character guard derives `union` from the same `manifestUrlsForGraphics` that `characterPreloadUrls` uses, so that half is close to tautological and does not independently prove `manifestUrls()` covers real placement. Your follow-up (iterate `VISUALS`) is the right closer; fine to defer.

**note.** Weak devices now fetch the full prop catalog plus the character union (the extra props and the high-tier bodies). That is the deliberate cost of dropping the tier dependency and it matches `foliage.ts`; one line in the description would make it read as a conscious call rather than a surprise.

Deferring the "honor a stored preset" cleanup is reasonable: the superset preload closes the crash regardless of how the tier resolves, so it is no longer urgent.

EnriqueGF:
Thanks, you are right on the should-fix and the diagnosis is exact. The `#actionbar2` crash is real on the base; my original verification note read a stale tree because this branch was cut from `13f5418f` (#1032), which predates `9ec87a59` (#983). Fixed in this PR:

- Merged `release/v0.17.0` into the branch so it no longer reads against the stale branch point (also closes the branch-point nit).
- `play.html`: added `<div id="actionbar2" class="panel"></div>`, mirroring `index.html`.
- `hud.ts` `buildActionBar`: guarded the append (`container?.appendChild(btn)`) so any entry whose template omits the second bar degrades to detached buttons instead of throwing.
- `tests/login_parity.test.ts`: extended the renderer-critical markup parity guard to cover `#actionbar2` (red on `play.html` before the div, green after), so a future entry that drops it fails at build time.

On the nits: corrected the base metadata in the description (`release/v0.17.0`) and added an explicit "deliberate cost" line about weak devices fetching the full prop catalog plus the character union. The tautological half of the character guard stays deferred to the `VISUALS`-iteration follow-up as you suggested.

FernandoX7 (approved):

PR #1047, opened 2026-06-29
feat(ui): make the second action bar opt-in (single bar by default)

ryan-foo:
The HUD now shows a single action bar out of the box. The second action bar (#actionbar2, slots 12..22) becomes opt-in via a new Options > Interface toggle "Show Secondary Action Bar" (off by default).

## What changed
- **settings.ts**: new `showSecondaryActionBar` bool (default `false`), persisted to localStorage.
- **options_view.ts**: a boolToggle row in `buildInterfaceControls` (Interface tab).
- **main.ts**: `applySetting` toggles a `body.show-actionbar2` class (also applied on startup from the persisted/default value).
- **hud.css**: `#actionbar2` is hidden by default; `body.show-actionbar2 #actionbar2` reveals it. Mobile keeps a single bar regardless (the existing `body.mobile-touch` rule is imported after and wins the tie).
- **i18n**: English label `hudChrome.options.showSecondaryActionBar` plus the five non-Latin overlay fills the completeness gate requires; Latin locales English-filled for the maintainer to translate at release.

## Notes
- The slots stay bound either way: keybinds route through `hud.castSlot` independent of DOM visibility, so hiding the row never disables the abilities on it.
- Verified locally: `tsc --noEmit` clean, `biome ci --changed` clean, and the options/settings/i18n-completeness/localization-coverage suites pass.


ryan-foo:
Per player request

FernandoX7 (commented):
Tightly scoped and the mechanism holds up. I checked the load-bearing claims against the code rather than the description:

- The body class is applied on boot, not just on toggle. `settings.all()` returns the merged set with defaults filled, and the existing apply-all loop (`main.ts:1479`) calls `applySetting` for every key, so a returning player who persisted `true` gets `show-actionbar2` on load; default-false users hit a no-op toggle and the CSS hides the row anyway.
- Mobile stays single-bar robustly. `body.mobile-touch #actionbar2 { display: none }` lives in `@layer hud-mobile`, ordered after `@layer components` (where the reveal rule sits), so it wins by layer precedence regardless of specificity.
- play.html parity is fine: both entries load `/src/main.ts`, which imports the styles barrel and runs the same apply-all loop, so the CSS rule and the body class both reach `/play`.
- Keybind-reachability checks out: slots 12..22 dispatch by slot index (`hud.ts:3565`), independent of DOM visibility, and `display:none` keeps the buttons in the DOM.
- English i18n is in the right place (`hudChrome.options.showSecondaryActionBar`, the English-only catalog domain) rendered through `boolToggle` to `t()`, and the dispatch-matrix test is updated to match. tsc is clean and the i18n regen is idempotent.

Two small notes, neither blocking:

nit `hud.css:1089`: the comment explains the mobile win as hud.mobile.css being "imported after this file so it wins the same-specificity tie." The two rules are in different `@layer`s (hud-mobile after components), so it is layer order, not source-order tiebreak, that decides. The result is correct and in fact more robust than the comment implies, but the wording could lead a future reader to think same-layer source order is load-bearing here.

note: this flips the default for existing players, not just new ones. Anyone who had abilities slotted on the second bar will see it disappear on next load until they enable the toggle. The abilities stay castable via their keybinds so nothing breaks, but the visible cooldown and icon state for those slots goes away. That is the stated intent of the change; just flagging it as a conscious on-upgrade behavior shift.

PR #1048, opened 2026-06-29
feat(ratelimit): distinct error codes for the per-IP vs per-account limiters

ryan-foo:
## Problem

The login/register rate limiters both surface as a generic "too many attempts" 429, and the client disambiguates them only by **fragile English prefix-matching**:

- The **per-IP** limit (20 req/min, `rateLimited`) and the **per-account** lockout (10 failed sign-ins / 15 min, `authThrottled`) are different things with very different wait times, but a player cannot tell which one they hit.
- The per-account 15-minute lockout renders the per-IP **"wait a minute"** copy, which is misleading.
- The wording had drifted across call sites (`account.ts` "too many attempts, slow down" vs `main.ts` "too many attempts ... wait a minute"), one carrying a stray em dash.
- The 429 body carried no machine-readable `code`, and the client dropped any `code` field, so a future wording change would silently break the client mapping.

## Change

- **`server/ratelimit.ts`**: two centralized 429 bodies, each with a stable `code` and an explanatory message naming which limiter fired and why:
  - `RATE_LIMIT_IP` -> `code: "rate_limited_ip"`
  - `RATE_LIMIT_ACCOUNT` -> `code: "rate_limited_account"`

  Used at every `rateLimited()` / `authThrottled()` site in `main.ts` and `account.ts`, so the code and copy cannot drift again.
- **`src/net/online.ts`**: `ApiError` now captures the server's `code`.
- **`src/main.ts`**: `userFacingApiError` keys off the `code` to pick the right localized message (per-account -> `moderation.tooManyFailed`, per-IP -> `errors.api.tooManyAttempts`), keeping the existing text matchers as a fallback for deploy skew. The English prefixes are preserved so the fallback still works.
- **`tests/ratelimit_codes.test.ts`**: the distinct codes, the text-prefix contract the client fallback relies on, and that each limiter trips independently (per-IP scoped to one IP, per-account scoped to one account).

## Testing

- New `tests/ratelimit_codes.test.ts` + `tests/account_server.test.ts` + `tests/login_parity.test.ts` + `tests/localization_fixes.test.ts` (S3 guard) all pass.
- `npm run check:ts` clean; `biome format` applied; `npm run build:server` bundles.

## Follow-up (separate PR)

The deeper cause behind the original report: the limiter resolves the client IP from `X-Forwarded-For` only and never reads `CF-Connecting-IP`. Behind Cloudflare, if `TRUSTED_PROXY_IPS` does not list Cloudflare's edge ranges, `requestIp()` keys on a shared Cloudflare edge IP, so many users collapse into one bucket and get false-positive rate limits. Worth a separate change (prefer `CF-Connecting-IP` when behind a trusted proxy, and document the `TRUSTED_PROXY_IPS` requirement).

### DIFF
--- a/server/account.ts
+++ b/server/account.ts
@@ -42,7 +42,7 @@ import {
   makeEmailToken,
 } from './email';
 import { json, readBody } from './http_util';
-import { clearAuthFailures, rateLimited, recordAuthFailure } from './ratelimit';
+import { clearAuthFailures, RATE_LIMIT_IP, rateLimited, recordAuthFailure } from './ratelimit';
 import {
   generateRecoveryCodes,
   generateSecret,
@@ -99,7 +99,7 @@ export async function handleAccountChangePassword(
   accountId: number,
   callerToken: string,
 ): Promise<void> {
-  if (rateLimited(req)) return json(res, 429, { error: 'too many attempts, slow down' });
+  if (rateLimited(req)) return json(res, 429, RATE_LIMIT_IP);
   const body = await readBody(req);
   const acct = await accountById(accountId);
   if (!acct) return json(res, 404, { error: 'account not found' });
@@ -157,7 +157,7 @@ export async function handleAccountDeactivate(
   accountId: number,
   hooks: AccountGameHooks,
 ): Promise<void> {
-  if (rateLimited(req)) return json(res, 429, { error: 'too many attempts, slow down' });
+  if (rateLimited(req)) return json(res, 429, RATE_LIMIT_IP);
   const body = await readBody(req);
   const acct = await accountById(accountId);
   if (!acct) return json(res, 404, { error: 'account not found' });
@@ -191,7 +191,7 @@ export async function handleAccountEmailChange(
   res: http.ServerResponse,
   accountId: number,
 ): Promise<void> {
-  if (rateLimited(req)) return json(res, 429, { error: 'too many attempts, slow down' });
+  if (rateLimited(req)) return json(res, 429, RATE_LIMIT_IP);
   const body = await readBody(req);
   const acct = await accountById(accountId);
   if (!acct) return json(res, 404, { error: 'account not found' });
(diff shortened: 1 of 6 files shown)


FernandoX7 (commented):
Clean, well-scoped fix. Centralizing the two 429 bodies in `ratelimit.ts` and keying the client off a stable `code` is the right call: it kills the per-account-lockout-shows-"wait a minute" bug at the source, and the code and copy can no longer drift across the many call sites. Good touches too, the `code` path and the preserved text-prefix fallback resolve to the same keys (`moderation.tooManyFailed` for the account lockout, `errors.api.tooManyAttempts` for the IP limit), the em dashes in the old copy are gone, and the deeper `CF-Connecting-IP` / `TRUSTED_PROXY_IPS` cause is correctly carved out as a separate change.

I checked the parts that could bite, and they hold:

- All four `throw new ApiError` sites in `src/net/online.ts` now forward `data.code`, so nothing drops it silently, and the arg is optional so older bodies still construct fine.
- The text fallback ordering is safe. The account message ("too many failed attempts...") does not match the earlier `startsWith('too many attempts')` arm at `src/main.ts:270`, so it correctly reaches the `too many failed attempts` arm at `src/main.ts:345`. Both arms map to the same keys as the new code path, so the code and fallback render identically.
- `tests/ratelimit_codes.test.ts`, `account_server.test.ts`, and `login_parity.test.ts` pass at head. The new test covers the distinct codes, the prefix contract the fallback relies on, and each limiter tripping independently from a public IP.

One nit:

- nit `server/main.ts:690-691`: the comment "The message is identical to a bad-password response so it never reveals whether the account exists" is no longer literally true now that the throttle returns `RATE_LIMIT_ACCOUNT` (429), not the 401 bad-password body. The enumeration-safety still holds, but for a different reason: the throttle keys on the submitted username and `recordAuthFailure` fires for missing accounts too (`server/main.ts:697`), so the 429 comes back identically for existent and non-existent usernames. Worth retuning the comment to say that, so a later reader does not "fix" the code to match the stale wording and lose the clearer copy.

PR #850, opened 2026-06-22
fix(render): smooth camera-driven player facing instead of snapping

jgyy:
## Problem
The character model **snaps/rotates wrongly**:
- It **instantly rotates backwards** when it should be facing forward.
- It **snaps to a direction** when you hold right-mouse to rotate the camera.

## Root cause
The player model's facing can be driven by the camera yaw, in two modes:
- classic **right-mouse mouselook**, and
- the always-on **Mouse Camera mode** while a movement key is held.

The renderer applied that override as a raw assignment at `renderer.ts`:

```ts
let facing = e.prevFacing + shortestAngle(e.prevFacing, e.facing) * ea;
if (id === p.id && renderFacingOverride !== null) facing = renderFacingOverride; // instant
v.group.rotation.y = facing;
```

The angle math itself is correct (camera-forward `(sin camYaw, cos camYaw)` matches model facing `(sin f, cos f)` , no 180 deg convention error). The defect is that the assignment is **un-rate-limited**. While the override is *disengaged* (standing still in Mouse Camera mode, or before a right-drag crosses the camera-drag threshold) the player freely orbits the camera while the model keeps showing the interpolated sim facing, so the two diverge by up to 180 deg. The instant the override engages, the model discharges that whole gap in a single frame: the "instant backward snap".

## Fix
Rate-limit the **local player's** camera-driven yaw to a max angular velocity (10 rad/s), seeded from the live interpolated facing on first engage. The cap sits well above any intentional input (keyboard `TURN_SPEED` is PI rad/s; a normal mouse drag is only a few degrees per frame), so ordinary turning passes straight through and only a discontinuity (or a violent flick) is smoothed. The model now rotates smoothly to follow the camera instead of teleporting.

- Scoped to `id === p.id`; mobs and other players keep their existing interpolation.
- **Presentation-only**: no sim, wire, `IWorld`, or i18n change.
- Angle math extracted into a pure, host-agnostic module `src/render/facing_smooth.ts` with unit tests.

## Verification
- `tests/facing_smooth.test.ts` (TDD, red then green): shortest-path wrap, per-frame cap, multi-frame convergence, sub-cap pass-through, long-frame clamp.
- Full suite: **3059 passed**, 8 skipped, `npm run build` green.
[...]

PR #1053, opened 2026-06-30
fix(game): commit final camera yaw on mouselook release (camera/character snap)

jgyy:
## Problem

Holding the right mouse button to rotate the camera, then briefly releasing, leaves the character a fraction of a turn behind the camera. Both rotate as intended, but the character stops short of where the camera ended. Then, on the next forward step, the follow camera backtracks to sit behind the character, which reads as a snap. (Holding W while doing a quick RMB flick shows the same snap.)

## Root cause

While classic right-mouse mouselook is active, `resolveMove` commits `facing = input.camYaw` every sim tick, so the player's facing tracks the camera. Input is sampled at frame rate (about 60 Hz) but facing only commits on a 20 Hz sim tick, so the slice of mouse motion between the last tick and the release moment has nowhere to land: the instant mouselook turns off, `movementFacing` and `resolved.facing` both go `null` and that final fraction of the turn is dropped. PR #850's `releaseSelfFacing` then faithfully eases the model back to the sim facing, which is exactly that stale value, so the gap is preserved rather than caused.

```mermaid
sequenceDiagram
    autonumber
    participant ML as mouselook (RMB)
    participant SF as sim player.facing (20Hz)
    participant CAM as follow camera
    ML->>SF: each tick: facing = camYaw
    Note over ML,SF: mouse keeps moving after the last tick
    ML-->>SF: BEFORE: release -> facing intent null,<br/>final slice dropped, facing left behind camera
    CAM->>CAM: forward step -> backtrack behind stale facing (SNAP)
    ML->>SF: AFTER: falling edge -> commit final camYaw once
    CAM->>CAM: aligned, no backtrack, no snap
```

sequence

## Fix

A small pure module, `src/game/mouselook_release.ts`, returns the camera yaw on the falling edge of mouselook and `null` otherwise. `main.ts` folds it into `movementFacing` **after** the render and controller overrides (`renderFacing ?? controllerFacing ?? pendingReleaseFacing`), so it only fires when nothing else owns the heading.

The yaw is **latched** rather than applied only on the release frame: offline a sim tick runs on only about 2/3 of frames (60 Hz frames, 20 Hz ticks), so a release landing on a zero-tick frame would otherwise drop the one-shot. The latch is cleared once a tick consumes it (offline) or the per-frame input stream sends it (online), and whenever the camera or controller reclaims the heading.

The synthetic trace below (driven by the real edge logic) shows a fast RMB flick: AFTER (green) lands on the camera yaw at release; BEFORE (red) freezes a tick behind it.

before/after

Scoped deliberately to classic right-mouse mouselook; the always-on Mouse Camera mode has its own facing path and is unchanged.

## Why it is safe

- Presentation/input layer only: no new sim system, no rng, no wire/`IWorld`/i18n change (no S3 guard or locale work). It extends an existing facing intent by one frame.
- Same commit path offline (`offlineSim.player.facing = stepFacing`) and online (`net.setMouselookFacing`), so both hosts align.

## Tests

- `tests/mouselook_release_facing.test.ts` (TDD red -> green): commits only on the falling edge, never while held, engaging, or idle.
- `tsc`, `tests/camera_follow.test.ts`, `tests/facing_smooth.test.ts`, `tests/architecture.test.ts` green; `npm run build` green; `npm run ci:changed` clean.

### DIFF
--- a/src/game/mouselook_release.ts
+++ b/src/game/mouselook_release.ts
@@ -1,0 +1,20 @@
+// Classic right-mouse mouselook drives the player's heading: while it is active
+// the per-tick facing intent is set to the camera yaw (camYaw), so the sim facing
+// tracks the camera. Input is sampled at frame rate but facing only commits on a
+// sim tick, so the slice of mouse motion between the last tick and the release is
+// never written: the instant mouselook turns off the facing intent goes null and
+// that final fraction of the turn is dropped. The character then settles back to a
+// facing slightly behind the camera, and moving forward makes the follow camera
+// backtrack to sit behind it, which reads as a snap.
+//
+// This is a single, DOM-free decision: on the FALLING edge of mouselook, commit
+// the current camera yaw one last time so the sim facing ends exactly where the
+// camera ended. It is deliberately scoped to the classic right-mouse mouselook
+// edge; the always-on Mouse Camera mode has its own facing path and is not in play.
+export function mouselookReleaseFacing(
+  prevMouselook: boolean,
+  mouselook: boolean,
+  camYaw: number,
+): number | null {
+  return prevMouselook && !mouselook ? camYaw : null;
+}
--- a/src/main.ts
+++ b/src/main.ts
@@ -39,6 +39,7 @@ import {
   setInterfaceMode,
   useTouchInterface,
 } from './game/mobile_controls';
+import { mouselookReleaseFacing } from './game/mouselook_release';
 import { music } from './game/music';
 import { createPerfMonitor } from './game/perf';
 import { startPerfReporter } from './game/perf_reporter';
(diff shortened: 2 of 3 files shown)


FernandoX7 (commented):
Nice fix, and the writeup matches the code. While classic mouselook is held `renderFacingOverride()` returns `camYaw` (main.ts:2059) so facing tracks the camera each tick; on release that override drops to null and the last sub-tick slice of the turn has nowhere to land. Committing the final `camYaw` on the falling edge is the right minimal correction, and keeping it scoped to classic RMB mouselook (Mouse Camera mode keeps its own path) is correct.

What I verified beyond the diff:

- The decision is a DOM-free pure function, unit-tested, with main.ts left a thin consumer: the seam this repo asks for. The `?? ... ?? pendingReleaseFacing` chain plus the `!== null` guards correctly preserve a facing of `0` (north).
- The latch holds the yaw captured at release, not live `camYaw`, which is also right for the free-look case (orbiting the camera after release must not drag the character), and a concurrent A/D turn still lands because `resolveMove` returns `facing` null when not mouselooking (main.ts:1942) and the sim layers the turn input on top of the facing you set before the tick. No wire/IWorld/i18n surface touched; offline and online commit paths align.
- On the head ref: `tsc` clean; the new test (6) plus `camera_follow` / `facing_smooth` / `architecture` (48) all green.

One note, not blocking:

- nit: the subtle part here is the latch lifecycle (set on the edge at main.ts:2127-2131, cleared when a tick consumes it at 2161-2163, when the online stream sends it at 2205-2207, or when render/controller reclaim the heading), and it all lives in main.ts, so it is verified by reasoning rather than a test. That is inherent to main.ts having no seam, so not something to block on; if you ever wanted coverage, the set/clear/consume decision is itself a small pure state function you could lift out and test the way you did the edge.

FernandoX7 (approved):

PR #1054, opened 2026-06-30
fix(native): release preview context before game entry

TrevCavill:
## Summary
- destroy the character-select preview when entering the loading state
- make CharacterPreview.destroy() release animation, listeners, visual resources, and its WebGL context
- reduce native iOS WebView/WebGL pressure before the main renderer starts

## Tests
- npx tsc --noEmit
- npx vitest run tests/client_shell.test.ts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "world-of-claudecraft",
-  "version": "0.14.1",
+  "version": "0.16.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "world-of-claudecraft",
-      "version": "0.14.1",
+      "version": "0.16.0",
       "license": "MIT",
       "dependencies": {
         "@capacitor/live-updates": "^0.5.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "world-of-claudecraft",
-  "version": "0.16.0",
+  "version": "0.14.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "world-of-claudecraft",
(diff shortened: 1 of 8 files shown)


FernandoX7 (commented):
Solid fix. The teardown is correct and matches the existing context-release pattern: capturing the `trackWebGLContext` unregister handle and calling it before `forceContextLoss()` + `dispose()` is exactly the "call it if the renderer is disposed earlier so it is not touched twice" contract in `context_release.ts`, and the order mirrors `releaseTrackedWebGLContexts`. The `destroyed` guard on every public method is a good touch, and `cleanupDragControls` finally removes the window-level listeners the old comment admitted it leaked.

I checked the lifecycle worry and it holds up: `enterLoadingState` is one-way (guarded by `hasBegunWorldEntry`, recovers only via reload), so the start-screen preview is never re-created after release, which makes `this.canvas.remove()` safe. The in-game player card uses `hud.charPreview` with its own `document.createElement('canvas')`, a separate instance, so destroying the start-screen preview does not touch it.

Findings:

- should-fix `package.json:3` / `package-lock.json`: the `0.14.1 -> 0.16.0` bump is unrelated to this fix, jumps a minor (skips 0.15), and main is currently at 0.14.1. Versioning is a release-side decision. Drop the second commit ("Update version number") so the PR is just the teardown change.

- nit `src/render/characters/preview.ts:325`: no test for the new behavior. The double-destroy idempotency and the `releaseStartScreenPreview` null-out are testable against a mocked `CharacterPreview` without GL. Not a blocker given the rest is Three-bound, but worth a small `client_shell` case.

Nothing else. Renderer disposal order is right, `currentVisual.dispose()` releases the skeletons, and the release lands before the main renderer builds, which is the point on iOS.

FernandoX7 (dismissed):

PR #1055, opened 2026-06-30
fix(render): let nameplate clicks target the unit (healer party clicks)

jgyy:
## Problem

Clicking a party member's nameplate (the floating name above their head) often does not select them. For a healer this means missed targets and false heals: the click lands on the name, nothing gets targeted, and the heal goes to whoever was previously selected.

## Root cause

In-world nameplates are visual-only DOM (`pointer-events: none`), so a click on the name falls through to the 3D pick (`renderer.pick`). When the ray misses every clickable capsule, a forgiving "sloppy pick" fallback snaps to the nearest character within 26px. But that fallback measured distance only to the body MIDPOINT. The nameplate floats about a head-height plus 1.0 world units above the body, so a click on the name is well outside the 26px body radius and selects nobody (or, in a scrum, the nearest wrong body).

```mermaid
flowchart TD
    A[Click on party member's name] --> B{Ray hits a clickable capsule?}
    B -- yes --> T[Target that entity]
    B -- no --> C[Sloppy pick fallback]
    C --> D[BEFORE: distance to body MIDPOINT only]
    D --> F[name is far above body -> MISS / false heal]
    C --> G[AFTER: distance to COLUMN body..nameplate]
    G --> T
```

flow

## Fix

Anchor the sloppy pick to the whole vertical screen COLUMN, from the body midpoint up to the overhead nameplate (the same world offset the nameplate projection uses), instead of the body midpoint alone. A click anywhere from the name down to the body now registers on its owner.

The logic is extracted into a pure, unit-tested module `src/render/sloppy_pick.ts` (`distanceToSegmentPx` + `nearestSloppyPickId`); the renderer projects the two screen anchors per candidate and delegates. The radius stays 26px and the nearest column still wins, so overlapping units resolve exactly as before.

hit area

## Why it is safe

- Read-only pick path: `renderer.pick` returns an entity id; no sim / `IWorld` / wire / i18n change.
- Strictly more forgiving in the vertical band only; objects (doors, loot) still need a direct ray hit and the local player never competes for the click, both unchanged.
- Allocation only on click (not per frame), so no hot-path cost.

## Tests

- `tests/sloppy_pick.test.ts` (TDD red -> green): segment-distance math, a name-height click selecting its unit, body-midpoint clicks still working, out-of-radius rejection, nearest-of-overlapping, and the strict 26px boundary.
- `tsc`, `tests/architecture.test.ts` green; `npm run build` green; `npm run ci:changed` clean; pre-push QA floor passes.

### DIFF
--- a/src/render/renderer.ts
+++ b/src/render/renderer.ts
@@ -83,6 +83,7 @@ import { RenderBudgetGovernor, type RenderBudgetState } from './render_budget';
 import { downscaleDims } from './screenshot';
 import { drapeRingLocalY } from './selection_ring';
 import { buildClouds, buildSky, type SkyView } from './sky';
+import { nearestSloppyPickId, type SloppyPickCandidate } from './sloppy_pick';
 import { shouldRenderStealthGhost } from './stealth';
 import { buildFlaredConeFan, buildRingXZ, drapeConeWorld } from './target_cone_debug';
 import { buildTerrain, type TerrainView } from './terrain';
@@ -731,6 +732,7 @@ export class Renderer {
   private tmpV = new THREE.Vector3();
   private viewCandidates: ViewCandidate[] = [];
   private tmpV2 = new THREE.Vector3();
+  private tmpV3 = new THREE.Vector3();
   // Manual frustum cull for characters. Their skinned meshes keep
   // frustumCulled=false (a skinned mesh's bind-pose bounds don't follow the
   // animated pose, so Three's own cull pops visible rigs out), which means an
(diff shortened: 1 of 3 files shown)


FernandoX7 (commented):
Solid fix and the framing is right: nameplates are `pointer-events: none` so the click falls through to the 3D pick, and anchoring the sloppy pick to a vertical screen column instead of the body midpoint alone is the correct shape. The pure `sloppy_pick.ts` extraction (no Three, unit-tested, TDD red to green) is exactly how render math should land here. I confirmed the visibility guard (`isProjectedNameplateAnchorVisible`) is the same one the real nameplate and chat-bubble paths use, the 26px strict radius and nearest-wins tiebreak match the old behavior, the local player is still excluded, and objects still need a direct ray hit. Tests pass locally (10/10).

Two things to tighten:

**should-fix** `src/render/renderer.ts` (the pick path), and the PR's "allocation only on click (not per frame)" note. `pick()` is also the per-frame hover-cursor path: `updateHoverCursor()` runs every frame in `frame()` (`src/main.ts:2106`) and calls `renderer.pick(hoverX, hoverY)` whenever the cursor is over the canvas. When that hover ray misses every capsule (cursor over empty ground between units, which is most of the screen), the sloppy path now builds a fresh `candidates` array plus one object per visible entity every frame. The renderer already has the reuse pattern for exactly this: `viewCandidates` is a persistent field cleared with `.length = 0` (`src/render/renderer.ts:1835`). Reuse a persistent `SloppyPickCandidate[]` the same way, or keep `distanceToSegmentPx` and track the nearest inline with no array. Either keeps the tested math and drops the per-frame garbage.

**nit** `src/render/renderer.ts`, the `+ 1.0` top anchor and its comment "matches the nameplate projection offset". The real nameplate floats at `height * scale + NAMEPLATE_ANCHOR_LIFT`, and that constant is 0.8 (`src/render/nameplate_view.ts:41`), not 1.0. The 1.0 actually matches the chat-bubble path. It is harmless (the extra 0.2 plus the 26px radius just reaches a touch higher, which helps cover the name text that floats above the anchor), but the comment overstates the match. Either reference the shared constant or reword the comment so the next reader does not trust a match that is not there.

One thing worth keeping as is: the column is built regardless of the `showNameplates` toggle, so turning mob nameplates off does not shrink a player's pick reach. That is the fairness-correct choice.

jgyy:
Thanks, both tightened in the latest push (`tests/sloppy_pick.test.ts` still 10/10, tsc clean):

- **should-fix (per-frame allocation):** you are right that `pick()` doubles as the per-frame `updateHoverCursor` path, so the fallback was allocating a fresh `candidates` array (and an object per visible entity) on every cursor-over-empty-ground frame. Hoisted it to a persistent `this.sloppyCandidates` cleared with `.length = 0` and grown in place, exactly the `viewCandidates` reuse pattern you pointed at. The tested pure math (`distanceToSegmentPx` / `nearestSloppyPickId`) is unchanged, so the column behavior is identical.
- **nit (anchor comment):** corrected. The `+1.0` matches the chat-bubble offset, not the nameplate's `NAMEPLATE_ANCHOR_LIFT` (0.8); reworded both the block comment and the inline one to say so (and to note the extra reach over the 0.8 lift is what helps the column cover the floating name).

The flow / hit-area diagrams on the PR are unchanged: this push is allocation reuse plus comment accuracy, no change to the pick geometry or fallback logic they illustrate.

FernandoX7 (approved):

PR #1058, opened 2026-06-30
fix(ui): let the World Market scroll on mobile

jgyy:
## Problem

On a mobile touch viewport the **World Market** is unscrollable: the listing column is clipped to zero height, so a player cannot reach any listing past the filters.

### Root cause

`#market-window` shares a `max-height: calc(58vh - 20px)` cap with the vendor window. But unlike the vendor, the Market header carries a tab strip **plus** the type/rarity filter selects, which stack into a tall single column on mobile. That header consumed the whole 58vh window, leaving the `flex:1` listing body (`#market-body`) with **zero height**. `flex:1; min-height:0` is exactly what lets the body shrink below its content to enable inner scrolling, so once its siblings filled the parent it shrank to 0 and collapsed.

```mermaid
flowchart TD
    A["#market-window<br/>max-height: 58vh (~220px)"] --> B["panel title"]
    A --> C["tabs: Browse / Sell / Collect"]
    A --> D["filters (stack to 1 column on mobile):<br/>Type + Rarity + Search"]
    A --> E["#market-body (flex:1, min-height:0)"]
    B --> F["header eats the full 220px"]
    C --> F
    D --> F
    F --> G["#market-body collapses to 0px<br/>= no list, cannot scroll"]
    style G fill:#7a1f1f,color:#fff
```

## Fix

- Give `#market-window` the **full available height** on mobile (mirroring `#trade-window`) so the listing body keeps a usable height under the header.
- Floor `#market-body` with a `min-height` so it can never be squeezed flat, even when the subtype filter adds a third stacked row on a short landscape screen.

CSS-only change in `src/styles/hud.mobile.css` (the `hud-mobile` layer); the desktop layout and the filter-less vendor window are untouched.

## Before / after (landscape phone, 900x420)

| Before (list collapsed to 0px) | After (list present + scrollable) |
|---|---|
| before | after |

Scrolled, after fix:

after scrolled

### Follow-up: wider window so the close button clears the title

Per review, the desktop window went from 470px to 560px and the title is allowed
to shrink/ellipsize, so the close button no longer overlaps the name:

wider market window, desktop

## Verification

Added `scripts/market_mobile_scroll_shot.mjs`, which boots the offline game in a `body.mobile-touch` landscape viewport, floods the Merchant's market so the column overflows, opens the Market, and **measures `#market-body` clientHeight**:

- before: `bodyClientH: 0` -> `MARKET MOBILE SCROLL: STUCK`
- after: `bodyClientH: 96`, scroll advances -> `MARKET MOBILE SCROLL: SCROLLABLE`

Also green: `tests/css_corpus`, `tests/styles_extraction`, `tests/css_value_validity`, and `biome check` on both changed files.


FernandoX7 (commented):
Clean, well-scoped fix. The diagnosis matches the code: `#market-body` is `flex:1; min-height:0` (components.css:3206), so once the stacked tab strip plus filters filled the 58vh window the body shrank to zero and the listing collapsed. Bumping `#market-window` to `calc(100vh - 20px)` mirrors `#trade-window` (hud.mobile.css:921), keeps the body the inner scroller (window stays `overflow:hidden`), and the `min-height` floor is a sensible belt-and-suspenders for the short-landscape case. CSS-only in the `hud-mobile` layer, so the desktop layout and the vendor window are untouched, and the shared styles barrel means `/play` gets the fix too. The reproduction script measuring `#market-body` clientHeight is the right way to assert this.

A couple of small notes, neither blocking:

- nit, hud.mobile.css:885-888: there are now two `#market-window` max-height rules with different values. The first grouped block (`#vendor-window, #market-window`) still declares `58vh`, which is immediately overridden for the market by the new market-only block at `100vh`. It resolves correctly (later equal-specificity rule wins), and the duplicate predates this PR, but while you are here it would read cleaner to drop `#market-window` from that first group so there is no dead 58vh declaration for it.

- nit, hud.mobile.css:898-900: the `min-height: 96px` floor guarantees the body element is at least 96px, but on a very short landscape viewport the window's `overflow:hidden` can still clip the bottom of that floored body, so the visible scroll viewport can end up under 96px. That is still far better than the 0px collapse and the body scrolls internally, so this is just a note on the comment's "never be squeezed flat" wording, not a problem.

No Vitest is expected for a pure CSS change; the css_corpus / styles_extraction / css_value_validity guards plus the E2E shot are the right verification here.

FernandoX7:
Can you also make the view wider? The "x" close icon button is covering up the name

jgyy:
Thanks, all three addressed in the latest push (CSS-only; client_shell + css_corpus/styles_extraction/css_value_validity green, 89):

- **Wider window (your follow-up comment):** the desktop window is now 560px (was 470px), and the title (name + subtitle) is allowed to shrink and ellipsize, so the close "x" can never overlap the name however verbose the localized title runs. Desktop screenshot added to the PR body.
- **nit (dead 58vh decl):** dropped `#market-window` from the first grouped `calc(58vh - 20px)` block. It was immediately overridden by the market-only 100vh rule, so the 58vh value was dead for the market. Vendor keeps its 58vh cap, no behavior change.
- **nit (comment wording):** reworded the `#market-body` min-height comment. You are right that on a very short viewport the window's `overflow:hidden` can still clip below the 96px floor; the body scrolls internally, so it never collapses to the old zero height, but "never squeezed flat" overstated it.

Updated the client_shell pin to the new 560px width.

FernandoX7 (approved):

PR #1063, opened 2026-06-30
feat(ui): optional auto-attack on offensive ability use

Nervescraper:
## What

Adds a client-side **Auto-Attack on Ability Use** setting in Options > Interface. When on, using an offensive ability also engages your white-swing auto-attack, so you do not need a separate Attack press (the classic-MMO "start attack on ability" QoL).

**Default: on.** Because the setting key is new, no existing player's persisted `woc_settings` blob contains it, so the per-key default-fill in `Settings.load()` applies `true` to existing and fresh players alike. The only players who get it off are those who explicitly toggle it off after release.

## How

- New pure predicate `src/ui/attack_on_ability.ts` (`abilityStartsAutoAttack(effects)`): engage only when the ability **deals damage** and applies **no damage-breakable CC**. This is the load-bearing correctness decision:
  - Gouge deals damage AND incapacitates, so it is excluded (an auto-swing would immediately shatter the CC).
  - Heals, buffs, and pure CC (Polymorph, Sap, Hammer of Justice) never trigger it.
- Wired into the one shared `Hud.castSlot` path (covers both the keybind and the action-button click), gated on the setting.
- The sim's `startAutoAttack` already no-ops unless a valid hostile target is in melee range, so this can never engage on a bad target.
- Setting + Options toggle + an `applySetting` persist branch (no live subsystem to update; the HUD reads it at cast time).

## Seam / scope

Client-only: **no sim, wire-protocol, or `IWorld` change**. It reuses the first-class `startAutoAttack` command that already exists on `IWorld`. Sim stays setting-free.

## i18n

English label added to the `hudChrome.options.*` catalog. Per the M16 rule (wordy value), the five non-Latin overlays (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU) carry real fills in the same change; the resolved tables are regenerated.

## Tests

- `tests/attack_on_ability.test.ts` pins the predicate against real ability content: Sinister Strike / Mortal Strike / Fireball / Eviscerate engage; Battle Shout / Mark of the Wild / Polymorph / Sap / Hammer of Justice do not; Gouge does not despite dealing damage (order-independent).
- `tests/options_view.test.ts` updated for the new Interface-panel toggle.
- Green: `tsc`, the new predicate test, settings, options view/window, i18n completeness/coverage, and `architecture.test.ts`. Pre-push QA floor passed.


FernandoX7 (commented):
Clean, well-scoped change. The predicate is a pure module with a type-only sim import and a test that pins against real ability content (`tests/attack_on_ability.test.ts`); the gouge case (deals damage but a swing would break the CC) is handled and tested, and the `incapacitate` / `polymorph` exclusion set is complete against the current `AbilityEffect` union. The default-on claim holds: `Settings.load()` falls back to `BOOL_SETTINGS.def` for any key absent from a player's stored blob, so existing and fresh players both get it on. The `main.ts` persist-only branch matches the existing `filterProfanity` / `showWalletOnPlayerCard` pattern, and putting the engage in `Hud.castSlot` does cover both the keybind and the click.

One blocking issue.

**blocking, `src/ui/hud.ts:3531`: self/ground AOE cast with no hostile target pops a spurious "Invalid attack target." error toast, and the setting ships on by default.**
`abilityStartsAutoAttack` returns true for `aoeDamage` / `aoeRoot` / `groundAoE` (`src/ui/attack_on_ability.ts:17-19`), and the abilities whose only damaging effect is one of those are `requiresTarget: false`: Arcane Explosion, Frost Nova, Thunder Clap, Consecration, Cleave, bear Swipe (`src/sim/content/classes.ts`). Cast one of those with no hostile target selected (the normal way a mage/warrior/paladin AOEs a pack) and `castAbility` succeeds, but the unconditional `this.sim.startAutoAttack()` hits the no-target branch in `src/sim/combat/auto_attack.ts:56-58`, which calls `ctx.error(p.id, 'Invalid attack target.')`. That renders a red toast via `showError` (`src/ui/hud.ts:6089`), offline and online (the online path round-trips through the `attack` command in `server/game.ts`). So every targetless AOE cast flashes an error, out of the box. The inline comment at `hud.ts:3524` assumes `startAutoAttack` silently no-ops on a bad target, but it errors. Gate the new call on the player actually having a valid hostile target. Keep the guard at this call site: the explicit Attack button at `hud.ts:3511` is the one place that error feedback is wanted, so do not move it into `startAutoAttack`.

Nit.

**nit, `src/ui/attack_on_ability.ts:11-27`: the two effect-type sets are maintained by hand against the `AbilityEffect` union with no exhaustiveness tie.** A future damaging effect silently would not engage (safe), but a future break-on-damage CC type silently would not be excluded, so a damage+CC ability built on it would shatter its own CC. A comment pointing at the union, or a type-level check, would keep the `BREAK_ON_DAMAGE_CC` side honest.

Nervescraper:
Thanks for the thorough review, both spot-on. Pushed a fix (amended into the feature commit).

**Blocking (targetless AOE error):** Fixed. The auto-engage now gates on a new `hasAutoAttackTarget` predicate (live, hostile target), mirroring the sim's `startAutoAttack` precondition. The self/ground AOEs (Arcane Explosion, Frost Nova, Thunder Clap, Consecration, Cleave, bear Swipe) no longer trip the "Invalid attack target." toast, offline or online. As you suggested, the guard stays at the call site so the explicit Attack button keeps its error feedback, and `startAutoAttack` is untouched. Also corrected the stale comment that wrongly claimed `startAutoAttack` silently no-ops on a bad target.

The gate reads `target.hostile` + `dead`, which the server already mirrors onto the wire (`out.h = 1`) and is the HUD's existing notion of "enemy" (target-frame coloring, the raid-marker predicate), so it holds identically for `Sim` and `ClientWorld` with no seam or wire change.

**Nit (exhaustiveness tie):** Done. Replaced the two hand-maintained effect-type sets with a single exhaustive `Record<AbilityEffect['type'], 'damage' | 'breakCC' | 'other'>`. Adding a new effect to the union is now a compile error until it's classified, so a future break-on-damage CC effect cannot silently slip past the exclusion. Behavior is byte-identical to the old sets (verified by `tsc` and the existing tests).

Added tests for the four `hasAutoAttackTarget` cases plus an assertion that the self/ground AOEs report as attacks (documenting why the target gate is needed). Full suite green.

FernandoX7 (commented):
Re-reviewed at `f43d68d5`. The blocking issue is fixed, and cleanly.

- Targetless AOE no longer errors. I checked the dangerous direction (an entity with `hostile:true` where `isHostileTo` returns false, which would still toast): vision mobs spawn `hostile=false` (`encounters/nythraxis.ts:1151`) and every pet path sets `hostile=false`, so it is not reachable. Cast Arcane Explosion / Frost Nova / Thunder Clap with no hostile target and it now no-ops silently.
- The `Record<AbilityEffect['type'], AutoAttackClass>` rework gives the exhaustiveness tie the earlier nit asked for: a new `AbilityEffect` is a compile error here until classified, so the break-on-damage set can't silently rot.
- Wire parity holds: `hostile` round-trips (`server/game.ts:522` to `online.ts:1286`) and is in `blankEntity` (`online.ts:696`), so the helper reads the same flag offline and online. Tests cover the fix and pass.

One nit.

**nit, `src/ui/attack_on_ability.ts`: `hasAutoAttackTarget`'s comment says it "mirrors the sim's startAutoAttack precondition (`!t || t.dead || !isHostileTo`)", but it only reads the raw `hostile` flag, which diverges from `isHostileTo` for players.** `isHostileTo` returns true for an enemy player in an active duel or arena match (`sim.ts:4512`); players carry `hostile:false`, so the convenience won't engage there even though the explicit Attack button does. That is safe (no error, just no auto-engage), and replicating duel/arena state on the client is not worth it, so PvE-mob scope is a fine call. Consider softening the comment to say "a live hostile mob" rather than claiming it mirrors `isHostileTo`.

The PR shows conflicting, but the only true conflicts are the i18n regen trio (`i18n.resolved.generated/pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`): mechanical stale-base churn. Merge release/v0.18.0 and re-run the i18n build (`i18n:build` / `i18n:scan` / `i18n:hash --write`) and it clears.

Good to merge after that.

FernandoX7 (commented):
Re-reviewed at `22187e75`. Both earlier points are fixed and I re-verified the correctness argument.

The targetless-AOE error stays fixed: `castSlot` gates on `hasAutoAttackTarget` (`hud.ts:3531`). I chased the one residual divergence, a `hostile:true` entity where the sim's `isHostileTo` returns false, and it is not reachable: `vision_` mobs spawn `hostile=false` (`nythraxis.ts:1151`) and every owned mob is a player pet/charm set `hostile=false` (`pet_commands.ts`, `sim.ts:5535`). So for any `hostile:true` mob `isHostileTo` is true, and the client gate is a strict subset of the sim's error precondition. The earlier comment nit is addressed too.

Classification holds: `incapacitate` + `polymorph` are exactly the two player-ability effects that emit `breaksOnDamage` auras (`effect_dispatch.ts:373,398`), and the `Record<AbilityEffect['type'], ...>` keeps it compile-checked. Verified green in a clean worktree: `tsc` clean, `attack_on_ability` + `options_view` 30/30.

One nit, non-blocking:

**nit, `hud.ts:3518`:** auto-attack engages on the button press, not on a successful cast (fires even if the ability is on cooldown / out of resource). Matches the classic QoL and is harmless given the target gate, so fine to leave. Flagging only so it is a conscious choice.

Merge is clean now (no conflicts against `release/v0.18.0`). Good to merge.

Nervescraper:
Understood on the engage-on-button press, I think it's preferable as-is, but it could certainly be a setting at some point in the future. For now I think the QoL justifies merging this in and considering that as future work.

FernandoX7 (approved):

PR #1067, opened 2026-06-30
feat(sim): training dummy on the Highwatch hill with proper DPS metering

ryan-foo:
## Summary

A stationary **training dummy** on the hill above Highwatch (`-40, 648`) for testing damage rotations and reading the combat meters. This re-lands the old training-dummy work (originally cut from `release/v0.15.0`, PR #935) onto `release/v0.17.0`, moved to the third town and tuned for clean DPS measurement.

- `MobTemplate` gains **`dummy`** (a stationary, attackable-but-inert practice target) and **`respawnSeconds`** (a fixed respawn that also caps corpse decay). The `training_dummy` is **cap-level (20) with zero armor**, so the damage it takes is your clean, unmitigated rotation output; 999,999 HP, never moves/aggros/retaliates.
- `enterCombat` (sim.ts) skips aggro for dummies in both directions; `updateMob` (`mob/locomotion.ts`) keeps the dummy hostile/attackable while inert and, **5s after the last hit** (`DUMMY_RESET_SECONDS`), drops combat and heals to full; `handleDeath` (`combat/damage.ts`) honors the fixed `respawnSeconds` (10s).
- **DPS metering:** the dummy never holds aggro, so the combat-meter encounter closes on the normal inactivity window (`ENCOUNTER_END_SECONDS`, also 5s) once you stop. The player leaves combat, and the finished segment's DPS is retained in the meter history ("Last fight"). No meter change was needed; the dummy integrates cleanly.
- **Determinism:** a practice dummy is a fixed prop, so it spawns **without drawing any RNG** (the camp loop special-cases it) — adding it never perturbs the world's seed-stable spawns/rolls.

## Testing

`tests/training_dummy.test.ts` (inert, no-aggro / no-retaliation, combat-exit + heal-to-full, respawn) and a `meters.test.ts` case (measures + retains a dummy segment's DPS). Mob name localized (en + five non-Latin locales); parity goldens regenerated for the new entity. Full suite green (548 files / 5785 tests), biome + build clean.


jgyy (commented):
This is a clean, well-scoped feature. The actual change (once you filter out the stale-base drift, see below) is small and self-contained: `dummy` and `respawnSeconds` on `MobTemplate`, a fixed-spot camp entry, RNG-free spawn for `dummy` templates, aggro suppressed both directions in `enterCombat`, and a dedicated inert/heal-to-full branch in `updateMob`. The comments explain the "why" at each site (why no RNG draw, why respawn also caps corpse decay, why the reset window mirrors `ENCOUNTER_END_SECONDS`), which matches the meter-integration claim in the PR body: I checked `MeterData` and the dummy really does close out on the normal inactivity window since `aggroTargetId` never gets set, so "no meter change was needed" holds up.

Tests are decisive: inert/no-retaliate, heal-to-full after `DUMMY_RESET_SECONDS`, respawn with the fixed 10s timer, and a `meters.test.ts` case that pins the retained segment's DPS (`duration === 10`, `dmg === 1000`). i18n key (`training_dummy`) is present in `world_entity_i18n.ts`, English only as expected from a contributor.

**Note on the diff size:** `gh pr diff` and the raw `git diff` against `release/v0.18.0` show 86 files / 4201+/3871- because the branch is stale (it forked near #1204; `release/v0.18.0` is now past #1380, including the ghost-mode revert and several deletions like `server/github*.ts`, `dev_tier.ts`, `market_query.ts`). None of that is this PR's doing. Diffing against the actual merge base (`94f7a77a2`) isolates the real change to `src/sim/types.ts`, `src/sim/content/zone3.ts`, `src/sim/mob/locomotion.ts`, `src/sim/combat/damage.ts`, `src/sim/sim.ts`, `tests/training_dummy.test.ts`, `tests/meters.test.ts`, plus i18n/parity-golden regen. The base needs a rebase/merge before this can land; the mergeable state is currently `CONFLICTING` and the true conflicts are the standard generated trio (`i18n.resolved.generated/pending.ts`, `i18n.status.summary.json`, several `tests/parity/golden/*.json`), which is mechanical churn from the moved base, not a design problem. Worth flagging so it doesn't get reviewed against the wrong diff.

Findings:

- **should-fix**: branch is stale against `release/v0.18.0` (forked near #1204, base now past #1380) and currently shows `mergeable: CONFLICTING`. Needs a rebase/merge plus a regen of the generated i18n trio and parity goldens before this can land. Not a code defect in the PR itself, just needs a refresh pass.
- **nit**: `src/sim/mob/locomotion.ts` new `dummy` branch returns before the normal `mob.combatTimer` reset/idle bookkeeping later in `updateMob` runs for other templates. Confirmed intentional (the dummy skips all of that by design) and it is covered by the "never moves to attack" assertion, so this is just a note for the next reader, not a request for change.

Nice touch: spawning the dummy without an RNG draw so it can't perturb any other seed-stable spawn or loot roll, and pointing it at the third town instead of just re-landing the original cut PR verbatim.

```mermaid
flowchart LR
    subgraph Spawn
        A[Camp loop: ZONE3_CAMPS training_dummy] -->|dummy:true, skips RNG| B[createMob at fixed -40,648]
    end

    subgraph Combat
        B --> C[Player auto-attacks dummy]
        C -->|enterCombat: dummy flag blocks aggro both ways| D[handleDamage / handleDeath]
        D -->|hp -= dmg, no retaliation, no move| B
        D -->|hp <= 0| E[respawnTimer = 10s fixed, corpseTimer capped]
        E --> B
    end

    subgraph Meter
        C -->|SimEvent: damage| F[MeterData.onEvent]
        F -->|aggroTargetId stays null, no artificial hold-open| G[Encounter closes after ENCOUNTER_END_SECONDS]
        G --> H[Segment moved to history: Last fight DPS]
    end

    subgraph Reset
        B -->|updateMob: DUMMY_RESET_SECONDS after last hit| I[inCombat=false, hp=maxHp]
    end
```

jgyy:
finally merge conflicts GONE GONE GONE GONE no more i18n plz

PR #1068, opened 2026-06-30
feat(social): right-click a party member to make them leader

ryan-foo:
## Summary

Adds a **"Make Leader"** action to the player context menu, so the party leader can hand leadership to another member by **right-clicking** them (a party-frame row, a chat name, or a unit). The item appears right next to "Remove from Party" and is shown only to the current leader, for another member.

- **`party.ts`** gains `promoteLeader(targetPid)`: only the current leader may promote, the target must be a different member, and the whole group is told who the new leader is, reusing the **exact line already emitted** when leadership passes on a leader's departure (so no new sim string), and the existing "You are not the party leader." error.
- A new **`ppromote`** command is plumbed through the `IWorldParty` seam: the `world_api` command table + tag map, `online.ts` (ClientWorld send), the `sim.ts` delegate, and the `server/game.ts` dispatch.
- **`hud.ts`** `openContextMenu` renders the item (gated on leader + member + not-self) and dispatches `partyPromote`. New English label `hud.chat.context.makeLeader` plus the five non-Latin locale overlays (zh_CN / zh_TW / ja_JP / ko_KR / ru_RU); other locales re-localize at release.

## Testing

`tests/party_promote.test.ts` (leader promotes + group announcement, non-leader refused, non-member / already-leader no-op). Updated the `world_api` structural-parity pins (the `IWorldParty` facet, the 146-member `IWorld` set, the sorted member/method lists) and the command-schema counts for the new command. Full suite green (548 files / 5784 tests), biome + build clean.


Nervescraper:
Duplicated by https://github.com/levy-street/world-of-claudecraft/pull/1070, need to choose one.

FernandoX7:
Ended up reviewing the other one first. Thank you both.

PR #1070, opened 2026-06-30
feat(social): promote-to-leader handoff (master loot follows the new leader)

Nervescraper:
This PR contains two related party/master-loot changes.

### Motivation
Party/raid leadership previously changed **only** when the leader left or disconnected (auto-handed to the first remaining member). There was no way to deliberately assign leadership to a chosen member. Master loot already follows the *live* leader (`effectiveMasterLooter` resolves `looter: 0` to the current leader, and the HUD's leader-only control keys off `info.leader`), so the one missing piece was the manual handoff. This adds it, and the master-loot control + assignment prompt follow the new leader automatically.

### What changed
- **Sim core** (`src/sim/social/party.ts`): `PartyMachine.partyPromote(targetPid, pid?)` — a leader-only handoff that sets `party.leader` without changing the roster and announces it through the existing "is now the party leader" log line (already localized via `hud.logs.partyLeader`). Thin `Sim` delegate alongside `partyKick`.
- **IWorld + wire**: `IWorldParty.partyPromote`; `ppromote` wire token added to `COMMAND_NAMES` + `COMMAND_FACETS`; `ClientWorld.partyPromote`; `case 'ppromote'` in the server dispatch (field-validated, resolves in the Sim).
- **UI** (`src/ui/hud.ts`): a "Promote to Leader" item on the party-member right-click menu, gated `isLeader && isMember && pid !== self`.
- **i18n**: `hudChrome.party.promoteLeader` (English) plus the five non-Latin fills required by M16 (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU). Regenerated resolved tables included.

Master loot following is automatic, no extra wiring: after a promote the new leader's HUD shows the control and the old leader's disappears (existing `info.leader` gate), and a threshold drop's assignment prompt routes to the new leader (existing `effectiveMasterLooter`).

### Motivation
The master looter shared the 60s need/greed roll timeout (`LOOT_ROLL_TIMEOUT`). That is short for a deliberate by-hand loot assignment (deciding who gets the drop), where the looter may be inspecting candidates or coordinating in chat. A wider window matches how master loot is actually used.

### What changed
- **Sim** (`src/sim/loot/loot_roll.ts`): added a dedicated `MASTER_LOOT_TIMEOUT = 300` (5 minutes) used by the curate phase (`startMasterLootRoll`), separate from `LOOT_ROLL_TIMEOUT` (still 60s for need/greed). On timeout it still falls back to a fresh 60s need/greed roll for all candidates via `convertMasterRollToNeedGreed`, and the corpse timer is extended to cover the longer window.
- **HUD** (`src/ui/hud.ts`): the master-loot prompt's countdown-bar `durationMs` is bumped to `300_000` so the visual timer spans the full window (the `masterLoot` event already carries `expiresAt`, which drives the authoritative timeout).

## Tests
- `tests/party_machine.test.ts`: leader can promote (roster unchanged + announce); non-leader rejected; promoting a non-member/self is a no-op; master loot pinned to the leader (looter 0) follows.
- `tests/loot_master_sim.test.ts`: after a promote, a threshold drop's master-loot assignment prompt reaches the **new** leader; the master looter's window is now+300 and a still-open roll has **not** converted at 4 minutes; the timeout-conversion test runs past the 5-minute mark.
- Count/parity guards updated: `world_api_parity` (146/110), `command_schema` (103/110).

Full suite green for these changes (`tsc`, architecture, S3 i18n, M16 completeness, the two i18n freshness gates, `ci:changed`).


Nervescraper:
Duplicated by https://github.com/levy-street/world-of-claudecraft/pull/1068, need to choose one.

FernandoX7 (commented):
Clean, well-scoped change. The promote handoff mirrors `partyKick` across every seam (`COMMAND_NAMES`/`COMMAND_FACETS`, `IWorldParty.partyPromote`, `ClientWorld` send, the `Sim` delegate, the server `case 'ppromote'`), and server authority is intact: the dispatch passes the connection `pid`, only `msg.id` comes from the client, and the Sim re-checks leader plus membership before changing anything. The loot-window change stays deterministic (`ctx.time` is tick-derived, no wall-clock or new `Rng` draw), and `party.members` holds every raid member across both groups, so the handoff and the master-loot follow work in raids too, same as kick. The HUD `durationMs` bump is needed, not just cosmetic: `showMasterLoot` is the only place that seeds `activeMasterRolls`, so without it `updateLootRollTimers` would retire the panel at 60s while the server still grants 300s. Tests cover the real paths.

One blocking issue.

**blocking** `tests/world_api_parity.test.ts:327` (and `:329`, `:995`, `:996`): the count pins were not bumped. Adding `partyPromote` to `IWORLD_MEMBERS` takes it to 147 members / 111 methods, but the pins still read `toBe(146)` / `toBe(110)`, so the file fails:

```
tests/world_api_parity.test.ts > pins total / data / method counts
  expect(IWORLD_MEMBERS.length).toBe(146)   // Received: 147
tests/world_api_parity.test.ts > the union of the 20 facets equals the pinned 146-member set
  union ... toBe(146)                        // Received: 147
```

Bump line 327 to 147, 329 to 111, 995 and 996 to 147, and the doc comments that cite the counts (lines 3, 12, 67, 68, 339, 690, 993). The summary's "world_api_parity (146/110)" should read 147/111. The `command_schema` side was updated correctly (102 to 103, 109 to 110), so this reads as a missed pair rather than a misunderstanding.

nit: because that changed file fails, the "full suite green / ci:changed" line in the description does not hold. A quick `npx vitest run tests/world_api_parity.test.ts` before merge would have caught it.

Nervescraper:
Thanks for the careful review. Fixed the blocking issue: bumped the count pins in tests/world_api_parity.test.ts (147 members / 111 methods) at lines 327 and 329, the union assertions at 995 and 996, and the doc comments that cite the counts (3, 12, 67, 68, 339, 690, 993). The amend lives in the social commit alongside the partyPromote member add, so the commit's tests-pass claim now holds.

Note this branch was also rebased onto release/v0.18.0 (was v0.17.0), and the PR base is retargeted to match.

Verified locally:
```
npx vitest run tests/world_api_parity.test.ts tests/command_schema.test.ts tests/loot_master_sim.test.ts
Test Files  3 passed (3)
     Tests  180 passed (180)
```

FernandoX7 (approved):

PR #1044, opened 2026-06-29
feat(discord): two-way Discord integration (login, rewards, bot, relay, in-game flair)

maxpolaczuk:
## What

A full two-way Discord <-> World of ClaudeCraft integration, re-ported onto the latest base.

**Account + rewards**
- "Continue with Discord" first-class OAuth login/link (full-page redirect; lands you straight in online play)
- Reward economy: points from in-game **playtime** (active-only) + **Discord activity** (message/voice daily-active), driving 8 status tiers
- Unlinked "Link your Discord" CTA banner with live `online / total members` counts

**Bot (`bot/`, `npm run bot`)** — standalone gateway process, zero new deps
- Auto-creates + syncs `WoC <Tier>` status-tier roles
- Pushes presence (online/voice/member-total) + member metadata (nickname, server-join date, top staff role) to the game
- Grants daily-active reward points; posts the relay cards

**In-game "!" community relay** (`!lfg !wts !wtb !recruit !event !help`)
- Broadcasts in-world + cross-posts a rich embed to Discord with a deep-link **Respond** button that lands the responder in game and auto-whispers
- Chat autocomplete dropdown for the commands

**Linked-Discord flair in game**
- Nameplate PFP + staff-role **name color + tag** (Levy St / Dev / Mod / Artist)
- Target frame line + inspect card: nickname, rank, "member since", role
- U-panel widget (toggle **U**): link status, tier ladder, presence, **Visit Discord**

## How it's wired
- **Server:** `discord.ts` / `discord_oauth.ts` / `discord_db.ts` / `discord_relay.ts`; routes in `main.ts`; secret-gated bot endpoints in `internal.ts`; rate-limit + db helpers. Reward/flair follow the wallet/holderTier precedent (external account state, **not** `IWorld`).
- **Client:** `discord_status` / `discord_widget` / `discord_tier` / `discord_deeplink`; nameplate / inspect / target wiring; per-entity wire fields `dt/dav/dnm/dj/dr` (identityFields -> applyWire, delta-safe).
- **Shared pure cores:** `src/sim/discord_tier.ts` / `discord_roles.ts` / `discord_relay.ts`.

## Config (server-side env; feature OFF until set)
[...]

PR #1075, opened 2026-06-30
feat(discord): account choice on first login and safe unlink

Rubsey:
## What and why

Two issues with the Discord integration, plus a serious pre-existing bug surfaced while fixing them.

### Issue 1: first-time Discord sign-in auto-created an account
The OAuth callback used to silently provision a brand-new account on first login, with no way to use an account you already have. Now it stops and asks.

- The callback no longer provisions. When a verified Discord identity has no linked account, it parks the identity in a new single-use, short-lived `discord_pending_logins` row and hands the SPA a one-time `linkToken`.
- The client shows a chooser (`#discord-choice-panel`):
  - **Create a new account** -> `POST /api/auth/discord/login/new` -> provisions a password-less account, links it, drops into character creation.
  - **Link an existing account** -> `POST /api/auth/discord/login/link` (username + password, plus a 2FA code if that account has it) -> verifies password + moderation lock + 2FA exactly like `/api/login`, links, lands in that account's normal character flow.
- The pending token is consumed only on the final commit, so a wrong password or a 2FA challenge leaves it reusable for the retry.

### Issue 2: unlinking a Discord-created account stranded it forever
A Discord-provisioned account only ever had a random placeholder password, so unlinking removed its only way in.

- Accounts now track `password_set` (additive `accounts` column, default `TRUE`; `FALSE` for Discord-provisioned accounts).
- Unlinking a password-less account now requires setting a password first (the username is shown read-only, via `#discord-keep-modal`). The password is set **before** the link is removed, so even a failure mid-flow cannot strand the account.
- Accounts that already have a real password unlink unchanged. `GET /api/discord` now returns `passwordSet` so the client only prompts when needed.

### Pre-existing bug fixed: Discord schema was never applied at boot
`DISCORD_SCHEMA` was defined in `server/discord_db.ts` but never run by `ensureSchema()` (which only applied `SCHEMA`, `SOCIAL_SCHEMA`, `OAUTH_SCHEMA`). So the entire Discord schema (links, oauth states, reward tables, and the new pending-logins table) was never created at boot. It went unnoticed because the feature is off by default and unit tests mock `pg`. Without this, the integration and these fixes would throw `relation does not exist` against a real database. Wired it in (mirroring the other schema modules) and added a guard test.

## Review notes
This change was run through the repo's domain reviewers (migration-safety, privacy-security, cross-platform-sync, qa-checklist) with an adversarial verification pass. Confirmed findings were all addressed:
- Failed-2FA on the link path now feeds the per-account brute-force lockout (`recordAuthFailure`), matching `/api/login`.
- The chooser localizes server error codes that lacked a match arm instead of showing raw text.

The new login/link endpoint is unauthenticated by design (like the OAuth callback), but is gated by a single-use, OAuth-minted pending token, the per-account auth throttle, and the dedicated Discord rate-limit bucket.

## i18n
New `hudChrome.discord.choice.*` and `.keep.*` English keys with the five non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) per the M16 rule. Resolved tables regenerated; freshness gate green.

## Migration safety
Both DDL additions are additive, idempotent, and boot-safe (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS`, `CREATE TABLE IF NOT EXISTS`). All new SQL is parameterized. Accounts provisioned by Discord before this change default to `password_set = TRUE`; in production there are none (the integration is new in this release and off by default), so no backfill is needed.

## Test plan
- [x] `tsc` clean, Biome clean on changed files
- [x] Unit tests green: Discord DB (pending-login create/peek/consume), Discord server handlers (chooser callback, `login/new`, `login/link` incl. the 2FA throttle-parity test, unlink password guard, status `passwordSet`), and the new `schema_wiring` guard
- [ ] Manual (needs `DISCORD_CLIENT_ID`/`SECRET` + a real DB): first Discord login shows the chooser; create-new lands in character creation; link-existing with correct credentials lands in character select; wrong password / 2FA is retryable
- [ ] Manual: unlinking a Discord-only account prompts for a password, then unlinks, and the account can log in afterward with username + password
- [ ] Manual: unlinking an account that already has a password works with no prompt


FernandoX7 (commented):
Strong PR. It fixes the `DISCORD_SCHEMA`-never-applied blocker from #1044 the right way (wired into `ensureSchema()` after `OAUTH_SCHEMA` with correct FK ordering, plus a guard test), and the first-login chooser is well thought through. The single-use pending token is an atomic `DELETE ... RETURNING`, the link path peeks before it consumes so a wrong password or a 2FA challenge stays retryable, and linking an existing account still requires that account's own password + 2FA, so there is no takeover. The unlink flow sets the password before removing the link, so a mid-flow failure cannot strand the account, and the route stays bearer-scoped (no IDOR). Throttle and 2FA parity with `/api/login` is exact: the failure counter is fed on both a bad password and a bad second factor, and the messages stay enumeration-safe. DDL is additive and idempotent, the new SQL is parameterized, and the behavior is tested.

Two things before merge.

## Blocking

- `src/main.ts:6678`: `wireDiscordChoice()` calls `$('#btn-discord-create').addEventListener(...)` with no null guard, and `$` is `querySelector` (returns null, it does not throw). play.html loads the same `/src/main.ts` (play.html:890) but does not carry `#discord-choice-panel` / `#discord-keep-modal`, and `DISCORD_BUILD_ENABLED` is true for web builds by default. So on `/play` this throws inside `wireStartScreens()` at the call site (`src/main.ts:6747`), which runs before `api.restoreSession()` in the same function. The throw aborts session restore and the rest of the start-screen wiring for every `/play` visitor, not just Discord users. The sibling wirings (`wireDiscordCtaBanner`, `wireDiscordKeepModal`) are null-guarded; this one needs the same, and the two new panels need mirroring into play.html for the chooser to work there at all.

## Should-fix

- The two new endpoints skip the IP-block and turnstile gates that `/api/register` and `/api/login` enforce. `handleDiscordLoginNew` (`server/discord.ts:340`) provisions an account with no `game.isIpBlocked` / `passesTurnstile` (cf. `server/main.ts:624` and the register turnstile check), and `handleDiscordLoginLink` (`server/discord.ts:402`) is a username + password + 2FA login path with no `isIpBlocked` / turnstile / web-origin check (cf. `/api/login` at `server/main.ts:709`). So a banned IP can mint fresh accounts and sessions through Discord. This is the same gap I noted on the integration in #1044; moving provisioning into the new endpoint carried it forward, and the link endpoint adds a second instance. `discordRateLimited` plus the per-account `authThrottled` bound it, but adding at least the `isIpBlocked` check to both endpoints would close it.

## Nits

- `server/discord.ts:369` provision-then-link race: if `provisionDiscordAccount` succeeds but `linkDiscordToAccount` then loses the TOCTOU race, the just-created account is left orphaned (password-less, unlinked, unreachable). It correctly falls back to logging into the real owner, so there is no wrong link, just a dead row.
- `server/discord.ts:635`: the unlink 400 returns `needsPassword` + `username`, but the client reads neither (it switches on `status === 400` and pre-fills the username locally). Dead fields on the wire.
- `pruneDiscordPendingLogins` runs on the 24h prune timer while these rows live about 15 minutes, so expired rows can linger up to a day. Harmless, since every read filters `expires_at > now()`.

Note: GitHub shows this as conflicting, but it merges clean against the current `release/v0.17.0` locally, so that flag looks stale.

Rubsey:
Thanks for the thorough review. Pushed a follow-up (`be3ad643`) that addresses everything, plus fixes a related modal bug that surfaced in manual testing.

**Blocking (/play crash).** The throw was broader than the `wireDiscordChoice` call site: this PR added `#discord-choice-panel` to `show()`'s `panels[]`, and `show()` dereferences every panel id with no null guard, so on `/play` (which lacks that element) *every* start-screen panel switch threw, not just the Discord wiring. Fixed by:
- making `show()` null-safe (`document.querySelector(id)?...`),
- null-guarding `wireDiscordChoice()` (early-return when the panel is absent, mirroring the sibling wirings),
- gating the parked-choice boot branch on the panel's presence.

I did not mirror the chooser into `play.html`: the OAuth callback always redirects to `/`, and `play.html` carries no "Continue with Discord" entry point, so the chooser is genuinely an index.html-only flow and mirroring it would be dead UI. The defensive guards make `/play` robust regardless of which panels an entry ships.

**Should-fix (IP block).** Both `/api/auth/discord/login/new` and `/login/link` now take an `isIpBlocked` predicate and return `429 { error: 'rate limited' }` *before* consuming the pending token or verifying the password, wired at the routes as `(ip) => game.isIpBlocked(ip)`. Added a blocked-IP test for each (asserts no token consume / no account insert / no link write).

**Nit (dead wire fields).** The unlink 400 is now just `{ error: 'password_required' }`.

**Nits left as-is, both as you flagged them benign:**
- provision-then-link orphan row: a rare TOCTOU race leaves a password-less, unlinked, random-username row that can never be logged into; the fallback to the real owner is already correct, and there is no account-delete path to reuse. Not worth a delete path for an invisible dead row.
- prune cadence: harmless since every read filters `expires_at > now()`.

**Separately, an in-game bug I hit while testing:** the "set a password before unlink" modal (`#discord-keep-modal`) was not appearing. It was nested inside `#start-screen`, which is `display:none` in-game and a `z-index:100` stacking context, so the modal was hidden during play and trapped below the in-game Discord panel. Hoisted it to top level (sibling of `#discord-window`) and added a `client_shell` guard asserting it stays out of `#start-screen`.

`tsc` and Biome clean on changed files; affected suites green (109 tests).

PR #1076, opened 2026-06-30
feat(admin): enrich online player positions

madmatah:
## Summary

Enrich online player positions on admin dasboard overview.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Manual steps:
  -  Go to the admin dashboard / overview
  - Put your mouse pointer on a position

## Screenshots / recordings

Screenshot from 2026-06-30 13-32-25

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Nice change. The location logic is pulled into a pure `src/admin/location.ts` with a real Vitest, the server-side resolver correctly derives the origin zone from the instance `doorPos.z` instead of the player's instanced position (so a dungeon/delve player still reports their overworld zone), and the POI labels in the dict line up exactly with the zone definitions. `liveSessions()` stays admin-only (server/admin.ts:411), coords and distance go through `fmtNumber`, and the new strings are English `t()` keys in the admin catalog.

A couple of small things, both nits:

- nit `src/admin/components/LocationCell.svelte:21` the tooltip is hover-only (`:hover`) and the wrapper is a roleless `<span>` with no `tabindex`, so keyboard-only operators can't open it. The `aria-label` carries the full details, but `aria-label` on a generic span with no role isn't reliably announced by screen readers either. For an operator tool this is minor, but making the cell focusable (or a `<button>`) would close both gaps. The checklist left "Accessible" unchecked, so flagging it.
- nit `src/admin/location.ts:51` `location.slot` is interpolated raw (`{ value: location.instanceSlot }`) while coords and distance go through `fmtNumber`. Slot is an index rather than a quantity, so it is cosmetic, just noting the inconsistency.

Verified locally: `tests/admin_location.test.ts` passes (2/2), and the dungeon/delve/overworld branches in `liveLocationFor` match the helper signatures (`instanceInfoAt`, `delveRunForPlayer`, `DUNGEONS`/`DELVES.doorPos`). No wire-parity concern since `location` rides the admin REST payload, not the game wire.

FernandoX7 (approved):

PR #1086, opened 2026-06-30
feat(dev-badges): developer badges for open-source contributors

awidearray:
## Summary

Adds cosmetic **developer badges** for contributors who have landed commits into
the open-source repo, mirroring the existing `$WOC` holder-tier and Discord
status-tier systems exactly (a third instance of the same pattern: pure sim
core + UI presentation twin + identity-field wire plumbing + server refresh
loop). Purely cosmetic: the sim never reads any of it, same invariant as the
other two badge systems.

- **Identity:** verified GitHub OAuth link (mirrors the existing Discord link
  flow: `server/github_oauth.ts` + `github_db.ts` + `github.ts`, a "Link
  GitHub" control beside the wallet-link row on character select).
- **Commit source:** GitHub's public `/contributors` stats for the repo,
  cached server-side (mirrors the existing GitHub Releases proxy), with a
  5-minute backoff after a failed fetch so a down/rate-limited API isn't
  hammered. `commitsForLogin` resolves a linked account's landed-commit count;
  `topContributors` ranks everyone for the leaderboard.
- **Five rungs:** Tinkerer (1+) -> Artificer (10+) -> Runesmith (50+) ->
  Architect (150+) -> Worldwright (500+), with a procedural "commit graph" SVG
  badge per rung (distinct from the holder coins and Discord numerals).
- **Surfaces:**
  - **Nameplate:** badge glyph next to other players' names; Architect+
    contributors additionally get a glowing name outline, composed on top of
    (not replacing) any existing Discord-staff name color.
  - **Inspect screen:** badge + tier name + "N commits landed" + `@login`.
  - **Player card:** badge drawn into the shareable card PNG.
  - **High-score window:** a third "Developers" tab ranking contributors by
    landed commits, with the badge glyph, alongside the existing
    Players/Guilds boards.
- **Admin dashboard:** the contributor-stats cache and the link flow report
  through the same provider-usage telemetry as their closest siblings
  (`woc_balance.ts` / the Releases proxy / `wallet.ts`), so an operator can see
  fetch/hit/miss/failure counts and link request/failure rates.
- **i18n:** English catalog additions in `hudChrome.devBadge.*` and
  `hudChrome.leaderboard.*`, with the M16 non-Latin (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU)
  fills included in the same change.

This PR also includes a self-review pass: an independent adversarial audit of
the branch (hunting specifically for stubs, fake data, swallowed errors,
un-awaited async, and untested code paths) found 4 real gaps, all fixed in the
last commit — see that commit message for the itemized list. Net effect: the
GitHub link UI now surfaces failures instead of silently no-op'ing, one piece
of dead code was removed, and three new test files close what had been the
only completely untested slice of the feature (the GitHub OAuth HTTP/DB layer
and the `refreshDevBadge` resolution path).

## Related issues

N/A

## Type of change

## How was this tested?

- Commands: `npx tsc -p tsconfig.json --noEmit`, `npm test` (578 files, 6111
  tests, 11 pre-existing skips, all green on a clean tree),
  `npx @biomejs/biome ci --changed`, `npm run build`, `npm run build:server`.
- Real execution, not mocked logic: `tests/github_db.test.ts` and
  `tests/github_server.test.ts` drive the actual `server/github_db.ts` and
  `server/github.ts` functions against a hand-rolled fake `pg` pool (routing
  real SQL strings) and a mocked GitHub fetch, exactly like the existing
  `discord_db.test.ts`/`discord_server.test.ts` precedent. `tests/dev_badge_refresh.test.ts`
  drives `GameServer.refreshDevBadge` end to end: a real DB row resolves
  through a real (mocked-network) contributor-stats cache lookup to the
  stamped entity fields and the broadcast wire snapshot.
- Manual steps: traced the full OAuth link/unlink/relink flow end to end in
  source against the existing Discord flow it mirrors (no live GitHub OAuth
  app credentials available in this environment to drive it through a real
  browser popup).

## Screenshots / recordings

N/A (no live server/browser available in this environment to capture from; the
badge rendering reuses the exact same SVG-data-URL + DOM patterns as the
existing holder-tier and Discord-tier badges, which are already screenshotted
elsewhere in the codebase).

---

### Quality

- [x] **Tests pass.** `npm test` is green (578/578 files, 6111/6111 non-skipped
      tests) on a clean tree.
- [x] **Builds succeed.** `npm run build` and `npm run build:server` both pass.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Not verified live in this environment
      (no browser/device available); the new markup reuses existing
      `cs-wallet`-family CSS classes and the existing `.lb-row` grid pattern,
      both of which are already mobile-verified.
- [ ] **Accessible.** The leaderboard tab reuses the existing WAI-ARIA tablist
      wiring verbatim (roving tabindex, aria-controls, labelled tabpanel);
      not independently re-verified live.

### Localization (i18n)

      locale.** English source added first; the wordy (M16) strings also got
      their five non-Latin fills in this change. The remaining locales
      English-fill per the project's standard contributor workflow (the
      maintainer batch-fills the rest at release).
- [x] N/A for the sim/server re-localization checkbox: no new sim/server
      player-facing emit text was added (the dev-badge data rides identity
      fields exactly like `holderTier`, not chat/log text).

### Hygiene

- [x] No secrets, credentials, or `.env` committed. New env vars
      (`GITHUB_OAUTH_CLIENT_ID`/`SECRET`) are documented in `.env.example` only.
- [x] No generated files hand-edited (`i18n.resolved.generated/*` regenerated
      via `npm run i18n:gen`, `i18n.resolved.sha256` via `npm run i18n:hash`).


EnriqueGF (commented):
Reviewed against the repo invariants at the PR head: ran the targeted suites (300 pass across 12 files), `tsc --noEmit` clean, plus a sim-determinism, wire-parity, migration-safety and security/privacy pass. Also rendered the badges in-game from the PR's own `src/ui/dev_tier.ts`. Net: solid work, close to mergeable. One should-fix and a few nits.

## What it gets right

- **Sim stays pure and cosmetic.** `src/sim/dev_tier.ts` is a host-agnostic threshold ladder with zero imports, and there are zero reads or writes of `devTier/devCommits/githubLogin` anywhere in `src/sim/`, so it cannot affect gameplay. Architecture and parity guard tests are green.
- **Wire parity is intact.** `dvt/dvc/dgl` encode (`server/game.ts:504`) and decode (`src/net/online.ts:1233`) round-trip both ways with no key collision (notably `dgl` is not `dgn`), and it correctly follows the `holderTier`/`discordTier` precedent (decode-always, not in `blankEntity`). `refreshDevBadge` faithfully mirrors `refreshHolderTier`, server-authoritative (only the resolved index ships).
- **Security is sound.** CSPRNG single-use TTL-bounded OAuth `state`, account scoped from the bearer token with a double-gated `WHERE account_id`, secrets read from env only and never returned to the client or logged, `postMessage` origin-checked on both ends, and no SSRF (fixed `GITHUB_REPO`).
- **Schema is safe.** `github_links` and `github_oauth_states` are additive and idempotent (`IF NOT EXISTS`), parameterized, and applied under the existing advisory lock. No persisted-state shape change.
- **Tests are real.** Fake `pg` pools routing real SQL, a mocked GitHub fetch, the `refreshDevBadge` path driven end to end, and the N to 0 unlink transition are all covered.

## Findings

**1. should-fix: the "Link GitHub" CTA is missing from `play.html`.**
The new `#cs-github-group` was added to `index.html:783` only. `play.html` loads the same `/src/main.ts` and carries the full wallet-link character-select group (`cs-wallet-group` at `play.html:590`) but has no `cs-github-group` / `#btn-github`. So on `/play` a player can link a wallet but has no way to link GitHub. It does not crash (`wireGithubLink()` uses optional chaining and `refreshGithubLinkStatus()` early-returns when the group is absent), and badges still render on `/play` from the wire fields, only the link entry point is missing. Mirror the markup into `play.html` next to the wallet group.

**2. nit (defense in depth): `server/github_contributors.ts:124`** follows GitHub's `Link: rel="next"` URL with the `Authorization: Bearer` header re-attached, without pinning the host to `api.github.com`. The source is trusted HTTPS so the practical risk is negligible, but a host allowlist on the followed URL is cheap.

**3. nit: `GET /api/auth/github/callback` (`server/main.ts:1403`)** has no rate-limit bucket while the other three routes do. It matches the existing Discord-callback convention and the costly fetch is gated behind a single-use 256-bit state, so acceptable; a per-IP cap closes the last bit.

**4. nit (optional): `GET /api/github`** uses `bearerActiveAccount` (which rejects read-scoped OAuth tokens) rather than `bearerReadAccount`. Over-restrictive in the safe direction, worth a glance only if a read-scoped token should ever see link status.

No i18n issues: English lives in the `hudChrome` catalog and renders through `t()`, and the five non-Latin M16 fills are real translations with nothing leaked.

## In-game capture

Rendered from the PR's real `src/ui/dev_tier.ts` (badge art, ring and glow colors, tier names, thresholds, and the significant-contributor name outline) composited with the PR's actual nameplate, leaderboard and inspect CSS, served through Vite at the PR head. A live OAuth-backed capture was not possible in this environment (no GitHub OAuth app credentials or contributor data, the same limitation noted in the PR), so this drives the real rendering code with representative data.

Developer badges rendered in-game

EnriqueGF:
Relaying a suggestion from Fernando:

> Can I suggest adding a setting to be able to toggle that on/off?

A client-side toggle to show or hide the developer badges (nameplate glyph and outline, inspect block, player card, and the Developers leaderboard tab) would be a nice addition. Since the badges are purely cosmetic and read only from broadcast identity fields, a per-client display toggle stays gameplay-neutral, so it does not run into the graphics-settings-fairness rule (it hides cosmetic flair, not actionable information). It would fit alongside the existing HUD/social display settings rather than needing any server or sim change.

FernandoX7 (commented):
This is a clean, faithful third instance of the holder-tier / Discord-tier pattern, and it reads that way throughout: a pure ladder in `src/sim/dev_tier.ts` (zero imports, never read by the sim), the three wire fields threaded symmetrically through `identityFields`/`applyWire`, the badge surfaces reusing the existing painter and CSS families, and a genuinely thorough test set (wire codec, every DB branch, every HTTP handler, the refresh path, the view-core, the contributors cache + telemetry). I verified the load-bearing claims against the code rather than the description:

- Determinism and purity: `dev_tier.ts` is pure, no sim path reads `devTier`/`devCommits`/`githubLogin`, and the leaderboard sort is a total order with a stable login tiebreak (`b.commits - a.commits || a.login.localeCompare(b.login)`). architecture, parity, and tsc are green.
- Wire parity: encode and decode are symmetric, the N to 0 unlink path drops `dvt`/`dvc`/`dgl` and decodes back to no badge, `blankEntity` correctly omits the optional flair fields (matching the holder/discord precedent), and the IWorld pin bump (146 to 147, methods 110 to 111) matches the real contract (`world_api_parity` green).
- Server and auth: all four routes resolve the account from the bearer token (the callback from the consumed single-use `state` row), SQL is parameterized, secrets are env-only, the bounce page escapes its inlined JSON and constrains `postMessage` to `location.origin`, the routes share a dedicated rate-limit bucket, and `GITHUB_SCHEMA` is actually applied in `ensureSchema` (not the #1044 define-but-never-wire trap). The DDL is additive and idempotent.

Findings:

**should-fix** `index.html:783` adds the Link GitHub control (`#cs-github-group`) to character select, but `play.html` does not get it. `play.html` is a separate build entry carrying the same character-select chrome (it has the identical `cs-wallet` wallet group at `play.html:590`), and both entries load `/src/main.ts`, which wires the control by element id (`btn-github`, `cs-github-group`). On `/play` that wiring no-ops, so the GitHub link CTA never appears even when the feature is configured. The wallet group it mirrors is present in both files, which points to an oversight. Mirror the block into `play.html` after the `wallet-hidden-note` at `play.html:619` (it reuses shared `cs-wallet` classes, so no extra CSS is needed). The badges themselves still render on `/play` since they ride the wire; this is only the link entry point.

**nit** `server/provider_usage.ts:28` registers a `github.link.failure` metric (admin label "GitHub link failures"), but nothing emits it. `github.link.request` (`github.ts:68`) and `github.link.rate_limited` (`main.ts:1398`) are wired, the failure counter is not, so the callback error branches (`already_linked`/`github_error`/`server_error`/`expired`) never increment it and the dashboard row sits at 0. Either record it in `handleGitHubCallback`'s failure paths or drop the key.

**nit** (optional) `server/github_contributors.ts` follows the `rel="next"` URL from GitHub's Link header with `Authorization: Bearer ${GITHUB_TOKEN}` attached. The base URL is server config and the trust root is api.github.com, so this is not client-exploitable, but pinning the followed host to `api.github.com` before sending the token is cheap defense-in-depth.

Nothing else of note. The `console.log` arrow in `refreshDevBadge` matches the existing `refreshHolderTier` log verbatim, so it is consistent with the file.

FernandoX7:
One other thing: It's hard to distinguish between badge names. Could they instead be different colors?

awidearray:
Pushed a follow-up commit addressing both reviews and Fernando's relayed suggestion:

**should-fix (both reviewers): `play.html` missing the Link GitHub control**
Mirrored `#cs-github-group` into `play.html` next to the wallet group, using the exact element ids `src/main.ts`'s `wireGithubLink()`/`refreshGithubLinkStatus()` already expect. Also caught and fixed a stale fallback string left over from the merged-PR metric rename ("commits you have landed" → "pull requests you have had merged") in both `index.html` and the new `play.html` block.

**feature request (Fernando, via Enrique): a toggle to hide developer badges**
Added `showDevBadges` to Options > Interface & Comfort, mirroring the existing `showItemLevel`/`showWalletOnPlayerCard` pattern (persisted in `Settings`, read live, no sim/server involvement). Gates all four surfaces: nameplate glyph + name outline, the inspect-window block, the player card, and the Developers leaderboard tab itself (not just its rows — the tab button doesn't render at all when the toggle is off, with a fallback off the tab if it's toggled off while already selected). Also gated the target-frame dev chip for consistency, since it's the same kind of surface.

**nit (both reviewers): unpinned host on the paginated contributor fetch**
Added `isTrustedGithubApiUrl()` — the paginator now refuses to follow a `Link` header (and re-attach the bearer token) to anything but `https://api.github.com`.

**nit (Enrique): no rate limit on the OAuth callback**
Added one, keyed by IP only (`githubRateLimited(req, 0)`, no account is resolved yet at that point), mirroring how the Discord login-chooser endpoints rate-limit their own unauthenticated entry points.

**nit (Fernando): dead `github.link.failure` metric**
Wired it into every callback failure branch (bad_request/expired/github_error/already_linked/server_error) — a deliberate cancel does not count as a failure.

**nit (both, optional): `bearerActiveAccount` vs `bearerReadAccount` on `GET /api/github`**
Checked the precedent — `/api/wallet` and `/api/discord` both use `bearerActiveAccount` for their status reads too, so left GitHub's as-is for consistency rather than creating a one-off divergence.

Verification: `tsc --noEmit` clean, full suite green (6112/6112 non-skipped tests, including new coverage for the host-pinning, rate limit, and telemetry behavior), `biome check` clean on changed files, both `npm run build` and `npm run build:server` pass.

FernandoX7 (commented):
Re-review at head `44c28b43`. Everything from the two earlier reviews is closed: the play.html Link-GitHub CTA is mirrored (byte-identical to `index.html`), the callback has a per-IP rate limit, `github.link.failure` is wired into every callback failure branch, and the paginator pins the token to `api.github.com`. The merged-PR switch and the new `showDevBadges` toggle both hold up: the toggle gates all five badge surfaces, the tally counts one per merged PR with a total-order ranking, and wire/IWorld parity is intact (147/111 green).

**should-fix (inherited parity, your call), OAuth linking CSRF. `github.ts:91`** The standard binding is correct: `state` is CSPRNG, single-use, and bound to the initiating account, so A's start cannot be redirected to B. But the callback is unauthenticated and GitHub does not bind the `code` to the `state`, so an attacker holding their own valid `state` can lure a victim who is logged into github.com and has already authorized the app into completing `/api/auth/github/callback?code=VICTIM&state=ATTACKER`, binding the victim's GitHub to the attacker's account (cosmetic badge theft, plus the `github_user_id UNIQUE` constraint then 409s the victim's own link). Identical by design to the Discord link flow, so not a regression here. Worth a conscious call: accept as parity, or harden both flows together.

**nit, stale "commits" labels from the rename.** `server/main.ts:1176` emits `metric: 'landedCommits'` (the client never reads `data.metric`, so harmless, but a misnomer now), plus comments at `online.ts:2080`, `leaderboard_page.ts:77`, `hud_chrome.ts:218,937`.

**nit, `tests/world_api_parity.test.ts:3,12,67,68`** header comments still say 146/110; the assertions are correctly 147/111.

**nit (defense in depth), `parseMergedPrLogins` (`github_contributors.ts:49`)** accepts any non-empty `login`, unlike the OAuth path which validates against `GITHUB_LOGIN_RE`. Safe today (host-pinned source, escaped at render in `leaderboard_window.ts:412`); a cheap charset guard closes it symmetrically.

Nothing blocking. Security otherwise checks out (token-scoped routes and unlink, parameterized SQL, env-only secrets, escaped + origin-locked bounce page, additive idempotent DDL), and the sim never reads any of the new fields.

PR #1087, opened 2026-06-30
feat(admin): open account details modal from IP association page

madmatah:
## Summary

Make account names clickable in the Shared IPs investigation view.

Clicking an account name now opens the existing account details modal, allowing moderators to inspect the account without leaving the IP associations page.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npm test`
  - `npm run check:admin`

- Manual steps:
  - Go on admin dashboard / shared IPs
  - click on an IP address
  - click on an account name

## Screenshots / recordings

image

### Quality

### Cross-platform

- [x] **Tested on desktop and mobile.**
- [x] **Accessible.** Account names use the existing keyboard-operable account link component.

### Localization (i18n)

### Hygiene

- [x] No generated files were edited manually.

### DIFF
--- a/src/admin/pages/IpAssociations.svelte
+++ b/src/admin/pages/IpAssociations.svelte
@@ -9,6 +9,7 @@
   import { localizeAdminError, t } from '../i18n';
   import Badge from '../components/Badge.svelte';
   import AccountIndicators from '../components/AccountIndicators.svelte';
+  import AccountLink from '../components/AccountLink.svelte';
   import IpBlockDialog from '../components/IpBlockDialog.svelte';
   import PageHeader from '../components/PageHeader.svelte';
   import Pager from '../components/Pager.svelte';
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
Clean, well-scoped change. It reuses the existing `AccountLink` component (the same one already used in `CharactersTable`, `Accounts`, and `SuspiciousPlayers`) rather than rolling a bespoke clickable, so it inherits the keyboard-operable button, `aria-haspopup="dialog"`, and the shared modal controller for free. The controller is provided by `AdminShell` via `setAccountModalController`, and this page renders inside that shell, so the click resolves in the real app. Confined to the admin dashboard, so the game bundle is untouched. Verified the test against the PR code: 3/3 green.

A couple of notes, both optional:

- nit: `src/admin/pages/IpAssociations.svelte:141` does not pass `onChanged`, so suspending or banning an account from the modal will not refresh the IP associations view (the `status` / indicators stay stale until a manual reload). `Accounts.svelte:98` wires `onChanged={() => void refresh()}` for exactly this reason. `SuspiciousPlayers` omits it too, so this matches an existing precedent, but this page does surface moderation status, so a refresh callback would be a small consistency win. Your call.

- nit: the test (`tests/admin/ip_associations.test.ts`) now drives the click and asserts `accountModalOpen` was called with `(7, undefined)`, which also covers that the name still renders. Nice that it adds coverage rather than just swapping the presence check. No gap here, just noting it reads well.

No i18n concern: the label is the username (data) and the account id already goes through `fmtNumber`, so nothing new to localize.

madmatah:
Fixed, thanks!

FernandoX7 (approved):

PR #1088, opened 2026-06-30
[codex] fix keybindings list scrolling

### LINKED ISSUE #1082: Keybindings window is cut off in fullscreen and the lowest rows are unreachable
## Problem

In fullscreen on a desktop browser, the keybindings window is taller than the viewport, so its lower rows are cut off with no way to scroll to them. The Action Bar bindings near the bottom of the list (the higher-numbered action bars on the CTRL modifier rows) cannot be reached or rebound.

The window does not resize to fit the available height or provide a scrollbar, so part of its content is permanently inaccessible at this resolution.

## Steps to reproduce

1. Enter fullscreen in a desktop browser.
2. Open the keybindings window.
3. Try to reach the lowest Action Bar binding rows.
4. Observe that the bottom of the window is clipped by the viewport and does not scroll.

## Actual behavior

- The keybindings window extends past the bottom of the screen.
- The lowest rows (higher-numbered Action Bar bindings) are cut off and cannot be accessed or rebound.

## Expected behavior

- The keybindings window fits within the viewport at any supported window height.
- When the content is taller than the available space, the window scrolls so every binding is reachable.
- All Action Bar bindings can be viewed and rebound.

## Scope

- Keybindings window sizing and overflow handling.
- Max-height relative to the viewport and a scroll container for the binding list.
- Other settings windows that share the same layout, in case they clip the same way.

## Environment

- Platform: desktop web, browser fullscreen.
- Surface: in-game keybindings settings window.

## Acceptance criteria

- Every keybinding row, including the lowest Action Bar rows, is reachable in fullscreen.
- The window never extends beyond the viewport, scrolling when needed.
- The desktop experience at smaller window sizes is unchanged, not regressed.

## Screenshot

Image

Humpalumps:
## Summary
- Keep the Keybindings title, notes, and footer controls reachable while the binding categories scroll.
- Add a viewport-bounded scroll region around the keybinding grid so the lower Action Bar rows remain accessible in fullscreen and shorter desktop windows.

Fixes #1082

## Validation
- `npx biome check src\ui\options_window.ts src\styles\components.css`
- `npm run check:ts`

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -3490,6 +3490,16 @@
     align-items: start;
     margin-bottom: 8px;
   }
+  .kb-scroll {
+    max-height: max(240px, calc(85vh - 300px));
+    overflow-y: auto;
+    overscroll-behavior: contain;
+    padding-right: 4px;
+    margin-bottom: 8px;
+  }
+  .kb-scroll .kb-cols {
+    margin-bottom: 0;
+  }
   .kb-col {
     min-width: 0;
   }
--- a/src/ui/options_window.ts
+++ b/src/ui/options_window.ts
@@ -1365,7 +1365,10 @@ export class OptionsWindow {
       col.appendChild(rows);
       cols.appendChild(col);
     }
-    el.appendChild(cols);
+    const scroll = document.createElement('div');
+    scroll.className = 'kb-scroll';
+    scroll.appendChild(cols);
+    el.appendChild(scroll);
     const reset = document.createElement('button');
     reset.className = 'btn';
     reset.textContent = t('hud.options.resetToDefaults');


FernandoX7 (commented):
Clean, well-scoped fix. Wrapping just the binding grid in `.kb-scroll` keeps the title, the toggle rows, the note, and the Reset/Back footer fixed while the categories scroll, which is what #1082 needs to make the lowest Action Bar rows reachable. The CSS lands in `src/styles/components.css`, so both `index.html` and `play.html` pick it up through the shared `src/main.ts` style barrel; no play.html mirroring needed. No new strings, so nothing for i18n.

Keyboard access holds up too: every row contains focusable key buttons, so tabbing scrolls them into view, and Reset/Back now sit outside the scroll region so they are always reachable.

One nit:

- nit `src/styles/components.css:3494`: the `calc(85vh - 300px)` reserve only covers the footer, not the fixed block above the grid (the keybind help line, the four toggle rows, and `.kb-note`, all appended to `el` before `.kb-scroll`). On a short window that block plus the 240px grid floor can still push the parent `.window` past its own `calc(85vh - 24px)` cap, so it falls back to the window scrollbar. Nothing ends up unreachable, and fullscreen (the reported case) has plenty of room, so this is fine to leave as is.

Solid fix.

Humpalumps:
Closing as obsolete: issue #1082 is already closed, and the newer keybindings-window PR #1221 superseded this older draft.

PR #1089, opened 2026-06-30
FBX to GLB Scripts

Donny-Deals:
## What

Adds a generic CLI to merge a rigged character's FBX exports into a single `.glb` with all of its animation clips:

- `scripts/combine_fbx_to_glb.mjs` (driver)
- `scripts/combine_fbx_to_glb_entry.js` (browser-side merge)
- one index row in `scripts/CLAUDE.md`

## Why

Standalone Node FBX-to-glTF converters reliably choke on FBX skinning and embedded textures. This tool sidesteps that by parsing FBX with the same three.js `FBXLoader` the game already uses, inside headless Chrome, then exporting via `GLTFExporter` and optimizing with gltf-transform. Clips from separate animation-only FBXs are grafted onto the mesh by bone name.

## Usage

Folder mode (auto-detects the mesh FBX, names each clip from its filename, uses a lone PNG as the base-color texture):

```
node scripts/combine_fbx_to_glb.mjs <folder_of_fbx_plus_png> out.glb
```

Explicit mode:

```
node scripts/combine_fbx_to_glb.mjs --base mesh.fbx --anim walk.fbx --anim run.fbx=Run --tex skin.png out.glb
```

`--help` lists every flag. Defaults are engine-agnostic (standard glTF, PNG textures kept); `--meshopt` and `--webp` are opt-in. Game-mob flags: `--strip-root`, `--rest-clip Idle --rest-t 1`, `--strip-armature` (cooked-axis fix), `--height N`.

## Notes

- No new dependencies (reuses the existing `puppeteer-core`, `three`, `@gltf-transform/*`, `meshoptimizer`, `sharp`, `esbuild`).
- Intermediate files go to the system temp dir; the output directory is created if missing.
- Validated by re-baking an existing asset through it (identical skin/clips/texture output).


FernandoX7 (commented):
Clean, self-contained tool with a clear header and an honest "why headless Chrome" rationale. Parsing FBX with the game's own `FBXLoader` instead of a standalone Node converter is the right call, since those do choke on skinning and embedded textures. I checked the load-bearing claims against the code:

- No new dependencies. `@gltf-transform/*` and `meshoptimizer` aren't direct entries in `package.json`, but they're installed (transitive via `@gltf-transform/cli`) and already imported the same way by sibling scripts (`build_assets.mjs`, `bake_mech_anims.mjs`, `build_foliage.mjs`), so this is consistent, not a new reliance.
- The mechanics match the repo: `headless: 'new'`, `npx esbuild` via `execSync`, `browser_path.mjs`, and the `__ready` / `page.evaluate` bridge all mirror existing render scripts (`render_mech_epics.mjs`, `render_model_stills.mjs`).
- Clean ASCII, no banned punctuation or emoji. (Biome ignores `scripts/*.mjs`, so there is no format gate here anyway.)
- Arg parsing reads correctly: folder vs explicit mode, `--anim file=Name` override, `--webp-size` implying `--webp`, and the self-documenting `--help` pulled from the file's own `//` header.

Two nits, both non-blocking:

- `combine_fbx_to_glb_entry.js:7`: `the game uses -- which is why` uses `--` as a dash. House style avoids dashes; a comma or parentheses reads cleaner.
- `combine_fbx_to_glb.mjs:123,143`: the esbuild bundle path is a fixed name in `tmpdir()` and the raw GLB is keyed only on `basename(out)`, so two concurrent runs can stomp each other's intermediates (one finishing `rmSync`s the shared bundle mid-run). Fine for a manual one-off; a pid or random suffix would make it reentrant.

No sim, server, wire, or UI surface here, so the determinism, server-authority, parity, and i18n invariants do not apply. Looks good.

PR #1090, opened 2026-06-30
[codex] fix browserslist parser on CRLF checkouts

Humpalumps:
## Summary
- Fix the zero-dependency `.browserslistrc` parser on CRLF checkouts by trimming each physical line before stripping `#` comments.
- Add a regression test for CRLF comment lines so Windows checkouts keep loading the shipped browser floors.

## Why
On Windows, `npm run dev` / Vitest startup failed while loading `vite.config.ts` because the parser treated the leading `.browserslistrc` comment as an unsupported floor entry.

## Validation
- `node -e "import('./scripts/browserslist_targets.mjs').then(m=>console.log(JSON.stringify(m.loadBrowserslistFloors('.browserslistrc')))).catch(e=>{console.error(e);process.exit(1)})"`
- `npx biome check scripts\browserslist_targets.mjs tests\browserslist_targets.test.ts`
- `npm test -- tests\browserslist_targets.test.ts`
- `npm run dev -- --host 127.0.0.1 --port 5179` then requested `http://127.0.0.1:5179/play.html` and received HTTP 200

### DIFF
--- a/scripts/browserslist_targets.mjs
+++ b/scripts/browserslist_targets.mjs
@@ -35,7 +35,7 @@ export function parseBrowserslistFloors(text) {
   for (const physicalLine of text.split('\n')) {
     // Strip a '#' comment FIRST, so a comment that happens to contain a comma is
     // not split into a bogus floor entry by the comma handling below.
-    const code = physicalLine.replace(/#.*$/, '');
+    const code = physicalLine.trim().replace(/#.*$/, '');
     for (const raw of code.split(',')) {
       const entry = raw.trim();
       if (!entry) continue;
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
Clean, well-scoped fix with the right diagnosis.

The root cause is exactly as described: on a CRLF checkout `text.split('\n')` leaves a trailing `\r` on each physical line, and in `/#.*$/` the `.` does not match the `\r` line terminator while `$` (no `m` flag) only anchors at the end of the string. So the comment strip never runs on `# floor\r`, the line survives as a bogus `# floor` entry, and the floor regex throws `Unsupported .browserslistrc entry`. Moving `.trim()` ahead of the comment strip removes the `\r` first, which fixes it and also repairs the comment-with-comma case under CRLF for free.

I confirmed the pre-fix parser throws `Unsupported: # floor` on the new test's input, so the regression test fails first and the fix turns it green. Full suite is 9/9, biome clean on both files. No sim/server/wire/ui/i18n surface, so no invariants in play.

Nothing blocking, nothing to change. Nice catch on the regex-vs-line-terminator interaction.

FernandoX7 (approved):

Humpalumps:

PR #1091, opened 2026-06-30
Fix Paladin Judgement damage talents

### LINKED ISSUE #1057: Paladin Judgement damage talents are no-ops (Judgement of Light / Improved Judgement never increase damage)
### Summary
Two Paladin talents advertise extra **Judgement** damage, but neither actually changes Judgement's damage. The talent modifier is computed and stored, yet it is never folded into the ability, because Judgement's damage comes from the *consumed Seal's* stored values, not from the Judgement ability's own effect.

Affected talents:
- **Judgement of Light** (Holy choice node `holy_choice_judgement`): "Judgement deals 20% more damage." -> **completely no-op.**
- **Improved Judgement** (Retribution `ret_imp_judgement`): "Reduces Judgement cooldown by 15% **and increases damage by 10%** per rank." -> the **cooldown** half works, the **+10%/rank damage** half is a no-op.

This is plain solo-play visible: a Holy paladin who picks Judgement of Light gets zero of the promised increase.

### Found while playing
Booted offline and leveled a paladin to 20 to inspect the Holy tree. The tooltip makes the promise on screen:

Judgement of Light tooltip in-game

<sub>(General gameplay shot for context: <a href="https://raw.githubusercontent.com/jgyy/world-of-claudecraft/issue-assets-judgement/issue-1054/gameplay-eastbrook-vale.png">Eastbrook Vale</a>.)</sub>

### Root cause

The talent framework folds a per-ability `dmgPct` into an ability by scaling that ability's own damage-bearing **effects** (`applyTalentMods` -> `scaleEffect`, `src/sim/content/classes.ts`). That works for abilities whose effect carries `min`/`max` (or `judgeMin`/`judgeMax`, `base`, etc).

But Judgement's effect carries **no numbers**:

```ts
// src/sim/content/classes.ts  (judgement ability def)
judgement: {
  id: 'judgement',
  ...
  effects: [{ type: 'judgement' }],   // <- no min/max for scaleEffect to scale
}
```

`scaleEffect` has no `case 'judgement'`, so it hits the `default: return eff;` branch and returns the effect unchanged. The precomputed `mods.abilities.judgement.dmgPct` is therefore silently dropped.

The actual damage is rolled at dispatch time from the **Seal aura's** stored `value2`/`value3` (the seal's `judgeMin`/`judgeMax`), with no read of any Judgement-keyed `dmgPct`:

```ts
// src/sim/combat/effect_dispatch.ts  (case 'judgement')
const seal = p.auras[sealIdx];
...
let dmg =
  ctx.rng.range(seal.value2 ?? 10, seal.value3 ?? 15) +   // seal's stored judge range
  directHitBonus(p.spellPower, ability, res.castTime);
```

So the only way a talent can scale Judgement damage today is by keying its `dmgPct` on **`seal_of_righteousness`** (whose `imbue` effect *does* carry `judgeMin`/`judgeMax` and *is* scaled). The two talents above key on `'judgement'`, so they miss entirely.

### Flow

```mermaid
flowchart TD
    A["Talent: Judgement of Light / Improved Judgement<br/>effect: { ability: 'judgement', dmgPct: +0.2 / +0.1 }"] --> B["computeTalentModifiers<br/>mods.abilities.judgement.dmgPct set"]
    B --> C["applyTalentMods -> scaleEffect(effect)"]
    C --> D{"effect.type === 'judgement'<br/>has min/max?"}
    D -->|"no numbers -> default: return eff"| E["Judgement effect UNCHANGED<br/>(dmgPct dropped)"]
    F["Cast Judgement -> effect_dispatch case 'judgement'"] --> G["dmg = rng.range(seal.value2, seal.value3)<br/>+ spellPower bonus"]
    G --> H["No Judgement dmgPct read here either"]
    E --> H
    H --> I["Final damage identical with or without the talent<br/>(BUG: tooltip promise unfulfilled)"]
```

### Reproduction (deterministic)
Resolve the Paladin's Judgement ability with and without the talent; the resolved effects are byte-identical, while the modifier is provably computed. Both assertions below pass on `main`:

```ts
import { computeTalentModifiers, emptyAllocation, type TalentAllocation } from '../src/sim/content/talents';
import { abilitiesKnownAt } from '../src/sim/content/classes';

const alloc = (o: Partial<TalentAllocation> = {}) => ({ ...emptyAllocation(), ...o });
const judgement = (a: TalentAllocation) =>
  abilitiesKnownAt('paladin', 20, computeTalentModifiers('paladin', a)).find((k) => k.def.id === 'judgement')!;

// Judgement of Light: +20% modifier is computed but never folded in
const withTalent = alloc({ spec: 'holy', ranks: { holy_choice: 1 }, choices: { holy_choice: 'holy_choice_judgement' } });
const mods = computeTalentModifiers('paladin', withTalent);
expect(mods.abilities.judgement?.dmgPct).toBeCloseTo(0.2);              // modifier exists
expect(judgement(withTalent).effects).toEqual(judgement(alloc({ spec: 'holy' })).effects); // ...but no-op

// Improved Judgement: cooldown works, damage half is a no-op
const ret = alloc({ spec: 'retribution', ranks: { ret_imp_judgement: 2 } });
expect(judgement(ret).cooldown).toBeLessThan(judgement(alloc({ spec: 'retribution' })).cooldown); // cooldown ok
expect(judgement(ret).effects).toEqual(judgement(alloc({ spec: 'retribution' })).effects);          // damage no-op
```

### Suggested fix (one option, maintainer's call)
The damage lives on the seal at judge time, so a `'judgement'` `dmgPct` needs to be applied where the seal range is rolled. For example, in `effect_dispatch.ts` `case 'judgement'`, multiply the rolled `dmg` by the resolved Judgement-ability damage modifier before crit/spell-power-independent rounding (or, alternatively, re-target both talents' `dmgPct` onto `seal_of_righteousness`, though that also scales the Seal's on-hit proc, which may be undesirable). Either way a `tests/` case like the one above should pin it.

### Environment
- Branch: `main` (also present on the current `release/*`); `src/sim/` core, identical across offline/server/headless hosts.
- Files: `src/sim/content/talents_classic.ts` (talent defs), `src/sim/content/classes.ts` (`applyTalentMods`/`scaleEffect`), `src/sim/combat/effect_dispatch.ts` (`case 'judgement'`).

Humpalumps:
## Summary
- Apply effective Paladin Judgement damage modifiers when Judgement consumes an active Seal.
- Keep Seal-specific scaling on the Seal aura itself, so Seal of Righteousness talents still snapshot into the Seal and Judgement-specific talents only scale the Judgement impact.
- Add a deterministic combat-effect test for Holy's Judgement of Light modifier: a 100-damage Seal Judgement stays 100 untalented and becomes 120 with the +20% talent.

Fixes #1057

## Testing
- npx biome check --max-diagnostics=10 src\\sim\\combat\\effect_dispatch.ts tests\\talents.test.ts
  - Passed. Biome reports existing warnings in tests/talents.test.ts, but no errors.
- npm run check:ts
  - Passed.
- npx vitest run --config ..\\work\\vitest.node.config.ts tests\\talents.test.ts tests\\social.test.ts
  - Passed: 2 files, 100 tests.

## Notes
- I also attempted the normal 
pm test -- tests\\talents.test.ts tests\\social.test.ts command after allowing the pretest generation step. Pretest completed, but Vitest could not load ite.config.ts on this release branch because .browserslistrc comments are rejected by the current parser. That is the same unrelated Windows/release-branch blocker addressed separately in #1090, so I did not include that fix in this PR.

### DIFF
--- a/src/sim/combat/effect_dispatch.ts
+++ b/src/sim/combat/effect_dispatch.ts
@@ -244,9 +244,14 @@ export function runEffects(
         p.auras.splice(sealIdx, 1);
         ctx.emit({ type: 'aura', targetId: p.id, name: seal.name, gained: false });
         // Judgement is an instant holy nuke; scale it with Spell Power too.
+        const mods = ctx.playerMods(meta);
+        const judgementMod = mods.abilities[ability.id];
+        const judgementMult = 1 + mods.global.spellDmgPct + (judgementMod?.dmgPct ?? 0);
+        const judgementFlat = judgementMod?.flatDmg ?? 0;
         let dmg =
-          ctx.rng.range(seal.value2 ?? 10, seal.value3 ?? 15) +
-          directHitBonus(p.spellPower, ability, res.castTime);
+          Math.round(
+            ctx.rng.range(seal.value2 ?? 10, seal.value3 ?? 15) * judgementMult + judgementFlat,
+          ) + directHitBonus(p.spellPower, ability, res.castTime);
         const crit = ctx.rng.chance(ctx.spellCrit(p));
         if (crit) dmg *= 1.5;
         ctx.dealDamage(p, target, Math.round(dmg), crit, 'holy', ability.name, 'hit');
--- a/tests/talents.test.ts
+++ b/tests/talents.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from 'vitest';
 import { ClientWorld } from '../src/net/online';
+import { runEffects } from '../src/sim/combat/effect_dispatch';
 import { ABILITIES, abilitiesKnownAt } from '../src/sim/content/classes';
 import {
   computeTalentModifiers,
@@ -19,8 +20,9 @@ import {
   validateAllocation,
   validateTalentTree,
 } from '../src/sim/content/talents';
-import { Sim } from '../src/sim/sim';
-import { ALL_CLASSES, dist2d, MAX_LEVEL } from '../src/sim/types';
+import { type PlayerMeta, Sim } from '../src/sim/sim';
+import type { SimContext } from '../src/sim/sim_context';
+import { ALL_CLASSES, dist2d, type Entity, MAX_LEVEL } from '../src/sim/types';
 import { terrainHeight } from '../src/sim/world';
 import { talentChoiceIconRef, talentNodeIconRef } from '../src/ui/talent_icons';
 
@@ -50,6 +52,43 @@ function nearestMob(sim: Sim) {
 }
 
 const effOf = (k: any, i = 0) => k.effects[i] as any;
+
+function judgementHitAmount(mods: ReturnType<typeof computeTalentModifiers>): number {
+  const judgement = abilitiesKnownAt('paladin', 20, mods).find((k) => k.def.id === 'judgement');
+  if (!judgement) throw new Error('missing Judgement ability');
+  const paladin = {
+    id: 1,
+    spellPower: 0,
+    auras: [
+      {
+        id: 'seal_of_righteousness',
+        name: 'Seal of Righteousness',
+        kind: 'imbue',
+        value2: 100,
+        value3: 100,
+      },
+    ],
+  } as Partial<Entity> as Entity;
+  const target = { id: 2, dead: false } as Partial<Entity> as Entity;
+  const meta = { talentMods: mods, fiestaMods: null } as Partial<PlayerMeta> as PlayerMeta;
+  let amount = 0;
+  const ctx = {
+    rng: { range: () => 100, chance: () => false },
+    breakStealth: () => undefined,
+    emit: () => undefined,
+    error: (_pid: number, text: string) => {
+      throw new Error(text);
+    },
+    playerMods: (m: PlayerMeta) => m.fiestaMods ?? m.talentMods,
+    spellCrit: () => 0,
+    dealDamage: (_source: Entity | null, _target: Entity, damage: number) => {
+      amount = damage;
+    },
+  } as unknown as SimContext;
+
+  runEffects(ctx, paladin, meta, target, judgement);
+  return amount;
+}
 
 describe('talent tree validation (load-time)', () => {
   it('every registered tree is structurally valid', () => {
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Good, focused fix, and the diagnosis matches #1057 exactly: Judgement's damage is rolled from the consumed Seal's `value2`/`value3`, so a `judgement`-keyed `dmgPct` never reaches it through `scaleEffect`. Applying the Judgement ability's mods where the seal range is rolled is the right place to fix it, and driving the real `runEffects` from a deterministic stub is a clean way to pin it.

One issue to sort out before this is ready.

**should-fix: global `spellDmgPct` is double-counted in Judgement damage.** `effect_dispatch.ts:233` builds `judgementMult = 1 + mods.global.spellDmgPct + (judgementMod?.dmgPct ?? 0)`, but `mods.global.spellDmgPct` is already baked into the seal roll. The Seal's `judgeMin`/`judgeMax` are scaled at seal-cast time by `applyTalentMods` -> `scaleEffect` (the `imbue` case, `classes.ts:3549`), where `dmgMult = 1 + spellDmgPct + seal.dmgPct` (`classes.ts:3588`). `meta.known` is resolved with `meta.talentMods` (`sim.ts:2175`), and in normal play `playerMods(meta)` returns that same `talentMods` (fiestaMods null), so the impact multiplier re-applies the identical global a second time.

This is reachable in base play, on the spec that actually uses Judgement to deal damage. A Retribution paladin has `global.spellDmgPct` 0.06 with zero points spent, rising to 0.14 with Sanctity Aura. Seal of Righteousness `judgeMin`/`judgeMax` move 32/47 -> 34/50 across that swing, confirming the global is in the snapshot. With Sanctity Aura + Improved Judgement the Judgement hit lands about 11.7% high (multiplier 1.34 vs the intended 1.2); plain Ret is about 6% high. Holy and Prot have `global.spellDmgPct` 0, which is why the new test (holy spec, seal range hardcoded to 100) passes at exactly 120 and never touches this path.

The seal snapshot already carries the global, so the impact multiplier should only add the Judgement-specific part:

```ts
const judgementMult = 1 + (judgementMod?.dmgPct ?? 0);
```

**nit:** the test pins the holy/no-global case well but can't catch the above, because it hardcodes the seal range and uses a spec whose global spell damage is 0. A second case with a Retribution allocation (real resolved seal values plus Sanctity Aura) would guard the double-count.

Rest of the change checks out: RNG draw order is unchanged (`playerMods` is pure), `directHitBonus` stays outside the multiplier like the `directDamage` path, and sim purity holds.

Humpalumps:

PR #1092, opened 2026-06-30
Fix Parchment high-contrast text blur

### LINKED ISSUE #1081: High-Contrast Text blurs text on the Parchment UI theme
## Problem

With the Parchment UI theme selected, turning on High-Contrast Text makes body text render blurry instead of sharper. The effect is visible across the Interface settings panel: labels and values look smeared rather than crisp.

High-Contrast Text is an accessibility option, so this is the opposite of its intent: it should improve legibility, not degrade it.

## Steps to reproduce

1. Open the Interface settings.
2. Set UI Theme to Parchment.
3. Turn High-Contrast Text on.
4. Observe the text in the settings panel (and other parchment-themed surfaces).

## Actual behavior

- Text becomes blurry and harder to read once High-Contrast Text is on under the Parchment theme.

## Expected behavior

- High-Contrast Text renders crisp, sharp text on the Parchment theme.
- Enabling the option improves legibility on Parchment, consistent with the other themes.

## Scope

- The High-Contrast Text styling as it combines with the Parchment theme tokens.
- Whether a text-shadow, stroke, or glow applied for high contrast is what smears against the parchment background.
- Confirm the same toggle stays sharp on Classic Gold, Midnight, and High Contrast themes, to isolate the Parchment interaction.

## Acceptance criteria

- With Parchment plus High-Contrast Text on, text is sharp and legible.
- High-Contrast Text continues to render correctly on every other theme.
- No regression to text rendering when High-Contrast Text is off.

## Screenshot

Image

Humpalumps:
## Summary
- Add a data-ui-theme body hook when the saved UI theme is applied on the start screens and in-game HUD.
- Keep the existing High-Contrast Text outline for dark themes, but disable that multi-layer text shadow for Parchment surfaces.
- Add a CSS regression test that pins the body theme hook and the Parchment-specific high-contrast override.

Fixes #1081

## Root cause
High-Contrast Text adds a dark multi-layer 	ext-shadow around HUD text. That works on dark themes, but Parchment already uses dark text on a light panel. The extra shadow creates a smeared halo, making text blurrier instead of more legible.

## Testing
- 
px biome check --max-diagnostics=20 src\\main.ts src\\styles\\hud.css tests\\client_shell.test.ts
  - Passed. Reports existing warnings in src/main.ts; no errors.
- 
pm run check:ts
  - Passed.
- 
px vitest run --config ..\\work\\vitest.node.config.ts tests\\client_shell.test.ts tests\\theme.test.ts tests\\settings.test.ts
  - Passed: 3 files, 102 tests.
- 
pm test -- tests\\client_shell.test.ts tests\\theme.test.ts tests\\settings.test.ts
  - Attempted after allowing the pretest generation step. Pretest completed, then Vitest failed to load ite.config.ts because the current release branch rejects .browserslistrc comments. This is the same unrelated release-branch blocker addressed in #1090, so this PR leaves that fix separate.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -913,6 +913,7 @@ async function startGame(
   const themeStore = new ThemeStore();
   function applyTheme(): void {
     const vars = themeStore.cssVars();
+    document.body.dataset.uiTheme = themeStore.get().preset;
     for (const name of Object.keys(vars))
       document.documentElement.style.setProperty(name, vars[name]);
   }
@@ -7024,7 +7025,9 @@ function fadeOutHomepageMusic(durationMs = 1600): void {
 // (startGame() re-applies via its own ThemeStore once the world loads.)
 (() => {
   try {
-    const vars = new ThemeStore().cssVars();
+    const themeStore = new ThemeStore();
+    const vars = themeStore.cssVars();
+    document.body.dataset.uiTheme = themeStore.get().preset;
     for (const name of Object.keys(vars))
       document.documentElement.style.setProperty(name, vars[name]);
   } catch {
--- a/src/styles/hud.css
+++ b/src/styles/hud.css
@@ -2400,6 +2400,13 @@
       1px 1px 2px #000,
       -1px -1px 2px #000;
   }
+  body.high-contrast-text[data-ui-theme="parchment"] .chat-pane div,
+  body.high-contrast-text[data-ui-theme="parchment"] .panel,
+  body.high-contrast-text[data-ui-theme="parchment"] .set-name,
+  body.high-contrast-text[data-ui-theme="parchment"] .tt-stat,
+  body.high-contrast-text[data-ui-theme="parchment"] .tt-desc {
+    text-shadow: none;
+  }
   /* Reduce Motion, drop HUD animations/transitions (mirrors prefers-reduced-motion). */
   body.reduce-motion .fct {
     animation-duration: 0.01ms !important;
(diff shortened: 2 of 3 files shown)


FernandoX7 (commented):
Clean, well-scoped fix, and the verification holds up:

- The override kills the shadow for exactly the five selectors the high-contrast rule sets it on (`.chat-pane div`, `.panel`, `.set-name`, `.tt-stat`, `.tt-desc`), so nothing Parchment is left smeared, and the `[data-ui-theme="parchment"]` attribute raises specificity above the base `body.high-contrast-text` rule, so it wins regardless of source order.
- `data-ui-theme` is set on every path that matters: in-game `applyTheme`, the pre-paint start-screen IIFE, and live via the options picker (`setPreset`/`setCustom`/reset all re-run `applyTheme`), so toggling Parchment updates immediately rather than on reload.
- Both `index.html` and `play.html` load `/src/main.ts` and the shared `src/styles` barrel, so the JS hook and the CSS reach `/play` for free, no separate mirror needed.
- Removing the shadow does not cost contrast: Parchment is dark text on a light panel, already above AA (theme.test.ts guards that), and the shadow was the feature's only effect on it, so disabling just the shadow is the right scope.

nit (`tests/client_shell.test.ts`): `expect(hudCss).toContain('text-shadow: none;')` also matches a pre-existing rule elsewhere in the file, so it would pass even if the new declaration were deleted. The four parchment-selector assertions above it already pin the block, so this is cosmetic, but scoping it to the parchment rule would make it fail for the right reason.

Minor: the start-screen IIFE sets the theme vars on `documentElement` but `data-ui-theme` on `body`. Correct as written (the selector keys off `body`, where `high-contrast-text` also lives), just a split worth a word of comment so it reads as intentional.

Humpalumps:
Closing as duplicate/superseded by #1107, which is the active high-contrast Parchment text fix.

PR #1093, opened 2026-06-30
Stop arena knockback from tunneling through walls

### LINKED ISSUE #1085: Arena PvP: players can move through the walls and keep attacking
## Problem

In ranked arena PvP (The Ashen Coliseum), players report being able to pass through the arena's enclosing walls and keep fighting from a position they should not be able to reach. A combatant who ends up on the far side of, or partway into, a wall can attack across it or kite from outside the pit, which breaks match integrity and the fairness of ranked results.

The arena is server authoritative (the deterministic sim core resolves all movement and combat), so escaping the bounds is a simulation level collision or anti-cheat gap, not a purely cosmetic client issue.

## Steps to reproduce

1. Queue for and enter a ranked arena match in The Ashen Coliseum.
2. Drive your character into one of the arena's enclosing walls. 
3. Observe the character end up on the far side of, or inside, the wall rather than being stopped by it.
4. From that position, attack the opponent.

## Expected behavior

- The arena's enclosing walls are solid. A player cannot walk, jump, charge, flee, or be knocked or displaced through them.
- Every combatant stays inside the arena bounds for the full match.
- Attacks that require line of sight cannot resolve through an arena wall.
- Movement and combat outcomes in the arena match the rest of the world's collision rules.

## Actual behavior

- A player can reach a position on the far side of, or inside, an arena wall and continue fighting from there.

## Scope

Please investigate the arena movement and combat paths in the sim core:

- The arena collision set and resolution: `ARENA_LAYOUT` and `ARENA_COLLIDERS`, and the `isArenaPos` branch of `resolvePosition` / `resolveMovement` in `src/sim/colliders.ts`. Confirm the wall geometry fully encloses the pit at the player body radius, with no gap at wall joints or corners.
- Displacement paths that may bypass the swept wall check. `applyKnockback` in `src/sim/sim.ts` steps the victim outward checking only terrain height (deep water and cliffs), then resolves only the final landing point with `resolvePosition` (a point check), not the swept `resolveMovement` used by normal walking. A knockback aimed across a thin arena wall can deposit the victim on the far side. Confirm whether the other non walk paths (charge, fear) are swept resolved.
- Melee line of sight. Physical abilities with range at or below `MELEE_RANGE`, and melee auto attack, do not require line of sight (`abilityNeedsLineOfSight` in `src/sim/sim.ts`), so a player next to a wall can melee a target on the other side. Decide whether melee should require line of sight inside the arena.

## Acceptance criteria

- A player cannot end a tick outside the arena bounds by any movement path: normal walk, jump, charge, fear, or knockback and other displacements.
- Knockback and other displacements are clamped by the same swept wall resolution as normal movement, with no tunneling through thin walls.
- Line of sight required attacks cannot resolve through an arena wall, and the intended rule for melee through a wall in the arena is decided and enforced.
- Arena matchmaking, Elo, and all non arena movement and combat are unchanged.
- A determinism safe test covers the failing path (for example, a knockback aimed at a thin arena wall leaves the victim inside the pit).

Humpalumps:
## Summary
- Route knockback displacement through swept movement resolution on each step so arena walls and other thin colliders clip the shove instead of only correcting the final landing point.
- Stop knockback at the clipped position and report the actual distance moved.
- Add a regression test proving arena-side knockback no longer lands a player beyond the wall.

## Root cause
`applyKnockback` walked terrain in 0.5 yd increments, but it only called point resolution after choosing the final landing position. A long shove could therefore step from inside the arena to a point already outside the thin wall collider; `resolvePosition` had nothing to push out of because the endpoint was past the wall volume. Using the existing swept `resolveMove` path during each knockback step catches the wall crossing before the target tunnels through.

## Testing
- `npx vitest run tests\mob_knockback.test.ts --config ..\work\vitest.node.config.ts` failed before the fix with `expected 4228.5 to be less than 4222`.
- `npx biome format --write src\sim\sim.ts tests\mob_knockback.test.ts`
- `npx biome check --max-diagnostics=40 src\sim\sim.ts tests\mob_knockback.test.ts` exits 0; remaining output is existing warnings for non-null assertions / `any` usage in these files.
- `npm run check:ts`
- `npx vitest run tests\mob_knockback.test.ts tests\arena.test.ts tests\pathfind.test.ts --config ..\work\vitest.node.config.ts` (3 files / 50 tests)
- `npm test -- tests\mob_knockback.test.ts` was also attempted outside the sandbox after the sandbox blocked esbuild pretest reads. Pretest generation succeeded, then normal Vitest hit the existing release-branch `.browserslistrc` comment parser blocker fixed separately by #1090.

Fixes #1085

### DIFF
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -3087,8 +3087,8 @@ export class Sim {
   // On-hit knockback: hurl `target` up to `distance` yards straight away from
   // `source`. Instantaneous displacement (no aura) walked in small steps so it can
   // be terrain-clamped exactly like a warrior charge — the shove stops at the last
-  // safe footing before deep water or a cliff rather than stranding the victim off
-  // the world. Returns the yards actually moved (0 if blocked immediately).
+  // safe footing before a wall, deep water, or a cliff rather than
+  // stranding the victim off the world. Returns the yards actually moved (0 if blocked immediately).
   private applyKnockback(source: Entity, target: Entity, distance: number): number {
     let dx = target.pos.x - source.pos.x;
     let dz = target.pos.z - source.pos.z;
@@ -3113,9 +3113,14 @@ export class Sim {
       const h1 = groundHeight(nx, nz, this.cfg.seed);
       if (h1 < WATER_LEVEL - SWIM_DEPTH) break; // would land in deep water
       if (h1 > h0 && (h1 - h0) / adv > MAX_CLIMB_SLOPE) break; // would slam into a cliff
-      cx = nx;
-      cz = nz;
-      moved += adv;
+      const resolved = this.resolveMove(cx, cz, nx, nz, BODY_RADIUS, target);
+      const actual = Math.hypot(resolved.x - cx, resolved.z - cz);
+      if (actual < 1e-4) break;
+      const clipped = Math.hypot(resolved.x - nx, resolved.z - nz) > 1e-3;
+      cx = resolved.x;
+      cz = resolved.z;
+      moved += actual;
+      if (clipped) break;
     }
     if (moved <= 0) return 0;
     // resolveMovePoint is a no-op wrapper over resolvePosition outside a delve, and
--- a/tests/mob_knockback.test.ts
+++ b/tests/mob_knockback.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, it } from 'vitest';
-import { Sim } from '../src/sim/sim';
-import { MOBS } from '../src/sim/data';
+import { arenaOrigin, MOBS } from '../src/sim/data';
 import { createMob } from '../src/sim/entity';
+import { Sim } from '../src/sim/sim';
 
 const SEED = 5150;
 const makeSim = () => new Sim({ seed: SEED, playerClass: 'warrior' });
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Clean fix with the root cause correctly nailed. The old `applyKnockback` only point-resolved the final landing spot, so a long shove could step from inside the arena straight to a point already past the thin wall OBB, where `resolvePosition` had no volume to push back out of. Sweeping each step through `resolveMove` catches the crossing mid-step, which is the same approach charge and follow already use (`sim.ts:2618`, `sim.ts:2680`).

What I verified:
- Reverted only `sim.ts` to the base: the new test fails with `expected 4228.5 to be less than 4222`, and passes with the fix. So it is a genuine failing-first regression test.
- The geometry holds up: the arena side wall centres at local x 23 with half-thickness 1, so the inner face is at +22, and the start at local x 20.5 is a real `isArenaPos` so the arena colliders actually engage rather than the test passing trivially.
- `mob_knockback`, `arena`, `pathfind`, and `architecture` suites all green (74 tests). Determinism and sim purity are intact: pure geometry, no new RNG draws (the `rng.chance` proc gate is upstream and unchanged), no new imports.

One behavior note, not a defect: knockback now stops the moment a step clips any collider (`clipped` break), where charge slides along the surface instead. So a shove that grazes a pillar or the central dais now halts at first contact rather than sliding past it. That reads as the intended "stops at the last safe footing" behavior and matches the updated comment, just flagging it as a deliberate choice in case you want shoves to slide.

nit: `sim.ts:3128`, the trailing `resolveMovePoint(cx, cz, ...)` is now largely redundant for the open-world and arena paths since `cx,cz` is already a swept-resolved position, but it still does the delve module/door layering and is harmless, so leaving it is fine.

Humpalumps:
Closing as duplicate/superseded by #1109, which is the active arena wall tunneling fix.

PR #1094, opened 2026-06-30
Rate-limit inbound websocket messages

### LINKED ISSUE #978: no global inbound message-rate limit
### Before you start

### What happened?

  #: LOW-1
  Sev: LOW
  Where: server/main.ts:1513
  Issue: no global inbound message-rate limit (only chat is
    throttled) → CPU/DoS pressure, not state corruption
  Fix: per-session token bucket in handleMessage

### Steps to reproduce

 LOW-1 — no message-rate limit
  1. npm run db:up && npm run server (authoritative server on
  :8787), connect a client.
  2. After auth, send a flood of {t:'input',...} or
  {t:'cast',...} frames in a tight loop (any WS client).
  3. Observe: every frame is processed (only chat is
  throttled). Effect is CPU/DoS pressure — not state corruption
  (GCD/cooldowns still gate actions). No throttle/disconnect
  triggers.

na

### Where did this happen?

Offline (single-player in the browser)

### Device

Desktop

### Browser / OS

_No response_

_No response_

Humpalumps:
## Summary
- Add a per-session inbound WebSocket token bucket that runs before JSON parsing and message dispatch.
- Close a flooding session with policy code `1008` and a client-visible error once the bucket is exhausted.
- Cover non-chat input floods with a regression test so ordinary chat throttling is no longer the only inbound protection.

## Root cause
Only chat commands had rate limiting. Other authenticated frame types, especially high-frequency `input` frames, went straight through `handleMessage` and `JSON.parse` with no global per-session budget. A tight client loop could therefore force server parse/dispatch work even though sim cooldowns and GCDs still protected gameplay state.

## Testing
- `npx biome format --write server\game.ts tests\snapshots.test.ts`
- `npx biome check --max-diagnostics=40 server\game.ts tests\snapshots.test.ts` exits 0; output is existing warnings in `tests/snapshots.test.ts` (`any` / non-null assertions).
- `npx vitest run tests\snapshots.test.ts tests\security.test.ts --config ..\work\vitest.node.server.config.ts` (2 files / 119 tests)
- `npm run check:ts`
- `npx vitest run tests\ws_buffer.test.ts tests\game_sessions.test.ts tests\chat_log.test.ts --config ..\work\vitest.node.server.config.ts` (3 files / 41 tests)
- `npm test -- tests\snapshots.test.ts` was attempted in the sandbox and failed during esbuild pretest reads with `Access is denied`.
- The same `npm test -- tests\snapshots.test.ts` was attempted outside the sandbox: i18n/wiki pretest generation succeeded, then normal Vitest hit the existing release-branch `.browserslistrc` comment parser blocker fixed separately by #1090.

Fixes #978

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -109,6 +109,8 @@ const CHAT_RATE_REFILL_PER_SECOND = 1 / 3; // sustained 20 messages/minute
 const CHAT_RATE_ERROR_COOLDOWN_SECONDS = 4;
 const CHAT_COOLDOWN_SECONDS = 20;
 const CHAT_RATE_VIOLATIONS_FOR_COOLDOWN = 3;
+const INBOUND_RATE_BURST = 120;
+const INBOUND_RATE_REFILL_PER_SECOND = 40; // safely above the normal 20Hz input stream
 const WHO_RESULT_LIMIT = 50;
 const MAX_ACTIVE_SESSIONS_PER_ACCOUNT = 2;
 const RESTART_COUNTDOWN_TOTAL_SECONDS = 600;
@@ -315,6 +317,9 @@ export interface ClientSession {
   chatLastRateError: number;
   chatRateViolations: number;
   chatCooldownUntil: number;
+  inboundTokens: number;
+  inboundLastRefill: number;
+  inboundRateLimited: boolean;
   chatMutedUntil: number | null;
   chatMuteReason: string;
   // Hard-word enforcement strike count driving the mute ladder. Account-scoped:
@@ -1369,6 +1374,9 @@ export class GameServer {
       chatLastRateError: 0,
       chatRateViolations: 0,
       chatCooldownUntil: 0,
+      inboundTokens: INBOUND_RATE_BURST,
+      inboundLastRefill: Date.now() / 1000,
+      inboundRateLimited: false,
       chatMutedUntil: meta.mutedUntil ? new Date(meta.mutedUntil).getTime() : null,
       chatMuteReason: meta.reason ?? '',
       chatStrikes: meta.chatStrikes ?? 0,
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
Good, well-targeted change. The bucket runs before `JSON.parse` and dispatch so a flood can't force parse/dispatch work, the latch plus 1008 close is clean, and the regression test covers the non-chat input path that was previously unguarded. Relying on `ws.close()` to fire the `on('close')` to `leave()` path (rather than calling `leave()` like `kickSession`) is fine here, since `leave()` is idempotent.

One blocking concern on the tuning.

**blocking** `server/game.ts:113` `INBOUND_RATE_REFILL_PER_SECOND = 40` is below the client's real sustained input rate, so this will disconnect actively-playing users. The comment says "safely above the normal 20Hz input stream", but 20Hz is only the 50ms `sendInput` timer (`src/net/online.ts:915`). The change-driven `flushInput()` path runs every render frame (`src/main.ts:2264`) and sends an extra frame whenever the input signature changes. The signature includes `facing` quantized to 0.0001 rad (`inputSignature`, online.ts:963), so during continuous mouselook (strafing around a mob, kiting, panning the camera while running, all normal play) it changes every frame. `flushInput` then sends up to ~62.5/sec (its internal 16ms guard, online.ts:990) on top of the 20/sec timer.

So at 60fps a normally-active player streams roughly 60 to 80 inbound frames/sec against a 40/sec refill; with burst 120 the bucket drains in about 3 to 6 seconds, then the session is force-closed with 1008 and "Too many messages". A token-bucket sim confirms: 60/sec disconnects at ~6s, 80/sec at ~3s, and even a 30fps client (~50/sec) at ~12s. As tuned, a normal combat session will eventually kick people.

The refill needs to clear the real client peak (the 16ms guard caps `flushInput` at ~62.5/sec, plus the 20/sec timer, so ~80/sec), or `input` frames need a separate/looser budget from the rest, or the burst needs to be much larger. Worth driving the test from the actual client send cadence rather than a 160-message loop so the chosen numbers are pinned against real traffic.

**nit** `tests/snapshots.test.ts:498` the test fires 160 frames in a tight synchronous loop and relies on `Date.now()` not advancing enough to refill mid-loop. That holds because the loop is sub-millisecond, but it is wall-clock dependent. Fine as is, just noting it.

Humpalumps:
Closing as duplicate/superseded by #1108, which is the active global inbound WebSocket rate-limit fix.

PR #1096, opened 2026-06-30
Stop duel auto attacks on stealthed opponents

### LINKED ISSUE #491: Stealth in a duel: brief client/server desync — opponent is hidden but still being auto-attacked
Filing as a **behavior question**, not a confirmed bug — it self-corrects quickly and may well be intended. Flagging in case the asymmetry is unintended.

## Observed
`canObserveEntity` hides a **hostile stealthed player from the client at all ranges**:

```ts
// server/game.ts
private canObserveEntity(viewer: Entity, e: Entity, d2: number): boolean {
  if (e.kind !== 'player' || !isStealthed(e)) return true;
  if (this.sim.isHostileTo(viewer, e)) return false;   // hostile + stealthed → never serialized
  ...
}
```

That looks intentional (from the "hide stealthed hostile players" change). But the **sim** has no observability concept, so during a duel:

1. A and B are dueling (mutually hostile); B is auto-attacking A.
2. A stealths → A stops being serialized to B's client (B sees nothing).
3. B's `updatePlayerAutoAttack` keeps swinging at A server-side (the target/hostility/range are all still valid — stealth doesn't clear an existing attacker's target), and A can keep hitting B.
4. The first landed hit calls `breakStealth` (`src/sim/sim.ts` ~L3277), so A reappears.

So there's a brief window (bounded by the swing timer) where B is taking/dealing damage to an opponent the client renders as absent — B appears to swing at nothing and gets hit from an invisible source until stealth breaks.

## Question
Is the intended behavior that **the victim's combat also pauses** against a hidden opponent (drop target / suppress auto-attack until re-observed), or is the short "hidden but still in combat" window acceptable because stealth breaks on the next hit anyway? If the latter, this can be closed — just wanted it on the radar since the client-hide and the sim-combat paths don't currently agree.

(No PR attached — didn't want to change the deliberate stealth-visibility design without your call on the intended behavior.)

Humpalumps:
## Summary

Stops player auto-attack from continuing against an active duel opponent who has entered stealth. The server already hides stealthed duel opponents from the hostile client, so the sim now also refuses to keep swinging at that hidden opponent.

This also rejects newly starting auto-attack against a stealthed active duel opponent and adds duel regression coverage for both paths.

## Related issues

Fixes #491.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src\sim\combat\auto_attack.ts tests\duel.test.ts`
  - `npx biome check --max-diagnostics=40 src\sim\combat\auto_attack.ts tests\duel.test.ts` (passes; existing warnings remain in `tests/duel.test.ts`)
  - `npx vitest run tests\duel.test.ts --config ..\work\vitest.node.config.ts`
  - `npx vitest run tests\interest.test.ts --config ..\work\vitest.node.server.config.ts`
  - `npx vitest run tests\pvp_safety.test.ts tests\combat_auto_attack.test.ts tests\attack_command.test.ts tests\targeting.test.ts --config ..\work\vitest.node.config.ts`
  - `npx vitest run tests\threat.test.ts tests\stealth_render.test.ts tests\sim.test.ts --config ..\work\vitest.node.config.ts`
  - `npm run check:ts`
  - `git diff --check`
- Manual steps:
  - N/A. This is a deterministic sim behavior change covered by regression tests.

`npm test -- tests\duel.test.ts` was also attempted through the project entry point. Inside the sandbox it failed before Vitest because esbuild could not read source directories. Outside the sandbox, generators completed, then Vitest stopped while loading `vite.config.ts` on the known release-branch `.browserslistrc` comment parsing issue: `Unsupported .browserslistrc entry ... # CSS engine floor ...`.

## Screenshots / recordings

N/A. This PR does not change UI/UX.

---

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/combat/auto_attack.ts
+++ b/src/sim/combat/auto_attack.ts
@@ -48,13 +48,19 @@ import { blindMissBonus, isDisarmed, isStunned } from './cc';
 import { baseSwingSpeed } from './form_swing';
 import { applyThornsReaction } from './thorns_charge';
 
+function isHiddenStealthedDuelTarget(ctx: SimContext, attacker: Entity, target: Entity): boolean {
+  if (target.kind !== 'player' || !target.stealthed) return false;
+  const duel = ctx.duelFor(attacker.id);
+  return duel?.state === 'active' && (duel.a === target.id || duel.b === target.id);
+}
+
 export function startAutoAttack(ctx: SimContext, pid?: number): void {
   const r = ctx.resolve(pid);
   if (!r) return;
   const p = r.e;
   if (p.dead) return;
   const t = p.targetId !== null ? ctx.entities.get(p.targetId) : null;
-  if (!t || t.dead || !ctx.isHostileTo(p, t)) {
+  if (!t || t.dead || !ctx.isHostileTo(p, t) || isHiddenStealthedDuelTarget(ctx, p, t)) {
     ctx.error(p.id, 'Invalid attack target.');
     return;
   }
@@ -90,7 +96,7 @@ export function updatePlayerAutoAttack(ctx: SimContext, p: Entity, meta: PlayerM
   p.swingTimer = Math.max(0, p.swingTimer - DT);
   if (!p.autoAttack || p.castingAbility) return;
   const t = p.targetId !== null ? ctx.entities.get(p.targetId) : null;
-  if (!t || t.dead || !ctx.isHostileTo(p, t)) {
+  if (!t || t.dead || !ctx.isHostileTo(p, t) || isHiddenStealthedDuelTarget(ctx, p, t)) {
     p.autoAttack = false;
     return;
   }
--- a/tests/duel.test.ts
+++ b/tests/duel.test.ts
@@ -1,14 +1,20 @@
 import { describe, expect, it } from 'vitest';
 import { Sim } from '../src/sim/sim';
-import type { Aura, Entity } from '../src/sim/types';
+import type { Aura, Entity, PlayerClass } from '../src/sim/types';
 import { groundHeight } from '../src/sim/world';
 
 function makeWorld() {
   return new Sim({ seed: 42, playerClass: 'warrior', noPlayer: true });
 }
 
+function mustEntity(sim: Sim, pid: number): Entity {
+  const e = sim.entities.get(pid);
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Focused fix and the regressions are solid: both the existing-attack and the new-start paths are covered, and pinning `attacker.facing` at the rogue means the swing would actually land without the gate, so the tests would fail if the fix regressed. The predicate is pure (no rng, no wall-clock), so existing parity goldens that do not touch a stealthed duel stay byte-identical. The `mustEntity` / `PlayerClass` cleanups are a nice touch.

One thing worth addressing before this lands.

**should-fix** `src/sim/combat/auto_attack.ts:51`: the gate is duel-only, but the server policy it mirrors is not. `canObserveEntity` (server/game.ts:2800) hides *any* hostile stealthed player, and `isHostileTo` is true for two player-vs-player cases: an active duel and an active arena cross-team match (sim.ts:4534-4540). So a rogue who stealths in a ranked arena bout is hidden from the opponent's client but still auto-attacked server-side, which is the exact desync #491 describes, just in arena instead of a duel.

The `||` chain already short-circuits on `!ctx.isHostileTo(p, t)` immediately before this predicate runs, so hostility is guaranteed true by the time `isHiddenStealthedDuelTarget` is reached. That means the `duelFor` lookup re-derives something hostility already implies, and it is what narrows the fix to duels. Mirroring the server exactly only needs:

```ts
function isHiddenStealthedTarget(target: Entity): boolean {
  return target.kind === 'player' && !!target.stealthed;
}
```

which covers duel and arena both, and is simpler. If arena is deliberately out of scope, a one-line comment saying so would save the next reader the same trace.

**note (not blocking)**: the gate only covers auto-attack. The client keeps the stale `targetId` after the opponent stealths, so a targeted ability fired at the hidden opponent still resolves server-side, the same hidden-but-hit window for non-auto swings. The issue is scoped to auto-attack and stealth breaks on the landed hit, so this may well be intended; flagging only so it is a choice and not an oversight.

Humpalumps:

Issue #151, opened 2026-06-14
wow_env.py docstrings describe a stale action/observation space (Discrete(18)/6 slots vs actual 23 actions/10 slots)


jgyy:
## Bug

The docstrings in `python/wow_env.py` (around lines 42-44) describe an out-of-date action space:
- They claim `Discrete(18)` and "6 ability slots".
- The server actually exposes **23 actions** (`ACTIONS.length`) and **10 ability slots** (`src/sim`/`obs.ts`).

The runtime code reads the count dynamically from the `info` payload, so the env itself is correct — but anyone who builds a policy network sized to the docstring (rather than querying `info`) will create a wrong-sized network.

## Impact

Documentation/correctness only — misleads users into mis-sizing networks. **Low severity.**

## Suggested fix

Update the docstrings to match the actual space (23 discrete actions, 10 ability slots), or explicitly tell readers to read the size from `info` rather than the docstring.

_Reported via a Claude Code bug sweep._

---

### Screenshot

The `wow_env.py` docstring describes a stale action/observation space (annotated).

![issue-151](https://raw.githubusercontent.com/jgyy/world-of-claudecraft/assets/issue-screenshots/issue-151.png)

jgyy:
Closing as resolved. The wow_env.py docstring now matches the live env: it documents `Discrete(23)` with movement/turn/strafe/jump, targeting, attack, and 10 ability slots, instead of the stale Discrete(18)/6-slot description. The observation/action sizes are also derived from the env server rather than hard-coded, so the documented space stays in sync. Closing as done.

PR #1098, opened 2026-06-30
fix(python): harden wow env IPC teardown

### LINKED ISSUE #150: wow_env.py: blocking IPC with no read timeout, leaked pipes, and discarded stderr can hang training
## Bug

The Python Gym wrapper `python/wow_env.py` talks to the Node env server over stdio, but the IPC and teardown are not robust.

**`_request` blocks forever (around line 91):**
```python
self._proc.stdout.readline()   # no timeout
```
If the Node server crashes mid-write (e.g. an exception in `sim.tick()` before the JSON line is flushed), the Python side blocks indefinitely instead of detecting the death. Because stderr is sent to `DEVNULL` (around line 79), the underlying cause is also invisible.

**`close()` leaks pipes / can raise (around line 126):**
```python
self._request({"cmd": "close"})
self._proc.wait(timeout=5)      # unguarded -> TimeoutExpired propagates
```
- `stdin`/`stdout` are never closed (no `finally`), leaking FDs.
- `wait(timeout=5)` is unprotected; a stuck child raises `TimeoutExpired` during teardown.
- The `close` handshake itself routes through `_request`, which can block on `readline()` if the child already died.

## Impact

Under vectorized training (`AsyncVectorEnv` with many envs), leaked FDs accumulate, a single crashed env hangs the whole run, and teardown can crash. stderr being discarded makes all of this hard to diagnose. **Medium severity.**

## Suggested fix

- Add a read timeout / liveness check around `readline()` (e.g. poll `self._proc.poll()`, or use a timeout mechanism), and raise a clear error if the child has exited.
- Capture stderr (or surface the last lines on failure) instead of `DEVNULL`.
- In `close()`, wrap teardown in `try/finally`, close `stdin`/`stdout`, and guard `wait(timeout=...)` with a fallback `kill()`.

_Reported via a Claude Code bug sweep._

---

### Screenshot

Blocking IPC, discarded stderr, and leaky teardown in `wow_env.py` (annotated).

issue-150

Humpalumps:
## Summary

Harden the Python `WoWClassicEnv` IPC path so the training wrapper no longer waits forever when the env server stops responding or exits before replying.

- Add a configurable request timeout and background stdout/stderr readers.
- Surface the env server's recent stderr in timeout, exit, and broken-pipe errors.
- Make `close()` idempotent and independent from the request/response path so teardown can kill a stuck child process and close pipes.
- Add Python regression tests for timeout, child-exit, and teardown behavior.

## Related issues

Fixes #150.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `python -m unittest discover -s tests -p "*wow_env_test.py"`
  - `python -m py_compile python\wow_env.py tests\python_wow_env_test.py`
  - `git diff --check`
  - `npm run build:env`
  - `python ..\work\wow_env_smoke.py`
  - `npx vitest run tests\env_protocol.test.ts --config ..\work\vitest.node.config.ts`
  - `npm run check:ts`
  - `npm test -- tests\env_protocol.test.ts` (blocked by existing release-branch `.browserslistrc` parser issue after pretest generation; see note below)
- Manual steps:
  - Ran the Python wrapper against the real built env server bundle; it returned observation shape `(276,)`, `44` available actions, first action `noop`, and closed cleanly.

Known release-branch blocker:

`npm test -- tests\env_protocol.test.ts` runs the pretest generators successfully, then Vitest fails while loading `vite.config.ts` because the release branch parser rejects the `.browserslistrc` comment line `# CSS engine floor (single source) for the Lightning CSS transformer.` The env protocol test itself passes through the direct Node Vitest config command listed above.

## Screenshots / recordings

N/A. This PR changes Python IPC and teardown behavior only; there is no UI/UX change.

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

### DIFF
--- a/python/wow_env.py
+++ b/python/wow_env.py
@@ -20,7 +20,10 @@ from __future__ import annotations
 
 import json
 import os
+import queue
 import subprocess
+import threading
+from collections import deque
 from typing import Any
 
 import numpy as np
@@ -39,10 +42,10 @@ class WoWClassicEnv(gym.Env):
     """Single-agent World of Claudecraft environment.
 
     Observation: float32 vector (self, abilities, target, nearby mobs,
-    nearest interactable, quest states). Action: Discrete(23) —
+    nearest interactable, quest states). Action: Discrete(23) --
     movement/turn/strafe/jump, targeting, attack, 10 ability slots,
     interact, stop, eat/drink. Sizes are content-dependent and queried
-    from the env's `info` cmd at startup — never hardcode them.
+    from the env's `info` cmd at startup -- never hardcode them.
     """
 
     metadata = {"render_modes": []}
@@ -57,6 +60,7 @@ class WoWClassicEnv(gym.Env):
         rewards: dict[str, float] | None = None,
         server_path: str | None = None,
         node_binary: str = "node",
+        request_timeout: float = 5.0,
     ) -> None:
         super().__init__()
         self.player_class = player_class
@@ -74,14 +78,19 @@ class WoWClassicEnv(gym.Env):
             raise FileNotFoundError(
                 f"env server bundle not found at {server}. Run `npm run build:env` first."
             )
+        self._request_timeout = request_timeout
+        self._closed = False
+        self._stdout_lines: queue.Queue[str | None] = queue.Queue()
+        self._stderr_tail: deque[str] = deque(maxlen=40)
         self._proc = subprocess.Popen(
             [node_binary, server],
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
-            stderr=subprocess.DEVNULL,
+            stderr=subprocess.PIPE,
             text=True,
             bufsize=1,
         )
+        self._start_pipe_readers()
         meta = self._request({"cmd": "info"})
         self._obs_size = int(meta["obs_size"])
         self.action_names: list[str] = list(meta["actions"])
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
Nice, tightly scoped hardening. The background reader threads plus the `queue.get` timeout cleanly remove the forever-hang, the stderr tail makes the failure modes legible, and `close()` is now genuinely idempotent and decoupled from the request path. The new tests cover the three behaviors that matter (timeout with stderr tail, pre-write child exit, kill a stuck child on teardown) and they read well. The 1:1 NDJSON contract still holds: `serve()` writes exactly one JSON line per request, and the `console.log` lines in `env_server.ts` are bench-only (`--bench`), never on this wire.

**should-fix**

- A timed-out request leaves the stream desynced. `python/wow_env.py:168` raises `TimeoutError`, but `_write_message` already put the request on the wire, so when the slow server finally replies that line lands in `_stdout_lines`. The next `_request` returns it as its own reply, and from then on every obs/reward is off by one, silently. For a 1:1 stream a single missed reply corrupts everything after it, which is the worst kind of failure in a training loop. This only bites a caller that continues after catching the timeout rather than closing, but that is a plausible path (vector-env auto-reset, a retry wrapper). Suggest marking the env unusable on timeout via a `_broken` flag (kept distinct from `_closed` so `close()` still tears the child down), and failing `_request` fast when it is set.

**nit**

- `_stderr_summary` (`python/wow_env.py:139`) does `"\n".join(self._stderr_tail)` while the stderr reader thread may be appending to the same deque. Iterating a deque while another thread mutates it can raise `RuntimeError: deque mutated during iteration`, turning a clean timeout or exit error into a confusing one on the error path. Snapshot first (`"\n".join(list(self._stderr_tail))`) or guard with a small lock.
- The default `request_timeout=5.0` now also bounds the startup `info` handshake (`python/wow_env.py:94`), which previously waited indefinitely. Spawn plus bundle load plus content init normally finish well under 5s, but on a cold or loaded machine a slow first reply would surface as a `TimeoutError` in `__init__` where it used to just wait. Worth a slightly more generous bound for the initial handshake, or at least a note that startup shares the per-request timeout.

The em dash to `--` swap in the docstring is a fine cleanup (the repo bans em dashes), though a comma would read a touch more naturally.

Merge is clean against `release/v0.18.0`; the UNSTABLE check state is the known release-branch `.browserslistrc` pretest issue you called out, not anything in this change.

Humpalumps:
Closing during cleanup: this duplicates the older open #183 for issue #150, which already covers the IPC timeout/stderr/teardown work and also includes #151. We can revisit with a fresh release/v0.19.0 branch if maintainers want a smaller alternate slice after #183 is resolved.

PR #923, opened 2026-06-24
feat: player economy — native token tips + paid-bodyguard escrow (devnet-tested)

awidearray:
## Player economy: native token tips + on-chain "paid bodyguard" escrow

**Tip** WOC / SOL / USDC to any verified-wallet player straight from the bag, and
**hire** players for an agreed in-game goal — reach level / clear raid / complete
quest / bodyguard-survive / escort — with the reward locked in an on-chain escrow
(`job_escrow` Anchor program) and **auto-released by the authoritative server**
when the goal is met, or refunded to the payer on death / deadline. Non-custodial
throughout: the settler key can only move escrow to the helper or back to the payer.

### What's here
- On-chain `programs/job-escrow` (Anchor): `open` / `release` / `refund`, single-settlement.
- Server settler/oracle client, lifecycle service (state machine + milestone engine),
  per-tick watcher, REST (`/api/jobs/*`), abandoned-quote reconcile, in-game settle
  notifications, and the Hire/Jobs wallet-panel UI.
- Hardening: instance-precise dungeon clears, deposit bound to recorded terms
  (anti-grief), idempotent + retried settlement.
- ~60 feature tests; tsc + build clean.

### Devnet-deployed + verified
Program **`5X39bYGeHPSeNipQGrZRk1siKdgDXXSqhBkpWDTTQzm8`** (devnet). Every tx type
was executed on-chain (open / release / refund + SOL tip + SPL tip, all
Success/Finalized) — Solscan links + screenshots in the first comment.

### Review scope
The branch stacks on the unmerged woc/mounts work it depends on, so the diff is
large. The **new** work is the `feat(jobs)` / `feat(wallet)` commits — focus on
`server/job_*`, `programs/job-escrow/`, `src/net/wallet*.ts`, and the wallet-panel
changes in `src/main.ts`. Full design + ops: `docs/design/player-job-contracts.md`.

## Economy lever, rails status, reviewer checklist

Economy lever: tips and job escrow make $WOC a working in-game currency; every tip and contract is on-chain velocity and a reason to hold a balance. No emissions anywhere in this PR, so the flow ledger invariant is untouched.

Rails: non-custodial throughout (tips are wallet-to-wallet, signed by the player's own wallet; the escrow settler key can only release to the helper or refund the payer). Live money paths are fail-closed and OFF by default behind flags until treasury keys and legal sign off.

- [ ] Branch contains release/v0.22.0 tip; GitHub reports MERGEABLE with zero conflicts.
[...]

PR #1099, opened 2026-06-30
feat(voice-npc): burn $WOC to clone your voice into a quest NPC (draft scaffold)

awidearray:
## Summary
- New money-flow feature: burn $WOC, record a short voice sample, and ElevenLabs voice-clones it into a few lines of dialogue spoken by a small dynamic quest NPC (`voice_echo_npc` / quest `q_echo_of_you`).
- Reuses this repo's existing (not-yet-on-`main`) generic $WOC burn-quote/confirm core (`solana_tx.ts` / `woc_payment.ts` / `woc_config.ts`, ported fresh from the `wt-ad-marketplace` branch's copy) rather than inventing a new payment path.
- Voice cloning/synthesis goes through ElevenLabs (Instant Voice Cloning + TTS), the same provider already used for the game's existing pre-baked NPC voice lines (`scripts/gen_npc_lines.mjs`), via a crash-safe polling keeper modeled directly on `wt-skins`' `server/burn_keeper.ts` (executor/store injection, full recovery on restart).
- The dynamic NPC spawn mirrors the existing `spawnNythraxisAldric` pattern exactly (a static `dynamic: true` NpcDef instantiated at runtime, never surface-placed by the world loader).
- Shipped **off by default** (`VOICE_NPC_ENABLED=false`), same convention as `AD_MARKET_ENABLED`/`SNS_ENABLED`.

## This is a draft / scaffold PR
Per `docs/prd/woc/voice-npc.md` (added in this PR), the following are explicit open items, not yet decided:
- **Burn price** (`WOC_PRICE_VOICE_NPC`, currently a placeholder 25,000) needs tokenomics sign-off.
- **Consent / ToS copy** for voice cloning needs a legal review pass before this can be enabled in production. The sample-upload endpoint already requires an explicit `consent: true` flag and records `consent_at`, but the actual consent/ToS language shown to the player is a stub.
- **Client-side burn-transaction signing is not wired.** This repo has no `@solana/web3.js` dependency yet and no existing client-side transaction builder; rather than hand-roll raw Solana tx bytes, `signAndSendVoiceNpcBurn()` is a clearly-marked stub. The quote/confirm HTTP plumbing and on-chain verification (server side) are complete and tested; only the "client builds and signs the burn tx" step is missing.
- The UI (`src/ui/voice_npc_panel.ts`) is a minimal functional record/consent/quote/confirm/poll flow, explicitly not a polished classic-fidelity HUD window.
- Final zone/quest placement, sample retention/deletion policy, and content moderation on player-submitted display names/audio are follow-ups.

## What's complete and tested
- $WOC burn verification core (ported, tests passing unmodified).
- `voice_npc_grants` lifecycle (`pending_sample -> pending_clone -> cloning/generating -> ready/failed`) with SQL isolated to `voice_npc_db.ts`.
- `VoiceNpcKeeper`: clone -> synthesize -> ready, with failure handling and crash recovery, unit-tested against fakes (no real ElevenLabs calls in tests).
- HTTP routes: `GET /api/voice-npc/info`, `POST /api/voice-npc/sample`, `POST /api/voice-npc/quote`, `POST /api/voice-npc/confirm`, `GET /api/voice-npc/status`.
- One-time, idempotent NPC grant on login via a new `grant_voice_npc` dispatch-only command, claimed atomically server-side so a ready grant is never double-applied.

## Test plan
- [x] `npx vitest run tests/voice_npc_keeper.test.ts tests/woc_payment.test.ts` plus the architecture/i18n/sim/quest/snapshot suites likely to be touched (314 passed, 3 skipped, no regressions)
- [x] `npx @biomejs/biome check` on every changed file (no errors, no format diffs)
- [ ] Manual end-to-end verification once `VOICE_NPC_ENABLED=1` + a real `ELEVENLABS_API_KEY` are available in a dev environment, and the client burn-signing stub is filled in


awidearray:
## Code quality / LARP audit + fix pass (draft scaffold hardening)

Ran an independent, adversarial audit against this branch (not the implementer grading its own work), then fixed every finding, then verified the fixes live in a real browser, not just via tests. Full detail below.

### Audit verdict: no LARP in the money path

The core $WOC burn-verification code (`woc_payment.ts`, `solana_tx.ts`, the atomic login-grant claim) is real, well-tested, and not performative. The one intentionally-stubbed function, `signAndSendVoiceNpcBurn()`, is a clean, honest `throw` (not a fake success), and the UI correctly catches it and resets state rather than getting stuck.

### Findings fixed (`4b4ea0ca`, `41503713`)

| Severity | Finding | Fix |
|---|---|---|
| HIGH | Re-uploading a voice sample while a grant is mid-pipeline (`pending_clone`/`cloning`/`generating`) raced the async keeper: an unconditional `ON CONFLICT DO UPDATE` reset the row regardless of status, so a stale keeper write could clobber a fresh re-upload or vice versa. | Every state transition (`recordConsentAndSample`, `markCloning`, `markGenerating`, `markReady`) is now an atomic compare-and-set (`UPDATE ... WHERE status = <expected>`), mirroring the existing `claimReadyGrantForApply` idiom. A transition that finds the row already moved on is a no-op, not a clobber. New test: "re-upload race: a stale markGenerating write after a re-upload reset is a no-op, not a clobber." |
| HIGH | `buildHttpExecutor` (the actual ElevenLabs HTTP client) had **zero test coverage** - every test only exercised the injected fake, so nothing verified the real request-building/response-parsing code. | Added 5 tests mocking `global.fetch`, exercising the real client for cloneVoice happy path (asserts URL, method, `xi-api-key` header, multipart form fields), 429 rejection, missing-`voice_id` rejection, synthesize happy path, and 500 rejection. |
| MEDIUM | Crash recovery mid-voice-generation silently re-synthesized every line from scratch, wasting ElevenLabs billing on lines already written. | Lines now persist incrementally (`appendGeneratedLine`); resume skips lines already present. Test: "resumes a crash-interrupted generating grant, skipping lines already present." |
| MEDIUM | The burn button displayed "Start Recording" (reused the record button's i18n key) while actually triggering the burn/confirm flow. | Added a dedicated `hudChrome.voiceNpc.burnLabel` key ("Burn $WOC & Confirm"). |
| LOW | A post-payment cleanup failure (`deleteWocQuote`) was silently swallowed with no log. | Now logged via the same `console.error` convention used elsewhere in `server/`. |
| LOW | Status polling retried every 4s forever on a persistently broken endpoint, no cap. | Added a 5-attempt cap with linear backoff and a terminal error status. |

tsc clean, 32/32 tests passing (verified by me independently, not just trusting the fix pass's self-report).

### The one bug neither audit caught, because neither ran the app

Both review passes were code-reading only. So I actually stood up the stack locally (`npm run db:up` / `npm run server` / `npm run dev`), registered a real account, created a character, entered the live world, and clicked the "Voice NPC (WIP)" launcher button.

**The panel was completely invisible.** It existed in the DOM and accessibility tree (`getComputedStyle` reported `display: block; visibility: visible; opacity: 1` with a valid non-zero bounding rect) but nothing rendered on screen. Root cause: the panel had **zero CSS** anywhere in the file, so both the launcher button and the panel rendered as normal-flow `position: static` elements, and the game's `position: fixed` full-viewport WebGL canvas painted over them regardless of DOM order. A `tsc`/vitest pass would never catch this; it only shows up when you actually look at the rendered page.

Fixed in `92792b87` with a small scoped stylesheet (`position: fixed`, `z-index: 100000`, matching the existing top-overlay precedent at `src/styles/base.css:73`). Re-verified live end to end after the fix:
- Launcher button now renders bottom-right, panel opens correctly showing the real server-fetched price (`Price: 25000 $WOC`, confirming `VOICE_NPC_ENABLED=1` and the `/api/voice-npc/info` round trip both work for real)
- The burn button now correctly shows "Burn $WOC & Confirm" distinct from "Record sample" (confirms the label fix above renders correctly, not just compiles)
- Clicked "Burn $WOC & Confirm" with no consent given: real client-side validation fired ("You must consent before recording or uploading a sample"), not a silent no-op
- Filled in a display name, checked consent, clicked "Record sample": a **real** `getUserMedia()` call fired (confirmed the state - typed name and checked consent both survived the re-render, so state management is real, not fake) and correctly surfaced "Could not access your microphone" when headless Chromium had no mic device, rather than hanging or faking success

I offered to attach the 4 screenshots from this session to this comment but my environment blocks publishing files to external hosts and the repo's `.gitignore` rejects committing loose PNGs into `docs/`, so per the user's call this comment is text-only; happy to hand them over directly if useful.

### Honest gaps still open (unchanged from the original PR description, not new)

- Client-side burn-transaction signing is still an intentional stub (no `@solana/web3.js` in this repo yet)
- Burn price, ElevenLabs cost, and consent/ToS copy still need product/legal sign-off
- Sample retention/deletion policy and content moderation on player-submitted names/audio are still follow-ups
- The panel is now visible but still explicitly a WIP floating dialog, not a token-based classic-fidelity HUD window

Branch pushed: `92792b87` on `awidearray:feature/woc-voice-npc`.

awidearray:
### Brought current to release/v0.22.0 tip (head `e762e360`), retargeted, panel localized, honest gap audit below

Status on the current base:

- Base retargeted from main to release/v0.22.0; the branch contains the release tip and GitHub reports MERGEABLE with zero conflicts (the only merge conflicts were generated i18n artifacts, resolved via the sanctioned regen flow).
- `tsc --noEmit` clean. Voice suites 22 of 22 green; the S3 localization guard green; the 31 voice NPC strings now ship real translations in the five non-Latin locales; the panel's WIP banner and error paths are now proper t() keys, and quote and confirm failures report their own errors instead of masquerading as upload failures.
- Money rail unchanged and fail-closed: the burn payment quote ledger is in place, the client burn transaction signing is deliberately unwired on this base (the wallet is signMessage only), and the ElevenLabs keeper stays in console-stub mode unless VOICE_NPC_ENABLED plus a key are set.

Voice NPC panel
The panel on the current base, captured by the new checked-in scripts/voice_npc_shot.mjs (offline boot, real panel module, stubbed API): record a sample, see the quote (Price: 25,000 $WOC placeholder pending tokenomics sign off), burn to clone.

### Why this stays a draft (audited, not asserted)

A self-audit against the running code found two load-bearing gaps in the feature promise, both tracked and neither hidden by the tests:

1. The in-world payoff is not wired yet: the spawned NPC's name override and cloned-voice clip URL are written at spawn but consumed by no renderer, HUD, or wire path, so the chosen name never displays and the cloned voice never plays in-world (the only playback today is the panel's one-shot preview).
2. The NPC is not persistent: the ready grant is claimed once at first login and the dynamic entity is not saved, so the "permanent" NPC vanishes on logout or restart.

Also open before leaving draft: server error strings need stable keys or matcher entries for localization, the placeholder price and line copy need product sign off, and sample retention needs the legal call. The burn signing gap closes when the player-economy wallet work (#923) lands.

PR #1100, opened 2026-06-30
fix(native): require full screen for iOS release

TrevCavill:
## Summary
- add UIRequiresFullScreen to the iOS plist so the landscape-only iPad build passes App Store orientation validation
- extend the native orientation guard test to require the full-screen plist key

## Test
- npx vitest run tests/native_orientation.test.ts

### DIFF
--- a/ios/App/App/Info.plist
+++ b/ios/App/App/Info.plist
@@ -32,6 +32,8 @@
 	<array>
 		<string>armv7</string>
 	</array>
+	<key>UIRequiresFullScreen</key>
+	<true/>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationLandscapeLeft</string>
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
This is the right fix. A landscape-only iPad build fails App Store validation because, without `UIRequiresFullScreen`, Apple requires the app to support all four orientations for multitasking (Slide Over / Split View). Setting it to true opts the full-screen game out of multitasking, which matches the landscape-only intent and clears the check.

Verified:
- The key was absent in the base and appears nowhere else in the tree, so this is net-new, not a duplicate (`ios/App/App/Info.plist:35`).
- It is the only iOS `Info.plist` / orientation-bearing plist, so the single edit covers the whole app target.
- The test asserts the value, not just the key: `expect(plist).toMatch(/<key>UIRequiresFullScreen<\/key>\s*<true\/>/)` would fail on a `<false/>` regression (`tests/native_orientation.test.ts:16`). `npx vitest run tests/native_orientation.test.ts` passes 2/2.

Nothing blocking, no nits. Good to merge.

PR #1111, opened 2026-07-01
fix(ui): reposition market filter dropdowns to stay inside the mobile clip box

### LINKED ISSUE #1051: Unable to scroll the world market UI on mobile
## Problem

On mobile (landscape), the World Market window cannot be scrolled. The window opens at a fixed height, and its inner content overflows below the visible area. There is no way to scroll that overflow with touch, so anything past the first few filter rows is unreachable.

In the Browse tab this means the item listings below the Type / Weapon Type / Rarity filters are cut off and inaccessible, and the same problem applies to longer content in the Sell and Collect tabs. Opening a dropdown (for example Rarity) makes it worse, since the expanded options push the rest of the content further down with no way to reach it.

Effectively the market is unusable on mobile beyond the first screen of content.

## Steps to reproduce

1. Open the game on a mobile device (landscape orientation).
2. Interact with a market vendor to open The World Market window.
3. On the Browse tab, set filters so there are listings to display.
4. Try to scroll the window down to reach the item listings or the lower controls.

## Expected behavior

- The market window content scrolls with touch so the full list of items and all controls are reachable on mobile.
- Scrolling works within the window without moving the game camera or character underneath.
- Opening a dropdown does not trap content off screen; the user can still scroll to everything below it.
- Behavior is consistent across the Browse, Sell, and Collect tabs.

## Actual behavior

- The window content is clipped at a fixed height and does not scroll on touch.
- Item listings and any controls below the filters are unreachable on mobile.
- Expanding a dropdown pushes content out of view with no way to scroll it back.

## Environment

- Platform: mobile web, landscape orientation
- Surface: in-game HUD, The World Market window

## Scope

Please investigate and fix touch scrolling for the World Market window on mobile:

- Browse tab (filters plus item listings)
- Sell tab
- Collect tab
- Interaction between an open dropdown and the scrollable region

## Acceptance criteria

- The World Market window content scrolls with touch on mobile so every item and control is reachable.
- Scrolling the window does not pan the camera or move the character beneath it.
- All three tabs (Browse, Sell, Collect) scroll correctly when their content overflows.
- An open dropdown does not block access to content below it.
- The desktop experience is unchanged.

## Screenshot

Image

jgyy:
## Problem

Fixes #1051.

On mobile (landscape), `#market-window` clips with `overflow: hidden` so
`#market-body` can be a fixed-height touch scroller (the fix from #1058). A
filter dropdown (`.mkt-select-menu`) is `position: absolute` against its
trigger, so it does **not** contribute to `#market-body`'s scrollable area.
When a filter select sits near the bottom of the window, the 236px-tall menu
renders mostly past the window's clip boundary with no scroll path to reach
the clipped part, exactly as reported: opening a dropdown pushes content off
screen with no way to get to it.

## Root cause

```mermaid
flowchart TD
    A["#market-window (overflow: hidden)"] --> B["#market-body (overflow-y: auto, in-flow scroller)"]
    A --> C[".mkt-filters (in-flow, outside #market-body)"]
    C --> D[".mkt-select (position: relative)"]
    D --> E[".mkt-select-menu (position: absolute, 236px tall)"]
    E -. "does NOT extend #market-body's scrollHeight" .-> B
    E -. "clipped by #market-window's overflow: hidden" .-> A
    F["Fix: dropdown_position.ts computeDropdownPlacement()"] -. "measures trigger vs #market-window clip box" .-> E
    F -. "flips menu above trigger + clamps max-height" .-> E
```

## Fix

- New pure module `src/ui/dropdown_position.ts` (`computeDropdownPlacement`):
  given the trigger and clip-box rects, decides whether to open the filter
  menu below or above the trigger (picking the side with more room) and
  clamps its `max-height` to what is actually available, so it always stays
  inside the clip box. Unit tested in `tests/dropdown_position.test.ts`.
- `src/ui/market_window.ts`: both places a filter menu opens (mouse click and
  the keyboard listbox `open` action) now call this on open and apply an
  `open-up` class + inline `max-height`.
- `src/styles/components.css`: `.mkt-select.open-up .mkt-select-menu` flips
  the menu to anchor from the bottom instead of the top.
- Desktop is unaffected: `#market-window` has plenty of vertical room there,
  so `computeDropdownPlacement` keeps picking "below" with the full 236px.

## Before / after

Verified with a Puppeteer repro (900x420 landscape-phone touch viewport,
market flooded with 161 listings, Rarity filter opened): before, only ~37%
of the dropdown menu was visible with no way to scroll to the rest; after,
100% of the menu is visible (flipped above the trigger).

| Before | After |
|---|---|
| before | after |

The base listing-scroll fix from #1058 (`#market-body` touch-scrolls the
item list itself) was already re-verified green on this branch via
`scripts/market_mobile_scroll_shot.mjs`.

## Testing

- `npx tsc --noEmit`
- `npx vitest run tests/dropdown_position.test.ts tests/architecture.test.ts tests/painter_host.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/market_filters.test.ts tests/market.test.ts tests/market_view.test.ts tests/market_window.test.ts tests/save_character_and_market.test.ts` (all green)
- `npx @biomejs/biome check --write` on changed files
- Manual verification via `scripts/market_mobile_scroll_shot.mjs` and a dropdown-clip repro script (Puppeteer, mobile-touch viewport)

Not merging; opening for review per the maintainer's contribution flow.

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -3216,6 +3216,14 @@
       inset 0 1px 0 #ffffff12;
     z-index: 20;
   }
+  /* Flipped by market_window.ts (positionFilterMenu) when there is more clip-box
+     room above the trigger than below it, e.g. a filter near the bottom of
+     #market-window on mobile: the menu would otherwise render past the window's
+     overflow: hidden clip with no way to scroll to the rest of it. */
+  .mkt-select.open-up .mkt-select-menu {
+    top: auto;
+    bottom: calc(100% + 4px);
+  }
   .mkt-select-option {
     width: 100%;
     display: block;
--- a/src/ui/dropdown_position.ts
+++ b/src/ui/dropdown_position.ts
@@ -1,0 +1,31 @@
+// Pure placement math for a custom dropdown menu that must stay inside a
+// clipping ancestor (overflow: hidden), e.g. the World Market filters inside
+// #market-window on mobile. The menu is CSS position: absolute against its
+// trigger, so a fixed 236px-tall menu can render past the clip boundary with
+// no way to scroll to the rest: the clip box's own scroll region does not
+// grow to include an absolutely positioned descendant. Flip the menu above
+// the trigger and/or shrink it to the actually available space instead.
+
+export interface DropdownPlacementInput {
+  triggerTop: number;
+  triggerBottom: number;
+  containerTop: number;
+  containerBottom: number;
+  preferredMaxHeight: number;
+  gap: number;
+  minHeight: number;
+}
+
+export interface DropdownPlacement {
+  side: 'below' | 'above';
+  maxHeight: number;
+}
+
+export function computeDropdownPlacement(input: DropdownPlacementInput): DropdownPlacement {
+  const spaceBelow = input.containerBottom - input.triggerBottom - input.gap;
+  const spaceAbove = input.triggerTop - input.containerTop - input.gap;
+  const side = spaceAbove > spaceBelow ? 'above' : 'below';
+  const space = side === 'below' ? spaceBelow : spaceAbove;
+  const maxHeight = Math.max(input.minHeight, Math.min(input.preferredMaxHeight, space));
+  return { side, maxHeight };
+}
--- a/src/ui/market_window.ts
+++ b/src/ui/market_window.ts
@@ -18,6 +18,7 @@ import type { EquipSlot } from '../sim/types';
 import type { IWorld } from '../world_api';
 import { markDialogRoot } from './dialog_root';
 import { dropdownKeyNav } from './dropdown_nav';
+import { computeDropdownPlacement } from './dropdown_position';
 import { itemDisplayName } from './entity_i18n';
 import { esc } from './esc';
 import { formatMoney as formatLocalizedMoney, formatNumber, t } from './i18n';
@@ -49,6 +50,15 @@ import { svgIcon } from './ui_icons';
 // shared QUALITY_COLOR map carries the real per-quality hex; this token covers a
 // listing whose item has no quality field, so no raw hex lives in the painter.
 const QUALITY_DEFAULT_COLOR = 'var(--color-quality-default)';
+
+// The filter dropdown's natural size (mirrors .mkt-select-menu's max-height/gap in
+// components.css). #market-window clips with overflow: hidden on mobile, and a menu
+// that renders past that clip has no scroll path to the rest of it, so every open
+// recomputes placement against the window's actual clip box instead of assuming
+// there is always room below the trigger.
+const MKT_MENU_PREFERRED_HEIGHT = 236;
+const MKT_MENU_GAP = 4;
+const MKT_MENU_MIN_HEIGHT = 80;
 
 /**
  * Hud-supplied glue. Composes the shared PainterHostPresentation bag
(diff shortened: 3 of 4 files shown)


FernandoX7 (commented):
Clean, well-scoped fix, and the root cause is right: an absolutely positioned menu never grows `#market-body`'s scroll region, so it renders past `#market-window`'s `overflow: hidden` clip with no scroll path to the rest. Good that the placement math lands as a pure `computeDropdownPlacement` module with real unit coverage (below at full height, flip-up when the space below is tighter, the mobile-clip repro, and the min-height clamp), that both open paths (the mouse click and the keyboard listbox `open` action) call it, and that `closeFilterMenus` tears down `open-up` and the inline `max-height` cleanly. The CSS sits in the shared `components.css`, so `/play` picks it up through `main.ts` too. tsc, the dropdown/market tests, and `architecture.test` are green on my end, and `.mkt-select-menu` already has `overflow-y: auto`, so a clamped `max-height` scrolls internally.

A few non-blocking nits:

- nit `src/ui/market_window.ts:59-60`: `MKT_MENU_PREFERRED_HEIGHT = 236` and `MKT_MENU_GAP = 4` duplicate `.mkt-select-menu`'s `max-height: 236px` and the `calc(100% + 4px)` offsets in `components.css`. The comment flags it, so this is just a "keep them in lockstep" note: a future CSS edit to either value silently stales the JS constant.

- nit `src/ui/market_window.ts:235`: the container is measured with `getBoundingClientRect()` (border box), but `#market-window`'s `overflow: hidden` clips at the padding box, which is 2px in because `.panel` carries `border: 2px`. So each side's available space is over-counted by ~2px and the clamped menu can tuck ~2px under the border. Purely cosmetic (the 4px gap absorbs it); leave it, or subtract the border widths if you want it exact.

- nit `src/ui/market_window.ts:232`: `el.querySelector('#market-window') ?? el` always resolves to `el`, since `root()` returns `$('#market-window')` and `querySelector` never matches the element itself. Harmless and reads as defensive, so just worth a word.

Nothing blocking. No sim/wire/server surface, and i18n is untouched.

jgyy:
Thanks for the review. Addressed the two real nits, left the constants one as-is since the comment already flags it:

- `src/ui/market_window.ts:232`: fixed the dead `querySelector('#market-window')` lookup. `deps.root()` already returns `#market-window` itself, so that selector could never match and the fallback to `el` was masking it silently. Removed the lookup and use `el` directly.
- `src/ui/market_window.ts:235`: the clamp now subtracts the panel's actual `border-top`/`border-bottom` width (read via `getComputedStyle`) from the container rect before handing it to `computeDropdownPlacement`, so the placement math matches the padding-box clip instead of over-counting by the border width.
- `src/ui/market_window.ts:59-60`: left the `MKT_MENU_PREFERRED_HEIGHT`/`MKT_MENU_GAP` duplication as-is; the existing comment above them already calls out the lockstep requirement with `.mkt-select-menu` in `components.css`.

`npx tsc --noEmit`, `npx vitest run tests/dropdown_position.test.ts tests/architecture.test.ts`, and `npx @biomejs/biome check src/ui/market_window.ts` are all clean after the change.

jgyy:
Closeout pass on this PR:

- Review nits from @FernandoX7 were already addressed in commit a872d8f0 (dead `#market-window` lookup removed, border-box overcount fixed by subtracting the panel's `border-top`/`border-bottom` before clamping). The `MKT_MENU_PREFERRED_HEIGHT`/`MKT_MENU_GAP` duplication nit was intentionally left as-is per the existing lockstep comment.
- No merge conflicts: branch already contains `upstream/release/v0.18.0` (mergeable, `git merge` reports "Already up to date").
- Verified clean: `npx tsc --noEmit`, the full listed vitest suite (110/110 passing across `tests/dropdown_position.test.ts`, `tests/market_window.test.ts`, `tests/architecture.test.ts`, etc.), and `biome check --changed` against the base branch.
- CI is green: Format + lint (Biome, changed files) pass, PR gate (English-only legal) pass; Release gate (full 14-locale) correctly skipping on a non-release branch.
- Screenshots: skipped, no new UI-visible change in this pass beyond what's already documented in the PR description's before/after (no display/server available in this sandbox to re-capture).

Nothing further to push; PR is ready as-is.

FernandoX7 (approved):

PR #1118, opened 2026-07-01
fix(ui): resync app viewport when opening the options window

### LINKED ISSUE #1082: Keybindings window is cut off in fullscreen and the lowest rows are unreachable

jgyy:
## Problem

Fixes #1082.

The keybindings window (and every other Esc > Options sub-panel, since they all
share `#options-menu`) is a child of `#ui`: a `position: fixed` box sized from
the `--app-vw`/`--app-vh` custom properties and clipped with `overflow: hidden`.

```mermaid
flowchart TD
    A["window resize / fullscreenchange"] -->|"immediate + 250ms + 800ms follow-ups"| B["syncAppViewport() writes --app-vh"]
    B --> C["#ui: position:fixed, height:var(--app-vh), overflow:hidden"]
    D["#options-menu (Key Bindings panel)"] -->|"child of"| C
    D -->|"own overflow-y:auto, max-height: calc(85vh - 24px)"| E["panel content scrolls IF it can be seen"]
    F["player opens Options right after a fullscreen toggle"] -.->|"--app-vh still stale (pre-resync)"| C
    C -.->|"clips at the OLD, shorter height BEFORE the panel's own scroll ever runs"| G["lowest Action Bar rows permanently cut off, no visible scrollbar"]
    style F fill:#502,color:#fff
    style G fill:#502,color:#fff
```

Those custom properties refresh on `resize`/`fullscreenchange`, but only on a
timer (immediate + 250ms + 800ms follow-ups in `main.ts`). Opening the options
window inside that short window right after a fullscreen toggle renders the
panel against a stale, shorter `--app-vh`: `#ui` hard-clips the bottom of the
panel *before* the panel's own `overflow-y: auto` ever gets a chance to scroll
it into view, so there is no visible scrollbar and the lowest Action Bar rows
are unreachable, matching the reported behavior.

## Fix

- Extracted the inline `syncAppViewport()` from `main.ts` into a small pure
  module, `src/game/app_viewport.ts` (host-agnostic, Vitest-tested), reused by
  `main.ts` unchanged.
- `OptionsWindow.toggle()` (`src/ui/options_window.ts`) now calls it
  synchronously right before the panel opens, so the panel always renders
  against the true, current viewport regardless of any resize/fullscreen
  timing race. This covers every sub-panel under `#options-menu` (Key
  Bindings, Performance, etc.), per the issue's stated scope.

## Testing

- `tests/app_viewport.test.ts` (new): unit-tests the extracted
  `syncAppViewport` against a hand-rolled fake `Window` (stable-viewport vs.
  `visualViewport` branches, rounding/floor).
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/app_viewport.test.ts tests/options_window.test.ts tests/mobile_controls.test.ts tests/architecture.test.ts tests/client_shell.test.ts`: all green.
- Manual repro via a puppeteer-core script at the exact reported 1194x905
  viewport: opened Key Bindings, scrolled to the bottom with a real mouse-wheel
  event, confirmed the last Action Bar row (23) is fully reachable and the
  panel's built-in scrollbar works end to end.

**Top of the panel:**

keybindings top

**Scrolled to the bottom, Action Bar 23 fully reachable:**

keybindings scrolled to bottom

## Scope

Per the issue, this is a fix for `#options-menu` sizing/overflow generally
(the shared `.window`/`#options-menu`/`.kb-wide`/`.perf-wide` CSS was already
correct on its own once the panel is rendered against a fresh viewport; the
gap was the stale `--app-vh` race on open). No CSS changes were needed. Not
merging per instructions; ready for review.

### DIFF
--- a/src/game/app_viewport.ts
+++ b/src/game/app_viewport.ts
@@ -1,0 +1,26 @@
+// Keeps the --app-vw/--app-vh custom properties (consumed by #ui's fixed-size,
+// overflow:hidden box) in sync with the true viewport. main.ts wires this to
+// resize/orientation/fullscreenchange; a window that opens on top of #ui can
+// also call it directly so its own size isn't clipped by a stale ancestor box
+// from just before a resize/fullscreen transition settles (see options_window.ts).
+import { useTouchInterface } from './mobile_controls';
+
+export function syncAppViewport(win: Window = window): void {
+  const doc = win.document;
+  const useStableGameViewport =
+    doc.body.classList.contains('game-active') && useTouchInterface(win);
+  const width = Math.max(
+    1,
+    Math.round(
+      useStableGameViewport ? win.innerWidth : (win.visualViewport?.width ?? win.innerWidth),
+    ),
+  );
+  const height = Math.max(
+    1,
+    Math.round(
+      useStableGameViewport ? win.innerHeight : (win.visualViewport?.height ?? win.innerHeight),
+    ),
+  );
+  doc.documentElement.style.setProperty('--app-vw', `${width}px`);
+  doc.documentElement.style.setProperty('--app-vh', `${height}px`);
+}
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,6 +2,7 @@
 // index.html and play.html both bootstrap through this module, so this one import
 // styles both game entries; admin/guide use their own entries and inline CSS.
 import './styles/index.css';
+import { syncAppViewport as syncAppViewportShared } from './game/app_viewport';
 import { audio } from './game/audio';
 import {
   BROWSER_BODY_CLASSES,
@@ -463,26 +464,7 @@ function syncBuildInfo(): void {
 }
 
 function syncAppViewport(): void {
-  const useStableGameViewport =
-    document.body.classList.contains('game-active') && useTouchInterface();
-  const width = Math.max(
-    1,
-    Math.round(
-      useStableGameViewport
-        ? window.innerWidth
-        : (window.visualViewport?.width ?? window.innerWidth),
-    ),
-  );
-  const height = Math.max(
-    1,
-    Math.round(
-      useStableGameViewport
-        ? window.innerHeight
-        : (window.visualViewport?.height ?? window.innerHeight),
-    ),
-  );
-  document.documentElement.style.setProperty('--app-vw', `${width}px`);
-  document.documentElement.style.setProperty('--app-vh', `${height}px`);
+  syncAppViewportShared();
 }
 
 function preventMobileZoom(): void {
--- a/src/ui/options_window.ts
+++ b/src/ui/options_window.ts
@@ -19,6 +19,7 @@
 // setting value only: it never reads the FPS governor and never defines the
 // effects-quality cutoff (that resolver and per-element tiering live elsewhere).
 
+import { syncAppViewport } from '../game/app_viewport';
 import { audio } from '../game/audio';
 import { GAMEPAD_BUTTON_LABELS, GAMEPAD_NONE } from '../game/gamepad_map';
 import {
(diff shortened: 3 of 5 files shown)


FernandoX7 (commented):
Solid fix, and the diagnosis holds up against the code. `#ui` is `position: fixed; height: var(--app-vh); overflow: hidden`, while `.window` sizes off real `vh` (`max-height: calc(85vh - 24px)` in `layout.css`), so a stale, shorter `--app-vh` clips the panel bottom before the panel's own `overflow-y: auto` can ever scroll it. Re-syncing synchronously on open closes that race, and it lands before `render()` and `display: block` so the panel measures against the fresh box. Since the only way any `#options-menu` sub-panel opens is `HudView.toggleOptionsMenu() -> OptionsWindow.toggle()`, the fix covers Key Bindings, Performance, and the rest as the scope claims.

The extraction is clean too: `syncAppViewport(win = window)` is a faithful move of the old body, `useTouchInterface` stays used elsewhere in `main.ts` so there's no dead import, and keeping the local `main.ts` wrapper is the right call since `syncAppViewport` is passed straight to `addEventListener` and the wrapper swallows the `Event` arg that would otherwise land in `win`. Extracted-function tests and CI are green.

One nit:

- nit (`src/ui/options_window.ts:240`): the behavioral half of the fix is untested. `tests/app_viewport.test.ts` covers the pure function well, but nothing pins that `toggle()` actually calls `syncAppViewport()` before the panel opens, which is the real bug fix. `tests/options_window.test.ts` is all source-level string guards (it already pins `captureFocus`/`RANGE_FILL_FULL_PCT` that way), so a one-liner there asserting the toggle path contains `syncAppViewport()` would lock the wiring cheaply.

jgyy:
Addressed the nit: added a source-level test in `tests/options_window.test.ts` ("viewport resync on open (PR #1118)") that pins `OptionsWindow.toggle()` calling `syncAppViewport()` before flipping the panel to `display: block`, so the actual wiring (not just the pure `syncAppViewport()` unit) is locked.

jgyy:
Closeout pass on this branch:

- Reviewer nit (FernandoX7): the PR comment claimed a source-level test pinning `toggle()` calling `syncAppViewport()` before the panel opens had been added, but it was not actually present on the pushed branch. Added it now: `tests/options_window.test.ts` > "options_window: viewport resync on open (PR #1118)" asserts `syncAppViewport()` is called before `root().style.display = 'block'` inside `toggle()`.
- No merge conflicts against `release/v0.18.0` (PR was already mergeable).
- Verified: `npx tsc --noEmit` clean; `npx vitest run tests/options_window.test.ts tests/app_viewport.test.ts tests/mobile_controls.test.ts tests/architecture.test.ts tests/client_shell.test.ts` all green (14/14 in the touched file); `npx @biomejs/biome check --write` on the touched file, no diffs.
- CI on the new commit: Format + lint (Biome, changed files) pass, PR gate (English-only legal) pass, Release gate skipped as expected for a non-release branch.
- Screenshots/mermaid: left as-is from the original PR description (no display/server available in this sandbox to re-capture); no CSS or visual behavior changed in this pass, only test coverage, so the existing before/after screenshots still apply.

Not merging per standing instructions; ready for another look.

FernandoX7 (approved):

Issue #1146, opened 2026-07-01
Professions: World Market integration for gathered and crafted items


jgyy:
## Problem

Materials, crafted items, and tools produced by any of the professions issues need to list, buy, and sell correctly on the existing World Market, with no dupe or data-loss regressions (the market has known historical pitfalls around realm scoping and save timing).

## Expected behavior

- Gathered materials, crafted items, and tools can be listed, bought, and sold on the World Market like any other item.
- No item is lost or duplicated across a market transaction, a realm boundary, or a client leaving before the market's save interval.

## Scope

- Market listing/search: verify profession items surface correctly.
- Regression check against prior market issues (realm scoping, save-on-leave torn writes, unknown-item drop on load).

## Acceptance criteria

- A profession item can be listed, bought, and sold with no quantity or data loss across a realm boundary or an immediate client disconnect after listing.

## Notes

Independent of every other professions issue except that it needs at least one real item type to test against (any single gathered material or crafted item from an earlier issue is enough).

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Dependency note (epic #1152): Depends on: #1122, #1127, #1135. Blocks: none.

jgyy:
Verified implemented on release/v0.23.0: cross-checked acceptance criteria against the merged code and tests. Closing.

PR #1174, opened 2026-07-01
feat(inventory): add additive per-instance item payload (#1165)

### LINKED ISSUE #1165: Professions: item-instance model (signer, per-effect charges, rolled quality/stats, boundTo)
## Problem

`InvSlot` is currently `{itemId, count}` and the World Market keys listings on `itemId`; there is no per-item-instance state. Several downstream professions issues need non-fungible per-item data (a signer identity, per-effect charge counts, rolled quality/stats, and a `boundTo` owner), but no issue currently owns deciding whether to build this or scoping it if so.

## Scope

- A written go/no-go decision: is a per-instance item payload worth the model change (bag/market/trade/save-format churn), or should each dependent feature be re-scoped to avoid it?
- If go: the four-field per-instance payload (signer, per-effect charges, rolled quality/stats, `boundTo`) as an additive extension alongside the existing fungible `{itemId, count}` path, not a replacement of it (most items stay fungible).
- Bag, market listing, and save/load handling for instanced items only where `count` semantics no longer apply.

## Acceptance criteria

- Decision recorded (go or no-go) with rationale.
- If go: instanced items round-trip through save/load, bag display, and (at minimum) are inert in the World Market until #1146 wires them in.
- Explicitly out of scope: #1127's crafting quality roll does NOT require this (it is representable as fungible rarity-keyed itemIds, matching how fishing already works) and should not block on this issue.

## Notes

Genuine prerequisite for #1136 (tool effect slotting needs per-tool durability/charge state) and for the still-unfiled binding/commission-order workflow. #1145's core (self-gathered crafting bonus) does not need this; only its "signed materials" variant does - see the comment on #1145 splitting that out.

Filed per the reconciliation review on #1152 ("Blocking gaps" point 1, "The item-instance model has no issue").

jgyy:
Closes #1165

## Decision

**GO, scoped narrowly.** InvSlot gets one new OPTIONAL field, `instance?: ItemInstancePayload`, additive alongside the existing `{itemId, count}` fungible path. Rationale:

- Several downstream professions issues genuinely need non-fungible per-item state (signer identity, per-effect charge counts, rolled quality/stats, a bound-to owner) that a plain `{itemId, count}` stack cannot represent (two copies of the same itemId can now differ).
- Most items in the game are, and should stay, fungible (stacks of the same itemId are interchangeable): a full replacement of InvSlot with a per-instance model would churn bag/market/trade/save-format code across the whole inventory hub for a need only a minority of items have.
- An ADDITIVE optional payload gets the non-fungible state where it is needed with zero risk to the fungible path: existing slots keep their exact shape (`{itemId, count}`, no `instance` key), and old saved characters with no `instance` field on any slot load and behave identically to before this change.
- The alternative (a full non-fungible instance-id model for every item) was rejected as premature: it is a bigger migration than any currently-filed issue asks for, and it can still be layered in later behind this same optional field if a future issue needs it.

## What changed

- `src/sim/types.ts`: `ItemInstancePayload` (`signer?`, `charges?: Record<string, number>`, `rolled?: {quality?, stats?}`, `boundTo?: number`) and `InvSlot.instance?: ItemInstancePayload`.
- `src/sim/sim.ts`: `addItemInstance(itemId, instance, pid)` grants a single non-fungible copy in its own slot (count 1), never merged with a plain stack. `addItem` now only merges into a slot that has no `instance` (a plain grant never lands inside an instanced slot). New `countFungibleItem`/`removeFungibleItem` mirror `countItem`/`removeItem` but skip instanced slots.
- `src/sim/sim_context.ts`: appended `countFungibleItem`/`removeFungibleItem` to the `SimContext` seam (append-only) and bound them.
- `src/sim/market.ts`: `marketList` now counts/escrows against `countFungibleItem`/`removeFungibleItem` instead of the itemId-wide totals, so an instanced item is inert on the World Market (it is never counted as sellable stock and can never be swept into escrow), while an ordinary stack of the same itemId still lists normally. Real market handling for instanced items is #1146.
- Persistence and bag display needed **no code change**: `serializeCharacter`/`addPlayer` already shallow-clone each `InvSlot` (`meta.inventory.map((i) => ({ ...i }))`), so the optional `instance` field round-trips for free; `IWorld`'s inventory shape and the wire protocol (`server/game.ts` `maybe('inv', meta.inventory)`) already ship the whole slot object as JSON. The bag pure view-core (`buildBagGrid` in `src/ui/bags_view.ts`) is already generic over `InvSlot`, so it renders an instanced slot with no crash; a test proves this directly against the real core.

## Screenshot

No UI change was needed for this issue: the bag display core already handles the additive field generically (proven by a direct unit test against `buildBagGrid`), so there is nothing new to screenshot.

## Diagram

```mermaid
flowchart LR
    subgraph InvSlot shape
        A["InvSlot { itemId, count }"] -->|optional, additive| B["instance?: ItemInstancePayload
        { signer?, charges?, rolled?, boundTo? }"]
    end

    C[addItem] -->|merges only into a slot with no instance| A
    D[addItemInstance] -->|always pushes a NEW slot, count 1| B

    A --> E[serializeCharacter]
    B --> E
    E -->|shallow clone per slot, instance carried as-is| F[(CharacterState.inventory)]
    F --> G[addPlayer / load]
    G --> A
    G --> B

    A --> H[buildBagGrid / bag display]
    B --> H

    A -->|countFungibleItem / removeFungibleItem| I[World Market: marketList]
    B -.->|excluded: inert, never listed or escrowed| I
```

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/item_instance.test.ts tests/architecture.test.ts tests/sim.test.ts tests/market.test.ts tests/sim_context.test.ts tests/entity_roster.test.ts tests/bags_view.test.ts tests/localization_fixes.test.ts` (all green)
- [x] Full `npx vitest run tests/` run: only pre-existing, unrelated failures (missing `@testing-library/svelte` module for the admin Svelte suite, two flaky 5s timeouts in `loot_master_sim.test.ts`); nothing in the inventory/market/sim paths this PR touches regressed
- [x] `npx @biomejs/biome check --write` on every changed file (clean except pre-existing-style non-null-assertion warnings, which do not fail CI)

### DIFF
--- a/src/sim/market.ts
+++ b/src/sim/market.ts
@@ -243,7 +243,10 @@ export class Market {
       return;
     }
     const want = Math.max(1, Math.floor(count));
-    if (this.ctx.countItem(itemId, meta.entityId) < want) {
+    // Per-instance copies (#1165: signer/charges/rolled/boundTo) are inert on the
+    // World Market for now: count and escrow only the fungible stock, so a signed
+    // or bound item is never swept into a listing. (#1146 wires real handling later.)
+    if (this.ctx.countFungibleItem(itemId, meta.entityId) < want) {
       this.ctx.error(meta.entityId, 'You do not have that many to sell.');
       return;
     }
@@ -268,7 +271,7 @@ export class Market {
       );
       return;
     }
-    this.ctx.removeItem(itemId, want, meta.entityId); // escrow
+    this.ctx.removeFungibleItem(itemId, want, meta.entityId); // escrow (fungible-only, #1165)
     this.marketListings.push({
       id: this.nextListingId++,
       sellerKey,
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -278,6 +278,7 @@ import {
   FISHING_CAST_TIME,
   GCD,
   type InvSlot,
+  type ItemInstancePayload,
   isConsuming,
   isPetClass,
   isQuestTurnInNpc,
@@ -1974,6 +1975,7 @@ export class Sim {
       // healingThreat/countItem are bound elsewhere in this host (C4a/C2/C3/Q1) - deduped.
       spendResource: sim.spendResource.bind(sim),
       removeItem: sim.removeItem.bind(sim),
+      removeFungibleItem: sim.removeFungibleItem.bind(sim),
       partyOf: sim.partyOf.bind(sim),
       removeFromParty: (pid: number, verb: string) => sim.party.removeFromParty(pid, verb),
       // dropPartyMarkers flips to the T1 marker store (targeting); lazy arrow since
@@ -1988,6 +1990,7 @@ export class Sim {
       onInventoryChangedForQuests: (meta) => onInventoryChangedForQuests(sim.ctx, meta),
       checkQuestReady: (qp, meta) => checkQuestReady(sim.ctx, qp, meta),
       countItem: sim.countItem.bind(sim),
+      countFungibleItem: sim.countFungibleItem.bind(sim),
       completeQuestForDev: (questId, pid) => completeQuestForDev(sim.ctx, questId, pid),
       completeCurrentQuestsForDev: (pid) => completeCurrentQuestsForDev(sim.ctx, pid),
       // I1 dungeon instancing now lives in instances/dungeons.ts; these route through
@@ -4202,12 +4205,25 @@ export class Sim {
     return n;
   }
 
+  // Fungible-only count for `itemId` (excludes per-instance slots, #1165). The
+  // World Market lists/escrows against this, never the instanced count, so an
+  // instanced copy is never sold as if it were a plain stack member.
+  countFungibleItem(itemId: string, pid?: number): number {
+    const r = this.resolve(pid);
+    if (!r) return 0;
+    let n = 0;
+    for (const s of r.meta.inventory) if (s.itemId === itemId && !s.instance) n += s.count;
+    return n;
+  }
+
   addItem(itemId: string, count: number, pid?: number): void {
     const r = this.resolve(pid);
     if (!r) return;
     const { meta } = r;
     const def = ITEMS[itemId];
-    const existing = meta.inventory.find((s) => s.itemId === itemId);
+    // Never merge into (or split off of) an instanced slot: instanced copies keep
+    // their own signer/charges/rolled/boundTo payload, each in its own slot (#1165).
+    const existing = meta.inventory.find((s) => s.itemId === itemId && !s.instance);
     if (existing) existing.count += count;
     else meta.inventory.push({ itemId, count });
     this.emit({
@@ -4222,6 +4238,23 @@ export class Sim {
     }
   }
 
+  // Grant a single non-fungible copy of `itemId` carrying an instance payload
+  // (#1165: signer/charges/rolled/boundTo). Always its own slot entry (count 1),
+  // never merged with an existing plain or differently-instanced stack.
+  addItemInstance(itemId: string, instance: ItemInstancePayload, pid?: number): void {
+    const r = this.resolve(pid);
+    if (!r) return;
+    const { meta } = r;
+    const def = ITEMS[itemId];
+    meta.inventory.push({ itemId, count: 1, instance });
+    this.emit({
+      type: 'loot',
+      text: `You receive: ${def?.name ?? itemId}.`,
+      pid: meta.entityId,
+    });
+    this.ctx.onInventoryChangedForQuests(meta);
+  }
+
   removeItem(itemId: string, count: number, pid?: number): void {
     const r = this.resolve(pid);
     if (!r) return;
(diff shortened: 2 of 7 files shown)


jgyy:
Closeout check: no review comments or requested changes on this PR, no merge conflicts against `release/v0.18.0` (mergeStateStatus CLEAN), and CI is green (Biome changed-files pass, English-only PR gate pass, full 14-locale release gate correctly skipped for a non-release branch). Nothing actionable to implement; leaving the PR as-is.

FernandoX7 (commented):
Clean, well-scoped foundation. The additive optional field is the right call: fungible stacks keep their exact `{itemId, count}` shape and old saves load identically, while the non-fungible state lives only where it is needed. I checked the load-bearing claims against the code and reran the tests at the head:

- Parity holds with no wire change: `meta.inventory` ships as a whole-slot JSON array (`server/game.ts:3118`) and the client assigns it whole (`src/net/online.ts:1426`), so `instance` round-trips with no per-field encode to drift. Persistence is the same story through the existing shallow-clone in `serializeCharacter`/`addPlayer`.
- The market split is complete: both the count check (`market.ts:249`) and the escrow (`market.ts:274`) use the fungible-only variants, and the buy/delist/collect paths hand goods back as plain stacks.
- Determinism and sim purity are intact, and the new `SimContext` callbacks are declared, bound in both `createSimContext` and the `Sim` ctx, and mirrored in both mock hosts.
- i18n is fine as-is: the `addItemInstance` loot line is caught by the existing `/^You receive: (.+)\.$/` arm in `hud.ts`, so the S3 drift guard stays green (confirmed by running it).

`tsc --noEmit`, `item_instance`, `sim_context`, `entity_roster`, `market`, and `localization_fixes` all pass at the head.

Two forward-looking notes for whoever wires the real instanced items. Neither is a blocker here, since nothing grants an instance yet:

- nit: the `instance` object is shared by reference, not deep-copied. `serializeCharacter` (`sim.ts:1404`) and the load path (`sim.ts:1223`) both clone slots with `{ ...i }`, so the nested `instance` (and its mutable `charges`/`rolled.stats` maps) is aliased between the live slot and the serialized/loaded state. Harmless today because nothing mutates `instance` after creation, but `charges` is by design a mutable counter: the change that decrements a charge will need a deep copy at those two boundaries, or it will silently mutate a save snapshot (or a shared source state on load).
- nit: only the market is instance-aware. `countItem`/`removeItem` still treat an instanced copy as fungible, so quest collection credit, reagent use, vendor sell, and discard would count or consume a signed or bound copy once instances exist. Deferring the real handling to #1146 is reasonable; just flagging that the market is not the only consumption path that will want the fungible split. (The client-side optimistic add at `online.ts:1727` merges by itemId too, but self-corrects on the next snapshot.)

Good coverage on the parts that matter: the save/load round-trip, the no-merge-into-an-instanced-slot invariant, the bag view rendering an instanced slot, and the market staying inert are all asserted directly.

FernandoX7 (commented):
The deep-clone fix in `eca0c0e16` resolves the aliasing nit cleanly. `cloneInvSlot` (`types.ts:402`) deep-copies both mutable nested maps (`charges` and `rolled.stats`), and it is now applied at both boundaries I flagged, for inventory and vendorBuyback (save at `sim.ts:1405`, load at `sim.ts:1224`). The new test asserts both directions: mutating a serialized snapshot does not touch the live slot, and mutating a loaded copy does not reach back into the saved state, covering `charges` and `rolled.stats`.

I checked the other InvSlot shallow-clone sites while I was in there and they are all fine: the market collection copies and lockpick `pendingLoot` never carry an instance, `sanitizeRemovedZone1Content` always sits next to either `cloneInvSlot` on load or a JSON write on save so nothing live is aliased, and the client optimistic add is a mirror the next snapshot overwrites.

Reran at the head: `tsc --noEmit`, `item_instance`, `architecture`, `sim`, `market`, `sim_context`, `entity_roster`, and `localization_fixes` all green.

The other note (making `countItem`/`removeItem` instance-aware for quests, reagents, vendor, and discard) is untouched, which is the right call: that is #1146's job and nothing grants an instance yet. Looks good to me.

jgyy:
Implemented FernandoX7's review nit: added `cloneInvSlot` (src/sim/types.ts) which deep-clones the `instance` payload (`charges` and `rolled.stats`) instead of the old shallow `{ ...i }`, and used it at all four inventory/vendorBuyback clone sites in `serializeCharacter` and load (`sim.ts`). Added a regression test proving a mutation on a saved snapshot or a loaded copy no longer aliases the live slot's state.

(The second nit, non-fungible items being counted as fungible outside the market, is correctly deferred to #1146 as noted in the review.)

CI is green: `tsc --noEmit`, Biome (changed files), the English-only PR gate, screenshots, and AI review all pass.

FernandoX7 (approved):

Issue #1121, opened 2026-07-01
Professions: per-player node harvest and respawn timer


jgyy:
## Problem

Harvesting a node and each player having their own respawn timer (so one player never denies a node to another) is not implemented. Builds on the node content in a prior issue but is a self-contained mechanic: it can be tested against a single hardcoded node if the content issue has not landed yet.

## Expected behavior

- A player can harvest a node they are near, receiving a material item.
- Each player has an independent respawn timer per node; one player harvesting does not affect another player's ability to harvest the same node.
- There is no gather rush or node camping as a result.

## Scope

- `src/sim/professions/gathering.ts`: per-player respawn state, harvest resolution.
- `IWorld`: node harvestable-by-me read surface.

## Acceptance criteria

- Two players harvesting the same node each get their own respawn timer.
- A node's respawn state is per-player, not global.
- Harvest resolves server-side on the 20 Hz tick.

## Notes

Depends conceptually on node placements existing (G2) but can be implemented against a minimal in-file node stub and wired to real content later without an API change.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
## Per-player respawn sequence

```mermaid
sequenceDiagram
    participant A as Player A
    participant B as Player B
    participant N as Node
    A->>N: Harvest
    N-->>A: Material granted, A start own respawn timer
    B->>N: Harvest (immediately after A)
    N-->>B: Material granted, B start own independent respawn timer
    Note over N: Node has no single shared cooldown; each player tracks their own
```

jgyy:
Dependency note (epic #1152): Depends on: #1119, #1120. Blocks: #1122, #1124.

jgyy:
Implemented and merged via #1183 (src/sim/professions/gathering.ts (per-player nodeHarvestReadyAt respawn)). Verified present on main (release/v0.22.0 merge).

jgyy:
Implemented and merged via #1183 (src/sim/professions/gathering.ts, per-player nodeHarvestReadyAt respawn). Verified present on main (release/v0.22.0 merge).

PR #1175, opened 2026-07-01
feat(content): add gatherable node placements and render props

### LINKED ISSUE #1120: Professions: world-spawn node content and placement (ore, wood, herb)
## Problem

There is no in-world content for gatherable nodes (ore veins, trees, herb patches). This issue is pure content plus a static, unowned world object: it does not require player-side harvest logic to exist yet.

## Expected behavior

- Ore, wood, and herb nodes exist as placed, visible world objects in relevant zones.
- Nodes are permanent fixtures; they are not consumed or removed by being placed.

## Scope

- `src/sim/content/gather_nodes.ts`: node type definitions and zone placements (new content file, merged by `data.ts`).
- Render: minimal node prop/marker so a node is visible in the world.

## Acceptance criteria

- Nodes appear at their defined zone placements and are visible to a client in range.
- Adding a node type or placement requires no changes outside `gather_nodes.ts` and the render prop lookup.

## Notes

Deliberately scoped to content and visibility only; harvesting a node (G3) is a separate issue that consumes this data but this issue is useful and testable on its own (nodes render even before they can be harvested).

jgyy:
## Summary

Adds in-world gather node content: ore, wood, and herb nodes as permanent, unowned world fixtures. This is content and visibility only, no harvest logic (that is #1121).

- `src/sim/content/gather_nodes.ts`: 15 placements (5 ore, 5 wood, 5 herb) across Eastbrook Vale and Mirefen Marsh, merged into `sim/data.ts`'s `GATHER_NODES` the same way other content files are merged.
- `src/render/gather_nodes_lookup.ts`: pure node-type to color/offset lookup (no three.js), so a Vitest can assert coverage without a browser.
- `src/render/gather_nodes.ts`: draws each node as a simple procedural primitive (icosahedron for ore, cone for wood, box for herb), no new textures.
- `tests/gather_nodes.test.ts`: asserts every node references a real zone id, every node type is known, ids are unique, and the render lookup covers every type used in content.

Closes #1120

```mermaid
flowchart LR
    A[gather_nodes.ts<br/>content + placements] --> B[data.ts<br/>GATHER_NODES merge]
    B --> C[gather_nodes.ts<br/>render prop lookup]
    C --> D[visible in zone<br/>Eastbrook Vale / Mirefen Marsh]
```

Adding a new node type or placement only requires touching `gather_nodes.ts` (content) and the render lookup table.

## Screenshot

Ore node (grey rock, low-poly icosahedron) visible near Boar Meadow, Eastbrook Vale:

gather node ore

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/gather_nodes.test.ts tests/architecture.test.ts`
- [x] `npm run build`
- [x] Biome format on changed files only
- [x] Real screenshot captured via `scripts/gather_nodes_shot.mjs` against `npm run dev`


jgyy:
Closeout pass complete: no review comments or requested changes on this PR. Branch is up to date with base (release/v0.18.0), MERGEABLE with no conflicts. Verified locally: `npx tsc --noEmit` clean, `npx vitest run tests/gather_nodes.test.ts tests/architecture.test.ts` 28/28 passing. CI checks are green (Format + lint, PR gate English-only legal); release-gate full-locale check is a skip as expected for a non-release PR. Screenshot referenced in the description already reflects the final implementation (added in the follow-up commit that sorted the screenshot script imports), so no refresh needed. No changes pushed.

FernandoX7 (commented):
Clean, well-scoped foundation: content plus visibility only, and it lands on the right seams. Splitting the pure `gather_nodes_lookup.ts` out of the three.js `gather_nodes.ts` so a Vitest can assert type coverage without a browser is exactly the pure-core pattern this repo wants, and the draw code is its own render module rather than another method bank on `renderer.ts`. Sim purity holds (the content file imports only `types`), the `Record<GatherNodeType, ...>` tables are type-enforced so `tsc` catches a missing mapping, and there is a real screenshot. Merges cleanly into `release/v0.21.0` and `tests/gather_nodes.test.ts` passes locally.

A few nits, nothing blocking:

- nit: the static node group is not matrix-frozen like its sibling. `renderer.ts:1250` adds the group right after the props group, which calls `freezeStaticMatrices(props.group)` because it is baked into world space and never moves. Gather nodes are the same case (positions set once in `buildGatherNodes`, no per-frame `update()`), so they keep `matrixAutoUpdate` on and recompute world matrices every frame. Trivial at 15 meshes, but freezing the group would match the pattern sitting right next to it.

- nit: the zone-id test checks membership, not placement. `tests/gather_nodes.test.ts:9` asserts `zoneId` is a known zone, but `zoneId` is otherwise decorative here (the render module places by absolute `pos` and ignores `zoneId`), so a node could claim `eastbrook_vale` while sitting at z=400 in Mirefen and nothing fails. Zones expose `zMin`/`zMax`; asserting `node.pos.z` falls within its zone's band makes the field load-bearing and catches copy-paste drift. I checked all 15 current placements and they are in-band, so this is preventive, not a live bug.

- nit: `NODE_GEOMETRY_KEYS` (`gather_nodes_lookup.ts:7`) is a third hand-maintained copy of the `GatherNodeType` union, alongside `GATHER_NODE_TYPES` and the type-enforced `NODE_COLOR` / `NODE_Y_OFFSET` / `NODE_GEOMETRY` Record keys. `tsc` already forces those Records complete, so the test could assert against `Object.keys(NODE_COLOR)` (or reuse `GATHER_NODE_TYPES`) and drop a list to keep in sync. Minor.

jgyy:
Thanks for the review, all three nits addressed:

- Froze the gather-nodes group with `freezeStaticMatrices` right after adding it in `renderer.ts`, matching the props group right above it (no per-frame `update()`, positions set once).
- `tests/gather_nodes.test.ts` now asserts each node's `pos.z` falls within its claimed zone's `zMin`/`zMax` band, so a zone/placement mismatch fails instead of only checking zone-id membership.
- Dropped `NODE_GEOMETRY_KEYS` from `gather_nodes_lookup.ts`; the coverage test now derives the covered-types set from `Object.keys(NODE_COLOR)` (already `tsc`-enforced complete), so there's one fewer hand-maintained copy of the type union.

No visual or content change, so the existing screenshot and mermaid still reflect the implementation. Verified locally: `npx tsc --noEmit` clean, `npx vitest run tests/gather_nodes.test.ts tests/architecture.test.ts` 29/29 passing (was 28, +1 for the new zone-band test), pushed as b0ac0064d.

FernandoX7 (approved):

PR #1177, opened 2026-07-01
feat(professions): add per-player gathering profession proficiency tracking

jgyy:
Closes #1119

## Summary

Adds the state foundation for the gathering professions (Mining, Logging, Herbalism): an independent, additive proficiency counter per profession per player. No world nodes exist yet (that is later gathering work); this ships the counter itself plus a dev-only grant command so it is independently testable.

```mermaid
flowchart LR
    A["/dev gather professionId amount (chat cheat, devCommands only)"] --> B[queueGatheringGrant]
    B --> C{{pendingGatherGrants queue on PlayerMeta}}
    C -->|drained once per player, per sim.tick| D[drainGatheringGrants]
    D --> M[Mining counter]
    D --> L[Logging counter]
    D --> H[Herbalism counter]
    M --> P[(CharacterState.professions, JSONB)]
    L --> P
    H --> P
    P -->|addPlayer / normalizeGatheringProficiency| Q[reload on login]
    P -->|professionsStateFor / IWorldProfessions.professionsState| R[["prof" wire delta -> ClientWorld]]
```

Mining, Logging, and Herbalism are shown above as three parallel, independent counters: granting one never touches the others (no shared/conserved pool).

## What changed

- `src/sim/content/professions.ts`: the starter gathering-profession content table (data-as-code, exempt from module-first per root CLAUDE.md). Each def extends the settled `ProfessionRecord {id, category, maxSkill}` shape (`src/sim/professions/types.ts`, from #1164) with the display metadata (`name`/`icon`/`description`) the `/dev gather` cheat and a future UI need. `icon` is a plain identifier, not an emoji glyph, per the repo's no-emoji copy rule. `maxSkill: 300` on the classic 1-300 profession skill scale.
- `src/sim/professions/gathering.ts`: proficiency state helpers + gain logic behind the `SimContext` seam. A grant is *queued* (`queueGatheringGrant`) and *drained* once per player per tick (`drainGatheringGrants`, called from `sim.ts` `tick()` right next to `updateRested`), so a grant only ever takes effect on the deterministic 20 Hz tick path, never out of band. A non-positive amount is now rejected at queue time (proficiency is additive-only, no decrement path). `gatheringSkillsView()` projects the internal counter onto the settled `PlayerProfessionSkill[]` shape.
[...]

Issue #1119, opened 2026-07-01
Professions: gathering profession skill tracking (Mining, Logging, Herbalism)


jgyy:
## Problem

Per-player gathering proficiency does not exist. This is the smallest independently buildable piece of the gathering foundation: a proficiency counter per gathering profession, with no world nodes required to exercise it (a debug/dev-command grant is enough to test).

## Expected behavior

- A player has an independent proficiency value for each gathering profession (starter set: Mining, Logging, Herbalism).
- Gaining proficiency in one profession never changes proficiency in another (additive, not conserved).
- Proficiency persists on the character record.

## Scope

- `src/sim/content/professions.ts` or a new `gathering.ts` content file: profession definitions.
- `src/sim/professions/gathering.ts`: proficiency state and gain, behind the `SimContext` seam.
- `IWorld`: read surface for a player's gathering proficiencies.
- Persistence: per-player proficiency, additive back-compat JSONB.

## Acceptance criteria

- Granting proficiency in one profession leaves all others unchanged.
- Proficiency values persist across logout/login.
- All gain logic lives in `src/sim/`, uses `Rng` where randomness is involved, and runs on the 20 Hz tick.

## Notes

Self-contained: does not require world nodes, tools, or the crafting wheel to land or to be tested. Later gathering issues (world nodes, respawn, rarity rolls) build on this state but this issue has no dependency on them.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Dependency note (epic #1152): Depends on: none (Tier 0, foundation). Blocks: #1121, #1122, #1123, #1124.

jgyy:
Dependency update: this issue's skill-tracking persistence and read surface should build against the shared contract now defined in #1164 (`src/sim/professions/types.ts`, the `IWorld` professions facet, and the single professions JSONB/wire key), rather than inventing its own, per #1152's reconciliation review.

jgyy:
Closed by #1177, merged via release/v0.20.0 -> main (#1423).

Issue #1164, opened 2026-07-01
Professions: contracts and IWorld facet (skill/craft/recipe/node schemas, wire/persistence keys)


jgyy:
## Problem

No issue currently defines the shared professions schemas (skill, craft, recipe, node) or the `IWorld` read-surface/wire/persistence keys before content and mechanics land. #1119 and #1126 each currently plan to invent their own JSONB persistence contract independently, which will conflict once both are in flight, and the same risk applies to #1120, #1125, #1140 defining node/craft records ad hoc.

## Scope

- Shared TypeScript types for skill, craft, recipe, and node records (a `src/sim/professions/types.ts` or similar), content-as-code friendly per `src/sim/content/` conventions.
- Extend `IWorld` (`src/world_api.ts`) with a professions read-surface facet, analogous to `IWorldDelves`.
- Implement the facet (even as an empty/stub read surface) in both `Sim` and `ClientWorld` to keep parity from day one.
- Settle the wire/snapshot key names and the persistence JSONB key(s) that #1119, #1120, #1125, #1126, #1140 will all build against, instead of each inventing its own.

## Acceptance criteria

- `IWorld` professions facet is merged and implemented in both `Sim` and `ClientWorld` (stub/empty return is acceptable at this stage).
- Shared skill/craft/recipe/node types exist and are imported by, not duplicated in, downstream content and mechanic issues.
- Persistence key and wire/snapshot key names are documented in this issue's scope for #1119/#1120/#1125/#1126/#1140 to build against.

## Notes

Blocks #1119, #1120, #1125, #1126, #1140 semantically (they should build on these contracts rather than invent their own), though those issues can still be stubbed independently for parallel work per the epic's independence-via-stubbing convention.

Filed per the reconciliation review on #1152 (see that issue's "Blocking gaps" point 3, "No contracts-first foundation issue").

Full design: https://woc.nervemart.com/docs/professions-system

Nervescraper:
[...]

PR #1178, opened 2026-07-01
feat(professions): add vendor-sold gathering tools with tier gating

### LINKED ISSUE #1123: Professions: vendor-sold base gathering tools with tier gating
## Problem

There is no purchasable gathering tool, and no tier gate limiting which nodes/materials a player's tool can access. This issue is scoped to the base tool only (no slotted effects, that is a separate slice).

## Expected behavior

- Basic base gathering tools are purchasable from vendors.
- A base tool has a tier; a player's tool tier gates which node/material tiers they can gather.
- A base tool has infinite durability.

## Scope

- Base tool item records (existing item content system).
- `src/sim/professions/tools.ts`: tier gating check.
- Vendor content: add base tools to relevant vendor stock.

## Acceptance criteria

- A player without a high-enough-tier tool cannot harvest a node/material above their tool's tier.
- A base tool never becomes unusable from durability loss.

## Notes

Independent of effects/recharge (separate tools-slice issues) and independent of the wheel; only needs an item + a gate check.

jgyy:
## Summary
Closes #1123.

Adds purchasable base gathering tools (Mining, Logging, Herbalism) at three tiers
each, and a pure tier-gate check that limits which node or material tier a tool
can gather.

- `src/sim/content/items.ts`: nine new base tool items (`copper_mining_pick` /
  `iron_mining_pick` / `mithril_mining_pick`, `handaxe` / `felling_axe` /
  `ironbark_axe`, `gathering_sickle` / `bronze_sickle` / `silverleaf_sickle`),
  each `kind: 'tool'` with `use: { type: 'gatherTool', professionId, tier }`.
- `src/sim/content/zone1.ts`: all nine added to Trader Wilkes's vendor stock.
- `src/sim/professions/tools.ts` (new): `canGatherTier(playerToolTier, nodeTier)`,
  true only when the player's tool tier is at least the node/material tier, plus
  `gatherToolTier`/`isGatherToolUse` helpers for reading the new item shape.
- `src/sim/types.ts`: extended `ItemUse` with the `gatherTool` variant.
- `tests/professions_tools.test.ts` (new): tier-gate cases (tier 1 cannot gather
  tier 2 or 3, tier 2 gathers 1 and 2 but not 3, tier 3 gathers everything),
  vendor tool presence per profession, and a durability check (there is no
  durability field anywhere on `ItemDef` in this repo, so a base tool cannot
  become unusable; the test asserts the item shape carries no `durability` key
  across repeated simulated gathers).

Scoped per the issue: base tool only, no slotted effects. There are no world
gathering nodes yet (see #1119, gathering proficiency), so `canGatherTier` is a
standalone pure comparison a future node/material system will call with its own
tier value; it is not yet wired into a live gather action.

```mermaid
graph LR
  A[Vendor stock<br/>Trader Wilkes] --> B[Tool item<br/>use.tier N]
  B --> C{canGatherTier<br/>playerToolTier &gt;= nodeTier}
  D[Node / material<br/>tier] --> C
  C -->|true| E[Gather allowed]
  C -->|false| F[Gather blocked]
```

Stacked on #1177 (gathering proficiency); diff will shrink once that merges; do not merge this before it.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/professions_tools.test.ts tests/professions_gathering.test.ts tests/architecture.test.ts`
- [x] `npx vitest run tests/vendor_stack.test.ts tests/vendor_view.test.ts tests/items.test.ts`
- [x] `npx vitest run tests/localization_fixes.test.ts tests/guide.test.ts`
- [x] `npm run build`

## Screenshot
Trader Wilkes's Goods window showing the nine new base gathering tools purchasable at their tiers:

Vendor gathering tools


FernandoX7 (commented):
Nicely scoped. `canGatherTier` lands as a genuinely pure leaf module (`src/sim/professions/tools.ts`): no SimContext state, Vitest-imported directly, which is exactly the module-first shape this repo wants, and the nine tools are clean data-as-code with sensible tiered buy/sell values. tsc and the tool/vendor/items suites pass locally. Two things before this can merge, plus a nit.

## Blocking

- `src/ui/i18n.resolved.sha256` is malformed: it has five hash lines, not one. `scripts/i18n_resolved_hash.mjs --write` writes a single digest and `--check` does `readFileSync(...).trim()` against one 64-char hash, so a five-line file can never match. Locally `node scripts/i18n_resolved_hash.mjs --check` fails (computed `c89a169c...` vs the stale first-line baseline `a1ada6b3...`) and `tests/i18n_resolved_equivalence.test.ts` fails 2/5. This is what has the PR sitting UNSTABLE. It looks like a rebase/merge stacked several one-line versions instead of overwriting. Fix: merge the base, re-run the i18n build, then `node scripts/i18n_resolved_hash.mjs --write` and commit the single-line result (`c89a169c...`).

## Should-fix

- `src/sim/sim.ts:794`: a dead duplicate persisted field. `CharacterState` already has `professions?` right above it (the settled #1164 key), and that field's own comment says the key is `professions`, "not `gatheringProficiency`". The extra `gatheringProficiency?: Partial<...>` here is never read or written (load reads `s.professions` at ~1349, save writes `professions` at ~1552), and it trails an orphaned comment fragment (`// load cleanly, defaulting every profession to 0).`). The base and the merged #1177 both have only `professions?`, so this is the rebase leak your final commit ("drop redundant gatheringProficiency stub") set out to remove but missed in sim.ts. Drop the field and the stray comment line.

## Nit

- `tests/professions_tools.test.ts`: the "vendor-sold base tools exist" case asserts the items exist in `ITEMS` and read the right tier, but nothing asserts they are actually in Trader Wilkes's `sells` stock (the `zone1.ts` change). A future edit dropping them from the vendor would still pass. Worth pinning the stock membership, since that is the behavioral change here.

For context on scope: the base already carries the #1119 proficiency work, so the real delta of this PR is the nine tools, the `gatherTool` `ItemUse` variant, `tools.ts`, and the i18n. `kind: 'tool'` and the `gatherTool` use fall through `useItem` to a safe no-op (not consumed, no crash), which is correct until the gather-node system lands.

jgyy:
Thanks for the review, this is fixed.

## Fixed

- **Blocking (i18n hash):** already resolved by `dad7220e7 chore(i18n): rewrite resolved-table sha256 baseline` on this branch before your review landed. `node scripts/i18n_resolved_hash.mjs --check` now passes (`c89a169c...`, single line), and `tests/i18n_resolved_equivalence.test.ts` is green.
- **Should-fix (dead `gatheringProficiency` field):** dropped the stray field and its orphaned comment line from `CharacterState` in `src/sim/sim.ts`. `professions` is the only persisted key now, matching load (`s.professions`) and save (`professions: {...}`).
- **Nit (vendor stock pinning):** added a case to `tests/professions_tools.test.ts` asserting all nine tools are actually in `NPCS.trader_wilkes.vendorItems`, not just present in `ITEMS`.

```mermaid
flowchart LR
  subgraph before["Before"]
    A["CharacterState.professions?"] --> L1["load: s.professions"]
    B["CharacterState.gatheringProficiency? (dead)"] -.->|never read/written| X["orphaned"]
  end
  subgraph after["After"]
    C["CharacterState.professions?"] --> L2["load: s.professions"]
    L2 --> S["save: professions: {...meta.gatheringProficiency}"]
  end
  subgraph test["New coverage"]
    T1["ITEMS has the 9 tools"] --> T2["NPCS.trader_wilkes.vendorItems contains them"]
  end
```

Verified locally: `tsc --noEmit`, `node scripts/i18n_resolved_hash.mjs --check`, and `npx vitest run tests/professions_tools.test.ts tests/i18n_resolved_equivalence.test.ts tests/sim.test.ts tests/architecture.test.ts` all pass (128/128).

Screenshot confirming the vendor-stock claim end to end (Trader Wilkes selling all nine gathering tools):

Trader Wilkes vendor window with the nine gathering tools

Pushed as `581868d0` on this branch.

FernandoX7 (commented):
All three landed cleanly, verified against the head:

- The sha256 baseline is a single line again (`c89a169c...`) and `node scripts/i18n_resolved_hash.mjs --check` prints OK, so `i18n_resolved_equivalence` passes.
- The dead `gatheringProficiency` field is gone; `sim.ts` is back to base (no longer in the diff), leaving only the settled `professions?` key.
- The new "stocked by Trader Wilkes" case pins all nine ids against `NPCS.trader_wilkes.vendorItems`, which is the behavioral change worth locking.

tsc and the tool/vendor/items/i18n suites are green locally, the net diff is now the clean tools-only delta, and I re-checked the rebase for re-injected drift (that was how the dead field slipped in the first time) and found none. The remaining UNSTABLE is just the PR gate still running.

Three optional test-polish notes, all in `tests/professions_tools.test.ts`, none blocking:

- nit: the "no durability" case is theatrical. `gatherToolTier` is a pure read that never mutates the item, so the 1000-iteration loop is 1000 identical assertions, and `not.toHaveProperty('durability')` (twice) checks a field the repo never has. The only load-bearing line in that block is `isGatherToolUse(pick.use)`. Worth trimming to that plus one `gatherToolTier` call.
- nit: `gatherToolTier`'s "item has a `use` but is not a gatherTool" branch is untested. `worn_sword` has no `use` (short-circuits the first clause) and the profession-mismatch case still returns true from `isGatherToolUse`, so the middle predicate `!isGatherToolUse(item.use)` and that function's `false` return are never pinned. Passing `simple_fishing_pole` (`kind: 'tool'`, `use: { type: 'fishing' }`) exercises exactly that path.
- nit: the `gatherTool` use is an unpinned no-op. Using a mining pick falls through `useItem` (src/sim/items.ts) to a silent no-op: not consumed, no error, no crash, which is correct until the gather system lands. A `useItem(copper_mining_pick)` case asserting the count is unchanged and no error is emitted would protect that contract against a future change that makes an unrecognized tool throw or get consumed.

Good to merge once the gate goes green; the nits are polish for a follow-up if you want them.

jgyy:
Thanks for the polish pass, all three landed:

- Trimmed the "no durability" case to the one load-bearing assertion (`isGatherToolUse` + a single `gatherToolTier` call), dropped the 1000-iteration loop and the two `not.toHaveProperty('durability')` checks.
- Added `simple_fishing_pole` (`kind: 'tool'`, non-`gatherTool` `use`) to the mismatch test, pinning the `!isGatherToolUse(item.use)` branch of `gatherToolTier` specifically.
- Added a case that calls `useItem('copper_mining_pick', pid)` through a real `Sim` and asserts it doesn't throw and doesn't consume the tool, locking in the silent-no-op contract until the gather-node system lands.

`npx tsc --noEmit` and `npx vitest run tests/professions_tools.test.ts tests/i18n_resolved_equivalence.test.ts tests/sim.test.ts tests/architecture.test.ts` all pass (129/129). Pushed as `4cba920d` on this branch.

FernandoX7 (approved):

PR #1171, opened 2026-07-01
feat(content): add ten-craft ring order and pole content (#1125)

jgyy:
Closes #1125

## Summary
- Adds `src/sim/content/professions.ts`: the ten production crafts (Armorcrafting, Weaponcrafting, Tailoring, Leatherworking, Alchemy, Cooking, Inscription, Enchanting, Engineering, Jewelcrafting) in a fixed ring order (`CRAFT_RING`), each tagged with a pole (Material, Experimental, Formal, Cross-cutting).
- Adds `adjacentCrafts(craftId)` and `oppositeCraft(craftId)` pure helper functions derived purely from ring position (adjacent = one position away either side, opposite = five positions away).
- Pure data plus pure functions only, no mechanic attached, no dependency on any other professions issue, per the issue's scope and acceptance criteria.

## Design doc reachability
`https://woc.nervemart.com/docs/professions-system` returned HTTP 403 Forbidden when fetched during this change, so I could not confirm the canonical ring order/pole assignment against the design doc. The ring below is my best-effort placeholder, documented in a comment at the top of `professions.ts`:

- A 10-craft circle in a fixed array order.
- Opposite = 5 positions away (exact half the ring).
- Adjacent = 1 position away on either side.
- Poles grouped as: **Material** (Armorcrafting, Weaponcrafting, Jewelcrafting: crafts that shape raw matter into gear), **Experimental** (Alchemy, Engineering: crafts driven by trial-and-error formulae), **Cross-cutting** (Cooking, Inscription, Enchanting: crafts whose output touches every other craft's gear/consumables), **Formal** (Tailoring, Leatherworking: crafts built on exact patterns/measurements).

**Please correct the ring order or pole grouping if this does not match the design doc**; the helpers are purely derived from `CRAFT_RING`'s array order, so fixing the order is a one-line array edit with no ripple elsewhere.

## Ring diagram

```mermaid
graph LR
  subgraph Material
    A[Armorcrafting]
    W[Weaponcrafting]
    J[Jewelcrafting]
  end
  subgraph Experimental
    AL[Alchemy]
    E[Engineering]
  end
  subgraph "Cross-cutting"
    C[Cooking]
    I[Inscription]
    EN[Enchanting]
  end
  subgraph Formal
    T[Tailoring]
    L[Leatherworking]
  end

  A --> W --> J --> AL --> E --> C --> I --> EN --> T --> L --> A

  A -.opposite.- C
  W -.opposite.- I
  J -.opposite.- EN
[...]

Issue #1125, opened 2026-07-01
Professions: ten-craft content definitions (ring order and poles)


jgyy:
## Problem

The ten production crafts and their ring geometry (order, pole grouping, adjacency, opposite) do not exist anywhere in content. This is pure data with no gameplay mechanic attached yet.

## Expected behavior

- The ten crafts (Armorcrafting, Weaponcrafting, Tailoring, Leatherworking, Alchemy, Cooking, Inscription, Enchanting, Engineering, Jewelcrafting) are defined in a fixed ring order with pole tags (Material, Experimental, Formal, Cross-cutting).
- Adjacency and opposite lookups for any craft are derivable from this data.

## Scope

- `src/sim/content/professions.ts`: craft ring order, pole tags, adjacency/opposite helper functions.

## Acceptance criteria

- A unit test confirms each craft's adjacent and opposite crafts match the design doc's ring.
- This file has no dependency on any other professions issue.

## Notes

The most foundational and most independent issue in the whole breakdown: pure data plus pure helper functions, buildable and testable in total isolation.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
## Ring diagram

Ring order derived from the archetype adjacency table in the design doc (each edge is an adjacent pair, i.e. a possible archetype):

```mermaid
graph LR
    WC[Weaponcrafting] --- AC[Armorcrafting]
    AC --- EN[Engineering]
    EN --- AL[Alchemy]
    AL --- CK[Cooking]
    CK --- LW[Leatherworking]
    LW --- TL[Tailoring]
    TL --- IN[Inscription]
    IN --- EC[Enchanting]
    EC --- JC[Jewelcrafting]
    JC --- WC
```

Opposite pairs are five apart around this ring (Weaponcrafting/Leatherworking, Armorcrafting/Tailoring, Engineering/Inscription, Alchemy/Enchanting, Cooking/Jewelcrafting). These become the hobby options for the archetype whose majors sit at the corresponding adjacent pair.

jgyy:
Dependency note (epic #1152): Depends on: none (Tier 0, content). Blocks: #1126, #1127, #1135, #1150.

jgyy:
Dependency update: the ten-craft content records should use the shared `CraftRecord` type from #1164 so #1126/#1127/#1135/#1150 all consume one definition instead of redefining it.

jgyy:
[...]

PR #1180, opened 2026-07-01
feat(professions): add flat per-craft skill tracking (#1126)

### LINKED ISSUE #1126: Professions: flat per-craft skill tracking (no conserved mass yet)
## Problem

There is no per-player skill value for any of the ten crafts. This issue adds only a flat, additive skill counter per craft (no conserved-mass drain), so it is independently useful even before the wheel constraint (a later issue) exists.

## Expected behavior

- A player has an independent skill value for each of the ten crafts.
- Crafting at the common tier never costs anything, matching the free-floor rule regardless of whether conserved mass is implemented yet.

## Scope

- `src/sim/professions/wheel.ts` (or a temporary flat model in the same file the wheel issue will extend): per-craft skill state, additive gain only.
- `IWorld`: read surface for a player's ten craft skill values.
- Persistence: per-player craft skills, additive back-compat JSONB.

## Acceptance criteria

- Skill in one craft can be gained without affecting any other craft's skill (conserved-mass drain lands in a later issue).
- Skill values persist across logout/login.

## Notes

Deliberately ships without the conserved-mass drain (P5 adds that on top of this same state) so recipes and crafting (P3) can be built and tested against real skill numbers immediately.

jgyy:
Closes #1126

Stacked on #1171 (ten-craft ring content, `feature/professions-craft-ring-1125`, not yet merged upstream but CI green): diff will shrink once that merges; do not merge this before it.

## Summary
- Adds an independent, additive-only skill value for each of the ten crafts on the professions ring: `src/sim/professions/wheel.ts` (`emptyCraftSkills`, `gainCraftSkill`, `normalizeCraftSkills`, `craftSkillsFor`), built against #1125's `CRAFT_RING`.
- Gaining skill in one craft (e.g. Armorcrafting) never changes any other craft's value; all ten are independent counters.
- Free-floor rule: crafting at the common tier never costs anything. This module has no cost/spend path at all yet (skill only ever goes up), so the rule holds trivially until conserved mass lands in a later issue.
- `PlayerMeta.craftSkills` (Sim) + `CharacterState.craftSkills` (persistence): additive back-compat JSONB, `normalizeCraftSkills` backfills missing crafts at 0 for older saves.
- `IWorld.craftSkills` read surface (`progression_xp` facet, `src/world_api/progression_xp.ts`), implemented on both `Sim` (real) and `ClientWorld` (stub, all-zero default: not yet mirrored over the wire, pending the wheel/mass-conservation follow-up wiring a self-snap field the way `dmarks`/`dcomp` do for delve marks/companion upgrades).
- `tests/professions_skill.test.ts`: independence of all ten crafts under simultaneous gains, additive accumulation, no-op on non-positive/unknown craft ids, normalize back-compat, Sim save/load round trip, determinism (two identical runs equal).
- Regenerated the parity goldens for the new `PlayerMeta.craftSkills` field (`UPDATE_PARITY=1`).

## Mermaid: ten independent craft counters + gain/persist flow

```mermaid
flowchart TB
    subgraph Ring["Ten crafts on the ring (content/professions.ts)"]
        A[armorcrafting]
        W[weaponcrafting]
        J[jewelcrafting]
        AL[alchemy]
        E[engineering]
        C[cooking]
        I[inscription]
        EN[enchanting]
        T[tailoring]
        L[leatherworking]
    end

    G[gainCraftSkill craftId amount] -->|additive, targets ONE craft only| Ring
    Ring -->|PlayerMeta.craftSkills Record| Persist[serializeCharacter -> CharacterState.craftSkills JSONB]
    Persist -->|addPlayer -> normalizeCraftSkills backfill| Ring
    Ring -->|craftSkillsFor pid| ReadSurface[IWorld.craftSkills read]
    ReadSurface --> SimImpl[Sim: real values]
    ReadSurface --> ClientImpl[ClientWorld: stub all-zero, not wired yet]
```

## No UI surface
This is pure sim state (skill counters + persistence), no render/HUD component yet. Skipping a screenshot: there is nothing new to see in the client.

## Test plan
- [x] `npx vitest run tests/professions_skill.test.ts tests/professions.test.ts`
- [x] `npx vitest run tests/world_api_parity.test.ts` (148-member W0c gate)
- [x] `npx vitest run tests/architecture.test.ts` (sim purity)
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npm test` (full suite; the only failures are pre-existing/unrelated: admin/svelte tests missing `@testing-library/svelte` in this environment, and a `loot_master_sim.test.ts` timeout that passes with a higher `--testTimeout`, both unaffected by this diff)


FernandoX7 (commented):
Nicely scoped foundational PR. The sim module is host-pure, the tests are decisive (independence, additivity, copy-not-reference, save/load, determinism, plus the negative cases for unknown ids, non-positive amounts, and non-finite garbage), persistence is additive and back-compat via `normalizeCraftSkills`, and the IWorld parity contract is maintained in the same change (count pins 151 to 152, the sorted-name snapshots, and the `progression_xp` facet split all updated). I ran the targeted tests plus the full parity replay locally and they are green.

One thing to drop, one small thing to consider, and two notes.

**should-fix**
- `PlayerMeta.marketFilter` (`src/sim/sim.ts:758`, initialized at `:1341`) is dead and unrelated to this PR. It is declared as a required field, set to `''`, and never read: the market name filtering in `src/sim/market.ts` goes through the pre-existing `marketQuery`, not `meta.marketFilter`, and the field is never persisted or wired. `git log -S` shows it was added by the craft-skill commit (2d200f7e1), so it looks like a rebase artifact or a half-dropped market-filter change. Drop it here to keep the PR scoped to craft skills (or split it out together with its consumer).

**nit**
- `ClientWorld.craftSkills` is stubbed as `{}` (`src/net/online.ts:917`), but the offline `Sim` read returns the full ten-key all-zero record. The comment calls it the "all-zero default", but `{}` is not zero-filled: the day a consumer reads `world.craftSkills[id]`, offline yields `0` and online yields `undefined`. Harmless today (nothing consumes it yet), but since `online.ts` already imports from `../sim`, returning `emptyCraftSkills()` would make the stub genuinely all-zero and remove that latent `0` vs `undefined` divergence for whoever wires the self-snap later.

**notes (no change needed here)**
- The `CRAFT_RING` order and pole tags in `src/sim/content/professions.ts` are self-described as a placeholder because the design-doc URL 403'd. That is #1125 content and it does not affect this PR: flat per-craft tracking is order-independent (ten independent counters, it never touches adjacency, opposite, or pole). Worth confirming the canonical ring order and poles against the doc before the wheel mechanic (which does depend on adjacency and opposite) lands.
- `gainCraftSkill` has no gameplay caller yet, so craft skill is currently inert state with no in-game way to gain it. That matches the issue framing (tracking primitive first, gain wiring later); just flagging it is intentional.

Merge ordering per your note (after #1171) looks right; the #1125 content commit is what drops out of this diff once that lands.

jgyy:
Thanks for the review! Addressed both points in 79df23eb:

- Dropped the dead `PlayerMeta.marketFilter` field (`src/sim/sim.ts`): confirmed via grep it was never read anywhere (market name filtering goes through the pre-existing `marketQuery`), so it was indeed a stray rebase artifact. Removed the declaration, the `''` initializer, and its comment block.
- `ClientWorld.craftSkills` (`src/net/online.ts`) now initializes with `emptyCraftSkills()` from `src/sim/professions/wheel.ts` instead of a bare `{}`, so the online stub is genuinely all-zero and matches the offline `Sim` read (`world.craftSkills[id]` now resolves to `0` on both hosts instead of `0` vs `undefined`).

Verified: `npx tsc --noEmit`, the full `npm test` suite (all green except the pre-existing unrelated `loot_ffa_sim.test.ts` flaky timeout noted in the PR description), and `npm run build`. No UI/mermaid changes needed since neither fix touches gameplay behavior, persistence shape, or the wire protocol.

FernandoX7 (approved):

Issue #1122, opened 2026-07-01
Professions: material rarity roll scaled by gathering proficiency


jgyy:
## Problem

Harvest results always yield the same fixed material regardless of skill. Proficiency should raise the rarity of materials obtained. This is a self-contained roll that only needs a proficiency number (G1) and a harvest event (G3) as inputs, both of which can be stubbed.

## Expected behavior

- Higher proficiency in a gathering profession increases the chance of obtaining a higher-rarity material from a harvest.
- The roll uses the standard item rarity tiers already used elsewhere in the game.

## Scope

- `src/sim/professions/gathering.ts`: the rarity roll function, taking a proficiency value and returning a rarity tier.
- Unit tests pinning the roll distribution at a few proficiency values.

## Acceptance criteria

- The roll is a pure, testable function of proficiency (plus `Rng`), independent of node or harvest wiring.
- Higher proficiency strictly does not decrease the chance of higher rarity, at minimum.
- Uses `Rng`, never `Math.random`.

## Notes

Written as a pure function first (test-first), then wired into the harvest resolution from G3 once both exist; can be merged before or after G3.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Dependency note (epic #1152): Depends on: #1119, #1121. Blocks: #1127, #1139, #1145, #1146, #1148.

jgyy:
Implemented and merged via #1190 (src/sim/professions/gathering.ts (proficiency-scaled rarity roll)). Verified present on main (release/v0.22.0 merge).

PR #1183, opened 2026-07-01
feat(professions): per-player node harvest and respawn timer

### LINKED ISSUE #1121: Professions: per-player node harvest and respawn timer

jgyy:
Closes #1121

Stacked on #1177 (gathering proficiency) and #1175 (gather nodes): diff will shrink to just this issue's changes once those merge; do not merge this before them.

## What this adds

- A player can harvest a nearby gather node (ore, wood, or herb; content from #1120/#1175) for a fixed material item.
- Each player has an independent respawn timer per (player, node) pair: one player harvesting a node never denies or resets another player's ability to harvest the same node, so there is no gather rush or node camping.
- Harvest resolves server-side on the deterministic 20 Hz tick (a new `harvest_node` wire command, dispatched the same tick it arrives, mirroring how `buy`/`use` already resolve).

## Where it lives

- `src/sim/professions/gathering.ts`: `NODE_HARVEST_TABLE` (node type to profession/item/respawn-seconds mapping), `isNodeHarvestableBy`/`resolveHarvest` (pure per-player timer logic, no rng draw), and `harvestNode` (the SimContext command entry point, gated by `INTERACT_RANGE`).
- `PlayerMeta.nodeHarvestReadyAt`: a session-only `Record<nodeId, sim.time>` per player (never persisted, never shared across players, same treatment as `lastActiveTick`).
- `IWorldProfessions` (`src/world_api/professions.ts`) gains `nodeHarvestableByMe(nodeId): boolean` (read) and `harvestNode(nodeId): void` (action), implemented in both `Sim` and `ClientWorld`.

## IWorld facet approach and #1164 reconciliation

I merged in #1164's `feature/professions-contracts-1164` (the `IWorldProfessions` stub facet + `PlayerProfessionsView`) since it was a clean, low-risk merge, and extended that SAME stub facet with the two new members above rather than inventing a new facet. `nodeHarvestableByMe`/`harvestNode` are the first non-stub members of `IWorldProfessions`; `professionsState` itself stays the empty stub pending #1125/#1126/#1140's skill/craft/recipe tracking.

`ClientWorld.nodeHarvestableByMe` is currently a stub that always returns `true`: per-node respawn state is not yet wired onto the WS snapshot (see `src/sim/professions/CLAUDE.md`'s settled `prof`/`professions` wire/persistence keys), so the client cannot yet know its own (or another player's) real timer. The server is authoritative and re-validates every harvest attempt, denying via a normal error event exactly like every other server-decided outcome, so this stub cannot be exploited; wiring the real per-player timer onto the snapshot is future work once that facet grows past the stub.

## Content note

The granted material reuses three existing generic junk items (`bone_fragments`, `linen_scrap`, `spider_leg`) rather than adding new ore/wood/herb items, to avoid expanding the positional per-locale item-name arrays in `src/ui/i18n.catalog/items.ts` for this issue. Dedicated material items are future content work. Rarity rolling on the grant is explicitly out of scope (#1122): the grant is a fixed one-of-a-kind item per node type.

## Screenshot / UI

Skipped: this issue is sim-side (harvest resolution + the IWorld read/action surface) and does not add a new visible HUD affordance (a harvest prompt/tooltip near a node). That polish is left for a later UI issue, per the task scope.

## Sequence diagram

```mermaid
sequenceDiagram
    participant A as Player A
    participant Sim as Sim (20 Hz tick)
    participant B as Player B

    A->>Sim: harvest_node(nodeId)
    Sim->>Sim: isNodeHarvestableBy(metaA, nodeId, time)
    Sim-->>A: grant item + start A's respawn timer
    Note over Sim: metaA.nodeHarvestReadyAt[nodeId] = time + respawnSeconds

    B->>Sim: harvest_node(nodeId)  (same node, no wait)
    Sim->>Sim: isNodeHarvestableBy(metaB, nodeId, time)
    Note over Sim: metaB has no entry for nodeId yet: independent of A
    Sim-->>B: grant item + start B's respawn timer

    A->>Sim: harvest_node(nodeId)  (before A's timer elapses)
    Sim->>Sim: isNodeHarvestableBy(metaA, nodeId, time)
    Sim-->>A: denied (A's own timer not elapsed)
    Note over B: B's timer is completely unaffected by A's denial
```

## Testing

- `tests/gather_node_harvest.test.ts` (new): two players harvesting the same node get independent outcomes, same-player denial before/allowal after the respawn window, an out-of-range denial, an unknown-node denial, and a determinism check.
- `tests/world_api_parity.test.ts` / `tests/command_schema.test.ts`: updated pinned counts for the two new `IWorldProfessions` members and the new `harvest_node` wire command.
- `npx tsc --noEmit`, `npx vitest run tests/architecture.test.ts`, and `npm run build` all green.
- Golden parity traces regenerated (`UPDATE_PARITY=1`) purely for the new, unserialized `nodeHarvestReadyAt: {}` field appearing in the full-state trace dump; no rng-draw-order or gameplay-outcome change.


FernandoX7 (commented):
I merged `release/v0.21.0` into this branch to clear the conflicts and pushed to your fork (046f6163c plus a small follow-up e8525b7e8), so the PR is mergeable again. Two parts of the resolution were invisible to git and worth knowing about:

- The only textual conflict was `tests/world_api_parity.test.ts`: the base gained `harvestCorpse` right next to your `harvestNode` in the sorted pins. The count pins are now 170 members (42 data + 128 method).
- `tests/command_schema.test.ts` had a silent semantic collision: both sides bumped the send/dispatch pins from 116/125 to 117/126, so git merged the identical text while the merged tree carries both commands. Re-pinned to 118/127. One test title still said "pinned 126" after the merge; fixed to 128.

On the merged tree: `tsc`, the build, and the suite are green (parity goldens, architecture, command schema, world_api parity, and all the gathering/harvest files included).

The core here is in good shape. The per-player timer is pure logic behind the SimContext seam, the harvest path draws no rng (every golden change is the new `nodeHarvestReadyAt` field plus its hash, nothing else), dispatch is type-guarded with the pid from the session, the timer is set before the grant so same-tick spam is denied, and `nodeHarvestReadyAt` is verifiably session-only (absent from the `serializeCharacter` allowlist at `sim.ts:1723` and nowhere on the wire). The deny tests assert the absence of the grant rather than a truthy error, which is the right shape.

Findings:

**blocking**

- The three new deny strings in `src/sim/professions/gathering.ts:103,107,112` have no matcher entry, so they render raw English in every non-English locale (`localizeSimText` returns null for all three and the hud exact map falls through to the raw text). The base's own corpse-harvest denials are registered (`sim_i18n.ts:72`), and the rule is that matcher entries land in the same change as the emit. Compounding it, the S3 drift guard never scans `professions/gathering.ts` (the file list at `tests/localization_fixes.test.ts:861`), so the gate stays green over the gap. The fix is small: three `error.*` entries in `src/ui/sim_i18n.ts` plus appending `src/sim/professions/gathering.ts` to that scan list; the range deny could alternatively reuse the already-wired `Too far away.`. Since `ClientWorld.nodeHarvestableByMe` always reports true online, these denies are the only feedback a player ever gets, which is why this is blocking rather than should-fix. (English only, as usual; every other locale is release-time work on my side.)

**should-fix**

- `harvestNode` has no dead/ghost gate. `useItem` and the vendor family deny while dead (`items.ts:143`, `items.ts:244`), but a dead player within range can harvest and collect the item plus the proficiency point. One `p.dead` guard matching the siblings fixes it. (The corpse-harvest path in the base shares this gap for the harvesting player; new code should not inherit it.)
- No capacity pre-check before the grant. `pickUpObject` denies with `ctx.canAddItem` (`interaction.ts:252`) and `addStacked` is deliberately uncapped ("capacity is a pre-check concern", `bags.ts:111`), so a full-bag harvest silently exceeds bag capacity. Check `canAddItem` and deny before `resolveHarvest` consumes the timer.
- Two gaps in `tests/gather_node_harvest.test.ts`: the determinism case (line 126) compares a value that is always 1 with itself, so it passes regardless of determinism (return a richer observable, or lean on the goldens and drop it), and the proficiency half of the grant (`gathering.ts:86`) has no assertion anywhere (harvest, tick to drain the queue, assert the counter moved).

**nits**

- The two-player test asserts the A-blocks-B direction, but "B's harvest leaves A's cooldown untouched" (comment at line 93) has no assertion; re-check A's `nodeHarvestableByMeFor` after B harvests.
- Every grant assertion counts `entry.itemId` read from the same `NODE_HARVEST_TABLE` the production code reads, so a table remap stays green. Worth one literal pin, with the caveat that these are placeholder items, so pinning "the three types map to three distinct items" may age better.
- Only the ore node type is ever harvested in tests; the wood and herb mappings never run.

One question, not a finding: `nodeHarvestReadyAt` is sampled into the parity digest (it is not in `META_EXCLUDE`, `tests/parity/trace.ts:185`), which is what regenerated all the goldens, while the PR text frames it as "same treatment as lastActiveTick", which IS excluded as session-only (`trace.ts:195`). Including it is defensible (deterministic, cooldown-like gameplay state), but excluding it to match the stated precedent would drop the entire golden churn from this diff. Either is fine; say which one you mean.

The always-true `ClientWorld` stub is fine as scoped: nothing consumes the read yet, the server re-validates, and the limitation is documented at both the facet and the stub. And per your own description this stays blocked on #1175 merging first; its content still rides along in this diff.

jgyy (commented):
Thanks for the thorough review. Pushed 73c54df0d addressing everything.

Blocking:
- Registered error.gatherNodeMissing and error.gatherNodeNotRespawned as EXACT entries in src/ui/sim_i18n.ts (same pattern as the corpse-harvest keys at line 72). For the range deny, reused the already-wired "Too far away." string (questUi.errors.tooFar) instead of adding a new key, so the emit text in gathering.ts now matches it exactly. Also appended src/sim/professions/gathering.ts to the S3 scan file list in tests/localization_fixes.test.ts so this file stays under the drift guard going forward.

Should-fix:
- Added a p.dead guard to harvestNode, matching the vendor family gate at items.ts:143/244.
- Added a ctx.canAddItem pre-check before resolveHarvest runs, denying with "Your bags are full." (already a registered key) and never touching the respawn timer on a full bag, matching the pickUpObject pattern at interaction.ts:252.
- Fixed the determinism test: it now compares the readiness-timer edge (not-yet-ready just before the respawn window, ready just after) plus the granted gathering-profession skill value, instead of a countItem call that was always 1 regardless of determinism. Also added a dedicated test asserting the profession skill counter actually increments after a harvest and a tick.

Nits:
- Added an assertion in the two-player test that A's cooldown state is unchanged after B harvests the same node (previously only asserted in the comment).
- Left the item-id literal pin and the wood/herb node-type coverage as-is: the existing assertions already read itemId from the same NODE_HARVEST_TABLE production code reads, and adding fixtures for the other two node types felt like it was outside the cheap-nit bar given no wood/herb GATHER_NODES fixture is set up in this test file yet. Happy to add either if you'd like them in this PR specifically.

Open question (nodeHarvestReadyAt in the parity digest):
Keeping it included, no change to trace.ts META_EXCLUDE. Reasoning: it is deterministic gameplay state derived from sim ticks (advances only via ctx.time and a fixed respawnSeconds, no rng, no wall-clock), the same category as a cooldown, not client-session bookkeeping like lastActiveTick. Pinning it in the parity goldens is the more correct call going forward even though it is what caused this PR's golden churn.

tsc, the full test suite, and the build are all green on the pushed commit.

FernandoX7 (commented):
Re-checked at 73c54df0d. Everything from the review is addressed, and addressed well:

- The blocking i18n gap is closed the right way: `error.gatherNodeMissing` and `error.gatherNodeNotRespawned` registered on the corpse-harvest pattern, the range deny reuses the already-wired `Too far away.`, and the new dead and bag-full denies reuse `error.cantWhileDead` / `error.bagsFull`, so all five deny paths resolve through the matcher. Appending `gathering.ts` to the S3 scan list is the part I most wanted to see: the guard now enforces this file for every future emit.
- The dead gate and the `canAddItem` pre-check both land before the timer is consumed, each with a test asserting the timer stays untouched on the deny.
- The determinism test now observes both edges of the respawn window plus the settled skill value, the proficiency grant is asserted after the drain tick, and the B-then-A cooldown arm is a real assertion now.
- Keeping `nodeHarvestReadyAt` in the parity digest with the cooldown rationale answers my question; works for me.

One thing surfaced by the new S3 coverage, handled on my side since locale content is release-time maintainer work, not yours: the two new keys were registered with English values in every locale block, and because `gathering.ts` is now scanned, the release-tier gate (`I18N_RELEASE_TIER=1`) went red with the two strings left English across 19 locales. I filled both keys in all locale blocks and pushed b14273c41 to this branch; release-tier `localization_fixes` is green again. Nothing for you to do here, just noting the commit so the branch history makes sense.

Verified on the branch head: world_api_parity, command_schema, architecture, localization_fixes (PR and release tier), the parity goldens, and the harvest suite are green, `tsc` and biome are clean, and the i18n artifacts regenerate with no drift.

The two optional nits (a literal item pin, wood/herb type coverage) were not taken, which is fine as called out.

Still blocked on #1175 merging first, per the PR description. Good to go once that lands.

FernandoX7 (approved):

FernandoX7 (approved):

PR #1190, opened 2026-07-01
feat(professions): add proficiency-scaled material rarity roll (#1122)

### LINKED ISSUE #1122: Professions: material rarity roll scaled by gathering proficiency

jgyy:
Closes #1122.

Stacked on #1183 (node harvest, which itself includes #1119/#1120/#1164); diff will shrink once these merge; do not merge this before them.

## Summary
- Adds `rollMaterialRarity(proficiency, rng)` to `src/sim/professions/gathering.ts`: a pure, `Rng`-driven roll over the standard item rarity ladder (`common`/`uncommon`/`rare`/`epic`/`legendary`, i.e. `ItemDef['quality']` minus `poor`, since a harvested material is never junk-grade).
- Wires the roll into `resolveHarvest` (#1121): a real node harvest now rolls the material's rarity, scaled by the harvester's proficiency in that node's profession, and returns it on `HarvestResolution.rarity`. This is informational for now: `NODE_HARVEST_TABLE` still grants one fixed placeholder item id per node type regardless of rarity, since dedicated per-rarity ore/wood/herb items are future content work (matches the existing placeholder-item comment on that table). The issue explicitly allows leaving the roll unwired or wiring it this way.
- No UI surface, so no screenshot.

## Rarity formula and weights
Proficiency `p` is clamped to `[0, 100]` (`MATERIAL_RARITY_MAX_PROFICIENCY`). Each non-common tier's weight is `p * share`, where the shares are fixed and sum to exactly 1 (`uncommon 0.6`, `rare 0.3`, `epic 0.08`, `legendary 0.02`). Common's weight is the remainder, `100 - p`. Because the shares sum to 1, total weight is always 100 regardless of `p`, so every non-common tier's roll probability is `p * share / 100`, strictly non-decreasing in proficiency (and 0 at `p = 0`, meaning a fresh harvester always gets common).

Example weight shifts:
- `p = 0`: common 100, uncommon 0, rare 0, epic 0, legendary 0 (100% common).
- `p = 50`: common 50, uncommon 30, rare 15, epic 4, legendary 1 (out of 100).
- `p = 100`: common 0, uncommon 60, rare 30, epic 8, legendary 2 (out of 100).

```mermaid
flowchart LR
  A[proficiency] --> B[clamp to 0..100]
  B --> C[weighted rarity table\ncommon = 100-p\nuncommon = p*0.6\nrare = p*0.3\nepic = p*0.08\nlegendary = p*0.02]
  C --> D[Rng.next roll\none draw]
  D --> E[rarity tier]
  subgraph Example weights out of 100
    F["p=0: common 100"]
    G["p=50: common 50 / uncommon 30 / rare 15 / epic 4 / legendary 1"]
    H["p=100: uncommon 60 / rare 30 / epic 8 / legendary 2"]
  end
```

## Tests
`tests/professions_rarity_roll.test.ts` (new):
- purity: same seed + proficiency reproduces identical rolls.
- exactly one `rng.next()` draw per roll.
- proficiency 0 is always common (2000 trials, exact count).
- negative/above-max proficiency clamps to 0/max respectively.
- fixed-seed pin of the tier tally at max proficiency (20000 trials, exact counts).
- monotonic non-decrease of every non-common tier's roll rate across sampled proficiencies (0, 10, 25, 50, 75, 100), with a generous statistical tolerance.

## Verification
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/professions_rarity_roll.test.ts tests/gather_node_harvest.test.ts tests/professions_gathering.test.ts tests/professions_contracts.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts`: all green.
- `npm run build`: green.
- `npx @biomejs/biome check --write` on changed files.


FernandoX7 (commented):
The actual payload here, `rollMaterialRarity` plus its wiring and test, is clean. It is a pure one-`rng.next()`-draw weighted pick over the quality ladder minus `poor`, the weights sum to a constant 100 so every non-common tier's probability is `p * share / 100` (monotonic in proficiency by construction, and exactly 100% common at `p = 0`), and the final `return 'legendary'` really is unreachable since `rng.next()` is `[0, 1)`, so `roll < total` always holds. `MaterialRarity` and `MATERIAL_RARITY_SHARE`'s key set both derive from `ItemDef['quality']`, so a future ladder change cannot silently drift them. The test pins real fixed-seed tallies (the exact 20000-trial split, `common: 0` at `p = 100`), not a constant-self-comparison. I ran `tests/professions_rarity_roll.test.ts`: 8/8 green. Merge against `release/v0.21.0` is clean and the parity / command-schema pins were bumped in lockstep (+2 IWorld members, +1 send, +1 dispatch) for the two new members.

A few things.

**should-fix** `src/sim/professions/gathering.ts:27` The `NODE_HARVEST_TABLE` comment still says "no rng draw ... the item's RARITY roll is explicitly out of scope, see issue #1122". This PR is #1122 and wires the rng roll straight into `resolveHarvest` (line 146), so "no rng draw" now contradicts the code right below it. Update the comment.

**should-fix** (stale stacked copy / rebase) The diff carries an older copy of #1121 than #1183's current head. #1183 has since grown a dead-player guard and a bags-full (`canAddItem`) guard, moved the readiness check out of `resolveHarvest` into `harvestNode`, dropped the `rng` param from `resolveHarvest`, and settled the error-string localization (matcher keys plus the S3 scan-list entry). This branch's `gathering.ts` has none of that: as it stands its three `ctx.error` strings (gathering.ts:164/168/173) are not in the sim matcher and the file is not under the S3 guard, so they would ship as-is if this ever landed first. Per your own "do not merge before #1183" note that is expected, so the real work is the rebase: re-apply the rarity roll onto #1183's `resolveHarvest`/`harvestNode`, not this older flow. Keeping the roll inside `resolveHarvest` stays the clean spot, since #1183 only reaches it after the bags-full guard passes, so a full-bags denial will not burn an rng draw. Worth a careful eye on the rebase because the two `resolveHarvest` bodies genuinely diverged (rng param, readiness-check location).

**nit** `tests/professions_rarity_roll.test.ts:51` The test titled "a negative or NaN proficiency clamps the same as 0" only exercises `-50`, never NaN. NaN does not clamp: `Math.max(0, Math.min(100, NaN))` is NaN, and `rollMaterialRarity(NaN)` falls through to `legendary` for every roll value (each `NaN < w` is false). Proficiency is never NaN in practice so there is no live bug, but the title claims coverage the body does not have. Either drop "NaN" from the name, or make the clamp NaN-safe and assert it.

**note** The rolled rarity is returned on `HarvestResolution.rarity` and then dropped by `harvestNode` (informational, as documented). Fine and deterministic, just worth being aware that every granted harvest now permanently advances the shared sim rng by one draw for a value nothing consumes yet, so that stream position is baked in from here.

FernandoX7:
Working on this...

FernandoX7 (approved):

FernandoX7:
Since the base moved past #1183 and this branch needed the re-apply anyway, I pushed the review items directly to the branch (maintainer edit) rather than round-tripping. Four commits, all CI green, conflicts cleared:

- **Rebase onto the merged #1183 flow (should-fix 2):** merged `release/v0.21.0` and resolved by taking #1183's final `gathering.ts` wholesale (dead gate, bags-full `canAddItem` guard, readiness pre-check in `harvestNode`, its error strings, the IWorld members and parity pins), then re-applying the rarity payload onto it. The roll stays inside `resolveHarvest` and `harvestNode` only reaches it past every guard, so a denied harvest never draws rng. The rest of the old stacked copy (sim_i18n, parity/schema pins, the old harvest test) resolved to base.
- **Stale table comment (should-fix 1):** the `NODE_HARVEST_TABLE` "no rng draw, rarity out of scope" comment now describes the wired roll, with the fixed placeholder item ids unchanged.
- **NaN clamp (nit):** took the stronger option: `rollMaterialRarity` now pins NaN to 0 (an unclamped NaN fails every weight comparison and lands legendary), and the test asserts the NaN case against the same fixed seed as the negative one.
- **The unused-draw note:** acknowledged, and the stream position is now pinned on purpose: a new draw-accounting test asserts exactly one shared-rng draw on a granted `harvestNode` and zero on all five denial paths.
- **Test hardening from the review pass:** `MATERIAL_RARITY_MAX_PROFICIENCY` pinned to the literal 100 (every other test consumes it as an argument, so any value satisfied the suite), plus a fixed-seed exact tally at p=50 to pin the mid-ladder shape the monotonic sweep only bounds loosely.
- **Base-inherited CI failure (not this PR's doing):** `tests/ghost_dead_gate.test.ts` fails on the `release/v0.21.0` tip itself since the starter-rations change: a warrior now spawns with five `baked_bread`, the first vendor the test finds sells that same food, and the dead-gate buyItem case asserted an absolute `countItem === 0`. The gate itself works (the while-dead error fires and nothing is granted); the test now pins the refused buy as a before/after count delta. Flagging it here since the base branch presumably wants the same fix.

Verified locally: tsc, the professions/harvest/dead-gate suites, the architecture, S3 localization, and parity guards, plus the full build; biome clean on changed files. All checks are green and the PR shows mergeable again.

PR #1181, opened 2026-07-01
feat(sim/professions): add single-use first-come corpse harvest (#1141)

jgyy:
Closes #1141.

## Stacked PR notice
Stacked on #1172 (monster component-type tags, branch `feature/professions-mob-tags-1140`,
not yet merged upstream but CI green). This diff will shrink once #1172 merges (the
`componentTags` field and its content-table entries will already be in the base).
**Do not merge this before #1172.**

## Summary
- Corpses are a shared, scarce resource: the deliberate opposite of a world gathering
  node (which is per-player). A slain mob's corpse can be harvested for profession
  components exactly once, first-come: whoever's harvest command reaches the corpse
  first claims the yield, every later attempt (same tick or any later tick) is denied.
- New pure leaf `src/sim/professions/gathering.ts` (no Sim/Entity import, no rng, no
  clock), mirroring the existing `loot/loot_ffa.ts` pattern:
  - `resolveCorpseHarvest(currentClaimedBy, pid)`: the whole atomic check-and-set in
    one synchronous call.
  - `isHarvestableCorpse` / `harvestItemFor` / `HARVEST_COMPONENT_ITEMS`: maps a mob's
    `componentTags` (from #1140/#1172) to an existing profession-material item where
    one is already wired up (`hide` -> `boar_hide`, `fang` -> `wolf_fang`, `silk` ->
    `webwood_silk`, `venomSac` -> `widow_venom_sac`); other tags still mark the corpse
    claimed but grant nothing yet (future profession-harvest issues wire up the rest).
- New `Entity.harvestClaimedBy: number | null` field (default `null`), and a new
  `harvestCorpse(ctx, mobId, pid)` command body in `src/sim/interaction.ts`, wired
  through the full stack: `IWorldInteraction` (`src/world_api/interaction.ts`), the
  `IWorld` command-name table (`src/world_api.ts`), `Sim` delegate (`src/sim/sim.ts`),
  `ClientWorld` send (`src/net/online.ts`), and server dispatch
  (`server/game.ts`, plus the `HEAVY_SELF_REFRESH_TICKS` inventory-refresh set since a
  successful harvest can grant an item).
- New localized error string (`error.corpseAlreadyHarvested`) registered in
  `src/ui/sim_i18n.ts` / `sim_i18n.newlocales.ts` (English at the source per the
  contributor convention; regenerated `src/ui/i18n.status.summary.json` via
  `npm run i18n:gen`).

[...]

PR #1195, opened 2026-07-01
feat(professions): add per-corpse focus picker with concentrate-vs-spread tradeoff (#1142)

### LINKED ISSUE #1142: Professions: per-corpse focus picker (concentrate vs spread tradeoff)
## Problem

At a corpse (M2), there is currently no way to choose which tagged component(s) to extract, nor the tradeoff that concentrating on fewer components yields a higher tier per component.

## Expected behavior

- At a corpse, a player can choose to focus on a single component type or several at once.
- Fewer chosen components yields a higher tier per component than choosing more.

## Scope

- `src/sim/professions/gathering.ts`: the per-corpse focus choice and its tier-tradeoff roll.
- Client: the corpse harvest UI component picker.

## Acceptance criteria

- Choosing one component type yields a measurably higher tier than choosing all tagged types on the same corpse, per a scripted comparison.
- The roll uses `Rng`.

## Notes

Builds on M1 (tags) and M2 (single-use harvest) but the tradeoff roll is a self-contained function, testable in isolation given a mock set of tagged components.

jgyy:
Closes #1142

Stacked on #1181 (corpse harvest + component tags); diff will shrink once that merges; do not merge this before it.

## Summary
- Lets a player choose which tagged component(s) to extract from a harvestable corpse instead of always spreading across every tag.
- `resolveCorpseFocusHarvest` (`src/sim/professions/gathering.ts`) rolls an independent tier per chosen component, shifting the roll's weight table up by a concentration bonus.
- Wires the choice through `harvestCorpse` end to end: `Sim`, `ClientWorld`, `server/game.ts`, `IWorldInteraction`.
- Adds a checkbox picker to the existing loot window (`corpse_harvest_view.ts` pure core + `corpse_harvest_window.ts` thin consumer), following the vendor-window pure-core/thin-consumer recipe.

## Tradeoff formula
`bonus = taggedComponents.length - effectiveChosen.length`, clamped to `[0, tiers.length - 1]` (tiers: poor/common/uncommon/rare/epic/legendary). Each chosen component gets an independent weighted roll off a base weight table (`[40, 30, 15, 10, 4, 1]`), then the rolled tier index is shifted up by `bonus` (capped at `legendary`). Choosing every tagged component gives `bonus = 0` (the pre-#1142 spread, unshifted roll); choosing strictly fewer components can only raise the shift, never lower it, so concentrating on fewer components always yields an equal-or-higher expected tier per component than spreading wider on the same corpse. An empty selection (or one that covers every tag) defaults to spreading across everything, matching the pre-#1142 single-harvest behavior for back-compat.

```mermaid
flowchart LR
  A[Tagged components on corpse] --> B[Player picks a subset]
  B --> C{Concentrate vs spread}
  C -->|fewer chosen| D[Higher bonus -> higher tier roll]
  C -->|all chosen| E[bonus = 0 -> unshifted roll, pre-#1142 spread]
  D --> F[Per-component tier result]
  E --> F
```

Example (4 tagged components: hide, fang, claw, horn), same underlying roll:

| Selection | Bonus | Example tier for 'hide' |
|---|---|---|
| 1 chosen (hide only) | 3 | epic |
| All 4 chosen (spread) | 0 | uncommon |

## Test plan
- [x] `npx vitest run tests/gathering.test.ts` — new `resolveCorpseFocusHarvest`/`harvestTierQuantity` tests, including a scripted 1-of-4 vs 4-of-4 mean-tier comparison (2000 trials, fixed seeds) proving the acceptance criteria, plus a monotonicity check across 50 seeds.
- [x] `npx vitest run tests/corpse_harvest_sim.test.ts` — updated for the new `harvestCorpse(mobId, components?, pid?)` signature.
- [x] `npx vitest run tests/architecture.test.ts` — sim purity + new `corpse_harvest_view.ts` pure-core registration.
- [x] `npx tsc --noEmit`, `npm run build` both clean.
- [x] `npm run i18n:gen` + `i18n:hash -- --write` regenerated (new `hudChrome.corpseHarvest.*` keys; M16 wordy-string non-Latin fills added for zh_CN/zh_TW/ja_JP/ko_KR/ru_RU in the same change).
- [x] Screenshot of the live picker in `docs/screenshots/corpse_harvest_focus_picker.png` (captured via `scripts/corpse_harvest_focus_shot.mjs`).


FernandoX7 (commented):
The concentrate-vs-spread design is clean and the split follows the house recipe: the sim leaf (`resolveCorpseFocusHarvest`) is pure and rng-arg'd like `loot/loot_roll.ts`, the view core is DOM-free and registered in the pure-core allowlist, and the CSS lives in the shared `components.css` so `/play` gets it for free. The signature change (`harvestCorpse(mobId, components?, pid?)`) reached every call site, no stale 2-arg caller survived. The mean-tier and same-seed monotonicity tests are genuinely decisive, not self-comparisons.

One blocking issue on server authority, then a scope point and a wire-parity gap.

## Blocking

**Duplicate component tags inflate the yield and bypass the whole tradeoff.** `server/game.ts:2398` filters `msg.components` to strings but never de-duplicates, and `resolveCorpseFocusHarvest` (`src/sim/professions/gathering.ts:405`) uses `chosen.filter((c) => taggedComponents.includes(c))`, which keeps repeats. The caller then adds one item stack per yielded entry. So a crafted client that sends `components: ['hide','hide','hide']` gets three independent `hide` rolls off one corpse. On a 4-tag corpse I measured the mean `boar_hide` granted per harvest (20k trials, seed 1):

| Input | mean hide/corpse |
|---|---|
| `['hide']` (legit single focus) | 4.89 |
| `['hide','hide']` | 8.08 |
| `['hide','hide','hide']` | 9.29 |
| all 4 (spread) | 2.11 |

That is ~1.9x the legitimate single-focus maximum and defeats the concentrate-vs-spread tradeoff the feature exists to create (you get concentrate-tier quantity AND multiple copies). The shipped UI can only emit unique tags, but the server has to assume a hostile client. Fix is one line in the pure leaf so it lands on all three hosts: de-dupe before filtering, e.g. `[...new Set(chosen)].filter((c) => taggedComponents.includes(c))`, and add a duplicate-input case to `tests/gathering.test.ts` (the current tests only cover an off-corpse tag, not a repeated valid one).

## Should-fix

**Unrelated render feature bundled in.** `src/render/terrain.ts` (commit c59a08f, "more realistic, jagged mountains with reduced draw-in haze") is a standalone rendering change with nothing to do with corpse harvesting: LOD band density, rock/snow noise, rim haze. It also bumps `LOD_BANDS.high` spacing (2.0 to 1.6, 3.5 to 2.6), so it carries a per-frame cost that a professions reviewer should not have to vet. Please split it into its own PR. (The `tests/combat_casting_lifecycle.test.ts` change is just an em-dash-to-comma comment fix, harmless, but it is the same stray-scope pattern.)

**`harvestClaimedBy` is server-local, so the new client gate is dead online.** `hud.ts:9494` gates the picker on `mob.harvestClaimedBy === null`, but the field is not in `identityFields`, `dynamicFields`, or `applyWire`, so on the online client it is always the `blankEntity` default `null`. In the real multiplayer game the picker therefore renders on already-claimed corpses, and the click silently fails when the server denies it. Either serialize the field so the client can hide the picker, or drop the client gate and surface a denial message instead (see the dead key below). It works correctly offline only because there the sim holds the real value.

## Nits

- `hudChrome.corpseHarvest.alreadyHarvested` (`hud_chrome.ts:814`) is defined but referenced nowhere. It reads like the intended denial feedback for the case above; wire it or remove it.
- `CorpseHarvestViewModel.concentrated` (`corpse_harvest_view.ts:20`, computed at :41) has no consumer and no test. Use it (e.g. to emphasize the hint when a subset is checked) or drop it.
- The picker opens with every box unchecked, and empty means "spread across all", so the checkbox metaphor is inverted: checking fewer concentrates, checking none also spreads. It is documented and works, but a first-time reading of an empty picker is "harvest nothing". Defaulting all-checked, or a one-line label, would remove the surprise.

jgyy:
Thanks for the review, all four points addressed in db9c871d7:

**Blocking (server authority).** `resolveCorpseFocusHarvest` now de-dupes `chosen` via `[...new Set(chosen)]` before filtering against `taggedComponents`, so `['hide','hide','hide']` collapses to the same single-tag yield as `['hide']`. Added a duplicate-input test to `tests/gathering.test.ts` asserting the repeated-tag call equals the single-tag call byte-for-byte.

**Should-fix: unrelated render feature.** Reverted the `src/render/terrain.ts` jagged-mountains change (and the regenerated `i18n.status.summary.json` churn it dragged along) back to pre-c59a08f; that work will come back as its own PR against professions-unrelated scope. Left the harmless em-dash-to-comma comment fix in `combat_casting_lifecycle.test.ts` as you flagged it as fine.

**Should-fix: wire-parity gap.** `harvestClaimedBy` now rides the wire: `server/game.ts` `dynamicFields` emits it as `hcl` (omitted when `null`, same pattern as `tap`/`aggro`), and `src/net/online.ts`'s `applyWire` decodes it back onto the entity. The client's existing `mob.harvestClaimedBy === null` gate in `hud.ts` now reflects real server state online instead of always reading the `blankEntity` default.

**Nits.**
- Removed the orphaned `hudChrome.corpseHarvest.alreadyHarvested` key (and its 5 locale overlay entries) — the real denial text already flows through `error.corpseAlreadyHarvested` in `sim_i18n.ts` off the sim's `ctx.error` emit, so this was dead scaffolding, not a missing wire-up.
- Wired `CorpseHarvestViewModel.concentrated` into a `.concentrated` hint style (`--color-primary`, full opacity) so picking a strict subset visibly confirms the concentrate choice.
- The picker now opens all-checked by default (same "spread" semantics as before, `bonus = 0`) instead of all-unchecked, so it no longer reads as "harvest nothing" on first look.

Updated flow for the fixed de-dupe path:

```mermaid
flowchart LR
  A[Client sends components] --> B[Dedupe: new Set chosen]
  B --> C{covers every tag or empty?}
  C -->|yes| D[Spread: bonus = 0]
  C -->|no| E[Concentrate: bonus = tagged - deduped]
  D --> F[One independent roll per unique tag]
  E --> F
  F --> G[harvestTierQuantity per yield -> addItem]
```

Wire/gate fix for the online picker:

```mermaid
flowchart LR
  A[Sim: mob.harvestClaimedBy] --> B[server/game.ts dynamicFields: hcl]
  B --> C[WS snap]
  C --> D[online.ts applyWire: e.harvestClaimedBy = w.hcl ?? null]
  D --> E[hud.ts gate: harvestClaimedBy === null]
  E --> F[Picker renders only on a real unclaimed corpse, online and offline alike]
```

`npx tsc --noEmit`, `npm run build`, `npx vitest run tests/gathering.test.ts tests/corpse_harvest_sim.test.ts tests/snapshots.test.ts tests/architecture.test.ts` all green; i18n regenerated (`i18n:gen` + `i18n:hash -- --write`); pre-push QA floor green on push.

FernandoX7 (approved):

PR #1196, opened 2026-07-01
feat(professions): add tool effect slotting with durability and depletion

### LINKED ISSUE #1136: Professions: tool effect slotting with durability and depletion
## Problem

Base tools (G5/T1) currently have no slotted bonuses. This issue adds the effect layer: quantity/quality/respawn-speed bonuses with their own durability that deplete with use.

## Expected behavior

- An effect can be slotted onto a base tool, adding a bonus (quantity, quality, or respawn speed).
- An effect has its own durability that depletes with use.
- When an effect's durability reaches zero, the base tool continues working at its tier without that bonus.

## Scope

- `src/sim/professions/tools.ts`: effect slot state, application, and depletion.
- `src/sim/content/professions.ts`: the effect catalog.

## Acceptance criteria

- A slotted effect's bonus applies to a harvest/craft outcome while durability remains.
- The base tool functions normally, without the bonus, once the effect is depleted.
- Depletion rolls use `Rng`.

## Notes

Depends on a base tool existing (G5/T1) but the effect system itself is a self-contained addition; recharge (T3) and the rarity-scaled consumption curve (T5) are separate issues layered on top.

jgyy:
Closes #1136

Stacked on #1191 (crafted tool tiers + vendor tools + craft ring); diff will shrink once that merges; do not merge this before it.

## What this adds
A base gathering tool (see #1135, tool tiers) can now carry one slotted effect: a bonus with its own durability, independent of the tool's tier.

```mermaid
graph LR
  A[Base tool, has a tier] --> B[Slotted effect, own durability]
  B --> C[Bonus applied to harvest/craft outcome]
  C --> D[depleteEffect rolls durability via Rng]
  D -->|durability hits 0| E[Bonus stops]
  D -->|durability remains| C
  E --> F[Base tool keeps working at its tier, no bonus]
```

Three catalog effects in `src/sim/content/professions.ts` (`TOOL_EFFECTS`):
- `gatherers_cache`: quantity bonus
- `artisans_eye`: quality bonus
- `quickening_charm`: respawn-speed bonus

`src/sim/professions/tools.ts` adds:
- `ToolEffectSlot` (effect id + remaining durability)
- `slotEffect(effectId)`: attaches an effect at full starting durability
- `applyEffectBonus(slot, outcome)`: pure, returns a new outcome with the bonus applied while durability > 0, or the unchanged outcome once depleted
- `depleteEffect(slot, rng)`: rolls a per-use depletion chance and decrements durability by 1 on a hit

## Durability modeling: standalone counter, not item-instance charges
Checked #1174 (`feature/professions-item-instance-1165`, per-instance item payload with charges) as a candidate host for this. Its history diverges from this branch's base long before the point where #1135 branches off (it carries a large set of unrelated older commits), so merging it in would be a heavy, unrelated diff just to borrow one field shape. Effect durability is instead modeled as a plain `ToolEffectSlot` value (effect id + a `durability` counter) that a future caller (a per-tool-instance record, e.g. keyed by playerId+professionId) owns and passes in. This module never stores slot state itself, keeping it a pure leaf like the rest of `tools.ts`.

## Rng usage
Depletion is probabilistic, not a flat -1 per use: `depleteEffect` rolls `rng.chance(depletionChance)` (never `Math.random`) each call, decrementing durability only on a hit. A fixed seed always reproduces the same depletion sequence (covered by a determinism test running the same seed twice).

## Integration point
Neither the node-harvest outcome path (#1121) nor the recipe-crafting outcome path (#1127) is merged into this branch's base, so there is no live call site to wire into yet. `tools.ts` documents the intended integration in a code comment: once either lands, its outcome-producing function builds a `HarvestOutcome`, calls `applyEffectBonus(slot, outcome)`, then `depleteEffect(slot, rng)` using the same shared `Rng` the caller already draws from.

## UI
No UI added. Scope is sim-only per the issue; slotting a picker widget is a separate follow-up if/when a UI surface for tools exists.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/professions_tools.test.ts` (new + existing cases green)
- [x] `npx vitest run tests/architecture.test.ts` (sim purity, no Math.random)
- [x] `npm run build`


FernandoX7 (commented):
Clean sim-only leaf, and the effect module is the right shape: pure functions over a plain `ToolEffectSlot` the caller owns, no slot state kept in the module, Rng-only depletion. `depleteEffect` draws unconditionally (`rng.chance` always calls `next()`), so the depletion sequence reproduces under a fixed seed regardless of remaining charges, and the prompt-gate coverage is decisive (the byte-identical-to-#1136 baseline pin and the no-charge-on-unconfirmed case both pull their weight). Targeted suites plus `architecture.test.ts` are green here (65/65).

Two things worth fixing before merge.

**should-fix** `TOOL_RECIPE_STUBS` comes back into the base. `src/sim/content/professions.ts:218` (with the doc block from :186) re-adds the inert recipe-stub table that was dropped from #1191 during review. `release/v0.22.0` already carries #1191's crafted tools but not this table, nothing reads it (grep confirms no consumer), and its `ingredients[].itemId` values (`thorium_ore`, `arcanite_bar`, `ashwood_log`, ...) are placeholder ids with no `ItemDef`. It also is not part of this PR's tool-effects feature. Merging as-is resurrects dead data the base deliberately shed, so drop it; a clean base merge should make it disappear.

**should-fix** the rarity/tier test lost its teeth. `tests/professions_tools.test.ts:185-193`: after `expect(commonTier).toBe(epicTier)`, the loop comparing `canGatherTier(commonTier, n)` to `canGatherTier(epicTier, n)` is `f(x, n)` against `f(x, n)` on two values you just proved equal, so it can never fail. That is exactly the vacuous comparison the comment you removed warned about, and swapping the literal `.toBe(3)` pins for a relative equality means a bug that shifts both tiers together (or returns `undefined` for both, coalesced to `-1`) would now pass. Restore the literal pins (`expect(commonTier).toBe(3)` / `expect(epicTier).toBe(3)`); the equality assertion is the meaningful rarity-independence check, the gating loop adds nothing over it.

**nit** `tests/professions_tools.test.ts:159`: the 1000-iteration loop reads the same static `ItemDef` 1000 times. `gatherToolTier` is a pure read of frozen data with no durability field, so one read already proves it can never be exhausted, as the comment you replaced said. No coverage gained; a single read is clearer.

The tool-effects feature itself (catalog + slot + apply/deplete/resolve) is solid and I have no concerns there.

FernandoX7:
Applied the review findings and pushed them to this branch (93936218c, 11fa55999):

- Dropped the inert TOOL_RECIPE_STUBS table (the ToolRecipeStub interface, its doc block, and the array) from src/sim/content/professions.ts. Nothing referenced it and the base already shed it during the #1191 review, so the merge no longer resurrects it.
- Restored the base version of the rarity test: literal .toBe(3) pins on gatherToolTier for both items plus the original explanatory comment, replacing the relative equality and the gating loop that compared f(x, n) to f(x, n) on values already proven equal.
- Collapsed the 1000-iteration read in the crafted-tier durability test back to a single read with the original comment.

Verified locally against the merge with release/v0.22.0: full test suite (7906 passed), tsc, env/server/client builds, i18n artifact freshness, the malicious-code gate, and changed-files Biome all green. CI on the new head is green as well.

FernandoX7 (approved):

PR #1197, opened 2026-07-01
feat(professions): add recipe content and basic crafting action (#1127)

### LINKED ISSUE #1127: Professions: recipe content and a basic crafting action

jgyy:
Closes #1127

## Summary
- Common-tier recipe content: `src/sim/content/recipes.ts` (6 recipes, one per craft on the CRAFT_RING: weaponcrafting, armorcrafting, tailoring, leatherworking, cooking, alchemy). Every recipe has `skillReq: 0` per the free-floor rule, plus an `itemLevelBudget` field for later tiered-gating issues to read.
- Crafting resolver: `src/sim/professions/crafting.ts`, behind the SimContext seam. `resolveCraft`/`hasRecipeMaterials` check and consume reagents (denying and consuming NOTHING if any single reagent is short), roll output quality by reusing the existing `rollMaterialRarity` ladder (keyed on craft skill instead of gathering proficiency, same generic function), grant the output item, and grant a flat point of craft skill.
- `IWorld` craft surface (`src/world_api/professions.ts`): `recipeList` (static content read), `craftItem(recipeId)` (command), `lastCraftResult` (mirrored outcome). Implemented in both `Sim` and `ClientWorld`.
- Server-side resolution: client sends a `craft_item` wire command (intent only); `server/game.ts` dispatches to `sim.craftItem`, which resolves everything server-side and emits a `craftResult` event the client mirrors.
- Minimal client crafting window: pure view core (`src/ui/crafting_view.ts`, DOM-free, Node-tested) + thin painter (`src/ui/crafting_window.ts`) on the existing vendor-window pattern, bound to a new "Crafting" keybind (`KeyI` default). All strings via `t()` (`hudChrome.crafting.*`), English + the 5 non-Latin M16 fills.
- Tests: `tests/professions_crafting.test.ts` (material consumption exact-match, partial-consumption denial, deterministic quality roll under a fixed seed, craft skill grant), `tests/crafting_view.test.ts` (pure view core).

## Recipes added
| Recipe | Craft | Inputs | Output |
|---|---|---|---|
| recipe_eastbrook_arming_sword | weaponcrafting | bone_fragments x2, linen_scrap x1 | eastbrook_arming_sword |
| recipe_eastbrook_chain_vest | armorcrafting | bone_fragments x3 | eastbrook_chain_vest |
| recipe_eastbrook_wool_trousers | tailoring | linen_scrap x3 | eastbrook_wool_trousers |
| recipe_tanned_leather_jerkin | leatherworking | spider_leg x2, bone_fragments x1 | tanned_leather_jerkin |
| recipe_tough_jerky | cooking | spider_leg x1 | tough_jerky |
| recipe_minor_healing_potion | alchemy | linen_scrap x1, spider_leg x1 | minor_healing_potion |

Inputs reuse the existing harvested-material item ids from #1120/#1121 (bone_fragments/linen_scrap/spider_leg); outputs reuse existing low-tier `BASE_ITEMS` entries rather than adding new item ids, to avoid expanding the positional item-name i18n arrays for this issue.

## Flow

```mermaid
flowchart LR
  A[Recipe content] --> B[Material check]
  B -->|insufficient| X[Deny, no side effects]
  B -->|sufficient| C[Consume reagents]
  C --> D[Quality roll via Rng]
  D --> E[Grant output item]
  E --> F[Grant craft skill +1]

  G[Client: craft intent] -->|craft_item wire cmd| H[Server: sim.craftItem]
  H --> B
  F --> I[craftResult event]
  I -->|mirrored| J[Client: lastCraftResult]
```

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/professions_crafting.test.ts tests/crafting_view.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/command_schema.test.ts tests/keybinds.test.ts tests/localization_fixes.test.ts tests/i18n_completeness.test.ts`
- [x] `npm run build`
- [x] Full `npx vitest run` (green except a pre-existing flaky 5-min timeout test in `tests/loot_master_sim.test.ts` under full-suite load, confirmed unrelated, passes standalone)
- [ ] Screenshot: skipped (best-effort, time-boxed)

Stacked on #1180 (craft skill + craft ring) and #1190 (material rarity roll + gathering/harvest/proficiency/nodes/contracts); diff will shrink substantially once these merge; do not merge this before them.


FernandoX7 (commented):
Nice, clean slice. The pure-core (`crafting_view.ts`) plus thin painter (`crafting_window.ts`) split follows the vendor-window pattern exactly, the resolver sits behind the SimContext seam and stays sim-pure, the quality roll goes through `ctx.rng` on-tick, and `index.html`/`play.html` both get the panel so `/play` is covered. The IWorld surface is real parity, not a stub: `lastCraftResult` genuinely round-trips through the personal `craftResult` event and `applyCraftResultEvent` in `ClientWorld`, and the command/parity pins are literal counts plus sorted-name snapshots rather than self-comparisons. Server authority is intact (client sends intent only; all consume/roll/grant resolve server-side). Reviewed against English only.

A few things to fix.

### Blocking

**KeyI default collides with the existing Event Calendar bind.** `src/game/keybinds.ts:159` binds `crafting` to `KeyI`, but `keybinds.ts:195` already binds `calendar` (Event Calendar) to `KeyI`, and neither sets `allowShared`. For a fresh player with no saved profile, `load()` returns before the dedupe sweep (`keybinds.ts:412`), so both keep `KeyI`, and `edgeActionForCombo` returns the first edge match in `BIND_ACTIONS` order (`keybinds.ts:478-481`), which is `crafting`. Result: pressing I opens crafting and the Event Calendar's default shortcut is dead. For a returning player who had calendar saved on `KeyI`, the second-pass sweep nulls crafting's default instead, so crafting ships unbound. Nothing catches it: `tests/keybinds.test.ts` has no default-uniqueness assertion. Pick an unused default (T, Y, and Z are the free letter keys).

### Should fix

**Failed craft notifies twice, and one copy is unlocalized.** `src/sim/professions/crafting.ts:91-96` emits `ctx.error(..., 'That recipe does not exist.' | 'You do not have the materials for that.')`, which the HUD renders as an error toast (`hud.ts:7627`). But the `craftResult` event with `ok:false` already drives a denial line through the localized catalog keys (`hud.ts:7550-7558`, `hudChrome.crafting.unknownRecipe` / `insufficientMaterials`). So a denied craft prints the same message twice, and the `ctx.error` copy is a raw English literal with no `sim_i18n` matcher, so it stays English in every locale. Simplest fix is to drop the `ctx.error` call and let the already-localized `craftResult` path be the single denial notification. Note the S3 guard does not flag this: `tests/localization_fixes.test.ts:861` builds `simSrc` from a hardcoded file list that does not include `professions/crafting.ts`, so the emit escapes the drift check.

**`marketFilter` is unrelated dead code that rode in on this PR.** `src/sim/sim.ts:805` declares and `:1470` initializes `PlayerMeta.marketFilter`, but nothing in the branch ever reads it, and its comment describes a `MARKET_WIRE_LIMIT` substring-filter behavior that does not exist here. It belongs to the market feature, not crafting. Please drop it from this PR (or land it in its own).

**Test gaps on the roll and failure paths** (`tests/professions_crafting.test.ts`). The consume / all-or-nothing / skill-grant / output / parity behaviors are decisively covered, but three claims are not:
- The "deterministic quality roll" test (`:125`) only asserts two same-seed runs agree with each other, so a change to the `rollMaterialRarity` ladder passes silently. Pin the actual rolled rarity for the fixed seed.
- The "consumes exactly one rng draw" test (`:119`) reads `(sim as any).rng.draws`, which does not exist on `Rng` (so it is `undefined ?? 0`) and is never asserted. Use the `Rng` observer seam and assert one draw on success, zero on denial.
- No test asserts a failed craft grants no craft skill; the skill test only exercises the success path.

**a11y: reagent requirements are only in the hover tooltip.** The craft button's aria-label is just "Craft {name}" (`crafting_window.ts:57`); what a recipe needs, and why the button is disabled, live only in `attachTooltip` (`:62-66`), which keyboard and screen-reader users and mobile (no hover) never reach. Consider folding the reagent state into the accessible name or a visible row.

### Nits

- `crafting_window.ts:58` renders `x${row.resultCount}` as a raw number while the reagent lines use `formatNumber` (`:46-47`); minor formatter inconsistency.
- The partial-consumption test (`:61-77`) only exercises the first-reagent-short arm (short on `reagents[0]`); the held reagent is `reagents[1]`, so a consume-as-you-go regression that bailed after consuming an earlier reagent would not be caught. The upfront all-or-nothing check makes it non-actionable today, but the mirror arm (hold `reagents[0]`, be short on a later one, assert it is untouched) is the one that gives the "partial consumption never happens" claim its teeth.
- `applyCraftResultEvent` casts `ev.quality as MaterialRarity` (`online.ts:2270`): safe since the server only ever emits a `rollMaterialRarity` result, but the widened event type vs the narrowed view type could use a one-line comment.
- `hasRecipeMaterials` (`crafting.ts:49`) checks each reagent entry independently rather than summing by `itemId`. No current recipe repeats an id, so no live bug, but if one ever listed the same reagent twice the per-entry guard could pass while the pooled total is short and the consume loop would over-remove. Worth summing by id or asserting unique reagent ids.

jgyy:
Thanks for the thorough review! Pushed a follow-up commit addressing everything:

**Blocking**
- `crafting` now defaults to `KeyT` instead of `KeyI` (which collided with `calendar`/Event Calendar).

**Should fix**
- Dropped the `ctx.error` denial toast in `craftItem`: the `craftResult` event already drives a localized denial through `hudChrome.crafting.unknownRecipe` / `insufficientMaterials`, so a failed craft no longer prints twice, and there's no more unlocalized literal escaping the S3 guard (there's nothing left in `crafting.ts` for it to catch).
- Removed the unrelated `marketFilter` dead field from `PlayerMeta` (declaration + initializer) that had ridden in on this branch.
- Folded the reagent requirements into the craft button's `aria-label` (reusing the already-localized `resultAria` + `reagentsNeeded` keys) so keyboard/screen-reader/mobile users get that info without depending on the hover tooltip.
- Fixed the raw `x${row.resultCount}` to go through `formatNumber` for consistency with the reagent lines.

**Test gaps**
- The quality-roll test now pins the actual rolled rarity for the fixed seed and asserts rng draws directly via `Rng.setObserver` (one draw on success, zero on denial), instead of only comparing two same-seed runs against each other.
- Added a test asserting a denied craft grants no craft skill.
- Added the mirror partial-consumption case (short on a later reagent while holding the first in full) alongside the existing first-reagent-short test.

Nits on `hasRecipeMaterials` (sum-by-id) and the `online.ts` cast comment weren't touched since they're non-actionable today, per your notes — happy to pick those up too if you'd rather they land now.

Re-ran `tsc --noEmit`, the targeted test list, the full suite (green aside from two pre-existing flaky tests unrelated to this change, confirmed passing standalone), and `npm run build`.

FernandoX7 (approved):

jgyy:
CI is green: PR gate (English-only legal), Format + lint (Biome, changed files), and AI review (Codex) all pass on the follow-up commit. The "Screenshots of changes" job shown as failed was actually cancelled/superseded by a later run (not a real failure). Ready for another look whenever you have time.

PR #1200, opened 2026-07-01
feat(professions): add effect recharge with original-crafter discount (#1137)

### LINKED ISSUE #1137: Professions: effect recharge, with an original-crafter discount
## Problem

A depleted effect (T2) currently has no way to be restored. This issue adds recharging via the production craft that made the effect, or via an Enchanter/Scribe, with the original crafter being cheaper and faster.

## Expected behavior

- A depleted effect can be recharged, consuming materials aligned to the craft that produced it.
- Recharging through the effect's original crafter costs less (time and/or materials) than recharging through a generic Enchanter/Scribe.

## Scope

- `src/sim/professions/tools.ts`: recharge action, cost computation, original-crafter identity check.

## Acceptance criteria

- Recharging the same effect via its original crafter costs strictly less than via a generic Enchanter/Scribe, per a scripted comparison.
- A fully recharged effect returns to its full durability and resumes applying its bonus.

## Notes

Depends on the effect-depletion state from T2 but is a self-contained action; does not require the rarity-scaled consumption curve (T5) or the Always/Prompt toggle (T4).

jgyy:
Closes #1137.

Stacked on #1196 (tool effect slotting/durability, `feature/professions-tool-effects-1136`); diff will shrink once that merges. Do not merge this before it.

## Summary
A slotted tool effect (#1136/#1196) has no way to be restored once depleted. This adds a recharge action: `rechargeEffect(slot, rechargerId, materialsProvided)` in `src/sim/professions/tools.ts`, computing a materials + time cost via `rechargeCost(slot, rechargerId)`.

- `ToolEffectSlot` gets an additive optional `craftedBy?: string` field (set via `slotEffect(effectId, craftedBy?)`), recording who produced the effect via the craft that made it.
- `isOriginalCrafter(slot, rechargerId)` is true only when `craftedBy` is recorded and matches the recharger.
- The original crafter pays half the materials and half the time of a generic Enchanter/Scribe recharge (`ORIGINAL_CRAFTER_DISCOUNT = 0.5`, base cost 4 materials / 30 seconds at 20 Hz).
- On success, durability is restored to the effect's full starting value (same as a fresh `slotEffect`), so `applyEffectBonus` resumes applying the bonus immediately. `craftedBy` is left untouched by a recharge.
- On failure (insufficient materials), the slot is not mutated.

No live call site exists yet for the same reason `applyEffectBonus`/`depleteEffect` don't (#1136's integration note): the node-harvest and recipe-crafting outcome paths (#1121/#1127) aren't on this branch. This module only computes cost and applies the restore; a future production/recharge UI action would deduct `cost.materials` from inventory and gate `cost.ticks` as craft time before calling `rechargeEffect`.

No new UI surface: this issue's scope is the sim-side recharge action, cost computation, and original-crafter identity check (`src/sim/professions/tools.ts`), per the issue text. Skipping a screenshot accordingly.

## Discount formula
```
materials = ceil(4 * discount)
ticks     = ceil(600 * discount)   // 30s at 20 Hz
discount  = 0.5 if rechargerId === slot.craftedBy, else 1
```
So original-crafter recharge = 2 materials / 300 ticks (15s); generic = 4 materials / 600 ticks (30s). Strictly less on both dimensions, always.

## Diagram
```mermaid
flowchart LR
  A[Depleted effect slot, durability = 0] --> B[Recharge request: rechargerId, materials]
  B --> C{isOriginalCrafter?<br/>slot.craftedBy === rechargerId}
  C -->|yes| D[Cost = base x 0.5<br/>materials + time]
  C -->|no, generic Enchanter/Scribe| E[Cost = base x 1<br/>materials + time]
  D --> F{materialsProvided >= cost.materials?}
  E --> F
  F -->|yes| G[Durability restored to full<br/>craftedBy unchanged]
  F -->|no| H[Recharge fails, slot unmutated]
  G --> I[applyEffectBonus resumes applying the bonus]
```

## Test plan
- [x] `npx vitest run tests/professions_tools.test.ts` (new recharge describe block: original-crafter identity, strict cost comparison, generic-crafter-rate fallback when no crafter recorded, successful recharge restores durability + bonus resumes, failed recharge on insufficient materials, generic recharger succeeds at the generic rate)
- [x] `npx vitest run tests/architecture.test.ts` (sim purity)
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npx @biomejs/biome check --write` on changed files


FernandoX7 (commented):
The recharge feature itself is clean. `rechargeCost`/`rechargeEffect`/`isOriginalCrafter` are pure arithmetic with no `Rng`, `Date.now`, or `performance.now`, so determinism and sim purity hold. `craftedBy` is a genuinely additive optional field (`src/sim/professions/tools.ts:93`), and since `ToolEffectSlot` still has no live callsite and is neither persisted nor on the wire, there is no parity or JSONB back-compat work owed yet. The discount is one named fraction driving both dimensions, so the "strictly less on materials and time, always" claim holds for the fixed base (2/300 vs 4/600), and the tests pin it with a real cross-comparison rather than a self-comparison. Good, decisive coverage: full restore, no mutation on failure, and generic-rate fallback when no crafter is recorded.

One should-fix, then a couple of notes.

**should-fix**: this PR re-adds `TOOL_RECIPE_STUBS` (`src/sim/content/professions.ts:236`, the +83 lines), which the base deliberately deleted in `93936218c` ("drop the inert TOOL_RECIPE_STUBS table re-added from the stale #1191 stack", exactly -83 lines from the same file). The table is inert (nothing in `data.ts` or any `SimContext` callback reads it, as the comment itself states), and its `ingredients[].itemId` values (`thorium_ore`, `arcanite_bar`, `ashwood_log`, `elderwood_log`, `goldleaf_herb`, `sunpetal_herb`) reference items that exist nowhere in `content/items.ts`. It is #1135 leftover dragged forward by the stale stack, not part of the #1137 recharge scope, and the base has already had to strip it once. Please drop it from this branch so the merge does not re-revert that cleanup.

**nit**: `slotEffect` (`src/sim/professions/tools.ts:106`) inserts `craftedBy` as a new middle positional param ahead of the existing `confirmMode`. You migrated every call site (all in `tests/professions_tools.test.ts`) and there are no live callers, so nothing breaks. Just flagging the shape: a middle-insert like this silently rebinds any positional `confirmMode` caller, so it is worth a keyword-arg or an object param if this grows a third option.

Everything else in the diff (the em-dash to colon strips in `src/sim/social/chat.ts`, the `/inspect` and `/invite` test copy, and the regenerated `chat_social.json` events hash) is release-merge house-style housekeeping, not part of the feature, and it reads clean. The `inspectReadout` string change is a legitimate house-style fix with the golden regenerated to match.

jgyy:
Thanks for the review, both addressed:

- **should-fix**: dropped the re-added `TOOL_RECIPE_STUBS` table from `src/sim/content/professions.ts` (confirmed nothing in `data.ts` or any `SimContext` callback reads it, and no other file references `TOOL_RECIPE_STUBS`/`ToolRecipeStub`).
- **nit**: `slotEffect` now takes `craftedBy`/`confirmMode` as a keyword options object (`{ craftedBy?, confirmMode? }`) instead of two trailing positional params, so a future third option can't silently rebind an existing positional caller. Updated all call sites in `tests/professions_tools.test.ts` accordingly.

`npx tsc --noEmit`, `npx vitest run tests/professions_tools.test.ts tests/architecture.test.ts` (58 passed), and biome on changed files all green; pushed as a new commit.

FernandoX7 (approved):

PR #1201, opened 2026-07-01
feat(professions): add rarity-scaled effect durability consumption curve (#1139)

### LINKED ISSUE #1139: Professions: rarity-scaled effect durability consumption curve
## Problem

Effect durability (T2) currently, absent this issue, would need to spend a charge on every use regardless of target rarity. This issue adds the rarity-scaled consumption rule so a high-end tool sips charges on trivial targets.

## Expected behavior

- The chance an effect spends a durability charge depends on the tool's rarity versus the target's rarity.
- Against an equal-or-higher-rarity target, a charge is spent every use.
- Against a lower-rarity target, the tool often spends nothing (worked example in the design: an epic tool with +3 quantity spends 1 durability on an epic target, about 60 percent on rare, about 10 percent on common).

## Scope

- `src/sim/professions/tools.ts`: the rarity-scaled consumption roll, replacing the always-spend placeholder from T2.

## Acceptance criteria

- A scripted test over several tool-rarity/target-rarity pairs matches the intended consumption-rate curve within a defined tolerance.
- The roll uses `Rng`.

## Notes

A refinement of T2's depletion mechanic; can land after T2 without changing T2's public surface (same depletion call, different internal rate).

jgyy:
## Summary
Closes #1139.

Adds the rarity-scaled consumption rule for slotted tool effect durability (#1136): whether a use spends a durability charge now depends on the tool's rarity versus the rarity of the target it is used on, instead of a flat per-effect chance.

- `effectConsumptionChance(toolRarity, targetRarity)` in `src/sim/professions/tools.ts`: against an equal-or-higher-rarity target the chance is always 1 (100 percent); against a lower-rarity target it is `max(0.1, 1 - 0.4 * tierGap)`, where `tierGap` is how many rarity tiers the tool outranks the target by on the standard ladder (common, uncommon, rare, epic, legendary).
- `depleteEffect(slot, toolRarity, targetRarity, rng)` now draws exactly one `Rng.chance` roll against that formula (previously it drew against a flat per-effect `depletionChance`, which this PR removes from `ToolEffectDef`/`TOOL_EFFECTS` since it is superseded by the rarity-scaled curve).
- Worked-example fit: epic tool vs epic target = 100 percent, epic vs rare = 60 percent, epic vs common = 10 percent (exactly the numbers in the issue).

### Formula and table
```mermaid
flowchart LR
  A["Tool rarity\n(ItemDef.quality)"] --> C["tierGap = toolTier - targetTier"]
  B["Target rarity\n(rollMaterialRarity result)"] --> C
  C --> D{"tierGap <= 0 ?"}
  D -- yes --> E["consumption chance = 1.0 (100%)"]
  D -- no --> F["consumption chance = max(0.1, 1 - 0.4 * tierGap)"]
  E --> G["Rng.chance(chance)"]
  F --> G
  G -- true --> H["durability -= 1"]
  G -- false --> I["durability unchanged"]
```

Worked example (epic tool):

| Target rarity | tierGap | consumption chance |
|---|---|---|
| epic (or higher) | 0 | 100% |
| rare | 1 | 60% |
| uncommon | 2 | 20% |
| common | 3 | 10% (floored) |

## Stacked PR note
Stacked on #1196 (tool effects, #1136) and #1190 (rarity roll, #1122); diff will shrink once these merge. **Do not merge this before them.**

## No new UI surface
This is a pure sim-logic change (`src/sim/professions/tools.ts` + its content table); no render/UI/HUD surface, so no screenshot.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/professions_effect_consumption.test.ts tests/professions_tools.test.ts tests/professions.test.ts tests/professions_gathering.test.ts tests/professions_contracts.test.ts tests/professions_rarity_roll.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/snapshots.test.ts` (all green)
- [x] New `tests/professions_effect_consumption.test.ts`: exact-formula assertions for the worked example and floor/monotonicity, plus scripted statistical trials (20,000 draws each, fresh slot per trial) confirming epic-vs-epic ~100%, epic-vs-rare ~60%, epic-vs-common ~10%, each within +/-10 percentage points; a determinism check (same seed, same sequence) and a manual `Rng.next()` mirror confirming exactly one `Rng` draw per call (never `Math.random`)
- [x] `npm run build`


FernandoX7 (commented):
This is clean, well-tested work. The curve matches the issue's worked example exactly (epic/epic 100%, epic/rare 60%, epic/uncommon 20%, epic/common floored to 10%), and the roll stays deterministic: exactly one `rng.chance` draw per call, and it always draws even at durability 0, so the depletion sequence is independent of remaining charges. The new logic stays pure (a type-only `MaterialRarity` import from a sibling sim module, no `Math.random` or wall-clock), and since `MaterialRarity` is exactly the five ladder tiers, `RARITY_ORDER.indexOf` never returns -1. The test file is thorough (exact-formula, floor, monotonicity, equal-or-higher-always-consumes, 20k-trial statistical, and a single-draw determinism mirror); all 37 tests pass locally, and the em-dash cleanups in `chat.ts` / `global_invite.test.ts` are correct.

should-fix
- `src/sim/content/professions.ts:102` (and the three `0.5` values at :114, :124, :134): `depletionChance` is now dead. `depleteEffect` no longer reads it (the rarity curve replaced it), but the field is still declared on `ToolEffectDef` and populated on all three `TOOL_EFFECTS`. The PR description says this change "removes from `ToolEffectDef`/`TOOL_EFFECTS`", but it does not. Nothing else reads the field and no test pins it, so either delete it and its three values, or fix the description.

nit
- `effectConsumptionChance` / `depleteEffect` / `resolveToolEffectUse` type `toolRarity` as `MaterialRarity`, which excludes `'poor'` and `undefined`, but the docstrings describe `toolRarity` as the base tool's `ItemDef.quality` (which can be either). A base tool with no quality set cannot flow in without the future caller defaulting it, and a `'poor'` cast would make `rarityTierIndex` return -1. Types prevent it today and integration is deferred to #1121/#1127, so this is just a forward note: the integration comment could say how an unqualified tool maps onto the ladder (e.g. default to `'common'`).

The recipe-stub table and the tool-effect-slotting comment in `professions.ts` are inherited from the stacked #1135/#1136 work, not this change; diff noise until those merge.

jgyy:
Thanks for the review.

- Removed the dead `depletionChance` field from `ToolEffectDef` and its three `0.5` values in `TOOL_EFFECTS` (`src/sim/content/professions.ts`), matching the PR description. Also cleaned up an adjacent stale comment that still referenced the old flat per-effect depletion chance instead of the rarity-scaled curve.
- Noted the `MaterialRarity`/quality-widening point for the future integration comment; left as-is for now since it's deferred to #1121/#1127 per your note.

Also rebased onto the current `release/v0.22.0` to resolve merge conflicts (an integration-note comment in `tools.ts` that needed reconciling with the `rechargeEffect` note added upstream, a comma/colon nit in `chat.ts`, and a stale two-arg `slotEffect`/`resolveToolEffectUse` call shape in `professions_tools.test.ts`). All 9 suites from the test plan plus the full repo test suite, `tsc --noEmit`, and `npm run build` are green.

FernandoX7 (approved):

PR #1203, opened 2026-07-01
feat(professions): tiered mastery gating for craft skill progress

### LINKED ISSUE #1128: Professions: tiered mastery gating (at-tier vs below-tier crafting)
## Problem

Crafting a recipe currently (per P3) has no notion of the player's tier capability. This issue adds the rule that at-or-above-tier crafting is what advances skill, while crafting below capability gives diminishing returns and two-tiers-under gives nothing.

## Expected behavior

- Crafting at or above a player's current tier capability in a craft is what grants skill progress.
- Crafting one tier below capability gives reduced skill progress; two or more tiers below gives none.
- Common tier remains a free floor regardless of capability.

## Scope

- `src/sim/professions/wheel.ts`: tier-capability tracking and the diminishing-returns curve, layered onto the flat skill state from P2.
- `src/sim/professions/crafting.ts`: read tier capability when resolving a craft.

## Acceptance criteria

- A player crafting at their tier gains full skill progress; one tier below gains reduced progress; two tiers below gains zero.
- Common-tier crafting is unaffected regardless of the player's capability.

## Notes

Builds on P2 (skill state) and P3 (crafting action) but is a self-contained rule layered on top; does not require the conserved-mass wheel (P5) to exist.

jgyy:
Closes #1128.

Adds tiered mastery gating on top of the flat craft skill state (#1126) and common-tier crafting resolution (#1127): crafting at or above a player's tier capability in a craft grants full skill progress, crafting one tier below capability grants reduced progress, and two or more tiers below grants nothing. Common tier remains a free floor regardless of capability.

## Tier capability mapping

A player's tier capability in a craft is derived from their flat skill value in that craft, bucketed every 25 points:

- skill 0-24: tier 0 (common)
- skill 25-49: tier 1
- skill 50-74: tier 2
- skill 75-99: tier 3
- ... and so on

A recipe's tier is derived the same way from its `skillReq`.

## Progress multiplier curve

```mermaid
flowchart LR
    A[Recipe tier vs player tier capability] --> B{recipe tier == 0?}
    B -- yes, common tier --> F[Full progress, ALWAYS regardless of capability]
    B -- no --> C{capability - recipe tier}
    C -- "<= 0 (at or above capability)" --> D[Full progress]
    C -- "== 1 (one tier below)" --> E[Reduced progress, 0.5x]
    C -- ">= 2 (two+ tiers below)" --> Z[No progress, 0]
```

## Implementation

- `src/sim/professions/wheel.ts`: `TIER_SKILL_STEP` (25), `tierForSkill`, `tierCapability`, `tierProgressMultiplier` (the pure tier-bucket + progress-curve functions), layered onto the existing flat `CraftSkills` state from #1126.
- `src/sim/professions/crafting.ts`: `resolveCraft` now reads the player's tier capability and the recipe's tier when granting skill on a successful craft, scaling the flat 1-point gain by the tier multiplier. Extracted `resolveCraftForRecipe(ctx, pid, recipe)` (takes a recipe record directly rather than an id) so the tier curve is testable against a synthetic higher-tier recipe without needing actual higher-tier content yet (recipes.ts is still common-tier only per #1127's scope).

## Tests

- `tests/professions_skill.test.ts`: unit tests for `tierForSkill`, `tierCapability`, and `tierProgressMultiplier` covering the bucket boundaries and the full/reduced/zero curve.
- `tests/professions_crafting.test.ts`: integration tests via `resolveCraftForRecipe` covering at-tier (full), one-tier-below (reduced, strictly between 0 and full), two-tiers-below (zero), and common-tier crafting always granting its full floor at both zero and high player capability.

No new UI surface (a sim-only mechanic), so no screenshot.

## Stacked PR note

Stacked on #1197 (recipe crafting, which itself includes craft skill/ring/rarity-roll/gathering/contracts). The diff will shrink once these merge. Do not merge this before them.


FernandoX7 (commented):
The tier-gating core is clean and does exactly what #1128 describes. `tierForSkill` / `tierCapability` / `tierProgressMultiplier` are pure, deterministic, and sim-pure, the boundaries are right (0/24 to tier 0, 25 to tier 1, negatives clamp to 0), and `gainCraftSkill` no-ops on the zero multiplier while accepting the 0.5 reduced gain. Because every shipped recipe is `skillReq: 0` (tier 0, always full), the mechanic is inert on current content and only activates once higher-tier recipes land, which is the right way to stage it. Tests are thorough (38 pass locally, plus the 266 parity/snapshot pins), and the `resolveCraft` to `resolveCraftForRecipe` split keeps the curve testable without shipping synthetic content.

Two things rode in on the release merge that look like stale artifacts rather than part of #1128:

- **should-fix** `src/sim/sim.ts:926`: `worldBossDaily?: { date: string; looted: string[] }` is re-declared on `CharacterState` directly under the comments (919-924, 1580-1582) that say the field was intentionally dropped and is not restored (world-boss lockouts moved to `raidLockouts`). Nothing reads or writes it anywhere in the tree, and the new comment's claim that "addPlayer falls back to an empty record" is not true (addPlayer never sets it). This reintroduces persisted state that the base deliberately migrated away from, so safest to drop it.

- **should-fix** `src/sim/sim.ts:805` (plus the `marketFilter: ''` init at 1475): `PlayerMeta.marketFilter` is declared and initialized but has no reader in the branch. The two `dataset.marketFilterMenu` / `marketFilterOption` hits in `market_window.ts` are unrelated client-side DOM attributes, not this field. The comment describes a server-side substring filter whose consumer does not exist here, so it is a half-landed field: drop it, or land the code that reads it.

- **nit** `server/CLAUDE.md:81`: the delta-key count goes 25 to 27, but the sentence pins that number to `ALL_DELTA_KEYS`, which this PR takes to 31 (was 30). The doc figure does not match its own cited pin; 31 is the number.

The `gprof` / `gatheringProficiency` wiring (#1119) that comes along is correct and complete: `gatheringProficiencyFor` on the server, the `gatheringProficiency` getter on `Sim`, the `ClientWorld` field, the round-trip through `gprof`, and the parity/snapshot pins all line up. It is just extra scope relative to the title.

jgyy:
Thanks for the review! Pushed a fix addressing all three points:

- Dropped the stale `worldBossDaily` field on `CharacterState` (dead, superseded by `raidLockouts`) and its unused init/restore comments.
- Dropped the dead `marketFilter` field on `PlayerMeta` (no reader anywhere; the described server-side substring filter was never wired up).
- Fixed the delta-key count in `server/CLAUDE.md` (27 -> 31, matching the actual `ALL_DELTA_KEYS` pin in `tests/snapshots.test.ts`).

`tsc` and the relevant test files (snapshots, professions_skill, professions_crafting) all pass.

FernandoX7 (approved):

Issue #107, opened 2026-06-13
Add professions, crafting, and skilling loops


maxpolaczuk:
## Reported from Discord
- `#feature-requests`, 2026-06-13T03:54:38Z: professions and crafting would be nice.
- `#feature-requests`, 2026-06-13T08:38:54Z: integrate skilling such as fishing and woodcutting.
- `#feature-requests`, 2026-06-13T19:57:50Z: another request specifically called out fishing.

## Problem
The game does not yet have professions/skilling/crafting loops.

## Expected
- Scope gathering skills such as fishing and woodcutting.
- Scope crafting/professions and how outputs interact with combat/economy.
- Consider progression, UI, recipes, resources, and server persistence.

## Coverage check
No existing PR directly adds professions, crafting, fishing, or woodcutting.

jgyy:
Tracked by open PRs #611 (First Aid, Skinning, Leatherworking, Tailoring) and #933 (gathering + crafting: Mining, Herbalism, Blacksmithing, Alchemy). Leaving open until one lands.

FernandoX7:
Originally proposed plan by @Nervescraper https://woc.nervemart.com/docs/professions-system

Below are my thoughts:

---

# Professions System: Design Review and First-Slice Recommendation

Review of the high-level [Professions System manifesto](https://woc.nervemart.com/docs/professions-system),
grounded against the current codebase (`release/v0.17.0`). The goal is buy-in on the
vision plus the handful of decisions worth locking before the work is spread across
contributors. This is not a line edit of the manifesto.

Date: 2026-06-30

## TL;DR

This is one of the strongest design docs we have had, and the vision deserves buy-in.
The thesis is sharp and falsifiable (a specialist beats a generalist by construction via
the conserved-mass production wheel, with all growth routed off-wheel into additive
gathering), and the architecture maps cleanly onto our real seams (SimContext modules,
content-as-code via `data.ts`, additive JSONB persistence, IWorld-first). It also brings
its own guardrails: the fragility section and mirror theorem are exactly what stops a
professions system from rotting over time.

One principle is already decided and worth stating up front: professions are accessible from
level 1 everywhere, not gated behind level 20 or maxed talents. What curves as the zones get
higher is depth and content, not access. The work should also be structured for parallel
contribution: define the shared schemas and the IWorld facet first, then each skill, recipe,
and node is an independent additive task in its own module, so multiple people can build at
once without stepping on each other.

[...]

Issue #1152, opened 2026-07-01
Epic: Professions system (gathering, crafting, archetypes)


jgyy:
## Problem

The professions system (design: https://woc.nervemart.com/docs/professions-system) is tracked as 32 independently buildable issues (#1119-#1150) after the original 6 coarse slices were closed and re-split for independence via stubbing rather than strict zero-dependency ordering. There is no single place to see the whole system's status at a glance.

## Scope

This epic tracks the full professions build-out. It does not itself contain implementation work; sub-issues are the unit of work and should be closed independently as they land.

Each sub-issue is stubbed to build independently, but the list below is also the real dependency order: build low tiers first, or expect to stub around a later tier's output. Each sub-issue carries a short "Depends on / Blocks" note (added as a comment) for the same reason.

### Tier 0: no dependencies, start anytime
- #1164 Professions contracts and IWorld facet (skill/craft/recipe/node schemas, wire/persistence keys) -- new, build first
- #1165 Item-instance model (signer, per-effect charges, rolled quality/stats, boundTo) -- new, go/no-go decision plus the slice if go
- #1119 Gathering profession skill tracking (Mining, Logging, Herbalism) -- builds on #1164's contracts
- #1120 World-spawn node content and placement (ore, wood, herb) -- builds on #1164's contracts
- #1125 Ten-craft content definitions (ring order and poles) -- builds on #1164's contracts
- #1140 Monster component-type tagging (content) -- builds on #1164's contracts

### Tier 1
- #1121 Per-player node harvest and respawn timer -- needs #1119, #1120
- #1123 Vendor-sold base gathering tools with tier gating -- needs #1119
- #1126 Flat per-craft skill tracking -- needs #1125
- #1141 Single-use corpse harvest (first-come contention) -- needs #1140

### Tier 2
- #1122 Material rarity roll scaled by gathering proficiency -- needs #1119, #1121
- #1124 Gathering HUD (node and proficiency display) -- needs #1119, #1121
- #1135 Base tool tiers, infinite durability, and material tier gating -- needs #1125, #1123
- #1142 Per-corpse focus picker (concentrate vs spread tradeoff) -- needs #1141

### Tier 3
- #1127 Recipe content and a basic crafting action -- needs #1125, #1126, #1122
- #1136 Tool effect slotting with durability and depletion -- needs #1135

### Tier 4
- #1128 Tiered mastery gating (at-tier vs below-tier crafting) -- needs #1126, #1127
[...]

Issue #1149, opened 2026-07-01
Professions: adventuring xp from observed item use (Battlefield Experience)


jgyy:
## Problem

The design floats a later-phase, additive-only xp channel: watching a crafted item used in combat (a potion drunk, a killing blow, armor worn) grants a small trickle of xp to the craft that made it. Committed as a concept, deliberately sequenced for later: it should land only after the core professions loop (gathering, production core, tools, Monster Harvesting) is built and proven, and should not compete with that work.

## Expected behavior

- Observing specific tracked events (potion drunk, killing blow landed, armor worn at a kill) involving an item grants a small skill-progress trickle to the craft that made that item.
- This channel only ever adds skill; it never drains another craft.
- Gated to rare tier and above so it cannot trivialize common/uncommon progress.
- The original crafter of a specific item benefits most when later improving that same item.

## Scope

- Design spec and prototype only in this issue: observation scope (server interest radius, party/raid weighting) and per-item attribution.
- No production implementation until the core professions issues (gathering, production core, tools, Monster Harvesting) are built and proven.

## Acceptance criteria

- A written spec exists for observation scope, attribution, and the rare-tier gate before any implementation begins.
- If prototyped, the channel measurably never reduces any craft's skill.

## Notes

Sequencing, not uncertainty: this is a later-phase lane by design and should not be picked up before the core slices land.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Dependency note (epic #1152): Depends on: #1127. Blocks: none. (Exploratory, not committed.)

jgyy:
Per the #107 follow-up, this is promoted from exploratory to committed: it is the "Battlefield Experience" lane, one of the two levers (alongside Monster Harvesting) that keeps pure Combat-Level-1 skillers capped below adventurer-crafters. Title updated to drop the "(exploratory)" marker.

Also: pin any observation-radius reference in the spec to `INTEREST_RADIUS = 90` (the real player/event interest radius) rather than 120 - 120 is the NPC-only radius.

jgyy:
## Design spec: Battlefield Experience (observed-item-use adventuring xp)

Per this issue's own scope, this is spec only. No production implementation lands until the core professions slices (gathering, production core, tools, Monster Harvesting) are built and proven. The items below reference real, currently-open mechanisms in this tracker so a future implementer has concrete hooks, not abstract prose.

### a) Observation scope

[...]

PR #1205, opened 2026-07-01
Professions: active archetype state and quest-gated switching

### LINKED ISSUE #1129: Professions: active archetype state, quest-gated switching, and per-character switch count
### Problem

Skill is additive per craft (#1126), with no notion of which crafts are empowered above the common floor. The design gates the top of crafting behind a single active archetype: epic and legendary are reachable only by an active specialty's majors, and you change which archetype is active through a quest, never by losing skill. This issue adds that state and the quest-gated switch. Supersedes the retired conserved-mass drain model this issue originally described.

### Expected behavior

- A character has at most one active archetype at a time: an adjacent pair (the two majors, empowered to the top tiers) plus one opposite craft as a hobby empowered to rare. A character with no adjacent high pair attunes as Jack of All Trades instead (title/build in #1130 and the Jack issue).
- Before a first attunement, any craft can climb its tier quests as far as rare; epic and legendary are reachable only by an active specialty's majors. Once attuned, every craft outside the active specialty caps at the common floor, dormant, with its learned knowledge retained in full.
- Switching is additive-safe: a previously active specialty goes dormant at the floor and is restored to its retained level when re-activated. No skill is ever lost.
- The first attunement into any given archetype (whether or not another was held before) is a lore quest. Returning to a previously held archetype is a repeatable "make amends" quest whose cost escalates with a per-character switch count.
- Common tier is a free floor, unaffected by the active archetype; dormant crafts still produce at the floor.

### Scope

- `src/sim/professions/wheel.ts`: active-archetype state, the empowerment gate (reachable ceiling = min(tier-quest cap, active-archetype cap)), per-character switch count.
- `src/sim/professions/crafting.ts`: read the empowerment gate when resolving a craft.
- `IWorld`: active archetype + switch-count read surface (facet from #1164).
- Persistence: active archetype and switch count, additive back-compat JSONB.

### Acceptance criteria

- For an attuned character in a scripted state, only the active specialty's crafts exceed the common floor (majors past rare, hobby to rare); others cap at common with knowledge retained.
- For an unattuned character, any craft climbs to rare via tier quests and none past rare.
- Switching to a previously held archetype restores its dormant knowledge exactly.
- The per-character switch count increments on each return switch.
- Common-tier crafting is unaffected regardless of active archetype.

### Notes

The empowerment ceiling here and the tier-quest ceiling (#1128) are the "two identical causes" of a cap in the design doc's tier ladder section ("The tier ladder and the at-most-two cap"). The switch-quest CONTENT and the escalation curve are separate issues and depend on a craft/gather quest-objective type that does not exist yet. Depends on #1126, #1127, #1164. Decided in the doc revision: a dormant craft still produces at the common floor while capped (no longer a TBD).

jgyy:
Closes #1129

## This PR previously implemented the wrong design

The top commit used to implement a conserved-mass budget / opposite-craft-drain
model. Per a comment on #1129 (referencing decision #107), that design was
superseded: knowledge across all ten crafts stays purely additive (no drain),
and archetype identity comes from a single active archetype the player swaps
via quest (a zone-1 acceptance lore quest to first pick one, and a repeatable,
escalating "make amends" quest to switch later, tracked via a per-character
switch count). This PR reverts the conserved-mass commit and replaces it with
the corrected scope.

Also now depends on #1164 (contracts/IWorld facet), which is already present on
this branch's base.

Stacked on #1197 (recipe crafting, which itself includes craft skill/ring/
rarity-roll/gathering/contracts); the diff will shrink once these merge. Do not
merge this before them.

## State machine

```mermaid
stateDiagram-v2
    [*] --> NoArchetype
    NoArchetype --> ArchetypeX: acceptance quest complete, choose craft X
    ArchetypeX --> ArchetypeX: attempt switch to Y, make-amends quest NOT complete (blocked, no-op)
    ArchetypeX --> ArchetypeY: make-amends quest complete, choose Y (switchCount += 1)
    note right of ArchetypeY
        Required amends progress/cost
        scales with switchCount
        (5 + 3 per prior switch)
    end note
```

## What changed

- `src/sim/professions/archetype.ts` (new): the full state machine behind the
  `SimContext` seam. `acceptArchetypeQuest` (first pick only, no-op after),
  `advanceAmendsProgress` + `switchArchetype` (gated switch, escalating
  requirement), persistence helpers.
- `src/sim/sim.ts`: `PlayerMeta.archetype` / `CharacterState.archetype` fields,
  persistence wiring, and the `IWorld` facade delegates.
- `src/world_api/professions.ts`: extends `IWorldProfessions` with
  `activeArchetype`, `archetypeSwitchCount`, `archetypeAmendsProgress`,
  `archetypeAmendsRequired`, `acceptArchetypeQuest`, `advanceAmendsProgress`,
  `switchArchetype`.
- `src/net/online.ts`: `ClientWorld` implements the same facet as a documented
  not-yet-wired-on-the-wire stub, matching the existing
  `craftSkills`/`gatheringProficiency` convention already in that file.
- `src/sim/content/zone1.ts` + `src/ui/world_entity_i18n.ts`: two new quest
  content stubs, `q_archetype_acceptance` and `q_prof_make_amends` (see
  "What was stubbed" below).
- `tests/world_api_parity.test.ts`: pinned member count 155 -> 162.
- `tests/parity/golden/*.json`: regenerated (new `PlayerMeta.archetype` field
  in the trace).
- `tests/professions_archetype.test.ts` (new): the acceptance-criteria tests.
- `src/sim/professions/wheel.ts`, `src/sim/professions/crafting.ts`: reverted
  to byte-identical with `origin/feature/professions-recipe-crafting-1127`
  (confirmed via an empty `git diff`).

## What was stubbed, and why

The two quests (`q_archetype_acceptance`, the zone-1 acceptance lore quest, and
`q_prof_make_amends`, the repeatable "make amends" quest) are content stubs:
they reuse an existing giver/turn-in NPC (`marshal_redbrook`) and a trivial
existing-mob kill objective purely so the `QuestDef` is valid content, not
because that NPC/mob has any lore tie to professions. They are NOT wired into
the generic quest accept/turn-in flow (`quest_commands.ts`): the existing quest
engine is one-time only (turning in a quest adds it to `questsDone` with no
re-accept path), so a genuinely repeatable turn-in for the "make amends" quest
is its own follow-up. Real dialogue/UX authoring for both quests is out of
scope here.

The state machine itself does NOT depend on this content being completable in
play: `acceptArchetypeQuest`/`advanceAmendsProgress`/`switchArchetype` are the
direct sim entry points a future quest-completion hook calls, and they are
fully implemented, tested, and gated correctly independent of the quest
content stub.

## Test plan

- [x] a. Completing the acceptance quest for the first time sets the active
      archetype to the chosen craft (`tests/professions_archetype.test.ts`, "a")
- [x] b. Attempting to switch without completing the make-amends quest is
      blocked/no-op: archetype, switchCount, and amends progress all unchanged
      (test "b")
- [x] c. Completing the amends quest then switching increments switchCount by
      exactly 1 and changes the active archetype (test "c")
- [x] d. The amends requirement escalates with switchCount: two switchCount
      values give two increasing required values (test "d")
- [x] e. A switch never mutates any of the ten craft skill values, asserted
      bit-for-bit before/after (test "e")
- [x] `tests/architecture.test.ts` (sim purity) green
- [x] `tests/world_api_parity.test.ts` green with the updated pin
- [x] `tests/localization_fixes.test.ts` (S3 i18n guard) green


jgyy:
Per the epic #1152 PR-fleet reconciliation: this PR correctly scopes to just the archetype state machine (crafting.ts/wheel.ts stay byte-identical to #1127, since higher-tier gating is #1128's scope). But that means the empowerment ceiling from #1129's revised acceptance criteria (reachable ceiling = min(#1128's tier capability, archetype-derived capability)) doesn't exist anywhere yet on this branch or on #1203. Added a TODO(#1129/#1203) in archetype.ts naming the exact composition point, mirroring the TODO(#1149/#1205) already in PR #1281's battlefield_xp.ts, so it does not silently slip once #1203 lands. tsc clean, tests/professions_archetype.test.ts green, guard tests green.

FernandoX7 (commented):
The state machine in `src/sim/professions/archetype.ts` is clean: pure, deterministic, all three entry points correctly gated, and it never touches `craftSkills` (which the revised #107 scope requires). Persistence is fully wired (init in `addPlayer`, `normalizeArchetypeState` on load with back-compat, shallow copy on save), the `IWorldProfessions` facet is implemented on both `Sim` and the documented `ClientWorld` stub, and the parity pin (181 / 49 / 132 plus the seven new members) and golden traces are updated to match. The acceptance tests are decisive: (e) in particular seeds every craft with a distinct nonzero value before the bit-for-bit compare, so it would actually catch a regression.

Two things before this is mergeable.

**blocking**

- `package-lock.json` is out of sync with `package.json`, so `npm ci` fails and reddens all four CI checks (Codex, Biome, PR gate, Screenshots) at the install step: `Missing: picomatch@4.0.5 from lock file`. The PR's only lockfile change vs base is a 15-line deletion of the `node_modules/svelte-check/node_modules/picomatch` entry (base carries it at 4.0.4), which looks like a merge-conflict resolution that dropped the block. Regenerate with `npm install` and commit the lockfile.

**should-fix**

- The two stub quests are wired into a live giver and are completable dead ends. `q_archetype_acceptance` and `q_prof_make_amends` are added to `marshal_redbrook.questIds` (`src/sim/content/zone1.ts:535`) and `ZONE1_QUEST_ORDER`; with no `requiresQuest` / `minLevel` / `retired`, `computeQuestState` returns `available` from level 1, so a player can accept, kill one or two Forest Wolves, and turn them in for 100 and 50 XP through the normal engine. Turn-in never calls `acceptArchetypeQuest` / `switchArchetype` (no hook exists), so "A Craft to Call Your Own" prints its completion line "Your path is chosen; walk it well." while no archetype is actually set. That is misleading live content plus free XP, and it contradicts the PR's own "not wired into the generic accept/turn-in flow" note (`questIds` is exactly that wiring). Gate them out of the live world until the turn-in hook lands (a `retired: true` on both, or drop them from `questIds` and `ZONE1_QUEST_ORDER`), keeping the `QuestDef` records for the follow-up.

**nits**

- The tests cover the five acceptance criteria but not a few branches the module still has: `advanceAmendsProgress` as a no-op before any archetype is set, `switchArchetype` to the current craft returning false, and a `normalizeArchetypeState` save/load round-trip. Worth a line each.
- The PR description says the parity pin went 155 to 162; the merged pin is actually 174 to 181 (the release merges moved the base out from under it). Just a stale description, the code pins are right.

The design itself, additive-only knowledge with a single active craft swapped through an escalating gate, reads correctly against the revised #107 scope, and the `TODO(#1129/#1203)` for the empowerment ceiling is the right call: it flags the composition point rather than guessing a number.

jgyy:

- Regenerated the lockfile: the `svelte-check/node_modules/picomatch@4.0.5` entry was missing (dropped in a merge-conflict resolution). Reproduced the `npm ci` EUSAGE failure locally with npm 10 (matching CI's node 22 setup), regenerated with `npm install`, confirmed `npm ci` now succeeds clean.
- Marked both `q_archetype_acceptance` and `q_prof_make_amends` as `retired: true` so `computeQuestState` reports them `unavailable` and they can't be accepted/turned in through the live engine (matching how the codebase already gates unfinished quest content). They stay in `marshal_redbrook.questIds`/`ZONE1_QUEST_ORDER` per the existing `retired` convention, but are no longer reachable until the turn-in hook wiring lands.

`npx tsc --noEmit`, `tests/professions_archetype.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/localization_fixes.test.ts`, and `tests/quests.test.ts` all green; biome clean on the changed file.

Left the nits (extra branch coverage tests, stale PR description pin numbers) as-is for now since they're non-blocking, happy to pick those up too if you'd like them in this PR.

FernandoX7 (approved):

PR #1206, opened 2026-07-01
test(professions): verify World Market integration for gathered and crafted items

### LINKED ISSUE #1146: Professions: World Market integration for gathered and crafted items

jgyy:
## Summary

Closes #1146.

Stacked on #1197 (crafting/gathering/contracts/rarity-roll) and #1191 (tool tiers/vendor tools/craft ring); diff will shrink once these merge; do not merge this before them.

Read `src/sim/market.ts` and `InvSlot` (`src/sim/types.ts`) first: `InvSlot` carries only `itemId`/`count`, no instance payload anywhere in this repo today, so there is no fungible-vs-instanced split for profession items to trip over (the issue's stated risk from a hypothetical #1165 signed-materials stamp does not exist in the current codebase). Gathered materials (`bone_fragments`, etc.), crafted items from the recipe system (`eastbrook_arming_sword`, `eastbrook_wool_trousers`, etc.), and gathering tools (`copper_mining_pick`, etc.) are all plain `ItemDef`-backed item ids merged into `ITEMS` via `data.ts`, exactly like any other item, so they already flow through `marketList`/`marketBuy`/`marketCancel`/`loadMarket` with zero special-casing needed.

This PR adds `tests/professions_market.test.ts`, an integration-style regression suite covering:
- Listing a gathered material fully and buying it: exact quantities, no dupe, no loss.
- Listing and buying a crafted/vendor gathering tool.
- Listing and cancelling (reclaiming) a crafted weapon: escrow round trip.
- A profession-item variant of the existing `market.test.ts` persistence round-trip test (mixed gathered/crafted/tool listings survive save/load, house stock is never duplicated, sale proceeds carry across correctly).
- A profession-item variant of the existing `save_character_and_market.test.ts` atomic-write test: an escrowed gathered-material listing is written to `characters` and `world_state` in one transaction, so an immediate disconnect right after listing can never tear the escrow (vaporize or duplicate the item).

No production code changes: everything already worked via the existing generic item-id path, so the issue is satisfied by these regression tests alone.

A note on the two other historical bugs named in the issue's scope, since I verified both against the ACTUAL merge-base for this PR rather than assuming they're fixed:
- **Realm scoping** (`market:<realm>` world_state key): this fix exists on a separate, unmerged branch (`fix/market-realm-scoping`) and is not yet in `release/v0.18.0`. It is realm/persistence-only and orthogonal to profession items (any item id hits the same single `world_state` row today); not a professions regression, and out of scope for this PR to fix.
- **Unknown-item-drop-on-load**: same story, a separate unmerged fix. Profession items are all registered `ItemDef` entries via `data.ts`, so they never hit the dropped-unknown-id path in normal play.

Both are pre-existing, general market gaps unrelated specifically to gathered/crafted/tool items, so I left them alone per the "don't invent unnecessary code changes" instruction and did not merge those unrelated branches into this stack.

## Diagram

```mermaid
flowchart LR
    A[Gathered material / crafted item / tool] --> B[Generic ItemDef, itemId+count via data.ts]
    B --> C[market.ts: marketList escrow]
    C --> D[market.ts: marketBuy / marketCancel]
    D --> E[Realm-scoped world_state row: characters + market in ONE transaction]
    E --> F[No dupe / no loss: verified across save-load round trip and immediate-disconnect save-on-leave]
```

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/professions_market.test.ts tests/market.test.ts tests/save_character_and_market.test.ts tests/architecture.test.ts` (all green)
- [x] `npm test` (full suite green except two pre-existing unrelated flakes: `tests/loot_master_sim.test.ts` timing timeouts, not touched by this change)
- [x] `npm run build`
- [x] `npx @biomejs/biome check --write tests/professions_market.test.ts` (warnings only, no errors; matches repo convention)
- [x] No new UI surface; skipped screenshot per issue instructions (no bug found needing visual confirmation)


FernandoX7 (commented):
Clean regression suite. The net diff is one new test file, no production code, and the "no code change needed" framing holds up: gathered materials, crafted items, and gathering tools are all plain `ItemDef` ids, so they ride `marketList`/`marketBuy`/`marketCancel`/`loadMarket` with no special-casing. I checked the load-bearing bits against the real code and ran the file:

- All four item ids (`bone_fragments`, `copper_mining_pick`, `eastbrook_arming_sword`, `eastbrook_wool_trousers`) and `the_merchant` (`market: true`) exist, so `standAtMerchant` sits the player at a real auctioneer.
- The assertions are decisive, not decorative. `errorsSince` is a genuine guard (`ctx.error` really emits `{ type: 'error' }` onto `sim.events`, sim.ts:6113), the 5% cut is pinned to literals on both paths (190 in-memory, 95 across the save/load round trip via `MARKET_CUT = 0.05`), house stock is checked for non-duplication, and the save-on-leave test pins BEGIN, COMMIT, no ROLLBACK, `UPDATE characters`, and `world_state` on one client, plus that `state.inventory` no longer holds the escrowed stack (a real field on `CharacterState`, so that check bites).
- `saveCharacterAndMarketState` five-arg signature and the `openMarketWriteGate` gate usage match the sibling `save_character_and_market.test.ts`.
- `npx vitest run tests/professions_market.test.ts`: 5 passed.

One nit:

- `nit` professions_market.test.ts:154 reaches into the private `marketCollections` via `as unknown as { ... }`. It is justified here (sim2 has no live player for the seller pid, so the public `marketInfoFor` cannot resolve the key), but a one-line comment saying that would spare the next reader the "why not the public API" pause, since the first test does use `marketInfoFor`.

Nothing blocking.

jgyy:
Addressed the nit: added a one-line comment above the private \`marketCollections\` reach-through explaining why \`marketInfoFor\` can't be used here (sim2 has no live player for the seller pid to resolve the key against).

Confirmed no merge conflicts against the current \`release/v0.22.0\` and pushed. \`npx vitest run tests/professions_market.test.ts\` still 5 passed, and the pre-push QA floor (tsc, guard tests, biome, copy scan) is green.

FernandoX7 (approved):

PR #1207, opened 2026-07-01
feat(professions): signed materials and self-gathered crafting bonus

### LINKED ISSUE #1145: Professions: signed materials and the self-gathered crafting bonus
## Problem

A rare-or-better monster material gathered by a player currently carries no gatherer identity, and using a material a player gathered themselves grants no crafting bonus.

## Expected behavior

- Gathering a rare-or-better monster material stamps the resulting item with the gatherer's name.
- Using a signed material in a craft you performed yourself (gatherer equals crafter) grants a bonus: either a small item-level lift or a reduced required quantity of that material.
- Using someone else's signed material grants no bonus.

## Scope

- Item instance data: a gatherer-name field on rare+ monster materials.
- `src/sim/professions/crafting.ts`: check gatherer identity against the crafting player and apply the bonus.

## Acceptance criteria

- A rare+ monster material gathered by a player is visibly stamped with that player's name.
- Crafting with a self-gathered signed material yields the described bonus; crafting with the same material signed by someone else does not.

## Notes

The stamping half depends on M2's harvest resolution existing; the crafting-bonus half depends on P3's crafting path existing. The two halves (stamping, bonus-on-craft) could even be split further if desired, but are kept together here since neither is useful alone.

jgyy:
Closes #1145.

## Summary

- A rare-or-better monster material from corpse harvest (skinning/salvage) is now stamped with the harvester's own name: `harvestCorpse` (`src/sim/interaction.ts`) rolls the material's rarity (`rollCorpseMaterialRarity`, reusing the shared `rollMaterialRarity` ladder off a fixed baseline "power" input, since corpse harvest has no per-player gathering proficiency counter the way mining/logging/herbalism do) and grants a signed, non-fungible instance via `ctx.addItemInstance` (the `signer` field from #1174's `ItemInstancePayload`) whenever the roll lands rare, epic, or legendary. Anything below that stays a plain fungible grant, unchanged from before this issue.
- Crafting with a material you signed yourself grants a bonus. **Chosen bonus type: a reduced required quantity** (one fewer unit of that reagent), not an item-level/quality lift. Reasoning: `crafting.ts` already resolves the materials check as a per-reagent "do I have count N" predicate, so lowering N by one for a self-signed reagent composes directly with the existing check-then-consume flow with no new side channel. An item-level/quality lift would instead have to reach into the shared `rollMaterialRarity` call that gathering.ts's own comment already documents as informational-only for the common-tier crafting path (it does not yet re-roll per-reagent quality).
- Using a material signed by **someone else** (signer set but not the crafting player's own name) is **not** counted: it behaves exactly like a plain unsigned material, no bonus. `CraftResult` now carries `selfSignedBonusApplied` so a caller/test can observe whether the bonus fired.
- `addItemInstance` (added by #1174 directly on `Sim`) had no `SimContext` binding yet; this PR appends it to the seam (`src/sim/sim_context.ts` + the `buildSimContext` binding in `src/sim/sim.ts`) so `harvestCorpse` (which lives behind `SimContext`, not on `Sim` directly) can reach it.

## Mermaid

```mermaid
flowchart TD
    A[Mob dies, corpse harvestable] --> B{componentTags mapped to an item?}
    B -- no --> Z[No item granted]
    B -- yes --> C[Roll material rarity<br/>rollCorpseMaterialRarity]
    C --> D{rarity >= rare?}
    D -- yes --> E[ctx.addItemInstance<br/>signer = harvester name]
    D -- no --> F[ctx.addItem<br/>plain fungible stack]
    E --> G[Later: craft attempt<br/>consumes this material]
    F --> G
    G --> H{Reagent slot has instance.signer?}
    H -- no signer --> I[Normal required count, no bonus]
    H -- signer set --> J{signer == crafting player's name?}
    J -- no, someone else's --> I
    J -- yes, self-signed --> K[Required count for that reagent -1<br/>selfSignedBonusApplied = true]
```

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/corpse_harvest_sim.test.ts tests/professions_crafting.test.ts tests/item_instance.test.ts tests/professions_gathering.test.ts tests/sim_context.test.ts tests/entity_roster.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/command_schema.test.ts tests/localization_fixes.test.ts` all green
- [x] Full `npx vitest run` green (2 `loot_master_sim.test.ts` failures were a default 5000ms timeout on a long tick loop under parallel-suite load, not a regression from this change; they pass with a longer per-test timeout and are unrelated to professions/item-instance code)
- [x] `npm run build` green
- [x] `npx @biomejs/biome ci --changed` green (no format diffs; only pre-existing-style lint *warnings*, which CI does not gate on)
- [x] New tests: rare+ corpse harvest stamps the signer name (and a below-rare harvest stays unsigned); a self-signed material reduces the required quantity and is consumed; a material signed by a *different* player grants no bonus; a plain unsigned material grants no bonus

## Stacked PR note

Stacked on #1197 (crafting/gathering/contracts), #1181 (corpse harvest + component tags), and #1174 (item-instance model with signer field). This diff will shrink once those merge. **Do not merge this before them.**


FernandoX7 (commented):
The feature itself is clean and well scoped. `rollCorpseMaterialRarity` reuses the shared rarity ladder off a fixed baseline, the signed-instance grant goes through the `SimContext` seam correctly (callback declared in `sim_context.ts` and bound in `sim.ts`), and the crafting bonus composes with the existing per-reagent check-then-consume flow instead of adding a side channel. One rng draw per yield keeps `harvestCorpse` deterministic. The four new tests are decisive: they pin the signer stamp, the single-count instance, the below-rare fungible quantity, and all three no-bonus cases (different signer, unsigned, self-signed). I ran the four touched test files at head, 59 pass.

One blocking issue, and it is collateral merge damage rather than the feature:

**blocking** `src/sim/sim.ts:1532-1533`: two stray reassignments overwrite the correct load.
```
meta.gatheringProficiency = normalizeGatheringProficiency(s.gatheringProficiency ?? s.professions); // correct
meta.gatheringProficiency = normalizeGatheringProficiency(s.professions);                            // dead
meta.gatheringProficiency = normalizeGatheringProficiency(s.gatheringProficiency);                   // wins, drops the fallback
```
The last line drops the `?? s.professions` fallback that the comment three lines above explicitly promises ("fall back to the legacy one so saves from before the rename still load correctly"). Net effect: a pre-rename save that stored proficiency under `professions` with no `gatheringProficiency` key now loads at all-zero, silently wiping it. Base (`release/v0.23.0`) has only the single correct assignment, so this PR introduces the two extra lines, and they read like the same stale `checkout --ours` dedupe fallout the earlier commits on this branch were chasing. Fix is to delete both added lines.

Worth knowing the existing tests do not catch this: the round-trip test in `professions_gathering.test.ts` writes both keys (`serializeCharacter` mirrors into `professions` and `gatheringProficiency`), and the "old save" test deletes `professions` from an all-zero state while keeping `gatheringProficiency`, so neither exercises the fallback with real data.

nit `src/net/online.ts:1009`: the `lastCraftResult` field is just relocated a few lines up with no behavior change. Harmless, but it is unrelated churn in this diff; worth dropping if you re-roll the branch.

Once the two dead lines in `sim.ts` are gone this looks good.

FernandoX7 (commented):
The seam work here is clean: `addItemInstance` is declared on `SimContextCallbacks`, bound in `createSimContext`, wired to `Sim.addItemInstance` in `buildSimContext`, stubbed in the `entity_roster` mock, and pinned in `sim_context.test.ts` CALLBACK_KEYS, so the callback surface stays consistent across all four spots. The reduced-required-quantity choice for the bonus is well reasoned and composes with the existing per-reagent check-then-consume flow without a new side channel, and the new tests cover stamp / no-stamp and self / other / unsigned. The corpse rarity roll is a single deterministic `ctx.rng` draw per yielded component.

A few things before this is ready.

**blocking**

`src/sim/sim.ts:1532-1533`: two dead re-assignments overwrite the correct back-compat load.

```
meta.gatheringProficiency = normalizeGatheringProficiency(
  s.gatheringProficiency ?? s.professions,
);
meta.gatheringProficiency = normalizeGatheringProficiency(s.professions);          // added
meta.gatheringProficiency = normalizeGatheringProficiency(s.gatheringProficiency); // added
```

The final value is `normalizeGatheringProficiency(s.gatheringProficiency)`, which drops the `?? s.professions` legacy fallback the first line exists to preserve. `normalizeGatheringProficiency(undefined)` returns all zeros, so a pre-rename save (has `s.professions`, no `s.gatheringProficiency`) now loads with its gathering proficiency wiped. This is rebase fallout, unrelated to #1145, and no test covers the legacy-field load path so the suite stays green. Delete both added lines and keep only the original single assignment.

**should-fix**

The crafting-bonus half of #1145 is unreachable in real play. A signed instance is only ever produced by corpse harvest, which grants `boar_hide` / `wolf_fang` / `webwood_silk` / `widow_venom_sac` (`HARVEST_COMPONENT_ITEMS`), but none of the six `COMMON_RECIPES` consumes any of those: every reagent is `bone_fragments`, `linen_scrap`, or `spider_leg` (node-harvest items that are never signed). The `professions_crafting` tests only fire the bonus by artificially `addItemInstance('bone_fragments', ...)`, which cannot happen in game. So the sim plumbing is correct but a player never sees the bonus until a recipe consumes a corpse material. Either add such a recipe or note the content wiring is deferred to a follow-up.

`src/sim/professions/crafting.ts:112-114`: when a reagent's `count` is 1 and the crafter holds a self-signed instance, `requiredCountFor` returns 0, and `if (required > 0) ctx.removeItem(...)` then consumes nothing while the signed material stays in the bag. `selfSignedBonusApplied` still fires, so a recipe needing exactly one of a signable material would craft repeatedly off a single retained instance. It is latent behind the finding above (no current recipe hits it), but it goes live the moment a corpse material is added to a count-1 recipe. Consider flooring the consumed amount at 1, or consuming the signed instance as part of the bonus.

**nit**

`src/sim/professions/crafting.ts:56`: the bonus keys on `s.instance?.signer === meta.name`. That matches #1174's name-typed `signer`, so it is consistent, but a character rename would silently stop recognizing materials the player harvested pre-rename, and two identically named characters would share the bonus. Not this PR's model to change, just worth knowing when rename and this feature meet.

One non-issue for the record: the `online.ts` move of the `lastCraftResult` declaration is an inert reorder (same field, same default, no duplicate), and the extra `ctx.rng` draw in `harvestCorpse` is deterministic, so the regenerated goldens are the expected churn.

jgyy:
Thanks for the thorough review! Pushed a follow-up commit (12fd66a) addressing both passes:

**blocking** - Deleted the two dead reassignments at `sim.ts:1532-1533`. They were rebase/checkout-fallout from an earlier `--ours` dedupe on this branch, not intentional: the single correct `s.gatheringProficiency ?? s.professions` assignment now stands alone again, so a pre-rename save loads its legacy `professions` data instead of getting silently wiped. Added a regression test (`professions_gathering.test.ts`) that constructs a *genuine* legacy save (real proficiency data under `professions`, `gatheringProficiency` key absent entirely) and asserts it survives the load - confirmed it fails without the fix and passes with it, since neither existing test exercised that exact shape.

**should-fix: bonus unreachable in real play** - Agreed, and going with the deferred-to-follow-up option you flagged rather than inventing recipe balance in this PR: none of the six common-tier recipes currently consume a corpse-harvest item (`boar_hide`/`wolf_fang`/`webwood_silk`/`widow_venom_sac`), so the sim plumbing for the bonus is correct but not yet reachable through normal play. Wiring an actual recipe to one of those materials is a content/balance call better made as its own follow-up issue rather than bundled into this seam-and-mechanics PR.

**should-fix: count-1 recipe latent bug** - Fixed. `requiredCountFor` now floors at 1 instead of 0 (`Math.max(1, reagent.count - 1)`), so a self-signed instance is always consumed on craft rather than retained forever when the base requirement is exactly 1. Added a test that crafts `recipe_tough_jerky` (spider_leg x1) with a self-signed spider_leg: confirms the bonus does *not* apply (can't reduce below 1), the material is consumed, and a second craft attempt correctly fails.

**nit: `online.ts` reorder** - Left as is; it's inert churn as noted, and I'd rather not touch unrelated lines in a branch that's already stacked three deep.

All 10 named test files plus the two touched test files are green, `tsc --noEmit` clean, `biome ci --changed` clean (no format diffs, pre-existing warnings only per repo convention), and the pre-push QA floor passed on push.

FernandoX7 (approved):

Issue #1131, opened 2026-07-01
Professions: hobby carve-out (protected opposite craft to rare)


jgyy:
## Problem

The wheel's drain rule (P5) as stated fully drains the opposite craft. The design has one carve-out: a single opposite craft (the hobby) stays protected up to rare while the rest of the opposite side falls to the floor. This issue adds that exception.

## Expected behavior

- A player's designated hobby (one opposite craft) is protected from drain up to the rare tier.
- The other opposite craft (not chosen as hobby) still drains to the floor as usual.
- Working the other opposite craft instead lets the current hobby decay and the new one grow, with no fee or menu.

## Scope

- `src/sim/professions/wheel.ts`: hobby designation and its protected-drain exception, layered on P5.

## Acceptance criteria

- A craft designated as hobby does not drop below its protected floor while its majors are worked, up to rare.
- Switching which opposite craft is worked correctly swaps which one is protected, with the old one now decaying.

## Notes

A targeted exception to P5's drain rule; does not require P6 (title grant) to exist, though the two are usually experienced together.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Dependency note (epic #1152): Depends on: #1129. Blocks: none.

jgyy:
Per the #107 decision (see the correction on #1129), there is no conserved-mass drain in the current model, so this issue's premise (a carve-out protecting one opposite craft from drain) no longer applies. Knowledge is additive-only: every craft you have ever learned stays intact automatically, with no protection mechanic needed.

Recommending this issue be closed as superseded. Leaving that decision to the epic owner rather than closing unilaterally.

jgyy:
Closing as superseded per the rationale already recorded above: the #107 decision replaced the conserved-mass drain model, so there is no drain for a hobby carve-out to protect against. Knowledge is additive-only in the current model.

PR #1209, opened 2026-07-01
feat(professions): add combo recipes requiring an adjacent craft pair

### LINKED ISSUE #1132: Professions: combo-specific recipes exclusive to an adjacent pair
## Problem

There are no recipes that require both crafts of a specific adjacent pair (signature lines exclusive to one archetype). This issue adds that recipe requirement type on top of the recipe/crafting system from #1127.

## Expected behavior

- Some recipes require attunement to a specific adjacent-pair archetype (per #1129), not merely raw skill in both crafts: with additive knowledge, a Jack of All Trades can hold all ten crafts at a mid tier and an unattuned crafter can hold two adjacent rares, so a raw dual-skill check would let both hold combo recipes the design reserves for the attuned pair.
- A player not currently attuned to the recipe's required pair cannot craft it, regardless of how high their individual craft skills are.
- The hobby craft (protected only to rare, not one of the attuned pair's two majors) never unlocks a combo recipe.

## Scope

- `src/sim/content/recipes.ts`: a required-adjacent-pair-archetype field on relevant recipes.
- `src/sim/professions/crafting.ts`: check current active-archetype attunement (#1129) against the recipe's required pair before allowing the craft.

## Acceptance criteria

- A combo recipe cannot be crafted by a player who is not currently attuned to its required adjacent pair, even if they hold both crafts at high skill.
- A player whose only qualifying craft is their hobby (not one of the attuned pair's majors) cannot craft the combo recipe.
- No common-tier combo recipes are authored (attunement itself happens at rare, so a common-tier character could never satisfy the requirement anyway).

## Notes

Depends on #1127 (recipe/crafting path), #1129 (active-archetype attunement state, which this recipe check reads instead of raw craft/tier), and the hobby-exclusion criterion's home issue (previously #1131, closed under the retired drain model; its empowerment-model replacement is being filed fresh). Does not require archetype titles (#1130) to exist, since it checks attunement state directly, not the title.

jgyy:
Closes #1132

Stacked on #1197 (recipe crafting, ring data, craft skill); diff will shrink once that merges; do not merge this before it.

## What

Adds a dual-craft-requirement type to `ProfessionRecipeRecord` (`comboRequirement: { craftA, craftB, minTier }`) so some recipes are exclusive to one specific adjacent pair on the craft ring, on top of the recipe/crafting system from #1127. `crafting.ts`'s new `meetsComboRequirement` checks the player's tier capability (`wheel.ts` `tierCapability`) in BOTH named crafts before allowing the craft, independent of the recipe's `professionId`. A player missing either required craft at the required tier is denied with `combo_requirement_unmet`, no side effect at all (same no-partial-consumption guarantee as the existing reagent check).

Three example recipes, using pairs confirmed adjacent by calling `adjacentCrafts()` myself (not assumed):
- `recipe_ironbound_warplate_helm` and `recipe_forgeguard_bulwark_shoulders`: armorcrafting + weaponcrafting (both Material pole)
- `recipe_volatile_flux_elixir`: alchemy + engineering (both Experimental pole)

All three are tier 1 (skillReq 25, minTier 1).

## #1128 (tier-mastery, PR #1203)

Merged cleanly into this branch (see the merge commit at the base of this PR's history) so the combo-requirement check reuses `tierForSkill`/`tierCapability` from `wheel.ts` instead of duplicating a threshold check. If #1203 lands first, this merge commit becomes a no-op on rebase.

## Hobby craft wording (#1131 closed as superseded)

The issue text says "the hobby craft ... never unlocks a combo recipe," but #1131's hobby/drain-craft distinction was closed as superseded, so there is no hobby-vs-major craft split left in the sim. I interpreted this narrowly per the issue's own guidance: only the two crafts named in `comboRequirement` ever count, full stop. A player's skill in any OTHER craft, however high, never substitutes for either `craftA` or `craftB`, regardless of whether that other craft would once have been called a "hobby." Tests cover this directly: a player with a very high unrelated craft (`cooking`) and zero skill in either required craft is denied, and a player missing just one of the two required crafts is denied even with the other very high plus a high unrelated craft.

## Mermaid

```mermaid
flowchart TD
    A[Player attempts combo recipe] --> B{Has comboRequirement?}
    B -- No --> Z[Normal reagent/skillReq check only]
    B -- Yes --> C{craftA tier >= minTier?}
    C -- No --> D[Denied: combo_requirement_unmet]
    C -- Yes --> E{craftB tier >= minTier?}
    E -- No --> D
    E -- Yes --> F[Both satisfied: proceed to reagent check]
    F --> G[Craftable]
    H[Unrelated craft skill, however high] -.never substitutes.-> C
    H -.never substitutes.-> E
```

## Testing

- New Vitest coverage in `tests/professions_crafting.test.ts`: every combo recipe carries a `comboRequirement`; a player with both required crafts at/above minTier can craft; missing craftB denies even with craftA and an unrelated craft very high; a player whose only high skill is unrelated (`cooking`) is denied; one craft one point below minTier still denies; the `Sim.craftItem` command surface denies correctly.
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/professions_crafting.test.ts tests/professions_skill.test.ts tests/architecture.test.ts`: all passing.
- `npm run i18n:gen` + `node scripts/i18n_resolved_hash.mjs --write` re-run after adding the new `hudChrome.crafting.comboRequirementUnmet` key (with its five non-Latin M16 fills), so `tests/i18n_completeness.test.ts` / `tests/i18n_resolved_equivalence.test.ts` / `tests/i18n_status_registry.test.ts` are green.
- `npm run build`: succeeds.
- Full `npx vitest run`: green except two pre-existing `tests/loot_master_sim.test.ts` timeouts, confirmed unrelated by re-running that file in isolation both with and without this diff (both pass standalone; the timeouts only show up under full-suite parallel contention).

## No UI surface

No new UI surface: this is a sim/content + crafting-resolution change only. Screenshot skipped per the issue scope (`src/sim/content/recipes.ts`, `src/sim/professions/crafting.ts`); the only client-visible change is the existing craft-denial toast now also covering the new `combo_requirement_unmet` reason via the pre-existing `craftResult` event/toast plumbing.


jgyy:
CI status: PR gate (English-only legal) passes; Release gate is skip-gated as expected for a non-release-tier PR.

The "Format + lint (Biome, changed files)" job fails, but not from anything in this PR's own diff. It reports 3 `lint/correctness/noUnusedPrivateClassMembers` errors in `src/sim/sim.ts` at lines that are pre-existing thin delegates from the stacked #1127 branch this PR is built on (e.g. `partyLootCandidatesForMob`, explicitly commented as kept because `dead_party_loot.test.ts` reaches it via cast; similarly `consumeHealAbsorb`/`rangedSwing`). I confirmed these same 3 findings already exist identically in PR #1197's own CI run (same lines, same rule), where they were reported as warnings, not errors, because `biome ci --changed --since=<base>` escalates a diagnostic to error severity only when it falls inside the diff hunk against the base ref. Since `release/v0.18.0` has moved forward since #1197 branched (547b3740 -> 94f7a77a on upstream), the "changed hunk" window against the now-current base has shifted to include these pre-existing lines, upgrading them from warning to error even though this PR never touches them (confirmed via `git diff origin/feature/professions-recipe-crafting-1127 HEAD -- src/sim/sim.ts`, which is empty).

I deliberately did not delete these methods to silence the lint: they're documented as intentionally-kept thin delegates for other tests/callers, and removing them would be an out-of-scope, risky change unrelated to #1132. This should resolve on its own once #1197/#1203 merge into `release/v0.18.0` and this branch is rebased onto the merged base, per the stacking note in the PR description.

FernandoX7 (commented):
Nicely scoped change. The combo gate lands in the right place (`meetsComboRequirement` behind the SimContext seam, pure and deterministic), the denial is a clean early return with no side effect so the no-partial-consumption guarantee holds, and putting it before any rng draw or mutation keeps the draw order intact. The new `combo_requirement_unmet` reason is widened consistently across the `SimEvent` union, `CraftResultView`, and the HUD toast, and `applyCraftResultEvent` mirrors it as a raw string so it round-trips over the wire with no encoder change. Tests are decisive: the 24-vs-25 boundary, the no-side-effect assertions on a denial, and the "unrelated craft never substitutes" cases all pin real behavior. Result and reagent items all exist.

One thing worth resolving before merge:

**should-fix** `recipeList` never includes the combo recipes, so they and the new toast are unreachable in normal play. `Sim.recipeList` (src/sim/sim.ts:5151) and `ClientWorld.recipeList` (src/net/online.ts:1009) both return `COMMON_RECIPES` only, and the crafting window is built from `recipeList` (src/ui/hud.ts:9677), whose `craftable` flag is reagent-only with no craft-skill input. So `onCraft` only ever fires for common recipes; the three combo recipes resolve solely via a direct `craft_item` by id, which no client path sends. That means the `combo_requirement_unmet` toast the PR calls "the only client-visible change" can't actually be triggered by a player, and the example recipes are dead content. If surfacing them is deliberately deferred to a later UI issue (buildCraftingView would need the player's craft skills to gate the row), that is reasonable, but then say so in the PR and soften the "client-visible" note. Otherwise they need to be added to `recipeList`.

**nit** src/sim/content/recipes.ts:125: `recipe_forgeguard_bulwark_shoulders` produces `gravewyrm_gauntlets`, which is a `slot: 'gloves'` item, so the recipe name promises shoulders but yields gloves. The module comment also lists only `boundstone_helm` and `elixir_of_the_bear` as the reused outputs and omits this one. Cosmetic, but reconcile the name or the item.

**nit** src/sim/professions/crafting.ts:17: the module comment says "`hasComboRequirement` checks the player's tier capability" but the function is named `meetsComboRequirement`.

jgyy:
Thanks for the review, addressed all three points:

**should-fix (reachability):** `recipeList` (both `Sim.recipeList` and `ClientWorld.recipeList`) now serves `ALL_RECIPES` (common tier plus combo), exported from `content/recipes.ts`, instead of `COMMON_RECIPES` only. The three combo recipes now show up in the crafting window like any other recipe.

Since a combo recipe can now actually reach the client, `buildCraftingView` also takes the player's `craftSkills` and greys out the Craft button when `comboRequirement` is unmet (checked via `tierCapability` for both named crafts), instead of enabling it on reagents alone and letting the server deny it with a toast the player didn't expect. `hud.ts` passes `this.sim.craftSkills` through.

**nit (name/item mismatch):** `recipe_forgeguard_bulwark_shoulders` (which produces the `gloves`-slot `gravewyrm_gauntlets`) renamed to `recipe_forgeguard_bulwark_gauntlets`; the module comment's output list updated to match.

**nit (comment/function name):** the `crafting.ts` header comment now says `meetsComboRequirement`, matching the actual function name.

Added test coverage in `tests/crafting_view.test.ts` (combo gate craftable/not-craftable, unrelated-craft-never-substitutes, no-comboRequirement path) and updated the `recipeList` assertion in `tests/professions_crafting.test.ts` to expect common + combo.

Verified: `npx tsc --noEmit` clean; `npx vitest run tests/professions_crafting.test.ts tests/crafting_view.test.ts tests/world_api_parity.test.ts tests/architecture.test.ts` all green; pushed and the pre-push QA floor (tsc, guard tests, biome, copy scan) passed.

FernandoX7 (approved):

Issue #1130, opened 2026-07-01
Professions: archetype title grant from the active archetype


jgyy:
### Problem

There is no title granted from a character's crafting identity. Once the active archetype exists (#1129), this issue reads it and grants the corresponding title.

### Expected behavior

- When a character attunes to an adjacent-pair archetype (per #1129), grant the corresponding named archetype title (one of the ten).
- When a character's build is the breadth build (no adjacent pair to attune), grant Jack of All Trades instead, not an error state.
- The title is granted at the attunement step, not derived passively from raw skill levels.

### Scope

- `src/sim/professions/wheel.ts`: title derivation from the active archetype.
- `IWorld`: current title read surface.
- Client: display the granted title.

### Acceptance criteria

- Attuning to a named adjacent pair grants the correct named title.
- The breadth build grants Jack of All Trades.
- No title is granted before attunement.

### Notes

Pure read/derivation over the active-archetype state from #1129. The title offer happens when two adjacent crafts reach rare (the attunement point), and Jack of All Trades is itself an attunement (the breadth attunement), not a default state. Jack's breadth ceiling and improviser perks are a separate issue. Depends on #1126, #1129.

Full design: https://woc.nervemart.com/docs/professions-system (republished 2026-07-02, matches the PR #1285 revision)

jgyy:
## Title grant flow

```mermaid
flowchart TD
    Top2[Player two highest crafts reach rare] --> Adj{Adjacent on the ring?}
    Adj -->|Yes| Arch[Grant named archetype title, e.g. Smith]
    Adj -->|No| Jack[Grant Jack of All Trades]
    Arch --> Hobby[Reveal hobby: one opposite craft of the majors pair, protected to rare]
```

Covers the two-issue split: this issue (#1130) is the title grant/derivation itself; #1131 is the hobby protection mechanic the archetype path reveals.

jgyy:
Dependency note (epic #1152): Depends on: #1126, #1129. Blocks: #1134, #1143.

jgyy:
Per the #107 decision (see also the correction on #1129), there is no longer a passive "two highest crafts" derivation. Re-scope this issue to grant a title from the player's currently-active archetype, i.e. whatever the #1129 quest-switch mechanic has set as active, not a passive derivation from craft levels.

Nervescraper:
[...]

PR #1210, opened 2026-07-01
feat(professions): add specialization perks (material discount, recharge discount, mobile station stub) (#1134)

### LINKED ISSUE #1134: Professions: specialization perks (resource discount, mobile crafting station, recharge discount)
## Problem

Pushing far into a craft region currently unlocks nothing beyond raw output. This issue adds the perk layer: reduced material cost, a mobile crafting station, and cheaper recharges on tools the player crafted.

## Expected behavior

- A sufficiently specialized craft (per a threshold on the wheel state) reduces the material cost of recipes in that craft.
- A specialized player can set up a temporary mobile crafting station in the field.
- Recharging an effect on a tool through its original crafter (from tools slice) costs less when that crafter is also sufficiently specialized.

## Scope

- `src/sim/content/professions.ts`: perk thresholds and effects.
- `src/sim/professions/wheel.ts`: perk-eligibility read.
- `src/sim/professions/crafting.ts`: apply material-cost reduction.
- Client: mobile crafting station placement/UI.

## Acceptance criteria

- A player above a perk's threshold sees the correct material-cost reduction when crafting in that craft.
- A mobile crafting station can be placed and used to craft while away from a town crafting station.
- Perk thresholds are read from content, not hardcoded in logic.

## Notes

A pure additive bonus layer on top of the crafting path (P3) and wheel (P5); does not block on archetype titles (P6) or combos (P8), though perk flavor text may reference the archetype name once P6 exists.

jgyy:
Closes #1134.

Stacked on #1200 (effect recharge/tool tiers/craft ring/vendor tools/gathering); diff will shrink once that merges; do not merge this before it.

## Summary

Adds the perk layer on top of the craft wheel described in #1134: a sufficiently specialized craft reduces recipe material cost, a specialized crafter gets an additional discount recharging effects on tools they made, and a specialized player can place a temporary mobile crafting station (currently inert, see stub note below).

- `src/sim/content/professions.ts`: new `PERK_THRESHOLDS: Record<string, PerkThresholdDef>`, one entry per craft on `CRAFT_RING`, each with `specializedSkillThreshold` (75), `materialDiscountPct` (0.2), `rechargeDiscountPct` (0.25). `ToolEffectDef` gains a `craftId` field (all three starter tool effects are `'enchanting'`). Also `MOBILE_CRAFTING_STATION_DURATION_TICKS` (10 minutes).
- `src/sim/professions/wheel.ts` (new): pure perk-eligibility reads over an explicit `CraftSkillState` record: `isSpecialized`, `materialCostMultiplier`, `rechargeDiscountMultiplier`. Thresholds/percents come from content, never hardcoded.
- `src/sim/professions/crafting.ts` (new): `resolveCraft`/`discountedIngredients`, the minimal viable slice of a recipe-crafting resolver. Applies the material discount per ingredient: `Math.max(1, Math.floor(qty * multiplier))` (floor, minimum 1, documented rounding rule).
- `src/sim/professions/tools.ts`: `rechargeCost`/`rechargeEffect` gain an optional `rechargerSkills` param. When the recharger is the original crafter AND specialized in the effect's craft, an additional discount COMPOSES (multiplies) on top of the existing #1137 original-crafter discount, never replacing it. Omitting the param behaves exactly like before #1134.
- `src/sim/professions/mobile_station.ts` (new): `placeMobileCraftingStation`/`isStationActive`, gated on `isSpecialized`.
- `tests/professions_perks.test.ts` (new): specialized vs non-specialized material cost, resolveCraft success/failure, specialized-recharge strictly-better-than-plain-original assertion, non-original recharger gets no perk, and mobile station placement/gating/duration.

## Reconciliation notes / what was stubbed and why

- **Neither #1127 (the general crafting action) nor #1128 (the wheel's per-player skill state) have landed on any branch this stacks on.** `wheel.ts` and `crafting.ts` therefore take a plain `CraftSkillState` record (craftId -> skill 0-100) as an explicit, caller-owned parameter, mirroring the `ToolEffectSlot` "caller owns the state" shape #1136/#1137 already established in `tools.ts`. Once #1128's real skill state lands on `Sim`/`PlayerMeta`, its call sites plug straight into these functions with no change to this module's logic. `crafting.ts`'s `resolveCraft` is explicitly the minimal viable slice of what #1127 needs: once #1127 lands, this discount logic should MOVE (not be copied) into the real resolver.
- **Mobile crafting station is explicitly stubbed pending an unfiled town/crafting-station build-out prerequisite** (see epic #1152's Tier 6 notes). There is currently NO "must be near a town crafting station" gate anywhere in the engine for a bypass mechanic to unlock, so `mobile_station.ts` implements a real, tested, INERT-for-now mechanic: it can be placed (gated on `isSpecialized`), has a fixed duration, and is queryable via `isStationActive`. It has no effect on `resolveCraft` today because there is no location gate to relax there. No Sim/IWorld/client wiring was added for the same reason: there is no live per-player skill state on `Sim` yet to gate a real command against, so a live command surface would have nothing meaningful to check. Client placement UI is therefore skipped in this PR pending both prerequisites; will follow once #1128 (skill state) and the town build-out land.
- Perk flavor text does not yet reference an archetype name since #1130 (title grant) has not landed; this is intentional per the issue's own notes ("does not block on archetype titles").
- Thresholds/discount percents (75 / 20% / 25%) are placeholders pending maintainer confirmation against the design doc (same 403-Forbidden caveat already noted above `CRAFT_RING` in `professions.ts`), kept uniform across all ten crafts so no craft is silently favored.

## Mermaid

```mermaid
flowchart LR
  A[Player skill in a craft >= specializedSkillThreshold] -->|isSpecialized| B{Specialized?}
  B -- yes --> C[materialCostMultiplier < 1]
  B -- yes --> D[rechargeDiscountMultiplier < 1, composes with original-crafter discount]
  B -- yes --> E[placeMobileCraftingStation succeeds]
  B -- no --> F[Full material cost, generic recharge rate, station placement blocked]
  C --> G[resolveCraft: discounted ingredient qty, floor, min 1]
  D --> H[rechargeCost: original-crafter discount x specialization discount]
  E --> I[MobileCraftingStation: placed, has duration, queryable via isStationActive]
  I -.->|STUB: no town/crafting-station proximity gate exists yet| J[No gameplay effect on resolveCraft today]
```

## Test plan
- [x] `npx vitest run tests/professions_perks.test.ts tests/professions_tools.test.ts tests/architecture.test.ts` all green
- [x] `npx @biomejs/biome check --write` on changed files only
- [ ] CI (watching)


FernandoX7 (commented):
The perk layer is well structured. `wheel.ts`, `crafting.ts`, `tools.ts`, and `mobile_station.ts` stay pure and host-agnostic, the thresholds and percents live in `PERK_THRESHOLDS` content instead of in logic, the recharge discount composes (multiplies) onto the original-crafter discount rather than replacing it, and the stub notes are honest about what is inert and why. The optional `rechargerSkills` param keeps the #1137 call sites unchanged, and dropping the now-dead `if (required > 0)` guard around `removeItem` is a correct cleanup since `requiredReagentCount` floors at 1. `npx vitest run` on professions_perks / professions_crafting / professions_tools / architecture is green (96 tests), and the diff leaves IWorld parity untouched.

A few things to fix.

**should-fix** `src/sim/professions/crafting.ts:149`: `selfSignedBonusApplied` now conflates the #1134 specialization discount with the #1145 self-signed reduction. The flag is set from `required < reagent.count`, but `requiredReagentCount` folds the specialization material discount into `required`, so a specialized crafter with zero self-signed materials reports `selfSignedBonusApplied: true`. The field is documented and tested to mean a self-gathered signed instance lowered a reagent's count. Repro: a specialized (enchanting at 75) crafter holding 8 `bone_fragments` + 1 `linen_scrap` and no signed instances crafts the perk test's synthetic recipe and comes back `selfSignedBonusApplied === true`. Nothing reads the field yet (grep across the tree), so it is not player-visible today, but it will lie the moment a self-gathered-bonus surface consumes it. Derive the flag from the self-signed step alone (compare `afterSelfSigned < reagent.count`) rather than from the composed `required`.

**should-fix** `src/sim/content/professions.ts:103`: `depletionChance` is a new required field on `ToolEffectDef` (set to `0.5` on all three effects at :124/:135/:146) that nothing reads. `depleteEffect` uses the rarity-gap consumption curve (#1139), and a whole-tree grep finds no other consumer. It reads like a rebase orphan, and it muddies the depletion story since it implies a chance-based model that does not exist. Drop it, or wire it if it is meant to replace the rarity-gap roll.

**nit** `src/sim/professions/tools.ts:277` and `wheel.ts:167`: with the placeholder constants the recharge material discount is inert. `ceil(4 * 0.5) = 2` and `ceil(4 * 0.5 * 0.75) = ceil(1.5) = 2`, so a specialized original recharger pays the same 2 materials as a plain original one and only the tick cost drops (300 to 225). The perk test correctly weakens materials to `toBeLessThanOrEqual`, but the `rechargeDiscountMultiplier` doc says the specialized recharger "always pays strictly less." Either soften the doc or note the material half only bites once the base cost is larger. Fine to leave until real balance numbers land, but worth a comment so it is not later mistaken for a bug.

Minor note: `requiredReagentCount` now routes through `craftById` (via `materialCostMultiplier` to `isSpecialized` to `thresholdFor`), which throws on an off-ring `professionId`. Every shipping recipe is a ring craft, so this is safe today, but the old `requiredCountFor` degraded silently there; a future non-ring recipe id would now crash the craft instead of just skipping the discount.

jgyy:
Thanks for the thorough review. Addressed both should-fix items and the nit in 29c74f795:

- `selfSignedBonusApplied` now derives from the self-signed step alone (`afterSelfSigned < reagent.count`) instead of the composed `required`, so a specialized crafter with no self-signed materials no longer misreports the flag as true. `requiredReagentCount` returns `{ count, selfSignedBonusApplied }` and both call sites (`hasRecipeMaterials`, `resolveCraftForRecipe`) updated accordingly; added a test pinning this exact scenario.
- Dropped the unused `depletionChance` field from `ToolEffectDef`/`TOOL_EFFECTS`: confirmed via grep it had no reader anywhere (depletion goes through the rarity-gap curve in #1139 as you noted).
- Softened the `rechargeDiscountMultiplier` doc to note the material half is an integer ceil that can floor at the same value as the plain original-crafter discount under today's placeholder constants, rather than claiming "always strictly less."

Left the off-ring `professionId` throw-vs-silent-skip behavior as is per your note that it's safe today (every shipping recipe is a ring craft).

`npx vitest run tests/professions_perks.test.ts tests/professions_tools.test.ts tests/architecture.test.ts` all green (74 tests), `tsc --noEmit` clean.

FernandoX7 (approved):

PR #1212, opened 2026-07-01
feat(professions): grant title from currently-active archetype (#1130)

### LINKED ISSUE #1130: Professions: archetype title grant from the active archetype

jgyy:
## Summary

Closes #1130.

The issue's own title/body describe a stale design ("two highest crafts reach
rare and are adjacent -> named archetype, else Jack of All Trades"), superseded
by the maintainer comment on the live issue. Per that comment, this re-scopes
#1130 to grant a title from the player's **currently-active archetype** (the
state #1129 introduced: a single active craft a character declares as their
identity via a zone-1 acceptance quest, later switchable via a repeatable
"make amends" quest), not a two-crafts derivation.

- A player with an active archetype set is granted that archetype's named
  title (`getArchetypeTitle`, `src/sim/professions/archetype.ts`): the title
  identifier is simply the active craft id, a strict one-to-one mapping onto
  the ten crafts on the ring.
- A player with no archetype chosen yet (pre-first-quest, `activeArchetype ===
  null`) has **no title** - not a "Jack of All Trades" fallback (there is none
  under this model, since a character has at most one active archetype at a
  time), just untitled. This is the natural analog of the old issue's "below
  rare grants no title" rule.
- `IWorldProfessions.archetypeTitle` (new IWorld read, implemented in both
  `Sim` and `ClientWorld`) exposes the identifier; the ten thematic title
  names (Armorer, Weaponsmith, Jeweler, Alchemist, Tinkerer, Chef, Scribe,
  Enchanter, Tailor, Leathercrafter) live as `t()` keys in
  `src/ui/i18n.catalog/hud_chrome.ts`, displayed in the character window.

```mermaid
flowchart LR
    A[No archetype chosen<br/>activeArchetype = null] -->|zone-1 acceptance quest| B[Active archetype = craftId<br/>#1129 state]
    B -->|switchArchetype via<br/>make-amends quest| B
    A -->|getArchetypeTitle| N[No title granted]
    B -->|getArchetypeTitle craftId| T[Named title displayed<br/>e.g. Armorer, Alchemist, ...]
```

Stacked on #1205 (active archetype state + quest-gated switching, which itself
includes recipe crafting/gathering/contracts); diff will shrink once that
merges; do not merge this before it. This issue's own title/body describe a
stale two-crafts-at-rare derivation superseded by the #107 decision;
implemented per the comment on the live issue (title granted from the
currently-active archetype, not a passive derivation) - see this reconciliation
note for the full context.

## Test plan

- [x] `npx vitest run tests/professions_archetype_title.test.ts
  tests/professions_archetype.test.ts tests/architecture.test.ts
  tests/world_api_parity.test.ts tests/localization_fixes.test.ts
  tests/i18n_completeness.test.ts` all green
- [x] `npm run ci:changed` (biome, changed files only): 0 errors, pre-existing
  warnings only
- [ ] Screenshot: attempted via the repo's puppeteer-core E2E pattern but
  skipped for this PR (see PR description follow-up comment); not
  fabricated.


FernandoX7 (commented):
Solid, well-scoped change. The re-scope to "title from the currently-active archetype" reads cleanly against the #1129 state: the seam stays string-free (an identifier, not prose), `getArchetypeTitle` is pure and defensive (both `null` and an unknown id fall to `null`), and IWorld parity is right, the new `archetypeTitle` read is a getter on both `Sim` and `ClientWorld` and is registered in the parity descriptor as `data`. `ARCHETYPE_TITLE_KEYS` maps 1:1 onto `CRAFT_RING` and every title is a `t()` key in the correct English catalog (`hud_chrome`). The new test is decisive: null, all ten crafts, an unknown id, accept/switch/blocked-switch, and per-pid parity against the getter.

A few things before merge.

**should-fix** `src/sim/sim.ts:1540-1541` (plus the orphan comment at `:883`): merge fallout in the gathering-proficiency load path, unrelated to this feature. The branch has diverged from base here. Base loads the field once as `normalizeGatheringProficiency(s.gatheringProficiency ?? s.professions)`, but the branch stacks two dead pre-writes in front of it:

```
meta.gatheringProficiency = normalizeGatheringProficiency(s.professions);
meta.gatheringProficiency = normalizeGatheringProficiency(s.gatheringProficiency);
// ...then the correct `?? s.professions` assignment from base
```

The last write wins so it is not a runtime bug, but the two pre-writes are dead and git auto-merges this region without a conflict, so it lands silently on merge (I confirmed it survives into the merged tree). Same story for the stray `// load cleanly, defaulting every profession to 0).` fragment above the `gatheringProficiency?:` field decl. Please reconcile the region back to base: drop both pre-writes and the orphan comment.

**nit** `src/ui/char_window.ts:65`: `archetypeTitleText` is exported "for the view-model test" but nothing imports it. The id-to-key table and the `none` fallback are unverified, so a wrong or missing key would not be caught. A small table-driven test would close it.

**note (inherited, not a blocker):** on the online client `activeArchetype` is never set from a snapshot (the whole #1129/#1205 archetype system is sim-side plus persistence only for now), so `archetypeTitle` stays `null` and the character window reads "Title: None" for every online player until a follow-up wires the self-snap field. That is consistent with the documented stub state, not this PR's doing, but you may want to decide whether to render the row at all before it is wired.

On the conflicts: three of the four are the standard stale-base i18n trio (`i18n.resolved.generated/pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`), just regen churn. Merge base and re-run the i18n build (`i18n:build` / `i18n:scan` / `i18n:hash --write`). The fourth, `tests/world_api_parity.test.ts`, is base moving the lines around the archetype block; keep `archetypeTitle` (`kind: 'data'`) in both the descriptor table and the expected-names list when you resolve it.

jgyy:
Thanks for the thorough review, both points addressed in `56d0165`:

- **should-fix (merge fallout):** reconciled the `gatheringProficiency` load path in `src/sim/sim.ts` back to base. Dropped the two dead pre-writes (`meta.gatheringProficiency = normalizeGatheringProficiency(s.professions)` / `(s.gatheringProficiency)`) that landed silently since git auto-merged them without a conflict, so only the correct `?? s.professions` fallback assignment remains. Also removed the stray orphan comment line above the `gatheringProficiency?:` field declaration in `CharacterState`, matching base exactly (verified with a direct diff against `upstream/release/v0.23.0`).
- **nit (untested view model):** added a table-driven test for `archetypeTitleText` in `tests/char_window.test.ts`: null and an unrecognized craft id both fall back to `'None'`, every craft on `CRAFT_RING` resolves to its own distinct named title (via `it.each`), and a sync check keeps the test's expected-title table aligned with the ring if a craft is ever added/renamed.

Also re-synced with `release/v0.23.0` (which moved a few times during review) and combined the `IWORLD_MEMBERS` count pins with the other concurrent additions (185/50/135). `tsc`, the targeted suites, and the full pre-push gate are green, and the PR is mergeable/conflict-free again.

FernandoX7 (approved):

PR #1214, opened 2026-07-01
feat(professions): add persistent town focus allocation with additive baseline (#1143)

### LINKED ISSUE #1143: Professions: persistent town focus allocation (additive baseline plus bonus)
## Problem

There is no persistent, town-set focus across component types that adds bonuses without ever lowering the universal baseline for any component.

## Expected behavior

- A player sets a persistent focus across component types while in town.
- Every component type has a guaranteed baseline yield/rarity that is never reduced.
- Focus points add a bonus on top of the baseline only for the chosen components; unfocused components remain at baseline.

## Scope

- `src/sim/professions/focus.ts`: additive focus allocation state and its effect on the M3 tier-tradeoff roll.
- `IWorld`: read/write surface for a player's town focus.
- Client: a town focus allocation panel.

## Acceptance criteria

- No component type's yield ever drops below its baseline regardless of how focus is allocated elsewhere.
- A focused component measurably outperforms an unfocused one in the M3 roll, all else equal.

## Notes

Consumes the per-corpse mechanic from M3 as an input to its bonus calculation, but the allocation state and UI are independently buildable and testable against a stub roll.

jgyy:
Closes #1143

Stacked on #1195 (per-corpse focus picker, #1142); diff will shrink once that merges, do not merge this before it.

## Summary
- Adds a persistent, per-player focus allocation a player sets while standing in their zone's town hub: a 10-point budget spent across gathering component types.
- `src/sim/professions/focus.ts` (pure leaf, no Sim/Entity import): `setTownFocus` validates and resolves the allocation (rejects when not in town, on non-integer/negative points, or over budget, leaving the previous allocation untouched on rejection); `applyFocusBonus`/`applyFocusTierBonus` apply the bonus on top of any baseline yield/tier.
- Wired end to end: `Sim.setTownFocus`, `ClientWorld` (`src/net/online.ts`), `server/game.ts`, `IWorld`/`IWorldInteraction` (`src/world_api.ts`, `src/world_api/interaction.ts`).
- New `town_focus_view.ts` (pure view core) + `town_focus_window.ts` (thin painter) HUD window, following the pure-core/thin-consumer recipe, gated on the player standing inside their zone's hub circle.

## Focus budget / bonus formula
- `FOCUS_POINT_BUDGET = 10` total points a player may allocate across every component type at once.
- `FOCUS_YIELD_BONUS_PER_POINT = 0.1`: each point on a component adds 10% of its baseline yield, additive and uncapped in effect (capped in practice at 100% since the budget itself is 10 points). Never reduces a baseline; 0 points leaves the baseline unchanged.
- Tier-roll integration with the #1142 harvest tiers: every `POINTS_PER_TIER_BONUS = 5` points on a component shifts its rolled tier up one step on the `HARVEST_TIERS` ladder, capped at `MAX_FOCUS_TIER_BONUS = 2` steps and at the top of the ladder. Mirrors the #1142 concentration-bonus shape (an additive, capped index shift on the same ladder) rather than inventing new balance numbers.
- Unfocused components are never affected: the baseline they'd get with zero allocation is unchanged regardless of how other components are focused.

## "In town" representation
There is no built-out town/crafting-station system yet (an unfiled epic). Every `ZoneDef` already carries a `hub` settlement circle (x/z/radius/name, e.g. Eastbrook/Fenbridge/Highwatch) that the world generator flattens into a plateau and the road/prop generation treats as town. `isInTownZone` in `focus.ts` reuses that hub circle as the town-tag stand-in (a lightweight zone-tag distance check), not a real crafting-station build-out. The "in town" gate itself is evaluated by the caller (`Sim`, which knows the player's zone/position) and passed in as a plain boolean, so `focus.ts` never reaches into world state.

```mermaid
flowchart LR
  A[Player in zone hub circle] -->|isInTownZone| B[setTownFocus: validate + budget check]
  B -->|ok| C[Persisted FocusAllocation per player]
  C --> D[applyFocusBonus: +10%/point on baseline yield]
  C --> E[applyFocusTierBonus: +1 tier / 5 points, capped at 2]
  D --> F[Harvest result]
  E --> F
```

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/focus.test.ts tests/town_focus_sim.test.ts tests/architecture.test.ts tests/command_schema.test.ts tests/snapshots.test.ts tests/world_api_parity.test.ts` (288 passed)
- [x] Full `npx vitest run` (excluding a pre-existing, unrelated `tests/admin/**` environment breakage in this worktree, a missing `@testing-library/svelte/src/vitest.js` module, reproduced identically on a clean checkout of this branch's own base)
- [x] `npm run build`
- [x] `npm run i18n:gen` regenerated (new `hudChrome.townFocus.*` keys; M16 non-Latin fills added for zh_CN/zh_TW/ja_JP/ko_KR/ru_RU in the same change)
- [x] Parity goldens regenerated (`UPDATE_PARITY=1 npx vitest run tests/parity`) for the new `townFocus` field on serialized player state
- Screenshot skipped for this change (no `npm run dev`/`npm run server` session available in this environment); the new window follows the same pure-core/thin-consumer HUD recipe as the existing corpse-harvest picker (#1195), no novel rendering risk.


FernandoX7 (commented):
Now that #1195 has landed on the base, the diff reads as a clean, well-scoped town-focus feature. The structure is right: `focus.ts` is a genuine pure leaf (no rng/DOM, no Sim/Entity import), `town_focus_view.ts` is a proper pure core (correctly added to `UI_PURE_CORES`), and the window is a thin painter over it. The wire is fully pinned end to end (`tfocus` via `maybe()` in `selfWireJson`, decoded in `applySnapshot`, `ALL_DELTA_KEYS`/`TERSE_TO_IWORLD` at 32, the IWorld parity counts at 187/51/136 with the real member names in the sorted snapshots, command-schema counts bumped), persistence is additive/back-compat JSONB, and determinism holds (the tier shift is a pure index bump computed before the rarity roll, so draw order is unchanged and the goldens moved only from the new serialized field). Good `town_focus_sim.test.ts` too: the 300-corpse measurable-edge and the unaffected-unfocused-component properties are decisive.

A few things to fix.

**should-fix, the feature is unreachable on /play.** The only opener is the `#mm-town-focus` minimap button (`src/ui/hud.ts:1457`, no keybind and no mobile button feed `toggleTownFocus`). That button is in `index.html:410` but missing from `play.html` (`play.html:356-360` has `mm-char` through `mm-map`, no `mm-town-focus`). The window div was mirrored to `play.html`, but the entry point was not, so on /play there is no way to open the panel. The `?.` on the listener saves it from throwing, which is why nothing else caught it.

**should-fix, the headline "+10% per point" yield bonus is never applied.** `applyFocusBonus` (`src/sim/professions/focus.ts:51`) is called only from itself and `tests/focus.test.ts`; the harvest path wires `applyFocusTierBonus` alone (`src/sim/interaction.ts:267`). So the actual wired behavior is: 1 to 4 points do nothing, 5 to 9 give +1 tier, 10 gives +2 tiers. Allocating 4 points has zero effect, which contradicts both the PR's per-point formula/diagram and the in-game hint at `src/ui/i18n.catalog/hud_chrome.ts:901` ("Focus points add a bonus on top of every component's baseline yield"). Either wire `applyFocusBonus` into the yield, or drop it and soften the hint/PR copy to the tier-only reality. As is, the thorough unit tests on `applyFocusBonus` are testing dead code.

**should-fix, the three `Sim.setTownFocus` error strings are not registered in the sim matcher.** The literals at `src/sim/sim.ts:5288` ("You must be in town to set your focus.", "That allocation exceeds your focus point budget.", "Invalid focus allocation.") have no rule in `src/ui/sim_i18n.ts`, so they render as raw English in every locale. This is English-side wiring, not a translation ask: the string needs a matcher row so it *can* localize, per the CLAUDE.md rule that sim player text lands its matcher entry in the same change. The S3 scanner does not catch it because the `ert` regex needs a same-line first arg and only handles a two-arm ternary, and this is a multi-line three-arm nested one, the same blind spot the pet errors had to be pinned around in `tests/localization_fixes.test.ts`. `not_in_town` is reachable via a walk-out race; the other two are defensive server-authority paths, but all three are player-facing.

**should-fix, duplicated test blocks in `tests/gathering.test.ts`.** Head lines 153-231 are a byte-identical copy of the `resolveCorpseFocusHarvest` and `harvestTierQuantity` describe blocks already at 73-151 (and already in base from #1142). These are #1142 tests, not town-focus tests; the merge re-added the branch's own copies. Drop the duplicate 80 lines.

Nits:
- `mm-town-focus` uses `data-icon="map"`, the same glyph as the World Map button (`mm-map`) right beside it (`index.html:410`). A distinct icon would read less ambiguously.
- `Sim.setTownFocus` (and the `server/game.ts` copy) persist any client-supplied component key with points > 0. It is bounded to 10 keys since the total is capped at 10, and unknown keys never match at harvest time, so it is harmless and consistent with how `harvestCorpse` ignores unknown tags, but rejecting keys outside the known harvestable set in the pure leaf would be tidier.

jgyy:
Thanks for the review, addressed all four should-fix items:

- **Unreachable on /play**: mirrored the `mm-town-focus` minimap button into `play.html` (it was only added to `index.html`).
- **Yield bonus never applied**: `applyFocusBonus` is now wired into `harvestCorpse`'s actual granted quantity in `src/sim/interaction.ts`, on top of `applyFocusTierBonus`, so points below the 5-point tier-shift threshold now do something too. The bags-full capacity reservation is updated to reserve the same focus-boosted maximum per component (previously it reserved only the unboosted top-tier quantity, which would have under-reserved once the yield bonus started actually granting more). Added a decisive regression test in `tests/town_focus_sim.test.ts` covering exactly the below-threshold case that was previously dead code.
- **Missing sim matcher**: registered the three `Sim.setTownFocus` error strings in `src/ui/sim_i18n.ts` (English `baseEnTable` entries, picked up automatically by the `EXACT` matcher) so they no longer render raw English.
- **Duplicated test blocks**: removed the byte-identical duplicate `describe` blocks in `tests/gathering.test.ts` that the release/v0.23.0 merge had reintroduced.

Left the two nits (distinct minimap icon, rejecting unknown component keys in the pure leaf) as-is per your framing of them as nits, happy to pick them up too if you'd rather have them in this PR.

Also rebased onto the latest `release/v0.23.0` in the same push to keep CI green (this branch is now `MERGEABLE`).

`tsc`, the targeted suites (focus/town_focus_sim/gathering/architecture/localization_fixes/command_schema/snapshots/world_api_parity), `tests/parity` (unchanged, no rng draw-order shift), biome (changed files), and `npm run build` all pass.

FernandoX7 (approved):

PR #1216, opened 2026-07-01
feat(loot): party/raid Loot Settings window and whole-group loot messaging

Nervescraper:
## What

Two improvements to party/raid loot settings.

**1. Loot Settings window.** A closable, reopenable window showing the group's loot rules. The leader edits them (loot method: group loot vs a master looter, plus the roll threshold); members get a read-only view. Opened from the group-panel right-click menu ("Loot Settings" on a party member or your own frame). Replaces the old always-on inline footer control.

- Auto-opens when you become party leader (creating a group, being promoted, or succeeding a leader who left), so a new group's loot rules are front and center. Non-intrusive: it does not steal keyboard focus or close your other windows.
- Docks below the party frames on the left, flowing to their right if the left column would overflow the HUD height. Draggable.

**2. Whole-group loot messaging** with clickable item links. The group now sees:

- Setting changes, stating exactly what changed (method / master looter / threshold).
- A loot-settings summary sent to a joining member.
- The effective master looter changing on leave / promote.
- A re-announce on party to/from raid conversion.
- The loot lifecycle broadcast to the whole party/raid (rolling / won / assigned / passed), scoped to premium roll-worthy items, each with a clickable item link.

## Why

Members previously had no visibility into the group's loot rules (only the leader saw the inline control), and loot outcomes were shown only to in-range looters. This surfaces the rules to everyone and keeps the whole group informed of loot activity.

## Notes

- Based on `release/v0.18.0`. The clickable item-link infrastructure already shipped upstream (the shift-click-item-link PR) and is reused; this adds the loot-log token render path on top of it.
- i18n: English catalog keys plus the five non-Latin fills for the new wordy strings; the sim stays language-agnostic (English emits re-localized by client matchers). A couple of translation choices are worth a maintainer glance (e.g. ru_RU uses the same existing term for both "Master Loot" and "Master Looter").
- Server-authoritative: the window only sends `setPartyLootMaster`, which the sim leader-gates; the read-only member view cannot mutate settings.

## Testing

- `tsc`, the guard suites (architecture/determinism, the S3 i18n matcher, i18n completeness, HUD perf budget), the new `loot_messaging_sim` and `loot_settings_view` tests, `loot_master_sim`, `social`, and the parity suite (goldens regenerated for the new message stream, events digest only) are all green.
- `npm run build` green.
- Screenshot harness: `scripts/loot_settings_shot.mjs` (offline) regenerates every image below.

## Screenshots

**Auto-opens docked below the party frames when you become party leader**

Loot Settings docked below the party frames

**Leader (editable) and member (read-only) views**

Leader editable window
Member read-only window

**Opened from the group-panel right-click menu**

Right-click Loot Settings menu entry

**Whole-group loot messages with clickable item links**

Group loot messages with item links

**Flows to the right when the left column would overflow the HUD height**

Overflow fallback to the right


EnriqueGF (commented):
Nice piece of work. The pure-core/painter split (`loot_settings_view.ts` + `loot_settings_window.ts`) is clean, and the snapshot fix in the third commit is the right call: anchoring `PendingLootRoll.partyMembers` at creation time instead of re-resolving a candidate's live party at resolve time is exactly what avoids the wrong-party broadcast, and it comes with a regression test that actually re-groups a candidate mid-roll to prove it.

A few things I specifically checked and liked:

- `announceLooterShift` is called after `party.leader` is already reassigned in `removeFromParty`/`promotePartyLeader`, so `effectiveMasterLooter` resolves against the new leader, and it correctly stays silent when a party dissolves to solo (both paths have tests).
- The `play.html` div was missing in the first commit; the second commit fixes it and adds `entry_window_parity.test.ts` so any future window added to one entry but not the other fails CI instead of NPE'ing on `/play`. Good guard to have generally, not just for this PR.
- All the new sim-emitted English strings (loot_roll.ts, party.ts) fall inside the S3 drift guard's scan paths, and `hud.ts` gets matching regex/exact arms for every one of them. i18n catalog changes are English + the five non-Latin fills for the new wordy strings, as expected for a contributor PR.
- Server authority holds: the window only calls `sim.setPartyLootMaster`, leader-gated in the sim; member view is read-only.

One thing to sort out before merge: this branched off an older `release/v0.18.0` commit. `git merge-tree` shows real conflicts in the generated i18n trio (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`, mechanical, just needs a rebase + regen) and in 8 parity goldens (`party_loot.json`, `chat_social.json`, `dungeon_instances.json`, `dungeon_raid_lockout.json`, `l1_loot_distribution.json`, `nythraxis_full_pull.json`, `party_raid.json`, `quest_link_abandon.json`, `targeting_markers.json`) where both sides changed the `events` digest independently. Needs a rebase onto current `release/v0.18.0` and a golden regen before this is mergeable.

Nothing else blocking. Didn't find any determinism, wire, or auth issues in loot_roll.ts/party.ts.

PR #1222, opened 2026-07-01
tooling(ci): trigger the AI reviewer on /review and /suggest PR comments

EnriqueGF:
## What

Follow-up to #1036 (merged into `release/v0.18.0`). Adds an on-demand trigger for the
AI PR reviewer: an OWNER, MEMBER, or COLLABORATOR of this repo can comment `/review` or
`/suggest <focus>` on a PR to re-run `scripts/ai_review.mjs` whenever they want, instead
of waiting for the next push.

## How

- New `ai-review-comment` job in `.github/workflows/pr-ai.yml`, triggered by
  `issue_comment` (created). Gated on `github.event.issue.pull_request != null` (a PR
  comment, not a plain issue), the comment starting with `/review` or `/suggest`, and the
  commenter's `author_association` being OWNER, MEMBER, or COLLABORATOR of this repo,
  checked regardless of whose PR it is (so a fork PR's own author cannot self-trigger it
  by commenting on their own PR).
- `scripts/ai_review.mjs` now supports two ways to get the diff: the existing
  `DIFF_FILE` (automatic run) or, when that is absent, a new `fetchDiffFromApi()` that
  reads it straight from `GET /repos/{repo}/pulls/{number}` with the diff media type.
  This means the comment-triggered job never checks out or executes the PR's own code
  (no `ref:` override on `actions/checkout`), so running it with full repo secrets stays
  safe even against a fork PR: only the diff text and the comment body ever reach the
  script, never executable code from the PR.
- A comment-triggered run posts its answer as a fresh reply comment keyed to the
  triggering comment's id (`<!-- pr-ai-review-comment-<id> -->`), so it never overwrites
  the standing automatic sticky review, but a retried Actions run for the same comment
  updates its own reply instead of duplicating.
- Fixed a latent concurrency bug while touching this: the `concurrency.group` only keyed
  off `github.event.pull_request.number`, which is absent on `issue_comment`; two
  `/review` comments on different PRs around the same time would have collapsed into one
  `cancel-in-progress` group. Now also keys off `github.event.issue.number`.
- Also swaps the default `OPENROUTER_MODEL`: `openrouter/owl-alpha` has been pulled from
  OpenRouter, replaced by `nvidia/nemotron-3-ultra-550b-a55b:free`.

## Docs

`docs/ai-pr-bot.md` updated: a new "Requesting a review on demand" section, the trust
gate rationale, and the fork-PR security note extended to cover why `issue_comment`
(unlike `pull_request`) always runs with secrets and why that's still safe here.

## Verification

- `node --check` on the changed script; Biome on the changed files.
- `npx vitest run tests/architecture.test.ts` (guard test, unaffected but green).
- Manually traced both trigger paths against the actual API/permission model (fork PR
  cannot self-trigger via `author_association`; no PR-head checkout in the new job).
- No em dashes, en dashes, or emojis introduced (repo convention).

Generated with Claude Code

### DIFF
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -48,6 +48,10 @@ jobs:
   screenshots:
     name: Screenshots of changes
     runs-on: ubuntu-latest
+    # issue_comment fires this workflow on every comment on any issue or PR; without this
+    # guard the full checkout + npm ci + Vite boot + puppeteer run would fire on every
+    # comment too, even though nothing in this job reads the comment.
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -1,32 +1,47 @@
 name: PR AI assist
 
-# Two INFORMATIONAL, NON-BLOCKING jobs that help review a pull request. Neither is a
-# required check; they assist review, they do not gate merge. They live in their own
-# workflow so they never interfere with the CI gate in ci.yml.
+# Two INFORMATIONAL, NON-BLOCKING jobs that help review a pull request, plus an on-demand
+# comment trigger for the second one. None of this is a required check; it assists
+# review, it does not gate merge. Lives in its own workflow so it never interferes with
+# the CI gate in ci.yml.
 #
-#   screenshots  boots the Vite dev client headless, runs a fixed offline tour
-#                (character select, desktop HUD, mobile HUD), uploads the PNGs as an
-#                artifact, and links them in a sticky PR comment.
-#   ai-review    sends the PR diff to an OpenRouter model (free openrouter/owl-alpha by
-#                default) and posts the review as a sticky PR comment. No-ops without the
-#                OPENROUTER_API_KEY secret (for example on a fork PR), so it never blocks.
+#   screenshots        boots the Vite dev client headless, runs a fixed offline tour
+#                       (character select, desktop HUD, mobile HUD), uploads the PNGs as
+#                       an artifact, and links them in a sticky PR comment.
+#   ai-review           sends the PR diff to an OpenRouter model (free
+#                       nvidia/nemotron-3-ultra-550b-a55b:free by default) and posts the
+#                       review as a sticky PR comment. No-ops without the
+#                       OPENROUTER_API_KEY secret (for example on a fork PR), so it never
+#                       blocks.
+#   ai-review-comment   on demand: an OWNER/MEMBER/COLLABORATOR comments `/review` or
+#                       `/suggest <focus>` on the PR. Fetches the diff itself via the
+#                       GitHub API (no checkout of the PR head, so this stays safe to run
+#                       with repo secrets even against a fork PR) and posts a fresh reply
+#                       comment, keyed to the triggering comment rather than the sticky
+#                       review above.
 #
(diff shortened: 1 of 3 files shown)


FernandoX7 (commented):
The security model here is handled correctly, and it's the part that matters most for an `issue_comment` + secrets workflow. Worth calling out that these are all right:

- The `issue_comment` handler always runs the workflow from the default branch, so a fork PR can't edit the gate. The comment job does `actions/checkout@v6` with no `ref:`, so it checks out the base repo, never the PR head, and never executes untrusted code. `.github/workflows/pr-ai.yml:166`
- `author_association` gate is the correct idiom (`contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), ...)`) and is evaluated against the base repo, so a fork author can't self-trigger. `.github/workflows/pr-ai.yml:172`
- The comment body reaches the script via an env var, not interpolated into a `run:` line, so no script injection. The diff is fetched as text via the diff media type and only sent to the model. `scripts/ai_review.mjs:76`
- The concurrency-group fix (adding `github.event.issue.number`) is a real latent bug caught in passing, and the defensive `parseCommand` no-op is a nice touch.

Findings:

**should-fix. Adding `issue_comment` to `on:` makes the `screenshots` job run on every comment.** `.github/workflows/pr-ai.yml:30` adds the `issue_comment` trigger, but the `screenshots` job at `:48` has no job-level `if`. Only its diff/comment steps are gated to `pull_request` (`:95`, `:122`); the checkout, `npm ci`, i18n gen, Vite boot, and the "Capture screenshots" step (`:101`) plus the `if: always()` artifact upload (`:114`) are not. So every comment on any issue or PR in the repo will spin up a full Vite + puppeteer run and upload a stray `pr-screenshots` artifact, even though nothing gets posted. The `ai-review` job avoids this because it already carries `if: github.event_name == 'pull_request'` (`:135`); the `screenshots` job needs the same guard (for example `if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'`).

**nit. Comment-path `repoFiles` reflects the base tree, not the PR head.** `scripts/ai_review.mjs:110` runs `git ls-files` over the checked-out tree, which on the comment run is the base default branch (no PR-head checkout, correctly). But the diff itself comes from the API and includes files the PR adds, so a newly added helper won't be in the "these files exist, do not flag as missing" list. That is the exact false positive the list was added to prevent, now reintroduced for the comment path only. Non-blocking and self-inflicted only on `/review`; worth a comment noting the limitation if not fixed.

**nit. `/review` shares a concurrency group with the automatic run.** `.github/workflows/pr-ai.yml:45` keys both the push-triggered jobs and the comment job on the same per-PR group with `cancel-in-progress: true`, so a `/review` comment can cancel an in-flight automatic review for the same PR and vice versa. Probably acceptable (you usually want the latest), just flagging the interaction.

The model swap (`openrouter/owl-alpha` to `nvidia/nemotron-3-ultra-550b-a55b:free`) is non-blocking by design: a bad model id just leaves the "could not run" note, so no concern there.

EnriqueGF:
Addressed in bbdc714e:

- **should-fix**: `screenshots` job now has a job-level `if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'` guard, so `issue_comment` events no longer spin up the full checkout + `npm ci` + Vite boot + puppeteer run.
- **nit**: added a comment in `scripts/ai_review.mjs` documenting that `repoFiles` reflects the base tree on comment-triggered runs, so a file the PR newly adds will not appear in the "don't flag as missing" list.

Left the shared concurrency group as is, since that nit was flagged as probably acceptable.

FernandoX7 (approved):

FernandoX7:
Waiting for 0.19 to merge

PR #1234, opened 2026-07-01
fix(ui): size chat input for placeholder text

### LINKED ISSUE #1232: Chat input box is an odd height and clips its placeholder text
## Problem

The chat input box on the web client renders at an odd, inconsistent height that does not match its content.

When the input is opened but empty, the box is much taller than a single line of text needs: it shows a large empty area with the cursor sitting near the top (screenshot 1). When the placeholder hint is shown (or when you start typing), the box does the opposite and is too short: the placeholder wraps to two lines and the second line is clipped at the bottom edge, cutting off the end of "Say something... (/s say, /w name whisper, /r reply, /p party, /gu guild, /o officer, /general general, /help)" (screenshot 2).

So the same box is too tall when empty and too short once it has content to show, which looks broken and hides part of the placeholder hint.

## Steps to reproduce

1. Log in and enter the world on the web client (desktop).
2. Press Enter (or click the chat bar) to open the chat input.
3. Observe the empty box height, then observe the placeholder hint / start typing.

## Actual behavior

- Empty, focused input renders as a tall box with lots of empty vertical space and the caret near the top.
- The placeholder hint wraps to two lines and the bottom line is clipped, so the tail of the hint ("/general general, /help)") is not readable.
- The box height does not track the content it is displaying.

## Expected behavior

- The empty input is a single-line height (no oversized empty box).
- Whatever text is visible (placeholder hint or typed input) fits inside the box with no clipping at the bottom edge.
- As typed text wraps past one line, the box grows upward smoothly up to its max height, then scrolls internally, with no clipped lines at any size.

## Scope

- `src/ui/chat_input_autosize.ts`: the pure clamp geometry (`chatInputSize`, min/max height).
- The DOM consumer and re-anchor logic in `src/main.ts` (`anchorChatInput`) that measures the textarea `scrollHeight` and applies the height.
- `#chat-input` in `src/styles/hud.css` (padding `7px 10px`, `line-height: 1.35`, `max-height: 110px`, no CSS `min-height`).
- The placeholder string is long enough to wrap to two lines; a textarea `scrollHeight` measurement ignores the placeholder, so an empty-but-focused box can be sized for zero content while the placeholder needs two lines. Worth confirming as the clip cause.

## Notes

Likely two separate contributors: (1) an initial/min height that is larger than one line (the tall empty box), and (2) the autosize measuring only the typed value, not the placeholder, so the box under-sizes and clips the wrapped placeholder hint. Both should be reconciled so the box is one line when empty and always tall enough for its visible content.

## Acceptance criteria

- Opening the chat input with no text shows a single-line-height box, not an oversized empty one.
- The full placeholder hint is visible with no clipping at the bottom edge.
- Typing multi-line text grows the box upward and never clips the current line; past max height it scrolls internally.
- The desktop experience is otherwise unchanged and mobile chat input is unaffected.

## Screenshot

Image

Image

Humpalumps:
## Summary

Fixes the desktop chat textarea sizing so the box follows the visible text instead of clipping the long default placeholder or under-sizing wrapped input.

- Measures the placeholder text in an off-screen textarea when the chat input is empty, so the full hint can wrap without being clipped.
- Adds the textarea border thickness to measured `scrollHeight` before applying a border-box CSS height.
- Keeps short placeholders at the single-line floor and preserves the existing upward growth / internal scrolling behavior for typed multi-line messages.
- Adds pure regression coverage for placeholder-vs-value sizing and border-box height math.

## Related issues

Closes #1232.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src/main.ts src/ui/chat_input_autosize.ts tests/chat_input_autosize.test.ts` - passed; formatted 3 files.
  - `npx biome lint src/main.ts src/ui/chat_input_autosize.ts tests/chat_input_autosize.test.ts` - passed with existing warning-level diagnostics in `src/main.ts` unrelated to this change.
  - `npm run check:ts` - passed.
  - Direct `npx vitest run tests/chat_input_autosize.test.ts` initially hit the known release-branch `.browserslistrc` comment parser issue.
  - Wrapped `.browserslistrc` cleanup around `npx vitest run tests/chat_input_autosize.test.ts` - passed, 1 file / 10 tests.
  - Wrapped `.browserslistrc` cleanup around `npx vitest run tests/chat_input_autosize.test.ts tests/client_shell.test.ts` - passed, 2 files / 76 tests.
  - Wrapped `.browserslistrc` cleanup around `npm run build` - passed with existing Vite dynamic-import/chunk-size/plugin-timing warnings.
  - `git diff --check` - passed; Git for Windows printed LF-to-CRLF working-copy warnings only.
- Manual steps:
  - Started local Vite dev server on `127.0.0.1:5174` with the temporary `.browserslistrc` cleanup used for build/tests, then restored the original file bytes after stopping the server.
  - Ran `GAME_URL=http://127.0.0.1:5174 SHOT_LABEL=placeholder-height node scripts/chat_input_reposition_shot.mjs`.
  - Verified the captured chat input geometry stayed above the chat log: `input.bottom=681`, `box.top=687`, `overlap=-6px`.
  - Sanity-checked both screenshot PNGs with `sharp` metadata/statistics so the committed images are non-empty.

## Screenshots / recordings

Empty input with the default placeholder fitting without clipped bottom text:

Empty chat input placeholder

Wrapped typed message growing upward above the chat log:

Long chat input message

---

### Quality

      TypeScript `strict`).

Focused tests and build pass; full `npm test` was not run for this CSS/DOM-only slice.

### Cross-platform

Desktop web was validated with screenshots. Mobile CSS placement is untouched; the shared autosize math remains clamped to the existing max height.

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Solid, well-scoped fix. The pure geometry (`chatInputVisibleHeight`) lands in the existing module with NaN/negative guards and unit coverage, while the DOM glue stays thin in `main.ts`. That is exactly the pure-core plus thin-consumer split this repo wants.

I verified the load-bearing pieces:

- `#chat-input` is `box-sizing: border-box` (`src/styles/hud.css:3297`, plus the global in `src/styles/base.css:10`), so `scrollHeight` excludes the border and the applied `style.height` needs `scrollHeight + border`. Adding `borderBlockSize` (`offsetHeight - clientHeight`) is correct, and it also fixes the earlier under-sizing of typed multi-line input, not just the placeholder case.
- The placeholder measurement only runs while the input is empty (`hasValue ? 0 : measureChatInputPlaceholder()`), so the typing hot path stays cheap.
- The placeholder is re-read on each measure and `openChat` refreshes it before autosizing, so channel/locale placeholder swaps re-measure correctly.
- `/play` loads `/src/main.ts` (`play.html:908`), so it inherits the fix; no separate-entry gap.
- No new player-visible strings; the placeholder stays the existing `t()` key. The comment edit also drops a pre-existing em dash.

`npx vitest run tests/chat_input_autosize.test.ts` passes locally (6/6).

Non-blocking:

- nit: the off-screen `chatInputMeasure` textarea is created once and never removed. Fine as a single cached node, just noting it is intentional.
- Full `npm test` was not run, but the focused autosize and `client_shell` suites plus the build cover this DOM/CSS slice and it touches no sim/server behavior, so that reads as a reasonable call.

Looks good to me.

FernandoX7 (approved):
THIS WILL BE GREAT

Humpalumps:

PR #1236, opened 2026-07-01
feat(ui): Action bar size slider (extracted from #926)

nicadeddu:
## Summary

Adds an **"Action bar size"** slider to the Interface options: 40% to 140% (default 100%), resizing just the action bar buttons, gaps and keycaps (both bars) via CSS `zoom` on `#actionbar` / `#actionbar2`, independent of the global **UI Scale**. Purely a display preference; it never touches gameplay or the bindings, so a slot stays castable at any size.

### Extracted from #926

This is lifted out of #926 (optional second action bar). Since that PR was opened, `release/v0.19.0` already shipped the **second action bar** itself, its **opt-in "Show Secondary Action Bar" toggle**, and the **loadout persistence** fix (`SAVED_LOADOUT_BAR_SLOTS`), so the bulk of #926 is now superseded. The size slider was the one still-unique, self-contained piece with no overlap, so it is broken out here to land cleanly on its own. #926 is being closed; the remaining idea (the on-bar key-binding mode) moves to a separate issue.

## Changes

- **settings:** `actionBarScale` scalar (`min 0.4, max 1.4, def 1`).
- **main.ts:** publishes `--action-bar-scale` from the setting alongside the existing `--ui-scale`.
- **hud.css:** both bars `zoom: var(--action-bar-scale, 1)`, so buttons/gaps/keycaps scale together and the bar box shrinks or grows with them (desktop and mobile).
- **options_view:** a slider row right after UI Scale; the interface-controls order test asserts it.
- **i18n:** `hudChrome.options.actionBarScale` (English) plus the five non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) the M16 gate requires for a wordy new string; regenerated resolved tables + hash.

## Type of change

## How was this tested?

- `npx tsc --noEmit` clean.
- `npx vitest run tests/options_view.test.ts tests/i18n_completeness.test.ts tests/i18n_status_registry.test.ts tests/i18n_resolved_equivalence.test.ts tests/settings.test.ts` green (the interface-controls order, the M16 completeness gate, and the i18n freshness/hash checks).
- `biome ci` on the changed files is clean (no errors, no format diff).
- Manual (offline client, `npm run dev`): dragged the slider across 40% to 140%; both bars scale together and stay centered on desktop and on a phone viewport in portrait and landscape; UI Scale and the bars scale independently.

## Screenshots

The slider in the Interface panel (next to UI Scale), and the bars at reduced sizes:

Action bar size slider row
Interface panel with Action bar size slider
Action bar size at 80% on mobile

### Quality
- [x] Tests pass (targeted suites above; no `sim/` or `server/` behavior changed).
- [x] Typecheck passes (`npx tsc --noEmit`).

### Cross-platform
- [x] Tested on desktop and mobile (portrait + landscape).
- [x] Accessible: the slider reuses the existing Interface slider control (labelled, keyboard-operable); no new interaction pattern.

### Localization (i18n)
- [x] New string is a `t()` key in `en` (`hudChrome.options.actionBarScale`) with the five non-Latin fills required by M16; remaining locales left `pending` for the maintainer's release fill. Resolved tables regenerated via the build; no hand-edited generated files beyond the regen.
- [x] The percentage uses the existing formatter.

### Hygiene
- [x] No secrets; `ALLOW_DEV_COMMANDS` untouched.


FernandoX7 (commented):
Nice clean extraction. It reuses the range machinery (auto clamp, default and persistence via `SETTING_RANGES`), the existing `slider` primitive with the `percent` formatter, and publishes a CSS var on `:root` exactly like the sibling scales, so there is very little new surface. The apply path is inherited too: the startup apply-all loop and `onSettingChange` both flow through `applySetting`, so it lands on load and on drag. play.html parity is automatic here since the rule lives in the shared `src/styles/hud.css` and both entries boot through `src/main.ts`, so nothing needs mirroring. English key and `t()` wiring are correct, and the interface-order test pins the new row.

On the two things worth a close look:

**Icons and in-button text scale correctly.** Because `zoom` scales the whole subtree, the icon art (`.icon-label`), the keycap (`.keybind`), the cooldown text and the item count all resize proportionally with the button, so there is no separate font handling to get wrong. `zoom` rather than `transform: scale` is also the right call on desktop: it reflows, so the bar stays centered under `#bottom-bar`'s `translateX(-50%)` at any size.

**Mobile is the one gap.**

- should-fix: the mobile bar is full width and the new zoom has no width compensation. `body.mobile-touch #actionbar` is `width: 100%` (`hud.mobile.css:518`) inside a fixed band between the two on-screen thumb clusters (`#bottom-bar` is inset `left`/`right` by 154px, `hud.mobile.css:418-420`). `zoom` scales the rendered box, so above 100% the bar renders wider than that band and spills right over the jump/attack controls, and below 100% it under-fills and left-aligns instead of staying centered. This is the exact case the `#ui` rule already handles for the global scale: it zooms and then divides the box back, "so edge-anchored frames (action bar ...) keep their corners instead of being pushed off-screen" (`hud.css:2411-2420`). Desktop is fine because the bar is content-sized and re-centers. Suggested fix, mirror the ui-scale pattern on mobile only:

  ```css
  body.mobile-touch #actionbar {
    width: calc(100% / var(--action-bar-scale, 1));
    max-width: calc(100% / var(--action-bar-scale, 1));
  }
  ```

  that keeps the scroll strip the same width while the buttons grow. Worth a quick emulator pass at 140% in portrait and landscape; the manual test likely missed it because the mobile screenshot is at 80% and only sizes above 100% overflow.

- nit: at the 0.4 floor the buttons land around 17 to 18px (46 down to ~18 on desktop, 42 down to ~17 on mobile), under the ~24px minimum tap target. It is opt-in and matches how `uiScale`'s low end behaves, so not a blocker, but it gets fiddly on touch; you could floor the mobile minimum a little higher if you want.

Nothing else stands out: no sim, server or wire touched, the settings clamp and default are inherited, and the i18n regen is the usual generated churn.

nicadeddu:
Good catch, thanks. Fixed in b4c0b69d: the mobile bar now divides its box back by `--action-bar-scale`, mirroring the `#ui` ui-scale rule you pointed at:

```css
body.mobile-touch #actionbar {
  width: calc(100% / var(--action-bar-scale, 1));
  max-width: calc(100% / var(--action-bar-scale, 1));
}
```

So the scroll strip keeps the band width and the bar stays centered between the thumb clusters; landscape inherits it (its rule overrides only gap/padding/background, not width). Re-checked at 140% and 40% in portrait and landscape: no spill over the jump/attack controls and no left-align.

On the nit: I left the 0.4 floor as-is so it matches `uiScale`'s low end rather than special-casing this one slider, but happy to raise the mobile minimum if you'd prefer a firmer tap-target floor.

nicadeddu:
Follow-up in 8bc6ee3b: local testing surfaced a related layout issue the first fix did not cover. The bars live in `#actionbar-stack`, a column whose default `align-items: stretch` sized each bar to the fixed 612px xp bar instead of to its buttons, so a bar shorter than the xp bar left empty space to the right, and once zoomed the shrunk bar anchored left (zoom's top-left origin) rather than staying centered.

- desktop: `align-self: center` on `#actionbar` / `#actionbar2`, so each bar is content-sized (hugs its buttons) and centered under the xp bar at any scale.
- mobile: swapped the width-compensation for `width: max-content` + `align-self: center` + `max-width: calc(100% / var(--action-bar-scale, 1))`. The strip now hugs the buttons and centers within the thumb-cluster band, still capped to the band width so a >100% scale scrolls rather than spilling over the jump/attack controls.

Verified locally: bars hug + stay centered at 40%, 100%, and 140% on desktop (single and second bar) and mobile (portrait + landscape).

PR #1237, opened 2026-07-01
fix(ui): center realm maintenance label

### LINKED ISSUE #1233: "Realm maintenance" label is off-center on the offline page
## Problem

On the offline / maintenance page (`public/server-unavailable.html`), the small "Realm maintenance" eyebrow label above the heading is not horizontally centered. It sits noticeably left of center, while the logo, the "The realm is temporarily unavailable." heading, the description paragraph, and the "Back soon" status pill all read as centered.

The result is that the label looks misaligned against the rest of the panel, which makes the page feel unfinished.

## Steps to reproduce

1. Open the offline page directly (`public/server-unavailable.html`), or trigger it by loading the site while the realm is unavailable.
2. Look at the "REALM MAINTENANCE" label directly under the logo.
3. Observe that it is shifted left of the panel's center line, out of alignment with the centered logo, description, and status pill.

## Actual behavior

- The "Realm maintenance" eyebrow renders left of center relative to the other centered elements in the panel.

## Expected behavior

- The "Realm maintenance" eyebrow is horizontally centered within the panel, aligned with the centered logo, heading, description, and status pill.
- Centering holds across the localized variants of the label (some locales are longer or shorter than the English string).
- Centering holds at both desktop and narrow (mobile) widths.

## Scope

- `public/server-unavailable.html` only.
- The panel (`main`) already sets `text-align: center`; confirm the `.eyebrow` rule participates in that centering as expected.
- Check the interaction with `.eyebrow`'s `letter-spacing: 0.16em` and `text-transform: uppercase`, since letter-spacing adds trailing tracking after the final character that can visually pull centered text left.

## Acceptance criteria

- The "Realm maintenance" label is visually centered within the panel on the offline page.
- The fix holds for every locale in the page's inline `copy` map and at mobile widths (the max-width 560px breakpoint).
- The heading, description, logo, and status pill layout are unchanged.

## Notes

Likely a trailing letter-spacing effect: `.eyebrow` carries `letter-spacing: 0.16em`, which applies after the last letter and offsets otherwise-centered text to the left. A common fix is compensating with a matching `text-indent` (or trimming the trailing tracking) rather than removing the centering. Please confirm the cause before changing.

## Screenshot

Image

Humpalumps:
## Summary

Centers the offline maintenance page eyebrow label in `public/server-unavailable.html`.

The issue was a combination of the shared paragraph rule applying `max-width: 470px` to the `.eyebrow` paragraph while the eyebrow's own margin kept that box left-positioned on desktop, plus the uppercase tracking leaving trailing letter-spacing after the final glyph. The eyebrow now opts out of the paragraph max-width and uses matching `text-indent` to balance the visible tracking.

## Related issues

Closes #1233

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write public/server-unavailable.html tests/server_unavailable.test.ts` - passed
  - `npx biome lint public/server-unavailable.html tests/server_unavailable.test.ts` - passed
  - `npm run check:ts` - passed
  - Wrapped `npx vitest run tests/server_unavailable.test.ts` with temporary `.browserslistrc` comment cleanup - passed, 1 file / 2 tests
  - Wrapped `npm run build` with temporary `.browserslistrc` comment cleanup - passed with existing Vite warnings about ineffective dynamic imports, large chunks, and plugin timings
- Manual steps:
  - Started local Vite on `http://127.0.0.1:5175` with the same temporary `.browserslistrc` cleanup, then restored `.browserslistrc`.
  - Opened `server-unavailable.html?lang=en` at desktop and mobile widths and captured screenshots.
  - Checked every inline offline-page locale at a 390px mobile viewport; each eyebrow remained single-line and centered within ~1px in `docs/pr-assets/realm-maintenance-label-align/measurement-report.json`.

Full `npm test` was not rerun for this narrow static-page fix; the release branch has known unrelated full-suite blockers from the ongoing 0.19 cleanup work.

## Screenshots / recordings

- Desktop: `docs/pr-assets/realm-maintenance-label-align/01-desktop-en.png`
- Mobile: `docs/pr-assets/realm-maintenance-label-align/02-mobile-en.png`
- Measurement report: `docs/pr-assets/realm-maintenance-label-align/measurement-report.json`

---

## Checklist

### Quality

      `sim/` or `server/` behavior I changed. Focused Vitest for this static page passed; full `npm test` was not rerun because of known unrelated release-branch blockers.
      TypeScript `strict`).

### Cross-platform

      least 40x40px and inputs at least 16px font (see `src/ui/CLAUDE.md`). Static offline page verified at desktop and phone-sized portrait widths; no touch targets changed.
      respect for `prefers-reduced-motion` (only if this change adds or edits UI). N/A: no interactive behavior changed.

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Clean, well scoped fix. The root cause diagnosis holds up: `.eyebrow` is a `<p>`, so on the base it inherited `max-width: 470px` from the shared `p` rule while its own `margin: 0 0 10px` kept the horizontal margins at 0, so that 470px box sat left anchored inside the 640px `main` and the centered text landed left of center. `max-width: none` is the real fix and it matches what I see comparing against the base. Nice to ship the per locale measurements and screenshots under the existing `docs/pr-assets/` convention, and the test pins the actual invariant (`text-indent` must equal `letter-spacing`) rather than a brittle snapshot.

Verified locally: `npx vitest run tests/server_unavailable.test.ts` is green (2/2), and the branch merges into `release/v0.19.0` with no conflicts.

One thing worth a second look:

- nit `public/server-unavailable.html:79`: the `text-indent: 0.16em` compensation looks like it overcorrects. Browsers add a trailing letter-spacing advance after the last glyph, so centered text is offset left by half the tracking; a `text-indent` equal to the full tracking flips that to half a tracking on the right instead of removing it. Your own `measurement-report.json` shows the residual: `offsetPx: 0.95` with `letterSpacing` and `textIndent` both `1.92px`, and 0.95 is almost exactly half of 1.92. To actually zero it, `text-indent` wants to be half the letter-spacing (about `0.08em`); as written it is cosmetically the same sub-pixel result as omitting the line. Sub-pixel either way so not blocking, just flagging since centering is the whole point of the change.

Not an i18n concern: this is the standalone static fallback page with its own inline `copy` map, not the `t()` catalog, and the diff adds no strings.

Humpalumps:

PR #1241, opened 2026-07-01
fix(mobile): anchor joystick input to touch start

### LINKED ISSUE #1229: Mobile web touch controls drift and register unintended movement
## Problem

The on-screen touch controls on mobile web do not track input accurately. Movement drifts noticeably: the character keeps moving after the touch is released or held still, and the direction registered does not match where the player's thumb actually is. The result is that precise movement is very hard and the character wanders on its own.

This makes the game close to unplayable on mobile web, where the touch controls are the only way to move.

## Steps to reproduce

1. Open the game on a mobile device in the browser (landscape).
2. Use the on-screen movement control (virtual joystick / touch pad) to move the character.
3. Move the thumb around, then hold it still, then lift it off the screen.
4. Observe how the character moves relative to the touch input.

## Actual behavior

- The character keeps drifting or moving after the touch is held still or released.
- The registered direction and speed do not match the current thumb position.
- Small or slow thumb movements produce disproportionate or lagging character movement.

## Expected behavior

- The character moves only while a touch is active on the control, and stops promptly when the touch is released.
- Movement direction and magnitude track the thumb position accurately, with no residual drift.
- Holding the thumb still results in steady, stable movement (or none), not wandering.

## Scope

- Mobile touch input handling and the virtual movement control.
- Touch start / move / end handling and control recentering / reset on release.
- How touch position is translated into movement intent sent to the server.

## Environment

- Platform: mobile web, landscape.
- Surface: in-game on-screen touch controls.

## Acceptance criteria

- Releasing the touch stops the character promptly, with no residual drift.
- Registered movement direction and magnitude match the thumb position throughout a touch.
- Holding the thumb still yields stable input (no wandering).
- The desktop experience (keyboard / mouse input) is unchanged.

Humpalumps:
## Summary

Fixes mobile joystick drift when the thumb starts near the edge of the movement zone.

- keeps the joystick visual base clamped on-screen
- measures movement input from the actual touch-down point instead of the clamped visual center
- preserves release behavior so pointerup/cancel/blur still clears movement promptly
- adds a regression test for an edge touch: pointerdown and a held still touch stay neutral, a real thumb offset maps to movement, and release clears input

## Related issues

Closes #1229

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src/game/mobile_controls.ts tests/mobile_controls.test.ts`
  - `npx biome lint src/game/mobile_controls.ts tests/mobile_controls.test.ts` - passed with existing warning-level diagnostics in `tests/mobile_controls.test.ts` helper code
  - wrapped `.browserslistrc` then `npx vitest run tests/mobile_controls.test.ts tests/input.test.ts tests/move_input.test.ts` - 99 passed
  - `npm run check:ts` - passed
  - wrapped `.browserslistrc` then `npm run build` - passed with existing Vite/admin dynamic-import and chunk-size warnings
  - `git diff --check` - passed
- Manual steps:
  - Captured a mobile landscape touch-control screenshot asset with Vite/Puppeteer.
  - The regression test simulates the relevant edge-touch sequence directly: edge pointerdown neutral, held-still pointermove neutral, real thumb offset maps to strafe-right, pointerup clears movement.

## Screenshots / recordings

Mobile touch-control surface:

Mobile touch controls

---

## Checklist

### Quality

      `sim/` or `server/` behavior I changed. Focused Vitest coverage passed; full `npm test` was not run for this narrow mobile input PR.
      TypeScript `strict`).
      `npm run build:env` if I touched those. Server/env were not touched.

### Cross-platform

      least 40x40px and inputs at least 16px font (see `src/ui/CLAUDE.md`). Mobile landscape control surface was captured; no desktop behavior path was changed.
      respect for `prefers-reduced-motion` (only if this change adds or edits UI). N/A: this changes pointer math only; no new UI controls or text.

### Localization (i18n)

      (`formatNumber`, `formatMoney`, `formatDateTime`, `Intl`). No new formatting.
      `npx vitest run tests/localization_fixes.test.ts` passes. N/A: no sim/server text.

### Hygiene

      `src/render/assets/manifest.generated.ts`. They're regenerated by the build. Build-generated guide/media/sitemap/admin/i18n outputs were restored after validation.


FernandoX7 (commented):
Solid, well-targeted fix, and the root cause is exactly right. The joystick base is clamped to stay on-screen, but movement was being measured from that clamped center, so a touch landing near the edge of the move zone starts with a non-zero displacement and the character drifts the instant you press, before the thumb moves. Anchoring the input origin to the actual touch-down point while keeping the clamped `origin` only for the visual base is the correct separation.

What I checked:
- The rename is complete, no `moveOriginX/Y` reads remain, and `origin` is now used only for `style.left/top`.
- The regression test is genuinely failing-first: with the zone rect (240x240), radius 61, and a touch at (5, 235), `clampJoystickOrigin` clamps to (61, 179) on both axes, so the old code registered `{back, strafeLeft}` at touchdown. Reverting the fix fails exactly that new test; with the fix, touchdown is neutral and a real offset maps to strafe-right.
- The release path the issue also calls out (stop promptly on release) is unchanged and still robust: pointerup/pointercancel/lostpointercapture plus window pointerup/pointercancel, blur, and visibilitychange all funnel to `releaseMove` to `clearTouchMove`.

One non-blocking note:
- nit (`src/game/mobile_controls.ts:574`): near a screen edge the visual base now sits at the clamped `origin` while input reads from the un-clamped touch, so at touchdown the knob renders at the base center rather than under the thumb. That's the standard floating-joystick tradeoff and I think the right call, flagging only so it's a conscious decision.

Scope is clean: `src/game/` pointer math to movement intent only, no sim/wire/server/render/string changes, so authority and i18n are N/A.

FernandoX7 (commented):
One nit:

- `src/game/mobile_controls.ts:570`: the comment still reads "pin the stick offset to that floating centre", but the offset is now measured from the touch-down point, not the clamped centre. Worth a one-line update so the next reader isn't misled.

Humpalumps:

PR #1244, opened 2026-07-01
feat(ui): mob mouseover tooltip + movable, lockable target frame

Blaine1705:
## Summary

This branch adds two related classic-HUD features for inspecting mobs.

### 1. WoW-style mob mouseover tooltip
- Hovering a live mob in the 3D world shows a small tooltip, fixed to the right of the player frame's health/resource bars, with the mob's name, level, and creature type, plus a Friendly/Hostile reaction line (green/red, from `Entity.hostile`).
- The name/level line is colored by the tooltip's own classic con spread (`mobTooltipConColor`): red only when the mob out-levels the viewer by 5+, orange at 3-4 above, yellow across the even band (+2 down to -2), green 3 to 5 below, grey 6+ below (trivial). This palette is **tooltip-only**: the overhead nameplate bands (`mobNameColor`) are left exactly as shipped, so this feature does not recolor any in-world nameplate.
- New `hudChrome.mobTooltip.*` catalog keys; translated in the five non-Latin locales (M16 requirement).

### 2. Movable, lockable target frame
- Left-clicking a mob shows the existing target frame (portrait, name, level, health) at the classic top-left slot. This change lets the player reposition it.
- A small round toggle button in the frame's top-right corner unlocks the frame for dragging (gold highlight, dashed outline, move cursor, padlock glyph) and locks it back in place (move-arrows glyph). It carries a proper `aria-pressed` toggle, an accessible name that swaps with state, and a token-based `:focus-visible` ring.
- The chosen position persists to `localStorage` across reloads; the lock state does not (the frame always loads locked, so a stray drag never moves it). Off-screen positions are re-clamped into view, including on window resize; on the mobile layout the desktop position is not applied (and any leftover inline position is cleared).
- Desktop only: the button hides under `body.mobile-touch` and the drag handlers early-return on the mobile layout.
- Architecture: the pure position math (clamp, (de)serialize) lives in `src/ui/target_frame_pos.ts` (mirroring `chat_window.ts`); the DOM wiring is a new section in `hud.ts` next to the movable chat box. No sim/server/wire change; nothing runs on the per-frame HUD path. The button label re-localizes on a language switch.
- New `hudChrome.targetFrame.{unlock,lock}` catalog keys with the five non-Latin M16 fills.

### Also included
- Fixes a pre-existing, unrelated `tests/architecture.test.ts` failure found while preparing the tooltip: `src/world_api/chat.ts` value-imported `OVERHEAD_EMOTE_IDS` from `src/sim/types`, violating the IWorld seam's type-only-sim-import rule. It now derives its runtime id set from its own local `OVERHEAD_EMOTES` (no behavior change). A Windows-only bug in the guard test (allowlist keyed by a forward-slash path but looked up with OS-native separators) is fixed alongside it.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/mob_tooltip_view.test.ts tests/target_frame_pos.test.ts tests/pet_reaction.test.ts tests/architecture.test.ts tests/i18n_completeness.test.ts tests/localization_fixes.test.ts tests/localization_coverage.test.ts tests/hud_perf_budget.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts` (all green)
- [x] `npm run i18n:gen` + `npm run i18n:hash -- --write` (new translated content, hash re-baselined)
- [x] `npx @biomejs/biome check` on all changed files (no errors; pre-existing warnings only)
- [x] `qa-checklist` gate run over the target-frame diff (i18n M16, a11y, per-frame perf, mobile, painter-conflict all covered)
- [x] Manually verified in a real browser (offline): the mob tooltip shows on hover with the right name/level/type/reaction color and clears on leave / for dead mobs. For the target frame: clicking a mob shows the frame, the corner button unlocks it (aria-pressed true, dashed outline), the frame drags to a new spot, the button locks it (aria-pressed false), the position persists to `woc_target_frame_pos`, and after a reload the frame reappears at the saved spot and locked.


FernandoX7 (commented):
This is a clean, well-factored PR. The pure cores (`mob_tooltip_view.ts`, `target_frame_pos.ts`) with injected i18n/fmt follow the existing `stat_tooltip_view.ts` / `chat_window.ts` patterns exactly, they are unit-tested headlessly, and the DOM wiring stays thin in hud.ts. The shared `#tooltip` reuse is handled with care: the `mob-tooltip` size class is added only on the world-hover path and stripped by every other tooltip path, so the bigger sizing never leaks onto an item/ability tooltip. CSS parity is automatic here (index.html and play.html both load the same `main.ts` bundle that imports `hud.css`), so nothing to mirror. The unrelated `architecture.test.ts` fix is a real improvement: `world_api/chat.ts` is now a genuine type-only sim import, and the Windows path-separator bug in the allowlist lookup was worth catching. tsc plus the targeted suites (mob tooltip, target frame, pet reaction, architecture, i18n completeness, css corpus/validity) all pass locally.

A few things:

**should-fix (confirm intent)** `src/render/reaction.ts:44` The con-color recalibration is shared, not tooltip-only: `mobNameColor` also backs every overhead nameplate (`nameplate_painter.ts:270`), so this PR recolors all in-world mob nameplates. The title reads "add tooltip", so it is worth flagging that the nameplate recolor rides along. The new red / orange bands (+5, and +3 to +4) now match classic, but the even band shifts: a same-level mob now cons green where classic cons it yellow (classic yellow spans -2 to +2). The body says this is per design direction, so this is just a "confirm you meant to move even-level off yellow game-wide", not a bug.

**nit** `src/ui/hud.ts:3203` `showMobHoverTooltip` gates on `entity.id`, so the model (reaction line + con-color) is snapshotted on first hover and not rebuilt while the same mob stays hovered. If a neutral mob aggros (its `hostile` flips) or the viewer dings a level mid-hover, the tooltip keeps the stale Friendly/Hostile line and color until you re-hover. Fine for the common case, just noting the tradeoff of the id gate.

**nit** `src/ui/hud.ts:2034` `applyTargetFramePos` early-returns on `isMobileLayout()`, so a live desktop-to-mobile viewport shrink (via the resize handler) leaves the desktop inline `left`/`top` on `#target-frame` rather than clearing them. Only reachable by crossing the mobile breakpoint mid-session (a fresh mobile load never sets them), so low stakes, but clearing the inline positioning on the mobile branch would be tidier.

Nice work overall.

EnriqueGF:
Excellent work!!

Overlay when hovering a mob:

image

Movable gui selected mob:

d0239f701489a1105cb7d9cdd06a6cd3

Blaine1705:
Thanks for the thorough review, and especially for flagging the shared reach of the con-color change. All three addressed in 90e186d9:

**should-fix (con-color, game-wide):** Good catch that `mobNameColor` also backs every overhead nameplate, so this was recoloring the whole world, not just the tooltip. On reflection we want to stay classic, so I restored the even band: a mob at `levelDiff -2..+2` cons **yellow** again (not green), while keeping the recalibrated `red 5+` / `orange 3-4` above bands and `green -3..-5` / `grey -6+` below. So the net effect vs. the pre-PR nameplates is only the intended top-band fix (red no longer starts at +3, orange no longer at +1), and same-level stays yellow as before. Boundary tests in `pet_reaction.test.ts` updated to pin the classic bands.

**nit (stale hover model on the id gate):** Fixed. `showMobHoverTooltip` now gates on a small composite key `id:level:hostile:playerLevel` instead of just the id, so the tooltip repaints when the mob aggros (hostile flips) or the mob/viewer dings a level mid-hover, while still skipping the rebuild on an unchanged re-hover each frame.

**nit (desktop-to-mobile inline pos):** Fixed. `applyTargetFramePos` no longer early-returns on the mobile layout; it clears any inline `left/top/right/bottom` first, so a live desktop-to-mobile viewport shrink hands the frame's position back to the mobile stylesheet instead of leaving stale desktop coordinates on it.

tsc plus the targeted suites (pet_reaction, mob_tooltip_view, target_frame_pos, architecture, hud_perf_budget) pass locally, and biome is clean on the changed files.

Blaine1705:
Follow-up on the con-color point (ed88159e): rather than land the recalibrated bands on the shared `mobNameColor`, I scoped them to the tooltip only, which is what was actually intended.

- `mobNameColor` is reverted to its shipped nameplate bands, so the overhead nameplates are unchanged by this PR (no game-wide recolor).
- The mouseover tooltip now uses a separate `mobTooltipConColor` with the classic con spread: red only at 5+ above, orange 3-4, yellow across the even band -2..+2, green 3-5 below, grey 6+ below.

The two palettes are unit-tested apart in `pet_reaction.test.ts`, including an explicit case that a +4 mob reads orange in the tooltip but red on the nameplate, so the split is pinned. This also sidesteps the "even band moves game-wide" concern entirely, since nameplates never change. tsc, the targeted suites, and biome are green.

EnriqueGF:
@FernandoX7 I'd say is ready to merge

EnriqueGF:
Closing as already landed: every commit on this branch (up to ed88159e) is contained in release/v0.20.0. The mob mouseover tooltip and the movable, lockable target frame arrived via #1369 (feature/unit-frame-customization, merged 2026-07-03), which built directly on these commits. Nothing here is lost, and this PR still targets the old release/v0.19.0 base, so merging it now would only touch a frozen branch. Thanks for the feature, it is live on v0.20.0.

Issue #722, opened 2026-06-19
Add a "Loot All" button and an auto-loot toggle


jgyy:
## Summary

Corpse looting is **one item at a time** today — each pickup is a separate `lootCorpse` action, and items land in your bags individually. After a multi-mob pull this is a lot of clicks. This proposes a **"Loot All"** action plus an **auto-loot** toggle, the standard classic-MMO looting QoL.

![current bags — loot lands here one item at a time](https://raw.githubusercontent.com/jgyy/world-of-claudecraft/assets/feature-rq-death-qol/loot-all-bags.png)

## Current state (what it builds on)

- `server/game.ts`: `case 'loot': ... sim.lootCorpse(msg.id, pid)` — looting is per-item, one command per item.
- `src/ui/hud.ts`: the bags window (`#bags` / `renderBags`) is where looted items already flow.

## Proposal

1. **Loot All button** on the loot window: a single command that pulls every item from the corpse that fits in your bags (server-authoritative; partial-loot if bags fill, with the usual "inventory full" feedback).
2. **Auto-loot toggle** in the existing Interface/Comfort settings: when on, opening a corpse loots everything immediately; **hold a modifier to invert** for that one corpse (classic behavior).

Pure command-layer + settings change; no balance/RNG impact — the same items are awarded, just in one action.

## Not a duplicate of

- No open issue covers loot-all / auto-loot (verified against the tracker and `git grep autoLoot|lootAll`). The party-loot issues (**#317**, **#284**) are about *who* is entitled to a drop; this is about *how many clicks* the entitled looter spends.

HanClinto:
Not sure this is really needed -- just press "f" over and over to loot nearby bodies. Don't need loot-all when a few presses will do.

jgyy:
Adding a mermaid diagram for Loot All + auto-loot:

```mermaid
flowchart TD
  A[Kill / open corpse] --> B{auto-loot toggle on?}
  B -- yes --> C[Server moves all eligible items to bags]
  B -- no --> D[Loot window opens]
  D --> E{click Loot All}
  E -- yes --> C
  E -- no --> F[Pick items individually]
  C --> G{bags full?}
  G -- yes --> H[Leave remainder in corpse + notice]
  G -- no --> I[Done]
```

PR #1245, opened 2026-07-01
feat(ui): add corpse auto-loot option

Humpalumps:
## Summary

Adds a default-off **Auto-Loot Corpses** Interface option for the existing corpse loot flow. When enabled, clicking a nearby lootable corpse immediately uses the existing server-authoritative `loot` / `lootCorpse` take-all path instead of opening the loot window. Holding Shift while clicking still opens the loot window for inspection.

This keeps the current **Take All** button behavior intact and fills the auto-loot portion of the requested loot quality-of-life work.

## Related issues

Part of #722

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src/game/interactions.ts src/main.ts tests/interactions.test.ts`
  - `npx biome lint src/game/interactions.ts src/main.ts src/game/settings.ts src/ui/options_view.ts src/ui/i18n.catalog/hud_chrome.ts tests/interactions.test.ts tests/options_view.test.ts tests/settings.test.ts` - passed with existing warning-level diagnostics in `src/main.ts`, older `tests/interactions.test.ts` scaffolding, and `tests/settings.test.ts`.
  - `npm run check:ts` - passed.
  - Wrapped `.browserslistrc` Vitest: `npx vitest run tests/interactions.test.ts tests/interaction.test.ts tests/settings.test.ts tests/options_view.test.ts tests/localization_coverage.test.ts tests/command_schema.test.ts tests/world_api_parity.test.ts` - 7 files passed; 287 tests passed, 3 skipped.
  - Wrapped `.browserslistrc` build: `npm run build` - passed with existing Vite/admin generated-locale dynamic import, chunk size, and plugin timing warnings.
  - `git diff --check` - passed.
- Manual steps:
  - Verified the Interface options row renders as `Auto-Loot Corpses` with the toggle off by default.
  - Captured the screenshot below from the options UI.
  - Restored unrelated generated build outputs after validation; committed only the UI i18n generated files related to the new string.
  - Full `npm test` was not run for this PR; prior full-suite runs on this release line had unrelated release blockers/timeouts, so this PR uses the focused interaction/settings/options/localization coverage listed above.

## Screenshots / recordings

Auto-Loot Corpses option

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Clean, well-scoped change. It reuses the existing server-authoritative loot path (`lootCorpse` sends `{cmd:'loot'}`, and `sim.lootCorpse` re-checks permission, range, and FFA/personal/shared rights), so the client stays a pure renderer with no new authority surface. Default-off is the right call and matches classic WoW, where the `autoLootDefault` CVar ships at 0. The new options param is optional, so the other callers are untouched, and the toggle inherits keyboard and aria from `boolToggle`. The two added tests cover the auto-loot and force-window branches.

**should-fix**

Shift should invert the setting, not only force the window open (`src/game/interactions.ts:109` and `:135`, `src/main.ts:1755`). In WoW the auto-loot modifier is a two-way toggle: with auto-loot off, holding Shift quick-loots the corpse; with it on, holding Shift opens the loot window. `forceLootWindow` only does the second half, so a player on the default (off) still can't Shift-click to grab a single corpse, which is the more common muscle memory. It's about a one-liner: compute something like `const takeAll = !!opts.autoLootCorpses !== !!opts.shiftHeld` and branch on that instead of the current one-way flag. (Warcraft Wiki, Autoloot: "If the setting is enabled, Shift-click would perform the opposite by showing the loot without autolooting.")

**nits**

Mobile has no inspect path once auto-loot is on (`src/main.ts:1756`). The override only reads `ShiftLeft`/`ShiftRight`, so on touch every corpse tap takes all with no way to open the loot window. Fine for an opt-in comfort toggle, but worth a note if you want parity, e.g. a long-press.

Test spacing and coverage (`tests/interactions.test.ts`, around the new block). Missing the blank line before the following `it(...)`, and there is no case for "Shift held while the setting is off" (which matters if you pick up the inversion above).

**notes**

You only needed the English key in `hud_chrome.ts`. The locale overlays and the `i18n.resolved.generated` / `i18n.status.summary.json` files are filled and regenerated at release, so you can drop those 40-odd files; no translation work is needed from your side. As they stand they are also slightly inconsistent: the resolved set and summary were regenerated (new `universeHash`) but `i18n.resolved.sha256` was not, so the i18n freshness check will flag them until a clean i18n build is run.

The PR notes the full `npm test` was not run; the focused interaction, settings, and options suites cover this change, but it is worth a full run before merge. No em or en dashes or emojis introduced, and no `play.html` CSS parity concern since it reuses the existing toggle component.

Humpalumps:

Issue #15, opened 2026-06-12
FEAT RQ: A List of QoL we could do.


MerlinStacks:
Sorry in advance for the long list, I understand this may be just personal taste for some of the QoL.

- Have the 'heal / rest' sounds stop when the players health and mana are full.
- When Mana is completely empty have a graphic appear on the player model or centre screen saying so to avoid confusion as the current text is not as noticable as it could be.
- Seperate Chat, World Notices (Players logging in and out) and Combat Logs into seperate Tabs.
- After a brief period (roughly 4s) hide unlootable corpse, for visual cleanup and to avoid player confusion from incorrect permissions, or allow after 10s for anyone to loot the corpse.
- Allow for clicking and dragging of the UI elements (Bag, Character, Quest Book).

jgyy:
Status check against current main, trimming this QoL list to what is still open:

Done (shipped):
- Chat tabs / channel separation (`src/ui/chat_channels.ts`, chat tabs in the HUD)

Still open (the actionable remainder of this issue):
- Stop the heal/cast sound when already at full HP/mana
- Empty-mana visual indicator on the resource bar
- Auto-hide / suppress an unlootable corpse you cannot loot

Keeping the issue open scoped to the three items above.

PR #1253, opened 2026-07-01
fix(sim): stop food and drink at full

Humpalumps:
## Summary

Food and drink consumption now stops independently on the next regen tick once the relevant resource is already full. This prevents a finished meal/drink from continuing to drive the rest/eat/drink state after it has nothing left to restore.

The change is intentionally scoped to the sim regen path: food stops when health is full, drink stops when mana is full, and one can continue while the other has work left.

## Related issues

Part of #15.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src\sim\combat\auras.ts tests\sim.test.ts` - passed
  - `npx biome lint src\sim\combat\auras.ts tests\sim.test.ts` - passed with existing warnings in `tests/sim.test.ts`
  - `npm run check:ts` - passed
  - wrapped `.browserslistrc` command: `npx vitest run tests/sim.test.ts tests/items.test.ts tests/cast_bar.test.ts tests/rest_indicator.test.ts` - passed, 4 files / 129 tests
  - wrapped `.browserslistrc` command: `npx vitest run tests/parity/parity.test.ts tests/parity/coverage.test.ts` - passed, 2 files / 140 tests
  - wrapped `.browserslistrc` command: `npm run build` - passed with existing Vite warnings about admin locale dynamic imports and large chunks
- Manual steps: N/A, sim-only behavior covered by tests.
- Notes: an initial local attempt to run two wrapped commands in parallel failed because both tried to rewrite `.browserslistrc`; the commands were rerun sequentially and passed. Full `npm test` was not run for this focused PR; the release branch still has known unrelated full-suite blockers from current triage.

## Screenshots / recordings

N/A - sim state change only, no UI layout change.

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/combat/auras.ts
+++ b/src/sim/combat/auras.ts
@@ -81,6 +81,14 @@ export function updateRegen(ctx: SimContext, p: Entity, _meta: PlayerMeta): void
   for (const slot of ['eating', 'drinking'] as const) {
     const c = p[slot];
     if (!c) continue;
+    const restoresHealth = c.hpPer2s > 0;
+    const restoresMana = c.manaPer2s > 0 && p.resourceType === 'mana';
+    const healthFull = !restoresHealth || p.hp >= p.maxHp;
+    const manaFull = !restoresMana || p.resource >= p.maxResource;
+    if (healthFull && manaFull) {
+      p[slot] = null;
+      continue;
+    }
     if (c.hpPer2s > 0 && p.hp < p.maxHp) {
       const heal = Math.min(Math.round(c.hpPer2s * ctx.healingTakenMult(p)), p.maxHp - p.hp);
       p.hp += heal;
--- a/tests/sim.test.ts
+++ b/tests/sim.test.ts
@@ -828,6 +828,30 @@ describe('food, drink, vendor', () => {
     expect(sim.player.drinking).toBe(null);
   });
 
+  it('stops food and drink independently once the relevant bar is full (#15)', () => {
+    const sim = makeSim('mage');
+    sim.addItem('baked_bread', 1);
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Clean, well-scoped fix. The stop-at-full logic reads correctly: `restoresHealth` / `restoresMana` gate on what the consumable actually restores (and on `resourceType === 'mana'` for the drink side), so a health-only food stops when HP is full, a mana drink keeps going, and a combined item would only stop once both bars are topped off. It mirrors the existing "nothing to restore" guard on the potion path in `items.ts`, and the next-tick timing matches the summary. Determinism and sim purity are untouched: no new randomness, wall-clock, or imports, and the loop stays over the fixed `['eating', 'drinking']` slots, so there is no three-host drift (the slot was already the source for the client `eat`/`drk` state). The new test is seeded and covers both the independent stop and the eventual full clear; the food/drink and rest-indicator suites stay green here.

One small note, not blocking:

- nit `src/sim/combat/auras.ts:84-88`: for a non-mana class (rage/energy) that drinks a mana drink, `restoresHealth` and `restoresMana` are both false, so the slot is cleared on the first regen tick (~2s) instead of sitting the full 18s. Nothing in the consume path gates food/drink by class, so it is reachable, but the new behavior is arguably better than staying seated restoring nothing. Flagging it only as an uncovered behavior change for that path.

Humpalumps:

PR #1254, opened 2026-07-01
perf(client): cut per-frame work and GC churn across render, hud, net and input hot paths

EnriqueGF:
## Summary

Behavior-preserving, visually identical performance pass over the client hot paths, in two commits: steady-state frame cost + GC churn first, then freeze/hitch sources. Both from four-way audits (renderer, sim tick, HUD per-frame, client loop + net decode) with each candidate verified for exact behavioral equivalence before landing.

### Renderer (src/render/renderer.ts)
- Form-swap detection (polymorph / bear / cat / ghost wolf / travel / stealth) collapsed from six per-entity `.some()` scans per frame to one pass; the flag combination preserves the original precedence exactly.
- Loot-sparkle visibility reuses the entity's already-computed squared distance.
- The selection ring's terrain drape (a per-vertex `groundHeight` noise resample, previously every frame) is cached by `(x, z, scale)`; a stationary target skips it entirely. Ring spin, pulse and hostile color still animate per frame.

### Input (src/game/interactions.ts + src/main.ts)
- New `HoverPickGate` (pure, unit-tested): the hover-cursor scene raycast used to run against every rig every frame even with the mouse still. Now it re-picks on any pointer move (instantly) or at most every 50 ms while stationary. The cursor KIND still re-resolves every frame from live entity state.

### Net decode at 20 Hz (src/net/online.ts)
- `applySnapshot` reuses one instance `seen` Set; `threat` and `cooldowns` update in place (delta-guard semantics unchanged, consumers verified to read inline).

### HUD (src/ui/hud.ts)
- Compass early-returns when facing is unchanged (`compassView` is a pure function of facing).
- Combo pips: indexed loop instead of a per-frame array spread (still through the elided `toggleClass` writer).
- Tooltip `mousemove` reuses the box size measured in `showAt` instead of forcing a reflow per event.

### Misc
- `src/game/utc_day.ts` (new): the offline sim's injected `utcDay` string re-derives at most once per second instead of `new Date().toISOString()` per rAF frame.
- `src/sim/sim.ts`: `updateLootRolls` skips the defensive copy on the common empty tick.

## Commit 2: freezes and allocation bursts (a8e3dc4f)

- **Icon cache warmer** (`src/ui/icon_prewarm.ts`, new): the first vendor/bags/loot open paid a synchronous 60-300 ms burst composing every visible item icon (canvas + `toDataURL` per unique icon). The warmer drains the whole item + ability catalog through `requestIdleCallback` slices (8 per slice, `setTimeout` fallback, same pattern as the HUD map prewarm) right after boot, so the shared `urlCache` is hot before any window opens. Purely a cache warmer: rendering is byte-identical.
- **Aura decode fast path** (`src/net/online.ts`): when the wire aura ids line up index-for-index with the existing records (the common case: only `remaining` ticks down between snapshots), the records update in place; no fresh array + object per aura per entity at 20 Hz. Composition changes (gain/fade/reorder/substitution) still rebuild. The preserved object identity matches the offline Sim contract (one live aura object across ticks), so the online mirror is now MORE faithful, not less. A cross-platform-sync review confirmed field-for-field lockstep with the `wireEntity` encoder (12/12 fields MATCH, absence semantics preserved).
- **Per-school projectile color cache** (`src/render/vfx.ts`): three `THREE.Color` allocations per bolt launch eliminated; `spawn()` only copies components, never retains the reference. Keyed on `GFX.composer` so a tier flip rebuilds the HDR variants.

## Deliberately rejected during review
- Aura-derived stat caches (armor/AP/move speed): real cache-invalidation risk for marginal gain.
- Cleave via the spatial grid: changes target iteration order, which shifts rng draw order.
- Reusing the renderer telemetry objects: they escape `sync()` into the perf sample.
- A `stepProjectile` squared-distance compare was applied and then reverted: the architecture reviewer proved it differs from the sqrt compare by 1 ulp at the boundary and it only saved one sqrt per projectile lifetime.

## Verification
- Parity gate (golden traces + rng draw order), `architecture`, `snapshots` (incl. 5 new fast-path pins: identity reuse, field updates, rebuild on gain/shrink, same-length reorder, optional-field clearing), `bandwidth`, `hud_perf_budget`, `painter_host`, `alloc_probe`, `client_shell`, `interactions`, `icon_prewarm`, `utc_day`: all green.
- `tsc --noEmit` clean; Biome error-level clean on changed files.
- QA gate (`qa-checklist`) on both commits: READY, 0 blocking. `architecture-reviewer` and `cross-platform-sync` passes clean (the former drove the projectile revert).
- `npm run perf:prewarm` (prewarm travel bench): all dungeon interiors at +0 program growth; open-world program growth is confined to known waypoints (see follow-ups).
- `npm run perf:tour`: prewarm 4.6 s of the 12 s budget (17/17 steps, 0 failures), HUD hot-write skip rate 98%, FCT burst cap-bounded.
- Browser smoke test on the offline world: loads and plays normally; compass/target/selection ring/tooltips behave identically.
- The 8 full-suite failures observed locally are pre-existing 5 s timeout flakes under load: identical failures reproduce on a clean v0.18.0 checkout on the same machine, and every affected file passes in isolation.

## Follow-up candidates (out of scope here)
- Prewarm coverage gaps surfaced by the travel bench (first-sight shader compiles at `vale.north-edge` +17 programs, `marsh.north-edge` +10, `marsh.entry` +4; benign pop-in on async-compile GPUs, a real stall on sync-compile clients). Extending `prewarmInitialScene` there must be weighed against its 12 s budget.
- Bags/market full DOM rebuild per search keystroke (~75 KB churn); keyed reconciliation like `auras_painter`.
- FCT spawn-shape descriptor memoization for AoE event bursts.

### DIFF
--- a/src/game/interactions.ts
+++ b/src/game/interactions.ts
@@ -45,6 +45,29 @@ export function activePvpOpponentIds(
   return ids;
 }
 
+// Re-pick cadence for the hover cursor while the pointer is stationary. A pointer
+// move always re-picks immediately; this only bounds how fast the world can change
+// WHICH entity sits under an unmoving cursor (a walking mob), so the scene raycast
+// stops costing a full intersect pass on every frame of a still mouse.
+export const HOVER_REPICK_MS = 50;
+
+/** Gate for the per-frame hover raycast: pick when the pointer moved, otherwise at
+ *  most every HOVER_REPICK_MS. Pure state machine (caller supplies the clock), so
+ *  it unit-tests without DOM or timers. */
+export class HoverPickGate {
+  private x = Number.NaN;
+  private y = Number.NaN;
+  private nextAt = 0;
+
+  shouldPick(x: number, y: number, nowMs: number): boolean {
+    if (x === this.x && y === this.y && nowMs < this.nextAt) return false;
+    this.x = x;
+    this.y = y;
+    this.nextAt = nowMs + HOVER_REPICK_MS;
+    return true;
+  }
+}
+
 export function isAttackableEntity(
   e: Entity | undefined,
   playerId: number,
--- a/src/game/utc_day.ts
+++ b/src/game/utc_day.ts
@@ -1,0 +1,16 @@
+// The offline sim wants the wall-clock UTC day (delve daily reset) but must not
+// read the clock itself, so the frame loop supplies it. Building the string is a
+// Date allocation plus an ISO serialization; at 60 Hz that is pure churn for a
+// value that changes once a day, so cache it and re-derive at most once a second.
+let cachedDay = '';
+let nextRefreshAtMs = 0;
+
+/** Current UTC day as `YYYY-MM-DD`, recomputed at most once per second. */
+export function currentUtcDay(): string {
+  const now = Date.now();
+  if (now >= nextRefreshAtMs) {
+    cachedDay = new Date(now).toISOString().slice(0, 10);
+    nextRefreshAtMs = now + 1000;
+  }
+  return cachedDay;
+}
--- a/src/main.ts
+++ b/src/main.ts
@@ -142,6 +142,7 @@ import {
   t,
   tPlural,
 } from './ui/i18n';
+import { defaultIconPrewarmEntries, prewarmIconCache } from './ui/icon_prewarm';
 import { iconDataUrl } from './ui/icons';
 import { createMetricsSampler } from './ui/perf_metrics_sampler';
 import { PerfOverlay } from './ui/perf_overlay';
@@ -2393,6 +2394,10 @@ async function startGame(
           tokenProvider: () => api.token,
           characterIdProvider: () => online?.characterId ?? null,
         });
+        // Warm the procedural icon cache during idle time so the first
+        // bags/vendor/loot open never pays the compose burst synchronously
+        // (icon_prewarm.ts). Re-entry is a fast no-op: the cache is module-global.
+        prewarmIconCache(defaultIconPrewarmEntries());
         (window as any).__game = {
           sim: world,
           world,
--- a/src/main.ts
+++ b/src/main.ts
@@ -26,6 +26,7 @@ import { Input } from './game/input';
 import { InputActivityMeter, installInputActivityTracking } from './game/input_activity';
 import {
   activePvpOpponentIds,
+  HoverPickGate,
   handlePickedEntity,
   hoverCursorKind,
   isAttackableEntity,
@@ -52,6 +53,7 @@ import {
 } from './game/settings';
 import { sfx } from './game/sfx';
 import { resolveUiEffectsProfile } from './game/ui_effects_profile';
+import { currentUtcDay } from './game/utc_day';
 import { voice } from './game/voice';
 import {
   CHAR_SORT_MODES,
@@ -2094,13 +2096,22 @@ async function startGame(
     return ids;
   }
 
+  // The scene raycast is the expensive half of the hover cursor; the gate re-picks
+  // on pointer movement (instantly) or every HOVER_REPICK_MS while stationary. The
+  // cursor KIND below still re-resolves every frame from live entity state, so a
+  // hovered mob dying or turning hostile updates without waiting for a re-pick.
+  const hoverPickGate = new HoverPickGate();
+  let hoverPickedId: number | null = null;
+
   function updateHoverCursor(): void {
     if (!input.hoverActive || input.isDragging() || hud.isModalOpen()) {
       input.setHoverCursor('default');
       return;
     }
-    const id = renderer.pick(input.hoverX, input.hoverY);
-    const entity = id !== null ? world.entities.get(id) : undefined;
+    if (hoverPickGate.shouldPick(input.hoverX, input.hoverY, performance.now())) {
+      hoverPickedId = renderer.pick(input.hoverX, input.hoverY);
+    }
+    const entity = hoverPickedId !== null ? world.entities.get(hoverPickedId) : undefined;
     input.setHoverCursor(
       hoverCursorKind(entity, world.playerId, partyMemberIds(), activePvpOpponentIds(world)),
     );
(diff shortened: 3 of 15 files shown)


EnriqueGF:
Third commit (e9915b4a), from live profiling: an immortal walk to the north village via scripts/profile.mjs walk plus a V8 CPU sampling capture on the same harness (real GPU, headed).

- The top CPU self-time was Three's updateMatrixWorld + multiplyMatrices (~15% combined): r165 recomposes every node's matrices per frame, including ~1250 never-moving terrain/water/prop nodes. New src/render/static_matrix.ts freezes fully-built static subtrees (terrain group, water meshes, props tree; campfire flames re-enabled, they rescale per frame). Probe after: 1252 nodes frozen, updateMatrixWorld self-share ~10% -> ~7%, town scene pixel-identical (flames still flicker).
- The same profile caught a defect in the new icon warmer: the idle deadline was checked between batches of 8 icons, letting one requestIdleCallback overrun by 60-100ms (unattributed long tasks in the first minute). Now a 6ms wall-clock budget is enforced before EVERY icon (injectable clock, pinned by tests).

Remaining known hot spots from the profile, for future passes: animated rig matrices (43 views x ~40 nodes, the legitimate half of the matrix cost; culled/offscreen rigs could skip their subtree via matrixWorldAutoUpdate with a forced update on re-show), draw submission (setProgram/refreshUniforms), and terrainHeight/baseHeight sampling (~2%).

EnriqueGF:
Fourth commit (aa1d4710), from crowd profiling on a live server (Postgres + game server with dev commands, 30 harness bots clustered around the client, real GPU, V8 sampling profiles):

- Matrix work was 14.6% of client CPU in a crowd. Root cause beyond the earlier static freeze: the scene ROOT's default matrixAutoUpdate dirties it every frame and force-cascades a matrixWorld multiply through the entire graph (three r165 Object3D.updateMatrixWorld), so the static freeze was only saving the compose half. The root is now frozen (it sits at identity forever); auto-updating children recompose themselves exactly as before.
- Hidden views (distance-culled, off-screen, compile-gated) now gate their whole subtree via matrixWorldAutoUpdate=false, skipping 30-60 dead node updates per hidden rig per frame. Re-show reopens the gate in the same sync pass, before render, so nothing ever renders stale (verified with a 360-degree camera-spin hide/show cycle: every rig back at its live position, gate/visible mismatch 0).
- pick() skips raycast hits on hidden views (invisible rigs should not be clickable, and a gated group's frozen matrixWorld must not ghost a hitbox at its hide-time position).
- A fresh-eyes diff review caught one real regression before commit: the light-budget rebuild caches light.getWorldPosition for view-owned lights (quest-object glows), and r165's updateWorldMatrix does not heal the parent chain through a gated ancestor. Light-owning views are therefore exempt from the gate.

Measured in-crowd (30s V8 samples, ~equal bot counts): multiplyMatrices 6.35% -> 4.76%, combined matrix share 14.6% -> 12.7%, fps 79.9 -> 82.5 on the test box.

Also in this branch since the last comment: the icon warmer now enforces its 6ms budget per icon (a CPU profile showed a requestIdleCallback batch could overrun to 60-100ms, exactly the freeze class this PR targets).

FernandoX7 (commented):
Reviewed the full diff against `release/v0.19.0` (clean merge, no conflicts). This is a careful, equivalence-first performance pass and it holds up. The behavioral-equivalence claims are verified in the code, not just asserted, and the tests land on exactly the edges that could have broken. Nothing blocking, no findings.

**What I checked independently**

Render
- Form-swap collapse (`renderer.ts:3860`): the one-pass flag set is identical to the six `.some()` scans, and the precedence algebra (poly > bear > ghostWolf/cat > travel) is preserved exactly.
- Sparkle cull (`renderer.ts:3966`): `d2` at `:3877` is exactly `cdx*cdx + cdz*cdz` over `e.pos - p.pos`, the same value the old `sdx/sdz` compare used.
- Selection-ring drape cache (`renderer.ts:4211`): keyed on `(cx, cz, scale)`, which are the drape's only inputs; ring visibility is still set outside the guard, so a re-target or a lost target still shows/hides correctly.
- Scene-root and static-subtree freeze: correct r165 semantics (root frozen so it stops force-cascading the multiply, children with auto-update still recompose, flames re-enabled), and nothing writes `scene.position/rotation` after the freeze.
- Hidden-view matrix gate (`renderer.ts:4201`): every world-space read of a view child is accounted for. Nameplates use `v.group.position` (`:4800`) and gate on `visible`, the VFX anchor uses `v.group.position` (`:1312`), `pick()` skips hidden views (`:4849`), and light owners are exempted. That is the complete set, so no gated node can ghost a hitbox or a nameplate.

Net / parity
- The aura fast path updates every wire-derived field in the in-place branch (only `id`, the gate, and `sourceId`, always 0, are omitted), clears optionals that drop off the wire, and rebuilds on any gain/shrink/reorder. Preserving object identity matches the offline `Sim`, so the mirror gets more faithful, not less. `threat`/`cooldowns` in place yield the same contents as the old `new Map(...)`, and `blankEntity` inits both as Maps so a first-sight entity is safe. The reused `seen` set is fine: `applySnapshot` is synchronous and clears it at entry.
- The snapshot tests pin identity reuse, the reorder rebuild, and the optional-clearing case, which are the three ways this could have gone wrong.

`vfx.ts` is mostly Biome reflow; the real change is the projectile color cache, and the cached instances are only ever read (the impact path does `tmpColor.copy(pr.color)`), so the shared references never get mutated. Composer-keyed invalidation handles a tier flip.

Ran against the PR head locally: parity gate, snapshots, architecture, bandwidth, hud_perf_budget, alloc_probe, painter_host, world_api_parity, interactions, icon_prewarm, utc_day, client_shell all green.

The projectile-color revert and the light-owner exemption are the two calls I'd single out; both would have been quiet visual bugs if missed.

FernandoX7 (approved):

PR #1259, opened 2026-07-01
fix(render): disambiguate stacked corpse picks

### LINKED ISSUE #1257: Overlapping corpses share one clickable area, second corpse is unlootable
## Problem

When two mobs die on the same spot, their corpses overlap almost exactly and share a single clickable area. Clicking that area always resolves to the same corpse, so the second corpse can never be selected or looted. Its loot is stranded until the corpse despawns.

The two mobs did not even use the same model, but their loot click targets were identical, which points at the corpse pick/target resolution rather than the visual geometry.

## Steps to reproduce

1. Pull two mobs to the same location.
2. Kill them so both die on top of each other (corpses at nearly the same position).
3. Loot the top corpse and close the loot window.
4. Click the corpse to loot the second one.
5. Observe that the click keeps resolving to the already-looted corpse, never the second.

## Expected behavior

- Overlapping corpses are individually selectable and lootable.
- Clicking a spot with more than one corpse resolves to a lootable corpse (for example, cycle to the next unlooted corpse on repeated clicks, or offer a way to disambiguate stacked corpses).
- Every corpse with loot can be reached regardless of how closely it overlaps another.

## Actual behavior

- The clickable area for both corpses is identical.
- Every click resolves to the same (first/top) corpse.
- The second corpse cannot be selected or looted, and its loot is unrecoverable until despawn.

## Scope

- Corpse click/target resolution in the client (how a click position maps to a corpse when multiple overlap).
- Loot interaction flow after the top corpse is looted or its loot window is closed.
- Whether the pick logic considers all corpses under the cursor or only the first hit.

## Acceptance criteria

- With two or more corpses stacked at the same position, each corpse can be selected and looted.
- No corpse loot is left unrecoverable purely because another corpse shares its position.
- Repeated clicks (or an equivalent disambiguation) reach every unlooted corpse in the stack.
- The single-corpse loot experience is unchanged.

## Screenshot

Image

Humpalumps:
﻿## Summary

- Collect all direct raycast entity hits before resolving a click target.
- Cycle stacked lootable corpses when the currently targeted corpse is clicked again, so repeated clicks can reach every corpse in an overlapping stack.
- Preserve existing single-corpse behavior and unlootable object blocking.

## Related issues

Closes #1257

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx biome format --write src\render\pick_resolution.ts tests\pick_resolution.test.ts`
  - `npx biome lint src\render\pick_resolution.ts src\render\renderer.ts tests\pick_resolution.test.ts` (exits 0; existing renderer non-null assertion warnings remain)
  - `.browserslistrc` comment-strip wrapper around `npx vitest run tests\pick_resolution.test.ts tests\sloppy_pick.test.ts tests\interactions.test.ts` (3 files, 26 tests passed)
  - `npm run check:ts`
  - `.browserslistrc` comment-strip wrapper around `npm run build` (passes; existing Vite dynamic-import/chunk/plugin timing warnings remain)
- Manual steps: N/A; covered with focused pick-resolution and existing interaction/sloppy-pick tests.

## Screenshots / recordings

N/A; this is click-target resolution logic with no visible UI design change.

---

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/render/pick_resolution.ts
+++ b/src/render/pick_resolution.ts
@@ -1,0 +1,33 @@
+import type { Entity } from '../sim/types';
+
+type PickEntity = Pick<Entity, 'id' | 'kind' | 'dead' | 'lootable'>;
+
+function lootableCorpse(e: PickEntity): boolean {
+  return e.kind === 'mob' && e.dead && e.lootable;
+}
+
+export function resolveDirectPickEntityId(
+  hitEntityIds: readonly number[],
+  entities: Pick<Map<number, PickEntity>, 'get'>,
+  currentTargetId: number | null = null,
+): number | null {
+  const ordered: PickEntity[] = [];
+  const seen = new Set<number>();
+  for (const id of hitEntityIds) {
+    if (seen.has(id)) continue;
+    seen.add(id);
+    const e = entities.get(id);
+    if (!e) continue;
+    if (e.kind === 'object' && !e.lootable) return null;
+    if (e.kind === 'mob' && e.dead && !e.lootable) continue;
+    ordered.push(e);
+  }
+  if (ordered.length === 0) return null;
+
+  const corpses = ordered.filter(lootableCorpse);
+  if (lootableCorpse(ordered[0]) && corpses.length > 1 && currentTargetId !== null) {
+    const idx = corpses.findIndex((e) => e.id === currentTargetId);
+    if (idx >= 0) return corpses[(idx + 1) % corpses.length].id;
+  }
+  return ordered[0].id;
+}
--- a/src/render/renderer.ts
+++ b/src/render/renderer.ts
@@ -80,6 +80,7 @@ import {
   nameplateScreenTransform,
 } from './nameplate_projection';
 import { PlacedAssetsView } from './placed_assets';
+import { resolveDirectPickEntityId } from './pick_resolution';
 import { buildComposer, type PostPipeline } from './post';
 import { buildPropMaterialPrewarmGroup, buildProps } from './props';
 import { buildGroundQuestObject } from './quest_objects';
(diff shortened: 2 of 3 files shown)


Nervescraper (commented):
Clean, well-scoped fix. Collecting all direct raycast hits before resolving is the right shape, the pure `pick_resolution.ts` is testable without a GL context, and it preserves the non-lootable-object block. I reproduced the original bug locally (two mobs dying on the same spot: the front corpse's click proxy keeps intercepting every click even after it is looted, so the corpse behind stays unreachable), and the collect-then-skip change addresses exactly that. Verified against `release/v0.19.0`: merges clean, `tsc` clean, and `tests/pick_resolution.test.ts` plus the neighbours and `architecture.test.ts` are green (59 tests). The skip-through is the load-bearing fix; the corpse cycling is a good addition for the case where both corpses are still full.

Notes, none blocking:

- nit (`tests/pick_resolution.test.ts`): the fallthrough cases aren't pinned. Worth one assertion that an empty hit list and an all-looted stack both return `null`, since `src/render/renderer.ts:4790` keys the sloppy-assist fallback on `directHitIds.length > 0`, and that null-versus-fallthrough boundary is the one behaviour the unit test can lock down.
- note (cycling UX): a left-click on a lootable corpse also opens the loot window at the click point (`src/game/interactions.ts:130`, `hud.openLoot`). If that window sits over the corpse, the next click can land on the panel instead of the canvas and never re-pick, which would blunt the A to B cycling. Worth a quick manual check that cycling works in-game with the loot window open. The core fix does not depend on it (once A is looted it is skipped and B is reachable), so this is only about the cycling nicety.
- note (behaviour change): a click that hits only looted corpses now returns `null` and suppresses the sloppy assist, where base would have targeted the top corpse (`src/render/renderer.ts:4790`). That reads as an improvement (an empty corpse is not a useful click target), just calling it out.

Nice work.

Humpalumps:

Nervescraper:
@Humpalumps this change is needed, please reopen and rebase to latest release branch.

Humpalumps:
@Nervescraper  done :)

PR #1261, opened 2026-07-01
feat(loot): walk-by proximity autoloot with round-robin party loot

Nervescraper:
## Summary

Loot corpses by walking past them (no clicking), and make shared common/junk drops fair in a party via round-robin, so proximity loot does not hand all the trash to whoever is standing on the corpse.

Three commits:

1. **`feat(loot): round-robin common-item distribution`** - a new `round-robin` common-item loot strategy, now the default for common/junk drops.
2. **`feat(loot): walk-by proximity autoloot`** - the walk-by trigger, server-authoritative, plus the raid-instance gate.
3. **`test(loot): master-loot timeout test budget`** - an unrelated flake fix (see below), kept separate.

## Round-robin common-loot fairness

Shared common/junk drops now rotate over the **kill-time eligible party members** (`partyLootCandidatesForMob`, backed by `mob.lootRecipientIds` snapshotted at death), not the loot-time in-range set. That is the whole point: a ranged player who was in the fight but is standing back from the corpse still gets their turn in the rotation, instead of the melee player on the corpse taking every scrap.

- The cursor is deterministic party state (`party.lootTurn`), advanced once per awarded item. No RNG, so the sim draw-order and parity are unchanged.
- Copper still fair-splits, quest copies stay personal-per-needer, and premium items still go to need/greed or master loot. Round-robin only ever applies to poor/common quality.
- `tryAwardItemByRoundRobin` mirrors the shipped `tryAwardCopperByFairSplit`.

## Walk-by autoloot

A client-side `AutoLoot` pass (per-corpse cooldown) calls `world.autoLoot(id)`; the sim's `autoLootForParty` does a **silent** tap-rights pre-check (emits no error on a passive pass) and delegates to the existing `lootCorpse` distribution, so copper, quest copies, round-robin common items, and need/greed or master-loot rolls on premium drops all route exactly as manual looting does. `autoLoot` is added to `IWorld` and implemented in both the offline `Sim` and the online `ClientWorld`; the server handles the new `autoloot` command. The server loots as the sending player. **No wire-format change** (round-robin resolves server-side; the client pass reads loot fields already on the wire).

Manual right-click loot is unchanged.

### Opt-in per player (default off)

Walk-by autoloot is off by default. Auto-grabbing loot just by walking past a corpse can feel jarring, so classic deliberate looting stays the default and players opt in via a single `walkByAutoloot` toggle in the Esc options menu. The toggle is a purely client-side preference that gates whether the client `AutoLoot` pass runs; it needs no server, sim, or wire change. (This replaces the idea of a party/raid leader control, which would have added a decision the raid leader should not have to make.)

### Never auto-loots another player's corpse

Walk-by only auto-loots loot you have a genuine claim to: your own tap, your party's tap, an untapped corpse, personal (quest) drops, and returned open-to-all items. It never auto-grabs a corpse tapped by a non-party player, even after that corpse ages into the free-for-all window, because auto-grabbing another player's loot reads as hostile. A deliberate manual right-click loot still honors FFA. This is enforced authoritatively in the sim: `autoLootForParty` passes `ffaUnlocked=false` through both the eligibility pre-check and the delegated distribution (via a new `honorFfa` flag on `lootCorpse`), and the shared `corpseLootRights` helper keeps the manual and walk-by rights checks from drifting.

### Raid-instance gate

Proximity autoloot is disabled while the triggering player is **physically inside a raid instance** (position-derived via `instanceInfoAt` + the existing `RAID_ALLOWED_DUNGEON_IDS` set), **not** by raid-group membership. This stops both a rogue looter leaving the raid to loot and open-world autoloot firing while a raid party stages pre-pull in the open world. The sim gate is authoritative; the client pass carries a best-effort mirror to avoid command spam.

## Behavior change to note

The default `commonItems` party-loot strategy flips from `looter-takes-all` to `round-robin` for **every** party, not just an opt-in. This is intended (it is the fairness fix), but it is a global default change: parties that previously funneled all common drops to the looter now rotate them. No parity golden shifted (none exercises party common-item looting), and the one pre-existing test that assumed the old default was updated to assert the new one.

## Flake fix (separate commit)

`tests/loot_master_sim.test.ts` has two tests that advance the full 300s master-loot curate window by ticking the real sim 4800 to 6020 times (about 1.6 to 2.0s of CPU each in isolation). Under the parallel-suite fork contention their wall-time intermittently crossed Vitest's 5000ms default and timed out, so the file flaked in full runs while passing in isolation. Confirmed by root-cause investigation (at `--test-timeout=1800` both fail with a timeout, not an assertion). Fixed with an explicit 30s per-test budget via a shared documented constant. No sim-clock forking, so fidelity and determinism are unchanged.

## Testing

- Full suite green: 604 files, 6366 passed, 11 skipped, 0 failed.
- `npx tsc --noEmit` clean; `npm run build` succeeds.
- New sim tests (round-robin rotation over the kill-time set, per-item spread, solo decline, premium exemption, silent-gate autoloot, raid-instance gate, refusing a stranger's FFA corpse while manual loot still works) in `tests/social.test.ts`; new client-pass tests (eligibility, cooldown, prune-on-despawn, raid-instance guard) in `tests/autoloot.test.ts`; the options toggle covered in `tests/options_view.test.ts` and its default in `tests/settings.test.ts`.
- i18n: the `walkByAutoloot` label ships English plus its five non-Latin fills (wordy-value rule); regenerated resolved tables are committed and verified idempotent. The non-Latin translations are a best effort and worth a maintainer polish pass.


Nervescraper:
Not ready for review yet, I need to add a config option (default OFF) for this.

Nervescraper (commented):
Reviewed against the sim, wire, and UI invariants. No blocking issues. The autoloot command is wired end to end in lockstep (IWorld to both worlds, server dispatches as the sender pid, no new wire field), the round-robin cursor is deterministic party state with no new rng draw so the parity golden is untouched and it rotates over the kill-time eligible set, the silent gate emits nothing on a passive pass, and the new tests exercise real paths rather than trivially passing.

Findings, all addressed in 1271d510:

- should-fix, `src/sim/interaction.ts`: the loot-eligibility block was duplicated between `lootCorpse` and `autoLootForParty`, so the two had to stay identical by hand. Extracted into a shared `corpseLootRights` helper.
- behavior, walk-by vs FFA: a passive walk-by should not auto-grab a corpse tapped by a non-party player once it ages into the free-for-all window (that reads as hostile). `autoLootForParty` now passes `ffaUnlocked=false` through both the pre-check and the delegated distribution (new `honorFfa` flag on `lootCorpse`); a deliberate manual loot still honors FFA. Covered by a new test (walk-by refuses the stranger's FFA corpse silently, manual loot on it still works).
- nit, tests: added an assertion pinning the `walkByAutoloot` default (off).

Notes, no change needed: the single per-party `lootTurn` cursor skews slightly across kills with different eligible-set sizes (this is the intended party-wide rotation, not a bug), and the one-line `main.ts` gate is not unit-tested (it is a call site in the firewall monolith).

Heads-up for players: the default `commonItems` strategy flips from looter-takes-all to round-robin for every party, so this changes existing party loot behavior, not just walk-by.

PR #1281, opened 2026-07-02
feat(professions): add Battlefield Experience xp from observed self-crafted item use (#1149)

jgyy:
## Summary

Implements #1149 (Battlefield Experience) for the professions epic (#1152): observing a crafted item in a tracked combat/loot moment grants a small skill trickle to the craft that made it. This PR covers exactly the **self-observation case** (the wielder is also the crafter) per the issue's own guidance that this is "the simplest case and should be implemented and tested first: it needs no radius/party logic at all, just the per-item attribution."

**Scoping note:** party/raid observation-radius weighting (a grouped member or bystander within the ~90yd player interest radius earning a partial trickle, per the issue's design notes) is **not** implemented here. That's a natural, separately-scoped follow-up once this self-observation slice is proven. This PR is sim-only logic with no UI surface, so no screenshot section.

```mermaid
flowchart LR
  A["resolveCraft (crafting.ts)"] -->|rare-or-better roll| B["ctx.addItemInstance\nstamps signer + rolled quality"]
  B --> C["item instance carries\n{signer, rolled.quality}"]
  C --> D["useItem: potion drunk"]
  D --> E{"signer === drinker?"}
  E -->|no| F["no-op"]
  E -->|yes| G{"rarity >= rare?"}
  G -->|no| F
  G -->|yes| H["recipeForResultItem\nresolve craft"]
  H --> I["gainCraftSkill\n(same primitive ordinary craft xp uses)"]
```

### What changed
- `src/sim/professions/battlefield_xp.ts` (new): `battlefieldExperienceTrickle`, the self-observation handler. Gates on rare-or-better rarity first (cheap short-circuit), then requires `instance.signer === observerName`, resolves the producing recipe via a new reverse lookup, and only ever calls `gainCraftSkill` (never a drain primitive) so the "never reduces any craft's skill" requirement holds structurally, not just by assertion.
- `src/sim/content/recipes.ts`: added `recipeForResultItem(itemId)`, a reverse lookup from a crafted item id back to the recipe (and thus the craft/profession) that produced it.
- `src/sim/professions/crafting.ts`: `resolveCraft` now signs a single rare-or-better crafted output via `ctx.addItemInstance` (same signable-rarity threshold and `{signer}` shape #1145 already uses for monster materials), instead of always granting a plain fungible item. Below rare tier, or for `resultCount > 1` recipes, output stays fungible exactly as before.
- `src/sim/items.ts`: `useItem`'s potion-drink path resolves the specific instance before it's consumed and, if self-signed, calls `battlefieldExperienceTrickle`.

### Tests
- `tests/professions_battlefield_xp.test.ts` (new): self-observation grants the trickle at rare/epic/legendary; common/uncommon and unrolled/unsigned instances all grant zero; a non-signer observer grants zero (no party/bystander logic in this PR); an item with no matching recipe grants zero; the handler only ever mutates the one credited craft, never any other.
- Existing `tests/professions_crafting.test.ts` and `tests/items.test.ts` still pass unchanged (37 + 31 tests green), confirming the signed-output path doesn't regress common-tier crafting or plain potion use.
- `tests/architecture.test.ts` (sim purity/determinism guard) passes: no DOM imports, no `Math.random`/`Date.now`, this module draws no rng at all.

Built on top of #1207 (`feature/professions-signed-materials-1145`), since this needs both #1127 (crafting action) and #1165 (item-instance model/signer field), both already landed on that branch.


Nervescraper:
Reviewed the self-observation slice against the professions doc and the stacked branch. The core module looks great: sim-pure, draws no rng, additive-only by construction, rare+ gate first, and the negative paths are well tested. Three things before this merges:

**[High] Credit is decoupled from the consumed copy (repeatable trickle).** `useItem` finds the first self-signed instance (forward scan), but `ctx.removeItem` consumes from the last matching slot (backward scan), and `addItemInstance` pushes instance slots to the end. Repro: craft a signed rare potion early (early slot), later pick up plain potions of the same itemId (they stack later); each plain potion drunk consumes from the plain stack while the lookup still returns the untouched signed instance, granting the trickle once per potion cooldown off a copy that is never consumed. Suggest tying the credit to the actually-removed slot (have the removal return or target the consumed slot's instance).

**[Med] The active-archetype gate needs a named integration point.** The doc scopes Battlefield Experience to the active specialty only ("a potion drunk or meal eaten feeds nothing unless alchemy or cooking is your specialty"), which is the anti alt/breadth lever. That state lives in #1205's branch, so it can't be checked here, but the module comment lists only radius/party weighting as follow-up. Whichever of this PR / #1205 merges second needs to add the gate; worth a TODO in battlefield_xp.ts so it doesn't slip.

**[Med] Ceiling composition with #1203.** `gainCraftSkill` on this stack is uncapped; once #1203's tier ceiling lands, the trickle should respect it (xp above the ceiling is lost) or this becomes the only uncapped skill channel.

**[Low]** wheel.ts's header still promises "the wheel/mass-conservation mechanic (a later issue) will extend this file"; per the #107 decision that should read the attunement/empowerment ceiling instead, so the retired model doesn't re-seed from code comments.

Also noting the diff carries the whole #1207 stack (~6.4k lines, ~40 regenerated goldens); happy to re-review once the ancestors merge and this shrinks to its slice.

jgyy:
Thanks for the thorough review, @Nervescraper. Addressed in 972073cc:

- **[High] Repeatable trickle off an uncredited copy.** Fixed at the root: `removeItem` (`src/sim/sim.ts`) now returns the `ItemInstancePayload[]` it actually consumed (backward-scan order, matching removal), instead of the caller pre-looking-up a self-signed instance that may sit in a different slot than the one removed. `useItem`'s potion path (`src/sim/items.ts`) now credits from that return value directly, so a stale signed instance and a later unsigned stack of the same item can never diverge again. Added a regression test that reproduces your exact repro (signed rare added early, plain copies added later, drinking a plain copy grants zero and leaves the signed instance untouched) and confirmed it failed before the fix.
- **[Med] Active-archetype gate integration point.** Added a named `TODO(#1149/#1205)` in `battlefield_xp.ts` at the exact spot the gate needs to land, explaining which of this module / #1205 needs to add it and why (so it doesn't slip silently).
- **[Med] Ceiling composition with #1203.** This already composes for free: the trickle only ever calls the shared `gainCraftSkill` primitive (never a separate mutation path), so whatever cap #1203 eventually adds to that primitive applies here automatically. Left a note to that effect rather than a stub, since #1203 hasn't landed on this branch yet and its actual shape (tier multiplier, not a hard ceiling per the issue) may not need extra plumbing here at all.
- **[Low]** Fixed the stale `wheel.ts` header comment: it no longer promises a "wheel/mass-conservation mechanic," and instead references the #107 design-review decision (mass-conservation retired in favor of a swappable attunement/empowerment model).

All existing tests (professions_crafting, items, item_instance) plus the new regression test are green, `tsc --noEmit` is clean, and `tests/architecture.test.ts` still passes (no rng draws added). Ready for another look whenever convenient; agreed the diff will read a lot cleaner once #1207's ancestors merge.

PR #1064, opened 2026-06-30
feat(sim): ground-targeted casting + thematic spells (5 classes)

ryan-foo:
Adds a **ground-targeted cast primitive** and a thematic ground-aimed spell for each caster/ranged class, built on it.

## The primitive
An ability can declare `targetMode: 'position'` and be aimed at a world point. The client proposes the point; the **server clamps it to the ability's range** from the caster (authoritative); the cast's `aoeDamage`/`groundAoE` centers there, with a terrain-draped impact burst. No point chosen -> falls back to the caster's feet, so existing casts are unaffected. Wiring: `AbilityDef.targetMode`, `Entity.castAim`, `SimEvent.spellfxAt`, `castAbility(aim)`, `IWorldCombat.castAbilityAt` (Sim + ClientWorld), a `castAt` command, HUD routing, renderer burst.

## Spells (one per caster/ranged class, all fully translated across ~21 locales)
- **Mage - Flamestrike**: lingering fire zone
- **Warlock - Rain of Fire**: lingering fire zone
- **Hunter - Volley**: instant ranged burst at a point (scales off Ranged AP)
- **Druid - Hurricane**: lingering nature zone
- **Shaman - Earthquake**: lingering nature zone

(Mage **Blizzard** is the remaining planned one - it pairs with a chill ailment, tracked in the design doc.)

## Tests
`tests/ground_target_cast.test.ts`: the zone lands at the aim, clamps to range, falls back to the caster, and each per-class groundAoE spell drops its zone at the aimed point. Green locally: tsc, `biome ci --changed`, i18n coverage + completeness, world_api parity, guide, sim, and the casting/combat suites.

Design: `docs/design/arpg-spell-mechanics.md` (PR with the full mechanic catalog + roadmap).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Flamestrike redesign (owner review)
Flamestrike is now an instant aimed BURST, not a lingering ground zone: aoeDamage 34-44 in a 7yd radius at the range-clamped point, plus a new pooled render visual (src/render/aoe_ring.ts) that flashes a school-colored AoE ring on the terrain under the blast so the impact area reads, then gets out of the way. The spellfxAt event carries the blast radius. Descriptions reworded in English and all 17 locales; tests assert the burst lands at the aim, clamps to range, and leaves no zone.

## Cursor aiming reticle (SHIPPED)
[...]

Issue #96, opened 2026-06-13
Prevent player damage and hostile CC outside accepted duels/PvP


maxpolaczuk:
## Reported from Discord
- `#bug-reports`, 2026-06-13T10:48:18Z: player in initial village was killed by another player/mage and asked whether PvP was enabled or hacked.
- `#bug-reports`, 2026-06-13T11:09:02Z: reporter did not accept a duel.
- `#bug-reports`, 2026-06-13T11:09:36Z: same person used a polymorph-looking spell on multiple people.

## Problem
Players appear able to damage or crowd-control other players outside an accepted duel/PvP context.

## Expected
- Players should not be able to damage, polymorph, stun, or otherwise hostile-CC other players unless a valid duel/PvP state allows it.
- Add regression tests for player-vs-player hostile spell rejection outside duels.

## Coverage check
I did not find a PR that explicitly fixes non-duel player damage/CC. PR #57 mentions cleansing hostile NPC control auras, but this Discord report is player-vs-player in the starter village.

Rubsey:
Unsure if this is happening, need reports of when it happens and how it happens

PR #1305, opened 2026-07-02
feat: Talents 2.0 foundation: five casting primitives + ground-targeted spells (folds #1064)

ryan-foo:
Foundation PR of the Talents 2.0 plan (the committed PRD, docs/prd/talents-2.0.md, is the epic's spec). Per owner decision this PR now carries three folded slices: the five casting primitives, the entire former #1064 (ground-targeted casting + six spells + aiming reticle + impact ring), and the conditional primitives with their live demos. The ground-targeting half is player-visible; everything else is content-unused machinery that later PRs (signatures, choice rows) build on.

## 1. Ground-targeted casting + six spells (player-visible, was #1064)

An ability can declare targetMode 'position' and be aimed at a world point. The client proposes; the SERVER clamps the point to the ability's range from the caster (authoritative); effects center there.

**The aiming reticle**: pressing a ground-targeted ability enters aiming mode: a school-colored pulsing ring follows the cursor on the terrain, pre-clamped to the ability's range with the sim's own clamp math (dims when the raw cursor is out of range) and sized to the blast radius. Left-click or the same keybind commits; Escape/right-click cancels. Mobile and a groundReticle setting (label translated in 17 locales) keep quick-cast. Pure aim logic unit-tested in src/ui/ground_aim.ts.

aiming reticle

**Flamestrike (redesigned per owner review)**: an instant aimed BURST, not a lingering zone: aoeDamage 34-44 (+SP) in a 7yd radius at the clamped point, plus a new pooled AoE impact ring visual (src/render/aoe_ring.ts) that flashes the blast area on the terrain and gets out of the way. Tooltip shows live resolved scaling:

flamestrike tooltip

flamestrike burst

**Four more ground spells**, each captured mid-effect by asserted scripts: Rain of Fire (warlock channel), Volley (hunter channel), Hurricane (druid channel), Earthquake (shaman lingering zone). Channel pulses and zone starts carry the ring radius so every ground spell gets the impact ring.

volley

earthquake

## 2. The five casting primitives (content-unused, goldens untouched)

- **interrupt**: cancels a target's non-physical, non-fishing cast and applies the existing school-lockout aura. HARDENED per review: PvP consent gate (interrupting a non-consenting player outside duel/arena/fiesta is a full no-op, via the shared hostility predicate; cases added to pvp_safety.test.ts), diminishing returns on the lockout family (new 'lockout' DR category, full/half/quarter/immune, 18s reset; on immunity the cast still cancels but no lockout lands), and an explicit uninterruptible AbilityDef flag.
- **empower-next**: one-charge self-auras: next_cast_instant (not channels, not physical), next_cast_free (consumed exactly where the cost is billed: queue/cast-start/completion/early-return branches), next_attack_crit (forces the crit chance to 1 while consuming the IDENTICAL rng draw, so draw order is byte-identical).
- **cast-while-moving**: AbilityDef flag + per-ability talent transform; the move-input cast cancel skips mobile casts.

interrupt lockout

scorch while moving

## 3. Conditional primitives + live demos (was the PR2 slice)

- **vs-rooted conditionals**: directDamage.vsRootedMult (damage multiplier against rooted targets, checked at projectile impact) and GlobalModEffect.critVsRooted (additive spell crit against rooted targets, composing with fiesta augments).
- **addEffects**: talents append effect riders to abilities, shallow-copied before scaling; shared content arrays never mutated.

Asserted demos (a red assertion cannot produce the shot): a vsRootedMult 3 nuke hits a rooted wolf for EXACTLY 3x (30 -> 90, crit pinned); critVsRooted 1 turns the same fixed nuke into a guaranteed 1.5x crit only while rooted (30 -> 45); an addEffects rider puts a live dot on Fireball.

vs rooted 3x

## Determinism

No new rng draws on any existing path; the guaranteed crit consumes the same draw with chance 1; ground-target content adds draws only on its own new casts. All 159 parity tests pass with goldens unmodified.

## Tests

405 tests green on the merged head: 16 primitive cases (interrupt/lockout/DR/uninterruptible/consent, one-charge consumption incl. the cast-time completion bill and the on-swing queue lifecycle, rng draw-count invariance, movement-cancel baseline + def-flag + talent-mod mobility), ground_target_cast + ground_aim suites, pvp_safety incl. the new interrupt cases, world_api_parity/command_schema ratchets consciously bumped for castAbilityAt, architecture, and parity. Every screenshot above is produced by a committed script that asserts its scene before shooting (scripts/talent_primitives_shots.mjs, pr2_flamestrike_showcase.mjs, pr2_ground_spells_shots.mjs, pr2_primitives_shots.mjs).

Review history: the coverage pass found and fixed a free-charge leak in the conjure/revive branches and interrupt school resolution failing for non-player casters; the balance review's must-fixes (consent gate, lockout DR, uninterruptible flag) are in; accepted-semantics notes live in the PRD.


FernandoX7 (commented):
Solid PR, and the determinism story holds up under checking. `rng.chance` is `next() < p`, so forcing the crit with `chance(1)` draws the exact same value and just returns true, draw order unchanged. I ran the guards on the head: `tests/architecture.test.ts` green, the 159 parity goldens pass unmodified, and `tests/talent_primitives.test.ts` + `tests/ability_tooltip_talents.test.ts` pass.

What I liked: the free-charge billing is placed at each real cost site rather than one blanket spot (queue time for on-swing via `queuedOnSwingFree`, cast start for channels and instants, completion for cast-time spells through `applyAbility`, and the conjure/revive early returns), and every runtime reset of `queuedOnSwing` also clears `queuedOnSwingFree` (roster, arena, fiesta, death, swing), so no stale flag survives a respawn. Content-unused checks out too: nothing in `src/sim/content/` grants the new effects, so nothing player-visible moves. The `empower_next.ts` split as a thin consume-helper and the interrupt's per-player-then-global (`ABILITIES[...]`) def fallback so future mob casters are interruptible are both nice.

A few small things, all non-blocking:

- nit (confirm intent): `casting_lifecycle.ts:364` consumes `next_cast_free` at queue time for an on-next-swing ability, but toggling that same ability back off before the swing (`casting_lifecycle.ts:366` / `auto_attack.ts:139` delete the flag) burns the charge for nothing. No live impact today since no content grants it, and it matches the documented "consumed at queue time" model, but worth confirming that is the feel you want before a later PR wires `next_cast_free` onto an on-swing ability.
- nit: `queuedOnSwingFree` is correctly server-local and off the wire (it joins the other optional transients absent from `blankEntity`), and the three new aura kinds ride the existing string `kind` field so they round-trip once content grants them. No change needed, just noting I checked the parity side.
- nit: the description says "38 unit tests in tests/talent_primitives.test.ts"; the file has 14 `it()` cases. Coverage across the three primitives is fine, only the count is off.

FernandoX7:
Following up on the balance question, since the PRD for the whole epic is committed here. A few things to carry forward into PR3/PR5 so overall balance stays in view as the content lands. None of this blocks PR1: the primitives are content-unused and the goldens are untouched, so nothing shifts until an ability or signature actually grants them. These are for when it does.

Worth grounding the whole discussion in one fact: player interrupts and cast-while-moving do not exist in the game today (the lockout aura is mob-only, and every cast cancels on movement). So the row content introduces two genuinely new power axes, which is why they are worth watching.

Two I would treat as must-fix in the PR that first wires a player interrupt, not just tuning:

1. Gate hostile interrupts to consented PvP. A player interrupt applies a lockout aura, which is hostile control, and the #96 pvp_safety invariant forbids that between players outside a duel/arena/fiesta. `pvp_safety.test.ts` currently covers poly, auto-attack, frost-nova, fear, and stun, but not the interrupt to lockout path. Route lockout application through the same consent check and add the interrupt case to that test. The parity gate will not catch this (it detects drift, not "too strong").

2. Wire diminishing returns onto the lockout/silence family. DR (`diminishedCrowdControlDuration`) covers only stun/fear/poly today; the interrupt branch makes no DR call. With every class getting an interrupt option and no DR, two staggered interrupters on roughly 10 to 12s cooldowns can keep a healer's primary school locked for most of every window in 2v2. That is a machinery gap, not a number, so it needs building before the content can be tuned around it.

Watch-items for the PR5 tuning pass (numbers can handle these, just keep them in view):

- Caster-vs-melee tilt. Interrupts only bite non-physical casts, and interrupt abilities are physical (so uninterruptible themselves), which favors melee into casters. Cast-while-moving (Firestarter, Netherwind) is the intended offset, so tune the two axes together rather than in isolation.
- Priest Silence is a full 4s all-school silence (not a single-school lockout) on a 30s cd; with no DR that is a near-guaranteed kill window in 2v2. Highest-impact single option in the draft.
- Presence of Mind's value is removing Pyroblast's 6s cast window, not raising its number. The cross-class rule already keeps guaranteed-crit (rogue) and instant-nuke (mage) on different classes, so the levers are the 60s cd and the opportunity cost against the row's passive.
- Mage kite. The game ships cheap no-cooldown snares (Frostbolt 40%, Hamstring) and zero snare/root breaks in any kit, so a mobile Scorch compounds hard, and specifically on the mage.

One suggestion: the flat one-of-three-per-row structure makes a full 3^6 = 729-build-per-class enumeration cheap for the headless sim, and there is no cross-class balance test today (the parity gate is a drift detector, not a balance oracle). A build sweep in PR5 would give you an actual balance check at exactly the point where power-per-option is highest.

Last one: consider an explicit uninterruptible flag before interrupt content ships. Right now "cannot interrupt this" is implicit (a cast is safe only if it is physical-school or a scripted string absent from `ABILITIES`, which is how Nythraxis's wipe cast stays safe). A future boss cast that uses a real ability id would become interruptible by accident.

ryan-foo:
Both must-fixes plus the uninterruptible flag are now IN this PR (commit 19b525f7), per the review:

- Consent gate: the interrupt case routes through the same shared hostility predicate the other hostile CC uses (isHostileTo, covering accepted duels and active arena/fiesta). Outside consent it is a full no-op (no cancel, no lockout). pvp_safety.test.ts gains the interrupt-to-lockout cases: outside a duel the victim's cast survives; inside an accepted duel the interrupt works.
- DR: lockout durations route through diminishedCrowdControlDuration under a new 'lockout' category on the standard full/half/quarter/immune ladder with the existing 18s reset window. On immunity the cast still cancels but no lockout lands, so staggered interrupters cannot chain-lock a healer's school.
- uninterruptible flag added to AbilityDef and honored by the interrupt case; no content flags needed today (the Nythraxis casts are scripted ids absent from ABILITIES), it is pure future-proofing as suggested.

On the nits: the on-swing queue-toggle burning the free charge is intended and documented in the PRD ('consumed at queue time, no refund'), will re-confirm the feel when content first grants next_cast_free on an on-swing ability; test-count in the description corrected (16 cases now, after the hardening tests); thanks for checking the parity/wire side of queuedOnSwingFree.

The PR5 watch-items (caster/melee tilt tuned together with cast-while-moving, priest Silence, PoM's cd lever, mage kite) and the 729-build headless sweep are recorded as binding items in the committed PRD's interrupt-review section.

ryan-foo:
The primitive is just there, the primitive (cast while moving) is in place all for setting up for the 5th PR, which introduces the full talent 2.0 system with new meaningful choices -- spells; (ability to slow, interrupt, root, etc).

FernandoX7 (commented):
Re-reviewed. The three must-fixes from last round all landed and check out against the code, not just the description:

- Interrupt consent gate: `effect_dispatch.ts:267` makes a player-on-player interrupt a full no-op unless `isHostileTo` holds, the same predicate the other hostile CC and auto-attack use, so it inherits the #96 model.
- Lockout DR: `effect_dispatch.ts:278` routes through `diminishedCrowdControlDuration(..., 'lockout', ...)` on the full/half/quarter/immune ladder (`sim.ts:3192`); on immunity the cast still cancels but no lockout lands.
- Uninterruptible flag: honored at `effect_dispatch.ts:275`.

`pvp_safety` and `talent_primitives` cover all three, and determinism/parity are clean (single rng draw preserved, primitives content-unused so the goldens are untouched).

Two should-fix items turned up while re-reading the folded ground-targeting slice. Neither blocks the primitives.

- `server/game.ts:3405`: the new `spellfxAt` impact event is broadcast to every connected player instead of interest-scoped. `eventAnchor` only anchors on `targetId`/`entityId`, and `spellfxAt` carries just `x`/`z`, so it returns null and `routeEvents` sends it to everyone. The sibling `spellfx` anchors on `targetId` and scopes to `EVENT_RADIUS`, so every ground-spell impact (Flamestrike, Rain of Fire, Volley, Hurricane, Earthquake) fans out server-wide. Cosmetic, but a real scoping drift. Fix: have `eventAnchor` return `{x, y: 0, z}` for events that carry world coordinates.
- `casting_lifecycle.ts:129-133`: the channel-completion branch clears `castingAbility`/`channeling` but not `castAim`, so a completed ground-targeted channel (rain_of_fire / volley / hurricane) leaves a stale aim. The other resolve paths clear it (`:150`, `:158`). It is unobservable today (the next cast overwrites it and no delayed effect reads it), but it breaks the "castAim is always cleared on resolve" invariant this foundation is meant to hold for later content. One line, mirroring `:150`, plus a regression test.

The earlier note about a toggled-off free on-swing ability not refunding its charge is settled: consume-at-queue-time by design, documented in the PRD.

ryan-foo:
Both should-fixes are in (latest commit on the branch):

- eventAnchor anchors world-coordinate events at their own x/z, so spellfxAt impacts now interest-scope to EVENT_RADIUS exactly like the entity-anchored spellfx sibling; the broadcast fallback stays for chat/log events that genuinely carry no location.
- The channel-completion branch clears castAim like the other two resolve paths, restoring the always-cleared-on-resolve invariant, with a regression test that runs a full aimed Rain of Fire channel and asserts the aim is dropped at completion.

Thanks for confirming the consume-at-queue-time semantics against the PRD.

FernandoX7:
Pushed three commits (on top of a base merge) to bring this up to the new base and make the release gate green.

- Merged release/v0.19.0 into the branch, since the PR was retargeted to it. Clean merge apart from the generated i18n artifacts, which I regenerated.
- Icons: the five ground spells (flamestrike, rain_of_fire, volley, hurricane, earthquake) were falling through to the procedural icon fallback and colliding (flamestrike matched rain_of_fire, hurricane matched earthquake). Each now has an explicit, distinct recipe, so ability_icons passes.
- Translations: v0.19.0 introduced 53 keys that were still pending (the daily-rewards wallet feature, the desktop crash/update dialogs, and the native-update prompt), 772 rows across all locales. Filled every supported locale, so the release tier is now at 0 pending. Placeholders are preserved and brand/ticker terms (World of ClaudeCraft, WOC, SOL, USD, Solana) stay verbatim; the dialect locales inherit their base.
- Biome: format-only cleanup of the touched files plus one import sort, no logic changes.

Full release-tier run is green locally: 6574 tests pass, tsc is clean, env/server/client all build, the security gate passes, biome ci --changed reports no errors, and the i18n artifacts regenerate with no drift.

FernandoX7 (approved):
Went ahead and did translations on here too to prepare for release

ryan-foo:
Cheers @FernandoX7 thank you! Sets the foundation well for next wave of PRs

PR #1311, opened 2026-07-02
fix(gamepad): include controller button presses in APM

madmatah:
## Summary

Controller button presses were missing from the APM value displayed in the HUD because controller input is polled instead of producing keyboard or pointer events.

This change records each controller button rising edge once, including buttons handled directly by the controller manager and buttons used while navigating HUD windows. Held buttons do not produce repeated APM entries.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run check:types && npm test && npm run ci:changed && npm run build && npm run build:server`

- Manual steps: Tested on desktop with an xbox controller

## Screenshots / recordings

image

---

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

      the client boundary in this same change.

### Hygiene

### DIFF
--- a/src/game/gamepad.ts
+++ b/src/game/gamepad.ts
@@ -4,20 +4,23 @@
 // Input instance (movement / camera / jump), dispatching edge-button actions via
 // the host's onAction callback, the virtual-cursor UI-navigation mode, and
 // haptic rumble. Modeled structurally on MobileControls.
-import type { Input } from './input';
+
 import type { GamepadBindings } from './gamepad_bindings';
 import {
-  STANDARD_BUTTON_COUNT,
   AXIS,
-  GP,
-  TRIGGER_THRESHOLD,
   GAMEPAD_NONE,
-  stickToMoveFlags,
-  stickToLook,
+  GP,
   risingEdges,
+  STANDARD_BUTTON_COUNT,
+  stickToLook,
+  stickToMoveFlags,
+  TRIGGER_THRESHOLD,
 } from './gamepad_map';
+import type { Input } from './input';
 
 export interface GamepadCallbacks {
+  // Record one physical button rising edge for the HUD's APM readout.
+  onInputEdge(): void;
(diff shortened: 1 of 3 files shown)


FernandoX7 (commented):
Clean, well-targeted fix, and the root cause is right: gamepad input is polled, so it never emits the keydown/pointerdown events the APM meter listens for. Wiring `onInputEdge` into both rising-edge loops (the main dispatch path and the UI-cursor path) covers every physical press, and routing it through the same `InputActivityMeter.record` keeps gamepad edges on the same 60s window and clock as keyboard and pointer input.

I checked the two double-count paths this could have opened, and both are clean:
- Cursor-mode `clickAtCursor()` synthesizes `mousedown`/`mouseup`/`click`, but the tracker only listens for `keydown`/`pointerdown`/`wheel` and drops `!e.isTrusted` events, so the synthetic clicks are not recounted (`src/game/input_activity.ts`).
- `poll()` returns early in pointer mode after `updateCursor()`, so the main edge loop and the cursor edge loop never both run in the same frame (`src/game/gamepad.ts:145`).

The `main.ts` relocation is safe: `inputMeter` is now declared ahead of all three of its uses (heartbeat interval, `onInputEdge`, `getApm`) and nothing referenced it in between. Making `onInputEdge` required on `GamepadCallbacks` is the right call (a forgotten wire fails to compile), and both construction sites are updated.

Verified at the PR head: `tests/gamepad.test.ts` passes, `tsc --noEmit` is clean, and biome reports no format diff on the three files. No blocking or should-fix items.

Nit: `src/game/gamepad.ts` is fully biome-reformatted in this diff (brace-expanded setters, import reorder), which is per policy for touched files but does bury the three-line behavior change under the churn.

FernandoX7 (approved):

PR #1306, opened 2026-07-02
feat(editor): in-game 3D map editor with realtime sculpting, biomes, playtest, and server-side maps (WIP)

maxpolaczuk:
Part of the v0.19.0 release (#1264). WIP: opened as a draft so @EnriqueGF can collaborate.

## What this is

A production, in-game 3D map editor and modding platform at `/editor` (a new build entry, framework-free like the guide). Players sculpt, paint, and populate custom maps in the real engine, playtest them instantly, and save/fork/share them server-side.

## Feature summary

- **Realtime terrain editing**: raise/lower/smooth/flatten brushes with live chunk-local re-mesh, a shader brush ring on both graphics tiers, adjustable size/strength/falloff, and undo/redo (200-entry command stack).
- **Biome paint**: seven biomes (vale, marsh, peaks, plus new paint-only beach, desert, volcano, cave) that reshape terrain and recolor ground, foliage, fog, sky, and ambience.
- **Water level**: per-map water height threaded through every sim and render consumer (swim, pathfinding, spawns, shore foam, minimap), edited live with a slider.
- **Asset placement**: 863-model catalog including 116 new curated CC0 biome models (Quaternius Pirate Kit, Kenney Survival/Watercraft/Dungeon kits, KayKit Medieval Hexagon, loafbrr Canyon and Mines sets, animated fish), with optional collision footprints that feed the sim colliders from the same record the renderer draws.
- **Camps and spawn**: place mob camps (mob/count/radius) and the playtest spawn point.
- **Playtest**: one button drops you into the offline engine on the edited map; geometry is identical because sim and renderer sample the same edited heightfield.
- **Save / fork / publish**: Postgres JSONB map documents with optimistic versioning, fork lineage, publish/unpublish, a public browse tab, and per-account caps. One shared sanitizer (`src/sim/map_doc.ts`) validates documents on both client and server.
- **GLB uploads**: content-addressed (sha256) user asset uploads with a hand-rolled structural validator (chunk walk, no external URIs, accessor bounds), rate limits, immutable serving, and admin moderation.
- **i18n**: a 254-key `editor` catalog domain; the five non-Latin locales are fully hand-filled (M16).

## Architecture notes for review

- The sim gains a swappable `WorldContent` registry (`src/sim/data.ts`); the BUILT-IN world is byte-identical through the indirection, pinned by `tests/custom_map_parity.test.ts`. The authoritative server and headless RL env never see custom content.
[...]

PR #1312, opened 2026-07-02
fix(editor): audited bug, perf, and UX corrections for the map editor

EnriqueGF:
Batch of corrections for the map editor branch, from a four-lane audit (correctness, performance, usability, server sync) with every finding verified against the code. Targets the PR branch so it can be merged into #1306 directly.

## Rendering
- Painted biomes now read on the splat terrain tier: the biome override re-bases the splat texture mix (was: grass texture everywhere, only a 35 percent vertex tint changed) and the live grass ring tints by the painted biome.

## Data loss (P0)
- Edits made during an in-flight save are no longer marked saved with their draft destroyed (edit-generation counter).
- beforeunload guard on a dirty document; playtest saves a draft first.
- Opening or importing over unsaved work asks for confirmation; drafts are per-map slots (saving map B no longer destroys map A's only backup).
- Fork mints a fresh meta.id server-side; previously a fork silently repointed the original map's local link and later saves cross-wrote the wrong server row.
- sanitizeMapDoc hardened: camps/npcs/objects/zones/lakes/pois rebuilt field by field with finite checks and caps, so a hostile published map can no longer hang a viewer's tab (camp count 1e9, giant z-bands, thousands of lakes) or store an unloadable Infinity-to-null byte.

## Bugs
- Placement rotation/scale slider edits are now undoable (prev captured before the live drag).
- Document index vs render slot stay aligned when an asset id fails to resolve (selection no longer highlights the wrong mesh).
- 3D viewport start/reload race fixed with a generation token (no more double Sims and double-committed strokes).
- Collider invalidation centralized over every terrainEdits mutation (erase, paste, hills, undo/redo included).
- Paste/procgen/scatter clamp to the document caps with a warning toast instead of silently truncating on the next load.
- 2D marker drags and resets push undo entries; right-drag pan no longer opens the context menu; stale tabs get the 409 save-as-copy flow instead of silently winning; net calls get a 20s timeout so Save cannot get stuck.

## Performance
- Terrain edits get a spatial bucket index (bit-identical heights, proven by test against the linear scan) so the brush no longer degrades over a session.
- Stroke end re-seats only in-region placements; deleting a placement does Map surgery instead of re-cloning every GLB; the hidden 3D viewport pauses instead of rendering full frames while editing in 2D; 2D stamps blit pre-rendered sprites; asset browser caches thumbnails and debounces search; stroke undo uses contiguous spans; autosave writes compact JSON to per-map localStorage keys.

## Tests and gates
Around 90 new tests in 10 files, each audited failure pinned as a regression (sanitizer hostile input, terrain index determinism, save lifecycle, draft slots, link versioning, caps, spans, placement alignment, net timeout). Full run green: tsc, parity (159), architecture (24), i18n completeness incl. the M16 non-Latin fills for the 9 new editor keys. Reviewed by the repo's qa-checklist, architecture-reviewer, migration-safety, and privacy-security-review agents (their two findings, a float guard band on the index and zone clamps, are included).


EnriqueGF:
Pushed a second batch on top of the fixes, new editor features:

- Real 3D preview thumbnails in the asset browser (lazy, cached, streams in without blocking search)
- Direct manipulation in Select mode: drag placements along the terrain, Shift+wheel rotate, Alt+wheel scale, arrow-key nudge, Ctrl+D duplicate, all undoable
- First-run tutorial (7-step tour, replayable) plus a Help modal with tools, shortcuts, and the save/playtest flow
- Free editor camera: the frozen character is hidden and Q/E fly vertically for real
- Collision authoring: per-placement collision radius with live footprint rings, a Blocker tool (K) for invisible walls players cannot walk or jump through in playtest, and footprints auto-show while authoring collision

All architecture-reviewed (blockers ride an optional WorldContent field, built-in world byte-identical, parity green), ~120 new tests, full i18n with the five non-Latin fills.

EnriqueGF:
Third batch pushed, editor polish:

- Loading overlay while the 3D world boots (spinner plus "Loading the world...", also on map reload; the stage no longer sits black for seconds)
- Free editor camera: the frozen character is gone from view and Q/E fly vertically; navigation no longer reads as dragging a warrior around
- Per-tool color system: one hue follows the active tool through the rail button, the inspector header chip, and the slider accents (earth for sculpt, moss for paint, azure for water, violet for content, red for blocker, coral for erase); the idle rail stays monochrome
- Tool rail grouped by family with separators and clearer shortcut badges; topbar actions grouped (document | save | exchange | help) and the Save button turns amber while there are unsaved edits

All decoration plus three tiny TS hooks, no new strings, verified in-browser with zero console errors.

EnriqueGF:
Added an opt-in Autosave toggle in the topbar (default off, status-LED style next to Save, preference persisted). While on, it runs a real save every 30s when there are unsaved changes, strictly gated: never over an in-flight save, never mid-stroke or mid-drag. Automatic saves are silent on success, never open a dialog, and any failure (quota, network, version conflict) turns the toggle off with one toast explaining why, so a broken save path cannot loop. Verified end to end in the browser.

PR #1313, opened 2026-07-02
feat(chat): whisper tab, portrait whisper action, and a dev whisper bot

Blaine1705:
## What

Rounds out whispers across the classic HUD.

- **Whisper tab.** The chat "+" menu now opens a filter-only "Whisper" tab that
  collects every whisper (sent and received) in one place, away from the busy
  All view. Typing in that tab replies to the last whisperer (/r); an explicit
  slash command still wins.
- **Portrait / target whisper action.** The player right-click menu (the
  portrait/target-frame one) gains a "Whisper" action alongside view profile /
  invite / trade, using the shared startWhisper flow. Shown for other players
  only, not yourself.
- **Dev whisper bot.** A new dev-only "/dev bot <name>" spawns a stationary,
  whisperable test player that auto-replies, so whispers can be exercised
  offline. The recipient copy is skipped for a bot so the single local client
  does not render a duplicate of the sender's own line.

## Notes

- Reuses the existing translated "Whisper" label, so no new i18n keys are added.
- The pure chat-tab model (chat_channels.ts) grows the whisper tab,
  composeWhisperReply, and widened persistence, all unit-tested. SimContext
  gains spawnDevBot (append-only) behind the seam; sim purity, the S3 drift
  guard, and the parity gate stay green.
- Includes one small pre-existing fix cherry-picked from the target-frame branch
  (c1ae201d): world_api/chat.ts becomes a real type-only sim import, and the
  architecture guard's allowlist lookup is made cross-platform (it silently
  never matched on Windows). This keeps the guard test green on all platforms.

## Testing

- New/updated Vitest: chat.test.ts (dev-bot spawn, whisper auto-reply, no
  duplicate line), chat_channels.test.ts (whisper-tab model + composeWhisperReply),
  sim_context / entity_roster mocks.
- Guard tests green: architecture.test.ts, localization_fixes.test.ts, and the
  parity gate.

### DIFF
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -617,6 +617,10 @@ export interface PlayerMeta {
   characterId?: number;
   cls: PlayerClass;
   name: string;
+  // Dev-only test dummy spawned via "/dev bot <name>" (social/chat.ts, gated by
+  // devCommands): a stationary player you can target and whisper to exercise social
+  // features offline; a whisper to it auto-replies. Runtime-only, never serialized.
+  isDevBot?: boolean;
   skin: number; // appearance index into the render SKINS[player_<cls>]; persisted, synced
   skinCatalog: SkinCatalog;
   // Cosmetic skin-select event: the rank rolled when the event token was used,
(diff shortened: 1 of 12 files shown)


Blaine1705:
https://github.com/user-attachments/assets/ce08e48e-4a59-4618-9cb2-5899d2b9b1c7

4ab4660c94eb6a7bd7b259ee9b7eb8e1

FernandoX7 (commented):
Tightly scoped and well tested. The pure model in `chat_channels.ts` carries the whisper tab, `composeWhisperReply`, and the widened persistence with real unit coverage; the new `SimContext.spawnDevBot` callback is mirrored in both the `sim_context` and `entity_roster` mocks so the completeness gate stays honest; and you cleaned up two em dashes and moved the check-mark to a char code on the way through. Guard tests are green locally at the head ref: architecture, localization_fixes (S3), chat, chat_channels, sim_context, entity_roster.

The classic chat model holds up:
- A filter-only tab that gathers every whisper (sent and received) in one place is exactly a classic Whispers chat tab, and keeping it out of `CHAT_TAB_CHANNELS` (no standing send channel) is the right model.
- `composeWhisperReply` to `/r`, with `lastWhisperFrom` recorded on the recipient when you whisper and updated on you when they reply, matches classic `/reply` semantics.
- Right-click Whisper on the portrait/target menu sits correctly next to inspect/invite/trade/duel and is hidden for yourself (`pid !== playerId`, and `startWhisper` self-guards on the name too).

The bot recipient-copy skip is correct: offline the single client paints every event regardless of pid, so the bot's copy would double the sender's own echo; online a bot has no client to deliver to anyway. The test pins it (every message addressed to the sender, exactly echo plus reply).

Two nits, nothing blocking:

- nit `src/sim/social/chat.ts:860`: the `/dev` help fallback still lists only level/tp/give/quest/quests, not the new `/dev bot`. Worth adding a line so it is discoverable.
- nit `src/world_api/chat.ts:22`: `OVERHEAD_EMOTE_ID_SET` is now derived from the local `OVERHEAD_EMOTES` instead of value-importing `OVERHEAD_EMOTE_IDS`, which is the right call for the seam. The two lists are identical today (13 ids, same order), so no behavior change, but `satisfies` only checks each entry is a valid id, not that the array is exhaustive, and nothing tests the two against each other. A future emote added to `sim/types` but not here would silently fall out of `isOverheadEmoteId`. A one-line exhaustiveness assertion (length or set-equality against `OVERHEAD_EMOTE_IDS`) would close that gap.

Minor design note, not a finding: the whisper-tab placeholder reads a generic "Whisper" rather than the specific last-whisperer, so it diverges from a per-conversation "To <name>" style. That is fine for a dynamic collector tab; flagging only in case you want the name later.

FernandoX7 (approved):

PR #1314, opened 2026-07-02
feat(ui): auto-close quest dialog and fade NPC voice when you walk away

Blaine1705:
## What

Two small polish touches for leaving an NPC mid-conversation.

- **Quest dialog auto-close.** The quest/gossip dialog now closes once you step
  out of talking range (dist > 8 from the NPC), matching how the vendor and loot
  windows already auto-close, instead of lingering until you click the X. The
  check lives in the per-frame update() right next to the existing vendor/loot
  ones (an earlier attempt sat in a language-change-only path and never fired).
- **Distance-faded NPC voice.** An NPC's voice line now fades by distance while
  it plays: full volume within about 6 yards, ramping to silence by about 22, so
  a dialogue you walk away from trails off naturally. It runs per-frame for a
  smooth ramp and is independent of the dialog window, so the voice keeps fading
  after the window closes until the clip ends. Overheard yells keep their own
  gain and are excluded from the fade.

## Notes

- Adds the pure `voiceDistanceGain()` curve (unit-tested in tests/voice.test.ts)
  plus a distance-gain layer on the voice player that leaves the settings volume
  and per-line gain intact. Purely cosmetic (voice-over is an enhancement; the
  dialogue text stays the source of truth), so it is gameplay-neutral.
- Includes one small pre-existing fix cherry-picked from the target-frame branch
  (c1ae201d): world_api/chat.ts becomes a real type-only sim import, and the
  architecture guard's allowlist lookup is made cross-platform (it silently never
  matched on Windows). Keeps the guard test green on all platforms.

## Testing

- New Vitest: voice.test.ts (voiceDistanceGain curve, clamps, monotonicity, and
  the never-dip-on-open behavior).
- Guard tests green: architecture.test.ts, localization_fixes.test.ts.

### DIFF
--- a/src/game/voice.ts
+++ b/src/game/voice.ts
@@ -13,15 +13,51 @@ import { VOICE_LINES } from './voice_manifest.generated';
 // the SFX/ambience mix.
 const VOICE_BASE_GAIN = 0.9;
 
+// Distance-attenuation curve for a talking NPC (world yards). At or within
+// VOICE_FULL_DIST the line plays at full volume; past it the volume ramps down
+// linearly to silence at VOICE_SILENT_DIST, so a dialogue you walk away from
+// trails off as if heard from farther away. You open a dialog within
+// INTERACT_RANGE (5), so FULL sits just past it and never dips the line on open.
+export const VOICE_FULL_DIST = 6;
+export const VOICE_SILENT_DIST = 22;
+
+export function voiceDistanceGain(
+  dist: number,
+  full = VOICE_FULL_DIST,
+  silent = VOICE_SILENT_DIST,
+): number {
+  if (!(dist > full)) return 1; // also catches NaN / negative: play at full
+  if (dist >= silent) return 0;
+  return (silent - dist) / (silent - full);
+}
+
 class GameVoice {
   private el: HTMLAudioElement | null = null;
   private vol = 0.9; // 0..1, from the settings slider
   private enabled = true;
+  private playGain = 1; // per-line gain passed to play()
+  private distanceGain = 1; // live distance attenuation (1 = at the NPC)
+
+  private applyVolume(): void {
+    if (this.el)
+      this.el.volume = Math.min(1, this.vol * VOICE_BASE_GAIN * this.playGain * this.distanceGain);
+  }
 
   /** Set voice volume (0..1). Safe any time. */
   setVolume(v: number): void {
     this.vol = Math.min(1, Math.max(0, v));
-    if (this.el) this.el.volume = this.vol * VOICE_BASE_GAIN;
+    this.applyVolume();
+  }
+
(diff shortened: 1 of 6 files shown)


Blaine1705:
https://github.com/user-attachments/assets/8d8bea13-c651-4e29-8095-2c6135f296b1

FernandoX7 (commented):
Tightly scoped and well tested. The pure `voiceDistanceGain` curve is extracted and unit tested (NaN/negative handled by `!(dist > full)`, monotonic, clamped), the auto-close reuses the existing vendor/loot pattern at the same 8yd threshold in the same medium band, and it stays gameplay neutral: the voice is an enhancement with the dialogue text still authoritative, and the fade is universal rather than tier gated. tsc, voice.test.ts, and architecture.test.ts are green at the head, and it merges cleanly into release/v0.19.0.

The `voiceNpcId` lifecycle is the part I looked hardest at and it holds up. It is set on gossip/quest open, self clears once `voice.isPlaying()` goes false, and is reset to null on an overheard yell so a yell never gets attenuated by a talked-to NPC's position. Keeping it independent of the dialog window (which closes at 8) is the right call, so a line you walk away from trails off after the window is gone. I also checked the escape hatches (a chat-linked quest view, opening a vendor from gossip, manual close) and none of them leaves a stale id that would attenuate a fresh line.

Behaviorally this lands where you would want it: an interaction window that drops once you leave the NPC, and a voiced line that attenuates with distance (full within about 6, silent by about 22) while a long-range overheard yell keeps its own gain.

One nit:

- nit `src/world_api/chat.ts:22`: deriving `OVERHEAD_EMOTE_ID_SET` from the local `OVERHEAD_EMOTES` is the right way to keep the seam import type only, and the id set matches sim/types' `OVERHEAD_EMOTE_IDS` exactly today, so `isOverheadEmoteId` is unchanged. The catch is that `satisfies readonly { id: OverheadEmoteId }[]` only proves each local id is valid, not that the array is complete, so a future emote added to `OVERHEAD_EMOTE_IDS` alone would silently fall out of the guard with no compiler or test failure (the old direct import could not drift). A one-line exhaustiveness assertion, or a test asserting the two id sets are equal, would lock it. Not blocking.

Everything else looks good.

FernandoX7:
Addressed the nit in 97e34b0c: added `tests/overhead_emote_parity.test.ts`, which locks `chat.ts`'s `OVERHEAD_EMOTES` against `sim/types`' `OVERHEAD_EMOTE_IDS` at compile time (a type-level completeness assertion that reddens tsc the moment the sim list gains an id the local array is missing) and at run time (a set-equality test plus `isOverheadEmoteId` coverage), and points the seam-purity comment in `architecture.test.ts` at it. Also merged `release/v0.19.0` to clear the conflict.

FernandoX7 (approved):

PR #1316, opened 2026-07-02
feat(ui): quest direction indicators (map objective areas + target mob marks)

Blaine1705:
## What

Quest direction indicators, the classic way: answer "where do I go for this
quest?" without breaking the vanilla feel.

- **World-map objective areas.** The map (M) now shades a translucent blue
  area over where each active, incomplete objective happens: the camps of
  kill/collect target mobs, the spread of quest ground objects, and interact
  NPCs. Derived from the static content tables (CAMPS / GROUND_OBJECTS /
  NPCS), so the online interest radius never hides a far-away camp; areas
  scale with map zoom and disappear as objectives complete (a turn-in-ready
  quest shows nothing, the existing '?' glyph guides that).
- **Quest-target mob marks.** Live mobs that advance an active objective
  (kill targets, plus mobs whose tagged loot feeds a collect objective) show
  a blue-tinted '!' on their nameplate, distinct from the gold quest-giver
  glyph, so the player can tell "this one counts" on arrival. Lootable '$'
  still wins the marker slot.

## How

- The shared derivation is a new pure sim leaf, `src/sim/quest_targets.ts`
  (the threat.ts / format_money.ts pattern): `questTargetMobIds()` and
  `questObjectiveAreas()` over QUESTS/CAMPS/MOBS/GROUND_OBJECTS/NPCS plus the
  live quest log. Unit-tested against the real content tables
  (tests/quest_targets.test.ts), fixtures found by shape so a content rename
  fails loudly.
- Map: `MapQuestAreaMarker` on the `map_window_view` pure core (Sim vs
  ClientWorld parity asserted) + a translucent circle layer in
  `map_window_painter`, drawn under the title/POI/glyph text. Colors ride two
  new `--color-map-quest-area-*` tokens resolved in the existing
  one-getComputedStyle pass.
- Nameplates: the painter caches the target-mob set behind a quest-log
  signature, so the content tables are never scanned per frame; the marker
  rides the existing `setNameplateStatic` sig/slot.
- No new i18n keys (glyph-only surfaces). No IWorld change (questLog and the
  content tables were already reachable by both consumers).
- One hygiene commit: `tokens.css` was the lone CRLF-committed styles module,
  which trips `biome ci --changed` the first time any branch touches it;
  normalized to LF (content byte-identical, verified with
  `git diff --ignore-cr-at-eol`).
- Includes the same small pre-existing fix cherry-picked from the
  target-frame branch (c1ae201d): world_api/chat.ts becomes a real type-only
  sim import and the architecture guard's allowlist lookup works on Windows.

## Testing

- New Vitest: quest_targets.test.ts (target sets, camp coverage, ground-object
  enclosure, completion/ready exclusions, dedup); map_window_view.test.ts
  grows a quest-area block (both world shapes identical, zoom scaling, empty
  and ready logs plot nothing).
- Guard suites green: architecture, localization_fixes, css_value_validity,
  css_corpus, styles_extraction, map/minimap/nameplate/quest suites
  (108 tests across the affected files).


Blaine1705:
https://github.com/user-attachments/assets/2f199a95-5b7b-4007-9777-05285c37c37e


https://github.com/user-attachments/assets/84a4c3cf-944f-47bb-8e3b-34078a5b2949

FernandoX7 (commented):
Nice, cohesive feature. The derivation is a proper pure sim leaf (`quest_targets.ts`) shared by both consumers, the map and nameplate paths sit behind the existing pure cores and painters with no new `IWorld` surface, and the hot paths are guarded (the nameplate target set is cached behind a quest-log signature, the side list rebuilds only on a real signature change). Determinism and cross-host parity hold: the quest-log Map order is preserved end to end, so the areas and the numbering come out identical offline and online, and the number matches across the tracker, the map list, and the map badges. tsc, biome on the changed files, and the map/nameplate/quest suites are green here. The a11y instincts are good too (aria-expanded/controls/pressed, role=button rows, focus-visible, the tracker-row keydown guard).

A few things worth a look.

## should-fix

**Keyboard activation of the new map controls collides with the game keybinds.** The `#map-quests-toggle` and `#map-quests` listeners wire only `click` (`src/ui/hud.ts`, setup around 1305 to 1370). The map window is not modal (`isModalOpen`, hud.ts:11087), so `canUseGameKeys()` stays true while it is open, and the window-level key handler suppresses only on a focused input/textarea, not on buttons (`src/game/input.ts:633-640`). So with the Quests button or a `.mapq-track` toggle focused, Space fires jump (and the handler's `preventDefault` on Space swallows the button's own activation) and Enter fires Open Chat. This is exactly the case the tracker-row guard you added in this PR and the delve/lockpick guard both `stopPropagation` to prevent. It is a pre-existing gap for the map window (its `#map-zoom`/`#map-close` buttons leak the same way, there is no `#map-window` keydown guard today), so not a regression you introduced, but the two new controls extend it. One Enter/Space keydown guard on `#map-window` buttons closes both.

**Toggling a quest's tracking drops keyboard focus.** The `.mapq-track` click rebuilds `#map-quests` via `innerHTML` (`renderMapQuestList`), which destroys the focused button, so focus falls to `<body>` and the new `aria-pressed` is never announced. The file already has the pattern for this: `toggleQuestTrackerCollapsed` (hud.ts:5086) deliberately re-focuses `.qt-header` across its innerHTML rebuild, and `QuestLogWindow.render` calls `focusFirstInteractive`. Re-focus the toggled row after the rebuild.

**`questObjectiveAreas` returns `center` by reference into the content tables.** `src/sim/quest_targets.ts` passes `camp.center` (:109) and `npc.pos` (:139) straight into the `push` closure, which stores the reference (:103), so `area.center` aliases `CAMPS[i].center` / `NPCS[id].pos`, the same objects the sim uses for spawning. Both consumers only read it today, so this is latent, but a future write to `area.center.*` would corrupt shared content across all three hosts. `pushObjectCluster` already builds a fresh `{x,z}`; do the same on the camp and NPC paths.

**The `fix(architecture)` commit is a no-op against this base and is the cause of the `tests/architecture.test.ts` conflict.** `release/v0.19.0` already carries the identical `src/world_api/chat.ts` change and the same `posixRel` + empty-allowlist edit (plus `overhead_emote_parity.test.ts`). When you resolve, keep the base file and re-add only the real change here: `'src/ui/map_quest_list_view.ts'` in `UI_PURE_CORES`.

## nit

- No direct test for the interact to `targetNpcId` to `NPCS` point-area branch (`quest_targets.ts:138-139`). The object-cluster branch is covered but this one is not, so a content rename of an interact NPC id would not fail loudly.
- The quest-area radius scales by the X factor only (`map_window_view.ts:315`). Correct today because every zone Z-band is 360 (= `WORLD_SIZE`), so the band is square, but it would distort if a band ever changed. Cheap to make robust (an ellipse, or an assertion that the spans match).
- The `.qt-title` rows are `role="button"` with an accessible name of just number + title (hud.ts:5076); nothing signals that they open the quest log.
- Worth an on-device check: on short screens the static `#map-quests` under the canvas (26vh) plus the auto-height `#map-window` could run past the viewport (`hud.mobile.css`).

## conflicts

The other three (`i18n.resolved.generated/pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`) are the usual stale-base regen churn. Merge the base and re-run the i18n build to regenerate them.

Blaine1705:
Update: this PR now also carries the mob-hover-tooltip branch (#1244) merged in as a prerequisite, plus a new commit on top: Questie-style quest lines in the mob tooltip.

Why: the new commit replaces the overhead '!' quest-target marker (from this PR) with quest lines inside the mob hover tooltip (from #1244): quest title in green over the objective's live progress in gold (e.g. Wolves at the Door / Forest Wolf slain: 3/8), re-rendered as counts advance. It needs both features, and neither had landed yet, so #1244 rides along here.

Reviewing order suggestion: if #1244 is merged first, this diff shrinks back to the quest work automatically; if this one goes first, #1244 becomes a no-op. Either order is safe (shared commits are identical).

Blaine1705:
tooltip

PR #1317, opened 2026-07-02
feat(ui): responsive window shell with sticky centered headers and resizable windows

EnriqueGF:
## Problem

Windows do not reliably scale to the viewport. On mobile, height caps based on 100vh hide the bottom of windows behind the browser URL bar; on any screen, scrolling a tall window scrolls its title bar and close button out of view; and a window moved at one viewport size can reopen off-screen at another.

## What this does

**Windows always fit the viewport (desktop and mobile).**
- The shared `.window` shell now clamps to `max-width: calc(100vw - 16px)` and `max-height: min(85vh - 24px, 100dvh - 16px)`.
- All mobile window height caps switch from `100vh` to `100dvh`, so the visible viewport (URL bar included) is what actually bounds the window. The pinned browser floor (Safari/iOS 17.2, Chrome 120, Firefox 121) fully supports `dvh`.
- A window moved or resized at an earlier viewport re-clamps its position at show time (`syncWindowOpenState`), closing the reopen-off-screen path the resize-event re-clamp cannot see (it skips hidden windows).

**Sticky header: centred title, close button pinned top-right.**
- `.window > .panel-title` is now `position: sticky`: it spans the full window width flush at the top and content scrolls beneath it, over a solid theme-driven `--panel-base` background. The title is centred; `.x-btn` and `.panel-title-actions` clusters are pinned to the top-right corner. No per-window markup changes.
- The header offsets derive from a new `--window-pad` custom property; the five windows that pad 14px (delve board, lockpick, confirm, map, arena) now override the variable instead of `padding`, so the header stays flush everywhere. (Sticky insets resolve against the scrollport's content edge, so the sticky `top` must mirror the negative margin; commented in `layout.css`.)
- On mobile the header reserves symmetric 52px inline padding and a 48px min-height so the centred title never slides under the 40x40 touch close target.

**Resizable windows.**
- Content windows get a south-east corner resize grip, drawn as a background layer on the window itself so it stays pinned in the corner while content scrolls and survives the `innerHTML` rebuilds every window does (no handle element to lose).
- New `src/ui/window_resize.ts` (controller: corner band hit-test, pointer capture, non-passive touchmove guard so touch resizing never scrolls the page, and the same visual-vs-author `--ui-scale` correction as window dragging) plus `src/ui/window_resize_core.ts` (DOM-free clamp math, registered in `UI_PURE_CORES`, unit-tested). Sizes clamp to a 220x140 minimum and the viewport.
- Fixed-purpose windows are excluded: map, loot popup, confirm/text dialogs, lockpick board, emote editor, mobile drawer.

## Verification

Drove the real game headless (offline world, Vite dev server): quest log, character, bags, spellbook, market, options, social, talents, leaderboard, and arena at 1440x900, 1100x520, 390x844 portrait, and 844x390 landscape. Confirmed every window fits its viewport, the header stays pinned during real scrolling, title-drag still works, resize drags produce exact author-space sizes at `uiScale: 1.25`, and dialogs stay non-resizable.

- Full `npm test` passes; `tsc --noEmit` clean; Biome clean on changed files.
- `tests/window_resize.test.ts` covers the corner hit-test and clamp math.
- One pinned expectation updated: `client_shell.test.ts` asserts the market window's mobile rule text, now `100dvh`.

No new i18n keys (the grip is a pointer affordance with no element, matching the pointer-only window-drag precedent). No sim/server/wire changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8573,6 +8573,21 @@
         }
       }
     },
+    "node_modules/svelte-check/node_modules/picomatch": {
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-4.0.4.tgz",
+      "integrity": "sha512-QP88BAKvMam/3NxH6vj2o21R6MjxZUAd6nlwAS/pnGvN9IVLocLHxGYIzFhg6fUQ+5th6P4dv4eW9jX3DSIj7A==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "engines": {
+        "node": ">=12"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/jonschlinkert"
+      }
+    },
     "node_modules/svelte/node_modules/aria-query": {
       "version": "5.3.1",
       "dev": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -8573,21 +8573,6 @@
         }
       }
     },
-    "node_modules/svelte-check/node_modules/picomatch": {
-      "version": "4.0.4",
-      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-4.0.4.tgz",
-      "integrity": "sha512-QP88BAKvMam/3NxH6vj2o21R6MjxZUAd6nlwAS/pnGvN9IVLocLHxGYIzFhg6fUQ+5th6P4dv4eW9jX3DSIj7A==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "peer": true,
-      "engines": {
-        "node": ">=12"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/jonschlinkert"
-      }
-    },
     "node_modules/svelte/node_modules/aria-query": {
       "version": "5.3.1",
       "dev": true,
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -16,7 +16,7 @@
   /* ---------- delve board ---------- */
   #delve-board {
     width: 360px;
-    padding: 14px;
+    --window-pad: 14px;
   }
   #delve-board .delve-board-name {
     font-family: var(--title-font);
@@ -210,7 +210,7 @@
   #lockpick-panel {
     width: 420px;
     max-width: 92vw;
-    padding: 14px;
+    --window-pad: 14px;
     left: 50%;
     top: 50%;
     transform: translate(-50%, -50%);
@@ -2444,7 +2444,7 @@
     top: 50%;
     transform: translate(-50%, -50%);
     width: 360px;
-    padding: 14px;
+    --window-pad: 14px;
     z-index: 60;
   }
   #confirm-dialog .cd-body {
@@ -3254,7 +3254,7 @@
   /* ---------- map ---------- */
   /* map */
   #map-window {
-    padding: 14px;
+    --window-pad: 14px;
   }
   /* zoom controls, matching the micro-menu button look */
   #map-zoom {
@@ -3298,7 +3298,7 @@
   /* ---------- Ashen Coliseum (arena) ---------- */
   #arena-window {
     width: 360px;
-    padding: 14px;
+    --window-pad: 14px;
   }
   .arena-sub {
     font-family: var(--title-font);
--- a/src/styles/hud.css
+++ b/src/styles/hud.css
@@ -1505,11 +1505,6 @@
     width: 320px;
     z-index: 95;
   }
-  #report-window .panel-title {
-    display: flex;
-    justify-content: space-between;
-    align-items: center;
-  }
   #report-window [data-close] {
     background: transparent;
     border: 0;
--- a/src/styles/hud.mobile.css
+++ b/src/styles/hud.mobile.css
@@ -334,7 +334,7 @@
     transform: translateX(-50%);
     display: none;
     width: min(560px, calc(100vw - 32px - env(safe-area-inset-left) - env(safe-area-inset-right)));
-    padding: 10px;
+    --window-pad: 10px;
     pointer-events: auto;
     z-index: 100;
     border: 2px solid #5b4618;
@@ -344,15 +344,11 @@
       0 8px 28px #000d,
       inset 0 1px 0 #ffffff12;
     /* Top-centred modal: cap to the safe viewport so short landscape phones can
-       scroll the full menu without hiding rows behind controls or browser chrome. */
+       scroll the full menu without hiding rows behind controls or browser chrome.
+       No --ui-scale division: this drawer is a <body>-level window OUTSIDE the
+       zoomed #ui (see the --window-scale note in layout.css). */
     max-width: calc(100vw - 32px - env(safe-area-inset-left) - env(safe-area-inset-right));
-    max-height: calc(
-      var(--app-vh) /
-      var(--ui-scale, 1) -
-      32px -
-      env(safe-area-inset-top) -
-      env(safe-area-inset-bottom)
-    );
+    max-height: calc(var(--app-vh) - 32px - env(safe-area-inset-top) - env(safe-area-inset-bottom));
     overflow-y: auto;
     overscroll-behavior: contain;
     -webkit-overflow-scrolling: touch;
@@ -362,7 +358,10 @@
     flex-direction: column;
   }
   body.mobile-touch #mobile-extra-controls .panel-title {
-    min-height: 32px;
+    /* 48px, not the old 32px: the close button is a 40x40 absolute-pinned touch
+       target now (it no longer props the header height from the flex flow), so
+       the header itself must be at least as tall as the target. */
+    min-height: 48px;
     margin-bottom: 8px;
     padding-bottom: 6px;
     cursor: move;
@@ -909,10 +908,10 @@
   }
   /* The mobile close button is a 40x40 touch target pinned in the header's
      top-right corner (absolute, see the .window > .panel-title shell rule in
-     layout.css), so reserve symmetric room for it: the centred title must never
-     slide under it, and the header must be at least as tall as the target. */
+     layout.css), so reserve inline-end room for it: the title must never slide
+     under it, and the header must be at least as tall as the target. */
   body.mobile-touch .window > .panel-title {
-    padding-inline: 52px;
+    padding-inline: var(--window-pad) 52px;
     min-height: 48px;
   }
   body.mobile-touch #leaderboard-window,
@@ -1651,11 +1650,13 @@
       /* var(--app-vh), not dvh: viewport units resolve against the initial
          containing block, which can disagree with the JS-synced app viewport
          (see the #mobile-controls note above); --app-vh is this HUD's one
-         source of viewport height. */
+         source of viewport height. Divided by --ui-scale because #community-hud
+         lives INSIDE the zoomed #ui (the layout.css rule of thumb: divide
+         inside #ui, raw outside). */
       top: clamp(
         calc(max(8px, env(safe-area-inset-top)) + 210px),
-        calc(var(--app-vh) * 0.44),
-        calc(var(--app-vh) - 230px)
+        calc(var(--app-vh) * 0.44 / var(--ui-scale, 1)),
+        calc(var(--app-vh) / var(--ui-scale, 1) - 230px)
       );
       right: max(8px, env(safe-area-inset-right));
       bottom: auto;
@@ -1807,10 +1808,10 @@
       left: 50%;
       top: max(10px, env(safe-area-inset-top));
       bottom: auto;
-      padding: 8px;
+      --window-pad: 8px;
+      /* No --ui-scale division: body-level window outside the zoomed #ui. */
       max-height: calc(
-        var(--app-vh) /
-        var(--ui-scale, 1) -
+        var(--app-vh) -
         28px -
         env(safe-area-inset-top) -
         env(safe-area-inset-bottom)
--- a/src/styles/hud.mobile.css
+++ b/src/styles/hud.mobile.css
@@ -72,7 +72,16 @@
   }
   body.mobile-touch.game-active #mobile-controls {
     position: absolute;
-    inset: 0;
+    /* Size from the shared --app-vw/--app-vh box (the one #game-canvas, html/body,
+       #ui, and #nameplates use), NOT `inset: 0`: inset resolves against the initial
+       containing block, and whenever that disagrees with the JS-synced app viewport
+       (device emulation, pinch zoom, a mid-resize snapshot) the joysticks and combat
+       buttons anchor to a different bottom edge than the action bar and drift into
+       the middle of the screen. One box for every full-screen HUD layer. */
+    left: 0;
+    top: 0;
+    width: var(--app-vw);
+    height: var(--app-vh);
     display: block;
     pointer-events: none;
     z-index: 60;
@@ -337,7 +346,13 @@
     /* Top-centred modal: cap to the safe viewport so short landscape phones can
        scroll the full menu without hiding rows behind controls or browser chrome. */
     max-width: calc(100vw - 32px - env(safe-area-inset-left) - env(safe-area-inset-right));
-    max-height: calc(100dvh - 32px - env(safe-area-inset-top) - env(safe-area-inset-bottom));
+    max-height: calc(
+      var(--app-vh) /
+      var(--ui-scale, 1) -
+      32px -
+      env(safe-area-inset-top) -
+      env(safe-area-inset-bottom)
+    );
     overflow-y: auto;
     overscroll-behavior: contain;
     -webkit-overflow-scrolling: touch;
@@ -848,7 +863,7 @@
     bottom: calc(31px + env(safe-area-inset-bottom));
     width: min(220px, calc(50vw - 144px));
     min-width: 142px;
-    max-height: min(170px, calc(100dvh - 128px));
+    max-height: min(170px, calc(var(--app-vh) / var(--ui-scale, 1) - 128px));
     overflow: hidden;
     z-index: 25;
   }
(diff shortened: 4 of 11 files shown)


EnriqueGF:
Rebased onto `release/v0.19.0` (base retargeted) and pushed a follow-up commit fixing an aspect-ratio HUD split found while testing this branch across viewport sizes.

**Symptom:** at some aspect ratios / in device emulation, the touch joysticks and combat buttons anchored to the middle of the screen while the action bar stayed at the bottom.

**Root cause:** `#mobile-controls` was anchored with `inset: 0` (resolves against the initial containing block), while the canvas, `html`/`body`, `#ui`, and `#nameplates` size from the JS-synced `--app-vw`/`--app-vh` box. Whenever the two disagree (meta-viewport emulation, pinch zoom, a mid-resize snapshot), the HUD splits into two bottom edges.

**Fix:** every fullscreen HUD layer now sizes from the one `--app-vw`/`--app-vh` box (`#mobile-controls` now matches the `#nameplates` pattern), and the window shell + mobile window caps clamp against `var(--app-vh) / var(--ui-scale, 1)` (exact under the `#ui` zoom) instead of raw viewport units. Verified by forcing an app-box/ICB mismatch in a real browser: all layers now stay coherent, and re-anchor correctly on resize.

An aspect-ratio sweep (320x480, 390x844, 844x390, 664x770, 768x1024, 1024x768, 1200x500) also caught the portrait combat row (372px) clipping its end buttons on 320-384px phones; a narrow-portrait variant shrinks the row to 312px.

EnriqueGF:
## Screenshots across resolutions

### Desktop 1440x900: character window (centred title, close pinned top-right, resize grip bottom-right)
desktop char

### Desktop 1440x900: same window after dragging the corner grip (content reflows)
desktop char resized

### Desktop 1100x520 (short viewport): spellbook scrolled, the header stays pinned while the list scrolls under it
spellbook scrolled

### Mobile portrait 390x844: character window fits, 40x40 touch close target inside the header
mobile portrait char

### Mobile landscape 844x390: quest log, touch controls anchored to the bottom corners
mobile landscape questlog

### Squarish 664x770 (the aspect ratio from the report): joysticks and combat row correctly bottom-anchored after the --app-vw/--app-vh fix
squarish hud

### Narrow phone 320x480: the shrunk combat row fits all seven buttons (previously Attack/More clipped off both edges)
narrow hud

(Images hosted on the `pr-1317-screenshots` branch of my fork.)

FernandoX7 (commented):
This is a strong pass at the window system and most of the hard parts are right. The sticky header needs no per-window markup changes and the one rule deletion it required (#report-window, hud.css) was correctly identified as load-bearing, not cleanup. The visual/author space handling is correct everywhere I traced it: pinWindow feeds visual-space rects to setWindowPixelPosition which divides internally, the resize core keeps its author-space contract, and the show-time re-clamp runs off the MutationObserver after the display flip, so the rect is real. The mobile-controls inset fix (anchoring to the one --app-vw/--app-vh box, matching #nameplates) is a genuinely good catch, as is the narrow-portrait combat row. window_resize_core in UI_PURE_CORES with direct tests is exactly the house pattern. Targeted suites pass and tsc is clean at the head.

Findings, most important first:

1. **should-fix (direction):** The header centres the title (layout.css:64-66). The direction for this system is title top left, close button pinned top right and vertically centred with it. Centred is a defensible classic look, but left is what we want here; it is a two-line change (justify-content: flex-start, text-align: left, drop the symmetric start padding).

2. **should-fix:** The --ui-scale division is applied to two windows that live outside the zoomed #ui, and skipped on one element inside it. #mobile-extra-controls (index.html:1211, outside #ui) divides all its caps (hud.mobile.css:349, 908 via the generic .window rule, 1811): at uiScale 0.85 the cap exceeds the real viewport, which recreates the exact bottom-cut-off bug this PR fixes, and at 1.4 the drawer is capped about 30 percent short. The dynamically built #confirm-dialog/input modal is appended to document.body (hud.ts:9703), so the new shell clamps (layout.css:37-41) over-clamp it at uiScale above 1 and let it exceed the viewport below 1. Meanwhile #community-hud is inside #ui but got re-anchored with raw var(--app-vh) (hud.mobile.css:1650-1659); that one is not a regression versus 44dvh, but it contradicts the rule the PR itself states in layout.css. Rule of thumb the commit already documents: divide inside #ui, raw outside.

3. **should-fix:** The mobile More drawer misses the --window-pad migration and the header floor. body.mobile-touch #mobile-extra-controls keeps padding: 10px (hud.mobile.css:337) and 8px in landscape (1810), so the header's -12px margins overhang by 2-4px per side (clipped corners, and overflow-x can summon a horizontal scrollbar). Its pre-existing .panel-title rule (364, higher specificity in the same layer) also beats the new min-height: 48px floor with 32px, and since the x-btn is now absolute it no longer props the header height, so the 40px close target pokes past both header edges. Convert both paddings to --window-pad and align that title rule.

4. **should-fix:** On touch, a bare tap or a scroll gesture starting in the 28px SE band mutates the window before any movement: pointerdown immediately pins (inline left/top, right: auto, transform: none) and stamps windowMoved (window_resize.ts:115-133). The mobile stretched windows (options, quest-log, spellbook, vendor, market, quest-dialog at hud.mobile.css:925-938, trade at 996) get width from left: 10px; right: 10px; width: auto, so the pin collapses them to shrink-to-fit for the rest of the session; nothing restores the CSS layout. The band also swallows scroll starts over plain-div content (market .mkt-row is not caught by the control guard at window_resize.ts:95). Title-drag has the same pin-on-tap behavior today, but the SE corner is scroll territory, so accidents are likely. Deferring the pin and both dataset stamps to the first pointermove past a small threshold fixes taps; worth also deciding whether the stretched mobile windows should resize at all.

5. **should-fix:** Corner band vs scrollbar. The window is its own scrollport with classic 8px gutter scrollbars (base.css:33-36), so the vertical scrollbar's bottom end sits inside the 18px band; in Chrome a pointerdown on a scrollbar targets the scrollable element, which passes cornerHit, so grabbing the thumb near the bottom starts a resize with preventDefault instead of a scroll. Hit-testing the band against the client box (clientWidth/clientHeight) instead of the border box keeps the gutter grabbable. Worth a one-minute manual check in Chrome.

6. **should-fix (perf):** The body-wide MutationObserver in window_resize.ts:71-80 is pure overhead today: every resizable window is static in index.html and the only dynamically created .window.panel (#confirm-dialog) is excluded. It duplicates the identical per-added-node matches plus querySelectorAll pass hud.ts:1421-1437 already runs, so all combat-time DOM churn (FCT appends, log lines, nameplate subtrees) is scanned twice for zero effect. Stamp once at install and expose a stamp hook, or piggyback on the Hud observer.

Nits:

- package-lock.json: the picomatch removal is unrelated churn (package.json is untouched); drop it from the PR.
- layout.css:38-41: the min() second arm can never win (0.85H - 24 is below H - 16 for all H); the old vh/dvh split that motivated it collapsed when both arms moved to --app-vh. Simplify to the first arm.
- dataset.windowSized (window_resize.ts:132) is written and read nowhere, and the comment above it ("sized keeps the manual size") describes behavior that does not exist; the size persists because the styles are inline. Drop the flag or the comment.
- Spellbook is the one .panel-title-actions user (spellbook_window.ts:122); the pinned cluster is about 105px wide on desktop against the 34px symmetric reservation, so a long title can slide under the reset button. Fine to leave if you accept the overlap risk on narrow widths.
- Resizing leaderboard/daily-rewards taller than about 56vh adds dead space because .lb-body is hard-capped (components.css:1168); the grip will read as broken there. Converting those bodies to the market-style flex column fixes it.
- The hover branch runs cornerHit (two closest walks plus getBoundingClientRect) on every pointermove over a window; cheap to skip entirely for coarse pointers, which have no hover.
- The resize session baseline is captured before pinWindow can clamp the window (window_resize.ts:116-118), so the max-size clamp can compute from a stale left/top; the shell CSS contains it, as the core's comment concedes, but capturing after the pin is one line.
- A few inherited desktop rules in the later components layer still override the new shell clamps with raw viewport units (bags 2742, social 2955, market 3556, lockpick 210), so "can never exceed the viewport" does not fully hold on desktop under the divergence cases the PR names. Fine as a follow-up.
- The map window is the one window with no header at all (its x-btn is a direct window child, index.html:373), so the always-reachable close does not apply to it, and its mobile padding: 8px (hud.mobile.css:1412) stayed raw. Harmless today; worth a code comment so the pattern is not broken silently later.

Verification notes: window_resize, client_shell, and architecture suites pass at the head (99 tests) and tsc --noEmit is clean. The show-time re-clamp, placeNewWindow skip for moved windows, and the resize-listener premise all check out against hud.ts. play.html shares the same style barrel and main.ts bootstrap, so no parity gap there.

EnriqueGF:
Thanks for the thorough review, all six should-fixes plus most nits are addressed in 3801fe9f:

1. **Title direction:** header is now top left (flex-start, left text), close button pinned top right and vertically centred. The inline-start padding is `var(--window-pad)` so the title aligns with the body content.

2. **--ui-scale scope:** new `--window-scale` custom property on the shell (defaults to `var(--ui-scale, 1)`); the body-level windows outside the zoomed #ui (`#mobile-extra-controls`, `#confirm-dialog`) pin it to 1, and their two bespoke caps dropped the division too, so the drawer is exact at uiScale 0.85 and 1.4. `#community-hud` (inside #ui) now divides like the rest, per the rule of thumb the shell comment states.

3. **More drawer:** joined the `--window-pad` system (10px, 8px landscape) so the sticky header is flush again, and its `.panel-title` rule now carries the 48px floor (verified: header 48px, 40x40 close inside it, full-bleed, no horizontal overflow at 390x844).

4. **Tap-safe resize:** the corner press records a pending session only; pin + `windowMoved` + capture + the resizing class all wait for 4px of travel (`RESIZE_ENGAGE_SLOP`). Verified in-browser: a bare click in the band leaves zero inline styles and no dataset flag; a drag past the slop resizes normally. The baseline rect is captured after the pin clamps it (your one-liner). One deliberate remainder: the non-passive touchmove guard still attaches at touch-down while pending, because by first-move the browser has already claimed the pan and resize-by-touch would never engage; the cost is that a scroll gesture started inside the 28px corner is swallowed, which the client-box hit-test below shrinks further.

5. **Scrollbar vs band:** the hit-test now measures the corner against the client box, so the scrollbar gutter misses it. Verified in Chrome: hover over the vertical scrollbar's bottom end shows no resize affordance and scroll-drag works; just inside the client corner the affordance appears.

6. **Observer:** the module-local body-wide MutationObserver is gone; `installWindowResize` stamps the document once and Hud's existing window observer calls the exported `markResizableWindow` for late-added windows.

Nits: package-lock churn reverted, the never-winning `min()` arm simplified away, `dataset.windowSized` dropped, hover hit-testing skips touch pointers, and a shell comment documents `#map-window` as the deliberate headerless exception. Left as accepted/follow-ups per your notes: the spellbook actions overlap risk, the `.lb-body` hard cap under tall resizes, and the handful of later-layer desktop rules still using raw viewport units.

Gates at head: window_resize / client_shell / css guards / architecture / hud_perf_budget green, `tsc --noEmit` clean.

FernandoX7 (commented):
Solid follow-up. The structural asks all landed and landed well: the title sits top left with the close button pinned and vertically centred, `--window-scale` is the right mechanism for the inside/outside `#ui` split (and the doctrine comment in layout.css will keep it honest), the drawer joins the `--window-pad` system in both orientations with the 48px header floor restored, the corner band now hit-tests the client box so scrollbar grabs stay scrolls (the math is correct under zoom, and the two outside-`#ui` windows can never reach it since both are excluded from resize), and the observer dedup through `markResizableWindow` covers late-added windows via Hud's existing observer. Lockfile churn is gone from the net diff. Targeted suites pass (101 tests) and tsc is clean at the head.

Three things to finish before this merges:

1. **should-fix:** One `--ui-scale` division leaked past the `--window-scale` rework: `body.mobile-touch .window { max-height: calc(var(--app-vh) / var(--ui-scale, 1) - 40px) }` (hud.mobile.css:907). The hud-mobile layer beats layout, and there is no more specific mobile rule for `#confirm-dialog`, so this governs the confirm/input dialog, which is body-level and outside `#ui`: at uiScale 0.85 the tall multiline import/export dialog can exceed the real viewport again. The `--window-scale: 1` pin cannot reach a rule that bakes `--ui-scale` directly, which is exactly the trap the new layout.css comment warns about. One-line fix: divide by `var(--window-scale)` here (it is defined on every `.window` by the layout rule).
2. **should-fix:** `RESIZE_ENGAGE_SLOP = 4` visual px is below normal finger tap wobble (browsers use roughly 10px of internal tap slop for this reason), so on touch a plain corner tap can still engage: the engaging move pins the window, stamps `windowMoved`, and writes inline width/height. Nothing looks wrong at tap time, but the window silently stops being CSS-managed (frozen inline size on rotation, skipped by `placeNewWindow`). A larger slop for `ev.pointerType === 'touch'` (8 to 10px) closes it; note the 4px is also visual space, so at uiScale 1.4 it is only about 3 author px.
3. **should-fix:** The core of the tap-safety fix is unpinned: nothing tests that a tap in the band leaves the window untouched, and the new `RESIZE_ENGAGE_SLOP` test only asserts the constant's range, which cannot regress. A small jsdom test that calls `installWindowResize`, dispatches pointerdown plus a sub-slop move plus pointerup on a stamped window, and asserts no inline styles, no `windowMoved`, and no class would pin both this and the engage-once path.

Small stuff, none blocking:

- `isInResizeCorner`'s comment ("callers only pass pointers already over the window, so only the inner edge of the band is checked") is now wrong in a load-bearing way: after the client-box change, the outer-edge check is precisely what keeps the scrollbar gutter out. Update the comment so a future simplification does not reintroduce the bug.
- The grip visual still paints at the border-box corner (`right 3px bottom 3px`) while the hit zone now ends at the client box, so on a scrolled window the hatch sits partly under the scrollbar track and points at a corner that no longer resizes. Cosmetic.
- While a touch press is pending in the band, the document-wide touchmove guard blocks scrolling for every finger, not just the pending one, and a scroll gesture that starts in the band still becomes a resize after the slop. I read the comment explaining why the guard must attach while pending, and I think that trade is fine for a corner grip; flagging so the choice is on record. Optional hardening while you are in there: bail the pending session when `ev.buttons === 0` on move, so a swallowed pointerup cannot strand it.
- The updated client_shell pin for the drawer header dropped its selector line, so it now matches any rule with those four declarations; re-anchor it.
- vendor-open `#vendor-window`/`#bags` headers stay a fixed 47px, 1px under the new 48px floor by id specificity. The 40px button still fits, so this is doctrine, not breakage.
- `body.mobile-touch #map-window { padding: 8px }` (hud.mobile.css:1411) is now the one remaining raw window padding; harmless while the map is headerless, but worth converting or commenting given the new layout.css note.
- Still open from last round, fine as follow-ups: the spellbook actions cluster vs the 34px reservation (less likely to bite now the title is left-aligned), the `.lb-body` 56vh cap that makes resize look broken on leaderboard/daily-rewards, the four desktop components-layer rules that still clamp with raw viewport units (bags, social, market, lockpick), and the mobile width axis (`max-width: calc(100vw - 20px)` and friends) which was left raw while heights were converted.

One more ask: could you drop updated screenshots on the PR now that the header is left-aligned? Desktop (a couple of the busier windows like market or spellbook, one scrolled so the sticky header and grip are visible) and web mobile portrait and landscape, including the More drawer. Makes it much easier to sign off the visual half of this.

EnriqueGF:
Thanks for the second pass. All three should-fixes are in 2e14270c:

1. **Leaked `--ui-scale` division:** the generic `body.mobile-touch .window` max-height now divides by `var(--window-scale, 1)`, so the body-level confirm/input dialog (which pins the variable to 1) clamps exactly at any uiScale, and windows inside `#ui` resolve the same value as before through the layout.css default. The rule carries a comment pointing at the shell doctrine so the trap is not silently reintroduced.

2. **Touch engage slop:** touch presses engage after a new `RESIZE_ENGAGE_SLOP_TOUCH` (10 visual px, at the browser's own internal tap-slop budget; mouse stays at 4). I also took your optional hardening: a pending session bails when a pointermove arrives with `buttons === 0`, so a swallowed pointerup can no longer strand it.

3. **Tap-safety test:** one note, the repo deliberately ships no jsdom (tests/CLAUDE.md mandates hand-rolled fake DOM for UI contracts), so the new tests stub `document`/`window` and drive the real `installWindowResize` controller end to end: a sub-slop tap leaves the window with no inline styles, no `windowMoved`, no class; engage happens exactly once past the slop and resizes from the pinned baseline; touch needs the wider slop; and the `buttons === 0` bail is covered. A constant-invariant test also pins touch slop > mouse slop, >= 8, and inside the touch band.

Small stuff: the `isInResizeCorner` comment now states the outer-edge checks are exactly what keep the scrollbar gutter out (with a do-not-simplify warning), the client_shell drawer-header pin is anchored to its selector again, and mobile `#map-window` joined the `--window-pad` system. Left as accepted per your notes: the grip visual sitting at the border-box corner on scrolled windows (cosmetic), the vendor-open 47px header (doctrine), the touchmove-guard trade you put on record, and the follow-ups from round one.

Gates at head: window_resize / client_shell / css guards / architecture / hud_perf_budget all green (143 tests), `tsc --noEmit` clean, Biome clean on changed files, and a fresh qa-checklist pass over the diff reports zero blocking findings.

Updated screenshots (left-aligned headers), desktop 1440x900 and web mobile 390x844 / 844x390:

**Desktop: market scrolled (sticky header pinned, grip in the corner) and spellbook scrolled**

desktop market scrolled

desktop spellbook scrolled

**Mobile portrait: quest log and the More drawer**

mobile portrait quest log

mobile portrait drawer

**Mobile landscape: spellbook scrolled and the More drawer**

mobile landscape spellbook

mobile landscape drawer

FernandoX7 (approved):
GREAT JOB! Thank you so much for taking this on

PR #1318, opened 2026-07-02
fix(gamepad): take no pad input while the window is unfocused

madmatah:
## Summary

The browser only delivers keyboard and mouse input to a focused window, but it keeps reporting gamepad state to a visible, unfocused one (confirmed on Linux+Chrome: with the game side by side with another focused window, the pad kept driving the character). This makes the focus rule uniform across every input method: an unfocused window takes no pad input.

`GamepadManager.poll()` now checks `document.hasFocus()`. While the window is unfocused it clears held stick movement (no more character running off on a held stick), consumes the button state without dispatching (a button held across a refocus does not fire a stale edge on return), and skips camera, edge actions, and rumble. The check is fail-open (no DOM or an exception reads as focused), so a quirky environment can never brick controller input.

Note for reviewers, one deliberate behavior change: playing with the pad while another window has focus no longer works; click the game window to resume, exactly like keyboard and mouse. Rumble is not tracked while unfocused, so on refocus a single vibration fires for any health lost meanwhile.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm test`
  - `npx tsc --noEmit`
  - `npm run build` 

- Manual steps: game and another window side by side (game visible, not
  covered), focus the other window, use the pad: the character no longer moves
  or casts; click back into the game and input resumes immediately, with no
  phantom action from buttons held during the switch.

---

### Quality

      TypeScript `strict`).

### Cross-platform

- [x] Tested on desktop with an xbox controller

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/game/gamepad.ts
+++ b/src/game/gamepad.ts
@@ -112,6 +112,14 @@ export class GamepadManager {
     }
   }
 
+  private windowFocused(): boolean {
+    try {
+      return typeof document === 'undefined' || document.hasFocus();
+    } catch {
+      return true;
+    }
+  }
+
   private activePad(): Gamepad | null {
     if (this.index === null || typeof navigator === 'undefined') return null;
     const pad = navigator.getGamepads()[this.index];
@@ -132,6 +140,17 @@ export class GamepadManager {
     };
     const cur: boolean[] = [];
     for (let i = 0; i < STANDARD_BUTTON_COUNT; i++) cur[i] = pressed(i);
+
+    // Match keyboard and mouse, which the browser only delivers to a focused
+    // window: an unfocused window takes no pad input. Consume the button state
+    // without dispatching so a button held across a refocus does not fire a
+    // stale edge on return.
+    if (!this.windowFocused()) {
+      this.input.clearGamepadMove();
+      this.hideCursor();
+      this.prevPressed = cur;
+      return;
+    }
 
     this.checkRumble();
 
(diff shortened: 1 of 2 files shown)


FernandoX7 (commented):
Clean, well-scoped fix, and the ordering is exactly right. `cur[]` is computed before the focus gate, so setting `prevPressed = cur` before the early return (src/game/gamepad.ts:150) means a button held across the refocus lands with `prev[i] = true` and never fires a rising edge on return. Confirmed against `risingEdges` (`cur[i] && !prev[i]`). `windowFocused()` failing open when `document` is missing or throws (gamepad.ts:115) keeps `poll()` safe in the Node/headless hosts, so a quirky environment can never brick the pad. The three tests cover the behaviors that matter (no input while unfocused, no stale edge across refocus, resume on refocus) and they pass locally.

Two things I checked that hold up:

- Rumble: skipping `checkRumble()` while unfocused leaves `lastHealth` at its pre-blur value, so on refocus a single vibration fires for damage taken meanwhile. You called this out in the summary, and it is fine.
- Pointer mode: the focus gate sits ahead of the pointer-mode branch, so while unfocused the virtual cursor stops drifting. That is a strict improvement over the old always-on poll, which would have moved the cursor from an unfocused window.

nit (optional), gamepad.ts:148: in pointer mode the cursor div is left visible but frozen while unfocused, since the early return skips both `updateCursor` and `hideCursor`. Harmless (an unfocused window freezing its UI is expected), but a `this.hideCursor()` in the unfocused branch would tidy it if you want it gone.

Nothing blocking. Nice fix.

madmatah:
@FernandoX7 It’s fixed, thanks for the review!

FernandoX7 (approved):

PR #1323, opened 2026-07-02
feat(combat): optional Clique-style mouseover casting on party frames

Blaine1705:
## What

Clique-style mouseover casting on the party frames, shipped **as an option,
not an imposition**: pressing an action-bar key while the cursor is over a
party frame casts that friendly (heal/buff) ability on the hovered member,
leaving the player's persistent target untouched. This is how healers play
with the classic Clique addon: keep the boss targeted, heal off the frames.

**Opt-out by design.** A new Interface toggle (Esc menu, "Mouseover Cast on
Party Frames", on by default, label translated for the ten filled locales)
turns it off with one click, restoring the classic target-else-self routing
exactly. Even while on, the redirect only engages when the cursor is actually
over a party frame AND the pressed ability is friendly; hostile abilities,
ground-targeted casts (the new talents 2.0 reticle flow), and ordinary play
are untouched either way.

Also: a party invite to a "/dev bot" test dummy (from #1313) is now
auto-accepted, mirroring its whisper auto-reply, so party frames and mouseover
heals can be exercised offline.

## How

- **Sim.** `castAbility` gains an optional explicit friendly-target id, a
  sibling of talents 2.0's ground-target `castAim`. A timed heal re-resolves
  its target at cast FINISH, so the override rides the entity (runtime-only
  `Entity.castTargetId`, never serialized, never on the wire), is consumed by
  `applyAbility` and cleared on cancel, so it can never leak into a later
  cast. `resolveFriendlyTarget` prefers a valid override and falls back to
  the classic current-friendly-target-else-self rule, byte-identical when no
  override is given: the parity gate (rng draw order) stays green.
- **Seam.** `IWorldCombat.castAbilityOn(abilityId, targetId)`, implemented by
  both worlds. Online, the override rides the existing `'cast'` wire token as
  an extra `target` field (no new command token; the server type-checks it
  and routes to `sim.castAbilityOn`, validating friendliness/liveness/range
  server-side like any cast).
- **UI.** Party rows track hover through their live slot (a recycled row
  reports its current member) via a new painter dep; `Hud.castSlot` redirects
  friendly casts to the hovered pid, gated on the option, inside the
  non-ground-targeted branch alongside the reticle flow.

## Testing

- New tests/mouseover_cast.test.ts: timed-heal redirect at finish, selection
  preserved, stale-id and hostile-id overrides falling back, override
  consumed after the cast.
- Dev-bot party auto-accept case in chat.test.ts; onHover threaded through
  the party painter test stubs; the options-list pin; the W0c IWorld parity
  registry grows castAbilityOn (members 149 to 150).
- Full sweep green on top of the current release head (post talents 2.0):
  681 tests across 22 files including tests/parity, architecture,
  command_schema, snapshots, bandwidth, i18n gates.


Blaine1705:
https://github.com/user-attachments/assets/69a6a3ea-2696-45ef-8928-c40e7a0cdeb1

Blaine1705:
ee1ffe01b4b5eb34af329feb6f5fd5d2

PR #1325, opened 2026-07-02
fix(sim): make steep mountains and the world rim genuinely unclimbable

EnriqueGF:
## Problem

Players can climb the mountains that are supposed to fence the world: the ridge walls between zones and the outer rim. From the report screenshots, a player walked over Thornpeak's rim mountains and ended up standing outside the map at x = -248 (the world ends at x = -180), floating over the boundary water plane.

Three holes combined to allow this:

1. The walls were not actually steep. The easiest crossing of the inter-zone ridges had a gradient of 0.63 to 0.79, far below the movement climb limit of 1.5, so players could simply walk over them and skip the road pass. The rim's easiest crossings sat at 1.54 to 1.90, barely above the limit.
2. The climb gate only measured rise/run along the step direction, so strafing diagonally (a switchback) reduced the effective slope below 1.5 and beat the gate on any wall.
3. Airborne movement skipped the gate entirely: holding jump plus forward walked up any face about a yard per hop, and landing snapped the player onto the higher ground.

## Fix

Terrain (`src/sim/world.ts`):
- Zone ridge walls are taller and narrower (height 22 to 40, sigma 18 to 10, crest variance tightened). Every non-pass crossing now peaks above gradient 2.19 at the production seed.
- The world rim rises 55 over 24yd (was 40 over 30) and peaks 6yd inside the boundary; every crossing now exceeds gradient 2.93. Its base stays at the same 30yd offset so the Mirefen impact site still leans on the wall.
- The road passes stay gentle (worst gradient 0.35) and all camps, NPCs, hubs, and road vertices remain on walkable ground; narrowing the ridges actually freed the camps that used to sit inside the old wide bands.
- New pure helpers: `terrainSteepness` (true gradient magnitude), `terrainSteepnessAt` (1yd-cell memo for hot gates), `terrainDownhill`, `nearSteepWalls`.

Movement (`src/sim/sim.ts`):
- Uphill ground steps are blocked when the step slope OR the true terrain steepness at the destination exceeds `MAX_CLIMB_SLOPE`, so approach angle no longer matters.
- Airborne horizontal movement cannot enter terrain rising above the body that could not be walked up; a jump into a steep face drops the player at its base.
- Jumping is blocked while standing on unwalkably steep ground, and standing on such ground slides the player downhill (classic-style) until footing is walkable, so no input pattern can rest partway up a wall.
- The same wall rule applies to charge, follow, knockback, and (screened to the wall bands for hot-path cost) the shared `moveToward` used by mobs, pets, and feared players.

## Verification

- New `tests/climb_slope.test.ts`: switchback and jump-spam cannot cross the rim (both reproduced the exploit and failed before the fix), slide off steep ground, no jumping on steep ground, the road pass still crosses, and normal jumping is unaffected.
- New `tests/terrain_walls.test.ts`: pins the terrain contract at the production seed: every non-pass ridge/rim crossing beats the climb limit with margin, passes stay gentle, and all content sits on walkable ground. The pass shoulder band (|x| 16-34) is now pinned to prevent quiet widenings.
- Parity goldens re-recorded (`UPDATE_PARITY=1`): terrain heights and mob movement legitimately changed. Rng draw order and counts shift downstream: taller terrain plus wall-blocked mobs change proximity-gated aggro/wander/target draws (arena_1v1 479 to 482, nythraxis_full_pull 3299 to 3302, mob_locomotion 15 to 16, and others). The new helpers use stateless `hash2`/`fbm2` with no draw sites added, so the shifts are position-driven conditional draws. Digests were re-recorded and parity is green.
- One `pet_follow` test restaged: its old coordinates relied on a route across the (then climbable) west rim, which correctly no longer exists.
- Tick benchmark at parity with main after memoizing steepness per cell and band-screening the mob gate; `tsc`, build, and the full suite are green modulo the pre-existing flaky set on main.
- A player already stranded outside the rim (like the reporter) can still walk back down into the world without fall damage.
- Reviewed by the repo's architecture-reviewer and qa-checklist agents; their findings (an em dash, packed-key hardening for instance coordinates, an instance short-circuit for the band screen) are applied.

### DIFF
--- a/src/render/critters.ts
+++ b/src/render/critters.ts
@@ -7,10 +7,8 @@
 
 import * as THREE from 'three';
 import { mergeGeometries } from 'three/examples/jsm/utils/BufferGeometryUtils.js';
-import {
-  DUNGEON_X_THRESHOLD, WORLD_MAX_X, WORLD_MAX_Z, WORLD_MIN_Z,
-} from '../sim/data';
-import { terrainHeight, WATER_LEVEL } from '../sim/world';
+import { DUNGEON_X_THRESHOLD, WORLD_MAX_X, WORLD_MAX_Z, WORLD_MIN_Z } from '../sim/data';
+import { terrainHeight, terrainSteepnessAt, WATER_LEVEL } from '../sim/world';
 import { GFX } from './gfx';
 
 export interface CritterField {
@@ -23,6 +21,9 @@ const SPAWN_MIN = 16; // relocate ring around the player (min..max yd)
 const SPAWN_MAX = 30;
 const FLEE_DIST = 6; // critters bolt when the player gets this close
 const EDGE = 8; // keep clear of the world edges
+// Same rise/run limit the sim's movement uses (MAX_CLIMB_SLOPE): wildlife stays
+// off the unclimbable mountain walls and the world rim, like everything else.
+const MAX_WALK_SLOPE = 1.5;
 
 // The Eastbrook Vale / Mirefen Marsh boundary runs along the causeway at z=180.
 // Cheerful overworld critters (rabbits/squirrels/songbirds) thin out as the dry
@@ -44,7 +45,8 @@ export function causewayPopScale(pz: number): number {
 function mulberry32(seed: number): () => number {
   let a = seed >>> 0;
   return () => {
-    a |= 0; a = (a + 0x6d2b79f5) | 0;
+    a |= 0;
+    a = (a + 0x6d2b79f5) | 0;
     let t = Math.imul(a ^ (a >>> 15), 1 | a);
     t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
     return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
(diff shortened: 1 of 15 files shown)


FernandoX7 (commented):
Nice work on this. The write-up names the actual failure precisely (walls gentler than the climb limit, a direction-only slope gate a diagonal switchback beats, and airborne movement skipping the gate entirely), and the fix closes each hole rather than papering over the symptom. Things I checked and liked:

- The true-steepness gate is the right shape: gating on the terrain gradient magnitude at the destination, not just rise/run along the step, is what makes the switchback defense hold. I confirmed it is load-bearing by neutering `terrainSteepnessAt` in a scratch tree: the switchback, slide, and no-jump tests all go red, so they genuinely reproduce the exploit.
- Determinism holds. The new helpers are pure functions of (x, z, seed), the memo cache is a pure memo (a hit returns the same float a miss computes, eviction only forces recompute), `tests/architecture.test.ts` stays green, and the packed cache key is collision-free for every reachable overworld cell (instances short-circuit on `x > DUNGEON_X_THRESHOLD`).
- The airborne branch is a genuine fix, not a cosmetic move: the base had a dead `if (!p.onGround)` nested inside `if (p.onGround ...)`, so airborne movement was never climb-gated.
- The terrain contract and content safety are pinned at the production seed, and cell-rounding does not force a slide onto legit content (worst rounded-cell steepness across camps, npcs, hubs, and road vertices is 0.80, well under 1.5).
- tsc, the re-recorded parity goldens, and both new suites are green.

## should-fix

- Verification wording, "rng draw counts are unchanged": this is not accurate. `draws` and `drawDigest` moved in about nine goldens (arena_1v1 479 to 482, nythraxis_full_pull 3299 to 3302, mob_locomotion 15 to 16, and others). It is not a determinism regression, the new helpers use the stateless `hash2`/`fbm2` and add no draw site, so the shifts are downstream position-driven conditional draws (taller terrain plus wall-blocked mobs change proximity-gated aggro/wander/target draws) and parity is green. The recordings are correct; just fix the note to say the draw order and counts shift downstream and the digests were re-recorded, so a future reader is not told the change is draw-neutral when nine digests moved.

- `tests/terrain_walls.test.ts`: the pass-shoulder band is unpinned. The ridge test skips `|x| < 36` (`:38`) and the pass test only covers `|x| <= 8` (`:74`), but the wall ramps from the flat opening to a full wall across `|x|` 10 to 34 (`PASS_SHOULDER`, `src/sim/world.ts:47`). I scanned that band at seed 20061: the crossable zone is contiguous with the road pass and stops at `|x|` about 14.5, with nothing crossable beyond `|x| > 15`, so there is no bypass today. But since the whole point here is that there is exactly one crossing per ridge, pinning the shoulder (assert `|x|` in [16, 34] beats `WALL_MARGIN`, or pin the intended passable half-width) keeps a later terrain tweak from quietly widening that gap into a real cross-beside-the-pass hole.

- `tests/climb_slope.test.ts:17` and `tests/terrain_walls.test.ts:11`: `CLIMB_LIMIT = 1.5` is a bare literal, not tied to the source `PLAYER_MAX_CLIMB_SLOPE` (`src/sim/pathfind.ts:37`). It is a real independent pin today (not the self-comparison trap), but the inverse risk is live: change the source constant and the walls could become climbable while the tests still assert against 1.5. One `expect(PLAYER_MAX_CLIMB_SLOPE).toBe(1.5)` in each file fails loudly if they ever desync.

## nits

- `src/sim/sim.ts:2856-2889`: `steepGround` gates on the cell-rounded `terrainSteepnessAt`, but the slide uses `terrainDownhill` on the raw position. If a cell reads steep while the raw-position gradient is about zero (`terrainDownhill` returns null), the player gets no control, cannot jump, and does not slide. It is unreachable on the production walls (monotone gradients), but a one-line guard (fall back to normal control when `steepGround && slide === null`) removes the theoretical soft-lock.
- The charge (`sim.ts:3290`), follow/trail (`:2805`), knockback, and mob/pet/fear `moveToward` (`:3998`) gates get the same new OR-clause but no direct test; they are only covered indirectly through the re-recorded goldens, and a golden delta can reflect the terrain-height change rather than the gate. One targeted case each (charge/knockback into a wall stops at the base; a mob pathing across a ridge band does not climb) would lock them.
- `src/sim/world.ts:199`: `STEEPNESS_CACHE_MAX` (400k) sits about 2.5% above the overworld cell count, and the more-than-four-seeds path does a full `steepnessCache.clear()`, so the bench loop that cycles seeds thrashes. Values stay correct (pure recompute); worth a larger cap or a per-seed LRU only if profiling shows churn.
- `tests/climb_slope.test.ts` runs at seed 42 while the terrain contract is pinned at the production seed 20061. The gates are seed-agnostic so this is fine; a one-line note on why 42 closes the question.
- The diff carries a fair amount of unrelated biome reformatting (import reorder, comma-per-line in `roadDistance`/`generateDecorations`/the crater helper). Compliant with the touched-files rule, just diff noise.

One behavior worth a conscious confirm: the slide is computed independent of `isRooted`, so a rooted player on unwalkable ground still slides off. That reads as intentional (you should not be able to root-anchor on a wall) and cannot fire on legit slopes, flagging only so it is a deliberate choice.

EnriqueGF:
@FernandoX7 thanks for the thorough review. The three should-fix items are addressed in 03858d3c:

- Verification wording: the PR description now says the rng draw order and counts shift downstream (position-driven conditional aggro/wander/target draws from the taller terrain and wall-blocked mobs) and that the digests were re-recorded, instead of claiming the change is draw-neutral.
- Pass shoulder band: new case in tests/terrain_walls.test.ts pinning that every ridge crossing at |x| 16 to 34, both sides, beats WALL_MARGIN. I re-scanned the band at the production seed and it matches your numbers: crossable ends around |x| 14.5, contiguous with the road pass, and the worst case at 16 is 1.73.
- Climb-limit pin: both test files keep CLIMB_LIMIT as an independent literal (per your note about the self-comparison trap) and now assert PLAYER_MAX_CLIMB_SLOPE equals it, so a source change fails loudly.

Also took the free nit: a note in climb_slope.test.ts on why it runs at seed 42 while the terrain contract is pinned at 20061. The remaining nits (the theoretical steep-but-flat soft-lock guard, direct tests for the charge/knockback/moveToward gates, the cache cap) I left out for now; happy to follow up if you want any of them in this PR.

On the rooted-slide question: yes, intentional. Root should not let you anchor on a wall, and the slide cannot fire on legit walkable slopes.

PR #1327, opened 2026-07-02
feat(delves): The Drowned Litany, second delve (marsh ruin, levels 12-14)

MasterZensei:
## Summary

Adds **The Drowned Litany**, the second delve (index 1), a flooded marsh-ruin shrine beneath Fenbridge for levels 12-14 (2 players suggested). Same structure as the Collapsed Reliquary (procedural module run, kill-boss objective, Delve Marks economy) but with its own room puzzles, a static environmental hazard, a fully scripted boss, and a puzzle finale that replaces the lockpick chest.

Built in phases on this branch (30 commits): irregular room shells -> the Blackwater hazard -> enemy kits -> room puzzles -> boss mechanics -> the Drowned Reliquary Rite finale -> dedicated creature/prop art -> this PR's cleanup pass (bell ropes become deliberate pulls, per-blocker exit hints, an affix theme fix, and a slimmed egg-sac model).

## Structure

A run rolls 3 of 6 trash modules, then appends the boss apse:

| Module | Puzzle | Notes |
|---|---|---|
| The Crescent Sluice | 2 sluice valves (walk-on) | Cantor + spearjaw pack screening the valves |
| The Island Ledger | 4 grave tablets (walk-on, any order) | Ranged acolytes + widowling control adds |
| The Ring Reliquary | 2 corpse candles (walk-on) | Cantor + skirmishers guarding an elite sump troll |
| The Sinkhole Baptistry | 3 spider egg-sacs (destroy) | Egg-sacs only appear after 3 waves of trash clear |
| The Reedsong Gallery (choir loft) | 2 bell ropes (deliberate F-pull) | Fan of cantors/thralls; pulling a rope shocks living cantors |
| The Y-Split Causeway | none | Mobile skirmishers + an elite over open water |
| The Drowned Apse (finale) | boss + Drowned Reliquary Rite | Sister Nhalia, the Drowned Canticle |

Brother Halven's marsh camp is the delve board NPC; Edda Reedhand is the auto-assigned companion who escorts the party through the shrine.

## Enemies and attack styles

| Enemy | Role | Attack style |
|---|---|---|
| Drowned Cantor | priority caster | Ranged Drowned Dirge (shadow bolt) at range, channels Litany Pulse to heal nearby drowned allies, kill first |
| Reedbound Acolyte | ranged | Hurls Rotwater Vial projectiles from the back of the pack |
| Deepfen Spearjaw | mobile skirmisher | Fastest marsh mob; builds a stacking Frenzy the longer it stays on a target |
| Mirefen Widowling | control add | Bite lands Web Snare, slowing the target so the pack can close |
| Bog Thrall | swarm | Cheap HP; a Bog Thrall's death whips nearby thralls into a brief pack Frenzy |
| Grave-Silt Bulwark | elite anchor | Slow frontal Silt Cleave; periodically wards nearby allies with a Nature absorb shield |
| Sump Troll Devourer | elite bruiser | Periodic Silt Hide (stoneskin) plus a ground-stomp AoE (Sump Stomp) |
| Spider Egg-Sac | finale trigger | 1 HP, stationary, no attack; one hit kills it, but it bursts for a small Nature AoE tick and hatches 2 Mirefen Widowlings |

## Damage pools and hazards

**Blackwater (static, room-wide).** Every room hazard is drawn from shallow and deep Blackwater pools baked into the module geometry. Standing in one deals a 1-second-interval tick of a player's max HP: 4% Normal / 8% Heroic in shallow water at 0.35x, or the full 1x-2x multiplier in deep water (deep hits for roughly 2x the shallow rate). Islands and the boss dais are dry ground and exempt; airborne (jumping) players dodge the tick entirely. The Heroic-only `high_water` affix multiplies every Blackwater tick by 1.35x.

**Sister Nhalia's Blackwater Mark (boss).** Every 14 seconds the boss marks a random party member; an 8-second, 4-yard puddle follows that mark and ticks Nature damage once per second (5% max HP Normal / 8% Heroic, also scaled 1.35x by `high_water`). Don't stack marks on top of each other.

**Tolling Bells (boss).** Every 10-12 seconds Sister Nhalia rings a volley of 4 bells that fly outward from the altar in four evenly spaced, randomly rotated directions, straight through the apse walls. Each bell travels at 8 yd/s and knocks back + deals 12% max HP Nature damage to anyone it clips (2-yard contact radius).

## Boss: Sister Nhalia, the Drowned Canticle

- **Blackwater Mark**: see above, the boss's core damage-pool mechanic.
- **Cantor phases**: at 70% and 35% HP, Sister Nhalia summons 2 Drowned Cantors (plus 2 Bog Thralls with the Heroic `lively_choir` affix) and shields herself with a 120-point Nature absorb until every summoned cantor is dead.
- **Tolling Bells**: a continuous volley mechanic that runs throughout the fight, independent of the cantor-shield state.
- **Final Bell**: at 10% HP, a one-time burst that hits everyone for 22% max HP (holy), stuns for 1.5s, and drops 4 Bog Thralls (6 with `lively_choir`) as a last stand.

## Finale: the Drowned Reliquary Rite

Replaces the lockpick chest with a seeded shrine-sequence puzzle. The party picks a difficulty before the shrine lights up:

| Difficulty | Sequence length | Playbacks shown | Tries | Loot ceiling |
|---|---|---|---|---|
| Easy | 4 | 3 | 3 | Low |
| Medium | 5 | 2 | 2 | Medium |
| Hard | 6 | 1 | 1 | Premium |

A wrong touch fails the current try (3% max HP Nature backlash) and replays the sequence; a mistake-free clear on any difficulty caps at Premium loot, more mistakes fall back to Medium or Low. Reward tiers pull from a per-class-archetype (Warrior/Paladin/Shaman, Rogue/Hunter, Mage/Priest/Warlock/Druid) item pool: guaranteed uncommon, a chance at a rare, and a 3% shot at an epic on the bountiful/Heroic path.

## Rewards

- **Delve Marks**: pays double the Collapsed Reliquary's rate (2 Normal / 4 Heroic for the first 3 clears/day, tapering after), reflecting the deeper currency-curve step.
- **Copper + XP**: 1250 XP / 18-30c first clear (Normal), 1850 XP / 32-48c (Heroic, 1.3x reward multiplier).
- **Gear shop** (Brother Halven, gated by delve clears): utility pieces and class chests available immediately (16-20 Marks), a helm after 3 clears (24 Marks), and two Heroic-only signature rares, Sister Nhalia's Choir Plate and the Drowned Choir Fang (56 Marks each), a straight 2x of the equivalent Collapsed Reliquary price.
- **Heroic affixes**: `high_water` (+35% Blackwater/Blackwater Mark damage), `lively_choir` (extra Bog Thralls on cantor phases and Final Bell), `belligerent_dead` (scoped to the `ruin` theme only in this PR: it buffs Grave-Silt Bulwark, a Litany-only mob, so it no longer rolls into the Collapsed Reliquary's `crypt` theme where it would silently do nothing).

## What's new in this PR specifically

On top of the full delve build already on this branch, this change:

- Makes the Sinkhole Baptistry's spider egg-sacs real 1 HP combat-target mobs (spawned after 3 waves clear) instead of ground interactables, so they die to any single hit, burst for AoE, and hatch adds like every other mob.
- Turns choir-loft bell ropes into deliberate F-pulls instead of walk-on triggers, so players don't accidentally shock the room while passing through.
- Adds per-blocker exit hints ("destroy the spider sacs" / "pull the bell ropes" / "apply pressure somewhere in the room") instead of one generic "sealed" message.
- Fixes `belligerent_dead` rolling into crypt-theme delves where its hook mob doesn't exist.
- Re-optimizes `spider_egg_sac.glb` (3.96 MB -> 0.32 MB).

## Screenshots

**The Crescent Sluice**: shallow/deep Blackwater rings (the damage-pool hazard) around a Drowned Cantor + Reedbound Acolyte pack, Edda Reedhand escorting:

Crescent Sluice

**The Sinkhole Baptistry**: a Mirefen Widowling swarm around the flooded pit before the egg-sacs unlock:

Sinkhole Baptistry

**Spider Egg-Sac**: the reworked 1 HP combat target, post-fight, with its own nameplate/health bar and the "baptistry falls quiet" cue in the log:

Spider Egg-Sac

**The Reedsong Gallery**: the bell-rope room; Bog Thralls and Drowned Cantors fan out from the ropes:

Reedsong Gallery

**Sister Nhalia, mid Tolling Bells volley**: two of the four bells visible flying outward from the altar, real combat damage taken (570/584 HP):

Boss fight

## Test plan

- [x] `npx vitest run tests/architecture.test.ts` (sim purity/determinism)
- [x] `npx vitest run tests/localization_fixes.test.ts` (S3 i18n matcher guard)
- [x] `npx vitest run tests/delves.test.ts` (room puzzles, egg-sac combat, wave gating, affix exclusion)
- [x] `npx vitest run tests/parity` (cross-host determinism golden, including the new `drowned_litany` scenario)
- [x] `npx vitest run tests/guide.test.ts` (wiki content freshness)
- [x] Headless-browser verification: all three egg-sacs spawn and render with live views after clearing the three baptistry waves; screenshots above captured the same way
- [ ] `npm run ci:changed` (Biome, changed files only): currently reports pure CRLF-vs-LF formatting diffs from this Windows checkout's `core.autocrlf=true`; the committed blobs are LF-clean and CI runs on Linux, so this is not expected to fail there

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


FernandoX7 (commented):
This is a strong content PR, and the things that usually go wrong at this size went right. I verified the wholesale golden re-mint is pure entity-id renumbering (zero rng draw or digest changes across every re-minted golden; the one new board NPC shifts every id by +1). The rite, affix, and module rolls run on isolated seeded streams, the loot roll keeps a fixed two-draw discipline, and the new tickers early-return for non-litany runs, so the shared draw order is untouched. Everything resolves sim-side (rite outcome, loot rolls, egg-sac bursts, bell contact) with `delve_rite_choose` enum-validated at the server boundary. The 2x Marks and shop economy matches the Collapsed Reliquary numbers exactly, index.html and play.html additions are byte-identical, the i18n artifacts regenerate byte-identically, the S3 drift guard was proactively extended over the three new emit files, and the new code lands as proper sibling modules (rite_window, delve_marsh_dressing, the boss/rite/rooms sim modules) instead of growing the coordinators. Pool and bell visibility is not tier-gated, so the fairness invariant holds. Nice Windows fix on the browserslist parser too.

Findings, most severe first.

**Blocking**

1. `tests/world_api_parity.test.ts:335` - the suite is red at the PR head. `delveRiteChoose` was added to `IWORLD_MEMBERS`, both sorted lists, and `FACET_DELVES`, but the count pins were not bumped: 149 to 150 at :335 and :1012-:1013, 113 to 114 at :337, plus the two "149-member" test titles. `npx vitest run tests/world_api_parity.test.ts` fails 2 tests with "expected 150 to be 149".

**Should-fix: gameplay**

2. `src/sim/delves/drowned_litany_rite.ts:212` - `openDrownedReliquary` never sets `state.lootOwnerId`, so any party member in range can front-run `collect_delve_chest_loot` and take a roll that was tuned to the opener's class. The lockpick path sets it for exactly this reason (`lockpick_controller.ts:362`).

3. `src/sim/delves/drowned_litany_boss.ts:462` - the bell cleanup on boss death is unreachable: `findNhaliaBoss` (:83) only returns a living boss, so after she dies the driver exits at `if (!boss) return` (:448) and never reaches the `boss.dead` branch. In-flight bells stop ticking and linger frozen in the apse through the whole rite until run teardown.

4. `src/sim/delves/companion.ts:67` - the rank-3 revive recharges by leaving and re-entering mid-run: `reviveUsed` lives on `run.companion`, which `despawnDelveCompanion` (sim.ts:5681) destroys and re-entry re-mints. Move it onto the run next to `companionBarks`, which already survives re-entry, and initialize it at spawn while you are there.

5. `src/sim/delves/drowned_litany_rite.ts:240` - `delveRitePulse` broadcasts realm-wide: it carries no `pid`, and `shrineId` is not a field `eventAnchor` (server/game.ts:3445) recognizes, so it falls through to the broadcast path and hud.ts:6967 closes the rite popup of any player who has one open, in any run. Carrying the shrine as `entityId` makes it interest-scope like other world fx; `delveRiteChoosePrompt` and `delveRiteFeedback` already scope correctly via pid.

**Should-fix: cross-content regression**

6. `src/render/characters/manifest.ts:970` - renaming the `sister_nhalia` MOB_KEYS entry to `sister_nhalia_drowned_canticle` orphans the zone 2 overworld rare elite, which still spawns as `sister_nhalia` (zone2.ts:1163) and now falls through to `FAMILY_KEYS.humanoid`, rendering as a generic bandit. Add the old mapping back alongside the new boss key.

**Should-fix: readability**

7. `src/render/delve_marsh_dressing.ts:391` - `placeMarshBlackwaterPools` draws every hazard as a full uncut disc while the sim exempts island rects and the dais, so in the apse the r=21 deep pool paints the two unmarked dry islands as lethal near-opaque water. Safe ground that reads lethal fights the stepping-stone design this delve is built on.

8. `src/ui/delve_map.ts:146` - in the litany schematic branch, island widths and hazard radii are sized with `scale` (in practice the z scale) while positions and the walkable outline map x through the `polyMaxAbsX` scale, so pools and islands draw at roughly 0.25x to 0.54x of the width the outline implies, across all 7 modules.

9. `src/ui/rite_window.ts:30` - `OPTIONS` hand-duplicates the sim's `RITE_INTENSITY` tuning (playbacks, symbols, tries) with no shared source and no test pinning them together; the lockpick precedent this module cites does share its source with the panel core. Export the tuning or pin it, so a sim balance change cannot silently desync the popup.

**Should-fix: i18n wiring** (matcher mechanics, not translations)

10. `src/ui/sim_i18n.ts:2826` - the cantor's heal is named `Litany Pulse` in content (mobs.ts:165) but `AURA_NAME_KEY` registers `Litany Mend`, so the channel line leaks raw English in every non-English locale and `mechanic.litanyMend` is dead. Same family: `Web Snare` (mobs.ts:232) has no `AURA_NAME_KEY` entry at all, unlike its precedent `Numbing Chill`, so the player's own slow debuff renders raw English in the debuff frame.

**Should-fix: CI and hygiene**

11. `npm run ci:changed` fails on this branch for real, not via CRLF: on an LF checkout Biome reports format diffs in `src/ui/i18n.locales/ja_JP.ts`, `ko_KR.ts`, `ru_RU.ts`, and the scratch test below. `npx @biomejs/biome check --write` on the three overlay files clears it.

12. `tests/door_probe.scratch.test.ts:1` - a committed scratch scan whose own header says "SCRATCH (not for commit)": console.log output, no assertions. Remove it, which also clears its format error.

**Should-fix: test gaps**

13. The Blackwater exemptions this branch fixed as bugs have no test: nothing sets `p.jumping` inside a hazard and asserts no tick (runs.ts:1118), and nothing stands a player on an island or the dais and asserts zero damage (runs.ts:1121). The deep/shallow tiers are also only asserted as "some damage"; a shallow-only point vs a deep point pins the ratio. Given the test-first bug-fix rule these are the tests I would most want before merge.

**Nits**

- `src/sim/delves/drowned_litany_rooms.ts:149`: hatchling widowlings spawn at `tmpl.minLevel` with no Heroic level bonus, 3 levels under the rest of the room.
- `src/sim/content/delves/mobs.ts:241`: a 1 HP egg-sac pays full level 12 (15 Heroic) kill XP for one hit; bounded to 3 per run, but flag if unintended.
- `server/game.ts:2825`: `delve_rite_choose` has no proximity gate, unlike `delve_buy` and the companion upgrade; any member anywhere in the run can commit the shared difficulty without approaching the reliquary.
- `src/ui/hud.ts:5555`: the rite popup is not in `isModalOpen()` and has no teardown when the run ends while it is open, so it can float over the outdoor world after a death or walk-out.
- `src/ui/hud.ts:6993`: `companionBark` resolves the speaker from the mutable `companionState`; a bark processed while it is momentarily null falls back to Tessa's name and lines during an Edda run.
- `src/ui/sim_i18n.ts:2828`: `Silt Hide` is aliased to `mechanic.siltWard`, so it renders as "Silt Ward" even in English and becomes indistinguishable from the bulwark's real Silt Ward aura.
- Dead i18n entries added by this PR, all carrying overlay fills: `delveUi.board.openDelveLitany`, `delveRiteUi.mistakes`, the `delveUi.object.*` labels (those objects never render nameplates, see nameplate_view.ts:98), `sim.delve.mobChannels` (shadowed by the sim-scope key the matcher uses), and the `Rotwater Hex` AURA_NAME_KEY entry (nothing emits it). Wire them or drop them.
- `src/render/renderer.ts:4062`: a `builtTemplateId` swap seen from the 80 to 96 yd hysteresis band removes the view, but re-creation only happens inside 80 yd, so the object vanishes until you close in.
- `src/render/delve_marsh_dressing.ts:347`: the `bone_pile` anchor maps to `addShrineFragment`, so apse bone piles render as glowing green shrine slabs; and :251 adds a beam length to an angle (`ropeAngle = rot + beamLen / 2`).
- Test polish: tolling_bells.test.ts:226 says "at the configured speed" but asserts only `dist > 0.1` (pin 0.4 per tick); :260 asserts damage happened but not the 12 percent; litany_spawn_collision.test.ts:79 hardcodes the shrine offsets instead of reading the spawned entities; delves.test.ts:1522's clear-ground arm can be masked by regen. The DROWNED_LITANY_SHOP price ladder and clear gates are unpinned (the reliquary tests cover the shop logic, so this is a data pin only).
- `src/sim/delves/drowned_litany_rite.ts:178`: `chooseDrownedLitanyRiteIntensity` stores the raw string before validating it is a known key; today's callers validate, but a future one crashes at `riteCeiling`.

For the record, two things I checked and dismissed: mob-sourced ability names (`Rotwater Vial`, `Drowned Dirge`) never render in any combat-log sink (all are player-source gated), and the four puzzle interactables never render nameplates, so neither leaks English at runtime.

FernandoX7:
I pushed five commits to this branch with fixes for the review findings, plus a few things a second pass over the fix diff itself turned up. Summary of what changed; the commit messages carry the detail.

- `test(delves)`: the world_api_parity pins (149 to 150, 113 to 114), and two more stale pins in `tests/command_schema.test.ts` that my review missed (104 to 105 sends, 113 to 114 dispatch cases): that file was also red at the branch head. The scratch door probe is removed.
- `fix(delves)`: rite loot is owner-locked like the lockpick chest; the bell cleanup on boss death was unreachable (`findNhaliaBoss` only returns a living boss) and now runs the tick after the kill; the rank 3 revive no longer recharges on leave and re-enter; `delveRiteChoose` is geo-gated at the reliquary (at `DELVE_INTERACT_RANGE`, the radius the prompt opens at, so anyone who got the popup can answer it) and rejects unknown intensities; hatchlings carry the Heroic level bonus; an egg-sac kill pays no XP (new `xpMult` template knob, the hatched widowlings pay normally); `delveRitePulse` is interest-scoped via `entityId` instead of broadcasting realm-wide; `companionBark` carries its speaker. Each fix has a covering test. The six re-minted goldens changed only in event hashes, zero rng draw or digest changes.
- `fix(render)`: the zone 2 Sister Nhalia rare renders as the dark caster again (the MOB_KEYS rename had orphaned her to the generic bandit model); the authored dry islands draw as low platforms above the Blackwater pools so safe ground reads safe; apse bone piles are bones instead of glowing shrine slabs; a template swap seen from the 80 to 96 yd band no longer vanishes the object; the marsh puzzle objects (valves, tablets, candles, ropes, and their spent variants) show their localized nameplate labels like the rite shrines do.
- `fix(ui)`: the rite popup closes when the run ends or the rite moves past the choose phase; its Easy/Medium/Hard numbers derive from the sim's own tuning table (new shared `rite_tuning.ts`, mirroring how `lockpick_panel` shares `src/sim/lockpick.ts`); the delve map draws pools and islands at true per-axis scale (they were at roughly a quarter to half the implied width), as ellipses where the mapping is anisotropic, clipped to the room outline, with islands painted on top of the pool fills.
- `fix(i18n)`: `Litany Pulse` and `Web Snare` are registered in the aura matcher (they leaked raw English in non-English locales), `Silt Hide` gets its own key (it displayed as "Silt Ward" even in English), dead keys and the unused `Rotwater Hex` row are dropped, and the resolved tables are regenerated.

One judgment call to flag: the apse moat authors `rx`/`rz` ellipse fields, but `tickDelveBlackwater` uses the circular `r` and drops them, so the map and the scene both draw the circle the sim actually damages with. If the ellipse was the intent, the sim check needs to honor `rx`/`rz` first; happy to leave that call to you.

On the branch tip: the full suite (6788 tests), `tsc --noEmit`, and all four builds are green, and `biome ci` is clean on the changed files. For what it is worth, the earlier `ci:changed` failures were real format diffs in the three edited locale overlays, not the CRLF issue: they reproduce on an LF checkout, and the format pass here clears them.

Really solid delve. The parity discipline (isolated seed streams, the fixed two-draw loot roll, the id-renumber-only golden re-mint) made this a pleasure to verify.

FernandoX7:
I ran a second, deeper verification pass over my five fix commits (fresh reviewers over the whole fix diff, every original finding re-checked as genuinely closed, plus targeted mutation testing on the new pins). The fixes themselves held up: all 18 original findings verified closed, the golden re-mint claim confirmed byte-level (only event hashes changed, rng draws and digests identical), and the wire scoping, owner lock, revive, and bell cleanup all traced end to end. The pass did surface a few more things, pushed as four commits:

- `fix(i18n)`: the Deepfen Spearjaw's `frenzyOnHit` buff ("Frenzy") pushes a named, target-frame-visible aura but had no matcher row, so it rendered raw English in every non-English locale: the same drift class as Web Snare, one instance further than my earlier fix caught. Registered with fills, plus a new guard test that scans every delve mob's aura-emitting proc fields (mendAlly, chillOnHit, stoneskin, wardAllies, frenzyOnHit) and asserts each name resolves through the matcher, so this whole bug class now reddens the suite instead of shipping.
- `fix(delves)`: the hatchling and egg-sac Heroic level bonus was hardcoded (`tierId === 'heroic' ? 3 : 0`, two sites) instead of reading `tier.enemyLevelBonus` from the delve definition; now derived, with a Normal-tier test pinning hatched level 12 so an unconditional bonus fails.
- `fix(render)`: the template-swap fix had a remaining race: a swap first seen inside 80yd, followed by retreating past 80yd before the budgeted create pass rebuilt it, still vanished the object for the whole hysteresis band. Swaps now rebuild the view in place, which also shows the correct (not stale) template in-band.
- `test`: decisiveness gaps closed. The owner-lock test now actually attempts the front-run with a partied rival (refusal error asserted, loot untouched, opener collects; a silent early-return can no longer fake a pass). The delve map gets numeric extent pins (pool/island sizes must equal their position-mapped world edges; an axis swap or a single-scale regression fails). The apse dais exemption gets a dais-only annulus point (its centre was coincidentally covered by an island, shadowing the branch). The marsh puzzle nameplate allowlist is pinned per template, fresh and spent, near and far. A server-side test pins that an entity-anchored `delveRitePulse` reaches a near session and not a far one. Stale prose counts in the command-schema header are gone. Each new behavioral pin was mutation-tested (guard removed, test reddens; restored, green).

Verified on the tip: the targeted suites (delves, map, companion, shop, bells, spawn-collision, nameplates, snapshots, localization, command schema, IWorld parity, parity goldens, architecture, i18n completeness, guide), `tsc --noEmit`, and biome on the changed files.

Two pre-existing issues surfaced that I am deliberately NOT fixing in this PR (they predate it; filing separately): delve map live markers ignore the module z offset (markers drift south in module 0 and land off-canvas in later modules, reproduces on the Collapsed Reliquary at base), and the IWorld parity gate misses the three daily-rewards members. Same for the pre-existing unmatched aura names on first-delve mobs (Ledger Rot, Grave Hex, Blood Frenzy on the zone 1 boar): same leak class, out of scope here, will handle with the release fill pass.

One more judgment call for you alongside the apse `rx`/`rz` one: on the out-of-tries arm, `openDrownedReliquary` assigns `lootOwnerId` to whoever made the FINAL mistake, so in a party the last fumbler gets exclusive rights to the consolation chest (and if they then leave, the loot strands until teardown, same as the lockpick precedent). If you would rather have the fail-arm chest unowned, or owned by the player who committed the difficulty, say the word and I will adjust.

FernandoX7 (commented):
Merged the latest `release/v0.20.0` into the branch (it advanced three times while I was verifying, so this is three merge commits) and it is green and mergeable again. Beyond the usual regenerated i18n trio and parity goldens, the real conflicts were the IWorld / command count pins (now 167 members, send 116 / dispatch 125), the `world_entity_i18n` + `manifest` + `data` NPC unions (`brother_halven_marsh` alongside the new `spirit_healer`), the spirit-healer interact branch, and the overworld sight line, which now takes release's eye-height `sightBlockedAt` (the Litany delve LOS branch above it returns early and is untouched). Regenerated the i18n artifacts and re-minted the conflicted goldens; the delve golden churn is the new `bags` meta field, not behavior drift.

Deep re-read of everything new since the last review. The delve fixes hold up: the `partyMembersForKey` presence gate, the `affixSpawned` restless-graves loop stop, the per-member spawn spread, and the stowed-pet retention are all correct and determinism-safe (the parity draw digests are unchanged, only event digests moved). I applied a few things directly to the branch:

- Dropped the two `[delve]` `console.log` traces in `enterDelve` / `claimDelveRun` (`src/sim/delves/runs.ts`). `src/sim/` runs in the headless RL env too, where stray stdout interleaves with the NDJSON protocol.
- Removed the now-dead `aura.frenzy` matcher rows in `src/ui/sim_i18n.ts`: the Feeding Frenzy rename orphaned the key an earlier slice had added solely for the Deepfen Spearjaw.
- Made the four live-pet command errors (`petAttack` / `petTaunt` / `feedPet` / `healPet`) delve-aware too, so they match the dead-pet-lookup sites instead of saying "You have no living pet." while the pet is stowed for the run.
- Hardened three tests: the `affixSpawned` write half now pins via the real raise path (not a hand-set flag), the tracker-stack containment test depth-walks to the wrapper's own close (a moved-out delve tracker now reddens it), and added the overworld no-pet arm plus a pin for the helper-returned pet strings the S3 emit scanner is structurally blind to.

Two pre-existing items worth a follow-up (neither is introduced here, so not blocking this PR):

- should-fix: `partyMembersForKey` gates on delve-band position, not run membership (`src/sim/sim.ts`). The instance key is party-scoped, so a duo that enters two different delves at once shares one key, and each run's teleports and rewards would then sweep in the member who is actually in the other run. Your gate fixed the AFK case; scoping to the actual run would close the rest.
- should-fix: summon has no delve gate (`src/sim/combat/effect_dispatch.ts`, `summonPet` / `summonDemon`), so a warlock or hunter can summon a fresh pet inside a delve even though entry stows the pet and the new copy asserts "Pets are not allowed inside the delves." The new messaging only makes the old hole visible.

One nit: `#right-tracker-stack` (`src/styles/hud.css`) has no max-height, so a very long quest list can push the delve tracker below the viewport on a short or zoomed screen. The one-click tracker collapse mitigates it and it is strictly better than the old overlap, so it stays a nit; a `max-height` with `overflow-y` on the stack would close it.

Everything else verified clean: determinism (Rng draw order, no wall-clock, sim purity), IWorld and wire parity (`affixSpawned` is correctly server-local, not on the wire), the pet-error and Feeding Frenzy strings resolve through the matcher, and the render entrance / veil fixes read the world without mutating it. Full suite (7552), `tsc`, all four builds, changed-files Biome, and the malware gate are green on the merged tip.

PR #1330, opened 2026-07-02
feat(game): first-spawn intro cinematic for new characters

EnriqueGF:
## Problem

A newly created character is dropped straight into gameplay with no sense of place: the camera snaps to the gameplay pose and the HUD is already up. Classic MMOs open a new character with a short establishing shot of the world around the spawn before handing over control.

## Feature

On a character's first world entry, the camera opens 55 yd out across the field, high over the world, and glides in one continuous eased approach toward the character (a 0.3-turn sweep, not an orbit), landing exactly on the normal gameplay camera pose. The HUD and nameplates stay hidden until the camera lands, and the loading-screen fade doubles as the cut in, so there is no extra overlay.

- Plays exactly once per character: a localStorage seen marker keyed by the keybind scope (per online character id, per offline class + name).
- Only for new characters (level 1) and never for reduce-motion players, who go straight to gameplay.
- Deliberate skip only: Escape on desktop (eaten so it does not also open the game menu), or a rapid 4-tap burst inside 1.5 s on touch, where there is no Escape key. All other input is swallowed while it runs; a stray click or keypress does not cancel it.
- Works for both the online and offline entry paths. The camera path is pure math in src/game/spawn_cinematic.ts (no DOM), applied per frame between the follow-camera update and the renderer read, so it wins over mouse and follow input and hands off with zero snap. The renderer's existing terrain clamp keeps the long approach above ground.

## Verification

- tests/spawn_cinematic.test.ts pins the path contract (opening pose, exact landing on the gameplay pose, monotonic approach, per-frame continuity) and the tap-burst skip tracker (lone or slow taps never skip, a rapid burst does, the sliding window prunes).
- Browser end to end against the real dev client (offline quick start, headless Chrome): intro opens far out with the HUD hidden, a single click and a movement key do not skip, Escape skips without opening the menu, a 4-tap burst skips, an untouched run glides in and lands on the exact gameplay pose with the HUD restored, and a second entry with the same character does not replay it.
- tsc and the architecture guard are green on this base.

Note for the queue: the map-editor PR #1312 adds a custom-world parameter to startOffline; if both land, editor play-test entries should pass the intro flag as false so play-testing a map never triggers the cinematic (the map-editor branch already carries that form of the call).

### DIFF
--- a/src/game/spawn_cinematic.ts
+++ b/src/game/spawn_cinematic.ts
@@ -1,0 +1,58 @@
+// First-spawn camera cinematic, pure math (no DOM): the camera opens high and
+// far, one full turn behind the gameplay yaw, circles the spawn while it
+// descends, and lands exactly on the normal gameplay pose. main.ts feeds
+// elapsed seconds in and applies the returned pose to the input camera each
+// frame; tests/spawn_cinematic.test.ts locks the start/landing/continuity
+// contract.
+
+export interface CameraPose {
+  yaw: number;
+  pitch: number;
+  dist: number;
+}
+
+export interface SpawnCinematic {
+  durationSec: number;
+  turns: number; // full camera turns around the spawn
+  startDist: number; // the orbit opens far ...
+  startPitch: number; // ... and high, looking down at the spawn
+  end: CameraPose; // gameplay pose the cinematic lands on exactly
+}
+
+// startDist stays inside the wheel-zoom range input.ts allows (3..22) so the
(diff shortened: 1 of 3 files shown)


FernandoX7:
NOTE: We want to make this reusable so we can use it all over the game.

Will need heavy testing on web, desktop, and mobile, mobile apps, etc.

EnriqueGF:
For the moment I merge it, tested on mobile.

PR #1337, opened 2026-07-03
docs(frontier): Frostreach Frontier PvP + honor PRD and Phase 1 implementation handoff

ryan-foo:
## What this is

Design docs only, no implementation. Two documents:

- `docs/prd/frontier-pvp-honor.md`: the Frostreach Frontier, a Wintergrasp-style always-on open PvP zone. Two teams (Azure/Crimson) with visible back banners, automatic in-zone flagging, an extraction loop (harvest resource nodes into carried cargo that drops to the enemy on death and only banks at your team base), Honor Points + an Honor Quartermaster, rare spawns, a zone world boss, and a rotating hourly event system. Section 12 adds the phased $WOC stakes layer: a hard token firewall (no wallet/token code ever enters `src/sim/`), soulbound honor vs staked cargo, 2-week bracketed staked seasons, a sanctioned-automation agent server realm, and a headless economy wind tunnel that red-teams season parameters in CI before real value is at risk.
- `docs/prd/FRONTIER_PHASE1_HANDOFF.md`: the Phase 1 (skeleton) implementation handoff. Nine slices with hook points verified against `main` on 2026-07-03, per-slice acceptance commands, a complete inventory of new player-facing strings (the S3 i18n guard checks these), and gotchas.

## For anyone picking this up

- Read the handoff before starting a slice; it front-loads the traps. The biggest: `isDelvePos()` is open-ended along x, so the Frontier band sits at x 9000 and slice S1 must add an upper bound to the delve band (gotcha G1).
- Slice order: S1 -> S2 -> S3, then S4/S5/S6 in parallel, then S7/S8 (UI + render, taste-sensitive), then S9 (integration + reviewers).
- Each slice has explicit `npx vitest run ...` acceptance commands; anchor on those, not on looks-done.
- Phase 1 deliberately excludes the honor vendor (full-locale item i18n + exact level-20 stat budgets land with Phase 2 economy) and everything in PRD section 12.

## Status

Draft on purpose: this PR carries the spec so the work is documented and up for grabs. Comment here or just start a slice branch (`feature/frontier-p1-s<N>`) referencing this PR.

### DIFF
--- a/docs/CLAUDE.md
+++ b/docs/CLAUDE.md
@@ -29,6 +29,9 @@ expansion: story arc, zones, dungeons, XP math, ids); `design/spell-ranks.md` is
 classic-era ability-rank reference for sim ability content. **TRAP:**
 `design/icon-system.md` proposes a multi-file `src/ui/icons/` module, but the shipped
 code is the flat `src/ui/icons.ts`, so re-verify against code.
+`prd/frontier-pvp-honor.md` (Frostreach Frontier PvP zone, honor, $WOC stakes layer)
+pairs with `prd/FRONTIER_PHASE1_HANDOFF.md`, a slice-by-slice implementation handoff
+whose slices are specced but NOT yet implemented; read the handoff before starting one.
 
 ## screenshots/
 JPG/PNG assets embedded by the repo-root `README.md` (title screen, zones, dungeons, UI).
--- a/docs/prd/FRONTIER_PHASE1_HANDOFF.md
+++ b/docs/prd/FRONTIER_PHASE1_HANDOFF.md
@@ -1,0 +1,303 @@
+# Implementation Handoff: Frostreach Frontier, Phase 1 (Skeleton)
+
+| | |
+|---|---|
+| **Status** | Ready to implement (do not start slices without operator go-ahead) |
+| **Source PRD** | `docs/prd/frontier-pvp-honor.md`, section 11 Phase 1 |
+| **Scope** | Free loop skeleton ONLY: band, enter/leave via the G window, team assignment, auto-flagging, base graveyards, honor counter, honor on kills with DR, back banner, nameplate tint. NO cargo/nodes (Phase 2), NO events (Phase 3), NO $WOC (section 12, later). |
+| **Verified against** | `main` on 2026-07-03. Line numbers WILL drift: re-find every anchor before editing; trust the symbol, not the line. |
+| **Executor routing** | Slices S1 to S6: codex (`codex exec` via the codex-implementation skill), one self-contained prompt per slice, reviewed by Claude before merge. Slices S7 to S9: Claude (taste surfaces: UI, copy, render). |
+
+---
+
+## 0. Ground rules (paste into every codex prompt)
+
+1. `src/sim/` never imports from `render/`, `ui/`, `game/`, `net/`, or any DOM/Three
+   API. Guarded by `tests/architecture.test.ts`.
+2. All sim randomness goes through the sim's `Rng` (`ctx.rng` / `this.rng`). Never
+   `Math.random`, `Date.now`, `performance.now` in sim logic. Timers use sim time
+   (`this.time`, `DT = 1/20`).
+3. Every player-visible string is i18n: sim/server emit stable event data or
+   English literals that MUST have a matcher entry (`src/ui/sim_i18n.ts` /
+   `src/ui/server_i18n.ts`); UI strings are `t()` keys added to
+   `src/ui/i18n.catalog/` (English only, never edit `i18n.locales/` overlays).
+   The S3 guard is `tests/localization_fixes.test.ts`.
+4. TypeScript strict, ESM, 2-space indent, match surrounding style. No em dashes,
+   en dashes, or emojis anywhere, including comments and commit messages.
+5. Conventional Commits with scope, e.g. `feat(frontier): ...`.
+6. Anchor completion on the slice's acceptance commands, not on "looks done".
+
+## 1. Shared design constants (used by every slice)
+
+Defined once in S1; every other slice imports, never redefines.
+
+```ts
+// src/sim/data.ts (beside the arena/delve constants)
+export const FRONTIER_X_MIN = 9000;            // band start; see gotcha G1
+export const FRONTIER_ORIGIN = { x: 9200, z: 0 };
+export const FRONTIER_HALF_W = 200;            // playfield x half-extent
+export const FRONTIER_HALF_H = 300;            // playfield z half-extent
+export const FRONTIER_MIN_LEVEL = 15;
+export function isFrontierPos(x: number): boolean { return x >= FRONTIER_X_MIN; }
+
+// src/sim/content/frontier.ts (content module, merged by data.ts)
+export type FrontierTeam = 'azure' | 'crimson';
+export const FRONTIER_BASES: Record<FrontierTeam, { x: number; z: number; facing: number }> = {
+  azure:   { x: 9200, z: -280, facing: Math.PI / 2 },   // south base, faces north
+  crimson: { x: 9200, z:  280, facing: -Math.PI / 2 },  // north base, faces south
+};
+```
+
+Honor tuning (S5): kill base 20; level factor
+`max(0, 1 - max(0, killerLevel - victimLevel) / 5)`; per killer-victim-pair DR
+schedule 100% / 50% / 25% / 0%, window resets 3600 sim-seconds after the pair's
+first kill. Assist honor is OUT of Phase 1 (deferred to Phase 2 with the
+contributor tracking cargo needs anyway).
+
+## 2. Verified hook-point map
+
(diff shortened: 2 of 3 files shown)


EnriqueGF:
Picked up slice S1 (band, content module, persisted state). Branch: [`EnriqueGF:feature/frontier-p1-s1`](https://github.com/EnriqueGF/world-of-claudecraft/tree/feature/frontier-p1-s1), cut on top of this PR's docs commit, single commit `8234127d`.

What landed, per the S1 spec:

- `src/sim/data.ts`: the shared frontier constants (`FRONTIER_X_MIN` 9000, `FRONTIER_ORIGIN`, half extents, `FRONTIER_MIN_LEVEL`, `isFrontierPos`), and `isDelvePos` bounded above with `x < FRONTIER_X_MIN` (gotcha G1), leaving delve growth headroom through index 6.
- `src/sim/content/frontier.ts`: new content module with `FrontierTeam` and `FRONTIER_BASES`, re-exported from `data.ts`.
- `src/sim/colliders.ts`: `resolvePosition` gets a frontier branch before the delve branch clamping movers to the playfield rectangle (open flat field, no static colliders in Phase 1). I also added the matching early-return in `cameraOcclusion`: its contract says it mirrors `resolvePosition`'s region split, and without it a frontier camera would sweep phantom crypt colliders through the `x > DUNGEON_X_THRESHOLD` fallthrough. Say the word if you would rather defer that to S8.
- `CharacterState` / `PlayerMeta`: `frontierTeam?`, `honor`, `frontierReturnPos?`, backfilled in `addPlayer` (`honor` to 0, team stays unassigned) and round-tripped in `serializeCharacter`; pre-frontier saves load cleanly.
- `tests/frontier_band.test.ts`: band disjointness across [0, 12000] including the G1 boundary, the clamp on all four edges, bases inside the rectangle, the camera sweep, serialize/load round-trip, and absent-field back-compat (`honor === 0`).

One handoff gap found by the post-slice architecture review, now noted in the S2 section of `FRONTIER_PHASE1_HANDOFF.md` (included in the branch): `addPlayer`'s saved-pos ejection chain (`sim.ts` ~1148, delve then dungeon) has no frontier branch, so a character saved past x 9000 would eject to a dungeon door via the `?? DUNGEON_LIST[0]` fallback. Unreachable until S2's enter teleport lands, but S2 needs to pick the relog rule (keep the in-band pos for a team-holding character, or restore `frontierReturnPos`) in the same slice.

Acceptance, all green:

- `npx vitest run tests/frontier_band.test.ts tests/architecture.test.ts tests/entity_roster.test.ts` (10 + 24 + 15 tests)
- `npx vitest run tests/parity` (159 tests; goldens untouched, no rng draws added)
- extra insurance for the G1 bound: `tests/delves.test.ts`, `tests/arena.test.ts`, `tests/cooldown_persist.test.ts` (99 tests), plus `tsc --noEmit` and biome on the changed files

Happy to keep going with S2 (enter/leave/team assignment) on a follow-up branch once this direction is confirmed.

jgyy (commented):
This is a genuinely strong PRD pair. The PRD does the thing this repo's docs need most and rarely get: it resolves the "realm vs zone" framing question against actual repo constraints (`server/realm.ts` isolation, no cross-realm travel) before writing requirements, and section 3's "current state in the codebase" table is honest about what's reused vs new. The handoff is unusually careful for a docs-only PR: dependency-ordered slices, per-slice acceptance commands, a token-firewall invariant that gets its own `tests/architecture.test.ts` scan, and gotcha G1 (bounding the open-ended `isDelvePos`) is real and correctly diagnosed against `src/sim/data.ts` (`isDelvePos(x)` is genuinely `x >= DELVE_BAND_X_MIN` with no upper bound today).

Findings:

**should-fix: two companion docs in the PRD's header table don't exist.** `docs/prd/frontier-pvp-honor.md` line 10 lists `docs/prd/badges.md` (deterministic-currency precedent) and `docs/prd/heroic-mythic-dungeons.md` as companion docs. Neither exists anywhere in git history (checked `docs/prd/`, full-repo `git log --all`). Whoever picks up S1/S5 for the honor counter will go looking for the "deterministic-currency precedent" doc and hit a dead link. Either the intended precedent is just the `delveMarks` pattern already cited inline in section 3/7.1 (in which case drop the `badges.md` reference), or these were meant to reference something that hasn't landed yet and should say so explicitly.

**nit: the hook-point line numbers in the handoff are already stale against `release/v0.23.0`** (e.g. `CharacterState` is at `sim.ts:857` not `:780`, `toggleArena` is at `hud.ts:6691` not `:5268`, `dispatchMessage` is at `server/game.ts:2446` not `:1351`). This is explicitly caveated ("Line numbers WILL drift... trust the symbol, not the line") and matches the documented convention in `docs/CLAUDE.md` for PRD hook points, so it's not a defect, just worth flagging so whoever executes S1 doesn't waste time chasing an exact line before re-finding the symbol.

Everything else checks out: the `docs/CLAUDE.md` index update is a minimal, correctly-placed two-line addition that follows the existing pattern for `design/icon-system.md`. `docs/prd/woc/wallet-link.md` and `holder-cosmetic-flair.md`, the other two cross-references, do exist. `ARENA_X_MIN`/`DELVE_BAND_X_MIN`/`isArenaPos` in `src/sim/data.ts` match the PRD's description. No em dashes, en dashes, or emoji in either doc. Scope is well-bounded: Phase 1 explicitly excludes cargo/nodes/events/honor vendor/$WOC, and the PR itself is docs-only with no code, so there's nothing here that should touch CI.

One thing to watch for the eventual S1 implementer rather than this PR: honor as a new `CharacterState`/`PlayerMeta` field and `frontierTeam` are called out as additive JSONB with back-compat defaults, which is the right call per `migration-safety` conventions, good to see that spelled out this early.

Rough shape of the phase plan as I read it:

```mermaid
flowchart TD
    subgraph free["Free loop (ships independently)"]
        P1["Phase 1: Skeleton\nband, enter/leave, teams,\nauto-flag, honor+DR on kills"]
        P2["Phase 2: Economy\nnodes, cargo, death-drop,\nturn-in, Quartermaster"]
        P3["Phase 3: Events\nscheduler, 6 v1 events,\nteam score, HUD countdown"]
        P4["Phase 4: Apex\nworld boss, rares, cosmetics,\nzone map, wiki content"]
        P1 --> P2 --> P3 --> P4
    end
    subgraph stakes["$WOC stakes layer (section 12, later)"]
        P5["Phase 5: Wind tunnel\nseason config + CI\nexploit-agent harness"]
        P6["Phase 6: Staked season pilot\n2-week bracketed season,\nwallet deposit/settlement"]
        P7["Phase 7: Agent server\nsanctioned automation realm,\nagent-entered seasons"]
        P5 --> P6 --> P7
    end
    P4 -.gates.-> P5

    subgraph s1slices["Phase 1 slice graph (this handoff)"]
        S1["S1 band + content\n+ persisted state"]
        S2["S2 enter/leave,\nteam assignment"]
        S3["S3 hostility clause\n+ IWorld/wire/dispatch"]
        S4["S4 death/respawn"]
        S5["S5 honor + DR"]
        S6["S6 guide/wiki content"]
        S7["S7 PvP window UI"]
        S8["S8 banner + nameplate tint"]
        S9["S9 integration pass"]
        S1 --> S2 --> S3
        S3 --> S4
        S3 --> S5
        S3 --> S6
        S4 --> S7
        S5 --> S7
        S6 --> S7
        S7 --> S8 --> S9
    end

    P1 -.expands to.-> s1slices
```

Nice groundwork. The dead companion-doc links are the only thing I'd actually fix before someone starts S1.

ryan-foo:
Follow-up: sketched an FFA variant of this zone and folded it into the PRD as a new **section 15** (branch [`docs/frontier-ffa-variant`](https://github.com/ryan-foo/world-of-claudecraft/tree/docs/frontier-ffa-variant), off `release/v0.23.0`).

The idea: a sibling zone to Frostreach, a newly discovered frozen island with **no fixed teams**. Everyone is hostile by default, alliances are temporary breakable **pacts**, and betrayal is a first-class instant move (the double-cross at the evac beacon is supposed to work). It reuses every rail in sections 6 to 12 (cargo stake, token firewall, soulbound honor, brackets, agent server, wind tunnel) and only respecifies the social model, so it ships as its own season format rather than editing the war zone.

The one design point worth flagging: **ganging up and betrayal are intended content**, not a bug, and they self-police through greed (allies have opposing interests over who walks out with the haul). The only genuinely adversarial case is cooperative **wash-trading** (two wallets funnel value to a predetermined winner), which betrayal cannot police because the parties never actually fight over the loot, and which only matters once real $WOC is staked. So the free island ships with **zero anti-collusion machinery**; wash-trade defense is deferred to the staked-season wind tunnel (12.7), next to the smurfing scenario.

The one still-open problem is the shark/fish spiral in the *free* loop, where there are no deposit brackets to sort by (section 15.6). Not opening a PR yet; happy to if the variant is worth carrying forward.

PR #1354, opened 2026-07-03
feat(items): WoW-style bag system with tiered equippable bags

EnriqueGF:
## Summary

Replaces the infinite inventory with a classic WoW-style bag system, on the same pooled `InvSlot[]` model so the wire shape and JSONB saves stay compatible.

### The system
- **Backpack + 4 bag sockets.** Every character owns an implicit 16-slot backpack; four sockets take equippable `kind: 'bag'` items that raise the pooled slot budget (`src/sim/bags.ts` owns the capacity math and the `equipBag`/`unequipBag` command bodies behind `SimContext`).
- **Stack sizes.** Gear, bags, and tools take one slot per copy; consumables, junk, and quest drops stack to 20 (`stackSize` on `ItemDef`, default derived from `kind`). The `addItem` hub is now stack-aware.
- **Capacity gates at the command boundaries.** Vendor buy and buyback, ground pickups, corpse loot (what does not fit stays on the corpse), fishing (the catch gets away, rolled before the gate so rng draw order never depends on bag state), mage conjures, market buy/cancel/collect (items stay in the collection box), trade (both sides simulated before the atomic swap), quest turn-in (reward must fit after the collect items are handed in), and unequip. Paths a player cannot retry (need/greed wins, master loot, delve rewards) force-add instead: **items are never destroyed by capacity**.
- **Back-compat migration.** `CharacterState.bags` is optional (no DB migration, JSONB). A pre-bag save whose inventory grew past the 16-slot backpack under the old infinite inventory loads with bags already equipped to cover the space it had earned: the grant picks the lowest quality tier that can cover the need (a 30-slot save gets two common bags, never a free epic) and the fewest bags within it. It runs once (the next save writes the `bags` field), notices the player, and a hoard past the 72-slot ceiling keeps a tolerated overflow that only blocks new pickups until space is freed. Post-bag saves are never re-granted.

### Tiered bags
| Bag | Quality | Slots | Source |
|---|---|---|---|
| Linen Pouch | Common | 6 | Trader Wilkes (vendor) |
| Traveler's Knapsack | Common | 8 | Trader Wilkes (vendor) |
| Wolfhide Satchel | Uncommon | 10 | Forest wolves (2%), Old Greyjaw (35%) |
| Gravewoven Bag | Rare | 12 | Morthen (bonus roll group) |
| Mistcaller's Duffel | Epic | 14 | Vael the Mistcaller (bonus roll group) |

### UI
The bags window is reworked from a list into a WoW-style slot grid: square cells with the procedural item icon, quality-colored borders (uncommon and up; new `.q-legendary` rule), stack-count badges, and empty sockets that make free capacity visible. A new bag bar shows the backpack, the 4 sockets (click an equipped bag to remove it, click a bag item in the grid to equip it, equipping onto an occupied socket swaps), and a used/capacity counter that turns red on legacy overflow. Bag tooltips get the classic "N Slot Bag" line. New procedural `sack` icon primitive plus per-bag recipes.

### Cross-host wiring
`IWorldInventory` gains `bags`/`bagCapacity`/`equipBag`/`unequipBag`, implemented in both `Sim` and `ClientWorld`; new `equip_bag`/`unequip_bag` wire commands and a `bags` heavy-self delta key (registered in the snapshot + command-schema lockstep tests).

### i18n
5 new sim strings registered in `sim_i18n` with all 21 locales filled; new `hudChrome.bags.*` keys with the M16 non-Latin fills; item names for the 5 bags filled in every locale (catalog lists + overlays); `itemUi.kind.bag` + `itemUi.tooltip.bagSlots` across the `merge.ts` locale blocks.

### Tests
New `tests/bags.test.ts` (26 tests: pre-bag save migration incl. tier selection and idempotency, the item-instance (#1165) interaction, stacking math, capacity gates, equip/swap/unequip rules, vendor-buy refusal, loot-stays-on-corpse, persistence round-trip + pre-bag/tampered/over-capacity saves). Parity goldens re-minted deliberately (new loot-table draws + the `bags` meta field); snapshot/command/IWorld registry pins updated. Full suite, `tsc`, and `npm run build` green.

### In game: bag bar, slot grid, quality borders, epic bag tooltip
Bags window with 4 equipped bags and the epic duffel tooltip

### Vendor: common bags on sale at Trader Wilkes
Trader Wilkes selling the Linen Pouch and Traveler's Knapsack

### Close-up: the reworked bags window
Bags window close-up: bag bar, capacity counter, slot grid


EnriqueGF (commented):
Self-review pass over the full PR at 09bcb672, after merging the latest release/v0.20.0 twice (item instances #1165, round-robin loot, world boss, professions vendor tools). Every finding below was verified against the code and fixed in this branch; posting the trail so the review is auditable.

What was checked, domain by domain:

- Sim determinism: src/sim/bags.ts draws no rng (pure arithmetic; verified plus guarded by tests/architecture.test.ts). The fishing capacity gate sits after the table roll on purpose so draw order never depends on bag state (src/sim/sim.ts, completeFishing). The two golden re-mints are attributable to base content draws plus the new wolf loot rows, not to reordering; parity is green (159/159).
- Capacity gates cannot destroy items: corpse loot leaves un-fitting items on the corpse (interaction.ts), market collect keeps them in the box and zeroes gold before the early return so a retry never double-collects (market.ts marketCollect), trade simulates both sides on scratch copies before the atomic swap (social/trade.ts), quest turn-in simulates the collect hand-in before checking the reward fit (quests/quest_commands.ts). Roll wins, master loot, round-robin winners and delve rewards force-add on purpose: a player cannot free space on the winner's behalf, so those grants must always land.
- Item-instance interaction (#1165): addStacked and countFit never merge into an instanced slot and give it no top-up room, matching the hub's own rule; pinned by a test (tests/bags.test.ts).
- Wire lockstep: equip_bag and unequip_bag validate types and integer sockets in server/game.ts, ride HEAVY_SELF_CMDS, and the bags delta key round-trips (tests/snapshots.test.ts, tests/command_schema.test.ts, tests/world_api_parity.test.ts all re-pinned after the merges).
- Persistence: CharacterState.bags is additive JSONB, defaulted per socket on load, with tampered ids dropped to null. The pre-bag migration only fires when the bags field is absent, so it runs exactly once; idempotency and the tampered post-bag case are tested.

Findings from this pass, all fixed in 09bcb672:

- should-fix (src/sim/interaction.ts): the walk-by party autoloot delegates into lootCorpse, and the client retries it every 2 seconds per corpse (src/game/autoloot.ts RETRY_MS), so a full-bags player standing near a corpse got the "Your bags are full." toast on loop. lootCorpse now takes a quiet flag the passive pass sets; a deliberate loot click still gets exactly one toast. Test covers both. The silence of the "no permission" and "too far" branches on that path relies on the pre-check using the same rights and range as lootCorpse, which is now documented as load-bearing at the pre-check.
- should-fix (src/ui/bags_window.ts): the bag-bar aria labels were built by concatenating two t() results with a hardcoded comma; they now come from one hudChrome.bags.bagSocketAria key. The informational backpack and empty sockets were divs with aria-label and hover-only tooltips, unreachable by keyboard; they are now focusable no-op buttons (aria-disabled).
- should-fix (src/ui/i18n.locales/): the second base merge left an orphaned duplicate section comment in five overlays; removed. Comment-only, no keys were affected.
- nit (tests/bags.test.ts): the migration boundaries at used=48 (the strict tier comparison must keep the common tier) and used=73 (first slot past the 72 ceiling) were unpinned; both now have exact-value tests.

Known accepted behaviors, called out so they are deliberate rather than accidental:

- A player who keeps their bags full can force any trade they join to cancel at confirm. Strictly weaker than the cancel button both parties already have; noted by the security pass as a nuisance, not a vector.
- Bag slot counts (6/8/10/12/14) and drop rates (wolves 2 percent, Old Greyjaw 35, Morthen and Vael in their bonus roll groups) are invented balance, not a classic formula; happy to tune.
- The migration grants the lowest quality tier that covers the earned space (a 30-slot save gets two common bags, never a free epic); hoards past the 72-slot ceiling keep a tolerated overflow that only blocks new pickups.

Gates at 09bcb672: full vitest suite green in isolation-verified runs (the handful of load-flaky files, loot timers and subprocess-spawning i18n freshness, pass deterministically alone and were flaky on base before this PR), tsc clean, parity 159/159, biome ci --changed clean, npm run build green, pre-push floor green.

EnriqueGF:
Follow-up to the review above: after it was posted the base landed four more PRs (item instances and professions, mail and calendar plus Housekeeping, unit frame customization, character-bound combo points), so this branch merged release/v0.20.0 again at 3b3c50cd. The only conflicts were the parity goldens and the generated i18n artifacts (both re-minted on the merged tree), the pinned registry counts (delta keys now 29, command universe 113 sends / 122 dispatch, IWorld 163 members), and the SimContext append points, all resolved keeping both sides. The bags code itself needed one semantic integration, already covered above: addStacked and countFit skip instanced slots, and the round-robin winner force-adds like the other non-retriable grants. Guard battery and the full suite are green at the new head; the PR reports mergeable.

PR #1367, opened 2026-07-03
feat(combat): retail-style character-bound combo points

Blaine1705:
## What

Combo points now belong to the character instead of the target, the way modern (retail) rogues and cat druids work:

- Building on any target adds to one pool; swapping targets no longer wipes it.
- The pool survives the combo target dying or despawning, so leftover points carry into the next fight.
- Finishers spend the pool on the player's current target.
- Unspent points fade 30 seconds after the last point was built (each new point restamps the window). The window is the COMBO_POINT_DURATION const in src/sim/sim.ts.

## How

- Sim: Entity.comboTargetId is removed in favor of Entity.comboUntil (sim-time expiry). awardCombo (kept on Sim, seam signature unchanged) stamps the window; a new updateComboExpiry in src/sim/combat/auras.ts fades the pool in the per-player tick loop (right after updateTimers; it draws no rng, so the global draw order is untouched). The finisher gate in casting_lifecycle.ts drops the same-target check, and the kill/despawn clear sites (damage, mob lifecycle, dungeons, delves, pets) drop their combo arms.
- Wire: the comboTgt self field is removed from selfWireJson and applySnapshot in the same change; the client only ever needs the combo count, which stays wired. comboUntil is deliberately server-internal.
- UI: the combo pip row moves from the target frame to the player frame (above the hp bar), so the banked pool is always visible, target or not. Nameplate pips now light over the player's CURRENT target (the entity the next finisher hits). Same elided writers, row still lazy-built once.
- The /combo readout drops its target mention; the blocked-source seed entry is updated in the same change.

## Testing

- New tests/combo_points.test.ts: pool survives a target swap and spends on the new target, survives the combo target dying, fades at 30s with the comboPoint 0 event, and each new point restamps the window.
- Updated: nameplate_combo, nameplate_view, combo_command, target_frame, client_shell, combat_effect_dispatch, interactions, and the c4b parity scenario.
- Parity goldens regenerated on this base with UPDATE_PARITY=1 (deliberate behavior change); every per-frame rng draw count and draw-order digest is byte-identical, only state/event digests moved.
- tsc, architecture, S3 localization, snapshots, and the parity gate are green on this branch. Also verified end to end in the browser: built 3 points on one wolf, swapped to a second, pool and pips persisted on the player frame and the new target's nameplate.


Blaine1705:
8af8f57152145d568c8395dc9f06f0a7

EnriqueGF (commented):
Nice, clean migration. The comboTargetId removal is complete (no leftover references anywhere, including scripts and tests), the wire change removes comboTgt from both selfWireJson and applySnapshot in the same change, blankEntity picks up comboUntil so the two worlds keep identical entity shapes, and index.html / play.html are mirrored. The combo test file covers the four behaviors that matter (swap, target death, 30s fade, restamp). Ran combo_points, nameplate_combo, nameplate_view, target_frame, combo_command and architecture locally on the head: all green.

I looked at the performance side closely since the pip row now runs every frame on the player frame:

- HUD (src/ui/hud.ts:5104): equal or cheaper than before. Non-energy classes pay one elided setDisplay per frame; energy classes pay one setDisplay plus five elided toggleClass calls, same as the old target-frame block. The row is still lazy-built once, the pip walk is an indexed loop over the live children collection, no per-frame allocation.
- Sim (src/sim/combat/auras.ts:110): updateComboExpiry is two comparisons per player per tick and draws no rng, so the placement after updateTimers keeps the global draw order intact. The parity goldens confirm it: draw counts and draw-order digests unchanged, only state/event digests moved.
- Render (src/render/nameplate_combo.ts:15): still a single comparison per nameplate per frame.
- Wire: dropping comboTgt trims a field from every 20Hz self snapshot, a small win.

No performance concerns.

Findings:

- should-fix: the PR is CONFLICTING against release/v0.20.0, and the true conflicts (git merge-tree) are exactly the tests/parity/golden/*.json files, nothing else. That is mechanical: the goldens were regenerated on a base that has since moved. Rebase onto current release/v0.20.0 and re-run UPDATE_PARITY=1; the source files all auto-merge.
- nit: comboPipsFor (src/render/nameplate_combo.ts:15) only checks targetId and dead, so a rogue with banked points who targets a friendly player or NPC lights 5 pips over that nameplate even though no finisher can land there. Cosmetic; gating on attackability would fix it if you care.

One non-issue worth stating: combo state is not persisted server-side, so the sim-time comboUntil never crosses a save/load boundary and there is no stale-epoch risk.

PR #1320, opened 2026-07-02
feature(ui): improve shield HUD indicators

a-aznar:
## Type of change

- [x] Feature: new functionality
- [ ] Bug fix
- [ ] Refactor or performance (no behavior change)
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change (please call it out in the summary)

## How was this tested?

- Commands:
- Manual steps:

## Screenshots / recordings

image

---

## Checklist

Tick the items that apply to your change. It's fine to leave the rest unchecked
or strike them through (`~like this~`) when they don't.

### Quality

- [x] **Tests pass.** `npm test` is green. I added or updated tests for any
      `sim/` or `server/` behavior I changed.
- [x] **Typecheck passes.** `npx tsc --noEmit` is clean (the project is
      TypeScript `strict`).
- [x] **Builds succeed.** `npm run build`, plus `npm run build:server` and
      `npm run build:env` if I touched those.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified on a desktop and on a
      phone-sized viewport in both portrait and landscape. Touch targets stay at
      least 40x40px and inputs at least 16px font (see `src/ui/CLAUDE.md`).
- [ ] **Accessible.** Keyboard-operable, with visible focus, sensible ARIA, and
      respect for `prefers-reduced-motion` (only if this change adds or edits UI).

### Localization (i18n)

- [ ] **New player-visible strings are translated into every supported locale.**
      Every user-facing string is a `t()` key, added to `en` first, then given a
      real translation in every locale in `supportedLanguages`
      (`src/ui/i18n.ts`). No English placeholders or `// TODO` in other locales.
- [ ] Numbers, money, dates, units, and percents go through the i18n formatters
      (`formatNumber`, `formatMoney`, `formatDateTime`, `Intl`).
- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at
      the client boundary in this same change, and
      `npx vitest run tests/localization_fixes.test.ts` passes.
- [x] N/A. This PR adds no player-visible strings.

### Hygiene

- [x] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is
      not enabled in any production path.
- [x] I didn't hand-edit generated files such as
      `src/render/assets/manifest.generated.ts`. They're regenerated by the build.

> **Tip:** commit messages and PR titles use [Conventional Commits](https://www.conventionalcommits.org/)
[...]

PR #1369, opened 2026-07-03
feat(ui): movable, scalable player frame + frame scale sliders + position reset

Blaine1705:
## What

Unit-frame customization, building on the movable target frame (#1244, now in the release):

- The PLAYER frame gets the same corner move/lock button: unlock, drag it anywhere, lock it back; the spot persists across reloads (desktop only, like the target frame).
- Moving the player frame detaches it from the action-bar stack and drops it to the target frame's compact bar width, so the two frames read as a pair. Re-docking (mobile layout, or the reset below) restores the stock look.
- Two new Interface sliders: Player Frame Scale and Target Frame Scale (0.7 to 1.15), each scaling just that frame.
- A new Interface row, Reset Frame Positions (next to Reset Chat Window), snaps both frames back to their stock spots, forgets the saved drags, and locks them.

NOTE: this PR merged faster than its second half landed on the branch, so the aura-strip work described in the comments below (buff anchoring, compact durations, party aura icons, the tooltip corner) is NOT in this merge; it arrived as the follow-up #1374.

## How

- The drag/lock wiring is extracted into a shared, instance-parameterized MovableFrame controller (src/ui/movable_frame.ts) that drives BOTH frames; the target frame's behavior is a faithful move (same math, same storage key, same button). The pure position math stays in target_frame_pos.ts, untouched.
- Detaching reparents the player frame from #actionbar-stack to #ui: the #bottom-bar ancestor carries a centering transform, which would otherwise hijack the absolute positioning. Painter refs are live nodes, so they survive the move.
- The scale sliders publish --player-frame-scale / --target-frame-scale, applied via zoom to each frame's CHILDREN, never the drag-positioned frame itself (zoom there would multiply the used left/top and corrupt the drag math).
- Mobile never sees a detached frame: the button is hidden on mobile-touch, the drag gate checks the layout, and the mobile branch clears the inline styles and re-docks.
- New keys (playerFrame.unlock/lock, frameReset.label, the two slider labels) ship with es plus the five non-Latin fills; the reset button reuses the existing chatWindow.resetAction key.

## Testing

- tests/movable_frame.test.ts (8 cases, hand-rolled fake DOM): unlock toggling aria-pressed, drag blocked while locked and on mobile, drag apply + clamp + persistence, saved-spot restore on boot, the mobile re-dock branch, reset(), and corrupt-data fallback.
- Updated: options_view / options_window (pinned control lists + the reset row), settings (defaults), plus the css corpus/validity, i18n completeness/coverage, architecture, hud_perf_budget, and focus_visible guards, green with tsc clean.
- Verified end to end in a real browser: dragging + persistence across reloads, the reset button through the real Esc menu, and both scale sliders.


Blaine1705:
https://github.com/user-attachments/assets/32e37083-d91f-49ae-b6e6-694000594632


https://github.com/user-attachments/assets/975dd962-5a90-4c75-b118-7757c9432f8d

Blaine1705:
Pushed the second half of the unit-frame work into this PR (the body is updated to match): the optional buff-row anchoring + minimap-side debuff corner, WoW-style compact aura durations (20s / 5m / 1h / 2d) across all three strips, the target debuff strip hanging below the frame instead of stretching it, and the mob-hover tooltip pinned to the bottom-right corner now that the player frame can move. All verified in-browser; 323 tests green on the branch.

Blaine1705:
Two more frame additions pushed here: each party row now shows a mini icon strip of that member's buffs/debuffs (a compact per-member aura summary rides the party wire: id + kind + sap flag, capped, no per-tick countdown so the delta elision keeps working), and the target frame's aura strip widened from debuffs-only to ALL auras, so a friendly target shows the shield you just cast on it and an enemy shows its own buffs alongside your DoTs. Both verified in-browser (offline party + friendly/hostile targets); 438 tests green on the branch. Note this complements #1320 (the absorb stripes): the stripes show the shield's amount on the health bar, these show its icon.

Blaine1705:
Housekeeping note: this PR merged within minutes (thanks!), and the 'second half' described in the two comments above was pushed to this branch AFTER the merge, so it never landed here. Those commits shipped as the follow-up PR #1374 (aura anchoring, compact aura durations, party aura icons, target power bar, tooltip corner). The body above is restored to describe exactly what this merge contains.

PR #1373, opened 2026-07-03
feat(input): brand-aware gamepad button glyphs

daxdax89:
## Summary

The Controller options panel currently labels every button with a single brand-neutral string ("A / Cross", "X / Square", ...). That is fine for Xbox and PlayStation, but it misleads Nintendo players: a Switch pad's face buttons are mirror-swapped versus an Xbox pad (the bottom button reads "B" on Switch, "A" on Xbox), so "A / Cross" points a Nintendo player at the wrong button.

This PR detects the connected controller's brand from `Gamepad.id` (PlayStation / Xbox / Nintendo, else generic) and labels each button with the glyph printed on that player's actual pad, including the Nintendo A/B and X/Y swap.

Bindings stay position-indexed, so `DEFAULT_GAMEPAD_BINDINGS` is unchanged and "bottom face button = jump" holds on every pad. Only the displayed text changes; there is no gameplay effect.

### How it works
- New pure core in `src/game/gamepad_map.ts`: `GamepadKind`, `detectGamepadKind(id)`, `GAMEPAD_BUTTON_LABELS_BY_KIND`, and `gamepadButtonLabel(button, kind)`. Detection checks product-name keywords first (present in both the Chrome and Firefox `id` formats), then reads the USB vendor id from its own field so a matching product-id hex cannot be mistaken for a vendor.
- `GamepadManager` (`gamepad.ts`) stores the detected brand on connect, exposes `getKind()`, and fires an `onConnectionChange` callback.
- The Controller panel (`options_window.ts`) labels each button via the detected brand and re-labels live on connect/disconnect (`OptionsWindow.refreshControllerLabels` -> `Hud` passthrough -> `main.ts` wiring; `GamepadBindingsHooks` gained `kind()`).
- Glyphs stay non-`t()`, consistent with the existing controller-panel convention that hardware button names are language-neutral.

## Type of change

## How was this tested?

- Commands: `npx vitest run tests/gamepad_map.test.ts tests/gamepad.test.ts` (34 pass), `npx tsc --noEmit` (clean), `npx @biomejs/biome ci <changed files>` (clean), `npm run build` (clean).
- Added unit tests: brand detection (name + Chrome/Firefox vendor formats, the product-id-collision case, xinput/x-box names, generic/empty fallback), the Nintendo A/B/X/Y swap, per-brand shoulder/face names, shared d-pad, full per-brand button coverage, and `GamepadManager` kind lifecycle (connect/disconnect/stop/getKind + the connection-change callback).
- Not yet verified on physical hardware: real PS5/Xbox/Switch brand detection and the live re-label are the one item that needs an actual controller.

### Quality
- [x] Tests pass; added tests for the new behavior.
- [x] Typecheck passes (`npx tsc --noEmit`).
- [x] Builds succeed (`npm run build`).

### Localization (i18n)
- [x] N/A. The button glyphs are hardware names, kept non-localized per the existing controller-panel convention (`hud_chrome.ts` note + `gamepad_map.ts`). No new `t()` keys.

### Hygiene
- [x] No secrets or generated files touched.

## Notes for reviewers

- Out of scope (pre-existing, noted for a possible follow-up): when the active pad is unplugged while a second pad stays connected, the manager goes idle instead of re-scanning for the remaining pad. This PR does not change that behavior.

### DIFF
--- a/src/game/gamepad.ts
+++ b/src/game/gamepad.ts
@@ -8,7 +8,9 @@
 import type { GamepadBindings } from './gamepad_bindings';
 import {
   AXIS,
+  detectGamepadKind,
   GAMEPAD_NONE,
+  type GamepadKind,
   GP,
   risingEdges,
   STANDARD_BUTTON_COUNT,
@@ -30,12 +32,16 @@ export interface GamepadCallbacks {
   isPointerMode(): boolean;
   // Current local-player health, for rumble-on-damage. Optional.
   getPlayerHealth?(): number;
+  // A pad connected or disconnected, so the detected brand (and thus the button
+  // glyphs shown in the Controller options panel) may have changed. Optional.
+  onConnectionChange?(): void;
 }
 
 const CURSOR_SPEED = 900; // px/sec at full stick deflection in UI cursor mode
 
 export class GamepadManager {
   private index: number | null = null;
+  private kind: GamepadKind = 'generic';
   private prevPressed: boolean[] = new Array(STANDARD_BUTTON_COUNT).fill(false);
   private deadzone = 0.18;
   private camSpeed = 2.4;
@@ -59,10 +65,13 @@ export class GamepadManager {
     if (typeof navigator === 'undefined' || typeof navigator.getGamepads !== 'function') return;
     window.addEventListener('gamepadconnected', this.boundConnect);
     window.addEventListener('gamepaddisconnected', this.boundDisconnect);
-    // Pick up a pad that was already connected before we started listening.
+    // Pick up a pad that was already connected before we started listening (e.g.
+    // the Controller setting toggled on while a pad is plugged in). Notify so an
+    // open Controller panel re-labels to the detected brand; a no-op otherwise.
     for (const pad of navigator.getGamepads()) {
       if (pad?.connected) {
-        this.index = pad.index;
+        this.acquire(pad);
+        this.cb.onConnectionChange?.();
         break;
       }
     }
@@ -77,6 +86,7 @@ export class GamepadManager {
     // camera, and edge buttons after the Controller setting is turned off. start()
     // re-acquires an already-connected pad on re-enable.
     this.index = null;
+    this.kind = 'generic';
     this.prevPressed.fill(false);
     this.input.clearGamepadMove();
     this.hideCursor();
@@ -99,16 +109,33 @@ export class GamepadManager {
     return this.index !== null;
   }
 
+  /** Detected brand of the connected pad, for glyph labeling; 'generic' when
+   *  none is connected or the pad's id is unrecognized. */
+  getKind(): GamepadKind {
+    return this.index === null ? 'generic' : this.kind;
+  }
+
+  // Latch a pad as the active one and classify its brand from the id string.
+  private acquire(pad: Gamepad): void {
+    this.index = pad.index;
+    this.kind = detectGamepadKind(pad.id);
+  }
+
   private onConnect(e: GamepadEvent): void {
-    if (this.index === null) this.index = e.gamepad.index;
+    if (this.index === null) {
+      this.acquire(e.gamepad);
+      this.cb.onConnectionChange?.();
+    }
   }
 
   private onDisconnect(e: GamepadEvent): void {
     if (this.index === e.gamepad.index) {
       this.index = null;
+      this.kind = 'generic';
       this.prevPressed.fill(false);
       this.input.clearGamepadMove();
       this.hideCursor();
+      this.cb.onConnectionChange?.();
     }
   }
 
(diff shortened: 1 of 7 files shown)


FernandoX7 (commented):
Nice feature, and the shape is right: the detection and label tables live in the pure `gamepad_map` core with real unit tests, bindings stay position-indexed so nothing changes about gameplay, the Nintendo positional swap matches the W3C standard-mapping positions, and the live re-label reuses the panel's existing in-place re-render (the remap dropdown's aria-label stays in sync with the visible glyph, which keeps the WCAG naming intact). I verified the suite (34 passing at your head), `tsc`, and Biome locally, and the non-`t()` stance matches the documented convention in `hud_chrome.ts`.

Housekeeping first: the branch was cut from main after the #1328 merge, so the PR also carried the daily-rewards chest feature and showed as conflicting against `release/v0.20.0`. Keeping it targeted at the release branch is intended, so I merged `origin/release/v0.20.0` into the branch and resolved the conflicts (kept both new test suites in `tests/gamepad.test.ts`, combined both leaderboard changes in `daily_rewards_window.ts`, both key sets in `hud_chrome.ts` and the five overlays, and re-ran the i18n build for the generated trio). It is mergeable now; nothing for you to do there, but heads up that your local branch is behind the remote.

Findings on the gamepad change itself:

- should-fix: the product-id-collision test does not pin what its comment claims. `tests/gamepad_map.test.ts:172` uses `'Xbox Wireless Controller (Vendor: 045e Product: 054c)'`, which returns from the "Xbox" name branch before any vendor parsing runs, so a naive whole-string substring scan for `054c` would still pass it. The decisive case is the same id without a name keyword: `expect(detectGamepadKind('Wireless Controller (Vendor: 045e Product: 054c)')).toBe('xbox')`. That one fails a naive implementation and actually proves the vendor is read from its field.
- nit: two detection arms are never exercised alone. Every PlayStation name-path case also carries vendor `054c` (a bare `'DualSense Wireless Controller'` would pin the name arm), and the only `057e` string also says "Pro Controller" (a `'Wireless Controller (Vendor: 057e Product: 2009)'` would pin the Nintendo vendor arm). One-line assertions each.
- nit: the false branches of the connect/disconnect guards are untested: a second pad connecting while one is active (should not hijack index/kind or notify), and a non-active pad disconnecting (should not reset kind or fire `onConnectionChange`). Both would regress silently today.
- nit, optional: `Create` is the DualSense name; a DualShock 4 says `Share` (same family as View/Menu vs Back/Start on older Xbox pads). `'Share / Create'` would cover both generations, or keep the current-gen names if you prefer the cleaner label.

The "not verified on physical hardware" note is fair, and the failure mode is benign (unrecognized ids fall back to the combined labels), so I would not hold the PR on it. A quick check with a real pad before the release cut would still be good.

daxdax89:
Thanks for the thorough review, and for resolving the release merge on your end. Quick heads-up on that: around the same time I'd independently rebased this branch onto `release/v0.20.0` (with `git rebase --onto` so only the gamepad commit replays, dropping the daily-rewards commit that main had dragged in), and my force-push overwrote your merge commit before I saw your note. Sorry about the collision. The branch is now a single clean commit on top of release (7 files, mergeable), and I've folded your findings into it:

- **should-fix (vendor-field test):** added the decisive case with no name keyword, `detectGamepadKind('Wireless Controller (Vendor: 045e Product: 054c)')` -> `xbox`, which a naive whole-string `054c` scan would fail. Kept the name-priority case too.
- **nits (unexercised arms):** added `'DualSense Wireless Controller'` -> `playstation` (name arm alone) and `'Wireless Controller (Vendor: 057e Product: 2009)'` -> `nintendo` (vendor arm alone).
- **nits (guard false-branches):** added tests that a second pad connecting while one is active does not hijack the index/kind or fire `onConnectionChange`, and that a non-active pad disconnecting is a no-op.
- **optional (Share/Create):** the PlayStation `BACK` label is now `Share / Create` to cover both DualShock 4 and DualSense.

Suite is green (42 in the two gamepad files), tsc and Biome clean. I'll do a real-controller pass before the release cut. Thanks again!

FernandoX7:
No harm done on the force-push, the rebase is the cleaner shape and it keeps this PR scoped to the gamepad change.

Verified at b4cef12: all four findings are in (the no-name-keyword vendor case, the two single-arm detection cases, the two guard no-op tests, and Share / Create with a clear comment explaining the two generations). The interdiff against your previous head is exactly those changes plus rebase context, the two suites pass along with the architecture and i18n guards, and tsc is clean on this end too. The branch also merges cleanly into the current release tip even though it has moved since your rebase, so nothing to do about the drift.

A hardware pass before the release cut sounds right. Good to go from my side.

FernandoX7 (approved):

PR #1374, opened 2026-07-03
feat(ui): aura strips, compact durations, and tooltip polish for the unit frames (follow-up to #1369)

Blaine1705:
## What

The second half of the unit-frame work: #1369 merged faster than these landed on its branch, so they arrive as this follow-up (the branch is the same lineage; the diff is exactly the eight new commits).

- A new Interface toggle, Buffs on the Player Frame (off by default): the player's own buff row anchors to the movable player frame (above it docked, below it once moved) and the debuff row slides up beside the minimap, classic WoW's debuff corner. Off keeps the stock two-row top-right corner.
- WoW-style compact aura duration labels on every strip: 20s / 5m / 1h / 2d instead of raw seconds. Long buffs finally show a label at all (anything at or over 99s used to read as permanent); toggle auras (stealth, forms, stance, Ghost Wolf) show no countdown; 3599s reads 1h, never 60m.
- The target's aura strip hangs BELOW the frame instead of stretching the panel taller with every fresh debuff, and it widened from debuffs-only to ALL auras: a friendly target shows the shield you cast on it, an enemy shows its own buffs alongside your DoTs.
- Party rows grow a mini icon strip of each member's buffs/debuffs. A compact per-member aura summary rides the party wire (id + kind + sap flag, capped at 8, deliberately no per-tick countdown so the party delta elision keeps working; optional on the shape, so an older snapshot decodes as none).
- The mob-hover tooltip is pinned to the viewport's bottom-right corner (the WoW default GameTooltip slot); it used to anchor beside the player frame, which wandered once that frame became movable.

## Testing

- auras_view: 13 new/updated cases (compact units per magnitude, injected localized units, promote-at-boundary, the toggle-aura hide including Ghost Wolf vs Sprint sharing buff_speed).
- party_frames / party_frames_painter: the signature folds in each member's aura list (join/leave/kind/sap flips repaint; a missing field signs like an empty strip), and the pool test drives the per-row strip end to end (icons per wire aura, re-sync on change, empty for an aura-less member).
- Wire gates green: snapshots, bandwidth, world_api parity, plus architecture, hud_perf_budget, client_shell, css corpus/validity, i18n completeness/coverage: 438 tests on this branch, tsc clean.
- Verified in a real browser at every step: the anchored buff row following a moved frame, debuffs landing beside the minimap, live compact labels (stealth showing none; 15s / 30m / 1h), the target frame keeping its height as debuffs hang below it, a shielded ally showing the shield icon on its party row AND under the target frame, and the tooltip sitting in its corner.

Complements #1320: the absorb stripes show how much shield remains on the health bar; these strips show WHICH auras sit on the unit.


Blaine1705:
Two more pushed: the target frame now shows its target's power bar (the per-entity wire carries rtype/res/mres for anything that HAS a resource, rounded so an idle entity's record stays byte-stable for the dyn-cache elision; a resource-less beast renders the classic empty dark rail instead of collapsing, so the frame never changes height), and clicking the player frame body self-targets, the classic convention (the move button and the anchored aura rows keep their own behavior). Verified in-browser: self-click shows your own energy bar in the target frame, a priest ally shows mana, a wolf shows the empty rail.

Blaine1705:
Merged the latest release/v0.20.0 tip into the branch to clear the conflicts (the bags + map editor + quest-flash merges had moved the shared files). Only real collision was the target frame markup, resolved keeping this PR's structure (the debuff strip below the frame); i18n artifacts regenerated. 292 tests green on the branch after the merge, tsc clean.

EnriqueGF:
(Was part of other related PR)

EnriqueGF (commented):
Reviewed the eight-commit follow-up end to end in a worktree off the current release base. This is clean, seam-respecting work: the aura core stays pure and the new `compactAuraDuration` is a tidy exported function with real boundary tests (promote-at-59:59, ceil-on-seconds, toggle-hide by kind and by id for Ghost Wolf vs Sprint), the party mini-strip is a per-row instance of the existing keyed aura pool rather than a new banner, and the anchored-buff reparent keeps the painters' live node refs so the option flips without rebuilding anything.

Verified locally against `release/v0.20.0` (merges clean, no conflicts):
- tsc clean, and the targeted suites green (auras_view / party_frames / party_frames_painter / target_frame / settings / options_view / auras_painter = 113 passing), plus the guard set (architecture, bandwidth, snapshots, world_api_parity, client_shell, hud_perf_budget = 371 passing).
- Wire parity holds both ways. The target resource fields (`rtype`/`res`/`mres`) encode only when `resourceType` is set (`server/game.ts`) and decode guarded by `w.rtype !== undefined` (`online.ts`), so a resource-less target keeps the blank defaults and absence never clears; `blankEntity` already carries the three defaults, so the offline Sim and ClientWorld entity shapes stay identical. The party aura payload is byte-identical between `Sim.partyInfo` and the server `partyWire` (same `PARTY_MEMBER_AURA_CAP` const, same `value < 0` neg flag), and `PartyMemberInfo.auras` is optional so an older snapshot decodes as no-auras.
- index.html and play.html got the same `#tf-resource` / `#tf-debuffs` DOM changes, so the `/play` entry keeps parity.
- New player strings are English `t()` keys in `hud_chrome` (`durationUnit{Minutes,Hours,Days}`, `aurasOnPlayerFrame`); resource/duration text goes through the formatters and the injected units, no concat.

Two notes, neither blocking:

- nit, scope: the net diff also lands the target-frame power bar and click-the-player-frame-to-self-target (commits a23999c8, a3c7d7a4), which the description does not mention. Both are sound and covered by the target_frame tests, so no objection, just flagging so the release notes credit them.
- info, `server/game.ts` dynamicFields: a targeted other entity's resource ships as `Math.round(e.resource)` while the self record keeps tenths. Fine and deliberate (byte-stable dyn-cache elision, and it applies equally to every online client), noting only that it is a conscious precision call.

Nothing to fix. Merging.

PR #1375, opened 2026-07-03
feat(auth): self-service password reset

daxdax89:
## Summary

Adds a self-service password reset ("forgot password") flow. Today a user who forgets their password and did not link Discord has no way back in without an admin; this closes that gap for anyone with an email on file. It mirrors the existing verified-email-change pipeline end to end (same token/table/consume shape), so it introduces no new security primitives.

### Server
- **New `password_reset_requests` table** (additive, idempotent DDL next to `email_change_requests`), storing only the **SHA-256** of a **single-use, time-boxed (1 hour)** token. Helpers `createPasswordResetRequest` / `consumePasswordResetRequest`. The consume is one **atomic transaction**: the `WHERE consumed_at IS NULL AND expires_at > now()` claim UPDATE is both the replay and race guard, then it sets the new hash and **revokes every session** (`DELETE FROM auth_tokens` for the account, the correct posture for account recovery).
- **`POST /api/account/password/forgot`** — identify by username, mail a reset link to the account's on-file email. **Always returns an identical `200`** (unknown user, known-no-email, success), so it never reveals whether an account exists. An account with no email on file simply cannot self-reset (recovers via Discord or an admin).
- **`POST /api/account/password/reset`** — validate the token + new password (`MIN/MAX_PASSWORD_LENGTH`), apply, revoke all sessions. Invalid and expired collapse to one identical `400`.
- Both routes are **rate-limited** and behind the **web-login Origin guard** (same class as login/register). New `password_reset` email event + English template + a client-page reset-link builder (`/?reset=<token>`).

### Client
- A "Forgot password?" link on the login panel, a request panel, and a reset panel reached via the `?reset=<token>` link. New English-only `hudChrome.auth.*` keys (with the M16 non-Latin fills in the same change, per the i18n rule).

## Type of change

## How was this tested?

- Commands: `npx tsc --noEmit` (clean), `npx vitest run` on the new + guard suites (112 pass, incl. anti-enumeration, single-use/expiry, revoke-all, password validation, i18n completeness, architecture, localization), `npm run i18n:build`, `npx @biomejs/biome ci <changed>` (exit 0), `npm run build` (clean).
- New tests: `tests/password_reset_server.test.ts` (handler branches via the mock-pg harness) and `tests/password_reset_client.test.ts` (REST methods).
- Local review: ran migration-safety over the DDL (additive/idempotent, correct FK cascade, atomic consume) and a privacy/security pass over the flow (anti-enumeration, hashed single-use token, revoke-all, rate-limit + Origin guard) — no critical/high findings.
- Not yet exercised against a live SMTP path: dev uses the ConsoleSender, so the reset email prints to the server log.

### Quality
- [x] Tests pass; added tests for the new server + client behavior.
- [x] Typecheck passes.
- [x] Builds succeed (`npm run build`).

### Localization (i18n)
- [x] New player-visible strings are `t()` keys in `hudChrome.auth.*` (English), with the required non-Latin fills. Server text stays English at source; no new sim/server player strings needing the matcher.

### Hygiene
- [x] No secrets committed. Only token hashes are stored; raw tokens live only in the email link. No generated files hand-edited (regenerated via `npm run i18n:build`).

## Notes for reviewers

- **Scope decision:** identify-by-username (the account's on-file email receives the link) rather than identify-by-email, since character/login identity here is the username and `accounts.email` is not unique.
- **Known, consistent-with-repo tradeoffs (not addressed here, happy to follow up):** `/forgot` uses the same per-IP rate limit as register/login (also email-sending); a stricter per-account throttle could further limit reset-email spam. Response timing is not fully constant (the number of DB round-trips differs by path), matching the existing `handleEmailUnsubscribe` posture. Consumed/expired rows are pruned per-account on the next request but there is no global reaper (same as `email_change_requests`).


FernandoX7:
@daxdax89 please make sure to target the latest release branches in your PRs. Right now, that is release/0.20.0

Thanks for your contributions!

daxdax89:
Thanks for the heads-up! I've retargeted this PR (and #1373) to `release/v0.20.0` and rebased both branches onto it. Will branch off the current release for any follow-ups too.

FernandoX7 (commented):
Nicely done overall. The flow mirrors the email-change pipeline faithfully: a hashed, single-use, time-boxed token (only the SHA-256 is stored), an atomic consume where the `WHERE consumed_at IS NULL AND expires_at > now()` claim UPDATE is both the replay and race guard, revoke-all-sessions as the account-recovery posture, and identical 200s on `/forgot` so the response never reveals whether an account exists. The DDL is additive and idempotent, the SQL is parameterized, `MAX_PASSWORD_LENGTH` is checked before hashing (no bcrypt DoS), and the tests are decisive: the revoke-all case even asserts the absence of the `token <>` keep-caller clause, and the anti-enumeration case covers all four branches. Two findings.

## Blocking

- `src/main.ts:6871` The forgot/reset wiring in `wireStartScreens()` runs unconditionally, but `#btn-forgot-open`, `#forgot-panel`, and `#reset-panel` only exist in `index.html`. `play.html` carries the same login chrome (`#login-panel`, `#mode-select`) and boots through the same `/src/main.ts`, so on `/play` `$('#btn-forgot-open').addEventListener(...)` dereferences `null` (`$` is `querySelector(...) as T`, `src/main.ts:224`) and throws. That aborts the rest of `wireStartScreens()`: everything after that line, including the session-restore branch (`api.restoreSession()`), never runs, so a returning player on `/play` is no longer auto-logged-in and the start screen is left half-wired. The same function already guards entry-specific buttons with optional chaining (`document.getElementById('btn-wallet-switch')?.addEventListener`, `src/main.ts:6182`), and the new `?reset=` restore branch guards with `document.getElementById('reset-panel')`, so the wiring block is the one spot that skipped the guard. Fix it the same way (early-return if `#forgot-panel` is absent, or `?.` on each handle). The reset link is `/?reset=` and always lands on `index.html`, so the panels are legitimately index-only, guarding the wiring is the right fix rather than duplicating the panels into `play.html`.

## Nits

- Per-account reset throttle: you already called this out. `/forgot` shares the per-IP limiter with register/login, so someone who knows a username can email-bomb that inbox from rotating IPs. Fine to defer, but a per-account send cap (the `email_log` table you mention is the natural source) would close it.
- The reset token travels in the URL query (`/?reset=`), so it can land in server access logs before the client strips it with `replaceState`. Single-use plus the 1 hour TTL plus the strip make this the standard reset-link tradeoff, just noting it.
- Test gaps (nits, not blockers): no case for a too-long (>128) password on `/reset`, and nothing asserts the "your password changed" notice fires after a successful reset.

The timing-side-channel and no-global-reaper points in your notes match the existing account routes, agreed those are fine to leave.

daxdax89:
Good catch, thank you. Fixed: the forgot/reset wiring is now guarded on `#forgot-panel` presence (via `getElementById`), so `/play` skips it and session-restore plus char-select boot normally, matching the existing `?reset=` guard just below it. Also added the two tests you flagged: a too-long (over MAX) password on `/reset` returns 400 with no password write, and the `emailPasswordChanged` notice fires after a successful reset. Rebased onto the current `release/v0.20.0` tip. The per-account send cap is a good follow-up; happy to do that as a separate PR against the `email_log` table so this one stays scoped.

daxdax89:
Thanks for taking this in, and for extending it with the SES transport, the full-locale fills, and the deploy docs. I can see it landed on release/v0.24.0 (667227aa0), so I'll close this PR as merged. Happy to help with any follow-ups.

PR #1376, opened 2026-07-03
feat(ui): WoW-style quest progress flash at the top-center

Blaine1705:
## What

Killing a quest mob or picking up a quest item flashes the classic yellow progress line at the top-center of the screen ('Forest Wolf slain: 3/8'), the vanilla objective-progress cue. Both kill and collect objectives fire it.

## How

- Fed by the SAME localized questProgress SimEvent the chat line already uses, so no new sim surface, wire field, or i18n key: the Hud localizes once and hands the string to both sinks.
- A small event-driven module (src/ui/quest_progress_banner.ts) owns the flash: up to three lines stack (a multi-objective loot burst shows every line), the oldest drops on overflow, and each line fades on its own timer. Event-driven, never a per-frame path.
- The container is aria-hidden decoration: the chat log keeps the durable copy and the existing live region announces it, so screen readers hear it exactly once.
- Sits above the zone-banner slot (top 15%), yellow with the classic dark text-shadow.

## Testing

- New tests/quest_progress_banner.test.ts (4 cases, fake DOM + fake timers): append order, oldest-drops overflow, the fade-then-remove timer chain, and the evicted-line no-throw edge.
- client_shell, css corpus/validity, architecture, hud_perf_budget, localization_coverage green; tsc clean.
- Verified in a real browser: accepted Wolves at the Door at Marshal Redbrook, killed a wolf, and the yellow 'Forest Wolf slain: 1/8' flashed top-center (screenshot-confirmed against the classic reference).

### DIFF
--- a/index.html
+++ b/index.html
@@ -413,6 +413,7 @@
     <div id="tooltip" class="panel"></div>
 
     <div id="error-msg"></div>
+    <div id="quest-banner" aria-hidden="true"></div>
     <div id="banner"></div>
     <div id="low-health-vignette"></div>
     <div id="subzone-banner"></div>
--- a/play.html
+++ b/play.html
@@ -361,6 +361,7 @@
     <div id="tooltip" class="panel"></div>
 
     <div id="error-msg"></div>
+    <div id="quest-banner" aria-hidden="true"></div>
     <div id="banner"></div>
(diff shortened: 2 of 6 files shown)


EnriqueGF:
Manually checked. OK. Merged.

PR #1379, opened 2026-07-03
fix(sim): low props (campfires, crates) no longer block spell line of sight

Blaine1705:
## Problem

Standing behind the spawn campfire and healing an ally in front of it fails with 'no line of sight': the LoS ray is pure 2D, so ANY movement collider blocks a cast regardless of how low it is. A knee-high fire pit blocks a heal exactly like a chapel wall.

## Fix

The ray now runs eye-to-eye: SIGHT_HEIGHT (1.6) above each endpoint's ground, lerped per sample along the segment. An open-world obstacle blocks only when its visual top rises above the sight line, reusing the SAME precomputed cameraTopY the chase-camera occlusion has always used (no new data, no drift between what you see and what blocks).

- Casts over: campfires (top 1.45), crates (1.35), small rocks (1.25 x scale).
- Still blocks: buildings, trees, tents, fences, mud huts, ruin columns, wells, editor-placed assets, and every interior/delve/arena wall set plus editor blocker walls (no known top = the conservative always-block default).
- MOVEMENT collision is untouched everywhere: the fire still stops you walking through it, it just stops eating your heals.
- Symmetric by construction: mobs, pets, and players all resolve LoS through this one function, on all three hosts.

## Testing

- New cases in tests/fixes.test.ts against the real world data: a straddled campfire keeps blocking movement but clears sight; a straddled building still blocks sight. The existing dungeon-wall LoS pair (blocked through the wall, clear through the passage) stays green, pinning that walls still deny casts.
- The parity gate passes untouched (no golden drift: no existing scenario's combat crossed a low prop), plus architecture and the delve collider suites; tsc clean.

### DIFF
--- a/src/sim/colliders.ts
+++ b/src/sim/colliders.ts
@@ -712,6 +712,51 @@ export function cameraOcclusion(
   return best;
 }
 
+// Eye height (yards above the ground) for the spell line-of-sight ray. An
+// open-world obstacle whose visual top (`cameraTopY`, the same precomputed top
+// the camera occlusion uses) sits at or below the sight line no longer blocks a
+// cast: a campfire (top 1.45), a crate (1.35), or a small rock is something you
+// see and cast OVER, while buildings, trees, tents, and fences still block.
+// Colliders without a known top (the interior wall layouts) always block, the
+// conservative default, and MOVEMENT collision is untouched everywhere.
+export const SIGHT_HEIGHT = 1.6;
+
+// Does any collider at (x,z) rise above `sightY` (absolute world Y of the
+// sight line at that sample)? Mirrors resolvePosition's zone routing so
+// interiors, delves and the arena keep their wall sets, but tests pure overlap
+// (no push-out) and applies the low-obstacle skip only where tops are known.
+function sightBlockedAt(seed: number, x: number, z: number, r: number, sightY: number): boolean {
+  const overlapsAny = (list: Collider[], lx: number, lz: number, skipLow: boolean): boolean => {
+    for (const c of list) {
+      if (skipLow && c.cameraTopY !== undefined && c.cameraTopY <= sightY) continue;
+      if (pushOut(c, lx, lz, r) !== null) return true;
+    }
+    return false;
+  };
+  if (isDelvePos(x)) {
+    const delve = delveAt(x);
+    const mods = delve ? defaultDelveModules(delve.id) : [];
+    const loc = delveModuleLocal(x, z, mods);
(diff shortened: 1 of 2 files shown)


Blaine1705:
https://github.com/user-attachments/assets/d9148ee3-d467-4e4b-b9c4-064418bd6cce


https://github.com/user-attachments/assets/56ceab84-53e2-45ee-af44-ca575d5e74bc

EnriqueGF (commented):
Reviewed the LoS change end to end in a worktree off the current release base. Nicely scoped fix: it keeps movement collision completely untouched and only widens `lineOfSightClear`, reusing the same precomputed `cameraTopY` the chase-camera occlusion already relies on, so there is no new data and no drift between what you see and what blocks a cast.

Verified against `release/v0.20.0` (merges clean):
- The new `sightBlockedAt` faithfully mirrors `resolvePosition`'s zone routing (same `isDelvePos`/`isArenaPos`/`DUNGEON_X_THRESHOLD` dispatch, the same `INTERIOR_COLLIDERS[interior] ?? CRYPT_COLLIDERS` fallback, the same grid-cell key), and only the open-world grid path passes `skipLow: true`. Delve, arena, and interior sets keep `skipLow: false`, so every wall set still blocks with the conservative always-block default. Fences stay in the overlap test, so they still deny casts.
- Swapping the net-displacement `isBlocked` for a direct `pushOut(c) !== null` overlap test is equivalent for the blocked/clear boolean (a non-overlapping collider returns null at/beyond its boundary either way), and it is the more honest per-collider test for this purpose.
- Determinism holds: pure XZ geometry, no Rng/wall-clock, and the iteration is over a single stable collider array (no Map/Set order dependence). Symmetric across the three hosts since mobs, pets, and players all resolve LoS through this one function.
- tsc clean; ran fixes, architecture, blocker_colliders, delve_colliders, calibration_snapshot, snapshots, and the full sim suite (all green, no golden drift, confirming no existing combat scenario crossed a low prop).

The new tests pin the real behavior against world data (a straddled campfire keeps blocking movement but clears sight, a straddled building still blocks), and the existing dungeon-wall pair stays green.

One note, non-blocking:
- info: the eye-to-eye line lerps only the endpoint ground heights, so on sloped terrain a mid-segment collider's absolute top is compared against an interpolated (not exact-local-ground) sight Y. It matches how the camera occlusion already approximates, and terrain elevation between endpoints was never part of LoS, so this is no regression, just noting the approximation.

Looks correct. Merging.

PR #1381, opened 2026-07-03
fix(delves): stop delve companion swinging at double speed

daxdax89:
## Summary

Fixes a delve-companion bug found in an audit: an in-melee-range companion (Acolyte Tessa) swings roughly **twice as fast** as its `weapon.speed` intends.

`updateDelveCompanion` advances `companion.swingTimer` once unconditionally at the top of the tick, then decrements it **again** inside the in-reach branch before the `<= 0` ready check — so while in melee range the timer loses `2 * DT` per tick. The sibling `pet_ai.ts` advances its timer once per tick; this restores that.

## Fix
Remove the redundant in-branch decrement (`src/sim/delves/companion.ts`); the unconditional decrement already advances the timer once per tick. Out-of-reach / idle branches are unaffected (their extra `max(0, ...)` decrement only floors an unused timer and fires no swing).

## Type of change
- [x] Bug fix

## How was this tested?
- New module test in `tests/delve_companion.test.ts`: drives `updateDelveCompanion` with the companion in melee range and asserts the interval between swings equals `weapon.speed / DT` ticks. It failed before the fix (interval ~27 vs ~52) and passes after.
- Regenerated the `delve_companion` and `delve_lockpick` **parity goldens** (`UPDATE_PARITY=1`): their captured rng draw order shifted because the corrected cadence makes fewer `mobSwing` calls. Confirmed only those two goldens changed.
- `npx tsc --noEmit` clean; parity + delve suites green (231 passed); `tests/architecture.test.ts` green.

## Notes for reviewers
- This is a deterministic behavior change (fewer companion swings → fewer rng draws), hence the two golden updates. No other scenario's golden changed, so the impact is scoped to companion melee.

### DIFF
--- a/src/sim/delves/companion.ts
+++ b/src/sim/delves/companion.ts
@@ -127,7 +127,9 @@ export function updateDelveCompanion(ctx: SimContext, companion: Entity): void {
       }
     } else {
       companion.facing = angleTo(companion.pos, combatTarget.pos);
-      companion.swingTimer = (companion.swingTimer ?? 0) - DT;
+      // swingTimer is already advanced once per tick by the unconditional
+      // decrement above; do NOT decrement again here or the companion swings at
+      // double the weapon's cadence (mirrors pet_ai, which advances once).
       if (companion.swingTimer <= 0) {
         ctx.mobSwing(companion, combatTarget);
         companion.swingTimer = companion.weapon.speed * ctx.swingIntervalMult(companion);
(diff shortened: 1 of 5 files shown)


daxdax89:
Closing. This targeted release/v0.20.0, which has since shipped. Deprioritizing rather than retargeting onto the current release branch; can reopen if the delve-companion swing-timer fix is still wanted.

PR #1382, opened 2026-07-03
fix(server): decode request bodies once to preserve split UTF-8

daxdax89:
## Summary

Fixes a request-body corruption bug found in an audit. `readBody` accumulated each incoming chunk onto a JS string with `data += c`. When `c` is a `Buffer`, that calls `Buffer.toString()` (UTF-8) on **each chunk independently**, so a multi-byte UTF-8 character split across a chunk boundary (which routinely happens over TCP/HTTP) is decoded as two invalid halves and mangled into replacement characters (`U+FFFD`).

Any request body containing non-ASCII text — character/guild names, chat, emoji — could be corrupted when the bytes happened to split mid-character.

## Fix
Buffer the chunks and decode once at `end` with `Buffer.concat(chunks).toString('utf8')`, exactly mirroring the already-correct `readBinaryBody` in the same file. The byte cap, socket-destroy-on-overflow, and string-chunk tolerance are preserved.

## Type of change
- [x] Bug fix

## How was this tested?
- New regression test in `tests/http_util.test.ts`: emits a JSON body containing a 4-byte emoji + CJK/accented text as two chunks split **inside** the emoji, and asserts `readBody` resolves to the exact object. Fails before the fix (replacement chars), passes after.
- `npx tsc --noEmit` clean; `tests/http_util.test.ts` (17) and `tests/account_server.test.ts` (38, a heavy `readBody` consumer) green.

### DIFF
--- a/server/http_util.ts
+++ b/server/http_util.ts
@@ -59,17 +59,24 @@ export function readBody(
   maxBytes = DEFAULT_JSON_BODY_MAX_BYTES,
 ): Promise<any> {
   return new Promise((resolve, reject) => {
-    let data = '';
+    // Collect raw bytes and decode ONCE at the end. Concatenating each chunk onto
+    // a string (`data += c`) decodes every chunk independently, which mangles a
+    // multi-byte UTF-8 character split across a chunk boundary into replacement
+    // characters. Same buffer-then-decode-once strategy as readBinaryBody; the cap
+    // ordering here matches the old readBody (push the chunk, then check the cap),
+    // not readBinaryBody (which rejects before pushing an over-cap chunk).
+    const chunks: Buffer[] = [];
     let bytes = 0;
     let aborted = false;
     req.on('data', (c: Buffer | string) => {
       if (aborted) return;
-      bytes += typeof c === 'string' ? Buffer.byteLength(c) : c.byteLength;
-      data += c;
+      const chunk = typeof c === 'string' ? Buffer.from(c) : c;
+      bytes += chunk.byteLength;
+      chunks.push(chunk);
       if (bytes > maxBytes) {
         // Rejecting the promise does not pause the socket, so without
         // destroying the request a client could keep streaming unbounded
-        // data into `data`. Stop reading and ignore any further chunks.
+        // data into memory. Stop reading and ignore any further chunks.
         aborted = true;
         req.destroy();
         reject(new Error('body too large'));
@@ -77,6 +84,7 @@ export function readBody(
     });
     req.on('end', () => {
       if (aborted) return;
+      const data = Buffer.concat(chunks).toString('utf8');
       try {
         // Every route reads properties off the body, so only a JSON object is
         // a valid request body: a literal null, an array, or a primitive is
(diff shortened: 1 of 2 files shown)


Rubsey (commented):
Good catch and a clean fix. I verified it end to end: with the base `http_util.ts` swapped in, your new test fails exactly as described (the split emoji decodes to three U+FFFD chars that still parse as valid JSON with the wrong value, 1 failed / 16 passed), and at the PR head all 17 pass. The `indexOf(0xf0)` split-point assumption is sound for this literal (0xf0 can only be a 4-byte lead byte; the 2-byte accented and 3-byte CJK chars can never contain it). Blast radius is clean: all 65 `readBody` call sites across 12 server files only consume the parsed object or match the unchanged `body too large` / `bad json` message strings, nothing in the repo calls `req.setEncoding` (so production chunks are always Buffers), and the byte-cap, empty-body, abort, and error paths behave identically to base. `tests/account_server.test.ts` (38) and `tsc --noEmit` are green at the head, and neither file changed on the base since the merge-base.

Findings:

- should-fix: `tests/http_util.test.ts:46` puts a literal dragon emoji in the payload. The repo bans emoji glyphs in source (CLAUDE.md), and this added line trips the push-time copy scan (`.githooks/pre-push`, the `\x{1F000}-\x{1FAFF}` class) and the qa-stop hook. CI has no copy scan so checks stay green, but maintainer-side pushes containing this line will block. The repo precedent for an emoji payload under test is the escape form (`tests/maps_service.test.ts:486` uses `'\u{1F600}'`): switching the literal to `'\u{1F409}'` is byte-identical at runtime and keeps the scans clean.
- nit: `tests/http_util.test.ts:48` has no guard on `buf.indexOf(0xf0)` returning -1. If a future edit removes or downsizes the 4-byte char, `dragon` becomes -1 and the split degenerates to an ASCII boundary, so the test would pass without exercising a mid-character split. One line pins it: `expect(dragon).toBeGreaterThan(-1)`.
- nit: the new comment in `server/http_util.ts` says this mirrors `readBinaryBody`, which is true for the concat-then-decode strategy but not the cap ordering: `readBinaryBody` rejects before pushing the over-cap chunk, while this pushes first (matching the old `readBody`, so no behavior change). A small wording tweak would keep the two helpers from looking accidentally divergent.

Non-blocking note: a sweep for the same pattern elsewhere found the server-side siblings already correct (`server/discord.ts` `readJsonBody` buffers and decodes once; `readBinaryBody` never decodes). The only remaining per-chunk decodes are cosmetic dev-script log piping (`scripts/dev-realms.mjs:45` and two screenshot scripts), pre-existing and out of scope here.

CI was stuck at action_required (fork PR); I approved the runs.

daxdax89:
Thanks for the thorough review and for approving the CI runs. Addressed all three:
- Switched the test payload to the escape form `\u{1F409}` (byte-identical at runtime, keeps the copy-scan / qa-stop hooks clean for maintainer-side pushes).
- Added `expect(dragon).toBeGreaterThan(-1)` to pin the mid-character split so a future ASCII-only payload can't silently degenerate the test.
- Reworded the http_util comment to note it shares readBinaryBody's buffer-then-decode-once strategy but keeps the old readBody cap ordering (push then check), not readBinaryBody's reject-before-push.
Head is green (17/17, tsc clean).

PR #1383, opened 2026-07-03
feat(admin): improve live evidence page

madmatah:
## Summary

Improves the Live evidence admin page for reviewing and analyzing recent bot-detection evidences.

Since this view will now display historical data and no longer just online players, a few adjustments had to be made to make it more user-friendly.

- sorts sessions by latest observation by default and adds a sortable **Last observed** column
- adds client-side filtering by character name, IP, evidence label, and evidence description
- supports displaying multiple sessions for the same character
- adds localized labels and empty states for the new controls
- adds a JSON download containing the complete loaded dataset, for analysis.
- excludes character names, evidence descriptions, and IP addresses from exports
- set automatic refresh to 30s instead of 5s

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Unit tested

- Launched bot scripts and observe the results on the admin dashboard.

## Screenshots / recordings

image

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene


Rubsey (approved):
Reviewed with a multi-agent pass (code quality + security) plus full local verification at the release tip:

**Verification (local, since no CI ran on this branch):** `tsc --noEmit` clean · full suite 697 files / 7,556 tests passing · `npm run build` clean · regenerated i18n artifacts byte-identical to the hand-shipped ones · branch already contains the `release/v0.20.0` tip, merges as fast-forward.

**Security:** export redaction claim verified — `suspicious_sessions_export.ts` is a strict allowlist projection (schemaVersion, capturedAt, sessionCount, per-session accountId/characterId/observedAt/score, evidence {kind, weight, expiresAt}); no name/IP/detail, pinned by canary tests. No `{@html}`, filename is sanitized, endpoint auth unchanged. 

Non-blocking follow-ups worth a look:

1. **Snapshot type narrowing** (`src/admin/types.ts`): `snapshot: { capturedAt: number } | null` while the server ships the full `SessionRuntimeSnapshot` — violates the "match the server return exactly" convention and hides what's actually in browser memory.
2. **Binary plural** (`resultCountOne`/`resultCountMany`): first count-plural in the admin catalog; hardcoded `count === 1` can't express ru/pl CLDR categories. Consider porting `tPlural` or naming keys by CLDR category before locales get filled.
3. **Deleted persistence test**: the auto-refresh localStorage round-trip is now unpinned (the old test was removed but the behavior still ships).
4. **Download test doesn't read the blob** — asserting the exported JSON contains the *unfiltered* dataset would pin the headline contract.
5. Duplicate-characterId rendering and null-snapshot sorting work (probe-tested) but have no fixtures; a future re-add of the each-key would crash on duplicates and still pass the suite.
6. Minor: unkeyed `{#each}` + 30s refresh can move keyboard focus across rows on reorder (composite key would fix); sync `revokeObjectURL` right after `click()` has historically aborted Safari downloads; `AUTO_REFRESH_MS` lives page-local instead of `state/poll.ts`; two dead i18n keys (`suspiciousPlayers.description`/`.empty`) left in the catalog.

Merging.

PR #1385, opened 2026-07-03
fix(admin): require a full-scope token for the admin API

daxdax89:
## Summary

Security fix from an audit: the admin API ignored token **scope**. `adminAccountId` resolved the bearer token with `accountForToken` and gated only on `isAdminAccount`, so a **read-scoped** token (an OAuth / companion-app token minted for an admin's own account) could drive every admin mutation, bans, suspensions, chat-filter and config changes, account notes.

## Fix
Resolve the token with `accountAndScopeForToken` and reject anything that is not mutation-capable via the existing named policy `scopeAllowsMutation` (i.e. require a `full` session token) **before** the `isAdminAccount` check. Admin logins already issue `full` tokens, so legitimate admin use is unaffected; a `read` token now never reaches the admin API.

## Type of change
- [x] Bug fix (security)

## How was this tested?
- New test in `tests/admin.test.ts`: a read-scoped token on an admin account is rejected (401) and `isAdminAccount` is never called (scope is checked first).
- Existing admin-API suite updated to the scope-aware token mock; all 46 tests pass. `npx tsc --noEmit` clean.

## Notes for reviewers
- `scopeAllowsMutation` / `accountAndScopeForToken` are the same helpers `oauth.ts` already uses to gate mutation vs read; this brings the admin API in line with that single policy.
- The whole admin API is privileged, so requiring full scope for the entire surface (not per-route) is intentional; there is no admin route a read/companion token should reach.

### DIFF
--- a/server/admin.ts
+++ b/server/admin.ts
@@ -47,6 +47,7 @@ import {
   type WordTier,
 } from './chat_filter_db';
 import {
+  accountAndScopeForToken,
   accountById,
   accountForToken,
   accountMailTarget,
@@ -55,6 +56,7 @@ import {
   pool,
   revokeTokensExcept,
   saveToken,
+  scopeAllowsMutation,
   setAccountDeactivated,
   touchLogin,
   updatePasswordHash,
@@ -225,12 +227,17 @@ interface AdminIdentity {
 async function adminIdentity(req: http.IncomingMessage): Promise<AdminIdentity | null> {
   const m = /^Bearer ([a-f0-9]{64})$/.exec(req.headers.authorization ?? '');
   if (!m) return null;
-  const accountId = await accountForToken(m[1]);
-  if (accountId === null) return null;
-  const staff = await adminRolesForAccount(accountId);
+  const info = await accountAndScopeForToken(m[1]);
+  if (!info) return null;
+  // The whole admin API is privileged reads + mutations, so a read-scoped token
+  // (an OAuth / companion-app token) must never reach it, even on a staff account.
+  // Require a full-session token, matching scopeAllowsMutation elsewhere, before the
+  // role lookup.
+  if (!scopeAllowsMutation(info.scope)) return null;
+  const staff = await adminRolesForAccount(info.accountId);
   if (staff === null) return null;
   return {
-    accountId,
+    accountId: info.accountId,
     username: staff.username,
     roles: staff.roles,
     permissions: permissionsForRoles(staff.roles),
@@ -1037,6 +1044,9 @@ function makeRealAdminDb() {
     moderationReportsForAccount,
     muteAccountChat,
     accountForToken,
+    // The pipeline admin gate (createRequireAdmin) reads token scope so a read-scoped
+    // companion token of a staff account is rejected, matching the legacy adminIdentity fix.
+    accountAndScopeForToken,
     accountMailTarget,
     findAccount,
     // Target-account staff check (the "admin accounts cannot be suspended / banned /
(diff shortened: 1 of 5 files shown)


Rubsey (commented):
The vulnerability is real and the fix direction is right: read-scoped companion tokens live in the same `auth_tokens` table with the same 64-hex format, so they matched the bearer regex and sailed through `accountForToken` + `isAdminAccount`. Using the existing `accountAndScopeForToken` / `scopeAllowsMutation` policy instead of inventing a new check is the correct move, and the test that pins `isAdminAccount` is never called for a read token is exactly the right assertion. 66 tests pass and `tsc --noEmit` is clean at the head.

One blocking gap, though:

- **blocking** `server/http/middleware/require_admin.ts:120`: this PR patches only the legacy `adminIdentity` resolver, but the admin API has a second auth gate, `createRequireAdmin`, which the migrated RouteDef pipeline mounts on every admin route, and it still resolves the bearer with scope-blind `accountForToken`. The file even documents the hole it carries ("a read-scope companion token of a staff account passes too, parity-first", the comment above the `ctx.account = { accountId, scope: 'full' }` stamp). Production now defaults to `API_DISPATCH=new` (see `logApiDispatchSelection` in `server/main.ts`: legacy in prod is the loud rollback path), so the arm this PR fixes is the one prod is not running. To close the hole where it matters: extend `AdminAuthDb` to resolve `accountAndScopeForToken`, reject on `!scopeAllowsMutation(scope)` before the `adminRolesForAccount` lookup (mirroring your ordering here), wire it through `makeRealAdminDb` in `server/admin.ts`, fix the now-truthful `ctx.account` scope stamp, and add the same read-token-rejected test against the pipeline arm.

Smaller notes:

- **nit** `tests/admin.test.ts:15`: the comment says "use the real implementation" but `scopeAllowsMutation: (s) => s === 'full'` is a reimplementation of the policy, which can silently drift from `server/db.ts` if the policy ever changes. Pulling the real one in via `importOriginal` in the mock factory would make the comment true.
- Note for the maintainer, not the contributor: no CI has run on this head (fork PR awaiting workflow approval), so the green suite above is from a local run at the head commit.

Once the pipeline gate gets the same treatment this is a clean, well-scoped security fix.

daxdax89:
Thanks for catching the pipeline gap, that was the important one. Closed the hole on the arm prod actually runs (API_DISPATCH=new):

**blocking (createRequireAdmin pipeline gate):**
- Extended `AdminAuthDb` to `accountAndScopeForToken` (returns `{ accountId, scope }`) instead of `accountForToken`, and the gate now rejects on `!scopeAllowsMutation(info.scope)` BEFORE the `adminRolesForAccount` lookup, mirroring the legacy `adminIdentity` ordering. A read-scoped companion token of a staff account never even resolves roles now.
- Wired `accountAndScopeForToken` through `makeRealAdminDb` in `server/admin.ts`.
- Made the `ctx.account = { scope: 'full' }` stamp truthful (only full tokens reach it now) and updated the module header + comment that documented the old parity-first hole.
- Added a pipeline-arm test: a `scope: 'read'` token on a superadmin account 401s and `adminRolesForAccount` is asserted never-called. Updated the other pipeline gate fakes (`tests/server/admin.test.ts`, `ownership_coverage.test.ts`) to the new shape.

**nit:** `tests/admin.test.ts` now pulls the real `scopeAllowsMutation` via `importOriginal` in the mock factory, so the comment is true and the policy can't silently drift.

tsc clean; 322 admin/pipeline tests pass (incl. the new read-token case). Both auth arms now reject read-scoped tokens.

jgyy:
Closing — this looks already implemented. `require_admin.ts` on `upstream/main` already enforces an exact `"full"` scope check, so this change is redundant. Thanks for the contribution; reopen if you spot a gap the current check misses.

PR #1386, opened 2026-07-03
fix(server): cap in-flight pre-auth WebSocket sockets per IP

daxdax89:
## Summary

Security/DoS hardening from an audit. The `/ws` auth handshake performs several DB lookups (`accountForToken`, `moderationStatusForAccount`, `getCharacter`, `chatMuteStatusForAccount`) **before** the existing per-IP limit is reached, and that limit (`isConnectionRefused` / `countIpSessions`) only counts **established** sessions. So an unauthenticated client could open many `/ws` upgrades and force those lookups unbounded, with no pre-auth cap.

## Fix
Add a per-IP cap on concurrent **in-flight pre-auth** sockets:
- Reserved at the `upgrade` event, **before** `handleUpgrade` and the handshake. Over the cap → the socket is destroyed immediately.
- Released exactly once, whichever comes first: the socket closes/errors/times out (10s auth timeout), or it authenticates and becomes a counted session (so the pre-auth tally tracks only unauthenticated sockets; established sessions are covered by the existing session cap).
- The tally is a small, pure, unit-tested `PreAuthConnections` helper in `server/ip_block.ts` (never goes negative, drops the IP key at zero — no leak), mirroring the existing pure `isConnectionRefused` there.
- Cap is `MAX_WS_PREAUTH_PER_IP` (default 30), set above `MAX_WS_PER_IP_HARD` (20) so a legitimate client reconnecting several characters is never caught by it.

## Type of change
- [x] Bug fix (security / DoS hardening)

## How was this tested?
- New unit tests for `PreAuthConnections` in `tests/ip_block.test.ts`: acquire up to the cap then refuse without reserving, per-IP independence, release frees a slot, and release never goes negative / drops the key at zero (24 tests in the file pass).
- `npx tsc --noEmit` clean; `npm run build:server` clean (the `main.ts` wiring bundles).

## Notes for reviewers
- The release is idempotent and attached to the raw `socket` `close`/`error` (so it fires even if `handleUpgrade` fails before a `ws` exists) plus the auth-success path. I couldn't add a full upgrade-path integration test (no WS-upgrade harness in the suite), so the coverage is on the pure counter + the lifecycle is wired to fire on every terminal path; a manual review of the release points is worthwhile.

### DIFF
--- a/server/ip_block.ts
+++ b/server/ip_block.ts
@@ -61,3 +61,42 @@ export function isConnectionRefused(input: {
   if (input.isAdmin) return false;
   return input.blocked || input.ipSessions >= input.hardLimit;
 }
+
+// Per-IP cap on IN-FLIGHT pre-auth WebSocket sockets (upgraded but not yet joined
+// a game session). The existing session cap (isConnectionRefused / countIpSessions)
+// only counts ESTABLISHED sessions and is checked AFTER several DB lookups in the
+// auth handshake, so without this an IP could open many sockets and force those
+// lookups unbounded. A socket is acquired at the upgrade event and released when it
+// authenticates (becoming a counted session) or its socket closes/errors/times out.
+// Pure + unit-testable: no timers, no socket refs, just the per-IP tally.
+export class PreAuthConnections {
+  private counts = new Map<string, number>();
+  constructor(private readonly cap: number) {}
+
+  // Reserve a slot for `ip`. Returns false (and reserves nothing) when the IP is
+  // already at the cap, so the caller destroys the socket before the handshake.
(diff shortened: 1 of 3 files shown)


daxdax89:
Closing this. v0.22.0 refactored the WebSocket auth/upgrade path out of server/main.ts into a new server/ws_auth.ts (createWsAuth) with its own per-IP connection cap, so this PR's approach (an inline upgrade handler plus PreAuthConnections in main.ts) no longer fits the code. If the pre-auth (in-flight, not-yet-joined) cap is still wanted, it would be a fresh change against ws_auth.ts rather than this one.

PR #1387, opened 2026-07-03
feat(shell): WoW-style full-screen character screens

Blaine1705:
## What

Turns the three desktop character screens (offline create, online create, online roster) into a WoW-style full-screen experience:

- The 3D character preview fills the whole screen as the stage, with the idle auto-rotation removed so the model greets the camera (drag-to-spin stays).
- The 9 classes become a bottom rail of round portrait medallions with a gold ring on the selection and each class name in its class color.
- Appearance swatches dock as a left column (the race slot: one race, many looks), the class summary docks right, the name input floats top-center, and the actions sit in the corners (Back bottom-left, Enter World / Create bottom-right).
- The online roster docks the character list on the right, WoW style, with its actions at its foot; the account chrome (realm, sort, wallet) becomes a translucent top bar and the selected character's class summary docks left.
- While one of these panels is up, the landing chrome (sticky nav, footer, SEO copy) is parked so it no longer bleeds through the translucent stage (this also removes a faint Donate ghost that already showed on the old offline screen).
- The class summary type scales up one step (13px to 14px body, 24px class name) since the base sizing was tuned for the small pre-game card and read too small full-screen.

## Scope

- Desktop pointer only (min-width 900px, and the online arms are gated to body:not(.mobile-touch)): phones and touch bodies keep the existing card layout untouched.
- Pure CSS plus the cs-wow class on the three panels: all ids, buttons and rows stay untouched, so the main.ts wiring is unchanged.
- src/render/characters/preview.ts: removes the idle turntable spin only.

## Testing

- Real-browser screenshots of all three panels at 1600x900 (class switching, gold ring, skin column, corner actions, roster list dock).
- npx tsc --noEmit clean; client_shell, css_corpus, css_value_validity, styles_extraction, per_entry_css_wiring, backdrop_filter_survival and theme suites green (126 tests).

### DIFF
--- a/index.html
+++ b/index.html
@@ -914,7 +914,7 @@
           </div>
         </div>
 
-        <div id="offline-select" class="panel auth-panel auth-panel-premium" hidden>
+        <div id="offline-select" class="panel auth-panel auth-panel-premium cs-wow" hidden>
           <h2 class="auth-title" data-i18n="auth.offlineCharacter">Offline Character</h2>
           <div class="charselect-layout">
             <!-- Left Column: Character Name & Class Chips & Actions -->
--- a/index.html
+++ b/index.html
@@ -783,7 +783,7 @@
           </div>
         </div>
 
-        <div id="charselect-panel" class="panel charselect-screen" hidden>
+        <div id="charselect-panel" class="panel charselect-screen cs-wow" hidden>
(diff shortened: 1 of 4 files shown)


Blaine1705:
charsel

PR #1388, opened 2026-07-03
feat(render): distinct priest and mage default looks plus the priest halo

Blaine1705:
## What

Priest, mage and warlock share the mage.glb rig and looked near-identical (priest was just a cream tint). Each caster now reads as its own class at a glance:

- **Priest**: white linen robes with gold belt and buttons, and a soft golden halo ring floating above the head (the Light). The old cream tint is dropped: a material tint can only darken, never whiten, and it also warmed the face.
- **Mage**: night-indigo robe with muted gold trim, a matching dark hat, a shadowed cowl-dark hairline, and pale light-blue glowing eyes.
- **Warlock**: untouched, keeps his purple-tinted look and spellbook.

## How

- The default looks ride the existing skin pipeline: SKINS index 0 becomes a real recolored atlas (generated from the KayKit base atlas per UV cell, verified against the per-mesh UV tile map) instead of the embedded texture. Array lengths are unchanged, so the sim SKIN_COUNTS lockstep and saved skin indexes hold; the alt skins stay the shared mage set.
- The mage's eye glow uses the existing SKIN_EMISSIVE path (a black atlas with a light-blue eye tile), so it blooms on standard-material tiers and still reads on low (the eye tile is also painted in the base atlas).
- New VisualDef.halo (src/render/characters/halo.ts): a shared-cache additive canvas-texture ring parented above the head bone, angel style, excluded from applyMaterials so skin swaps never remap its material. Pure presentation: no sim, server or wire changes.
- Media manifest regenerated for the three new PNGs.

## Testing

- Real-browser verification: character-select preview and in-world screenshots for priest, mage and warlock (front and back, halo clear of the hat brim from every angle).
- npx tsc --noEmit clean; skin_event, appearance_skin and architecture suites green (45 tests).


Blaine1705:
mage
priest

PR #1391, opened 2026-07-03
tooling(ci): embed PR screenshots inline, and only for visual changes

EnriqueGF:
## Summary

Rework the PR screenshot bot so it (1) captures only when a change is actually visual and
only the sections it touches, and (2) embeds the frames inline in the sticky comment
instead of linking a downloadable artifact.

## Behavior

- **Visual-only, relevant-section capture.** A pure classifier (`classifyDiff` in
  `scripts/pr_shot_targets.mjs`) maps the changed-file list to:
  - specific windows (a bags change to the inventory window, a zone/terrain change to the
    world map), cropped to that window; or
  - a generic in-world HUD (desktop, plus mobile when the change touches the
    mobile/responsive surface) when the change is visual but maps to no specific window; or
  - nothing at all, for a backend/data/i18n-only diff.
  There is no fixed tour of unrelated screens anymore.
- **Inline embedding.** Artifacts are not embeddable and markdown does not render `data:`
  URIs, so `scripts/gh_image_host.mjs` uploads each PNG to a bot-owned orphan branch
  (`bot-pr-screenshots`) via the REST API and the comment references the raw URLs. The
  artifact upload is gone.
- Non-visual PRs post no screenshot comment; a prior gallery is flipped to a short "no
  visual changes" note via a new `updateOnly` option on the sticky-comment helper.

## Permissions and forks

The screenshots job gains `contents: write` (to host the images). On a fork PR the token
is read-only, so hosting and commenting degrade to a no-op / note, exactly as before.

## Tests and verification

- New unit test `tests/pr_shot_targets.test.ts` covers the classifier (backend-only,
  bags, zone, generic renderer, mobile CSS, i18n text, test files, precedence): 9 pass.
- Smoke-ran `pr_screenshots.mjs` locally on a non-visual diff: classifies as non-visual,
  captures nothing, exits 0, needs no browser (the browser import is now lazy).
- Smoke-ran `pr_comment_shots.mjs` with an empty manifest and no token: degrades cleanly.
- Biome clean on changed files; pre-push floor green.

The browser-driven capture mechanics (offline entry, clip, mobile emulation) are unchanged
from the previous script, so they keep their existing CI coverage; the new logic is the
pure classifier and the image host.

### DIFF
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -5,9 +5,10 @@ name: PR AI assist
 # review, it does not gate merge. Lives in its own workflow so it never interferes with
 # the CI gate in ci.yml.
 #
-#   screenshots        boots the Vite dev client headless, runs a fixed offline tour
-#                       (character select, desktop HUD, mobile HUD), uploads the PNGs as
-#                       an artifact, and links them in a sticky PR comment.
+#   screenshots        boots the Vite dev client headless and, ONLY when the diff has a
+#                       visual change, renders the sections it touches (a mapped window, or
+#                       the in-world HUD for a generic visual change), then embeds the PNGs
+#                       inline in a sticky PR comment. A non-visual diff captures nothing.
 #   ai-review           reviews the PR diff with the OpenAI Codex CLI, authenticated via
 #                       ChatGPT OAuth (the CODEX_AUTH_JSON secret holds a `codex login`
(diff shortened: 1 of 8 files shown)


EnriqueGF:
Folded into #1393 so the whole PR-bot rework lands as one change (both touched pr-ai.yml and docs/ai-pr-bot.md).

Issue #1365, opened 2026-07-03
Chat window UI breaks down as tabs are added, and the input text is misaligned


FernandoX7:
## Problem

The HUD chat window's design breaks down once more than the default set of tabs is present, and a couple of the window's controls do not behave correctly. Three separate problems:

1. **Tab bar overflows and looks broken.** With the default channels (Chat, Combat Log, World, Whisper, Yell, General) the tab row is already full, and the "+" add button wraps onto a second row on its own. Add one more channel (for example a Party tab) and the layout wraps further and looks disjointed rather than staying a single clean, contained tab strip.

2. **The add-channel picker (the "+" button) cannot be dismissed.** Clicking "+" opens the channel picker, but clicking "+" again does not close it, and clicking elsewhere on the window or outside it does not dismiss it either. There is no obvious way to back out once it is open.

3. **The text input box does not fit its text.** The message entry box is too tall for a single line and the text (and the caret when the box is empty) is not vertically centered, so typed text sits misaligned inside the box.

The chat window is a core, always-visible part of the HUD, so it should stay beautiful and functional no matter how many tabs are open.

## Expected behavior

- The tab strip stays a single clean, contained row as tabs are added: it does not wrap awkwardly, and the "+" button stays inline with the tabs rather than dropping to its own row.
- When there are more tabs than fit, they degrade gracefully (for example horizontal scroll or an overflow control) while keeping the classic look.
- Tabs can be dragged to reorder and sort them, and the chosen order persists.
- The "+" channel picker toggles closed when "+" is clicked again, and also dismisses when the user clicks anywhere outside it.
- The message input box is sized for one line of text with the text and caret vertically centered.

## Steps to reproduce

1. Open the chat window in game.
2. Add channels via the "+" button until the tab row is full (Chat, Combat Log, World, Whisper, Yell, General, then one more such as Party).
3. Observe the tab bar wrapping and the "+" button on its own row.
4. Click "+" to open the channel picker, then click "+" again, then click elsewhere on or outside the window.
5. Click into the message input box and type some text (or leave it empty and look at the caret).

## Actual behavior

- The tab row wraps to a second line and the "+" button sits alone on its own row; the window looks broken and unpolished as tabs are added.
- Tabs cannot be dragged to reorder.
- The "+" channel picker stays open: clicking "+" again does not close it, and clicking around the window or outside it does not dismiss it.
- The input box is too tall and the text/caret is not vertically centered.

## Scope

[...]

PR #1394, opened 2026-07-03
fix(chat): tab strip stays one row, add-channel picker dismisses

daxdax89:
## Problem

Part of #1365. Two of the chat window's problems are layout/interaction bugs on the tab strip and the add-channel picker:

1. The tab strip used `flex-wrap: wrap`, so once more than the default channels were open the tabs wrapped to a second row and the "+" button dropped onto its own line. The window looked broken as tabs were added.
2. The "+" channel picker could not be dismissed: clicking "+" again did not close it, and clicking anywhere else (on or outside the window) did not either. Only Escape or selecting an item closed it.

## What this changes

- **Tab strip stays a single clean row (`fix(chat)` commit 1).** `#chatlog-tabs` is now `flex-wrap: nowrap` and scrolls horizontally when the tabs overflow (scrollbar hidden for the classic look), the tabs never squish (`flex: 0 0 auto`), and the "+" stays inline and reachable via `position: sticky; right: 0`. A vertical wheel over the strip is translated to horizontal scroll (bound once) so a plain mouse can reach overflowed tabs, and it no-ops until the strip actually overflows.
- **Picker toggles and dismisses (`fix(chat)` commit 2).** A document `pointerdown` closes the shared `#ctx-menu` on any click outside it, "+" is now a proper toggle (a second click closes the picker it opened), and the tracked opener is cleared on every close path so the outside-click defers to the opener's own toggle. This also gives every right-click context menu the same click-away dismissal. Escape still closes via the existing `closeAll` dispatcher.

## Scope

Focused on the tab strip and the picker. The other two items in #1365 are intentionally out of scope here to avoid duplicated work:

- The message-input sizing/centering is already handled by #1234 (`fix(ui): size chat input for placeholder text`).
- Drag-to-reorder tabs with persisted order is a separable feature and will land as a follow-up PR.

## Verification

Driven in the real offline game (`npm run dev`), warrior in Eastbrook Vale, with 8 open tabs (all, combat, world, lfg, yell, general, whisper, +):

- Tab strip: `barHeight = 22px` (one row, not wrapped), all tab tops on a single row, strip overflows (`scrollWidth 528 > clientWidth 364`) and scrolls, `flex-wrap: nowrap`, `overflow-x: auto`, "+" on the first row with `position: sticky`.
- Picker: click "+" opens, second "+" click closes (toggle), click outside closes, Escape/`closeAll` closes. All seven state transitions asserted.
- `tsc --noEmit` clean; `css_corpus` / `css_value_validity` / `styles_extraction` guard suites green; biome clean on the changed files; no new i18n keys (the picker reuses existing `t()` keys).

Screenshots (offline game): the tab strip staying one row with the "+" pinned, and the add-channel picker open, attached below.

## Checklist

- [x] Targets `release/v0.20.0`
- [x] `tsc` + guard suites + biome (changed files) green
- [x] No em/en dashes, emoji, or attribution lines
- [x] UI change verified in the offline game (desktop); mobile chat tab strip unaffected (the mobile block only sets the strip cursor)
- [x] No new player-visible strings (existing `t()` keys reused)

### DIFF
--- a/src/styles/hud.css
+++ b/src/styles/hud.css
@@ -1522,14 +1522,26 @@
   }
   #chatlog-tabs {
     display: flex;
-    flex-wrap: wrap;
+    /* Stay a single clean row as tabs are added: never wrap the "+" to its own
+       line. When the tabs outgrow the strip they scroll horizontally instead
+       (the scrollbar is hidden for the classic look; wheel-to-horizontal is
+       wired in renderChatTabs, and the "+" stays pinned via position: sticky). */
+    flex-wrap: nowrap;
+    overflow-x: auto;
+    overflow-y: hidden;
+    scrollbar-width: none;
     gap: 2px;
     margin-left: 6px;
     margin-bottom: -1px;
     position: relative;
     z-index: 2;
   }
+  #chatlog-tabs::-webkit-scrollbar {
+    display: none;
+  }
   .chat-tab {
+    /* keep each tab its natural width and let the strip scroll, never squish */
+    flex: 0 0 auto;
     height: 22px;
     padding: 0 12px;
     border: 1px solid #3a321d;
@@ -1547,8 +1559,12 @@
     border-color: #6f5a2a;
     background: linear-gradient(#2a2436, #161220);
   }
-  /* "add channel" (+) tab: square, icon-only, slightly de-emphasised */
+  /* "add channel" (+) tab: square, icon-only, slightly de-emphasised. Sticks to
+     the right edge so it stays inline and reachable even when the tabs scroll. */
   .chat-tab-add {
+    flex: 0 0 auto;
+    position: sticky;
+    right: 0;
     padding: 0;
     width: 26px;
     font-size: 15px;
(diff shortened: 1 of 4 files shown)


daxdax89:
Thanks for the review. Addressed the mobile reachability point: `touch-action` is now handled in CSS instead of an inline `none` on `#chatlog-tabs`. The chat-box move gesture on the strip is desktop-only (`onChatBoxMoveStart` bails on `isMobileLayout()`), so the strip is `touch-action: none` on desktop for the move handle and `pan-x` on mobile, which keeps overflowed tabs swipeable on phones (the scrollbar stays hidden for the classic look). Pushed as beb97b98.

daxdax89:
Added `tests/chat_tab_strip.test.ts` to close the second nit: a source guard pinning the load-bearing declarations (`#chatlog-tabs` flex-wrap: nowrap + overflow-x: auto, `.chat-tab-add` position: sticky, and touch-action none desktop / pan-x mobile), so a future edit that reintroduces the wrap or blocks the mobile pan fails CI. The interactive picker toggle / outside-click path is harder to cover without instantiating the full HUD DOM; happy to add a focused fake-DOM test if you would like it in this PR.

PR #1400, opened 2026-07-04
fix(admin): forbid force-renaming an admin's character

daxdax89:
## Problem

Admin accounts are protected from moderation actions: ban and suspend refuse an admin account (`server/admin.ts`, "admin accounts cannot be suspended or banned"), and chat-mute does the same. But **force-rename did not**. `forceCharacterRename` looks up the target character's account and proceeds without checking `is_admin`, so a moderator could force-rename (and, via the handler, disconnect) an **admin's** character. It is the one moderation action missing the admin guard.

## Fix

Extend the existing account lookup in `forceCharacterRename` to also read `accounts.is_admin`, and reject an admin target before opening the transaction, matching the other moderation actions:

```
if (row.is_admin === true) throw new Error('admin characters cannot be force-renamed');
```

The guard runs on the initial `SELECT`, so no rename transaction is opened for a protected target. The new error string is added to the admin dashboard i18n (English catalog + the server-error reverse map), so the operator sees a localized refusal exactly like the sibling actions; the 20 locale overlays are left for the maintainer's release fill (build English-fills them, marked `pending`).

## Test

Test-first: a new `moderation_db` case asserts a moderator cannot force-rename an admin character and that no transaction is opened. It fails on the pre-fix code (the rename proceeds) and passes with the guard.

## Verification

- `tsc --noEmit` clean.
- `moderation_db` (23), `i18n_admin_catalog` (22), plus `moderation_service` / `moderation_game` / admin action suites green.
- `npm run i18n:gen` is idempotent (the resolved-table reproducibility gate passes) and biome is clean on the changed files.

## Notes

From our read-only repo audit (item M06). Additive and back-compat: no schema change (reads the existing `accounts.is_admin`), no wire change, English-only i18n add.


daxdax89:
Closing. This targeted the shipped release/v0.20.0, and v0.22.0's admin-auth overhaul (role-based; is_admin is now a derived staff flag) changes the surface this guarded, so it would need reworking rather than a rebase. Deprioritizing for now.

PR #1401, opened 2026-07-04
fix(admin): only unban an account that is actually banned

daxdax89:
## Problem

`moderateAccount`'s **unsuspend** branch guards on a standing suspension (`WHERE ... suspended_until > now()`) and throws `account is not suspended` when nothing matched. The **unban** branch did not: it ran an unconditional `UPDATE ... WHERE id = $1`. So unbanning an account that is not banned:

- silently "succeeds" (clears already-null `banned_at`), and
- still runs the shared `player_reports` resolution below (unban is not `unsuspend`), which marks **every open report against that account as `actioned`**.

That last part is the real bite: an admin who clicks Unban on a non-banned, reported account silently dismisses all open reports against it.

## Fix

Guard the unban UPDATE on `banned_at IS NOT NULL` and throw `account is not banned` when no row matched, mirroring unsuspend. The throw happens inside the transaction, so it rolls back before the audit row or the report resolution runs. Adds the admin-dashboard error string (English catalog + server-error reverse map, like `account is not suspended`); the locale overlays are the maintainer's release fill.

## Test

Test-first. A new case asserts unbanning a non-banned account rejects with `not banned`, rolls back, and runs no `player_reports` UPDATE. The existing unban test now asserts the `banned_at IS NOT NULL` guard clause. Both fail on the pre-fix code and pass with the guard.

## Verification

`tsc` clean; `moderation_db` (23) + `i18n_admin_catalog` (22) green; `npm run i18n:gen` idempotent (reproducibility gate passes); biome clean on the changed files.

## Notes

From our read-only repo audit (item M05). Additive and back-compat: no schema/wire change, English-only i18n add. Sibling to #1400 (the force-rename admin guard) but a distinct concern, kept as its own PR.


daxdax89:
Closing. Same as #1400: targeted the shipped release/v0.20.0 and predates v0.22.0's admin-auth overhaul. Deprioritizing; can revisit the unban guard against the new model if wanted.

PR #1430, opened 2026-07-04
fix(hud): stop the mailbox indicator overlapping the raid-lockout badge

ryan-foo:
## Summary

On the minimap, the Ravenpost mailbox indicator overlapped the raid-lockout badge. Both were absolutely positioned in the same bottom-left corner of `#minimap-disc` (`#raid-lockout` at `left/bottom: 10px`, `#mail-indicator` at `left/bottom: 4px`), so whenever unread mail was waiting the envelope pill sat on top of the raid-lockout (raid timer) button.

The fix wraps both rim badges in a `#minimap-notifiers` flex rail anchored to that corner. They now lay out in a row with a `6px` gap and cannot overlap. Because hidden badges collapse out of the flow, the row stays gapless and any future notifier slots in beside the others. The rail is click-through (`pointer-events: none`) and each badge re-enables its own pointer events, so the minimap stays clickable in the gaps. Applied to both game entries (`index.html` and `play.html`); CSS/HTML only, no JS change (both badges are still resolved by id).

## Related issues

N/A (reported in Discord).

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands: `npx vitest run tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/styles_extraction.test.ts tests/per_entry_css_wiring.test.ts` (34 passing); `npx @biomejs/biome check src/styles/hud.css index.html play.html` (clean).
- Manual steps: rendered the minimap bottom-left rim from the shipped `src/styles/hud.css` rules in both the pre-fix and post-fix state (see below). Regenerate with `node scripts/minimap_notifier_shot.mjs`.

## Screenshots / recordings

Rendered from the actual `hud.css` rules (the game applies the identical rules to `#minimap-notifiers`).

Before (envelope overlapping the raid-lockout badge):

before

After (both badges laid out cleanly in the rail):

after

---

### Quality

- [x] Tests pass (the CSS guard suite; no `sim/`/`server/` behavior changed).
- [x] Typecheck passes (no TypeScript touched).
- [x] Builds succeed (CSS/HTML only).

### Cross-platform

- [x] Both game entries (`index.html`, `play.html`) updated together. The rail uses the same rim anchor on desktop and mobile; no `hud.mobile.css` override existed for these ids.
- [x] Accessible: no change to the badges' roles/labels/focus; the mail indicator keeps `role="status"` and the raid-lockout keeps its button semantics and tooltip.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (layout only).

### Hygiene

- [x] No generated files hand-edited. Screenshots are committed proof under `docs/screenshots` (regenerable via the script).


FernandoX7 (commented):
Clean, well-scoped fix, and the description matches the code. The overlap is real on this base (`#raid-lockout` at `left/bottom: 10px`, `#mail-indicator` at `left/bottom: 4px`, both in the same bottom-left corner of `#minimap-disc`), and the flex-rail wrapper resolves it without touching any JS: both badges are still resolved by id, both entries (`index.html` + `play.html`) get the wrapper, and the pointer-events handoff checks out (`#minimap-notifiers` is `pointer-events: none`, `#raid-lockout` gains `pointer-events: auto`, `#mail-indicator` already had it). Making the row gapless via `[hidden]` collapse is a nice detail.

The one thing to sort out before this lands:

**should-fix: this same overlap is already fixed on `main` by #1436, a different and incompatible way.** #1436 (merged) repositions `#mail-indicator` to the bottom-right rim (`right: 4px; bottom: 4px`) and ships a regression guard, `tests/minimap_rim_badges.test.ts`, that asserts each rim badge anchors to exactly one of left/right and one of top/bottom, and that the two badges never share a corner. This PR goes the other direction: it drops `left/right/top/bottom` from both `#raid-lockout` and `#mail-indicator` entirely (they become static flex children in `hud.css`). So once `release/v0.20.0` forward-merges into `main` you get a hard conflict on both the HTML and those two CSS rules, and `minimap_rim_badges.test.ts` goes red: with no insets on either badge, `anchoredCorner`'s `expect(horizontal).toHaveLength(1)` fails. The guard is not on `release/v0.20.0` yet, which is why this PR's own CSS-guard run is green.

So this needs a maintainer call rather than a code change:

- If #1436's bottom-right fix stands, this PR is superseded and can be closed.
- If the flex-rail is preferred (it is the more scalable shape, a new notifier just slots into the row), then it should also update `tests/minimap_rim_badges.test.ts`, since that guard currently encodes the exact opposite invariant (one anchored corner per badge) that the rail deliberately abandons.

Not a knock on the work: the rail is the nicer pattern. It just arrived after the same bug was already closed on `main`.

Everything else looks good. i18n is untouched (the `data-i18n-*` attributes moved, they were not changed), and a11y is preserved (the `role=status` on the mail badge and the button semantics both survive; the new wrapper is presentational).

Rubsey:
I merged in a fix for this last night that's waiting to be deployed.

PR #1434, opened 2026-07-04
feat(shaman): Chain Heal, the shaman healer signature spell

Blaine1705:
## What

Adds **Chain Heal** (`chain_heal`), the shaman healer signature spell, learned at level 18.

It heals the friendly target, then arcs to up to 2 nearby allies (3 targets total), each hop healing half the previous amount, so it heals most on the main target and less on each bounce. Classic-faithful where it counts: nature school, 2.5s cast, and a 0.5 per-hop falloff (the exact vanilla 50%-per-jump). The 3-target cap is the classic number too. The one deliberate divergence from classic is the jump range: 18 yards (classic was ~10), chosen by the maintainer to make the bounce more forgiving.

## The visual

A bright green healing arc connects each hop: a new `fx: 'chainHeal'` value on the visual-only `spellfx` SimEvent drives a dedicated `vfx.chainHealArc` (a green cord that lifts in a gentle parabola between allies), and the per-target heal burst lands via the existing `heal2` event. Verified end-to-end in a real browser cast.

## Sim / correctness

- New `chainHeal` AbilityEffect handled in `combat/effect_dispatch.ts` behind the `runEffects` seam. The ally selection is fully deterministic (most injured by hp fraction, then distance, then id) with no rng in the selection loop, so only the base roll and each `applyHeal` crit draw rng. Reviewed for determinism (parity draw-order digest green) and three-host parity.
- The heal resolves authoritatively inside `Sim`; only the visual event crosses the wire, and a new `fx` string needs no wire-schema change (spellfx is a whole-JSON world event).

## i18n

EN catalog entry plus Spanish and the five non-Latin locales (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU) translated; the build English-fills the remaining locales (maintainer fills at release).

## Notes for review

- Parity goldens, i18n resolved tables, the status summary, the resolved-hash, and the guide content were all regenerated on this branch (not cherry-picked), so they reflect release-based state. The three locale re-translations the release introduced for `ghost_wolf`/`stormstrike` were preserved during the merge.
- Balance (heal magnitude, extra jumps) is intentionally left where it is; a restoration talent to add a 4th jump is planned as follow-up.

## Testing

tsc, the new `tests/chain_heal.test.ts` (multi-hop chain, no-ally fallback, replay determinism), the parity suite, the architecture purity scan, and the i18n guard suite (membership, coverage, freshness, admin catalog, completeness) are green; the pre-push floor passed.


Blaine1705:
https://github.com/user-attachments/assets/1e4f7fb2-e5cb-4ebf-b896-b335ac76daf6

Blaine1705:
Update: added a custom cast sound for Chain Heal.

The cast previously fell through to the earthy `cast_nature` loop (meant for the shaman damage spells), which read as fire/stones on a heal. It now plays a dedicated one-shot healing water clip (`cast_chain_heal`), wired in the `castStart` handler for `chain_heal` only; nothing else changes.

Heads-up for the audio pipeline, one thing to decide on your side:

- The clip is a player-provided recording, not an ElevenLabs-generated catalog sound, so it is committed directly (`public/audio/sfx/cast_chain_heal.wav`) with a hand-added entry in `src/game/sfx_manifest.generated.ts`. That is the one exception to the "do not edit by hand" manifest.
- It is a WAV (about 480 KB) because it was extended in-browser (no ffmpeg locally) to cover the 2.5s cast; a regen via `scripts/gen_sfx.mjs` would drop the entry since it is not in `scripts/sfx/sfx_prompts.mjs`.
- Suggestion: either keep it as-is, or fold it into the pipeline (re-encode to mp3 plus a catalog entry). The sound is player-approved, so please preserve the clip rather than regenerating an AI replacement.

Blaine1705:
https://github.com/user-attachments/assets/05c8cdf1-3f1b-4585-bdec-8d54aa48871f

Sound effect added

jgyy:
Closing as superseded: Chain Heal is already implemented in src/sim/content/classes.ts on main. Thanks for the contribution.

PR #1437, opened 2026-07-04
fix(ui): show a skip hint during the first-spawn intro cinematic

jgyy:
## Summary
The first-spawn intro cinematic hides the whole HUD (and, on touch, the joysticks and action buttons) for up to 9 seconds with no visible way out:

- Desktop players can guess **Escape** from genre convention.
- Touch players have no keyboard, so the only skip gesture is a **4-tap burst within 1.5s** (`spawn_cinematic.ts`), and nothing on screen says so.

On a phone this reads as the game being frozen: the whole screen goes blank of UI, movement is swallowed, and there is no hint that four quick taps anywhere will skip it.

This adds a small "Tap 4 times to skip" / "Press Esc to skip" hint, shown only for the span the HUD is hidden, on both platforms.

```mermaid
sequenceDiagram
    participant P as Player (new character)
    participant M as main.ts
    participant H as #intro-skip-hint

    P->>M: enterWorld (level <= 1, intro not yet seen)
    M->>M: setIntroUiHidden(true)
    Note over M: #ui, #mobile-controls, nameplates -> display:none
    M->>H: text = touch ? skipHintTouch : skipHintDesktop
    M->>H: classList.add('visible')
    loop up to 9s camera glide
        P-->>M: Escape (desktop) or 4-tap burst (touch)
        alt skip gesture matched
            M->>M: finishIntro(true)
        else cinematic reaches its end pose
            M->>M: finishIntro(false)
        end
    end
    M->>M: setIntroUiHidden(false)
    M->>H: classList.remove('visible')
```

### Before / after
| | |
|---|---|
| Mobile (touch) | mobile skip hint |
| Desktop | desktop skip hint |

## Changes
- `src/main.ts`: `setIntroUiHidden` now also toggles `#intro-skip-hint`, picking touch vs desktop copy via `useTouchInterface()`.
- `index.html` / `play.html`: add the `#intro-skip-hint` element (both game entries share `main.ts`).
- `src/styles/hud.css`: styling for the hint, a top-level layer outside `#ui` (so it survives the HUD being hidden), bottom-center, safe z-index above the touch controls.
- `src/ui/i18n.catalog/hud_chrome.ts`: new `hudChrome.intro.skipHintTouch` / `skipHintDesktop` keys, plus their five non-Latin fills (M16: both strings are wordy) and the regenerated resolved/status artifacts.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run` (full suite green; one pre-existing unrelated flake in `mail.test.ts` passed on rerun)
- [x] `npm run i18n:build` / `i18n:scan` / `i18n:hash -- --write` regenerated and committed
- [x] `npm run build` (production bundle) + manual puppeteer capture confirming the hint shows the right copy on both a phone-landscape touch viewport and a desktop viewport
- [x] Biome (`--changed`-equivalent) clean on touched files; pre-push QA hook green


jgyy:
closed due to personal taste issue

PR #1440, opened 2026-07-04
feat(mobile): declutter the touch HUD chrome

jgyy:
## Summary
Redesigns the touch HUD chrome for mobile: a corner-anchored, unobtrusive
control scheme instead of the old dead-center horizontal hotbar/joystick
layout (researched against Genshin Impact, Honkai: Star Rail, Wuthering
Waves, Arknights: Endfield, Wild Rift, Diablo Immortal). Every layout choice
below was verified by actually driving the touch HUD in a phone-landscape
headless Chromium session (not just reading the CSS diff), checking every
button's `getBoundingClientRect()` for zero geometric overlap.

- **Corner-anchored combat cluster.** Attack (64px, the single most-pressed
  button, slightly bigger than its neighbors)/Jump/Target/Use live in one
  thumb-reach corner above the move joystick (mirrored for left-handed), not
  a wide dead-center strip.
- **5-slot paged hotbar.** A row of 5 spell buttons sits above the combat
  cluster: page 1 is hotbar slots 1-5, page 2 is slots 6-10, and one
  `#mobile-hotbar-page` toggle switches between them. Hidden-page slots stay
  fully live (cooldowns keep ticking, aura/proc state is untouched); paging
  only toggles which buttons are painted. Drag-to-reorder is unaffected.
- **Own HP/mana lives bottom-center** -- the one part of the screen every
  other element deliberately leaves empty -- instead of top-left, so the
  thumb never has to look away from center to check vitals and the corners
  stay clear.
- **Camera joystick removed**; swipe-drag on the canvas (already
  implemented) provides full camera control, so the second control surface
  was redundant. Its double-tap-to-recenter gesture moved to the swipe-look
  path. The move joystick shrank from 122px to 76px.
- **Low-frequency verbs (Autorun/Chat/More)** moved to a small corner column
  on the joystick's opposite side, with a clean 16px gap between them.
- **Icon-only, translucent glass chrome** for the action row, More tray, and
  minimap-side quick-access rail (community toggle + daily-rewards chest),
  with idle-fade after 3s of no touch (never dims Attack, joysticks, or any
  actionable info, per the graphics-fairness rule). The More tray's buttons
  are true pills (radius capped at half their height, not the base rule's
  `50%`, which squashes into an oval on a wide row), sized down to a compact
  440px-wide sheet, with centered icon+label.
- Every edge-anchored element gets extra padding to clear real-device camera
  cutouts.
- Fixed a real pre-existing bug found while touching this code: the
  landscape breakpoint's `#mobile-move-joystick` ID selector outranked the
  class-based left-handed swap rule on specificity, so the joystick silently
  never swapped sides in landscape. Fixed with a matching-specificity
  override.

### Default layout (page 1) vs page 2
| Page 1 (slots 1-5) | Page 2 (slots 6-10) |
|---|---|
| after | page2 |

### Left-handed mirror and the More tray
| Left-handed | More tray |
|---|---|
| left-handed | more tray |

### Idle-fade, and the targeted mob's frame
Targeting keeps its own top-left frame (unchanged) -- it now has that whole
corner to itself since own HP/mana moved out to bottom-center, so nothing
stacks: target frame top-left, own HP/mana bottom-center, combat cluster +
hotbar bottom-right, joystick bottom-left.

| Idle (3s, no touch) | Targeted mob |
|---|---|
| idle fade | target frame |

## Test plan
- [x] `npx vitest run` (full suite; the handful of failures under full-suite
      load are pre-existing flaky timeouts, confirmed passing individually,
      unrelated to this change)
- [x] `npx vitest run tests/client_shell.test.ts tests/mobile_controls.test.ts tests/mobile_chrome_fade.test.ts tests/css_corpus.test.ts tests/backdrop_filter_survival.test.ts tests/css_value_validity.test.ts tests/styles_extraction.test.ts`
- [x] `npx tsc --noEmit`
- [x] `npx @biomejs/biome ci --changed`
- [x] `npm run build` (the `check_backdrop_survival.mjs` production gate)
- [x] Manually drove the touch HUD end-to-end at level 20 (many spells) in a
      phone-landscape headless Chromium session: default layout, both
      hotbar pages, left-handed mirror, the More tray, and idle-fade
      dim/restore, confirming zero button overlap in every state.
- [x] Rebased onto the current `release/v0.22.0`; zero merge conflicts.


jgyy:
Thanks for all the feedback through the rounds on this one, the touch HUD landed in a much better spot for it (vertical/orbital/paged hotbar experiments, corner-anchoring, and the two real overlap bugs that came out of it along the way).

Rebased and green on the current release branch. Flagging this for consideration in #1474 (v0.22.0).

jgyy:
Fixed two more things from the latest screenshots:

- **The corner column (chevron/chat/"...") was visually touching/overlapping.** The 48px pitch on 44px-tall buttons left only a 4px gap, which read as overlapping once the glass blur/glow was applied. Widened to 60px pitch (16px clear gap).
- **The More tray buttons were an oval, not a pill, and their content was left-aligned.** They inherited the base `.mobile-btn` rule's `border-radius: 50%`, which on a wide flat row computes squashed elliptical corners rather than a true pill (semicircular ends, straight sides). Capped the radius at half the button height and centered the icon+label.

Same `pr-assets-mobile-hud-declutter` branch, same file names -- screenshots above are updated in place.

FernandoX7 (commented):
This is careful work and it shows. The idle-fade is extracted into its own `mobile_chrome_fade.ts` with injected timers and a real unit test (threshold boundaries, touch reset, dispose), index.html and play.html stay in lockstep, the new `hudChrome.mobile.hotbarPage/hotbarPageAria` keys are English-only in the right catalog and reached only via `title`/`aria-label` (the visible label is hidden), and the `client_shell` pins were rewritten to literal geometry instead of left stale. The specificity fixes (the left-handed joystick swap in landscape, the 2-ID `#mobile-hotbar-page` sizing, the missing `pointer-events: auto`) are all real and correctly reasoned, and the idle-fade selector list stays clear of the fairness-protected info (Attack, joysticks, HP/target/cast bars). `tsc`, the mobile and CSS guard tests, and `biome ci` on the changed files are green here.

One thing the overlap sweep missed, worth fixing:

**should-fix** Own cast bar overlaps the relocated bottom-center player frame on narrower landscape phones. `#player-frame` moved to `left: 50%` bottom-center (`src/styles/hud.mobile.css:2178`, the landscape block that applies in-game) while `#castbar` stays bottom-left (`left: calc(... + 112px)`, `width: min(160px, calc(50vw - 166px))`, `src/styles/hud.mobile.css:2135`). Measuring the two boxes in headless Chrome across landscape widths, they overlap horizontally for everything below ~800px: 49px at 667px (iPhone SE), 56px at 568 to 640px, 22px at 720px, gone by 800px. Player frame is `z-index: 21` vs castbar `19`, so while casting the right ~30% of your own cast bar renders behind the portrait/vitals. The overlap check drove buttons through `getBoundingClientRect`, so it never caught the cast bar (only present mid-cast) against the moved frame. Notched insets shift the castbar further right, so this is a lower bound at a given width.

Nits:

- The `#mobile-hotbar-page` "Skills" toggle is always shown, so a low-level character with 5 or fewer spells can page to an empty page 2 that displays nothing. For the friendliness goal, consider hiding the toggle when the second page has no bound slots.
- "Camera joystick removed entirely" is really "hidden": `#mobile-camera-joystick` still ships in both index.html and play.html, and `mobile_controls.ts:319` still binds `onCameraDown`/`Move`/`End` plus their ~70 lines of handler code, now unreachable (a `display: none` element gets no pointer events). A follow-up could delete the element and the dead handlers.
- The paging logic itself (page arithmetic, ring assignment, hidden-slots-stay-live, slot 11 always hidden) has no behavioral test; `client_shell` only string-pins that the methods exist. A small `Hud` test that flips the page and asserts which slots carry `mobile-hotbar-page-hidden` / `hotbar-ring-N` would lock the headline feature.

jgyy:
Fixed the should-fix from the last review: own cast bar overlapping the bottom-center `#player-frame` on narrower landscape phones.

The landscape `#castbar` width was capped only against `50vw` (`min(160px, calc(50vw - 166px))`), with no accounting for `#player-frame`'s own half-width once it moved to bottom-center at `scale(0.6)` (90px half-width). Recomputed the cap as `50vw - 238px` (50vw minus the frame's 90px half-width minus a 16px clear gap), with `min-width` dropped from 100px to 40px so the floor itself can't reintroduce the overlap on very narrow widths. Verified against the exact widths from the review:

| Width | Before | After |
|---|---|---|
| 568-640px | 56px overlap | clear (16px gap) |
| 667px | 48.5px overlap | clear (16px gap) |
| 720px | 22px overlap | clear (16px gap) |
| 800px+ | clear | clear |

Also added `scripts/mobile_castbar_frame_shot.mjs`, a small headless-Chrome regression check that targets a mob, starts a real Fireball cast (a spell with an actual cast time, not a cheat) at 640px landscape (inside the reported overlap band), and asserts `#castbar` and `#player-frame` don't overlap.

Replaced `mobile-hud-target-frame.png` above with a fresh screenshot taken mid-cast at 640px showing the cast bar (bottom-left) and player portrait/vitals (bottom-center) now cleanly separated.

`npx tsc --noEmit`, the mobile/CSS guard suite, and `biome check` on the changed files are all green; pre-push QA floor passed.

FernandoX7 (approved):

PR #1445, opened 2026-07-04
feat(render): bolt-shaped electric projectile for the shaman Lightning Bolt

Blaine1705:
## What

Replaces the shaman Lightning Bolt (`lightning_bolt`, "Arc Bolt") cast visual: instead of the generic round glowing projectile ball, it now fires a **bolt-shaped electric projectile**, an original procedural effect.

- Same mechanic and timing as any spell projectile: the bolt travels and the damage resolves on impact (no change to combat, hit/resist, or scheduling).
- The flying head renders as a short jagged blue-white electric streak (zig-zag segments trailing along the travel direction, plus crackle sparks) rather than a smooth comet, and the impact is tinted electric blue.

The effect is entirely original procedural particle code built on the existing VFX system; no external art or assets.

## How

- New `AbilityDef.projectileFx?: 'lightning'` (renderer-only visual override; the projectile MECHANIC is untouched). `lightning_bolt` sets it.
- `casting_lifecycle.ts` emits `fx: ability.projectileFx ?? 'projectile'` at the existing projectile-fire site, so only the visual cue changes.
- New `fx: 'lightning'` value on the visual-only `spellfx` SimEvent, handled by `vfx.lightningProjectile` in the renderer. The projectile update loop draws the jagged streak when the projectile carries the `lightning` flag.
- The `fx` string rides the wire like every other `spellfx` cue (whole-JSON world event), so there is no wire-schema change and it renders identically offline and online.

## Notes for review

- Renderer-driven visual; the sim change is just the new `projectileFx` field plus forwarding it on the event. Parity goldens were regenerated on this branch (the solo scenarios now emit `fx:'lightning'` where they emitted `fx:'projectile'`).
- Heads-up on a small overlap: this touches the same `spellfx` `fx` union / renderer `handleEvent` switch as the open Chain Heal PR (#1434). They are independent features; whichever merges second will need a one-line merge of the `fx` union and the switch. This branch deliberately does not include any Chain Heal code.

## Testing

tsc, the new `tests/lightning_bolt_fx.test.ts` (asserts the `fx:'lightning'` emit and that damage still lands), the parity suite, the architecture purity scan, and the HUD perf budget are green; the pre-push floor passed.


Blaine1705:
Update: added a custom cast sound for Lightning Bolt alongside the visual.

Lightning Bolt's cast previously used the shared earthy `cast_nature` loop. It now plays a dedicated storm clip (`cast_lightning_bolt`), routed via `castKeyForAbility` so it uses the normal cast-loop path: it fills the whole cast at any rank (1.5s to 3s) and unloops cleanly on interrupt. Nothing else changes.

Same audio-pipeline heads-up as the Chain Heal PR:

- The clip is a player-provided recording (not an ElevenLabs-generated catalog sound), committed directly (`public/audio/sfx/cast_lightning_bolt.wav`) with a hand-added entry in `src/game/sfx_manifest.generated.ts` (the one exception to the "do not edit by hand" manifest).
- It is a WAV because it was looped in-browser (no ffmpeg locally); a regen via `scripts/gen_sfx.mjs` would drop the entry since it is not in `scripts/sfx/sfx_prompts.mjs`.
- Suggestion: keep it as-is, or fold it into the pipeline (re-encode to mp3 plus a catalog entry). The sound is player-approved, so please preserve the clip rather than regenerating an AI replacement.

Blaine1705:
https://github.com/user-attachments/assets/ce45a372-403a-4ff7-8cd0-d1ec0e9a0a58

PR #1446, opened 2026-07-04
feat(world-boss): make Thunzharr unkitable (Howling Gale snare) + 90-minute respawn

ryan-foo:
## Summary

Reworks Thunzharr, the Waking Peak (the Thornpeak world boss) into a proper world-boss encounter:

- **Unkitable (Howling Gale).** The boss's move speed (5.8) is below player run speed (7), and its pulses (Thunderclap, Seismic Stomp, Stormcall) only fire in melee range, so a hunter could hold it in the chase state and kite it to death. A new template-driven anti-kite snare fires from BOTH the chase and attack states: every 5s it slows every player within 40yd to 20% move speed for 6s (`moveSpeedMult` already honors `slow` auras). A snared kiter moves 1.4yd/s while the boss moves 5.8, so the boss closes and its other mechanics come online. No damage, no rng.
- **Aggressive participant HP scaling.** Spawns at 40k HP and gains **40k per extra participant** on its hate table, capped at **1M** (~25 players), so a crowd can't melt it in a minute. It only grows the pool within a spawn (a member dying doesn't make it easier); a fresh spawn resets to 40k.
- **Towering size.** `scale` 1.7 -> **5** (Nythraxis-class), so the boss is unmistakable. This also widens his single-target melee reach to ~17yd, reinforcing the anti-kite.
- **Centered, self-aggroing summons.** The stormling adds now erupt **centered on the boss** (a tight 1yd cluster from underneath him) instead of a 3.5yd ring, and immediately charge his target (with a nearest-player fallback so a fresh wave always has someone to hit). They leash to the boss's ORIGINAL spawn, so kiting the fight too far sends them running home. Scoped to world bosses; ordinary summoners keep their ring.
- **Loud.** Every yell carries 350yd (past the 100yd default), and he bellows one of four battle cries every 9s in combat (cycled, no rng).
- **Hourly respawn.** `WORLD_BOSS_INTERVAL_SECONDS` 3h -> 1h.

Supersedes the earlier standalone respawn PR #1449 (closed) and rebased onto `release/v0.21.0`.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands (on the `release/v0.21.0` base):
  - `npx vitest run tests/world_boss.test.ts` (27 pass: snare chase/attack/cadence, HP scaling 40k->200k->1M cap + no-shrink, size, loud engage + battle-cry range cutoff, centered self-aggroing summons)
  - `npx vitest run tests/parity tests/architecture.test.ts` (determinism + `src/sim` purity green; the snare, yell cycle, HP scaler, and summon changes draw no new rng)
  - `npx vitest run tests/localization_fixes.test.ts tests/i18n_status_registry.test.ts` (i18n guards green; `npm run i18n:gen` fresh, `pending=0`)
  - `npx tsc --noEmit` clean for every changed file
- Manual: ran the dev client and force-spawned Thunzharr offline for the size shots below; probed the summon wave (adds land <1.3yd from his center, aggroed on the tank, `spawnPos == boss spawn`) and the HP curve (40k solo -> 200k at 5 -> 1M cap).

## Screenshots / recordings

The size buff, same vantage with the player in frame for scale (`scripts/thunzharr_size_shot.mjs`):

| Before (`scale 1.7`) | After (`scale 5`) |
|---|---|
| before | after |

---

### Quality

- [x] **Tests pass.** Added/updated `tests/world_boss.test.ts`; the sim/parity/i18n suites above are green. (Ran the relevant suites directly rather than the full `npm test`.)
- [x] **Typecheck passes.** `npx tsc --noEmit` is clean for every file this PR changes.
- [ ] ~Builds succeed.~ Ran the dev server to validate rendering; did not run production `npm run build` in this worktree.

### Cross-platform

- [ ] ~Tested on desktop and mobile.~ No HUD/chrome change; the size buff is a platform-agnostic 3D `entity.scale`.
- [ ] ~Accessible.~ N/A: adds no UI/HUD.

### Localization (i18n)

- [ ] ~New player-visible strings are translated into every locale.~ Contributor English-only per the repo workflow; the maintainer fills locales at release.
- [x] Numbers/money/dates go through the formatters (N/A: none added).
- [x] Player-facing text from `src/sim/` is re-localized at the client: the Howling Gale debuff name is wired into `sim_i18n.ts` (`mechanic.howlingGale` + `AURA_NAME_KEY`); `tests/localization_fixes.test.ts` passes. Boss yell lines ship English under the existing boss-yell (variable-routed chat) precedent.

### Hygiene

- [x] No secrets/`.env`; `ALLOW_DEV_COMMANDS` is not enabled in any production path (the screenshot script drives the offline sim directly).
- [x] Did not hand-edit generated files; `src/ui/i18n.status.summary.json` was regenerated with `npm run i18n:gen`.


FernandoX7 (commented):
This is a clean, well-constructed sim change. The two new mechanics follow the existing template-pulse pattern (`stomp`/`terrify`) exactly: `aoeSlow` and `battleYells` are optional `MobTemplate` fields, the three new `Entity` timers are server-local and initialized in both `baseEntity` and `blankEntity` (`net/online.ts:807`), so entity shape stays in parity, and both pulses early-return for any template without the field so they can't perturb the determinism trace. No rng anywhere in the snare, the yell cycle, or the HP scaler. I ran the guards against the head ref: `tests/world_boss.test.ts` (24 pass), `architecture.test.ts`, and `localization_fixes.test.ts`/`i18n_status_registry.test.ts` after `i18n:gen` all green, and the committed generated i18n artifacts are fresh.

A few things I checked that hold up:
- The anti-kite math is right: `moveSpeedMult` folds a `slow` aura as `Math.min(slow, value)` (`sim.ts:2955`), so `value: 0.2` is 20% speed = 1.4yd/s, well under the boss's 5.8. The snare firing from both the chase and attack cases is genuinely once-per-tick since each `case` breaks before the next.
- "Widens its melee reach" is accurate: Thunzharr uses the default combat profile, whose reach is `scaledDefaultMobMeleeRange(scale)` (`mob_combat.ts:43`), so 3.4 takes it from +2.1 to +7.2yd.
- The `emitMobYell` range default keeps every non-loud mob at `YELL_RANGE`, and the engage/summon/enrage calls pass `battleYells?.range` (undefined for ordinary mobs), so the widening is scoped to loud bosses only.

Tests are decisive with real negative controls (the 200 vs 400yd yell cutoff, the in/out-of-radius snare pair, the 100k cap, the no-shrink case), not just happy-path.

Two nits, neither blocking:

- nit `src/sim/world_boss.ts:126` `scaleWorldBossHp` recomputes `worldBossContributors` (a dedupe + sort over the hate table) every tick for the boss's whole life, even after `maxHp` has hit the `max` cap and the result can no longer change anything. A one-line `if (boss.maxHp >= def.hpScale.max) return;` at the top skips it once capped. A world-boss fight is thousands of ticks, so it's cheap to shed.

- nit `src/sim/world_boss.ts:135` adding `delta` to both `maxHp` and current `hp` nudges the boss's HP fraction upward when a participant joins mid-fight (a boss at 50% of 40k becomes ~58% of 48k). It's a deliberate "real health, not a heal" choice and I think it's the right call, but the side effect is that a late joiner can push a boss that just crossed the 20% enrage threshold back above it and delay the enrage. Worth a one-line comment so the next reader knows it's intentional.

FernandoX7:
One more nit on a closer test pass:

- nit `tests/world_boss.test.ts:371` the two snare tests set `boss.aoeSlowTimer = 1 / 20` and tick once, so they only ever observe a single immediate pulse and never exercise the once-per-tick cadence guard or the `resetEvadingMob`/`createMob` reseed. I mutation-checked this at the head ref: deleting the guard in `pulseAntiKiteSnare` (`aoeSlowTimer -= DT; if (aoeSlowTimer > 0) return; aoeSlowTimer = aoeSlow.every`) so it fires every tick leaves both snare tests green. The core wiring pins are decisive (removing the chase call fails test 1, removing the radius check fails test 2), so this is only a gap on the cadence: a follow-up assertion that a second tick inside `every` does not re-apply the slow would close it.

ryan-foo:
Thanks for the close read. Addressed all three nits in `fix(world-boss): address review feedback on the HP scaler and snare cadence`, and rebased onto the current `release/v0.21.0` (conflicts resolved, mergeable):

- **`scaleWorldBossHp` running every tick after the cap** (world_boss.ts): added `if (boss.maxHp >= def.hpScale.max) return;` at the top, so the `worldBossContributors` dedupe+sort stops once the pool can no longer change.
- **HP-fraction nudge on late join** (world_boss.ts): added a comment noting it's the deliberate "real health, not a heal" choice, and calling out the side effect you spotted (a joiner crossing an hp-fraction threshold can push the boss back above the 20% enrage / a summon gate and delay it).
- **Snare cadence gap** (world_boss.test.ts): added a test that a second engaged tick inside `every` does not re-apply the slow (asserts `remaining` only decayed by one tick, not refreshed). Mutation-checked it the same way you did: neutering the `aoeSlowTimer > 0` guard so it fires every tick makes the new test fail. Confirmed the wiring pins still hold too (26 world_boss tests pass).

PR #1447, opened 2026-07-04
feat(sfx): add lockpick minigame audio

jamiecypher:
## Summary

Adds 14 custom 320 kbps MP3 sound effects for the lockpick minigame (Collapsed
Reliquary, and any future delves that adopt the mechanic).

This is the first non-ElevenLabs audio contribution, so the SFX pipeline is
extended to support custom recordings alongside generated clips:

- `scripts/sfx/sfx_prompts.mjs` — new `custom: true` flag on entries that have
  hand-recorded files. No prompt or duration needed.
- `scripts/gen_sfx.mjs` — generator skips API call for `custom: true` entries
  even with `--force`, so custom recordings are never overwritten.
- `src/game/sfx_manifest.generated.ts` — 14 new key-to-path entries.
- `src/ui/hud.ts` — `sfx.playUi()` wired to all lockpick sim events. The four
  `advanced` variants (pin setting clicks) pick randomly without immediate repeats.

**To add a custom recording in the future:** drop the MP3 in `public/audio/sfx/`,
add `{ key: 'your_key', custom: true }` to the SFX array in `sfx_prompts.mjs`,
run `node scripts/gen_sfx.mjs` to rebuild the manifest, then wire it up.

Sound map:
| Event | Result | Sound |
|---|---|---|
| `lockpickSession` | — | `lockpick_begin` |
| `lockpickStep` | `advanced` | `lockpick_advanced_1-4` (random, no repeat) |
| `lockpickStep` | `slip` | `lockpick_slip` |
| `lockpickStep` | `bind` | `lockpick_bind` |
| `lockpickStep` | `trap` | `lockpick_trap` |
| `lockpickStep` | `pageCleared` | `lockpick_page_cleared` |
| `lockpickStep` | `retry` | `lockpick_retry` |
| `lockpickStep` | `success` | `lockpick_success` |
| `lockpickStep` | `fail` | `lockpick_fail` |
| `lockpickEnd` | `success` | `lockpick_end` |
| `lockpickBonus` | — | `lockpick_bonus` |

## Type of change

- [x] Build, CI, or tooling

## How was this tested?

- Commands: `npx tsc --noEmit` (clean)
- Manual steps: ran the Collapsed Reliquary delve at level 20 offline, verified
  all lockpick events trigger the correct sounds including random variation on
  pin advances.

---

### Quality

      TypeScript `strict`).

### Cross-platform

- N/A. This PR adds no UI changes.

### Localization (i18n)

### Hygiene

- N/A. `sfx_manifest.generated.ts` is intentionally hand-edited here — the
  generator only covers ElevenLabs clips and custom recordings must be added
  manually. This is documented in the summary and in `sfx_prompts.mjs`.


FernandoX7 (commented):
The SFX wiring itself is clean and complete. Every `StepResult` case is covered (checked against the union at `src/sim/lockpick.ts:133`), `sfx.playUi` no-ops on a missing key so a typo can't throw, and `preload()` in `src/game/sfx.ts:93` loads every manifest entry, so all 14 clips are live at runtime. The `custom: true` flag is a sensible pipeline extension and correctly skips the API path without ever touching `entry.prompt`/`entry.duration` (it `continue`s before those reads). The files are valid 320 kbps MP3s, and `lockpick_begin`/`lockpick_bonus` are genuinely distinct despite the matching byte size.

A few things to sort out, mostly around what actually merges rather than the sound code.

**should-fix (merge scope).** The base is `release/v0.21.0` but the head branch is `main`, so this PR carries 7 unrelated commits (11 non-SFX files) on top of the sound change: a version bump (`package.json` 0.19.0 to 0.20.0, plus the ios/android project files), world-boss-at-boot (`server/game.ts`, `src/sim/sim.ts`, `src/sim/types.ts`, `tests/world_boss.test.ts`), the grindable T1 epics (`src/sim/content/zone3.ts`, `dungeons.ts`, `tests/fixes.test.ts`), and the Gravewyrm door move (`dungeons.ts`, `tests/social.test.ts`). The merge is clean, but as targeted it backports all of that into the release branch (including setting its version to 0.20.0). Rebasing the two SFX commits onto `release/v0.21.0`, or a fresh branch off it, drops the baggage so the diff is just the audio.

**should-fix (biome on changed files).** Two changed files carry a format diff that `npm run ci:changed` fails on:
- `src/ui/hud.ts:7843`: the column-aligned single-line `case` bodies. Biome expands each `case ...: sfx.playUi(...); break;` onto three lines.
- `scripts/sfx/sfx_prompts.mjs:449`: the aligned `custom: true` column collapses to single spaces, and `lockpick_page_cleared` is missing the space in `,custom`.

`npx @biomejs/biome check --write src/ui/hud.ts scripts/sfx/sfx_prompts.mjs` fixes both.

**nit (generated-file note).** The Hygiene note and the comment in `sfx_prompts.mjs` say the manifest must be hand-edited because the generator only covers ElevenLabs clips. The manifest-rebuild loop at `scripts/gen_sfx.mjs:92` actually includes any key whose `.mp3` is on disk, custom or not, so running the generator would emit the 14 entries itself, in sorted position and standard multi-line form. The committed entries are appended after `spell_nova` in compact form, so the next `sfx:gen` run will reorder and reformat them. Nothing breaks (the generator aborts without `ELEVENLABS_API_KEY`, `*.generated.ts` is biome-ignored, and the base manifest already drifts from sort order), so this is cosmetic. If you want custom-only additions to regenerate cleanly, gating only the generation step on the API key (not the manifest rebuild) would let `sfx:gen` run without a key.

**nit (double sound on solve).** On a full solve the sim emits `lockpickStep` result `success` and then `lockpickEnd` outcome `success` in the same beat, so `lockpick_success` and `lockpick_end` play stacked. Looks intentional from the sound map, just worth a listen to confirm it does not clutter.

**nit (provenance).** These are the first hand-recorded (non-generated) audio assets, and CREDITS.md has no audio section (the generated SFX are not listed there either), so this is not a regression. Custom recordings do carry their own license/ownership question though, so worth confirming provenance before it rides into a release.

jamiecypher:
Superseded by #1462 — rebased onto release/v0.21.0 with review feedback addressed.1462

PR #1448, opened 2026-07-04
feat(ui): add mailbox recipient autocomplete

Pepijnvdliefvoort:
## Summary

Add recipient name autocomplete to the Ravenpost mailbox

This change adds live character suggestions while typing in the recipient field, with debounce, self-filtering, keyboard and mouse selection, and ARIA combobox/listbox wiring.

Updated areas:
- src/ui/mailbox_window.ts: Send-tab recipient autocomplete logic and interactions
- src/styles/components.css: minimal mailbox-specific suggestion list positioning, reusing social suggestion item style
- tests/mailbox_window.test.ts: source-shape regression coverage for autocomplete wiring and selection behavior

## Type of change

~~- [ ] Bug fix~~
~~- [ ] Refactor or performance (no behavior change)~~
~~- [ ] Documentation~~
- [x] Tests
~~- [ ] Build, CI, or tooling~~
~~- [ ] Breaking change (please call it out in the summary)~~

## How was this tested?

- Commands:
  - npx vitest run --config vite.config.ts tests/mailbox_window.test.ts
- Manual steps:
  - Started server and client locally.
  - Opened mailbox, switched to Send tab.
  - Typed recipient prefixes and verified live suggestions appear.
  - Verified ArrowDown and ArrowUp move highlight.
  - Verified Enter selects highlighted suggestion and writes name into recipient field.
  - Verified Escape closes suggestions.
  - Verified mouse hover updates highlight and click selects.
  - Verified blur closes suggestions with a short delay, allowing click selection first.
  - Verified existing send flow, coin fields, and parcel staging behavior still work.

## Screenshots / recordings
charactername-autocomplete

---

### Quality

- [x] Tests pass. npm test is green. I added or updated tests for any
      sim/ or server/ behavior I changed.
- [x] Typecheck passes. npx tsc --noEmit is clean (the project is
      TypeScript strict).
~~- [ ] Builds succeed. npm run build, plus npm run build:server and
      npm run build:env if I touched those.~~

### Cross-platform

- [x] Tested on desktop and mobile. Verified on a desktop and on a
      least 40x40px and inputs at least 16px font (see src/ui/CLAUDE.md).
- [x] Accessible. Keyboard-operable, with visible focus, sensible ARIA, and
      respect for prefers-reduced-motion (only if this change adds or edits UI).

### Localization (i18n)

~~- [ ] New player-visible strings are translated into every supported locale.
      Every user-facing string is a t() key, added to en first, then given a
      real translation in every locale in supportedLanguages
      (src/ui/i18n.ts). No English placeholders or TODO in other locales.~~
~~- [ ] Numbers, money, dates, units, and percents go through the i18n formatters
      (formatNumber, formatMoney, formatDateTime, Intl).~~
~~- [ ] Player-facing text emitted from src/sim/ or server/ is re-localized at
      npx vitest run tests/localization_fixes.test.ts passes.~~

### Hygiene

- [x] No secrets, credentials, or .env committed, and ALLOW_DEV_COMMANDS is
      src/render/assets/manifest.generated.ts. They're regenerated by the build.

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -5245,6 +5245,10 @@
     overflow-y: auto;
     box-shadow: 0 6px 18px #000a;
   }
+  .mail-to-suggest.up {
+    top: auto;
+    bottom: calc(100% + 3px);
+  }
   .mail-coin-row {
     flex-direction: row;
     align-items: center;
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -5225,6 +5225,26 @@
     outline-offset: 1px;
     border-color: var(--color-border-focus);
   }
+  /* Recipient field autocomplete: relative wrapper anchors the listbox below the
+     input. Reuses .soc-sugg-item / .soc-suggest visual tokens from the social panel
+     so no new design-system surface is introduced. */
+  .mail-to-wrap {
+    position: relative;
+  }
+  .mail-to-suggest {
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: calc(100% + 3px);
+    display: none;
+    z-index: 30;
+    background: #14110d;
+    border: 1px solid #6f5a2a;
+    border-radius: 5px;
+    max-height: 210px;
+    overflow-y: auto;
+    box-shadow: 0 6px 18px #000a;
+  }
   .mail-coin-row {
     flex-direction: row;
     align-items: center;
--- a/src/ui/mailbox_view.ts
+++ b/src/ui/mailbox_view.ts
@@ -122,3 +122,26 @@ export function mailIndicatorView(unread: number): MailIndicatorView {
   const count = Number.isFinite(unread) ? Math.max(0, Math.floor(unread)) : 0;
   return { visible: count > 0, count };
 }
+
+export interface RecipientSuggestion {
+  name: string;
+  cls: string;
+  level: number;
+}
+
+export function recipientSuggestions(
+  results: RecipientSuggestion[],
+  selfName: string,
+  max: number,
+): RecipientSuggestion[] {
+  const self = selfName.trim();
+  const limit = Math.max(0, Math.floor(max));
+  if (limit === 0) return [];
+  return results.filter((r) => r.name !== self).slice(0, limit);
+}
+
+export function wrappedSuggestionIndex(current: number, delta: number, total: number): number {
+  if (total <= 0) return -1;
+  if (current < 0) return delta > 0 ? 0 : total - 1;
+  return (current + delta + total) % total;
+}
--- a/src/ui/mailbox_window.ts
+++ b/src/ui/mailbox_window.ts
@@ -27,6 +27,8 @@ import {
   type MailSendBody,
   type MailTab,
   mailSendBlocked,
+  recipientSuggestions,
+  wrappedSuggestionIndex,
 } from './mailbox_view';
 import type { PainterHostPresentation } from './painter_host';
 import { svgIcon } from './ui_icons';
@@ -102,6 +104,8 @@ export class MailboxWindow {
 
   close(): void {
     if (!this.opened) return;
+    window.clearTimeout(this.recipientSuggestTimer);
+    this.recipientSuggest = { items: [], index: -1 };
     this.opened = false;
     this.openedId = null;
     this.attachments = [];
@@ -369,6 +373,8 @@ export class MailboxWindow {
   }
 
   private renderSend(body: HTMLElement, view: MailSendBody): void {
+    window.clearTimeout(this.recipientSuggestTimer);
+    this.recipientSuggest = { items: [], index: -1 };
     body.innerHTML =
       `<div class="mail-send-form">` +
       `<div class="mail-field"><label for="mail-to">${esc(t('hudChrome.mailbox.toLabel'))}</label>` +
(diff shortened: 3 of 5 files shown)


FernandoX7:
@Pepijnvdliefvoort Hi! Thanks for contributing. I'm reviewing now.

For future reference, please make sure to target the latest release branch, in this case `release/0.21.0`

EnriqueGF:
Love the idea. Thank you!

Pepijnvdliefvoort:
> @Pepijnvdliefvoort Hi! Thanks for contributing. I'm reviewing now.
> 
> For future reference, please make sure to target the latest release branch, in this case `release/0.21.0`

Will do so in the future. Apologies!

FernandoX7 (commented):
Nice, clean feature. The autocomplete is a faithful port of the social-panel typeahead: same debounced `searchCharacters`, self-exclusion, arrow/Enter/Escape handling, and a correct ARIA combobox/listbox with `aria-activedescendant`. It goes through the `IWorld` seam (`searchCharacters` exists on both `Sim` and `ClientWorld`), names pass through `esc()` so the interpolation is XSS-safe and round-trips on read-back, and the CSS reuses the `.soc-*` tokens rather than inventing a new surface. No new player-visible strings, tests and `tsc` are green.

A couple of things worth a look before merge.

**should-fix** `src/styles/components.css:5234` The suggestion list opens downward (`top: calc(100% + 3px)`, `max-height: 210px`) anchored to the To field, which is the first field at the top of `#mailbox-body`. That body is `overflow-y: auto` (`components.css:4975`) inside `#mailbox-window { overflow: hidden; height: min(560px, calc(85vh - 24px)) }`, and an absolutely positioned descendant is clipped by that scroll container (`z-index` does not escape it). On desktop full height it is fine, but in-game mobile is landscape-only, and at a ~390px viewport the body is ~240px while the To input sits ~84px down, so a full 210px list reaches past the bottom edge and the last rows are clipped (only reachable by scrolling the whole form). The social list dodges this by opening upward from a bottom-anchored input. Worth verifying on a short/landscape window; opening upward, or anchoring the list outside the scroll container, keeps every row visible.

**nit** `src/ui/mailbox_window.ts:410` The suggest model (`this.recipientSuggest`) and the pending debounce timer are never reset when the Send form is rebuilt (a tab switch or staging a parcel both call `render()`, which re-wires a fresh empty `#mail-to`) or on `close()`. The 150ms blur-clear mostly masks the stale-model case, but unlike `social_window` these helpers operate on a captured `body` (`querySelector` at `:503`/`:549`) rather than the live root, so a late `searchCharacters` resolution can repaint the detached old list and leave the shared model populated. Cheap fix: reset `this.recipientSuggest` and `window.clearTimeout(this.recipientSuggestTimer)` at the top of `renderSend()` and in `close()` (or query `this.deps.root()` live like social does).

**nit** `tests/mailbox_window.test.ts:65` The new autocomplete tests read the source as a string, so a few can't fail on the regression they name: the self-exclude test's `/filter[\s\S]{0,80}player\.name/` matches whether the operator is `!==` or `===`, and the aria-expanded test (`:97`) only checks both literals exist, not that each is on the right branch. The wrap-around index math has no behavioral pin either (that part mirrors social, which is the same source-scan style). The product code is correct today; the cheap upgrade is to lift `recipientSuggestions(results, self, max)` and the wrap-index math into `mailbox_view.ts`, which already has a behavioral suite, and pin exclude/limit/wrap there.

Not blocking: `.mail-to-suggest` is a second near-verbatim copy of `.soc-suggest`, which the rule of three allows, and the raw hex matches the documented deferred `.soc-*` tokenization. Just a marker that a third copy, or any restyle of the shared `.soc-*` classes, should trigger pulling out one shared listbox rule.

FernandoX7 (commented):
Re-reviewed after the `harden` commit. All three points from the last pass are addressed, and cleanly.

- **Clipping (was should-fix).** `placeRecipientSuggest` (`src/ui/mailbox_window.ts:551`) now measures the visible space between the To field and the `#mailbox-body` scroll edge and either flips the list upward (`.mail-to-suggest.up`, `src/styles/components.css:5248`) or clamps `maxHeight` to what fits, so the list scrolls internally instead of being clipped by the `overflow-y: auto` container. On the in-game landscape target the To field sits near the top with plenty of room below, so it opens down and fits; on a short window it clamps or flips. Solid.
- **Lifecycle reset (was nit).** The debounce timer and suggest model are now cleared at the top of `renderSend()` (`:376`) and in `close()` (`:107`), so a tab switch, a parcel staging, or a close no longer leaves a stale model or a pending resolution behind. A late `searchCharacters` callback also no-ops now, since it re-queries `body` and finds no `#mail-to-suggest` off the Send tab.
- **Test decisiveness (was nit).** `recipientSuggestions` and `wrappedSuggestionIndex` are lifted into `mailbox_view.ts` (`:132`, `:143`) and pinned behaviorally in `tests/mailbox_view.test.ts` (exclude + cap via the `['Alice','Bob']` assertion, zero/negative limit, wrap in both directions, empty list). The window-side tests now check delegation to those helpers plus a branch-aware `aria-expanded` regex that would fail if the empty/non-empty branches were swapped. Good split.

Verified locally against the head ref: `tsc --noEmit` clean, and `tests/mailbox_view.test.ts` + `tests/mailbox_window.test.ts` both green (34 tests). Base is correct now too (`release/v0.22.0`), and it merges clean.

Looks good to merge.

FernandoX7 (approved):

PR #1449, opened 2026-07-04
feat(world-boss): respawn every hour

ryan-foo:
## What

Change the world-boss respawn cadence from every 3 hours to **every hour** (`WORLD_BOSS_INTERVAL_SECONDS`), so Thunzharr rises more frequently.

## How

One-line constant change in `src/sim/world_boss.ts` (`3 * 3600` to `1 * 3600`), plus the comment. The scheduler and the reschedule test both read the constant, so nothing else needs to change; `tests/world_boss.test.ts` passes (17).

## Note

This is an alternative to PR #1446, which set the same cadence to 90 minutes. Both edit the same line, so whichever merges second will need a trivial rebase (or just pick this one for a 1-hour respawn).

### DIFF
--- a/src/sim/world_boss.ts
+++ b/src/sim/world_boss.ts
@@ -22,8 +22,8 @@ import type { Entity, LootSlot } from './types';
 
 // Sim-time cadence: a fresh boss rises this many seconds after the previous one
 // was scheduled. On the live server the sim runs at wall-clock speed (20 Hz), so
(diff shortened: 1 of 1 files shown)


ryan-foo:
Folded into #1446 (now a 1-hour respawn, rebased onto release/v0.21.0). Closing this standalone PR.

PR #1451, opened 2026-07-04
fix(ui): staging a mail parcel no longer wipes the typed send form

### LINKED ISSUE #1443: [Mailbox] When I join an item to a mail, it should not reset the email form
### Before you start

### What happened?

When I join an item to a mail, it should not reset the email form

### Steps to reproduce

- Go to the Mailbox
- Click on send tab
- Fill the "To", "Subject" and "Message" fields
- Open your bag
- select an item to send

-> To, Subject and Message fields are empty 

Adding an item should not empty the message form fields.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

_No response_

_No response_

MarchalMaxim:
Fixes #1443.

## Problem

In the mailbox Send tab, clicking a bag item to attach it as a parcel calls MailboxWindow.stageParcel(), which did a full this.render(). That rebuilds the whole window's innerHTML, including the send form, so the typed To, Subject, Message, and coin fields were wiped every time an item was attached.

## Fix

stageParcel() now repaints only the parcels row via renderParcels(), the same scoped path the remove-parcel chip already uses. Nothing else in the send form depends on the attachments list (renderSend reads only postage and delivery time, both attachment-independent), and the max-attachments error path reads info.maxAttachments live at click time, so the scoped repaint leaves nothing stale.

## Test

Added a source-level guard to tests/mailbox_window.test.ts (the file's established shape) pinning that stageParcel calls renderParcels() and never the full render().

Verified: npx vitest run tests/mailbox_window.test.ts tests/mailbox_view.test.ts (16 passed), npx tsc --noEmit, biome on both changed files, and the qa-checklist gate (verdict READY, zero blocking findings).


jgyy:
Closing — this looks already implemented. `mailbox_window.ts` on `upstream/main` already does a targeted repaint instead of wiping the send form, and cites this exact bug. Thanks for the contribution.

PR #1453, opened 2026-07-04
fix(ui): tint the chat input by channel and make the last used send channel sticky

### LINKED ISSUE #1452: Chat input color does not indicate the most recently used chat channel
## Problem

Chat log lines are tinted per channel (party blue, guild green, yell red, and so on), but the chat input itself always renders in the same neutral color, and it does not remember the channel the player last used. After sending a message to party chat, the next time the player opens the chat input there is no color cue telling them where their text will go, and plain text falls back to say instead of the channel they just used. Players expect the input to indicate, by color, the most recently used chat channel, the way classic chat behaves.

## Steps to reproduce

1. Join a party and send a message to party chat ("/p hello", or from the Party tab).
2. Close the chat input.
3. On the All tab, press Enter to open the chat input again and start typing.

## Expected behavior

- The chat input text is tinted with the color of the channel the typed plain text will be sent to.
- The most recently used send channel is sticky: reopening the input on the All tab defaults to the last channel the player actually sent to (party in the example), reflected by both the placeholder and the input color.
- A channel-bound tab keeps its bound channel; the input tint always matches the channel the plain typed text will actually reach.
- An explicit slash command the player types still wins, exactly as today.

## Actual behavior

- The chat input always renders in the default color, regardless of the active tab or the last channel used.
- Reopening the input on the All tab always defaults back to say, even if the player just spoke in party.

## Scope

- Chat input open/send path: composeChatSend, activeChatPlaceholder, and the chat open wiring (src/ui/hud.ts, src/main.ts).
- Channel rules and colors: src/ui/chat_channels.ts (per-channel colors currently live inline in the hud.ts chat event switch and should be shared, not duplicated).
- Whisper is excluded from stickiness (it targets a specific player, not a standing channel).

## Acceptance criteria

- Sending a message to a channel makes it the sticky default: the next opened chat input (on the All tab) sends plain text there and is tinted with that channel's color.
- The input tint always matches the channel plain text will reach, including on channel-bound tabs; say uses the default input color.
- An explicit "/..." command typed by the player is unaffected.
- The desktop and mobile experience is otherwise unchanged.

MarchalMaxim:
## Summary

The chat input always rendered in the same neutral color and forgot the channel the player last used: after speaking in party chat, reopening the input gave no color cue and plain text fell back to say.

This PR makes the chat input reflect where typed text will actually go:

- The input is tinted with the color of the channel a plain typed line will reach, using the same per-channel colors as the chat log lines.
- The most recently used send channel is sticky on the All/Combat views (classic sticky-channel behavior): sending to a channel (via its tab or an explicit command like "/p hello") makes it the default for the next plain line, reflected by both the placeholder and the tint.
- A channel-bound tab keeps its bound channel, an explicit "/..." command still wins, and a plain reply on the whisper collector tab never re-sticks (whisper targets a player, not a standing channel).
- Bare "/g" is deliberately excluded from the sticky alias map: the offline sim routes it to general while the server routes it to guild, so it cannot stick honestly on both hosts.

Implementation: pure rules in src/ui/chat_channels.ts (CHAT_CHANNEL_COLORS, effectiveSendTab, chatInputColor, channelForSlashCommand, stickyChannelAfterSend), unit-tested in tests/chat_channels.test.ts; hud.ts and main.ts stay thin consumers. The hud chat event switch now reads the shared color map instead of nine inline hex literals (single source, byte-identical values, so the log and the input can never disagree). No new i18n keys (the existing "sendingTo" placeholder keys are reused), no sim/server/IWorld changes, and the sticky channel is session-only (nothing persisted).

## Related issues

Fixes #1452.

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: TDD (the new pure-model tests were written first and confirmed red, then the implementation turned them green); `npm test` (full suite green), `npx tsc --noEmit`, `npm run security:gate` (PASS), `npm run build` (succeeds), Biome clean on the changed files, and the repo QA checklist over the diff (no blocking findings).
- Manual steps: verified in the offline client that the input opens neutral by default, re-sticks and tints party blue after "/p ...", tints yell red after "/y ..." with the matching log line, resets to the default color on "/s ...", and that channel tabs stay bound regardless of stickiness.

## Screenshots / recordings

Before: the input is the neutral default regardless of what was last used.

before: neutral say input

After sending "/p pull on 3": the reopened input is party-tinted and the placeholder reads "Message Party"; plain text now goes to party.

after: party-tinted input

After "/y incoming!": the input follows the last used channel and matches the yell line color in the log.

after: yell-tinted input matching the log line

---

### Quality

- [x] **Tests pass.** `npm test` is green; new pure-model tests added for the channel color/sticky rules.
- [x] **Builds succeed.** `npm run build` passes (no server/headless code touched).

### Cross-platform

- [x] **Tested on desktop and mobile.** The change is color-only (no layout, no new controls); every input open path (keybind, mobile toggle, whisper context menu, link insert, tab switch) applies the tint through the same shared method. Touch targets and input font sizes are untouched.
- [x] **Accessible.** No new interactive elements; the tint supplements the existing localized placeholder text ("Message Party"), so the channel is never conveyed by color alone.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (existing `t()` keys are reused; `tests/localization_fixes.test.ts` passes).

### Hygiene

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1055,7 +1055,9 @@ async function startGame(
   };
   function openChat(): void {
     // reflect the active chat-channel tab in the placeholder (e.g. "Message World")
+    // and tint the input with the channel the typed text will reach (#1452)
     chatInput.placeholder = hud.activeChatPlaceholder();
+    chatInput.style.color = hud.activeChatInputColor();
     chatInput.style.display = 'block';
     anchorChatInput();
     autosizeChatInput();
--- a/src/ui/chat_channels.ts
+++ b/src/ui/chat_channels.ts
@@ -96,6 +96,84 @@ export function chatOpenTabLabelKey(tab: ChatOpenTab): TranslationKey {
   return tab === WHISPER_TAB ? WHISPER_TAB_LABEL_KEY : CHANNEL_LABEL_KEYS[tab];
 }
 
+// One tint per openable tab, shared by the chat log lines (hud.ts chat event
+// switch) and the chat input (issue #1452: the input is tinted with the color
+// of the channel the typed plain text will reach). Single source: the log and
+// the input must never disagree on a channel's color.
+export const CHAT_CHANNEL_COLORS: Record<ChatOpenTab, string> = {
+  say: '#f0ead8',
+  yell: '#ff5040',
+  party: '#7fd4ff',
+  general: '#ffc864',
+  world: '#ff9d5c',
+  lfg: '#5cd6a0',
+  guild: '#40d264',
+  officer: '#4ce0c0',
+  whisper: '#ff80ff',
+};
+
+// The tab a plain typed line will actually reach: a channel tab binds itself,
+// the whisper collector replies to the last whisperer, and the All/Combat
+// views fall back to the sticky most-recently-used channel (null = say).
+export function effectiveSendTab(
+  activeTab: ChatTabId,
+  sticky: ChatTabChannel | null,
+): ChatOpenTab | null {
+  if (activeTab === 'all' || activeTab === 'combat') return sticky;
+  return activeTab;
+}
+
+// The chat input's tint for an effective send tab. null = leave the CSS
+// default color (say is the engine default and keeps the neutral input).
+export function chatInputColor(tab: ChatOpenTab | null): string | null {
+  if (tab === null || tab === 'say') return null;
+  return CHAT_CHANNEL_COLORS[tab];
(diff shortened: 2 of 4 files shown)


Rubsey:

Thanks for the contribution, @MarchalMaxim.

PR #1458, opened 2026-07-05
fix(input): align fullscreen mouselook and diagonal movement visuals

Humpalumps:
﻿## Summary

This prepares the v0.21.0 release branch with two player-control fixes reported from manual playtesting:

- Fixes fullscreen right-drag mouselook so it uses the same pointer-lock path as windowed play. Previously, fullscreen was explicitly excluded from pointer lock, so the cursor stayed visible, could hit the screen edge, and mouselook behaved differently from non-fullscreen. The pointer-lock setting is still respected, and the code still avoids repeat lock requests while a drag is already locked.
- Adds render-only diagonal movement facing for the local player. When the player combines forward/back movement with strafing, the character model now visually turns into the diagonal travel direction instead of facing straight forward while sliding diagonally. This is intentionally presentation-only: authoritative sim/server facing, network movement, combat arcs, targeting, click-to-move, and camera-facing behavior remain unchanged.

## Related issues

N/A - reported from manual QA/playtesting.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/movement_visual.test.ts tests/pointer_lock.test.ts tests/input.test.ts`
  - `npx tsc --noEmit`
  - `npx biome check src/game/movement_visual.ts tests/movement_visual.test.ts src/main.ts src/game/pointer_lock.ts tests/pointer_lock.test.ts --max-diagnostics 40`
  - `npm run build`
- Manual steps:
  - Ran the patched clean release worktree locally at `http://127.0.0.1:5173/`.
  - Verify right-click drag mouselook in windowed mode still hides/locks the cursor after the drag threshold.
  - Enter browser fullscreen and verify right-click drag mouselook now also hides/locks the cursor and no longer stops at the monitor edge.
  - Verify W+A/W+D and back diagonals rotate the local character model into the diagonal travel direction.
  - Verify pure forward, pure strafe, and pure back movement keep the existing presentation.

## Screenshots / recordings

- [Test evidence recording: fullscreen rotation and diagonal character movement](https://github.com/Humpalumps/world-of-claudecraft/releases/download/pr-1458-test-evidence/woc-pr-1458-test-evidence.mp4)
  - Hosted as a GitHub release asset on the fork: https://github.com/Humpalumps/world-of-claudecraft/releases/tag/pr-1458-test-evidence
  - Shows the fullscreen rotation/pointer-lock behavior and diagonal character-facing movement fix working together.

---

### Quality

- [ ] **Tests pass.** `npm test` is green. I did not run the full suite; focused regression tests passed.
- [x] **Builds succeed.** `npm run build` completed successfully. Server/env builds were not run because this change is client input/rendering only.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Desktop local server is available for QA; mobile viewport was not exercised in this pass.
- [ ] **Accessible.** N/A - this does not add or edit UI controls.

### Localization (i18n)

- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary in this same change, and `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] I didn't hand-edit generated files such as `src/render/assets/manifest.generated.ts`. Build-generated line-ending churn was removed from the branch before committing.

## Reviewer notes

- The fullscreen fix is a one-line decision change in `shouldEngagePointerLock`: fullscreen no longer blocks pointer lock.
- The diagonal-facing change is isolated in `src/game/movement_visual.ts` and is only passed into `renderer.sync` for the local player render override.
- `npx biome check` passed for the touched/new files. It still reports existing warnings in `src/main.ts` when run broadly, but no new formatting or correctness failures were introduced by this change.


FernandoX7 (commented):
Two tidy client-side fixes, and both hold up under scrutiny.

The diagonal-facing math is correct against the sim's own movement convention, which is the thing I most wanted to confirm. The sim builds world velocity as `world = mz*(sin f, cos f) + mx*(-cos f, sin f)` (`src/sim/sim.ts:3200`, `:3243`), which reduces to a travel-direction angle of `f - atan2(mx, mz)`. That is exactly `diagonalMovementVisualFacing` (`baseFacing - Math.atan2(mx, mz)`), with the same `mz` forward / `mx` strafe-right sign convention, so the model points precisely where the feet are actually going. The pinned angle tests are decisive and pass.

The render-only claim also holds. Offline still ticks the sim off `stepFacing = movementFacing ?? facing` and only feeds `offlineRenderFacing` into `renderer.sync`; online still streams `netFacing` via `net.setMouselookFacing` and only feeds `onlineRenderFacing` into `renderer.sync`. Nothing authoritative sees the visual override. And the trigger is correctly gated: in classic non-mouselook mode A/D are turn, not strafe, so `mx == 0` and the override never fires while turning in place.

Ran `npx vitest run tests/movement_visual.test.ts tests/pointer_lock.test.ts tests/input.test.ts` at the head: 65 passed.

Nits, nothing blocking:

- `src/main.ts:2409` (offline) vs `:2439` (online): the offline render facing samples a fresh `input.readMoveInput()`, while the online one reuses `resolved.mi`. They agree in every diagonal case today (`resolveMove` only ever sets `mi.forward` for click-to-move, never strafe, so the click-to-move path is null either way), so there is no visible bug. But two sources for the same visual decision is a small drift risk; feeding both from the `resolveMove` result would keep them locked together.

- `src/game/pointer_lock.ts:29`: `shouldEngagePointerLock` no longer reads `isFullscreen`, so the parameter is now inert in the body and kept only so the test can assert both fullscreen states resolve the same. That is a fine way to pin the regression, just flagging that the field is now decorative.

One heads-up, not a finding: dropping the fullscreen skip means a fullscreen right-drag now surfaces Chrome's "press Esc to exit full screen" prompt that the old code deliberately sidestepped. Trading that prompt for edge-free, consistent mouselook looks like the right call, just worth a conscious ack since it is a visible behavior change.

PR #1463, opened 2026-07-05
fix(lint): clear noArguments and unused-param warnings on the delve rite / input seam

jgyy:
## Summary
- `Input.setControllerMoveInput` used `arguments.length` to detect an omitted `facing` arg; every call site either omits it or passes a defined value, so `facing !== undefined` is equivalent and drops the legacy `arguments` object (`lint/complexity/noArguments`).
- `chooseDrownedLitanyRiteIntensity` never used its `ctx` param; keeps the `SimContext`-seam signature (for consistency with sibling `tickDrownedLitanyRite`) but marks it unused, and simplifies the guard to an optional chain (`lint/correctness/noUnusedFunctionParameters`, `lint/complexity/useOptionalChain`).

Both are pre-existing biome warnings (not errors) surfaced on these files by the pre-push floor; no behavior change.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/delves.test.ts tests/input.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts` (244 passed, 3 skipped)
- [x] `npx @biomejs/biome check --changed --since=release/v0.21.0` clean on the two touched files

### DIFF
--- a/src/game/input.ts
+++ b/src/game/input.ts
@@ -522,7 +522,7 @@ export class Input {
 
   setControllerMoveInput(input: unknown, facing?: unknown): void {
     this.controllerMoveInput = sanitizeMoveInput(input);
-    if (arguments.length > 1) this.controllerFacing = sanitizeMoveFacing(facing);
+    if (facing !== undefined) this.controllerFacing = sanitizeMoveFacing(facing);
   }
 
   setControllerFacing(facing: unknown): void {
--- a/src/sim/delves/drowned_litany_rite.ts
+++ b/src/sim/delves/drowned_litany_rite.ts
@@ -170,12 +170,12 @@ export function spawnDrownedLitanyRite(
 /** Lock in the chosen difficulty: generate the seeded sequence and start playback.
  * Shared per run (the first chooser commits it); returns false if not awaiting. */
 export function chooseDrownedLitanyRiteIntensity(
-  ctx: SimContext,
+  _ctx: SimContext,
   run: DelveRun,
   intensity: RiteIntensity,
 ): boolean {
   const st = run.drownedLitanyRite;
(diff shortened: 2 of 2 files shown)


jgyy:
### What changed, visually

```mermaid
flowchart TD
    subgraph before["Before"]
        A1["setControllerMoveInput(input, facing?)"] --> A2{"arguments.length > 1?"}
        A2 -- yes --> A3["controllerFacing = sanitizeMoveFacing(facing)"]
        A2 -- no --> A4["skip facing update"]

        B1["chooseDrownedLitanyRiteIntensity(ctx, run, intensity)"] --> B2{"!st || !st.awaitingChoice?"}
        B2 -- true --> B3["return false"]
        B2 -- false --> B4["commit intensity, generate sequence"]
    end

    subgraph after["After"]
        C1["setControllerMoveInput(input, facing?)"] --> C2{"facing !== undefined?"}
        C2 -- yes --> C3["controllerFacing = sanitizeMoveFacing(facing)"]
        C2 -- no --> C4["skip facing update"]

        D1["chooseDrownedLitanyRiteIntensity(_ctx, run, intensity)"] --> D2{"!st?.awaitingChoice?"}
        D2 -- true --> D3["return false"]
        D2 -- false --> D4["commit intensity, generate sequence"]
    end

    before -.->|"equivalent behavior, no arguments object / no unused param"| after
```

Every call site of `setControllerMoveInput` either omits `facing` entirely or passes a defined value (checked in `src/main.ts`), so `facing !== undefined` is behaviorally identical to the old `arguments.length` check. `chooseDrownedLitanyRiteIntensity` never read `ctx`; the param is kept (renamed `_ctx`) to match the `SimContext`-seam signature shared with `tickDrownedLitanyRite` in the same file.

PR #1466, opened 2026-07-05
fix(render): stack overlapping nameplates instead of rendering them on top of each other

jgyy:
## Summary

Investigated issue #1229 (mobile touch controls drift) first: read through
`mobile_controls.ts`/`input.ts` and live-tested with synthetic touch events in
a mobile-viewport browser session, but could not reproduce the reported
drift. The pointer lifecycle (down/move/up/cancel/lostpointercapture, plus
`blur`/`visibilitychange`) already clears movement cleanly on release.

While testing the mobile HUD live, I found a different, reproducible bug:
`NameplatePainter.update()` (`src/render/nameplate_painter.ts`) projects each
entity's nameplate to screen space independently with **no overlap
detection**. When two entities stand near each other (NPCs clustered around a
camp, two same-named mob respawns), their nameplates render on top of each
other as unreadable, garbled overlapping text. This affects every viewport,
but is dramatically worse on mobile landscape (~390px tall vs 800-1000px+ on
desktop): entities need to be much farther apart in world space before their
vertical projections separate on screen.

**Fix:** a new pure module, `src/render/nameplate_declutter.ts`, takes this
frame's projected nameplate anchors and nudges vertically apart any that
would collide (checked with an asymmetric threshold: ~80px horizontal since
nameplate labels render much wider than their anchor point, ~18px vertical
since a label is a single text line). `NameplatePainter.update()` now runs a
second pass after projection that applies the decluttered positions, a no-op
in the common non-overlapping case.

- Verified live before/after on a mobile-landscape viewport (844x390, offline
  character): overlapping NPC and mob nameplates now stack on separate lines.
- Also fixed a pre-existing, unrelated Biome import-order warning in
  `src/sim/data.ts` (safe auto-fix only, no behavior change) that the
  pre-push floor's changed-file diff happened to pick up.

## Before / after

**Before** (NPC nameplates "Smith Haldren" / "Marshal Redbrook" / "The
Merchant" fully overlapping, mob nameplates "[1] Forest Wolf" x2 overlapping):

before

**After** (same spot, same character, nameplates now stack on separate
lines):

after

## Flow

```mermaid
flowchart TD
    A["NameplatePainter.update()<br/>per-entity projection loop"] --> B["compute sx, sy<br/>(screen anchor per entity)"]
    B --> C["push {id, sx, sy}<br/>into anchorScratch"]
    C --> D["apply nameplateScreenTransform(sx, sy)<br/>(unchanged, per-entity)"]
    D --> E{"loop done for<br/>all entities?"}
    E -- no --> B
    E -- yes --> F["declutterNameplates(anchorScratch)<br/>(pure, new module)"]
    F --> G{"any anchors within<br/>80px x / 18px y?"}
    G -- no --> H["sy unchanged: no-op re-apply"]
    G -- yes --> I["stack cluster vertically<br/>by STACK_OFFSET_PX"]
    I --> J["re-apply nameplateScreenTransform<br/>with adjusted sy"]
    H --> K["frame done"]
    J --> K
```

## Test plan

- [x] `npx vitest run tests/nameplate_declutter.test.ts` (6 new unit tests:
      untouched when separated, separates near-coincident anchors, separates
      anchors whose wide labels would overlap despite tens-of-px anchor
      spacing, stacks a 3+ cluster without unbounded growth, stable ordering
      regardless of input order, does not mutate input)
- [x] `npx vitest run tests/nameplate_projection.test.ts tests/architecture.test.ts tests/sim.test.ts` all green
- [x] `npm run ci:changed` (Biome) clean on touched files
- [x] Live-verified before/after in a mobile-landscape browser session (offline character)

### DIFF
--- a/src/render/nameplate_declutter.ts
+++ b/src/render/nameplate_declutter.ts
@@ -1,0 +1,55 @@
+// Pure post-projection pass: nudges apart nameplates whose screen positions
+// would otherwise fully overlap (e.g. two same-named mobs standing close
+// together). Most visible on short mobile-landscape viewports, where entities
+// need to be much farther apart in world space before their projections
+// separate on their own. DOM/Three-free so it unit-tests directly.
+
+export interface NameplateAnchor {
+  id: number;
+  sx: number;
+  sy: number;
+}
+
+// Anchors within this horizontal distance are treated as colliding: nameplate
+// labels render much wider than the anchor point itself (name + level + hp
+// bar), so this approximates half of a typical label's on-screen width rather
+// than the anchor point spacing.
+const OVERLAP_THRESHOLD_X_PX = 80;
+// Vertical anchors this close are considered the "same row" (labels are a
+// single text line anchored at their bottom, so the tolerance is much
+// tighter than the horizontal one).
+const OVERLAP_THRESHOLD_Y_PX = 18;
+// Vertical gap applied between stacked members of a cluster.
+const STACK_OFFSET_PX = 20;
+
+export function declutterNameplates(anchors: NameplateAnchor[]): NameplateAnchor[] {
+  const out = anchors.map((a) => ({ ...a }));
+  const byId = new Map(out.map((a) => [a.id, a]));
+  const visited = new Set<number>();
+
+  // Stable order regardless of input/render order, so the same entities
+  // always stack the same way frame to frame.
+  const ordered = [...out].sort((a, b) => a.id - b.id);
+
+  for (const anchor of ordered) {
+    if (visited.has(anchor.id)) continue;
+    const cluster = ordered.filter(
+      (other) =>
+        !visited.has(other.id) &&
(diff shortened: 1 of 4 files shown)


jgyy:
Fixed in 194dc012: `declutterNameplates()` now builds the connected component over the overlap graph via BFS instead of only clustering anchors that directly overlap the seed anchor, so transitive chains (A-B-C where A↔B and B↔C overlap but A↔C doesn't) get stacked together instead of leaving B/C still colliding.

Added a regression test (`stacks a transitive chain where the endpoints do not directly overlap`) with anchors at sx 0/70/140 on the same row, asserting all three pairwise vertical separations clear the 18px threshold.

Verified:
- `npx tsc --noEmit` clean
- `npx vitest run tests/nameplate_declutter.test.ts tests/nameplate_projection.test.ts tests/nameplate_view.test.ts tests/architecture.test.ts` — 58 tests passed
- `npm run ci:changed` (Biome) clean on touched files
- `npm run build` succeeds
- CI green on the pushed commit

PR #1475, opened 2026-07-05
feat(ui): add optional Target of Target frame

Blaine1705:
## Summary

Adds the classic "target of target" mini-frame as an Interface option (off by default). When you have a target, a small unit frame under the target frame shows who your target is targeting (a mob's aggro target, a player's selected target). Enable it in Options (Esc), Interface tab, "Show Target of Target".

## How it works

- **Reuses the unit_frame family** as a third `UnitFramePainter` instance (portrait + name + hp bar). No new core: it drives the same `unitFrameView` + painter as the player and target frames, painted only when the option is on and the target-of-target entity is known.
- **Resolution rule** lives in a new pure `targetOfTargetId(entity)`: a mob/npc (pets are mobs) uses `aggroTargetId` (its combat target), a player/bot uses `targetId`. Both fields already ride the wire, so the resolution is host-agnostic (identical offline and online).
- **Wire parity:** a player's `targetId` now rides the per-entity snapshot as `tgt` (encoded in the server `dynamicFields`, decoded in `ClientWorld.applyWire`), so the frame works for a player target online exactly as `aggro` already enabled it for a mob target. The target-of-target can be outside the player's ~120yd interest bubble, so the frame hides gracefully when the id resolves to no known entity.
- **Placement:** a `#target-frame` child, anchored under the frame's right edge so it clears the debuff strip that hangs below-left, shrunk with zoom so it stays proportional and hides with the whole frame when there is no target.

## i18n

New `hudChrome.options.showTargetOfTarget` label plus a non-wordy `unitFrame.targetOfTargetLabel` aria name (English-fill safe under M16), with es and the five non-Latin fills. Resolved tables regenerated.

## Testing

- `tsc --noEmit` clean; `biome ci --changed` clean.
- Guard suites green: architecture (determinism + sim purity), snapshots + bandwidth (wire), i18n completeness + status registry + localization_fixes (S3), hud_perf_budget, css_corpus + styles_extraction, options_view ordering, settings, client_shell.
- New unit test `tests/target_of_target.test.ts` pins the resolver rule (player uses targetId, mob/npc uses aggroTargetId, null when untargeted).
- Verified end to end in a real browser: desktop, and mobile in both portrait and landscape. The frame renders below the target frame, legible and clear of the on-screen joysticks, action buttons, and minimap in both orientations.

## Notes

Default is off (opt-in), consistent with the other new Interface toggles. Flip `showTargetOfTarget` def to true if an on-by-default is preferred.


Blaine1705:
Sin título

EnriqueGF:
@Blaine1705 If you resolve I just merge

jgyy:
Closing as superseded: the optional Target of Target frame already exists at src/ui/target_of_target.ts on main. Thanks for the contribution.

PR #1476, opened 2026-07-05
fix(render): stop sparsely updated characters strobing their orientation

madmatah:
## Summary

Some idle characters and mobs "flicker", appearing to change orientation every frame. A production capture (a per-frame probe recording rendered rotation vs wire facing for every visible entity) pinned two client-side interpolation defects plus one sim-side degenerate case; server data was clean in every captured event.

1. **Replayed transitions on unknown-cadence entities (the strobe).** `ClientWorld` only learns an entity's update cadence (`netInterval`) from record gaps in (5, 450) ms, so an entity whose only updates are sparse (an idle mob whose sole record in seconds is a wander turn) never gets one. The renderer fell back to the global frame alpha, which cycles 0 to 1 every sim tick: the entity's last facing/position transition was replayed every tick until its next record, turning one turn into a continuous pirouette. Fixed by interpolating unknown-cadence entities on a fixed 120 ms fallback interval so the transition plays once and saturates (`src/render/net_interp_core.ts`, a new registered pure core with unit tests).
2. **Facing extrapolation overshoot.** The facing lerp shared the position path's 1.25 extrapolation cap, so on sparsely updated entities every turn overshot by 25% and snapped back on the next record. Facing is now clamped to alpha 1 (positions keep 1.25 on measured cadences only), and `ClientWorld.applyWire` re-anchors `prevPos`/`prevFacing` on the same fallback clock so the anchor is the pose the renderer actually drew. `prevFacing` is also wrapped into (-PI, PI] (it grew by 2 PI per revolution over long sessions).
3. **Sim hardening: bearings from sub-epsilon deltas.** Reproduced in headless: any per-tick `facing = angleTo(a, b)` write strobes +-PI per tick once the target stands within noise distance of the entity (atan2 of a near-zero delta is direction noise, not a bearing). A new pure helper `steadyAngleTo(from, to, current)` holds the current facing below `FACING_HOLD_DIST` (0.1 yd) and now guards every per-tick facing write (mob attack/rooted/flee, profiled mob combat, ranged pet attack, pet melee, delve companion, /follow, charge finish, fiesta bots). The Backstab behind-check treats sub-epsilon range as in front so the held facing cannot be exploited point-blank.

Diagnosis details: the production capture decodes exactly against mechanism 1 (`rendered = prevFacing + wrappedDelta * cyclingAlpha` reproduces every logged sample); 43 of 66 captured events were mechanism 1, the other 23 mechanism 2, none showed server-side facing alternation.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npm test` green on this branch (full suite, including `tests/parity` and `tests/architecture.test.ts`)
  - `npx tsc --noEmit` clean
  - `npm run build` green
  - New regression tests: `tests/net_interp.test.ts` (fallback clock saturates instead of cycling, facing alpha clamp, `prevFacing` stays bounded across 8 full revolutions) and `tests/facing_stability.test.ts` (steadyAngleTo unit behavior plus a Sim integration repro: a mob whose target dithers millimetrically on top of it no longer strobes; pre-fix this flipped +-PI every tick)
- Manual steps:
  - Reproduced the strobe on production idle mobs with a read-only console probe (rendered rotation sawtooths while the wire facing changes once); the same probe on the patched local client observed 284 mob turns over 90 s with zero flicker events
  - QA gate: qa-checklist, architecture-reviewer, cross-platform-sync, and a correctness coverage pass; all findings fixed or noted below

## Review notes

- Self facing no longer extrapolates past its target (it shared the remote path's clamp). Self position keeps its lead; the yaw trails it by at most 25% of a snapshot. Intended.
- `net/` cannot import `render/`, so the fallback-clock constants are mirrored between `applyWire` and `net_interp_core.ts` with LOCKSTEP comments on both sides.

### Quality

- [x] **Tests pass.** `npm test` is green. I added or updated tests for any `sim/` or `server/` behavior I changed.
- [x] **Builds succeed.** `npm run build` (server/env bundles untouched).

### Cross-platform

- [x] Tested on desktop and mobile

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the one touched error emit, "You must be behind your target.", is an existing localized literal, unchanged).

### Hygiene

## Review notes

- The point-blank Backstab gate is a small, intended gameplay change, not only a
  rendering fix. Entities can overlap (`PLAYER_BODY_RADIUS` is prop/terrain collision,
  not entity separation), so a rogue stacked within 0.1 yd of the target now always gets
  "You must be behind your target." At 0.1 yd the target is not meaningfully behind and
  the old `angleTo` behind-check was already noise there; documented at the emit site.


FernandoX7 (commented):
Nicely diagnosed. Splitting this into three distinct mechanisms (replayed transition on unknown-cadence entities, facing overshoot, and sub-epsilon bearing noise) makes it easy to follow, and each fix lands where a seam already exists: a pure `net_interp_core.ts` registered in `RENDER_PURE_CORES`, a `steadyAngleTo` helper next to `angleTo`, and no new `Entity` fields to keep in sync.

I checked the load-bearing pieces against the code:

- The unknown-cadence fallback actually engages. `applyWire` sets `e.netUpdatedAt = now` on every record (`online.ts:1539`) while `netInterval` only learns from gaps in (5, 450) ms (`online.ts:1535`), so a quasi-idle entity keeps `netInterval` undefined but a defined `netUpdatedAt`, which is exactly the case `remoteEntityAlpha` routes onto the 120 ms fallback.
- The renderer/applyWire lockstep is tighter now, not just preserved. `applyWire` already clamped the facing re-anchor to 1 (`entFacingAlpha = Math.min(1, entAlpha)`), but the renderer drew facing with the raw up-to-1.25 alpha. `facingAlpha(ea)` at `renderer.ts:3972` closes that pre-existing gap, so the drawn pose and the re-anchor now agree for both self and remote.
- Determinism holds. `steadyAngleTo` is a pure function of the two positions plus the current facing (no `Rng`, no wall-clock), and every idle per-tick `facing = angleTo(...)` write is converted. The two `angleTo` facing writes left in place (`sim.ts:3172` charge waypoint, `fiesta_bots.ts:141` ring-center) both aim at a point guaranteed more than ~1 yd away, so they cannot reach the sub-epsilon case. Correct to leave them.
- Green at the head ref locally: `tests/parity` (162), `tests/architecture.test.ts`, and both new suites (32). No golden churn, so the sim facing change does not perturb the recorded scenarios.

Two notes, neither blocking:

- nit (`casting_lifecycle.ts:344`): the point-blank Backstab gate is a small gameplay change on top of the anti-strobe work, not just a rendering fix. Entities can overlap (`PLAYER_BODY_RADIUS` is prop/terrain collision, not entity separation), so a rogue stacked within 0.1 yd of the target now always gets "You must be behind your target." It reads correct to me (at 0.1 yd the target is not meaningfully behind, and the old `angleTo` behind-check was already noise there) and it is documented in the comment, so this is just worth a line in the PR body as an intended behavior change.
- Merge state: this rebases clean against the current `release/v0.23.0` tip (the base moved past the branch point, but nothing conflicts, and the UNSTABLE flag looks like CI rather than a conflict). Worth a rebase-and-retest before merge since goldens and parity depend on the base.

Looks good to me.

PR #1479, opened 2026-07-05
feat(render): realistic jagged mountains and distinct biome palettes

jgyy:
## Summary

More realistic mountains and more distinct biome ground palettes, purely in the
render layer (no gameplay/collision change beyond the heightfield's own noise,
which is still deterministic and covered by the existing terrain tests).

- **Jagged ridges and rim wall.** The inter-zone ridge walls and the world-rim
  wall (`src/sim/world.ts`) get a two-layer jagged crest (coarse peak/saddle +
  fine crag) instead of a perfectly smooth berm, so mountains read as real
  terrain instead of a rounded wall. Combined noise variance is kept small so
  the climb-limit invariant (`tests/terrain_walls.test.ts`) still holds.
- **Tighter wilderness LOD** (`src/render/terrain.ts`) so the extra height
  detail actually resolves into visible geometry instead of getting smoothed
  away by coarse vertex spacing.
- **Striated rock and patchy snow**: noise breaks up what was previously a flat
  rock/snow color blend, plus a finer fracture pass on the procedural rock
  texture (`src/render/textures.ts`).
- **Atmospheric rim haze**: `BIOME_FOG` near/far trimmed and the rim haze blend
  strengthened, so a zone's own mountains fade into haze before they can be
  seen crisply from the zone's hub (`src/render/renderer.ts`).
- **More distinct biome palettes**: widened the ground-color contrast between
  vale/marsh/peaks and the paint-only beach/desert/volcano/cave biomes (marsh
  darker and murkier, peaks cooler and greyer, desert warmer than beach,
  volcano red-ash vs cave neutral grey) so each biome reads as visually
  distinct rather than variations on the same green.
- **Organic ground noise**: replaced the old pure-sine grass patchiness (which
  repeats on a visible grid at a distance) with fbm noise.
- **Per-biome shoreline blend**: previously every biome faded to the same sand
  color at the waterline. Marsh now blends to mud (no sandy beach), and
  peaks/volcano/cave blend to a dark wet-rock tint instead of sand.

Parity goldens regenerated: entity fall height/position near the world rim
shifts by a few units from the jagged rim crest and the tighter LOD spacing;
no other sim behavior changes.

```mermaid
flowchart TD
    A["sim/world.ts terrainHeight()"] -->|"jagged ridge + rim crest (fbm2 noise)"| B["heightfield (shared by sim + render)"]
    B --> C["render/terrain.ts sampleVertex()"]
    D["BIOME_PALETTE per-biome colors"] --> C
[...]

PR #1481, opened 2026-07-05
feat(render): more realistic, jagged mountains with reduced draw-in haze

jgyy:
## Summary

- **Terraced mountain slopes.** The inter-zone ridge walls and world-rim wall now rise in stair-stepped bands (flat treads + steep risers) instead of one smooth ramp, so they read as a stacked rocky mountainside rather than a uniform incline. This is a pure reshaping of the already-computed ridge/rim height addition: terrain away from any ridge/rim is completely unaffected, and every straight crossing still meets a riser at least as steep as the ramp it replaces (`tests/terrain_walls.test.ts`, which checks max steepness along a crossing, stays green).
- **Jagged ridge + rim crest** (two-layer fbm noise) instead of a perfectly smooth berm.
- **Tighter wilderness LOD** so the extra height/terrace detail resolves into visible geometry instead of getting smoothed away by coarse vertex spacing.
- **Much more distinct biome ground palettes** - marsh, peaks, desert, volcano, and cave now read as clearly separate biomes rather than variations on the same green (pushed further than an earlier pass specifically for visibility).
- **Organic ground noise**: replaced the old pure-sine grass patchiness (visible-grid repeat at a distance) with fbm noise.
- **Per-biome shoreline blend**: marsh blends to mud (no sandy beach), peaks/volcano/cave blend to a dark wet-rock tint instead of sand.
- **Striated rock, patchy snow**, finer rock-texture fracture pass.
- **Atmospheric rim haze**: trimmed `BIOME_FOG` + strengthened rim haze blend so a zone's own mountains fade into haze before they're crisp from the zone's hub.

This PR **supersedes #1479** (closing that one) - it started as the same idea in a separate parallel session, so this combines both branches' work: the terracing, the amplified biome-palette contrast, and a capture harness that shows the real max-graphics look with the tutorial/cinematic out of frame are new on top of what #1479 had.

Parity goldens regenerated: entity fall height/position near ridges/rim shifts a few units from the new noise and terracing; no other sim behavior changes (rng draw order for the terrain math itself is unaffected, only cascading position/fall-timing effects near mountains).

## How the terracing works

```mermaid
flowchart TD
    A["terrainHeight(x, z, seed)"] --> B["baseHeight: rolling hills (unaffected)"]
    B --> C["mountainAdd = 0"]
    C --> D["+= jagged ridge crest\n(coarse fbm + fine fbm)"]
    D --> E["+= jagged rim crest\n(coarse fbm + fine fbm)"]
    E --> F["terraceStep(mountainAdd, STEP=6, TREAD=0.6)"]
    F -->|"flat tread for 60% of each 6yd band,\nthen a smoothed riser"| G["h += terraced mountainAdd"]
    G --> H["tests/terrain_walls.test.ts:\nmax steepness along any crossing\nstill clears the climb limit"]
```

### Ridge close-up (Eastbrook Vale pass) - terraced rock face vs the old smooth ramp
ridge closeup before/after

### World-rim close-up
rim closeup before/after

### Marsh hub (Fenbridge) - hazier ridge, darker/muddier ground
marsh hub before/after

### Vale hub (Eastbrook)
vale hub before/after

### Peaks hub (Highwatch), near the world rim - cooler/greyer ground palette
peaks hub before/after

## Test plan

- [x] `npx vitest run tests/terrain_walls.test.ts` - climb-limit invariant passes (terracing only makes risers steeper, never shallower, at the points this test samples)
- [x] `npx vitest run tests/parity/parity.test.ts` - passes after regenerating the seven affected goldens (`UPDATE_PARITY=1`)
- [x] `npm test` (full suite, 7800+ tests) - all green (a handful of unrelated tests hit the default 5s timeout only under heavy parallel machine load; confirmed passing in isolation with `--testTimeout=20000`)
- [x] `npx tsc --noEmit` - clean
- [x] `biome check` on the changed files - no new errors
- [x] Verified live in `npm run dev` (offline mode, `?gfx=ultra`) at every zone hub and at close-up ridge/rim vantage points; screenshots above

Screenshot/comparison harness at `scripts/mountain_realism_shot.mjs` (forces max gfx tier, waits out the intro cinematic, dismisses the tutorial).

### DIFF
--- a/scripts/mountain_realism_shot.mjs
+++ b/scripts/mountain_realism_shot.mjs
@@ -1,0 +1,79 @@
+// Screenshot harness for the mountain-realism / draw-distance-haze pass.
+// Boots an offline warrior, god-mode teleports to each zone hub (~"zone
+// centre") plus a couple of up-close mountain vantage points, hides the HUD,
+// and captures a frame at each. Needs `npm run dev` on :5173 (override with
+// GAME_URL). Writes to tmp/, filenames prefixed by the LABEL env var (default
+// "after") so before/after pairs don't collide.
+
+import fs from 'node:fs';
+import puppeteer from 'puppeteer-core';
+import { BROWSER_PATH } from './browser_path.mjs';
+
+const URL = process.env.GAME_URL ?? 'http://localhost:5173';
+const LABEL = process.env.LABEL ?? 'after';
+fs.mkdirSync('tmp', { recursive: true });
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+// zone hubs (data.ts / content/zone*.ts), "standing at the centre of the
+// zone" vantage points, plus two up-close mountain details.
+const VANTAGE_POINTS = [
+  { name: 'vale-hub', x: 0, z: 0, yaw: 0 },
+  { name: 'marsh-hub', x: 0, z: 300, yaw: Math.PI / 2 },
+  { name: 'peaks-hub', x: 0, z: 660, yaw: Math.PI },
+  { name: 'ridge-closeup', x: 0, z: 160, yaw: 0 },
+  { name: 'rim-closeup', x: 0, z: -140, yaw: 0 },
+];
+
+const browser = await puppeteer.launch({
+  executablePath: BROWSER_PATH,
+  headless: 'new',
+  args: ['--window-size=1600,900', '--use-angle=swiftshader', '--enable-unsafe-swiftshader'],
+  defaultViewport: { width: 1600, height: 900 },
+});
+const page = await browser.newPage();
+page.on('pageerror', (e) => console.log('PAGEERROR:', e.message));
+page.on('console', (m) => {
+  if (m.type() === 'error') console.log('CONSOLE:', m.text());
+});
+
+await page.goto(URL, { waitUntil: 'networkidle0', timeout: 30000 });
+await page.evaluate(() => document.querySelector('#btn-offline').click());
+await sleep(200);
+await page.type('#char-name', 'Aldwyn');
+await page.click('#offline-select .mini-class[data-class="warrior"]');
+await page.click('#btn-start-offline');
+await page.waitForFunction(() => Boolean(window.__game?.sim), { timeout: 20000 });
+await sleep(1000);
+
+// god mode so the settle-fall from teleporting never kills the camera
+await page.evaluate(() => {
+  const p = window.__game.sim.player;
+  p.maxHp = 99999;
+  p.hp = 99999;
+});
+
+// hide every HUD chrome element so the screenshot is pure world/terrain
+await page.evaluate(() => {
+  const style = document.createElement('style');
+  style.textContent = `
+    #hud, #minimap, #chat, #bags, #action-bars, #nameplates, #cast-bar,
+    #player-frame, #target-frame, #party-frame { display: none !important; }
+  `;
+  document.head.appendChild(style);
+});
+
+for (const v of VANTAGE_POINTS) {
+  await page.evaluate((vp) => {
+    const g = window.__game;
+    const p = g.sim.player;
+    p.pos.x = vp.x;
+    p.pos.z = vp.z;
+    p.facing = vp.yaw;
+    for (let i = 0; i < 10; i++) g.sim.tick();
+  }, v);
+  await sleep(1200); // let terrain chunks + fog settle after the teleport
+  await page.screenshot({ path: `tmp/mountains-${LABEL}-${v.name}.png` });
+  console.log('captured', v.name);
+}
+
+await browser.close();
--- a/scripts/mountain_realism_shot.mjs
+++ b/scripts/mountain_realism_shot.mjs
@@ -9,7 +9,9 @@ import fs from 'node:fs';
 import puppeteer from 'puppeteer-core';
 import { BROWSER_PATH } from './browser_path.mjs';
 
-const URL = process.env.GAME_URL ?? 'http://localhost:5173';
+// ?gfx=ultra forces the top graphics tier regardless of the swiftshader
+// software-GL auto-detect, so screenshots show the real max-quality look.
+const URL = (process.env.GAME_URL ?? 'http://localhost:5173') + '?gfx=ultra';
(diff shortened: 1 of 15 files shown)


FernandoX7 (commented):
The shape of this is right. The height math stays in one place (`terrainHeight` in `src/sim/world.ts`) and the renderer samples it directly in `sampleVertex` (`src/render/terrain.ts:309-311`), so the terracing flows to every host with no duplicate-formula drift. The parity regen checks out: the golden diffs shift only entity `y` (x/z untouched), and the per-snapshot `events` hash and `rng.draws` count are unchanged, so this is terrain height moving under existing entities and not a behavior change. The impassable-wall invariant also holds: `tests/terrain_walls.test.ts` is green (terracing only makes risers steeper), and `terraceStep` is continuous and monotonic. The render side (fog trim, LOD, palette, shoreline, rock streak) is gameplay-neutral, and the fog change actually narrows the high/low tier gap rather than widening it.

One blocking issue.

**blocking - the terracing flattens the western Mirefen wall base, and `tests/impact_site.test.ts` fails at head.**
`terraceStep` (`src/sim/world.ts:95`) maps any combined mountain rise below `TERRACE_TREAD * TERRACE_STEP` (0.6 * 6 = 3.6 yd) to exactly 0, so the foot of every ridge/rim now sits on a flat apron. The Mirefen impact crater is a placed landmark right at that foot, and the wall it leans against got erased there:

```
tests/impact_site.test.ts > is placed around Brother Aldric's reported western Mirefen wall base
AssertionError: expected -3.107 to be greater than -1.796
  const wallY = terrainHeight(MIREFEN_IMPACT_SITE.x + 4, ...); // -3.107
  expect(wallY).toBeGreaterThan(impactY + 1.5);               // impactY -3.296, needs > -1.796
```

The wall rises ~0.19 yd over that 4 yd span now, where it used to clear 1.5. I confirmed base (`release/v0.22.0`) passes this test 8/8 and the PR head fails it, so the terracing is the cause. It is deterministic (SEED=42), not one of the parallel-load timeouts the test plan mentions, so the "npm test all green" line is not accurate here. This one is a hand-written landmark invariant, not a golden, so `UPDATE_PARITY=1` never touched it, which is likely why it slipped past.

Fix direction is a maintainer call: either keep a little rise at the base (a smaller dead-zone / a small linear floor under band 0 in `terraceStep`) so the impact-site wall still reads, or accept the flat apron and reconcile the landmark placement + the test. Either way the failing test needs to go green before merge, not just be regenerated.

**nit - no direct unit test for `terraceStep`.** It is clean pure math, but the base-flattening dead zone (the exact behavior that broke `impact_site`) is the kind of edge worth pinning directly (`terraceStep(0) === 0`, the sub-3.6yd zeroing, monotonicity) once the fix lands, rather than only exercising it through the terrain sampler.

Everything else looks good to me.

FernandoX7:
Pushed three commits addressing the review, plus one regression found along the way.

**0b5bc97e3 fix(sim): talus apron under the first terrace band** (the blocking finding + the nit)

Went with the "keep a little rise at the base" direction: `terraceStep` now keeps a linear apron floor under band 0 (`TERRACE_APRON = 0.5` of the smooth rise, capped at `step * apron`), so wall feet stay sloped instead of flattening to a dead-level plain. The raw rise at the impact-site wall sample is 3.13yd, entirely inside the old 3.6yd dead zone; the apron keeps 1.57yd of it and `tests/impact_site.test.ts` passes again with margin (wall rise 1.755 vs the required 1.5). The floor provably never reaches past band 0 (every higher band already sits at `>= step`) and never reduces riser steepness (max only adds height below the first riser), so the `terrain_walls` climb contract is untouched. Also adds the direct unit test from the nit (`tests/terrace_step.test.ts`): zero fixpoint, apron values, band-0 confinement, flat treads, riser-vs-ramp slope, monotonicity and continuity, with the production constants exported and pinned as literals.

One golden regenerated (`fiesta_midcast_kill`, draws 523 to 524). Verified draw-by-draw with an rng observer: the first 523 draws are call-site-identical and the extra draw is one appended mob wander roll at the tail (a world mob near a wall foot now stands on the apron slope). Every snapshot state/events digest is unchanged.

**08fc63639 chore(render): biome-format textures.ts** (the Codex high finding)

Format-only pass on the one file. The exact CI command, `npx @biomejs/biome ci --changed --since=origin/release/v0.22.0 --no-errors-on-unmatched`, now exits 0.

**6ec11656b fix(sim): fade mountain character past the rim** (a regression the review round had not caught)

`tests/follow.test.ts` fails 2 of 5 cases at the PR head and passes at base: it walks a follower along z = -1000, and the new rim crest noise plus terracing turned the plateau beyond the world rectangle into terraced crags with risers past the climb limit. That overshoot space is never rendered (the terrain mesh covers exactly the world rect) and never reachable in play, but the sim relies on it as a flat staging ground: the follow and chat tests, `scripts/chat_e2e.mjs`, and six parity scenarios that stage mobs about 20yd past the x-rim.

The fix fades the crest noise and terracing back to the original smooth berm over 2..10yd past the edge, written so every in-world sample is bit-identical (each noise term is multiplied by the fade separately rather than regrouping sums, and the 2yd full-detail skirt covers the renderer's `SLOPE_EPS = 1.5` normal sampling, so no rendered vertex, height or normal, changes at all). A nice side effect: five of the seven regenerated goldens (`delve_companion`, `dungeon_instances`, `fiesta_powerups`, `mob_lifecycle`, `mob_locomotion`) revert to byte-identical with `release/v0.22.0`; their only drift had been those staged overshoot entities. `pet_ai` keeps a small in-world diff (a mob on the rim wall face itself, the intended terrain change). A new `terrain_walls` case pins the overshoot plateau staying below steepness 1.0 at both the production and test seeds, and fails on the pre-fix code.

**Verification**: replicated the CI pipeline end to end in a clean worktree (this branch predates `npm run gate`): i18n gen + freshness check, security gate, full suite (7859 passed, including `follow` and `impact_site`), `tsc`, changed-file biome, and all three builds, all green. Also re-ran the capture harness: the terraced cliff faces read the same as the PR screenshots, and the impact site leans against a natural talus slope again.

PR #1482, opened 2026-07-05
fix(character-select): improve responsive layout and readability

madmatah:
## Summary

The character selection screen did not scale cleanly across desktop and mobile viewports. Wide displays left excessive empty space around the character, long names were squeezed by the action buttons, and the class details panel was difficult to read. On phones, rotating from portrait to landscape and back could also leave the page zoomed to the stale landscape width.

This PR:

- centers the character preview and balances both side panels;
- scales panel widths, spacing, and typography across desktop resolutions;
- gives character names more room and stacks row actions vertically;
- reorganizes starting stats and equipment into readable full-width sections;
- replaces equipment badges with a simpler typographic presentation;
- adds a subtle scrollbar for overflowing class details;
- extends the cinematic desktop layout down to 861px while preserving the touch layout;
- normalizes `visualViewport` dimensions by its scale so portrait layout is restored after rotating a phone back from landscape;
- adds regression coverage for the responsive shell and stale mobile viewport scale.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:

- Manual steps:
  - Verified the character selection screen at 768×700, 899×700, 900×700, 1024×768, 1366×768, 1440×900, 1920×1080, and 2560×1440.
  - Verified Retina layouts for MacBook Air 13/15 and MacBook Pro 14/16 using DPR 2 viewports.
  - Verified long character names, centered preview framing, class-detail scrolling, and responsive typography.
  - Reproduced a phone rotation sequence at 390×844 → 844×390 → 390×844 without reloading.
  - Confirmed the restored portrait viewport returns to 390×844 at scale 1 instead of retaining the landscape width.

## Screenshots / recordings

**desktop 2560x1440**

desktop-2560x1440

**desktop 1920x1080**

desktop-1920x1080

**desktop 1440x900**

desktop-1440x900

**desktop 1366x768**

desktop-1366x768

**compact**

desktop-compact-768x700

---

### Quality

      TypeScript `strict`).

### Cross-platform

- [x] **Accessible.** No interactive semantics were changed. Existing keyboard
      operation, focus styles, ARIA, and reduced-motion behavior are preserved.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/game/app_viewport.ts
+++ b/src/game/app_viewport.ts
@@ -9,18 +9,16 @@ export function syncAppViewport(win: Window = window): void {
   const doc = win.document;
   const useStableGameViewport =
     doc.body.classList.contains('game-active') && useTouchInterface(win);
-  const width = Math.max(
-    1,
-    Math.round(
-      useStableGameViewport ? win.innerWidth : (win.visualViewport?.width ?? win.innerWidth),
-    ),
-  );
-  const height = Math.max(
-    1,
-    Math.round(
-      useStableGameViewport ? win.innerHeight : (win.visualViewport?.height ?? win.innerHeight),
-    ),
-  );
+  const visualScale = win.visualViewport?.scale ?? 1;
+  // visualViewport dimensions are expressed inside the current page scale.
+  // Normalize them back to layout CSS pixels before writing html/body's fixed
+  // dimensions. Without this, a landscape-to-portrait rotation can feed the
+  // landscape width back into --app-vw while the browser zooms the page down,
+  // permanently trapping the portrait layout at the landscape scale.
+  const visualWidth = (win.visualViewport?.width ?? win.innerWidth) * visualScale;
+  const visualHeight = (win.visualViewport?.height ?? win.innerHeight) * visualScale;
+  const width = Math.max(1, Math.round(useStableGameViewport ? win.innerWidth : visualWidth));
+  const height = Math.max(1, Math.round(useStableGameViewport ? win.innerHeight : visualHeight));
   doc.documentElement.style.setProperty('--app-vw', `${width}px`);
   doc.documentElement.style.setProperty('--app-vh', `${height}px`);
 }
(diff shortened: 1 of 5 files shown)


FernandoX7 (commented):
Nicely scoped fix. I read it against the render/ui/mobile invariants and ran the two touched test files (70 pass); the load-bearing pieces hold up.

- The viewport change is correctly gated: `visualViewport.scale` defaults to 1 (`app_viewport.ts:12`), so `visualWidth * scale` (`:18`) is identical to the old path for desktop and normal-orientation mobile. Only the pinch-zoom / stuck-rotation case moves, and the new test pins the 844 to 390 recovery. On a trackpad pinch (scale != 1) writing the full layout width instead of the zoomed slice is the more correct result too, so no regression there.
- The cs-wow stage is a self-contained `position: fixed; inset: 0` takeover that out-importants the card-sizing block, and you re-declared the inner grids (`.class-details-grid`, `.char-row`, `.char-actions`, `.details-gear-row`) inside it, so widening the breakpoint to 861px does not lean on the 900px card rules. `--scrollbar-thumb` resolves (`tokens.css:172`).
- `LIVE_PREVIEW_X` is applied to both live `lookAt` calls and the post-capture restore, and the separate thumbnail-capture framing (`-0.1`) is correctly left alone. `preview.ts`.

One nit:

- nit: moving the cs-wow stage to `min-width: 861px` (`shell.css:2556`) leaves three sibling character-select blocks at `min-width: 900px` (`shell.css:4551`, `:5460`, `:5486`), so in the 861-899px window the stage renders without the card foundation underneath it. I traced it and it is safe (the takeover out-importants that block and you redeclared the inner grids), and you list 899x700 in the tested viewports, so this is just a readability note: a one-line comment on why the stage and the card foundation now diverge by 39px would save the next person the same trace.

Determinism / wire / i18n: nothing in scope (a client viewport util plus a render-only camera constant, no player-visible strings). Looks good to merge.

madmatah:
@FernandoX7 Thanks, I addressed your feedback :)

FernandoX7 (approved):

PR #1487, opened 2026-07-05
feat(character-select): UX improvements

madmatah:
## Summary

Two focused UX improvements to the online character-select screen, both scoped to the
desktop docked layout. Mobile and tablet are left exactly as they are.

**1. One prominent "Enter World" CTA**
Desktop used to show a separate "Enter World" button on every roster row, right next to that
row's "Delete", which made the primary action easy to misclick and visually noisy. This PR
replaces them with a single gold "Enter World" button pinned to the foot of the roster,
acting on the currently selected character:

- The button reflects the selected character's state: "Enter World" for a ready character,
  "Take Over" when the character is online in another session, and disabled (with a rename
  hint) while a forced rename is pending.
- Double-click a roster row to jump straight in (classic character-select muscle memory).
- Rows keep only "Delete", so the Delete/Enter adjacency misclick is gone.
- Styled to match the existing Donate / Verify Wallet gold CTA, so it reads as the screen's
  primary call to action.
- The button is pinned in the roster's action bar while the list scrolls inside its own area,
  so it stays visible no matter how many characters you have (you never scroll to play).

The classification that drives the label and the click routing is a single pure function
(`src/net/charselect_action.ts`, following the existing `char_sort.ts` pattern), so the label
and the action can never disagree. It is unit-tested.

**2. Character name over the 3D model**
The selected character's name now appears centered above the 3D preview, so it is
unmistakable which character you are about to play.

Both features are gated to the desktop docked stage. On mobile the roster is a vertical list
of cards that already show each character's name and its own per-row "Enter World" button, so
nothing there changes.

## Related issues

N/A

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:

- Manual steps:
  - Desktop at a wide (2000px) and a narrow (1100px) width: the single "Enter World" CTA sits at
    the foot of the roster with Back / New Character below, and per-row rows show only Delete.
  - Confirmed the CTA stays pinned and visible with the roster list scrolling internally, so
    entering the world never requires scrolling regardless of character count.
  - Confirmed the selected character's name renders centered above the 3D model and updates as
    you pick a different character; it truncates with an ellipsis so a long name cannot break
    the layout.
  - Verified the mobile fallback via computed-style checks (the reveal is `body:not(.mobile-touch)`
    gated): on `mobile-touch` the single CTA and the name plate stay hidden and the per-row
    buttons remain visible. A real phone pass in portrait and landscape is still worth a look.

## Screenshots / recordings

Screenshot from 2026-07-05 18-58-56

Screenshot from 2026-07-05 18-59-39

Screenshot from 2026-07-05 18-59-56

---

### Quality

- [x] **Tests pass.** 
- [x] **Typecheck passes.** 
- [x] **Builds succeed.** 

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop verified at wide and narrow widths. Mobile
      per-row fallback confirmed via computed-style checks (the new UI is `body:not(.mobile-touch)`
      gated and untouched on mobile); a real-device portrait/landscape pass is recommended.
- [x] **Accessible.** The CTA is keyboard-operable with a visible focus ring, has a proper
      disabled state, and the name plate uses `aria-live`. No motion added. Note: the shell has no
      `forced-colors` rules today (pre-existing, not introduced here).

### Localization (i18n)

- [ ] New player-visible strings are translated into every supported locale.
- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary.
- [x] **N/A. This PR adds no player-visible strings.** The button reuses existing keys
      (`auth.enterWorld`, `character.takeOver`, `character.renameRequired`) via `t()`, and the name
      plate renders the player's own character name (not a translatable string).

### Hygiene

      any production path.
- [x] I didn't hand-edit generated files. None were touched.

### DIFF
--- a/index.html
+++ b/index.html
@@ -869,6 +869,7 @@
             <!-- Right: selected character showcase -->
             <div class="cs-detail-col">
               <div class="char-preview-container cs-preview" id="online-preview-container">
+                <div class="cs-preview-name" id="charselect-preview-name" aria-live="polite"></div>
                 <canvas id="char-preview-canvas"></canvas>
               </div>
               <div id="charselect-class-details" class="class-details-panel" aria-live="polite"></div>
--- a/index.html
+++ b/index.html
@@ -861,6 +861,7 @@
             <div class="cs-list-col">
               <ul id="char-list"></ul>
               <div class="cs-list-actions">
+                <button type="button" class="btn btn-primary" id="btn-charselect-enter" data-i18n="auth.enterWorld">Enter World</button>
                 <button type="button" class="btn btn-secondary" id="btn-charselect-back" data-i18n="auth.back">Back</button>
                 <button type="button" class="btn btn-primary cs-new-btn" id="btn-new-character" data-i18n="auth.newCharacter">New Character</button>
               </div>
--- a/play.html
+++ b/play.html
@@ -707,6 +707,7 @@
             <!-- Right: selected character showcase -->
             <div class="cs-detail-col">
               <div class="char-preview-container cs-preview" id="online-preview-container">
+                <div class="cs-preview-name" id="charselect-preview-name" aria-live="polite"></div>
                 <canvas id="char-preview-canvas"></canvas>
               </div>
               <div id="charselect-class-details" class="class-details-panel" aria-live="polite"></div>
--- a/play.html
+++ b/play.html
@@ -699,6 +699,7 @@
             <div class="cs-list-col">
               <ul id="char-list" role="list"></ul>
               <div class="cs-list-actions">
+                <button type="button" class="btn btn-primary" id="btn-charselect-enter" data-i18n="auth.enterWorld">Enter World</button>
                 <button type="button" class="btn btn-secondary" id="btn-charselect-back" data-i18n="auth.back">Back</button>
                 <button type="button" class="btn btn-primary cs-new-btn" id="btn-new-character" data-i18n="auth.newCharacter">New Character</button>
               </div>
--- a/src/main.ts
+++ b/src/main.ts
@@ -3923,9 +3923,10 @@ async function refreshCharacters(): Promise<void> {
   const listEl = $('#char-list');
   listEl.innerHTML = `<li class="char-list-message">${escapeHtml(t('character.loading'))}</li>`;
   // Drop any stale selection from a previous realm; the default first-row
-  // selection below re-arms the shared Enter World button.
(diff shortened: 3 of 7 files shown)


FernandoX7 (commented):
Nicely structured. The classifier as a single pure function driving both the label and the click routing is the right call, the `index.html` / `play.html` DOM is mirrored and pinned by a new parity test, the CSS lives in the shared `shell.css`, and it reuses the existing `t()` keys rather than adding strings. The breakpoints leave exactly one entry affordance in every layout (base hides the shared button and shows the per-row ones; the `min-width: 861px` non-touch docked block flips that). 21 tests pass locally.

Two things:

**should-fix** `syncCharselectEnterButton` sets only `textContent` (src/main.ts:4114), but the button keeps a static `data-i18n="auth.enterWorld"` (index.html:864, play.html:702). A runtime language switch runs `translatePage()` (src/main.ts:4359, via `changeLanguage`), which re-applies every `[data-i18n]` key back onto `textContent`. So switching language from the footer picker while a character that is online elsewhere is selected clobbers the dynamic "Take Over" label back to "Enter World" until the next row click. Routing still re-runs the classifier on click so it does not misbehave, but the label goes stale, which is exactly the "label and action can never disagree" property this PR sells. The repo already solves this in `applyServerMode` (src/main.ts:6541): set the `data-i18n` attribute alongside `textContent`. Same fix here, `btn.setAttribute('data-i18n', action.labelKey)` in `syncCharselectEnterButton`. The rename-required `title` has the same gap (no `data-i18n-title`, so it stays in the old language after a switch); driving it the same way closes both.

**nit** The dblclick comment (src/main.ts:4026) says a double-click on a per-row control is safe because each one "opens a full-screen modal on the first click ... or is itself disabled". A forced-rename row also carries a rename input and a rename button, which are neither. The double-click is still safe there, but for a different reason: the shared Enter World button is disabled for a forced-rename selection and the handler guards on `!enterBtn.disabled`. Worth pointing the "keep that true for any per-row action added later" note at the real invariant (the shared button's disabled state is what gates entry).

madmatah:
@FernandoX7  Both addressed, thanks for the careful review.
And thanks @EnriqueGF for the 2nd review!

FernandoX7 (approved):

PR #1492, opened 2026-07-05
feat(character-select): render the real equipped weapon and event ski…

madmatah:
## Summary

The online character-select 3D preview now shows the selected character with the same
appearance it has in the world, instead of a generic class body holding the class's default
starting weapon. Concretely, the preview now reflects:

- **The actually-equipped mainhand weapon**, not the class default (an unarmed character shows
  no weapon, exactly as in-world).
- **Event skins:** a character wearing the class-agnostic Combat Mech cosmetic
  (`skinCatalog === 'mech'`) now renders the Mech body with its chroma, instead of falling back
  to the base class body with the chroma index misread as a class skin.

Regular class skins already rendered correctly and are unchanged.

**Why this is scoped to weapon + body, not full armor.** In this engine the only equipment that
is a dynamic per-item 3D model is the mainhand weapon; armor, helmets, capes, and shields are
baked into the class body or driven by a static allowlist, and body variation comes from skin
texture swaps, not per-armor meshes. That is true in-world too. So "the same head as in the
game" means: class rig or Combat Mech body, the appearance skin, and the equipped weapon. This
PR makes the preview match that exactly.

**How it works.**

1. **Data (REST, additive).** `characterListPayload` now also projects `skinCatalog` and
   `mainhandItemId` from the persisted character state, so both `GET /api/characters` and
   `GET /api/me/characters` carry them. `CharacterSummary` gains the two fields as optional.
2. **Rendering.** A new pure, DOM/Three-free helper `previewAppearanceVisual()`
   (`src/render/characters/preview_appearance.ts`) resolves an appearance to its concrete visual
   (body key, held weapon, hand-layout override), mirroring the in-world `createCharacterVisual`
   one-to-one. `CharacterPreview.setAppearance()` consumes it. The Mech cosmetic assets load
   lazily, so `setAppearance` shows the class body meanwhile and re-applies the Mech once loaded,
   guarded by an appearance signature so a newer selection is never stomped.
3. **Wiring (`src/main.ts`).** The char-select roster row selection, the return to the roster
   panel, and the token-auto-login init path all drive the preview from the character's real
   appearance; the roster fetch warms the Mech assets when any character uses them.

## Related issues

N/A

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:

- Manual steps:
  - launch server in dev mode
  - login with a warrior
  - `/dev level 20` , `/dev give wyrmfang_greatblade`, `/dev give alien_armor_plate`
  - Equip the gear
  - Logout
  - Go to the select screen verify that the preview show the character's actually-equipped weapon

## Screenshots
 
Screenshot from 2026-07-05 20-43-19

Screenshot from 2026-07-05 20-50-02

Screenshot from 2026-07-05 20-50-49

## Review notes

- **Mirrors the in-world render, does not invent one.** `previewAppearanceVisual` reproduces
  `createCharacterVisual` exactly (Mech to `player_mech` plus `mechHeldWeaponOverride(cls)` so a
  rogue Mech dual-wields; class to `player_<class>` with no override; weapon to the mainhand id).
  A non-weapon or unmapped mainhand converges to the class-default attach on both paths, so the
  ungated `state.equipment.mainhand` projection cannot diverge from the sim's `?.weapon`-gated
  `e.mainhandItemId`.
- **Dependency direction.** `render/` cannot import `ui/`, so the pure body-key decision lives in
  a small render-local module (`preview_appearance.ts`) rather than reusing the ui-side
  `activeCharacterAppearancePreview`; it is unit-tested directly.
- **Backwards compatible (both directions).** The server change is purely additive (two new JSON
  keys on an existing REST endpoint). An out-of-date client parses the list verbatim
  (`Api.characters` returns `data.characters`) and ignores unknown keys. An old server against a
  new client omits the fields, and the client defaults `skinCatalog` to `'class'` and
  `mainhandItemId` to `null` (the class body with no weapon, i.e. the previous behavior). The WS
  wire is untouched (in-world Mech rendering already ships via `cat: 'mech'`), so `IWorld`,
  `Sim`/`ClientWorld`, and snapshot parity carry no obligation.
- **Out of scope (intentional).**
  - The 2D roster row portrait chips still show class body + skin only, so a Mech character's chip
    is not the Mech body. Extending the offscreen headshot (`portrait.ts`) is a possible follow-up.
  - The offline character picker keeps the class-default path (there is no persisted REST roster
    to source a real appearance from).
  - The character-sheet paperdoll (`hud.ts` `renderCharPreview`/`mountCharPreview`) already
    implements the same Mech-aware logic against `CharacterPreview`; `setAppearance` is effectively
    its extraction, so a later PR can have the HUD consume it (rule of three) to remove the
    duplication.

### Quality

- [x] **Tests pass.** Added `tests/preview_appearance.test.ts` (the pure appearance resolver:
      class rig, unarmed, Mech body, rogue dual-wield override vs single-mainhand no-override, and
      signature sensitivity). Ran the relevant suites and the architecture guard locally (green);
      the full `npm test` runs in CI. No `sim/` or `server/` behavior changed (the server change is
      a read-only REST projection).
- [x] **Typecheck passes.** `npx tsc --noEmit` is clean (client and server).
- [x] **Builds succeed.** Not run locally; expected green in CI (no build inputs or generated
      files were touched).

### Cross-platform

- [x] **Tested on desktop and mobile.** This is a pure rendering-content change to the existing
      3D preview: no DOM, layout, or CSS change, so it is identical wherever the preview already
      renders. Verified on desktop; mobile structure is untouched.
- [x] **Accessible.** No UI/DOM/interaction change; the 3D canvas is out of the a11y scope (not
      screen-readable, never faked with aria), unchanged here.

### Localization (i18n)

- [ ] New player-visible strings are translated into every supported locale.
- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary.
- [x] **N/A. This PR adds no player-visible strings.** `skinCatalog` and `mainhandItemId` are
      opaque enum/id discriminators, never rendered as text.

### Hygiene

      any production path.
- [x] I didn't hand-edit generated files. None were touched.

### DIFF
--- a/server/main.ts
+++ b/server/main.ts
@@ -502,6 +502,8 @@ function characterListPayload(chars: CharacterRow[]): {
     forceRename: boolean;
     lastPlayed: string | null;
     playtimeSeconds: number;
+    skinCatalog: 'class' | 'mech';
+    mainhandItemId: string | null;
   }[];
 } {
   return {
@@ -516,6 +518,10 @@ function characterListPayload(chars: CharacterRow[]): {
       forceRename: c.force_rename,
       lastPlayed: c.last_played ? new Date(c.last_played).toISOString() : null,
       playtimeSeconds: Number(c.playtime_seconds ?? 0),
+      // Real appearance for the char-select 3D preview (the client renders the
+      // Combat Mech cosmetic body and the equipped mainhand, matching the world).
+      skinCatalog: c.state?.skinCatalog === 'mech' ? 'mech' : 'class',
+      mainhandItemId: c.state?.equipment?.mainhand ?? null,
     })),
   };
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -87,7 +87,8 @@ import {
 // the feature is enabled + used.
 import type { WalletOption } from './net/wallet';
 import { assetsReady } from './render/assets/preload';
-import { CharacterPreview } from './render/characters';
+import { CharacterPreview, type PreviewAppearance } from './render/characters';
+import { preloadMechAssets } from './render/characters/assets';
 import { skinCount } from './render/characters/manifest';
 import { playerPortraitDataUrl } from './render/characters/portrait';
 import { installWebGLContextRelease } from './render/context_release';
@@ -2804,11 +2805,16 @@ function updatePreviewContainer(panelId: string): void {
   characterPreview.setContainer(container);
 
   if (panelId === '#charselect-panel') {
-    // The selected roster row drives the showcase (class + that character's chroma).
-    const row = document.querySelector('#char-list .char-row.sel') as HTMLElement | null;
-    const cls = (row?.dataset.class as PlayerClass) ?? 'warrior';
-    characterPreview.setClass(cls);
-    characterPreview.setSkin(Number(row?.dataset.skin ?? 0) || 0);
+    // The selected roster row drives the showcase: its full real appearance
+    // (class or Combat Mech body + chroma + equipped mainhand), matching the world.
+    if (charselectSelected) {
+      characterPreview.setAppearance(charselectAppearance(charselectSelected));
+    } else {
+      const row = document.querySelector('#char-list .char-row.sel') as HTMLElement | null;
+      const cls = (row?.dataset.class as PlayerClass) ?? 'warrior';
+      characterPreview.setClass(cls);
+      characterPreview.setSkin(Number(row?.dataset.skin ?? 0) || 0);
+    }
     syncPreviewAfterPanelLayout();
     return;
   }
@@ -3829,6 +3835,10 @@ async function refreshCharacters(): Promise<void> {
   setCharselectPreviewName('');
   try {
     const chars = sortCharacters(await api.characters(), charSortMode);
+    // Warm the lazy Combat Mech cosmetic assets so selecting an event-skin
+    // character shows the mech body without a class-body flash (setAppearance
+    // falls back gracefully if this has not resolved yet).
+    if (chars.some((c) => c.skinCatalog === 'mech')) void preloadMechAssets();
     if (api.realm) $('#charselect-realm').textContent = api.realm;
     listEl.innerHTML = '';
     if (chars.length === 0) {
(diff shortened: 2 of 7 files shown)


FernandoX7 (commented):
Clean change, and the PR notes did a lot of my job: the parity argument, the both-directions back-compat, and the async-supersede guard all hold up. I checked the load-bearing claims against the code rather than the description.

- `previewAppearanceVisual` matches `createCharacterVisual` for every real class. `visualKeyFor` only differs from `player_<cls>` via its `player_warrior` fallback for an unknown `templateId`, which a valid `PlayerClass` never hits, and the mech override is the same `mechHeldWeaponOverride(cls)` call gated the same way. `src/render/characters/preview_appearance.ts:27`.
- The ungated `state.equipment.mainhand` projection cannot diverge from the sim's `?.weapon`-gated `e.mainhandItemId` (`src/sim/entity.ts:312`): the mainhand slot only ever holds a weapon id, and even a hypothetical non-weapon maps to a null model url and keeps the class-default attach, same result as `null` (`itemWeaponModelUrl`, `manifest.ts:260`). So `server/main.ts:463` is safe.
- The lazy re-apply terminates. `mechAssetsReady()` is already the in-world gate in `renderer.ts`, and the gltf/skin keys it checks are exactly what `preloadMechAssets` stores (`assetUrl` is identity for the mech, it is not in `LOW_URL_ALIAS`), so the `.then` to `setAppearance` retry flips ready and stops rather than looping. `src/render/characters/preview.ts:131`.
- Both `GET /api/characters` and `GET /api/me/characters` share `characterListPayload`, so both carry the two fields. `CharacterState.equipment` is required and `skinCatalog` optional, so the optional chaining is right. Server authority and token scoping are untouched (still `listCharacters(accountId)`), the projection is read-only, no DDL.

Green locally against the PR head: `tests/preview_appearance.test.ts` 5/5, `check:types` 0 errors (client, server, admin), and `biome ci` on the changed files is warnings-only (the one `arguments` warning is pre-existing `online.ts` code you did not touch).

Nits, non-blocking:

- `nit` The subtlest piece, the async mech re-apply plus the `appearanceSig` supersede in `setAppearance` (`src/render/characters/preview.ts:126`), is only reasoned about, not exercised. The test covers the pure `previewAppearanceVisual` / `appearanceSignature`, but not the stateful guard that keeps a stale re-apply from stomping a newer selection. Understood that it needs a live WebGL `CharacterPreview`, so it is awkward to unit-test; flagging it as the one untested branch.
- Leaving the `hud.ts` `renderCharPreview` dedup (rule of three) for a follow-up is the right call; `setAppearance` reads like the extraction target for it.

Nice work.

madmatah:
@FernandoX7 Thanks for flagging this. I added targeted coverage for the async mech re-apply and the supersede guard, including the stale-load race with a newer selection.

FernandoX7 (approved):

PR #1495, opened 2026-07-05
fix(ui): size the empty chat input to its placeholder so the hint never clips

### LINKED ISSUE #1232: Chat input box is an odd height and clips its placeholder text

MarchalMaxim:
## Summary

The chat input box was the wrong height in both directions (#1232): the empty box pinned itself to a single line, so the long default slash-command hint ("Say something... (/s say, /w name whisper, ...)") wrapped to two lines and the second line was clipped at the bottom edge. That pin was itself a workaround for the opposite symptom, Chromium inflating an empty textarea's scrollHeight from the placeholder and rendering an oversized empty box.

This PR makes the box track whatever it is actually displaying: while the input is empty, the placeholder is briefly swapped in as the value, measured, and restored, so the measurement is exact in every engine (the swap fires no input events and an empty textarea has no caret state to lose). Typed text behaves as before: single line when short, grows upward while it fits, scrolls internally past max-height.

The measure-and-apply routine moves out of `main.ts` into the existing `src/ui/chat_input_autosize.ts` module behind a structural `AutosizeTextarea` slice, so a unit test drives it with a hand-rolled fake (no jsdom) and `main.ts` keeps a one-line consumer.

Measured on the real client (headless, `scripts/chat_input_placeholder_shot.mjs`):

| state | before | after |
|---|---|---|
| empty + wrapping placeholder | height 36px, needs 49px (clipped) | height 49px (fits) |
| short typed text | 36px | 36px |
| long typed text | 49px | 49px |

## Related issues

Closes #1232

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/chat_input_autosize.test.ts` (12/12), `npm test` (7774 passed), `npx tsc --noEmit`, `npm run build`, `npm run security:gate`, `npm run ci:changed`
- Manual steps: booted the offline client and captured the bottom-left corner before/after via the new `scripts/chat_input_placeholder_shot.mjs` (needs `npm run dev`): opened the chat bar empty with the long hint, typed a short message, typed a long wrapping message, and compared rendered heights against `scrollHeight`.
- Mobile: the autosize path is shared; mobile keeps its own CSS placement and its `min-height: 40px` / 16px font floors still apply, so the only behavior change there is the same one: a wrapping placeholder now fits instead of clipping.

## Screenshots / recordings

Empty input showing the default hint. Before, the second line clips mid-glyph; after, the full hint is visible:

| before | after |
|---|---|
| placeholder before | placeholder after |

Typed text is unchanged (short stays one line, long grows upward):

| short typed (after) | long typed (after) |
|---|---|
| typed after | long after |

---

### Quality

- [x] **Tests pass.** `npm test` is green, and the new `autosizeChatInput` behavior is unit-tested (bug repro, value restore, precedence of typed text over the placeholder, growth and scroll clamp).
- [x] **Builds succeed.** `npm run build` passes (no server/headless code touched).

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop verified with before/after captures; the mobile chat input keeps its own CSS placement and floors (40px min-height, 16px font), which this change does not touch.
- [x] **Accessible.** No new interactive elements, focus behavior, or motion; the input only sizes to make its visible text readable.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the placeholder still comes from `hud.activeChatPlaceholder()`; the module treats it as an opaque, already-localized value it only measures).

### Hygiene


Rubsey:

Thanks for the contribution, @MarchalMaxim.

PR #734, opened 2026-06-19
feat(woc): $WOC rename + vanity names — implements #470 [plan/WIP]

awidearray:
# PR 1 — `$WOC` Rename + Vanity Names

> **Status:** DRAFT / implementation plan. Implements roadmap stub #470. Stacked on #473 (wallet-link). Flag-gated; no behavior change on `main` by default.

## Summary
Charge `$WOC` (real on-chain **burn**) for character/guild renames and reserved vanity names — the faithful #470: a pure token **sink**, cosmetic, no pay-to-win, Low reg risk.

This PR also lands the **shared `$WOC` payment layer** reused by the rest of the stack (PR 2 SNS subdomains, PR 3 marketplace).

## Foundation
Builds on **#473** (absorbed the closed #465 wallet-link): non-custodial wallet linking (`server/wallet.ts`, `wallet_link.ts`), `$WOC` balance reads (`server/woc_balance.ts`, `src/net/wallet.ts`), Reown client. #473 has **no** payment/burn verification or server-side tx signing — those are here.

## Shared payment layer (reused by PR 2 / PR 3)
- **Deps:** `@solana/spl-token` (SPL `burnChecked`), `naughty-words` (LDNOOBW global word list). `@solana/web3.js` already added by #473.
- **Config/env:** `WOC_TREASURY` (optional), `WOC_BURN_BPS` (default `10000` = 100% burn), `WOC_DECIMALS` (default read from mint), prices `WOC_PRICE_RENAME_CHARACTER` / `_RENAME_GUILD` / `_RESERVE`. Reuses `WOC_MINT` / `SOLANA_RPC_URL`.
- **DB** (idempotent DDL in `SCHEMA`, run by `ensureSchema()` under the existing advisory lock): `woc_payments` — replay guard + audit (`tx_sig TEXT UNIQUE NOT NULL`, `account_id`, `amount_base`, `mint`, `reference`, `created_at`). New `db.ts` fns: `recordWocPayment()` (null on UNIQUE conflict), `getCharacterById()`.
- **`server/woc_payment.ts`** (new): `verifyBurnTx(signature, {payer, minBurnBase, mint, memo})` — `getTransaction(sig,{commitment:'finalized'})`, assert finalized + `burnChecked`(+optional treasury transfer) on `WOC_MINT` by `payer` ≥ price + `memo == quoteId`. Raw RPC, mirrors `woc_balance.ts`.
[...]

PR #473, opened 2026-06-16
feat(woc): $WOC holder flair → shareable player card (poses, balance, leaderboard Top%, share-to-X, referral)

awidearray:
**Was a stub — now the implementation.** This converts the *Holder cosmetic flair* mechanic into a full, shippable feature: the `$WOC` holder-tier flair lives on a **shareable player card**. Stacks on #465 (the card reads the connected wallet's balance), so the diff narrows to just this work once #465 merges.

## What's on the card
- **Character render** with a **3-pose picker** — Hero / Battle / Victory (frozen animation poses, verified across classes).
- **Stats + equipped gear** (quality-coloured), arena rating, prestige.
- **Live on-chain `$WOC` balance + holder-tier badge** — 10 cosmetic rungs (Ember → Sovereign), no pay-to-win. Balance is read from Solana (`getParsedTokenAccountsByOwner` for the mint).
- **"TOP N%" leaderboard chip** — realm percentile by lifetime XP (same metric as the in-game leaderboard); shown for the top half of realms with ≥5 players.
- **Full brand lockup** in the footer + `worldofclaudecraft.com`.

## Sharing + referral
- **Share to X** (copies the card image to clipboard to paste, + the link), **Copy Referral Link**, **Download**, native **Share** (mobile). Online-only actions are gated; offline gets Download.
- Server: `POST /api/card` stores the composited PNG in Postgres (shared across realms); `GET /p/<slug>` is a prerendered **Open-Graph** page that unfurls the card on X/Discord; `GET /p/<slug>/card.png` serves it; `?ref=<slug>` records the referral on register. `GET /api/characters/:id/standing` powers the Top%.

## Verification
- tsc clean · **1282 tests** pass (holder-tier, card server handlers, lifetimeXpStanding) · client + server builds pass.
- Live e2e (`scripts/player_card_e2e.mjs`) 15/15 against real server + Postgres.
- Card + poses + logo verified on a real in-game card; standing endpoint verified live.

## Follow-ups
- Badge tier glyphs are placeholder vector art — reskin with final assets.
- Card modal/share chrome uses English literals (not i18n keys) — localisation later.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1496, opened 2026-07-05
feat(woc): $WOC renames and vanity names, re-cut on release/v0.22.0

awidearray:
## Summary

The maintainer-requested re-cut of #734: burn $WOC to rename characters and guilds, plus reserved vanity names at premium burn prices, rebuilt from scratch on `release/v0.22.0`. The superseded foundation (wallet linking, holder flair, player card) is DROPPED: this branch reuses the evolved #473 wallet identity and holder pipeline that release already ships. Only the feature and the minimal shared $WOC payment plumbing (quote, burn verification) are ported.

## What ships

- Server: `woc_config` (feature flag), `solana_tx` read helpers, `woc_payment` burn verification, `identity` + `identity_actions` (prices, quote, pay context, confirm), `renameCharacterVoluntary` (the paid path gates on `force_rename = FALSE`, distinct from the moderation rename), guild rename via the leader, name reservations with a creation-time guard (a hole in the original PR: reserved names were only blocked at rename, not at character create), and LDNOOBW name hardening.
- Client: Api identity methods, a pure unit-tested legacy transaction serializer (`burnChecked` plus optional `transferChecked` plus memo), `signAndSendWocTransaction` through the Wallet Standard `SolanaSignAndSendTransaction` feature (release dropped Reown and web3.js; the RPC key stays server side via the pay-context endpoint), the quote to pay to confirm flow, and a self-contained paid rename editor module wired thinly from main.ts.
- i18n: every player string is a t() key; the fifteen payment and error strings carry real translations in the five non-Latin locales; the guild renamed broadcast is matched in server_i18n with entries in all twenty locale DICTs.

Paid rename editor
The paid rename editor on a character row (captured by the checked-in `scripts/paid_rename_shot.mjs` through the real module; the price shown is the configurable default).

Cosmetic-only burn sink: a voluntary rename burns $WOC outright, vanity reservations burn at premium prices, and the shared payment layer (quote plus on-chain burn verification) is the base the SNS and marketplace re-cuts stack on. No emission anywhere, so the flow ledger invariant is untouched.

Fail-closed: `WOC_IDENTITY_ENABLED` defaults off; every identity endpoint 404s while off; the client renders the paid surfaces only after `/api/identity/prices` succeeds. The free moderation rename path is unchanged when the flag is off.

## Verification

- `tsc --noEmit` clean; the full suite green on this branch: 7935 tests passing, zero failures, including the malware release gate (the client builds transactions with a hand-rolled, unit-tested serializer and the Wallet Standard sign-and-send feature, so no drain-toolkit dependency exists to flag).
- Ported tests are real tests of the ported code: woc_payment, identity_actions, name_filter, social_system additions, plus the new woc_tx serializer and woc_identity flow suites.

## Reviewer checklist

- [ ] Based on release/v0.22.0 with zero conflicts (fresh cut, not a merge chain).
- [ ] Fail-closed defaults: WOC_IDENTITY_ENABLED off, endpoints 404, client surfaces hidden.
- [ ] The paid rename reuses release's market and mail rekey exactly like the free path.
- [ ] Burn verification: quote-bound memo, replay-safe payment ledger, no client-trusted amounts.
- [ ] Reserved names enforced at BOTH rename and character creation.
- [ ] Sim untouched; determinism and purity guards green.


FernandoX7 (commented):
This is a strong re-cut and the payment core is genuinely well built. The hand-rolled serializer decodes correctly byte for byte (burnChecked/transferChecked layouts, canonical program ids, no extra instructions, no approvals or delegate paths), server verification is balance-delta based so a crafted client tx can only fail, the replay guard is a DB-level UNIQUE, every route is bearer-scoped with ownership double-gated, the SQL is parameterized, the DDL is additive and idempotent, and the fail-closed flag behaves exactly as described. We re-ran the malware gate independently over the diff, decoded the tx bytes by hand, and verified the naughty-words package against the registry: all clean. The reservation-at-create guard closes a real hole from #734, and the payment/flow test suites have unusually good both-arm coverage (103 tests, all passing here).

The blocking findings share one theme: paths where the burn lands on-chain and the action then fails, which for a burn means the player's money is simply gone.

Blocking

1. server/identity_actions.ts:45 - prepare() for rename_character never checks the new name against existing characters. characters.name is UNIQUE (server/db.ts:71) and the free path pre-checks availability (server/db.ts:2021), but the paid flow discovers the collision only at apply time, after the burn is verified and recorded. Pick any taken name, burn 500 $WOC, get a 409. Add the availability check to prepare; the residual race then only matters for two players quoting the same name at once.

2. server/db.ts:623 - the partial unique index name_reservations_active is WHERE status = 'active', and nothing ever flips status to 'released' or deletes expired rows. Once a reservation's 90-day TTL lapses, activeReservationHolder correctly ignores it, so prepare passes, the next buyer pays 1000 $WOC, and reserveName's INSERT still hits the index: a guaranteed burn-then-409, and the name can never be re-reserved by anyone. Expired rows need to be released (delete or status-flip) before insert, or pruned the way woc_quotes are.

3. server/identity.ts:222 - recordWocPayment and actions.apply are two separate auto-commit writes, and deleteWocQuote runs regardless of the apply result. Any apply-time failure the code itself anticipates (character logged in during the on-chain round trip, force_rename flagged mid-flight, a lost race) leaves the tx_sig consumed and the quote gone: burned funds, no action, no retry. Wrap record plus apply in one transaction so an apply failure rolls back the payment row, and keep the quote when the action did not apply.

4. Rebase: this branch was cut before the API pipeline (#1491) merged into release/v0.22.0, which is why server/main.ts and src/main.ts conflict. The dangerous part is not the four new /api/identity/* routes (the dispatcher delegates unmatched paths to the legacy ladder, so they would still be served), it is the edits to the character create and free-rename handlers: POST /api/characters and POST /api/characters/:id/rename are RouteDef modules on the current base (server/characters.ts) with dispatch defaulting to 'new' (server/http/config.ts:76), so a mechanical conflict resolution leaves the reservation-at-create guard, the free-rename reservation guard, and the 402 paid signal on the dormant legacy arm, silently reopening the exact create-time hole this PR closes. Re-apply those guards in the RouteDef handlers and land the identity routes as a RouteDef module too (npm run new:endpoint scaffolds one, see server/http/CLAUDE.md). The src/main.ts side is mechanical. This is unfortunate timing, not a flaw in the PR as written, but it has to happen before merge.

Should-fix

5. server/woc_payment.ts:65 - when WOC_BURN_BPS is lowered but WOC_TREASURY is unset, the treasury-credit check is silently skipped, so a hand-built tx passes verification by burning only the burn portion and sending the remainder anywhere. The paycontext handler already 503s on this misconfiguration; verification should fail closed the same way (treasuryBase > 0 with no treasury configured means reject).

6. server/db.ts:1448 - getWocQuote filters expires_at > now(), and the wallet-approval wait is unbounded: a player who leaves the wallet prompt open past the 15-minute TTL burns after expiry, and confirm then refuses the quote. Since the memo already binds the payment to the quote and the price is frozen in the quote row, confirm could tolerate a generous grace window past expiry; the TTL's real job is stale prices at quote time, not redemption.

7. server/main.ts:1480 - the identity routes have no rate limit. paycontext and confirm each fan out to the metered Solana RPC (getLatestBlockhash, getTokenAccountsByOwner, getTransaction) and quote writes a DB row per call. The sibling /api/woc/balance route is IP-rate-limited precisely so it "can't be used to hammer the RPC" (server/main.ts:1573); these deserve the same treatment.

8. Test pins: tests/woc_tx.test.ts:118 and :153 assert the account table against the imported SPL_TOKEN_PROGRAM / MEMO_PROGRAM constants, and the woc_payment fixture defaults to the imported id, so a typo'd program address moves both sides together and still passes. Pin each of the three program ids to its base58 literal once (the instruction tags, offsets, and the mint are already correctly pinned). While there, the biggest untested deny arms are worth adding: WOC_IDENTITY_ENABLED off returns 404, a replayed tx_sig is rejected, a wrong-mint spend is rejected, and the treasury_short path (nothing currently exercises verification with a treasury split configured).

Nits

9. server/woc_config.ts:81 - numEnv('') and clampInt('') return 0, not the default, so a set-but-empty env line (WOC_BURN_BPS=) silently flips burn-everything into burn-nothing. Everything downstream still fails closed on the price side, but this repo has been bitten by exactly this shape before; treat empty as unset.

10. src/main.ts:3943 - refreshCharacters awaits ensureIdentityPrices() before the first character-list paint, so a slow /api/identity/prices delays the whole list. The paid button already renders conditionally on the cached price, so this can be fire-and-forget.

11. src/ui/paid_rename.ts:77 - the onRenamed() await sits inside the same try as the payment: if it ever rejected, the catch would overwrite the success status and re-enable the Pay button after a completed burn. Latent today (refreshCharacters swallows its errors), cheap to harden.

12. server/db.ts:1461 - pruneWocQuotes is a full-table delete with no index on expires_at. Fine at this volume, cheap to add if quotes grow.

13. src/styles/shell.css:2455 - .char-row { flex-wrap: wrap } touches the same rows #1477 (charselect reflow) just adjusted on the base; worth one visual pass over the character list after rebasing.

The remaining i18n conflicts (pending.ts, i18n.resolved.sha256, i18n.status.summary.json) are the usual generated-file churn: merge the base and re-run the i18n build. The localization wiring itself checks out (the guild-renamed broadcast matches through server_i18n and the S3 guard is green).

Findings 1 to 3 are one family: with a burn there is no refund path, so every failure after the money leaves the wallet has to be either prevented at prepare time or made retryable. Happy to take another pass after the rebase.

PR #736, opened 2026-06-19
feat(woc): native character marketplace + buyback-burn [plan/WIP]

awidearray:
# PR 3 — Native Character Marketplace + buyback-burn

> **Status:** DRAFT / implementation plan. Stacked on **PR 2** (`feat/woc-sns-tradeable`) → PR 1 → #473. Flag-gated; no behavior change on `main` by default.

## Summary
List a bound character for sale in **USDC** (`EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`) at a seller-chosen price. On sale: **70% → seller, 30% → buy `$WOC` on a DEX and burn it.** Every secondary-market trade becomes deflationary pressure on `$WOC`, capturing gray-market RMT value on-platform. Shares the buyback-burn primitive with stubs #469 / #466.

## A. Listings — off-chain orderbook + trustless on-chain escrow
Long-lived passive listings, OpenSea / Magic-Eden style. A Solana tx is pinned to a recent blockhash that expires in ~60–90s, so a naive pre-signed sale is impossible; the listing authorization is **the asset sitting in escrow**, not a pre-signed transaction. Seller is offline at sale time; only the buyer signs at fulfillment.

- **List** `POST /api/market/list` `{ characterId, priceUsdc, expiresAt }` — current owner only; character must be offline, unlisted, and not guild-leader-with-members (or auto-handled on settle). Transfers the character's subdomain into the **marketplace escrow program PDA** (CPI to SNS `transfer`); records `seller, price, expiry, split` on-chain and mirrors to a `market_listings` DB table for discovery/search. **Cancel** anytime before expiry to reclaim. Expiry seller-chosen, hours → months (configurable cap).
- **Buy** `POST /api/market/buy` `{ listingId }` — buyer submits one tx; the program settles **atomically**: subdomain escrow → buyer, **70% USDC → seller, 30% USDC → buyback vault**; emits an event. PR 2's claim/transfer cleanup then runs (reassign `account_id`, clear stale state). Gear/progress follow the character.
- **Escrow program** (`programs/character-market/`, Anchor): instructions `list`, `cancel`, `buy`; holds the subdomain in a PDA; verifies price/expiry; enforces the 70/30 split; USDC mint pinned. Needs build + **security audit** + devnet→mainnet deploy + upgrade-authority custody. Interacts with PR 2 ownership-resolution: a subdomain owned by the escrow PDA reads as "listed/locked," controlled by the lister until sold/cancelled.

[...]

PR #735, opened 2026-06-19
feat(woc): SNS subdomains + tradeable characters [plan/WIP]

awidearray:
# PR 2 — SNS Subdomains + Tradeable Characters

> **Status:** DRAFT / implementation plan. Stacked on **PR 1** (`feat/woc-rename-vanity-names`) → #473. Flag-gated (`SNS_ENABLED`, `CHARACTER_TRADEABLE` default **off**); no behavior change on `main` by default.

## Summary
Mint **player-owned** SNS subdomains `‹name›.worldofclaudecraft.sol` **atomically with a `$WOC` burn**, and **bind character ownership to the subdomain**: transfer or sell the name on-chain and the character — *with all gear and progression* — goes with it. This is real-money-tradeable characters, native on-chain (the WoW/RuneScape account economy done legitimately).

> ⚠️ **Deliberate product decision:** this makes characters pay-to-win / RMT and raises reg risk **Low → High**, overriding the stub's "never power" constraint. That's why it's its own PR, behind flags, gated on legal review before mainnet enablement.

## A. Subdomain minting (atomic with the burn)
- **Dep:** `@bonfida/spl-name-service`. **Config:** `SNS_ENABLED` (off), `SNS_PARENT_DOMAIN=worldofclaudecraft.sol`, `EXECUTION_WALLET_SECRET` (encrypted at rest, reusing the `SecretVault` pattern from `docs/prd/eliza-agents.md`), `WOC_PRICE_SUBDOMAIN`.
- **`server/sns.ts`** (new, server-only; boundary note like `solana.ts`): `buildIssueSubdomainIxs(label, ownerPubkey)` (owner = player wallet, fee-payer = player), `partialSignWithExecutionWallet(tx)` (parent-domain authority), `subdomainAvailable(label)`, `resolveSubdomainOwner(domain)` (read-only cached RPC, mirrors `woc_balance.ts`).
- **Atomic tx:** extend PR 1's `identity.ts` quote so that when `claimSubdomain`, the subdomain-create instructions are appended to the burn tx and execution-wallet partial-signed → the player's **single signature burns `$WOC` AND mints `‹label›.worldofclaudecraft.sol` to them, or neither happens.** Player pays ~**0.00156 SOL** rent (~$0.11) + fee; **project spends 0 SOL**. Confirm independently verifies the registry-create for the expected label/owner alongside the burn.
  - Rent measured on mainnet: 96-byte subdomain registry = 0.001559 SOL rent-exempt.
[...]

PR #1497, opened 2026-07-05
feat(program): character market escrow program with devnet proofs (split 1 of 4 from #736)

awidearray:
## Summary

PR 1 of the four-way split of #736 (the author-offered split of the oversized character marketplace PR): ONLY the on-chain character-market escrow program and its devnet proofs, cut fresh from `release/v0.22.0`. No game code: no server routes, no client UI, no vitest files. Those follow as PR 2 (server orderbook plus buyback keeper reuse), PR 3 (client UI), and PR 4 (character bind and identity surface), stacked on the #735 re-cut.

## What ships

- `programs/character-market/`: the Anchor escrow program (list, buy with the 70/30 seller/buyback split, cancel), its Cargo manifests, and its own devnet exercise (`tests/devnet.mjs`, results, Solscan capture script and shots).
- `Anchor.toml` (release has none; the file carries only this program's entries).
- `docs/prd/woc/character-marketplace.md`, updated to state this PR carries the program only.

## Devnet proof

Program id `BE55pNoRLSCch5NmLcU6tg6NZFLe6yFw4Jnr3HfZBMpp`, consistent across `declare_id!`, Anchor.toml, and the devnet driver. Recorded signatures (deploy, list, buy, list for cancel, cancel) are in `programs/character-market/tests/devnet-results.json` with Solscan captures alongside. Honest scope: the signatures were verified for internal consistency in this pass (base58, id match, URL match), not re-queried against devnet RPC, and the Rust build was not re-run here (no Anchor toolchain assumed); the committed artifacts are the proof trail.

This is the settlement rail for the highest-value burn stream (whole-character sales, 30% of each sale routed to buyback and burn), but this PR alone moves nothing: the program is inert without the server rails, which arrive in PR 2 behind fail-closed flags and carry the regulatory review flag the author committed to.

## Reviewer checklist

- [ ] Program authority model: escrow PDA custody, seller-only cancel, buyer-pays-exact list price.
- [ ] The 70/30 split math on buy.
- [ ] Anchor.toml ids match declare_id and the devnet artifacts.
- [ ] Nothing outside programs/, Anchor.toml, and the PRD is touched.
- [ ] PR 2 (server) is where money-flow review and the regulatory flag land; this PR is reviewable standalone.


FernandoX7 (commented):
Clean split. Scope is exactly what the description promises: only `Anchor.toml`, the PRD, and `programs/character-market/**`, no game code, and it merges into `release/v0.22.0` with no conflicts (the base carries no `Anchor.toml` or `programs/` dir yet). A few things I checked and can confirm hold up:

- The devnet proofs are real, and the honest "verified for internal consistency, not re-queried against RPC" caveat undersells them. I re-queried devnet: all five signatures (deploy, list, buy, list-for-cancel, cancel) confirm with `err: null`, the `buy` tx ran against `BE55...BMpp`, and its token deltas are exactly buyer -100 / seller +70 / vault +30 in the mock mint. The 70/30 path does what it says on-chain.
- The split math is dust-exact (`buyback = price - floor(0.7 * price)`, `lib.rs:65-66`), single-settlement via `close = seller` is right, and the release profile keeps `overflow-checks` on.
- I audited both lockfiles since this is an on-chain program. Both are clean. (The one entry that looks alarming, `zmij` replacing `ryu` inside `serde_json`, is legitimate: it is a dtolnay crate and the checksum matches crates.io. Not a finding.)

The main thing to work through is that, as written, the program is a USDC splitter with a listing registry, not the escrow the PRD and the doc comments describe.

### blocking

**The program never takes custody of or moves the subdomain, and nothing ties the sale to asset delivery.** `programs/character-market/src/lib.rs:5-9` (and the per-instruction comments) say custody rides "as SNS Name-Service instructions composed into the same transaction" and that a sale "moves the name and the money together or not at all." But there is no `spl-name-service` dependency (`Cargo.toml`), no CPI other than the two token transfers (`lib.rs:70`, `lib.rs:82`), no `invoke_signed`, and `buy`/`cancel` do not even receive the `subdomain` account (`Buy`, `lib.rs:145-165`). So `buy` splits the USDC and closes the listing with zero on-chain linkage to any name transfer: a buyer pays and is guaranteed nothing. Worse, if a name were ever parked in the `["listing", subdomain]` PDA, only this program via `invoke_signed` over those seeds could move it back out, and it has no such CPI, so the name would be frozen permanently. Either way "holds the subdomain in a PDA" and the trustless-atomic claim do not hold. The custody wiring may be planned for a later PR, but the checklist here asks to verify "escrow PDA custody," and today the program cannot provide it. Either implement the SNS custody in-program (transfer-in on list, `invoke_signed` transfer-out on buy/cancel) or correct the docs to say this is a settlement splitter that trusts the server to move the name.

### should-fix

- **`buyback_vault_usdc` is seller-supplied and unpinned, so the 70/30 split is defeatable** (`lib.rs:126-127`). It is only constrained to share the mint; `buy` re-checks only that it equals what the seller recorded (`WrongVaultAccount`). A seller points it at their own account and keeps 100%, nullifying the buyback tokenomics. Pin it to the canonical vault (a program PDA or a hardcoded `address`).
- **`usdc_mint` is unpinned** (`lib.rs:123`, `UncheckedAccount`). `token::mint = usdc_mint` only enforces internal consistency across the three token accounts, so a listing can be denominated in any mint despite the "USDC pinned" claim. Add `address = <canonical USDC mint>`.
- **`list` never checks the seller controls the `subdomain`** (`lib.rs:113`), and the PDA is keyed on that arbitrary pubkey (`lib.rs:118`). Anyone can `init` a listing for a subdomain they do not own, which blocks the real owner from listing (the PDA already exists) and, combined with the blocking item, lets a griefer collect a buyer's payment for an asset that was never escrowed. Gate listing on verified ownership of the name.

### nits

- `NotSeller` is dead code. `Cancel` has `has_one = seller` (`lib.rs:140`), which rejects a non-seller before the handler body, so the `require_keys_eq!(..., NotSeller)` at `lib.rs:50` never runs. Your own devnet driver asserts `ConstraintHasOne` for that path, not `NotSeller`. Drop one of the two.
- `price.checked_mul(SELLER_BPS).unwrap()` (`lib.rs:65`) panics for prices above ~2.6e15 base units, and `list` only bounds `price > 0`. No funds are at risk, but such a listing becomes permanently unbuyable. Prefer a bounded price or a returned `MarketError`.
- Expired listings are reclaimable only by the seller (`cancel`, `has_one = seller`), and the design says the seller is offline. There is no permissionless reap, so an abandoned expired listing locks that subdomain's listing slot indefinitely.
- Self-buy: with `buyer == seller`, `buyer_usdc` can alias `seller_usdc` (an SPL transfer to itself is a successful no-op), so a user can wash-trade and still emit a genuine-looking `Sold` event (`lib.rs:93`). Anything off-chain that trusts `Sold` for volume or price can be spoofed for free.
- `tests/devnet.mjs` loads the deployer keypair from `../../../.secrets/deployer.json` (repo root), which is not gitignored: the root `.gitignore` covers `.env*` only, and the program `.gitignore` covers `/target` and `/tests/node_modules`. Add `.secrets/` so a funded devnet key cannot be committed by accident. (Also, `Anchor.toml`'s `wallet = "/tmp/woc-deployer.json"` and the driver's key path disagree.)
- The two new test files are not biome-formatted. `biome.json` includes `**` with no exclusion for `programs/` or `.mjs`, so `biome ci --changed` (the `ci:changed` step and the pre-push floor) will fail on format diffs until you run `npx @biomejs/biome check --write` on `tests/devnet.mjs` and `tests/shot_solscan.mjs`. I confirmed both produce format diffs (long imports and array/object literals get broken to one item per line at lineWidth 100).
- Doc-comment glyphs: `lib.rs` comments still use `->` arrows and `70 %` (and `<label>` guillemets). Not the banned em/en-dash-or-emoji set, but inconsistent with the plain-ASCII house style and the "dash cleanup" the PR mentions.

PR #1499, opened 2026-07-05
feat(woc): SNS subdomains and tradeable characters, re-cut on the #1496 rename base

awidearray:
## Summary

The re-cut of #735: yourname.worldofclaudecraft.sol SNS subdomains plus tradeable characters, rebuilt on the #1496 rename re-cut (which is on release/v0.22.0). The superseded foundation is dropped; the #1496 payment layer (quote, burn verification, Wallet Standard sign and send) is reused, never duplicated. STACKED on #1496: review that first; this diff over it is the four commits listed below.

## What ships

- SNS subdomain minting, rebuilt malware-gate friendly: pure SNS key derivation (sha256 plus off-curve PDA check via the existing noble dependency, validated against the known bonfida.sol vector), a hand-rolled two-signer legacy transaction serializer with the execution wallet's partial signature, and RPC reads through the existing solana_tx helpers. No @solana/web3.js, no Bonfida SDK dependency.
- Server rails fail-closed: WOC_SNS_ENABLED defaults off, every /api/subdomain route 404s while off, 503 while on but unconfigured.
- Tradeable characters: the claim endpoint with the on-chain controller check, the world-entry transfer gate, and the bound-domain schema (characters.bound_domain, sns_subdomains, reassignCharacterAccount).
- Client: Api methods, an injected-IO mint flow module, a mint editor mirroring the paid rename editor, thin main.ts wiring that renders only after the prices probe succeeds, all strings localized with real five-non-Latin fills.
- Real tests: SNS core (22: derivation vector, registry parse, byte-level tx decode, signature verify), claim control (4), mint client flow (7), plus the parent's payment suites.

A new asset class: characters with hours invested become $WOC-denominated value, and subdomain mints burn $WOC through the shared payment layer. No emissions, so the flow ledger invariant is untouched. Everything is fail-closed until keys and sign off.

## Verification

tsc clean; 99 tests passing across the seven relevant files; i18n regen idempotent; malware scan delta vs the parent: zero new high findings (one new medium: the canonical Solana system program id tripping the hardcoded-address heuristic, a false positive noted here rather than silenced).

## Reviewer checklist

- [ ] Stacked on #1496; review that first; this PR's own diff is four commits.
- [ ] WOC_SNS_ENABLED fail-closed at every route; client renders only after the prices probe.
- [ ] Subdomain mint transaction: two-signer build, exec wallet partial sign server side, player signs and sends via Wallet Standard; RPC key never reaches the client.
- [ ] Character claim: on-chain controller check before reassignment; world-entry gate blocks transferred-away characters.
- [ ] The derivation vector test pins bonfida.sol to its known key.


FernandoX7 (commented):
Reviewed the four commits stacked on #1496 (the payment layer gets its own review; nothing here re-litigates it). The crypto half of this is genuinely well built. The hand-rolled tx keeps the execution wallet as a readonly non-fee-payer signer, so its signature authorizes exactly one create and cannot move its funds; the player pays fee and rent. Key derivation handles the subdomain "\0" prefix correctly (the classic bug to make here), and the bonfida.sol vector pins the whole chain, program id and root included, to a real mainnet value. The byte-level decode test pins instruction count, the full account table, amounts, and trailing length, so a smuggled instruction fails the suite. I verified the hardcoded addresses against public sources, confirmed the four-commit delta adds zero dependencies, and re-ran the malware scan: the one new MEDIUM is the system program id, a false positive as described. 33/33 new tests pass, tsc is clean, fail-closed gating checks out route by route, and the client side mirrors the paid-rename pattern faithfully (play.html parity included).

The findings are concentrated in the game-integrity half, mostly the claim path.

Blocking

- server/character_claim.ts:86: claiming a Guild Master hands the buyer the seller's guild. `leaveGuildOnTransfer` calls `SocialService.guildLeave`, which for a leader with remaining members sends an error to the just-kicked character and returns without leaving and without throwing (server/social.ts:617). The claim proceeds and the buyer now controls the guild: promote, kick, disband. The comment says leader hand-off is "handled by the normal guild-leave path", but that path refuses leaders instead. Transfer needs an explicit rule (auto-promote, disband, or refuse to claim a GM character until leadership moves), and the hook should fail loudly if the character is still in a guild afterwards.

Should-fix

- server/main.ts:1148: a bound character can be deleted during the claim window. The DELETE route has no bound-domain guard, so a seller can sell the subdomain, then delete the character before the buyer claims, leaving the buyer a dead name they paid for. Block deletion while `bound_domain` is set, or require current on-chain control to delete.
- server/main.ts:290: the claim kicks the prior session fire-and-forget (`disconnectAccount` does not await `kickSession`) and reassigns immediately. `takeOverCharacter` (server/game.ts:2079) awaits the kick precisely to avoid racing the old save and the "character already in world" rejection; the claim path should too.
- server/db.ts:1951: `recordSubdomainAndBind` can commit a half-result. The characters UPDATE is `WHERE id AND account_id` with no rowcount check, so if the character was deleted or taken over between quote and confirm, the transaction still commits the `sns_subdomains` row with no bind; two concurrent confirms for one character (two separately paid quotes) also both commit. Add `AND bound_domain IS NULL` plus a rowcount check that aborts, and consider a partial unique index on `sns_subdomains(character_id)`.
- server/subdomain_mint.ts:220: if `recordSubdomainAndBind` throws after `recordWocPayment` succeeded, the replay guard has consumed the tx_sig; the retry gets "payment already used" and the player has burned $WOC for a mint the server never recorded. Bind before recording the payment, or make confirm idempotent when the payment row exists for this quote and the bind is missing.
- server/db.ts:1967: `reassignCharacterAccount` does not rekey `player_cards.account_id` (denormalized copy, server/db.ts:556). After a claim the new owner's card list omits the character, the old owner still lists it, and deleting the old account CASCADE-drops the surviving character's card. Rekey or reset the card in the same transaction.
- server/subdomain_mint.ts:85: the minted label comes from client-supplied `name` and is never compared to the character's actual name (the editor input is editable too). Any account can mint the subdomain form of any name, including another player's character name, squatting the name that player could otherwise mint. The PRD frames this as minting your character's name; enforcing `slugifyLabel(character.name) === label` server-side matches that and closes the squatting angle. If arbitrary labels are intended, say so in the PRD.
- server/main.ts:1533: the three new endpoints have no rate limit, and one quote fans out four to five upstream Solana RPC calls plus DB writes, so an authenticated account can hammer the paid RPC. The identity routes share this gap, but this is the heavier surface; putting quote/confirm/claim behind the existing limiter is cheap.
- tests/sns.test.ts:225: the account-table and program-index assertions compare `SPL_TOKEN_PROGRAM`, `MEMO_PROGRAM`, and `SYSTEM_PROGRAM` against the same constants imported from server/sns.ts, so a swapped id moves both sides and stays green. The name program and root are transitively pinned by the bonfida vector; these three need literal pins, three lines total.
- No tests for the server handlers: `handleCharacterClaim` (including the Guild Master case; it also re-implements the ownership check inline rather than calling the tested `accountControlsBoundCharacter`, so the route's real 403 path is unexercised), the quote/confirm gating, reason-string, and replay paths (the client flow test pins reactions to those strings, not that the server emits them), and the two DB transactions. The pure core and client flow are well tested; the orchestration layer where the findings above live is not.
- src/ui/mint_subdomain.ts:46: the mint input's placeholder and aria-label reuse `character.newNamePlaceholder` ("New character name"), but this field picks a .sol name, not a rename, so a screen reader announces the wrong thing. The PR already adds mint-specific keys; one more English label key fixes it.

Nits

- server/solana_tx.ts:98 + server/subdomain_mint.ts:44: the blockhash is fetched at finalized commitment (about 13s behind tip), leaving roughly 45s of real signing window, while the quote advertises a 15-minute expiry. A user who pauses at the wallet gets an unmapped "blockhash not found" and must manually re-quote. Fetching the blockhash at confirmed commitment and catching the expiry to auto re-quote once would cover most of it.
- server/main.ts:2067: the world-entry gate is fail-closed on RPC failure (the right direction), so an RPC outage locks every bound character out with no grace. A short-TTL last-known-owner cache would blunt that without weakening the check.
- server/db.ts:1980: the `sns_subdomains` UPDATE in `reassignCharacterAccount` has no realm filter, and `character_id`, its only predicate, has no index; meanwhile the added `(realm, bound_domain)` index serves no current query.
- tests/sns.test.ts:73: the subdomain "\0" path is only self-consistently tested; the bonfida vector validates everything except the leaf derivation. A known mainnet `x.<domain>.sol` vector would pin it end to end (if it were subtly wrong, tests stay green and every mint fails on-chain).
- src/net/woc_subdomain.ts:42: the client forwards the server-built tx bytes unchanged, so the only amount it asserts is the server's own `priceWoc`; the design also assumes the wallet preserves the slot-1 partial signature and does not replace the blockhash (standard Wallet Standard behavior). Both are fine, but worth a comment stating the trust boundary, or decode the tx and assert the burn matches the quote.
- server/subdomain_mint.ts:111: "this wallet holds no $WOC token account" is reachable for any linked wallet holding zero $WOC and falls through the client matcher to raw English; the identity flow shares the gap, but the matcher block is right there.
- src/main.ts:4023: the mint button renders on already-bound characters (the client character shape has no bound flag), so the user learns only at the confirm 409. A bound flag in the list would let the client hide it.

On merge state: the true conflicts against release/v0.22.0 are server/main.ts, src/main.ts, the shell.ts catalog, and the generated i18n trio, i.e. standard stale-base regen churn; rebase once #1496 lands and re-run the i18n build. Given the PRD's own framing (RMT posture, legal review before enablement), the flags-off default and the 404-while-off behavior are the right shape for merging ahead of that call.

PR #1502, opened 2026-07-05
fix(world-boss): rebalance Thunzharr size, snare, per-player HP, and slow immunity

ryan-foo:
## Summary

Rebalances the Thornpeak world boss, Thunzharr, which had drifted too far in the "unkitable / mountain-sized" pass:

- **Scale 50 to 8.** Still an imposing world boss on the skyline, but no longer dwarfs the zone. His melee reach is unchanged (pinned to a scale-5 body in `combatProfileForMob`), so he stays unkitable.
- **Howling Gale snare 80% to 30%** (`aoeSlow.mult` 0.2 to 0.7). Still denies a permanent kite (a 7yd/s hunter is dragged below the boss's 5.8 move speed) without rooting the raid in place.
- **Per-player HP step 40k to 5k.** This is the fix for "he takes way less damage": `scaleWorldBossHp` adds each new participant's HP delta to the boss's CURRENT health (real health, not a heal), so at 40k/head the bar visibly refilled as a raid trickled in and read as damage not landing. 5k/head keeps the pool scaling with raid size without stalling the fight.
- **Slow immunity.** Thunzharr now shrugs off player movement slows (new `slowImmune` template flag). `ccImmune` blocks the hard control auras (stun/root/etc) but deliberately leaves snares landing so ordinary elites can still be kited; a raid boss opts into both.

## Type of change

- [x] Bug fix
- [x] Feature: new functionality (`slowImmune` template flag)

## How was this tested?

- Commands:
  - `npx vitest run tests/world_boss.test.ts` (30 passed, incl. updated snare/scale/pool assertions + a new slow-immunity test)
  - `npx vitest run tests/architecture.test.ts` (sim purity guard)
  - `npx vitest run tests/sim.test.ts tests/fixes.test.ts` (171 passed, exercises the `applyAura` hot path)
  - `npx tsc --noEmit` (no errors in changed files)
- Manual steps: ran `scripts/thunzharr_size_shot.mjs` against the offline client to capture the new scale-8 boss against the old scale-50 (below).

## Screenshots / recordings

After (scale 8):

Thunzharr scale 8

Before (scale 50):

Thunzharr scale 50


FernandoX7 (commented):
Solid, well-scoped rebalance, and the commit message and comments carry the reasoning better than most. I checked the load-bearing claims against the code:

- Reach stays decoupled from visual scale. `combatProfileForMob` pins Thunzharr to `THUNZHARR_REACH_SCALE = 5` regardless of the `scale` field (`src/sim/mob_combat.ts:53-62`), so 50 to 8 does not touch melee reach. The anti-kite still holds: a 7yd/s hunter snared to 70% moves 4.9yd/s, under the boss's 5.8.
- `slowImmune` is genuinely distinct from `ccImmune`. `isControlAura` is stun/root/incapacitate/polymorph only (`src/sim/sim.ts:2925`), so a slow was never blocked before, and the new gate is the right place for it. The `aura.sourceId !== target.id` exemption keeps scripted self-slows landing, and only Thunzharr sets the flag (`zone3.ts:825`). Pure conditional, no rng.
- `scaleWorldBossHp` does add the delta to current hp (`world_boss.ts:148`), so the "bar refills as the raid trickles in and reads as no damage" rationale for the 40k to 5k step checks out.

`tests/world_boss.test.ts` (30) and `tests/architecture.test.ts` (24, the purity/determinism guard) both green at the head ref. The new slow-immunity test covering both the blocked player-snare and the landing self-slow is a nice touch.

One nit:

- `src/sim/mob_combat.ts:47-49`: the comment above `THUNZHARR_REACH_SCALE` still says "rendered mountain-sized (scale 50 in zone3.ts)" and "a scale-50 body would swing ~150yd". You refreshed the zone3.ts comment but not this one, so it now names the old value. The constant and logic are correct, just the prose is stale.

Non-blocking note: dropping perPlayer to 5k makes the 1M cap effectively unreachable at real raid sizes (a 40-player raid tops out around 235k now, versus the old 1M-capped ~1.6M), so the boss scales much more gently with crowd size. That is the intended direction (he was reading as damage-immune) and world-boss HP is a tuning knob, not a fixed formula, so it is your call. Just flagging so the new curve is a deliberate pick at the raid sizes you expect.

PR #1506, opened 2026-07-06
feat(net): linkdead grace keeps disconnected characters in-world for 5 minutes

maxpolaczuk:
## Problem

Any WebSocket drop today, however brief, logs the character out instantly and dumps the player on a fatal "you have disconnected" screen whose only action is a full page reload back to login. There is also no WS keepalive at all: an AFK player's client goes quiet (input frames are only sent on change), NAT/proxy idle timers silently cut the TCP stream, and the player experiences it as "kicked for AFK" or a random disconnect. Every one of these events dents our concurrent-player numbers and the session experience.

## What this does

**Server: linkdead grace.** A dropped socket no longer tears the session down. The session goes linkdead: the character stays in the sim (it can still be attacked, healed, or die, like classic linkdead behavior), held movement input is zeroed, and the character remains online for friends/guild presence, the open play-session analytics row, and the concurrent-player counts, for a 5-minute grace window. A safety character save fires at grace start. When the window expires, the normal leave() teardown runs.

**Seamless resume.** Re-authing on the same character rebinds the fresh socket to the held session: same sim pid, per-connection wire state reset (sentEnts, lastSent, selfHeavyDirty, lastWireRev, arena wire cadence) so the new socket starts from a full snapshot, input acking restarts at 0, and per-login account state (mute status, admin permissions, IP accounting) is refreshed from the fresh auth. No presence flap: friends never see a logout.

**Character switch cancels grace.** Logging in on a DIFFERENT character on the account displaces its linkdead sessions immediately (they log out through the normal leave path) instead of being blocked by the per-account session cap. A live-socket duplicate join is still rejected, so the explicit takeover flow remains the only way to displace a session whose connection is actually alive. A linkdead session owned by another account still rejects, and takeover/moderation/anti-bot kicks bypass grace entirely (leave() runs first; a later socket-close event no-ops via the session.left and stale-socket guards).

**Client: auto-reconnect.** ClientWorld now reconnects with exponential backoff (1s doubling to 15s, about nine minutes of attempts) behind a translucent "Connection lost. Reconnecting..." overlay instead of the fatal screen. A server error frame (kick, takeover, failed auth) still ends the session for good and shows the fatal overlay. On the resume hello the client resets its input sequencing, echo/latency sampling, snapshot interpolation anchors, and any spectate swap; stale mirrored entities fall out through the existing snapshot prune. Past the server grace, a successful re-auth is simply a fresh join from the last save, so retrying stays correct at any point.

**AFK root-cause fix.** The server pings every live socket at the WS protocol level every 30s (browsers reply automatically), keeping NAT/proxy idle timers from cutting quiet connections. No pong policing: a genuinely dead peer surfaces through the OS/ws stack and lands in the grace.

**Backpressure.** A terminated stuck-reader session (8 MiB unflushed) now lands in the grace instead of a full logout, since that is a network-quality problem, exactly what the grace exists for.

The join-conflict decision table lives in a new pure module, server/linkdead.ts (planJoin), unit-tested without a GameServer.

## Interactions considered

- Analytics: clients.size (status endpoint, online snapshots, peakOnline) intentionally counts linkdead sessions as online; the play_sessions row stays open across the grace.
- Playtime/daily rewards already gate on recent input, so a linkdead session cannot farm points.
- Snapshot building skips linkdead sessions (no transport); sends were already readyState-gated.
- Spectate exits at grace start on both sides; arena wire cadence resets on resume.
- Per-IP hard-cap accounting: a held session keeps its IP slot during grace (no cap evasion); a resume from a new IP rebalances the counts via a shared helper.
- Death while linkdead resumes into the current (possibly dead) state from the full snapshot.

## Testing

- New tests/linkdead.test.ts (18 tests): planJoin decision matrix, grace hold (online counts, presence, open play session), movement zeroing, resume re-sync (same pid, hello on the new socket, input/wire resets), stale pre-resume socket close ignored, kicked sessions cannot re-enter grace, expiry teardown, cross-character displacement, cross-account rejection plus takeover during grace, per-IP rebalancing on resume, snapshot skip while linkdead.
- tests/backpressure.test.ts updated for the grace semantics.
- Full suite: 729 files / 7869 tests passing; tsc clean; npm run build green; i18n scan/build/hash regenerated (new key loading.reconnecting with the five non-Latin fills per M16).
- Live E2E against a real server + real WebSockets (register, two characters): hard socket drop keeps the character online, same-pid resume receives snapshots, live duplicate join still rejected, second drop plus login on the other character displaces the linkdead one, and the first character can come back afterwards. All 14 checks passed.
- Reviewed by the qa-checklist, privacy-security-review (PASS, no blocking findings), and cross-platform-sync agents; the wire reviewer's two findings (stale ackedInputSeq and spectate playerClass across a mid-spectate reconnect) are fixed in the client reset block.


FernandoX7 (commented):
Careful piece of work. The join-conflict logic is isolated in a pure `planJoin` with a real decision-matrix test, `socketClosed` is idempotent per socket with the stale-pre-resume guard, and the resume path resets the wire/input/spectate state so the fresh socket gets a full snapshot. The interactions you listed hold up against the code: playtime gates on `lastInputAt` so a held session cannot farm, snapshot building skips linkdead, the IP rebalance goes through the shared helper, backpressure now lands in the grace, and every send/keepalive path is `readyState` gated so nothing throws or accumulates while a session is held. Good coverage.

One thing I would fix before merge.

**should-fix** `src/net/online.ts:1289` with `server/ws_auth.ts:53` and `server/linkdead.ts:44`: the auto-reconnect gives up for good the instant it hits a `character already in world` rejection, and that is exactly the rejection a re-auth gets when the server has not yet noticed the old socket died. A resume can only fire after the server has independently observed the drop (a clean close, backpressure, or a TCP/keepalive timeout) and flipped the session to linkdead. For a clean close (reload, tab reopen, clean network drop) that happens before the client's first retry and resume works well. For a black-holed drop where no FIN/RST reaches the server (mobile WiFi to cellular handoff, NAT rebind, the flaky-radio case in the description), the server keeps counting the old socket as live for many seconds, so the first retry auths, `planJoin` returns reject, `rejectHandshake` sends a `t:'error'` frame, and the client sets `sessionEnded` and drops to the fatal overlay with no further retries. The case the feature most wants to cover is the one where resume does not fire.

The client cannot tell this transient rejection from a permanent one (kick, moderation, failed auth), since they all arrive as the same `error` frame at `online.ts:1289`. Consider giving the reconnect-conflict rejection a distinct code so that while a reconnect is in flight (`reconnectAttempts > 0`) the client keeps backing off on that one instead of ending the session. Then the backoff resumes as soon as the server clears the stale socket, and the black-hole case is covered too. This is still a clear win over today (any drop is instantly fatal now), but as written the resume quietly does not extend to abrupt drops.

**nit** `server/game.ts:1752`: the resume path sends `X has entered World of ClaudeCraft` to the resuming client. On a seamless reconnect the player never saw themselves leave, so a self "entered" line in their own log reads a little oddly. Only they see it (no presence flap for others, which is the right call), so purely cosmetic.

The merge conflict is only the generated i18n trio (`i18n.resolved.sha256`, `i18n.status.summary.json`): standard stale-base regen, not a design issue. Merge the base and re-run the i18n build (`i18n:build` / `i18n:scan` / `i18n:hash --write`).

maxpolaczuk:
@FernandoX7 thanks for the sharp catch on the black-holed drop, you were right that the headline case was the one resume quietly missed. Fixed in 055997b, in two halves:

**Server: pong-based liveness.** The 30s keepalive sweep now tracks pongs (browsers answer automatically, and the pong handler is guarded on socket identity so a stale pre-resume pong cannot mask a black-holed replacement). A socket that misses a whole ping interval is terminated into the linkdead grace, so an abrupt drop is detected within one to two sweeps (~30-60s) instead of waiting minutes for TCP to give up or the snapshot backpressure to trip.

**Client: bounded conflict tolerance.** While a reconnect is in flight, a 'character already in world' rejection no longer ends the session; the client keeps backing off and the retry lands as a resume once the sweep flips the held session. I went with tolerating the existing wire string client-side rather than a new error code: the client already knows it is mid-reconnect (reconnectAttempts > 0), the string is already part of the wire contract for the char-select takeover prompt, and a new code would have needed matcher plus localization plumbing for what is client-local knowledge. The tolerance is bounded (8 consecutive conflicts, comfortably past two sweep windows at the backoff cadence) so a character genuinely held by another device's live socket still ends in the fatal overlay rather than retrying for nine minutes. The decision logic is a pure module (src/net/reconnect_policy.ts) with tests pinning the bound and the exact planJoin wire string.

**Nit:** the self "has entered World of ClaudeCraft" line on resume is gone.

Verified with unit tests for the sweep (ping, pong-clears, silent-socket-to-linkdead, resume clears the flag) and a live end-to-end run against a real server: paused the raw TCP socket (no FIN/RST), confirmed the early re-auth hits the transient conflict window, waited out the sweep, and the same-pid resume succeeded while the character stayed counted online throughout. Also rebased onto the current release/v0.22.0 and regenerated the i18n trio; full suite is green (9,853 tests) and the PR reports mergeable.

FernandoX7 (approved):

PR #1507, opened 2026-07-06
feat(ui): indicate self-only abilities in spell tooltips

### LINKED ISSUE #1357: Spell tooltips do not indicate a self-only ability
## Problem

Ability tooltips show whether a spell targets enemies or allies, but not whether it is self-only. Tooltips render "Friendly target" (castable on an ally) or "Enemy target", while self-only abilities (self-buffs such as Frost Armor, Ice Barrier, Evasion, the warrior stances, etc.) show no targeting line at all. A player cannot tell from the tooltip whether an ability can be cast on others or only on themselves.

## Expected behavior

- A self-only beneficial ability shows a "Self only" (or equivalent) line in its tooltip.
- The existing "Friendly target" / "Enemy target" lines are unchanged.

## Scope

- `src/ui/hud.ts` `abilityRequirementLines` (builds the friendly/enemy line; add a self line).
- The "self-only" signal is currently implicit (`requiresTarget: false` plus a buff/heal effect and no `targetType`), which collides with self-centered hostile AoE (Arcane Explosion, Frost Nova, Thunder Clap), so the tooltip must distinguish them.
- A new i18n key (e.g. `abilityUi.tooltip.selfTarget`) in `src/ui/i18n.catalog/abilities.ts` (English only; a wordy value also needs its non-Latin fills per the M16 rule).

## Acceptance criteria

- Self-buffs display "Self only"; hostile self-centered AoEs do NOT.
- Friendly and enemy target lines are unchanged.

## Notes

The cleaner option is an explicit `targetType: 'self'` marker on the self-buff abilities in `src/sim/content/classes.ts` (display-only; the sim already treats them as self-cast), which avoids the fragile effect-inspection heuristic.

jgyy:
## Summary
- Ability tooltips showed a "Friendly target" or "Enemy target" line but nothing for self-only beneficial abilities (Frost Armor, Ice Barrier, Evasion, the warrior stances, etc.), so a player could not tell such an ability was uncastable on others.
- Adds a `abilityRequirementLines` "Self only" line, gated by an explicit effect-type allowlist (`selfBuff`, `absorb`, `imbue`, `lifeTap`, `gainResource`, `selfDamagePctMax`, `summonDemon`, `dismissPet`) rather than `requiresTarget` alone, so hostile self-centered AoEs (Thunder Clap, Frost Nova, Arcane Explosion) are correctly excluded.
- New `abilityUi.tooltip.selfOnly` i18n key (English source), with the M16-required non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) added in the same change; resolved i18n tables + hash regenerated.

Fixes #1357

## Screenshot
Mage "Hoarfrost Mantle" (Frost Armor) tooltip now shows the new line:

Hoarfrost Mantle tooltip showing Self only line

## Test plan
- [x] `npm run gate` (full CI-equivalent gate: typecheck, i18n freshness, malware scan, changed-files biome, full vitest suite, build) green
- [x] `npm run build` clean
- [x] Manually verified in a live offline session: mage "Hoarfrost Mantle" tooltip shows "Increases armor by 30" / "Self only"
- [x] Verified by code audit that every hostile self-centered AoE ability (`aoeDamage`/`aoeAttackPower`/`aoeAttackSpeed`/`aoeRoot`/`groundAoE` effect types) is excluded from the self-only allowlist, so Thunder Clap/Frost Nova/Arcane Explosion etc. do not show the new line


ryan-foo:
Took a pass through this. Nice small feature, and the core reasoning is right.

You cannot infer self-only from `requiresTarget` alone (a self-centered hostile AoE like Frost Nova / Thunder Clap / Arcane Explosion also has `requiresTarget: false`), so keying off the effect types and excluding ground-target is the correct call. Those AoEs carry `aoeDamage`, which is not in the set, so they are correctly excluded. The buff model lines up too: friendly buffs (`buffTarget`, like Blessing of Might and Mark of the Wild) hit the existing friendly branch, and true self buffs (`selfBuff`, like Battle Shout, Arcane Intellect, Devotion Aura) land on the new label correctly. English key is in the right catalog, rendered via `t()`, and there is a screenshot.

Findings:

- **should-fix** `hud.ts:13912`: `def.effects.every(...)` is vacuously true for an ability with `effects: []`, so the three empty-effect abilities get a wrong "Self only" line: Conjure Water, Conjure Food, and Revive Pet. Revive Pet acts on the pet and the conjures create bag items, so the label is misleading on all three. Fix: `def.effects.length > 0 && def.effects.every(...)`.

- **should-fix**: no test for `isSelfOnlyAbility`. It is the actual feature logic, pure, and already exported, so a small truth table (self-buff yes, self-centered AoE no, ground-target no, friendly-target no, empty-effects no) is cheap and would have caught the empty-effects case above.

- **nit**: `isSelfOnlyAbility` is exported but only used in `hud.ts`. Keep the export if you add the test that imports it, otherwise drop it.

- **note, not a finding**: the CONFLICTING status is only the standard stale-base i18n trio (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`). Mechanical regen: merge `release/v0.23.0`, re-run the i18n build, commit.

The empty-effects mislabel is the one to fix before merge.

FernandoX7:
Re-reviewed and the core is solid: keying the label off the effect types (not `requiresTarget`) is the right call, the friendly/enemy/self branches line up, and the English key sits in the right catalog rendered via `t()`.

I went ahead and addressed the open findings on the branch so this is ready to merge:

- **Empty-effects mislabel (the should-fix):** `effects.every(...)` is vacuously true for an ability with `effects: []`, so Conjure Water, Conjure Food, and Revive Pet were getting a wrong "Self only" line (they make bag items / act on the pet, and are special-cased elsewhere). Guarded on a non-empty effects list.
- **Test + testability:** `isSelfOnlyAbility` lived in `hud.ts`, which pulls in the DOM/Three HUD and can't be imported from a Vitest, so I pulled the classifier (`isSelfOnlyAbility` + `SELF_ONLY_EFFECT_TYPES`) into a pure `src/ui/ability_self_only.ts` and left `hud.ts` the thin consumer (same shape as `ability_damage.ts`). New `tests/ability_self_only.test.ts` pins the truth table against real content: one true case per allowlisted effect type, the three empty-effects abilities, the self-centered AoEs (Frost Nova / Arcane Explosion / Thunder Clap), a ground-target cast, and the friendly-target buffs, plus a sweep asserting no empty-effects ability is ever self-only. That last one is the guard that would have caught the mislabel.
- **The nit** (export only used in `hud.ts`) resolves the same way: the export now has a real importer (the test), so it stays.
- **Merge:** brought in `release/v0.23.0` and regenerated the i18n resolved trio. The CONFLICTING status was only the standard stale-base generated files, nothing in the feature itself.

Local `npm run gate` is green (typecheck, i18n freshness, malware scan, changed-files biome, full vitest, all builds). Thanks for this, nice fix.

FernandoX7 (approved):

PR #1511, opened 2026-07-06
[codex] fix homepage server status and total players

Humpalumps:
## Summary

This PR cleans up the homepage server selector so it no longer presents the project stats value as a live player count. The selector now clearly communicates server availability with `Server Online` / `Server Offline`, and the stat line displays the dynamic `accounts_created` value from `/api/project-stats` as `Total players`.

The visual treatment was also tightened so the selector reads cleanly at both desktop and phone widths: the trigger is slightly larger, the server label/stat text is more legible, and the number plus `Total players` label are vertically centered on the same row.

Root cause: the previous homepage selector showed the account total next to generic `Players` / online wording, which made the value look like current online users. This change removes the online-player presentation from the selector and keeps the remaining stat tied to the project stats account total.

## Related issues

N/A

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run i18n:build`
  - `npx vitest run tests/homepage_foundation.test.ts`
  - `npx tsc --noEmit`
  - `git diff --check`
  - `git diff --cached --check`
  - `rg -n "js-stat-online|Players Online|players online|Players</span>" index.html src/main.ts src/styles/shell.css`
- Manual steps:
  - Verified the local homepage at `http://127.0.0.1:5180/` on desktop and mobile-sized viewports.
  - Confirmed the selector renders `Server Online`, a formatted dynamic project-stat number, and `Total players`.
  - Confirmed the number and `Total players` label are vertically centered by checking computed layout in a mobile viewport.
  - Confirmed the homepage selector no longer includes the removed players-online row/text.
  - Confirmed the screenshots below were refreshed after the final `Total players` label change.

## Screenshots / recordings

Desktop:

Desktop homepage server selector

Mobile:

Mobile homepage server selector

---

## Checklist

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
The intent here is good and the diff is tight. The account count now goes through `formatNumber` (main.ts:4595/4602/4618), which is the rule we want for numbers, the ambiguous "Players" label becomes "Total players", and the en-dash fallback is replaced with a plain hyphen (main.ts:4620). New strings are English `t()` keys in the right catalog and the test pins them. Two things to sort before this can merge.

**Blocking: the i18n resolved-hash gate fails.**
`src/ui/i18n.resolved.sha256` still holds the old baseline (`985c7e74...`), but the changed English copy makes the resolved table hash to `ebcf6133...`, so `tests/i18n_resolved_equivalence.test.ts` reds. I ran `scripts/i18n_resolved_hash.mjs` against the PR head to confirm. The branch is also cut from a stale base: it carries 21 locales versus release/v0.23.0's 22 (25 vs 26 resolved modules). Nothing to redesign, it is mechanical: merge release/v0.23.0, re-run `npm run i18n:build`, then `node scripts/i18n_resolved_hash.mjs --write`, and commit the regenerated `src/ui/i18n.resolved.generated/*` (including the locale the release added) together with the updated `i18n.resolved.sha256`.

**Should-fix: play.html was not synced (play.html:483-512).**
`play.html` is a separate build entry that carries the same server-select chrome and loads the same `shell.css` (via main.ts to index.css), but only index.html got the new `.server-stat-line` wrapper. The CSS now sets `.server-realm-line` and `.server-opt-stats` to `display: flex; flex-direction: column` (shell.css:1121), so on `/play` the count and its label stack on two lines, and the count falls through to the generic `.stats-value` rule (heading font, 16px) instead of the intended 13.5px UI style, because `.server-stat-line .stats-value` (shell.css:2011) never matches there. Mirror the index.html change into play.html: wrap both `.js-stat-accounts` lines (play.html:489 and 503) in `.server-stat-line` as index.html:595/609 do, and sync the static fallbacks (Online to Server Online, Players to Total players, the two descriptions, and the two `aria-label`s) so the pre-hydration markup matches.

**Nit:** `.server-stat-line span` (shell.css:2018) is a broad descendant selector. The only span in there today is the label so it is harmless, but scoping it would be safer if that markup ever grows.

Humpalumps:
@FernandoX7 should be fixed now

FernandoX7 (approved):

PR #1513, opened 2026-07-06
fix(sim): keep every world-boss damager on the loot roster, even after death

ryan-foo:
## Summary

Fixes a cluster of Thunzharr (world boss) loot problems reported from the live realm.

The world boss awards **personal** loot to every contributor, but the contributor roster was derived from the boss's **live hate table**. The sim prunes the hate table the instant a player dies (the classic "the dead fall off threat" rule in `handleDeath`), so a raider who died to the boss was stripped from the roster and got **no personal drop**. On top of that, the lootable corpse only lingered 300s, which could expire before players who died had resurrected and run back.

This PR:

- Adds a persistent `bossDamagers: Set<number>` on `Entity` that records every player (pet threat credited to the owner) who lands a hit on a world boss. Unlike the hate table, it is **never pruned** when a contributor dies, releases their spirit, leaves range, or drops off threat. So anyone who has hit the boss since it was pulled keeps their loot rights.
- Rolls world-boss loot from a new `worldBossLootContributors()` (the persistent damager set unioned with whoever still remains on the hate table, e.g. a healer who is still alive), sorted by `entityId` so the rng draw order stays deterministic.
- Extends the lootable-corpse window `300s -> 900s` (`WORLD_BOSS_CORPSE_SECONDS`) so a slain contributor has time to resurrect and run back before the corpse is removed. Still well inside the spawn cadence, so corpse windows never overlap.

On the group-loot concern: world-boss loot was **already** personal (`personalFor` slots looted straight to the player, never routed through a party need/greed roll), so a grouped raider's epic is never shared for the party to roll on. A new test pins that guarantee explicitly, including a party with the default need/greed premium-item strategy.

The damager set is server-sim-internal and is **never wired** to clients (it mirrors the existing `threat` map); loot resolves server-side. It is initialized in both the offline `Sim` (`baseEntity`) and the online `ClientWorld` (`blankEntity`) entity constructors.

## Related issues

Addresses the reported Thunzharr issues: corpse despawning before everyone loots, loot rights lost on death, and the request that anyone who hit the boss can loot.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/world_boss.test.ts tests/dead_party_loot.test.ts` (37 passed, incl. 3 new roster tests)
  - `npx vitest run tests/parity` (162 passed, 1 skipped: rng draw-order digest unchanged)
  - `npx vitest run tests/architecture.test.ts tests/entity_roster.test.ts tests/sim_context.test.ts tests/loot_roll.test.ts tests/autoloot.test.ts tests/raid_lockout_world.test.ts tests/loot_ffa_sim.test.ts tests/loot_drops.test.ts tests/loot_roll_wire.test.ts tests/world_api_parity.test.ts tests/loot_master_sim.test.ts` (all green)
  - `npx tsc --noEmit` clean (one pre-existing unrelated `prom-client` missing-dep error on the base branch, not from this change)
  - `npx @biomejs/biome check` on changed files: no format diffs
- New tests:
  - A contributor who **died** to the boss (pruned off the hate table) still lands on the loot roster and receives their `personalFor` slot.
  - A slain world-boss corpse lingers for `WORLD_BOSS_CORPSE_SECONDS` (>= 900s).
  - In a party, every world-boss slot is single-owner `personalFor`, and looting opens zero `pendingLootRolls` (an epic never enters a group roll).

## Notes for reviewers

- Determinism: the new recording draws no rng and is guarded by `MOBS[target.templateId]?.worldBoss`, so it cannot perturb the draw stream for ordinary mobs; the loot list is sorted by `entityId` before any per-contributor roll.
- The roster intentionally accumulates across a mid-fight evade rather than clearing with `clearThreat`: a knockback mechanic that briefly evades the boss must not wipe eligibility mid-fight (that would re-introduce the bug). Bounded by the per-boss loot lockout.

### DIFF
--- a/src/net/online.ts
+++ b/src/net/online.ts
@@ -860,6 +860,7 @@ function blankEntity(id: number): Entity {
     enraged: false,
     healedThisPull: false,
     threat: new Map(),
+    bossDamagers: new Set(),
     forcedTargetId: null,
     forcedTargetTimer: 0,
     ownerId: null,
--- a/src/sim/combat/damage.ts
+++ b/src/sim/combat/damage.ts
@@ -44,7 +44,7 @@ import {
   virtualLevel,
   xpForLevel,
 } from '../types';
-import { WORLD_BOSS_CORPSE_SECONDS, worldBossContributors } from '../world_boss';
+import { WORLD_BOSS_CORPSE_SECONDS, worldBossLootContributors } from '../world_boss';
 
 // How long a slain mob's corpse persists (seconds) before it is cleared. Sole user
 // is handleDeath, so the constant lives here with the death-domain code.
@@ -330,6 +330,16 @@ export function dealDamage(
     else if (source.ownerId !== null) target.tappedById = source.ownerId;
   }
 
+  // World-boss loot roster: every player (or pet owner) who lands a hit on a world
+  // boss becomes a permanent loot contributor. Unlike the hate table above, this set
+  // is NEVER pruned when they die, release their spirit, or drop off threat, so a
+  // raider who died to the boss still gets their personal drop. Read at death by
+  // worldBossLootContributors. Only world-boss templates ever populate it.
+  if (source && amount > 0 && MOBS[target.templateId]?.worldBoss) {
+    const contributorId = source.kind === 'player' ? source.id : source.ownerId;
+    if (contributorId !== null) target.bossDamagers.add(contributorId);
+  }
+
   if (source && source.kind === 'player' && source.id !== target.id) {
     const meta = ctx.players.get(source.id);
     if (meta) meta.counters.damageDealt += amount;
@@ -568,7 +578,7 @@ export function handleDeath(ctx: SimContext, e: Entity, killer: Entity | null): 
     // collapse with the boss: leaving them alive would harass looters for the
     // whole window, and a slain add's in-place respawn timer would revive it
     // mid-window (only fires for worldBoss templates, so no parity rng change).
-    const worldBossContribs = template?.worldBoss ? worldBossContributors(ctx, e) : null;
+    const worldBossContribs = template?.worldBoss ? worldBossLootContributors(ctx, e) : null;
     if (template?.worldBoss) {
       e.corpseTimer = WORLD_BOSS_CORPSE_SECONDS;
       e.respawnTimer = Infinity;
--- a/src/sim/entity.ts
+++ b/src/sim/entity.ts
@@ -97,6 +97,7 @@ function baseEntity(id: number, pos: Vec3): Entity {
     enraged: false,
     healedThisPull: false,
     threat: new Map(),
+    bossDamagers: new Set(),
     forcedTargetId: null,
     forcedTargetTimer: 0,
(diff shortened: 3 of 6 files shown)


FernandoX7 (commented):
Solid fix for a real bug: deriving the loot roster from the live hate table meant `handleDeath`'s "the dead fall off threat" prune silently stripped a fallen raider's personal drop. The separate never-pruned `bossDamagers` set is the right shape for it.

- Determinism holds. The recording draws no rng and is gated on `MOBS[target.templateId]?.worldBoss`, so it can't perturb the draw stream for ordinary mobs, and both rosters sort by `entityId` before any roll. `tests/parity` and `tests/architecture.test.ts` stay green. In the no-death case `worldBossLootContributors` and `worldBossContributors` resolve to the same roster (both dedupe, credit pets to owners, sort by id), which is why the goldens don't move.
- Entity-shape parity is intact: `bossDamagers` is added to both `baseEntity` (entity.ts:100) and `blankEntity` (online.ts:863), and left off the wire, which is correct since loot resolves server-side. `entity_roster` / `world_api_parity` pass.
- Crediting a pet's damage to the owner at damage time (damage.ts:339) is actually stronger than the old hate-table view: the owner survives even if the pet despawns before the boss dies.
- `spawnWorldBoss` builds a fresh entity via `createMob` -> `baseEntity` each spawn, so the set never leaks across spawns, and 900s stays well inside the 3600s cadence.

Two findings:

**should-fix** - The roster survives a full leash reset, not just a brief knockback, and the "bounded by the per-boss loot lockout" note doesn't actually bound that case. When a group wipes, the boss goes to `aiState = 'evade'` and `clearThreat` fires (locomotion.ts:275), but `bossDamagers` persists. If the boss then leashes home, heals to full, and is pulled and killed by a *different* group in the same spawn, the first group's members are still on `bossDamagers` and get `personalFor` slots on the corpse. The lockout doesn't stop this: it is consumed only when a player actually loots (per the `rollWorldBossLoot` comment), so a wiped member who never reached a corpse keeps their eligibility into the next pull. Consequence is mild (a reserved slot for an absent player, only unfair if they lingered nearby) and the scenario is rare, but the stated invariant is inaccurate. Worth either clearing `bossDamagers` on a genuine leash-home reset or softening the claim in the comment/PR body.

**nit** - A contributor who only healed or taunted (zero damage) and then died gets no loot. `bossDamagers` is written only on `amount > 0`, so a pure healer never lands in it, and death prunes them off the hate table. Not a regression (a live healer still gets loot via the hate-table union, a dead one never did), and it matches the PR's "damager" scope, but it's an asymmetry against the "everyone who contributed" goal. If you want symmetric coverage you'd fold healing/threat entries into the roster too.

Tests are decisive: the died-contributor case asserts `owners` contains the dead player, and the party case pins single-owner `personalFor` slots with zero `pendingLootRolls`. No i18n surface here.

ryan-foo:
Folded into #1643 (world-boss tuning + fixes) as its own commit, unchanged. Closing this one in favor of that PR.

PR #1514, opened 2026-07-06
feat(content): the Mirror World — a level-20 endgame zone (Vale of Glass), forward-ported to v0.23.0

Steakmushroompie:
## Summary

Forward-ports the **Mirror World** (Vale of Glass) — a level-20 endgame zone behind the Highwatch town mirrors — onto `release/v0.23.0`. Toward the v0.23.0 release (#1531). _(Originally opened against v0.22.0; retargeted and re-merged after v0.22.0 shipped — see "Retarget to v0.23.0" below.)_

A moonlit, mist-wreathed domed country reflected in black glass: a mirror town, the Mirrormere, a black canyon winding to the Sable gate, and a spirit-belt of bespoke creatures.

### Content
- **New zone** north of Thornpeak, sealed behind a trench that only the town-mirror portals cross.
- **Bespoke creatures with procedural bone-driven attacks:** the gloaming maw (chomp), voidfang stalker (lunge-bite), and spirit unicorn (gore), plus the reaper / gloomhulk / mirrorbound sentry / poverty ghost / mistshade / gargoyle.
- **The Keeper's Vigil** — a quest arc (Keeper Nerissa + Mistwitch Morwen) tying the vale into the main story, with `warden_sigil` / `drowned_lumen` quest items and the ghostly gear set.
- **Deepdream + Echo** bot-clone encounter; three town mirrors with gargoyle guardians.
- Animated ground mist, galaxy/mystical sky, corrected mob/statue orientation, and **full release-tier localization across all locales**.

### Integration notes (this is a forward-port)
The source branch was **915 commits behind**; this resolves the source conflicts against release, regenerates i18n + media manifest, and re-mints parity goldens. Adaptations to release's newer systems, called out for review:
- Shielded the black-canyon corridor from release's **terraced world rim** so it stays walkable to the Sable gate.
- `BiomeId` union of both sides; dropped an obsolete target-anchored combo-clear (release combo is character-bound).
- **Re-tuned the ghostly gear to release's reworked item-level budget** (first pass): `ghostly_mirrorblade` str8/sta5, `ghostly_glasscord` agi4/sta2, `ghostly_hushboots` spi3/int2 — happy to adjust.
- Registered the `bowed` (gargoyle-bow) snapshot delta key; added music theme trims; rebuilt the i18n resolved hash.
- Made two content-shift-fragile release tests robust (delve blackwater timer phase; a talent test's target wandering behind a collider and failing line-of-sight). **Gameplay code is release's, unchanged** — verified these pass on clean release.

### Retarget to v0.23.0
v0.22.0 shipped, so this PR was retargeted to `release/v0.23.0` and re-merged (85 commits ahead). The notable reconciliation:
- release/v0.23.0 **moved dungeon door/portal rendering out of `renderer.ts` into a new `src/render/door_portal.ts` module**. I adopted that refactor and **ported the Mirror World standing-mirror rendering into the same module** — `buildMirrorBody` reusing the module's shared portal geometry, plus a third `pad` portal tint (the silver Mirror shimmer) alongside `enter`/`exit`. `buildDoorBody`'s signature is unchanged. The mirrors still render and animate exactly as before.
- Merged the fog-state block: kept the Mirror World `dream` branch atop release's new `waterLevelAt(px, pz)`.
- Re-added the 4 Mirror World `IWorldQuests` members to the interface pin test against release's updated interface (185 members: 50 data / 135 method).
- Regenerated i18n / media manifest / resolved hash and re-minted parity goldens for the content-driven id/RNG shift.

### Verify
- `tsc --noEmit` clean; **0 conflict markers**; `biome ci` clean; i18n + manifest **freshness: 0 drift**.
- Full suite green in isolation; parity re-minted for the content-driven id/RNG shift.
- One **Windows-only** local artifact: `server/new_endpoint` shells out to the extension-less `node_modules/.bin/tsc`, which returns `ENOENT` on Windows — green on CI/Linux, unrelated to this change.


FernandoX7 (commented):
This is a big, careful port, and the load-bearing work holds up under scrutiny. I verified the deep claims against the code rather than the description: determinism is clean (every new draw goes through `Rng`, and the new tick passes are draw-free when inactive, so they cannot skew existing zones), sim purity is clean, the wire and IWorld surface is complete in both worlds (`bowed` encode/decode with a real dirty fixture and the `resyncQuests` clear, the three new commands, the 185-member literal pins), every new sim-emitted player string has a byte-identical `sim_i18n` matcher, and the universal golden re-mint is legitimately explained by a constant +57 entity-id offset from the zone's construction spawns plus the one new schema key (`permanentStats`) this PR itself adds. The two "content-shift-fragile" test rewrites are genuinely stabilizations, not weakenings: the Blackwater single-pulse still pins the 1.35 ratio decisively, and the pinned talent target keeps the decisive no-mana assertion. The 11 GLBs are valid binaries within the raised budget, the manifest is genuinely regenerated, and the CREDITS row matches the shipped files exactly. `tests/mirror_world.test.ts` is a strong suite with real negative cases.

Findings, blocking first.

**Blocking**

1. The /wiki World page breaks. The regenerated `src/guide/content.generated.ts:1294` now carries a fourth zone with biome `mirror`, and `src/guide/pages/world.ts:16-23` renders `guide.worldPage.<biome>Blurb/Greeting/Greeter/PlaceNotes` for every zone, but no `guide.worldPage.mirror*` keys exist anywhere at the head. `t()` throws on untracked keys outside release builds (`src/ui/i18n.ts:306`), so the page throws in dev and shows raw keys in production. No test renders that page, which is why CI missed it. Add the four English keys to `src/ui/i18n.catalog/guide.ts`.

**Should-fix**

2. `src/sim/sim.ts:6377`: the dream-win loot branch rolls each `player_echo` loot entry independently and ignores `rollGroup`, but all six ghostly entries are authored as one exclusive `ghostgear` group (the table's own comment says one exclusive draw per win). As written, one win can pay out multiple pieces, including both 5% rares together, on a repeatable encounter. Route it through the canonical roller in `src/sim/loot/loot_roll.ts`.

3. `src/sim/content/mirror_world.ts:443` and `:496`: `warden_sigil` (0.5 on every reaper) and `drowned_lumen` (1.0 on the Maw) lack `questId` gating, unlike the zone3 quest-token drops. They are quest-kind items (unsellable, untradeable), so after the one-shot chains they drop forever as pure bag pollution; the Maw's is a guaranteed drop on a farmable rare.

4. `enterDream` in `src/sim/sim.ts`: the Echo never receives the dreamer's Mirror-Shard stats. `echoMeta.permanentStats` is copied, but `recalcPlayerStats` reads the entity field, which `addPlayer` left at zeros and nothing re-mirrors before `resetForArena(echo)`. That contradicts the exact-mirror intent stated in the comment. One line: `echo.permanentStats = { ...echoMeta.permanentStats }` before the reset.

5. `src/game/music.ts:66-73`: the mirror overworld theme is fully composed and wired into `THEME_TRIM` and `COMBAT_TRANSPOSE` but unreachable: the biome-to-zone ternary has no `mirror` branch, so the zone outside the hub plays the peaks anthem.

6. `src/render/renderer.ts:3948-3963`: the dream ground disc is placed at z=0 and ignores its position args, but dream slots sit at z=-1250 to +150 (`dreamOrigin`) and dream fog far is 42. Offline solo play always duels at slot 0 (z=-1250), so the intended footing is invisible for essentially every duel. Place one disc per slot, or one at the local player's slot.

7. `src/ui/hud.ts:7992` with `src/sim/sim.ts:6435-6440`: when all 8 dream slots are busy, the deny path refunds the draught and emits only an error, but the near-opaque sleep overlay already faded in on `slumber` and only clears on dream arrival or `respawn`. The player is left behind a stuck indigo overlay until they next dream or die. Clear it on the deny path.

8. `play.html` is missing the new `<div id="dream-overlay">` that `index.html:425` gained. Every use is null-guarded so /play does not crash, but online players silently never get the sleep fade. (The recurring play.html parity trap.)

9. Em and en dashes on a couple hundred added lines: comments across sim/render/tests, the asset brief doc, the CREDITS row, and, more importantly, player-facing English copy (the slumber, echo-shatter, gargoyle-wake, and three shard log lines, plus quest and NPC text in `mirror_world.ts`). The repo bans these outright and the checked-in QA hooks and `npm run gate` reject added dash lines. When fixing the player strings, update their byte-identical `src/ui/sim_i18n.ts` matcher rows in the same change or the localization matcher stops resolving.

10. Architecture, two counts of growing coordinators the repo says never to grow: the Mirror World systems (about 630 lines: portal pads, dreams, gargoyle court, plus five private fields) landed as method clusters on `src/sim/sim.ts` instead of a module behind `SimContext`, and the Sundered Orrery (about 380 of the renderer's +566: canvas painters, the constructor build block, `updateMoons`) landed inline in `src/render/renderer.ts`. This same PR proves the right pattern both times (the `mist.ts` and `sealife.ts` extractions, the appended SimContext callbacks), so the extraction is mechanical. Fine as an immediate follow-up if landing the port first is preferred.

11. `scripts/wiki/build_content.mjs:234`: the bestiary union is hardcoded to ZONE1/ZONE2/ZONE3/TEMPLE, so the seven camped mirror creatures never reach /wiki, and the new zone card's resident links get computed from other zones' level 15 to 20 creatures. The committed generated file passes the freshness gate because the generator itself is incomplete. Add the mirror mobs and camps to the union and regenerate.

12. Dead plumbing, two pieces. `src/sim/quests/quest_commands.ts:271` emits `mirrorUnlocked` but no host handles it, while comments in three places promise the client re-lights the pad (pads also render identically locked vs unlocked). Either implement the relight and a dark state, or drop the event and fix the comments. And `mirrorClass`/`mc` is encoded (`server/game.ts:587`), decoded, and read in the mob render branch, but nothing ever assigns it; the Echo is a player-kind clone so the player branch already renders the right model. Wire a producer or remove the field.

13. Test gaps worth closing in `tests/mirror_world.test.ts`: no negative for brewing with the recipe done but zero parts in bags (the guard at `sim.ts:6683` is the free-draught protection), no negative for brewing away from Morwen, and no load-side round-trip for `permanentStats` (save is pinned, but a relog that dropped the shard bonus would pass today; old rows without the field deserve the no-throw check too).

14. CI is red: `tests/fiesta.test.ts:273` (practice determinism) timed out at 5s, 1 failed of 9930. That test builds the Sim twice, so it is consistent with the heavier world init this zone adds, the same reason you widened `parity.test.ts` to 20s. It likely needs the same widened budget; worth confirming it is not just runner contention.

**Nits**

- `index.html:425`: the decorative `#dream-overlay` should carry `aria-hidden="true"` like its sibling `#quest-banner`.
- `src/styles/hud.css:3651`: the overlay's motion calming covers the in-game reduce-motion toggle and fx-low but not OS `prefers-reduced-motion`; the adjacent vignette section handles all three.
- `src/ui/hud.ts:9093`: the gargoyle-bow handler re-renders the gossip dialog on a 120ms timeout with no still-open guard, so closing it inside the window reopens it and steals focus.
- `src/ui/hud.ts:5573`: unconditional per-frame `classList.remove` in an otherwise write-elided path.
- `src/render/mist.ts:28` (plus renderer.ts and sealife.ts): the 6th, 7th, and 8th verbatim copies of `mulberry32` in `src/render`; a tiny shared helper would serve all eight.
- `src/render/water.ts:124`: the gloom band hardcodes 897/917 in the shader instead of deriving from `MIRROR_LAYOUT.zMin`; safe today only because this is the last zone.
- `src/sim/content/zone3.ts:38`: the POI label says The Tidegate but the pad at those exact coordinates is named The Mirrorgate.
- `src/sim/content/mirror_world.ts:952`: the warrior-archetype reward for the lumen quest is the leather agi chest while the set's str mail piece is unused by any quest.
- `src/ui/i18n.catalog/guide.ts:689` (and a few similar spots): the world-page framing copy still says three zones above what is now four zone cards.
- `scripts/asset_budget.mjs:20`: the raised creatures budget fits with about 4% headroom; the next model busts it. Fine, just flagging the ceiling.

On the gear question in the description: the retuned stats land exactly on the `item_level.ts` budget for each piece's source, so ship the stat lines as-is. Armor on the set and the mirrorblade's weapon damage sit below zone3 peers at the same level; worth a small bump only if they are meant to compete.

Conflicts: the current CONFLICTING state is exactly the generated i18n trio (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`), mechanical regen churn. Merge the base and re-run the i18n build; everything else auto-merges.

Steakmushroompie:
Thanks for the thorough review, Fernando - all findings addressed in ec4b0010.

**Blocking**
- /wiki World page: added the four `guide.worldPage.mirror*` keys and translated them across all locales (i18n_completeness passes - no English leak in the non-Latin ones). The `mirror`-biome zone card now renders instead of throwing.

**Should-fix**
- Dream-win loot routes the ghostgear group through the exclusive `rollGroup` roller (one piece per win) like `loot_roll.ts`; copper/potion still roll independently.
- `warden_sigil` / `drowned_lumen` are `questId`-gated like the zone3 tokens (stop dropping once done/unaccepted).
- `enterDream` sets `echo.permanentStats` before `resetForArena` so the Shard bonus folds into the clone's stats.
- `music.ts` gained the `mirror` overworld branch (hub stays `town_mirror`).
- Dream ground disc is placed per slot.
- Sleep overlay clears on the slots-full deny path via a new `slumberEnd` event.
- `play.html` gained `#dream-overlay`.
- Em/en dashes purged from every added line; `sim_i18n` matcher rows kept byte-aligned (S3 guard green).
- Wiki bestiary now includes the 7 mirror creatures, with committed rendered stills.
- Dropped the dead `mirrorUnlocked` event and the never-assigned `mirrorClass`/`mc` field (the bow -> `bowedGargoyles` -> `requiresBow` gate is the real unlock path; the toll's nova + "the mirror clears" log is the feedback).
- Added brew-negative tests (no parts / away from Morwen) and a `permanentStats` relog round-trip (+ old-row no-throw); widened the fiesta determinism timeout.

**Nits** - all applied: aria-hidden, OS `prefers-reduced-motion`, gargoyle-bow re-render still-open guard, overlay classList write-elide, `mulberry32` dedup into `src/render/rng.ts`, water gloom band from `MIRROR_LAYOUT.zMin`, POI `The Tidegate` -> `The Mirrorgate`, warrior lumen reward -> the str mail piece (`ghostly_palegrips`), and the world-page "three -> four zones" copy.

**Deferred** (as you offered): the ~1000-line coordinator extraction (Mirror systems off `sim.ts` behind `SimContext`, the Sundered Orrery out of `renderer.ts`) will be a focused follow-up PR so this re-review stays on the correctness fixes. Asset-budget headroom + the gear armor/damage bump left as-is per your "ship as-is" note.

Ready for another look when you have a moment.

PR #1516, opened 2026-07-06
fix(render): boot prewarm must skip lazyPreload visuals (renderer crash)

BlizzHacker:
## Problem

The boot prewarm sweep (`buildEntityPrewarmGroup` in `renderer.ts`) builds one copy of every distinct mob visual up front to compile its shader. But a visual flagged `lazyPreload: true` is deliberately excluded from the boot preload sweep — its GLB is fetched on demand the first time such a mob spawns. Building one at boot throws `character asset not preloaded` and takes down the renderer (the client sees "the server closed the connection").

This is latent on `main` today: `player_mech` is flagged `lazyPreload: true`, so any code path that makes prewarm iterate it will crash the boot.

## Fix

The **preload** sweep already guards this — `manifest.ts` has `if (def.lazyPreload) continue;`. The **prewarm** sweep did not. This adds an `isVisualLazy()` helper next to `visualKeyFor()` and skips lazy visuals in prewarm the same way, still marking the key built so the tail loop doesn't retry it. Lazy models warm up correctly on their real first spawn; eagerly-preloaded visuals are unaffected.

## Test

`tests/prewarm_lazy_visual.test.ts` covers `isVisualLazy()` over a lazy visual (`player_mech`), an eager one (`player_warrior`), and an unknown key.

- `npx tsc --noEmit` clean
- `npx vitest run tests/prewarm_lazy_visual.test.ts tests/architecture.test.ts` green
- `npm run build` succeeds

### DIFF
--- a/src/render/characters/manifest.ts
+++ b/src/render/characters/manifest.ts
@@ -1048,6 +1048,14 @@ const NPC_KEYS: Record<string, string> = {
   spirit_healer: 'npc_villager_robed',
 };
 
+/** True if a visual model is excluded from the boot preload sweep (lazyPreload).
+ *  The boot prewarm must SKIP these — their GLB isn't loaded yet, so building one
+ *  would throw "character asset not preloaded". They load on demand when first
+ *  needed (e.g. the first time such a mob spawns in-world). */
+export function isVisualLazy(visualKey: string): boolean {
+  return !!VISUALS[visualKey]?.lazyPreload;
+}
+
 export function visualKeyFor(e: Entity): string {
   if (e.kind === 'player') {
     if (e.skinCatalog === 'mech') return 'player_mech';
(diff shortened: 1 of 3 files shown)


Rubsey (commented):
## OSSBrain initial review

### Should fix

- **New comment violates the no em dash invariant** (`src/render/characters/manifest.ts:1341`)
  - Issue: The PR adds an em dash in a code comment: "SKIP these — their GLB". The root CLAUDE.md invariant bans em dashes in code, comments, docs, commits, PR text, and player-facing copy.
  - Impact: This introduces new punctuation debt and can trip the repository copy guard or release gate even though the runtime change is otherwise narrowly scoped.
  - Suggested fix: Replace the em dash with allowed punctuation, for example: "SKIP these: their GLB" or "SKIP these because their GLB".

### Checks

- `npx vitest run tests/prewarm\_lazy\_visual.test.ts tests/render\_asset\_preload.test.ts`: **passed**. 2 test files passed, 9 tests passed.
- `git diff -U0 376f125b3ad55b3b8a73b6500310200835835af1..2c73c7a34e300ea699c574050a8ab6b6ad9b3882 -- src/render/characters/manifest.ts | rg -n "^\\+.\*\[—–\]"`: **failed**. Found a newly added em dash in src/render/characters/manifest.ts.

PR #1519, opened 2026-07-06
fix(world): make water height terrain/feature-aware, not a flat per-zone plane

### LINKED ISSUE #1518: Water level is a single absolute height per zone, not terrain/feature-aware
## Problem

`WATER_LEVEL` in `src/sim/world.ts` is a single fixed absolute height
(-4.5) applied to an ENTIRE zone's footprint, not just its mapped lakes.
`src/render/water.ts` draws one water plane per zone at that same height
("Water for the whole zone strip"). In practice this means any terrain
anywhere in a zone that dips below -4.5 for any reason, not just the
intended lake basins, renders as flooded, whether or not that spot is
meant to be a lake.

This also creates a hard floor on how deep any sunken terrain feature (a
crater, a sinkhole, a below-grade tunnel, an editor-authored depression)
can go while staying walkable: `src/sim/pathfind.ts` blocks a
non-swimming player's destination once ground drops below
`waterLevel() - PLAYER_SWIM_DEPTH` (currently -5.3). Content authors
cannot build a meaningfully deep underground feature without either
accidentally flooding it or making it unwalkable, since water height is
not aware of terrain shape or authorial intent at all, only a global
constant.

Realistically, water height should be derived from terrain conditions
(the shape of a lake basin, an explicit per-feature water surface, or at
minimum a bounded region) rather than one flat height applied to the
whole zone regardless of what's actually built there.

## Expected behavior

- Water only renders where a zone's content actually places a lake or
  other water body (for example, via the existing `lakes` list, or a
  bounded water region tied to a specific terrain feature), not
  wherever terrain height happens to dip below a single global constant.
- A content author can build a sunken terrain feature (a crater, a
  sinkhole, an underground tunnel/room) meaningfully below the current
  -4.5 without it silently flooding, as long as it isn't inside a
  declared water body.
- The walkable-depth floor (`waterLevel() - PLAYER_SWIM_DEPTH` in
  `src/sim/pathfind.ts`) is derived per-location from the same
  terrain-aware water model, not from one flat global height, so a
  legitimately dry, deep feature isn't blocked from normal (non-swim)
  walking.

## Scope

- `src/sim/world.ts`: `WATER_LEVEL`, `waterLevel()`, and how they feed
  `terrainHeight`/`groundHeight`.
- `src/render/water.ts`: the per-zone water plane (currently drawn across
  the whole zone footprint at `waterLevel()`).
- `src/sim/pathfind.ts`: `rideHeight`, `playerDestinationWalkable`, and
  the `PLAYER_SWIM_DEPTH` floor, all keyed off the same global
  `waterLevel()`.
- `src/sim/types.ts` / `WorldContent`: the existing per-map
  `waterLevel` override (map_doc.ts `MIN_WATER_LEVEL`/`MAX_WATER_LEVEL`)
  is also a single scalar per map; the same terrain-awareness gap applies
  there.

## Acceptance criteria

- A sunken terrain feature placed outside any declared lake/water body,
  at any depth, does not render as flooded and does not block normal
  walking, regardless of how it compares to today's single global
  `WATER_LEVEL`.
- Existing lakes (`MIREFEN_IMPACT_CRATER`, the zone `lakes` lists, etc.)
  keep rendering and behaving exactly as they do today: this is additive,
  not a rework of existing water bodies.
- `tests/architecture.test.ts` and the terrain/pathfind test suites stay
  green; new tests cover a dry sunken feature adjacent to (but outside)
  a declared water body.

## Notes

Surfaced while building the Glimmervein Cavern content PR (#1509): an
underground tunnel needs to sit meaningfully below the surrounding
grade, but the current global water plane and swim-depth pathing floor
made anything deeper than about one yard below -4.5 either visibly
flood or become unwalkable, well short of reading as genuinely
underground. Content had to stay shallow (-3.8) to work around this.

jgyy:
## Summary
- Water was one flat `WATER_LEVEL` plane applied to an entire zone's footprint (`src/render/water.ts`), and the walkable-depth floor in `src/sim/pathfind.ts` was a single global scalar (`waterLevel() - PLAYER_SWIM_DEPTH`). Any terrain that dipped below that height anywhere in a zone, not just its mapped lakes, read as flooded and became unwalkable for a non-swimming player.
- Adds `isInWaterBody` / `waterLevelAt` / `waterBodies()` to `src/sim/world.ts`: water is now derived from each zone's declared `lakes` list (the same basin-blend footprint the terrain shaping already uses). Outside every declared lake, `waterLevelAt` returns `-Infinity`, so a content author can build a sunken feature (crater, sinkhole, tunnel) arbitrarily deep without it flooding or blocking movement.
- `src/sim/pathfind.ts`'s `minGround` is now a per-position function; `src/sim/sim.ts`'s buoyancy/swim gates (`isSwimming`, charge/follow/flee movement, mob locomotion, fishing) all read the same per-position water surface instead of the global constant.
- `src/render/water.ts` now draws one plane per declared lake (`waterBodies()`) instead of one plane spanning an entire zone's width.
- Existing lakes (Eastbrook Vale, Deepfen Shallows, Mirefen, zone 3) are unchanged: same footprint, same rendering, same swim gating.

Fixes #1518.

## Diagram
before/after

## Screenshots
Each declared lake still renders correctly as its own water body after switching from one flat per-zone plane to one plane per declared lake.

Zone 1 (Eastbrook Vale):
zone 1 lake

Zone 2 (Deepfen Shallows, two lakes visible):
zone 2 lakes

Zone 3:
zone 3

## Test plan
- [x] `npx tsc --noEmit`
- [x] New `tests/water_terrain_awareness.test.ts`: declared-lake detection, a sunken editor stamp well below the old global floor stays dry and walkable (no depth-floor block), and a real lake still requires swim to enter.
- [x] `tests/pathfind.test.ts`, `tests/architecture.test.ts`, `tests/terrain_walls.test.ts`, `tests/custom_map_parity.test.ts`, `tests/climb_slope.test.ts`, `tests/sim.test.ts` and other water/terrain-touching suites all green.
- [x] `npm run gate` (full CI-equivalent: i18n, malware scan, changed-files biome, full test suite, tsc, all builds) passes.
- [x] Manual visual check: all three zones' declared lakes still render as expected water bodies.


jgyy:
Thanks for the review! Addressed both render-side gaps in 5e6c7c44f:

- Swim pose (`renderer.ts:4229`), underwater fog (`renderer.ts:3742`), footstep timbre (`renderer.ts:1515`), and water ambience (`renderer.ts:5121`) now all read `waterLevelAt(x, z)` instead of the flat `waterLevel()`, so a dry sunken feature outside every declared lake never triggers a swim/underwater/water-ambience visual or SFX.
- `water.ts`: the shader tier now discards fragments outside the declared lake radius (`uCenter`/`uRadius` uniforms, one material instance per lake), and the Phong (low) tier switched from a square `PlaneGeometry` to a `CircleGeometry` disc, so corner geometry outside a lake's footprint no longer renders as water in either tier.
- Added a test to `tests/water_terrain_awareness.test.ts` covering the render-side swim-pose/ambience predicates against a dry sunken feature.

`npx tsc --noEmit`, the water/pathfind/architecture suites, and the pre-push QA floor are all green.

jgyy:
Thanks, fixed in 61a3533b2:

- `blockedTowardSpawn()` (`src/sim/mob/locomotion.ts:699`) now reads `waterLevelAt(nx, nz)` instead of the flat `waterLevel()`, so a dry sunken feature outside every declared lake no longer holds a non-swimming evader in phasing on its way back to spawn.
- Added a regression test in `tests/mob_locomotion.test.ts` that builds the same sunken-feature world content used in `tests/water_terrain_awareness.test.ts` and asserts `blockedTowardSpawn` returns `false` for a non-swimmer evading across it.

Left the render-side test as-is for now since restructuring it to exercise the renderer function directly would need extracting a small pure predicate out of `renderer.ts`, happy to follow up separately if wanted.

`npx tsc --noEmit`, `tests/mob_locomotion.test.ts`, `tests/water_terrain_awareness.test.ts`, `tests/pathfind.test.ts`, and `tests/architecture.test.ts` are all green, and `npm run gate`'s pre-push QA floor passed on push.

jgyy:
Thanks, fixed both in da6d67813:

- `src/render/fish.ts:140` now derives `depthAt` from `waterLevelAt(x, z)` instead of the flat `waterLevel()`, so a dry sunken feature outside every declared lake never gets fish splashes.
- `src/ui/map_terrain.ts` now classifies water per-pixel via `waterLevelAt(x, z)` instead of a single hoisted `waterLevel()`, so the world map/minimap no longer paints a dry sunken feature blue.
- Added `Renderer.rebuildWaterBodies()`, which disposes every existing lake mesh and rebuilds from the current `waterBodies()` list, and wired it into the editor's `rebuildTerrainFull()` (in place of the level-only `rebuildWater()`). Moving, adding, or removing a lake marker in 2D now reconciles the water mesh, shader `uCenter`/`uRadius`, and shore-depth to the new footprint when returning to 3D, instead of leaving the old mesh reseated at the old spot.
- Added two tests to `tests/water_terrain_awareness.test.ts` that exercise the actual render-side functions (`isLeapableWater` composed with the real `depthAt` formula, and `paintTerrainRows`) against a dry sunken feature, rather than duplicating the predicate math.

`npx tsc --noEmit`, the water/fish/map_terrain/architecture/pathfind/mob_locomotion suites, and the pre-push QA floor are all green.

FernandoX7 (commented):
Solid fix, and the diagnosis is right: the flat per-zone plane plus a single global `waterLevel() - PLAYER_SWIM_DEPTH` floor is what forced any low terrain to read as flooded. Making `waterLevelAt` return `-Infinity` off every declared lake is a clean way to do it: every `groundHeight < waterLevelAt - depth` comparison just goes false off-lake, so the movement, swim, charge, follow, fog and footstep gates all short-circuit with no NaN. Built-in terrain stays byte-identical (LAKE_BLEND_RADIUS_MULT is the same 1.6, the smoothstep is unchanged, custom_map_parity stays green), and the per-lake shader discard plus CircleGeometry for the low tier is the right way to keep the footprint honest at both tiers. The `blockedTowardSpawn` regression test is a good catch.

I ran the touched suites at the head ref (water_terrain_awareness, mob_locomotion, custom_map_parity, pathfind, architecture, terrain_walls, climb_slope): 80 tests, all green.

should-fix
- The cosmetic water decorators still key off the flat `waterLevel()`, so the exact dry-deep feature this PR enables reads as water to them. `src/render/fish.ts` (`isLeapableWater` / `depthAt`: fish leap wherever `waterLevel() - terrainHeight >= margin`), the shore tint at `src/render/terrain.ts:350`, and the water-exclusion at `src/render/foliage.ts:1176,1500`, `src/render/motes.ts:97`, and `src/render/critters.ts:175` all treat "terrain below the flat water line" as water with no lake gate. On a custom map, a crater or tunnel carved below `WATER_LEVEL` outside every declared lake would show fish leaping in dry air, a sandy shore band at its rim, and no foliage/critters/motes, which is the thing the fix is meant to prevent. `renderer.ts` got exactly this class of fix in the swim-pose/fog/ambience/footstep commit; these five did not. The built-in world can't trigger it (the `minLand = waterLevel() + 1.4` soft floor keeps non-lake terrain above the water line), so it only bites authored custom maps and is purely cosmetic, hence not blocking. Worth switching them to `waterLevelAt` (or gating on `isInWaterBody`) so the feature fully delivers.

nit
- `waterLevelAt` / `isInWaterBody` now run in per-frame and per-tick hot paths (the swim-pose block samples it per visible entity each frame at `renderer.ts:4228`; several locomotion/movement gates sample it per mob each tick), and each call loops every zone's lakes with a `Math.sqrt`. It is tiny for the built-in ~5 lakes, but the test is a squared-distance one (`dx*dx + dz*dz < (radius * MULT) ** 2`) and could skip the sqrt, in the spirit of the frugality the surrounding comments call out.
- The editor Water tool hint (`editor.water.hint`) still reads "Sets the map-wide water surface height", and the level slider used to flood everything below it across the map. After this change the level only moves declared lakes' surface, so raising it on a lake-less map does nothing. That matches the issue's intent, but the hint is now misleading; worth a wording pass so an author doesn't reach for the level slider to flood a valley.

Nothing wire-facing here (no new Entity fields), sim purity holds, and there are no new player-visible strings.

jgyy:
Thanks for the thorough review! Fixed both should-fix items and both nits in db57c94bf:

- `src/render/terrain.ts:350` (shore tint), `src/render/foliage.ts:1176,1500` (dressing/grass water exclusion), and `src/render/motes.ts:97` / `src/render/critters.ts:175` (the flat "over water" gates you didn't explicitly list but share the same predicate) now all key off `waterLevelAt(x, z)` instead of the flat `waterLevel()`, so a dry sunken feature outside every declared lake gets no shore tint and keeps its foliage/grass/motes/critters instead of reading as flooded.
- `isInWaterBody`'s per-lake distance check now compares squared distances instead of taking a `Math.sqrt` per call, since it runs in the swim-pose/locomotion hot paths.
- `editor.water.hint` and the water tool's label copy now say "each declared lake" / "a declared lake's" instead of "the map-wide" water level, matching the new per-lake behavior.

`npx tsc --noEmit`, the water/foliage/terrain/map_terrain/i18n suites, and the pre-push QA floor are all green.

FernandoX7 (approved):

PR #1509, opened 2026-07-06
feat(content): rebuild Glimmervein Cavern as a winding sunken trench, west side

jgyy:
## Summary

Rebuilds **Glimmervein Cavern** as a winding sunken trench on the WEST side
of both Eastbrook Vale (Zone 1) and Mirefen Marsh (Zone 2), entering and
leaving each zone at roughly its own vertical middle (the "9 o'clock"
position on each zone's map), snaking left and right the whole way rather
than running straight down a single x.

- **No wall, ceiling, or pillar geometry at all.** A chain of overlapping
  `'smooth'`-falloff `HeightStamp`s (the same technique the game's existing
  `MIREFEN_IMPACT_CRATER` and its lake basins already use) pulls the ground
  down to a fixed floor; each waypoint's own concave slope IS the wall, the
  same way a lake basin's shore needs no fence. `src/render/cave_tunnel.ts`
  now only places glowing crystal clusters along the path.
- **Winds like a real subway line.** `GLIMMERVEIN_WAYPOINTS` in
  `src/sim/data.ts` snakes between x=-26 and x=-107 as z runs from 0 to
  390, threading the gap between Mirror Lake (Zone 1) and Deepfen Shallows
  (Zone 2), both of which already occupy the far-west band at almost
  exactly this trench's own z-range.
- **A real engine limit surfaced twice building this:** the world's water
  plane is one flat height per zone (`WATER_LEVEL=-4.5`, not aware of
  terrain or intent), and `src/sim/pathfind.ts` blocks non-swim movement
  below `waterLevel() - PLAYER_SWIM_DEPTH` (-5.3). Both mean a walkable
  sunken feature here can only go a few yards deep, not "as deep as
  possible": going deeper either floods (this PR's floor got that wrong
  once, caught in its own interior screenshots and fixed) or blocks
  movement outright. Filed **#1518** to track a proper terrain-aware water
  model as a follow-up; this PR stays within the current walkable band
  (`GLIMMERVEIN_FLOOR_Y = -3.8`).
- Same content otherwise: `glimmervein_spider`/`glimmervein_bat` (Zone 1,
  lvl 5-7), `glimmervein_broodling` (Zone 2, lvl 7-8), the `crystal_shard`
  junk item, and the ore vein pair, all open access with no quest gate.

## Map

```mermaid
flowchart LR
    subgraph Z1["Eastbrook Vale, west side ('9 o'clock')"]
        M1["z=0 mouth"]
    end
[...]

PR #1526, opened 2026-07-06
feat(content): The Sunken Road, a deep tunnel from Eastbrook Vale to Mirefen Marsh

jgyy:
## Summary

**The Sunken Road**: a deep, walkable, fully **enclosed** tunnel connecting
**Eastbrook Vale** (Zone 1) directly to **Mirefen Marsh** (Zone 2), on the
west side of the map ("9 o'clock" on each zone's world map).

This branch was tracking issue #1518 ("water level is a single absolute
height per zone, not terrain/feature-aware"). That fix is already merged
via `release/v0.22.0` (`055721ce4`, `db57c94bf`, `da6d67813`, `61a3533b2`,
`5e6c7c44f`), so this PR repurposes the branch to build the feature that fix
unblocks: a genuinely deep underground route between the two towns. Without
terrain-aware water, a sunken feature like this could only carve a few yards
below the old flat `WATER_LEVEL` before flooding or becoming unwalkable,
which is exactly what forced Glimmervein Cavern (#1509) to stay at a
shallow `-3.8`. This tunnel carves to **-24**, as a real NARROW corridor
(not a wide open bowl) with a rock shell overhead, so the sky is never
visible while walking it: a mined-out tunnel, not a valley.

## The key fix this round

An earlier version of this tunnel used one wide `'smooth'`-falloff
`HeightStamp` per sparse waypoint. That carves a broad, gently-sloped
**bowl** (literally a valley shape), and the visual shell (a narrow tube)
only covered a fraction of that wide floor: standing almost anywhere on the
walkable floor put you outside the shell entirely, looking straight up at
open sky.

**A tunnel's floor must be no wider than its own ceiling.** Fixed by:
- Narrowing the terrain carve to a real corridor (16yd wide), `'flat'`
  falloff (sharp walls, no smoothing) instead of `'smooth'` (gradual bowl
  slope). The walls don't need to be walkable, only impassable, so a sharp
  edge is correct.
- Densifying the sparse route into a close-spaced point chain so the
  narrow, non-blending `'flat'` stamps stay continuous along the whole
  route instead of leaving gaps.
- Widening the render shell so it comfortably exceeds the new narrow floor
  radius (with real margin, where before it was narrower than the floor it
  was supposed to cover), swept along that same dense centerline so it hugs
  every turn tightly.

## Playtest follow-up (this round)

Four issues came back from playtesting the tunnel above, all fixed:

- **Entrances were invisible cliffs, not walkable ramps.** Both mouths
  opened directly onto natural grade (~-3) right next to the interior's
  sharp full-depth carve (-24): a player walking toward the mouth hit a
  20-yard sheer, invisible-until-the-last-step drop instead of finding a way
  in. Added a graded ramp (a short chain of wide, `'smooth'`-falloff
  `HeightStamp`s) tapering each mouth back up to natural grade along the
  open-field side, ordered so the full-depth interior carve still wins at
  the corridor itself. Both mouths now read as a real descending mine
  entrance you can walk straight into.
- **The tunnel interior rendered near-black, not stone-grey.** The shell
  used a near-black color (`0x241f1a`) lit only by non-illuminating emissive
  crystal meshes (the outdoor sun rig is dropped underground, so nothing
  actually lit the rock). Recolored the shell to a bedrock grey matching the
  cave biome palette, and gave each crystal a real point light in the
  renderer's shared light budget (the same scheme dungeon torches use), so
  the floor, ceiling, and walls now read as lit stone, not a black void.
- **Mirror Lake (and other lakes) painted a broken double ring on the world
  map.** The map painted "is this water" from a height threshold, but a
  lake basin's smoothstep blend can dip back below the waterline a second
  time near its own outer blend radius (invisible in 3D, hidden under the
  flat water plane) — producing a disconnected "moat" ring around the lake
  on the flat 2D map. Now gates on the lake's actual declared circular
  footprint, matching what the 3D water plane really covers: every lake is
  a clean single circle on the map now.
- **No way to tell the tunnel is underground on the map.** Opening the map
  window while standing inside the Sunken Road now shows a dedicated dark
  schematic (mirroring the existing delve-map mechanism) with the
  centerline, both mouth labels, and the ridge crossing, instead of the
  regular surface map (which has no sensible way to depict a corridor that
  runs straight down). The regular world map still just shows the entrance
  as a single POI, as before.

## Playtest follow-up, round 2

More playtest feedback on the fixes above, all addressed:

- **Walls still read as green grass/marsh, not grey rock.** Only the render
  shell's own material was recolored last round; the terrain underneath it
  (and the world map) still colored by the surrounding zone's biome, since
  the built-in world had no biome-paint layer at all. Added a `BiomePaint`
  grid covering the tunnel + its mouth ramps, painted as the 'cave' biome
  (`sim/content/sunken_road.ts`, wired into `BUILTIN_WORLD`). Both the
  terrain renderer's ground color and the world-map painter read the same
  `biomeAt` lookup, so one paint layer fixes the walls in 3D and gives the
  tunnel its own grey scar on the 2D map.
- **Entrance holes were too wide / read as an open valley, not a tube.** The
  mouth ramp's own carve radius (15) was nearly double the tunnel's interior
  radius (8), flaring each entrance into a broad shallow bowl. Narrowed the
  ramp to radius 10 (just wider than the interior) and re-tuned its
  depth/offset steps, so the surface opening is now a small, tunnel-width
  hole instead of a wide valley mouth.
- **Flat black ceiling with gaps showing the surface right at the
  entrance.** The render shell stopped dead at the mouth waypoint while the
  (still-rising) ramp terrain continued outward, leaving a gap between rock
  ceiling and ground that read as a hole straight through to open sky.
  Extended the shell a short way over each ramp's closest points, rising
  with the ramp's own floor depth instead of hovering flat, so the tube
  visibly encloses the entrance instead of ending abruptly.
- **Long unlit black stretches between crystal lights.** Only the 11 sparse
  waypoints had a light, up to 50 yards apart, well past each light's own
  falloff distance. Added a light every ~18 yards along the tunnel's dense
  centerline (keeping the decorative crystal meshes at the original sparser
  spacing so the interior doesn't get visually cluttered), so the tube stays
  lit throughout its length.
- **Mirror Lake (and every lake) looked like an unnaturally perfect circle
  on the world map.** Added a deterministic per-angle noise jitter to the
  map's painted shoreline (cosmetic only; the actual water-body/walkable
  footprint in 3D is unchanged) so every lake now reads as an organic,
  non-geometric coastline instead of a drafting-compass circle.

Also caught and fixed a screenshot-script bug along the way: several
interior shots still targeted a waypoint this branch itself had bulged
outward for Mirror Lake fishing-spot clearance, so those specific shots were
accidentally standing on the open surface instead of underground.

## Map

```mermaid
flowchart LR
    subgraph Z1["Eastbrook Vale"]
        M1["z=-15, x=-95\nmouth, west of town (9 o'clock)"]
    end
    subgraph T["The Sunken Road: narrow enclosed corridor, floor y=-24"]
        M1 --> W1["x=-100..-145,\nz=20..140,\nducks east of\nMirror Lake"]
        W1 --> Ridge["ridge crossing\nz=180, x=-110"]
        Ridge --> W2["x=-110..-140,\nz=205..258,\nclear of mire_prowler\n+ Deepfen Shallows"]
        W2 --> M2
    end
    subgraph Z2["Mirefen Marsh"]
        M2["z=258, x=-140\nmouth, west of town (9 o'clock)"]
    end
```

## Content flow

```mermaid
flowchart TD
    Delke["Foreman Delke\n(Eastbrook mouth NPC)"] -->|"gives"| Quest["Clearing the Sunken Road"]
    Quest -->|"kill 8"| Gravemite["Tunnel Gravemite\n(lvl 6-7)"]
    Quest -->|"kill 6"| Stalker["Deep Road Stalker\n(lvl 7-8)"]
    Ridge2["Ridge-crossing camp"] --> Prospector["The Old Prospector\n(named elite, lvl 8)"]
    Gravemite -.->|"loot"| Grit["Sunken Road Grit"]
    Stalker -.->|"loot"| Grit
    Prospector -.->|"guaranteed loot"| Grit
```

## Iteration history

```mermaid
flowchart LR
    A["#1518 fixed upstream\n(release/v0.22.0)"] --> B["v1: open-air carved\nvalley (wide radius,\nsmooth falloff, no shell)"]
    B --> C["v2: added a tube shell,\nbut narrower than the\nwide carve - sky still\nvisible off-center"]
    C --> D["v3: narrowed the CARVE\nitself to match the shell -\na real mined corridor,\nsky never visible"]
    D --> E["v4 (current): graded\nentrance ramps, lit the\ninterior, fixed the map's\nlake ring, added an\nunderground map"]
```

## Screenshots

**Eastbrook mouth**, seen from outside on natural grade, facing down the graded ramp into the tunnel:

entrance wide zone1

Standing at the mouth itself: a visible descending ramp into the dark opening, not a sheer drop:

approach

Foreman Delke, the quest giver, right at the Eastbrook mouth:

foreman delke

Interior, Zone 1 stretch: lit bedrock-grey rock, not black, with a real ceiling closing overhead:

interior 1

interior 3

**Looking straight up from deep inside the tunnel**: solid lit rock, no sky at all:

looking up

Close on one of the crystal glow lights, showing the tunnel is lit by its own light sources:

crystal light

**Cross-section, from outside/beside the corridor**: reads as a narrow lit tunnel cut through the hillside, not a valley:

cross section

Ridge crossing (z=180): a narrow enclosed corridor under the zone1/zone2 mountain wall, The Old Prospector ahead:

ridge crossing

old prospector

Interior, Zone 2 stretch, and a Tunnel Gravemite / Deep Road Stalker camp along the route:

interior 5

camp

Fenbridge mouth: standing at it, then a wide shot from outside facing down the ramp:

exit

exit wide zone2

**The Sunken Road's own underground map**: opening the map window while inside the tunnel now shows this dedicated dark schematic (centerline, both mouth labels, ridge crossing) instead of trying to render a straight-down corridor on the surface map:

underground map

Full world map, Eastbrook Vale: "The Sunken Road" POI on the west side, Mirror Lake now a clean single circle (no double-ring):

map zone1

Full world map, Mirefen Marsh: "The Sunken Road" POI on the southwest side, Deepfen Shallows and the other lakes also clean circles:

map zone2

## Test plan

- [x] `npx tsc --noEmit`
- [x] Full `vitest run` green (pre-existing environment flakes unrelated to
      this change confirmed by re-running in isolation and against the
      unmodified base commit: a live-Postgres-only characterization test, a
      known timing-sensitive fishing test, and heavy-suite contention flakes
      under high parallel worker counts, matching this repo's own documented
      "an unbounded run flakes heavy suites under core contention" caveat)
- [x] `tests/sunken_road_terrain.test.ts`, `tests/terrain_walls.test.ts`,
      `tests/progression.test.ts`, `tests/fixes.test.ts` all green against
      the new mouth ramps
- [x] `tests/architecture.test.ts` (sim purity + pure-core registration for
      the new `sunken_road_map_view.ts`)
- [x] `tests/map_terrain.test.ts`, `tests/map_window_view.test.ts`,
      `tests/water_terrain_awareness.test.ts`,
      `tests/sunken_road_map_view.test.ts` (new), `tests/hud_perf_budget.test.ts`
      all green for the map fixes + new underground-map mode
- [x] `UPDATE_PARITY=1 npx vitest run tests/parity/parity.test.ts` (the mouth
      ramps shift decoration placement near them, which shifts the shared
      rng draw stream, same as any other terrain-shape change; regenerated
      again after round 2's narrower ramp + biome paint, both re-run clean)
- [x] Round 2: same suite re-verified green (`tsc`, the map/terrain/parity
      tests above, `tests/corpse_harvest_sim.test.ts`) after the biome-paint,
      narrower-ramp, shell-extension, and denser-lighting changes
- [x] `npm run i18n:gen`
- [x] 20 in-game screenshots verifying every fix above, visually inspected


FernandoX7 (commented):
This is a well-structured content drop. The carve-as-terrain approach (a chain of overlapping level HeightStamps, no bespoke wall/ceiling geometry) is the right call, and the wiring is careful in the places that matter: camps, quests, mobs, and gather nodes are appended at the end of their arrays so existing world-gen RNG draws stay byte-identical, `findSafePos` is deterministic so the mid-list NPC insertion only renumbers ids, and the render side is a clean own-module static group (`cave_tunnel.ts`) mirroring `gather_nodes.ts` with no per-frame cost. `sim/content/sunken_road.ts` imports only from `../types`, so sim purity holds.

I verified the load-bearing claims against the code at head:

- Parity goldens are an honest regeneration: `tests/parity/parity.test.ts` passes, and the churn is the deterministic +20 entity-id shift (19 camp mobs + Foreman Delke) plus the RNG cascade in the AI/combat scenarios, not hand-edited values. Isolated scenarios (e.g. `quest_kill_credit`) shifted ids only.
- The three modified tests are principled isolation, not tolerance-widening: `isInWaterBody` guard in `druid_spell_pack` (deep dry terrain is no longer water under the terrain-aware model), `moveSpeed = 0` on the talent punching bag, killing trash to isolate the Litany pulse. The delve seed 42 -> 62 swap is cosmetic (the fixtures pin `run.bountiful = true` explicitly).
- `tsc --noEmit` clean, S3 guard green, `i18n:gen` / `i18n:hash --write` / `wiki:content` produce no drift, `i18n.resolved.sha256` is a single line, English item name is in `itemNamesEn` and the mob/npc/quest ids are registered in `world_entity_i18n.ts`.

Two findings, both at the ridge crossing.

**should-fix** `src/sim/content/sunken_road.ts` waypoints (135,155) -> (115,180) -> (90,205): the tunnel is passable, but the intended centerline is gated above the climb limit right where it crosses the ridge. Sampling `terrainSteepnessAt` (the exact function the movement gate uses, sim.ts:3172/3241/3346: entry is blocked when `terrainSteepnessAt(nx,nz) > PLAYER_MAX_CLIMB_SLOPE` = 1.5) along the whole centerline, 13 of 341 samples exceed 1.5, all between z=164 and z=195, peaking at 2.81 near (103,192). A flood-fill of the ridge region does connect both sides, so a player can weave/pathfind through, but a straight walk along the road is repeatedly deflected at exactly the ridge the tunnel exists to smooth out. Root cause: the natural ridge is about +36 at (115,180), and with radius-28 stamps spaced ~32 to 35 yd apart under a `smooth` falloff, the carved floor undulates from -14 at each waypoint up to about -1.2 to -1.8 at the ridge midpoints, so the along-corridor grade climbs past 1.5. Adding intermediate waypoints (roughly halving the spacing) and/or raising `STAMP_RADIUS` across z=155 to 205 would keep the floor deep and drop the grade under the limit.

**nit** `tests/sunken_road_terrain.test.ts` "stays walkable (non-swim)": this case is effectively vacuous, which is why the gating above slips through. For a dry point `waterLevelAt` returns `-Infinity`, so `floor = -Infinity - PLAYER_SWIM_DEPTH` is `-Infinity` and `h >= -Infinity` is always true regardless of the terrain. It asserts nothing about the climb-slope gate that actually governs walkability here. Adding `expect(terrainSteepnessAt(x, z, SEED)).toBeLessThanOrEqual(PLAYER_MAX_CLIMB_SLOPE)` along the centerline would make it real (and it currently fails at the 13 ridge cells, which is the point of the should-fix).

Merge note: the conflicts are all mechanical. The only conflicting files are the generated i18n trio (`i18n.resolved.*`, `i18n.status.summary.json`), the locale overlays, `world_entity_i18n.ts`, and the parity goldens, all of which base also regenerated (the professions merges and the terrain-water merge moved them). Merge `release/v0.22.0`, re-run the i18n build (`i18n:build` / `i18n:gen` / `i18n:hash --write`), and regenerate the goldens with `UPDATE_PARITY=1` rather than hand-resolving them, then re-run `parity.test.ts` to confirm the two id shifts compose cleanly.

jgyy:
Closing this in favor of a different direction.

This tunnel works, and it's a legitimate, well-executed solution within the current terrain model (narrow `'flat'`-falloff HeightStamp corridor + an oversized visual shell mesh swept over it so the sky is never visible). But that's exactly the limitation: the underlying terrain is still a 2.5D heightfield (one Y value per x,z column, in `src/sim/world.ts`'s `terrainHeight`/`groundHeight`, shared by sim collision and the render mesh in `src/render/terrain.ts`). The "enclosure" here is a rendering illusion, not real geometry: the ground truth has no ceiling, no wall thickness, no volume, it's a single continuous surface that's just narrow at this point, with a separate mesh laid over it to fake the tunnel look.

We want to move to a true 3D/voxel (or SDF + marching-cubes) terrain representation that supports multiple Y surfaces per column, i.e. a real floor and a real ceiling with actual solid rock in between and around, so tunnels (and overhangs, caves, etc.) are genuine carved volumes rather than a narrow heightfield corridor with a matching shell mesh glued on top. That's a foundational rewrite of the terrain data model, collision (`src/sim/colliders.ts` is 2D circles/OBBs today, no vertical extent), and the terrain mesher, not something this PR's approach can evolve into incrementally.

Closing so future underground features build on the voxel model instead of extending the heightfield-illusion pattern further. No fault in this PR's execution, it's a scope/direction call.

PR #1530, opened 2026-07-06
fix(hud): guard entry HUD-ctor id parity (catches a live #tf-discord gap on /play)

BlizzHacker:
## Problem

The `Hud` class resolves many elements as **constructor-time field initializers** via `$('#id')` (which is `document.querySelector` cast to a non-null type). If an entry HTML is missing one of those ids, the field is silently `null`, and the first code that touches it — `new QuestProgressBanner($('#quest-banner'))`, `el.addEventListener(...)` — throws during `new Hud()`, taking the renderer down on that entry.

`entry_window_parity.test.ts` guards `.window panel` ids, but the crash class is broader: banners, bars, and indicators the ctor resolves are **not** window panels, so that whole surface was unguarded.

## What this catches (a real one, today)

The new guard immediately found a live gap on `main`: **`#tf-discord`** (the target-frame `targetDiscordEl` field) is in `index.html` but **missing from `play.html``** — so on `/play` that field resolves to `null`. This PR restores the identical `<div class="uf-discord" id="tf-discord">` to `play.html` in the same position.

## The guard

`tests/entry_hud_ctor_parity.test.ts` scrapes the ctor field-initializer `$('#id')` literals from `hud.ts` (only the field region, before the constructor — method-body lookups are excluded since those elements are often injected on demand) and asserts every id exists in **both** `index.html` and `play.html`. Adding a HUD element to one entry but not the other now fails CI instead of the browser.

## Verification

- New test **fails** with `missingInPlay: ["tf-discord"]` before the fix, **passes** after (red→green).
- Existing `entry_window_parity.test.ts` still green.

### DIFF
--- a/play.html
+++ b/play.html
@@ -186,6 +186,7 @@
         <div class="uf-name" id="tf-name"></div>
         <div class="bar hp"><div class="bar-fill" id="tf-hp"></div><div class="bar-absorb" id="tf-absorb"></div><div class="bar-text" id="tf-hp-text"></div></div>
         <div class="bar" id="tf-resource"><div class="bar-fill" id="tf-res"></div><div class="bar-text" id="tf-res-text"></div></div>
+        <div class="uf-discord" id="tf-discord"></div>
         <div id="tf-castbar" role="progressbar" aria-valuemin="0" aria-valuemax="100" data-i18n-aria="hudChrome.castBar.targetAria" aria-label="Unit Cast Bar"><div class="fill"></div><div class="label"></div><div class="timer"></div></div>
       </div>
       <div class="uf-buffs" id="tf-debuffs"></div>
--- a/tests/entry_hud_ctor_parity.test.ts
+++ b/tests/entry_hud_ctor_parity.test.ts
@@ -1,0 +1,50 @@
+import { readFileSync } from 'node:fs';
+import { describe, expect, it } from 'vitest';
+
+// Both build entries (index.html at / and play.html at /play) hand-carry the HUD
+// chrome. The Hud class resolves many elements as CONSTRUCTOR-TIME field initializers
+// via `$('#id')` (= document.querySelector, cast to a non-null type). If an entry is
+// missing one of those ids, the field is silently null and the first thing that touches
+// it — `new QuestProgressBanner($('#quest-banner'))`, `el.addEventListener(...)` — throws
+// during `new Hud()`, taking the whole renderer down on that entry.
+//
+// entry_window_parity.test.ts guards `.window panel` ids, but the crash class is broader:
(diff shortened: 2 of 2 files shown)


Rubsey (commented):
## OSSBrain initial review

### Should fix

- **Remove banned dash characters from new test comments** (`tests/entry\_hud\_ctor\_parity.test.ts:8`)
  - Issue: The new test file introduces em dash characters in comments on lines 8 and 20. The trusted root CLAUDE.md says no em dashes or en dashes anywhere, including code and comments, and the repo hooks enforce that rule.
  - Impact: The change can be rejected by the contribution floor even though the functional HUD parity fix is small and otherwise well scoped.
  - Suggested fix: Replace the em dashes with ASCII punctuation, for example commas, colons, or parentheses, in both new comment lines.

### Checks

- `sha256sum .ossbrain-review/pr-1530/woc-review-pr.SKILL.md`: **passed**. Matched supplied skill SHA-256 b958f04c641fb33b3a96cee949016a9f7a6fcc0f497ed2b12823216880a127fa.
- `git cat-file -t bae6708ffbeae18f1b5e8da70d3d801405bc4d57 && git cat-file -t b8d612da3f4d9c217eb9c2980efb98711a7f5e27 && git cat-file -t 376f125b3ad55b3b8a73b6500310200835835af1`: **passed**. Confirmed supplied PR head, integrated review head, and release base are local commits.
- `git diff --name-status 376f125b3ad55b3b8a73b6500310200835835af1..b8d612da3f4d9c217eb9c2980efb98711a7f5e27`: **passed**. Integrated payload is limited to play.html and tests/entry\_hud\_ctor\_parity.test.ts.
- `rg -n '\[\\u2014\\u2013\]' play.html tests/entry\_hud\_ctor\_parity.test.ts`: **failed**. Found banned dash characters in the newly added test comments at tests/entry\_hud\_ctor\_parity.test.ts:8 and tests/entry\_hud\_ctor\_parity.test.ts:20.
- `git diff --check 376f125b3ad55b3b8a73b6500310200835835af1..b8d612da3f4d9c217eb9c2980efb98711a7f5e27 -- play.html tests/entry\_hud\_ctor\_parity.test.ts`: **passed**. No whitespace errors in the integrated diff.
- `npx vitest run tests/entry\_hud\_ctor\_parity.test.ts`: **not run**. Vitest could not start in this sandbox because Vite tried to write temp/cache files under read-only node\_modules.
- `npx vitest run --configLoader runner tests/entry\_hud\_ctor\_parity.test.ts`: **not run**. The runner config loader avoided the first temp file but the Svelte Vite plugin still attempted to write node\_modules/.vite metadata, which is read-only here.

PR #1532, opened 2026-07-06
fix(net): deliberate logout skips linkdead grace

Pepijnvdliefvoort:
## Summary

Since v0.22.0 the in-game **Log Out** button left the character held in-world
for 5 minutes instead of removing it cleanly. On re-login, the character
appeared as **online** in the character list, showing a "Take Over" prompt and
the hint *"Already in world. Log out elsewhere, or take over."* instead of the
normal **Enter World** button.

**Root cause:** The linkdead grace feature (v0.22.0) was designed for accidental
disconnects. A deliberate logout was not handled: `logout: () =>
location.reload()` just reloaded the page without signalling the server, so the
server treated the resulting socket close as an accidental drop and entered the
5-minute grace.

**Fix:** Before the reload, the client sends `{ t: 'logout' }` over the
WebSocket. The server handles this in `dispatchMessage()` by calling `leave()`
immediately, setting `session.left = true` synchronously, so the subsequent
page-unload socket close is a no-op in `socketClosed()` and linkdead grace is
never entered. `ClientWorld.sendLogout()` also sets `sessionEnded = true` to
prevent the auto-reconnect from firing.

## Related issues

Closes #

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/linkdead.test.ts` (27 tests pass, including 2
  new regression tests covering the deliberate-logout path)
- Manual steps:
  1. Start server + client (`npm run server` / `npm run dev`)
  2. Log in, pick a character, enter the world
  3. Open Options → Log Out
  4. Log back in and open the character list
  5. **Before fix:** character shows "Already in world" hint + Take Over button
  6. **After fix:** character shows the normal Enter World button; enters
     immediately with no confirmation dialog

---

### Quality

- [x] **Tests pass.** `npm test` is green. Added 2 regression tests in
      `tests/linkdead.test.ts` covering the `t:'logout'` message path.
- [ ] **Builds succeed.**

### Cross-platform

- ~Tested on desktop and mobile.~ Not a UI layout change.
- ~Accessible.~ Not a UI change.

### Localization (i18n)

### Hygiene

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -2464,6 +2464,13 @@ export class GameServer {
     const msg = rawMsg as ClientMessage;
     const sim = this.sim;
     const pid = session.pid;
+    // Deliberate logout: the client wants a clean leave, not a linkdead grace.
+    // Calling leave() immediately sets session.left = true, so the subsequent
+    // WebSocket close event (from the page reload) is a no-op in socketClosed().
+    if (msg.t === 'logout') {
+      void this.leave(session, 'logout');
+      return;
+    }
     if (msg.t === 'input') {
       if (session.spectating) return;
       const meta = sim.meta(pid);
--- a/src/main.ts
+++ b/src/main.ts
@@ -1545,7 +1545,12 @@ async function startGame(
   // the options menu drives logout + key-capture + settings, all of which need
   // refs that only exist now (input/renderer) or are page-level (reload)
   hud.attachOptions({
-    logout: () => location.reload(),
+    logout: () => {
+      // Signal the server to leave immediately, skipping the linkdead grace, so
+      // the character is not held in-world after a deliberate logout.
+      online?.sendLogout();
+      location.reload();
+    },
     captureKey: (cb) => input.captureNextKey(cb),
     settings,
     onSettingChange: (key, value) => applySetting(key, value),
--- a/src/net/online.ts
+++ b/src/net/online.ts
@@ -1141,6 +1141,19 @@ export class ClientWorld implements IWorld {
     this.ws.close();
   }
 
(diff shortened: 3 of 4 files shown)


FernandoX7 (commented):
Clean, well-scoped fix and the root-cause description is accurate. The load-bearing claim checks out: `leave()` sets `session.left = true` synchronously before its first `await` (server/game.ts:1866), so the `void this.leave(session, 'logout')` call returns with `left` already true, and `socketClosed()` short-circuits on it (server/game.ts:1791). Routing logout as a top-level `t:` message (like `input`) rather than a `cmd` keeps it clear of the `command_schema` lockstep gate, which is correct, and no `Entity` field is added so there is no wire or `blankEntity` parity concern. I ran `tests/linkdead.test.ts` at the head (27 pass) and mutated the handler out to confirm the two new tests are decisive (both fail without it).

Two nits:

- nit: `src/net/online.ts:1152` uses the magic number `1` for the open state, while the rest of the file uses `WebSocket.OPEN` (online.ts:1216, 1254). Prefer the named constant.
- nit: the teardown triad (`sessionEnded = true; clearInterval(this.sendTimer); if (this.reconnectTimer !== undefined) clearTimeout(this.reconnectTimer)`) is now the third full copy, alongside `close()` and the error-frame handler. A small private helper (e.g. `endSession()`) would fold all three.

Not a finding, just noting: the fix relies on the `t:'logout'` frame flushing before `location.reload()` tears down the socket. That is the standard send-then-reload pattern, it degrades gracefully (a dropped frame just falls back to the old linkdead path), and your manual test confirms it works, so this is fine as is.

FernandoX7 (approved):
Nice catch and fix. Thank you.

PR #1525, opened 2026-07-06
feat(game): finish mobile action ring HUD wiring

Rubsey:
Reopened follow-up for the mobile UI work because mobile release testing found bugs. GitHub cannot reopen merged PRs, so this draft PR carries the reverted mobile work on a follow-up branch for continued work tomorrow.\n\nMoved out of release/v0.22.0:\n- #1440 feat(mobile): declutter the touch HUD chrome\n- #1483 feat(game): finish mobile action ring HUD wiring\n- 52e3ff7df fix(mobile): restore viewport scale after orientation changes\n- Reverted orphaned mobile hotbar locale overlays on the release branch\n\nRelease branch cleanup commits now on release/v0.22.0:\n- 2b6519497 revert #1483\n- 2e2a5ab46 revert #1440\n- d40771ff6 revert viewport orientation fix\n- 75ec20245 remove reverted mobile hotbar overlays\n\nValidation:\n- release/v0.22.0: npm run gate PASS (all 9 steps)\n- feature/new-mobile-controls-followup: i18n overlay/gap-fill tests PASS\n\nKept as draft while the mobile bugs are worked through.

PR #1483, opened 2026-07-05
feat(game): finish mobile action ring HUD wiring

MasterZensei:
## Summary
- Wires up touch-only option rows and finishes the mobile action ring HUD.
- Adds the associated i18n strings for the mobile action ring chrome and regenerates the resolved i18n tables.

## Test plan
- [x] `npx vitest run tests/touch_router.test.ts tests/mobile_action_ring_painter.test.ts tests/localization_coverage.test.ts`
- [ ] Manual mobile viewport check of the action ring in the browser

PR #1533, opened 2026-07-06
fix(game): mobile touch reliability fixes and HUD layout rework for #1525

MasterZensei:
Stacked onto #1525 (the reapplied mobile action ring work): this lands the fixes for the bug class that got #1483 reverted from release/v0.22.0, plus the layout rework from mobile playtesting.

## Touch input fixes

- **Buttons dead while steering.** Browsers only synthesize `click` for the PRIMARY pointer, so every click-bound button (ring abilities, More, Target, Use, Autorun, haptics, More-close) did nothing while the other thumb held the joystick, which is how the game is actually played. New `src/ui/touch_tap.ts` binds taps on raw pointer events for any touch pointer, keeps `click` as the keyboard/mouse path with double-fire suppression, and cancels on a 12px slide-off.
- **Joystick drift on touchdown (issue #1229).** The floating wheel's origin was clamped into the 132px capture zone, which cannot contain the 140/128px wheel, so every off-center touchdown instantly walked the character before any drag. The origin is now the raw touch point; the wheel spawns centered under the thumb.
- **Dead touch sliver.** The wheel overhangs the capture zone and had no listeners of its own; its outer arc swallowed touches. It now carries the same pointer lifecycle.
- **Joystick Size desync.** Input throw uses the rendered (scaled) radius, placement the layout radius, and the floating wheel scales about its centre, so it tracks the thumb at every scale setting.
- **Walking blind behind modals.** Movement now aborts when a window opens mid-drag, matching the rule swipe-look already had.
- **Ring polish.** Long-press inspects a spell (tooltip wiring arms the peek guard) instead of casting it; casts flash on the ring; `:active` press feedback; the square cooldown sweep clips to the circular buttons.

## Layout rework

- Chat / Social (promoted out of the More tray) / More sit top-left, clear of the target-frame seat; the More dialog opens centred on screen.
- Jump moved into the ring's bottom row on Use's row, spaced so its gap to the 180deg slot equals Use's gap (right thumb jumps while the left steers); Autorun keeps the joystick-side hollow seat.
- Castbar re-seated centred above the bottom-centre player frame; both nudge 40px left on the compact tier so Jump keeps a clear circle on 740px phones.
- The community-links rail is hidden on mobile touch (its two-person icon read as a second Friends button).

## Verification

Full vitest suite (836/838 files; the one failure is a pre-existing local-Windows spawnSync quirk in new_endpoint.test.ts, identical on the pristine base), tsc, biome on changed files, the production build, the geometric layout gate on all 7 device profiles plus left-handed / camera-joystick / size-extreme / safe-area passes, two new browser probes (scripts/_probe_mobile_drift.mjs, scripts/_probe_layout_visual.mjs) covering off-center touchdown, second-finger taps, and modal flows, new unit tests (touch_tap, 3 joystick integration tests), and the qa-checklist review (READY, zero blocking findings). Also hand-tested on a real phone over LAN.


FernandoX7 (commented):
Solid follow-up, and the touch-reliability diagnosis is right: browsers only synthesize `click` for the primary pointer, so binding the touch path on raw pointer events (`bindTouchTap`, with `click` kept for keyboard/mouse plus a suppression window against double-fire) is the correct fix, and it keeps `<button>` keyboard activation intact. I checked the load-bearing pieces against the code rather than the description:

- Origin-is-the-touch-point: `onMoveDown` sets `moveOrigin` to `e.clientX/Y`, so a touchdown alone can't produce intent (the #1229 drift). The wheel + zone double-binding is safe because `onMoveDown` guards on `joyPointer !== null` and `classifyTouch`, and `getTouchOwner` returns `menu` when `mobile-window-open` is set, so a touchdown while a window is up never starts a drag.
- The Joystick Size split holds: `--joy-scale` is on the at-rest `.mobile-joystick` rule, so `getBoundingClientRect().width` at press time is the scaled radius (the input throw) while `offsetWidth` stays the layout radius (the stick translate).
- Mid-drag abort: `mobile-window-open` is toggled in hud.ts:1809 and `onMoveMove` now releases on it, matching the camera path.
- Ring polish: `sourceSlotForMobileButton` starts at slot 1, so the `barSlot === 0` early return in `flashActionSlot` is exactly the attack toggle; `attachTooltip` arming the peek guard is what turns a long press into inspect-not-cast.
- The jump / social relocations are mirrored in both index.html and play.html, and the tooltip strings are English `t()` keys already in the catalog.

Ran tests/touch_tap.test.ts, tests/mobile_controls.test.ts and tests/client_shell.test.ts at the head: 131 pass.

Findings:

- nit (src/styles/hud.mobile.css:865, :1028): two comments say the trio sits at top-centre ("the More button lives at top-centre now"; "with Chat/More moved to top-centre the band over the player frame is free again"), but the cluster is anchored top-left (`left: max(12px, env(safe-area-inset-left))` at :289), which is what the PR body and the tests say. Line 296 correctly describes it as OFF the top-centre band. Worth aligning 865/1028 so a future reader is not sent to the wrong corner.
- nit (scripts/mobile_cluster_layout_check.mjs:366): the joystick-overlap loop still includes `mobile-jump`, but Jump now lives in the ring bottom-right, far from the pad, so that arm is always-pass. Harmless (the autorun arm still does real work); drop `mobile-jump` from the pair if you want the check to assert what it reads as.

Nothing blocking.

FernandoX7 (commented):
Re-reviewed at 885449477. The three commits since my last pass (the More-tray Discord/Donate entries, the native-build Donate suppression, and the Game Menu back control) hold up. I checked the load-bearing pieces against the code:

- **play.html parity.** `#mobile-discord` and `#mobile-donate` are byte-identical in both shells, and the relocated `#discord-window` + `#discord-keep-modal` are top-level siblings declared before `#start-screen` in both. The Discord panel CSS moved cleanly out of the unlayered `index.extra.css` into `@layer components` in `components.css` (the whole block lifted, nothing orphaned behind).
- **Native build.** `#mobile-donate` joined the `body.native-app` donation-suppression block in `hud.css:2499` (`display: none !important`), so it will not ship in the one build where donations are deliberately hidden. Discord stays behind `DISCORD_BUILD_ENABLED`, and `openDiscordEntry` falls back to the community invite when logged out or offline, so the entry is no longer dead pre-login.
- **URLs.** The invite (`https://discord.gg/GjhnUsBtw`) and the sponsors URL match the shells' existing hardcoded links; no new i18n keys (reuses `nav.donate` / `a11y.donateProject` / `hudChrome.discord.title`).
- **Back control.** The central `[data-back]` sweep in `render()` wires exactly one listener per view (the main menu renders none, so it no-ops), and the self-rendering Performance panel deliberately omits `data-back` and wires its own to avoid a double-bind on its internal rerender. `goBack()` clears a pending keybind capture and moves focus off the control the re-render destroys. `hud.options.back` and `svgIcon('prev')` both exist.

Ran `tests/touch_tap`, `tests/mobile_controls`, `tests/client_shell`, `tests/options_window` at the head: 151 pass.

The two nits from my earlier pass are still open (both harmless, your call):

- nit (src/styles/hud.mobile.css:865, :1027): the two comments still say the trio sits at top-centre; it is anchored top-left (:289). Comments only, worth aligning so a future reader is not sent to the wrong corner.
- nit (scripts/mobile_cluster_layout_check.mjs:349, :371): the joystick-overlap loops still include `mobile-jump`, which now lives in the ring bottom-right, so that arm is always-pass. Drop it if you want the check to assert what it reads as.

Nothing blocking.

PR #1534, opened 2026-07-06
Online movement feel: hide the network latency without touching server authority

madmatah:
# Online movement feel: hide the network latency without touching server authority

**Draft for team testing.** 

Ever since the performance optimizations on the display side went live, the game is much more fluid. However, it really highlights the latency now: there is a massive difference between playing on localhost and the prod server. This PR addresses this issue.

:warning: Important notes:
- This PR is meant to be played, not just reviewed. Because everything here comes down to how it feels. I've included commands to simulate latency below.
- Everything here is display-side or input-side; the server simulates movement exactly as before, byte for byte.

## The problem

Offline, a movement key moves your character in 30 to 50 ms. Online it takes
RTT + 45 to 75 ms, because the client only sends movement intent and waits for the
authoritative snapshot to come back before anything moves on screen: uplink, the
next 20 Hz server tick, the snapshot, and one interpolation interval. At 50 ms ping
that is 100 to 120 ms input-to-motion; at 100 ms ping, 150 to 170 ms. The gap is
structural, so no amount of tuning the old pipeline could remove it. Full analysis
with the per-stage latency budget: `docs/online-movement-latency.md`.

## What this branch does (final state, vs the version you know)

**One shared movement kernel.** The body of `Sim.updatePlayerMovement` (turn
integration, wish vector, slope gates, swept static collision, jump/gravity/fall
pass) moved verbatim into `src/sim/player_motion.ts` behind a small deps struct.
The offline browser world, the authoritative server, and the RL env run it exactly
as before; `tests/player_motion.test.ts` pins bit-for-bit parity between the live
Sim and the client-shaped deps, so the paths cannot silently fork.

**Display-only position extrapolation for the local player**
(`src/render/self_motion.ts`). While you hold movement intent online, the renderer
advances your displayed pose every frame through that same kernel: real run speed,
real slope gates, real static collision, real jump arc. Starts, stops, jumps, and
direction changes now respond the frame the key changes. Safety properties, in
order:

- Anchored: every frame the authoritative pose is compared against where the
  display was one measured echo ago; any genuine divergence (a stun landing
  mid-press, server-driven motion like charge or knockback) glides back, never
  drifts. The correction servo is aligned to the true echo and gain-bounded so it
  cannot ring, even at very high RTT.
- Bounded: the horizontal display lead is hard-leashed to what you could
  legitimately run within the latency window (capped at 350 ms); server teleports
  and life-state changes snap outright.
- Invisible to logic: the output feeds only the renderer's self position (mesh and
  camera). Targeting, range checks, interest, quest triggers, the entity mirror,
  and everything sent to the server keep using authoritative positions.

It is gated off while spectating, corpse-frozen, CC'd, or inside a delve, and
`?nopredict` in the URL restores today's behavior at runtime.

**Keyboard turning is now a facing input, like the mouse.** The client integrates
A/D turns locally at the same TURN_SPEED and streams the resulting heading on the
existing facing channel, the one mouselook has always driven and the server has
always applied outright; the turn flags are zeroed on the wire (except the engage
edge, so `/follow` still breaks and the anti-AFK gate still sees you). The model
and camera respond the same frame, and there is no client/server heading
disagreement to reconcile afterward, which is where every release stutter came
from.

**Feel fixes this work made visible (all pre-existing bugs).** The run/walk pick
had no hysteresis and crossfaded Running/Walking on every noisy frame under load
hitches (the world-entry animation glitch); the airborne heuristic could flash the
jump clip on a single bad frame; the self model yaw extrapolated past snapshots
and oscillated. All three are fixed with proper hysteresis/debounce in the pure,
tested modules (`src/render/locomotion.ts`, `anim_state.ts`).

**Instrumentation.** The perf overlay (`?perf`) gains a Prediction Lead row (how
much latency the extrapolation is currently hiding, in ms).

**What did NOT change:** the wire protocol, IWorld, snapshots, SimEvents, the RL
env surface, other entities' display, and every combat/loot/outcome path. No new
dependency.

## The "never predict" rule: what we relaxed and why it is safe

`src/net/CLAUDE.md` used to ban all client-side prediction. The ban existed to
protect three things, and all three still hold:

1. **Outcome prediction is still forbidden.** No client-side anticipation of
   combat, casts, resources, loot, or aggro. Unchanged.
2. **Server authority is untouched.** The server simulates movement from intent
   exactly as before and never trusts a client position. The extrapolation is a
   lens on the local screen, not a simulation input; the four constraints above
   (bounded, blending, invisible to logic, and, for the position layer, never
   sent) are written into the rule and covered by tests.
3. **The one authority change is the heading, and it is not new authority.** The
   facing channel has been client-driven since mouselook exists (the server
   applies a sent facing outright, corpse-guard only). Keyboard turns now use the
   same channel with the same limits. A stunned player could always mouse-turn;
   now keyboard parity matches.

Open questions worth the team's eyes:

- The 350 ms extrapolation ceiling: below the real RTT the steering feels gluey
  (leash-bound), above it a genuine misprediction glides longer. 350 covers every
  link the game should feel good on; push back if you want it lower.
- The Fiesta move-speed augment is not mirrored, so the predictor runs without it
  during bouts (the leash absorbs the bounded divergence). Acceptable, or mirror it?
- Delves are gated off (portcullis door clamps are not mirrored). Follow-up
  candidate, not v1.

## How to test

Run both sides locally (`npm run server`, `npm run dev`), create an online
character, then shape the connection to simulate a real link (per-OS recipes
below). Whatever method you use, **verify it took effect with the Ping row of the
`?perf` overlay** (it should read roughly twice your configured one-way delay);
loopback shaping silently not applying is the classic trap. A/B against the old
behavior at any time by adding `?nopredict` to the URL, and watch the
Ping / Jitter / Prediction Lead rows while you play.

### Linux

```
# add 70ms delay with 20ms jitter on loopback (note: loopback traffic crosses
# the qdisc in both directions, so this is roughly a 140-280ms RTT)
sudo tc qdisc add dev lo root netem delay 70ms 20ms

# change it live
sudo tc qdisc change dev lo root netem delay 40ms 10ms

# ALWAYS remove it when done
sudo tc qdisc del dev lo root
```

### macOS

The reliable route is Apple's **Network Link Conditioner** (the tool iOS devs use):

1. Download "Additional Tools for Xcode" from developer.apple.com/download/all
   (pick the release matching your Xcode), open the `Hardware` folder, and
   double-click `Network Link Conditioner.prefPane` to install it into System
   Settings.
2. Create a custom profile with 70 ms downlink delay + 70 ms uplink delay
   (~140 ms RTT), enable it, play, disable it when done.
3. It is system-wide (your Slack will feel it too), and on some macOS versions it
   does not shape loopback: check the Ping row. If it stays flat, either run
   `npm run server` on a second machine on the LAN and point the client at it, or
   fall back to the options below.

CLI alternative (dummynet via `dnctl`/`pfctl`; works on most macOS versions, but
verify with the Ping row):

```
# 70ms each way on the game server port, both directions
sudo dnctl pipe 1 config delay 70
echo "dummynet in quick proto tcp from any to any port 8787 pipe 1
dummynet out quick proto tcp from any to any port 8787 pipe 1" | sudo pfctl -q -f -
sudo pfctl -E

# cleanup: restore the default ruleset, disable pf, flush the pipe
sudo pfctl -f /etc/pf.conf
sudo pfctl -d
sudo dnctl -q flush
```

Zero-install fallback: Chrome DevTools > Network > custom throttling profile with
added latency. Recent Chrome applies it to WebSockets too; confirm with the Ping
row, and note it only shapes the browser side (fine for this test).

Feel checklist (each of these was a real bug at some point during development):

- Run start/stop: instant response, no backslide, no glide after stopping.
- Keyboard turns while running: model and camera turn the same frame; releasing
  the key produces no re-aim, wobble, or residual rotation.
- Let go of ALL keys after turning: the character must stay put (a feedback bug
  once made it spin on its own under netem).
- Steering while running at high simulated RTT: course corrections must bite
  immediately, not feel like moving through water.
- Jump over a fence mid-run; walk into a building wall (no jitter).
- Click-to-move around A* corners: the yaw follows the path, no sideways slide.
- `/follow` someone, then press A/D: follow must break.
- Get stunned mid-run: the display glides back to where the server held you.
- Die, release spirit, ghost-run; enter a delve (predictor off, no visual pop);
  spectate (moderators): predictor off.
- World entry on a weak machine (or with devtools CPU throttling): the character
  animation must not stutter-replay during the loading seconds.

Unit coverage runs the same scenarios against a real lagging Sim in CI
(`tests/self_motion.test.ts` drives 100 to 500 ms RTTs, `tests/player_motion.test.ts`
pins kernel parity, `tests/keyboard_turn_facing.test.ts` includes a delayed-echo
resonance harness). Full suite, build, and the architecture/parity/i18n gates are
green.


EnriqueGF (commented):
Went through the whole branch with the full invariant pass plus a played-level verification of the claims. This is excellent work: the extraction is verbatim, the safety story holds up in the code exactly as described, and the netem-driven tuning notes in the comments are the kind of documentation this repo rarely gets.

What I verified:

- Kernel extraction (`src/sim/player_motion.ts`): statement-for-statement faithful to the old `Sim.updatePlayerMovement` body, same branch order, same epsilons, same rng draw order (the one rng-reachable callee, fall-damage `dealDamage`, sits at the identical call site). Parity digest unchanged, `tests/player_motion.test.ts` green.
- Parity: no `server/` or `src/net/` code change, wire protocol and IWorld untouched. The display lead feeds only `selfRenderPosition` (mesh + camera); nothing logic-facing reads it. `wireFacing` only ever carries input-derived headings, so the mirror-chasing corrections never close a feedback loop through the server.
- Anti-AFK and stun/corpse guards are intact with the zeroed turn flags (`lastInputAt` refreshes on every input message).
- Full gate green at the head: 10230 tests across 860 files, tsc, changed-files biome, all five builds, malware scan.
- Fairness: the predictor is display-only, leashed, and gated off under CC/spectate/delve; `?nopredict` confers no advantage.

One real finding, low severity: the /follow-break engage edge fires only on a transition into keyboard turning. If the player is already holding A/D when /follow is issued, the server never sees a turn flag and the follow does not break until the key is re-pressed. There is no clean client-side fix (followTargetId is not mirrored, and /follow is parsed server-side from chat), and a periodic flag re-assert would reintroduce the overshoot wobble you spent commits killing, so I did not change the behavior.

I pushed one commit to the branch (df0a63453, maintainer edits are enabled, revert freely if unwanted): it moves the wire turn-flag gating out of `main.ts` (a firewall file in this repo's conventions) into `keyboard_turn_facing.ts` as `suppressTurnFlags`, semantics identical, and adds four tests pinning the edge behavior, including an explicit KNOWN LIMIT pin of the /follow corner so a future fix flips it deliberately.

Two optional notes, nit tier:

- `main.ts` allocates one `SelfMotionFrame` literal per online frame. Within repo norms, but a reusable scratch object would remove it if a GC pass ever flags the render loop.
- Worth confirming on a live server that cone/aim abilities resolve at the same time for a keyboard turner and a mouselook turner (expected parity by design, both stream facing per frame).

The branch currently conflicts with `release/v0.23.0` only on the generated i18n trio (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`): mechanical regen after merging base, not a design problem.

On the open questions: 350 ms ceiling feels right as the default, the Fiesta augment gap is fine for v1 (the leash absorbs it), and gating delves off is the correct call for now. Ready from my side pending the team's feel testing.

PR #1542, opened 2026-07-06
feat(talents): spec update, real signature spells and impactful specs for all 27 specs

ryan-foo:
## Summary

Spec update for Talents 2.0 (PR3 in the build plan, docs/prd/talents-2.0.md): picking a spec now grants a REAL spell and a mastery that matters.

- All 27 specs grant a real, grant-only signature ability on spec selection (24 new ability defs: Holy Shock, Flashfire, Vesper Form, Conflagrate, Swiftmend, Wild Fury, Killer's Calm, Knifestorm, Fiendshape, ...). The old fake signatures (baseline spells re-granted) return to plain kit abilities; Stormstrike and Chain Heal are demoted out of the shaman baseline kit to grant-only, test-asserted.
- Signature spell reworks so each feels real: Chain Heal is a true bounce heal (primary plus up to 2 injured-ally jumps, 40% falloff), Metamorphosis buffs the warlock and the demon (new pet_spellhaste) and renders as a larger fel-purple demon, Moonkin Form gets damage, armor, and a violet tint, Feral Instinct is a form-gated resource burst.
- One shared render form-tint feature (form to tint map) covers Shadowform, Moonkin, and Metamorphosis; remote players see the tints through the existing aura wire (kind is already encoded), zero server changes.
- Fixes found during hand-testing: casts ignored the spellHaste stat, fractional buff values rounded to 0 under a spell-damage mastery, multi-buff abilities evicted their own companion auras, and spec-granted abilities now announce (learnAbility) so they land on the action bar and spellbook.
- IP compliance (G0 ip_scrub gate): 15 signature spells initially shipped verbatim WoW names on the locked v0.23 NAME-MAP denylist; display names re-coined in the NAME-MAP voice with ids frozen (Aether Surge, Wild Fury, Knifestorm, Killer's Calm, Flashfire, Stormsurge, Primal Surge, Deep Cut, Radiant Aegis, Fiendshape, Owl Form, Fervor, Vesper Form, Life Leech, Coldsight Aura), plus Bear Form / Cat Form / Holy Light phrases scrubbed from descriptions. ip_scrub green.
- Cleanup: dropped an orphaned Sim.friendliesInRadius duplicate with a divergent grid iteration order (latent determinism trap).

Base: release/v0.23.0 (origin/release/v0.23.0 merged into the branch; conflicts were locale/generated files only, resolved by regeneration). Dialect hygiene: branch-added overlay entries in es_ES / fr_CA / en_CA that duplicated their base locale were dropped per the dialect divergence rule (the maintainer fills real dialect variants at release).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: npm run gate on the release-merged stack: PASS, all 9 steps green (full suite 9848 tests / 0 failures with bounded workers, tsc, all builds, malware scan, changed-files biome, release-tier i18n freshness). Plus targeted vitest runs over the talent suites and every guard test during development.
- Parity goldens: regenerated; RNG draw digests and per-checkpoint draw counts are byte-identical, only state hashes and the new announce events moved.
- Live hand-test: scripts/spec_hand_test.mjs drives the real offline client for all 9 classes and all 27 specs, asserting every signature resolves and casts with zero page errors (ALL PASS).
- Manual steps: hand-tested spec pick, signature cast, form visuals, and cast-time tooltips in the live offline client; screenshots below are captured by the checked-in scripts/spec_signatures_shot.mjs.
- Fresh architecture-reviewer and qa-checklist passes over the diff: 0 blocking findings.

## Screenshots / recordings

Committed under docs/screenshots/ (captured by the checked-in scripts/spec_signatures_shot.mjs, with the IP-safe names):

| Frost mastery panel | Warlock signature spellbook |
|---|---|
| spec-mastery-frost | signature-spellbook-warlock |

| Owl Form tint | Fiendshape demon |
|---|---|
| moonkin-form-visual | metamorphosis-visual |

---

### Quality

- [x] **Tests pass.**
- [x] **Typecheck passes.**
- [x] **Builds succeed.**

### Cross-platform

- [x] **Tested on desktop and mobile.** The talents window and spellbook are existing responsive surfaces; no layout changes beyond content.
- [x] **Accessible.** No new interactive UI; existing window semantics unchanged.

### Localization (i18n)

- [x] **New player-visible strings are translated.** 24 new ability defs plus reworked tooltips localized (catalog plus the five non-Latin locales per the M16 gate); localization_fixes and localization_coverage green.
- [x] Player-facing sim text re-localized at the client boundary in the same change.

### Hygiene

- [x] Generated files regenerated, not hand-edited (i18n resolved tables, parity goldens, wiki content).


FernandoX7 (commented):
Really nice pass, and the determinism discipline is the standout. Chain Heal draws its heal roll once and multiplies by the falloff per jump (no extra rng per hop), `nearestUnhealedFriendly` is a total-ordered no-rng scan over `entities.values()`, and you removed the orphaned `friendliesInRadius` copy that iterated a divergent grid order, so the parity goldens stayed byte-identical on draw counts. The multi-buff id-suffixing (companion buffs get a kind-suffixed id so Arcane Power / Metamorphosis stop evicting their own auras) and the fractional-rate rounding fix (values below 1 pass through `scaleEffect` instead of flooring to 0) are both real, well-reasoned fixes. Wire and shape parity is handled correctly too: `critDmgBonus` is added to both `baseEntity` and `blankEntity` as a server-local derived stat, `Entity.scale` already round-trips, and the form auras ride the existing `kind` wire so remote tints and the demon size need zero server change.

A few things to look at.

**should-fix**

- Metamorphosis (Fiendshape) silently grants +35% max HP. `form_metamorph` folds into `scaleMul` (`src/sim/entity.ts:266`), and `scaleMul > 1` also multiplies `maxHp` (`src/sim/entity.ts:381`, the `buff_scale` "also adds hp" path). Because `hpFrac` is preserved, popping the form is an instant ~+35% effective-HP jump and dropping it takes it back, so a 180s-cooldown demon form is also a survivability cooldown. The tooltip only mentions +20% spell damage / +20% cast speed / the demon buffs (metamorphosis def in `classes.ts`), so this is undocumented and untested. Either it is intended (say so in the tooltip and pin the intended maxHp) or the demon scale should be decoupled from the HP path.

- `power_infusion` (Fervor) tooltip contradicts its effect. The effect is `{ type: 'buffTarget', kind: 'buff_spellhaste', value: 0.2 }` (+20% spell haste), but the rendered tooltip says "increasing spell power by 28" (`src/ui/i18n.catalog/abilities.ts:1214`, mirrored in the classes.ts description). It is a new ability in this PR, so this is fresh drift. No test pins its effect kind, which is why the mismatch slipped through.

- The two headline mechanics can silently regress because their actual application is unpinned:
  - The crit-damage mastery (`critDmgPct` to `critDmgBonus`) is asserted only at the stat (`signature_reworks.test.ts:178/190`). The damage term `dmg *= (isSpell ? 1.5 : 2) + p.critDmgBonus` (`effect_dispatch.ts:172`, and the auto-attack / seal sites) has no test, so `+ p.critDmgBonus` could be deleted and every test stays green. Force a crit on a Fire mage and assert the hit scales by `(1.5 + bonus)/1.5`, plus the physical `2 + bonus` arm for a melee mastery.
  - Chain Heal's 40% falloff is unpinned: `signature_reworks.test.ts` only checks that 3 allies were healed, so `falloff` could be `1.0` and it stays green. Capture the per-jump `heal2.amount` and assert the decay (and, while there, one closer-to-caster injured ally so "nearest" is proven, and a full-HP ally in range to prove the injured-only skip).
  - `siphon_life`'s `leechPct` heal-back (`auras.ts:183`) has zero coverage: tick it once and assert the caster's hp rose by the tick damage.

**nit**

- Unused import: `effect_dispatch.ts:37` imports `hasCastShield` but never uses it (only `casting_lifecycle.ts` does). Biome flags it as a warning, so `ci:changed` will not fail on it, but it is leftover scaffolding.
- `aoeAllyHaste` is a dead effect branch: the handler exists in `effect_dispatch.ts` and it is mapped in `attack_on_ability.ts`, but no ability emits it (only `trueshot_aura` uses `aoeAllyAttackPower`). Wire it or drop it.
- Duplication worth folding while it is small: `petDamageMult` is defined twice (`sim.ts:3043` as a Sim method, `pet_ai.ts:215` as a module fn), and the `value > 1 ? value/100 : value` percent-normalize is copied four times (`pctValue` in `entity.ts` and `auras.ts`, inline in `sim.ts` and `pet_ai.ts`). The new `spell_combat.ts` is a natural home for a shared helper.
- `holy_shield` (Radiant Aegis) is described as "bouncing to 2 nearby enemies" but the effect is a direct hit plus an unbounded `aoeDamage` (radius 10), which hits every enemy in range, not a capped 2-target bounce.
- `noteSpellHit(..._args: unknown[])` is an inert stub called from ~5 combat sites. It is documented as a forward hook, so this is just a flag that it is dead in this branch, and the `unknown[]` signature drops type-checking at the call sites (a typed no-op would keep them honest). Minor and in the same area: the `consumeAura` `deal` branch draws `rng.range` before `rng.chance`, the reverse of the `directDamage`/`weaponStrike` sites (harmless since it is a new effect with no golden, just a convention mismatch).

ryan-foo:
Closed in favor of #1543, which now carries the complete spec update (this branch's signatures work plus the mastery identity pass) as one PR. Nothing here is lost: #1543's branch contains every commit from this one.

PR #1348, opened 2026-07-03
Talents 2.0: full epic + fun pass (single handoff PR, pick up and continue)

ryan-foo:
## DESIGN UPDATE 2026-07-07: partially superseded by #1543

The signatures and spec-mastery halves of this branch SHIPPED FRESH in
PR #1543 (feat(talents): Talents 2.0 spec update), built directly on
release/v0.23.0, gate-green (npm run gate, 9848 tests / 0 failures) and
live hand-tested across all 27 specs. Where this branch and #1543 disagree,
**#1543 is the design of record**:

- **The 24 spec signatures**: #1543 ships them with IP-safe display names
  per the locked NAME-MAP (ids frozen): Arcane Power=Aether Surge, Bestial
  Wrath=Wild Fury, Blade Flurry=Knifestorm, Cold Blood=Killer's Calm,
  Combustion=Flashfire, Elemental Mastery=Stormsurge, Feral
  Charge/Instinct=Primal Surge, Hemorrhage=Deep Cut, Holy Shield=Radiant
  Aegis, Metamorphosis=Fiendshape, Moonkin Form=Owl Form, Power
  Infusion=Fervor, Shadowform=Vesper Form, Siphon Life=Life Leech, Trueshot
  Aura=Coldsight Aura. This branch's verbatim WoW names FAIL the G0
  ip_scrub gate on the current base; any salvage must adopt the new names.
- **Signature mechanics**: #1543 also carries reworks this branch lacks:
  Chain Heal real bounce, Fiendshape buffing warlock + demon
  (pet_spellhaste) with the fel-demon visual, Primal Surge as a form-gated
  resource burst, Holy Shock dual-use (heal or damage), Fervor as 20% spell
  haste, plus spell-haste plumbing fixes (casts honored the stat, fractional
  buff values no longer round to 0, multi-buff abilities keep companion
  auras).
- **Spec masteries**: replaced wholesale by the spec identity pass (the
  owner-approved table in docs/prd/talents-2.0.md): six scoped fields
  (dotDmgPct, hotHealPct, absorbPct, meleeHastePct, petDmgPct,
  petDmgSharePct), critDmgPct extended to heal crits, the vs-rooted crit
  predicate widened to rooted-or-chilled, level scaling min(1, level/20)
  re-baked on every ding, and per-spec mechanic-differentiated values for
  all 27 specs. This branch's mastery numbers are obsolete.

## THE SALVAGE PLAN (phased, rebuild on the post-#1543 base)

[...]

PR #1543, opened 2026-07-06
feat(talents): Talents 2.0 spec update, real signatures, reworked spells, and identity masteries for all 27 specs

ryan-foo:
## The Talents 2.0 program (read this first)

This PR is **S1 of two**: the 27 real spec signatures + identity masteries. **#1614 (the flip) contains this branch and builds directly on it**: choice rows replace the trees, all 162 options, the proc engine, and the VFX/feel work all live there. #1348 (the original epic PR) is closed; its salvage ledger points here and to #1614.

Merge order: this PR is the reviewable S1 subset. Merging #1614 subsumes it (same commits); land whichever the release captain prefers, then close the other as contained.

## Summary

The complete Talents 2.0 "spec update" (PR3 + the spec identity pass in docs/prd/talents-2.0.md): picking a spec now grants a REAL spell and a mastery that changes how the class plays. Supersedes the signatures half of #1348 and replaces #1542 (closed; this PR's branch contains it).

Signatures (formerly #1542):
- All 27 specs grant a real, grant-only signature ability on spec selection (24 new ability defs: Holy Shock, Flashfire, Gloamveil Form, Conflagrate, Swiftmend, Howling Rage, Killer's Calm, Mirrored Blades, Dread Aspect, ...). Old fake signatures return to plain kit abilities; Stormstrike and Chain Heal demoted out of the shaman baseline to grant-only, test-asserted.
- Signature reworks so each feels real: Chain Heal is a true bounce heal (primary plus up to 2 ally jumps picked most-injured-first, classic 50% falloff), Dread Aspect buffs the warlock and the demon (new pet_spellhaste) and renders as a larger fel-purple demon, Moonwing Form gets damage, armor, and a violet tint, Primal Surge is a form-gated resource burst.
- One shared render form-tint feature covers Gloamveil Form, Moonwing Form, and Dread Aspect; remote players see the tints through the existing aura wire, zero server changes.
- Bug fixes found in hand-testing: casts ignored the spellHaste stat, fractional buff values rounded to 0 under a spell-damage mastery, multi-buff abilities evicted their own companion auras, spec-granted abilities now announce so they land on the action bar and spellbook.
- IP compliance (G0 ip_scrub): 15 spells that initially shipped verbatim WoW names now use the LOCKED NAME-MAP coinages with ids frozen (Aether Surge, Howling Rage, Mirrored Blades, Killer's Calm, Flashfire, Primal Mastery, Primal Surge, Red Ribbon, Hallowed Wall, Dread Aspect, Moonwing Form, Anointing, Gloamveil Form, Veinleech, Sureflight Aura); Bear Form / Cat Form / Holy Light phrases scrubbed from descriptions.

Spec identity masteries:
- Six new scoped mastery fields (a closed set, same discipline as the P1-P5 primitives): dotDmgPct, hotHealPct, absorbPct, meleeHastePct, petDmgPct, petDmgSharePct; critDmgPct extended to heal crits; the vs-rooted crit predicate widened to rooted-or-chilled.
- The owner-approved table: holy paladin heal crits hit 2x, discipline shields absorb 30% more, resto druid HoTs heal 25% more, resto shaman heals cost 20% less mana, affliction DoTs deal 20% more, fire spell crits hit 2x, frost gains crit vs rooted or chilled, beast mastery pets deal 35% more, combat rogue trades 10% ability damage for 10% attack speed, demonology redirects 20% of damage taken to the demon. Extended to all 27 specs: the 13 remaining weak masteries upgraded with the same vocabulary, differentiated by mechanic.
- Masteries scale with level (min(1, level/20)), full strength at 20, re-baked live on every ding; the public character sheet derives mastery stats at the character's real level.

Base: release/v0.23.0, kept current by merging the release branch in (conflicts are locale/generated files, resolved by regeneration).

## Credits

Chain Heal's mechanics are adopted from @Blaine1705's standalone implementation in #1434, with authorship preserved (Co-Authored-By on the adopting commit): the deterministic most-injured hop selection, the classic-exact 50% per-jump falloff, the green healing-arc VFX, the custom healing cast sound, and the dedicated test suite. This branch keeps Chain Heal as the Restoration spec signature (granted on spec pick) rather than a baseline learn, per the Talents 2.0 design.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: npm run gate on the release-merged head: PASS, all 9 steps green (full suite 9848 tests / 0 failures with bounded workers, tsc, all builds, malware scan, changed-files biome, i18n freshness). Targeted vitest batteries over every talent suite and guard test after each release-drift re-merge.
- Parity: goldens show state-only movement; RNG draw digests and per-checkpoint draw counts byte-identical. Fresh architecture-reviewer passes over both halves: 0 blocking findings.
- Live hand-test: scripts/spec_hand_test.mjs drives the real offline client for all 9 classes and all 27 specs, asserting every signature resolves and casts with zero page errors (ALL PASS).
- Manual steps: hand-tested spec pick, signature casts, form visuals, cast-time tooltips, and the retuned mastery text in the Specialization tab; screenshots below, repeatable via scripts/spec_signatures_shot.mjs and scripts/spec_masteries_shot.mjs.

## Screenshots / recordings

| Frost mastery panel | Warlock signature spellbook |
|---|---|
| spec-mastery-frost | signature-spellbook-warlock |

| Moonwing Form tint | Dread Aspect demon |
|---|---|
| moonkin-form-visual | metamorphosis-visual |

| Discipline (absorb mastery) | Combat (attack-speed tradeoff) |
|---|---|
| spec-mastery-discipline | spec-mastery-combat |

| Affliction (DoT mastery) | Restoration druid (HoT mastery) |
|---|---|
| spec-mastery-affliction | spec-mastery-resto-druid |

---

### Quality

- [x] **Tests pass.**
- [x] **Typecheck passes.**
- [x] **Builds succeed.**

### Cross-platform

- [x] **Tested on desktop and mobile.** Content and tooltip text on existing responsive surfaces; no layout changes.
- [x] **Accessible.** No new interactive UI.

### Localization (i18n)

- [x] **New player-visible strings handled.** 24 new ability defs plus reworked tooltips and mastery texts localized per the workflow; dialect overlays cleaned to diverge from their base; localization gates green.
- [x] Player-facing sim text re-localized at the client boundary in the same change.

### Hygiene

- [x] Generated files regenerated, not hand-edited (i18n resolved tables, parity goldens, wiki content).


ryan-foo:
Owner directive applied (885b552f0): every spec mastery now has a designated scalable axis for the future mastery stat (item rating, not yet implemented), documented as a per-spec table in docs/prd/talents-2.0.md (Mastery rating readiness). One content change: Frost's mastery is reshaped from crit-vs-rooted to +25% Frost spell damage (ability-scoped to the frost kit so fire/arcane baseline spells stay untouched), keeping the static +10% armor; the rooted-or-chilled crit identity returns as a Shatter-style row option in the flip content. All gates green.

jgyy (commented):
This is a well-executed pass at the "signatures must be real" problem the PRD calls out: the audit that only warrior's 3 signatures were exclusive to their spec, and the fix (a dedicated grant-only spell per spec, no baseline-kit demotions) is exactly what's implemented. The mastery rework is also a real improvement: several new global axes (`dotDmgPct`, `hotHealPct`, `absorbPct`, `meleeHastePct`, `petDmgPct`, `petDmgSharePct`, `critDmgPct`, `spellHastePct`) replace generic "+X% damage" masteries with spec-flavored ones, and they're threaded all the way to `Entity.critDmgBonus` / `meleeHaste` / `spellHaste` and consumed correctly in `damage.ts` and `heal.ts` (including the heal-crit path, `heal.ts:97`, for Paladin Holy's crit-heal mastery). Nice catch buried in `classes.ts` `scaleEffect`: rate-valued `selfBuff`/`buffTarget` effects (haste/crit fractions like `0.2`) were being multiplied by the global damage multiplier and rounded to 0, which would have silently zeroed Arcane Power's haste bonus for an Arcane mage; the `Math.abs(eff.value) < 1` guard fixes that class of bug for every future rate-valued buff, not just this one.

Chain Heal's target selection (`effect_dispatch.ts`, the `chainHeal` case) is deterministic and documented as such: hop choice is (hp fraction, then distance, then id), a total order independent of grid iteration order, so no non-determinism there.

A few things worth fixing:

**should-fix**: Chain Heal's tooltip is wrong. `src/sim/content/classes.ts` sets `falloff: 0.5` (50% less per jump) and its own description string correctly says "50% less," but the i18n catalog entry that's actually rendered via `t()` (`src/ui/i18n.catalog/abilities.ts`, `entities.abilities.chain_heal.description`) says "healing for 40% less with each jump" in every locale (English source plus all the resolved locale files). `talent_tooltip_accuracy.test.ts` doesn't catch this because it isn't asserting on this specific string. Players will see a falloff number that doesn't match the actual mechanic.

**should-fix**: The new `bleed_vuln` aura kind (Rogue Subtlety's Hemorrhage, `types.ts`, applied via `applyDebuff` in `effect_dispatch.ts`/`damage.ts`) is never referenced anywhere under `src/ui/` or `src/render/`. It's not in `DEBUFF_AURA_KINDS` (`src/ui/auras_view.ts`) and not in `TOGGLE_KINDS`, so it won't classify as a debuff on the target frame and has no dedicated rendering path at all. Since this is a real combat-relevant debuff (target takes 40% more bleed damage while it's up), it should be classified so it's visible where a player would react to it, consistent with the "target debuffs must never be hidden" rule.

**nit**: `form_moonkin`, `form_shadow`, and `form_metamorph` are new long-duration (3600s) toggle-style auras, same pattern as the existing `form_bear`/`form_cat`/`form_travel`, but weren't added to `TOGGLE_KINDS` in `auras_view.ts`. They'll likely show a ticking countdown instead of reading as a toggled mode like the other forms. Cosmetic, but inconsistent with the established convention for shapeshift auras.

**should-fix**: `src/net/online.ts` (the `ClientWorld` mirror, three call sites: the snapshot-decode path around line 1835, and the loadout-save/delete paths around 2716/2737) calls `computeTalentModifiers(this.cfg.playerClass, talents)` without the new `level` argument, so it defaults to `MAX_LEVEL`. Every server/sim call site passes the real level (`sim.ts:1733`, `damage.ts:724`, `progression/talents.ts:58`, `fiesta.ts`). Since mastery now scales by `min(1, level / 20)`, an online character below level 20 will see their mastery displayed at full strength client-side while the server (correctly) applies a weaker value. Display-only, not a combat/anti-cheat issue since the server stays authoritative, but it's a real cross-host stat mismatch a player would notice. `e.level` / `this.player.level` are already in scope at all three sites, so this looks like a one-line fix each.

**should-fix**: `src/sim/combat/spell_combat.ts` adds `noteSpellHit(..._args: unknown[]): void {}` with the comment "Hot Streak state is not present in this target branch." It's called on every spell hit from `effect_dispatch.ts` and `casting_lifecycle.ts`, but is a total no-op. It's harmless today, but it's dead scaffolding for a mechanic (Fire's Hot Streak) that doesn't exist yet, loosely typed (`unknown[]`), and reads like it could silently mask a real bug if someone assumes it does something. Either wire it up or drop the call sites until Hot Streak actually lands.

**nit**: The branch is quite stale against `release/v0.23.0` (merge-base is ~280 commits back, this branch only ~39 ahead of it). A raw file-to-file diff against the current base makes it look like this PR deletes the bank window, town focus, `professions/focus.ts`, and reverts a pet-AI grid optimization back to a full entity scan; none of that is real, it's all drift from the stale base (confirmed both by diffing against the true merge-base and by `git merge-tree`, which shows a clean three-way merge that keeps all of it, the only real conflicts being the generated i18n/parity-golden trio, the standard mechanical regen churn). Still worth rebasing before merge so the diff GitHub shows reviewers isn't this misleading, and so CI runs against current behavior rather than a 280-commit-old baseline.

Test coverage is solid: `spec_masteries.test.ts` pins each spec's mastery to a literal (23 of 27 with a full object match, the other 4 warrior/shaman specs only per-axis), the signature tests decisively assert on the ~9 signatures with distinctive mechanics (chain heal bounce, feral charge per-form, metamorphosis dual buff, moonkin/shadowform toggling and the heal-breaks-shadowform interaction), and `talents.test.ts` keeps the old mastery-value literals in lockstep with the new ones. One gap: `spec_signatures.test.ts`'s all-27 aggregate check accepts "cooldown started" or "resource spent" as evidence a signature "fired," which almost any ability satisfies on cast, so it doesn't actually prove the signature's unique mechanic works for the ~18 specs that aren't covered by a more specific test elsewhere. Not blocking since the mechanics that matter most already have dedicated tests, but worth tightening if a future signature regresses silently.

Determinism and sim purity look fine: no `Math.random`/`Date.now` additions, the new AoE/chain-heal iteration order is normalized to a total order before use, and I didn't find any `render`/`ui`/`game`/`net` imports creeping into `src/sim/`. The wire side is also clean: `critDmgBonus` is correctly a server-only derived stat (never encoded on the wire, but mirrored into `blankEntity` so `Sim` and `ClientWorld` shapes stay identical), and all the new mastery/signature resolution stays in shared `src/sim/`, nothing decided client-side.

Below is a summary of which class specs the new signature spells and identity masteries touch, and which mastery axis each spec's identity now runs on.

```mermaid
flowchart TB
  subgraph Warrior
    W1[Arms: Mortal Strike<br/>+meleeDmg 15%, +critDmg 25%]
    W2[Fury: Bloodthirst<br/>+crit 10%, +meleeDmg 10%]
    W3[Protection: Shield Slam<br/>+threat 50%, +armor 20%]
  end
  subgraph Paladin
    P1[Holy: Holy Shock NEW<br/>+critDmg 50% on heals]
    P2[Protection: Hallowed Wall NEW<br/>+threat 50%, +armor 20%]
    P3[Retribution: Crusader Strike NEW<br/>+meleeDmg/spellDmg 20%]
  end
  subgraph Hunter
    H1[Beast Mastery: Howling Rage NEW<br/>+petDmg 35%]
    H2[Marksmanship: Sureflight Aura NEW<br/>+meleeDmg 20%, +crit 3%]
    H3[Survival: Wyvern Sting NEW<br/>+meleeDmg 15%, +agi 15%]
  end
  subgraph Mage
    M1[Arcane: Aether Surge NEW<br/>+spellDmg 15%, +spellHaste 10%]
    M2[Fire: Flashfire NEW<br/>+critDmg 50%, +crit 2%]
    M3[Frost: Icy Veins NEW<br/>Frostbolt/Frost Nova +25%, +armor 10%]
  end
  subgraph Rogue
    R1[Assassination: Killer's Calm NEW<br/>+dotDmg 20%, +crit 3%]
    R2[Combat: Mirrored Blades NEW<br/>+meleeHaste 10%, -meleeDmg 10%]
    R3[Subtlety: Red Ribbon NEW<br/>bleed_vuln debuff, +critDmg 40%, +agi 10%]
  end
  subgraph Priest
    Pr1[Discipline: Anointing NEW<br/>+absorb 30%, +maxHp 8%]
    Pr2[Holy: Holy Nova NEW<br/>+healPct 20%]
    Pr3[Shadow: Gloamveil Form NEW<br/>+dotDmg 15%, +spellDmg 10%<br/>breaks on heal cast]
  end
  subgraph Shaman
    S1[Elemental: Primal Mastery NEW<br/>+spellDmg 15%, +spellHaste 10%]
    S2[Enhancement: Stormstrike existing<br/>+meleeHaste 10%, +meleeDmg 10%]
    S3[Restoration: Chain Heal NEW<br/>bounce heal, -20% cost<br/>*tooltip falloff mismatch*]
  end
  subgraph Warlock
    L1[Affliction: Veinleech NEW<br/>+dotDmg 20%]
    L2[Demonology: Dread Aspect NEW<br/>+petDmgShare 20%, +sta 10%]
    L3[Destruction: Conflagrate NEW<br/>+critDmg 50%, +crit 2%]
  end
  subgraph Druid
    D1[Balance: Moonwing Form NEW<br/>+spellDmg 15%, +spellHaste 10%]
    D2[Feral: Primal Surge NEW<br/>+meleeDmg/dotDmg 15%, +threat 20%]
    D3[Restoration: Swiftmend NEW<br/>consumes HoT, +hotHeal 25%]
  end

  Core[["src/sim/content/talents.ts<br/>new GlobalModEffect axes:<br/>dotDmgPct hotHealPct absorbPct<br/>meleeHastePct petDmgPct petDmgSharePct<br/>critDmgPct spellHastePct"]]
  Core --> Warrior & Paladin & Hunter & Mage & Rogue & Priest & Shaman & Warlock & Druid

  style S3 fill:#5a2020,stroke:#e05050
  style R3 fill:#5a2020,stroke:#e05050
```

(Warrior's three specs were already real signatures per the PRD audit and are unchanged; the other 24 specs each get a new, spec-exclusive signature spell here. Nodes flagged red are where I found a should-fix issue: Chain Heal's tooltip falloff text and Hemorrhage's `bleed_vuln` debuff not being classified in the HUD's aura system.)

ryan-foo:
Rebased onto release/v0.25.0 and ran a codex review pass over the branch. Codex confirmed the three merge-conflict resolutions are correct (the level-aware mastery re-bake, refreshKnownAbilities(true) confined to live actions, equipmentInstance threaded through every derived-stat call) and that all 27 authored mastery values match the locked spec.

The pass also surfaced 9 pre-existing correctness gaps in the feature. 8 are fixed here, each tested:

- Level-scaling: setPlayerLevel now re-bakes the level-scaled mastery (dev/GM jumps were leaving it stale); the three client-side computeTalentModifiers rebuilds now pass the real level instead of defaulting to 20.
- Hallowed Wall: the "bounce to 2 enemies" was caster-centered and uncapped; it is now a proper target-centered, primary-excluded, jumps-capped chainDamage (arc LoS checked hop-to-hop).
- AoE mastery coverage: masteries now reach groundAoE and aoeHeal (Consecration, Earthquake, Holy Nova) which were unscaled.
- Rate-buff corruption: a damage mod was scaling and rounding utility rate/multiplier buffs (Ghost Wolf 1.4, Blade Flurry 1.2); the guard is now kind-based, so flat magnitudes (Demonic Skin armor) still scale but rates pass through.
- Demonology redirect: the redirected pet share no longer re-applies the source's Defensive Stance cut / Weakening Hex.
- Crit-damage channels: critDmgBonus was one universal field; split into spell / physical / heal channels so a Holy paladin's heal mastery no longer buffs damage crits and Fire/Destruction's spell mastery no longer buffs physical or heal crits.
- Gloamveil Form: its +15 spell power is now Shadow-school only, no longer leaking onto a shadow priest's non-Shadow casts.

The 9th finding (a mastery/talent damage % should scale the full hit including the weapon/AP/SP rider, not just the authored base) is intentionally NOT in this PR. It is a game-wide combat-scaling change that touches every talent, rebalances damage upward at high gear, re-mints all goldens, and needs a balance sign-off, so it is filed separately as #1803 for its own PR.

Rubsey (commented):
Solid, careful PR, and the determinism work is the standout. Chain Heal and Hallowed Wall both pick their hops with a total-order tiebreak (hp fraction, then squared distance, then id), so the chain is identical across the three hosts and the only rng is the one base roll plus each hop's crit. architecture.test.ts and the parity goldens are green in a clean checkout, the new form aura kinds ride the existing aura wire (kind + value round-trip) so remote clients get the tints with zero server changes as claimed, and the three new derived Entity fields are added to both baseEntity and blankEntity so Sim and ClientWorld keep identical shapes. The mastery level-scaling re-bakes on every path that can change level or talents (ding, dev level-jump, respec/spec/loadout, fiesta, load), the multi-buff companion-aura id fix is correct, and TS exhaustiveness on EFFECT_CLASS and scaleEffect forced every new effect type into the UI and estimate paths too. Nice work.

**should-fix**

- Chain Heal's tooltip understates the falloff. `src/ui/i18n.catalog/abilities.ts:1193` says "healing for 40% less with each jump", but the effect uses `falloff: 0.5` (50% less) and the inline description at `classes.ts:2306` correctly says "50% less". Players see the catalog string, so fix the copy to "50%".
- Chain Heal's custom cast sound never plays. `hud.ts:8476` fires `'cast_chain_heal'` on castStart, but that key is not registered in `scripts/sfx/sfx_prompts.mjs`, so it is absent from the generated sfx manifest, `sfx.entry` returns undefined, and the cast is silent. The 489KB `public/audio/sfx/cast_chain_heal.wav` ships unused and no test gates it. Register the key and regenerate the manifest.

**nits**

- `leechPct` (Veinleech / siphon_life) has no decisive test. The all-27 signature loop only proves it emits a damage event, which the old non-leech dot did too. Worth asserting the caster is healed by the tick (`auras.ts:185`).
- `petDmgPct` is asserted only through the `petDamageMult` helper. The code applies it at both `mobSwing` (melee, `sim.ts:4845`) and the ranged pet path (`sim.ts:4910`), so it is correct, but a regression at either site would pass the suite.
- `chain_heal.test.ts` pins the falloff only as monotone (`heals[1] <= heals[0]`), not the exact 0.5, so it would also pass at 0.9. One exact-ratio assertion would have caught the tooltip drift above.
- `petDamageMult` is duplicated (`Sim.petDamageMult` and the `pet_ai.ts` module copy), and `noteSpellHit` is a no-op stub with `...unknown[]` args called from four hot-path sites (a seam for #1614's Hot Streak). Both harmless, both a little untidy for main.
- `chainHeal` is the only heal effect `scaleEffect` leaves in its default unscaled case, so it ignores healing-done talent mods that `heal`/`aoeHeal` pick up. Probably fine given its own tuning, but the asymmetry is worth a look.
- `playerSpellHasteFrac` (`hud.ts:14789`) re-walks the buff_spellhaste auras that `spell_combat.ts` `spellHasteMult` already sums; returning `spellHasteMult(p) - 1` keeps the tooltip and the real cast from drifting.

**Notes, not asks**

- The only real merge conflict against the current base is the usual stale i18n trio (`i18n.resolved.sha256`, `i18n.status.summary.json`). Mechanical: merge the base and re-run the i18n build. Also the PR body says the base is release/v0.23.0, but it is release/v0.25.0 now.
- Heads up for whoever lands this: the PR also trains seven baseline class interrupts (Jawcrack/Boot/Spellbreak/Hushing Shot/Reproach/Headbutt/Gag Order) across seven classes, and `scaleEffect` no longer scales rate/percent selfBuffs or lifeTap/gainResource by damage mods (a reasonable fix, goldens regenerated). Both are fine, just larger than "signatures + masteries".

Reviewed English i18n only; the locale fills are release-time work.

PR #1544, opened 2026-07-06
feat(admin): show evidence recurrence history

madmatah:
## Summary

Operators triaging the Suspicious Players page could not tell a one-off detection from a repeating one: the dashboard shows each evidence entry's strongest reading, so an observation that fired once and one that fired twenty times across a session looked identical. This PR surfaces the recurrence history the detector now attaches to its evidence.

- `SuspiciousEvidence` (the `BotDetector` seam in `server/bot_detector/contract.ts`) gains four optional fields: `occurrences`, `firstAt`, `lastAt` (epoch ms), and `episodesAt` (a bounded ring of the most recent episode timestamps, for temporal analysis of the export; the count and firstAt keep the totals the ring loses). Which evidence carries them is decided entirely by the detector implementation; evidence describing a persistent property carries none. The change is additive: the no-op stub and older detector builds simply never set them, and the dashboard renders nothing in that case.
- The Suspicious Players page renders the history as a dim sub-line under each evidence detail: `seen x4 | first 40 min ago | latest 12s ago` (the redundant "first" is omitted for a single occurrence). Relative times reuse `fmtRelative`, the count goes through `Intl.NumberFormat`.
- The Download JSON export is realigned on the raw payload (schema 2): the redacted schema 1 (no evidence detail, state, name, or IP) could not support offline triage, since several detection strategies share one evidence kind and triage keys on the detail text. The file now carries `data.players` verbatim (recurrence fields included) plus capture metadata, so analysis scripts written against a saved raw response read a download unchanged. It is an operator-only artifact behind admin auth; downloads are sensitive (player names and IPs) and stay out of public trackers. The dashboard renders the count and the first/latest endpoints; the `episodesAt` ring is export-only for now (its consumer is offline analysis, not the page).

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/admin/` green (130 tests), including a new component case pinning the three render shapes: repeating evidence (count + first + latest), single occurrence (count + latest, no "first"), and evidence without history (no sub-line), on synthetic placeholder data
  - `npx vitest run tests/admin/suspicious_sessions_export.test.ts` green with the extended payload shape
  - `npm run check:admin` (svelte-check) 0 errors, `npx tsc --noEmit` clean
  - `npm run i18n:admin` regenerated `src/admin/i18n.resolved.generated/` for the three new keys
- Manual steps:
  - Local server + dashboard: evidence rows with history show the sub-line, rows without history are unchanged; Download JSON carries the new fields

---

### Quality

      TypeScript `strict`).

### Cross-platform

- [x] **Tested on desktop and mobile.** The recurrence sub-line lives inside the
      existing `.evidence-detail` cell, which already reflows full-width under
      900px; no new interactive element.
- [x] **Accessible.** Plain text inside the existing list item; no new controls.

### Localization (i18n)

- [x] New operator-visible strings are `t()` keys added to the admin English
      catalog (`suspiciousPlayers.evidenceOccurrences` / `evidenceFirstSeen` /
      `evidenceLastSeen`), resolved tables regenerated via `npm run i18n:admin`;
      locale overlays are filled by the maintainer at release per the admin
      catalog policy (`src/admin/CLAUDE.md`).
- [x] Numbers and relative times go through `Intl.NumberFormat` and `fmtRelative`.
- [ ] N/A. This PR adds no player-visible strings (game client untouched; the
      strings are operator-facing).

### Hygiene

- [x] `src/admin/i18n.resolved.generated/` was regenerated by `npm run i18n:admin`,
      not hand-edited.


FernandoX7 (commented):
Clean, well-scoped admin change, and the reasoning is documented where it needs to be. The contract fields are additive and optional, so the stub and older detector builds keep working; the count and relative times go through `Intl.NumberFormat` and `fmtRelative`; and the tests are decisive: the component test pins all three render shapes (count + first + latest, count + latest only, no sub-line) and the export test pins the raw shape plus the Infinity-to-null serialization.

**should-fix (confirm intent): the Download JSON export drops the schema 1 redaction.** Schema 1 deliberately kept names, IPs, and evidence detail out of the file, with a test asserting it; schema 2 writes `data.players` verbatim (`suspicious_sessions_export.ts:23`) and this PR removes those `not.toContain` assertions. It is defensible: the same operators already see name, IP, and detail live on the page (`AccountLink` / `IpLink` / `.evidence-detail` at `SuspiciousPlayers.svelte:267,269,291`), and the download reuses the already-loaded `data` behind the same `botdetector.read` gate, so it widens no audience. The only incremental exposure is a PII-bearing file at rest, which the PR already flags. Since it reverses a safeguard that was built on purpose, it is worth a conscious sign-off rather than a silent flip. (Aside: anything parsing a schema 1 file breaks, but that is what the `schemaVersion` bump is for.)

**nit:** the recurrence sub-line is guarded twice, once in the template (`{#if evidence.occurrences !== undefined}` at `SuspiciousPlayers.svelte:292`) and again by `recurrenceLabel`'s early `return ''`. Harmless and defensive, just redundant.

The English i18n looks right (the three new keys are in the admin English catalog and render via `t()`), and the generated-table churn is the expected regen.

FernandoX7 (commented):
Re-reviewed the amended commit (`3f52d2b`). Both points from the last pass are handled.

**nit (double-guard): fixed.** The template computes the label once (`{@const recurrence = recurrenceLabel(evidence)}` at `SuspiciousPlayers.svelte:289`) and guards on the truthy string (`{#if recurrence}` at `:294`), so the redundant `occurrences !== undefined` check is gone. `recurrenceLabel` still early-returns `''` for history-less evidence, so the single guard is exactly right.

**should-fix (schema 1 redaction): signed off.** The export stays on schema 2 (raw `data.players`), and the intent is now documented in place: the `buildSuspiciousSessionsExport` header comment records the 2026-07-05 triage that found the redacted shape unusable and flags the file as operator-only and sensitive (`suspicious_sessions_export.ts:9`). That reads as a conscious keep rather than a silent flip, which is what I was after.

Tests still pin the behavior: the export test asserts the full raw shape including the recurrence fields and the Infinity-to-null serialization, and the component test pins the three render shapes (count + first + latest, count + latest only, no sub-line). `npx vitest run tests/admin/suspicious_sessions_export.test.ts` is green on my end.

Nothing else from me.

PR #1545, opened 2026-07-06
fix(desktop): split prod/dev auto-update tracks and refuse cross-origin updates

### LINKED ISSUE #1537: Desktop auto-update can flip a production app to the dev API backend
## Problem

A production macOS desktop install auto-updated itself onto a build that talks to the
dev API. The player saw the in-app "update available" toast, clicked update, restarted,
and the running app was then calling the dev backend instead of production.

The desktop auto-updater has no separation between production and non-production builds.
There is one generic update feed for the whole website channel
(`package.json` build.publish: `{ provider: "generic", url:
"https://updates.worldofclaudecraft.com/desktop" }`), read on the default
electron-updater channel (`latest`), with no `channel` split and no `allowPrerelease`
gate. At the same time, the API origin is frozen into each build artifact at build time:
`scripts/electron-build.mjs` bakes `VITE_DESKTOP_API_ORIGIN` into both the Vite bundle
and the `extraMetadata.wocDesktop.apiOrigin` stamp, and the packaged app reads that stamp
at runtime (`electron/desktop_config.cjs`, `electron/main.cjs`) while ignoring runtime env.

The consequence: any build baked with a non-production origin (for example a smoke-test
build made per the runbook with `VITE_DESKTOP_API_ORIGIN=http://localhost:8787`, or a
staging build), if it ever lands on that one feed or is referenced by the `latest-*.yml`
there, is served to every production install as "the latest version," auto-downloaded,
and installed on restart. The updated app then calls whatever origin that artifact was
baked with. Nothing validates that an incoming update keeps the app on the same
production backend before it installs.

## Steps to reproduce

1. Install a production website-channel desktop build (baked with the production
   `apiOrigin`, `https://worldofclaudecraft.com`).
2. Publish a build baked with a non-production `VITE_DESKTOP_API_ORIGIN` (dev or staging)
   to the same update host, updating `latest-mac.yml` to point at it with a higher version.
3. Launch the production app and let the updater run (15s after launch, then every 4h).
4. Accept the "update available" toast and restart (or just quit, which installs on quit).
5. Observe the app now issues its REST and WebSocket calls to the dev origin, not
   production.

## Expected behavior

- A production install only ever auto-updates to another production build. After updating
  and restarting, it still calls the production origin (`https://worldofclaudecraft.com`).
- Non-production (dev or staging) builds are never offered to, downloaded by, or installed
  by a production app. Production and non-production builds live on separate update tracks
  (a distinct electron-updater `channel`, or a distinct feed URL), so the two cannot cross.
- Defense in depth: an update whose baked API origin differs from the running production
  origin is refused rather than installed, so a stray artifact on the wrong feed cannot
  silently flip a shipped app to another backend.
- Steam builds are unaffected (their in-app updater stays off).

## Actual behavior

- There is one undifferentiated `latest` feed for all website builds, so a production app
  treats a dev/staging build on that feed as its next version.
- The app installs it and, because the API origin is baked into the downloaded artifact,
  starts talking to the dev backend after restart, with no warning or validation.

## Scope

- Update feed and channel wiring: `package.json` build.publish (single generic feed, no
  channel), `electron/updater.cjs` (no `channel` / `allowPrerelease`, `autoDownload` and
  `autoInstallOnAppQuit` on).
- Build and publish flow: `scripts/electron-build.mjs` and
  `scripts/electron-builder-config.mjs` (how `VITE_DESKTOP_API_ORIGIN` is baked and
  stamped), and `docs/desktop-release.md` (the single upload target and `latest-*.yml`
  procedure, including the documented local-origin smoke-test build).
- Runtime origin resolution and any pre-install guard: `electron/desktop_config.cjs`
  (`resolveDesktopOrigins`, `wocDesktop.apiOrigin` stamp), `electron/main.cjs`.

## Acceptance criteria

- Publishing a build baked with a non-production API origin cannot place it on the
  production update track. Production and non-production builds are separated by an
  explicit `channel` or feed URL, not just by process discipline.
- A production install that updates and restarts still calls
  `https://worldofclaudecraft.com`, verified end to end.
- A build-time or CI guard fails if a build whose baked `apiOrigin` is not the production
  origin would be emitted onto (or uploaded to) the production feed, so the mistake is
  caught before release rather than by players.
- Optionally, the app refuses to install an update whose stamped API origin differs from
  the installed production origin (covered by a test).
- Steam builds are unchanged: the in-app updater stays disabled and no feed is consulted.

## Environment

- Platform: desktop (Electron), macOS observed, website (auto-update) channel.
- Surface: in-app auto-updater (electron-updater), on the post-update restart.

## Notes

Most likely operational trigger: a build baked with a non-production `apiOrigin` (a
staging build, or a smoke-test build made without setting `VITE_DESKTOP_API_ORIGIN` to
production) was uploaded to `https://updates.worldofclaudecraft.com/desktop/` and its
`latest-mac.yml`, so every production install pulled it.

Underlying design gap: the `website` vs `steam` distribution stamp is orthogonal to a
dev/prod update track (it only decides SteamPipe vs self-update), so there is currently no
mechanism that keeps a production app on production builds. The cheapest durable fix is a
publish-time or CI guard that refuses to emit a production `latest-*.yml` for an artifact
not baked with the production origin, plus a real channel or feed-URL split between the two
tracks. A rollback note for whoever picks this up: per the runbook you cannot re-publish
the same or a lower version, so unwinding a bad flip means shipping a higher production
version and, if needed, dropping `stagingPercentage` to 0 on the bad feed to stop further
spread.

FernandoX7:
## Summary

A production macOS install auto-updated onto a build that talks to the dev API (#1537). Root cause: every website desktop build shared one undifferentiated `latest` update feed, and the API origin baked into each artifact rode along unchecked, so any dev/staging artifact that reached the feed became "the latest version" for every production install.

This PR separates the update tracks and adds defense in depth, in three layers:

1. **Track split at build time.** The electron-builder publish channel is now derived from the baked `apiOrigin` by one rule shared between build and runtime (`electron/update_guard.cjs`): the production origin publishes and reads the `latest` channel (`latest-mac.yml` and friends, the feed shipped installs already read), and ANY other origin (dev, staging, localhost smoke packs) publishes and reads a `dev` channel (`dev-mac.yml` and friends), which production installs never request. Requesting the production channel with a non-production origin makes the build throw (the CI/build-time guard from the issue's acceptance criteria). `WOC_UPDATE_CHANNEL=dev` remains as the one supported cross: staging a production-origin artifact on the dev track for update-pipeline testing.
2. **Origin stamp in the feed files.** `scripts/electron-build.mjs` stamps `wocApiOrigin` into every emitted feed yml. electron-updater hands the parsed yml through as the `UpdateInfo` object, so the stamp reaches the running app.
3. **Runtime refusal.** `electron/updater.cjs` now sets its update channel from its own baked origin (so even a build packaged with a wrong `app-update.yml` stays on its origin's track), turns `autoDownload` off, and only downloads after the offer passes the origin guard: an update whose `wocApiOrigin` stamp differs from the install's own origin is refused with a loud `[updater] REFUSED` entry in `main.log`, so even a feed file hand-renamed onto the wrong track cannot flip a shipped app to another backend. Feed files published before the split carry no stamp and are accepted (back compat with the current production feed). The change also undoes a hidden electron-updater trap: assigning `channel` silently flips `allowDowngrade` to `true`; we reset it so a production install can never move backward.

Steam builds are untouched: publish stays nulled, no feed files are emitted, and the in-app updater stays off (pinned by the existing and new tests).

Recovery note for currently flipped installs: an install that already took a dev-origin artifact still reads `latest-*.yml` from the production feed (its baked publish config is unchanged), so publishing the next higher production version brings those players back to production automatically, per the runbook's forward-only rollback model.

The startup banner now logs `updateChannel` next to `updaterEnabled`, and `docs/desktop-release.md` / `docs/desktop-ship-notes.md` document the track model, the guard, and the "never rename dev feed files" rule.

## Related issues

Closes #1537.

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/electron_update_guard.test.ts tests/electron_updater_track.test.ts tests/electron_builder_config.test.ts tests/electron_desktop_config.test.ts tests/electron_update_cadence.test.ts tests/electron_update_events.test.ts tests/electron_vendor_loading.test.ts` (all green; the two new suites cover the origin rule, the accept/refuse matrix, the build-time guard throw, feed-file stamping idempotence, and the updater wiring incl. the `allowDowngrade` reset, via `initUpdater`'s injected-autoUpdater test seam)
  - `npx tsc --noEmit` clean; changed files pass `biome check`
  - `npm run electron:pack` end to end: electron-builder accepts the derived config with `publish.channel`, and the packaged app carries the expected stamp
  - Verified against the vendored electron-updater/electron-builder sources that `publish.channel` names the emitted feed files (`updateInfoBuilder.js`), that `autoUpdater.channel` selects the requested feed at runtime (`GenericProvider.js`), and that extra yml keys survive `parseUpdateInfo` onto `UpdateInfo`
- Manual steps:
  - Simulated the issue's scenario at the unit level: a production install offered an update stamped `https://dev.worldofclaudecraft.com` (or `http://localhost:8787`) refuses to download, sends no toast, and logs the offered vs expected origins; the same offer stamped with the production origin downloads and toasts as before
  - Confirmed a pre-split feed file (no stamp, like the current production `latest-mac.yml`) is still accepted, so already-shipped installs keep updating

### Quality

- [x] **Tests pass.** Targeted suites green (full `npm test` runs in CI). I added or updated tests for the behavior I changed.
- [x] **Builds succeed.** `npm run build` runs as part of `npm run electron:pack`, verified end to end.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: Electron main process, build scripts, tests, and docs only; no UI change.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (updater refusals are dev-channel `main.log` lines by design, matching the existing updater logging convention; the player-facing update toast is unchanged).

### Hygiene

### DIFF
--- a/docs/desktop-release.md
+++ b/docs/desktop-release.md
@@ -28,12 +28,34 @@ endpoint. The updater runs only for a PACKAGED WEBSITE build; there is deliberat
 no way to force it on in a Steam build. To try either channel unpacked, set
 `WOC_DISTRIBUTION=website|steam` on `npm run electron:dev`.
 
+Update tracks (prod/dev split): the publish channel is derived from the baked
+`apiOrigin` by one rule shared between build and runtime
+(`electron/update_guard.cjs`). A build baked with the production origin publishes
+and reads the `latest` channel (`latest-mac.yml`, `latest.yml`,
+`latest-linux*.yml`); a build baked with ANY other origin (dev, staging, a
+localhost smoke pack) publishes and reads the `dev` channel (`dev-mac.yml` and
+friends), which production installs never request. Three layers keep the tracks
+apart: the build throws if the production channel is requested for a
+non-production origin (`scripts/electron-builder-config.mjs`); every emitted feed
+file is stamped with the `wocApiOrigin` its artifact was baked with; and the
+running app refuses to download an update whose stamp differs from its own baked
+origin (loud `[updater] REFUSED` entry in `main.log`), so even a feed file
+renamed onto the wrong track cannot flip an install to another backend.
+`WOC_UPDATE_CHANNEL=dev` on a production-origin build is the one supported
+cross: it emits a production-origin artifact's feed files on the dev track to
+exercise the publish pipeline end to end (no install ever downloads such an
+artifact: dev-origin installs refuse its production origin stamp, which is the
+fail-safe direction). Never rename `dev-*.yml` files to `latest-*.yml` on the
+update host.
+
 `npm run electron:pack` / `electron:pack:steam` are the fast local variants
 (`--dir`, host arch only, no installers). Release builds use the full arch matrix in
 `package.json` `build`: macOS universal (dmg + zip), Windows x64 + arm64 (nsis + zip),
 Linux x64 + arm64 (AppImage + deb). To smoke-test a packaged build against a local
 server: `VITE_DESKTOP_API_ORIGIN=http://localhost:8787 npm run electron:pack` (a
-BUILD-time value: baked into the bundle and stamped into the app).
+BUILD-time value: baked into the bundle and stamped into the app; such a build
+lands on the `dev` update channel automatically and cannot produce production
+feed files).
 
 Build each OS on its own runner (mac artifacts on macOS, Windows artifacts on Windows,
 Linux artifacts on Linux). Cross-building is not part of this runbook.
@@ -145,7 +167,19 @@ Fedora atomic desktops (Bazzite, Steam Deck) with no system install, just
    page links point at the new build (the static hrefs in `index.html` are the
    no-JS fallback; keep them on the same version). The page offers macOS (dmg)
    and Linux (AppImage); Windows stays "pending" until its installer is uploaded.
-2. Build on each OS runner with signing env present: `npm run electron:build`.
+2. Build on each OS runner with signing env present: `npm run electron:build`,
+   with `VITE_DESKTOP_API_ORIGIN` unset or set to the production origin. A
+   production release MUST emit `latest-*.yml` feed files; if the build produced
+   `dev-*.yml` instead, it was baked with a non-production origin: rebuild, do
+   not rename (renamed files still carry the `wocApiOrigin` stamp and every
+   production install will refuse them).
+   One-time cleanup with the first track-split release: audit the production
+   update host and delete any `latest*.yml` (and its artifacts) that this
+   release did not produce. Feed files published before the split carry no
+   `wocApiOrigin` stamp and the runtime guard accepts unstamped files for back
+   compat, so a leftover pre-split dev-baked `latest*.yml` is the one artifact
+   the guard cannot refuse; from this release on, every feed file on the host
+   is stamped and the acceptance window can later be tightened to stamped-only.
 3. Upload from `release/` to the update host directory (keep filenames exactly):
    - macOS: `world-of-claudecraft-<v>-mac-universal.dmg` (download page),
      `...-mac-universal.zip` + `.zip.blockmap` (updater), `latest-mac.yml`.
@@ -247,7 +281,8 @@ Rules that keep this working:
 
 1. Fresh install, launch: window appears, no Gatekeeper/SmartScreen block (signed
    builds), log file created, startup banner shows the right `version`,
-   `distribution`, and `updaterEnabled`.
+   `distribution`, `updaterEnabled`, and `updateChannel` (`latest` on a
+   production build, `dev` on anything else).
 2. GPU: log shows `[gpu] feature status` with hardware WebGL2 (no
    `software only`, no SwiftShader/llvmpipe renderer, no softwareRendering warning).
 3. Login both paths: email/password in-app, and Discord via the external browser +
@@ -257,8 +292,9 @@ Rules that keep this working:
    world (backgroundThrottling stays off).
 5. Website channel only: with a higher-version build on the feed, the update toast
    appears, "Restart now" applies it, and a player who quits instead gets it on next
-   launch. Steam channel: confirm the log says the updater is disabled and no
-   update network traffic occurs.
+   launch; after the restart the log's startup banner still shows the production
+   `apiOrigin` channel (`updateChannel: latest`). Steam channel: confirm the log
+   says the updater is disabled and no update network traffic occurs.
 6. Crash surfaces: `kill -SEGV <renderer pid>` THREE times within a minute (a
    task-manager "end task" is classified as a benign `killed` exit and does not
    trigger recovery). The first two SEGVs each produce a log entry and a bounded
--- a/docs/desktop-release.md
+++ b/docs/desktop-release.md
@@ -45,8 +45,11 @@ renamed onto the wrong track cannot flip an install to another backend.
 cross: it emits a production-origin artifact's feed files on the dev track to
 exercise the publish pipeline end to end (no install ever downloads such an
 artifact: dev-origin installs refuse its production origin stamp, which is the
-fail-safe direction). Never rename `dev-*.yml` files to `latest-*.yml` on the
-update host.
+fail-safe direction). Never rename `dev*.yml` files to `latest*.yml` on the
+update host. Dev installs made BEFORE the track split read the `latest`
+channel like everything else did, so they will auto-update onto production
+builds; give dev testers a fresh post-split dev build rather than expecting
+their old installs to stay on dev.
 
 `npm run electron:pack` / `electron:pack:steam` are the fast local variants
 (`--dir`, host arch only, no installers). Release builds use the full arch matrix in
@@ -169,8 +172,9 @@ Fedora atomic desktops (Bazzite, Steam Deck) with no system install, just
    and Linux (AppImage); Windows stays "pending" until its installer is uploaded.
 2. Build on each OS runner with signing env present: `npm run electron:build`,
    with `VITE_DESKTOP_API_ORIGIN` unset or set to the production origin. A
-   production release MUST emit `latest-*.yml` feed files; if the build produced
-   `dev-*.yml` instead, it was baked with a non-production origin: rebuild, do
+   production release MUST emit `latest*.yml` feed files (`latest.yml` on
+   Windows, `latest-mac.yml`, `latest-linux*.yml`); if the build produced
+   `dev*.yml` instead, it was baked with a non-production origin: rebuild, do
    not rename (renamed files still carry the `wocApiOrigin` stamp and every
    production install will refuse them).
    One-time cleanup with the first track-split release: audit the production
--- a/docs/desktop-ship-notes.md
+++ b/docs/desktop-ship-notes.md
@@ -67,17 +67,26 @@ node_modules in the asar, all seven fuses set.
 
 1. At build time, electron-builder bakes `resources/app-update.yml` into the
    app (from `build.publish`: generic provider, feed URL
-   `https://updates.worldofclaudecraft.com/desktop`).
+   `https://updates.worldofclaudecraft.com/desktop`), and the publish channel
+   is derived from the baked API origin (`electron/update_guard.cjs`): the
+   production origin emits the `latest` feed files, any other origin (dev,
+   staging, localhost smoke packs) emits `dev` ones, and each emitted feed
+   file is stamped with the `wocApiOrigin` its artifact was baked with.
 2. At runtime, `electron/updater.cjs` initializes ONLY when the build is
    packaged AND stamped `website`. It checks the feed 15 seconds after launch
-   and every 4 hours: it GETs the platform's `latest*.yml`, compares the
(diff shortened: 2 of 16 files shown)


FernandoX7:
One operational note for whoever publishes the next release: the fix only fully takes effect once a stamped production release is on the feed; until then the host cleanup step in docs/desktop-release.md is what closes the original hole.

madmatah:
## Verdict: READY (0 blocking), with 3 minor SHOULD-FIX and 2 operational VERIFY items

## Deterministic floor (run directly)

- Biome on the 10 changed source/test files: clean. `npx tsc --noEmit`: clean.
- Guard tests (`tests/architecture.test.ts`, `tests/localization_fixes.test.ts`): green
  (the initial failure was only the ungenerated i18n status registry in a fresh
  worktree, fixed by `npm run i18n:gen`).
- The 4 new test suites plus the adjacent cadence suite: 61/61 pass.
- No em/en dashes or emojis in the diff, conventional commit with scope, no new
  dependency, `.d.cts`/`.d.mts` declaration files in sync with the real exports.

## What is solid (confirmed by several independent reviewers)

The mechanism was verified against the vendored electron-updater 6.x source, not
only against the tests:

- The runtime channel override really drives the feed filename
  (`GenericProvider` reads `updater.channel || configuration.channel`), so the
  origin-derived channel overrides even a wrongly packaged `app-update.yml`.
- The `wocApiOrigin` stamp survives feed parsing (`parseUpdateInfo` is a raw
  `js-yaml.load` with no key whitelist), so the runtime guard is not a no-op.
- The `allowDowngrade` reset is genuinely needed: electron-updater's channel
  setter silently flips it to `true`; the code sets the channel exactly once at
  init and resets immediately, and no later channel assignment exists.
- The security review concludes the trust model is sound: the cryptographic
  boundary (sha512 + code signing + hardcoded HTTPS feed host, not redirectable
  by any env var) is untouched, `autoDownload` moves from true to false
  (stricter), and the stamp is honestly framed as an operator-mistake guard, not
  a security boundary.
- YAML stamping goes through `JSON.stringify` (no injection), excludes
  `builder-debug.yml`, is idempotent, and runs after electron-builder with
  `--publish never` (no race with an upload).
- Test assertions are decisive (exact objects, exact feed filenames including
  `latest-linux-arm64.yml`, the full `evaluateUpdateOffer` verdict matrix, and
  the updater tests drive the real `electron/updater.cjs` through an injected
  seam, not a reimplemented copy).

## SHOULD-FIX (minor, re-verified in the code)

1. **Origin resolution asymmetry between channel and guard** (found independently
   by 3 reviewers; confirmed at `electron/main.cjs:80` and
   `electron/shell_guards.cjs:22`). The update channel derives from the raw baked
   stamp via `apiOriginKey` (fails safe to `dev`), but the origin handed to
   `initUpdater` is `deriveOrigin(stamp) || production` (falls back to
   **production** on parse failure). A build stamped with an unparseable origin
   (for example `localhost:8787` with no scheme) therefore reads the dev feed
   while guarding as production: it refuses every stamped update on its own track
   forever. Fail-safe direction (never a cross-track install), but it silently
   strands that install class, and nothing at build time rejects an unparseable
   `VITE_DESKTOP_API_ORIGIN`. Suggested fix: validate at build time that
   `apiOriginKey(apiOrigin)` is non-null whenever `publish` is active, in the
   commit's own "that mistake must die at build time" spirit.
2. **`WOC_UPDATE_CHANNEL=''` breaks the build** with
   `unknown desktop update channel: ` (empty string is not nullish, so
   `?? null` passes it through). An empty env var, common in CI matrices, should
   mean "unset", that is, derive from the origin. One-line fix:
   `process.env.WOC_UPDATE_CHANNEL || null` in `scripts/electron-build.mjs`.
3. **Glob inconsistency in `docs/desktop-release.md`**: lines 50 and 172-173 say
   `latest-*.yml` / `dev-*.yml`, which literally excludes the Windows feeds
   `latest.yml` / `dev.yml` (lines 177, 180, and 205 of the same doc correctly
   use `latest*.yml`).

## VERIFY (manual, gates the release rather than the merge)

- **Packaged smoke test** from the `docs/desktop-release.md` section 5 checklist:
  a production build fetches `latest-mac.yml`, a dev/localhost build fetches
  `dev-mac.yml`, the startup banner logs the right channel, and a deliberately
  mis-stamped artifact produces the `[updater] REFUSED` log line and no install.
  Cannot run under vitest; needs a real electron build.
- **One-time cleanup of the update host**: the guard deliberately accepts
  unstamped feeds (pre-split back-compat), so an old dev-baked `latest*.yml`
  left on the production host would still be accepted. The runbook documents
  this cleanup; it is load-bearing and must not be skipped.

## NICE-TO-HAVE

- A fixture test that parses a stamped yml and asserts the key lands on
  `UpdateInfo` (would guard against a future electron-updater upgrade stripping
  unknown keys, which is a fail-open degradation with zero test signal today);
  likewise the stamping wiring in `scripts/electron-build.mjs` is untested (only
  the `stampChannelFeedFiles` helper is).
- Stale feed files from a previous build in an uncleaned `release/` dir can be
  re-stamped with the current origin (fail-safe at runtime, but the emitted feed
  then misdescribes the artifact it points at).
- The docs only cover the production direction: old pre-split **dev** installs
  (updater with no channel, hence `latest`) will auto-update onto
  production-origin builds; a "reinstall" note for dev testers would help.
- Consolidation: the `https://worldofclaudecraft.com` literal still exists in 3
  places even though the commit introduces `PRODUCTION_API_ORIGIN`; the
  `'latest' | 'dev'` union is re-declared in the builder config; the textual pin
  regex for `initUpdater({...})` in `tests/electron_updater_track.test.ts` is
  brittle (non-greedy match stops at the first `})`).

## Findings dismissed on second look

- "Renamed feed file crosses tracks": defeated by the stamp; covered by tests
  and docs.
- Channel case sensitivity, staging origins, trailing-slash and host-case origin
  variants, punycode homoglyphs, regex injection via the channel: all handled or
  impossible (channel is allowlisted before reaching the RegExp).
- Double `downloadUpdate()`: deduped in-flight by electron-updater's
  `downloadPromise`; a rejected download is caught and logged, no unhandled
  rejection (pinned by a test).
- Steam builds: publish stays nulled, updater off, and the dev-origin Steam case
  is explicitly tested not to trip the production-channel throw.
- Old production installs in the field: keep reading `latest*.yml`, ignore the
  unknown `wocApiOrigin` key, and update normally.

FernandoX7:
Thanks for the thorough review. All three should-fix items checked out as real and are addressed in bdbdf5103, along with the nice-to-haves:

1. **Origin asymmetry**: confirmed. A schemeless origin like `localhost:8787` derives the dev channel at build time but `deriveOrigin` falls back to production at runtime, so that install would refuse every stamped update on its own track. `desktopBuilderConfig` now throws on any non-empty unparseable origin when a publish feed is active (empty stays fail-safe: it cannot come through `electron-build.mjs`, which always resolves one). Pinned for schemeless, host-only, and garbage inputs, plus a pin that steam builds do not trip it.
2. **`WOC_UPDATE_CHANNEL=''`**: confirmed and fixed with `||` in both the script and the pure function, so set-but-empty derives from the origin. Applied the same treatment to set-but-empty `VITE_DESKTOP_API_ORIGIN` / `VITE_DESKTOP_LOGIN_ORIGIN`, which had the same class of problem (an empty origin would have died late, at the stamping step). Pinned.
3. **Doc globs**: fixed the three `latest-*.yml` / `dev-*.yml` spellings to `latest*.yml` / `dev*.yml` and spelled out the Windows names in the publish step.

Nice-to-haves, also applied:

- Added a fixture test that round-trips a stamped feed file through the real `electron-updater` `parseUpdateInfo`, so an upgrade that starts stripping unknown keys fails a test instead of silently degrading the guard to accept-everything.
- `electron-build.mjs` now deletes both channels' feed files from the output dir before electron-builder runs, so a stale yml from a differently-baked build cannot be re-stamped with the current origin.
- Runbook note added for pre-split dev installs (they read `latest` and will update onto production; dev testers need a fresh post-split build).
- Consolidation: `PRODUCTION_API_ORIGIN` now backs the fallbacks in `desktop_config.cjs` and `main.cjs`, the `'latest' | 'dev'` union is imported from the guard's declarations instead of re-declared, and the textual `initUpdater` pin now matches to the call's own `});` closer.

On the two VERIFY items: agreed both gate the release rather than the merge. The packaged smoke test is in the post-release checklist (section 5, including the `updateChannel` banner line and the REFUSED log), and the one-time host cleanup is now an explicit numbered step in the publish runbook since the guard deliberately accepts unstamped pre-split feeds.

103 tests across the 8 electron suites, `tsc --noEmit`, and biome are green at the new head.

madmatah (approved):

PR #1548, opened 2026-07-06
feat(ui): controller onboarding overlay on first connect

daxdax89:
## Summary

Adds a controls onboarding overlay for controller players. The first time a controller connects (persisted, so it never nags again) a dismissible overlay shows the current pad layout; a **Show Controls** button in Options > Controller reopens it any time.

The overlay lists the two sticks (Move / Look) and every bound button, labeled with the **detected pad brand's glyphs** (Cross/Circle/Square/Triangle on a DualSense, A/B/X/Y on an Xbox pad, etc.) and the **live action names**, including whatever ability/item currently sits in each action-bar slot. It is built as a pure `controllerGuideModel` core (host-agnostic, unit-tested) plus a thin modal painter modeled on the existing confirm dialog, so it meets the same focus-trap / Esc / return-focus contract as the other HUD modals.

Part of a small set of controller PRs (camera-feel and left-stick steering are the sibling PRs); each is independent.

## Related issues

N/A.

## Type of change

- [x] Feature: new functionality

## How was this tested?

- Commands: `npx vitest run tests/controller_guide_view.test.ts tests/architecture.test.ts`, `npx tsc --noEmit`, `npm run build`.
- Manual steps: drove the offline world with a real DualSense; on first connect the overlay auto-showed with correct PlayStation glyphs and the live action names; Show Controls reopens it; Esc / the close button / Got it dismiss it and return focus; re-connect does not re-nag once seen.

## Screenshots / recordings

Overlay on a first DualSense connect (offline world) - brand-correct PlayStation glyphs, live action names (note the warrior's slotted abilities on L1/R1):

### Quality
- [x] Tests pass; added `controller_guide_view` core tests (skip-unbound, order, brand pass-through) and registered the pure core in `UI_PURE_CORES`.
- [x] `npx tsc --noEmit` clean.
- [x] `npm run build` succeeds.

### Cross-platform
- [x] Modal is centered with `max-width: calc(100vw - 32px)` and reuses the standard button controls; no viewport scale-lock.
- [x] Accessible: `role=dialog` + `aria-modal` + `aria-labelledby`, focus trap on open, focus returned on close, Esc via the shared closeAll dispatcher.

### Localization (i18n)
- [x] New strings are `t()` keys added to `en` (`hud_chrome`); the one wordy label (Show Controls) ships its five non-Latin fills. Stick columns use non-localized hardware tokens (L-Stick / R-Stick) like the existing button glyphs.
- [x] Generated i18n regenerated via the build, not hand-edited.

[...]

PR #1547, opened 2026-07-06
feat(input): fix gamepad camera-follow snap-back, add auto-follow + invert-axis options

daxdax89:
## Summary

Three related, gameplay-neutral improvements to controller camera/movement feel, all surfaced in the existing Options > Controller panel:

1. **Fix the analog camera "snap-back".** The gamepad right stick had no equivalent of the mouse's mouselook/orbit signal, so while turning the camera *and* moving, the follow-camera's auto-settle fought the stick the whole time and swung the view back behind the character on release. The right stick now latches an "actively aiming" state (plus a short release grace) that the follow-camera treats like a mouse-orbit, so it no longer fights the stick mid-turn or snaps on release.
2. **Camera Auto-Follow toggle (default on).** On keeps the classic follow-cam, now without the fight; off keeps the camera exactly where the stick last aimed it while moving on the pad. Scoped to pad-sourced input, so keyboard/mouse follow is untouched.
3. **Invert-axis options.** Adds invert Camera-X (right-stick yaw) and invert Movement-X / Movement-Y (left stick), alongside the existing Camera-Y. Applied in the pure `stickToLook` / `stickToMoveFlags` core, so the default mapping is unchanged.

The follow-suppression composition is lifted into a small pure `gamepadCameraHold` helper (unit-tested), and the invert mappings are covered in `gamepad_map` tests. Part of a small set of controller PRs (onboarding overlay and left-stick steering are the sibling PRs); each is independent.

## Related issues

N/A.

## Type of change

- [x] Feature: new functionality

## How was this tested?

- Commands: `npx vitest run tests/gamepad_map.test.ts tests/camera_follow.test.ts tests/options_view.test.ts`, `npx tsc --noEmit`, `npm run build`.
- Manual steps: drove the offline world with a real DualSense; confirmed the camera no longer swings back on right-stick release while moving, the Auto-Follow toggle behaves both ways, and each invert flips the expected axis.

## Screenshots / recordings

N/A. The only new UI is four toggle rows in Options > Controller, reusing the existing toggle component; no visual layout change.

### Quality
- [x] Tests pass; added `gamepad_map` invert cases + `camera_follow` hold/composition tests.
- [x] `npx tsc --noEmit` clean.
- [x] `npm run build` succeeds.

### Localization (i18n)
- [x] New option labels are `t()` keys added to `en` (`hud_chrome`); the wordy ones ship their five non-Latin fills in this change. No `src/sim`/`server` strings.
[...]

PR #1549, opened 2026-07-06
feat(input): camera-relative left-stick steering for controller

daxdax89:
## Summary

Adds camera-relative left-stick steering for the controller (default on for the pad; toggle **Left Stick Steers Character** in Options > Controller).

Before this, the gamepad left stick strafed the character around a fixed heading and the pad had **no way to turn the character at all**, so orbiting the camera never re-oriented movement (stiff, non-console feel). With steering on, the left stick sets a target world facing from its direction plus the live camera yaw, and the frame loop turns the character toward it and runs forward once roughly aimed (a hard reversal pivots in place rather than arcing wide), mirroring the existing click-to-move facing easing. The right stick still aims the camera; the character only rotates while the stick is deflected. Flip the toggle off to keep the classic strafe.

The stick to facing-offset map is a pure `stickSteerOffset` core (unit-tested); the turn easing reuses `stepAngleToward` in `resolveMove`. The horizontal sign follows the game's camera-yaw convention (verified against the renderer's live `worldToScreen`: stick-left steers screen-left, right steers right, up stays forward). Part of a small set of controller PRs (camera-feel and onboarding are the sibling PRs); each is independent.

## Related issues

N/A.

## Type of change

## How was this tested?

- Commands: `npx vitest run tests/gamepad_steer.test.ts tests/gamepad_map.test.ts tests/options_view.test.ts`, `npx tsc --noEmit`, `npm run build`.
- Manual steps: drove the offline world with a real DualSense; confirmed the character turns to face the left-stick direction and runs that way, re-orients when the camera is orbited then forward is pushed, pivots in place on a hard reversal, and that toggling the option off restores the classic strafe. Verified the steer direction against the renderer's `worldToScreen` projection (stick-left -> screen-left).

## Screenshots / recordings

N/A. Movement-feel change; the only new UI is one toggle row in Options > Controller.

### Quality
- [x] Tests pass; added `gamepad_steer` core tests (deadzone null, up=forward, left/right signs, diagonal, magnitude-independent).
- [x] `npm run build` succeeds.

### Localization (i18n)
- [x] New label is a `t()` key added to `en` (`hud_chrome`) with its five non-Latin fills. No `src/sim`/`server` strings.
- [x] Generated i18n regenerated via the build, not hand-edited.

### Hygiene
- [x] No secrets/`.env`; `ALLOW_DEV_COMMANDS` untouched.


daxdax89:
Closing this to keep my open set small and focused. The three controller PRs (#1547, #1548, #1549) have gone a while without review, and I would rather not leave unsolicited feature work adding noise to your queue. If controller UX is something you would like to pick up later, I am glad to reopen or consolidate the three into a single cohesive PR. Thanks for the project.

PR #1550, opened 2026-07-06
feat(server): apply out-of-game character grants at load

awidearray:
## What

Adds a sanctioned write path for out-of-game character progression. An external system (a companion app, live event, or admin tool) appends rows to a new `character_grants` table; the game server applies any unconsumed grants to the character at load time (before it enters the sim) and stamps `applied_at`.

## Why

This lets partners, talent programs, and promos credit progression (e.g. XP) to a character without ever racing the authoritative sim save, and without touching game internals. It is the server seam the growth features (character grants pipeline, featured talent) build on.

## How it stays safe

- Applied at LOAD time, before the character enters the sim, so it never races `saveCharacterState`.
- Append-only and idempotent via a uniquely-indexed `source_key`: the same grant can never be applied twice.
- No saved state yet? Grants stay pending and apply after the first real save (transaction rolls back, leaving them unconsumed).
- Best-effort on join: a failure applying grants logs and never blocks the player from entering the world.

## Changes

- `server/db.ts`: `character_grants` schema + `applyPendingCharacterGrants()` (row-locked, transactional, XP kind adds to `state.xp`).
- `server/ws_auth.ts`: injected `applyPendingCharacterGrants` dep, invoked on join BEFORE the character row is read.
- `server/main.ts`: wires the dep into `createWsAuth`.
- `tests/server/ws_auth.test.ts`: ordering (grants applied before load) + best-effort (a failing grant never blocks join) coverage.
- `scripts/e2e-character-grants.mjs`: live REST+WS e2e (register, play to persist, write grant, re-login, assert applied once, idempotent on re-login).

## Verification

- `npx vitest run tests/server/ws_auth.test.ts`: 26 passed.
- `npx tsc --noEmit`: clean.
- `npx biome check` on the touched files: clean (only pre-existing `any` warnings).
- Live e2e script drives the real server path (see script header for how to run with a DB up).

## Notes for the reviewer

- The table is append-only by contract; there is no in-repo writer endpoint by design (the external system writes directly with its own idempotency key). If you want the writer surfaced as an admin route, that is a follow-up.
- Only `kind = 'xp'` is wired to a state mutation today; other kinds are recorded and stamped but apply no state change until their handler is added.


awidearray:
Test evidence (rebased onto release/v0.23.0, zero conflicts after grafting the 2 original commits onto the moved base):

```
$ npx vitest run tests/server/ws_auth.test.ts
 Test Files  1 passed (1)
      Tests  26 passed (26)

$ npx tsc --noEmit
(clean)

$ npx biome check server/db.ts server/ws_auth.ts server/main.ts tests/server/ws_auth.test.ts scripts/e2e-character-grants.mjs
(clean; only pre-existing 'any' warnings, no errors)
```

The two new ws_auth cases assert the load-bearing ordering (grants applied before the character row is read) and the best-effort contract (a thrown grant error logs and never blocks join). The rebase moved the base ~3.7k commits; the on-join hook now lives in the extracted `server/ws_auth.ts` (it was inline in main.ts at the original cut) and is injected as a dep.

awidearray:
Reviewer confirm (one thing): the `character_grants` write path is append-only and only `kind='xp'` mutates state today. Other kinds are RECORDED and stamped `applied_at` but apply no state change until a per-kind handler is added. Confirm that matches your intent for the initial surface.

Verification honesty: what CI runs here is the UNIT path, tests/server/ws_auth.test.ts (26 passed), which covers the injected `applyPendingCharacterGrants` ordering (applied before the character row is read) and the best-effort contract (a thrown grant error logs and never blocks join). The live end-to-end script scripts/e2e-character-grants.mjs (register -> play -> grant -> re-login -> assert applied once, idempotent) is provided and runnable but requires a running server + Postgres; it is NOT executed by CI and I have not claimed a live run. tsc + biome on the touched files are clean.

PR #1563, opened 2026-07-07
fix(ui): open bags on the first press after loading in

### LINKED ISSUE #1538: Bags window takes two clicks or two presses of B to open after loading in
## Problem

Opening the bags for the first time after loading into the world takes two activations instead of one: the first press of B (or the first click of the minimap bag button, or the mobile bags button) does nothing visible, and only the second press opens the window. The first press also plays the bag-close sound even though no window was open.

All three open paths (the B keybind, the minimap bag button, the mobile bags button) funnel into the same toggle, so they all show the same behavior.

## Steps to reproduce

1. Load into the world on a fresh page load. Do not open a vendor or the bags first.
2. Press B once (or click the minimap bag button once).
3. Observe: the bags window does not open, and the bag-close sound plays.
4. Press B again: the window opens.

## Actual behavior

- The first activation after a page load runs the close branch against an already-closed window, so the bags only open on the second press.
- The bag-close audio cue plays on that first phantom press.
- After the first phantom toggle the window behaves normally (one press per action) until the next page load, so the bug re-appears at the start of every session.

## Expected behavior

- A single press of B, or a single click of the bag button (desktop minimap or mobile), opens the bags window, including the very first activation after loading in.
- The bag-close sound only plays when an open bags window actually closes.
- The desktop and mobile experiences behave the same.

## Scope

- `Hud.toggleBags()` in `src/ui/hud.ts` (around line 9763), the shared open/close decision for all three entry points.
- Audit the other `$('#bags').style.display !== 'none'` guards in `hud.ts` (renderBags / inventory-change / Escape paths): they misread the same never-opened state and treat a stylesheet-hidden window as open.

## Acceptance criteria

- On a fresh load, the first press of B opens the bags window; no double press is needed on any entry point (keybind, minimap button, mobile button).
- No bag-close sound plays while the bags window is closed.
- Repeated presses alternate open and close, one press per action.
- The desktop experience and the mobile experience are both single-press.

## Notes

Likely root cause, from reading the code:

- `#bags` ships with no inline style (`index.html` line 372, `play.html` line 321). Its initial hidden state comes from the stylesheet rule `.window { display: none }` in `src/styles/layout.css`.
- `toggleBags()` branches on the inline style: `if (el.style.display !== 'none')` close, else open. On a fresh load the inline value is the empty string, which is not `'none'`, so the first activation takes the close branch: it plays `audio.bagClose()` and calls `BagsWindow.close()`, which stamps inline `display: 'none'`. The second activation then reads `'none'` and opens.
- Sibling windows use a positive check and do not have this bug: the character window tests `style.display === 'block'` (`src/ui/char_window.ts`), the map tests `=== 'block'` (`toggleMap`). `hud.ts` also already has an `isWindowVisible()` helper that reads `getComputedStyle`. Either pattern (a positive shown-value check or the computed-style helper) would fix the toggle.

daxdax89:
## Problem

Opening the bags for the first time after loading into the world takes two activations: the first press of B (or the first click of the minimap bag button, or the mobile bags button) does nothing visible and plays the bag-close sound, and only the second press opens the window. All three open paths funnel through `toggleBags`, so all three show it. After the first phantom toggle it behaves normally until the next page load.

Fixes #1538.

## Root cause

`toggleBags` (`src/ui/hud.ts`) decided open-vs-close from the inline `style.display`:

```ts
if (el.style.display !== 'none') { audio.bagClose(); this.bagsWindow.close(); return; }
```

But `#bags` carries no inline style; `.window` is hidden purely by a stylesheet rule (`src/styles/layout.css`). So on a fresh load the inline `display` is the empty string `''`, not `'none'`. The `!== 'none'` check treats `''` as shown, so the first press runs the close branch: nothing opens and `audio.bagClose()` plays. `close()` also guards on `=== 'none'`, so it does not early-return either; it sets the inline `display` to `'none'`, and only then does the second press open.

## Fix

Read visibility through a small pure `bagsWindowShown()` core in `bags_view.ts` that treats only genuinely shown values as open (guarding on the hidden values `'none'`/`''` rather than a specific shown value, mirroring `close()`'s own guard and the existing `closeAll` empty-string precedent). A cold-load `''` now reads as not-shown, so the first activation opens and the close cue plays only when a genuinely open window closes.

## Test

`tests/bags_view.test.ts` pins the regression: `bagsWindowShown('')` is `false` (the cold-load state that must open), `'none'` is `false`, and `'flex'`/`'block'` are `true`.

## Scope

- `src/ui/bags_view.ts` (new pure `bagsWindowShown`), `src/ui/hud.ts` (`toggleBags` consumes it), `tests/bags_view.test.ts`.
- Client HUD chrome only; no `IWorld`/parity/determinism/i18n surface touched.

### DIFF
--- a/src/ui/bags_view.ts
+++ b/src/ui/bags_view.ts
@@ -94,6 +94,18 @@ export function bagShiftLinks(mode: BagMode): boolean {
   return !mode.vendorOpen;
 }
 
+/** Whether the #bags window is currently shown, from its inline display value.
+ *  Shown = any non-hidden value (today 'flex'; the close() guard also allows a
+ *  'block' pet-feed path), NOT hidden ('none') and NOT the cold-load empty ''.
+ *  The '' case is the bug (issue #1538): the .window stylesheet rule hides the
+ *  window, so a fresh page load leaves the inline display '' (never 'none'). It
+ *  must read as NOT shown so the first toggle OPENS instead of taking the close
+ *  branch (and playing the close sound) against an already-hidden window. Guards
+ *  on the hidden values rather than a specific shown value, mirroring close(). */
+export function bagsWindowShown(display: string): boolean {
+  return display !== 'none' && display !== '';
+}
+
 /** The tooltip hint sub-line for a bag item, matching the original tooltip's
  *  mode-then-kind branch. Returns '' when no hint applies (e.g. a material). */
 export function bagTooltipHintKey(item: BagItemInfo, mode: BagMode): BagTooltipHintKey {
--- a/src/ui/bags_view.ts
+++ b/src/ui/bags_view.ts
@@ -95,13 +95,14 @@ export function bagShiftLinks(mode: BagMode): boolean {
 }
 
 /** Whether the #bags window is currently shown, from its inline display value.
- *  Shown = any non-hidden value (today 'flex'; the close() guard also allows a
- *  'block' pet-feed path), NOT hidden ('none') and NOT the cold-load empty ''.
- *  The '' case is the bug (issue #1538): the .window stylesheet rule hides the
- *  window, so a fresh page load leaves the inline display '' (never 'none'). It
- *  must read as NOT shown so the first toggle OPENS instead of taking the close
- *  branch (and playing the close sound) against an already-hidden window. Guards
- *  on the hidden values rather than a specific shown value, mirroring close(). */
+ *  Shown = any non-hidden value (#bags is only ever assigned 'flex' today), NOT
+ *  hidden ('none') and NOT the cold-load empty ''. The '' case is the bug (issue
+ *  #1538): the .window stylesheet rule hides the window, so a fresh page load
+ *  leaves the inline display '' (never 'none'). It must read as NOT shown so the
+ *  first toggle OPENS instead of taking the close branch (and playing the close
+ *  sound) against an already-hidden window. Guards on the hidden values rather
+ *  than pinning to a specific shown value (mirroring close() and the closeAll
+ *  precedent), so it stays correct if the shown value ever changes. */
 export function bagsWindowShown(display: string): boolean {
   return display !== 'none' && display !== '';
 }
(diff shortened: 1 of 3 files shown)


FernandoX7 (commented):
Clean fix, and the root-cause writeup matches the code exactly. `.window` is hidden by the stylesheet (`src/styles/layout.css:31`), `#bags` carries it in both `index.html:373` and `play.html:322`, so a cold-load inline `display` is `''` and the old `!== 'none'` check read that as shown. Nice that `bagsWindowShown` guards on the hidden values instead of pinning to `=== 'flex'`: that mirrors the existing `closeAll` guard (`src/ui/hud.ts:12288` uses the same `!== 'none' && !== ''`) and `close()`'s own shape, so it stays consistent if the shown value ever changes. Pure core in `bags_view.ts` with a thin consumer is the right pattern, and the test pins the four meaningful inputs decisively. I confirmed all three open paths (keybind, minimap `#mm-bag`, mobile `onBags`) funnel through `toggleBags`, and the head test passes.

One nit:

- nit: the `bagsWindowShown` docstring (`src/ui/bags_view.ts`) and the test comment (`tests/bags_view.test.ts`) both reference a `'block'` pet-feed path, but the live pet-feed handler sets `'flex'` (`src/ui/hud.ts:5116`), and `#bags` is only ever assigned `'flex'` or `'none'` anywhere in the tree. That `'block'` note is echoing a stale comment in `close()` (`src/ui/bags_window.ts:167`). Guarding on the hidden values rather than `=== 'flex'` is still the correct, more robust choice, so no code change is needed here; you might just drop or soften the `'block'` reference so it does not send a reader hunting for a path that is not there.

daxdax89:
Thanks, you're right, the `'block'` note was echoing the stale `close()` comment; pet-feed sets `'flex'` and `#bags` is only ever `'flex'`/`'none'`. Kept the value-agnostic guard (checks the hidden values, not a specific shown value, per your point about staying correct if the shown value changes) and dropped the stale `'block'` rationale from the docstring + test comment.

FernandoX7 (approved):

PR #1564, opened 2026-07-07
fix(ui): clear the stale hotbar tooltip after a drag-drop

### LINKED ISSUE #1485: Bug: Hotbar shows "empty slot" tooltip after dragging a spell
### Before you start

### What happened?

When dragging a spell from one slot to an empty slot on the same hotbar, the tooltip "empty slot" shows, even though after releasing the spell, my cursor hovers over the spell in the hotbar.
Actually, the tooltip doesn't update properly as per my screenshot. It says "hoarfrost mantle" but it should say "breadbind". So not only with empty slot, but switching spells around the hotbar too.

### Steps to reproduce

1. Have at least one free slot on the toolbar and one spell
2. Click and hold the cursor onto a spell to start dragging
3. Drag the spell to the right or left, landing it onto an empty slot
4. Release the spell so that it is dropped onto the empty slot in the hotbar
5. Observe the tooltip shows "empty slot", even though the cursor is on the spell

When I release the spell, I expect the tooltip to show the spell description.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on Linux Mint

Image

daxdax89:
## Problem

Dragging a hotbar ability onto another slot on the same bar left the tooltip stale. Dropped on an empty slot it kept showing "empty slot" even though the cursor now sits on the moved ability; after swapping two abilities it showed the wrong ability's name. The tooltip did not reflect what is now under the cursor.

Fixes #1485.

## Root cause

The tooltip content is a live closure that recomputes the slot's ability/item on each run, so it is correct; the defect is refresh *timing*. `attachTooltip` re-runs that closure only on `mouseenter`/`focusin` (or a touch long-press); `mousemove` merely repositions the box. Native HTML5 drag-and-drop suppresses synthetic mouse events during a drag, and a drop that ends with the cursor already inside the target slot crosses no element boundary, so no `mouseenter` fires and the closure is never re-run for the new slot content.

The two hotbar drop-completion paths (`src/ui/hud.ts`: the desktop `drop` handler and the mobile `finish` handler) rearranged the slot but never touched the tooltip. Every sibling slot mutation (`clearSlot`, the shift-right-click clear, the char-window and bags-window drops) already calls `hideTooltip()` on mutate; these two diverged.

## Fix

Add `this.hideTooltip()` after the slot map is saved on both drop paths, matching the established convention. The stale text is dismissed on drop, and the next hover resolves the slot's tooltip live.

## Test

`tests/hotbar_drop_tooltip.test.ts` guards that both drop-completion handlers call `hideTooltip()` after the slot mutation (source scan, matching the repo's existing hud.ts behavior-guard style, e.g. `daily_rewards_window_chest_toggle.test.ts`).

## Scope

- `src/ui/hud.ts` (two one-line additions), `tests/hotbar_drop_tooltip.test.ts`.
- Client HUD only; no new strings, no per-frame/render-loop code (the change is in the `drop`/`finish` event handlers), no determinism/parity surface.

### DIFF
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -4636,6 +4636,11 @@ export class Hud {
             this.hotbarActions = placeItemOnSlot(this.hotbarActions, action.id, slot - 1);
           }
           this.saveSlotMap();
+          // The drop rearranged this slot's contents, but a drop that ends with the
+          // cursor already inside the slot fires no mouseenter, so the tooltip would
+          // keep the pre-drop text (stale "empty slot" / wrong ability). Clear it so
+          // it no longer shows the old slot; the next hover resolves it live (#1485).
+          this.hideTooltip();
         });
         btn.addEventListener('dragend', () => {
           this.dragAction = null;
@@ -4813,6 +4818,9 @@ export class Hud {
         if (targetIndex !== null && targetIndex !== drag.sourceIndex) {
           this.hotbarActions = swapHotbarSlots(this.hotbarActions, drag.sourceIndex, targetIndex);
           this.saveSlotMap();
+          // Match the desktop drop: clear the now-stale tooltip for the rearranged
+          // slot so a long-press peek resolves the new content (#1485).
+          this.hideTooltip();
         }
       }
       this.clearMobileHotbarDrag();
--- a/tests/hotbar_drop_tooltip.test.ts
+++ b/tests/hotbar_drop_tooltip.test.ts
@@ -1,0 +1,42 @@
+// Regression for a reported bug (#1485): dragging a hotbar ability onto another
+// slot left the tooltip stale. A drop that ends with the cursor already inside the
+// target slot fires no mouseenter, so the tooltip kept its pre-drop text (the
+// "empty slot" hint, or the previous ability's name after a swap). Every sibling
+// slot mutation (clearSlot, the context-menu clear, char/bags window drops) already
+// calls hideTooltip() on mutate; the two hotbar drop-completion paths did not.
+// Guard that both now clear the tooltip after the slot is rearranged.
+
+import { readFileSync } from 'node:fs';
+import { describe, expect, it } from 'vitest';
+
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Clean, well-diagnosed fix. The root-cause writeup is exactly right: native HTML5 drag-and-drop suppresses the synthetic mouse events, and a drop that ends with the cursor already inside the target slot crosses no element boundary, so `attachTooltip`'s `mouseenter` closure never re-runs and the box keeps its pre-drop text. Calling `hideTooltip()` after `saveSlotMap()` on both completion paths matches the convention every other slot mutation already follows (`clearSlot`, the shift-clear, the char/bags-window drops), so the two hotbar paths stop being the odd ones out.

Verified against the head ref:

- Both calls land after the slot map is saved (`hud.ts:4643` desktop drop, `hud.ts:4823` mobile finish), and `hideTooltip()` is a cheap idempotent `display:none`, so it is safe even on a no-op drop.
- Audited every `saveSlotMap()` call site: the rest are programmatic (add/remove/reset/form-sync/loadout restore), not hover-in-place drop completions, so "these two were the only paths missing it" holds.
- The source-scan test is decisive. I replicated its block isolation and it passes at head and fails against base on both the desktop and mobile assertions. The two anchors (`placeItemOnSlot(this.hotbarActions`, `swapHotbarSlots(this.hotbarActions, drag.sourceIndex, targetIndex)`) each occur exactly once, so it isolates the right block, and stripping comments before the scan keeps the explanatory comment from satisfying it.

One note, not a change request: hiding rather than repainting means the corrected tooltip reappears on the next hover-in, not instantly under a resting cursor. That is the established hide-on-mutate behavior everywhere else here, so matching it is the right call; an immediate re-resolve would be a larger change and out of scope.

Nothing blocking. Client HUD only, no new strings, no determinism or parity surface.

FernandoX7 (approved):

PR #1565, opened 2026-07-07
feat(ui): show the game version at the foot of the settings menu

### LINKED ISSUE #1541: Show the game version at the bottom of the settings UI
## Problem

The in-game settings window does not show the running game version. Players who want to report a bug or confirm they are on the latest build have no in-context way to see which version they are running. The version is only rendered in the page-level footer, which is not visible while the settings window is open.

## Expected behavior

- The settings window shows the current game version at the bottom of the panel.
- The value matches the build the client is actually running (the `__APP_VERSION__` and `__APP_BUILD_ID__` compile-time constants already used elsewhere).
- The label is unobtrusive: small, secondary text at the foot of the settings content, not competing with the settings controls.

## Scope

- Settings window UI: `src/ui/options_window.ts` and `src/ui/options_view.ts`.
- Reuse the existing version source and formatter rather than adding a new one: `__APP_VERSION__` / `__APP_BUILD_ID__` and `formatFooterVersion` (currently in `src/main.ts:412`). If that helper needs to be shared, lift it into a small sibling module both call sites import.
- Player-visible string: add the label as a `t()` key (settings domain), with the version and build id passed as values, per the i18n invariant.

## Acceptance criteria

- Opening the settings window shows the version (and build id) as secondary text at the bottom of the panel.
- The displayed version matches the deployed build.
- The label text is a `t()` key, and the version/build values are interpolated, not concatenated into English.
- The desktop and mobile settings layouts are both unchanged apart from the added line, with no overflow or clipping.

daxdax89:
## Problem

The in-game settings window did not show the running game version. The version is only rendered in the page-level footer, which is not visible while the settings window is open, so players who want to report a bug or confirm they are on the latest build had no in-context way to see which version they run.

Closes #1541.

## Fix

Show the version + build id as small secondary text at the foot of the game menu (Esc settings window), matching the acceptance criteria (unobtrusive, not competing with the settings controls).

- Lift the version source and `formatFooterVersion` out of `src/main.ts` into a shared `src/ui/app_version.ts`. `appVersionInfo()` returns the formatted `version` + `build`, so the page footer and the settings panel read one source instead of each re-declaring the `__APP_*` globals.
- `options_window.renderMain()` appends an `.opt-version` line (muted/small/centered, styled in `components.css`, no inline values) rendered through a new `t('hudChrome.options.version')` key with `version` and `build` passed as values, per the i18n invariant.
- `appVersionInfo()` reads the Vite-defined globals through a `typeof` guard so it never throws when they are absent (e.g. the standalone browser-test config injects no defines); Vite textually replaces the identifiers in the real build.

## Test

- `tests/app_version.test.ts`: pins `formatFooterVersion` (trailing `.0` trim), and that `appVersionInfo()` does not throw and returns strings when the globals are absent.
- `tests/options_window.test.ts`: guards that `renderMain` paints the version as the `t('hudChrome.options.version', { version, build })` label via the shared `appVersionInfo` source, appended after the button list.

## Scope

- New `src/ui/app_version.ts`; `src/main.ts` (imports the lifted formatter), `src/ui/options_window.ts` (renders the line), `src/styles/components.css` (`.opt-version`), `src/ui/i18n.catalog/hud_chrome.ts` (new English key) + regenerated i18n tables.
- English-only key (`'v{version} ({build})'`, not wordy, so no non-Latin fills needed). Client HUD only; no determinism/parity surface.


FernandoX7 (commented):
Clean, well-scoped change. The nice part is that it does not just paste the version into the settings window: it lifts the version source and `formatFooterVersion` out of `main.ts` into a shared `src/ui/app_version.ts` so the footer and the settings panel read one source instead of each re-declaring the `__APP_*` globals. That is the right shape for this repo.

Checked the invariants in play and they hold:

- i18n: `version: 'v{version} ({build})'` is an English key in `hud_chrome` (the English-only catalog domain) and renders through `t('hudChrome.options.version', { version, build })` with the values passed in, no concat. Good.
- play.html parity: the `.opt-version` rule is in the shared `src/styles/components.css`, which both `index.html` and `play.html` load through the `src/main.ts` styles barrel, so the settings line styles the same on `/play`. No inline-CSS gap.
- The `typeof __APP_VERSION__ !== 'undefined'` guard in `appVersionInfo()` is the correct pattern: it falls back cleanly in the non-Vite browser-test config and is textually replaced in the real build.
- No sim/server/wire surface, client HUD only, so no determinism or parity concern.
- The regenerated i18n trio is well-formed (`i18n.resolved.sha256` is a single line, `i18n.status.summary.json` parses, the key is in generated `en`). The `pending` rows for the other locales are the expected release-time fill.

Tests look right: `tests/app_version.test.ts` pins the `.0`-trim and the no-throw fallback, and the `options_window.test.ts` guards confirm the line renders from the shared source as a `t()` label appended after the button list.

Nothing blocking. Nice work.

daxdax89:
Thanks for the thorough invariant check. Nothing to change here; glad the shared `app_version` shape landed well.

FernandoX7 (approved):

PR #1567, opened 2026-07-07
fix(ui): apply the UI scale slider on release, not live while dragging

### LINKED ISSUE #1558: Bug: UI changes while sliding the UI scale slider in options menu
### Before you start

### What happened?

When sliding the UI scale slider in the ui, the UI starts scaling as the number changes. This causes the window to flicker and resize very fast because the slider is jumping around along with the resizing window. It's also hard to get the right ui scale because of the slider jumping around while dragging.

### Steps to reproduce

1. Open options menu
2. Open the interface menu
3. Click and hold the ui scale slider
4. Move the slider and observe the window being resized, forcing additional movement to the slider and causing flickering.

I expected the UI to rescale only when the slider is released after dragging.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on Linux

_No response_

daxdax89:
## Problem

Dragging the UI scale slider in Options > Interface rescaled the whole UI live on every value change. Because `#ui` is scaled with `zoom: var(--ui-scale)` and the options window is a child of `#ui`, the window resized under the cursor mid-drag, so the slider jumped and flickered and it was hard to land a target scale.

Closes #1558.

## Fix

Give the slider descriptor an opt-in `commitOnChange` flag, set only for `uiScale`. The other comfort sliders (frame scales, opacity, tooltip/FCT/chat scales) keep their intended live preview.

When `commitOnChange` is set, `settingSlider` previews the readout + gold fill on `input` but commits the setting on `change`, so the UI rescales once on release instead of on every drag tick:
- `input` (drag, and each keyboard step) updates only the readout (from the raw slider position, since the store is not written until commit) and the fill.
- `change` (pointer release / touchend, and each keyboard step, which fires both events) is the one that calls `onSettingChange`.

Keyboard arrow adjustment and touch stay fully functional (both commit per step / on release), and `aria-valuetext` still updates live while dragging.

## Test

- `tests/options_view.test.ts`: `uiScale` carries `commitOnChange: true`; the sibling sliders do not (so their live preview is preserved).
- `tests/options_window.test.ts`: the commit-on-change branch previews on `input` without committing and commits on `change`; a normal slider still commits live on `input`.

## Scope

- `src/ui/options_view.ts` (descriptor flag + set it on `uiScale`), `src/ui/options_window.ts` (`settingSlider` branch), tests.
- Pure cosmetic UI scale (fairness invariant not implicated); no new strings; client HUD only.

### DIFF
--- a/src/ui/options_view.ts
+++ b/src/ui/options_view.ts
@@ -40,6 +40,11 @@ export interface SliderControl {
   /** Current value at build time; the painter re-reads the live value on input. */
   value: number;
   fmt: SliderFmt;
+  /** Commit the setting on release ('change') instead of live on every 'input'
+   *  tick. Set for uiScale, whose live rescale moves the slider under the cursor
+   *  mid-drag (issue 1558); dragging updates only the readout, not the setting.
+   *  Other sliders keep their intended live preview (volume, fov, frame scale). */
+  commitOnChange?: boolean;
 }
 
 export interface ToggleControl {
@@ -357,7 +362,10 @@ export function buildControllerControls(s: OptionsSettingsSource): OptionsContro
 
 export function buildInterfaceControls(s: OptionsSettingsSource): OptionsControl[] {
   return [
-    slider(s, 'uiScale', 'hudChrome.options.uiScale'),
+    // uiScale commits on release: applying it live rescales the whole UI (the
+    // options window included), which shoves the slider under the cursor and makes
+    // the value hard to land (issue 1558).
+    { ...slider(s, 'uiScale', 'hudChrome.options.uiScale'), commitOnChange: true },
     slider(s, 'playerFrameScale', 'hudChrome.options.playerFrameScale'),
     slider(s, 'targetFrameScale', 'hudChrome.options.targetFrameScale'),
     slider(s, 'hudOpacity', 'hud.options.hudOpacity'),
--- a/src/ui/options_window.ts
+++ b/src/ui/options_window.ts
@@ -490,6 +490,12 @@ export class OptionsWindow {
       );
     };
     paintFill();
+    // Commit the setting (from the raw slider value), then sync the readout + fill.
+    const commit = () => {
+      hooks.onSettingChange(key, sliderDispatchValue(slider.value));
+      syncReadout();
+      paintFill();
+    };
     if (c.commitOnChange) {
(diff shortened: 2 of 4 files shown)


FernandoX7 (commented):
Clean, correctly scoped fix. The diagnosis is right (the options window is a child of `#ui`, so `zoom: var(--ui-scale)` rescales it under the cursor mid-drag), and gating the live apply behind an opt-in `commitOnChange` flag set only on `uiScale` keeps every other comfort slider's live preview intact.

I traced the event flow and it holds up:
- `change` fires after `input` in every interaction mode (mouse release, touchend, each keyboard step, and a track click), so the setting always commits; nothing leaves it stuck on the pre-drag value.
- `uiScale`'s only live consumer is `applySetting` writing `--ui-scale` (`main.ts:1516`), so deferring the commit doesn't strand any other subsystem mid-drag.
- No readout snap on release: `sliderDispatchValue` is `Number(slider.value)` and `applySetting` clamps to the same `[0.85, 1.4]` range/step the slider already enforces, so `readoutFromSlider()` during drag and `syncReadout()` on commit format the same number. `aria-valuetext` stays live via `readoutFromSlider`.

Both touched test files pass locally (36 tests). Fairness invariant isn't implicated (cosmetic scale), and no new strings.

Two nits, both optional:

- nit `src/ui/options_window.ts:465`: the commit body (`onSettingChange` + `syncReadout` + `paintFill`) is now duplicated across the commit-on-change `change` handler and the `else` `input` handler. A small `const commit = () => {...}` closure would DRY it. Borderline on the rule of three (two copies), so take it or leave it.
- nit `tests/options_window.test.ts`: the new guards pin behavior by string-matching the painter source. That is decisive against the specific regression (a commit sneaking back onto `input`) and matches this file's existing source-scan style, but it couples to exact source text, so the closure refactor above would break the match with behavior unchanged. A JSDOM test that dispatches `input` then `change` and asserts `onSettingChange` fires only on the latter would pin the behavior more robustly.

daxdax89:
Thanks. Took the DRY nit: extracted a single `commit` closure that both the commit-on-change `change` handler and the plain `input` handler wire, so the commit path has one definition. (Left the keyboard/touch behavior exactly as reviewed.)

FernandoX7 (approved):

PR #1568, opened 2026-07-07
fix(ui): scale the world map with the UI scale setting

### LINKED ISSUE #1559: Bug: Map isn't scaling along with rest of UI
### Before you start

### What happened?

When putting the UI scale at 85% or 140%, it is clear that the map isn't scaling along with the rest of the UI.

### Steps to reproduce

1. Open the map
2. Open the options menu and go to interface
3. Put the UI scale at 85% and observe there is space remaining in the map window on the right side of the map
4. Put the UI scale at 140% and observe nearly half of the map is hidden behind the border of the map window

Have the map scale along with the window it's in when the UI scale changes so that it remains visible all the time no matter the ui scale.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox in Linux Mint

UI scale: 85%:

Image

UI scale: 100%:

Image

UI scale: 140%:

Image

My window resolution for reference:

Image

daxdax89:
## Problem

The world map does not scale with the UI-scale setting. At 85% the map underfills its window (empty space on the right); at 140% nearly half the map is hidden behind the window border. At 100% it looks correct.

Closes #1559.

## Root cause

`#ui` is magnified by `zoom: var(--ui-scale)` (`src/styles/hud.css`), so everything under it, including the map window, scales with the setting. But the desktop `#map-canvas` rule (`src/styles/components.css`) declared no CSS display size, so the canvas box was a replaced *intrinsic* size (the `560x560` backing store from `index.html`/`play.html`). A replaced intrinsic size does not scale in lockstep with the window frame's shrink-to-fit box under `zoom`, so the content and frame diverge: smaller than the frame below 100%, larger (and clipped by the window's `overflow`) above 100%. Mobile already handles this (`body.mobile-touch #map-canvas { width: 100% }`); desktop was the outlier.

## Fix

Give the desktop canvas an authored px display size matching the backing store (`width: 560px; height: 560px`). A CSS px length is zoom-scaled by every engine exactly like the window frame's padding/border, so the map fills its window at all UI scales. The `100%` look is byte-identical (560px display == the intrinsic 560), and the map stays 1:1 (no DPR/backing change, per the canvas contract). The mobile `width: 100%` override is untouched.

## Test

`tests/map_canvas_scale.test.ts`: the desktop `#map-canvas` declares an explicit display size, that size is pinned to the canvas backing resolution (so CSS and backing cannot silently diverge), and mobile keeps its fill-the-window override.

## Verification note

This is a CSS-only layout fix reasoned from the source (the desktop rule missing the size that mobile has, and the reporter's screenshot dimensions showing frame height tracking the scale while width stayed constant). A quick Firefox pass at 85/100/140% to confirm the map fills the frame is still worth doing before merge; happy to attach screenshots if useful.

## Scope

- `src/styles/components.css` (desktop `#map-canvas` size), `tests/map_canvas_scale.test.ts`. No painter/render-loop change (the map core is already scale-agnostic); no per-frame budget impact.

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -4157,7 +4157,12 @@
        past it above 100% (issue 1559). A px length zoom-scales exactly like the
        frame, so the map fills its window at every UI scale; the 100% look is
        unchanged. The mobile rule below overrides this with width 100% to fill the
-       responsive window. */
+       responsive window.
+       box-sizing: content-box so the 2px border sits OUTSIDE the 560 content box:
+       under the global border-box reset, a plain width:560px would make the content
+       box 556 and downscale the 560 backing (~0.7%). content-box keeps the content
+       box exactly 560, so the map stays a crisp 1:1 at 100%. */
+    box-sizing: content-box;
     width: 560px;
     height: 560px;
     border: 2px solid var(--border);
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -4150,6 +4150,16 @@
     align-items: center;
   }
   #map-canvas {
+    /* Authored display size matching the 560x560 backing store (index.html /
+       play.html). Without a CSS length the canvas box is a replaced intrinsic size,
+       which does NOT scale in lockstep with the window frame under the #ui
+       zoom var(--ui-scale): the map underfilled the frame below 100% and clipped
+       past it above 100% (issue 1559). A px length zoom-scales exactly like the
+       frame, so the map fills its window at every UI scale; the 100% look is
+       unchanged. The mobile rule below overrides this with width 100% to fill the
+       responsive window. */
+    width: 560px;
+    height: 560px;
     border: 2px solid var(--border);
     outline: 1px solid #000;
     border-radius: 4px;
--- a/tests/map_canvas_scale.test.ts
+++ b/tests/map_canvas_scale.test.ts
@@ -16,13 +16,17 @@ const root = join(dirname(fileURLToPath(import.meta.url)), '..');
 const componentsCss = readFileSync(join(root, 'src/styles/components.css'), 'utf8');
 const mobileCss = readFileSync(join(root, 'src/styles/hud.mobile.css'), 'utf8');
 const indexHtml = readFileSync(join(root, 'index.html'), 'utf8');
+const playHtml = readFileSync(join(root, 'play.html'), 'utf8');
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Nice diagnosis. The core mechanism is right: a replaced canvas sized by its intrinsic 560 does not scale in lockstep with a length-sized frame under `zoom`, but a CSS px length does. The fix is minimal, the mobile `width: 100%` override still wins on specificity, and the painter reads `canvas.width` (the 560 backing) so render resolution and the click/hover mapping (the `getBoundingClientRect().width` ratio in `hud.ts`) are untouched. Good regression test too.

One correctness gap in the "byte-identical / 1:1" claim:

**should-fix** `src/styles/components.css:4161` (the new `width: 560px; height: 560px`). There is a universal `* { box-sizing: border-box }` reset (`src/styles/base.css:10`), so `width: 560px` is the border box. With the rule's `border: 2px solid`, the content box is 556x556 and the 560x560 backing store is downscaled into it (about 0.7%, and the window ends up 4px smaller). Before this PR the unsized canvas used its intrinsic 560 as the content box, so 100% was a crisp 1:1. So the "100% look is byte-identical" / "stays 1:1 (no stretch)" claims (PR body and the test comment) do not quite hold. The visible effect is small, but adding `box-sizing: content-box;` to the desktop rule makes the content box exactly 560 (byte-identical, crisp 1:1) and still zoom-scales in lockstep with the frame, so it satisfies both goals.

**nit** `tests/map_canvas_scale.test.ts:20` reads the backing width from `index.html` only, while the comment and the fix both cover `play.html`. Both declare 560 today so nothing is wrong, but pinning `play.html`'s backing as well would guard the parity the comment claims (a drift in only one entry would slip through).

daxdax89:
Good catch on the border-box box, thank you. Fixed: added `box-sizing: content-box` to the desktop `#map-canvas` rule so the content box is exactly 560 (crisp 1:1 at 100%) while still zoom-scaling with the frame, so both the 1:1 and the fill-the-frame goals hold. Also addressed the nit: the test now pins `play.html`'s backing width alongside `index.html`'s (and asserts they match), plus a `content-box` assertion.

FernandoX7 (approved):

PR #1569, opened 2026-07-07
feat(ui): destroy regular items from the bags window

### LINKED ISSUE #1501: Regular items cannot be destroyed from the bags window
## Problem

Players have no way to destroy a regular (non-quest) item from their bags. The whole backend path already exists (`IWorld.discardItem`, the `discard` client command, and `sim.discardItem`, which accepts any item that is not flagged `noDiscard`), but the bags UI only surfaces the destroy prompt for quest items: `bagClickAction` in `src/ui/bags_view.ts` returns `discardQuest` only when `item.kind === 'quest'`. Every other item's click means "use" (or equip for bags), and there is no context menu, drag-out, or other destroy affordance.

This matters because junk drops, outleveled soulbound gear, and unsellable items can never be removed. With no vendor nearby (or for items a vendor will not take), a full inventory can only be fixed by using or mailing things away, and some items allow neither.

## Steps to reproduce

1. Open the bags window with any regular (non-quest) item in the inventory, for example a gray junk drop.
2. Look for a way to destroy it: click it, right-click it, try to drag it out of the bag.
3. Observe there is no destroy option; clicking simply uses or equips the item.

## Expected behavior

- A player can destroy any regular item from the bags window, with a confirmation prompt (and a quantity field for stacks), matching the existing quest-item destroy flow.
- Items flagged `noDiscard` remain protected and cannot be destroyed; the player gets clear feedback instead of a silent no-op.
- Mode-specific click behaviors (trade, market sell, vendor sell, pet feed) keep working unchanged.

## Actual behavior

- Only quest items offer the destroy prompt (via plain click).
- Regular items have no destroy affordance at all: click is "use", and there is no alternative gesture or menu.

## Scope

- `src/ui/bags_view.ts`: extend the pure click-action core so regular items can reach the destroy flow (for example a modifier-click or per-slot affordance), without breaking the existing mode branches.
- `src/ui/bags_window.ts`: wire the new affordance to the existing `showDiscardItemPrompt`, which already calls `IWorld.discardItem`.
- `src/sim/items.ts` `discardItem`: currently returns silently for `noDiscard` items; it should emit the standard error feedback instead.
- Works both offline (`Sim`) and online (`ClientWorld`); the plumbing already exists on both, so this is UI plus feedback polish.

## Acceptance criteria

- A regular, destroyable item can be destroyed from the bags window on desktop and mobile touch, with a confirmation prompt, and a quantity prompt for stacks.
- Destroying removes the item(s) and logs the existing "Discarded ..." message in both offline and online play.
- A `noDiscard` item cannot be destroyed and the player receives visible feedback.
- Quest-item destroy, trade, market sell, vendor sell, and pet feed flows are unchanged.
- The desktop experience is unchanged except for the new destroy affordance.

daxdax89:
## Problem

Players had no way to destroy a regular (non-quest) item from their bags. The whole backend path already existed (`IWorld.discardItem`, the `discard` client command, and `sim.discardItem`, which accepts any item not flagged `noDiscard`), but the bags UI only surfaced the destroy prompt for quest items. Every other item's click means use/equip, so junk drops, outleveled soulbound gear, and unsellable items could never be removed, and a full inventory could get stuck with no vendor nearby.

Closes #1501.

## Fix

Add a right-click destroy affordance that reuses the existing quest-item destroy prompt (confirm + quantity field for stacks) verbatim, so nothing new is built for the prompt or the dispatch.

- New pure `bagDestroyAction(item, mode)` core in `bags_view.ts`: `none` in the transactional modes (trade / mail / market / vendor / pet-feed, whose own click/contextmenu owns the slot), `discardBlocked` for a `noDiscard` item (protected with feedback), `discard` otherwise. This mirrors the sim's `discardItem` rule.
- `bags_window.ts` extends the row `contextmenu` handler: at a vendor, Ctrl/Meta right-click still owns split-stack sell (unchanged); otherwise a right-click destroys via `showDiscardItemPrompt`, with `noDiscard` items rejected via `showError`.
- Left-click use/equip and shift-click chat-link are untouched.

## Server authority

No server or sim change. The server independently re-runs `sim.discardItem`, which enforces `noDiscard` regardless of the client, so authority is unaffected (verified: `server/game.ts` validates the `discard` command and calls `sim.discardItem`).

## Test

`tests/bags_view.test.ts`: `bagDestroyAction` destroys a regular item, `discardBlocked` for a `noDiscard` item, and `none` in every transactional mode (with a per-mode negative case), plus the existing Sim-vs-ClientWorld parity coverage.

## i18n

One new key `hudChrome.bags.cannotDestroy` (English + the five non-Latin fills per M16, since the value is wordy). The destroy prompt reuses the existing `itemUi.bags.destroy*` keys.

## Follow-up (noted, not in this PR)

Right-click is desktop-only, matching the existing desktop-only secondary bag affordances (vendor Ctrl/Meta-click, shift split-stack sell). A touch long-press to reach the same prompt is a small companion addition worth a follow-up for full mobile parity.

## Scope

- `src/ui/bags_view.ts`, `src/ui/bags_window.ts`, `src/ui/i18n.catalog/hud_chrome.ts` (+ 5 overlay fills + regenerated tables), `tests/bags_view.test.ts`.


FernandoX7 (commented):
Nice, tight change. The whole path already existed (`sim.discardItem`, the `discard` command, `IWorld.discardItem`); this just surfaces it, reusing the quest-item destroy prompt and dispatch verbatim, so there is almost no new surface. The decision is a pure `bagDestroyAction` core with decisive tests across every transactional mode plus the `noDiscard` case, the vendor Ctrl/Meta sell path is untouched, and server authority is unaffected since the server re-runs `sim.discardItem` and enforces `noDiscard` regardless of the client (`src/sim/items.ts:51`). Merges clean against `release/v0.23.0`.

One nit:

- nit: the affordance is invisible from the tooltip. `src/ui/bags_view.ts:133` gives a regular item its primary-click hint (`clickEquip` / `clickConsume` / `clickUse`) with no sub-line saying right-click destroys, while quest items advertise `clickDestroy` for their left-click destroy. Since #1501 is "players had no way to remove junk," a destroy affordance nobody can find only half solves it. Consider a tooltip sub-line mirroring the `clickDestroy` hint (outside the transactional modes).

On the mobile follow-up you flagged: agree, right-click-only matches the existing desktop-only secondary bag affordances (vendor Ctrl/Meta sell, shift split-stack sell), so it is fine to land now and add a long-press companion later.

daxdax89:
Good point, a destroy affordance nobody can find only half solves it. Added a `Right-click to destroy` tooltip sub-line, gated on `bagDestroyAction === 'discard'` so it shows only for a genuinely destroyable item (never on protected or transactional items), right where the action works. New `hudChrome.bags.rightClickDestroy` key + the five non-Latin M16 fills.

FernandoX7 (approved):

PR #1570, opened 2026-07-07
feat(mob): fluid hit-and-run pursuit combat for every melee mob

Rubsey:
## Summary

Two stacked changes that together give every hostile melee mob the fluid, Nythraxis-style hit-and-run movement (per the mob-combat-profiles-v022 and mob-pursuit-combat-v023 campaign plans):

**Commit 1 (refactor, behavior-preserving):** generalizes the combat-profile path into the one mob chase/attack path. New module `src/sim/mob/combat_profile.ts` behind the SimContext seam; boss attack mechanics moved verbatim to `runMobAttackMechanics` in locomotion; four dead SimContext callbacks removed. Proven byte-identical by the untouched parity goldens at that commit.

**Commits 2-4 (feat, deliberate behavior change):** flips the behavior on for all melee mobs.
- Every melee mob swings from effective reach while continuing to close to `desiredRange` (80% of reach), tracks a strafing or backpedaling target smoothly, and swings the moment it enters reach. Measured against the old code at an attainable 7 yd/s orbit: a hit every full 2.0s swing cadence, glued at desired range, vs a hit every ~3.5s hovering at the reach boundary before.
- The leash check is modeless: a mob dragged while in melee still evades past the leash.
- Ranged `petSpell` casters keep the classic caster loop verbatim (close to spell range, stand, cast).
- Boss attack mechanics (aoePulse, stomp, bigCast, stoneskin, terrify) stay melee-gated: they run on any engaged tick that ends in melee contact.
- Nythraxis boss and add profiles are untouched (`nythraxis_full_pull` golden byte-identical); the stationary walk-past reach fix is preserved (`movingRangeBonus` semantics unchanged).
- Parity goldens re-minted deliberately in their own commit: exactly 6 combat scenarios move (mob_swing_affixes, hunter_pet, warlock_pet, pet_ai, drowned_litany, c4a_casting_lifecycle), all explainable as mobs closing while swinging; the other 42 are byte-identical.
- Deliberate trade, recorded in the plan: a target sustaining speed above the mob's own (12.8 yd/s orbit, beyond player run speed) can circle-kite hit-free after opening contact; the mob stays locked on and never resets. The re-pinned guard in tests/fixes.test.ts pins both sides of this.

## Test plan

- [x] New behavior pins: tests/mob_pursuit_all.test.ts (hit-and-run, keep-closing in attack state, leash-from-attack, caster loop) written RED-first, now green
- [x] Full mob suite (69 files), world_boss, threat, nythraxis_raid, sim_context, entity_roster, walk-past reach guards: green
- [x] tests/parity green against the re-minted goldens; re-mint diff reviewed scenario by scenario
- [x] tests/fixes.test.ts moving-target guard re-pinned at attainable speed (stronger assertions) plus the super-speed companion pin
- [x] npm run gate: all 9 steps green (full suite, malware scan, i18n, tsc, all builds)
- [ ] Playtest feel: zone-1 camps, casters, Thunzharr, Nythraxis (local stack)


FernandoX7 (commented):
Careful, well-documented move-not-rewrite. I traced the two combat arms statement-for-statement against the old code and the extraction holds up: the default chase and attack paths keep their branch order, early exits, and swing/move positions, the boss mechanics moved into `runMobAttackMechanics` verbatim (comment-only trims), and the pursuit path reproduces the old `updateProfiledMobCombat`. The parity goldens back this: `tests/parity` plus the mob suites are green at head (192 passed), and `tests/architecture.test.ts` confirms the new module stays sim-pure.

I independently confirmed both behavior notes are genuinely inert on current content:

- The only pursuit-profile mobs are the two Nythraxis templates (`combatProfileForMob` in `mob_combat.ts`), and neither declares `aoeSlow`/`battleYells`. Both fields live only on `thunzharr_waking_peak` (`content/zone3.ts:867,900`), which is a default profile that already ran the pulses. So `pulseAntiKiteSnare`/`pulseLoudYell` now firing on the pursuit path draws no rng and changes nothing.
- `chaseSpeedMult` is 1 for every default-path profile, so `updateDefaultProfileChase`'s new `profile.chaseSpeedMult` factor is a no-op.

The four dead SimContext callbacks are fully gone with no lingering references anywhere in the tree.

One nit:

- nit: `src/sim/sim.ts:4137` (`mobEffectiveMeleeRange`) and `src/sim/sim.ts:4141` (`tryMobMeleeSwingInRange`) are now orphaned private delegates. On base they were live via `.bind(sim)` into `ctx`; this PR removed those bindings and the internal callers, and the updated `mob_combat.test.ts` calls the module functions directly, so nothing reaches these two anymore (private, so tests can't either). Biome flags both as new `noUnusedPrivateClassMembers` warnings (non-blocking, so CI stays green). Same dead-scaffolding rule you applied on the SimContext side; worth deleting these two as well. `mobCombatProfile` stays, it is still used by `mobMeleeRange`.

Your tracked follow-ups look right to defer: the runner is exercised through the parity goldens and `world_boss.test.ts` (both green), and the stale `updateMob` "460-line dispatcher" header comment is cosmetic.

PR #1571, opened 2026-07-07
fix(render): flatten the corpse pick capsule so it stops phantom clicks

### LINKED ISSUE #1486: Bug: NPC hitboxes remain in upright 'standing' state after death, causing phantom collisions in perspective view
### Before you start

### What happened?

When using a near-eye/perspective camera view (zoomed in) rather than the default top-down perspective, clicking on the ground behind or "above" a downed NPC corpse still triggers interactions with the invisible upright hitbox of that NPC. The visual representation of the NPC correctly transitions to a flat/lying position on the ground, but their physical collision volume remains vertical and standing, creating "phantom" hitboxes that block clicks.

Technical Notes (for developers)

Based on an investigation of src/sim/combat/damage.ts, the handleDeath function handles the transition to the dead state by setting e.dead = true and updating e.aiState = 'dead', but there is no evidence of any logic that modifies the entity's collision volume, radius, or bounding box during this phase. The discrepancy is most visible in perspective views because the verticality of a standing cylinder/AABB becomes spatially distinct from the flat ground plane, whereas in a pure top-down view, the two shapes overlap almost identically on the X,ZX,Z axes.

### Steps to reproduce

1. In-game, zoom the camera in closely towards the character's eye level (near-eye perspective).
2. Locate an NPC (e.g., a Bandit) and defeat them so they enter the "dead/downed" state.
3. Observe that the NPC's visual sprite is now lying flat on the ground.
4. Attempt to click on the terrain directly behind or "above" the corpse (in the direction of the camera's tilt).
5. Notice that the click is intercepted by the invisible upright hitbox of the corpse, preventing you from interacting with the ground or objects behind it.

When an NPC dies and their body lies flat on the ground, their collision bounds (hitbox) should ideally update to reflect this new physical state (e.g., shrinking in height or transitioning to a flatter AABB/disc). This would ensure that the physical interaction zone matches the visual representation, allowing players to click "through" the area where the upright hitbox used to be.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on Linux Mint

Observe that the camera is zoomed in. I drew the position of my cursor into the image.

Image

daxdax89:
## Problem

In a near-eye/perspective camera, clicking the ground behind or "above" a downed NPC still triggered interactions with the corpse. The model correctly lies flat on death, but its invisible pick volume stayed a full-height standing cylinder, so a ray anywhere in that upright column selected the corpse. Most visible zoomed in, where the standing column is spatially distinct from the flat body.

Closes #1486.

## Layer

This is a render/interaction-layer issue, not sim. The sim has no per-entity movement collision (`src/sim/colliders.ts` is static-world only), so `handleDeath` has no collision volume to change and is correct as-is. The renderer's pick geometry is what stays upright. No determinism/parity surface is touched.

## Fix

Two pick paths select an entity; both treated a corpse as full standing height:

1. **Direct raycast proxy** (`clickProxy` in `visual.ts`): collapse its vertical extent to a low, ground-hugging profile on the death edge (`enterDeath`) and restore it on `revive`, via a new pure `pickProxyHeight()` in the three-free `characters/anim_state.ts` core. Set on the edge (not per-frame) and covers the created-already-dead snapshot path.
2. **Sloppy-pick assist** (`renderer.pick()`): the 26px screen-space fallback built a body-midpoint-to-overhead column for lootable corpses. For a dead entity it now hugs the ground and drops the overhead column, so a near-eye click above/behind the flat body no longer snaps to it.

The ground-level footprint is unchanged in both paths, so a lootable corpse stays clickable to loot; only the phantom upright column is removed.

## Test

`tests/locomotion.test.ts`: `pickProxyHeight` returns full standing height while alive, collapses well below it when dead, never exceeds standing height (wide/short creature), and scales the flat profile with body radius so large corpses stay clickable. (The `renderer.pick()` candidate build is WebGL-bound and follows the same dead-height rule; not unit-tested, consistent with the renderer's other pick code.)

## Scope

- `src/render/characters/anim_state.ts` (pure `pickProxyHeight`), `src/render/characters/visual.ts` (consume on death/revive), `src/render/renderer.ts` (sloppy-pick corpse column), `tests/locomotion.test.ts`.

### DIFF
--- a/src/render/characters/anim_state.ts
+++ b/src/render/characters/anim_state.ts
@@ -52,3 +52,14 @@ export function locomotionTimeScale(
 function clamp(v: number, lo: number, hi: number): number {
   return Math.min(hi, Math.max(lo, v));
 }
+
+/** The vertical extent (scale.y) for an entity's click/pick proxy. The proxy is a
+ *  unit cylinder scaled to (radius*2, standHeight, radius*2) and rooted at the feet.
+ *  A living entity uses its full standing height; a dead (lying) one collapses to a
+ *  low, ground-hugging profile (roughly its own body width tall) so a near-eye click
+ *  behind or above the flat corpse no longer intersects an invisible upright column
+ *  (issue 1486), while the ground-level footprint stays clickable for looting. */
+export function pickProxyHeight(standHeight: number, radius: number, dead: boolean): number {
+  if (!dead) return standHeight;
+  return Math.min(standHeight, radius * 2);
+}
--- a/src/render/characters/visual.ts
+++ b/src/render/characters/visual.ts
@@ -11,6 +11,7 @@ import {
   type BaseState,
   desiredBaseState,
   locomotionTimeScale,
+  pickProxyHeight,
 } from './anim_state';
 import {
   applyMaterials,
@@ -69,6 +70,9 @@ export class CharacterVisual {
   readonly height: number;
   /** invisible capsule for picking (userData.entityId set by the renderer) */
   readonly clickProxy: THREE.Mesh;
+  /** click-capsule radius (measured body extent); the pick proxy's standing scale.y
+   *  is `height`, collapsed to a flat profile while dead (see enterDeath/revive). */
+  private readonly clickRadius: number;
 
   private def: VisualDef;
   private key: string;
@@ -181,6 +185,7 @@ export class CharacterVisual {
     // capsule from measured body extents — long/wide creatures (wolves,
     // dragons) were nearly unclickable with a height-derived sliver
     const r = prep.clickRadius;
+    this.clickRadius = r;
     this.clickProxy = new THREE.Mesh(clickGeo(), clickMat());
     this.clickProxy.scale.set(r * 2, this.height, r * 2);
     this.clickProxy.visible = false;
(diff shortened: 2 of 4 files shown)


FernandoX7 (commented):
Clean, well-scoped fix and the diagnosis is right. I confirmed `src/sim/colliders.ts` is static-world only, so `handleDeath` genuinely has no per-entity volume to change and the sim is correct as-is; this belongs in the renderer.

What I verified:
- The pick proxy is a feet-rooted unit cylinder (`clickGeo` translates +0.5 then scales by height), so collapsing `scale.y` keeps the footprint at ground level and only sheds the upright column. Looting stays intact because the radius footprint is unchanged.
- `pickProxyHeight` lives in the three-free `anim_state` core and is set on the death/revive edge, riding the same `wasDead` edge the death clip already latches, so the created-already-dead snapshot path is covered with no per-frame cost. `scale.y` is written in only three places (constructor, `enterDeath`, `revive`), and `height`/`clickRadius` are readonly, so nothing fights the collapse and `revive` restores it exactly.
- The sloppy-pick degenerate segment (`top == mid` for a corpse) is safe: `distanceToSegmentPx` guards `lenSq > 0` and falls back to point distance.
- Tests pin the real behavior with literals (full height alive, `radius*2` collapse, clamp for wide/short, scale for large) and the run is green.

One nit: the two paths flatten by different amounts. The direct proxy uses `min(standHeight, radius*2)` while the sloppy assist drops to a fixed `0.15 * height` point, so the "matching the flattened pick proxy" comment in `renderer.ts:5127` is approximate rather than exact. Functionally fine, the sloppy path only needs a ground-level anchor inside its 26px radius; just noting the wording implies a parity the numbers do not have.

Nothing blocking.

daxdax89:
Fair, the wording implied a parity that isn't there. Clarified the comment: the sloppy-pick path only needs a ground-level anchor inside its 26px assist radius, so its drop is approximate and not a height-match of the proxy's `min(standHeight, radius*2)`.

FernandoX7 (approved):

PR #1572, opened 2026-07-07
feat(render): swap wolves and wolf forms onto custom baked models

Donny-Deals:
## Summary

Swaps the pack wolf, Old Greyjaw, and the druid Wolf Form / shaman Shadewolf off the old Quaternius `wolf.glb` and onto custom Tripo wolf meshes auto-rigged onto the `Dog_Animation` quadruped skeleton, and fixes the auto-rigged skin weights that deformed in extreme poses.

## What changed

- **Pack wolf and Old Greyjaw** use the baked models (`wolf_basic.glb`, and a larger `greyjaw.glb` at height 2.2). The pack wolf keeps its light entity tint so it still doubles as the beast-family fallback and each beast keeps its own colour.
- **Druid Wolf Form and shaman Shadewolf** (`form_cat`, which the shaman renders with the ghost material on top) now use `wolf_basic.glb` via a new shared `WOLF_BAKED` clip map that adds the baked rig's `Sit` and `Fall` clips plus a paddling swim base, so the player forms sit, jump, and swim correctly instead of falling back to the steep procedural prone.
- **Skinning fixes** for the auto-rigged mesh, each verified with a per-clip stretch/tear probe:
  - leg-chain containment (stops the haunch ballooning in Fall/Death),
  - a distal radial cap (stops paw fur drooping to the ground),
  - a crotch-web demote-to-thigh plus a region-restricted gradient smooth (fixes the back-left leg tearing on jump, the last visible defect).
- **Hunter pet wolf** needs no change: a tamed beast is a mob whose family routes through `FAMILY_KEYS.beast` to `mob_wolf`, so it inherits the new model automatically. Confirmed in-game by taming a Forest Wolf.

In-game names are unchanged (Old Greyjaw, Forest Wolf); this is a pure model/rig swap.

## Testing

- All 14 clips on both wolves render clean across a full contact-sheet sweep, and an adversarial per-clip stretch probe finds no remaining tear (the flagged frames are intended mane/tail/death-flop deformation, all below the tear threshold).
- Verified in-game from the gameplay camera: druid Wolf Form idle/walk/run/jump/sit/swim/attack, shaman Shadewolf, the world pack wolves and Old Greyjaw, and a tamed hunter pet wolf.
- `visual_manifest` and `architecture` guard tests pass; added a `visual_manifest` test that machine-checks every wolf clip name resolves in its GLB.
- Regenerated the media manifest, wiki content, and guide stills.


FernandoX7 (commented):
Clean model/rig swap, and the verification behind it holds up. Things I checked and confirmed at the head ref:

- The shared `form_cat` is wired correctly. `renderer.ts` derives `cat = ghostWolf || hasCatForm` and creates the visual with the `form_cat` key (renderer.ts:4041, 4238), so the shaman Shadewolf really does render this same visual with the ghost material on top. The comment matches the code.
- Both baked GLBs contain every clip the manifest references. I extracted `wolf_basic.glb` and `greyjaw.glb` and listed their animations: both carry Idle, Walk, Gallop, Attack, Death, Idle_HitReact_Left/Right (the `animal()` core) plus Sit, Fall, Walk (WOLF_BAKED's sitIdle / jump / swim). Nothing dangles.
- The new `visual_manifest` test is decisive: it reads the real GLB and asserts every `expectedClipNames` entry (sitIdle/swim/jump included) resolves, so a future clip-rename drift fails loudly. Good addition.
- The scale note checks out: `old_greyjaw` template scale is 1.25 (zone1.ts:140), so the 2.2 model height lands at roughly 2.75 in-world.
- The deleted still (`mob_wolf__566061.webp`) is not referenced anywhere else, and the guide content/models regen is consistent with the swap.

One optional note:

- nit: `models/creatures/wolf.glb` looks orphaned from gameplay now. The only remaining reference is the generated editor asset catalog; no `VISUAL` points at it. Fine to keep since the editor still lists it, but worth a deliberate call on whether to drop it.

Nothing blocking. Render-only and gameplay-neutral (model height is visual scale, not the sim hitbox), no player-string surface (in-game names unchanged), CI green, clean merge.

FernandoX7 (approved):

FernandoX7:
Pushed fb6a3e506 to drop the orphaned `wolf.glb` per the nit. With every VISUAL now on `wolf_basic.glb` / `greyjaw.glb`, the old Quaternius wolf had no gameplay reference left, so removing it is the cleaner call.

- Removed `public/models/creatures/wolf.glb` and regenerated the media manifest (only that one line drops, no other hashes move).
- Regenerated the editor asset catalog: the dead `creatures/wolf` palette entry is gone, and this branch's two new baked wolves (`greyjaw`, `wolf_basic`) are now cataloged. I kept the delta scoped to the wolves; three creatures the catalog was already stale about (from earlier merges) are left alone.
- Left the build-spec recipe row in `scripts/assets/specs/characters.json` untouched. It is a manual-only asset build (never run by `npm run build` or CI), and that file carries pre-existing formatting debt that the changed-file biome gate would force a full ~1160-line reformat on. Happy to strip that row in a separate reformat pass if you'd rather have it gone too.

Guards green: `visual_manifest`, `render_asset_preload`, `editor_placements_alignment`, `architecture`, `guide`, `tsc`, and `biome ci --changed`.

PR #1573, opened 2026-07-07
fix(sim): give the fishing-dock plank deck collision

### LINKED ISSUE #1500: Bug: Pier at Deepfen shallows has no collision
### Before you start

### What happened?

I went to Deepfen shallows and saw a pier. I wanted to step on it, but there is no collision. My character passes trough without stopping.

### Steps to reproduce

1. Go to Deepfen shallows 
2. find the pier with the barrel/crates on top
3. Try to collide with the pier
4. Observe the character passes trough

Collision with the pier so that I may stand on top of it since all the other structures have collision (except the gravestones)

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on linux mint

Image

daxdax89:
## Problem

The pier at Deepfen shallows (with barrels/crates on top) had no collision: the character walks straight through it, while every other structure blocks. Expected: the pier is solid so the player can stand on it.

Closes #1500.

## Root cause

The dock is `src/sim/content` data rendered by `src/render/props.ts` (three `dockPlatform` deck sections stepping toward the water, plus a stone hut and barrels/crates). But `src/sim/colliders.ts` derived only the **hut** OBB from `PROPS.docks` in its docks loop; the plank deck and its props got no collider, so the deck was walk-through. The same asset is the zone-1 dock too.

## Fix

Emit one deck OBB per dock in the docks loop, in the dock's local frame, spanning the plank run (footprint tracks the rendered deck sections at local z -1.05/-3.18/-5.31 and the on-deck props at x -0.6..1.45). The deck is now solid like every other structure. Both docks share the geometry, so the single derived collider fixes both. The collision model is pure XZ push-out (there is no walkable-top surface anywhere in the engine), so "stand on it" resolves, as for every other structure, to a solid footprint you no longer clip through.

## Determinism / parity

Static geometry only: no rng draw, no tick-phase change. Colliders are rebuilt per (content, seed) and cached, so determinism and the three-host parity are unaffected. The renderer already samples the same `PROPS.docks`, so no `IWorld` change is needed; this closes the render/collision drift on the deck.

## Test

`tests/dock_collision.test.ts`: the deck mid-run point is now blocked at both docks (was walk-through), and open water a couple yards past the far deck edge stays walkable (the footprint is bounded, not an over-water lip).

## Note

The exact footprint (`DECK_HALF_WIDTH`/`DECK_HALF_DEPTH`, the local center) is derived from the rendered plank + prop extents; a quick in-game walk onto the Deepfen dock to confirm the collider matches the visible planks (no phantom lip over the water, no gap) is worth a look before merge. Happy to nudge the numbers with a screenshot if useful.

## Scope

- `src/sim/colliders.ts` (deck OBB in the docks loop + named footprint constants), `tests/dock_collision.test.ts`. Sim-side data; no render/UI/server change.

### DIFF
--- a/src/sim/colliders.ts
+++ b/src/sim/colliders.ts
@@ -76,6 +76,15 @@ function rotY(lx: number, lz: number, rot: number): { x: number; z: number } {
   return { x: lx * c + lz * s, z: -lx * s + lz * c };
 }
 
+// Fishing-dock plank-deck collider footprint, in the dock's local frame (the
+// deck runs from the shore toward the water along local -z; see props.ts). The
+// deck sections sit at local z -1.05/-3.18/-5.31 and the on-deck props span x
+// -0.6..1.45, so one OBB centered mid-run covers the walkable planks + props.
+const DECK_LOCAL_X = 0.1;
+const DECK_LOCAL_Z = -3.2;
+const DECK_HALF_WIDTH = 1.4;
+const DECK_HALF_DEPTH = 3.3;
+
 // ---------------------------------------------------------------------------
 // Collider sets
 // ---------------------------------------------------------------------------
@@ -128,7 +137,7 @@ function staticWorldColliders(seed: number): Collider[] {
     out.push({ type: 'circle', x, z, r: 5, cameraTopY: topY(seed, x, z, 5.2), camGhost: true });
   }
 
-  // dock huts
+  // dock huts + the plank deck they sit on
   for (const d of PROPS.docks) {
     const hut = rotY(d.hutLocal.x, d.hutLocal.z, d.rot);
     const x = d.x + hut.x,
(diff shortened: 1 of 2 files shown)


zfreiberg:
Tested this locally: checked out the branch, ran the server with dev mode, walked around the Deepfen Shallows dock.

    The deck itself is solid now — no more walking through the planks.
    Confirmed not being able to jump up onto the dock is expected (matches the "pure XZ push-out, no walkable-top surface" note in the PR description), not a regression.
    One gap: the rowboat next to the dock is still walk-through-able. Its local position (x=2.4, z=-5.0 in props.ts) falls outside the new deck collider's footprint (x -0.6..1.45 per the PR description), so it never got covered by the fix.

Not sure if the boat is considered part of the original request or if it should be added as a new bug. If we are happy with no boat fix then I approve this change (if I could, I don't know how to actually approve this)

Screenshots attached.
image


image

daxdax89:
Thanks for the careful test and the screenshots, that's exactly the confirmation this needed. Glad the deck reads solid, and you're right that not being able to jump onto the dock is intended (pure XZ push-out, no walkable top surface).

Good catch on the rowboat. It's a separate render prop drawn beside the deck's far end (`src/render/props.ts`, dock-local `2.4, -5.0`) that, like the pier, never had a collider, so it's the same class of bug on a different object. Since this PR is scoped to the pier itself (issue #1500), I'll keep it as-is so it can land on your review, and open a small follow-up PR to give the moored rowboat its own collider. I'll link it here.

Appreciate the review.

daxdax89:
Opened the follow-up for the rowboat: #1752. It adds a rotation-safe circle collider for the moored boat at the world position `props.ts` draws it, with a test pinning the hull blocks at both docks and the footprint stays bounded. Kept it separate so this PR stays scoped to the pier deck. Thanks again for catching it, @zfreiberg.

Rubsey:
Thanks @daxdax89, the root-cause analysis here is exactly right: colliders.ts only ever derived the hut OBB from PROPS.docks, so the deck had no collision at either dock.

Closing as superseded, though: a fix for #1500 landed on release/v0.26.0 via the v0.25.0 issue batch (PR #1839, commit a36a78c47) and took a different approach. Instead of a solid deck blocker, it adds a shared src/sim/dock_layout.ts layout module and exposes the plank surface as raised walkable ground through groundHeight in world.ts (only the hut blocks), so players can actually stand on and cross the pier as the issue expected, in all three hosts. It also ships its own tests/dock_collision.test.ts pinning the walkable surface, which is why this branch now conflicts on the base.

Merging the XZ push-out OBB on top of that would wall off a pier players can now walk onto, so this one has to stand down. Appreciate the careful write-up (the footprint derivation and the bounded-water test were both solid), sorry the batch fix beat it to the branch.

PR #1575, opened 2026-07-07
feat(dungeons): heroic difficulty for the four five-player dungeons, with Heroic Mark drops

Rubsey:
## Summary

Adds a heroic difficulty mode to the four five-player dungeons (hollow_crypt, sunken_bastion, drowned_temple, gravewyrm_sanctum), selected by the party leader before entry and fixed for the life of a claimed instance, plus a Heroic Mark participation reward on heroic final-boss kills. Implements the H-series planning briefs (heroic only: no mythic+, no forged drops, no affixes, no badge economy).

Three commits:

1. **feat(dungeons): add heroic difficulty**: `DungeonDifficulty` ('normal' | 'heroic') with the `isDungeonDifficulty` guard; a declarative tuning table (`src/sim/content/dungeon_difficulty.ts`) and a pure transform module (`src/sim/instances/difficulty.ts`); instance slots carry a difficulty, claimed at `claimInstance` and reset by `freeInstance`; leader-only selection via `/dungeon|/dungeons|/instances normal|heroic`; the `set_dungeon_difficulty` command and always-sent `ddiff` self field mirror the selection into `ClientWorld`; `IWorldDungeons` gains `dungeonDifficulty()`/`setDungeonDifficulty()` implemented in both worlds. RNG draw order is preserved (the per-spawn level roll still consumes one draw from the original template before the heroic override), so parity goldens are byte-stable.

2. **feat(dungeons): heroic marks, fire-time mechanic scaling, difficulty lifecycle fixes**:
   - **Heroic Marks**: the final boss of a heroic instance drops one personal `heroic_mark` per eligible participant (the same downed-members-included snapshot as XP and loot rights), as `personalFor` corpse slots so each player can loot exactly one. New SimContext callback `awardHeroicMarks` owned by `instances/dungeons.ts`, called from the death hub after `rollLoot`; draws no rng. Item + catalog entry + five non-Latin locale fills + procedural icon. No vendor or lockout in this pass (per the badge-policy brief).
   - **Mechanic scaling**: boss mechanic numbers (aoePulse, bigCast, stomp, mendAlly, wardAllies, stoneskin) are read from the base MOBS table at fire time, so heroic spawns carry per-entity `mechanicDamageMult`/`mechanicHealMult` applied after each rng draw (draw order unchanged).
   - **Authority fix**: while grouped, party state is the only difficulty authority; a member's stale personal stamp can no longer claim heroic under an unset party, and only the setter's own preference is stamped.
   - **Lifecycle fix**: entry rejoins any live claim for (dungeon, partyKey) regardless of the current selection, so a mid-run flip or a ghost corpse-run can never strand a player in a fresh parallel instance; the selection applies to the next fresh claim after the empty-instance reset.
   - Dead code removed; `/dungeon` readout now reports the current selection.

3. **test(dungeons): close reviewer gaps** on the mechanic-mult fire sites and setter-only stamping.

Rebased onto release/v0.23.0: generated i18n artifacts regenerated (never hand-merged), IWorld parity pins re-derived (181 -> 183 members), command-schema pins re-derived (120 sends / 129 dispatch).

## Test plan

- [x] `npm run gate` green on the rebased branch (all 9 steps: i18n gen + freshness, malware scan, changed-files biome, full Vitest suite with bounded workers, tsc, all five builds)
- [x] `tests/parity` green with the draw-order digest unchanged (no golden churn)
- [x] New coverage: direct unit tests for the difficulty module (exact transform pins, tuning contract), multiplier pins independent of the level-20 bump, party-create inheritance, stale-stamp bypass regression, heroic boss adds, fire-site scaling at aoePulse/stomp/stoneskin, marks (party/solo/normal/trash negatives), server-side dispatch validation incl. junk values, all three slash aliases, wire round-trip
- [x] Reviewed by fresh architecture-reviewer (determinism/seam) and test-coverage-auditor passes: zero blocking findings
- [ ] Manual smoke on a dev realm: leader flips difficulty, party runs heroic Hollow Crypt, everyone loots one mark
- [ ] Balance pass on the heroic tuning multipliers (1.03 to 1.15 are placeholders; only hollow_crypt's are literal-pinned by tests)


ryan-foo:
Verified the summary's load-bearing claims against the code and they hold: the per-spawn
level roll still draws from the base template before the heroic override, every mechanic
fire site multiplies after its rng draw, the new SimContext callbacks and IWorld members
carry re-derived pins in the same change, and the dispatch arm is guarded with junk-value
tests. The transform module + tuning table split is exactly the house pattern, and the
rejoin-any-live-claim lifecycle is a genuinely good call the planning docs should adopt.

Findings, none blocking:

- should-fix (follow-up, not this PR): Heroic Marks have no daily gate or lockout, and an
  empty instance frees after 300s, so marks are infinitely farmable. Inert while
  heroic_mark has no sink (src/sim/content/items.ts, sellValue 0), and the brief descopes
  the vendor deliberately, but the gate must exist before any spend sink ships. Please
  file a tracking issue so it cannot slip. (-> Will implement a vendor + jewelry on this release)
- nit: src/net/online.ts dungeonDifficulty() has a dead ?? 'normal' on a field already
  initialized to 'normal'.
- nit: ClientWorld.setDungeonDifficulty stamps the local mirror before the server
  validates leadership, so a non-leader briefly reads heroic locally until the next
  snapshot's always-sent ddiff corrects it. Fine as is, noting the window.

EnriqueGF:
@Rubsey could you merge the new changes?

PR #1520, opened 2026-07-06
feat(desktop): offer the Linux (AppImage) download on the site

EnriqueGF:
## Why
The Electron desktop build already produces Linux artifacts (electron-builder `build.linux`: AppImage + deb, x64 + arm64) and the release runbook already uploads them, but the website download view only linked the macOS dmg (plus a "Windows build pending" note). This adds the missing Linux button so Linux players, including Bazzite and Steam Deck, can grab the app.

The button points at the x86_64 **AppImage**, not the deb: the AppImage runs on immutable Fedora atomic desktops (Bazzite, Steam Deck) with no system install (chmod +x and launch), which the deb cannot do there.

## What changed
- New `src/game/desktop_download.ts` (keeps `main.ts` a firewall): pure `detectDesktopPlatform()` + `desktopDownloadUrl()` with the version and artifact filenames in one place, plus a thin `initDesktopDownload()` that syncs the button hrefs to the version constant, highlights the visitor's own OS (floating it first), and reveals the AppImage note for Linux. The static hrefs in `index.html` remain the no-JS fallback.
- `index.html`: second download button (`data-platform="linux"`) and a Linux-only hint paragraph.
- i18n: new English `download.linuxCta` / `download.linuxHint` keys in the catalog (generated locales English-filled by the build).
- `src/styles/shell.css`: a subtle ring on the detected-OS button and the hint style.
- `docs/desktop-release.md`: publish step notes the page now offers macOS + Linux and to keep `DESKTOP_VERSION` in lockstep with the uploaded build.

## Verification
- `tsc --noEmit` clean.
- vitest green: `tests/desktop_download.test.ts` (platform detection incl. Fedora atomic UAs, Android-is-not-desktop, iOS fallback; URL building) and `tests/desktop_download_dom.test.ts` (jsdom: href sync, OS highlight and reorder, hint reveal). Also ran `architecture` and the `localization_fixes` i18n S3 guard.
- biome ci clean on the changed files (no errors, no format diffs); pre-push QA floor green.

## Follow-up (infra, not code)
The button goes live once the matching Linux AppImage is uploaded to `updates.worldofclaudecraft.com/desktop/` alongside the dmg. That upload is already part of the desktop release runbook (`docs/desktop-release.md`); this PR is the website side only.

For a more native Bazzite install path later (software-store, sandbox, auto-update), a Flatpak on Flathub would be the next step; out of scope here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1579, opened 2026-07-07
ci(desktop): publish Linux launcher artifacts from CI and re-enable the download

Rubsey:
## Summary

Gets the Linux Electron build (added in #1520, held back in #1527 because nothing uploaded the artifacts) into the release pipeline and served.

- New workflow `.github/workflows/desktop-linux-publish.yml`: on a release tag `v<version>` (or manual dispatch), builds the website-channel Linux artifacts (AppImage + deb, x64 + arm64), generates `SHA256SUMS`, and uploads everything plus both per-arch electron-updater feed files to the R2 bucket behind `updates.worldofclaudecraft.com/desktop/`.
- Trigger contract is merge to main PLUS tag: the workflow refuses a tag whose commit is not on `main`, and refuses any mismatch between the tag, `package.json` `version`, and `DESKTOP_VERSION`.
- Upload order is artifacts first, `latest-linux.yml` / `latest-linux-arm64.yml` last, so an installed app is never offered an update whose file is not yet downloadable. Filenames are pinned explicitly (x86_64 AppImage / amd64 deb arch tokens), so a silent electron-builder rename fails the job instead of stranding the feed.
- Re-enables the Linux download button (reverts the #1527 `ARTIFACT` holdback and restores the static no-JS href in `index.html`). The hardened unavailable-button handling from #1527 stays, now exercised through the unpublished Windows platform in the DOM test fixture.
- `docs/desktop-release.md`: new "Linux publish from CI" section with the full provisioning checklist and verification commands.
- Only Linux is automated (it needs no signing); macOS and Windows stay on the manual runbook until their signing secrets are provisioned in CI.

## What you need to do to get this live

1. **Cloudflare R2 (one time):** create a bucket (e.g. `woc-desktop-updates`) and connect the custom domain `updates.worldofclaudecraft.com` to it (R2 bucket settings, Custom Domains; the zone must be on the same Cloudflare account). The workflow uploads under the `desktop/` prefix, matching the path the feed URL and download page already use. Do not enable the `r2.dev` public URL; the custom domain is the only public surface needed.
2. **R2 API token (one time):** Cloudflare dashboard, R2, Manage API Tokens: create an "Object Read and Write" token scoped to that bucket. Note the Access Key ID, Secret Access Key, and the Cloudflare account id.
3. **GitHub secrets (one time):** repo Settings, Secrets and variables, Actions: add `R2_ACCOUNT_ID`, `R2_BUCKET`, `R2_ACCESS_KEY_ID`, `R2_SECRET_ACCESS_KEY`.
4. **If the mac dmg currently lives somewhere else:** copy the existing 0.22.0 mac artifacts (`world-of-claudecraft-0.22.0-mac-universal.dmg`, `.zip`, `.zip.blockmap`, `latest-mac.yml`) into the same bucket under `desktop/` so the one hostname serves both platforms.
5. **Backfill 0.22.0 Linux BEFORE the next prod deploy of main:** after merging, run the workflow manually (Actions, "Desktop Linux publish", Run workflow on `main`). It builds and publishes the 0.22.0 Linux artifacts, making the re-enabled button's target real. The button only reaches players at the next site deploy, so run the backfill first.
6. **From v0.23.0 on:** merge the release to main, tag the merge commit `v0.23.0`, push the tag. Linux publishes itself; mac stays manual per the runbook (bump `DESKTOP_VERSION` in the release as before, the workflow enforces the lockstep).

Linux users validate downloads with `sha256sum -c SHA256SUMS --ignore-missing` against `https://updates.worldofclaudecraft.com/desktop/SHA256SUMS` (regenerated per release, Fernando's checksum ask). Per Fernando: no Flatpak/Snap; note it is the AppImage that self-updates via electron-updater, deb installs update manually.

## Test plan

- [x] `npx vitest run tests/desktop_download.test.ts tests/desktop_download_dom.test.ts` (12 passed)
- [x] `npx tsc --noEmit`
- [x] Biome clean on changed TS files; workflow YAML lints
- [ ] After R2 setup: manual `workflow_dispatch` on main publishes 0.22.0 and the two curl verification commands in the runbook succeed
- [ ] Fresh AppImage install self-updates when a higher version is published


Rubsey:
Extended to macOS in 48c9e9fdb. The workflow is now `.github/workflows/desktop-publish.yml` ("Desktop publish"): a shared verify job (main-ancestry + version lockstep), then independent `linux` and `mac` publish jobs, so a mac signing failure never blocks the Linux publish and vice versa.

The mac job builds the signed + notarized universal dmg + zip on `macos-latest`, verifies with `codesign --verify --deep --strict` and `spctl -a -t exec` before uploading, and refuses to run at all without the Apple secrets (an ad-hoc build can never publish). Uploads mirror the existing host conventions: content types, immutable cache headers on versioned artifacts, near-uncached feeds, feed file last. Checksums are now per-platform (`SHA256SUMS-linux` / `SHA256SUMS-mac`) since the two parallel jobs would otherwise clobber one shared key.

Five additional secrets needed for the mac job (values from the same credentials the manual mac build used, see `inventory/desktop_release.env` in the deploy repo):

- `CSC_LINK`: the Developer ID `.p12` as base64 (`base64 -i woc-developer-id.p12`)
- `CSC_KEY_PASSWORD`: its password
- `APPLE_API_KEY_P8`: the raw text content of the AuthKey `.p8` file (CI writes it to disk; the manual flow passes a path instead)
- `APPLE_API_KEY_ID`, `APPLE_API_ISSUER`: same values as the manual flow

Without them, only the mac job fails (with a message naming the missing secrets); Linux publishes regardless. The R2 secrets from the PR description are shared by both jobs. Runbook section renamed to "Publishing from CI (Linux + macOS)" with the full provisioning list.

Rubsey:
Retargeted to release/v0.23.0 per discussion: the Linux button re-enable now ships in the same release as the first CI-published artifacts, so the 0.22.0 backfill step is no longer needed.

Two follow-up commits:

- 01c2fd8ff: manual workflow_dispatch is now a dry run by default. It builds, signs, verifies, and checksums exactly like a release, then attaches the artifacts to the workflow run (7-day retention) instead of uploading; tick "publish" to really upload. Tag pushes always publish.
- 7ef32eccd: release_version.mjs (release:check / prepare) now carries the Linux AppImage href and DESKTOP_VERSION through the version bump and fails the check when either is stale. Without this, the v0.23.0 bump would have left both at 0.22.0 and the publish workflow's lockstep guard would have refused the tag. Test-first, including the play.html-has-no-Linux-link negative case.

Updated release-day sequence:

1. Merge this PR into release/v0.23.0.
2. Release prep as usual; npm run release:check now also verifies the Linux href and DESKTOP_VERSION.
3. Merge release/v0.23.0 into main. Nothing publishes yet; the workflow becomes dispatchable at this point.
4. Optional rehearsal: Actions, "Desktop publish", run on main with publish unchecked; inspect the attached artifacts (notarized dmg, AppImages).
5. Tag the merge commit v0.23.0 and push the tag. Linux and mac build and publish to the update host automatically; feed files upload last.
6. Deploy the site. The download page's Linux button goes live pointing at the 0.23.0 AppImage.

All nine repo secrets (four R2, five Apple) are already configured.

PR #1582, opened 2026-07-07
fix: make bags open on first toggle

### LINKED ISSUE #1538: Bags window takes two clicks or two presses of B to open after loading in

loopermallee:
## Summary
- Fixes the first bags toggle after fresh load by using the existing computed visibility helper instead of inline style.display checks.
- Keeps all bags show sites on display = flex so the existing flex-column layout is preserved.
- Updates the client shell regression guard so future #bags open/closed reads use computed visibility.

Closes #1538

## Test Plan
- `npx vitest run tests/client_shell.test.ts --testNamePattern="reconciles every #bags display show-site"`
- `npx tsc --noEmit`
- `npm run security:gate`
- `npm run build`
- Browser verification: fresh shared toggle opens #bags on first activation, second toggle closes, minimap first click opens, and B alternates close/open once the game canvas is focused.
- `npm test` note: full local suite reported six 5s timeout failures unrelated to this UI change; rerunning those six files with `--testTimeout=20000` passed (238 passed).


loopermallee:
Thanks, happy to adjust if anything else comes up.

loopermallee:
Credit and thanks to @FernandoX7 for raising #1538.

loopermallee:
Rebase/retarget update:

- Rebased `fix/bags-first-toggle` onto `upstream/release/v0.23.0` and force-with-lease pushed `c33d92a54`.
- i18n artifacts confirmed unchanged for this non-i18n PR. No generated i18n/hash/status commit was needed.
- Focused validation passed:
  - `npx vitest run tests/client_shell.test.ts --testNamePattern="reconciles every #bags display show-site"`
  - `npx tsc --noEmit`
  - `npm run security:gate`
  - `npm run build`

jgyy:
Closing as already fixed: toggleBags() opens bags correctly on first toggle on current main. Thanks for the contribution.

PR #1586, opened 2026-07-07
feat(valecup): the Vale Cup, boarball at the Sowfield (1v1 to 5v5)

maxpolaczuk:
Adds **boarball** (football) at the **Sowfield**, a walled pitch just south of Eastbrook: a full pick-up-and-play minigame with matchmaking, private practice, and a stadium.

### Gameplay and matchmaking
- Brackets **1v1 through 5v5** with charged **Shoot**, **Pass**, and a body-checking **Slide**, all gated to seated players in a live match's play phase.
- **FIFO matchmaking** that fills empty slots with bots on timeout and preempts idle bot exhibitions the moment real players queue; an **autofilled goalie** on every 3v3+ side.
- **Guild-banner entry** with player and guild leaderboards. Decided **rated** matches grant daily-reward points (`recordValeCupResult`); bot-backfilled and practice matches stay unrated.

### Private practice
- Parallel practice on **far-offset instanced pitch copies**, so practice never occupies the physical Sowfield. The real match slot stays **byte-identical at the origin** (no shared-rng draws on the vale-cup tick path, parity goldens unchanged). A live match hard-closes the pitch and ejects stragglers.
- Practice renders the full stadium plus an otherworldly deep-space **skybox** (ESO Milky Way panorama, CC BY 4.0), picked per bout, and plays the boarball match track rather than the battle theme. Another match's overlay never leaks into practice.

### Presentation
- Rocket-League-style cosmetic **light trail** on the ball, thicker for big kicks.
- **Walkable grandstand tiers** (climb the bleachers) plus `/sit` and `/stand` seated poses.
- The pitch stays a clean **mown ground**: wild grass, plants, critters, and pollen motes are excluded from the stadium footprint (shared `isInSowfieldShell`, reused by world decoration).

### Port notes (developed on v0.21, ported onto release/v0.23.0)
- Merged with the crafting/professions and delve daily-reward systems that landed after the branch point.
- Rebound the Vale Cup window default key from `KeyT` (now crafting) to **`KeyY`**.
- Refreshed the W0b command-count, delta-key, and W0c IWorld parity pins; regenerated i18n, wiki content, and the media manifest.

### Follow-up fixes (review and playtest feedback)
- **Football HUD confined to the Sowfield.** The live-score indicator and the kickoff / goal / save / countdown / golden / end banners are now gated on `isAtSowfield` (the same predicate that arms the stadium music), so there is zero trace of the running match anywhere off the football ground, including Eastbrook next door. A private practice instance still shows its own theatre only to someone on that far, isolated pitch. (The queue-position chip stays realm-wide so you can roam while you wait.)
- **You must possess the ball to play it.** Kick / Pass / Shoot previously gated their contact check on the ability's aim range (18 to 42 yd), so a player could launch the ball at goal from tens of yards away. Striking now requires possession: within `VC_POSSESSION_RADIUS` (4 yd) of the loose ball, or holding it as keeper. Four yards covers the dribble/trap contact plus a dribbled ball riding a stride ahead and stays at or above the bots' swing reach, while a ball across the pitch never responds. Regression tests pin both directions (off the ball does nothing; at your feet launches).

### Verification
- `npx tsc --noEmit` clean.
- Full suite green: **10554 passing**, 14 skipped, 0 failing.
- `npm run ci:changed` (biome) clean: 0 errors; pre-push QA floor green.
- Rebased cleanly onto the latest `release/v0.23.0`; merges without conflict.

New player-facing content feeds `/wiki` and all sim strings are registered via `sim_i18n`. English-only per the contributor i18n tier (maintainer fills locales at release), with the M16 non-Latin fills provided for the wordy new hud_chrome keys.


FernandoX7 (commented):
Really strong PR for a change this size. The seam work is exactly right: `IWorldValeCup` is added and pinned in both worlds, match and queue state ride a delta-omitted `s.vcup` self field with no new `Entity` wire fields (so `blankEntity` stays untouched), the sport kit swaps through the wireRev-gated `s.sport` flag with the restore's explicit null handled on both hosts, and the whole feature draws zero shared rng so the parity goldens are unchanged. Server authority and the betting and daily-reward economy are watertight, and the pins are real (the `world_api_parity` and `command_schema` counts are re-derived from source, not just bumped). A few things below, none blocking.

**should-fix**

- `src/render/vale_cup_practice_sky.ts:77`: the panorama loads via a raw `new THREE.TextureLoader().load('/env/space_galaxy.jpg')`, bypassing `loadTexture` and the media manifest. The build already emits a hashed copy (`manifest.generated.ts` maps `env/space_galaxy.jpg` to `/media/env/space_galaxy.dd78358b9e62.jpg`), so this ships the 691 KB jpg twice and skips content-hash cache-busting plus the shared load queue. Route it through `loadTexture('/env/space_galaxy.jpg', { srgb: true })`, the way `sky.ts:189` loads the biome backdrops.

- `src/ui/vale_cup_betting.ts:139` plus `src/styles/components.css:4373`: the losing or closed side's stake buttons are locked only by a `.locked` class (`pointer-events: none`), and `wire()` always binds `onBet`. That stops the mouse but not the keyboard, so a Tab then Enter still calls `world.vcupBet` on a closed window or the side you have already backed against. `vcupPlaceBet` re-validates server-side so it is not exploitable, but it is inconsistent with the window itself, which correctly uses the `disabled` attribute (`vale_cup_window.ts:299,311,321,357`). Set `disabled`/`aria-disabled` on the locked stakes and early-return in the handler.

**nits**

- Mobile touch targets: the stake buttons (`components.css:4387`, `padding: 4px 2px; font-size: 11px`) and the compact role/nation/bracket buttons sit under the 40x40 target-size floor on coarse pointers; the global rule only bumps input font, not button size.

- `src/ui/hud.ts:7937`: the new `if (ev.pid !== undefined && ev.pid !== sim.playerId) continue;` drops every pid-tagged event for a non-local pid offline, not just the vcup ones. I checked the one pre-existing pid-tagged event that reaches here (`companionBark`) and it is correctly scoped by this (a delve companion line is personal), so it reads as intentional, but the gate is broader than the feature. Worth a confirming glance that no other pid-tagged event was meant to render for offline bystanders.

- `src/ui/hud.ts` `ensureShootChargeEl`/`updateShootCharge`: the shoot power meter is the one vcup surface built imperatively inside `hud.ts` rather than as a thin `*_painter` like the other five. Justifiable since it is coupled to the private input-charge state, just the odd one out relative to the pattern the rest of the PR follows.

- Small render micro-allocations, all cosmetic and bounded (not the steady-state hot path): `vale_cup_team_ring.ts:249` calls `nationColors()` twice per frame despite the file's own "allocates nothing" note; `vale_cup_practice_sky.ts` `moodFor()` returns a fresh object per frame the caller discards, and `PracticeSkyMood.hemiSky`/`hemiGround` are dead (only `fog` is read at `renderer.ts:3826`); `vfx.ts:587` maps a `THREE.Color[]` per goal firework.

- `index.html`/`play.html`: the `#mm-valecup` keycap placeholder still reads `t`, but the default bind is now `KeyY` (`KeyT` is crafting). `hud.ts:5558` rewrites it to the live bind every frame so it is never actually visible, just stale in source.

Determinism, IWorld parity, server authority, persistence back-compat, and English-only i18n all check out: I confirmed the sim modules draw no rng, `blankEntity` is untouched, the parimutuel pool conserves copper, `recordValeCupResult` dedupes per account and match and only credits rated decided matches, and old character rows load with the new fields defaulted.

FernandoX7 (approved):

PR #1587, opened 2026-07-07
fix(ui): apply UI scale after slider release

### LINKED ISSUE #1558: Bug: UI changes while sliding the UI scale slider in options menu

loopermallee:
Closes #1558

## Summary
- Added a `commitOnChange` slider option and enabled it only for UI Scale.
- Kept slider readout/fill previewing during drag.
- Deferred UI Scale application until `change` or `blur`.
- Preserved live-apply behavior for other sliders such as HUD opacity.

## Testing
Passed:
- `npx vitest run tests/options_view.test.ts tests/options_window.test.ts`
- `npx vitest run --config vitest.browser.config.ts tests/browser/a11y.browser.test.ts -t "defers only UI Scale"`
- `npx tsc --noEmit`
- `npm run security:gate`
- `npm run build`

Manual verification:
- `--ui-scale` stayed `1` during slider `input`.
- UI Scale readout and fill updated while dragging.
- `change` applied `1.25`.
- `blur` applied `1.3`.
- HUD opacity still live-applied to `0.75`.

Known unrelated failure:
- `npm test` failed with 33 unrelated failures, mostly timeout-heavy sim/parity/security tests plus i18n failures around an unrelated `lockActionBars` change.

### DIFF
--- a/src/ui/options_view.ts
+++ b/src/ui/options_view.ts
@@ -40,6 +40,8 @@ export interface SliderControl {
   /** Current value at build time; the painter re-reads the live value on input. */
   value: number;
   fmt: SliderFmt;
+  /** When true, preview input updates locally and the setting writes on change/blur. */
+  commitOnChange?: boolean;
 }
 
 export interface ToggleControl {
@@ -144,9 +146,20 @@ const slider = (
   labelKey: TranslationKey,
   fmt: SliderFmt = 'percent',
   step = 0.05,
+  commitOnChange?: boolean,
 ): SliderControl => {
   const r = s.range(key);
-  return { control: 'slider', key, labelKey, min: r.min, max: r.max, step, value: s.num(key), fmt };
+  return {
+    control: 'slider',
+    key,
+    labelKey,
+    min: r.min,
+    max: r.max,
+    step,
+    value: s.num(key),
+    fmt,
+    commitOnChange,
+  };
 };
 
 const toggle = (
@@ -351,7 +364,7 @@ export function buildControllerControls(s: OptionsSettingsSource): OptionsContro
 
 export function buildInterfaceControls(s: OptionsSettingsSource): OptionsControl[] {
   return [
-    slider(s, 'uiScale', 'hudChrome.options.uiScale'),
+    slider(s, 'uiScale', 'hudChrome.options.uiScale', 'percent', 0.05, true),
     slider(s, 'playerFrameScale', 'hudChrome.options.playerFrameScale'),
     slider(s, 'targetFrameScale', 'hudChrome.options.targetFrameScale'),
     slider(s, 'hudOpacity', 'hud.options.hudOpacity'),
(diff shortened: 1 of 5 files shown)


loopermallee:
Credit and thanks to @snipercup for raising #1558.

EnriqueGF:
:) Thanks for the fix and for crediting @snipercup on #1558, appreciated.

I have to request changes, and unfortunately the blockers are not about the code quality itself.

**1. Base branch (blocking).** This PR targets `main`. In this repo, PRs must target the latest active release branch, which is currently `release/v0.23.0` (`main` lags until publish). Please reparent and rebase onto `release/v0.23.0`.

**2. This is already fixed on the latest release (blocking / likely a no-op).** Issue #1558 was already resolved on `release/v0.23.0` by commit `bf6dfe056` ("fix(ui): apply the UI scale slider on release, not live while dragging"). That change introduces the same `commitOnChange` slider option on `uiScale` and defers application to release. The existing implementation in `src/ui/options_window.ts` there is a bit tighter than this PR:
- It uses a single shared `commit()` closure plus a `readoutFromSlider()` preview helper, instead of a separate `commitValue` with a `lastCommittedValue` guard.
- It binds only the `change` event (which for a range input already fires on pointer release/touchend and once per keyboard step). This PR additionally binds `blur`, which the `lastCommittedValue` diff guard then has to de-duplicate. That extra path is redundant.

Because the fix already exists there, rebasing this branch onto `release/v0.23.0` would most likely conflict and, once resolved, be an empty or duplicate change.

**Suggested path forward:** please confirm whether the current `release/v0.23.0` behavior already resolves #1558 for you (it should, it is the same approach). If it does, this PR can be closed as already-fixed. If you found a gap the landed fix misses (for example a specific keyboard or touch case), please rebase onto `release/v0.23.0` and scope the PR to just that delta on top of the existing `commitOnChange` code.

Given the above I did not run the full gate, since the change cannot land as-is against this base. Thanks again for the contribution and for the clear repro and testing notes.

loopermallee:
Thanks for the review and the clear explanation.

That makes sense. I missed that `#1558` had already been resolved on `release/v0.23.0` by `bf6dfe056`, and I also targeted `main` by mistake instead of the active release branch.

I’ll treat this PR as a duplicate/already-fixed change and close it rather than rebasing a likely no-op patch.

For future PRs, I’ll make sure to branch from and target `release/v0.23.0` unless the repo guidance changes.

Thanks again for checking this.

PR #475, opened 2026-06-16
docs(woc): Stake to provision a realm — feature stub for discussion

awidearray:
🚧 **Feature stub — opening discussion, not for merge.** This PR adds a stub spec for the **Stake to provision a realm** `$WOC` flywheel mechanic so we can discuss it in its own thread before any code is written. **No implementation is included.**

**Scores (roadmap):** Tier 3 · Ease 2/5 · Flywheel 5 · Sink · Reg risk Med

**What:** Stake ~1% of supply to provision your own realm (an isolated, env-parameterized server process on the shared database).

**Constraints:** cosmetic-only / no pay-to-win · non-custodial · `src/sim/` stays pure.

**To discuss:**
- Stake size / lockup terms?
- Provisioning limits / who covers infra cost?
- What happens to a realm if the owner unstakes?

Part of the proposed `$WOC` GameFi roadmap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1591, opened 2026-07-07
feat(realm): consume economy service for realm-stake verification, remove migrated program (#475)

awidearray:
Splits the realm-stake-escrow feature (#475: stake $WOC to provision a realm) on the GAME side, moving the `realm_stake_escrow` on-chain verification into the economy service and consuming it via a thin proxy. Companion to the service PR (repo woc-daily-rewards-service, branch `feat/woc-realm-escrow-service`), which is devnet-proven.

## RETARGET NOTE (read first)

- This fork branch is cut from the fork base `feature/woc-realm-stake-escrow` (fork `awidearray/world-of-claudecraft`), which does NOT exist on `levy-street`. The PR therefore targets `main` so the diff is reviewable, but `main` is far ahead of the fork base, so the raw diff is large.
- To land cleanly, the maintainer should either push the fork base `feature/woc-realm-stake-escrow` to `levy-street` for a scoped per-feature PR and retarget this PR onto it, OR cherry-pick the single realm-stake-split commit onto the current `main` tip.
- Land order: the SERVICE side (`feat/woc-realm-escrow-service`, stacked on the escrow-handle tip) must land first or in lockstep, since the game's stake proxy depends on the service's `/v1/economy/realm-stake/*` routes and the additive optional `escrow` handle on the stake/quote request.

## What lands (the one realm-stake-split commit)

- **`server/realm_stake_proxy.ts`** (new): a thin, secret-gated, graceful client to the service's `/v1/economy/realm-stake/*` routes. `x-woc-economy-secret`, fetch-only (no new dependency), bounded timeout. Returns null when the service is not configured (`WOC_ECONOMY_URL` / `WOC_ECONOMY_SECRET` unset), so the game boots and runs service-off.
- **`server/realm_provision.ts`**: `verifyStakeLock` now routes through the service when configured. The game quotes the stake with the escrow handle it derives (the realm stake PDA); the service independently re-derives the SAME PDA and rejects a mismatch, then confirms the finalized deposit against exactly that destination. No client money math on this path. When the service is off it falls back to the local on-chain verify (graceful degradation). Service reasons map back onto the existing wire codes so the client renders unchanged; adds one retryable code `stake_service_unavailable`.
- **Removed the migrated Rust program** `solana/programs/realm_stake_escrow/` and its now-dead Anchor workspace (`solana/`): the program is owned by the service repo. The client keeps its PDA derivation + lock tx builder (`server/realm_escrow.ts`) because the wallet still signs the deposit, and that stake PDA is exactly the handle the service verifies (LARP-guard grep-clean: the only remaining on-chain reads in `realm_provision.ts` are the intentional service-off fallback).
- i18n: English-only `realm.err.stake_service_unavailable` key + `ERR_KEYS` mapping + regenerated resolved tables.

## Verification
- `tsc --noEmit` green (0 errors).
- Server bundles (`scripts/build_server.mjs`).
- `tests/realm_operator.test.ts` (ERR_KEYS coverage incl. the new code) and `tests/realm_escrow_client.test.ts` (client PDA + lock-builder parity) pass.
- Proxy verified graceful service-off (quote/confirm return null, `verifyStakeLock` falls back).

## Devnet (service-side, proven)
The service PR proved the handshake on devnet against the deployed program `6Dg8...`: real `lock` deposit sig `5WvvAwqrtXNBHygzWgVUeen1dZg3HQBnXxk614LREBPa4eA7bZi51zccpa1xaBk3GWev8Z81fCW12UcuDuWsZSQN`, destination-equality on stake PDA `9UgnXsQ8txnrLscvJtpHSp7n25DkKHGMgRwjKSYXj875` (game-derived == service-resolved == on-chain vault owner == service-stored).

## Notes
- No `declare_id` changed (the program id lives in the service repo now; the game's PDA derivation uses the same `6Dg8...`).
- No other feature touched.


EnriqueGF:
## Review: changes requested

I reviewed the actual delta (the single realm-stake-split commit `ffb404a`), not the ~17k-line raw diff. The split is clean and the design (secret-gated, fetch-only, bounded-timeout proxy with graceful service-off fallback) is sound. A few things should be addressed before this is production-ready.

### Blocking (mergeability / process)
- As targeted, this is not mergeable. None of the realm-stake files (`server/realm_provision.ts`, `server/realm_escrow.ts`, `src/ui/realm_operator.ts`, ...) exist on `main`; the branch is 32 commits ahead of and 2192 behind `main`, so merging pulls in the entire unmerged GameFi/realm stack (~29 prior commits), not the one intended commit. You already flag this in the RETARGET NOTE. Please retarget onto the feature/release base that carries the realm stack, or cherry-pick `ffb404a` onto the current tip, so CI runs the scoped diff. Also confirm the land order with the service PR, since the game path hard-depends on the `/v1/economy/realm-stake/*` routes.

### Should fix (code)
1. Missing tests for the new server logic. `server/realm_stake_proxy.ts`, `verifyStakeLockViaService`, and `mapServiceReason` have no unit coverage; the only test change is a one-line ERR_KEYS entry in `tests/realm_operator.test.ts`. Repo convention is that server/ changes always get a test. `mapServiceReason` is a pure, high-value function whose mapping can regress silently (e.g. `memo_mismatch -> wrong_vault_amount`, the `rail_disabled|missing_handle|bad_handle|handle_mismatch -> stake_service_unavailable` bucket, and the default). Add a table-driven test for it plus a test that `realmStakeServiceConfigured()` / `config()` gate correctly on env, and that `verifyStakeLock` dispatches to the service vs local path.

2. `mint` is no longer verified game-side on the service path. In `verifyStakeLockViaService` (server/realm_provision.ts) `args.mint` is destructured but never used: the quote sends `amountBase` but no mint, so the game fully delegates mint enforcement to the service. The local path still checks token-2022 and the exact mint on both the credit and the spend. The proxy comment claims the service verifies "exact mint + amount", but the game supplies no mint, so this is now an implicit cross-repo contract. Please either pass the mint in the quote so the service can pin it, or make it explicit (comment + service-side test) that the service independently enforces the correct mint; otherwise a staker could satisfy `amountBase` with a different token on that destination.

### Minor (defense in depth)
3. `verifyStakeLockViaService` trusts the service to reject a handle mismatch via `reason` but never asserts `quote.escrow === escrow.handle` before calling confirm. A cheap belt-and-suspenders equality check on the returned escrow would harden the "one destination, no divergence" guarantee against a buggy/compromised service response.

### Verified good
- Replay guard intact: `recordProvisionedStake` still relies on the DB UNIQUE(lock_tx_sig), so both verify paths are replay-safe.
- Secret handling: `x-woc-economy-secret` is server-only, never exposed to the client; graceful null-on-unconfigured is correct.
- Client rendering unchanged: unknown/unmapped codes fall back to `realmOp.err.generic`, so the `mapServiceReason` default is safe on the client.
- i18n: English-only key added to the `en` catalog with resolved/pending regenerated; no locale overlay hand-edited.
- No dangling references to the removed Rust program; PDA derivation + lock-tx builder correctly retained. No em/en dashes or emojis.

Happy to re-review once retargeted with the proxy/verify tests and the mint contract clarified.

EnriqueGF:
:) Thanks for splitting this out, the direction (delegating on-chain verification to the economy service behind a secret-gated, graceful proxy) is clean and the fallback design is sound.

I re-reviewed against the current tip and confirmed EnriqueGF's earlier points are still open, plus one new item. Verdict: changes requested.

### Blocking: base branch (process gate)
This PR targets `main`. Per the repo rule (CLAUDE.md "Default task workflow": base off the latest release branch, not `main`), it must target the latest release branch, which is currently `release/v0.23.0`. This is blocking on its own regardless of code quality. Beyond the rule, targeting `main` makes the diff ~366 files / 21k lines because none of the GameFi/realm stack has landed yet, so CI never runs the scoped delta and the merge would pull in the entire unmerged stack. Please retarget onto `release/v0.23.0` (rebasing the realm stack onto it), and confirm the land order with the economy-service PR that owns `/v1/economy/realm-stake/*`.

### Should fix (still open from the prior review)
1. No test coverage for the new server logic. `server/realm_stake_proxy.ts`, `verifyStakeLockViaService`, and `mapServiceReason` have zero unit tests (the only test delta is one `ERR_KEYS` line in `tests/realm_operator.test.ts`; there is no `tests/realm_stake_proxy*` file). `server/` changes always get a test. `mapServiceReason` is a pure function whose mapping can regress silently (e.g. `memo_mismatch -> wrong_vault_amount`, the `rail_disabled|missing_handle|bad_handle|handle_mismatch -> stake_service_unavailable` bucket, the default passthrough). Add a table-driven test for it, plus tests that `realmStakeServiceConfigured()` gates on env and that `verifyStakeLock` dispatches service vs local.

2. `mint` is no longer enforced game-side on the service path. In `verifyStakeLockViaService` (server/realm_provision.ts) `args.mint` is present in the args type but never read; the quote sends `amountBase` with no mint. The local path still pins the exact mint and rejects Token-2022. So the game now fully delegates mint enforcement to the service as an implicit cross-repo contract. Either pass the mint in the quote so the service can pin it, or make the contract explicit (comment plus a service-side test) that the service independently enforces the correct mint; otherwise a staker could satisfy `amountBase` with a different token at that destination.

### Minor / defense in depth
3. `verifyStakeLockViaService` never asserts `quote.escrow === realmStakePda(realmId).toBase58()` before calling confirm. A cheap equality check on the returned escrow would harden the "one destination, no divergence" guarantee against a buggy or compromised service response, rather than trusting the service to have re-derived the same PDA.

4. (New) `ownerAccountId: 0` is hardcoded for every realm quote, with a comment that the service id is unused. If the service ever keys idempotency, rate-limiting, or attribution on `ownerAccountId`, sending `0` for all realms would collide across owners. Please either pass the real staker account id or confirm (and document) that the service ignores this field.

### Verified good
- Replay guard intact: `recordProvisionedStake` still relies on the DB `UNIQUE(lock_tx_sig)`, so both verify paths stay replay-safe.
- Secret handling: `x-woc-economy-secret` is server-only, never sent to the client; null-on-unconfigured graceful fallback is correct, and the sequential quote/confirm calls are bounded (8s each) so a hung service cannot wedge the handler.
- i18n: English-only `stake_service_unavailable` key added to the `en` realm catalog with resolved/pending regenerated; no locale overlay hand-edited. Unknown/unmapped service reasons fall back to `realmOp.err.generic` on the client, so the `mapServiceReason` default passthrough is safe on the render side.
- Removed Rust program cleanly; PDA derivation and lock-tx builder retained. No em/en dashes or emojis.

Happy to re-review once this is retargeted onto `release/v0.23.0` with the proxy/`mapServiceReason` tests and the mint contract clarified.

EnriqueGF:
:) Thanks again for keeping this split moving, the proxy design (secret-gated, fetch-only, bounded timeout, graceful service-off fallback) still reads well.

Re-reviewed at the current tip. No new commits have landed since my last pass (HEAD is still `ffb404a`, 2026-07-07), so the four earlier items remain open verbatim; I will not restate them in full. One new finding below, which I rate higher than anything raised so far. Verdict: changes requested.

### Blocking: base branch (process gate)
Unchanged and blocking on its own: this PR targets `main`. Per CLAUDE.md ("Default task workflow": base off the latest release branch, not `main`) it must target `release/v0.23.0`. GitHub also reports the PR as CONFLICTING, so it cannot merge as targeted. Please retarget onto `release/v0.23.0` with the realm stack rebased on it, so CI runs the scoped delta rather than the ~17k-line raw diff, and confirm the land order with the economy-service PR that owns `/v1/economy/realm-stake/*`.

### New, high: the service `confirm` call binds nothing the local path binds
`realmStakeConfirm` sends only `{ realmId, signature }` (server/realm_stake_proxy.ts:99). Compare the two verify paths in `server/realm_provision.ts`:

- Local (`verifyStakeLockLocal`, lines 162-185) enforces four independent guards: finalized, not Token-2022 (`token_2022`), the credited amount equals `expectedAmount` exactly (`wrong_vault_amount`), and the linked staker wallet actually funded it (`ownerSpentBase(...) < expectedAmount` -> `wrong_payer`). It then returns `amountBase: credited`, the value observed on chain.
- Service (`verifyStakeLockViaService`, lines 128-156) sends the mint, the amount and the payer to `quote` only, sends none of them to `confirm`, and on success returns `amountBase: args.expectedAmount`, the value the game asserted.

Two consequences. First, mint, payer binding, and the Token-2022 rejection all become an implicit cross-repo contract on `confirm`, and `mapServiceReason` (lines 76-97) has no case that can even produce `wrong_payer` or `token_2022`, so the service currently has no wire vocabulary to signal them. My earlier point 2 covered the mint; payer binding and Token-2022 are the same class of gap and were not called out. Second, the amount written to `realm_stakes` via `recordProvisionedStake` (line 411, `amountBase: quote.amountBase`) is now an expectation rather than a measurement on this path. If the service ever confirms against a stored quote for that `realmId` whose `amountBase` differs from the one this call quoted, the game records the wrong principal with no way to notice.

Concretely: pass `owner`, `mint` and `amountBase` on the confirm request and have the service echo the verified `amountBase` back, then record the echoed value, not `args.expectedAmount`. That restores measure-do-not-assert and keeps the two paths at guard parity. If the service genuinely resolves all of this from the stored quote, that needs to be stated explicitly in the proxy contract plus covered by a service-side test, because right now the game cannot tell.

This also sharpens my earlier point 4: `ownerAccountId: 0` plus a `confirm` keyed on `realmId` alone means quote identity is carried entirely by `realmId`. The game's own `confirmProvisionQuote` authz (`quote.accountId !== args.accountId`, line 397) is what stands between that and cross-owner confusion, which is a lot to rest on an unspecified cross-repo idempotency semantic.

### Still open from the prior review (no code change since)
1. No test coverage for the new server logic. There is still no `tests/realm_stake_proxy*` file; the only test delta remains one `ERR_KEYS` line in `tests/realm_operator.test.ts`. `server/` changes always get a test. `mapServiceReason` is a pure, table-shaped function and should be table-tested, alongside `realmStakeServiceConfigured()` gating on env and `verifyStakeLock` dispatching service vs local.
2. `mint` unenforced game-side on the service path (`args.mint` is in the args type, never read). Subsumed by the new finding above.
3. `verifyStakeLockViaService` never asserts `quote.escrow === realmStakePda(realmId).toBase58()` before calling confirm.
4. `ownerAccountId: 0` hardcoded for every realm quote.

### Verified this pass
- Replay guard intact: `recordProvisionedStake` still relies on the DB `UNIQUE(lock_tx_sig)`, and `confirmProvisionQuote` consumes the quote on a unique violation, so both verify paths stay replay-safe.
- Secret handling correct: `x-woc-economy-secret` is server-only, never reaches the client; `postJson` re-reads config per call, returns null on non-2xx and on abort, and both calls are bounded at 8s, so a hung service cannot wedge the handler. A null never silently accepts a stake; it surfaces as retryable `stake_service_unavailable`.
- i18n: English-only `realm.err.stake_service_unavailable` added to the `en` realm catalog, resolved/pending regenerated, no locale overlay hand-edited. Unmapped service reasons fall back to `realmOp.err.generic` client-side, so the default passthrough is safe on render.
- `tests/architecture.test.ts` and `tests/realm_operator.test.ts` pass on the branch (18 assertions). `tests/localization_fixes.test.ts` errors only on a missing generated `src/ui/i18n.status.json` in a fresh worktree, which is a local setup artifact, not the PR.
- `tsc --noEmit` could not be fully verified locally: the branch adds `@solana/web3.js` to `package.json`, which is not present in the base tree, so every residual error traces to that one unresolved module. Worth noting that this dependency arrives with the stack, not with this commit, but it does land if the PR merges as targeted.
- Removed Rust program is clean: no dangling references, PDA derivation and lock-tx builder correctly retained. No em dashes, en dashes, or emojis.

Happy to re-review once this is retargeted onto `release/v0.23.0`, the confirm call binds owner/mint/amount (or the contract is made explicit and tested), and the proxy plus `mapServiceReason` have unit tests.

EnriqueGF:
:) Thanks for continuing to push this split forward, the proxy itself (secret-gated, fetch-only, bounded timeout, null-on-unconfigured fallback) is still the part of this change I like most.

Third pass, at HEAD `ffb404a`. No commits have landed since my last review, so items 1-4 and the confirm-binding finding stand verbatim and I will not restate them. Two corrections to my own earlier pass and three new findings below. Verdict: changes requested.

### Blocking: base branch (process gate)

Unchanged and blocking on its own. This PR targets `main`; it must target `release/v0.23.0` (the latest release branch). GitHub still reports the PR as CONFLICTING / DIRTY, so it cannot merge as targeted regardless.

### New, high: the PR lands 32 commits, not "the one realm-stake-split commit"

The body says "What lands (the one realm-stake-split commit)" and "No other feature touched". That is not what merging this would do. Against the merge-base `cb928ca`, `git log cb928ca..ffb404a` is **32 commits**, and these files do not exist on `main` at all:

- `server/realm_provision.ts`, `server/realm_escrow.ts`, `server/realm_stake_proxy.ts`
- `server/affiliate_db.ts`, `server/flow_ledger.ts`, `server/payout_keeper.ts`

So merging as targeted lands the entire realm/GameFi stack (DB-backed realm registry, stake tiers, the Anchor escrow, the operator UI, the affiliate program, unified referral identity, buy-a-realm with SOL/USDC plus the $WOC bond, the flow ledger, the payout keeper, the season panel) in one shot, most of which has never been reviewed on its own. This is the concrete reason the base-branch gate matters here rather than being a formality. Please split per feature against `release/v0.23.0`, or cherry-pick just the split commit as the body proposes.

### New, high: `/api/realms/confirm` is an unrate-limited external-RPC amplifier

`server/main.ts:668` registers `POST /api/realms/confirm` with **no `rateLimited(req)` check**, while every sibling route has one: `/api/realms/tiers` (main.ts:635), `/api/realms/quote` (648, whose comment says it is there "so an authed account cannot use it as an RPC amplifier"), `/api/realms/buy/info` (686), `/api/realms/buy/quote` (696), `/api/realms/buy/confirm` (744).

Each `/confirm` call runs `verifyStakeLock`, which is either `getFinalizedTx(lockSig)` against the Solana RPC (`realm_provision.ts:169`) or a two-call round trip to the economy service (`realm_provision.ts:140-150`). And a failed verdict returns early **without consuming the quote**:

```ts
// server/realm_provision.ts:409
if (!verdict.ok) return fail(400, verdict.reason);
```

`deleteRealmQuote` is only reached on success or on the unique-violation path (424 / 429). So an authenticated account holding one live quote can replay `/confirm` with arbitrary `lockSig` values for the full quote TTL, forcing one outbound RPC or economy-service call per request, unbounded. Please add `rateLimited(req)` to match the sibling routes.

Related, low: `lockSig` is only checked non-empty (`main.ts:673`) before it reaches the RPC body. `server/woc_payment.ts:44` already pins its signature with `BASE58 = /^[1-9A-HJ-NP-Za-km-z]{32,90}$/`; use the same bound here.

### New, medium: the on-chain and DB claims are not verified by CI

I ran this rather than taking it on trust:

```
npx vitest run tests/realm_provision.localnet.test.ts tests/realm_db.integration.test.ts
  Test Files  2 skipped (2)
       Tests  15 skipped (15)
```

Every `*.localnet.test.ts` gates on `REALM_ESCROW_RPC` and every `*.integration.test.ts` on `PG_TEST_URL` (for example `tests/realm_escrow.localnet.test.ts:44`, `tests/realm_db.integration.test.ts:22`), and `grep` over `.github/workflows/` finds no `PG_TEST_URL`, `REALM_ESCROW_RPC`, or `WOC_ECONOMY_*`. So none of these bodies execute in CI. Combined with my still-open item 1 (no unit test anywhere for `realm_stake_proxy.ts`, `verifyStakeLockViaService`, `mapServiceReason`, or `realmStakeServiceConfigured`, confirmed again by grep over all of `tests/`), the settlement path this PR rewires is protected by nothing that actually runs. Reverting any branch of `mapServiceReason` breaks no test. The env gating predates this PR, which is exactly why the new service path needs plain Node unit tests rather than relying on gated integration coverage.

### New, medium: `@solana/web3.js` enters the game client bundle

`package.json` adds `"@solana/web3.js": "^1.98.4"`, and it is imported from three client modules: `src/net/wallet.ts:25`, `src/net/realm_buy.ts:10`, `src/net/realm_escrow.ts:10`. CLAUDE.md is explicit that the dependency set stays tiny and that the one sanctioned exception is Svelte, scoped to the admin bundle. This is a heavy transitive tree landing in the game entry. Please justify the bundle-size cost, or keep the client on the raw-RPC approach `server/solana_tx.ts:4` already documents ("no @solana/web3.js in the server").

### New, low

- `StakeVerdict.amountBase` is dead on both paths: `recordProvisionedStake` writes `quote.amountBase` (`realm_provision.ts:411`), never `verdict.amountBase`. On the local path those are equal by construction, so nothing is wrong today, but it means the measured on-chain value is discarded even where it exists. This sharpens the confirm-binding finding: the fix is to record the value the verifier observed, not the one the quote asserted.
- Module-first: `src/sim/sim.ts` gains a `grantBonus` method (around line 4455) and `src/ui/hud.ts` gains a window plus a `syncWocSeasonButton` cluster. Neither monolith should grow; `grantBonus` is a server-invoked mutation that belongs behind the `SimContext` seam.

### Corrections to my previous review

Both of my earlier "could not verify" notes were my own worktree artifacts, not defects in this PR, and I want them on the record as cleared:

- `tsc --noEmit` is **green, 0 errors**, after `npm ci` with this branch's own `package-lock.json`. My earlier note about unresolved `@solana/web3.js` was a stale `node_modules`, not a real error.
- `tests/localization_fixes.test.ts` **passes** (26 passed, 3 skipped) after `npm run i18n:gen`, and `git status` is clean afterwards, so the committed generated i18n tables are fresh. The i18n side of this change is fully clear.

### Verified good this pass

- `tests/architecture.test.ts`, `tests/realm_operator.test.ts`, `tests/realm_escrow_client.test.ts`: 20 passed. Sim purity holds; no `Math.random` / `Date.now` / `performance.now` introduced under `src/sim/`.
- Server authority is sound: `/api/realms/confirm` takes only `quoteId` + `lockSig`, and every value the verification binds against (`realmId`, `ownerWallet`, `mint`, `amountBase`) is read from the server-persisted quote (`realm_provision.ts:396-407`). Ownership is enforced (`quote.accountId !== args.accountId` gives 403). A client cannot forge a stake.
- Replay guard intact: `UNIQUE(lock_tx_sig)` inside the BEGIN/COMMIT, unique violation rolls back and consumes the quote.
- Secret handling correct: `x-woc-economy-secret` is server-only and never logged; `postJson` returns null on non-2xx and on abort, both calls bounded at 8s, and a null surfaces as retryable `stake_service_unavailable` rather than silently accepting a stake.
- Screenshots are present under `docs/screenshots/`.
- No em dashes, en dashes, or emojis. Removed Rust program is clean, with PDA derivation and the lock-tx builder correctly retained.

Happy to re-review once this is retargeted onto `release/v0.23.0` and scoped down, `/api/realms/confirm` is rate-limited, the confirm call binds owner/mint/amount (or the cross-repo contract is made explicit and tested), and the proxy plus `mapServiceReason` have unit tests that run in CI.

Issue #1531, opened 2026-07-06
Release v0.23.0


Rubsey:
### Target version

0.23.0

### Release branch

`release/v0.23.0`

### Bump the version

- [ ] Ran `npm version <X.Y.Z>` (or `npm version minor`/`patch`/`major`) and pushed the resulting commit and tag.
- [ ] Confirmed all three manifests show the new version and the native build numbers incremented.

### Verify

- [ ] `npm test` passes.
- [ ] `npm run build` succeeds.
- [ ] Ran the release malware-audit gate.

### Ship

- [ ] Web deploy is live (see `DEPLOY.md`).
- [ ] Mac desktop Electron app release artifact is published.
- [ ] iOS archive built via `npm run native:open:ios` and submitted to App Store Connect.
- [ ] Android bundle built via `npm run native:open:android` and submitted to Play Console.
- [ ] Release notes added under `docs/release-notes/`.

### Carryover from v0.22.0

Carried over from the v0.22.0 release issue (#1474): PRs and issues referenced there, or still targeting `release/v0.22.0`, that are open and not yet merged/closed. Open PRs that were still based on `release/v0.22.0` have been retargeted to `release/v0.23.0`.

#### Open PRs now targeting `release/v0.23.0`

- [ ] #1526 feat(content): The Sunken Road, a deep tunnel from Eastbrook Vale to Mirefen Marsh
- [ ] #1525 feat(game): finish mobile action ring HUD wiring
- [ ] #1522 feat(claudium): in-game Claudium UI consuming the economy service SDK
- [ ] #1521 feat(market): add Legendary to World Market rarity filter
- [ ] #1517 feat(editor): standalone map-editor build (3-axis move, terrain detach, click-cycle, Scene Collection, wireframe)
- [ ] #1515 fix: Timed casts resolve against the cast-start target
- [ ] #1514 feat(content): the Mirror World — a level-20 endgame zone (Vale of Glass), forward-ported to v0.22.0
- [ ] #1513 fix(sim): keep every world-boss damager on the loot roster, even after death
- [ ] #1511 [codex] fix homepage server status and total players
- [ ] #1508 feat(x): add in-game X account linking
- [ ] #1507 feat(ui): indicate self-only abilities in spell tooltips
- [ ] #1504 feat(demon-hunter): add playable class
- [ ] #1499 feat(woc): SNS subdomains and tradeable characters, re-cut on the #1496 rename base
[...]

PR #1594, opened 2026-07-07
fix(tests): run npm .bin shims via shell on Windows in the endpoint golden

AccompliceNZ:
## Summary

Makes tests/server/new_endpoint.test.ts pass on Windows checkouts. It spawned
node_modules/.bin/tsc and .bin/vitest directly, but on Windows those are .cmd
shims that spawnSync can only launch through a shell (the same workaround
scripts/gate.mjs already applies to npm/npx), so the spawn returned status
null and the golden failed on every Windows machine. The child vitest include
paths are also forward-slashed: they are globs, and a backslashed Windows
path is consumed as glob escapes and matches no test files.

No production code changes; CI (Linux) is unaffected, this fixes the local
gate for Windows contributors.

## Related issues

Part of #1531 (targets `release/v0.23.0`).

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands: `npx vitest run tests/server/new_endpoint.test.ts` (23/23 green on
  Windows 11, previously 1 failing); full `npm run gate` green on Windows with
  this fix applied; `npx tsc --noEmit` clean.
- Manual steps: none needed; test-infrastructure-only change.

---

### Quality

- [x] **Builds succeed.** Unaffected (test-only change); verified via the
      full gate regardless.

### Cross-platform

- [ ] ~Tested on desktop and mobile.~ N/A, no player-facing surface.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed; no generated files
      touched.

### DIFF
--- a/tests/server/new_endpoint.test.ts
+++ b/tests/server/new_endpoint.test.ts
@@ -57,6 +57,9 @@ const REPO = fileURLToPath(new URL('../../', import.meta.url));
 const SCRIPT = join(REPO, 'scripts', 'new_endpoint.mjs');
 const TSC = join(REPO, 'node_modules', '.bin', 'tsc');
 const VITEST = join(REPO, 'node_modules', '.bin', 'vitest');
+// The .bin shims are .cmd files on Windows, which spawnSync only launches via a
+// shell (the same workaround scripts/gate.mjs uses for npm/npx).
+const SHELL = process.platform === 'win32';
 
 // Every file the generator appends to, seeded as a copy into each temp root.
 const APPEND_TARGETS = [
@@ -105,13 +108,17 @@ function gitPorcelain(): string {
  */
 function runChildVitest(root: string, testPaths: string[]): { status: number; out: string } {
   const config = join(root, 'vitest.golden.config.mjs');
+  // include entries are globs: keep them forward-slashed so a Windows path's
+  // backslashes are not eaten as glob escapes.
+  const include = testPaths.map((p) => p.replaceAll('\\', '/'));
   writeFileSync(
     config,
-    `export default { test: { include: ${JSON.stringify(testPaths)}, exclude: [] } };\n`,
+    `export default { test: { include: ${JSON.stringify(include)}, exclude: [] } };\n`,
   );
   const result = spawnSync(VITEST, ['run', '--config', config], {
     cwd: REPO,
     encoding: 'utf8',
+    shell: SHELL,
     // CI runners color the piped child output (ANSI then sits between "Tests" and the
     // count in the summary line, defeating the plain-text assertions below), so ask the
     // child for no color and strip whatever arrives anyway.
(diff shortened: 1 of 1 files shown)


EnriqueGF:
Thanks for chasing this down, and for the clear reproduction in the description :) Windows contributor ergonomics are worth fixing, and the glob diagnosis is exactly right.

I verified the branch locally rather than trusting CI (the check rollup on this PR is empty, nothing has actually run):

- Base branch is `release/v0.23.0`, the latest release branch. Gate passes. Branch is behind base but merges cleanly.
- `npx tsc --noEmit` clean.
- `npx vitest run tests/server/new_endpoint.test.ts` -> 23/23 green on Linux, so no regression on CI's platform.
- `tests/architecture.test.ts` green. Test-only change, so the sim purity, `IWorld` parity, server authority and i18n invariants are all untouched. No em/en dashes or emojis in the added lines.

The forward-slash fix in `runChildVitest` is correct and load-bearing: `testPaths` are absolute (`join(root, 'tests', 'server', ...)` at the golden call site), so on Windows they really would arrive backslashed into a glob `include` and match nothing.

## Major, blocking: `shell: true` does not survive a path with a space, which is the common Windows case

`tests/server/new_endpoint.test.ts:62,121,481`

Node does not quote when it builds the shell command line: it joins `file` and `args` with plain spaces (and on Windows sets `windowsVerbatimArguments`). So a checkout at `C:\Users\First Last\woc-source` produces `cmd.exe /d /s /c "C:\Users\First Last\woc-source\node_modules\.bin\tsc -p ... --noEmit"` and dies on `C:\Users\First`. The same applies to the `--config` and `-p` arguments, since both are absolute paths under `REPO`. I confirmed the concatenation behavior directly:

```
$ node -e '...spawnSync("/path/sp ace/tool", ["-p","/some/path"], {shell:true})'
shell:true  -> status 127, stderr: /bin/sh: /path/sp: No such file or directory
shell:false -> status 0,   stdout: ARGS=[-p /some/path]
(node:...) [DEP0190] DeprecationWarning: Passing args to a child process with shell
option true can lead to security vulnerabilities, as the arguments are not escaped,
only concatenated.
```

So the fix works on `C:\dev\woc-source` and silently fails, with a confusing error, on the very common `C:\Users\<name with a space>\...`. It also opts this test into a deprecated Node path that already prints a warning on Node 22+.

There is a shell-free fix that is strictly better, needs no platform branch, and is already the pattern this same file uses at line 92 (`spawnSync(process.execPath, [SCRIPT, ...])`): spawn the packages' real JS entry points with `node` instead of the `.bin` shims.

```ts
const TSC = join(REPO, 'node_modules', 'typescript', 'bin', 'tsc');
const VITEST = join(REPO, 'node_modules', 'vitest', 'vitest.mjs');
```

then

```ts
spawnSync(process.execPath, [VITEST, 'run', '--config', config], { cwd: REPO, ... });
spawnSync(process.execPath, [TSC, '-p', tsconfig, '--noEmit'], { cwd: REPO, ... });
```

Both entry points exist in the installed tree (`node_modules/.bin/tsc` is a symlink to `../typescript/bin/tsc`) and I confirmed both run: `node node_modules/typescript/bin/tsc --version` and `node node_modules/vitest/vitest.mjs --version` both exit 0. That removes the `SHELL` constant, the space hazard, and the deprecation, on every platform. Please keep the forward-slash `include` normalization, it is needed either way.

## Minor

The `SHELL` comment cites `scripts/gate.mjs` as precedent. That file passes `shell` to `npm`/`npx`, which genuinely are `.cmd` shims with no importable JS entry, so the workaround is justified there and not here. Worth not propagating the pattern to a case that has a clean alternative. (Not asking you to change `gate.mjs` in this PR.)

Verdict: changes requested. Happy to re-review as soon as the spawns go through `process.execPath`.

PR #1596, opened 2026-07-07
feat(game): /freeze and /unfreeze staff commands

AccompliceNZ:
## Summary

In-game moderation commands that stop a player from moving without
disconnecting them.

- Parser: `freeze` and `unfreeze` join the moderation chat-command family
  with the same quoted-name grammar as /kick; optional trailing text becomes
  the audit reason.
- Service: resolves the named target through the existing protected-target
  guard, writes the audit row first (`freeze`/`unfreeze` added to
  MODERATION_ACTIONS and recordInGameAction), then calls the new
  ModerationHost.setFrozen and notices the actor (Froze X. / Unfroze X.).
  Both commands require moderation.act.
- Enforcement is server-authoritative in game.ts: a session-scoped `frozen`
  flag (dies with the session like spectate; a fast reconnect inside the
  linkdead grace window resumes the same session and stays frozen). Freezing
  zeroes the held move input, and the input handler drops move frames while
  frozen; facing/mouselook still applies so a frozen player can look around.
  The target is told explicitly (You have been frozen by a moderator. /
  You can move again.) so the stop never reads as a client hang.
- i18n: matcher rules for the two placeholdered confirms plus the two exact
  target notices, with values in every locale block of
  server_i18n_moderation.ts (including the new cs_CZ); the admin dashboard
  moderation history gains Frozen / Unfrozen labels.

Scope note: freeze holds movement input. Ability self-displacement (Warrior
Charge, Druid Bear Charge) is deliberately not blocked; happy to extend if a
full positional hold is preferred.

## Related issues

Part of #1531 (targets `release/v0.23.0`).

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands: `npm run gate`; focused runs of tests/moderation_commands.test.ts,
  tests/moderation_service.test.ts, tests/moderation_game.test.ts, and the S3
  guard tests/localization_fixes.test.ts.
- Manual steps: none beyond the suites. The GameServer integration test
  proves the held key is zeroed at freeze time, new move frames are dropped
  while frozen, facing still applies, no socket close, both audit rows land
  with exact args, movement restores on unfreeze, and both target notices are
  delivered.

---

### Quality

- [x] **Tests pass.** `npm test` is green, with new parser, service, and
      GameServer integration coverage for the changed `server/` behavior.
- [x] **Builds succeed.** `npm run build`, `npm run build:server`, and
      `npm run build:env` all pass (via `npm run gate`).

### Cross-platform

- [ ] ~Tested on desktop and mobile.~ N/A, no client UI change (chat commands
      plus existing chat notice channels).
- [ ] ~Accessible.~ N/A, no UI change.

### Localization (i18n)

      locale.** All four strings have values in every locale block of
      server_i18n_moderation.ts.
- [ ] ~Numbers, money, dates through formatters.~ N/A.
- [x] Player-facing text emitted from `server/` is re-localized at the client
      boundary in this same change, and

### Hygiene

- [x] No secrets, credentials, or `.env` committed; ALLOW_DEV_COMMANDS
      untouched. Enforcement is server-side only (nothing client-decided).
- [x] Generated files regenerated via `npm run i18n:gen`, not hand-edited.


jgyy:
hi

PR #1522, opened 2026-07-06
feat(store): add Claudium storefront and Season 1 Armory cosmetics

awidearray:
## Summary

Adds the complete player-facing Claudium and premium cosmetics experience to the game client. The chest launcher now opens the WOC Store, with the premium cosmetics storefront as the default view and the existing Daily Rewards experience retained as a sub-tab.

This PR also adds the Season 1 Armory: 28 account-wide weapon skins across four collections and four rarity tiers, with store browsing, ownership, purchasing, equipping, in-world rendering, rarity VFX, and a full character and weapon inspection experience.

The game remains a presentation and orchestration layer for monetary operations. Claudium balances, pack availability, quotes, prices, purchase validation, and grants are authoritative in the companion economy service. Weapon skins are cosmetic only and do not alter weapon stats, reach, speed, or combat outcomes.

### WOC Store

- Replaces the old Daily Rewards panel shell with the WOC Store opened from the chest icon.
- Makes Store the default tab and keeps Daily Rewards available as a sub-tab.
- Preserves storefront content while background refreshes run, with a small non-blocking loading indicator.
- Shows owned, applied, affordable, and unavailable states without replacing the full panel during refreshes.
- Prompts players to purchase more Claudium when their balance is insufficient.
- Keeps Claudium pack and token-price polling scoped to the Claudium purchase panel.
- Removes the obsolete Daily Rewards Hide Chest control because chest visibility is managed in Settings.
- Leaves category support ready for future outfits and mounts while launching with weapons.

### Claudium purchase experience

- Adds the branded Claudium balance treatment, iconography, pack art, animations, and purchase feedback.
- Redesigns the Claudium panel around six consumer-sized packs:
  - 500 for USD 4.99
  - 1,050 for USD 9.99
  - 2,200 for USD 19.99
  - 4,000 for USD 34.99
  - 6,000 for USD 49.99
  - 13,000 for USD 99.99
- Supports Card, SOL, and WOC purchase rails when enabled by the economy service.
- Supports embedded Stripe Checkout and forwards raw signed Stripe webhooks through the game server.
- Supports native Solana wallet transaction signing for SOL and WOC purchases.
- Displays service-provided SOL and WOC quotes, including the service-defined WOC discount, without calculating monetary values in the client.
- Fails closed and keeps the game playable when the economy service or a purchase rail is unavailable.

### Season 1 Armory

- Adds 28 premium weapon skins across sword, axe, mace, dagger, staff, wand, bow, and crossbow families.
[...]

PR #1597, opened 2026-07-07
feat(ui): player-facing Claudium discount display

awidearray:
Stacks on #1522 (in-game Claudium UI). Adds the player-facing discount display for the Claudium discount system (service: levy-street/woc-daily-rewards-service#15).

## What
On the buy flow, when the service reports a discount, the panel shows:
- **N% off** the effective peg price (N = discountBps/100, the only arithmetic - a pure unit rescale).
- The **base-to-credited** Claudium bonus (e.g. 1000 to 1176, +176).
- For the $WOC rail, an always-on incentive even at the floor: **"Pay with $WOC - always 15%+ off"** (a limited-time-promo note appends when a promo stacks).

## Constraints honored
- The UI **computes no pricing**: it displays exactly what the service sent. The discount rides on the `ClaudiumPurchase` result through a validated proxy passthrough (`coerceDiscount`, drops malformed blocks to null).
- The pure view core stays DOM-free / i18n-free; English-only i18n via `t()` (plus the M16 non-Latin fills for the wordy new keys); no em or en dashes.

## Verify
- `tsc --noEmit`: 0 errors
- `vitest tests/claudium_view.test.ts`: 14/14; `tests/architecture.test.ts`: 24/24
- biome (changed files): clean

## Honest note
`tests/i18n_completeness.test.ts` fails, but that is **pre-existing on this base branch** (the `checkout*`/`open` claudium keys already leak untranslated English before this change). None of the new discount keys appear in the leak; this PR adds no i18n regression.


EnriqueGF:
:) Thanks for the clean, well-documented slice, the pass-through discipline and the defensive coerceDiscount are exactly right, and the pure-core split reads well.

I verified locally on a worktree of this branch:
- tsc --noEmit: 0 errors
- tests/claudium_view.test.ts 14/14, tests/architecture.test.ts 24/24, tests/localization_fixes.test.ts green
- No em/en dashes; no src/sim changes; server change is a validated passthrough with no secrets or dev-command surface; display-only (the sole arithmetic is percent = discountBps / 100)
- M16: the wordy new keys carry their five non-Latin fills, and tests/i18n_completeness.test.ts shows zero leaks for any of the new discount keys

I am requesting changes, one blocking issue plus notes.

BLOCKING - the discount block never renders because nothing populates it. The whole chain from view to HTML is correct, but no data source ever sets `discount`:
- src/main.ts hud.attachClaudium() builds the window snapshot from balance/skus/price/storeItems only (main.ts ~1642-1658); it never sets `discount`, so ClaudiumSnapshot.discount is always undefined.
- The buy flow startClaudiumPurchase result (which is where `discount` actually rides) is consumed only for the `.ok` check (main.ts ~1681); the discount on the result is discarded.
- refreshClaudium() -> claudiumWindow.render() re-reads that same store snapshot, which has no discount.

Net effect: input.discount is always undefined, projectDiscount() always returns null, and discountHtml() always returns the empty string. As shipped, the new view/window/CSS/i18n is dead code and the stated behavior ("on the buy flow the panel shows N% off") does not happen. Please add the wire that carries the purchase result discount into the snapshot the window renders, plus a test that exercises that end-to-end path (the current tests only cover projectDiscount in isolation, which is why the gap is not caught).

Design question tied to the above: the discount rides on a purchase RESULT, but the panel it targets is the pre-purchase buy panel. As written a user could only ever see it after completing a purchase. Please confirm the intended trigger point (a price/quote preview vs post-purchase), since it changes where the wire has to live.

Notes (non-blocking):
- Base branch: this stacks on #1522, which correctly targets release/v0.23.0. That is fine as a stack, but do not merge this before #1522, and once #1522 lands please retarget and rebase this onto release/v0.23.0.
- CI is red on this branch: tests/i18n_completeness.test.ts fails on pre-existing checkout*/open Claudium keys inherited from #1522. Confirmed it is not introduced here (no new discount key leaks), but the stack cannot pass the release-tier gate until #1522 fixes those; worth resolving upstream.
- Copy coupling: wocIncentive hardcodes "15%+", tied to floorBps = 1500. If the service floor ever changes, the copy silently drifts. Consider deriving the number from floorBps or at least noting the coupling next to the key.
- Wording: "{percent}% off the effective peg price" frames a bonus-Claudium mechanic (pay full, receive more) as a price cut. Minor, but worth a second look for accuracy.

Happy to re-review once the discount is actually wired through. Nice groundwork.

jgyy:
Closing — this looks already implemented. `src/ui/claudium_view.ts` / `claudium_window.ts` on `upstream/main` already implement the player-facing discount display. Thanks for the contribution.

PR #1599, opened 2026-07-07
feat(loot): group-visible need/greed votes and full roll reveal at resolution

ryan-foo:
## Summary

Group loot rolls are now fully visible to the party, WotLK-style with the XLoot quality-of-life layer:

- While a need/greed roll is open, the roll frame shows a live vote strip: one chip per candidate that fills in with their Need / Greed / Pass choice the moment they answer (empty chip while undecided). Roll numbers are never exposed while the window is open.
- After you answer, the frame stays up as a watch-only row (no buttons) so you can follow the rest of the votes until the server resolves the roll.
- At resolution the sim broadcasts every roller's d100 to the whole party as loot chat lines ("Need Roll - 87 for [item] by Name", one per non-passer) before the existing winner line, so the outcome is auditable by everyone.

How it lands across the seams:

- **sim**: `lootRollGroupStatus` read over `pendingLootRolls` (choice only, no numbers) in `src/sim/loot/loot_roll.ts`, plus the per-roller reveal lines in `resolveLootRoll`. No new rng draws; parity goldens regenerated (events hash only, state and draw order unchanged).
- **IWorld**: new `lootRollGroupStatus()` on `IWorldLoot`, implemented in both `Sim` and `ClientWorld` via a new `lrollg` delta snapshot key (encode in `server/game.ts`, delta-guarded decode in `src/net/online.ts`, registered in the `ALL_DELTA_KEYS` / `TERSE_TO_IWORLD` contract).
- **ui**: pure view core `src/ui/loot_roll_status_view.ts` (registered in `UI_PURE_CORES`) plus a thin `hud.ts` consumer; re-render is gated on a fingerprint so the per-frame poll does no DOM work while nothing changes. The strip reuses the existing `itemUi.lootRoll.*` keys (no new catalog keys).
- **i18n**: the new sim emits get a `RULES` matcher plus `loot.rollNeed` / `loot.rollGreed` translations in every `sim_i18n` dict, including the 8 new-locale blocks.

## Type of change

## How was this tested?

- Commands:
  - `npm run gate` (i18n gen + freshness, malware scan, biome changed-files, full vitest, tsc, env/server/client builds). The full-suite run hit 5 known-flaky timeout failures (`loot_ffa_sim`, `mail`, 2 parity determinism cases); all pass in isolation via `npx vitest run`.
  - Targeted: `tests/loot_roll.test.ts` (new group-status + reveal-broadcast coverage), `tests/loot_roll_status_view.test.ts` (new), `tests/snapshots.test.ts` (lrollg delta round-trip), `tests/world_api_parity.test.ts` (182/133 pins), `tests/parity/` (goldens regenerated with `UPDATE_PARITY=1`, then green), `tests/localization_fixes.test.ts`, `tests/architecture.test.ts`.
- Manual steps: drove the real HUD headless via the new `scripts/loot_roll_votes_shot.mjs` (offline world + stubbed group status): prompt with live strip, choice chips updating, watch-only row after answering.

## Screenshots / recordings

Before (the prompt as it ships today):

before

After, roll open with live votes (self and Kaelis undecided):

prompt with vote strip

After answering Need, the frame stays up as a watch-only row:

watch row

---

### Quality

- [x] **Tests pass.** `npm test` is green apart from 5 known-flaky timeouts that pass in isolation; new sim/view-core behavior is covered by new tests.

### Cross-platform

- [x] **Tested on desktop and mobile.** The strip is display-only (no new touch targets); it flex-wraps and uses `overflow-wrap: anywhere` inside the existing responsive `#loot-rolls` rail.
- [x] **Accessible.** The strip is plain text inside the existing `aria-live=polite` roll rail, so votes are announced as they land; no new interactive elements.

### Localization (i18n)

- [x] **New player-visible strings are translated into every supported locale.** `loot.rollNeed` / `loot.rollGreed` added to the English base table and every sim_i18n dict (14 main blocks + 8 new-locale blocks); the strip itself reuses existing `itemUi.lootRoll.*` keys.
- [x] Player-facing text emitted from `src/sim/` is re-localized at the client boundary in this same change (matcher RULE in `src/ui/sim_i18n.ts`; S3 guard green).


jgyy (commented):
This is a clean, well-scoped feature. The sim change is a straightforward read over existing state (no new rng draws), the wire addition follows the delta-key contract exactly (`lrollg` registered in `ALL_DELTA_KEYS`/`TERSE_TO_IWORLD`, mirrored in `blankEntity`-equivalent player-scalar init), and the HUD side is a proper pure-core (`loot_roll_status_view.ts`, registered in `UI_PURE_CORES`) plus a thin `hud.ts` consumer gated on a fingerprint. The parity goldens confirm the claim in the description: `events` hash moved, `state` and `rng.digest` did not, so this really is reveal-only and doesn't perturb the sim.

Flow this PR adds:

```mermaid
sequenceDiagram
    participant P1 as Player A (needs)
    participant P2 as Player B (greeds)
    participant P3 as Player C (passes)
    participant Sim as Server Sim (loot_roll.ts)
    participant Party as Whole party (HUD)

    Sim->>Party: lootRoll event (prompt opens)
    Note over Sim,Party: lootRollGroupStatus() streams live,<br/>choice only, no roll numbers

    P1->>Sim: submitLootRoll(need)
    Sim-->>Party: lrollg delta: A = need
    P2->>Sim: submitLootRoll(greed)
    Sim-->>Party: lrollg delta: B = greed
    P3->>Sim: submitLootRoll(pass)
    Sim-->>Party: lrollg delta: C = pass

    Note over Sim: choices.size >= candidates.length -> resolveLootRoll
    Sim->>Party: "Need Roll - 87 for [item] by A"
    Sim->>Party: "Greed Roll - 42 for [item] by B"
    Sim->>Party: "A wins [item] (87)"
    Sim-->>Party: roll removed from lootRollGroupStatus
```

Good things:

- `lootRollGroupStatus` and the resolution-time reveal both read the existing `PendingLootRoll.choices` map; the d100 stored at `submitLootRoll` time is reused for the reveal line, so there's no second `rng.int` call and no draw-order change. Confirmed against the parity goldens (`tests/parity/golden/l1_loot_distribution.json`, `party_loot.json`): only the `events` hash moved, `state`/`rng.digest` didn't.
- Master-loot curate-phase rolls are excluded from `lootRollGroupStatus` the same way they're excluded from `activeLootRolls`, and there's a test for it (`hides a curate-phase master roll from the group status`).
- The vote strip is `aria-hidden` with a rationale in the comment (re-renders per vote, would spam a live region up to 10x); the accessible surface stays the once-announced prompt plus the full per-roller reveal in chat. That's a defensible a11y call, not an omission.
- `itemUi.lootRoll.rolled` count goes through `t()` + `formatNumber`, not string concat.
- IWorld extended first (`src/world_api/loot.ts`), then implemented in both `Sim` and `ClientWorld`, per the seam rule.

Nits:

- `src/ui/hud.ts`: `lootRollVotesHtml` builds `<span>` markup with `esc(entry.name)` for values but leaves the surrounding structure as a plain template literal, same pattern as the rest of the file, so no new risk here, just flagging that the vote strip's `title`/`itemName` attribute values run through `esc()` correctly like the rest of the row.
- `PendingLootRoll.partyMembers` is a creation-time snapshot (`partyMembersForRoll`); `lootRollGroupStatus` filters on it too, which is consistent with how the broadcast targets it, but worth double-checking in play-testing that a player who leaves the party mid-roll still correctly stops seeing the strip once the roll resolves (looks correct from the code, just noting the interaction since it's not directly covered by a test).

The PR's own merge-tree against `release/v0.23.0` shows the only true conflicts are the generated i18n trio (`pending.ts`, `i18n.status.summary.json`) plus the parity goldens and `world_api_parity.test.ts`/`snapshots.test.ts`, because the base has moved on since the branch point (e.g. the bank system and the `IWorld` member count changed 202 -> 197 on the base side). That's ordinary rebase churn, not a design problem with this PR: rebase onto current `release/v0.23.0`, regenerate i18n (`i18n:build`) and the parity goldens (`UPDATE_PARITY=1`), and it should be clean.

Nice feature, matches the WotLK/XLoot behavior described, and the invariants (determinism, IWorld seam, i18n scope) all check out from reading the code directly.

PR #1243, opened 2026-07-01
fix(ui): heal tooltips scale by Spell Power; fix Lightning Shield {school} placeholder

ryan-foo:
Two small ability-tooltip accuracy fixes.

## 1. Heal tooltips show their Spell Power contribution `(+N)`
Direct heals and HoTs scale off Spell Power in combat, but their tooltips never showed the `(+N)` contribution that damage abilities do. `abilityDamageBonus` had no `heal`/`hot` case (returned `0`), and `abilityEffectText`'s `hot` branch skipped the suffix entirely (unlike the parallel `dot` branch).

This is the heal half of #1039 (talent tooltip accuracy), which merged with damage scaling only. Added `heal` + `hot` cases mirroring `effect_dispatch.ts` exactly:
- direct heal: `directHealBonus(spellPower, castTime)`
- pure HoT: `hotTickBonus(spellPower, duration, interval) x ticks` for the displayed total
- `0` for a HoT that rides a direct heal (Regrowth), matching the sim's `hybridHeal` double-dip guard

## 2. Lightning Shield reflect line: raw `{school}` placeholder
The reflect line read **"Deals 29 {school} damage to attackers"** (see screenshot in the issue). The ability tooltip appends the aura effect summary passing only `{kind, value}`, so `auraEffectTooltipHtml` had no school to splice into `{school}`.

`auraEffectTooltipHtml` now resolves the school from the **source ability** (authoritative and always present client-side), and the call site passes the ability id. Same fix also corrects the **online buff-frame tooltip**, where the wire mirror drops an aura's school and the reflect read "Physical" instead of "Nature".

## Notes
- Both reuse existing i18n keys (`hudChrome.abilityScaling.bonus`, `hudChrome.auraEffect.school.*`), so no new strings.
- Tests: `ability_damage.test.ts` now covers heal / pure-HoT / hybrid-HoT scaling (replacing the stale "heals do not scale in this PR" test). tsc + biome + architecture green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1603, opened 2026-07-07
fix(ui): make ability tooltips match their actual effects

ryan-foo:
## Summary

A full pass over every ability tooltip to make it match what the cast actually does. An audit of all 157 ability descriptions against the sim's combat math (`effect_dispatch.ts`, `casting_lifecycle.ts`, `spell_scaling.ts`) found six classes of drift; this PR fixes the display pipeline, the descriptions, and the translations, and adds a guard test so the drift cannot come back.

**Pipeline fixes** (`src/ui/ability_damage.ts`, `hud.ts`, `entity_i18n.ts`):

- Heals and pure HoTs now show the `(+N)` Spell Power callout that combat actually adds (`directHealBonus` / `hotTickBonus`); a HoT riding a direct heal stays suppressed, matching the sim's no-double-dip rule. (This absorbs #1243, including its Lightning Shield `{school}` placeholder fix.)
- `groundAoE` (Earthquake, Consecration), `sunder` (Sunder Armor, Expose Armor), and `lifeTap` are now readable by the damage placeholder. Before, those tooltips rendered an empty string where the number should be: "searing nearby enemies for  Holy damage".
- Channelled AoE (Rain of Fire, Hurricane, Volley) now uses the channel-tick coefficient for its `(+N)`, matching the per-tick combat path.
- Three new placeholders join `$d`: `$o` (over-time total), `$b` (resolved buff value), `$t` (resolved duration). All fill from the rank- and talent-resolved ability, so the prose updates as the player ranks up.

**Content fixes** (`classes.ts` + the EN catalog + all 18 carrying locale overlays):

- Hybrid nukes (Immolate, Flame Shock, Garrote) no longer splice the direct-hit number into the sentence that describes the DoT. Immolate rank 3 used to read "Burns the enemy for 11 Fire damage and an additional 38 over 15 sec": the 11 was stale rank-1 data and the 38 was the direct hit shown as the burn (real burn total: 60).
- Rank-drifting hardcoded numbers (Power Word: Shield's absorb, Battle Shout's attack power, Lay on Hands' heal, Life Tap's conversion, the weapon imbues, and about twenty more) are now placeholders.
- Factual corrections: bear form says armor +90% (the real `recalcPlayerStats` multiplier; it claimed +65%), faerie fire says 30 sec (the real sunder aura duration; it claimed 40), expose armor no longer claims combo-point scaling it does not have, Bloodthirst/Shield Slam state their 60%/50% weapon damage, and the channelled/pulsing AoEs state their real cadence.

**Guard test** (`tests/ability_tooltip_consistency.test.ts`): every placeholder must resolve at every rank, only the four supported placeholders may appear, and every bare number in a description must hold at every rank (a number that drifts with rank must become a placeholder). Combat-constant numbers are allowlisted with their source of truth named.

**i18n:** the changed descriptions are re-translated in all 18 overlays that carry them (token parity verified per key per locale); generated tables and hashes regenerated.

## Related issues

Closes #1243 (its heal/HoT scaling and Lightning Shield school fixes are included here verbatim).

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: `npm run gate` (i18n gen + freshness, malware scan, changed-files biome, full vitest, tsc, env/server/client builds). Final full-suite run after rebasing onto the current `release/v0.23.0` tip: 855 test files, 10159 tests, all green.
- Validation: the audit itself was codex-driven (two independent passes: the display-pipeline math vs `effect_dispatch.ts`/`casting_lifecycle.ts`, and a per-ability sweep of all 157 descriptions), and a final codex review of the finished diff verified the math mirroring, the picker/renderer sharing, the description claims against the combat code, and token parity across all 18 overlays. Verdict: pass.
- Targeted: `npx vitest run tests/ability_tooltip_consistency.test.ts tests/ability_damage.test.ts tests/localization_coverage.test.ts tests/i18n_completeness.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/guide.test.ts`
- Manual steps: booted the offline game before and after, hovered the audited tooltips in the spellbook at max level (screenshots below, captured by the committed `scripts/tooltip_parity_shot.mjs`).

## Screenshots / recordings

Burning Pact (Immolate) rank 3, before and after: the stale 11, and the direct hit shown as the burn total.

| Before | After |
|---|---|
| before | after |

Holy Ground (Consecration): the empty damage number, now the real per-pulse range with its Spell Power callout and cadence.

| Before | After |
|---|---|
| before | after |

Last Rite (Lay on Hands) rank 2, before and after: hardcoded 250 vs the real 600 plus the new heal Spell Power callout.

| Before | After |
|---|---|
| before | after |

More pairs in `docs/screenshots/tooltip_parity_*` (Life Tap, Rain of Fire, Devotion Aura).

---

### Quality

- [x] **Tests pass.** `npm test` is green (15 known load flakes pass in isolation). New guard suite added; `ability_damage` heal/HoT coverage extended.

### Cross-platform

- [x] Tooltip content only; layout and interaction untouched. Desktop verified via the screenshot script; the tooltip component itself is unchanged.
- [x] **Accessible.** No new UI surface; the spellbook aria summaries pick up the same corrected text.

### Localization (i18n)

- [x] **New player-visible strings are translated into every supported locale.** No new keys; the ~50 changed descriptions are re-translated in all 18 overlays that carry them, with per-key token-set parity verified. `localization_coverage` and `i18n_completeness` pass.
- [x] Numbers keep flowing through `formatAbilityNumber` / the i18n formatters.
- [x] Sim/server emit text untouched; `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] Generated i18n artifacts regenerated via `npm run i18n:gen` + `npm run i18n:hash -- --write`, never hand-edited.


jgyy (commented):
This is a genuinely good fix, and the scope matches the claim. I diffed against the actual merge base (`git merge-base pr-1603-review upstream/release/v0.23.0`), not the raw two-branch diff, since the branch predates a couple of unrelated features (Bank window, Town Focus): the naive diff makes it look like this PR deletes those, but `git merge-tree` confirms the real merge auto-resolves cleanly and keeps both, so that's a non-issue, just a stale-branch artifact.

What I checked and confirmed:

- `src/sim/content/classes.ts`: every changed hunk touches only the `description` string field. No numeric gameplay leaf (`value`/`value2`/`value3`/`dmgPct`/`cooldown`/effect fields) moved. The "bear form +90%" and "faerie fire 30 sec" fixes are text catching up to values that were already live (`recalcPlayerStats` 1.9x, the `effect_dispatch.ts` 30s sunder window), not balance changes. Good, this stays within the "don't invent balance numbers" invariant because nothing numeric changed.
- `src/ui/ability_damage.ts`: the new placeholder pickers (`abilityPrimaryEffect`, `abilityOverTimeEffect`, `abilityBuffValue`, `abilityDurationValue`) are pure functions over `ResolvedAbility`, shared between `hud.ts` and the new guard test, which is the right way to prevent the picker logic from drifting between the two call sites.
- `tests/ability_tooltip_consistency.test.ts`: walks every rank of every class ability via `abilitiesKnownAt`, checks every placeholder resolves and every bare number in a description matches a resolved-effect leaf at that rank, with a short, individually-justified allowlist for true constants that live outside effect data (charge stun, taunt window, stance threat bonus). This is exactly the kind of guard that keeps the fix from regressing.
- The spellbook live-refresh commit (`tickOpen`, `rerenderPreservingView`) is a legitimate companion bug: since the talents window no longer closes the spellbook, an open spellbook could show stale cost/cast/cooldown after a talent point. The content-signature approach (id/rank/cost/castTime/cooldown) avoids rebuilding on every online snapshot echo (reference equality would thrash), and scroll/focus are explicitly preserved across the rebuild.
- i18n: the catalog edits (`src/ui/i18n.catalog/abilities.ts`) are English-only description string changes, as required; the 18 locale overlays are the maintainer's non-English fills to review at release time, not something I checked here.
- Ran `npx vitest run tests/ability_tooltip_consistency.test.ts tests/ability_damage.test.ts tests/localization_coverage.test.ts tests/i18n_completeness.test.ts tests/architecture.test.ts tests/spellbook_window.test.ts` locally on the PR head: 117 passed, 3 skipped, all green. `npx tsc --noEmit` clean.
- Merge conflict scope: the only real conflict against current `release/v0.23.0` is the generated `src/ui/i18n.status.summary.json`, which is mechanical regen churn (merge base + rerun `i18n:gen`/`i18n:hash --write` clears it), not a design problem.

Nits, non-blocking:

- `abilityPrimaryEffect` / `abilitySecondaryEffect` in `ability_damage.ts` are exported but I didn't see `abilitySecondaryEffect` consumed outside the guard test and `hud.ts`'s fallback path; worth a second look that both pickers stay reachable from the same call site so they can't diverge silently, but this is already covered by the guard test's own strategy so it's a nit, not a gap.
- The spellbook's `knownSig` string concatenation runs every frame while the window is open; fine at ability-list scale (dozens of entries), just flagging it as the kind of thing to watch if the known list ever grows much larger.

Nice piece of work: the picker-sharing design and the rank-indexed guard test are the right shape for a bug class like this (text drifting from the sim's real math), and the fix is scoped tightly to tooltip text and its resolution, not gameplay.

PR #1604, opened 2026-07-07
fix(delves): clear restlessPending on module advance to prevent softlock

Pepijnvdliefvoort:
## Summary

`restless_graves` is a Heroic affix for the Collapsed Reliquary that raises a weak Bonewalker ~3 seconds after a regular mob dies. The 3-second delay created a softlock:

1. Player kills the last mob in room N -- the spawn is queued in `run.restlessPending`.
2. All mobs are dead, so the exit portal opens.
3. Player walks through; `advanceDelveModule` calls `spawnDelveModule`, which resets `run.mobIds` and spawns room N+1's mobs -- but **did not clear `run.restlessPending`**.
4. Three seconds later `tickDelveRestlessGraves` fires, spawns the Bonewalker at the old room's position, and pushes its id into the new `run.mobIds`.
5. The Bonewalker is alive in a room the player has already left. `tryOpenDelveExitPortal` sees a live mob in `run.mobIds` and never opens the exit for room N+1. Permanent softlock.

Fix: add `run.restlessPending = [];` to `spawnDelveModule` so any spawns queued for the old room are discarded on module transition.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/delves.test.ts` -- 136/136 pass.
- Manual steps: Entered the Collapsed Reliquary on Heroic, with `restless_graves` rolled, killed the last mob in the first chamber and immediately walked through the tombstone passage. Confirmed the second chamber cleared normally and the exit opened. Repeated across several seeds to exercise different module orderings.

## Screenshots / recordings

---

### Quality

- [x] **Tests pass.** Added regression test `restless_graves pending spawn is discarded on module advance (no softlock in next room)` to `tests/delves.test.ts`. `npm test` is green.
- [x] **Typecheck passes.**
- [x] **Builds succeed.**

### Cross-platform

- ~Tested on desktop and mobile.~ N/A -- sim-only change, no UI.
- ~Accessible.~ N/A.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/delves/runs.ts
+++ b/src/sim/delves/runs.ts
@@ -485,6 +485,10 @@ export function spawnDelveModule(ctx: SimContext, run: DelveRun): void {
   run.exitPortalOpen = false;
   run.rewardChestId = null;
   run.surfaceExitId = null;
+  // Discard any pending restless_graves spawns from the room just left. If they
+  // were allowed to fire after the module transition they would land in the new
+  // room's mob list and permanently block the exit-portal clear check.
+  run.restlessPending = [];
   clearDrownedLitanyBossState(run);
   clearDrownedLitanyRiteState(run);
   run.litanyBaptistry = undefined;
--- a/src/sim/delves/runs.ts
+++ b/src/sim/delves/runs.ts
@@ -486,6 +486,10 @@ export function spawnDelveModule(ctx: SimContext, run: DelveRun): void {
   run.exitPortalOpen = false;
   run.rewardChestId = null;
   run.surfaceExitId = null;
+  // Discard any pending restless_graves spawns from the room just left. If they
+  // were allowed to fire after the module transition they would land in the new
+  // room's mob list and permanently block the exit-portal clear check.
+  run.restlessPending = [];
   clearDrownedLitanyBossState(run);
   clearDrownedLitanyRiteState(run);
   run.litanyBaptistry = undefined;
--- a/tests/delves.test.ts
+++ b/tests/delves.test.ts
@@ -695,6 +695,43 @@ describe('delve interactables and affixes', () => {
     expect(run.restlessPending.length).toBe(0);
   });
 
+  it('restless_graves pending spawn is discarded on module advance (no softlock in next room)', () => {
+    // Regression: spawnDelveModule did not clear restlessPending. A spawn queued
+    // in room N would fire after the player moved to room N+1, land in run.mobIds,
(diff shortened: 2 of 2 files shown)


EnriqueGF:
:) Thanks for the clean, well diagnosed fix Pepijn.

I reviewed this end to end against the real code on a checkout of the branch and it is production ready.

**Base branch gate:** passes. The PR targets `release/v0.23.0`, which is the latest release branch, and it merges cleanly against its current tip (verified with `git merge-tree`, no conflicts) even though the branch is based on an older point of that release.

**What I verified:**
- **Root cause and fix placement.** `spawnDelveModule` is the single choke point reached by both run init and `advanceDelveModule`, so clearing `run.restlessPending` there discards stale room-N spawns exactly when the module transitions. Correct seam, minimal change.
- **No intra-tick race.** In `tickDelveRun`, `tickDelveModuleExit` (which drives `advanceDelveModule` into `spawnDelveModule`) runs at runs.ts:889, before `tickDelveRestlessGraves` at runs.ts:893. On the tick the player steps through the portal the pending list is cleared first, so the restless tick is a no-op that same tick. No stale Bonewalker can leak into the new room.
- **Determinism.** `tickDelveRestlessGraves` draws no rng (it creates mobs via `nextId++`), and clearing an array touches no rng stream. `tests/parity` stays green (162 passed), so the global draw order is unchanged.
- **Regression test is decisive.** I removed the one-line fix and the new test fails as expected (`restlessPending.length` is 1 instead of 0, and it also asserts no live `reliquary_bonewalker` appears after ticking past the 3s delay). With the fix restored, `tests/delves.test.ts` is 136/136.
- **Guards and typecheck.** `npx tsc --noEmit` clean; `tests/architecture.test.ts` green; the S3 i18n guard (`tests/localization_fixes.test.ts`) green (no new player-visible strings, as claimed).
- **Hygiene.** No secrets, no dev-command gating touched, no generated files hand-edited, no em or en dashes or emojis in the diff. Module-first respected: an edit to an existing `delves/runs.ts` function, no monolith growth.

Nothing blocking. One optional note for a tidy merge: the branch sits a few hundred commits behind the release tip, so a rebase before merge would keep history clean, but it applies without conflict as is.

Approving.

jgyy:
Closing — this looks already implemented. `src/sim/delves/runs.ts` on `upstream/main` already clears `restlessPending` on module advance. Thanks for the contribution.

PR #1607, opened 2026-07-07
fix(talents): stop percent mods from rounding away ratio-valued aura buffs

Rubsey:
## Problem
Many players reported druids shifting into Fleet Form and moving at plain run speed, and a restoration shaman recently reported the same with Shadewolf. Respeccing appeared to clear it in some cases. The healing-spec correlation in the reports is real and is the key to the bug.

## Root cause
`abilitiesKnownAt` folds talent modifiers into each resolved ability via `scaleEffect`, whose `selfBuff`/`buffTarget` arm applied `Math.round(value * dmgMult + flat)` to every aura kind. That is correct for flat magnitudes (armor, attack power, stamina) but destroys kinds whose value is a multiplier or a 0..1 fraction:

- A restoration mastery (druid or shaman, +14% healing) makes `healMult !== 1`, which triggers the scaling pass. Speed buffs scale by `dmgMult`, which is still 1, so Fleet Form / Shadewolf's 1.4 becomes `Math.round(1.4) = 1`: the form applies, the visual shows, and the player runs at exactly normal speed.
- A balance/elemental mastery (+10% spell damage) rounds the same 1.4 up to `Math.round(1.54) = 2`: double speed.
- A combat rogue mastery (+10% melee damage) rounds Ghostfoot's 0.5 dodge fraction up to 1 (100% dodge for the duration) and deletes stealth's 0.5 movement penalty.

This also explains the respec observation: wiping talent points removes class-tree percent mods, so the next shift resolves the unscaled 1.4 again. A character whose percent comes from the spec mastery itself needs to leave the spec, which is why respec only sometimes cleared it.

## Fix
Add `RATIO_VALUED_AURA_KINDS`, the set of aura kinds whose value is a multiplier or fraction rather than a flat magnitude (speed, forms, stealth, dodge, haste, attack speed, vulnerability fractions, and the rest of the documented ratio kinds), and pass those effects through untouched in both `scaleEffect` and the `buffPct` arm of `applyTalentMods`. Flat magnitude buffs keep scaling exactly as before (pinned by tests). No live `buffPct` talent targets a ratio kind, so that guard changes nothing today.

Notes:
- Auras are not persisted (`serializeCharacter` saves no aura array), so no migration is needed and no stale broken values survive a deploy; the next shift after the deploy is correct.
- The change is in the shared sim core, so offline, server, and headless RL hosts all pick it up identically. Parity golden-trace gate is green.

## Test plan
- [x] New `tests/talent_aura_multiplier_scaling.test.ts`: RED before the fix (6 failures reproducing the reported symptoms, including `moveSpeedMult === 1` for a resto druid in Fleet Form and a resto shaman in Shadewolf), GREEN after
- [x] Pins that flat-magnitude scaling is unchanged (Oakhide armor under spell damage, Steadfast Aura under `buffPct`)
- [x] Full local suite green (860 files / 10256 tests), including the parity golden-trace gate
- [x] `tsc --noEmit` clean, changed files `biome ci` clean

### DIFF
--- a/src/sim/content/classes.ts
+++ b/src/sim/content/classes.ts
@@ -1,4 +1,4 @@
-import type { AbilityDef, AbilityEffect, PlayerClass, Stats, WeaponInfo } from '../types';
+import type { AbilityDef, AbilityEffect, AuraKind, PlayerClass, Stats, WeaponInfo } from '../types';
 import type { TalentModifiers } from './talents';
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jgyy:
Closing — this looks already implemented. `classes.ts` on `upstream/main` already has `INTEGRAL_BUFF_KINDS` handling this, so percent mods no longer round ratio auras incorrectly. Thanks for the contribution.

PR #1611, opened 2026-07-07
feat(sim): add 3D voxel density field engine as a foundation for tunnels

jgyy:
## Summary

First phase of a voxel-terrain refactor: adds a true 3D voxel **density field** layered on top of the existing `(x,z) -> y` heightfield, plus a seam-free chunked mesher.

This PR previously included a temporary verification swap of the production terrain renderer (`renderer.ts`) to `buildVoxelTerrain(...)`, plus a 22-location screenshot tour across all three zones and every dungeon door / delve marker, to prove the engine's output against the real world before it ships anywhere. That swap has done its job and **has now been reverted in this PR** (not deferred to the merge): `renderer.ts` builds `buildTerrain(...)` again, and the screenshot tour (which showed the temporary voxel renderer, not the shipped client) has been removed from this description since it no longer represents in-tree behavior. `voxel_terrain.ts` stays in the tree, unwired, as a foundation for the production-render phase.

```mermaid
graph TD
  A["world.ts terrainHeight(x,z,seed) existing 2.5D heightfield"] --> B["voxel.ts voxelDensity(x,y,z,seed) = y - terrainHeight(x,z,seed)"]
  C["content/tunnels.ts hand-authored capsule waypoints (test fixture only)"] --> B
  B --> D["voxel_mesh.ts meshVoxelChunk() chunked Surface Nets, overlap-padded, outward-facing normals"]
  D --> E["VoxelMesh positions / normals / indices"]
  E --> H["voxel_terrain.ts buildVoxelTerrain(seed) whole-world chunk grid (unwired, foundation only)"]
  B -.future phase.-> F["colliders.ts / pathfind.ts 3D collision + tunnel pathing"]
  C -.future phase.-> G["real tunnel content rendered in-game"]
```

- **`src/sim/voxel.ts`**: `voxelDensity(x, y, z, seed)`, pure and deterministic. Outdoors it is exactly `y - terrainHeight(x, z, seed)`, so it is byte-identical to the existing heightfield and every existing (x,z)-only consumer (`colliders.ts`, `pathfind.ts`, `mob/locomotion.ts`) is untouched.
- **`src/sim/voxel_mesh.ts`**: a chunked Surface Nets mesher with correct outward-facing normals.
- **`src/render/voxel_terrain.ts`** (unwired foundation, not called from `renderer.ts`): meshes the entire playable world from `voxelDensity`, using overlap padding to close chunk seams; kept in-tree for the next phase.

## Chunk seam contract

Vertex agreement between adjacent chunks is not the same as a seam-free surface: `meshVoxelChunk` only emits a face quad when all four surrounding cells are inside the chunk, so a chunk meshed to exactly its own bounds never emits the quads that straddle a neighbor, cracking the shared face even though every boundary vertex value matches. `voxel_terrain.ts` closes this by meshing every chunk one voxel cell past its own bounds on every side (`OVERLAP_VOXELS`), so the overlap band is meshed by both neighbors and the shared face is covered twice instead of zero times. The module comment and tests now state this contract directly instead of the stronger (incorrect) claim that plain vertex agreement is sufficient.

## Real GPU, ultra graphics tier verification (historical, not shipped)

Earlier revisions of this PR verified the engine with a real-GPU screenshot tour (not software rendering) at the `ultra` graphics tier, confirming 45-60 FPS and no seam cracks across all 22 locations, including the previously-flagged thin-sliver spot at the peaks north rim. That verification build is not part of the shipped diff; see the commit history for the tour script and captures if you want to re-run it locally.

## Test plan
- [x] `npx vitest run tests/voxel.test.ts tests/voxel_mesh.test.ts` -- density purity, tunnel carving, mesh determinism, normal direction, exact-abutting-cracks-but-overlap-padding-seals seam behavior
- [x] `npx vitest run tests/architecture.test.ts` -- `src/sim/` purity (no DOM/render imports, no `Math.random`) still holds
- [x] `renderer.ts` builds `buildTerrain(...)` (production heightfield renderer), confirmed by reading the diff -- no verification swap remains


FernandoX7 (commented):
This is honest, well-scoped foundation work. The density field is a pure function of (x, y, z, seed), sim purity and determinism hold (no Math.random, no wall-clock, fixed-order loops; the architecture guard passes at your head), the capsule carve is correct CSG subtraction, tunnels are declarative data-as-code, and I checked the Surface Nets corner/edge tables and the X/Y/Z quad-winding parity: the math is right. The commit history shows real debugging (the inverted gradient normal, the resolution-mismatch seams, the overlap padding) rather than claims, and the 22-location real-GPU tour is exactly the evidence a terrain change needs. CI is green and all 33 tests pass at your head.

Findings:

**blocking: the verification swap has done its job, revert it in this PR, not after.** `src/render/renderer.ts:1206` still builds `buildVoxelTerrain(...)` behind a stubbed `TerrainView`. You marked it revert-before-merge yourself and the tour evidence is captured, so the revert should land as a commit here rather than a thing the merge has to remember. While the swap is live it degrades the shipped client in concrete ways:
- `src/render/voxel_terrain.ts:11-15` loads all 8 albedo textures unconditionally and always uses the 8-sampler triplanar `MeshStandardMaterial`, with no graphics-tier gating; production `terrain.ts` drops to plain Lambert with zero splat textures on the low tier. The module comment openly assumes a real hardware WebGL2 context, which the shipped client cannot promise (software-GL and low-end devices exist).
- The stub no-ops `update()`, which in production is the per-frame fog-distance chunk cull (`terrain.ts:990`, called from `renderer.ts:2059` and `4598`), so fully-fogged terrain becomes pure overdraw.
- The world editor constructs this same `Renderer` (`src/editor/3d/viewport.ts:181`) and opens on the stub: the sculpt brush ring (`setBrush`/`clearBrush`) and the live per-drag region re-mesh (`rebuildRegion`, via `renderer.ts:4961`) are silent no-ops, and the full-rebuild arm (`renderer.ts:4979`) still calls `buildTerrain(...)`, so a map load swaps the voxel terrain back to the heightfield mid-session.
- The fixture tunnel carves a real opening near (60, 150) while grass/props/entities still seat against `terrainHeight`, so the PR's "nothing is shipped as new in-game content" only holds once the swap is gone.
Once reverted, `voxel_terrain.ts` staying in-tree unwired is fine for a foundation slice.

**should-fix: the mesher's "why chunk boundaries never crack" comment describes a property the bare mesher does not have.** `src/sim/voxel_mesh.ts:1-13` argues world-space purity makes exact-abutting chunks seam-free. Vertex agreement is real, but the three face passes (lines 160-203) iterate only interior grid edges, so a quad whose four cells span two chunks is emitted by neither chunk. I verified this empirically: partition a sphere volume into 8 exact-abutting octant chunks and the vertices match the single-chunk mesh exactly (416 = 416), but 204 of 828 triangles are missing and zero triangles bridge the shared plane (the single mesh has 72 there). Your own render side knows this: the overlap padding in `voxel_terrain.ts` is what actually closes the seams (the padded variant produces bridging triangles again), and the hairline-crack commits corroborate it. Since this is the foundation module future phases build on, the comment should state the real contract: chunks meshed to exactly their own bounds crack at shared faces, and callers must mesh overlap-padded volumes the way `voxel_terrain.ts` does.

**should-fix: the seam test is vacuous.** `tests/voxel_mesh.test.ts:42-53` filters vertices to `|x| < 1e-6` on the shared plane, but Surface Nets vertices are cell-interior averages and never land on the chunk boundary: for the sphere fixture both filtered sets are empty (the closest vertex sits at |x| = 0.5, dead center of the boundary cell), so the containment loop runs zero times and asserts nothing. It would pass with the two surfaces arbitrarily far apart. A decisive replacement is cheap and pins the mechanism the renderer really relies on: mesh an exact-abutting pair and an overlap-padded pair over the same volume, and assert the padded union has triangles bridging the shared plane (asserting the compared sets are non-empty in whatever agreement check remains).

**should-fix: nothing asserts normal direction, and that is the exact bug this PR already hit once.** Commit `a02b9555d` fixed `gradientNormal` computing the reverse finite difference; re-introducing that bug passes all 9 tests today because no test reads `mesh.normals`. One assertion on the sphere fixture closes it: for every vertex, `dot(normal, vertexPos) > 0`.

Nits:
- `tests/voxel.test.ts:39-49`: the tunnelBounds test only checks each waypoint center is inside the box, which any radius >= 0 satisfies; an implementation that dropped the radius expansion still passes. Assert `b.minX <= w.x - w.radius` and friends.
- `tests/voxel.test.ts:11-12`: the module claims density is exactly `y - terrainHeight` outdoors, but the test only brackets it at +/-1; `toBeCloseTo(y - terrainHeight(x, z, seed))` pins the equality. Similarly the two determinism tests double-call a pure function in one process; a golden literal would also catch wrong-but-deterministic drift.
- `src/sim/voxel.ts:57`: the radius lerp (tunnel taper) is never sampled mid-segment; every tunnel test hits a waypoint exactly.

Big picture, for the follow-up phases: terrain is not render-only here (the server runs `terrainHeight` for movement and pathfinding through the same sim), so 3D collision and tunnel-aware pathfinding land in the deterministic sim, and the production renderer pass needs LOD plus a place in the tier budgets (the verification build's ~200 merged draw calls would eat about half the lowest tier's whole call budget, and terrain is not a bucket the FPS governor can shed today). One small cross-host note for then: `voxel_mesh.ts` uses `Math.hypot`, which is not guaranteed correctly rounded across engines; fine while the mesher is render-only, but if mesh output ever becomes authoritative sim state, prefer `Math.sqrt`. None of that blocks this slice once the swap is reverted; with the revert and the seam-test fix this is a solid foundation to build on.

jgyy:
Thanks for the thorough review. Pushed a follow-up commit (`a13acf7ea`) addressing everything:

- **blocking (verification swap):** reverted. `renderer.ts` builds `buildTerrain(...)` again; the `buildVoxelTerrain(...)` stub and its `TerrainView` no-ops are gone, so graphics-tier gating, the per-frame fog LOD cull, and the editor's live sculpt-brush re-mesh are all back to production behavior. `voxel_terrain.ts` stays in-tree unwired, as you said is fine for this foundation slice.
- **should-fix (seam comment):** rewrote the module comment in `voxel_mesh.ts` to state the real contract: chunks meshed to exactly their own bounds crack at the shared face because the face-quad passes only fire when all four surrounding cells are inside the chunk; callers must mesh overlap-padded volumes, as `voxel_terrain.ts` does.
- **should-fix (vacuous seam test):** replaced it with a decisive pair of tests: `bridgingTriangleCount` counts triangles with vertices on both sides of the shared plane. One test asserts exact-abutting chunks produce zero bridging triangles (the crack), the other asserts one-voxel-cell overlap-padded chunks produce at least one (the seal) -- reproduces your sphere-octant finding directly.
- **should-fix (no normal-direction test):** added a regression test asserting `dot(normal, position) > 0` for every vertex on the sphere fixture, so a reintroduced inverted-gradient bug (like `a02b9555d`) fails CI.
- **nits:** `tunnelBounds` test now asserts the box edges against `waypoint +/- radius`, not just that the center is contained; the outdoors-density test now pins `voxelDensity(x, y, z, seed)` to the exact `toBeCloseTo(y - terrainHeight(...))` equality at a non-surface `y`, instead of only bracketing it.

Left as-is for a future phase, per your big-picture note: 3D collision/pathfinding, LOD + tier budget for the eventual production render pass, and the `Math.hypot` note in `voxel_mesh.ts` (fine while render-only).

Also updated the PR description: the verification-swap section now states it's reverted, and removed the 22-shot screenshot tour images since they showed the temporary swapped renderer rather than the shipped diff.

`npx vitest run tests/voxel.test.ts tests/voxel_mesh.test.ts tests/architecture.test.ts` and `npx tsc --noEmit` both clean at the new head.

PR #1612, opened 2026-07-07
fix(server): REALM_NAME set only in .env was silently ignored

Blaine1705:
## What

A realm named via `REALM_NAME` in `.env` always booted as the default realm. `server/realm.ts` resolves `REALM` into an exported const at import time, but the `.env` load lived inside `server/db.ts`'s module body, which evaluates after `realm.ts` in `main.ts`'s import graph, so the variable was not set yet when it was read.

## Fix

- New `server/env.ts`: the one `.env` / `.env.local` bootstrap (`loadServerEnv()`, runs once on import), with a comment warning that any module reading `process.env` at import time needs it imported above.
- `server/main.ts` imports it FIRST, before anything that can pull `realm.ts`.
- `server/db.ts` replaces its inline `process.loadEnvFile` calls with the same bootstrap, so entry points that only import the db keep their existing behavior.

## Tests

- `tests/server/env_bootstrap.test.ts`: pins the loader behavior (loads `.env` from cwd, no-op without one) and the import order in both `main.ts` and `db.ts` (the order test is the one that reproduces the bug: it fails on the old tree).
- `tests/server/http/characterization.test.ts`: pins `REALM_NAME` before importing main. With the bootstrap actually working, a developer's machine-local `.env` realm would otherwise leak into the goldens; this mirrors how the file already handles the Discord env keys.

Found while standing up a LAN realm: `REALM_NAME=Lan-mode` in `.env` kept booting as Claudemoon. Verified end to end on Windows: with only `.env` set, `/api/status` now reports the configured realm.

### DIFF
--- a/server/db.ts
+++ b/server/db.ts
@@ -30,20 +30,9 @@ import { USER_ASSETS_SCHEMA } from './user_assets_db';
 // consumers (the market tests) keep importing it from ./db unchanged.
 export { marketStateKey } from './market_backfill';
 
-try {
-  process.loadEnvFile?.();
-} catch {
-  // .env is optional; production usually injects DATABASE_URL directly.
-}
-try {
-  // Local-dev convenience: also load .env.local so the server can reuse the
-  // client's VITE_* values (e.g. the Solana RPC + $WOC mint) for the in-world
-  // holder-tier reads. Existing keys from .env are not overwritten. In
-  // production these come from real env vars (SOLANA_RPC_URL / WOC_MINT).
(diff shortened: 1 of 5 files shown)


EnriqueGF:
@FernandoX7 As I dont have access to production I'd love to know If this can cause problems, do we have any REALM_NAME set?

jgyy:
Closing as already fixed: server/realm.ts now has a DEFAULT_REALM_NAME fallback, so REALM_NAME set only in .env is no longer silently ignored. Thanks for the contribution.

PR #1613, opened 2026-07-07
feat: Protect Yumi, a 3v3/5v5 maze objective PvP mode

MasterZensei:
## Summary

Protect Yumi is a new unranked team PvP bracket pair (yumi3/yumi5) built on the arena system, the fiesta pattern: each team guards a passive 5000 HP cat familiar in a fixed competitive maze; every 60 seconds both cats teleport simultaneously to separated maze points; the first team to destroy the enemy cat wins. Downed players bench for a flat 10 seconds and revive beside their cat. At 10 minutes sudden death latches: teleports freeze and an escalating damage ramp plus bleed guarantees a winner (pre-pulse HP, then damage dealt, then a per-match rng coin; never a draw).

## How it is built

- One sim, three hosts: all match logic lives in src/sim/social/yumi.ts behind SimContext, with state riding ArenaMatch.yumi and a per-match rng stream, so existing golden traces stay byte-identical (parity suite green).
- The maze (src/sim/yumi_maze_layout.ts) is a deterministic fixed-seed 13x13 braided grid (rotationally symmetric, zero dead ends, five plazas) in its own instance band (x 8000..12000, two-sided so the Vale Cup practice pitches at x 30000 stay disjoint), with collision, spell LoS, camera occlusion, minimap raster, and renderer walls all derived from the one layout.
- Hostility is two small arms in isHostileTo/isFriendlyTo: the enemy cat is attackable, the own cat heal/shield-targetable through the existing friendly-cast path; player PvP rides the existing arena machinery.
- Wire: ArenaInfo.match.yumi on the IWorld seam plus four per-pid SimEvents (a 1/s yumiStatus heartbeat carries the live numbers, so the online client needs no mirror changes). No new commands or delta keys.
- Presentation: torch-lit night-arena maze with team beacons, team arrows over every fighter, a match HUD strip (both cat bars, match clock to sudden death, teleport countdown, collapse toggle, phone-portrait flow), Coliseum queue tabs, maze minimap raster, wiki section, and full i18n with the five non-Latin fills.
- Yumi wins are excluded from the arena daily-reward task (unranked objective mode; fiesta keeps its historical counting).
- Yumi ships as a dedicated Meshy-generated cat familiar (scale-baked, emissive cleared, meshopt + 1024 webp: 28.3 MB source to 479 KB, in line with the other creature GLBs). Her one authored clip, the block, is mapped as the hit reaction, so she blocks whenever she is struck; docs/prd/protect-yumi-assets.md tracks the remaining optional dressing assets.

## Screenshots

Match start at the blue spawn plaza: team arrows over every fighter, the match strip with both cat bars, the clock to sudden death, and the teleport countdown.

Match start

A torch-lit corridor right after the simultaneous teleport: the relocation banner, the maze minimap with the mode title, and the enemy cat's nameplate down the hall.

Corridor after a teleport

Victory: the enemy cat falls, the strip clears, and everyone ports home after the aftermath.

Victory

Yumi herself, blocking while under attack (her Meshy block clip rides the hit-react slot).

Yumi blocking under attack

Phone portrait (375px): the clock row wraps above two fluid bars with the collapse toggle.

Mobile HUD

## Verification

- New suites: tests/yumi_maze_layout.test.ts (19 invariants incl. BFS connectivity + band edges), tests/yumi_match.test.ts (queue, hostility matrix, teleports, respawn, sudden death, teardown, same-seed replay), tests/yumi_match_view.test.ts. Full suite 9.8k+ green locally except the known Windows-only new_endpoint tsc-spawn failure; tsc and all five builds green.
- Browser E2E: scripts/yumi_maze_verify.mjs (needs npm run dev) drives the whole mode in a real browser: queue six, maze build, real-keyboard wall containment, own-cat heal, engage via the right-click path, simultaneous teleport with separation, kill to victory banner, teardown, plus HUD checks (banner suppression, m:ss clock, collapse, 375px fit); 20/20 green including at ?gfx=high.
- Post release-merge audit (release/v0.23.0 with Vale Cup merged in): both features' suites green side by side (598 tests), the maze band capped east of the practice pitches, and reviewer agents (architecture/determinism + cross-host parity) reported zero blocking findings.


FernandoX7 (commented):
Really strong piece of work. The sim slice is a faithful copy of the fiesta two-stream pattern: the whole mode stays off the shared RNG on its own per-match stream, so the parity goldens stay byte-identical, and it lands behind SimContext as a proper module (live ctx views, append-only callbacks, no method cluster grafted onto sim.ts). The maze generator is a clean deterministic leaf, the wire needed no ClientWorld changes (s.arena mirrors match.yumi verbatim, the four events drain through the generic queue and every one carries a pid so they route per-participant, never realm-wide), and the enemy-cat attackability fix through activePvpOpponentIds is exactly right given the cat rides the wire with hostile=false. i18n is complete on the English side (every sim-emitted string has a matcher rule, S3 green), graphics fairness holds (beacons and both cat bars paint on every tier), and the daily-rewards exclusion is pinned by a decisive test with a 2v2 control. I ran the yumi suites, the architecture/purity guard, the S3 guard, and the parity goldens against the PR head: all green.

A few things to tighten, none blocking.

**should-fix**

- `CREDITS.md` is not updated for the new Meshy asset. `public/models/creatures/yumi_cat.glb` ships in this PR, but the Meshy-creature attribution row (currently the Drowned Litany models) is not extended to include it, and `CREDITS.md` is not in the diff. This repo credits each Meshy creature by filename, so add `yumi_cat` to that row.

- The headline "never a draw" branch is untested. The simultaneous double-kill path in `src/sim/social/yumi.ts` (the `dieA && dieB` case that calls `resolveYumiTiebreak` and kills the loser's cat) is never exercised: the sudden-death test in `tests/yumi_match.test.ts:330-331` uses uneven hp (240/120), so the cats bleed out sequentially and the winner falls out of the ordinary path. A winner/loser inversion on the kill line (`winner === 'A' ? catB : catA`) would leave both the `resolveYumiTiebreak` unit test and the sequential sudden-death test green while the wrong team wins a same-tick double death. Add a case with equal hp below the first pulse (and equal damage-dealt) that asserts exactly one cat dead, `state === 'over'`, `draw === false`, and the survivor matches the resolver.

- The three match timers are self-compared. `tests/yumi_match.test.ts` asserts `down.seconds` against `YUMI_RESPAWN_SECONDS` (:315) and drives the teleport/sudden-death clocks off `YUMI_TELEPORT_EVERY`/`YUMI_SUDDEN_AT` (:249, :335), all imported from the same module the production code uses, so a balance edit (respawn 10 to 30, cadence 60 to 30, sudden 600 to 300) moves both sides and every assertion still passes. Pin the magnitudes to literals the way `YUMI_HP` already is (`expect(YUMI_RESPAWN_SECONDS).toBe(10)` etc., or assert the emitted `seconds === 10`).

**nits**

- `src/sim/social/yumi.ts:10` says "a bout can never draw," but a same-tick mutual total disconnect resolves to `winner = null` in `updateArena` and `endArenaMatch` emits `arenaEnd` with `draw: true`. It is inconsequential (unranked, nobody left to see it) and the competitive paths are all genuinely draw-free, but the absolute claim is technically false; soften the wording or resolve mutual abandonment deterministically.

- A few coverage gaps beyond the two above: the hostility matrix has no per-dimension negatives for a benched or dead controller, or the countdown/over states (`yumiCatHostileTo`/`yumiCatFriendlyTo`); the non-takedown bench safety at `src/sim/combat/damage.ts:434` (a yumi fighter bottomed out by a non-cross-team source benches instead of dying) has no test; and there is no golden pinning the maze seed or shape, so an accidental `YUMI_MAZE_SEED` change would silently rotate the "identical every match" map while the structural invariants still hold. All would-slip-silently, none is the actual fix.

- `src/sim/yumi_maze_layout.ts:8` still describes a "4.5yd pitch: 3.5yd clear corridors + 1yd walls"; the constant is now 6.75 (5.75 clear) after the 1.5x rescale. The constant's own comment on `YUMI_MAZE_PITCH` is correct; just the file header is stale.

- Optional, and inherited from fiesta so not a regression: the per-match seed `ctx.tickCount * 2654435761 + matchId * 40503` loses low-bit precision once the product passes 2^53 (~47h of 20 Hz uptime), thinning seed entropy on long-lived servers. It stays deterministic across hosts, so this is only an entropy note; `Math.imul`-based mixing would preserve it if you ever revisit both call sites.

One thing I confirmed rather than flagged: narrowing `isDelvePos` to `x < YUMI_BAND_X_MIN` changes collision-band routing for `x >= 12000` (the Vale Cup practice pitches at 30000 no longer match it), but practice players are clamped to their pitch copy and the real match runs at origin, so it is inert and parity is green. The two-sided band cap is the right call.

FernandoX7 (commented):
One more on a closer pass of the render/wire seam, and this one I would call should-fix:

**The cat objective beacon lags the cat by up to ~10s online, but is instant offline.** `src/render/yumi_maze.ts` `update()` sets each beacon from `world.arenaInfo?.match?.yumi.yumiA/yumiB.x/z` only, and that structure rides the arena wire at `ARENA_WIRE_HZ = 0.1` (`server/game.ts:214`, gated at `:3884`), so online it refreshes every 10s. The `yumiTeleport` event carries fresh `toX/toZ`, but the only consumer is the VFX burst in `renderer.ts` (~:2980); nothing folds the new position into the beacon, and `yumiStatus` (the 1 Hz heartbeat) carries HP and timers but no cat x/z. Offline the Sim's `arenaInfo` getter is polled every frame, so the beacon snaps on the same tick.

Effect: after each 60s teleport (`YUMI_TELEPORT_EVERY`), the online beacon marks the old cell for up to 10s while the cat sits at the new one. The cat entity model and the minimap marker stay fresh at 20 Hz when the cat is inside interest range, but the maze corner-to-corner (~90yd footprint, ~127yd diagonal) can exceed `NPC_INTEREST_RADIUS = 120`, so from a far corner the stale beacon is the only indicator, on a mode whose whole loop is hunting the enemy cat. The beacon's own docstring calls its position "actionable info," so it should not be host-dependent.

Fix is small: fold `yumiTeleport`'s `toX/toZ` into a client position cache the beacon reads (the position half of the event was plumbed, just never consumed), or let the beacon read the wired cat entity when it is in range. Since the cat only ever moves on a teleport, folding the event gives an instant online update.

Everything else in my first pass stands; the rest of the parity spine (verbatim `s.arena` mirror, per-pid event routing, no new Entity/wire field, format enumeration, the i18n matchers) checks out.

MasterZensei:
Going to do those, the next step after is testing it on the dev server with 6 and preferably 10 players.

EnriqueGF:
## Review: APPROVED (production-ready)

Thorough pass over the whole diff plus a determinism, tri-host parity, and server/security fan-out. No blocking or should-fix issues. This is a well-built, self-contained mode that respects the repo's seams and invariants.

### What I verified

**Determinism and the SimContext seam.** Match logic is a sibling module (`src/sim/social/yumi.ts`) behind `SimContext`, with state riding `ArenaMatch.yumi` and live `ctx` views on `Sim`; five append-only callbacks bound with correct arg order, no method cluster on `sim.ts`. All randomness goes through `Rng`: the maze uses a local seeded `Rng`, and the per-match stream `new Rng((tickCount*2654435761 + matchId*40503)>>>0)` follows the fiesta two-stream rule, so the shared draw order is untouched and golden traces stay byte-identical (parity suite green). The sudden-death tiebreak (pre-pulse HP, then damage dealt, then the per-match coin) is deterministic and never draws.

**Tri-host parity.** `ArenaInfo.match.yumi` is implemented in both Sim (`yumiMod.yumiMatchInfo`) and mirrored structurally by `ClientWorld`, sent in the rate-limited arena field. The four new per-pid SimEvents (`yumiStatus` heartbeat, `yumiDown`, `yumiSuddenDeath`, `yumiTeleport`) are routed to the owning pid and consumed by the HUD, so the "no ClientWorld mirror changes" claim holds: structure rides the arena field, live HP/timers ride the heartbeat.

**Server authority, security, persistence.** `msg.format` is sanitized to a fixed allowlist (unknown falls back to 1v1). Win/loss resolves entirely server-side; `endArenaMatch` sets `ranked = !fiesta && !yumi`, so Yumi never moves the Elo ladder. The daily-reward exclusion returns 0 for yumi3/yumi5 before any write, and `format` is the server-set match format, so a Yumi bout cannot be laundered into a ranked grant. No secrets, no raw SQL, no schema or migration changes; the new queues and `ArenaMatch.yumi` are runtime-only and `ArenaFormat` is widened additively.

**i18n.** Every sim emit has a matcher in `sim_i18n.ts` (including the leader-only wildcard); the S3 drift guard is green. The `yumi_cat` template resolves through `world_entity_i18n`.

**Gameplay.** The loop is coherent and free of major loose ends: symmetric braided maze (no dead ends), simultaneous 60s cat teleports keep offense from camping a fixed point, a flat 10s bench avoids permanent elimination, and sudden death at 10:00 (frozen teleports plus an escalating damage-taken ramp and neutral bleed) guarantees a winner with no draw. Turtling on a passive 5000 HP cat is therefore hard-capped.

**Tests.** yumi_match, yumi_maze_layout, yumi_match_view, architecture, parity, daily_rewards, and the S3 localization guard all pass locally.

### Non-blocking notes (for a later pass, none gate merge)

1. The enemy cat is drawn on the minimap and its position is exposed in `yumiMatchInfo` for both teams, so there is no fog of war on the objective. The comment in `minimap_painter.ts` says this is deliberate; worth a quick confirmation that always-visible enemy-cat position is the intended feel rather than a scout-and-find dynamic.
2. Balance of the 5000 HP passive cat versus a focused 3 to 5 player burst is only exercised mechanically by the E2E, not tuned; it deserves a real playtest to confirm a rush cannot trivialize the PvP.
3. The daily-reward exclusion is a hardcoded `yumi3 || yumi5` string check rather than a shared `isRewardable(format)` allowlist. Fine today; a future third unranked format would have to remember this site. Consider centralizing.
4. Minor wording: the `yumi.ts` header says "nothing here touches the shared sim stream" and "never draws", which is true for the shared/golden stream; the tiebreak can draw once from the per-match stream on an exact tie. Reads fine, flagged only so a future reader does not misread it as absolute.

Nice work. Approving.

EnriqueGF (approved):
Approved. Reviewed for determinism/SimContext, tri-host parity, server authority, security, persistence, i18n, and gameplay soundness; all green. Merged release/v0.23.0 into the branch (kept both the Protect Yumi and Vale Cup daily-reward tests, regenerated the i18n tables); tsc, parity, yumi, daily_rewards, and the i18n guards pass locally.

PR #1609, opened 2026-07-07
Talent Choices engine + warrior rows (collaboration branch for Talents 2.0)

Blaine1705:
# Talent Choices engine + warrior rows (collaboration branch for Talents 2.0)

Draft PR so we can collaborate on one branch, as agreed on Discord. This is Champete's playtested talent-rows work, rebased conceptually on `release/v0.23.0` (full merge, all gates green). Not intended to merge as-is yet; it is the shared starting point.

### Choice-row engine (MoP-style)
- `src/sim/content/talent_rows.ts`: the row framework. Six choice rows at levels 5/8/11/14/17/20 (`ROW_LEVELS`), one pick per row, respec-free swapping out of combat.
- `rowPicks` on `PlayerMeta`, persisted and validated server-side (`pickRowTalent` is combat-locked and level-gated; the server re-validates every pick).
- `computeModifiersWithRows` folds row picks into the same flat `TalentModifiers` pipeline the point tree uses; hot paths never walk the tree.
- Full three-host parity: offline `Sim`, authoritative server, `ClientWorld` mirror (`tal.rowPicks` on the snapshot self), plus the golden-trace parity gate regenerated on the merged code.
- The old point tree is intentionally KEPT alongside the rows until all 9 classes have rows.

### Warrior: 18 live row options (all playtested on a live realm)
Levels 5 to 20, three options per row: Victory Rush, Double Charge (2-charge system on ability cooldowns, wire key `chg`), Bladestorm (full-duration uninterruptible channel with spin animation), Battle Trance (20% per-swing proc, next Heroic Strike or Slam free), Redhand builder, Sanguine Aura, Direhowl, Second Wind, Piercing Howl, Lingering Dread (fear with break-on-damage threshold), and more. Balance numbers follow classic-era formulas.

### Base-kit additions (warrior)
- Pummel at level 12: free, and refunds 10 rage when it actually interrupts a cast (`rageOnInterrupt`).
- Heroic Leap at level 16.
- Rallying Cry at level 18: +20% max HP for allies within 40 yards (`aoeAllyMaxHp` + `buff_maxhp_pct`, current HP scales with the recalc).

### New sim mechanics the rows needed (all generic, reusable by other classes)
- `Entity.charges` (multi-charge cooldowns) + `maxCharges` on `armAbilityCooldown` + timer refund.
- `requiresAuraKind` ability gate and `selfHealPctMax` effect (Victory Rush).
[...]

PR #1614, opened 2026-07-07
feat(talents): the Talents 2.0 flip, choice rows replace the point trees (all 9 classes)

ryan-foo:
## The Talents 2.0 program (read this first)

This PR is **S2, the flip, and the program's trunk**: it CONTAINS #1543 (S1: signatures + masteries) and everything since (proc engine, three row-quality waves, talent-moment VFX, adversarial balance suites, the salvaged fun-pass roadmap in docs/prd/talents-2.0-fun-pass.md). #1348 is closed with a salvage ledger; #1609 is the consolidation partner for the playtested warrior tunings. Still open from the fun-pass design: the S3 streak engine (noteSpellHit seam), Hot Streak, Ice Lance, facing enforcement.

## Summary

The Talents 2.0 tree flip: choice rows become the ONLY talent allocation and the legacy point trees are deleted. Design source: `docs/prd/talents-2.0.md` (Part 4, PR5). Builds on the 27 spec signature spells + spec masteries shipped in #1543 (this branch contains that work; it is the S1+S2 flip on top).

DRAFT on purpose: this is the consolidation base for the choice-row work, targeted at a PTR realm for open playtesting before anything reaches the main realm. See also @Blaine1705's #1609 (the same engine plus 18 live-playtested warrior rows): the plan is to consolidate into one branch, with the playtested warrior tunings and mechanics from #1609 merged on top of this base, credited to their author.

What the flip changes:

- `TalentAllocation` becomes `{ spec, rows }` (ranks/choices deleted). `validateAllocation(cls, alloc, playerLevel)`: rows unlock at levels 5/8/11/14/17/20, no point budget.
- `computeTalentModifiers` = spec signature grant + mastery (level-scaled `min(1, level/20)`, re-baked on every ding) + `accumulateRowEffects`. Combat never walks trees.
- 162 row options across all 9 classes (`choice_rows_classic.ts`), 51 ported ability defs, every granted spell cast-proven per class (`tests/row_grants_cast.test.ts`, 53/53).
- Migration: legacy `{ ranks, choices }` saves load clean and drop to a free re-pick (`{ spec, rows }`); build strings move to `TALENT_BUILD_VERSION = 2`, v1 imports are rejected as incompatible.
- UI: the talents window becomes Specialization | Choices tabs (Choices default); tree painting deleted; pure view core rewritten (`talents_view.ts`).
- i18n: full coverage for the new option names and readouts (option-name title overrides in every locale dict, CJK/Cyrillic hand-verified for language purity).
- Determinism: parity goldens regenerated; per-frame draws and drawDigest are byte-identical to `release/v0.23.0`'s goldens in every scenario (state/events-only drift).

## Related issues

Part of the Talents 2.0 program (`docs/prd/talents-2.0.md`). Companion to #1543 (signatures + masteries) and #1609 (consolidation partner). Supersedes the design direction of #1348.

## Type of change

- [x] Breaking change (talent build strings v1 rejected; legacy saves migrate to a free re-pick)

## How was this tested?

- Commands: `npm run gate` (full suite, tsc, builds, changed-file biome, i18n freshness) green on the merged tree; `UPDATE_PARITY=1 npx vitest run tests/parity/parity.test.ts` then draw/drawDigest verified byte-identical vs the release goldens; targeted suites: `tests/talents.test.ts`, `tests/talents_view.test.ts`, `tests/choice_rows_engine.test.ts`, `tests/spec_masteries.test.ts`, `tests/spec_signatures.test.ts`, `tests/row_grants_cast.test.ts`, `tests/localization_fixes.test.ts`, `tests/talent_tooltip_accuracy.test.ts`, `tests/architecture.test.ts`.
- Manual steps: scripted 8-step playthrough (`scripts/talents_flip_playtest.mjs`): level a warrior 1 to 20 picking rows at each unlock, spec pick grants the signature, legacy-save migration to free re-pick, clear/respec, then a mage run (frost rows + freeplay combat). Screenshots below are from that run.

## Screenshots / recordings

All nine classes, the Choices tab at level 20 with one pick per row, plus a combat moment with a row buff live:

| Class | Picker | Moment |
|---|---|---|
| Warrior | w | wm |
| Paladin | p | pm |
| Priest | pr | prm |
| Shaman | s | sm |
| Mage | m | mm |
| Rogue | r | rm |
| Hunter | h | hm |
| Druid | d | dm |
| Warlock | wl | wlm |

Talent-moment VFX (new light + particle systems, captured live):

| Moment | When it fires | Shot |
|---|---|---|
| Proc surge | an every-Nth-cast empowerment arms | f1 |
| Ward bloom | a big-hit ward or cheat-death save | f2 |
| Echo burst | a stored heal-echo fires below 35% | f3 |
| Detonation | Earthen Jolt eats the Cinder Jolt DoT | f4 |

---

### Quality

- [x] **Tests pass.** `npm test` is green. I added or updated tests for any `sim/` or `server/` behavior I changed.

### Cross-platform

- [x] **Tested on desktop and mobile.** Talents window verified at phone-sized viewports in both orientations.
- [x] **Accessible.** Keyboard-operable rows with roving focus, visible focus ring, ARIA on tabs and options.

### Localization (i18n)

- [x] Numbers and percents go through the i18n formatters.
- [x] Player-facing text emitted from `src/sim/` is re-localized at the client boundary (`sim_i18n.ts` rows-readout matcher), `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets, credentials, or `.env` committed; `ALLOW_DEV_COMMANDS` untouched in production paths.
- [x] Generated files (parity goldens, i18n resolved tables, guide content, sha256, status summary) regenerated by the build scripts, not hand-edited.


FernandoX7 (commented):
This is a big, load-bearing change (the tree flip plus the signature and mastery work under it), and it holds together unusually well. I reviewed against the head ref, not the description, and the load-bearing claims check out:

- Determinism is preserved byte-identically. In the regenerated goldens the top-level `drawDigest` and every per-frame `rng.draws`/`rng.digest` are unchanged; only the `state` hashes move (mastery and spec numbers). The new machinery earns that: `sweptReposition` is rng-free, `chainHeal`'s hop pick is an exact min over (hp fraction, distance, id) so it is order-independent with no rng draw, and the `incapacitate` generalization to `ability.fearDr` keeps the fear draw order intact.
- Migration is safe. A legacy `{spec, ranks, choices}` save loads to `{spec, rows:{}}` (a free re-pick), the spec is preserved when the level allows it, `accumulateRowEffects(... alloc.rows ?? {})` tolerates the missing field so nothing crashes, and v1 build strings are rejected by the `TALENT_BUILD_VERSION` bump rather than throwing at the user.
- Server authority holds. `talentAllocationFromWire` is shape-only; the real gate is `validateAllocation(..., r.e.level)` against the true server-side level, and rows cannot grant a spec signature, so a tampered client cannot unlock a gated row or self-grant a signature.
- Wire and seam are clean. `critDmgBonus` is a server-resolved combat stat, correctly kept off the wire but present in `blankEntity` for shape parity; aura kinds ride the wire as strings so the 12 new `AuraKind` values inserted mid-union are order-safe; `resolveMove` is exposed on the `SimContext` seam (append-only) and bound from `Sim`. New sim files import sim-only, and the new pure view core is in `UI_PURE_CORES`.

Findings:

**should-fix (correctness, dev-gated)** `src/sim/sim.ts:2892` (`setPlayerLevel`). It sets `r.e.level`, then re-runs `recalcPlayerStats(..., this.playerMods(r.meta))`, but never re-bakes `meta.talentMods`. `playerMods` just returns the cached `meta.talentMods`, so the spec-mastery magnitude (now scaled by `Math.min(1, level/20)`) stays frozen at the old level's scale until the next apply, respec, ding, or relog. The ding path was deliberately patched for exactly this (`src/sim/combat/damage.ts:722` re-bakes `computeTalentModifiers(meta.cls, meta.talents, p.level)` before the stat pass), so the level-setter is an asymmetric miss. It is behind `ALLOW_DEV_COMMANDS` and stales identically on every host (not a determinism break), but `/dev level` is how PTR testers will jump characters, so their masteries would read wrong until a respec, which undercuts the playtest. Fix: `r.meta.talentMods = computeTalentModifiers(r.meta.cls, r.meta.talents, r.e.level)` before the `recalcPlayerStats` call.

**should-fix (a11y)** `src/ui/talents_window.ts:257-292`. The Choices-tab picker sets `role="radiogroup"` with `role="radio"` option buttons and `aria-checked`, but has no roving `tabindex` and no arrow-key handler. The sibling spec tab in the same file (`paintSpecTab`, ~300-341) implements the full radiogroup contract (roving tabindex plus a `rovingTarget` keydown handler), so this reads as an oversight: every option becomes its own tab stop and arrows do nothing inside the group, which is the interaction the checklist's "roving focus" line promises. It is still operable (native buttons, Tab reaches them, Enter/Space picks, visible focus ring), so not blocking. Related: `tests/browser/keyboard_nav.browser.test.ts` dropped the old choice-node roving tests and added nothing for `.tal-row-opt`, so the behavior is unpinned even once roving lands.

**nit** `src/sim/combat/effect_dispatch.ts:40`. `hasCastShield` is imported but never used in this file (only `noteSpellHit` and `spellDamageMultFromAuras` are; the real `hasCastShield` caller is `casting_lifecycle.ts:182`). Dead import, worth dropping so the changed-files biome pass stays quiet.

**nit** `src/net/online.ts:2725,2746`. The optimistic `computeTalentModifiers(this.cfg.playerClass, this.talents)` omits the level argument, so it defaults to `MAX_LEVEL`. Harmless (the granted-ability set is level-independent and the next `tal` snapshot overwrites `known`), but passing `this.player.level` would make both sites self-evidently correct.

**nit** The `value > 1 ? value / 100 : value` percent-normalization is copied in four places (`src/sim/entity.ts:180` as `pctValue`, `src/sim/combat/auras.ts:63`, `src/sim/sim.ts:3242`, `src/sim/pet/pet_ai.ts:219`). Rule of three: one shared helper would be worth it, and it pins the one ambiguous point (a value of exactly 1 reads as 100%, not 1%).

**note** `noteSpellHit` in `src/sim/combat/spell_combat.ts` is a deliberate no-op (Hot Streak deferred to the S3 streak engine) called at several combat sites. Fine as scaffolding for the draft, just flagging that it is currently inert so it does not get lost when the streak engine lands.

Given this is a draft aimed at a PTR realm as the consolidation base, none of the above blocks that goal. Nice work keeping the RNG draw order stable through a change this size.

ryan-foo:
Thanks for a genuinely useful review @FernandoX7, all five landed in cbf463a5c:

- setPlayerLevel now re-bakes talentMods before the stat recalc (mirroring the ding path), with a regression test, so /dev level on the PTR reads masteries correctly.
- The Choices picker has the full radiogroup contract now (roving tabindex + arrow keys via rovingTarget, same pattern as the spec tab), pinned in the window token test and in keyboard_nav.browser.test.ts.
- Both online.ts optimistic sites pass the live level; pctValue is one shared helper with the exactly-1-is-100% pin; the dead import had already fallen out in a later pass.
- noteSpellHit stays as documented scaffolding for the S3 streak engine, agreed on keeping it visible.

Since you reviewed, the branch also grew: all nine classes' flagged row options moved onto a deterministic proc engine (rhythms, wards, echoes, detonations, a real cheat death), Ice Block became a true stasis, and the talent moments got dedicated VFX, with the 9-class picker/moment screenshot matrix in the PR body. Draw order stayed byte-identical through all of it, which your review made much easier to keep honest.

jgyy (commented):
This is a big, well-executed flip and it reads like it. The module boundaries are right where CLAUDE.md wants them: `talent_procs.ts` and `spell_combat.ts` are new sibling modules behind `SimContext`, not method clusters bolted onto `sim.ts`; `choice_rows.ts`/`choice_rows_classic.ts` are pure data-as-code; `talents_view.ts` stays a DOM/Three-free pure core. The row-effect fold (`accumulateRowEffects`) iterates `Object.entries(rows)` on integer-like keys, which JS always orders ascending regardless of host, so that's deterministic without needing an explicit sort. `talent_procs.ts` draws no rng itself (tick-based ICDs/counters); the proc responses that do heal/crit still resolve through the normal `ctx.applyHeal` rng path, which is fine, it's just inserting new deterministic-trigger draws into the existing stream, and the parity goldens were regenerated to match.

Checked the parts that usually bite a flip like this:

- **Wire/parity:** `meta.talents` (`{spec, rows}`) round-trips whole through `tal.alloc` in `server/game.ts` / `src/net/online.ts`; `talentPoints()` is reimplemented identically on both hosts off the new row economy (`rowsUnlockedAtLevel`/`rowsPicked`). No IWorld drift.
- **Server authority:** `talentAllocationFromWire` (`server/game.ts:294`) only does shape normalization; `validateAllocation` re-checks authoritatively in the shared sim module on apply. Build-string import rejects v1 by version tag.
- **Migration:** confirmed in `src/sim/sim.ts:1736` and `server/character_sheet.ts` (`normalizeAllocation`) — both drop legacy `ranks`/`choices` to `rows: {}` and pass through `repairAllocation`, so old saves load clean into a free re-pick, spec preserved when still valid at level. Consistent across both persisted-state entry points.
- **i18n:** new tab/label strings are real `t()` keys in `src/ui/i18n.catalog/game.ts`; the 162 row option names/descriptions follow the existing `talent_i18n.ts` manifest pattern used for the old tree nodes, not a new inline concat. (English-only scope, as usual.)
- **No em dash/en dash introduced.** I grepped the diff for the dash characters: every instance in the `+` lines is an exact-match relocation of a comment that already existed pre-PR (verbatim in the matching `-` line, e.g. the `KnownAbility` doc comment moved between files), nothing new.

Two small things worth a look, neither blocking:

- `tests/talents_window.test.ts`'s roving-tabindex/ARIA coverage for the new Choices rows asserts on painter output structure (`rovingTarget`, `aria-selected`, etc.) rather than driving actual keypresses through focus. It proves the wiring exists but not that arrow-key nav actually moves focus at runtime. Given the pure-core/painter split this is a reasonable trade-off, just flagging it as a should-fix if you want stronger behavioral proof before this reaches the main realm.
- The migration silently resets legacy `{ranks,choices}` allocations to an empty `rows` (free re-pick) with no explicit refund/notice event beyond whatever the client shows on next `/talents` open. Sounds intentional per the PR description ("drop to a free re-pick") and it's consistent across both load paths, just calling it out since it is real, if expected, data loss for anyone who had ranks banked under v1.

Given the DRAFT status and the stated plan to consolidate with #1609 before this touches the main realm, this looks solid enough for that PTR-realm playtesting goal. Nice work threading `computeTalentModifiers`'s new `level` parameter (mastery scaling) through every call site (`fiesta.ts`, loadouts, respec) without missing one.

Diagram of what actually flipped:

```mermaid
flowchart LR
    subgraph OLD["Before: point trees"]
        direction TB
        O1["TalentAllocation<br/>{ ranks, choices }"] --> O2["walk Class + Spec trees<br/>per node, per rank"]
        O2 --> O3["computeTalentModifiers<br/>(no level param)"]
    end

    subgraph NEW["After: choice rows (this PR)"]
        direction TB
        N1["TalentAllocation<br/>{ spec, rows }"] --> N2["spec signature grant<br/>+ level-scaled mastery"]
        N1 --> N3["accumulateRowEffects<br/>162 options, 9 classes<br/>choice_rows_classic.ts"]
        N2 --> N4["computeTalentModifiers(cls, alloc, level)"]
        N3 --> N4
        N4 --> N5["flat TalentModifiers<br/>(baked once, hot path never walks a tree)"]
    end

    OLD -. "TALENT_BUILD_VERSION 1 -> 2<br/>v1 builds rejected<br/>legacy saves -> free re-pick" .-> NEW

    N5 --> S["sim.ts / progression/talents.ts<br/>applyTalentAllocation (server-authoritative)"]
    S --> W["server/game.ts <-> src/net/online.ts<br/>tal.alloc wire round-trip"]
    S --> U["src/ui/talents_view.ts (pure core)<br/>+ talents_window.ts<br/>Specialization | Choices tabs"]
    N3 -.->|"new every-Nth-cast /<br/>threshold procs"| P["src/sim/combat/talent_procs.ts<br/>(new sibling module, no rng)"]

    classDef removed fill:#4a1d1d,stroke:#a33,color:#eee
    classDef added fill:#123a24,stroke:#2a8,color:#eee
    class O1,O2,O3 removed
    class N1,N2,N3,N4,N5,P added
```

ryan-foo:
Superseded by #1756, which is the current Talents 2.0 choice-pass PR on the same branch (targeting the v0.24.0-ptr line). Closing this older duplicate; work continues on #1756.

PR #1615, opened 2026-07-07
fix(terrain): close LOD seams, smooth mountain relief and ground cliff scatter

madmatah:
## Summary

A pass over how the zone-rim and inter-zone mountain walls look and feel, from
issues spotted while playing around Thornpeak Heights and Mirefen Marsh. Five
problems, one common cause: the steep walls are far from every hub and hold the
heightfield's highest frequencies, so they were the worst-served part of the
world. Two commits:

1. `fix(terrain): close LOD seams, smooth mountain relief and ground cliff scatter`
2. `fix(sim): ease the player body off cliff faces (wall standoff)`

**What was wrong, and the fix:**

1. **See-through seams.** At LOD-band borders the coarse side (a chord) and the
   fine side diverged; measured holes reached 3.3yd against a flat 0.3yd skirt, so
   you saw sky through the wall. Skirts now drop by `0.3 + slope * spacing`, and
   every wall chunk is promoted to the densest LOD band regardless of hub distance.
   Re-measured: 44 differing-spacing borders, 0 uncovered gaps.
2. **Sharp, faceted relief.** The wall-chunk LOD promotion resolves the terraced
   risers, and each quad now splits along the diagonal whose endpoints are closest
   in height, so the fold follows a ridge edge instead of sawing across it.
3. **Repetitive checkerboard on distant peaks.** The snow ramp was narrow relative
   to the 6yd terrace step (alternate treads went fully white / fully bare) and the
   macro normal map was sampled unfiltered (distant shimmer). The snow ramp was
   widened and noise-broken; the normal map now uses mipmaps + anisotropy (magFilter
   stays linear, so no loss up close).
4. **Floating rocks and trees.** `generateDecorations` had no slope gate, so scatter
   landed on near-vertical faces and jutted into the air (and large rocks / trunks
   planted invisible colliders there). It now rejects any candidate on terrain
   steeper than the climb limit. On the built-in seed this drops 141 props, all on
   rim/ridge walls, none in the rolling interior.
5. **The player sinking into cliff faces.** Heightfield collision samples only the
   player's center: the climb gate blocks the center from climbing a wall, but
   nothing keeps the body's width clear of one. A body-radius "wall standoff" now
   eases the center off any wall within reach.

**Design notes (the seams touched):**

- **Terrain is render-only; the heightfield is untouched.** Problems 1 to 3 live in
  `src/render/terrain.ts`. Sim and renderer share the same `terrainHeight`, so
  collision, movement, and parity are unaffected: pure presentation.
- **Decorations are one deterministic source, consumed by three systems.**
  `generateDecorations` (`src/sim/world.ts`) feeds the renderer, the minimap, AND
  collision (`colliders.ts`: large rocks and trunks are solid). So the cliff filter
  had to live at the source, not the renderer: a render-only filter would hide the
  visual but leave a phantom collider on the wall and a stray minimap dot.
- **The movement kernel is shared by the server and the client predictor.** The wall
  standoff lives in `stepPlayerMotion` (`src/sim/player_motion.ts`), the kernel both
  the authoritative server Sim and the client self extrapolator
  (`src/render/self_motion.ts`) run. Putting it there (not on `sim.ts`) keeps client
  prediction in lockstep with the server; a Sim-only standoff would rubber-band the
  player near walls. The standoff is a pure ring sample of the shared heightfield
  (`terrainWallStandoff` in `world.ts`), bounded to one body radius, never nudging
  onto steeper ground, a no-op on open ground and flat instanced floors. It draws no
  rng, so the tick draw order and the MV1 kernel parity gate are unchanged.

**Scope / known limits:** the standoff covers the normal WASD + gravity path; charge,
follow, and fear resolve separately (rare, brief, server-authoritative). A
sub-body-radius cosmetic clip can remain when hugging a cliff at a low camera angle
(the model is slightly wider than the collision radius); left as is, within
classic-MMO norms.

## Related issues

N/A (reported in-play, no filed issue).

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run` (full suite, 10,860 passed) and `npx vitest run tests/fixes.test.ts`
    for the new terrain/decoration/standoff tests.
  - `npx tsc --noEmit` clean; Biome on changed files clean.
  - `npm run build`, `npm run build:server`, `npm run build:env` all succeed.
- Manual / numeric steps:
  - Measured LOD-seam holes vs skirt coverage before/after (44 borders, 0 uncovered).
  - Measured the decoration slope distribution (141 props removed, all on walls).
  - Before/after screenshots at the reported vantages (seams, relief, distant
    checkerboard, floating props) on the high graphics tier. The geometry changes
    (skirts, LOD promotion, triangulation, scatter filter, standoff) apply to both
    tiers by construction; the snow/mipmap change is high-tier only (the Lambert
    tier has no normal map or splat snow).
  - The wall-clip case is verified by an end-to-end Sim test rather than a
    screenshot: near walls the climb gate and downhill slide dominate the exact
    resting frame, so the test is the decisive evidence (it fails without the fix).

## Screenshots / recordings

**Seam before:**

seam-before

**Seam after:**

seam-after

**Relief before:**

relief-before

**Relief after:**

relief-after

---

### Quality

- [x] **Tests pass.** `npm test` green; added tests for the `sim/` changes
      (decoration slope filter and the wall standoff) in `tests/fixes.test.ts`.
- [x] **Typecheck passes.** `npx tsc --noEmit` clean.

### Cross-platform

- [x] ~Tested on desktop and mobile / touch targets.~ N/A: no HUD/UI or touch
      surface. Terrain rendering + sim only; geometry changes are graphics-tier
      independent, snow/mipmap is high-tier only.
- [ ] ~Accessible.~ N/A: no UI added or edited.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed; `ALLOW_DEV_COMMANDS` not
      enabled anywhere.
- [x] No hand-edited generated files (`manifest.generated.ts` etc.).

### DIFF
--- a/src/render/self_motion.ts
+++ b/src/render/self_motion.ts
@@ -26,7 +26,12 @@
 // against a real lagging Sim.
 
 import { resolveMovement } from '../sim/colliders';
-import { moveSpeedMult, type PlayerMotionDeps, stepPlayerMotion } from '../sim/player_motion';
+import {
+  BACKPEDAL_MULT,
+  moveSpeedMult,
+  type PlayerMotionDeps,
+  stepPlayerMotion,
+} from '../sim/player_motion';
 import { DT, type Entity, type MoveInput, RUN_SPEED } from '../sim/types';
 
 // Latency cap on the extrapolation window: at least one snapshot-ish interval
@@ -66,6 +71,8 @@ export const SELF_MOTION_DEADBAND_YD = 0.05;
 export const SELF_MOTION_SNAP_DIST_SQ = 6 * 6;
 const MAX_FRAME_DT = 0.25; // matches the main-loop frame clamp
 const LEASH_SLACK_YD = 0.05;
+const BLOCKED_INTENT_PROGRESS_FRACTION = 0.45;
+const BLOCKED_INTENT_MOVE_FRACTION = 0.75;
 // Pose-history ring: enough to look SELF_MOTION_CAP_MAX_MS into the past with
 // headroom even on high-refresh displays (128 entries covers 267 ms at 480 fps
 // and over 2 s at 60 fps).
@@ -92,6 +99,29 @@ interface Vec3Like {
 }
 
 const clamp = (n: number, min: number, max: number): number => Math.max(min, Math.min(max, n));
+
+function horizontalIntent(
+  facing: number,
+  input: MoveInput,
+): { x: number; z: number; step: number } | null {
+  let mx = 0;
+  let mz = 0;
+  if (input.forward) mz += 1;
+  if (input.back) mz -= 1;
+  if (input.strafeLeft) mx -= 1;
+  if (input.strafeRight) mx += 1;
+  const len = Math.hypot(mx, mz);
+  if (len === 0) return null;
+  mx /= len;
+  mz /= len;
+  const sin = Math.sin(facing);
+  const cos = Math.cos(facing);
+  return {
+    x: mz * sin - mx * cos,
+    z: mz * cos + mx * sin,
+    step: RUN_SPEED * (mz < 0 ? BACKPEDAL_MULT : 1) * DT,
+  };
+}
 
 export class SelfMotionPredictor {
   /**
@@ -273,13 +303,30 @@ export class SelfMotionPredictor {
     inp.strafeLeft = frame.moveInput.strafeLeft;
     inp.strafeRight = frame.moveInput.strafeRight;
     inp.jump = frame.moveInput.jump;
+    const intent = horizontalIntent(frame.displayFacing, inp);
+    let blockedHorizontalIntent = false;
     this.acc = Math.min(this.acc + dt, MAX_FRAME_DT);
     while (this.acc >= DT) {
+      const beforeX = actor.pos.x;
+      const beforeZ = actor.pos.z;
       actor.prevPos.x = actor.pos.x;
       actor.prevPos.y = actor.pos.y;
       actor.prevPos.z = actor.pos.z;
       actor.facing = frame.displayFacing;
       stepPlayerMotion(this.deps, actor, inp);
+      if (intent) {
+        const stepX = actor.pos.x - beforeX;
+        const stepZ = actor.pos.z - beforeZ;
+        const progress = stepX * intent.x + stepZ * intent.z;
+        const moved = Math.hypot(stepX, stepZ);
+        const expectedStep = intent.step * moveSpeedMult(actor, 0);
+        if (
+          progress < expectedStep * BLOCKED_INTENT_PROGRESS_FRACTION &&
+          moved < expectedStep * BLOCKED_INTENT_MOVE_FRACTION
+        ) {
+          blockedHorizontalIntent = true;
+        }
+      }
       this.acc -= DT;
     }
     const frac = this.acc / DT;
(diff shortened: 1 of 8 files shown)


FernandoX7 (commented):
This is a strong set of fixes, and the design notes in the description match the code. I verified the load-bearing claims against the branch: the decoration filter lives at the one shared source (placement is hash-based, so the skip shifts nothing for the renderer, minimap, or colliders), the standoff draws no rng and sits inside the shared kernel so the server Sim and the client predictor stay in lockstep by construction, the instanced-floor no-op holds (`groundHeight` returns a flat `DUNGEON_FLOOR_Y` past the threshold, and the predictor is gated off in delves anyway), the wall-band constants match `world.ts` (ridge influence at `RIDGE_SIGMA * 3`, rim onset 30yd inside the edge), and `rebakeNormalRegion` does a full `needsUpdate` re-upload so the new mip chain regenerates. I also mutation-tested the new tests: disabling the decoration gate and the kernel standoff call each fails its test, and the end-to-end Sim test is what catches the kernel wiring. Good coverage shape.

One should-fix and one nit:

- **should-fix** `src/sim/player_motion.ts:325`: the standoff writes `p.pos.x/z` directly, bypassing the collider sweep. Every other horizontal position write in the kernel and Sim goes through `resolveMove` (player_motion.ts:220, sim.ts:3340/3409/3744), which is also the only place fence crossing is enforced. A direct write can push the body into a prop or building collider at a wall foot (the new slope gate clears props off the faces, but not off walkable ground adjacent to one), and can step across a fence line outright; while the player stands still nothing corrects the overlap. Routing the nudge through `deps.resolveMove(p.pos.x, p.pos.z, s.x, s.z, BODY_RADIUS, p, false)` before the walkability guard keeps the convention, and it stays parity-safe: the client predictor binds the real `resolveMovement` (self_motion.ts:145).
- **nit** `src/sim/world.ts:724`: the comment says the constant is "Pinned as a literal by tests/fixes.test.ts", but the test imports `DECORATION_MAX_SLOPE` and filters by it, so a changed threshold would still pass. Either add `expect(DECORATION_MAX_SLOPE).toBe(1.5)` (or use the literal in the test filter) or drop the comment line.

One note on the description, not the code: the widened snow ramp lives in the vertex-color path, and the Lambert tier renders vertex colors too, so the low tier gets the new snow as well. That is fine (the tiers stay matched); only the "snow/mipmap change is high-tier only" sentence is half right (the mipmap part is).

madmatah:
@FernandoX7 Thanks for the careful review and mutation testing.

I addressed the should-fix by routing the terrain standoff through `resolveMove(...)` before applying it, so the nudge now preserves the same collider/fence convention as the rest of the movement kernel. While testing the wall-foot case further, I also found a small sawtooth when forward intent and terrain standoff fought each other at a wall. The standoff now preserves tangential slide, but no longer lets a repeated wall push undo forward progress once the body is already clear.

I also tightened the local self-motion predictor for blocked horizontal intent so it does not visually lead into an obstacle and then reconcile back while the key is held.

For the nit, I pinned `DECORATION_MAX_SLOPE` explicitly in the test instead of relying on an imported threshold indirectly. I will also update the PR description wording around the snow ramp: the mipmap part is high-tier only, but the vertex-color snow ramp applies to the low tier too.

FernandoX7 (approved):

PR #1616, opened 2026-07-07
test(sim): pin mob corpse loot decay

### LINKED ISSUE #1539: Un-looted mob loot is permanently lost when the mob respawns in place
## Problem

When a mob is killed but not looted in time, its corpse and loot are destroyed the moment the mob respawns in place. A mob's corpse is the same entity as the mob (the dead entity carries its own `loot`/`lootable` state), so the in-place respawn reuses that entity id and clears the loot. Any un-collected drops on that specific corpse are lost permanently, with no second chance.

Today the window is roughly 60 seconds (the corpse timer). After that the mob respawns in the same spot and the loot is gone. This affects only the skipped corpse; looting other mobs afterward still works.

## Steps to reproduce

1. Kill a mob that drops loot, and do not loot it.
2. Wait for the corpse to decay and the mob to respawn in the same spot (about 60 seconds).
3. Try to loot. The loot option is gone and the drops are unrecoverable.

## Actual behavior

- The un-looted corpse is removed and its loot is cleared when the mob respawns in place.
- Drops the player had loot rights to are lost, not recoverable from the next kill's fresh roll.

## Expected behavior

- The intended corpse-decay behavior for un-looted loot is confirmed and documented, matching classic-era MMO expectations.
- A player who tapped a mob gets a fair, clearly bounded window to loot before the corpse decays.
- If loot is going to be destroyed on respawn, that outcome is predictable (the corpse visibly decays and the loot affordance behaves consistently), not a surprise where an alive-again mob sits where the un-looted corpse was.

## Scope

- Corpse/loot lifetime and the respawn gate: `src/sim/mob/locomotion.ts` (respawn condition), `src/sim/mob/lifecycle.ts` (`respawnMob` clears loot), `src/sim/combat/damage.ts` (`handleDeath` sets `corpseTimer`/`respawnTimer`), `src/sim/loot/loot_roll.ts` (loot generation).
- Loot rights and the loot command: `src/sim/interaction.ts` (`lootCorpse`).

## Acceptance criteria

- The intended behavior for un-looted corpses on respawn is confirmed and documented (working as designed, or a change is made).
- If a change is made, an un-looted corpse is not silently overwritten by an in-place respawn before its documented window ends.
- `tests/mob_lifecycle.test.ts` asserts that `respawnMob` clears loot state (`lootable` false, `loot` null, `tappedById` null, `lootRecipientIds` unset), not just position/hp/state. This side effect is currently unpinned.

## Notes

- Root cause is structural, not a stale-client bug. The client loot affordance and the server `lootCorpse` gate both re-derive from live `dead`/`lootable` each frame, so once the mob is alive again there is correctly no loot option. The loss happens in the sim when `respawnMob` clears `loot`/`lootable` on the reused entity.
- Current window: a lootable corpse blocks respawn until its corpse timer elapses (`respawnTimer <= 0 && (corpseTimer <= 0 || !lootable)`, roughly 60 seconds), so the player has about a minute. This may already be intended corpse-decay behavior; filing to confirm and to close the test gap.
- Scope confirmed by the reporter: only the skipped corpse is affected; looting later kills still works.

loopermallee:
## Summary

- Documented the finite corpse window for normal unlooted mob loot.
- Pinned `respawnMob` cleanup for reused mob entities.
- Added regression coverage for lootable corpse blocking, expired corpse loot cleanup, and non-lootable corpse early respawn in `tests/mob_lifecycle.test.ts`.

## Related issues

Closes #1539.

## Type of change

- [ ] Bug fix
- [x] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/mob_lifecycle.test.ts tests/mob_locomotion.test.ts tests/loot_ffa_sim.test.ts` (passed, 3 files / 26 tests)
  - `npx vitest run tests/architecture.test.ts` (passed)
  - `npm run check:ts` (passed)
  - `git diff --check` (clean)
- Manual steps: N/A, sim-only behavior and tests.

## Screenshots / recordings

N/A, no UI changes.

---

### Quality

- [ ] **Tests pass.** `npm test` is green. I added or updated tests for any `sim/` or `server/` behavior I changed.
- [ ] **Builds succeed.** `npm run build`, plus `npm run build:server` and `npm run build:env` if I touched those.

### Cross-platform

- [ ] N/A. This PR adds no UI behavior.

### Localization (i18n)

### Hygiene


loopermallee:
The patch documents and tests the existing corpse-decay behavior without changing production behavior.

When a mob dies, `handleDeath` marks it dead, starts corpse and respawn timers, snapshots who can loot it, and rolls loot. `CORPSE_DURATION` is documented as the finite window where unlooted corpse loot can remain available.

While the mob is dead, `updateMob` decrements its corpse, respawn, and loot FFA timers. A normal non-instance mob can respawn only when its respawn timer is ready and either:

- the corpse timer has expired, or
- the corpse is no longer lootable, such as after being fully looted

When respawn happens, `respawnMob` reuses the same mob entity id for the fresh spawn. That is the important cleanup boundary: it clears corpse-specific state like `lootable`, `loot`, `lootRecipientIds`, `tappedById`, and `harvestClaimedBy`, then restores the mob to a live idle state at its spawn position with full HP.

The new tests pin all of that:

- `respawnMob` clears old corpse loot state.
- `updateMob` does not respawn a still-lootable corpse before corpse expiry.
- `updateMob` does respawn after corpse expiry and clears uncollected loot.
- non-lootable corpses can respawn once the respawn timer is ready, even if the corpse timer still has time left.

TLDR: the patch confirms classic-style behavior. **Loot left on a corpse expires with that corpse. A respawned mob is a clean new kill, not a continuation of the old corpse.**

loopermallee:
Credit and thanks to @FernandoX7 for raising #1539.

loopermallee:
Rebase/retarget update:

- Rebased `codex/issue-1539-corpse-decay` onto `upstream/release/v0.23.0` and force-with-lease pushed `f15123a99`.
- i18n artifacts were confirmed unchanged for this non-localization/sim PR; unrelated generated status drift from the release scanner was left out of scope.
- Focused validation passed:
  - `npx vitest run tests/mob_lifecycle.test.ts tests/mob_locomotion.test.ts tests/loot_ffa_sim.test.ts`
  - `npx vitest run tests/architecture.test.ts`
  - `npx tsc --noEmit`
  - `git diff --check`
  - `npm run security:gate`
  - `npm run build`

EnriqueGF:
Thanks for taking this on, and for the clear write-up on the corpse-decay semantics :)

I reviewed this against the real code in a worktree rather than the PR description. The base branch gate passes: this targets `release/v0.23.0`, which is the latest release branch, and the head is correctly rebased on top of it. I confirmed `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, the parity suite, and `tests/mob_lifecycle.test.ts` + `tests/mob_locomotion.test.ts` + `tests/loot_ffa_sim.test.ts` (all green), and that the production diff really is comment-only, so the "no behavior change" claim holds.

I also mutation-tested the new assertions. Three of them are genuinely decisive: dropping `|| !mob.lootable`, dropping `mob.loot = null` from `respawnMob`, and forcing respawn regardless of `corpseTimer` each fail exactly the intended test. That is good work, and better than most pin PRs.

That said, this is a docs-and-tests PR, so coverage and comment accuracy are the whole deliverable, and both have gaps I confirmed by hand.

1. (Should fix) The `respawnTimer` conjunct of the very condition this PR pins is never exercised in its blocking state. `src/sim/mob/locomotion.ts:96` reads `mob.respawnTimer <= 0 && (mob.corpseTimer <= 0 || !mob.lootable)`, and all four new tests set `respawnTimer = 0`. I deleted `mob.respawnTimer <= 0 &&` from the guard and the full `tests/mob_lifecycle.test.ts` still passed 13/13. Please add the negative case: `respawnTimer = 5` with `corpseTimer = 0.01` (or `lootable = false`), asserting `mob.dead === true`. Both OR arms are covered; the AND arm is not.

2. (Should fix) The new comment at `src/sim/mob/locomotion.ts:94-95` is factually incomplete. It says a lootable corpse blocks in-place respawn "only while its corpse window is active", but corpse-window expiry is necessary and not sufficient: the respawn also requires `respawnTimer <= 0`, and it never happens at all for instance mobs (`!isInstanceMob`). As written the comment would mislead the next reader into the same gap as finding 1.

3. (Minor) The new `CORPSE_DURATION` comment at `src/sim/combat/damage.ts:50-51` reads as though 60 seconds is the corpse window, but 60 is only the default. `handleDeath` itself caps it to `template.respawnSeconds` for fixed-respawn mobs and overrides it for world bosses, pets, and demons, and `src/sim/loot/loot_roll.ts` both extends it (`Math.max(corpseTimer, LOOT_ROLL_TIMEOUT + 2)` at lines 284, 330, 506, 643) and shrinks it (`Math.min(corpseTimer, 4)` at line 648 once the corpse is emptied). Since documenting this window is the stated purpose of the change, please say that 60 is the default and name the overrides. The revision also dropped the old "sole user is handleDeath, so the constant lives here with the death-domain code" rationale, which is still true and explained the file placement; worth keeping a clause of it.

4. (Minor) The inclusive boundary is unpinned. Changing `mob.corpseTimer <= 0` to `mob.corpseTimer < 0` leaves all 13 tests green, because the expiry test uses `corpseTimer = 0.01`, which decrements to `-0.04`. A case landing exactly on `0` would pin the comparison this PR is about.

5. (Nit) No test spawns a mob at `spawnPos.x > DUNGEON_X_THRESHOLD` to pin that a dungeon corpse does not in-place respawn. Pre-existing and outside the stated scope, but it is the last unpinned arm of the same condition, so it is cheap to close while you are here.

Nothing here touches i18n, the wire protocol, `IWorld`, persistence, or the renderer, and there are no UI changes, so no screenshots are needed. No determinism or sim-purity concerns: the parity draw digest is unchanged and no rng draw order moved.

Findings 1 and 2 are what I would like to see addressed before merge. Once the `respawnTimer` negative case lands and the locomotion comment names both co-gates, this is good to go.

PR #1617, opened 2026-07-07
fix(mobile): HUD parity for every desktop window, a measured overlap gate, and the charcreate scroll clip

MasterZensei:
## Problem

Continuation of the mobile HUD work from #1578, picking up its named open follow-ups plus
the author's handoff notes. Three concrete problem groups:

1. **Character creation clipped and unscrollable on mobile.** The landscape height clamp on
   `#charcreate-panel` combined with `overflow: hidden` on its columns cut the appearance
   swatches and the Back/Create buttons off with no way to reach them (the panel next door,
   charselect, already solved this with scrollable columns).
2. **Desktop HUD elements without mobile treatment.** Nine windows (calendar, crafting,
   mailbox, emote editor, arena, vale cup, delve board, leaderboard, loot settings) had no
   `body.mobile-touch` layout at all, floating combat text used desktop sizing and rise
   distance on phone viewports, and the chrome idle fade leaked its dimmed state across
   interface-mode switches.
3. **"Manual scanning" for HUD overlaps does not scale.** #1578 shipped after hand-checking
   a handful of layouts; the remaining overlap surface (persistent chrome in a real party
   with a target and auras, every window state, seven device profiles) needs a repeatable
   measured gate, not eyeballs.

## Fix

Nine commits, all mobile-scoped (`body.mobile-touch`, `scripts/`, `tests/`); desktop is
untouched (the FCT change pins desktop `hud.css` byte-identical with a test).

1. **Charcreate scrolls.** The charcreate columns take the charselect treatment
   (`overflow-y: auto`, `min-height: 0`) under the same panel clamp. Verified reachable at
   844x390, 926x428 and 932x430 on both `index.html` and `play.html`; a CSS invariant test
   pins the columns scrollable.
2. **Chrome idle-fade leak fixed.** `dispose()` now un-dims its target and the fade handle's
   lifecycle moved into `setActive` (created on activate, disposed on deactivate), so a
   flip to desktop mode can no longer strand `mobile-chrome-idle` on body. Covered by
   fake-timer tests for the full flip cycle. (Two other suspected chat/joystick state leaks
   were investigated and refuted in code; the audit exercises the chat toggle live.)
3. **Compact floating combat text on touch.** 14px base / 20px crit (still scaled by the
   player's Combat Text Size setting) with a shorter rise via duplicated `-mobile`
   keyframes, keeping the compositor-driven translate animation static per the hud.css
   rule. Only `font-size` and `animation-name` are overridden; the low-fx crit shed is
   re-pointed for the mobile layer. Cosmetic density only: every number still spawns, so
   the graphics-fairness rule holds.
4. **Overlap audit tooling.** The cluster check's geometry (edge gaps, circle math, device
   profiles) moved to a shared `scripts/lib/overlap_geometry.mjs` (behavior-identical,
   proven by byte-identical output), and a new `scripts/mobile_hud_overlap_audit.mjs` runs
   two passes across seven device profiles: a strict persistent-chrome pass (real 5-member
   party via `sim.partyInvite`/`partyAccept`, forced target, buff and debuff auras, meters
   open) with pairwise gap floors, and a window matrix that opens every window through its
   real hud toggle and asserts on-screen bounds plus a reachable 40px close control
   (`MATRIX_ALL=1` runs every window at every profile; `--gate` promotes it to exit-1).
   Settling is a bounded two-identical-samples poll, so repeated runs are deterministic.
5. **Nine windows join the mobile sheet base.** Edge-pinned via the existing grouped sheet
   rule; mailbox and calendar take the market-window full-height template (their inner flex
   bodies own the scroll), the leaderboard gets a max-height cap, and touch floors were
   added only where controls measured under 40px. A new coverage guard scrapes every
   `.window` id from both HTML entries plus dynamically-created windows and fails any
   future window that has neither a real mobile rule (pin/size/floor property required)
   nor an explicit exception with a reason.
6. **Player frame cleared off the Jump button.** The audit's measured findings (not the
   predicted ones): 17.9px of overlap at 740x360 and a 2.1px gap on 4:3 tablets. Fixed
   with tier-scoped seats (a max-width 800px compact override and a tablet nudge, cast and
   swing bars moving in lockstep), landscape-gated so browser portrait keeps the centred
   seat, with the four clean profiles byte-unchanged.

## Round 2: collapsible party chip, chat priority, and a keyboard dismiss

Four more commits, same scoping rules (mobile-only, desktop untouched):

1. **Party frames collapse behind a "Party" chip.** The member stack plus Leave Party
   used to always occupy the top-left; it now sits behind a compact chip (SVG chevron,
   40px floor) at the stack anchor, keeping the below-target offset. Tap to expand the
   full stack under the chip, tap again to collapse; default collapsed, persisted in
   localStorage, and a pure user toggle (never gated on fx tier, reduce-motion, or the
   governor: party HP stays one tap away). State logic is a pure core
   (`src/ui/party_collapse.ts`) with a thin builder and write-elided painter wiring;
   the chip is never built on desktop.
2. **Chat gets the space while open.** The party UI yields to the chip while chat is
   shown (transient, the persisted preference restored on close), and the chat log,
   tabs and composer are excluded from the chrome idle fade so they stay at full
   contrast while in use.
3. **The Android/iOS keyboard can be dismissed without closing chat.** A chevron on the
   composer row blurs the input (dropping the keyboard on both platforms); the
   keyboard-viewport applier reflows chat to its resting seat and re-tapping the input
   raises the keyboard again. The composer blur path is refactored onto a pure seam so
   the intentional dismiss and the close path stay distinct; Enter-send and Escape are
   unchanged. The static button ships in both entries with a base desktop hide.
4. **Review nits closed:** the chat composer draft now survives a redundant touch
   re-activation (a foldable resize crossing the touch-query thresholds), the
   window-transform guard's media-block scan is nesting-proof, and the window coverage
   scrape is attribute-order-tolerant.

Two new `hudChrome` keys (`partyChip`, `hideKeyboard`) ship with their five non-Latin
fills per the M16 wordy-string rule; resolved artifacts and the sha256 pin regenerated.

## Round 3: chat usability, top-band fixes, unit menu, and the 0.85 chrome scale

Three more commits from the next usability pass (all body.mobile-touch, desktop untouched):

1. **Touch-first chat.** The desktop 22px tab strip (scaled 0.92, about 20px effective) and
   its 26px add-channel button now meet the 40px floor with 16px text, keeping the
   issue-1365 invariants (nowrap, pan-x scroll, sticky add); the chatlog seat compensates
   so the 4-line cap is pixel-identical. The composer uses a short mobile placeholder
   instead of the desktop command list, gains padding so text stays clear of the dismiss
   chevron, and the chevron shows only while a keyboard is actually up. The chat-open
   contrast pin now out-specifies the idle fade for the Chat button itself.
2. **Compact Vale Cup chrome.** The queue indicator and in-match strip had no mobile rules
   and floated mid-screen; both get small top-edge seats (the arena-status stacking
   exception preserved).
3. **Consumables flow sideways again** (owner decision recorded in the design comment),
   anchored to the toggle; while a pet is out the row drops one band
   (body.mobile-pet-active, value-diffed) because the pet bar keeps the top-centre line:
   the three-way band contention is not solvable by nudges at 740px. The pet bar itself
   gave up its 0.88 landscape scale so commands measure a true 40x40, verified by tapping
   stances on a real summoned pet.
4. **Double-tap a friendly target opens the unit menu** (Invite to Party, Whisper, and the
   rest), the same menu as desktop right-click, via a slop-guarded double-tap helper; pet
   and hostile targets mirror their desktop menus too. Tall menus now clamp on-screen.
   Invite verified live end to end (party formed offline; the same IWorld call drives the
   online pinvite command).
5. **The whole touch HUD chrome scales to 0.85** behind one token (--mobile-chrome-scale)
   folded multiplicatively into the chrome sizing, composing with the player's button and
   joystick size settings. Every interactive control still measures >= 40px at the new
   default (the compact ring page-toggle clamps at max(40px, ...)); FCT, the composer
   font, the pet bar, menu items, and the modal windows are deliberately excluded. The
   cluster gate (7 profiles, slider extremes, left-handed, safe-area) and the overlap
   audit (default and MATRIX_ALL) pass.

A post-round code review by the maintainer surfaced 7 findings plus 6 nits, all fixed in
the closing commit. The structural one: the party chip could never span the member columns
(grid-column against implicit columns), so 3+ member parties flowed a frame beside the
chip; member rows now live in a .party-rows wrapper under the chip (desktop unchanged via
display: contents, painter pool re-parents with zero steady-state moves), and the audit
now measures the chip against every member row so the nested-pair exemption can never hide
this class again. The audit also became strictly load-bearing on the chat path: real
pointer taps (the old .click() was a no-op), hit-tested chip taps via elementFromPoint,
the simulated keyboard writing the body var exactly like the real applier with a
composer-actually-docked assertion, and a hard failure if the chat cycle completes on zero
profiles.

## Verification

- **Overlap audit**: exit 0 on all seven profiles (iphone-13, pro-max, pixel-7, galaxy-s8
  landscapes; 720p laptop; 4:3 tablet; FHD), including the full `MATRIX_ALL` sweep: 94
  window-open checks, zero violations. Three consecutive clean runs (no flakes).
- **Combat scenario** (844x390 and 740x360, offline, 5-member party, ~26s sustained
  combat): FCT renders 14px/`fct-rise-mobile` with concurrent floaters peaking at 12
  (cap 64); player/target/jump rects drift 0.0px through combat; chat toggle mid-combat
  hides and restores the joystick with no stuck state.
- **Load check**: Docker unavailable locally, so the heaviest offline scene stood in for
  the bot raid: 5-member party plus 20 hostile mobs over three combat cycles on a phone
  viewport. Frame time flat from idle to combat (the absolute floor is SwiftShader
  software GL in headless), FCT bounded, DOM nodes plateau after the pools warm up, JS
  heap stable across cycles.
- **Unit and guard tests**: every changed behavior has a test (charcreate scroll pin,
  chrome-fade flip cycle, FCT desktop-byte-identical pin, window coverage guard, landscape
  gate pin). Full suite green locally except three pre-existing environment items listed
  below.
- **QA gates**: the repo's qa-checklist agent returned READY on the round 1 diff (zero
  blocking, zero should-fix), plus an independent Phase 6 code review whose one real
  finding (portrait orientation gate) is fixed in this branch. Round 2 got its own
  qa-checklist pass: its single blocking finding (a stale i18n sha256 pin) was fixed
  before commit, everything else PASS across all categories.
- **Round 2 verification**: 232 unit tests across the touched suites (48 new or
  changed cases: collapse-state resolver including corrupt-storage fail-safes, painter
  chip lifecycle including desktop neutrality and disband, dismiss-vs-close blur
  boundary, draft-survives-reactivation); the extended overlap audit drives the chip
  collapse/expand cycle, the chat-open yield and clearance, and the simulated keyboard
  dismiss at every profile, exit 0 across all seven.

## Known items, named rather than silently skipped

- **Meters overlay over party/player frames on short landscape phones**: pre-existing
  (matches the accepted bags-over-player-frame pattern from #1578 round 2); the audit
  notes it as an allowed toggled-overlay case; a separate follow-up will re-seat it.
- **Quest tracker in landscape**: deliberately `display:none` (map badges carry quests on
  short landscape); the audit proves the state populates and documents the skip.
- **Stale perf-tour anchor**: `hud_perf_budget` ARM 3 fails at 265 writes vs the 153
  anchor committed 2026-06-24 on a live tour, with a healthy 0.985 elision skip rate; this
  branch touches nothing in the per-frame write path (CSS and event handlers only), so the
  anchor is stale relative to earlier release-branch changes. Left for a maintainer
  re-baseline rather than edited here.
- **Pre-existing browser-suite failures** (outside this diff): `#party-leave` at 28px and
  `.community-toggle` at 0px in the target-size test, and a char-window professions-state
  crash in the a11y test's data setup.
- **`tests/server/new_endpoint.test.ts`** fails on this Windows dev machine (POSIX spawn,
  pre-existing, unrelated).
- **Meters/party vs the Autorun button with no window open**: hardening the audit's chip
  tap to be hit-tested exposed that the leader's auto-opened Loot Settings window (which
  hides #mobile-autorun via mobile-window-open) had always masked a pre-existing
  meters-vs-autorun (and party-vs-autorun at 740px) proximity when no window is up. The
  audit preserves the original measurement context and the meters re-seat follow-up
  already filed covers this corner.

## Screenshots

All current-state captures at 844x390 landscape unless noted (probe-driven offline
sessions, chrome woken out of the idle fade, windows closed first). Every screenshot
from earlier rounds has been replaced with this set, taken at the final commit.

### Character create: bottom reachable again (the fixed clip bug)
The two entry shots are byte-identical PNGs, which is the parity point: index.html and
play.html share the charcreate markup and the one shell.css fix, and render identically.
charcreate create button reachable, index
charcreate create button reachable, play

### Newly-sheeted windows
calendar as a mobile sheet
vale cup sheet at 740x360

### Compact combat text and seated frames in combat
fct in combat
party frames with target

### Player frame cleared off the Jump button (the audit's findings)
player frame vs jump at 740x360
player frame vs jump at 1024x768

### Chat: two-state toggle, party yield, and the keyboard dismiss
chat open, joystick hidden, dismiss chevron on the composer
chat closed, joystick back, party chip restored
chat open, party yielded
keyboard up, dismiss chevron on the docked composer
keyboard-open docked column: tabs, log, composer reserved

### Party chip: collapsed, expanded, and the fixed 3-member layout
party chip collapsed
party chip expanded
3-member party, chip alone on its line

### Round 3: chat tabs, vale cup, consumables, unit menu, pet bar
40px chat tabs and short placeholder
composer text clear of the dismiss chevron
compact vale cup indicator
consumables flowing sideways
double-tap unit menu with Invite to Party
pet bar with a real pet, consumables band-dropped

### The 0.85 chrome scale
full HUD at the new default scale
combat at 740x360


FernandoX7 (commented):
Sorta Solid follow-up to #1578. The scoping discipline is the best part: everything is `body.mobile-touch`, `scripts/`, or `tests/`, desktop is left alone, and the FCT change pins `hud.css` byte-identical with a real test rather than asserting it in prose. Every behavior change lands with a decisive test (charcreate scroll pin, chrome-fade dispose cycle, FCT desktop/mobile values, the window-coverage guard, the landscape-gate pin), and the coverage guard strips comments and requires a real pin/size/floor property so a commented id or a lone z-index cannot spoof it. I ran the six changed test files locally: 155 pass.

Verified the load-bearing claims:

- **Chrome-fade leak fix is correct.** `dispose()` clears `CHROME_FADE_IDLE_CLASS` before stopping the timer, and the handle lifecycle now lives in `setActive` (armed with `??=` on activate, disposed and nulled on deactivate). The `pointerdown` forwarder already read `this.chromeFade` dynamically behind `if (this.active)`, so nulling it in desktop mode is a safe no-op. No stranded idle class across a flip.
- **FCT stays within the fairness rule.** Only `font-size` and `animation-name` are overridden; size and rise distance are exactly the cosmetic-richness knobs CLAUDE.md lets a mode shed, every number still spawns. The low-fx crit re-point (`:root[data-fx-level="low"] body.mobile-touch .fct.crit`) outranks `body.mobile-touch .fct.crit` on specificity, so the low-tier crit shed is preserved on touch too.
- **The 9 newly-sheeted windows are fully covered.** The seven without an explicit cap in the sheet rule still inherit `max-height` from the base `body.mobile-touch .window` rule (`hud.mobile.css:1764`) and get `overflow-y: auto` from the sheet, so none can grow past the viewport with unreachable controls (the charcreate class of bug). mailbox/calendar correctly switch to `overflow:hidden` + inner-body scroll.
- **Lockfile change is the minimal picomatch entry** (17 lines, additive, dev/optional/peer), the same repair #1578 needed. No production surface.
- **Charcreate split is clean:** charselect keeps `overflow:hidden` (its children scroll), charcreate columns get `overflow-y:auto`, panel-scoped ids so they do not collide.

A few nits, nothing blocking:

- `src/game/mobile_controls.ts:499`: `exitChatReply()` now runs on every `setActive(true)`, not only on a real desktop-to-touch transition (there is no `active === this.active` guard). A redundant re-activation while a composer is open would clear the in-progress chat text. The trigger is narrow (the `PHONE_TOUCH_QUERY` width/height thresholds crossing while staying in touch, e.g. a foldable or a coarse-pointer resize), and the old code already reset the chat classes on every activation, so impact is minor, but a transition guard would make it airtight.
- `tests/mobile_window_transform.test.ts`: `mediaPrelude` finds the enclosing media via `lastIndexOf('@media')`, which would grab an inner nested `@media` if one ever sat between the block open and the rule. Fine for the current flat structure, just a touch fragile if these blocks gain nesting.
- `tests/mobile_window_coverage.test.ts`: `windowIdsFromHtml` only matches `id="..."` before `class="...window..."`. All 27 windows in the markup are id-first today so the guard is complete, but a future class-first window would slip the scrape. A one-line reorder-tolerant regex would close it.

The named known-items read correctly to me: the `hud_perf_budget` anchor is genuinely stale relative to this diff (CSS and event handlers only, nothing in the per-frame write path), and the pre-existing target-size / a11y-setup failures are outside the change. Leaving the anchor for a maintainer re-baseline is the right call.

jgyy:
Tested this branch on a real phone (mobile landscape) and found three follow-up issues, now fixed in MasterZensei/world-of-claudecraft#4 (opened against this branch, merged in `release/v0.23.0` and resolved the generated-artifact i18n conflicts along the way):

1. **Character creation was broken on mobile.** `#offline-select`'s appearance-swatch row was the only flex child with an explicit `min-height:0` among its landscape-layout siblings, so on a short landscape phone it collapsed to a few px instead of scrolling, and the swatch portraits (still `overflow:visible`) spilled over the Enter World/Back buttons - blocking character creation. Gave the columns the same `overflow-y:auto` treatment `#charcreate-panel` already uses.
2. **Chat composer wasn't centered.** The active composer inherited the read-only log's left-edge seat. Centered it horizontally; the log itself is untouched.
3. **Left move joystick "moved" oddly on touch.** The wheel intentionally floats exactly under the thumb (a prior clamped-origin attempt caused the shipped v0.22.0 #1229 input-drift regression), but a touch anywhere in the large move zone could teleport the visible wheel clear across the screen. Clamped only the *drawn* position to a small radius of its resting spot; the input origin stays the raw touch point, so there's no reintroduced drift.

Also fixed while testing: the ability/item tooltip popped on every action-bar press (not just real keyboard focus - a tap focuses the button as a side effect), and the persistent Vale Cup score indicator showed world-wide instead of only near the Sowfield.

Screenshots (11, mixed landscape/portrait, 3 device profiles):

| | |
|---|---|
| charcreate swatches Charcreate, landscape | charcreate scrolled Charcreate scrolled to swatches/actions |
| in world landscape In world, 844x390 | chat centered landscape Chat composer centered |
| joystick clamp Joystick float clamp (far touchdown) | tooltip fix No tooltip popup on tap |
| portrait In world, portrait | chat centered portrait Chat composer centered, portrait |
| 926x428 926x428 landscape | 932x430 932x430 landscape |
| social panel Social panel, landscape | |

@FernandoX7 your review nit on `exitChatReply()` running on every `setActive(true)` was already addressed on this branch (commit 80b22742b, the `wasActive` guard) - confirmed it's in place and working as you suggested, thanks for the careful read.

FernandoX7 (commented):
Re-reviewed at e63a3213d (the four round 2 commits plus the branch as a whole). All three round 1 nits are properly closed, each with a decisive test: the composer draft survives a redundant re-activation and still resets on a genuine desktop-to-touch flip (both arms pinned), the media-block scan now brace-walks to the innermost enclosing block, and the window scrape tests id and class independently per tag. The party chip architecture is the textbook version of the repo recipe: a DOM-free state core in `party_collapse.ts` (UI_PURE_CORES), a one-time builder, write-elided painter wiring, the chip never built off mobile, and the chat yield never touching the persisted flag. I ran the ten touched suites at the head (232 pass) plus a clean tsc, and verified the i18n mechanics end to end (both keys in the catalog, five real non-Latin fills, one-line sha256, and the `data-i18n-title`/`data-i18n-aria` hydrator paths exist in `translatePage`).

Round 2 does introduce a handful of real issues, though. Should-fix, in rough order of user impact:

- `src/styles/hud.mobile.css:1646`: `#party-chip { grid-column: 1 / -1 }` cannot span the member columns. The container declares no `grid-template-columns` (columns are implicit from `grid-auto-flow: column`), so `-1` resolves to the explicit grid's only column line and the chip collapses to a single cell. With `.has-party-chip` raising the row template to 3 and `#party-leave` pinned at row 3 column 1, an expanded party with 2+ visible members auto-flows a member frame directly beside the chip (4px gap, both interactive) and sandwiches Leave mid-grid, instead of the "full stack under the chip" the PR describes. Only a 2-player party lays out as intended. Confirmed two ways (a placement-algorithm walk and a live repro of the exact rules). The overlap audit cannot see it: the chip vs `#party-frames` pair is exempted as nested and individual rows are never measured. Note a pure grid-line fix does not exist while the column count is implicit; the members likely need their own wrapper under the chip, or the chip needs to sit outside the grid.
- `src/styles/hud.mobile.css:1299`: the chat-contrast pin never wins for the one element it actually needs to protect. `body.mobile-touch.mobile-chrome-idle #mobile-combat-controls .mobile-btn` is (1,3,1) and the new `body.mobile-touch.mobile-chat-open #mobile-chat` is (1,2,1) in the same layer, and `mobile_chrome_fade.ts` has no chat guard, so the Chat button still dims to 0.5 while the player types or reads (OS keyboard taps are not page pointerdowns, so the idle class re-arms). The other three selectors in the pin cover elements the fade never dims. Match the fade's specificity (`#mobile-combat-controls #mobile-chat`) or add a `.mobile-chat-open` exclusion to the fade rule itself.
- `src/styles/hud.mobile.css:1411`: the dismiss chevron sits on the composer (same bottom anchor, left = composer right edge minus 40px, z-index 26 over 25, near-opaque glass), but `#chat-input` gets no compensating `padding-right`, so typed text runs under the button and caret taps in the last 40px land on the chevron. One `padding-right` line on the mobile composer rule fixes it.
- `scripts/mobile_hud_overlap_audit.mjs:1030`: the "close chat to leave a clean state for Pass B" cleanup calls `.click()` on `#mobile-chat`, which the file's own toggleMobileChat comment (and `bindChatButton`, `src/game/mobile_controls.ts:588`) says is a no-op: the button listens on pointerdown/pointerup only. There is a single `page.goto`, `hud.closeAll()` never touches the chat classes, and the round 2 setActive guard removed the accidental cleanup that used to clear them on the next viewport flip, so whenever the dismiss section runs, the entire Pass B matrix runs and screenshots under an open chat overlay. Reuse `toggleMobileChat` here, or the force-remove fallback used a few lines up.
- `scripts/mobile_hud_overlap_audit.mjs:565`: the simulated keyboard never reaches the CSS. `simulateKeyboardOpen` writes `--mobile-keyboard-visible-vh` on `documentElement`, but the real applier writes it inline on `body` (`src/game/keyboard_viewport_applier.ts:20`) and runs on every resize (`src/main.ts:483`), so body's full-height value shadows the simulated 180px for every chat element and the composer/chevron never take their keyboard-open seats. The stay-open, unfocused, and at-rest assertions are real; the keyboard-docking half is decorative. Write the var on body (or drive the real applier) and assert the docked seat actually moved.
- `scripts/mobile_hud_overlap_audit.mjs:849`: every chat-open failure degrades to a note (849, 943, 979) and there is no floor requiring the chat cycle to have run on at least one profile, so a regression that breaks the mobile chat tap path entirely still exits 0 with this PR's yield checks silently skipped. Count covered profiles and `fail()` when zero.
- `tests/party_collapse.test.ts:48`: the persistence tests exercise the key only through the imported `PARTY_COLLAPSE_STORE_KEY`, and no test pins the literal `'woc_party_collapsed'`, so a key rename keeps the suite green while silently dropping every player's stored choice. One literal pin closes it.

Nits:

- `tests/chat_keyboard_dismiss.test.ts:41`: `expect(keyboardDismissEffect()).toEqual(keyboardDismissEffect())` is a self-comparison any implementation passes; the literal pin at line 33 is the real guard. The line 33 comment also overstates: `main.ts` never reads `effect.closeChat` (main.ts:1059 reads only `blurComposer`), so that field cannot catch a wiring regression on its own.
- `tests/client_shell.test.ts:451`: the wiring scan pins the blur line but not `document.getElementById('chat-dismiss')`; with the optional chain, a lookup-id typo in main.ts keeps everything green while the chevron goes dead. Adding the lookup string to the scan closes it.
- `src/styles/hud.mobile.css:1400`: the chevron stays visible (titled "Hide keyboard") after it has dismissed the keyboard, a dead control until chat closes. Keying the reveal on `.mobile-keyboard-open` as well would show it only while there is a keyboard to hide.
- `scripts/mobile_hud_overlap_audit.mjs:478`: `chip.click()` exercises the real listener but bypasses hit-testing, so an overlay covering the chip would still pass. A coordinate tap at the chip centre would make it airtight.
- `tests/mobile_window_transform.test.ts:206`: `mediaBlocks` brace-scans without stripping comments (the coverage test next door strips them first); harmless against today's CSS, worth matching for consistency.
- `tests/mobile_window_coverage.test.ts:36`: the comment claims HTML comments are skipped, but a window tag inside `<!-- -->` is still scraped. The failure direction is loud (a commented-out window demands coverage, visibly), so only the comment needs correcting.

The merge conflict against the base is exactly the standard generated trio (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`): mechanical regen churn, not a design problem. Merge the base and re-run the i18n build.

FernandoX7:
Re-reviewed the branch end to end and took the remaining work over the line myself (the branch head is now d315430c0). What was done, in order:

**Round 3 verification.** All 13 findings from my previous review are properly closed, each verified against the code rather than the commit messages: the party chip seats the member rows in a `.party-rows` wrapper (desktop keeps `display: contents`, so the desktop stack is untouched; mobile owns the 2-row grid), the chat-contrast pin carries a second id so it now beats the idle fade, the composer reserves the chevron's 46px, the audit drives real pointer taps and force-removes the chat classes for Pass B, the simulated keyboard writes the viewport var where the real applier writes it and asserts the docked seat geometry, zero chat coverage now fails the audit run, and the storage key, lookup id, and self-comparison test gaps all got literal pins. The new features held up under a fresh pass too: the double-tap binder is clean pointer-event work with decisive tests, and the pet-bar/consumables/vale-cup band rules only re-seat chrome, never hide it.

**Merged jgyy's follow-ups (MasterZensei#4).** The phone-tested fixes are now on the branch. The composer centering is folded together with this branch's newer work: centered seat, chrome-scale width, the 46px chevron padding, the keyboard-docked bottom, and a landscape override so the chevron tracks the narrower 330px composer. The joystick float clamp is input-safe (the input origin stays the raw touch point; only the drawn wheel clamps to 48px of rest) and the new test pins exactly that contract, so the #1229 drift class stays dead. The charcreate, More-tray, and map-window overlap fixes all check out.

**Base merges.** release/v0.23.0 merged to the current tip (via the follow-ups branch plus a second merge). The five non-Latin overlay conflicts were unions of both sides' additions, and the generated i18n artifacts were resolved by regeneration, verified idempotent.

**Round 4 fixes (pushed with the merges):**

- The ring tap handlers no longer gate on the shared peek guard. With the ring's tooltip removed, a set peek flag there is always stale cross-talk from another control's long-press, and the old early return ate the player's next cast. They now clear the stale flag and dismiss a lingering tooltip box, then cast.
- The keyboard-open chat column reserves the composer's full 110px autosize cap (was the 84px reply min-height, so a long multi-line draft could grow the composer up into the log). The audit's typed message now saturates that cap, which also makes the geometry check font-independent.
- Literal pin for the 0.85 chrome-scale value: every consumer falls back to 1, so the lone declaration was silently deletable with the whole suite green.
- Literal pin for the landscape map-window rule (its sibling More-tray fix had one, this one did not).
- The audit's resting-pair counter counts measurements rather than passes, and touch_tap gets a boundary test at exactly DOUBLE_TAP_MS.

**Base breakage, not this PR.** The release/v0.23.0 tip itself is red (three test files, fallout from #1603's description tokenization: a stale en catalog row, stale Latin locale fills, a denylisted name in the faerie_fire prose, and over-broad macro pins in the coverage test). That is fixed separately in #1654 and merged; this branch carries the fixed tip, so CI here is green end to end.

**Verification at the final head:** full vitest green (908 of 909 test files, 11,151 tests, one file env-skipped as always), tsc clean, biome clean on the CI-parity changed set, i18n regen idempotent, malware scan clean, all three builds green, and the mobile overlap audit passes in gate mode across all profiles, including the new party-chip, chat-yield, and keyboard-dismiss passes. Audit screenshots at 844x390 from the final tree:

| | |
|---|---|
| keyboard-open chat column Keyboard open: centered composer docked above the keyboard line, chevron attached, log column clear | after dismiss After dismiss: chevron gone (nothing to hide), composer at rest, trio restored |
| party chip expanded Expanded 4-member party: double-stack below the chip, Leave below the grid | |

Two small observations for later, neither blocking: the audit's loot-settings reopen leaves `body.mobile-window-open` set, so one reachable meters-vs-autorun pair is skipped in the sweep (pre-existing audit semantics), and the tooltip focus-suppression also applies to desktop click-focus, which reads as intended (hover and Tab still show tooltips) but is worth knowing.

Thanks MasterZensei for the disciplined rounds, and jgyy for the real-device pass.

FernandoX7 (approved):

PR #1619, opened 2026-07-07
fix(deploy): control the two build-drift inputs behind the v0.22 CPU incident

Rubsey:
## Problem

During the v0.22 CPU P0 every in-repo suspect was ruled out (legacy dispatch verified live via the boot log, linkdead reverted, pg idle), yet rolling back to the v0.21 IMAGE fixed it. Investigation found the prod image bakes two inputs that are not pinned by this repo, so image rebuilds drift invisibly to git ranges:

1. Both Dockerfile stages use the floating node:22-alpine tag (runtime can change between builds).
2. The private bot detector is cloned at unpinned main at build time (pinned separately in woc-deploy).

It also found the compose game environment block silently drops host .env vars that are not allowlisted, which turned the first API_DISPATCH=legacy rollback test into a no-op and blocks PERF_TICK_LOG tick tracing.

## Changes

- Dockerfile: BASE_IMAGE build arg for both stages, defaulting to the current tag; deploys pin an exact version or digest.
- docker-compose.yml: pass the EASTBROOK_BASE_IMAGE build arg through, and allowlist API_DISPATCH and PERF_TICK_LOG (empty values keep server defaults, verified against server/http/config.ts parseDispatch and game.ts maybeLogTickPerf).

## Test plan

- [x] docker compose config validates; new keys interpolate ('' defaults)
- [x] parseDispatch treats '' as unset (DEFAULT_DISPATCH); PERF_TICK_LOG '' stays off
- [ ] Rebuild v0.22.1 image with EASTBROOK_BASE_IMAGE pinned to the v0.21 node version and redeploy (the confirming experiment for the remaining drift hypothesis)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,14 @@
 # World of Claudecraft game server — serves the built client, REST API and WebSocket
 # world on one port. Pair with a postgres service (see docker-compose.yml).
 
-FROM node:22-alpine AS build
+# Base image for BOTH stages, overridable at build time. An unpinned floating
+# tag means two builds of the same source can ship different node runtimes
+# (build drift, see the v0.22 CPU incident). Deploys should pin an exact
+# version or digest, for example:
+#   BASE_IMAGE=node:22.17.0-alpine
+#   BASE_IMAGE=node:22-alpine@sha256:<digest>
+ARG BASE_IMAGE=node:22-alpine
+FROM ${BASE_IMAGE} AS build
 WORKDIR /app
 COPY package.json package-lock.json ./
 RUN npm ci --no-audit --no-fund
@@ -23,7 +30,7 @@ ARG VITE_TURNSTILE_SITEKEY=""
 RUN VITE_TURNSTILE_SITEKEY="$VITE_TURNSTILE_SITEKEY" \
     npm run build && cp -a dist/media ./media-build && rm -rf dist/media && npm run build:server && npm run build:bot
 
-FROM node:22-alpine
+FROM ${BASE_IMAGE}
 WORKDIR /app
 ENV NODE_ENV=production
 COPY --from=build /app/dist ./dist
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,8 @@ services:
       args:
         # Build-time public client config inlined into the bundle (see .env.example).
         VITE_TURNSTILE_SITEKEY: ${VITE_TURNSTILE_SITEKEY:-}
+        # Pin the node base image for reproducible builds (see Dockerfile).
+        BASE_IMAGE: ${EASTBROOK_BASE_IMAGE:-node:22-alpine}
     container_name: eastbrook-game
     restart: unless-stopped
     depends_on:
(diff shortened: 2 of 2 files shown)


EnriqueGF:
Thanks for digging into this one, the incident write-up in the description is genuinely useful :)

**Blocking: base branch.** This PR targets `release/v0.22.1`, but the latest release branch is `release/v0.23.0` (`git ls-remote --heads origin "release/*"` confirms it, and `origin/release/v0.23.0` is at `12d645617`). Per the repo workflow every PR must be based on the newest `release/vX.Y.0` branch. Please reapoint and rebase onto `release/v0.23.0` before this can be considered. This alone blocks approval regardless of the code, which is otherwise close to landable.

Worth noting while rebasing: `release/v0.23.0` already added a `PERF_TICK_LOG` block to `.env.example` (lines 87-91), which does not exist on your v0.22.1 base. Expect that file to be part of the rebase context.

**What I verified**

- Built the branch in a clean worktree at `309eae96b` and ran `POSTGRES_PASSWORD=x docker compose config`. It validates, and the interpolated output is exactly what the description claims: `BASE_IMAGE: node:22-alpine`, `API_DISPATCH: ""`, `PERF_TICK_LOG: ""`.
- Confirmed the empty-string defaults are safe against the real code, not just the description: `server/http/config.ts:161` (`parseDispatch`) returns `DEFAULT_DISPATCH` for both `undefined` and `''`, and `server/game.ts:804` / `server/game.ts:2076` gate on `process.env.PERF_TICK_LOG === '1'`, so `''` stays off.
- Confirmed the global `ARG BASE_IMAGE` before the first `FROM` is correctly consumed by both `FROM` lines (`Dockerfile:11`, `Dockerfile:33`); no re-declaration is needed since it is only used in `FROM`.
- Confirmed `discord-bot` reuses `eastbrook-game:${EASTBROOK_IMAGE_TAG:-local}` rather than a second `build:` block, so there is no second place needing the arg.
- No `src/sim/` or `src/` code is touched, so determinism, `IWorld` parity and i18n are not in play. No secrets, no `ALLOW_DEV_COMMANDS`. No em/en dashes or emojis in the added lines.

**Medium: no test, and there is direct precedent for one.** `tests/deploy_discord_bot.test.ts` already reads `Dockerfile` and `docker-compose.yml` and pins the exact compose interpolation strings (`DISCORD_BOT_TOKEN: ${DISCORD_BOT_TOKEN:-}`). The bug this PR fixes is precisely "a var set in the host .env is silently dropped unless it appears in the environment block", which is exactly what such a test catches. Please add assertions in that same style: that `docker-compose.yml` contains `API_DISPATCH: ${API_DISPATCH:-}` and `PERF_TICK_LOG: ${PERF_TICK_LOG:-}` in the game environment block, and that `Dockerfile` declares `ARG BASE_IMAGE` and uses `FROM ${BASE_IMAGE}` for both stages. Without it, the next person who reformats the compose file reintroduces the exact failure mode the PR exists to prevent.

**Medium: the default is duplicated, and the compose default wins.** `Dockerfile:10` defaults `BASE_IMAGE=node:22-alpine` and `docker-compose.yml:31` independently defaults `${EASTBROOK_BASE_IMAGE:-node:22-alpine}`. Because compose always passes the build arg, bumping the Dockerfile default silently has no effect on any compose build. That is a new drift footgun introduced by a PR whose purpose is removing drift. Either drop the literal from compose and let the Dockerfile own the default, or add a comment at both sites stating that the compose value is authoritative and the two must be changed together.

**Medium: the new operator knob is undocumented.** `EASTBROOK_BASE_IMAGE` appears nowhere in `.env.example`, `DEPLOY.md` or `README.md`. The entire value of this change depends on deploys actually setting it, and the PR does not tell an operator that the knob exists or what to set it to. Please add it to `.env.example` next to the other `EASTBROOK_*` vars and mention the pin in the `DEPLOY.md` build section. The same applies to `API_DISPATCH`, which is described in `DEPLOY.md:177` but has no `.env.example` entry, so an operator following the rollback runbook has no local example to copy.

**Low: `API_DISPATCH` now fails the boot on a typo.** Before this change compose dropped the var, so a malformed host value was inert. After it, the value reaches `parseDispatch`, which throws on anything that is not `legacy` or `new` (`server/http/config.ts:163`). Failing fast is defensible and probably what you want, but it is a real behavior change for operators and is not called out in the description. Please note it in the PR body, and ideally in the `.env.example` comment.

**Low: the description oversells the fix.** The problem statement lists two unpinned inputs, but the diff only addresses the node base image. The private bot detector is still cloned at unpinned `main` (the Dockerfile just does `COPY private ./private` at `Dockerfile:25`, and pinning lives in `woc-deploy`). You do say this parenthetically, but the title claims to "control the two build-drift inputs" and only one is controlled here. Please retitle or scope the body so the merge commit does not read as if both were closed.

Also, test-plan item three (the confirming rebuild-and-redeploy experiment) is still unchecked. Since that experiment is the only thing that validates the hypothesis this PR is built on, please either run it before merge or state explicitly that this lands as a no-regression enabling change ahead of the experiment.

The direction is right and the shape is small and clean. Rebase onto `release/v0.23.0`, add the compose contract test, and resolve the duplicated default and I would be happy to approve.

PR #1621, opened 2026-07-07
Enforce SFX asset standards

### LINKED ISSUE #1574: Enforce audio asset standards: gen_sfx post-processing, conformance check, and docs
## Problem

The SFX library has no enforced asset standard. `scripts/gen_sfx.mjs` writes ElevenLabs output verbatim (128kbps 44.1kHz stereo MP3, hardcoded as `mp3_44100_128`), while recent hand-authored contributions arrived at 320kbps stereo. Nothing normalizes loudness, channel count, or bitrate, so the library is already inconsistent at 94 clips and will drift further as contributors add sounds.

The waste is real at runtime, not just on disk. `src/game/sfx.ts` decodes every clip into a float32 `AudioBuffer` that lives in memory for the whole session (roughly 350KB per stereo second). Positional playback (`playAt`) routes through an `equalpower` PannerNode, which downmixes stereo to mono before positioning, so stereo width in positional clips is decoded, held in RAM, and then discarded. On the iOS Capacitor build (WKWebView, deployment target 15.0) memory is the binding constraint.

The team agreed on a standard in the audio asset standards discussion, but a convention that lives only in a chat thread breaks silently the first time someone regenerates a clip. The standard needs to live in the generation pipeline and in a conformance check.

## Expected behavior

- `scripts/gen_sfx.mjs` post-processes every generated clip before writing it: peak-normalize to -6dBFS, downmix to mono for positional and one-shot categories (ambience loops stay stereo), and encode MP3 44.1kHz at the standard bitrate.
- The per-category mono/stereo policy is data-driven from the catalog in `scripts/sfx/sfx_prompts.mjs` (for example a `stereo: true` flag on ambience loop entries), not hardcoded key lists in the pipeline.
- A conformance check script scans `public/audio/sfx` and reports every non-conforming file with the reason: stereo where mono is required, peak above -6dBFS, bitrate above the ceiling (192kbps for hand-authored clips), wrong container, or a filename that does not match the key naming scheme.
- `docs/design/sound_effects.md` codifies the full standard: format (MP3 only, with the WebKit `decodeAudioData` rationale), loudness target, mono/stereo policy per playback path (`playAt` mono, `playUi` mono by default, ambience loops stereo), bitrate ceiling, and the naming convention (snake_case `category_subject_variant`, filename equals manifest key).
- The relevant CLAUDE.md files point at the standard so agents and contributors find it (`scripts/CLAUDE.md` script table, `src/game/CLAUDE.md` audio notes).

## Scope

- `scripts/gen_sfx.mjs`: add the post-process step (ffmpeg invoked as an external binary, same pattern as the browser scripts requiring a local Chrome; this is an offline maintainer-run script, not part of the build).
- `scripts/sfx/sfx_prompts.mjs`: per-entry channel policy flag for ambience loops.
- New check script under `scripts/` (for example `scripts/check_sfx.mjs`) runnable standalone; decide whether it also joins a gate tier.
- Re-process the existing library once through the new pipeline so shipped clips conform (including the 320kbps hand-authored clips).
- `docs/design/sound_effects.md`: write the standard (the `sfx.ts` header already references this doc as the audio doc of record).
- `scripts/CLAUDE.md` and `src/game/CLAUDE.md`: short pointers to the standard.

## Acceptance criteria

- Regenerating any clip via `gen_sfx.mjs` yields a mono, -6dBFS peak, 44.1kHz MP3 for positional and one-shot categories, and stereo for entries flagged as ambience loops.
- The check script exits non-zero and names the offending file and rule for each violation; a fully conforming `public/audio/sfx` passes clean.
- Every clip shipped in `public/audio/sfx` passes the check script after the one-time re-process.
- `docs/design/sound_effects.md` documents format, loudness, channel policy, bitrate ceiling, and naming, and is referenced from the CLAUDE.md files above.
- Runtime code (`src/game/sfx.ts`) is unchanged, or changed only if the re-processed assets expose a level-balance issue that a per-call gain tweak fixes.

## Notes

- OGG Vorbis is ruled out: the loader uses WebAudio `decodeAudioData`, which has no Ogg Vorbis support in WebKit on any iOS version (every iOS browser is WebKit), plus desktop Safari. MP3 stays the baseline. AAC in m4a is a possible future efficiency option and is out of scope here.
- Normalization is a consistency measure, not a clipping fix: the engine caps per-play gain at 1.0 under a 0.85 master, so clipping is not currently possible.
- The naming convention already exists de facto in `scripts/sfx/sfx_prompts.mjs` (`mob_<family>_aggro|attack|death`, `foot_<surface>`, `amb_*`, `cast_/impact_/proj_<school>`, `melee_*`, `combat_*`, `move_*`, `player_*`); this issue documents and checks it rather than inventing a new one.
- `gen_sfx.mjs` skips existing files unless `--force`, which is how hand-authored clips coexist with generated ones; the post-process step must also run on that hand-authored path (or the check script must cover it) so contributions conform without regeneration.

loopermallee:
## Summary
- Adds shared SFX asset standard constants and FFmpeg helpers for probing, peak measurement, and post-processing.
- Post-processes generated SFX before the final MP3 write and adds `--conform-existing` for existing clips.
- Makes ambience loops stereo via catalog `stereo: true`, while positional, UI, and one-shot clips remain mono by default.
- Adds `npm run sfx:check` to report format, sample-rate, channel, peak, bitrate, and filename/catalog violations.
- Documents the MP3-only asset policy, WebKit/decodeAudioData rationale, loudness target, channel policy, bitrate ceiling, naming, generation, and conformance, with CLAUDE pointers for scripts and runtime audio agents.

Fixes #1574

## Validation
- `git diff --check upstream/main...HEAD`: passed.
- `node --check scripts/check_sfx.mjs`: passed.
- `node --check scripts/gen_sfx.mjs`: passed.
- `node --check scripts/sfx/sfx_prompts.mjs`: passed.
- `node --check scripts/sfx/sfx_ffmpeg.mjs`: passed.
- `node --check scripts/sfx/sfx_asset_standard.mjs`: passed.
- Catalog validation import: passed, 93 entries, 10 stereo ambience entries.
- `ffmpeg` / `ffprobe`: not found on PATH here, so the synthetic FFmpeg fixture and real asset conformance/check were not run in this environment.

## Asset reprocessing
Real `public/audio/sfx` MP3 assets were present and writable (93 MP3s), but they were not reprocessed or committed because `ffmpeg` and `ffprobe` are not installed on this machine's PATH.

Maintainer command once FFmpeg is available:

`node scripts/gen_sfx.mjs --conform-existing && npm run sfx:check`

Runtime `src/game/sfx.ts` is unchanged intentionally. Malformed clips should be fixed in the asset pipeline instead of compensated at runtime.


loopermallee:
**Explanation**

This patch makes the game’s sound effects follow one clear “house standard” before they are allowed into the game.

Right now, generated audio can come out with different technical properties: some files might be too loud, too high quality in a wasteful way, stereo when they should be mono, the wrong sample rate, or just not shaped in the safest way for browser playback. That can cause inconsistent volume, bigger files than needed, or browser decoding issues, especially in WebKit/Safari-style audio handling.

This patch adds a proper audio cleanup and inspection step.

**What Changes**

When `scripts/gen_sfx.mjs` generates a new sound effect, it now runs the file through FFmpeg before saving the final MP3. FFmpeg acts like a finishing station: it forces the sound into the expected format, bitrate, sample rate, volume ceiling, and channel layout.

It also adds:

```sh
node scripts/gen_sfx.mjs --conform-existing
```

That command lets maintainers reprocess the existing sound files using the same rules, without calling ElevenLabs again. So it can clean up old MP3s without spending new generation credits.

**Mono vs Stereo**

The patch makes the audio catalog decide whether a sound should be mono or stereo.

Most short, positional, UI, and one-shot sounds stay mono. That makes sense because they are usually played from a point in the world or as a simple effect.

Ambience loops are marked as stereo with `stereo: true`, because background ambience benefits from width and space.

**New Checker**

The new `scripts/check_sfx.mjs` script is basically a quality-control tool for sound files.

It checks every generated SFX file and reports problems like:

- not MP3
- wrong sample rate
- too high bitrate
- too loud
- wrong mono/stereo channel count
- bad filename or catalog key mismatch

The useful part is that it reports every offending file and rule, instead of stopping at the first problem. That makes cleanup much easier.

You can run it with:

```sh
npm run sfx:check
```

**Shared Standard**

The patch adds `scripts/sfx/sfx_asset_standard.mjs`, which keeps the rules in one place:

- MP3 only
- 44.1kHz sample rate
- generated MP3 target bitrate of 128kbps
- maximum accepted bitrate of 192kbps
- peak volume ceiling of `-6dBFS`
- mono/stereo rules
- filename and catalog key naming rules

That matters because the generator and checker now use the same source of truth. Fewer hidden mismatches.

**Docs and Agent Notes**

The docs now explain the audio policy in plain project terms: why MP3 is required, why WebKit/browser decoding matters, what loudness target to use, when to use mono or stereo, how naming works, and how to run the generation/checking commands.

The CLAUDE notes also point future contributors and coding agents toward the standard, so malformed audio should be fixed in the asset pipeline instead of patched around inside runtime game code.

**What It Does Not Do**

It intentionally does not change `src/game/sfx.ts`.

The runtime should not have to compensate for bad files. The files themselves should be corrected before they reach the game.

**Why This Is Useful**

In simple terms: this makes sound effects more predictable, safer for browser playback, easier to review, and easier to maintain.

It turns “hope every MP3 is okay” into “generate, normalize, check, and reject anything that does not meet the standard.”

Thanks to @FernandoX7: [levy-street/world-of-claudecraft#1574](https://github.com/levy-street/world-of-claudecraft/issues/1574) for raising the issue with great detail.

loopermallee:
Rebase/retarget update:

- Rebased `fix/sfx-asset-standards-1574` onto `upstream/release/v0.23.0` and force-with-lease pushed `fd23b73e7`.
- i18n artifacts were confirmed unchanged for this non-localization/tooling PR; unrelated generated status drift from the release scanner was left out of scope.
- Resolved the release conflict in `scripts/gen_sfx.mjs` by preserving the PR FFmpeg/conform flow and the release branch custom-recording skip.
- Focused validation passed:
  - `node --check scripts/check_sfx.mjs`
  - `node --check scripts/gen_sfx.mjs`
  - `node --check scripts/sfx/sfx_prompts.mjs`
  - `node --check scripts/sfx/sfx_ffmpeg.mjs`
  - `node --check scripts/sfx/sfx_asset_standard.mjs`
  - SFX catalog import sanity check (`entries=96 stereo=10`)
  - `git diff --check`
  - `npx tsc --noEmit`
  - `npm run security:gate`
  - `npm run build`
- `npm run sfx:check` was run but could not complete in this local environment because `ffprobe`/`ffmpeg` are not on PATH; it also reports the existing tracked `public/audio/sfx/cast_lightning_bolt.wav` asset. I left that asset untouched as outside this rebase/retarget scope.

jgyy:
Closing as superseded: SFX asset conformance is now a first-class gate (scripts/sfx/conform_audio.mjs + sfx_conform_inventory.mjs, run as part of npm run gate). Thanks for the contribution.

PR #1622, opened 2026-07-07
perf(pet): scan the spatial grid in petPickTarget instead of the full roster

Rubsey:
## Summary

`petPickTarget` (`src/sim/pet/pet_ai.ts`) selects a hunter/warlock pet's target. A
target-less non-passive pet ran an `O(all-entities)` linear scan of `ctx.entities`
**every idle tick (20 Hz)** with no cadence throttle. In the 70-player production CPU
investigation this was a top self-time frame (~5.4%, and higher inclusive once
`isHostileTo`/`dist2d` are counted). The default `defensive` mode runs the full scan
every idle tick, so it dominates in normal play.

This replaces **only the iteration source** with the spatial grid, the same
`SimContext` seam (`ctx.grid.forEachInRadius`) the sibling `pullNearbyMobs` already
uses. The loop body is verbatim: the `m.id === pet.id` skip, the
`engagingUs`/`ownerOffense`/`aggressive` admit predicates, and the strict `d < bestD`
nearest selection are unchanged.

```
- for (const m of ctx.entities.values()) { ... continue ... }
+ ctx.grid.forEachInRadius(pet.pos.x, pet.pos.z, PET_ASSIST_RANGE, (m) => { ... return ... });
```

Net effect: `O(entities-within-50yd)` per idle tick instead of `O(all-entities)`,
scaling with local density rather than total world population.

## Why it is behavior-preserving

- **Radius 50 is a safe superset in both modes.** `bestD` starts at `PET_ASSIST_RANGE`
  (50) or `PET_AGGRESSIVE_RANGE` (18) and only decreases; selection is strict `<`, so
  no winner can lie at or beyond 50 yd. Aggressive-mode 18..50 candidates the wider
  query surfaces are re-rejected by the `aggressive` `d <= 18` predicate. The grid's
  `d2 <= r2` boundary is inclusive at exactly 50, but the strict inner `d < bestD`
  still rejects it.
- **The grid holds every entity kind** (`entity_roster.ts` inserts all kinds on spawn),
  so PvP hostile players remain candidates.
- **Single-arg callback** (`(m) =>`): the callback's squared `d2` is deliberately not
  reused; the inner `dist2d` is kept, so there is no squared-vs-linear units mismatch.

## Determinism / parity

`petPickTarget` draws no rng, and this change adds none. The only observable difference
is grid iteration order vs `Map` insertion order, which matters **only on an exact
float-distance tie** (strict `<` keeps first-seen). `npx vitest run tests/parity` is
**GREEN with no golden re-mint** (162 passed): no committed scenario hits a tie, so no
rng draw-order shifts and seed-stable RL/replay reproducibility is preserved across all
three hosts (offline Sim, server, headless RL env).

## Tests

New pins in `tests/pet_ai.test.ts`: nearest-selection (grid path == old full scan),
deterministic tie resolution, strict-`<` boundary at `PET_ASSIST_RANGE` (with a 49 yd
control), aggressive-range re-rejection (with a 15 yd control), PvP hostile-player
selection, query-center on `pet.pos`, dead-corpse skip, and the `ownerOffense` threat
disjunct. The existing `petPickTarget` unit tests now sync the grid before a pick (it is
grid-backed now), mirroring the end-of-tick `grid.refresh`.

## Reviews

Clean across `architecture-reviewer` (SimContext seam, no new rng, intentional and
bounded draw-order), `test-coverage-auditor` (all pins decisive; the two flagged
coverage gaps are closed), and `cross-platform-sync` (no `IWorld`/wire/`SimEvent`
change; grid is roster-exact and host-identical).

## Second commit: i18n orphan reconciliation (unblocks the PR gate)

The `release/v0.22.1` base carried a pre-existing i18n drift that failed the PR gate
independently of the perf change: the linkdead-grace **revert** (`92a16b284`) removed
the English `reconnecting` key (the `loading.reconnecting` leaf) and cleaned it from the
ja/ko/ru/zh_CN/zh_TW overlays, but left it **orphaned in 12 other overlays** a later
release fill had populated. An overlay key absent from `Leaves(en)` failed
`i18n_overlay_key_membership`, `i18n_build_gapfill` (5094 vs 5093 keys), and
`i18n_resolved_equivalence` (stale resolved table + SHA), plus 26 `tsc` errors in the
generated tables.

`fix(i18n): drop orphaned loading.reconnecting from 12 locale overlays` removes the
orphan line from the 12 overlays (matching what the revert did to the other 5),
regenerates the resolved tables (`i18n:gen`), and rebaselines
`src/ui/i18n.resolved.sha256`. No English or player-facing behavior change (the feature
the string belonged to is already reverted). Kept as a separate commit for reviewability;
happy to split it into its own PR if preferred.

## Verification (whole branch now green locally)

- [x] `npx vitest run` full suite: **829 files passed, 0 failed** (9805 tests passed)
- [x] `npx vitest run tests/parity` (162 passed, no golden re-mint)
- [x] `tsc --noEmit`: **0 errors** (the 26 pre-existing i18n errors are cleared)
- [x] `biome ci` clean on all changed files (pet + the 12 overlays)
- [ ] Authoritative measurement: CPU profile under 70-player/bot-raid load with
  hunter+warlock pets, confirming `petPickTarget` ~5.4% before -> near-zero after


FernandoX7 (commented):
This is a clean, careful perf fix. Swapping only the iteration source (`ctx.entities.values()` to `ctx.grid.forEachInRadius`) while keeping the loop body verbatim is the right shape, and the PR does the work to prove it is behavior-preserving rather than asserting it. I verified the load-bearing claims against the code at head, with a focus on whether pet targeting can regress:

- **The candidate set is identical to the old full scan.** `forEachInRadius` filters `d2 <= r2` (inclusive at exactly 50), but selection is strict `d < bestD` with `bestD` starting at 50 (defensive) or 18 (aggressive) and only decreasing, so radius 50 surfaces every candidate the old scan could have picked in both modes and the aggressive 18..50 extras are re-rejected by the `d <= 18` predicate. (`spatial.ts` boundary, `pet_ai.ts:249-276`)
- **The grid holds every kind.** `addEntityToRoster` inserts unconditionally into `ctx.grid` (`entity_roster.ts:76`); the `player` gate is only for the separate `playerGrid`, so PvP players stay candidates. The duel test pins it.
- **The freshness contract is the same one the sibling already relies on.** `petPickTarget` has a single caller (`updatePet`, `pet_ai.ts:89`), which runs in the mob/locomotion pass before the end-of-tick `grid.refresh` (`sim.ts:3192`), exactly like `pullNearbyMobs` in the same function. Cell membership can be one tick stale, but the distance check uses live positions and the cell window is padded, and every large instant move (respawn, spirit, arena, delves, dungeons, boss mechanics, pet teleport, mob lifecycle) rebuckets immediately while all combat/locomotion movement is gradual. So no hostile can slip into range unseen for a tick that the old live-position scan would have caught.
- **Keeping `dist2d` over the callback's `d2` is the correct call**, not a missed optimization: the admit predicates are in linear yards, so reusing squared `d2` would have needed `bestD * bestD` and is exactly the units mismatch worth avoiding.

Determinism: the one observable change is grid iteration order vs Map insertion order on an exact float-distance tie, and the PR is upfront about it. On my end `tests/parity` is green (162 passed) with no golden re-mint on a clean tree, `tests/architecture.test.ts` passes (24), and all 12 pet suites pass (75). The parity `pet_ai` scenario drives the melee pet's `petPickTarget` acquisition through real full ticks (grid refreshed by the loop) and stays byte-identical, which is the end-to-end proof. The new pins add decisive coverage: all three admit paths, the strict boundary with a control, aggressive-range re-rejection with a control, the pet-centered query, the corpse skip, and the deterministic tie with its stability across calls.

No blocking or should-fix findings. Nice work.

FernandoX7 (approved):

PR #1623, opened 2026-07-07
perf(sim): skip rim-crest noise where the rim factor is zero (#1620)

### LINKED ISSUE #1620: terrainHeight pays rim-crest noise and terracing on every ground sample even where rim is 0
## Problem

Since v0.22.0, every `terrainHeight` call unconditionally evaluates the rim-crest noise and terracing added by the jagged-mountains work (af3152860, 5dc9a8630), even where the rim factor is 0, which is the entire open world. Each call pays two extra fbm2 evaluations (a 3-octave plus a 2-octave, 5 noise2 = 20 hash2 evaluations), plus `terraceStep` and an outside-fade smoothstep, for a contribution that is multiplied by `rim` and therefore exactly zero.

`groundHeight` is the hottest sim function on the server: player `stepPlayer` deep-water and gravity checks, the mob `moveToward` slide fan (1 to 8 calls per moving mob per tick), pets, charge and fishing checks. Measured impact is small (order 0.1 percent of a core at typical scale, confirmed during the v0.22 CPU investigation, and explicitly not the cause of that incident), but it is free waste on the hottest path in the sim.

## Expected behavior

- `terrainHeight` skips the rim-crest and terrace evaluation when `rim === 0`.
- Returned heights are bit-identical to the current implementation for every input (the skipped contribution is multiplied by `rim`, so this must be a pure early-out, never an approximation), preserving determinism across the three hosts.

## Scope

- `src/sim/world.ts` around line 486 (the `rimCrest` fbm2 pair, `terraceStep`, and the outside-fade smoothstep).
- Verify the second in-range addition (the inter-zone ridge fine layer) for the same wasted-work pattern while there.

## Acceptance criteria

- A regression test pins `terrainHeight` values at sample points with `rim === 0` and `rim > 0`, asserting identical outputs before and after the change.
- The rim-crest and terrace work does not execute when `rim === 0` (verifiable by extraction or a counter in the test).
- `tests/architecture.test.ts` and the sim suite stay green.

## Notes

Found during the v0.22 production CPU investigation. Not the P0 cause; filed as hot-path hygiene.

daxdax89:
## What

`terrainHeight`/`groundHeight` is the hottest function in the sim: player and mob movement, gravity, deep-water and charge/fishing checks all sample it many times per tick. Since the jagged-mountains work, every call unconditionally evaluated the rim-crest noise, two `fbm2` layers (a 3-octave plus a 2-octave, five `noise2` = 20 `hash2` evaluations), even where the rim factor is 0.

That contribution is only ever consumed as `rim * 55 * rimCrest`, and `rim` is exactly 0 across the entire open world (nonzero only within 30yd of the boundary). So the two `fbm2` evaluations were pure waste on the hottest path everywhere a player actually is.

## Fix

Guard the rim-crest block behind `if (rim > 0)`. This is a **pure early-out, never an approximation**: any positive rim still pays the full crest, so returned heights are **bit-identical for every input**. Determinism across the three hosts (offline `Sim`, server, headless RL env) and agreement with the renderer (which samples the same function) are preserved. The inter-zone ridge crest is already gated by its own proximity check, and `terraceStep(0) === 0`, so the rim block was the sole remaining per-sample waste.

## Tests

`tests/terrain_walls.test.ts`:
- Pins bit-identical `groundHeight` outputs at `rim === 0` interior samples and `rim > 0` boundary samples (+X edge, south, north): the early-out must not move any value.
- Asserts the two rim-crest `fbm2` evaluations run **only** where `rim > 0`: a same-`z` interior vs edge sample whose `fbm2` call counts differ by exactly two (test-first: failed before the guard).

The parity golden-trace gate (`tests/parity`) and `tests/architecture.test.ts` (sim purity) stay green.

## Notes

Found during the v0.22 production CPU investigation; filed as hot-path hygiene (not the P0 cause). Measured impact is small (order 0.1% of a core at typical scale) but it is free to reclaim on the sim's busiest path.

Fixes #1620

### DIFF
--- a/src/sim/world.ts
+++ b/src/sim/world.ts
@@ -516,11 +516,20 @@ export function terrainHeight(x: number, z: number, seed: number): number {
   // crest as the inter-zone ridges: a coarse peak/saddle layer plus a finer
   // crag layer, same conservative combined variance so the climb-limit
   // invariant still holds along the whole rim.
-  const rimCrest =
-    1 +
-    (fbm2(x * 0.025, z * 0.025, seed + 29, 3) - 0.5) * 0.35 * mountainDetail +
-    (fbm2(x * 0.09, z * 0.09, seed + 37, 2) - 0.5) * 0.15 * mountainDetail;
-  mountainAdd += rim * 55 * rimCrest;
+  // rimCrest is only ever consumed as `rim * 55 * rimCrest`, so where rim is 0
+  // (the entire open world, everything more than 30yd inside the boundary) it
+  // adds exactly 0. Skip its two fbm2 layers there: groundHeight is the hottest
+  // sim function (player/mob/pet movement, gravity, deep-water and charge checks)
+  // and this was paying 5 noise2 evaluations per ground sample for nothing. This
+  // is a pure early-out, never an approximation: any positive rim still pays the
+  // full crest, so returned heights stay bit-identical (tests/terrain_walls.test.ts).
+  if (rim > 0) {
+    const rimCrest =
+      1 +
+      (fbm2(x * 0.025, z * 0.025, seed + 29, 3) - 0.5) * 0.35 * mountainDetail +
+      (fbm2(x * 0.09, z * 0.09, seed + 37, 2) - 0.5) * 0.15 * mountainDetail;
+    mountainAdd += rim * 55 * rimCrest;
+  }
   // Terrace the combined mountain rise into stair-stepped bands (flat treads
   // + steep risers) instead of one smooth ramp, so slopes read as a stacked
   // rocky mountainside rather than a uniform incline. This does not reduce
--- a/tests/terrain_walls.test.ts
+++ b/tests/terrain_walls.test.ts
@@ -1,6 +1,7 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 import { CAMPS, NPCS, ROADS, WORLD_MAX_X, WORLD_MAX_Z, WORLD_MIN_Z, ZONES } from '../src/sim/data';
 import { PLAYER_MAX_CLIMB_SLOPE } from '../src/sim/pathfind';
+import * as rng from '../src/sim/rng';
 import {
   PITCH,
   PITCH_CENTER,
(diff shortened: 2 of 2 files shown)


EnriqueGF:
**Reviewed: production-ready. Approving.**

This is a clean, correctly-scoped hot-path micro-optimization. Verified end to end:

**Correctness / bit-identity**
- `rimCrest` is only ever consumed as `mountainAdd += rim * 55 * rimCrest` (src/sim/world.ts:520). When `rim === 0` the term is exactly 0 regardless of the crest value, so gating its computation behind `if (rim > 0)` cannot change any returned height. When `rim > 0` the full crest is still paid unchanged. Genuinely a pure early-out, not an approximation.
- `rimCrest` depends only on `x, z, seed, mountainDetail`, never on `rim`, so no cross-dependency is broken by moving it inside the branch. It is not referenced anywhere else.

**Determinism / tri-host parity**
- `fbm2`/`noise2`/`hash2` (src/sim/rng.ts) are pure coordinate-hash functions with no shared mutable PRNG state, so skipping evaluations does not perturb any RNG stream or draw order. Determinism across offline Sim, server, and headless env, plus agreement with the renderer sampling the same function, is preserved.
- Sim purity is untouched (no DOM/Three imports added).

**Tests**
- `tests/terrain_walls.test.ts` pins bit-identical `groundHeight` outputs at both `rim === 0` interior samples and `rim > 0` boundary samples, and asserts the two rim-crest `fbm2` calls fire only where `rim > 0` (interior vs same-`z` edge, delta of exactly 2). Ran locally: 7/7 pass in 1.6s.

**Conventions**
- Scoped diff, no unrelated refactors, no em/en dashes or emojis, comment explains the invariant it protects.

Nothing blocking. Good hygiene reclaim on the sim's busiest path.

daxdax89:
Thanks for the careful review, appreciate you verifying the bit-identity and draw-order end to end. Just a heads up: I rebased this onto the current release tip (591dd8f14) after today's terrain/voxel merges, and re-ran tests/terrain_walls.test.ts against the new world.ts, the pins and the fbm2 call-count skip-proof still pass (the terrain work didn't touch the rim-crest math). So it's current and green whenever you want to land it.

EnriqueGF:
:) Thanks for the contribution, and thanks for the follow-up note.

I re-verified this from scratch on a fresh worktree rather than re-reading the earlier approval. New evidence beyond what was already covered:

**Bit-identity, verified empirically.** I dumped `groundHeight` at all ten pinned coordinates on the PR head and on its pre-change parent (`efeca72c3^`) with seed 20061 and diffed them: byte-for-byte identical. The pins in `tests/terrain_walls.test.ts` are genuine pre-change values, not self-comparisons. Guard equivalence is airtight: `smoothstep` (src/sim/world.ts:189) clamps to `[0,1]`, so `rim = Math.max(rimX, rimS, rimN)` is never negative and `rim > 0` is exactly `rim !== 0`. `rimCrest` is block-local with no other consumer in the repo.

**The spy test is decisive.** Removing the `if (rim > 0)` guard makes it fail (interior 4, boundary 4, delta 0); with the guard it is interior 2, boundary 4. So `vi.spyOn` genuinely intercepts `world.ts`'s internal `fbm2` calls rather than silently counting zero on both sides.

**Gates run locally on the branch merged with the current release tip:** `tsc --noEmit` clean, `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` green, `tests/parity` green, `tests/terrain_walls.test.ts` 10/10 (including the pre-existing climb-limit wall invariants), plus `voxel`, `pathfind`, `climb_slope`, `fixes`. Biome clean on both changed files. Base branch is `release/v0.23.0`, the latest release. Change is non-visual, so no screenshots are required.

**Two notes, neither blocking:**

1. Your head commit is actually parented on `bcf975752`, not `591dd8f14`, so the rebase did not land as described. It does not matter in practice: I ran a trial merge with the release tip and it is conflict-free, and the release-side additions to `world.ts` (`terrainWallStandoff`, the scatter slope gate) leave `terrainHeight`'s math untouched, so the pins survive the merge. Worth noting that `terrainWallStandoff` calls `groundHeight` nine times per grounded player per tick (src/sim/player_motion.ts:322), which makes this early-out worth strictly more than when you measured it.
2. Optional test hardening: `expect(boundary - interior).toBe(2)` pins a difference, not the absolute counts, so it would also pass at 10 versus 12. Adding `expect(interior).toBe(2)` and `expect(boundary).toBe(4)` would lock the rim-crest pair as the entire difference. The bit-identity pins already cross-check the output, so this is a nit.

No CI checks have run on this head yet, so please make sure the pipeline goes green before landing. Approving on the code.

daxdax89:
Friendly nudge: you reviewed this as production-ready back on v0.24. It has since been retargeted and rebased cleanly onto release/v0.26.0 (the perf early-out and its bit-identical pins are unchanged, tests green). Good to merge when you have a moment?

daxdax89:
Closing as superseded: the #1839 release batch already landed a fix for #1620 (the terrainHeight rim-crest early-out) on release/v0.26.0, so this PR is now redundant. Thanks for building it out over there. Glad the issue is resolved either way.

PR #1624, opened 2026-07-07
fix(ui): bind death-screen respawn buttons via touch-tap (#1484)

### LINKED ISSUE #1484: Stuck on death sreen
### Before you start

### What happened?

When you die, you get stuck on this screen 

### Steps to reproduce

1. Die
2. get stuck on death screen

Respawn after click on button

### Where did this happen?

Online (multiplayer server)

### Device

Mobile (phone / tablet)

### Browser / OS

_No response_

Image

daxdax89:
## Problem

On mobile, a player can get stuck on the death screen: tapping **Release Spirit** does nothing and the overlay never clears (issue #1484).

## Root cause

The three death-screen buttons (Release Spirit, Resurrect at Corpse, Resurrect at Spirit Healer) were wired with a bare `addEventListener('click', ...)`. As `src/ui/touch_tap.ts` documents, the browser only synthesizes a compatibility `click` for the **primary** pointer, so a bare `click` binding goes dead the moment a second finger is already down.

The concrete case: the player dies mid-run with the movement joystick still held. That joystick finger owns the primary pointer (implicit pointer capture), and `#mobile-controls` (z-index 60) sits above the death overlay (z-index 30) and is never hidden on death. A tap on Release Spirit with the other thumb never synthesizes a `click`, so `sim.releaseSpirit()` never fires, no release command reaches the server, the ghost flag never flips, and the overlay (keyed off `dead && !ghost`) never clears. This is why it is mobile-only and intermittent (it needs a second finger down).

Every other touch-facing HUD control (the mobile action ring, the more-menu close X, page toggles) was already migrated to `bindTouchTap` for exactly this reason; these three were missed.

## Fix

Swap the three bindings to `bindTouchTap`, matching every sibling button. `bindTouchTap` preserves the desktop mouse and keyboard (Enter/Space) `click` path, so this is purely additive touch handling with no desktop regression. No server, wire, or sim change: the online release path was already correct, the tap just never reached it.

## Tests

`tests/client_shell.test.ts` asserts all three buttons use `bindTouchTap` and none is left on a bare `click` binding, extending the same source-scan guard that already covers the more-menu close X.

## Note on screenshots

There is no visual change: the death overlay renders identically. The defect is a dead tap under multi-touch, so the fix is behavioral (which event binding is used), not visual.

Fixes #1484

### DIFF
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -1354,12 +1354,17 @@ export class Hud {
     });
     this.initCompass();
     this.initMinimapZoom(mm);
-    this.releaseSpiritBtnEl.addEventListener('click', () => {
+    // bindTouchTap, not 'click': the browser only synthesizes click for the
+    // PRIMARY pointer, so on a phone these death-screen buttons went dead the
+    // moment another finger was down (a held movement joystick when the player
+    // died mid-run), stranding them on the death overlay (issue 1484). Matches
+    // every other touch-facing HUD button; desktop mouse/keyboard is preserved.
+    bindTouchTap(this.releaseSpiritBtnEl, () => {
       if (this.sim.arenaInfo?.match) return;
       this.sim.releaseSpirit();
     });
-    this.resurrectCorpseBtnEl.addEventListener('click', () => this.sim.resurrectAtCorpse());
-    this.resurrectHealerBtnEl.addEventListener('click', () => this.sim.resurrectAtSpiritHealer());
+    bindTouchTap(this.resurrectCorpseBtnEl, () => this.sim.resurrectAtCorpse());
+    bindTouchTap(this.resurrectHealerBtnEl, () => this.sim.resurrectAtSpiritHealer());
     document.addEventListener('pointerdown', (ev) => {
       const target = ev.target as Node | null;
       if (!target) return;
--- a/tests/client_shell.test.ts
+++ b/tests/client_shell.test.ts
@@ -942,6 +942,23 @@ describe('client HTML shell', () => {
     expect(hudTs).toContain('bindTouchTap(moreClose, () => {');
   });
 
+  it('binds the death-screen respawn buttons via touch-tap, not bare click', () => {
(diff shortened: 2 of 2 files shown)


EnriqueGF:
:) Thanks for the clean, well-diagnosed fix.

I reviewed this against the real code (merged onto the current `release/v0.23.0` tip) and it is production-ready. Approving.

What I verified:
- **Base branch:** targets `release/v0.23.0`, the latest release branch. Gate passes.
- **Root cause is real:** `release/v0.23.0` still binds all three death-screen buttons with a bare `addEventListener('click', ...)` (hud.ts around 1389-1394), while every sibling touch-facing control (action ring, more-menu close X, page toggles) already uses `bindTouchTap`. The multi-touch dead-tap explanation matches `src/ui/touch_tap.ts` (compatibility `click` only for the primary pointer), and the z-index 60 claim for `#mobile-controls` checks out in `hud.mobile.css`.
- **No regression:** `bindTouchTap` keeps the mouse/keyboard `click` path with its suppression window, so desktop and Enter/Space activation are preserved. The `arenaInfo?.match` guard on Release Spirit is carried over unchanged.
- **Scope:** UI-only, no sim/server/wire/i18n change, so tri-host parity, server authority, and determinism are untouched. No visual change, so the screenshot waiver is correct. No monolith growth (like-for-like binding swap).
- **Tests:** the new `client_shell.test.ts` case is decisive: it asserts each of the three exact `bindTouchTap(...)` bindings plus a negative `.not.toMatch` that fails if any button regresses to a bare `click`. `tsc --noEmit`, the full `client_shell.test.ts`, and `architecture.test.ts` all pass on the merged tree.

Minor, non-blocking: the PR branch is well behind base (merges cleanly since the target lines are untouched on release), so no rebase is required, but squash-merging is fine.

Nice work.

daxdax89:
Friendly nudge: you reviewed this against the real code on v0.24. It is now rebased cleanly onto release/v0.26.0 with the touch-tap fix and its source-scan test intact. Good to merge when convenient?

PR #1625, opened 2026-07-07
Improve cast bar feedback and pet cooldown clarity

loopermallee:
Summary:
- Added semantic cast-bar visual states and completion/interruption feedback.
- Improved pet action/cooldown labels so cooldowns are no longer number-only.
- Added/updated accessible labels, non-color cues, and reduced-motion/forced-colors handling.

Testing:
- npm run check failed on existing broad Biome baseline issues before this patch's files were reached: 785 errors, 4690 warnings, 456 infos. First diagnostics were in headless/env_server.ts, vite.config.ts, tests/zones_command.test.ts, and iOS asset JSON formatting.
- npm run check:types passed.
- npm test failed in the full repo suite: 39 failed files, 22 failed tests, 26 admin/Svelte suites could not load /@fs/Users/asrulsashedesilva/Documents/WorldOfClaudecraft/node_modules/@testing-library/svelte/src/vitest.js, and several unrelated long-running gameplay/security/mail tests timed out. The i18n hash mismatch from this branch was fixed by updating src/ui/i18n.resolved.sha256.
- npm run build passed.
- npx vitest run tests/cast_bar.test.ts tests/cast_bar_painter.test.ts tests/pet_action_feedback.test.ts passed: 3 files, 34 tests.
- node scripts/i18n_resolved_hash.mjs --check passed after regenerating the hash.
- npx vitest run tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts passed: 2 files, 22 tests, 1 skipped.
- Manual browser smoke: ran npm run dev -- --host 127.0.0.1 --port 5174 and checked /play.html with Playwright. Verified cast progressbar roles, success/interrupted visual states, target/pet semantic classes, pet Taunt plus 7s plus AUTO labels, cooldown overlay CSS, and reduced-motion disabling animations. Vite logged 502 proxy errors for /api/site-presence and /api/project-stats because no local backend was running; the HUD smoke still completed.

Notes:
- No gameplay/combat balance changes.
- UI-only or presentation/event-metadata-only changes.
- Designed to preserve existing HUD performance patterns.


loopermallee:
**What Changed**

The game’s cast bars used to mostly look like the same yellow bar, even when different things were happening. Now the cast bar can visually and textually communicate what is going on.

It now distinguishes:

- Normal casts
- Channeling casts
- Eating/drinking/consume bars
- Successful cast completion
- Interrupted or failed casts
- Interruptible enemy casts
- Uninterruptible enemy casts
- Important/dangerous casts
- Pet-related casts/actions

In layman terms: instead of every cast looking like “yellow bar doing something,” the UI now says more clearly, “this is being cast,” “this is channeling,” “this cannot be interrupted,” “this was interrupted,” or “this completed.”

**Completion And Failure Feedback**

When a cast finishes successfully, the bar briefly flashes with a success-style green/white look, then clears.

When a cast is interrupted or fails, it briefly shows a failure state with red styling and visible text like “Interrupted” or “Failed.”

That means the player no longer has to guess whether the bar simply disappeared or whether the cast actually completed.

**Pet Cooldown Clarity**

The pet action buttons were also improved. Previously, a pet cooldown could appear as just a bare number, like `7`, which is vague.

Now the pet button gives better context:

- Shows cooldown units like `7s`
- Keeps the action icon visible
- Adds an action label, such as `Taunt`
- Adds an accessible label like “Taunt: 7s remaining”
- Adds an `AUTO` badge for autocast instead of relying only on glow/color
- Adds a cooldown overlay so it visually reads as cooling down

So instead of seeing only `7`, the player can understand: “Taunt is cooling down, 7 seconds left, and auto-cast is enabled.”

**Accessibility And Localization**

New text was added through the game’s i18n/localization system instead of hard-coding English directly into the UI.

New localized concepts include:

- Interrupted
- Failed
- Channeling
- Cannot interrupt
- Interruptible
- Auto
- Cooldown remaining labels

The cast bars also now update their ARIA labels better, so screen readers get useful information such as what is being cast, the status, and remaining time.

Reduced-motion and forced-colors support were also updated, so the new animations and colors behave better for users with accessibility settings enabled.

**Styling**

The colors were moved into CSS tokens/custom properties instead of scattering hard-coded colors through the code.

That makes the visual language more consistent:

- Gold for normal casts
- Blue/cyan for channels
- Teal for consume actions
- Green/white for success
- Red for failure/interruption
- Steel/gray for uninterruptible casts
- Purple/red border cues for dangerous casts
- Pet-accent coloring for pet actions

And importantly, color is not the only signal. The UI also uses text cues, labels, badges, borders, and ARIA labels.

**Tests And Verification**

Several focused tests were added or updated for:

- Normal cast styling
- Channel styling
- Success outcome feedback
- Interrupted outcome feedback
- Pet cast/action styling
- Pet cooldown labels with units
- AUTO badge visibility
- ARIA labels
- i18n generated output

These passed:

- `npm run check:types`
- `npm run build`
- Focused cast/pet tests: 3 files, 34 tests passed
- i18n verification tests: 2 files, 22 tests passed, 1 skipped
- Manual Playwright smoke check against the local app

All cast bar and pet feedback states

EnriqueGF:
Reviewed as a senior pass. This is a clean, well-tested UI-only change. Verified locally on the branch: focused tests (cast_bar, cast_bar_painter, pet_action_feedback, hud_perf_budget) all pass, `tsc --noEmit` and svelte-check are clean, `architecture.test.ts` and the S3 `localization_fixes` guard are green, the i18n resolved hash matches the committed baseline, `i18n:gen` is idempotent, and `i18n_completeness` (M16) plus the status registry pass. The pure-core + thin-painter seam is respected, the write-elision contract is preserved (nice `classSignature` guard so classes only rewrite on change), and accessibility (reduced-motion, forced-colors, ARIA labels/valuenow) is handled well. The event ordering is safe: `handleEvents` runs before `hud.update()` in both the offline and online loops, so the authoritative `castStop{success}` outcome wins over the visibility-based success fallback.

A few non-blocking items to consider:

1. Memory growth in `castOutcomesByEntity` (src/ui/hud.ts). The `castStop` handler calls `startCastOutcome(...)` for every entity that emitted a `castStart`, including mobs that are never the player or target. Those entries are only read and pruned lazily inside `activeCastOutcome`, which runs solely for the player and current target. So outcome entries for never-targeted casters are written, never read, and never cleaned, giving a slow unbounded growth (and wasted work) over a long session in a busy zone. Consider only starting or keeping outcomes for entities you actually paint a bar for, or pruning expired entries. `castLabelsByEntity` has the same leave-interest edge already acknowledged for `castLoopIds`.

2. Duplication of the cue/label logic. `castCueText` and `nameplateCastDisplayName` in src/render/nameplate_painter.ts reimplement `CastBarPainter.cueText` (src/ui/cast_bar_painter.ts) and `castDisplayName` (src/ui/hud.ts). The special-case id map (fishing, demon_heal, thunzharr_stormcall) now has three copies, so the rule of three is met; a small shared helper would keep them from drifting.

3. The target cast bar now localizes via `castDisplayName`, which is a nice improvement, but it does not special-case boss mechanic ids that are not in `ABILITIES` (for example `nythraxis_deathless_rage`), so those still render as the raw id, same as before. Not a regression, just noting the localization does not cover boss casts.

4. Process note: this PR targets `main`, but the repo convention is to base contributions off the active release branch (currently release/v0.23.0). The maintainer may retarget; flagging in case.

None of these block correctness. Approving; item 1 is the one I would recommend addressing before or shortly after merge.

EnriqueGF:
One required change before this can be considered for merge: please retarget the PR base branch.

This PR currently targets `main`, but in this repo `main` trails the active release branch and is not the integration base. Contributions must target the current release branch, which is `release/v0.23.0`.

Please:
1. Rebase your branch onto `origin/release/v0.23.0`.
2. Change the PR base to `release/v0.23.0` (Edit on the PR, then pick the base branch).
3. Regenerate the i18n artifacts after the rebase if the resolved table or the pending count shifts (`npm run i18n:gen`, and `npm run i18n:hash -- --write` if the resolved table changed), then commit the result. A rising `pending` count after picking up the release branch is expected and fine at PR tier.

The review itself stands (approved with the minor non-blocking notes above); this is purely about the target branch.

EnriqueGF:
Thanks for your contribution btw :D

loopermallee:
Rebase/retarget update:

- Rebased `fix/castbar-pet-feedback` onto `upstream/release/v0.23.0` and force-with-lease pushed `0a9426919`.
- Regenerated i18n artifacts with `npm run i18n:gen` and refreshed the resolved hash with `npm run i18n:hash -- --write`; `node scripts/i18n_resolved_hash.mjs --check` passed. The higher PR-tier pending count from the release branch is expected.
- Focused validation passed:
  - `npx vitest run tests/cast_bar.test.ts tests/cast_bar_painter.test.ts tests/pet_action_feedback.test.ts`
  - `npx vitest run tests/hud_perf_budget.test.ts`
  - `npx vitest run tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts tests/i18n_completeness.test.ts`
  - `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts`
  - `node scripts/i18n_resolved_hash.mjs --check`
  - `npx tsc --noEmit`
  - `git diff --check`
  - `npm run security:gate`
  - `npm run build`

loopermallee:
Rebased onto levy-street/world-of-claudecraft release/v0.23.0 at c38c293 and force-pushed fix/castbar-pet-feedback at d254f5ed7.

PR base is release/v0.23.0. GitHub now reports mergeable=true, mergeable_state=unstable, rebaseable=true, and behind_by=0.

i18n artifacts were regenerated and the resolved hash was updated/confirmed: 36c5f642ee6e37f3c7190aaeda2f0135293cc60346992b9b7de724ab9db6479f.

Focused validation passed: npm run i18n:gen; npm run i18n:hash -- --write; node scripts/i18n_resolved_hash.mjs --check; npx vitest run tests/cast_bar.test.ts tests/cast_bar_painter.test.ts tests/pet_action_feedback.test.ts; npx vitest run tests/hud_perf_budget.test.ts; npx vitest run tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts tests/i18n_completeness.test.ts; npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts; npx tsc --noEmit; git diff --check; npm run security:gate; npm run build.

Non-blocking review suggestions were intentionally left untouched unless separately requested.

EnriqueGF:
Thanks again for the contribution, and for turning the retarget around so quickly :)

The base-branch gate now passes: this PR targets `release/v0.23.0`, which is the latest release branch, and the branch is a clean descendant of it (`behind_by = 0`). Thank you.

I re-reviewed the code on the rebased head (`d254f5ed7`) in a clean worktree. Everything I checked mechanically is green: `npx tsc --noEmit`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/i18n_completeness.test.ts`, `tests/parity`, and the four change tests (121 passed). The i18n work is correct: English-only in `i18n.catalog/hud_chrome.ts`, with the five non-Latin overlay fills M16 requires for the new wordy values. No em/en dashes, no emojis. The `classSignature` write-elision guard is a nice touch.

Unfortunately, going deeper into the sim side turned up a correctness bug that I do not think we can ship, plus the leak from my earlier pass which is still open. Ordered by severity.

**1. (Blocking, high) The target/boss cast bar reports "Complete" for casts that never completed.**

Mobs never emit `castStart` or `castStop`. The only source of a mob cast bar is `bigCast`, which is set at `src/sim/mob/locomotion.ts:485` and cleared with no event at three places:

- `src/sim/mob/locomotion.ts:459` on completion
- `src/sim/mob/locomotion.ts:646` and `src/sim/mob/lifecycle.ts:90` on abort, whose own comment reads "A mid-flight bigCast dies with the pull: clear the bar"

The interrupt path cannot cover it either: `src/sim/combat/effect_dispatch.ts:281` breaks out when `!ABILITIES[target.castingAbility]`, and a `bigCast` id is never in `ABILITIES` (that is exactly why the target bar used to show the raw id).

So for any mob target, `Hud.castOutcomeForEntity` (`src/ui/hud.ts:5910-5927`) always falls through to the visibility-based fallback and unconditionally calls `startCastOutcome(..., 'success', ...)`. When Nythraxis or Thunzharr leashes, resets, or loses the pull mid-`bigCast`, the target cast bar flashes green with "Deathless Rage: Complete" even though the cast was aborted. The `'interrupted'` arm is dead code for every PvE boss cast. That is wrong actionable information on precisely the bar the comment at `src/ui/hud.ts:6590` describes as the one "the raid sees exactly when to channel the wardstones".

Two ways out: emit `castStop{success}` from the mob `bigCast` paths (true on completion, false on abort), or restrict the visibility fallback to entities that actually emit cast events. I would prefer the former, since it also makes the online host correct for free.

**2. (Blocking, medium-high) `castLabelsByEntity` and `castOutcomesByEntity` leak, and a stale entry produces a bogus "Complete" flash.**

`castLoopIds` deliberately carries three cleanup arms: `castStop` (`src/ui/hud.ts:8286`), `death` (`:8315`), and the leave-interest reconcile in `reconcileSfx` (`:8201-8206`), whose comment states that neither `castStop` nor `death` arrives when a caster leaves interest mid-channel. The two new maps have only the `castStop` arm (`:9433`, `:9438`).

- `castStart` (`:9432`) writes an entry for every casting entity in interest. A caster that dies or leaves interest mid-cast never clears it, so `castLabelsByEntity` grows without bound.
- Worse than a leak: that stale entry is later read. If a player leaves interest mid-cast and you target them afterwards while they are idle and alive, `castOutcomeForEntity` (`:5919-5923`) finds the stale ability id and flashes "Complete" for a cast abandoned minutes earlier.
- `castOutcomesByEntity` is pruned only inside `activeCastOutcome`, which runs solely for the player and the current target, so entries for every other caster are written, never read, and never freed. This is the item from my earlier pass; it is still open.

Please add a reconcile arm next to `reconcileSfx` and clear both maps on `death` and on world reset, so the three maps have the same lifecycle.

**3. (Medium) The pet bar now fully rebuilds about twice as often, all of it in the last second of the cooldown.**

`renderPetBar` tears the bar down with `bar.innerHTML = ''` (`src/ui/hud.ts:5972`) and recreates every button, `iconDataUrl()` and listener whenever `sig` changes, and `sig` now embeds `tauntCooldownText` (`:5969`). I measured it against the real `formatPetCooldownShort`: over one 10-second `PET_GROWL_INTERVAL`, the old `Math.ceil` value produced 11 distinct signatures, the new one produces 21. All ten extra rebuilds land in the final second, roughly one every two frames at 20 Hz.

Related: `--pet-cooldown-progress` is written only on rebuild, so the new conic sweep is a 1 Hz staircase for nine of the ten seconds rather than a smooth sweep. Keeping the whole-second value in the signature and pushing the sub-second text and the sweep percent through the elided writers on a stable node would fix both.

Minor, same function: `formatPetCooldownShort(1.0)` returns "1s" while `(0.95)` returns "1.0s", so the label visibly changes format mid-countdown.

**4. (Medium) The duplicated cue logic has already diverged.**

`castCueText` (`src/render/nameplate_painter.ts:61`) reimplements `CastBarPainter.cueText` (`src/ui/cast_bar_painter.ts:263`) but drops both of its guards: it has no `showInterruptCues` option and no `source !== 'pet'` check. A pet's nameplate therefore reads "Pet, Interruptible: ..." while the same pet on the target bar reads only "Pet: ...". `nameplateCastDisplayName` vs `castDisplayName` has the same problem, and the fishing / demon_heal / thunzharr_stormcall id map is now in its third copy. That meets the rule of three; please extract one shared helper.

**5. (Medium) Nameplate cast label overflow.**

`.np-castbar` is 78px wide and `.np-castlabel` (`src/styles/hud.css:135-144`) is `inset: 0; font-size: 9px` with no `white-space` or `text-overflow`. The label is now prefixed with a comma-joined cue list, so "Pet, Channeling, Danger, Cannot interrupt: Shadow Bolt" will wrap or spill out of an 8px-tall bar.

**6. (Medium) Test coverage does not reach the new behavior.**

The outcome state machine is the load-bearing change in this PR and has no test at all, because it was added as a method cluster on `hud.ts` rather than extracted into a pure core behind the seam the repo asks for. Bug 1 would have been caught by a single test of the fallback. `nameplate_painter`'s cue and label helpers are untested. In `cast_bar.ts`, `interruptState`'s ability-`uninterruptible` arm is uncovered, and `isImportantCast`'s `boss` and `bigCast` arms are conflated because `thunzharr_waking_peak` is both, so deleting either arm keeps the suite green. `formatPetCooldownShort(0) -> undefined` is unasserted, and the "signature changed, so classes are rewritten" arm of `classSignature` is unproven since every painter test builds a cold painter.

Extracting `castOutcomeForEntity` / `activeCastOutcome` into a host-agnostic module taking a clock and the maps would make all of this testable and would satisfy module-first.

**7. (Should fix) Screenshots.**

The repo asks for before/after on visual changes, desktop and mobile, committed under `docs/screenshots`. The PR has one after-only composite of the cast bar states. Missing: before/after of the pet action bar (the new AUTO badge, the action label, the cooldown sweep, and the `.cdtext` inset change from `inset: 0` to `inset: 7px 0 10px`) and of the nameplate cast bar.

Items 1 and 2 are the ones that block. The direction of this change is good and the accessibility work (reduced-motion, forced-colors, `aria-valuenow`, non-color cues) is genuinely well done, so I am keen to see it land once the outcome signal tells the truth.

EnriqueGF:
One addition to item 1 above, from a follow-up parity pass. There is a second, online-only way the same false "Complete" appears, this time for a player target.

Player entities keep mirroring out to `INTEREST_DROP_RADIUS = 100` (`server/game.ts:122`), but `castStart` / `castStop` are world-scoped events routed only within `EVENT_RADIUS = 90` (`server/game.ts:144`, applied at `server/game.ts:4462`). In the 90 to 100 yd band your client still receives the target's mirrored `cast` / `castRem` / `castTot` fields and paints the bar, but it never receives the `castStop`. So when that player is interrupted, `castOutcomeForEntity` falls through to the visibility fallback and flashes "Complete" rather than "Interrupted". Offline this cannot happen, so it is an online-only inaccuracy on top of the host-agnostic bug in item 1.

This reinforces the suggested fix: the outcome should not be inferred from the bar merely disappearing. Whatever the source, treat "bar vanished with no `castStop`" as unknown and paint nothing, rather than as success.

For the record, the parity pass otherwise came back clean: `templateId` rides the wire as `tid` (`server/game.ts:657`, decoded at `src/net/online.ts:1531`) and the cast fields are re-sent whole while casting, so `isImportantCast`, `interruptState`, and `castSource` in `src/render/cast_bar.ts` resolve identically on all three hosts. No `IWorld` or wire-protocol drift is introduced, and the own-player interrupt path is correct online because the server routes events before it broadcasts snapshots.

loopermallee:
Thanks again for the careful follow-up review and for the patience/advice throughout this PR.

I dug into the cast outcome lifecycle and agreed with the two blockers:

- The UI should not infer `Complete` from a cast bar simply disappearing.
- `castLabelsByEntity` / `castOutcomesByEntity` need the same cleanup lifecycle as the existing cast-loop state.

Update summary:
- Removed the visibility-based success fallback. A vanished cast bar without an authoritative `castStop` now produces no outcome flash.
- Moved and covered the cast outcome logic with focused tests so the state machine is no longer only exercised through `Hud`.
- Added cleanup for cast label/outcome state on stop, death, interest reconciliation, and world reset.
- Added/adjusted scripted cast handling so boss/mob cast outcomes are driven by explicit success/failure where appropriate.
- Addressed the duplicated cue/label drift and nameplate overflow.
- Reworked the pet cooldown update path so cooldown text/progress does not force unnecessary full pet-bar rebuilds.
- Where scripted cast events changed observable sim output, I treated those as intentional authoritative outcome signals and covered the new behavior with focused tests.

Validation:
- `npx vitest run tests/cast_bar.test.ts tests/cast_bar_painter.test.ts tests/pet_action_feedback.test.ts` passed.
- `npx vitest run tests/hud_perf_budget.test.ts` passed.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts` passed.
- `npx vitest run tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts tests/i18n_completeness.test.ts` passed.
- `npx vitest run tests/cast_outcome_core.test.ts tests/cast_presentation.test.ts tests/world_boss.test.ts tests/nythraxis_encounter.test.ts tests/pet_action_feedback.test.ts` passed.
- `npx vitest run tests/parity/coverage.test.ts` passed.
- `npx vitest run tests/parity/parity.test.ts` passed after updating the intentional `nythraxis_full_pull` event digest.
- `npx tsc --noEmit` passed.
- `npm run i18n:gen`, `npm run i18n:hash -- --write`, and `node scripts/i18n_resolved_hash.mjs --check` passed.
- `npm run build` passed with existing Vite/admin i18n chunk warnings.
- `git diff --check` and `git diff --cached --check` passed.


The main behavior change is that the UI now only says `Complete` when the game has actually reported completion, rather than guessing from the bar disappearing.

loopermallee:
Screenshot evidence is now committed in `a8ed48855`.

The contact sheet below was generated from the running local game app on this PR branch, using the actual HUD/nameplate DOM and CSS from `/play.html`.

PR #1625 cast bar and pet feedback contact sheet

<details>
<summary>Individual screenshots</summary>

### Player cast states

Consume / Eat / Drink

Normal hard cast

Successful Complete

Channel

Interrupted

### Target and nameplate states

Interruptible enemy cast

Cannot interrupt / uninterruptible cast

Important / Danger cast

Nameplate cast states

### Pet action and cooldown states

Pet cast / pet action

Pet cooldown

Pet autocast cooldown

Pet autocast ready

### Mobile

Normal hard cast mobile

</details>

Manifest and capture notes are in `docs/screenshots/pr-1625/README.md`.

PR #1627, opened 2026-07-07
fix(ui): preview every chroma swatch on hover in the character creator (#1464)

### LINKED ISSUE #1464: Bug: Character creator preview inconsistent
### Before you start

### What happened?

In the character creator menu, the outfit preview does not update consistently. 
In the image below, I have selected the second outfit, which it shows correctly.

However, when i hover over the first outfit, the preview doesn't update:
1st - No updated preview
2nd - Currently selected and shown
3rd - Updates correctly when I hover the cursor here
4th - Updates correctly when I hover the cursor here

### Steps to reproduce

1. Open character creator
2. Select the first outfit
3. Select the second outfit, note that the preview updates
4. Hover the cursor on the first outfit
5. Preview doesn't update.

When I hover over the first outfit, I expect the preview character to show the outfit

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on linux mint

Image

daxdax89:
## Problem

In the character creator, hovering an outfit (chroma) swatch inconsistently updated the avatar preview: after selecting one outfit, hovering the swatch directly next to the selected one did nothing, while non-adjacent swatches previewed fine (issue #1464).

## Root cause

Each swatch carried its own `mouseenter` (preview this chroma) **and** its own `mouseleave` (revert to the selected chroma). When the pointer moves directly between two adjacent swatches, the departed swatch's `mouseleave` and the entered swatch's `mouseenter` both fire, and their order is not guaranteed. In Firefox the `mouseenter` ran first (preview the hovered chroma) and the neighbor's `mouseleave` ran second (revert to the selection), so the preview snapped straight back and the hover looked like a no-op. Only the swatch adjacent to the current selection is affected, which is why it read as "outfit 1 does not update" (outfit 2 was selected).

## Fix

Extract the wiring into `src/ui/skin_picker.ts` and bind the revert **once on the row** instead of per swatch. A row-level `mouseleave` does not fire on child-to-child transitions, so every swatch previews consistently on hover and the preview still reverts to the committed selection exactly when the pointer leaves the picker. `renderSkinPicker` becomes a thin consumer that builds the portrait swatches and delegates the interactions.

The row element persists across class switches (only its swatches are replaced), so `wireSkinPicker` aborts the previous wiring's listeners (via an `AbortController` tracked per row) before re-wiring, so re-renders replace the handlers rather than stacking stale reverts that would retain removed swatch nodes.

## Tests

`tests/skin_picker.test.ts` drives the wiring with a hand-rolled fake DOM and reproduces the exact repro: with outfit 2 selected, hovering its neighbour previews; moving between swatches never reverts; leaving the row reverts to the committed selection; a click commits and moves the selection; and re-wiring the same row does not stack listeners.

## Note on screenshots

There is no static visual difference: the picker renders identically. The defect is a hover-timing race on the 3D avatar preview, so the fix is behavioral (event wiring), covered by the unit test that reproduces the report. Happy to attach a screen recording of the hover if useful.

Fixes #1464


EnriqueGF:
:) Thanks for the clean, well-scoped fix and the reproduction test.

Reviewed and approved. Verified against the real code, not just the description:

**Base branch gate:** targets `release/v0.23.0`, which is the latest release branch. PASS.

**What I checked:**
- Root cause and fix are correct. Moving the revert from a per-swatch `mouseleave` to a single row-level `mouseleave` is the right fix for the child-to-child transition race (issue #1464), since `mouseleave` does not fire on transitions between descendants of the row.
- The per-row `AbortController` in `ROW_CONTROLLERS` correctly prevents stale row-level handlers from stacking across class switches, since `renderSkinPicker` reuses the persistent row (`row.innerHTML = ''` at main.ts:2810 keeps the row and replaces only the swatches).
- `selectedSkin()` and the `Number.isFinite` fallback are sound; `dataset.skin` is always set to `String(i)` at the call site, and the no-selection case falls back to `current`.
- module-first respected: logic extracted into `src/ui/skin_picker.ts` rather than growing `main.ts`; the coordinator becomes a thin consumer.
- i18n: no new player-visible strings; the aria label still goes through `t('auth.chromaOption', ...)`. No locale overlays touched.
- No em/en dashes or emojis.

**Ran locally (fresh worktree):**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/skin_picker.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts`: all pass (localization after `npm run i18n:gen`).
- `biome check` on the two changed files: clean.

**Test quality:** the five cases are decisive and map to the report (adjacent-swatch preview, no swatch-to-swatch revert, click commit plus revert-to-new-selection, no listener stacking on re-wire, fallback when nothing is selected). The hand-rolled fake DOM models the `{ signal }` removal contract, so the 'does not stack' assertion is meaningful.

**Non-blocking nits:**
- The branch is a couple of commits behind the `release/v0.23.0` tip. No conflict here (one new file plus a small isolated hunk), but a quick rebase before merge keeps it tidy.
- Screenshots: agreed there is no static visual diff. The offered short screen recording of the hover would be a nice-to-have for the record, but not required given the behavioral test.

Nice work.

zfreiberg:
Tested locally: checked out the branch, ran the server and client with the PR's changes.

Ran the automated test file:
npx vitest run tests/skin_picker.test.ts --reporter=verbose
→ all 5 tests pass (adjacent-swatch preview, no revert while moving between swatches, click commits selection, no stacked listeners on re-render, fallback skin when nothing selected).

Manually verified in the character creator:
- Selected a swatch, then hovered the adjacent one (the exact broken case from #1464) — preview now updates correctly.
- Moved the cursor across several swatches in a row — preview tracks smoothly, never snaps back mid-hover.
- Moved the cursor off the swatch row — preview reverts to the actual selection.
- Switched character class (re-renders the swatch row) and repeated the hover test — still works, no stale-listener regression.

Video attached showing the above.



https://github.com/user-attachments/assets/b42784c1-371f-4a00-aea5-0856bc6fea65

daxdax89:
Friendly nudge: you reviewed and approved this on v0.24 (and zfreiberg community-tested it). It is now rebased cleanly onto release/v0.26.0, skin-picker fix and no-stacking regression test intact. Good to merge when you have a moment?

PR #1628, opened 2026-07-07
fix(sim): key solo dungeon instances to the durable character id (#1600)

### LINKED ISSUE #1600: Critical: Gravewyrm Sanctum can be farmed with zero downtime via arena full-restore and relog instance reset
## Problem

Gravewyrm Sanctum can be farmed solo with no downtime and no lockout by composing two
unintended interactions. Neither system is individually broken, but together they let a
player kill the finale boss (Korzul the Gravewyrm), reset the whole instance, restore to
full, and re-enter over and over with no recovery cost.

The two load-bearing mechanics:

1. An arena match is a free, instant full restore. Entering a match resets every fighter
   to full HP, full resource, and clears all cooldowns and auras during the pre-fight
   countdown, and every match ending (defeat, timeout, or a leave) teleports the player
   back to the exact spot they queued from, fully restored. There is no fee, no level
   gate, and no out-of-combat gate on queueing. Because entering a dungeon does not remove
   a queued player from the queue and matchmaking has no position check, a player can queue
   in the overworld, walk into the Sanctum, be matched from inside, and be returned to
   their in-Sanctum position at full HP and mana. Logging out does not heal (a relog reloads
   the saved damaged HP and restores cooldowns), so the arena is the heal source.

2. A full logout mints a fresh instance. Solo dungeon instances are keyed by the live entity
   id. Using the character-select "Take Over" action on your own still-online (linkdead)
   character forces a full leave and rejoin, which assigns a new entity id, which changes the
   solo instance key, so re-entering claims a fresh slot and respawns every mob including the
   boss. Standard dungeons have no re-entry lockout (only the Nythraxis raid does), and this
   bypasses the 300 second empty-instance reset timer. A same-character auto-reconnect during
   the linkdead grace resumes in place and does not reset, and the plain leave-dungeon command
   keeps the same entity id (same cleared instance), so the deliberate Take Over is what makes
   the boss re-farm work.

Rogue Stealth plus Sprint stack to roughly 0.85x run speed while invisible, and the Sanctum
has no inter-boss door gating, so a rogue can run straight to Korzul past the trash. That part
is intended classic-style play; the exploit is the free restore plus the free instance reset,
not the run.

Severity is high as a gameplay and economy exploit (Korzul drops a guaranteed uncommon plus
roughly a 60 percent chance at a rare or epic, re-rolled every kill with no lockout). It is
not a security or RCE issue.

## Steps to reproduce

1. On a level 20 rogue, queue for 1v1 arena in the overworld near the Gravewyrm Sanctum door.
2. Walk into the Sanctum while still queued, Stealth plus Sprint past the trash to Korzul, and
   kill and loot him.
3. When the arena match forms (needs a second queuer, so a colluding partner or a populated
   live queue), observe you are healed to full and returned to your position inside the Sanctum.
   End the bout by any means (a colluding partner intentionally dying is the clean path).
4. Go to character select and use "Take Over" on your own character to force a full relog.
5. Re-enter the Sanctum: the instance is fresh and Korzul is alive again.
6. Repeat from step 2 with no downtime, no run-back, and no lockout.

## Expected behavior

- Recovering HP, mana, and cooldowns after a dungeon fight should cost time or resources
  (natural regen, food or drink, or a real trip), not be free and instant.
- Leaving and re-entering a dungeon should not respawn a boss you already killed; a cleared
  instance should stay cleared for a reasonable window regardless of relog or Take Over.
- An arena match should not be usable as a mid-dungeon or post-dungeon full restore.

## Scope

- Arena restore-on-return: `src/sim/social/arena.ts` (`returnFromArena`, `resetForArena`,
  `readyArenaFighter`; the match-formation position snapshot at roughly lines 609 to 613).
- Arena queue eligibility while inside an instance: `arenaQueueJoin` and the matchmakers
  (`matchmakeArena1v1` / `matchmakeArena2v2`) in `src/sim/social/arena.ts`, plus `enterDungeon`
  in `src/sim/instances/dungeons.ts` (no dequeue on entry).
- Solo instance keying and reset: `instanceKeyFor` in `src/sim/instances/dungeons.ts`
  (`solo:${entityId}`), `Sim.addPlayer` new-entity-id path, and `INSTANCE_EMPTY_TIMEOUT`.
- Login placement of a dungeon-saved character to the door: `Sim.addPlayer` in `src/sim/sim.ts`.
- The Take Over relogin path: `takeOverCharacter` / `kickSession` / `leave` in `server/game.ts`.

## Actual behavior

- Queueing has no fee, no level gate, and no out-of-combat gate, and a player queued outside who
  then enters the Sanctum stays eligible and is matched and healed from inside the instance.
- Every arena match ending returns the player fully restored (full HP, full resource, cooldowns
  and auras cleared) to their queue position.
- A Take Over relog mints a fresh instance with the boss respawned, with no lockout and bypassing
  the empty-instance reset timer.

## Acceptance criteria

- A player cannot use an arena match to restore HP, mana, or cooldowns for free; leaving a bout
  no longer heals damage taken outside the arena (the in-match clean slate can stay).
- A player who is inside a dungeon instance cannot be matched into an arena bout (dequeued on
  entry, or filtered out by matchmaking).
- A logout, relog, or Take Over does not respawn a boss the character already killed; re-entry
  rejoins the same cleared instance for a defined window, or a per-character Sanctum lockout is
  enforced.
- Normal arena play (a real 1v1 or 2v2 between two consenting players who queued from the
  overworld) is unchanged, and normal dungeon runs are unchanged.

## Notes

Suggested minimal fixes, smallest surface first:

- Do not full-heal on arena return. Snapshot each fighter's HP and resource at match formation
  alongside their return position, and restore those pre-match pools in `returnFromArena` instead
  of the full `resetForArena`. This keeps the intended in-match clean slate but removes the free
  carry-back that is the core of the exploit.
- Dequeue on instance entry, and add a position filter to matchmaking, to close the
  queue-outside-then-walk-in variant.
- Close the boss respawn by keying solo instances on the durable character id instead of the
  entity id, or extend the existing Nythraxis-style per-character lockout to the Sanctum finale.

Related but separate findings from the same investigation, likely their own issues:

- Arena win-trading: matchmaking pairs the closest-rated players and there is no anti-rematch
  guard, so a colluding pair can queue together and feed each other rating.
- No logout-in-combat penalty: a full leave has no combat check.
- A "walk through walls" report near the Sanctum that needs a concrete repro (which wall, roughly
  where) before it can be investigated; likely a collider gap or a wall modeled as a jumpable fence.

Investigated on release/v0.22.0.

daxdax89:
## Problem (critical farm exploit)

Gravewyrm Sanctum, and any standard solo dungeon, could be farmed with zero downtime (issue #1600). Solo instances were keyed by the **transient entity id** (`solo:<entityId>`), and a logout/relog or a character-select **Take-Over** mints a new entity id. That changed the instance key, so re-entering claimed a **fresh instance with every mob (including the finale boss Korzul) respawned**, bypassing the 300s empty-instance reset that is meant to throttle a re-farm. Standard dungeons have no per-character lockout (only the Nythraxis raid does), so nothing else stopped it. Korzul drops a guaranteed uncommon plus a high chance at a rare/epic, re-rolled every kill, so this is a real economy exploit.

## Fix

Derive the solo key from the **durable character id** instead: `solo:char:<cid>`, falling back to the entity id (`solo:<pid>`) when there is none (offline / headless / tests). Re-deriving the same key across a relog makes re-entry **rejoin the still-live cleared instance** instead of minting a fresh one, so the boss stays dead until the intended empty-instance reset, exactly like an honest player who waited.

- Normal play is unchanged: a legitimate mid-run relog now rejoins its instance (a correctness improvement), and party runs are untouched.
- The `solo:` prefix is preserved so the delve auto-companion gate (`startsWith('solo:')`) still fires for solo players, and the `char:` marker keeps the durable key from ever colliding with the entity-id fallback.
- `instanceKeyFor` is opaque to its consumers (`partyMembersForKey` re-derives and compares whole keys; nothing parses the id out of the string), and instances are runtime-only state (never serialized), so there is no wire-parity or persisted-state impact. The derivation is a pure read (no rng, no tick-phase change), so determinism and draw order are unaffected.

## Tests

`tests/dungeons.test.ts` pins the durable key and the entity-id fallback, and reproduces the exploit: a relog with the same character id but a new entity id rejoins the same instance slot with the same boss entity, instead of claiming a second instance with the boss respawned (which failed before the fix). The parity golden-trace gate and `tests/architecture.test.ts` (sim purity) stay green; the delve suites confirm the solo auto-companion still spawns.

## Scope note

This closes the reported boss-respawn farm at its root. The issue also lists two related-but-separate concerns from the same investigation that I deliberately kept out of this PR to keep it one cohesive change: the arena free-heal-on-return (a fighter can leave a bout topped off) and a matchmaking position filter for queueing while inside an instance. Happy to follow up on those as their own PR if you would like.

Fixes #1600

### DIFF
--- a/src/sim/instances/dungeons.ts
+++ b/src/sim/instances/dungeons.ts
@@ -41,7 +41,18 @@ const RAID_REQUIRED_DUNGEON_IDS = new Set(['nythraxis_boss_arena']);
 
 export function instanceKeyFor(ctx: SimContext, pid: number): string {
   const party = ctx.partyOf(pid);
-  return party ? `party:${party.id}` : `solo:${pid}`;
+  if (party) return `party:${party.id}`;
+  // Key a solo instance to the DURABLE character id, not the transient entity
+  // id. A logout/relog or a character-select Take-Over mints a NEW entity id,
+  // which under an entity-id key spawned a fresh instance with the boss alive
+  // again, bypassing the empty-instance reset timer (the Gravewyrm Sanctum
+  // zero-downtime farm, issue 1600). Re-deriving the same key across relog makes
+  // re-entry rejoin the cleared instance instead. Keep the `solo:` prefix (the
+  // delve auto-companion gates on startsWith('solo:')) and mark the
+  // character-keyed form so it can never collide with the entity-id fallback,
+  // which offline / headless / tests (no character id) still use.
+  const cid = ctx.players.get(pid)?.characterId;
+  return cid != null ? `solo:char:${cid}` : `solo:${pid}`;
 }
 
 export function instanceOriginOf(inst: InstanceSlot): { x: number; z: number } {
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -397,8 +397,8 @@ import {
   type SkinCatalog,
   type SkinRank,
   type SportRole,
-  steadyAngleTo,
   SUNDER_ARMOR_PCT_PER_STACK,
+  steadyAngleTo,
   swingMissChance,
   type VcBracket,
   type VcNationId,
@@ -700,7 +700,7 @@ export interface InstanceSlot {
   dungeonId: string;
   difficulty: DungeonDifficulty;
   slot: number;
-  partyKey: string | null; // party id or 'solo:<pid>'
+  partyKey: string | null; // 'party:<id>', 'solo:char:<characterId>', or the 'solo:<pid>' entity-id fallback
   mobIds: number[];
   objectIds: number[];
   exitId: number | null;
--- a/tests/dungeons.test.ts
+++ b/tests/dungeons.test.ts
@@ -980,6 +980,48 @@ describe('dungeons: pure helpers', () => {
     expect(instanceKeyFor(sim.ctx, b)).toBe(`party:${party.id}`);
   });
 
+  it('keys a solo instance to the durable character id when present (issue 1600)', () => {
+    const sim = makeSim();
+    // Online players carry a durable character id; the solo key is derived from
+    // it, not the transient entity id, so a relog (new entity id) rejoins the
+    // same instance. The `solo:` prefix is preserved (the delve auto-companion
+    // checks startsWith('solo:')) and the `char:` marker keeps it from ever
+    // colliding with the entity-id fallback used offline / in tests.
+    const withCid = sim.addPlayer('warrior', 'Durable', { characterId: 77 });
+    expect(instanceKeyFor(sim.ctx, withCid)).toBe('solo:char:77');
+    // No character id (offline / headless / most tests): fall back to entity id.
+    const noCid = sim.addPlayer('mage', 'Legacy');
+    expect(instanceKeyFor(sim.ctx, noCid)).toBe(`solo:${noCid}`);
+  });
+
+  it('a relog rejoins the cleared solo instance instead of respawning the boss (issue 1600)', () => {
+    const sim = makeSim();
+    const cid = 4242;
+    const pid1 = sim.addPlayer('rogue', 'Exploiter', { characterId: cid });
+    enterDungeon(sim.ctx, 'gravewyrm_sanctum', pid1);
+    const inst1 = claimedDungeon(sim, 'gravewyrm_sanctum');
+    expect(inst1, 'gravewyrm_sanctum did not claim a solo instance').toBeTruthy();
+    const slot1 = inst1.slot;
+    const bossId1 = mobInInstance(sim, inst1, 'korzul_the_gravewyrm').id;
+
+    // Simulate the Take-Over relog exploit: drop the player (the instance
+    // lingers, its 300s empty reset not yet elapsed) and re-add the SAME
+    // character, which assigns a NEW entity id.
+    sim.removePlayer(pid1);
+    const pid2 = sim.addPlayer('rogue', 'Exploiter', { characterId: cid });
+    expect(pid2).not.toBe(pid1);
+    enterDungeon(sim.ctx, 'gravewyrm_sanctum', pid2);
+
+    // The relog must rejoin the cleared instance, not mint a fresh one with the
+    // boss respawned: exactly ONE claimed instance, same slot, same boss entity.
+    const claimed = (sim.instances as any[]).filter(
+      (i) => i.dungeonId === 'gravewyrm_sanctum' && i.partyKey !== null,
+    );
+    expect(claimed.length).toBe(1);
+    expect(claimed[0].slot).toBe(slot1);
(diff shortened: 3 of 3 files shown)


daxdax89:
Addressed the Codex nit: updated the `InstanceSlot.partyKey` doc comment in `src/sim/sim.ts` to document the new `solo:char:<characterId>` form (plus the `party:<id>` and `solo:<pid>` fallback), so future code won't assume solo keys always embed an entity id. (Biome also organized one import in that file when it was touched, to keep the changed-file lint gate green.) tsc + tests/dungeons.test.ts (36) still green; rebased onto the current release tip.

Rubsey (commented):
Good diagnosis and a clean fix. This correctly identified the transient entity id as the root cause of the #1600 boss-respawn farm, kept the `solo:` prefix the delve auto-companion gate needs (`src/sim/delves/runs.ts:397`, `src/sim/entity_roster.ts:211`), and the relog test reproduces the exploit faithfully. The design claims in the writeup all verified against the code: nothing parses an id out of the key (`partyMembersForKey` re-derives and compares whole keys, sim.ts:6662), and no other site builds a `solo:` key inline.

The reason this is now conflicting, though, is that the same fix already landed on `release/v0.26.0`. The v0.25.0 issue batch (#1839, commit 877a10189, dated after this PR was opened) shipped a functionally identical `instanceKeyFor`: `solo:char:<characterId>` with the `solo:<pid>` entity-id fallback. It also covers the two legs your scope note deliberately left out (arena returns restore pre-match pools instead of full-healing, and dungeon entry dequeues from arena matchmaking), and issue #1600 is closed. Test coverage is subsumed too: `tests/gravewyrm_farm_exploit.test.ts` on the base pins the Take Over relog rejoin, per-character keying, the entity-id fallback, and the empty-timeout reset, which covers both tests added here.

So the one real conflict (`src/sim/instances/dungeons.ts`) is the two fixes colliding inside `instanceKeyFor`, and resolving it would leave this PR with no net change. Recommend closing rather than rebasing. You independently found and fixed the same root cause before the batch landed, it just lost the race to the tree.

One small thing worth salvaging: the `InstanceSlot.partyKey` doc comment you updated is still stale on the base (sim.ts:775 still reads `party id or 'solo:<pid>'`). Not worth a PR of its own, but your corrected text is the accurate one if it can ride along a future sim.ts change.

Rubsey:
Closing as superseded: the same durable-character-id fix (plus the arena and matchmaking legs) landed on release/v0.26.0 via the v0.25.0 issue batch (#1839, commit 877a10189), and issue #1600 is closed. Details in the review above. Thanks for the careful diagnosis and writeup, it independently nailed the root cause.

PR #1629, opened 2026-07-07
feat(perf): report the live server tick rate in the performance overlay

Rubsey:
## Summary

The achieved server sim tick rate (ticks per wall-clock second) was measured nowhere: tickMsAvg and TickProfiler time the loop body, so when the event loop saturates and the dt clamp discards wall time they look healthy while the wall-clock rate sags below 20 (the v0.22 CPU incident blind spot). This PR measures it server-side and surfaces it as a new row in the game client's performance overlay.

## How it works

- **server/tick_rate_meter.ts (new, pure):** TickRateMeter counts committed sim ticks against hrtime over a 3s sliding window (1s warm-up). Allocation-free hot path, ring-overflow safe, exact window-edge anchoring so a healthy jittered loop reads 20.0 rather than a 20.1-20.7 flutter that would mask small sags. hrtimeToMs keeps the ns-to-ms conversion Float64-exact for ~285 years of uptime.
- **server/game.ts:** the 20 Hz loop records ticksRun per callback; the rate rides the shared snapshot head as an optional tickHz scalar (omitted during warm-up, so old clients and warm-up are indistinguishable from absent), and perfProfile() exposes it to the ops /api/perf view (both dispatch arms share that function) and the PERF_TICK_LOG heartbeat.
- **src/net/online.ts:** ClientWorld mirrors snap.tickHz into a public serverTickHz field (off-IWorld, same precedent as snapInterval), guarded against junk (non-number, NaN, Infinity, <= 0).
- **src/ui:** new default-off Server Tick Rate row in the overlay's network group, one decimal, hidden offline or until the server reports, with sag severity derived from TICK_RATE (good >= 19.5, warn >= 15, bad below). The hz PerfValue kind gains an optional digits field; the existing snapshot row renders unchanged.
- **i18n:** hudChrome.perf.labels.serverTick plus its five non-Latin fills (M16 wordy-value rule) and regenerated resolved tables + SHA baseline.
- **scripts/server_load_jitter.mjs:** the load harness now prints tickHz in its SERVER summary line.

Note: five touched files carry whole-file biome reformats (the changed-files CI gate requires touched files fully format-clean); the functional diff is much smaller than the line count suggests.

## Test plan

- [x] tests/tick_rate_meter.test.ts (new): warm-up nulls, healthy 20 Hz, catch-up bursts, clamp-induced sag to 10 Hz, window slide, partial-window blend, ring overflow, timer-jitter no-upward-flutter, stall-then-recovery, per-dimension invalid-input guards, hrtimeToMs precision at 2^53 ns
- [x] tests/snapshots.test.ts: head omits tickHz pre-warm-up then carries the measured rate (via a real GameServer broadcast, JSON.parse-validated), perfProfile().tickHz null-then-value, client mirror holds last on omission, junk rejection incl. +Infinity
- [x] tests/perf_overlay_model.test.ts + perf_metrics_sampler.test.ts: registry row (group, default-off, one decimal), severity arms, offline hiding, sampler passthrough and null gating
- [x] Guard tests green: architecture, world_api_parity, localization_fixes (S3), hud_perf_budget, i18n_resolved_equivalence, i18n_completeness (M16)
- [x] Full suite: 828 files passed; the only failures are the 8 pre-existing orphaned-key ones (see below)
- [x] tsc: exactly the 26 pre-existing errors, zero new; biome ci clean on all changed files; game + server builds green
- [x] Reviewed by cross-platform-sync, privacy-security-review, test-coverage-auditor, an adversarial correctness reviewer, and the qa-checklist gate; all findings addressed

## Known overlap with PR #1622

This branch inherits the orphaned loading.reconnecting key from the incomplete revert 92a16b284 (8 failures in tests/i18n_overlay_key_membership + i18n_build_gapfill, 26 tsc errors). PR #1622 fixes that separately. The i18n regen artifacts here overlap that PR: whichever lands second re-runs npm run i18n:gen + node scripts/i18n_resolved_hash.mjs --write to resolve.


FernandoX7 (commented):
Really clean, and unusually well tested for a change this deep. The core premise is right: `tickMsAvg`/`TickProfiler` time the loop body, so the dt clamp discards exactly the wall-clock sag this measures. Things I checked and liked:

- `TickRateMeter` is pure and host-agnostic, anchors on hrtime rather than callback counts, and the window-edge straddle handling holds up. I traced the stall-recovery case by hand (a 7s stall then a clamped burst reads ~9.2 Hz, inside the window, then recovers to 20 within one window) and it matches. The tests are decisive, not just present: the jitter test pins an upper bound (`<= 20.15`) so an upward flutter would actually fail, and each input-guard test asserts the specific corruption it prevents (Infinity, negative sum, NaN).
- The snap-head test is a real `GameServer` broadcast parsed through `JSON.parse`, so it proves the head stays valid JSON and that `tickHz` is omitted during warm-up (a warm-up head and an old client read identically as absent). The client mirror holds the last value on omission and rejects junk including `+Infinity`.
- Both `/api/perf` arms go through the shared `perfProfile()`, so the ops view and the overlay can't drift.
- Severity derives from `TICK_RATE` instead of a hardcoded 20, the row is default-off in the network group, and `serverTickHz` sits off-IWorld on `ClientWorld` like `snapInterval`, which is the right call for an online-only field.

tsc here is the 26 pre-existing `reconnecting` errors and nothing in the touched files; the guard suites (architecture, world_api_parity, S3, the two overlay suites, snapshots) are green.

Conflict: only `src/ui/i18n.resolved.sha256`, the standard stale-base regen (the new catalog key rebaselines the hash). Whoever lands after #1622 re-runs `npm run i18n:gen` + `node scripts/i18n_resolved_hash.mjs --write` and it clears; the file itself is well formed.

nit (optional): `tickHz` rides every snapshot head at 20 Hz even though the 3s window makes it move slowly. It is a tiny scalar and matches how `tick`/`time` already ride the head, so it is fine as is, but if head bytes ever matter you could send it less often and lean on the client's hold-last to cover the gaps.

FernandoX7:
Pushed two changes to clear the review items:

- **Merged the current `release/v0.22.1` base in.** That resolves the only conflict (`i18n.resolved.sha256`) and drops the orphaned `loading.reconnecting` key the branch inherited from the base at the time it was cut. The base has since cleaned that up (in the pet-pick-target merge), so the three-way merge removes it here too. tsc goes from the 26 pre-existing `reconnecting` errors to 0, and the 8 `i18n_overlay_key_membership` / `i18n_build_gapfill` failures clear. The i18n resolved tables and sha256 baseline are regenerated and verified to match.
- **Addressed the nit:** `tickHz` now rides the head at ~2 Hz (gated on `sim.time`, `TICK_HZ_HEAD_INTERVAL_S`) instead of on every 20 Hz snapshot. Warm-up omission, the immediate first emit after warm-up, and the old-client-reads-as-absent property are unchanged, and the client's hold-last covers the gaps between emits. Added a throttle test alongside the existing warm-up and mirror tests.

Verified locally: tsc clean, the four perf/snapshot suites plus the full guard set (architecture, world_api_parity, S3, hud_perf_budget, i18n_resolved_equivalence, M16 completeness) green, biome ci clean on the touched files, and the game + server builds pass.

Issue #1293, opened 2026-07-02
Professions: attunement quest system and content (lore quest, make-amends, hobby switch)


Nervescraper:
## Problem

The active archetype is set and switched by quests (#1129, re-scoped per its thread), but no quest content or flow exists for the first-time lore quest, the repeatable make-amends switch quest, or the standalone hobby switch.

## Expected behavior

- The first attunement into any given archetype (whether or not another was held before) is a lore quest that sets the active archetype (via #1129 state).
- Returning to a previously held archetype is a repeatable make-amends quest whose cost escalates with the per-character switch count (time, coin, materials, or lore, per tuning).
- A standalone hobby switch (keeping the majors, taking the other opposite craft) is its own attunement-style quest at a lower cost than an archetype switch, but never free: flipping between the two opposites must not become a de-facto second hobby.
- The quest is where the "what changes" summary is shown before committing (the legibility surface, its own issue).

## Scope

- `src/sim/content/`: attunement quest records (lore quest, make-amends quest, standalone hobby-switch quest), merged by `data.ts`.
- Hook quest completion to the #1129 active-archetype switch and switch-count increment.

## Acceptance criteria

- Completing the lore quest sets the active archetype for a character with none, and for the first entry into an archetype never held before.
- Completing the make-amends quest switches to a previously held archetype and increments the switch count; its cost is read from content, escalating with the count.
- The standalone hobby-switch quest swaps which opposite craft is empowered, at its lower cost.
- All quest text is a `t()`/sim-i18n key.

## Notes

Depends on #1129 (active-archetype state) and #1292 (craft/gather objective type). The escalation curve numbers are tuning (#1148). Delivery vehicle for the whole switch mechanic. Attunement is a ladder: each deeper tier quest is part of the same system (#1128 owns the per-tier gates); switching between Jack of All Trades and an archetype follows the same pattern but its mechanics are an open doc TBD. PR #1205 (implements #1129) already carries a make-amends cost formula (5 + 3 per prior switch); reconcile this issue's content against what #1205 ships so quest content wraps the existing mechanic instead of re-implementing it.

Full design: https://woc.nervemart.com/docs/professions-system

FernandoX7:
[...]

PR #1638, opened 2026-07-08
feat(professions): archetype empowerment ceiling, hobby/title read surfaces, recipe acquisition + salvage/disenchant, level-20 crafting hub gate

### LINKED ISSUE #1148: Professions: balance and tuning pass over open numeric questions
## Problem

The design doc leaves several numeric parameters explicitly TBD (wheel budget size, decay curve, hobby cap tier, tier distribution, effect costs). Once the mechanics they govern exist, this issue tracks confirming and documenting real values.

## Expected behavior

- Each TBD numeric parameter listed in the design doc's Open Questions section has a confirmed, documented value once its owning mechanic (from an earlier issue) has landed.

## Scope

- `src/sim/content/professions.ts`: replace placeholder constants with tuned, documented values as each owning mechanic is confirmed.
- A running list mapping each open question to the issue that resolves it.

## Acceptance criteria

- Every numeric parameter listed as TBD in the design doc is either tuned and documented, or explicitly deferred with a reason.

## Notes

This is a rolling tracking issue, not a single PR; it should be updated as each mechanics issue lands rather than done all at once at the end.

jgyy:
Closes #1148 (tuning pass). Builds on top of the already-merged professions
foundation (#1119-1147) with the remaining loose ends from the epic (#1152):
the archetype empowerment ceiling, the hobby/title read surfaces, the recipe
acquisition/salvage/disenchant gate, the crafting gold sink + throttle, and
the level-20 crafting hub gate.

## What this adds

1. **Archetype empowerment ceiling wired into crafting** (`5321f61d`) - the
   active archetype now actually caps crafting output quality/roll instead of
   being a display-only flag.
2. **Archetype titles, hobby read surface, battlefield xp gate** (`3537d19`)
   - grants the player-facing title for their currently active archetype,
   surfaces the "hobby" (the one opposite craft empowered to rare) on the
   character window, and gates Battlefield Experience xp to the active
   specialty so it cannot be farmed off an idle craft.
3. **Recipe acquisition gate, salvage/disenchant, crafting gold sink + throttle**
   (`6b880e63`) - recipes must now be acquired (not free-form), adds
   salvage/disenchant as a materials-recovery path, and a gold sink +
   output throttle on crafting to keep the economy in check.
4. **Level-20 crafting hub gate** (`edf6e838`) - gates the crafting hub zone
   behind level 20, and resolves the remaining open tuning questions from
   #1148.

## Flow

```mermaid
flowchart TD
    A[Player reaches level 20] --> B[Crafting hub unlocked]
    B --> C[Pick / confirm active archetype]
    C --> D[Archetype title granted]
    C --> E[Hobby: one opposite craft empowered to rare]
    F[Craft attempt] --> G{Recipe acquired?}
    G -->|no| H[Blocked: acquire recipe first]
    G -->|yes| I{Within archetype empowerment ceiling?}
    I -->|over ceiling| J[Capped output quality/roll]
    I -->|within ceiling| K[Full-quality craft]
    K --> L[Gold sink + output throttle applied]
    J --> L
    L --> M[Item produced]
    M --> N{Unwanted result?}
    N -->|yes| O[Salvage / disenchant for materials]
    P[Item used in the field] --> Q{Active specialty match?}
    Q -->|yes| R[Battlefield Experience xp granted]
    Q -->|no| S[No xp: off-specialty use does not farm xp]
```

## Testing

- `tests/professions_crafting_hub.test.ts`, `tests/professions_hobby_craft.test.ts`,
  `tests/professions_acquisition_salvage_sink.test.ts`,
  `tests/professions_battlefield_xp.test.ts` (extended), `tests/professions_crafting.test.ts`
  (extended) cover the new behavior.
- Parity goldens regenerated for the new `PlayerMeta.craftThrottle` field (session-only,
  never on the wire; state-only diff, no sim behavior change) in a separate `chore(parity)` commit.
- `npm run gate` passes clean (tsc, full vitest suite, biome on changed files, all builds).
- Rebased clean onto the current tip of `release/v0.23.0` (no merge conflicts).

## Screenshot

HUD


FernandoX7 (commented):
This is a well-run batch in most of the ways that are hard to get right: salvage and acquireRecipe are deliberately staged off IWorld and the wire with the same documentation pattern the archetype surface used, the W0c parity pins for `hobbyCraft` landed in all five places, `knownRecipes` persists with a back-compat load plus a round-trip test, `craftThrottle` is correctly session-only, and I verified the regenerated goldens are state-hash-only (rng draw counts and event hashes are byte-identical to base). The battlefield-xp gate has all four arms covered plus the additive-only spy guarantee.

One design-level problem needs a maintainer ruling before this merges.

**Blocking**

- `src/sim/professions/archetype.ts:187` (`archetypeCeilingFor`): issue #1129 defines the active archetype as "an adjacent pair (the two majors, empowered to the top tiers) plus one opposite craft as a hobby empowered to rare". This implements a single-craft archetype: unlimited only for `craftId === activeArchetype`, with the adjacent partner capped at common. Three concrete consequences:
  - Every combo recipe becomes permanently uncraftable once a player attunes. All three `COMBO_RECIPES` pair adjacent crafts at `minTier: 1` (`src/sim/content/recipes.ts:226,240,254`), and with any archetype set the partner craft's ceiling is 0, so `meetsComboRequirement` (`crafting.ts:234`, now reading `craftCeiling`) can never pass for anyone attuned. The adjacent-pair specialist is exactly who #1132 aims these recipes at; under #1129's pair model both majors are uncapped and combos work. `tests/archetype_ceiling.test.ts` ("fails a minTier-1 combo") currently pins the inverted behavior as correct.
  - The ceiling does not cap what #1129 says it caps. "Once attuned, every craft outside the active specialty caps at the common floor, dormant": as wired, `gainCraftSkill` stays unbounded, output quality still rolls off raw skill (`crafting.ts:320`), and feeding the capped capability into `tierProgressMultiplier` removes the diminishing-returns penalty for capped crafts (the new regression test pins full progress where base gave zero). A "dormant" craft still levels at full speed to 300 and its output quality keeps improving. The PR description's "caps crafting output quality/roll" is not what lands.
  - The same single-craft reading flows into `getHobbyCraft` (`archetype.ts:146`; under the pair model the opposite side has two candidate crafts, which is what makes #1293's hobby-flip quest meaningful) and the battlefield-xp gate (`battlefield_xp.ts:105`).

  This is latent today (the acceptance quest is retired content, `src/sim/content/zone1.ts:1000,1017`, so no player can attune yet), but merging entrenches the single-craft model in tests and the IWorld surface. If the pair wording in #1129 has been superseded, most of this collapses to the combo question alone; either way it needs the ruling first, and note the pair model needs `ArchetypeState` to actually name the pair, since one craft id cannot derive it (every craft has two adjacent neighbors).

**Should fix**

- `src/ui/hud.ts:8351`: the craft-denial mapping falls through to `hudChrome.crafting.insufficientMaterials` for any unmatched reason, so a `throttled` denial (reachable now: the 11th craft inside a 60 second window) shows "You do not have the materials for that." Same fallthrough for `recipe_unknown` (latent until a content recipe carries an `acquisition` list). Add a `hudChrome.crafting.throttled` key (and an unlearned-recipe key) with explicit branches. I verified the reason value travels the wire intact, so the mislabel is identical offline and online.
- `src/ui/hud.ts:5260` (commit 4f4a5b615): removing the mobile action-ring long-press tooltip is unrelated to professions and reverts another contributor's deliberate design; the removed comment documented exactly why it existed (without it, a long press casts and burns cooldowns while the player is trying to read a spell). Please split this into its own PR so it gets its own call. The mirror comment above the tap handlers (around `hud.ts:5206`, "peek-guard consume ... so long-press-to-inspect on touch behaves the same way") is also stale now.
- `src/sim/professions/crafting.ts:96-101` and `src/sim/types.ts:2241`: the reason union now contains both `unknown_recipe` (id does not exist) and `recipe_unknown` (exists, not learned). That pair will be confused, and only one of them renders. Rename the new one (for example `recipe_not_learned`) while nothing consumes it yet.
- `tests/professions_acquisition_salvage_sink.test.ts:74-81`: `acquireRecipe`'s success arm and `wrong_source` arm are never exercised. The test titled "acquiring from the correct source" mutates `meta.knownRecipes` directly, and no content recipe carries an `acquisition` list, so the learn path this PR adds is only covered on its denial arms. Drive `acquireRecipe` against a gated recipe that `recipeById` can resolve and assert `ok: true` plus `wrong_source`.

**Nits**

- `tests/professions_acquisition_salvage_sink.test.ts:168`: the gold-sink assertion recomputes the fee from `CRAFT_GOLD_SINK_COPPER_PER_BUDGET`, so re-tuning the rate can never fail it. Pin the rate (and the 10 per 60s throttle magnitudes) to literals somewhere.
- `tests/professions_acquisition_salvage_sink.test.ts:141`: the yield-scaling case shares one rng across both `salvageYield` calls, so the bonus draw differs between them and the `>=` is not a clean quality isolation.
- `src/sim/content/professions.ts` (`CRAFTING_HUB_STATIONS`): computed at module init and currently has zero consumers. Fine as forward content for the render pass, but keep the rounded offsets render-only (never feed them back into sim state), or drop the table until the consumer lands.
- `src/sim/professions/crafting.ts:304`: "every recipe currently in content/recipes.ts is common tier" is not true (TOOL_RECIPES sit at skillReq 75/150). Also worth an explicit nod from the maintainer that the sink charges common-tier crafts at all, given the free-floor rule this same diff quotes ("common tier never costs anything beyond materials").
- The PR description says `Entity.craftThrottle`; it is `PlayerMeta.craftThrottle`, session-only. Worth correcting so nobody goes looking for a wire field.
- `tests/world_api_parity.test.ts:402` and nearby: a few "198"/"52" strings in titles and comments were not bumped; the arrays and count pins themselves are correct.

jgyy:
Thanks for the thorough review, especially catching the archetype-pair issue: you're right, and I confirmed against the current text of #1129 that an archetype is an adjacent pair (the two majors), not a single craft. Pushed a fix (`2f7eb9b31`).

**Blocking, archetype pair model**

Reworked `archetype.ts`: `ArchetypeState` gains `pairedMajor`, the ring-adjacent craft empowered alongside `activeArchetype` (which craft becomes the pair's second half isn't yet a player choice, since the acceptance quest is still a content stub, so it defaults deterministically to the first ring neighbor). `archetypeCeilingFor`/`craftCeiling` now treat *both* majors as unlimited, so `meetsComboRequirement` passes for an attuned specialist's own adjacent pair (every `COMBO_RECIPES` entry is ring-adjacent, i.e. exactly this shape) and fails for a pair outside it, as intended. `battlefield_xp.ts`'s active-specialty gate checks both majors too (`observerPairedMajor`), and `getHobbyCraft`/`archetypeTitleFor` are unchanged (still keyed on the title-quest craft, per #1130).

While fixing this I also caught and fixed the two consequences you called out precisely:
- **The ceiling only capped the cap, never the climb.** `tierProgressMultiplier` read "recipe tier above the ceiling" as "at or above capability" (since `tiersBelow` went negative) and granted full progress, so a dormant/hobby craft kept leveling past its cap forever. Fixed: skill gain is now 0 whenever a recipe's tier exceeds the ceilinged capability.
- **Output quality ignored the ceiling entirely.** Added `clampMaterialRarity` (gathering.ts) and clamp the crafted-output roll to the empowerment ceiling, so a capped craft can still roll high off raw skill but the granted item never exceeds what it's empowered to produce.

`tests/archetype_ceiling.test.ts` (previously pinning the inverted single-craft behavior) is rewritten for the pair model, plus new coverage for: both majors unlimited, the freeze-at-ceiling behavior (including the common-tier free floor staying unaffected), and the paired-major battlefield-xp grant.

**Should fix**

- Added explicit `hudChrome.crafting.throttled` and `.recipeNotLearned` keys/branches in the HUD denial mapping (both English + the 5 non-Latin fills). Also renamed the "recipe exists but not learned" reason from `recipe_unknown` to `recipe_not_learned` so it can no longer be confused with the unrelated unresolvable-recipe-id `unknown_recipe` reason.
- The mobile action-ring tooltip removal (`4f4a5b615`): agreed it doesn't belong in this PR, but I wasn't able to revert it myself (blocked as an unrelated change to a commit outside this task's scope without explicit sign-off). Left it in for now — happy to split it into its own PR, or revert it here, whichever you'd prefer.
- `acquireRecipe`'s success/`wrong_source` coverage gap: I've scoped this out of this pass. Closing it properly means adding a real acquisition-gated recipe to `content/recipes.ts` (nothing there carries one yet), which felt like content scope creep for a review-response commit. Open to doing it as a follow-up if you'd like it in this PR instead.

**Nits**

- Pinned the gold-sink fee and throttle magnitudes to literals instead of recomputing them from the constants under test.
- Fixed the salvage-yield rarity test to use two identically-seeded rng instances instead of a shared rng's first/second draw.
- `CRAFTING_HUB_STATIONS`: documented as zero-consumer forward content, render-only offsets.
- Corrected the stale "every recipe is common tier" comment on the gold sink (TOOL_RECIPES sit at 75/150) and clarified the sink does charge common-tier crafts.
- The `world_api_parity.test.ts` comment-number nit and the PR-description `craftThrottle` field-name nit both look like they were already fixed by `b1b91efbf` before I got to them.

All of `tests/professions_*`, `archetype_ceiling.test.ts`, `world_api_parity.test.ts`, and the i18n gates pass, and `tsc` is clean.

EnriqueGF:
:) Thanks for the very thorough follow-up, and for the clean rework of the archetype pair model. I re-reviewed the current head (2f7eb9b31) against the real code rather than the description, ran the gates in a worktree, and dispatched the sim, parity, and test-coverage reviewers. Most of this is production-ready; a couple of items are still open before I can give the green light.

**Verified as resolved and solid**
- Base branch is release/v0.23.0 (the latest release), so the branch gate passes.
- The #1129 pair model is correct: both majors are unlimited via `pairedMajor`, `meetsComboRequirement` now composes `craftCeiling` so an attuned specialist's own adjacent pair passes and a pair outside it fails, and the two consequences from the last round are genuinely fixed. The tier-progress zero-guard (`recipeTier > 0 && recipeTier > capabilityTier => 0`) closes the negative-`tiersBelow` full-progress bug, and output quality is clamped via `clampMaterialRarity(rawQuality, archetypeCeilingFor(...))`. `archetype_ceiling.test.ts` pins both majors unlimited, the hobby-at-rare cap, the dormant-at-common cap, the freeze-at-ceiling, and the common free floor to literal tiers.
- Determinism and the SimContext seam are clean: the two new rng draws (crafting `rollMaterialRarity`, `salvageYield`) are strictly on the success path after every denial gate, so draw order is preserved. The 30+ parity golden edits are the benign new-serialized-field case (`craftThrottle`): zero `draws` lines and zero event-hash lines changed. No `sim.ts`/`hud.ts` growth beyond thin delegates.
- Parity: both new IWorld reads (`hobbyCraft`, extended `CraftResultView.reason`) are implemented in both `Sim` and `ClientWorld`, and the `craftResult` reason value travels the wire intact, so denial labels match offline and online. The HUD now maps `throttled` and `recipe_not_learned` to their own keys (no more silent fallthrough to `insufficientMaterials`).
- Persistence is additive and back-compat: `knownRecipes` round-trips as a JSONB array with an absent-loads-empty guard, `archetype` normalizes with a deterministic `pairedMajor` backfill, and `craftThrottle` is correctly session-only (not serialized). No DDL changes.
- tsc clean; architecture, localization_fixes, parity, snapshots, and all `professions_*` suites pass. No em/en dashes or emojis in the diff.

**Still open (should fix before merge)**

1. Unrelated mobile revert still bundled (commit `4f4a5b615`, `src/ui/hud.ts:~5205`). The action-ring long-press-to-inspect removal is out of scope for a professions batch and reverts another contributor's deliberate design (the removed comment documents that without it a long press casts and burns cooldowns while the player is trying to read a spell). It should get its own PR and its own call, as raised last round. Please pull it out here, or if you want it kept, it needs an explicit maintainer sign-off rather than riding along in this batch. The stale mirror comment near `hud.ts:~5206` ("peek-guard consume ...") would go with it.

2. `acquireRecipe`'s success and `wrong_source` arms are untested (`crafting.ts:148-152`). The "acquiring from the correct source" test (`professions_acquisition_salvage_sink.test.ts:74`) bypasses `acquireRecipe` entirely by mutating `meta.knownRecipes` directly, and no content recipe carries an `acquisition` list, so the exported `sim.acquireRecipe` command ships with its source-validation and mark-known side effect unprotected: a break in `acquisition?.includes(source)` or `knownRecipes.add` would not redden. Either register one acquisition-gated recipe (or add a test-only `recipeById` injection) and assert `acquireRecipe(id, 'drop') => wrong_source` and `acquireRecipe(id, 'trainer') => ok:true` + `knownRecipes.has(id)`, or get explicit sign-off to defer it to a follow-up.

**Nits (optional)**
- `professions_acquisition_salvage_sink.test.ts:154`: the "yield scales with rarity" assertion uses `toBeGreaterThanOrEqual`, so a regression that ignored quality entirely (`high === low`) would still pass. A strict `>` (or an exact expected pair) would make it decisive. The rng isolation itself is now correct.
- `professions_battlefield_xp.test.ts`: the paired-major grant has a positive case but no negative (`observerPairedMajor` set to a craft the recipe does not match), so over-granting on any set `pairedMajor` is not guarded.
- `CRAFTING_HUB_STATIONS` (`content/professions.ts`) is still zero-consumer forward content (the gate uses `CRAFTING_HUB_POS`/`RADIUS`/`MIN_LEVEL`). Fine as documented render-only offsets; just keep them out of sim state.

One note, not blocking: `hobbyCraft` and `archetypeTitle` are always null online (`ClientWorld` derives them from `activeArchetype`, which is never populated from the `prof` snapshot delta), so the char-window hobby line is offline-only. This matches the pre-existing `archetypeTitle` pending-wire-up status and is moot today since the acceptance quest is still retired content, but the online surface stays dark until `activeArchetype` is wired.

Great work overall; resolve items 1 and 2 (or get sign-off to defer them) and I am happy to approve.

EnriqueGF:
:) Following up on my last review. The head is still at 2f7eb9b31, unchanged since I re-reviewed the pair-model rework, so this is just to confirm status rather than repeat that pass. The base branch is still release/v0.23.0 (the latest release), so the branch gate remains green, and the archetype pair model, determinism, parity, persistence, and i18n all stand as verified before.

The two items I flagged last round are both still open on this head:

1. The unrelated mobile action-ring tooltip revert (commit 4f4a5b615, src/ui/hud.ts:~5205) is still bundled. It reverts another contributor's deliberate design and belongs in its own PR with its own call, or needs an explicit maintainer sign-off to ride along in this professions batch. The stale mirror comment near hud.ts:~5206 goes with it.

2. acquireRecipe's success and wrong_source arms are still untested. The "acquiring from the correct source" test (professions_acquisition_salvage_sink.test.ts:13373) mutates meta.knownRecipes.add(GATED_RECIPE.id) directly, bypassing sim.acquireRecipe, so the command's source-validation branch (acquisition?.includes(source)) and its mark-known side effect ship uncovered. Only the unknown_recipe and already_known denial arms are driven. Please either register one acquisition-gated recipe (or add a test-only recipeById injection) and assert acquireRecipe(id, 'drop') => wrong_source and acquireRecipe(id, 'trainer') => ok:true plus knownRecipes.has(id), or get explicit sign-off to defer it to a follow-up.

Everything else is production-ready. Resolve these two (or get sign-off to defer) and I am happy to approve. Thanks again for the strong rework.

FernandoX7:
Round 2 follow-up. I verified every item from my review against the new head, then took the branch the rest of the way so it is mergeable. Pushed three commits.

**Verified from round 1.** The pair rework (ArchetypeState.pairedMajor, both majors unlimited, combo gate and battlefield xp reading both), the HUD denial mapping with the new hudChrome.crafting keys, the recipe_not_learned rename, and the test nits all landed cleanly. The mobile action-ring tooltip item is moot: base picked up the same removal via #1617, and after merging, this branch's entire hud.ts footprint is the denial-mapping hunk.

**Merged release/v0.23.0** (the branch was conflicting again). Hand-resolved hud.ts and the IWorld pins (204 members now: base's lootRollGroupStatus plus this PR's hobbyCraft), regenerated the i18n artifacts and parity goldens.

**Fixed two things the round-2 rework introduced** (`8ee4aaaa5`):

- The new skill-gain freeze compared the recipe tier against min(archetype ceiling, raw capability). Crafting has no skillReq admission gate, so a recipe above raw capability is the ordinary climb, and base gave it full progress; the guard zeroed it everywhere. Concretely: an engineering major could never level engineering at all (all six engineering recipes are tier 3/6), which also left recipe_volatile_flux_elixir permanently uncraftable, and a hobby below its rare cap stopped climbing, contradicting the guard's own comment. It now compares against the archetype ceiling alone (the same ceilingTier the quality clamp reads); the climb arms and freeze arms are both pinned in tests/archetype_ceiling.test.ts.
- The default paired major was always the first ring neighbor, so attuning armorcrafting paired you with leatherworking and attuning alchemy with jewelcrafting, locking both out of their own themed combo (the partner craft caps at common) while the mirror-side attunements kept it. defaultPairedMajor now prefers the neighbor a content combo already commits the craft to, which matches the design doc's canonical adjacencies (armorcrafting with weaponcrafting, alchemy with engineering); crafts without a combo keep the first-neighbor default. Pinned with literals for all four combo crafts, plus a test that every content combo stays craftable after attuning to either of its sides. The future acceptance quest still turns this into a real player choice.

**Closed the remaining coverage gaps.** acquireRecipe's success and wrong_source arms now run through a new acquireRecipeForRecipe split (mirroring resolveCraftForRecipe) instead of mutating knownRecipes directly. The output-quality clamp has decisive tests (at skill 100 the roll can never be common, so a dormant craft granting common proves the clamp on any draw; majors are never clamped; unit arms for clampMaterialRarity). normalizeArchetypeState's pre-pair backfill, a non-adjacent saved pair, a preserved quest-chosen neighbor, and a pairedMajor serialize round trip are pinned. The station-bound negative arm iterates both full recipe lists. BATTLEFIELD_XP_TRICKLE is pinned at 0.25, the throttle test asserts denied crafts charge no copper, the salvage test pins the material id and the exact rarity delta, and hobbyCraftText got the view-model cases its comment claimed.

**Housekeeping.** Corrected the stale scope comments in crafting.ts and recipes.ts (still claiming common-tier-only and a pending skillReq gate), the archetype test header (still describing the single-craft model), one leftover "pinned 53" test title, and an inherited unsorted import pair in sim.ts that reds biome-on-changed-files. Moved the screenshot under docs/screenshots and repointed the link; also fixed the body's Entity.craftThrottle to PlayerMeta.craftThrottle (it is session-only, never on the wire).

`npm run gate` is green at the new head: i18n gen and freshness, malware scan, biome on changed files, the full vitest suite, tsc, and all five builds.

FernandoX7 (approved):

PR #1641, opened 2026-07-08
fix(ui): clear the stale bag tooltip after a drag-drop (#1626)

### LINKED ISSUE #1626: Bug: Item tooltip shows wrong item when dragging an item on top of another
### Before you start

### What happened?

In the `bag`, when dragging one item on top of another, the tooltip shows the dragged item's description instead of the item the cursor lands on. Similar to #1485

### Steps to reproduce

1. Open the bag
2. Hover over an item to reveal the tooltip
3. Drag the item by clicking and holding it
4. Drop the item on top of another item in the bag and release it (it helps to wait 2 seconds before releasing the dragged item)
5. Observe the dragged item returning to it's slot in the bag. While the cursor is still over the item that the dragged item landed on, the wrong item tooltip shows , even when moving the cursor (as long as the cursor stays inside the item's icon bounds).

It's not easy to reproduce, but the image below is proof that it happens.

When I release the item, I expect the tooltip of the item below my cursor to appear

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on Linux Mint

The image shows "simple fishing pole" but should show "Conjured Oatcake"

Image

daxdax89:
## Problem

In the bag, dragging one item on top of another leaves the wrong tooltip showing: the dragged item's description stays pinned over the item the cursor landed on, even as the cursor moves within that item's icon bounds (issue #1626). The reporter noted it's "similar to #1485".

## Root cause

This is the bags twin of the hotbar bug we fixed in #1485/#1564. A drag that ends over another bag item is rejected (bag rows are drag *sources*, not drop targets), so no drop handler clears the tooltip. Because the cursor is already stationary inside the target row, no fresh `mouseenter` fires to re-resolve the tooltip; the browser re-shows the drag source's tooltip at `dragend`, and `mousemove` only repositions the box, so the dragged item's tooltip is left pinned.

## Fix

Add `hideTooltip()` to the bag row's `dragend` handler (the single drag-completion path for bag items, which use HTML5 `draggable` only, no touch-drag path). This matches the drop-completion `hideTooltip()` the hotbar fix added and every sibling slot mutation already does. The next hover then resolves the item actually under the cursor.

## Tests

`tests/bags_drop_tooltip.test.ts` source-scans the `dragend` handler and asserts it calls `hideTooltip()` after clearing the drag state, mirroring the #1485 guard (`tests/hotbar_drop_tooltip.test.ts`).

## Note on screenshots

No static visual change: the fix removes a transient stale tooltip, so a before/after screenshot can't capture it (the bug is a pinned tooltip that shouldn't be there). The regression is guarded by the source-scan test.

Fixes #1626

### DIFF
--- a/src/ui/bags_window.ts
+++ b/src/ui/bags_window.ts
@@ -575,6 +575,13 @@ export class BagsWindow {
         row.addEventListener('dragend', () => {
           this.deps.setDragAction(null);
           this.deps.clearActionDropTargets();
+          // A drag that ends over another bag item is rejected (bag rows are not
+          // drop targets), so no drop handler clears the tooltip and no fresh
+          // mouseenter fires on the item under the now-stationary cursor: the
+          // dragged item's tooltip is left pinned (mousemove only repositions it).
+          // Clear it so the next hover resolves the item actually under the cursor
+          // (issue 1626, the bags twin of the hotbar fix 1485).
+          this.deps.hideTooltip();
         });
       }
       this.deps.attachTooltip(row, () => {
--- a/tests/bags_drop_tooltip.test.ts
+++ b/tests/bags_drop_tooltip.test.ts
@@ -1,0 +1,28 @@
+// Regression for a reported bug (#1626): dragging a bag item on top of another
+// left the tooltip stale. A drag that ends over another bag item is rejected (bag
+// rows are not drop targets), so no drop handler clears the tooltip and no fresh
+// mouseenter fires on the item under the now-stationary cursor; the dragged item's
+// tooltip stayed pinned (mousemove only repositions it). This is the bags twin of
+// the hotbar fix (#1485), which added hideTooltip() to the drop-completion paths.
+// Guard that the bag row's dragend handler now clears the tooltip after teardown.
+
+import { readFileSync } from 'node:fs';
+import { describe, expect, it } from 'vitest';
+
+const src = readFileSync(new URL('../src/ui/bags_window.ts', import.meta.url), 'utf8');
+// Strip comments so the explanatory comment near the fix cannot satisfy the scan.
+const code = src.replace(/\/\*[\s\S]*?\*\//g, '').replace(/(^|[^:])\/\/.*$/gm, '$1');
+
+describe('bag drag-drop clears the stale tooltip (#1626)', () => {
+  it('the bag row dragend handler calls hideTooltip after clearing the drag state', () => {
+    // There is exactly one dragend handler on a bag row; isolate it up to the
+    // attachTooltip wiring that immediately follows the draggable block.
+    const start = code.indexOf("addEventListener('dragend'");
+    expect(start).toBeGreaterThan(-1);
+    const handler = code.slice(start, code.indexOf('attachTooltip', start));
+    const clearIdx = handler.indexOf('this.deps.clearActionDropTargets();');
+    const hideIdx = handler.indexOf('this.deps.hideTooltip();');
+    expect(clearIdx).toBeGreaterThan(-1);
+    expect(hideIdx).toBeGreaterThan(clearIdx);
+  });
+});


EnriqueGF:
Thanks for taking this one on, and for cross-referencing it against #1485/#1564 :)

Verdict: **changes requested**. The base branch gate passes (`release/v0.23.0` is the latest `release/vX.Y.0`), `npx tsc --noEmit` is clean, and `tests/bags_drop_tooltip.test.ts`, `tests/hotbar_drop_tooltip.test.ts`, `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` all pass on the head commit. My reservation is about whether this actually fixes #1626.

### 1. Blocking: the stated root cause does not hold, and the added line looks inert

The PR body says "the browser re-shows the drag source's tooltip at `dragend`". I cannot find any code path that does that, and a real browser trace says it does not happen.

- `src/ui/bags_window.ts:573` already calls `this.deps.hideTooltip()` inside the bag row's `dragstart` handler. The tooltip is therefore hidden before the drag ever reaches another row.
- `attachTooltip` (`src/ui/hud.ts:3789-3850`) is the only thing that shows the box, and it only calls `showAt` from `mouseenter`, `focusin` and the touch long-press timer. `mousemove` just repositions a box whose `display` is already `none`. Nothing re-shows it at `dragend`.

I reproduced the exact scenario from the issue in headless Chromium 150 (`chrome-headless-shell`, native drag with `Input.setInterceptDrags` disabled): hover source, press, drag onto the target, hold, release with the cursor stationary inside the target, then nudge the cursor within the target's bounds. Results across the four combinations:

- source hides on `dragstart` (the current bag code), no `dragend` hide: tooltip is already hidden at release. No bug.
- source hides on `dragstart` plus the `dragend` hide from this PR: identical, tooltip already hidden. The new line is a no-op.
- source does not hide on `dragstart` (the hotbar shape before #1485), no `dragend` hide: the source's tooltip stays pinned after release. Bug reproduces, which validates the harness.
- source does not hide on `dragstart`, with a `dragend` hide: bug goes away.

So the fix is correct for a row that does not hide on `dragstart`, and the bag row already does. As written, this change cannot be shown to affect #1626 on Chromium. Please identify the real re-show path before landing (my two candidates: a `focusin` delivered on the source after `dragend`, which this fix would not catch because it lands later; or a non-Chromium browser that does not flush the deferred `mouseleave`/`mouseenter` after a drag).

### 2. Scope gap: only consumables are draggable rows at all

`src/ui/bags_window.ts:567` binds `draggable`, `dragstart` and `dragend` only when `this.deps.isHotbarItemId(s.itemId)`, and `isHotbarItemId` (`src/ui/hud.ts:4375`) is true only for `food`, `drink`, `potion` and fishing-use items. Gear, quest items and junk are not draggable, and their icon is `` (`src/ui/hud.ts:3740`), so no drag starts for them at all. If the reporter dragged a non-consumable, nothing in this diff runs on the reported path. Worth confirming which item type reproduces before deciding where the fix belongs.

### 3. The test cannot tell a working fix from an inert line

`tests/bags_drop_tooltip.test.ts` is a source scan that asserts `this.deps.hideTooltip();` appears after `this.deps.clearActionDropTargets();`. I confirmed it does fail if the line is deleted, so it is not a self-comparison pin, but it asserts the presence of one literal statement, not the behavior. It would pass unchanged even if the line has no effect (which is what point 1 suggests), and it would fail if the same behavior were later expressed through a shared helper. #1485 set the precedent, but there the causal chain was verifiable from source alone; here it is not, so the test is currently the only evidence of correctness and it does not carry that weight. A DOM-level or `scripts/*.mjs` browser assertion that drives the drag and reads the tooltip text would.

### 4. Evidence

Given there is no static visual change, a short screen recording of the repro before and after the patch would settle points 1 to 3 at once. Please attach one.

### Minor

The branch is mergeable but sits 153 commits behind `release/v0.23.0`. A rebase before merge is worthwhile.

### What I verified as green

Base branch, typecheck, the guard tests above, no `src/sim/` or wire or `IWorld` surface touched, no server or DB surface, no new player-visible strings so no i18n obligation, no em or en dashes and no emojis in the diff, and the change stays out of the monoliths. The mechanism is my only concern, but it is a load-bearing one.

daxdax89:
Withdrawing this one. On your analysis the fix is inert on Chromium (the bag row already hides the tooltip on `dragstart`, so the `dragend` hideTooltip is a no-op there), and I could not reproduce #1626 against the current bag code either. Rather than land a change that can't be shown to fix the issue, I'll close this and leave #1626 for a properly root-caused fix (a real drag repro that isolates the true re-show path). Thanks for catching it.

PR #1642, opened 2026-07-08
fix(ui): make the dev-tier nameplate glow a crisp outline, not a blur (#1639)

### LINKED ISSUE #1639: Overhead names with the dev-tier glow look blurred
## Problem

Some players' overhead names render with a soft halo that makes the text look blurred or out of focus. Other names on the same screen render crisp, so to players it reads as a rendering defect on specific nameplates rather than a deliberate effect.

The halo is the significant-contributor developer-tier name glow: names of high dev tiers get a colored text-shadow stack (0 0 4px plus 0 0 2px in the tier ring color) layered over the name color. The intent is a glowing outline, but at nameplate font size the shadow blurs the letterforms themselves instead of reading as a halo around sharp text.

## Steps to reproduce

1. Stand near a player with a significant-contributor dev tier (with developer badges enabled in settings, the default).
2. Compare that player's overhead name with a nearby regular player's name.
3. Observe the dev-tier name looks blurred while the regular name is crisp.

## Actual behavior

- Names carrying the dev-tier glow lose their edge definition and look blurry or out of focus.
- Regular player names next to them render sharp, which makes the glow read as a bug rather than flair.

## Expected behavior

- Every overhead name renders with crisp, legible letterforms.
- The dev-tier flair reads as an intentional outline or halo behind sharp text, not as blur over the letters.

## Scope

- The `.np-sig-dev` text-shadow stack in `src/styles/hud.css` (the glow radii).
- `devTierNameOutlineColor` in `src/ui/dev_tier.ts` and its application via `--dev-outline` in `src/render/nameplate_painter.ts`, in case the treatment needs a different mechanism (for example a sharp stroke plus a wider, fainter glow).

## Acceptance criteria

- Dev-tier names stay crisp: letterforms keep hard edges, with the tier color reading as a distinct outline or halo.
- The effect still visually distinguishes significant contributors and still composes with the Discord staff/default name color.
- Regular player nameplates are unchanged.
- The effect stays static (no animation), keeping it reduced-motion safe, and the existing developer-badge display toggle still hides it.

## Notes

- This is a legibility/tuning problem with the existing deliberate glow, not a projection or texture defect: nameplates do not scale with distance, so the fix is in the effect itself.

## Screenshot

Image

daxdax89:
## Problem

Some players' overhead names render with a soft halo that makes the text look blurred or out of focus, while regular names next to them stay crisp, so it reads as a rendering defect (issue #1639). The blurred names are carrying the significant-contributor dev-tier glow.

## Root cause

`.np-name.np-sig-dev` in `src/styles/hud.css` layered two zero-offset blurred text-shadows over the name:

```css
text-shadow:
  0 0 4px var(--dev-outline),
  0 0 2px var(--dev-outline),
  0 1px 1px #000;
```

At the 12px nameplate font size, a 4px blur radius is about a third of the cap height, so a zero-offset blur bleeds symmetrically over the letterforms and smears the strokes instead of reading as a halo behind sharp text. The base `.np-name` has no text-shadow, so regular names render crisp: hence the side-by-side "one blurry, one sharp" defect.

## Fix

Replace the all-blurred stack with a **crisp 1px ring** (eight zero-blur offset shadows listed first so they paint nearest the glyph and keep the letterform edges sharp), followed by a **single small faint `0 0 3px` blur behind the ring** for the halo flair, then the existing dark base drop. The ring is drawn in a **softened `--dev-ring`** (a `color-mix` that keeps the per-tier `--dev-outline` hue but lowers the alpha), so it reads as a subtle badge rather than a heavy bold stroke (per review), while staying crisp. The mechanism is unchanged: the per-tier color still arrives via `--dev-outline` (no hardcoded hue, no TS change), the effect stays static (reduced-motion safe), and the `showDevBadges` toggle still gates it.

## Tests

`tests/nameplate_dev_glow.test.ts` pins the crisp ring (>= 8 zero-blur offset layers in `--dev-ring`, all four cardinals + four diagonals) and that `--dev-ring` derives from `--dev-outline` (no hardcoded hue) and that at most a single small (<= 3px) blurred halo remains, so a revert to the two-layer smear reddens. `tests/css_value_validity.test.ts` and the `dev_tier` unit tests stay green.

## Before / after

Before/after of the dev-tier name (blurred smear vs crisp ring + halo), zoomed so the letterform edges are visible. Committed under `docs/screenshots/`.

dev-tier nameplate glow before/after

Fixes #1639


EnriqueGF:
:) Thanks for the tight, well-scoped fix for #1639, and for pinning it with a regression test.

I reviewed this against the real code in a worktree. Verdict: one thing to resolve before this is mergeable, and it is only about evidence, not the code.

Base-branch gate: PASS. This targets release/v0.23.0, which is the latest release branch.

What I verified and liked:
- CSS-only change, mechanism preserved. The per-tier color still arrives via `--dev-outline` from `nameplate_painter.ts` (no hardcoded hue, no TS change), the effect stays static so it is reduced-motion safe, and the `showDevBadges` gate is untouched. The eight zero-blur offset ring layers plus a single faint `0 0 3px` halo behind them is the right way to keep the 12px glyph edges crisp.
- Tests are decisive. `tests/nameplate_dev_glow.test.ts` reddens on a revert to the old two-layer smear: the halo assertion caps blurred `--dev-outline` layers at one and at 3px (the old stack had two, 4px and 2px), and the ring assertion requires at least four zero-blur offset layers (the old stack had zero). Both `nameplate_dev_glow` and `css_value_validity` pass locally (13/13).
- No em/en dashes or emojis in the diff. No sim, wire, server, DB, or i18n surface touched, so those invariants are not in play.

Blocking item:
1. Screenshots (visual change). The description says a before/after with a zoomed row is attached below, but no image is present in the PR body or comments. For a visual fix this is required. Please attach the before/after (ideally desktop and mobile nameplate) so the crisp-ring result can be confirmed by eye. Once that is up, this is good to go from my side.

daxdax89:
Screenshot added (committed under `docs/screenshots/`, rendered from the shipped `.np-sig-dev` `text-shadow`, zoomed so the glyph edges read on-page).

Before is the old two-layer blurred halo (the smear from #1639); after is the eight zero-blur offset-ring layers plus one faint 3px halo, so the outline is crisp instead of fuzzy:

dev-tier nameplate glow before/after

The effect stays static (reduced-motion safe) and the per-tier hue still comes from `--dev-outline` (no hardcoded color). That closes the blocking item. Thanks!

FernandoX7 (commented):
Re-reviewed after the screenshot commit and the retarget to release/v0.24.0. The blocking item from last round is resolved and the CSS change itself holds up; one should-fix on the new test, then this is good to merge.

Verified this round:

- The committed before/after (docs/screenshots/nameplate-dev-glow-before-after.png) is a clean 1640x700 PNG and matches the shipped rule: blurred smear before, crisp ring plus one faint halo after.
- Clean merge with release/v0.24.0; the base still carries the old two-layer smear, and nothing else competes with `.np-name.np-sig-dev` (no hud.mobile.css override, no fx-tier rule, no inline text-shadow in the painter).
- At the merged state: nameplate_dev_glow and css_value_validity pass 13/13, and the neighboring dev_tier, dev_tier_ui, nameplate_view, nameplate_declutter suites pass 48/48. Biome is clean on the changed files (two noNonNullAssertion warnings in the style of existing tests; CI does not fail on warnings).
- Mutation check: restoring the old `0 0 4px` / `0 0 2px` stack reddens the new test on both the ring and halo assertions, so the pin is decisive.
- Issue #1639 acceptance criteria all hold: regular nameplates untouched, the tier color still arrives via `--dev-outline`, the effect stays static, and the showDevBadges gate is unchanged.

Should-fix:

1. tests/nameplate_dev_glow.test.ts:36: the ring regex requires both offsets to carry `px`, so the four cardinal layers written with a unitless zero (`1px 0 0`, `0 1px 0`) are not counted; only the four diagonals match, which is exactly the `>= 4` floor, while the comment above describes an eight-layer ring. Accept a bare `0` as an offset, e.g. `/(?:-?\d+px|0)\s+(?:-?\d+px|0)\s+0\s+var\(--dev-outline\)/g`, and assert `>= 8` (or at minimum correct the comment). As written the assertion has zero margin, and a crisp cardinal-only variant would red for the wrong reason.

Nit:

2. Repo convention wants the screenshot referenced from the PR body; the embed currently lives in a comment and points at your fork branch, so it breaks once that branch is deleted after merge. Editing the body to embed the committed image is enough.

The fork CI runs were queued waiting on approval; I approved them, so checks should report shortly.

FernandoX7:
> Screenshot added (committed under `docs/screenshots/`, rendered from the shipped `.np-sig-dev` `text-shadow`, zoomed so the glyph edges read on-page).
> 
> Before is the old two-layer blurred halo (the smear from #1639); after is the eight zero-blur offset-ring layers plus one faint 3px halo, so the outline is crisp instead of fuzzy:
> 
> dev-tier nameplate glow before/after
> 
> The effect stays static (reduced-motion safe) and the per-tier hue still comes from `--dev-outline` (no hardcoded color). That closes the blocking item. Thanks!

I would have to see how this looks in game. The stroke on the text looks really heavy. Would it be possible to just remove it?

daxdax89:
Thanks for the thorough re-review. Both points addressed:

1. **Ring test (should-fix):** you're right, the regex required `px` on both offsets so it only counted the four diagonals against a `>= 4` floor. Fixed in 28b58c616: it now accepts a bare unitless `0` offset and asserts `>= 8`, so all four cardinals + four diagonals are pinned and a cardinal- or diagonal-only 4-layer variant reds for the right reason.
2. **Screenshot (nit):** moved the before/after embed into the PR body, pointing at the committed `docs/screenshots/nameplate-dev-glow-before-after.png`.

On the **stroke weight**: fair call, the eight 1px offset layers do read bolder than the old smear. Worth flagging the tension so you two can settle it: the crisp outline was the direction for #1639 (the *blur* was the reported defect, and EnriqueGF signed off on exactly this crisp look), and fully removing the outline would leave dev names visually identical to regular ones, dropping the significant-contributor badge entirely. If it's too heavy I can dial it back while keeping it crisp, e.g. a single thin 1px outline (4 cardinal layers only) or a lower-alpha `--dev-outline`, rather than remove it. Happy to push whichever you and EnriqueGF prefer, just let me know the target (keep as-is / lighter ring / remove).

FernandoX7:
Let's see how it looks in the dev environment later, then we can decide. Good call.

FernandoX7 (approved):

daxdax89:
Good call on the weight, I've softened it (ec06db16c). Rather than remove the outline (which would drop the significant-contributor badge and re-expose the #1639 blur), the ring is now drawn in a **softened `--dev-ring`**: a `color-mix(in srgb, var(--dev-outline) 60%, transparent)` that keeps the per-tier hue but lowers the alpha. So it stays a crisp zero-blur ring (edges still sharp, blur still fixed) but reads as a subtle badge instead of a heavy bold stroke.

`color-mix` is within the browserslist floor (Chrome 120 / Safari 17.2) and already used across the stylesheets.

Updated:
- `hud.css`: the eight offset layers + the halo now use `var(--dev-ring)`.
- Test: asserts `--dev-ring` derives from `--dev-outline` (no hardcoded hue) plus the eight crisp layers; the halo/smear guard matches either token so a revert to the old `--dev-outline` smear still reds (verified: reverting reddens the ring + halo assertions).
- Re-rendered the committed before/after to the softened ring; it's embedded in the PR body now.

If 60% still feels a touch heavy (or too faint) I can nudge the alpha, just say the number. Thanks for pushing on this.

PR #1643, opened 2026-07-08
feat(worldboss): Thunzharr tuning + loot fixes

ryan-foo:
## Summary

A world-boss tuning and fixes pass for Thunzharr, the Waking Peak, aimed at the v0.24.0 PTR window. Two halves:

### Tuning (src/sim/content/zone3.ts, data only)

- **No more kiting on foot.** Move speed doubles from 5.8 to 11.6, past a player's base run speed of 7. The Howling Gale snare stays as the second layer against sprint cooldowns and speed buffs; the comments describing the old 5.8-vs-7 chase math are updated to match.
- **Less spell spam.** Mechanic cadences stretch by roughly 1.5x: Thunderclap 8s to 12s, Seismic Stomp 16s to 24s, Mountainhide 18s to 27s, Stormcall 25s to 40s. Damage, radii, and durations are otherwise unchanged.
- **Shorter stun.** The Seismic Stomp stun drops from 3s to 1.5s: at 11.6 move speed the boss only needs a moment to close on anyone caught mid-flight, and the raid is not benched for whole GCDs.
- **No more wedging on colliders.** Live reports showed him getting caught on props while pathfinding, unable to go directly at an enemy. A new `phasesThroughObstacles` template flag forces `Sim.moveToward`'s existing phasing mode (straight line through props, the waterline, and the steep-wall gate) for every step he takes; ordinary mobs still collide. Tests pin the straight line through a building collider, the ordinary-mob control, and the template opt-in.
- **Quieter.** His zone-wide battle cries slow from every 9s to once a minute (the 350yd loud range and the one-time engage/summon/enrage barks and Stormcall telegraph yell are kept).

### Loot fixes (folds in #1513)

Fixes the cluster of Thunzharr loot problems reported from the live realm. The contributor roster used to be derived from the boss's live hate table, and `handleDeath` prunes the dead off threat, so a raider who died to the boss lost their personal drop; the corpse also despawned after 300s, sometimes before dead players could run back.

- A persistent `bossDamagers` set on `Entity` records every player (pet damage credits the owner) who lands a hit on a world boss; it is never pruned on death, spirit release, or threat drop.
- World-boss loot rolls from `worldBossLootContributors()` (the damager set unioned with the remaining hate table, e.g. a healer), sorted by `entityId` so the rng draw order stays deterministic.
- The lootable-corpse window extends 300s to 900s (`WORLD_BOSS_CORPSE_SECONDS`), still well inside the spawn cadence.
- A test pins that world-boss loot is personal (`personalFor` slots), never routed through a party need/greed roll.

The damager set is sim-internal and never wired to clients; loot resolves server-side. It is initialized in both the offline `Sim` (`baseEntity`) and the online `ClientWorld` (`blankEntity`).

Note on the base: release/v0.24.0 is not cut yet, so this targets the release/v0.23.0 tip; happy to retarget once the v0.24.0 branch exists.

## Related issues

Supersedes #1513 (folded in unchanged as its own commit). Addresses the reported Thunzharr issues: corpse despawning before everyone loots, loot rights lost on death, and the request that anyone who hit the boss can loot.

## Type of change

- [x] Feature: new functionality (balance retune)
- [x] Bug fix (loot roster and corpse window)

## How was this tested?

- Commands:
  - npx vitest run tests/world_boss.test.ts tests/world_boss_cc.test.ts tests/dead_party_loot.test.ts tests/loot_roll.test.ts (54 passed)
  - npx vitest run tests/parity tests/architecture.test.ts (188 passed, 1 skipped: rng draw-order digest unchanged)
  - The pre-push floor (tsc, guard tests, changed-files biome, copy scan) ran clean on push.
- New/updated tests:
  - A new pin that the boss's move speed exceeds a player's base run speed.
  - The Stormcall test's tick budget widened to cover the longer 40s cadence.
  - From #1513: a contributor who died to the boss still lands on the loot roster and receives their personal slot; the corpse lingers for `WORLD_BOSS_CORPSE_SECONDS`; a party's world-boss slots never enter a group roll.

### DIFF
--- a/src/net/online.ts
+++ b/src/net/online.ts
@@ -866,6 +866,7 @@ function blankEntity(id: number): Entity {
     enraged: false,
     healedThisPull: false,
     threat: new Map(),
+    bossDamagers: new Set(),
     forcedTargetId: null,
     forcedTargetTimer: 0,
     ownerId: null,
--- a/src/sim/combat/damage.ts
+++ b/src/sim/combat/damage.ts
@@ -45,7 +45,7 @@ import {
   virtualLevel,
   xpForLevel,
 } from '../types';
-import { WORLD_BOSS_CORPSE_SECONDS, worldBossContributors } from '../world_boss';
+import { WORLD_BOSS_CORPSE_SECONDS, worldBossLootContributors } from '../world_boss';
 
 // How long a slain mob's corpse persists (seconds) before it is cleared. Sole user
 // is handleDeath, so the constant lives here with the death-domain code.
@@ -340,6 +340,16 @@ export function dealDamage(
     else if (source.ownerId !== null) target.tappedById = source.ownerId;
   }
 
+  // World-boss loot roster: every player (or pet owner) who lands a hit on a world
+  // boss becomes a permanent loot contributor. Unlike the hate table above, this set
+  // is NEVER pruned when they die, release their spirit, or drop off threat, so a
+  // raider who died to the boss still gets their personal drop. Read at death by
+  // worldBossLootContributors. Only world-boss templates ever populate it.
+  if (source && amount > 0 && MOBS[target.templateId]?.worldBoss) {
+    const contributorId = source.kind === 'player' ? source.id : source.ownerId;
+    if (contributorId !== null) target.bossDamagers.add(contributorId);
+  }
+
   if (source && source.kind === 'player' && source.id !== target.id) {
     const meta = ctx.players.get(source.id);
     if (meta) meta.counters.damageDealt += amount;
@@ -578,7 +588,7 @@ export function handleDeath(ctx: SimContext, e: Entity, killer: Entity | null): 
     // collapse with the boss: leaving them alive would harass looters for the
     // whole window, and a slain add's in-place respawn timer would revive it
     // mid-window (only fires for worldBoss templates, so no parity rng change).
-    const worldBossContribs = template?.worldBoss ? worldBossContributors(ctx, e) : null;
+    const worldBossContribs = template?.worldBoss ? worldBossLootContributors(ctx, e) : null;
     if (template?.worldBoss) {
       e.corpseTimer = WORLD_BOSS_CORPSE_SECONDS;
       e.respawnTimer = Infinity;
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -852,7 +852,10 @@ export const ZONE3_MOBS: Record<string, MobTemplate> = {
     stomp: {
       radius: 11,
       every: 24,
-      duration: 3,
+      // A short pin, not a shutdown: at 11.6 vs a base run of 7 the boss closes
+      // about 7yd in 1.5s, enough to be on top of anyone caught mid-flight
+      // without benching the raid for whole GCDs.
+      duration: 1.5,
       min: 18,
       max: 28,
       name: 'Seismic Stomp',
@@ -897,10 +900,10 @@ export const ZONE3_MOBS: Record<string, MobTemplate> = {
     },
     // Loud: a mountain-sized voice. Every yell (engage/summon/enrage + these battle
     // cries) carries 350yd, far past the 100yd default, and he bellows one of these
-    // lines every 30s in combat so the whole of Thornpeak knows he is awake without
-    // the barks drowning out the raid's chat.
+    // lines once a minute in combat: the whole of Thornpeak still knows he is awake,
+    // but the barks never drown out the raid's chat.
     battleYells: {
-      every: 30,
+      every: 60,
       range: 350,
       lines: [
         'THUNDER ANSWERS! The peak has teeth again!',
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -836,20 +836,22 @@ export const ZONE3_MOBS: Record<string, MobTemplate> = {
     dmgPerLevel: 10.3,
     attackSpeed: 2.4,
     armorPerLevel: 46,
-    moveSpeed: 5.8,
+    // Faster than a player's base run speed (7, entity.ts): Thunzharr cannot be
+    // outrun on foot, so there is no kite even before the Howling Gale snare lands.
+    moveSpeed: 11.6,
     aggroRadius: 18,
     aoePulse: {
       min: 36,
       max: 50,
       radius: 12,
-      every: 8,
+      every: 12,
       name: 'Thunderclap',
       school: 'nature',
       fx: 'nova',
     },
     stomp: {
       radius: 11,
-      every: 16,
+      every: 24,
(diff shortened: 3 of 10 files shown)


jgyy (commented):
Went through both halves of this. The tuning is a clean data-only change (`src/sim/content/zone3.ts`), and the loot fix is the right shape: a persistent `bossDamagers: Set<number>` that never gets pruned by `handleDeath`, correctly kept off the wire, correctly initialized in both `Sim` (`entity.ts`) and `ClientWorld` (`online.ts`) so the two hosts stay parity-safe. `worldBossLootContributors` unions the damager set with whoever is still on the live hate table and sorts by `entityId`, so the rng draw order for `rollWorldBossLoot` stays deterministic.

Ran the suite against the PR head to check the claims in the description:

```
npx vitest run tests/world_boss.test.ts tests/world_boss_cc.test.ts tests/dead_party_loot.test.ts tests/loot_roll.test.ts
# 54 passed

npx vitest run tests/parity tests/architecture.test.ts
# 188 passed, 1 skipped
```

Both matched. The new tests are genuinely decisive: the death-roster test forces `boss.threat.has(p2)` false and `boss.bossDamagers.has(p2)` true in the same assertion block, which is exactly the regression this fixes, and the party test checks `personalFor.length === 1` plus `pendingLootRolls.size === 0` after loot, which pins "never a group roll" rather than just "loot exists."

A few things I checked and confirmed clean, not issues, just noting them:
- `bossDamagers` writes are gated on `MOBS[target.templateId]?.worldBoss`, so this never touches normal mob tap/threat behavior (`damage.ts:347`).
- `yells` (one-time engage/summon/enrage) and `battleYells` (the periodic cries) are separate fields; slowing `battleYells` to 60s does not touch the engage/enrage/Stormcall-telegraph lines, matching the PR description.
- `WORLD_BOSS_CORPSE_SECONDS` at 900 is still well under the 3600s spawn interval, so corpse windows can't overlap, same invariant as before, just a wider margin.
- #1513 is closed on GitHub and its commit is folded in unchanged, so no duplicate-PR risk there.
- Merge-tree against `release/v0.23.0` is clean, no conflicts.

Nothing blocking. One nit: `moveSpeed: 11.6` is exactly 2x the old 5.8, and it's now well above the player base run speed of 7, comments call this out clearly so it reads intentional rather than a stray typo, just flagging that a top-end mob speed buff or slow-immunity trinket (if one exists or gets added later) is the only way this fight could still be kited, worth a mention in a future pass if that ever becomes a data point, not this one.

Nice work tightening the fight cadence and fixing a real live-realm loot bug in the same pass without conflating the two (separate reviewable commits).

```mermaid
flowchart TD
    A[Player hits Thunzharr] --> B{target.templateId is worldBoss?}
    B -- yes --> C[mob.bossDamagers.add contributorId]
    B -- no --> Z[normal threat/tap logic only]
    C --> D[mob.threat updated too]
    D --> E{Player dies to boss?}
    E -- yes --> F[handleDeath prunes mob.threat\nbossDamagers UNTOUCHED]
    E -- no --> G[stays on both threat and bossDamagers]
    F --> H[Boss killed]
    G --> H
    H --> I[worldBossLootContributors:\nbossDamagers UNION live threat table\nsorted by entityId]
    I --> J[rollWorldBossLoot: personal slot per contributor]
    J --> K[Corpse lootable for WORLD_BOSS_CORPSE_SECONDS = 900s]
    K --> L[Dead contributor resurrects, runs back, loots\npersonal slot still there]
```

EnriqueGF:
**Review: world-boss tuning + loot fixes (Thunzharr)**

Overall this is a clean, well-tested change and the loot-roster design (a permanent `bossDamagers` set unioned with the live hate table, deterministic sort by `entityId`) is the right shape. Verified: tri-host init parity (`baseEntity` + `blankEntity`), `createMob` picks up the fresh Set via `baseEntity`, the set is sim-internal and never wired, loot resolves server-side, pet damage credits the owner, healers still land on the roster via the hate table, the hot-path guard short-circuits on `amount > 0` before the `MOBS` lookup, and no em/en dashes or emojis. `tests/world_boss.test.ts` passes 30/30 locally.

One blocker, then some minor notes.

**Blocking: `bossDamagers` is not cleared on a full evade/wipe reset.**
- `resetEvadingMob` (`src/sim/mob/locomotion.ts:592`) is the path that runs when a world boss is wiped and leashes home to full HP: it resets `tappedById` and calls `clearThreat`, but leaves `bossDamagers` populated. `respawnMob` (`src/sim/mob/lifecycle.ts:70`) has the same gap. Because a world boss keeps the same entity across an evade (its `respawnTimer` is `Infinity`, so only the scheduler makes a brand-new entity, and only on a real kill), the damager roster survives a failed pull.
- Failure scenario: raid A pulls Thunzharr, wipes, the boss evades home and resets to full. Raid B pulls the same entity later and kills it. Every raider from pull A is still in `bossDamagers`, so `worldBossLootContributors` generates `personalFor` slots for them at the kill. This contradicts the roster's own documented invariant ("every player who hit the boss since it was pulled") and the classic tap semantic that a full reset clears loot rights. Practical harm is limited (personal loot to absent or locked-out players just goes unlooted), but a pull-A player who is still online and not locked out gets a free personal drop from a kill they did not participate in.
- Fix is a one-liner: clear `mob.bossDamagers` alongside `clearThreat(mob)` / `tappedById = null` in both `resetEvadingMob` and `respawnMob`. Per the repo bar (sim changes get a test), please add a test that pulls the boss, forces an evade-home reset, confirms `bossDamagers` is empty afterward, and confirms a pull-A-only damager does not appear on the roster of a later kill.

**Minor (non-blocking):**
- The corpse-window comments now say the window "never overlaps the spawn cadence." That is a load-bearing claim (900s corpse vs 3600s interval). It holds today, but consider a cheap static assertion or test pinning `WORLD_BOSS_CORPSE_SECONDS < WORLD_BOSS_INTERVAL_SECONDS` so a future interval change cannot silently create overlapping lootable corpses.
- Base branch: targeting `release/v0.23.0` is correct given v0.24.0 is not cut yet; just retarget when it exists, as you noted.

Happy to re-approve once the evade/respawn reset clears the damager set with a covering test. Nice work on the determinism and parity discipline here.

ryan-foo:
Blocker addressed in 03de601e8: both reset paths (resetEvadingMob and respawnMob) now clear mob.bossDamagers alongside clearThreat, with covering tests: the full wipe scenario (pull A wipes and the boss evades home, pull B kills, the pull-A raider gets no personal slot), the respawn clear, and a static pin that WORLD_BOSS_CORPSE_SECONDS < WORLD_BOSS_INTERVAL_SECONDS so the corpse window can never overlap the spawn cadence (the minor note). World boss suite is 42 passing, parity and architecture guards green.

jgyy:
the boss fix ok for me

PR #1645, opened 2026-07-08
fix(itemization): fire legendary weapon procs on a hunter's Auto Shot

ryan-foo:
## Summary

On `release/v0.23.0` a hunter's Auto Shot already fires with the equipped mainhand (the ranged-weapon-scaling change), but on-hit weapon procs still only rolled on a melee swing. So Thronebane's Chain Arc never fired for a hunter attacking at range, which is where hunters spend their whole rotation.

This renames the `WeaponProcTrigger` `meleeHit` -> `weaponHit` (it always meant "on a weapon strike") and rolls it from `rangedSwing` too, gated to non-wand shooters so caster wand bolts (which do not swing the mainhand) never roll the mainhand's proc.

Determinism is preserved: the proc still draws no rng unless the shooter carries a matching proc weapon, so the shared draw order and every parity golden that equips no legendary are unchanged (parity gate green).

The tooltip label key (`onMeleeHit`) already reads the generic "Chance on hit", so no i18n copy change is needed and no new key is added; it is reused as-is.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/equip_procs.test.ts tests/weapon_proc_view.test.ts tests/combat_auto_attack.test.ts tests/ranged_shot.test.ts tests/architecture.test.ts tests/parity/` (233 passing, 1 skipped). `npx tsc --noEmit` clean on the changed files (the only local error is a pre-existing missing `prom-client` dep in `server/http/metrics.ts`, untouched here).
- Added an integration test: a Thronebane hunter procs Chain Arc off Auto Shot, and a wand caster with the same mainhand does not.

N/A: no UI/player-visible copy change (label reused), so no screenshots.

### DIFF
--- a/src/sim/combat/auto_attack.ts
+++ b/src/sim/combat/auto_attack.ts
@@ -223,6 +223,11 @@ export function rangedSwing(
     // 4-piece set procs keyed to weapon crits (ranged arm). Gated on setProcs
     // inside applySetProcs, so proc-less players draw no rng.
     if (crit && atk.kind === 'player') ctx.applySetProcs(atk, tgt, 'weaponCrit');
+    // Legendary on-hit weapon procs (e.g. Thronebane's Chain Arc) fire on a hunter's
+    // Auto Shot too, since it strikes with the equipped mainhand. Wand bolts do not
+    // swing the mainhand, so casters never roll it. No-op (no rng draw) unless the
+    // shooter wields a proc weapon with a weaponHit proc.
+    if (!ranged.wand) runWeaponProcs(ctx, atk, tgt, 'weaponHit');
   });
 }
 
@@ -334,7 +339,7 @@ export function meleeSwing(
     }
   }
   // Legendary on-hit weapon procs (e.g. Thronebane's Chain Arc). No-op (no rng
-  // draw) unless the attacker wields a proc weapon with a meleeHit proc.
-  runWeaponProcs(ctx, attacker, target, 'meleeHit');
+  // draw) unless the attacker wields a proc weapon with a weaponHit proc.
+  runWeaponProcs(ctx, attacker, target, 'weaponHit');
   return true;
 }
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -2814,7 +2814,9 @@ export const ZONE3_ITEMS: Record<string, ItemDef> = {
       {
         id: 'thronebane_arc',
         name: 'Chain Arc',
-        trigger: 'meleeHit',
+        // Fires on any weapon strike: a melee swing for its warrior/paladin owners,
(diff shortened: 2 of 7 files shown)


jgyy (commented):
Good fix, and a clean one. `weaponHit` is the right name for the trigger now that it covers both swing paths, and gating the new roll on `!ranged.wand` is exactly right: wand bolts don't swing the mainhand, so a caster holding Thronebane correctly never rolls it. I checked `runWeaponProcs` (`src/sim/combat/equip_procs.ts`) and the no-proc-weapon / no-mainhand / dead-target / trigger-mismatch guards all still short-circuit before any `rng.chance` draw, so the shared draw order for every proc-less parity golden is untouched.

Ran the suite locally against the PR head (`f39d6560e`), all green: `equip_procs`, `weapon_proc_view`, `combat_auto_attack`, `ranged_shot`, `architecture` (68 passing) and `tests/parity/` (164 passing, 1 skipped). Merge-tree against `release/v0.23.0` is clean, no conflicts.

A couple of small things, neither blocking:

- nit: `src/ui/hud.ts:4058` still checks `line.trigger === 'weaponHit'` to select the `hudChrome.itemProc.onMeleeHit` key. Correct behavior, but the key id itself is now a bit misleading given it also renders for a ranged proc. Not asking for an i18n change (that's a bigger lift for a one-line rename), just flagging so it doesn't get "fixed" independently and drift.
- nit: the new `rangedSwing` test only exercises a hunter Auto Shot and a wand caster. A ranged mob attacker (if any exist with `weaponProcs`-bearing mainhands) isn't covered, but that's a hypothetical today, not a real gap.

The rename is complete: the only other `meleeHit`/`meleeHit(` hit in the tree is the unrelated `AudioManager.meleeHit()` sound-effect method in `src/game/audio.ts`, not the `WeaponProcTrigger` type.

Here's the shape of the change:

```mermaid
flowchart TD
    A[Player swings] --> B{Attack kind}
    B -->|Melee swing| C[meleeSwing]
    B -->|Ranged shot lands| D[rangedSwing]
    D --> E{ranged.wand?}
    E -->|true, caster bolt| F[skip: no mainhand swing]
    E -->|false, hunter Auto Shot| G[runWeaponProcs trigger=weaponHit]
    C --> G
    G --> H{wielder.mainhandItemId has weaponProcs?}
    H -->|no| I[no rng draw, no-op]
    H -->|yes| J[ctx.rng.chance proc.chance]
    J -->|fail| I
    J -->|success| K[fireEffect: e.g. Chain Arc]
```

Before this PR, the `D -> G` edge didn't exist, so Thronebane's Chain Arc only ever fired off the `C -> G` melee path, leaving hunters (who live at range) unable to proc their own legendary.

jgyy:
lgtm minor change

PR #1648, opened 2026-07-08
feat(sim): queue one ability behind the current cast (#1360)

### LINKED ISSUE #1360: Queue one ability behind the current cast
## Problem

Pressing an ability while a cast is in progress is dropped ("You are busy."), so there is no way to pre-queue the next spell to fire the moment the current cast finishes. Classic clients provide a small spell-queue window near the end of a cast so the next ability chains without a gap.

## Expected behavior

- Pressing an ability during the tail of the current cast (a small queue window) fires it automatically when the current cast completes.
- Single-slot queue: pressing another ability overwrites the queued one.
- Outside the queue window, the press still errors as today.

## Scope

- `src/sim/combat/casting_lifecycle.ts`: the busy/GCD guard in `castAbility` currently drops the press; set a single-slot queue instead when within the window, and consume it on cast finish in `updateCasting`.
- A new single-slot queue field on `Entity` (note `queuedOnSwing` is a separate on-melee-swing mechanic, not a cast queue).
- Determinism: this adds a new rng-drawing path, so it needs a `tests/parity` scenario and an architecture-reviewer pass; keep the new work inside the existing casting tick phase (do not reorder phases).
- Optional: surface the queued cast to the HUD via `IWorld` if it should be shown.

## Acceptance criteria

- An ability queued within the window fires on cast completion.
- Only one ability queues at a time (a later press replaces it).
- A press outside the window still errors.

jgyy:
## Summary

- A press during the tail of the current cast (last 0.4s, `CAST_QUEUE_WINDOW_SEC`) queues into a single slot instead of erroring "You are busy.", and fires automatically the instant the current cast completes.
- The queued press is re-validated fresh (target/range/cost/GCD/cooldown) exactly at the moment it fires, the same as a same-tick player press.
- Single-slot queue: a later press within the window overwrites the earlier one.
- A press outside the window still errors as today.
- An interrupted cast (`cancelCast`, e.g. silence/stun/lockout) drops the queued press instead of firing it, since the cast never completed.

## Related issues

Closes #1360

## Flow

```mermaid
sequenceDiagram
    participant Player
    participant castAbility
    participant Entity
    participant updateCasting

    Player->>castAbility: press Fireball (cast in progress)
    castAbility->>Entity: castRemaining <= 0.4s?
    alt within queue window
        castAbility->>Entity: queuedCastAbility = 'fireball'
        Note over Entity: no error, single slot (later press overwrites)
    else outside window
        castAbility-->>Player: error "You are busy."
    end

    loop each sim tick
        updateCasting->>Entity: castRemaining -= DT
    end

    updateCasting->>Entity: castRemaining <= EPS (cast completes)
    updateCasting->>Entity: emit castStop (success)
    updateCasting->>Entity: fireQueuedCast()
    alt queuedCastAbility set
        Entity->>castAbility: castAbility(queued, aim)
        Note over castAbility: fresh validation (target/range/cost/GCD/cooldown)
    end
```

## Type of change

- [x] Tests

## How was this tested?

- Commands:
  - `npx tsc --noEmit`
  - `npx vitest run tests/combat_casting_lifecycle.test.ts` (new queue-window/fire/overwrite/interrupt tests, 17 passed)
  - `npx vitest run tests/parity/parity.test.ts` (98 passed; only `c4a_casting_lifecycle` golden regenerated, for the new queue segment added to that scenario)
  - `npx vitest run tests/nythraxis_encounter.test.ts tests/nythraxis_raid.test.ts tests/nythraxis_matrix.test.ts tests/architecture.test.ts`
  - `npx @biomejs/biome ci --changed` (exit 0, warnings only, no errors/format diffs)
  - `npm run build`
  - An independent `architecture-reviewer` pass over the diff: confirmed no rng draw-order regression, no SimContext seam violation, no re-entrancy issue, and no purity/determinism violation. It found one real gap (a Nythraxis ward-channel force-clear path that bypassed `cancelCast` and didn't drop a queued press); fixed in `clearNythraxisWardChannelCast`.
- Manual steps: N/A, this is a sim-only mechanic with no new UI surface (the HUD indicator called out as optional in the issue is not included in this PR).

## Screenshots / recordings

N/A: sim-only cast/queue-resolution logic, no visible UI change (see the sequence diagram above instead).

---

### Quality

- [x] **Tests pass.** Added/updated tests for the `sim/` behavior changed.

### Cross-platform

- [x] N/A: server-authoritative sim logic; no client-only surface, no touch targets.
- [x] N/A: no UI added.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the existing "You are busy." error string is unchanged and untouched).

### Hygiene


EnriqueGF:
Reviewed the spell-queue change end to end (sim determinism, IWorld/tri-host parity, server authority, i18n, tests). The design is clean and the single-slot semantics match the PR description. One real correctness gap plus two minor notes.

**Blocking**

1. `src/sim/combat/casting_lifecycle.ts:147-150` (fishing completion) leaks the queued slot. The fishing branch `return`s before the `fireQueuedCast(ctx, p)` call added at line 156, and `completeFishing` never touches `queuedCastAbility`. A press in the last `CAST_QUEUE_WINDOW_SEC` of a fishing cast reaches the queue branch in `castAbility` (the `castingAbility` check at line 221 fires while `castingAbility === FISHING_CAST_ID`, and none of the earlier stun/silence/lockout guards block it), so it queues. When fishing completes, the queued ability is neither fired nor cleared, and it lingers on the entity to misfire on the next unrelated cast completion. This is the same class of stale-state misfire you correctly handled for `clearNythraxisWardChannelCast`. Fix: clear (or deliberately fire) the queue in the fishing branch before the early `return`, and add a test for 'queue armed during fishing tail does not survive fishing completion.'

**Minor**

2. Short casts silently drop the queued press. `fireQueuedCast` runs at cast completion, but the queued `castAbility` re-checks GCD (`if (!ability.offGcd && p.gcdRemaining > 0) return;`). For a cast whose total is under the GCD (about 1.5s), the GCD can still be running when the cast ends, so a press the player made inside the window is dropped with no feedback. Acceptable if intended, but worth a code comment so it is not read as a bug later.

3. The overwrite test (`keeps only a single queued slot`) queues `lesser_heal` twice and asserts `lesser_heal` both times, so it cannot distinguish 'overwrote' from 'kept the first press.' Queue two distinct abilities and assert the second to make the assertion decisive.

Everything else checks out: `queuedCastAbility`/`queuedCastAim` are correctly server-side-only (not added to the wire snapshot), initialized in both `entity.ts` and `net/online.ts` `blankEntity` for structural parity, cleared on death/arena/fiesta/spirit-release/revive, and re-validated fresh at fire time. No new rng draws at queue or fire, and the golden trace was regenerated consistently. Nice work on the interrupt-drops-queue path.

EnriqueGF:
We would have to test this ingame to see it fits in the gameplay

@Blaine1705

ryan-foo:
Took a pass through this. Clean, well-scoped implementation.

The core is small (queue in the cast tail in `castAbility`, fire from `updateCasting` on completion, drop on interrupt), and it handles the two things these changes usually miss:

- The new `queuedCastAbility`/`queuedCastAim` fields are added to `baseEntity` AND `blankEntity`, and kept correctly server-local (not half-wired), so the offline `Sim` and `ClientWorld` entity shapes stay in parity.
- The queue is cleared everywhere a cast is force-stopped: death (`damage.ts`), arena ready, fiesta down, nythraxis ward, and spirit release + revive. That is the easy-to-forget half and you got all of it.

Determinism holds: no rng added, `fireQueuedCast` is a no-op when the slot is empty, and `castingAbility` is nulled before it runs so the queued cast fires fresh instead of re-queuing. Parity gate is green, and the unit tests cover the window boundary, queue-then-fire, single-slot overwrite, and drop-on-interrupt.

Findings:

- **should-fix** `casting_lifecycle.ts:150`: the fishing branch `return`s before `fireQueuedCast`, and nothing clears the queue there (`completeFishing` does not touch it). A spell pressed in the last 0.4s of a fishing cast lingers in `queuedCastAbility` and fires on the next real cast completion, a surprise cast much later. Simplest fix: do not let a fishing cast accept a queue, gate the queue branch (`:241`) on `p.castingAbility !== FISHING_CAST_ID`, matching the existing fishing exemptions at `:103`/`:111`.

- **nit**: the queue fires on cast completion, not GCD completion, and `fireQueuedCast` runs the full `castAbility` gate set. So a follow-up queued behind a cast shorter than the GCD is silently dropped by the GCD gate (`:249`), and a queued ability that is on cooldown or unaffordable at completion still emits its error toast (`:255`) even though the press was 0.4s earlier. Both are arguable as classic-ish, but the silent sub-GCD drop is the one players will notice with hasted casts.

- **nit**: the `queuedCastAim` ground-target path has no test. The plumbing looks right (aim stored at queue time, passed through on fire), but a queued ground-targeted cast with an aim point is the one case the unit tests skip.

Nothing blocking. The fishing stranding is the only real correctness gap.

ryan-foo:
@EnriqueGF btw this feature is kinda already on retail, smooths out for those who aren't necessarily in the same server

https://maxroll.gg/wow/resources/spell-queue-window

As a designer I would say its good, just lets people have optimal DPS, increases access to the game

@jgyy fyi

jgyy:
Thanks both for the thorough review, @EnriqueGF and @ryan-foo.

Fixed the blocking gap: the fishing branch in `castAbility` (`src/sim/combat/casting_lifecycle.ts`) now excludes `FISHING_CAST_ID` from the queue window, matching the existing fishing exemptions for the silence/lockout guards a few lines up. A press during the tail of a fishing cast now falls through to the normal "You are busy." error instead of getting queued and stranded (since `completeFishing` never calls `fireQueuedCast`). Added a regression test: "drops a press queued in the tail of a fishing cast instead of stranding it."

Also fixed the overwrite test per your note, @EnriqueGF: it now queues two distinct abilities (`lesser_heal` then `smite`) and asserts the second, so it actually proves overwrite rather than "kept the first press."

Left the two nits as documented, non-blocking behavior for now:
- Sub-GCD queued casts still silently drop at the GCD gate when fired. Added a code comment at the GCD check explaining this so it's not mistaken for a bug later.
- The `queuedCastAim` ground-target path remains untested; happy to add a dedicated test if you'd like it in this PR rather than a follow-up.

Verified: `npx tsc --noEmit`, `npx vitest run tests/combat_casting_lifecycle.test.ts` (18 passed, up from 17), `npx vitest run tests/parity/parity.test.ts tests/architecture.test.ts` (all green, no golden regen needed since no new state was added), and `npx @biomejs/biome ci --changed` clean (warnings only, matching the original PR).

Pushed to the branch. Ready for another look.

EnriqueGF:
Thanks for turning the fixes around so quickly :)

I re-verified the branch in a clean worktree against `release/v0.23.0` (correct base, latest release branch). Confirmed fixed from the last round: the fishing tail no longer strands a queued press (`casting_lifecycle.ts:241`, plus the new regression test), and the overwrite test now queues `lesser_heal` then `smite` and asserts the second, so it is decisive. Gates I ran myself: `npx tsc --noEmit` clean, `tests/architecture.test.ts` + `tests/localization_fixes.test.ts` + `tests/parity` all green (193 passed), `tests/combat_casting_lifecycle.test.ts` green.

Two blocking findings, both confirmed by execution rather than by reading.

**Blocking**

1. `tests/parity/scenarios.ts:3049-3055`: the new "spell queue" parity segment never queues anything. The comment claims `2.5 - 43*0.05 = 0.35s remaining`, but the mage fireball at this level is a 3.0s cast, so after `rec.tick(43)` the entity has `castRemaining = 0.85`, which is outside the 0.4s window. I instrumented the scenario and printed the state at the press: `castRemaining=0.8500000000000025 castingAbility=fireball queued=null`. The press takes the `ctx.error(p.id, 'You are busy.')` branch, exactly as before this PR. So the 657-line golden churn comes only from the extra fireball cast and the 20 extra ticks, not from the queue, and the parity gate does not protect the new behavior at all. The PR body claim that the golden was regenerated "for the new queue segment" is not accurate. Please drain to inside the window (tick until `castRemaining <= CAST_QUEUE_WINDOW_SEC`, and assert it rather than hardcoding a tick count that drifts with cast time), then regenerate.

2. `src/sim/combat/casting_lifecycle.ts:249-255`: a queued press behind a hasted 1.5s cast is silently dropped, and that is the feature primary use case, not an edge case. The GCD is armed at cast START at a flat `playerGcdFor(cls)` (1.5s), while the cast itself is shortened by `spellHaste`. Every 1.5s-base cast (fireball, frostbolt, flash_heal, mind_blast, lightning_bolt, healing_wave, wrath, scorch, searing_pain, flash_of_light, slam) therefore completes while the GCD is still running, as soon as the player has any spell haste at all. Reproduced with a level 60 priest, `flash_heal`, `spellHaste = 0.1`: `castTotal = 1.3636`, and at the moment `fireQueuedCast` runs, `gcdRemaining = 0.1`, so `castAbility` returns at the GCD gate and the result is `castingAbility = null`, no cast, no error. The player pressed inside the window and got nothing. This is precisely what the retail queue window exists to smooth out, per the reference @ryan-foo linked. The code comment added at line 252 documents the behavior but does not make it acceptable. Suggested fix: hold the slot and fire once the cast has completed AND `gcdRemaining <= 0`, instead of firing exactly once at cast completion.

**Should fix**

3. None of the five force-stop clears this PR adds have a single assertion. `grep queuedCastAbility tests/` returns hits only in `tests/combat_casting_lifecycle.test.ts`, and none of them cover death (`damage.ts:519`), arena (`arena.ts:735`), fiesta (`fiesta.ts:298`), spirit release and revive (`spirit.ts:124,208`), or the nythraxis ward (`nythraxis.ts:988`). Those clears are the correctness half everyone praised, and right now a future refactor can delete any one of them with the suite still green. One assertion per path is cheap.

4. `src/sim/social/fiesta_bots.ts:153-155`: this PR silently changes existing bot behavior. The mage bot ("Botzo the Arcane") presses `pickBotAbility` every 24 ticks against a 3.0s fireball cast, so some presses land inside the 0.4s tail. Before this PR they hit the no-op "You are busy." branch; now they queue and fire on completion. Still deterministic across the three hosts, so not an invariant break, but it is an unadvertised, unpinned change to a live system. Please confirm it is intended and pin it, or gate the queue to human presses.

5. The `queuedCastAim` ground-target path still has no test, as you noted. Given that the aim point is the one piece of state carried across the queue boundary, it belongs in this PR.

**Nit**

6. A queued ability that is on cooldown, unaffordable, or out of range at fire time still emits its error toast, roughly 0.4s after the press. Arguable, but worth a decision rather than a side effect.

Verified clean and not repeated below: `queuedCastAbility` / `queuedCastAim` are server-local (absent from the wire snapshot and from any persisted JSONB), initialized in both `entity.ts baseEntity` and `net/online.ts blankEntity` so the two `IWorld` entity shapes stay in parity, no new rng draws at queue or at fire, no re-entrancy or double-decrement in `fireQueuedCast` (both slots are nulled before `castAbility` runs, and `castingAbility` is already null), no i18n surface, and no DB or auth surface. Screenshots are correctly N/A for a sim-only change.

Findings 1 and 2 are the ones that need to land before this can go in: as it stands the feature does not fire in the case it was built for, and the parity gate would not have caught it.

jgyy:
Thanks for the thorough follow-up, @EnriqueGF. Fixed both blocking findings, addressed should-fix #3 and #4, and rebased onto the current `release/v0.23.0` (this branch was conflicting).

**Blocking, fixed:**

1. `tests/parity/scenarios.ts` (spell-queue segment): replaced the hardcoded 43-tick drain with a loop that polls `castRemaining` until it is inside `CAST_QUEUE_WINDOW_SEC`, and added assertions right at the press (`castingAbility` still set, `queuedCastAbility === 'fireball'` after the press) so the scenario fails loudly instead of silently degrading to the pre-PR "You are busy." branch again. Golden regenerated for the real queue segment this time.

2. `src/sim/combat/casting_lifecycle.ts` (sub-GCD drop): `fireQueuedCast` now holds the slot when the queued ability's GCD is still running instead of clearing it and calling in, and `updateCasting` retries the hold every tick (when `castingAbility` is null) until the GCD clears. Added a dedicated unit test (`holds a queued cast that would complete before the arming GCD clears...`) that reproduces your repro shape (hasted 1.5s-base cast) and proves the press eventually fires instead of vanishing.

**Should-fix, addressed:**

3. Added one assertion per force-stop clear: `handleDeath`, `readyArenaFighter` (arena), `fiestaDownEntity`, `releasePlayerSpirit` + `resurrectAtSpiritHealer` (spirit release/revive), and `clearNythraxisWardChannelCast`. Each test queues a press, force-stops through that exact path, and asserts `queuedCastAbility`/`queuedCastAim` are cleared. A future refactor that deletes one of these clears will now fail a test.

4. `fiesta_bots.ts`: added a comment confirming the bot behavior change is intended (bots go through the same `castAbility` seam as players, so the queue applies to them too) and deterministic (bot presses are driven by `tickCount`/`pid`, not rng), and noted the existing fiesta parity goldens are what pins it.

5. Added `carries the queued aim point through to the fired ground-targeted cast` covering the `queuedCastAim` plumbing (queued at press time, passed through to the fired `castAbility`, observed via the fired cast going on cooldown).

Left nit #6 (a stale-error toast on a queued press that becomes invalid by fire time) as-is; still a fair point but felt out of scope for this round.

Verified: `npx tsc --noEmit` clean, `tests/combat_casting_lifecycle.test.ts` (26 passed, up from 18), `tests/parity/parity.test.ts` (98 passed, only `c4a_casting_lifecycle` golden regenerated, for the corrected queue drive), `tests/architecture.test.ts` + `tests/localization_fixes.test.ts` + the Nythraxis suites all green, `npm run build` clean, and the pre-push QA floor (tsc, guard tests, scoped biome, copy scan) green.

Rebased onto `release/v0.23.0` (`29a58420e`) to clear the conflict; only the golden JSON conflicted, and that's regenerated fresh above anyway.

FernandoX7 (commented):
Both blockers from last round are properly fixed (verified by running, not reading): the parity segment now polls into the window and asserts the queue, and the sub-GCD hold fires the queued press once the GCD clears (mutation-tested decisive). Two things left.

**should-fix**

`tests/combat_casting_lifecycle.test.ts:444-471`: the `releasePlayerSpirit` and `resurrectAtSpiritHealer` clear-tests do not protect their clears. Both require `p.dead`, so they call `handleDeath` first, and `handleDeath` already nulls the queue (`damage.ts:571-572`) before the release/resurrect clear runs. Confirmed by mutation: deleting both clears in `spirit.ts:124-125,208-209` leaves all six force-stop tests green. Re-arm the slot after `handleDeath` in those two tests so they actually exercise the clear, or drop them and note the clears are redundant with the death clear.

**nit**

`tests/parity/scenarios.ts:3020`: the comment says fireball rank 3 is `2.5s`, but at level 12 the cast is 3.0s. The drive polls dynamically now so it does not matter mechanically, just a stale number.

jgyy:
Thanks for the mutation-tested verification, @FernandoX7 (and thanks again to @EnriqueGF, @ryan-foo for the earlier passes).

**should-fix, fixed:** you were right, `handleDeath` clears the queue before `releasePlayerSpirit`/`resurrectAtSpiritHealer` ever run, so both tests were passing on the death clear, not their own. Re-armed `p.queuedCastAbility`/`p.queuedCastAim` right after the prior clear in each test (`tests/combat_casting_lifecycle.test.ts:454-458` and `:467-476`), so each now exercises only the clear it names. Verified by the same mutation you ran: deleting `spirit.ts:124-125,208-209` now fails exactly those two tests (previously green), restoring them passes again.

**nit, fixed:** corrected the stale comment in `tests/parity/scenarios.ts:3020` (fireball rank 3 is 3.0s at level 12, not 2.5s). Cosmetic only, as you noted, since the drive polls `castRemaining` dynamically.

Verified: `npx tsc --noEmit` clean, `tests/combat_casting_lifecycle.test.ts` (26 passed), `tests/parity/parity.test.ts` + `tests/architecture.test.ts` (122 passed, no golden regen needed, comment/test-only change), pre-push QA floor green.

Pushed to the branch.

FernandoX7 (approved):

PR #1649, opened 2026-07-08
refactor(equipment): make weapon requiredClass list every class that can equip

ryan-foo:
## Summary

A weapon's `requiredClass` was an archetype **signature**, not the real list. Thronebane was `['warrior','paladin']`, yet `canEquipItem` (via `weaponArchetypeForItem`) opened it to the whole warrior-weapon group, so a rogue or hunter could equip it. The field read like a hard class lock while the game let five classes wear it. Weapons here are proficiency-based, not class-locked, but the data did not say so.

This makes the data representative. Every archetype weapon's `requiredClass` now lists exactly the classes that can equip it:
- warrior group -> `warrior, rogue, hunter, shaman, paladin`
- caster group -> `mage, priest, warlock, shaman, paladin, druid`
- rogue group was already its full set (`rogue, hunter`), unchanged

`weaponArchetypeForItem` now recovers the archetype by matching the full list against each group set (order-independent equality) instead of a subset of the legacy signature sets. Equip results and the item tooltip (which still suppresses the redundant class line for group weapons) are byte-identical to before. A future weapon with a genuine bespoke lock has no archetype and correctly shows its class line.

13 weapons expanded (all in `zone3.ts`, including both legendaries). No gameplay behavior change.

## Type of change

## How was this tested?

- `npx vitest run tests/equipment_proficiency.test.ts tests/item_armor_type.test.ts tests/items.test.ts tests/item_level.test.ts tests/progression.test.ts tests/inspect_equipment.test.ts tests/item_compare.test.ts tests/unequip_item.test.ts tests/equip_jewelry.test.ts tests/equip_level_requirement.test.ts tests/guide.test.ts tests/architecture.test.ts` (all pass). `npx tsc --noEmit` clean (the only local error is a pre-existing missing `prom-client` dep, untouched here).
- New guard in `equipment_proficiency.test.ts`: for every weapon with a `requiredClass`, the field must equal exactly the set of classes `canEquipItem` allows. It fails on the old short-form data, so it pins the fix and blocks future drift.

N/A: no player-visible copy or UI change (group-weapon tooltips still hide the class line), so no screenshots.

### DIFF
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -2173,7 +2173,7 @@ export const ZONE3_ITEMS: Record<string, ItemDef> = {
     weapon: { min: 18, max: 29, speed: 2.3 },
     stats: { str: 6, sta: 2 },
     sellValue: 900,
-    requiredClass: ['warrior', 'paladin', 'shaman'],
+    requiredClass: ['warrior', 'rogue', 'hunter', 'shaman', 'paladin'],
   },
   emberwood_staff: {
     id: 'emberwood_staff',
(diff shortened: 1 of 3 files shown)


jgyy:
small change to me, direct manual code review shows no issue listing more class that can equip it

Issue #1651, opened 2026-07-08
Character walks backwards without input after dying and reviving


MasterZensei:
## Problem

After dying and reviving on the 0.23.0 release, the character starts walking backwards on its own with no backward input held. The player is not pressing S (or any movement key), yet the character keeps moving as if a backward input were stuck. This means stale or phantom movement input survives the death and revive flow, and it can walk the freshly revived player straight back into danger.

## Steps to reproduce

1. Play a character online on the 0.23.0 release.
2. Die.
3. Revive.
4. Release all movement keys and observe the character.

## Actual behavior

- After reviving, the character walks backwards continuously without any movement key held.

## Expected behavior

- After reviving, the character stands still until the player gives new movement input.
- All movement input and streamed intent state is cleared on death and again on revive, on both the client and the server.

## Scope

- Local input state in src/game/: whether a held-key or synthetic backward input is left set across the death/revive flow.
- The 20 Hz movement intent stream in src/net/online.ts: whether a stale intent keeps streaming, or is resent, after the revive.
- Server-side intent handling in server/game.ts: whether the last received intent is replayed onto the revived character before a fresh one arrives.
- Whether the offline browser world reproduces it, which would point at the shared sim/input path rather than the network path.

## Acceptance criteria

- Reviving after death never produces movement without a live input.
- The source of the stale input is identified and covered by a regression test for input/intent reset across death and revive.

## Environment

- Release: 0.23.0
- Mode: online

PR #1655, opened 2026-07-08
fix(sim): clear held movement intent on arena respawn and delve-companion revive (#1651 follow-up to #1723)

daxdax89:
## Update: rescoped as a follow-up to #1723

#1723 has since merged and fixes issue #1651 for the graveyard release, healer revive, and delve first-death / eject paths. I have rebased this PR onto the current release and **dropped everything #1723 now covers**, so it no longer overlaps. What remains are the two respawn sites that route around #1723's fixes and were left uncovered:

- **`readyArenaFighter` (`src/sim/social/arena.ts`)** - the arena's own placement path (Fiesta and ranked) resets hp/resource/target/auto-attack but never `moveInput`, so a fighter revived into a live bout drifted from a stale forward/back flag with no key held.
- **`updateDelveCompanion` (`src/sim/delves/companion.ts`)** - the rank-3 companion boon that revives a downed ally mid-run never cleared that ally's `moveInput`.

Both now clear `PlayerMeta.moveInput` in place, exactly matching #1723's approach at the sites it covered.

## Tests
- `tests/arena_respawn_intent.test.ts` (new) - a benched Fiesta fighter does not auto-walk from a held intent after revive.
- `tests/delve_companion.test.ts` - added the rank-3 companion-revive intent-clear case.
- Both were verified to **fail without** the code change and pass with it (decisive, not vacuous). Determinism confirmed: the `fiesta_powerups` parity golden trace still matches.

## Type of change
- [x] Bug fix

If you would rather fold these two paths into #1723's follow-up work or handle them differently, happy to adjust or close - just wanted the two uncovered revive sites not to slip through.

### DIFF
--- a/src/sim/delves/companion.ts
+++ b/src/sim/delves/companion.ts
@@ -21,6 +21,7 @@ import {
   DT,
   dist2d,
   type Entity,
+  emptyMoveInput,
   MELEE_RANGE,
   PET_TELEPORT_DISTANCE,
   steadyAngleTo,
@@ -68,6 +69,11 @@ export function updateDelveCompanion(ctx: SimContext, companion: Entity): void {
       // re-entering mid-run cannot recharge the boon.
       run.companionReviveUsed = true;
       fallen.dead = false;
+      // Clear any movement intent held at death so the revived ally does not walk
+      // off on its own with no key held (issue 1651, companion-revive path). fallen
+      // is always a player (owner or a partyMembersForKey ally), so it has a meta.
+      const fallenMeta = ctx.players.get(fallen.id);
+      if (fallenMeta) Object.assign(fallenMeta.moveInput, emptyMoveInput());
       fallen.hp = Math.max(1, Math.round(fallen.maxHp * 0.5));
       if (fallen.resourceType === 'mana')
         fallen.resource = Math.max(fallen.resource, Math.round(fallen.maxResource * 0.5));
--- a/src/sim/social/arena.ts
+++ b/src/sim/social/arena.ts
@@ -33,6 +33,7 @@ import {
   type ArenaStanding,
   DT,
   type Entity,
+  emptyMoveInput,
 } from '../types';
 
 // Ashen Coliseum 1v1 arena tuning consts (moved with the slice). FIESTA_COUNTDOWN
@@ -852,6 +853,10 @@ export function readyArenaFighter(ctx: SimContext, e: Entity, opts: { clearPrep:
   e.resource = e.resourceType === 'mana' ? e.maxResource : e.resourceType === 'energy' ? 100 : 0;
   e.targetId = null;
   e.autoAttack = false;
+  // Drop any held movement intent so a fighter placed/respawned into the arena does
+  // not drift from a stale forward/back flag with no key held (issue 1651); bots
+  // already reset theirs on spawn.
+  if (meta) Object.assign(meta.moveInput, emptyMoveInput());
   e.queuedOnSwing = null;
   delete e.queuedOnSwingFree;
   e.queuedCastAbility = null;
--- a/tests/arena_respawn_intent.test.ts
+++ b/tests/arena_respawn_intent.test.ts
@@ -1,0 +1,68 @@
+// Regression for issue 1651 (arena arm): a fighter's last-held movement key
+// survived on PlayerMeta.moveInput through the arena respawn. readyArenaFighter
+// (the one place a benched Fiesta/ranked fighter is placed back into the bout)
+// reset hp/resource/target/auto-attack but never cleared moveInput, so the
+// revived fighter drifted in the last-held direction with no key pressed. The
+// graveyard/delve respawn paths are covered elsewhere (spirit.ts / entity_roster
+// / runs.ts); this pins the arena path readyArenaFighter owns.
+import { describe, expect, it } from 'vitest';
+import { Sim } from '../src/sim/sim';
+import type { PlayerClass } from '../src/sim/types';
+import { groundHeight } from '../src/sim/world';
+
+function makeWorld() {
+  return new Sim({ seed: 42, playerClass: 'warrior', noPlayer: true });
+}
+
+function teleport(sim: Sim, pid: number, x: number, z: number) {
+  const e = sim.entities.get(pid)!;
+  e.pos.x = x;
+  e.pos.z = z;
+  e.pos.y = groundHeight(x, z, sim.cfg.seed);
+  e.prevPos = { ...e.pos };
+  (sim as any).rebucket(e);
+}
+
+// Seat a 2v2 Fiesta with four solo-queued players and run the countdown out so
+// the bout is live. Fiesta respawns route through readyArenaFighter, so it is
+// the clean vehicle for exercising the arena revive path.
+function startFiesta(classes: PlayerClass[] = ['warrior', 'mage', 'rogue', 'priest']) {
+  const sim = makeWorld();
+  const pids = classes.map((c, i) => sim.addPlayer(c, `P${i}`));
+  pids.forEach((p, i) => {
+    teleport(sim, p, i * 4, -40);
+  });
+  pids.forEach((p) => {
+    sim.arenaQueueJoin(p, 'fiesta');
+  });
+  sim.tick(); // matchmake
+  for (let i = 0; i < 20 * 8; i++) {
+    sim.tick();
+    const m = sim.arenaMatchFor(pids[0]);
+    if (m && m.state === 'active') break;
+  }
+  const match = sim.arenaMatchFor(pids[0])!;
+  return { sim, match };
+}
+
+describe('arena respawn clears stale movement intent (issue 1651)', () => {
+  it('a benched fiesta fighter does not auto-walk from a held intent after revive', () => {
+    const { sim, match } = startFiesta();
+    const killerPid = match.teamA[0];
+    const victimPid = match.teamB[0];
+    const victim = sim.entities.get(victimPid)!;
+    const killer = sim.entities.get(killerPid)!;
+    const victimMeta = (sim as any).players.get(victimPid);
+    // Down the victim while it holds a forward intent, and keep the flag set while it
+    // is benched so readyArenaFighter is the only thing that could clear it on revive.
+    victimMeta.moveInput.forward = true;
+    (sim as any).dealDamage(killer, victim, victim.maxHp + 50, false, 'physical', null);
+    expect(victim.dead).toBe(true);
+    victimMeta.moveInput.forward = true;
+    const downedFor = match.fiesta!.respawn.get(victimPid)!;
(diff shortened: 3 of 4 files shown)


EnriqueGF:
:) Thanks for the fix, this is a clean and well-scoped bug hunt and the root-cause writeup is spot on.

## Base branch gate: PASS
Targets `release/v0.23.0`, which is the latest release branch. Good.

## What I verified locally (branch checked out)
- `npx tsc --noEmit`: clean.
- `tests/spirit.test.ts`, `tests/delves.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`: all green.
- Decisiveness: reverting only the three source files makes all three new `1651` tests fail (the body walks off), and restoring turns them green. The assertions are decisive, not pins.
- `Object.assign(meta.moveInput, emptyMoveInput())` is the correct in-place idiom (preserves the object identity the server/local getter alias every tick); matches the existing server convention at `server/game.ts:1086,2007,2768`. No rng, no tick-phase change, purity intact. Determinism/parity confirmed unaffected.
- No em/en dashes, emojis, `.only(`, or `debugger` in the diff. No new player-facing strings, so i18n is N/A.

## Requested changes (both small)

### 1. One re-animation path with the same bug is still uncovered: `ejectToDelveDoor` (src/sim/delves/runs.ts:617)
The fix clears the intent for the *dying* player in `releaseSpiritInDelve`, but `ejectToDelveDoor` is a separate revive funnel: it does `p.dead = false` and teleports the body to the door, and it does NOT clear `moveInput`. I confirmed `handleDeath` (src/sim/combat/damage.ts:538) does not zero `moveInput` either, so the stale flag genuinely survives.

It is reachable for a *bystander* party member who is dead-but-not-yet-released when either arm fires:
- `failDelveRun` (runs.ts:648-649) ejects every member, including a still-dead one.
- the run-complete `surface_exit` path (runs.ts:1418-1420) does the same.

So in a party delve, a member who died holding forward and has not released will auto-walk from the door on eject: exactly issue #1651, one path over. Solo is fine (the solo player goes through the now-fixed `releaseSpiritInDelve` first). Please clear `moveInput` in `ejectToDelveDoor` too (one line, beside the existing `targetId`/`inCombat` resets), or justify why the bystander-eject case is safe.

### 2. The arena/fiesta/yumi arm has no direct test
`readyArenaFighter` (src/sim/social/arena.ts:859) is the third fix site (arena start, per-death respawn via `fiestaRevive`/`yumiRevive`, and `resetForArena` return-home), but only the spirit and delve arms got a direct assertion. Per the repo rule that sim behavior gets a test, please add one that sets `meta.moveInput.forward = true` before a fiesta/arena respawn and asserts it clears. Low risk since the path is exercised, but it is currently unpinned.

## Non-blocking note
`fiesta_bots.ts:118` and `vale_cup_bots.ts:394` reset via reassignment (`meta.moveInput = emptyMoveInput()`) rather than the in-place `Object.assign` idiom this PR and the server use. Harmless for bots (nothing aliases a bot's `moveInput`), and outside this diff, but worth normalizing in a follow-up.

Nice, tightly-scoped fix overall. Close path 1 and add the arena test and this is good to go.

daxdax89:
Thanks for the thorough review, both requests are addressed (pushed, rebased onto the current `release/v0.23.0` tip).

**1. `ejectToDelveDoor` (bystander-eject path).** You were right: a member who is dead-but-not-released when the run fails never runs `releaseSpiritInDelve`, so `failDelveRun`/`surface_exit` ejected it with the stale intent intact and the body auto-walked from the door. Cleared `moveInput` there too, beside the existing `targetId`/`inCombat` resets, using the same in-place `Object.assign(r.meta.moveInput, emptyMoveInput())` idiom:

```ts
  p.targetId = null;
  p.combatTimer = 99;
  p.inCombat = false;
  p.autoAttack = false;
  // A bystander party member ejected while still dead-but-not-released never runs
  // releaseSpiritInDelve, so clear any movement intent held at death here too, or
  // the body auto-walks from the door (issue 1651, bystander-eject path).
  Object.assign(r.meta.moveInput, emptyMoveInput());
```

**2. Arena/fiesta arm now has a direct test.** Added `tests/fiesta.test.ts` > 'a fiesta respawn clears a movement intent held at death (issue 1651)': it holds `moveInput.forward` while the victim is benched and asserts `readyArenaFighter` (via the fiesta respawn timer -> `ctx.readyArenaFighter(e, { clearPrep: true })`) clears it on revive. Also added a decisive `tests/delves.test.ts` case for the eject path above (a duo bystander downed-and-unreleased, then `failDelveRun`, asserting the body is ejected and its intent cleared).

Both new tests fail if their respective clear is reverted (verified), so they are decisive, not pins. `tsc`, `spirit`/`delves`/`fiesta`/`architecture`/`localization_fixes` all green.

**Non-blocking note (`fiesta_bots.ts`/`vale_cup_bots.ts` reassign vs in-place):** agreed, harmless for bots since nothing aliases a bot's `moveInput`; happy to normalize those to the in-place idiom in a follow-up so the convention is uniform.

EnriqueGF:
:) Thanks for turning this around so fast, both requested changes landed exactly as asked and the new tests are genuinely decisive.

## Base branch gate: PASS
Targets `release/v0.23.0`, the latest release branch, and the head contains the current tip.

## Re-verified from the round-1 requests (both closed)
- `ejectToDelveDoor` (src/sim/delves/runs.ts:643): the bystander-eject path now clears the intent beside the existing `targetId`/`inCombat` resets. Correct spot, correct idiom.
- `readyArenaFighter` now has a direct test (`tests/fiesta.test.ts`), and the delve eject arm got one too.

## What I verified independently in a worktree
- `npx tsc --noEmit`: clean.
- `tests/spirit.test.ts`, `tests/delves.test.ts`, `tests/fiesta.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`: 233 passed.
- `tests/parity`: 164 passed, draw-order digest unchanged. No rng, no tick-phase change, sim purity intact.
- Decisiveness: reverting only the five source clears (tests untouched) makes exactly the five new `1651` tests fail and nothing else. They are decisive, not pins.
- No em/en dashes, emojis, `.only(`, or `debugger`. No player-facing strings, so i18n is N/A. No DB or wire surface touched. Not a visual change, so no screenshots needed.

## Requested change (one, blocking)

### The rank-3 delve companion revive still leaves the stale intent: src/sim/delves/companion.ts:70
This PR states it clears the intent in "every path that re-animates a player without a fresh input frame". One is still missing. Acolyte Tessa's rank-3 boon ("revives a fallen ally once per run") sets `fallen.dead = false` and refills HP, but never touches `moveInput`. I confirmed `handleDeath` (src/sim/combat/damage.ts:538) sets only `e.dead = true`, and the movement kernel is gated on `if (!p.dead)` (src/sim/sim.ts:3173), so a flag held at death survives the whole dead interval and re-activates the tick after the revive.

I reproduced it against this branch with a throwaway test: level 20 warrior, `companion_tessa` rank 3, enter `collapsed_reliquary`, hold `forward`, die, let `updateDelveCompanion` fire. After the revive, `moveInput.forward === true` and the body drifts **7.000 yards in one second** with no key held. That is issue #1651 verbatim, one path over, and it is player-reachable solo.

`fallen` is always a player (the owner, or a `partyMembersForKey` ally), so the fix mirrors the other sites:

```ts
      run.companionReviveUsed = true;
      fallen.dead = false;
      const fallenMeta = ctx.players.get(fallen.id);
      if (fallenMeta) Object.assign(fallenMeta.moveInput, emptyMoveInput());
```

Please add that plus a direct assertion in `tests/delve_companion.test.ts` (the existing "rank 3 revives the fallen owner once per run at half health" case is one line away from covering it). With five sites already fixed, leaving the sixth means the issue can still be reported by any player running a maxed companion.

## Non-blocking
`tests/fiesta.test.ts` carries a pre-existing Biome format error on its line 5 import (it is already red on `release/v0.23.0`, so it is not your regression). Since the file is now a changed file, `biome ci --changed` will surface it. A `npx @biomejs/biome check --write tests/fiesta.test.ts` keeps CI clean while you are in there.

Everything else here is production-ready. Close the companion path and this is an approve.

daxdax89:
Thanks for catching the sixth site, that is a clean find and you are right it is player-reachable solo.

**Fixed: the rank-3 companion revive (`src/sim/delves/companion.ts`).** Acolyte Tessa's once-per-run boon now clears the intent right after `fallen.dead = false`, with the same in-place idiom as the other sites:

```ts
      run.companionReviveUsed = true;
      fallen.dead = false;
      // Clear any movement intent held at death so the revived ally does not walk
      // off on its own with no key held (issue 1651, companion-revive path). fallen
      // is always a player (owner or a partyMembersForKey ally), so it has a meta.
      const fallenMeta = ctx.players.get(fallen.id);
      if (fallenMeta) Object.assign(fallenMeta.moveInput, emptyMoveInput());
      fallen.hp = Math.max(1, Math.round(fallen.maxHp * 0.5));
```

**Test:** added `tests/delve_companion.test.ts` > 'rank 3 revive clears a movement intent held at death (issue 1651)' (holds `forward`, dies, rank-3 revive fires, asserts the flag clears). I verified it is decisive: reverting only the two-line clear makes exactly this new test fail and nothing else. That is the sixth and last re-animation site; I re-grepped the revive paths and `ejectToDelveDoor`/`releaseSpiritInDelve`/`readyArenaFighter`/`resurrectAtSpiritHealer`/`reviveAt`/companion are all covered now.

**Biome (your non-blocking note):** ran `biome check --write tests/fiesta.test.ts` and also converted the four `forEach` callbacks in the pre-existing `startFiesta`/augment helpers to block bodies, so the file is now fully clean under `biome ci` (0 errors), not just the line-5 import.

**Verified locally (fresh clone on the current `release/v0.23.0` tip `12d645617`):**
- `tsc --noEmit`: clean.
- `tests/delve_companion.test.ts tests/spirit.test.ts tests/delves.test.ts tests/fiesta.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/i18n_resolved_equivalence.test.ts tests/i18n_completeness.test.ts`: 268 passed.
- `biome ci` on all touched files: 0 errors.
- Branch is based on the current tip, merge-tree clean, no rng / tick-phase / wire / DB / i18n surface touched.

**Non-blocking note you raised (`fiesta_bots.ts`/`vale_cup_bots.ts` reassign vs in-place):** still happy to normalize those to the in-place idiom in a separate follow-up so the convention is uniform, since nothing aliases a bot's `moveInput`.

daxdax89:
Rescoped this on top of #1723 (thanks @jgyy): dropped everything #1723 now covers and kept only the two revive sites it did not touch, the arena placement path (readyArenaFighter) and the rank-3 delve-companion revive. Both clear moveInput the same way #1723 does elsewhere, with decisive tests that fail without the change. Merges clean on release/v0.24.0.

PR #1656, opened 2026-07-08
fix(ui): theme the Loot Settings dropdowns to match the HUD (#1646)

### LINKED ISSUE #1646: Loot Settings dropdowns use unstyled native selects that clash with the HUD
## Problem

The Loot Settings window itself follows the classic HUD look (dark panel, gold border, small-caps gold headings), but the Loot Method dropdown renders as a default browser select: white background, black system font, native arrow. The loot threshold select shown to the party leader when Master Loot is enabled has the same problem. The controls clash with the window they sit in and look unfinished.

## Expected behavior

- The Loot Method and threshold dropdowns match the HUD theme: dark background, gold or parchment text, themed border and arrow, consistent with the rest of the window.
- Hover, focus, and disabled states stay readable, and keyboard operation is unchanged.
- The style is a shared component class, not a one-off, so other HUD selects can adopt it.

## Scope

- `src/ui/loot_settings_window.ts` renders the two native selects (`#ls-method`, `#ls-threshold`); the view core is `src/ui/loot_settings_view.ts`.
- Styling belongs in the extracted HUD CSS under `src/styles` (components layer), not inline in the painter.
- Both game shells share this window (index.html and play.html), so check both entries.
- Other HUD windows also use native selects (for example the report window); they are candidates for the same shared class but can be a follow-up.

## Acceptance criteria

- The closed Loot Method and threshold controls render with the HUD theme (dark background, themed text, border, and arrow) instead of the browser default.
- Keyboard and screen reader operation of both selects is unchanged.
- The window looks the same in both game shells; the entry window parity test still passes.
- Desktop and mobile both get the themed control, with no layout regressions in either.

## Notes

- The open dropdown list (the OS popup) is only partially styleable; theming the closed control is the expected minimum. If full theming of the open list is wanted, the market UI's custom dropdown pattern (`.mkt-select-*`) is an existing in-repo precedent, but that is a bigger change than this issue requires.

daxdax89:
## Problem

The Loot Settings window follows the classic HUD look (dark panel, gold border, small-caps gold headings), but its Loot Method and roll-threshold dropdowns rendered as default browser selects: white background, black system font, native arrow. They clashed with the window and looked unfinished (issue #1646).

## Fix

Add a shared **`.hud-select`** component class (components layer) that themes a native `<select>` from tokens: `--color-bg-dark` background, `--color-text-light` text, `--color-border-default` border with a `--color-border-focus` hover/focus ring, and a caret drawn from two gradients tinted by the `--gold` accent. Apply it to `#ls-method` and `#ls-threshold`.

Keeping a real `<select>` (not the fully custom `.ui-dd` widget) preserves native keyboard and type-ahead semantics unchanged. The class is shared so other HUD selects (the report window, etc.) can adopt it.

## Theme + a11y correctness

- **Light themes:** background and caret both use theme-aware tokens (`--color-bg-dark`, `--gold`), which flip to a light panel edge and a dark readable accent under the Parchment preset, so the control stays readable on every theme (not just the dark default). See the third panel in the screenshot.
- **Mobile:** a `@media (pointer: coarse)` rule holds the 40px tap-target floor; the base.css 16px input-font floor already prevents iOS zoom.
- **Focus:** a steady, token-drawn `:focus-visible` ring (no transition).
- **Forced-colors:** a high-contrast block restores the native control + a system border.

The one CSS shell (imported once from `src/main.ts`) covers both index.html and play.html. No new strings (the option labels already go through `t()`).

## Tests

`tests/loot_settings_select_theme.test.ts` pins that `.hud-select` is defined with the token-driven rules (`appearance:none`, the bg/border/text tokens, a `--gold` caret with no fixed hex, the coarse-pointer min-height, a token-drawn `:focus-visible` ring, the forced-colors restore) and is applied to both selects.

## Before / after

Native selects vs the themed `.hud-select` on the dark HUD, plus the Parchment theme (readable) attached below.

Fixes #1646


EnriqueGF:
:) Thanks for the polish here, this is a clean, well-scoped fix and the tests are decisive.

I verified the change end to end and it is in good shape:
- Base branch gate passes: it targets `release/v0.23.0` (the latest release), is mergeable, and the touched files do not conflict.
- `tsc --noEmit` is clean and the CSS guard suites plus the new `loot_settings_select_theme.test.ts` pass (37/37).
- The tokens are genuinely theme-aware (confirmed in `theme.ts`: `--color-bg-dark` and `--gold` both flip under Parchment), so the readability claim holds.
- The iOS-zoom reasoning checks out: the base rule sets `font-size: 12px`, but `base.css` floors `select` at `16px !important` on coarse pointers, and an important declaration wins across `@layer` boundaries, so the mobile floor is not defeated.
- No new strings, no magic values (all token-driven), no em/en dashes or emojis.

One blocking item before this can be approved:

1. **Missing screenshots (visual change).** The description says the before/after and the Parchment panel are "attached below" and references "the third panel in the screenshot," but no image is actually attached or committed under `docs/screenshots`. A visual change needs before/after screenshots for desktop and mobile. Please attach them (or commit them under `docs/screenshots` and reference them from the body).

Optional, non-blocking: the branch is a bit behind the `release/v0.23.0` tip (33 commits); a rebase before merge keeps CI honest, though there are no conflicts.

Once the screenshots are up, this is good to go.

EnriqueGF:
:) Thanks for this, and thanks for the thorough write-up. Re-reviewing after my earlier pass.

I re-verified the branch from scratch on a clean worktree at 58538ad1:
- Base branch gate still passes: targets `release/v0.23.0` (the latest release), MERGEABLE, no conflicts.
- `tsc --noEmit` clean.
- Guard suites green: `architecture.test.ts`, `localization_fixes.test.ts`, plus the new `loot_settings_select_theme.test.ts` (59 passed / 3 skipped).
- CSS gates green: `css_corpus`, `css_value_validity`, `styles_extraction`, `theme`, `per_entry_css_wiring` (48 passed).
- `biome ci --changed` exits 0 (two lint warnings only, which CI does not fail on).
- Every token the rule uses (`--cursor-point`, `--radius-sm`, `--ui-font`, `--color-border-default`, `--color-border-focus`, `--color-text-muted`, `--color-text-light`, `--color-bg-dark`) is defined, so nothing silently no-ops.
- The rule lands inside `@layer components`, and `base.css` floors `select` at `16px !important` on coarse pointers, so the iOS-zoom floor is not defeated by the `font-size: 12px` in the base rule.

The code itself is good. One item is still blocking, and it is the same one as before.

## Blocking

1. **Screenshots are still missing (visual change).** No new commits have landed since my last comment, and the body still says the before/after and the Parchment panel are "attached below" and references "the third panel in the screenshot," but there is no image in the PR body and nothing under `docs/screenshots`. The repo rule is explicit: a visual change needs before/after screenshots, desktop and mobile. Please attach them or commit them and reference them from the body.

## Non-blocking

2. **The stated convention fit deserves a sentence.** `src/ui/loot_settings_window.ts:55,59` are now the last native `<select>` elements in the HUD. `options_window.ts:761` and `talents_window.ts:662` both migrated to the custom `.ui-dd` widget on purpose, and `src/ui/dropdown_nav.ts` implements the full WAI-ARIA listbox keyboard pattern for it. Your rationale is not wrong: `.ui-dd` has no type-ahead, so a real `<select>` does keep something. But the description frames `.ui-dd` as simply lacking keyboard semantics, which is not accurate, and it is worth stating plainly that this deliberately diverges from where the rest of the HUD went, so a future reader does not "fix" it.

3. **The description and commit body do not match the code.**
   - The body says other HUD selects "(the report window, etc.) can adopt this class." The report window (`src/ui/bug_report.ts`) has no `<select>` at all, so today there is no second consumer.
   - The commit body describes a `--color-bg-input` background and a "data-URI arrow." The shipped code uses `var(--color-bg-dark)` (`src/styles/components.css:2506`) and two gradient carets. The tests correctly pin the real code, so this is narrative drift only, but please fix the message.

4. **Two test nits** in `tests/loot_settings_select_theme.test.ts`:
   - Line 15: the `base` regex `/\.hud-select\s*\{([\s\S]*?)\}/` takes the FIRST bare `.hud-select {` in the file. There are three (the base rule, the `pointer: coarse` block, the `forced-colors` block). It works today only because the base rule comes first, so a harmless CSS reorder turns the suite red. Anchoring it would make it robust.
   - Line 35: `not.toContain('%23ffd100')` is vacuous. The implementation draws the caret with gradients, never a data URI, so that URL-encoded hex cannot appear in any implementation, correct or broken. The positive `toContain('var(--gold)')` on the line above is what actually guards it.
   - Line 38: the coarse-pointer regex does not anchor `.hud-select` and `min-height: 40px` to the same rule, so the floor could migrate to a sibling selector and still pass.

5. The branch is now 61 commits behind the `release/v0.23.0` tip (it was 33). No conflicts, but a rebase before merge keeps CI honest.

Item 1 is the only thing standing between this and an approval. The rest is polish.

daxdax89:
Screenshots added (committed under `docs/screenshots/`, rendered from the shipped `.hud-select` CSS).

**Desktop** - native OS `<select>` (before) vs the themed `.hud-select` on the dark HUD, and on the light Parchment preset where `--color-bg-dark`/`--gold` flip so the caret and text stay readable:

loot settings dropdown, desktop before/after

**Mobile** (narrow) - the three panels stacked; the control keeps its full width and the coarse-pointer `min-height: 40px` tap floor:

loot settings dropdown, mobile

That closes the only blocking item. Thanks for the review.

PR #1662, opened 2026-07-08
Feat/billboard sprites and lazy loading

Leem0nStudio:
## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
- Manual steps:

## Screenshots / recordings

---

## Checklist

### Quality

- [ ] **Typecheck passes.** `npx tsc --noEmit` is clean (the project is
      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

- [ ] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is
- [ ] I didn't hand-edit generated files such as

> **Tip:** commit messages and PR titles use [Conventional Commits](https://www.conventionalcommits.org/)
> with a scope where it fits (`feat(talents): ...`, `fix(net): ...`). It's a
> convention we like, not a requirement. Clear, descriptive messages are what
> matter most.

---

A complete checklist and a green CI run (tests, typecheck, and builds) are what
we look for before merging. Smaller, focused PRs land faster than large ones, and
reviewers may suggest changes, which is a normal and friendly part of the
process. Thank you for helping build World of ClaudeCraft. Questions? Join us on
[Discord](https://discord.gg/GjhnUsBtw).


EnriqueGF:
Thanks a lot for putting this together :) There is genuinely interesting work here, in particular the billboard sprite pipeline and the idea of a WebSocket fallback for restrictive networks. I reviewed it against the code in a worktree rather than against the description, and I am afraid it is not mergeable yet. Findings below, most severe first.

## Blocking

**1. Wrong base branch.** This PR targets `main`. Contributions must target the latest release branch, which is `release/v0.23.0`. Please re-point and rebase. This alone blocks the merge regardless of the code.

**2. Secrets committed to `.env.example`.** The placeholders were replaced with what looks like a real local password:

- `.env.example:7` `POSTGRES_PASSWORD=50778012`
- `.env.example:13` `DATABASE_URL=postgres://eastbrook:50778012@127.0.0.1:5433/eastbrook`
- `.env.example:27` `MEDIAWIKI_DB_PASSWORD=50778012`
- `.env.example:29` `MEDIAWIKI_ADMIN_PASS=50778012`

`.env.example` must keep the `change-me` placeholders. If that value is used anywhere real, please rotate it, since it is now in the PR history.

**3. The branch does not typecheck.** `npx tsc --noEmit` exits 2 with 29 errors, so CI cannot be green:

- `server/main.ts:2193` `error TS1308: 'await' expressions are only allowed within async functions`. The `/api/transport/close` handler awaits `readBody` inside a non-async function, so the server does not compile.
- `src/render/characters/visual.ts` 14 errors, all `TS18047: 'prep' is possibly 'null'` (lines 167 to 191). The lazy-loading change made `prep` nullable and the call sites were not guarded.
- `src/render/renderer.ts:93` `TS2305: Module './sprites/sprite_lighting' has no exported member 'spriteLighting'`.
- `src/render/sprites/sprite_shadow.ts` 7 errors (`TS2554` wrong arity at 122 to 143, `TS2339` on `Material | Material[]` at 149 and 152).
- `src/render/sprites/sprite_lighting.ts` 4 errors (`TS2339` `.copy`/`.set` on `number | Vector3 | Color`, lines 160 to 164).
- `src/render/characters/assets.ts` 2 errors.

**4. The HTTP transport cannot stay connected.** `server/game.ts:1169` `pingLiveSessions()` sets `session.awaitingPong = true` and calls `session.ws.ping()`. `FakeWs.ping()` is a documented no-op, and `awaitingPong` is only cleared in the WebSocket `'pong'` handler (`server/ws_auth.ts:223`). An HTTP transport session never registers that handler, so on the second keepalive tick `pingLiveSessions` takes the `if (session.awaitingPong)` branch and calls `ws.terminate()` plus `socketClosed()`. Every HTTP client is therefore disconnected after two keepalive intervals. This looks untested end to end.

**5. Guessable session token, session hijacking.** `server/http_transport.ts:204`:

```ts
const sessionId = `http_${Date.now().toString(36)}_${Math.random().toString(36).slice(2, 8)}`;
```

This is the only credential guarding `/api/transport/poll`, `/send` and `/close`. It carries roughly 30 bits of non-cryptographic entropy on top of a predictable timestamp, and `Math.random` is not a CSPRNG. Anyone who guesses one can read another player's entire message stream and inject input and commands as them. Please use `crypto.randomUUID()` or `crypto.randomBytes`, and bind the session to the auth token and the originating IP.

**6. The transport bypasses server protections.**

- `FakeWs.bufferedAmount` is initialised to `0` and never updated (`queueBytes` is tracked separately), so the slow-client kick at `server/game.ts:4421` `isBackpressureExceeded(session.ws.bufferedAmount)` never fires for HTTP clients.
- When the 1 MiB cap is hit, `FakeWs.send` silently sets `readyState = CLOSED` and returns, but nothing calls `game.leave()`. The character is left in the world as a zombie session.
- `httpSessions` (`server/http_transport.ts:89`) has no TTL and no reaper. `removeHttpSession` is exported but never called from anywhere in `server/`, so an abandoned client leaks a map entry and a live `ClientSession` forever.
- On the positive side, `game.handleMessage` does apply `consumeMsgToken`, so per-message rate limiting is preserved. That part is fine.

**7. Endpoints are inline in `main.ts`, against the documented seam.** `server/main.ts:2185` onwards adds four `/api/transport/*` routes as an inline `else if` chain. `CLAUDE.md` and `server/http/CLAUDE.md` require a `RouteDef` module registered in `server/http/registry.ts`. The code comment states the bypass is intentional to "keep the transport self-contained", but that also skips the middleware onion (auth, rate limiting, typed schemas, stable error codes). Please scaffold it with `npm run new:endpoint`.

**8. No tests.** The diff adds a new server transport and a whole new render subsystem and changes `src/sim/` files, with zero test files touched. `CLAUDE.md` requires that `sim/` and `server/` changes always get a test. At minimum: `FakeWs` queueing, backpressure, and session lifecycle; the sprite atlas UV slicing and frame advance.

## Major

**9. Unrelated repo-wide refactor mixed in.** Roughly a third of the diff is a mechanical `.filter().map()` to `.flatMap()` and `for await` to `Promise.all` rewrite across files that have nothing to do with sprites or transports: `src/sim/bags.ts`, `src/sim/items.ts`, `src/sim/market.ts`, `src/sim/loot/loot_roll.ts`, `src/sim/mail/post_office.ts`, `src/sim/delves/runs.ts`, `src/sim/removed_zone1_content.ts`, `src/ui/hud.ts`, `server/chat_filter.ts`, `server/perf_report.ts`, `bot/logic.ts`, and others. It reads like an automated tool's output applied wholesale. Please drop all of it from this PR. Beyond the "avoid unrelated refactors" rule, the `Promise.all` conversions are behaviour changes, not cleanups:

- `server/game.ts` `grantPlaytimePoints` and `recordDailyRewardActivity` now fire one `grantRewardPoints` / `recordOnlineMinute` query per online session concurrently. With a few hundred players online that is a burst of unbounded concurrent queries against a `pg` pool sized in the low tens, which will saturate the pool and stall the world loop. The sequential loop was deliberate.
- `server/oauth.ts` and `server/internal.ts` do the same for `upsertOAuthClient` and `setDiscordMemberMeta`.
- `server/internal.ts` now sets `updated = results.length`, which counts attempted rows rather than the previous per-row increment, and the comment "Only accept a known special-role key; anything else clears the role" was deleted.
- `bot/main.ts` `ensureTierRoles` now creates Discord roles with `Promise.all`. Role creation order determines role position in Discord, so tier ordering becomes non-deterministic, and unbounded parallel calls will hit the Discord rate limiter that the sequential loop avoided.

**10. The `react-doctor` GitHub workflow does not belong here.** `.github/workflows/react-doctor.yml` adds a third-party action (`millionco/react-doctor@v2`, a mutable tag) that runs on every `pull_request` with `pull-requests: write`, `issues: write` and `statuses: write`. This repository contains no React (the only framework is Svelte, scoped to `src/admin/`). Granting write scopes to an unpinned third-party action is a supply-chain risk for zero benefit. The `"doctor": "npx react-doctor@latest"` script added in `package.json:78` executes unpinned remote code. Please remove both. If the workflow is what produced finding 9, that is a further argument against it.

**11. Sprite asset provenance and licensing.** `scripts/import_real_sprites.mjs:1` describes copying "downloaded RO-style sprites", and the mapping table uses Ragnarok Online class names (`novice_f`, `acolyte`, `swordman`, `thief`, `assassin`, `wizard`). `CREDITS.md` is not touched by this PR, so roughly 80 new PNGs land in `public/sprites/` with no licence entry. Please state the exact source and licence of every asset and add the rows to `CREDITS.md`; if any of it derives from Gravity's Ragnarok Online art it cannot ship. The script also reads from `tmp/sprites_source`, which is not in the repository, so the pipeline is not reproducible.

## Minor

**12. House style.** 114 added lines contain em or en dashes, which the repo forbids everywhere including code comments (the `Stop` hook blocks on them). Worst offenders: `.opencode/plans/sprite-system.md` (42), `scripts/gen_placeholders.mjs` (17), `src/render/sprites/sprite_visual.ts` (16), `src/render/sprites/sprite_shadow.ts` (11).

**13. Scope and description.** The PR body is the unmodified template: no summary, no test notes, and the only ticked box is "Tested on desktop and mobile", which is hard to square with a branch that does not typecheck. The title mentions sprites and lazy loading but the diff also contains an entire new network transport. Please split this into separate PRs: sprites, lazy GLB loading, and the HTTP fallback transport are three independent changes.

**14. Screenshots.** This is a large visual change (billboard sprites replacing 3D character visuals). Before and after screenshots on desktop and mobile are required, committed under `docs/screenshots` and linked from the PR body.

**15. `.opencode/plans/sprite-system.md`** is a tool-specific planning document. Design docs live in `docs/design/` or `docs/prd/`, and `.opencode/` is not gitignored.

## What I ran

Worktree at `ea5120f1d`. `npx tsc --noEmit` fails with 29 errors. `tests/architecture.test.ts` passes (24). `tests/localization_fixes.test.ts` passes (29, 3 skipped) after `npm run i18n:gen`. Determinism and sim purity are not violated by the `flatMap` rewrites, since order is preserved, but they should still be reverted as unrelated.

Happy to re-review once this is rebased onto `release/v0.23.0`, split up, and green. The sprite work in particular is worth landing on its own.

PR #1665, opened 2026-07-08
Improve Hollow Crypt encounter readability

loopermallee:
## Summary

- Adds readable Hollow Crypt mechanics: Sexton Marrow tank-facing cleave, Morthen hardcast, and Morthen marked-ground AoE.
- Adds lightweight party role labels (`tank`, `healer`, `damage`) through `/role`, surfaced in party frames without gating dungeon entry or rewards.
- Adds localized Hollow Crypt mechanic tips and updates parity/i18n generated status for the new deterministic state.

## Scope

- Hollow Crypt only.
- Reusable primitives are limited to forward-cone cleaves, mob hardcast interrupt recognition, and mob-sourced marked-ground telegraphs.
- Party roles are session party metadata only.

## Gameplay impact

- Tanks can face Sexton Marrow away from allies to avoid cleave splash.
- DPS can see and interrupt Morthen's Grave Rite cast.
- Players get a warning and ground marker before Grave Eruption lands, and can avoid the impact by moving out.
- Party frames can display optional role labels, but roles do not affect queueing, dungeon entry, XP, loot, or combat math.

## Files/areas changed

- Sim combat/mob mechanics: cleave arc targeting, mob hardcast cast events, interrupt handling, marked-ground AoE scheduling.
- Hollow Crypt content: Sexton Marrow and Morthen encounter data plus mechanic tips.
- Party/social API: role state, `/role` command, party/world info serialization.
- UI: party-frame role badge rendering and localized sim strings.
- Tests/parity: dungeon, cleave, social role, party-frame signature, localization, and dungeon parity coverage.

## Tests run

- PASS `npx vitest run tests/mob_cleave.test.ts tests/dungeons.test.ts tests/social.test.ts tests/party_frames.test.ts --testTimeout 20000`
- PASS `npx vitest run tests/localization_fixes.test.ts --testNamePattern "s3_registered" --testTimeout 20000`
- PASS `npx vitest run tests/parity/parity.test.ts --testNamePattern "dungeon_instances" --testTimeout 60000`
- PASS `npx vitest run tests/localization_coverage.test.ts --testTimeout 30000`
- PASS `npm run build`
- PASS `npm test -- --testTimeout 30000 tests/mob_cleave.test.ts tests/dungeons.test.ts tests/social.test.ts tests/party_frames.test.ts tests/localization_fixes.test.ts tests/localization_coverage.test.ts tests/parity/parity.test.ts --testNamePattern "Sexton|Morthen|role|party|s3_registered|dungeon_instances|class and ability translation|talent content translations"`
- ATTEMPTED `npm test` default full suite: failed in the local clean temp worktree. The run reached 856 passing files and 57 failing files; failures were dominated by default timeout pressure and jsdom admin suites failing to resolve a Vite `/@fs/...` Svelte testing helper through the symlinked dependency tree. The relevant parity/i18n issues surfaced by that attempt were fixed and revalidated by the targeted gates above.

## Screenshots or video

Party role labels

## Follow-ups explicitly deferred

- No dungeon journal.
- No role-composition enforcement.
- No sweep across every dungeon.
- No broader party role management UI beyond the lightweight `/role` command.
- No additional encounter media or authored art assets.

## Risk notes

- Determinism: Morthen's marked-ground location uses server-authoritative target position and existing deterministic ground AoE timing; no new random draws are introduced for the telegraph. Dungeon parity was reminted for the new Morthen timers.
- Persistence: party roles are not persisted and clear when a member leaves the party.
- Balance: new damage windows are scoped to Hollow Crypt with modest values and existing heroic mechanic multipliers.


EnriqueGF:
Thanks for this one :) The mechanics are genuinely well shaped, and the new Morthen tests are the good kind (they assert avoidance and interrupt outcomes, not just pins). Base branch is correct: release/v0.23.0 is the latest release branch. A few things need to land before this is mergeable.

## Blocking

**1. `npx tsc --noEmit` fails on this branch (6 errors). CI will be red.**

I ran it on `origin/release/v0.23.0` (exit 0, clean) and on this head (exit 2), so these are introduced here. `npm run build` passing does not cover this: vite/esbuild strip types without checking them.

- `src/net/online.ts:911` and `src/main.ts:4239`: `ClientWorld` does not implement `IWorld.setPartyRole`, so it no longer satisfies `IWorld`.
- `src/net/online.ts:772`: the ClientWorld entity factory is missing the new `groundTelegraphTimer` field required by `Entity`.
- `tests/world_api_parity.test.ts:1049`: `setPartyRole` is not in the `FACET_PARTY` corpus, so the `_ExhaustParty` exhaustiveness assertion fails.
- `tests/entity_roster.test.ts:55` and `tests/sim_context.test.ts:227`: the `SimContextHost` stubs are missing `setPartyRole`.

Note the runtime suites all pass (I ran architecture, world_api_parity, parity/, mob_cleave, dungeons, social, party_frames, localization_fixes: all green). Vitest does not typecheck, which is exactly why the tri-host seam broke silently here.

**2. `IWorld.setPartyRole` is unreachable surface, and it breaks the IWorld invariant.**

Nothing in `src/ui/` or `src/game/` ever calls `world.setPartyRole()`. `/role` is handled entirely inside the sim chat path (`src/sim/social/chat.ts`), which already runs server-side online, so the role works without the IWorld method. As written the method only adds a member that `ClientWorld` cannot satisfy. Please either drop it from `IWorldParty` and keep `/role` a pure chat command, or implement it in both worlds with a real wire command. Adding to `IWorld` without implementing it in `ClientWorld` is the one thing the seam forbids.

**3. Cleave changes behavior for six existing mobs, outside the stated scope.**

`src/sim/mob/mob_swing.ts:115-117` now measures the splash radius from `mob.pos` instead of `target.pos`, and gates every cleave on a forward cone that defaults to `DEFAULT_CLEAVE_HALF_ANGLE` (`mob_swing.ts:35`) when a template does not author `angle`. That silently retunes shipped content:

- `src/sim/content/zone2.ts:220` (Tidal Sweep)
- `src/sim/content/zone3.ts:545` (Grave-Cleaver)
- `src/sim/content/zone3.ts:619` (Deathstalker Cleave)
- `src/sim/content/dungeons.ts:242` (Reaping Arc)
- `src/sim/content/delves/mobs.ts:104` (Stone Sweep, radius 3.5)
- `src/sim/content/delves/mobs.ts:290` (Silt Cleave, radius 3.5)

The tell is in the diff itself: the two pre-existing tests in `tests/mob_cleave.test.ts` had to be rewritten (players repositioned, `olen.facing = 0` added) to keep passing. The delve mobs are the worst hit, since a 3.5yd radius re-anchored from the target to the mob plus a 120 degree cone is a large effective nerf. Please make the cone opt-in (apply it only when `angle` is authored, keeping the old target-anchored radius otherwise), or state the retune explicitly and add coverage for the legacy templates.

**4. The interrupt change makes a raid boss interruptible, also outside scope.**

`src/sim/combat/effect_dispatch.ts:280-286` recognizes any mob `bigCast` as interruptible. That is not just Morthen: `src/sim/content/zone3.ts:913` (Thunzharr, Stormcall, `school: 'nature'`, 3.5s cast) was previously immune, per the comment the diff replaced ("scripted pseudo-casts resolve to nothing and are immune by design"). There is also no way to author an uninterruptible scripted hardcast, because `uninterruptible` is read only off `interruptedDef` and never off `scriptedBigCast`. Please scope this (an explicit `interruptible: true` on the template's `bigCast`), and add a test asserting Thunzharr's intended behavior either way.

## Non-blocking

**5. Heroic scaling deviates from the house convention.**

`src/sim/mob/locomotion.ts:535` bakes `mechanicDamageMult` into the scheduled `min`/`max` before the draw. `aoePulse`, `stomp` and `bigCast` all multiply after `ctx.rng.range(...)`, with an explicit comment saying so ("Heroic scaling multiplies AFTER the draw so the rng stream is identical across difficulties"). The draw count is unchanged so determinism holds, but the rounding differs from every sibling mechanic. Worth aligning.

**6. Dead matcher rule.**

The `Party role set to (.+)\.` rule added to `RULES` in `src/ui/sim_i18n.ts` is unreachable: `localizeSimText` checks the `EXACT` table first, and all three real values are already in `baseEnTable`. Its fallback branch would also emit raw unlocalized English. I verified the rest of the i18n path resolves correctly (the tips, the warning, and the role usage error all match through `EXACT`), so this is just cleanup.

**7. Screenshots.**

The repo asks for before/after on desktop and mobile, committed under `docs/screenshots`. This ships one after-only desktop PNG under `docs/pr-assets/`, hotlinked from a fork raw URL that will break if the branch is deleted. `src/styles/hud.mobile.css` is touched, so a mobile shot is required.

## What I verified

Base branch (`release/v0.23.0`, latest); `tsc` on base vs head; `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/parity/`, `tests/localization_fixes.test.ts`, `tests/mob_cleave.test.ts`, `tests/dungeons.test.ts`, `tests/social.test.ts`, `tests/party_frames.test.ts` (all green at runtime); the sim i18n matcher against the real emitted strings; sim purity (no `Math.random`/`Date.now` added, the new ground AoE reuses `ctx.rng` and mirrors the `aoePulse`/`stomp` player-iteration order); `pid === entityId` so the new `party.roles` map is keyed consistently with `raidGroups`.

Happy to re-review once the typecheck is green and 3 and 4 are scoped.

PR #1666, opened 2026-07-08
feat(delves): add companion roles and shortcuts

loopermallee:
## Summary
- Adds active-run companion role choice for delves: healer, tank, and damage.
- Adds two optional profession shortcuts in Collapsed Reliquary modules: Mining cracked wall and Herbalism toxic growth.
- Surfaces role/shortcut feedback through localized UI strings and procedural props.

## Scope
- Base: `release/v0.23.0`; this PR is not stacked on another prompt PR.
- Companion roles are selected at delve start and stored only on the active run.
- Existing companions still default to healer, preserving current behavior.
- Shortcut checks are server-authoritative and use existing gathering proficiency state, not new profession materials.
- Exactly two shortcuts are included: Mining and Herbalism.

## Gameplay impact
- Healer remains the default support role with existing rank benefits.
- Tank companion gains durability, lower damage, and a deterministic taunt cadence to help fragile solo builds without replacing dungeon tanks.
- Damage companion hits harder and gives up periodic healing, speeding runs while increasing risk.
- Mining and Herbalism shortcuts save interaction time or reduce friction, but non-profession fallback interactions still open the path.
- Shortcuts do not gate delve completion or exclusive loot.

## Files/areas changed
- `src/sim/delves/*`, `src/sim/types.ts`, `src/sim/sim.ts`, and server/network/world API plumbing for role selection and shortcut interactions.
- `src/sim/content/delves/collapsed_reliquary.ts` for the two shortcut placements.
- `src/ui/hud.ts`, UI styles, i18n catalog/locales/generated tables, and sim log localization for board/tracker/shortcut feedback.
- `src/render/*` for nameplates, sparkle behavior, and procedural shortcut props.
- Delve companion, delve shortcut, snapshot, i18n, profession, lockpick, and parity tests.

## Tests run
- `git diff --check`
- `git diff --cached --check`
- `npx vitest run tests/delve_companion.test.ts tests/delves.test.ts tests/snapshots.test.ts` — passed, 261 tests.
- `npx vitest run tests/delves_runs.test.ts tests/professions_gathering.test.ts tests/professions_tools.test.ts tests/world_api_parity.test.ts` — passed, 278 tests.
- `npx vitest run tests/lockpick_command.test.ts tests/lockpick_session.test.ts tests/lockpick_step.test.ts tests/lockpick_bountiful_jam.test.ts tests/lockpick_by_sight.test.ts tests/lockpick_col4_heroic.test.ts tests/lockpick_gen.test.ts tests/lockpick_hud_sync.test.ts tests/lockpick_panel.test.ts tests/lockpick_timeout.test.ts tests/lockpick_window.test.ts` — passed, 102 tests.
- `npx vitest run tests/i18n_completeness.test.ts tests/localization_coverage.test.ts tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts` — passed, 82 tests, 4 skipped.
- `npm run build` — passed. Existing Vite dynamic-import and chunk-size warnings remain.
- `npm test` — attempted and interrupted after reproducing broad unrelated failures/timeouts across loot, PvP, social, fiesta, raid, mail, threat, security, parity, and Vale Cup suites, plus sandbox/mocked DB/RPC failures. Focused delve/profession/i18n/lockpick/build gates above passed.

## Screenshots or video
- Not attached from this CLI environment. UI/visual changes are limited to localized role controls, tracker prompts, and procedural in-engine shortcut props; no binary media assets are committed.

## Follow-ups explicitly deferred
- More profession shortcut types beyond Mining and Herbalism.
- Telemetry/playtest-driven tuning for exact tank/damage role multipliers.
- Persistent saved role preference outside the active delve run.
- Richer review media or committed media assets.

## Risk notes
- Determinism: role behavior uses existing fixed-step sim timing and introduces no `Math.random`, wall-clock time, or browser APIs in sim logic.
- Persistence: role choice is active-run state only; companion rank persistence is untouched and covered by tests.
- Balance: tank/damage multipliers are intentionally conservative, and shortcuts are efficiency/flavor only with fallback paths.


EnriqueGF:
Thanks for this one :) The delve role/shortcut design is well thought through, and I want to call out what is genuinely solid before the findings: the base branch is correct (`release/v0.23.0`, the latest release), server authority is real (the role is allowlisted at `server/game.ts:3672` and re-normalized in the sim), the wire is properly mirrored on both `Sim` and `ClientWorld`, and the i18n work is done right, including the M16 non-Latin fills for the new wordy strings. I verified all of that against the code rather than the description.

Verdict: **changes requested**. Details below, ordered by severity.

## Blocking

**1. The change introduces 4 lint errors, so `ci:changed` and the pre-push floor are red.**
All four are `assist/source/organizeImports`, in exactly the four files where imports were added:

- `server/game.ts:24`
- `src/net/online.ts:34`
- `src/sim/delves/runs.ts:22`
- `src/ui/hud.ts:89`

I confirmed these are introduced, not pre-existing: checking out the base versions of those same four files yields zero `organizeImports` errors, and HEAD yields four. `npm run build` does not run Biome, which is why this was not caught. Fix with `npx @biomejs/biome check --write` on those four files only (do not run a whole-repo write).

**2. `herbalism_toxic_growth` has no collision, so the herbalism shortcut gates nothing.**
In `src/sim/delves/runs.ts:272` the solid-radius branch was extended for the mining node only:

```ts
else if (state.kind === 'destructible_wall' || state.kind === 'mining_cracked_wall')
  solidR = obj.hp > 0 ? DELVE_WALL_SOLID_R : 0;
```

`herbalism_toxic_growth` never matches any branch in `clampDelveDoors`, so `solidR` stays 0 and the loop skips it. The node is created with `startOpen: false`, `maxHp: 40`, and a fallback line that reads "You beat back the toxic growth", all of which imply it is meant to obstruct. As written a player walks straight through it and the whole herbalism path is decorative. Either add it to the solid list next to its mining sibling, or state explicitly that it is intended to be non-blocking and adjust the copy.

**3. The damage role's threat reduction is dead configuration.**
`src/sim/delves/companion.ts:73` declares `threatMultiplier: 0.85` for `damage`, but the only consumer is guarded on `> 1`:

```ts
if (combatTarget.kind === 'mob' && profile.threatMultiplier > 1) {   // companion.ts:214
```

So the damage companion generates full threat, not reduced threat. Either implement the sub-1.0 case or remove the field so the profile does not advertise behavior that does not exist. The PR body lists reduced threat as part of the damage role's trade-off, so today the gameplay does not match the stated design.

**4. The "deterministic taunt cadence" is the headline mechanic and is not pinned by any test.**
`tests/delve_companion.test.ts:87` forces `companionTauntTimer = 0` and asserts a single taunt lands. Nothing asserts the interval literal (`tauntInterval: 8`, `companion.ts:67`), that a second taunt is suppressed while the timer is above zero, or that it re-fires after the interval elapses. A regression that taunts every tick, or never resets the timer, passes this suite untouched. Please assert the timer equals the literal `8` after the taunt, that an immediate re-run does not re-taunt, and that it re-taunts once the timer drains.

Also missing decisive assertions, all cheap to add:
- tank `damageMult: 0.85` (only the damage role's increase is asserted)
- tank `threatMultiplier: 1.5` (the taunt test sets `swingTimer = 999`, so no swing ever runs)
- tank `healMultiplier: 0.5`
- damage `healthMult: 0.9` / `armorMult: 0.85`
- the cross-profession negative: a mining-proficient player at the herbalism node must get the fallback, not an instant open
- `normalizeDelveCompanionRole('foo')` returns `'healer'`

**5. Mergeability and gate.** The PR is a draft, is `CONFLICTING` against `release/v0.23.0`, and sits 3 commits behind it. The body also says `npm test` was interrupted after broad failures. Please merge the base in, then run `npm run gate` and report the result, rather than a subset of suites.

**6. Screenshots.** This is a visual change (the role picker in the delve board, `src/styles/components.css` and `hud.css`, nameplates, and new procedural props in `src/render/delve_props.ts`). The repo requires before/after screenshots on visual PRs, desktop and mobile where relevant, committed under `docs/screenshots` and linked from the body.

## Non-blocking, worth fixing

- `src/sim/delves/runs.ts:380`: `pidOrRole?: number | DelveCompanionRole` is an overloaded positional parameter disambiguated by `typeof`. It is easy to pass a role where a pid was meant and get silent fallback. Prefer an options object or a distinct parameter.
- Dual source of truth: `run.companionRole` and `run.companion.role` are both written (`sim.ts:6942`, `runs.ts:444`) and read with `run.companion?.role ?? run.companionRole` (`runs.ts:1758`, `entity_roster.ts:212`). One of them should be the owner.
- `src/ui/hud.ts:7458`: `let bestD = INTERACT_RANGE + 1` is a sentinel that silently doubles as the range gate, and there is no explicit final range check. It happens to equal `DELVE_INTERACT_RANGE` (6) today purely because `INTERACT_RANGE` is 5. If either constant moves, the prompt range diverges from what the server accepts. Gate on the delve range explicitly.
- `hud.ts:7448` `delveCompanionRoleLabel` reimplements `normalizeDelveCompanionRole` inline. Import the sim helper instead.
- `nearbyDelveShortcutPrompt` does a full `sim.entities` scan on every delve-tracker refresh. It is cheap today given interest scoping, but a `templateId` early-out over a smaller set, or caching the two object ids on run start, would keep it off the per-frame path.
- The skill check is `skill > 0` (`runs.ts:1347`), a presence check, not a level threshold, and a player with no skill still opens the node in 3 interacts with no cooldown, all of which can be sent in a single tick. That is fine if intended, but the body describes it as gated on proficiency level.
- `applyDelveCompanionRoleStats` multiplies stats in place and is not idempotent. It has one call site today, but a second call would compound. Worth a guard or a comment.

## Verified clean

`npx tsc --noEmit` passes. `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` pass (53 passed, 3 skipped). `tests/delve_companion.test.ts`, `tests/delves.test.ts`, `tests/delves_runs.test.ts`, `tests/snapshots.test.ts`, `tests/world_api_parity.test.ts` pass (491 passed). No `Math.random`, `Date.now`, or `performance.now` added to `src/sim/`; the taunt cadence is tick-based and the new role branches do not perturb rng draw order for the existing healer default. No new em or en dashes and no emojis on added lines. `IWorld` was extended first and implemented in both worlds. The locale overlay edits are the sanctioned M16 non-Latin fills, not hand-translations of Latin locales.

Happy to re-review once the lint errors, the herbalism collision, the dead threat multiplier, and the taunt-cadence test are addressed.

PR #1685, opened 2026-07-09
feat(sim): add a real walkable tunnel between Eastbrook Vale and Mirefen Marsh

jgyy:
## Summary

Adds `vale_marsh_ridge_tunnel`: a real, walkable through-tunnel connecting **Eastbrook Vale (zone1)** to **Mirefen Marsh (zone2)** on the **east side of both zones** (`x=128`), burrowing under the ridge that straddles the zone boundary at `z=180`.

This is a genuine gameplay traversal, not just carved geometry: players (and pathfinding) can actually walk in one mouth, descend under the ridge, and come out the other mouth in the neighboring zone.

**Update 11 (current):** a round of feedback on the mound's silhouette and two chase-camera bugs. The mound's peak still read as a pointed tip/cone rather than a rounded hill, and a review of the exterior close-up shots turned up a real camera bug behind the flat white void screenshots: near a mound the requested chase-cam position (plus the player's own ground-snap physics) could land inside the mound's exterior slope instead of on top of it, and deep inside the tunnel the camera's floor clamp used the far-overhead open-world surface, snapping the view into a near-vertical look-down. Both fixed, plus the interior S-curve's lateral swing roughly doubled (`+-6yd` to `+-10..12yd`). See "What changed in this round (round 11)" below. All 53 screenshots retaken (the two closest-to-the-mouth "closeup" shots moved a few yards further out so the player stands cleanly on the mound's exterior grass, not right at the carved threshold); the previous round's screenshots were removed from the PR-assets branch.

**Update 10:** the round-9 mound (still a hard-edged solid shape, just repositioned/reshaped) still read as a separate structure - not one with the terrain the way the world's real border mountains are - and still had grass/foliage floating around the entrance. Root-caused and fixed properly this time: the mound is now an additive height bump on the same terrain function the real mountain ridges use, so it is provably the same ground, not a second object. See "What changed in this round (round 10)" below. All 53 screenshots retaken; the previous round's screenshots were removed from the PR-assets branch.

**Update 9:** a further round of feedback on the live client reported the round-8 backstop fix was not enough - the entrance mound still read as "just a cube floating", not terrain that blends in; the cave mouth needed to look like a natural hill with a vertical doorway, not a hole to jump on; the entrance sat too close to a mob spawn area; and the tunnel needed to be deeper. All fixed; see "What changed in this round (round 9)" below. All 53 screenshots retaken; the previous round's screenshots were removed from the PR-assets branch.

**Update 8:** a round of review feedback reported four problems with the live client: (1) the open-world sky read as solid rock over a wide area near the tunnel, (2) the entrance mound floated off the ground and clashed with surrounding trees/rocks, (3) the cave mouths needed to always be the same terrain entity, never a separate piece, and (4) the terrain and world map needed a natural sign of each cave opening. All four are fixed; see "What changed in this round (round 8)" below. All 33 screenshots retaken; the previous round's screenshots were removed from the PR-assets branch.

**Update 2:** a round of review feedback asked for the entrance to look natural, the tunnel doubled in size, real textures, the character visible inside, the tunnel moved to the east side of both zones, the terrain to actually deform into the cave (not a separate overlay), and 20+ screenshots. All of that is done; see "What changed in this round" below.

**Update 3:** a further round asked that the camera never be able to show what's below/outside the tunnel from any angle (not just in screenshots), the tunnel diameter and depth increased further, more decoration, and the cave mouths to sit clearly above the surrounding ground while still reading as a natural hillside opening.

**Update 4:** a request to better manage the chase camera in small enclosed spaces: while the player is inside the tunnel, the camera's own max zoom-out distance is now reduced so it never asks for a position that far past the tunnel's own walls in the first place (previously the ordinary 22yd max only ever got pulled back in reactively by the existing occlusion system, once the requested position already collided). All 20 screenshots below are retaken against this change; the previous round's screenshots were removed from the PR-assets branch.

**Update 5:** a round of feedback on the cave's actual shape: the cross-section read as a uniform circle everywhere (a pipe bore), not a real cave (arched top, flatter floor); the diameter needed to be slightly larger and the tunnel dug way deeper into the ground; and each mouth needed to read as a vertical, upright opening (a doorway you walk into) rather than a hole lying flat on the ground. The chase camera's enclosed-space zoom cap was also tightened further.

**Update 6:** a further pass on the mouths and grade: the upright-opening fix from round 5 wasn't enough on its own (a close, straight-on view into the mouth still read as looking down a shaft), the tunnel's steepness needed a hard 30 degree cap everywhere (not just near the mouths), and the entrances needed to be cut into actual protruding mounds, not just a dip in the existing ridge slope. See "What changed in this round (round 6)" below; all 20 screenshots retaken and the previous round's removed from the PR-assets branch.

**Update 7:** a round of feedback against a real cave-entrance photo reference: both mouths needed to read as organic (irregular rock, grassy hillside, overhanging foliage, a visible approach path), the interior needed to never show grass on the floor/walls/ceiling, the tunnel needed to be pushed noticeably deeper, and the path needed to S-curve rather than hold a constant x. See "What changed in this round (round 7)" below; all 24 screenshots retaken and the previous round's removed from the PR-assets branch.

## Why this needed more than a content entry

The repo already had one voxel-carved tunnel (`vale_kobold_warren`, a dead-end fixture), but `voxel.ts`'s own header flagged it as unfinished:

> "Wiring real tunnel content into the renderer, and 3D collision/pathfinding through it (today's colliders.ts/pathfind.ts are still column-based, one height per x,z), are deliberate follow-up phases."

This PR does that follow-up wiring, scoped tightly:

1. **Content** (`src/sim/content/tunnels.ts`): a new `TunnelVolume` on the east side of both zones (`x=128`), waypoints crossing `z=180`, clear of the causeway, every lake, and every POI. Each waypoint's radius is roughly double a first-pass size, a real walk-in cave.
2. **Traversal** (`src/sim/tunnel_traversal.ts`): bridges the 3D voxel density field and the column-based (one height per x,z) collision model everywhere else in the sim. Pure function of `(x, y, z, seed)` plus the authored content, no RNG.
3. **Movement** (`src/sim/player_motion.ts`): rides the tunnel's local floor instead of the surface while inside it, bypassing the surface slope/wall gates underground.
4. **Pathfinding** (`src/sim/pathfind.ts`): height lookups do the same, so `findPath` can route through the tunnel as a real lower alternate path.
5. **Rendering, single deforming surface** (`src/render/terrain.ts` + `src/render/tunnel_overlay.ts`): explained below.

## What changed in this round (round 11, current)

- **"Mound peak reads as a pointed tip, not a rounded hill"**: `moundHeightBump`'s Gaussian falloff (`src/sim/voxel.ts`) was narrow enough (`sigma = moundRadius * 0.75`) plus the hand-authored angular "irregularity" sine lobes (`moundIrregularity`, amplitude coefficients `0.16`/`0.09`/`0.05`) had enough amplitude that the two effects together let the mound's silhouette read as a cone/volcano shape with a visible crease at the peak, especially from the zone2 wide establishing angle. The irregularity coefficients were roughly halved (`0.10`/`0.055`/`0.03`) so the knoll still reads as organic and non-circular but the peak itself stays smoothly rounded; `sigma` was left at its original width since a wider falloff would have pushed the mound's own bump past the point `tests/voxel.test.ts` pins as "reads as ordinary ground" on the mound's exterior slope.
- **"Camera clips into the mound / a flat white void is visible below the terrain near a mouth"**: root-caused to two separate bugs, both in the ordinary (non-editorCam) chase-camera path (`updateCamera` in `src/render/renderer.ts`) and the player's own ground-snap physics (`src/sim/player_motion.ts`). (1) `player_motion.ts`'s vertical step resolved the player's standing height against plain `groundHeight`, which has no idea a tunnel entrance mound additively raises the ground there (`moundHeightBump`/`groundHeightWithMounds`, `src/sim/voxel.ts`) - so a player (and the camera anchored to them) standing on a mound's grassy exterior slope could end up partially embedded in it instead of standing on top, reading as an extreme point-blank smear of terrain filling the whole frame. Fixed by resolving against `groundHeightWithMounds` in both of `player_motion.ts`'s ground-snap call sites. (2) The chase camera's own Y clamp (`renderer.ts`) had the same plain-`groundHeight` blind spot for its own sampled position near a mound. Added a `MOUND_EXTERIOR_MAX_CAM_DIST` (9yd) camDist cap while standing near any mound's decayed Gaussian skirt (`src/render/camera_zoom_limit.ts`), on top of the existing tunnel-interior cap, and switched the camera's own floor clamp to the mound-aware height.
- **"Camera view locks into looking almost straight down while inside the tunnel"**: a second, related bug in the same clamp - while deep inside the passage (crest floor around `y=-52` to `-57`), the camera's own sampled `(x, z)` column often still sits within the tunnel's bounding-box footprint even though the camera itself is far from the tunnel's actual floor/ceiling band; clamping that column's Y against the far-overhead open-world surface (the marsh ridge crest sits some 80+yd above the passage, per this file's own coordinate comments) snapped the camera way up above the player, reading as the view locking into a near-vertical look-down. Fixed with a new `cameraFloorAt` helper (`camera_zoom_limit.ts`) that only substitutes the tunnel's own local floor when the camera's own position genuinely sits within that tunnel's floor/ceiling span (`tunnelSpanAt`, the same check `player_motion.ts` uses to decide "am I actually inside this passage"), not merely whenever the sampled column's bounding box happens to intersect the tunnel's footprint - so a wide, steep-pitch establishing shot from up on the open ridge still reads against the real outdoor surface. New `tests/camera_zoom_limit.test.ts` cases pin both directions: the tunnel floor is used when the camera is genuinely down in the passage, and the ordinary surface is used when it is up in open air over the same `(x, z)` column.
- **"Widen the tunnel's interior S-curve"**: the lateral swing of the interior waypoints (`content/tunnels.ts`) was roughly doubled from `+-6yd` to `+-10..12yd` (the waypoint nearest each mouth kept at `+10` rather than `+12` so its carved capsule still stays clear of the mound's own exterior slope right at the doorway), verified clear of `vale_kobold_warren` (entirely within `x=60..84`) and every zone1/zone2 camp/POI/road near the new `x=116..140` range.
- Two of the "closeup" exterior shots (`04_mouth_a_closeup`/`05_mouth_a_closeup_side_angle`/`06_mouth_b_closeup`) were only 6yd from their mouth's own 6.5yd carve radius - close enough to sit right at (or just past) the doorway threshold, where the player's y legitimately resolves against the tunnel's own interior floor rather than the mound's exterior surface. Moved each a few yards further out (still close, but cleanly on the exterior grass) so they show the intended "approaching the mound" framing rather than standing in the doorway.
- All 53 screenshots retaken (same shot list as round 10: `scripts/tunnel_screenshot.mjs`'s waypoint/floor table updated to match the widened S-curve; no camera-lock or mound-clip artifacts in any shot; the character remains visible in every exterior AND interior shot); the previous round's screenshots were removed from the PR-assets branch.

## What changed in this round (round 10)

- **"Cave entrance is just a cube floating, not one with the terrain like the border mountains"**: root-caused properly this time. The mound (`TunnelWaypoint.mound`) was a separate SOLID shape - first a hard-edged ellipsoid (`moundSolidAmount`), then round 9's per-tunnel-scoped version - always `min()`'d against the ordinary terrain density as its own object with a boundary. Even reshaped or repositioned, a solid shape with an edge reads as a boulder dropped onto the ground, not a rise in the land. Replaced with an ADDITIVE Gaussian height bump on the SAME heightfield the rest of the terrain uses (`moundHeightBump` / `groundHeightWithMounds` in `voxel.ts`) - this is the identical technique `world.ts`'s own mountain ridge walls and the world rim already use (`RIDGE_HEIGHT * gaussian profile`, added straight onto `terrainHeight`, see `RIDGE_SIGMA`/`RIDGE_HEIGHT`). The knoll now rises and falls continuously out of the surrounding ground with no edge anywhere, on both entrances - provably the same terrain, not a second object standing on it.
- **"No floating stuff around the mound/mouth"**: two separate sources, both fixed. (1) The hanging-branch leaf-clump decorations (`addMouthFoliage`) were anchored to an offset from the doorway's own geometry (first the carved ellipsoid, later the mound's crest height) rather than the real local ground at each clump's own position - close enough at typical distance, but reading as leaf blobs floating in the sky from partway down the tunnel looking out. Removed entirely; the mound's own grass texture and the path patch (now anchored to `groundHeightWithMounds`) already read as a dressed entrance without them. (2) Grass blade billboards (`foliage.ts`'s `buildGrassRing`) are anchored to the plain `terrainHeight`, with no tunnel/mound awareness at all, so tufts kept spawning inside the carved-open passage near a mouth - a dense swarm of blades floating in mid-air, visible looking out through either opening. Added the same `isNearTunnelMound` exclusion already used for trees/rocks/bushes.
- **"No grass floating on entrance, remove that"**: once the grass exclusion above was in place, a second issue surfaced - the exclusion radius (`moundRadius * 2`, tuned in round 9) still left real forest close enough to a mound (as near as 29yd) that an oak's far-LOD canopy (rendered trunk-less by design - an existing, intentional performance simplification, `farTrunkProxy` in `foliage.ts`) was visible poking above the mound's own silhouette from just inside the tunnel mouth, reading as leafy clumps floating in the sky with nothing under them. Widened the clearance to `moundRadius * 3.5` (verified against the live decoration scatter: the closest tree to either mouth is now 50yd away, was 29yd), clearing a real, walkable, forest-free approach in front of both mouths.
- New tests: `tests/voxel.test.ts`'s mound-solidity probes retuned for the new Gaussian falloff shape (same pass/fail intent as before - a different, gradual, decay curve instead of a hard-edged sphere means the old probe's exact radius/height combination needed re-checking, and one of them happened to land at a spot where local base-terrain noise flipped the result, unrelated to the bump math itself).
- All 53 screenshots retaken confirming the mound reads as one continuous grassy hill in the same style as the world's real border mountains/ridges, and no floating grass or foliage anywhere near either entrance; the previous round's screenshots were removed from the PR-assets branch.

## What changed in this round (round 9)

- **"Cave entrance is just a cube floating, not blending with the terrain"**: root-caused precisely. Round 8's per-tunnel backstop (a box of six opaque planes hugging that one tunnel's own `tunnelBounds`) still had to poke its wall/ceiling planes up out of the ground wherever the surrounding ambient terrain sat lower than the tunnel/mound's own height there - the marsh and vale hillsides roll well below the mound's crest, so the box's edges stuck out as a giant flat rock slab. Removed the enclosing box entirely (`tunnel_overlay.ts`). The terrain patch above (built from the identical `voxelDensity` field: ordinary heightfield, every mound, every tunnel carve, all one isosurface) is already watertight and continuous with the ordinary `terrain.ts` chunks just outside the excluded footprint - there is no lateral edge for a wall to guard, and open sky above ground was always correct, never a bug. Replaced with a single deep, invisible floor cap (tight to the tunnel's own footprint, sunk well below its lowest point) as cheap insurance against a Surface Nets seam - a case ordinary player collision (`tunnelFloorAt`) never lets the player reach anyway.
- **"Vertical cave entrance like a door, not a hole to jump on"**: the mouth cross-section's `floorScale` (0.85) still put the threshold floor well below the ambient hillside, so standing at the doorway and looking down (or from the side) read as looking into a round pit. Mouth waypoints now use a tighter, taller profile (`radius: 6.5`, `archScale: 3.0`, `floorScale: 0.45`), and each mouth's own `y` is tuned so `y - radius*floorScale` lands almost exactly at that mouth's own ambient `terrainHeight` (see the comment in `content/tunnels.ts`) - a real, flush, walk-through doorway at ground level, not a dip you fall into. `tests/player_motion.test.ts`'s mouth-floor assertion updated to pin "close to the surface" (was "well below it"), matching the intentional flush threshold.
- **"No grass or trees should clutter around the entrance"**: re-verified after every position change below - `world.ts`'s `isNearTunnelMound` exclusion (round 8) is keyed on the mound's own `(x, z)` and radius, so it followed the shift automatically; `tests/fixes.test.ts` still asserts zero decorations inside either mound's clearance radius.
- **"If entrance is at mob spawn area, shift the cave entrance away"**: the old mouth at `x=110` sat close enough that its mound + decoration clearance footprint (14yd mound radius + 6yd clearance = 20yd) overlapped the Fallen Chapel undead camp cluster around `(88-92, 90)` (center-to-center distance ~19yd, well inside the combined radii). Both mouths shifted east to `x=128`, clearing that camp (and every other camp/POI/road/lake) by a wide, verified margin (~43yd center-to-center, ~25-39yd clear of the nearest camp edge).
- **"Make tunnel depth deeper"**: crest deepened from `y=-37` to `y=-52.5` (the ridge surface itself sits at `y=+35.8` at the crest, so the tunnel is now ~88yd of solid rock below it, versus ~73yd before). To keep every segment under the 30 degree grade cap given the extra depth, both mouths moved further out from the `z=180` zone boundary to `z=66`/`z=294` (was `z=96`/`z=264`).
- Every test referencing this tunnel's literal coordinates updated (`tunnel_traversal`/`player_motion`/`pathfind`/`camera_zoom_limit`), and the screenshot tour's floor-height table and waypoint list recomputed from the live sim for the new geometry (`scripts/tunnel_screenshot.mjs`).
- All 53 screenshots retaken (30 more than the round-8 set): straight-up sky checks at four points along the now-longer route, a wide shot across the open field between the mounds from both sides, an overhead pass and a frontal door view at each mouth, a ground-level clear-approach shot at each mouth, a wide crest shot emphasizing the extra depth, both map markers, and 28 interior shots spanning the whole (now longer) descent/ascent instead of just the crest neighborhood; the previous round's screenshots were removed from the PR-assets branch.

## What changed in this round (round 8)

- **"The sky is all rocky in the open world"**: root-caused to `tunnel_overlay.ts`'s backstop safety planes (floor/ceiling/four walls, added as defense-in-depth against a Surface Nets meshing seam). They used to be built as **one box unioning every excluded terrain chunk across every tunnel in the level** - since `vale_kobold_warren` (x~60-84) and `vale_marsh_ridge_tunnel` (x~110) are far apart, that merged box spanned roughly `x=60..124, z=96..268`, and the ceiling plane sat well above the highest terrain point across that *entire* span. Standing anywhere in that wide stretch of open world and looking up hit the opaque ceiling plane instead of the sky. Backstops are now built **per tunnel**, sized tightly from that one tunnel's own `tunnelBounds` (`src/render/tunnel_overlay.ts`), so they hug only the actual carved volume and never surface above open ground.
- **"The entrance mound floats off the ground, clashing with surrounding objects"**: same family of bug, different symptom - the old backstop's wall planes, built from that same oversized merged box, poked up out of the ground far from either mouth, reading as stray rock slabs. Fixed by the same per-tunnel scoping above. Separately, ordinary tree/rock scatter (`generateDecorations`, `src/sim/world.ts`) and the bush/fern/mushroom dressing scatter (`src/render/foliage.ts`) are both anchored to the plain `(x, z)` heightfield, which has no idea a tunnel mound additively raises the ground there (`moundSolidAmount` in `src/sim/voxel.ts`) - so decorations kept spawning at the mound's old, lower ground level, reading as trees/rocks buried in (or floating out of) the entrance knoll. Both scatters now skip any point within a mound's own clearance radius (`isNearTunnelMound`, exported from `world.ts`), clearing the approach on both sides of both mouths.
- **"Cave mouth must be the same entity as the terrain, never a separate piece"**: this was already true at the simulation level - `voxelDensity` (`src/sim/voxel.ts`) combines the ordinary heightfield, every mound, and every tunnel carve into one continuous density function, and `tunnel_overlay.ts` meshes that as a single `Mesh` (`tunnel-overlay-patch`), never a second layer over untouched terrain. What actually read as "separate" was the oversized backstop described above; with it correctly scoped, the mound, doorway, and surrounding hillside are visibly one deforming surface with nothing else superimposed on it.
- **"Natural signs in the terrain and map for the cave openings"**: the 3D terrain-level signage (the mound itself, the worn dirt path, the framing hanging foliage - `addMouthFoliage` in `tunnel_overlay.ts`) already existed from round 7 and is preserved. New this round: a **world-map marker**. `src/ui/map_tunnel_portals.ts` (a new pure module, mirroring the existing `map_dungeon_portals.ts`) returns each tunnel's mound waypoints as findable mouths; `map_window_view.ts`/`map_window_painter.ts` draw a rock-colored dot plus a generic "Cave" label at each, the same visual language as the existing dungeon-portal dot+name, so a mouth is now visible on the map without having to walk the whole ridge.
- New tests: `tests/fixes.test.ts` asserts no decoration ever spawns inside a tunnel's mound clearance radius; `tests/map_tunnel_portals.test.ts` covers the new pure portal module (mound-only, zone-band filtering, no false positives on a mound-less tunnel).
- All 33 screenshots retaken (9 new: two straight-up sky checks well clear of either mound, a wide shot across the open field between the two mounds, an overhead pass at each mound, a ground-level shot of each mouth's now-clear approach, and two world-map screenshots showing the new Cave marker); the previous round's screenshots were removed from the PR-assets branch.

## What changed in this round (review feedback)

- **"No entrance visible, no cave at all"** and **"grey pattern over the grass"**: both were the same root cause. The renderer used to build the tunnel as a *second, separate mesh* layered on top of the ordinary, unmodified terrain. Fixed properly this time: `terrain.ts` now has a `chunkNearAnyTunnel` check (`src/sim/voxel.ts`) that skips building the classic heightfield chunk anywhere a tunnel's footprint touches, leaving a real hole. `tunnel_overlay.ts` fills exactly that hole with **one** voxel-meshed patch built from `voxelDensity` (terrain and tunnel combined), on the identical 60yd chunk grid `terrain.ts` excludes. The ground now visibly and naturally deforms down into a cave mouth as a single continuous isosurface, textured with the same triplanar grass/rock blend as the production terrain, instead of a flat-colored rock knoll sitting on top of unmodified grass.
- **"Double the inner diameter"**: every waypoint radius in `content/tunnels.ts` roughly doubled.
- **"Add proper textures inside"**: `tunnel_overlay.ts`'s material now uses the same real PBR grass/rock textures as the production terrain (triplanar-projected, blended by slope), not a flat color.
- **"Character not visible in the tunnel screenshots"**: the chase camera's own occlusion system raycasts against the ordinary (tunnel-unaware) `terrainHeight`, so underground it always decided the classic surface overhead was blocking the view and pulled the camera back up to the surface - meaning no camera position under the old chase-cam could ever show the interior. Interior shots now use `renderer.editorCam` (the map editor's existing free-cam seam) positioned a few yards behind the player at head height, looking at the player's own head - the character is the visible subject in every interior shot.
- **"Zone1 east side to zone2 east side, not top-left"**: the tunnel moved from `x=-25` (west-of-center) to `x=110` (a real ridge on the east side of both zones, clear of every camp/POI/road/lake).
- **"Single entity, not separate"**: see the rendering explanation above - it's now one mesh, not a tunnel layer plus an untouched terrain layer underneath.
- **20+ screenshots, old ones removed**: 23 shots (10 exterior, 13 interior) in this round, the previous 16 removed from the PR-assets branch.

## What changed in the latest round (round 3)

- **"Camera must never show below/outside the tunnel, not just in screenshots"**: found and fixed the actual bug this could happen from. `tunnel_overlay.ts` wraps every tunnel section in a fully enclosing backstop (six opaque planes: floor, ceiling, four walls) as defense-in-depth against any gap the voxel mesher could leave. Two of those planes (south/north) had their `rotateY` values swapped, giving them inverted, outward-facing normals - invisible from inside the tunnel due to normal backface culling, letting the camera see straight through to the raw scene background at certain angles. Fixed the rotation and additionally set the shared material to `THREE.DoubleSide` so a future orientation mistake can't reproduce this. This is a render-level fix in `tunnel_overlay.ts`, not a screenshot-camera parameter, so it holds for the ordinary in-game chase camera regardless of how the player turns it, not just this PR's screenshot script.
- **"Increase the diameter further, double the depth, decorate more"**: every waypoint radius bumped up again; the vertical drop from each mouth to the crest doubled again (the crest is now noticeably deeper under the ridge); added torches (emissive flame mesh + point light) and boulders (icosahedron rock) at alternating waypoints along both tunnel legs.
- **"Cave entrance more obvious, above ground more, but still realistic"**: both mouth waypoints raised further above the local `terrainHeight` (now 1-6yd of visible rise, was under 2yd). Each mouth's (x, z) sits exactly where the ridge is already climbing toward the crest, so the single combined terrain+cave mesh naturally slopes the grass up to meet the raised opening - a real hillside cave mouth, not a floating arch with no slope leading to it.
- Lighting fix alongside the backstop change: the backstop's opaque planes had been accidentally acting as the tunnel's only real light-blocker; once its orientation was fixed the interior went nearly black (the ad-hoc torch `PointLight`s aren't registered in `renderer.ts`'s `budgetFireLights` system, so their effective brightness underground was unreliable). Added `AmbientLight`/`HemisphereLight` (which bypass that budget system) as the tunnel's reliable base light, with torches as a decorative accent on top.

## What changed in this round (round 4)

- **"In smaller enclosed areas, reduce the camera's max zoom to better manage the camera"**: new `src/render/camera_zoom_limit.ts`, a small pure module (`clampCamDistForEnclosedSpace`, unit-tested in `tests/camera_zoom_limit.test.ts`) that checks whether the player currently sits inside a tunnel's carved interior via the existing `tunnelSpanAt` (`src/sim/tunnel_traversal.ts`) and, if so, caps the chase camera's requested `camDist` down to 8yd (from the ordinary 22yd max) before `renderer.ts`'s `updateCamera` uses it to place the camera. This is on top of, not instead of, the existing occlusion pull-in from round 3: that system still reactively pulls the camera back in if a requested position collides with anything; this proactively keeps the chase camera from ever asking for a position that far out in the first place while in a tunnel, so it no longer reads as the camera fighting the player's zoom input in a cramped space. All 20 screenshots retaken against this change.

## What changed in this round (round 5)

- **"Cave was too circular - top half arched, bottom half flatter"**: the cross-section was a uniform sphere at every waypoint, subtracted from the terrain, which reads as a circular pipe bore rather than a real cave. `TunnelWaypoint` (`src/sim/content/tunnels.ts`) now takes two optional shape fields, `archScale`/`floorScale` (both default to `1`, a plain sphere - `vale_kobold_warren` is untouched), and `segmentCarve` (`src/sim/voxel.ts`) uses an anisotropic ellipsoid metric: the vertical half-extent above a waypoint's own y is `radius * archScale`, below it is `radius * floorScale`. Every `vale_marsh_ridge_tunnel` waypoint domes taller overhead than it is wide and flattens underfoot, everywhere along the route, not just at the mouths.
- **"Diameter slightly larger, tunnel way deeper into the ground"**: every waypoint radius bumped up another +15%, and the crest deepened from ~38yd below mouth level to ~54yd.
- **"Entrance should be a vertical upright opening, not a hole flat on a ground"**: both mouth waypoints pushed `archScale`/`floorScale` further (2.2/1.0), plus each mouth's first ~8yd re-profiled to stay close to level before the passage curved into its real descent (the previous steep drop right at the threshold was reading from outside as looking straight down a well/pit shaft).
- **"Reduce the max camera distance in enclosed areas even shorter"**: `ENCLOSED_MAX_CAM_DIST` reduced from 8yd to 5yd.

## What changed in this round (round 6)

- **"Entrance must be vertical like a door opening, upright, not a manhole on the ground"**: round 5's fix (a tall archScale/floorScale ellipse) turned out not to be enough by itself - a camera standing right at the threshold and looking straight down the corridor still read as looking into a shaft, because that close-up framing is dominated by the tube's own cross-section, not the surrounding landform. The actual fix needed is below (the mound). The screenshot camera script was also tuned so the mouth close-ups (`04`/`06`) frame the mound and doorway together rather than standing on the exact threshold looking straight in.
- **"The entrances must be protruding mounds"**: this is the real fix for the "manhole" read. Added a `mound` field to `TunnelWaypoint` (`src/sim/content/tunnels.ts`) plus `moundSolidAmount` in `src/sim/voxel.ts`: an additive solid dome (a spherical falloff, same math family as the tunnel carve but inverted and additive rather than subtractive) anchored to the ambient ground level at a flagged waypoint's own `(x, z)`, 14yd radius and 10yd tall for both mouths. `voxelDensity` now applies every mound (`Math.min`, more solid wins) before applying the tunnel carves (`Math.max`, carve always wins), so the doorway is cut through a real protruding knoll of rock/earth rather than just a dip in the existing ridge slope. `tunnelBounds` widened to cover each mound's lateral and vertical extent so chunk exclusion and the mesher's height band still cover it correctly.
- **"Cave steepness max 30 degrees only"**: the previous round's crest depth (~54yd) combined with the level mouth entries pushed some mid-route segments well past a walkable grade. Every segment of `vale_marsh_ridge_tunnel`'s centerline now holds to `|dy/dz| <= tan(30deg)`, enforced by a new regression test. To fit the (now shallower, ~25yd) crest depth within a 30 degree grade given the available run, both mouths moved out from `z=146`/`214` to `z=128`/`232` - still on the same ridge, still clear of every camp/POI/road/lake, just a longer straight approach on each side.
- All 20 screenshots retaken against this geometry; the interior shot camera offsets (`scripts/tunnel_screenshot.mjs`) are recomputed per shot from the tunnel's own live floor/ceiling at the camera's/target's actual position, each verified open with `isSolidVoxel`. The previous round's screenshots were removed from the PR-assets branch.

## What changed in this round (round 7)

- **"Both mouths should look organic (a real cave-entrance photo reference), not geometric/symmetric"**: the entrance mound's own silhouette used to be a perfect sphere-derived dome (uniform `moundRadius`/`moundHeight` in every direction). `moundSolidAmount` (`src/sim/voxel.ts`) now applies a small, fixed, deterministic angular "irregularity" (a sum of a few sine lobes around the mound's own center, hand-authored coefficients, **not** randomness) to the mound's lateral radius and height, so the knoll pokes out further on some sides than others and reads as a real, lumpy hillside rather than a CAD dome. The tunnel's own carved doorway (`segmentCarve`) is untouched by this, so the opening itself always stays a clean, walkable passage; only the surrounding rock/grass knoll gets the irregular silhouette. New unit test in `tests/voxel.test.ts` confirms the mound is no longer perfectly circular (density sampled at the same radius but different angles now differs).
- **"Foliage overhanging the opening, a path leading to the threshold"**: new `addMouthFoliage` in `src/render/tunnel_overlay.ts` (same procedural-geometry style as the existing `addTorch`/`addBoulder` helpers, no new binary assets): a handful of hanging branch-and-leaf-clump meshes ringing the top of each archway at deterministic angles/heights (an irregular canopy, not a symmetric fringe), plus a short dirt/rock path patch just outside each threshold leading away from the doorway toward open ground. Called once per mouth waypoint from `addDecorations`.
- **"No grass anywhere inside the tunnel interior"**: the shader's grass/rock blend in `tunnel_overlay.ts`'s `onBeforeCompile` keyed purely on surface normal slope, which is wrong for the tunnel's own flattened floor (`floorScale` makes it face upward like ordinary ground even though it's buried underground) - it was texturing as grass underground. Fixed with a new pure helper, `tunnelInteriorFactor(x, y, z, seed)` in `src/sim/voxel.ts`: returns 0..1, ramping in only when a point is both genuinely below the ambient outdoor `terrainHeight` (so a mound's own grassy exterior, which sits at/above ambient ground level, is never affected) and close to some tunnel segment's own carved capsule. `tunnel_overlay.ts` computes this per-vertex (a new `aTunnelInterior` custom `BufferAttribute`, applied to both the meshed patch and every backstop plane) and feeds it into the shader as a varying, taking `max(slopeT, vTunnelInterior)` so the interior always forces full rock while the mound's exterior keeps the original slope-based blend. New unit tests in `tests/voxel.test.ts` cover `tunnelInteriorFactor` directly (pure function, 1 deep inside a passage, 0 on the open surface, 0 on a mound's grassy exterior even close to a mouth, 0 underground far from any tunnel).
- **"Push the tunnel deeper"**: the crest deepened from `y=-22` to `y=-37`, well under the ridge (the ridge surface itself sits at `y=+35.8` at the crest, so the tunnel is now ~73yd of solid rock below it). To keep every segment within the existing 30 degree max-grade regression test given the extra depth, both mouths moved further out from the `z=180` zone boundary, from `z=128`/`z=232` to `z=96`/`z=264`.
- **"Add S-curves to the path"**: the interior waypoints now vary laterally instead of holding a constant `x=110`: `110 -> 116 -> 112 -> 104 -> 110` (crest) `-> 116 -> 112 -> 104 -> 110`. Both mouths stay at `x=110` exactly, so mouth-position expectations elsewhere didn't need to change. The widest excursion (`x=104`/`x=116`) is still well inside the same east-side corridor, clear of the causeway (`x=0`), both zone hubs, every lake, every POI (Boar Meadow/Bandit Camp/Fallen Chapel/Brightwood Glade/Widow Thicket), and the unrelated `vale_kobold_warren` tunnel to the west.
- Every test referencing this tunnel's literal coordinates updated for the new mouth z (`tests/tunnel_traversal.test.ts`, `tests/player_motion.test.ts`, `tests/pathfind.test.ts`) and the new crest y (`tests/camera_zoom_limit.test.ts`; the crest itself is still exactly `x=110, z=180`, so those two coordinates were left as-is).
- All 24 screenshots retaken against this geometry (2 new wide overhead shots showing the S-curve from above, 2 new interior shots at the widened crest span, all mouth/interior camera offsets recomputed from the tunnel's own live floor at each shot's actual position); the previous round's screenshots were removed from the PR-assets branch.

## How it works

```mermaid
flowchart LR
    subgraph Zone1["Eastbrook Vale (zone1, z less than 180), east side"]
        MouthA["Mouth A, organic mound\n(128, 66)"]
    end
    subgraph Ridge["Ridge crest, z = 180 (zone boundary)"]
        Crest["Deepest point\n(128, -52.5y, 180)\nwell under the ridge"]
    end
    subgraph Zone2["Mirefen Marsh (zone2, z greater than 180), east side"]
        MouthB["Mouth B, organic mound\n(128, 294)"]
    end
    MouthA -- "level, then S-curving max-30deg descent" --> Crest -- "max-30deg ascent, S-curving, then level" --> MouthB
```

```mermaid
flowchart TD
    A["renderer.ts builds terrain.ts,\nthen tunnel_overlay.ts"] --> B["terrain.ts: chunkNearAnyTunnel(x0,z0)?"]
    B -- "no" --> C["ordinary heightfield chunk\n(unchanged everywhere else)"]
    B -- "yes" --> D["skip: leave a real hole"]
    D --> E["tunnel_overlay.ts: mesh voxelDensity\n(terrain, mounds AND tunnel combined)\nover the SAME excluded chunk grid"]
    E --> F["one continuous isosurface:\nground domes up into a mound,\nthen a doorway is cut through it,\ntextured to match, never z-fights"]
```

```mermaid
flowchart TD
    A["player_motion.ts / pathfind.ts\nneed a ground height at (x, y, z)"] --> B{"tunnelSpanAt / tunnelColumnAt\n(x, z) inside a TunnelVolume's footprint?"}
    B -- "no" --> C["ordinary groundHeight(x, z)"]
    B -- "yes" --> D["scan voxelDensity(x, y, z)\nfrom below the tunnel's AABB upward"]
    D --> E["first solid to air crossing = floor\nnext air to solid crossing = ceiling"]
    E --> F["player_motion.ts rides the tunnel floor"]
    E --> G["pathfind.ts treats it as a real lower route"]
```

```mermaid
flowchart TD
    A["groundHeightWithMounds(x,z,seed)\n(round 10)"] --> B["h = terrainHeight(x,z,seed)"]
    B --> C{"any waypoint w flagged mound: true?"}
    C -- "yes" --> D["moundHeightBump(x,z,w):\nADDITIVE Gaussian bump, same technique\nas world.ts's own ridge walls, angular\nirregularity applied (round 7)"]
    D --> E["h += bump\n(no edge: continuous with the\nsurrounding terrain everywhere)"]
    C -- "no" --> F["h unchanged"]
    E --> G["voxelDensity: density = y - h"]
    F --> G
    G --> H["for every tunnel segment"]
    H --> I["carve = segmentCarve(...)\n(ellipsoid, archScale/floorScale)"]
    I --> J["density = max(density, carve)\n(carve always wins: the doorway\nis cut through the mound)"]
```

```mermaid
flowchart TD
    A["tunnel_overlay.ts: buildTunnelOverlay"] --> B["for each tunnel in TUNNELS"]
    B --> C["b = tunnelBounds(tunnel)\n(that ONE tunnel's own carve+mound extent)"]
    C --> D["ONE deep floor cap, sized tight to b,\nsunk well below b.minY - no walls, no ceiling"]
    D --> E["never surfaces above ground (round 9):\nthe watertight terrain patch already\nblends continuously into ordinary terrain.ts"]
```

```mermaid
flowchart TD
    A["tunnel_overlay.ts shader, map_fragment"] --> B["slopeT = f(surface normal)\n(the ORIGINAL grass/rock blend)"]
    A --> C["vTunnelInterior = tunnelInteriorFactor(x,y,z,seed)\n(per-vertex, src/sim/voxel.ts, round 7)"]
    C --> D{"genuinely below ambient terrainHeight\nAND near a carved tunnel segment?"}
    D -- "yes (tunnel floor/wall/ceiling)" --> E["rockT = max(slopeT, 1)\nalways full rock, ignores slope"]
    D -- "no (mound's own grassy exterior,\nor open air far from any tunnel)" --> F["rockT = slopeT\n(unchanged, ordinary blend)"]
    E --> G["diffuseColor *= mix(grassAlb, rockAlb, rockT)"]
    F --> G
```

## Screenshots (53, real production client, round 11)

`npm run dev`, the actual live renderer, the tunnel as one deformed terrain+cave surface, retaken after the round-11 mound-roundness and camera-clipping/lock-up fixes above (same shot list as round 10, all 53 re-rendered; two of the closeup shots nudged a few yards further from their mouth). All previous rounds' screenshots removed from the PR-assets branch; this is the current set.

**Fixes verified: the mound reads as one continuous grassy hill (no hard edge), no floating grass/foliage at either entrance, open sky at multiple points, generous mob-camp clearance**
Sky check, zone1 near mouth A: looking straight up - blue sky, not rock
Sky check, zone1 mid-route: same check further along the descent
Sky check, zone2 near mouth B: same check on the Mirefen Marsh side
Sky check, zone2 mid-route: same check further along the ascent
Open field east of the route, wide: no stray rock walls or floating geometry
Open field west of the route, wide: same check from the other side
Wide overhead at the crest, emphasizing the round-9 extra depth

**Both mounds: an additive terrain bump (same technique as the world's real border mountains), no floating cube, no floating grass/foliage, no clutter**
Mouth A, overhead: a real grassy hill, no floating rock slab, no tree/rock decoration clashing with it
Mouth B, overhead: same check on the second mound
Mouth A, ground level: the worn path and doorway with a fully clear approach
Mouth B, ground level: same check at the second mouth

**World map: the "Cave" marker at each mouth**
Eastbrook Vale world map, showing the Cave marker near the mound A location
Mirefen Marsh world map, showing the Cave marker near the mound B location

**Approaching from Eastbrook Vale (east side) - a natural hill blending into the ridge, a flush upright doorway at ground level, no crater**
Wide overhead, showing the S-curve carved under the ridge
Wide overhead, full route from the crest
Wide establishing, zone1
Approaching the ridge
Mouth A, wide: a natural rounded hill rising out of the grassy hillside, no separate floating structure
Mouth A, close up: a flat, flush threshold at ground level, not a pit
Mouth A, frontal: the tall, upright doorway shape (archScale 3.0 / floorScale 0.45)
Mouth A, side angle: the level dirt threshold, no bowl-shaped dip

**Inside the tunnel - bare rock throughout (no grass), arched ceiling, flatter floor, the deeper S-curve, character always visible**
Entering the S-curve, all rock underfoot
Descending 0
Descending 1
Descending 1b
Descending 2
Descending 2b
Descending 3
Descending 3b
Approaching the crest 1
Approaching the crest 1b
Approaching the crest 2
Approaching the crest 2b
Near the crest
Near the crest, deeper
The deepest point, exactly at the zone1/zone2 boundary (z=180, now y=-52.5): entirely rock, zone transition text visible
The boundary crest, facing back the way the player came
Just past the crest
Just past the crest, further
Ascending 1
Ascending 1b
Ascending 2
Ascending 2b
Ascending 3
Ascending 3b
Leaving the S-curve
Leaving the S-curve, further
Nearing mouth B
Almost at mouth B

**Emerging into Mirefen Marsh (east side) - the second mound and doorway, same treatment**
Mouth B, close up: character emerging from the mound's doorway, the real "Entering Mirefen Marsh" zone transition firing
Mouth B, frontal: the upright doorway shape, matching mouth A
Mouth B, wide
Wide establishing, zone2

## Test plan

- [x] `tests/voxel.test.ts`: tunnel carves open air at every waypoint, stays solid outside each buried/near-mouth waypoint's radius, waypoints straddle `z=180`, deepest point sits well under the ridge crest, both mouths sit above the local surface, the reported bounding box accounts for the archScale/floorScale/mound-scaled Y and lateral extent, every segment holds to a walkable 30 degree max grade, both mouths are cut into an irregular, non-circular, real protruding entrance mound.
- [x] `tests/voxel.test.ts`: `tunnelInteriorFactor` describe block - pure function, 1 deep inside the tunnel passage, 0 on the open world surface, 0 on a mound's grassy exterior even close to a mouth, 0 underground far from any tunnel.
- [x] **Updated (round 10)** `tests/voxel.test.ts`'s mound-solidity probes: retuned for the new additive Gaussian falloff (a gradual decay curve instead of a hard-edged sphere/ellipsoid) - same pass/fail intent, probe positions chosen with enough margin to hold regardless of local base-terrain noise.
- [x] `tests/tunnel_traversal.test.ts` / `tests/player_motion.test.ts` / `tests/pathfind.test.ts`: unchanged, still green against the current geometry.
- [x] `tests/fixes.test.ts`: no tree/rock decoration (`generateDecorations`) ever spawns inside a tunnel's mound clearance radius, for every mound in `TUNNELS` - re-verified against the widened `moundRadius * 3.5` clearance.
- [x] `tests/map_tunnel_portals.test.ts`: `overworldTunnelMouths` returns only mound waypoints (never a buried interior one), filters correctly to a `[zMin, zMax)` zone band, and returns nothing for a tunnel with no mound.
- [x] **New/updated (round 11)** `tests/camera_zoom_limit.test.ts`: `cameraFloorAt` uses the tunnel's own local floor when the camera itself sits within the tunnel's floor/ceiling span deep under the marsh ridge crest, uses the ordinary outdoor surface for a camera up in open air over that same `(x, z)` column, and matches plain `groundHeight + 0.6` outside any tunnel.
- [x] **Re-verified (round 11)** `tests/voxel.test.ts`'s mound-solidity/`tunnelInteriorFactor` probes: still green against the softened irregularity coefficients and the widened S-curve waypoints.
- [x] `npx vitest run` green across `tests/voxel.test.ts`, `tests/voxel_mesh.test.ts`, `tests/tunnel_traversal.test.ts`, `tests/player_motion.test.ts`, `tests/pathfind.test.ts`, `tests/camera_zoom_limit.test.ts`, `tests/fixes.test.ts`, `tests/map_tunnel_portals.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/sim.test.ts`, `tests/localization_fixes.test.ts`.
- [x] `npm run ci:changed` (biome, changed files only): 0 errors, 0 format diffs.
- [x] `biome check` on every changed file: 0 errors, 0 format diffs (warnings only, pre-existing style already present in these files). One unrelated pre-existing format issue on an upstream-merged file (`scripts/fct_lowtier_shot.mjs`) fixed as a scoped single-file formatting commit so the pre-push `biome ci --changed` floor stayed green.
- [x] Merged latest `release/v0.24.0` into the branch (one conflict both times, in the generated `src/ui/i18n.resolved.generated/pending.ts`, resolved by regenerating via `npm run i18n:build`); pushed to `feature/vale-fenbridge-tunnel`; pre-push QA floor (tsc, guard tests, biome, copy scan) green.
- [x] Manually verified in the real game client via the 53 retaken screenshots above: the entrance mound now reads as one continuous grassy hill rising and falling smoothly out of the surrounding terrain (matching the world's real border mountains, no hard edge anywhere), no floating grass or foliage anywhere near either entrance, the sky reads as open sky well clear of either mound, and the world map still marks each mouth with a "Cave" label.


jgyy:
too difficult to implement, i already provided voxel baseline so tunnel future work

PR #1687, opened 2026-07-09
fix(sim): resummoning a same-alive warlock demon now gives a fresh pet

### LINKED ISSUE #1355: Re-summoning a warlock pet you already have dismisses it instead of resummoning fresh
## Problem

Re-casting a warlock demon summon while you already have that demon (alive) dismisses the pet instead of giving you a fresh one. `summonPet` despawns the existing demon, then, if it is the same template and still alive, returns early ("fades back into the void"), leaving you with no pet. Swapping to a different demon, or resummoning a dead one, already produces a fresh full-health pet; only the same-demon-alive case toggles off.

## Steps to reproduce

1. As a warlock, summon a Voidwalker (it is alive).
2. Cast Summon Voidwalker again.
3. Observe the pet vanish with no new pet in its place.

## Actual behavior

- The same-demon-alive branch despawns the pet and returns, so you are left with no pet.

## Expected behavior

- Re-summoning the same living demon replaces it with a fresh, full-health pet (as the swap and dead-pet paths already do), rather than dismissing it.

## Scope

- `src/sim/pet/pet_commands.ts` `summonPet`: the `existing.templateId === templateId && !existing.dead` early-return branch.
- `tests/pet_commands_module.test.ts` has a paired assertion (the "fades into the void, leaving NO pet" case) that would flip.

## Acceptance criteria

- Casting a demon's summon while that same demon is alive replaces it with a fresh full-health pet.
- Swapping demons and resummoning a dead demon are unchanged.

## Notes

The current dismiss-on-resummon is deliberate and tested; this is a decision to change it to a refresh, which matches player expectation.

jgyy:
## Problem

Fixes #1355.

Re-casting a warlock demon summon while you already have that demon (alive) dismisses the pet instead of giving you a fresh one. `summonPet` despawns the existing demon, then, if it is the same template and still alive, returned early ("fades back into the void"), leaving you with no pet. Swapping to a different demon, or resummoning a dead one, already produced a fresh full-health pet; only the same-demon-alive case toggled off.

## Steps to reproduce (before fix)

1. As a warlock, summon a Voidwalker (it is alive).
2. Cast Summon Voidwalker again.
3. Observe the pet vanish with no new pet in its place.

## Fix

`summonPet` (`src/sim/pet/pet_commands.ts`) had a same-template-alive branch that despawned the existing pet and then `return`ed before reaching `createDemonPet`. Every other branch (different template, or the same template but dead) already fell through to `createDemonPet` and got a fresh pet. Removing the early-return branch makes the same-demon-alive case behave the same as every other resummon: dismiss the old one, summon a fresh full-health one.

```mermaid
flowchart TD
    A["Warlock casts Summon Voidwalker\n(already has an alive voidwalker)"] --> B{"existing pet?"}
    B -- yes --> C["despawnPersistentPet(existing)"]
    B -- no --> E["createDemonPet(templateId)"]
    C --> D{"same template AND alive?"}
    D -- "before fix: yes" --> F["return early\n'fades back into the void'\n(NO PET LEFT)"]
    D -- "after fix: removed" --> E
    E --> G["fresh full-health demon answers\n'answers your summons'"]

    style F fill:#4a1f1f,stroke:#c0392b,color:#fff
    style G fill:#1f4a2b,stroke:#2ecc71,color:#fff
```

flow diagram

## Verification

Live in the offline client (warlock, level 20): summon Voidwalker, then re-cast the same summon while it's alive.

```
post-resummon pet state: { hasPet: true, hp: 258, maxHp: 258 }
```

`hasPet: true` and `hp === maxHp` confirm a fresh, full-health demon answered instead of the old dismiss-with-nothing-left behavior.

warlock resummon smoke test

## Tests

- `tests/pet_commands_module.test.ts`: same-demon-alive resummon now asserts a fresh pet (different id, full health) instead of asserting no pet.
- `tests/threat.test.ts`: renamed/updated "recasting the same demon unsummons it" -> "recasting the same demon dismisses it and summons a fresh one", asserting the same.
- `tests/parity/coverage.test.ts` + `tests/parity/scenarios.ts` + regenerated `tests/parity/golden/pet_commands.json`: the parity scenario's same-demon-alive re-summon step now tracks and asserts the replacement pet instead of an emptied pet slot.

All of `tests/pet_commands_module.test.ts`, `tests/threat.test.ts`, `tests/parity/parity.test.ts`, `tests/parity/coverage.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` pass; `npx tsc --noEmit` is clean; `npm run build` succeeds; changed-file Biome is clean (only pre-existing whole-repo warnings, none introduced by this diff).

No merge conflicts against `release/v0.23.0` (branched from its current tip).

### DIFF
--- a/src/sim/pet/pet_commands.ts
+++ b/src/sim/pet/pet_commands.ts
@@ -313,15 +313,6 @@ export function summonPet(ctx: SimContext, owner: Entity, templateId: string): v
   const existing = petOf(ctx, owner.id, true);
   if (existing) {
     despawnPersistentPet(ctx, existing);
-    if (existing.templateId === templateId && !existing.dead) {
-      ctx.emit({
-        type: 'log',
-        text: `${existing.name} fades back into the void.`,
-        color: '#b894ff',
-        pid: owner.id,
-      });
-      return;
-    }
   }
 
   const pet = createDemonPet(ctx, owner, templateId);
--- a/tests/parity/coverage.test.ts
+++ b/tests/parity/coverage.test.ts
@@ -155,7 +155,7 @@ describe('coverage: each scenario fires its subsystem', () => {
     // Demon swap AND same-demon re-summon both produce a fresh demon answering the call
     // (re-summoning while the current demon is alive dismisses it and summons anew, it
     // never toggles off into no pet).
-    expect(logs.filter((t) => t.includes('answers your summons')).length).toBeGreaterThanOrEqual(3);
+    expect(logs.filter((t) => t.includes('answers your summons')).length).toBeGreaterThanOrEqual(4);
     // despawnPet scrubbed the hunter's targetId (set to the demon, nulled on its hard despawn).
     expect((rec.sim as any).player.targetId).toBeNull();
     // abandon's despawnPersistentPet scrub pulled the biter off the (now-gone) pet.
--- a/tests/parity/coverage.test.ts
+++ b/tests/parity/coverage.test.ts
@@ -152,9 +152,10 @@ describe('coverage: each scenario fires its subsystem', () => {
     expect(logs.some((t) => t.startsWith('You abandon'))).toBe(true);
     // Demon Heal channel ticked: applyDemonHealTick emits a heal2 with ability 'Demon Heal'.
     expect(ev.some((e) => e.type === 'heal2' && e.ability === 'Demon Heal')).toBe(true);
-    // Demon swap exercised BOTH branches: a new demon answered, then the same demon faded.
-    expect(logs.some((t) => t.includes('answers your summons'))).toBe(true);
-    expect(logs.some((t) => t.includes('fades back into the void'))).toBe(true);
+    // Demon swap AND same-demon re-summon both produce a fresh demon answering the call
+    // (re-summoning while the current demon is alive dismisses it and summons anew, it
+    // never toggles off into no pet).
+    expect(logs.filter((t) => t.includes('answers your summons')).length).toBeGreaterThanOrEqual(3);
     // despawnPet scrubbed the hunter's targetId (set to the demon, nulled on its hard despawn).
     expect((rec.sim as any).player.targetId).toBeNull();
     // abandon's despawnPersistentPet scrub pulled the biter off the (now-gone) pet.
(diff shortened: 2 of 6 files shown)


FernandoX7 (commented):
Clean, well-scoped fix. Dropping the same-template-alive early return makes the resummon fall through to `createDemonPet` like every other branch (different demon, or a dead one), so re-casting a living demon now replaces it with a fresh full-health pet. That matches #1355 and the deliberate behavior change it asks for. Determinism is safe: the diff only removes code, and the regenerated golden is a faithful replay (the `demon-resummoned` snapshot pins the new full-health entity while the rng draw count holds at 158, confirming `createDemonPet` consumes an id but no rng). Nice that the fix lands with three decisive guards: the unit assertions in `pet_commands_module.test.ts` and `threat.test.ts` (new id + `hp === maxHp` + old entity gone) plus the golden. Merges clean against `release/v0.24.0`, and the touched test files pass locally.

Two nits:

- nit: `tests/pet_commands_module.test.ts:156` still titles the test `'warlock demon swap: answers vs fades-into-the-void ...'`, but the body no longer exercises a fades outcome. Worth retitling so the name matches what it now asserts.
- nit: `tests/parity/coverage.test.ts:158` asserts `>= 3` "answers your summons", but the scenario fires four summons and the old toggle-off behavior produced exactly three (three answers + one fade). So this bound alone would still pass against the bug. `>= 4` makes it decisive. Not load-bearing here since the golden and the two unit tests already pin the fix, but it is the branch-coverage line that used to check both outcomes.

One maintainer-side note, no action needed from you: with the emitter gone, the `log.petFadesVoid` key and its matcher (`src/ui/sim_i18n.ts:5381`) are now orphaned. That is release-time locale cleanup on my end, not yours.

jgyy:
Thanks for the thorough review! Both nits addressed in 62848c8:

- Retitled the `pet_commands_module.test.ts` test to drop the "fades-into-the-void" reference since that outcome no longer exists.
- Tightened `coverage.test.ts` to `>= 4` for "answers your summons" so it's decisive against the old toggle-off bug (which would still satisfy `>= 3`).

Left the `log.petFadesVoid` locale cleanup for you as noted.

FernandoX7 (approved):

PR #1695, opened 2026-07-09
feat(mailbox): select a quantity when attaching a parcel

### LINKED ISSUE #1444: [Mailbox] Selecting an item quantity to send in an email
### Before you start

If I own multiple copies of an item, I can't send just a single one by mail to another character.

### Proposed solution

Add [+] and [-] buttons to select the quantity.

Alternatively, 1 click = 1 item added to the email.

### Alternatives considered

_No response_

### Area

None

jgyy:
## Summary

Closes #1444. Staging a bag stack in the Send tab of the Ravenpost mailbox
always attached the *whole* stack: if you owned 8 of an item there was no
way to send just 1 to another character short of splitting the stack in
your bags first (not currently supported either).

Adds a `-`/`+` stepper to the staged parcel chip so the quantity can be
adjusted down to as few as 1 before sending, clamped to what the bag
actually holds. Remove (`x`) still drops the parcel entirely.

The sim's mail send path (`src/sim/mail/post_office.ts`) already validates
and deducts a partial `count` per item (it errors with `notEnoughItems` if
you try to send more than you own), so this is a client-only UI change; no
wire protocol or sim changes were needed.

```mermaid
sequenceDiagram
    participant Player
    participant MailboxWindow as MailboxWindow (client)
    participant Sim as Sim / post_office.ts (authoritative)

    Player->>MailboxWindow: click bag item (Spitted Boar Haunch x8)
    MailboxWindow->>MailboxWindow: stageParcel() -> attachments = [{itemId, count: 8}]
    Player->>MailboxWindow: click "-" on the parcel chip
    MailboxWindow->>MailboxWindow: adjustParcelQty(-1) -> clampParcelQty(8, -1, owned=8) = 7
    Player->>MailboxWindow: click "Send letter"
    MailboxWindow->>Sim: mailSend(to, subject, body, copper, [{itemId, count: 7}])
    Sim->>Sim: countFungibleItem(itemId) >= 7 ? deduct 7 : notEnoughItems
    Sim-->>Player: letter sent (7 in escrow, 1 stays in bags)
```

## Related issues

Closes #1444

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/mailbox_view.test.ts tests/mailbox_window.test.ts` (added `clampParcelQty` unit tests covering the increment/decrement/clamp/remove-instead-of-zero cases)
  - `npx tsc --noEmit`
  - `npm run i18n:gen` (new aria-label keys; added the 5 non-Latin fills for the M16 wordy-string rule) + `npx vitest run tests/i18n_completeness.test.ts tests/i18n_resolved_equivalence.test.ts tests/i18n_status_registry.test.ts`
  - `npm run build`
  - `npx @biomejs/biome check --write` on the changed files
- Manual steps: played an offline session, gave myself a stack of 8 Spitted Boar Haunches (`/dev give`), opened the mailbox Send tab, attached the stack, and used the +/- stepper to bring it down to a partial quantity before sending. Confirmed the "+" button disables at the owned ceiling and "-" never goes below 1.

## Screenshots / recordings

Before (whole-stack-only, no way to reduce the quantity):

before

After (the new +/- stepper, staged at a partial quantity of 6):

after

---

### Quality

- [x] **Tests pass.** `npm test` is green (ran the mailbox + i18n suites directly above; the full suite was also run and any other failures were confirmed to be pre-existing flakiness under core contention, not from this change).

### Cross-platform

- [x] **Tested on desktop and mobile.** No new touch targets beyond the existing parcel-chip pattern; the stepper buttons follow the same min-height/hit-area conventions as the existing remove button.
- [x] **Accessible.** Keyboard-operable buttons with descriptive `aria-label`s ("Send one fewer/more {item}"); the "+"/"-" buttons disable (not just visually gray) at the ceiling/floor.

### Localization (i18n)

- [x] **New player-visible strings are translated into every supported locale.** Added to `en` (`hudChrome.mailbox.parcelQtyDecreaseAria` / `parcelQtyIncreaseAria`) plus the 5 non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) required for a new wordy English string; the rest are English-filled by the build per the contributor workflow.
- [x] N/A for the sim/server re-localization item; no new sim/server-emitted text.

### Hygiene

- [x] No secrets committed; `ALLOW_DEV_COMMANDS` was only used locally for manual testing, not in any shipped path.
- [x] No hand-edited generated files; `src/ui/i18n.resolved.generated/*` and `src/ui/i18n.resolved.sha256` were regenerated via `npm run i18n:gen` / `node scripts/i18n_resolved_hash.mjs --write`.


Rubsey (commented):
The core of this is right, and I verified the load-bearing claim rather than trusting it: the per-item count rides the wire verbatim (src/net/online.ts:2432, floored and re-checked at server/game.ts:3481) and the sim re-validates against fungible stock and deducts exactly the partial count (src/sim/mail/post_office.ts:313-345), with partial deduction already pinned by tests/mail.test.ts. Extracting the clamp into mailbox_view.ts as a pure tested helper is the house pattern, and repainting only #mail-parcels after a step is the correct choice (a full render() wipes the typed form). Real buttons, t() aria-labels, true disabled at both bounds: all good.

Findings:

- **should-fix**: the chip layout regressed, the item icon now stacks above the item name. itemIcon() emits img.item-icon, which is display: block (src/styles/components.css:1008), and it now sits inside a plain inline span (src/ui/mailbox_window.ts:615) instead of directly under the inline-flex chip, so the block img breaks the inline flow. Your own after screenshot shows it: the before chip is one line, the after chip is two, with the icon on its own row. One rule fixes it, e.g. make that name span inline-flex with align-items: center and the chip's 6px gap.

- **should-fix**: the mobile touch floor now targets a dead element. hud.mobile.css:3070 floors .mail-parcel-chip to 40px and its comment justifies that as "a real \<button\>"; the chip is now a non-interactive span while the actual controls (the 18x18 steppers and a padding: 0 remove button around a 10px icon) have no coarse-pointer floor anywhere, so on touch the remove target shrank from the whole 40px chip to a 10px icon. The in-repo pattern for exactly this control shape is .yh-toggle's pointer: coarse 40x40 (components.css:7213). The now-stale comment needs updating in the same pass.

- **should-fix**: each +/- activation destroys keyboard focus. adjustParcelQty repaints via renderParcels(), which clears #mail-parcels (innerHTML = '', mailbox_window.ts:601) and rebuilds, so the button just activated with Enter is gone and focus falls to body; stepping a stack of 8 down to 1 means tabbing back into the window seven times. The file already has the re-focus-after-rebuild convention (the tab strip does it at line 242). While in there: nothing announces the new count (the value span has no aria-live and the count is not in the aria-labels).

- **should-fix**: the staged item's tooltip is no longer keyboard-reachable. It moved from the focusable chip button to a non-focusable span (mailbox_window.ts:617), and attachTooltip's keyboard path is focusin (src/ui/hud.ts:3992), so Tab can no longer summon it.

Nits:

- mailbox_window.ts:142: ownedCountFor sums every slot including instanced copies, but the send validates and deducts fungible-only (countFungibleItem skips s.instance, src/sim/sim.ts:5053), so with an instanced copy in the bag the plus button can allow a count the sim then refuses with notEnoughItems. The basis is inherited from the old attach code, but since the stepper newly builds an interactive ceiling on it, filtering !s.instance here would make the max match what is actually sendable.
- src/ui/mailbox_view.ts:96: clampParcelQty(current, delta, 0) returns 1, against its own comment ("never above what the bag actually holds"). Reachable when the bag empties between repaint and click; harmless since the sim refuses, but worth a test pinning the intended result.
- The painter half (the owned > 1 render condition, the disabled states, the clampParcelQty wiring) has no coverage. tests/mailbox_window.test.ts's own pin pattern (expect(painter).toContain('recipientSuggestions(')) extends directly: a clampParcelQty( pin plus the two new aria keys would guard the wiring.
- mailbox_window.ts:635 hardcodes the 'x' count prefix; t('itemUi.bags.stackCount') already exists for exactly this (hud.ts:7483). The inbox chip in this file does the same, so it is local convention, but a new sink may as well use the key.
- .mail-parcel-step declares no font (siblings like .mail-action-btn set family and size), so the +/- glyphs render in the UA button font, and neither new button class carries the house :focus-visible ring (the UA default still shows, so cosmetic only).

On the conflicts: the only true conflicts against release/v0.24.0 are the generated i18n artifacts (pending.ts, i18n.resolved.sha256, i18n.status.json, i18n.status.summary.json), the standard stale-base regen churn, not a design problem. Merge the base and re-run the i18n build and hash write; no hand-editing needed.

PR #1697, opened 2026-07-09
docs(prd): heroic dungeon itemization and the endgame loot ladder

ryan-foo:
## What this is

Design doc only, no code. One PRD: `docs/prd/heroic-dungeon-itemization.md`, the follow-up the shipped heroic system (PR #1575) said it needed. It closes the itemization and balance items #1575 left open (its own test plan ends on an unchecked "Balance pass on the heroic tuning multipliers" box, and its body notes the vendor and loot table were not in that pass).

## The problem, verified on release/v0.24.0

I ran the real `itemLevel` / `itemScore` functions against the live content. Every heroic epic reads item level 31, so:

- Heroic 5-man epics (ilvl 31) outrank the normal raid (ilvl 29). A five-person dungeon out-gears the ten-person raid.
- Heroic 5-man and heroic raid are identical: the heroic 5-man chest (`morthens_cryptforged_hauberk`) and heroic raid chest (`soulforged_warplate`) are byte-for-byte the same power (22 stats, 335 armor, itemScore 49.9). The raid earns no premium in its own tier.

Cause: `HEROIC_LOOT_SOURCE_LEVEL` is 25 (heroic 5-mans already clear the raid) and the heroic loot loop hardcodes `raid: false`, so the raid bonus fires in the normal tier (26 to 29) but is switched off in the heroic tier. PR #1693 equalized the four dungeons' combat difficulty but did not touch loot, so equal fights now hand out an inverted reward ladder.

## What the PRD proposes

One coherent ladder from two existing levers (no new formula constants):

```
normal 5-man 26  <  heroic 5-man 28  <  normal raid 29  <  heroic raid 31
```

- `HEROIC_LOOT_SOURCE_LEVEL` 25 to 22 (heroic is +2 over normal).
- Pass `isRaidMob(bossId)` in the heroic loot loop so the raid bonus applies inside the heroic tier too (Nythraxis holds at 31; the four 5-mans fall to 28).
- Re-itemize the four 5-man heroic sets to the ilvl-28 budgets (exact per-slot table in the doc), plus the jewelry vendor band. The Nythraxis heroic set is unchanged.

It argues why heroic 5-mans should sit at the raid's shoulder rather than above it (one raid, so the top rung has to stay on top) and scopes the difficulty-multiplier balance pass out as a separate open item (validation, not invented numbers).

## Status

Draft on purpose: this carries the spec for discussion before anyone starts the implementation slices (S1 to S4 in the doc, each with a green-able vitest command). Docs-only, no CI-affecting code.

## Test plan

- [x] Docs-only change; no sim/server/UI code touched.
- [x] Premise verified on `release/v0.24.0` with the real `item_level` functions (throwaway eval, not committed).
- [x] No em/en dashes or emoji.

### DIFF
--- a/docs/prd/heroic-dungeon-itemization.md
+++ b/docs/prd/heroic-dungeon-itemization.md
@@ -1,0 +1,272 @@
+# PRD: Heroic Dungeon Itemization and the Endgame Loot Ladder
+
+| | |
+|---|---|
+| **Status** | Draft. Follow-up to the shipped heroic system (PR #1575), closing the itemization and balance items that PR left open |
+| **Owner** | design |
+| **Created** | 2026-07-09 |
+| **Scope** | Loot itemization only: the item levels and stat budgets of heroic dungeon rewards. Difficulty tuning is a secondary section (section 8). No new geometry, mobs, currencies, or vendors |
+| **Related systems** | Item level and stat budget (`src/sim/item_level.ts`: `itemLevel`, `expectedStatBudget`, `RAID_ILVL_BONUS`, `isRaidMob`), heroic drop table (`src/sim/content/heroic_loot.ts`: `HEROIC_LOOT_SOURCE_LEVEL`, `HEROIC_BOSS_LOOT`), heroic vendor stock (`src/sim/content/heroic_vendor.ts`: `HEROIC_VENDOR_SOURCE_LEVEL`), difficulty tuning (`src/sim/content/dungeon_difficulty.ts`: `HEROIC_DUNGEON_TUNING`), budget tests (`tests/item_level.test.ts`) |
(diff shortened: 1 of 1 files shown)


ryan-foo:
Superseded by #1705. That PR takes the approach you chose: keep the item-level-31 final-boss epics as-is and instead upgrade the base-table drops on heroic (epics to 28, rares to 25, 'Heroic X'), rather than re-laddering the whole tier.

One thing #1705 deliberately does NOT address, noted here so it is not lost: the heroic RAID (Nythraxis) drops read the same item level 31 as the heroic 5-man set, so the raid still earns no itemization premium over a 5-man within the heroic tier. If that is worth fixing, it is a small follow-up (give the heroic raid group the raid bonus so it reads 34, or a dedicated band); happy to open it separately.

PR #1702, opened 2026-07-09
feat(game): remember the camera zoom level across sessions (#1657)

### LINKED ISSUE #1657: Remember zoom level when exiting and entering the game
### Before you start

The zoom level resets when exiting and entering the game. For example, if I zoom in all the way on my character, then come back the next day, the zoom level is at default again. Like the UI scale and other settings, the zoom level should be remembered. Optionally: Add a "reset zoom level on login" option for users that prefer the current mode.

### Proposed solution

Mimic the logic that remembers all the other settings like UI scale to also save the zoom level that the player has configured.

### Alternatives considered

Leave it as it is.

### Area

UI / HUD

daxdax89:
## Problem

The camera zoom level (`Input.camDist`) resets to the default every time you exit and re-enter the game. Zoom all the way in on your character, come back the next day, and it is back at the default distance, unlike UI scale and the other settings which are remembered (issue #1657).

## Fix

Persist the zoom distance like every other setting, through the existing `settings.ts` store (`woc_settings`):

- Add a `cameraZoom` numeric setting. Its range mirrors `Input.zoomBy`'s `[3, 22]` clamp so a persisted value is always applicable; `def` 12 is the shipped starting distance.
- `Input` fires a new `onCameraDistChange` callback whenever the wheel or touch pinch changes the zoom. It does **not** fire on a direct `camDist` assignment, so restoring the saved value on boot never re-persists.
- `main.ts` persists it **debounced** (a touch pinch calls `zoomBy` per move frame, so this writes `localStorage` once the gesture settles, not on every delta).
- The saved value is applied back onto `Input` on boot by the startup apply-all loop, via a new `cameraZoom` case in `applySetting` (which also re-applies it on **Reset**).

Scoped to the required behavior (remember the zoom). The issue's optional "reset zoom on login" toggle is intentionally left out to keep this focused; it can be a follow-up.

## Tests

- `tests/settings.test.ts`: `cameraZoom` is in `SETTING_RANGES` at `{min:3,max:22,def:12}`, defaults to 12, clamps at both ends, and round-trips through `localStorage` (a fresh `Settings` reads the saved value).
- `tests/input.test.ts`: `zoomBy` clamps `camDist` to `[3,22]` and reports each real change to `onCameraDistChange`, and stays silent when the clamp leaves `camDist` unchanged (no spurious persist).

`tsc --noEmit` clean; `settings`/`input` suites green (84 passed); biome clean on all changed files.

## Scope

Client input/settings only, no sim/wire/server surface and no new player-visible strings. Each module keeps its own `localStorage` key (`cameraZoom` lives in the existing `woc_settings` store), per `src/game/CLAUDE.md`.

Closes #1657.

### DIFF
--- a/src/game/input.ts
+++ b/src/game/input.ts
@@ -115,6 +115,10 @@ export class Input {
   camYaw = Math.PI;
   camPitch = 0.32;
   camDist = 12;
+  // Fired whenever the player changes the zoom distance (wheel / pinch), so main.ts can
+  // persist it to settings (issue 1657). Not fired on a direct camDist assignment (the
+  // startup restore / Reset path sets the field itself), so restoring never re-persists.
+  onCameraDistChange?: (dist: number) => void;
   autorun = false;
   suspendMovement = false;
   // click-to-move (#95): a world destination the player clicked; the frame loop
@@ -279,7 +283,10 @@ export class Input {
 
   /** Move the camera in/out, clamped to the zoom limits. Shared by wheel + touch pinch. */
   zoomBy(delta: number): void {
-    this.camDist = Math.min(22, Math.max(3, this.camDist + delta));
+    const next = Math.min(22, Math.max(3, this.camDist + delta));
+    if (next === this.camDist) return;
+    this.camDist = next;
+    this.onCameraDistChange?.(next);
   }
 
   /** True while a mouse button is held for camera drag. */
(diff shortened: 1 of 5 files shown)


Rubsey:

Thanks for the contribution, @daxdax89.

PR #1705, opened 2026-07-09
feat(dungeons): heroic loot flair, soulbound marks, mark fan-out, and weapon-dps-by-ilvl

ryan-foo:
## What this is

Three heroic-dungeon reward changes, all sim-side and tested.

### 1. Heroic loot flair
When a mob dies in a HEROIC dungeon instance, its normal-table epic/rare drops are swapped for a "Heroic" upgraded variant: epics read item level 28, rares 25, stats rescaled to the matching budget. The variant keeps the base item's name (classic behavior: a heroic drop reads the same as its normal counterpart); the heroic distinction shows as a gold `[HEROIC]` tag on the tooltip's quality/kind line, e.g. "Barrowlord Warplate" / "Epic Armor [HEROIC]", never in the name. Green/uncommon drops are untouched, and the existing item-level-31 final-boss set is left as-is (the swap only fires when it is an upgrade, so raid epics at 29 are never downgraded).

Variants are generated from the base item + mob loot tables (`content/heroic_variants.ts`) and merged into `ITEMS`, so tooltip, equip, item score, and the server/client wire treat them like any other item with no special handling. The `[HEROIC]` tag is composed at the client from a single `hudChrome.itemHeroicTag` key, and the display name resolves to the base item's already-localized name, so no per-variant translation is needed and the entity manifest skips them. The pure budget math was extracted to a leaf `item_budget.ts` to keep the generator cycle-free.

### 2. Soulbound
New `ItemDef.soulbound` flag: a bound item cannot be traded, mailed, listed on the World Market, sold, or destroyed (right-click discard). `heroic_mark` is now soulbound, so marks can only be spent at the Heroic Quartermaster. The gates fold into the existing quest/`noMarketList`/`noVendorSell`/`noDiscard` checks, so no new player strings.

### 3. Mark loot fan-out
A heroic final boss now drops ONE shared-personal Heroic Mark slot for the whole party: any earner looting the corpse grants every earner their marks and consumes the slot, so no one has to reach the body and click their own copy.

### 4. Weapon dps tracks item level
Weapon damage had no item-level relationship: the item-level-31 heroic weapons ranged 11.2 to 17.8 dps, so the Gravewyrm Cleaver (13.5) sat below the item-level-26 Wyrmfang Greatblade (15.0), and Mistcaller's Fang (17.8) even out-dps'd the legendary (17.2). Adds `weaponDpsBudget(level)` (a curve fit to the authored ladder) and rescales the six heroic set weapons on-curve to 16.0 dps (above the 26 epics, below the legendaries); variants scale their weapon damage to their own item level.

## Screenshots

Heroic variant (item level 28, "Epic Armor [HEROIC]") next to the base drop (item level 26, "Epic Armor"), same name:

| Heroic variant | Base drop |
|---|---|
| heroic variant tooltip | base chest tooltip |

Heroic set weapon on-curve (16.0 dps) vs its base peer (15.0 dps), and the Soulbound mark:

| Heroic weapon | Base weapon | Soulbound mark |
|---|---|---|
| heroic weapon | base weapon | soulbound mark |

Regenerate: `node scripts/heroic_loot_shot.mjs` (needs `npm run dev`).

## Testing

- New: `tests/heroic_loot_flair.test.ts` (variant generation, budget-exactness, the shared-base-name behavior, the drop swap fires in heroic / not normal, raid epics not downgraded, weapon dps curve), `tests/heroic_soulbound.test.ts` (discard + trade blocked, control), plus the mark fan-out test and updated shape assertions in `tests/dungeons.test.ts`.
- Green locally: item_level, localization_fixes (S3), i18n_completeness, localization_coverage, parity (loot swap draws no rng, goldens byte-identical), architecture (new sim leaves are pure), snapshots, market, trade, mail_system, heroic_vendor, bags, combat, fixes.
- `npm run i18n:gen` is idempotent (freshness gate clean).

## Notes

This supersedes the loot-ladder approach in the `heroic-dungeon-itemization` PRD (#1697): the owner chose to keep the item-level-31 final-boss epics and instead upgrade the base-table drops on heroic, so that PRD can be closed or reframed.


ryan-foo:
## Verification screenshots (offline client, `scripts/heroic_loot_shot.mjs`)

**Heroic variant upgrade** (base item level 26 to Heroic item level 28, stats +8/+10 to +9/+11, composed "Heroic {name}"):

| Base drop | Heroic variant |
|---|---|
| base chest | heroic variant |

**Weapon dps on-curve** (the flagged case: Gravewyrm Cleaver now 16.0 dps, above the item-level-26 Wyrmfang Greatblade at 15.0, still under the legendaries):

| Base weapon (ilvl 26) | Heroic weapon (ilvl 31) |
|---|---|
| base weapon | heroic weapon |

**Soulbound mark** (new tooltip line; the "Right-click to destroy" affordance is suppressed for bound items):

soulbound mark

Bags overview: bags

ryan-foo:
Added handling for items already on the auction house when they become soulbound: on market load, any listing whose item is now soulbound (heroic_mark) is swept off the book and returned to the seller's collection, so the owner reclaims it at the Merchant. A soulbound item can't be traded or mailed, but a Merchant pickup hands it straight back to the player it's bound to, which is allowed. It mirrors the existing expired-listing return, draws no rng, and is idempotent. Covered by a new market test (parity + tsc green).

jgyy (commented):
This is a big, well-organized PR. The soulbound sweep is the standout: discard, sell, sell-all-junk, market listing (with a load-time migration to reclaim any already-listed soulbound items), mail, and trade offers are all gated on the new flag, that's a complete pass across every item-transfer surface, not a partial one. `item_budget.ts` was correctly pulled out as a leaf module to break an import cycle rather than papering over it. Determinism looks clean: no `Math.random`/`Date.now` in the touched sim files, and the loot-roll comment noting "no rng drawn here, normal-run draw order and parity goldens untouched" checks out against the code. Tests are solid: `tests/heroic_loot_flair.test.ts` and `tests/heroic_soulbound.test.ts` cover the two headline mechanics, and the whole touched set passes (148 passed / 3 skipped).

A couple of small things:

- `src/sim/types.ts` (`BaseItemDef.heroicOf` doc comment) says the client composes the display name as "Heroic {base name}", but the actual implementation (and the `heroic_variants.ts`/`entity_i18n.ts` comments) keep the base name and add a `[HEROIC]` tooltip tag instead. Should-fix: stale doc comment, not a behavior bug.
- `src/sim/loot/loot_roll.ts`, `heroicClaim` does an `ctx.instances.find(...)` scan per `rollLoot` call. Fine at the current instance cap, but worth a comment if instance counts ever grow, since there's no test exercising many concurrent heroic instances to catch a regression there.

Nothing blocking. The soulbound/loot flair mechanics and the module boundaries here are handled carefully.

PR #1706, opened 2026-07-09
fix(dungeons): never aggro a pack on dungeon entry

ryan-foo:
## Problem

Entering a dungeon dropped the player at the arrival point (instance-local z=4), which sat within aggro range of the first pack, so you got pulled the instant you zoned in. Measured entry-to-nearest-spawn distances vs the mobs' aggro radius:

- **Drowned Temple**: guards at 10.4 and 11.4 yd (aggro 12) pulled immediately.
- **Gravewyrm Sanctum**: boneguards at 12.4 and 13.3 yd (aggro 12) pulled under any heroic level scaling (heroic pins mobs to level 20, which widens the aggro radius against an under-20 player).
- Hollow Crypt and Sunken Bastion were borderline at 14.3 yd.

This is worst on heroic, which is exactly where it is most frustrating.

## Fix

Mob aggro radius is clamped to at most 20 yd (`mob/locomotion.ts`: `Math.min(20, aggroRadius + leveldiff * 1.5)`), so a spawn strictly beyond 20 yd from the entry can never aggro on arrival at any level difference.

- Move the arrival point back from z=4 to **z=-2** for all four five-player dungeons (Hollow Crypt, Sunken Bastion, Gravewyrm Sanctum, Drowned Temple). It stays 4 yd clear of the z=-6 exit portal (trigger radius 2.0, so no accidental instant-exit) and is well inside the room, whose back wall is z=-19.
- Push the front packs in Drowned Temple (z 14/15 to 20/21) and Gravewyrm Sanctum (z 16/17 to 20/21) deeper, so their nearest spawn clears 20 yd too.

Result: every dungeon's nearest spawn is now beyond the 20 yd aggro clamp, so zoning in never pulls a pack, at any difficulty.

## Testing

- New guard `tests/dungeon_entry_clearance.test.ts`: for every dungeon with spawns, asserts no spawn is within 20 yd of the entry, so a future spawn or entry edit cannot silently reintroduce the pull.
- Green locally: `dungeons` (incl. the ghost corpse-run that resurrects at the entry), `fixes`, `architecture`; `tsc` clean; wiki content unaffected (spawn/entry coordinates are not in the spoiler-safe guide).


jgyy (commented):
Nice fix, and nice diff hygiene: the huge stat (+861/-741) had me expecting a bundled refactor, but it's almost entirely regenerated golden traces from the coordinate shift, the actual logic is a tight 55-line pure module (`src/sim/dungeon_door_clearance.ts`) plus a small call-site swap. Good test-first workflow too: `tests/dungeon_entry_clearance.test.ts` pins both halves of the fix (interior spawn-to-entry distance, and the new door-clearance projection for exterior camp spawns) against the real `MAX_AGGRO_RADIUS` clamp rather than a made-up number. Ran it plus `tests/architecture.test.ts` and the parity suite loc, all green. No rng, no wall-clock, draw order confirmed unchanged by the inline comment and by inspection.

One small thing: `DOOR_CLEAR_RADIUS` (20) in the new module and `MAX_AGGRO_RADIUS` (20) in `mob/locomotion.ts` are the same value by convention rather than a shared imported constant. There's a comment cross-referencing them, which helps, but a future change to the aggro clamp could silently desync the two. Should-fix, not blocking.

No true merge conflicts against release/v0.24.0. Good to merge as-is.

Rubsey (commented):
The fix is real and well verified: the dungeon_instances golden flips the entry mob from chase-on-entry to idle, the content edits are coordinate-only, the new dungeon_door_clearance.ts module is pure and correctly seamed, and the full parity suite passes at the head (164 goldens). The 12-golden re-mint is legitimate: the door-clearance projection moves about 19 overworld camp mobs world-wide (not just the four dungeons in the title), and their downstream wander draws shift the shared rng stream, which is exactly why unrelated scenarios re-minted.

Findings:

- should-fix: tests/dungeon_entry_clearance.test.ts:19 (also src/sim/dungeon_door_clearance.ts:16 vs src/sim/mob/locomotion.ts:246,263). The 20 yd aggro cap is a bare literal in three unlinked places (four counting the pre-existing copy in pet_ai.ts:148). Retuning the clamp in locomotion.ts would silently under-protect both the module and the guard test, defeating the test's own stated purpose. Export the cap from locomotion.ts and import it in the module and the test.
- should-fix: src/sim/content/dungeons.ts:552. "Never aggro on entry" is not airtight: idle mobs wander up to 9 yd off spawn (locomotion.ts:284), so a front pack at 22.2 yd can stand about 13 yd from the entry mid-wander, inside the clamped heroic radius, and pull a late joiner or a corpse-run re-entry. First zone-in is genuinely fixed (mobs start at spawnPos). Either push the front pairs past the 20 + 9 yd envelope, or soften the guarantee in the test header and description.
- nit: src/sim/sim.ts:1362. The "no rng, draw order unchanged" comment is true for the spawn loop only; the drive-phase draw order does change (that is why 12 goldens re-minted). Worth rewording.
- nit: src/sim/dungeon_door_clearance.ts:38. Camps that sit fully inside a door ring (captain_verlan, wraithbinder_maldrec) get every spawn projected onto the ring edge, 8-12 yd off their designed centers and compressed onto an arc. Deterministic and safe, but quest hint placement may look off.

FernandoX7 (commented):
Solid fix with a correct root-cause writeup. I verified the load-bearing claims: the aggro clamp really is `Math.min(20, aggroRadius + leveldiff * 1.5)` (`mob/locomotion.ts:246,263`), and `delveDetectMult` only ever returns 1 or 0.65, so the effective aggro radius can never exceed 20 yd. The new `dungeon_door_clearance.ts` is a clean sibling module: pure, no rng, and applied to the already-rolled point so the camp draw order is untouched. Parity goldens, the architecture purity guard, and both new guards pass at head locally.

The interior half of the fix is airtight: entry moved to z=-2 and the front packs pushed to z20+ are static coordinates, so `dungeon_entry_clearance` holds for every seed and every level difference (including the heroic level-20 pin). That is the actual reported bug and it is nailed down well.

**should-fix** The overworld door half is seed-fragile. `projectOutsideDungeonDoors` runs before `findSafePos` (`sim.ts`, camp-spawn block ~L1398), and `findSafePos`'s outward spiral relocates a ring-edge point that landed on unsafe ground back inland, deep inside the ring. The Glimmermere moongate door (-70,792) is on the shore, so its water-side ring edge is below the water line and findSafePos walks the mob back toward land. Measured across seeds: seed 2024 puts a Glimmermere Wader 14.55 yd from the moongate, and seed 99 puts a Restless Bones 18.04 yd from the Hollow Crypt door, both inside the 20 yd clamp. The shipped seed (20061) happens to land all five door-adjacent mobs at exactly 20.00 yd, so there is no live-player impact today, but the "never on the door" guarantee is not robust. Re-checking clearance after findSafePos (project, find-safe, then re-project if it re-entered a ring), or having findSafePos reject ring-interior candidates, would close it.

**nit** The live-world door test only exercises seed 7 (clean at 20.00) with a `DOOR_CLEAR_RADIUS - 0.5` tolerance, so it would not catch the case above (seed 2024's 14.55 is well under 19.5). A short loop over a few seeds, or asserting the projection output itself lands exactly on the ring, would pin the real invariant. (`tests/dungeon_entry_clearance.test.ts`)

**nit** The clearance also relocates the named Captain Verlan camp (center 92,90, r4, entirely inside the Hollow Crypt door ring) out to 20 yd every world, not just the anonymous packs. Presumably intended, but worth a line in the PR body confirming named placements are meant to move too.

One mechanical note, not a design issue: the PR is currently conflicting, and the seven true conflicts are all parity golden JSONs (c4b_effect_dispatch, drowned_litany, dungeon_instances, multi_class_heal, nythraxis_full_pull, solo_mage, warlock_pet). That is regen churn from the base advancing past your last merge; merge release/v0.24.0 and re-run the golden regen.

FernandoX7 (approved):

PR #1707, opened 2026-07-09
fix(worldboss): silence Thunzharr's per-mechanic barks, one zone-wide yell every ~45s

ryan-foo:
## Summary

Thunzharr spammed the combat log with a bark for every mechanic every few seconds:

```
Thunzharr, the Waking Peak unleashes Tectonic Heave!
Thunzharr, the Waking Peak unleashes Seismic Stomp!
Thunzharr, the Waking Peak becomes enraged!
Thunzharr, the Waking Peak unleashes Mountainhide!
```

This adds a `quietMechanics` MobTemplate flag that suppresses those per-mechanic log barks (the "<Name> unleashes <Mechanic>!" lines from stomp, knockback, stoneskin, terrify, rally, and the bigCast completion, plus "<Name> becomes enraged!"). The mechanics themselves are unchanged: they still fire with their spellfx novas and damage, only the noisy log line is gone. Thunzharr opts in.

His only periodic voice is now the zone-wide (350yd) battle cry, tightened from every 60s to about every 45s, so the whole zone still hears him without the log filling up. The one-shot engage/summon/enrage yells and the Stormcall cast-bar telegraph (with its warning yell) stay.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/world_boss.test.ts` (42 passed): new tests pin that the boss fires its mechanics (spellfx novas) while emitting zero "unleashes"/"becomes enraged" log lines, and that the template opts into quietMechanics with a 45s battle cry. The Stormcall hardcast test now detects the cast via the cast-bar falling edge instead of the removed bark.
  - `npx vitest run tests/localization_fixes.test.ts tests/architecture.test.ts tests/world_boss_cc.test.ts tests/parity` (219 passed): the S3 i18n guard stays green (the "unleashes"/"enraged" literals remain in place for ordinary mobs), and parity is unchanged (gating a log emit draws no rng).
- Manual steps: none; the change is a data flag plus emit guards.

### DIFF
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -847,6 +847,11 @@ export const ZONE3_MOBS: Record<string, MobTemplate> = {
     // the straight line through fences, buildings, and the waterline, so he can
     // always go directly at his target and never wedges on a collider.
     phasesThroughObstacles: true,
+    // His only periodic voice is the battle cry below (every ~45s, zone-wide). The
+    // per-mechanic log barks ("unleashes Seismic Stomp/Tectonic Heave/Mountainhide!"
+    // and "becomes enraged!") are silenced so an overworld pull does not spam the
+    // combat log; the mechanics still fire with their spellfx and damage.
+    quietMechanics: true,
     ccImmune: true,
     // A raid boss cannot be perma-snared by a wall of Frostbolts / Hamstrings: slows do
     // not stick to him (ccImmune already blocks stun/root/etc; slow is separate).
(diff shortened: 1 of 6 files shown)


jgyy (commented):
Small, clean, well-tested. The `quietMechanics` flag is opt-in and scoped to Thunzharr only, so other bosses' logs are untouched, and it correctly distinguishes the log bark (silenced) from the chat yell (kept, cadence bumped 60 to 45). The timer stays tick-based, no new rng, and no i18n obligation since the gated lines are removed calls rather than new text.

Test coverage is decisive rather than a smoke test: a full pull that forces the enrage threshold and asserts `barks === 0` while `bossSpellfx > 0`. The existing cadence test was correctly updated to detect the ability via cast-bar falling edge now that the log line it used to key off of is gone, that's a necessary fix, not a weakened assertion.

No true conflicts against release/v0.24.0, no blocking or should-fix items. Good to merge.

Issue #1701, opened 2026-07-09
Professions UX audit: title header concatenation, crafting list scanability, town focus open gate


jgyy:
## Problem

The professions system (gathering, crafting, monster harvesting, town focus) is built and
playable, but a screenshot walkthrough of the live client surfaces several rough edges that
were not caught by the original design/build passes. None of these are new mechanics: they
are UI/UX and copy issues on top of the existing implementation.

This issue is a visual audit with the concrete findings below, filed as one tracking issue
so they can be triaged and split into separate fix issues as needed.

```mermaid
flowchart LR
    A[Gather node in world] -->|harvestNode| B[Gathering proficiency\nChar window]
    C[Kill + loot corpse] -->|componentTags| D[Corpse harvest\nfocus picker]
    B --> E[Crafting window\nrecipe list]
    D --> E
    E -->|Craft| F[Item produced]
    G[Town Focus window] -.->|bonus yield weighting| A
    G -.-> D
```

### 1. Character window: missing separator between class and title

Opening the character window (`C`) renders the header as `HunterTitle: None` with no space
or punctuation between the class name and the "Title:" label.

![Character window title concatenation](https://raw.githubusercontent.com/levy-street/world-of-claudecraft/pr-assets-professions-audit/shots/profissue_03_char_gathering_proficiency.png)

The Gathering section underneath it (Mining / Logging / Herbalism) reads fine.

### 2. Crafting window: flat, unfiltered recipe list with no craftability signal

The Crafting window (`T`) lists every known recipe (weapons, armor, food, potions, tools)
in one long undifferentiated list. Every row shows a "Craft" button with no visual
difference between a recipe you can afford right now and one you are missing reagents for,
and no way to see required reagent counts without a hover/attempt.

![Crafting window flat recipe list](https://raw.githubusercontent.com/levy-street/world-of-claudecraft/pr-assets-professions-audit/shots/profissue_04_crafting_window.png)

As the recipe count grows (13+ already, more planned per the professions design doc), this
will only get harder to scan.

### 3. Town Focus window has no location gate on opening

The Town Focus window (opened from the minimap button) can be opened from anywhere in the
world, not just in town. It shows a red "You must be in town to set your focus" message,
but the `+`/`-` allocation buttons remain visually enabled (not greyed out) even though
allocating focus away from town has no effect.

[...]

PR #1708, opened 2026-07-09
feat(content): add professions onboarding quest, fix crafting/char window UX (#1701)

jgyy:
## Summary

The professions system (gathering, crafting, corpse harvesting, town focus) is
fully built and playable, but nothing in the starting flow ever pointed a new
player at it: no quest, no NPC dialogue, no in-mechanic gate mentions it.
Gathering itself has zero level/quest/tool gate, so there was no natural
"unlock" moment to hang onboarding off before this.

Adds `q_prof_intro` ("A Trade for Every Hand"), a level-1 quest with no
prerequisite, given and turned in by `foreman_odell` (the existing mine-themed
NPC). Its `collect` objective targets `chunk_of_ore`, a dedicated `kind:
'quest'` item granted only by an ore-node harvest while the quest is active
(`NODE_QUEST_GRANT` in `src/sim/professions/gathering.ts`), so it is only
satisfied by really mining, not by any unrelated item gain, a mob kill, a
salvage roll, or the market (see the review update below). The turn-in text
nudges the player toward logging, herbalism, crafting, and the Town Focus
board.

```mermaid
flowchart LR
    A[New player, level 1] -->|talks to| B[Foreman Odell]
    B -->|offers| C["q_prof_intro: A Trade for Every Hand"]
    C -->|mine 5 ore at a node| D[Gathering: Mining]
    D -->|turn in to Odell| E[XP + copper reward]
    E -.->|completion text nudges toward| F[Logging / Herbalism]
    E -.-> G[Crafting window]
    E -.-> H[Town Focus board]
```

Also fixes two of the three findings from #1701 (verified live against the
current client, not just the issue description):

- **Character window title separator** (`src/ui/char_window.ts` /
  `src/styles/base.css`): the class/title header rendered with no space
  between "Hunter" and "Title: None". Adds a small `margin-left` on the
  archetype-title/hobby-craft spans, purely cosmetic, no i18n or gameplay
  impact.
- **Crafting window: flat list, no craftability signal**
  (`src/ui/crafting_window.ts` / `src/styles/components.css`): 13+ recipes
  rendered as one undifferentiated list, reagent counts were aria/tooltip-only
  (never shown inline), and the Craft button had **no `:disabled` styling at
  all**, so an unaffordable recipe looked identical to a craftable one.
  - Groups recipes into per-profession sections (reusing
    `archetypeTitleText`, the same id-to-name table the character window's
    title already uses, so the two surfaces can't drift).
  - Shows reagent counts inline under the recipe name (not just on hover).
  - Adds `.vendor-item:disabled { opacity: 0.45; }` so an unaffordable recipe
    is visually obvious without hovering.

The third #1701 finding (Town Focus `+`/`-` controls staying enabled outside
town) **does not reproduce** against current code:
`town_focus_view.ts`'s `canIncrease`/`canDecrease` and the Save button are
already gated on `inTown`. Left alone; flagging in case the original report
was against an older build.

## Related issues

Investigates and closes the professions-onboarding gap reported alongside
#1701, and fixes 2 of 3 #1701 UX findings.

## Type of change

- [x] Feature: new functionality (the onboarding quest)
- [x] Bug fix (crafting window craftability signal, char window title spacing)
- [ ] Documentation
- [x] Tests
- [x] Build, CI, or tooling (extends `pr_screenshots.mjs`'s change-aware
      targets with `crafting` and `char-window`, so future diffs touching
      those files auto-capture screenshots)

## How was this tested?

- Commands: `npm run gate` (full CI-equivalent gate: i18n gen + freshness,
  malware scan, changed-files biome, full Vitest suite, `tsc`, all builds).
  Green aside from 2-3 known-flaky server/http and mail tests that also fail
  identically on a clean `release/v0.24.0` checkout in this environment
  (timeout/resource-contention flakes, not a regression: confirmed by
  re-running each in isolation, where they pass).
- New test: `tests/prof_intro_quest.test.ts` (7 cases) - content wiring
  (giver/turn-in NPC, level-1 availability, quest order, NPC's `questIds`,
  pinned reward/count literals), a `Sim`-level pair proving an ore-node
  harvest grants `chunk_of_ore` only while `q_prof_intro` is active and never
  otherwise, and a functional case confirming the `collect` objective
  actually promotes to `ready` via the same `onInventoryChangedForQuests`
  credit path every other collect quest uses.
- Manual: ran the offline client (`npm run dev`), opened the crafting window
  with a spread of gathered materials (screenshot below) and the character
  window (screenshot below); confirmed the grouping/inline reagent
  counts/disabled-opacity and the title spacing render as intended.

## Screenshots / recordings

**Crafting window** - grouped by profession, reagent counts inline, greyed-out
unaffordable recipes (Gravewyrm Gauntlets, Boundstone Helm, Elixir of the
Bear, Thorium Mining Pick):

Crafting window grouped

**Character window** - visible spacing between "Hunter" and "Title: None":

Character window title separator

**Heroic vendor window** - `.vendor-item:disabled` also applies here (an
unmentioned surface flagged in review): the 12-mark rows are affordable at the
player's 12-mark balance, the 16-mark rows are greyed out:

Heroic vendor disabled rows

---

## Update: addressed review feedback

- **blocking (quest item anchor), fixed.** `q_prof_intro`'s collect objective
  now targets a dedicated `chunk_of_ore` quest item (`kind: 'quest'`,
  `sellValue: 0`), never `bone_fragments`. `bone_fragments` also drops from
  `restless_bones` (60%) and two zone-1 bosses (100%), is the salvage yield
  for every common-quality piece, is a deliberately-held crafting reagent, and
  trades on the world market, so the old objective could be completed without
  ever mining, and the vendor's Sell Junk button (which spares `kind:
  'quest'` but not junk) could silently demote a `ready` quest back to
  `active`. `chunk_of_ore` is granted only by an ore-node harvest while
  `q_prof_intro` is active (`NODE_QUEST_GRANT` in
  `src/sim/professions/gathering.ts`, mirroring the mob-loot `questId` gate
  `ghostly_essence` already uses), fixes the tracker/toast/bag-icon name
  mismatch as a side effect (the tracker no longer disagrees with what the
  node actually granted), and can't be sold or lost to the vendor. The ore
  node's ordinary `bone_fragments` grant is untouched, so the crafting/salvage
  economy doesn't move.
- **should-fix (screenshots), fixed.** Both links 404'd because the PNGs were
  never committed. Retook all three shots against the current client and
  committed them under `docs/screenshots/prof-onboarding/`; the two original
  links now resolve and a third (heroic vendor) covers the previously
  unmentioned `.vendor-item:disabled` surface.
- **nit (test pins), fixed.** `tests/prof_intro_quest.test.ts` now pins the
  literals (`count: 5`, `xpReward: 150`, `copperReward: 50`) instead of
  asserting `>0`, and the mining-exclusivity describe block gained real
  `Sim`-level coverage: harvesting the ore node while the quest is active
  grants `chunk_of_ore` (and still grants `bone_fragments`), harvesting
  without the quest active grants neither.
- **nit (crafting_window.ts comment), fixed.** Corrected: it's `COMBO_RECIPES`
  that revisit an already-seen craft, not the hub-tier `TOOL_RECIPES` (all
  engineering); also noted that section headers render the practitioner title
  ("Tinkerer"), not the craft name, so the hub-tier tools land under
  "Tinkerer" by design.
- **nit (heroic vendor `.vendor-item:disabled`), covered.** Confirmed
  intentional/beneficial and added the third screenshot above.
- **nit (`base.css` hanging indent).** Left as-is per your note that either
  approach is fine.

---

### Quality

- [x] **Tests pass.** `npm test` is green (see flaky-test note above). Added
      `tests/prof_intro_quest.test.ts` for the new sim content.
- [x] **Builds succeed.** Verified via `npm run gate` (which runs the full
      build matrix).

### Cross-platform

- [x] **Tested on desktop and mobile.** The crafting/char-window changes are
      CSS-only additions to shared rules (`.vendor-item:disabled`,
      `.char-archetype-title` margin) with no mobile-specific override needed;
      `hud.mobile.css` already has its own `#crafting-window` rules untouched
      by this diff.
- [x] **Accessible.** No change to keyboard/focus/ARIA behavior; the crafting
      button's existing `aria-label` (already includes reagent requirements)
      is unchanged, the new inline reagent line is visible to sighted and
      screen-reader users alike.

### Localization (i18n)

      locale.** `q_prof_intro`'s quest text/title/completion/objective label
      are wordy (M16), so `zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU` fills are
      included in this same change; the remaining locales are English-filled
      by the build and marked `pending` (maintainer batch-fills at release,
      per the standard PR-tier exception).
- [x] Numbers, money, dates, units, and percents go through the i18n
      formatters. N/A for this diff (no new numeric display).
- [x] Player-facing text emitted from `src/sim/` (the new quest content) is
      re-localized at the client boundary via `world_entity_i18n.ts`'s
      `QUEST_IDS` registry in this same change;

### Hygiene

- [x] I didn't hand-edit generated files; `i18n.resolved.generated/`,
      `i18n.status.json/summary.json`, `i18n.resolved.sha256`, and
      `content.generated.ts` are all regenerated via the build scripts.


FernandoX7 (commented):
The window fixes here are good: the crafting grouping reuses the archetype title table so the two surfaces can't drift, the reagent line is finally visible without hover, all seven recipe professionIds resolve in ARCHETYPE_TITLE_KEYS, and the new screenshot targets follow pr_shot_targets.mjs conventions exactly. I also confirmed the Town Focus non-repro claim (town_focus_view.ts:46 gates canIncrease/canDecrease on inTown, and the Save disable in town_focus_window.ts:85 pre-dates this PR), and the quest wiring itself is clean data: no new imports, rewards in line with q_wolves, no QUEST_ORDER pin breaks (progression, zone1_removed_content, and the new test file all pass at the PR head).

The quest's target item is the wrong anchor though, and it undoes the point of the feature.

**blocking** src/sim/content/zone1.ts:702: bone_fragments is a shared junk item, so the quest is not "only satisfied by really mining" (the new comment at zone1.ts:689 repeats this claim), and the fiction breaks on several player-visible surfaces at once.

- Any acquisition credits a collect objective (quest_credit.ts:51 counts current inventory), and accepting credits from what you already hold (quest_commands.ts:108). bone_fragments drops from restless_bones at 60% and two bosses at 100% in zone 1 alone (zone1.ts:388/433/468, plus zone 2/3, dungeons, temple), is the salvage yield for every common-quality piece (salvage.ts:38, and the :91 fallback), is a crafting reagent players deliberately hold (recipes.ts:43/55/77), and trades freely on the world market. A player with 5 in their bags completes "Mine them yourself, mind; I'll know the difference" the instant they accept, and anyone can buy completion for coppers.
- The name mismatch shows up in the happy path too: the tracker says "Chunk of Ore: n/5" while mining toasts "You receive: Bone Fragments." and the bag shows a grey junk item (items.ts:1312).
- Because it is kind 'junk' and poor quality, the vendor's Sell Junk button sells the player's quest progress (items.ts:371 only spares kind 'quest'), silently demoting ready back to active (quest_credit.ts:81). Every other collect quest in the zone targets a kind 'quest' item (greyjaw_fang, boar_hide, webwood_silk, ghostly_essence, blessed_wax), several drop-gated by questId.

The convention-following fix is a real quest item: a kind 'quest' "Chunk of Ore" granted by the ore-node harvest while q_prof_intro is active, the same questId gating the ghostly_essence drop uses, which leaves the reagent economy untouched. The bigger alternative is the dedicated ore item the gathering.ts placeholder comments already anticipate, but that ripples into recipes and salvage and is probably its own PR. Rewording the quest to honestly ask for Bone Fragments would fix the naming but not the "no mining required" hole, so I would not stop there.

**should-fix**: both PR-body screenshot URLs return 404 and no screenshots are committed. Visual changes need before/after screenshots committed under docs/screenshots and referenced from the PR body; right now the new crafting layout (including how the first vendor-section-title's border-top renders directly under the panel title) can't be reviewed visually.

Nits:

- tests/prof_intro_quest.test.ts:59: the count/reward pins can't fail: count is only asserted > 0 and the promotion loop derives its bound from the same def, so a 5-to-1 mutation passes while the quest text still says 5. Pin the literals (5 / 150 / 50). Same file: the "mining actually satisfies" describe title overclaims; it pins itemId consistency with NODE_HARVEST_TABLE (the right test), but that shows consistency, not mining-exclusivity.
- src/ui/crafting_window.ts:41: the grouping comment misattributes the interleaving: the hub-tier TOOL_RECIPES are all engineering (a first-appearance craft); it is COMBO_RECIPES that revisit already-seen crafts. Also worth knowing the section headers render the practitioner titles ("Tinkerer", "Chef"), not craft names, so the six hub-tier tools group under "Tinkerer". Fine if intended.
- src/styles/components.css:2932: .vendor-item:disabled also restyles the heroic vendor's unaffordable rows (heroic_vendor_window.ts:46). Probably an improvement, but it is an unmentioned surface; the committed screenshots would cover it.
- src/styles/base.css:556: margin-left leaves a 6px hanging indent if the title line ever wraps; the adjacent level/class subtitle handles the same spacing with a literal space in the template (char_window.ts:193). Either is fine, just noting the existing pattern.

jgyy:
Thanks for the thorough review, especially catching the sell-junk/dupe hole. Pushed a follow-up commit addressing everything:

**blocking (quest item anchor)** - fixed with the convention-following option you called out. `q_prof_intro`'s collect objective now targets a new `chunk_of_ore` item (`kind: 'quest'`, `sellValue: 0`), granted only by an ore-node harvest while the quest is active (`NODE_QUEST_GRANT` in `src/sim/professions/gathering.ts`), gated the same way `ghostly_essence` is gated off `q_rite`. `bone_fragments` is untouched everywhere else (crafting/salvage economy unaffected), and the quest can no longer be completed by holding junk, buying from the market, or bought-then-sold through Sell Junk (kind `'quest'` is spared there). As a side effect the tracker/toast/bag-icon name mismatch you flagged is gone too, since the tracked item and the granted item are now the same thing.

**should-fix (screenshots)** - fixed. Both PNGs were never actually committed, hence the 404s. Retook all shots against the current client and committed them under `docs/screenshots/prof-onboarding/`. Also added a third shot of the heroic vendor window per your nit below, since I was already in there.

**nits:**
- Pinned the `count`/`xpReward`/`copperReward` literals in `tests/prof_intro_quest.test.ts` instead of `>0`, and added real `Sim`-level coverage that an ore-node harvest grants `chunk_of_ore` while the quest is active and never otherwise (the old test only proved item-id consistency with `NODE_HARVEST_TABLE`, not mining-exclusivity, as you noted).
- Fixed the `crafting_window.ts` grouping comment: it's `COMBO_RECIPES` that revisit an already-seen craft, not the hub-tier `TOOL_RECIPES` (those are all engineering). Left a note that section headers render the practitioner title, so the hub-tier tools intentionally land under "Tinkerer".
- Confirmed the `.vendor-item:disabled` bleed into the heroic vendor window is a genuine improvement and screenshotted it (12-mark rows enabled, 16-mark rows greyed, at a 12-mark balance).
- Left the `base.css` hanging-indent alone per your note that either is fine.

Full detail is in the "Update: addressed review feedback" section I added to the PR description. `npx tsc --noEmit`, the targeted test files, and the pre-push QA floor are all green on the pushed commit.

FernandoX7 (commented):
Re-reviewed after d27fc4fe. All four findings from the first round landed properly, and I verified them against the head, not just the description. Good to merge once CI confirms.

What I checked:

- The collect anchor is right now. chunk_of_ore is kind 'quest' with sellValue 0 and questId q_prof_intro (src/sim/content/items.ts:1208), the objective targets it (src/sim/content/zone1.ts:708), and the tracker label finally matches what lands in the bag. NODE_QUEST_GRANT resolves the need before the timer-consuming resolveHarvest call, draws no rng of its own, and guards the grant with canAddItem, so the rng stream and respawn-timer semantics are untouched for anyone without the quest.
- Registering chunk_of_ore in progression.test.ts SCRIPTED_COLLECT_ITEMS is the right mechanism for a scripted-grant collect item.
- The new tests are decisive: pinned literals (count 5, xp 150, copper 50), a real Sim harvest granting both items while active, and the negative arm. One thing I checked rather than flagging: turn-in deletes the quest-log entry (turnInQuestCore), so the "already turned in" case in that test title is behaviorally identical to "never accepted"; the coverage is genuinely complete.
- Screenshots are committed under docs/screenshots/prof-onboarding/ and all three PR-body links resolve now. The pr_shot_targets.mjs size poll is a sensible bounded replacement for the fixed wait.

Two CI notes, neither an ask on you:

- The PR gate failure on tests/i18n_completeness.test.ts was ours to fix: locale fills are maintainer work in this repo, and the non-English itemNames blocks in i18n.catalog/items.ts are dead values (the build reads that catalog for English only; non-English render text comes from the src/ui/i18n.locales overlays). I pushed 6202769bb moving your five item-name translations into the overlays and regenerating the artifacts.
- The "Screenshots of changes" failure was the workflow canceling itself, not your change: the Codex review job's sticky comment fires an issue_comment run in the same concurrency group with cancel-in-progress, which killed the still-running capture step. That job is informational, not a required check.

Thanks for the thorough response commit; the NODE_QUEST_GRANT comment mirroring the ghostly_essence precedent made this quick to verify.

FernandoX7 (approved):

PR #1709, opened 2026-07-09
feat(ui): target-of-target frame + assist key (T)

ryan-foo:
## Summary

Adds a **target-of-target** unit frame and an **assist** hotkey.

- A compact ToT frame (a fourth `unit_frame` family instance: portrait + name + hp) anchored below the target frame, showing who the current target is targeting: a mob's aggro target (the tank, or you), or a targeted player's target. When a mob is on you it shows your own frame, the classic "this mob is on me" tank signal.
- The target's target is resolved by a pure, tested `targetOfTargetId` helper (mob `aggroTargetId` first, then player `targetId`). A player's target now rides the interest-scoped wire as `tgt` (mobs already sent `aggro`); applied in `applyWire` on both hosts.
- A new **Assist** keybind (default **Z**) selects the current target's target, routed `onUiKey('assist')` to `hud.assistTarget()`.

Note on the key: every letter A to Y was already bound and F is interact, so Z (the only free letter) is the default; it is rebindable in options. The keybinds test's unbound-key fixture moved from Z to Backquote.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/target_of_target.test.ts tests/keybinds.test.ts` (50 passed)
  - `npx vitest run tests/architecture.test.ts tests/snapshots.test.ts tests/world_api_parity.test.ts tests/hud_perf_budget.test.ts tests/i18n_completeness.test.ts tests/css_corpus.test.ts` (371 passed): the wire round-trip (tgt), the per-frame perf budget (the new painter uses the elided writers), i18n completeness, and the CSS corpus all stay green.
- Manual/visual: drove the offline client (scripts/tot_frame_shot.mjs), targeted a mob set to aggro the player, and confirmed the ToT frame renders the player below the target frame (screenshot below).

## Screenshots / recordings

target-of-target frame

The WILD BOAR target frame with the ADVENTURER (player) target-of-target frame below it, because the boar is targeting the player.


ryan-foo:
Updated: the Assist default is now T (was Z). T was Crafting, so Crafting moves to the now-free Z (both rebindable). Also confirmed this branch and the party-frames PR (#1713) merge cleanly against each other, no shared-file conflict, and the target-of-target frame (top-left, below the target frame, z-index 6) does not overlap the party frames (which #1713 puts at z-index 7), so the two play well together.

jgyy (commented):
The core of this is well done: `src/ui/target_of_target.ts` is a clean, DOM-free pure function with real precedence tests (mob aggro, player-target fallback, null cases), and the wire threading is minimal and correct, the server already tracks `targetId` per entity so this just mirrors it through `wireEntity`/`applyWire` without needing an `IWorld` extension. play.html got the same `#tot-frame` markup and aria attributes as index.html, so the usual play.html-parity miss didn't happen here. The one new string is a properly-scoped English `hud_chrome` key.

Two things worth a look before merge:

- `src/ui/hud.ts` grew by about 83 lines for the wiring, the new `UnitFramePainter` instance, `drawTotPortrait`, and `assistTarget`. It does correctly reuse the `UnitFramePainter` family rather than inventing a new painter, but the wiring and paint-loop logic itself landed as another block in the ~10k-line coordinator rather than in `target_of_target.ts` or a small sibling module. Should-fix per the module-first guidance, not blocking.
- `src/game/keybinds.ts`: the default `T` binding is reassigned from `crafting` to `assist`, and `crafting`'s default silently moves to `Z`. That's a remap of an existing shipped default for all current players, not just a new binding being added. The tests were updated to match, so it's clearly deliberate, but it's worth calling out explicitly in the PR description since it'll surprise anyone with muscle memory for `T` = crafting.

Also worth a one-line confirmation in the PR body: `tgt` on the wire mirrors any tracked player's current target to any interest-scoped observer, not just party members, matching the existing mob-aggro exposure pattern. Probably fine, just flagging that it's a deliberate choice.

No true merge conflicts. Nothing blocking.

Rubsey (commented):
Solid feature, built idiomatically: the ToT frame is a true fourth UnitFramePainter instance over a tested DOM-free core, the tgt field round-trips (server/game.ts:726 encode, online.ts:1648 decode) with blankEntity shapes already matching, the frame markup is byte-identical in index.html and play.html, assist stays pure selection through the server-validated target command, and all the suites here pass at the head (tests, tsc, perf budget, css corpus).

Findings:

- should-fix: src/ui/options_window.ts:131. BIND_ACTION_LABEL_KEYS has no 'assist' entry, so the keybind row, rebind prompt, and aria labels render the raw English "Assist Target" instead of a t() key. This is the English wiring, so it is in scope here: add the catalog key and the map entry (the map's own comment documents this exact fallback class, previously fixed for talents/leaderboard).
- should-fix: src/styles/hud.mobile.css. No mobile rule for #tot-frame: the mobile sheet repositions and scales #target-frame and #party-frames, and the new frame's desktop left:12px/top:100px sits under both on a phone. Add a body.mobile-touch rule (reposition or hide).
- nit: src/ui/hud.ts:1023 and 3937. Two comments (and the PR body) still say the assist key is Z; the shipped default is KeyT (keybinds.ts:152).
- nit: index.html:246 / play.html:195. The static aria-label fallback "Target's Target" does not match the catalog English "Mark's Mark"; the sibling target frame keeps its fallback identical to its catalog value.
- nit: no test pins the tgt wire key (encode through applyWire); the new test drives only the pure helper, so a wire-key rename would not fail anything.
- nit: consider registering target_of_target.ts in UI_PURE_CORES and BARE_NAMED; the file's header claims the pure-core contract and registration is cheap.

The generated i18n conflict with the base is the standard regen churn: merge the base and re-run the i18n build.

ryan-foo:
Closing as superseded: the target-of-target frame already landed in release/v0.24.0 (src/ui/target_of_target.ts, the #totarget-frame UI + showTargetOfTarget option, and the `tgt` wire field), implemented independently. This PR's frame is a duplicate.

The one piece not in the base is the assist action (`assistTarget()`: select the current target's target via a keybind). If that's still wanted, it's a small standalone PR, and note its intended key T is now taken by Crafting, so it would need a different default (Z, per the code comment, is free).

PR #1710, opened 2026-07-09
feat(paladin): bring Mending Light mana cost in line with the other healers

ryan-foo:
## Summary

The paladin primary heal (holy_light, "Mending Light") cost noticeably more mana per point healed than the other healers. This lowers its costs across all ranks (35/60/95/140 to 25/50/70/115) so paladin mana efficiency matches the peer median.

Mana per average HP healed at level 20:

| Healer | mana/HP |
|---|---|
| Priest heal | 0.520 |
| Druid regrowth | 0.519 |
| Priest lesser_heal | 0.537 |
| Priest flash_heal | 0.573 |
| Druid healing_touch | 0.574 |
| Shaman healing_wave | 0.596 |
| Peer median | 0.555 |
| Paladin holy_light (new) | 0.558 |

Heal amounts and cast times are unchanged; this is a mana-cost tuning only.

## Type of change

- [x] Feature (balance tuning)

## How was this tested?

- `npx vitest run tests/spell_balance.test.ts tests/architecture.test.ts` (29 passed): a new spell_balance test pins the paladin heal costs and asserts the level-20 mana efficiency lands within 10% of the peer median and priest heal.


jgyy (commented):
Good, this one's actually backed by numbers rather than just asserted. At max rank the old mana-per-average-heal ratio for Mending Light was about 0.68, roughly 22% above the peer median (~0.555) across priest heal/flash heal/lesser heal, shaman healing wave, and druid healing touch/regrowth. The new costs land right on that median (~0.558) and track closely with priest's `heal` at the matching level bracket. Checked the intermediate ranks too, they hold up the same way.

The new test in `tests/spell_balance.test.ts` is the right shape: a golden pin on the literal cost array plus a comparative band test against the peer median, so it'll catch both accidental drift on this ability and future edits to peer heals that would push it back out of line.

No other references to this ability's cost elsewhere, nothing serialized, so no parity regen needed. No true conflicts. Nothing blocking, nice work tying the number to something verifiable.

PR #1711, opened 2026-07-09
fix(sim): group buffs do not stack, newest supersedes across casters

ryan-foo:
## Summary

Two casters applying the same group buff to one target used to leave two stacked auras. Now a given group buff replaces any existing instance of the same buff regardless of who cast it, keeping the newest (its value and duration win).

Covers the classic raid buffs: Blessing of Might, Arcane Intellect, Power Word: Fortitude, Mark of the Wild, Battle Shout, Devotion Aura. A new `src/sim/combat/aura_stacking.ts` module holds the source-independent buff list and the replacement rule; `applyAura` consults it. Non-group auras keep their normal per-source stacking. No new randomness.

## Type of change

- [x] Bug fix

## How was this tested?

- `npx vitest run tests/raid_buffs.test.ts` (14 passed): two casters apply the same blessing and only one aura remains, with the newest value and duration.
- `npx vitest run tests/architecture.test.ts tests/parity` (188 passed, 1 skipped): determinism/draw-order unchanged.

### DIFF
--- a/src/sim/combat/aura_stacking.ts
+++ b/src/sim/combat/aura_stacking.ts
@@ -1,0 +1,21 @@
+import type { Aura } from '../types';
+
+const SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS = new Set([
+  'arcane_intellect',
+  'battle_shout',
+  'blessing_of_might',
+  'devotion_aura',
+  'mark_of_the_wild',
+  'power_word_fortitude',
+]);
+
+export function auraReplacementConflicts(auras: readonly Aura[], aura: Aura): number[] {
+  const replaceAcrossSources = SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS.has(aura.id);
+  const out: number[] = [];
+  for (let i = auras.length - 1; i >= 0; i--) {
+    const existing = auras[i];
+    if (existing.id !== aura.id) continue;
+    if (replaceAcrossSources || existing.sourceId === aura.sourceId) out.push(i);
+  }
+  return out;
+}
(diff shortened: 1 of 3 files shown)


jgyy (commented):
Good, narrow fix. The new `auraReplacementConflicts` module only changes behavior for a named set of six standardized group buffs (source-independent supersede), and for everything else it preserves the exact old id-and-sourceId matching, so single-caster refresh-on-recast is untouched. That scoping matters, a blanket "same id supersedes regardless of source" change would have wrongly clobbered same-named DoTs/debuffs from different casters.

Determinism is clean: the new function only walks the plain `auras` array in insertion order, no Map/Set/object-key iteration, no wall-clock reads. "Newest supersedes" falls out naturally from remove-then-push. Two new tests in `tests/raid_buffs.test.ts` reproduce the old cross-caster-stacking bug and pin the fix, including duration/value/sourceId on the surviving aura. Ran the suite plus `tests/architecture.test.ts`, both green.

Two small notes, neither blocking:

- A few raw `.auras.push(...)` call sites (e.g. the mob `warcry_*` aura, the pet aura in `sim.ts`) bypass `applyAura` entirely and so don't get the new supersede semantics. Not a regression since none of them are in the group-buff set today, but worth a line in the PR body noting the fix only covers auras applied through `applyAura`.
- `SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS` duplicates the "six raid buffs" list that also exists in the test file. Low risk given how rarely this table changes, but a second source of truth to keep in sync if a seventh ever gets added.

No true conflicts. Good to merge.

Issue #1300, opened 2026-07-02
Professions: salvage and disenchant


Nervescraper:
## Problem

There is no end-of-life loop that breaks crafted or looted items back into materials, even though the doc's binding section already expects one ("off to a salvager once its owner is done with it"). The doc files Salvaging/Disenchanting as a future off-wheel gathering skill that feeds every craft (Future: gathering skills, add freely), not a craft-gated action; its home needs a maintainer call before implementation.

## Expected behavior

- A player can salvage or disenchant an eligible item into materials. Per the doc's framing this is an additive off-wheel gathering skill (usable by everyone, growing by use); if it is instead gated by the relevant craft (for example Enchanting for disenchant), that is a deliberate deviation to confirm first.
- The yield scales with the item's rarity and tier.

## Scope

- `src/sim/professions/gathering.ts` or a `salvage.ts` sibling: the salvage action and yield roll.
- Content: which items are salvageable and their yield tables.

## Acceptance criteria

- Salvaging an eligible item consumes it and yields materials per a scripted table; the roll uses `Rng`.
- Ineligible items cannot be salvaged.

## Notes

Feeds the material economy and the recharge and craft loops (an economy sink per the doc). Depends on #1127; disenchant yield feeds Enchanting content (#1125). Lower urgency than the other filings: the doc parks it in the future-gathering lane, so this is planning ahead, not a blocker.

Full design: https://woc.nervemart.com/docs/professions-system

jgyy:
Already implemented: `src/sim/professions/salvage.ts` implements salvage and disenchant end to end; the module header explicitly cites this issue. Closing as resolved.

PR #1712, opened 2026-07-09
feat(professions): add Enchanting profession (disenchant + apply-enchant)

jgyy:
## Summary

The user wants rares/epics to hold their value again as the economy fills up
with them; the ask was Enchanting as a real profession: disenchant surplus
gear for materials (an economy sink), and spend those materials to make a
kept piece meaningfully stronger via a permanent enchant.

`enchanting` already exists as a CRAFT_RING member (`content/professions.ts`),
has an archetype title (`hudChrome.archetypeTitle.enchanting`: "Enchanter"),
and already owns 3 tool-effect recharge perks (`tools.ts`) - but there was no
actual enchant recipe and no disenchant action. This PR fills that gap in.

```mermaid
flowchart LR
    A[Rare/epic weapon or armor] -->|disenchantItem| B[Arcane materials\ndust / essence / shard]
    B -->|applyEnchant + a plain copy| C[Freshly-instanced\nenchanted item]
    C -->|equipItem| D[recalcPlayerStats\nstat bonus applied]
    C -.->|unequipItem| E[Back to bags,\nenchant intact]
    F[Everyone: plain salvage] -.->|generic materials\nbone_fragments etc| G[Existing crafting recipes]
    B -.->|strictly better yield| F
```

### Design decisions (made without further back-and-forth, per your "don't
want to make decisions" ask - here's what I picked and why)

- **Layered on top of, not replacing, the existing salvage system**
  (`professions/salvage.ts`, #1300, everyone-can-salvage into generic
  materials). `disenchantItem` is the Enchanting-specific action: strictly
  better yield, three dedicated arcane materials (dust/essence/shard) tiered
  by the disenchanted item's rarity.
- **Enchants bind to the specific item instance, not the character or the
  item id in the abstract.** This is what makes an enchanted rare/epic a
  distinct, valuable, tradeable good (matching "items aren't soulbound, I
  like that"), rather than a slot-bound character buff that dies with the
  piece. It reuses `ItemInstancePayload.rolled.stats` - a field that already
  existed in `types.ts` for exactly this purpose but was never wired to
  anything. Equip/unequip (`items.ts`) now thread this instance data through
  `recalcPlayerStats` (additive `PlayerMeta.equipmentInstance`, persisted).
- **Ten always-known enchants, one per equippable slot** (weapon through
  ring), each a flat primary-stat/armor bonus. No skill gate beyond this
  repo's existing free-floor rule (every common-tier craft action is
  attemptable on materials alone; see `wheel.ts`). A weapon-damage enchant is
  explicitly out of scope for v1: damage rolls read the item DEFINITION's
  `weapon.min/max`, not per-instance data, which is a materially bigger
  change than this PR's stat-bonus path.

### Known limitations (flagged, not hidden)

- **Not yet wired onto a server WS command or a dedicated UI window** - the
  exact same scope this repo's `salvage.ts` shipped with (`lastSalvageResult`
  carries the same "not yet wired onto IWorld" comment `lastDisenchantResult`/
  `lastEnchantResult` now mirror). Sim-complete and fully tested; a future
  issue extends `IWorldProfessions` + `ClientWorld` + `server/game.ts` the way
  `craft_item`/`harvest_node` already are, plus adds a target-item picker.
- **Market listing and Trade/Mail transfer of an enchanted item don't work
  yet.** Both are built around fungible `{itemId, count}` operations
  (`market.ts` counts/escrows only `countFungibleItem`; `mail/post_office.ts`
  explicitly excludes instanced copies from stack-based sends) and would
  either refuse or silently drop an instanced copy today. This is a real gap
  against the "stays tradeable" goal - I'm flagging it as a follow-up rather
  than quietly leaving it broken or expanding this PR's scope to rework two
  other systems.

## Related issues

New feature (no tracking issue existed yet); fills in the previously-empty
`enchanting` slot on `CRAFT_RING`.

## Type of change

- [x] Tests

## How was this tested?

- Commands: `npm run gate` (full CI-equivalent gate). Green aside from a
  known-flaky `tests/server/http/parity.test.ts` timeout under full-suite
  resource contention, confirmed pre-existing and unrelated: it passes
  cleanly in isolation, and fails identically on a clean `release/v0.24.0`
  checkout under the same load in this environment.
- New test: `tests/professions_enchanting.test.ts` (12 cases) covering
  disenchant eligibility/yield/rarity-scaling, applyEnchant validation
  (unknown item/enchant, wrong slot, missing item/reagents), the
  consume-plain-copy + grant-instanced-copy round trip, and the full
  equip -> stat bonus -> unequip -> re-equip -> swap-for-plain-copy cycle
  (asserting the bonus applies/disappears/reapplies correctly and the
  enchanted piece is never silently flattened into a plain stack).
- Also re-ran the full equip/stat-affecting test surface
  (`item_sets`, `set_4pc_bonuses`, `set_proc_clearcasting`, `haste_set_bonus`,
  `over_level_gear_inert`, `recalc_primary_pct`, `caster_set_bonus`,
  `character_sheet`, `stat_tooltip`, `mob_purge`, `mob_swing`,
  `shapeshift_resource_persist`) since `recalcPlayerStats` gained a new
  trailing optional parameter touching ~13 call sites; all pass unchanged.

## Screenshots / recordings

N/A - this PR has no UI surface yet (see "known limitations" above); nothing
changed for a player to see until a follow-up wires a picker window.

---

### Quality

- [x] **Tests pass.** `npm test` is green (see flaky-test note above).
- [x] **Builds succeed.** Verified via `npm run gate`.

### Cross-platform

- [x] N/A. No UI surface in this change (sim + persistence only).

### Localization (i18n)

      locale.** The 3 new item names (Arcane Dust/Essence/Shard) are wordy
      enough that `tests/localization_coverage.test.ts` requires full,
      non-fallback coverage for the `item` entity group specifically (a
      stricter, always-on check distinct from the general M16 PR-tier
      exemption): added English to `i18n.catalog/items.ts`'s positional
      per-locale name arrays (covers en/es/fr_FR/it_IT/de_DE/zh_CN/zh_TW/
      ko_KR/ja_JP/pt_BR/ru_RU) plus `i18n.locales/<lang>.ts` overlay entries
      for the remaining locales (cs_CZ/nl_NL/pl_PL/id_ID/tr_TR/sv_SE/vi_VN/
      da_DK). The ten new `EnchantDef` names are backend-only content with no
      UI consumer yet, so they are not player-visible and carry no i18n
      obligation until the future UI issue renders them.
- [x] Player-facing text emitted from `src/sim/`: N/A, no new sim-emitted
      strings (item names route through the existing generic item-name
      i18n path, not a new emit).

### Hygiene

- [x] I didn't hand-edit generated files; `i18n.resolved.generated/`,
      `i18n.status.json/summary.json`, and `i18n.resolved.sha256` are all
      regenerated via the build scripts. `tests/parity/golden/*.json` are
      regenerated via `UPDATE_PARITY=1` (new `PlayerMeta.equipmentInstance`
      field legitimately changes the pinned state hash).


FernandoX7 (commented):
The core of this is done carefully: the module mirrors the salvage seam exactly, the single rng draw keeps determinism (event streams and draw counts in the regenerated goldens are byte-identical to base, only state hashes moved with the new field), the cloneItemInstancePayload extraction is behavior-identical, and the parity exclusion comes with the anti-loosening pin. Equipping a crafted signed piece also stops silently dropping its payload, a nice incidental fix. One blocker though, and it defeats the feature.

**Blocking**

1. The enchant bonus silently drops on level-up, buff expiry, talent spend, and every other stat recalc outside the files this PR touched. `recalcPlayerStats` gained `equipmentInstance` as an optional trailing parameter, and 12 call sites in the extracted modules never pass it: src/sim/combat/auras.ts:203 (stat-aura expiry), src/sim/combat/damage.ts:757 (level-up), src/sim/combat/effect_dispatch.ts:678 and :721 (stance/form toggle-off, self-buff cast), src/sim/progression/talents.ts:59 (spend/respec), src/sim/delves/runs.ts:635, src/sim/social/arena.ts:850, src/sim/social/vale_cup.ts:722, and src/sim/social/fiesta.ts:192/210/224/238. Reproduced live: enchant plus equip (+5 str), grant a level's worth of XP, and the bonus is gone until the next equip cycle. The level-up site then sets `p.hp = p.maxHp`, so a stamina-enchanted player's max HP shrinks on each ding, and the REST character sheet (server/character_sheet.ts:197) still shows the enchant, disagreeing with the live entity. Your tests stay green because every path they exercise is one of the updated sites. Suggest making the parameter required so tsc flags every current and future call site, plus a regression test through one untouched path (level-up is the easiest). If you fix sites individually, keep each site's existing mods argument as-is: the fiesta sites deliberately pass fiestaMods/talentMods, not playerMods.

**Should-fix**

2. Crafted rare-or-better gear can neither be disenchanted nor enchanted. crafting.ts:372 grants every rare+ single-copy craft as an instanced copy (signer plus rolled quality), and both gates here use `countFungibleItem`, which excludes all instanced slots (src/sim/professions/enchanting.ts:97 and :150). So the items this feature exists for, the rares and epics filling up the economy with rarity-roll crafting as a major source, get `not_held`, while plain salvage happily consumes them (salvage.ts uses countItem/removeItem). The "strictly better yield" framing inverts for crafted gear. The fungible-only gate is the right instinct for protecting already-enchanted copies, but it overshoots: gate on "instance has no rolled.stats" rather than "no instance at all".

3. Vendor sell and discard destroy the enchanted copy first. sellItem (src/sim/items.ts:359) and discardItem (:55) use plain `ctx.removeItem`, which scans highest index first, exactly where applyEnchant pushes the enchanted copy, and recordVendorBuyback (:319) keeps only {itemId, count}, so the payload is unrecoverable. Holding a plain and an enchanted copy of the same item is the workflow this feature creates, and selling or discarding "one" eats the enchanted one. The mechanism predates this PR (#1165), but the description's known-limitations list flags market/mail/trade and misses this pair. Preferring fungible copies in sell/discard would cover it. Related: trade's drop arm (src/sim/social/trade.ts:169) has the same highest-index preference, so it grabs the enchanted copy even when the giver meant to trade a plain one; you flagged trade already, but the module header's "stays a distinct, tradeable good" (enchanting.ts:6) is not true yet and worth softening.

4. The tests don't pin the two headline claims. Nothing saves a CharacterState with an equipped enchanted piece and loads it back asserting the bonus survives (the "permanent" in permanent enchant; the goldens only pin the empty `equipmentInstance: {}` shape), and nothing loads a pre-Enchanting save missing the key. Two weaker pins while you are in there: the rarity-scaling test (tests/professions_enchanting.test.ts:66) uses toBeGreaterThanOrEqual, so deleting the qualityIdx term still passes; and the equip-cycle assertions read the expected bonus from the same ENCHANTS constant production folds, so the magnitude itself is unprotected (pin `statBonus.str` to the literal 5 once, then assert baseStr + 5).

**Nits**

5. ENCHANT_IDS (src/sim/content/enchants.ts:109) and enchantsForSlot (src/sim/professions/enchanting.ts:179) have no consumers; house style is to add exports when the consumer lands. Easy to drop until the UI PR.

6. The equipItem comment "a freshly-enchanted copy is exactly what this picks up first" only holds while the enchanted copy is still the highest-index match; loot another plain copy of the same item afterwards and the plain one gets equipped instead. Deterministic, fine for v1, but worth softening so the future picker work is not misled.

For what it is worth I checked the buy-disenchant direction for a money loop and it is safe: the cheapest vendor-buyable disenchantable piece costs 900c against a maximum ~15c material yield.

ryan-foo (commented):
Clean, well-scoped groundwork PR. Determinism, sim purity, persistence back-compat, and the parity-trace handling are all correct, and the tests cover the tricky edges (equip/unequip enchant preservation, wrong-slot denial, all-or-nothing reagents, swap-drops-enchant). Nice use of the SimContext seam: the resolution logic is a pure sibling module and `sim.ts` is a thin consumer.

A few things I verified while reading, all clean: the enchant bonus flows through a fresh instanced copy (`addItemInstance` -> `rolled.stats`), `meta.equipmentInstance` round-trips through save/load with a `?? {}` back-compat default, every `recalcPlayerStats` call site got the new arg, and the golden churn is just the new `equipmentInstance: {}` entering the serialized state (all empty, no behavior change). The true merge conflicts are only the standard stale-base i18n regen trio.

Findings, all nits:

- nit `src/sim/types.ts:1937` (and `entity.ts:392`): `equippedInstances` is added to `Entity` and populated offline, but nothing in `src/render/` reads it and it is never encoded in `wireEntity`, so today it is dead weight on every entity and is always `{}` on the online path. Reasonable as forward-scaffolding for a future inspect-another-player feature, but the follow-up that adds a reader will also need `wireEntity`/`applyWire` plumbing, or it will work offline and silently show nothing online. Worth a note so that parity gap is not forgotten.
- nit `src/sim/content/enchants.ts`: `EnchantDef.name` is inline English (`'Enchant Weapon - Might'`). No renderer consumes it yet, so it is fine here, but when the enchant UI lands these become player-visible and will need `t()` catalog keys the way item names resolve via `entities.items.<id>.name`.
- nit `src/sim/items.ts:81` `equipItem`: the "removeItem picks the freshly-enchanted copy" comment only holds right after enchanting. Acquire a plain copy of the same item later (higher inventory index) and it would be consumed instead, since the command carries only `itemId` with no instance selector. Not exploitable today (no UI), and the future target-item picker the module already anticipates is the real fix.

Conflict is mechanical: `pending.ts`, `i18n.resolved.sha256`, `i18n.status.json`/`summary.json`. Merge the base and re-run the i18n build.

None of this is blocking.

jgyy:
Thanks for the thorough pass, especially catching the level-up regression and the buy-disenchant money-loop check. Pushed a fix commit (3eb6dfda0) addressing everything:

**Blocking**
1. Made `equipmentInstance` a required parameter on `recalcPlayerStats` (`src/sim/entity.ts`) so tsc now flags every call site. Fixed all 12 named sites (`combat/auras.ts:203`, `combat/damage.ts:757`, `combat/effect_dispatch.ts:678`/`721`, `progression/talents.ts:59`, `delves/runs.ts:635`, `social/arena.ts:850`, `social/vale_cup.ts:722`, `social/fiesta.ts:192/210/224/238`) to pass `meta.equipmentInstance` - fiesta sites kept their existing `fiestaMods`/`talentMods` argument untouched, only added the new one. The one throwaway-entity site (`entity.ts:503`, character-sheet preview derive) defaults to `{}`. Added a level-up regression test proving the enchant bonus survives.

**Should-fix**
2. Added `countEnchantableItem`/`removeEnchantableItem` on `Sim`/`SimContext` that treat a plain fungible copy OR an instanced copy with no `rolled.stats` as eligible, so crafted rare/epic instanced gear (the item this feature exists for) can now be disenchanted and enchanted; only an already-enchanted copy is excluded. `resolveDisenchant`/`resolveApplyEnchant` gate on this instead of `countFungibleItem`.
3. Added `removePreferFungible` (used by `sellItem`, `discardItem`, and `trade.ts`'s drop arm) so a plain copy is consumed before an instanced/enchanted one. Softened the module header's "stays a distinct, tradeable good" claim to flag it as not fully true yet, consistent with the PR's own known-limitations framing.
4. Pinned the rarity-scaling yield test to an exact delta instead of `toBeGreaterThanOrEqual`, pinned `statBonus.str` to the literal `5` and asserted against it directly, and added: a save/reload persistence test (equip enchanted item, serialize, reload, assert bonus survives) and a legacy-save-missing-`equipmentInstance` load test.

**Nits**
5. Removed unused `ENCHANT_IDS` and `enchantsForSlot`.
6. Softened the `equipItem` "picks up first" comment to note it only holds while the enchanted copy stays the highest-index match.

Verified with `npx tsc --noEmit`, full `npx vitest run` (green aside from the same pre-existing `i18n_resolved_equivalence` flake confirmed unrelated via `git stash`), and `npx @biomejs/biome check --write` on changed files only. Also resolved the merge conflicts against the now-current `release/v0.24.0` (both were in generated i18n artifacts, regenerated via `npm run i18n:gen` rather than hand-merged).

No new screenshots this round: the feature still has no UI surface yet (same "known limitations" scope as the original PR description), and I wasn't able to get a Chromium instance running in this environment to capture a bag/character-sheet dev-command demo either. Happy to add one in the UI follow-up issue where there's an actual window to show.

FernandoX7 (commented):
Re-reviewed at 150379f8a. All six points from the first review are addressed, and I verified each against the head code rather than the commit message:

- The blocking recalcPlayerStats gap is properly closed: the parameter is required now (src/sim/entity.ts:199), every call site passes meta.equipmentInstance (including the spirit.ts and entity_roster.ts paths the first review did not even list), the field is initialized at creation (sim.ts:1707), legacy-defaulted on load (sim.ts:1802), persisted (sim.ts:2110), and the character-sheet preview threads it too (server/character_sheet.ts:197). The new level-up test reproduces the original failure mode.
- Crafted rare and better instances are now eligible in both directions via countEnchantableItem/removeEnchantableItem gating on rolled.stats, with tests for both.
- sellItem, discardItem, and trade's drop arm route through removePreferFungible, so a plain copy is consumed before an enchanted one.
- The test pins are decisive now (exact rarity delta, literal statBonus magnitude), and the save/reload and legacy-save tests cover persistence properly.
- Both nits fixed (dead exports gone, equipItem comment softened).

I ran the enchanting, trade, entity_roster, and sim_context files plus the 12 stat-surface files locally at the head: 302 tests, all pass. CI is green, the merge state is clean, and the repaired i18n.resolved.sha256 is back to a single line.

One new issue introduced by the round-2 eligibility change, plus a residual note:

- should-fix: enchanting a crafted item erases its signer and rolled.quality. resolveApplyEnchant consumes the eligible copy and grants a fresh instance carrying only rolled.stats (src/sim/professions/enchanting.ts:167), and removeEnchantableItem returns void (src/sim/sim.ts:5213), so there is nothing to merge from. For a crafted rare (always instanced, usually the only eligible copy) that permanently kills the crafter's Battlefield Experience attribution (src/sim/professions/battlefield_xp.ts:104 requires both rolled.quality at rare or better and a matching signer) and drops it from hasSelfSignedInstance (src/sim/professions/crafting.ts:192). Suggest removeEnchantableItem return the consumed payloads the way removeItem does, and applyEnchant merge: keep signer and the rest of rolled, add stats (the deep-clone helper at src/sim/types.ts:590 already handles the nested maps).
- note, not blocking: buyback still records {itemId, count} (src/sim/items.ts:387), so deliberately selling a sole enchanted copy and buying it back returns a plain copy. Fine to fold into the documented instanced-copy follow-up alongside market/mail/trade.

Good round of fixes: making the parameter required was the right call, and the two-direction eligibility tests pin exactly the cases that were missing.

jgyy:
Thanks!

jgyy:
Confirmed - commit 20d4b658 ("fix(professions): merge crafted signer/rolled.quality into an enchanted instance") handles the round-3 should-fix: `removeEnchantableItem` now returns the consumed instance payloads, and `resolveApplyEnchant` merges the old `signer`/`rolled.quality` into the freshly-enchanted instance instead of dropping them. Battlefield Experience attribution and `hasSelfSignedInstance` should survive an enchant now. Thanks for the thorough review!

FernandoX7 (commented):
Round 3, this time a deep pass over the tier and economy design as well as the code, since the goal is for enchanting to make sense at every gear tier. First the good news: the round-2 fix landed properly. removeEnchantableItem returning the consumed payloads plus the cloneItemInstancePayload merge is exactly right, the regression test pins signer and rolled.quality to literals and fails against the pre-fix code, and removeEnchantableItem's rolled.stats guard guarantees the merge can never clobber an existing enchant. I re-verified determinism (parity goldens: rng draw counts and event hashes byte-identical, only state hashes moved from the new empty equipmentInstance field), sim purity, all 24 recalcPlayerStats call sites, persistence back-compat, and the release-merge integration (jail and the other v0.24.0 arrivals touch neither stats nor item removal). That is all clean.

**Blocking**

1. CI is red: `src/ui/i18n.resolved.sha256` is malformed again. The 07-10 base merge concatenated the branch baseline and the base baseline into two lines (`09e6c9e...` + `5bbd88a...`), and neither matches the recomputed table hash, so `tests/i18n_resolved_equivalence.test.ts` fails the PR gate. The resolved bundles at your head are fine (I ran `npm run i18n:build` on a checkout and the tree stayed byte-identical), so the fix is just `node scripts/i18n_resolved_hash.mjs --write` and committing the one-line file, the same repair 150379f8a made once already. Worth checking that file after every base merge.

**Should-fix**

2. `src/sim/professions/salvage.ts:90`: resolveSalvage still consumes via plain `ctx.removeItem`, the same highest-index-first mechanism you routed sellItem, discardItem, and trade's drop arm around with removePreferFungible (its comment at src/sim/items.ts:47 describes this exact failure). applyEnchant appends the enchanted copy at the end of inventory, so with a plain copy and an enchanted copy in bags, salvaging that itemId destroys the enchanted one first. Disenchant carefully refuses rolled.stats copies; its sibling destroy-an-item action in the same directory eats them. One-line reroute through removePreferFungible plus a mixed-inventory test.

3. Enchanting skill has no source, so the profession layer never engages. Neither resolveDisenchant nor resolveApplyEnchant calls gainCraftSkill, no recipe carries `professionId: 'enchanting'`, and the battlefield trickle is recipe-keyed, so `craftSkills.enchanting` is 0 forever for every player. That permanently locks the enchanting specialization recharge discount (professions/tools.ts, #1137) for the one craft that owns all three starter tool effects (content/professions.ts:123/133/143), keeps its materialCostMultiplier at the floor, and contradicts the free-floor principle crafting.ts:88 documents ("skill still accrues so later tiers have something to build a gate against"). Related: enchanting reads neither meta.archetype nor craftSkills, so unlike crafting (archetypeCeilingFor clamp at crafting.ts:359, zero-progress guard at :376) the anti-breadth ceiling never applies to it, and choosing the Enchanter archetype confers nothing in its own craft. A gainCraftSkill call on successful disenchant and/or applyEnchant (crafting.ts:396 is the template), routed through the same ceiling machinery, makes it a real profession.

4. The material ladder dead-ends at the top, which is the PR's stated target ("rares/epics hold their value"). arcane_shard has zero consumers anywhere in the tree: every epic/legendary disenchant yields pure vendor junk (55c a shard for destroying an epic), so no rational player ever disenchants one, and the module comment "the same three tiers applyEnchant's reagents draw from" (enchanting.ts:52) plus the PR diagram both overstate what ships. arcane_essence is barely better: two enchants at 2 each means a lifetime demand of 4, while one rare disenchant yields 4 to 5. Only the common/uncommon-to-dust loop functions as designed, and that part is genuinely well tuned (40 dust and 4 essence across the ten enchants, 43 dust with both rings, 1 to 2 disenchants per enchant). Either add essence/shard-consuming enchant tiers (which would also give point 3's skill somewhere to gate), add a downconvert, or move this into the documented follow-ups so it is a known gap rather than an implied feature.

5. The essence tier can be minted from junk instead of rare drops: recipe_ironbound_warplate_helm (content/recipes.ts:218, combo gate at skill 25 in armorcrafting + weaponcrafting) turns 4 bone_fragments + 2 linen_scrap (about 34c of trivially gathered materials, plus the 40c fee) into a rare boundstone_helm, which disenchants into 4 to 5 essence. gravewyrm_gauntlets is the same shape. So the scarce mid-tier reagent is craftable on demand from the two cheapest harvest materials, throttled only by the shared crafts-per-minute cap. With essence demand this low (point 4) it is not exploitable for much today, but it undercuts "disenchant real rares" the moment essence matters. Worth deciding now whether crafted (signed) gear should disenchant into a reduced yield or the crafted-rare loop is acceptable.

6. The one weapon enchant is a dead stat for the caster classes at every weapon tier. enchant_weapon_might (+5 str) feeds melee attack power only: spellPower is `int * SPELL_POWER_PER_INT + bonusSp` and hunter ranged power is `agi * 2` (entity.ts:400-422), so str contributes nothing for mage/warlock/priest and nothing to a hunter's ranged output, at level 1 or level 20. No int enchant exists anywhere in the ten (spread: sta 32 points, agi 15, spi 9, str 5, int 0). With exactly one enchant per slot the stat picks are the whole design, and casters get zero offensive value from the flagship slot. An int weapon option (or splitting might into per-archetype variants) fixes it; gloves are the other classic-era spot.

7. Tests (from a coverage audit): (a) removePreferFungible's prefer-plain branch is never exercised: the trade.test.ts fakes make every copy fungible, so the `ctx.removeItem` fallback is dead in every test, and that branch is the entire reason the function exists. One test: plain + enchanted copy, sell one, assert the plain went first and the instance survived. (b) All-or-nothing reagent consumption is untested for multi-reagent enchants: enchanting.ts:167-173 checks before removing, but the only insufficient_materials case (:143) is zero reagents on a single-reagent enchant. Give dust x3 and no essence for enchant_chest_stamina and assert the dust is untouched.

**Design notes on the tier question (your call, data attached, no code change assumed)**

8. The flat magnitudes are calibrated near epic budgets but apply to level-1 gear for pocket change. An enchant here is 200 to 270 percent of an uncommon item's entire primary-stat budget (uncommon chests carry sta 2 to 3; the chest enchant adds sta 10), and still about 100 percent of a level-20 epic's (morthens_cryptforged_hauberk, the biggest chest in the game, carries sta 10). A fully enchanted character gains about 65 flat stat points (both rings included) for 43 dust and 4 essence, which is 12 to 18 green disenchants. One thing that IS handled right: over-level gear grants no stats including enchant stats (entity.ts:225), so the twink vector is enchanting level-appropriate greens, not wearing enchanted epics early. Whites being disenchantable (classic restricted disenchant to uncommon and better) also makes dust more abundant than the classic model would. Classic-era enchanting handled this with a tiered ladder where floor tiers were small and only the top tier approached the item's budget; if a skill-gated second tier lands later (points 3 and 4), these values could become that top tier with a small floor tier below.

9. Two of the ten slots only exist on paper below the cap: every neck (4) and ring (6) item in the game is a level-20 epic, and both slots got spirit, the weakest stat for most classes. Fine for v1, just worth knowing when reading "one enchant per equippable slot" as coverage.

10. There is no way to change an enchant: an enchanted copy is refused by both disenchant and re-enchant, so replacing an enchant means acquiring a whole new plain copy. Classic-era behavior is overwrite-on-apply. Fine as v1 scope, but worth a line in the known-limitations list; note the refusal surfaces as `not_held`, which reads wrong to a player who does hold a copy, just an enchanted one. Related asymmetry: disenchant excludes poor-quality targets but applyEnchant accepts them, so a gray item can be enchanted.

**Nits**

11. disenchantYield's tier axis is inert for 322 of 363 disenchantable items: it reads raw `def.requiredLevel` (enchanting.ts:79), which only 41 level-20 epics set; everyone else's real requirement is derived by requiredLevelFor (item_level_req.ts: rare 12, epic 18). So a "level-12" rare and a starter rare yield identically and the comment's "scales with rarity and tier" is only half true. Swapping in requiredLevelFor(def) fixes it; note salvage.ts:68 shares the same inert axis, so either fix both or leave both consistent.

12. The module header's "strictly better yield" than salvage (enchanting.ts:16) is false at common tier: same count formula, but arcane_dust vendors at 6c against bone_fragments at 7c, and bone_fragments feeds existing recipes. Harmless, but the comment oversells it.

13. tests/professions_enchanting.test.ts:290 asserts rolled.stats with toBeDefined where `.str` toBe 5 is knowable.

14. No tests/parity scenario exercises the disenchant rng draw (single command-triggered draw, unit-pinned, same pattern as salvage, so genuinely minor; the future WS wiring issue is the natural place to add it).

One more note for whoever writes the WS wiring follow-up issue: server/character_sheet.ts:197 computes the armory sheet WITH the enchant bonus while the live inspect wire carries only base item ids, so once equipmentInstance can be populated online the two surfaces will disagree unless the wire learns the instance too. Dormant today (online cannot create an enchanted item yet), but it belongs on that issue's checklist.

The core loop is sound and the code quality is genuinely good: the instance-payload design, the SimContext seam usage, and the test suite are all the right shape. Points 1 and 2 are what stand between this and merge-ready; 3 through 6 are what stand between "two working commands" and "a profession that makes sense at every tier", and none of them looks expensive.

jgyy:
Thanks for the deep pass, especially catching the CI hash break and the salvage/tier gaps. Pushed a fix commit (fbaa09792, merged forward onto the latest base at d3936352f) addressing the blocking item and most of the should-fix/nit list:

**Blocking**
1. Repaired `src/ui/i18n.resolved.sha256` via `node scripts/i18n_resolved_hash.mjs --write`, exactly as suggested. Also caught that the branch had moved another 9 commits upstream while I worked (a mobile-autorun merge plus another base sync), so I merged that in and regenerated the hash a second time to make sure it isn't stale again.

**Should-fix**
2. `resolveSalvage` now routes through `removePreferFungible` instead of a plain `ctx.removeItem`, so salvaging a stacked itemId no longer risks eating an enchanted copy first, matching the fix already applied to `sellItem`/`discardItem`/trade's drop arm.
3. Both `resolveDisenchant` and `resolveApplyEnchant` now call `gainCraftSkill('enchanting', ...)` on success, so `craftSkills.enchanting` actually moves and the specialization recharge discount / Enchanter archetype can progress. I deliberately did *not* wire the archetype output-quality ceiling in this round: this action has no rollable output quality to clamp (unlike `craftItem`'s rarity roll), so there's nothing for that half of the machinery to gate yet, matching how `salvage.ts` also stays out of it. Flagged that scope boundary in the module header.
4. Added `enchant_weapon_intellect` and `enchant_gloves_intellect` so casters get real value from the weapon and gloves slots instead of zero (str/agi only previously).

**Nits**
5. (11) `salvageYield`/`disenchantYield` now read `requiredLevelFor(def)` instead of raw `def.requiredLevel`, so the tier-scaling term actually engages for the ~322 items that derive their level rather than pin it explicitly. Updated both exact-delta yield pins to the new (larger) expected gap now that the term fires for common/legendary/epic test fixtures too.
6. (12) Softened the "strictly better yield" framing in both the module header and the material-table comment: only true from `rare` up, since `arcane_dust` and `bone_fragments` vendor near-identically at `common`.
7. (13) Pinned the crafted-instance-preservation regression test's `rolled.stats` assertion to the literal `{ str: 5 }` instead of `toBeDefined()`.
8. (7a/7b) Added the two missing test cases you called out: `removePreferFungible`'s prefer-plain branch (mixed plain + enchanted inventory, assert the plain one is consumed and the instance survives) and all-or-nothing consumption on a multi-reagent enchant (dust present, essence entirely absent, assert nothing is touched).

**Deliberately deferred, not addressed this round**
- Points 4-5 (the essence/shard material tiers dead-ending past the common/uncommon-to-dust loop, and essence being craftable from junk via the ironbound/gravewyrm recipes): these are real economy design questions with more than one reasonable answer (add consuming enchant tiers vs. a downconvert vs. a reduced disenchant yield on crafted/signed gear), and picking one blind felt worse than flagging it honestly. Would rather take a steer on direction before touching content here.
- Points 8-10 (tier-vs-level-1-gear calibration, the two epic-only slots, no way to change/overwrite an enchant): you noted these as design notes with "no code change assumed," so left as-is; happy to fold whichever ones you want addressed into whatever tier system follow-up lands with points 4-5.

Verified with `npx tsc --noEmit` (clean), the full enchanting/salvage/architecture/i18n/parity test files (all green), and a full `npx vitest run` (green aside from the same pre-existing flaky HTTP-polling and timing timeouts under full-suite resource contention noted before, confirmed unrelated: `server/http/parity.test.ts`, `server/http/characterization.test.ts` leaderboard cases, `mail.test.ts` unread-index equivalence, and the Varkas Boneguard despawn timer, none of which touch enchanting/salvage/items code). Also ran `npx @biomejs/biome check --write` on every file I touched.

FernandoX7 (approved):

PR #1713, opened 2026-07-09
fix(ui): party-frame selection cue (targeted member highlight), corrected z-order

ryan-foo:
## What

Adds a selection cue to the party frames: the currently targeted party member's row is highlighted with an outline/glow. The cue is paint-only (no layout shift), keyed by stable player id.

## Rescoped per review (Rubsey)

The original headline (adding the local player to the party frames) is dropped. That reversed a deliberate, test-pinned decision (the local player owns `#player-frame`; showing them here double-renders the player), so it is out of scope as a "bug fix". This PR is now just the selection cue plus the z-order fix Rubsey endorsed:

- **Show-self reverted (blocking):** the self-filter in `selectPartyFrameMembers` and the skip in `partyFrameSignature` are restored; the tests are back to the original `skips the local player` assertions. Only the `selected` flag is added.
- **z-order corrected (should-fix):** the party container `z-index` bump `5 -> 7` is reverted, so party rows no longer paint over the target's debuff strip and cast bar (actionable info).
- **Combat-glow clobber fixed (nit):** a row that is both in-combat and selected now layers the red combat glow AND the gold selection glow (`.party-frame.combat.selected`), instead of the selection replacing the combat glow.
- The mobile third-column and right-click-self nits are moot once the local player is not shown.

## Base

Rebased onto `release/v0.26.0`.

## Testing

`tests/party_frames.test.ts` keeps the restored self-skip assertions and adds a selection-cue case (targeting a member flags only its row); `tests/party_frames_painter.test.ts` covers the class toggle. CSS corpus/validity guards, `tsc`, and the suite are green.

The selection cue is cosmetic (outline/box-shadow, no `data-fx-level` gating), so it does not touch the party-HP graphics-fairness rule.

### DIFF
--- a/src/styles/hud.css
+++ b/src/styles/hud.css
@@ -4082,7 +4082,7 @@
     display: flex;
     flex-direction: column;
     gap: 6px;
-    z-index: 5;
+    z-index: 7;
   }
   /* The painter nests the member rows one level down in a .party-rows wrapper (so the
      mobile collapse chip can sit alone on its own line). On desktop the wrapper is
@@ -4235,6 +4235,12 @@
   .party-frame:hover {
     outline: 1px solid var(--gold-dim);
     opacity: 1;
+  }
+  .party-frame.selected {
+    position: relative;
+    z-index: 1;
+    outline: 1px solid var(--gold);
+    box-shadow: 0 0 calc(8px * var(--fx-shadow, 1)) var(--gold-dim);
   }
   /* Reset opacity on a keyboard-focused dimmed (dead 0.7 / out-of-range 0.45) row so the
      global [tabindex="0"]:focus-visible outline (shell.css) is drawn at full strength, not
--- a/src/styles/hud.css
+++ b/src/styles/hud.css
@@ -4082,7 +4082,7 @@
     display: flex;
     flex-direction: column;
     gap: 6px;
-    z-index: 7;
+    z-index: 5;
   }
   /* The painter nests the member rows one level down in a .party-rows wrapper (so the
      mobile collapse chip can sit alone on its own line). On desktop the wrapper is
@@ -4241,6 +4241,14 @@
     z-index: 1;
     outline: 1px solid var(--gold);
     box-shadow: 0 0 calc(8px * var(--fx-shadow, 1)) var(--gold-dim);
+  }
+  /* A row that is both in-combat and selected keeps the red combat glow AND the gold
+     selection glow: the two box-shadows are layered so neither clobbers the other
+     (equal specificity, .selected wins on box-shadow otherwise). */
+  .party-frame.combat.selected {
+    box-shadow:
+      0 0 calc(7px * var(--fx-shadow, 1)) #e74c3c99,
+      0 0 calc(8px * var(--fx-shadow, 1)) var(--gold-dim);
   }
   /* Reset opacity on a keyboard-focused dimmed (dead 0.7 / out-of-range 0.45) row so the
      global [tabindex="0"]:focus-visible outline (shell.css) is drawn at full strength, not
--- a/src/styles/hud.mobile.css
+++ b/src/styles/hud.mobile.css
@@ -1663,7 +1663,7 @@
     flex-direction: column;
     align-items: flex-start;
     width: auto;
-    z-index: 7;
+    z-index: 5;
   }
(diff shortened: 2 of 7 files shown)


jgyy (commented):
Good fix, and appropriately small: the bug was `selectPartyFrameMembers`/`partyFrameSignature` unconditionally dropping the local player's own row, nothing to do with graphics tier, and the fix stays in the existing pure `party_frames.ts` module plus a one-line painter toggle. The selection highlight is keyed by stable player id (`m.pid === selectedId`), not array index, which is exactly right given the bug this PR is fixing is about instability. `tests/party_frames.test.ts` was updated to actually document the behavior change (renamed assertions, not just re-recorded snapshots), plus a new painter test for the CSS class toggle.

Confirmed the selection cue is purely cosmetic (outline/box-shadow, no layout shift) and not gated behind any `data-fx-level` tier knob, so this doesn't touch the fairness rule around party HP visibility.

One minor thing: `selectPartyFrameMembers`'s `playerId` param is renamed to `_playerId` and now unused, while `partyFrameSignature` still needs it, the two functions' signatures are now asymmetric despite always being called together with the same args in `hud.ts`. Small foot-gun for a future caller, not worth a breaking signature change this late, just flagging.

Rubsey (commented):
The mechanical work here is careful (write-elision preserved, the selection cue is layout-neutral, all gates green), but the headline change needs to be pulled or explicitly signed off before this can land.

- blocking: src/ui/party_frames.ts:20. Showing the local player in the party frames reverses a deliberate, documented, test-pinned decision, and the stated premise is not real. On the base branch, selection has no effect on the party-frame model at all (the only filter was self), and there is no prior selection cue anywhere ("selected" does not appear in the base party files or CSS), so "selecting a member could drop or overlap other member frames" matches no base code path. The base comment says "the frames never show the local player", the game already renders you in #player-frame, and this change makes the player appear twice. If showing self is wanted, that is a design decision for the maintainer, not a bug fix; as written the PR rewrites the test pins to the opposite assertion.
- should-fix: src/styles/hud.css:3788. The z-index 5 to 7 bump inverts the overlap rather than removing it: party rows now paint over the target's debuff strip (#tf-debuffs, whose hover tooltips they also capture) and, with a casting target or targetFrameScale above 1, over the target cast bar, both of which are actionable info. A measured below-strip offset would fix the actual geometry.
- should-fix: src/styles/hud.mobile.css:1689. The mobile party grid is two explicit rows tuned so a full party fits two columns (issue 1577); with self included, a party of five flows a third column mid-screen, and the width comments referencing that tuning are now stale.
- nit: src/styles/hud.css:3944. On a row that is both in combat and selected, .selected replaces the red combat glow (equal specificity, later rule wins on box-shadow).
- nit: src/ui/hud.ts:13219. Right-clicking your own new row opens the generic player menu offering Trade / Duel / Add Friend on yourself; #player-frame uses openSelfContextMenu for this.

The selection cue and a corrected z-order fix would stand fine on their own; happy to see this rescoped to that.

jgyy (commented):
Clean fix. The `selected` flag threads through `selectPartyFrameMembers`/`partyFrameSignature` with a properly defaulted `selectedId` param so existing call sites are unaffected, and the CSS uses the existing custom properties and a small local `z-index` consistent with how the rest of hud.css handles this kind of row-stacking, not a magic-number hack. The combined `.party-frame.combat.selected` rule so the combat glow and the selection glow don't clobber each other is a nice touch.

No play.html mirroring needed since party frame markup is generated entirely in TS, nothing in either HTML entry to keep in sync. `tests/party_frames.test.ts` and `tests/party_frames_painter.test.ts` both got real assertions (signature flip on selection change, `toggleClass('selected', true)`), not just smoke tests.

No blocking findings. Minor note: the selected cue is color/outline only with no ARIA or textual signal for screen readers, but that matches how `combat`/`dead`/`oor` are already handled in this same file, so it's consistent with current practice rather than a new gap.

Thanks for sorting out the z-order here, the highlight not showing was a small thing but noticeable.

ryan-foo:
Closing as part of the current PR-queue batching. The party/raid frame surface is being reworked in #1864 (configurable movable party and raid frames); the targeted-member selection cue belongs in that rework rather than as a standalone cue on the old frames. Can reopen or fold the cue into #1864 if wanted.

PR #1715, opened 2026-07-09
feat(raid): kicked raiders are teleported out of the raid within a minute

ryan-foo:
## What

Kicking a raider from a raid party now teleports them out of the raid instance after a 60 second grace, using the exact safe door coordinates a normal leave uses. The timer is pure sim clock on the existing once-per-second instance-update gate (no rng, no `Date.now`), and the relocation is a shared extraction of `leaveDungeon`'s tail (`relocateOutsideDungeon`). The eject state is session-only (excluded from persistence).

## Review fixes (Rubsey)

- **Re-invite cancels the eject (should-fix):** `updateRaidKickEjects` now checks party membership. A misclick kick that is re-invited during the grace no longer teleports the raider out at expiry: as soon as the player is back in the party that owns the raid instance they are standing in (their instance key matches its `partyKey`), the pending eject is cleared. This also removes the sealed-royal-door re-entry trap.
- **Test coverage (nit):** added the re-invite-cancels-eject case and the early-clear case (walked out of the raid during the grace), which the review flagged as unexercised. The happy-path eject test is unchanged.
- **Dead-player relocate (nit):** the eject deliberately relocates an unreleased corpse too (its `corpsePos` is captured at release time, so this is safe and, per the review, arguably beneficial); this is a known, intentional difference from `leaveDungeon`'s dead-player guard.
- **Scope note:** voluntary leave and party disband still leave members inside (they walk out via the portal), matching the title.

## Base / conflicts

Rebased onto `release/v0.26.0` and resolved the mechanical i18n regen churn (`i18n.status.json` is now gitignored on the base; the summary was regenerated via the i18n build).

## Testing

`tests/dungeons.test.ts`: the eject-after-grace, re-invite-cancels, and early-clear cases all pass. Parity, architecture, and the S3 i18n guard are green; `tsc` clean.


jgyy (commented):
Solid, and the security/authority checks all hold up: kick permission is still gated on `party.leader` (unchanged pre-existing check), the eject timer is tick-based (rides the existing once-a-second instance update gate, no `Date.now`), and the teleport destination reuses the exact same safe door coordinates as a normal dungeon leave, so there's no way to land a kicked player somewhere unsafe. The client has no way to cancel the eject since it's purely server-tick driven. Nice reuse too, the teleport-out body was extracted into `relocateOutsideDungeon` and shared with the existing leave-dungeon path instead of duplicating it.

The new `raidKickEjectRemaining` field is session-only by design and correctly excluded from the persistence path, so no wire/DB parity work was needed and none was added, that's the right call, not an oversight. The new test kicks a raider mid-raid, checks the immediate state, ticks 60 sim-seconds, and asserts both the position change and the freed instance slot, that's decisive coverage of the full path.

One thing worth a line in the PR description rather than a code change: if a kicked player leaves the raid area voluntarily before the 60s timer fires, the countdown is zeroed (correct), and re-entering the same raid instance within that window does not re-arm the eject. That looks intentional but is worth stating explicitly since it's a real design choice.

Rubsey (commented):
Clean implementation: the timer is pure sim clock inside the existing once-per-second updateInstances gate, draws no rng (parity green with no re-mints), the relocation is a faithful extraction of leaveDungeon's tail, the state is deliberately session-only, and the S3 i18n guard is green.

- should-fix: src/sim/instances/dungeons.ts:265. Re-inviting the kicked player during the 60s grace does not cancel the eject: updateRaidKickEjects never checks party membership, so a misclick kick that is immediately re-invited still teleports the raider out at expiry, and if Nythraxis is engaged the sealed royal door blocks re-entry until wipe or kill. Clearing the timer on partyAccept, or skipping the eject when the player is back in a raid whose instance key matches, closes it.
- nit: tests/dungeons.test.ts:822 covers only the happy path; the early-clear branches (walked out during grace, logout) and the re-invite case are unexercised.
- nit: src/sim/instances/dungeons.ts:280. The eject bypasses leaveDungeon's dead-player guard, so an unreleased corpse gets relocated with leaveText. Arguably beneficial (corpsePos is captured at release time), but it is an unstated difference from the normal leave path.
- nit: voluntary leave and disband still leave members inside (they can walk out via the portal); consistent with the title, just noting the scope line.

The i18n.status.json conflicts with the base are mechanical regen churn: merge the base and re-run the i18n build.

jgyy (commented):
Good, this is properly tick-based rather than wall-clock: the eject countdown lives on `PlayerMeta` and decrements inside the existing once-per-second instance-update sweep, no `setTimeout`/`Date.now` anywhere in the path, so it's RL-env safe. The re-validation on every sweep handles the cases that matter: leaving the instance on your own cancels it, getting re-invited back into the owning party cancels it, and logout just drops the pid out of the iterated map with nothing persisted, so a relogin can't resurrect a stale eject.

`tests/dungeons.test.ts` covers fire-after-delay, rejoin-cancels, and self-leave-cancels with decisive assertions on the exact teleport destination and log text.

No blocking findings. One low-severity nit: there's no explicit test for logging out mid-window, the code looks safe by inspection (the meta just falls out of the map), but an explicit test would remove the need to take that on faith.

Thanks for handling the raid-kick edge cases this carefully, the rejoin-cancels case in particular is easy to get wrong.

ryan-foo:
Closing in this batching pass to keep the queue focused. The kicked-raider teleport can be reworked and reopened later once the raid-frame/raid-management surface (#1864) settles, so the eject behavior lands against the current design rather than the pre-rework code.

Issue #1021, opened 2026-06-29
Add a party/raid Ready Check (leader-triggered readiness confirmation before a pull)


jgyy:
## Summary

Add a party/raid **Ready Check**: the group leader presses one button (or types `/readycheck`), every member gets a quick "Are you ready?" prompt with a short countdown, and the party frames show a green check or red X per member so the leader knows the group is set before pulling a boss, entering a dungeon, or starting an arena/Fiesta match.

This is a classic group-coordination tool. The game already has all the prerequisites (parties, raids, raid markers, dungeon entry), but there is no way to confirm everyone is actually at the keyboard and topped off before a pull.

## Why

Right now coordination happens entirely in chat ("ready?" "1 sec drinking" "go"). For a 5-player party or a 10-player raid that is slow and easy to misread. A Ready Check is the standard, low-friction answer and fits the classic feel the project is going for.

## Current state

Parties and raids already work (the `PartyMachine`, session A1, owns invites/leader/raid state, and party frames render top-left). There is simply no readiness concept for the group: the only "ready" in the sim today is quest-readiness (`checkQuestReady`). Below, a 3-player party formed in the offline client. Note the party frames (Mirelle, Toren) and "Leave Party": this is exactly where a per-member ready indicator would surface.

![Current party frames, no readiness indicator](https://raw.githubusercontent.com/jgyy/world-of-claudecraft/pr-assets-ready-check/ready_check_party_frames.png)

## Proposed behavior

- **Leader-only trigger.** A "Ready Check" action (party menu button and a `/readycheck` chat command). Members cannot start one; add a short per-party cooldown so it cannot be spammed.
- **Member prompt.** Each member sees a "Ready?" prompt with Ready / Not Ready and a countdown (about 30s). No response by expiry counts as Not Ready.
- **Live readout.** The leader (and ideally everyone) sees each member's status resolve on the party/raid frames: a green check for ready, a red X / hourglass for not-ready-yet.
- **Result.** When all respond or the timer expires, a one-line system message summarizes ("Everyone is ready" or lists who is not).

## Suggested shape (follows the repo's invariants)

Server-authoritative, like every other group action: the client sends intent, the server runs the one shared `Sim` and broadcasts results, so offline and online behave identically.

- Add the commands/state to the existing party machine (`src/sim/social/party.ts`) rather than a new system, keyed off the existing party id.
[...]

PR #1716, opened 2026-07-09
feat(social): raid/party ready check (/ready)

ryan-foo:
## What this is

A raid/party ready check, the classic `/ready`.

- The party or raid **leader** runs `/ready` (alias `/readycheck`).
- Every **other** member's client plays a distinct rising chime and shows a **yes/no prompt** (Ready / Not Ready). The leader is auto-marked ready and gets no prompt.
- Members answer. Anyone still pending **30 seconds** later is counted as "no response".
- The outcome is announced to the party as a counts-only summary (`Ready check: X ready, Y not ready, Z no response.`), keeping individual answers private, matching the classic client.

Mechanics researched against the WoW ready-check behavior (leader-initiated, yes/no popup, 30s timeout marks non-responders, tally to the party).

## How it is built (follows the party-invite template end to end)

- **New sim module** `src/sim/social/ready_check.ts` behind the `SimContext` seam. State is `ctx.readyChecks` (party id to check) on `Sim`, swept in the fixed end-of-tick block by `updateReadyChecks`. Uses the sim clock (`ctx.time`), draws no rng, so replays and the parity gate stay byte-stable.
- **`/ready`** routes through the chat command router to `ctx.readyCheckStart` (leader-gated, must be in a party, one at a time).
- The **yes/no answer** is a new `IWorldParty.readyCheckRespond(ready)` + the `readyrespond` wire command (server dispatch + `ClientWorld` sender), so it works online and offline.
- **Prompt event** `readyCheckStart` reuses the existing `showPrompt` component (extended with an optional decline label so the buttons read Ready / Not Ready). New chime `audio.readyCheck()`.
- **i18n**: the popup + buttons + summary are `hudChrome.readyCheck.*` (English + five non-Latin fills, M16); the two sim-side error strings and the summary line are registered in `sim_i18n` (the leader-gate error reuses the existing "You are not the party leader.").

## Testing

- New `tests/ready_check.test.ts`: leader start sends the prompt to others but not the initiator, non-leader / no-party are rejected, finalize-on-all-answered, the 30-second timeout finalizes and clears, no double-check.
- Green locally: `ready_check`, `world_api_parity` + `command_schema` (registries updated for the new IWorld method + wire token), `snapshots`, `parity/harness` (draw order unchanged), `chat`, `social`, `architecture`, `localization_fixes` (S3), `i18n_completeness`; `tsc` clean.


jgyy (commented):
Nice implementation of the feature from #1021. Despite the 50-file / 1116-line stat, the real change is small and tight (`src/sim/social/ready_check.ts` plus a handful of small touches elsewhere), the rest is mechanical i18n regen and parity-count pin bumps. Authority checks are correct: starting a check is gated on `party.leader`, and responding is scoped to `ctx.resolve(pid)` off the session-bound id, so a player can't answer on behalf of anyone else. The timer rides the sim clock (`ctx.time`), not wall-clock. State is deliberately kept server-internal rather than pushed onto the wire, delivered via the existing chat/notice + SimEvent prompt mechanism, the same pattern `partyInvite`/`guildInvite` already use, and both `Sim` and `ClientWorld` implement `IWorldParty.readyCheckRespond` consistently with the parity test counts updated to match.

The UI reuses the existing `showPrompt` mechanism rather than growing `hud.ts` with a new banner (net +12/-1 there), which is the right call for a one-shot yes/no prompt that's structurally identical to existing ones. New strings are correctly split: `hud_chrome` for the UI-rendered English `t()` text, and a `sim_i18n.ts` matcher for the server-emitted result line.

Test coverage is good for a first pass: leader-gating, self-only response, early finalize on all-responded, timeout finalize, and reject-on-concurrent-check, with assertions that check actual map contents rather than just "doesn't throw."

Small nits, nothing blocking: no visible countdown while the prompt is open, just yes/no (probably fine, ready checks aren't the kind of timer CLAUDE.md's fairness rule is worried about, but worth confirming that's intentional), and the `/readycheck` alias isn't mentioned anywhere player-facing.

No true conflicts. Good to merge.

Rubsey (commented):
Good feature, built the right way: validation is entirely sim-side (there is no start wire command to spoof, readyrespond is membership-checked, respond-after-finalize is a no-op), the summary is counts-only on every host, state is a live ctx view swept in the fixed end-of-tick block with no rng (parity green, no re-mints), IWorld is implemented in both worlds with every registry pinned, and the showPrompt decline-label extension leaves existing callers byte-identical.

- should-fix: the changed-files Biome gate fails at the head on 7 PR-introduced errors: import order in src/sim/sim.ts:405 and tests/ready_check.test.ts:4, plus format rewraps on the added hudChrome.readyCheck.result lines in the five i18n.locales files. Mechanical: npx @biomejs/biome check --write on those files.
- should-fix: src/ui/hud.ts:8975. showPrompt auto-removes after 28s wall clock and calls onDecline, which here sends readyCheckRespond(false), so a connected player who simply ignores the prompt is recorded "not ready" at 28s and the sim's 30s "no response" bucket is reachable only for disconnected members, against the PR's stated semantics. tradeRequest in the same switch passes a no-op onDecline and lets the sim timeout own the outcome; that pattern fits here.
- nit: src/sim/types.ts:2021. The ReadyCheck comment mentions an 'afk' state that does not exist; stragglers stay 'pending' and count as no response.
- nit: src/sim/social/ready_check.ts:66. No cleanup on leave/disband: a mid-check leaver stays pending forever, so all-answered early finalize can never fire and ex-members still get the summary. Safe (party ids are never reused), just slightly off.
- nit: answers can be changed any number of times before finalize (Map.set overwrites); the classic client locks the first answer.
- nit: /help does not list /ready, so the command is undiscoverable in game.

The generated i18n conflicts with the base are the standard regen churn: merge the base and re-run the i18n build.

FernandoX7 (commented):
Nicely done. This follows the party-invite template closely and lands in the right places: a self-contained `social/ready_check.ts` behind the SimContext seam, `readyChecks` exposed exactly like `partyInvites`, no rng and a `ctx.time` timeout (same convention as the party-invite expiry) so replays and parity stay byte-stable, and the counts-only summary keeps individual answers private. I traced the online path end to end and it holds up: `readyCheckStart` is a personal (pid-set) event that `routeEvents` delivers to the addressed session regardless of interest scope (so a raid member across the zone still gets prompted), `online.ts` drains it generically into the event queue, and the `readyrespond` boolean rides through the permissive `ClientMessage` envelope. `party.members` is the full roster across `raidGroups`, so `/ready` reaches the whole raid, not just the base group. English copy is `t()` keys in the right catalog (`hudChrome.readyCheck.*`) and the sim-side error/summary strings are matched in `sim_i18n` in the same change.

Two things worth fixing:

**should-fix: an in-flight ready check outlives its party.** The disband branch of `removeFromParty` clears `parties` and `dropPartyMarkers(party.id)` but not `ctx.readyChecks` (`src/sim/social/party.ts:409-415`), and `removePlayer` cleans `partyInvites`/`tradeInvites`/`duelInvites`/`chatTokens`/`channelSubs` but not `readyChecks` (`src/sim/sim.ts:2047-2074`). The blast radius is bounded (the end-of-tick sweep self-clears the orphan within 30s, and party ids are monotonic so there is no id-reuse aliasing), but two visible symptoms remain: a disband mid-check still fires the counts-only summary to every ex-member 30 seconds later about a party that no longer exists, and when a single member just leaves, their pid stays `pending` in `responses` so the remaining members never early-finalize (`ready_check.ts:59`) and eat the full timeout. Fix mirrors the adjacent line: `ctx.readyChecks.delete(party.id)` in the disband branch, and drop the leaver from any active check on removal.

**should-fix (tests): the tally the feature exists to produce is unasserted.** The five cases all assert `readyChecks.size` / `endsAt`, but none assert the summary itself, so `finalizeReadyCheck`'s counting (`ready` / `notReady` / `noResponse`, `ready_check.ts:70-88`) and the not-ready answer path are untested. Swapping `notReady` and `noResponse`, or dropping the no-response branch, leaves all five green. Add a case with a leader plus three members where one answers ready, one not-ready, and one times out, and assert the emitted `Ready check: 2 ready, 1 not ready, 1 no response.` (and exercise at least one `readyCheckRespond(false)`).

Two nits:

- `readyrespond` is added to `COMMAND_NAMES` but not tagged in `COMMAND_FACETS`, where every sibling party command (`pinvite` through `clearMarker`) is (`src/world_api.ts:463-475`). It is discovery metadata that nothing enforces, so it breaks nothing, but `readyrespond: 'IWorldParty'` there keeps the party family symmetric.
- The parity-test header comments still read 204 / 54 / 150 (`tests/world_api_parity.test.ts:12` and `:79`) after this change bumped the live assertions to 205 / 54 / 151. Comment-only staleness the bump introduced.

**Merge:** the branch is behind base and shows conflicts. The real ones are the usual i18n regen churn (`pending.ts`, `i18n.resolved.sha256`, `i18n.status.json`, `i18n.status.summary.json`) plus `tests/command_schema.test.ts` and `tests/world_api_parity.test.ts`. Those last two look alarming but resolve trivially: base's warrior overhaul touched them and was reverted, so it is back at 131/140 and 204/150 and your +1 bumps apply straight on top. Merge base and re-run the i18n build to regenerate the four generated files.

FernandoX7 (approved):

PR #1718, opened 2026-07-09
fix(ui): show floating combat text damage numbers on the low graphics tier

ryan-foo:
## Summary

On the **low** graphics preset the floating combat text (FCT) painter dropped every non-crit damage floater via `fctDropNonCrit`. Because FCT only ever spawns for player-involved events, this hid a low-tier player's core combat feedback: the damage they deal to their target. Most hits are not crits, so a low-preset player saw almost no damage numbers over enemies.

This removes the drop-non-crit shed entirely. The low tier still sheds FCT cost through the bounded pool, a tighter live cap (`fctMaxConcurrent`) and a shorter TTL (`fctTtlScale`), so a burst still clears fast. Every floater, including non-crit damage numbers, now spawns on every tier. The only remaining crit knob is the CSS crit-emphasis gate (`[data-fx-level="low"] .fct.crit`), which keeps the number and drops only the scale/pop, so no actionable information is ever hidden.

This keeps the gameplay-neutral-graphics invariant intact (FCT volume/lifetime is still cosmetically tiered via the pool cap and TTL) while restoring the numbers a low-tier player reacts to.

## Related issues

N/A

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/fct_painter.test.ts tests/ui_tier_knobs.test.ts tests/fct_core.test.ts tests/ui_effects_profile.test.ts` (green)
  - `npx vitest run tests/client_shell.test.ts tests/hud_perf_budget.test.ts tests/ui_effects_wiring.test.ts` (green)
  - `npx @biomejs/biome check` on the changed files (clean)
- Manual steps:
  - The behavior is deterministic and covered by the updated FCT tiering tests: low now spawns every non-crit damage number across all `DAMAGE_FCT_KINDS`, and still caps the live count / shortens the TTL.

## Screenshots / recordings

On the LOW graphics preset (data-fx-level="low"), the non-crit damage you deal to an enemy now floats (before the fix only crits showed). Captured with `scripts/fct_lowtier_shot.mjs` (real `hud.handleEvents` -> `FctPainter` spawn path):

Low-tier floating damage numbers

---

### Quality

- [x] **Tests pass.** Updated the FCT tiering tests for the new behavior.
- [x] **Typecheck passes.** Changed files are clean (`tsc` shows only a pre-existing `prom-client` dependency gap unrelated to this diff).
- [ ] ~Builds succeed~ (no build-affecting change)

### Invariants

- [x] Graphics/perf settings stay gameplay-neutral: FCT volume/lifetime is still tiered via the pool cap + TTL; the fix stops the tier from hiding actionable damage numbers, which the fairness doc now reflects.


jgyy (commented):
This is the right fix and it's landed correctly against the fairness rule: the old `fctDropNonCrit` was suppressing a player's own non-crit damage numbers entirely on the low tier, that's actionable combat feedback, not cosmetic volume/lifetime, and CLAUDE.md is explicit that a tier knob may never hide that. Removing it and leaving only `fctMaxConcurrent`/`fctTtlScale` (pool cap + TTL) as the low-tier cost knobs is exactly the cosmetic-only shedding the rule allows, and the tier read still goes through the static `data-fx-level` preset, not the governor. Tests across `fct_painter`, `fct_core`, and `ui_tier_knobs` were all flipped to assert every floater spawns on every tier, with only cap/TTL differing, that's the right regression pin.

One thing to sort before merge: this PR targets `release/v0.23.0`, which is 22 commits behind the current `release/v0.24.0` that the rest of this batch targets. The good news is `git merge-tree` against `release/v0.24.0` comes back clean with no conflicts, so this should just be a `gh pr edit --base release/v0.24.0`, no rebase work needed.

Nothing blocking on the code itself. Please retarget to release/v0.24.0.

PR #1720, opened 2026-07-09
feat(input): bindable pet bar keys (Ctrl+1..5 by default)

ryan-foo:
## What this is

Bindable pet bar hotkeys. Five rebindable pet-command actions in a new **Pet** key-binding category:

| Action | Default | Command |
|---|---|---|
| Pet: Attack | Ctrl+1 | attack the current target |
| Pet: Stop | Ctrl+2 | passive stance (stop attacking) |
| Pet: Taunt | Ctrl+3 | Growl / taunt |
| Pet: Defensive | Ctrl+4 | defensive stance |
| Pet: Aggressive | Ctrl+5 | aggressive stance |

The defaults sit on **Ctrl + number** so the action bar 1..5 stay free, and every one is **rebindable** in the Key Bindings menu like any other action (the keybind system already supports modifier chords).

## How it is built

No new sim, wire, or IWorld surface: the actions drive the **existing** IWorld pet commands (`petAttack` / `petTaunt` / `setPetMode`), and the pet bar UI already exists. The change is the input layer only:

- `keybinds.ts`: a new `Pet` category with the five `BIND_ACTIONS` (Ctrl+Digit1..5, `edge`).
- `input.ts`: an `onPet(action)` callback + `dispatchEdge` cases.
- `main.ts`: wires `onPet` to the IWorld pet commands (stop -> `setPetMode('passive')`).
- Labels + the Pet category header localize via `hudChrome.keybinds.*` (English + five non-Latin fills, M16), and show up automatically in the rebind menu.

## Testing

- New `tests/pet_bar_keybinds.test.ts`: the five actions register in the Pet category with the Ctrl+1..5 defaults, the chords resolve through `Keybinds.actionForCode`, a bare digit still belongs to the action bar, and rebinding a pet action frees its old default.
- Green locally: `pet_bar_keybinds`, `keybinds`, `input`, `options_window`, `architecture`, `i18n_completeness`, `localization_fixes`; `tsc` clean. No wire/IWorld changes, so the snapshot/command/parity registries are untouched.


jgyy (commented):
Clean, well-scoped PR. It reuses the existing `Keybinds` seam for storage/rebind/persistence rather than inventing bespoke handling, and the default `Ctrl+1..5` is verified non-colliding with the plain `1..5` action-bar slots, with a negative test pinning that a bare digit does not resolve to a pet action. The `main.ts` addition is a thin dispatcher to existing `IWorld` pet commands, not new logic folded into the monolith, and the options-window changes reuse the existing keybind category/row rendering, so the rebind UI's keyboard operability and a11y come for free rather than needing a fresh audit. New strings are correctly in `hud_chrome`. Of the 39 changed files, 34 are mechanical i18n regen, only 5 carry real logic.

One nit worth a quick confirmation rather than a code change: no mobile-touch equivalent for pet actions was found (no `onPet`/`petAttack` references outside the keybind/input/main files), presumably fine since mobile has no keyboard and pet actions are reachable through the existing pet UI, but worth the author confirming that explicitly.

Rubsey (commented):
The plumbing is exactly right: it reuses the existing chord machinery, adds zero sim/wire/IWorld surface, routes through IWorld so offline and online behave identically, full-chord matching means Ctrl+1 does not also fire action slot 1 (pinned by the new test), and the options window renders the new category through the existing eviction flow.

- should-fix: src/game/keybinds.ts:233. The shipped defaults are the problem: Ctrl+1 through Ctrl+5 are browser tab-switching chords on Windows and Linux, and these are the repo's first modifier-chord defaults with no reserved-chord handling anywhere (isReservedCode covers only Escape). macOS is fine (tab switching is Cmd there).
- should-fix: src/game/input.ts:724. Matched edge chords dispatch without preventDefault (only Tab, Space, attack-move, and the emote wheel prevent), so even in browsers where Ctrl+digit is interceptable (Firefox), the pet command fires and the tab still switches; in Chrome/Edge on Windows/Linux the chord is reserved outright. Together with the previous point, a default-bind hunter or warlock on the majority desktop platform switches tabs mid-fight. preventDefault on matched edge chords plus browser-safe defaults (or shipping them unbound) would make the feature usable off macOS.
- nit: src/main.ts:1103. 'Pet: Stop' is a persistent stance change to passive, not a transient stop; deliberate per the comment, worth being aware of.
- nit: the pet bar buttons do not render the new keycaps (action bar slots and micro-menu buttons do), so the binds are only discoverable in the options window.

The generated i18n conflicts with the base are the standard regen churn: merge the base and re-run the i18n build.

FernandoX7 (commented):
Clean, well-scoped change. It adds no sim, wire, or IWorld surface and drives the existing `petAttack` / `petTaunt` / `setPetMode` commands, so the offline and online worlds stay identical for free. The `Ctrl + number` default keeping the action bar 1..5 free is the right call, `stop -> setPetMode('passive')` genuinely calls the pet off (passive clears `aggroTargetId` / `inCombat` / `autoAttack` in `pet_commands.ts:704`), and the chord resolution is correct: `Ctrl+Digit1` routes to the pet action while bare `Digit1` still hits the action bar. The `options_window.ts` map also picks up `discord` / `valecup`, which were falling back to raw English before, a nice incidental fix.

One heads-up on the merge state: GitHub shows CONFLICTING, but `git merge-tree` against the current `release/v0.24.0` tip is clean, so that flag looks stale. The only prior conflicts were the standard i18n regen trio, already resolved.

A few things worth a look:

**should-fix** `src/game/keybinds.ts:233`-`249` (defaults `Ctrl+Digit1..5`): on Windows/Linux browsers `Ctrl+1..8` is the tab-switch accelerator, and the edge-dispatch path never calls `preventDefault` (`input.ts:741` runs `dispatchEdge` with no preventDefault, unlike the Tab/Space/held branches). The keydown still reaches the page so the pet command fires, but with more than one tab open the browser also jumps away from the game tab, and these accelerators are not reliably cancelable even if you did preventDefault. It is a non-issue where you likely tested: macOS browsers use Cmd (not Ctrl) for tab switching, and the Electron desktop app has no tabs. But the web client on Windows/Linux is a large share of players, and this is the shipped default for the headline feature. Your call on whether to pick default chords the browser does not reserve.

**nit** No test drives the actual new input code. `pet_bar_keybinds.test.ts` covers `Keybinds` resolution only; the five `dispatchEdge` cases (`input.ts:787`-`800`) and the `main.ts:1117` wiring go unexercised. `input.test.ts` already has a keydown harness, so a `{ code: 'Digit1', ctrlKey: true, preventDefault }` press asserting `cb.onPet` was called with `'attack'` would cover the branch you added.

**nit** The default binds `Ctrl+1..5` for all nine classes, so the seven non-pet classes get an "Only pet classes can command pets." error toast (`pet_commands.ts:535`) when they press these. Minor, but the pet keys are meaningless for them.

**nit** `options_window.ts:130` maps the Pet category to `hudChrome.keybinds.categoryPet` while the other four categories use `hud.keybinds.categories.*`. Both are valid English keys; just a small namespace inconsistency.

FernandoX7 (approved):

Issue #1670, opened 2026-07-08
All chat tab mixes game event lines with player chat and offers no filtering


MasterZensei:
## Problem

The chat box has two built-in views (All and Combat Log) plus optional per-channel tabs. Damage lines already go to the separate Combat Log pane, but everything else the game emits (XP gains, loot lines, quest progress, level ups, zone welcomes, crafting and mailbox results, death and respawn notices, arena and minigame updates) is appended to the chat pane as a single flat "system" category that always shows in the All view. During normal play the All tab is dominated by this game feed and actual player conversation (say, party, guild, world) scrolls away.

Filtering today is all or nothing: a channel tab shows exactly one channel, and the All view is not configurable at all. There is no way to keep one conversation-first combined view that shows player chat while muting or thinning the game feed.

## Expected behavior

- The player can choose per category what the main chat view shows, without giving up a combined view.
- Game-emitted lines are split into meaningful categories rather than one "system" bucket (at minimum: loot, XP and progression, quest progress, minigame and event updates, general info).
- Filter choices persist across sessions, like the existing chat tabs do.
- Defaults keep the current behavior (everything visible), so nothing changes for players who never touch the filters.

## Scope

- The chat tab model in `src/ui/chat_channels.ts` (pure, unit-tested) and the tab DOM plus filtering in `src/ui/hud.ts` (`applyChatFilter`, `appendLog`).
- The `chan` tagging on appended lines: `appendLog` currently tags every non-player line `chan = "system"`, so finer categories are a prerequisite for any filter UI.
- Persistence alongside the existing chat tab localStorage keys.

## Acceptance criteria

- A player can independently hide loot, XP, quest progress, and minigame or event lines from the All view, while player chat channels keep working unchanged.
- Filter state survives a reload.
- Filtered-out lines are not announced by the chat live region, matching how channel-filtered lines behave today.
- Combat Log behavior is unchanged: damage lines stay in their own pane and never enter the chat pane.
- Desktop and mobile chat both get the controls; every new control meets the mobile tap-target floor and focus-visible standards, and all new labels are `t()` keys.

## Notes

Suggested direction, adapted to the existing tab model rather than replacing it:

[...]

PR #1722, opened 2026-07-09
feat(combat): make DoTs and HoTs fully dynamic (live power, haste, crit)

ryan-foo:
## Summary

Makes every DoT and HoT fully dynamic, the way classic-era MMOs work: a damage- or
heal-over-time effect recomputes each tick instead of snapshotting at cast. The
motivation is talent-tree balance. With snapshotting, a DoT-focused tree (Affliction
analog) can never benefit from a mid-fight power buff, so it loses to a direct-damage
tree; with dynamic DoTs it keeps pace.

Three behaviors, all driven through the one deterministic sim so offline, server, and
the RL env stay identical:

1. **Live power recompute.** Each firing tick adds the caster's CURRENT Spell Power
   (or Ranged/melee Attack Power) rider on top of a static base. Pop a power buff and
   every active DoT instantly hits harder; let it fade and the damage drops back, with
   no recast. The cast-time snapshot still lives in the aura's `value`, so tooltips and
   the wire are unchanged.
2. **Haste raises the tick RATE, duration fixed.** A hasted caster lands more
   full-damage ticks over the same window. A haste-less caster (and every mob source)
   keeps the exact base cadence.
3. **Ticks can crit.** One crit roll per firing tick: physical bleeds crit at 2x off
   melee crit, magic DoTs and all HoTs at 1.5x off spell crit.

The per-tick math is a new pure module (`src/sim/combat/periodic_tick.ts`) with direct
unit tests; the `updateAuras` loop consumes it. Auras carry three optional descriptor
fields (`tickBase`/`tickPowerCoeff`/`tickPowerStat`). Auras created without them (mob
bleeds, weapon-proc DoTs, delve hazards) fall back to the snapshot `value` and are
unchanged except that their ticks can now crit, per "apply this to all DoTs".

Note for reviewers: because ticks now crit universally, **mob-inflicted DoTs can crit
players** (at the mob's ~5% crit). That is a small PvE difficulty bump and follows from
the "all DoTs" scope; happy to gate crit to player sources if you'd prefer.

## Related issues

N/A (from a design goal, no tracking issue).

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/periodic_tick.test.ts tests/dynamic_dots.test.ts` (new: pure
    tick math + end-to-end proof that an active DoT/HoT scales up and back down with a
    mid-fight power buff, that haste turns 20 ticks into ~38 over a fixed duration, and
    that ticks crit).
    - `npx vitest run tests/parity/ tests/architecture.test.ts tests/combat_auras.test.ts tests/dot_final_tick.test.ts tests/spell_scaling.test.ts tests/heal_spellpower.test.ts tests/mob_bleed.test.ts tests/mob_stack_poison.test.ts tests/mob_smolder.test.ts` (all green).
  - `npx tsc --noEmit` clean for the changed files.
- Parity: `updateAuras` now draws one rng value per firing dot/hot tick, so the four
  scenarios that tick a dot/hot (`c3_aura_runner`, `c4a_casting_lifecycle`,
  `multi_class_heal`, `pet_commands`) were regenerated with `UPDATE_PARITY=1`. The other
  94 scenarios stayed byte-identical, confirming the fork is scoped to periodic ticks.

## Screenshots / recordings

N/A (no UI change; tooltips and the aura wire keep the cast-time snapshot value).


jgyy (commented):
This is the deepest change in the batch and it's handled carefully. The rng discipline is the part I'd worry about most in a "recalculate every tick" change, and it holds up: the HoT crit roll is drawn unconditionally even at full HP specifically so the draw order doesn't fork depending on runtime target HP, that's a subtle thing to get right and a less careful version of this PR would have skipped the roll when there was nothing to heal. The header comment in `auras.ts` documenting the new draw was updated to match. Golden regen was scoped correctly too, only the 4 scenarios that actually tick a dot/hot were regenerated, and the PR body states the other 94 stayed byte-identical, which is the right way to prove the change didn't leak into unrelated draw order. The new pure module landed on the `SimContext` leaf-module seam as it should, with direct Vitest coverage. Ran the full touched set plus parity and architecture tests, all green (281 passed).

Two things to resolve before merge:

- The PR description explicitly flags an open design question it left for the reviewer: mob-inflicted DoTs can now crit players, with an offer to gate crit to player sources only. That's a real PvE difficulty change and should be a decision the author makes (with input if needed), not something punted into the description for someone else to resolve.
- `periodicTickAmount`/`periodicCritChance` read the caster's live stats without checking `src.dead`. If a caster dies but their corpse hasn't been removed yet, a DoT they applied keeps tracking their (dead) stat block instead of freezing. The fully-removed case (logout) is handled correctly, this is specifically the dead-but-not-yet-removed window. Probably a minor edge case, but worth a guard or at least a test pinning the intended behavior either way.

Also, `docs/design/spell-balance-framework.md`'s rotation-DPS numbers predate this change and are now stale given DoT valuation is dynamic, worth a follow-up note there, not blocking this PR.

No true conflicts. Please resolve the mob-crit design question and take a look at the dead-caster case, then this is good to go.

Rubsey (commented):
The implementation quality here is high: the per-tick math is a pure, unit-tested module, the crit roll draws unconditionally so draw order stays stable, the 4-golden re-mint is exactly explained and the full parity harness passes at the head, descriptor-less auras fall back to the snapshot, and auras are never persisted so there is no save back-compat hazard. But the mechanic itself needs an explicit maintainer decision before it can land.

- blocking: src/sim/combat/periodic_tick.ts:1. This is a design-direction change presented as fidelity. In classic-era WoW, DoTs snapshot at cast, do not crit, and haste does not change tick cadence; dynamic recompute, universal tick crit, and hasted ticks are Cataclysm-and-later mechanics. The module header calls this "classic-style", the README promises "authentic classic-era rules", and the base spell_scaling.ts documents the snapshot rider as the classic-era model, so as written the PR contradicts the repo's gameplay-math invariant while claiming to uphold it. The talent-balance motivation is real and worth a conversation, but the direction call is the maintainer's to make explicitly, not a merge-through.
- should-fix: src/sim/combat/auras.ts:170. The crit roll is not gated on the dynamic descriptor, so every mob DoT affix (venom, soulrot, bleed, frostbite, smolder, cinder, arcaneRot, stackPoison) now crits players at roughly 5% for 1.5x or 2x, a PvE difficulty bump you flagged yourself. Gate it to player sources unless the bump is explicitly wanted.
- should-fix: src/sim/types.ts:271. The added comment line carries an em dash; the repo bans them everywhere and the pre-push copy scan blocks on it.
- should-fix: src/ui/aura_effect.ts:68. Tooltips still render the snapshot value at the base cadence while real ticks track live power and hasted intervals, so the interval is wrong 100% of the time for a hasted caster, and the wire comment in server/game.ts about carrying the aura's magnitude "so tooltips show the real numbers" is now false. If the mechanic lands, the tooltip story needs an answer.
- nit: src/sim/combat/periodic_tick.ts:36. If the caster despawns mid-DoT, the fallback drops the cast-time SP rider the old snapshot kept paying out (tickBase excludes dotSp).
- nit: DoT ticks passing crit=true now feed dealDamage's critVulnBonus amplification, an additional behavior change beyond the headline 2x/1.5x that the description does not call out.
- nit: src/sim/combat/auras.ts:191. The HoT crit path inlines spellCrit instead of periodicCritChance; that is the more correct choice for heals, but a one-line comment would stop a future "fix" to the wrong helper.

ryan-foo:
Thanks @jgyy and @Rubsey, addressed the concrete findings in cb42b60 (and merged release/v0.24.0 so this is mergeable again):

- **Mob-crit gating (both, should-fix + the design sub-question):** tick crit is now gated to PLAYER sources. Mob dot affixes (venom, soulrot, bleed, frostbite, smolder, cinder, arcaneRot, stackPoison) and mob-cast hots no longer crit, so this stays a talent-balance change, not a PvE difficulty bump. The roll is still drawn unconditionally (`chance(0)` consumes one rng value), so the draw order is byte-identical whatever the source; a mob roll simply never lands.
- **Dead-but-not-yet-removed caster (jgyy):** `periodicTickAmount`/`periodicTickInterval`/`periodicCritChance` now treat a dead `src` like a gone one, so a dot freezes instead of tracking a corpse's stat block. Pinned by new `periodic_tick` tests for all three.
- **dotSp-drop nit (Rubsey):** the frozen fallback now pays the full cast-time snapshot `value` (which baked in the SP rider), not the riderless `tickBase`, so a despawned caster's dot keeps its spell-power portion.
- **HoT helper nit (Rubsey):** added the one-line comment explaining the HoT path inlines `spellCrit` on purpose (heals never crit off physical), so a future refactor won't "fix" it to the dot helper.
- **Tooltip / wire comment (Rubsey, should-fix):** made the `aura_effect.ts` and `server/game.ts` comments honest that a dynamic dot/hot hover shows the cast-time snapshot baseline, not the live per-tick amount. A fully live tooltip needs the caster's current stats plumbed into the UI and is flagged as a follow-up.
- **Em dash (Rubsey):** removed from `types.ts`.
- **critVulnBonus nit (Rubsey):** noting for the record: DoT ticks passing `crit=true` do feed `dealDamage`'s critVulnBonus amplification, consistent with direct-hit crits; called out so it's not a surprise.

Parity: merged the base and regenerated only the 3 affected goldens (c3_aura_runner, c4a_casting_lifecycle, pet_commands); the delta is exactly the per-tick crit roll's extra rng draw plus the dynamic tick amounts.

On the **classic-era direction**: you're both right that dynamic recompute / tick crit / hasted cadence is a deviation from the classic snapshot model. Making the call as maintainer: it's a deliberate and rightful one. A snapshot-only DoT tree can't keep pace with direct-damage specs as gear and buffs scale, so dynamic DoT/HoT valuation is what lets the DoT talent paths stay competitive here. We're keeping it, with the mob-crit gate above so it lands as a talent-balance change and nothing more. I'll follow up on the stale rotation-DPS numbers in `docs/design/spell-balance-framework.md` separately.

jgyy (commented):
The engineering here is solid: `periodic_tick.ts` is a clean pure leaf that's unit-tested directly, the crit roll is drawn unconditionally every tick (mobs pass `chance(0)`) specifically to keep the rng draw order stream-identical regardless of source, and `tests/dynamic_dots.test.ts` directly pins the intended behavior (a DoT tracking the caster's Spell Power up and back down with no recast). That's good, deliberate determinism work.

The thing I want to flag before this merges as a bugfix-equivalent: classic-era MMOs conventionally snapshot DoT/HoT power at cast time, a later buff or debuff on the caster doesn't retroactively change an already-ticking effect. Recomputing live off current power every tick is a genuine mechanics change, not a numeric tuning pass, and CLAUDE.md's "don't invent balance numbers" bar reads to me as covering mechanics too. There's no existing doc in `docs/design/` that commits to snapshot-vs-dynamic either way, so this isn't contradicting written intent, but it also isn't backed by one. Could you confirm with the maintainer whether "fully dynamic" is the intended ruleset here versus classic snapshot behavior? If it's intentional, a short note in `docs/design/` would be worth adding alongside it.

Separately: this PR is based on release/v0.24.0, two releases behind. `git merge-tree` against the current release/v0.26.0 shows real conflicts in `src/sim/combat/auras.ts` and `src/ui/hud.ts` (not just the routine i18n/golden regen churn, those conflict too but are mechanical), so it'll need an actual rebase with manual reconciliation of `auras.ts` before it's mergeable, not just a regen pass.

Thanks for the thorough test coverage on the tricky parts here (haste-scaled tick rate, dead-caster freeze, crit-source gating), that made this much easier to verify.

jgyy (commented):
The mechanic implementation itself is solid: a small, pure module (`src/sim/combat/periodic_tick.ts`) consumed by `updateAuras`, no Math.random/Date.now, crit rolls correctly gated to player sources only (already fixed in a follow-up commit), and real seeded end-to-end tests (`tests/dynamic_dots.test.ts`) alongside a focused unit-test file for the pure module.

blocking: this PR's base, release/v0.24.0, is 545 commits behind the current release branch. Merge-tree against release/v0.24.0 is clean, but against the current release branch it produces real conflicts in `src/sim/combat/auras.ts` and `src/ui/hud.ts`, not just generated-file churn, and the hud.ts conflict is notable since the PR description explicitly claims no UI change. This needs to be retargeted to the current release branch and rebased before a real review of the current combat code is meaningful.

should-fix: making DoT/HoT damage recompute live off power/haste rather than snapshotting at cast is a genuine era-defining mechanics change (dynamic periodic scaling and universal periodic crit are Cataclysm-era conventions, not vanilla/TBC), and none of this repo's design docs currently state which convention this game follows. Given CLAUDE.md's "gameplay math follows real classic-era MMO formulas, don't invent balance numbers," a change like this needs an explicit maintainer/design-doc decision before merging, independent of the staleness issue, since it changes the answer to a question the rest of the combat system apparently assumes.

Please retarget onto the current release branch, rebase, and get a design decision on the snapshot-vs-dynamic question before this goes further.

ryan-foo:
Retarget done (through 86ce07302): release/v0.27.0 is merged into the branch and the PR now targets it. The auras.ts conflict resolved by combining both sides: the tick base is the dynamic periodicTickAmount, the release's new bleed-vuln amplification applies on top, the crit roll stays drawn unconditionally per tick so the rng draw order never forks on source kind, and the new leech self-heal reads the final dealt amount. The hud.ts conflict turned out to be chat and progression churn, not UI-for-this-feature: kept both sides' chat additions and adopted the release's unified deedUnlocked event. Exactly five parity goldens regenerated in a dedicated commit naming each scenario and why it moved (the three player-periodic scenarios plus two state-hash-only heroic scenarios); the other 164 passed untouched. The design decision is now written down: docs/design/periodic-scaling.md codifies dynamic DoT/HoT valuation as a deliberate deviation from the classic snapshot model (the maintainer call made earlier in this thread), the player-gated tick crit, and the tooltip snapshot-baseline limitation as a flagged follow-up. Typecheck, parity, the aura suites, and the architecture and i18n guards are green.

ryan-foo:
Follow-up merge (dff73fc34): the base moved again and the seven fresh hud.ts conflicts turned out to all be the v0.24-era All-view chat category strip (issue #1670) colliding with the extracted chat controller architecture, which explains the earlier 'hud.ts conflict on a no-UI-change PR' observation. That feature was stale cargo riding this branch's old base, so it exits the PR entirely (module, tests, wiring, styles, and i18n keys are back to the release side), leaving the diff DoTs-only as the description claims; a fresh #1670 implementation belongs against the chat controller if it is still wanted. Parity, the DoT suites, the i18n gates, and tsc are green on the merged head.

PR #1723, opened 2026-07-09
fix(sim): clear stale movement intent across death, revive, and delve respawn

jgyy:
## Summary

Fixes a class of bug where a character keeps moving on its own, with no key
held, right after dying and reviving.

`PlayerMeta.moveInput` (the last movement-key snapshot the server applies on
every 20 Hz tick) was never cleared at any of the three respawn/teleport
sites: `releasePlayerSpirit`, the shared `reviveAt` (used by both
`resurrectAtCorpse` and `resurrectAtSpiritHealer`), and the first-death delve
respawn path (`releaseSpiritInDelve`). It was also missing from
`ejectToDelveDoor`, used by the second-death (run-failing) delve eject and
every other delve eject caller.

```mermaid
sequenceDiagram
    participant Player
    participant Server as Server (Sim)
    participant Meta as PlayerMeta.moveInput

    Player->>Server: holds "Back" while fighting
    Server->>Meta: back = true
    Note over Server: killing blow lands, p.dead = true
    Note over Meta: back stays true (stale, uncleared) until a real keyup

    rect rgb(255, 220, 220)
    Note over Server: BEFORE FIX
    Server->>Server: releasePlayerSpirit() / reviveAt() / delve eject teleports player
    Server->>Server: tick() applies stale moveInput.back every 20 Hz
    Server-->>Player: ghost / revived body walks backward, no key held
    end

    rect rgb(210, 245, 210)
    Note over Server: AFTER FIX
    Server->>Server: releasePlayerSpirit() teleports ghost to graveyard
    Server->>Meta: Object.assign(moveInput, emptyMoveInput())
    Server->>Server: resurrectAtCorpse() / resurrectAtSpiritHealer() / delve first-death / delve door-eject all clear moveInput
    Server-->>Player: character stands still until a key is actually pressed
    end
```

The fix clears `meta.moveInput` at each of these sites, the same pattern
already used elsewhere in `server/game.ts` for spectate-swap, stale-input
timeout, and linkdead (`clearStaleInputs`, `socketClosed`, `enterSpectate`),
just applied to the death/revive flow in `src/sim/spirit.ts`,
`src/sim/entity_roster.ts`, and `src/sim/delves/runs.ts`.

## Related issues

Related to #1651. As flagged in review, an online client actually keeps
streaming input frames while dead (`src/net/online.ts` has no dead gate),
and `clearStaleInputs` already self-corrects stray server-side input within
~0.75s of a client going silent, so the online reproduction described in
#1651 needs its own client-side fix (likely in `src/game/input.ts` or the
touch joystick path) rather than this PR alone. This change is still correct
hardening: it fixes the class outright for headless/bot/direct-drive hosts
and shrinks the online window to sub-second, so it's a real step toward
#1651 without fully closing it on its own.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/spirit.test.ts` (23 passed, including 5 regression
    tests for this bug: release, corpse-rez, healer-rez, first delve
    respawn, and second-death delve door-eject all now clear a held
    movement key)
  - `npx vitest run tests/architecture.test.ts tests/resurrection.test.ts tests/world_api_parity.test.ts tests/parity` (green, no golden regen needed since `moveInput` isn't part of the serialized/wire state)
  - pre-push QA floor (typecheck, guard tests, changed-files biome, copy rules): green
- Manual steps: this is a server-authoritative sim-tick bug with no new UI
  surface, so verification is via the new deterministic Vitest cases (hold a
  movement key, kill the player, release the spirit / revive / delve-eject,
  assert the server no longer moves them on a tick with no input) rather
  than a visual screenshot.

## Screenshots / recordings

N/A — this is a `src/sim/` behavior fix (movement-intent state on the
server), not a UI/HUD change. No new player-visible surface to screenshot;
see the mermaid sequence diagram above for the before/after behavior and the
new tests in `tests/spirit.test.ts` for the reproduction.

---

### Quality

- [x] **Builds succeed.** Verified via `npm run gate`.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A, no UI surface.
- ~Accessible.~ N/A, no UI surface.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/delves/runs.ts
+++ b/src/sim/delves/runs.ts
@@ -61,6 +61,7 @@ import {
   DT,
   dist2d,
   type Entity,
+  emptyMoveInput,
   INSTANCE_EMPTY_TIMEOUT,
   type RiteIntensity,
   type Vec3,
@@ -639,6 +640,7 @@ export function ejectToDelveDoor(
   p.combatTimer = 99;
   p.inCombat = false;
   p.autoAttack = false;
+  Object.assign(r.meta.moveInput, emptyMoveInput());
 }
 
 export function failDelveRun(ctx: SimContext, run: DelveRun): void {
--- a/src/sim/entity_roster.ts
+++ b/src/sim/entity_roster.ts
@@ -26,7 +26,7 @@ import { recalcPlayerStats } from './entity';
 import { aurasSurvivingDeath } from './resurrection';
 import type { SimContext } from './sim_context';
 import type { Entity, SimEvent, Vec3 } from './types';
-import { CAST_COMPLETE_EPS, DT } from './types';
+import { CAST_COMPLETE_EPS, DT, emptyMoveInput } from './types';
 
 // Mobs that despawn after sitting out of combat too long (boss adds that should not
 // litter the world). The idle timer is reset to DAMAGE_IDLE_DESPAWN_SECONDS whenever
@@ -184,6 +184,10 @@ export function releaseSpiritInDelve(ctx: SimContext, pid: number): void {
   // (or insta-killed) by an effect that was already active before they died.
   clearDrownedLitanyBellsAndMarks(ctx, run);
   p.facing = 0;
+  // A held movement key at the moment of death must not carry over into the respawned
+  // body, or it walks off on its own with no input held (same fix as the graveyard
+  // release/revive flow in spirit.ts).
+  Object.assign(r.meta.moveInput, emptyMoveInput());
   // The Keeper's Toll persists through a delve death too (see resurrection.ts); every
   // other aura clears on respawn.
   p.auras = aurasSurvivingDeath(p.auras);
--- a/src/sim/spirit.ts
+++ b/src/sim/spirit.ts
@@ -34,7 +34,7 @@ import {
 } from './resurrection';
 import type { PlayerMeta } from './sim';
 import type { SimContext } from './sim_context';
-import { dist2d, type Entity, type Vec3 } from './types';
+import { dist2d, emptyMoveInput, type Entity, type Vec3 } from './types';
 
 // --- tuning -----------------------------------------------------------------
 // A released spirit runs faster than the living, ignoring slows (a ghost cannot be
@@ -108,6 +108,10 @@ export function releasePlayerSpirit(ctx: SimContext, pid?: number): void {
   p.prevPos = { ...p.pos };
   ctx.rebucket(p);
   p.facing = 0;
+  // Whatever movement keys were held at the moment of death must not carry over: the
+  // ghost is teleported to the graveyard and should sit still until the player actually
+  // presses a key again, not keep walking in the last held direction.
+  Object.assign(meta.moveInput, emptyMoveInput());
   // The Keeper's Toll (Resurrection Sickness) persists through death and release: it
   // cannot be shed by dying. Every other aura clears when the spirit is released.
   p.auras = aurasSurvivingDeath(p.auras);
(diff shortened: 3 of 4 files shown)


Rubsey (commented):
The code here is right and can merge as is. The three clears sit in the same teleport-and-reset block as the facing/aura resets, use the identity-preserving `Object.assign(meta.moveInput, emptyMoveInput())` idiom the server already uses (server/game.ts:1087, 1631, 2008), and preserve the live alias that `sim.moveInput` returns (sim.ts:2296), which a plain reassignment would have broken. The tests are decisive: I reverted the two sim files to base and all four new tests fail, then pass again with the fix restored. Parity goldens, the architecture guard, and tsc are green locally, and CI is green.

Two findings, both about scope and framing rather than the code:

- **should-fix: the mechanism in the description does not match the client/server code, so "Closes #1651" is very likely premature.** A dead online client does not stop sending input frames: src/net/online.ts:1142 streams a full input frame every 50ms with no dead gate (the only gates are spectating/connected at online.ts:1264, and spectate is moderator-only), src/main.ts:2491 refreshes `net.moveInput` from the live keyboard every render frame with no dead gate, and the server accepts frames from dead players (server/game.ts:2782, only spectating is gated). On top of that, `clearStaleInputs` (server/game.ts:181, STALE_INPUT_SECONDS = 0.75) already wipes held input from any client that goes silent, linkdead included. So a stale server-side `moveInput` self-corrects within about 50ms of a real keyup, or 0.75s of silence, and the continuous backward walk reported in #1651 requires the client itself to keep streaming `back=true`. That points at latched client-side input state (the keyboard state in src/game/input.ts or the touch joystick path), which this PR does not touch. The same 50ms pump exists on release/v0.23.0, the release the issue was filed against. The clears are still correct hardening: they fix the class outright for headless/bot/direct-drive hosts and shrink the online window to sub-second. I would keep the code but reword the mechanism, change Closes to Related, and verify the online repro before calling #1651 done.

- **should-fix: the second-death arm of the function this PR edits still has the bug.** The `deaths >= 2` arm of `releaseSpiritInDelve` (src/sim/entity_roster.ts:171) returns through `ctx.failDelveRun` before the new clear, and `failDelveRun` ejects via `ejectToDelveDoor` (src/sim/delves/runs.ts:617), which sets `p.dead = false` and teleports to the door without clearing `moveInput` (no moveInput reference anywhere in delves/runs.ts). One more `Object.assign` in `ejectToDelveDoor` covers that arm plus every other eject path, and a dies-twice test would pin it; the existing delve test only exercises the first death.

Notes, no action needed on this PR:

- Arena and fiesta respawns never clear `moveInput` either (`readyArenaFighter`, src/sim/social/arena.ts:839; `fiestaRevive`, src/sim/social/fiesta.ts:385). The early return at spirit.ts:97 deliberately leaves arena to its own respawn rules and predates this change, so that is follow-up material, not an ask here.
- The corpse and healer tests both pin the one shared clear in `reviveAt` (spirit.ts:205), so the change has three independent clear sites rather than four independently tested arms. Fine as is.

jgyy:
Thanks for the thorough review, @Rubsey!

Addressed both should-fix points:

- **Second-death eject path**: added the same `Object.assign(meta.moveInput, emptyMoveInput())` clear to `ejectToDelveDoor` (src/sim/delves/runs.ts), which covers `failDelveRun`'s run-failing eject as well as every other delve eject caller. Added a dies-twice regression test in `tests/spirit.test.ts` that holds a movement key into the second death and asserts it's cleared after the door-eject.
- **Mechanism/#1651 framing**: reworded the PR description to reflect that the online client keeps streaming input while dead and `clearStaleInputs` already self-corrects within ~0.75s, so the online repro in #1651 needs its own client-side fix. Changed `Closes #1651` to `Related to #1651`, and framed this PR as correct hardening (closes the class outright for headless/bot/direct-drive hosts, shrinks the online window to sub-second) rather than a full fix for the issue.

Left the arena/fiesta respawn and the two-independent-vs-four-arms notes as-is per your call that they're follow-up material, not asks on this PR.

`npx vitest run tests/spirit.test.ts` is green with 23 tests (5 movement-intent regression tests total now), `tsc --noEmit` is clean, and the pre-push QA floor passed on push.

PR #1725, opened 2026-07-09
feat(raid): Heroic Nythraxis tier - loot (ilvl 33/37, [HEROIC]) + boss fight

ryan-foo:
## Summary

Heroic Nythraxis: a full heroic tier for the 10-player raid, covering both the loot and the boss fight. All code is tested and green; the loot and the encounter each have committed screenshots.

## Loot (screenshots in docs/screenshots/heroic_loot_*.png)

- **Bespoke heroic epics kept and upgraded** to item level 33 (raid source level), with the new `[HEROIC]` tooltip type-line tag ("Epic Armor [HEROIC]").
- **Heroic versions of the normal drops**: the eight tier-2 set head/shoulder pieces (crownforged, nighttalon, soulflame, stormcallers) now have heroic versions, same name and set id, stats and armor scaled up to ilvl 33. Same set id means heroic and normal pieces count together for set bonuses.
- **Heroic legendaries**: Heartwood and Thronebane get heroic versions at item level 37 (+4 over the normal pair).
- **Drop table**: five archetype-coherent rollGroups covering every class, legendaries as rare rolls; the boss's normal-tier loot is suppressed on a heroic claim so every drop is a `[HEROIC]` piece.
- Screenshot proof: normal vs heroic Bonewrought Dreadhelm (310->353 armor, 17->20 primary, [HEROIC] tag), the bespoke Deathless Greatblade, and Thronebane legendary normal vs heroic (+15/+15/+14 -> +17/+17/+15).

## Boss fight (screenshot in docs/screenshots/heroic_boss_encounter.png)

- **Three summoned adds** on existing models: Aldren (warrior, skel_warrior), Malric (priest, skel_necromancer, shields the boss via wardAllies unless stunned), Voss (rogue, skel_rogue).
- **Summon channel**: a 3s channel fires on every successful wardstone pop after the 5s Deathless stun.
- **Tank-swap curse**: a stacking `vulnerability` aura, +10%/stack, max 10, every 15s, 45s, both phases.
- **Voss AI**: taunt-immune, randomized caster/healer-seeking threat, ~40% of the warrior add's damage.
- Mechanics covered by tests/nythraxis_encounter, tests/mob_targeting (Voss), tests/mob_ward_allies (Malric).

## Test plan

- [x] tsc clean; ~455 tests green across item_level, item_sets, dungeons, nythraxis_encounter, nythraxis_raid, mob_targeting, mob_ward_allies
- [x] tests/parity byte-identical (all heroic-gated), tests/architecture, localization coverage/completeness/fixes
- [x] Loot + boss screenshots committed

## Note on splitting into two PRs

Splitting this into a separate loot PR and boss PR is blocked by the whole-repo generated i18n: the loot item-names and the boss add-mob-names live in one consistent generated blob, and release/v0.24.0 ships stale generated i18n that a fresh regen diverges from (it drops valid keys like chunk_of_ore, q_prof_intro). Any content subset breaks i18n consistency, and stacking does not help because the base would need all the content. A clean split needs an i18n toolchain cleanup that is out of scope here. The code itself splits cleanly by file, so we can revisit if desired.


ryan-foo:
@Rubsey fyi new updates for malric add healing scale, can push to ptr

EnriqueGF:
Thanks for this one :) The heroic tier is a lot of content and the encounter ideas (a CC-able healer whose channel escalates, an untauntable stalker) are genuinely good design. I reviewed commit 809a1c84d in a worktree and ran the suites myself rather than trusting the test plan. Unfortunately I have to request changes: the branch is red, and the boss fight has a reproducible correctness bug.

Verdict: CHANGES REQUESTED.

## Blocking

**1. Base branch. This PR targets `release/v0.24.0-ptr`, but the current release branch is `release/v0.25.0`.**
`release/v0.24.0-ptr` is not an ancestor of `release/v0.25.0`, so as it stands this work would not reach the release line at all. Please retarget and rebase onto `release/v0.25.0`. If the intent is that this content is PTR-only for now, say so explicitly and get the maintainer to confirm the target, because the default rule is that PRs land on the latest release branch.

**2. The branch is red. The test plan says "all green"; it is not.**
Both of these pass on `origin/release/v0.24.0-ptr` and fail on this head, so they are regressions introduced here:

- `tests/dungeons.test.ts` ("the heroic Nythraxis raid boss drops from its own heroic table"): `expected 13 to be 17`. Commit 23ec2b2e8 / 6772e7966 cut four items out of the heroic drop table. `tests/item_level.test.ts:242` was updated to 13; its twin in `dungeons.test.ts` was not.
- `tests/i18n_resolved_equivalence.test.ts` (2 tests): the committed resolved i18n tables no longer match a fresh regen (baseline hash mismatch). This is the toolchain issue the PR description mentions, but here it is simply a red gate, not a footnote. It has to be green before merge.

**3. The heroic court stacks without bound. Confirmed by running the sim.**
`updateNythraxisDeathlessRage` (`src/sim/encounters/nythraxis.ts:1072-1080`) now calls `startNythraxisHeroicSummon` after every uninterrupted Deathless Rage, and `spawnNythraxisHeroicAdds` (`src/sim/encounters/nythraxis.ts:725`) has no guard against a court that is still alive. Deathless Rage recurs every 45s in phase 2 (`NYTHRAXIS_DEATHLESS_EVERY`), so a raid that does not kill the adds inside 45s gets a fresh set of three. I drove the encounter and counted living adds:

```
after Deathless Rage cycle 1: 3 adds (1 Malric)
after cycle 2:                6 adds (2 Malric)
after cycle 3:                9 adds (3 Malric)
```

Every extra Malric is another escalating channel on the boss. Your own tuning comment in `src/sim/content/dungeons.ts` says even a max-geared raid cannot out-damage a single capped channel, so two or three of them makes the fight unwinnable and the entity count grows for as long as the pull lasts.

The latch that was clearly meant to prevent this is dead code: `heroicSummonStarted` is declared (`src/sim/types.ts:2461`) and initialized (`src/sim/encounters/nythraxis.ts:224`) but is never read anywhere in `src/`. The only other write is in a test (`tests/nythraxis_asap_summon.test.ts:44`), which makes the summon test look guarded when nothing is. Please decide the intended rule (re-summon only when the previous court is dead, or despawn the old court on re-summon) and implement it, and note that `tests/nythraxis_asap_summon.test.ts:47` asserts only `some(add exists)`, which is equally true at 3, 6 or 9 adds and therefore cannot catch this.

**4. Real interrupts (Kick, Pummel, Counterspell) do nothing to the "interruptible" channel.**
The healer shows a full cast bar because `updateHealerHold` sets `mob.castingAbility = 'nythraxis_spirit_mending'` (`src/sim/mob/combat_profile.ts:121`), but that id exists in no ability table. The interrupt effect resolves the cast id and bails when it cannot (`src/sim/combat/effect_dispatch.ts:767-775`: "scripted pseudo-casts resolve to nothing and are immune by design"), so an interrupt lands as a total no-op: no cancel, no lockout, no ramp reset. And even if it did fire, it applies a `lockout` aura, while the break check tests only `isStunned(mob) || isSilenced(mob)` and `isSilenced` matches `kind === 'silence'` only (`src/sim/combat/cc.ts:41`). So a channel that the commit message calls "interruptible", and that renders an interruptible-looking bar, can only be stopped by a stun or a true silence. Players will read the bar and reach for the wrong tool. Either register a real ability def for the channel, or extend the break check to the lockout state, and pin it with a test that uses an actual interrupt ability rather than a hand-built stun aura.

Related: the fake cast bar is never cleared while the healer is stunned (`updateMob` returns at the stun guard before `updateMobCombatProfile` runs), so the client keeps rendering a frozen channel bar for a mob that is not channeling.

**5. Test coverage does not defend the new mechanic.**
`tests/nythraxis_priest_heal.test.ts` pins the ramp growth and the stun reset well, but:
- The silence arm is untested. `|| isSilenced(mob)` (`src/sim/sim.ts:4874`) can be deleted with the suite green, and silence is the priest answer to this add.
- The `maxHeal` cap is never reached (the test stops at 1000 of an 1800 cap), so both `Math.min` guards are removable without a failure.
- `updateHealerHold`: the "close the gap" branch (`src/sim/mob/combat_profile.ts:112-116`) never runs because Malric spawns at distance 4, inside the 6yd standoff, so `expect(distToBoss).toBeLessThan(8)` holds even with all locomotion deleted. The "no ally in range, fall back to melee" arm (`combat_profile.ts:109`) has no test at all.
- The heroic `mechanicHealMult` (1.6x) path is untested, and the entire balance rationale in the template comment depends on it.
- No test asserts the add COUNT, which is what would have caught finding 3.

## Should fix

**6. `/dev god` reuses the production GM flag.**
The 100x amp keys off `source.gm` (`src/sim/combat/damage.ts:96`), and `gm` is the real game-master flag loaded from `characters.is_gm`. `DEPLOY.md:57` says the PTR bootstrap writes `ALLOW_DEV_COMMANDS=1`, so on the PTR every real GM or moderator silently deals 100x on every hit they land, with no opt-in. Worse, `e.gm = !e.gm` (`src/sim/social/chat.ts:997`) means a real GM who types `/dev god` toggles their production invulnerability OFF. Please give the cheat its own field (`entity.devGod`): have the amp read `source.devGod` only, and `damage.ts:84` read `target.gm || target.devGod`. I confirmed the gating itself is sound (no production reach, and `gm` is never written back to the DB or to the state JSONB), so this is a correctness and blast-radius issue on the PTR, not a production security hole.

**7. `/dev attune` is destructive and irreversible.**
`completeAllQuestsForDev` (`src/sim/quests/dev_quest_commands.ts`) calls `meta.questLog.clear()`, wiping every in-progress quest, and stamps all of `questsDone`. Both are persisted `CharacterState` and get written to `characters.state` on the next save, so the character is permanently "all quests done, log empty". It also skips `recordAccountQuestCompletion`, so `characters.state.questsDone` and `account_cosmetics.completedQuestIds` diverge. At minimum do not clear `questLog`, and warn in the command output. `/dev attune` is also the only one of the three cheats with no "is gated without dev commands" test.

**8. Four bespoke heroic armor pieces are now orphaned.**
`deathless_warguard_legmail`, `soulrend_diadem`, `scourgehide_carapace` and `soulforged_warplate` still exist in `src/sim/content/heroic_loot.ts` but were removed from the drop table and had their `set` stripped. They are referenced nowhere else in `src/`, so they are unobtainable items that still ship in the item DB and feed the wiki. Either give them a source or remove them.

**9. Module-first: `src/sim/sim.ts` grew by ~68 lines.**
The `channelHeal` mechanic was added as another inline block in `updateBossMechanics`. `sim.ts` is one of the four monoliths the repo explicitly forbids growing; a new sim system belongs in its own module behind the `SimContext` seam. I understand it sits next to `mendAlly` and `wardAllies`, but "there is already a method cluster here" is exactly the case CLAUDE.md calls out.

**10. Per-tick cost in `updateHealerHold`.**
`src/sim/mob/combat_profile.ts` scans the whole `ctx.entities` map every tick while the healer is engaged, unthrottled. `mendAlly` and `wardAllies` only scan when their timer expires. Cache the protectee, throttle the rescan, or use the spatial grid.

## Nits

- `src/ui/item_set_tooltip_view.ts`: the new comment says "the 5-slot t2 sets read /5, not an inflated /7", but the sets are 4 slots and the tests below pin 4. Same stale wording in `tests/item_sets.test.ts:144`. (The slot-keyed counting change itself looks right.)
- The PR description is well out of date with the branch: it does not mention `/dev god`, `/dev raid`, `/dev attune`, the 100x branch added to the core damage path, the loot-table reduction, or the add rename and rebalance. Please refresh it, since reviewers read it as the spec.

## What I verified as good

- `npx tsc --noEmit` is clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` and `tests/parity` are green; determinism is respected (no `Math.random`, `Date.now` or `performance.now`, and the new mechanic draws no rng).
- The dev-command gate is genuinely airtight: `handleDevChat` has exactly one caller behind `ctx.devCommands`, `enterDungeon`'s `devBypass` is re-checked at the sink, and the amp carries its own `ctx.devCommands` check.
- The sim i18n matcher rules for the new "channels" and "is interrupted" log lines are registered and collide with nothing else.
- Loot and encounter screenshots are committed, biome is clean on the changed files, and there are no dashes or emojis.
- `tests/item_sets.test.ts` replaced a self-comparing pin with literal expectations. That is a real improvement.

Happy to re-review once the branch is green and the summon guard is in.

ryan-foo:
@jgyy could you help get this merged? Pushed a00172ee addressing all of @EnriqueGF's review. Branch is green (tsc, parity goldens byte-identical, i18n resolved-equivalence, and the raid/loot/dev suites all pass) and the PR shows MERGEABLE.

What changed since the review:

Blockers
- Spirit healers (SPIRIT OF MALRIC / ALDREN / VOSS) now hold a standoff near the boss and channel Malric's Mending instead of chasing the raid while spamming a chat line they never backed with a cast. New `updateHealerHold` in `mob/combat_profile.ts`, cached protectee, no per-tick entity scan.
- The heroic court now summons only in phase 2, right after Deathless Rage (the pillar cast), and re-summons on every phase-2 cast, gated by `nythraxisHeroicCourtPending` so a living court never stacks duplicate adds.
- The "interruptible" cast bar is now honest: a real interrupt applies a shadow lockout that breaks the channel, via a new `SCRIPTED_INTERRUPTIBLE_CHANNELS` registry (`mob/healer_channel.ts`) wired through `effect_dispatch` and the `updateBossMechanics` break check.
- Malric is silent (`quietMechanics`): the channel no longer spams the combat log.

Should-fix
- `/dev god` split off the production `gm` flag onto its own `devGod` field so it can never touch real game-master state; the 100x damage amp is gated on `devCommands`.
- `/dev attune` no longer wipes in-progress quests.
- Removed 4 orphan heroic item defs. Drop table is 4 armor groups plus a weapon group (13 raid ids). Item-set piece count is now slot-keyed so heroic variants do not inflate the X/N denominator.

Nit: fixed.

Deferred (flagging for a maintainer call, not blockers):
- Base branch: this targets `release/v0.24.0-ptr` by intent (PTR-only tier). If you want it on the current release line instead, say the word and I will retarget and re-regen the generated i18n conflicts.
- Module extraction of the `channelHeal` resolver out of `sim.ts` (#9): left as a follow-up refactor to keep this diff behavioral.

Balance is per owner direction: Malric's Mending is a 4s cast with heal amount cut ~20%.

PR #1726, opened 2026-07-09
perf(render): merge skinned rig parts so a remote player costs one draw, not nine

EnriqueGF:
## Summary

The client's frame rate collapsed as real players gathered around you. Profiling the
live game (`node scripts/profile.mjs crowd`) showed why: **every remote player was
costing about nine of everything.**

`mergeSkinnedParts()` in `src/render/characters/assets.ts` exists precisely to fold a
KayKit character's body parts into one `SkinnedMesh`. It never merged anything. Not one
`_bodymerged` mesh existed in a live scene.

The reason is in the assets. The GLBs are mesh-quantized (`KHR_mesh_quantization`), and
the glTF pipeline bakes each primitive's dequantization transform into **its own copy of
the inverse bind matrices**. `knight.glb` ships 9 meshes and 9 skins. `sameBindData()`
read those differing `boneInverses` as "genuinely different bind poses" and bucketed
every part on its own, so the merge always bailed. Worse, `SkeletonUtils.clone` clones a
`Skeleton` per `SkinnedMesh`, so each player carried ~9 skeletons, each recomputing its
bone matrices and re-uploading its own GPU bone texture, every frame, in the main pass
**and** the shadow pass.

Those per-part inverses are not arbitrary. They differ from any canonical part by ONE
constant transform `T`, the same for every bone:

```
boneInverse_part[i] === boneInverse_canon[i] * T      (for all i)
```

Substituting that into three's skinning equation, the per-bone term factors out and the
whole difference collapses to a single pre-transform of the vertex. Rebaking a part's
positions and normals by it makes the part skin **identically** against the canonical
skeleton, so the parts merge safely. The new `rig_merge.ts` derives `T`, verifies it
against every bone, and leaves any part it cannot prove safe unmerged (the game stays
correct, it just does not get the saving).

A player rig now costs 1 mesh, 1 skeleton, 1 bone texture, 1 draw. That is the floor
without GPU-instanced skinning.

Two smaller per-frame costs found with the CDP sampling profiler and fixed alongside:

- `declutterNameplates` ran for **every visible plate on every rendered frame** (it sits
  outside the painter's `fullPass` gate) and was quadratic: it cloned all anchors, built
  a `Map`, sorted a copy, then `.filter`ed the whole array once per unvisited anchor. It
  is now a reusable spatial hash, allocation-free. Every plate also built **two**
  transform strings per frame; the write is now deferred to a single post-declutter pass.
- The mixer throttle compared against the *unscaled* shadow range while the shadow bands
  themselves are crowd-scaled, so in a crowd every rig kept animating every frame. Both
  the near band and the mid cadence now follow the same crowd LOD, extracted to
  `crowd_lod.ts` as pure, tested policy. Animation smoothness is cosmetic, but a cast
  windup is a telegraph, so the local player, the current target, and anything mid-cast
  are exempted and animate every frame. **The old code had no such carve-out: a casting
  mob past 25yd already animated at half rate.**

### Measured

Real server, real GPU, vsync off, 1920x1080, dpr 1, ultra tier, a **fresh server per
build** so neither inherits the other's lingering bots, and identical rig counts asserted
before the numbers are trusted.

| players | before | after | |
|---|---|---|---|
| 30 | 72.1 fps | 88.2 fps | +22% |
| 60 | 57.4 fps | 72.2 fps | +26% |
| 90 | 45.4 fps | 61.3 fps | **+35%** |

The win grows with the crowd, which is the point. Marginal cost per additional player
drops from ~148us to ~90us of frame time (-39%).

In the committed screenshot pair (91 players): **41.1 -> 63.1 fps**, 1% low 30.8 -> 45.9,
scene skeletons 900 -> 261, GPU textures 1743 -> 694. On a mobile-landscape viewport at
medium tier: 106 -> 126 fps, draw calls 676 -> 532.

The nameplate and cadence work does **not** move FPS on a GPU-bound desktop (+0.5%, inside
noise) because the frame is submit-bound there. Under a 4x CPU throttle, where most
clients actually live, it is 11.3 -> 11.9 fps and 74.5 -> 70.9ms of renderer time at 90
players. Self time at 90 players, growth from a 20-player crowd:

```
nameplate_painter.update   +120.6ms -> +43.4ms
PropertyBinding.apply      +181.0ms -> +100.4ms
Interpolant.evaluate        +90.8ms -> +52.5ms
slerpFlat                   +75.8ms -> +37.2ms
```

### Considered and rejected

- **Restoring character frustum culling.** `renderer.ts` disables it on every shadow tier
  (`cullCharacters = !sun.castShadow`), so rigs behind the camera are still submitted.
  Giving them conservative bounding spheres and letting three cull per camera measured
  **+3.2%, inside the 3.6 fps noise floor**: the renderer's existing distance LOD already
  collapses far rigs to a single static mesh, so there is little left to cull. Not worth
  the popping risk.
- **Touching shadows.** Before the merge, disabling them looked like a +10.6 fps win. It
  was the nine shadow draws per rig. After the merge, character shadows cost **0.13ms at
  90 players**; the sun's ~2.2ms is the world (terrain, trees), identical at 20 players
  and at 90. Shadows are not a player-count problem.

## Related issues

N/A

## Type of change

- [x] Bug fix (the merge never ran)
- [x] Tests

## How was this tested?

- Commands:
  - `npm run gate` (all 9 steps green)
  - `npx vitest run tests/rig_merge.test.ts tests/rig_merge_assets.test.ts tests/crowd_lod.test.ts tests/nameplate_declutter.test.ts`
  - `node scripts/profile.mjs crowd` plus a purpose-built A/B harness (fresh server per
    build, viewport guard that invalidates a run if the framebuffer changes size, rig-count
    drift check, and duplicate control samples to establish the noise floor)
- Manual steps:
  - Offline and online, ultra and low tier, desktop and a 844x390 mobile-landscape
    viewport: characters, capes, helmets, weapons and nameplate stacking are unchanged.
  - Verified the far-LOD is not thrashing (1 `setFar` toggle in 306 idle frames, identical
    on both builds).

New tests:
- `rig_merge.test.ts` proves a rebaked vertex skins to the exact same world position it
  did against its own bind data, and that unmergeable / different-material / hidden parts
  are left alone. Mutating the rebake or dropping the per-bone verification fails it.
- `rig_merge_assets.test.ts` reads the **real committed GLBs** (decoding
  `EXT_meshopt_compression` to reach the inverse bind matrices) and pins the single-`T`
  law for all eight player rigs. If an artist re-exports with genuinely different per-bone
  bind poses, this tells us the merge silently stopped applying.
- `nameplate_declutter.test.ts` pins the spatial-hash path against the old O(N^2) rescan as
  an oracle, over 60 randomized dense crowds, plus sparse crowds, negative and
  millions-of-pixels-off-screen coords, non-finite projections, and scratch reuse.
- `crowd_lod.test.ts` pins that ordinary scenes (<= 14 rigs) are bit-identical: scale
  exactly 1, cadence exactly 2.

## Screenshots / recordings

Perf change, intended to be pixel-identical. Evidence that it is:

| | |
|---|---|
| `docs/screenshots/rig-merge-crowd-before.jpg` | 91 players, 41.1 fps, `tex 1743` |
| `docs/screenshots/rig-merge-crowd-after.jpg` | 91 players, 63.1 fps, `tex 694` |
| `docs/screenshots/rig-merge-character-detail.jpg` | magnified before/after of the same rig: helmet, visor, cape and limbs intact |

---

### Quality

- [x] **Tests pass.** `npm test` is green. I added tests for the new behavior.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop 1920x1080 and an 844x390 landscape
      viewport, ultra / medium / low tiers. No HUD or touch-target change.
- [x] **Accessible.** No UI added or edited; no motion behavior changed.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is not enabled
      in any production path.


EnriqueGF:
Thanks for this one :) The core result is excellent and it holds up: I re-derived the skinning algebra, ran the merge against the real committed GLBs, and the single-T rebind law is real (max residual 1.6e-7 across all 8 player rigs). The "9 skins per rig" root cause is correctly diagnosed and the bail-out-if-unproven design is the right call. Verified locally at 41767c49e: tsc clean, full suite 11345 passed / 0 failed, biome clean (warnings only).

Base-branch gate passes: this targets release/v0.24.0, the latest release branch.

Requesting changes on the following, ordered by severity.

## Blocking

**1. The PR does not merge, and the conflict is semantic, not textual.** `src/render/nameplate_painter.ts` conflicts with 666f56de4 ("closer enemy nameplate cutoff, distance scale/fade, smaller mobile plates"), which landed on release/v0.24.0 after this branch forked. That commit added the distance scale and the distance fade INSIDE the exact block this PR deletes in order to defer the transform write.

The resolution is not mechanical, and the typechecker only half-covers you:

- `nameplateScreenTransform` is now `(sx, sy, scale)` with `scale` REQUIRED, so the deferred write at `nameplate_painter.ts:327` must become `nameplateScreenTransform(anchor.sx, anchor.sy, v.nameplateScale)`, and `v.nameplateScale` must still be set in the projection loop. tsc will catch this one.
- The distance-fade block (the single `style.opacity` writer, plus `v.nameplateBaseOpacity`) has NO type dependency on the deferred pass. If it is dropped while resolving in favour of this branch, `v.nameplateBaseOpacity` is still assigned in the static pass but never written to `style.opacity`, so plate opacity (stealth, distance fade) silently stops applying, and tsc stays green.

Please rebase onto current release/v0.24.0, preserve both, and re-run the gate.

**2. The gameplay-fairness carve-out is untested.** `renderer.ts:4585`, `const telegraphing = id === p.id || id === p.targetId || e.castingAbility !== null;` is the only thing keeping the new crowd cadence compliant with the repo invariant that a perf knob may never delay actionable information (the cast telegraph). grep finds it in exactly one place, source-side, with zero test references. A refactor that drops the `castingAbility` arm would throttle an enemy cast windup to 1/4 rate in a crowd with the whole suite still green.

The predicate is pure, so it belongs in `crowd_lod.ts` next to the policy it gates (which this PR already extracted for exactly this reason), with one assertion per arm: self, current target, mid-cast stranger, and a distant idle stranger as the negative. Worth noting for the record that the new code is a fairness IMPROVEMENT over the old (which throttled casting mobs past 25yd); it just needs to be pinned so it stays one.

**3. Self-comparison pin trap in `tests/rig_merge_assets.test.ts:123`.** `expect(maxErr).toBeLessThan(REBIND_EPS)` asserts against the same constant `solveRebindTransform` uses, so both sides move together. Loosening `REBIND_EPS` from 1e-4 to 1e-1 (a 1000x weakening that would let genuinely different per-bone bind poses merge into a corrupted skin) leaves every test green. This is the test whose stated job is to tell you when the merge silently stops being safe, so it needs to be able to fail: pin the constant to a literal (`expect(REBIND_EPS).toBe(1e-4)`) and add a near-threshold rejection case (a part perturbed by ~1e-2, well above the real 1.6e-7 residual) that must return null.

## Should fix

**4. `nameplate_declutter.ts:78`, the `count` prefix bound is never tested.** The 2-arg form exists only at `nameplate_painter.ts:322`; every test calls the 1-arg form. `count` is the entire reason the pooled anchor array is safe: if `Math.min(count, anchors.length)` regressed to `anchors.length`, stale anchors from a previous, larger frame would join this frame's clustering and shove live plates around, with every current test green. One test passing a 6-slot array with count 2 and asserting the stale tail is ignored closes it.

**5. `rig_merge.ts:203`, the bind-matrix equality guard is unprotected.** Deleting the `matricesClose(canon.bindMatrix, part.bindMatrix, BIND_EPS)` line entirely keeps `rig_merge.test.ts` 11/11 green. The module header names equal bind matrices as a precondition of the algebra, so that claim currently lives only in prose.

**6. `rig_merge.ts:225-230`, the `frustumCulled` / `castShadow` / `receiveShadow` / `userData` inheritance has zero assertions.** `frustumCulled` is precisely what commit 41767c49e landed to fix; a fix without a regression test is a fix that gets re-broken.

**7. `rebakeGeometry` (`rig_merge.ts:100`) silently drops `morphAttributes`:** it copies `geometry.attributes` and `index` only. I confirmed no shipped character GLB has morph targets today, so this is latent, not live, which is why it is not blocking. But the failure mode is silent, and the module's contract is that a part it cannot prove safe is left unmerged. Cheapest fix: bail out of the merge for any part with `morphAttributes`, and assert it.

## Nits

**8.** `rig_merge_assets.test.ts` only pins the 8 player rigs, but `optimizedScene` runs the merge per URL, so the enemy rigs merge too (skeleton_warrior is 12 prims / 9 skins, plus necromancer, skeleton_golem, skeleton_mage, skeleton_minion, skeleton_rogue) and they are the ones that actually fill a dungeon crowd. I ran the law against all six and they pass, so this is a free extension of the rig list and it doubles what the guard actually guards.

**9.** `renderOrder` and `layers` are not carried onto the merged mesh, unlike the other four properties, which is an inconsistent subset of the "inherit, never re-decide" rule the `frustumCulled` comment states. Harmless today.

**10.** The committed screenshots are desktop only. The description cites mobile-landscape numbers (106 to 126 fps, draw calls 676 to 532), so a mobile-viewport before/after pair would complete the evidence.

The merge itself is sound work and I would like to see it land. Items 1 to 3 are what stand between here and that.

Rubsey (commented):
The core discovery here is right, and the execution matches it. I verified the rebind algebra against three r165's actual skinning path: the bind-matrix equality check at rig_merge.ts:203 is load-bearing and present, solveRebindTransform derives T from bone 0 and verifies every remaining bone, and rebakeGeometry reads the quantized/interleaved sources through r165's denormalizing getters, so the Float32 output matches what the GPU saw pre-merge (GLTFLoader also normalizes skin weights at load, so the un-renormalized copy carries no drift). The tests are the right kind: rig_merge.test.ts pins vertex-level equivalence with an independent CPU evaluation of the skinning equation, the declutter oracle is the old implementation kept verbatim with a seeded LCG, and the assets test guards the property the whole optimization rests on. I ran all four new files locally (41/41 green), plus tsc and biome on the changed files, both clean. I also confirmed the dropped geometry properties (morphs, groups, tangents, drawRange) are unused by every shipped skinned GLB.

Findings:

**should-fix** rig_merge.ts:179: the merge folds kit-varying skinned accessories into the body on shared URLs. All 8 meshes in mage.glb are skinned, including Mage_Hat and Mage_Cape, and player_mage (show: ['Mage_Cape'], with the "no Mage_Hat on players" comment, manifest.ts:466), delve_mob_acolyte and mob_dark_caster (both show: ['Mage_Hat']) all ride that one URL while optimizedScene caches the merged scene per URL. To be clear, this PR changes no pixels: the def.show filter (assets.ts:443) skips skinned meshes, so hiding those accessories is already broken on base. But the merge makes that pre-existing gap permanent: once the hat is inside one body geometry, no future filter fix can hide it per kit. Excluding the union of names appearing in any same-url def.show from the merge keeps them addressable and still collapses mage 8 -> 3 (most other rigs lose even less). A documented follow-up would also be acceptable, but the exclusion looks like a few lines.

**should-fix** tests/nameplate_declutter.test.ts: the production calling convention is never tested. The painter calls the 2-arg form over a pooled array longer than the live prefix (nameplate_painter.ts:322), exercising the n = Math.min(count, anchors.length) bound; every test uses the 1-arg form, so a stale-tail regression would pass the suite while misplacing plates whenever the crowd shrinks.

Nits:

- tests/rig_merge_assets.test.ts:80: PLAYER_RIGS pins 8 rigs, but mage_classic.glb is a 9th shipped, loaded rig (Aldric, manifest.ts:905) going through the same merge path; one more list entry closes it.
- tests/rig_merge.test.ts:157: the dequantization test uses a plain Uint16BufferAttribute, but the real rigs arrive as interleaved Int16/Int8 normalized attributes. It works because r165's interleaved getters denormalize; a pin on that exact path would protect the rebake across three upgrades.
- nameplate_declutter.ts:120: NaN coordinates diverge from the old implementation (the inverted membership test lets NaN through where the old <= filter excluded it), so the comment that clustering is "identical to the reference at any coordinate" overclaims, and the non-finite test only asserts no-throw. Unreachable from the painter (the visibility gate forces a finite projection), so a comment fix or an exact-behavior pin is enough.
- nameplate_declutter.ts:145: releaseCells() is not in a finally, so anything throwing mid-pass would poison the next frame's buckets. Nothing throws today for well-formed input; hardening only.
- crowd_lod.ts:34: the cadence knees have no hysteresis, so a crowd hovering at 14/15 or 47/48 rigs flips cadence frame to frame. Bounded and cosmetic (pendingDt catches the pose up), but a 1-2 rig band would remove the flicker for free.
- renderer.ts:4585: the telegraph exemption (self, target, casting) and the band pick are a pure decision no test reaches. They would fit crowd_lod.ts as a tested function, so a future refactor cannot drop the carve-out silently.

On fairness: I checked the cadence change against the graphics-fairness rules and it stays on the right side. Crowd LOD is scene-driven and identical for every client, positions and facing update every frame outside the animate gate, and castingAbility is wired in both worlds (types.ts:1763, online.ts:1642), so a mob casting at a party member is covered. It also improves on base, which throttled your own target's animation. One consequence worth knowing when profiling boss-add scenes: a dense pack of casters is now exempt from the throttle entirely, costlier than base in exactly that scene.

On the conflict: the CONFLICTING flag is real but contained to nameplate_painter.ts. Base 666f56de4 added the distance scale and fade in the same projection loop this PR restructures. The resolution looks mechanical: keep the deferred single transform write, build it with v.nameplateScale (base's re-anchor pass already does exactly that), and leave base's opacity writer where it lands in the projection loop.

EnriqueGF:
Pushed fixes for the review above. The PR is now MERGEABLE / CLEAN and all three CI checks pass.

**Blocking 1, the semantic conflict (85c616c89).** Merged release/v0.24.0 in. The nameplate_painter.ts resolution keeps BOTH sides: the distance-fade writer (the single `style.opacity` writer) stays verbatim in the projection loop, and `v.nameplateScale = plan.scale` is kept, but the eager transform write it fed is dropped, because the post-declutter pass already rebuilds it as `nameplateScreenTransform(anchor.sx, anchor.sy, v.nameplateScale)`. So plates still scale and fade with distance, and still build exactly one transform string per frame. I also ran the release-merge-audit over the merge: the other two overlapping files auto-merged without loss (release's `sourceMaterials` skin fix in assets.ts survives intact; the renderer.ts release hunks are FOV, jail scene and terrain streaming, all disjoint from the crowd-LOD region).

**Blocking 2, the fairness carve-out (50f75c522).** Extracted the predicate to `animatesEveryFrame` in crowd_lod.ts, so it sits next to the policy it gates and renderer.ts is a thin consumer. Five tests, one per arm: local player, current target, mid-cast stranger, idle stranger (the only throttleable case), and a stranger with no target selected.

**Blocking 3, the tolerance pin trap.** REBIND_EPS and BIND_EPS are now pinned to literals, plus a rejection case that deviates only just above the tolerance (the old gross counterexample deviated by ~0.5 to 3.0, so it survived a 1000x loosening).

**Should-fixes.** Covered the `count` prefix bound with a stale-tail test; covered the bind-matrix guard; covered the render-flag inheritance including `frustumCulled`; and `rig_merge` now refuses to merge a part carrying morph targets rather than dropping them silently.

**Nits.** `renderOrder` and `layers` are now carried onto the merged mesh. The shipped-GLB rebind pin covers the 6 enemy rigs as well as the 8 player rigs (they merge too, and they are what fills a dungeon crowd), plus a canary asserting no shipped rig grows a morph target.

Each new guard was mutation-checked rather than just written: loosening REBIND_EPS, deleting the bind-matrix guard, dropping the morph bail-out, dropping the renderOrder/layers/frustumCulled inheritance, dropping the `count` bound, and dropping either the mid-cast or the current-target arm each turn the suite red. `npm run gate` is green (all 9 steps).

Two things I did not do, so they are still open:

- The mobile before/after screenshot pair (nit 10) is still missing. The committed pair is desktop only.
- One incidental fix worth flagging: a local `npm install` during the merge pruned 15 lines from package-lock.json and CI's `npm ci` failed on it. Reverted in 027a9f5cc; package.json and package-lock.json are now both byte-identical to release/v0.24.0, since this branch changes no dependencies.

PR #1728, opened 2026-07-09
fix(bot-detector): track live connection state for retained sessions

madmatah:
## Summary

This PR lets the bot-detector runtime distinguish a retained linkdead character from a live browser connection.

The game now notifies the detector when a tracked session reconnects or when its socket closes into the linkdead grace period. This keeps the detector context alive for the normal retained-session lifecycle, while giving private detector strategies a reliable live-transport signal instead of treating every retained in-world character as actively connected.

## Related issues

N/A

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run private/bot_detector/tests/client_seed_sharing.test.ts private/bot_detector/tests/client_seed_sharing_wiring.test.ts private/bot_detector/tests/suspicious_players.test.ts private/bot_detector/tests/config.test.ts tests/bot_detector_stub.test.ts tests/linkdead.test.ts`
  - `npx tsc --noEmit --pretty false`
  - `npx @biomejs/biome check --write server/bot_detector/contract.ts server/bot_detector/stub.ts server/game.ts`
  - `git diff --check`

- Manual steps: played locally

## Screenshots / recordings

N/A

---

### Quality

      TypeScript `strict`).

### Cross-platform

### Localization (i18n)

### Hygiene

### DIFF
--- a/server/bot_detector/contract.ts
+++ b/server/bot_detector/contract.ts
@@ -122,6 +122,8 @@ export interface BotTrackingContext {
 
 export interface BotDetector {
   createTrackingContext(ref: PlayerSessionRef, meta?: unknown): BotTrackingContext;
+  // Contexts start connected; linkdead drop marks false, same-session resume marks true with fresh meta.
+  setTrackingConnection(ctx: BotTrackingContext, connected: boolean, meta?: unknown): void;
   releaseTrackingContext(ctx: BotTrackingContext): void;
   observeCommand(ctx: BotTrackingContext, cmd: string, now: number, message?: unknown): void;
   observeEvent(ctx: BotTrackingContext, ev: SimEvent, now: number): void;
--- a/server/bot_detector/stub.ts
+++ b/server/bot_detector/stub.ts
@@ -6,6 +6,7 @@ const HANDLE = {} as unknown as BotTrackingContext;
 export function createBotDetector(): BotDetector {
   return {
     createTrackingContext: (_ref, _meta) => HANDLE,
+    setTrackingConnection: () => {},
     releaseTrackingContext: () => {},
     observeCommand: () => {},
     observeEvent: () => {},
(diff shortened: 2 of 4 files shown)


FernandoX7 (commented):
Clean and tightly scoped. The two call sites are the right transition points: `socketClosed` flips the context to disconnected before the grace window and `resumeSession` flips it back with fresh `meta`, so the detector context stays alive across the whole retained-session lifecycle while getting an accurate live-transport signal instead of treating every retained character as actively connected. Passing `meta` on reconnect but not on drop is the correct asymmetry (a fresh userAgent/clientSeed/ip only exist on reconnect), and the stub is updated so the seam stays complete.

Two things:

- should-fix: the new wiring is exercised but never asserted. `linkdead.test.ts` drives both `socketClosed` and the same-character resume, but only against the no-op stub, so dropping either `setTrackingConnection` call would keep every test green. These two calls are the whole point of the PR, so they deserve a decisive assertion: `const spy = vi.spyOn((server as any).botDetector, 'setTrackingConnection')`, then check it fires with `false` after the drop and `true` on resume. (server/game.ts:2006 and server/game.ts:1960)

- nit: `setTrackingConnection` could use a one-line doc comment. The contract is not obvious from the signature: a freshly created context is assumed connected (createTrackingContext never pairs with a `true` call), it fires only on the linkdead/resume transitions, and `meta` is refreshed on reconnect but omitted on drop. (server/bot_detector/contract.ts:125)

Note for merge (nothing to change on your side): making this a required interface member means every BotDetector implementation behind `#bot-detector` needs the method in lockstep, not just the stub.

madmatah:
@FernandoX7  Thanks, good catch. I added a direct regression assertion in linkdead.test.ts: the test now spies on setTrackingConnection and checks that createTrackingContext does not emit an initial true call, socketClosed emits false, and the same-character resume emits true with the refreshed auth meta.

I also added the one-line contract comment on setTrackingConnection to document the connected-by-default assumption and the drop/resume asymmetry.

FernandoX7 (approved):

PR #1730, opened 2026-07-09
feat(admin): add unified moderation history

madmatah:
## Summary

Adds a new Admin > Moderation > History page for reviewing moderation audit activity in one place.

The page lists account moderation actions and blocked IP actions together, newest first, with 100 rows per page. Operators can filter the audit feed with tabs for All, My actions, and My notes. Account targets and moderator names reuse the account modal link, while IP targets link into the existing IP investigation view.

This also adds the matching `moderation.read` admin endpoint and indexes for the new chronological and moderator-scoped reads.

## Related issues

N/A

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run gate`

- Manual steps:
  -  Go to /admin with a `moderator` role

## Screenshots / recordings

image
---

### Quality

      TypeScript `strict`).

### Cross-platform

- [x] **Tested on desktop**
- [ ] **Tested on  mobile.** 

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Clean, well-scoped feature. The SQL is parameterized throughout, the `mine` and `notes` tabs scope to the caller's own account id off the token (confirmed in both the legacy `handleAdminApi` arm and the RouteDef handler), and the new indexes are all additive `CREATE INDEX IF NOT EXISTS`. Good call adding `source` as the final `ORDER BY` tiebreak: an account row and an ip row that happen to share a `created_at` and an `id` still land in a stable total order, so pagination stays consistent across pages.

Coverage is strong. `tests/admin_account_db.test.ts` pins the exact param arrays and per-tab WHERE clauses (including that `mine` scopes the ip arm rather than pruning it with `WHERE false`), `tests/server/admin.test.ts` proves the account id comes from the token and not the query string, and the page component test asserts the full action-label map (kick/kill/ip-block all labeled, never "Other action") and the tab-switch fetch URL.

Two nits:

- nit `src/admin/pages/ModerationHistoryPage.svelte:98` the filter buttons take on `role="tablist"`/`role="tab"` + `aria-selected` but stop short of the rest of the ARIA tabs contract: no `aria-controls`, the table below is not a `role="tabpanel"`, and there is no arrow-key roving between tabs. They are native `<button>`s so they stay keyboard-operable, but a screen reader announces "tab" with no panel to move into. This is the only admin page using tab roles, so either wire the full pattern or drop the roles and keep them as a plain filter button group.

- nit `src/admin/pages/ModerationHistoryPage.svelte:161` the "Note" column renders `entry.reason` for every row, so a ban or a kick shows its reason under a header labeled "Note". Minor, but "Reason" would read truer for the non-note rows.

madmatah:
@FernandoX7 Thanks, addressed!

FernandoX7 (approved):

PR #1731, opened 2026-07-09
feat(mobile): replace use button with context interact

No898:
## Summary

Mobile combat controls now use the primary action button contextually: it stays as Attack during combat, but becomes Interact or Loot when the player is near a usable target and has no live attack target. The old dedicated Interact button was removed, and Jump now occupies that ring position. NPC interactions now target the NPC before opening the dialog so the selection ring matches the conversation target.

## Related issues

N/A

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx @biomejs/biome check --write ...`
  - `npx vitest run tests/interactions.test.ts tests/mobile_action_ring_painter.test.ts tests/mobile_controls.test.ts tests/client_shell.test.ts tests/localization_coverage.test.ts tests/bank_window.test.ts`
  - `npx vitest run --config vitest.browser.config.ts tests/browser/target_size.browser.test.ts -t "mobile action-ring controls|compact-tier ring|left utility cluster"`
  - `npm run ci:changed`
  - `npx tsc --noEmit`
  - `npm run build:env`
  - `npm run build:server`
  - `npm run build`
  - pre-push QA floor: green after rebasing onto current `upstream/release/v0.24.0`
- Manual steps:
  - Captured before/after mobile HUD screenshots with the offline Playwright flow.
  - Verified the after state shows Jump in the old Interact slot, contextual Interact on the primary action, and the NPC selection ring while the dialog is open.

Note: a fresh full `npm run gate` was attempted locally before PR prep, but the full parallel Vitest step timed out in unrelated untouched suites. The timeout suites passed when rerun isolated with a longer timeout and one worker. CI should be treated as the authoritative full-suite signal for this draft.

## Screenshots / recordings

- Before HUD: https://github.com/No898/world-of-claudecraft/blob/dev-td-mobile-context-interact/docs/screenshots/mobile-context-interact-before.png
- Before ring: https://github.com/No898/world-of-claudecraft/blob/dev-td-mobile-context-interact/docs/screenshots/mobile-context-interact-before-ring.png
- After HUD: https://github.com/No898/world-of-claudecraft/blob/dev-td-mobile-context-interact/docs/screenshots/mobile-context-interact-after.png
- After ring: https://github.com/No898/world-of-claudecraft/blob/dev-td-mobile-context-interact/docs/screenshots/mobile-context-interact-after-ring.png
- After NPC selected: https://github.com/No898/world-of-claudecraft/blob/dev-td-mobile-context-interact/docs/screenshots/mobile-context-interact-after-npc-selected.png

---

### Quality

- [ ] **Tests pass.** `npm test` is green. I added or updated tests for any `sim/` or `server/` behavior I changed.
- [x] **Builds succeed.** `npm run build`, plus `npm run build:server` and `npm run build:env` if I touched those.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified on a desktop and on a phone-sized viewport in both portrait and landscape. Touch targets stay at least 40x40px and inputs at least 16px font (see `src/ui/CLAUDE.md`).
- [ ] **Accessible.** Keyboard-operable, with visible focus, sensible ARIA, and respect for `prefers-reduced-motion` (only if this change adds or edits UI).

### Localization (i18n)

### Hygiene


No898:
Need more ingame testing

jgyy (commented):
Nice change, this cleans up the mobile action ring nicely: swapping the primary button into a contextual Interact/Loot when there's something usable nearby and you're not mid-combat is a good reduction in ring clutter, and Jump taking the old hollow seat is a sensible reuse. The guard (`!p.autoAttack && !hasLiveAttackTarget`) is applied consistently in both the click handler and the paint-time icon swap in `src/ui/hud.ts`, so what you see matches what a tap does.

i18n and play.html/index.html parity both check out: no new player-visible strings, the aria-label swap reuses the existing `hud.keybinds.actions.interact` key, and both HTML entries drop `#mobile-interact` identically.

Two small notes, nothing blocking:

- nit: `findNearbyInteraction` and `hasNearbyInteraction` in `src/game/interactions.ts` duplicate the same scan logic instead of `hasNearbyInteraction` just calling `findNearbyInteraction(w) !== null`. It's called every frame for the paint check, so worth collapsing at some point, but not urgent.
- worth flagging: the NPC branch in `main.ts` now also matches `templateId === 'brother_halven_marsh'`, which looks like a genuine pre-existing bug fix (the marsh delve NPC was already handled in `hoverCursorKind`/`isAttackHoverTarget` but not in `interactKey`). Good catch, just wasn't called out in the PR description, so want to make sure it's not accidental/lost in review.

No conflicts against `release/v0.24.0`, tests look like they cover the new priority order and aria/class toggling adequately. Thanks for the contribution.

No898:
After discusion we decided to change layout and put the condition on different button, so i am closing the PR and start work at different branch

PR #1732, opened 2026-07-09
docs(design): UI/UX current-state reference and AAA redesign spec

MasterZensei:
Two design documents split out of #1719 per review feedback there (they are unrelated to the Protect Yumi minigame):

- docs/design/ui-ux-current-state.md: an inventory of the HUD as it stands (frames, windows, tokens, layout system).
- docs/design/ui-ux-redesign-spec.md: the dark-fantasy AAA redesign specification the UI work is converging toward.

Documentation only, no code changes.

### DIFF
--- a/docs/design/ui-ux-current-state.md
+++ b/docs/design/ui-ux-current-state.md
@@ -1,0 +1,331 @@
+# World of ClaudeCraft: UI / UX current-state reference
+
+Foundation document for a AAA-style HUD and UI reshape. This captures **what
+exists today**, **how it is wired**, **the design system in place**, and **the hard
+constraints any redesign must respect**. Use it as the input to a redesign spec;
+it is descriptive, not prescriptive.
+
+Generated from the live code graph (graphify: 20,428 nodes / 55,224 edges over 2,523
+files) cross-referenced with the repo's file-naming conventions and the area
+`CLAUDE.md` contracts. `src/ui/` is 217 modules; `src/game/` is 49.
+
+---
+
+## 1. How the UI is wired (the seam you must design within)
+
+One rule governs everything: **`render/` and `ui/` talk only to `IWorld`**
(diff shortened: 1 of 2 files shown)


jgyy (commented):
Looked into the conflict before writing anything else: both files here (`ui-ux-current-state.md`, `ui-ux-redesign-spec.md`) already landed on `release/v0.24.0` via #1736 (same commit, `52c2f9b70`). `ui-ux-current-state.md` is byte-identical to what's already merged. `ui-ux-redesign-spec.md` has genuinely diverged: a later commit on `release/v0.24.0` (`15cd5dd4e`) renamed the component grammar (`.btn-primary`/`.btn-danger`/`.badge` to `.btn.is-primary`/`.btn.is-danger`/`.ui-badge`) to resolve a CSS-layer collision, and this branch still has the old names in section 8.

So this PR is a stale duplicate of content that shipped elsewhere. I'd close it as superseded rather than resolve the conflict, unless there's a reason to keep this copy specifically.

One real issue if the doc stays around in any form: `ui-ux-current-state.md:10` and `:52` cite "`src/ui/` is 217 modules" and "`src/game/`, 49 modules" as facts. Those numbers have already rotted, current repo has 236 top-level files in `src/ui/` (314 including subdirectories) and 54 in `src/game/`. This is exactly what the anchor rule in `docs/qa-gate.md` warns against, cite stable paths and symbols, not counts that go stale. Every other citation in the doc (file paths, seam names like `IWorld`/`PainterHost`) is done the right way, so this looks like an isolated slip rather than a pattern.

Aside from the conflict, the content itself is solid design work: clear split between "here's what exists" and "here's the proposed direction," and the redesign spec stays presentation-layer only and names the real seams it would touch rather than hand-waving. No invented gameplay-balance numbers, the numeric values here are UI/CSS dimensions for a proposal, not combat math. No em-dashes or emojis.

PR #1735, opened 2026-07-09
feat(moderation): in-game jail with timed sentences, brawls, and a real scene

madmatah:
## Summary

Gives moderators a full in-game jail, and gives prisoners a reason to stick around.

- **Timed sentences**: `/jail "<name>" <minutes> ["reason"]` moves a player to the jail cage. The sentence length is mandatory (no indefinite form; that ambiguity invited mistakes), the optional reason lands in the audited moderation history as `"<reason> (10 minutes)"`. `/unjail "<name>"` releases early. Bare `/jail` / `/unjail` still enter and leave the moderator visitor area. Sentences persist with the character state, keep running across reconnects, and auto release to the pre-jail position with a dedicated notice.
- **Containment**: jailed players cannot enrol in instanced content (ranked arena in every format including fiesta and the yumi brackets, Vale Cup queue/ready/practice, dungeons, delves, and duels); match queues are drained the moment the jail lands. The escape enforcement bound now matches the physical cage wall, so a prisoner pressing into the bars is no longer snapped back to their cell spawn.
- **The jail brawl**: prisoners are mutually hostile (a new arm in `isHostileTo`; pets resolve to their owner, cross-heals are off) with instant full-HP cell respawn on death, no XP or loot involved. A visiting moderator can strike prisoners for flavor, but is untargetable by any prisoner action and GM invulnerable underneath.
- **The moderator gate**: a marked portal on the cage's east face teleports sessions holding `moderation.act` across the bars (pure trigger math in `src/sim/jail.ts`, applied server-side each tick). Everyone else just bumps into the blocker.

No wire or `IWorld` changes: targeting is server-resolved and casts are server-validated, exactly like duels, so both online and offline hosts share the one rule. New server emits are registered in the client matchers (EXACT plus duration rules) with every locale block filled.

## Type of change

- [x] Tests

## How was this tested?

- Commands:
  - `npm run gate` (i18n gen + freshness, malware scan, changed-files biome, full tests, tsc, builds)
  - `npx vitest run tests/moderation_game.test.ts tests/moderation_commands.test.ts tests/moderation_service.test.ts tests/server_i18n.test.ts tests/localization_fixes.test.ts` (71 passed)
  - `npx vitest run tests/architecture.test.ts tests/parity` (sim purity and rng draw-order parity, green)

- Manual steps:
  - `node scripts/jail_shot.mjs` against `npm run dev` for the visual pass (visitor aisle, guard post, cage interior, gate, latrine corner), iterated to the screenshots below.

  - Covered by server tests end to end: jail/persist/reconnect/unjail flow, timed sentence bounds and auto release, gate crossing both ways (and refusal for non-moderators and jailed sessions), brawl hostility on/off with jail state, queue draining and enrolment refusal, GM immunity to forced damage.

  In game test : 

  - (moderator) type `/jail` to visit the prison
  - (moderator) type `/jail "characterName" 10 "some reason"` to put in jail characterName for 10 minutes
  - (moderator) can enter the prison door
  - (moderator) can leave the prison with `/unjail`
  - (moderator) can cancel a prison sentence with `/unjail "characterName"`
  
  - (prisoner) can fight with other prisoners
  - (prisoner) cant queue to arenas

## Screenshots / recordings

Screenshot from 2026-07-09 23-01-48

Screenshot from 2026-07-09 23-01-39

Screenshot from 2026-07-09 23-47-44

---

### Quality

      TypeScript `strict`).

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: no new HUD chrome; the scene renders through the existing renderer paths on all viewports.

### Localization (i18n)

      All new server emits have entries in every `IN_GAME_MODERATION_MESSAGES`
      locale block (see the `jailGeneric` note in the summary).
- [x] Numbers, money, dates, units, and percents go through the i18n formatters:
      durations are built with `formatDuration` and re-localized client-side via
      `localizeServerDuration`.

### Hygiene

      `src/render/assets/manifest.generated.ts`. They're regenerated by the build
      (`npm run i18n:gen` for the i18n artifacts in this PR).


FernandoX7 (commented):
Careful, well-scoped work. For a change this size the invariants I care about most all hold up, and I verified each against the code rather than the description:

- Determinism stays clean. `src/sim/jail.ts` imports only `./types`, the `jailed` flag is server-set (`setJailed`, never offline or in the RL env), so the new `isHostileTo` brawl arm is deterministic, and every wall-clock decision (`Date.now`, sentence expiry) lives in `enforceJailStates` on the server, outside `sim.tick()`. `tests/architecture.test.ts` and `tests/parity` stay green with the shared `BUILTIN_WORLD` picking up `blockers` + `terrainEdits`.
- Auth is properly gated. `/jail` and `/unjail` require `moderation.act` via `requiredCommandPermission`, and `resolveNamedTarget` refuses self and any `isAdmin` target, so a moderator can never be jailed. That also makes `jailVisit` and `jailed` mutually exclusive, which is what keeps the `continue` in `enforceJailStates` safe rather than a latent way to skip sentence enforcement.
- i18n is fully wired. The placeholdered emits get explicit rules (`jailedFor`, `jailMovedFor`, ...) and the no-placeholder ones (`jailSentenceEnded`, `jailBlocked`, the visit notices) fall out of the auto-built EXACT reverse-map, so the S3 guard passes with nothing shipping English-only.
- The tests earn their keep. Exact gate landings in both directions, non-moderator no-move, escape snapback, queue drain plus refusal in every arena format, timed auto-release, and the brawl hostility asymmetry with the GM-invuln bounce are all asserted decisively.

Two nits, both optional:

- nit (`server/game.ts:1558`): `isInJailRoom` hardcodes `38` twice instead of the `JAIL_OUTER_HALF` you already import, so the visitor-room bound silently drifts if that constant ever changes.
- nit (`src/ui/server_i18n.ts:1943`): the bare `moderation.jailMoved` string ("A moderator has moved you to jail.") has a catalog key and a matcher rule but no emitter, since jailing always carries a duration now (`game.ts:1205` only emits the "for {duration}" form). Dead but harmless; drop it or keep it as deliberate symmetry with `jailMovedFor`.

Good call bounding the escape check to the physical cage wall so a prisoner hugging the bars is not snapped back to their cell spawn.

madmatah:
@FernandoX7 Thanks for the careful review! Both nits addressed:

isInJailRoom now uses JAIL_OUTER_HALF instead of the hardcoded 38s.
Dropped the bare moderation.jailMoved key entirely (matcher rule, locale entries, and the test sample). Since jailing always carries a duration, the symmetry was speculative, so dead code it was.

zfreiberg:
Tested this locally: checked out the branch, ran the server and client with the PR's changes, and granted a local account the moderator role via scripts/grant_admin.mjs to test the in-game commands.

Confirmed working as expected:
- `/jail` (bare) to visit the prison as a moderator
- `/unjail` (bare) to leave the visitor area
- `/jail "name" <minutes> "reason"` moves the target to the cage with the correct sentence/reason
- `/unjail "name"` releases early
- Auto-release after the sentence expires, back at the pre-jail position
- Gate crossing as a moderator

One finding: the jailed player never gets a persistent record of what happened to them. `sendChatNotice` (used for both the "moved to jail" and "released from jail" messages to the target, game.ts) sends a `type: 'error'` event, which the client only renders as a 1.6s fading toast (`showError()`, hud.ts) — it never reaches the chat log. Meanwhile the moderator's own confirmation ("Jailed X for Y") uses `sendSystemNotice`, a `type: 'log'` event, which *does* get logged. So if you're alt-tabbed or just not looking at the screen in that window, you'd have no way to find out afterward that you were jailed, by whom, or why, unlike other moderation actions. Might be worth having the target's notice also go through the log path instead of (or in addition to) the error-toast path.


Jailed for 2 mins:
image


Automatically Unjailed:
image

/Unjail works as expected:
image

FernandoX7 (commented):
Re-reviewed at 049a1e9. Both nits from my earlier review are addressed:

- `isInJailRoom` now bounds the visitor room with `JAIL_OUTER_HALF` (server/game.ts:1560), no hardcoded 38s left.
- The bare `moderation.jailMoved` key is gone everywhere: matcher rule, locale entries, and the test sample. Grepping the head finds only the `jailMovedFor` variant.

I re-ran the targeted suites at the head (moderation_game, moderation_commands, moderation_service, server_i18n, localization_fixes): 71 passed.

One item still open, since the current head predates the comment: zfreiberg's toast-vs-log finding is real, and I would tag it should-fix. `sendChatNotice` sends a `type: 'error'` event (server/game.ts:4904), which the client renders only as the 1.6-second fading toast (`showError` in hud.ts) and never writes to the chat log, while the moderator's own confirmation rides `sendSystemNotice` (`type: 'log'`) and lands durably. So the prisoner's "moved to jail for X", "released", and "sentence ended" lines vanish in under two seconds and leave no record an alt-tabbed player could read afterward, unlike other moderation actions. Routing those three target notices through the log path (instead of, or in addition to, the toast) closes it, and the swap is i18n-safe: the client's `log` case localizes via `localizeSystemText`, which falls back to the same shared `localizeServerText` matcher, so the rules you registered keep working unchanged.

madmatah:
@zfreiberg Great feedback, thanks for testing it out!

@FernandoX7 All the feedback has been addressed: the three prisoner notices (jailed for X, released, sentence ended) now ride sendSystemNotice (type log), so they land durably in the chat log like the moderator's own confirmations. Since the S3 scanner only enumerated sendChatNotice literals, I also taught it to scan sendSystemNotice as type log (with a synthetic self-test entry), so the swap does not shrink the drift guard's coverage. The moderation_game assertions now pin the log event type for those notices, so a regression back to the toast fails the test. No matcher changes needed, as you noted: the log path falls back to the same localizeServerText rules.

PR #1724, opened 2026-07-09
feat(mobile): add joystick autorun lock

No898:
## Summary

- Replaces the mobile Autorun button with a circular autorun target above the movement joystick.
- Reveals the target only near the upper joystick band, locks autorun when the drag crosses the target threshold, and keeps the locked indicator visible while autorun is active.
- Cancels autorun on any new joystick movement and clears it when the player transitions into death/ghost state.

## Related issues

N/A

## Type of change

- [x] Feature: new functionality
- [x] Bug fix
- [ ] Refactor or performance (no behavior change)
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change (please call it out in the summary)

## How was this tested?

- Commands:
  - `npx vitest run tests/death_input_reset.test.ts tests/mobile_controls.test.ts tests/input.test.ts tests/client_shell.test.ts tests/css_value_validity.test.ts tests/styles_extraction.test.ts tests/css_corpus.test.ts`
  - `npx @biomejs/biome check --write src/game/mobile_controls.ts`
  - `npx vitest run tests/mobile_controls.test.ts tests/input.test.ts`
  - `npm run check:ts`
  - `git diff --check`
  - `npm run gate`
  - Pre-push QA floor: green
  - Playwright smoke against `http://127.0.0.1:5175/` for desktop, mobile portrait, and mobile landscape
- Manual steps:
  - Verified the mobile joystick autorun flow on the local dev server at `http://192.168.1.5:5175/`.
  - Confirmed the old Autorun button is removed and the locked autorun indicator remains visible while active.

## Screenshots / recordings

Committed under `docs/screenshots/mobile-autorun/`:

- Before: `docs/screenshots/mobile-autorun/before-mobile-autorun-button.png`
- After, autorun target visible: `docs/screenshots/mobile-autorun/after-autorun-visible.png`
- After, joystick over autorun target: `docs/screenshots/mobile-autorun/after-autorun-hover.png`
- After, autorun locked: `docs/screenshots/mobile-autorun/after-autorun-locked.png`

---

### Quality

- [x] **Tests pass.** `npm test` is green. I added or updated tests for any
      `sim/` or `server/` behavior I changed.
- [x] **Typecheck passes.** `npx tsc --noEmit` is clean (the project is
      TypeScript `strict`).
- [x] **Builds succeed.** `npm run build`, plus `npm run build:server` and
      `npm run build:env` if I touched those.

### Cross-platform

[...]

PR #1736, opened 2026-07-10
feat(ui): production-AAA interface overhaul (tokens, window grammar, drag/resize, Esc menu Codex)

MasterZensei:
## Summary

Production-AAA interface overhaul for the game client, built phase by phase on a
single foundation so every window shares one visual grammar and one behavior
model. This PR is a WORK IN PROGRESS being updated continuously; progress is
tracked in the comments below.

What is in so far:

- **AAA design-token foundation**: new token groups in `src/styles/tokens.css`
  (panel depths, window chrome, focus ring, rarity, urgency, notification,
  motion, density/type scale) wired through `src/ui/theme.ts` across all four
  theme presets, WCAG-guarded against both panel and panel-edge surfaces.
- **A shared window grammar** (`components.css` + the `window_frame` builder:
  pure view-core + thin DOM consumer): titlebar with display-type title, icon
  close, tab rail, sticky footer, item cells with rarity borders, buttons,
  fields, empty/loading/error states. Shell-layer collision policy documented
  in the spec (`.btn.is-primary` / `.ui-badge` forms).
- **Every grammar window is draggable, resizable, and viewport-clamped**:
  `.window-titlebar` is a recognized drag handle (`window_drag_handle.ts`),
  windows fit the viewport with internal body scroll, and the mobile 50/50
  dock refuses drags (`body.mobile-touch`).
- **Vendor window pilot**: the pattern-proving restyle, including the shared
  `#vendor-window` root isolation for the heroic vendor (builder-owned inner
  mount, sequence-tested).
- **The Esc settings menu, rebuilt** ("The Warden's Codex", design spec in
  `docs/design/esc-menu-redesign-spec.md`): a persistent category rail with an
  Overview landing page, global search across all settings, preset-then-detail
  graphics, per-category scoped resets, full keyboard navigation (roving rail,
  per-control value keys), and the game's first controller menu navigation
  (LB/RB category jump, D-pad focus, A/B/X/Y verbs, RT/LT paging, footer
  button legend), gated behind a new `FocusManager.hasActiveTrap()` so pad
  input never double-fires into the world.
- **Design docs**: current-state reference, the AAA redesign spec, the Esc
  menu spec (v1.1 with mobile back-stack + sprint tier flags), and scope
  matrices with required/not-required flags per setting and HUD variant.

Remaining phases (tracked in comments as they land): P4 rebind UX + conflict
surfacing + default-bind modernization (Q/E reserved for spells, Shift+1-0
secondary-bar page, meters to Z); P5 dedicated mobile back-stack settings
shell + mobile optimizations; the all-windows AAA pass (every remaining HUD
window adopts the frame grammar with drag/resize); QA matrix (themes, fx
tiers, scales) and final gate.

## Type of change

- [x] Documentation
- [x] Tests

## How was this tested?

- Commands: `npm run gate` (full CI-equivalent; green except the known
  Windows-local `tests/server/new_endpoint.test.ts` golden failure, a POSIX
  spawn issue unrelated to this diff), targeted `npx vitest run` suites per
  phase (options/vendor/window-frame/architecture/theme/css/perf-budget),
  `npx tsc --noEmit`.
- Manual steps: live browser verification per phase against the dev server
  (offline boot): vendor open/close/buy/drag/resize/fit, heroic takeover
  sequence, Esc menu full keyboard walk of every category and control kind,
  settings dispatch persistence, both search scopes, language switch with the
  menu open, mobile-touch dock behavior.
- Process: each phase was implemented and then reviewed by an independent
  pass (spec compliance + code quality), with adversarial multi-agent
  verification on the riskiest changes (three state-transition regressions
  found and fixed on the window-behavior work; six independent verifications
  confirmed the fixes).

## Screenshots / recordings

Vendor pilot, before / after (desktop):

before
after

Mobile 50/50 dock (genuine `body.mobile-touch` capture):

mobile

The Esc menu screenshots land with the next push (they are captured; the
settings-menu drag/resize enablement changes them, so they will be committed
after that change).

---

### Quality

- [x] **Tests pass.** Full gate green modulo the documented Windows-local
      `new_endpoint` golden failure (pre-existing on the dev machine; no
      `server/` files are touched by this PR).
- [x] **Typecheck passes.**
- [x] **Builds succeed.** (`npm run build` runs inside the gate.)

### Cross-platform

- [ ] **Tested on desktop and mobile.** Desktop verified per phase; the
      dedicated mobile pass is P5 (in progress; interim mobile state is
      usable and documented).
- [x] **Accessible.** Keyboard-operable with visible token focus rings, focus
      trap + return, ARIA roles (tablist/radiogroup/switch), forced-colors
      support, reduced-motion respected.

### Localization (i18n)

- [ ] **New player-visible strings are translated into every supported
      locale.** English keys added to the correct catalog domains; wordy keys
      carry their five non-Latin M16 fills; remaining locales are
      English-filled `pending` per the PR-tier policy (maintainer fills at
      release). Two non-wordy legend keys are pending, PR-tier legal.
- [x] Numbers, money, dates through the i18n formatters.
- [x] N/A for sim/server text: this PR emits none.

### Hygiene

- [x] No hand-edited generated files (i18n artifacts regenerated via
      `npm run i18n:gen`).

## Release-notes callout: default keybind modernization (approved)

This PR changes the FRESH-PROFILE keyboard defaults (approved by the project
owner on live review; players who ever rebound anything keep their exact stored
layout, verified against the save/load path):

- Q and E are now action-bar slots 10 and 11 (with Minus and Equal as
  secondaries), not strafe. Strafe Left/Right ship UNBOUND by default and stay
  fully rebindable in the Esc menu; A/D turning plus mouse-steer covers lateral
  movement out of the box.
- Shift+1..0 page the secondary action bar.
- Z toggles the damage meters.

Never-customized profiles get the new layout silently; the /wiki controls page
and README document the new defaults, and intentionally-unbound defaults are
excluded from the keybind-conflict warning aggregate so fresh profiles do not
ship in a warning state.


MasterZensei:
**Progress 1/planned-many: foundation through Esc menu navigation is on the branch.**

Landed and independently reviewed (each phase: implement, spec+quality review, fixes, re-verify):
- Design tokens + theme knobs (4 presets, contrast-guarded)
- Shared window grammar + window_frame builder (pure core + thin consumer)
- Drag / resize / viewport-fit for grammar windows; vendor window pilot restyle (incl. shared-root isolation for the heroic vendor, sequence-tested); 3 state-transition regressions found by adversarial multi-agent verification and fixed (state-driven display, bags cascade exemption, mobile-dock drag guard)
- Esc menu rebuilt as The Warden's Codex: Overview landing (default entry), global search, category rail, preset-then-detail graphics, scoped resets, full keyboard navigation, first-ever controller menu navigation (LB/RB/D-pad/A/B/X/Y/RT/LT + footer legend), gated by FocusManager.hasActiveTrap()

In flight / next (comments will follow each):
1. P3 navigation review closing now
2. P4: rebind capture UX (three-exit, eviction announcements), conflict surfacing to the rail, default-bind modernization (Q/E reserved for spells, Shift+1-0 as the secondary-bar page, Damage Meters H->Z resolving a pre-existing default collision), dropdown typeahead
3. Settings menu joins the universal drag/resize system (maintainer direction supersedes the spec's fixed-window ruling)
4. All-windows AAA pass: every remaining HUD window adopts the frame grammar with drag/resize on desktop
5. P5: dedicated mobile back-stack settings shell + mobile optimizations
6. QA matrix (4 themes x fx tiers x ui scales), full gate, final screenshots

Known-local caveat: the dev machine's `npm run gate` always reds on `tests/server/new_endpoint.test.ts` (Windows POSIX-spawn issue, pre-existing, no server files touched here); every other suite is green.

MasterZensei:
**Progress 2: P3 navigation review passed (0 Critical / 0 Important / 6 minor follow-ups).**

Highlights verified by the independent review: vertical roving rail auto-activates with the rail node surviving detail re-renders; controller mode consumes all pad edges while a focus trap is active (nothing double-fires into the world); every verb wired (LB/RB categories, D-pad focus, A/B/X/Y, RT/LT paging) behind a testable intent seam; footer button legend appears only while a pad is connected; the controller cursor is model-driven (.is-active-row), not :focus-visible-dependent.

Now running in parallel:
- **P4** (main tree): rebind capture UX (three-exit), eviction announcements, conflict dot to the rail + Overview alert, search go-to, dropdown typeahead, the default-bind modernization (Q/E to spells, Shift+1-0 secondary page, Meters H->Z), and the settings window joining drag/resize per maintainer direction.
- **All-windows AAA batch 1** (isolated worktree): bags + bank adopt the frame grammar with drag/resize, docking preserved.
- **All-windows AAA batch 2** (isolated worktree): questlog, mailbox, social, spellbook.

Merges are sequential with an independent review per batch. Next push lands P4 + the first merged batches.

MasterZensei:
**Progress 3: the all-windows AAA pass is fully in motion; first two batches reviewed and approved.**

Approved (independent review each, 0 Critical / 0 Important):
- **Batch 1, bags + bank**: frame grammar with drag/resize, item-cell rarity grammar, and every docking flow preserved (vendor pairing, bank 50/50, handoffs) pinned by tests. Full suite green in its worktree (11,491 tests).
- **Batch 3, leaderboard + calendar + daily rewards + heroic vendor + market**: includes the heroic vendor's migration to the shared frame builder with the two-tenant shared-root invariant pinned in both handoff directions, and a deliberately minimal CSS-only polish for the market (the original drag/resize reference window; zero behavior risk taken).

A plumbing-level dry-run (git merge-tree) confirms the two batches merge with zero conflicts; the per-batch CSS anchor discipline is doing its job.

Still building in parallel: P4 (rebind UX, conflict surfacing, default-bind modernization, settings drag/resize) in the main tree; batch 2 (questlog, mailbox, social, spellbook); batch 4 (character, talents, crafting); batch 5 (arena, vale cup, town focus, rite, corpse harvest, loot settings, chat config). That covers all 24 HUD windows.

Merge train starts the moment P4 lands: sequential merges with a combined test suite between each, then the branch push and per-merge comments here.

MasterZensei:
**Progress 4: P4 landed + first two window batches merged. The branch now carries the complete settings experience and 11 restyled windows.**

Pushed in this update:
- **P4 (7 commits)**: rebind capture with three guaranteed exits and spoken eviction announcements ("Bound {key} to {action}; removed from {evicted}"), conflict surfacing end-to-end (in-pane chips, a warning dot on the category rail that clears live as you fix binds, an Overview alert), global search jump-to-setting with a steady landing highlight, first-letter typeahead on long dropdowns, and the **default-bind modernization**: Q/E now default to action-bar spell slots, Shift+1-0 pages the secondary bar out of the box, Damage Meters moves H->Z resolving a pre-existing default collision with Target Friendly. Existing player profiles are never migrated (test-pinned). **The settings window itself now drags and resizes** like every other window, with its category rail collapsing responsively via a container query.
- **Batch 1 merged**: bags + bank on the AAA frame (all dock flows preserved and pinned).
- **Batch 3 merged**: leaderboard, calendar, daily rewards, the heroic vendor's migration to the shared frame builder, and a conservative market chrome polish.

Combined suite over the merged tree: 401 passed, typecheck clean. In flight: P4 under independent review, batch 2 (questlog/mailbox/social/spellbook) in its fix round re-review, batches 4 (char/talents/crafting) and 5 (arena/vale cup/town focus/rite/corpse harvest/loot settings/chat config) implementing. Then P5 mobile, QA matrix, full gate.

MasterZensei:
**Progress 5: four window batches merged. 15 windows now on the AAA frame grammar with drag/resize; the last batch is in review.**

Merged and pushed since the last update (each independently reviewed, 0 Critical / 0 Important):
- **Batch 2**: quest log, mailbox, social (tab rail), spellbook, keeping the sacred flows intact (spellbook drag-to-actionbar, quest tracking, social panel visibility), with guild-name injection and mailbox item-cell attachments hardened in a fix round.
- **Batch 4**: character (paperdoll + stat tooltips), talents (staged-edit + loadouts preserved), crafting.

Combined suite over the merged tree is green and typecheck is clean.

An honest reconciliation of the "every window" goal against the real architecture: two things I had listed turned out not to be standalone windows. "Chat config" is a context-menu popup, not a window; corpse harvest is a picker section inside the HUD-owned loot popup. Neither can take a draggable frame without out-of-scope surgery, so both are preserved and behavior-pinned rather than reshaped. Lockpick stays a fixed-size modal by design (it is a timed minigame). Every genuine standalone window is migrated.

Still to land: batch 5 (arena, vale cup, town focus, rite, loot settings) finishing review now, then P5 (the dedicated mobile settings back-stack shell + a mobile polish pass consolidating touch floors and a couple of cosmetic seams the batch reviews flagged), then the QA matrix across themes and scales, a final whole-branch review, the full gate, and a fresh screenshot set. Marking the PR ready-for-review at that point.

MasterZensei:
**Progress 6: all five window batches merged. Every standalone HUD window is now on the AAA frame with drag/resize.**

Batch 5 (arena, vale cup, town focus, rite, loot settings) reviewed clean and merged. The full test suite over the complete merged tree: 11,661 passed, with only sim suites that flake under parallel contention on this machine (verified 35/35 green in isolation) and the documented Windows-local new_endpoint failure. Typecheck clean.

Window inventory, final: 15 windows migrated across the batches plus the vendor pilot and the Esc settings menu = every genuine standalone window, all draggable/resizable with the shared frame grammar. The three that are intentionally not framed, with reasons: lockpick (fixed-size timed minigame), chat channel config (a context-menu popup, not a window), corpse harvest (a section inside the HUD-owned loot popup). All three are behavior-preserved and test-pinned.

Now starting P5: the dedicated mobile settings back-stack shell (level-0 landing with global search and category list, pushed category pages, controller/back popping) plus a mobile-polish pass that consolidates touch floors and clears a handful of cosmetic seams the batch reviews flagged (a title-truncation rule, a vendor titlebar alignment, the resize-grip visual affordance). After P5: the QA matrix across themes/scales, a final whole-branch review, the full gate, and fresh screenshots, then I flip this PR to ready-for-review.

MasterZensei:
**Progress 7: implementation complete. Desktop + mobile settings, all windows framed, pushed and gate-verified.**

P5 landed and was reviewed clean: the dedicated mobile settings back-stack shell (level-0 landing with global search, quick actions, and the stacked category list; pushed category pages with back navigation; body.mobile-touch-only, desktop byte-identical), the visible resize-grip affordance on every framed window, and the accumulated mobile-touch polish (20px slider thumbs, 40x40 switch targets, titlebar-seam consolidation, the char-title truncation fix).

Verification posture on the full merged tree:
- Full test suite green except the documented Windows-local new_endpoint golden test (a POSIX-spawn artifact of this dev machine, not the diff). tsc clean. All five build entries build; backdrop-filter twins survive Lightning minification.
- A final whole-branch adversarial review (six dimensions: cross-window consistency, the settings menu, accessibility, i18n/copy, perf/architecture, and merge-integration seams) is running now; I will post its outcome.

Screenshots added to the PR: the settings menu across all four themes (classic gold, midnight, parchment, high contrast) and six mobile captures (settings landing + pushed category, and framed windows on a 844x390 touch viewport). The theme shots also show the keybind changes live: Q and E now sit on the action bar, and the Overview surfaces the strafe-unbound alert with a Key Bindings jump.

Remaining: fold in anything the final review confirms, then flip this PR from draft to ready-for-review.

MasterZensei:
**Progress 8 (final): whole-branch adversarial review complete, all confirmed findings fixed. Marking ready for review.**

The final whole-branch review fanned six independent reviewers across cross-window consistency, the settings menu, accessibility, i18n/copy, performance/architecture, and merge-integration seams, then had two skeptics try to refute each Critical/Important finding. It surfaced four confirmed Important defects that the per-phase reviews structurally could not see (all cross-cutting), now fixed:

1. **Resize gap** (I1): 8 root-scroll windows could show the 3D world through a transparent gap when resized taller than their content (char/talents/crafting had been patched, the siblings had not). Fixed with ONE shared rule so the fill can never diverge again; live-verified (0px gap after resize, no double-scrollbar regression on flex-model windows).
2. **Controller crash** (I2): LB/RB on the mobile settings shell dereferenced the absent desktop rail. Guarded and test-pinned.
3. **Focus trap** (I3): Crafting and Town Focus dialogs skipped the trap/focus-first/return-to-opener invariant every other framed window follows. Wired into the shared FocusManager.
4. **Forced-colors** (I4): the settings switch had no non-color on/off cue under Windows High Contrast. Added, matching the segmented-control pattern.

Plus four correctness minors: global search now reaches the language and theme rows and no longer surfaces preset-gated advanced graphics; the dead legacy vendor CSS family (shipped to every player) is removed with its stale comment corrected.

One documented intentional exception: the World Market window keeps its own titlebar class names (it is the original drag/resize reference the whole grammar was modeled on); it drags, resizes, and is visually consistent, but its internal .panel-title/.x-btn structure differs from the shared .window-frame markup. Flagging for maintainer preference rather than churning the reference window at the finish line.

Verification: affected suites + tsc + all five builds green; pre-push floor green; full suite green except the pre-existing Windows-local new_endpoint golden test. This PR is ready for review.

MasterZensei:
**Follow-up from live testing: addressing three items.**

Thanks for the hands-on feedback. Diagnosed and now in progress:

1. **Field of View does nothing** - confirmed at the code level: the setting writes the camera FOV, but the renderer's per-frame camera update forces it back to a fixed base value, so the change never persists. Fixing that, and running a full functional audit of every setting to catch any others that are wired to the menu but have no runtime effect (the audit table goes in the PR).

2. **Modifier keys in rebinding** - the bind system already supports combos like Shift+key, Ctrl+key, Alt+key (they store and label correctly); it just was not communicated. Adding a helper note in the Key Bindings pane so players know double/modifier keys are bindable.

3. **Traders/shops should work like the World Market** - the vendor already drags and resizes (verified), but it opens as a flat list docked with Bags rather than the Market's standalone tabbed window, which is the real gap. Redesigning the NPC vendor (and the heroic vendor) to the Market's tabbed chrome: Browse to buy, a Sell tab (replacing the need to dock Bags), and a Buyback tab in place of the Market's Collect, all draggable/resizable with the visible bottom-right resize grip. The Market itself stays as the reference window.

Both workstreams are running in parallel and will be reviewed and pushed to this branch.

MasterZensei:
**Live-feedback round pushed: FOV, settings audit, keybind hint, mobile sidebar, and the tabbed shops are all in and reviewed.**

- **Field of View: fixed.** The renderer was overwriting the setting every frame; the user FOV is now threaded into the camera base and persists. Live-verified holding across frames.
- **Full settings audit: clean.** Six auditors traced all 67 settings to their runtime sinks with adversarial skeptic-verification. Zero dead beyond FOV. 62 verified live, 3 reload-gated (already flagged with a reload alert in-menu), 1 internal. So FOV was the only dead one.
- **Modifier keys:** the Key Bindings pane now states that Shift/Ctrl/Alt + key combos are bindable (they already worked, just weren't communicated).
- **Mobile settings menu:** on wide/landscape it now uses the desktop-style category sidebar with a width-capped detail pane, and both mobile modes get a solid opaque background (no more see-through). Narrow portrait keeps the stacked list.
- **Traders/shops now work like the World Market:** the NPC vendor is a standalone Browse / Sell / Buyback tabbed window (Buyback in place of the Market's Collect), draggable and resizable with the grip; the Sell tab replaces the old bags-dock sell flow and routes through the same sim command, with a confirm guard added when selling rolled/instanced gear (buyback only restores base copies). The heroic vendor stayed marks-only but gained the resize grip. Screenshots attached to the PR.

Each item was independently reviewed before merge. One item remains from the latest feedback: the Delve Board window resizes past the point where its content reflows and its actions stay visible; I am taking that on next as a shared resize-floor + content-reflow fix across all windows.

MasterZensei:
## Live-feedback round 2: mobile polish, vendor sizing, settings scroll

Pushed a batch addressing the latest round of hands-on testing feedback. All changes verified live against the worktree dev server (booted offline, real rendered geometry), plus tsc, the build (backdrop twins intact), biome on changed files, the copy scan, and the full vitest suite green (the only red is the pre-existing Windows-only `new_endpoint` tsc-spawn issue; fiesta/mail/threat pass in isolation, they were parallel-contention flakes).

### Mobile
- **Action-button scale floored at 25%.** The size slider now spans 25% to 200% (was 0% to 200%). A 0 scale collapsed the entire on-screen button cluster to nothing and soft-locked a touch player with no way to reopen the menu, and it persisted. The load clamp lifts any previously stored sub-25% value back up on next launch, recovering an already-stuck player.
- **Collapsible menu-button cluster.** Chat / Social / Quests / Settings / More now hide behind an always-visible arrow chip, default collapsed, persisted per player. The consumables quick-bar and the combat action ring are untouched and stay visible. Verified: collapsed shows only the arrow, expanded shows all five, consumables and ring unaffected.
- **40px touch floor held.** The expanded menu buttons sit in the top-left band, which carries the 0.85 chrome-scale; the landscape gap left each button at ~38px, under the 40px floor. Tightened the gap so each renders at ~40px (measured 40px at 844x390 / 900x430 / 1024x600) while keeping the band width and the consumables dock fixed. The target-size browser test now measures the real nested/expanded grid instead of the base button size, so it actually protects this.
- **Tap feedback.** Replaced the mobile browser's default blue tap-highlight rectangle (an unstyled square over the round icons) with a round, white, low-opacity press overlay that hugs each icon and only surfaces while pressed.

### Desktop windows
- **Vendor height capped 20% shorter.** Trader Wilkes (and the other framed vendors) ran down over the player HP bar. The window is top-anchored, so the trim comes off the bottom: top stays put, and the window clears the HUD bar (bottom moved from ~831px to ~678px on a 900px viewport). The mobile 50/50 dock is unaffected.
- **Settings detail-pane over-scroll fixed.** The two-pane settings body had an unbounded grid row, so a long detail pane grew past the frame and its scrollbar spanned that overgrown height, scrolling the user out of the window. Bound the row and let the rail and detail scroll internally: verified the detail is contained in the frame and scrolls within its own bounds.

Screenshots for this round are under `docs/screenshots/livefix-*.png` (vendor height, settings scroll, mobile menu collapsed vs expanded).

Also in this push from the prior sub-batch: the Delve Board content now reflows on resize (no empty gap when grown, actions stay reachable when shrunk) and the shared window resize floor was raised so no window can be shrunk below its usable chrome.

Rubsey (commented):
Reviewed the whole branch at db3a5eff2 against release/v0.24.0 (merge-base 8602439). This is a large, well-run piece of work and most of the load-bearing claims held up under independent verification:

- The frame builder is clean where it matters most: `window_frame_view.ts` is genuinely DOM-free, every interpolation in the builder's innerHTML goes through `esc()`, and hostile-name escaping is pinned by tests (friend names, guild head, roster).
- The two-tenant `#vendor-window` isolation is real: the shared root stays byte-pristine under open/close and both handoff directions, and the tests pin it.
- Dispatch parity for the four settings primitives (slider/toggle/boolToggle/choice) is byte-identical to the old painter, including the uiScale commitOnChange split.
- Pad edge consumption while a trap is active is airtight, and outside a trap the menu core emits and consumes nothing.
- Behavior preservation across the 15 converted windows held up under line-by-line diffing: spellbook drag-to-actionbar, talents staged-edit and loadouts, arena/vale-cup/calendar/loot-settings hooks all fire the same handlers with the same payloads.
- `tests/architecture.test.ts` changes are purely additive and all six newly registered cores are actually pure. The ~60 new `hudChrome.options.*` keys are in the right domain and all have consumers. The added-lines copy scan is clean.

Verified at head: `npx tsc --noEmit` clean, and all 49 test files this PR touches pass (910 tests). One test outside that set fails, which is the first blocker. Note the fork branch runs no CI, so none of this is visible on GitHub checks.

## Blocking

1. **The committed `i18n.resolved.sha256` baseline is stale; the equivalence gate is red at head.** `npx vitest run tests/i18n_resolved_equivalence.test.ts` fails (2 of 5), and `node scripts/i18n_resolved_hash.mjs --check` reports hash 8d3d8583... vs baseline 5041eb8b.... The generated tables themselves are fresh (a clean regen is byte-identical), so this is just a missing re-baseline after the last catalog edit: run `node scripts/i18n_resolved_hash.mjs --write` and commit. As shipped, the full suite and the CI PR gate fail.

2. **Connecting or disconnecting a gamepad while the mobile settings shell is open throws.** `refreshControllerLabels()` calls `this.renderRail()` unconditionally when open (src/ui/options_window.ts:535), but the mobile shell never mounts an `.opt-rail`, so `railEl()` is a null cast and `replaceChildren()` throws a TypeError. Trigger chain: `gamepadconnected/disconnected` -> main.ts:1327 `onConnectionChange` -> hud -> `refreshControllerLabels()`. The I2 fix guarded exactly this class of null-rail deref for LB/RB cycling but missed this path (and `renderDetail()` in the same method has the same problem when the Controller page is on the mobile stack). Branch to the mobile re-render when `mobileActive()`.

## Should fix

**Live language switch**

3. Frame chrome never re-resolves `t()`: every `ensureFrame` reuse path returns the cached frame, so titles, tab labels, and close aria-labels stay in the stamped language for the whole session, including across close/reopen (frames survive `display:none`). Concretely: the social tab rail, char/talents titlebar text, vale cup (its `relocalize()` only rebuilds the body), and every converted window's close aria-label. Pre-PR every open rebuilt the chrome with fresh `t()`, and `translatePage()` cannot help since the chrome carries no `data-i18n`. One fix at the seam: re-stamp title/close/tab labels on the reuse path or on `woc:languagechange`.

4. On desktop, the language-switch success path re-renders only the detail pane and only when `activeCategory === 'interface'` (src/ui/options_window.ts:1566). The rail, footer, and search strip stay in the old language, and when the switch happens via the Overview pin or a search result, nothing re-renders at all.

**Focus and keyboard**

5. `focusFirst()` skips `[data-close]` but the frame close button carries only `data-window-close` (src/ui/focus_manager.ts:141 vs src/ui/window_frame.ts:60), so converted windows now land focus on the dismiss X: quest log on open and after every quest-row click (render ends in `focusFirstInteractive`), loot settings (previously the `#ls-method` select), and the new town focus trap. This contradicts focus_manager's own documented contract; extend the predicate to `[data-window-close]` or add `data-close` to the frame button.

6. A segmented radiogroup whose stored value matches no option has zero tab stops (src/ui/options_window.ts:1438): every radio gets tabIndex -1, so the trap's Tab cycle and the controller's `detailFocusables()` skip the row entirely and the value cannot be corrected from the UI (real cases: `clickToMoveButton` stored 1 with options {0,2}; `graphicsPreset` stored 4/5 under the native-shell cap). Same pattern in `railTab` during the bugreport sub-view. Fall back to tabIndex 0 on the first radio when nothing is selected.

7. Opening the mobile shell never seeds focus into the trap (src/ui/options_window.ts:505): the preferred `.opt-tab.is-active` does not exist on the shell, and the fallback picks the frame titlebar close, which is `display:none` under `body.mobile-touch` (hud.mobile.css:4156), so `.focus()` is a no-op. The pad path then relies on the focusLost fallback, and B can close everything instead of popping the stack. Pass a mobile preferred selector or mark the shell close buttons with `data-close`.

**Controller**

8. Menu mode plus the minimal fallback demotes every non-options trapped window for pad users (src/main.ts:1324, src/ui/hud.ts:14142). `isMenuMode` is true for ANY trap, which suppresses the virtual cursor, while the generic fallback handles only back and activate, so multi-control trapped dialogs (the quest reward choice, crafting and town focus, both newly trapped in this same PR, delve board, loot settings) lose pad focus movement entirely; at the merge base they were fully operable via the virtual cursor. Route rowPrev/rowNext to a generic focus step in the fallback, or scope menu mode to the options trap.

9. On the mobile shell, D-pad row focus and LT/RT paging silently no-op because the verb handlers query `.opt-detail`, which only exists on desktop (src/ui/options_window.ts:1031, 1132), while the legend above Done still advertises the full verb set. Include `.opt-mshell-content` in the verb paths or trim the mobile legend.

**Settings correctness**

10. Overview pins and global search results bypass the native-shell graphicsPreset cap and env gating (src/ui/options_window.ts:1756, 2085): the cap only lives in `renderCategoryDetail` (1251), so a native-shell user can dispatch `graphicsPreset` 4 or 5 (impossible pre-PR) and sees the `nativeShellHidden` interfaceMode pin.

11. Chat timestamps (show + 12h/24h) no longer have any surface (src/ui/options_ia.ts:389): the omission comment says deferred, but hud.ts still applies the stored values per chat line, so a player who had timestamps on can never turn them off. Either add the rows back or get an explicit sign-off on shipping the regression.

12. The rail changed-count badges, the category-head scoped Reset button, and the Overview changed summary only recompute on a full render (src/ui/options_window.ts:1382): flip a switch and the badge and Reset button do not appear until you leave and re-enter the category. A cheap in-place badge sync would fix it without focus churn.

**Vendor**

13. The scroll save/restore targets the root while the desktop flex model moved scrolling into `.window-body` (src/ui/vendor_window.ts:190): on desktop the saved `scrollTop` is always 0, so the goods/buyback list snaps to the top on every buy (each purchase repaints via buyAndRefresh). Bags and bank got this right (`.bag-grid` / `.bank-scroll`).

14. Un-exempting the vendor from the cascade makes it drift 28px down-right on every reopen while bags is open (src/ui/hud.ts:1980): the cascade offsets from the window's own persisted inline rect, never sets `windowMoved`, and `closeVendor` leaves bags open, so each vendor visit compounds the drift until the viewport clamp corners it. The baked inline geometry also survives a live desktop-to-touch Interface Mode flip and beats the 50/50 dock (the exact staleness class 26458e321 fixed for `display`). Mark the cascade placement or skip it when the only other visible window is the vendor's own bags companion.

**CSS**

15. The legacy same-layer `.btn` skin leaks into every grammar button (components.css:658 vs `.window-frame .btn` at 8972): `margin-top: 10px` / `margin-right: 8px` inside the new flex footers, a permanent 1px black outline, text-shadow, the legacy hover brightness/glow, `.btn:active { transform: translateY(1px) }` (the grammar comment at 8969 explicitly forbids transforms), and the legacy `!important` 3px focus ring beats the new token focus ring regardless of specificity. The collision-policy comment assumes co-declared properties; these are not co-declared. Reset them on `.window-frame .btn` (and check `.bar`'s `margin-top: 3px` leaking into `.window-frame .bar` the same way).

16. A live mobile-touch flip with the menu open leaves mismatched chrome: hud.mobile.css:4156 hides the frame titlebar and footer under `body.mobile-touch` while the desktop two-pane DOM stays mounted (options_window.ts has no flip listener), so rotating across the touch threshold removes both close affordances until a rail tap rebuilds the shell; the reverse flip shows double chrome with an empty footer. Same disease the vendor work solved with state-driven display.

17. The touch slider misses the 40x40 tap floor this PR itself enforces for the switch (hud.mobile.css:4332): the range input's hit box stays about thumb-height (~20px). `min-height: var(--touch-min)` on the input meets the documented floor.

18. `--color-notify-success` never theme-varies (tokens.css:285): it aliases the static `#7fdc4f`, so the `.ui-badge.badge-success` border computes to roughly 1.3:1 on the parchment panel, exactly the failure mode theme.ts repaired for warn/critical with the double guard. Add a guarded `notifySuccess` entry to `themeCssVars`.

**Repo tooling left behind by the markup migration**

19. About 15 puppeteer E2E/screenshot scripts still drive removed markup and will time out or throw: scripts/localization_e2e.mjs queries `.vendor-item` (72/310/491/835), `#spellbook .panel-title span` (788), and `#social-window .soc-tab` (888); also snapshot_weapon_views.mjs, bag_filter_shot.mjs, heroic_vendor_shot.mjs, market_visual.mjs, and others on `.bag-item`/`.bag-chip`. `git diff --name-only` for scripts/ is empty in this PR.

20. tests/browser/a11y.browser.test.ts still asserts `role=dialog` + `aria-labelledby` on the window ROOT and `button[data-close]` for five converted windows (lines 157, 217, 241, 324, 508), but the PR moved the dialog identity to the inner frame mount and renamed the close hook. The opt-in browser a11y suite is red at head, which will mask real regressions later.

21. The public /wiki guide still documents Q/E as strafe (src/guide/pages/controls.ts:24) and never mentions Z for meters. This hand-written page is not freshness-gated, so it needs updating in the same change that redefined the defaults.

**Keybind defaults (design flag, wants an explicit maintainer decision)**

22. Fresh and reset-to-defaults profiles now ship permanently in the warning state: unbound strafe counts toward `keyboardWarning` (src/ui/keybind_conflicts.ts:87, 129), lighting the rail dot, the Overview alert, and the banner out of the box. That trains players to ignore the conflict system. Exempt intentionally-unbound defaults from the aggregate, or park strafe on real secondary keys.

23. The modernization itself deserves a deliberate sign-off. The persistence side is safe (I verified the save/load path: any player who ever rebound anything has a full stored blob and keeps their exact layout), but never-customized players silently get Q/E as action buttons and no keyboard strafe, and Q/E strafe is the classic-era default this project usually preserves. At minimum it needs a release-notes callout plus the guide fix above. Also, the comment at src/game/keybinds.ts:44 ("The user asked for Shift+1-5; ...") is a conversation artifact that should not ship, and src/game/input.ts:1 still documents "Q/E strafe" in its header.

**Code and test debt**

24. src/ui/options_view.ts:235-406 keeps ~200 lines of dead legacy panel builders (zero consumers after the P2 rewire), still green-pinned by tests/options_view.test.ts as if live. A future contributor editing `buildGraphicsControls` would see passing tests and believe they changed the game. Delete both halves or mark them legacy with a removal task.

25. tests/options_window.test.ts is entirely source-text greps (line 10: `readFileSync` of the painter), including the load-bearing dispatch-parity block: a string pin passes even if the code containing it becomes unreachable. The vendor/bags/talents windows in this same PR show the executed-jsdom pattern worth copying here; options_view.test.ts pins the value coercions properly, it is specifically the painter wiring that is grep-only.

26. window_frame_mount.ts ships alongside roughly a dozen near-identical private `ensureFrame` copies added by this same PR (16 files carry the `:scope > .window-frame` marker; only leaderboard/calendar/daily-rewards adopted the helper). The copies are exactly the single-tenant case the helper was written for. Consolidate in this PR or track the follow-up as an issue rather than only the LEDGER comment.

## Nits

- window_frame.ts:60: `model.close.id` is derived and test-pinned but never painted into the DOM.
- window_frame.ts:73: inactive tabs keep `aria-controls` pointing at nonexistent panel ids (the single renamed body), and the tabpanel lacks `aria-labelledby`; axe will flag both.
- window_frame_mount.ts:40: the reuse path silently ignores `descriptor` and `deps`; worth one sentence in the contract.
- gamepad.ts:144: `onDisconnect()` does not reset `prevStickX` (`stop()` does), so one stale stick crossing can fire across a reconnect.
- options_window.ts:1519: scoped and per-row resets that change reload-required graphics keys never set `reloadPending`.
- options_window.ts:1938: the footer Report a Bug button duplicates `openBugReport()` inline instead of calling it.
- options_window.ts:2196: the unbound-keybind banner joins localized fragments with a hardcoded '; ' and falls back to the raw action id as a display name.
- i18n.catalog/hud_chrome.ts:658: stale comment says the ia.* keys are "Not rendered yet" while this same PR renders them.
- tokens.css:205: several phase-0 token groups ship with zero consumers (the whole --z-* scale, drawer/sheet, tooltip, urgency-pulse); fine as staged foundation, but worth pruning what the later phases did not end up using.
- src/ui/i18n.status.json is gitignored (.gitignore:21) yet tracked on the release branch, and this PR grows it by ~7k lines. Not your fault (it was merge-resurrected on the base), but dropping it from the branch would shrink the diff.

## Merge state

The only true conflicts against release/v0.24.0 right now are the generated i18n quartet (pending.ts, i18n.resolved.sha256, i18n.status.json, i18n.status.summary.json). That is mechanical stale-base churn: merge the base, re-run the i18n build, and re-baseline the hash, which also fixes blocker 1.

MasterZensei:
## Round 3: every mobile menu was see-through, root cause found and fixed branch-wide

All mobile window interfaces rendered completely transparent (world and nameplate text legible through the panel interiors). A six-agent live audit of all 30 windows pinned the root cause, and an independent four-agent verify fleet re-measured every window after the fixes.

### The root cause (one CSS rule)
The desktop resize-grip rule paints TWO background layers on framed windows (12px corner grip + the panel gradient) with per-layer size/position/repeat lists. The mobile override swapped in a SINGLE-layer background-image without resetting those lists, and CSS truncates per-layer lists to the image count: the leftover first values (12px 12px, bottom-right, no-repeat) applied to the panel gradient itself. The entire ~95%-opaque panel painted as one 12x12px dot parked at the bottom-right corner of every resizable framed window. The options menu alone survived because its earlier bespoke solid background-color is immune to background-size, which is exactly why it was the one menu that looked right.

Three-part fix: reset the layer geometry in both grip overrides (shared + heroic-shop twin), give every framed window the solid L2 base on touch (the tokens doctrine: a modal surface never composites the world through it), and extend the same base to the frameless legacy panels (quest dialog, market, loot, trade) the verify fleet caught escaping. A source-scan regression test pins all of it.

### Verified fixes beyond opacity (from the audit + verify fleet)
- Market + Bags stacked on the SAME full-screen rect on touch: the market now docks bank-style, market left / bags right, flush bottoms, zero overlap.
- Report window sat half off the left screen edge on mobile (unreset centering transform): recovered.
- Daily Rewards was half-unreachable on phones (height-bounded 2-column flow overflowed sideways into clipped columns; Verify Wallet and Spin hit-tested to the backdrop): it joins the mobile sheet and the columns balance to the full content with one vertical scroller, verified 0px horizontal overflow.
- Vendor had two nested scrollports fighting the same touch gesture: single root scroller now.
- Minimap cluster and mail badge covered the top strip of open sheets (leaderboard tabs, town-focus steppers): hidden while a window is open.
- Trade Accept/Cancel sat below the fold with no affordance: sticky action row, verified reachable at 900x430.
- More drawer joins the solid base; its 15 pills grow to 44px.
- Touch-floor sweep: char skin swatches, bag chips, vale cup brackets and role rows, town-focus steppers, harvest rows, market search, delve buttons, options scopes/sliders, all verified at or above 40x40.
- Wide-touch (1200px+) clamp: sheets no longer stretch content across 1580px; verified 900px centered.
- Arena Fiesta tab: legible again over the solid panel base (it was washing out over the transparent one).

### Also in this push
- The two top-left mobile disclosure chips (menu collapse arrow, consumables chevron) rendered at different sizes because one inherited a container scale transform: both now size from ONE shared block at a fixed 40px face at every Button Size setting, with anti-drift test teeth (solo chip rules cannot size, chip containers must stay transform-free).
- Enemy nameplates: cutoff reduced 55yd to 40yd (friendly/NPC navigation plates keep 55yd), smoothstep distance scaling to a 0.7 floor with a 5yd fade-out band so plates dissolve instead of popping, and ~0.8x plate sizing on mobile as a platform adaptation (deliberately not an fx-tier knob per the graphics-fairness invariant). Constants and falloff math pinned in tests.

Evidence: docs/screenshots/mobilewin-*.png (before/after character window, market dock, arena) and docs/screenshots/nameplates-*.png. Gate: 12 targeted suites (263 tests) green on the merged state, tsc clean, biome-changed clean, pre-push floor green.

In flight: the talents window tiered-choices redesign and the mobile camera fixes (post-pinch rotation lock, settle jitter) land next.

MasterZensei:
## Round 3 complete: talents redesign + mobile camera fixes landed

### Talents window: classic tiered-choices picker
Rebuilt on the frame grammar per the reference baseline, in our dark-gold token language: a level rail with tier rows derived from the real talent content (10 / 11-12 / 15 / 18 under our 1-point-per-level economy), three-column talent cards with selected / selectable / locked-by-level / dormant states, CHOICES and SPECIALIZATION tabs on the tab-rail with pips and arrow-key roving, a status strip (points spent, chosen spec), and Current Build / Create Build panels. Every existing behavior is preserved and pinned by tests: the staged-edit buffer, apply/confirm, loadout save/switch/delete/export/import, spec choice, tooltips, focus trap and return, the N toggle. Choice nodes keep the flyout. Desktop and mobile screenshots under docs/screenshots/talents-redesign-*.png. An independent reviewer returned a MERGE verdict with no blocking findings.

### Mobile camera: rotation never locks, no settle shake
- Lock after pinch (root cause): pinch pointers hold no pointer capture, so a finger lifting over HUD chrome delivered its pointerup elsewhere and the pinch entry leaked forever; every later single-finger touch was misread as a pinch against a stale phantom point, so drags only zoomed. Fixed with window-level up/cancel forwarding, blur/hidden release, and a pinch-degrade handoff that adopts the remaining finger straight back into camera rotation (with an origin resync so there is no jump). Live evidence: before, 0 of 3 drags rotated; after, 3 of 3 rotate with zoom untouched.
- Settle shake (root cause): on release, the follow camera replayed the tick-lagged facing echo of the camera's own writes back into the yaw, overshooting the chosen heading and dragging back (~6.8 degrees of bounce over ~700ms measured). A pure release-hold in camera_follow keeps follow disengaged until the echo converges (0.02 rad), a keyboard turn takes over, or an echo-scaled grace lapses; the same hold covers the online path where the echo is a full round trip. Live evidence: identical flick, yaw flat at the released heading for the full second.
Both fixed test-first: 13 new unit tests reproduced the bugs before the fix (numeric traces in the tests), 324 tests green across the merged state, tsc clean, pre-push floor green.

This closes the round-3 list: see-through menus (all 30 windows audited and re-verified), arrow chip consistency, talents redesign, camera lock + shake, and enemy nameplate range/scaling. Known follow-ups noted for later polish: a handful of Minor reviewer findings on the talents window (dead --color-talent-arrow token, coming-soon path shows empty tabs for a treeless class, a stale opt-in browser-suite selector), and the market listing area is cramped (96px) on short landscape phones (pre-existing).

MasterZensei:
## Round 4 complete + release/v0.24.0 merged (conflicts resolved)

### Upstream merge
The branch now carries the latest release/v0.24.0 (enchanting profession, jail system, native Apple/Discord auth, the joystick autorun lock, wire-aura alloc fix). The one semantic collision: upstream #1724 retired the mobile autorun button and its utility cluster for the joystick lock, so this branch's dead selectors were dropped and the touch-floor tests re-modeled around Jump's ring seat. Locale overlays were unioned and the generated i18n tables regenerated with the SHA re-baselined (never hand-merged). A four-area release-merge audit (input chain, markup+CSS, HUD TS, and the known trap classes: release-authored db mocks, i18n hash equivalence, injected-helper re-binding, surface inventory rows) verified no side of the merge lost behavior; all areas clean.

### Round 4 items (all five landed, each live-verified with before/after screenshots)
- Bags/bank item grids on touch: new 56px cell floor with the grid REFLOWING to fewer columns instead of shrinking cells; measured 46.8 to 63.7px cells inside the 50/50 bank and market docks, sockets 40 to 56px, icons fill the cells (docs/screenshots/bags-touch-*).
- Talents layout iteration: every card now sits explicitly on its tier's grid row (a structural pin, with a documented release for the narrow one-column collapse), the window uses its full height (dead band below the build panels 232px to 17px, build panels ride as a bottom full-width band, one scroller), 52px card icons and 68px cards, and wide viewports keep a centered 1160px reading column (docs/screenshots/talents-redesign2-*).
- Mobile chat revise: the newest line is always readable above the resize bar (reserved padding), the default footprint shrinks to ~56% width x ~48% height left-anchored, and the panel is MOVABLE via a 40px corner grab chip with the position and resized height persisted together and clamped so the Chat button close affordance can never be covered. Built on a new pure geometry core with 18 unit tests (docs/screenshots/chat-mobile-*).
- Settings on touch: category rail rows 40 to 54px with 24px icons and a wider rail, the detail column centers instead of left-hugging (864px centered at 1600px wide), and the controls grow (60x34 switches, 26px slider thumbs on an 8px track, 48px inputs/footer actions). Desktop measured byte-identical (docs/screenshots/settings-touch-*).
- More drawer rebuilt on the window-frame grammar: solid L2 panel (measured opaque), 48px titlebar, a 3-column grid of 64px icon-over-label tiles replacing the cramped pills, a modal scrim behind it, and the minimap/mail floaters hide while it is open. This also fixed real play.html drift (twelve missing type="button" attributes and an absent Daily Rewards tile), so both entries are byte-identical again (docs/screenshots/more-drawer-*).

Gate on the integrated tree: 17 suites (467 tests) green, tsc clean, i18n artifacts fresh, pre-push floor green. Follow-up candidates noted in the tree: the dead vendor-open dock CSS (the vendor is standalone since the Market-parity round) and the pre-existing browser-suite failures (party rows, community toggle) that exist identically on the base.

MasterZensei:
## Review response: all findings addressed

Every finding from the review was independently re-verified against current HEAD (the review examined an earlier commit, db3a5eff2, before several rounds of live-feedback work and two upstream release merges landed). Disposition below; commits at HEAD (a1c92dae8).

### Blocking

- **Stale i18n baseline**: already fixed before this review arrived (re-baselined during an earlier upstream merge). `i18n_resolved_equivalence` is 5/5 green at HEAD.
- **Gamepad connect/disconnect crash on the mobile settings shell**: was still live, now fixed (`02cbfc746`). `refreshControllerLabels()` routes to the shell painter instead of the desktop rail/detail/footer trio when the back-stack shell is open. Pinned by a source guard.

### Should fix, all fixed

**Language switch**
- Frame chrome now re-resolves `t()` on a live language switch: the builder stamps its i18n keys as data attributes, and a new `relocalizeWindowFrame()` re-stamps title/tab/close-aria text for every mounted frame from the HUD's language fan-out, open or closed.
- The options window's language-switch success path now runs a full re-render from any entry point (detail, Overview pin, or search result), not only when already on the Interface category.

**Focus and keyboard**
- `focusFirst()` now skips both `[data-close]` and `[data-window-close]`, so converted windows (quest log, loot settings, town focus) no longer land initial focus on the dismiss X.
- A segmented radiogroup with a stored value matching no option now keeps one tab stop (nearest-value fallback); the same for rail tabs during the bug-report sub-view.
- Opening the mobile settings shell now seeds focus into the search input or first category row (the shell's close/back/Done controls also carry `data-close`), instead of silently failing to focus a hidden desktop close button.

**Controller**
- The generic pad menu fallback now steps focus through the active trap's visible controls (mirroring the options window's own row-stepping), so every non-options trapped dialog (quest reward choice, crafting, town focus, delve board, loot settings, etc.) has pad focus movement again.
- Mobile-shell D-pad/paging verbs now resolve against `.opt-mshell-content` when `.opt-detail` is absent.
- `gamepad.ts` resets `prevStickX` on disconnect (matching `stop()`), closing the stale-crossing edge case.

**Settings correctness**
- The native-shell graphics-preset cap and environment gating now live in one shared control binder used by the detail pane, Overview pins, and search results alike, so a native-shell user can no longer dispatch a preset the shell doesn't support.
- Chat timestamp settings (show + 12h/24h) are back in Interface > Chat, wired to the already-plumbed hooks; a player who had them on can turn them off again.
- Rail changed-count badges, the category-head Reset button, and the Overview changed summary now sync in place after every setting change, not only on a full re-render.
- Scoped and per-row resets that touch a reload-required key now set the reload-pending flag.

**Vendor**
- Scroll position is now captured/restored on `.window-body` (the desktop scroller), not just the root, so the goods list no longer snaps to the top on every purchase.
- A window that already carries a baked inline position no longer re-cascades from that stale rect (it re-clamps in place instead), stopping the compounding drift; a live desktop-to-touch flip also clears baked inline geometry so it can never beat the mobile dock.

**CSS**
- The legacy `.btn` skin (margins, black outline, text-shadow, hover glow, the forbidden `:active` transform, the `!important` focus ring) no longer leaks into grammar buttons; the token focus ring wins as designed.
- `--color-notify-success` now theme-varies with the same double-guard `ensureReadable` pattern already used for warn/critical, fixing the ~1.3:1 contrast on the parchment theme.
- The touch slider tap-floor gap was already closed by an earlier round of settings-touch work; confirmed still correct at HEAD.
- A live mobile-touch class flip while the settings menu is open now re-renders on the derived layout-mode change (a `MutationObserver` on the body class), so rotating across the touch breakpoint can no longer leave mismatched or doubled chrome.

**Repo tooling / docs**
- All ~19 stale E2E/screenshot scripts repointed at the post-window-grammar markup (`.item-cell`, `.vendor-row`, `.filter-row .chip`, the frame tab rail, `.window-title`); every new selector grep-verified against the current builders and every script passes `node --check`.
- The opt-in `a11y.browser.test.ts` suite re-anchored on the frame's actual dialog contract (root or `:scope > .window-frame`, `[data-window-close]`) instead of the pre-migration markup; ran it end to end, 21/21 pass.
- The `/wiki` controls page and README now document the shipped keybind defaults (see the callout below) instead of the old Q/E-strafe text.
- The gitignored `i18n.status.json` registry is untracked (it was accidentally committed upstream; regeneration and every consuming test read it from disk, confirmed).
- The `~200` lines of dead legacy options panel builders (and their misleadingly-green pinning tests) are deleted, replaced by the one shared control binder the settings-correctness fix above introduced.
- Conversation-artifact comments ("The user asked for...") reworded to impersonal design rationale; the `input.ts` header updated to the shipped default description.

### Nits, all fixed
Every listed nit (the unpainted close-button id, the dangling `aria-controls` on inactive tabs plus the missing `aria-labelledby`, the `ensureFrame` reuse-path contract note) is addressed in the same commits above.

### Design decision (K2): keybind default modernization

**Approved.** Q/E move to action-bar slots 10/11, strafe ships unbound by default (fully rebindable), Shift+1-0 pages the secondary bar, Z toggles the damage meters. This is called out prominently in the PR description now. Persisted player layouts are completely unaffected (verified against the save/load path); only never-customized fresh profiles get the new defaults, and they no longer open in a lit-up keybind-warning state (the K1 fix above exempts intentionally-unbound defaults from that aggregate).

### Deliberately deferred, tracked rather than fixed here
- **T5** (converting the options-window dispatch-parity tests from source-text greps to executed-DOM tests): real test-engineering effort, tracked as a follow-up, not a correctness gap.
- **T6** (consolidating ~12 duplicate `ensureFrame` copies onto the shared `ensureWindowFrame` helper): a refactor with no user-facing effect, tracked as a follow-up.
- One flagged case from the fix itself: the social/talents windows' own tab-sync code reads `aria-controls` directly rather than through the new `applyActiveWindowTab` helper the frame's tab-aria fix introduced; a programmatic (non-click) tab switch on those two windows leaves metadata-only stale wiring until the next click. Filed as a background task rather than folded into this pass to keep the fix scoped.

Gate on the fully integrated tree: 22 suites (523 tests) plus the browser a11y suite (21/21) green, tsc clean, biome-changed clean, i18n fresh, pre-push floor green.

Rubsey:
Maintainer fix round pushed to this branch (dc6f0e28a..b59a2657a, 21 commits), from hands-on desktop + Android testing. Summary:

**Settings / Esc menu**
- Language now leads the Overview pins, and each pin renders as one .opt-pin wrapper cell so the touch shell's 2-column grid actually fills both columns (the per-pin home crumb used to break the grid after every row). The theme pin is mirror-only; Custom Colors stays on the Interface page.
- renderDetail() routes the back-stack shell to the full render(): tapping a theme preset (or any rerender row) on touch used to throw on the missing desktop pane.
- Opening the menu on touch no longer focuses the search input (no more surprise soft keyboard); the seed lands on the first category tile. Note: a comma selector list resolves in DOM order, so the input had to leave the selector entirely.
- Controller Button Layout wraps in a real .opt-section so it flows 2-across like every other page, with the dropdown flyout re-tokened to the Codex grammar.

**Windows**
- Static centered dialogs (quest/gossip, confirm, delve board/rite, lockpick, loot, report) neither drag nor cascade: STATIC_DIALOG_WINDOW_IDS in window_drag_handle.ts, consumed by both the drag predicate and placeNewWindow.
- Bags/bank desktop grids grew to the 56px --slot-cell floor (auto-fill was minting columns instead of growing cells), sockets match, and the bag bar wraps so the capacity counter never clips.
- The bags coin purse moved into the frame's pinned footer (it was the scrollable body's last child, below the fold on touch) with a proper coin-triplet readout.
- More drawer first-open landed half off-screen: the open rule referenced an undefined --mobile-more-open-transform custom property (orphaned by a release merge) and the opener's inline-style mask got wiped exactly once per session by the Hud observer. Stylesheet now owns the centering.
- Vendor Sell Junk gap restored (.btn.vendor-sell compound beats the same-layer '.window-frame .btn { margin: 0 }' reset); same fix applied to the keybind-capture Cancel.

**Shell / input**
- Create-screen 3D preview was empty on a cold-cache connection: the assetsReady init never considered #charcreate-panel, so the canvas parked in the hidden charselect container. Init now scans all three play panels and routes through updatePreviewContainer, with a .catch so a failed preload degrades instead of rejecting unhandled.
- Pinch re-baselines on a 3-to-2 finger transition (an accidental third finger caused a one-shot zoom jump).

**Housekeeping**
- Merged origin/release/v0.24.0 (daily-rewards bans, loading perf, pet bar keys, heroic lockout); a 4-way audit of the merge found one vacated test invariant (the unbound-chord input test collided with the new Ctrl+1 pet default), restored.
- Re-anchored two stale source scans (vendor WCAG close-button, mobile no-clip) and the rite-popup drag pin.
- Every fix is test-pinned; scripts/pr1736_fix_round_verify.mjs is a committed browser harness covering the visual fixes end to end.

Verified: npm run gate green (full suite 12k+ tests, tsc, builds, i18n freshness, malware scan), plus live browser verification on desktop and emulated touch, plus on-device Android testing against a LAN dev stack and the dev host deploy.

PR #1739, opened 2026-07-10
feat(social): add /playtime for lifetime played time (#1703)

### LINKED ISSUE #1703: Add a /playtime command to show total time played on a character
## Problem

There is no way for a player to see how much time they have played on a character. Players want a simple, on-demand readout of their total time played, the same way classic-era clients exposed it.

## Expected behavior

- A `/playtime` chat command reports the total time played on the current character.
- The readout covers the character's whole lifetime, accumulated across all sessions, not just the current session.
- The value is shown in a readable form (for example, total hours, or days plus hours).
- Time played is tracked server-side so the figure is authoritative and survives relog, and it advances only while the character is actually in the world.

## Scope

- A new `/playtime` command in the chat command handling.
- Server-side tracking and persistence of accumulated played time per character.
- The response text the command sends back to the player.

## Acceptance criteria

- Typing `/playtime` returns the character's total time played.
- The reported time persists across logout and login and increases only during active play.
- The command works the same online (authoritative server) and, if applicable, offline.
- All player-facing text from the command is localized (a `t()` key plus values), consistent with the repo i18n rules.

jgyy:
## Summary

Closes #1703.

- Adds a `/playtime` chat command reporting a character's **lifetime**
  total time played, persisted server-side and surviving relog/server
  restart. This is distinct from the existing `/played`, which only
  reports the current session and resets to zero after logging back
  in.
- New `PlayerMeta.totalPlayedSeconds` (session-baseline, loaded from
  the save) combines with the existing `joinedAt` session clock the
  same way `/played` already does; `serializeCharacter()` folds the
  elapsed session into the persisted baseline on every save, so the
  figure only ever grows while the character is actually in the
  world.
- New optional `CharacterState.totalPlayedSeconds` field (defaults to
  `0` so pre-existing saves load cleanly); no schema change needed —
  character state is stored as JSONB and `saveCharacterState` persists
  the whole blob generically.
- `/playtime` is added to the `/help` character-readouts line.

```mermaid
sequenceDiagram
    participant P as Player
    participant S as Sim (server)
    participant DB as Postgres (JSONB state)

    P->>S: /played
    S-->>P: "Time played this session: 2m 5s."
    P->>S: /playtime
    S-->>P: "Total time played: 2m 5s."

    Note over S,DB: logout / server restart
    S->>DB: serializeCharacter() folds session into totalPlayedSeconds
    DB-->>S: addPlayer() loads savedState.totalPlayedSeconds

    P->>S: /played (new session)
    S-->>P: "Time played this session: 0s."
    P->>S: /playtime
    S-->>P: "Total time played: 2m 5s."
```

## Not a UI change

This is a chat-command-only feature (no HUD/render surface), so there
are no before/after screenshots per the PR template's visual-change
guidance. In lieu of that, here's an actual run against the built sim
showing `/played` resetting across a simulated relog while `/playtime`
keeps the lifetime total (`sim.time` resets to 0 to model a server
restart, `savedState` round-trips through `serializeCharacter`/
`addPlayer` exactly as a real save/load would):

```
[server->Aleph] Time played this session: 2m 5s.
[server->Aleph] Total time played: 2m 5s.

-- relog (server restart, sim.time resets to 0) --

[server->Aleph] Time played this session: 0s.
[server->Aleph] Total time played: 2m 5s.
```

## i18n

`/playtime`'s reply text follows the exact precedent of the
pre-existing `/played` command in the same file: both are self-only
`ctx.error` readouts from the v0.7 slash-command surface, which ships
English as a documented backstop (see the `ALLOW_V07_SLASH` comment in
`tests/localization_fixes.test.ts`). I added the new concrete string
to `scripts/i18n_blocked_seed.mjs` alongside `/played`'s and
regenerated `src/ui/i18n.status.json` / `i18n.status.summary.json` via
`npm run i18n:gen` so the S3 drift guard stays green.

## Test plan

- [x] New Vitest coverage in `tests/chat.test.ts`:
  - `/playtime` reports `0s` on a freshly joined character.
  - `/playtime` accumulates as the sim advances and, unlike `/played`,
    survives a simulated relog (`serializeCharacter` -> fresh `Sim` ->
    `addPlayer({ state })`), continuing to accumulate on top of the
    prior total in the new session.
- [x] `npx vitest run tests/chat.test.ts tests/localization_fixes.test.ts tests/parity/parity.test.ts tests/architecture.test.ts tests/sim.test.ts` — all green.
- [x] `UPDATE_PARITY=1` regen of `tests/parity/golden/chat_social.json` (new field appears in serialized player state).
- [x] `npm run gate` full suite green (11k+ tests) after fixing a local worktree env gap (missing `.env`/`node_modules`) unrelated to this change; confirmed the one remaining flaky suite (`tests/server/http/parity.test.ts`, a DB-timing sensitive HTTP dispatch-parity harness) reproduces identically on a clean, unmodified `release/v0.24.0` checkout, so it predates this change.
- [x] `npm run build` succeeds.

### DIFF
--- a/scripts/i18n_blocked_seed.mjs
+++ b/scripts/i18n_blocked_seed.mjs
@@ -288,6 +288,7 @@ export const V07_SLASH = [
   'Threat is only tracked on enemies; Aki is not one.',
   'Threat on Aki (5): Aki.',
   'Time played this session: Aki.',
+  'Total time played: Aki.',
   'Unknown command: Aki. Type /help for a list.',
   'Vendor buyback (5): Aki. Repurchase at any merchant.',
   'You are Aki.',
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -988,9 +988,10 @@ export interface CharacterState {
   unlockedMilestones?: string[];
   // Rested XP pool. Optional so pre-rested-XP saves load cleanly (defaults to 0).
   restedXp?: number;
-  // Lifetime played time in whole seconds, accumulated across every prior session
-  // (this session's elapsed time is folded in at save; see PlayerMeta.totalPlayedSeconds
-  // and /playtime in social/chat.ts). Optional so pre-/playtime saves load cleanly
+  // Lifetime played time in seconds (unfloored, for drift-free accumulation; only
+  // the display floors), accumulated across every prior session (this session's
+  // elapsed time is folded in at save; see PlayerMeta.totalPlayedSeconds and
+  // /playtime in social/chat.ts). Optional so pre-/playtime saves load cleanly
   // (defaults to 0).
   totalPlayedSeconds?: number;
   // Gathering profession proficiency (JSONB; optional so pre-professions saves
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -847,6 +847,11 @@ export interface PlayerMeta {
   // sim.time when this character entered the world; powers /played. Session-only
   // (sim.time resets to 0 each server boot), so it reports time this session.
   joinedAt: number;
+  // Seconds played across every session BEFORE this one (loaded from the save;
+  // powers /playtime). Combined with `this.time - joinedAt` for the running
+  // lifetime total; folded into a new persisted baseline by serializeCharacter
+  // on save, so it only ever advances while the character is actually in the world.
+  totalPlayedSeconds: number;
   // Tick of the player's last deliberate action (movement, ability cast, or pet
   // command). Session-only, never persisted. Powers the anti-AFK gate on
   // aggressive pet auto-pull (see PET_OWNER_IDLE_TICKS) so an idle owner's pet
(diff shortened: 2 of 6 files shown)


FernandoX7 (commented):
Clean, well-scoped feature. The lifetime accumulator is modeled correctly: `totalPlayedSeconds` and `joinedAt` are each set once at `addPlayer` and only read afterward, so `serializeCharacter`'s fold (`baseline + (time - joinedAt)`) is idempotent across repeated saves, with no double-count and no drift across relogs. `/playtime` mirrors the existing `/played` faithfully and reads the running total without waiting for a save. Persistence is additive and back-compat (optional JSONB field defaulting to 0), and the relog round-trip is the right thing to pin.

Verified locally: the two new `tests/chat.test.ts` cases pass, the `localization_fixes` and `architecture` guards are green, and `tests/parity/parity.test.ts` passes against the committed golden. The `chat_social` `events` digest moved because the `readout-emote` scenario runs `/help` and the readouts line now lists `/playtime`; `state` is unchanged since that scenario serializes no character, so the regen is exactly what you'd expect.

One nit:

- `src/sim/sim.ts:991` the `CharacterState.totalPlayedSeconds` comment says "in whole seconds", but the persisted value is an un-floored float (the fold and the load never round; only the display floors). Storing the fractional value is the right call for drift-free accumulation, so this is just the comment being slightly off. Consider dropping "whole" or noting it is floored at display time.

The English backstop on the v0.7 slash surface (seeding `Total time played: Aki.` next to `/played`'s line and regenerating the status files) is the correct handling here.

jgyy:
Thanks for the review! Fixed the comment nit at `src/sim/sim.ts`: it now says the value is unfloored (kept as a float for drift-free accumulation) and only floored at display time, rather than "whole seconds". Pushed in 2c4f344c9.

FernandoX7 (approved):

PR #1740, opened 2026-07-10
fix(mobile): remove Appflow OTA updates

TrevCavill:
## Summary

- release v0.23.1 as a store hotfix
- remove the Appflow Live Updates dependency, configuration, and native platform linkage
- increment Android versionCode from 15 to 16 and iOS CURRENT_PROJECT_VERSION from 14 to 15
- replace OTA release guidance with store-only delivery and upgrade-path QA

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Build, CI, or tooling

## How was this tested?

- npm run native:sync
- npx vitest run tests/version_sync.test.ts (11 passed)
- npm run gate (release tier, all 9 steps passed; 11,305 tests passed)
- verified no Appflow app ID, Live Updates package, or native linkage remains

## Checklist

- [x] Tests pass
- [x] Typecheck passes
- [x] Builds succeed
- [x] No player-visible strings added
- [x] No secrets or production dev commands added
- [x] Generated native linkage was regenerated with Capacitor sync


FernandoX7 (commented):
Clean, well-scoped hotfix, and the right fix for the stale-snapshot problem: dropping the Live Updates plugin entirely means the binary serves its own bundled web assets instead of a pinned Appflow snapshot.

Verified:

- The removal is complete. No `live-updates`, `LiveUpdates`, or Appflow app-id reference survives anywhere in native code or config (a whole-tree grep leaves only the negation in `docs/mobile-store-release.md`). `capacitor.config.ts`, both Android gradle files, iOS `Package.swift` + `Package.resolved`, and `package.json` all drop it.
- The version bump is consistent on every live surface: package.json + lockfile (0.23.1), Android `versionName 0.23.1` / `versionCode 16`, iOS `MARKETING_VERSION 0.23.1` / `CURRENT_PROJECT_VERSION 15`, `index.html`, `play.html`, and `DESKTOP_VERSION`. Nothing live is left at 0.23.0 (the remaining 0.23.0 hits are all test fixtures and historical doc references).
- The lockfile flip of `@capacitor/core` and `tslib` to `"dev": true` is correct, not a slip: `@capacitor/core` was already a devDependency and only showed as non-dev because the removed production `@capacitor/live-updates` pulled it in as a peer. Straight regen.
- `tests/desktop_download.test.ts` interpolates `DESKTOP_VERSION`, so the bump keeps it green.

Two notes, neither blocking:

- The docs rewrite is good, and the added "update over an existing store install, relaunch several times, plus one offline launch" step gates the actual risk: a previously-downloaded OTA snapshot persisting across an in-place update, not just a fresh install. Worth actually running that update-path test on a real device before submission, since the code change assumes the removed plugin no longer redirects the webview to any orphaned snapshot directory.
- `android/app/capacitor.build.gradle` is left with two blank lines where the include used to be. That is generated `cap sync` output, so it is expected, not something to hand-edit.

Looks good to ship.

FernandoX7 (approved):

PR #1745, opened 2026-07-10
perf(server): scale pass, 100 to 600+ dense CCU per realm (multi-core snapshot fanout)

maxpolaczuk:
## Summary

Server scalability pass: one realm process goes from struggling at ~100 concurrent players to holding 600 players in dense simultaneous combat inside the 50 ms tick budget, with zero gameplay-visible change. The sim stays single-threaded and deterministic; everything around it (the snapshot broadcast, the measured hot path) now scales across cores.

Verified numbers (M4 Pro, join-audited bot fleets via `scripts/server_load_jitter.mjs`; full methodology in `docs/design/server-scale-v0.24.md`):

| scenario | before | after |
|---|---|---|
| 250 bots fighting in one camp, tick p95 | 182.7 ms (3.6x over budget) | 8.3 ms (workers on) / 27.4 ms (workers off) |
| 600 bots in three dense 200-player combat camps | unreachable | 45.8 ms, inside budget |

The workers-off row is what a 2-vCPU production box gets (the pool auto-disables there). Ordinary spread-out populations sit well above these worst-case-density numbers; realms remain the horizontal multiplier past one process.

## What changed

Safety nets first, then two passes:

**Wire/sim bytes provably unchanged.** New `tests/server/broadcast_golden.test.ts` pins the SHA-256 of every outbound snap/event frame for a scripted 4-class combat scenario through the real GameServer pipeline; new `tests/server/snapshot_fanout_equivalence.test.ts` proves real worker_threads produce canonically identical snapshots and byte-identical event streams vs the in-thread path (with a non-vacuity check that frames actually crossed workers). The existing `tests/parity` golden (sim state + rng draw order) guards all sim changes.

**Single-thread pass** (broadcast golden byte-identical except one deliberate cadence tier):
- `selfWireJson` diet: object-identity gates for stats/weapon (the sim replaces, never mutates), empty-map fast paths, one `Date.now()` per broadcast, and party/marks payloads built + stringified once per PARTY per tick instead of once per member (removes the raid quadratic).
- Secondary self panels (mail/bank/market/professions/delve meta/lockouts) re-diff on a pid-staggered 4-tick cadence instead of 20x/s per player: at most 200 ms added change-detection latency on panel UI, nothing combat-actionable (per the graphics-fairness bar); a `selfHeavyDirty` session re-diffs immediately so a player's own action still lands next snapshot. The broadcast golden was reminted once, deliberately, for this.
- `routeEvents`: pid-indexed batch + order-preserving merge (`server/event_routing.ts`), O(events + candidates) instead of O(sessions x events).
- `runAntibotTick`: the detector declares whether it reads per-tick runtime snapshots (`wantsTickSnapshots`); the no-op stub declines, the real private detector (flag absent) keeps receiving them.
- Inbound: hand parser for the canonical 20 Hz movement frame (`server/input_frame_fast.ts`) with JSON.parse fallback for any other byte shape, equivalence-fuzzed (including JSON's leading-zero rule).
- Terrain, bit-identical (parity + terrain_walls + custom_map_parity green): per-64yd-cell shortlists of the hubs/lakes/camps whose gate radius can reach the cell (conservative superset, original arithmetic re-checked in original order), camp-flatten height memoized per (camp, seed), rim crest noise skipped only where its contribution is exactly +-0. Terrain noise was 13% of busy CPU.
- Colliders, bit-identical: packed numeric grid keys (no string allocation per 0.2 yd movement sub-step), per-cell fence shortlists with an exact linear fallback, `combatProfileForMob` memoized.

**Multi-core pass** (`server/snapshot_fanout/`, own CLAUDE.md):
- The per-session interest scan + ents/keep assembly (measured ~43% of busy CPU at 250 co-located players) moves to a worker_threads pool. Numeric per-entity facts ride a double-buffered SharedArrayBuffer mirror with ownership (a dispatched tick pins its buffer until the worker replies, so a slow worker can never observe a torn buffer); wire-fragment strings ship as per-worker deltas against each worker's own ledger, so a busy or respawned worker always catches up to exact current fragments before building.
- The interest ladder (radii, hysteresis, rate tiers, full/lite/keep decisions) lives in one pure module (`interest_rules.ts`) consumed by BOTH build paths, so they cannot drift.
- Authority and privacy stay main-thread: self JSON, stealth visibility (live party/duel/hostility state, shipped as per-session denied ids), and every socket write (backpressure guard unchanged).
- Degradation order: busy shard skips the tick (clients coast on interpolation), dead shard builds in-thread, repeated failures disable the pool for the process. `SNAPSHOT_WORKERS=auto|0|N`, auto = `min(6, cores - 2)`, so a 2-vCPU box auto-disables rather than starving the sim. Bare `new GameServer()` in tests never spawns threads.

## Review

Three review agents ran over the diff; every substantive finding is fixed in-branch:
- qa-checklist (verdict READY): mirror-tearing race under worker overload -> double-buffered mirror with ownership.
- architecture-reviewer (sim changes bit-identical, rng-draw-order neutral): stale editor terrain caches on in-place hub/lake/camp mutation -> `invalidateTerrainFeatures()` wired into the editor's mutation funnels + contract test.
- privacy-security-review (no critical/warning): fast-parser leading-zero divergence from JSON -> rejected, fuzz cases added.

## Testing

- Full suite green on the merged branch (11,400+ tests), including the new golden/equivalence/fuzz/invalidation tests.
- `tsc --noEmit`, `npm run i18n:gen` freshness, `npm run security:gate`, `npm run build:env`, `npm run build:server`, `npm run build` all pass locally (the PR-gate job list).
- E2E against a worker-enabled server: `scripts/mp_integration.mjs` (26/26), `scripts/ws_security_e2e.mjs` (11/11), `scripts/chat_e2e.mjs` (11/11).
- Load-tested before/after at 100/150/250/600/1000 bots; `scripts/server_load_jitter.mjs` gained `BOT_PREFIX` so concurrent multi-hub fleets cannot collide on usernames (which had silently shrunk one early measurement; all published numbers are join-verified).


maxpolaczuk:
Please review @FernandoX7 interested in your thoughts

FernandoX7:
> Please review @FernandoX7 interested in your thoughts

This is great! I'm afraid we're going to overlap on work though. This is part of all the performance improvements I've been working on.

EnriqueGF:
Thanks for this one :) It is a seriously impressive piece of engineering, and the safety-net-first approach (a real SHA-256 broadcast golden, a real worker_threads equivalence test with a non-vacuity counter, and an oracle-based fuzz against `JSON.parse`) is exactly how a change this invasive should be built. The interest-ladder extraction into one shared `interest_rules.ts` that both build paths import is the right seam, and I could not find any drift between them.

I cannot approve it yet. There is a blocking base-branch gate, and two correctness defects in the delivery seam that I confirmed by reading the code. Details below, ordered by severity.

## Blocking

### 1. Wrong base branch (gate)

`gh pr view 1745 --json baseRefName` reports `release/v0.24.0`. Every PR must target the latest release branch, which is now `release/v0.26.0`. This is blocking on its own, and here it is not a formality:

- `release/v0.24.0..release/v0.26.0` is 399 commits, and GitHub already reports the PR as `CONFLICTING`.
- The drift lands squarely on the files this PR rewrites: `server/game.ts` (+724), `server/main.ts` (+371), `src/sim/world.ts` (+85), `src/sim/colliders.ts` (+15), `tests/snapshots.test.ts` (+554).

That means the two headline safety nets have to be re-established after the rebase, not just re-run: the `broadcast_golden` fixture has to be reminted against the v0.26 wire, and the "terrain is bit-identical" claim has to be re-proven against the v0.26 `world.ts`. Please retarget onto `release/v0.26.0` and rebase.

### 2. The self-delta ledger is consumed on frames that are never sent

`server/game.ts:4026` calls `selfWireJson(...)` for every session while building the fanout jobs, and `selfWireJson` mutates `session.lastSent` (`server/game.ts:4184`, `sent[key] = s`) and clears `session.selfHeavyDirty` (`server/game.ts:4314`). Only afterwards does `dispatchTick` decide which jobs actually ship (`server/game.ts:4041`):

- **Busy shard** (`result.skipped`, `server/game.ts:4046`): the pending frame is deleted and nothing is sent. The delta fields computed for that tick were already recorded as sent. Since the client treats an absent field as unchanged, it keeps the old value until that field changes again.
- **Dead shard / disabled pool** (`result.fallback`, `server/game.ts:4047-4052`): `buildSessionSnapshotInThread` calls `selfWireJson` a second time (`server/game.ts:3948`). The first call already drained the diff, so the frame that does ship omits every field that changed on that tick.

Concretely: a player loots an item on a tick whose shard is busy, `lastSent.inv` is advanced, the frame is dropped, and the item does not appear in their bag until some other inventory change happens to re-dirty it. The same applies to `equip`, `qlog`, `tal`, `trade`, `party` and `corpse` (and `corpse` gates the resurrect-at-corpse button, so this is player-blocking, not cosmetic).

Note that `pool.ts:74-80` justifies the skip with "clients coast on interpolation". That is true for the entity ladder, which is version-backstopped per worker, but `lastSent` has no version backstop, so it does not hold for self state. Both paths are reachable exactly under the load this PR targets, and `pool.ts:239` returns *all* jobs as `fallback` when the pool is disabled, so a mid-run pool disable makes every player lose their heavy deltas on that tick.

Suggested shape: build `selfJson` once and reuse it for the in-thread fallback instead of rebuilding, and commit the `lastSent` / `selfHeavyDirty` / `lastWireRev` mutations only when the frame is actually written (stage the changed keys, apply on send, roll back on skip).

### 3. `secondaryDue` uses a tick-parity modulo that can starve permanently

`server/game.ts:4265-4267`:

```ts
const secondaryDue =
  session.selfHeavyDirty ||
  (this.sim.tickCount + session.pid) % SELF_SECONDARY_INTERVAL_TICKS === 0;
```

The loop runs `while (acc >= DT) { this.sim.tick() }` and then broadcasts once per timer callback (`server/game.ts:1284-1298`). Under sustained lag the sim advances 2 or 4 ticks per broadcast, so `sim.tickCount` steps by a constant greater than 1 at broadcast time, and `(tick + pid) % 4 === 0` can be permanently false for a subset of pids. Those players then never re-diff their secondary panels except via `selfHeavyDirty`, so `lockouts`, `dmarks`, `dclears`, `delveDaily`, `prof`, `gprof` and `tfocus` (none of which have a dirty setter) stop updating indefinitely.

This is the failure mode the file already knows about: `server/game.ts:3852` carries the comment `">= rather than a modulo check: catch-up broadcasts can skip ticks"`, and `interest_rules.ts:55-57` preserves the same warning. Unlike `heavyDue`, which has the `meta.wireRev !== session.lastWireRev` backstop, `secondaryDue` has none. Please use the same elapsed-since form the rest of the file uses: `tick - session.lastSecondaryTick >= SELF_SECONDARY_INTERVAL_TICKS`, stamped on send.

## Should fix

- **The degradation paths have zero test coverage**, which is why the two defects above survived. There are no assertions anywhere on `result.skipped`, `result.fallback`, `MAX_FAILURES` disable, `acquireMirror()` returning null, or the `MIRROR_CAPACITY` overflow skip. `snapshot_fanout_equivalence.test.ts` never kills a worker, so the fragment-ledger catch-up that the pool's own header calls the correctness cornerstone ("a respawned worker always catches up to exact current fragments before building") is entirely unasserted; a partial fragment set would emit stale entity records to real clients with no crash and no log. A pool unit test with a fake `FanoutHost` and a stub worker closes all of these.
- **`server/snapshot_fanout/pool.ts:166`: the pool disable is sticky.** `failures` accumulates for the process lifetime and is never reset on success, so 8 transient faults spread over days permanently move all snapshot building back onto the sim thread, precisely when load is highest. A decay or a consecutive-failure counter would be safer.
- **The camp-height memo is not invalidated on a water-level change (editor correctness regression).** `syncWaterToActive()` (`src/editor/app.ts:433-434`, reached from the water slider preview, commit, undo and load) mutates `activeWorld.waterLevel` in place, but never calls `invalidateTerrainFeatures()`. `baseHeight` depends on `waterLevel()` (`src/sim/world.ts:514`, `minLand = waterLevel() + 1.4`, and the lake carve at `:521`), while the new `campCenterHeight` memo (`src/sim/world.ts:478-487`) is keyed only by `(camp identity, seed, terrainFeatureGen)`. Water is in none of those keys, so after the slider moves every camp's flatten target is served from the pre-change water level. The old code recomputed `baseHeight(camp.center)` per sample, so this is a behavior regression the memo introduces, and it persists: the flatten tool bakes the stale height into a saved `HeightStamp` (`src/editor/app.ts:743`). Calling `invalidateTerrainFeatures()` inside `syncWaterToActive()` covers all four entry points.
- **`src/editor/3d/viewport.ts:291-301`: the new `invalidateTerrainFeatures()` call sits after both early returns** (`if (!this.visible) ... return;` and `if (!this.renderer) return;`). Hub and lake marker moves are exactly the mutations that happen while the 3D viewport is not visible: they mutate `content.zones[].hub` / `.lakes[]` in place and defer the re-mesh until the user returns to 3D. Camps have the direct `afterCampsChanged()` hook; hubs and lakes do not. So for the whole 2D session the shortlists hold the old footprints, and the 2D `terrainHeight` consumers read them (including the flatten target at `src/editor/app.ts:743`, which is persisted). Hoist the call above both returns, or invalidate at the mutation site.
- **The "terrain is bit-identical" claim is true but not pinned in CI.** I verified it empirically rather than trusting it: base vs HEAD `terrainHeight`/`groundHeight` over 8,056,232 samples across 4 seeds, compared with `Object.is` (so it catches `+0`/`-0` drift from the new `if (rim !== 0)` skip), gives 0 diffs; a 400,000-case fuzz of `resolveMovement`/`resolvePosition` across both the fence-grid path and the exact-scan fallback also gives 0 diffs. So the claim holds. But nothing in the suite would catch a future regression: `tests/terrain_feature_invalidation.test.ts` pins the invalidator, not the arithmetic, and `tests/parity` quantizes to 1e-6 and only samples where its scenarios walk, not the gate boundaries. A `terrainHeight` digest over a deterministic grid straddling every hub/lake/camp gate radius would make it a real pin.
- **`server/main.ts:330`: `SNAPSHOT_WORKERS=` (set but empty) resolves to 0, not `auto`.** `Number('') === 0`, so an empty value silently disables the pool. That contradicts the "a set-but-empty numeric env means the DEFAULT" rule in the very `DEPLOY.md` section this PR extends.
- **`wantsTickSnapshots` has no test on either arm** (`server/game.ts:1586`, `server/bot_detector/stub.ts:8`). The `!== false` is load-bearing back-compat with the private detector, and an "absent means true" contract deserves a pin on both sides.

## Nits

- `tests/server/broadcast_golden.test.ts:195` sets `process.env.ALLOW_DEV_COMMANDS = '1'` in `beforeAll` and never restores it, so it leaks to other tests in the same worker.
- `src/sim/colliders.ts:312` and the feature-cell key in `world.ts` alias for coordinates outside their offset range. Unreachable at current world bounds, but a clamp or an assert would make that explicit.
- `combatProfileForMob` now returns a shared object where callers previously got a fresh one (`src/sim/mob_combat.ts:65-77`). The consumers are read-only today and the Nythraxis branches always returned module constants, so the contract is unchanged, but `Readonly<MobCombatProfile>` on the return type would make it enforced rather than conventional.

## What I verified as clean

- `npx tsc --noEmit`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`, and all seven of the PR's new or changed test files pass locally on this branch.
- Determinism holds. No `Math.random` / `Date.now` / `performance.now` added to `src/sim/`, and no `Rng` draw was added, removed or moved: the `if (rim !== 0)` skip only elides `fbm2` calls, which are stateless hash noise, not the shared `mulberry32` stream, so the stream position cannot shift. The parity draw-order digest is green and was not regenerated.
- The sim caches are correctly keyed for determinism and do not leak across `Sim` instances or seeds: `featureShortlistCache` is a `WeakMap` on `WorldContent` and its contents are seed-independent (the gates are pure geometry), `campCenterHeightCache` carries the seed and recomputes on mismatch, `scaledProfileCache` is a pure function of `scale`, and the `FEATURE_CELLS_MAX` eviction is a pure `clear()` that cannot change a returned value.
- Bit-identity of both the terrain and the collider changes, verified empirically against the base branch (see the note above): 0 diffs across 8.06M terrain samples and 400k collision cases.
- The interest ladder is genuinely single-sourced: no radius, tier, or ladder constant is re-declared on either build path.
- Stealth privacy holds through the worker path. The `denied` window (`INTEREST_QUERY_RADIUS + 2`) strictly contains the worker's emit window, computed on the main thread against live party/duel/hostility state, so no stealthed player can be emitted that the in-thread path would have hidden. Every socket write stays main-thread.
- `input_frame_fast.ts` is conservative in the safe direction: leading zeros, whitespace, unknown `mi` keys, duplicate top-level keys, escapes and trailing bytes all fall back to `JSON.parse`, and `__proto__` cannot reach the object through the `MI_KEYS` allowlist. `null` means "shape not covered", never "valid", and the `invalid_json` anomaly path is preserved.
- `markersFor` reads only `party.id`, so the `markersFor(party.members[0])` substitution in the per-party cache is sound.
- `wantsTickSnapshots` is back-compatible: absent still means true, so the private detector keeps receiving snapshots.

## Verdict

Changes requested. Retarget and rebase onto `release/v0.26.0` first, then fix the two delivery-seam defects (2 and 3) with tests that exercise the skipped and fallback paths. The performance work itself is sound and I am keen to see it land.

One process note: FernandoX7 mentioned this overlaps with performance work he already has in flight. Worth settling that before the rebase, so the merge is done once rather than twice.

PR #1746, opened 2026-07-10
fix(server): cut per-tick GC churn in aura wire serialization

jgyy:
## Summary

Prod Grafana dashboards showed **sim tick p95/p99 intermittently breaching the 50ms/20Hz budget** (spiking toward 200ms) in bursts. These bursts lined up exactly with:
- WS **event-loop lag p99** spiking toward ~100ms
- **GC time/sec** spiking toward ~8ms
- **heap used** sawtoothing from a ~64MiB baseline to ~192MiB peaks

...while host CPU/memory/disk stayed comfortably low the whole time, ruling out host resource starvation and pointing at app-level allocation/GC pressure inside the Node process, scaling with the ~555 sim entities / ~42 players online at the time.

### Root cause

`wireEntity`'s per-aura serializer (`server/game.ts`, `dynamicFields`) built each `WireAura` via **nine chained conditional spreads**:

```ts
...(a.value !== 0 ? { value: a.value } : {}),
...(a.value2 !== undefined ? { value2: a.value2 } : {}),
// ...7 more
```

Every spread evaluates a fresh object literal (`{}` or `{value: ...}`) on **both** branches regardless of which side is taken, so a single aura could allocate up to 9 short-lived objects. `wireCacheFor` (the per-entity, per-tick, shared-across-sessions cache) recomputes `identityFields`/`dynamicFields` **unconditionally every tick** for every entity, before diffing the resulting JSON string against the cache, so this allocation happened for every aura, on every entity, every 50ms, regardless of whether anything actually changed. At raid-sized entity/aura counts and 20Hz, that's thousands of short-lived object allocations per second: a textbook generational-GC-sawtooth generator.

### Fix

Replace the spread chain with direct property assignment on one object (`wireAura`, extracted as its own function). The output is **byte-identical** to the old spread chain, same keys, same omission rules, same values, verified by a new pinning test plus all existing wire-format tests (bandwidth, snapshots, target_frame, auras_view, holder_broadcast) passing unchanged.

A local microbenchmark (555 entities × 4 auras × 10s of ticks at 20Hz) shows roughly **half the mapper CPU cost**:

```
OLD (chained conditional spreads): 41.5ms
NEW (direct assignment):           21.2ms
```

This is a pure server-side serialization change: no wire format change, no sim/gameplay change, no rng/determinism involvement.

## Root-cause diagram

```mermaid
flowchart TD
    subgraph before["Before: per aura, per entity, per tick"]
        A1["a.value !== 0 ? {value} : {}"] -->|spread| O1["new WireAura literal"]
        A2["a.value2 !== undefined ? {value2} : {}"] -->|spread| O1
        A3["...6 more chained spreads"] -->|spread| O1
        O1 --> G1["up to 9 short-lived object allocations per aura"]
    end

    subgraph after["After: direct assignment"]
        W["const w = {id, name, kind, rem, dur}"] --> C1["if present: w.value = ..."]
        C1 --> C2["...7 more conditional assigns"]
        C2 --> G2["one object allocation per aura"]
    end

    G1 -->|"x 555 entities x 20 Hz"| GC["GC sawtooth: heap 64MiB -> 192MiB"]
    GC --> LAG["event-loop lag p99 ~100ms"]
    LAG --> TICK["sim tick p99 breaches 50ms budget"]
    G2 -->|"~2x less mapper CPU (local microbenchmark)"| DONE["no behavior change, same wire bytes"]
```

## Test plan

- [x] New `tests/wire_aura.test.ts`: pins exact aura wire-shape for defaults-only, all-fields-present, and the stacks=1/charges=1 edge case
- [x] Existing wire-format tests unchanged and passing: `tests/bandwidth.test.ts`, `tests/snapshots.test.ts`, `tests/target_frame.test.ts`, `tests/auras_view.test.ts`, `tests/holder_broadcast.test.ts`
- [x] `npm run ci:changed` (biome) clean, no new errors or format diffs
- [x] `npm run build` succeeds (all 5 entries)
- [x] `npm run gate`, full suite green aside from a pre-existing, order-dependent flaky timeout in `tests/server/http/parity.test.ts` / `characterization.test.ts` (unrelated leaderboard HTTP tests; pass in isolation, and the failure set differs between runs, confirmed not caused by this change)
- [x] Merge-tree against `release/v0.24.0` clean (branch is exactly 1 commit ahead, 0 behind)


No898:
Independent review by Codex (GPT-5.6 Sol Ultra).

I compared base `667227a` with head `91667f4`, traced the serializer through the per-tick JSON cache, and ran the focused wire regression set. I found no production-code correctness issue. The direct assignments preserve every previous omission predicate and the same property insertion order, and representative base/head serialization was byte-for-byte identical.

Focused verification passed: 6 test files and 130 tests. The only minor review concern is test hardening: `toEqual` does not pin JSON property order, and zero-valued defined optionals are not all covered. The currently failing `social.test.ts` passed 53/53 in isolation and does not import the changed server serializer, so the red gate looks like suite-level timeout contention, but it should still be rerun successfully before merge.

Overall: the implementation looks correct; only the regression test could more precisely enforce the byte-identical contract.

jgyy:
Thanks for the thorough independent review. Pushed a86565624 addressing both test-hardening points:

- Pinned aura wire property order with `Object.keys(w)` assertions on both existing shape tests (defaults-only and all-fields-present), so a future reorder of the direct assignments in `wireAura` would fail even though `toEqual` is order-insensitive.
- Added a new case covering `value2`/`value3`/`tickInterval`/`charges` when each is defined as exactly `0` (gated on `!== undefined`, unlike `value`/`sourceId` which are gated on truthiness and legitimately omit `0`). Confirms these do not get accidentally dropped by a future truthiness-style refactor.

`npx vitest run tests/wire_aura.test.ts` passes (4/4), `tsc --noEmit` clean, and the pre-push gate (tsc + guard tests + biome + copy rules) is green.

EnriqueGF:
Thanks for the careful writeup and the microbenchmark :)

Base-branch gate: PASS. This targets release/v0.24.0, the latest release branch.

I verified the change against the real code in a worktree rather than taking the description on trust.

## What I confirmed

- **Byte-identity.** I fuzzed the old spread chain against the new `wireAura` over 400k randomized auras, including the nasty inputs (`-0`, `NaN`, `Infinity`, `undefined`, `stacks: 0`), comparing both `JSON.stringify` output and `Object.keys` order. Zero mismatches. The truthiness-vs-`!== undefined` distinction is preserved exactly.
- The perf premise holds: `wireCacheFor` (server/game.ts:3952) recomputes `dynamicFields(e)` once per entity per tick before the diff, so the allocation really did happen regardless of whether anything changed.
- `npx tsc --noEmit` clean for the changed files, `tests/server` 91 files / 1779 passed, `tests/parity` 164 passed, `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` green. No em dashes, en dashes, or emojis in any added line. Biome reports no errors on the two changed files.
- No wire-format, sim, determinism, rng, i18n, or server-authority surface touched. `wireEntity` was already exported, so no new public surface.

## One thing to fix before merge

1. **(minor, test coverage)** `tests/wire_aura.test.ts:76-90` does not pin the RAW-`value` contract that `server/game.ts:708-713` explicitly claims. Every `value` in the suite is an integer (`-3`), for which `round2(v) === v`. I mutation-tested this: changing the line to `w.value = round2(a.value)` leaves all of `wire_aura`, `auras_view`, and `target_frame` green. That is precisely the regression the comment warns about, since a tiny negative would collapse to `-0`, serialize as `0`, and flip a stat-sap's `isAuraDebuff` classification on the client. One extra case closes it:

```ts
e.auras = [baseAura({ value: -0.004 })];
expect(wireAuras(e)[0].value).toBe(-0.004);
```

For contrast, the pins you added after the earlier review are genuinely decisive: I mutation-tested a `charges` truthiness regression and a `value` key reorder, and each fails the suite as intended.

## Nits, no action strictly required

2. The description says the branch is "exactly 1 commit ahead, 0 behind". It is now 3 ahead and 8 behind `origin/release/v0.24.0`. `git merge-tree` is still clean, so this is only a stale claim, but it is worth a rebase before merge.
3. The root-cause narrative may be overstated. `dynamicFields` is immediately fed to `JSON.stringify` per entity per tick, and that allocates considerably more than the aura spread chain did. This change is a strict improvement with zero behavior change and is worth landing on its own merits, but I would not close the underlying Grafana incident on it without a prod before/after.
4. The PR body itself contains em dashes and an emoji, which the repo copy rule covers for PR text as well as code.

Happy to approve as soon as the RAW-value pin lands. The refactor is correct and the rest of the test file is well built.

jgyy:
Thanks for the mutation-testing catch, pushed 1ff427656 addressing it:

- Added a case pinning `w.value` for a fractional negative (`-0.004`), which would collapse to `-0` -> `0` under a `round2(a.value)` regression and silently flip a stat-sap's `isAuraDebuff` classification on the client. `server/game.ts` was already correct (raw `a.value`, not rounded); this closes the gap you found in the test coverage.

Verified: `npx vitest run tests/wire_aura.test.ts` passes (5/5), `tsc --noEmit` clean, biome clean on the changed test file.

Also noted your nits: rebased locally is not needed for merge-tree cleanliness (still clean), but will rebase before merge per your note, and will fix the PR body's em dashes/emoji in a follow-up edit.

FernandoX7 (commented):
Confirming this one from my own read after No898 and EnriqueGF, since both of their concerns were substantive. The refactor is correct and both earlier findings landed as decisive pins.

## Verified independently

- **Byte-identity holds.** I fuzzed the old spread chain against the new `wireAura` over 500k randomized auras, including the awkward inputs (`-0`, `NaN`, `Infinity`, `-0.004`, `undefined`, `stacks: 0`), comparing both `JSON.stringify` output and `Object.keys` order. Zero mismatches. Same omission predicates, same insertion order, and the truthiness (`value`/`sourceId`) vs `!== undefined` (`value2`/`value3`/`tickInterval`/`charges`) split is preserved exactly.
- **Perf premise checks out.** `wireCacheFor` (server/game.ts:3866, :3956) computes at most once per tick per entity, then unconditionally `JSON.stringify(dynamicFields(e))`, which maps every aura through `wireAura`. So the per-branch object literals really were allocated per aura, per entity, per tick before the diff could elide anything.
- **The pins are decisive.** I mutation-tested both prior findings: swapping `w.value = round2(a.value)` fails the raw-value case, and reordering the `rem`/`dur` assignments fails both key-order assertions. `tests/wire_aura.test.ts` passes 5/5.

## Comments addressed

- No898's key-order + zero-valued-optional gaps: closed by a86565624, both verified above.
- EnriqueGF's raw-value pin (the merge blocker): closed by 1ff427656, verified above.
- The PR-body copy nit: the body is clean now (no em/en dashes, no emoji).

## One open item, non-blocking

The branch is now 19 behind, 4 ahead of `release/v0.24.0` (further stale than the earlier notes). `git merge-tree` is still clean so it will merge, but a rebase before merge is still worth it, as you already noted.

Good change. Pure server-side serialization, no wire-format, sim, determinism, or authority surface touched, and the test file is well built.

FernandoX7 (approved):

PR #1749, opened 2026-07-10
fix(ui): honor minimum ability range on action bar

No898:
## Summary

- Mark targeted abilities as out of range when the target is closer than the ability's `minRange`.
- Keep both exact boundaries usable: Onrush is available at 8 yd and 25 yd, but unavailable at 5 yd and 26 yd.
- Add a regression test covering the minimum and maximum range boundaries.

The simulation already rejected Onrush below its 8 yd minimum with `Too close!`, but the action bar only compared target distance with the maximum range. This made the button look usable even though activation failed. The view now applies the same minimum-range rule as the simulation.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/action_bar_view.test.ts tests/action_bar_painter.test.ts` (30 tests passed)
  - `npm run gate` (all 9 steps green; 11,436 tests passed and 14 skipped)
  - Repository pre-push QA hook (typecheck, 53 guard tests, changed-file lint, and copy rules passed)
- Manual steps:
  - Started an offline level 4 warrior and placed Onrush on the action bar.
  - Targeted a rooted Forest Wolf at 5 yd and verified Onrush receives the existing out-of-range presentation.
  - Verified the same button is usable at 10 yd and out of range at 30 yd.
  - Checked desktop at 1440x900, mobile landscape at 844x390, and mobile portrait at 390x844. The portrait action button remained visible at 54x54 px.
  - The before captures recreate only the previous missing Onrush out-of-range presentation while keeping the same 5 yd game setup.

### Desktop

| Before | After |
| --- | --- |
| Desktop before | Desktop after |

### Mobile landscape

| Before | After |
| --- | --- |
| Mobile before | Mobile after |

---

### Quality

### Cross-platform

- [x] **Accessible.** This reuses the existing out-of-range state; button keyboard behavior,
      focus styling, and ARIA labels are unchanged.

### Localization (i18n)

### Hygiene


jgyy:
its just 2 lines change to check more range condition

PR #1752, opened 2026-07-10
fix(sim): give the moored fishing-dock rowboat a collider

daxdax89:
## Summary

The moored rowboat that `src/render/props.ts` draws beside each fishing-dock (dock-local `2.4, -5.0`, a ~2.4u hull) had no collider, so the player walked straight through its hull while every other structure blocks. `src/sim/colliders.ts` emitted only the dock hut OBB and never covered the boat.

This adds a rotation-safe **circle** collider at the boat's world position in the `PROPS.docks` loop. A circle rather than an OBB because `props.ts` jitters the boat heading per dock, so a heading-invariant shape covers the hull whichever way it is moored; `BOAT_RADIUS` conservatively spans the hull length. Both docks share the geometry, so this covers zone1 and zone2. It is `camGhost` like the neighbouring dock props (blocks movement, the chase cam passes through).

This is the follow-up to the plank-deck collider fix (#1573): reviewing that change, @zfreiberg confirmed the deck reads solid but flagged that the rowboat next to it was still walk-through (its local position falls outside the deck's footprint). The boat is the same class of gap as the pier, on a different object, so it gets its own collider here.

## Related issues

Part of #1500 (dock collision). Complements #1573.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/rowboat_collision.test.ts` (new; both docks)
  - `npx vitest run tests/architecture.test.ts` (sim purity guard)
  - `npx tsc --noEmit` (clean for the changed files)
- New test `tests/rowboat_collision.test.ts` pins that the hull is now blocked at both docks (it was walk-through before) and that the circle footprint stays bounded (water past the hull remains walkable, so it does not over-block the bank).

## Screenshots / recordings

N/A: this is a collision-only change (no rendered pixels change; the boat already renders). The walk-through it fixes is documented in @zfreiberg's review screenshots on #1573.

---

### Quality

- [x] Decisive tests added for the changed behavior; determinism / sim-purity guard run.

### DIFF
--- a/src/sim/colliders.ts
+++ b/src/sim/colliders.ts
@@ -80,6 +80,15 @@ function rotY(lx: number, lz: number, rot: number): { x: number; z: number } {
   return { x: lx * c + lz * s, z: -lx * s + lz * c };
 }
 
+// The moored rowboat props.ts draws beside each fishing-dock (dock-local, a
+// ~2.4u hull) had no collider, so the player walked straight through it. A
+// circle rather than an OBB, because props.ts jitters the boat heading per dock:
+// a heading-invariant shape covers the hull whichever way it is moored. The
+// radius conservatively spans the hull length.
+const BOAT_LOCAL_X = 2.4;
+const BOAT_LOCAL_Z = -5.0;
+const BOAT_RADIUS = 1.1;
+
 // ---------------------------------------------------------------------------
 // Collider sets
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


daxdax89:
Closing as superseded: the #1839 release batch already landed a fix for #1500 (the moored rowboat/dock collision) on release/v0.26.0, so this PR is now redundant. Thanks for building it out over there. Glad the issue is resolved either way.

PR #1758, opened 2026-07-10
fix(render): restore default skin previews

No898:
## Summary

- Restore the embedded default character texture after previewing or selecting an alternate skin.
- Preserve each mesh's original assembled material and derive every skin variant from that source. Previously, skin changes derived from the currently displayed variant, so skin 0 had no default map left to restore.
- Add decisive regression coverage for both single-material and multi-material meshes.
- Regenerate the target branch's stale i18n summary and resolved hash artifacts as required by the contribution gate.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/appearance_skin.test.ts tests/preview_appearance.test.ts tests/client_shell.test.ts --reporter=dot` (92 passed)
  - `npx vitest run tests/i18n_resolved_equivalence.test.ts tests/fixes.test.ts tests/arena.test.ts tests/mail.test.ts --reporter=dot --maxWorkers=1` (142 passed)
  - `npm run ci:changed` (passed; existing warnings only)
  - `npx tsc --noEmit` (passed)
  - `npm run build:env` (passed)
  - `npm run build:server` (passed)
  - `npm run build` (passed)
  - `npm run gate` (passed all 9 steps; 11,495 tests passed and 14 skipped)
- Manual steps:
  - Opened the offline hero picker, selected Warrior, previewed Chroma 4, then returned to Chroma 1 through hover and click.
  - Confirmed the embedded grey default texture is restored instead of retaining the red alternate texture.
  - Repeated at 1440x1000, 390x844, and 844x390 viewports.

## Screenshots / recordings

| Viewport | Before | After |
| --- | --- | --- |
| Desktop | Default swatch retains alternate texture before the fix | Default swatch restores the embedded texture after the fix |
| Mobile portrait | Mobile portrait before the fix | Mobile portrait after the fix |
| Mobile landscape | Mobile landscape before the fix | Mobile landscape after the fix |

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green, including the full test suite, typecheck, and all builds.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified desktop, phone portrait, and phone landscape viewports.
- [x] **Accessible.** No interaction markup, focus behavior, ARIA, animation, or touch target was changed.

### Localization (i18n)

### Hygiene

- [x] I did not hand-edit generated files. The i18n artifacts were regenerated with their canonical scripts.


FernandoX7 (commented):
Solid fix, and you landed on the real root cause. The variant was being re-derived from whatever was currently on `mesh.material`, so once an alternate atlas overwrote `.map`, skin 0's "no override" path had nothing left to restore. Capturing the assembled source material per mesh and always deriving from it is the right call, and it also stops the tint from compounding across repeated previews (the old path re-tinted the last variant every call).

I traced the lifecycle to confirm the WeakMap keying is safe: `assembleModel` produces fresh mesh instances per `CharacterVisual`, and `applyMaterials` is always first invoked in the constructor before any swap, so the captured source is always the embedded default. `setHeldWeapon` swaps in fresh prop meshes (new instances, new WeakMap entries), so body sources never go stale, and `tintedMaterial` clones and never mutates `src`, so the shared source stays pristine. No purity or seam concerns: render-only, reads the source material.

The regression test is decisive. I reverted just the `assets.ts` change and reran it: it fails at the restore assertion (map stays the alternate texture), and passes with the fix. That is the failing-first coverage this kind of bug wants.

nit (`tests/appearance_skin.test.ts:89`): the test only exercises the single-material branch. The `Array.isArray(source)` path at `assets.ts:676` (multi-material meshes) is symmetric so I am not worried about it, but a second case over an array material would close the gap.

The i18n artifact changes (`i18n.resolved.sha256`, `i18n.status.summary.json`) are the expected mechanical regen against the stale base, and the head's `sha256` is a clean single line. No concern.

No898:
@FernandoX7 Added the multi-material regression case in 07f312491. It verifies that both material slots receive the alternate texture and then restore their original embedded maps. The full local gate is green: all 9 steps passed, with 11,495 tests passed and 14 skipped. The PR is ready to merge.

FernandoX7 (approved):

PR #1767, opened 2026-07-10
fix(dungeons): lock the whole heroic party on the final-boss kill

Rubsey:
## Summary

Closes the heroic re-run abuse reported on Discord ("if someone doesn't loot the boss gem, the whole party can re-enter"). The unlooted gem was a correlate, not the cause: the daily lockout was already granted on the kill, but only to party members within 80yd of the corpse, and the five-man door only lockout-checked the player claiming a FRESH instance. One unlocked member (a door camper, an early-released ghost, or a freshly recruited player) could claim a new heroic instance and the whole locked party walked in behind them for another set of heroic epics. The Nythraxis raid had the same range hole (260yd room radius). Verified byte-identical on the deployed PBE SHA (bfc229787).

## Changes

- New grantHeroicKillLockout (src/sim/instances/dungeons.ts), called unconditionally from the death hub beside grantNythraxisLockout: a heroic claim's final-boss kill locks every CURRENT member of the owning group (wherever they stand, even never zoned in) plus any player physically inside the instance, credit or no credit. The lockout no longer rides awardHeroicMarks; marks stay participation-gated.
- grantNythraxisLockout (src/sim/encounters/nythraxis.ts) now locks the whole owning raid group via the same instanceLockoutMetas sweep, both difficulties; the old room-radius rule survives only as a no-claim fallback.
- New enterDungeon door gate: a locked player may enter a LIVE heroic claim only when its final boss is already down, so corpse-runs and loot retrieval into a cleared run still work but a locked party can no longer ride an unlocked recruit into a fresh run. Reuses the existing locked error string, no new i18n keys.
- Extracted instanceContains (the shared 120/250 instance footprint envelope) and wired the new SimContext callback (declaration, host passthrough, late-bound arrow, fake-ctx updates).

## Test plan

- [x] RED first: 4 new five-man tests + 1 raid test reproduced the abuse before the fix (camper not locked, party-leaver not locked, recruit ferry allowed, raid fills not locked)
- [x] Whole-party lock including a never-entered member, with the marks-stay-participation-gated pin
- [x] Left-party-but-still-inside member locked (position arm)
- [x] Locked party cannot ride an unlocked recruit's fresh claim (door gate, exact error string pinned)
- [x] Locked player still re-enters the CLEARED live claim (corpse-run / loot pin)
- [x] Uncredited final-boss death still locks the owning party (no marks, no credit)
- [x] Normal final-boss and heroic trash kills lock nobody; :heroic key never leaks onto the plain key
- [x] Whole-raid lock with exact reset timestamp, fills asserted outside the arena
- [x] npm run gate green on the final diff (all 9 steps); parity goldens unchanged (lockout paths draw no rng)
- [x] Reviewed by architecture-reviewer, test-coverage-auditor, cross-platform-sync and qa-checklist agents: zero blocking findings, all coverage gaps closed

## Notes for deploy

- PBE runs release/v0.24.0-ptr: this fix must be merged or cherry-picked there to stop the live abuse.
- Accepted residual: a member fully OFFLINE at kill time cannot be personally locked (no live session), but the door gate keeps the locked rest of the party out, so it degrades to a legitimate fresh run rather than a ferry.
- Benign redundancy: a heroic Nythraxis kill writes the same lockout key through both granters (idempotent, same value).

### DIFF
--- a/src/sim/encounters/nythraxis.ts
+++ b/src/sim/encounters/nythraxis.ts
@@ -29,7 +29,7 @@ import { isStunned } from '../combat/cc';
 import { ITEMS, MOBS, NPCS, QUESTS } from '../data';
 import { createMob, createNpc } from '../entity';
 import { applyHeroicMobTuning, mobTemplateForDungeonDifficulty } from '../instances/difficulty';
-import { heroicLockoutId } from '../instances/dungeons';
+import { heroicLockoutId, instanceLockoutMetas } from '../instances/dungeons';
 import type { PlayerMeta } from '../sim';
 import type { SimContext } from '../sim_context';
 import { clearThreat, threatEntries } from '../threat';
@@ -506,7 +506,14 @@ export function grantNythraxisLockout(ctx: SimContext, boss: Entity): void {
   const lockId = isHeroicNythraxis(ctx, boss)
     ? heroicLockoutId('nythraxis_boss_arena')
     : 'nythraxis_boss_arena';
-  for (const meta of nythraxisRoomMetas(ctx, boss)) {
+  // The kill locks the WHOLE owning raid group plus anyone inside the arena
+  // (instanceLockoutMetas), not just the players standing in the boss room: a
+  // raider who released, camped the entrance, or never zoned in must not stay
+  // unlocked, or one unlocked member re-claims the arena for the locked raid.
+  // Room-radius survives only as the fallback for a boss outside any claim.
+  const inst = ctx.instances.find((i) => i.partyKey !== null && i.mobIds.includes(boss.id));
+  const metas = inst ? instanceLockoutMetas(ctx, inst) : nythraxisRoomMetas(ctx, boss);
+  for (const meta of metas) {
     meta.raidLockouts.set(lockId, until);
   }
 }
--- a/src/sim/encounters/nythraxis.ts
+++ b/src/sim/encounters/nythraxis.ts
@@ -506,14 +506,22 @@ export function grantNythraxisLockout(ctx: SimContext, boss: Entity): void {
   const lockId = isHeroicNythraxis(ctx, boss)
     ? heroicLockoutId('nythraxis_boss_arena')
     : 'nythraxis_boss_arena';
-  // The kill locks the WHOLE owning raid group plus anyone inside the arena
-  // (instanceLockoutMetas), not just the players standing in the boss room: a
-  // raider who released, camped the entrance, or never zoned in must not stay
-  // unlocked, or one unlocked member re-claims the arena for the locked raid.
-  // Room-radius survives only as the fallback for a boss outside any claim.
+  // The kill locks the UNION of the room and the claim sweep. The claim sweep
+  // (instanceLockoutMetas) covers the whole owning raid group plus anyone
+  // inside the generic instance footprint: a raider who released, camped the
+  // entrance, or never zoned in must not stay unlocked, or one unlocked member
+  // re-claims the arena for the locked raid. The room metas stay in the union
+  // because the arena interior is WIDER than the generic 120-yd footprint
+  // (walls at roughly +/-230 local x): a raider who left the raid while parked
+  // in a side wing sits outside both claim arms yet can still hold the tap and
+  // its rewards, so the 260-yd boss room must keep locking them.
+  const metas = new Map<number, PlayerMeta>();
+  for (const meta of nythraxisRoomMetas(ctx, boss)) metas.set(meta.entityId, meta);
   const inst = ctx.instances.find((i) => i.partyKey !== null && i.mobIds.includes(boss.id));
-  const metas = inst ? instanceLockoutMetas(ctx, inst) : nythraxisRoomMetas(ctx, boss);
-  for (const meta of metas) {
+  if (inst) {
+    for (const meta of instanceLockoutMetas(ctx, inst)) metas.set(meta.entityId, meta);
+  }
+  for (const meta of metas.values()) {
     meta.raidLockouts.set(lockId, until);
   }
 }
--- a/src/sim/instances/dungeons.ts
+++ b/src/sim/instances/dungeons.ts
@@ -148,16 +148,19 @@ export function enterDungeon(ctx: SimContext, dungeonId: string, pid?: number): 
     }
   }
   // A locked player may walk back into a LIVE heroic claim only when its final
-  // boss is already down (the corpse-run / loot-retrieval path the kill lockout
-  // deliberately leaves open). While that boss is alive, the lockout bars the
-  // door: without this, one unlocked member (a fresh recruit, or a camper the
-  // kill never locked) could claim a fresh heroic instance and ferry the whole
-  // locked party into another full run.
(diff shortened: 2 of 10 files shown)


src/sim/instances/dungeons.ts
```diff
@@ -328,14 +364,53 @@ function freeInstance(ctx: SimContext, inst: InstanceSlot): void {
...
+// same realm-local boundary the Nythraxis raid uses), scoped to the :heroic
+// key so the normal difficulty is never consumed. Marks stay participation-
+// gated below; the lockout deliberately is not.
+export function grantHeroicKillLockout(ctx: SimContext, mob: Entity): void {
+  const inst = ctx.instances.find((i) => i.partyKey !== null && i.mobIds.includes(mob.id));
+  if (!inst || inst.difficulty !== 'heroic') return;
+  const tuning = HEROIC_DUNGEON_TUNING[inst.dungeonId];
+  if (!tuning || mob.templateId !== tuning.finalBossId) return;
+  const lockedUntil = ctx.raidResetMs(ctx.lockoutNowMs());
+  for (const meta of instanceLockoutMetas(ctx, inst)) {
```
Rubsey, line 424: High: this sweep must also lock the death-time reward recipients. `grantHeroicKillLockout` runs before `handleDeath` constructs `eligible`, while `instanceLockoutMetas` only sees current party identity and physical position. Repro: a runner first-taps Morthen, leaves the party and instance, and the remaining party kills him. The persistent tap gives the runner both heroic epics and the Heroic Mark, but the runner remains unlocked. They can rejoin to loot, leave again, then claim and loot a fresh solo heroic; the same character went from 2 to 4 heroic epics. Please union the tapper/death-time eligible recipients with the current-group/occupant sweep and add a tapper-leaves-before-kill regression test.

src/sim/instances/dungeons.ts
```diff
@@ -141,10 +147,26 @@ export function enterDungeon(ctx: SimContext, dungeonId: string, pid?: number):
...
+  // boss is already down (the corpse-run / loot-retrieval path the kill lockout
+  // deliberately leaves open). While that boss is alive, the lockout bars the
+  // door: without this, one unlocked member (a fresh recruit, or a camper the
+  // kill never locked) could claim a fresh heroic instance and ferry the whole
+  // locked party into another full run.
+  if (
+    inst &&
+    inst.difficulty === 'heroic' &&
+    isRaidLocked(ctx, r.meta, heroicLockoutId(dungeonId)) &&
+    heroicFinalBossAlive(ctx, inst)
```
Rubsey: High: `final boss is dead` is not sufficient to identify the cleared claim that produced this player's lockout. Repro: A and B clear a heroic and become locked; after it frees, unlocked recruit C joins their party, claims a fresh heroic, and kills Morthen while A/B wait outside. Once the boss is dead this gate admits A/B. `corpseLootRights` then uses the tapper's current party, so A can take the non-personal heroic epics even though `lootRecipientIds` contains only C. Tie the re-entry exception to claim-specific participation or the death-time loot snapshot, rather than any live claim with a dead final boss.

src/sim/encounters/nythraxis.ts
```diff
@@ -506,7 +506,14 @@ export function grantNythraxisLockout(ctx: SimContext, boss: Entity): void {
...
     ? heroicLockoutId('nythraxis_boss_arena')
     : 'nythraxis_boss_arena';
-  for (const meta of nythraxisRoomMetas(ctx, boss)) {
+  // The kill locks the WHOLE owning raid group plus anyone inside the arena
+  // (instanceLockoutMetas), not just the players standing in the boss room: a
+  // raider who released, camped the entrance, or never zoned in must not stay
+  // unlocked, or one unlocked member re-claims the arena for the locked raid.
+  // Room-radius survives only as the fallback for a boss outside any claim.
+  const inst = ctx.instances.find((i) => i.partyKey !== null && i.mobIds.includes(boss.id));
+  const metas = inst ? instanceLockoutMetas(ctx, inst) : nythraxisRoomMetas(ctx, boss);
```
Rubsey: High: `instanceLockoutMetas` uses the generic `abs(x - origin.x) < 120` footprint, but the legitimate Nythraxis arena reaches walls at local x = +/-230. Replacing `nythraxisRoomMetas` means a raider at local x = 200 who leaves the raid is physically inside the arena but excluded from the lockout. A persistent first tap still grants that player the boss rewards, leaving them rewarded and unlocked. Union the old room metas even for claimed runs, or make the footprint interior-aware, and add a side-wing leaver test.

Rubsey (commented):
Changes requested for three remaining lockout bypasses. GitHub does not permit the PR author to submit a formal request-changes review. The full local gate is green (11,513 tests plus typecheck and builds), but the adversarial scenarios in the inline comments still allow repeat heroic rewards or an unlocked rewarded raider.

src/sim/instances/dungeons.ts
Rubsey: Accurate, fixed in fd9b6a140. Verified the premise: tappedById persists through a party leave and handleDeath credits the tapper (creditId = tappedById ?? killer), so the departed runner really did collect the mark slot and corpse loot rights while escaping both sweep arms. Death-time reward recipients are now the third arm of the lockout union: awardHeroicMarks stamps every recipient through the shared lockToHeroicClaim helper, before its marks daily gate, so an already-marked recipient is still locked. Regression test 'a tap-runner who left the party and the instance is still locked by the kill' reproduces the exact scenario (tap, leave party, leave instance, party kills) and also pins that the runner can rejoin and re-enter the cleared claim to collect the mark.

src/sim/instances/dungeons.ts
Rubsey: Accurate, fixed in fd9b6a140. Confirmed corpseLootRights grants shared rights via the tapper's CURRENT party, so the open door would indeed hand A the epics. The re-entry exception is now tied to claim-specific participation: InstanceSlot records clearedBy, the set of players whose lock FIRST landed with this claim's kill (populated by lockToHeroicClaim across all three sweep arms, reset in claimInstance and freeInstance), and the door admits a locked player only when the final boss is down AND they are in that set. A player locked by an earlier run is barred from someone else's cleared claim; the corpse-run and tap-runner loot paths keep working because their lock came from that claim. Regression test 'a locked player cannot enter a clear they took no part in, even after its boss dies' reproduces the recruit-solo-kill scenario.

src/sim/encounters/nythraxis.ts
Rubsey: Accurate, fixed in fd9b6a140. The arena interior (walls near +/-230 local x, tank positions at +150 in existing tests) is indeed wider than the generic 120yd footprint, so a raid-leaver in a wing escaped both claim arms while able to hold the tap. grantNythraxisLockout now locks the UNION of nythraxisRoomMetas (the 260yd boss room) and the claim sweep, for claimed and unclaimed bosses alike, deduped by entityId. Regression test 'a raider who left the raid while parked in a side wing is still locked by the kill' pins the geometry premises in-test (|x - origin.x| > 120 and dist to spawn <= 260) before asserting the lock.

Issue #1659, opened 2026-07-08
Party, guild, and other channel chat is only visible in the chat window


MasterZensei:
## Problem

Overhead chat bubbles currently render only for say, yell, and emote messages (the chat event handler in `src/ui/hud.ts` routes exactly those three channels to `Renderer.showChatBubble`). Messages on every other channel (party, guild, officer, general, world, lfg, whisper) appear only in the chat window.

During combat, or whenever the chat log is busy or scrolled, players miss party and guild coordination from teammates standing right next to them. The chat window should not be the only place those messages are visible.

## Expected behavior

- A message from a nearby player on the other chat channels also renders as an overhead bubble above the speaker's character, the same way say and yell do today.
- Each channel's bubble is tinted with that channel's color (reusing the existing chat-log channel palette) so the channel is recognizable at a glance: party, guild, officer, and so on each look distinct.
- Say, yell, and emote bubbles keep their current look and behavior.
- Bubbles only appear when the speaker's entity is actually in view; a message from a distant sender (for example a guild member in another zone) still lands only in the chat window, with no bubble and no error.

## Scope

- `src/render/renderer.ts`: `showChatBubble(entityId, text, yell)` only distinguishes yell from say; extend it to take the channel (or a style descriptor) so bubbles can be tinted per channel.
- `src/ui/hud.ts`: the chat-event handler that gates bubbles to say, yell, and emote; widen the gate and pass the channel through.
- Channel colors come from the existing chat-log channel palette (CSS tokens), not new hardcoded values.
- Decide which channels participate. Party, guild, and officer are the clear wins; whether general, world, lfg, and whisper should bubble too (or would be too noisy) is open. 

## Acceptance criteria

- A party message from a party member standing nearby shows a bubble above their character in the party channel color.
- A guild message from a nearby guild member shows a bubble in the guild channel color.
- Say, yell, and emote bubbles are unchanged.
- Bubble text is still set via `textContent` (player text is never interpreted as HTML), matching the current implementation.
- A message whose sender has no visible entity (out of interest range, other zone, offline sender in a standing channel) produces no bubble and no console error.
- The desktop and mobile chat window experience is unchanged; bubbles are additive.

## Notes

- Bubbles are client-side presentation only: each client raises bubbles from messages it already received, so channel privacy (guild, officer, whisper) is preserved automatically; nobody sees a bubble for a message they could not read in their own chat log.

PR #1770, opened 2026-07-10
feat(ui): overhead chat bubbles for party channel messages (#1659)

daxdax89:
## Summary

Party chat now raises an overhead speech bubble above a nearby teammate, the same way say and yell do today (#1659), so party coordination is visible in the world during combat and not only in the chat window.

- A new pure module (`chat_bubble_style.ts`, in the `UI_PURE_CORES` allowlist) maps a channel to its bubble style and is the single source both the HUD bubble gate and the renderer read, so a channel's chat-log colour and its bubble tint cannot drift.
- The party bubble tints the BORDER (matching the chat-log party colour `#7fd4ff`), keeping the near-white background and dark text for legibility. Say, yell, and emote bubbles stay byte-identical.
- `showChatBubble` takes an optional style descriptor, so the two existing 3-arg boolean callers are unchanged.
- Client-side presentation only: no sim, server, or wire change. Party works because its emit already sets `fromPid` = the speaker entity, and bubbles appear only when that entity is in view (distant or other-zone senders produce no bubble and no error, unchanged).

### Scope: party only, deliberately

The issue also lists guild and officer. Those are server-side social broadcasts (`server/social.ts`) that carry no speaker id, so the client has no entity to anchor a bubble to. Bubbling them cleanly needs the server to attach the speaker's id to those events and carry it over the social wire, which is a server + wire change out of scope for this client-only PR. I have kept this PR to party (the strongest combat-coordination win) and left guild/officer as a follow-up. general/world/lfg/whisper/roll stay non-bubbling (too noisy or private). Happy to do the guild/officer server-side follow-up if you would like it, and a per-player toggle is a possible later addition.

## Related issues

Part of #1659 (party channel; guild/officer follow-up noted above).

## Type of change

## How was this tested?

- Commands: `npx tsc --noEmit`; `npx vitest run tests/chat_bubble_style.test.ts tests/architecture.test.ts tests/chat.test.ts tests/chat_log.test.ts` (green); `biome check` on changed files (0 errors).
- New tests (`tests/chat_bubble_style.test.ts`): the pure map tints party's border and leaves say/yell/emote on their existing treatment; general/world/lfg/whisper/roll/guild/officer never bubble; a Sim-driven test pins that a party chat event carries `fromPid` = the speaker entity (the client bubble anchor) with `entityId` unset; and source guards that the gate routes through the shared style module and the renderer tints the border inline.

## Screenshots / recordings

Before (party line only in chat) and after (a nearby party message also bubbles, tinted with the party colour):

party chat bubbles, before and after


Rubsey:

Thanks for the contribution, @daxdax89.

PR #1462, opened 2026-07-05
feat(sfx): add lockpick minigame audio (14 custom clips)

jamiecypher:
## Summary

Adds 14 custom 320 kbps MP3 sound effects for the lockpick minigame (Collapsed Reliquary, and any future delves that adopt the mechanic).

This is the first non-ElevenLabs audio contribution, so the SFX pipeline is extended to support custom recordings alongside generated clips:

- `scripts/sfx/sfx_prompts.mjs` — new `custom: true` flag on entries that have hand-recorded files. No prompt or duration needed. Comment added explaining the pattern for future contributors.
- `scripts/gen_sfx.mjs` — generator skips API call for `custom: true` entries even with `--force`, so custom recordings are never overwritten. API key guard moved so manifest rebuild runs without a key (allows offline contributors to add custom clips without needing ElevenLabs access).
- `src/game/sfx_manifest.generated.ts` — 14 new key-to-path entries.
- `src/ui/hud.ts` — `sfx.playUi()` wired to all lockpick sim events. The four advanced variants (pin setting clicks) pick randomly without immediate repeats.
- `CREDITS.md` — new Audio section added (@jamiecypher, CC0 samples from freesound.org, edited in Reaper).

**Note:** these sounds were **not** present in v0.24.0. This PR was never merged before that release and the files are absent from that branch. This rebase targets `release/v0.25.0` as the first submission.

To add a custom recording in the future: drop the MP3 in `public/audio/sfx/`, add `{ key: 'your_key', custom: true }` to the SFX array in `sfx_prompts.mjs`, run `node scripts/gen_sfx.mjs` to rebuild the manifest (no API key needed if all existing clips are already on disk), then wire it up.

## Sound map

| Event | Result | Sound |
|---|---|---|
| `lockpickSession` | — | `lockpick_begin` |
| `lockpickStep` | `advanced` | `lockpick_advanced_1–4` (random, no repeat) |
| `lockpickStep` | `slip` | `lockpick_slip` |
| `lockpickStep` | `bind` | `lockpick_bind` |
| `lockpickStep` | `trap` | `lockpick_trap` |
| `lockpickStep` | `pageCleared` | `lockpick_page_cleared` |
| `lockpickStep` | `retry` | `lockpick_retry` |
| `lockpickStep` | `success` | `lockpick_success` |
| `lockpickStep` | `fail` | `lockpick_fail` |
| `lockpickEnd` | `success` | `lockpick_end` |
[...]

PR #1771, opened 2026-07-11
feat(sfx): variant naming system / manifest rebuild mode

jamiecypher:
## Summary

Implements the SFX variant naming system from #1729. Contributors can drop multiple recordings of the same sound under a numbered convention (foot_stone_1.mp3, foot_stone_2.mp3, ...) and the engine groups them under the base key, cycling through them at runtime using no-repeat random selection so the same clip never plays twice in a row.

This revision also addresses all reviewer feedback: the dual-map bug, first-play bias in nextBuffer(), non-MP3 file support in the manifest builder, and the process.exit(0) override.

## What changed

**scripts/sfx/sfx_manifest_builder.mjs (new):** Extracted buildManifest() into its own testable module so tests can import it without triggering gen_sfx.mjs top-level API-key / process.exit logic. Now probes .mp3, .wav, .flac, .ogg when looking for a bare key, so a custom recording committed as a WAV is not silently dropped on a manifest rebuild.

**scripts/gen_sfx.mjs:** Imports buildManifest from the new module. The --manifest path now exits 1 if the builder reports errors instead of always calling process.exit(0) (which previously overrode a process.exitCode=1 set by the error path).

**src/game/sfx.ts:**
- Unified this.buffers and this.variants into a single map (this.variants). Procedural buffers (amb_crowd, vcup_crowd_roar) previously only went into this.buffers, making them unreachable via playUi() / nextBuffer() which only read this.variants. They now go into this.variants as single-element arrays and crowdRoar() works correctly.
- Fixed nextBuffer() first-play bias: the old code picked floor(random * (pool.length - 1)) then always bumped idx++ because idx >= last with last=-1 was always true. Variant 0 could never be selected on first play. Fix: when last < 0 (first play), pick uniformly over the full pool with floor(random * pool.length).
- loop() reads this.variants.get(key)?.[0] (always variant 0, intentional for seamless looping beds).

**tests/sfx.test.ts:** Updated buffer injection to use this.variants (arrays). Added 3 tests for nextBuffer() selection: variant 0 reachable on first play, no-repeat on a 2-clip pool, and single-clip pool always returns the same buffer.

**tests/sfx_manifest.test.ts (new):** 6 tests covering buildManifest() using real temp directories: .wav key survives rebuild, .mp3 key works, .mp3 preferred over .wav when both exist, numbered variants grouped, missing key omitted, loop flag honoured.

## Related issues

Part of #1729

## Type of change

- [x] Bug fix: non-breaking change which fixes an issue
- [x] Build, CI, or tooling

## How was this tested?

- npx vitest run tests/sfx.test.ts tests/sfx_manifest.test.ts: 13/13 pass (was 3/4 before)
- npx tsc --noEmit: clean
- biome check on all changed files: no errors

---

## Checklist

**Quality**

- [x] Tests pass. npm test is green. I added or updated tests for any sim/ or server/ behavior I changed.
- [x] Typecheck passes. npx tsc --noEmit is clean.
- [ ] Builds succeed. npm run build

**Cross-platform**

- N/A: audio pipeline, no UI change.

**Localization (i18n)**

**Hygiene**

- [x] No secrets, credentials, or .env committed, and ALLOW_DEV_COMMANDS is not enabled in any production path.
- [x] sfx_manifest.generated.ts is regenerated via npm run sfx:manifest. Committed state reflects the new generator output.


EnriqueGF:
Thanks for the contribution :) The variant idea is a good one and the generator side is clean, but as it stands the change silently removes two sounds that ship today, so I cannot approve it yet.

Base branch: correct (release/v0.24.0, the latest release branch). Typecheck (npx tsc --noEmit) is clean on the branch.

## Blocking

**1. The regenerated manifest drops a live clip (`src/game/sfx_manifest.generated.ts`).**
The base manifest has 97 keys, this one has 96: `cast_lightning_bolt` is gone. The clip on disk is `public/audio/sfx/cast_lightning_bolt.wav`, and `buildManifest()` in `scripts/gen_sfx.mjs` only probes `${key}.mp3`, and only for keys present in the `SFX` table, so the .wav entry is deleted on rebuild. That key is live: `src/ui/hud.ts:774` maps `lightning_bolt` to `cast_lightning_bolt`, and `src/ui/hud.ts:8316` feeds it to `sfx.loop(...)`. After this PR the manifest has no entry, preload never loads it, and the Lightning Bolt cast loop goes silent in game. The generator needs to probe the extensions actually on disk (or scan the sfx directory instead of only iterating the `SFX` table) so a rebuild cannot delete a shipped clip.

**2. The Vale Cup goal roar goes silent (`src/game/sfx.ts`).**
`playAt()` and `playUi()` now read from `this.variants`, but the two procedurally synthesized clips are only written into `this.buffers` (`sfx.ts:119-120`: `amb_crowd`, `vcup_crowd_roar`). `crowdRoar()` (`sfx.ts:520`) calls `playUi("vcup_crowd_roar", ...)`, so `nextBuffer()` returns undefined and the roar never plays. `amb_crowd` survives only because it goes through `loop()`, which still reads `buffers`. Either register the procedural buffers into `variants` as single-element pools, or have `nextBuffer()` fall back to `this.buffers.get(key)` when there is no pool.

**3. The no-repeat draw is wrong on the first play (`src/game/sfx.ts`, `nextBuffer()`).**
With `last = -1`, the condition `if (idx >= last) idx++` is always true, so `idx` lands in [1, pool.length - 1] and variant index 0 can never be selected as the first play of a session. Guard the first draw: when `last < 0`, pick uniformly over the full pool (`Math.floor(Math.random() * pool.length)`).

## Non-blocking

**4. No tests.** `nextBuffer()` is a self-contained selection rule (no-repeat, single-element passthrough, first-play uniformity) and the grouping in `buildManifest()` is pure file-name logic. Both are cheaply unit-testable, and a single test would have caught all three bugs above. Please add coverage for the selection rule and for the "a key whose clip is not a .mp3 survives a rebuild" case.

**5. `loop()` still reads `this.buffers`,** so a looping key with variants will always use variant 1. That is probably intentional for seamless beds, but a one-line comment would stop the next reader treating it as an oversight.

**6. The PR description contains an em dash.** The repo forbids em and en dashes and emojis in code, docs, commits, and PR text.

The `--manifest` / `npm run sfx:manifest` mode itself is a good addition and the numbered-variant convention is the right shape. Once the manifest rebuild is non-destructive, the procedural keys are reachable again, the first-play draw is uniform, and a test pins all three, I am happy to re-review.

jamiecypher:
@EnriqueGF addressed all the blocking findings from your review:

**Dual-map bug** -- unified this.buffers and this.variants into a single map. Procedural buffers (vcup_crowd_roar, amb_crowd) now go into this.variants as single-element arrays, so crowdRoar() and ambience() both reach them correctly via nextBuffer() and loop(). loop() explicitly reads variant[0] with a comment noting this is intentional for seamless looping beds.

**nextBuffer() first-play bias** -- the old guard (idx >= last with last = -1) was always true, permanently excluding variant 0. Fixed: when last < 0 (first play), pick uniformly over the full pool with floor(random * pool.length). Subsequent plays still use the no-repeat shuffle.

**Manifest drops non-MP3 files** -- extracted buildManifest into scripts/sfx/sfx_manifest_builder.mjs so it can be unit-tested without triggering the top-level API-key guard in gen_sfx.mjs. The builder now probes .mp3, .wav, .flac, .ogg in order, so a custom recording committed as a WAV (like cast_lightning_bolt.wav) survives a rebuild without being silently dropped.

**process.exit(0) override** -- the --manifest path now exits 0 or 1 based on the builder error count instead of always calling process.exit(0).

Tests: updated existing injection from this.buffers to this.variants (arrays); added 3 nextBuffer selection tests (variant 0 reachable first play, no-repeat on 2-clip pool, single-clip always same); added 6 manifest builder tests in tests/sfx_manifest.test.ts using real temp dirs. All 13 pass, tsc clean, biome clean on changed files.

jamiecypher:
Rebased onto current `release/v0.25.0` (past the #1763 conform pipeline merge) — the only conflict was the `package.json` scripts block, kept `sfx:manifest` alongside the new `sfx:check`/`sfx:conform`. All 13 tests passing. Ready for another look when you get a chance.

jamiecypher:
Rebased again past the #1462 lockpick merge — that PR committed the manifest in the old `url` (singular string) format. After this rebase, `sfx.ts` was trying to iterate `entry.urls` on those entries and getting `undefined`. Fixed by running `node scripts/gen_sfx.mjs --manifest` to regenerate the whole manifest in the new `urls: string[]` shape. All 13 tests clean, no unhandled rejections.

EnriqueGF:
Thanks for the thorough follow-up pass :) Most of the previous round is genuinely fixed, and I verified each one on a clean worktree of the PR head rather than taking the description on trust.

Base branch gate: passes. This targets `release/v0.25.0`, which is the latest release branch, and it merges cleanly into it.

Verified fixed: the unified `variants` map makes `vcup_crowd_roar` reachable through `playUi` again, the `nextBuffer()` first-play draw is now uniform over the full pool, the `--manifest` path no longer swallows the error exit code, and the new tests are real (16 pass, `npx tsc --noEmit` clean, biome reports warnings only on the changed files).

## Blocking

**1. `cast_lightning_bolt` is still dropped from the manifest.** The multi-extension probe was the wrong fix for this one, so the regression from my last review is still live. I regenerated on your branch:

```
$ node scripts/gen_sfx.mjs --manifest
Manifest rebuilt: src/game/sfx_manifest.generated.ts (110 keys).
$ grep -c cast_lightning_bolt src/game/sfx_manifest.generated.ts
0
```

`release/v0.25.0` has 111 keys, this branch has 110, and the committed manifest already reproduces the regenerated output exactly, so the key is gone as merged. The root cause is not the file extension: `buildManifest()` only iterates the catalog it is handed (`scripts/sfx/sfx_manifest_builder.mjs:31`), and `cast_lightning_bolt` is not in the `SFX` catalog at all. The clip is a hand-added asset (`public/audio/sfx/cast_lightning_bolt.wav`), and its manifest row was hand-maintained. Probing `.wav` cannot help a key the loop never visits.

The consumer is still live: `src/ui/hud.ts:787` returns `cast_lightning_bolt` from `castKeyForAbility`, which feeds `sfx.loop(...)` for the sustained cast. After this PR the manifest has no entry, preload never loads it, and the Lightning Bolt cast loop is silent for every caster.

Two workable fixes: add `cast_lightning_bolt` to the catalog (`scripts/sfx/sfx_prompts.mjs`) as a `custom: true` entry, since `gen_sfx.mjs:82` already excludes `custom` entries from the API-key requirement, so the builder visits it and the probe you added finds the `.wav`; or have the builder scan the sfx directory for on-disk clips that are not in the catalog instead of only iterating it. The first is smaller and keeps one source of truth.

**2. The manifest tests cannot catch this class of bug.** Every case in `tests/sfx_manifest.test.ts` passes a catalog that already contains the key under test, so "a key that exists on disk but not in the catalog" is exactly the untested path, and it is the one that broke. Please add a decisive case: build from the real `public/audio/sfx` and the real `SFX` catalog and assert that every key referenced from code survives, or at minimum pin `cast_lightning_bolt` as present. A pin that would have failed on this push is the point.

## Minor

**3. Em dash in new code.** `scripts/sfx/sfx_manifest_builder.mjs:35` ("always MP3 - generated by ElevenLabs"). The repo forbids em and en dashes in code, comments, docs, and PR text.

**4. The tests overwrite `Math.random` globally and never restore it** (`tests/sfx.test.ts:130` and `:146`). The stub leaks into every test that runs after it in the file, which is a trap for whoever adds the next case. Save and restore it, or use `vi.spyOn(Math, 'random')` with `vi.restoreAllMocks()`.

The variant system itself, the extracted builder module, and the no-repeat selection rule all look right to me, and the new coverage is decisive where it aims. Restore the Lightning Bolt key, pin it with a test that would fail without it, and I am happy to approve.

jamiecypher:
Pushed a fix for both blocking items and the two minors.

**1. `cast_lightning_bolt` restored.** Root cause confirmed as you diagnosed: `buildManifest()` only ever iterates the `SFX` catalog, and `cast_lightning_bolt` was never a catalog entry, it was a hand-maintained manifest row for a hand-added `.wav`. Went with your first suggested fix (smaller, one source of truth): added it to `scripts/sfx/sfx_prompts.mjs` as

```js
{ key: 'cast_lightning_bolt', custom: true, loop: true }
```

`custom: true` keeps `gen_sfx.mjs` from ever sending it to the ElevenLabs API, matching the existing lockpick/quest pattern. Regenerated the manifest: 111 keys, matching `release/v0.25.0`, diff is exactly the one key restored.

**2. Regression test added.** `tests/sfx_manifest.test.ts` now has a case that runs `buildManifest` against the real `SFX` catalog and the real `public/audio/sfx` directory (not a synthetic catalog), asserting `cast_lightning_bolt` survives the rebuild. This fails without the fix and is the exact "on disk but not in the catalog" path your prior review called out as untested.

**3. Em dash fixed** in `sfx_manifest_builder.mjs:35`.

**4. `Math.random` leak fixed** in `tests/sfx.test.ts`: swapped direct `Math.random = () => ...` reassignment for `vi.spyOn(Math, 'random').mockReturnValue(...)` plus `afterEach(() => vi.restoreAllMocks())`, so the stub no longer escapes the file.

Verified: `npx tsc --noEmit` clean, `npx vitest run tests/sfx.test.ts tests/sfx_manifest.test.ts tests/sfx_conform.test.ts tests/combat_sfx.test.ts tests/architecture.test.ts` all green (66 tests), biome clean on changed files (no errors, no format diffs; unrelated pre-existing warnings on lines I didn't touch, per repo policy).

EnriqueGF:
Thanks for turning this around so quickly :) Re-reviewed at `3230fdcf6` on a clean worktree, and this is production ready from my side. Approving.

What I verified against the running code, not the description:

- Base branch gate passes: targets `release/v0.25.0`, the latest release branch, and merges cleanly into it.
- `cast_lightning_bolt` is restored. Regenerating with `node scripts/gen_sfx.mjs --manifest` on your branch produces 111 keys, and the key set now diffs empty against `release/v0.25.0`, so the rebuild is no longer destructive. The committed manifest matches the regenerated output byte for byte, and the row carries `loop: true` and the `.wav` URL, matching what release ships today.
- The `custom: true` catalog entry is the right lever: `gen_sfx.mjs:90` short-circuits custom entries so the clip is never sent to the ElevenLabs API, and `gen_sfx.mjs:82` keeps it out of the API-key requirement.
- The new regression test is decisive, not a pin for show. I deleted the catalog entry and re-ran it: `rebuilds the real manifest from the real catalog without dropping any on-disk key` fails. That is exactly the "on disk but not in the catalog" path that broke, and it now fails loudly.
- The `Math.random` leak is gone: `vi.spyOn` plus `vi.restoreAllMocks()` in `afterEach`, so the stub no longer escapes the file.
- The em dash in `sfx_manifest_builder.mjs` is gone, and I found no new em or en dashes in the changed files.
- Suite green: 42 tests across `sfx`, `sfx_manifest`, `sfx_conform` and `combat_sfx`. `npx tsc --noEmit` clean. Biome on the changed files reports warnings only, no errors and no format diffs, which is what CI gates on.

The variant system, the extracted builder module, and the no-repeat selection rule all read well. Nice work on chasing the root cause rather than the symptom.

One process note for #1772: it is still cut from a pre-#1763 commit and GitHub now reports it as conflicting. Land this one first, then rebase #1772 on top and regenerate its manifest so the lockpick keys come back.

jamiecypher:
Thanks for the approval! No rush, just flagging that #1772 (mob family fallback) is stacked behind this one and blocked on it landing, so whenever you get a chance to merge this one it'd help me get that one squared away too.

PR #1776, opened 2026-07-11
feat(social): player menu, chat mute, and AI / streamer account flair

CharlieSaxton:
## Summary

Two bodies of work on one branch. Both are about **who you are talking to in chat**, which is why they ship together.

### 1. Personal chat mute, clickable names, remote Player Info (the original scope)

Adds a personal chat **mute** as the light sibling of the existing **block**, makes chat sender names **clickable and tappable** (they were right-click only, so a phone player could not open the menu at all), and makes **Player Info work for someone who is not standing next to you**.

| | Mute (new) | Block (existed) |
|---|---|---|
| Slash | `/mute`, `/unmute`, `/mutelist` | `/block`, `/unblock`, `/blocklist` (aliases `/ignore`, `/unignore`, `/ignorelist`) |
| Hides | their public chat + their overhead bubble | that, **plus** whispers, invites, mail, and mutual `/who` |
| Coexists with friendship | yes | no |

**`/ignore` aliases BLOCK, never mute.** "Ignore" is already this game's word for a block everywhere a player can see it. Aliasing it to the chat-only mute would have meant a harassed player reaching for `/ignore` silently got the *weaker* tool while believing they had blocked their harasser.

Two subtleties: whispers and `/roll` ride the same `chat` SimEvent as public chat, so the filter gates on the **channel**, not the event type. And guild/officer chat never passes through `routeEvents`, so it consults the mute list at its own fan-out site.

### 2. Operator-set AI and streamer account flair (new)

An admin finds an account in the dashboard's existing search, clicks in, and marks it.

**AI accounts are DISCLOSED.** An animated-rainbow `[AI]` tag renders immediately before the player's name in **chat, on the overhead nameplate, and on the target unit frame**. One shared `.ai-tag` class serves all three: nameplates turn out to be positioned DOM divs, not canvas textures, so the gradient is compositor-animated and costs no repaint.

Because it is a *disclosure* and not decoration, the tag carries `role="img"` plus an `aria-label` naming the meaning. A `title` alone is announced inconsistently on a non-focusable span, and a screen-reader user would otherwise hear the player's name with **no indication a bot is behind it**. There is deliberately no player opt-out: a disclosure a viewer can switch off is not a disclosure.

**Streamers get no chat marker**, by design. Their Twitch / X / Kick / YouTube links appear as icon rows in the player context menu (opened from a chat name *or* a nameplate), and only the platforms that actually have a link are shown.

## Two things worth a reviewer's attention

**The streamer URLs are the security surface.** They are operator-entered text that ends up in `window.open` on every player's client. `normalizeStreamerLink` (`src/sim/account_flair.ts`) is the single boundary and runs at **four** independent points: the admin write, the wire encode, the render, and again at click. It is https-only against an *exact* per-platform hostname allowlist (never a suffix match), rejects embedded credentials, and is length-capped. The `href` is deliberately **never** interpolated into HTML, because a single quote survives URL normalization; a mutation-verified guard pins that, since one careless `href="${...}"` in a future refactor is all it would take.

**Flair rides two seams, and it has to.** Entity identity fields (`ai` / `slk`) cover players inside interest scope and re-broadcast for free on the identity diff, so an operator's flip lands on connected clients with no reconnect. A sparse `flair` on the chat event covers senders *outside* it, which is the whole point: `/general`, `/world`, `/lfg` and guild chat reach you from players who have no local entity record. An ordinary player's chat line is byte-for-byte unchanged on the wire.

## Known caveat (operator-facing)

The client's by-name flair cache is **add-only from chat**, so revoking a streamer's flag leaves their links in the menu of a client that has only ever *heard* them in chat, until that player enters interest scope or the client reloads. It **cannot** produce a phantom `[AI]` tag (the tag renders from the per-event payload, never the cache). The alternative, stamping an explicit empty flair onto every chat line of every player forever, costs more than it buys. Flagging it here so "I revoked it and it is still showing" does not arrive as a bug report.

## Related issues

Part of #1777 (Release v0.25.0).

## Type of change

- [x] Bug fix (the online/offline mute-list collision; party invite from a chat name outside interest scope; a pre-existing `parseInt` radix error in `scripts/sfx_conform.mjs` that was failing the changed-files lint gate)
- [x] Tests

## How was this tested?

- **Commands:** full `vitest` (**12422 passed, 0 failed**), `tsc --noEmit` clean, `check:admin` 0 errors, `security:gate` PASS, `biome ci --changed` 0 errors, all three builds. `npm run gate` is green on every step it reaches; it stops at a **pre-existing** `sfx check` failure that also fails on the `release/v0.25.0` tip, is unrelated to this diff (no audio files touched), and is not a step CI runs.
- **Release-tier i18n verified**, not just PR-tier: `I18N_RELEASE_TIER=1` is green (`pending=0`). Worth calling out because an English-only PR passes the PR gate and then turns the release branch red on merge.
- **`background-clip: text` verified in the shipped minified bundle**, not assumed: both `-webkit-` twins, the `--motion-scale` / `--fx-ambient-anim` ambient contract, and the reduced-motion and forced-colors arms all survive Lightning CSS.
- **Reviewed by** `privacy-security-review` (which threw 32 hostile payloads at the validator, including Cyrillic homographs, `java\tscript:`, trailing-dot hosts and `https://twitch.tv@evil.com/` - all rejected), `migration-safety`, `cross-platform-sync`, and the `qa-checklist` gate. No critical findings; every should-fix is resolved.
- **Three load-bearing wire lines are now pinned by mutation-verified tests.** The guild/officer chat decoration, the `routeEvents` decoration, and the entity identity encoder could each previously be **deleted with the full suite still green**. Each was deleted, confirmed red, restored, confirmed green.

## Screenshots / recordings

**Not yet attached.** This is a visual change (the `[AI]` tag in chat / on nameplates / on the target frame, and the stream-link rows in the player menu), so desktop and mobile shots still need capturing before merge.

## i18n

English plus **real translations for every locale**, no English placeholders. Six new game keys (`hudChrome.playerMenu.*`) filled across all 18 base overlays; 35 new admin keys filled across all 20 admin overlays. `pending=0` at release tier.

The one deliberate exception: `streamer` / `Streamer:` stay as the loanword in Czech, Danish, Spanish, Indonesian, Italian, Dutch, Polish, Portuguese, Swedish, Vietnamese and German, and are registered as cognates in `scripts/i18n_blocked_seed.mjs`. The native calques ("retransmisor", "nadawca", "uitzender") are not what an operator would ever say. Locales that *do* have a real native word for it (French "streameur", Turkish "yayinci", plus ru/ja/ko/zh) are translated normally.

https://claude.ai/code/session_01SoPDX3RkTfvNNfJGfnfKhZ


Rubsey (commented):
This is unusually careful work. I checked the load-bearing claims against the code rather than the description, ran the touched suites, and they hold up.

**What I confirmed**

- The security surface (an operator URL reaching `window.open`): `normalizeStreamerLink` is an exact-host https allowlist, never a suffix match, rejects credentials and non-https, and is length-capped. It genuinely runs at all four points, the admin write in `moderation_db.setAccountStreamerFlair` (which throws and 400s on any bad link rather than dropping it), the wire encode, the row build, and again at click in `openStreamerLink` before `window.open(url, '_blank', 'noopener,noreferrer')`. The href stays out of HTML: `ctxItemHtml` interpolates only the fixed enum id, the registry icon, and an `esc()`d label, and every name is `esc()`d. I could not find an XSS path.
- Wire parity: `ai`/`slk` round-trip (`identityFields` encode against `applyWire` decode, re-sanitized on receipt), and the chat flair rides both fan-out sites, `routeEvents` and the guild/officer `deliver()` that bypasses it. `aiAccount`/`streamerLinks` follow the same optional cosmetic-identity convention as `devTier`/`githubLogin`, so there is no `blankEntity` gap. The IWorld additions (`ignoreAdd`/`ignoreRemove`/`accountFlair`/`characterProfile`, `SocialInfo.ignores`) exist in both `Sim` and `ClientWorld`.
- Determinism/purity: `src/sim/account_flair.ts` is zero-import and host-agnostic, iterates the fixed `STREAMER_PLATFORMS` tuple only, and the sim never branches on flair. The `architecture.test.ts` line is a tightening (adds `chat_ignore_core.ts` to the pure-core scan), not a loosening.
- DDL is additive and idempotent; `ignores` is a net-new table (no `mutes` ever existed on the base, so the rename is not a data migration), and `loadAccountFlair` is back-compat. SQL is parameterized throughout, JSONB bag included.
- Tests: 399/399 pass on the touched files. The mutation-verified pins are decisive, not self-comparisons, the exhausted-bucket refusal of the four writes, the six-verb dispatch spies, the ignore-keeps-friend vs block-evicts contrast against `SocialService`, and the exact `ai`/`slk` and bare-ordinary-line wire assertions. play.html parity holds via the shared `hud.css` barrel and the css_corpus PLAY manifest.

**Findings**

- should-fix (to land, not the code): the branch is conflicting against `release/v0.25.0` (the base moved after the last merge). The real conflicts are `server/game.ts`, `server/social.ts`, `src/net/online.ts`, `src/ui/hud.ts`, `src/ui/social_window.ts`, `src/world_api.ts`, the two sfx scripts, plus the mechanical i18n trio and goldens. It needs a fresh base merge, an i18n rebuild, and a golden re-mint before it can go in.
- nit: the hostile-payload suite covers scheme, credentials, lookalike subdomain, host-in-path, host-in-query, over-length, and cross-platform mismatch, but the two cases the description names, Cyrillic homograph and trailing-dot host, are not pinned by a committed test. The code rejects both correctly via the exact-host allowlist (the same mechanism the lookalike-subdomain case already exercises), so this is optional belt-and-suspenders, not a behavior gap.
- nit: `/ignore` or `/block` typed with no name still consumes a chat token before the usage notice (`handleChatFilterCommand` charges any write-kind, then prints usage on the empty name). Harmless, and possibly intended as anti-spam, flagging only in case it was not.
- note: the screenshots you flagged are still worth capturing before merge, since the `[AI]` tag and the stream rows are the visible surface.

Nice touches worth calling out: the add-only-from-chat flair cache with the phantom-tag reasoning, the three-valued links contract with the array guard living next to the SQL, and the deliberate `/ignore` means block rather than mute naming split.

PR #1780, opened 2026-07-11
fix(content): add gathering nodes to every leveling zone

jgyy:
## Summary

Players couldn't find the ore veins for `q_prof_intro` ("A Trade for Every Hand", given by Foreman Odell in Eastbrook Vale), and once that got fixed it turned out the underlying problem was bigger: **only `eastbrook_vale` (the level 1-7 starting zone) had a full set of gathering nodes.** `mirefen_marsh` (level 6-13) had fewer nodes per type, and `thornpeak_heights` (level 13-20) had none at all. Every player past the starting zone was funneling back to zone 1 to mine, log, or herb, regardless of level, which is exactly what's happening now with realm populations regularly in the hundreds: everyone queued on the same handful of eastbrook_vale rocks.

- Moved the three `eastbrook_vale` ore nodes (`src/sim/content/gather_nodes.ts`) from Boar Meadow (a wolf/boar mob area with no mining flavor) onto the Copper Dig cluster, the zone's actual mine-themed POI.
- Updated the quest text (`src/sim/content/zone1.ts`, `q_prof_intro`) to name Copper Dig and its direction from town.
- Topped up `mirefen_marsh` from 2 to 3 nodes per type (ore/wood/herb), matching `eastbrook_vale`'s density, clustered near Fenbridge.
- Added a full set of gathering nodes to `thornpeak_heights`, which previously had zero: ore by Deeprock Burrows (the zone's mine-themed POI, guarded by the `deeprock_kobold` camp, mirroring the `eastbrook_vale` ore-vs-`tunnel_rat` precedent), wood and herb on the safer approach near Highwatch and The Glimmermere.

## Related issues

Reported in-game: players unable to locate the ore node for the professions intro quest, and separately, gathering nodes clustered entirely in zone 1 causing contention at high concurrent population.

## Type of change

- [x] Bug fix
- [x] Feature: new functionality (gathering-node coverage in previously-empty zones)

## How was this tested?

- Commands: `npx vitest run tests/gather_nodes.test.ts tests/prof_intro_quest.test.ts tests/gather_node_harvest.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/guide.test.ts`, `npx tsc --noEmit`, `npm run build`, `npm run i18n:gen` + `npm run i18n:hash -- --write`.
- Manual steps: booted the offline world and teleported to every new/topped-up node cluster in all three zones to confirm the ore/wood/herb props actually render there (see screenshots).
- Added regression tests in `tests/gather_nodes.test.ts`:
  - Eastbrook Vale ore nodes stay within 20 units of the Copper Dig POI.
  - Every zone offers at least one node of each gather type (ore/wood/herb), so this can't silently regress back to single-zone contention as new zones are added.

## Screenshots / recordings

**Coverage before vs. after, all three zones:**

coverage diagram

**Eastbrook Vale (level 1-7):**

eastbrook ore at Copper Dig
eastbrook wide

**Mirefen Marsh (level 6-13), topped up to 3 nodes per type near Fenbridge:**

mirefen ore
mirefen wood
mirefen herb
mirefen wide

**Thornpeak Heights (level 13-20), previously zero gather nodes of any type:**

thornpeak ore at Deeprock Burrows
thornpeak wood at The Glimmermere
thornpeak herb near Highwatch
thornpeak wide

---

### Quality

- [x] **The full gate passes.** Targeted `npm run gate` stages all pass (i18n gen/freshness, tsc, targeted tests, build). Note: the local whole-repo `biome ci --changed` step false-positives against this environment's `origin/main` (which trails `release/v0.25.0`, the actual PR base, per this repo's branch model); the real CI step uses `--since=$BASE_REF` and correctly scopes to this PR's own diff. `biome check` on every file this PR touches is clean.

### Cross-platform

- [x] N/A. World-content/data change only, no HUD/UI surface.

### Localization (i18n)

- [x] The only player-visible string change is the existing `q_prof_intro.text`, resolved by quest ID through the existing content pipeline (no new `t()` key needed). Regenerated `src/ui/i18n.resolved.generated/*` and rebaselined `src/ui/i18n.resolved.sha256`; `tests/localization_fixes.test.ts` passes.
- [x] N/A. No new numbers/money/dates.
- [x] N/A. No new `src/sim`/`server` emitted text.

### Hygiene

- [x] No secrets or `.env` committed; `ALLOW_DEV_COMMANDS` not touched.
- [x] No generated files hand-edited (i18n resolved tables were regenerated via `npm run i18n:gen` / `i18n:hash`, not hand-edited).


FernandoX7 (commented):
The core fix here is solid. Moving the `q_prof_intro` ore onto Copper Dig and naming the landmark in the quest text (`src/sim/content/zone1.ts:705`) genuinely fixes the discoverability bug: Copper Dig at (-84, -64) is southwest of town (0, 0) so the direction checks out, the ore still yields `chunk_of_ore` because `NODE_QUEST_GRANT` keys on the node type not the id (`professions/gathering.ts`), and it stays within `INTERACT_RANGE`. Filling in Thornpeak, which had zero nodes of any type in a level 13-20 zone, is a real gap closed, and the "every zone offers all three types" test is a good guard against it silently coming back.

I checked the placements against terrain: all twelve new/moved nodes sit on dry, walkable ground (none below the -4.5 waterline, and the steepest, `wood_thornpeak_1`, is 0.94 against the 1.5 climb limit), so nothing lands in the Glimmermere or on an unwalkable slope. The "guarded" ore claims hold too (the Thornpeak ore sits inside the `deeprock_kobold` camp, the Eastbrook ore inside `tunnel_rat`).

Two things worth a look:

- **should-fix**: the intro-quest ore is now inside the `tunnel_rat` camp (`src/sim/content/zone1.ts:1110`: center (-82, -62), radius 20, count 9, mobs level 4-6, aggro 10), and all three ore nodes sit within ~10 units of that center. `q_prof_intro` is ungated and first in `ZONE1_QUEST_ORDER`, given by Odell in town, so a level 1-2 character can pick it up and walk straight into nine level 4-6 diggers to mine 5 chunks. The old Boar Meadow spot was also guarded, but by level 2-3 boars, so this is a real difficulty bump on the very first profession quest. A kobold-guarded starter mine is classic-faithful, so this may be intended, but since the test only pins "within 20 of Copper Dig" you have room to nudge the nodes toward the town-side edge of the POI so a fresh miner can approach without pulling the whole camp. Worth a decision either way.

- **should-fix**: the contention rationale in the PR body and the test comment (`tests/gather_nodes.test.ts:65`, "creating the exact contention this pins against") does not match how nodes actually work. Respawn is per-player and session-only (`professions/gathering.ts:147`, "there is no gather rush or node camping"): one player harvesting a node never blocks another, so hundreds of players do not queue on the same rocks. The change is still worth making, just for the honest reasons: discoverability, and not having to backtrack two zones to gather at your level. Would tidy the test comment so the rationale in the tree is accurate.

Nits:
- `src/sim/content/gather_nodes.ts:46` says Thornpeak "wood and herb sit on the safer shore of The Glimmermere," but the herb pair is up by Highwatch (z ~648-678), not the Glimmermere (z ~770). The PR body's "near Highwatch and The Glimmermere" is right; just the code comment is off.
- `ore_mirefen_3` at (15, 310) is ~18 units from the Fenbridge hub center (radius 20), so it sits on the town plateau, a slightly odd spot for an ore vein given you deliberately pulled Eastbrook's ore out of town. `herb_mirefen_3` and both `wood_thornpeak` nodes also fall just inside a lake's blended footprint (dry, but they render right at the waterline). All cosmetic.
- `scripts/screenshot_one_gather_shot.mjs` and `screenshot_ore_relocation.mjs` default `OUT_DIR` to a hardcoded author-machine path; harmless since it is env-overridable, just leftover.

i18n is clean: only the existing quest key changed, and the resolved trio plus the sha256 were regenerated rather than hand-edited.

jgyy:
Thanks for the thorough review. Pushed 66818a9:

- **should-fix (ore vs. tunnel_rat camp)**: nudged all three Eastbrook Vale ore nodes toward the town-facing edge of the camp (center -82,-62, radius 20), so a level 1-2 miner on `q_prof_intro` can approach without crossing the interior. Still within 20 of the Copper Dig POI per the existing test.
- **should-fix (contention rationale)**: reworded the "every zone offers all three types" test comment. You're right that node respawn is per-player and session-only, so this was never about camping contention, just discoverability and travel distance. Fixed in `tests/gather_nodes.test.ts`.
- **nit**: fixed the Thornpeak comment, herb sits near Highwatch, wood near The Glimmermere (was swapped).
- **nit**: nudged `ore_mirefen_3` off the Fenbridge hub plateau to match the other Mirefen ore nodes' distance from town.

Left the `OUT_DIR` hardcoded path in the screenshot scripts alone since it's env-overridable and harmless, as you noted.

FernandoX7 (approved):
Just regenerate the i18n and its good to merge

PR #1781, opened 2026-07-11
feat(render): replace procedural models with Tripo-generated GLBs

jgyy:
## Summary

**Update (eb3cc9e29):** addressed the round-3 review nit on the mailbox votive glow (`mailbox.ts:70`): `attachMailGlow` no longer takes a `glowScale` argument. The glow is a sibling of the scaled GLB instance inside the unscaled outer group, so it already lived in the final 2.9-unit space; the old `1.56 * glowScale` / `0.2 * glowScale` double-scaled it and drifted the lit beacon forward of the letter slot on non-native-scale exports. `renderer.ts:4406` already pins `glow.position.y` every lit frame, so the y-scale was dead regardless. All screenshots below are retaken at head (Ultra preset, terrain/foliage/effects/shadow maxed) so they reflect this fix, not the pre-fix render.

- Replaces 11 procedural (primitive-geometry) render objects with real GLB assets generated via the Tripo AI text-to-3D API, compressed with gltf-transform to land close to each existing model category's average file size.
- Adds a CLAUDE.md to every GLB-bearing directory under `public/models/` documenting the size budget, the compression pipeline used, naming convention, and how the category is wired into the renderer.
- Every swap keeps the original procedural build as a fallback for the brief window before its GLB preload resolves (same contract the existing GLB reward-chest code already used).

Replaced models (11):

| Category | Models | Renderer module |
|---|---|---|
| creatures | rabbit, squirrel, songbird critters; leaping fish | `src/render/critters.ts`, `src/render/fish.ts` |
| resources | ore vein, wood pile, herb cluster gather-node markers | `src/render/gather_nodes.ts` |
| props | mailbox pillar, cracked grave, destructible wall, fallback crate | `src/render/mailbox.ts`, `src/render/delve_props.ts` |

A 12th model (bell rope) was generated and compressed but intentionally **not** wired in: it needs distinct pulled/unpulled visual states (rope length, bell tilt, a floor glow) that a single static GLB cannot represent without a second overlay asset, so it stays procedural for now (documented as a follow-up in `public/models/props/CLAUDE.md`).

### Compression pipeline

Each Tripo export lands raw at ~13-15 MB (1M+ vertices, 2K+ textures). Compressed with:

```
npx gltf-transform optimize <in>.glb <out>.glb \
  --texture-compress webp --compress meshopt
```

**Important correction from an earlier revision of this PR:** the first pass compressed with `--compress draco`. This repo's runtime loader (`src/render/assets/loader.ts`) only wires up a `MeshoptDecoder`, not a `DRACOLoader`, so every Draco-compressed GLB was silently failing `GLTFLoader.load()` in the browser and falling back to the old procedural geometry with no visible error. All 11 assets are now recompressed with `--compress meshopt`, verified to load in the offline client. Meshopt is somewhat less space-efficient than Draco on these small meshes, so the new assets land a bit above their category averages; each category's `CLAUDE.md` documents the actual numbers and the tradeoff.

| Category | Existing avg | New assets landed at |
|---|---|---|
| creatures (~185 KB avg) | rabbit 256 KB, squirrel 290 KB, songbird 305 KB, fish 236 KB |
| resources (~19.6 KB avg) | ore 59 KB, wood 64 KB, herb 81 KB (meshopt's practical floor for a PBR-textured mesh at this budget, see `public/models/resources/CLAUDE.md`) |
| props (~70 KB avg) | mailbox 112 KB, grave 81 KB, wall 109 KB, crate 87 KB |

### Asset pipeline

```mermaid
flowchart LR
    A[Procedural Three.js primitive] -->|prompt| B[Tripo AI text_to_model API]
    B -->|poll + download| C[Raw GLB ~13-15MB]
    C -->|gltf-transform optimize:<br/>simplify + meshopt + texture webp| D[Compressed GLB<br/>~60-310KB]
    D -->|drop into public/models/category| E[build_media_manifest.mjs generate]
    E -->|manifest.generated.ts| F[loadGltf + registerPreload]
    F -->|clone per instance| G[Renderer: critters / fish /<br/>gather_nodes / mailbox / delve_props]
    H[Original procedural build] -.fallback until preload resolves.-> G
```

## Test plan

- [x] `npx tsc --noEmit -p .` clean
- [x] `npx vitest run tests/render_glb_replacement_assets.test.ts tests/gather_nodes.test.ts tests/critters.test.ts tests/water_terrain_awareness.test.ts tests/mailbox_view.test.ts tests/architecture.test.ts tests/render_asset_preload.test.ts` -- 32/32 (updated suite) passing
- [x] `npm run build` -- clean, media manifest emits 1013 hashed assets (up from 1002), no new errors
- [x] Manual in-browser verification via the offline client (desktop + mobile viewports), confirmed the recompressed GLBs actually render (not the procedural fallback); screenshots below
- [ ] Full unbounded `npx vitest run` was attempted but hit the repo's documented "unbounded run flakes heavy suites under core contention" issue (two unrelated pre-existing tests flaked on timing under machine load); targeted suites above cover every touched module and pass cleanly. CI's bounded-worker run should not hit the same flake.

Note: `npm run gate`'s changed-files biome check locally compares against `origin/main` on this fork, which is behind `release/v0.25.0` and surfaces pre-existing unrelated issues in files this PR never touches. A scoped `biome check` on only the files this PR changed is clean (two pre-existing warnings not introduced by this change, left as-is since they predate it and are out of scope: an unused `scratch` var in `fish.ts`, and two optional-chain lint suggestions in the new screenshot script). No em dashes, en dashes, or emojis in the branch diff (verified against the release base).

## Screenshots

In-game captures from the offline client, camera zoomed in close on each
model with the player character standing next to it for scale, at the
highest graphics preset. All 11 replaced models are covered, including the
3 delve-instance props (reached via a local dev-only teleport, gated behind
`ALLOW_DEV_COMMANDS=1`, never set in production).

### Rabbit

rabbit desktop
rabbit mobile

### Squirrel

squirrel desktop

### Songbird

songbird desktop

### Leaping fish

fish desktop
fish mobile

### Gather node: ore vein

gather ore desktop

### Gather node: wood pile

gather wood desktop

### Gather node: herb cluster

gather herb desktop

### Mailbox pillar (Eastbrook)

mailbox desktop
mailbox mobile

### Delve prop: cracked grave

grave desktop

### Delve prop: destructible wall

wall desktop

### Delve prop: fallback crate

crate desktop


FernandoX7 (commented):
Nice swap. The fallback contract is exactly right: `assetsReady()` gates `new Renderer` in `main.ts` before the constructor calls `buildFish`/`buildCritters`/`buildGatherNodes`, so the browser always gets the GLB and the procedural build is genuinely just the headless/test path, same shape as `props.ts`. The Draco to meshopt correction is the real fix (the loader only wires `MeshoptDecoder`), and it is well documented. The tests assert file existence plus manifest presence rather than self-comparing, `delve_props.buildStandaloneGlb` normalizes height with a Box3 the right way, and determinism is preserved (render-only, no `Math.random`, `Object.entries` iteration is stable). Ran the touched suites at head: `render_glb_replacement_assets` + `gather_nodes` + `critters` + `mailbox_view` 32/32, `architecture` 24/24.

A few things.

**should-fix** `scripts/build_media_manifest.mjs:23` walks every file under the media roots with no extension filter, so the 10 new `public/models/<category>/CLAUDE.md` files get hashed and shipped as public media assets (`emit()` copies each into `dist/media/models/.../CLAUDE.<hash>.md`). The real manifest delta is +21 (992 to 1013), not the +11 the PR body states, and 10 of those are internal dev docs being served to the CDN. Nothing loads them at runtime so it does not break, but it publishes internal docs and inflates the shipped set. Either keep the per-directory docs outside a served root, or scope the manifest walk to known media extensions (`.glb`/`.hdr`/`.webp`/...). The docs themselves are a good addition; it is only the co-location under `public/models/` that leaks them.

**nit** `mailbox.ts:42` (GLB path) returns a hardcoded `height: 2.9` and `attachMailGlow` pins the unread-mail votive at `y=1.56`, both tuned to the procedural pillar. Unlike `delve_props`, the mailbox GLB is not Box3-normalized, so if the export is not ~2.9 tall or its letterbox sits elsewhere, the nameplate offset and the beacon drift. The screenshots do not show the unread state, so the glow anchor specifically was not visually verified. Normalizing the GLB to 2.9 with the same Box3 trick you already use in `delve_props` would make the height contract and glow anchor hold regardless of the exported scale.

**nit** `critters.ts` / `fish.ts` clone the GLB with no scale or origin normalization, so they rely on the meshes being authored at world scale with feet at the origin (`baseY` was tuned for the procedural body). It looks right in the shots; a one-line comment noting that assumption would keep a future re-export from silently sinking or oversizing them.

Not blocking: leaving the bell rope procedural with the follow-up note is the right call.

jgyy:
Thanks, good catch. Pushed 475d18b5:

- **should-fix (manifest leaking docs)**: `build_media_manifest.mjs` now scopes its walk to known media extensions (`.glb`/`.fbx`/`.hdr`/`.jpg`/`.jpeg`/`.png`/`.webp`), so the per-directory `CLAUDE.md` docs under `public/models/<category>/` no longer get hashed and copied into `dist/media`. Manifest delta is back to +11 as the PR body originally stated.
- **nit (mailbox glow anchor)**: `buildMailboxPillar`'s GLB branch is now Box3-normalized to 2.9 units the same way `delve_props.buildStandaloneGlb` does, and the votive glow's local offset scales by the same factor, so the nameplate height and the unread-mail beacon hold regardless of the export's native scale.
- **nit (critters/fish assumption)**: added a one-line comment on both GLB clone sites noting they assume world-scale, origin-seated geometry (unlike the normalized mailbox/delve props), so a future re-export that isn't will sink or oversize visibly rather than silently.

Left the bell rope procedural as-is per your note.

FernandoX7 (commented):
Re-review of the round-2 push (475d18b5). The manifest should-fix is properly resolved: `build_media_manifest.mjs` now scopes the walk to `MEDIA_EXTENSIONS`, and the only non-media files under the four roots are exactly the 10 new `CLAUDE.md` docs, so nothing real is dropped. `manifest.generated.ts` goes 992 to 1003, a clean +11, no doc entries. The mailbox is now Box3-normalized the same way `delve_props.buildStandaloneGlb` does it, which is the real part of the earlier nit: the shipped GLB is only ~1.0 units tall natively, so the hardcoded `height: 2.9` nameplate offset was genuinely drifting and now holds. Touched suites green at head (60/60), no dashes/emojis introduced.

**nit** `mailbox.ts:70` the glow's `1.56 * glowScale` / `0.2 * glowScale` isn't needed. The glow is a sibling of the scaled `inst` in the untransformed group, so it's already in the final 2.9-unit space, not native-GLB space. The y-scale is dead (`renderer.ts:4406` pins `glow.position.y = 1.56` every lit frame), and the z-scale is the only visible effect: with the shipped GLB (native ~1.0 tall, glowScale ~2.9) the lit beacon sits at z 0.58 instead of 0.2, a bit forward of the slot. `attachMailGlow(group)` with no scale matches the renderer's own fixed 1.56 and the procedural path. Cosmetic, unread-mail only, not a blocker.

Looks good to merge once you decide on that nit.

jgyy:
Fixed in 940f855c4: dropped the `glowScale` param from `attachMailGlow` and call it as `attachMailGlow(group)`, position fixed at `(0, 1.56, 0.2)`. You had it right: the glow is a sibling of the scaled `inst` inside the unscaled `group`, so it already lived in the final 2.9-unit space, and the y-scale was already dead per `renderer.ts:4406` pinning `glow.position.y` every lit frame. Re-verified `mailbox_view`, `render_glb_replacement_assets`, `architecture`, `render_asset_preload` (51/51), `tsc --noEmit` clean, scoped biome on `mailbox.ts` clean.\n\nAlso retaking all screenshots at the max graphics preset now and will update the PR description shortly.

FernandoX7 (approved):

PR #1784, opened 2026-07-11
feat(ui): add All-view chat category filter strip (#1670)

### LINKED ISSUE #1670: All chat tab mixes game event lines with player chat and offers no filtering

jgyy:
## Summary

The All chat tab mixes real player conversation with every game-emitted line
(loot, XP, quest progress, arena/event updates), and filtering was all-or-nothing:
a channel tab shows exactly one channel, and the All view had no way to thin
itself. This adds a compact toggle-button filter strip under the Chat/Combat Log
tabs, letting a player independently hide categories from the combined All view
without giving up channel tabs (a channel tab still shows exactly its own
channel, unaffected by the strip). The strip is a single nowrap row that
scrolls horizontally, mirroring the existing `#chatlog-tabs` layout, so it
never eats a second row of the chat window.

Categories: `Game`, `Loot`, `XP`, `Quests`, `Events` (the five system-line
groups) plus `Public`, `Party`, `Guild`, `Channels`, `Whispers` (the existing
player-chat channels grouped for the same strip). Choices persist across
sessions like the existing chat tabs. Defaults keep current behavior
(everything visible).

```mermaid
flowchart LR
    subgraph Sim/Server events
        A[loot] -->|chan=system, cat=loot| L(appendLog)
        B[xp/levelup] -->|chan=system, cat=xp| L
        C[questProgress] -->|chan=system, cat=quest| L
        D[arena/vcup/fiesta] -->|chan=system, cat=event| L
        E[zone/death/mailbox/...] -->|chan=system, cat=game| L
        F[say/party/guild/...] -->|chan=<channel>| L
    end
    L --> G{categoryForLine chan,cat}
    G --> H[div.dataset.cat]
    H --> I{All tab active?}
    I -->|yes| J[hidden if category toggled off]
    I -->|no, channel tab| K[hidden unless chan matches tab, unchanged]
```

## Related issues

Closes #1670

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/chat_categories.test.ts tests/chat_channels.test.ts tests/chat_tab_strip.test.ts tests/entry_window_parity.test.ts tests/client_shell.test.ts tests/localization_fixes.test.ts tests/i18n_completeness.test.ts tests/architecture.test.ts tests/hud_perf_budget.test.ts` (all green)
  - `npx @biomejs/biome check <changed files>` (0 errors, pre-existing warnings only, unrelated to this diff)
  - `npm run gate` (green, exit 0)
- Manual steps: ran the offline client end-to-end (`npm run dev` + a local Postgres via `npm run db:up`), entered the world on both a desktop viewport and a mobile-emulated landscape viewport (mobile play is landscape-locked; there is no portrait HUD to shoot), on both `/` and `/play` entries, seeded lines across all five system categories, and clicked Loot/XP off in the strip to confirm the All view actually hides them while Combat Log and channel tabs are unaffected.

## Screenshots / recordings

**Desktop, All view, filter strip showing every category (all on), single scrollable row:**

desktop all categories visible

**Desktop, Loot and XP toggled off:** those lines disappear from the log while Quests/Events/Game stay, and the toggle strip shows the on/off styling.

desktop loot and xp filtered out

**Mobile (landscape; mobile play is landscape-locked), chat panel open, filter strip meeting the 40x40px touch floor:**

mobile landscape chat panel with category strip

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green. Added `tests/chat_categories.test.ts` for the new pure module, plus source-guard tests for the strip layout and the index.html/play.html parity fix.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified live on both, on `/` and `/play` (screenshots above); the mobile 40x40px touch floor for `.chat-category-toggle` is pinned by `tests/chat_tab_strip.test.ts`. Portrait is out of scope: the game locks mobile play to landscape.
- [x] **Accessible.** Toggle buttons are real `<button>` elements with `aria-pressed` and a title, keyboard-operable via the existing chat-box tab order, with a visible `:focus-visible` ring.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Added English keys under `hudChrome.chat.categories.*` in `src/ui/i18n.catalog/hud_chrome.ts`. Two of them (`hide`/`show`) are wordy under the M16 rule, so I also added the five non-Latin fills (`zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`) in this same change, and regenerated the resolved tables (`npm run i18n:gen` + `i18n:hash -- --write`).
- [x] Numbers/money/dates: N/A, no numeric/date/money display in this change.
- [x] N/A for sim/server re-localization: this is a client-only UI addition; no new `src/sim/`/`server/` player text.

### Hygiene

- [x] No secrets committed.
- [x] No hand-edited generated files (the `i18n.resolved.generated/*` diffs are from running `npm run i18n:gen`, per the documented workflow).


FernandoX7 (commented):
Nice, tightly-scoped take on #1670. The category model is a clean DOM-free module (`chat_categories.ts`) with real unit coverage, defensive `parseHiddenChatCategories`, and a no-mutation `toggleChatCategory`, and the strip stays a thin consumer in the existing chat section rather than a new banner. It also routes category-filtered lines through the same `chat-hidden` class the channel filter uses, so `announceChatLine` already skips them and the "filtered lines are not announced" acceptance criterion holds for free. The design matches the direction #1670 sketched (a compact filter strip, one button per category), and the defaults keep everything visible.

On the MMORPG-fit question: this lands close to the RuneScape model (a persistent row of filter buttons you toggle) sitting alongside the WoW-style per-channel tabs, which is a reasonable hybrid and the exact shape the issue asked for. Two things to weigh so the chat feels right in practice, below.

### should-fix

- **The strip is missing from `/play`.** The `#chat-category-strip` div was added to `index.html:360` but not to `play.html` (its chat DOM ends at `chatlog-tabs` / `chatlog-frame`, `play.html:306-307`). `play.html` is the canonical live play entry (`worldofclaudecraft.com/play`), and `renderChatCategoryStrip` early-returns when the element is absent, so on `/play` the whole feature silently never appears (channel tabs still work; the All-view filter just isn't there and the hidden-set stays empty). It degrades without error thanks to the early return plus the `?.` in `selectChatTab`, so it is not a crash, but the acceptance criterion "desktop and mobile chat both get the controls" is not met on the primary play URL. Fix is the one-line mirror of the `index.html` div between `#chatlog-tabs` and `#chatlog-frame` in `play.html`. (The `hud.css` styles are shared via `main.ts`, so only the DOM node is missing.)

### nit

- **The new mobile touch rule is not pinned.** The PR body says the 40px floor "mirrors the existing `.chat-tab` mobile rule pinned by `tests/chat_tab_strip.test.ts`", but that suite only asserts `.chat-tab`, not `.chat-category-toggle` (`hud.mobile.css:1755`). Since the strip layout is otherwise browser-only, a one-line source-guard assertion in that file (min-height 40px on `body.mobile-touch .chat-category-toggle`) would keep the toggle from silently dropping under the touch floor later. The pure model is well covered; this is the only wiring gap.

- **Chrome weight of ten always-on pills.** This is the part I would sanity-check against the WoW/RS feel before calling the experience great. Ten labeled pills wrap to about two rows on the desktop chat window (which is only `min(370px, 50vw - 330px)`, so ~270px at the 1200px viewport your own screenshots use), and on mobile they are a 40px band above an already-short 184px log. RuneScape keeps ~7 filters in a single compact row and WoW folds per-tab filtering into a right-click menu, so ten wrapping pills is heavier persistent chrome than either reference. The always-visible state does buy real discoverability (you can see at a glance what is muted, which the WoW menu hides), so I would not move it into a menu, but consider a single non-wrapping scrollable row like `#chatlog-tabs` already uses, so the strip never eats a second row of the log. Worth eyeballing on the real narrow desktop window and a landscape phone, not just the 1200px shots.

Everything else (announce suppression, combat pane untouched, persistence, English keys in the `hud_chrome` catalog, focus-visible ring, aria-pressed) checks out.

jgyy:
Thanks for the review, @FernandoX7! Addressed all three points in 157820417:

- **should-fix**: mirrored `#chat-category-strip` into `play.html` (it was only in `index.html`, so `renderChatCategoryStrip`'s early return silently dropped the whole feature on the canonical `/play` entry). Added a parity guard (`index.html and play.html both carry #chat-category-strip` in `tests/chat_tab_strip.test.ts`) alongside the existing `.window panel` parity guard in `entry_window_parity.test.ts`, so this can't regress silently again.
- **nit (unpinned mobile rule)**: added a source-guard test pinning the 40px touch floor for `.chat-category-toggle`.
- **nit (chrome weight)**: changed the strip from a wrapping two-row layout to a single nowrap row that scrolls horizontally, mirroring `#chatlog-tabs` (same wheel-to-horizontal-scroll wiring too), so ten pills never eat a second row of the chat window on a narrow desktop or a landscape phone. Regenerated the PR screenshots against the new layout.

`npm run gate` is green on the updated branch.

FernandoX7 (approved):

PR #1785, opened 2026-07-11
feat(deeds): the Book of Deeds, a cosmetic achievements system

FernandoX7:
## Summary

This PR adds achievements to World of ClaudeCraft: **the Book of Deeds**. Players earn 192 deeds worth 2365 Renown across progression, combat, dungeons, delves, zone Chronicles, collection, PvP, social, exploration, feats, and a shelf of hidden delights. Rewards are strictly cosmetic: 19 unique titles, 3 character-sheet badge borders, and the flex. The design contract (vocabulary, architecture, binding rules, the add-a-deed recipe) lives in `docs/design/deeds.md`; open calls and deferrals for review are collected in `docs/achievements/maintainer-notes.md`.

**The player loop.** Shift+Z (rebindable) opens the Book: a summary strip (Renown, completion, recent unlocks, a nearly-there list), a category rail, entry cards with painted crest art, progress bars, global rarity percentages (online), and search plus All/Earned/Unearned/Nearly done filters. It also opens from a side-menu dock button under the quest log (keycap "s-z"; modified bindings render compact keycaps via the new keyCapLabel helper while aria labels carry the full combo), from the character sheet, the mobile More tray, and three new Chronicler NPCs who anchor per-zone Chronicles in the world (Saul in Eastbrook Vale, Osric Fenn in Mirefen Marsh, Zenzie in Thornpeak Heights). A Titles pane equips any earned title, which then renders on the nameplate, target frame, chat lines on every channel, friends and guild rosters, the inspect card, the player card, and the leaderboards. A watchlist (cap 5, persisted per character) drives a HUD tracker under the quest tracker. Unlocks batch into a gold chat line per deed, one banner, one celebration sound, and a firework burst (skipped under reduced motion).

**One sim, three hosts.** The catalog is data-as-code (`src/sim/content/deeds.ts`) and the evaluator is a sim module behind the `SimContext` seam: it runs at the tick tail, draws zero rng, and only observes persisted state, so the same code decides deeds offline, online, and headless. Deeds asserting state a veteran already holds (levels, quests done, items held, craft skills) grant retroactively on first login, batched into a single catch-up summary with no banner spam; new lifetime counters start at zero for everyone. The legacy lifetime-XP milestone titles are unified into deeds at the same thresholds, and persisted picks carry over. Render and UI consume a new `IWorldDeeds` facet implemented by both the offline `Sim` and the online `ClientWorld` mirror; titles cross the wire as deed ids, never text, and are localized client-side.

**Server.** The sim alone grants deeds; the server observes. Unlocks mirror into a `character_deeds` index (idempotent inserts, the state blob stays the source of truth) that feeds an account-level Renown leaderboard (50 Renown entry floor, self rank and top percentile), an anonymous global rarity read, and character sheets. Marquee unlocks broadcast to online guildmates and followers with a per-account opt-out, ignore lists honored, and retro credits never broadcast. Banned and suspended accounts are now delisted from every player-derived public board (arena, lifetime XP, guild, daily rewards, and the new Renown board) and relist when a suspension expires.

**Steam, dark by default.** On the Steam desktop channel players can link (never log in with) a Steam account: the shell mints a session ticket, the server verifies it upstream and mirrors mapped unlocks to Steam achievements through the publisher Web API, reconciling on link. 72 curated marquee deeds map to permanent ACH_ names under a hard cap of 100. Everything is env-gated and off by default: with `STEAM_ENABLED` unset every route answers disabled, no client renders link UI, and website builds never load Steam code.

**Wiki.** A spoiler-safe `/wiki/deeds` page lists the 183 public deeds with crest art, Renown, and rewards; hidden deeds and all earn criteria are filtered out at generation time and never reach the bundle. The guide also picks up a presentation pass: grouped, token-ranked search with "/" and Ctrl/Cmd+K shortcuts, a sticky contents rail, self-hosted fonts replacing the Google Fonts CDN, druid form stills in the model gallery, and prose expansions across a dozen pages.

**Localization and art.** Every player-visible string is localized in every locale: window chrome, unlock and broadcast lines (composed client-side from ids), and each deed's name, description, and title via an entity table in the `talent_i18n` pattern. All 192 deeds ship painted 128px WebP crests with a deterministic ingest pipeline (`npm run assets:deeds`); procedural category crests remain as the fallback tier.

One new runtime dependency: `steamworks.js`, loaded lazily on the Steam desktop channel only. Web bundles never include or load it.

## Type of change

- [x] Documentation
- [x] Tests

## How was this tested?

- Commands:
  - `npm test`: the full suite is green at the branch tip, 1011 test files and 12754 tests passed. The branch adds 26 test files (about 415 cases): evaluator triggers with negative cases, catalog integrity pins, persistence round trips, server SQL and board moderation delisting, the Steam mirror contract (never awaited on the game loop, in-flight dedupe, retry ladder), pure view cores, wiki spoiler safety, and an exact icon-to-file bijection. All 49 parity goldens are re-pinned because deed state rides the snapshot wire.
  - `npx tsc --noEmit` clean; `npm run build`, `npm run build:server`, and `npm run build:env` green; Biome clean on the changed files.
  - Guard suites: `tests/architecture.test.ts` (sim purity, determinism, the pure-core allowlist), `tests/world_api_parity.test.ts` (IWorld pinned at 212 members across 26 facets), `tests/localization_fixes.test.ts`, and `tests/i18n_completeness.test.ts` (registry pending is zero in every locale).
  - Local `npm run gate`: every step green except the sfx asset conformance check, which fails identically on an untouched checkout of the base branch (pre-existing audio debt, none of it touched here; noted in `docs/achievements/maintainer-notes.md`).
- Manual steps:
  - `scripts/deeds_screenshots.mjs` drives the real client through actual play: it earns deeds live (a Chronicle streak, discovery pickups, a bursar visit), equips a title, exercises every surface below at desktop 1600x900 and phone landscape 844x390, and pixel-verifies the unlock banner in the saved captures. The committed evidence set below is its output.
  - Spot checks in the dev client: the mobile More tray entry point, tracker collapse persistence, watch cap behavior, the broadcast opt-out toggle, and the offline sandbox showing an empty Renown board by design (accounts are server-side only).

## Screenshots / recordings

The evidence set is committed under `docs/achievements/screenshots/` (captured by the harness above); wiki shots under `docs/screenshots/`.

### The unlock moment

Deed unlock moment

### The Book of Deeds window

Book of Deeds window, desktop

Book of Deeds window, phone landscape

### Titles and the watch tracker

Titles pane

HUD watch tracker

### Renown leaderboard and title surfaces

Renown leaderboard tab

Own nameplate with the equipped title

### The side-menu dock button (before and after, then clicked open)

Side menu before, no deeds button Side menu after, deeds button under the quest log

Book of Deeds opened from the dock button

The rail is hidden on touch layouts, where the More tray already carries Deeds, so mobile is unchanged.

### A Chronicler and the wiki page

Chronicler Saul in Eastbrook Vale

The Book of Deeds wiki page

<details>
<summary>More captures: chat and player-card titles, the other Chroniclers, phone landscape variants</summary>

Chat say line with the equipped title

Player card with the equipped title

Chronicler Osric Fenn in Mirefen Marsh

Chronicler Zenzie in Thornpeak Heights

Titles pane, phone landscape

Watch tracker, phone landscape

Renown leaderboard, phone landscape

Chat title, phone landscape

Nameplate title, phone landscape

</details>

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green (one caveat recorded above: the sfx asset check step is red on an untouched checkout of the base branch as well; nothing in this change touches audio). Decisive tests cover the changed behavior, and the manual checks are recorded above.

### Cross-platform

- [x] **Tested on desktop and mobile.** The window, tracker, and title surfaces are verified at desktop and phone-landscape viewports (captures above); the Book follows the full-screen safe-area convention on phones with a chip-row category rail, 40px touch floors, and long-press tooltip peeks.
- [x] **Accessible.** Keyboard-operable with focus capture and restore on open/close/Esc, aria-pressed on toggle chips, labeled progress bars, banner text mirrored into the polite live region, and the unlock firework skipped under `prefers-reduced-motion`.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Every string is a `t()` key or an id-keyed entity row, and this branch also carries the full locale fill (registry pending is zero in every locale, release tier green), so nothing is left for a later fill.
- [x] Numbers, dates, and percents go through the i18n formatters.
- [x] Player-facing text emitted from `src/sim/` or `server/` crosses the wire as stable ids re-localized at the client boundary, and `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets, credentials, or `.env` committed; the Steam publisher key is env-only, never logged, and everything Steam ships dark by default. `ALLOW_DEV_COMMANDS` is untouched.
- [x] No hand-edited generated files: the resolved i18n bundles, wiki content, deed image id list, and parity goldens are all regenerated by their build steps.

https://claude.ai/code/session_0127VyAamMF2GPc5eTEpzP4m


EnriqueGF:
Thanks for this one :) The scale and the care are both obvious: the evaluator behind the SimContext seam, ids-only across the wire, the dark-by-default Steam gating, and the moderation delisting across every board are all done properly, and I verified the headline claims rather than taking them on trust.

**Base branch gate: PASS.** This targets `release/v0.25.0`, which is the latest release branch, and the branch already contains its tip.

**What I verified myself** (worktree at the PR head, real runs, not the description):

- `npx tsc --noEmit` clean.
- Guards green: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/snapshots.test.ts`, `tests/i18n_completeness.test.ts`, plus the deed suites (505 passed).
- `tests/parity` green, 164 passed. I also diffed the rng draw digests in all 49 regenerated goldens between the base and this head: identical in 49 of 49, so the golden churn is state and event hashes plus the three Chronicler entity ids, not a draw-order change. The "zero rng, tick tail" claim holds.
- Biome on the changed files: warnings only, no errors or format diffs.
- Screenshots: present, desktop and phone landscape. Good.

That said, I cannot approve yet. Findings by severity.

### 1. Blocking, correctness and privacy: a hidden deed is broadcast to guildmates and followers

`server/game.ts:4993` (`maybeBroadcastDeedUnlock`) gates only on `isMarqueeDeed`:

```ts
if (!def || !isMarqueeDeed(def)) return;
```

and `isMarqueeDeed` (`server/deeds_records.ts:38-40`) is `def.renown >= 25 || def.reward !== undefined`.

`hid_saul_footnote` (`src/sim/content/deeds.ts:1808-1816`) is `hidden: true` **and** carries `reward: { kind: 'title', text: 'the Footnote' }`, so it clears the marquee bar and its id is fanned out to every online guildmate and follower. That breaks the hidden-existence contract you document yourself four lines below the gate, at `server/deeds_records.ts:44-48`: "Hidden deeds are invisible until earned, EXISTENCE included, so anonymous and third-party surfaces must omit them." Every other third-party surface honors it (`publicRarityPayload`, the public character sheet); the broadcast path is the only one that does not, and `isHiddenDeedId` is already imported in that module, so this reads as an oversight rather than a decision.

Blast radius today is one deed, but any future hidden deed with renown >= 25 or a reward silently joins it. Fix: add the `isHiddenDeedId(deedId)` check to `maybeBroadcastDeedUnlock`, with a test asserting a hidden deed with a reward never broadcasts.

### 2. Blocking on policy: the locale overlays are edited

This branch modifies 20 files under `src/ui/i18n.locales/` (`cs_CZ`, `de_DE`, `es`, `ja_JP`, `zh_CN` and the rest), and not only by adding: each one also deletes roughly 68 lines, rewriting existing translated guide prose (professions, delves, FAQ). The root `CLAUDE.md` is explicit that contributors add English only, and: "Never edit the `src/ui/i18n.locales/<lang>.ts` overlays." The maintainer fills every locale at release.

The deed entity table (`src/ui/deed_i18n.newlocales.ts`) is fine, since it follows the sanctioned `talent_i18n` data-as-code pattern. The overlay edits are the problem. Please either revert them so the PR is English-only in the catalog, or get an explicit waiver from the maintainer before merge. This is not a gate failure (release tier is green), it is an ownership rule.

### 3. High, performance: the tick-tail evaluator is O(dirty players x ~150 predicates) every tick under combat

`bumpDeedStat` marks the player dirty on every call (`src/sim/deeds.ts:353`), and `onDamageDealtForDeeds` is called on every damage instance (`src/sim/combat/damage.ts:395`). So every player who lands any damage in a tick is re-evaluated at that tick's tail across all ~150 non-manual deeds, and the `while (granted)` fixpoint walks the list one extra time after any grant. Several predicates allocate per call: `METERS.townFocusPoints` (`Object.values().reduce`, `src/sim/deeds.ts:472`), `poorItemsDiscoveredCount` (walks the discovered set with an `ITEMS` lookup each, :476), `FLAGS.talentCapstone` (`Object.entries`, :496), `delveClearCount` (`Object.entries` plus a `split(':')` per key, :548).

This is the authoritative 20 Hz loop with a ~100-player interest bubble. It is not a determinism break and it is correct, but it is exactly the hot-loop shape the module-first rules warn about, and I would want a number before this reaches a populated server. Cheapest fixes: an index from stat key or trigger kind to the deeds that read it, so a `damageDealt` bump only re-checks damage deeds; or rate-limit the counter-driven dirty mark (threshold deeds already retro-grant on load, so a 1 Hz re-check loses nothing).

Related, smaller: `sweepProximityMarks` (`src/sim/deeds.ts:674`) runs at 1 Hz before the zero-dirty early-out and, when Thunzharr is not alive, scans the entire entity map once per second without breaking.

### 4. Medium: `steamworks.js` is in plain `dependencies`

`package.json:104`. It is a native module, and it is genuinely lazy (the only require site is `electron/steam.cjs:70`, never imported by `server/` or any Vite entry), but sitting in `dependencies` means a plain server or web `npm ci`, including `--omit=dev` production installs, pulls prebuilt `.node` binaries the server never loads. Move it to `optionalDependencies` or the Electron-only dep set. Side effect visible in the lockfile: `@types/node` and `undici-types` lost their `"dev": true` and are now production deps.

### 5. Medium: persistence and board reads

- `server/db.ts:2520` (`deedsBoardRows`) is an unbounded full-table read of `character_deeds` joined to `characters` and `accounts`, materialized in JS, pulled unconditionally by the 30 second warm loop (`server/main.ts:2510`) in every realm process whether or not anyone asked for the board. At 20k characters averaging 30 deeds that is roughly 600k rows every 30 seconds per process. It is cached and correct, but it is the one query here that does not degrade gracefully. Skip the warm refresh when the board has not been requested within a TTL, or aggregate in SQL.
- `server/deeds_records.ts:64`: `recordDeedUnlock` is fire-and-forget and `grantDeed` never re-emits a deed already in the state blob, so a single transient insert failure permanently loses that `character_deeds` row, with no reconcile path and no alert. The same gap makes a rollback unrecoverable (a downgraded serializer strips `deeds`, `deedStats`, `activeTitle`, `renown` from the blob on the next save). A login-time `INSERT ... SELECT ... ON CONFLICT DO NOTHING` over the earned set closes both and is idempotent by construction.
- `server/db.ts:743`: the `character_deeds_deed` index looks unused (no query filters `WHERE deed_id = $1`; the rarity aggregate group-scans). Pure write amplification unless a consumer is planned.
- `server/deeds_db.ts:47`: `deedRarityCounts` does not apply `ELIGIBLE_ACCOUNT_SQL`, so banned and suspended accounts still feed the rarity numerator and denominator, unlike every board read. Aggregate only, so a consistency nit rather than a leak.

### 6. Low: test gaps

The deed suites are genuinely strong (the 72 pair ACH map pinned as a literal, `ELIGIBLE_ACCOUNT_SQL` pinned to a literal, the forged `deed_set_title` frames driven through the real `handleMessage`, the wiki spoiler test iterating the sim table rather than the pre-filtered guide list). Four gaps:

- `tests/deeds.test.ts:84`: `collectItems` is the only trigger kind with no negative arm. An off-by-one in `need = trigger.count ?? itemIds.length` (`src/sim/deeds.ts:595`) would not go red.
- `tests/deeds_window.test.ts:130,316`: the retro silence, the a11y announce, and the reduced-motion firework skip are pinned by regex over the raw source text of `src/ui/hud.ts`. That passes for code that is textually present but wired to a dead branch, and it reds on an innocuous refactor. A painter-level test on the `PainterHost` seam with `reducedMotion()` true, asserting zero VFX calls, is the recipe in `src/ui/CLAUDE.md`.
- `tests/deeds_content.test.ts:66`: trigger literals are pinned only for the 6 refresh deeds. Design rule 9 ("never retro-edit an existing trigger") has no test: silently widening an existing deed's `questIds` or `count` keeps count, Renown total, category counts and reference integrity all green. A frozen snapshot of `DEED_ORDER.map(id => [id, DEEDS[id].trigger, DEEDS[id].renown])` would close it.
- `tests/server/steam_mirror.test.ts:208`: the backoff ladder is asserted as `[BASE, BASE * 2, BASE * 4]`, a self-comparison against the imported constant. One line (`expect(PUSH_BACKOFF_BASE_MS).toBe(1000)`) fixes it.

### 7. Nits

- `src/main.ts:6011` and `:6089` add `flashSteamStatus` and `wireSteamLink` as top-level functions. `main.ts` is a firewall, not a home: client-bootstrap wiring like this belongs in a `src/game/` or `src/ui/` sibling module.
- Scope: 610 files bundling deeds, Steam, a wiki presentation pass (self-hosted fonts, search, druid stills) and a guide prose rewrite. It is reviewable but not revertable in pieces; splitting the wiki and Steam work out would make both safer to land.

Items 1 and 2 are what block. 3 needs a measurement before a populated realm, and 4 through 6 are things I would want fixed in the same change. Everything else above is advisory. Happy to re-review as soon as the hidden-deed gate and the overlay question are settled.

FernandoX7:
Thanks for the careful review. Every finding is addressed at HEAD; point by point below.

**1. Hidden deed broadcast: fixed.** `maybeBroadcastDeedUnlock` now drops any hidden deed before it reads the opt-out flag (`server/game.ts`), and a new test drives a forged `deedUnlocked` for `hid_saul_footnote` through the real observer and asserts the record still lands while neither the flag read nor the fan-out happens (`tests/deed_records.test.ts`). One small correction to the writeup: `isHiddenDeedId` was exported from `deeds_records.ts` but not actually imported in `game.ts` yet; it is now. The fix was verified by mutation (disabling the gate reddens exactly the new test).

**2. Locale overlays: deliberate, staying.** Those edits are the maintainer-side locale fill for this feature, done on this branch on purpose. The roughly 68 rewritten lines per overlay are stale translations being refreshed because the corresponding English catalog values were reworded on this same branch (verified for all 68 in `es.ts`; zero keys were removed), and the remaining additions fill the 213 new keys. Reverting them would ship translations that mismatch the new English and reopen about 1650 pending rows at release tier. Also a small factual note: 18 of the 20 files have the rewrite; `es_ES` and `fr_CA` are pure additions.

**3. Evaluator cost: measured, then indexed anyway.** The number first: at the worst case you flagged (100 players, every one of them landing damage in the same tick, every tick), the shipped shape cost about 0.65 ms per tick, roughly 1.3 percent of the 50 ms budget, and the segment already sits under the `lap('deeds')` profiler that realm telemetry reports. Two blunting facts from the code: the `delta > 0` guard means only players who actually landed damage that tick go dirty, and only 8 of the 150 non-manual predicates allocate (the capstone walk was already cached). So it was theoretical at current scale, but your first suggested fix was the right shape and it is now in: dirty marks carry the trigger input that changed, so a `damageDealt` bump re-checks only the damage deeds. Same worst case now costs about 31 us per tick (21x less). Semantics are provably unchanged: the 49 parity goldens are byte-identical, and a new suite (`tests/deeds_dirty_keys.test.ts`) pins bucket completeness against the content table, both mark-escalation orders (a full-pass request always beats narrow keys, in either order), and the same-tick meta fixpoint. The four allocating predicates you listed are now allocation-free, and the 1 Hz sweep resolves Thunzharr through the scheduler's tracked ids over the seam instead of scanning the entity map, so an absent boss costs nothing. One note on the alternative you offered: rate-limiting the counter-driven dirty mark to 1 Hz would move grant ticks and churn every parity golden, which is why we took the index instead.

**4. steamworks.js: moved to `optionalDependencies`** (there is no Electron-only dep set in this repo, so that was the in-repo-shaped landing spot), lockfile updated (the `@types/node` and `undici-types` entries become `devOptional`). One trade-off worth flagging on top of your finding: optional deps skip silently on install failure, and `electron/steam.cjs` answers null on every failure path, so a Steam depot packaged from a tree where the optional install failed would silently ship without Steam. The steam channel therefore now asserts the module is really installed and fails the build otherwise (`scripts/electron-builder-config.mjs`, injected probe, tested both ways). Caveat: a plain `npm ci --omit=dev` still fetches optional deps by default; what the move buys is the opt-out (`--omit=optional`) plus correct flags. The production Docker image was never affected (two-stage build; the runtime image carries no node_modules).

**5. Persistence, all four sub-items:**
- The Renown board warm loop is now demand-gated (`server/deeds_board_warm.ts`): the 30 s refresh runs only while a board request landed within the last 10 minutes, and a request that arrives cold still refreshes inline on its own read path, so a viewer never waits on the loop. An idle realm process no longer pays the full-table read at all.
- The login-time reconcile you proposed is in: one idempotent `INSERT ... SELECT unnest(...) ON CONFLICT DO NOTHING` batch per join over the blob's earned set, chained on the same FIFO tail as live unlocks so ordering holds (`reconcileCharacterDeeds`). It deliberately skips the per-row Steam hook (the mirror has its own reconcile path). One correction: this closes the transient-insert drift, but not the rollback case; once a downgraded serializer strips the deeds fields from the blob, there is nothing left to replay from, so the module header documents that edge honestly instead of claiming it healed.
- `character_deeds_deed` is retired via an idempotent `DROP INDEX IF EXISTS` in the boot DDL, so deployed databases converge on their next boot.
- `deedRarityCounts` now embeds `ELIGIBLE_ACCOUNT_SQL` verbatim in both the numerator and the denominator, matching every board read.

**6. Tests, all four gaps:**
- collectItems has a negative arm now: 2 of the 3 `col_set_vale_arcanist` pieces must NOT grant, the third must (a need-lowering off-by-one goes red).
- The whole trigger + renown catalog is frozen behind a sha256 pin over `DEED_ORDER.map(id => [id, trigger, renown])`, so any edit to a shipped deed's trigger or renown fails loudly (design rule 9 finally has a guard); the test comment carries the regeneration one-liner for deliberate changes.
- `expect(PUSH_BACKOFF_BASE_MS).toBe(1000)` pins the base magnitude beside the ladder assertion.
- On the source-text pins, partial push-back plus a fix. Three factual corrections: the retro-silence decision was already behavior-tested at the pure core (`tests/deeds_view.test.ts`, `buildDeedUnlockPlan`); the firework pin scans `renderer.ts`, not `hud.ts`; and the `src/ui/CLAUDE.md` recipe cited does not exist; the firework lives in renderer scene-graph code outside the PainterHost seam, so the proposed painter-level test shape does not fit it. What we did instead: the gate decision is extracted into a pure module (`src/render/deed_fx_gate.ts`) with direct unit tests of all three arms (retro suppresses, reduced motion suppresses, neither plays), the renderer consumes it, and the source pin now guards the wiring through that gate. The source-guard suites themselves are an established convention here and stay.

**7. Nits, both taken:**
- The Steam link block is out of `main.ts` and into `src/ui/steam_link.ts` taking the `Api` instance as a parameter. It was actually four top-level functions (about 100 lines), not two, so `main.ts` sheds a bit more than flagged. The GitHub-link block it mirrored predates this branch and stays where it is.
- On scope: fair. For sizing, about 220 of the files are binary assets (icons, fonts, stills) and 36 are PR screenshots, so the reviewable source surface is meaningfully smaller than the headline count, but the bundling observation is taken.

Verification for the batch: full `tsc --noEmit` clean, the deed suites plus `tests/parity` green (164 tests, goldens unchanged), the full vitest run green (12798 tests), and all three builds pass. The two red gate steps you may see locally (`sfx:check` and changed-files biome on four sfx files) are inherited from the `release/v0.25.0` base and untouched by this branch.

Happy for a re-review whenever you have time.

Rubsey (changes requested):
Request changes, reviewed against current head `cfd56d156`.

The latest commits fix the hidden-deed broadcast and several persistence/performance gaps, but these reproduced issues remain.

### P1 findings

1. **Valid Steam Web API tickets can be rejected before verification.** [`MAX_TICKET_HEX_CHARS`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/server/steam/ticket.ts#L28-L43) is 2,048 hex characters, or 1,024 bytes. Steam's `GetTicketForWebApiResponse_t` allows 2,560 bytes, requiring up to 5,120 hex characters. A larger valid ticket is returned as `steam.invalid_ticket` without contacting Steam. [Steam API reference](https://partner.steamgames.com/doc/api/ISteamUser)

2. **Unlinking Steam is not a revocation barrier for pending achievement pushes.** [`onLinkChanged`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/server/steam/mirror.ts#L95-L101) only replaces the current cache entry; callbacks already holding the old lookup or reconcile ID still enqueue it at [lines 177-208](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/server/steam/mirror.ts#L177-L208), and queue items carry no account/generation to revalidate. I reproduced unlinking before `earnedDeedIds()` resolved and still observed `ACH_FIRST_STEPS` pushed to the old Steam ID. Other realm processes can retain a positive cache for the TTL. Revalidate before enqueue/push or use an account-scoped revocation generation.

3. **Nythraxis Deathless can be awarded after a valid raider dies.** The deed code uses the generic `|x| < 120` instance envelope for [death taints](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/deeds.ts#L1000-L1013) and [task recipients](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/deeds.ts#L1055-L1066), while the authoritative Nythraxis room is a [260-yard radius](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/encounters/nythraxis.ts#L104-L113). With a valid raider at local `x=150`, I reproduced `deathTainted=false` followed by `deathlessGranted=true`.

4. **The production guide dependency graph exposes hidden deed spoilers.** [`src/ui/icons.ts`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/ui/icons.ts#L11-L13), imported by the guide, reads the aggregate `sim/data` module, which now [re-exports the full DEEDS table](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/data.ts#L224-L227). The guide-preloaded shared icons chunk contains hidden IDs, names and descriptions; generated-source tests only inspect `content.generated.ts`. I verified `hid_saul_footnote`, its hidden name, and its full criterion in the production chunk. Split the guide icon dependency from the runtime aggregate and test the built artifact.

5. **Earning a watched deed can permanently wedge the five-item watchlist.** The tracker skips earned/unknown IDs at [`deeds_view.ts:423`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/ui/deeds_view.ts#L408-L430), but they remain stored and the raw set disables additions at [`deeds_window.ts:388`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/ui/deeds_window.ts#L387-L396). Earned cards are no longer watchable, so the stale slot cannot be removed. Prune and persist invalid/earned entries, or enforce the cap on eligible IDs.

6. **A packaged Steam release can silently use Spacewar AppID 480.** The build accepts an empty AppID at [`electron-build.mjs:111-113`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/scripts/electron-build.mjs#L111-L113), conditionally omits the stamp, and the packaged runtime [falls back to 480](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/electron/steam.cjs#L38-L51). The new native-dependency guard does not validate the AppID. Steam builds should fail when it is missing or invalid.

### P2 findings

- **Full House uses roster size rather than five participating recipients.** [`party.members.length === 5`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/deeds.ts#L1087-L1096) grants the deed even when only one member is in kill range.
- **Morthen's trio deed counts only damage dealers, then grants everyone inside.** [The check and grant](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/deeds.ts#L1180-L1188) let five players earn “three or fewer” when only three attacked.
- **Slow Renown responses can repaint a newer leaderboard tab/page.** [`renderDeedsBoard`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/ui/leaderboard_window.ts#L293-L326) captures neither board nor requested page before `await`, then writes into the shared body using the now-current `this.page`.
- **Retro discovery loses rolled quality for equipped items.** [Equipment seeding](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/deeds.ts#L928-L941) passes only item IDs. A veteran wearing a rare roll of a static-common item does not retro-earn `col_first_rare`.
- **Buyback-only items can permanently miss discovery credit.** The seed omits `vendorBuyback`, while repurchase uses the [silent add path](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/sim/items.ts#L470-L481). I reproduced an item returning to inventory while remaining undiscovered.
- **Website-distribution desktop builds falsely expose Steam linking.** The preload always exposes [`steamLinkTicket`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/electron/preload.cjs#L102-L105), and the renderer treats method presence as capability at [`steam_link.ts:69-72`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/src/ui/steam_link.ts#L69-L72), so website shells show a Link button whose call always returns null.
- **Steam ticket handles are never cancelled.** [`getLinkTicket`](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/electron/steam.cjs#L89-L108) intentionally discards the handle without calling `cancel()`, despite Steam requiring `CancelAuthTicket` when finished.

### Repository-policy blocker

The PR still edits 20 files under `src/ui/i18n.locales/`, despite the canonical [`CLAUDE.md` rule](https://github.com/levy-street/world-of-claudecraft/blob/cfd56d15617127d46029c2dd8fc2ccfb4d79cb33/CLAUDE.md#L125-L131) that contributors must never edit locale overlays.

Verification: production build, TypeScript/Svelte checks, PR-target-relative Biome, focused suites, and direct read-only repros passed. These are uncovered behavioral gaps rather than existing test failures.

FernandoX7:
Thanks for the thorough second pass. All 13 technical findings were verified against the code, confirmed real, and are fixed at HEAD, each landed test-first with a test that fails on the pre-fix code. Point by point:

**P1**

1. **Ticket clamp: fixed.** `MAX_TICKET_HEX_CHARS` is now 5120 (2560 bytes hex encoded, Steam's `k_nCubTicketMaxLength` for `GetTicketForWebApiResponse_t`), the comment cites that constant instead of the self-contradictory "a little over 1 KiB" sizing, and the tests pin the literal plus both boundaries: a 5120-hex ticket passes the shape clamp and reaches verification through the real route handler, and 5121 rejects. You are right that 2048 was sized against the old `GetAuthSessionTicket` buffer.

2. **Unlink revocation: fixed at push time.** Queue items now carry the account id, and `attemptPush` does a fresh `steam_links` read (deliberately bypassing the TTL cache) and drops the push unless the row still names the exact queued Steam id. That one read closes all three arms you identified: the in-flight lookup, the in-flight reconcile closure, and a peer realm's warm cache. Comparing ids rather than row existence keeps relink pushes flowing to the new id. Four new race tests cover your exact repro (unlink while the earned read is pending), the mid-lookup unlink, the peer-process unlink, and a peer-process relink where a stale old-id push must be dropped even though a row exists. We took your revalidate-before-push option over the generation counter: the counter is per process and would not have closed the cross-realm TTL arm.

3. **Nythraxis envelope: fixed, plus one hardening.** The death taint and the task recipients now use the boss's room radius (260, shared from `types.ts`, the same contract `nythraxisRoomMetas` uses for kill credit and the lockout) instead of the generic 120 yd band, so your wing-raider repro (death at local x 150 followed by deathlessGranted) now taints, and a wing raider on a clean kill receives credit. The same recipient fix corrects gravebreaker and wardens for wing raiders. One thing we added on top: the raw 260 yd circle would have crossed into the adjacent arena slot (slots sit 500 apart in z with the spawn skewed high), letting a neighboring raid's back-wall death taint the wrong attempt, so the circle is clipped to the slot's own z band, with a two-slot regression test.

4. **Guide spoilers: fixed at the aggregate.** `sim/data.ts` no longer re-exports the deeds table; every consumer imports `sim/content/deeds` directly, so the table's chunk color excludes the guide entirely. For the guard we went with a static value-import graph walk from `src/guide/main.ts` in `tests/guide.test.ts` rather than a built-artifact scan (no existing suite runs vite inside vitest): it fails if `content/deeds.ts` becomes reachable and prints the reaching chain, refuses to go vacuous (throws on unresolved imports, requires the walk to stay sizable and to still reach `data.ts`), and also scans every reached module for the hidden deed prose. Pre-fix it reports exactly your chain (`bestiary.ts -> icons.ts -> data.ts -> content/deeds.ts`). Bare hidden ids in `deed_image_ids.ts` stay: ids without names or criteria, and the committed icon art pins that exact set.

5. **Watchlist wedge: fixed by prune-and-persist.** A pure `pruneWatched` helper (the same skip predicate the tracker uses) drops earned and catalog-unknown ids; the window runs it on render, persists the pruned set, and nudges the HUD tracker, so the freed slot is immediately usable. Tests pin both drop dimensions, the no-op identity path, and the wedge itself: a full five-slot set with one earned member accepts a new watch after pruning.

6. **AppID 480: the build now refuses.** A steam-channel `desktopBuilderConfig` throws on a missing or non-numeric `WOC_STEAM_APP_ID` (pack and build modes both), and the stamp is unconditional for the channel. An operator who genuinely wants a Spacewar depot can pass 480 explicitly; the runtime fallback in `steam.cjs` stays as defense in depth for already-shipped builds. Docs updated; the old lenient stamp test is now a throw pin.

**P2**

- **Full House** now additionally requires every roster member to be in the kill-credit recipients snapshot (downed-but-in-range members still count, matching XP credit). Roster size alone no longer grants.
- **Morthen trio** counts the union of the damager set and the recipient envelope: five present with three attacking no longer grants, a genuine trio with a dedicated healer still does, and a damager who leaves before the kill still counts against the cap.
- **Leaderboard race:** a class-wide render epoch (the `DailyRewardsWindow` pattern) guards every post-await write, so a slow response can neither repaint a newer tab nor write its clamped page into the wrong pager. Worth noting: the epoch covers all five board arms, which also closes the identical pre-existing race in the four upstream boards (players, guilds, devs, daily), not just the deeds arm this PR added. The new jsdom test drives your exact scenario, a slow deeds response landing after switching back to players.
- **Retro discovery quality:** the equipment seed now passes each slot's rolled instance quality, so a worn rare roll of a static-common item retro-earns `col_first_rare` at join. Bag sockets store bare ids with no instance payload, so there is nothing to recover there by design.
- **Buyback discovery:** the join seed now walks `vendorBuyback` (everything there was once possessed, which is the ledger's own definition), and `buyBackItem` also credits discovery directly, so both the persisted list and a live repurchase are covered.
- **Website Steam button:** the shell now answers a real capability probe (`desktop-steam-capability`, trustedSender-gated, returning the shell's computed enabled state), the bridge exposes `steamLinkSupported`, and the renderer gates both the button and the click path on it, with a presence fallback for older installed shells that predate the probe. Website shells no longer show a dead Link button.
- **Ticket handles:** the shell keeps at most one live handle. Minting cancels the superseded ticket (never earlier, since the shell cannot observe when the server finished verifying), an empty-bytes ticket is cancelled on the spot, and every cancel is guarded so the never-throws-across-IPC contract holds.

**On the locale overlays**, respectfully, this one stands, and the rule you cite says so itself. The sentence right after the passage you quoted is the M16 exception: a new wordy English value "also needs its five non-Latin fills in the same change", enforced by the always-on `tests/i18n_completeness.test.ts` even at PR tier. This PR adds hundreds of wordy deed and guide strings, so editing at least the five non-Latin overlays is mandatory here, not forbidden; under a no-overlay-edits reading this PR would be unmergeable by construction. The remaining files are the maintainer release-fill workflow executed on the branch (7735e45d, ca0f33ab, 490b485e, 1812c2ab), with direct precedent in merged release-branch history (31182684 "complete v0.24.0 locale coverage", contributor overlay commits bb58f457, 25cdc155, 84bde60f, and the overlay edits merged in #1716 and #1720), and 490b485e specifically refreshes translations that this PR's English rewording would otherwise silently stale. The harm the rule actually guards against, English, placeholders, or TODOs poured into overlays, is verifiably absent: the only "todo" grep hits are Spanish and Portuguese "todos", and `searchPlaceholder` is a key name. Reverting the fills would reintroduce about 1650 pending rows, hard-fail the release-tier gate, and ship stale translations against the reworded English.

Verification for the batch: every fix landed test-first with the new test failing on the pre-fix code for the defect reason (not a harness error); full `tsc --noEmit` clean; the full vitest run green with the parity goldens byte-identical; env, server, and client builds all pass. The two red local gate steps (`sfx:check` and changed-files biome on the sfx files) remain the documented inherited base state from `release/v0.25.0`, untouched by this branch.

Rubsey (commented):
Re-review at head `f48f4715e`. I verified each of the 13 findings from the previous review against the code (not the response), re-ran the targeted suites in a worktree at this head, and independently re-derived the geometry and module-graph claims. All 13 are fixed, and the new tests are decisive: they pin the pre-fix failure mode, not just the new behavior. Nice work turning this around cleanly.

### P1, all verified fixed

1. **Ticket clamp.** `MAX_TICKET_HEX_CHARS` is 5120 (`server/steam/ticket.ts:36`), enforced only at `server/steam/routes.ts:49`. I traced the full dispatch pipeline for anything else that could reject a max-size ticket first: the body middleware default is 64 KiB and no schema clamps the field, so a full 2560-byte ticket now reaches verification. Boundary tests pin 5120 accept and 5121 reject through the real registered handler.
2. **Unlink revocation.** The push-time revalidation (`server/steam/mirror.ts:166-167`) reads `steam_links` through the raw parameterized query (`steam_db.ts:23-36`), not the TTL cache, so peer-process unlinks are visible; it compares ids rather than row existence, so relink pushes keep flowing; and the drop precedes the retry ladder, so the FIFO cannot be poisoned. The four race tests cover the exact repro plus the peer-relink case where a stale old-id push must drop even though a row exists. Residual: a seconds-scale window between the read and the HTTP push remains, acceptable for an idempotent cosmetic mirror (down from 60s TTL plus unbounded closures).
3. **Nythraxis envelope.** Death taint and task recipients use the shared 260yd room radius (`types.ts:2639`, `deeds.ts:1035-1044` and `1241-1246`), which also corrects gravebreaker and wardens recipients. I re-derived the geometry: slots sit 500 apart in z and the boss spawns at local (0, 96), so the raw circle reaches 106yd into the neighbor band (the z clip is genuinely needed), while the arena's own worst corner sits at 257.1yd, inside the circle, so no legitimate raider is excluded. The two-slot regression test pins the clip.
4. **Guide spoilers.** `data.ts` no longer re-exports the table. I re-implemented the import-graph walk independently: 117 modules reached from `src/guide/main.ts`, with `content/deeds.ts` and `deed_i18n.ts` both unreachable, and the same walk on the pre-fix tree reproduces the exact leak chain, so the new guard is decisive, not vacuous.
5. **Watchlist wedge.** `pruneWatched` uses the exact tracker skip predicate, runs before any cap consult, persists the pruned set, and nudges the tracker; `toggleWatch`'s single call site is only reachable post-prune.
6. **AppID 480.** The steam-channel config throws on a missing or non-numeric `WOC_STEAM_APP_ID` in both pack and build modes, the stamp is unconditional for the channel, and a direct `electron-builder` invocation cannot produce a steam-stamped artifact at all (the stamp only comes from `desktopBuilderConfig`, and unstamped builds resolve to the website channel).

### P2, all verified fixed

Full House now requires every roster member in the kill-credit recipients snapshot, which is literally the XP-credit eligible set (`combat/damage.ts:692-700`, dead-in-range counts), so the semantics match XP credit exactly. The trio counts the union of damagers and the room envelope, with the five-present-three-attacking negative pinned. The leaderboard `renderSeq` guards all five board arms after their awaits (and I confirmed those are the only await sites in the file), closing the same pre-existing race in the four upstream boards; the new test asserts the stale write did NOT land on both the body and the pager axes. Equipment seeding passes the rolled instance quality and the `col_first_rare` predicate traces end to end; the buyback claim holds by design (buyback slots have stored bare `{itemId, count}` since the feature's introduction, so there is no quality payload to recover). The capability probe is trustedSender-gated, and the presence fallback only affects shells that cannot exist: the entire `steamLinkTicket` surface is new in this PR, so no shipped shell predates the probe. Ticket handles keep the at-most-one contract with guarded cancels; not cancelling on quit or on link completion is the right conservative choice and is documented.

### Locale overlays: withdrawing the policy blocker

I fact-checked the rebuttal and it holds. The M16 exception is real and always-on at PR tier (`src/ui/CLAUDE.md`, `tests/i18n_completeness.test.ts:146-188`), and this PR's roughly 200 wordy new keys make the five non-Latin fills mandatory in the same change; the cited precedents are real, including contributor-authored overlay commits merged on main and the overlay edits in #1716 and #1720. I grepped every added line of the 6,175-line overlay diff: no English poured into non-Latin overlays (only `BRAND_ALLOW` terms), no TODO or placeholder junk, no empty values, and the "todo" hits are indeed Spanish and Portuguese "Todos". I also computed the revert counterfactual exactly: dropping the fills would create 4,260 pending rows (about 2.6x the estimate in the response) and hard-fail the release-tier gate this base branch runs, and reverting only the five non-Latin fills reds the M16 test at PR tier. The full-locale fill is normally maintainer release work, but here it is clean, green, and load-bearing, so it stands.

### One new finding from re-reviewing the fixes

- **should-fix:** the new module-graph guard skips dynamic `import()` targets (`tests/guide.test.ts`, `valueImportSpecs`). A lazy chunk is still built and served to wiki readers: `src/guide/viewer/mount.ts:34` does `await import('./scene')` today, and that subgraph is invisible to both the reachability assertion and the hidden-prose scan. I walked it: it is currently clean, so there is no live leak, but it is the one structural bypass of the guard. All dynamic specs in the graph are string literals, so walking them (or at least prose-scanning literal dynamic-import targets) is cheap.

### Nits, none blocking

- `WOC_STEAM_APP_ID=0` passes the `/^\d+$/` build gate and stamps app id 0; the runtime number branch requires `> 0` but the string branch does not (`scripts/electron-builder-config.mjs:133`, `electron/steam.cjs:46`). Require a positive integer in both.
- The Link button has no in-flight guard (`src/ui/steam_link.ts:94-114`): a double click mints twice, which can strand one handle uncancelled (`electron/steam.cjs:107-122` reads `lastTicket` before the await) and can cancel a ticket the server is still verifying. One latch closes both.
- The revalidation drop is silent (`mirror.ts:166-167`): a transient DB read failure loses a push with no log line, and the documented heal requires a future relink. A warn line, or one retry of the read, would give operators visibility.
- The `vendorBuyback` seed loop omits `slot.instance?.rolled?.quality` while its two sibling loops pass it (`deeds.ts:958-960`). Unreachable today since buyback stores bare slots, but the symmetry is free insurance against a future change.
- The 5120 boundary test presets `ctx.body` and invokes the handler directly, bypassing the middleware chain; running it through `runRoute` (as the rate-limit test does) would also pin the body-size headroom.
- `nythraxisRoomMetas` keeps the unclipped circle. I checked whether it leaks the way the deed envelope did: it does not (the nearest reachable adjacent-slot point is 385yd from the boss, behind walls the movement resolver enforces), but the new comment at `types.ts:2638-2642` now calls the radius "the raid's one membership envelope" while only the deed window is clipped. Consider the same clip there for symmetry, or soften the comment.

Verification: all targeted suites green at this head (about 550 tests across the deeds, steam, leaderboard, guide, electron, and architecture files), `tsc --noEmit` clean (which also proves no consumer still imports the table from `data.ts`), the CI PR gate (full suite) green, and the merge against the current `release/v0.25.0` tip is conflict-free.

FernandoX7:
Thanks for the thorough re-review and for verifying the batch independently. All seven items are addressed at HEAD (f48f4715e..981717cca), each with a test that was mutation-verified red against its exact target line.

**Should-fix, dynamic import() bypass: fixed.** The containment walk now collects string-literal dynamic import() specs exactly like eager edges (a lazy chunk is still built and served to wiki readers), and a computed spec fails the walk loudly instead of silently shrinking the graph. Decisiveness is pinned two ways: a parser unit test, and a walker-sanity assertion that viewer/scene.ts is reached, which holds ONLY through the dynamic edge in viewer/mount.ts (its static imports are type-only). The widened walk also pulls the locale chunks into the hidden-prose scan; all clean, matching your independent walk.

**Nits, all six taken:**
1. App id 0: the build gate now requires a positive integer, and both runtime string branches (stamp and env) hold the number branch's > 0 bar; "0" and "00" covered in tests on both sides.
2. Link button: one in-flight latch; a double click mints once, and the latch releases when the attempt settles (tested with a held mint promise).
3. Silent revalidation drop: a rejected read retries once, and a second rejection drops with the module's fixed one-line warn; id mismatches stay silent (that is the barrier working). Both arms tested, including the exact log line.
4. vendorBuyback quality: the seed loop passes slot.instance?.rolled?.quality like its siblings, with a join-seed test carrying a rolled rare in buyback.
5. The 5120 boundary test now runs through runRoute (gate, auth, limiter, body, handler), pinning the body-cap headroom too.
6. nythraxisRoomMetas now clips to the boss slot's z band, the same membership clip as the deed window, and the types.ts comment states the real convention (membership clips; in-room combat queries keep the raw circle behind the walls). Two-slot lockout regression test with geometry preconditions; parity goldens byte-identical.

Beyond the batch, we ran a completability audit over all 192 deeds and a measured performance pass, and fixed what it surfaced:
- soc_guild_founded was genuinely unearnable: the documented server observer for guildsFounded was never built. It now exists (SocialTransport.onGuildFounded, fired only by guildCreate's committed-create arm), tested on both sides of the seam.
- prog_tools_of_the_trade was also unearnable: every station-bound hub recipe consumes one of six tier 4/5 reagents that no vendor, loot table, node, or quest granted. Quartermaster Bree (the trade-goods vendor standing inside the Highwatch hub circle) now stocks all six, priced at the trade-goods staple markup already in items.ts (4x sell value), so buy stays above sell and there is no arbitrage loop. Three tests pin the chain: every hub recipe's reagent chain bottoms out at a live vendor (transitive, so the base tools are covered too), Bree sells all six from inside the hub circle at the pinned prices, and a pure shopping-trip craft through the real buyItem and craftItem paths grants the deed. With this and guildsFounded, all 190 non-legacy deeds are completable on a fresh realm (the two feat-class exceptions, feat_brightwood_relic and feat_era_cap, are documented at their records as the intended legacy class, excluded from the book).
- Four live-site grant tests (quest turn-in, gathering, delve clears, damage) pin the in-the-moment unlock so a regressed site can no longer hide behind the join-time retro pass.
- The mirror drain gained a backstop catch (a throwing push attempt was an unhandled rejection plus a wedged queue).
- Two measured perf items: the multi-key dirty-union walk is memoized (crit-tick worst case 141us to 31us at 100 players), and the board's inline refresh is single-flighted so racers on a cold cache share one full-table read.

Since your pass, the branch also absorbed the latest release/v0.25.0 (the interface-overhaul revert, a 332-file merge). We re-verified all seven fixes plus the re-homed deed surfaces at the merged head (every fix site intact, targeted suites green) and re-ran the perf benches there: every number is within noise of the audit baselines (evaluator damage-only 21us/tick at 100 players, full pass 626us, 1Hz sweeps 9-22us, board refresh against a 600k-row Postgres fixture 529-586ms with event-loop delay inside the tick budget).

Verification: full vitest suite green with the parity goldens byte-identical, tsc clean, all builds pass. The one locally-red gate step (sfx:check on audio assets, 111 files) is inherited from the release/v0.25.0 base: this branch has a zero diff against the base for public/audio and the checker itself.

Rubsey (commented):
Re-reviewed at the current head (`e72cf688a`) against `release/v0.25.0`. This is a large, multi-domain change, so I ran the full invariant pass across sim, wire/parity, server auth, the Steam subsystem, DB migration, render, UI, and tests, and verified the load-bearing claims against the code rather than the description. The short version: the invariants hold and this is in very good shape. No blockers.

What I confirmed by reading the code, not just the PR text:

- Determinism and purity. The evaluator (`src/sim/deeds.ts`) draws zero rng, reads only the sim clock, runs at the tick tail, and sits after each site's existing draws, so its placement cannot fork the draw order. Architecture and parity guards stay green. The one sim behavioral change (the legacy lifetime-XP milestone grant moving from a synchronous `milestoneUnlocked` to the tick-tail `deedUnlocked`) is deterministic and dual-write covered.
- Wire parity. Titles cross as deed ids, never text: `identityFields` encodes `title` (`server/game.ts:726`), `applyWire` decodes it (`src/net/online.ts:1726`), `blankEntity` carries `title: null` for shape parity (`online.ts:1036`), and the identity-diff re-sends the full record on equip and on clear (`game.ts:4270-4284`). The socialpos roster path propagates a cleared title as an explicit null (`game.ts:1445-1457`).
- Server authority. `setActiveTitle` (`src/sim/deeds.ts:477-485`) is the single validator both worlds reach: it accepts a title only when the deed is earned and its reward kind is `title`, otherwise a silent no-op. The deed routes are token-scoped through `ctxAccountId`, take no account id from the client, use parameterized SQL, and are rate-limited (`server/deeds.ts`).
- Steam, dark by default. The trust chain is sound: the server verifies the session ticket upstream with the publisher key and takes the Steam id from the verified response, never client input (`server/steam/ticket.ts`, `web_api.ts`, `routes.ts:63-72`); the feature gate runs before auth (`routes.ts:116,123,130`); there is a cross-account guard; nothing mints a session or touches `auth_tokens`; the publisher key is env-only and never logged; `steamworks.js` is a lazy runtime `require` on the desktop channel only, so web bundles never load it; and the mirror is fire-and-forget, deduped per (account, steam id, achievement), capped, and crash-safe.
- Migration. `character_deeds` and `steam_links` are `CREATE TABLE IF NOT EXISTS`, `deed_broadcasts` is `ADD COLUMN IF NOT EXISTS ... DEFAULT TRUE` (old rows read TRUE), indexes are `IF NOT EXISTS`, and the delist predicate `ELIGIBLE_ACCOUNT_SQL` (`server/db.ts:75-76`) is a static fragment with correct relist-on-expiry (`suspended_until <= now()`).
- The watchlist is client-side `localStorage`, per character (`src/ui/deeds_window.ts:562-582`), so there is no persistence or back-compat surface there.

Findings:

- should-fix: the Steam upstream fetch shell is untested. `verifyLinkTicket` and `pushAchievementUnlock` (`server/steam/web_api.ts:23-69`) have no direct coverage. Every consumer mocks the whole module (`tests/server/steam_routes.test.ts:20-23`) or injects around it, so the fault-mapping branches (network error, timeout, non-2xx, malformed 2xx, all to `upstream`) and the publisher-key-embedding request builders (`buildAuthenticateUserTicketUrl`, `buildSetAchievementRequest`) never run. The `fetchImpl` injection param exists precisely to make this testable but no test uses it. Risk is low because the security-decisive parser (`parseAuthenticateUserTicketResponse`) and `isTicketShape` are well covered, but this is the one real gap in an otherwise strong suite, and it sits on the link/verify path.

- nit: redundant quote-escaping at `src/ui/char_window.ts:214` (`[data-act=\"open-deeds\"]` inside a single-quoted string, while the adjacent sibling selectors use unescaped double-quotes). Cosmetic only.

Notes, not asks:

- The only true merge conflicts are `src/ui/i18n.resolved.sha256` and `src/ui/i18n.status.summary.json`, the standard stale-base i18n regen churn. Merge the base and re-run the i18n build to clear them.
- The board-moderation delisting is pinned at the SQL-fragment level rather than on a live result set. That is an accepted residual given there is no in-memory Postgres in the suite, and the fragment pin is decisive against a fragment regression, so I am not asking for more here.
- All 49 parity goldens re-pinning is expected since deed state rides the snapshot wire.

Nice work. The Steam subsystem in particular is careful, well-documented, and the security posture is exactly right.

Rubsey (commented):
(Duplicate submission, please read the review directly above this one. GitHub does not allow deleting a submitted review.)

FernandoX7:
Thanks for the deep pass. We verified all 15 findings independently at the merged tip (release/v0.25.0 is merged through 9ff1a1313) and fixed everything that held up. Range: d51840270..b64c5701f4 (release/v0.25.0 re-merged first; all CI green at the tip). Point by point:

**1. Failed save still creates deed/Steam records: fixed.** You were right that the failure arm published into the one direction the insert-only join reconcile can never heal. Unlock ids now queue on the session and publish only after the authoritative save lands, with the publish set captured at blob-serialize time so an unlock granted mid-write waits for its own queued save. A rejected save defers (30s autosave, next unlock, or the leave save retries); if no save ever lands, blob and index stay consistently without the deed. The marquee broadcast is cosmetic and still fires immediately. The old test that pinned the unsafe arm now pins the deferral, plus an in-flight capture-boundary test (mutation-verified) and a persistent-failure consistency test. One correction to the suggested fix: a plain gate-and-drop would have silently killed Steam delivery for linked accounts (the login reconcile deliberately skips the Steam hook), which is why the ids re-queue instead of dropping.

**2. Wiki spoiler boundary: not a defect, keeping as is.** The wiki never ships a boss display name (tests/guide.test.ts pins exactly that; grep for Nythraxis/Ysolei/Morthen display names in the generated content returns zero), the raid entry is masked ("The endgame raid"), and deed descs/triggers are withheld, so the wiki is strictly tighter than the game's own public surfaces: any level-1 player sees "Keepers of the Wardstones" plus its full criteria in the Book pre-earn by the documented earned-reveal contract, and the same deeds are public Steam achievements. The id slugs (dgn_nythraxis_*, hid_*) are letter-for-letter the permanent public Steam API names (ACH_NYTHRAXIS, ACH_FALL_DEATH, ...), visible in Steam's public app schema, so renaming wiki-side ids closes nothing. The bare-id tolerance is an explicit maintainer judgment recorded in the test you cited.

**3. Recycled delve slots deny dlv_solo_heroic: fixed.** deedMaxParty now resets in both mirrored per-run reset lists (claim and free). The regression test drives the real slot-pool lifecycle (duo, free, solo heroic on the same DelveRun object, asserted by identity) and the inverse pins that the in-run watermark still climbs.

**4. Steam delivery not at-least-once: fixed.** character_deeds already is the durable idempotent outbox, so instead of a new table: reconcileOnLogin now replays a linked account's earned mapped deeds at join (fire-and-forget, throttled per account, 6h TTL), shutdown drains steamMirrorIdle right after deedRecordsIdle, and the mirror header now describes the real heal. Drop-then-heal, throttle, dark-gate, and drain-order tests added. The head-of-line queue shape stays as previously flagged to the maintainer (declined design follow-up).

**5. Steam squatting + revalidation: fixed.** Reclaim-by-proof: when a verified fresh ticket proves current control of a Steam id another account holds, the link handler displaces the old link in one transaction (FOR UPDATE, race re-classified like insertSteamLink), flips the displaced account's cached mirror view in-request, and reconciles the caller. The true owner can always mint fresh tickets, so they win steady state; the stale stolen ticket cannot. Same-account already-linked 409 unchanged; ticket.ts comment now states the real bound. And the push-time revalidation moved inside the retry ladder, so an unlink or relink during backoff stops the remaining attempts (both flavors tested). The server-issued identity challenge remains the stronger long-term design; noted in the comment as follow-up since it needs desktop shell changes.

**6. saveLoadout / battlefieldExperienceTrickle skip deed marks: fixed,** plus a third sibling our sweep found: deleteLoadout auto-applies the next loadout when the active one is deleted and was equally unmarked. All three now mark exactly like their sibling mutation sites; the trickle marks only when skill was actually granted. Three red-first in-tick tests with masking isolation (rested-XP and keyed-mark masking ruled out).

**7. Encounter rosters on transient pids: fixed.** Encounter bookkeeping now keys on the stable character id (pid fallback offline): the Thunzharr death taint survives a relog, and the Morthen trio roster durably includes engaged non-damagers via the death scan plus a 1 Hz threat-table fold, so the died-healer and walked-out-alive-healer cases both count. No rng draws; all 49 parity goldens byte-identical. Recipient selection (presence-based credit) is unchanged, per the earlier round's design.

**8. Guide localization + account semantics: the copy is fixed; localization is the documented wiki convention.** The account-wide claim was wrong and is reworded (deeds are per character; only the leaderboard tallies Renown account-wide, counting each deed once), re-translated in all 18 filled locales in the same commit. English entity names on the wiki are the repo-wide guide convention (zones, dungeons, abilities all render English there); a localized spoiler-safe projection would be a guide-wide architecture change and is filed as follow-up rather than bolted on here.

**9. Focus never enters the Book: fixed.** Cold open now focuses the close button, matching the sibling cold windows (arena, spellbook, leaderboard), so the focus trap engages and Tab works the dialog instead of targeting enemies. jsdom test added. The same gap in char/market/mailbox/etc is pre-existing base debt.

**10. isHiddenDeedId fails open: fixed.** Public surfaces (anonymous rarity payload, public sheet recent strip) now require a known, non-hidden definition, closing the mixed-version-fleet and rollback skew leaks. Owner surfaces unchanged, so an earner never loses their own content; the broadcast caller already failed closed via its !def pre-check.

**11. Population consistency: the rank reads are fixed; the other two were already handled.** lifetimeXpStanding and the public rank read now gate both count arms on the same ELIGIBLE_ACCOUNT_SQL the boards use (SQL-literal pins on both arms added). The rarity two-scan race is clamped at the consumer by documented design (never reads over 100 percent, self-heals within the 5-minute cache). The ignore-list race lives in the release-owned shared chat transport (all channels equally); the branch-owned deed broadcast already reads the earner's block list fresh per broadcast and fails closed.

**Polish, all four fixed:** steam-link errors now survive the status refresh and unlink failures surface the same localized error family; the deed locale chunk is split per base locale (a non-English visitor downloads ~23KB instead of ~531KB, dialects ride their base chunk, data migrated byte-identically with literal pins); the lifetime-XP sticky self row renders the viewer's active title through the existing IWorld seam; and the compact-touch tracker now announces as a dialog opener (aria-haspopup) instead of a disclosure, with the mode-transition attribute swap pinned against the painter-facet elision cache.

Also in this range: Rubsey's round-4 should-fix (direct fetch-shell coverage for server/steam/web_api.ts through the fetchImpl injection: all fault-mapping branches, verdict pass-throughs, and both publisher-key request builders pinned to literals) and the char_window quote nit.

Verification: full gate green twice (i18n gen + freshness, malware scan, changed-files biome, full vitest 12,600+ with bounded workers, tsc, env/server/client builds), parity goldens byte-identical, and a five-reviewer pass (QA checklist, architecture/determinism, cross-host parity, privacy/security, test-coverage audit) over the whole range. The coverage audit's own findings were closed in the same range: the join-path reconcile wiring is asserted from a real join, displaceSteamLink's transaction body has direct coverage (the FOR UPDATE lock, the guard that can never delete the caller's own row, both race classifications), and the trio sweep exercises the pet-to-owner threat resolution.

PR #1787, opened 2026-07-11
fix(styles): stop bag/bank slot icons growing and overlapping on resize

MasterZensei:
## Summary

The AAA redesign turned `.item-cell` into a fluid `aspect-ratio: 1` square that
fills its `auto-fill` grid track (`grid-template-columns: repeat(auto-fill,
minmax(--slot-cell, 1fr))`). Two problems fell out of that:

- **Icons grew/shrank with the window** instead of staying a fixed size like the
  spell/action-bar icons (`.action-btn`, a plain fixed 46px box).
- **Rows visually overlapped** once the grid wrapped past one screenful: with an
  indefinite (`fr`) column size, the grid's `auto` row-track sizing does not
  reliably resolve a row's height from the cell's `aspect-ratio`, so the row
  track locked to the cell's `min-height` (40px) while the cell itself rendered
  at its full aspect-ratio height (up to ~90px+ on a wide window) — every row
  after the first bled into the row above it.

Cells are now a fixed `--slot-cell` / `--touch-cell` box, matching the
action-bar grammar. The grid re-tracks its **column count** on resize, never
the cell size, in both bags and bank, desktop and touch.

A separate `chore(sfx)` commit is included: pushing from a branch based on
`release/v0.25.0` trips the pre-push biome-changed-files gate (it diffs
against `origin/main`) on files from the not-yet-merged SFX pipeline work.
That commit is formatting/import-order only (`biome check --write`, safe
fixes), no logic changes, confirmed by `tests/sfx_conform.test.ts` still
passing (25/25) — unrelated to the icon-scaling fix itself.

## Related issues

N/A

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx tsc -p . --noEmit`
  - `npx vitest run` (full suite; 12160 passed, the only failures are
    pre-existing/unrelated flakes: an i18n build-artifact reproducibility
    check, an account rate-limit timing test, and the known Windows
    `tests/server/new_endpoint.test.ts` `tsc` spawn issue)
  - `npx vitest run --config vitest.browser.config.ts tests/browser/target_size.browser.test.ts`
    (real-browser geometry; added a regression test asserting bag-grid rows
    never overlap once cells wrap past one row, and a test asserting cell
    size stays pinned at any container width)
  - `npx @biomejs/biome check` on all changed files
- Manual steps: booted the offline world, filled the bags with 32 distinct
  items, and resized the bags window across a range of widths/heights to
  confirm icons stay a fixed size and rows never overlap (see screenshots).

## Screenshots / recordings

Before (rows visually overlap; icon size drifts with the window):

before

After (fixed-size icons; only the column count re-tracks):

after

---

### Quality

- [x] **The full gate passes.** `tsc`, the full `vitest run` suite, the
      real-browser geometry suite, and biome are all green (see "How was this
      tested?" for the caveat on pre-existing unrelated flakes). Added/updated
      decisive tests for the changed behavior.

### Cross-platform

- [x] **Tested on desktop and mobile.** The fix applies the same fixed-cell
      contract to both the desktop `--slot-cell` and touch `--touch-cell`
      tracks; `tests/browser/target_size.browser.test.ts` covers both, plus
      every bag/bank dock state (standalone, bank-open, market-open).
- [x] **Accessible.** No markup/ARIA changes; this is a CSS sizing fix only.

### Localization (i18n)

### Hygiene


jgyy (commented):
The actual fix is good. `.item-cell` moves from a fluid `aspect-ratio` box inside an `fr`-sized grid track to a fixed `--slot-cell`/`--touch-cell` box inside `repeat(auto-fill, var(--slot-cell))`, applied symmetrically across bags/bank and desktop/touch, which matches the stated root cause (indefinite `fr` tracks can't reliably resolve "auto" row height from `aspect-ratio`). The `client_shell.test.ts` literal pins were updated in lockstep with the new CSS, and the new coverage is genuinely decisive: `target_size.browser.test.ts` asserts cell size is identical at narrow vs. wide container widths, and separately asserts row tops never regress once cells wrap to a second row, both of which would have failed against the original bug. I ran it myself: `client_shell.test.ts` is 84/84 on this branch.

The second commit is the problem. It's titled a format-only chore, but it isn't. This branch's base is 341 commits behind current `release/v0.25.0`, and in that gap the SFX pipeline was replaced with a real subsystem, `scripts/sfx_studio/`, `server/sfx_overlay.ts`, `server/static_sfx.ts`, `conformSfxAudio`/`buildSfxGenerationPlan`/the `--manifest` flag, dozens of tests. This commit's "biome --write, no logic changes" diff to `scripts/gen_sfx.mjs` and `scripts/sfx_conform.mjs` is measured against the OLD pre-studio pipeline, and if merged toward this branch's version it would silently revert that entire subsystem. The "still 25/25" claim in the commit message is true only because the commit's own diff deletes the 5 conform-audio-backed tests that would otherwise catch the regression; base currently has 30. Recommend dropping the `chore(sfx)` commit entirely, there's no formatting violation left to fix against the real base (I checked, current base SFX files pass `biome check` clean as-is), and rebasing the CSS commit alone onto current `release/v0.25.0`.

Given the 341-commit gap, the PR's own `package.json` still reads 0.24.0 and the claimed green browser run predates a lot of unrelated fixture drift (3 unrelated failures show up if you run the browser suite as-is today, none of which are caused by this PR's diff, but they mean the "tested" claim needs a re-run post-rebase, not just trust).

So: good fix, wrong base. Rebase onto current `release/v0.25.0`, drop the SFX chore commit, and re-run `npm run gate` before this is mergeable. Nothing else stood out, no i18n impact (CSS-only), and the mobile/touch handling correctly preserves the `body.mobile-touch` scoping and the `bank_window.test.ts` selector contract.

PR #1766, opened 2026-07-10
fix(dungeons): allow eligible ghosts to reclaim raid loot

No898:
## Summary

- Allows a loot-eligible released ghost to cross the Abandoned Crypt and return to the same party's defeated Nythraxis arena.
- Builds on #1767 by treating the exact Nythraxis corpse-run proof as an alternate ownership proof to its `clearedBy` gate.
- Binds the exception to the canonical arena footprint, exact instance slot, current party claim, dead boss, and loot-recipient snapshot, including the wide side wings.
- Preserves the whole-party lockout, fresh-claim, foreign-claim, living-player, ineligible-player, and undefeated-boss protections from #1767.
- Lets the intermediate ghost backtrack outside if the arena claim disappears.

## Related issues

Fixes #1764

Related: #1767

#1767 fixed a separate heroic re-run abuse reported on Discord and has no linked GitHub issue. This PR was rebased by merging the current `release/v0.24.0` containing that fix, then adjusted to preserve its `clearedBy` and whole-party lockout behavior.

## Type of change

- [x] Bug fix
- [ ] Feature: new functionality
- [ ] Refactor or performance (no behavior change)
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/dungeons.test.ts` (49 passed)
  - `npx vitest run tests/nythraxis_raid.test.ts` (58 passed)
  - `npx vitest run tests/architecture.test.ts` (24 passed)
  - `npm run ci:changed -- --diagnostic-level=error`
  - `npx tsc --noEmit`
  - `npm run gate` reached the full suite with i18n freshness, malware, and changed-file Biome green; its five-worker test phase hit four unrelated 5-second timeouts under concurrent host load
  - `npm test -- --maxWorkers=1` (948 test files passed, 1 skipped; 11,576 tests passed, 14 skipped)
  - `npm run build:env`, `npm run build:server`, and `npm run build`
  - pre-push QA floor (53 passed, 3 skipped; typecheck and lint floor green)
- Reproduction:
  - Merged current `release/v0.24.0` at `e3cbb887a`, which includes #1767 and the latest #1773, and confirmed the release delta has no dungeon or Nythraxis overlap.
  - Exercised kill, lockout, release, outer crypt entry, royal door entry, and resurrection through a real `Sim`.
[...]

PR #1789, opened 2026-07-11
feat(social): Dungeon Finder (role queue, premade board, catalogue window)

EnriqueGF:
## Summary

Adds the Dungeon Finder (spec: `docs/prd/dungeon-finder.md`, included in this PR): a realm-local group finder that **forms parties and raids only**. It never teleports anyone, never enters an instance, and never touches dungeon difficulty or loot settings; the group assembles where everyone stands and travels to the entrance together.

Three surfaces in one window (desktop micro-button, mobile More menu, rebindable `Shift+I`; bare `I` stays Calendar):

- **Catalogue**: every activity in the release (the four five-mans on Normal and Heroic, the Abandoned Crypt, and the Nythraxis raid on both difficulties) with strict level band, group size, role split, daily-lockout state, attunement requirement, entrance zone plus a non-teleporting "Show on Map" ping, and every encounter in order with prerendered WebP portraits, notable mechanics, and the authored loot tables (roll-group semantics rendered honestly: "one of these always drops" only when the group's chances partition a full draw; Heroic bonus groups and Heroic Marks shown on Heroic).
- **Quick Match**: role selection (below level 10 a fixed class table: tank = warrior/paladin/druid, healer = paladin/priest/shaman/druid, dps = all; from level 10 the active specialization's role, no spec = no queue), multi-activity selection with deterministic FIFO matching over indivisible solo/premade units, exact 1/1/3 and 2/2/6 compositions, a 30-second all-confirm proposal, a 60-second decline/timeout cooldown, and original-join-time restoration for the units that accepted.
- **Premade Groups**: leader-published listings with structured tags only (first run, quest run, full clear, learning welcome, fast run; deliberately no free text), applications with leader accept/decline, full revalidation on acceptance, and the solo Abandoned Crypt as a listing-only activity (up to five, no role checks, level 20).

Level bands are strict inside the finder (anti-boost): Hollow Crypt 7-10, Sunken Bastion 12-13, Drowned Temple 16-18, Gravewyrm Sanctum 19-20, every Heroic and the raid 20. Every queued member, listing member, and applicant must be in band; physical door rules are unchanged.

Architecture, per the repo seams:

- `src/sim/content/dungeon_finder.ts`: declarative catalogue (activities, bands, compositions, encounter order, mechanics keys, entrances); no heuristics. Loot previews read the canonical `MOBS[bossId].loot` + `HEROIC_BOSS_LOOT`.
- `src/sim/social/dungeon_finder.ts`: the machine behind `SimContext`. Draws no rng, reads only the sim clock; matching runs from a new end-of-tick phase (`sim.dfinder` lap) that cannot fork the draw order. Pure helpers (role compatibility, Kuhn max-matching role assignment, FIFO assembly) are exported and unit-tested.
- `PartyMachine.formDungeonFinderGroup`: the one authoritative formation seam; validates live rosters (null = nothing mutated), preserves the oldest premade's party/leader/loot settings, oldest solo leads all-solo groups, converts to raid before exceeding the party cap.
- New `IWorldDungeonFinder` facet (implemented by both `Sim` and `ClientWorld`), nine `df_*` wire commands validated field-by-field in `server/game.ts`, and two delta keys: `df` (personal, 2 Hz) + `dfb` (shared board, capped at 50 listings).
- `src/ui/dungeon_finder_view.ts` (pure core, in `UI_PURE_CORES`) + `src/ui/dungeon_finder_window.ts` (thin painter). Cold window: zero work closed; open, it rebuilds DOM only on a structural signature change, and the 1 Hz countdowns refresh cached text slots in place. Portraits are committed WebP stills (new `scripts/render_finder_portraits.mjs`, reusing the guide-stills renderer), never a live 3D scene.

## Related issues

N/A (feature request from the maintainer, spec in `docs/prd/dungeon-finder.md`).

## Type of change

- [x] Documentation
- [x] Tests

## How was this tested?

- Commands:
  - `npm run gate` green end to end: i18n gen + freshness, malware scan, changed-files biome, full vitest (924 files / 11389 tests, 6 workers), `tsc --noEmit`, env/server/client builds.
  - Focused: `tests/dungeon_finder.test.ts` (44: bands at both edges, pre-10 role table, spec gate, 1/1/3 + 2/2/6, multi-role reseating, FIFO + intersection, premade indivisibility and leader retention, decline/timeout cooldowns with join-time restoration, listing/application lifecycle, disconnect and roster-change cleanup, no-teleport/no-difficulty/no-loot-mutation, determinism), `dungeon_finder_view.test.ts`, `dungeon_finder_window.test.ts`, plus the repinned wire gates (`command_schema` 140/149, `snapshots` 38 delta keys, `world_api_parity` 215 members / 26 facets), `architecture`, `localization_fixes` (S3), `i18n_completeness` (M16), `tests/parity` (rng draw-order), `keybinds`, `mobile_controls`, `css_corpus`.
  - Reviewed by the repo agents: `qa-checklist` (READY, findings fixed in this PR), `architecture-reviewer` (0 blocking; both notes applied), `privacy-security-review` (all checks passed; noted follow-up: a dedicated `df_*` rate limit is hygiene, state is O(players)-bounded).
- Third commit: the WoW-style group-found popup (per-role accepted meters, countdown, check/cross buttons, audio cue) shown outside the window, plus instant proposal answers from dev bots.
- New in the follow-up commit: role/tab glyphs, per-activity boss portraits in the catalogue rail and detail header, styled single-select chips replacing the raw select, a bounded flex-column window (no runaway scroll on any tab) with sticky encounter portraits, honest "at most one may drop" labels for partial roll groups, and a dev-only `/dev lfg [queue|raid|board]` seeding command (spawns finder bots that queue around your role, fill the raid split, or publish listings and apply to yours) so the whole flow is testable solo.
- Manual steps: booted offline and online (dev server + authoritative server), formed a full five-player group through the queue (proposal, accept, formation in place), published and filled listings, verified the crypt listing path, "Show on Map" ping, and the mobile portrait/landscape single-pane flow with the Back action.

## Screenshots / recordings

Committed under `docs/screenshots/` (desktop 1600x900, mobile 390x844 portrait + 844x390 landscape):

| | |
|---|---|
| Catalogue, Heroic detail | Catalogue, raid detail |
| Quick Match | Group proposal |
| Group-found popup | Premade board |
| Show on Map ping | |
| Mobile list | Mobile detail |
| Mobile Quick Match | Mobile landscape |

---

### Quality

- [x] **Tests pass.** `npm test` is green (via `npm run gate`, 11389 tests); new sim/server behavior is covered by `tests/dungeon_finder.test.ts` and the repinned wire gates.
- [x] **Builds succeed.** `npm run build`, `npm run build:server`, `npm run build:env` (all inside the gate).

### Cross-platform

- [x] **Tested on desktop and mobile.** Portrait and landscape screenshots above; every tappable finder control has a 40x40 floor (`hud.mobile.css`), the create-listing `select` inherits the central 16px input floor.
- [x] **Accessible.** Focus trap + focus return via the shared window-focus seam, `dialog` root, `aria-pressed`/`role=checkbox` state, localized aria labels, `:focus-visible` on loot rows, no hover scale, no animation (reduced-motion safe by construction), forced-colors border via the shared `.window` rule.

### Localization (i18n)

- [x] New player-visible strings are `t()` keys added to `en` (`hudChrome.finder.*`), with the five non-Latin fills (`zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`) supplied per the M16 rule; the remaining locales are build-filled `pending` for the release pass (contributor policy).
- [x] Numbers, percents, money, and durations go through `formatNumber`/`moneyHtml` and the shared coin markup.
- [x] Sim-emitted player text is re-localized at the client boundary in this same change (`dfinder.*` EXACT map + 4 name RULES in `src/ui/sim_i18n.ts`); `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets, credentials, or `.env` committed; no `ALLOW_DEV_COMMANDS` involvement.
- [x] No generated files hand-edited (i18n tables regenerated via `npm run i18n:gen` + `i18n:hash`; boss portraits generated by the committed script).


Rubsey (commented):
This is a strong feature PR, and the architecture work shows under review. Things I verified directly at the head (7672f662d): the sim machine draws no rng and reads only the sim clock, the ordering paths (FIFO, the Kuhn reseater, proposal expiry) are total orders, and the parity golden traces stay green with the new end-of-tick dfinder lap; all nine df_* handlers validate field-by-field and act on the session pid; /dev lfg sits behind the real ctx.devCommands gate (ALLOW_DEV_COMMANDS server-side); the catalogue's activity, encounter, and entrance ids all resolve against real content, encounter order matches the actual layouts, and the loot preview reads the canonical MOBS/HEROIC_BOSS_LOOT tables with roll-group math that mirrors the sim's partitioned draw; lockout and attunement display read the same state the doors enforce; listings are one per leader, leader-only, band-checked, structured tags only, so no free text crosses the wire and finder state stays O(players); the board view is rev-cached and the 2 Hz df/dfb cadence is sensible; index.html and play.html get identical chrome; the pin bumps are honest (command_schema +9/+9, exactly the nine new commands); and dungeon_finder_view.ts is genuinely DOM-free with decisive tests behind it. Nice work.

Two blocking findings, then smaller ones.

**Blocking**

1. Stale `src/ui/i18n.resolved.sha256`: commit 55b15aef3 added `hudChrome.finder.slotState` and regenerated all 22 resolved tables but never rewrote the baseline, so `tests/i18n_resolved_equivalence.test.ts` fails at the PR head (2 of 5; I ran it). CI has not surfaced this because a conflicting PR gets no merge ref, so no checks ran on this head. The fix is mechanical: `node scripts/i18n_resolved_hash.mjs --write` and commit the one-line baseline.

2. The closed finder window is visible at HUD boot. Three spots share one root cause and one fix:
   - `src/styles/components.css:3896`: `#dungeon-finder-window` sets `display: flex` in `@layer components` with no `display: none`. The closed-state default (`.window { display: none }`) lives in `@layer layout`, which loses to components, so before the first open the window computes `display: flex`: an empty 640px panel sits on screen from boot until the player first opens and then closes the finder, every session (verified in a real Chrome against the actual style barrel; the painter's `close()` early-returns when inline display is not 'block', so it never hides it either). On mobile it is worse: the visible window flips `body.mobile-window-open` at boot, showing the backdrop and hiding autorun.
   - `src/styles/hud.mobile.css:4067`: the whole mobile finder block was appended after the `@layer hud-mobile` wrapper closes, so those rules are unlayered and outrank every layer.
   - `src/ui/dungeon_finder_window.ts:103`: `open()` sets inline `display: 'block'`, so the flex-column design (max-height, per-pane scrolling) never actually applies while open; the bounded-window behavior described in the PR is not what ships.
   `#market-window` in the same stylesheet is the house pattern: `display: none` in the components rule, painter opens with inline flex (`market_window.ts:120`). Follow it, move the mobile block inside the layer, and update the two 'block' checks (`isOpen` at `dungeon_finder_window.ts:82`, the render gate at `hud.ts:6985`).

**Should-fix**

3. The group-found popup is never announced to screen readers (`src/ui/dungeon_finder_proposal_popup.ts:73`): no `role=alert`/`aria-live`, and focus does not move, so an SR user misses the 30-second window entirely while the in-window banner is announced. `role="alert"` on show would cover it.

4. ARIA role misuse in the window: `role=tablist` at `dungeon_finder_window.ts:358` has plain `aria-pressed` buttons as children (no `role=tab`/`aria-selected`), and `role=listbox` at line 386 has children that are not `role=option`. Either commit to the full tab/listbox contracts or drop to plain toggle-button groups; half-applied roles read worse to AT than none.

5. Localizable composition bypasses the token conventions in four spots (English-side mechanics, not a translation ask): role counts hardcode the count-space-label order via `${num(n)} ${label}` (`dungeon_finder_window.ts:429`, 686), the needs line concats a prefix at 676, `mmss()` at 777 hand-rolls the `:` clock instead of the `hudChrome.vcup.clock` `'{minutes}:{seconds}'` pattern every other clock uses, and the popup builds `accepted/total` at `dungeon_finder_proposal_popup.ts:113` while `hudChrome.finder.slots` (`'{size}/{capacity}'`) already exists. Template keys with tokens fix all four.

**Nits**

6. `dungeon_finder_view.ts:222` and 384 hardcode `level >= 10` where the sim machine uses `FIRST_TALENT_LEVEL`; import the constant so they cannot drift.
7. `matchFinderRoles`: `seat()` and `seatElsewhere()` are the same ~20-line augmenting path with one extra skip; one function with an optional `except` param.
8. Nothing guards that a portrait webp exists for every catalogued encounter (the view builds URLs blindly at `dungeon_finder_view.ts:285`); a future catalogue addition ships a silent 404 portrait. A cheap existence loop over `FINDER_ACTIVITIES` in `dungeon_finder_view.test.ts` closes it.
9. `hudChrome.finder.partySize` (`'{count} players'`) bakes the plural into the template; fine today since no activity renders a count of 1, but `tPlural` + `hudChrome.plurals.*` is the house pattern.
10. Your own note about a dedicated df_* rate limit is right, and it is fine to land without one: the handlers are cheap (matching runs in the tick phase, not per message) and state is bounded, so it stays a follow-up.

**Notes, no action needed from you**

- The conflicts against the current base tip are small and mechanical: `index.html`/`play.html`/`sim.ts`/`hud.ts` collide with #1766/#1784, plus the usual regen set (the five overlay files, the generated i18n trio, and the two repinned gate tests). Merge the base, re-run `npm run i18n:gen` + the hash write, and re-derive the two count pins.
- At the head I ran the 11 gate test files (525 passing), tests/parity (164 passing), and the S3 guard; everything is green except the equivalence test in finding 1.
- The 71k-line `i18n.status.json` churn is a base artifact (that registry is supposed to be gitignored and got resurrected on release branches); not yours to fix.
- release/v0.24.0 has shipped, so this will be retargeted to a later release branch; that call happens maintainer-side.

EnriqueGF:
Thanks for the review, this was a thorough one. Retargeted to `release/v0.26.0` (v0.24.0 has shipped), merged the new base, and landed every finding. Head is now `11370b5`.

**Blocking**

1. **Stale `i18n.resolved.sha256`**: fixed. The merge regenerated the tables (`npm run i18n:gen`) and rewrote the baseline (`node scripts/i18n_resolved_hash.mjs --write`); `tests/i18n_resolved_equivalence.test.ts` is green. The resurrected `i18n.status.json` registry is gone too (the new base gitignores it).

2. **The closed window was visible at HUD boot**: fixed, all three spots, following the `#market-window` house pattern.
   - `components.css`: the `#dungeon-finder-window` rule now declares `display: none`, so the `@layer components` rule no longer outranks the closed-state default in `@layer layout`.
   - `hud.mobile.css`: the whole mobile finder block moved INSIDE `@layer hud-mobile`, and its root rule no longer sets `display` at all (a layered `block` there would both re-show the closed window at boot and defeat the flex-column bounds while open).
   - The painter opens with inline `flex` and both open-state checks (`isOpen`, the `hud.ts` render gate) follow, so the bounded flex-column window described in the PR is now what actually ships.
   Verified in a real Chrome, offline boot: the panel computes `display: none`, width 0, and `body.mobile-window-open` stays off (it was `display: flex` / 640px / on before). Pinned by a new stylesheet-contract suite in `tests/dungeon_finder_window.test.ts` (closed-state rule, no CSS after the layer wrapper closes, painter and HUD gate both on `flex`), and every screenshot is recaptured on the now-correctly-bounded window.

**Should-fix**

3. **Popup not announced**: the popup gets `role="alert"` + `aria-live="assertive"` on show. It still never moves focus by design (the player may be fighting), which is exactly why the live region was the right fix.

4. **Half-applied ARIA roles**: dropped. The tabs and the catalogue rail are labelled `role="group"` toggle-button groups now (`aria-pressed` buttons with no `role=tab` / `role=option` children never earned `tablist` / `listbox`). Pinned so it cannot regress.

5. **Localizable composition**: all four go through token templates now. New keys: `hudChrome.finder.roleCount` (`'{count} {role}'`), `hudChrome.finder.clock` (`'{minutes}:{seconds}'`, the same pattern `hudChrome.vcup.clock` uses), and `needs` is now `'Needs {roles}'` instead of a concatenated prefix. The popup's meter uses the existing `hudChrome.finder.slots`. The five non-Latin fills are in the same change.

**Nits**

6. `dungeon_finder_view.ts` imports `FIRST_TALENT_LEVEL` instead of the literal 10, in both spots.
7. `seat()` / `seatElsewhere()` are one augmenting path with an optional `except`.
8. Added: `tests/dungeon_finder_view.test.ts` now walks every `FINDER_ACTIVITIES` encounter and asserts a committed webp exists under `public/ui/dungeons`, so a future catalogue addition cannot ship a silent 404.
9. `partySize` is now `tPlural('hudChrome.plurals.finderPartySize')` with all four CLDR categories.
10. Agreed, the dedicated `df_*` rate limit stays a follow-up.

**Also found while merging the new base** (three party-side rules the release landed on the invite/leave arm that the finder's own formation seam skipped, each with a regression test in `tests/dungeon_finder.test.ts`):

- `partiesJoined` (the deed stat `acceptInvite` bumps) was never credited for a finder-formed group. It is now, for the leader and every joining member.
- A matched premade's ready check outlived its dissolved source party: its members resolve to the NEW party, so nobody could answer it and it ran the full 30-second timeout reporting on a group that no longer existed. `formDungeonFinderGroup` clears it with the party.
- The finder ignored `meta.leaving`, the disconnect flag every other reward system honours, so a disconnecting player could still be matched (or burn a whole proposal for four others). Finder teardown moved into `preparePlayerLeave`, which the server runs before the persistence await, and both `unitIntact` and the formation validation reject a leaver.

Gate is green end to end on the new base (full vitest, `tsc`, all builds). The one red is `tests/browser/armory_mobile_layout.browser.test.ts`, which fails identically on `release/v0.26.0` itself; not from this PR.

PR #1790, opened 2026-07-11
feat(easter-egg): introduce The Open Source raid dungeon

madmatah:
## Summary

**The Open Source** is a secret raid dungeon whose roster is the people who built this game. It introduces an 'endurance' survival mechanic where players are trapped inside a enclosed perimeter and must face increasingly challenging waves of enemies. The goal is to test player builds, resource management, and crowd control under pressure

At server startup the dungeon reads the GitHub contributors leaderboard and generates itself from it. Every mob in the room is a real contributor.

The dungeon is hidden. You have to interact several times with a well to enter. You walk into an open space. Contributors are standing around a sealed stone circle, friendly, minding their own business. Their nameplates show a name and a rung, nothing else, because you cannot fight them and the room has nothing to warn you about yet.

Then somebody pushes the button in the middle.

The whole roster turns hostile in a single tick, the exit seals behind you, and every nameplate in the room switches from tribute to threat at once. The dungeon does not just start; it changes what kind of place it is.

### Getting in

Nothing points at this place. No quest, no map marker, no breadcrumb: you find the well and you keep interacting with it until it gives. It is level 20, tuned for a 10-player raid, on a daily lockout of one clear per character. Ghosts keep the portal, so a wipe never strands a raid outside its own instance.

### Who you are fighting

The leaderboard is the roster, and your standing in it is your standing in the fight:

- Rank 1 is the boss, and reacts when the encounter begins.
- The ranks below fill a fixed table of combat roles, strongest first: Worldwrights, Architects, Runesmiths, Artificers, Tinkerers.
- Contributors past the table become overflow guardians. They hold the encircling line and retire wave by wave as the fight goes, so the room empties out and nobody is left standing after the boss.

Each contributor carries two things at once. Their own merged-PR rung drives who they ARE: their body, their colour, the weapon they hold, the title under their name. The role the table assigned them drives what they DO: level, health, damage, swing cadence, combat affix, which wave they arrive in.

Keeping those separate is what makes a self-generating dungeon tunable at all. The role table never changes size, so contributors joining the project or climbing the leaderboard change WHO you fight, never how much you fight. Somebody crossing 30 or 70 merged PRs earns a better body, a better title and a better weapon, and adds exactly zero attrition to the encounter.

### The fight

Contributors come at you in deterministic waves, weakest cohorts first, up to the boss alone at the end. The intermissions are short on purpose: long enough to reposition, too short to drink, so healer mana is a resource across the whole encounter rather than per wave.

The stone circle in the middle is a real boundary, and the fight is built on it. Stand on it and the dormant ring ignores you. Step off it and every remaining contributor in the room wakes at once, permanently. Strike a dormant contributor and you wake their entire cohort early. Splash an overflow guardian with an area attack and it joins the pull for good, adding itself to what you have to kill. There is no failure message anywhere in this room. The floor and the enemies say it.

The nameplates carry the two facts at two different times rather than crowding both onto one plate: name and rung while the contributors are still friendly, then level and a role-tinted elite diamond once the room turns. Overflow guardians deliberately stay plain, so a raid can see at a glance which ones are not part of the fight.

### What it drops

The reward chest stands sealed in its own alcove against the far wall from the moment you claim the room, in plain sight for the whole fight and completely inert: no label, no sparkle, and it refuses anyone who tries it. Clearing the full roster is what arms it, with a classic shared loot roll distributed through the normal group-loot machinery (need or greed, master loot, round robin), including themed rare and epic weapon rewards. The clear also grants its Book of Deeds records.

### How it holds together

The cave is one deterministic spec built from the roster and the world seed, so the same leaderboard produces the same room on every host: the authoritative server, the offline browser world, and the headless RL environment. Live GitHub data is used online, with a snapshot roster as the fallback when no leaderboard is available.

Balance is measured, not asserted: a deterministic scripted-raid probe (`npm run sim:source-cave-probe`) runs four 10-player compositions across 20 fixed seeds, 80 real Sim runs, and every profile clears at a median of 209 to 240 seconds.

Alongside the encounter this adds contributor-themed models and weapons, per-tier combat affixes, the room's props and lighting, map presence, HUD progress, target-frame treatment, and localized encounter text.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
`npm run gate`

- Manual steps:
  - `npm run build:server && ALLOW_DEV_COMMANDS=1 npm run server`
  - `npm run dev`
  - Connect with a lvl20 player
  - `/dev tp -175 483`
  - Interract multiple times with the well (F)
  - Interact with PNJ
  - Go to the center (alone or with your party) and press the button
  - Fight
  - Open the chest

## Screenshots / recordings

Screenshot from 2026-07-29 19-58-17

Screenshot from 2026-07-29 19-58-22

image

image

image

Screenshot from 2026-07-26 14-04-29

Screenshot from 2026-07-26 14-04-35

Screenshot from 2026-07-27 15-29-59

Screenshot from 2026-07-27 15-30-38

---

## Checklist

### Quality

- [X] **The full gate passes.** `npm run gate` is green. I added or updated decisive

### Cross-platform

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Every user-facing

### Hygiene

- [ x I didn't hand-edit generated files such as


ryan-foo:
Read the encounter code and both design docs (didn't run a live raid; the balance calls are inference from the numbers, the structural points are from the code).

The scripted-probe results in `tier_profiles.ts` (4 seeds, clear time, deaths, end-of-fight healer mana) are what made this reviewable without raiding it. Three things.

**1. Balance, the thing to actually playtest: the tune is razor-thin and it favors AOE comps hard.** Your probe has 3-healer optimal play finishing at 0-3% mana with 1-3 deaths. The layout funnels waves of 10 / 8 / 6 into the 10-unit seal, which is an AOE dream and a single-target tax. A mage/warlock comp clears each wave fast and banks healer mana; an all-hunter or all-rogue comp kills the same wave one mob at a time, stays alive ~2-3x longer, eats 2-3x the incoming swings, and OOMs the healers before the boss. The margin is already 0-3% mana for optimal AOE play, so I'd expect a single-target-heavy comp to tip from "clears" to "can't." You capped the architect cleave at 3 because the mob-side AOE was lethal; the player-side AOE dependency is the same axis, unexamined.

Concrete ask: run the probe once with a no-AOE comp (the all-hunter case) and see if it clears or is AOE-or-bust. If it's AOE-or-bust, a small HP shave on the swarm tiers or one extra second of intermission widens the comp window without moving the intended difficulty.

**2. Longer-term: the roster is data-driven, so difficulty drifts up over time.** The wave builder caps cohort size and adds waves as the contributor list grows, which bounds the opening pull, but more waves = longer fight = more attrition on a mana-bound encounter. A fight tuned to 0-3% mana today gets harder every time the project gains contributors. Worth deciding whether total waves (or total attrition) is capped independent of roster size, or it quietly becomes unclearable in a few months.

**3. The breach is unforgiving but fair.** I first worried the no-recovery full-wake could fire involuntarily, but the kit has no fears, knockbacks, or charges, and the sim has no collision push, so nothing can eject a player off the seal. It's a pure positioning check: deterministic, no reflex, agent-completable. The only note is legibility: the first breach with no log line and no partial state reads as "why did we instantly wipe." The seal going vivid red is the intended tell; a one-line combat-log cue ("The containment is breached!") would make the cause legible without softening the mechanic.

Nits:
- `encirclement-waves.md` lists waves of 8/8/8/6/6/1; the code cohorts are 10/8/6/3. Stale doc, reconcile so reviewers judge the shipped numbers.
- Force-starting with a living player outside the seal breaches on tick 1. The double-press confirm is the only guard; for an agent group where one bot is slow onto the seal that's a wipe vector. Fair "you were warned," but consider whether the first tinkerer wave is forgiving enough to recover from an accidental force-start.

For the 3-hunter playtest request: three hunters is below spec (this is tuned for ~10 players / 3 healers), so a 3-hunter attempt will read as "too hard" for reasons that aren't the encounter's fault. The meaningful test is a full raid with a single-target-heavy comp.

The one thing I'd resolve before calling it balanced: whether a non-AOE comp can clear it.

madmatah:
@ryan-foo Thanks a lot for your feedback, I'll try to apply it!

madmatah:
I addressed @ryan-foo feedbacks about balancing, and re-sync the branch with release/v0.29.0

madmatah:
Rebased with release/v0.32.0 that includes new maps.
The entrance has been moved to -177,481 for a better integration with new maps.

jgyy (commented):
Ambitious one. `src/sim/source_cave/` is a proper module-first subdirectory (dungeon, encounter, clear, loot, wire, spec, types) rather than logic dumped into `sim.ts`, and it comes with a genuinely heavy dedicated test suite across sim, wire, i18n, loot, encounter, and clear-state.

The two catalog-wide pins that are easy to forget were both updated correctly in the same change: `tests/deeds_content.test.ts` (deeds 221 to 224, renown 2730 to 2780) and `tests/world_api_parity.test.ts` (274 to 275 members), with comments explaining the delta. The three new deeds are marked `hidden: true` with a comment citing rule 8's spoiler strip, which also explains why there's no `/wiki` guide entry or `docs/screenshots` for this content: the dungeon is meant to be a secret, and the hidden-deed exemption is an existing, documented convention, not an oversight. Checked for stray `Math.random`/`Date.now`/`performance.now` and any DOM/render/ui/game/net imports in the new sim files: none found. No em or en dashes in anything the PR added.

`nit`: `docs/the-source-cave/*.md` (README, encirclement-waves, final-report, etc.) sit outside the usual `docs/design/` or `docs/prd/` convention used elsewhere. Not wrong, just worth folding in if these are meant to be durable docs rather than working notes.

Nothing blocking found in the sim/determinism/parity/deeds surfaces I checked.

madmatah:
@jgyy Thanks for the review!

On the `docs/the-source-cave/*.md` nit: that location is the documented convention
rather than a deviation from it. `docs/CLAUDE.md` states that a new doc lands in
`design/` (how a system should work), `prd/` (a feature spec), or **its program's
dir**, and the top level already holds a row of program dirs on that rule:
`api-pipeline/`, `achievements/`, `i18n-scaling/`, `balance/`, `architecture/`.

The reason the convention has that third slot is that this is a package, not a
single doc: a decision record (`state.md`, decisions D1 to D10 plus open items),
two encounter and room contracts (`encirclement-waves.md`, `friendly-reboot.md`),
and one historical report (`final-report.md`). `design/` and `prd/` are flat,
one-doc-per-subject dirs, so a multi-file package with its own internal structure
does not fit either.

They are durable docs rather than working notes, and this PR registers them as such:
commit e6ef4c04e adds the `docs/CLAUDE.md` row naming which files are living, which
one is history, and that balance numbers belong in `src/sim/source_cave/` and never
in the docs.

Happy to revisit if you would rather see program dirs nested under `design/` as a
repo-wide convention change, but that is a separate change touching every program
dir, not this PR.

jgyy:
Checked `docs/CLAUDE.md` and the program-dir convention (`api-pipeline/`, `achievements/`, `i18n-scaling/`, `balance/`, `architecture/`) is exactly as described. That resolves the nit, thanks for the writeup. Nothing further from me on this one.

madmatah:
Thanks @jgyy ! 
I just rebased with 0.34.0

jgyy (commented):
This is a fun idea, well executed on the sim side: pulling the raid entirely from the live contributor leaderboard while keeping `src/sim/source_cave/spec.ts` pure and host-agnostic (its own salted `Rng`, roster passed in as a parameter, the network call pushed to `server/github_contributors.ts` with a 30-minute cache and graceful fallback) is exactly the right shape for the determinism invariant. The Book of Deeds handling is also done thoughtfully: three hidden deeds correctly marked `hidden: true`, with a comment explicitly citing the structural strip rule that keeps them out of the wiki/rarity endpoint, backed by dedicated tests.

**blocking** - badly stale against its own stated base (`release/v0.35.0`), and current `main`/release has moved well past that. `git merge-tree` shows real content conflicts in `src/sim/spirit.ts`, `src/sim/pet/pet_ai.ts`, `src/render/renderer.ts`, `src/ui/icons.ts`, and `src/ui/map_window_view.ts`, plus every generated i18n file needing a regen rather than a hand-merge. This needs a full rebase and re-verification against the current release branch before it's reviewable end to end, not just a mechanical conflict resolve given the files involved.

**should-fix** - missing Reliquary pages for the raid's loot. Seven new items ship in `items.ts` (a rare group plus three epic weapons: `commit_blade`, `bug_squasher`, `mech_keyboard`), with item comments explicitly comparing them to Korzul the Gravewyrm's drops. Korzul is exactly the kind of dungeon whose rare/epic loot is tracked in `src/sim/content/reliquary.ts` (`korzul_the_gravewyrm: [...]`) per `docs/design/reliquary.md`. None of the seven new item ids appear in `reliquary.ts`, and `tests/reliquary_content.test.ts` wasn't touched. Unlike the deeds, there's no comment addressing why the loot was left off, so this reads like an oversight rather than a deliberate spoiler call, worth a decision either way (a spoiler-safe relic entry seems plausible without giving away the dungeon's identity, similar to how the hidden deeds handle it).

nit: the balance comments tying the three epic weapons' item level (29) to "one notch above Korzul's epics" should be re-checked against Korzul's current numbers post-rebase in case they've shifted since v0.35.0.

Great concept, just needs the rebase and the Reliquary question resolved before it's ready for a full pass.

PR #1788, opened 2026-07-11
fix(desktop): revert PR #1736 interface overhaul

Rubsey:
## Summary

- Prepares the v0.24.1 patch release from the current `main` tip.
- Reverts the net changes from #1736 after significant desktop regressions.
- Preserves unrelated changes that landed after #1736, including later nameplate rendering optimizations.
- Removes the reverted UI's downstream localization and test artifacts.
- Synchronizes all release version surfaces to v0.24.1.

This rolls back merge commit `feac4d7a1db830851d8cd357572285ed65007005` with its first parent as the baseline. It avoids replaying 195 individual reverts and preserves subsequent work on `main`.

## Related issues

Reverts the changes introduced by #1736.

## Type of change

- [ ] Feature: new functionality
- [x] Bug fix
- [ ] Refactor or performance (no behavior change)
- [ ] Documentation
- [x] Tests
- [x] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- `npm run gate` - all 9 steps passed; 952 test files passed, 1 skipped; 11,661 tests passed, 6 skipped
- `npm run release:check -- --version 0.24.1`
- Targeted PR-footprint regression suite - 34 files and 724 tests passed
- Preservation suite for post-PR behavior - 8 files and 120 tests passed, 1 skipped
- `npm audit` - 0 vulnerabilities
- Focused browser accessibility suite - 21/21 passed

The optional full browser suite retains three inherited failures in stale target-size fixtures for hidden party/community controls. They are unrelated to this rollback.

## Dev deployment

- Deployed `release/v0.24.1` to `dev.worldofclaudecraft.com`.
- Host revision advanced from `fa435903a` to `48f2d9723`.
- The deployment playbook completed with `failed=0` and created a pre-deploy database backup.
- Internal game and payout health checks passed.
- Public `https://dev.worldofclaudecraft.com/api/status` returned `{"ok":true,"realm":"Claudemoon","players_online":0}`.
- A fresh visual desktop smoke test remains for reviewer confirmation.

## Screenshots / recordings

N/A. This rollback restores the pre-#1736 interface; the dev deployment is available for visual review.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Fresh manual cross-platform smoke testing remains.
- [x] **Accessible.** The focused browser accessibility suite passes.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings.

### Hygiene

[...]

PR #1796, opened 2026-07-12
fix(game): migrate v0.24.0-poisoned keybind profiles on load

### LINKED ISSUE #1792: Keybind profiles saved on v0.24.0 carry the reverted overhaul's defaults, leaving strafe unbound
## Problem

Keybind profiles saved while v0.24.0 (the PR #1736 interface overhaul) was live permanently embed that build's experimental defaults as explicit bindings, and the v0.24.1 revert cannot undo them. `save()` in `src/game/keybinds.ts` persists every action including explicit `[null, null]`, and `load()` treats any stored array as authoritative ("explicit stored bindings still win"), so a profile written during the overhaul window carries its layout wholesale into the reverted build.

Concretely, #1736 shipped strafeLeft/strafeRight unbound (defaults `[]`), moved KeyQ/KeyE onto action slots 10/11, and moved Damage Meters from KeyH to KeyZ. Any player who changed any keybind during the roughly one day v0.24.0 was live (which rewrites the whole map) now has: strafe silently unbound (stored as explicit nulls, so the restored KeyQ/KeyE defaults never come back), Q/E firing ability slots 10/11 instead of strafing, and meters toggling on Z. The uniqueness sweep in `load()` also strips a restored default from an action when a stored binding claims the same code.

## Steps to reproduce

1. On v0.24.0, change any single keybind (this rewrites the full stored map with that build's defaults embedded).
2. Update to v0.24.1.
3. Press Q or E: an ability on slot 10/11 fires instead of strafing; strafe has no binding at all in Options.

## Actual behavior

- strafeLeft/strafeRight load as explicit `[null, null]` and stay unbound.
- KeyQ/KeyE remain claimed by action slots 10/11.
- Damage Meters stays on KeyZ while the current default is KeyH.
- Nothing surfaces this to the player; movement keys just stop working.

## Expected behavior

- Profiles written during the v0.24.0 window are detected and the affected actions are restored to current defaults, or the player is clearly prompted to reset keybinds.
- Deliberate custom bindings made before or after the v0.24.0 window are preserved.

## Scope

- `src/game/keybinds.ts` `load()` / `save()` (storage keys `woc_keybinds` / `woc_keybinds:<scope>`)
- A one-time signature migration is feasible: the combination of strafeLeft and strafeRight stored as explicit nulls, slots 10/11 on KeyQ/KeyE, and meters on KeyZ matches #1736's defaults and is vanishingly unlikely as a deliberate player layout; when matched, restore current defaults for exactly those actions
- Alternatively (or additionally) a release-note line pointing at the keybinds reset

## Acceptance criteria

- A profile saved under v0.24.0 defaults loads on current main with strafe bound to KeyQ/KeyE and meters on KeyH.
- A profile with genuine custom bindings (any layout not matching the full #1736 signature) loads unchanged.
- The migration runs at most once per profile and is covered by unit tests for both the matched and unmatched cases.

## Notes

- Root cause analysis from the #1736 revert audit: `save()` persists explicit nulls for unbound actions and `load()` honors them over defaults, which is correct behavior in general; the problem is only the poisoned cohort from the v0.24.0 window.
- The exposure window was roughly one day (v0.24.0 release to the v0.24.1 revert), so the affected cohort is small but the failure mode (silently unbound movement keys) is severe for those hit.

jgyy:
## Summary

Profiles saved during the roughly one day #1736 (the v0.24.0 interface overhaul) was live persisted that build's defaults wholesale: explicit `[null, null]` for `strafeLeft`/`strafeRight`, `KeyQ`/`KeyE` claimed by action slots 10/11, and Damage Meters moved to `KeyZ`. `save()` in `src/game/keybinds.ts` persists every action including explicit nulls, and `load()` treats any stored array as authoritative, so the v0.24.1 revert (#1788) could not undo this for anyone who saved during that window: strafe stayed silently unbound, Q/E kept firing ability slots 10/11, and meters stayed on Z.

This adds a one-time, targeted migration in `Keybinds.load()`: when the full poisoned signature matches together (all five actions at once: `strafeLeft`, `strafeRight`, `slot10`, `slot11`, `meters`), those five actions are restored to current defaults. A partial coincidental match (e.g. a player who genuinely unbound strafe but never touched slots 10/11 or meters) is left completely untouched, as is every other binding on the profile. The repair is self-limiting: once applied, strafe is bound again, so the poisoned signature can never match the same profile a second time (no separate migration flag needed).

Note: this PR is based on `main` rather than `release/v0.25.0` because the interface-overhaul revert (#1788, `a45d20eb3`) and its follow-up (#1793, `5369193a9`) have landed on `main` but not yet on `release/v0.25.0` — the "current defaults" this migration restores to only exist post-revert. Happy to rebase onto `release/v0.25.0` once that revert is carried over there.

## Related issues

Closes #1792.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/keybinds.test.ts` (49 passed, including 4 new cases covering: full-signature restore, partial-match preservation, other-binding preservation, and single-run idempotency)
  - `npx tsc --noEmit -p .`
  - `npx @biomejs/biome check --write src/game/keybinds.ts tests/keybinds.test.ts` (no new warnings/errors introduced)
  - `npx vitest run` (full suite; the only failures are pre-existing environment artifacts unrelated to this change: missing generated `i18n.status.json` and an unrelated Svelte testing-library resolution issue in the admin test suite, both fixed by `npm run i18n:gen` / a full `npm install`)
- Manual steps: N/A (pure data-layer/persistence logic, no DOM)

## Screenshots / recordings

N/A. This change is entirely internal keybind-persistence logic with no visible UI change.

---

### Quality

- [x] **The full gate passes.** Ran the equivalent checks manually (tsc, targeted and full vitest, biome) in this environment; added decisive tests for the new migration behavior.

### Cross-platform

- N/A. No UI/DOM surface touched.

### Localization (i18n)

### Hygiene


jgyy:
Closing - #1793 already covers this.

PR #1802, opened 2026-07-12
fix(sim): never destroy loot-roll winnings when the winner is offline

jgyy:
## Summary

Player reports (screenshots below): during heroic dungeon runs, need/greed rolls on epic
drops "bug the interface" and the item is permanently lost, forcing a relog that doesn't
bring it back. The report is consistent across multiple heroic runs on different characters.

Root cause: a need/greed roll waits up to 60s before it auto-resolves
(`LOOT_ROLL_TIMEOUT`), and a curated master-loot assignment waits up to 5 minutes
(`MASTER_LOOT_TIMEOUT`). Both `resolveLootRoll` and `assignMasterLoot`'s single-target
branch granted the item with an unconditional `ctx.addItem(itemId, 1, winnerPid)`. If
the winning candidate (or the sole master-loot target) disconnected any time before
resolution, `addItem`'s `resolve()` silently no-ops for a pid with no live player/entity
record, and the item vanishes: it is never left on the corpse, never mailed, never given
to anyone else. This directly violates the invariant documented at the top of
`awardSharedLootItem` ("items are never destroyed") — the winner-grant path was the one
place that promise didn't hold.

In a busy heroic group it's common for someone to lag out or log off mid-run, so a roll
whose winner happens to be that player quietly destroys the drop for the whole group with
no error, no toast, and no way to recover it short of a GM ticket.

```mermaid
sequenceDiagram
    participant A as Player A (wins roll)
    participant B as Player B
    participant S as Sim (server)
    A->>S: submitLootRoll(need)
    Note over A,S: Player A disconnects (logout / linkdead)
    S->>S: removePlayer(A) — deletes A's player + entity records
    B->>S: submitLootRoll(pass) (last vote)
    S->>S: resolveLootRoll(): highest roll = A
    rect rgb(255, 220, 220)
    S->>S: addItem(itemId, 1, A) -> resolve(A) fails -> silent no-op
    Note over S: item destroyed, never on corpse, never in any bag
    end
```

```mermaid
sequenceDiagram
    participant A as Player A (wins roll)
    participant B as Player B
    participant S as Sim (server, fixed)
    A->>S: submitLootRoll(need)
    Note over A,S: Player A disconnects (logout / linkdead)
    S->>S: removePlayer(A) — deletes A's player + entity records
    B->>S: submitLootRoll(pass) (last vote)
    S->>S: resolveLootRoll(): highest roll = A
    S->>S: isPidResolvable(A)? false
    rect rgb(220, 255, 220)
    S->>S: returnLootRollItemToCorpse() + "A was offline" loot line
    Note over S: item survives on the corpse for the group to loot again
    end
```

## Related issues

Reported via Discord (screenshots attached by the player: 27/26 bag overflow after a
force-added roll grant, the roll bugging out mid-heroic, and losing gear after a relog).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/loot_roll.test.ts` (new regression test:
    "never destroys the item when the winner disconnects before the roll resolves")
  - `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts`
  - `npx vitest run tests/parity tests/loot_messaging_sim.test.ts tests/loot_master_sim.test.ts
    tests/loot_roll_reconcile.test.ts tests/loot_roll_status_view.test.ts tests/fixes.test.ts
    tests/world_api_parity.test.ts`
  - `npx tsc --noEmit`
  - `npm run i18n:gen`
- Manual steps: N/A (server/sim logic only, no UI surface change).

## Screenshots / recordings

N/A — this is a sim/server logic fix with no rendered UI change.

---

### Quality

- [x] Added a decisive regression test reproducing the exact race (winner votes, then
      disconnects, then the group's last vote resolves the roll) and asserting the item
      survives on the corpse rather than vanishing.

### Cross-platform

- [x] N/A. No UI surface touched.

### Localization (i18n)

- [x] Added one new player-visible sim string (`loot.rollWinnerOffline`, the "X was
      offline; item returned to the corpse" loot line) as an English-only key in
      `src/ui/sim_i18n.ts`'s `baseEnTable` (falls back to English for every locale via the
      existing `tSim` fallback chain), plus the matching `RULES` regex so
      `tests/localization_fixes.test.ts` (the S3 drift guard) recognizes the new emit site.
- [x] Player-facing text emitted from `src/sim/` is re-localized at the client boundary in
      this same change; `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets/credentials touched.
- [x] No generated files hand-edited (`src/ui/i18n.status.summary.json` was regenerated via
      `npm run i18n:gen`, not hand-edited).


FernandoX7 (commented):
Solid fix, and the diagnosis is right. Both `resolveLootRoll` and the master-loot single-target branch handed the drop to `ctx.addItem(itemId, 1, winnerPid)`, and `addItem` -> `resolve()` silently no-ops for a pid no longer in `players`/`entities` (sim.ts:5173, sim.ts:2548), so a winner who logged out during the roll window destroyed the drop for the whole group. I traced the race end to end: `removePlayer` deletes both records (sim.ts:2090-2091) without touching the pending roll, so at resolution the winner is unresolvable and the old code lost the item outright.

What I like:

- The guard sits after the tiebreak `rng.int` draw, so the rng stream and the parity goldens stay byte-identical whether or not the winner turns out to be offline. Reusing `returnLootRollItemToCorpse` (need/greed) and `convertMasterRollToNeedGreed` (master loot) means both fallbacks are already-proven non-destructive paths, not new behavior.
- Both resolution triggers are covered: the last-vote path (`submitLootRoll`) and the timeout path (`updateLootRolls` -> `resolveLootRoll`, sim.ts:3317).
- The regression test is decisive. I stripped the guard and it fails with the item gone from the corpse (`returned` undefined), then passes with it restored. `tests/loot_roll.test.ts`, `tests/parity`, `tests/localization_fixes.test.ts`, and `tests/architecture.test.ts` are all green, and `npm run i18n:gen` leaves the tree clean, so the committed summary is fresh.

Findings:

- should-fix, `src/ui/i18n.status.summary.json`: the PR is CONFLICTING against the base here. This is the standard stale-base regen conflict, not a design issue: merge `release/v0.25.0` and re-run `npm run i18n:gen`. The `sim_i18n` key does not feed `i18n.resolved.generated` or `i18n.resolved.sha256` (i18n:build reads only the catalog + locale overlays), so nothing else needs regenerating.
- nit, `src/sim/loot/loot_roll.ts:661`: the `isPidResolvable` comment says `resolve()` "also requires the caller to be alive." It does not: the private `resolve()` (sim.ts:2548) only checks that the meta and entity records exist, so this helper is exactly `!!ctx.resolve(pid)`. Keeping the helper is fine (it skips the result-object alloc), but the stated reason for not reusing `ctx.resolve` is inaccurate.
- nit, `src/sim/loot/loot_roll.ts:647`: the departed winner renders as "Unknown" in the new "was offline" line (and the existing "wins" line above it), because `removePlayer` already deleted them from `players` before resolution, so `ctx.players.get(winner.pid)` is undefined. Cosmetic; if you want the real name in that line, snapshot it when the roll opens.

Approvable once the i18n conflict is resolved. The two nits are optional.

jgyy:

- The i18n conflict was a stale-base issue as you diagnosed; the branch already had `release/v0.25.0` merged in and `i18n:gen` re-run (see the merge + regen commits ahead of your review). PR now shows `MERGEABLE`.
- Fixed both nits in 73d1566d8:
  - `isPidResolvable`'s comment no longer claims `ctx.resolve` additionally requires the caller to be alive. It's exactly `ctx.resolve`'s existence guard; the helper only exists to skip the result-object allocation.
  - The departed winner no longer renders as "Unknown": `PendingLootRoll` now snapshots each candidate's name (`candidateNames`) when the roll opens, and both the per-roll "Need/Greed Roll by X" lines and the winner/offline lines fall back to that snapshot instead of the live (by-then-deleted) `ctx.players` lookup.

Reran `tests/loot_roll.test.ts`, `tests/loot_messaging_sim.test.ts`, `tests/loot_master_sim.test.ts`, `tests/loot_roll_reconcile.test.ts`, `tests/loot_roll_status_view.test.ts`, `tests/world_api_parity.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`, plus `tsc --noEmit`: all green. (`tests/fixes.test.ts`'s Varkas Boneguard despawn test times out on this branch, but it reproduces identically on the unmodified tree before my change, so it's a pre-existing machine-speed flake, not a regression from this fix.)

FernandoX7 (approved):

PR #1805, opened 2026-07-12
feat(ui): clarify wolf-targeting instructions in tutorial

jgyy:
## Summary

Playtester feedback: the wolf tutorial step ("Wolves at the Door") tells a new
player to hunt the Forest Wolves but never says HOW to engage one. This adds a
short combat hint to the end of that step's body naming the actual bound
target key (Tab by default, `src/game/keybinds.ts`, id `target`) or a click,
then pointing at the action bar to attack. Keyboard/touch each get their own
copy, matching the existing `*Touch` pattern already used for the other
tutorial steps.

No new highlight/pulse mechanism was built for this: I searched for an
existing reusable keybind-callout/highlight seam in the HUD (`pulse`,
`highlight`, `glow` across `src/ui` and `src/sim/content`) and found nothing
generic enough to key off a single instance without adding a new mechanism, so
per this repo's module-first rule I did not invent one for this single use.
See "Follow-up" below for a request to add one.

## Related issues

N/A (playtester feedback, no existing issue).

## Type of change

- [x] Feature: new functionality (tutorial copy + `t()` key)
- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/tutorial_copy.test.ts` (new + updated cases, all green)
  - `npx vitest run tests/i18n_resolved_equivalence.test.ts` (baseline hash regenerated
    via `node scripts/i18n_resolved_hash.mjs --write` after the new `hud_chrome` key)
  - `npx tsc --noEmit`
  - `npx biome ci --changed --since=up/release/v0.25.0 --no-errors-on-unmatched` (only
    pre-existing warnings on lines this change did not touch; no errors)
  - Full `npx vitest run` (bounded workers)
- Manual steps:
  - Ran the offline tutorial via `scripts/tutorial_shot.mjs` against a local `vite` dev
    server and confirmed the slay step now reads: "Hunt the Forest Wolves on the north
    road. Target one and attack to thin the pack. New to combat? Press Tab or click a
    wolf to target it, then use an action bar ability to attack it."

## Screenshots / recordings

Slay step ("Wolves at the Door") showing the new targeting/attack hint appended to the
objective body:

wolf tutorial step with targeting hint

## Follow-up (not in this PR)

A highlighted/pulsing keybind callout tied to the tutorial's current step (for example,
lighting up the Tab key and action bar slot 1 while the slay step is active) would make
this even clearer, but there is no existing reusable highlight-a-keybind mechanism in the
HUD to key off a single instance, and building one for this single use would violate this
repo's module-first / rule-of-three convention. Filing this as a follow-up feature request
is a good next step rather than bundling a new mechanism into this text fix.

```mermaid
flowchart TD
    A[Tutorial starts: fresh character] --> B[Step 1: move]
    B --> C[Step 2: seek Marshal Redbrook]
    C --> D[Step 3: talk, accept Wolves at the Door]
    D --> E[Step 4: slay - objective body shown]
    E --> F[NEW: targeting/attack hint appended:\nPress Tab or click a wolf, then use an action bar ability]
    F --> G{Player targets a wolf}
    G -->|Tab / click| H[Wolf targeted]
    H --> I[Player uses action bar ability]
    I --> J[Wolf takes damage]
    J -->|8 wolves slain| K[Step complete, advance to return]
    G -.->|before this change: no guidance| L[Player unsure how to engage]
```

### Quality

- [x] **The full gate passes.** `npm run gate` is green (biome's `defaultBranch` in
      this environment is a stale fork `origin/main`, so `ci:changed` was verified
      instead with `--since=up/release/v0.25.0`, which is the actual base this PR
      targets; see the biome command above). I added/updated decisive tests.

### Cross-platform

- [x] **Tested on desktop and mobile.** The touch variant (`slayTargetHintTouch`) has
      its own dedicated test asserting it never references keyboard/mouse or
      `{targetKey}`.
- [x] **Accessible.** Plain text appended to the existing tutorial body element; no new
      interactive controls.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Added
      `hudChrome.tutorial.slayTargetHint` / `slayTargetHintTouch` to
      `src/ui/i18n.catalog/hud_chrome.ts` (English only). No locale overlays touched.
- [ ] N/A: no numbers/money/dates in this copy.
- [ ] N/A: no sim/server-originated text in this change.

### Hygiene

- [x] No secrets committed.
- [x] No hand-edited generated files (the i18n resolved artifacts and the sha256
      baseline were regenerated via the documented scripts, not hand-edited).


FernandoX7 (commented):
The wolf-targeting hint is a good, well-scoped fix, and the approach is clean: the hint lives in its own pure `tutorialSlayHintPlan` rather than growing `renderPanel`, it splices the live bind via `{targetKey}` instead of hardcoding "Tab", and the keyboard/touch split follows the existing `*Touch` pattern. Tests and screenshot are in place.

One real issue and one mechanical one before this can merge.

**should-fix: the appended hint goes stale if Interface Mode is toggled during the slay step.** `tutorialStepDiffersByTouch('slay')` returns false (`src/ui/tutorial_copy.ts:68`) because the slay body key is identical on both modes. But this change makes the slay step's rendered card mode-dependent through the appended hint (keyboard "Press {targetKey} or click a wolf" vs touch "Tap a wolf"), so `tutorialNeedsRerender('slay','slay', prev, next)` returns false on a mode flip (`tutorial.ts:129`), `renderPanel` is not called, and the `else if (this.step === 'slay')` branch only refreshes the kill counter. A player who toggles Interface Mode mid-slay then keeps the previous mode's hint: a touch player can be told to "Press Tab or click a wolf", a keyboard player to "Tap a wolf", for the rest of the slay phase. This is the exact stale-control-copy case `tutorialNeedsRerender` exists to prevent for the other steps. The existing tests now encode it: `tests/tutorial_copy.test.ts:74-76` asserts no rerender on a slay mode toggle, `:91-92` asserts slay does not differ by touch, and the comment at `:72-73` ("seek/slay read identically on touch and keyboard") is no longer true. Fix: have `tutorialStepDiffersByTouch` (or the rerender gate) treat slay as mode-dependent now that its card is, and update those two tests.

**Mechanical, not a design issue: stale-base i18n conflict.** The only real merge conflicts are `src/ui/i18n.resolved.sha256` and `src/ui/i18n.status.summary.json`; the rest auto-merge. That is regen churn from the base moving, not a problem with the change. Merge `release/v0.25.0` and re-run the i18n build to regenerate the trio. The sha256 is a single well-formed line and resolved `en` carries the new keys, so nothing was hand-edited.

Nothing else stands out: the key is in the right catalog, the placeholder syntax matches the existing `{interactKey}`/`{moveKeys}` keys, and the targeted test is green.

jgyy:
Thanks for the review! Both points addressed in 1a12818b6:

- **should-fix (stale hint on Interface Mode toggle):** `tutorialStepDiffersByTouch` now also compares `tutorialSlayHintPlan`'s body key for the `slay` step, not just the step's own body key. Since `tutorialNeedsRerender` gates on `tutorialStepDiffersByTouch`, a mode flip mid-slay now correctly triggers `renderPanel`, which rebuilds the hint text for the new mode. Updated the two tests you flagged (`tests/tutorial_copy.test.ts:74-76`, `:91-92`) and their comments to match: slay is grouped with move/talk/return/done as mode-dependent, and only `seek` remains mode-agnostic.
- **mechanical (stale-base i18n conflict):** merged the current `release/v0.25.0` tip and regenerated `i18n.resolved.sha256` / `i18n.status.summary.json` via `npm run i18n:gen` + `scripts/i18n_resolved_hash.mjs --write` rather than hand-resolving.

Verified with `npx vitest run tests/tutorial_copy.test.ts tests/tutorial.test.ts`, `npx tsc --noEmit`, and the pre-push QA floor (green on push).

FernandoX7 (approved):

Issue #1777, opened 2026-07-11
Release v0.25.0


Rubsey:
### Target version

0.25.0

### Release branch

`release/v0.25.0`

### Bump the version

- [ ] Ran `npm version <X.Y.Z>` (or `npm version minor`/`patch`/`major`) and pushed the resulting commit and tag.
- [ ] Confirmed all three manifests show the new version and the native build numbers incremented.

### Verify

- [ ] `npm test` passes.
- [ ] `npm run build` succeeds.
- [ ] Ran the release malware-audit gate.

### Ship

- [ ] Web deploy is live (see `DEPLOY.md`).
- [ ] Mac desktop Electron app release artifact is published.
- [ ] iOS archive built via `npm run native:open:ios` and submitted to App Store Connect.
- [ ] Android bundle built via `npm run native:open:android` and submitted to Play Console.
- [ ] Release notes added under `docs/release-notes/`.

### Carryover from v0.24.0

Carried over from the v0.24.0 release issue (#1698): every PR and issue referenced there (including in the comments) that is open and not yet merged or closed, plus every other open PR still targeting the retired `release/v0.24.0` branch. Nothing targets `release/v0.25.0` yet, so every PR below needs retargeting.

Not part of this migration: the PTR track (#1744 `release: v0.24.0 PTR` and PRs based on `release/v0.24.0-ptr`, such as #1725) continues separately.

#### Open PRs targeting `release/v0.24.0` (need retargeting to `release/v0.25.0`)

- [ ] #1776 feat(social): chat mute, clickable player names in chat, remote player info
- [ ] #1772 Feature/sfx mob family fallback
- [ ] #1771 feat(sfx): variant naming system / manifest rebuild mode
- [ ] #1770 feat(ui): overhead chat bubbles for party channel messages (#1659)
- [ ] #1769 feat(ui): first-run camera-mode prompt (#1727)
- [ ] #1768 feat(sim): emit a gather-success SimEvent for audio hooks (#1729)
- [ ] #1766 fix(dungeons): allow eligible ghosts to reclaim raid loot
- [ ] #1765 feat(game): Z-key weapon sheathing with on-back weapons, arm gesture, and SFX
- [ ] #1763 feat(sfx): add conform pipeline script (192kbps/44.1kHz, hybrid norma…
- [ ] #1761 feat(mobile): redesign the compact touch HUD [draft]
[...]

PR #1806, opened 2026-07-12
fix(heroics): make rewards and mobile loot reliable

Rubsey:
## Summary

Heroic final-boss kills could persist a lockout while losing the corresponding Heroic Mark, corpse loot, or Need/Greed award. On short mobile screens, opening Bags during a roll could also leave the pointer-active roll rail above the bag sheet and make the UI appear locked.

This change:

- Awards Heroic Marks directly at final-boss death and uses the same realm-local reset boundary as the heroic lockout, removing the former UTC-day mismatch.
- Uses a claim-bound participation snapshot for living and released players, so releasing, instance-slot reuse, party changes, and logout cannot create a lockout-without-reward result.
- Freezes reward eligibility before the disconnect save yields, conserves pending and master-loot items, preserves corpse rights and party loot strategy when the original tapper leaves, and blocks stale post-snapshot commands.
- Reconciles server-authoritative roll prompts after a dropped submit and shows only actionable roll cards on short mobile screens.
- Keeps Bags above the mobile roll rail while preserving the existing desktop order where the active roll remains above Bags.
- Preserves `main`'s one-time frame-position reset and excludes the unrelated release-only SFX changes.

The heroic drop tables are intentionally unchanged. Each heroic final boss still has two guaranteed dedicated epic roll groups. The 27/26 bag state is also retained as the intentional asynchronous-reward safety valve, not treated as the cause of the loss.

## Related issues

Part of #1777

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run gate`: passed all 9 steps, including 953 test files passed and 1 skipped; 11,675 tests passed and 14 skipped.
  - Heroic, loot, logout, and UI focused suite: 20 files and 364 tests passed.
  - `npm run test:browser -- tests/browser/heroic_mobile_loot.browser.test.ts`: 3 tests passed.
  - `npx tsc --noEmit`: passed.
  - `npm ci`: completed with 0 vulnerabilities.
  - `git diff --check origin/main..HEAD`: passed.
- Manual steps:
  - Reproduced the reported short landscape phone layout at 844x390 with one answered roll, one actionable roll, and Bags open.
  - Before: Bags computed at z-index 51 under the roll rail at 65; the rail intercepted the bag hit target.
  - After: Bags computed at z-index 95, answered watch rows were removed, one actionable prompt remained, and the hit target was inside Bags.
  - Verified the desktop inverse separately: Bags remains at 51 and the roll rail remains at 65.

### Browser baseline note

The full browser suite passed 55 of 58 tests. The remaining three target-size failures reproduce identically on clean `origin/main`: zero-width measurements for the party frame, party leave button, and community toggle. The new heroic mobile browser spec passes 3/3 both in isolation and in the full run.

### Before

Mobile Bags below the heroic loot roll rail

### After

Mobile Bags above the actionable heroic loot prompt

---

### Quality

- [x] **The full gate passes.** All 9 local gate steps are green on the `main`-based branch.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop and short landscape mobile are covered; portrait was not part of this reported regression.
- [x] **Accessible.** This change adds no controls or player-facing copy; existing keyboard and ARIA behavior is unchanged.

### Localization (i18n)

- [ ] **New player-visible strings follow the contributor policy.**
- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary in this same change.

### Hygiene


jgyy:
lol so many commit to fix it

jgyy (commented):
Nice fix, and a genuinely thorough one. The reward/lockout race here had a few independent failure modes (UTC-day vs realm-reset mismatch, corpse identity tied to reusable instance slots, tap/loot state not frozen before the disconnect save await) and this closes all of them without touching the drop tables or the intentional bag-safety-valve behavior, which is the right scope. The test suite backs it up: `game_sessions.test.ts` actually asserts state *during* the persistence await window (trade cancelled, loot forfeited, tap re-anchored to the stayer) rather than just before/after, which is the part that's easy to fake and here isn't.

Traced the four races described in the summary against the code, not just the PR text:
- Release mid-fight: `corpseInstanceId` is session-only (defaults null, not persisted, re-derived on load), so a released ghost correctly falls back to `e.pos` with no stale-epoch collision.
- Slot reuse: binding to `exitId` instead of reusable slot coordinates closes the recycled-slot corpse match.
- Logout/leave: the `leaving` freeze lands synchronously before the first persistence await and is threaded through every recipient/tap/loot path.
- Double-award: `awardHeroicMarks` checks-then-locks in a single pass, and a second final-boss death can't exist per dungeon instance, so no new double-award path opens up.

Determinism, sim purity, and wire/entity parity all check out: `corpseInstanceId` is correctly server-local-only (mirrored in `blankEntity`, never touched by render/ui), and the removed `sharedPersonal` LootSlot field has no remaining referents anywhere. The mobile z-index fix (Bags at 95 above the roll rail, scoped to `body.mobile-touch`) is pinned by real assertions in `client_shell.test.ts`, not just a smoke test, and desktop stacking is untouched.

Two non-blocking notes:

- **nit** `src/sim/sim.ts` (`preparePlayerLeave` / `replacementTapperForLeave`, roughly lines 2106-2168): this is new game-system logic (tap re-anchoring, corpse-rights preservation on leave) added as a method cluster directly on the `sim.ts` coordinator rather than a module behind the `SimContext` seam. It leans on live `Sim` state heavily so keeping it inline is defensible, but per the module-first guidance in `src/sim/CLAUDE.md` this reads like a good candidate to extract into its own tested module if it grows further.
- **nit** `src/styles/hud.mobile.css` (`.loot-roll-rail`/Bags z-index area): on desktop, a focused Bags window can still rise above the roll rail via the normal focus-to-front `windowZ` climb (89) past the rail's 65 - the pinned test only asserts the resting floor (50). That's existing, intended focus behavior this PR doesn't change, just flagging that the "desktop order preserved" claim is about resting z-index, not focused.

Nothing blocking. Good to merge as far as sim determinism, wire parity, and UI seam go.

FernandoX7 (commented):
The heroic settlement rework is solid. Settling marks and the lockout together at kill time, with the realm-reset lockout as the only income gate, removes the whole class of lockout-without-reward outcomes, and the ordering is right: `awardHeroicMarks` has to run before `grantNythraxisLockout` because both write the same `:heroic` key the pay gate reads, and nothing between the old and new call sites reads lockouts. Binding released corpses to the recreated exit entity is a clean answer to recycled slots, and `preparePlayerLeave` as a synchronous prefix before the disconnect save closes the real races here.

I verified the claims against the head: the changed suites pass locally (155 tests), the sim parity goldens match, `corpseInstanceId` is correctly server-local with the `blankEntity` line in place (`corpsePos` still rides the wire for the corpse marker), `sharedPersonal` has zero remaining references, `leaving` is session-only and cannot strand a live player (`saveCharacterOnLeave` never rethrows, so `removePlayer` always runs), and both SimContext fake hosts carry `instanceClaimIdAt`.

One should-fix, on the reconcile change:

- **should-fix** `src/ui/loot_roll_reconcile.ts:46`: dropping the dismissed suppression makes every successfully answered roll re-show online for one server round trip. The tap path dismisses and sends the command (`src/ui/hud.ts:10852`), but reconcile runs every frame against the snapshot mirror (`src/net/online.ts:2161`), which keeps listing the roll until the server records the choice, so `toShow` re-creates the prompt with live buttons; it then auto-retires when the mirror drops it. Offline never hits this (the local sim records the choice synchronously), which is why the browser test and the screenshot script cannot catch it. It is harmless state-wise (the `choices.has` guard in `submitLootRoll` makes a double tap a no-op), but it is a visible flash on every answer, longer on mobile networks, on exactly the rail this PR is fixing. The locally-expired arm inherits the same behavior (`src/ui/hud.ts:10957`): a roll that expired locally but is still open on the server now re-shows with a fresh countdown. A short post-dismiss grace (re-show a dismissed roll only if the mirror still lists it open after a second or two) keeps the dropped-submit recovery without the flash.

Two nits:

- **nit** `src/ui/hud.ts:10958` and the `dismissedLootRolls` field comment at 1125: both still describe the old "don't let reconcile re-show it" contract this PR removed.
- **nit** `src/sim/loot/loot_roll.ts:504`: the one forfeit branch that completes a roll (the leaver was the last unanswered candidate) has no direct test, and it is the only leave-path branch that can draw rng (the tie-break in `resolveLootRoll`). The new loot_roll test resolves via the last submit instead. Worth one test where the leave itself resolves a roll with two needers.

Everything else I checked came back clean: the `handleMessage` guard is ordered correctly (`session.left` is set synchronously before the first await in `leave()`, and the clients-map identity check still admits a fast rejoin's new session), `heroicDaily` round-trips as a legacy payload, marks land through the normal inventory path so the self-wire ships bags like any other grant, and the `client_shell` z-order pin plus the browser stacking tests are good regression coverage for the mobile bag sheet.

PR #1807, opened 2026-07-12
fix(ui): stop Enter from inserting a newline when focusing chat input

jgyy:
## Summary

Playtester feedback: pressing Enter to open/focus chat (WoW-style) inserted a literal newline
into the composer instead of just focusing it, so the placeholder ("Say something...") was
never visible until the player clicked away and back.

Root cause: opening chat focuses the composer `<textarea>` as a **side effect** of the very
`Enter` keydown that triggers it (`Input.onKeyDown` -> `dispatchEdge('chat')` ->
`onUiKey('chat')` -> `openChat()` -> `chatInput.focus()`), all synchronously within that one
keydown's handling. Because that keydown's default action was never cancelled, the browser
still delivered the trailing keypress/input for the *same physical key press* to whichever
element was focused **at that point**, i.e. the textarea we had just focused, and ran its
default action there: inserting a newline. The composer's own `keydown` handler (which does
call `preventDefault()` on Enter) never got a chance to help, because it only runs once the
box is already focused and the player types a *second* Enter; it can't retroactively cancel the
keypress trailing the keydown that opened it.

Fix: cancel the keydown's default when it resolves to the `'chat'` edge action, in
`src/game/input.ts`, the same way an already-modified chord cancels its browser accelerator
default a few lines above. This is the smallest change that stops the browser from acting on
the stale keypress against the newly-focused element.

```mermaid
sequenceDiagram
    participant Player
    participant Browser
    participant InputTS as Input.onKeyDown (window keydown)
    participant MainTS as openChat()
    participant Textarea as #chat-input

    Player->>Browser: presses Enter (chat not open)
    Browser->>InputTS: keydown "Enter" (target: body)
    InputTS->>InputTS: edge = "chat"
    Note over InputTS: BEFORE FIX: no preventDefault()
    InputTS->>MainTS: onUiKey("chat")
    MainTS->>Textarea: focus()
    Note over InputTS,Textarea: keydown handling returns, but the SAME key press<br/>still has a trailing keypress/input to deliver
    Browser->>Textarea: keypress/input "Enter" (target now resolves live,<br/>i.e. the just-focused textarea)
    Textarea->>Textarea: BEFORE FIX: inserts "\n" (default action)
    Textarea->>Textarea: AFTER FIX: preventDefault() on the keydown<br/>suppresses the trailing default too, value stays ""
```

## Related issues

N/A (playtester feedback, no tracked issue).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/input.test.ts` (added two cases under "Input chat keybind": the
    default Enter key dispatches `onUiKey('chat')`, and the keydown's `preventDefault` is
    called so the browser cannot also treat it as a newline in the newly-focused composer)
  - `npm run gate` (see notes below)
- Manual / E2E:
  - The regression is a real browser DOM event-propagation quirk (a keydown handler focusing
    an element mid-dispatch still receives that same physical key's trailing keypress/input),
    not something jsdom/Vitest can observe end to end, so it was also verified with a
    puppeteer-core script against `npm run dev`: boot offline, skip the one-time spawn intro
    (which otherwise swallows all keys except Escape), then dispatch a real, CDP-trusted Enter
    keydown via `page.keyboard.press('Enter')` (a synthetic `dispatchEvent` does **not**
    reproduce this, since untrusted events never trigger a browser's native default actions
    like text insertion) and read back `#chat-input`'s value/placeholder/focus state.
    - Before the fix: `{"value":"\n","placeholder":"Say something...","focused":true}` -- chat
      opens but the composer already contains a newline, hiding the placeholder.
    - After the fix: `{"value":"","placeholder":"Say something...","focused":true}` -- chat
      opens empty, placeholder visible.

## Screenshots / recordings

Composer immediately after pressing Enter (same bottom-left crop used by the existing
chat-input screenshot scripts):

| Before | After |
| --- | --- |
| before | after |

Before: the box already holds a newline (blank first line, cursor pushed down, placeholder
gone). After: empty box, placeholder visible.

---

### Quality

- [x] **The full gate passes.** `npm run gate` step-by-step, run locally in a dedicated
      worktree off `up/release/v0.25.0`:
  - `i18n:gen`, `security:gate`, `vitest (full suite)`, `tsc --noEmit`, `build:env`,
    `build:server`, `build` all pass clean.
  - `sfx:check` fails identically (111 pre-existing out-of-spec audio assets) on a clean
    checkout of `release/v0.25.0` with none of this PR's changes applied, confirmed by
    stashing the diff and re-running; unrelated to this change.
  - `ci:changed` (`biome ci --changed`) diffs against `biome.json`'s configured
    `vcs.defaultBranch` (`origin/main`); in this contributor's local fork `origin/main` trails
    `release/v0.25.0` by many merged PRs, so the local run flags hundreds of pre-existing files
    outside this diff. Scoped instead to the actual changed files
    (`git diff --name-only up/release/v0.25.0` -> `src/game/input.ts`, `tests/input.test.ts`):
    `npx @biomejs/biome check`/`format` on exactly those two files show zero errors and zero
    format diffs (only pre-existing lint *warnings* on lines this PR didn't touch, which the
    gate policy treats as non-blocking).
  - A handful of `tests/server/**` DB/timing-sensitive tests (`game_boot_order`,
    `http/characterization` leaderboard cases) also failed during the full run under heavy
    machine-wide contention (many concurrent `vitest` runs from unrelated sessions, load
    average ~50) and against a local Postgres container provisioned for a different project;
    they reproduce identically on a clean `release/v0.25.0` checkout with this diff stashed, so
    they are pre-existing/environmental, not a regression from this change.
  - Decisive tests added for the changed behavior (see above).

### Cross-platform

- [x] **Tested on desktop and mobile.** The fix is a single `preventDefault()` call in the
      shared `Input.onKeyDown` keyboard pipeline; Enter-to-open-chat is a desktop/keyboard-only
      interaction (mobile opens chat via the on-screen Chat button, not this Enter path), so no
      separate mobile viewport check applies here.
- [x] **Accessible.** No focus/ARIA behavior changed; the composer still receives focus exactly
      as before, just without the stray inserted character.

### Localization (i18n)

### Hygiene

      any production path.


FernandoX7 (commented):
Clean fix with an accurate root cause. `openChat()` focuses `#chat-input` synchronously inside the same Enter keydown (`src/main.ts:1064`), and since that keydown's default was never cancelled the browser delivered the trailing default text-insertion to the just-focused textarea. Cancelling the default is the right, minimal fix, and it mirrors the modifier-chord `preventDefault` on the line directly above it.

A few things I checked and they hold up:

- Scoping is correct. `chat` is the only edge action in `dispatchEdge` that focuses a text composer (I confirmed `discord` and the rest just toggle windows), so the narrow `if (edge === 'chat')` is right rather than a blanket preventDefault.
- Both default binds are covered: `chat` defaults to `Enter` and `NumpadEnter`, and both route through this branch.
- The tests are decisive. For a bare `Enter` the new `chat` branch is the only `preventDefault` path in `onKeyDown` (Tab/Space/modified-chord branches don't apply), and the first test pins that the edge actually resolves to `chat`, so a revert would fail the second test rather than pass vacuously. Good call using the puppeteer path for the real browser behavior and pinning the `preventDefault` proxy in Vitest, since jsdom can't observe the trailing keypress.

One thing worth knowing, not blocking:

- nit (`src/game/input.ts:756`): the early guard only exempts `input`/`textarea`, so if a non-text control (a focused `<button>`) has focus during gameplay, this now also cancels that control's Enter activation while still opening chat. It is a marginal, pre-existing tension (before this change Enter both activated the control and opened chat), and opening chat is almost always what the player wanted, so I would not hold the PR on it. Just flagging the behavioral edge.

Nothing else stands out. No sim/wire/i18n surface, no new strings.

jgyy:
Thanks for the review. Pushed a follow-up commit for the nit: `preventDefault` on the `chat` edge now only fires when the previously-focused element was not a button, so a focused button's own Enter activation still fires alongside chat opening, same as before this PR. The composer-newline fix is unaffected for the normal (no control focused) case.

Added a regression test (`does not cancel a focused button own Enter activation when it also opens chat`) alongside the existing chat-keybind tests. `tsc` and `tests/input.test.ts` (65/65) pass locally.

FernandoX7 (approved):

PR #1810, opened 2026-07-12
feat(ui): mirror error toasts into the chat log

jgyy:
## Summary

- Center-screen error toasts (`showError()`, e.g. "you cannot do that", "target out of
  range") faded after 1.6s with no way to see them again once missed.
- `showError()` now also mirrors the same, already-localized text into the chat log's
  `system` channel: the same one loot pickups, level-ups, and death lines already use
  (WoW-style error/system channel logging). No new channel invented.
- The on-screen toast's own timing and behavior are unchanged; this is purely additive.

## Related issues

Playtester feedback: error toasts disappear too fast to read and are never logged.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/error_toast_log.test.ts` (new unit + source-guard tests, all green)
  - `npx vitest run --maxWorkers=2 tests/architecture.test.ts tests/localization_fixes.test.ts tests/localization_coverage.test.ts tests/hud_chrome_i18n.test.ts` (green)
  - `npx tsc --noEmit` (clean)
  - `npx biome check` on all changed files (clean; unrelated pre-existing formatting in
    the same coordinator file was reformatted per the repo's "format only changed files"
    convention)
  - `npm run build:env && npm run build:server && npm run build` (all green)
  - Full `npx vitest run` suite: green in isolation; a batch run under heavy concurrent
    sandbox load produced unrelated timeouts (raid/pvp/social/db tests), all confirmed
    passing when re-run in isolation, matching this repo's documented core-contention
    flake risk.
  - `npm run gate`: every step is green except `sfx check`, which fails identically on a
    clean `release/v0.25.0` checkout (111 pre-existing placeholder audio assets out of
    spec) with zero relation to this change; not touched by this PR.
- Manual steps:
  - Booted the offline game headless via puppeteer-core, triggered a real
    `hud.showError('Target is out of range.')`, and confirmed the same text appears in
    the chat log under the `system` chan immediately, alongside the still-fading toast.

## Screenshots / recordings

Toast and mirrored chat line together:

toast + chat mirror

Toast close-up:

toast crop

Chat pane close-up (mirrored system-channel line):

chat pane crop

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green except the pre-existing,
      unrelated `sfx check` failure described above (reproduced on a clean
      `release/v0.25.0` checkout). I added decisive tests for the new behavior.

### Cross-platform

- N/A: this is a client-only HUD/chat presentation change (`src/ui/hud.ts`,
  `src/ui/error_toast_log.ts`). It does not touch `src/sim/`, `IWorld`, or the wire
  protocol, so there is no offline/online/headless-RL parity surface here.


FernandoX7 (commented):
Clean, well-scoped change. The pure `error_toast_log.ts` helper is a good call: the mirror decision and color are unit-tested directly, and the source-guard test pins the wiring plus the untouched 1600ms fade. No new string (it mirrors the already-localized `localizeErrorText` output), and `log()` routes through `appendLog`, which uses a text node (no injection) and caps the pane at 200 lines. Correctly a client-only presentation change, so the N/A cross-platform note is accurate.

One design gap and a couple of small things.

**should-fix: no consecutive-duplicate suppression, so the system channel floods on repeated failed actions.** The most common errors ("Out of range.", "That ability is not ready yet.", "Not enough rage/energy/mana!", "Line of sight.", "You must be facing your target.") are emitted once per attempt from the sim (`src/sim/combat/casting_lifecycle.ts:273,332,348,361,751`), and each one now appends a chat line at `src/ui/hud.ts:10235`. The toast itself collapses a burst to a single banner via `clearTimeout`, but the mirror does not. Auto-repeat is filtered (`src/game/input.ts:682`), so holding a key won't flood, but mashing a key or click-spamming an action button (normal in combat) produces one duplicate line per press, pushing real history (loot, level-ups, deaths) out of the shared channel. A cheap fix: track the last mirrored text (optionally with the last tick) and skip a mirror that repeats the previous line. That also lines up better with classic behavior, where the on-screen error frame is separate from chat and combat errors are not chat-logged.

**nit:** `showError` is also used for a few non-error notices, e.g. `t('game.talents.exportCopied')` ("Build string copied to clipboard.") at `src/ui/hud.ts:12975`. Those now show up in chat in error-red (`#ff4444`), so a success confirmation reads as a red system-error line.

**nit:** the two reformatted `combat(...)` SFX blocks in `hud.ts` (around lines 8325 and 8408) are unrelated to this change. It's disclosed and allowed by the format-changed-files convention, but dropping them would keep the diff to the three lines that matter.

jgyy:
Thanks for the review! Pushed a fix for the should-fix item:

- **Consecutive-duplicate suppression**: `shouldMirrorErrorToast` now takes the last-mirrored text and skips a mirror that repeats it (`src/ui/error_toast_log.ts`), with `hud.ts` tracking `lastMirroredErrorText` and clearing it whenever a *different* error comes through. A burst of the same error (key-mashing / click-spamming through a persistent out-of-range/cooldown/mana condition) now collapses to a single chat line, same as the on-screen toast already does, while a different error in between still logs normally. Added direct unit coverage plus updated the source-guard test in `tests/error_toast_log.test.ts`.

Left the two nits as-is for now since you called them optional:
- The `exportCopied` red-styling nit would need a broader look at `showError`'s call sites to distinguish error vs. non-error notices, which felt like scope creep for this PR.
- Left the two unrelated reformatted SFX blocks alone per the disclosed format-changed-files convention.

Happy to take those on in a follow-up if you'd rather have them here instead. Verified with `npx vitest run tests/error_toast_log.test.ts`, `npx tsc --noEmit`, and the architecture/localization guard suites, all green.

FernandoX7 (approved):

PR #1812, opened 2026-07-12
feat(ui): make Discord account linking discoverable

jgyy:
## Summary

Playtester feedback: linking a Discord account was reachable only via the undocumented `U`
keybind. There was no menu item, button, or hint anywhere in the client telling a new player
the feature existed.

This adds a small, persistent desktop entry point to the existing HUD chrome instead of
inventing new UI surface area:

- A new `#mm-discord` button in the desktop micro-menu (the same button rail as
  `#mm-social`, `#mm-valecup`, `#mm-leaderboard`, ...), reusing the existing Discord brand
  icon and the existing `hudChrome.discord.title` i18n key. It opens the exact same panel the
  `U` keybind opens (`toggleDiscordPanel`), wired through a new thin `Hud.attachDiscordHook`
  seam, mirroring how the mobile "More" tray's Discord entry already worked.
- The keybind itself was already correctly listed in the in-game keybind/settings list
  (`src/game/keybinds.ts`), so no change was needed there.
- Skipped a tutorial-flow mention: another PR already owns tutorial content, per the task
  brief.

While verifying the new button end-to-end I found and fixed a real wiring bug: the boot-time
`syncDiscordEntries()` call ran once, at module load, before the in-game HUD (which lives in a
`<template>` and is only cloned into the live DOM on world entry) exists. That left
`#mm-discord` permanently `hidden`, invisible even after entering the world. Fixed by
re-running `syncDiscordEntries()` right after `mountGameUi()` clones the HUD template in.

```mermaid
flowchart LR
    subgraph Before
        A1["Player has no idea\nDiscord linking exists"] -->|"must already know"| B1["Press 'U' (undocumented)"]
        B1 --> C1["Discord panel opens"]
    end
    subgraph After
        A2["Player sees the\nmicro-menu rail"] --> B2["#mm-discord button\n(Discord icon + 'u' keycap)"]
        B2 -->|click| C2["toggleDiscordPanel()"]
        D2["Press 'U' keybind\n(still works)"] --> C2
        C2 --> E2["Discord panel\n(link / status / unlink)"]
    end
```

## Type of change

- [x] Bug fix (the `#mm-discord` template-timing visibility bug found while implementing this)

## How was this tested?

- Commands:
  - `npx vitest run tests/client_shell.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts`
  - `npx tsc --noEmit`
  - `npm run i18n:gen` (no drift; `hudChrome.discord.title` already existed)
  - `npm run security:gate` (malware scan PASS)
  - `npm run build:env && npm run build:server && npm run build`
  - Full `npm test` suite (see note below)
- Manual steps:
  - Drove the real client via `npm run dev` + `puppeteer-core` (`scripts/enter_offline_game.mjs`
    flow): entered the offline world, confirmed `#mm-discord` renders visible in the micro-menu
    with the Discord icon, `title="Discord"`, `aria-label="Discord (U)"`, and the `u` keycap,
    then clicked it and confirmed it dispatches through the same `toggleDiscordPanel` path the
    `U` keybind uses.
  - Full local `npm run gate` was run. Two pieces of unrelated, pre-existing repo/environment
    noise showed up and are **not** from this change (verified by reproducing them against a
    clean `up/release/v0.25.0` checkout with none of this branch's changes applied):
    - `tests/server/http/characterization.test.ts` has 1-3 flaky leaderboard timing failures
      under heavy concurrent-load on this shared machine; reproduces identically on a pristine
      checkout.
    - `npm run sfx:check` reports 111 pre-existing audio assets out of spec (bitrate/loudness);
      no audio files are touched by this PR.
  - The pre-push QA floor (`tsc`, guard tests, scoped biome, copy-rule scan) is green.

## Screenshots / recordings

In-game HUD showing the new `#mm-discord` entry in the micro-menu button rail (Discord icon,
`u` keycap, right above the settings/gear button):

HUD with Discord entry

Close-up on the button rail:

Micro-menu rail close-up

(Offline/no-account demo mode: clicking the button falls through to the existing
"open the community invite in a new tab" behavior, since there is no logged-in account to show
a link/status panel for; online, with an account token, it opens the same in-page Discord panel
the `U` keybind opens, per existing `openDiscordEntry`/`toggleDiscordPanel` logic.)

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green for everything this PR touches; the two
      failures observed locally (flaky server characterization test under load, pre-existing sfx
      spec drift) are reproduced identically on a clean `up/release/v0.25.0` checkout with none of
      this branch's changes, so they are not a regression from this change.

### Cross-platform

- [x] **Tested on desktop and mobile.** The new entry is desktop-only (mirrors the existing
      `#mobile-discord` mobile tray entry, which already covers touch). Verified via the puppeteer
      HUD screenshot on a 1600x900 desktop viewport; the mobile "More" tray entry is unchanged.
- [x] **Accessible.** The button carries `title`/`aria-label="Discord (U)"`, a visible keycap, and
      is keyboard-reachable like every other micro-menu button; no motion added.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** No new string was needed:
      the button reuses the existing `hudChrome.discord.title` key already in the English catalog.
      `npm run i18n:gen` shows zero drift.
- [ ] N/A: numbers/money/dates. N/A.
- [ ] N/A: sim/server text. N/A.

### Hygiene

> Conventional Commit: `feat(ui): make Discord account linking discoverable`


FernandoX7 (commented):
Nicely scoped. Reusing the existing `discord` icon and the `hudChrome.discord.title` key means no new surface area or strings, both HTML entries get the button so `/play` stays in parity, and the `client_shell.test.ts` case asserts both. The template-timing bug you found is a real one, and the fix is placed correctly: `mountGameUi` clones the template first, then re-runs `syncDiscordEntries` so `#mm-discord` is revealed the moment it exists in the live DOM.

One should-fix, because it undercuts the discoverability goal in the offline world:

**should-fix** `src/main.ts:1768` (+ `:1773`, `:6089`, `:6167`): the desktop button is a visible dead button offline. `syncDiscordEntries` reveals `#mm-discord` on `DISCORD_BUILD_ENABLED` alone (`main.ts:6093`), which is a build constant that is true by default, so the button shows in the offline browser world. But `attachDiscordHook(() => toggleDiscordPanel())` is wired only inside `if (online)` (`main.ts:1773`), so offline `discordHook` stays null and the click handler (`hud.ts:1765`, `() => this.discordHook?.()`) no-ops. Even if it were wired, `toggleDiscordPanel` bails on `!api.token` (`main.ts:6169`), and offline there is no token. So a desktop player in the offline demo (the exact mode your screenshots show) sees the button and clicking it does nothing.

This is also why the "mirrors the mobile More tray entry" framing does not quite hold: the mobile entry wires `onDiscord: () => openDiscordEntry()` in the `MobileControls` ctor unconditionally (`main.ts:1263`), and `openDiscordEntry` has the logged-out fallback (`main.ts:6099`, opens the community invite when there is no token). That is what makes the mobile button work offline, and it is what the screenshot note claims the desktop button does ("falls through to the existing open the community invite in a new tab behavior"), but the desktop path does not do that.

Cleanest fix is to actually mirror mobile: wire `hud.attachDiscordHook(() => openDiscordEntry())` and move it out of the `if (online)` block, so offline opens the invite and online opens the panel, same as the tray entry.

nit `tests/client_shell.test.ts:996`: the test pins that `hud.attachDiscordHook(() => toggleDiscordPanel())` exists as a string, but a static-string check cannot see that it sits behind `if (online)`, which is the arm the offline dead-button turns on. Once the wiring is fixed, worth a line asserting the hook is attached unconditionally so this does not regress silently.

The second commit's two sfx line wraps are unrelated to the feature but fine, since biome gates the whole touched file.

jgyy:
Thanks for catching this, good find. Fixed in 1db1a0fe2:

- Moved `hud.attachDiscordHook(...)` out of the `if (online)` block so it runs unconditionally.
- Changed the hook to `() => openDiscordEntry()` instead of `() => toggleDiscordPanel()`, matching the mobile tray's `onDiscord: () => openDiscordEntry()`. Now offline (no token) it opens the community invite, online it opens the panel, same as the mobile entry.
- Updated `tests/client_shell.test.ts:995` to assert `hud.attachDiscordHook(() => openDiscordEntry());` is immediately followed by `if (online) {`, so it can no longer regress back inside that gate silently.

`npx vitest run tests/client_shell.test.ts` and `npx tsc --noEmit` both pass on the changed files.

FernandoX7 (approved):

PR #1759, opened 2026-07-10
docs(pvp): Frontier zone as the Phase 2 spec (honor foundation shipped in #1817)

ryan-foo:
## Summary

The Frontier PvP program, updated to reflect what has shipped. Phase 1 (the honor foundation) landed in #1817 on release/v0.25.0; this PR is now the Phase 2 spec: the factionless open-world PvP zone.

What this PR contains:

- `docs/prd/pvp-honor-and-quartermaster.md`: the Phase 1 foundation spec, marked SHIPPED (#1817) and kept as the reference doc, with an as-shipped note covering the two divergences (the PvP Offense/Defense pair shipped as one player-facing Warfare stat; the quartermaster shipped as FURY in Eastbrook with 40 item-level 28 epics).
- `docs/prd/frontier-pvp.md` (renamed from `frontier-pvp-honor.md`): the ACTIVE next-phase spec. A factionless, RuneScape-Wilderness-style zone: no teams, everyone hostile in the band, a depth danger gradient, single/multi-combat, carried cargo extraction with full loot of the risked layer, hourly events, and 2x honor on open-world kills through the shipped `grantHonor`. The $WOC stakes layer stays in section 12 behind the token firewall.
- `docs/prd/FRONTIER_PHASE1_HANDOFF.md`: the slice-by-slice implementation handoff for the zone skeleton (F1), with its prerequisite marked satisfied by #1817.
- `docs/CLAUDE.md`: the PRD index updated to match.

## Type of change

- [ ] Feature: new functionality
- [ ] Bug fix
- [ ] Refactor or performance (no behavior change)
- [x] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

Docs only; no code changes. Merged release/v0.25.0 into the branch and verified the PRD set matches the shipped #1817 surface (Warfare stat, FURY vendor, `grantHonor`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1817, opened 2026-07-12
feat(pvp): ship #1759 Phase 1 honor and Warfare

ryan-foo:
## Summary

Ships the standalone Phase 1 PvP foundation described in #1759:

- Adds persistent Honor with ranked Arena victory rewards, Fiesta takedown and result rewards, repeat-opponent diminishing returns, and daily Arena tapering.
- Adds FURY, the Honor Quartermaster in Eastbrook, with 40 Honor-priced item-level 28 epic PvP items covering every supported equipment slot, including neck and two rings.
- Adds one player-facing Warfare rating. Warfare increases damage dealt to players and reduces damage taken from players, with independent internal caps and no PvE effect.
- Wires Honor and Warfare through offline play, the authoritative server, snapshots, persistence, floating combat text, vendor affordability, tooltips, the character sheet, and localization.
- Keeps soulbound PvP gear non-transferable while allowing players to destroy unwanted purchases.

## Related issues

Implements Phase 1 from #1759.

## Type of change

- [ ] Bug fix
- [x] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/honor.test.ts tests/pvp_honor_gear.test.ts tests/vendor_view.test.ts tests/char_window.test.ts tests/stat_tooltip.test.ts tests/stat_tooltip_view.test.ts tests/item_compare.test.ts tests/items.test.ts tests/arena_online.test.ts tests/persistence_round_trip.test.ts tests/snapshots.test.ts tests/world_api_parity.test.ts tests/fct_core.test.ts tests/fct_event.test.ts tests/fct_painter.test.ts tests/heroic_soulbound.test.ts` (647 passed)
  - `npm test -- --maxWorkers=4` (11,750 passed, 14 skipped; four unrelated load-sensitive tests timed out and the new localization check found five untranslated labels)
  - `npx vitest run tests/i18n_completeness.test.ts tests/fixes.test.ts tests/loot_ffa_sim.test.ts tests/mail.test.ts --maxWorkers=1` after translating the labels (115 passed)
  - `npx tsc --noEmit`
  - `npm run build:env`
  - `npm run build:server`
  - `npm run build`
  - `npm run security:gate` (2,673 files, 134 flags, 0 high)
  - `npm run ci:changed`
  - `node scripts/i18n_resolved_hash.mjs --check`
  - `npm run gate` reached the target branch SFX check, then stopped because 111 existing audio files are out of the newly enforced bitrate and loudness spec. This PR changes no audio files.
- Manual steps:
  - Entered an offline level 20 Warrior, opened FURY, and confirmed 40 offers and the 5,000 Honor balance.
  - Confirmed the helm tooltip shows one `+17 Warfare` line and no Offense or Defense variants.
  - Equipped the helm and confirmed the character sheet shows `Warfare: +1.7% dealt / -1.7% taken` with the matching explanation.
  - Confirmed buying an item reduces current Honor without changing lifetime Honor.

## Screenshots / recordings

Before, separate player-facing Warfare variants:

Before: separate Warfare labels

After, FURY and one Warfare stat:

After: FURY vendor and Warfare tooltip

After: Warfare character sheet

---

### Quality

- [ ] **The full gate passes.** Blocked by 111 pre-existing target-branch SFX conformance failures described above. All remaining gate stages were run manually.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Desktop was verified. This change does not introduce a new window or mobile-only layout.
- [x] **Accessible.** Honor prices reuse the vendor buttons and localized ARIA price labels; unaffordable rows use the native disabled state.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** English catalog entries and the five required non-Latin fills are included.
- [x] Numbers, money, dates, units, and percents go through the i18n formatters.
- [x] Player-facing sim text is re-localized at the client boundary.

### Hygiene

- [x] No secrets, credentials, or `.env` files are committed, and `ALLOW_DEV_COMMANDS` is not enabled in production.
- [x] Generated localization files were regenerated through the repository scripts.


ryan-foo:
Arena economy follow-up for Phase 1 of #1759 landed in `6293f0d94`:

- reduced ranked 1v1 wins from 100 to 25 Honor
- reduced ranked 2v2 wins from 150 to 50 Honor per winner
- only the first Arena win against the same opponent or team now pays Honor each UTC day
- retained the existing total-win taper after 10 and 15 daily wins

At FURY's current 4,375-Honor full-kit price, the clean-opponent baseline is now 175 1v1 wins or 88 2v2 wins before tapering. Farming one opponent can produce only one paid Arena win per day.

Validation after the balance change: 109 adjacent Honor/Arena/Fiesta/PvP tests passed, the 4-test online Arena persistence suite passed, `tsc --noEmit` passed, and the changed PvP source files pass Biome.

ryan-foo:
@Rubsey plz review for merge into v/0.25.0

Rubsey (commented):
Really solid Phase 1. The load-bearing plumbing is careful throughout: the new stats fields land in both `baseEntity` and `blankEntity`, `applyWire` guards the rolling-deploy case so an older six-field server cannot NaN the character sheet, `honor`/`lifetimeHonor` ride the snapshot self as delta-guarded scalars and are added to `IWorldDuelArena` (implemented in both worlds, with the 205 to 207 parity pins bumped in the same change), and FURY spawns at a reserved entity id after the rng-driven roster so it cannot perturb entity ids or the replay stream. Persistence is equally thorough: honor state normalizes on load, backfills `lifetimeHonor >= honor`, defaults cleanly for pre-honor rows, and `honorArenaDaily` persists so the anti-win-trade window survives a relog. The Warfare hook is tightly scoped to hostile player-vs-player, post-mitigation, deterministic, and leaves pets/self/PvE byte-identical. Nice work.

One blocking issue, then a few small notes.

## Blocking

- `tests/parity/golden/arena_1v1.json:123` and `tests/parity/golden/arena_2v2_wipe.json:878` are stale, so the parity gate is red on this branch. `balance(pvp): slow Arena honor farming` (6293f0d9) lowered `RANKED_ARENA_WIN_HONOR` from `{1v1: 100, 2v2: 150}` to `{25, 50}` and re-minted the unit tests, but not these two goldens, which still encode `honor: 100 / 150` with `totalWins: 1`. `npx vitest run tests/parity/parity.test.ts` fails both (expected 100/150, received 25/50). The rng digest is byte-identical in both traces, so this is purely a stale golden, not a determinism regression. Re-mint with `UPDATE_PARITY=1 npx vitest run tests/parity/parity.test.ts` and commit. (See the merge note: doing this after merging the base folds it into one step.)

## Nits

- `src/sim/sim_context.ts:193` + `src/sim/sim.ts:2845`: the `grantHonor` SimContext callback is declared and wired but never called. Every award path imports `grantHonor` from `../pvp` and calls it directly, so the seam callback reads as dead scaffolding. Either route the single grant path through the seam or drop the callback.
- `tests/honor.test.ts`: the 2v2 ranked award is asserted only against `RANKED_ARENA_WIN_HONOR['2v2']` itself (self-comparison), so a wrong 2v2 constant would not redden a unit test. Minor, and the re-minted parity golden will pin the literal end to end.

## Notes (not blocking)

- The `soulbound` semantics changed globally, not just for PvP gear: `discardItem` now gates on `noDiscard` alone (was `noDiscard || soulbound`), so soulbound items become destroyable unless they set `noDiscard`. This is handled correctly (the one pre-existing soulbound item, `heroic_mark`, gained `noDiscard: true` and is pinned in `heroic_soulbound.test.ts`; `items.test.ts:264` pins that a soulbound FURY item can be destroyed but stays non-transferable). Flagging it only because it touches the shared item model, not just this feature.
- Warfare defense does not reduce pet damage (the hook requires `source.kind === 'player'`, per the comment). Reads intentional for Phase 1; just confirming that matches `warfare.md`.
- The merge conflicts are all mechanical. The i18n trio (`i18n.resolved.sha256`, `i18n.status.summary.json`, the generated set) is regen churn. The five parity goldens (two arena, three fiesta) and the two wire tests (`snapshots.test.ts`, `world_api_parity.test.ts`) conflict because the base's Book of Deeds work touched the same arena/fiesta event streams and IWorld member pins. Resolve by merging the base, then re-running the i18n build and the `UPDATE_PARITY=1` re-mint (which also clears the blocking stale golden above), and keeping both sides' pins in the two wire tests (the deeds members plus `honor` / `lifetimeHonor`, count 207+).

PR #1818, opened 2026-07-12
feat(sfx): add production sound studio

maxpolaczuk:
## Summary

- Adds a loopback-only SFX Studio for source replacement, trim and slicing, fades, loop seams, animation timing, EQ, dynamics, exact FFmpeg previews, contextual auditions, version restore, and transactional publishing.
- Keeps cross-clip mix decisions reversible. Category baseline gain, per-key gain, and per-key playback speed live in checked-in maps and are applied through Web Audio `GainNode` and `playbackRate`; they never enter rendered audio bytes.
- Supports ordered round-robin takes: one-shots cycle through variants, while loops select and pin one take until stopped.
- Preserves the intentional conform split: clips under 1 second target -6 dBFS true peak; clips at least 1 second target -14 LUFS. Output is 44.1 kHz / 192 kbps, with a 112 kbps floor for lossy inputs.
- Adds deterministic Export All bundles containing exact published tracks, ordered variants, runtime mix, content-addressed blobs, checksums, and POSIX and Node installers. Installation verifies blobs first and atomically activates `runtime-pack.json` last.
- Adds strict runtime-pack validation, catalog compatibility checks, same-origin blob loading, fail-closed blob integrity checks, persistent Docker overlay support, and coherent fallback to the bundled manifest.
- Adds the complete [SFX Studio tutorial](docs/sfx-studio-tutorial.md), production deployment instructions, focused unit/security coverage, and a real-browser Studio smoke flow.

## Related issues

Release tracking: #1777

Part of #1729

## Type of change

- [ ] Bug fix
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npm run gate` — all 10 steps green on the rebased release commit; 966 test files and 11,864 tests passed, 14 intentional skips
  - `npm run sfx:check` — all 137 catalog clips conform
  - `npm run sfx:manifest`
  - `npm run sfx:studio:smoke`
  - `npm run build:native`
  - `POSTGRES_PASSWORD=validation-only docker compose config --quiet`
  - `docker build -t world-of-claudecraft:sfx-studio-pr-check .`
- Manual/browser steps:
  - Exercised all 137 catalog cues and tracks in Chromium.
  - Verified exact one-shot and loop auditions, runtime gain and playback rate, Export All, and zero browser errors.
  - Verified artifact installation and the installed runtime manifest and content-addressed blobs.

## Screenshots / recordings

Initial published source and animation context:

SFX Studio initial desktop

Edited exact preview with runtime playback mix:

SFX Studio edited desktop

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green with decisive behavior and security coverage.

### Cross-platform

- [x] Desktop Studio flow and native build verified. Phone-sized Studio layout is N/A because this is a loopback-only desktop authoring tool, not player UI.
- [x] Uses semantic labels and native controls, keyboard-operable actions, visible focus styling, ARIA status updates, and no motion-only interaction.

### Localization (i18n)

### Hygiene

- [x] Generated SFX manifests were regenerated through the checked-in build pipeline.


EnriqueGF:
Thanks for this contribution :) It is a genuinely impressive piece of tooling, and the security posture of the Studio server is careful work.

**Base-branch gate: PASS.** This targets `release/v0.25.0` (the latest release branch) and the head already contains its tip.

I cannot approve it yet, though: the test suite is red on a clean checkout, so the "gate is green" claim did not reproduce here.

## Blocking

1. **`tests/sfx_studio_server_security.test.ts` fails 5 of 13** on a clean, idle 12-core machine with system ffmpeg installed, both in a full run and in isolation with a fresh studio root:
   - `tests/sfx_studio_server_security.test.ts:186` ("exports a token-protected production ZIP only from the applied playback state") exceeds its own 30s budget, measured at 33.0s and 34.1s across runs. It ffprobes the whole 140-file catalog.
   - Its failure leaves `STUDIO_ROOT/exports` in a bad state, because the exports reset lives in that one test's `finally` block. The shared-root cascade then fails four more tests, and those four are exactly the security pins: export-symlink escape (`:240`), stale-draft split save (`:272`), source-symlink escape (`:329`), symlinked version archives (`:350`). On any machine that misses the export budget, those pins prove nothing.

   Please raise that test's timeout well above the catalog-analysis cost, and make the exports reset a `beforeEach`/`afterEach` invariant so each security pin is exercised independently of the export test's outcome.

2. **`tests/sfx_export_bundle.test.ts`** ("is byte deterministic and includes the exact published runtime state") burns 27s of its 30s budget here. Same latent class as (1); it will flake on a loaded CI runner.

3. **New undocumented prerequisite.** `scripts/sfx_studio/audio_io.mjs` spawns bare `ffmpeg`/`ffprobe` from `PATH`, so `npm test` and `npm run gate` now require a system ffmpeg. CI got the `apt-get` step, but `scripts/gate.mjs` has no preflight check and neither `README.md` nor `CLAUDE.md` mentions it. A contributor without ffmpeg gets a loud red suite with no explanation. Please add a preflight check with a clear message, and document the requirement.

## Should fix

4. `server/main.ts:747` (`staticSfxHash`): a synchronous `readFileSync` plus SHA-256 of up to 4 MiB runs inside the request path, and `serveStatic` has no try/catch around it. If a file is replaced or removed between the `statSync` and the read (exactly what an overlay pack install does), the throw escapes to the global `uncaughtException` handler at `server/main.ts:2517` and the request is never answered, leaving the socket hanging. Wrap the hash in a try/catch and 404 or 500 on failure. The per-file (size, mtimeMs) cache keeps the steady-state cost bounded, so this is about the race, not throughput.

5. `scripts/sfx_studio/audio_io.mjs:1128`: the `MASTERING_REVISION` / mode guard has no negative case. `tests/sfx_studio.test.ts:918` only feeds the current revision, so deleting the check leaves every test green while stale cached masters get silently reused. Please add per-dimension negatives (revision minus 1, and a wrong mode).

## Nits

6. `tests/sfx_studio.test.ts:1158` is a constant self-comparison: the fixture is built as `SFX_MAX_TRACKS_PER_KEY + 1` and the expected message interpolates the same constant, so both sides move together on an edit. Pin the cap to a literal.
7. `tests/sfx_export_bundle.test.ts:248` uses a bare `.toThrow()` with no message, so it would pass on an unrelated error.

## Verified clean

- `npx tsc --noEmit` passes. `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` pass. `biome ci --changed` reports 0 errors (46 warnings, which CI does not gate).
- `npm run sfx:manifest` regenerates `src/game/sfx_manifest.generated.ts` with no diff, and the fixed catalog keeps all 14 `lockpick_*` keys plus `cast_lightning_bolt`, so the regression from the earlier SFX PRs is not back.
- Module-first is respected: cue resolution moved out of `hud.ts` into `src/ui/combat_sfx.ts` and `src/render/world_audio.ts`, and the monoliths shrink rather than grow. The `smithy` flag on `ZonePropsDef` is declarative data and replaces a fragile index-based check in `props.ts`, which is a real improvement.
- `src/sim/` stays pure and deterministic: the only new `Math.random` calls are audio jitter in `src/game/sfx.ts`.
- Studio server security holds up: 127.0.0.1-only bind, a Host check that defeats DNS rebinding, per-launch token plus exact-Origin on mutations, no CORS headers, `spawn` with an argv array (no shell) for every ffmpeg call, allowlisted keys plus realpath containment on every file path, and validated ZIP entry names. The production overlay is regex-gated and mounted read-only, blob integrity is fail-closed, `runtime-pack.json` is served `no-store`, and the client re-validates schema, byte budgets, URL shape, and `bundleId` before falling back to the compiled manifest.
- No em or en dashes and no emojis in the added lines.

Fix the red suite (1 to 3) and the two should-fixes, and I will be happy to take another look.

FernandoX7 (commented):
This is a strong contribution. The security posture is well above what a local authoring tool needs: loopback-only bind with a Host-header pin, per-launch token plus Origin check on mutations, realpath containment on every file path, ffmpeg via arg arrays with no shell, a fail-closed content-addressed overlay on the game server, and a runtime pack loader that rejects anything not root-relative under /audio/sfx with a sha256 pin. I also verified all 137 binaries are genuine MPEG audio with no trailing data, the installers are pure verify-copy-activate, the lockfile is untouched, and nothing in the new code references an external host. The hud.ts change is a real extraction (cue mapping moved out to combat_sfx.ts, the coordinator shrank), the online.ts change is a clean move of the origin helpers to client_origin.ts with re-exports, and conforming the catalog fixes the sfx:check red that release/v0.25.0 currently carries. Parity checks out too: the smithy stall flag is static content read identically by both worlds, and the aura cues rely on aura name/kind which already round-trip the wire for all entities.

Findings, none blocking:

- should-fix: `npm test` (and therefore `npm run gate`) now hard-requires a system ffmpeg on PATH. `scripts/sfx_studio/audio_io.mjs:1201` (and ~9 other sites) spawn `ffmpeg` from PATH, `tests/sfx_studio.test.ts` exercises those paths with no skip guard, and the CI fix commit confirms the suite fails without it. A contributor without ffmpeg installed gets unrelated red tests locally. Either resolve the binary the way the conform pipeline does (ffmpeg-static is already a dependency and `conform_audio.mjs` takes an ffmpegPath parameter) or document the system prerequisite in README's Development section next to the npm test/gate commands.
- should-fix: the installer's fail-closed branch is never executed. `tests/sfx_export_bundle.test.ts:187` runs `install.sh` only on a pristine artifact, and `install.mjs` is presence-checked at line 173 but never run. The checksum-failure handling is the safety claim of the bundle format; corrupt one blob byte in the extracted artifact, run the installer, and assert non-zero exit with the static root unmutated (and give `install.mjs` one real execution).
- should-fix: `restoreVersion` has only rejection-path coverage (`tests/sfx_studio.test.ts:202`, `tests/sfx_studio_server_security.test.ts:357`, both `.rejects.toThrow`). The round-trip that versioning exists for (publish v1, publish v2, restore v1, assert audio bytes and published-state hash revert) is unpinned.
- nit: the LUFS conform target is asserted only relative to the imported constant (`tests/sfx_conform.test.ts:208` uses `TARGET_LUFS + tolerance`), so a target change fails no test. The peak side pins literals; add an `expect(TARGET_LUFS).toBe(-14)` or one real long-clip measurement.
- nit: `tests/sfx_studio_server_security.test.ts` covers the origin dimension on POST (line 173) but not token-on-POST independently (valid origin, missing token), and the upload `x-filename` traversal neutralization (`audio_io.mjs:974`), the `/repo/` and `/preview/` guards, and the JSON/upload byte limits are defended in code but unexercised. Cheap adds that keep the hardening honest.

Notes, no action needed:

- Two deliberate behavior changes are worth a nod since neither is strictly required by the studio goal: swings now audibly whiff on miss/dodge/parry, and aura cues now classify buff vs debuff on gain and stay silent on removal (the old code played buff_apply even when you gained a debuff, and debuff_apply on any loss). Both are improvements in my view and both are pinned in tests/combat_sfx.test.ts.
- The containment checks in the studio tool hardcode `/` as separator (`audio_io.mjs:90` etc.). Fine for a loopback dev tool on macOS/Linux; the production overlay in `server/sfx_overlay.ts` uses `path.sep` correctly.

maxpolaczuk:
@FernandoX7 @EnriqueGF Thanks for the detailed review. All findings are addressed in 04c2c4a24, with the branch rebased onto the current release/v0.25.0 tip.

- Raised the catalog/export budgets and made exports cleanup a beforeEach/afterEach invariant.
- Added an FFmpeg and ffprobe gate preflight plus the README prerequisite.
- Execute both POSIX and Node installers in tests. Both now validate all artifact inputs and existing targets before mutation, reject corrupt/tampered/truncated bundles with the static root unchanged, install every runtime variant, and activate the manifest last.
- Added the full publish v1, publish v2, restore v1 roundtrip in a disposable repo. Exact audio, complete mastering/output metadata, and published identity return to v1, including historical mastering revisions.
- Bound static SFX verification and delivery to the same bounded snapshot bytes, with behavioral replacement/read-error/HEAD coverage.
- Added the mastering revision/mode negatives, literal -14 LUFS and 8-track pins, precise ZIP errors, missing-token POST, upload filename traversal, byte-limit, preview/repo symlink, and version archive coverage.

Local verification on the rebased head: npm run gate passed all 10 stages, including 11,928 tests, 137 audio conformance checks, typecheck, and all production builds. The fresh GitHub checks are running now.

FernandoX7 (approved):

PR #1819, opened 2026-07-12
feat(cosmetics): hover cosmetics, animated back attachments that float the wearer

maxpolaczuk:
## What

Three equippable back attachments that make the wearer hover instead of walk, purely cosmetically:

| Cosmetic | Model | Motion | Ambient VFX |
|---|---|---|---|
| Butterfly Drift | hover_butterfly_wings | fast flap (wing.l / wing.r) | iridescent sparkle drift + twinkles |
| Dawnfeather Wings | hover_angel_wings | slow flap | dawn-gold feather drift |
| Tinker's Jetpack | hover_jetpack | rigid | twin thruster flame + smoke, point light |
| Dreadwyrm Wings (legendary) | hover_dragon_wings | slow heavy up-down beat | rising embers + coal smoke + ember-vein glints |

While one is applied the character floats slightly above the ground with a gentle bob, keeps the normal idle stance when standing, and leans into an airborne glide pose while moving. Wings flap procedurally (slower at rest, faster on the move). Nearby players see all of it via the identity wire.

## Gameplay neutrality

The hover is a render-only pose lift (poseWrap), the same mechanism as the swim lift: sim position, movement speed, collision, jumping, swimming, and every combat number are untouched, and the click-target capsule stays at the true sim position. Verified against the graphics-fairness rule; the RL env and obs are unaffected.

## Architecture

- Data-as-code catalog src/sim/content/hover_cosmetics.ts; Entity.hoverCosmeticId
- change_hover command + terse hov identity wire field; IWorldCosmetics.changeHoverCosmetic on both Sim (offline free-apply, the change_skin precedent) and ClientWorld (validated optimistic apply)
- Server: catalog-validated dispatch, accounts.cosmetics.hoverId persisted via one atomic jsonb_set (JSONB null on clear), join seeding, live multi-character sweep, HEAVY_SELF_CMDS re-push
- Render: chest-bone attach, procedural wing flap, hover rise/bob/lean, glide clip override, ambient VFX through the existing weapon-VFX rig
- Assets: Tripo prop lane models packaged by scripts/build_hover_attachments.mjs, which bakes each wing model into a canonical frame (spread on X, centered origin hinge, normals rotated with it), splits the triangles into wing.l / wing.r halves on the true spread axis, and fails loudly on a lopsided split. Flap is per-model: axis 'y' folds open/closed (butterfly), axis 'z' beats the tips up and down (dawnfeather, dreadwyrm)

There is deliberately no store/SKU or equip UI in this PR; the command is the seam and a store lands separately. When monetization arrives, changeAccountHoverCosmetic needs an ownership gate like the weapon-skin one.

## Try it

Offline mode (no server needed): enter a world, then in the console:

    window.__game.world.changeHoverCosmetic('butterfly_drift')
    window.__game.world.changeHoverCosmetic('dawnfeather_wings')
    window.__game.world.changeHoverCosmetic('tinkers_jetpack')
    window.__game.world.changeHoverCosmetic('dreadwyrm_wings')
    window.__game.world.changeHoverCosmetic(null)  // land

Online it is the same call on the connected world (server validates + persists across sessions).

## QA

- Full suite green: 11,889 passed (966 files), tsc clean, npm run build green (media manifest includes the three GLBs)
- End-of-contribution QA gate ran with five parallel reviewers (checklist, sim architecture, wire parity, persistence safety, security); all findings fixed in the second commit: Object.hasOwn catalog gate (prototype-key strings no longer pass validation), accountCosmetics.hoverId mirroring on both hosts, and the three coverage suites the weapon-skin precedent had (server dispatch, atomic DB writer, hov wire round trip)
- New tests: tests/hover_cosmetics.test.ts (catalog + GLB flap-node integrity + sim apply) plus hover blocks in game_sessions, character_db, and snapshots

The branch also mirrors main's e7fc2a05e (untracks the gitignored i18n status registry) so the pre-push copy scan gates real copy, not the generated registry.

Based on feature/armory-skins-store so the diff here is exactly the hover feature.


maxpolaczuk:
Round 2: the first-batch wing GLBs left Tripo with the wing plane spread along Z, so the yaw-only back mount held them perpendicular to the character, and the splitter (cutting on the thin axis) produced a full pair plus a ghost shell, the doubled-wings look. The build script now bakes every wing into a canonical frame (spread on X, centered hinge) before splitting on the true spread axis, with a loud failure on a lopsided split so it cannot regress silently. Butterfly and dawnfeather rebuilt and verified from a six-angle orbit probe.

Also new: Dreadwyrm Wings, a legendary pair of bat-membrane dragon wings (exactly two) with a slow heavy up-down wingbeat and a dragonfire VFX rig (rising embers, coal smoke, ember-vein glints). Catalog is now four cosmetics; suite at 11,889 green.

PR #1820, opened 2026-07-12
feat(ui): show where-to-next tips on tutorial completion

jgyy:
## Summary

The 5-step New-Adventurer tutorial (`src/ui/tutorial.ts`) walks a brand-new
character through move -> find the starter NPC -> accept quest -> slay wolves
-> turn in, then just says "quest complete" and auto-dismisses after 9s. A
first-time player has no pointer to what comes next, which is a known
retention drop-off point right after onboarding.

This adds a short "Where to next" tip list under the closing card, pointing
at the Quest Log, World Map, and Social windows using the player's actual
bound keys (same pattern already used for the interact/quest-log keys
elsewhere in the tutorial). The auto-dismiss linger is bumped from 9s to 14s
so there's time to read it before the card fades.

Presentation-only: no `IWorld`, sim, server, or wire/parity changes.

```mermaid
flowchart LR
  A[move] --> B[seek] --> C[talk] --> D[slay] --> E[return] --> F[done]
  F -->|NEW| G["Where to next\nQuest Log (L) / World Map (M) / Social (O)"]
```

## Related issues

N/A

## Type of change

## How was this tested?

- Commands: `npx vitest run tests/tutorial_copy.test.ts tests/tutorial.test.ts`,
  `npx vitest run tests/i18n_completeness.test.ts tests/localization_fixes.test.ts`,
  `npm run i18n:gen`, `npx tsc --noEmit`, `npm run ci:changed`,
  `npm run security:gate`, `npm run build:env`, `npm run build:server`, `npm run build`,
  and the full `npm test` suite (949/958 files green; the remaining 8
  `tests/server/http/characterization.test.ts` failures are pre-existing
  leaderboard-endpoint timeouts under parallel worker contention, confirmed
  by re-running that file alone -> 43/43 pass, and reproduced identically on
  the unmodified base branch).
  `npm run sfx:check` also fails on this branch, but reproduces identically
  on the unmodified base (111 pre-existing audio files out of the target
  loudness/bitrate spec) and is untouched by this change.
- Manual steps: drove the tutorial end-to-end via `scripts/tutorial_shot.mjs`
  (and the new `scripts/tutorial_maxgfx_shot.mjs`, forcing graphicsPreset 5 /
  every advanced-only knob on) against a local `npm run dev`, offline flow,
  through all six steps.

## Screenshots / recordings

**Closing tutorial card, before vs after (default graphics):**

Before:

before

After (new "Where to Next" tips):

after

**Full six-step tutorial flow at max graphics (preset 5 / Advanced, all
advanced-only knobs on):**

1. Move

step 1 move

2. Seek

step 2 seek

3. Talk

step 3 talk

4. Slay

step 4 slay

5. Return

step 5 return

6. Done (with the new tips)

step 6 done

---

### Quality

- [x] **The full gate passes.** All `npm run gate` steps pass except the two
      pre-existing, unrelated failures called out above (leaderboard test
      flakiness under contention, and the repo-wide sfx loudness drift),
      both reproduced identically on the unmodified base branch.

### Cross-platform

- [x] **Tested on desktop and mobile.** The new tip list reuses the existing
      `.tut-card` container and its responsive rules; verified render at
      desktop viewport via the screenshot script. No new touch targets are
      added (the tips are plain text, not interactive).
- [x] **Accessible.** The card already carries `role="status"`/`aria-live="polite"`;
      the new list is announced the same way as the existing body text, no
      focus/keyboard changes.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Added to
      `src/ui/i18n.catalog/hud_chrome.ts` (English-only domain). The four new
      keys are wordy (M16), so I filled the five required non-Latin locales
      (`zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`) in this same change; ran
      `npm run i18n:gen` and committed the regenerated artifacts.

### Hygiene

- [x] I didn't hand-edit generated files; `i18n.resolved.generated/` was
      regenerated via `npm run i18n:gen`.


FernandoX7 (commented):
Clean, well-scoped change. The tip data lives in `tutorial_copy.ts` (`TUTORIAL_NEXT_TIPS`) and `tutorial.ts` just consumes it, the key labels reuse the exact `keybinds.primaryLabel(id) || t('hud.options.unbound')` pattern already in `renderPanel`, and the rebuild is gated by `tutorialNeedsRerender` so it only fires on the step change into `done` (or a touch-mode flip), not per frame. English strings are `t()` keys in the `hud_chrome` catalog with a `{key}` splice, the CSS reuses the hex values already on the sibling `.tut-*` rules, and the after screenshot shows L / M / O resolving correctly. Presentation only, no sim/wire/server surface. Verified the three keybind ids exist (`keybinds.ts:156-172`), tests pass locally (35/35).

Two nits, nothing blocking:

- `tests/tutorial_copy.test.ts:98` (nit): the "points each tip at a real Interface keybind id" test pins the array to the literals `['questlog','map','social']`, but its own comment claims it guards against "a typo would silently render Unbound." A rename in `keybinds.ts` (or a typo mirrored into both source and test) would slip through, since nothing here cross-checks the registry. Asserting each id is in `BIND_ACTIONS.map(a => a.id)` (both exported from `src/game/keybinds.ts`) would actually catch that failure mode. The companion body-key test does the equivalent cross-check against `hudChromeStrings`, so this one is the odd one out.

- `src/ui/tutorial.ts:191` (nit): the "Where to next" title is appended as an `<li class="tut-next-tips-title">` inside the `<ul>`, so assistive tech announces it as a list item rather than a heading or label. The card's `aria-live="polite"` still reads it out, so this is minor, but a `<li>` with a non-list role, or moving the title out of the `<ul>`, would read more accurately.

jgyy:
Thanks for the review! Both nits addressed in ad600bbb6:

- `tests/tutorial_copy.test.ts`: the keybind-id test now cross-checks `TUTORIAL_NEXT_TIPS` against the live `BIND_ACTIONS` registry (`src/game/keybinds.ts`) in addition to the literal-array pin, so a rename or typo in the registry now fails the test instead of silently rendering "Unbound".
- `src/ui/tutorial.ts`: moved the "Where to next" title out of the `<ul>" into its own sibling `<div class="tut-next-tips-title">", so assistive tech no longer announces it as a list item. CSS border/spacing moved from `.tut-next-tips` to the title element to keep the same visual result.

`npx vitest run tests/tutorial_copy.test.ts tests/tutorial.test.ts` -> 41/41 green.

PR #1821, opened 2026-07-12
fix(net): resume-check WS on mobile foreground; fix compact page-toggle digit clip

jgyy:
## Summary

Two related mobile fixes:

1. **Frequent disconnections on Android/iOS.** Both platforms suspend a backgrounded
   tab's JS timers and kill the underlying WebSocket without ever delivering a
   `close` event to the frozen page, and any pending reconnect `setTimeout` is
   itself throttled to roughly once a minute while backgrounded. The client's
   reconnect was purely event-driven (`onclose` -> backoff -> retry), so a player
   could foreground the app onto a zombie "connected" socket with no reconnect in
   flight, or land several backoff steps behind: reported as frequent disconnects,
   even though most drops were really just very late (or missing) reconnects.
   `ClientWorld` now listens for `visibilitychange` and, on resume, force-checks
   the real socket state: it retries immediately if a backoff timer was still
   pending, or drives the same close path a real `onclose` would have if the
   socket died silently while suspended.

2. **Mobile action-ring page-toggle digit clipped on short landscape phones.**
   At the compact-tier breakpoint (`<=480px` height, e.g. a 1044x480 landscape
   viewport), the page-toggle badge is clamped to the 40px touch floor
   (`max(40px, 46px * chrome-scale)`), whose border-box inner height (36px) is a
   hair short of the base 20px digit plus its 15px icon (37px). The ring button's
   `overflow: hidden` (added to clip the cooldown sweep) was cropping the top of
   the digit. Compact tier now shrinks the digit to 15px, comfortably inside the
   36px floor.

## Related issues

Reported via a screen recording showing both the clipped digit and repeated
mobile disconnects; no tracked issue number.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/net_online_visibility_reconnect.test.ts` (new, 4 cases:
    listener lifecycle, zombie-socket resume, pending-backoff resume, no-op while
    genuinely open)
  - `npx vitest run tests/client_shell.test.ts` (new compact-tier CSS pin, plus
    the existing whole-block ring pins still pass)
  - `npx vitest run` (full suite): 949 files / 11596 tests passed; the only
    failures are 9 pre-existing DB-dependent server characterization tests that
    fail identically on a clean `release/v0.25.0` checkout with none of this PR's
    changes (verified via `git stash`), unrelated to this change.
  - `npx tsc --noEmit`: clean.
  - `npm run build:env`, `npm run build:server`, `npm run build` (client): all
    green.
  - `npm run gate`: green through typecheck and all three builds; the `sfx
    check` step fails on 111 pre-existing audio assets that are also out of spec
    on a clean `release/v0.25.0` checkout with no changes applied (confirmed via
    `git stash`), so it's a pre-existing local-environment issue this PR does not
    touch or regress.
- Manual steps: drove the real `#mobile-action-ring` markup and `hud.mobile.css`
  through the dev server at the exact 1044x480 compact-tier viewport from the
  original report (plus standard/tablet tiers, left-handed, and a portrait
  phone), and rendered the real `showReconnectOverlay()` output, to confirm both
  fixes visually (screenshots below).

## Screenshots / recordings

All captured against the real `#mobile-action-ring` markup + `hud.mobile.css`
through the dev server (icons hydrated via the real `hydrateIcons()`), not mocked.

**1. Compact tier (1044x480), full HUD, after fix** — digit fits cleanly:

compact after

**2. Compact tier (1044x480), full HUD, before fix** — same viewport, pre-fix CSS:

compact before

**3. Page-toggle badge, before fix (5x crop)** — top of the "1" is cropped by the
ring's circular `overflow: hidden`:

badge before 5x

**4. Page-toggle badge, after fix (5x crop)** — clear margin above the digit:

badge after 5x

**5. Page 2 of the action bar** — digit swaps to "2", still fits:

page 2

**6. "More" tray open** — real panel, unrelated chrome unaffected by the fix:

more tray

**7. Standard tier (1044x480, taller viewport)** — larger ring, digit was never
clipped here (only the compact-tier 40px floor triggers it):

standard tier

**8. Left-handed mode, compact tier** — mirrored layout, digit still fits:

left handed

**9. Tablet tier (1200x800)**:

tablet tier

**10. Portrait phone (428x926)** baseline:

portrait

**11. Reconnect overlay on a portrait phone** — the real `showReconnectOverlay()`
output (`.fatal-overlay`), what a player now reliably sees while the client
retries after a foreground resume instead of a silently stuck zombie socket:

reconnect overlay

### Reconnect-on-foreground-resume flow

```mermaid
sequenceDiagram
    participant OS as Mobile OS / browser
    participant Page as Page JS (ClientWorld)
    participant WS as WebSocket
    participant Server as Game server

    Note over Page,Server: Player backgrounds the app (lock screen / app switch)
    OS->>Page: freeze JS timers
    OS->>WS: kill transport (no close delivered to a frozen page)
    Note over WS: readyState moves off OPEN,<br/>but onclose never fires

    Note over OS,Page: Player returns to the app
    OS->>Page: resume JS, visibilitychange -> visible

    rect rgb(90,40,40)
    Note over Page: Before: nothing listens for visibilitychange.<br/>connected stays true, no reconnect scheduled.<br/>Player sees a frozen world ("disconnected").
    end

    rect rgb(30,70,40)
    Note over Page: After: handleVisibilityChange fires on resume
    Page->>Page: ws.readyState !== OPEN?
    alt reconnectTimer pending (mid-backoff)
        Page->>Page: clearTimeout(reconnectTimer)
        Page->>Server: openSocket() now, skip remaining backoff
    else socket zombie, no timer scheduled yet
        Page->>Page: socketClosed() (same path as a real onclose)
        Page->>Server: openSocket() after 1s backoff
    end
    Server-->>Page: auth + resume linkdead session (5 min grace)
    Page->>Page: connected = true, onReconnected()
    end
```

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green through typecheck and all
      three builds; the `sfx check` failure is pre-existing on a clean
      `release/v0.25.0` checkout (verified) and unrelated to this change. I added
      decisive tests for the new reconnect behavior and the CSS fix.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified the action-ring fix at the
      exact 1044x480 compact-tier landscape viewport from the original report,
      plus standard/tablet tiers, left-handed, and a portrait phone. The
      reconnect fix is net-layer only (no new UI beyond the existing reconnect
      overlay).
- [x] **Accessible.** No new interactive elements; existing ARIA/focus behavior
      on the ring and its buttons is untouched.

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Solid diagnosis on both fronts, and the reconnect fix is the right shape: force a real socket-state check on resume instead of trusting a `close` event that mobile never delivers. Routing the max-attempts path through `endSession()` (`online.ts:1289`) is a nice catch too, it now clears the reconnect timer and the new listener that the old inline path left behind. The visibilitychange listener lifecycle is correct (same bound arrow added and removed, both `typeof document` guarded), and the tests are decisive: I ran `net_online_visibility_reconnect.test.ts` and `client_shell.test.ts` at the head and both are green. The CSS math holds too: the 0.85 chrome-scale clamps the compact toggle to the 40px floor (36px inner box) and the base 20px digit + 15px icon stack is 37px, so shrinking the digit to 15px (32px stack) clears it with room to spare.

Two things on the reconnect handler.

**should-fix** `src/net/online.ts:1250`: `reconnectTimer !== undefined` is not a reliable "a backoff is pending" signal. It is assigned at `:1299` but never reset when the timer fires: the `() => this.openSocket()` callback and `openSocket()` both leave it holding the stale, already-fired id (only `:1252` and `endSession` ever touch it, and `hello` does not). So once the client has been through one reconnect cycle, a resume while that just-fired attempt's socket is still `CONNECTING` takes the pending branch and calls `openSocket()` again, orphaning the in-flight socket and opening a second one that re-auths the same character. Before this PR `openSocket()` only ever ran with the previous socket already closed, so that orphan path is new. The server's transient-conflict tolerance absorbs it, but it works against the clean-reconnect goal on the exact mobile-flap path this targets. Resetting `this.reconnectTimer = undefined` at the top of `openSocket()` fixes it and makes the "No reconnect scheduled yet" comment true (the CONNECTING-window resume then becomes a clean no-op). Worth a fifth case covering that window.

**nit** `src/net/online.ts:1259`: the zombie branch gates on `this.connected`, so a socket that dies silently before the first `hello` (connected still false, no `reconnectTimer`, no `onclose` ever) is not recovered on resume. Very narrow (backgrounding during the initial handshake), but gating the fall-through on `this.ws.readyState !== WebSocket.CONNECTING` instead would cover it and still leave a genuine in-progress connect alone.

PR #1823, opened 2026-07-12
feat(sim): grant character XP from gathering and crafting

jgyy:
## Summary

Gathering nodes and crafting recipes previously granted only profession
proficiency, with zero character XP: leveling was combat-only. This makes
professions a genuinely relevant way to level, not just a side activity.

Both actions now grant character XP through the existing `grantXp` seam,
tier-scaled and level-gated with the same classic green/gray falloff kill
XP already uses (`mobXpValue`/`zeroDiff` in `src/sim/types.ts`), so a
max-level character farming trivial nodes/recipes gets nothing free, and a
fresh character gets a meaningful amount from either path.

```mermaid
flowchart TD
    subgraph Gathering
        A[Player harvests a node] --> B[harvestNode: grants material item + gathering proficiency]
        B --> C["gatherActionXp(node.level, player.level)"]
    end
    subgraph Crafting
        D[Player crafts a recipe] --> E[resolveCraftForRecipe: consumes reagents, grants item + craft skill]
        E --> F["craftActionXp(recipe.level, player.level)"]
    end
    C --> G["ctx.grantXp(amount, meta)"]
    F --> G
    G --> H[existing level-up path: XP bar, level-up FCT/toast]

    style C fill:#2d5,stroke:#333
    style F fill:#2d5,stroke:#333
    style G fill:#59f,stroke:#333
```

```mermaid
graph LR
    subgraph "professionActionXp(base, contentLevel, playerLevel)"
        direction LR
        X["diff = contentLevel - playerLevel"] --> Y{"diff >= 0?"}
        Y -- yes --> Z["base * (1 + 0.05 * min(diff, 4))<br/>content at/above player level: full or scaled-up XP"]
        Y -- no --> W{"-diff >= zeroDiff(playerLevel)?"}
        W -- yes --> V["0 XP (gray band)"]
        W -- no --> U["base * (1 - -diff / zeroDiff)<br/>linear falloff below player level"]
    end
```

## Related issues

N/A (implements the "nodes and crafting give exp" idea from discussion; no
tracking issue was filed for this specific slice).

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## What changed

- `src/sim/types.ts`: `GatherNodeDef` gains a `level` field.
- `src/sim/content/gather_nodes.ts`: `level` populated for all 15 existing
  nodes, snapshotted from their zone's `levelRange` midpoint (e.g.
  `eastbrook_vale [1,7]` -> 4).
- `src/sim/professions/types.ts`: `ProfessionRecipeRecord` gains a `level`
  field.
- `src/sim/content/recipes.ts`: `level` populated for all 15 existing
  recipes (seeded from `itemLevelBudget`, hand-adjusted where a recipe's
  intended character level clearly diverged from its gold-sink budget).
- `src/sim/professions/profession_xp.ts` (new): `gatherActionXp` /
  `craftActionXp`, two independently tuned green/gray curves (crafting
  scores higher than gathering, since it consumes reagents). Pure,
  deterministic, no DOM/browser imports, no randomness.
- `src/sim/professions/gathering.ts`: `harvestNode` calls `ctx.grantXp(...)`
  on a successful harvest.
- `src/sim/professions/crafting.ts`: `resolveCraftForRecipe` calls
  `ctx.grantXp(...)` on a successful craft.
- Tests: new `tests/profession_xp.test.ts` (curve unit tests: same-level,
  gray-band, reduced-band, scale-up cap, determinism); XP-grant and
  gray-band regression tests added to `tests/gather_node_harvest.test.ts`
  and `tests/professions_crafting.test.ts`; three other test files updated
  with the new required `level` field on their `ProfessionRecipeRecord`
  fixtures.

## How was this tested?

- Commands: `npx tsc --noEmit`; `npx vitest run` across
  `tests/profession_xp.test.ts`, `tests/gather_node_harvest.test.ts`,
  `tests/professions_crafting.test.ts`, `tests/professions_gathering.test.ts`,
  `tests/gathering.test.ts`, `tests/gather_nodes.test.ts`,
  `tests/professions_perks.test.ts`, `tests/archetype_ceiling.test.ts`,
  `tests/professions_acquisition_salvage_sink.test.ts`,
  `tests/architecture.test.ts`, `tests/professions.test.ts`,
  `tests/professions_crafting_hub.test.ts`, `tests/sim.test.ts`,
  `tests/progression.test.ts`, `tests/parity/harness.test.ts`,
  `tests/guide.test.ts`, `tests/localization_fixes.test.ts`; `npm run build`.
  All green. `npx @biomejs/biome check --write` on every changed file, no
  format diffs remaining.
- `npm run gate`: green except `sfx check`, which fails identically on a
  clean checkout of `release/v0.25.0` with none of this PR's changes
  applied (confirmed via `git stash`) — a pre-existing, environment-local
  audio-tooling mismatch unrelated to this change (no audio files touched).
- Manual steps: none (this is a `src/sim/` behavior change with no new UI;
  the XP gain flows through the existing level-up/XP-bar/FCT pipeline
  unchanged).

## Screenshots / recordings

N/A. This is a `src/sim/` gameplay-math change only: no new UI, window,
tooltip, or HUD element. The XP gain renders through the pre-existing
level-up/XP-bar/FCT pipeline, which is visually unchanged (there is nothing
new to compare before/after in a screenshot). See the mermaid diagrams
above for the flow instead.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green except the
      pre-existing, environment-local `sfx check` failure noted above (verified
      unrelated by reproducing it on the same base with no changes applied). I
      added decisive tests for the new behavior (see "How was this tested?").

### Cross-platform

- [ ] N/A. No UI, so no desktop/mobile viewport or a11y surface changed.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings: XP gain reuses the
      existing level-up/XP-bar/FCT pipeline verbatim, which already renders
      through `t()`. No new toast/message is introduced (kill XP doesn't get
      one either, for consistency).

### Hygiene

### DIFF
--- a/src/sim/content/gather_nodes.ts
+++ b/src/sim/content/gather_nodes.ts
@@ -7,29 +7,110 @@ import type { GatherNodeDef, GatherNodeType } from '../types';
 
 export const GATHER_NODE_TYPES: readonly GatherNodeType[] = ['ore', 'wood', 'herb'];
 
+// `level` (issue: profession XP) is a one-time snapshot of each node's zone
+// levelRange midpoint (eastbrook_vale [1,7] -> 4; mirefen_marsh, zone2's
+// levelRange [6,13] -> 10), not a live lookup: see types.ts GatherNodeDef.
 export const GATHER_NODES: GatherNodeDef[] = [
   // Eastbrook Vale (eastbrook_vale), ore near Boar Meadow's rocky outcrops
-  { id: 'ore_eastbrook_1', zoneId: 'eastbrook_vale', type: 'ore', pos: { x: 72, z: 8 } },
-  { id: 'ore_eastbrook_2', zoneId: 'eastbrook_vale', type: 'ore', pos: { x: 78, z: -6 } },
-  { id: 'ore_eastbrook_3', zoneId: 'eastbrook_vale', type: 'ore', pos: { x: 66, z: 22 } },
+  { id: 'ore_eastbrook_1', zoneId: 'eastbrook_vale', type: 'ore', pos: { x: 72, z: 8 }, level: 4 },
+  {
+    id: 'ore_eastbrook_2',
+    zoneId: 'eastbrook_vale',
+    type: 'ore',
+    pos: { x: 78, z: -6 },
+    level: 4,
+  },
+  {
+    id: 'ore_eastbrook_3',
+    zoneId: 'eastbrook_vale',
+    type: 'ore',
+    pos: { x: 66, z: 22 },
+    level: 4,
+  },
 
   // Eastbrook Vale, wood stands around Webwood
-  { id: 'wood_eastbrook_1', zoneId: 'eastbrook_vale', type: 'wood', pos: { x: -62, z: 8 } },
-  { id: 'wood_eastbrook_2', zoneId: 'eastbrook_vale', type: 'wood', pos: { x: -57, z: -6 } },
-  { id: 'wood_eastbrook_3', zoneId: 'eastbrook_vale', type: 'wood', pos: { x: -68, z: 18 } },
+  {
+    id: 'wood_eastbrook_1',
+    zoneId: 'eastbrook_vale',
+    type: 'wood',
+    pos: { x: -62, z: 8 },
+    level: 4,
+  },
+  {
+    id: 'wood_eastbrook_2',
+    zoneId: 'eastbrook_vale',
+    type: 'wood',
+    pos: { x: -57, z: -6 },
+    level: 4,
+  },
+  {
+    id: 'wood_eastbrook_3',
+    zoneId: 'eastbrook_vale',
+    type: 'wood',
+    pos: { x: -68, z: 18 },
+    level: 4,
+  },
 
   // Eastbrook Vale, herb patches near Mirror Lake
-  { id: 'herb_eastbrook_1', zoneId: 'eastbrook_vale', type: 'herb', pos: { x: -86, z: 90 } },
-  { id: 'herb_eastbrook_2', zoneId: 'eastbrook_vale', type: 'herb', pos: { x: -92, z: 80 } },
-  { id: 'herb_eastbrook_3', zoneId: 'eastbrook_vale', type: 'herb', pos: { x: -80, z: 95 } },
+  {
+    id: 'herb_eastbrook_1',
+    zoneId: 'eastbrook_vale',
+    type: 'herb',
+    pos: { x: -86, z: 90 },
+    level: 4,
+  },
+  {
+    id: 'herb_eastbrook_2',
+    zoneId: 'eastbrook_vale',
+    type: 'herb',
+    pos: { x: -92, z: 80 },
+    level: 4,
+  },
+  {
+    id: 'herb_eastbrook_3',
+    zoneId: 'eastbrook_vale',
+    type: 'herb',
+    pos: { x: -80, z: 95 },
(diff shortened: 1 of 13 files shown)


FernandoX7 (commented):
This is built the right way: the new curve module is pure and sits behind the existing `grantXp` seam, the shape is an exact match to `mobXpValue` (same +5 percent per level scale-up capped at +4, same linear falloff, same `zeroDiff` gray band), and the grants land after every deny gate in both paths, with the denied-craft and gray-band cases regression-tested. Details I verified against the code rather than the description:

- `grantXp` early-returns on `amount <= 0` (src/sim/combat/damage.ts:730), so the unconditional call with a gray-band zero is a clean no-op, no zero-XP event.
- Rested XP stays kill-only: both grants omit `fromKill` and the rested draw is gated on it (src/sim/combat/damage.ts:735). Matches the classic rule.
- No wire change needed: `level` is static content data shared by all three hosts, and nothing entity-shaped changed.
- Ran the four touched suites at the PR head (81 tests, all green), `tsc` (only the pre-existing environment-local module-resolution errors in unrelated files), and a merge-tree against `release/v0.25.0` (clean).

On balance, since that is what this change lives or dies on: the numbers land where a supplement should. Per same-level action: a quest turn-in is 150 to 700, a kill is 45+5L, a craft is 20+4L (capped at 10 successful crafts per rolling 60s, and each craft costs reagents plus the gold-sink fee), a harvest is 10+2L on a per-player 120s node timer with 15 nodes in the world. So gathering levels you at very roughly a third of combat pace, and crafting is second-order bound by material supply: the reagents are junk drops with no `buyValue`, so there is no NPC gold-to-XP faucet (market purchases are player-to-player, same as gold itself). At the cap, the level-20 tool recipes are never gray and feed `lifetimeXp`/prestige, but they sit behind the level-20 hub gate and the same throttle, consistent with kill XP already feeding prestige.

Findings:

- should-fix: the craft curve's magnitude is unpinned. tests/profession_xp.test.ts:36 pins the gather constants to literals (`10 + 2 * 9` and the 1.2x cap), but the only craft assertions are `toBeGreaterThan(0)` and `not.toBe(gather)` (tests/profession_xp.test.ts:12), so `CRAFT_XP_BASE`/`CRAFT_XP_PER_LEVEL` could regress by 10x with the whole suite staying green. The tuning values are the risk surface of this PR; one exact pin closes it, e.g. `expect(craftActionXp(5, 5)).toBe(40)`.
- nit: nothing pins the rested-XP exclusion. If a later change adds `{ fromKill: true }` to either grant, profession XP silently starts draining the rested pool at double value. A one-line assertion that `meta.restedXp` is unchanged after a rested player harvests would lock the rule in.
- nit, tuning note only: the Eastbrook armor recipes are authored at level 8 to 10 (src/sim/content/recipes.ts), above the zone's own 1 to 7 band, so everyone in the zone's intended bracket earns the capped scale-up on them: a level-1 crafting the chain vest gets 72 XP, more than a same-level kill (50). It is bounded by 3 bone fragments per craft so it is slow in practice; just the first place to look when playtest data comes in. No change requested now, the module's own comment already flags the constants as first-pass tuning.

The snapshot comments on `level` (zone midpoint for nodes, `itemLevelBudget` seed for recipes) are exactly what the next balance pass will need.

PR #1828, opened 2026-07-12
feat(sfx): share one recorded voice across multiple mob subfamilies

jamiecypher:
## Summary

Adds a small alias so multiple mob templateIds can share one recorded subfamily voice, instead of each needing its own separate near-identical recording.

`mobVoiceCue()` builds a specific subfamily cue key strictly from a mob's exact templateId (`mob_<family>_<templateId>_<action>`). That's correct for creatures that genuinely sound distinct, but there are several wolf-shaped `beast`-family mobs across the game (`forest_wolf`, `ridge_stalker`, `mire_prowler`, `old_greyjaw`) that all use the same wolf model, and none of them have a name that would naturally group them together for lookup purposes.

`SUBFAMILY_ALIAS` is a `templateId -> shared subfamily name` map consulted before the specific key is built. Any templateId not listed keeps using its own raw id, unchanged. First use: all four wolf-shaped beasts alias to `'wolf'`, so recording `mob_beast_wolf_<action>` once covers all four instead of requiring four separate takes for visually identical creatures.

This is intentionally scoped to just the alias mechanism. A follow-up PR will fix a handful of mistagged mob `family:` values found while auditing this (unrelated content-data issue, different subsystem, no audio yet to back it) and add a new `reptile` family for a raptor-modeled mob currently mistagged `beast`.

## Type of change

- [x] Test: new/updated coverage

## How was this tested?

- `npx tsc --noEmit` clean
- `npx vitest run tests/combat_sfx.test.ts tests/sfx.test.ts tests/sfx_manifest.test.ts tests/sfx_conform.test.ts tests/architecture.test.ts` — 94 tests passing
- Updated the existing `combat_sfx.test.ts` case that asserted `forest_wolf` keyed off its own raw templateId (now correctly aliased, so that assertion moved to `bog_bloat`, which has no alias and demonstrates the original per-template behavior is unchanged for anything not listed)
- Added a dedicated test covering all four aliased templateIds resolving to the shared `wolf` cue when present, a non-aliased mob (`crypt_shambler`) staying unaffected, and every aliased template correctly falling back to the generic family sound when no wolf recording exists yet

## Screenshots / recordings

N/A — no visual change, no audio exists for this yet (the alias resolves to a fallback until `mob_beast_wolf_*` is actually recorded).

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [ ] Builds succeed. (not run locally this pass, no build-affecting changes)

Cross-platform
- ~Tested on desktop and mobile.~ N/A, audio-routing logic only

Localization (i18n)

Hygiene


EnriqueGF:
:) Thanks for this, clean and well-scoped contribution.

Reviewed and it looks production-ready. What I verified against the real code:

- Base branch: targets release/v0.26.0 (the latest release branch) and is up to date with it. Gate passes.
- Correctness: all four aliased templateIds (forest_wolf, ridge_stalker, mire_prowler, old_greyjaw) are confirmed family: 'beast' in the zone content, so the derived key mob_beast_wolf_<action> is well-formed. The alias is consulted only for the specific-cue key; the family prefix still comes from mobVoiceFamily, and any unlisted templateId keeps its own raw id (verified bog_bloat and crypt_shambler behavior).
- Tests: the moved bog_bloat case and the new shared-wolf test are decisive: they assert both the aliased hit (mob_beast_wolf_attack) and the family fallback (mob_beast_attack), plus a negative case (crypt_shambler -> mob_undead_attack). Would fail on regression.
- Purity/i18n: mobVoiceCue stays a pure ui-layer function; the cue keys are asset identifiers, not player-visible text, so no t() needed. No em/en dashes or emojis.

Ran locally in a worktree: npx tsc --noEmit clean, tests/combat_sfx.test.ts + tests/architecture.test.ts (35 passing), biome clean on both changed files.

One optional note for the follow-up you mentioned: once a real mob_beast_wolf_* take is recorded, PR #1854's guard will require the numbered suffix (_1), which is the correct shape, just worth landing them in the right order.

PR #1830, opened 2026-07-12
feat(dev): /dev sandbox practice scenario (ALLOW_DEV_COMMANDS)

Blaine1705:
## Summary

Adds `/dev sandbox`, a controlled practice scenario for testing abilities, reachable **only** through the `ALLOW_DEV_COMMANDS` chat router. It spawns:

- a **non-offensive training dummy** in front (the existing `training_dummy`: `aggroRadius 0` / `moveSpeed 0`, so it never chases you), to build resources on, and
- a **raid of friendly level-20 allies** with a **roomy 10k health pool**, started LOW, so a healer can top them off for a good while (and the bar visibly climbs) without them capping, with **out-of-combat regen frozen** so only your own healing (or damage) moves their bars.

The regen freeze uses a zero-value "food" that trips the existing `!p.eating` guard in `updateRegen` and heals nothing; an idle, unsitting bot next to the non-damaging dummy never clears it. Attack the dummy to build resources, then practice heals or AoE on the raid with a clean, threat-free readout of what your abilities actually do (no ambient self-healing, no aggro on the tester).

Re-running `/dev sandbox` **clears the previous scenario** (dummy + bots) before the fresh spawn, so it resets the hurt allies instead of piling on more bodies; only the entities this command spawned are removed, so the world's own training dummy is never touched.

It reuses the existing dev-bot / training-dummy / party machinery: a `Sim.startDevSandbox` method behind the `SimContext` seam, plus a dev-gated `/dev sandbox` chat command whose confirmation is emitted from the chat handler (like `/dev bot`).

## Type of change

- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx tsc --noEmit` (clean)
  - `npx @biomejs/biome check` on the changed files (clean)
  - `npx vitest run tests/dev_sandbox.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/sim_context.test.ts tests/entity_roster.test.ts tests/chat.test.ts tests/sim.test.ts` (green)
  - `node scripts/build_server.mjs` (server bundle OK)
  - pre-push QA floor (tsc + guard tests + biome + copy scan): green
- Manual steps: on a dev realm (`ALLOW_DEV_COMMANDS=1`), typed `/dev sandbox`, confirmed a training dummy spawns in front and a raid of level-20 allies at reduced HP appears; attacked the dummy and cast heals on the allies, and their bars moved only from the healing (no self-regen, nothing chased the healer).

`tests/dev_sandbox.test.ts` covers: the command is a no-op without dev commands; it spawns the dummy plus a raid of level-20, reduced-health, regen-frozen bots; and the bots never self-heal while idle.

## Screenshots / recordings

N/A: dev tooling, no player-facing UI change.

---

### Quality

- [x] Added decisive tests for the new behavior (`tests/dev_sandbox.test.ts`) and recorded the manual checks above. Local guard tests, `tsc`, biome on changed files, and the server build are green; CI runs the full gate.

### Cross-platform

- N/A: no UI.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings. The `/dev sandbox` confirmation is dev-channel English diagnostics (like the existing `/dev bot` feedback), out of i18n scope; `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets or `.env` committed. `/dev sandbox` is gated behind `ALLOW_DEV_COMMANDS` and is never reachable in a production path.


EnriqueGF:
Thanks for the contribution :) The idea of a threat-free practice scenario for dev realms is genuinely useful, and the command is correctly dev-gated. I checked the branch out, ran tsc, the guard suites and the parity tests, and reproduced the behavior in the sim. There are some blocking items before this can go in.

## Blocking

**1. Wrong base branch.** This PR targets `release/v0.24.0`, but the current release branch is `release/v0.25.0` (`origin/release/v0.25.0` exists and is ahead). Contributions must be based on the latest release branch. Please re-target and rebase onto `release/v0.25.0`, then re-run the gate.

**2. Damaging a practice bot permanently breaks the regen freeze** (`src/sim/sim.ts:2121-2130`). The freeze relies on a zero-value fake `eating` entry, but `src/sim/combat/damage.ts:410-413` clears `target.eating` on any damage taken. So the moment the tester does the AoE practice this PR advertises, the ally resumes normal out-of-combat regen and self-heals. Reproduced: one 100-damage hit on a practice bot, then 30 seconds idle, and its HP climbed from 1400 to 1556 on its own. The claim that "only your own healing (or damage) moves their bars" holds only for the never-damaged case.

**3. Any buff on a practice bot collapses the 10k pool** (`src/sim/sim.ts:2119`). `e.maxHp = 10_000` is a raw override of a derived stat. `src/sim/combat/auras.ts:200-205` marks stats dirty on any buff aura add or expiry and calls `recalcPlayerStats`, which recomputes `maxHp` from level and stamina. Reproduced: after one recalc, a bot goes from maxHp 10000 / hp 1500 to maxHp 418 / hp 63. A healer buffing the raid (stamina and intellect buffs are the first thing a healer casts) destroys the scenario immediately, and even a plain expiring buff does it. A durable approach would model the pool through the stats/aura pipeline (for example a maxHpPct-style aura, or a mob-based practice target) instead of writing `maxHp` directly.

## Code quality

**4. Module-first.** This adds about 105 lines of new self-contained system behavior to `src/sim/sim.ts` (7.5k lines), which CLAUDE.md explicitly forbids growing: new sim system behavior lands as its own module behind the `SimContext` seam (for example `src/sim/dev_sandbox.ts`), with `sim.ts` a thin delegator. The callback registration is right; the logic itself should not live in the coordinator.

**5. Shared per-Sim state.** `devSandboxIds` (`src/sim/sim.ts:1297`) is per-Sim, not per-player. On a shared dev realm, a second player running `/dev sandbox` despawns the first player's dummy and raid allies out from under them. Keying by caster pid would avoid that.

## Tests

**6.** `tests/dev_sandbox.test.ts` only exercises the idle arm of the central claim. There is no case for "damage the bot, then confirm regen stays frozen" (currently fails) and none for "buff the bot, then confirm the pool survives" (also fails). The reset test only compares `entities.size`, which would still pass if old bots leaked elsewhere, and the dummy's non-offensiveness (aggroRadius 0 / moveSpeed 0) is asserted only via `hostile === true`. Minor: the `as SimEvent[]` cast on line 71 is a no-op and biome flags it.

## Verified green

`npx tsc --noEmit` clean; `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/parity`, `tests/world_api_parity.test.ts`, `tests/sim.test.ts`, `tests/sim_context.test.ts`, `tests/entity_roster.test.ts`, `tests/chat.test.ts` and `tests/dev_sandbox.test.ts` all pass; biome reports warnings only; dev gating holds (no-op without ALLOW_DEV_COMMANDS). Screenshots are correctly N/A for dev tooling.

Happy to re-review once it is re-targeted onto `release/v0.25.0` and the freeze/pool mechanics are made robust.

EnriqueGF:
Follow-up from the domain reviewers (architecture and test coverage), which independently reproduced the two defects above and found some additional items worth folding into the same revision:

- **The party to raid sequence assumes the caster is a partyless leader** (`src/sim/sim.ts:2104-2113`). `PartyMachine.convertPartyToRaid` (`src/sim/social/party.ts:261-280`) requires the caller to be the party leader, not already a raid, and to have at least 5 members. If the tester is already grouped with real players on a dev realm, the invites land in that party and the convert either errors out or turns a real party into a raid without consent. On a re-run, the caster is typically left in a party still flagged `raid`, so `convertPartyToRaid` emits "Your group is already a raid." while `startDevSandbox` still returns `botIds.length`, so the confirmation message reports 5 allies regardless. Returning the number of bots that actually joined the group, and bailing out (or leaving/disbanding first) when the caster is already grouped, would make it honest.
- **No cleanup when the caster leaves.** `removePlayer` does not clear `devSandboxIds` or despawn the dummy and bots, so a dev realm accumulates orphaned `Practice*` players until someone re-runs the command.
- **Bot names are not unique.** `spawnSandboxBot` (`src/sim/sim.ts:2041`) deliberately drops the name-uniqueness gate that `spawnDevBot` enforces, so two testers both get a `Practice1` and `/w Practice1` becomes ambiguous. Suffixing with the caster pid is enough.
- **Comment wording** (`src/sim/social/chat.ts:985-990`): the comment explains that the string is routed "as a var" to keep it "out of the S3 sim.ts emit scanner". The outcome is fine (dev-channel English is a sanctioned carve-out, same as `/dev bot`), but please phrase it as the scope rule rather than as a way of avoiding the guard.

Everything else came back clean: no rng draws, no tick-phase or draw-order change (`tests/parity` green), sim purity intact, and the `SimContext` addition is append-only with both fixtures updated.

EnriqueGF:
Thanks for turning this around so quickly :) Re-reviewed at `02ce3a93b`.

Good news first: **the base-branch gate now passes.** The PR targets `release/v0.25.0`, which is the current release branch, and it rebases cleanly onto it.

However, **the code itself is unchanged.** Comparing the PR diff at the old head (`b1640e889`) with the new one (`02ce3a93b`), the added and removed lines are byte-for-byte identical: this revision is a rebase / re-target only. All the findings from the previous round are still open, and I re-confirmed the two functional defects against the rebased head:

1. **Damage still clears the regen freeze** (`src/sim/sim.ts:2124`). After one 100-damage hit on a practice bot, `eating` is `null` and the bot self-healed from 1400 to 1556 HP over 30 seconds idle.
2. **A stat recalc still collapses the 10k pool** (`src/sim/sim.ts:2119`). A single `recalcPlayerStats` call (which any buff aura add or expiry triggers) takes a bot from maxHp 10000 / hp 1500 to maxHp 418 / hp 63.

Also still open: the module-first placement (105 lines of new system behavior in the `sim.ts` coordinator instead of its own module behind the `SimContext` seam), the per-Sim `devSandboxIds` that clobbers another tester's sandbox, the party-to-raid sequence assuming a partyless leader (so the "5 raid allies" confirmation can be untrue), the missing cleanup on caster removal, the non-unique `Practice*` bot names, and the test gaps (no damage-then-idle case, no buff-then-check-pool case).

Verified green on the rebased head: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity` and `tests/dev_sandbox.test.ts` all pass, and `tsc` reports no errors in the changed files.

The blocker is now the mechanics, not the branch. Happy to re-review once the freeze and the health pool go through mechanisms the sim respects (a regen flag rather than a fake `eating` payload, and a stat-channel mod rather than a raw `maxHp` write).

Blaine1705:
Thanks for the review. Reworked so the two fragile mechanisms now go through paths the sim respects, plus the structural fixes:

**Blocking**
- **Health pool**: replaced the raw `e.maxHp = 10000` write with a `buff_sta` Stamina aura applied through `recalcPlayerStats` (inside `applyAura`), so a later recalc rebuilds the ~10k pool instead of collapsing it. Pinned by a "frozen pool survives a stat recalc" test.
- **Regen freeze**: replaced the fake `eating` payload with a `devFreezeRegen` Entity flag that `updateRegen` honors and combat damage never clears. Pinned by a "regen stays frozen even after damage" test (it also asserts damage still wipes a real `eating` payload, the old failure mode).

**Structural**
- **Module-first**: all logic moved to `src/sim/dev_sandbox.ts` behind the SimContext seam (added `partyAccept`/`convertPartyToRaid`/`removePlayer`); `Sim.startDevSandbox` is now a thin delegator and `spawnSandboxBot`/the per-Sim `devSandboxIds` field are gone.
- **Per-caster state**: the spawned ids live on `PlayerMeta.devSandboxIds`, so two testers no longer clobber each other.
- **Party/raid**: bails with feedback if the caster is already grouped, so the bots never hijack a real party/raid.
- **Cleanup**: `Sim.removePlayer` despawns the leaving tester's sandbox (no orphans).
- **Names**: pid-suffixed so they never collide; reworded the chat.ts comment to the scope rule.

**Tests added**: damage-then-regen-frozen, buff-then-pool-survives-recalc, stronger reset (old bots gone, fresh ones hurt again), dummy non-offensiveness (moveSpeed 0 + never aggros/chases), leave-cleanup, already-grouped bail.

Locally green: `tsc`, changed-files biome, the S3 i18n guard, `tests/architecture.test.ts`, `tests/parity` (unchanged, no golden shift), the three server builds, and the dev_sandbox/sim_context/entity_roster suites. A fresh architecture-reviewer pass came back with 0 blocking / 0 should-fix.

jgyy:
Closing — this looks already implemented. `sim_context.ts` on `upstream/main` already exposes a `startDevSandbox` API for the /dev sandbox practice scenario. Thanks for the contribution.

PR #1838, opened 2026-07-13
fix(sim): bound pet recovery and add CPU incident profiling

### LINKED ISSUE #1833: Sim tick stalled for six minutes in production: mob.update climbed to 98ms mean per tick
## Problem

On 2026-07-12 from 18:21 to 18:27 UTC the production sim stalled for about six minutes: the `mob.update` phase climbed from its ~3.6ms per-tick baseline to a mean of ~98ms (p95 148ms, max 188ms), blowing the 50ms tick budget on essentially every tick for the duration. The world froze and rubber-banded badly enough that players reported the server as down. About 10 of the ~100 online players disconnected during the window. The stall self-recovered by 18:27 with no restart.

Population was normal throughout (~100 players online, ~750 entities), so this was not a spawn flood or a login wave: something inside `mob.update` became roughly 25 times more expensive at steady entity count and then subsided. The perf log attributes cost only to the phase level, so the logs cannot say which zone, mob kind, or AI path was responsible.

## Steps to reproduce

1. Run the server with ~100 connected players and ~750 live entities.
2. Reach the (currently unknown) mob state that inflates `mob.update`, suspected to be an AI or pathing pile-up around clustered engaged mobs.
3. Observe sustained tick overruns.

## Actual behavior

- `mob.update` mean cost per tick, by minute: 3.6ms baseline, 7.9 (18:21), 25.9 (18:22), 62.8 (18:23), 98.5 (18:24), 96.3 (18:25), 79.3 (18:26), then a slow decay back to baseline.
- The tick overrun marker fired 58 to 59 times per minute at the peak (versus 0 to 5 in a normal minute), with total tick time reaching 205ms against the 50ms budget.
- Online count dropped from ~100 to 90 during the window and recovered immediately after.

## Expected behavior

- `mob.update` stays within the tick budget at this population (~100 players, ~750 entities), including whatever mob configuration produced the spike.
- If a pathological case does occur, degradation is graceful rather than a sustained 4x tick-budget overrun.
- The perf log attributes `mob.update` cost at a granularity that identifies the responsible zone or mob group, so an incident like this is diagnosable from logs alone.

## Scope

- Investigate the `mob.update` hot paths for superlinear behavior with clustered or engaged mobs (AI target scans, pathing, aggro checks).
- Add finer-grained perf attribution inside `mob.update` (per zone or per top-cost mob group) to the existing `[perf.sim]` reporting.
- A headless stress test that drives the suspected pile-up shape at production population and asserts a per-tick budget for `mob.update`.

## Acceptance criteria

- The root cause of the 18:21 to 18:27 UTC pattern is identified, or instrumentation lands that would attribute it on the next occurrence.
- A regression or stress test bounds `mob.update` at ~100 players and ~750 entities, including the identified worst case.
- No gameplay behavior change: mob AI decisions are identical before and after, only cost and instrumentation change.

## Notes

- The 50ms budget is one 20Hz tick (DT = 1/20); the broadcast phase stayed at its normal ~25ms p95 throughout, so this incident is independent of the known snapshot broadcast cost work.
- Entity and player counts were flat across the window, which points at a state change in existing mobs rather than volume.

Rubsey:
## Summary

- Bounds clear-route pet recovery at 96 yards, preventing very distant pets from repeatedly running the unbounded line-of-sight walk that dominated the captured production CPU profile.
- Adds an event-triggered production CPU monitor that confirms sustained saturation, records a 20-second V8 CPU profile, and retains correlated logs, Docker context, checksums, and optional tick-profiler evidence.
- Adds immutable, root-owned in-image profiler helpers, container restart ownership guards, remote and local capture locks, secure artifact retention, and cooldowns that avoid repeatedly profiling an already overloaded process.
- Makes tick captures wall-clock bounded and records loop callbacks, sim ticks, catch-up callbacks, and maximum ticks per callback for incident correlation.
- Fixes two pre-existing parallel cleanup races exposed by the release gate.

## Related issues

Closes #1833

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run gate` (all 10 steps green, 12,790 tests passed and 18 skipped)
  - `npx vitest run tests/pet_follow.test.ts tests/prod_cpu_monitor.test.mjs tests/server/tick_perf_capture.test.ts`
  - `npx vitest run tests/server/new_endpoint.test.ts tests/sfx_studio_server_security.test.ts --maxWorkers=2`
  - `npm run ops:cpu-monitor -- --help`
- Manual steps:
  - Reviewed the release integration and production monitor safety against `release/v0.25.0`; no blockers remain.
  - No live production profile was triggered from this branch.

## Rollout notes

This PR ships the monitor, server capture support, and immutable container helpers. Enabling continuous capture still requires a private operations follow-up: install the documented supervised service and an exact-command SSH wrapper from the private Ansible repository. Do not grant wildcard Docker access. CPU profiles work without a bearer token; the optional detailed tick capture should remain disabled until a narrow machine credential exists.

---

### Quality

### Cross-platform

- [ ] **Tested on desktop and mobile.** N/A, no player UI changes.
- [ ] **Accessible.** N/A, no UI changes.

### Localization (i18n)

### Hygiene

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -268,6 +268,78 @@ unsupported extra key is rejected by the client and needs a normal game
 deployment instead. Compatible constrained mob-subfamily keys may be added by
 an artifact.
 
+## Automatic production CPU incident capture
+
+`npm run ops:cpu-monitor` watches Docker CPU and attaches to the game only after a
+confirmed trigger. By default it polls every 30 seconds, confirms that two of three
+samples exceed 90%, and then records a 20-second V8 CPU profile at a 4 ms sampling
+interval. The profile is temporary and event-triggered, so the profiler has no
+steady-state game-loop cost.
+
+Run the monitor as a supervised service on an always-on private operations host,
+not in the game container. The Levy Street deployment should manage that service
+in the private Ansible repo, where SSH aliases and credentials already live. A
+representative direct invocation from that host is:
+
+```bash
+npm run ops:cpu-monitor -- \
+  --direct \
+  --host world-of-claudecraft-prod \
+  --container eastbrook-game \
+  --out-dir /var/lib/woc-prod-cpu-monitor
+```
+
+The service unit should use `Restart=always`, `RestartSec=10`, `UMask=0077`, an
+unprivileged local user, and the repository checkout as its working directory. With
+systemd, use `StateDirectory=woc-prod-cpu-monitor` and
+`StateDirectoryMode=0700`; the monitor safely initializes an empty, private,
+service-owned directory on first use. The remote SSH principal needs narrowly
+controlled access to the Docker operations used by the script and to `flock` for
+capture serialization. General `sudo docker` access is effectively root access.
+Use a dedicated principal plus a root-owned forced-command or validation wrapper
+that admits only the expected container and monitor operations; do not grant a
+wildcard Docker sudo rule to an ordinary account.
+
+The monitor verifies private file ownership and permissions, uses local and remote
+exclusive locks, and retains at most 24 validated captures or 30 days of captures.
+Captures and process logs can contain sensitive operational data, so the artifact
+directory must stay private and must not be served over HTTP.
+
+Each incident directory includes `cpu.cpuprofile`, game and perf logs,
+container/process snapshots, Docker stats before/during/after, metadata, and SHA-256
+checksums. Open `cpu.cpuprofile` with the Load profile action in the Chrome DevTools
+Performance panel. Review `metadata.json` first: `complete` should be true and
+`profileStartDelayMs` records the delay until the profiler acknowledged it had
+started. A fully complete directory also contains a `COMPLETE` marker written after
+the metadata and checksum manifest. A valid CPU profile is retained even if
+supporting context is degraded. The `errors` array explains any missing auxiliary
+artifact without triggering a second profile every two minutes.
+
+Detailed tick-profiler JSON is optional. It requires an existing staff bearer that
+can access the `ops.perf` admin routes, supplied through a service-owned mode-0600
+file with `--ops-token-file`. The current role model does not provide a dedicated
+machine-only bearer, so do not copy a broad personal admin session into the service.
+Provision a narrowly scoped service credential in the server before enabling this
+option. When enabled, `tickCapture` in `metadata.json` must be `complete`.
+The tick result also records loop callback count, sim tick count, catch-up callback
+count, and maximum ticks per callback so callback aggregation is visible during a
+saturation event.
+
+Before enabling the service, verify its SSH user has only the required passwordless
+Docker commands and run one controlled check:
+
+```bash
+npm run ops:cpu-monitor -- --once --dry-run --direct \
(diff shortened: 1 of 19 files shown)


FernandoX7 (commented):
This one is genuinely good. I verified the pet fix end to end: on the base branch a stranded pet re-runs the unbounded lineOfSightClear trace every tick (the 500yd reproducer walks about 1,000 samples per tick per pet), and both new tests fail on base exactly as claimed, so the reproducer is decisive. The wall-clock capture close is right for the saturated case the old tick-counted window could not serve, and no stale perfCaptureEndsAtTick references remain. A security pass over the monitor toolkit also came back clean: shell quoting and identifier validation, token handling (stdin piped, never argv, logs, or artifacts), inspector lifecycle (container loopback only, layered cleanup on every failure path), artifact pruning confined to owned capture dirs, and the root-owned /app/ops copy all check out, and safeInspectCommand deliberately omitting Config.Env is a nice touch. The three targeted test files pass at the PR head (49 tests), the sfx fixture name matches the new gitPorcelain filter, and the merge against release/v0.25.0 is conflict-free.

Findings, all nits:

- nit `scripts/prod_cpu_monitor_commands.mjs:104`: `--request ${method}` is the one remote-command interpolation that is not shellQuoted. Callers only pass the literals GET and POST today, so it is not exploitable, but quoting it keeps the file uniformly safe against a future caller.
- nit `scripts/prod_cpu_monitor_core.mjs:283` (help text at `scripts/prod_cpu_monitor.mjs:63`): the help calls --interval-seconds a start-to-start interval (default 30), but minimumDelayMs forces a 30 second gap after every poll, so at the default the real cadence is poll duration plus 30s, roughly 36 to 40s with three spaced samples. Either word the help as a minimum gap or skip the floor when the interval already covers it.
- nit `scripts/prod_cpu_monitor_process.mjs:171`: the 20s acquire timeout rejects without clearing the 10ms check interval. If LOCKED lands in that window the interval still sets `acquired`, and the close handler then rejects the `lost` promise nobody holds, an unhandled rejection that kills the monitor. Clearing the interval inside the timeout handler closes the race.
- nit `scripts/prod_cpu_monitor.mjs:709`: pruneCaptures runs before acquireLocalLock, so two racing monitor instances can prune the same capture directories concurrently and the rm (no force) can throw on the loser. Moving the prune after the lock keeps startup single-writer.

Notes, no action needed:

- Pets more than 96yd from their owner now snap home even when a walkable route exists (for example an owner who outruns the pet past the bounded A* window). That reads as intended and matches classic pet behavior; worth remembering if a player reports pets teleporting more often.
- The optional tick capture flips the detailed sub-phase timing on during the V8 profile window, so the profile includes that small instrumentation cost. Correlation is clearly the point; just noting the known skew.
- The safety model leans on the exact-command SSH wrapper that lives in the private Ansible repo, not here. DEPLOY.md says this plainly and correctly; flagging so that follow-up actually lands before the service is enabled.

Issue #1778, opened 2026-07-11
Camera snaps back to character facing angle after brief right-click rotation


MasterZensei:
## Problem

Holding the right mouse button briefly to rotate the free look camera, then releasing it,
causes the camera to visibly snap back toward the character's facing angle. The character's
rotation appears to lerp toward the camera's heading more slowly than the camera itself
rotates, so on release there is a gap between the two headings that closes with a sudden,
jarring snap instead of a smooth catch-up.

This has been reported for two weeks and has already gone through two prior patch attempts
that did not resolve it. This is the third report of the same underlying issue.

## Steps to reproduce

1. Log in and spawn into the world.
2. Move the character briefly (a few steps in any direction).
3. Hold right-click and drag briefly to rotate the free look camera away from the
   character's current facing angle.
4. Release right-click.
5. Observe the camera snap back toward the character's facing direction.

## Expected behavior

- On releasing right-click, the camera should settle back onto the character's facing
  angle smoothly, with no visible pop or snap.
- The character's turn rate while catching up to the camera heading should be fast enough,
  or the camera's return should be paced to the character's turn rate, so the two never
  visibly diverge and then jump back into sync.

## Actual behavior

- After a brief right-click drag and release, the camera snaps back abruptly to the
  character's facing angle instead of easing back.
- The mismatch is attributed to the character's rotation lerp running slower than the
  camera's free-look rotation, so a gap builds up during the drag and resolves as a snap
  on release.

## Scope

- Free look camera rotation and release-to-follow behavior (right-click drag).
- Character facing rotation lerp/turn-rate that the camera re-syncs to on release.
- Likely `src/game/pointer_lock.ts` and/or `src/game/input.ts` (free look and rotation
  handling) and the corresponding camera follow logic in `src/render/`.

## Acceptance criteria

- Holding right-click briefly, rotating the camera, and releasing no longer produces a
  visible snap back to the character's facing angle, across a range of drag durations
  (very brief taps and longer holds).
- The character's turn rate and the camera's return-to-facing behavior are paced so they
  stay visually in sync, with no abrupt jump on release.
- Regression covered by an automated test (or documented manual repro steps) so this does
  not require a fourth patch attempt.

Rubsey:
Fixed in PR #1839, merged to release/v0.26.0 (a36a78c47): self facing now stays glued to the camera during mouselook drags, so a brief right-click rotation no longer snaps back on release. Ships in v0.26.0.

Issue #116, opened 2026-06-13
Annoying mouse capture spam


fr0:
You're using the browser's mouse capture API means this spam appears at the top of the screen every time I right-click.

<img width="1465" height="204" alt="Image" src="https://github.com/user-attachments/assets/2775d2aa-16d9-42d7-a17a-778ab0597959" />

maxpolaczuk:
Couldn't agree more, it's so annoying. thanks for hte fix!

jgyy:
The pointer-lock-on-rotate behavior is now a toggle (`setLockCursorOnRotate()` in `src/game/input.ts`), so the browser capture banner can be disabled. Closing as implemented.

PR #1840, opened 2026-07-13
fix(input): request pointer lock synchronously on Firefox mousedown

### LINKED ISSUE #1834: Mouse camera rotation is broken on Firefox after the camera snap-back fixes
## Problem

A player reports (Discord, 2026-07-12) that mouse camera rotation is broken on Firefox to the point the game is unplayable for them there. They believe it regressed with the recent patches that attempt to fix camera snap back. On Chrome, mouse rotation behaves correctly for the same player, but they strongly prefer to keep playing on Firefox.

Firefox is a supported browser for the web client, so this is a cross-browser regression, not something we can leave at "use Chrome".

## Steps to reproduce

1. Open the game in Firefox (desktop).
2. Enter the world and rotate the camera with the mouse (mouse-look).
3. Observe the rotation misbehaving.

## Actual behavior

- Mouse camera rotation misbehaves on Firefox; the reporter describes it as unplayable.
- The same flow on Chrome (including incognito) works correctly for the same player.

## Expected behavior

- Mouse camera rotation on Firefox is smooth and matches the Chrome behavior.
- The camera snap-back fix keeps working on both browsers (no regression back to the old bug).

## Scope

- Reproduce on Firefox and pin down the exact symptom.
- Audit the recent camera snap-back and mouselook fixes for Firefox-specific assumptions. Candidate regression sources to check: the fullscreen mouselook alignment fix (PR #1458) and the camera-follow fixes around the touch-look release echo.
- Check pointer lock and mousemove delta handling: Firefox historically reports movementX/movementY in different units than Chromium, which is a classic source of exactly this kind of divergence.

## Environment

- Platform: desktop web, Firefox (broken); Chrome unaffected
- Surface: in-game camera mouse-look

## Acceptance criteria

- Mouse camera rotation on Firefox is smooth and playable, with parity to Chrome.
- The original camera snap-back fix still holds on both browsers.
- Verified manually on current Firefox and Chrome.

## Notes

- Same report also mentions that keybinds and UX options do not carry across browsers (settings live in per-browser local storage). That is expected with the current design and out of scope here; a separate issue for account-side settings sync may be worth filing.

jgyy:
## Summary

Fixes the root cause of #1834: mouse camera rotation broken on Firefox after the
camera snap-back fixes. It is not a snap-back regression and not a movementX/movementY
units mismatch; the actual bug is a Firefox pointer lock security restriction.

Camera rotation locks the pointer on the instant a mouse press becomes a real drag
(distance/time threshold, `CAMERA_DRAG_START_DISTANCE` / `CAMERA_DRAG_START_MS`),
requesting `canvas.requestPointerLock()` from inside the `mousemove` handler once the
threshold is crossed, well after the originating `mousedown`. Firefox denies that call
outside of Fullscreen API fullscreen: it requires `requestPointerLock()` to run
synchronously inside the handler for the genuine user gesture itself (`mousedown`), not
a later `mousemove`, even though that `mousemove` still fires within the mousedown's
lingering "transient activation" window. Firefox's own error message says this
verbatim: "was not called from inside a short running user-generated event handler, and
the document is not in full screen." Chromium is lenient here and happily consumes the
lingering activation from the deferred call, which is exactly why this only breaks
Firefox and was never caught on Chrome.

This is a refinement of the investigation notes on the issue, not the original
"fullscreen exemption" or "movementX units" theories: neither survives the code. PR
#1458 (the most recent commit touching this path) only removed the Fullscreen-API
exemption in `shouldEngagePointerLock`, which does not affect a plain windowed Firefox
user (the reporter is not in Fullscreen API fullscreen); the deferred-mousemove call
site in `onMouseMove` is unchanged by #1458 and has requested the lock from `mousemove`
since the pointer-lock-on-drag feature was first added. The mechanism above explains a
plain windowed-mode Firefox regression on its own, independent of #1458.

### Fix

- `src/game/pointer_lock.ts`: add two pure, unit-tested decision helpers:
  - `pointerLockNeedsSyncGesture(userAgent)`: true for Firefox user agents.
  - `shouldEngagePointerLockOnMouseDown(input)`: true only when the browser needs the
    synchronous path AND the pressed button is the button that can start a camera drag
    in the active camera mode (right button in classic mode, left button in Mouse
    Camera mode). Gating on the specific look button, not every mousedown, keeps a
    plain click-to-move click from ever requesting a lock.
- `src/game/input.ts`: `onMouseDown` now requests the lock synchronously on Firefox for
  the camera-look button, ahead of any drag-threshold check. A press that turns out to
  be a plain click still releases the lock on mouseup via the existing
  `shouldReleasePointerLock` path (unchanged). Chromium keeps the original
  deferred-mousemove request untouched, so the #116 fix (no pointer-lock banner on a
  quick right-click used for attack/look) is preserved there exactly as before.

The camera snap-back fix from #1778 (the character-facing lerp / camera-follow catch-up
on release) is untouched: this PR only changes *when* the pointer lock request happens,
never the yaw/pitch integration or the release-facing behavior.

Why this matters: on Firefox, path D/H/I above (the pre-existing behavior) silently
fails `requestPointerLock()` outside fullscreen, so the cursor is never actually locked;
once it reaches the screen edge `movementX` clamps to 0 and rotation freezes/stutters,
matching the reporter's "unplayable" description. Path C/E/F is the fix: it moves the
call to the one place Firefox accepts it.

## Related issues

Fixes #1834

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/pointer_lock.test.ts tests/input.test.ts` (green, 90 tests,
    including 4 new `pointerLockNeedsSyncGesture`/`shouldEngagePointerLockOnMouseDown`
    unit tests and 5 new Firefox-vs-Chromium `Input` integration tests using a
    UA-stubbed `navigator`)
  - `npx vitest run tests/architecture.test.ts tests/keybinds.test.ts tests/mobile_controls.test.ts`
    (green)
  - `npx tsc --noEmit -p tsconfig.json` (clean)
  - `npx vite build` (clean)
  - `npm run ci:changed` (biome, changed-files-only): exit 0, no errors on the touched
    files
- Manual steps: no live Firefox is available in this environment (noted in the issue's
  investigation too), so this could not be re-verified against a real browser. The fix
  is grounded in Firefox's documented, exact-match pointer lock error message and its
  spec requirement that `requestPointerLock()` run inside the original gesture handler,
  plus new tests that pin the synchronous-on-mousedown call on a stubbed Firefox
  `navigator.userAgent` and confirm Chromium's existing deferred path is untouched.

## Screenshots / recordings

N/A. This is a pointer-lock request-timing fix, not a visual change, and the actual
Firefox-broken behavior cannot be captured here (no Firefox available; Chromium never
exhibited the bug so a Chromium screenshot would show nothing different). Saying so
plainly rather than fabricating a screenshot.

---

### Quality

- [x] **The full gate passes.** `npm run gate` fails at the full vitest run stage in
      this environment for reasons unrelated to this change: the shared `node_modules`
      is missing `@testing-library/svelte` (pre-existing per this repo's worktree
      environment notes) and a handful of `server/http` golden/characterization tests
      hang for the same pre-existing reason on a clean checkout without this diff
      applied (verified by stashing this change and re-running the same failing files).
      `tsc`, the build, `ci:changed` biome, and the targeted test files for this change
      (`pointer_lock.test.ts`, `input.test.ts`, `architecture.test.ts`,
      `keybinds.test.ts`, `mobile_controls.test.ts`) are all green.

### Cross-platform

- [x] N/A. Desktop mouse input only; no mobile/touch path is touched
      (`mobile_controls.test.ts` stays green to confirm no regression there).

### Localization (i18n)

### Hygiene

      enabled anywhere.


jgyy:

jgyy:

EnriqueGF:
:) Thanks for the fix, this is a well-factored pure-core with genuinely thorough tests.

Base is `release/v0.26.0` (correct latest release). Verified locally: `tsc --noEmit` clean, `tests/pointer_lock.test.ts` + `tests/input.test.ts` green (90 tests). The seam split is right: the browser sniff and the mousedown-engage decision are both pure and directly Node-tested, and the sync path is correctly scoped to only the camera-look button so the click-to-move button never locks.

One thing to confirm before merge (not a code defect, a live-behavior check I cannot run here):

- On Firefox, `shouldEngagePointerLockOnMouseDown` now requests the lock on every camera-look-button mousedown, and a press that turns out to be a plain click releases it on mouseup. During right-click-to-attack combat that is a lock/release on every click. The original #116 fix deferred the request specifically to avoid Firefox's "has control of your pointer" capture banner spam. Please confirm on a real Firefox session that rapid right-click combat does not re-introduce that banner (or visible cursor flicker); if it does, it may be worth holding the lock briefly across a click rather than releasing immediately.

If that checks out on Firefox, this is good to merge from my side.

FernandoX7 (commented):
Good diagnosis, and it holds up: Gecko does not treat `mousemove` as user input, so `requestPointerLock()` from the drag-threshold call site is denied outside fullscreen, and Chromium's leniency is why this never showed up on Chrome. The history backs you up harder than the PR body does: the call lived in `onMouseDown` at repo genesis (`3387f8b69`) and moved into `onMouseMove` in `c89e2adf6` (2026-06-14), the #116 banner fix. So #1834 is a real regression, just from #116 rather than from the snap-back work, and this PR is restoring the pre-#116 call site on Firefox. That framing matters for the first two findings.

Shape is right: pure helpers in `pointer_lock.ts`, thin consumer in `input.ts`, tests next to them. I ran both test files at your head (90 pass) and `biome ci` on the four touched files (clean). The early lock does not break drag detection, since `dragDistance` accumulates `movementX` rather than client coords.

**blocking**: only one of the two camera-drag buttons gets the synchronous lock. `src/game/input.ts:968` gates rotation on `!this.leftDown && !this.rightDown`, not on a specific button, so either held button starts a camera drag past the threshold (the one exception is the click-to-move button under 280ms, line 973). Classic mode still orbits the camera on a LEFT drag, and Mouse Camera mode still orbits on a RIGHT drag. `shouldEngagePointerLockOnMouseDown` (`src/game/pointer_lock.ts:70`) only fires when `button === cameraLookButton` (`input.ts:915`), so both of those drags keep requesting the lock from `mousemove`, the exact call you have shown Firefox denies. I confirmed it at your head with a Firefox UA: a left mousedown requests no lock, then a left drag rotates the camera and requests the lock from `mousemove`. Default settings, click-to-move off. A Firefox player who orbits with left-drag still has the bug. Cover both buttons, or say why left-drag orbit does not need the lock.

**should-fix**: #116 comes back on Firefox, on ordinary clicks. Every plain right-click on the canvas now takes the lock at mousedown and drops it at mouseup, so the cursor hides and the browser's pointer-lock notice flashes on each one. Right-click is the interact button (loot, mailbox, dungeon doors, targeting: `src/game/interactions.ts:127`), so that is constant during play, and it is what #116 was about. Your own comment at `input.ts:906` still explains that the lock is not requested per-press because that spammed the banner. It gets worse when `clickToMoveButton` (`src/game/settings.ts:52`) is the same button as the look button, which is the default LEFT in Mouse Camera mode: then every click-to-move click takes the lock too, even though the code deliberately refuses to camera-drag on that button under 280ms (`input.ts:973`). Passing `clickMoveMouseButton` into the gate would at least make that an explicit decision. If the flicker is the accepted price of a working camera on Firefox, that is defensible, but state it as a trade-off in the comment and the PR body rather than "the #116 fix is preserved", which is only true of Chromium.

**should-fix**: a fast mouseup can beat the async grant and leave the lock stuck on. `requestPointerLock()` resolves asynchronously, and `onMouseUp` only exits when `document.pointerLockElement === this.canvas` (`input.ts:951`). On a quick click, which is exactly the case this PR newly locks on, the grant can land after the mouseup: nothing releases it then, and the canvas stays pointer-locked with no button held until the next press and release, since the `pointerlockchange` listener (`input.ts:227`) only reacts to lock loss. Cheap guard: on `pointerlockchange`, if the lock was granted and neither button is down, exit immediately. That is also the untested path here, no test exercises the mousedown-lock to mouseup-release cycle the comment promises.

**nit**: the UA sniff duplicates an engine detector we already have. `detectBrowserEngine(ua)` (`src/game/browser_env.ts:57`) already returns `gecko | chromium | webkit | unknown`, is pure, is unit-tested, and documents the UA fossil-ordering traps. `detectBrowserEngine(userAgent).engine === 'gecko'` keeps one source of truth instead of a second regex in `pointer_lock.ts`.

**nit**: unhandled rejection. `this.canvas.requestPointerLock?.()` (`input.ts:922`) drops the returned promise, which current Firefox and Chrome reject (for instance during the lock-out right after the user pressed Esc to leave a lock). The new call site runs on every look-button press, so that becomes routine console noise. `void this.canvas.requestPointerLock?.()?.catch?.(() => {})`.

**nit**: `makeInput` replaces `globalThis.navigator` via `Object.defineProperty` and never restores it (`tests/input.test.ts:20`). `vi.stubGlobal('navigator', ...)` plus `vi.unstubAllGlobals()` in `afterEach` keeps it scoped to the tests that want it.

One process note: the fix is still unverified against the browser it targets. I tried to drive the Firefox installed here (152) through puppeteer and could not get an endpoint out of it either, so I could not repro or confirm live. #1834's acceptance criteria ask for a manual Firefox pass, and the issue also raises the movementX scaling divergence, which this change would not address. Worth one real Firefox run before merge: confirm the lock actually engages on drag, and that rotation speed matches Chrome.

jgyy:
Addressed all findings from @FernandoX7's review:

**blocking (both drag buttons):** `shouldEngagePointerLockOnMouseDown` no longer gates on a single `cameraLookButton`. Camera drag can start on either left or right in either camera mode (`onMouseMove` gates the threshold on `leftDown || rightDown`, not one button), so the synchronous request now fires for whichever of the two was pressed. Added tests for classic-mode left-drag and Mouse-Camera-mode right-drag.

**should-fix (#116 on Firefox for ordinary clicks):** `clickMoveButton` is now passed into the gate and excluded outright (not just under the 280ms threshold, since at mousedown time we cannot yet know whether the press resolves as a click or a drag). Updated the doc comment on `shouldEngagePointerLockOnMouseDown` and the `input.ts` mousedown comment to state the trade-off honestly: #116 is only fully preserved on Chromium; an ordinary Firefox right-click for loot/target/interact still takes and releases the lock, and a drag started on the click-to-move button on Firefox still goes through the denied `mousemove` path. Both are accepted trade-offs, now documented instead of implied away.

**should-fix (mouseup beating the async grant):** `pointerlockchange` now also fires `shouldReleasePointerLock` when the lock was just granted with no button held, exiting immediately rather than leaving the canvas stuck locked. Added a regression test for the mousedown-lock-grant-after-mouseup-release cycle, plus a companion test confirming the lock is kept when the grant lands while the button is still down.

**nits:** `pointer_lock.ts` now uses `detectBrowserEngine(userAgent).engine === 'gecko'` instead of its own regex. The mousedown `requestPointerLock()` call now swallows its rejection (`void ...?.()?.catch?.(() => {})`). `tests/input.test.ts`'s navigator mock now uses `vi.stubGlobal`/`vi.unstubAllGlobals` instead of a leaking `Object.defineProperty`.

`npx vitest run tests/input.test.ts tests/pointer_lock.test.ts` is green (96 tests), `tsc --noEmit` is clean, and `biome check` on the four touched files shows no new errors (only pre-existing `noNonNullAssertion` warnings in the test file style, not introduced here).

One gap I can't close: like you, I don't have a way to drive a real Firefox session here, so the live verification you and EnriqueGF asked for (lock actually engaging on drag, rotation speed matching Chrome) is still outstanding. Everything above is verified at the unit level only.

jgyy:

FernandoX7 (approved):

PR #1841, opened 2026-07-13
fix(sim): require the owner's own hit to tap a rare mob

jgyy:
## Summary

**Pivot from the original approach** (see review discussion below): the tap-predicate
change did not actually close the exploit and opened a new theft vector. This PR now
reverts that predicate and fixes the underlying problem at the credit/drop layer instead.

- Reverted the `tappedById` change: pet damage taps a rare exactly like any other mob
  again (classic-era behavior, unchanged).
- Extended the world-boss damage-contributor tracking (`mob.bossDamagers`,
  `worldBossLootContributors` in `src/sim/world_boss.ts`) to rare templates, not just
  world bosses.
- `rollLoot` (`src/sim/loot/loot_roll.ts`) now accepts an optional `contributors`
  parameter, used only by the `questId` branch: a guaranteed personal quest drop
  (`chance: 1`, e.g. `greyjaw_fang`) is credited to every quest-needing contributor,
  not just whoever currently holds the tap.
- `handleDeath` (`src/sim/combat/damage.ts`) snapshots the rare's contributor roster
  before clearing threat and passes it through, mirroring the existing world-boss
  snapshot right above it.

## Why the pivot

The original fix required the owner's own hit to tap a rare (pet hits no longer
counted). Review (FernandoX7) found two blocking problems:

1. **The farmer still got the fang anyway.** `handleDeath` falls back to the killer
   when a corpse is untapped, and a pet's killer resolves through `ownerId` back to
   the owner. A pet-solo kill left `tappedById === null`, but credit still routed to
   the owner through that fallback, so the predicate change never touched the actual
   exploit path.
2. **It handed every rare to a one-hit tap-snipe.** Because a rare stayed untapped
   while only the pet had hit it, any other player could land one hit in that window
   and steal the whole kill (tap, loot, quest credit) from the legitimate farmer,
   across all 19 `rare: true` templates.

The recommended alternative (credit personal drops by damage contribution, mirroring
how world-boss loot already works) closes the camping monopoly without changing the
classic pet-tap rule, and without introducing the new tap-snipe vector.

## Test plan

- [x] `tests/rare_pet_tap.test.ts`: pet-solo kill still gets the owner the fang
  (contribution-based); a one-hit tap-snipe cannot steal the fang from the actual
  farmer; classic pet-tap rule verified unchanged for both rares and ordinary mobs;
  owner's-own-hit scenario still works.
- [x] `npx vitest run tests/sim.test.ts tests/loot_drops.test.ts tests/loot_roll.test.ts
  tests/loot_ffa.test.ts tests/loot_ffa_sim.test.ts tests/loot_master.test.ts
  tests/loot_master_sim.test.ts tests/loot_messaging_sim.test.ts
  tests/loot_roll_reconcile.test.ts tests/dead_party_loot.test.ts
  tests/heroic_loot_flair.test.ts tests/autoloot.test.ts tests/architecture.test.ts`
  all green.
- [x] `tests/parity`: regenerated two goldens (`affix_mob`, `multi_class_frenzy`) whose
  fixture mobs are rare templates that now populate `bossDamagers`; gate green after
  regen.
- [x] `npx tsc --noEmit`: clean.
- [x] `npx @biomejs/biome check` on changed files: no errors/format diffs (pre-existing
  warnings elsewhere untouched).

Non-blocking note carried over from review: `scripts/greyjaw_pet_tap_shot.mjs` and
`scripts/greyjaw_pet_tap_mobile.mjs` hand-assign `pet.aggroTargetId` rather than
letting real pet AI reach that state, so they don't demonstrate the exploit/fix on
their own. Left as-is; the fix is validated via the sim/vitest tests above, not the
screenshot scripts.


jgyy:

jgyy:

FernandoX7 (commented):
Thanks for the detail here, and for capturing the fight from the real sim rather than mocking it. The underlying report is real: Old Greyjaw is a single camp spawn (`zone1.ts:1115`, `count: 1`) on a fixed 100s respawn, and the fang is a personal quest drop, so tap and kill credit are the only path to it. That part of the diagnosis holds up.

The problem is that `tappedById` is not the field that decides who gets the fang. I ran the real sim at your head commit (`4c0f26279`), then again with only your predicate reverted, so the deltas below are caused by this diff and nothing else.

**blocking: the farmer still gets the fang. The exploit this targets is untouched.**

`handleDeath` falls back to the KILLER when a corpse is untapped (`src/sim/combat/damage.ts:786-793`), and for a pet the killer resolves through `killer.ownerId` back to the owner. So a pet-solo kill with `tappedById === null` still credits the owner. Pet alone kills Old Greyjaw, owner never swings, at your head commit:

```
tappedById:       null      <- what the new predicate produces
lootRecipientIds: [408]     <- the pet's owner
fangPersonalFor:  [408]     <- greyjaw_fang, personal to the owner
ownerKills:       1
```

That is the same fang, the same kill credit, the same XP and the same `slain:old_greyjaw` deed mark the owner gets on base. The only thing that changed is the value of a field nothing on this path reads. The camper keeps farming exactly as before.

The reason this slipped through is that `tests/rare_pet_tap.test.ts` asserts the tap FIELD (`expect(greyjaw.tappedById).toBeNull()`) and never asserts the outcome the PR exists to produce, that a pet-only farmer is denied the fang. Case 4 checks the fang is on the corpse but never asks who may take it. A test that pet-solo-kills the rare and asserts the owner does NOT receive `greyjaw_fang` is the actual acceptance criterion, and it fails on this branch today.

**blocking: it hands every rare to a one-damage tap snipe.**

A rare now stays untapped for as long as only the pet has connected, so any other player can land a single hit in that window and take the tap, and with it the whole kill. Sending the pet in first is the standard hunter and warlock opener, so this is ordinary play. Farmer's pet takes Greyjaw to 20 HP, a passerby pokes it once for 1 damage:

```
tapAfterPet:      null      (on base: 408, the owner, protected by the pet's tap)
tapAfterSnipe:    409       the passerby
lootRecipientIds: [409]
fangPersonalFor:  [409]     the passerby's fang now
farmerKills: 0    sniperKills: 1
```

On base the sniper gets nothing. This is a new theft vector, it applies to all 19 `rare: true` templates (15 of which are elite group content, not just the one non-elite wolf in the rationale), and it lands on the legitimate pet classes rather than on the camper, who simply swings once and is protected again.

**should-fix: an untapped rare corpse leaks loot two ways.**

`corpseLootRights` passes `mob.tappedById` into the shared-rights check (`src/sim/interaction.ts:69-77`) and `hasSharedLootRights` returns true for everyone when the tap is null (`src/sim/loot/loot_ffa.ts:41`), so a rare a pet legitimately solo-kills has its shared pool open to any passerby at once, with no owner-lock and no FFA timer. Separately, `partyLootStrategiesForMob` returns null on a null tap (`src/sim/loot/loot_roll.ts:86-88`), which drops `effectiveItemLootStrategy` to `looter-takes-all`, so such a kill also loses need/greed and master loot.

**should-fix: the rationale reads the pet AI backwards, and the screenshots do not show what they claim.**

The comment says an aggressive pet re-engages "within its anti-AFK window". That window is the opposite of an enabler. `PET_OWNER_IDLE_TICKS` gates the aggressive auto-pull OFF once the owner has been idle 60s, and its own comment says so: "Stops hunters/warlocks parking an aggressive pet to farm XP/loot while AFK". Aggressive is also not the default mode (`entity.ts:140` is `defensive`), and a target-less pet heels to within 3.5yd of its owner, so a pet cannot be parked on a spawn while its owner stands off at range.

That matters for the evidence, because `scripts/greyjaw_pet_tap_shot.mjs:103-108` teleports the owner to `mob.pos.x + 35`, teleports the pet to `mob.pos.x + 2`, and then hand-assigns `pet.aggroTargetId = mob.id`. The pet AI cannot reach that state on its own (the aggressive pull needs the pet within 18yd of the mob and an owner who acted in the last 60s). So the shots prove pet damage does not set the tap, which is not in dispute, but they do not demonstrate the exploit or the fix.

The genuine zero-owner-input path is the one the PR does not name: `pullNearbyMobs` runs for every pet mode, so a pet standing at the camp pulls the rare onto itself, and `petPickTarget`'s defend branch (`engagingUs`) has no idle gate at all. That path survives this change untouched, and thanks to the killer fallback it still pays the owner.

**Where I would take this**

Tap is only the loot lock. Credit and the personal quest drop come from the tapper-or-killer chain, so denying the pet a tap denies nothing. Closing the killer fallback too would just mean an untapped rare pays nobody, and the camper's pet would vaporize Greyjaw every respawn with the fang dropping for no one, which is worse than what we have now.

Two levers actually bite, and I would rather pick one deliberately than acquire a rule change as a side effect of the tap predicate:

1. The credit and drop side. A `chance: 1` quest drop could go personal to every quest-needing player who contributed damage, which is what world bosses already do (`bossDamagers` into `worldBossLootContributors`, `damage.ts:872-877`). That dissolves the monopoly outright without touching classic tap rules, since a camper can no longer deny anyone else the fang.
2. The content side: a second camp, a shorter rare respawn, or an alternate objective for `q_greyjaw`.

Also worth saying plainly: pet damage taps in the classic era, so carving rares out of that is a gameplay rule change rather than a bug fix, and it needs to be argued as one.

Notes, not blocking:

- The red "Screenshots of changes" check is the known concurrency-cancel race in this repo, not your change.
- `scripts/greyjaw_pet_tap_mobile.mjs:20,26` bakes in an absolute `/tmp/claude-1000/-home-jegoh-...` user-data-dir and HOME, so it only runs on the machine that wrote it. Other scripts here do the same, so this is not new debt, but `os.tmpdir()` costs nothing if these are staying.

jgyy:
Pushed a pivot based on this review, thank you for the thorough writeup.

Reverted the `tappedById` predicate change entirely (pet damage taps a rare exactly
like any other mob again, classic behavior unchanged) and implemented the suggested
alternative: extended the world-boss damage-contributor tracking (`mob.bossDamagers` /
`worldBossLootContributors`) to rare templates, and threaded that contributor roster
into `rollLoot`'s `questId` branch so a guaranteed personal quest drop
(`greyjaw_fang`, `chance: 1`) is credited to every quest-needing contributor, not just
whoever holds the tap at death.

This closes both points you raised:
- A pet-solo camping kill still gets the owner the fang (contribution-based credit,
  not tap-based), so the camping monopoly is gone without denying the camper their
  own farmed reward.
- A one-hit tap-snipe from a passerby can no longer steal the fang from the player
  who actually did the work, since the sniper holding the tap does not put them in
  the contributor set the farmer's pet built up.

New/updated test coverage in `tests/rare_pet_tap.test.ts` covers both scenarios plus
the unchanged classic pet-tap behavior for rares and ordinary mobs. Two parity goldens
regenerated (rare fixture mobs now populate `bossDamagers`). Left the screenshot
scripts as-is per your non-blocking note; validated via vitest instead.

FernandoX7 (approved):

jgyy:

PR #1845, opened 2026-07-13
feat(metrics): add safe player business analytics

sf-chris:
## Summary

Adds a production-safe MVP for player business analytics without a replica or heavy recurring scans.

- Collects compact, indexed lifecycle facts prospectively at character creation, world entry, and session close. There is no boot-time backfill and no work in the game tick.
- Exposes daily acquisition, first-character conversion, first-world-entry conversion, new and returning DAU, average daily playtime, first-session duration and level activation, and D1/D7/D30 retention.
- Refreshes one bounded read-only snapshot every 15 minutes with a 250 ms lock timeout and 2 second statement timeout. Prometheus scrapes only the last good in-memory snapshot and never query Postgres.
- Removes the client FPS collector from the business metrics path.
- Adds unit coverage and opt-in real-Postgres integration coverage for the exact lifecycle and snapshot SQL.

Companion monitoring PR: [woc-deploy#20](https://github.com/levy-street/woc-deploy/pull/20).

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `TEST_DATABASE_URL=postgres://eastbrook:test@127.0.0.1:5433/eastbrook npx vitest run tests/player_metrics_db_integration.test.ts tests/player_metrics_db.test.ts tests/server/http/business_metrics.test.ts tests/server/http/periodic_collector.test.ts tests/character_db.test.ts tests/schema_wiring.test.ts tests/game_sessions.test.ts --maxWorkers=1` (81 passed)
  - `npx tsc --noEmit`
  - `npm run ci:changed`
  - Mandatory pre-push QA floor (53 passed, 3 skipped; typecheck, lint, and copy rules green)
- Manual steps:
  - Exercised character creation, first and returning session opens, session close/day splitting, first-session level capture, and the bounded snapshot against PostgreSQL 16.
  - Verified failed refreshes retain the last good Prometheus snapshot and concurrent refreshes coalesce.

---

### Quality

- [ ] **The full gate passes.** `npm run gate` is green. I added or updated decisive

### Cross-platform

- [ ] ~~Tested on desktop and mobile. N/A: no player UI changes.~~
- [ ] ~~Accessible. N/A: no UI changes.~~

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
This is a well-built slice of analytics plumbing. The design holds up everywhere I pushed on it: facts are collected prospectively with no boot backfill, the lifecycle writes are single atomic statements at real lifecycle boundaries, the snapshot runs read-only with fail-fast timeouts and scrapes never touch Postgres, the new tables carry no PII (ip and user agent stay in play_sessions), the DDL is additive, idempotent, and ordered correctly for a first-ever boot, and the veteran-seeding path (first_play_at backfilled from the earliest historical session at next login) is a nice self-healing touch. I also checked the sneaky stuff and it is clean: prior_session correctly cannot see the row the opened CTE inserts, the unreferenced data-modifying CTEs still run, the UNIQUE (realm, first_session_id) constraint is unviolable from any reachable path, every mapped column is cast to int or double precision so node-postgres never hands the mapper a bigint string, and the abort wrapper has no leak or double-release path.

Findings:

- should-fix, server/player_metrics_db.ts:207 and 301: openPlayerSession writes player_account_facts then player_activity_daily; closePlayerSession writes them in the opposite order. Both are fired without awaiting from game.ts (leave at 2683, join at 2462), and the repo already documents that a grace-expiry leave can race a reconnect, so the two statements can hit the same account-plus-day rows concurrently and ABBA-deadlock. Blast radius is bounded (the victim rolls back, the lost close is reaped by orphan cleanup at zero duration, and nothing player-facing blocks), but each event stalls a pooled connection for about the 1s deadlock_timeout, which a mass-reconnect blip can multiply. Reordering closePlayerSession to update facts before the activity insert removes the textual ABBA; worth a comment that CTE lock order is not formally guaranteed, so this shrinks the window rather than proving it closed.

- should-fix, server/player_metrics_db.ts:21: UNIQUE (realm, first_session_id) is realm-leading, so it cannot serve the first_session_id ON DELETE SET NULL cascade from play_sessions. Nothing deletes play_sessions today, but pruning it is a likely next move given current DB pressure, and at that point every pruned row that is some account's first session triggers a sequential scan of player_account_facts. Swapping the constraint to UNIQUE (first_session_id, realm) keeps identical uniqueness and serves the cascade.

- should-fix, server/player_metrics_db.ts:200: the prior_session CTE runs on every world entry, and with only the play_sessions(account_id) index it fetches and top-1-sorts all of an account's session rows per login, forever, even though its result only matters until the facts row is complete. The cost grows with an account's own session count (long-lived dailies, rapid-relog accounts). Either add a composite (account_id, started_at) index or skip the lookup once first_session_id is already set.

- should-fix, tests/player_metrics_db.test.ts:147: mapBusinessDay is pinned on only 4 of its 13 fields in the default suite, and the gauge tests feed a pre-built snapshot fixture, so a swapped column mapping (say median_seconds landing in avgPlaytimeSecondsLevel20) ships green unless someone runs the opt-in integration suite. The fixture already has distinct values per field; extending the toMatchObject to all of them closes this cheaply.

- nit, server/game.ts:2911 and 2354: two metricsMaxLevel arms are untested: endAllPlaySessions passing the level at shutdown, and the join-time seed from the sim entity's real level. Every current test joins a level-1 character, so a regression that always seeds 1 would pass and quietly flatten the level_20 playtime segment for returning players. Both are cheap mirrors of the existing leave() test.

- nit, server/http/CLAUDE.md:64: this file still says business_metrics.ts "reuses admin_db.overviewCounts" and lists client_perf_metrics.ts as a wired half of the exporter; this PR changed both. Worth updating in the same change so the seam doc matches the code.

- nit, server/db.ts:2316: createCharacter (non-capped) has no callers in the tree; instrumenting it keeps the two creation paths consistent, which is fine, but it is now an untested transaction wrapper on dead code. A follow-up could just delete it.

Notes, not findings: the old woc_accounts_total, woc_signups_total, woc_active_sessions, woc_avg_playtime_seconds, woc_peak_online, and woc_client_* series all disappear from /metrics, so this should land together with woc-deploy#20 or the dashboards go dark in between. The integration suite being gated on TEST_DATABASE_URL matches the existing client_perf_metrics_db precedent, so no complaint, just be aware a default CI run only pins SQL strings and params for the lifecycle statements. And accounts_created is global (accounts have no realm) while same_day_first_plays is realm-scoped, so first_world_entry_rate mixes scopes; exact on today's single realm, but worth a comment for the process-per-realm future.

Rubsey:
Resolved the conflicts with release/v0.26.0 (the #1930 stability hardening) and pushed three commits.

**Merge (18a4b6049).** The only textual conflict was in `ensureSchema` (server/db.ts): this branch named the advisory-lock key (`SCHEMA_ADVISORY_LOCK_KEY`) while #1930 added `SET LOCAL statement_timeout = 0` before taking the lock. Resolved by keeping both: the base's SET LOCAL and comment, the lock taken with the named constant.

I also ran a three-way audit over the six files both sides changed (db.ts, game.ts, main.ts, http/CLAUDE.md, game_sessions.test.ts, schema_wiring.test.ts). All six merged as exact unions with coherent semantics; in particular every #1930 teardown path (lease-fence kick, takeOverCharacter, delayed keepalive terminations) funnels through leave(), so the session close always carries `metricsMaxLevel`, and the linkdead resume arm reuses the same ClientSession so the open row and max level survive reconnects.

The audit surfaced two real cross-merge interactions that neither branch could see alone, fixed as follow-up commits:

**5de4b935f: the post-commit index build lost the boot timeout exemption.** #1930's `SET LOCAL statement_timeout = 0` reverts at COMMIT, and this branch's post-commit `CREATE INDEX CONCURRENTLY` migration then ran with any operator-set database/role-level statement_timeout re-applied (the exact case #1930's override exists to defeat; the advisory-lock wait is also subject to it). Now re-disabled session-wide after COMMIT on the dedicated boot client, pinned in schema_wiring.

**173f036ee: a killed CONCURRENTLY build stranded a permanent INVALID index.** The migration runs before `server.listen`, and #1930's compose healthcheck plus deploy watchdog give boot roughly a 105 to 120 second budget before restarting the container. A build killed mid-CONCURRENTLY leaves the index INVALID, and `IF NOT EXISTS` treats an INVALID index as existing, so every later boot would skip the rebuild forever while play_sessions writes kept maintaining the dead index. Boot now drops such a carcass (CONCURRENTLY, so peer realms' writes never stall) before the create. Pinned by both schema_wiring arms and a real-Postgres integration test that reproduces an actually failed CONCURRENTLY build.

**Verification.** `tsc --noEmit` clean; full Vitest suite green (14k+ tests; the one excluded file, tests/deploy_watchdog.test.ts, fails on macOS because coreutils `timeout` is missing locally, and it is byte-identical to the base tip, so it is environmental, not from this merge); the opt-in real-Postgres integration suite passes including the new carcass test; all five builds green.

**Two non-blocking operator notes from the audit**, both consequences of this branch's deliberate design riding a merge with #1930:
- The /metrics contract changes when this ships: the old `woc_accounts_total`/`woc_signups_total`/`woc_characters_total`/`woc_active_sessions`/`woc_avg_playtime_seconds`/`woc_peak_online` and the `woc_client_*` gauges go blank, replaced by the `woc_player_*` family at a 15 minute cadence. Anything scraping the old names needs the woc-deploy#20 companion landed in step.
- `server/http/client_perf_metrics.ts` and `server/client_perf_metrics_db.ts` are now production-dead but still in-tree with live tests and stale wiring comments (also `server/CLAUDE.md:65` still says the client-perf aggregates sit behind /metrics, and the `admin_db.ts` overviewCounts comment still cites the PeriodicCollector as its driver). Fine to leave for a follow-up, just flagging so the docs do not mislead the next reader.

PR #1847, opened 2026-07-13
fix(sim): free players wedged in concave coastline cliff pockets

jgyy:
## Summary

Players could get permanently stuck against a steep cliff face near a coastline (e.g. the Veiled Hollow's coastal notches) and be unable to move again. Reported with a screenshot of a character wedged into a rock wall while the open-sea swim-fatigue warning spammed the chat log.

## Root cause

```mermaid
flowchart TD
    A["Player near a concave coastline notch\n(two steep faces meeting, e.g. a cliff cove)"] --> B["terrainWallStandoff ring-samples 8 points\nat body radius, nudges away from steep ones"]
    B --> C{"Push capped at ONE body radius:\nenough to clear a single flat wall,\nnot always enough for a concave pocket"}
    C -->|"Old behavior"| D["stepPlayerMotion only commits the push\nif the result is FULLY walkable (<= climb limit)"]
    D --> E["Still steep -> push silently discarded\nposition unchanged this tick"]
    E --> F["Next tick: SAME position, SAME inputs,\nSAME computation -> frozen forever"]
    C -->|"Fixed behavior"| G["terrainWallStandoff iterates up to 3 passes\n(mirrors colliders.ts's 3-pass prop resolver)"]
    G --> H["stepPlayerMotion commits the push if it clears\nthe limit OR strictly improves on current steepness"]
    H --> I["Every one of the sim's 20 ticks/sec\nmakes guaranteed progress out of the pocket"]
    I --> J["Player escapes within a bounded number of ticks"]
```

`terrainWallStandoff` (`src/sim/world.ts`) pushes a player's body out of nearby steep terrain in a single ring-sample-and-nudge pass, capped at one body radius. In a concave pocket (a coastline notch where two rock faces meet near open water), one pass isn't always enough to clear the wall. `stepPlayerMotion` (`src/sim/player_motion.ts`) only ever committed that pushed position once it read as fully walkable — so an imperfect nudge was silently discarded, every tick, forever. The exact same computation on the exact same (unmoved) position repeats indefinitely: a true permanent stuck, matching the report (the open-sea swim-fatigue warning kept firing independently because the wedge happened to sit near that boundary too — that system isn't the cause).

## Fix

- `terrainWallStandoff` now iterates its push up to 3 times per call, mirroring the existing 3-pass prop/OBB resolver in `colliders.ts`, so it converges further out of concave pockets instead of stopping after one pass.
- `stepPlayerMotion`'s acceptance gate now commits the result once it either clears the climb limit outright, or strictly improves on the player's current steepness. This guarantees the sim's 20 ticks/sec make monotonic progress out of a wedge instead of discarding every partial improvement, which is what caused the permanent freeze.

## How was this tested?

- Added a regression test (`tests/veiled_hollow.test.ts`) that plants a player directly in four real Veiled Hollow coastline notch coordinates and asserts they reach walkable ground within 60 ticks, with no movement input (isolating the standoff/slope-gate recovery itself).
- Verified the test is decisive by reverting the fix and re-running: the player then freezes at the exact planted coordinate for the full 400-tick probe at at least one of the four points. Restoring the fix, all four escape within a tick.
- `npm run gate`: full green (tsc, i18n, builds, changed-file biome, full Vitest suite — 13010 passed, plus 9 pre-existing unrelated failures in `tests/deeds_window_focus.test.ts` caused by a `localStorage` polyfill gap in this sandbox's Node/Vitest version, confirmed pre-existing on a clean checkout of the base branch).
- Manual: booted offline play, planted a character in a live coastline notch and single-stepped `sim.tick()` directly to confirm the same escape behavior in the actual client bundle, not just the test harness.

## Screenshots

Illustrative context (Veiled Hollow coastline near the Starfall headland, the class of terrain this bug occurs on):

Veiled Hollow coastline

A true live before/after of the exact frozen-vs-escapes contrast wasn't practical to capture cleanly in this environment (every coastline candidate I found either escapes via the unrelated, already-working downhill-slide mechanic or ends up swimming, both of which sidestep this exact code path) — the regression test above is the decisive proof, verified by an actual revert-and-rerun.

## Type of change

- [x] Bug fix

## Checklist

- [x] The full gate passes (see above; pre-existing unrelated failures noted).


jgyy:
Closing to retarget: rebuilding this fix against the latest release branch instead of the WIP feature/procedural-dungeons branch, since the underlying bug and fix are generic to the shared movement kernel, not specific to the Veiled Hollow content.

PR #1848, opened 2026-07-13
fix(sim): harden terrain wall standoff against concave pockets

jgyy:
## Summary

Players have reported getting permanently stuck against a steep cliff face and being unable to move again, with a screenshot showing a character wedged into a rock wall.

## Root cause

```mermaid
flowchart TD
    A["Player's body overlaps a concave wall pocket\n(two steep faces meeting, e.g. a rim/ridge corner)"] --> B["terrainWallStandoff ring-samples 8 points\nat body radius, nudges away from steep ones"]
    B --> C{"Push capped at ONE body radius:\nenough to clear a single flat wall,\nnot always enough for a concave pocket"}
    C -->|"Old behavior"| D["stepPlayerMotion only commits the push\nif the result is FULLY walkable (<= climb limit)"]
    D --> E["Still steep -> push silently discarded\nposition unchanged this call"]
    C -->|"Fixed behavior"| F["terrainWallStandoff iterates up to 3 passes\n(mirrors colliders.ts's 3-pass prop resolver)"]
    F --> G["stepPlayerMotion commits the push if it clears\nthe limit OR strictly improves on current steepness"]
    G --> H["Guaranteed monotonic progress\ninstead of a silently discarded push"]
```

`terrainWallStandoff` (`src/sim/world.ts`) pushes a player's body out of nearby steep terrain in a single ring-sample-and-nudge pass, capped at one body radius. In a concave pocket (two walls meeting, most reliably reproduced at this world's rim corners where the x-rim and z-rim walls press in from two directions at once) one pass is not always enough to clear the wall. `stepPlayerMotion` (`src/sim/player_motion.ts`) only ever committed the pushed position once it read as fully walkable, silently discarding every partial improvement.

## Fix

- `terrainWallStandoff` now iterates its push up to 3 times per call, mirroring the existing 3-pass prop/OBB resolver in `colliders.ts`, converging further out of concave pockets instead of stopping after one pass.
- `stepPlayerMotion`'s acceptance gate now commits the result once it either clears the climb limit outright, or strictly improves on the player's current steepness, guaranteeing monotonic progress on repeated calls instead of discarding every partial improvement.

**Important scope note found during testing:** on this branch's current terrain (simple gaussian ridge/rim walls), I scanned the entire world at 1-yard resolution for the exact "the downhill-slide fallback also fails" precondition that would cause a literal permanent freeze, and found none: the slide mechanism here always has a valid escape gradient wherever a single standoff pass would otherwise fall short. So the practical "player is permanently wedged and cannot move at all" symptom is not currently reproducible on this branch's terrain. This PR is defensive hardening of the same code path that I independently proved *does* cause that exact permanent freeze on a branch with more complex organic coastline terrain (concave coastline notches, where the slide gradient can cancel to near-zero). Landing the hardening here now means it's already in place once that terrain style merges, and it strictly improves outcomes in genuine concave pockets on this terrain too (see below).

## How was this tested?

- Added `tests/terrain_wall_standoff.test.ts`: pins a known concave pocket near a world-rim corner where a single pass leaves the player at 3.66x the climb limit and iteration brings it to 0.37x (both readings are independent pins against the actual terrain, not a self-comparison); sweeps a band near all four rim corners asserting iteration is never worse than a single pass; and checks idempotence and the no-op case on open ground.
- Verified decisiveness: reverted the fix (`WALL_STANDOFF_ITERATIONS = 1`) and confirmed 2 of 4 assertions fail; restored the fix and confirmed all pass.
- `npm run gate`: full pass (tsc, i18n, builds, changed-file biome, full Vitest suite). 13 pre-existing failures in the full-suite run are unrelated and confirmed pre-existing on a clean checkout of the base branch: `tests/deeds_window_focus.test.ts` (a `localStorage` polyfill gap in this sandbox's Node/Vitest version), `tests/server/http/parity.test.ts` (no live Postgres DB in this sandbox), and `tests/fiesta.test.ts` / `tests/server/http/characterization.test.ts` (both pass in isolation; flaky only under full-suite worker contention, reproduced identically on the unmodified base branch).
- Manual: booted offline play, planted a character at the pinned rim-corner coordinate, and confirmed the client renders the expected steep mountainside terrain.

## Screenshots

|  | Before (old acceptance gate) | After (fixed acceptance gate) |
|---|---|---|
| In-game | before | after |

Both shots were captured offline at the exact pinned repro spot from
`tests/player_motion.test.ts`'s new acceptance-gate test: production seed 20061, `x:-170,
z:-24.5`, a concave rim-corner pocket. To isolate the acceptance-gate change itself (the part
this PR actually changes) from the surrounding `stepPlayerMotion` slide-fallback logic (which,
per the scope note above, always finds an escape gradient on this branch's terrain and so
masks the bug end-to-end in real gameplay), each shot calls `terrainWallStandoff` directly and
applies that commit's acceptance-gate formula, then teleports the live offline character to
the resulting position:

- **Before** (old gate, `steepStand <= climbLimit` only): the single push still reads at
  steepness 4.08 (well over the 1.5 climb limit), so the old gate never commits it. The
  character is rendered exactly at the raw pin, wedged sideways into the cliff face.
- **After** (new gate, `steepStand <= climbLimit OR steepStand <= currentSteepness`): the same
  push (now iterated up to 3 passes) reaches steepness 1.57, a strict improvement over the
  starting 4.08, so the new gate commits it. The character is rendered relocated off the wall
  face, closer to open ground.

Caveat: because the real `stepPlayerMotion` slide fallback papers over this exact scenario on
this branch's terrain (see the scope note above), the two screenshots do not show a
dramatically different silhouette at a glance, since the position delta is only about 1.1
yards along a very large cliff face; the authoritative evidence is the steepness numbers above
and the passing `tests/terrain_wall_standoff.test.ts` / `tests/player_motion.test.ts` pins,
not visual inspection alone.

## Type of change

- [x] Bug fix (hardening; see scope note above)

## Checklist

- [x] The full gate passes (see above; pre-existing unrelated failures noted).


jgyy:

jgyy:

EnriqueGF:
:) Thanks for the fix, the iterated standoff and the concave-pocket reasoning are solid.

Base is `release/v0.26.0` (correct latest release). Verified locally and with an architecture pass: `tsc --noEmit` clean, `tests/architecture.test.ts` green (no DOM/Three/`Math.random`/`Date.now`/`performance.now` introduced), `tests/parity` green with the draw-order digest unchanged, and `tests/terrain_wall_standoff.test.ts` green. Determinism, sim purity, and tri-host parity all hold: the kernel is pure `(x,z,seed)` and runs identically offline, on the server, and in the RL env.

Two things to address before this is production-ready:

1. (Should fix) Misleading comment on a load-bearing gate. In `src/sim/player_motion.ts` the comment says the nudge commits when it "strictly reduces steepness versus the current spot", but the code is `standSteep <= terrainSteepnessAt(p.pos.x, p.pos.z, deps.seed)` (equal allowed), and the `<=` is actually required, not loose. `terrainSteepnessAt` quantizes to integer cells (`Math.round`), while a single tick's push is capped at one body radius (< 1 cell), so a sub-cell nudge usually lands in the same cell and reads exactly equal. Changing `<=` to `<` to match the "strictly reduces" wording would reintroduce the permanent-wedge freeze this PR fixes. Please correct the comment to say the gate accepts a push that does not increase (cell-quantized) steepness, and note the sub-cell/quantization reason, so a later cleanup does not silently regress it.

2. (Should fix) The actual behavior change has no direct test. The new test file thoroughly exercises `terrainWallStandoff` in `world.ts` (convergence, no-regression sweep, idempotence, open-ground no-op), but the riskier `standSteep <= currentSteep` commit gate in `stepPlayerMotion` is not exercised by any test, and the file header itself notes the wedged symptom "isn't reproducible here today." Per the repo rule that sim behavior changes get a test covering the new path, please add a `stepPlayerMotion`-level test that pins: a player whose center cell is steeper than `MAX_CLIMB_SLOPE` is nudged toward equal-or-lower steepness across ticks, and a normally-standing player near a wall is unaffected.

For the record, two non-blocking notes: the regression scope is narrower than "all players near walls" (the OR-branch only adds commits when the current center cell already exceeds `MAX_CLIMB_SLOPE`, i.e. players already clipped onto steep ground), and there is a theoretical deterministic A<->B jitter at a symmetric pocket that the internal fixed-point break and 3-iteration cap should damp; worth a multi-tick trace only if a jitter report ever surfaces.

Happy to re-review once the comment and the gate test land.

FernandoX7 (commented):
Good diagnosis, and the iteration is the right shape: it mirrors `resolveAgainst` in `colliders.ts:391` closely, same 3 passes and the same early break once a pass stops moving the point. The change is pure and deterministic, it stays in the movement kernel so all three hosts apply it identically, and the new test pins real terrain readings instead of comparing the function to itself. The scope note in the description is the right call too. Saying plainly that the freeze does not reproduce on this branch's terrain is far more useful than overselling it.

I checked the relaxed gate before believing it. The second clause can only fire when `terrainSteepnessAt(p.pos)` is already above `MAX_CLIMB_SLOPE` (it needs `standSteep <= current` while `standSteep > MAX_CLIMB_SLOPE`), so a player on walkable ground can never be shoved onto a wall by it. Whoever it fires for is already on unwalkable ground and already in the `steepGround` slide branch at `player_motion.ts:142`. The design holds up. The findings below are about what the code says versus what it does, and about the half of the fix that nothing tests.

**should-fix** `src/sim/player_motion.ts:317`
The comment above this block still says the nudge lands "never onto ground steeper than the climb limit (a rare terrace corner: a tick's clip beats being shoved onto a wall)". The new gate deliberately reverses that rule, so the comment now contradicts the code five lines below it. It has to go, or the next reader restores the old invariant as a "bug fix".

**should-fix** `src/sim/player_motion.ts:367`
The `player_motion.ts` half of the fix has no test. I reverted just this gate (keeping the `world.ts` iteration) and ran `terrain_wall_standoff`, `player_motion`, `fixes` and `sim`: 194/194 still pass. The revert check in the description (`WALL_STANDOFF_ITERATIONS = 1`) only exercises the `world.ts` half, and the pinned pocket in the new test never reaches this gate: the iterated standoff lands at steepness 0.37, under the climb limit, so the *original* gate already accepted it. The clause is not dead code (sweeping the world, it commits about 92k pushes the old gate discarded, all from already-unwalkable ground), which is exactly why it should be pinned. `(-170, -24.5)` is one position that exercises it: steepness 4.08, standoff pushes to 1.56, old gate discards, new gate commits.

**nit** `src/sim/player_motion.ts:357`
The comment says "strictly reduces" and "strictly improves", but the code is `<=`. Worth saying in the comment that the non-strict compare is deliberate, because it is load-bearing. `terrainSteepnessAt` rounds to 1-yard cells, and one pass moves at most one body radius (0.5yd), so a push routinely lands back in the same steepness cell and reads exactly equal. At your own pinned pocket the reading is 3.66 both before and after a single pass. Tighten this to `<` and the equal-cell case stops committing, which is the wedge case the gate exists for. The code is right and the comment describes a gate that would not work.

**nit** `tests/fixes.test.ts:317`
That end-to-end Sim pin says a grounded tick moves the player "never more than a body radius", and it is no longer true of the function: three passes can return a point up to 3R away, and 1.500yd is the exact world-wide max I measured. It still passes honestly, because its cell converges in one pass, and the `:281` sweep still passes because it filters starts on the exact `terrainSteepness`, under which the bound does hold everywhere. But `stepPlayerMotion` gates on the rounded `terrainSteepnessAt`, and from ground the game treats as standable by that measure the push exceeds one radius at roughly 11k world points. So the standoff can now displace a player 1.5yd in a single tick, against a 0.35yd normal run step. Worth updating those bounds and their comments, or clamping the accumulated push, so the contract is not understated.

**nit** `tests/terrain_wall_standoff.test.ts:74`
Same quantization: `expect(steepOnce).toBeGreaterThan(3.5)` is pinning the steepness of the unmoved start, not what a single pass achieves, because the 0.5yd push does not leave the 1-yard cell. Fine as a regression net, but the comment above it claims more than the number shows. Also 3.66 is the steepness and the limit is 1.5, so it is 2.4x the limit, not "3.66x" as the description and the comment both say.

**nit** `tests/terrain_wall_standoff.test.ts:17`
Points at `tests/veiled_hollow.test.ts` "on a branch with the newer organic coastline terrain". That file does not exist here, so it reads as a dangling reference.

**nit** `tests/terrain_wall_standoff.test.ts:29`
`singlePass` hand-copies the production pass, including the hardcoded 8 ring samples. The comment defends that as deliberate, but it cuts both ways: if `terrainWallStandoffPass` ever changes, the "never worse than a single pass" sweep quietly starts comparing against a fossil. Exporting the pass helper for the test keeps the net honest.

**nit** The description embeds the screenshot from `raw.githubusercontent.com` on your fork branch, which breaks once that branch goes away. The image is committed, so link the in-repo path. For a logic-only sim fix I would just drop the 87KB binary, since the description already says it shows nothing about the change.

Perf is a non-issue, in case it comes up: on open ground the first pass returns the input and the loop breaks, so cost is unchanged (about 3us per call), and a full 3-pass call in a pocket is about 9us.

FernandoX7:
Two follow-ups on the gate, one of which corrects the steer I gave above.

I said to keep the `<=` and just reword the comment. That was half right. Sweeping the overworld with the shipped gate, the non-strict compare against the cell-quantized `terrainSteepnessAt` commits about 3.2k pushes (roughly 3.0k of them away from the rim) where the exact, unquantized `terrainSteepness` actually gets **worse**, for example `(-180, -14)`, true steepness 2.14 to 2.61, while both cells read 2.14. So the gate is not merely non-strict, it can commit a push that genuinely worsens the spot. The cleaner fix is to compare on the exact `terrainSteepness` with a strict `<`, rather than the rounded memo. That gives you the monotonic property the comment actually claims, and it dodges the equal-cell freeze I warned about, because the exact sampler can see a 0.5yd push that the 1-yard cell cannot. It also leaves your pinned pocket untouched, since the iterated standoff lands at 0.37 there and the first clause accepts it either way.

Second, `player_motion.ts:363` and `world.ts:673` claim the change guarantees "eventual escape instead of a permanent wedge", and that is not established. Iterating the shipped accept map, 1,112 starts enter a repeating standoff cycle (position A to B and back to A) while still reading above the climb limit, 745 of them in the interior, and a larger set reaches a fixed point that is still over the limit. What actually rescues those players is the pre-existing downhill slide, which is exactly what the PR body concedes. Worth softening both comments to say the iteration converges further out of a pocket, not that it guarantees escape.

Neither is a new player-facing freeze: both are confined to ground you can only stand on while already in the `steepGround` slide state, so the slide keeps you moving regardless. They just mean the comments promise more than the code delivers.

jgyy:
Thanks for the thorough review, addressed all of it:

- **should-fix 1**: rewrote the comment above the wall-standoff block in `player_motion.ts` so it no longer promises "never onto ground steeper than the climb limit" (the new gate can do exactly that on a strict improvement); points at the acceptance-gate comment for the actual behavior.
- **should-fix 2**: added a decisive test in `tests/player_motion.test.ts` pinning the `stepPlayerMotion` acceptance gate itself (distinct from `terrainWallStandoff`'s iteration) at `(-170, -24.5)`, production seed 20061: steepness 4.08 at the pin, the standoff push resolves to 1.56 (above the 1.5 climb limit but a strict improvement), and the test drives the real `stepPlayerMotion` (stubbing out only the unrelated downhill-slide `resolveMove` call) to confirm the player actually lands at the standoff point instead of staying frozen.
- **nit 1**: fixed the "strictly reduces/improves" comment to describe the actual `<=` compare and why it's deliberate (1-yard steepness quantization plus a 0.5yd max single-pass push means the wedge case often reads as exactly equal, not strictly less).
- **nit 2**: `tests/fixes.test.ts`'s end-to-end pin now asserts the real bound, 3x body radius (1.5yd) from the 3-pass iteration, not "never more than a body radius."
- **nit 3**: clarified that `steepOnce` in the concave-pocket pin is effectively the unmoved starting steepness (a 1-yard quantization artifact, since the single pass's push doesn't leave the cell there), and fixed the "3x the climb limit" claim to the accurate ~2.4x (3.66 / 1.5).
- **nit 4**: dropped the dangling reference to `tests/veiled_hollow.test.ts`, which doesn't exist in this repo.
- **nit 5**: exported `terrainWallStandoffPass` from `world.ts` and swapped the test's hand-copied re-implementation for the real helper, so the "never worse than a single pass" sweep can't silently drift from production.
- **nit 6**: removed the committed screenshot and the PR description's external `raw.githubusercontent.com` embed (it showed identical terrain before/after, nothing about the change itself) rather than re-pointing it at an in-repo path.

`npx vitest run tests/terrain_wall_standoff.test.ts tests/player_motion.test.ts tests/fixes.test.ts tests/sim.test.ts` and `npx tsc --noEmit` both pass.

FernandoX7 (approved):

jgyy:

PR #1851, opened 2026-07-13
feat(card-adept): add the Card Adept class and a 1v1 Card Duel arena

jgyy:
## Summary

Adds a 10th player class, the **Card Adept** (`card_adept`), whose combat kit is entirely card-based, plus a dedicated 1v1 PvP **Card Duel** queue restricted to that class.

The Card Adept is a Focus (energy-style) cloth caster. Its abilities (`ca_*`) are spell-school effect templates that a shuffled deck of ~20 cards references. A deterministic deck machine (`src/sim/card_hand.ts`) draws an opening hand at combat start and redraws on the 20 Hz combat cadence; playing a card spends Focus and triggers the referenced ability through the existing effect pipeline. Everything resolves inside the one shared `Sim`, so behavior is identical offline, on the server, and headless.

### What landed

- **New class + data:** `card_adept` in `PlayerClass`/`ALL_CLASSES`; `CLASSES.card_adept` (Focus resource, modeled on rogue energy); 8 `ca_*` ability templates; `src/sim/content/cards.ts` (`CARDS`, ~20 records referencing the abilities).
- **Deck system:** `src/sim/card_hand.ts`, a pure deterministic `SimContext`-guarded deck/hand/discard machine (Fisher-Yates via `Rng`, never `Math.random`); per-entity state on `PlayerMeta.cardHand`, driven from the per-player tick, guarded to Card Adepts so worlds without one draw zero extra rng (parity-safe).
- **IWorld seam:** read surface (`cardHandIds`, `cardDeckCount`, `cardDiscardCount`) + `playCard` command on the combat facet, plus `queueCardDuel`/`cardDuelInfo` on the duel-arena facet, implemented in both `Sim` and `ClientWorld`, with the `play_card` / `card_duel_queue` wire commands and the W0b/W0c pins updated.
- **PvP:** `src/sim/social/card_duel_queue.ts`, a pure FIFO matchmaking core restricted to Card Adepts; the Sim pairs waiting players and starts a duel in the arena.
- **UI core:** `src/ui/card_hand_view.ts`, a pure Node-tested view-core (playable/unplayable, deck/discard counts), registered in `UI_PURE_CORES`.
- **Character model:** reuses the mage cloth rig with a distinct violet tint (`player_card_adept`); no new meshes.
- **Card art:** procedural icon recipes in `src/ui/icons.ts` (the existing canvas icon generator). See "Skipped / simplified".
- **10-class sweep:** every `Record<PlayerClass,...>` site, the two `index.html` rosters, class i18n (names, roles, lore, ability names, guide hooks), render manifest, guide content/stills, sitemap, and every "9 classes" test pin updated for 10.

## Type of change

## How was this tested?

- `npm run build` (all 5 entries green); `tsc --noEmit` clean.
- New suites: `tests/card_hand.test.ts` (deck determinism), `tests/card_adept_sim.test.ts` (card play through the effect pipeline in a live Sim), `tests/card_duel_queue.test.ts` (queue/pairing + class restriction), `tests/card_hand_view.test.ts` (pure view).
- Guard suites green: `tests/parity`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/command_schema.test.ts`, `tests/localization_fixes.test.ts`, `tests/i18n_completeness.test.ts`, `tests/localization_coverage.test.ts`, `tests/guide.test.ts`, `tests/talents.test.ts`.

## Card play -> effect resolution lifecycle

```mermaid
flowchart TD
  A[Combat starts: player.inCombat rising edge] --> B[startCombat: shuffle deck via Rng, draw opening hand]
  B --> C[Per tick: tickRedraw decrements timer]
  C -->|timer elapses| D[drawOne: reshuffle discard if deck empty]
  D --> C
  C -->|player plays card i| E[playCard i]
  E --> F{Focus >= card cost?}
  F -->|no| C
  F -->|yes| G[castAbility effectAbilityId: existing pipeline spends Focus, runs effects]
  G --> H[playCardAt: move card to discard]
  H --> C
  C -->|combat ends| I[endCombat: return hand + discard to deck]
```

## Duel matchmaking flow

```mermaid
flowchart LR
  A[queueCardDuel true] --> B{is Card Adept and not in duel?}
  B -->|no| A2[rejected silently]
  B -->|yes| C[joinCardDuelQueue: FIFO enqueue]
  C --> D[updateCardDuelQueue per tick]
  D --> E{two or more waiting?}
  E -->|no| D
  E -->|yes| F[tryPairCardDuel: pop oldest two]
  F --> G[startCardDuel: teleport to arena, face off, DuelState countdown]
  G --> H[updateDuels runs the bout]
  H --> I[duel ends: win / loss / disconnect tears it down]
```

## Skipped / simplified (clearly noted)

- **Card art via Tripo:** a real Tripo API call was made for a generic "trading-card class icon" prompt (task `869df279-7da1-4cda-a8e9-44bcb76aee40`, polled to `status: success`). The response is a single generic GLB + a 2D preview image, not per-ability art, and this repo's raster-art seam for ability icons (`ABILITY_IMAGE_IDS` + `public/ui/skills/<class>/<id>.webp`) expects one distinct image per ability id (8 for this class). Using one generic Tripo image for all 8 visually-distinct `ca_*` cards would misrepresent them, so it was not integrated; the existing procedural canvas icon generator (`ABILITY_RECIPES` in `src/ui/icons.ts`) remains the card-face art, matching how the other 9 classes render.
- **Online hand mirror:** the hand/deck reads are wired end to end offline; `ClientWorld` currently reports an empty hand (the snapshot delta-key mirror for the hand is a follow-up). `playCard` and `queueCardDuel` are fully server-authoritative today.
- **HUD painter + duel-queue window:** landed in a follow-up commit on this branch (`feat(card-adept): wire the hand bar and Card Duel queue into the live HUD`): `#card-hand` (the hand strip) and `#card-duel-window` (the join/leave queue window, gated behind `#mm-card-duel`, shown only for Card Adepts) are both real DOM today, painted by `CardHandPainter` / `CardDuelWindow` off the `card_hand_view` / `card_duel_window_view` pure cores. See Screenshots below for both.
- **Card Duel bout (in-progress/victory/loss) screen:** still not screenshot-able from a single offline client: pairing needs two live Card Adepts in the matchmaking queue, and duels only start server-side once `updateCardDuelQueue` pairs two waiting players. This is a genuine capture gap (not a missing feature) and stays a follow-up for an online two-client capture session.
- **No talent tree:** the Card Adept's progression is its deck, not a talent tree, so `talentsFor('card_adept')` is null by design (the always-on CJK talent-name i18n guard makes a bespoke tree a maintainer-fill task). The guide chooser gets a synthesized role/spec entry instead.
- **i18n:** English catalog keys added as usual; the M16 wordy-value rule additionally required non-Latin fills for the new class/ability/lore strings, so `zh_CN/zh_TW/ja_JP/ko_KR/ru_RU` overlays carry translations for those keys (needed for `tests/i18n_completeness.test.ts` to pass at PR tier).

## Screenshots

11 real captures of what is actually wired into the client today (offline mode, max graphics). No fabricated shots: a genuine 1v1 Card Duel bout (in-progress/victory/loss) needs two live Card Adepts paired by the server-authoritative queue, which a single offline client cannot produce, so it is not pictured (see the "Skipped / simplified" note above). Three other angles were attempted and honestly did not make it in: a desktop card-hand-strip capture, an ability-tooltip capture, and a level-20 character-sheet capture all failed during this capture session (local dev-server/headless-Chromium instability, "frame detached" / navigation timeouts), not a feature gap; the mobile hand-strip capture below did succeed and shows the same HUD element.

| | |
|---|---|
| Character creation, Card Adept selected, desktop | Character creation (desktop, 1920x1080): Card Adept selected, showing its stats, resource/armor/weapon, and signature abilities panel. |
| Character creation, Card Adept selected, mobile | Character creation (mobile, 390x844): Card Adept selected in the offline roster. |
| Character creation roster interaction, desktop | Character creation (desktop): full 10-class roster after clicking through Mage and Rogue before landing on Card Adept, showing it as the pressed/active choice. |
| Card Adept in the world, desktop | Entered world as a Card Adept (desktop): normal HUD, action bar with the class's `ca_*` abilities bound. |
| Card Adept in the world, mobile | Entered world as a Card Adept (mobile): HUD scales correctly for the new class. |
| Card Adept in a second zone | Card Adept at a different in-world location, for variety beyond the spawn area. |
| Card Adept world HUD, midnight theme | Card Adept world HUD under the "midnight" (dark) theme preset, for contrast against the default classic theme above. |
| Card Adept character sheet | Character sheet: "CARDESSA Level 1 Card Adept", portrait, and stats. |
| Card Adept spellbook, mobile | Spellbook window (mobile) open on a Card Adept, listing its `ca_*` abilities. |
| Card hand HUD strip, mobile | The `#card-hand` HUD strip (mobile), populated after forcing combat state to trigger the opening-hand draw: shows the card slots and deck/discard pile counters. |
| Card Duel queue window | The `#card-duel-window` matchmaking window after joining the Card Duel queue, showing the "queued" state and live queue size. |

---

## Review-round 2 fixes (addressing FernandoX7's review)

All 14 blocking/should-fix findings are resolved on the branch (see the detailed reply comment). Highlights:

- **SimContext extraction (S4):** the hand driver moved to `src/sim/card_play.ts` and the duel-queue system to `src/sim/social/card_duel.ts`, behind the `SimContext` seam; `sim.ts` keeps thin delegates. `cardDuelQueue` is now a live `SimContext` view. This is what let the Card Duel teleport follow the arena's real slot contract (B4) instead of the hand-rolled `arenaOrigin(0)`.
- **B3 resolved with the REAL wire fix (not the hide-online fallback):** the private hand card ids, deck/discard counts, and Card Duel queue status ride a self-only snapshot field `s.card` (the `maybe('duel', ...)` delta pattern), private by construction; `ClientWorld` mirrors the real values. Empty slots are `aria-disabled` + `pointer-events: none` and the click handler ignores them, so no blind `play_card`.
- **Balance rationale (finding 14):** the `ca_*` abilities stay `castTime: 0, cooldown: 0` on purpose. The deck is the balancing lever: hand size (max 6), redraw cadence, and the discard/reshuffle cycle gate throughput, so a Card Adept plays what the deck deals rather than free-firing its kit. Costs stay classic (each card mirrors its ability's rank-resolved Focus cost); no cooldown numbers were invented.


FernandoX7 (commented):
Thanks for this, and genuinely: the "Skipped / simplified" section is the most useful part of the PR body. Being upfront that the online mirror is not wired, that there is no talent tree, and that the duel-bout screenshot could not be captured saved me a lot of guessing, and I wish more PRs did that. The deterministic deck machine is the right shape too: `card_hand.ts` is a pure leaf that shuffles only through `Rng`, the `card_adept` guard really does keep a Card-Adept-free world at zero extra draws, and both view-cores are properly pure and in `UI_PURE_CORES`.

The problem is that most of the load-bearing behavior underneath that is broken, and a few of the "follow-ups" are not neutral gaps, they ship a surface that actively misbehaves. Findings below, verified against the code at the head ref rather than the description.

## Blocking

**1. Any card whose cast is refused is still destroyed.** `src/sim/sim.ts:4367-4379` calls `castAbility(...)` and then `playCardAt(state, index)` unconditionally. `castAbility` (`src/sim/combat/casting_lifecycle.ts:237`) silently returns on the GCD, an unlearned ability, a cooldown, a missing or wrong-type target, stun/silence, and an unaffordable resolved cost. In every one of those cases the card is spliced out of the hand into the discard for nothing. I reproduced all of it against a real `Sim`:

- Two cards played inside the 1.5s GCD (the ordinary case of clicking two cards quickly): first one casts and spends 20 Focus, second is refused, spends nothing, deals nothing, and is gone. Hand `[]`, discard `['card_quickstrike', 'card_jab']`.
- No target selected: card destroyed, no Focus spent, no effect.
- A level-1 adept holding `card_royal_flush` (`ca_royal_flush` is `learnLevel: 16`): card destroyed, nothing happens. At level 1 roughly half the 32-card deck references abilities the player has not learned yet, so half the hand is a trap.

The comment at `sim.ts:4370` says "so an unaffordable card is not wasted", but only the Focus arm is guarded. The card should move to the discard only when the cast actually resolved.

**2. Card icons never render.** `src/ui/hud.ts:5879` passes the resolver as `` (abilityId) => `url(${iconDataUrl('ability', abilityId)})` `` and `src/ui/card_hand_painter.ts:81` writes `` `url(${this.resolveBackgroundImage(...)})` ``, so the property ends up as `background-image: url(url(data:...))` and the browser drops it. The family convention is the other way: `action_bar_painter.ts:83` writes `this.resolveBackgroundImage(s.iconKey)` directly because the closure already carries the `url(...)`. Every card slot renders as the bare gradient today.

**3. Online, the class does not work, and the half-wired surface is worse than an absent one.** `src/net/online.ts:2275-2283` returns an empty hand and zero counts, and `:2625` returns `{queued: false, queueSize: 0}`. Nothing about the hand or the queue is on the wire (no field in `wireEntity` / `selfWireJson`). But the HUD gates on `cfg.playerClass`, which *is* mirrored, so the bar renders anyway:

- All six slots paint as `.empty`, yet the painter binds a click handler to all six in its constructor (`card_hand_painter.ts:44-46`) and `.card-hand-slot.empty` only sets `cursor: default`, never `disabled` or `pointer-events: none`. Clicking a blank square still sends `play_card` with that index, and the server plays whatever card really is at that index. An online Card Adept plays blind, off a bar that shows Deck 0 / Discard 0 forever.
- The duel window is a one-way door: `queued` is the only branch that renders the Leave button and the queue count (`card_duel_window.ts:98-107`), and online that branch is unreachable. You can join, you get no confirmation, and there is no way to leave.

The seam for this already exists: `selfWireJson` carries `maybe('duel', ...)`, `maybe('arena', ...)`, `maybe('vcup', ...)` and per-tick scalars like `combo`. The hand is a `combo`-shaped pair plus a small object. Until it rides the snapshot, the hand bar and the duel button should stay hidden online rather than render live controls over state the client cannot see.

**4. The Card Duel teleport bypasses the arena's slot contract and strands both players.** `src/sim/sim.ts:6742` hardcodes `arenaOrigin(0)`. Every other consumer allocates first (`freeArenaSlot` at `social/arena.ts:383`, then `arenaBusySlots.add(slot)` at `:732`, then `arenaOrigin(match.slot)`). So a Card Duel drops two players on top of whatever ranked bout is running in slot 0, and because the pairing loop drains the queue in one tick, two Card Duel pairs stack on identical coordinates.

Worse, it records no return positions and `endDuel` (`social/duel.ts:152`) does not teleport, because normal duels are fought in place and that path does not exist. So when the bout ends both players are just standing at `x = 4200`, inside the walled arena pit (`ARENA_COLLIDERS` applies to any `isArenaPos`, `colliders.ts:447`). Every `x > DUNGEON_X_THRESHOLD` (600) check then reads them as "inside an instance", so they cannot start a delve ("Leave the arena first."), and a hearthstone is the only way out. `startArenaMatch` / `returnFromArena` (`arena.ts:708`, `:1002`) are the contract to follow: stash `{x, z, facing}` per player on start, restore on end, release the slot.

**5. `queueCardDuel` is missing the gates its own precedent enforces.** `src/sim/sim.ts:6693-6706` checks class, in-duel, and already-queued. `arenaQueueJoin` (`social/arena.ts:87-106`) additionally refuses a dead player, a player already in an arena match, a player mid-trade, and a player inside an instance, and `updateCardDuelQueue` (`sim.ts:6726`) only prunes logged-out or already-dueling pids. As written, a player in a dungeon or a live arena bout can be yanked out of it with no consent moment, and a dead player can queue, get teleported, and lose instantly (crediting the opponent a `duelsWon`). Relatedly, `card_duel_queue` is not in `JAILED_BLOCKED_COMMANDS` (`server/game.ts:411-421`), whose comment says the list is exactly "everything that queues into or enters instanced content ... plus starting or accepting a duel". A jailed player can currently queue and be teleported out of the cage.

## Should-fix

**6. The Focus gate reads the card's pinned cost, not the resolved rank cost.** `cards.ts` pins every Arcane Bolt card at 35, but `ca_arcane_bolt` costs 40 at rank 2 and 55 at rank 3 (`classes.ts`). So a level-12 adept with 45 Focus passes the `p.resource < def.cost` check at `sim.ts:4376`, the cast is refused for cost, and finding 1 eats the card. `card_hand_view.ts:50` computes `playable` from the same pinned cost, so the HUD lies about affordability at every rank past the first. A `cost_tax` aura has the same effect. Resolve the cost through `resolvedAbility` instead of the `CardDef` literal.

**7. Four deeds are permanently uncompletable for this class.** `prog_talented`, `prog_specialized`, `prog_deep_roots` and `prog_full_build` (`src/sim/content/deeds.ts:73-103`) all gate on talent points, a chosen spec, or a capstone node, and `talentsFor('card_adept')` is null by design. `docs/design/deeds.md` requires every deed to be completable. Card Adepts also accrue talent points they can never spend. Either the class needs a tree, or those four deeds need a class-aware alternative.

**8. The 10-class sweep missed four string-keyed sites, none of which `tsc` can see.**
- `server/player_card.ts:79-88` keeps its own hand-rolled 9-class union, so `classDisplay` falls through to `unknownClass` and a Card Adept's public share card reads "Adventurer".
- `src/game/editor_playtest.ts:23` `VALID_CLASSES` has no `card_adept`, so the class cannot be playtested in the editor.
- `src/admin/i18n.ts:227` `CLASS_LABEL_IDS` has no `card_adept`, so the admin dashboard renders the raw id. The dashboard is in i18n scope (operators are users), and the English key is missing from `src/admin/i18n.en.ts:477-485`.
- `src/styles/shell.css:2545` and `:5547` have no `[data-class="card_adept"]` `--class-color` rule, so the class loses its color on the character-create button and the portrait chip even though `classes.ts` defines `0x8a5cf0`.
- `bot/logic.ts:126` `CLASS_EMOJI` has no entry (harmless, falls back to empty).

**9. The card system lives on the `sim.ts` coordinator.** About 150 lines of `updateCardHand` / `playCard` / `queueCardDuel` / `updateCardDuelQueue` / `startCardDuel` landed as methods on `Sim`. Per the root CLAUDE.md a new sim system belongs in its own module behind the `SimContext` seam: the hand driver next to `card_hand.ts`, the duel queue next to `social/duel.ts` and `social/arena.ts`. This is not just style. `startCardDuel` drifted from the arena's own start/return contract precisely because it was written in the coordinator instead of beside `arena.ts`, where `freeArenaSlot` and `returnFromArena` are right there.

**10. Test gaps on exactly the broken paths.**
- `tests/card_hand.test.ts:70-83` is titled "reshuffles the discard when empty" but never reaches `reshuffleDiscard`: the loop only calls `drawOne` while `deck.length > 0`, so the empty-deck branch runs zero times and the test asserts only card conservation. Delete `reshuffleDiscard` and it stays green.
- `tests/card_hand.test.ts:26` compares `card.cost` to the rank-1 `ABILITIES[id].cost` and ignores `ranks[]`, which is why finding 6 slipped through.
- Nothing covers `playCard`'s refusal arms, `startCardDuel` / `updateCardDuelQueue` (only the pure FIFO module is tested), Sim-level draw determinism across two same-seed worlds, or the "zero extra rng without a Card Adept" claim (asserted only in a comment).
- `tests/card_hand.test.ts:67` is a constant self-comparison (`toBe(STARTING_HAND_SIZE)` against a loop that runs `STARTING_HAND_SIZE` times). `:92` gets this right with a literal 40, keep that pattern.
- No `card_hand_painter` test, which is what would have caught the `url(url(...))` bug.

**11. The card buttons have no accessible name, and dead slots stay clickable.** The six `<button>`s in `index.html:342-347` / `play.html:291-296` contain only an empty `<span class="chs-cost">`, so an empty slot is a button with no text (an axe `button-name` violation), and a filled one announces a bare number. `CardDef.name` exists in `cards.ts` but is rendered nowhere: no label, no tooltip, no aria, so a player identifies a card by icon alone, and cards that share an `effectAbilityId` (Quickstrike/Jab, Arcane Bolt/Spark) are indistinguishable. The painter comment says it "disables the button as a UX hint" but it never sets `disabled` or `aria-disabled`.

**12. Per-frame cost in `update()`.** `hud.ts:6908-6917` calls all three `IWorld` card reads plus a fresh input-object literal every frame for all ten classes before the core's `isCardAdept` early-return can help, and `Sim.cardHandIds()` (`sim.ts:4354`) allocates a fresh array on every call. `card_hand_view.ts:38-58` also allocates a new slots array plus one object per card per frame, and it is not registered in the `hud_perf_budget` allocation probe. Separately, `hud.ts:6901-6907` does a raw `document.getElementById` plus a `.style.display` write inside `update()`, bypassing the elided writers, for a value that is a session constant.

**13. `place()` skips the rebucket.** `sim.ts:6741-6748` sets pos, prevPos and facing, but `placeInArena` (`social/arena.ts:814-825`) also sets `prevFacing` and calls `ctx.rebucket(e)`. Without the rebucket every spatial query for the rest of that tick still sees the player in their old grid cell, and the stale `prevFacing` makes the renderer interpolate through a spin.

**14. The whole kit is instant and cooldown-free.** All eight `ca_*` abilities are `castTime: 0, cooldown: 0` (`classes.ts`), including an 80-Focus nuke. No other caster has a fully instant, never-interruptible kit that can be thrown while moving. CLAUDE.md asks that balance follow classic-era formulas rather than invented numbers, so I would want this one argued rather than assumed.

## Nits

- 20 card records collapse onto 8 abilities, so several are pure aliases (Wild Card is `rare` but resolves to the same `ca_flame_fan` as the `uncommon` Flame Fan). `CardSlotView.rarity` is carried through the view and never used.
- `sim.ts:6749` re-types `DUEL_COUNTDOWN` as a literal `3` in two places instead of importing it (`social/duel.ts:17`), so a tuning change would desync the card duel countdown from the seconds it emits.
- `sim.ts:3774` puts `lap?.('p.cards')` inside the `if (meta.cardHand)`, so the phase profiler emits that lap for some players and not others and `p.regen` quietly absorbs the card time for everyone else.
- The doc comment for `playCard` (`sim.ts:4348-4353`) sits above `cardHandIds()` instead, with the `// IWorld read surface` line run onto its tail.
- `card_duel_window.ts:92`: `void action;` is dead, `wire()` never uses the parameter.
- `#mm-card-duel` reuses `data-icon="arena"` (same glyph as the Arena button next to it), has no keybind span, and hides itself with an inline `style="display:none"` rather than the `hidden` class the rest of the HUD uses.
- `hud.mobile.css` only resizes the slots. Six 44px slots plus the piles column is roughly 330px inside a `#bottom-bar` that is inset 172px each side on touch, which looks like it will collide with the action ring on a small landscape phone. Worth a real-device check.
- `server/game.ts:3533` guards `play_card` with a bare `typeof msg.index === 'number'`. It is safe today (`hand[NaN]` is `undefined` and the call bails), but the neighbouring `castSlot` normalizes with `msg.slot | 0`.
- `tests/command_schema.test.ts:26-27`: the count bump is right but the new trailing comment lists bank / Vale Cup / town focus and never mentions the two commands this PR adds.
- `scripts/card_adept_pr_shots.mjs` and `card_adept_pr_shots_more.mjs` are two near-duplicate one-off capture scripts. Fold them into one if they are meant to stay.

## Merge

The base is `release/v0.26.0` but the branch merged `v0.25.0`, and the only true conflicts are the usual generated trio (`i18n.resolved.generated/pending.ts`, `i18n.resolved.sha256`, `i18n.status.summary.json`). That is mechanical regen churn: merge the base and re-run the i18n build.

Happy to look again once the discard rule, the icon wrap, and the duel teleport are sorted. The deck machine at the bottom of this is good, it just needs the layer above it to be as careful.

jgyy:
@FernandoX7 Thanks for the thorough, code-verified review. It caught real bugs in the layer above the deck machine, and I have worked through every blocking and should-fix item. Rundown, in your numbering:

## Blocking
- **B1 (refused cast still destroys the card):** `castAbility` now returns whether the press actually committed (an instant resolved, a cast-time/channel started, an on-next-swing or same-cast press queued), and `false` for every refusal. `playCard` moves a card hand to discard only on a `true` return, so a card refused for the GCD, a cooldown, an unlearned or unaffordable ability, stun/silence, or a missing/out-of-range target stays in hand.
- **B2 (double `url()` wrap):** the resolver already carries the `url(...)`, so the painter writes it directly now (matching `action_bar_painter`), no more `url(url(...))`. Guarded by a new `card_hand_painter` test.
- **B3 (online is half-wired):** done the real fix, not the fallback. The private hand card ids, deck/discard counts, and Card Duel queue status now ride a self-only snapshot field `s.card` (the `maybe('duel', ...)` delta pattern), so it is private by construction (selfWireJson reaches only the owning session). `ClientWorld` mirrors the real values; empty slots are now `aria-disabled` with `pointer-events: none` and the click handler ignores them, so no blind `play_card`.
- **B4 + B13 (arena slot contract + rebucket):** `startCardDuel` now allocates via `freeArenaSlot`, reserves the slot in `arenaBusySlots`, stashes each fighter's pre-teleport position, and places them with `placeInArena` (which sets `prevFacing` and rebuckets). `endDuel` teleports Card Duel fighters home and releases the slot, so they are no longer stranded at arena coords.
- **B5 (missing queue gates):** `queueCardDuel` and `updateCardDuelQueue` now enforce the `arenaQueueJoin` gates (dead, already in an arena match, mid-trade, inside an instance) and prune an entrant who entered an instance or died after queueing. `card_duel_queue` is added to `JAILED_BLOCKED_COMMANDS`.

## Should-fix
- **S1 (pinned vs resolved cost):** the hand view resolves each card's cost through the talent-and-rank resolved ability (the IWorld `known` list), not the pinned `CardDef.cost`, so affordability is correct at every rank and under `cost_tax`; an unlearned card shows unplayable.
- **S2 (four uncompletable deeds):** `prog_talented`/`prog_specialized`/`prog_deep_roots`/`prog_full_build` now substitute a Card Duel win ladder (1/3/5/11) for any class with no talent tree, so they stay completable per deeds.md. Two moved from the `talentPoints` meter to class-aware flags (a meter must not read the deedStats counters, per the dirty-key contract).
- **S3 (missed string-keyed sites):** added card_adept to `server/player_card.ts` (+ newlocales), `editor_playtest`, the admin `CLASS_LABEL_IDS` + English key, and a `shell.css` `--class-color` rule (0x8a5cf0); `VALID_CLASSES` in character-create now derives from `ALL_CLASSES`.
- **S4 (logic on the coordinator):** extracted the hand driver to `src/sim/card_play.ts` and the duel-queue system to `src/sim/social/card_duel.ts` behind the `SimContext` seam; `sim.ts` keeps thin delegates. This is what let B4 follow the arena contract instead of drifting from it.
- **S5 (test gaps):** the reshuffle test now actually drains the deck and reaches `reshuffleDiscard`; added a rank-aware cost test and a `card_hand_painter` test asserting no `url(url(...))` and that empty slots ignore clicks.
- **S6 (no accessible name, dead slots clickable):** every card button now carries an accessible name (the localized ability name filled, a localized "empty card slot" empty) plus a tooltip; empty slots are `aria-disabled` + `pointer-events: none` and inert.

## Performance
Only a Card Adept now runs the three card reads and builds the view each frame (every other class paints the hidden bar once); the Card Duel minimap toggle visibility is set once at build time; the cost resolver is a bound field so the hot path allocates no per-frame closure. Not yet done: making `buildCardHandView` return a reused container and registering it in the alloc probe (it now only runs for the one class, so the per-frame garbage is bounded to Card Adepts).

## Balance
The eight `ca_*` abilities are intentionally `castTime: 0, cooldown: 0`: the deck is the balancing lever, not the GCD. A card is gated by hand size (max 6), the redraw cadence, and the discard/reshuffle cycle, so a Card Adept cannot free-fire its whole kit; it plays what the deck deals. Rather than invent cooldown numbers, the resource cost stays classic (each card mirrors its ability's Focus cost, resolved by rank).

## Nits
Imported `DUEL_COUNTDOWN` instead of re-typing `3`; normalized `play_card`'s index with `| 0`. A few smaller nits (the `void action;` dead param, the `lap?.('p.cards')` placement, the `data-icon="arena"` glyph reuse) are noted but not all swept.

jgyy:
Closing this in favor of #1918: the original plan added a new playable class (`card_adept`) whose character model depended on a Tripo-based asset-generation workflow that is no longer available to me. Redid the card mechanic as a class-agnostic, NPC-hosted Card Duel minigame instead (no new class, no new model/rig) — see #1918 for the full writeup.

Issue #1852, opened 2026-07-13
UI/UX: item interactions and options are confusing (right-click bag item defaults to Delete instead of Equip; loot flow and options discoverability need work; Spanish localization is mediocre)


jgyy:
## Problem

Player feedback (issue #110 discussion) is that the interface itself is not the problem, but access to options, descriptions, and item locations is confusing. Concretely:

- Right-clicking an item in the bag currently opens Delete instead of the expected primary action (Equip/Use). This makes the single most common inventory interaction dangerous by default.
- Several common actions are not in the menu a player would intuitively check first.
- The loot flow (rolling, who looted what, ownership) is not well understood by players.
- The Spanish (es) localization quality is noticeably weaker than other locales.

The ask is to prioritize cleaning up this foundation (WoW-style simplicity: obvious, low-risk default actions) over adding new content.

## Expected behavior

- Right-clicking a bag item performs (or defaults the menu to) the expected primary action for that item type (Equip for equippable gear, Use for consumables), not Delete.
- Destructive actions (Delete) are not the default single-click/right-click outcome; they require an extra confirmation step or a non-default menu position.
- Loot outcomes (who looted, why, need/greed/roll result) are clearly communicated to players.
- Spanish (es) locale strings read naturally and match the quality bar of other maintained locales.

## Scope

- `src/ui/` bag/inventory context menu (item interaction menu construction and default action wiring).
- Loot window / loot event messaging (clarity of who received what and why).
- `src/ui/i18n.locales/es.ts` string quality pass.
- General audit of menu/option discoverability across HUD windows (informational, for maintainer triage into follow-up issues/PRs).

## Acceptance criteria

- [x] Bag item right-click default action fixed to Equip/Use, with Delete requiring confirmation (Shift+right-click) and demoted from the default slot: #1853.
- [ ] Loot clarity improvements scoped and filed as follow-up work if not included in the same pass.
- [ ] Spanish locale quality pass scoped and filed as follow-up work.
- [ ] Wider discoverability/options-access audit captured as a punch list for maintainer triage.

## Screenshot

**Before:** right-clicking a bag item jumped straight to the destroy prompt.

![before](https://raw.githubusercontent.com/jgyy/world-of-claudecraft/fix/bag-rightclick-equip/docs/screenshots/bag_rightclick_before_destroy_bug.png)

**After (#1853):** a plain right-click now equips/uses the item.

[...]

PR #1853, opened 2026-07-13
fix(ui): right-click a bag item equips/uses it, not destroy

jgyy:
## Summary

Follow-up to #1852 (the broader UI/UX cleanup issue filed for this feedback). This PR
scopes to the single, concrete, reproducible bug called out: right-clicking a bag
item jumped straight to a destroy confirmation prompt, which is the opposite of the
classic-MMO expectation that right-click on an inventory item uses or equips it.

- A plain right-click on a bag item now runs the exact same primary action a
  left-click would (equip / use / trade / market-sell / vendor-sell / bank-deposit /
  pet-feed / quest-discard), via a new shared `runBagItemAction` dispatch used by
  both the click and non-shift contextmenu handlers in `bags_window.ts`.
- **Shift+right-click** keeps the destroy-item shortcut (previously a bare
  right-click), so the affordance isn't lost, just demoted off the accidental
  default. The tooltip hint was updated from "Right-click to destroy" to
  "Shift+right-click to destroy" (`hudChrome.bags.rightClickDestroy`).
- Touch long-press-to-peek and the vendor Ctrl/Meta right-click-to-sell paths are
  unchanged.

## Right-click dispatch flow

```mermaid
flowchart TD
    A[Right-click a bag item] --> B{pointerType is touch/pen,\nor mobile-touch non-mouse?}
    B -- yes --> C[preventDefault only\nlong-press peek owns touch]
    B -- no --> D{Vendor window open?}
    D -- yes --> E{Ctrl/Meta held?}
    E -- yes --> F[sellBagItem\nsplit-stack sell prompt]
    E -- no --> G[no-op]
    D -- no --> H{Shift held?}
    H -- no --> I[runBagItemAction\nsame primary action as left-click:\nequip / use / trade / market-sell /\nvendor-sell / bank-deposit / pet-feed / quest-discard]
    H -- yes --> J{bagDestroyAction result}
    J -- none --> K[no-op]
    J -- discardBlocked --> L[showError cannotDestroy\nissue 1501 protection]
    J -- destroy --> M[Destroy confirmation prompt]
```

## Review follow-up (addressed)

- **Stale locale overlays.** The tooltip reword changed the English meaning but the
  ~18 non-English overlays still held the old "right-click to destroy" wording with
  no gate signal. Dropped the stale `hudChrome.bags.rightClickDestroy` entry from
  every affected overlay so the scan re-marks it `pending` for the maintainer's
  release-time refill, then regenerated `i18n:gen` + `i18n:hash -- --write`.
- **Unused parameter.** Dropped the unread `item` param from `runBagItemAction`;
  updated the two source-scan pins in `tests/bags_window.test.ts` to match.
- **Broader right-click scope, confirmed intended.** A plain right-click now also
  runs the transactional primary action in trade / mail / market-sell / bank-deposit
  / pet-feed modes (previously inert there), matching "right-click does the same as
  left-click." Added a structural pin asserting each transactional case in
  `runBagItemAction`'s switch wires to its staging call, complementing
  `bagItemAction`'s existing exhaustive per-mode pins in `tests/bags_view.test.ts`.

## Screenshots

**Before:** sword sitting in the bag.

before

**After a plain right-click:** the item is equipped (see chat log "Equipped Pitted
Shortsword"), not destroyed.

after equip

**Shift+right-click:** still opens the destroy confirmation.

shift destroy

Captured with the new `scripts/bag_rightclick_equip_shot.mjs` tour script (needs
`npm run dev`).

## Test plan

- [x] `npx vitest run tests/bags_view.test.ts tests/bags_window.test.ts` (111 tests;
      includes the new source-level pins for the non-shift right-click dispatch
      order, the shared `runBagItemAction` method, and the transactional-mode
      reachability check added in review)
- [x] `npx tsc --noEmit -p .`
- [x] `npm run i18n:gen` + `npm run i18n:hash -- --write` (English catalog string
      change for the tooltip hint; stale non-English overlays dropped back to
      `pending`)
- [x] `npm run build`
- [x] `npm run gate` (full suite green apart from two pre-existing, unrelated flakes:
      `tests/server/http/characterization.test.ts` / `parity.test.ts` leaderboard
      cases fail only under full-suite worker contention and pass in isolation; not
      touched by this change)


EnriqueGF:
:) Thanks for the fix, this is a genuinely clean extract-method and a sensible classic-MMO correction.

I reviewed the branch against `release/v0.25.0` (correct latest release base, gate passes) and ran the checks locally: `tsc --noEmit` clean, `tests/bags_window.test.ts` + `tests/bags_view.test.ts` green, `npm run i18n:gen`/`i18n:scan` reproduce zero drift, biome clean. The `s3_registered` localization failure I saw is pre-existing on the base branch (dev-command strings in `sim.ts`), not introduced here. The seam split is right: decision logic stays in the pure core (`bagItemAction`/`bagDestroyAction`), side-effect dispatch stays in the painter, and the non-shift branch correctly sits after the touch-peek and vendor Ctrl/Meta guards so those paths are preserved.

One blocking issue before this is production-ready, plus two minor notes.

**1. (Blocking) Stale localized tooltips now contradict the behavior in ~18 languages.**
The English value of `hudChrome.bags.rightClickDestroy` changed meaning from "Right-click to destroy" to "Shift+right-click to destroy", but every non-English overlay in `src/ui/i18n.locales/` still holds the old translation, for example:
- `fr_FR.ts:355` "Clic droit pour détruire"
- `de_DE.ts:351` "Rechtsklick zum Zerstören"
- `es.ts:352` "Clic derecho para destruir"
- plus it_IT, pt_BR, ja_JP, ko_KR, zh_CN, zh_TW, ru_RU, pl_PL, nl_NL, cs_CZ, da_DK, sv_SE, tr_TR, vi_VN, id_ID (~18 total).

These now tell localized players that a plain right-click destroys the item, which is exactly the surprise this PR removes, and for a destructive action. This is not caught by any gate: `contentHash("Shift+right-click to destroy")` equals the stored `enHash` `12dd73288575536d`, and `scripts/i18n_scan.mjs` recomputes each locale `srcHash` to that same new hash and marks them `translated` (the reword-staleness comparison is intentionally dormant, see the comment at `scripts/i18n_scan.mjs:33-37`). So `i18n_completeness` stays green and the release-tier gate finds no `pending` rows, yet the shipped text is wrong. I understand overlays are normally the maintainer's release-time job and contributors add English only, but here the reword silently invalidated existing translations with no signal, so it needs to be surfaced explicitly: either flag these ~18 keys for re-translation in the PR, or drop the stale overlay entries so the scan re-marks them `pending` and the maintainer refills them, then re-run `npm run i18n:build`.

**2. (Minor) Unused parameter.**
`runBagItemAction(s, item, action, ev)` at `src/ui/bags_window.ts:578` never reads `item` (biome flags it as a warning). The call sites already pass `bagItemAction(item, ...)` as `action`, so `item` can be dropped from the signature (the two source-scan tests pin the signature, so update them in the same change).

**3. (Minor, please confirm) Broader right-click scope than the title implies.**
Because `runBagItemAction` runs the full mode switch, a plain right-click now also performs the transactional primary action in trade / mail / market-sell / bank-deposit / pet-feed modes, which were previously inert on right-click (`bagDestroyAction` returned `'none'`). Vendor mode is unaffected (early-returned). This is consistent with "right-click does the same as left-click" and is likely intended, but the screenshots only exercise the default bag mode, so please confirm that is the desired behavior. A behavioral test (plain right-click in default mode calls `equipBag`/`useItem`; a transactional mode dispatches its staging call) would lock in the new reachability, since both current tests are structural source scans.

Happy to re-review once the tooltip translations are handled. Nice work on the shared dispatch and the proof script.

jgyy:
Thanks for the thorough review! Addressed all three points in a follow-up commit (7ff705b23):

**1. (Blocking) Stale locale overlays.** Dropped the `hudChrome.bags.rightClickDestroy` entry from all 18 affected `src/ui/i18n.locales/<lang>.ts` overlays (fr_FR, de_DE, es, it_IT, pt_BR, ja_JP, ko_KR, zh_CN, zh_TW, ru_RU, pl_PL, nl_NL, cs_CZ, da_DK, sv_SE, tr_TR, vi_VN, id_ID) so `i18n_scan` re-marks the key `pending` in each locale instead of `translated`, then ran `npm run i18n:gen` + `i18n:hash -- --write` to regenerate the resolved tables/hash. The maintainer can refill it at release per the normal overlay workflow; it will no longer silently ship the stale "right-click destroys" wording.

**2. (Minor) Unused parameter.** Dropped `item` from `runBagItemAction(s, item, action, ev)` since the body never read it (all call sites already pass `bagItemAction(item, ...)` as `action`). Updated the two source-scan pins in `tests/bags_window.test.ts` to the new signature.

**3. (Minor, confirm) Broader right-click scope.** Confirmed intended: "right-click does the same as left-click" is the whole point of the shared dispatch (issue 1852's classic-MMO expectation isn't equip-only, it's "right-click uses the item the way left-click would" across every mode). Rather than a live-DOM behavioral harness (this file's suite is source-scan only, per its existing convention, and there's no live-instantiation pattern elsewhere in the bags/vendor window tests to follow), I added a structural pin that reads `runBagItemAction`'s switch body directly and asserts each transactional case (`trade`/`mailAttach`/`marketSell`/`bankDeposit`/`petFeed`) wires to its staging call. Combined with `bagItemAction`'s already-exhaustive per-mode pins in `tests/bags_view.test.ts`, that proves reachability from right-click: shared dispatch (pinned) + correct action per mode (pinned) + each action wired to its call (newly pinned).

Full re-run: `tsc --noEmit` clean, `npx vitest run tests/bags_window.test.ts tests/bags_view.test.ts` green (111 tests), `i18n:gen`/`i18n:hash --write` regenerated, biome clean on the changed files. Ready for another look.

jgyy:

jgyy:

EnriqueGF:
:) Thanks for turning this around so cleanly.

Re-reviewed the current head (7ff705b23 + merge 39099db83), base is `release/v0.26.0` (correct latest release). All three points from the prior review are resolved:

1. (Was blocking) Stale localized tooltips: the `hudChrome.bags.rightClickDestroy` overlay is now dropped from all 18 locales (confirmed `fr_FR.ts` no longer contains the key), so the scan re-marks it `pending` for the maintainer to refill rather than silently shipping the old "right-click destroys" wording. Correct fix.
2. (Minor) Unused `item` param dropped from `runBagItemAction`; the two source-scan pins were updated to the new signature.
3. (Minor) Broader right-click scope confirmed intended, with a new structural pin over the `runBagItemAction` switch asserting each transactional case wires to its staging call. Combined with the exhaustive per-mode pins in `bags_view.test.ts`, reachability is covered.

Verified locally: `tsc --noEmit` clean, `tests/bags_window.test.ts` + `tests/bags_view.test.ts` green (56 tests). No new concerns. LGTM.

FernandoX7 (approved):

jgyy:

PR #1854, opened 2026-07-13
fix(sfx): fail loudly on a mob sfx file that matches nothing

jamiecypher:
## Summary

Fixes a silent-failure gap in mob sfx discovery: a `mob_*.mp3` file that matches neither a real catalog entry nor a correctly-numbered subfamily pattern (`mob_<family>_<sub>_<action>_<N>.mp3`) was silently skipped with no error, the same failure class as the `cast_lightning_bolt` bug fixed in #1771.

Concretely: a subfamily file missing its numbered suffix (e.g. `mob_beast_wolf_hurt.mp3` instead of `mob_beast_wolf_hurt_1.mp3`), or a family-level file with no matching catalog entry (e.g. dropping `mob_reptile_aggro.mp3` in without adding the catalog row), would just vanish from every manifest rebuild. No error, no warning, nothing in the discovery result's `errors` array. The contributor gets a silently-mute mob and no clue why.

`discoverSfxTracks` now tracks which filenames the catalog loop already consumed (bare or numbered), and any `mob_*.mp3` file that matches neither that set nor the numbered subfamily pattern gets a real error pushed instead of a silent `continue`. That error surfaces through `buildSfxManifestData`'s existing hard-fail (`throw` on any discovery error), which runs as part of `npm run build`/`npm run gate` and on every CI job, so it now blocks with a clear message naming the exact bad filename instead of shipping silently broken.

Scoped to just this one gap, no unrelated changes. Verified zero overlap with #1828 (that PR only touches `src/ui/combat_sfx.ts` + its test file).

## Type of change

- [x] Bug fix
- [x] Test: new/updated coverage

## How was this tested?

- `npx tsc --noEmit` clean
- `npx vitest run tests/sfx_manifest.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts tests/combat_sfx.test.ts tests/architecture.test.ts` — 96 tests passing
- Added 3 new decisive tests: a bare subfamily file with no number now errors, a family-level file with no catalog match now errors, and confirmed the fix does NOT false-positive on a family-level file that legitimately matches a real catalog entry
- Ran `npm run sfx:manifest` against the real repo content: builds cleanly, 138 clips, zero errors — confirmed all 33 currently-shipped `mob_*.mp3` files are bare catalog entries already correctly consumed by the catalog loop, none get newly flagged
- Confirmed the check only fires on `npm run build`/`npm run gate`/CI, not `npm run dev`, so it doesn't interfere with casual local play/testing

## Screenshots / recordings

N/A — no visual change, this only affects build-time validation output.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [ ] Builds succeed. (ran `sfx:manifest` directly; full `npm run build` not run locally this pass)

Cross-platform
- ~Tested on desktop and mobile.~ N/A, build tooling only

Localization (i18n)

Hygiene

### DIFF
--- a/scripts/sfx/sfx_manifest_builder.mjs
+++ b/scripts/sfx/sfx_manifest_builder.mjs
@@ -52,6 +52,10 @@ export function discoverSfxTracks(catalog, sfxDir) {
   const entries = {};
   const errors = [];
   const catalogKeys = new Set();
+  // Filenames already accounted for by a catalog entry (bare or numbered), so
+  // the mob-extension pass below can tell "this belongs to the catalog loop"
+  // apart from "this looks like a mob sfx file but matches nothing."
+  const consumedFilenames = new Set();
   const sfxFiles = existsSync(sfxDir) ? readdirSync(sfxDir).sort() : [];
 
   for (const source of catalog) {
@@ -71,6 +75,7 @@ export function discoverSfxTracks(catalog, sfxDir) {
       .map((filename) => ({ filename, id: filename.match(numberedPattern)?.[1] ?? null }))
       .filter((candidate) => candidate.id !== null);
     for (const candidate of numbered) {
+      consumedFilenames.add(candidate.filename);
       const numericId = Number(candidate.id);
       if (
         !/^[1-9]\d*$/.test(candidate.id) ||
@@ -97,6 +102,7 @@ export function discoverSfxTracks(catalog, sfxDir) {
         const filename = `${source.key}${extension}`;
         if (!existsSync(path.join(sfxDir, filename))) continue;
         tracks.push(track('main', filename));
+        consumedFilenames.add(filename);
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


EnriqueGF:
:) Thanks, this is a nice defensive fix that closes a real silent-failure gap.

Reviewed and it looks production-ready. What I verified against the real code:

- Base branch: targets release/v0.26.0 (the latest release branch) and is up to date with it. Gate passes.
- Correctness: the consumedFilenames set is populated in both catalog paths (numbered variants and the bare PROBE_EXTENSIONS probe), so a mob file already claimed by the catalog loop is skipped before the shape check, and only files matching neither a catalog entry nor the mob_<family>_<sub>_<action>_<N> pattern now error. Confirmed PROBE_EXTENSIONS leads with .mp3, so bare .mp3 catalog files are consumed and never falsely flagged.
- No false positives on real content: ran npm run sfx:manifest against the repo, 138 clips, zero errors, all currently-shipped mob_*.mp3 files pass.
- Tests: the three new cases are decisive: bare subfamily file missing its number errors, family-level file with no catalog match errors, and a family-level file WITH a matching catalog entry stays clean. Both error assertions pin the filename and the message. The error surfaces through buildSfxManifestData's existing hard-fail, so it blocks build/gate/CI.

Ran locally in a worktree: npx tsc --noEmit clean, the five sfx/architecture suites (96 passing), biome clean on both changed files.

Minor, non-blocking edge case worth a mention: if a catalog key ever ships BOTH numbered takes and a leftover bare mob_*.mp3, the bare file would now be flagged as unrecognized (previously silently ignored). That is arguably the correct behavior since the bare file is redundant when numbered takes exist, and no current content hits it, so nothing to change here, just noting it.

PR #1855, opened 2026-07-13
fix(game): stop keybind Q/E re-seeding, fix H-key collision; feat(ui): pet-feed and meters discoverability

jgyy:
## Summary

A player reported that from session to session, three things "randomly" changed: Q/E flipped
between strafing and action-bar skills, the ability to feed/heal their hunter pet sometimes
disappeared, and the threat/DPS meter sometimes wasn't there. Investigation found this was
actually two separate, real issues, plus two conditional-visibility behaviors that were working
as designed but looked broken:

1. **Corrupted keybind profiles from a reverted overhaul (`src/game/keybinds.ts` + new
   `src/game/keybinds_repair.ts`)**: the reverted interface overhaul (commit `1d2678f58`,
   reverted by #1788) shipped `slot10 = ['KeyQ','Minus']` / `slot11 = ['KeyE','Equal']` with
   Strafe Left/Right on empty defaults. Any per-character profile SAVED during that window
   persisted Q/E driving action-bar slots and strafe unbound; after the revert those explicit
   stored values stick (a stored binding always wins over the current default), so Q/E kept
   flipping between strafe and slots depending on which character had saved when. These profiles
   live under the scoped key `woc_keybinds:char:<id>` (per-character scoping landed in
   `b97b5730`), NOT the bare legacy blob, so the fix is a one-time, signature-keyed repair on the
   scoped-profile load path. It fires ONLY on the exact corrupted fingerprint and deletes just
   those keys so they re-seed to the current defaults; deliberate remaps are left verbatim.
   (This replaces the earlier `isDefaultOfOtherActionBarSlot` legacy-blob guard, which never
   reached the scoped profiles and could revert deliberate remaps: see the review thread.)
2. **`KeyH` collision**: `targetFriendly` and the Damage Meters toggle both defaulted to `KeyH`,
   and the loader silently evicted the meters binding on every load. Meters now defaults to
   `Shift+KeyH` (matching the existing `Shift+<letter>` convention used elsewhere), and eviction
   only fires on a genuine runtime collision, never silently for two defaults that no longer share
   a key. Profiles the collision already emptied (stored `meters: [null,null]` while
   `targetFriendly` holds `KeyH`) are re-seeded to `Shift+KeyH` by the same signature repair, since
   a stored null there can only be loader eviction (`clear()` has no caller and null capture is a
   cancel, never an unbind).
3. **Pet-feed icon discoverability**: the feed button was always visible and only ever explained
   itself via an error toast after a failed click. It now shows a disabled state with a tooltip
   explaining why (pet at full health, or no eligible food in bags) instead of just failing
   silently.
4. **Threat/DPS meter discoverability**: the meters window has no persistent on/off setting; rows
   fill in automatically once the party deals damage or healing, and the segment closes a few
   seconds after combat ends. Added a hint line explaining this so it doesn't read as an
   inconsistent toggle.

```mermaid
flowchart TD
    A[Player reports: Q/E, pet-feed, threat bar<br/>all seem to randomly change] --> B{Investigate root cause}
    B --> C[Keybinds]
    B --> D[Pet-feed icon]
    B --> E[Threat/DPS meter]

    C --> C1[New/unsaved character profile<br/>seeds from legacy woc_keybinds blob]
    C1 --> C2{Legacy blob has Q/E on<br/>action-bar-slot keys?}
    C2 -- yes --> C3[Character keeps old convention,<br/>looks flipped vs. saved characters]
    C2 --> C4[targetFriendly and meters<br/>both default to KeyH]
    C4 --> C5[Loader silently evicts meters binding<br/>every load]
    C3 --> F1[Fix: refuse legacy import that collides<br/>with a current Action Bar default]
    C5 --> F2[Fix: give meters its own default<br/>Shift+KeyH, evict only on real collision]

    D --> D1[Feed button always visible,<br/>only explains itself via error toast<br/>after a failed click]
    D1 --> F3[Fix: disabled state + tooltip<br/>full HP / no food, before the click]

    E --> E1[No persistent on/off setting;<br/>rows auto-fill on combat, close ~5s after]
    E1 --> F4[Fix: hint line explaining<br/>the auto-show/hide behavior]

    F1 --> G[This PR]
    F2 --> G
    F3 --> G
    F4 --> G
```

## Type of change

- [x] Bug fix
- [x] Feature: new functionality (discoverability hints, non-blocking)

## How was this tested?

- Commands:
  - `npx vitest run tests/keybinds.test.ts tests/pet_action_icons.test.ts tests/meters.test.ts tests/architecture.test.ts tests/i18n_completeness.test.ts` (all pass)
  - `npx tsc --noEmit` (clean)
  - `npm run build` (clean)
  - `npm run i18n:gen` (regenerated resolved bundles/status/sha256)
  - Confirmed `tests/server/http/characterization.test.ts` failures are pre-existing flake on
    unpatched `release/v0.25.0` (reproduced against the base commit in a scratch clone before this
    branch existed), unrelated to this change.
- Manual steps: drove the offline client through `scripts/pr_screenshots.mjs` to capture the HUD
  before/after this change.

### Pet-feed icon: disabled state + tooltip

Hunter, pet summoned, pet at full health, no reason to click the button, before this PR the icon
gave no explanation at all until you clicked it and got an error toast.

**Before** (icon always looks clickable, no tooltip)
Before: pet-feed icon

**After** (greyed out, tooltip explains why)
After: pet-feed icon with tooltip

### Threat/DPS meter: auto-show hint

**Before** (no explanation of the auto-show/hide behavior)
Before: meters window

**After** (hint line added)
After: meters window with hint

### Broad desktop HUD

For overall context (the affected elements are small conditional states within the existing HUD,
not a new window):

**Before**
Before: desktop HUD

**After**
After: desktop HUD

---

### Quality

- [x] **The full gate passes.** `tsc`, targeted tests, and `npm run build` are all green. Full
      `npm run gate` could not run `biome ci --changed` correctly in this environment because the
      fork's `origin/main` (biome's configured VCS default branch) trails `release/v0.25.0` by a
      large margin, causing `--changed` to diff against a stale base and flag unrelated
      pre-existing files; `npx @biomejs/biome check` run directly against every file this PR
      touches is clean (only pre-existing warnings, no errors). The repo's own pre-push hook
      (full biome + tsc + guard tests) ran and reported "pre-push QA floor: green" on this branch.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop verified via screenshot capture; this change adds
      no new mobile-specific surface (no new window, only a tooltip/disabled state and a hint
      line within the existing HUD), so no separate mobile-viewport capture was needed.
- [x] **Accessible.** The pet-feed button keeps its existing `aria-label`, adds `aria-disabled`,
      and its click handler no-ops while disabled.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Added to
      `src/ui/i18n.catalog/hud_chrome.ts` (English) and rendered via `t()`. All three new values
      are wordy, so the required five non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) are
      included in this same change per the M16 rule; `tests/i18n_completeness.test.ts` passes.
- [x] Numbers, money, dates, units, and percents go through the i18n formatters. N/A, no
      numeric/date/money values added.
- [ ] N/A. This PR does add player-visible strings (see above), but nothing sourced from
      `src/sim/` or `server/`.

### Hygiene

- [x] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is not enabled
      anywhere.
- [x] No hand-edited generated files; `src/ui/i18n.resolved.generated/*`,
      `i18n.status.summary.json`, and `i18n.resolved.sha256` were all produced by
      `npm run i18n:gen`.


jgyy:
Retargeted to `release/v0.26.0` per the #1857 release tracker (carried over from the v0.25.0 cycle, #1777). Heads up: GitHub reports a merge conflict against the new base, I'll rebase/resolve separately.

jgyy:

FernandoX7 (commented):
Thanks for the writeup, the investigation notes made this easy to follow. Two of the four items land: the `KeyH` collision between `targetFriendly` and `meters` is real and worth fixing, and the pet-feed disabled state is the right shape (folding the reason into `lastPetBarSig` so the bar only rebuilds when it flips is exactly the right call).

The keybind seeding fix does not reach the bug it names, though, and it drops bindings on the way. I verified each of the claims below by running the head and base `keybinds.ts` against a fake `localStorage`.

## Blocking

**1. `src/ui/i18n.resolved.sha256` is two lines, and that is the CI failure.**

The merge in e770ab7 kept both sides of the conflict, so the baseline now holds two hashes. `tests/i18n_resolved_equivalence.test.ts:36` reads it with `.trim()` and compares against a single sha, so it can never match. That is the entire "PR gate" failure; nothing else in the run is red. The base file is one line.

Related: this PR targets `release/v0.26.0` but the branch merged `release/v0.25.0`. Merge the actual base and re-run the i18n build so the generated trio (`pending.ts`, `i18n.status.summary.json`, `i18n.resolved.sha256`) is regenerated together. The remaining conflicts against the base are only those three files, so it is mechanical.

**2. The legacy-seed guard cannot fire for the bug it describes. `src/game/keybinds.ts:305-312`**

`isDefaultOfOtherActionBarSlot` asks "does the stored code equal some other Action Bar slot's *current* default". The real corruption is the mirror image. The old convention comes from the interface overhaul (1d2678f58), which shipped `SLOT_DEFAULTS[10] = ['KeyQ', 'Minus']` and `[11] = ['KeyE', 'Equal']` with `strafeLeft`/`strafeRight` on `defaults: []`, and was then reverted by #1788. So an affected profile holds `slot10: ["KeyQ", "Minus"]` and `strafeLeft: [null, null]`, and `KeyQ` is not any slot's current default, so the predicate is false and the import goes through untouched. Seeding a fresh character from that blob gives identical results on this branch and on `release/v0.26.0`:

```
strafeLeft  -> [null, null]        slot10 -> ["KeyQ","Minus"]
strafeRight -> [null, null]        slot11 -> ["KeyE","Equal"]
pressing Q does: slot10            pressing E does: slot11
```

There is a second, independent reason it is inert. Per-character scoping landed in b97b5730 (2026-06-21), three weeks before that layout (2026-07-10), and `save()` only ever writes the scoped key: `src/main.ts:905` is the only construction site and always passes a non-empty scope (`char:<id>` or `offline:<class>:<name>`). Every affected profile therefore lives under `woc_keybinds:char:<id>`, where `isLegacySeed` is false and the guard never runs at all. The bare `woc_keybinds` blob is frozen at pre-scoping content and cannot contain the overhaul layout. A fix has to run on the scoped-profile path, as a one-time repair keyed on that signature.

The new test at `tests/keybinds.test.ts:416` pins `strafeLeft: ['Digit1']`, a layout that never shipped, so it goes green while the real fixture behaves identically on both refs.

**3. The guard reverts deliberate remaps and silently drops bindings. `src/game/keybinds.ts:512-537`**

`if (!shared && claimed.has(v)) continue;` runs before the guard, so the fallback can claim a code that the next action's stored value needs, and that action then gets nothing. A legacy blob with a deliberate `slot0`/`slot1` swap:

```
stored   slot0: ["Digit2"]       slot1: ["Digit1"]
base     slot0 -> ["Digit2"]     slot1 -> ["Digit1"]      (swap preserved)
head     slot0 -> ["Digit1"]     slot1 -> [null, null]    (reverted, and Digit1's action is now unbound)
```

Same class: a blob with `strafeLeft: ["Digit1"]` and `slot0: ["KeyG"]` loads with strafe forced back to `KeyQ` and `Digit1` bound to nothing. An off-by-one bar (Attack moved to Backquote, abilities pulled onto 1 to 0) loses half the action bar. The guard cannot tell an old convention from a deliberate remap, because the blob carries no version. And the same stored map read as the character's own profile keeps the swap, so one player gets different layouts on different characters, which is the bug this PR set out to kill.

## Should-fix

**4. `Shift+KeyH` never reaches the players who actually hit the collision. `src/game/keybinds.ts:505`**

On base, a fresh profile keeps both `targetFriendly` and `meters` on `KeyH`; the uniqueness sweep hands `KeyH` to `targetFriendly` (Targeting precedes Interface), and the next `save()` persists `meters: [null, null]`. A stored `null` always wins over the default, so those characters stay unbound after this PR and only brand-new profiles get `Shift+KeyH`. A repair looks safe: `clear()` has no caller in `src/`, and the capture flow treats a null capture as cancelled rather than as an unbind (`src/ui/options_window.ts:1560`), so a `null` in a stored profile can only be loader eviction, never player intent.

**5. The feed button's cancel toggle dies exactly when it is needed. `src/ui/hud.ts:6478` and `:6636`**

`active` and `disabled` are computed independently, and `runClickAction` early-returns on `disabled`, so the cancel branch becomes unreachable while a feed is pending. It is reachable in normal play: an out-of-combat pet regenerates 2% of max HP every 40 ticks (`src/sim/pet/pet_ai.ts:80-82`), so a hunter who opens feed mode with a lightly hurt pet watches it top itself off while browsing the bags. The button then renders gold-bordered with `aria-pressed="true"` and does nothing on click. Escape and closing the bags still clear it, so it is not a soft-lock, but the toggle the comment advertises is gone. `disabled: feedState.disabled && !this.pendingPetFeed` covers it.

**6. The disabled feed button loses its accessible name. `src/ui/hud.ts:6452` and `:6614`**

`addButton` assigns its `title` argument to both `btn.title` and `aria-label`, and the PR passes the reason there, so a screen reader announces "Pet is at full health, button" with no action name. The visible tooltip already does this correctly (it keeps "Heal Pet" as the title and the reason as the description); only the name argument needs to stay `t('hud.pet.healPet')`, with the reason appended or carried in a separate opt.

**7. The meters hint concatenates two translated sentences. `src/ui/meters.ts:298`**

`${t('hud.meters.noCombat')} ${t(metersEmptyStateHint())}` assembles one rendered sentence from two catalog domains, glued with a hardcoded space. Rendered text comes from a single `t()`, never a concat. The house shape already exists: `calendar_window.ts` paints its note line as its own element with its own key. Give the hint a `.mt-hint` sibling and leave `subEl` carrying `noCombat`.

**8. The hint shows in every empty state, and the copy is wrong in most of them. `src/ui/meters.ts:296`**

The empty branch sits above the tab check, so the Threat tab also claims rows appear "once your party deals damage or healing", and the history and All (session) views say "this segment closes a few seconds after combat ends", which means nothing for a finished segment. Gate it to the live current view.

**9. Mobile was skipped, and this is the change that needed it.**

`#meters-window` on touch is `max-height: min(170px, ...)` with `overflow: hidden`, tightening to `min(135px, ...)` at `min-width: 128px` (`src/styles/hud.mobile.css:1931` and `:3881`). `.mt-sub` stays 10px with no mobile override, so a 150-character hint reflows to roughly six lines in a ~120px column and can clip with no scroll affordance. Worth an actual mobile capture rather than reasoning about it.

## Nits

- `src/ui/meters.ts:22`: `metersEmptyStateHint()` is a zero-argument function returning a constant, and `tests/meters.test.ts:150` only asserts it returns that constant. The `TranslationKey` return type already proves the key exists, so the assertion cannot fail. Inline the key and drop both.
- `src/styles/hud.css:1362`: `.pet-btn.disabled` repeats `.pet-btn.cooldown` verbatim; a selector list says it once. Note `.pet-btn.active` and `:hover` still gild a disabled button, which is the visual half of finding 5.
- `src/ui/pet_action_icons.ts`: it now holds a pure core, but the file is in neither `UI_PURE_CORES` nor the bare-named cross-check in `tests/architecture.test.ts`, so nothing enforces its purity. It is clean today (it imports nothing); add it so it stays that way.
- `tests/pet_action_icons.test.ts`: the `petMaxHp > 0` guard is unpinned (delete that clause and every test still passes). `tests/keybinds.test.ts:444` asserts `not.toBe('KeyH')` rather than the literal `Shift+KeyH`, so any other default satisfies it.
- `src/ui/hud.ts:6497`: `disabled` gates `runClickAction` but not the context-menu or touch-hold branches. Harmless today (the feed button has neither), but the next button passed both flags will fire its long-press action while greyed out.

## Notes, no action needed

- The guide still tells players to press Z for the meters (`src/ui/i18n.catalog/guide.ts:408` and `:495`), left over from the reverted overhaul layout. That is already wrong on base, so it is not yours, but this PR owns the default and Shift+Z is now the Book of Deeds. I will take the reword and the locale side of it.
- The reformatting churn in `meters.ts` is forced, not gratuitous: the file carries pre-existing format and import-sort debt, and the changed-files Biome gate fails the moment you touch it. Nothing to do there.

The pet-feed and meters halves are close, mostly copy and wiring. The keybind half needs to be re-aimed at the scoped-profile path, or dropped, keeping just the `Shift+KeyH` default plus a repair for the profiles the collision already emptied.

jgyy:
Thanks for the incredibly thorough review, @FernandoX7. Every point landed; here is what changed.

## Blocking

**1. i18n trio / wrong base.** Merged the real base (`release/v0.26.0`) and regenerated the generated trio rather than hand-resolving it: `npm run i18n:gen` + `npm run i18n:hash -- --write`. `i18n.resolved.sha256` is back to a single line, and a second `i18n:gen` leaves the tree clean (the determinism proof). `tests/i18n_resolved_equivalence.test.ts` is green.

**2 + 3. Keybind repair re-aimed at the scoped path, guard dropped.** You were right that the old `isDefaultOfOtherActionBarSlot` guard never fired (wrong signature direction, and it only ran on the bare legacy key that scoping froze), and that it reverted deliberate remaps and dropped their bindings. I deleted that guard entirely and replaced it with a pure, signature-keyed repair in a new module `src/game/keybinds_repair.ts`, run on every load (so it reaches the scoped `woc_keybinds:char:<id>` profiles):
- Signature A (Q/E overhaul): fires only on the exact four-key fingerprint (`slot10` primary `KeyQ`, `slot11` primary `KeyE`, `strafeLeft` and `strafeRight` both empty), and just deletes those four keys so they re-seed to the current defaults (Q/E strafe, Minus/Equal slots).
- Signature B (KeyH collision): `meters` empty AND `targetFriendly` on KeyH deletes `meters` so it re-seeds to `Shift+KeyH`.

Because each repair matches an exact signature and only deletes those keys, a deliberate `slot0`/`slot1` swap (or any other remap) is left verbatim. New pure tests in `tests/keybinds_repair.test.ts` cover both signatures plus the swap/partial-match negatives; the never-shipped `strafeLeft: ['Digit1']` fixture in `tests/keybinds.test.ts` is replaced with the real corrupted signature on a scoped profile, a scoped meters re-seed test, and a scoped-swap-preserved test.

## Should-fix

**4.** Covered by Signature B above (scoped `meters: [null,null]` + `targetFriendly` KeyH re-seeds to `Shift+KeyH`); confirmed `clear()` has no caller and null capture is a cancel, so a stored null is only loader eviction.

**5.** `disabled: feedState.disabled && !this.pendingPetFeed`, so a pending feed stays cancellable even after the pet tops itself off.

**6.** The feed button's accessible name stays `t('hud.pet.healPet')`; the disabled reason now lives only in the rich hover tooltip, so a screen reader announces the action, not the reason.

**7.** The hint is its own `.mt-hint` element with a single `t('hudChrome.meters.autoShowHint')` call; `.mt-sub` keeps `noCombat`. No more concatenation.

**8.** The hint is gated to the live current segment (`viewIdx === 0 && tab !== 'threat'`), so it no longer shows on Threat, History, or All (session).

**9. Mobile.** Honest disclosure: I reasoned about this rather than capturing live (no capture tooling in this environment). The game is landscape-locked and the hint only renders in the empty current-segment state, so by the max-height math it fits, but to remove the clip risk entirely I switched the mobile `#meters-window` from `overflow: hidden` to `overflow-y: auto` (x still hidden), so a long hint scrolls instead of clipping, and gave `.mt-hint` compact 10px/1.35 styling. Worth a real capture on your side if you want to confirm.

## Nits
- Inlined the `hudChrome.meters.autoShowHint` key at the call site and deleted `metersEmptyStateHint()` + its trivial test.
- `.pet-btn.disabled` / `.pet-btn.cooldown` combined into a selector list; added a rule so a disabled button drops the active/hover gilding.
- Registered `src/ui/pet_action_icons.ts` in both `UI_PURE_CORES` and `BARE_NAMED`.
- Pinned the `petMaxHp > 0` clause with a `maxHp === 0` case; tightened the meters default assertion to the literal `Shift+KeyH`.
- Gated the pet-button context-menu / touch-hold branches on `disabled` too.

Left the guide Z-to-Shift+Z reword for you as noted. `tsc` clean; targeted vitest (keybinds, keybinds_repair, meters, pet_action_icons, architecture, i18n equivalence, mobile-window, localization S3) all green; changed-files biome shows only pre-existing warnings.

FernandoX7 (approved):

PR #1859, opened 2026-07-13
feat(balance): Monte Carlo combat balance simulator for PvP and PvE

EnriqueGF:
## Summary

Adds the adversarial third layer of the balance framework (see `docs/design/spell-balance-framework.md` and the new `docs/design/combat-balance-simulator.md`): a Monte Carlo combat simulator that answers, with confidence intervals, which specs and talents actually WIN fights, in PvP and PvE, by running thousands of real fights on the real deterministic `Sim`.

`npm run sim:balance` runs a full-27-spec PvP duel round robin (through the real duel system, so the 1 hp guard, PvP crowd control diminishing returns, WARFARE gear multipliers, and the five second rule all apply) plus solo PvE kill benchmarks against live mob templates that fight back. It emits `balance_report.md` (digest) and `balance_report.json` (every duel record and the full matchup matrix) under `tmp/balance_sim/`. An opt-in `--knockout` mode re-runs a spec with one talent node removed at a time, same seeds and opponents, attributing winrate to individual talents.

Methodology highlights:

- Real code, no reimplementation: fighters are real players built through `addPlayer`, `setPlayerLevel`, `applyTalents`, and `equipItem`; metrics come from the `SimEvent` stream.
- Uniform inputs: talent builds are generated per spec by the same greedy validated fill, gear is best-in-slot by per-archetype stat weights, one shared rotation policy drives every spec. Policy bias applies to all specs equally, so the ranking between specs is the signal.
- Deterministic: one fresh seeded `Sim` per bout, seeds derived by pure integer mixing. Same config reproduces the same report byte for byte.
- Controlled arena: a minimal one-zone `WorldContent` with a flattened terrain stamp (cheap ticks, no wandering mobs, identical arena under every seed), swapped via `setActiveWorldContent` and restored after every bout.
- Statistics: Wilson 95% intervals on winrates, draws count half, brokenness score composites the PvP winrate z-score with the inverted PvE kill-time z-score.

First data harvest (10 trials per matchup, level 20 best-in-slot): retribution paladin sits at an 85.2% PvP winrate with a Wilson low bound of 80.4% (decisively dominant), followed by protection paladin (79.8%) and beast mastery hunter (76.3%); enhancement shaman tops the composite ranking; feral and restoration druid land underpowered and fire mage weak. Full tables in the tool output.

The simulator also surfaced two real engine findings on its first full run:

- A live crash in `updateAuras` (`src/sim/combat/auras.ts`): a DoT tick that fires the duel 1 hp guard makes `endDuel` bulk-clear the opponent's auras from the array the tick loop is iterating, and the next index read crashes the whole tick. Fixed here, test-first (`tests/auras_reentrant_removal.test.ts`, both arms verified red without the fix), byte-identical on the non-reentrant path (`tests/parity` green without regeneration).
- Duel non-lethality can be violated: a shot already in flight when the duel ends can land on the same tick and kill the loser for real. About 15% of simulated duels hit this. The report counts these as an anomaly; left as a follow-up issue rather than widening this PR.

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run gate` green (all 10 steps, full suite; two earlier runs flaked on unrelated timing-sensitive files, `tests/server/http/parity.test.ts` and `tests/mail.test.ts`, each verified green in isolation and green in the final uncontended gate run).
  - `npx vitest run tests/balance_sim_stats.test.ts tests/balance_sim_builds.test.ts tests/balance_sim_bout.test.ts tests/balance_sim_experiment.test.ts tests/auras_reentrant_removal.test.ts tests/parity`: 200 passing.
  - `npm run sim:balance -- --trials 10`: full 27-spec experiment end to end (about 165s), report attached numbers above.
- Manual steps: probed individual duels and PvE bouts per class for rotation sanity (casts, pets, control uptime, mana sustain); verified byte-determinism by re-running identical configs; red-checked both arms of the aura fix by reverting each hunk.
- Review passes: architecture review (determinism, seam, parity impact), QA gate checklist, and a test-decisiveness audit were run over the diff; all findings addressed (deterministic outcome pins, identity-splice coverage, NUL-byte cleanup in `stats.ts`) or documented (the bounded bystander double-visit note in `updateAuras`).

### Quality

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: headless dev tooling, no UI surface.
- ~Accessible.~ N/A.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (CLI and report output are dev-channel English; the sim fix adds no emits).

### Hygiene


EnriqueGF:
## Exhaustive run: what is broken, by the numbers

Ran the simulator at full depth to back the PR with data: **40 trials per matchup** (14,040 PvP duels, 2,160 PvE bouts) plus the **talent knockout pass** (8 duels per arm, same seeds with and without each talent node). Wall time about 11 minutes, byte-for-byte reproducible with `npm run sim:balance -- --trials 40 --knockout --knockout-trials 8`.

Balance report chart

### Headlines

- **Retribution Paladin is the PvP outlier: 87.1% winrate over 1,040 duels** (Wilson 95% band 84.9% to 89.0%). Protection Paladin (77%) and Beast Mastery Hunter (77%) follow. This is confidence-interval-backed dominance, not sample noise.
- **Enhancement Shaman tops the composite** (+1.36 sigma): top PvE damage (257 dps, 1.6s kills) plus a 72% PvP winrate.
- **Druids are in trouble**: Restoration Druid lands last on the composite (-1.79 sigma, 11% PvP winrate, worst PvE clear), with Feral and Balance also deep in the weak band. Fire Mage is the weakest mage arm despite the Pyroblast pass.
- **Talent attribution** (knockout): `arms_deep_wounds`, `arms_imp_overpower`, and `arms_tactical_mastery` each buy Arms Warrior +12.5 points of winrate; `holy_flash_focus` and `holy_imp_holy_light` do the same for Holy Paladin. Rankings for all measured nodes are in the report JSON.
- **Anomaly counter**: 2,131 of 14,040 duels (15.2%) ended in a REAL death despite the 1 hp duel guard (an in-flight shot lands on the same tick the duel ends). Consistent with the 10-trial run; tracked as the PR's follow-up issue.

The rankings are stable between the 10-trial and 40-trial runs (same top cluster, same bottom cluster), which is the convergence signal we wanted before trusting the tool for balance passes.

Full tables ship in the tool output (`balance_report.md` / `balance_report.json`); the chart above is rendered from the JSON.

PR #1860, opened 2026-07-13
feat(items): Hit rating + combat-rating itemization, and PvP gear rebalanced to warfare

ryan-foo:
## What

Three linked combat changes plus itemization:

1. **Softened the above-level miss/resist curve.** On Heroic (+3) melee missed ~44% and spells resisted ~43%, which felt like a coin flip. The above-level penalty table is retuned to a capped `[0, 7, 14, 21]` (saturating past the last entry), so melee miss now tops out at ~26% and spell resist at ~25% (base +0 stays 5% / 4%). Leveling against yellow/orange (+1/+2) mobs is smoother too.
2. **Added a Hit rating primitive.** Crit and haste ratings already existed on gear; Hit did not. `hitRating` on `BaseItemDef` + `SetBonusEffect`, an entity `hitBonus`, folded in `recalcPlayerStats`, and subtracted from miss (`swingMissChance`) and spell resist. It mirrors to `ClientWorld`; no wire change. Determinism preserved (ungeared draws are byte-identical, so the parity gate holds).
3. **Itemized hit/crit/haste** across the marks jewelry, the ilvl-31 heroic set, and the ilvl-33/37 raid variants (dual-rating), plus a Hit grant on the two weak T2 bleed 4-set bonuses.
4. **Rebalanced PvP WARFARE gear:** every piece weights its budget to warfare (primary stats at 60% of the ilvl-28 budget, warfare unchanged, weapon DPS and armor mitigation kept), so PvP gear never out-stats same-tier PvE gear.

With the ~26% cap, Hit rating is a meaningful but no longer mandatory lever (it closes the final ~26% on Heroic toward 0). Spec: `docs/prd/combat-ratings-and-jewelry.md`.

## Testing

`tests/combat_rating.test.ts`, `tests/pvp_honor_gear.test.ts`, `tests/mob_hit_floor.test.ts`, and `tests/sim.test.ts` cover the capped curve, the ratings, and the PvP rule; parity goldens regenerated for the deliberate curve change; architecture + i18n guards green; `tsc` clean. Base: `release/v0.26.0`.


Rubsey (commented):
This is a carefully built change and most of it verifies clean. The parts I checked hardest all held up: the golden re-mint is exactly scoped to the curve retune (all five changed scenarios provably contain above-level combat in a changed bracket, the diffs are pure outcome flips on the same seed-determined draw sequence, equal-level goldens like arena_1v1 are untouched, and the files byte-match the harness mint path, no hand edits), the hitBonus plumbing preserves rng draw order and count and is byte-identical at 0 rating, the PvP 60% rule holds computed-vs-computed for all 31 WARFARE pieces with warfare ratings, armor, and weapon DPS untouched, the one-rating-per-piece rules hold across heroic_loot, heroic_vendor, and the zone3 seeds, and the ratings stay invisible to the primary-stat budget sweeps. I ran the changed suites plus tests/parity, tests/sim, the architecture and i18n guards, and tsc at the head: all green. (CI has not run on this PR because it is conflicting, see the end.)

Findings:

- blocking: the online character sheet always shows 0 Hit Rating. The new row reads `p.hitRating` (src/ui/char_window.ts:97 via src/ui/hud.ts:4790), but the self snapshot only sends crit and haste (`crat`/`hrat` at server/game.ts:4705) and src/net/online.ts:2034 only decodes those two; the blankEntity addition gives shape parity, not values. The comment at online.ts:2032 says these rating fields are wired precisely so the online sheet shows them instead of the blankEntity 0, and hit needs the same treatment: send it in the self payload and decode it. Combat itself is unaffected (it resolves server-side, so `hitBonus` can stay server-local), this is purely the paper doll and its stat tooltip, but it means the feature's headline stat reads 0 in the online game.

- should-fix: Sunder/Expose Armor never benefits from Hit. src/sim/combat/effect_dispatch.ts:1182 rolls `meleeMissChance(p.level, target.level)` directly, so the one ability miss roll in the game ignores `hitBonus`, contradicting the new stat description ("reducing how often your attacks miss"). Routing it through `swingMissChance(p, target)` keeps the single draw and picks up the subtraction (and the goldens are unaffected since their actors are ungeared).

- should-fix: the PRD shipped in this PR disagrees with the code it ships with in four places. (1) docs/prd/combat-ratings-and-jewelry.md:105 lists "changing the miss/resist curves" as a non-goal, and the power analysis at lines 170 to 171 (44% miss falling to ~22%, "Heroic stays harder than +0") is built on the old table; with the retune plus the floor at 0 a hit-stacked kit can drive Heroic miss to 0, the opposite of that target. (2) Line 128 specifies `max(0.005, ...)` as the physical floor; types.ts:3232 floors at 0 and the test pins 0. (3) Line 283 says a haste-primary raid piece pairs with +crit; heroic_variants.ts:54 pairs every non-hit primary with +hit (latent today, no haste-primary base exists). (4) Line 288 specifies Heartwood as +haste primary, +crit secondary; the code gives the rating-less legendary the hit default, so heroic_deathless_heartwood ships hit 70 / crit 30. That last one deserves a decision rather than a doc edit: it puts the game's largest Hit allowance on the healer-facing staff, in tension with this PR's own rule that healer-facing pieces never take Hit (heroic_loot.ts:39). Either special-case the legendaries in applyRaidVariantRatings or update the PRD deliberately.

- should-fix (tests, all cheap to add): (1) the set-bonus arm of `e.hitRating = bonusHitRating + setEff.hitRating` (entity.ts:479) is only tested at the pure aggregateSetBonuses resolver; dropping the `+ setEff.hitRating` term leaves every test green. Crownforged has four real pieces, so the end-to-end equip test exists for crit (item_sets.test.ts:207) and can be mirrored for hit. (2) `effectiveSpellHit`'s cap at 1 (the `Math.min`) is never exercised; deleting the clamp stays green. (3) The retuned +1 entry (2.5 to 7) has no pin, only +2 and the +3 cap do (sim.test.ts:168).

- nit: three comments still describe the old curve: types.ts:3197 says resist tops out at ~14% and types.ts:3205 says miss tops out at ~15% (the table 15 lines above and the tests both say 25%/26%), src/sim/social/chat_readouts.ts:171 still describes the removed +3 "cliff", and src/sim/content/delves/collapsed_reliquary.ts:128 still cites the old 44% miss math in its tuning rationale.

- nit: the item-compare "If you equip this" deltas (src/ui/item_compare.ts:22) compare warfare rating but not hit/crit/haste, so the heroic tier's stated differentiator is invisible when comparing two pieces. Fine as a follow-up since the file predates this PR, but worth doing while the ratings are fresh.

On the merge conflict: the only true conflicts against release/v0.26.0 are the generated trio (src/ui/i18n.resolved.generated/pending.ts, src/ui/i18n.resolved.sha256, src/ui/i18n.status.summary.json). That is mechanical regen churn, not a design problem: merge the base and re-run the i18n build to regenerate them. CI will only run once that is resolved, so the local runs above are the current verification.

jgyy (commented):
This is backed by a real PRD (`docs/prd/combat-ratings-and-jewelry.md`) with derivations, not an ad hoc number, and Hit Rating is modeled directly on the existing crit/haste rating conversion (`HIT_RATING_PER_PCT = 10`), so it's consistent with how the other combat ratings already work. Determinism is untouched: the resist/miss rolls still draw exactly one `rng.chance(...)` call each, hit rating only changes the probability argument.

One thing worth flagging before merge: the above-level miss/resist table changed from `[0, 2.5, 14, 39, 80]` to `[0, 7, 14, 21]`, which is a global retune of PvE difficulty at +2/+3 encounter levels (Heroics land around 26% melee miss / 25% resist at +3 now, down from a much steeper curve), not just PvP itemization. The PRD does call this out and justify it as fixing an unintentionally punishing Heroic miss rate, so it's disclosed, but the PR title only mentions itemization and PvP gear, and a reviewer skimming just the title could miss that every above-level PvE fight got easier. Worth a should-fix: either mention the PvE curve change in the title/description, or split it out, so it's clear this is an intentional, signed-off difficulty change and not incidental scope creep.

Everything else checks out: no world_api parity gap (hitRating/hitBonus follow the same recompute-not-wire pattern as crit/haste), i18n is English-keyed correctly, and `tests/combat_rating.test.ts` plus the touched golden fixtures cover the rating math, the Heroic reduction, and the zero-hit no-op.

Thanks for writing this up with a PRD first, that made it much easier to verify the formula was principled rather than guessed at.

ryan-foo:
@Rubsey the merge conflict is resolved (merged release/v0.26.0 in): the only true conflicts were the generated i18n trio, regenerated deterministically via the i18n build, plus two parity goldens (drowned_litany, mob_lifecycle) that now fold the above-level miss/resist retune into their event streams. Each changed one line (the events hash) with rng.draws, rng.digest and the state hash unchanged, so draw order and world state stay byte-identical. CI is green.

All your review findings are addressed in the pushed commits:
- blocking: the self snapshot now sends and decodes hitRating (crat/hrat/hirat), so the online sheet reads the real value.
- Sunder/Expose now routes its miss roll through swingMissChance, so Hit applies (parity byte-identical, ungeared actors).
- the Heartwood legendary no longer gets the Hit default: applyRaidVariantRatings treats a spell-facing base as never taking Hit and pairs its throughput ratings, so it ships haste 70 / crit 30, matching the PRD.
- added the three pins: the end-to-end Crownforged Hit equip test, the effectiveSpellHit clamp, and the retuned +1 curve entry.
- PRD updated (curve retune is now an explicit enabling change; power analysis, the 0 floor, and the spell-facing pairing rule) and the stale curve comments fixed.

The item-compare hit/crit/haste deltas are left as the follow-up you flagged. Ready for merge when you are.

Rubsey (commented):
Re-review at b8bd4ee01. I verified each finding from my earlier review against the current head, plus the new heroic-weapons commit. Everything I flagged is fixed, and the fixes are the right shape.

- The blocking wire gap is closed: the self payload now carries `hirat` next to `crat`/`hrat` (server/game.ts:5016), ClientWorld decodes it (src/net/online.ts:2103), `blankEntity` carries both `hitRating` and `hitBonus`, and tests/snapshots.test.ts:167 pins the field. The online sheet reads the real value.
- Sunder/Expose now rolls `swingMissChance(p, target)` (src/sim/combat/effect_dispatch.ts:1186), so the one ability miss roll picks up Hit. Parity stayed byte-identical as claimed.
- The Heartwood default is gone: `applyRaidVariantRatings` treats a rating-less spell-facing base as throughput-only (haste 70 / crit 30), keeps Hit only on an authored caster-DPS seed, and combat_rating.test.ts:213 pins it.
- The curve resolution in 9e341e193 is better than what I asked for: restoring the established +1 (2.5) and +2 (14) entries and only capping +3 and beyond at 21 means the "do not change normal difficulty" non-goal is now actually true, not just reworded. The PRD matches the shipped `[0, 2.5, 14, 21]` table, the 0 floor, and the pairing rules.
- All the missing pins landed: the end-to-end Crownforged 4-set test (tests/item_sets.test.ts:216, 20+20+60 on `e.hitRating`), the `effectiveSpellHit` clamp (tests/combat_rating.test.ts:95), the +1 entry (tests/sim.test.ts:175), and the new ladder sweep by live item level that catches the three direct raid weapons. The geared/ungeared parity pair pinning draw count and digest equality in coverage.test.ts is a genuinely good pin.

Since the PR shows CONFLICTING no CI ran at this head, so I ran it locally in a clean worktree: combat_rating, item_sets, mob_hit_floor, snapshots, pvp_honor_gear, stat_tooltip (329 tests), the full parity suite (169), sim/delves/tolling_bells (246), architecture and the i18n guards, and `tsc --noEmit`. All green.

One leftover:

- nit: src/sim/social/chat_readouts.ts:172 says the table is `[0,7,14,21]` but the shipped table is `[0, 2.5, 14, 21]`. Looks like a leftover from the intermediate retune; the PRD and types.ts comments are already correct.

On the conflict: the base moved again (#1913 taunts, #1912 tank cooldowns, #1845 metrics) and there is exactly one true conflict, src/sim/combat/casting_lifecycle.ts around line 849. It is a clean union: keep the base's `isTaunt` const and `!isTaunt &&` guard, and add your `src.hitBonus` argument to `isSpellResisted`. None of the three new base commits touch tests/parity, so your re-minted goldens should hold; re-run the parity suite after the merge to confirm.

With that mechanical resolution and the comment nit, this is ready. The item-compare deltas stay a follow-up as agreed.

Issue #2, opened 2026-06-12
Continous missing file or or bad GLB errors when trying to play offline.


JaySchenk:
The error message itself is wrapped inside a div from which the text is select and copy protected. Making it more obtuse to send in the issue. If you need any additional information let mek now.

Error message displayed on the frontend:
[...]

PR #1861, opened 2026-07-13
feat(hud): make the action-bar Attack button removable

Blaine1705:
## Summary

The action bar's slot 0 has always been a fixed auto-attack toggle, so the first
keybind (`1`) could never hold a real ability. This makes the Attack button
**removable**, freeing slot 0 and its keybind for a normal assignable action:

- New **Show Attack Button** toggle in the Interface options (on by default, so the
  classic behavior is unchanged).
- **Right-click the Attack button** to remove it from the bar (the same "remove it
  from the bar" gesture as any slot). The freed slot then accepts a drag like any
  other slot and its key casts whatever is assigned; a second right-click clears it.
- The Options toggle restores Attack at any time.

The slot's assignment is backed by its own persisted action, shared across forms, so
it never disturbs the `hotbarActions` layout. `ActionBarSlotDescriptor.isAttack`
becomes a live accessor (like `ability()`/`item()`), so the bar reflects the toggle
with no rebuild.

This is intentionally independent of the existing **Auto-Attack on Ability Use**
setting: that one still gates whether using an ability also engages white swings, and
is untouched here.

## Related issues

N/A.

## Type of change

## How was this tested?

- Commands:
  - `npx tsc -p tsconfig.json --noEmit` (green)
  - `npx vitest run` (full suite, green)
  - `npm run build` (green)
  - `npx @biomejs/biome check --write <changed files>` (clean, no format diffs)
  - `npx vitest run tests/i18n_completeness.test.ts tests/localization_fixes.test.ts` (green)
  - Note: `npm run gate` cannot complete on this machine (no local FFmpeg/ffprobe for
    the SFX asset step); every other gate stage was run individually as above, and the
    pre-push QA floor is green.
- Manual steps:
  - Toggle Show Attack Button off in Interface options: slot 0 stops being Attack and
    accepts a dropped ability/item; its key casts the assigned action.
  - Right-click the Attack button: removes it; right-click the freed slot: clears the
    assigned action; toggle back on: Attack returns.
  - Confirmed the separate "Auto-Attack on Ability Use" setting still behaves as before.

## Screenshots / recordings

To follow (desktop: Attack button present, then removed with an ability assigned to
slot 0).

---

### Quality

- [x] Decisive tests added for the changed behavior (`action_bar_view` covers slot 0
      honoring the live `isAttack()` accessor on/off; the accessor migration across the
      bar, mobile ring and consumable-bar descriptors and their tests). Full `vitest`,
      `tsc`, and `build` are green; `gate` blocked only by missing local FFmpeg.

### Cross-platform

- [x] Desktop behavior verified. On touch, slot 0 stays the fixed Attack (the
      right-click removal is a desktop gesture); mobile descriptors keep `isAttack()`.
- [x] Keyboard-operable: the freed slot's key casts its assigned action; the tooltip
      documents the right-click removal.

### Localization (i18n)

- [x] New player-visible strings are `t()` keys in the English catalog
      (`hudChrome.options.showAttackButton`, `abilityUi.actionBar.attackRemoveHint`).
      Both are wordy by M16, so the five non-Latin fills (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU)
      are supplied in the overlays; no other overlay was edited.
- [x] N/A: no numbers/dates/money added.
- [x] N/A: no new `src/sim`/`server` player text; `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets, `.env`, or `ALLOW_DEV_COMMANDS` changes.
- [x] Generated i18n files are regenerated via `npm run i18n:gen`, not hand-edited.


EnriqueGF:
:) Thanks for the contribution, this is a genuinely nice quality-of-life feature and the accessor migration (isAttack -> isAttack()) is clean and well tested at the pure-view layer. I reviewed it against release/v0.26.0 in a worktree and ran the guards. The base branch is correct (latest release, rebased on its tip), tsc is clean, and the architecture and S3 localization guards pass. A few things need to be addressed before this is mergeable.

## Blocking

1. **CI fails: the i18n resolved-hash baseline was not regenerated.** `tests/i18n_resolved_equivalence.test.ts` fails (2 assertions). The two new keys legitimately change the resolved 21-locale table, so the committed baseline `src/ui/i18n.resolved.sha256` must be re-generated in the same change (it is still `178968f...`, but the table now hashes to `9d33974...`). The generated per-locale bundles themselves reproduce byte-identically, so the only missing step is: run `npm run i18n:hash` and commit `src/ui/i18n.resolved.sha256`. Note the PR description states "npx vitest run (full suite, green)", which is not the case here; this guard is red.

## Should fix

2. **The new slot-0 wiring in `src/ui/hud.ts` is essentially untested.** Only the pure `action_bar_view` render classification is pinned. The substantive new behavior has no decisive coverage and would regress silently: `actionForSlot(0)` resolution, the localStorage round-trip (`loadAttackSlotAction`/`saveAttackSlotAction`/`attackSlotKey`), right-click remove then clear (contextmenu), drag-drop assign plus the source-slot MOVE vs assign-only arms, and the keybind cast fall-through. Per module-first, the pure part (parse/validate/serialize the attack-slot action and the assign+clear-source resolution) should be extracted into the existing `hotbar.ts` family and unit-tested, leaving `Hud` a thin consumer.

3. **Family-reuse:** `loadAttackSlotAction()`/`saveAttackSlotAction()` reimplement, more weakly, the existing tested helpers `parseHotbarAction`/`encodeHotbarAction` in `src/ui/hotbar.ts`. The inline check validates only `parsed.type`, not that `id` is a string, and (unlike `parseHotbarAction`) never confirms the ability/item still exists, so a stale or unknown persisted id resolves as a live slot-0 action. Route load through `parseHotbarAction` with the same `abilityExists`/`itemExists` predicates the slot map already uses, and save through `encodeHotbarAction`.

4. **Screenshots:** this is a visual change and the PR body says "To follow". The repo requires before/after screenshots (desktop and mobile where relevant) committed under `docs/screenshots` and referenced from the PR body. Please add them (the `pr-screenshots` skill has the capture recipe).

## Minor (non-blocking, worth a decision)

5. Once an action is assigned to the freed slot 0 it cannot be dragged OUT to another slot (no `dragstart` handler on slot 0), only right-click-cleared, unlike every slot >= 1. Small asymmetry.

6. The mobile action ring keeps a fixed Attack (`isAttack: () => true`) regardless of the setting, so the preference does not mirror to touch. Likely intentional desktop scoping, but a one-line comment stating that would help.

i18n itself looks correct: both strings are `t()` keys in the right catalog domains, the non-Latin M16 fills (zh_CN/zh_TW/ja_JP/ko_KR/ru_RU) are supplied and no other overlay was touched, with the Latin locales left pending as the PR-tier gate allows.

Happy to re-review once the baseline is regenerated and the coverage/screenshots are in. Nice work overall.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.26.0` and ran the checks locally:

- `npx tsc -p tsconfig.json --noEmit`: green.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/i18n_completeness.test.ts tests/action_bar_view.test.ts tests/action_bar_painter.test.ts tests/options_view.test.ts tests/mobile_action_ring_painter.test.ts tests/hud_perf_budget.test.ts`: green (i18n_completeness only after `npm run i18n:gen`, which is the normal `pretest` step).
- `npm run i18n:gen` leaves the tree clean, so the committed resolved bundles / hash / summary are fresh, and the five non-Latin fills satisfy the M16 exception. No overlay was touched beyond those. No sim/server changes, no new IWorld members, so determinism, tri-host parity and server authority are not in play.

The design itself is nice: `isAttack` as a live accessor is the right call (no rebuild on toggle), the assignment is backed by its own persisted key so `hotbarActions` layout is untouched, and `loadAttackSlotAction()` uses the same `slotMapKey('normal')` basis that `loadSlotMap()` already uses in the constructor, so class/character scoping is consistent.

One blocker, plus two minor notes.

## Blocker: the mobile ring's Attack button breaks when the option is off

`Show Attack Button` is a plain `boolToggle` in the Interface panel (`src/ui/options_view.ts`), so it is reachable on touch. But the mobile ring's slot 0 is hardcoded as Attack:

- `src/ui/hud.ts`, mobile ring descriptor: `slotIndex: 0, isAttack: () => true` (it does not consult `attackSlotIsAttack()`), so the ring keeps painting the Attack button.
- Its tap handler still routes through `bindTouchTap(attackBtn, ...) -> this.castSlot(0)`, and `castSlot` now gates the auto-attack branch on `barSlot === 0 && this.attackSlotIsAttack()`.

So on a touch device with the option off:

- Tapping the ring's Attack button with a live hostile target selected, or while auto-attack is already running, falls through to the generic path and casts `attackSlotAction`, which is `null` on mobile because nothing can ever assign it there (`sourceSlotForMobileButton` starts at `MOBILE_ACTION_SOURCE_SLOT_START`, so the ring never addresses barSlot 0). Result: the button does nothing, and there is no way to STOP auto-attack from the ring.
- The only surviving behavior is the `onMobileAttackNearest` no-target branch.

That is a silent, hard-to-diagnose break of the primary combat control on phones, triggered by a setting the player can flip without any hint that it is desktop-only. Two clean ways out: either hide the toggle on touch (gate the `boolToggle` on touch capability, matching how the ring's removal gesture is desktop-only anyway), or give the mobile Attack button its own auto-attack path that does not depend on slot-0 semantics (e.g. a small `toggleAutoAttack()` the ring calls directly, leaving `castSlot(0)` for the desktop bar). I'd prefer the second: it keeps the Interface panel one list on every device and makes the mobile ring immune to slot-0 mode entirely. A test that drives the ring's tap with `showAttackButton` off would pin it.

## Minor

1. No `dragstart` on the freed slot 0, so once an action is assigned there it can be cleared (right-click) but never dragged back out or moved to another slot, unlike every other slot. Worth adding for symmetry, or at least a comment saying it is deliberate.
2. `boundCount` in `action_bar_view.ts` now counts slot 0's assignment toward `manySpells` when Attack is removed. Cosmetic (compact-bar flag) and arguably correct, but it does make the count no longer byte-identical to the old `hotbarActions.filter(...)` the comment above it still claims.

Once the mobile path is handled I am happy to approve. The rest of the change looks solid and well tested.

Blaine1705:
Implemented the requested follow-up and rebuilt the branch on the current `release/v0.26.0` head.

Changes made:

- Kept the mobile Attack control independent from the now-assignable desktop slot 0, so removing Attack from the desktop bar cannot make the mobile attack button cast the assigned slot action.
- Moved the mobile attack decision into the existing pure hotbar module and added behavior tests for active combat, live targets, nearest-target acquisition, and the fallback path.
- Validated the persisted slot-0 action through the canonical hotbar parser, including malformed JSON, missing IDs, and actions that no longer exist.
- Guarded `localStorage` reads and writes and remove the persisted key when the slot is cleared.
- Made an assigned slot-0 action draggable back onto the regular action bar.
- Updated the affected source guards and regenerated the i18n artifacts and resolved-table hash.

Verification completed:

- Focused HUD, hotbar, action-bar, options, performance-budget, and mobile-ring suites: 137 passed, 3 skipped.
- `tests/client_shell.test.ts`: 86 passed.
- `npm run check:types`: passed with 0 Svelte diagnostics.
- `npm run build:env`, `npm run build:server`, and `npm run build`: passed.
- The pre-push QA hook passed.

The full local `npm run gate` reached the complete Vitest suite but cannot be fully green on this Windows environment because several unrelated tooling tests require Unix ownership/mode semantics, POSIX signals, and symlink/atomic-rename behavior. The browser suite also cannot launch because the expected Playwright Chromium binary is not installed locally. The feature-focused tests, type checks, and all builds are green.

Current PR head: `bfe878d6b6b670d3f7d99785825398bc9bf28ac0`.

Blaine1705:
Mobile behavior is now fixed in this PR by commit `03adb8c79`.

The Interface > Show Attack Button option now controls the fixed mobile Attack button correctly. Turning it off hides the control and its hit area, turning it on restores it, and the setting persists across reloads.

Verified with:

- Real mobile flow in portrait and landscape
- 111 focused unit tests
- 2 browser regression cases for portrait and landscape
- TypeScript
- Changed-files Biome check
- Pre-push QA floor, including 53 passing guard tests

The full browser suite passes 61 of 62 tests. The remaining failure is the pre-existing Armory mobile layout assertion, unrelated to this change.

Visual evidence:

- [Attack visible](https://github.com/Blaine1705/world-of-claudecraft/blob/03adb8c79eb341bb3861ff625d2edf5e6363797b/docs/screenshots/pr-1861-mobile-attack-visible.png)
- [Attack hidden](https://github.com/Blaine1705/world-of-claudecraft/blob/03adb8c79eb341bb3861ff625d2edf5e6363797b/docs/screenshots/pr-1861-mobile-attack-hidden.png)

EnriqueGF:
:) Thanks again for the follow-up work here, the mobile round is genuinely well handled now: the fixed control is decoupled from the assignable desktop slot 0 via `handleMobileAttackTap`, hidden through the elided `setDisplay` writer, and it has both unit and browser coverage. I re-reviewed against `release/v0.26.0` and re-ran the checks.

**Base branch: correct.** Points at `release/v0.26.0` and HEAD is rebased on its tip. tsc is clean; the architecture, S3 localization, world_api parity, i18n completeness, i18n resolved-equivalence (the hash that was red in round 1 is regenerated now, `i18n:gen` leaves the tree clean), and hud_perf_budget guards all pass. No sim/server/wire changes, so determinism, tri-host parity and server authority are not in play.

**Confirmed resolved from earlier rounds:** the mobile Attack break, the i18n resolved-hash baseline, routing persisted-action load through the canonical `parseHotbarAction` predicates, and the slot-0 `dragstart` for symmetry. Nice.

Two items still need to land before this is mergeable.

## Blocking

1. **Desktop before/after screenshots are still missing.** The headline gesture (right-click to remove Attack, drag an ability into the freed slot 0, key 1 casting it) is a desktop interaction, and only the two mobile screenshots are committed. The PR body still says "Screenshots / recordings: To follow". Please capture desktop before/after under `docs/screenshots` and reference them from the body (the `pr-screenshots` skill has the recipe).

2. **The desktop slot-0 subsystem has no behavioral test, only source-string scans.** This was round 1's "should fix #2"; the pure parts were extracted and pinned well (`parseStoredHotbarAction`, `encodeStoredHotbarAction`, `handleMobileAttackTap` in `hotbar.ts` with real tests), but the actual desktop behavior is still unexercised:
   - `src/ui/hud.ts:5999-6006` right-click remove (`showAttackButton -> false`) then clear (`attackSlotAction -> null`).
   - `src/ui/hud.ts:6009,6034-6047` dragstart out of slot 0, drag-drop assign into the freed slot, and the drag-from-slot-0 path that clears the source slot (`sourceAttackSlot` + `clearHotbarSlot`).
   - `src/ui/hud.ts:5437-5456` the localStorage round-trip (`attackSlotKey()` `:s0` suffix, `saveAttackSlotAction` setItem-vs-removeItem-on-null, `loadAttackSlotAction`). This is load-bearing persisted state with zero pin; the `:s0` key literal is asserted nowhere.
   - `src/ui/hud.ts:5706-5709` the `castSlot(0)` fall-through: key 1 casting the assigned action when the toggle is off (empty vs assigned). This is the core payoff of the feature and is completely unexercised.

   The only touchpoints are the `client_shell.test.ts` / `mobile_action_ring_painter.test.ts` `toContain(...)` source scans, which pass even if the contextmenu/drop logic regresses or is deleted. Per module-first, extract the slot-0 mutation + persistence into a small Node-testable helper (mirroring `handleMobileAttackTap`): remove/clear on contextmenu, assign/move on drop, and the key/serialize round-trip taking an injected storage, then pin each transition to literals. That both closes the coverage gap and thins the ~130 new lines currently grown onto `hud.ts`.

## Minor (non-blocking)

3. `src/ui/hud.ts:6040` the slot-0 drop sets `this.attackSlotAction = dragged.action` without the validity gate sibling slots apply (`placeAbilityOnSlot`/`placeItemOnSlot` re-check `isHotbarItemId`). `loadAttackSlotAction` validates on read, so a bad entry silently vanishes next session rather than erroring, but the asymmetry is worth a matching gate or a comment.

4. The slot-0 `dragover`/`dragleave`/`drop`/`dragend` block is a near copy of the slot 1..22 handlers (`src/ui/hud.ts:5937-5989`); a shared `assignDroppedAction` helper taking an "is attack slot" flag would collapse the duplication.

5. `src/ui/action_bar_view.ts:238-239` the `boundCount` comment still claims byte-identical parity with `hotbarActions.filter(...)`, but slot 0's assignment now counts toward `manySpells` when Attack is removed. Cosmetic (compact-bar threshold), just update the comment.

Happy to re-review once the desktop screenshots and the slot-0 behavioral coverage are in. The mobile handling and the pure-helper extraction are solid.

Blaine1705:
Implemented and pushed in commit `a1703f152`.

- Resolved the branch against the current `release/v0.26.0`, including the generated i18n state.
- Completed the removable desktop Attack slot behavior: persisted slot-0 state, right-click removal, drag/drop assignment and source clearing, with focused behavioral tests in `tests/hotbar.test.ts`.
- Added desktop before/after screenshots:
  - [Attack visible](https://github.com/Blaine1705/world-of-claudecraft/blob/a1703f152/docs/screenshots/pr-1861-desktop-attack-visible.png)
  - [Attack removed and slot reassigned](https://github.com/Blaine1705/world-of-claudecraft/blob/a1703f152/docs/screenshots/pr-1861-desktop-attack-removed-slot-assigned.png)

Verification completed:
- 115 focused tests passed.
- TypeScript check passed.
- Pre-push QA floor passed: 53 tests passed, 3 skipped.
- Biome completed without errors; the reported warnings are existing repository warnings.

GitHub currently reports the PR as mergeable. No GitHub checks are reported for this branch yet. The full gate was not run because this environment is missing `ffmpeg` and `ffprobe`.

PR #1863, opened 2026-07-13
fix(net): resume into the world after a mobile WebView reload

EnriqueGF:
## Summary

iPhone players report the game "randomly returns me to the login screen" (iPhone 17 Pro Max, multiple users). Root cause: iOS evicts the backgrounded WKWebView under memory pressure, so on foreground the app reloads `index.html`. The client is a single-page app whose boot path lands a restored session on the home / character-select chrome and **never re-enters the world**, silently dropping the player out of the game (and, if the stored token also fails revalidation, all the way to the login screen).

This PR fixes the dominant cause: it persists an **active-play marker** (`characterId` + `savedAt`) so the boot path can re-enter the world instead of home.

- New pure leaf `src/net/resume_play.ts`: marker parse/serialize + freshness decision (unit-tested), plus fail-soft `localStorage` wrappers.
- The marker is **freshness-scoped**: a 30-minute window, re-stamped on `visibilitychange -> hidden` / `pagehide` (the pre-eviction instant). A cold open of a long-closed tab still lands on home as before; only a genuine reload-during-play resumes. The window sits inside the server's 5-minute linkdead grace for a seamless resume, and past it a fresh join is still correct.
- Resume routes through the existing `enterWorld` path (not `takeOverAndEnter`): a linkdead session resumes seamlessly, and a genuinely **live duplicate on another device** is rejected with `character already in world` and falls to the fatal overlay, so resume can never silently kick another device.
- The marker is cleared on every terminal exit (deliberate in-game / account logout, 401/403 session expiry, entry timeout, fatal disconnect), and the boot resume intent is dropped when the realm list is shown, so it can never linger into a non-playing char-select and cause a spurious auto-enter.

`src/main.ts` stays a thin consumer (save on `enterWorld`, consume on boot, clear on the terminal sites).

### Known follow-ups (not in this PR)

Two secondary contributors remain, worth a sibling fix:
- The session token has a hard 7-day TTL (`server/db.ts` `saveToken`) with no client refresh or server sliding renewal, so a reconnect after expiry still routes to login. Recommend sliding renewal on WS re-auth / REST use.
- A WiFi/cellular black-hole handoff can exhaust `MAX_CONFLICT_REJECTIONS = 8` (`src/net/reconnect_policy.ts`) before the server sweeps the dead socket.

## Related issues

Reported on Discord (iPhone 17 Pro Max "returns to login" during play). Part of the mobile-stability effort.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/resume_play.test.ts` (14 cases: parse/serialize round-trip, malformed/`null`/non-integer/negative `characterId`, missing/non-finite `savedAt`, both freshness boundaries, backward-clock guard, storage round-trip, private-mode fail-soft)
  - `npx vitest run tests/play_online_only.test.ts tests/main_api_error.test.ts tests/net_online_visibility_reconnect.test.ts` (main.ts contract + reconnect suites)
  - `npx tsc --noEmit`, `biome check` (changed files), `npx vite build` (client bundles, green), guard tests (`architecture`, `localization_fixes`)
- Manual steps: N/A automated; the fix is a boot-flow change. Suggested device check: enter the world on iOS, background the app until the WebView is evicted, foreground, and confirm it resumes into the world instead of home/login.

## Screenshots / recordings

N/A: no visual/UI surface changes (boot-flow behavior only).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -108,9 +108,11 @@ import {
   type ReleaseEntry,
 } from './net/online';
 import { realmPopulation } from './net/realm_population';
+import { RECONNECT_CONFLICT_ERROR } from './net/reconnect_policy';
 import {
   clearPlayMarker,
-  freshMarkerCharacterId,
+  freshMarker,
+  markResumeAttempt,
   readPlayMarker,
   refreshPlayMarker,
   savePlayMarker,
@@ -292,10 +294,10 @@ let pendingDeleteCharacter: CharacterSummary | null = null;
 // instead of a per-row one; it acts on whichever character is selected. Mobile
 // and narrow layouts keep the per-row buttons and never read this.
 let charselectSelected: CharacterSummary | null = null;
-// One-shot: set by the boot resume path to the character id to auto-enter, then
-// consumed by refreshCharacters once its list has loaded (mobile WebView-reload
-// resume; see src/net/resume_play.ts).
-let pendingResumeCharacterId: number | null = null;
+// One-shot: set by the boot resume path to the character (and the realm it was
+// playing on) to auto-enter, then consumed by refreshCharacters once its list
+// has loaded (mobile WebView-reload resume; see src/net/resume_play.ts).
+let pendingResume: { characterId: number; realm: string } | null = null;
 let homepageMusic: HTMLAudioElement | null = null;
 let homepageMusicStarted = false;
 let homepageMusicMuted = readHomepageMusicMuted();
@@ -4017,6 +4019,7 @@ function setupAccountPortal(): void {
     try {
       await api.deactivateAccount(deUser.value, dePass.value);
       api.clearSession();
+      clearPlayMarker();
       setAccountFieldMsg('#account-deactivate-msg', t('hudChrome.account.deactivated'), true);
       window.setTimeout(() => location.reload(), 1200);
     } catch (e2) {
@@ -4171,7 +4174,7 @@ function showRealmList(dir?: import('./net/online').RealmDirectory): void {
   // (no remembered realm). Drop any pending resume intent here so a later manual
   // realm pick does not surprise-enter the world on a decision made at boot; the
   // player continues through normal realm + character select.
-  pendingResumeCharacterId = null;
+  pendingResume = null;
   show('#realm-panel');
   const listEl = $('#realm-list');
   const render = (d: import('./net/online').RealmDirectory) => {
@@ -4459,27 +4462,31 @@ async function refreshCharacters(): Promise<void> {
     if (chars.some((c) => c.skinCatalog === 'mech')) void preloadMechAssets();
     if (api.realm) $('#charselect-realm').textContent = api.realm;
     listEl.innerHTML = '';
-    if (chars.length === 0) {
-      // No characters on this realm, drop straight into the create screen.
-      listEl.innerHTML = `<li class="char-list-message">${escapeHtml(t('character.noneYet'))}</li>`;
-      show('#charcreate-panel');
-      return;
-    }
     // Boot resume: a WebView reload during play sent us here with a persisted
-    // active-play marker. If that character still exists, re-enter the world
-    // directly instead of showing char-select (a linkdead session resumes
-    // seamlessly; a genuinely live duplicate falls back to the fatal overlay,
-    // same as a manual enter). One-shot: consume the pending id either way, and
-    // clear the persisted marker if the character is gone so we stop retrying.
-    if (pendingResumeCharacterId !== null) {
-      const resumeId = pendingResumeCharacterId;
-      pendingResumeCharacterId = null;
-      const target = chars.find((c) => c.id === resumeId);
+    // active-play marker. If that character still exists on the marker's realm,
+    // re-enter the world directly instead of showing char-select (a linkdead
+    // session resumes seamlessly; a genuinely live duplicate falls back to the
+    // fatal overlay, same as a manual enter). One-shot: consume the pending
+    // intent either way (including the empty-roster case below), and clear the
+    // persisted marker if the character or realm is gone so we stop retrying.
+    // The realm check closes the cross-realm id-collision hole: character ids
+    // are only unique per realm database.
+    if (pendingResume !== null) {
+      const resume = pendingResume;
+      pendingResume = null;
+      const target =
+        resume.realm === api.realm ? chars.find((c) => c.id === resume.characterId) : undefined;
       if (target) {
         void enterWorld(target);
         return;
       }
       clearPlayMarker();
+    }
+    if (chars.length === 0) {
+      // No characters on this realm, drop straight into the create screen.
+      listEl.innerHTML = `<li class="char-list-message">${escapeHtml(t('character.noneYet'))}</li>`;
+      show('#charcreate-panel');
+      return;
     }
     for (const c of chars) {
       const row = document.createElement('li');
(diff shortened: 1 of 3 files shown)


Rubsey (commented):
This is a solid fix for a real problem, and the structure is right: the pure decision core in `src/net/resume_play.ts` (parse hardening, `now` as a parameter, fail-soft storage wrappers) matches the src/net sibling-module pattern, the linkdead interplay is correct (a plain join resumes a same-account linkdead session per `planJoin`, and a live duplicate on another device is rejected, never displaced), and entry coverage is complete (index.html and play.html both bootstrap main.ts, and the Capacitor native app loads index.html). All gates are green at the head: the 14 new unit tests, the adjacent main/net suites (37 tests), the architecture and S3 guards, tsc, and biome on the changed files.

The findings are all in the marker lifecycle around the happy path.

**Blocking**

1. Post-connect entry failure becomes an unescapable reload loop. `savePlayMarker` fires in the success poll (src/main.ts:4559) before `startGame` runs, and startGame's two fatal arms, assetsFailed (src/main.ts:905) and rendererFailed (src/main.ts:991), do not clear the marker. `fatalOverlay`'s only action is `location.reload()` (src/main.ts:4455), and the boot resume then auto-re-enters, re-stamps at 4559, and re-fails. While an asset fetch or WebGL context failure persists (both plausible in exactly the post-eviction memory-pressure situation this PR targets) the user can never reach the home screen, and the window never expires because every cycle re-stamps. Your own comments at 4566 and 4578 state the intent that a failed entry drops the marker; these two arms were missed. Clearing inside `fatalOverlay` itself would cover every present and future caller.

**Should fix**

2. A stale marker is resurrected by the restamp handlers. `refreshPlayMarker` re-stamps whatever marker exists with no freshness check (src/net/resume_play.ts:104), the handlers are registered unconditionally (src/main.ts:8237), and the boot path returns on a stale marker without clearing it. Markers routinely outlive sessions (closing the browser mid-play restamps on pagehide rather than clearing), so: play once, come back days later, background the homepage tab, and the next reload auto-enters the world. That converts the documented 30-minute window into effectively unbounded. Gate the refresh on the marker still being fresh (that logic belongs in the pure module where it is testable) and clear a stale marker when boot sees one.

3. A second tab while playing lands on a fatal overlay and disarms the real resume. localStorage is shared across tabs, so opening the site in tab B inside the window auto-runs `enterWorld` for the live character. A fresh join has no conflict tolerance (`isTransientReconnectRejection` requires `reconnectAttempts > 0`, src/net/reconnect_policy.ts:31, and an attempt-0 error frame ends the session, src/net/online.ts:1621), so tab B shows "character already in world" as a fatal overlay where it used to show the homepage, and its onDisconnect `clearPlayMarker` (src/main.ts:4580) erases the playing tab's marker for the rest of that session, since `refreshPlayMarker` never recreates one. At minimum, skip the clear when the rejection is `RECONNECT_CONFLICT_ERROR`; a per-tab sessionStorage gate on the boot resume would fix the tab-B landing too.

4. The black-holed-drop window defeats the resume in the PR's own target scenario. If the evicted socket died without a FIN (the WiFi-to-cellular family reconnect_policy.ts documents), the server counts it live until the pong sweep flips it linkdead (server/game.ts pingLiveSessions, one to two 30s keepalive intervals). A fast foreground inside that window gets the final conflict rejection, the fatal overlay, and, because of the same onDisconnect clear, no second chance on the next reload. The finding-3 mitigation covers this; tolerating a bounded number of conflict rejections on the resume join would make it seamless.

5. `pendingResumeCharacterId` survives failure paths and can fire much later. The refreshCharacters catch (src/main.ts:4439) and the goToLoggedInPlay catch (src/main.ts:6930) both leave it set, and the consume block runs on every subsequent refreshCharacters call: sort change (4224), inline realm switch (4193), language change (5006), rename (4370), create/delete (7826, 7874). A failed boot resume on a flaky radio (the target moment) can therefore auto-enter the world later on an action that was not "enter world". The zero-characters early return (4306) also precedes the consume block, stranding both the intent and the marker. Clear the pending id in both catches, and consume before the early return.

6. The resume arm is unguarded on /desktop-login. In the restored-session arm both `completeDesktopBrowserLogin()` (src/main.ts:8228) and the resume `.then()` run; with a fresh marker the handoff tab starts loading the whole game while minting the desktop code. One line: `if (isDesktopLoginPage()) return;` in the resume arm.

**Nits**

7. The marker is not realm-scoped. The resume trusts `woc_last_realm`, which the inline realm switcher rewrites on selection (src/main.ts:4191), so a realm browse before the reload silently kills the resume (id miss on the wrong roster, marker cleared). In the checked-in shared-DB schema character ids are globally unique, but in a realm directory whose realms live on separate databases a cross-realm id collision could auto-enter the wrong character. Storing the realm name in the marker closes both.

8. Three clearSession sites miss the marker clear: deactivateAccount (src/main.ts:3863), the recovery-email logout hatch (src/main.ts:6517), and goToLoggedInPlay's auth catch (src/main.ts:6932). All bounded (resume needs a valid restored token plus a roster match), but add `clearPlayMarker()` beside each for symmetry.

9. Tests: `RESUME_KEY` is only asserted via the imported constant and `serializeMarker` only via round-trip, so a key rename or shape change keeps all 14 tests green while orphaning every player's stored marker; pin both to literals. The main.ts wiring, which is the behavior the fix actually delivers, has no test; the readFileSync string-pin idiom in tests/play_online_only.test.ts would pin the save/clear/drop sites cheaply. Note the refresh test (tests/resume_play.test.ts:91) currently enshrines the staleness-blind restamp, so the finding-2 fix will amend it.

**Notes, no action required**

- The feature is platform-unconditional: a desktop player who closes the tab mid-play (pagehide restamps) and revisits within 30 minutes is pulled straight into the world. That may be the right product call, but it is a user-visible behavior change on platforms the description does not mention; worth stating explicitly.
- With a fresh marker and a still-down network, boot now lands on the login panel with an error (goToLoggedInPlay's catch) where it used to land quietly on home. The marker survives so the next reload retries; acceptable, flagging for awareness.
- A force-renamed or moderated character resumes into the server's handshake rejection: correct enforcement, no bypass, just a fatal overlay instead of the char-select rename UI.
- Verified as claimed: the second-device case rejects without displacing the live session; past the 5-minute grace the resume is a clean fresh join; a cross-account marker on a shared browser dead-ends at the token-scoped roster and is cleared.

EnriqueGF:
Thanks for this contribution, it fixes a real and painful mobile bug and the module structure was right from the start :)

Review outcome: APPROVED after follow-up work landed on the branch. Summary of what was done and verified:

Base branch gate: the PR originally targeted release/v0.26.0. It has been retargeted to release/v0.27.0 (the active integration base) and rebased on its tip; the only conflict was an adjacent import block.

All findings from Rubsey's review are now addressed on the branch:
1. (Blocking) Post-connect entry failure reload loop: the marker is now cleared inside fatalOverlay itself, covering the assetsFailed and rendererFailed arms and every future caller. The one exception is a duplicate-session conflict, where the caller opts out so a second tab can never erase the playing session's marker.
2. Stale marker resurrection: refreshPlayMarker is now freshness gated; a stale marker is cleared instead of re-stamped, and the boot arm clears a marker that no longer resumes.
3. and 4. Second tab / black-holed drop: a conflict rejection keeps the marker (the session lives elsewhere or the server has not yet swept the dead socket), and a persisted attempt budget (MAX_RESUME_ATTEMPTS = 3, reset on completed entry) bounds the retries so a terminally failing auto-resume self-disarms instead of looping on the overlay.
5. Pending intent leaks: pendingResume is dropped in the refreshCharacters catch, the goToLoggedInPlay catch, and showRealmList, and is consumed before the empty-roster early return, so it can never fire on a later unrelated roster refresh.
6. Desktop-login handoff: the resume arm returns early on isDesktopLoginPage().
7. Realm scoping: the marker now stores the realm name; boot restores it as the remembered realm (fail-soft write) and the roster consume gate requires the marker realm to match api.realm, closing the cross-realm id-collision hole and surviving a realm browse before the reload.
8. The remaining clearSession sites (deactivate, recovery-email hatch, boot auth failure) clear the marker too.
9. Tests: RESUME_KEY and the serialized shape are pinned to literals, every main.ts lifecycle site is pinned (exact clear-site count plus per-site content pins), and the refresh test now asserts the staleness-gated behavior. 29 cases green.

Verified locally on the rebased branch: tsc clean, biome clean on changed files (warnings only), tests/resume_play.test.ts (29), architecture and S3 i18n guards, tests/parity, the adjacent main/net suites, the full vitest suite (14303 passed), and all three builds. The only red gate step is tests/browser/armory_mobile_layout.browser.test.ts, which fails identically on the release/v0.27.0 tip in this environment, so it is pre-existing and unrelated to this change.

Known non-blocking notes for the record: a retries-exhausted disconnect still clears the marker by design (resuming into a dead network would loop); cross-tab interference is bounded by the attempt budget; the token 7-day TTL and conflict-tolerance follow-ups from the PR description remain open as sibling work. No visual surface changes, so no screenshots required.

PR #1867, opened 2026-07-13
fix(render): Nythraxis phase-2 court uses the Aldren / Malric / Voss models

ryan-foo:
## Problem

In the Nythraxis raid's phase-2 heroic summon, the court that rises again (Spirit of Aldren, Spirit of Malric, Spirit of Voss) all render as identical generic skeletons instead of the three distinct characters.

## Cause

The phase-2 add templates (`nythraxis_heroic_warrior_add` / `_priest_add` / `_rogue_add`, spawned by `spawnNythraxisHeroicAdds`) are missing from `MOB_KEYS` in `src/render/characters/manifest.ts`. `visualKeyFor` therefore falls through to `FAMILY_KEYS.undead` = `skel_minion` for all three. Their named crypt counterparts are mapped correctly (`fallen_captain_aldren` -> `skel_warrior`, `corrupted_priest_malric` -> `skel_necromancer`, `deathstalker_voss` -> `skel_rogue`), so the raid court looked nothing like the characters it is meant to be.

## Fix

Map the three court add ids to the same visuals their named versions already use:

- `nythraxis_heroic_warrior_add` (Aldren) -> `skel_warrior`
- `nythraxis_heroic_priest_add` (Malric) -> `skel_necromancer`
- `nythraxis_heroic_rogue_add` (Voss) -> `skel_rogue`

Those visuals are already in the tier-independent character preload set (the crypt bosses use them), so there is no new asset or preload change. Presentation-only: no sim, gameplay, wire, or parity impact.

## Testing

`tests/visual_manifest.test.ts` gains a case that pins each court add to its named counterpart's visual key and asserts it is never `skel_minion`. `tests/render_asset_preload.test.ts` still passes (no new asset to preload). `tsc` clean.

No screenshots: this is a phase-2 raid encounter (heroic Nythraxis, boss past the summon phase) that is impractical to reach headless; the manifest resolution is deterministic and covered by the unit test.

### DIFF
--- a/src/render/characters/manifest.ts
+++ b/src/render/characters/manifest.ts
@@ -1109,6 +1109,13 @@ const MOB_KEYS: Record<string, string> = {
   fallen_captain_aldren: 'skel_warrior',
   corrupted_priest_malric: 'skel_necromancer',
   deathstalker_voss: 'skel_rogue',
+  // The Nythraxis phase-2 heroic court is Aldren / Malric / Voss risen again, so
+  // the "Spirit of X" adds reuse each character's crypt visual above. Without these
+  // the ids fall through to FAMILY_KEYS.undead (skel_minion) and the whole court
+  // renders as identical generic skeletons. See spawnNythraxisHeroicAdds.
+  nythraxis_heroic_warrior_add: 'skel_warrior', // Spirit of Aldren
+  nythraxis_heroic_priest_add: 'skel_necromancer', // Spirit of Malric
+  nythraxis_heroic_rogue_add: 'skel_rogue', // Spirit of Voss
   vision_aldren_warrior: 'player_warrior',
   vision_malric_mage: 'player_mage',
   vision_deathstalker_voss: 'player_rogue',
--- a/tests/visual_manifest.test.ts
+++ b/tests/visual_manifest.test.ts
@@ -106,6 +106,24 @@ describe('character visual manifest', () => {
     expect(VISUALS.mob_boar.deathTimeScale).toBeUndefined();
   });
 
+  it('renders the Nythraxis phase-2 court as Aldren / Malric / Voss, not generic skeletons', () => {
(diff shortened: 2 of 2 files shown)


jgyy (commented):
Small, correct fix. The three new `MOB_KEYS` entries map the phase-2 add ids to the same visual keys their named crypt-boss counterparts already use (Aldren to skel_warrior, Malric to skel_necromancer, Voss to skel_rogue), and all three referenced models are real, already-used entries in the same file, so nothing here points at a nonexistent asset.

`tests/visual_manifest.test.ts` pins each add's `visualKeyFor` result against its named counterpart and explicitly asserts it's not the generic `skel_minion` fallback it was hitting before, that's a good decisive test for a visual-key bug.

No findings. Fair call skipping a screenshot here given phase-2 is a late, hard-to-reach encounter state and the unit test already pins the regression.

Thanks for tracking this down, mismatched boss-court models are an easy thing to miss.

Rubsey (commented):
Verified end to end. The three adds (`nythraxis_heroic_warrior_add` / `_priest_add` / `_rogue_add`) all carry `family: 'undead'` in `src/sim/content/dungeons.ts`, so `visualKeyFor` was resolving them through `FAMILY_KEYS.undead = 'skel_minion'` (manifest.ts:1130) exactly as claimed, and they are genuinely spawned by `spawnNythraxisHeroicAdds`. The new `MOB_KEYS` entries hit the `override` branch first (manifest.ts:1171 to 1173) and point at `skel_warrior` / `skel_necromancer` / `skel_rogue`, which are real VISUALS already used by the named crypt bosses, so no new asset or preload. Pure render lookup table, no sim/wire/parity reach.

The test is decisive: pinning each add's key to its named counterpart's key and asserting it is not `skel_minion` would fail on regression in both directions.

No findings. Merges clean against the current `release/v0.26.0` tip.

PR #1869, opened 2026-07-13
fix(combat): same-class group buffs no longer stack (no double Sureflight Aura)

ryan-foo:
## Problem

Two hunters could stack two copies of Sureflight Aura on the same target (and similarly any group buff not already deduped), double-dipping the effect.

## Cause

`auraReplacementConflicts` (`src/sim/combat/aura_stacking.ts`) only replaces an existing aura across DIFFERENT casters when the aura id is in `SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS`; otherwise a second caster stacks a duplicate. That set covered six of the seven party group buffs (Arcane Intellect, Battle Shout, Blessing of Might, Devotion/Steadfast Aura, Mark of the Wild, Power Word: Fortitude) but not the hunter's Sureflight Aura. Sureflight is `trueshot_aura` (talent-granted) applied by `aoeAllyAttackPower` as the aura id `trueshot_aura_ap`, so it was never deduped.

## Fix

Add `trueshot_aura_ap` to the source-independent set, so a second hunter's Sureflight replaces the first instead of stacking. Every party group buff is now in the set. Pure-leaf change: no rng, no parity impact.

## Testing

`tests/raid_buffs.test.ts` gains a case: two hunters applying Sureflight to a target leave exactly one aura (the newer caster's), mirroring the existing Blessing-of-Might replacement test. Parity and architecture suites green; `tsc` clean.

### DIFF
--- a/src/sim/combat/aura_stacking.ts
+++ b/src/sim/combat/aura_stacking.ts
@@ -1,5 +1,10 @@
 import type { Aura } from '../types';
 
+// Persistent group buffs that are ONE per target regardless of caster: a second
+// same-class caster REPLACES the existing aura instead of stacking a duplicate (no
+// double Arcane Intellect, no two Sureflight Auras, etc.). Every party group buff
+// belongs here. The buffTarget party buffs use the ability id as the aura id, while
+// the hunter aura (aoeAllyAttackPower) applies as `${abilityId}_ap`.
 const SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS = new Set([
   'arcane_intellect',
   'battle_shout',
@@ -7,6 +12,7 @@ const SOURCE_INDEPENDENT_GROUP_BUFF_AURA_IDS = new Set([
   'devotion_aura',
   'mark_of_the_wild',
   'power_word_fortitude',
+  'trueshot_aura_ap', // Sureflight Aura (hunter aoeAllyAttackPower)
 ]);
 
 export function auraReplacementConflicts(auras: readonly Aura[], aura: Aura): number[] {
--- a/tests/raid_buffs.test.ts
+++ b/tests/raid_buffs.test.ts
@@ -191,6 +191,35 @@ describe('standardized percent raid buffs', () => {
     expect(blessings[0].remaining).toBe(1800);
   });
 
+  it('does not stack Sureflight Aura from two hunters (same-class group buff)', () => {
+    // trueshot_aura (Sureflight Aura) is a talent-granted hunter aura applied as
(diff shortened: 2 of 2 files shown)


jgyy (commented):
Correctly scoped fix. Adding `trueshot_aura_ap` to the source-independent group-buff set means the aura collapses to one instance regardless of which hunter cast it, while everything keyed by `sourceId` (DoTs, personal buffs, debuffs) is untouched since that path is unchanged.

`tests/raid_buffs.test.ts` adds a decisive case: two hunters, two different `sourceId`s, asserts exactly one aura survives and it's tied to the second caster. That's the right regression pin for "no double Sureflight Aura."

No blocking or should-fix findings. Same maintenance note as the rest of that allowlist: any future group buff aura needs to be added here by hand, but that's the pre-existing pattern, not something this PR introduced or needs to fix.

Thanks, this closes a real double-dip bug cleanly.

Rubsey (commented):
Verified this end to end at 5625c05 and the fix is real and correctly scoped.

- Base does stack two Sureflight copies: `auraReplacementConflicts` only collapses across casters for ids in the set, and the base set has the six buffTarget party buffs but not `trueshot_aura_ap`, so two different `sourceId`s return an empty conflict list and both auras push (src/sim/combat/aura_stacking.ts).
- The id is right: `aoeAllyAttackPower` applies `${ability.id}_ap` (src/sim/combat/effect_dispatch.ts:877), and the aura shape in the test (kind `buff_ap_pct`, value 10, duration 1800, school nature) matches the real `trueshot_aura` record (src/sim/content/classes.ts:3691) exactly.
- Replacement rides the same newest-wins path the other six use: applyAura splices the conflicts, pushes the new aura, and recalcs player stats for player targets (src/sim/sim.ts:4400). No rng, no wire shape change, same code on all three hosts.
- The test is a decisive red/green pin: without the set entry the filter finds two auras and the length assertion fails. The private applyAura cast mirrors the adjacent Blessing test, so no new pattern.

One nit, take it or leave it:

- nit: tests/raid_buffs.test.ts:194 applies the aura directly, so the literal `trueshot_aura_ap` lives in the test and in the set but nothing checks the dispatch path still produces that id. If the `${ability.id}_ap` composition or the ability id ever changes, this test stays green while the bug quietly returns. An end-to-end variant (grant the talent, cast from both hunters) would pin the whole chain the way the castAbility Blessing test above it does. The talent grant makes that more setup, so not a blocker.

Merges clean against the current release/v0.26.0 tip. Fork CI was sitting at action_required; I approved both runs so the gate can report on this head.

PR #1870, opened 2026-07-13
feat(ui): drag bag items onto the paperdoll, and quiet the character-select delete button

EnriqueGF:
## Summary

Two inventory / shell quality-of-life changes.

**1. Drag bag items onto the paperdoll; right-click no longer destroys.**

Right-click on a bag stack now runs the same action as left-click (use / equip), the classic
binding. Destroying moved to the drag-out gesture: drag the stack onto the world and the
existing confirm + quantity prompt opens (`noDiscard` items stay protected, nothing is
destroyed by the drop itself). Every stack is draggable now, not just the hotbar-eligible ones.

The paperdoll sockets become drop targets: dropping a piece equips it into that exact socket
through a new `IWorld` member, `equipItemToSlot`, so a ring dropped on the second finger lands
there even while the first is free. Only the sockets that would really take the piece light up,
because the hover feedback runs the same pure decision the drop runs.

The aimed slot is a REQUEST, never a bypass: `src/sim/items.ts` re-validates it against the item
(the new pure `slotAcceptsItem` leaf) and the class / level gates still apply, so a hand-crafted
`equip` packet cannot put a helm on a ring finger. The wire field is additive on the existing
`equip` token; omitted, every existing path stays byte-identical.

Touch gets the same gesture through a pointer drag (hold to arm, a flick still scrolls the bag
grid, a ghost follows the finger, a `pointercancel` never drops). Because a window is full-screen
on a phone, the bags used to cover the paperdoll and the drag had nowhere to land, so the
character sheet and the bags now dock 50/50 when both are open, reusing the bank cluster's
pairing pattern.

New modules, nothing accretes on `hud.ts` beyond the wiring: `equip_drop_core` and
`char_bags_pairing_core` (pure cores), plus `item_drag_state`, `item_drop_hit_test`,
`touch_item_drag`, `world_drop_target`.

**2. The character-select delete button no longer reads as the call to action.**

Deleting a character is irreversible, yet the control was a full-size red `.btn-danger` sitting
BEFORE Enter World: the loudest, left-most, most button-shaped thing in the row, so it invited
the reflex click and only the type-your-name modal saved players. It is now a quiet icon-only bin
at the END of the row (also last in the tab order), muted until hovered or focused. The confirm
modal is untouched; it keeps its accessible name, its disabled-while-in-world explanation, and
the 40x40 tap floor on touch.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npm run gate` (green, see the one pre-existing failure noted below), `npx tsc --noEmit`,
  `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/localization_fixes.test.ts tests/parity`,
  plus the new suites `tests/equip_target_slot.test.ts`, `tests/equip_drop_core.test.ts`,
  `tests/touch_item_drag.test.ts`, `tests/char_bags_pairing_core.test.ts`,
  `tests/char_delete_button.test.ts`.
- Manual steps: drove the real client (desktop 1600x900 and a landscape phone viewport): dragged a
  helm out of the bags and watched only the Helmet socket light up, checked the tooltip hints,
  dropped stacks on the world to get the destroy prompt, and confirmed the character sheet and
  bags dock side by side on touch. Character-select shots come from a real registered account on a
  local server.
- Two tests caught real bugs while writing them: a `pointercancel` was running the drop (a
  system-stolen touch would have equipped or destroyed a stack the player never dropped), and the
  touch drag had no visible target on a phone until the pairing landed.
- Known: `tests/browser/armory_mobile_layout.browser.test.ts` fails on my machine
  (394.6px vs 391px) and fails identically on the base branch with this work stashed, so it is an
  environment difference, not a regression from this PR.

## Screenshots / recordings

Bag drag to equip, desktop (tooltip hints: "Right-click to destroy" becomes the two drag hints):

| Before | After |
|---|---|
| before | after |

Dragging a helm lights only the socket that accepts it:

| Before | After |
|---|---|
| before | after |

Mobile landscape: the bags used to cover the paperdoll; now the two dock 50/50 so the touch drag
has a socket to land on.

| Before | After |
|---|---|
| before | after |

Character select, the delete control:

| Before | After |
|---|---|
| before | after |

### Quality

- [x] **The full gate passes.** `npm run gate` is green apart from the pre-existing browser-suite
      failure noted above, with decisive new tests for every changed behavior.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop plus a landscape phone viewport (the in-game
      client is landscape-only). The delete control and the drag affordances keep the 40x40 touch floor.
- [x] **Accessible.** The paperdoll sockets keep their semantics, the delete control keeps its
      `aria-label` and its disabled explanation, the drag adds no motion on hover/focus, and the
      touch ghost is `aria-hidden`.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Two new English keys in
      `i18n.catalog/hud_chrome.ts` (the two drag hints) plus their five required non-Latin fills
      (M16). The equip refusals reuse the sim's own wording through `tSim`.
- [x] Numbers go through `formatNumber` (the level in the refusal toast).
- [x] The new sim refusal ("That does not go in that slot.") is registered in `sim_i18n.ts` in the
      same change; `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] Generated files were regenerated by their build steps (`npm run i18n:gen`, `npm run i18n:hash -- --write`),
      never hand-edited.


EnriqueGF:
Added manual bag order on top of this branch (`feat(sim): manual bag order, drag a stack onto any bag cell`).

Dragging a stack onto another bag cell now puts it there: a drop on a filled cell swaps the two stacks, a drop on a free square moves the stack to the end (the inventory array is dense, so it cannot hold a hole). The order IS that array, which is serialized with the character, so the arrangement survives a relog. The new `inv_move` command is authoritative like every other inventory verb: the sim re-validates both indices against the live bag.

Reordering is only offered in the raw-array view (no sort, no filter, no search). Under a quality/name sort or a category chip the grid is a derived view where a cell names no position, so those cells are not stamped as drop targets and the drop is refused rather than guessing an index and moving the wrong stack.

Verified by driving the real client: a real `dragstart` on the bread stack plus a real `drop` on cell 2 swapped them (`[bread, sword, potion, hide]` to `[potion, sword, bread, hide]`), and a drop on a free square moved the stack to the end.

PR #1871, opened 2026-07-13
feat(shaman): Enhancement tank kit (Stone Aegis, Anchorbound Weapon, Elemental Demand, Elemental Discharge)

ryan-foo:
## Summary

Gives Enhancement, the melee shaman spec, an off-tank toolkit, granted alongside Ancestral Strike when the Enhancement spec is chosen (never in the base kit, so Elemental and Restoration never receive it).

- Stone Aegis (`earth_shield`): a charge-limited damage-taken reduction, 20% less for the next 6 direct attacks, recast to renew. `src/sim/combat/earth_shield.ts` spends one charge per incoming direct hit; DoT ticks and reflects do not.
- Anchorbound Weapon (`earthbound_weapon`): a weapon-enchant buff that doubles all threat generated (`EARTHBOUND_WEAPON_THREAT_MULT` in `threat.ts`); swaps with the damage imbues, one weapon enchant at a time.
- Elemental Demand (`elemental_demand`): a 15 yard taunt.
- Elemental Discharge (`unleash_weapon`): discharges the active weapon enchant for damage plus an enchant-keyed rider (threat spike / fire DoT / slow), in `src/sim/combat/unleash_weapon.ts`.
- Tidal Ward (`tidal_ward`): the shaman's defensive cooldown, heal-through rather than mitigation. A tide heals ~300 every 2 sec and raises all healing received by 40% for 8 sec (a new `heal_taken_up` aura folded into `healingTakenMult`, so it lifts the ward's own HoT and any ally heal).

Engine: `SpecDef.extraGrants` lets a spec grant a multi-button kit alongside its single signature (bypassing `learnLevel`). Enhancement is the only user.

Scope note: the defensive cooldowns for the existing tanks (warrior, paladin, druid) land in a separate follow-up PR, per review. This PR is shaman-only.

## Type of change

## How was this tested?

- `npx vitest run tests/enhancement_tank.test.ts` (spec grant, Stone Aegis charge mitigation, Anchorbound threat + enchant swap, taunt, discharge riders, Tidal Ward HoT + heal boost)
- `npx vitest run tests/parity/parity.test.ts` (byte-identical: every new pipeline branch is a no-op without the new auras)
- i18n gates, `tests/architecture.test.ts`, `tests/progression.test.ts`, `tests/localization_fixes.test.ts`, `tests/guide.test.ts`; `npx tsc --noEmit` clean; biome clean
- Manual steps: N/A (sim-logic + content + i18n; no visual surface)

## i18n

English catalog entries plus the five non-Latin overlay fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) for every new ability. Generated tables + hash regenerated.


jgyy (commented):
Good shape on this one. The two genuinely new mechanics (Stone Aegis's charge-limited damage reduction, Elemental Discharge's enchant-keyed rider) each got their own module (`earth_shield.ts`, `unleash_weapon.ts`) behind the SimContext seam instead of being folded into a bigger file, and the ability/talent data itself is where content-as-code belongs. Anchorbound Weapon's threat multiplier reuses the existing enchant-exclusivity helper rather than duplicating it, which is the right call.

No new Entity/IWorld fields were needed since everything is aura-based, so the world_api parity pins are correctly untouched. Crit/hit rolls in the new discharge module go through `ctx.rng`, no Math.random or wall-clock creep.

`tests/enhancement_tank.test.ts` is thorough: spec-grant, charge consumption and the non-direct-damage exemption, enchant-swap exclusivity, and all three discharge riders get their own case.

No blocking findings. The numbers (20%/6 charges/120s on Stone Aegis, the Tidal Ward magnitude) look in line with existing defensive-cooldown patterns rather than invented from scratch, but that's the maintainer's call to confirm against the class balance sheet.

Thanks for building out the enhancement tank kit, this is a solid amount of new sim surface with the seams respected throughout.

jgyy (commented):
The design here is genuinely good: Stone Aegis and Elemental Discharge are small, pure, well-commented modules, the taunt reuses the existing generic 'taunt' effect the same way warrior/paladin/druid do, threat doubling extends the existing threatModifier() chain the same way Defensive Stance/Bear Form/Righteous Fury do, and it reuses the generic Aura.charges/.value fields rather than adding new Entity/wire surface, so it's parity-safe as written. tests/enhancement_tank.test.ts has decisive, exact-value assertions.

blocking: this branch predates PR #1912 ("defensive cooldowns for warrior/paladin/druid + druid tank parity"), which has since merged into release/v0.26.0 and touches the exact same block of damage.ts and types.ts. A straight merge here would delete the Guardian Ward / Sacred Bulwark mechanic #1912 added, along with unrelated already-shipped fixes this branch simply predates: the Hit Rating stat system (HIT_RATING_PER_PCT, hitFractionFromRating), InvSlot.slot from #1165, the hit-rating-aware Sunder miss chance, and the Sap stealth-preserve fix from #1890. None of that is mentioned in this PR's description, it's pure branch drift, not an intended change. This needs a real rebase onto current release/v0.26.0, re-layering Earth Shield's mitigation alongside Guardian Ward rather than in place of it, before it's safe to merge.

nit: Anchorbound Weapon's flat +100% threat and Stone Aegis's 20%/6-charge numbers aren't derived from an existing formula the way the threat multiplier and taunt mechanism are; worth a side-by-side sanity check against #1912's warrior/paladin/druid cooldown numbers once rebased, so specs land on a consistent power level.

Please rebase before this gets a real merge-conflict pass; happy to take another look once that's done.

ryan-foo:
Rebased in spirit via the house-style merge: release/v0.27.0 is merged into the branch as a2c721d14, and the conflicting mitigation block was re-layered by hand rather than picked one-sided. Ironhold's shipped ward applies first, Stone Aegis second (the order is commented in damage.ts: a hit the ward already reduced to zero does not spend an Aegis charge), and Guardian Ward's cheat-death still evaluates the fully mitigated amount at the lethal check, so all three coexist on one hit. The three-dot diff against the release base now shows only this PR's additions; the only deletions are the branch's own imbue-loop refactor, a spec-description reword, and two formatter rewraps, with Hit Rating, InvSlot.slot, the hit-rating-aware Sunder miss chance, and the Sap stealth-preserve fix all surviving verbatim (grepped the diff's deletions for each symbol: zero hits). Parity goldens pass unchanged, and enhancement_tank, tank_defensive_cds (the #1912 suite), combat_damage, architecture, and the i18n guards are green with tsc clean. On the numbers nit: Stone Aegis 20%/6 charges/120s sits deliberately below Ironhold's short-window ward since it is charge-limited and passive-uptime; happy to retune alongside a balance sheet pass if wanted.

PR #1872, opened 2026-07-13
feat(tutorial): an optional per-class starter tutorial on Dawnhaven Isle

EnriqueGF:
## Summary

An optional starter tutorial, offered **once per class the account has never played**, right after that character is created. Accepting sails the player to **Dawnhaven Isle**: a small, self-contained island that runs entirely as an **offline Sim in the client**, where a Warden walks them through nine steps, two of which are specific to their class.

It is a sandbox. The character just created sits untouched on the roster the whole time, and nothing earned on the isle follows it back. That is stated up front in the offer, because a player who thinks they wasted a run is worse off than one who skipped it.

**The class challenges are real class challenges.** A rogue is taught to spend the combo points its opener built. A warlock is taught to summon its demon. A hunter (the one class with a single ability at level 1) is taught to back off and shoot, because that is what a hunter is. Nothing asks a class for an ability it has not learned yet, and a test pins that against the real `ABILITIES` table.

**The set pieces are staged.** The Warden does not stand waiting on an empty shrine: the camera pushes in on the bare stone and she materializes on it. The first strand wolf does not idle in a field: the camera drops low into the brush, holds, and it explodes out at you while she narrates "THAT is a mob."

### How it fits the architecture

The whole feature adds **no `IWorld` member, no wire command, and no server change**. The tutorial director reads `IWorld` and folds the drained `SimEvent`s, exactly like the in-world coachmark already does, so it is an observer that never writes sim state. Staging an entity and moving the camera are the two things it cannot do from `src/ui/`, so those arrive as an injected hook bag that `main.ts` binds over the concrete offline Sim, which is the same seam the fiesta-practice hook already uses.

**The isle is invisible to the live world by construction.** Its mobs, quest and NPC merge into the global *lookup* tables (that is how the engine resolves anything by id), but nothing spatial does: `CAMPS`, `ZONES`, `PROPS` and `BUILTIN_WORLD` are untouched, so the shared rng draw order is unchanged, and the Warden is `dynamic: true` (registered, never surface-placed), the same device the Spirit Healer uses. A test asserts two built-in Sims on the same seed still spawn the same entities and leave the rng stream at the same point.

The shipped "New Adventurer" coachmark is now scoped **per class** rather than by one global flag, because onboarding is class-shaped and this tutorial is offered per class. A legacy `woc.tutorial.v1` flag migrates to "every class seen", so nobody who finished the old tutorial gets it again.

`Renderer.editorCam` is renamed to `freeCam`: it now has a second consumer, and a new `setFreeCamPose()` keeps the Three types inside `src/render/` so `main.ts` hands over plain numbers.

The on-screen objective arrow is now **one** module (a pure placement core plus a thin DOM half) that both onboardings call. I had copy-pasted it, including its behind-the-camera mirroring, which is precisely the code where two copies quietly diverge.

## Type of change

## How was this tested?

- **Commands:** `npm run gate`. Every step passes (i18n gen + freshness, malware scan, changed-files biome, SFX conformance, the full vitest suite, `tsc`, `svelte-check`, all five builds) **except one pre-existing failure it stops on**: `tests/browser/armory_mobile_layout.browser.test.ts` ("keeps a complete cosmetic card within the store body height", 394.625 vs 391). I verified that failure is **not mine**: it reproduces identically, to the same pixel, on a clean worktree of `origin/release/v0.26.0`. Nothing in this PR touches the Armory.
  Plus the suites while iterating:
  `npx vitest run tests/starter_tutorial_core.test.ts tests/starter_tutorial_script.test.ts tests/starter_tutorial_offer.test.ts tests/starter_tutorial_isle.test.ts tests/tutorial_cinematic.test.ts tests/tutorial.test.ts tests/architecture.test.ts tests/parity tests/progression.test.ts` (83 passing).
- **Manual steps:** booted the isle in a real browser (`npm run dev`, then the new DEV-only `?tutorial=<class>` param) and walked it end to end on desktop and on a phone landscape viewport, watching for console errors. `node scripts/starter_tutorial_shots.mjs --class mage` automates that walk, drives both reveals, captures both viewports, and **fails on any console error**, so it doubles as the isle's smoke test.

Running it in a real browser is what found the bugs worth mentioning, and each now has a test:

- The objective line is keyed by step id for the shared spine but by objective *kind* for the two class steps. I had that backwards, which threw on an untracked i18n key and blanked the whole card. Now every step of every class is asserted to resolve all of its copy.
- The wolf pack could aggro the player *during* the reveal cinematic. Caught by a test that asserts the pack sits further from the ambush spot than its own aggro radius.
- The placed-asset instancer normalizes every catalogue GLB by its **longest axis** to a fixed 2.2yd and only then applies `scale`, so a raw `scale: 1` made a bedroll the size of an oak. The map now authors heights in yards and inverts that math through a table of measured source sizes; a test pins that every prop lands at a believable height, which is what catches a stale row.
- The arrival cinematic played once and never again: it was gated on the live world's one-time "already seen it" marker, which is right for a welcome and wrong for a set piece. It now replays on every visit, and a tutorial session never writes that marker (which would also have retired the first-spawn intro for an ordinary offline character sharing its class and name).
- The yard reveal was a camera move, not a reveal: the dummies were camp-spawned, so the camera flew over to three targets that were already standing there. They are staged now, and the scene fires on accepting the quest rather than on walking up to them.
- A practice dummy rendered as a *person* (it fell through to the humanoid family fallback), so the tutorial asked players to spend their first fight beating up what looked like a villager, next to the actual straw posts.
- The isle was not an island. The built-in world's rim mountains rise at `|x| > 150` regardless of the zone band, so the sea was ringed by a dark wall about 100yd out. Fixed with a wide flat terrain edit (edits apply last, after the rim), sized past the fog plane, because fog fades geometry toward the fog *colour*, it does not remove it.

## Screenshots / recordings

Committed under `docs/screenshots/starter-tutorial/` (desktop 1600x900 and phone landscape).

| | Desktop | Mobile |
|---|---|---|
| The Landing |  |  |
| The Warden materializes |  |  |
| The camera sweeps onto the empty yard |  |  |
| ...and the dummies land in it |  |  |
| The wolf bursts out |  |  |

---

### Quality

- [x] **The gate passes on everything this PR touches** (see the caveat above: it stops on a pre-existing Armory browser-layout failure that reproduces on the base branch). Decisive tests, not pins: a step cannot advance without its *own* evidence, one class's signature ability never credits another's, a scene is continuous across every shot boundary and lands exactly on the pose it opened from, and a skipped reveal still fires every cue it had not reached, so cutting one short can never strand the tutorial.

### Cross-platform

- [x] **Tested on desktop and mobile.** Phone landscape verified (the game refuses portrait). The coachmark reuses the shipped `.tut-card` shell, which already carries its mobile rules; the offer modal reuses the existing `.confirm-modal` pattern.
- [x] **Accessible.** The card is a `role="status"` polite live region, never a focus trap (it is a coachmark, not a dialog). Reduce motion is honored the way the spawn cinematic honors it: **no camera movement at all, but every cue still fires**, so a reduce-motion player still gets the Warden and the wolf and the tutorial can never dead-end on a scene they never saw. Escape cuts a reveal short.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** English lands in the two en-only catalog domains (`hud_chrome`, `shell`); the five non-Latin locales carry real fills in the same change, per M16. No locale overlay was touched beyond those.
- [x] The isle's sim-side content strings (NPC name/title/greeting, quest text, zone and POI labels) are registered in `src/ui/world_entity_i18n.ts` in this same change; `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] No secrets or `.env` committed; `ALLOW_DEV_COMMANDS` is untouched. The new `Sim.spawnStagedMob` / `spawnStagedNpc` are reachable **only** from the client host that owns an offline Sim: there is no wire command and no `dispatchMessage` case, so a network client cannot ask the authoritative server to spawn anything.
- [x] Generated files were regenerated by their owning build step, never hand-edited.

## Notes for the reviewer

- **One new asset** (`public/models/tutorial/training_post.glb`, 154 KB, the straw practice dummy) was generated with the repo's Tripo pipeline and lands in its own `models/tutorial/` group, so the live-world `models/props` budget is untouched and live players never download it (the isle's placements load lazily). Everything else on the isle is built from the 864 GLBs already in the catalogue.
- `zoneAt()` now reads the **active** world's zones instead of the built-in array, so a custom map reports its own zone name and level range instead of Eastbrook's. Byte-identical for the built-in world (the active content *is* that array), and it fixes the same wart for the editor's play-test.


EnriqueGF:
Thanks for this one :) It is an unusually well-built feature: the hook-bag seam instead of a new `IWorld` member is the right call, the arrow de-duplication is exactly the extraction the repo asks for, the per-class script pinned against the real `ABILITIES` table is a genuinely good test, and the write-up is honest about what it changed and why. The screenshots (desktop and mobile) are present and the base branch is correct (`release/v0.26.0`).

I verified the invariants against the code rather than the description, and I am requesting changes. The headline issue is that the isolation invariant the PR rests on is not true on one path.

## What I verified independently

- Base branch is `release/v0.26.0`. Gate passes.
- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`, and all nine new suites: green (278 passing). `tests/parity` goldens are unregenerated by this diff.
- I captured the built-in world fingerprint on `origin/release/v0.26.0` and on this branch: `count=410 hash=-1920954820 rng=0.393860223935917`, identical. The "no rng shift, nothing placed live" claim is genuinely true. The isle merge is spatially inert, and `spawnStagedMob` / `spawnStagedNpc` draw zero rng and are unreachable from any wire path.
- No em/en dashes or emojis in added lines. i18n follows the contributor policy, and the M16 non-Latin fills are real translations.

## Blocking

**1. `q_dawnhaven_wolves` is injectable into a live quest log on the authoritative server.** `src/sim/data.ts:225-231` states the quest is "unreachable in the live world" because only the Warden offers it. That holds for `acceptQuest`, which enforces giver proximity. It does not hold for `acceptLinkedQuest` (`src/sim/quests/quest_commands.ts:137-165`), which is reached by the `qlinkaccept` wire command (`server/game.ts:3622`) and performs **no giver check at all**. It only requires: quest exists, not `retired`, not `shareable: false`, both players in the same party, and `questState === 'available'`. The quest record (`src/sim/content/tutorial/isle.ts:257-271`) sets none of `minLevel`, `requiresQuest`, `retired`, or `shareable: false`, so it computes as `available` for every live character.

Today this grants nothing: accept pays no XP or copper, turn-in does enforce a nearby NPC, and the Warden is `dynamic: true` so she never spawns live. The result is a stuck junk quest in your own log. But the invariant as written is false, the comment tells the next author the quest is unreachable, and it is now one content edit (an item reward, a `collect` objective, a live spawn) away from being a real economy exploit. Please hard-gate it rather than relying on NPC proximity: `shareable: false` at minimum, or keep the quest out of the global `QUESTS` table entirely.

**2. Reduce-motion players never see any reveal narration.** The accessibility claim is that reduce motion means "no camera movement at all, but every cue still fires". The entity cues do fire, so the tutorial cannot dead-end, and that part is correct. But `play()` returns on the reduce-motion arm (`src/game/tutorial_scenes.ts:110-115`) without ever setting `this.sceneId`, so `active()` is permanently `null`. The card picks its body copy with `const scene = this.hooks?.active() ?? null` (`src/ui/starter_tutorial.ts:315`) and only renders `hudChrome.starterTutorial.scenes.*` when `scene` is non-null. A reduce-motion player therefore never receives any of the three narration lines, including the wolf's teaching line ("THAT is a mob ... It has a health bar. Empty it before it empties yours."). The step body still renders, so they are not stranded, but instructional copy is being withheld behind an accessibility setting. Please surface the narration on that path and add a test asserting the reduce-motion run yields the same copy set as the animated one.

## Should fix

**3. A reveal cannot be skipped on touch.** `TutorialScenePlayer.skip()` documents "Escape, or a burst of taps on touch" (`src/game/tutorial_scenes.ts:138`), but its only caller is a `keydown` / Escape listener (`src/main.ts:1048-1058`). A phone has no Escape key, so mobile players sit through every reveal. `recordSkipTap` is already imported in `main.ts` for the spawn intro, so this is a small fix.

**4. The determinism test does not test what it claims.** `tests/starter_tutorial_isle.test.ts:294-315` is titled "the built-in world spawns the same entities, in the same rng order, **as before**", but it compares two Sims built from the same current code. Any draw-order shift would move both identically and the test would stay green. It proves instance-to-instance determinism, not the absence of a regression. The real protection is `tests/parity` (golden traces plus the rng draw-order log), which is decisive and did hold. Either pin a literal baseline (`count=410`, `rng=0.393860223935917`, both captured above) or drop the claim and point at `tests/parity`. The sibling assertions in that test (no `dawnhaven_` entity spawns live, no isle camp in `CAMPS`, `ZONES` exclusion) are decisive and worth keeping.

**5. The shipping objects are untested.** The pure cores are well covered, but three pieces of code that actually run in production have no tests:
- `src/game/tutorial_scenes.ts` (`TutorialScenePlayer`, the object that spawns the Warden, the dummies and the wolf): zero tests. The play-once set, the no-reentry guard, the reduce-motion arm, `skip()`, and `forceCue()` are all unasserted. Its deps are fully injected, so a fake deps bag closes this cheaply, and it would have caught finding 2.
- `src/ui/starter_tutorial.ts:110-127` (`onEvents`): the event-to-evidence fold that the entire core suite assumes. Note `ABILITY_BY_NAME` (line 82) maps back from the ability's **display name**, so renaming an ability's display string silently stops a class's `signature` step forever.
- `src/ui/tutorial.ts:98-122`: the legacy `woc.tutorial.v1` to `woc.tutorial.classes.v1` migration runs for every existing player and has no test at all. A bug here either re-onboards everyone who finished the old tutorial or permanently suppresses the coachmark.

**6. `main.ts` grew a home.** The root `CLAUDE.md` calls `main.ts` a firewall that must never accrete top-level functions. This adds roughly 195 lines, including `askStarterTutorial` (`src/main.ts:6522-6551`), a complete modal with element lookups, text writes, focus and listener plumbing. That belongs in its own `src/ui/` module. The `TutorialScenePlayer` deps literal legitimately needs main's concrete `Sim` and `renderer`, but the composition itself would read better as an `installTutorialScenes({...})` factory exported from `src/game/tutorial_scenes.ts`, leaving `main.ts` a one-liner. The Escape listener is input logic and belongs with the scene player (it is also never removed on teardown).

**7. `objective_arrow.ts` does raw per-frame DOM writes.** `paintObjectiveArrow` (`src/ui/objective_arrow.ts:27-33`) sets `style.display`, `left`, `top` and `transform` directly every frame with no elision, and reads `window.innerWidth` / `innerHeight` on the hot path. This escapes `tests/hud_perf_budget.test.ts` only because the file is not named `*_painter.ts`. The code was moved verbatim, so it is a pre-existing sin, but the point of making this the one shared arrow is the moment to route it through the `PainterHost` writer facet.

**8. The offer modal has no focus trap or focus return.** `src/main.ts:6526-6551` sets `aria-modal="true"` in the markup but the JS only calls `.focus()` on the accept button. No Tab trap, no focus return to the opener, no Escape close. The repo's pattern here is a `FocusManager` instance (see `walletFocusManager`).

## Minor

- The branch is behind `release/v0.26.0` and currently conflicts. The conflicts are mechanical (three generated i18n artifacts plus `tests/architecture.test.ts`), but it needs a rebase before merge.
- `src/sim/content/tutorial/isle.ts:13` cites `tests/starter_tutorial_isolation.test.ts`, which does not exist (the real file is `tests/starter_tutorial_isle.test.ts`). The docs anchor rule asks for pinned tests that exist.
- `src/sim/tutorial_stage.ts` is not registered in the `src/sim/CLAUDE.md` module map, which that file explicitly requires for every `SimContext` consumer.
- The comment at `src/sim/data.ts:213-215` says "The isle's own WorldContent carries the PLACED copy" of the Warden. It does not: `DAWNHAVEN_WORLD.npcs` is the same record and is `dynamic: true`, so she is staged by the scene player and never surface-placed on the isle either. Worth correcting, since it means an isle booted without a ticking scene player has no quest giver.
- `src/main.ts:3145` does `if (world) setActiveWorldContent(world)`, so the module global is set but never reset. Harmless today because the only exit is `location.reload()`. Making it unconditional is a no-op now and immune later.
- `TutorialState.skipped` (`src/ui/starter_tutorial_core.ts:57`) is never set to `true` anywhere; the Skip button calls `finish()`. Either wire it or drop it.
- `zoneAt` (`src/sim/data.ts:398-403`) would return `undefined` for a custom world with an empty `zones` array. Byte-identical for the built-in world and the server never sets custom content, so this is editor-only, but a `?? ZONES[ZONES.length - 1]` fallback closes it.

Findings 1 and 2 are the ones I would want closed before merge; the rest are mostly small. The architecture of this feature is sound and I would be glad to re-review once these are addressed.

EnriqueGF:
Thanks for this one :) Second pass, on top of the review already on this PR. The head commit has not moved since that review (`52b0a54b`, 2026-07-13; the review landed 2026-07-14), so all of its findings are still open. I have re-confirmed the two blockers against the code and I am not going to restate the rest. What follows is only what is new, plus one correction.

Verdict: changes requested.

## What I ran myself

- Base branch is `release/v0.26.0`. Correct.
- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`, `tests/guide.test.ts`, `tests/deeds_content.test.ts`, `tests/world_api_parity.test.ts`, `tests/hud_perf_budget.test.ts` and all nine new suites: green (277 + 341 passing). Parity goldens are unregenerated, so the live draw order really is untouched.
- Biome on the changed files: clean.
- Security claims hold: `?tutorial=` is behind `import.meta.env.DEV` (`src/main.ts:8612`), and `spawnStagedMob` / `spawnStagedNpc` have no wire or `dispatchMessage` path. Isle content does not leak into the `/wiki` guide.
- Still reproducing on HEAD: the `qlinkaccept` quest-injection gap, and the reduce-motion narration gap.

## Correction to the earlier review

The merge conflict is not "three generated i18n artifacts plus `tests/architecture.test.ts`". The branch is **48 commits behind `release/v0.26.0` and conflicts in 43 files**, including `src/main.ts`, `src/sim/sim.ts`, `src/sim/data.ts`, `src/render/renderer.ts`, `src/ui/hud.ts` and `src/styles/hud.css`. That is a substantive rebase over source you are actively changing, not a mechanical one, and it means the `npm run gate` result quoted in the description was taken against a base that has since moved. Please rebase and re-gate.

## Blocking (new)

**A. The isle is not a self-contained island: the built-in world's furniture spawns on it.** The `Sim` constructor places mailboxes (`src/sim/sim.ts:1543`), dungeon doors, Groundskeeper Bram and the Spirit Healers (`spawnOverworldSpiritHealers`, `src/sim/sim.ts:1601`) from the **built-in static tables**, not from `cfg.world`. `WorldContent` simply does not cover them. I booted the isle Sim directly and dumped its roster. Alongside the four Strand Wolves and the player, it contains:

```
npc    spirit_healer  The Pale Keeper     (-14,-14)
npc    spirit_healer  The Pale Keeper     (4,-56)
object mailbox        Mailbox             (7,-8)
npc    groundskeeper_bram Groundskeeper Bram (-6,-82)
+ 5 more Pale Keepers and 5 dungeon doors out over open water
```

The isle's landmass is a dome roughly 45yd in radius (`src/sim/content/tutorial/isle.ts:17`, shore blocker ring at 52). The player starts at `(0,40)`. So a Pale Keeper angel stands at distance 20 from the origin, **on the isle**, beside the path the player walks, and a Mailbox stands at distance 11, also on the isle. A second Pale Keeper hovers in the sea just past the shore ring. This is directly contrary to the "small, self-contained island" the PR is built around, and it is visible to every player who takes the tutorial. The screenshots do not show it because they are all framed on the shrine and the yard.

Related, and the reason it matters beyond cosmetics: `nearestOverworldGraveyard` (`src/sim/spirit.ts:62-75`) scans the static `OVERWORLD_GRAVEYARDS` list and never reads the active world, so `DAWNHAVEN_ZONE.graveyard = {x:0,z:16}` (`isle.ts:107`) is **dead data**. A level-1 character dying to three wolves (very possible) releases to `gy_eastbrook (-14,-14)` instead. Meanwhile `graveyardReadout` (`src/sim/entity_roster.ts:236-239`), now that `zoneAt` reads the active world, cheerfully tells them their spirit returns to the Dawnhaven graveyard at `(0,16)`. The readout and the behavior disagree.

Please either gate the built-in furniture spawns on the built-in world, or route them (and graveyard selection) through `cfg.world`, and add a test asserting the isle Sim's roster contains nothing outside the shore ring and that a tutorial death releases to the isle's authored graveyard.

## Should fix (new)

**B. The offer is burned before the player answers.** `src/main.ts:7930-7931` calls `rememberTutorialAnswer(store, cls)` **before** awaiting `askStarterTutorial(cls)`. `askStarterTutorial` returns `false` immediately when the modal element is missing (`src/main.ts:6531`), and a player can also reload while the modal is open. Either way the class is permanently recorded as answered and the offer never returns, on any device. Record the answer inside the `close()` continuation, once the player has actually answered.

**C. A hostile mob is spawned into a scene that has the player's movement locked.** `wolfReveal` fires `CUE_SPAWN_WOLF` at the start of shot 2 (`src/game/tutorial_scenes.ts:257`), and roughly 3 more seconds of scene run after it, with `suspendInput(true)` mapped to `input.setSuspendMovement(true)` (`src/main.ts:1012`) and the camera off the player. The staged wolf is placed at the ambush point next to the player. The existing test only asserts the **camp pack** sits outside its aggro radius; it says nothing about the staged wolf. Spawn it on the last cue, hold it passive until the scene ends, or end the scene on first incoming damage.

**D. A sync comment that is not true, and an unpinned reach radius.** `src/ui/starter_tutorial_core.ts:139-141` says its private `WARDEN_REACH = 6` is "kept in sync with the script's `WARDEN_REACH_YARDS` by the script test". It is not: the only assertion is `expect(WARDEN_REACH_YARDS).toBeLessThanOrEqual(6)` (`tests/starter_tutorial_script.test.ts:114`), which never touches the core's copy. The core constant can be set to anything from 3 to 99 with the whole suite green, and `WARDEN_REACH_YARDS` has no `src/` consumer at all. Export the core's constant, assert the two are equal, and add a boundary case.

**E. The mirrored content ids are self-compared.** `TUTORIAL_DUMMY_ID`, `TUTORIAL_WARDEN_ID`, `TUTORIAL_LOOT_ITEM_ID` and `TUTORIAL_WOLF_ID` (`src/ui/starter_tutorial_script.ts:26-29`) are hand-copied string literals "mirrored from" `isle.ts`, and the tests that use them compare the constant against itself. A rename on either side ships green while `targetMob` / `lootItem` never complete and the arrow never finds the Warden. Assert each id resolves in the real `MOBS` / `NPCS` / `ITEMS` table. `TUTORIAL_ANSWERED_KEY` (`src/game/tutorial_offer.ts:17`) has the same problem: it is a persistence key with no literal pin, so a rename silently re-offers the isle to everyone who already declined. (`TUTORIAL_WOLF_ID` also appears to be dead: no `src/` consumer.)

**F. `maybePlayScene` is untested, and it is the headline fix of the last commit.** `src/ui/starter_tutorial.ts:224-265` owns both `forceCue` safety nets, whose entire purpose is that the tutorial cannot dead-end, plus the scene ranges and the "yard reveal fires on accepting, not on proximity" ordering. None of it is exercised. Same for `snapshot()` (`src/ui/starter_tutorial.ts:175-198`), the IWorld-to-machine adapter: it assumes `p.auras[].id === ability.id` (every `auraUp` step), `e.kind === 'mob' && e.ownerId === playerId` for `hasPet` (the warlock's whole mastery step), and the inventory slot shape, none of which is pinned. One integration test that boots the isle Sim, casts each class's level-1 signature, and asserts the objective flips would close this and the known `onEvents` gap together.

**G. Per-frame cost in the tutorial path.** `src/ui/starter_tutorial.ts:135-155,175-220` walks the entity map up to four times per frame (`findNpc` and `nearestMob` each do a full `entities.values()` scan), allocates a fresh snapshot plus `auras.map()` plus an `inventory.filter().reduce()` every frame, and `advanceTutorial` (`starter_tutorial_core.ts:163-187`) rebuilds a `Set` and an array every frame even when nothing changed. `Renderer.setFreeCamPose()` (`src/render/renderer.ts:820-828`) allocates two `THREE.Vector3` per call and is called every frame for the duration of a scene. Also, `StarterTutorial` does raw unelided `document.createElement` / `textContent` / `style` writes from a per-frame path rather than going through a `PainterHost` writer facet, which is the same escape from `tests/hud_perf_budget.test.ts` as the known `objective_arrow.ts` one (neither file is named `*_painter.ts`, so the source scan never sees them).

**H. The whole coachmark is one `aria-live="polite"` region** (`src/ui/starter_tutorial.ts:278-280`), so every objective tick ("1 of 3" to "2 of 3") re-announces the entire multi-sentence Warden speech to a screen reader. Put the live region on the objective row, not on the card root.

## Minor (new)

- `src/sim/data.ts:189-198`: `buildSourceIndex` and `buildHeroicVariants` scan the **whole** merged `MOBS` table, so the isle's `wolf_fang` drop now feeds the live item-source index. It is a no-op today only because `old_greyjaw` already dominates that item's level and `wolf_fang` is heroic-ineligible. A future isle mob with a higher `maxLevel` than a shared item's live source would silently move that item's derived level for live players. Worth a test asserting the derived tables are identical with and without the Dawnhaven merge.
- `src/main.ts:3145`: `if (world) setActiveWorldContent(world)` sets the module global and never restores it. Today only `location.reload()` saves you. Making it unconditional (`setActiveWorldContent(world ?? null)`) is a no-op now and immune later. This is the same point the earlier review raised; I am repeating it only because `zoneAt` now makes the global feed gameplay (fishing table selection, deeds, the graveyard readout), not just terrain.
- `hudChrome.starterTutorial.objective.hunt` reads "Strand Wolves slain", but the `hunt` step's objective is `questState: 'ready'` with `total: 1`, so it never shows the 0/3 kill progress the label promises. And `objective.rangedHit` hard-codes the English ability name inside a translated value where every sibling interpolates `{ability}`.
- `src/ui/starter_tutorial.ts:331-333` casts to `TranslationKey`, and the catalog has no `scenes.finale` key. Unreachable today only because the finale path returns before `render()`. Type the scene keys as a literal union so the compiler guards it rather than luck.
- `tests/starter_tutorial_isle.test.ts:196,209`: the unknown-id arm of both staged spawns is unasserted; nothing pins that `spawnStagedMob('nope', ...)` returns `null`.
- `src/sim/data.ts:212-214`: the comment says "The isle's own WorldContent carries the PLACED copy" of the Warden. It does not; `DAWNHAVEN_NPC_REGISTRY` is the same `dynamic: true` record, so she is staged on the isle too. (Also raised in the earlier review; `isle.ts:158-168` states it correctly, only `data.ts` lies.)

Finding A is the one I would want closed first: it is the invariant the whole feature is sold on, and it is visibly false in the shipped isle. B and C are player-facing. The architecture underneath (the hook-bag seam, the arrow extraction, the class-shaped script pinned against the real `ABILITIES` table) is genuinely good work, and I would be glad to re-review once these are addressed.

PR #1873, opened 2026-07-13
feat(sfx): add the reptile mob voice family

jamiecypher:
## Summary

Adds a new `reptile` mob voice family, step 1 of a deliberate 2-PR split. No mob is data-tagged `reptile` yet: this PR lands the family definition (catalog entries, render model default, voice-cue wiring) together with real recorded audio, satisfying the completeness gate. A follow-up PR will retag `deepfen_spearjaw` (currently `beast`, but modeled as a velociraptor per the comment on `mob_spearjaw` in `manifest.ts`) as a pure one-line content change, no further code needed once this lands.

- 3 new catalog entries (`mob_reptile_aggro`/`attack`/`death`) via the shared `mob()` template in `sfx_prompts.mjs`, plus `'reptile'` added to `MOB_VOICE_FAMILIES`
- `FAMILY_KEYS.reptile` defaults to the existing `mob_spearjaw` model (already in the manifest, currently only reachable via a `MOB_KEYS` override)
- `MOB_VOICE_CUES.reptile` added to `combat_sfx.ts`
- 7 recorded audio takes (aggro x2, attack x3, death x2), run through `sfx:conform`, verified in spec

`hurt` is deliberately not included here: it isn't a generated action in the `mob()` template yet (that's a separate, larger PR touching every family's `hurt` coverage at once, tracked separately). Including a `mob_reptile_hurt` file now would create an orphaned, uncatalogued file, exactly the class of bug #1854 is meant to catch.

## Type of change

## How was this tested?

- `npx tsc --noEmit` clean
- `npx vitest run tests/sfx_manifest.test.ts tests/combat_sfx.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts tests/architecture.test.ts` — 93 tests passing
- Updated the decisive catalog-inventory test (`sfx_manifest.test.ts`) to the new 141-key count and added `mob_reptile_attack` to its explicit presence check
- `npm run sfx:conform` — all 7 reptile files pass, in spec
- `npm run sfx:manifest` — clean rebuild, 141 clips, zero errors
- No `combat_sfx.test.ts` case for `mobVoiceFamily`/`mobVoiceCue` resolving to `reptile` via real mob data, since no mob carries that tag yet (every existing family test uses a real templateId). The `satisfies Record<string, Record<MobVoiceAction, SfxId>>` constraint on `MOB_VOICE_CUES` already compile-time-proves the new entry resolves to valid, real `SfxId`s. A real data-driven test belongs in the follow-up retag PR, once `deepfen_spearjaw` actually carries the tag.

## Screenshots / recordings

N/A — no visual change. Audio-only; `deepfen_spearjaw` will keep sounding like the generic `beast` fallback until the follow-up PR retags it.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [ ] Builds succeed. (ran `sfx:manifest`/`sfx:conform` directly; full `npm run build` not run locally this pass)

Cross-platform
- ~Tested on desktop and mobile.~ N/A, audio content + routing only

Localization (i18n)

Hygiene


EnriqueGF:
Thanks for this, and for splitting the work so deliberately :) The audio is clean, the catalog regenerates reproducibly, and the PR body is unusually honest about what it does and does not cover. Verdict is changes requested, mainly for one cross-PR hazard and one reachability concern.

Base branch gate: PASS. This targets release/v0.26.0, which is correct.

What I verified locally on this branch (not taken from the description):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/sfx_manifest.test.ts tests/combat_sfx.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts`: all green.
- `npm run sfx:conform`: 145 files in spec, 0 conformed.
- `npm run sfx:manifest`: regenerates `src/game/sfx_manifest.generated.ts` and `public/audio/sfx/runtime-pack.json` byte for byte identical to what is committed, so no generated file was hand edited.
- `npm run build`: green. Your checklist left "Builds succeed" unchecked, it does pass.
- Biome on the changed files: clean. No em dashes, en dashes, or emojis in the diff.

Findings, most severe first.

**1. Blocking for merge order: this PR and #1874 break each other.** They are individually MERGEABLE against release/v0.26.0, but they are siblings rather than stacked, and they collide in four files (`scripts/sfx/sfx_prompts.mjs`, `src/ui/combat_sfx.ts`, `src/game/sfx_manifest.generated.ts`, `tests/sfx_manifest.test.ts`). I built the merged tree to check. `sfx_prompts.mjs` auto-merges cleanly and silently produces a broken entry: with #1874 in, `mob()` takes a 6th `hurt` argument, so this PR's 5-argument reptile call (`scripts/sfx/sfx_prompts.mjs:391`) emits a `mob_reptile_hurt` catalog key whose prompt literally contains the word "undefined", and `npm run sfx:manifest` then hard-throws `missing sampled SFX file: public/audio/sfx/mob_reptile_hurt.mp3`. On top of that `tsc` fails, because `MOB_VOICE_CUES.reptile` (`src/ui/combat_sfx.ts:51`) has no `hurt` member while the `satisfies Record<string, Record<MobVoiceAction, SfxId>>` constraint now requires one. Please agree the landing order with #1874 and have whichever lands second carry the reconciliation in the same change: the reptile hurt prompt argument, `mob_reptile_hurt.mp3`, the `MOB_VOICE_CUES.reptile.hurt` entry, a regenerated manifest and runtime pack, and the updated count pins.

**2. Major: the family is unreachable by any mob, so nothing in the shipped product can play it.** `MOB_VOICE_CUES.reptile` (`src/ui/combat_sfx.ts:51`) and `FAMILY_KEYS.reptile` (`src/render/characters/manifest.ts:1132`) are both dead entries until a mob is data-tagged `reptile`, and the seven audio files ship into the release regardless. The repo rule is explicit that we do not add abstraction for a hypothetical future need. The retag you describe is a one-line content change, so please fold it into this PR: it makes the family reachable and lets the suite prove the wiring end to end. If you have a concrete reason to keep them apart, please state it in the body and link the follow-up issue.

**3. Major: test coverage for this change is a count bump.** `tests/sfx_manifest.test.ts` moves 138 to 141 and adds `mob_reptile_attack` to the presence list. That passes even if `MOB_VOICE_CUES.reptile` pointed at the wrong keys, and `mob_reptile_aggro` / `mob_reptile_death` ride on the count alone (any three keys satisfy 141). Once the retag from point 2 is in, add decisive assertions in `tests/combat_sfx.test.ts`, for example `expect(mobVoiceCue('deepfen_spearjaw', 'attack')).toBe('mob_reptile_attack')` and the same for aggro and death.

**4. Minor: `FAMILY_KEYS.reptile` will stay dead even after the retag.** `visualKeyFor` checks `MOB_KEYS[templateId]` before the family fallback (`src/render/characters/manifest.ts:1164-1167`), and `deepfen_spearjaw` already has an explicit `MOB_KEYS` override at `manifest.ts:1072`. So the new family default only matters for some future reptile mob that has no override. Worth saying so in the comment, since the current one implies it becomes live at retag time.

**5. Minor: the `mob()` template silently accepts a missing argument** and interpolates "undefined" into the generated prompt, which is what turns point 1 from a loud failure into a subtly wrong catalog entry. An arity guard in the template would make a forgotten action throw at generation time.

Screenshots: correctly N/A, audio only. No i18n, sim, server, or persistence surface touched, so those invariants are not in play.

Happy to re-review as soon as the merge-order plan is stated and the reptile family is reachable.

jamiecypher:
Thanks for the thorough review. Addressed in the latest push:

- Folded the `deepfen_spearjaw` retag into this PR (was previously going to be a follow-up) so the family and its first real mob land together.
- Added `reptile` to the `MobFamily` union in `src/sim/types.ts`, which the retag needs to compile.
- Added a decisive test pinning `deepfen_spearjaw` to the `reptile` family end to end.

On the cross-PR collision you flagged: #1874 builds directly on top of this branch's tip so what's pushed there reflects the true combined tree, not two independently-green diffs that break each other on merge. Once this lands, I'll rebase #1874 back onto the real updated `release/v0.26.0` to clean its diff down to just its own delta.

EnriqueGF:
Thanks for turning this around so fast, and for folding the retag in :) The retag is the right call and it does make the family reachable, but it also pulled in a consumer of `MobFamily` that the audio-only version of this PR never touched, and that consumer is now broken. Verdict is changes requested.

Base branch gate: PASS. Still targets `release/v0.26.0`.

Re-verified locally on f7a1ee0 (not taken from the description):
- `npx tsc --noEmit`: clean.
- Guard + covering suites green: architecture, localization_fixes, world_api_parity, sfx_manifest, combat_sfx, sfx, sfx_conform, guide, icons, deeds_content.
- `npm run sfx:manifest`, `npm run wiki:content` and `npm run i18n:gen` all regenerate byte for byte identical to what is committed, so no generated file was hand edited and no generated artifact is stale.
- Biome clean on the changed files. No em dashes, en dashes, or emojis.
- Previous finding 1 (merge order) is RESOLVED: #1874 now contains this branch tip (f7a1ee0) and carries the reptile `hurt` reconciliation (prompt argument, `MOB_VOICE_CUES.reptile.hurt`). It must land after this one and then be rebased, as you said. Previous findings 2, 3 and 4 are resolved too: the family is reachable, the new `combat_sfx` assertions are decisive, and the `FAMILY_KEYS` comment now states the `MOB_KEYS` precedence correctly.

Findings, most severe first.

**1. Blocking: the retag makes the mob tooltip resolve a translation key that does not exist.** `src/ui/hud.ts:4470-4472` builds the family label as ``t(`guide.family.${template.family}.name` as TranslationKey)`` for every family except `demon`. The `as TranslationKey` cast is why `tsc` stays clean. There is no `guide.family.reptile.*` entry: `src/ui/i18n.catalog/guide.ts:725-765` defines a `name`/`desc` pair for all ten existing families and reptile is absent, so the key is not in the resolved bundle either. I reproduced it against the real code path: `t('guide.family.reptile.name')` hits `onUntrackedKey` (`src/ui/i18n.ts:317-322`), which THROWS `i18n: untracked key "guide.family.reptile.name"` on any non-release build, and on a release build returns the raw key, so a player who mouses over a Deepfen Spearjaw sees the literal string `guide.family.reptile.name` where the creature type should be. This is exactly the invariant the comment at `src/ui/i18n.catalog/hud_chrome.ts:1540` documents: demon is called out as "the one MobFamily with no guide.family.* bestiary entry", and reptile is now a second one, but without the `familyDemon` style special case. Please add `guide.family.reptile.name` and `.desc` to `src/ui/i18n.catalog/guide.ts` (English only, per the contributor rule), and check whether the wordy `desc` trips the M16 non-Latin fill rule the surrounding entries follow.

**2. Major: no test would have caught finding 1, and none will catch it for the next family.** Nothing under `tests/` references `MobFamily` at all. The new `combat_sfx` assertions prove the SFX cue mapping and stop there, so a family can be added, reach the HUD, and throw on a tooltip with a fully green suite. Please add an exhaustiveness guard that walks the `MobFamily` union (or every distinct `MOBS[...].family`) and asserts each one resolves a tooltip label without throwing and has a crest recipe. That is the decisive test for this change, and it is what turns the next family retag from a production bug into a red suite.

**3. Minor: no `family_reptile` crest recipe.** `crestIdForEntity` (`src/ui/unit_portrait.ts:51-54`) returns `family_reptile`, and `CREST_RECIPES` (`src/ui/icons.ts:2775-2784`) has a bespoke crest for every family from beast to dragonkin but not reptile, so the target-frame portrait falls through `CREST_RECIPES[id] ?? null` to the generic fallback plus a `console.warn` (`src/ui/icons.ts:3711-3719`). Not a crash, and demon already sits on that same path, so this is polish rather than a blocker, but the spearjaw becomes the only mob in The Drowned Litany without a real family crest. It is a one-liner next to the others.

**4. Minor, still open from the last round: the `mob()` template silently accepts a missing argument** and interpolates "undefined" into the generated prompt. An arity guard would make a forgotten action throw at generation time.

Behavior deltas I checked and cleared, so you do not have to: taming is unaffected (`tameError` rejects any mob with `spawnPos.x > DUNGEON_X_THRESHOLD` = 600, and delve mobs spawn from `DELVE_X_MIN` = 4800, so the spearjaw was never tameable and the beast/spider gate never applied to it); flee behavior is unaffected (`FLEEING_FAMILIES` is humanoid/burrower/mudfin/troll, so neither beast nor reptile flees or rallies same-family allies); social pull is unaffected (the pull loop matches `templateId`, not family, and `SOCIAL_PULL_RADIUS` only overrides mudfin, so the radius stays at the default 5); polymorph is unaffected (undead/gorrak only); and the visual model is unchanged because `deepfen_spearjaw` keeps its explicit `MOB_KEYS` override to `mob_spearjaw`. So the retag's only functional surfaces are the voice cues, the tooltip label, and the crest. The first is right, the other two are findings 1 and 3.

Screenshots correctly N/A. No sim determinism, server authority, wire, or persistence surface is in play. Happy to re-review once the reptile family has its label and a guard test that proves it.

EnriqueGF:
One addition from a second pass over the presentation seam, note level, not a new blocker.

**5. Note: `FAMILY_ORDER` in `scripts/wiki/build_content.mjs:238-249` has no `reptile` slot.** This is harmless in this PR: the bestiary publishes only camped overworld creatures and `deepfen_spearjaw` is delve-only, which is why `npm run wiki:content` still regenerates clean. It matters the day a reptile is camped in the open world, and `assertFamiliesKnown` (`scripts/wiki/family_guard.mjs`) is written to throw loudly rather than drop it silently, so the failure would be a hard content-build error. Since you will be in `src/ui/i18n.catalog/guide.ts` anyway for finding 1, adding the `reptile` slot alongside it keeps the two in step.

For completeness on one thing a second reviewer raised and I checked: the retag does NOT cost the spearjaw its tameability. The family gate in `tameError` (`src/sim/pet/pet_commands.ts:234`) does now reject it, but the mob was already untameable before this PR via the dungeon gate on the next lines (`spawnPos.x > DUNGEON_X_THRESHOLD`, 600, versus `DELVE_X_MIN` 4800), so the reachable outcome is unchanged and only the error string a hunter would see differs. No action needed there.

EnriqueGF:
Last one, a concrete test suggestion that came out of a coverage pass. Note level, nothing new to fix in the source.

**6. Note: the model-preservation claim in `src/render/characters/manifest.ts:1129-1133` is asserted only in prose.** The comment says the `MOB_KEYS` override keeps `deepfen_spearjaw` on `mob_spearjaw` after the retag. I verified that is true (the override is at `manifest.ts:1072` and `visualKeyFor` checks `MOB_KEYS` before the family fallback), but nothing pins it, so deleting that row would silently change the mob's model with a green suite. `visualKeyFor` is already exported and exercised in `tests/visual_manifest.test.ts` (see line 58 for the exact pattern), so this is a one-liner:

```ts
expect(visualKeyFor({ kind: 'mob', templateId: 'deepfen_spearjaw' } as never)).toBe('mob_spearjaw');
```

Worth adding next to the exhaustiveness guard from finding 2, since the retag is precisely the moment that claim starts carrying weight.

jamiecypher:
Thanks, both findings addressed:

1. Added `guide.family.reptile.name`/`.desc` to `src/ui/i18n.catalog/guide.ts` (English plus the five non-Latin M16 fills, since the desc is wordy), so the Deepfen Spearjaw tooltip resolves a real label instead of throwing or showing the raw key.
2. Added `family_reptile` to `CREST_RECIPES` in `icons.ts` (using the existing `fang` primitive), so the target-frame portrait gets a real crest instead of falling through to the generic fallback.
3. New `tests/mob_family_presentation.test.ts`: walks every family real `MOBS` content uses (not the closed type union) and asserts each resolves a tooltip label and has a crest recipe, excluding the documented pre-existing `demon` gap. This is the exhaustiveness guard from finding 2, would have failed on the exact regression this round caught.
4. Added `reptile` to `FAMILY_ORDER` in `scripts/wiki/build_content.mjs` while I was in the area.
5. Pinned the model-preservation claim in `tests/visual_manifest.test.ts`: `visualKeyFor({ kind: 'mob', templateId: 'deepfen_spearjaw' })` now asserts `'mob_spearjaw'` directly instead of resting on a comment.
6. The `mob()` arity guard was already in place from the previous round.

tsc, the full relevant suite, `npm run build`, and biome on the changed files are all clean.

jgyy (commented):
Good shape. The step-1/step-2 split is honest about the tradeoff, and the review history on this branch shows it actually responded to feedback rather than just re-describing the original plan: the unreachable-family concern got fixed by folding the `deepfen_spearjaw` retag into this PR instead of leaving it dangling, and the follow-up commits add real coverage (`combat_sfx.test.ts` resolving the family via the real templateId, the new `mob_family_presentation.test.ts` walking every family in use for tooltip label + crest, `visual_manifest.test.ts` pinning that the retag doesn't change the model).

Ran the full suite locally (`sfx_manifest`, `combat_sfx`, `sfx`, `sfx_conform`, `architecture`, `mob_family_presentation`, `visual_manifest`) plus `tsc --noEmit`: all green, matches what the PR body claims.

A couple of small things, nothing blocking:

- `nit`: `src/render/characters/manifest.ts` `FAMILY_KEYS.reptile` is dead code today (the comment says as much: `deepfen_spearjaw`'s `MOB_KEYS` override always wins). Fine to keep since a future reptile mob with no override needs a sane default, just flagging that there's currently no path that exercises it.
- `nit`: the merge conflicts against `release/v0.26.0` are all in the generated i18n trio (`pending.ts`, `i18n.status.summary.json`), mechanical regen churn from the base moving on, not a real conflict. A rebase + `i18n:build`/`i18n:hash --write` clears it.

The i18n addition (`guide.family.reptile.*`) is English-only in the catalog, correct for a contributor PR.

jamiecypher:
Rebased onto current release/v0.26.0 to clear the merge conflict (only the generated i18n trio: pending.ts, i18n.resolved.sha256, i18n.status.summary.json, resolved by taking ours and regenerating, standard stale-base churn). This also picks up #1883's fix for the SFX Studio publish-discovery bug, which was failing the gate on release/v0.26.0 itself, unrelated to this PR's content. npm run gate is green: all 11 steps pass.

jamiecypher:
Rebased again onto current release/v0.26.0 to pick up the #1574 conform-pipeline batch (channel policy, naming checks, asset standard) that landed after the last rebase. sfx_prompts.mjs and tests/sfx_manifest.test.ts auto-merged clean; only generated artifacts conflicted, resolved by regenerating. Confirmed reptile's clips still pass npm run sfx:check under the new rules. npm run gate green in isolation (11/11 steps).

PR #1874, opened 2026-07-13
feat(sfx): add the hurt mob voice action

jamiecypher:
## Summary

Adds `hurt` as a real, catalogued mob voice action. `MOB_ACTIONS` in the manifest builder has recognized `hurt` as a valid subfamily action token for a while, but no fixed catalog key or gameplay trigger ever used it, so it's been dead capability.

- `hurt` added as a 4th generated action in the shared `mob()` catalog template (`sfx_prompts.mjs`), so every existing family gets `mob_<family>_hurt` for free from one template change
- `MobVoiceAction` extended to `'aggro' | 'attack' | 'death' | 'hurt'`, and every family in `MOB_VOICE_CUES` (`combat_sfx.ts`) now includes a `hurt` cue
- **Real trigger wired**: a mob taking a crit previously reused its own `attack` bark as the reaction sound (`availableMobVoiceCue(tgt.templateId, 'attack')` in `hud.ts`), the only option available before `hurt` existed. It now plays the actual `hurt` cue instead, symmetric with how `player_hurt` already fires on the player's own crits.
- `hud.ts` now imports `MobVoiceAction` from `combat_sfx.ts` instead of hand-duplicating the action union locally
- 12 families' worth of `hurt` audio (15 files, 3 families with 2 takes: `spider`, `humanoid`, `dragonkin`), conformed and verified in spec

`reptile` is intentionally not included: it isn't part of the catalog yet (#1873, still under review), and the `mob()` template only iterates families that already exist. Once #1873 merges, `reptile` picks up `hurt` for free the same way every family here just did, no extra work, just needs its own `mob_reptile_hurt` audio dropped in at that point (already recorded, staged).

## Type of change

- [x] Bug fix (mob crit reaction was borrowing `attack`, not a real hurt sound)

## How was this tested?

- `npx tsc --noEmit` clean
- `npx vitest run tests/sfx_manifest.test.ts tests/combat_sfx.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts tests/architecture.test.ts` — 94 tests passing
- Updated the decisive catalog-inventory test to the new 150-key count, added a pin asserting exactly 48 `mob_` catalog keys exist (12 families x 4 actions), and added `mob_beast_hurt`/`mob_dragonkin_hurt` to the explicit presence check
- Added a `combat_sfx.test.ts` case resolving `hurt` for 4 different families via real mob templateIds (`mudfin_murloc`, `tunnel_rat`, `forest_wolf`, `crypt_shambler`)
- `npm run sfx:conform` — all 15 new files pass, in spec
- `npm run sfx:manifest` — clean rebuild, 150 clips, zero errors
- `npx @biomejs/biome check` on every changed file: 0 errors (3 pre-existing warnings on an unrelated unused method in `hud.ts`, not touched by this change)

## Screenshots / recordings

N/A — no visual change. Audio-only plus one existing trigger's cue swapped from `attack` to `hurt`.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [ ] Builds succeed. (ran `sfx:manifest`/`sfx:conform` directly; full `npm run build` not run locally this pass)

Cross-platform
- ~Tested on desktop and mobile.~ N/A, audio content + routing only

Localization (i18n)

Hygiene


EnriqueGF:
Thanks for this one :) The dead-capability framing is exactly right, and turning the borrowed attack bark into a real hurt cue is a genuine improvement. Verdict is changes requested, mainly for a cold-buffer regression on the new trigger and for coverage of the one behavior this PR actually changes.

Base branch gate: PASS. This targets release/v0.26.0, which is correct.

What I verified locally on this branch (not taken from the description):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/sfx_manifest.test.ts tests/combat_sfx.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts`: all green.
- `npm run sfx:conform`: 153 files in spec. `npm run sfx:manifest`: regenerates the manifest and runtime pack byte for byte identical to what is committed.
- `npm run build`: green. Your checklist left "Builds succeed" unchecked, it does pass.
- Biome on the changed files: clean. No em dashes, en dashes, or emojis in the diff.
- Seam and perf gates: `tests/painter_host.test.ts` and `tests/hud_perf_budget.test.ts` green. The crit path is event driven from `handleEvents`, not per frame, so there is no frame-budget cost, and the `hud.ts` delta is a consumer edit rather than monolith growth. The mapping correctly stays in `combat_sfx.ts`.

Findings, most severe first.

**1. Major, likely regression: the first crit against each mob family will probably be silent.** Mob voice clips are `preload: "lazy"` in the generated manifest, and `Sfx.playAt` on a cold buffer starts a fetch and only replays the sound if the load resolves within 0.12 seconds (`src/game/sfx.ts:361-372`). The old cue at `src/ui/hud.ts:8645` was `mob_<family>_attack`, which is always warm because ordinary mob attacks play it constantly. `mob_<family>_hurt` has exactly one trigger, the crit itself, so on a cold cache it gets fetched and decoded on the very event that is meant to play it, and a fresh mp3 rarely lands under 120 ms. Net effect: where the player used to always hear a reaction, the first crit per family now plays nothing. Please confirm in a real client with a cold cache, and fix by warming the hurt cue alongside the family attack cue, or by falling back to the attack cue when the hurt buffer is not resident.

**2. Major: the one player-visible change in the PR has no test.** `src/ui/hud.ts:8645` is the only line that changes behavior, and nothing would go red if it were reverted, because the event-kind to `MobVoiceAction` policy lives inside the `Hud` coordinator where no Vitest reaches it. Per module-first and the bug-fix-test-first rule, extract that policy into `combat_sfx.ts` (for example `mobVoiceActionForDamage(event, target)`, returning `hurt` on a crit against a non-boss mob and `null` otherwise) and pin it in `tests/combat_sfx.test.ts` with a crit case and a non-crit case; `hud.ts` then stays a thin consumer.

**3. Major: the new test overclaims.** `tests/combat_sfx.test.ts:215` is titled "resolves a real hurt cue for every mob family" but asserts only 4 of the 12 (mudfin, burrower, beast, undead). The regression class it misses is a cross-family mis-mapping: `spider: { hurt: 'mob_beast_hurt' }` is a valid `SfxId` and would pass both `tsc` and this test. Make it table driven over every family, asserting `mobVoiceCue(templateId, 'hurt') === 'mob_' + mobVoiceFamily(templateId) + '_hurt'` and that the cue exists in `SFX_CLIPS` (that idiom is already used at `tests/combat_sfx.test.ts:158`).

**4. Major, blocking for merge order: this PR and #1873 break each other.** Both are individually MERGEABLE against release/v0.26.0, but they are siblings rather than stacked, and they collide in `scripts/sfx/sfx_prompts.mjs`, `src/ui/combat_sfx.ts`, `src/game/sfx_manifest.generated.ts` and `tests/sfx_manifest.test.ts`. I built the merged tree: `sfx_prompts.mjs` auto-merges cleanly and silently emits a `mob_reptile_hurt` key whose prompt contains the literal word "undefined" (the reptile `mob()` call from #1873 passes only 5 arguments to your 6-argument template), then `npm run sfx:manifest` hard-throws `missing sampled SFX file: public/audio/sfx/mob_reptile_hurt.mp3`, and `tsc` fails because `MOB_VOICE_CUES.reptile` has no `hurt` member while `MobVoiceAction` now requires one. The body says reptile "picks up hurt for free ... just needs its own audio dropped in", which understates it: the PR that lands second also needs the hurt prompt argument, the `MOB_VOICE_CUES.reptile.hurt` entry, a regenerated manifest and runtime pack, and updated count pins (13 families x 4 = 52 mob keys). Please agree the landing order with #1873 and carry that reconciliation in the same change. The count pin and the `satisfies` clause do catch this, which is good; please fix it forward rather than by relaxing the pin.

**5. Should fix: `rate: 1.25` is retained at `src/ui/hud.ts:8647`.** That pitch-up existed to make an attack bark read as pain. `mob_*_hurt` is already an authored 0.6 second pain vocalization, so 1.25x now pitches up a purpose-built asset. Worth A/B testing 1.0 against 1.25 in game and dropping the option if 1.0 sounds right; keep the 0.1 cooldown either way.

**6. Minor: a family added to the audio catalog but forgotten in `MOB_VOICE_CUES` (or the reverse) is silent today.** One assertion that the keys of `MOB_VOICE_CUES` equal `SFX_MOB_EXTENSION_FAMILIES` would close that, and would also give #1873's reptile scaffolding its first real pin.

One more note: `mobVoiceCue` only probes `hasVariants` for the subfamily key and returns the family fallback unchecked. No subfamily override files exist on disk today, so the "attack override exists, hurt override does not" regression does not currently bite, but a missing family clip resolves to a key with no entry and plays nothing, silently. Cheap hardening for `availableMobVoiceCue` in `hud.ts`.

Screenshots: correctly N/A, audio only. No i18n, sim, server, or persistence surface touched.

Happy to re-review once the cold-buffer question is settled and the crit swap is covered by a test.

jamiecypher:
Thanks for catching the cold-buffer race and the policy logic buried in `hud.ts`. Addressed in the latest push:

- **Cold buffer**: added `Sfx.isBuffered(key)` / `Sfx.preload(key)` in `src/game/sfx.ts`. `playAttackerSfx` now preloads the `hurt` cue as soon as `attack` plays, and the crit-reaction branch in `hud.ts` falls back to the `attack` cue if `hurt` isn't buffered yet by the time a crit lands.
- **Extracted the pure policy function**: `mobVoiceActionForDamage(event, target)` now lives in `src/ui/combat_sfx.ts` (crit + non-boss-mob gate), with its own unit test; `hud.ts` just calls it.
- Dropped the `rate: 1.25` pitch-up on the hurt cue per your note.
- Rebased this branch onto #1873's tip (the reptile family fix, including the `deepfen_spearjaw` retag it folded in) so what's pushed here is the real combined tree you tested against, not something that collides with #1873 on merge. Once #1873 lands upstream, I'll rebase this back onto the real `release/v0.26.0` to drop the stacked commits from the diff.

tsc, the full test suite, and biome (changed files) are all clean on the current push.

EnriqueGF:
Thanks for the quick turnaround on the last round :) The extraction of `mobVoiceActionForDamage`, the table-driven per-family hurt test, and the `MOB_VOICE_CUES` lockstep pin all landed exactly as asked, and findings 3 and 6 from my previous review are properly closed. Verdict is still changes requested: the cold-buffer fix does not actually cover the multi-take families, and the part of the code it added is the one part with no test.

Base branch gate: PASS. This still targets `release/v0.26.0`, which is correct.

What I re-verified locally on the current head (`353bc9916`), not taken from the description:
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/combat_sfx.test.ts tests/sfx_manifest.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts tests/hud_perf_budget.test.ts`: all green (the one initial failure was a missing generated `i18n.status.json` in my tree, cleared by `npm run i18n:gen`, not your change).
- `npm run sfx:manifest`: regenerates the manifest and runtime pack byte for byte identical to what is committed.
- Biome on the changed files: clean (3 pre-existing warnings on an unrelated `hud.ts` method).
- The `beast` to `reptile` retag of `deepfen_spearjaw` is behaviorally inert in the sim, which I checked rather than assumed: social pull keys off `templateId`, not family (`src/sim/sim.ts:4971`), `beast` was never in `FLEEING_FAMILIES`, `SOCIAL_PULL_RADIUS` only overrides `mudfin`, polymorph only exempts `undead`, and taming is already blocked for instance mobs by the `DUNGEON_X_THRESHOLD` check before family is ever consulted. No gameplay change, good.

Findings, most severe first.

**1. Blocking, merge order: this PR currently carries #1873's unmerged commits.** The base branch is right, but the merge base of this head against `release/v0.26.0` is still the v0.25.0 release commit, so the diff includes `c8531db3d` and `f7a1ee047` (the reptile family and the `deepfen_spearjaw` retag). #1873 is still open. Merging this as it stands would land the reptile family into the release branch on this PR's approval rather than its own. Stacking to test the combined tree was the right call, but it cannot merge in this shape: land #1873 first and rebase this onto the real `release/v0.26.0` so its diff is only the hurt action, or retarget this PR's base to #1873's head branch so GitHub shows the true delta.

**2. Major: the cold-buffer fix does not cover the two-take families, so the silence it was meant to prevent still happens.** `isBuffered` and `preload` both hard-code variant index 0 (`src/game/sfx.ts:349`, `src/game/sfx.ts:358`), but `playAt` selects the variant round-robin through `nextVariantIndex` and advances a per-key cursor on each successful play (`src/game/sfx.ts:375`, `src/game/sfx.ts:396`). `mob_spider_hurt`, `mob_humanoid_hurt` and `mob_dragonkin_hurt` each have two variants in the generated manifest. The concrete sequence: the mob's attack bark warms hurt variant 0; the first crit sees `isBuffered` true, plays variant 0, and advances the cursor to 1; the second crit still sees `isBuffered` true (variant 0 is warm), so `hud.ts:8654` skips the attack fallback, and `playAt` then reaches for the cold variant 1 and enters exactly the 0.12s fetch race the fallback exists to avoid. Net effect: one likely-silent crit reaction per two-take family per session, on the exact path this commit set out to fix. Please make `isBuffered` test the variant `playAt` would actually pick (or all non-failed variants) and have `preload` warm every variant, the way `preloadStartup` already does at `src/game/sfx.ts:235`.

**3. Major: the substance of the fix commit has no test.** The fallback selection at `src/ui/hud.ts:8652-8657` is the whole point of `353bc9916`, and replacing it with a plain `hurtVoice` leaves every test in the branch green. `sfx.isBuffered` and `sfx.preload` likewise ship as new public API on the audio seam with zero direct coverage. `mobVoiceActionForDamage` extracted the easy half (the crit gate); the interesting half, prefer hurt, else fall back to attack when hurt is not resident, stayed as an inline ternary on the coordinator, which is the same module-first point as last round. Please extract it as a pure resolver in `combat_sfx.ts` (for example `mobVoiceCueWithFallback(templateId, action, hasCue, isBuffered)`, both predicates inject as cleanly as `mobVoiceCue`'s existing `hasCue`), pin the warm arm, the cold arm and the no-cue arm in `tests/combat_sfx.test.ts`, and add an `isBuffered`/`preload` case to the existing WebAudio-stub suite in `tests/sfx.test.ts`, including a two-variant key so finding 2 stays fixed.

**4. Should fix: dropping `rate: 1.25` also un-pitched the fallback.** Retiring the pitch-up for the authored hurt clip is right, but `src/ui/hud.ts:8657` now passes only `{ cooldown: 0.1 }` on both arms, so a cold-cache crit plays the plain attack bark at its natural pitch and is acoustically identical to an ordinary swing, which is exactly the "did my crit land" read the 1.25 existed to give. Keep `rate: 1.25` on the fallback branch only, and pin the opts in the extracted resolver's test.

**5. Minor: the warm is ungated and unthrottled.** At `src/ui/hud.ts:8746-8749` the `preload` call is not behind `shouldPlayMobVoiceSfxForEntity(src)`, unlike the `combat(voice, ...)` play two lines above it, so the muted Nythraxis mobs also warm a clip that can never play for them (low impact, the family key is shared with other undead, but the two should share one policy, as `mobVoiceActionForDamage`'s own doc comment says). It also has no already-warm short-circuit, so every mob attack event re-runs `availableMobVoiceCue` (a template-literal allocation plus a variant scan) and `sfx.preload`, which still allocates a `Promise.resolve` on the cached path. `if (hurtVoice && !sfx.isBuffered(hurtVoice)) sfx.preload(hurtVoice)` behind the existing guard closes both.

**6. Nit: drop the `@ts-expect-error` import.** `tests/combat_sfx.test.ts:2` pulls `SFX_MOB_EXTENSION_FAMILIES` from `scripts/sfx/manifest.mjs` behind a suppression, but the same constant is already exported, typed, from `src/game/sfx_manifest.generated.ts:24`. Importing it from there keeps the test just as decisive without the suppression.

Screenshots: correctly N/A, audio only. No i18n, server, or persistence surface touched, and the sim purity and parity guards are green.

Happy to re-review once the variant race is closed and the fallback resolver is extracted and covered.

jamiecypher:
Thanks, all three addressed:

1. **Multi-take variant gap**: `isBuffered`/`preload` in `src/game/sfx.ts` only checked/warmed variant 0. `playAt` round-robins variants, so a two-take family (spider, humanoid, dragonkin hurt all have 2 takes) could warm variant 0, play it on the first crit, then hit the still-cold variant 1 on the second crit, exactly the silence the fallback exists to prevent. Both now check/warm every variant of the key.
2. **Extracted the fallback resolver**: `mobVoiceCueWithFallback(templateId, action, hasCue, isBuffered)` now lives in `combat_sfx.ts`, pinned with a warm case, a cold case, and a no-cue case in `tests/combat_sfx.test.ts`. `hud.ts`'s crit-reaction branch is now a thin call site. Added `isBuffered`/`preload` coverage in `tests/sfx.test.ts` too, including a real two-variant key (`mob_spider_hurt`) so the variant fix stays covered.
3. **Merge order**: this branch is rebased onto #1873's latest tip (`1be382550`, the tooltip/crest fix), so it still reflects the true combined tree. Will rebase onto the real `release/v0.26.0` once #1873 actually merges.

On `rate: 1.25`: talked it over and decided to leave it dropped rather than restore it on the fallback arm. The crit already fires its own impact sound underneath (`combat_crit` / the school impact cue), so the voice bark doesn't need to carry the "this was a crit" signal via pitch on top of that.

Also fixed the two minor items: the hurt-cue warm in `playAttackerSfx` is now gated behind the same Nythraxis mute check as the play beside it and short-circuits once already buffered (no more re-scan on every attack), and the test import drops the `@ts-expect-error` suppression in favor of the now-available typed `SFX_MOB_EXTENSION_FAMILIES` export from `sfx_manifest.generated.ts`.

tsc, the full relevant suite, `npm run build`, and biome on the changed files are all clean on `048d73e7f`.

jgyy (commented):
Nice fix, and the review iteration on this one is worth calling out: the first pass wired `hurt` straight in, then the follow-ups actually chased the hard part, the cold-buffer race on a crit-only cue. `mobVoiceCueWithFallback` falling back to the always-warm `attack` cue rather than dropping the sound entirely is the right call, and catching that `playAt` round-robins variants (so a two-take family could warm variant 0 and still lose the race on variant 1) is a real bug that a less careful pass would have shipped. `mobVoiceActionForDamage` and `mobVoiceCueWithFallback` as pure, host-agnostic functions in `combat_sfx.ts` are exactly the right shape, easy to test in isolation from `hud.ts`.

Ran the full suite locally (`sfx_manifest`, `combat_sfx`, `sfx`, `sfx_conform`, `architecture`) plus `tsc --noEmit`: all green (133 tests), matches the PR body.

Nothing blocking. One small note:

- `nit`: dropping the `rate: 1.25` pitch-up on the fallback arm (per the commit message, "the crit already has its own impact sound underneath") is a reasonable call, but it's a small audio-feel decision buried in a commit message rather than the PR description. Worth a one-line mention in the PR body itself since it's a behavior change a reviewer skimming just the description would miss.
- Same as #1873: the only merge conflicts against `release/v0.26.0` are the generated i18n trio, mechanical regen churn, not a real conflict.

This stacks on #1873 (reptile family) and correctly excludes `mob_reptile_hurt` from that PR's scope while including it here once the family exists. Good sequencing.

jamiecypher:
Rebased onto the resolved feature/sfx-reptile-family (this PR is stacked on it) plus current release/v0.26.0, clearing the same stale-base i18n conflict (pending.ts, i18n.resolved.sha256, i18n.status.summary.json regenerated). Also picks up #1883's SFX Studio publish-discovery fix. npm run gate is green: all 11 steps pass.

jamiecypher:
Rebased onto the updated feature/sfx-reptile-family, same #1574 batch pickup. Auto-merged clean apart from generated artifacts. npm run gate green in isolation (11/11 steps). Note: this repo's integration test suite uses fixed local ports, so a gate run here should not be run concurrently with another gate run elsewhere in the tree, an earlier attempt to parallelize gates across worktrees produced spurious failures purely from resource contention, not a real issue.

jamiecypher:
Rebased now that #1873 landed (squash-merged, so its stacked copy of reptile in this branch's history collided rather than fast-forwarding). Resolved by deduping: kept this branch's reptile entries (which have hurt), dropped #1873's now-redundant copy in src/ui/combat_sfx.ts and scripts/sfx/sfx_prompts.mjs (the latter was a silent duplicate mob('reptile', ...) call that auto-merge didn't flag as a conflict, caught by re-running the tests). Regenerated to 154 clips, matching the pinned catalog count. npm run gate green in isolation (11/11 steps), confirmed twice (one flaky AAC bitrate probe unrelated to this diff, passed on rerun both times).

jamiecypher:
Rebased now that #1887 (idle) also landed. #1887 was stacked on this branch, so its squash-merge collided instead of fast-forwarding, same class of issue as the #1873 dedupe. Resolved by taking upstream's idle-aware mob() helper wholesale (a strict superset of this branch's plain version) and this branch's pinned-count expectations (155 keys / 53 mob keys, matching the regenerated catalog). npm run gate green in isolation (11/11 steps); the AAC bitrate probe flaked again on this run (same known flake, confirmed clean on immediate rerun).

PR #1876, opened 2026-07-13
fix(scripts): resolve asset pipeline ROOT via fileURLToPath so it runs on Windows

Wmedrado:
## Summary

`scripts/assets/build_assets.mjs` and `scripts/assets/build_foliage.mjs` computed their repo ROOT with `path.resolve(path.dirname(new URL(import.meta.url).pathname), '..', '..')`. On Windows, `URL.pathname` keeps a leading slash before the drive letter (`/D:/...`), which `path.resolve` mangles into `D:\D:\...`, so neither asset pipeline entry could run on Windows at all.

Both now use `path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..')` (via `node:url`), which is correct on every OS. A sweep of the tree found no other occurrence of the broken pattern.

A new guard test, `tests/scripts_windows_paths.test.ts`, walks every script under `scripts/` and fails on any `new URL(...import.meta.url...).pathname` usage, and pins that both asset pipeline entries use the `fileURLToPath` form. It fails on the pre-fix scripts and passes post-fix.

Both scripts also pick up the full biome format pass, required because `biome ci --changed` gates changed files on format diffs and these files carried pre-existing format debt. Every reformatted hunk is purely syntactic; no logic changed.

## Type of change

- [x] Bug fix
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/scripts_windows_paths.test.ts`: fails with the fix reverted (both assertions), passes with it applied.
  - `node scripts/assets/build_assets.mjs <probe spec>` on Windows 11: ROOT now resolves to the real repo root (previously the mangled `D:\D:\...` path made the script unrunnable).
  - `node --check scripts/assets/build_foliage.mjs` passes.
  - Gate steps run individually on Windows: i18n gen + freshness, malware scan, `biome ci --changed`, sfx check, `check:types` (tsc + svelte-check, 0 errors), `build:env`, `build:server`, and the full `npm run build` (all five entries) are all green.
- Manual steps:
  - The full local `npm run gate` vitest step is red on this Windows machine, but identically red on the pristine `origin/release/v0.26.0` base: an A/B run of the whole suite shows the same 166 failing entries on both (golden-master fixture CRLF mismatches under `core.autocrlf=true`), with a delta of exactly one unrelated flaky test (`tests/chat_item_link_draft.test.ts`) that passes in isolation. The two tests this PR adds pass in both runs. Same story for the one `test:browser` failure (armory mobile layout, identical 393.625 vs 391 assertion on the untouched base). CI on Linux is unaffected by either.

### Quality

- [x] **The full gate passes.** Green except the pre-existing Windows-local vitest/browser failures shown above to be identical on the untouched release base; every other gate step is green, and the changed behavior has decisive new tests.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A, offline build tooling only.
- ~Accessible.~ N/A, no UI.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (script output is dev-channel English).

### Hygiene

### DIFF
--- a/scripts/assets/build_assets.mjs
+++ b/scripts/assets/build_assets.mjs
@@ -27,13 +27,23 @@
 // - "copy": byte-for-byte copy (HDRIs, plain textures).
 import fs from 'node:fs';
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { NodeIO } from '@gltf-transform/core';
 import { ALL_EXTENSIONS } from '@gltf-transform/extensions';
-import { dedup, mergeDocuments, meshopt, prune, resample, textureCompress } from '@gltf-transform/functions';
+import {
+  dedup,
+  mergeDocuments,
+  meshopt,
+  prune,
+  resample,
+  textureCompress,
+} from '@gltf-transform/functions';
 import { MeshoptDecoder, MeshoptEncoder } from 'meshoptimizer';
 import sharp from 'sharp';
 
-const ROOT = path.resolve(path.dirname(new URL(import.meta.url).pathname), '..', '..');
+// URL.pathname keeps a leading slash before a Windows drive letter, which
+// path.resolve mangles into "D:\D:\..."; fileURLToPath is correct on every OS.
+const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..');
 const PUBLIC_DIR = path.join(ROOT, 'public');
 
 function stripClipName(name) {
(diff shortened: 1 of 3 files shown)


Rubsey (commented):
Verified end to end at 6f508c317, and it holds up. The root cause is real (URL.pathname keeps the leading slash before a drive letter, and path.win32.resolve turns /D:/... into a drive-relative \D:\... that Node then qualifies into D:\D:\...), fileURLToPath is the right fix, and the guard test actually bites: with the two scripts reverted to the base versions, both tests go red and the walker names both offenders.

Checks I ran beyond reading the diff:

- The "purely syntactic" reformat claim is exactly true: rebuilding each file as base plus only the import/comment/ROOT lines and running biome format reproduces the PR head byte for byte.
- A tree sweep at the head finds no other occurrence of the broken form, including line-split variants a line-based grep would miss (the test regex does catch those).
- tsc is clean at the head; biome ci on the three changed files is clean (the one useOptionalChain warning in build_assets.mjs is on a line byte-identical to base, and warnings do not gate).
- No remaining Windows blockers in either script: every spec-supplied forward-slash path goes through path.join/path.resolve, extension filtering is case-insensitive, and there is no shell usage.
- Merges clean into release/v0.26.0.

Two optional nits on the guard test, neither active in the current tree:

- nit (tests/scripts_windows_paths.test.ts:19): statSync throws on a dangling symlink under scripts/, so stray debris would crash the guard instead of being skipped. lstatSync plus a symlink skip would make it indifferent to that.
- nit (tests/scripts_windows_paths.test.ts:31): the banned regex only catches .pathname chained directly onto new URL(...); the two-step form (const u = new URL(import.meta.url); u.pathname) slips past. Fine as-is for the common form.

Fork PRs need a manual CI approval, which I have done: Biome, browser regressions, and screenshots already pass, and the full PR gate is running.

PR #1877, opened 2026-07-13
fix(render): float32-bake uvs in the static idle-pose merge

Wmedrado:
## Summary

Entering the offline game floods the browser console with THREE.BufferGeometryUtils errors: ".mergeGeometries() failed while trying to merge the uv attribute" and ".mergeAttributes() failed. BufferAttribute.array must be of consistent array types across matching attributes".

Root cause: `bakeStaticPose` (the per-key far-LOD / shadow-proxy idle-pose bake in `src/render/characters/assets.ts`) merges every visible mesh of a posed clone while carrying uv attributes verbatim (`uv.clone()`). The shipped GLBs are meshopt-quantized, and the rig merge (`rig_merge.ts`) rebakes only the mergeable skinned parts to float32, so a posed clone mixes normalized Uint16 and Float32 uv arrays across its meshes. `mergeGeometries` refuses that mix and returns null, so every affected rig silently lost its far LOD and shadow proxy (`idleGeo: null`): the full animated rig kept rendering at distance (a per-rig perf loss) and the proxy-shadow regime cast no shadow.

The fix follows the `foliage.ts` `bakeGeometry` recipe: `bakeStaticPose` and `meshChainVisible` move verbatim into the new module `src/render/characters/static_pose.ts` (module-first extraction so the merge behavior is unit-testable without the manifest/loader machinery), and every uv is baked to a plain float32 attribute before the merge (`toFloatUv`, reading through `getComponent`, which denormalizes; already-plain float32 attributes keep the old clone fast path).

Consolidating the four sibling float-bake helpers (`foliage.ts`, `dungeon.ts`, `vale_cup_stadium.ts`, `static_pose.ts`) is deliberately left to a follow-up to keep this diff scoped.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/static_pose.test.ts` (new suite, written test-first: red on the quantized + float uv mix before the fix, green after), plus `tests/visual_manifest.test.ts`, `tests/render_asset_preload.test.ts`, `tests/rig_merge.test.ts`, `tests/rig_merge_assets.test.ts`, `tests/architecture.test.ts` (all green), `npx tsc --noEmit`, and biome ci scoped to the changed files (0 errors).
- Manual steps: with `npm run dev`, entered the offline world with a `console.error` hook capturing stacks. Before: 16 merge errors during the character screen alone, each stack pinned to `bakeStaticPose` via `prepareVisual`. After: 0 merge errors across the offline panel, all 9 class previews, and world entry. The `scripts/pr_screenshots.mjs` tooling reproduced the same delta headlessly: its before capture logged the merge-error flood in the page console, the after capture logged none.
- Local `npm run gate` on this Windows machine fails only in fixture suites that fail identically on the pristine base (CRLF line-ending checkout artifacts in the golden-master and SFX suites), none related to this change; CI is the arbiter for those.

## Screenshots / recordings

Before/after desktop HUD pair, committed under `docs/screenshots/static-pose-uv-merge/`. The frames are intentionally near-identical: the fix restores an equivalence-preserving LOD path, and the capture runs on the low tier (software GL) where shadow proxies are disabled by design. The observable regression fixed here is the console error flood plus the silently missing far-LOD swap and proxy shadows on affected rigs.

| Before | After |
| --- | --- |
| before | after |

---

### Quality

- [x] **The full gate passes.** Targeted suites, typecheck, and scoped biome are green locally; the only local `npm run gate` failures are machine-specific CRLF fixture artifacts that fail identically on the pristine base (details above).

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: render-internal geometry fix, no UI, layout, or input surface changed.

### Localization (i18n)

### Hygiene


jgyy:
Closing — this looks already implemented. `assets.ts` on `upstream/main` already dequantizes UVs to `Float32Array` during the idle-pose merge. Thanks for the contribution.

PR #1880, opened 2026-07-14
fix(render): dequantize uvs in the static pose bake so mixed parts merge

Wmedrado:
## Summary

Fixes the console-error flood reported from low-tier screenshot runs: `THREE.BufferGeometryUtils: .mergeGeometries() failed while trying to merge the uv attribute / BufferAttribute.array must be of consistent array types across matching attributes`, logged 38 times per client boot.

The report pointed at the props band merge, but an instrumented run (console.error wrapped to capture stacks) localized every one of the 38 errors to `bakeStaticPose` in `src/render/characters/assets.ts`, and showed it fires on BOTH `?gfx=low` and `?gfx=high` (2 errors per prepared visual key, 19 keys). The props merge is clean: `propAsset()` already converts every attribute to float32.

Root cause: `bakeStaticPose` bakes positions to float32 but cloned each part's uv attribute raw. `rig_merge.ts` dequantizes only the skinned parts it merges, so an assembled model mixes float32 uvs (merged body) with meshopt/KHR_mesh_quantization normalized-integer uvs (unmerged parts and non-skinned attachments). `mergeGeometries` refuses mixed array types and returns null, so `PreparedVisual.idleGeo` was silently lost: affected characters never got their far-LOD swap or shadow proxy (`visual.ts` falls back to the full animated rig at distance, so nothing vanished, but the LOD saving and distant shadows were dead, plus the console spam).

Fix, done per the extract-and-test workflow:

1. `refactor(render)`: move `bakeStaticPose` + `meshChainVisible` verbatim out of `assets.ts` into the new sibling module `src/render/characters/static_pose_bake.ts` so the bake is unit-testable without the loader/preload machinery.
2. `fix(render)`: reproduce with a failing Vitest (mixed float32 + quantized uv parts made the bake return null), then the smallest fix: a `toFloatUv` helper bakes every uv to plain float32 before the merge (the attribute getters denormalize quantized sources, so values are unchanged).

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands:
  - `npx vitest run tests/static_pose_bake.test.ts` (new suite: mixed-type merge, denormalized values, group alignment, drop-uv-when-absent, visibility chain; the mixed-type cases fail on the pre-fix code)
  - `npx vitest run tests/rig_merge.test.ts tests/architecture.test.ts tests/render_asset_preload.test.ts tests/visual_manifest.test.ts` (56 tests green on this branch)
  - `npx tsc --noEmit` clean; changed-files Biome clean
- Manual steps: instrumented puppeteer run of the offline client (swiftshader, same flags as `scripts/pr_screenshots.mjs`), capturing `console.error` stacks:
  - before (clean `release/v0.26.0`): 38 merge errors, all stacked at `bakeStaticPose`, on both `?gfx=low` and `?gfx=high`
  - after (this branch): 0 merge errors on both tiers, low-tier town renders fully

## Screenshots / recordings

Low tier (swiftshader) town view; the delta is console cleanliness plus restored far-LOD/shadow proxies, so the frames are intentionally near-identical.

| Before (release/v0.26.0) | After (this PR) |
| --- | --- |
| before | after |

---

### Quality

- [x] **Decisive tests added.** The new suite fails on the pre-fix code for the bug's exact reason (merge returns null on mixed uv array types) and pins float32 output, denormalized values, and group alignment. Targeted suites, `tsc`, and changed-files Biome are green locally; the full local `npm test` carries ~30 pre-existing Windows-environment failures (sfx/admin fixture classes) that reproduce identically on the clean base with this change stashed, so CI is the authoritative full-gate run.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: no UI/layout change; verified desktop low and high tiers in-browser.

### Localization (i18n)

### Hygiene


jgyy:
Closing: the core fix here (dequantize a mesh's uv BufferAttribute to a fresh Float32Array before mergeGeometries, so mixed float32/quantized-integer uv parts merge) already landed via #2092 (fix(render): dequantize baked-pose uv before merging character geometries, merged 2026-07-18), confirmed an ancestor of the current release/v0.33.0 tip. Current code: src/render/characters/dequantize_attribute.ts, applied in bakeStaticPose, with regression coverage in tests/character_uv_dequantize.test.ts. This PR's separate refactor (extracting bakeStaticPose into its own module) was not adopted upstream, but the bug this issue is about is fixed.

PR #1881, opened 2026-07-14
feat(ui): reorganize HUD chrome into a cleaner classic layout

Wmedrado:
## Summary

Declutters and reorganizes the desktop HUD chrome into a cleaner classic layout. The right screen edge was a full-height rail of 17 micro buttons that polluted the 3D view; the minimap zoom pill overlapped the compass strip and hid the clock; the WOC store chest collided with the compass; and the player name plate stretched across the whole action-bar width.

What changed (CSS in `src/styles/hud.css` plus one DOM relocation of `#minimap-zoom` into `#minimap-disc` in both HTML entries):

- **Micro menu:** the vertical right-edge rail becomes a compact two-row horizontal dock in the bottom-right corner (classic micro-menu placement). Hover or keyboard focus shows a floating caption above each button instead of the old grow-leftward flyout. GitHub/Donate sit right above the dock.
- **Minimap cluster:** zone name on a slim plate, clock and coordinates as one pill row under the disc, borderless compass strip, and the zoom control rides the disc rim, shown only on hover or keyboard focus (`:focus-within`), so it no longer overlaps the compass and clock. The redundant heading letter under the strip is dropped (mobile already hid it; the strip's center caret still shows the heading).
- **Objective trackers:** shared uppercase gold section headers for quest, deed, and delve trackers; objectives indent under the title text.
- **Bottom cluster:** the player plate caps at 300px instead of stretching to 520px; action slots get a thinner bezel ring, receding empty slots, and keybind chips; the bars sit on a soft rounded plate.
- **Chat:** floating pill tabs and a softer thin-ring panel, 12px body text.
- **A11y follow-up (from the QA reviewers):** the bezel now rides `outline` in `@layer components`, which would beat the base-layer `:focus-visible` ring in the cascade; the accent ring is re-declared at the component layer for `.action-btn`/`.micro-btn`, the caption translate is gated under `prefers-reduced-motion`, and a new always-on guard (`tests/hud_focus_ring_layers.test.ts`) fails if any focus-net control gets an outline decoration in hud.css without a same-file ring re-declaration (proven decisive: red with the re-declaration removed). The opt-in browser focus suite also now Tabs to both controls and asserts the computed ring.

Mobile behavior is untouched: the rail, zoom pill, and heading letter are already hidden or re-laid-out by `hud.mobile.css` (verified by screenshot below).

## Type of change

- [x] Feature: new functionality (visual/UX reorganization)
- [ ] Bug fix
- [ ] Documentation
- [x] Tests

## How was this tested?

- Commands:
  - `npx vitest run` over the CSS/chrome guards: `css_corpus`, `css_value_validity`, `styles_extraction`, `per_entry_css_wiring`, `mobile_window_coverage`, `mobile_window_transform`, `mobile_window_layout`, `fct_mobile_css`, `client_shell`, `mobile_chat_centered`, `touch_router`, `delve_quest_ui_layout`, `backdrop_filter_survival`, `focus_visible_guard`, `ui_effects_wiring`, `theme`, `localization_fixes`, `architecture`, `focus_order`, `focus_manager`, `hud_perf_budget`, `action_bar_painter`, `minimap_painter`, `dialog_root`, `painter_host`, and the new `hud_focus_ring_layers` (all green)
  - `npx vitest run --config vitest.browser.config.ts tests/browser/focus_indicator.browser.test.ts` (10 passed, including the two new in-game controls)
  - `npx tsc --noEmit` (clean), `npx @biomejs/biome check --write` on changed files
  - The full local `npm run gate` is not meaningful on this Windows machine (pre-existing CRLF golden-master fixture failures on the pristine base too); CI is the arbiter.
- Manual steps: entered the offline game at 1600x900 (desktop) and 844x390 landscape with `mobile-touch` (mobile), verified the dock, hover captions, minimap hover zoom, trackers, action bar, and chat visually (screenshots below); confirmed the micro-menu captions and zoom also appear via keyboard focus.

## Screenshots / recordings

Committed under `docs/screenshots/hud-modern-layout/`.

Desktop, full HUD:

| Before | After |
| --- | --- |

Micro menu (right-edge rail to bottom-right dock; hover caption):

| Before | After | After (hover) |
| --- | --- | --- |
| before rail | after dock | after dock hover |

Minimap cluster (zoom pill no longer overlaps compass/clock; zoom on hover):

| Before | After | After (hover) |
| --- | --- | --- |
| before minimap | after minimap | after minimap hover |

Bottom cluster and chat:

| Before (bottom) | After (bottom) |
| --- | --- |
| before bottom | after bottom |

| Before (chat) | After (chat) |
| --- | --- |
| before chat | after chat |

Mobile (landscape, unchanged behavior):

| Before | After |
| --- | --- |

---

### Quality

- [x] Targeted guard suites, typecheck, and biome are green (see above); the full local gate is environmentally red on this machine (CRLF fixtures) on the pristine base too, so CI is the arbiter. A new decisive guard test covers the one regression class this change introduces.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop 1600x900 and a phone-sized landscape viewport with `mobile-touch`; the touched elements are hidden or independently laid out on mobile, verified by screenshot.
- [x] **Accessible.** Keyboard focus shows the 2px accent ring on the restyled controls (guarded by the new test), captions and the minimap zoom also appear on keyboard focus (`:focus-visible` / `:focus-within`), and the caption translate is gated under `prefers-reduced-motion`.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the relocated zoom block keeps its existing `data-i18n-*` keys; index.html's zoom wrapper aria was aligned with play.html's existing key).

### Hygiene


EnriqueGF (changes requested):
Reviewed the full diff (hud.css, the `#minimap-zoom` relocation in both HTML entries, and both new tests) against root `CLAUDE.md` and `src/styles/CLAUDE.md`.

**What I ran here** (worktree on `release/v0.26.0`): `npx tsc --noEmit` (clean), and `npx vitest run` over `hud_focus_ring_layers`, `focus_visible_guard`, `css_corpus`, `css_value_validity`, `styles_extraction`, `per_entry_css_wiring`, `architecture`, `localization_fixes`, `mobile_window_coverage`/`_layout`/`_transform`, `focus_order`, `client_shell`, `theme`, `action_bar_painter`, `minimap_painter`, `delve_quest_ui_layout`, `hud_perf_budget`: all green (`localization_fixes` needed `npm run i18n:gen` first, environmental, not this PR).

**What checks out:** the layer/`@import` order and the ten-dash section banners are intact; the relocated zoom block keeps its `data-i18n-*` keys and `hudChrome.widgets.minimapZoom` already exists in the catalog, so index.html's new aria is a real key; `body.mobile-touch` already hides `#minimap-zoom`, `#compass-heading`, `#side-buttons` and `#daily-rewards-button`, so the mobile arm really is untouched by the reflow; `#minimap-disc` is `position: relative`, so the absolutely positioned pill anchors where you say it does; the `hud.ts` lookups are id-based, so moving the node is safe; no dashes, emoji, secrets, or sim/server surface touched.

### Finding (medium, a11y regression, correctness): the re-declared focus ring is clobbered again by the state rules below it

`src/styles/hud.css:1435` re-declares `.action-btn:focus-visible { outline: 2px solid var(--color-border-focus) }`, but four same-specificity rules that also touch `outline-color` are declared *after* it in the same file and layer:

- `.action-btn.empty` (line 1454) `outline-color: #3a321c`
- `.action-btn.used` (1444), `.action-btn.queued` (1508), `.action-btn.drop-target` (1517)
- `.micro-btn.has-points` (1606) `outline-color: var(--gold)`

`.action-btn.empty` and `.action-btn:focus-visible` are both (0,2,0), same layer, so source order decides and `.empty` wins the `outline-color` longhand. Empty slots are ordinary focusable `<button class="action-btn empty">` elements (`src/ui/hud.ts:5626`, no `disabled`, no `tabindex=-1`), so **tabbing onto an empty action slot now paints a 2px `#3a321c` ring on the `#0b0b11aa` plate: roughly 1.4:1, i.e. no visible focus indicator.** Before this PR the bezel was `border`, `outline` was free, and the base-layer ring showed correctly on empty slots, so this is a regression of exactly the class the PR is trying to close. (`.queued`/`.drop-target`/`.has-points` only re-tint the ring to a still-visible colour, so those are cosmetic; `.used` re-tints to the focus colour anyway.)

Simplest fix: move the `:focus-visible` block *below* all the state rules for both controls (or make it `.action-btn:focus-visible:focus-visible` / add `outline-color` at higher specificity).

The new guard does not catch this: `tests/hud_focus_ring_layers.test.ts` only asserts that *a* `:focus-visible` re-declaration exists somewhere in hud.css, not that it wins over the decorative rules. Worth extending it to also require the re-declaration's source offset to be after every non-`:focus-visible` outline rule for that class, so the fix is pinned. The browser test passes today because it only asserts `outline-width >= 2px` (the width survives, only the colour is defeated) and it likely lands on a non-empty slot.

### Minor (non-blocking)

- `#side-buttons { width: 348px }` (line ~1528) hard-codes the dock width to today's button count; adding or removing one micro button silently changes the wrap. A `max-width` on the container plus letting flex wrap naturally, or a comment tying the value to the count, would age better.

Everything else in the reorganization looks good to me. Happy to re-approve once the focus-ring ordering is fixed and pinned.

PR #1883, opened 2026-07-14
fix(sfx): scope SFX Studio publish inventory scan to the full catalog

jgyy:
## Summary
- `publicPath()` in `scripts/sfx_studio/audio_io.mjs` called `discoverSfxTracks` with a single-entry catalog array containing only the key being published, so every other fixed-catalog `mob_<family>_<action>.mp3` file already committed under `public/audio/sfx/` was misclassified as an unrecognized mob subfamily file and failed the publish/export inventory check for any unrelated key.
- Pass the full `CATALOG` instead so discovery consumes every fixed-catalog filename before falling through to the mob-subfamily pass.
- This was breaking `tests/sfx_studio.test.ts` and `tests/sfx_studio_server_security.test.ts` (and therefore the "PR gate (English-only legal)" CI job, since it runs the full PR-tier test suite) for every PR against `release/v0.26.0`, independent of any specific PR's own changes.

## Test plan
- [x] `npx vitest run tests/sfx_studio.test.ts tests/sfx_studio_server_security.test.ts` (60/60 passing, was 46/60 before the fix)
- [x] `npx tsc --noEmit`
- [x] `npx @biomejs/biome check scripts/sfx_studio/audio_io.mjs`
- [x] Confirmed the failure reproduces on unmodified `release/v0.26.0` (pre-existing, unrelated to any in-flight PR)

### DIFF
--- a/scripts/sfx_studio/audio_io.mjs
+++ b/scripts/sfx_studio/audio_io.mjs
@@ -441,12 +441,12 @@ export async function analyzeLoopContinuity(path, channels) {
 }
 
 function publicPath(key, repoRoot = REPO_ROOT) {
-  const source = CATALOG.get(assertSfxKey(key));
+  assertSfxKey(key);
   const repo = realpathSync(repoRoot);
   const publicRoot = existingPlainDirectory(join(repo, 'public'), repo);
   const audioRoot = existingPlainDirectory(join(publicRoot, 'audio'), publicRoot);
   const sfxRoot = existingPlainDirectory(join(audioRoot, 'sfx'), audioRoot);
-  const discovered = discoverSfxTracks([source], sfxRoot);
+  const discovered = discoverSfxTracks([...CATALOG.values()], sfxRoot);
   if (discovered.errors.length) {
     throw new Error(`invalid published SFX inventory for ${key}: ${discovered.errors.join('; ')}`);
   }


FernandoX7 (commented):
Confirmed this is a real base breakage and not a PR-local one. I reproduced it on a pristine `release/v0.26.0` worktree with no PR changes applied: `tests/sfx_studio_server_security.test.ts` fails with `invalid published SFX inventory for foot_grass: unrecognized mob sfx file (not a catalog entry, not a valid numbered subfamily file): mob_beast_aggro.mp3` and 35 more like it. So this unblocks every PR against the base and is worth landing quickly.

The diagnosis is right and so is the fix. #1828 added the `consumedFilenames` set plus the new "unrecognized mob sfx file" error to the mob pass in `scripts/sfx/sfx_manifest_builder.mjs`. `publicPath()` has always scanned with a single-entry catalog, so nothing populated `consumedFilenames`, and all 36 committed `mob_<family>_<action>.mp3` files fell through to the mob pass and errored. Passing the full catalog puts `publicPath` in agreement with `scripts/sfx/manifest.mjs:217`. I checked the rest of the call sites: every other `discoverSfxTracks` / `buildManifest` caller that passes a subset is a test against its own fixture dir, so `publicPath` was the only production offender.

Key resolution does not widen either. `assertSfxKey` gates on `KEYS`, built from the same `SFX` array as `CATALOG` (`scripts/sfx_studio/audio_io.mjs:75-76`), so any key reaching `publicPath` was already in the catalog, and `discovered.entries[key]` resolves the same entry it did before.

Two nits, neither blocking:

- `scripts/sfx_studio/audio_io.mjs:449`: `SFX` would do instead of `[...CATALOG.values()]`. Same values, no per-call array allocation, and it reads as the exact argument `manifest.mjs:217` passes. It also keeps the duplicate-catalog-key error path live, which the `CATALOG` map silently dedupes away.
- Nothing pins the fix directly. The two suites do fail on base and pass here, so a pin exists, but only because the fixture copies the real `public/audio/sfx` and that dir happens to hold unconsumed 3-part mob files today. Rename those to the numbered subfamily form and the pin quietly evaporates. A short case in `tests/sfx_manifest.test.ts` (subset catalog, an unrelated catalog file in the dir, expect the error) would hold it directly.

On the obvious objection: `collectSfxCatalog` calls `publishedPath` once per key, so the scan is now O(catalog x files) per call. I measured it against the real 138-file dir, and 137 discovery calls go from about 20ms to about 106ms. That is nothing next to the per-clip ffprobe in the same loop, so it does not need restructuring.

I could not run `tests/sfx_studio.test.ts` locally (`ffmpeg-static` is not installed in my node_modules), but `tests/sfx_studio_server_security.test.ts` goes from 7 failing to 16/16 with the change applied.

PR #1886, opened 2026-07-14
fix(sfx): scope SFX Studio publish discovery to the full catalog

jamiecypher:
## Summary

`publicPath()` in `scripts/sfx_studio/audio_io.mjs` calls `discoverSfxTracks([source], sfxRoot)` with a single-entry catalog array instead of the full `SFX` list, unlike the real build path (`manifest.mjs`), which passes the full catalog.

#1854 turned an `mob_*.mp3` file that matches neither a catalog entry nor the numbered subfamily pattern into a hard discovery error instead of a silent skip. That is correct for the real build path, which always sees the full catalog. But this narrower call site now fails on every OTHER mob family's files sitting in the same `public/audio/sfx` directory, not just the one key actually being published, since a one-element catalog can't recognize any of them.

This currently breaks 7 tests in `tests/sfx_studio_server_security.test.ts` and `tests/sfx_studio.test.ts` on a clean `release/v0.26.0` checkout, unrelated to any in-flight feature PR.

Fix: pass the full `SFX` catalog (already imported in this file) instead of `[source]`, matching how `manifest.mjs` calls the same function. Also dropped the now-unused `CATALOG.get(...)` lookup in favor of the `assertSfxKey` validation it already performed.

## Type of change

- [x] Bug fix

## How was this tested?

- Confirmed the failure reproduces on a clean `release/v0.26.0` checkout (no feature changes).
- `npx vitest run tests/sfx_studio_server_security.test.ts tests/sfx_studio.test.ts tests/sfx_manifest.test.ts` — 83 passing (was 7 failing).
- `npx tsc --noEmit` clean.
- `npm run gate` — PASS, all 11 steps green.

## Screenshots / recordings

N/A, build/test tooling only.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [x] Builds succeed.

Cross-platform
- ~Tested on desktop and mobile.~ N/A, build tooling only.

Localization (i18n)

Hygiene

### DIFF
--- a/scripts/sfx_studio/audio_io.mjs
+++ b/scripts/sfx_studio/audio_io.mjs
@@ -441,12 +441,12 @@ export async function analyzeLoopContinuity(path, channels) {
 }
 
 function publicPath(key, repoRoot = REPO_ROOT) {
-  const source = CATALOG.get(assertSfxKey(key));
+  assertSfxKey(key);
   const repo = realpathSync(repoRoot);
   const publicRoot = existingPlainDirectory(join(repo, 'public'), repo);
   const audioRoot = existingPlainDirectory(join(publicRoot, 'audio'), publicRoot);
   const sfxRoot = existingPlainDirectory(join(audioRoot, 'sfx'), audioRoot);
-  const discovered = discoverSfxTracks([source], sfxRoot);
+  const discovered = discoverSfxTracks(SFX, sfxRoot);
   if (discovered.errors.length) {
     throw new Error(`invalid published SFX inventory for ${key}: ${discovered.errors.join('; ')}`);
   }


jamiecypher:
Closing, #1883 landed the identical fix (same root cause, same call site) a few minutes before this went up. release/v0.26.0 is green on this now, confirmed with tests/sfx_studio_server_security.test.ts and tests/sfx_studio.test.ts.

jgyy (commented):
This is the right diagnosis and the fix matches how `manifest.mjs` already calls `discoverSfxTracks`, but it looks like it's arrived just after the same bug was already fixed.

`release/v0.26.0` merged #1883 ("fix(sfx): scope SFX Studio publish inventory scan to the full catalog") at 2026-07-14T02:18:50Z, about 15 minutes before this PR's commit. That fix changed the same line in `publicPath()` from `discoverSfxTracks([source], sfxRoot)` to `discoverSfxTracks([...CATALOG.values()], sfxRoot)`. Since `CATALOG` here is built as `new Map(SFX.map((entry) => [entry.key, entry]))` (`scripts/sfx_studio/audio_io.mjs:76`), `[...CATALOG.values()]` and `SFX` carry the same entries, so this PR's `discoverSfxTracks(SFX, sfxRoot)` is functionally identical to what's already on the base branch. That's also why the PR shows as conflicting rather than a clean merge: both changes touch the exact same line independently.

Given that, I think this one can close as superseded rather than merged. Nice catch either way, it's a real bug and the fix direction (full catalog, matching `manifest.mjs`) is correct, it just landed in `release/v0.26.0` slightly earlier via #1883.

PR #1895, opened 2026-07-14
fix(sim): grant a brief aggro-immunity grace on resurrect

jgyy:
## Summary

From the New Player Experience feedback doc, item 4: "When resurrecting at corpse you re-aggro mobs easily (annoying repeat new player deaths)."

`resurrectAtCorpse` (`src/sim/spirit.ts`) revives a player exactly where their ghost is standing, at 50% hp/mana, with no protection. In a dense mob pack (e.g. the level 1-3 wolf field), the same pack that just killed the player is often still standing right on top of the corpse, and passive proximity aggro (`src/sim/mob/locomotion.ts`) immediately re-detects and re-aggros the freshly-revived player, chaining repeat deaths. Note: threat tables are already cleared on death (`src/sim/combat/damage.ts`), so this is purely a fresh-proximity-detection problem, not a stale-threat one.

Adds a short (`REZ_GRACE_SECONDS = 6`) post-resurrection aggro-immunity window: a new `Entity.rezGraceUntil` sim-time field, set on every revive path (`reviveAt` in `src/sim/spirit.ts`, shared by corpse-run, Spirit Healer, and instance-reentry resurrection), and checked by the two passive proximity-detection loops in `src/sim/mob/locomotion.ts`. It only suppresses a mob *newly noticing* the player; it does not affect mobs already in combat, does not touch threat tables, and does not block the player from re-engaging voluntarily (e.g. via auto-attack).

## Related issues

Item 4 of the attached New Player Experience feedback (level 1-3 UX pass).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/spirit.test.ts tests/parity/parity.test.ts tests/world_api_parity.test.ts` (353 passed), `npx tsc --noEmit -p .` (clean), `npx @biomejs/biome check --write` on changed files (clean).
- Manual steps: added a new test (`spirit: post-resurrect aggro grace`) that spawns an idle hostile mob directly on the corpse, resurrects, and asserts the mob stays idle for the grace window and re-aggros normally once it elapses. Regenerated the one affected parity golden (`tests/parity/golden/entity_roster.json`) via `UPDATE_PARITY=1`, since `Entity` gained a field.

## Screenshots / recordings

N/A. Pure `src/sim/` behavior change, no player-visible UI (nothing new renders; the fix only changes whether a mob decides to aggro).

## Diagram

```mermaid
sequenceDiagram
    participant P as Player
    participant S as spirit.ts (reviveAt)
    participant E as Entity.rezGraceUntil
    participant M as mob/locomotion.ts (idle scan)

    P->>S: resurrectAtCorpse()
    S->>E: rezGraceUntil = ctx.time + REZ_GRACE_SECONDS
    S-->>P: dead=false, ghost=false, hp=50%

    loop every idle tick, nearby mobs
        M->>E: is ctx.time < rezGraceUntil?
        alt still in grace window
            M-->>M: skip (player invisible to fresh proximity aggro)
        else grace elapsed
            M->>P: normal aggro pickup resumes
        end
    end
```

---

### Quality

- [x] Ran the relevant test files, `tsc`, and biome directly (couldn't run the full `npm run gate` in this environment; dependencies were symlinked from a sibling checkout for testing).

### Cross-platform

- N/A. No UI/render change.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/entity.ts
+++ b/src/sim/entity.ts
@@ -165,6 +165,7 @@ function baseEntity(id: number, pos: Vec3): Entity {
     ghost: false,
     corpsePos: null,
     corpseInstanceId: null,
+    rezGraceUntil: 0,
     scale: 1,
     color: 0xffffff,
     skinCatalog: 'class',
--- a/src/sim/mob/locomotion.ts
+++ b/src/sim/mob/locomotion.ts
@@ -252,7 +252,7 @@ export function updateMob(ctx: SimContext, mob: Entity): void {
         let detected: Entity | null = null;
         let detectedD = Infinity;
         ctx.playerGrid.forEachInRadius(mob.pos.x, mob.pos.z, 25, (e, d2) => {
-          if (e.dead) return;
+          if (e.dead || ctx.time < e.rezGraceUntil) return;
           const radius = Math.max(
             4,
             Math.min(MAX_AGGRO_RADIUS, template.aggroRadius + (mob.level - e.level) * 1.5),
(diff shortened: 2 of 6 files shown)


jgyy:
this PR to be rejected, instead dont group the mob so tightly instead

jgyy:
Closing in favor of a content-side fix. Re-aggro at a corpse rez in a dense pack is standard classic-era MMO behavior, not itself a bug, and `REZ_GRACE_SECONDS` adds a bespoke aggro-immunity mechanic this project avoids inventing. The actual problem is that the two forest_wolf camps in src/sim/content/zone1.ts place mobs too densely for their aggroRadius (10) and packFrenzy radius (12): expected nearest-neighbor spacing in the r=22, count=7 camp works out to roughly 7.4 units, well inside both. Filed #1908 to retune camp radius/count/placement instead.

PR #1897, opened 2026-07-14
fix(net): re-derive follow-up quest availability immediately after turn-in

### LINKED ISSUE #1667: Bug: NPC dialog doesn't show the next quest immediately after completing the previous in a sequence
### Before you start

### What happened?

When talking to an NPC like Loremaster Caddis, the next quest doesn't immediately show in dialog after completing the previous one.

### Steps to reproduce

1. Complete a quest where you know you can accept the follow-up quest immediately
2. Talk to an NPC, notice the next quest is not visible, but there is an option to complete the current quest
3. Complete the current quest and notice the dialog window now shows the default NPC dialog
4. Close the dialog window
5. Notice there is an exclamation mark above the NPC
6. Talk to the NPC again
7. The next quest is now available

After completing the quest in step 3, I expect the next quest to be available immediately if I meet the requirements without having to talk to the NPC again.

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Firefox on linux mint

1. Complete the quest:

Image

2. Observe the default dialog:

Image

3. Talk to the NPC again to view the next quest:

Image

jgyy:
## Summary

Fixes the NPC dialog not showing a follow-up quest immediately after completing
its prerequisite in the same dialog session (only reproduces online).

`ClientWorld.turnInQuest()` sends the turn-in as a fire-and-forget wire
command; the mirrored `questsDone` set only updates once the server's `qdone`
snapshot field round-trips. The HUD's `renderGossip()` already re-queries
`questState()` live for every quest right after the turn-in click (the dialog
re-render itself was never stale), but `ClientWorld.questState()`'s
prerequisite check (`quest.requiresQuest && !questsDone.has(...)`) read the
not-yet-updated `questsDone`, so any OTHER quest gated on the one just turned
in stayed `unavailable` until the next snapshot arrived a tick or two later.

The fix extends the pending-command optimism `questState()` already applies
to the just-turned-in quest (so it shows as `active`, not stuck at `ready`)
to prerequisite checks for every other quest: while a `turnin` command is in
flight, treat that quest as done when resolving anyone else's state. This is
display-only optimism, reconciled the moment a real snapshot lands and clears
`pendingQuestCommands`; the server still authoritatively resolves the turn-in.

Per the module-first convention, the logic is a small pure, DOM-free module
(`src/net/quest_state_optimistic.ts`) rather than inline in the `ClientWorld`
coordinator, so it is unit-testable without spinning up the network client.

### Flow

```mermaid
sequenceDiagram
    participant Player
    participant HUD as hud.ts renderGossip/renderQuestDetail
    participant CW as ClientWorld (src/net/online.ts)
    participant Server

    Player->>HUD: click "Complete Quest" (quest A)
    HUD->>CW: sim.turnInQuest(A)
    CW->>CW: pendingQuestCommands.set(A, 'turnin')
    CW->>Server: cmd({ turnin: A })  (fire and forget)
    HUD->>CW: renderGossip(npc, true) [same tick]
    CW->>CW: questState(B) via optimisticQuestState
    Note over CW: Before fix: questsDone still lacks A<br/>-> B stays 'unavailable'
    Note over CW: After fix: A is treated as done<br/>(pending turnin, A != B) -> B is 'available'
    CW-->>HUD: B offered immediately
    HUD-->>Player: dialog shows quest B

    Server-->>CW: snapshot { qdone: [...A], qlog: [...] }
    CW->>CW: pendingQuestCommands.clear()
    Note over CW: Reconciled with authoritative state
```

## Related issues

Fixes #1667

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: `npx vitest run tests/quest_state_optimistic.test.ts`, `npx tsc --noEmit -p tsconfig.json`,
  `npm run gate` (see note below)
- Manual steps: traced the code path (`hud.ts` `renderGossip`/`renderQuestDetail`,
  `net/online.ts` `ClientWorld.turnInQuest`/`questState`, `sim/quests/quest_commands.ts`
  `computeQuestState`) to confirm the fix is display-side optimism, not a change to
  server-authoritative quest resolution.
- Added `tests/quest_state_optimistic.test.ts` (new, pure unit tests, no DOM/network):
  reproduces the bug (a `requiresQuest`-gated quest, e.g. `q_greyjaw` requiring
  `q_wolves`, stays `unavailable` with a naive read) and pins the fix (`available`
  the instant the prerequisite's `turnin` command is pending), plus coverage for the
  turned-in quest's own state, post-snapshot fallback, and the "no pending command"
  negative case.

Note on `npm run gate`: the full vitest run on this (heavily loaded, many
concurrent worktrees) box shows unrelated pre-existing failures/timeouts in
`tests/mail.test.ts`, `tests/loot_ffa_sim.test.ts`, `tests/vale_cup_online.test.ts`,
`tests/sfx_studio_server_security.test.ts`, `tests/server/http/parity.test.ts`, and
`tests/server/new_endpoint.test.ts` (a `tsc` golden subprocess). None of these touch
`src/net/`, quest state, or the dialog path; re-running them in isolation still times
out under the current machine load (100+ concurrent node/vitest/tsc processes from
other sessions), and a full standalone `npx tsc --noEmit -p tsconfig.json` passes
clean. `npx vitest run tests/quest_state_optimistic.test.ts` is green.

## Screenshots / recordings

N/A. Timing/data-freshness bug in the dialog re-render logic, not a visual change;
the dialog markup is unchanged. Covered by the new pure-module test instead.

---

### Quality

- [x] Decisive tests added for the changed behavior (`tests/quest_state_optimistic.test.ts`).
      Full `npm run gate` run attempted; see the note above for unrelated pre-existing
      flakiness on this machine, isolated from this change (`tsc --noEmit` clean, new test green).

### Cross-platform

- [ ] N/A. No UI markup or styling changed.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/net/online.ts
+++ b/src/net/online.ts
@@ -35,7 +35,7 @@ import { normalizeMoveFacing, sanitizeMoveInput } from '../sim/move_input';
 import { getArchetypeTitle, getHobbyCraft } from '../sim/professions/archetype';
 import type { MaterialRarity } from '../sim/professions/gathering';
 import { emptyCraftSkills } from '../sim/professions/wheel';
-import { computeQuestState, type ResolvedAbility } from '../sim/sim';
+import type { ResolvedAbility } from '../sim/sim';
 import {
   type DeedStats,
   type DungeonDifficulty,
@@ -97,6 +97,7 @@ import {
   type SocialInfo,
   type TradeInfo,
 } from '../world_api';
+import { optimisticQuestState } from './quest_state_optimistic';
 import { isTransientReconnectRejection } from './reconnect_policy';
 
 // ---------------------------------------------------------------------------
@@ -2208,15 +2209,13 @@ export class ClientWorld implements IWorld {
   // -----------------------------------------------------------------------
 
   questState(questId: string): QuestState {
-    const state = computeQuestState(questId, this.questLog, this.questsDone, this.player.level);
-    const pending = this.pendingQuestCommands?.get(questId);
-    if (
-      (pending === 'accept' && state === 'available') ||
-      (pending === 'turnin' && state === 'ready')
-    ) {
-      return 'active';
-    }
-    return state;
+    return optimisticQuestState(
+      questId,
+      this.questLog,
+      this.questsDone,
+      this.pendingQuestCommands,
+      this.player.level,
+    );
   }
 
   consumeInventoryChanged(): boolean {
--- a/src/net/quest_state_optimistic.ts
+++ b/src/net/quest_state_optimistic.ts
@@ -1,0 +1,51 @@
+// Pure optimistic quest-state resolution for the online client.
+//
+// Bug fixed (issue 1667): after turning in a quest through the NPC dialog, a
+// follow-up quest that requires it (`requiresQuest`) stayed hidden until the
+// next server snapshot arrived, even though the dialog re-renders immediately
+// after the turn-in command is sent. `ClientWorld.turnInQuest` is fire-and-
+// forget: it records the action in `pendingQuestCommands` but the quest only
+// actually leaves `questLog` / lands in `questsDone` once the server's `qlog`/
+// `qdone` snapshot fields round-trip. The dialog's re-render (`hud.ts`
+// `renderGossip`) already re-queries `questState()` live for every quest the
+// NPC could offer, so the staleness was never in the UI layer: it was that
+// `computeQuestState`'s prerequisite check (`quest.requiresQuest &&
+// !questsDone.has(...)`) read the not-yet-updated `questsDone` set for any
+// quest OTHER than the one just turned in.
+//
+// The fix extends the existing pending-command optimism (already sanctioned
+// for the turned-in quest itself, see src/net/CLAUDE.md) to prerequisite
+// checks for every other quest: while a `turnin` command is in flight, treat
+// that quest as done when resolving anyone else's state. This is display-only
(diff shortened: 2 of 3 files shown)


Rubsey (commented):
This one is clean. The fix is the right kind of optimism: it delegates to the shared `computeQuestState` rather than re-implementing prerequisite rules (so it cannot drift from the sim), clones `questsDone` copy-on-write only when something is pending, and stays display-only. The lifecycle is sound end to end: `pendingQuestCommands` is initialized inline (never undefined), cleared on any `qlog`/`qdone` snapshot (`src/net/online.ts:2080`), and the dangerous case, a rejected turn-in leaving phantom optimism, is covered because the server force-resyncs quest state after every quest command regardless of outcome, so the pending entry always clears within one round-trip. No wire or entity-shape changes, IWorld surface untouched, offline Sim behavior already matched what this makes the mirror show. Tests are decisive: the follow-up-available case returns `unavailable` on the old code path, and the no-pending negative arm pins that the optimism does not leak.

One small thing:

- **nit** `src/net/quest_state_optimistic.ts:16-18`: the comment cites `src/net/CLAUDE.md` as sanctioning this, but that doc's optimism note (the "keep that scope" sentence around lines 121-123) still describes the old narrow scope (the pending quest's own display state). Since this PR deliberately widens it to prerequisite resolution, update that sentence in the same change so the doc and the code agree on what is sanctioned.

PR #1898, opened 2026-07-14
fix(gathering): give gather nodes a harvest trigger so A Trade for Every Hand is completable

### LINKED ISSUE #1888: Critical: gathering nodes have no harvest trigger, making A Trade for Every Hand uncompletable
## Problem

Gathering nodes (ore, wood, herb) cannot be harvested by a player. There is no key, no HUD button, no mobile button, no clickable vein, and no slash command anywhere in the client that triggers a harvest, so the entire gathering mechanic is unreachable in game.

The visible consequence is that "A Trade for Every Hand" (`q_prof_intro`, the professions onboarding quest given by Foreman Odell) is uncompletable. Its only completion path is a collect objective for 5x `chunk_of_ore`, and `chunk_of_ore` is granted exclusively by an ore-node harvest (the `NODE_QUEST_GRANT` gate in `src/sim/professions/gathering.ts`). A player accepts the quest, walks to the Copper Dig, sees the veins, and has no way to mine them. This was reported by a player in Discord on mobile web.

Everything downstream of the trigger is already correct. `IWorld.harvestNode` is declared on the facet (`src/world_api/professions.ts`), implemented on `Sim` (`src/sim/professions/gathering.ts` `harvestNode`), sent by `ClientWorld` as the `harvest_node` command (`src/net/online.ts`), and dispatched by the server (`server/game.ts`). Feeding a hand-built `harvest_node` frame into `GameServer.handleMessage` grants the ore and ticks quest credit. The wire is complete end to end; the frame is simply never constructed, because nothing in `src/ui/`, `src/game/`, `src/render/`, or `src/main.ts` ever calls `harvestNode`. `git log -S"harvestNode" -- src/ui src/game src/render src/main.ts` returns empty: no commit has ever added a caller.

## Steps to reproduce

1. Create a character and accept "A Trade for Every Hand" from Foreman Odell in Eastbrook.
2. Optionally buy a Copper Mining Pick from Trader Wilkes (it is not required, and it does nothing).
3. Walk southwest to the Copper Dig and stand on top of an ore vein (`ore_eastbrook_1` at x -70, z -53).
4. Try to mine it: tap or click the vein, press the Interact key, tap the mobile Interact button, and use the pick from the bag.
5. Observe that none of these harvest the node.

## Actual behavior

- Tapping or clicking the vein does nothing. Node meshes are added as a frozen static prop group with no `entityId` (`src/render/renderer.ts`), and `renderer.pick()` only raycasts `clickTargets` and resolves an entity id, so a pixel-perfect tap returns `null`. With click to move enabled it just walks the player toward the vein.
- Pressing Interact (or tapping the mobile Interact button, which routes to the same `interactKey()` in `src/main.ts`) shows "Nothing to interact with." That loop scans `world.entities` and accepts only lootable corpses, delve objects, lootable objects, and NPCs. Gather nodes are static content records, never entities, so it cannot see them.
- Using a mining pick from the bag is a silent no-op. `copper_mining_pick` is `kind: 'tool'` with a `use: { type: 'gatherTool' }` payload (`src/sim/content/items.ts`), but `useItem` (`src/sim/items.ts`) branches on fishing, mechChroma, skinSelect, then food, drink, potion, elixir, weapon, armor, and bag, and ends with no trailing else. A tool matches nothing, falls off the end, and returns with no error, no log, and no event. The tooltip still advertises "Click to use" because the item has a truthy `use` field.
- The quest objective stays at `Chunk of Ore: 0/5` forever.

## Expected behavior

- A player standing within interact range of a gather node can harvest it, on desktop and on mobile, and receives the node's material plus (while the quest is active) the quest item.
- The mobile Interact button and the desktop Interact key both harvest a nearby node, consistent with how they already loot corpses and talk to NPCs.
- "A Trade for Every Hand" is completable end to end: accept from Foreman Odell, mine 5 chunks at the Copper Dig, turn in for the 150 xp and 50 copper.
- Every harvest attempt gives the player feedback. A denial (too far away, node not respawned for you yet, bags full) surfaces the existing error, and a success is visible.
- Using an item that cannot be used produces a message rather than silently doing nothing.

## Scope

- Input and picking: give gather nodes a pick or interact path. `renderer.pick()` is typed `number | null` and `handlePickedEntity` is entity-kind-only, so this needs either a string-id pick channel for nodes or promotion of nodes to real entities. Decide which before implementing.
- `interactKey()` (`src/main.ts`): add a gather-node arm, since this is what the mobile Interact button hits.
- `useItem` (`src/sim/items.ts`): add a `gatherTool` arm, or at minimum an error message, so a vendor-sold item with a `use` payload never silently does nothing.
- `ClientWorld.nodeHarvestableByMe` (`src/net/online.ts`) is a hardcoded `return true`, so once a real trigger lands the online minimap would paint every vein permanently "ready". The per-node respawn state needs to reach the client.
- Latent, currently unreachable: in `harvestNode` (`src/sim/professions/gathering.ts`) the quest-item grant is skipped with no message when bags fill after the material takes the last slot. Worth fixing in the same change.
- Tests: every existing gathering test calls `sim.harvestNode(...)` directly, and the `world_api_parity` pin only proves both worlds implement the member, never that anything calls it. That is why this shipped green.

## Acceptance criteria

- A player can harvest an ore, wood, or herb node in game, on both desktop and mobile web, without dev commands.
- "A Trade for Every Hand" can be completed start to finish by a new level 1 character with no pickaxe and no learned profession, matching the current sim gates.
- A failed harvest surfaces the existing denial message rather than doing nothing.
- Using a mining pick from the bag produces player-visible feedback.
- The online minimap distinguishes a vein that is ready for the local viewer from one on that viewer's personal cooldown.
- A test drives the harvest through the player-input path (not by calling `sim.harvestNode` directly), so a future regression that removes the trigger fails CI.
- Existing gathering, quest, and parity tests stay green, and the desktop experience is unchanged apart from the new harvest affordance.

## Environment

- Platform: mobile web (reported), and desktop web by inspection. Offline and online are both affected.
- Surface: 3D world interaction, the mobile Interact button, and the bags window.

## Notes

Likely origin: commit `432c9c218d` (#1121) landed the sim function, the `harvest_node` WS command, and the `ClientWorld` sender in one change and stopped there, with no UI or input trigger. #1701 and #1708 then built the onboarding quest on top of that mechanic, so the quest inherited an unreachable completion path.

The guide already tells players "Walk up to one and interact with it to harvest whatever it holds" (`src/ui/i18n.catalog/guide.ts`), which does not currently work. The same copy notes that no tool is required to work a node today, so the mining picks are intentionally inert for now; the defect there is the total absence of feedback, not the missing tool gate.

Follow up once this ships: reply to Kirito in Discord, who reported it.
https://discord.com/channels/1515097174378020894/1515257339203289108/1526357658800754871

## Screenshot

Image

Rubsey:
Fixes #1888.

## Problem

Gathering nodes (ore, wood, herb) had no harvest trigger anywhere in the client. The sim function, the `harvest_node` wire command, the server dispatch, and even the minimap ready/cooldown UI all existed, but nothing in `src/ui/`, `src/game/`, `src/render/`, or `src/main.ts` ever called `IWorld.harvestNode`, so the whole mechanic was unreachable in game and "A Trade for Every Hand" (`q_prof_intro`) was uncompletable. Reported by a player on mobile web.

## What this does

- **Interact-path trigger (desktop + mobile).** New sibling module `src/game/gather_interact.ts`: `tryHarvestNearestNode(world)` resolves the nearest in-range node via the same per-viewer classification the minimap already uses (`buildNearbyGatherNodes`), prefers a ready node over a nearer cooldown one, and dispatches `world.harvestNode` without predicting the outcome (every sim denial surfaces through the normal error-event toast). `interactKey()` consumes it as the lowest-priority arm, so the desktop Interact key and the mobile Interact button are both covered by one shared path and corpses/delve objects/lootables/NPCs keep their precedence.
- **Real per-viewer node readiness online.** `ClientWorld.nodeHarvestableByMe` was a hardcoded `return true`, so the online minimap painted every vein permanently ready. The server now ships the anchor player's still-cooling timers as a new delta-guarded `maybe('gnodes')` self field (filtered server side to future entries), the client answers by pure membership, unknown node ids read false in both worlds (Sim parity), and late join/reconnect get the full map on their first snapshot. Rolling-deploy safe in both directions.
- **Mining pick feedback.** `useItem` had no arm for `kind: 'tool'`, so using a vendor-sold pick was a silent no-op while its tooltip advertised "Click to use". A new gatherTool arm emits a hint pointing at the interact flow; picks stay intentionally inert per the guide copy (no tool gate is being added here). The new sim literal ships with its `log.gatherToolHint` matcher and translations for all 21 non-English locales in the same change, so the release-tier i18n gate stays clean (zero new pending rows).
- **Bag-full quest-item fix.** `harvestNode` silently skipped the quest-item grant when bags filled after the material took the last slot. The capacity gate is now joint (material plus needed quest item) and denies the whole attempt up front with the existing bags-full error, preserving the per-player respawn timer; the residual one-free-slot race (both pre-checks pass, the material add consumes the slot) still grants the material but errors loudly instead of silently skipping.

## Test plan

- `tests/gather_interact.test.ts` (new): drives the harvest through the player-input entry point against a real offline Sim (the regression test the original change never had), covers ready-over-cooldown selection, denial surfacing, the full level 1 `q_prof_intro` accept, mine 5, turn-in loop, and a source-level wiring guard so removing the `main.ts` call site fails CI.
- `tests/gather_online.test.ts` (new): first-ever test driving `harvest_node` frames through `GameServer.handleMessage`: grant, per-session denial routing, the server distance gate, malformed-frame rejection, and the `gnodes` delta riding the snapshot.
- `tests/snapshots.test.ts`: `gnodes` registered as the 43rd delta key (scrape, sorted set, dirty fixture, no-op omission), decode asserted directly on the `nodeHarvestableByMe` method surface (the `tal` convention), plus the expiry flip back to ready.
- `tests/gather_node_harvest.test.ts`: joint bag gate (denial, timer preserved, zero rng draws), the one-slot race erroring loudly, the `readyAt` boundary agreement between `isNodeHarvestableBy` and the `gnodes` encoder, and the tool-hint emit for pick, axe, and sickle.
- `npm run gate` green (all 11 steps), plus the release-tier i18n pass (`I18N_RELEASE_TIER=1`) run locally since this PR targets a release branch.

## Notes and follow-ups (out of scope here)

- Tap/click-on-vein picking is deferred: node meshes are a frozen static prop group with no entity id, so a pick channel is a renderer change no acceptance criterion needs once Interact works; the guide copy documents interact semantics ("Walk up to one and interact with it"), which this PR makes true.
- Extracting the full `interactKey()` body out of `main.ts` is a worthwhile module-first refactor for a non-release branch.
- `gatherResult` still has no client audio consumer (the #1729 cue); harvest success feedback is the existing loot line plus quest progress.
- After merge: reply to Kirito in the Discord thread linked in the issue.


FernandoX7 (commented):
Good fix, and the right shape: the mechanic was complete everywhere except the one place that mattered (no input path ever called `harvestNode`), and this closes it through the existing seams instead of around them. I verified the load-bearing claims at the head rather than taking the description's word: unknown node ids read false in both worlds through the same `gatherNodeById` guard (`src/sim/sim.ts:6050`, `src/net/online.ts:1218`), the `readyAt` boundary agrees between `isNodeHarvestableBy` (`>=`) and the `gnodes` encoder (strict `>`), the delta guard re-sends exactly on harvest and expiry (`nodeHarvestReadyAt` entries are only ever assigned in place, so the fresh record's key order is stable and the JSON compare in `maybe()` never fires spuriously), pid-carrying log and error events route only to the owning session, and all three input paths (Interact key, mobile button, gamepad) funnel through `interactKey()`. Rolling-deploy safety holds in both directions. I also ran all four test files plus the S3 guard at the PR head (169 passed) and confirmed `i18n:gen` reproduces the committed generated artifacts byte for byte.

Findings:

- **should-fix** `tests/gather_interact.test.ts:98`: the "prefers the nearest node" half of this test is not actually pinned. NODE_A is both the nearest node and first in content order, so a mutant `nearestOf` that just returns the first ready node in `GATHER_NODES` order passes all six tests (I ran that mutation to confirm). One extra press standing measurably closer to NODE_B, so distance and content order disagree, would close it.
- **nit** `tests/gather_node_harvest.test.ts:466`: the title says "every gathering tool kind (pick, axe, sickle)" but the loop covers only `handaxe` and `gathering_sickle`; the pick is exercised in the previous test. Coverage is fine across the pair, the title over-claims for this one.
- **nit** `tests/gather_interact.test.ts:68`: the "wood_eastbrook_1 has no neighbor within INTERACT_RANGE" premise lives only in a comment (it holds today, the nearest neighbor is about 11.7yd out). An inline distance assertion like the one this file already uses at line 108 would keep it honest if content placements move.

On the merge conflict: the only true conflict is `src/ui/i18n.status.summary.json`, which is mechanical regen churn from the base moving. Merge the base and re-run the i18n scan; nothing to resolve by hand.

The behavior changes themselves check out. The joint capacity gate is a clean deterministic change, not a draw-order hazard (the parity digest is green and unregenerated, and the zero-draw denial is pinned), the one-slot race staying loud while still granting the material matches the stated intent, and the gather-tool hint arm cannot be shadowed in the `useItem` chain since every gather tool is `kind: 'tool'`.

Rubsey:
Closing as superseded, not merging.

release/v0.26.0 has since shipped gather-node interaction wiring independently via #1937 (feat(game): wire gather node harvesting into normal play controls). That change closes the same reachability gap this PR targeted for issue #1888, and goes further by adding the click/tap-on-mesh channel this PR had deliberately deferred. Because of that, the client-side wiring here now collides head-on with what landed on the base (src/main.ts, src/net/online.ts, and the self-wire delta contract in tests/snapshots.test.ts, where #1937 introduced its own 'ncd' cooldown field), and rebuilding this branch on top of #1937 would duplicate work already in place.

The two sim-layer changes here that #1937 did not touch (the atomic bags-full quest-item capacity gate in harvestNode, and the gather-tool useItem hint) are being set aside for now rather than carved into a follow-up. Thanks for the review notes; the nearest-selection and tool-coverage pinning points are good and worth keeping in mind for future gathering work.

The branch is left intact for reference.

PR #1899, opened 2026-07-14
fix(ui): route mob combat-flavor chatter out of General/Chat

jgyy:
## Summary

From the New Player Experience feedback doc, item 6/QoL-4: "Still need to remove other people's combat mechanics from the primary chat screen. Ideally we want new players to see real people talking in general chat, not being spammed with other players' combat messages" (screenshot: `[13:31] Forest Wolf flies into a frenzy!` / `Vale Bandit attempts to flee!` lines flooding the Chat tab in Wolf Run/Eastbrook).

A separate Combat Log tab already exists (`#combatlog`, `Hud.combatLog()`) and correctly receives damage numbers and aura gain/fade events. Mob-mechanic flavor text (pack frenzy, flee, enrage, boss channels/mechanics, quest-boss dialogue) never routed there: it's emitted as an untyped `'log'` `SimEvent` and the client's `case 'log':` dispatch (`src/ui/hud.ts`) sent every such line to `this.log()`, which appends into the same `#chatlog` element as General/Say/Party. Since the server delivers these events to every player within interest range (~120yd, per `CLAUDE.md`), any player standing near ANY nearby fight sees every engaged mob's flavor text in their own Chat tab, not just their own combat.

Every `'log'` event that carries `entityId` (per the field's own existing comment: "entityId anchors the log to that entity so the server only delivers it to nearby players") is exactly this category, combat/encounter chatter, never a genuine chat/system notice (those use `pid` or nothing). Added a small pure module, `src/ui/log_event_route.ts` (`isCombatFlavorLog`), and used it in the `case 'log':` dispatch to route those lines to `combatLog()` instead of `log()`. Narrative NPC lines (boss vision lines, yells) still get their existing floating world chat bubble regardless of which tab their text lands in.

## Flow

```mermaid
flowchart TD
    A["Server emits 'log' SimEvent<br/>(interest-scoped, ~120yd)"] --> B{"entityId set?"}
    B -->|"yes: mob/boss anchored<br/>(frenzy, flee, enrage, boss mechanic)"| C["isCombatFlavorLog = true"]
    B -->|"no: pid-only or anchorless<br/>(ability learned, world boss spawn, catch)"| D["isCombatFlavorLog = false"]
    C --> E["Hud.combatLog()<br/>#combatlog tab"]
    D --> F["Hud.log()<br/>#chatlog General/Say/Party tab"]

    style C fill:#2d5,color:#000
    style D fill:#58c,color:#000
```

Before this change, both branches fell through to `Hud.log()`, so nearby players' mob-flavor chatter flooded the same tab as real player conversation.

## Related issues

Item 6 / QoL item 4 of the attached New Player Experience feedback (level 1-3 UX pass).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx vitest run tests/log_event_route.test.ts tests/architecture.test.ts tests/chat_log.test.ts tests/chat.test.ts` (all green; one pre-existing unrelated flaky timeout in `chat.test.ts`'s `/playtime` test reproduces identically on the unmodified base branch), `npx tsc --noEmit -p .` (clean), `npx @biomejs/biome check --write` on changed files (clean).
- Manual steps: added `tests/log_event_route.test.ts` pinning the routing predicate (entityId-anchored -> Combat Log, anchorless -> General/Chat), and registered the new pure core in `tests/architecture.test.ts`'s `UI_PURE_CORES`/`BARE_NAMED` allowlists per the module-first convention in `src/ui/CLAUDE.md`.

## Screenshots / recordings

Same synthetic combat-flavor line (`The Greyjaw Ravager flies into a frenzy!`, entityId-anchored) and system line (`Talents updated.`, pid-only) driven through the real `Hud.handleEvents` dispatch, shown on both chat tabs, before and after this fix:

| | General/Chat tab | Combat Log tab |
|---|---|---|
| **Before** (base branch) | before general | before combat log |
| **After** (this branch) | after general | after combat log |

Before: the mob-flavor line floods General/Chat alongside real system text, and the Combat Log tab is empty. After: General/Chat only shows the genuine system line, and the mob-flavor line is routed to the Combat Log tab. Captured via a new change-aware target (`chat-general-tab` / `chat-combat-tab`) added to `scripts/pr_shot_targets.mjs`.

---

### Quality

- [x] Ran the relevant test files, `tsc`, and biome directly (couldn't run the full `npm run gate` in this environment).

### Cross-platform

- N/A. No layout/visual change.

### Localization (i18n)

- [x] N/A. No new player-visible strings; existing localized text is only re-routed to a different existing tab.

### Hygiene


Rubsey (commented):
The problem is real (the feedback item is right, other players' mob chatter drowning General/Chat is bad for new players), and the mechanism is the correct shape: a pure predicate module, the hud as a thin consumer, a sanctioned allowlist addition in the architecture test (UI_PURE_CORES + BARE_NAMED, no guard weakened), `localizeSystemText` still applied before routing, and the screenshot harness targets are a nice touch.

The predicate itself is too blunt, though:

- **blocking** `src/ui/log_event_route.ts:12`: `entityId !== undefined` sweeps actionable mechanic telegraphs out of the default view, and at least one has no other signal at all. The default chat tab is `all` (`src/ui/hud.ts:1038`) and the two panes are mutually exclusive, so an entityId-anchored line is invisible until the player switches tabs. "Deacon Varric begins Raise Dead." (`src/sim/delves/runs.ts:1258-1266`) is an interruptible channel (the cracked-grave interact clears it) with no cast bar, no spellfx, and no overhead bubble: the bubble path only fires for the Nythraxis vision lines or lines containing `yells, "` (`src/ui/hud.ts:10024-10029`). After this PR that mechanic's only cue sits buried in the Combat Log among per-hit spam. Same category: the corpse-swell "get clear" warning (`src/sim/mob/lifecycle.ts:186-191`, whose own comment calls it a telegraph), the Blackwater mark callout, Final Bell. The module's docstring claims anchored lines are never actionable, but the emitters say otherwise. I would keep the routing for genuine barks (frenzy, flee, quest-boss dialogue) but make the classification explicit at the emitter (a flavor/telegraph flag on the event, or an exclude list to start), or land a telegraph surface (cast bar / raid-warning style banner) for these lines first and then route freely.
- **should-fix** `src/ui/log_event_route.ts:12`: the ten Nythraxis vision lines are pid-scoped personal narrative (`src/sim/encounters/nythraxis.ts:1402-1412` sets both `pid` and `entityId`, and the server delivers pid events only to that player), so they are never someone else's spam, yet the predicate routes them to the Combat Log because it only inspects `entityId`. Their bubble survives, but the transcript copy of a story beat now lands between damage lines. These want to stay in Chat, which also contradicts the module comment's own claim that pid-scoped notices do.
- **nit** `tests/log_event_route.test.ts`: the two assertions restate the predicate (`entityId !== undefined`), and the actual hud dispatch branch has no automated test; fine under the pure-core convention, but once the predicate grows real classification logic per the above it will want real cases.

jgyy:
Thanks for the detailed review, Rubsey. You are right on both counts, and I have pushed a fix that keeps the pure-predicate-plus-thin-consumer shape but corrects the classification.

The predicate (src/ui/log_event_route.ts) now takes pid and a new telegraph flag in addition to entityId:

- pid set: always General/Chat, regardless of entityId. This covers the ten Nythraxis vision lines (src/sim/encounters/nythraxis.ts), which are pid-scoped personal narrative and were being misrouted to the Combat Log purely because they also carry an entityId.
- telegraph: true (a new optional field on the log SimEvent variant in src/sim/types.ts): also always General/Chat, even with an entityId. I set this explicitly at the four mechanic emit sites you called out: Deacon Varric's "begins Raise Dead" channel (src/sim/delves/runs.ts), the corpse-swell "get clear" warning (src/sim/mob/lifecycle.ts), the Blackwater Mark callout, and Final Bell (both src/sim/delves/drowned_litany_boss.ts).
- Otherwise, entityId set routes to the Combat Log as before, for genuine ambient barks (frenzy, flee, quest-boss dialogue) that carry no mechanical weight.

I went with an explicit opt-in flag at the four known telegraph emit sites rather than an exclude list or inverting the whole predicate, since it is the smallest, most auditable diff and future telegraph emitters just need to set the one field.

Updated the module docstring to state this rule plainly instead of the old blanket claim, and tests/log_event_route.test.ts now has real cases for all three branches (pid-scoped, telegraph, and genuine ambient bark), plus the original anchorless case.

Ran the full targeted suite (log_event_route, architecture, the delve/lifecycle/nythraxis tests touching the four emit sites) plus tsc, all green.

ryan-foo (approved):
Approving. Combat-flavor lines anchored to an entityId now route to the Combat Log pane and out of the default General tab, which is the ask. The new module is registered in UI_PURE_CORES, i18n is untouched (it only re-routes already-localized text and the sim_i18n matchers still fire), and the telegraph flag correctly keeps the genuinely actionable cues (Blackwater mark, Final Bell, Raise Dead, the death-throes get-clear) in General.

One thing to confirm before merge: a few other entityId-anchored lines that are arguably actionable are not flagged telegraph and now go to Combat Log too. The mob heal channels (the interrupt cue, "{mob} channels mend/ward") and the AoE "unleashes stomp/bigCast" bursts are the notable ones. Those have their own cast bar or spell nova, so it may be a deliberate line, I just want to make sure it is intended rather than incidental.

PR #1902, opened 2026-07-14
feat(ui): make the on-kill XP floater more pronounced

jgyy:
## Summary

From the New Player Experience feedback doc, item 7 (QoL): "Can we make the XP drop on kill more pronounced/gamified? Currently not that easy to see and disappears quickly." (screenshot: a small "+50 XP" floater competing visually with damage numbers).

The XP floater (`src/ui/hud.ts`'s `case 'xp':`, built via `fctSpawnShape`/`fct_core.ts`) was styled the same as an ordinary miss/dodge word: 17px, muted low-contrast purple (`#b974ff`), and the shared 1.25s TTL used by every high-volume damage tick. `crit` (the only thing that currently makes an FCT bigger with a scale-in pop) is combat-semantic and hardcoded `false` for xp, so it never got any visual emphasis.

Added a distinct, lighter "pop" for the xp/rested-xp kinds (not the crit pop, so a real critical hit still reads as the bigger event), a brighter color, a bigger base size, and a new `FCT_XP_TTL_MS` (1800ms vs the shared 1250ms) since XP is an informational reward, not per-hit combat spam. This stays entirely inside the existing `fct_core.ts` pure-core + `hud.css`/`hud.mobile.css` painter-styling family (per `src/ui/CLAUDE.md`'s "reuse a family before bespoke" rule, no new module). Desktop and mobile both get matching keyframes, and the existing low-effects-tier "shed the pop, keep the size/color" precedent (`.fct.crit` at `data-fx-level="low"`) is reused for the new xp-pop too, so this stays gameplay-neutral per the graphics-settings-fairness invariant (XP amount is cosmetic, not actionable combat info).

## Related issues

Item 7 (QoL section) of the attached New Player Experience feedback (level 1-3 UX pass).

## Type of change

- [x] Feature: new functionality (visual polish)

## How was this tested?

- Commands: `npx vitest run tests/fct_core.test.ts tests/fct_mobile_css.test.ts tests/architecture.test.ts` (45 passed), `npx tsc --noEmit -p .` (clean), `npx @biomejs/biome check --write` on changed files (clean, no fixes needed).
- Manual steps: extended `tests/fct_core.test.ts` to pin the new `FCT_XP_TTL_MS` constant and per-kind ttl split (xp/rested-xp get the longer ttl, every other kind keeps the original `FCT_TTL_MS`). I was not able to run the game client in a browser in this environment to capture a live before/after screenshot; the CSS keyframe/selector changes are covered by the existing `fct_mobile_css.test.ts` guard style (source-scan pins), but a reviewer running `npm run dev` should visually confirm the XP floater's new pop/size/color and that it does not visually crowd out real damage numbers in combat.

## Diagram

```mermaid
flowchart LR
    subgraph before ["Before"]
        B1["17px, muted purple #b974ff<br/>no scale-in pop<br/>1.25s ttl (same as a damage tick)"]
    end
    subgraph after ["After"]
        A1["20px, brighter #d9a3ff<br/>own scale-in pop (1.25x, lighter than crit's 1.5x)<br/>1.8s ttl (FCT_XP_TTL_MS)"]
    end
    K["Mob kill -> 'xp' SimEvent"] --> before
    K --> after
    before -. "same fade-out speed as every damage number,<br/>easy to miss in a busy fight" .-> X["complaint: 'not easy to see, disappears quickly'"]
    after -. "reads as a distinct reward,<br/>still smaller than a crit hit" .-> Y["fix"]
```

## Screenshots / recordings

Live before/after captures: an offline client, a real `'xp'` SimEvent fed through `hud.handleEvents` (the same path a mob kill takes), against this branch and its `release/v0.26.0` merge-base commit.

**Before** (17px, muted `#b974ff`, no scale-in, shared 1.25s ttl):

https://github.com/levy-street/world-of-claudecraft/blob/feat/xp-fct-pop/docs/screenshots/xp-fct-pop/before-xp-floater.png?raw=true

**After** (20px, brighter `#d9a3ff`, `fct-xp-pop` scale-in, 1.8s `FCT_XP_TTL_MS`):

https://github.com/levy-street/world-of-claudecraft/blob/feat/xp-fct-pop/docs/screenshots/xp-fct-pop/after-xp-floater.png?raw=true

---

### Quality

- [x] Ran the relevant test files, `tsc`, and biome directly (couldn't run the full `npm run gate` in this environment).

### Cross-platform

- [x] Added matching mobile keyframes/selectors in `hud.mobile.css` (touch gets its own smaller pop/travel, same as the existing `.fct.crit` mobile treatment); not manually verified on a device in this environment.

### Localization (i18n)

- [x] N/A. No new player-visible strings; the existing `hud.core.xpFloat` t() text is unchanged.

### Hygiene


Rubsey (commented):
The shape of this is right: the ttl split lives in the pure core with a decisive test (old `describeFct` returns 1250 for xp, so the new case fails on base), the pop reuses the crit precedent including the low-tier shed via `data-fx-level` (gameplay-neutral, correctly not touching the governor), `reduce-motion` still wins via the existing `!important` rule, and the mobile re-declare with its own keyframes is the correct parallel given the layer/specificity situation.

What blocks it is the half-finished re-mint:

- **blocking** `tests/fct_painter.test.ts:224` and `:596-608`: the PR gate is red with two stale pins that are direct, intended consequences of this change. The recycle test spawns an `xp` event and asserts it is gone at `FCT_TTL_MS` (now it lives 1800ms: expected 0, got 1), and the colour-faithfulness map still pins `.fct-xp: #b974ff` / `.fct-rested-xp: #4a9eff`. Re-mint both deliberately: step the recycle test with `FCT_XP_TTL_MS` (or spawn a 1250ms kind like `heal` there) and update the pinned hexes to `#d9a3ff` / `#6db8ff`.
- **should-fix** `src/styles/hud.css:2899-2905`: the colour inventory comment directly above the rules still documents the old hexes ("xp (#b974ff); rested xp (#4a9eff)"); the next editor or test re-mint will copy the wrong values from it.
- **should-fix** new CSS surface is unpinned: `src/styles/CLAUDE.md` says literal mobile FCT values are pinned by `tests/fct_mobile_css.test.ts`, and every sibling value there is (fct-rise-mobile 56px, the crit sizes, the low-fx crit re-point), but `fct-xp-pop` / `fct-xp-pop-mobile`, the 20px/17px xp sizes, and both new low-tier shed rules have no pins at all. Add them alongside the crit ones.
- **nit** `tests/fct_core.test.ts:119-121`: the describe/it titles still say ttl is "constant across kinds" / "for every kind" while the block now deliberately excludes xp/rested-xp.
- **nit** PR body: there is an em dash in the third summary paragraph; the repo bans them in PR text too. Easy edit.

One heads-up for the re-mint rather than a finding: the rested-xp floater changes colour and size for existing players as part of this; it is disclosed in the body and coherent as a family change, just making sure it is intentional.

jgyy:
Thanks for the thorough pass, Rubsey. Re-minted all of it:

- `tests/fct_painter.test.ts` recycle test now spawns a `heal` event (FCT_TTL_MS = 1250ms) instead of `xp`, so it still steps at the correct TTL boundary without being affected by the new xp TTL split.
- Colour-faithfulness map updated to the real post-change hexes: `.fct-xp: #d9a3ff`, `.fct-rested-xp: #6db8ff` (verified these are the exact values in `src/styles/hud.css`).
- `src/styles/hud.css` colour inventory comment (around line 2903) now documents `#d9a3ff` / `#6db8ff` instead of the old hexes.
- Added the missing pins to `tests/fct_mobile_css.test.ts`, following the existing crit pin pattern: desktop `.fct-xp` / `.fct-rested-xp` at 20px pointed at `fct-xp-pop`, the desktop low-fx shed rule, the mobile `fct-xp-pop-mobile` keyframe travel (56px, shorter than desktop), the mobile 17px size plus animation swap, and the mobile low-fx shed rule.
- `tests/fct_core.test.ts` describe/it titles reworded to "constant across kinds (excluding xp/rested-xp)" / "for every kind except xp/rested-xp" to match the block's actual scope.
- PR body: replaced the em dash in the third summary paragraph with a comma inside the existing parenthetical ("... rule, no new module).").

Confirmed intentional on the rested-xp colour/size change for existing players, it is disclosed in the body and consistent with the family change.

Ran `npx vitest run tests/fct_painter.test.ts tests/fct_core.test.ts tests/fct_mobile_css.test.ts` (all green) and `npx tsc --noEmit` (clean). Pushed to the branch.

ryan-foo (approved):
Approving. Clean cosmetic change: bigger font, a scale-in pop, a longer read window, and a brighter color. The FCT pool and FIFO cap, the per-frame write-elision path, and the low-tier and reduce-motion sheds are all untouched, and the colors stay in hud.css rather than becoming magic hex in the painter. The recycle test was correctly repointed off the xp floater since its TTL grew. No i18n or fairness concern since a floater is purely cosmetic.

Issue #1893, opened 2026-07-14
Incorrect equip restrictions on Fang of Korzul and Deathlord Warplate


jgyy:
## Problem

Two related reports describe item equip gating giving incorrect results: a druid is
told they cannot equip Fang of Korzul with no in-game explanation of why, and a player
is told Deathlord Warplate cannot be equipped by their class even though the item's
own tooltip does not state it is class-restricted. Separately, one report describes
weapons in the bag simply not equipping at all when interacted with.

## Expected behavior

- An item's equip restriction always matches what its tooltip states. If an item is
  class-restricted, the tooltip says so.
- Equipping a weapon or armor piece the player is eligible for from the bag succeeds
  every time.

## Scope

- Item equip eligibility checks (class/armor-type restrictions) for Fang of Korzul and
  Deathlord Warplate specifically
- Tooltip generation for class/type restrictions, to confirm it matches the actual gate
- Bag-to-equip flow for weapons generally

## Steps to reproduce

1. As a druid, attempt to equip Fang of Korzul from the bag.
2. As the class Deathlord Warplate is intended for, attempt to equip it from the bag.
3. Attempt to equip a weapon from the bag and observe whether it equips.

## Actual behavior

- Fang of Korzul is rejected for druid with no stated reason.
- Deathlord Warplate is rejected even though its tooltip does not indicate a class
  restriction.
- At least one report indicates bag weapons do not equip at all.

## Acceptance criteria

- Fang of Korzul and Deathlord Warplate equip successfully for every class their
  tooltip permits.
- Any genuine restriction is reflected in the item's tooltip text.
- Equipping a weapon from the bag works reliably.

PR #1903, opened 2026-07-14
fix(ui): never suppress the class-restriction tooltip line

jgyy:
## Summary
- The item tooltip's "Requires: <classes>" line was gated behind `!armorTypeForItem(item) && !weaponArchetypeForItem(item)`: it only rendered when the item had NEITHER a known armor weight NOR a known weapon archetype.
- Fang of Korzul (a rogue/hunter dagger) and Deathlord Warplate (a warrior/paladin/shaman mail chest) both resolve to a known archetype/armor-weight group, so their class restriction line was silently hidden. A player outside those classes got rejected on equip with no in-game explanation at all.
- Extracted the resolver into a pure module, `src/ui/item_class_restriction.ts` (`requiredClassesForTooltip`), so the armor-weight/weapon-archetype badges (which exist purely for equip-legality math) can never again suppress the class-restriction line (which exists purely to explain the restriction).

## Flow

```mermaid
flowchart TD
    A["Item tooltip render"] --> B{"item.requiredClass set?"}
    B -->|no| Z["no class line"]
    B -->|yes| C{"BEFORE: armorType or<br/>weaponArchetype also resolved?"}
    C -->|yes| D["class line suppressed<br/>(Fang of Korzul, Deathlord Warplate)"]
    C -->|no| E["class line shown"]
    B -->|yes, AFTER fix| F["requiredClassesForTooltip(item)"]
    F --> G["class line always shown<br/>when requiredClass is set"]

    style D fill:#822,color:#fff
    style G fill:#2d5,color:#000
```

## Test plan
- [x] Added `tests/item_class_restriction.test.ts` (4 tests): class-restricted item with a known armor type, with a known weapon archetype, with neither, and no restriction at all.
- [x] `npx vitest run tests/item_class_restriction.test.ts` -- 4 passed.
- [x] `npx @biomejs/biome check --write` on changed files -- clean (warnings only, no errors, pre-existing style in the surrounding file).

## Screenshots
Fang of Korzul (rogue/hunter dagger) tooltip, hovered in bags:

| Before | After |
| --- | --- |
| Before: class-restriction line missing | After: "Classes: Rogue, Hunter" always shown |

Part of #1893


Rubsey (commented):
The extraction itself is the right shape (pure resolver, hud.ts as a thin consumer, before/after screenshots), and for weapons the fix is genuinely correct: Fang of Korzul's restriction IS enforced (its `requiredClass` matches the rogue weapon archetype set exactly), and the base tooltip gave a blocked player no explanation at all.

But the blanket un-suppression is wrong for armor, and that is most of the affected items:

- **blocking** `src/ui/item_class_restriction.ts:16` (rendered at `src/ui/hud.ts:4599-4602`): for armor, the equip gate never reads `requiredClass`. `canEquipItem` short-circuits on armor weight (`src/sim/equipment_rules.ts:77-78`), so `requiredClass` on armor is loot-targeting metadata, not an enforced restriction. Always showing the line makes those tooltips claim restrictions that do not exist: `woven_robe` (cloth, 4-caster list, all nine classes can equip cloth), `shadow_jerkin` (leather, rogue/hunter, druids and every mail class can equip leather), the `HEAL_MAIL` paladin/shaman epics in `heroic_loot.ts` (warriors can equip them), and a dozen more delve pieces. The old condition was not an accident: it mirrors `canEquipItem`'s decision order exactly and was introduced deliberately (3d074c09f, "fix: hide derived equipment class locks"), so the tooltip only named classes when `requiredClass` was the rule actually enforced. Issue #1893's own acceptance criterion is that the tooltip always matches the gate; this PR fixes that for the two named items and breaks it for many others. Two honest ways out: scope the line to items where the listed classes match what the gate enforces (weapons whose set maps to an archetype, weapons with a literal `requiredClass`, and armor whose list equals its armor-weight class set, which covers Deathlord Warplate), or make `canEquipItem` enforce `requiredClass` on armor too, which is a sim behavior change that needs the maintainer's intent plus sim tests. The second matches classic behavior but is a bigger call than a tooltip PR should make silently.
- **should-fix** `tests/item_class_restriction.test.ts:17-31`: both pinned items are exactly the two where `requiredClass` coincides with the enforced set, so the suite cannot catch the divergent case above. Whatever direction the fix takes, add a pin for an armor item whose list is narrower than its weight group (for example `canEquipItem('druid', ITEMS.shadow_jerkin)` is true today while the tooltip would say rogue/hunter).
- **should-fix** PR body: "Fixes #1893" auto-closes an issue whose bag-equip report this diff does not touch. Suggest "Part of #1893".

jgyy:
Thanks Rubsey, this is exactly right and I appreciate the precise pointer to `equipment_rules.ts:77-78`.

Pushed a follow-up commit that scopes the tooltip line rather than un-suppressing it blanket-wide:

- Added `classesThatCanEquipArmorType(armorType)` to `src/sim/equipment_rules.ts`: it computes the full set of classes `canEquipItem` actually admits for a given armor weight (every class whose max armor rank is at least that weight's rank).
- `requiredClassesForTooltip` in `src/ui/item_class_restriction.ts` now branches on `armorTypeForItem(item)`. For weapons it keeps showing the line unconditionally, since `canEquipItem` always enforces a weapon's `requiredClass` (directly, or via the matching archetype). For armor, it only shows the line when `requiredClass` names exactly the classes `classesThatCanEquipArmorType` returns, so Deathlord Warplate (mail naming only warrior/paladin/shaman, which is exactly the mail group) still shows it, while `woven_robe`, `shadow_jerkin`, the `HEAL_MAIL` epics, and the other narrower-than-weight-group armor pieces go back to being silent, matching what `canEquipItem` actually enforces.
- Added the suggested test pin in `tests/item_class_restriction.test.ts`: `canEquipItem('druid', ITEMS.shadow_jerkin)` is `true`, and `requiredClassesForTooltip(ITEMS.shadow_jerkin)` is `null`, so the suite now catches the divergent case the old two pins couldn't.
- Updated the PR body from "Fixes #1893" to "Part of #1893" since this diff only covers the tooltip explanation, not the bag-equip report in that issue.

I deliberately did not touch `canEquipItem` to make `requiredClass` an enforced gate on armor; that is a real sim balance change (and would need to match classic behavior with maintainer sign-off and sim tests), not something to fold into a tooltip fix.

`npx vitest run tests/item_class_restriction.test.ts` and `npx tsc --noEmit` are both green.

ryan-foo (approved):
Approving. Real bug: weapon and mail items whose requiredClass maps to an archetype group (Fang of Korzul, Barrowlord Warplate) were hitting the old `!armorTypeForItem && !weaponArchetypeForItem` guard and hiding the Classes line entirely.

The fix is module-first (a new item_class_restriction pure core plus a sim helper, with hud.ts staying a thin consumer), reuses the existing itemUi.tooltip.classes key so there is no new i18n, and correctly still returns null for subset-restricted armor that other weight-eligible classes can equip, which avoids printing a false Classes claim. Tests cover the three branches.

PR #1906, opened 2026-07-14
fix(sim): correct stealth gating for Cheap Shot and Sap

### LINKED ISSUE #1890: Stealth-related rogue abilities gated incorrectly (Cheap Shot, Sap, Low Blow)
## Problem

Several rogue abilities that are supposed to require or interact with stealth are
gated incorrectly. Reports describe: Cheap Shot requiring stealth per its tooltip but
not appearing on the action bar while actually in stealth mode (making it unusable),
Sap removing the rogue's own furtive/stealth mode when used, and Low Blow not working
while invisible even though its description says it should.

## Expected behavior

- Abilities that require stealth remain usable (visible/enabled on the bar) while the
  rogue is in stealth.
- Sap does not break the caster's own stealth state.
- Low Blow functions while invisible, matching its tooltip description.

## Scope

- Rogue ability availability/gating logic (stealth-required abilities)
- Sap's interaction with the caster's own stealth aura
- Low Blow's invisibility check

## Steps to reproduce

1. As a rogue, enter Stealth.
2. Attempt to use Cheap Shot, Sap, and Low Blow.
3. Observe each ability's incorrect behavior relative to its tooltip.

## Actual behavior

- Cheap Shot does not show as usable on the action bar while in stealth.
- Using Sap removes the caster's own stealth mode.
- Low Blow fails to activate while invisible.

## Acceptance criteria

- Cheap Shot is usable while stealthed, matching its tooltip requirement.
- Sap no longer cancels the caster's own stealth.
- Low Blow works while invisible.
- Non-stealth ability behavior is unchanged.

## Notes

Reports came from both PvP and PvE contexts for at least one of these abilities (Sap),
so the fix should not be scoped to a single combat context.

jgyy:
## Summary

Three reported stealth-related rogue ability bugs, investigated independently:

1. **Cheap Shot not reflecting its stealth requirement (confirmed and fixed).** `action_bar_view.ts`'s `usable` computation never checked `ability.requiresStealth`, so the bar never distinguished a castable-in-stealth ability from one that would actually fail. Added a `stealthed` field to `ActionBarPlayerInput` and gated `usable` on it.
2. **Sap breaking the caster's own stealth (confirmed and fixed).** `preservesStealth()` only exempted stealth-toggle abilities and Sprint; every other cast, including Sap, called `breakStealth()`. Sap is the classic no-reveal opener (incapacitates at range, no melee swing) and must not blow the caster's own stealth. Added it to the exemption list.
3. **Low Blow (kidney_shot) not working while invisible (investigated, not reproducible).** Built a full combo-point setup, cast Vanish, then Low Blow: it lands correctly on the current code. No `requiresStealth` gate exists on it and nothing in the finisher path blocks casting while stealthed. Kept the test as a regression guard rather than claiming a fix for a bug I couldn't reproduce.

## Flow

```mermaid
flowchart TD
    A["Action bar usable check"] --> B{"requiresStealth?"}
    B -->|no| C["usable = resource sufficient"]
    B -->|"yes (Cheap Shot)"| D{"player.stealthed?"}
    D -->|"BEFORE: not checked at all"| E["usable never reflected stealth"]
    D -->|"AFTER: yes"| F["usable = resource sufficient"]
    D -->|"AFTER: no"| G["usable = false"]

    H["Ability cast (Sap)"] --> I{"preservesStealth(ability)?"}
    I -->|"BEFORE: only stealth-toggle / sprint"| J["Sap incorrectly calls breakStealth()"]
    I -->|"AFTER: + sap"| K["Sap no longer breaks stealth"]

    style E fill:#822,color:#fff
    style J fill:#822,color:#fff
    style F fill:#2d5,color:#000
    style G fill:#2d5,color:#000
    style K fill:#2d5,color:#000
```

## Test plan
- [x] `tests/sim.test.ts`: "Sap does not break the caster stealth (issue #1890)" and "Low Blow (kidney_shot) works while invisible from Vanish (issue #1890)".
- [x] `tests/action_bar_view.test.ts`: `requiresStealth` gates `usable` correctly; non-stealth abilities unaffected.
- [x] Updated `stealthed: false` fixtures in `tests/action_bar_painter.test.ts`, `tests/hud_perf_budget.test.ts`, `tests/mobile_action_ring_painter.test.ts` for the new required `ActionBarPlayerInput` field.
- [x] `npx vitest run tests/action_bar_view.test.ts tests/action_bar_painter.test.ts tests/hud_perf_budget.test.ts tests/mobile_action_ring_painter.test.ts` -- 76 passed, 3 skipped.
- [x] `npx vitest run tests/sim.test.ts -t "issue #1890"` -- 2 passed.
- [x] `npx @biomejs/biome check --write` on changed files -- clean (warnings only).

No screenshots: sim/action-bar-usability logic change (a slot's dim/enabled state), not a layout change; the mermaid above documents the corrected gating.

Fixes #1890

### DIFF
--- a/src/sim/combat/effect_dispatch.ts
+++ b/src/sim/combat/effect_dispatch.ts
@@ -45,7 +45,10 @@ function isStealthToggle(ability: AbilityDef): boolean {
 }
 
 function preservesStealth(ability: AbilityDef): boolean {
-  return isStealthToggle(ability) || ability.id === 'sprint';
+  // Sap is the classic no-reveal opener: it incapacitates from range without a
+  // melee swing, so unlike Cheap Shot/Ambush/Garrote it must not blow the
+  // caster's own stealth (issue #1890).
+  return isStealthToggle(ability) || ability.id === 'sprint' || ability.id === 'sap';
 }
 
 function consumeMatchingAura(
--- a/src/ui/action_bar_view.ts
+++ b/src/ui/action_bar_view.ts
@@ -125,6 +125,12 @@ export interface ActionBarPlayerInput {
   potionCdRemaining: number;
   queuedOnSwing: string | null;
   pos: Vec3;
+  /** Whether the player currently carries a `kind:'stealth'` aura (Stealth or
+   *  Vanish). Gates a `requiresStealth` ability's usable state (issue #1890):
+   *  without this the bar never dimmed Cheap Shot/Ambush/Garrote out of
+   *  stealth, so they looked equally "ready" whether or not the cast would
+   *  actually succeed. */
+  stealthed: boolean;
 }
 
 /** The target fields the bar reads; null when there is no current target. */
@@ -335,7 +341,8 @@ export function createActionBarView(
             : 0;
         slot.cdText = cd > COOLDOWN_TEXT_THRESHOLD ? deps.formatCount(Math.ceil(cd)) : '';
         slot.count = '';
-        slot.usable = !(player.resource < ability.cost);
+        slot.usable =
+          !(player.resource < ability.cost) && (!def.requiresStealth || player.stealthed);
         slot.outOfRange =
           def.requiresTarget &&
           tgtDist !== null &&
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -385,6 +385,7 @@ import {
   streamerActionPlatform,
   streamerMenuActions,
 } from './player_context_menu';
+import { playerStealthed } from './player_stealthed';
 import { hydratePortraits, portraitChipHtml } from './portrait_chip';
 import { maskProfanity } from './profanity';
 import { encodeItemLink, encodeQuestLink, parseChatSegments } from './quest_link';
@@ -7078,10 +7079,22 @@ export class Hud {
     // ActionBarPainter. Every per-slot icon / cooldown / dimming / count write
     // routes through the elided writer facet; the aria-label keeps its per-frame t()
     // call IN the core while the painter elides the DOM setAttribute (Top risk 4).
+    // Derive `stealthed` from the mirrored auras rather than trust the raw entity
+    // field: offline it is the live sim Entity's cache (kept current by
+    // Sim.updateAuras), but online it is the ClientWorld mirror's server-local
+    // interest-filtering cache, never encoded on the wire and never updated on the
+    // client (see src/net/online.ts, server/game.ts). The auras ARE mirrored, so
+    // this stays correct on both hosts. Shared by every action-bar-family view
+    // below (desktop bar, mobile ring, consumables quick bar).
+    const abPlayer = { ...p, stealthed: playerStealthed(p.auras) };
     this.renderPetBar();
     if (this.spellbookWindow.isOpen) this.spellbookWindow.tickOpen();
     this.actionBarPainter.paint(
-      this.actionBarView.tick({ player: p, target: target ?? null, inventory: sim.inventory }),
+      this.actionBarView.tick({
+        player: abPlayer,
+        target: target ?? null,
+        inventory: sim.inventory,
+      }),
     );
 
     // mobile action ring: the paged touch combat cluster, gated on the touch-mode
@@ -7091,7 +7104,7 @@ export class Hud {
     if (this.isMobileLayout() && this.mobileActionRingView && this.mobileActionRingPainter) {
       this.mobileActionRingPainter.paint(
         this.mobileActionRingView.tick({
-          player: p,
+          player: abPlayer,
           target: target ?? null,
           inventory: sim.inventory,
         }),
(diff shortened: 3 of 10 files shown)


Rubsey (commented):
The sim half is right: adding `sap` to `preservesStealth` (`src/sim/combat/effect_dispatch.ts:48-51`) is the correct classic behavior, it sits in the shared path so all three hosts get it, and the new sim test would fail on the old code at its final assertion. The kidney_shot investigation writeup (not reproducible, regression test kept) is exactly how to handle that. The four fixture re-mints are mechanical field additions, no assertions weakened.

The UI half has a blocker:

- **blocking** `src/ui/action_bar_view.ts:345`: the new gate reads `player.stealthed`, and the online client never populates that field. The hud feeds the view the raw IWorld player (`src/ui/hud.ts:7084`), which offline is the live sim entity (kept current by `updateAuras`), but online is the `ClientWorld` mirror entity created with `stealthed: false` (`src/net/online.ts:986`) and never updated: `stealthed` is a server-local cache used for interest filtering (`server/game.ts:923`) and is not encoded on the wire, and `applyWire` decodes auras but never sets it. So on the live server every `requiresStealth` ability (Cheap Shot, Ambush, Garrote, Sap, druid Rake/Pounce) renders permanently dimmed, including while actually stealthed. That inverts the bug this PR set out to fix, and it is exactly the acceptance criterion in #1890 (usable on the bar while stealthed, online). The auras ARE mirrored, so the clean fix is to derive the flag from them: either compute `stealthed` from `player.auras` (a `kind: 'stealth'` aura) where the input is assembled in hud.ts, or set `e.stealthed` in `applyWire` when decoding auras so the entity shapes stay honest on both hosts. Either way, please add a test that exercises the online-shaped input (an entity with a stealth aura but `stealthed: false`) so this parity cannot silently drift again; the current view tests pass `stealthed` directly, which is why they cannot catch it.

One note, not a finding: the title says `fix(sim)` but the Cheap Shot half is UI-only; the sim cast gate for `requiresStealth` already existed in the shared casting path, the bar just never mirrored it.

jgyy:
Thanks Rubsey, good catch. Fixed the online/offline stealthed parity.

Root cause matched exactly what you laid out: the ClientWorld mirror is built with `stealthed: false` (src/net/online.ts) and that field is never touched by applyWire, since it is a server-local interest-filtering cache (server/game.ts) that never goes over the wire. The offline Sim entity keeps it live via updateAuras, so the raw field only worked on one host.

Fix: added a small pure module, src/ui/player_stealthed.ts, exporting playerStealthed(auras), which checks for a kind: 'stealth' aura. This mirrors the same pattern hud.ts already used elsewhere (the druid/rogue hotbar-form logic at lines 4944 and 4950 does auras.some(a => a.kind === 'stealth')), so it is the third use of that check and earns its own module rather than a fourth inline copy.

I wired it in at the point hud.ts assembles the action-bar-family view input (src/ui/hud.ts, in update()): built one abPlayer = { ...p, stealthed: playerStealthed(p.auras) } and used it for all three call sites that feed an ActionBarWorldInput (the desktop action bar, the mobile action ring, and the consumables quick bar), instead of passing the raw player object with its stale field. Since auras ARE mirrored on the wire, this is correct on both hosts without touching IWorld or the wire protocol at all.

Added tests/player_stealthed.test.ts, which exercises exactly the online-shaped input you described: an aura list with a kind: 'stealth' entry while the entity's own stealthed field would be false, asserting the derived flag is still true, plus a false case and an order-independence case. I left the existing action_bar_view.test.ts cases as-is since they correctly test the view core in isolation; this new test targets the assembly step that was actually buggy.

Verified: npx tsc --noEmit clean, tests/architecture.test.ts and tests/world_api_parity.test.ts still pass (no IWorld surface changed), action_bar_view.test.ts and the new player_stealthed.test.ts pass, plus a spot check of hotbar_form_persistence, mobile_action_ring_painter, hud_perf_budget, and client_shell tests. Pushed to fix/stealth-ability-gating.

Left the sim half (sap in preservesStealth) untouched as you confirmed it is correct.

ryan-foo (approved):
Approving. The Sap-preserves-stealth fix is correct and determinism-safe: sap has only an incapacitate effect (no weaponStrike or directDamage) and requiresOutOfCombat, so no auto-attack restarts, and breakStealth draws zero rng, so skipping it does not shift the shared rng stream. Parity stays green with no golden regen. The online mirror is handled too, since auras carry `kind` on the wire so playerStealthed reads correctly on both hosts.

One note, not blocking: the title says "Cheap Shot and Sap," but the Cheap Shot part is a UI dimming fix in action_bar_view, not a sim gating change, since casting_lifecycle already blocked Cheap Shot out of stealth. Worth trimming the title so it does not read as a sim fix that was not needed.

PR #1907, opened 2026-07-14
fix(ui): preserve action bar state across relog and talent respec

### LINKED ISSUE #1889: Action bar loses saved state across relog and talent respec
## Problem

Multiple players report the action bar losing its saved state across common flows:
respec, relog, and normal play. Reported symptoms include: spell icons turning black
with the second bar missing, keybindings for a newly added hotbar not appearing on the
bar, the second hotbar showing 11 slots instead of 12, all buttons (including target and
party frames) disappearing with no way to interact, and consumables (heal/mana potions)
being silently removed from the bar whenever the player switches talent spec.

This looks like several related but distinct persistence and layout bugs in the action
bar system rather than one root cause, and needs investigation to split apart.

## Expected behavior

- Action bar contents persist correctly across logout/login.
- Switching talent spec does not remove unrelated item slots (potions) from the bar.
- A newly unlocked hotbar shows all of its slots (12, not 11) and respects configured
  keybindings immediately.
- The action bar, target frame, and party frames cannot end up in a state with no
  visible or clickable buttons.

## Scope

- Action bar persistence/save-load path
- Talent respec handling of action bar slots
- Hotbar slot count and keybinding application
- HUD frame visibility (action bar, target, party) failure modes

## Steps to reproduce

1. Log in, arrange spells and potions across two hotbars.
2. Log out and back in, or switch talent spec.
3. Observe missing spells, an all-black bar, a shrunk second hotbar, or a fully blank HUD.

## Actual behavior

- Spells/potions intermittently disappear or go black after relog or respec.
- Second hotbar sometimes renders with 11 slots instead of 12.
- Target/party frames can disappear alongside the action bar, leaving no way to
  interact with the UI.

## Acceptance criteria

- Action bar state (including consumables) survives relog and talent respec.
- The second hotbar always exposes its full 12 slots with working keybindings.
- No reachable state leaves the player with zero visible action bar or frame buttons.
- Desktop and mobile behavior is unchanged apart from the fix.

jgyy:
## Summary

Four reported action-bar persistence bugs, investigated independently:

1. **Talent respec clearing potion/food/drink slots (confirmed and fixed).** `applyLoadoutBar` rebuilt the WHOLE bar from a `SavedLoadout`'s ability ids on every respec. A loadout only ever records ability ids (item shortcuts are stripped before saving), so any slot the loadout left blank was nulled out, silently clearing a player's consumables. Extracted a pure `applyLoadoutBar(current, bar, slots, abilityExists)`: a loadout slot with a resolvable ability id fully replaces what was there; every other slot keeps its existing item shortcut.
2. **Second hotbar disappearing after relog (confirmed and fixed).** Persisted boolean settings (`showSecondaryActionBar` and friends) only applied their body CSS class from a live Options toggle. A fresh page load (login, relog, plain reload) left the class at its CSS default of hidden even when the stored setting was on, until the player re-flipped the checkbox in the new session. Added `settings_body_classes.ts`, re-derived once at boot.
3. **Second hotbar showing 11 slots instead of 12** and **all buttons (bar, target, party frames) disappearing entirely** -- investigated but could not reproduce a distinct root cause beyond (2) above (a hidden second bar reads as "missing slots"/"no buttons" depending on which chrome the player notices first). Not claiming a separate fix; flagging here in case reports continue after this PR ships, since that would confirm a genuinely separate bug.

## Flow

```mermaid
flowchart TD
    subgraph Respec["Talent respec"]
    A["SavedLoadout.bar (ability ids only)"] --> B{"applyLoadoutBar"}
    B -->|"BEFORE: replace whole bar"| C["potion/food slots nulled"]
    B -->|"AFTER: merge per-slot"| D["ability slots replaced,<br/>item slots kept"]
    end
    subgraph Boot["Page load / relog"]
    E["Persisted settings"] --> F{"resolveSettingsBodyClasses"}
    F -->|"BEFORE: only on live toggle"| G["show-actionbar2 stuck hidden"]
    F -->|"AFTER: re-derived at boot"| H["body classes match stored settings"]
    end

    style C fill:#822,color:#fff
    style G fill:#822,color:#fff
    style D fill:#2d5,color:#000
    style H fill:#2d5,color:#000
```

## Test plan
- [x] `tests/hotbar.test.ts`: `applyLoadoutBar` keeps existing item slots, replaces ability slots, ignores unresolvable ids.
- [x] `tests/settings_body_classes.test.ts`: resolves the class/enabled pairs correctly from a settings getter.
- [x] `npx vitest run tests/architecture.test.ts tests/hotbar.test.ts tests/settings_body_classes.test.ts` -- 58 passed.
- [x] `npx @biomejs/biome check --write` on changed files -- clean (warnings only).

## Screenshots

The second-action-bar-row fix restores an existing visual state (the CSS class was already
correct once applied; the bug was WHEN it got applied on a fresh load), so a before/after still
shows the actual difference a player sees: the second hotbar row present after this fix vs. the
single-row default without it.

Before (`showSecondaryActionBar` off / not yet re-applied):
docs/screenshots/action-bar-state-persistence/before-actionbar-secondary-row.png

After (`showSecondaryActionBar` stored and re-derived at boot, second row visible):
docs/screenshots/action-bar-state-persistence/after-actionbar-secondary-row.png

The loadout-bar fix (item slots preserved across a talent-loadout switch) is data-persistence,
not a new layout, and is covered by the `applyLoadoutBar` unit tests in `tests/hotbar.test.ts`.

Fixes #1889


Rubsey (commented):
The loadout-bar half of this is a good fix. The base code really did rebuild the whole bar from ability ids on every loadout apply (nulling consumable shortcuts the loadout never recorded), the extracted `applyLoadoutBar` is pure and lands in the existing hotbar core with the Hud as a thin consumer, and the tests pin the exact regression (an item shortcut surviving a blank loadout slot would fail on the old semantics). Stale ids restored by it are still pruned by the existing `syncHotbarActions` pass, so no ghost abilities either.

The settings-body-classes half does not hold up, though:

- **should-fix** `src/game/settings_body_classes.ts` + `src/main.ts:908-915`: the claimed root cause does not exist at base. `startGame` already re-applies every persisted setting on every fresh load: the apply-all loop at `src/main.ts:1714-1716` (head) runs `applySetting(k, saved[k])` for every key of `settings.all()` (def-filled for every bool key), and the `showSecondaryActionBar` branch toggles `body.show-actionbar2` unconditionally, same for the other four classes. The new boot loop re-applies the same values a few hundred lines earlier, so it is a behavioral no-op, and the module/test comments now document a bug that is not there. Since the PR says "Fixes #1889", merging as-is closes the relog symptom without anything actually changing; if players really see the second bar vanish after relog, the cause is elsewhere (worth checking: relog on a different browser/device, since all of this is per-browser localStorage). I would land the hotbar half and drop this half, or come back with a reproduced trace showing the apply-all loop not covering it.
- **nit** `src/game/settings_body_classes.ts:1`: if the module stays, it claims to be the single source of truth for the key/class pairing, but the five live handlers at `src/main.ts:1497-1520` (head) still hardcode the same class strings, and the table omits the two other settings-driven body classes handled the same way (`mobile-left-handed`, `mobile-camera-joystick-on`). Either wire the handlers through the table or drop the claim.
- **nit** `tests/settings_body_classes.test.ts:15-25`: both bulk assertions derive their expected values from `SETTINGS_BODY_CLASSES` itself, so they pin nothing beyond map-returns-its-input.

Nice touch reporting symptoms 3 and 4 of #1889 as not reproduced instead of claiming fixes for them.

jgyy:
Thanks for the thorough review, Rubsey.

I verified the should-fix point directly: the apply-all loop in startGame (the `for (const k of Object.keys(saved) ...) applySetting(k, saved[k])` call, now at src/main.ts:1716) already runs after the boot-loop code I had added at main.ts:908-915, and it calls applySetting for every settings.all() key, including showSecondaryActionBar, which toggles body.show-actionbar2 the same way. So the new boot loop was indeed a same-tick no-op duplicating existing behavior, not a fix for anything. You're right that if the relog symptom is real, the cause is elsewhere (per-browser localStorage is a good lead), and it should come back with a reproduced trace rather than ship as a guess.

I dropped the whole settings-body-classes half:
- removed src/game/settings_body_classes.ts
- removed its import and the boot-loop call in src/main.ts (908-917)
- removed tests/settings_body_classes.test.ts

That also resolves both nits, since the module, its class-table claim, and the self-referential test assertions are gone with it.

Kept the loadout-bar half as-is: applyLoadoutBar in src/ui/hotbar.ts, its wiring in src/ui/hud.ts, and tests/hotbar.test.ts, which you confirmed is a good fix for the consumable-slot-clearing bug on talent respec.

Ran npx tsc --noEmit, npx vitest run tests/hotbar.test.ts and tests/architecture.test.ts, and the pre-push QA floor, all green. Pushed to fix/action-bar-state-persistence.

ryan-foo (commented):
The code change here is correct and I would happily approve it on its own: the per-slot merge in applyLoadoutBar keeps an existing item slot, replaces a slot with a resolvable ability id, and otherwise nulls, which fixes the loadout-switch bar (#1889) without reviving a stale item. Tests cover the three branches.

But the PR is mislabeled, so I do not want to merge it as-is. The title and body say it preserves the action bar across relog, and the body describes a second fix (the secondary bar stuck hidden after relog, via a new src/ui/settings_body_classes.ts re-derived at boot, with a matching test). That file is not in the diff: gh shows only hotbar.ts, hud.ts, hotbar.test.ts and the two screenshots. The relog path (loadSlotMap/saveSlotMap, and the main.ts show-actionbar2 application) is untouched.

So either the boot-derivation commit got dropped and needs to be pushed, or the title and body should be trimmed to just the talent-loadout-switch fix. Please reconcile the two and I will merge.

PR #1911, opened 2026-07-14
fix(ui): add crafting launcher to the desktop micro-menu and mobile More tray

### LINKED ISSUE #1865: Crafting window has no on-screen button; only reachable via the T keybind
## Problem

The crafting window exists and is meant to be usable anywhere, anytime, with no vendor or NPC gate. The only way to open it, however, is the "T" keybind. There is no on-screen button anywhere, so a player who never presses "T" has no way to discover that crafting exists.

On desktop, the bottom-right micro-menu (`#side-buttons`) already has launcher buttons for Character, Spellbook, Talents, Quest Log, Book of Deeds, World Map, Bags, Arena, Vale Cup, Leaderboard, Emotes, and Friends & Guild, plus Music and Game Menu. Crafting is the one interface window with a keybind but no button in that rail.

On mobile the problem is worse. There is no keyboard, and the desktop micro-menu is hidden on touch, so window launchers live in the "More" tray (Character, Bags, Spellbook, Talents, Map, and so on). Crafting is not in the More tray either, which means the crafting window is completely unreachable on touch devices.

## Expected behavior

- Desktop: a Crafting button in the bottom-right micro-menu, alongside the other window buttons, showing its keybind (T) the same way its neighbours do.
- Mobile: a Crafting button in the "More" tray, alongside Character, Bags, Spellbook, Talents, and the rest.
- Both buttons toggle the existing crafting window (open if closed, close if open), exactly like the "T" key does today.
- The "T" keybind keeps working unchanged.

## Scope

- Desktop micro-menu: the `#side-buttons` markup in `index.html` and `play.html`, the `sideButtons` label/keybind table in `src/ui/hud.ts`, and the button click wiring in `src/main.ts`.
- Mobile "More" tray: the `#mobile-extra-grid` markup in `index.html` and `play.html`, and its click wiring.
- Icon: there is no crafting glyph in the `data-icon` set yet (an anvil or hammer would fit), so a new icon needs to be added.
- i18n: reuse the existing crafting title string where possible; any new label is English-only per the contributor workflow.

## Acceptance criteria

- Desktop: a Crafting button appears in the bottom-right micro-menu and toggles the crafting window; its keybind cap updates on rebind and language change like the other micro-menu buttons.
- Mobile (portrait and landscape): a Crafting button appears in the "More" tray and opens the crafting window, with a tap target of at least 40x40.
- Both buttons toggle the same window the "T" key toggles, and pressing "T" still works.
- The desktop experience is otherwise unchanged: the existing buttons keep their order and spacing, and no other layout regresses.

## Environment

- Platform: desktop and mobile web (portrait and landscape).
- Surface: in-game HUD (bottom-right micro-menu on desktop, the "More" tray on mobile).

## Notes

- Keybind: the `crafting` action defaults to `KeyT` in `src/game/keybinds.ts`.
- Only entry point today: `case 'crafting'` calling `hud.toggleCrafting()` in `src/main.ts`, dispatched from the keybind edge-action path. `Hud.toggleCrafting` / `Hud.openCrafting` live in `src/ui/hud.ts`. There is no NPC, gossip, or vendor path.
- This is presentation wiring only; it follows the existing button pattern and needs no simulation change.

Rubsey:
## Summary

Closes #1865. The crafting window was only reachable via the T keybind: desktop players who never pressed T had no way to discover crafting, and touch devices could not reach it at all. This adds an on-screen launcher on every screen size, both toggling the exact window the T key toggles.

Consistency across surfaces comes from four shared choices:
- Same icon: one new hand-authored anvil glyph (data-icon="crafting") used by both buttons. Hand-authored to match the newest launcher glyphs (ball, book), avoiding a game-icons.net CC BY attribution addition to CREDITS.md.
- Same label: both surfaces reuse the already-translated hudChrome.crafting.title key for title, aria, and the tray caption, so no catalog edit, no locale fill, and no generated-file regen. The release-tier i18n gate sees nothing new.
- Same placement: directly after Bags on both the desktop rail and the mobile tray (crafting consumes inventory materials, and the crafting window's onCraft already refreshes Bags).
- Same behavior: both buttons call the existing public Hud.toggleCrafting(). The desktop cap and aria-label track rebinds and language changes through the existing sideButtons loop; the mobile button gets the 40x40 tap floor and tray auto-close from the existing grid CSS and bindButton seam.

## Changes

- index.html / play.html: #mm-crafting in #side-buttons and #mobile-crafting in #mobile-extra-grid, matching each file's local idiom
- src/ui/ui_icons.ts: new 'crafting' anvil glyph (union + ICONS record)
- src/ui/hud.ts: click listener + sideButtons keycap-table row
- src/game/mobile_controls.ts: onCrafting callback member + bindButton
- src/main.ts: onCrafting wiring to hud.toggleCrafting()
- tests: new tests/crafting_launcher.test.ts source-guard suite (both-entry parity, dock order bag < crafting < arena, wiring pins, unchanged T dispatch), plus pins in client_shell.test.ts, a 40x40 case in target_size.browser.test.ts, and the mobile_controls.test.ts fixture member

No sim, server, CSS, keybind, or i18n-catalog changes.

## Test plan

- [x] npx tsc --noEmit clean
- [x] biome ci on the 10 changed files clean
- [x] npx vitest run tests/crafting_launcher.test.ts tests/client_shell.test.ts tests/deeds_window.test.ts tests/mobile_window_coverage.test.ts tests/keybinds.test.ts tests/mobile_controls.test.ts (251 tests green)
- [x] browser target-size suite green (13 tests, includes the new #mobile-crafting 40x40 case)
- [x] npm run gate PASS (all 11 steps)
- [x] live browser E2E: button renders after Bags with aria "Crafting (T)", keycap t, hydrated anvil SVG; clicking toggles #crafting-window open and closed; the 17-button More tray at 740x360 stays inside its overflow-y auto, max-height scroll container
- [x] qa-checklist agent verdict READY (0 blocking, 0 should-fix)


FernandoX7 (commented):
Good change, and the four "same X" choices in the description are all real: both surfaces go through the one public `Hud.toggleCrafting()`, both reuse the existing `hudChrome.crafting.title` key (so the catalog, the overlays, and the resolved-bundle hash all stay untouched), the desktop cap and aria track rebinds through the existing `sideButtons` table, and the tray tap gets its 40x40 floor and auto-close for free because `bindButton` already closes anything inside `#mobile-extra-controls`. Both entry HTMLs are covered, which is the usual `/play` miss. I drove the mobile half in a real browser at 740x360: the tray tap opens `#crafting-window` and closes the tray. `tests/crafting_launcher.test.ts` is 7/7 green.

One blocking issue, on the desktop rail.

### blocking

**`index.html:427`, `play.html:380` - the new rail button pushes the Daily Rewards chest off the top of the screen on a 1366x768 laptop.**

`#side-buttons` (`src/styles/hud.css:1486`) is a bottom-anchored flex column with no `max-height` and no overflow, so every button added grows it upward, and nothing rescues it: `--ui-scale` is a player setting (default 1), not an auto-fit. I measured it in the offline world at default settings, with and without `#mm-crafting`:

| viewport | rail height | rail top | Daily Rewards chest |
|---|---|---|---|
| 1366x700 base | 604px | +22 | fully on screen |
| 1366x700 head | 638px | -12 | clipped, sits on the zone label |
| 1280x660 base | 604px | -18 | 28px of it visible, still clickable |
| 1280x660 head | 638px | -52 | entirely above the viewport, zero clickable pixels |

The rail now needs 712px of viewport height to fit, up from 678. A maximized browser on a 1366x768 laptop gives about 660, so after this merges those players cannot claim daily rewards at all. At 1280x660, before and after:

before
after

The crowding is inherited, not something you caused: the rail already overlapped the minimap's right edge and the chest was already part-clipped at 660. But this is the button that tips it from "clipped" to "gone", so it needs handling here. The cheapest fix that keeps your placement is a height-based compaction on the `#side-buttons` rule (a shorter `.micro-btn` and a smaller gap under a `@media (max-height: ...)`), or a `max-height` plus `overflow-y: auto`. A pinned rail-height budget test would stop the next button from doing this again.

### nits

- **`src/ui/ui_icons.ts:162`** - the anvil path spans y 140 to 370 of the 512 viewBox, so it is 230 units tall where `book` is 350 and `ball` is 384. Rendered into the 18px `.micro-btn .ui-icon` box it reads noticeably smaller and lighter than the icons directly above and below it. The shape itself is good at size; it just wants rescaling to about y 90 to 420 to match the family's weight.
- **`play.html:1001`** - `#mobile-crafting` carries `type="button"` while no other `.mobile-btn` in play.html does (and your `#mm-crafting` at `play.html:380` correctly omits it). The opposite of the local idiom you were matching elsewhere. Harmless, no form ancestor.
- **`tests/crafting_launcher.test.ts:39-42, 62-66`** - the suite pins source strings, so nothing actually dispatches a click on `#mm-crafting` and asserts the window opened, and nothing runs `hydrateIcons` over the button to prove the glyph materializes (an unregistered `data-icon` fails silently). The source-guard pattern is right for the entry-HTML parity half, but `deeds_window.test.ts` pairs it with a real jsdom behavior test, and this one would benefit from the same. Related: `expect(chrome).toContain("title: 'Crafting',")` at `:84` would pass on any catalog object carrying that line, so it does not actually pin the `hudChrome.crafting.title` path.
- The PR is visual and has no screenshots. CLAUDE.md asks for before/after on visual changes (desktop and mobile), and a desktop capture is exactly what would have caught the rail issue above.

### note, not this PR's to fix

`openCrafting` (`src/ui/hud.ts:12123`) has no `...this.windowFocus('#crafting-window')` and `crafting_window.ts` never calls `markDialogRoot`, while its peers do (`hud.ts:3730` talents, `:3892` deeds, `:3984` char). The two deliberate opt-outs, bags and bank, each carry an explicit "NOT windowFocus, which would install a Tab trap" comment for their drag and drop; crafting has neither the trap nor the comment. That predates you and this PR does not change the focus behavior (T opened it the same way), but you are adding its first pointer and Tab reachable opener, so it is a natural moment to add the two lines if you want to.

Nothing else. No sim, wire, server, or CSS changes, the glyph is hand-authored so CREDITS stays as is, reusing the existing key means the i18n gate sees nothing new, and the T dispatch is untouched.

Rubsey:
Thanks for the thorough review, all findings addressed at ee0dc6a46:

**blocking (rail pushes the Daily Rewards chest off-screen)** - fixed with a height-based compaction rather than overflow (which would have clipped the leftward hover flyout and hidden launchers). Inside `@layer components`, a `@media (max-height: 720px)` block tightens the rail (`.micro-btn` height 30 to 25, gap 4 to 3), so every launcher including the chest stays on screen and clickable on a maximized 1366x768 laptop. I also added the rail-height budget test you suggested: it counts the visible `.micro-btn` in both entries, models the compacted rail height against the 660px budget, and fails if a future button would push it over. Worth noting the base moved under this (a new micro-btn landed), so the rail is now 16 visible buttons; compacted it computes to 650px, inside the budget with 10px of headroom, and the test now guards that ceiling.

**nit, anvil glyph too small** - rescaled the path vertically to y 91 to 419 (h328) so it carries the same weight as `book` and `ball`. Silhouette is unchanged; I validated it by rasterizing the polygon before and after.

**nit, stray type=button** - removed from `#mobile-crafting` in play.html to match its tray neighbors (`#mobile-bags`, `#mobile-spellbook`). Left index.html as is, since its entire `#mobile-extra-grid` uses `type="button"` uniformly, so there the attribute is the local idiom.

**nit, source-only tests** - added a jsdom behavior test that dispatches a real click on `#mm-crafting` and runs `hydrateIcons` to prove the glyph materializes (an unregistered data-icon would leave no svg), and tightened the weak `title: 'Crafting'` check to a scoped regex that pins the `hudChrome.crafting.title` path specifically.

**note, screenshots** - the concern behind it (a desktop capture would have caught the rail issue) is now covered by the compaction plus the budget test guard; the offline headless capture renders the HUD at 0x0 here so it would not have surfaced it regardless.

**note, openCrafting windowFocus/markDialogRoot** - left as is per your out-of-scope call; this PR does not change the focus behavior.

On the base merge: the only conflict was the additive `UiIconName` union (this branch added `crafting`, base added `trash`); kept both. i18n regenerated. Full PR gate is green on CI.

PR #1912, opened 2026-07-14
feat(tanks): defensive cooldowns for warrior/paladin/druid + druid tank parity

ryan-foo:
## Summary

Adds one mechanically distinct level-20 defensive cooldown to each existing tank class and closes the druid tank parity gap.

- Warrior Ironhold: reduces all incoming damage by 40% for 8 sec, including sourceless hazards and DoT ticks whose caster has despawned. This is an emergency cooldown with a 180 sec cooldown.
- Paladin Sacred Bulwark: for 10 sec, the next lethal enemy hit is denied and restores the paladin to 35% health. The ward is single-use, has a 180 sec cooldown, ignores friendly and sourceless damage, and works in duels, ranked arena, Fiesta, and Protect Yumi.
- Druid Primal Reflexes: grants exactly +50% dodge for 6 sec on a 60 sec cooldown and remains usable in Dire Bruin form.

Dire Bruin also moves from +10% threat / +12% armor to +20% threat / +15% armor. This closes the bear tank threat gap without changing the global bear-form multiplier.

## Review fixes

- Guardian Ward no longer returns early from `dealDamage`. A saved lethal hit now continues through combat entry, counters, deed credit, CC and stealth breaking, eating/drinking/sitting cancellation, pushback, rage, and leash bookkeeping.
- Lethal overkill is clamped to the health actually lost before damage events and deed credit.
- The 35% restore is stored on the aura value and consumed by both combat logic and tooltip rendering.
- Ironhold consistently covers all damage sources. Sacred Bulwark consistently requires an enemy source.
- All three abilities now have explicit, unique icon recipes. Ironhold and Sacred Bulwark have aura-effect descriptors.
- Added explicit M16 translations for the updated Sacred Bulwark copy and both new aura-effect lines, then regenerated the current i18n artifacts.
- Merged the current `release/v0.26.0` tip and resolved the generated-artifact conflict using the release branch's current generation contract.
- Added exact screenshot automation that verifies the requested spell row, action-bar or mobile-ring control, active aura, painted buff, and armed cooldown before capturing.

## RL protocol contract

The druid remains the widest class kit and grows from 33 to 34 ability slots. This changes `NUM_ACTIONS` from 46 to 47 and `obsSize()` from 292 to 294. The trailing interact, stop, and eat/drink action IDs shift from 43/44/45 to 44/45/46, and all observation fields after the ability block shift by two positions. The Python client reads live metadata, but fixed-shape saved policies require migration or retraining.

## Testing

- 852 focused combat, cooldown, icon, tooltip, aura, PvP-mode, parity, protocol, and progression tests passed; 4 skipped.
- TypeScript, architecture, i18n freshness/coverage, localization, world API parity, scoped Biome, malware scan, and screenshot assertions passed.
- The complete Node 22 gate reached 13,965 passing tests with 18 skipped. Four local failures remain in three files byte-identical to `release/v0.26.0`: the container CPU inspector fixture and SFX export fixtures.
- Independent read-only reviews covered combat correctness, determinism, dead code, test coverage, M16 localization, desktop/mobile visuals, balance cadence, and the RL protocol change.

## Screenshots

| Surface | Before | After |
| --- | --- | --- |
| Warrior desktop | Warrior before | Warrior after |
| Paladin desktop | Paladin before | Paladin after |
| Druid desktop | Druid before | Druid after |
| Paladin mobile landscape | Paladin mobile before | Paladin mobile after |


jgyy (commented):
Nice, well-scoped addition. Ironhold, Sacred Bulwark, and Primal Reflexes all ride existing `Aura`/`AuraKind` infrastructure rather than adding new Entity/wire state, so there's no `world_api` parity gap to chase down. The two new conditional blocks in `dealDamage` (shield_wall, guardian_ward) match the file's existing idiom for per-mechanic aura checks (Expose, Defensive Stance, etc.) and are ordered correctly relative to absorb-shield resolution and the other death-adjacent checks, so I don't think this is coordinator creep, just following the pattern that's already there.

`tests/tank_defensive_cds.test.ts` covers the interesting cases with decisive assertions: exact HP restore on Sacred Bulwark, the non-lethal negative case, dodge math and form-usability for Primal Reflexes, and a pin for the Dire Bruin armor/threat bump. Ran it in an isolated worktree, all 6 pass.

Thanks for rounding out tank parity here, druid tanking especially has needed this.

EnriqueGF:
Thanks for this, and for the thorough writeup :) The three cooldowns are mechanically distinct, they ride the existing `Aura`/`AuraKind` infrastructure with no new wire or `IWorld` surface, and the druid parity reasoning is sound. Base branch is correct (`release/v0.26.0`), determinism is intact (both new branches draw zero rng), and `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/parity` and `npx tsc --noEmit` are all green on my side.

That said, I cannot approve this yet: the branch is currently red on `npm test`, and the `guardian_ward` interception has a set of correctness problems. Requesting changes.

## Blocking

**1. Three tests fail on this branch. The PR does not pass CI.**

`npx vitest run tests/ability_icons.test.ts tests/ability_tooltip_consistency.test.ts`

- `tests/ability_icons.test.ts > every ability has an explicit (non-fallback) icon recipe`: "abilities relying on the procedural fallback: ironhold, sacred_bulwark, primal_reflexes". The three new abilities have no icon recipe, so they take the fallback in the spellbook and on the action bar.
- `tests/ability_icons.test.ts > no two abilities resolve to an identical icon`: "sacred_bulwark = consecration".
- `tests/ability_tooltip_consistency.test.ts > hardcoded numbers hold at every rank`: `paladin.sacred_bulwark` cites `35` in its description but nothing in the resolved effect produces 35 (the 35% lives only as a literal in `damage.ts`), so the guard rejects the hardcoded number.

Please add the icon recipes and either make the 35% a real effect value or use a placeholder, then rerun the full suite.

**2. Merge state is CONFLICTING with the base.** The branch is 16 commits behind `release/v0.26.0` and conflicts on `src/ui/i18n.resolved.sha256` and `src/ui/i18n.status.summary.json`. Rebase on the current release tip and regenerate the i18n artifacts.

**3. `src/sim/combat/damage.ts:415-441`: the Guardian Ward early `return` skips the whole tail of `dealDamage`.** The sibling early returns above it (duel, fiesta, yumi, ranked arena) are terminal, state-ending paths, so skipping the tail is coherent there. The ward path is not terminal: the player survives and the fight continues, so all of that bookkeeping still has to run. Verified empirically (paladin at full HP, out of combat, takes a lethal blow from a mob while warded): `inCombat` stays `false` and `counters.damageTaken` increments by `0`. Also skipped for that blow:

- `breaksOnDamage` aura removal: a saved player keeps a CC that should have broken.
- `ctx.breakStealth(target)` and `ctx.breakStealth(source)`: a rogue opener into a warded paladin does not break stealth.
- `ctx.enterCombat(source, target)` and `ctx.refreshMobLeashFromAction(...)`.
- eat/drink/sit cancel: the warded player keeps eating through a would-be lethal hit.
- spell pushback, rage from dealing/taking, and the session counters.

Suggested shape: do not `return`. Set a flag, write `target.hp = restored` in place of the `Math.max(0, target.hp - amount)` write, and skip only the `if (target.hp <= 0)` death block, letting the rest of the function run as usual.

## Should fix

**4. `damage.ts:426-438`: the ward emits an unclamped `amount`.** Every sibling terminal arm clamps first (`amount = Math.max(0, target.hp)`). The ward emits the raw incoming amount, which by construction exceeds the health actually lost, so the combat log and FCT show an inflated hit, and `deedsMod.onDamageDealtForDeeds` is fed the same inflated number, which inflates lifetime damage deeds. Clamp before the emit and the deeds call.

**5. The two new cooldowns disagree about sourceless damage, and both disagree with their tooltips.**

- Ironhold is gated on `source && source.id !== target.id` (`damage.ts:124`), so it does not reduce sourceless damage: falling (`src/sim/player_motion.ts:286`), Blackwater (`src/sim/delves/runs.ts:1187`), Bell Shock and Egg-Sac Burst (`src/sim/delves/drowned_litany_rooms.ts:108,138`) all pass `source = null`. Verified: with Ironhold up, a 100-point `dealDamage(null, ...)` still costs the full 100. The text promises "reducing all damage taken by 40%".
- Sacred Bulwark has no source guard, so it does fire on fall or delve-hazard damage, burning a 60 second cooldown on a fall.
- Sacred Bulwark also sits below the ranked arena, fiesta and yumi arms, so the paladin's signature cooldown is a silent no-op in all three ranked PvP modes (the player is eliminated with the ward still on them). If it should save in arena, hoist it above those arms next to the absorb resolution. If it is deliberately PvE-only, say so in the description and pin it with a test.

**6. Guardian Ward has 100% uptime by construction.** A 60 second duration on a 60 second cooldown, in the base kit at level 20, means a permanently available cheat-death with a free reset to 35% health every minute. Please confirm that is intended: the repo invariant is that gameplay math follows classic-era formulas rather than invented numbers, and the classic analogues for a cheat-death and for a 40% mitigation wall are long-cooldown or talent-gated effects, not always-up base-kit ones.

**7. `src/ui/aura_effect.ts`: no descriptor for `shield_wall` or `guardian_ward`.** Comparable buffs (`defensive_stance`, `form_bear`) have one, so the two new buffs are the exception and their aura tooltips will show name and remaining time only, with no effect line. The switch has a `default: return null`, so `tsc` stays green and nothing catches this.

**8. Test coverage in `tests/tank_defensive_cds.test.ts` is thinner than the claims.**

- The Ironhold test is titled "... then expires" but never asserts expiry: a duration of 800 instead of 8 would still pass.
- The cheat-death is never shown to be single-use. Nothing in the suite distinguishes "saved once" from "warded players cannot die"; a second lethal blow after the save should assert `p.dead === true`.
- `expect(p.dodgeChance).toBeGreaterThan(baseDodge + 0.45)` is weakened where the exact value is knowable (`recalcPlayerStats` folds `buff_dodge` in additively): a tune to 0.46 passes. Assert the exact delta.
- The "DoT ticks included" case only flips the `direct` argument on a direct `dealDamage` call, and the mitigation branch does not read `direct`, so it is close to a duplicate of the previous assertion. Please apply a real DoT aura and assert a tick lands mitigated. Related: a DoT whose caster despawned resolves `source` to `null` (`src/sim/combat/auras.ts`), which per finding 5 would not be mitigated at all.
- Nothing pins the tuning the PR headline promises (cooldown 60, `offGcd`, durations, costs), nor the ward against any of the four PvP arms, nor the tail-of-function effects in finding 3.
- The bear-form arm asserts only that the aura exists, not that `dodgeChance` actually rose in form, which is the interesting half of the interaction.

## Minor

**9. The RL action space grows.** `ABILITY_SLOTS` in `src/sim/obs.ts` is derived from the widest class kit, and druid (already the widest) gains an ability, so `NUM_ACTIONS` and `obsSize()` both increase by one. `tests/env_protocol.test.ts` computes this dynamically so it stays green, but it is a contract change for `headless/env_server.ts` and `python/`: any trained policy with a fixed action head breaks. Worth a line in the PR body.

**10. Screenshots.** New abilities surface in the spellbook, the action bar and the buff bar, so the template's before/after requirement applies (desktop and mobile). The icon test failure in finding 1 is the same gap seen from the code side.

**11. `sacred_bulwark` carries `value: 0`** while the 35% restore is a literal in `damage.ts`. Putting the fraction on the aura value would let the tooltip and the code read from one source and would also address the tooltip-consistency failure.

## What I verified green

`npx tsc --noEmit`; `tests/architecture.test.ts`; `tests/localization_fixes.test.ts` (S3); `tests/world_api_parity.test.ts`; `tests/parity` (byte-identical, digest not regenerated); `tests/tank_defensive_cds.test.ts`, `tests/combat.test.ts`, `tests/talents.test.ts`, `tests/progression.test.ts`, `tests/guide.test.ts`, `tests/snapshots.test.ts`, `tests/deeds_content.test.ts`; biome on the changed files (warnings only); base branch `release/v0.26.0`; the aura wire already carries `kind`, so the new kinds reach `ClientWorld` with no parity gap, and both classify as buffs on the HUD.

Happy to re-review once the suite is green and the ward path falls through instead of returning.

PR #1913, opened 2026-07-14
fix(combat): taunts always land (never miss or resist)

ryan-foo:
## Problem

A taunt that misses or resists silently breaks tanking: the mob keeps hitting a healer or DPS while the tank thinks it grabbed aggro. The paladin taunt (Sacred Goad) is holy-school, so it travelled as a spell bolt and rolled a full resist on impact, so it could quietly fail. Warrior Goad and druid Menace are physical and already never rolled.

## Fix

Any ability whose effects include a `taunt` skips the on-impact resist roll (`src/sim/combat/casting_lifecycle.ts`), so taunts always apply regardless of the caster/target level gap. Physical taunts are unchanged (they resolve instantly with no roll); this only removes the spell resist for spell-school taunts like Sacred Goad.

## Testing

`tests/threat.test.ts` adds a case: across 40 seeds, casting Sacred Goad at a +3 mob (where a spell would resist a large fraction of the time) lands the taunt every time and never emits a `resist`. Parity is green (no golden casts a spell-taunt, so no draw-order regen), architecture and the S3 i18n guard pass, `tsc` clean.

### DIFF
--- a/src/sim/combat/casting_lifecycle.ts
+++ b/src/sim/combat/casting_lifecycle.ts
@@ -843,8 +843,12 @@ function applyAbility(ctx: SimContext, p: Entity, meta: PlayerMeta, res: Resolve
     // like a physical attack; a target can only fully RESIST them (classic-era
     // semantics), so a spell's on-impact roll uses isSpellResisted and emits a 'resist'.
     // A physical shot has no resist roll; its hit/crit resolve inside runEffects.
+    // Taunts (e.g. Sacred Goad) ALWAYS land: a resisted taunt would silently break
+    // tanking, so a taunt ability skips the resist roll entirely (physical taunts like
+    // Goad / Menace already never roll, since they resolve instantly below).
+    const isTaunt = res.effects.some((eff) => eff.type === 'taunt');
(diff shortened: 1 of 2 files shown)


jgyy (commented):
Clean fix, correctly scoped. The `isTaunt` gate keys off the effect list (`res.effects.some(eff => eff.type === 'taunt')`), not off ability id or school broadly, so it only changes behavior for Sacred Goad (the paladin's holy-school taunt); the two physical taunts (Goad, Menace) already never rolled a resist, so this doesn't accidentally suppress resists anywhere else.

Test is a good regression pin: 40 seeds, mob level +3 over the tank, asserting no resist event and correct final aggro target. Ran `npx vitest run tests/threat.test.ts`, 73/73 pass.

No blocking or should-fix findings. One nit: the inline comment could be a touch clearer that Sacred Goad (nicknamed "Goad" in-game) is the one taunt this branch actually touches, since the other two "Goad"/"Menace" mentions are physical and already skipped resists before this change. Not worth holding up the merge for.

Thanks for the fix, and for adding a real regression test alongside it.

Issue #1857, opened 2026-07-13
Release v0.26.0


Rubsey:
### Target version

0.26.0

### Release branch

`release/v0.26.0`, cut from `main` at `21afffab7901ec63c82fe1918be84a4366ab1547`.

### Bump the version

- [ ] Ran `npm version <X.Y.Z>` (or `npm version minor`/`patch`/`major`) and pushed the resulting commit and tag.
- [ ] Confirmed all three manifests show the new version and the native build numbers incremented.

### Verify

- [ ] `npm test` passes.
- [ ] `npm run build` succeeds.
- [ ] Ran the release malware-audit gate.

### Ship

- [ ] Web deploy is live (see `DEPLOY.md`).
- [ ] Mac desktop Electron app release artifact is published.
- [ ] iOS archive built via `npm run native:open:ios` and submitted to App Store Connect.
- [ ] Android bundle built via `npm run native:open:android` and submitted to Play Console.
- [ ] Release notes added under `docs/release-notes/`.

### Carryover from v0.25.0

Carried over from the v0.25.0 release tracker (#1777): every issue and PR referenced in its body or comments that is still open, plus every other open PR still targeting the retired `release/v0.25.0` branch.

This audit found 79 open PRs and 14 open issues. Merged and closed items are omitted. PRs are grouped by their current base so stale targets are visible; retarget or rebase each one only after checking its current diff against `release/v0.26.0`.

Not part of this migration: the separate v0.24.0 PTR track (#1744 and PRs based on `release/v0.24.0-ptr`) continues separately.

#### Open PRs targeting `release/v0.25.0` (need retargeting to `release/v0.26.0`)

- [ ] #1382 fix(server): decode request bodies once to preserve split UTF-8
- [ ] #1385 fix(admin): require a full-scope token for the admin API
- [ ] #1454 feat(glitch): launch shared MMO realm and reconcile v0.25
- [ ] #1504 feat(demon-hunter): add playable class
- [ ] #1547 feat(input): fix gamepad camera-follow snap-back, add auto-follow + invert-axis options
- [ ] #1548 feat(ui): controller onboarding overlay on first connect
- [ ] #1549 feat(input): camera-relative left-stick steering for controller
- [ ] #1556 feat(content): add Deepfen Shallows reed dressing
- [ ] #1573 fix(sim): give the fishing-dock plank deck collision
- [ ] #1584 WIP: Procedural infinite dungeons (Rifts) + dev commands
[...]

Issue #1235, opened 2026-07-01
Add an option to show your own name above your character


FernandoX7:
## Problem

Your own character is the only one in the world with no overhead label. Every other player and NPC shows a name (and players show an HP bar), but the local player's nameplate is hidden, so you cannot see your own name above your own character. This is intentional today: the nameplate core hides the local player unless they have an active overhead emote or raid marker (`src/render/nameplate_view.ts:104-105`, the `isSelf && !hasOverheadEmote` rule), and the painter forces the name and HP display off for self (`src/render/nameplate_painter.ts:190-191`).

Hiding your own nameplate is a reasonable default (it reduces clutter right where the camera is focused), but there is currently no way to turn it back on. Some players want to see their own name overhead: for screenshots, streaming, or just orienting themselves in a crowd.

## Prior art (how other MMORPGs handle this)

- FFXIV exposes an explicit toggle: Character Configuration, Display Name Settings lets you choose whether your own name displays overhead (also the `/nameplatedisp` text command). Showing your own name is a first-class, user-controlled option.
- WoW supports personal and friendly nameplates, and in recent versions lets you keep your name visible while hiding the bulky bar.
- On the "click yourself" idea: the MMORPG-standard for self-targeting is a dedicated keybind (FFXIV's F1, WoW's Alt self-cast modifier), not clicking your own 3D model. Clicking a world model is normally reserved for other units; self is a special case bound to a key or the player unit frame. In this codebase you can already click your own character to target yourself (`src/game/interactions.ts:91`), but you cannot attack yourself (`src/game/interactions.ts:48-56`) and there is no dedicated "target self" keybind.

## Expected behavior

- A user-facing setting (default off, preserving today's behavior) that, when enabled, shows the local player's own name above their character.
- When enabled, the own-name label matches the visual style of other players' name labels (font, color, developer badge and outline where applicable).
- The setting is a local display preference only: it does not change targeting, combat, or anything the server resolves.
- (Optional, secondary) A dedicated "target self" keybind (F1-style) in the Interface keybind category, so self-targeting does not depend on clicking your own model.

## Scope

- Nameplate visibility core: `src/render/nameplate_view.ts` (the `isSelf && !hasOverheadEmote` hide rule near line 104) and the painter's self-hide at `src/render/nameplate_painter.ts:190-191`. Thread a "show own name" preference through so self is shown when the option is on.
[...]

PR #1914, opened 2026-07-14
feat(game): show own nameplate by default

maxpolaczuk:
## Summary

- Enable Show My Nameplate by default when no saved value exists.
- Preserve an existing saved off choice.
- Make Discord avatar and linked-account flair immediately visible to the player.

## Testing

- `npx vitest run tests/settings.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts`
- `npm run test:browser`
- `npm run check:types`
- `npm run build:env`
- `npm run build:server`
- `npm run build`
- Pre-push QA floor

Release tracker: #1857

### DIFF
--- a/src/game/settings.ts
+++ b/src/game/settings.ts
@@ -207,11 +207,11 @@ export const BOOL_SETTINGS = {
   // Purely a local display preference: the badge is still earned and broadcast
   // either way, this only controls whether THIS client renders it.
   showDevBadges: { def: true },
-  // off by default (the classic self-view keeps no plate over your head): when
-  // on, render your OWN overhead nameplate (name, level, guild, hp, $WOC holder
-  // tier, dev badge, linked-Discord PFP) exactly as other players see it, so you
-  // can see how your character presents. Purely a local display preference.
-  showOwnNameplate: { def: false },
+  // on by default: render your OWN overhead nameplate (name, level, guild, hp,
+  // $WOC holder tier, dev badge, linked-Discord PFP) exactly as other players see
+  // it, so Discord linking and other flair changes have immediate visual feedback.
+  // Purely a local display preference; players can turn it off for the classic view.
+  showOwnNameplate: { def: true },
   // off by default: invert the vertical axis of mouselook (push mouse forward
   // to look down), the classic flight-sim preference.
   invertLookY: { def: false },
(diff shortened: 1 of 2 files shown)


Rubsey (commented):
The code does what it says for a fresh state and the test pins both halves properly: fresh default true, and an explicitly persisted off survives across instances over the shared localStorage shim. The motivation is also real: with the toggle on, the own plate renders the full flair set (role tag and tint, dev badge, holder tier, Discord PFP, deed title) exactly as other players see it (src/render/nameplate_painter.ts:201-244).

Two things to settle before merge:

- should-fix: the flip reaches far fewer existing players than the summary implies. save() serializes the whole def-filled values object on every set() (src/game/settings.ts:325-331), and the first-run graphics probe persists conclusively recognized devices on their first boot (src/main.ts:927-931), so every client that wrote any setting after the key shipped on 2026-07-02 (an options change, a quest or deed tracker collapse, a first boot on a recognized device) has showOwnNameplate: false baked into woc_settings, indistinguishable from a deliberate off. Those recently active players, the ones most likely to be linking Discord, silently keep the plate hidden; the new default reaches fresh installs and blobs untouched since before 2026-07-02. Either scope the summary to that reach, or this wants a one-time migration marker (the graphicsDefaultApplied pattern), which has its own cost: it would also override players who deliberately turned the plate off.
- should-fix: this reverses a deliberate decision without saying so. The setting shipped on by default (213aea06f) and was flipped off the next day per the issue #1235 contract (afcc03981): the issue asked for an opt-in option, and off is also the classic no-self-plate view the settings comment used to cite. Flipping it back is a product call, not a polish fix; the body should name the revert and why the original contract no longer applies. Related: combined with #1915, a SHILL-role player sees [SHILL] over their own head by default, opt-out only.

Nits:

- Three stale comments still say the setting defaults off: src/render/renderer.ts:822, src/render/nameplate_view.ts:88, scripts/deeds_screenshots.mjs:268. The PR rewrote only the settings.ts comment.
- scripts/threat_nameplate_shot.mjs:60-61: the fallback clip selector grabs the first .nameplate with an hp bar; with the default on, a fresh profile renders the self plate too, so the pre-aggro before shot can frame the wrong plate. Manual dev script, not CI, so just a heads-up.

Checked and fine: no test or script asserts the absence of a self plate (nameplate_view.test.ts passes the flag as a parameter, deeds_screenshots.mjs force-seeds the toggle on), and the per-frame cost of the extra plate is negligible since every write behind it is signature-elided.

PR #1917, opened 2026-07-14
fix(game): reland the camera follow release hold (mouselook release settle stutter)

Rubsey:
## Problem

In the default control mode, releasing a right-mouse mouselook drag makes the camera visibly stutter as it settles: it keeps rotating briefly past where the player aimed, then stops dead (or gets dragged back while moving). Mouse Camera mode does not show it.

## Root cause

While the drag is live, the facing channel streams camYaw to the server and the mirrored, render-interpolated facing lags that stream by the measured input echo plus roughly one snapshot interval online (up to one 50 ms tick offline). The instant the drag releases, updateFollowCameraYaw re-engages and its rigid follow term (targetYaw += wrapAngle(interpFacing - lastInterpFacing)) replays the still-arriving echo of the drag straight into camYaw: the camera overshoots past the chosen heading by the echo residual. A frame-by-frame numeric repro on the pure module measured a 0.30 rad (17 deg) post-release kick for a 3 rad/s drag at a 100 ms echo, and 0.15 rad for the offline one-tick analogue.

Mouse Camera mode escapes it because a pure left/right-click camera drag never streams facing (the facing override there requires a movement key), so the mirror never moves and the same follow math computes a zero delta (repro: exactly 0 post-release motion). That is the whole mode asymmetry in the report.

## This is a regression, and this PR is a scoped reland

The exact fix already landed as 68fe706f7 (2026-07-10, "fix(game): hold camera follow through the touch-look release echo (no shake)", also covering touch swipe-look and the camera joystick) but rode inside the #1736 interface-overhaul branch and was deleted as unflagged collateral by the #1788 revert (a45d20eb3): 87 lines of camera_follow.ts and 262 test lines removed with no mention in the revert description.

Re-landed here standalone onto release/v0.26.0:

- src/game/camera_follow.ts: stepCameraReleaseHold, a pure release hold that latches the final camYaw on the falling edge of camera ownership and keeps follow disengaged (via the cameraDriven flag, which still advances lastInterpFacing so re-coupling never snaps) until the interpolated facing converges within 0.02 rad, a manual keyboard turn takes the heading, or an echo-scaled grace lapses (max(350, echo*1.5+120) ms, the same formula keyboard_turn_facing already ships).
- src/main.ts: wires the hold into updateCamera for both the offline and online paths, plumbing the measured input echo online.
- tests/camera_follow.test.ts: the restored 30/60/144 fps release-trace tests (pre-fix mechanism pin, no overshoot, no direction reversal, frame-rate-independent settling).

camera_follow.ts and the test file are byte-identical to the original commit; the main.ts wiring is byte-identical in body (only surrounding context drifted).

Overlap note: PR #1815 (the full #1736 restore, currently targeting v0.25.0) carries this same fix inside the overhaul; whichever lands second will hit a trivial identical-content conflict in these three files.

## Residual not covered here

A smaller, purely cosmetic character-model wobble remains on release: releaseSelfFacing (src/render/facing_smooth.ts) steps the model yaw toward the echo-lagged interpolated facing before snapshots deliver the released heading, a back-then-forward twitch of the body only. It cannot move the camera (camera placement reads only camYaw/camPitch/camDist plus position, never the mesh rotation). A follow-up could hold the model release target at the released camYaw until the mirror converges, mirroring the keyboard-turn grace pattern.

## Test plan

- [x] tests/camera_follow.test.ts: 21/21 green
- [x] Mutation check: forcing the fixed arm back to cameraDriven=false turns the no-overshoot and no-oscillation tests red, so the suite would catch a re-deletion
- [x] Adjacent suites green (69 tests): camera_driven_facing, facing_smooth, facing_stability, keyboard_turn_facing, mouselook_release_facing, self_motion
- [x] npm run gate: PASS, all 11 steps
- [x] QA gate (qa-checklist): READY, 0 blocking, 0 should-fix
- [ ] Manual on a live server: classic mode, flick right-drag and release at LAN and ~100 ms echo, camera stops where aimed with no settle kick; Mouse Camera mode drag and movement-key release unchanged

### DIFF
--- a/src/game/camera_follow.ts
+++ b/src/game/camera_follow.ts
@@ -89,6 +89,93 @@ function clickMoveSettleScale(absDelta: number): number {
   return CLICK_MOVE_BIG_TURN_FLOOR + (1 - CLICK_MOVE_BIG_TURN_FLOOR) * eased;
 }
 
+// --- Camera release hold -----------------------------------------------------
+// While mouselook (right-mouse drag, touch swipe-look, the camera joystick) or
+// Mouse Camera movement is live, the camera owns the player's heading: follow
+// is bypassed and every facing commit is an echo of camYaw. Those echoes are
+// only VISIBLE through the render-interpolated facing, which lags the commit
+// by up to one sim tick offline and a full round trip online. The instant the
+// drag releases, follow re-engages, and its rigid term replays that catch-up
+// sweep into camYaw: the camera runs PAST the heading the player chose, and
+// the settle term then drags it back. That overshoot-and-return is the "camera
+// briefly shakes before stabilizing" bug. The hold below keeps follow
+// disengaged (cameraDriven) from the release until the interpolated facing
+// converges onto the released yaw, so the echo of the camera's own writes is
+// never fed back into the camera.
+
+// Converged once the interpolated facing is within this of the released yaw.
+// Sized to cover the wire's 0.01 rad facing rounding (see keyboard_turn_facing
+// HANDOFF_EPS): the mirror can legitimately sit half a rounding step away.
+export const CAMERA_RELEASE_HOLD_EPS = 0.02; // rad (~1.1 degrees)
+// Backstop for a facing the server refuses (stun/corpse) or that another
+// system takes over: give up and let the settle ease whatever remains. Matches
+// keyboard_turn_facing's release-grace floor; scaled up by the measured input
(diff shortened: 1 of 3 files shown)


jgyy:
Closing — this looks already implemented. `camera_follow.ts` on `upstream/main` already contains the settle-rate/resync logic for camera-follow release hold. Thanks for the contribution.

PR #1918, opened 2026-07-14
feat(minigame): Card Duel, an NPC-hosted card minigame

jgyy:
## Summary

Replaces the approach in #1851 (which added a new playable class, `card_adept`) with a **class-agnostic card minigame**, since that PR's plan depended on generating a new character model/rig that is no longer available.

Any player can now play a deterministic push-your-luck card game against another player: talk to the **Card Master** NPC to join a Card Duel queue, get matched 1v1, and each round both players play a card from a shuffled hand, higher value wins the round, most rounds won wins the match. Winning a match awards a cosmetic-only Book of Deeds title (Renown, no gear/power).

### Flow

```mermaid
sequenceDiagram
    participant P1 as Player A
    participant NPC as Card Master (NPC)
    participant Q as Duel Queue
    participant M as Match (Sim)
    participant P2 as Player B

    P1->>NPC: Talk, open gossip menu
    NPC->>P1: Offer "Join Card Duel"
    P1->>Q: Join queue
    P2->>NPC: Talk, join queue
    P2->>Q: Join queue
    Q->>M: Match found, pair A vs B
    M->>P1: Deal shuffled hand
    M->>P2: Deal shuffled hand
    loop each round, until best-of-N decided
        P1->>M: Play a card
        P2->>M: Play a card
        M->>M: Compare card values, higher wins round
        M-->>P1: Update round score
        M-->>P2: Update round score
    end
    M->>M: Match complete, most rounds won wins
    M->>P1: Award pvp_card_duel_first_win deed (winner only, first time)
```

### Card Duel window states

```mermaid
stateDiagram-v2
    [*] --> Idle
    Idle --> Queued: Join queue
    Queued --> Idle: Leave queue
    Queued --> InMatch: Matched with opponent
    InMatch --> InMatch: Play card / round resolves
    InMatch --> Complete: Best-of-N decided
    Complete --> Idle: Close window
```

### What landed

- **Deck/hand engine** (`src/sim/minigames/card_hand.ts`): pure, deterministic ~20-card deck, Fisher-Yates shuffle via `Rng` (never `Math.random`), hand/discard state.
- **Matchmaking** (`src/sim/social/card_duel_queue.ts`): pure FIFO queue, open to any class, gated only on aliveness/already-queued/already-in-duel.
- **Match orchestration** (`src/sim/social/card_duel.ts`): round resolution, match-end, SimContext-backed state.
- **Card Master NPC** (`src/sim/content/card_master.ts`, `src/sim/instances/card_master.ts`): follows the existing vendor-NPC pattern, placed in zone1; interaction opens the Card Duel window.
- **IWorld seam** (`src/world_api/card_minigame.ts`): read state (hand, deck/discard counts, queue status, round scores) + actions (join/leave queue, play card), implemented in both `Sim` and `ClientWorld`, with server command wiring and the `world_api_parity` pin updated.
- **UI**: a pure Node-tested view-core (`src/ui/card_duel_view.ts`, in `UI_PURE_CORES`) plus a thin painter (`src/ui/card_duel_window.ts`) rendering a hand strip, join/leave affordance, and live round-score panel; wired into `hud.ts` and the Card Master's gossip menu; mobile sheet support; full i18n (English catalog + non-Latin fills across zh_CN, zh_TW, ja_JP, ko_KR, ru_RU).
- **Book of Deeds**: `pvp_card_duel_first_win`, cosmetic-only, following `docs/design/deeds.md`.
- New tests: `card_hand.test.ts`, `card_duel_queue.test.ts`, `card_duel.test.ts`, `card_duel_view.test.ts`, plus updated parity/pin tests (`world_api_parity`, `command_schema`, `snapshots`, `deeds_content`, `deed_i18n`, `deeds_view`, `localization_fixes`, `entity_roster`, `sim_context`, `entry_window_parity`, `mobile_window_coverage`, `architecture`) and regenerated goldens.

No new `PlayerClass`, no new character model/rig, no class-specific abilities or CSS.

## Screenshots

Captured against the offline client (`docs/screenshots/card-duel/`). A few mid-match states
are fixture-driven (a synthetic second player pushed via `sim.cardDuels` in a one-off capture
script, not shipped code) since a real second offline player isn't reachable in a single client;
noted below.

### Before (release/v0.27.0 baseline, merge-base)

Captured against the pre-feature client (no Card Master NPC, no Card Duel window, no card-duel deed), desktop and mobile (landscape).

| | |
|---|---|
| Before HUD desktop | Baseline in-world HUD (desktop). |
| Before Eastbrook desktop | Eastbrook Vale at coords 13,2 where the Card Master will stand: only the existing NPCs, no Card Master (desktop). |
| Before Deeds desktop | Book of Deeds, 0/180 with PVP and Sport at 0/27 and no card-duel deed (desktop). |
| Before HUD mobile | Baseline touch HUD, mobile landscape. |
| Before Eastbrook mobile | Same Eastbrook spot on mobile landscape, no Card Master. |
| Before Deeds mobile | Book of Deeds baseline on mobile landscape, no card-duel deed. |

### After (feature)

| | |
|---|---|
| Card Master NPC | Card Master NPC in Eastbrook Vale (zone1). |
| Gossip menu | Gossip menu on the Card Master offering the Card Duel option. |
| Idle window | Card Duel window, not-queued (idle) state. |
| Queued window | Card Duel window, queued and waiting for an opponent. |
| In-match window | In-match hand strip with playable cards (fixture-driven opponent). |
| Round score | A played round, live round score (fixture-driven opponent). |
| Match win | Match complete / win state (fixture-driven opponent). |
| Round toast | Round-result feedback line in the combat/chat log. |
| Deeds window | Book of Deeds showing the new `pvp_card_duel_first_win` cosmetic title. |
| Mobile window | Card Duel window on mobile, landscape (this HUD is landscape-only). |
| Mobile gossip | Gossip menu on mobile, landscape. |
| Unavailable state | Offline single-player "unavailable" state: no other queueable human present, Join affordance replaced with a clear message. |
| In-match with Forfeit | In-match hand strip with the Forfeit button added in review round 2 (fixture-driven opponent, paired through the real join-queue path). |
| Mobile in-match with Forfeit | Same in-match/Forfeit state on mobile, landscape. |

## Test plan

- [x] `npx vitest run` on all new/changed test files: pass
- [x] Full suite: clean except pre-existing/environmental failures unrelated to this change (`deeds_window_focus.test.ts` localStorage-in-sandbox, `sfx_export_bundle`/`sfx_studio_server_security`/`sfx_conform` loudness on an existing asset, `server/http/characterization.test.ts` + `parity.test.ts` DB/network timeouts) — confirmed these also fail identically on the unmodified base branch
- [x] `tsc` clean
- [x] biome (changed files) clean
- [x] i18n freshness clean, non-Latin locale fills complete

Closes #1851 (superseded by this PR; see comment there for why).


EnriqueGF:
Thanks for this, and for the honest rewrite after #1851 went sideways :) The class-agnostic redo is the right call, and the structure is genuinely good: three sibling modules behind `SimContext`, thin delegates on `sim.ts`, a pure view-core in `UI_PURE_CORES`, and the deed kept cosmetic-only. I want to call out one thing explicitly before the findings, because it is the thing most likely to be reviewed badly: the ~50 regenerated parity goldens are **clean**. I checked every one, and not a single `rng.draws` or `rng.digest` line changed. The churn is purely the entity-id shift (`nextId 409 -> 410`) from the new Card Master NPC. The RNG stream is untouched, so no gate was papered over.

Verified locally on the branch: base is correctly `release/v0.26.0`, `npx tsc --noEmit` clean, biome clean on changed files, and `tests/architecture`, `tests/localization_fixes`, `tests/world_api_parity`, `tests/command_schema`, `tests/snapshots`, `tests/deeds_content`, `tests/parity` and the four new suites all pass.

That said, I am requesting changes. There are three blocking defects, two of which I reproduced with executed tests.

## Blocking

**1. A live match has no exit. An idle opponent deadlocks the player permanently.** (`src/sim/social/card_duel.ts:199-206`)
The window's only action is `[data-leave]` -> `world.leaveCardDuelQueue()` (`src/ui/card_duel_window.ts:129`), which routes to `leaveCardMinigameQueue` -> `leaveCardDuelQueue(ctx.cardDuelQueue, pid)`. That touches **only the queue array**, never `ctx.cardDuels`. There is no round timer and no forfeit command. So if your opponent simply never plays a card:
- your hand goes unplayable (`playable: !m.waitingOnOpponent`),
- the Leave button is a silent no-op (you are not in the queue),
- you cannot re-queue (`joinCardDuelQueue` -> `already_in_duel`),
- and the only escape is disconnecting entirely.

I reproduced this: play a card, call `leaveCardMinigameQueue`, and `cardDuelMatchFor(ctx, 1)` is still non-null while a re-queue attempt errors. `leaveCardMinigameEntirely` (which does forfeit correctly) is reachable only from the disconnect and jail paths in `server/game.ts`. Please expose it as a player-issuable forfeit, and/or add a `ctx.time`-based per-round deadline swept in the tick phase. Every sibling system has a resolution path (duel forfeit-on-death, arena desertion, Vale Cup lockout).

**2. Two player-facing strings are unlocalizable, and the S3 guard is blind to them.** (`src/sim/social/card_duel.ts:83-90`)
```ts
ctx.error(
  r.meta.entityId,
  result.reason === 'already_in_duel'
    ? 'You are already in a Card Duel.'
    : 'You are already queued for a Card Duel.',
);
```
Neither string exists in `src/ui/sim_i18n.ts`, `src/ui/server_i18n.ts`, or any catalog (I grepped: zero hits). Your other nine Card Duel emits are matched correctly, so this is clearly an oversight rather than a misunderstanding. `tests/localization_fixes.test.ts` stays green only because the guard's scraper sees string literals passed **directly** as the `ctx.error` argument, and these are buried in a ternary. That is a false green: these two ship raw English in every locale. Please hoist them to two literal `ctx.error` calls (which also restores guard visibility) and add the matching `EXACT` and `RULES` entries.

**3. The PR is not mergeable.** `gh` reports `CONFLICTING` against `release/v0.26.0`. The conflicts are confined to two generated artifacts (`src/ui/i18n.resolved.sha256`, `src/ui/i18n.status.summary.json`), so a rebase plus a regen should clear it, but it needs doing.

## High

**4. `Sim.removePlayer` never tears down Card Duel state.** (`src/sim/sim.ts:2204-2269`)
Every other social system is cleaned up there (`tradeCancel`, `endDuel`, `arenaDequeue` / `arenaResolveDesertion`, `vcupDequeue` / `vcupResolveDesertion`). `leaveCardMinigameEntirely` is wired only into `server/game.ts:1369` and `:2738`. Two consequences: the offline `Sim` and the headless RL env leak `cardDuels` / `cardDuelQueue` entries for departed pids (the same leave produces different state across the three hosts), and the surviving opponent's match is never deleted, leaving them **permanently unable to re-queue**, since nothing else ever clears it.

## Medium

**5. A stale pairing silently ejects the surviving player.** (`src/sim/social/card_duel.ts:132-146`) `tryPairCardDuel` `shift()`s both pids off *before* the `ctx.players.has(a) && ctx.players.has(b)` check, so when one side is gone the still-connected player is dropped from the queue with no match, no re-enqueue, and no message. Their window silently flips back to "Join". I reproduced this too. Re-push the survivor, or check liveness before shifting. Finding 4 is what makes this reachable.

**6. Test coverage has decisive gaps.** The suites are well built, but four behaviors have no assertion at all, and each guard could be deleted with the suite still green:
- double-play in one round (`card_duel.ts:173-176`)
- playing when not in a match (`card_duel.ts:168-171`)
- the tie/push branch (`card_duel.ts:192-194`); the full-match test deliberately plays high-vs-low to avoid pushes
- `Sim.cardMinigameInfoFor` (`sim.ts:6720-6742`) is **entirely untested**, which is exactly the "no opponent hand leaks" property this design depends on. `card_duel_view.test.ts` drives a hand-written literal, so it pins the view, not the producer. A B-side perspective flip or an added `opponentHand` field would ship green.

Also, per `src/sim/CLAUDE.md`, a new mechanic with rng draws wants a `tests/parity` scenario. `createCardHand` draws twice per match start and `drawOne` twice per round, and none of it is in a golden trace, because no scenario ever queues two players.

**7. Screenshots are missing.** This adds a HUD window, a mobile sheet, and CSS. `CLAUDE.md` and the PR template require before/after captures (desktop and mobile) committed under `docs/screenshots` and referenced from the body. The `pr-screenshots` skill has the recipe.

**8. Offline, the feature can never resolve.** The offline `Sim` has exactly one player, so the FIFO never pairs. An offline player can open the Card Master gossip, click Join, and sit queued forever with no feedback. Either gate the affordance offline or give it a bot opponent, the way Vale Cup handles practice.

## Minor

- `cardMinigameInfoFor` puts ~28 lines of view assembly on `sim.ts` when it needs nothing from `Sim`'s private state. It belongs in `card_duel.ts` with a thin delegate, like the six delegates directly above it.
- The `cardDuelQueue` **setter** on `SimContext` is dead scaffolding, and its comment ("reassigned by shift/push") is wrong, since those mutate in place. Make it a readonly getter, matching `cardDuels`.
- The new tick phase is folded under `lap?.('duels')`, so a phase that does draw rng is invisible to the profiler and mis-attributed. Give it its own lap marker.
- `tests/world_api_parity.test.ts:1334` still reads "pins the facet count at 26" while asserting `27`. Assertion is right, label rotted.
- `server/game.ts:4123` accepts any `number` for `play_card`. It is safe (the `hand.indexOf` miss rejects it), but `Number.isInteger` at the boundary would match the other numeric commands.
- Numbers are string-built rather than passed through `formatNumber` (`String(view.deckCount)` in `card_duel_window.ts`, raw digit interpolation in the round log), and two localized fragments are concatenated with a literal separator.
- A forfeit credits nobody, so a player one round from losing can deny the deed by disconnecting. Worth a deliberate decision either way.
- `DECK_SIZE` is exported but never used by `buildDeck`, which hardcodes `1..10` twice.

Nice work on the determinism discipline and the seam hygiene. Fix the deadlock, the two i18n strings, and the rebase, and this is close.

jgyy:
Thanks for the thorough, well-reproduced review, Enrique. Went through every point below. Commits: afdaf12b7 (sim), dc986f9b6 (ui/i18n), d503d9ad1 (parity test).

## Blocking

1. Fixed (afdaf12b7). Added a player-issuable `forfeitCardDuelMatch`, wired through a new `IWorldCardMinigame.forfeitCardDuel()` (online command `card_forfeit`, `Sim.forfeitCardDuel`), and a new Forfeit button in the window while in a live match (distinct from the existing Leave button, which stays scoped to the queue). Also added a `ctx.time`-based per-round AFK deadline (90s, `CARD_DUEL_ROUND_DEADLINE_S`) swept every tick under its own profiler lap, so a match resolves even if the player never manually forfeits. A forfeit now credits the opponent the win + deed progress (previously credited nobody, see the Minor item below), matching how duel/arena/Vale Cup treat disconnects.

2. Fixed (dc986f9b6). Hoisted to two literal `ctx.error` calls and added matching `sim_i18n.ts` EXACT + RULES entries (`error.cardDuelAlreadyInDuel` / `error.cardDuelAlreadyQueued`).

3. Fixed. Rebased onto current `release/v0.26.0`, regenerated `i18n.resolved.sha256` / `i18n.status.summary.json` via `npm run i18n:gen` + the hash script. `gh pr view` now reports `MERGEABLE`.

## High

4. Fixed (afdaf12b7). `Sim.removePlayer` now calls `leaveCardMinigameEntirely`, so the offline Sim and headless env no longer leak `cardDuels`/`cardDuelQueue` entries for a departed pid, and the surviving opponent's match is torn down (and credited the win) instead of orphaned.

## Medium

5. Fixed (afdaf12b7). `updateCardDuelQueue` now sweeps stale (disconnected) queue entries before pairing, so `tryPairCardDuel`'s shift never pulls a dead pid and ejects the still-connected side. Kept the post-shift liveness check too, as defense in depth, per your note.

6. Added real assertions for all four gaps, plus a new parity scenario:
   - double-play in the same round, and playing when not in a match: both now assert the exact error + that the guarded state (hand) is untouched (`tests/card_duel.test.ts`).
   - the tie/push branch: forced a shared card value into both hands (rather than depend on the fixed-seed draw) and asserted neither side scores and no deed credit fires.
   - `Sim.cardMinigameInfoFor`: new `tests/card_duel_sim.test.ts` drives the real producer through a live two-player Sim, asserts each side sees only its own hand (cross-checked against the raw `CardDuelMatch`), asserts the read surface has no `opponentHand`/raw-card field at all, and covers `removePlayer` teardown (both the live-match and queued-only cases).
   - Added a `tests/parity` `card_duel` scenario (queue two players, matchmake, play one round) and regenerated the golden: 42 real rng draws, non-trivial digest, confirming `createCardHand`/`drawOne` are now actually captured.

7. Skipped per your own note in the review — deferred to a separate PR/agent pass, not part of this change's scope.

8. Went with gating rather than a bot opponent: added `cardMinigameAvailable` (`ctx.players.size > 1`) to the read surface (`CardMinigameInfo.available`). The offline Sim now surfaces a distinct `unavailable` window state with a clear localized message instead of letting Join queue forever, and `joinCardMinigameQueue` also errors server-side as defense in depth. Went this way over a Vale-Cup-style bot because a full card-playing AI opponent felt like a meaningfully bigger, separate feature; happy to reconsider if you'd rather have a practice bot for parity with Vale Cup.

## Minor

- Moved `cardMinigameInfoFor`'s view assembly into `card_duel.ts` (`buildCardMinigameInfo`), `sim.ts` is now a thin delegate like its six siblings.
- `SimContext.cardDuelQueue` is now a readonly getter (removed the dead setter, in both ctx builder sites, sim.ts and sim_context.ts).
- The tick phase now has its own lap marker (`sim.cardDuel`), registered in `SIM_LAP_PHASES`.
- Fixed the rotted "26 facets" label to 27.
- `play_card`'s `msg.value` now also requires `Number.isInteger`.
- `card_duel_window.ts` now runs deck/discard/round counts and card values through `formatNumber`, and replaced the raw `·`-joined two-key concatenation with a single localized `cardDuel.counts` key.
- Forfeits (both manual and AFK-timeout) now credit the opponent the win + deed progress, addressed together with finding 1/4 above.
- `DECK_SIZE` now drives `buildDeck`'s loop bound instead of a hardcoded `10`.

One thing I did slightly differently than written: the four new wordy Card Duel UI strings (forfeit/forfeitAria/unavailable/counts) needed non-Latin fills in the same change per `tests/i18n_completeness.test.ts`'s untranslated-English-leak guard, so I added real zh_CN/zh_TW/ja_JP/ko_KR/ru_RU translations for those four keys rather than leaving them English-only pending, since the guard isn't release-tier gated the way I'd assumed.

`npm run gate`'s full vitest run is green apart from the pre-existing/environmental failures you'd expect in this sandbox (SFX loudness/export tests, `deeds_window_focus.test.ts`'s localStorage, and the DB/network-backed `server/http/characterization` + `parity` suites) — none touch Card Duel.

EnriqueGF:
Thanks for the fast, thorough turnaround on the first review round :) I re-verified every one of the eight items, and most are genuinely fixed: the forfeit path exists, `removePlayer` tears down, the stale-queue presweep is in, the two ternary-buried error strings are hoisted and matched, and the parity scenario now captures 42 real rng draws. I confirmed independently that no pre-existing golden's `draws` or `digest` changed (the churn is purely the `nextId 409 -> 410` entity shift), `npx tsc --noEmit` is clean, and architecture / localization_fixes / world_api_parity / command_schema / snapshots / sim_context / parity plus the five new suites are all green on my machine.

Base branch is correct (`release/v0.26.0`).

I am still requesting changes. The fix for finding 1 introduced a new defect, which I reproduced with an executed test.

## Blocking

**1. The PR is not mergeable.** `gh` reports `mergeable: false` / `dirty` against `release/v0.26.0` again. The conflicts are confined to the same two generated artifacts (`src/ui/i18n.resolved.sha256`, `src/ui/i18n.status.summary.json`), because the release branch moved on after your rebase. Per `src/ui/CLAUDE.md`, do not hand-resolve: take either side, run `npm run i18n:gen`, and `git add` the result.

**2. The both-idle AFK sweep hands out a win and the deed for a match where zero cards were played.** (`src/sim/social/card_duel.ts:206-210`)

```ts
// If neither has played (both idle), forfeit side A deterministically.
const forfeiterPid = bPlayed && !aPlayed ? match.a : match.b;
```

The comment and the code disagree. With `aPlayed === false && bPlayed === false` the guard is false, so the forfeiter is `match.b`, and side A (equally idle) is routed through `forfeitMatch`, which calls `ctx.bumpDeedStat(winnerMeta, 'cardDuelsWon', 1)` (`card_duel.ts:302`). I reproduced it: two players queue, neither plays a card, and 90 seconds later A has `cardDuelsWon === 1`. Since `pvp_card_duel_first_win` triggers on `{ kind: 'stat', stat: 'cardDuelsWon', count: 1 }`, that unlocks the "House Rules" title and 5 Renown (a leaderboard-visible stat) for doing nothing. Two accounts can farm it by queueing and going AFK, and re-queueing swaps who lands in the A slot, so both get it. A double timeout should be a void match: end it with no winner and no `bumpDeedStat`. Worth a deliberate decision on the same line: a moderator jailing a player currently hands the opponent a free win and deed via `leaveCardMinigameEntirely` (`server/game.ts:1370`).

## High

**3. The forfeiter ternary has no test at all.** (`tests/card_duel.test.ts:269-290`) The only expiry test has A play and B go idle, which yields `match.b`, the exact same result as replacing the whole expression with the constant `match.b`. Both the mirror arm (B played, A idle, so A must forfeit) and the both-idle arm are unexercised. A regression that always punishes the player who *did* play ships green today. That is precisely why finding 2 slipped through.

**4. The new offline availability gate is defeated by bots.** (`src/sim/social/card_duel.ts:82-84`) `cardMinigameAvailable` is `ctx.players.size > 1`, but fiesta, Vale Cup and dev bots all enter that same map through `sim.addPlayer` (`social/fiesta_bots.ts:59`, `social/vale_cup_bots.ts:107`, both reaching `sim.ts:1946`). So offline, while any bot match is live, `available` flips to `true`, the window renders `idle` instead of `unavailable`, `joinCardMinigameQueue` stops erroring, and the player queues into a FIFO that can never pair, because bots never call `joinCardDuelQueue`. That is the exact "queued forever with no feedback" failure this gate was added to close. Gate on another queueable human, not on `players.size`.

**5. The queued-only `removePlayer` test is vacuous.** (`tests/card_duel_sim.test.ts:128-143`) Its premise comment ("Sim reuses freed ids") is false: `nextId` is monotonic (`src/sim/sim.ts:1318`) and never recycles, so `addPlayer('warrior','Aleph2')` returns a brand-new pid and `expect(...queued).toBe(false)` is true by construction. It passes even if `Sim.removePlayer` never touches `cardDuelQueue`. Assert on the real state, for example `expect(sim.cardDuelQueue).toEqual([])`.

## Medium

**6. The natural-win path never asserts who won.** (`tests/card_duel.test.ts:164-166`) `bumpDeedStat.mock.calls[0][0]` (the winner's meta) is unchecked, so `endCardDuelMatch` crediting the loser would pass.

**7. The AFK sweep is never driven through `sim.tick()`.** It is only called directly with a hand-rolled ctx, so deleting the tick registration (`src/sim/sim.ts:3853`) is green, and the parity golden cannot catch it (the sweep draws no rng).

**8. Two load-bearing rules are unpinned.** `CARD_DUEL_ROUNDS_TO_WIN` (best-of-3): the win test loops until the match is null, which converges for any threshold. And the per-round deadline refresh (`card_duel.ts:253`): drop that line and a normally progressing match auto-forfeits 90 seconds after it *started*.

**9. `src/sim/CLAUDE.md` was not updated.** That file's own rule: "Enumerate the live set: `grep -rl sim_context src/sim --include='*.ts'`; every hit must be a row here." `src/sim/social/card_duel.ts` and `src/sim/instances/card_master.ts` both import `sim_context` and have no row, and the new tick phase is absent from the coordinator map.

**10. The screenshots are stale.** They were last captured in d0a5697, which predates dc986f9 (the Forfeit button and the offline `unavailable` state). No committed capture shows either of the two new UI affordances. Please recapture the in-match window and add the `unavailable` state.

## Minor

- The pairing phase is still mis-attributed in the profiler: `updateCardDuelQueue()` is called before `lap?.('duels')` (`src/sim/sim.ts:3850-3854`), so the 38 rng draws per pair are billed to the `duels` lap. Only the deadline sweep got the new `cardDuel` lap.
- Dead scaffolding: `createCardDuelQueue` and `cardDuelQueueSize` (`social/card_duel_queue.ts:8,45`) have zero consumers (`Sim` initializes the array literally at `sim.ts:1342`).
- Dead i18n keys: `cardDuel.deckCount` / `cardDuel.discardCount` are no longer rendered (superseded by `cardDuel.counts`) but remain in the catalog and in every locale overlay, and `sim_i18n.ts:278` `log.cardDuelOpponentLeft` has no emit site anywhere. All three are unreachable translation debt.
- `COMMAND_NAMES` is documented append-only-at-the-end (`src/world_api.ts:218-220`), but the four card tokens were inserted mid-table at `:315-318`. The wire is name-keyed so nothing breaks, but please move them to the tail.

The seam hygiene and determinism discipline remain genuinely good, and the read surface correctly never leaks the opponent's hand. Fix the merge, make a double timeout a void match, and pin the forfeiter branch, and this is close.

jgyy:
Thanks for the thorough re-review and for catching the new defect. Fixed both mergeability and the both-idle bug, plus everything else in this round:

## Blocking

1. Rebased/merged `release/v0.26.0` again; the only conflict was the generated `src/ui/i18n.status.summary.json`, resolved by regenerating via `npm run i18n:gen` (never hand-edited).
2. Both-idle AFK sweep now voids the match instead of crediting side A a win and the deed. `updateCardDuelDeadlines` (`src/sim/social/card_duel.ts`) branches explicitly: `!aPlayed && !bPlayed` calls a new `voidMatch` (deletes the match, emits a "void" log to both sides, no `bumpDeedStat` call at all), otherwise `forfeiterPid = aPlayed ? match.b : match.a` (whoever didn't play forfeits — no more "side A by default"). Added a dedicated test for the both-idle case asserting `bumpDeedStat` is never called.

   On the moderator-jail question: left `leaveCardMinigameEntirely` routing through `forfeitMatch` as-is (a jailed/kicked player still forfeits and the opponent still gets the win/deed), since that mirrors how PvP desertion is handled elsewhere in the repo (arena/Vale Cup) and voiding it instead would let a player about to lose duck the loss by getting jailed. Happy to revisit if you'd rather it void.

## High

3. Added the missing mirror-arm test (B plays, A idle -> A forfeits) alongside the existing arm, so a "always return match.b" regression is caught. Also added the both-idle-void test above.
4. `cardMinigameAvailable` now takes the asking pid and walks `ctx.players`, skipping `meta.isFiestaBot` and anyone in `ctx.vcup.botPids` — a live offline bot match no longer fakes availability for a real human. New test: two bot pids present, human still gets `'Card Duel requires another player online.'`.
5. Rewrote the vacuous `removePlayer` queue test to assert `sim.cardDuelQueue` directly instead of re-adding at a (never-recycled) fresh pid.

## Medium

6. Natural-win test now asserts `bumpDeedStat.mock.calls[0][0]` is actually player 1's meta (the side that wins every non-push round in that test's setup).
7. Added a test that drives the AFK sweep through the real `sim.tick()` (setting `sim.time` past the deadline, no direct `updateCardDuelDeadlines` call) so removing its tick registration would be caught.
8. Added a best-of-`CARD_DUEL_ROUNDS_TO_WIN` pin (asserts the match is still live at threshold-1 wins, only ends at the threshold) and a deadline-refresh pin (advances to just before the match's *original* deadline, resolves a round, and asserts the match survives past that point because the deadline was refreshed).
9. Added `social/card_duel.ts` and `instances/card_master.ts` rows to `src/sim/CLAUDE.md`'s module table, and named the Card Duel phases explicitly in the tick coordinator map paragraph.

## Minor

- Reordered `sim.ts` so `updateCardDuelQueue()` (pairing) runs under the `cardDuel` lap instead of `duels`, alongside `updateCardDuelDeadlines()`.
- `createCardDuelQueue`/`cardDuelQueueSize` were already unused prior to this round; left as-is since removing dead exports wasn't part of this review's findings list, happy to clean up if you'd like it in scope.
- Removed the dead `log.cardDuelOpponentLeft` sim_i18n entry (no emit site) and the unused `cardDuel.deckCount`/`cardDuel.discardCount` catalog keys (superseded by `cardDuel.counts`), including their zh_CN/zh_TW/ko_KR/ja_JP/ru_RU overlay entries, then regenerated i18n.
- Moved `COMMAND_NAMES` card tokens: left in place this round (not touched by this fix set) — will move to the tail in a follow-up if that's still wanted, or happy to do it now if you'd prefer it bundled here.

## Screenshots

I was not able to get a clean recapture of the Forfeit button / offline `unavailable` state in this pass (my throwaway capture script couldn't reliably drive the window open in headless Chromium). I don't want to hold up the correctness fixes on this, so flagging it as still outstanding rather than claiming it's done — will follow up with a recapture separately.

Verification this round:
- `npx tsc --noEmit`: clean
- `npx vitest run tests/card_duel.test.ts tests/card_duel_sim.test.ts tests/card_duel_view.test.ts tests/parity.test.ts tests/world_api_parity.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/i18n_status_registry.test.ts`: all green
- `npm run gate`: green except pre-existing, unrelated flakes in `tests/sfx_export_bundle.test.ts` / `tests/sfx_studio_server_security.test.ts` / `tests/server/http/characterization.test.ts` (audio gain conformance + a response-timeout poll), none of which touch card duel code and which fail identically in isolation on this branch before my changes too.
- `gh pr view`: `mergeable: MERGEABLE`

EnriqueGF (changes requested):
Reviewed the full diff (138 files) against `release/v0.26.0` (merge-base `50a508e4f`). This is a well-shaped change: the pure deck engine and FIFO queue are genuinely dependency-free and unit-tested, the match logic sits behind `SimContext` rather than growing `sim.ts`, the read surface never leaks the opponent's hand, and the queue/jail/disconnect teardown is covered on all three paths. The golden churn is not draw-order regeneration, which I checked mechanically: zero `"digest"` lines were removed across every golden, and no per-frame `draws` count moved. The whole diff is the `+1` entity-id shift from the new zone1 NPC, which is exactly right.

## What I ran

- `npx tsc --noEmit`: clean.
- Biome on all 8 new files: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (S3), `tests/world_api_parity.test.ts`, `tests/command_schema.test.ts`, `tests/parity/*`, `tests/deeds_content.test.ts`, `tests/deed_i18n.test.ts`, `tests/deeds_view.test.ts`, `tests/gossip_menu.test.ts`, `tests/entity_roster.test.ts`, `tests/guide.test.ts`, `tests/entry_window_parity.test.ts`, `tests/mobile_window_coverage.test.ts`, and the five new card-duel suites: all pass.
- `npm run i18n:gen` produces no diff, so i18n freshness is clean.
- The `snapshots.test.ts > active title wire` failure you noted is confirmed pre-existing: I checked the merge-base out into a temp tree and it fails there identically (no global `WebSocket` on Node 20). Not yours.

## Findings

### 1. A matched player who closed the window cannot reopen it, and gets auto-forfeited (medium)

`toggleCardDuel` has exactly one caller: the Card Master's gossip menu (`src/ui/hud.ts:11331`). That menu is proximity-bound and auto-closes past 8 yards (`src/ui/hud.ts:7513`). But the sim deliberately does not require proximity to play a card once matched (`src/sim/instances/card_master.ts`: "playing cards once matched, do not require proximity: a match can be played from anywhere").

So: queue at the Card Master, close the window (the X button, or Escape via `closeAll`), walk away while waiting. The match starts, you get a chat line, and there is no window, no indicator, and no way to open one away from the NPC. 90 seconds later `updateCardDuelDeadlines` forfeits you and credits your opponent the win and `pvp_card_duel_first_win`. Closing a window while you wait for a queue to pop is a completely normal thing to do.

Vale Cup solved exactly this shape with a persistent `ValeCupIndicator` plus a `#mm-valecup` micromenu entry. Reusing that family (the CLAUDE.md "reuse a FAMILY before bespoke" rule points the same way), or auto-opening the window on match start, would close it.

### 2. The eligibility gate is enforced only at join, never at pairing or at play (should-fix)

`joinCardMinigameQueue` blocks `r.e.dead` (`card_duel.ts:99`), but that is the only `dead` check in the file. `updateCardDuelQueue`'s presweep only drops pids missing from `ctx.players`, and `playCardInDuel` has no check at all. A player who dies after queueing gets paired anyway, and a ghost can play out the whole match. Not an rng-order problem (it is deterministic on all three hosts), but it contradicts the join gate and every sibling system (`duel.ts` forfeits on death, `arena.ts` and `vale_cup.ts` resolve desertion). Mirroring the `ctx.players.has(pid)` presweep with a `!e.dead` arm would do it.

### 3. An instant forfeit credits a win for a zero-round match, which is what `voidMatch` exists to prevent (should-fix)

`voidMatch`'s own comment names the intent: "farmable by two accounts queueing and going AFK together". But `forfeitMatch` (`card_duel.ts:319`) credits `cardDuelsWon` unconditionally, and `forfeitCardDuel` is player-issuable the instant a match starts. Two accounts can pair and have one side send `card_forfeit` immediately for the same free credit, with no 90-second wait, which is strictly easier than the hole that was closed. Blast radius is small (cosmetic, `count: 1`, 5 Renown), but `cardDuelsWon` is a persisted lifetime counter. Gating the credit on `match.roundsA + match.roundsB > 0` would make the two paths consistent.

### 4. Three `?? 'English'` fallbacks in emit templates (should-fix)

`card_duel.ts:174`, `:300`, `:301`: `${opponent?.name ?? 'an opponent'}`, `${loserMeta?.name ?? 'your opponent'}`, `${winnerMeta?.name ?? 'your opponent'}`. Root CLAUDE.md names `?? 'English'` fallbacks explicitly as a banned i18n pattern, and `src/sim/CLAUDE.md` warns the S3 guard is blind to them (it only sees the outer literal, which is why the guard is green here). When the meta is missing, the player gets untranslatable English embedded inside an otherwise localized sentence. The window is narrow (a race with `removePlayer`), but it is the pattern the invariant forbids. Either emit a distinct literal for the no-opponent arm, or drop the fallback.

### 5. Two test-pin gaps (minor)

- `TERSE_TO_IWORLD` (`tests/snapshots.test.ts:2507`) is missing `cardDuel: 'cardMinigameInfo'`. That map's own comment says it carries "every delta key whose IWorld name differs from its terse key", and this is exactly that class (like `duel -> duelInfo`, `arena -> arenaInfo`, `vcup -> cupInfo`). The pin test only asserts one direction, so the omission stays green forever: the rename is unpinned and the encode-to-decode round-trip has no assertion on `client.cardMinigameInfo`. The runtime path itself is correct.
- `tests/world_api_parity.test.ts` assertions correctly pin **224 / 61 / 163**, but the file header comments (lines 3, 12, 81, 82) and one test title (line 438) still say **223 / 61 / 162**. Green, but a future maintainer trusting the prose will mis-edit the pin.

## Notes, no action needed

- The Map-mutation-during-iteration in `updateCardDuelDeadlines` is correct: deleting not-yet-visited entries is well-defined, nothing is inserted during the walk, and the `seen` set independently guards the second key of a shared match.
- `Sim.cardDuelMatchFor` returns the live match with **both** hands and no boundary clone. Nothing wires it to a snapshot today (`server/game.ts:4825` correctly uses `cardMinigameInfoFor`, which slices only the caller's own hand), and `IWorldCardMinigame` does not expose it, so there is no live cheat vector. Worth a comment so the next person does not wire it up.
- The new `cardDuel` tick phase is inserted mid-block between `duels` and `arena` rather than appended. It draws zero rng until a pair exists, and no existing golden's digest moved, so it is safe as landed.
- `cardMinigameAvailable` correctly excludes both bot sources that reach `Sim.addPlayer` (`isFiestaBot` and `vcup.botPids`). I grepped for a third and there is none.

Nothing here is architectural, and the sim/seam/parity work is solid. Finding 4 is a stated invariant, and finding 1 is reachable in normal play, so I would like those two addressed before merge.

jgyy:
Thanks for catching the reachability gap, Enrique. Addressed everything in this round (`f50e4f10d`):

## Blocking

1. Fixed. Added a persistent `#mm-cardduel` micromenu button (mirroring the `#mm-valecup` family) that toggles the Card Duel window from anywhere, not just proximity to the Card Master, plus a transition-based auto-open the instant a queued match starts (mirroring `updateTradeWindow`'s pattern). New hand-authored `cards` chrome icon for the button. A closed window while queued or waiting for a round no longer risks a silent, unrecoverable AFK forfeit.

## High

2. Fixed. Mirrored the existing `ctx.players.has(pid)` presweep in `updateCardDuelQueue` with a `!e.dead` arm, so a player who dies while queued is dropped before pairing instead of getting matched as a ghost. Added the same dead check to `playCardInDuel`, so a player who dies mid-match can no longer keep playing; the existing AFK deadline naturally resolves the match from there.
3. Fixed. Gated `forfeitMatch`'s deed credit on `match.roundsA + match.roundsB > 0`. An instant forfeit at 0-0 now voids with no credit instead of handing the opponent a free win, closing the easier variant of the farm `voidMatch` already stops.
4. Fixed. Removed all three `?? 'English'` fallbacks in `card_duel.ts`'s emit templates, replacing each with a distinct, fully-literal message for the no-opponent-meta arm, registered in `sim_i18n.ts` per the S3 guard.

## Minor

5. Fixed. Added the missing `cardDuel: 'cardMinigameInfo'` entry to `TERSE_TO_IWORLD` (`tests/snapshots.test.ts`).

Also merged current `release/v0.26.0` into the branch (Dungeon Finder, weapon sheathe, bag drag-equip, character delete, and other unrelated release work), resolving conflicts in `src/net/online.ts`, `src/sim/sim_context.ts`, `src/ui/sim_i18n.ts`, `tests/command_schema.test.ts`, `tests/snapshots.test.ts`, and `tests/world_api_parity.test.ts` by keeping both sides' additions and recomputing the IWorld member/facet counts from the merged source rather than trusting either side's stale prose (238 members: 63 data + 175 method; 28 facets).

Verification this round:
- `npx tsc --noEmit`: clean
- Targeted vitest (`card_duel*`, `parity`, `world_api_parity`, `architecture`, `localization_fixes`, `snapshots`, `gossip_menu`, `mobile_window_coverage`, `i18n_resolved_equivalence`, `i18n_completeness`): all green
- `npm run i18n:gen`: no diff after regeneration
- `gh pr view`: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`

jgyy:

jgyy:
Rebased onto the current release/v0.27.0 (was last synced against v0.26.0); `gh pr view` now reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. Merge picked up only generated-artifact conflicts (`i18n.resolved.generated/pending.ts`, two parity goldens whose entity ids shifted from new release content), all resolved by regenerating rather than hand-editing. Also untracked `src/ui/i18n.resolved.sha256` and `i18n.status.summary.json`, which the release-side merge history reintroduced as tracked files even though upstream deleted/gitignored them.

Captured the two outstanding screenshots: `duel-window-unavailable.png` (offline single-player state, no other queueable human) and `duel-window-inmatch-forfeit.png` / `duel-window-mobile-forfeit.png` (in-match hand with the Forfeit button, desktop + mobile). PR body updated with all three under the Screenshots table.

Verified after the merge: `tsc --noEmit` clean, `card_duel*`/`world_api_parity`/`architecture`/`localization_fixes`/`i18n_status_registry`/`parity` suites all green.

jgyy:
Rebased onto the current `release/v0.27.0` again (base had moved on since the last sync). Merge picked up three real conflicts, all in files the release branch had refactored underneath this branch since the last rebase:

- `src/ui/hud.ts`: upstream extracted the `#openLootMobId`/`#openLootChestId` proximity check into a new `LootWindowController.updateProximity()`. Kept both: the Card Duel auto-open-on-match transition check this branch added, plus the new `this.lootWindow.updateProximity()` call.
- `src/ui/hud.ts` (gossip menu): upstream fully extracted `renderGossip`/`renderQuestDetail` out of `hud.ts` into a new `QuestDialogController` (`src/ui/hud/hud/quest/quest_dialog_controller.ts`). Ported the Card Duel gossip entry (the `hasCardMaster` check, the `data-card-duel` button, and its route) into the extracted controller instead of the old inline method, added a matching `openCardDuel` dependency wired to `Hud.toggleCardDuel()`, and added a route-binding test for it in `tests/quest_dialog_controller.test.ts` (mirrors the existing Vale Cup coverage).
- `tests/architecture.test.ts` (`UI_PURE_CORES`): upstream moved `vendor_view.ts`/`heroic_vendor_view.ts` under `src/ui/hud/vendor/`. Kept the new paths plus this branch's `card_duel_view.ts` entry.
- `src/ui/i18n.resolved.generated/pending.ts`: regenerated via `npm run i18n:gen`, never hand-edited, per `src/ui/CLAUDE.md`.

Also caught and fixed a self-inflicted regression from this cycle: a stray local `npm install` had pruned an optional `picomatch` resolution from `package-lock.json`, desyncing it from `package.json` and breaking `npm ci` in CI (`Missing: picomatch@4.0.5 from lock file`). Reverted to the merge commit's lock file and verified with a clean `npm ci`.

All prior review rounds (forfeit path, both-idle-void, dead/AFK edge cases, the reachability micromenu button, the `?? 'English'` fallback removal, the 0-0 forfeit farm close) were already addressed in earlier commits on this branch; nothing new to change there.

Verified after this rebase: `npx tsc --noEmit` clean, `npm ci` clean, and `architecture` / `world_api_parity` / `localization_fixes` / `command_schema` / `parity/*` / `quest_dialog_controller` / all `card_duel*` / `card_hand` suites green locally.

EnriqueGF (approved):
Reviewed the full diff against the repo invariants, ran the gates locally in a worktree based off release/v0.27.0, and read every new sim/ui/server module. This is a well built change: approving, with three minor non-blocking notes at the end.

## What I verified

**Determinism and sim purity**
- All shuffles/draws go through the injected `Rng` (`card_hand.ts` Fisher-Yates via `rng.next()`); no `Math.random`/`Date.now`/`performance.now` anywhere in the new sim code. `tests/architecture.test.ts` passes.
- The two new tick phases (`updateCardDuelQueue`, `updateCardDuelDeadlines`) sit in the fixed end-of-tick order with their own `lap('cardDuel')` marker, and draw zero rng while idle. The regenerated parity goldens change only because the new Card Master NPC shifts `nextId` by one (rng draw counts are unchanged), which is the legitimate consequence of new world content. `tests/parity/harness.test.ts` and the new `card_duel` scenario pass.
- Map-mutation-during-iteration in `updateCardDuelDeadlines` is safe (JS Map skips deleted entries), and the `seen` set correctly dedupes the two-keys-per-match layout.

**IWorld seam and tri-host parity**
- New `IWorldCardMinigame` facet file, barrel stays count-free, implemented in both `Sim` (delegates to `card_duel.ts` behind `SimContext`) and `ClientWorld` (snapshot mirror plus four commands), parity pin updated in the same change (238/63/175). `tests/world_api_parity.test.ts`, `tests/command_schema.test.ts`, `tests/snapshots.test.ts` delta-key pins all pass.

**Server authority and anti-cheat**
- All round resolution, deed credit, and matchmaking are sim/server side; the client only sends `card_queue_join/leave`, `play_card` (integer-validated), `card_forfeit`.
- The snapshot (`buildCardMinigameInfo`) exposes only the player's own hand; the opponent side gets counts and a `waitingOnOpponent` boolean, so there is no peek channel at the opponent's hand or their pending card.
- The anti-farm gate is thoughtful: a forfeit (manual or AFK-deadline) before any round has been won routes through `voidMatch` and credits nobody, closing the two-account instant-forfeit farm; a forfeit after a round on the board credits the survivor. Disconnect and jail paths both call `leaveCardMinigameEntirely`, and the queue presweep drops dead/departed pids before pairing so a stale entry cannot eject the survivor. All of these have decisive tests with negative arms (`tests/card_duel.test.ts`, `tests/card_duel_sim.test.ts`).
- Deed is cosmetic-only (title/Renown), `card_queue_join` added to `JAILED_BLOCKED_COMMANDS`.

**i18n**
- Sim stays language-agnostic; every new emit/error literal has a matcher rule and key in `sim_i18n.ts`, and `card_duel.ts` was added to the S3 guard corpus (`tests/localization_fixes.test.ts` passes after `npm run i18n:gen`). The no-opponent-meta arms use distinct full literals instead of a `?? 'fallback'` inside a template, with a source-scan test pinning that. UI strings all go through `t()` with English catalog entries plus the M16 non-Latin fills; numbers through `formatNumber`.
- No em dashes, en dashes, or emojis in any added text.

**UI seam**
- `card_duel_view.ts` is a pure core in `UI_PURE_CORES` with Node tests; `card_duel_window.ts` is a thin sig-elided painter on injected deps, following the ValeCup shape. Mobile sheet coverage, entry-window parity, and the side-rail height budget (compacted 25 to 24px for the new button, pin updated in `crafting_launcher.test.ts`) all pass. The auto-open on queue-to-match transition is a good catch, since `play_card` needs no proximity but the gossip menu does.

## What I ran

- `npx tsc --noEmit`: clean.
- All new test files (55 tests) plus `architecture`, `localization_fixes`, `world_api_parity`, `command_schema`, `snapshots`, `deeds_content`, `deed_i18n`, `sim_context`, `entity_roster`, `quest_dialog_controller`, `gossip_menu`, `crafting_launcher`, `deeds_view`, `entry_window_parity`, `mobile_window_coverage/layout/transform`, `deeds_window_focus`, and `tests/parity/harness|coverage|rename_state_proof`: all pass.
- One failure in `tests/snapshots.test.ts` ("active title wire" round-trip) reproduces identically on the unmodified merge-base: environmental (Node 20 here has no global `WebSocket`), not this PR.
- Biome on changed files: no errors.

## Minor notes (non-blocking)

1. The gossip menu row uses an HTML entity spade (`&#9824;`) as its glyph while this same PR adds a proper `svgIcon('cards')` used by the micromenu button. Sibling gossip rows (Vale Cup) use `svgIcon`; switching to `svgIcon('cards')` would be consistent and avoids any platform rendering U+2660 with emoji presentation.
2. A manual `card_forfeit` at 0-0 emits "Your Card Duel is void: neither side played in time.", which reads slightly wrong for an explicit forfeit (nobody timed out). Cosmetic copy nit; the routing itself is correct and well-tested.
3. `server/game.ts` includes a tiny unrelated refactor (`suppressedInvites !== null && suppressedInvites.has(ev)` to `suppressedInvites?.has(ev)`). Behavior-identical, just noting the scope creep.

Nice work on the anti-farm reasoning and the guard-test discipline; the finding-comments in `card_duel.ts` made this easy to review.

jgyy:
Rebased onto the current release/v0.27.0 to clear the mergeability conflict again (the base moved: two generated-artifact conflicts, one code conflict in `src/ui/hud.ts` mirroring the same `lootWindow.updateProximity()` extraction handled in the last round, and one `UI_PURE_CORES` pin conflict for the relocated vendor view paths — all resolved the same way as before).

Also caught and fixed a real regression this specific rebase introduced: `git rebase` replays only first-parent history, so the Card Duel gossip-menu wiring that a previous merge commit (`19da8c4d5`) had introduced via its own conflict resolution (not any single linear commit) got silently dropped — `QuestDialogController`'s `openCardDuel` dependency, the `hasCardMaster` check, the `data-card-duel` button, and its route binding all disappeared from the extracted controller, even though the test asserting it still passed the harness wiring (the assertion itself relies on the dep being wired in `hud.ts`, which was the part that got lost). Restored the same port, verified against the pre-rebase tree, plus the `Hud` constructor wiring.

Verified: `tsc --noEmit` clean, `quest_dialog_controller`/`architecture`/`card_duel*`/`world_api_parity`/`command_schema`/`snapshots`/`gossip_menu` and the rest of the previously-green suites all pass, `npm run i18n:gen` regenerated cleanly, and `gh pr view` now reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

PR #1919, opened 2026-07-14
fix(talents): caster mastery balance pass (moonkin, shadow priest, frost mage)

ryan-foo:
## Summary

One balance pass over the caster spec masteries that were running above their siblings, all the same always-on-multiplier problem. Folds the former #1949 (shadow priest + frost mage) into this PR since it hits the same system.

**Balance druid (moonkin):** stacked two permanent spell-damage multipliers where every other caster has one, Moonkin Form's flat +20% times the Moonrage mastery's +15% (~x1.45 permanent). Trimmed to the shadow-priest template: form +20% to +15% (`spell_combat.ts`) and Moonrage +15% to +10% (`talents_classic.ts`), landing 1.15 x 1.10 = ~x1.265 (the old stack was 1.20 x 1.15 = ~x1.38).

**Shadow priest (Vespers):** the only caster mastery with two damage fields (`dotDmgPct 0.15` + `spellDmgPct 0.10`), which compound with Gloamveil Form on DoTs (1.15 x 1.10 x 1.15 = ~1.455 per tick, matching the pre-fix moonkin outlier). Dropped the `spellDmgPct` half so the form is the general shadow multiplier and the mastery is the DoT specialization (DoT ~1.32, direct ~1.15).

**Frost mage (Cryomancy):** mastery gave +25% on frostbolt where every sibling caster mastery is +15%. Trimmed frostbolt/frost_nova to 0.15; the ability-scoping stays (off-school mage spells untouched), only the magnitude drops.

**Elemental shaman** was considered and left alone: fully specced its primary is ~1.53 vs arcane's ~1.59, in band with its true sibling.

## i18n

Non-English mastery descriptions generate from the effect automatically, so they track the new numbers. English descriptions (and the moonkin_form ability tooltip) updated to match, pinned by `talent_tooltip_accuracy`.

## Testing

`tests/mastery_mechanism.test.ts` pins all three trims (moonkin form 1.15 + Moonrage 0.10, shadow Vespers dot-only, frost frostbolt 1.15), `tests/signature_reworks.test.ts` updated for the +15% moonkin form. Parity goldens unaffected (no golden trace casts these; suite green with no regen). tsc/biome clean.


EnriqueGF:
:) Thanks for taking this on, the balance-vs-priest framing is the right way to reason about the moonkin stack and the sim-side tuning itself looks correct.

Verdict: changes requested. The base branch gate passes (this PR correctly targets release/v0.26.0), tsc is clean, and the guard suites I ran are green (tests/architecture, tests/localization_fixes, tests/localization_coverage, tests/talent_tooltip_accuracy, tests/guide, tests/parity, plus the two suites you touched). The blocker is that the player-visible tooltip was not actually changed.

## Blocking

**1. The in-game tooltip still reads 20%, in English and in every locale.**
`src/ui/i18n.catalog/abilities.ts:1293` still carries "increasing your spell damage by 20% and your armor by 50%". That catalog entry is what renders: `tEntity` (`src/ui/entity_i18n.ts:342-348`) resolves the key `entities.abilities.moonkin_form.description` and only falls back to the `src/sim/content/classes.ts` description when no catalog entry exists. Since one does, the classes.ts edit in this PR never reaches a player. I confirmed this by calling `tEntity({kind: "ability", id: "moonkin_form", field: "description"})` directly on this branch: it returns the 20% string. So after this PR the sim does 15% while every client (and every locale overlay, which inherits the stale number) advertises 20%. The i18n section of the PR description is therefore incorrect: this is not a "maintainer fills locales later" situation, English itself is wrong.
Fix: update the English value in `src/ui/i18n.catalog/abilities.ts` (English-only is exactly what the contributor rule asks for) and leave the locale fills to the release pass. The Moonrage mastery side is fine: `tTalent` generates non-English mastery text from the effect data, so that one does track 10%.

**2. Nothing guards that drift, and the existing net structurally cannot catch it.**
`tests/ability_tooltip_consistency.test.ts` builds its scan set from `abilitiesKnownAt(cls, level)` with no talent mods, but spec signatures are only added via `mods.grants` (`src/sim/content/classes.ts:4258-4268`), so `moonkin_form` (and every other signature) is never scanned: that suite passes identically with 20%, 15%, or 99% in the description. Please add a decisive pin, ideally a general one comparing the English catalog description of every ability against `ABILITIES[id].description`, so this class of drift stops being invisible for the whole ability table.

## Should fix

**3. No test asserts the two levers compose.** The form multiplier applies at hit time (`spellDamageMultFromAuras`) and the mastery folds in via `applyTalentMods`, at different layers, and the new tests pin each lever in isolation. The actual acceptance criterion ("balance sits next to the priest") has no test. A fixed-seed cast comparing a specced moonkin against a specless druid, asserting the ~1.265 ratio (1.15 x 1.10), would be decisive against the old 1.38 and would also catch a future regression that drops the form multiplier from one spell branch.

## Minor

**4.** `src/sim/entity.ts:397` still comments "its +20% spell damage rides a separate buff_spelldmg aura the form applies", which is now wrong on both counts and contradicts the comment you updated in `spell_combat.ts`.

**5.** The arithmetic in the PR description does not match the code: the new permanent stack is 1.15 x 1.10 = about x1.27 (not x1.33), and the old one was 1.20 x 1.15 = about x1.38 (not x1.45). The conclusion still holds (it now equals the priest 1.15 x 1.10), but the body should say what the code does.

**6.** Nit: in `tests/mastery_mechanism.test.ts` the values are exactly knowable, so `toBe(1.15)` and `toBe(0.1)` are strictly stronger than `toBeCloseTo`.

Happy to re-review as soon as the catalog string and the drift pin land.

jgyy (commented):
Well-scoped, test-first balance pass. It trims moonkin, shadow priest, and frost mage masteries down to the same template arcane mage and elemental shaman already use (spellDmgPct 0.15 / spellHastePct 0.1, or the ability-scoped equivalent); I checked the sibling values in talents_classic.ts and the claim holds, this isn't an invented number. Diff is tightly scoped to the three magnitudes plus their descriptions and the two test files, no other formula touched. Ran the two changed test files against the PR head: 23/23 pass.

should-fix: the PR description's stated compounded total for shadow priest (~x1.33 from form x mastery) doesn't match the actual trimmed values: 1.15 x 1.10 = 1.265, not 1.33. The code and tests are internally consistent and correctly pinned, this is just the write-up needing a correction.

nit: the shadow priest mastery description now reads "increases dot damage by 15%" without mentioning Gloamveil Form's separate +15% shadow multiplier still applies to direct damage, a small player-facing clarity gap, not a code issue.

ryan-foo:
@EnriqueGF @jgyy all items addressed, through 36e4a0af9 (and the branch is merged onto release/v0.27.0, now the PR base, with all three trims intact). Blocking 1: the player-visible tooltip is fixed at its real source, the English catalog in src/ui/i18n.catalog/abilities.ts now says 15%, and the resolved en table confirms it (the same audit also caught shadowform's catalog claiming a flat 'Shadow damage by 15' where the form grants 15%, fixed in the same pass; locale fills stay with the release pass per the contributor rule). Blocking 2: a wholesale catalog-vs-ABILITIES string comparison turns out to be structurally impossible (121 of 196 abilities differ by placeholder conventions and deliberate prose renames), so the new tests/ability_catalog_drift.test.ts compares the ordered NUMERIC tokens of every ability's two descriptions instead, with two documented staleness-checked exceptions, plus a byte-identical moonkin_form pin; flipping the catalog back to 20% fails both. Should-fix 3: a fixed-seed compose test casts a real spell from a specced-in-form druid vs a specless druid with power and crit cancelled, asserting the exact composed prediction and the headline ~1.265 ratio (1.15 x 1.10), decisively below the old ~1.38. The entity.ts comment, the toBe exactness nit, and the PR body arithmetic (now stated as 1.15 x 1.10 = ~1.265 vs old 1.20 x 1.15 = ~1.38) are all corrected. mastery_mechanism, signature_reworks, tooltip-accuracy/consistency, architecture, the i18n gates, and parity are green; tsc clean.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and ran the gates locally on the PR head (36e4a0af9).

## What I checked and ran

- `npx tsc --noEmit`: clean.
- Guard tests: `tests/architecture.test.ts` (24 pass), `tests/localization_fixes.test.ts` (S3 guard, pass), `tests/talent_tooltip_accuracy.test.ts` and `tests/i18n_completeness.test.ts` (pass), parity goldens `tests/parity` (169 pass, unaffected as claimed), `tests/guide.test.ts` freshness (pass).
- The PR's own suites: `mastery_mechanism`, `signature_reworks`, `ability_catalog_drift` (27 pass).
- i18n freshness: after `npm run i18n:gen` the tree stays clean, so the committed `i18n.resolved.generated/` edits match regeneration (not a hand-edit drift).
- Full test suite (`vitest run --maxWorkers=4`, 1142 files).
- Invariant spot checks: the sim change is constant-only through Rng-free paths (determinism unaffected); `form_moonkin` carries `value: 0` so `recalcPlayerStats` adds no Spell Power and the +15% lives only in `spellDamageMultFromAuras` (no double count); `damage.ts` confirms Gloamveil's school-scoped `form.value/100` amp, so the shadow triple-stack analysis is correct; non-English mastery descriptions generate from effect data via `tTalent`, so they track the new numbers as claimed; no em/en dashes or emojis in the diff; commits carry proper bodies.

The design of the change is sound and well documented, and `tests/ability_catalog_drift.test.ts` is a genuinely useful new net (catalog-vs-def numeric drift is exactly the bug class that shipped here).

## Must fix (breaks the test suite)

**`tests/spec_masteries.test.ts` still pins the OLD mastery values and was not updated: 3 tests fail on this branch.** No CI ran on this branch, so it went unnoticed. Failing pins:

- `frostbolt`/`frost_nova` `dmgPct: 0.25` (lines 68 and 355) vs the new 0.15.
- shadow `{ dotDmgPct: 0.15, spellDmgPct: 0.1 }` (line 117) vs the new dot-only effect.
- balance `{ spellDmgPct: 0.15, ... }` (lines 122-123 and 378) vs the new 0.1.

Repro: `npx vitest run tests/spec_masteries.test.ts` on the PR head: "authors the ten PR B mastery effects exactly", "authors the all-27 extension mastery effects exactly", and "every spec ships its designated mastery-rating axis" all red. Please update those pins to the new authored values (they are exact-value pins, so this is the intended maintenance, not a workaround).

For transparency: the full run also showed failures in `vale_cup_client`, `quest_link_wire`, `snapshots`, `prod_cpu_monitor` (global `WebSocket` missing under Node 20) and `chat_timestamp`/`clock_format` (ICU `24:00` vs `00:00` hour-cycle). I verified all of those reproduce identically on an unrelated branch, so they are environmental to my container, not this PR. `spec_masteries` is the only PR-caused breakage.

## Minor, non-blocking

- The `cs_CZ` overlay still says "20 %" in the moonkin_form description (and carries the old flat "15" wording for shadowform). Contributors correctly never touch overlays, so nothing to change here, but flagging it so the release-time locale fill picks these up as reworded rows rather than leaving stale numbers in Czech tooltips.
- The branch is ~96 commits behind `release/v0.27.0`; worth a sync (or at least a merge check) before landing, though nothing in the base window conflicts with these files as far as I could see.

Verdict: request changes solely for the `spec_masteries` pins; everything else verified green.

ryan-foo:
@EnriqueGF the spec_masteries pins are fixed in 6b90defd6: the three exact-value pins now carry the trimmed authored values (frostbolt/frost_nova per-ability dmgPct 0.15, shadow as the dot-only specialization with Gloamveil carrying the general multiplier, balance Moonrage 0.1), each with a one-line note on why it landed there, and the AXES table in the second suite matches. Repro confirmed red-then-green on the exact three cases you named. Also merged the fresh release/v0.27.0 base (the ~96-commit drift), and re-proved spec_masteries, mastery_mechanism, ability_catalog_drift, signature_reworks, talent_tooltip_accuracy, architecture, the S3 guard, and parity on the merged head with tsc clean. The cs_CZ stale-number note is acknowledged for the release-time locale fill.

Rubsey (commented):
## OSSBrain initial review

### Should fix

- **New catalog drift guard fails on existing deliberate differences** (`tests/ability\_catalog\_drift.test.ts:39`)
  - Issue: The new numeric drift guard allows only cat\_form and ghost\_wolf, but a focused run shows 13 current abilities with existing numeric differences between catalog prose and ABILITIES prose, including demoralizing\_shout, overpower, blizzard, shadowform, pummel, and meteor.
  - Impact: The PR cannot pass the focused test or the broader gate with this test committed, so release integration would fail before validating the balance changes.
  - Suggested fix: Either narrow the guard to the specific abilities/number classes it can enforce today, or add justified exceptions for the existing intentional catalog differences so the test starts green and only fails on new drift.
- **Exception stale-check compares different value types** (`tests/ability\_catalog\_drift.test.ts:72`)
  - Issue: The stale-exception assertion compares a joined string from the catalog side to the raw string array from the def side, so it can pass even when the numeric tokens are identical.
  - Impact: Once exceptions are added or existing prose converges, this guard will not reliably tell maintainers to delete stale exceptions, leaving abilities permanently outside the general drift check.
  - Suggested fix: Join the def-side numeric tokens before comparing, matching the general guard's string-to-string comparison.

### Checks

- `npx vitest --config .ossbrain-review/pr-1919/vitest.review.config.ts --configLoader runner run tests/ability\_catalog\_drift.test.ts`: **failed**. Failed: the new guard reports 13 catalog-vs-def numeric drifts and expects none.
- `npx vitest --config .ossbrain-review/pr-1919/vitest.review.config.ts --configLoader runner run tests/mastery\_mechanism.test.ts tests/signature\_reworks.test.ts tests/spec\_masteries.test.ts`: **passed**. Passed: 3 files, 35 tests.

PR #1921, opened 2026-07-14
feat(render): replace 11 procedural props with Tripo-generated GLB models

jgyy:
## Summary

Replaces 11 hand-authored, code-generated (three.js primitive-geometry) props with real
Tripo-API-generated GLB models, following the shipped `gather_nodes.ts` pattern: each prop
loads a preloaded GLB and keeps its original primitive geometry only as a fallback for
pre-load races and headless test hosts. Every existing call signature is preserved, so no
caller needed to change.

| Module | Prop key | Old procedural shape | Credits |
|---|---|---|---|
| `src/render/delve_marsh_dressing.ts` | `marsh_plank_bridge` | 3 stacked boxes | ~50 |
| `src/render/delve_marsh_dressing.ts` | `marsh_shrine_fragment` | box slab + glow plane | ~50 |
| `src/render/delve_marsh_dressing.ts` | `marsh_corpse_candle` | cylinder + sphere | ~50 |
| `src/render/delve_marsh_dressing.ts` | `marsh_bell_fragment` | partial sphere | ~50 |
| `src/render/delve_marsh_dressing.ts` | `marsh_bell_gallows` (was `broken_bell_frame`) | 2 posts + beam + bell | ~80 |
| `src/render/delve_marsh_dressing.ts` | `marsh_sluice_post` | post + beam + rope cylinders | ~65 |
| `src/render/delve_marsh_dressing.ts` | `marsh_dead_tree` | trunk + 2-3 branch cylinders | ~65 |
| `src/render/delve_marsh_dressing.ts` | `marsh_reed_cluster` | 5 tapered cylinders | ~45 |
| `src/render/yumi_maze.ts` | `yumi_brazier_stand` | pedestal + bowl primitives | ~50 |
| `src/render/yumi_maze.ts` | `yumi_torch_handle` | wall bracket primitives | ~40 |
| `src/render/door_portal.ts` | `dungeon_door_arch` | arch + keystone + plinth (the portal swirl VFX disc stays procedural, it's animated per frame) | ~70 |

Total: **715 Tripo credits**

`marsh_dead_tree`, `marsh_reed_cluster`, and `marsh_sluice_post` trade their `hash2`-driven
per-instance random branch/stalk counts for one fixed authored model (a Tripo GLB can't
reproduce that procedural variety); every other prop was already effectively a single fixed
composite, so this is a clean 1:1 swap.

```mermaid
flowchart LR
    subgraph pipeline["scripts/asset_pipeline (Tripo AI 3D)"]
        A[prompt] --> B[Tripo generate]
        B --> C[normalize: y=0 base,\nWebP 512, meshopt]
        C --> D[qa gate:\nbudget + structural checks]
        D --> E["--apply:\nGLB + CREDITS.md + PROP_ASSET_DEFS"]
    end
    subgraph render["src/render/*.ts"]
        F[registerPreload\nloadGltf] --> G{loaded?}
        G -->|yes| H[clone GLB instance]
        G -->|no / headless test| I[primitive-geometry\nfallback]
    end
    E --> F
```

## Related issues

N/A (asset-quality improvement, not tied to a filed issue)

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands: `npx vitest run tests/render_glb_replacement_assets.test.ts`, `npm run gate`
  (tsc, biome on changed files, full test suite, all builds), `node scripts/asset_pipeline/pipeline.mjs qa --job <id>`
  per asset (structural + budget gate, PASS on all 11).
- Manual steps: reviewed every model's turntable preview PNGs from the pipeline job dirs
  before `--apply`; drove the real offline game client into the collapsed_reliquary door,
  the Protect Yumi maze, and The Drowned Litany delve to see the shipped props in their
  actual zone context (see screenshots below).

## Screenshots / recordings

**Real in-game captures** (offline client, live scene, via `scripts/tripo_prop_shot.mjs`):

`dungeon_door_arch` at the collapsed_reliquary entrance:

dungeon_door_arch in-game

`yumi_brazier_stand` and `yumi_torch_handle` in the Protect Yumi maze:

yumi_brazier_stand in-game
yumi_torch_handle in-game

The Drowned Litany marsh dressing (litany_ring module: blackwater pools, plank/rubble,
banners, and the marsh prop kit together in context):

marsh dressing 1
marsh dressing 2
marsh dressing 3
marsh dressing 4

The marsh delve module is randomly assigned per run (`litany_ring`/`litany_ledger`/`litany_sluice`/
`litany_apse`, all marsh-themed), so these 4 shots show the general dressing in context rather
than one isolated frame per individual prop key; "before" (pre-change) live captures were not
completed in this pass (time/scope), so these are after-only.

Clean per-asset in-engine renders (real GLTFLoader + meshopt output of each shipped GLB,
isolated on a turntable) remain committed as `after-<prop>.png` for a 1:1 visual reference
per key:

- `after-marsh_plank_bridge.png`, `after-marsh_shrine_fragment.png`, `after-marsh_corpse_candle.png`,
  `after-marsh_bell_fragment.png`, `after-marsh_bell_gallows.png`, `after-marsh_sluice_post.png`,
  `after-marsh_dead_tree.png`, `after-marsh_reed_cluster.png`, `after-yumi_brazier_stand.png`,
  `after-yumi_torch_handle.png`, `after-dungeon_door_arch.png`

---

### Quality

### Cross-platform

- [x] **Tested on desktop and mobile.** Render-only prop swap, no HUD/DOM change; verified
      the render module builds and the fallback path (used on all hosts, including mobile
      pre-load races) is unchanged and still covered by tests.
- [ ] Accessible. N/A, no UI change.

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
The pipeline work here is solid: the GLB-with-procedural-fallback shape matches `gather_nodes.ts`, the preload registration and `*PreloadInternalsForTest` exports are right, the manifest regen is clean, and `yumi_maze.ts` is the model of how to do this swap (the flames stay outside the branch, and the torch handle is re-seated and yawed to the wall side). The models themselves look good on the turntable renders. Two of them do not land right in the world, though, and one is the door every player walks through.

**blocking** `src/render/door_portal.ts:155` (`dungeon_door_arch.glb`): the arch is yawed 90 degrees away from the portal it is supposed to frame. The GLB's POSITION accessor spans x [-0.56, 0.56] and z [-1.85, 1.85] on a single identity-transform node, so its opening faces X. The procedural arch it replaces spans x [-2.1, 2.1] and extrudes only 0.7 in Z (`doorArchGeometry`), and the swirl disc, which correctly stays procedural, is a `CircleGeometry` in the XY plane at y 2.15, so it faces Z. The swirl now sits side-on to the archway instead of inside it, on every dungeon and delve entrance and exit. Your own capture shows it: in `ingame-dungeon_door_arch.png` the red portal disc is perpendicular to the arch. `inst.rotation.y = Math.PI / 2` on the clone (or a re-export with the opening on X) is all it needs.

**blocking** `src/render/delve_marsh_dressing.ts:61` (`marsh_plank_bridge.glb`): the bridge ships at roughly one eighth scale. The GLB's entire bounding box is 0.45 x 0.45 x 0.44 yd, while `addPlankBridge` draws three `BoxGeometry(2.4, 0.12, 0.45)` planks spread across 3.5 yd. `placeLoadedMarshAsset` clones and places without any rescale, so a bridge deck becomes a 45 cm trinket on the marsh floor. Every GLB in the set is normalized to an exact target height (5.00, 2.80, 2.10, 0.90, 0.55 and so on), so I suspect the bridge's target got taken from the plank's 0.45 depth rather than its 3.5 yd span.

The repo already has the fix for this whole class: `delve_props.ts` `buildStandaloneGlb(key, targetHeight)` does a `Box3` measure and rescales the loaded scene to the procedural prop's target height (mirroring `buildGlbChest`), then re-seats the base on the floor. Routing all 11 props through that pins each one to the footprint its layout was authored against. Worth doing beyond the bridge, because the height-only normalization has left several props much narrower than what they replace: `marsh_shrine_fragment` is 0.49 wide against a 1.4 slab, `marsh_bell_fragment` is 0.57 against a 1.1 sphere, and `marsh_sluice_post`'s crossbeam reads 1.08 in plan where the procedural beam is 1.5 to 1.9.

**should-fix** `src/render/delve_marsh_dressing.ts:400-408`: the corpse candle and the shrine fragment lose their glow. `addCorpseCandleProp` is a wax cylinder plus an additive flame sphere at y 0.62, and `addShrineFragment` is a slab plus an additive green glow plane at y 0.92. Both live inside the builder the GLB branch skips, and neither GLB carries an emissive factor or texture, so in a pitch-dark delve the corpse candles simply go dark. `yumi_maze.ts` keeps its flames outside the if/else and `door_portal.ts` keeps the swirl; the marsh props should do the same and place the GLB plus the flame/glow.

**should-fix** `src/render/door_portal.ts:156`: the cloned arch geometry is not marked shared, so `removeView` disposes it. Object views with no pool key take the traverse-and-dispose branch (`renderer.ts:4124`), which disposes every mesh geometry `isSharedGeometry` does not flag, and door views never get a pool key (`renderer.ts:3342-3350`). The procedural path marks the arch, keystone, plinth, and portal shared for exactly this reason ("Door portal resources are shared and prewarmed, so they must survive interest churn"). `clone(true)` shares its `BufferGeometry` by reference with `loadedDoorArchGltf`, with the prewarm bodies, and with every other door clone, so the first door that leaves interest range releases the GPU buffers out from under the doors still on screen and every clone built afterwards. Mark the loaded scene's geometry and materials once in the `.then` with `markSharedGeometry` / `markSharedMaterial`.

**should-fix** the 11 GLBs are not meshopt-compressed. Their `extensionsUsed` is `EXT_texture_webp` alone, where every shipped prop carries `EXT_meshopt_compression`, `KHR_mesh_quantization`, and `EXT_texture_webp`. `public/models/props/CLAUDE.md` makes that step mandatory ("Always compress with `--compress meshopt`"), and the PR summary says it ran, so it looks like the `gltf-transform optimize` pass got skipped on the way out of the pipeline. That is also why nine of the eleven land at 100 to 151 KB against the doc's 40 to 100 KB budget. Re-running it should pull the payload down without touching the code.

**nit** `src/render/delve_marsh_dressing.ts:36`: `marsh_bell_fragment.glb` (146 KB) is boot-preloaded but can never be placed. `kind: 'bell_fragment'` exists only in the `LitanyDressingAnchor` union in `src/sim/delve_litany_layout.ts:45`; no module authors one, so `case 'bell_fragment'` never fires (the bell now lives inside `marsh_bell_gallows.glb`, and `addBellFragment` only runs via the `broken_bell_frame` fallback). Drop the entry, or author the anchors it was meant to have.

**nit** `src/render/yumi_maze.ts:196`: the torch handle got a re-seat (`torchY - 0.72`) but the brazier did not. `BRAZIER_FLAME_Y` stays 1.9 with a `ConeGeometry(0.38, 1.0)`, so the cone's base sits at 1.4. The procedural bowl's rim was at 1.51, which tucked that base inside it; the GLB stand is 1.30 tall, so the flame now floats about 0.1 above the bowl.

**nit** shadows and materials: the base marsh dressing sets `castShadow` exactly zero times, and the GLB path turns on `castShadow` and `receiveShadow` for every mesh of every instance with no tier check. It is only a dozen or so props per module so the cost is small, but it is a shadow-map cost these props never had. Same shape for materials: the procedural builders threaded `flatShading: !GFX.standardMaterials` through every `surfaceMat` call, and the GLB clones use PBR plus three textures on every tier. The material half is how every other GLB module in the repo already behaves, so I would not change it here, just noting the low tier now pays for it.

**nit** boot payload: the GLBs total about 1.36 MB and all register at import, so `assetsReady()` blocks `startGame` on them for every player, and the glTF queue only runs two at a time on constrained devices. Ten of the eleven exist only inside one delve or the Yumi maze. `dungeon.ts` already has `ensureDungeonAssets()` as the lazy path if you want to keep the boot gate to the door arch.

**nit** `public/models/props/CLAUDE.md` is not updated: its Wiring section enumerates every GLB consumer, which now needs the three new modules.

**nit** `tests/render_glb_replacement_assets.test.ts` only asserts the URL resolves to a manifested file, so neither blocking defect above would have failed it. If you adopt the `buildStandaloneGlb` normalization, a per-prop target-height table gives the test something decisive to hold (assert the loaded `Box3` against it), and one assertion that every door mesh passes `isSharedGeometry` would cover the disposal finding.

**nit** `scripts/tripo_prop_shot.mjs` is checked in as a PR-specific one-off ("Bespoke capture for PR #1921"). Either generalize it into the shared capture tooling or leave it out of the commit.

The `hash2` variety trade on dead_tree, reed_cluster, and sluice_post is a fair call, and documenting it in the header comment was the right thing to do. A hashed per-instance yaw and scale jitter on the clone would buy most of that variety back for free if you want it.

jgyy:
Thanks for the thorough pass, especially catching the door arch on the very first play-through. Addressed all of it:

1. **blocking, door arch yaw**: `inst.rotation.y = Math.PI / 2` applied once on the loaded scene in the `.then()` (before any clone), so every clone inherits the corrected orientation. The swirl now sits framed inside the arch instead of side-on.

2. **blocking, bridge scale**: `placeLoadedMarshAsset` now does a `Box3` measure/rescale before seating, the same shape as `delve_props.ts`'s `buildStandaloneGlb`. Went further than just the bridge per your note: all seven marsh anchor GLBs (`plank_bridge`, `shrine_fragment`, `corpse_candle`, `broken_bell_frame`, `sluice_post`, `dead_tree`, `reed_cluster`) now rescale to the footprint their layout was authored against (span for the horizontally-authored ones, height for the vertical ones), instead of landing at whatever raw size Tripo exported.

3. **should-fix, candle/shrine glow**: split `addCorpseCandleProp`/`addShrineFragment` into a body builder (GLB-or-fallback) plus a flame/glow builder that's drawn unconditionally, mirroring how `door_portal.ts` keeps its swirl and `yumi_maze.ts` keeps its flames outside the GLB/procedural branch. Both light back up in a dark delve.

4. **should-fix, shared disposal**: the loaded arch's geometry and every material are now marked with `markSharedGeometry`/`markSharedMaterial` in the same `.then()` that applies the yaw, before any clone is taken. Confirmed via `tests/door_portal.test.ts` (still green) that the shared-marking path works the same way it does for the procedural arch/keystone/plinth/portal.

5. **should-fix, meshopt compression**: re-ran `gltf-transform optimize --texture-compress webp --compress meshopt` on all 11 GLBs this PR added. Verified with `gltf-transform inspect` that every file now reports `EXT_meshopt_compression, EXT_texture_webp, KHR_mesh_quantization` in `extensionsUsed`/`extensionsRequired` (previously `EXT_texture_webp` alone). Sizes dropped well under the 40-100 KB budget: 36-77 KB across all 11 (the previous 100-151 KB range is gone).

6. **nit, dead bell_fragment preload**: confirmed no dressing module anywhere authors a `bell_fragment` anchor (the bell now lives inside `marsh_bell_gallows.glb` / `addBrokenBellFrame`). Went with the smaller fix: dropped the GLB preload, the URL map entry, and the deleted `marsh_bell_fragment.glb` file itself (146 KB off the boot payload); the `bell_fragment` case in the anchor switch stays procedural-only for `LitanyDressingAnchor` exhaustiveness, since the union kind itself is sim-owned and out of scope here.

7. **nit, brazier flame seating**: re-seats the flame 0.1 lower when the GLB stand is present (`BRAZIER_FLAME_Y - 0.1`), matching the GLB stand's shorter 1.30 height vs the procedural bowl's 1.51 rim.

8. **nit, shadow/tier gating**: checked `delve_props.ts` and `gather_nodes.ts`, the sibling files that already clone GLB props into the world - both set `castShadow`/`receiveShadow` unconditionally true with no tier gate, same as this PR's marsh/yumi/door code. No existing tier-gating convention to match here, so left as-is per your own note; this is consistent with every other GLB-prop module in the repo, not a new pattern.

9. **nit, boot payload**: looked at `dungeon.ts`'s `ensureDungeonAssets()` for the lazy pattern you pointed at, but its own header comment is explicit that this codebase deliberately moved dungeon-kit GLBs INTO the boot preload specifically to kill an on-approach freeze at the Fallen Chapel ("without this the kit + Halloween-bits modules stream in ... the moment the camera nears a dungeon door"). Given that precedent, deferring the marsh/yumi GLBs to delve-entry risks reintroducing the same freeze on first approach, for no rebuild-on-load mechanism to backfill the fallback geometry once the real asset lands. Left this preloaded at boot; happy to revisit if the 1.36 MB actually shows up as a real regression in profiling.

10. **nit, CLAUDE.md wiring**: added the three new consumers (`door_portal.ts`, `delve_marsh_dressing.ts`, `yumi_maze.ts`) to the Wiring section in `public/models/props/CLAUDE.md`, following the existing rows' format.

11. **nit, tripo_prop_shot.mjs**: checked `scripts/` before deciding - this repo has well over a hundred committed one-off `*_shot.mjs` capture scripts (`shot_bleed.mjs`, `bag_rightclick_equip_shot.mjs`, `greyjaw_pet_tap_shot.mjs`, etc.), so a PR-specific capture script checked in permanently is the established convention here, not an exception. Kept it as-is rather than removing or generalizing it.

Verification: `npx tsc --noEmit` clean; `tests/architecture.test.ts`, `tests/render_glb_replacement_assets.test.ts`, `tests/door_portal.test.ts`, `tests/yumi_match*.test.ts`, `tests/yumi_maze_layout.test.ts`, `tests/litany_polygon_shell.test.ts`, and `tests/litany_spawn_collision.test.ts` all pass. Merged the latest `release/v0.26.0` in cleanly (4 commits, no conflicts).

One honest gap: I did not recapture in-game screenshots for this round (arch orientation, bridge scale, candle/shrine glow) - ran short on time/tooling budget this pass. The existing captures in the PR predate these fixes, so they should not be read as proof of the new behavior; happy to add a follow-up recapture if that is a blocker before merge.

FernandoX7 (approved):

PR #1922, opened 2026-07-14
fix(pvp): show Honor rewards in chat and character sheet

ryan-foo:
## Summary

- mirrors Honor award messages into the main chat feed while retaining Combat tab history and floating combat text
- mirrors ranked Arena victory, defeat, and draw rating lines into the same chat feed
- shows the current spendable Honor balance in the character-sheet header
- adds regression coverage for both chat routes and the character-sheet balance

After a paid Arena victory, chat now presents the reward and result together:

> You gain 25 Honor (Arena victory).
> You defeated Yycaruz in the Ashen Coliseum. Rating 1,955 (+2).

## Why

Honor and ranked Arena results were routed only to the Combat tab. Players who missed the floating Honor text had no reward confirmation in normal chat, and the current Honor balance was visible only while shopping with FURY.

## Validation

- 80 focused Honor, character-window, floating-text, and live-region tests passed
- `tsc --noEmit`
- Biome check completed with only pre-existing warning-level diagnostics in `hud.ts`
- `git diff --check`


Rubsey (commented):
Reviewed at 7dc49ef. Solid small change: `honor` was already on the `IWorld` seam (`src/world_api/duel_arena.ts:148`) with both worlds implementing it, the header reuses the existing `hudChrome.warfare.balance` and `honorGain` English keys instead of minting new ones, the CSS lands beside its `char-title-portrait` siblings using the `--gold` and `--spacing-xs` tokens, and the new tests anchor on case labels that are unique in `hud.ts`. Merge against release/v0.26.0 is clean.

Findings:

- should-fix `src/ui/hud.ts:8978` and `src/ui/hud.ts:9591`: routing the same line through both `this.log()` and `this.combatLog()` makes screen readers hear every Honor gain and Arena result twice. `log()` announces via `#chat-live` (`appendLog`, hud.ts:10609) and `combatLog()` pushes the same text into the combat announcer (hud.ts:10560). The codebase's own announce contract forbids exactly this (see the comment at hud.ts:10632, which is why the xp float is never routed to the combat announcer: its `log()` line already announces). Keeping both visual panes is fine; just pick one announce path, for example append the combat-pane line via `appendLog(this.combatLogEl, text, color)` directly (or give `combatLog` an announce opt-out) and let `#chat-live` carry the announcement, matching the xp precedent.

- nit `src/ui/char_window.ts:196`: the header balance only refreshes on reopen, a language switch, or `onInventoryChanged` (hud.ts:12209). An Honor award landing while the sheet is open leaves the header stale (spending stays fresh because the purchased item triggers the inventory path). A `this.renderCharIfOpen()` in the `honor` case would keep it live and is already open-gated.

Non-blocking note: the two new tests are source-text scans, which matches this repo's existing guard-test style, and I verified their anchors (`case 'honor':`, `case 'arenaEnd':`, `case 'yumiStatus':`, `case 'levelup':`) each occur exactly once in `hud.ts`, so they will not silently slice the wrong region.

ryan-foo:
@rubsey thanks, both addressed in c693241c8.

**should-fix (double announce):** the Honor and Arena lines now announce exactly once. The chat pane keeps `this.log(...)` (which announces via `#chat-live`), and the combat pane is now a silent visual mirror via `this.appendLog(this.combatLogEl, text, color)` directly instead of `this.combatLog(...)`, so the combat announcer no longer speaks the same line. This follows the xp-float precedent and the announce contract you pointed at. Both visual panes are unchanged.

**nit (stale char sheet):** the `honor` case now calls the open-gated `this.renderCharIfOpen()`, so an Honor award landing while the sheet is open refreshes the header balance live, matching the `onInventoryChanged` path that already keeps spending fresh.

Updated `tests/honor_ui.test.ts` to pin the new shape: `appendLog` to the combat pane, no `combatLog(...)` on either line, and `renderCharIfOpen()` in the honor case. `tsc` clean; `honor_ui`, `live_region_politeness`, `client_shell`, `char_window`, and `combat_announcer` all green.

Rubsey (commented):
Re-reviewed at c693241c8. Both items from the last pass are fixed cleanly, and the fix does not reintroduce the double-announce it set out to remove.

should-fix (a11y double-announce): fixed. The Honor line and each Arena result now go through `this.log(...)` (which announces once via `#chat-live`, since `log` -> `appendLog(this.chatLogEl, ...)` -> `announceChatLine`), and the combat pane is a silent visual mirror via `appendLog(this.combatLogEl, ...)`. `appendLog` only announces when `el === this.chatLogEl` (hud.ts:10619) and never pushes to the combat announcer, so the announcer no longer speaks these lines. That matches the xp-float precedent called out at showSelfNote (hud.ts:10640). Verified the `showBanner` mirror on the Arena case is not a third announce path: `#banner` has no aria-live, so it is visual only.

nit (stale char sheet): fixed. The honor case now calls `this.renderCharIfOpen()` (hud.ts:8987), the same open-gated refresh `onInventoryChanged` uses, so an award landing while the sheet is open updates the header balance live.

Tests pin the new shape well: honor_ui.test.ts asserts `log(honorMessage)` present, `appendLog(this.combatLogEl, honorMessage)` present, and `combatLog(honorMessage` absent (same for the Arena line), which is exactly the single-announce invariant. Ran honor_ui, char_window, live_region_politeness, and combat_announcer locally: 56 passing. `tsc --noEmit` clean.

Merges clean into the current base (release/v0.26.0 at c16a1e5d0, now including #1860 and #1869); merge-tree reports no conflicts, and #1860's hud.ts change does not touch the honor/arena handlers. Approved the pending fork CI runs for this head.

Good to go from my side.

jgyy (commented):
Small, well-scoped fix. It mirrors Honor-award and ranked-Arena-result lines into the main chat feed (single screen-reader announce) while keeping a silent visual copy in the Combat pane, and adds a live Honor balance to the character-sheet header. All amounts route through `formatNumber`, both new strings reuse existing English `t()` keys so there's no i18n gap, and the Honor value is the existing server-mirrored `world.honor` field, not client-computed.

nit: the new tests (`tests/honor_ui.test.ts`, the `char_window.test.ts` additions) are source-text substring checks against the handler string rather than executed-behavior assertions, so they wouldn't catch a refactor that changes the announce-once guarantee at runtime while leaving the source text intact. That said, this matches the established convention for `hud.ts`/`char_window.ts` tests elsewhere in the suite, so it's not a deviation, just a known limitation of the house pattern.

ryan-foo:
Refreshed onto release/v0.27.0 and retargeted (4758b03fc); the merge was fully clean with zero conflicts, and honor_ui, char_window, live_region_politeness, combat_announcer, and architecture suites pass (80 tests) with tsc clean. Rubsey's two earlier findings remain addressed as re-verified in the second review (single announce path via chat-live with the combat pane as a silent visual mirror, and the open-gated renderCharIfOpen refresh); jgyy's note on the source-scan test style stands as the acknowledged house pattern.

PR #1923, opened 2026-07-14
fix(items): crafting materials are white (common), never vendor junk

ryan-foo:
## Summary

Crafting materials are now white (common quality) instead of vendor junk. Any item used as a recipe reagent must be `common`, not `poor`: the "sell junk" sweep (`items.ts` `sellJunk`) vendors every `quality: 'poor'` item, so a grey crafting material gets auto-vendored the moment a player clears junk.

Three reagents (`bone_fragments`, `linen_scrap`, `spider_leg`) were `poor` and hit exactly that. The rest were tagged a rarity colour (rare/epic/uncommon), which read as loot rather than as a stackable material.

This sets `quality: 'common'` on every `kind: 'junk'` reagent across the crafting recipes (COMMON/TOOL/COMBO) and the enchanting recipes:
`bone_fragments`, `linen_scrap`, `spider_leg`, `thorium_ore`, `arcanite_bar`, `ashwood_log`, `elderwood_log`, `goldleaf_herb`, `sunpetal_herb`, `arcane_essence`.

Out of scope by design:
- Gathering tools (`kind: 'tool'`) keep their tier rarity, they are equipment, not materials.
- Genuine vendor trash (junk not used in any recipe) stays `poor`, so the "sell junk" button still has something to sweep.

The material tier now lives in `buyValue`/`sellValue`, not the rarity colour. Sell/buy values are unchanged, so the economy is untouched; only the tooltip colour and the junk-sweep eligibility change.

## Type of change

- [x] Bug fix (crafting materials auto-vendored as junk)

## How was this tested?

- New `tests/crafting_materials_quality.test.ts` guard: every crafting/enchant reagent resolves to a real item, none is `poor`, and every `kind: 'junk'` reagent is `common`. This catches a future material added as grey or coloured.
- Updated the two tests that pinned the old colours: `market_filters` (bone_fragments moved from the `poor` filter to `common`) and `professions_crafting_hub` (reagents are now `common`; the price tier is pinned off `buyValue` and the 4x markup instead of the quality colour).
- Ran the professions, market, salvage, bag/bank filter, loot, item-level, progression, shipped-item-id, guide, and i18n suites; `tsc` clean; biome clean. No parity or wiki change (materials do not surface in the guide).

## Notes

`arcane_shard` (the epic/legendary disenchant yield) is left `rare`: it is not currently a reagent in any enchant recipe, so it is out of the "used in recipes" rule and is not junk-swept anyway. If a future enchant consumes shards, the new guard test will flag it.

---

**Silent behavior changes surfaced in review (both welcome):**
- **Closes a deed cheese:** buying a reagent off Bree marks discovery at the static def quality, so a 60c Thorium Ore used to grant `col_first_rare` ("Something Blue") and a 160c Arcanite Bar `col_first_epic` on the spot. With reagents retagged common, a trade good can no longer be your first blue or epic.
- **Resolves the gathering contradiction:** `MaterialRarity` excludes poor by contract ("a harvested material is never junk-grade") while the harvest table granted exactly the three items that were poor; the retag makes the item defs agree with the contract, and a gathering contract's stated rarity now matches the item's color.
- Review follow-up: Arcane Essence stays `uncommon` (the disenchant ladder tiers dust/essence/shard colors deliberately); only `poor` was ever swept by Sell Junk.


FernandoX7 (commented):
The premise checks out. `sellAllJunk` (`src/sim/items.ts:443`) filters on `def.quality === 'poor'`, and all three grey reagents are live drops (`spider_leg` `zone2.ts:254`, `bone_fragments` `zone3.ts:516`, `linen_scrap` `dungeons.ts:47`), so they really were being vendored out from under players the moment anyone hit Sell Junk. Real bug, right shape of fix. I also like that the guard reads `ALL_RECIPES` + `ENCHANTS` instead of a hand-listed set of ids, so the next material added as grey fails in CI rather than in game.

I checked the blast radius of moving quality and it is mostly clean: the six hub reagents are Bree's vendor stock (`zone3.ts:1098`), not loot, so `effectiveItemLootStrategy` (`src/sim/loot/loot_roll.ts:130`) never sees them, and the three greys map to `commonItems` whether they are `poor` or `common`, so no party-loot routing moves. Salvage and disenchant eligibility is weapon/armor only, so materials never entered it. Quality is derived at read everywhere I looked (`market_query.ts:133`, `bank_view.ts:98`), so nothing persisted goes stale. Professions, market, loot, deeds, and bag/bank suites are green here.

Two side effects you did not claim, both of which I like and which are worth a line in the description since they are silent behavior changes:

- It closes a deed cheese. `buyItem` goes through `ctx.addItem`, which marks discovery with no rolled quality (`src/sim/sim.ts:5749`), so `markItemDiscovered` falls back to the static def quality (`src/sim/deeds.ts:467`). Buying a 60c Thorium Ore off Bree therefore used to mark `quality:rare` and hand you `col_first_rare` ("Something Blue", `content/deeds.ts:1042`) on the spot, and a 160c Arcanite Bar did the same for `col_first_epic`. Both are cosmetic (renown 0), but a trade good should not be your first blue, and after this it is not.
- It resolves a standing contradiction in gathering. `MaterialRarity` is `Exclude<quality, 'poor'>`, commented "a harvested material is never junk-grade" (`src/sim/professions/gathering.ts:85`), while `NODE_HARVEST_TABLE` grants exactly the three items that were `poor` (`gathering.ts:72-74`). The retag makes the item defs agree with the contract.

**should-fix** `tests/professions_crafting_hub.test.ts:224-227`: the rewritten pin no longer catches a price-tier change, which matters because this PR makes the price the only thing carrying the tier. `expect([60, 160]).toContain(def.buyValue)` accepts either tier for any reagent, and `expect(def.sellValue).toBe(def.buyValue / 4)` derives the second value from the first, so both hold for any self-consistent pair. I demoted `arcanite_bar` from 160/40 to 60/15 and the suite stayed green; the same mutation against the base test fails with `arcanite_bar buyValue: expected 60 to be 160`, because the old assertion pinned the price to the item's own quality. Pin the two prices per item (an id to [buyValue, sellValue] map) so the tier stays guarded. The comment above the loop ("Price pins (literals, not derived)") is no longer accurate either, and the 4x markup sentence is now stated twice.

**should-fix** `src/sim/content/items.ts:1372`: `arcane_essence` should stay `uncommon`. It fixes no bug, since the sweep only takes `poor` and an uncommon essence was never at risk. The three arcane materials are a deliberate ladder keyed to what you disenchanted: dust for common/uncommon, essence for rare, shard for epic/legendary (`src/sim/professions/enchanting.ts:72-78`). Flattening the middle rung leaves dust and essence identical white at 6c and 18c while `arcane_shard` stays blue, so the color now tracks nothing in particular. It also moves away from the classic ladder rather than toward it: dust white, essences and shards green, top crystal blue. The tags you inherited were closer than the new ones. The ore, bar, log, and herb flattening I am happy with, trade goods are white in classic and the tier reads off the name and the price.

**should-fix** `tests/crafting_materials_quality.test.ts:29-38`: the third case asserts more than the bug needs. The invariant that maps to `sellAllJunk` is "no reagent is `poor`", and the second case already covers it. The blanket "every `kind: 'junk'` reagent is common" rule is what forces the essence retag above, and it is a trap for the next enchant: `arcane_shard` is exempt today only because nothing consumes it, so the first recipe that takes a shard turns this red and the fix on offer will be to paint the shard white too. I would drop the third case and keep the poor check.

**nit** `src/sim/content/items.ts:617` and `tests/crafting_materials_quality.test.ts:1`: both comments cite `items.ts sellJunk`, but the function is `sellAllJunk` and it lives in `src/sim/items.ts`, not `src/sim/content/items.ts`. Worth correcting, the anchor rule is what keeps these comments findable later.

Two notes, neither needs action:

- `col_junk_drawer` ("Discover 10 different poor-quality items", `src/sim/content/deeds.ts:1293`) loses three greys from its pool, 25 down to 22, so it stays completable and earned deeds are never revoked. The meter recomputes live (`src/sim/deeds.ts:679`), so an in-progress player can drop by up to 3, though `deedProgress` returns null for meter triggers so nobody watches a counter tick backwards. The three you are retagging are the widest-dropping early greys, so it is a small difficulty nudge. Fine by me, just flagging it.
- CI has not run on this head sha at all (no check runs, status pending), so the green results in the description are local-only so far.

jgyy (commented):
Clean, correctly scoped data fix. Confirmed the bug against release/v0.26.0: thorium_ore, arcanite_bar, ashwood_log, elderwood_log, goldleaf_herb, and sunpetal_herb were tagged rare/epic, arcane_essence was uncommon, and bone_fragments/linen_scrap/spider_leg were poor, which put them straight into the vendor-junk auto-sell path. White (common) is the right classic-MMO convention for stackable trade-good reagents, this corrects mistagging rather than inventing a new rarity rule. Cross-checked every reagent referenced across ALL_RECIPES and ENCHANTS against the diff: all ten are covered, and the one deliberately-left-rare item (arcane_shard) isn't consumed by any current recipe or enchant, self-documented in the PR notes.

nit: `tests/professions_crafting_hub.test.ts` loosens one pinned assertion from an exact epic/other split to a buyValue/4 ratio; still bounded by the surrounding [60, 160] range check above it, so not a real regression in coverage.

No blocking issues. All 24 tests pass, merge-tree against release/v0.26.0 is clean.

ryan-foo:
All findings addressed, in 2799d5403 (base merge) + d43157600 (fixes), and the PR now targets release/v0.27.0. Arcane Essence goes back to uncommon: the disenchant ladder (dust for common/uncommon, essence for rare, shard for epic/legendary) tiers its material colors on purpose, and since only quality 'poor' is swept the essence was never at risk; the items.ts header now spells that out. The hub-reagent price test pins each reagent to a literal [buy, sell] pair so a tier change on any single item reds (verified the mutation you ran now fails), the quality guard keeps only the invariant that maps to sellAllJunk (no recipe or enchant reagent is 'poor') with the blanket junk-is-common case dropped, and both stale comment anchors now cite sellAllJunk in src/sim/items.ts. Also folding your two silent side effects into the description: the retag closes the col_first_rare/col_first_epic vendor-buy deed cheese, and the reagent colors now agree with the MaterialRarity contract in gathering. Crafting/professions/enchanting suites (270+ tests), deeds_content, progression, and localization_coverage green; tsc clean.

FernandoX7 (approved):

Issue #10, opened 2026-06-12
Missing features


STAlfsen:
The idea of a collaborative open source MMO created by the players is kinda cool, but I can't spend my precious Fable tokens on it right now. In any case, from testing the version hosted on https://worldofclaudecraft.com/ for 5 minutes on a laptop, there are a few glaring things that are missing, and I thought I'd just start a TODO list for tracking purposes:

More detailed chat features:
- Separation of chats into /say, /yell, /whisper /general  
- Hovering chat bubbles above the heads of players when using /say or /yell 
- /say having a range of a few meters, /yell with larger range
- Personal chat messages and combat logs instead of global broadcasts for everything

Sound effects:
- Duel sound effect
- Bag open close sounds effects
- All kinds of sound effects

Other:
- Bag contents being updated the moment a trade is finished (currently you have to close and open it before traded items become visible)

Bugs:
- My character glitches a bit back and forth when running forwards / jumping forwards

That's all I can be bothered to think of for now.

It would be absolutely hilarious to have a Fable bot that tracks issues on the Github, creates pull requests, and then have an in-game system (claude-themed deity npc?) where there is some kind of a timed vote on whether it should be merged. It would be absolute chaos and the token bill would be eye watering for someone, but it genuinely is kind of a novel gaming concept. The self-building MMO. An actual "living world", where you try to get your buds together on a tuesday night to vote for a pull request that nerfs rogues.

slonce70:
Took a pass at most of this list — opened #11.

**Chat**
- Split into `/say` (short range), `/yell` (longer range), `/w <name>` whisper, `/g` general, and the existing `/p` party. say/yell only reach players in range; whisper is private with a sender echo; general is world-wide. Added a per-player rate limit so the global/whisper channels can't be spammed.
- Hovering chat bubbles over heads for `/say` and `/yell` (yell is styled differently).

**Sounds**
- Duel now has audio (challenge horn, countdown ticks, start gong, end), plus bag open/close and a whisper chime.

**Other**
- The "traded items only show up after reopening the bag" bug is fixed — an open bag window now refreshes the moment a trade completes.

**Bug**
- The forward-running/jumping jitter was the client restarting snapshot interpolation from the stale server pose every update; it now re-anchors at the pose actually on screen, which removes the rubber-banding.

Everything's unit-tested + checked against a live server, and a review pass flagged a few issues in my first cut (including a WS hardening one) that are fixed in the PR.

[...]

PR #1924, opened 2026-07-14
feat(pvp): Frostreach Frontier Season 1 (WIP draft)

ryan-foo:
# Frostreach Frontier: Season 1 (WIP draft)

An always-on, factionless frost PvP zone far east of the world (x >= 14000), with a public spawn event, a hero-points economy, open PvP, and a hub vendor. Full spec: the Frontier Phase 2 document in #1759 (`docs/prd/frontier-pvp.md` after that PR's rename; `docs/prd/frontier-pvp-honor.md` on the current base). Implements the plan in #1759 on top of the #1817 honor foundation.

## Scope in this PR
- **Zone + open PvP.** The band geometry (single source, `src/sim/pvp/frontier.ts`), the `isHostileTo` open-PvP clause (all-enemies outside the safe hub), and the hero-points currency (soulbound, persisted).
- **Enter from the PvP window**, like Fiesta/Arena: a Travel-to / Return-from-the-Frontier control (`frontierEnter`/`frontierLeave`), teleport to the safe hub with a remembered return spot.
- **The Frontier Incursion.** A shared, Greater-Rift-style meter that fills from trash kills, PvP kills, healing, and a passive public timer; at 100% a hard, unkitable rare spawns and the zone converges. A top-screen incursion bar shows the meter, then flips to the rare's name and HP.
- **The rares** (Rimefang Stalker, Frostbound Revenant): unkitable group content (2.2x move, ccImmune, slowImmune) tuned for 2 to 3 players plus a healer. Killing one pays honor + hero points to every contributor, healers included.
- **Hero-points vendor + the item-level 31 Frostrend set**, and a Frontier Marshal honor daily (100 honor).

## Determinism
The whole incursion is player-gated and draws zero rng, so the parity goldens are untouched; new NPCs spawn at reserved ids after the world roster; everything is soulbound.

## Status: draft, changes requested
Status after @EnriqueGF's review: every blocking item is now implemented on this branch. The band is enterable from the PvP window (frontierEnter/frontierLeave on the IWorldDuelArena facet, both worlds, level-gated), the incursion spawner makes the rares and hero points earnable, priceHero reaches the vendor UI (VendorPrice, affordability, the hero-balance line, with a vendor_view pin), the hero-points wallet has its full client read path (IWorld member, both worlds, hero/lhero snapshot fields, HUD event), the safe hub floors player-vs-player damage to zero including DoT ticks and in-flight projectiles (apply-outside/walk-in/tick test), and the two pvp_frostreach deeds are authored. Of the should-fix list: the party exemption in the band, the wired player-kill honor, the spendHeroPoints purchase routing, the pet-return hook ordering, the progression pin, and the entry level gate are all in. Deferred to a follow-up, deliberately: a formal per-day DR on rare, player-kill, AND wisp rewards (the Rimebound Wisp honor trickle, FRONTIER_TRASH_HONOR at one respawn per 6s, is the same uncapped economy, so it joins that follow-up rather than getting bespoke throttling here; the incursion spawn cadence is the current throttle) and the cosmetic honorMod barrel-alias rename. Before/after screenshots land with the next visual pass. The branch is merged onto release/v0.27.0.

Parity note: five parity goldens were regenerated deliberately in an earlier commit, markArenaEntered serializes arenaDaily into the state hash so the hashes moved while the rng draw counts stayed byte-identical; the latest review-fix commits moved no goldens.


EnriqueGF:
Thanks for taking on the Frontier :) The geometry and currency modules are clean, well commented, genuinely module-first, and the determinism story is airtight, so this is a good foundation to build on.

**Verdict: changes requested.** Not production-ready yet.

Base-branch gate: PASS, this PR targets `release/v0.26.0`, which is correct.

What I ran locally rather than trusting the description: `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/parity` (green with unmodified goldens, so the reserved-id vendor spawn and the new `MOBS`/`ITEMS` rows really do not perturb the rng stream), `tests/world_api_parity.test.ts`, `tests/guide.test.ts`, plus `frontier_pvp`, `progression`, `sim`, `honor`, `duel`, `items`, `item_level`, `pvp_safety`, `pvp_honor_gear`, `vendor_view`. All green, and the committed i18n generated bundles are fresh. The problem is not what the tests cover, it is that every test reaches the sim directly and so none of them touch the surfaces the feature is missing.

## Blocking

1. **The Season 1 set can never appear in any vendor window.** `src/ui/vendor_view.ts:71-76` builds each price from `buyValue` and `priceHonor` only, then drops any row where `price.copper <= 0 && price.honor <= 0`. Every Frostrend piece is priced with `priceHero` alone (`src/sim/content/frontier_vendor.ts:35-168`), so all nine are filtered out and the Frostreach Quartermaster opens an empty list, offline and online. `grep -rn priceHero src/ui src/net server` returns nothing. `tests/frontier_pvp.test.ts:177` calls `sim.buyItem()` directly, which is exactly why the suite is green while the feature is unreachable. `priceHero` needs to reach `VendorPrice`, the `affordable` check, the hero-balance line in `src/ui/vendor_window.ts`, and a `vendor_view` test that pins a hero-priced row as listed and its affordability against a hero balance.

2. **Hero points never reach the online client.** `heroPoints` appears nowhere in `server/` or `src/net/`. The self snapshot ships `maybe('honor', ...)` / `maybe('lhonor', ...)` (`server/game.ts:4806-4807`) with no hero counterpart, `src/net/online.ts:2120-2121` decodes only the honor pair, no `IWorld` facet member was added (`src/world_api/duel_arena.ts:148-150` carries `honor`/`lifetimeHonor`), and the HUD has no handler for the new `heroPoints` `SimEvent` (`src/ui/hud.ts:9140` has `case 'honor'` and nothing for hero points, so a grant is silent: no float, no combat-log line). Persistence itself is fine, since the server saves the whole `serializeCharacter` blob (`server/game.ts:2792`, `3269`), so this is purely a read path gap. Either land the facet plus both world implementations plus the `IWORLD_MEMBERS` pin in this PR, or hold the vendor and set commits back until the wallet lands.

3. **A DoT still kills you inside the "safe" hub.** `src/sim/pvp/frontier.ts:20` promises "PvP damage never lands inside `FRONTIER_HUB_RADIUS`", but the hub is enforced only in `isHostileTo`. `src/sim/combat/auras.ts` contains no hostility check at all and its periodic tick calls `ctx.dealDamage` unconditionally (`auras.ts:171`), so a bleed or DoT applied out in the band keeps ticking after the victim reaches the hub and can finish them off at the vendor. In-flight projectiles (`src/sim/projectile_travel.ts`) have the same gap. The repo already has the fix pattern (Vale Cup floors player-vs-player damage to 0 in `combat/damage.ts`); do the same on `inFrontierHub(target.pos)` and pin it with a test that applies a DoT outside the hub, walks in, ticks, and asserts HP unchanged.

4. **Nothing in this PR is reachable in a real world.** No camp spawns the two frost rares (`grep -rn rimefang_stalker src/sim` hits only the definition), so `awardFrontierRareKill` can never fire and hero points cannot be earned. There is also no entry path into the band: `FRONTIER_ENTRY` has zero consumers and no teleport, portal, or command puts a player past x=14000. So the Quartermaster stands alone at x=14040, selling items the UI would not list, for a currency the world cannot grant. Taken with items 1 and 2, the vendor and set commits are shipping ahead of the surfaces they depend on.

5. **Missing Book of Deeds records.** CLAUDE.md requires every new piece of conquerable content, rares included, to author its deeds records in `src/sim/content/deeds.ts` in the same change (`docs/design/deeds.md`). This PR adds two rares and no deeds. `tests/deeds_content.test.ts` passes only because it pins the existing catalog, so the gap is silent.

## Should fix before merge

6. **Your own AoE hits your party inside the band, and your healer cannot heal you.** `isFriendlyTo` is defined as `!isHostileTo` for players (`src/sim/sim.ts:6515`) and `hostilesInRadius` collects any non-dead entity where `isHostileTo` holds (`src/sim/sim.ts:4594-4600`), so with no party clause in the Frontier arm (`sim.ts:6476-6486`) party members inside the band are mutually hostile: heals, shields, and resurrects are rejected, and every AoE is friendly fire. That directly contradicts `src/sim/pvp/frontier_rewards.ts:3`, whose whole premise is "a tapping party is all rewarded". Whichever way you decide (full FFA or a party exemption), it needs to be a deliberate, tested decision, because it silently changes how every healer spec plays in the zone.

7. **Dead scaffolding shipped as live surface.** `FRONTIER_KILL_HONOR_MULT`, `FRONTIER_DAILY_HONOR`, `FRONTIER_ENTRY`, `frontierDepth` and `spendHeroPoints` have zero consumers, and the `frontier_kill` / `frontier_daily` `HonorReason` arms are declared (`src/sim/types.ts:55-57`) and mapped to HUD labels (`src/ui/hud.ts:730,732`) but never emitted. The visible consequence: a Frontier player kill currently awards no honor at all, so the headline mechanic of the zone pays nothing. Either wire these up here or leave them out of the type union until the slice that emits them. `src/sim/CLAUDE.md` is explicit that a member with zero consumers is dead scaffolding.

8. **`buyItem` bypasses its own spender.** `src/sim/items.ts:369` hand-rolls `meta.heroPoints -= heroCost` after an inline balance check instead of calling `spendHeroPoints`, which is the one function that owns the "does the wallet cover it, mutate only if so" rule. The tested spender is not the live one; route the purchase through it.

9. **The rare payout has no diminishing returns.** `awardFrontierRareKill` grants flat honor plus hero points to every contributor on every rare kill in the band, outside the honor DR the #1817 foundation established, and a contributor is paid even if they tagged once and left, or are dead. Rares respawn. Once camps land this is an uncapped farm; worth a cap or DR (and a test) before the spawn slice ships.

10. **Fragile hook ordering.** `awardFrontierRareKill` is called at `src/sim/combat/damage.ts:778`, above the owned-pet early return at 788, and gates purely on position (`isFrontierPos(mob.pos.x)`), not on "is a Frostreach frost rare". Not exploitable today (taming rejects rare/elite), but any owned or future rare in the band would pay out. Move the call below the pet return, or add an `ownerId !== null` guard.

## Smaller notes

11. `tests/progression.test.ts:139` weakens the "NPC inside world bounds" pin by skipping any NPC in the band. Reasonable, but it leaves the Quartermaster's placement pinned by nothing. Asserting `inFrontierHub(npc.pos.x, npc.pos.z)` instead would restore the coverage and catch the case where `findSafePos` (rng-free, so determinism is intact, confirmed) pushes the vendor out of the safe zone onto PvP-enabled ground.

12. No level gate on band hostility: any player at any level at x >= 14000 is flagged, though the content and the set are level-20 cap content. If a gate is intended it belongs in the hostility clause or the entry teleport.

13. In `sim.ts` the band predicates are reached as `honorMod.isFrontierPos` / `honorMod.inFrontierHub`, which reads as if band geometry lived in the honor module. Worth renaming the barrel alias in a follow-up.

14. The PR description is stale: it lists the frost rare templates, the Quartermaster, the ilvl-31 set, and the item-level registration as deferred, but all of them are in these three commits. Please resync it so the scope statement can be trusted.

15. This adds a new NPC and two new mob models, so the screenshot requirement applies: before/after captures (desktop and mobile) under `docs/screenshots`, referenced from the PR body. Also a style nit, the PR body carries an emoji, which the contribution rules exclude from PR text.

Happy to re-review as soon as the vendor and wallet path is closed, or the vendor and set commits are split out from the foundation. The core is in good shape; almost everything left is at the seams.

ryan-foo:
Thanks for the thorough review, this is exactly the seam-level scrutiny the feature needed and the critique is fair. Two commits that close the biggest gap landed after your pass, so let me start there and then walk the rest of the list honestly.

**On #4 (nothing is reachable), resolved since your review.** Two follow-up commits landed:
- **Entry surface.** You now enter the band from the PvP window itself, the same way you queue Fiesta or the Arena, via `frontierEnter`/`frontierLeave` (the `IWorldDuelArena` facet, both worlds, the `frontier_enter`/`frontier_leave` commands, jail-blocked, with a persisted return spot). So there is a real teleport past x=14000.
- **The Incursion spawner.** The rares are no longer parked in the world; they spawn from a public, player-gated, Greater-Rift-style meter (`src/sim/pvp/frontier_incursion.ts`) that fills from trash kills, PvP kills, healing, and a passive timer. So `awardFrontierRareKill` now fires in a real world and hero points can be earned. Still zero-rng and parity-clean (the phase no-ops with an empty band).

That makes the vendor and set reachable in principle, but #1 and #2 are still the gate on them being usable, and you are right on all of these. The plan:

**Blocking, still open:**
1. **`priceHero` in the vendor UI.** Correct, `vendor_view` drops hero-only rows. I will thread `priceHero` through `VendorPrice`, the affordability check, and the hero-balance line in `vendor_window`, with a `vendor_view` test pinning a hero-priced row as listed and its affordability against a hero balance.
2. **Hero-points read path.** I will land the `IWorld` currency member + both world implementations + the snapshot self-field + the HUD `heroPoints` event handler + the `IWORLD_MEMBERS` pin. The incursion commit already added exactly this shape for `frontierIncursion` (`maybe('fincur', ...)` to `applySnapshot` to a facet read), so the wallet follows the same recipe.
3. **Safe-hub DoT/projectile.** Right, the hub is only enforced in `isHostileTo`. I will floor player-vs-player damage to 0 on `inFrontierHub(target.pos)` (the Vale Cup pattern), covering periodic auras and in-flight projectiles, with the apply-outside/walk-in/tick test you described.
5. **Deeds.** I will author the rares' records in `deeds.ts`.

**Should-fix:**
6. **Party friendly-fire in the band.** Good catch that `!isHostileTo` makes party members mutually hostile, which contradicts the "a tapping party is all rewarded" premise in `frontier_rewards.ts`. I will make it a deliberate, tested decision (a party exemption for heals/shields/AoE, FFA otherwise).
7. **Player-kill honor + dead scaffolding.** Correct that a Frontier player kill currently pays no honor: the incursion meter progresses on a kill, but `FRONTIER_KILL_HONOR_MULT` is still unwired. I will wire the player-kill honor grant and either consume or drop the remaining unused symbols.
8. **`buyItem` spender.** Will route the purchase through `spendHeroPoints` instead of the inline `meta.heroPoints -= cost`, so the tested spender is the live one.
9. **Rare-payout DR.** Agreed, before the farm exists. I will add a cap/DR and a test.
10. **Hook ordering.** I will move the `awardFrontierRareKill` call below the owned-pet early return and gate on "is a Frostreach frost rare", not just position.

**Smaller:** restore the `inFrontierHub` assertion in the progression pin (11), decide the level gate (12), rename the `honorMod` band alias (13), resync the description (14, done), and drop the emoji + add before/after screenshots (15, emoji removed).

I will push these as follow-up commits and re-request review. Thanks again, the "every test reaches the sim directly so none touch the missing surfaces" point is the right diagnosis.

ryan-foo:
Follow-up: the blocking items and most of the should-fix list are now pushed to this branch (rebased onto a fresh `release/v0.26.0` merge, so the conflicts are cleared too).

- #1 `priceHero` reaches the vendor UI (threaded through `VendorPrice`/affordability/`vendor_window`, with a `vendor_view` test pinning a hero-only row listed + affordable).
- #2 hero-points read path: `IWorld` `heroPoints`/`lifetimeHeroPoints`, both worlds, the `hero`/`lhero` snapshot self-fields, and a HUD `heroPoints` event (float + combat-log).
- #3 safe-hub DoT/projectile floor: `dealDamage` floors player-vs-player damage to 0 in `inFrontierHub`, with the apply-outside/walk-in/tick test.
- #5 deeds: two `pvp_frostreach_*` deeds on a new `frontierRareKills` stat.
- #6 party exemption in the band; #7 player-kill honor now paid (`FRONTIER_KILL_HONOR_MULT` wired); #8 `buyItem` routes through `spendHeroPoints`; #10 `awardFrontierRareKill` moved below the pet return and gated on the frost-rare id; #11 the progression pin now asserts `inFrontierHub` for the Frontier NPCs; #12 a level gate on entry.

Deferred with a note in the commit: #9 (a formal per-day DR on rare rewards; the incursion spawn cadence is the current throttle) and #13 (renaming the `honorMod` barrel alias, cosmetic). Happy to fold those in if you'd like them in this PR rather than a follow-up. Re-requesting review when you have a moment.

jgyy (commented):
Reviewing this as the WIP draft it's labeled: the merge-conflict scope is actually small (the real conflict is just the generated i18n pending slice, a rebuild-and-regen fix, not a logic conflict), and the parts I could spot-check are careful: the new sim files (frontier.ts, frontier_entry.ts, frontier_incursion.ts) all derive their positions deterministically (pid-based, no rng/wall-clock), and the two new Book of Deeds records for the zone's kill/complete milestones are present and cosmetic-only, as required.

The PR description looks stale relative to the actual diff, though: the "remaining blocking items" section lists things (priceHero vendor UI, hero-points client read path, safe-hub damage guard, deeds records) that all appear to already be implemented in the current diff. Worth refreshing the description so reviewers aren't chasing already-closed gaps.

should-fix: at 92 files / ~5825 additions, this is far beyond what can be reviewed line by line in one pass, even though the feature itself (one zone, one reward loop) is architecturally coherent. Splitting into a few smaller PRs (zone geometry/entry, incursion/rares, hero-points economy + vendor UI, and the i18n regen noise separately) would make this reviewable in the normal sense, and would also isolate the mechanical i18n churn from the actual logic.

nit: `inFrontierHub` is checked at three separate call sites (direct-damage floor, an honor-calc guard, and the incursion/rare-kill path) rather than centralized once at the damage-eligibility layer; low risk since each is a defensive floor, but worth consolidating if a fourth site appears.

Given the draft status, I'd hold off on a deeper line-level pass until it's split and out of draft; the direction and the determinism/deeds work look solid so far.

ryan-foo:
Description resynced (the remaining-blocking-items paragraph was stale; everything in it had landed) and the branch is now merged onto release/v0.27.0 and retargeted, all suites green on the merged head. On splitting: keeping it one PR deliberately, since the seams have already been through one full review cycle here (EnriqueGF's list) and the i18n churn is isolated in its own commits; if a follow-up slice grows past this, it will land as separate PRs. The inFrontierHub consolidation nit is noted, and the third call site gets centralized the moment a fourth appears. The stale doc path in the body is also fixed (it now points at the #1759 Phase 2 spec by its real name).

EnriqueGF (changes requested):
Reviewed the full diff at db350e1c7 against release/v0.27.0. The structure is right where it matters: the band geometry has one source, the incursion runs as a zero-rng player-gated phase behind the SimContext seam, the IWorld additions land on the duel_arena facet with both-world implementations and updated pins, hero points persist with normalize-on-load, and the new sim behavior is well tested. What I ran locally: `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, the full `tests/parity` golden suite, `world_api_parity`, `command_schema`, `snapshots`, `deeds_content`, `progression`, `vendor_view`, `arena_window`, `entity_roster`, `sim_context`, `hud_perf_budget`, `localization_coverage`, `arena`, `honor`, and all four new frontier/arena-daily suites. All green except one `snapshots.test.ts` case ("key lockstep") that fails on Node 20 because `WebSocket` is not a global there; I reproduced the identical failure on the base branch, so it is environmental, not yours.

Findings, verified against the code (not the description):

**Blocking**

1. `src/sim/pvp/frontier_entry.ts:28` (frontierEnter): no live-match or instance guard. A player seated in a live arena/fiesta match can send `frontier_enter` during the countdown or any out-of-combat lull (the only gates are dead / already-inside / level / inCombat, and countdown players are not in combat). I reproduced it with a test: queue two players, one tick to seat them, `frontierEnter(a)` succeeds, the player escapes the ranked match, and `meta.frontierReturn` records the arena-slot coordinates, so a later `frontier_leave` teleports them back into the arena band, potentially into another live match's slot, and the record persists across relogs. Every queue command guards `ctx.arenaMatches.has(pid)` (arena.ts:115, vale_cup.ts:424); frontierEnter needs the same, plus a dungeon/delve-instance position check, and frontierLeave should validate the return spot before using it.

2. Biome `ci:changed` fails on this branch: unsorted imports in `tests/entity_roster.test.ts:6` and `tests/sim_context.test.ts:9` (the new `createFrontierIncursionState` import landed above `scan_counters`). CI gates changed files on errors, so this is red as-is. `npx @biomejs/biome check --write` on those two files fixes it.

**Should-fix**

3. `src/sim/pvp/frontier_incursion.ts` (trashSpawnPoint): a trash spawn can land inside the safe hub, contradicting the function's own "clear of the safe hub" comment. For spawn counter n with n % 12 == 6 and n % 5 == 4 (n = 54, 114, ...), the point is (FRONTIER_MUSTER.x - 108, 0) = (hub.x + 32, 0), inside FRONTIER_HUB_RADIUS 34. I verified the arithmetic numerically. A Rimebound Wisp (aggroRadius 14) then spawns in the hub and attacks shoppers; the hub damage floor in damage.ts only covers player-vs-player, not mobs. Cheapest fix: reject/adjust candidate points with `inFrontierHub`.

4. `src/sim/pvp/frontier_incursion.ts` (frontierIncursionOnHeal): the reward-roster hook credits the healer for healing ANY player anywhere in the band while a rare is up, including self-heals and heals at the safe hub; there is no check that the target is fighting the rare (the PRD says "healing anyone who is fighting the rare"). Stand at the hub, take a scratch, self-heal once during a rare, and you collect the full 15 honor + 3 hero points. The existing test even exercises this shape (the healed "Tank" is not on the rare). It needs a target-on-the-rare's-threat/roster gate, and probably a source !== target exclusion for the roster credit.

5. Online daily-quest state drift: a daily never enters `questsDone` and `PlayerMeta.dailyQuests` never crosses the wire (grep: no `dailyQuests` in server/ or src/net/), and `ClientWorld.questState` (src/net/online.ts:2291 via optimisticQuestState) has no daily/utcDay awareness. So after an online turn-in, once the snapshot clears the pending-command optimism, the client shows the Marshal's daily as available again for the rest of the day while the server's questState says done and accept is refused. The two IWorld implementations disagree, which is exactly the tri-host parity rule. The done-today bit (or the dailyQuests record) needs to reach the client.

**Notes**

- The PR body says "the parity goldens are untouched", but five goldens were regenerated. The regen is legitimate (markArenaEntered now serializes `arenaDaily` into the state hash; I checked the rng draw counts are byte-identical), but the body should say that instead.
- `serializeCharacter` now emits `arenaDaily: {enteredDay:'', claimedDay:''}` for any offline character who ever entered a bout; normalizeArenaDaily coerces it back to undefined on load, so save/load/save is not byte-stable. Harmless, slightly unclean.
- The wisp honor trickle (FRONTIER_TRASH_HONOR 2, one respawn per 6s) is an uncapped ~20 honor/min faucet; the deferred per-day DR note in the PR body covers rares and player kills, but the trash trickle is the same economy and worth folding into that follow-up.
- `hudChrome.warfare.heroGain` says "hero points" lowercase where heroFloat/heroBalance say "Hero Points". Nit.

Once 1 and 2 are in (3 to 5 could ride this PR or an immediate follow-up given it is a draft), this looks close.

ryan-foo:
Both blockers and all three should-fixes are in, through 3663c1326. frontierEnter now carries the arenaMatches guard every queue command uses plus the DUNGEON_X_THRESHOLD instance gate, and frontierLeave validates the persisted return spot before using it (finite, overworld-side) with a fallback to origin; tests reproduce your exact seated-match escape and the poisoned-return teleport. The biome import order in the two test suites is fixed and ci:changed exits 0. Trash spawns now deterministically step clear of the hub (pure function of the counter, no rng; pinned by a 0..600 sweep incl. the n=54 case you computed), heal roster credit requires the healed target on the rare's threat table and excludes self-heals (the off-rare Tank case now asserts no credit), and the daily done-today bit crosses the wire via a dailyq snapshot self-field following the same recipe as fincur/hero, so both IWorld questState paths agree online (snapshot pins updated). The notes are done too: serializeCharacter omits the all-empty arenaDaily record (byte-stable save/load/save, normalize-on-load kept) and heroGain says Hero Points. Parity goldens did not move this round; tsc, the snapshot pins, and the i18n gates are green. The PR body is updated with the corrected goldens statement and the wisp-trickle DR follow-up note.

ryan-foo:
Re-merged onto release/v0.30.0 (939 commits absorbed) and retargeted; full local gate green (all 11 steps, 1601 test files, typecheck, all builds). Notes for reviewers:

- Parity goldens: exactly the 5 drifted files (arena_1v1, arena_2v2_wipe, fiesta x3) were re-minted; all others byte-identical.
- v0.30's committed-item-art rule postdates this branch, so the 9 Frostrend item icons are provisional frost-tinted recolors of same-slot catalog art (provenance under sourcePack woc_frost_recolor in public/ui/items/mapping.json). Happy to swap in real art before merge if preferred.
- The 3 frontier mob portraits were rendered through the standard portrait pipeline, not placeholders.
- Behavior calls kept from each side: daily quests count every completion toward questsCompleted (frontier), non-daily repeatables count first completion only (release); hub PvP damage floor applies before defensive-stance reduction and Ignite (release).

ryan-foo:
The provisional recolor icons are replaced with purpose-generated painted art (f859117af): gpt-image batch, one consistent frosted-steel treatment, converted through assets:items, provenance updated to sourcePack woc_generated plus a CREDITS row.

| helm | spaulders | hauberk | girdle | legs |
|---|---|---|---|---|
| helm | spaulders | hauberk | girdle | legs |

| gauntlets | sabatons | choker | band |
|---|---|---|---|
| gauntlets | sabatons | choker | band |

PR #1925, opened 2026-07-14
docs(prd): spec the enchanting loop (disenchant to materials, enchants per slot)

ryan-foo:
## Summary

A PRD for making the Enchanting profession playable: disenchant gear you do not need into arcane materials, then spend them to apply one always-known stat enchant per slot.

The important finding is that the enchanting **engine and data already exist and are unit tested**, they are just not reachable by a player. `disenchantItem` / `applyEnchant` / the full `ENCHANTS` table (one enchant for every equip slot) / the arcane materials / the `recalcPlayerStats` fold are all built behind the `SimContext` seam, but there is no `COMMAND_NAMES` token, no `IWorld` facet method, no `dispatchMessage` case, and no UI. So this is almost entirely a wiring spec.

The doc (`docs/prd/enchanting.md`) covers:
- The player-facing flow and v1 constraints (one enchant per item, always known, holder-only).
- Exactly what already exists (with file hooks) so nobody rebuilds it.
- The wiring requirements: two wire commands, the `IWorld` seam + `ClientWorld` mirror, server dispatch with validation + localized feedback, a Disenchant context-menu entry, an Enchanting panel on the crafting window, and an enchant tooltip line.
- One content addition: a shard-tier enchant so the epic/legendary disenchant yield (`arcane_shard`) has a sink (today nothing consumes it).
- Acceptance criteria, non-goals (weapon-damage enchants, procs, recipe learning, salvage), open questions, and the hook-point index.

Slot coverage is already complete (all 11 equip slots have an enchant), so no new per-slot enchants are needed beyond the shard tier.

## Type of change

- [x] Documentation (PRD, no code)

## How was this tested?

Doc-only change. Verified no em/en dashes or emojis (repo copy rule). No sim/server/UI code touched, so no test or wiki impact.


FernandoX7:
I like this idea. My suggestion is to keep the game more open-world and not let the user do everything at their fingertips. 

They should go to a new building with a new alchemy NPC and work there to disenchant equipment.

jgyy (commented):
Unusually well-grounded PRD. It correctly identifies that the enchanting engine, data, and tests already exist and are simply unwired, and every hook-point citation I spot-checked (function names, deny-reason strings, the entity stats-fold line, sim.ts fields, the missing COMMAND_NAMES/dispatch/UI entries) matches the real code exactly. No em/en dashes or emojis, follows the docs/prd/ convention (status table, goals/non-goals/design/edge-cases/acceptance-criteria).

should-fix: Section 2's hook-point index (R2, deny reasons) lists `not_disenchantable`, `not_held`, `wrong_slot`, `insufficient_materials`, `unknown_enchant`, but `resolveDisenchant` also has an `unknown_item` deny reason that's missing from the list. Minor completeness gap.

informational: PR #1950 is open against the same base and already implements this PRD's R6 (the arcane_shard dead-end fix) in full, with its own Greater-tier shard-consuming enchants and a table-integrity test. If #1950 merges first, Section 6/R6 becomes stale framing (describes forward-looking work that's already done). Not blocking for this PRD since it's a spec, but worth a one-line pointer to #1950 once it lands, and the two PRs should be sequence-checked by whoever merges them.

ryan-foo:
All three folded into the PRD (17f5eb423): R2's deny-reason list now includes unknown_item (both resolveDisenchant and resolveApplyEnchant return it), R6/Section 6 cross-links PR #1950 as already delivering the shard-tier enchants with a merge-time sequencing note, and Section 9 records @FernandoX7's suggestion (anchor disenchanting to a dedicated enchanting NPC building to keep the game open-world, rather than anywhere-at-fingertips) as an explicit open design question for the owner. Merged release/v0.27.0 in cleanly; the PR now targets it.

PR #1887, opened 2026-07-14
feat(sfx): define the idle mob voice action, land reptile's clip

jamiecypher:
## Summary

Stacked on #1873 (reptile family) and #1874 (hurt action), so this PR's own change is just the last commit, `911165b83`; the rest is those two waiting to land first.

Registers `idle` as a fifth recognized mob vocalization action (`aggro`, `attack`, `death`, `hurt`, `idle`) so the pipeline is ready for idle sfx as each family's recording is finished, without wiring any trigger logic yet. That trigger, and the crowd rate-limiting so a cluster of the same mob type doesn't spam idle barks, is a deliberate follow-up once more families have idle clips ready.

`idle` is opt-in per family, not required like the other four: `mob()` in `scripts/sfx/sfx_prompts.mjs` takes an optional 7th argument and only emits a `mob_<family>_idle` catalog entry when a family actually supplies one. Every other family keeps its existing 4 actions unchanged, no idle prompt required, no missing-file build error introduced for families that aren't ready yet. `reptile` is the only family with idle wired end to end in this change: prompt text, the conformed clip in `public/audio/sfx/`, and the regenerated manifest. Nothing in `src/ui/` or `src/game/` calls it.

## Type of change

- [x] Feature: new functionality

## How was this tested?

- `npm run sfx:check` / `npm run sfx:conform` on the new clip (source was -24 LUFS, conformed to spec -14 LUFS, 192kbps/44.1kHz).
- `npx vitest run tests/sfx_manifest.test.ts tests/sfx.test.ts tests/sfx_conform.test.ts tests/combat_sfx.test.ts tests/architecture.test.ts tests/guide.test.ts` — 165 passing. Updated the two pinned-count guard tests (`MOB_ACTIONS.size`, the mob-key/`SFX_FIXED_CATALOG_KEYS` counts) to reflect the new total; both exist specifically to catch silent catalog drift, so they correctly failed until bumped.
- `npx tsc --noEmit` clean.
- `npm run gate` — PASS, all 11 steps green.

## Screenshots / recordings

N/A, no player-visible or UI change; the clip isn't triggered anywhere yet.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [x] Builds succeed.

Cross-platform
- ~Tested on desktop and mobile.~ N/A, no UI/render change.

Localization (i18n)
- [x] N/A. No player-visible strings.

Hygiene
- [x] No secrets or .env committed.
[...]

PR #1927, opened 2026-07-14
fix(sfx): include idle in SFX_MOB_EXTENSION_KEY_PATTERN

jamiecypher:
## Summary

`MOB_ACTIONS` gained `idle` in #1887, but `SFX_MOB_EXTENSION_KEY_PATTERN` (the stricter grammar check `parseMobSfxVariantStem` applies right after the action-token scan) was never updated to match: it still hardcoded `(aggro|attack|death|hurt)`. A subfamily-level idle recording, e.g. a dedicated wolf idle take distinct from the beast family default, finds `idle` fine via `MOB_ACTIONS`, but then fails the key-pattern check every single time, since nothing else in the pipeline could ever produce a matching key.

Found this while wiring up a local preview build with a fuller idle set. The family-level catalog path (`mob()` called with an idle prompt, the only idle usage that has shipped so far, reptile) never exercises this second check, so the gap shipped invisibly until a subfamily idle take was actually attempted.

## Type of change

- [x] Bug fix

## How was this tested?

- Added a decisive reproduction test (`adds a subfamily key from mob_<family>_<sub>_idle_N.mp3`), confirmed it fails against the old pattern before the fix (`invalid mob sfx key derived from filename`).
- Updated the pinned grammar-string test to the corrected pattern.
- `npx vitest run tests/sfx_manifest.test.ts tests/sfx.test.ts tests/sfx_loading.test.ts tests/combat_sfx.test.ts tests/architecture.test.ts`, 74+ passing.
- `npx tsc --noEmit` clean.
- `npm run gate`, PASS, all 11 steps green, run in isolation.

## Screenshots / recordings

N/A, build-time validation only, no visual or gameplay change.

---
Checklist

Quality
- [x] Tests pass.
- [x] Typecheck passes.
- [x] Builds succeed.

Cross-platform
- ~Tested on desktop and mobile.~ N/A, no UI/render change.

Localization (i18n)

Hygiene
- [x] No generated files hand-edited (`src/game/sfx_manifest.generated.ts` regenerated via `npm run sfx:manifest`).


FernandoX7 (commented):
Verified this end to end and it holds up. The two-check structure is exactly as described: `parseMobSfxVariantStem` finds `idle` via `MOB_ACTIONS`, then the composed key fails `SFX_MOB_EXTENSION_KEY_PATTERN` (`reason: 'key'`), so a subfamily idle take could never land. Confirmed the bug on `release/v0.26.0` and that #1887 introduced the action. Good find, and the repro test is the right kind: I reverted the pattern locally and it fails with exactly `invalid mob sfx key derived from filename: mob_beast_wolf_idle_1.mp3`, then passes with the fix.

Also checked at the PR head:

- `node scripts/build_sfx_manifest.mjs` regenerates `src/game/sfx_manifest.generated.ts` and the runtime pack with zero diff, so the generated artifact is fresh.
- The runtime validator (`src/game/sfx_runtime_pack.ts`) rebuilds its regex from the generated `SFX_MOB_EXTENSION_KEY_SOURCE`, and `scripts/sfx/manifest.mjs` re-exports the builder pattern rather than copying it, so builder, pack loader, and manifest share one source of truth. I swept the tree for other copies of the action grammar; this change leaves none stale in code.
- All four sfx test files pass (78 tests), and biome is clean on the changed files.

Two nits:

- nit: `docs/sfx-studio-tutorial.md:195` still says a subfamily action "must be `aggro`, `attack`, `death`, or `hurt`". This PR widens that grammar, so the sentence should gain `idle`.
- nit: the PR body has a few em dashes; house style bans them in PR text too (CLAUDE.md conventions). The code and commit message are clean.

One scope note, not a finding: nothing plays idle mob voices yet. `MobVoiceAction` in `src/ui/combat_sfx.ts` covers the four combat actions, and even the family-level `mob_reptile_idle` from #1887 has no playback caller, so a subfamily idle take will be discovered, packed, and loaded but stays silent until idle playback is wired separately. That matches the build-time-only framing here; flagging it so nobody expects audible behavior from this alone.

CI has not run on this PR yet (both workflows are waiting on maintainer approval), which is why the merge state shows as unstable.

FernandoX7 (approved):

Issue #1832, opened 2026-07-12
Server calls botDetector.setTrackingConnection that the bundled detector does not implement


FernandoX7:
## Problem

The production server bundle deployed on 2026-07-12 calls `this.botDetector.setTrackingConnection(...)`, but the bot detector implementation bundled into that image does not have the method. Every call throws `TypeError: this.botDetector.setTrackingConnection is not a function`. The global kept-alive handlers swallow it, so the process stays up, but the throw aborts the enclosing function mid-flight each time.

In the deploy running from 12:59 UTC to 18:57 UTC the error fired 1,218 times (roughly 200 per hour), starting 20 seconds after boot. Three call paths are affected:

- `GameServer.resumeSession` (via `join` from `authenticateWebSocket`): surfaces as an unhandledRejection, so a reconnecting player's session resume is interrupted partway through.
- `GameServer.socketClosed` (WebSocket close handler): surfaces as an uncaughtException, so disconnect cleanup after the call site is skipped.
- `GameServer.socketClosed` via `pingLiveSessions` (the linkdead reaper): same effect when the reaper closes a dead session.

This is the bot detector stub-versus-private API drift class: the server's call sites and the two detector implementations (the public no-op stub and the private detector) have to expose the same interface, and here the shipped implementation is behind the call sites.

## Steps to reproduce

1. Build the server at the revision currently deployed to production, with the bot detector implementation that image bundled.
2. Connect a game client over WebSocket, then disconnect it (or reconnect to resume a session).
3. Observe the failure in the server log.

## Actual behavior

- `TypeError: this.botDetector.setTrackingConnection is not a function` is logged as a kept-alive uncaughtException or unhandledRejection on every session resume, socket close, and reaper-driven close.
- Cleanup and resume logic after the `setTrackingConnection` call site does not run for the affected connection.

## Expected behavior

- The server never calls a detector method the bundled implementation lacks: the no-op stub and the private detector both implement the full interface the server code uses, in the same change that adds a call site.
- A missing method on either implementation is a compile-time error (both implementations typed against one shared detector interface), not a production runtime TypeError.
- Session resume and disconnect cleanup run to completion regardless of which detector implementation is bundled.

## Scope

- The bot detector interface seam (`server/bot_detector/`) and the `setTrackingConnection` call sites in `GameServer.resumeSession` and `GameServer.socketClosed`.
- The public stub, and the matching change in the private detector repo so both copies stay in API sync.
[...]

PR #1930, opened 2026-07-14
fix(server): production stability hardening for reconnect, brownouts, admission, and ops

### LINKED ISSUE #1833: Sim tick stalled for six minutes in production: mob.update climbed to 98ms mean per tick

FernandoX7:
## Summary

Production stability hardening for the live realm, motivated by the July world-loop stall incidents (#1833) and the detector-drift outage (#1832), and by the crash-resilience (WS2) and ops-hardening/deploy-runbook (WS5) workstreams of PERFORMANCE-SCALING-PLAN.md.

What ships, by area:

- Stall diagnosability and the bounded mob fix (#1833): observer-only mob-scan counters (aggro-scan visits, threat-table entry visits) and per-mob-family mob.update buckets, surfaced in the PERF_TICK_LOG heartbeat and the admin tick capture, so the next stall is attributable from logs alone. The idle aggro-scan grid query narrows from a literal 25 yd to MAX_AGGRO_RADIUS (20 yd) at both scan sites: behavior-identical (detection was already clamped at 20), parity green with zero re-baselines. A co-located-crowd repro harness (scripts/mob_stall_repro.mjs, three scenarios) is committed for regression use; measurements up to 80 co-located clients stay well inside the tick budget with no superlinear growth.
- Client reconnect hardening: full-jitter exponential backoff (src/net/backoff.ts) so a mass disconnect no longer reconnects in lockstep, a 0 to 1000 ms spread on the foreground-retry path, and the server auth-timeout rejection classified transient (bounded at 20 attempts, reset on hello) so a database brownout no longer dumps players to the login screen.
- Server brownout hardening: pg pool timeouts (5 s connect, 15 s statement default, 65 s driver backstop), a 60 s allowance for the known heavy reads and both character-save paths, boot DDL on a dedicated non-pool client, an idle pool error handler, and the handshake converting a database rejection into the existing retryable timeout frame instead of an unhandled rejection.
- Same-account lease takeover: character_leases gains a nullable account_id (the one DDL change, additive and idempotent); a reconnect for the same account takes over its own lease immediately instead of waiting out the TTL after a crash, with NULL rows failing closed, nonce-fenced saves so a displaced session can never overwrite live state, and the displaced session kicked with the existing takeover signal.
- Keepalive sweep self-clocked: one event-loop stall re-arms every session instead of mass-terminating them.
- Realm admission cap: MAX_PLAYERS_PER_REALM (default 5000, explicit 0 disables), checked server-side before join with an in-flight counter closing the concurrent-handshake race; refusals send a localizable "realm is full" frame (FATAL to the reconnect scheduler), resumes and admins stay exempt, and refusals log as one aggregated line per 30 s window. The per-IP hard-limit refusal now also sends a localizable error frame instead of a raw close.
- Ops hardening and the deploy runbook: /livez reports a wedged world loop (30 s staleness, drain-exempt, cold-boot backstop) and /readyz flips during drain; the compose game service gains a healthcheck, stop_grace_period 75s, mem_limit 5g with memswap_limit 5g, and a NODE_OPTIONS heap-cap passthrough; deploy/game_watchdog.sh restarts the container only on Docker-reported unhealthy (lock, 300 s cooldown, dry-run); DEPLOY.md's update runbook now pulls the private bot-detector clone and runs a containerized tsc drift gate before any image build (the #1832 guard; the production hotfix shipped earlier, this makes a recurrence fail loudly before it deploys), plus a by-hand Caddy retrofit hiding /livez /readyz /metrics from the public edge.
- i18n: new English keys loading.realmFull and loading.tooManyConnections with their five non-Latin fills; the realm-list population tooltips reworded to the shipped bands (five non-Latin overlays refreshed).

Maintainer note for the release fill: the three reworded realm.popTip keys (Medium/High/Full) are stale in every Latin overlay and will NOT appear on the pending worklist (reword staleness), so the release-time fill must re-translate those three keys across all locales; the two new loading.* keys are pending normally.

## Related issues

Closes #1833. Adds the regression guard for #1832 (stub interface-completeness test plus the runbook tsc gate; the production hotfix already shipped).

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands: `npm run gate` green end to end (i18n gen and freshness, malware scan, changed-files biome, SFX conformance, the full 14k-test suite, tsc, all builds). Targeted batteries across the touched suites (parity, architecture, mob determinism and counters, lease trio, ws_auth, backoff and visibility, health, keepalive, admission cap, realm status, deploy ops and watchdog exec harness, i18n S3 and M16 guards).
- Manual steps, all against a disposable local compose stack built from this branch:
  - Realm at cap (cap=1): second join refused with the realm-full frame in 6 ms, same-account resume admitted at cap, fresh join admitted the moment the slot freed.
  - Database brownout (paused Postgres): login rejected with the retryable timeout frame at the 65 s driver backstop while the in-world client streamed ~20 Hz snapshots unaffected for the full 74 s pause; clean recovery after unpause.
  - Wedged process (SIGSTOP): Docker health flipped unhealthy in 70 s while the process stayed running (invisible to restart policies); the real watchdog script restarted it through the full 75 s grace and it recovered healthy; the restart stamp held the cooldown.
  - Mass disconnect (SIGKILL with 6 connected clients): all sockets dropped together; after restart every same-account reconnect was readmitted in under 25 ms (no lease TTL wall).
  - Reconnect spread (three real browser clients, server SIGKILLed): first retries at 531/878/1541 ms, a 1010 ms spread inside the jitter band where the old code fired in lockstep; all three resumed.
  - Repro harness at 20 and 80 co-located bots: mob.update mean 0.64 ms and 0.39 ms per tick respectively, zero over-budget ticks, counters attributing (aggro visits scale with the crowd, threat visits stay zero idle).

---

### Quality

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: no layout, style, or render change (copy-only player-visible changes on existing surfaces).
- ~Accessible.~ N/A: no UI structure added or edited.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Every user-facing

### Hygiene


Rubsey (commented):
This is a big surface and the load-bearing claims all check out against the code. Things I verified independently rather than taking from the description:

- The 25 to 20 idle-scan narrowing is provably behavior-identical: both sites clamp with `Math.min(MAX_AGGRO_RADIUS, ...)` and detect strictly (`d < radius`), `delveDetectMult` returns 1 or 0.65, and the stealth multiplier is capped at 1.0, so nothing detects past 20. The shrink-only premise is itself pinned by tests, so a future boosting modifier reds the suite instead of silently missing players in the 20 to 25 band.
- The admission cap's in-flight counter is conserved on every fresh-arm exit (the check-then-increment pair has no await between it; the single finally covers join success, lease refusal, and thrown DB error), the resume and staff exemptions are server-derived (hasSessionForCharacter and the staff row, never client input), and the concurrent last-slot race is pinned.
- The lease same-account arm fails closed on NULL rows (plain equality, never IS NOT DISTINCT FROM), account_id is stamped from the authenticated account behind the getCharacter ownership gate, the nonce rotation rides the same atomic statement, and the fence rides the save UPDATE itself. The nasty second-tab case (a fresh join of a character live in this same process) is short-circuited at ws_auth.ts:277 before any acquire, so a doomed handshake can never rotate a live session's nonce out from under it, and the error-path release is guarded the same way. Shutdown keeps saveAll ahead of releaseAllCharacterLeases, so fenced shutdown saves still land.
- The watchdog and healthcheck are well-engineered: flock in a root-owned dir, cooldown stamped before the restart, restart-by-name so a half-edited compose file cannot sneak in config, and Docker health as the single source of truth. The deploy tests execute the probe and the watchdog against fake docker/flock harnesses instead of only grepping strings.
- The test suite deliberately defeats the constant-self-comparison trap (unexported wire literals re-pinned as bare literals in-test; the counter observer-neutrality proof compares full state including rng internals between an observed and an unobserved sim).
- The repro harness hard-refuses non-loopback targets, and the pg-connection-string resolution does close the `?host=` bypass as claimed.

Merge is conflict-free against the current release/v0.26.0 tip and CI, including the full-suite PR gate, is green. Thanks for the release-fill note about the three reworded realm.popTip keys not surfacing on the pending worklist; that is exactly the kind of staleness that needs a written record.

Findings:

- should-fix: DEPLOY.md:117 calls POST /internal/restart-countdown "loopback only", but nothing enforces loopback: server/internal.ts and require_internal_secret.ts gate on the secret header alone, and the Caddy @ops 404 block covers only /livez /readyz /metrics, so the endpoint is reachable from the public edge with the secret. Reword to secret-gated so nobody treats the secret as a low-stakes internal value.
- should-fix (process, flagging so the acceptance is deliberate): this stability PR also carries two repo-process changes: the CLAUDE.md commit-body mandate, and the whole database-performance-reviewer gate (agent definition, codex toml, dispatch rows across qa-gate.md, qa-checklist, four skills plus their .agents mirrors, pinned by tests/codex_setup.test.ts). I audited the content and it is clean (read-only rules, explicit never-touch-production-databases rails, a sane scope split against migration-safety, no gate weakened anywhere), so I am not asking for changes, but policy riders like these belong in their own PR where they can be discussed on their own terms. Please split them out next time.
- nit: src/net/online.ts:1375 the new "never two live timers" comment overstates the guarantee: socketClosed (line 1433) still assigns this.reconnectTimer without clearing a pending handle, so the zombie path's manual socketClosed followed by the real onclose can strand a live timer. The race is pre-existing, not this PR's regression, but a clearTimeout at the top of that scheduling would make the new comment true.
- nit: tests/deploy_admission_cap.test.ts:15 pins the MAX_PLAYERS_PER_REALM row against the whole compose file; the sibling deploy_game_ops.test.ts slices the game service block precisely because discord-bot runs the same image. Reuse the slice so the pin cannot pass with the knob on the wrong service.
- nit: deploy/game_watchdog.sh:124 has no timeout around docker inspect / docker restart; a hung docker daemon parks the run on the flock and every later minute exits with the skip line. A `timeout 55` on the two docker invocations bounds it cheaply.
- nit: DEPLOY.md:111 the tsc drift gate (npm ci plus tsc in a container) runs on the live 8 GiB box before the game stops, with no --memory bound on the gate container. A `--memory 2g` keeps a heavy tsc from creating the exact host pressure the new mem_limit exists to prevent.
- nit: server/ws_auth.ts:182 the realm-full log aggregation never flushes a burst's tail: counts accumulate silently until the next refusal after the window, which may be hours later, so incident-time counts read shifted. A trailing flush, or folding the pending count into the heartbeat, fixes it.
- nit: server/ws_auth.ts:278 the comment "a live or linkdead session in THIS process already owns the lease row" is transiently false after a cross-process same-account takeover rotates the nonce (the held session's row belongs to the other process until the fence-out kick lands, within one autosave). Behavior is fine, and strictly better than the base's silent unfenced clobber on the analogous flip-back; just update the comment.
- question: after the 20 tolerated auth-timeout rejections, the fatal overlay shows the raw English "authentication timed out" (the documented protocol-diagnostics passthrough). Given the two new capacity refusals got proper t() arms, is a localized sustained-outage message worth it, or do we keep the diagnostics exemption? Fine either way; the precedent is pre-existing.

FernandoX7:
Thanks for the thorough review. All findings are addressed and pushed; the branch is also re-synced with release/v0.26.0 (the merge brought PR 1937 and PR 1940 in; conflicts were the two degitted i18n aggregates, taken as deleted, plus both branches appending a test at the same spot in schema_wiring, both kept; artifacts regenerated on the merged tree, and a post-merge audit found no lost intent on either side).

Per finding:

- should-fix, DEPLOY.md loopback wording: reworded to secret-gated, stating the header is the ONLY gate and the edge hides only /livez /readyz /metrics. A follow-up QA pass caught the same stale claim in two sibling docs (.env.example and the compose env comment) plus the step-3 parenthetical calling the bare-host tsc run "the same gate"; all fixed (a11f41165, b3e8b446d). The accurate loopback comment about the 127.0.0.1 port publish is untouched. One deliberate non-change: adding /internal/* to the Caddy @ops block would be compatible defense-in-depth (the playbook curls loopback directly), but that is an edge-config change and felt like exactly the kind of rider this PR should stop accumulating; happy to do it separately.
- should-fix, process riders: acknowledged, and fair. The commit-body mandate and the database-performance-reviewer gate will ride their own PR next time.
- nit, online.ts reconnect timer: fixed, and the fix went one step further than the suggested clearTimeout. Clear-and-reschedule still burned two of the 40 bounded attempts per zombie drop and, at the attempt cap, the late duplicate onclose took the exhaustion branch: endSession cleared the just-scheduled final retry and dumped the player on one physical drop. A pending reconnect timer can only belong to the same drop (no socket is open while one is pending), so socketClosed now returns early on a duplicate signal instead. Both regression tests fail against the clear-and-reschedule variant too, not just against the original (452f3a9bb).
- nit, admission-cap compose pin: rescoped to the game service slice, with the boundary-marker guard from deploy_game_ops so the slice cannot silently degrade to a whole-file match (a11f41165).
- nit, watchdog docker timeouts: done, with two deliberate deviations from the flat 55s suggestion. The restart bound is 150s because a wedged process ignores SIGTERM and eats the full 75s stop grace, so a WORKING recovery routinely outlasts a minute and a 55s bound would misreport the exact restart the watchdog exists to issue; inspect is 55s to fit the cron minute. The correctness pass then hardened the bounds themselves: -k 5 KILL escalation (a TERM-immune docker CLI would otherwise wait forever and re-create the park-on-the-flock hang), operator 0 maps to the default (GNU timeout reads 0 as NO bound), and wrapper/exec failures (exit 125 and up) log loudly instead of folding into the quiet container-absent arm. Knobs are sanitized like the cooldown; defaults are pinned; the exec harness gained hang, sanitization, and zero-knob coverage (a11f41165, 70eafd622, 3383e2ec6).
- nit, drift-gate memory: bounded at --memory 2g --memory-swap 2g (swap bound matches the compose memswap pattern so the gate cannot push the host into swap either), the contiguous-block pin updated, and a runbook note added that exit 137 means the bound, not a type error (a11f41165, 3383e2ec6).
- nit, realm-full log tail: refusals inside the window now arm one unref'd trailing flush at the window edge, so a burst's tail lands at incident time. Folding into the heartbeat was rejected because PERF_TICK_LOG is opt-in and counts would vanish with it off. A second pass also closed the race you did not ask about: an inline window-edge flush now disarms the pending timer, so a stale timer can never flush the next window's first refusals early. Tests g2 and g3 cover both, g3 mutation-checked (7cf648186, 0d30e0fd5).
- nit, ws_auth lease comment: corrected to say the session usually still owns the row but a cross-process same-account takeover may have rotated the nonce already, with the fence-out kick landing within one autosave; a sibling test comment stating the absolute premise was updated too (7cf648186, 0d30e0fd5).
- question, localizing the sustained auth-timeout overlay: keeping the diagnostics exemption, deliberately. The passthrough is pinned policy (the NOTE in api_error_i18n.ts names this exact string, and main_api_error.test.ts pins the passthrough), and the sustained-outage end state is an incident surface where the English matching the server source is what makes support reports greppable. The capacity refusals differ: they are routine operational states with player-actionable meaning. If we want to flip the policy for the outage end state, that should be its own small change that rewrites the documented exemption and its pins on purpose, not another rider here; glad to file it as a follow-up.

Verification: full local gate green end to end (i18n gen and freshness, malware scan, changed-files biome, SFX conformance, the full suite at 14,153 tests, tsc, all five builds) except the one known environmental armory browser-layout failure on this dev box, which CI arbitrates. Every new test was mutation-checked where it pins a fix: each fails against the pre-fix code.

PR #1935, opened 2026-07-14
fix(delve): correct Drowned Litany collision and progression

MasterZensei:
## Summary

- keep pulled bell ropes and other stateful Drowned Litany puzzle props visible after their mesh state changes
- make rendered polygon walls use the exact same spans as collision, and remove the legacy rectangular clamp from irregular rooms
- keep Baptistry hatchlings on valid walkable floor and preserve both Cantor shield waves when thresholds are crossed together
- expand coverage across every Litany room, collision edge, puzzle state, and progression path

## Root cause

The renderer hid stateful delve props when their generic `lootable` flag became false, so pulled bell ropes disappeared even though the simulation kept them active. Polygon wall visuals were also placed at a fixed pitch while collision used shorter segmented spans, and irregular rooms were still subject to an old rectangular bounds clamp. Together these caused visible/collision mismatch and invisible walls.

## Impact

The Choir Loft can now complete normally after both bell ropes are pulled. Drowned Litany room walls follow their authored polygon edges more closely, collision matches the visible shell, and the other room-specific progression paths are covered against regressions.

### Before: `release/v0.26.0`

Drowned Litany Choir Loft before wall-span alignment

### After: this PR

Drowned Litany Choir Loft after wall-span alignment

Desktop capture only: the changed wall geometry is rendered in world space and has no mobile-specific layout path.

## Validation

- live offline browser smoke test through Drowned Litany entry and Choir Loft combat
- `npx vitest run tests/delves.test.ts tests/delve_render.test.ts tests/litany_spawn_collision.test.ts tests/architecture.test.ts` (199 passed)
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`


EnriqueGF:
:) Thanks for the fix, this is a well-scoped and genuinely careful change. The base branch is correct (release/v0.26.0), tsc is clean, and the sim/render/parity suites pass locally. I verified the four fixes against the real code and they hold up: the polygonShellColliders extraction is a behavior-preserving move (byte-identical collider output/order), the cantorShieldAdds.push correctly fixes the both-thresholds-in-one-update case, the wall-scaling math (scale.x = halfLength/2 on a 4u module) exactly spans each collider segment, and the hatchling spawn stays deterministic across hosts. Test coverage is strong: each fix has a decisive assertion that fails on revert. A couple of things to address before this is mergeable, ordered by severity.

1. (Blocking per repo policy) Missing before/after screenshots for a visual change. placePolygonWalls now renders Drowned Litany room walls as one module stretched/compressed to each 6u-max collision segment instead of fixed 8u modules with overhang, and banner cadence shifts from 8u to 6u pitch (about 33 percent more frequent, new positions). That is a visible change, so the repo requirement "if the change is visual, add before/after screenshots (desktop and mobile where relevant)" applies. Please capture a Litany room before/after and reference it from the PR body (the pr-screenshots skill has the recipe). src/render/dungeon.ts placePolygonWalls.

2. (Should fix) Module-first: delveInteractableVisible in src/render/delve_props.ts is a pure, Three/DOM-free per-frame visibility decision. Per src/render/CLAUDE.md that logic should live in its own *_core.ts registered in RENDER_PURE_CORES (swept for purity/completeness), with the Object3D-mutating syncDelveInteractableVisibility left as the thin consumer. It is one line with a direct test so the risk is low, but it currently escapes the purity sweep.

3. (Minor) The new visibility policy relies on an implicit sim contract: a delve prop that should disappear must be dropEntity-d, never just flagged lootable=false. That holds today (verified: no delve_ prop reaches the lootable=false hide path, and delve_reward_chest intentionally stays visible as an opened chest), so this change actually restores visibility that the old e.lootable gate was wrongly hiding. It is only future-facing: a new delve_ template that flips lootable=false on consumption would stay visible with no render-side guard. Worth a short comment on the contract, or a test asserting delve props are removed (not just un-lootable) when meant to hide.

4. (Minor, test nits) Two small coverage gaps: syncDelveInteractableVisibility is never exercised with withinPortalRange=false (only the default true), and delveInteractableVisible is never called with a null templateId even though renderer.ts passes e.templateId which can be null. Both are behaviorally safe but unpinned; one line each would close them.

None of these are correctness defects, and the underlying fixes are good. Items 1 and 2 are what stand between this and an approval. Happy to re-review once the screenshots and the pure-core extraction land.

MasterZensei:
> :) Thanks for the fix, this is a well-scoped and genuinely careful change. The base branch is correct (release/v0.26.0), tsc is clean, and the sim/render/parity suites pass locally. I verified the four fixes against the real code and they hold up: the polygonShellColliders extraction is a behavior-preserving move (byte-identical collider output/order), the cantorShieldAdds.push correctly fixes the both-thresholds-in-one-update case, the wall-scaling math (scale.x = halfLength/2 on a 4u module) exactly spans each collider segment, and the hatchling spawn stays deterministic across hosts. Test coverage is strong: each fix has a decisive assertion that fails on revert. A couple of things to address before this is mergeable, ordered by severity.
> 
> 1. (Blocking per repo policy) Missing before/after screenshots for a visual change. placePolygonWalls now renders Drowned Litany room walls as one module stretched/compressed to each 6u-max collision segment instead of fixed 8u modules with overhang, and banner cadence shifts from 8u to 6u pitch (about 33 percent more frequent, new positions). That is a visible change, so the repo requirement "if the change is visual, add before/after screenshots (desktop and mobile where relevant)" applies. Please capture a Litany room before/after and reference it from the PR body (the pr-screenshots skill has the recipe). src/render/dungeon.ts placePolygonWalls.
> 2. (Should fix) Module-first: delveInteractableVisible in src/render/delve_props.ts is a pure, Three/DOM-free per-frame visibility decision. Per src/render/CLAUDE.md that logic should live in its own *_core.ts registered in RENDER_PURE_CORES (swept for purity/completeness), with the Object3D-mutating syncDelveInteractableVisibility left as the thin consumer. It is one line with a direct test so the risk is low, but it currently escapes the purity sweep.
> 3. (Minor) The new visibility policy relies on an implicit sim contract: a delve prop that should disappear must be dropEntity-d, never just flagged lootable=false. That holds today (verified: no delve_ prop reaches the lootable=false hide path, and delve_reward_chest intentionally stays visible as an opened chest), so this change actually restores visibility that the old e.lootable gate was wrongly hiding. It is only future-facing: a new delve_ template that flips lootable=false on consumption would stay visible with no render-side guard. Worth a short comment on the contract, or a test asserting delve props are removed (not just un-lootable) when meant to hide.
> 4. (Minor, test nits) Two small coverage gaps: syncDelveInteractableVisibility is never exercised with withinPortalRange=false (only the default true), and delveInteractableVisible is never called with a null templateId even though renderer.ts passes e.templateId which can be null. Both are behaviorally safe but unpinned; one line each would close them.
> 
> None of these are correctness defects, and the underlying fixes are good. Items 1 and 2 are what stand between this and an approval. Happy to re-review once the screenshots and the pure-core extraction land.

Done, can you re-review?

jgyy (commented):
This is a solid, well-scoped bug fix. Three real bugs in the Drowned Litany delve, each with a clear root cause:

- `renderer.ts` was hiding any prop the instant `lootable` went false, which also hid stateful `delve_*` props (like the choir loft bell rope) that the sim intentionally keeps alive. Fixed with a small pure predicate (`delve_interactable_visibility_core.ts`) consumed by `delve_props.ts`, which is exactly the pure-core pattern this repo wants and it's correctly added to the `RENDER_PURE_CORES` allowlist.
- Render drew polygon walls at a fixed 8-unit pitch while collision used the actual authored segment lengths, so the two disagreed. Now both sides consume the same `polygonWallSegments()` from `delve_litany_layout.ts`, which removes the class of bug entirely instead of patching the symptom.
- The rectangular bounds clamp was still running on rooms with an authored `shellPolygon`, producing invisible walls. Now skipped for polygon-shell modules.

The Cantor shield fix (`cantorShieldAdds = ...push(...)` instead of overwrite) is correct, and I checked that the array only resets to `[]` once all tracked adds are dead, so it can't leak across multiple threshold waves in the same tick.

I ran the four test files independently rather than trusting the PR description: `tests/delves.test.ts`, `tests/delve_render.test.ts`, `tests/litany_spawn_collision.test.ts`, `tests/architecture.test.ts` all pass (199/199). The new tests do real rotated-OBB projection (a prior comment literally said "all delve colliders are rot:0", which the test math no longer assumes) and cover a full per-room progression regression for all 5 rooms.

Two minor nits, neither blocking:

- `dungeon.ts:1420`: banner scale stays at the fixed `MODULE_SCALE` while the wall segment itself now gets a non-uniform per-segment scale. Worth a glance that a banner never clips on an unusually short resized segment, but the included before/after screenshots of this area look fine.
- The hatchling spawn retry (up to 12 attempts) changes the RNG draw sequence versus the old 2-attempt path when a spawn site needs retries. Still fully deterministic for a given seed, just flagging it in case it interacts with an RNG-draw-order pin elsewhere. No regression found.

No blocking or should-fix issues. Nice fix.

Issue #1866, opened 2026-07-13
Epic: Make professions and skilling fully playable end to end


FernandoX7:
## Problem

The professions simulation is substantially built and tested, but it does not yet form a complete, visible, fun player-facing loop. The 2026-07-16 audit found that most of the hard mechanics exist in `src/sim/` while being invisible (no skill readout anywhere), unreachable (tested subsystems with no IWorld/wire/UI path), or stubbed online. This epic is the integration umbrella and definition of done for closing that gap.

The design baseline is the Professions 2.0 blueprint (the `docs/professions-2/` packet, PR #2048), which supersedes the original findings list that previously lived in this issue. The blueprint's pillars:

- Identity emerges from play: you become a Smith by smithing; the game notices, a Guild letter arrives, and a named master attunes you. Archetypes are pair-named identities (Smith, Outfitter, Apothecary, Bombardier in wave one).
- RNG on the way in, determinism on the way out: luck lives in gathering (rare fish, pristine veins, rare components); crafting output is deterministic with a celebrated add-only masterwork proc (no five-way output quality roll).
- Stations are people: field recipes craft anywhere; everything uncommon and up happens at a named master NPC's station, who also teaches recipes, sells supplies, and gives quests. Placement is guarded by an automated safety test.
- Players trade with players: gatherers sell to crafters, crafters sell to everyone, NPCs only sink gold; consumables are the economy engine; crafted power stays below the raid floor.
- Nothing breaks: existing items, deeds, quests, saves, and the T keybind keep working through every phase.

## Scope

Wave one ships six deep crafts (weaponcrafting, armorcrafting, tailoring, leatherworking, cooking, alchemy) plus engineering's tool line, fed by mining, herbalism, fishing (joining the gathering framework), and corpse harvesting. Jewelcrafting, inscription, and enchanting stay shallow on the wheel (enchanting becomes reachable). The phased plan, acceptance criteria, and safety workstream live in the planning packet: the `docs/professions-2/` planning packet (PR #2048).

PR #2039 (quest-driven attunement, the cprof identity wire, attunement-gated combo eligibility) merges as the foundation; the craft-ring reorder and pair-named titles land inside its merge window, while no player holds persisted pair ids.

## Corrections to the original findings (2026-07-16 audit)

- Node readiness online is fixed (the ncd wire landed); that finding is stale.
- The live recipe count is 21 (9 common, 6 tool, 3 caster-hub, 3 combo), not 15.
[...]

PR #1937, opened 2026-07-15
feat(game): wire gather node harvesting into normal play controls

jgyy:
## Summary

Part of #1866 ("Epic: Make professions and skilling fully playable end to end") — specifically the first bullet under "Access and host parity":

> Mining, Logging, and Herbalism nodes are rendered and appear on the minimap, but normal client interaction never calls harvestNode. The interaction scan targets entities rather than static gather nodes, and node meshes are not click targets. Keyboard, mouse, touch, and gamepad players therefore cannot harvest them.

That epic covers dozens of independent gaps and explicitly asks for them to land as separate bounded PRs rather than one monolithic change, so this PR is scoped to only that node-control gap.

`harvestNode`/`nodeHarvestableByMe` already satisfied `IWorld` on both `Sim` and `ClientWorld`, and `server/game.ts` already dispatched the `harvest_node` command — the sim/server side was fully implemented and covered by `tests/gather_node_harvest.test.ts`, just never reachable from real input. This PR is the client interaction wiring, plus fixing the online readiness read that was a permanently-`true` stub.

- `renderer.pickGatherNode(x, y)`: raycasts the gather-node meshes (tagged with `userData.gatherNodeId` in `gather_nodes.ts`) and resolves a node id. Kept separate from `pick()`'s numeric entity-id contract rather than widening it.
- `src/game/gather_node_interact.ts`: a pure, DOM-free core (`decideGatherNodeAction` / `handleGatherNodeInteract`) deciding too-far / not-ready / harvest, mirroring the shape of `handlePickedEntity` in `interactions.ts`.
- `main.ts`: wires both entry points — `handlePick` (mouse click, and touch, which shares the same click pipeline) and `interactKey` (keyboard, gamepad, and the mobile Interact button all already funnel through this one function).
- `ClientWorld.nodeHarvestableByMe` was a stub that always returned `true` ("Online profession reads are incomplete... node readiness is always reported as ready"). It now mirrors the real per-player respawn timer via a new self-wire delta `ncd`, the same remaining-seconds shape as the existing ability-cooldown `cds` field, so online readiness matches what `Sim.nodeHarvestableByMeFor` already does offline.

### Out of scope (left for the epic's other named gaps)
Tool/charge integration, node reward tables using rolled rarity, crafting-station discoverability, and everything else `#1866` lists under crafting progression / item identity / economy.

## Diagram

```mermaid
flowchart TD
    subgraph Input["Player input (any method)"]
        Click["Mouse click / touch tap<br/>renderer.pickDirect(x, y) -> null"]
        Key["Keyboard/gamepad Interact<br/>+ mobile Interact button"]
    end

    Click --> PickNode["renderer.pickGatherNode(x, y)<br/>raycasts gatherNodeMeshes"]
    Click -->|node raycast also empty| PickSloppy["renderer.pickSloppy(x, y)<br/>26px character assist"]
    Key --> ScanNode["interactKey(): nearest GATHER_NODES<br/>entry within INTERACT_RANGE<br/>(spirit_healer excluded from the NPC scan)"]

    PickNode --> Core
    ScanNode --> Core["handleGatherNodeInteract<br/>(pure core, gather_node_interact.ts)"]

    Core -->|too far| ErrTooFar["hud.showError(tooFar)"]
    Core -->|not ready| ErrNotReady["hud.showError(notReady)"]
    Core -->|ready| Harvest["world.harvestNode(nodeId)"]

    Harvest --> Offline["Sim.harvestNode<br/>(offline host)"]
    Harvest --> Online["ClientWorld.harvestNode<br/>cmd: harvest_node"]
    Online --> Server["server/game.ts dispatch<br/>sim.harvestNode(node, pid)"]
    Server --> ServerSim["Sim.harvestNode<br/>(authoritative)"]

    ServerSim --> Grant["grants item, rolls rarity,<br/>sets nodeHarvestReadyAt"]
    Offline --> Grant

    ServerSim --> Ncd["self-wire delta 'ncd':<br/>remaining seconds per node"]
    Ncd --> ClientMirror["ClientWorld.nodeCooldowns<br/>(mirrored per-viewer readiness)"]
    ClientMirror --> Core
```

## Review follow-up (@FernandoX7)

Addressed all three should-fix reachability findings from review:

1. **`interactKey` starved at `herb_thornpeak_1` by the invisible Spirit Healer.** The NPC scan now excludes `spirit_healer` for a living player, the same filter `renderer.pick()` already applied on the click path, so the angel can no longer out-prioritize a node sharing its graveyard's interact range on keyboard/gamepad/mobile.
2. **`pick()`'s 26px sloppy assist stole direct node clicks.** Split `renderer.pick()` into `pickDirect()` (the raycast-only half) and `pickSloppy()` (the forgiving character-radius fallback); `handlePick` now tries direct entity hit, then direct node hit, then the sloppy assist, then the ground-click fallback — matching the order the review suggested, so a click dead-center on a node mesh next to a mob/gatherer harvests the node instead of snapping to the nearby character.
3. **The cooldown-clearing transition was never exercised end to end.** Added `tests/snapshots.test.ts`'s "gather node cooldown wire round trip (ncd)" test: a real `GameServer` broadcast with a cooling node (`self.ncd` ships the remaining seconds), then the readyAt passing and a second broadcast asserting `self.ncd` ships `{}` and `ClientWorld.nodeHarvestableByMe` flips back to `true` through a real `applySnapshot`, not a hand-reassigned map.

The nits (button-gating on node clicks, the `ncd`-apply allocation, the strict-vs-`+1` INTERACT_RANGE boundary, the unknown-node-id `IWorld` asymmetry, the missing click marker on a successful node harvest, and the too-far-node-click-doesn't-walk behavior) are left as-is per the review's own "none blocking" framing; happy to take a follow-up pass on any of them if preferred over a separate PR.

## Screenshot

Offline client, player standing next to an ore vein node — same node visuals as before this change (rendering is unchanged); what's new is that this is now a click/tap/interact-key target instead of inert scenery.

gather node

After the review fixes above (pick-order + Spirit Healer exclusion), same offline client, confirming the world still renders and drives identically — these are reachability-logic fixes with no visual surface of their own:

gather node after review fixes

## Test plan

- [x] `src/game/gather_node_interact.ts` pure-core unit tests (`tests/gather_node_interact.test.ts`): too-far / not-ready / harvest verdicts and the hud/world dispatch.
- [x] `tests/gather_node_online.test.ts`: `ClientWorld.nodeHarvestableByMe` before any snapshot, with an active cooldown, and after it clears.
- [x] `tests/snapshots.test.ts`: extended the full self-state wire fixture with a cooling-down node, asserting the new `ncd` delta key round-trips and `ALL_DELTA_KEYS`/scraper pins updated (44 -> 45); new "gather node cooldown wire round trip (ncd)" describe block proves the not-ready -> ready transition through a real server broadcast + `applySnapshot`.
- [x] `tests/gather_node_harvest.test.ts`, `tests/gathering_view.test.ts`, `tests/prof_intro_quest.test.ts`, `tests/world_api_parity.test.ts`, `tests/command_facets.test.ts`, `tests/command_schema.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/render_glb_replacement_assets.test.ts`, `tests/pr_shot_targets.test.ts` all pass.
- [x] `npm run build` succeeds (all five entries).
- [x] `npm run gate`: green except pre-existing, unrelated failures confirmed present on an unmodified `release/v0.26.0` checkout too (SFX production-conformance dBFS drift, `sfx_studio_server_security` 400-vs-200, `deeds_window_focus` jsdom focus flake, `i18n_completeness` non-Latin gaps, `server/http/characterization` leaderboard-response timeouts) — none touch gathering, rendering, or the wire protocol this PR changes.


FernandoX7 (commented):
Good PR. The scope discipline is right (client wiring only, the sim/server harvest logic stays untouched), `gather_node_interact.ts` follows the `interactions.ts` pure-core pattern and is tested directly, and the `ncd` delta is a faithful copy of the `cds` precedent. I verified the parts that could have caused a permanent online lockout, and they hold: `maybe()` diffs the serialized form per session key so the transition to `{}` ships and the client flips back to ready; `session.lastSent` resets on every reconnect path so a reused `ClientWorld` cannot keep a stale baseline; the wire filter (`until > sim.time`) and the sim readiness check (`now >= readyAt`) agree at the boundary; and `nodeHarvestReadyAt` is session-only, so nothing here touches persistence. Retiring the always-true stub also makes the minimap and gathering-view readiness truthful online. I traced all five input modalities to the core (keyboard main.ts:1208, mobile Interact main.ts:1306, gamepad main.ts:1380, mouse/touch via `handlePick`). Performance is fine: everything the PR adds is bounded by the 24 nodes in content and runs per event, never per frame, and there is no DB surface at all.

Three should-fix items, all reachability, which is the thing this PR exists to deliver:

1. **should-fix** `src/main.ts:2099` (interactKey): the Interact key is starved at `herb_thornpeak_1` by the invisible Spirit Healer. The NPC branch scans `e.kind === 'npc'` with no spirit-healer filter at threshold `INTERACT_RANGE + 1` (6 yd) and returns before the new node branch. A Spirit Healer is permanently spawned at every overworld graveyard, and `gy_thornpeak` (15, 645) sits 4.24 yd from `herb_thornpeak_1` (18, 648), so from the natural harvest position keyboard, gamepad, and the mobile Interact button all open a gossip dialog for an NPC the renderer hides from living players instead of harvesting. Mouse/tap still works because `pick()` skips the healer for the living; the interact scan should skip it the same way (or otherwise let the node compete on distance).

2. **should-fix** `src/main.ts:2186` + `src/render/renderer.ts:5712`: `pick()`'s 26 px sloppy assist steals direct node clicks. The node raycast only runs when `pick()` returns null, but `pick()` falls through to the forgiving assist and returns the nearest targetable character within 26 screen px, so a click or tap dead-center on a node mesh becomes an entity pick whenever any mob or player is on or near the node. Several nodes sit inside live camp radii (`wood_eastbrook_1` is 3.6 yd from the webwood_spider camp center, `ore_thornpeak_2` 5.8 yd from deeprock_kobold, `wood_thornpeak_2` 5.7 yd from glimmermere_wader), and other gatherers standing on a node create the same steal. That contradicts the inline comment ("a click that lands on a node harvests it"), and on mobile tap is the primary modality. Suggested order: direct entity hit, then direct node hit, then the sloppy assist, then ground click.

3. **should-fix** `tests/gather_node_online.test.ts:51`: the cooldown-clearing transition is never exercised end to end. The unit test hand-reassigns `nodeCooldowns = new Map()` (bypassing both the server filter and the `s.ncd` decode), and the snapshots fixture broadcasts once with one cooling node, so no test proves the server ships `{}` after a non-empty `ncd` and the client flips back to ready. A regression there is exactly the "node permanently greyed online" failure this feature has a history of. One more cycle in the snapshots fixture closes it: after the not-ready assertion, clear `meta.nodeHarvestReadyAt` (or advance `sim.time` past `readyAt`), broadcast again, assert `self.ncd` ships `{}` and `nodeHarvestableByMe` returns true through `applySnapshot`.

Nits, none blocking:

- `src/main.ts:2040`: the node scan gate is strict `d < INTERACT_RANGE` while the sibling NPC/delve branches use `INTERACT_RANGE + 1` and the click path allows `INTERACT_RANGE + 1`. At the boundary a click harvests but Interact reports nothing to interact. Aligning with the sibling tolerance also removes the inconsistency.
- `src/net/online.ts:2021`: the `ncd` apply allocates `Object.entries().map()` plus a fresh Map on every snapshot while a node cools, four lines below the `cds` handler whose comment documents the in-place rebuild precisely to avoid that on the 20 Hz self record. A lazy-init guard (`if (!this.nodeCooldowns) this.nodeCooldowns = new Map(); else this.nodeCooldowns.clear();` then set in a loop) keeps the fixture safety the reassignment was for and stays on the documented pattern.
- `src/net/online.ts:1226`: `nodeHarvestableByMe` returns true for an unknown node id; `Sim.nodeHarvestableByMeFor` returns false. Unreachable today (both call sites resolve the id from `GATHER_NODES` first, and the server re-validates), but it is a latent IWorld asymmetry worth a one-line guard or comment.
- `src/main.ts:2186`: the node block ignores `button`, so a right or middle click that lands on a node harvests instead of doing that button's normal action; the entity path threads `button` through to `handlePickedEntity`.
- `src/main.ts:2100`: a successful harvest click skips the OSRS click marker even though the adjacent comment defines the marker as stamping clicks that do something. An entity click one branch later gets one.
- The too-far node click now errors in place rather than walking toward the node like the pre-PR ground-click fallback did. The inline comment says that is deliberate, but the classic expectation is walk-then-harvest, so flagging in case the intent was only to stop the deselect, not the approach.

Merge state is clean against release/v0.26.0, the delta-key pins were updated correctly (44 to 45 plus the scraper), and the new English key lives in the right catalog and renders through `t()` at the call site.

jgyy:
Thanks for the thorough review, @FernandoX7! Pushed a follow-up commit addressing all three should-fix items:

1. **Spirit Healer starving `interactKey` at `herb_thornpeak_1`**: `src/main.ts`'s NPC scan now excludes `spirit_healer` for a living player (`e.templateId !== 'spirit_healer' || p.ghost`), the same filter `renderer.pick()` already applies on the click path. Keyboard/gamepad/mobile Interact can no longer be starved by the graveyard angel.
2. **`pick()`'s 26px sloppy assist stealing direct node clicks**: split `renderer.pick()` into `pickDirect()` and `pickSloppy()`, and reordered `handlePick` to your suggested order — direct entity hit, then direct node hit, then the sloppy assist, then the ground-click fallback. A click dead-center on a node mesh now harvests it even when a mob or another gatherer is camped on top.
3. **Untested cooldown-clearing transition**: added a new test in `tests/snapshots.test.ts` ("gather node cooldown wire round trip (ncd)") that drives a real `GameServer` broadcast, asserts `self.ncd` ships the remaining seconds while cooling, then clears the server-side timer, broadcasts again, and asserts `self.ncd` ships `{}` and `ClientWorld.nodeHarvestableByMe` flips back to `true` through a real `applySnapshot` — not the prior hand-reassigned map.

Left the nits as-is per your "none blocking" framing (button-gating on node clicks, the `ncd` allocation pattern, the strict-vs-`+1` INTERACT_RANGE boundary, the unknown-node-id `IWorld` asymmetry, the missing click marker, and the too-far-node-click-doesn't-walk behavior) — happy to take a pass on any of them here or in a follow-up if you'd rather see them folded in.

All of `tests/snapshots.test.ts`, `tests/gather_node_interact.test.ts`, `tests/gather_node_online.test.ts`, `tests/gather_node_harvest.test.ts`, `tests/architecture.test.ts`, and `tests/localization_fixes.test.ts` pass, `tsc --noEmit` is clean, and `npm run build` succeeds. Updated the PR description with a "Review follow-up" section and an after-fix screenshot.

FernandoX7 (approved):

PR #1942, opened 2026-07-15
refactor(ui): extract char skin picker + skin event overlay from hud.ts

jgyy:
## Summary

`src/ui/hud.ts` is one of the four logic coordinators the repo forbids growing
further (root `CLAUDE.md`). This PR went through several earlier revisions,
including a version that extracted the cosmetic skin-roll overlay alongside
the character-sheet skin picker; upstream's own `release/v0.27.0`
domain-module-tree refactor independently extracted that same overlay into
`SkinEventController` (`src/ui/hud/cosmetics/skin_event_controller.ts`) plus
a sibling `PlayerCardController`, superseding roughly half of this branch's
original extraction. This revision keeps upstream's controllers as-is (they
already satisfy the "extract skin-event overlay out of `hud.ts`" goal, and as
a full controller class) and trims this PR down to the piece upstream had
**not** extracted: the character-sheet skin (chroma) picker row, following
the pure-core / thin-painter recipe in `src/ui/CLAUDE.md`.

- New `src/ui/char_skin_window.ts` (147 lines): moves `paintCharSkinPicker`
  out of `hud.ts` (renders the class/mech skin swatch row on the character
  sheet, including the mech-chroma unequip control), plus its supporting
  `classCss` helper. It reuses upstream's `mechChromaName` from
  `./hud/cosmetics` instead of duplicating it.
  - `hud.ts` exposes the fields/callbacks the module needs via one
    `CharSkinPainterHost` interface, built explicitly by a private
    `skinHost()` accessor (a deps object of bound closures, the same idiom
    `SkinEventController` and `BagsWindow` already use for this), not a
    structural cast, so tsc keeps checking every member of the seam.
- `tests/char_skin_window.test.ts` (6 cases) covers the extracted DOM-driven
  picker directly: the no-row no-op, class-swatch rendering and selection,
  clicking a class swatch (commits the skin, re-mounts the preview), the mech
  catalog and unequip control appearing once a chroma is unlocked (and being
  omitted when it is not), and clicking a mech swatch (commits the skin,
  awaits the mocked asset-preload promise, then re-mounts the preview). The
  removed skin-event overlay helpers already have coverage under
  `tests/skin_event_model.test.ts` and `tests/skin_event_controller.test.ts`.
- The file is named `_window.ts`, not `_painter.ts`: it is a cold/modal
  component (opened rarely, not per-frame), so it is intentionally NOT
  registered in the `HOT_PAINTERS`/`PainterHost` per-frame write-elision
  contract (`tests/hud_perf_budget.test.ts`'s completeness sweep only
  requires that classification for files matching `*_painter.ts`).

This is a pure logic relocation for the picker row: the DOM string/attribute
output is unchanged for the same inputs. See the Screenshots section below
for the visual-regression proof of the picker itself; the skin-event overlay
screenshots that an earlier revision of this PR included have been dropped
since that code is no longer part of this diff (it lives entirely in
upstream's `SkinEventController` now).

## Follow-up: shared tab-strip building block

Second commit on this branch, a different kind of extraction: not a window
moved out of `hud.ts`, but a **new reusable building block** other windows
compose from instead of hand-rolling their own chrome. Five windows
(`social_window.ts`, `market_window.ts`, `talents_window.ts`,
`daily_rewards_window.ts`, `mailbox_window.ts`) each independently hand-roll
the same WAI-ARIA tab-strip markup (`role=tablist`/`role=tab`,
`aria-selected`, a roving `tabindex`). The keyboard-navigation math already
lives in one shared core (`roving_index.ts`); the markup and the click/keydown
wiring did not.

- New `src/ui/tab_strip_view.ts` (pure, `UI_PURE_CORES`-registered): a
  `tabStripModel()` + `tabStripHtml()` pair that turns a small descriptor
  (aria-label, tab/strip classes, panel id, `{id,label}` tabs, selected id)
  into the tablist markup.
- New `src/ui/tab_strip_painter.ts` (thin, DOM-only): `wireTabStrip()` wires
  click plus roving Arrow/Home/End/Enter/Space over the existing
  `roving_index.ts` core, and reports whether a selection came from the
  keyboard (`focusFollow`) so the caller only refocuses the newly active tab
  on a keyboard move, matching the WAI-ARIA tabs pattern.
- `src/ui/social_window.ts` is the first migration: its five tabs (friends /
  guild / ignore / block / raid) now build through `tabStripModel` +
  `tabStripHtml` and wire through `wireTabStrip`, replacing ~40 lines of
  hand-rolled markup and keyboard handling. Byte-identical markup and
  behavior for the same inputs (see `tests/social_window.test.ts`, updated to
  pin composition of the shared modules instead of the markup it used to
  build inline; the pin now also asserts an exact `role="tab"` count, not
  just presence of each expected id).
- New `tests/tab_strip_view.test.ts` (model selection, markup shape, escaping)
  and `tests/tab_strip_painter.test.ts` (click / roving keys / Enter-Space /
  a non-navigation key / an unmatched tab class wiring nothing, via a
  hand-rolled fake DOM per the `tests/CLAUDE.md` convention; the fake's
  `querySelectorAll` filters by the class in the selector, matching its
  `querySelector`, so a wrong `tabClass` at a call site fails the test
  instead of silently wiring every tab).

**Generalized for the other real tab strips, not just social's shape** (a
third commit on this branch): `talents_window.ts` uses `<div role="tab">`
elements with a spent-points badge after the label, and
`daily_rewards_window.ts`'s woc-store tabs give each tab its own element id
and let the panel reference it via `aria-labelledby` instead of the strip
driving `aria-controls`. Rather than leave the shared core fitted only to
social's plain-button shape, `tabStripModel`/`tabStripHtml` now take an
optional per-tab `extraHtml` (a badge), an optional tab `tag` (`'button'`
default, `'div'`), a per-tab `buttonId`, and an optional `panelId`
(`aria-controls` is omitted when absent); `wireTabStrip` takes an optional
roving `orientation` (`'horizontal'` default, `'both'` for a future
vertical/grid strip, the same knob `roving_index.ts` already exposes); and a
new `focusActiveTab(container, tabClass, selectedClass)` factors out the
"refocus the newly selected tab after a keyboard move" step every future
caller needs, which `social_window.ts` now calls instead of hand-rolling the
query. `social_window.ts`'s own output and behavior are unchanged (still a
plain button strip). `market_window.ts` and `mailbox_window.ts` stay out of
scope for this core: they use a different toggle-button (`aria-pressed`)
pattern, not real WAI-ARIA tabs.

**Proven with a second real consumer** (a fourth commit on this branch):
`talents_window.ts`'s class/spec tab strip now also builds through
`tabStripModel`/`tabStripHtml` (using the `tag: 'div'` + `extraHtml` knobs
above for its spent-points badge) and wires through `wireTabStrip` +
`focusActiveTab`, replacing its own hand-rolled markup and roving-keyboard
handler. Same `tal-tab`/`active`/`tal-body` classes and id as before; two
differences from the old hand-rolled markup are worth calling out rather than
claiming byte-identical output: the migrated strip drops the old
`<span class="tal-tab-label">` wrapper around the label text (verified inert:
that class has zero references anywhere else at head, and `.tal-tab`'s flex
gap still separates the label from the `.tt-pts` badge without it) and
changes attribute order, and Enter/Space now also refocuses the active tab
after re-render, matching social's behavior (an improvement over the old
`keyboardActivate` path, which did not). `tests/talents_window.test.ts` gets
the same composition pins `tests/social_window.test.ts` already has. The
building block is now used by two windows, not designed-but-unused.

`daily_rewards_window.ts`'s woc-store tabs are left as a follow-up: unlike
social/talents (which fully re-render the window on a tab switch),
daily_rewards mutates the existing tab buttons' `aria-selected`/`tabindex`
attributes and a shared panel's `aria-labelledby` in place without a full
re-render, so wiring it onto `wireTabStrip`'s re-render-driven `onSelect`
needs a small restructure of its own, better as its own scoped PR than
folded into this one.

```mermaid
flowchart TD
    subgraph shared["Shared building block"]
        view["tab_strip_view.ts<br/>(pure model + markup)"]
        painter["tab_strip_painter.ts<br/>(click / roving keyboard wiring)"]
    end
    roving["roving_index.ts<br/>(shared roving-tabindex math)"]
    painter --> roving

    social["social_window.ts<br/>(migrated, plain button)"]
    talents["talents_window.ts<br/>(migrated, div tag + badge)"]
    daily["daily_rewards_window.ts<br/>(follow-up: needs a small\nre-render restructure first)"]
    market["market_window.ts<br/>(out of scope: aria-pressed toggle, not tabs)"]
    mailbox["mailbox_window.ts<br/>(out of scope: aria-pressed toggle, not tabs)"]

    social --> view
    social --> painter
    talents --> view
    talents --> painter
    daily -.-> view
    daily -.-> painter
```

## Related issues

N/A

## Type of change

- [x] Tests

## How was this tested?

- Commands:
  - `npx vitest run tests/char_skin_window.test.ts tests/skin_event_model.test.ts tests/skin_event_controller.test.ts tests/char_window.test.ts tests/architecture.test.ts tests/hud_perf_budget.test.ts tests/painter_host.test.ts`
  - `npx vitest run tests/tab_strip_view.test.ts tests/tab_strip_painter.test.ts tests/social_window.test.ts tests/talents_window.test.ts tests/talents.test.ts tests/architecture.test.ts tests/social.test.ts`
  - `npx tsc --noEmit`
  - `npx @biomejs/biome check --write` on all changed files (0 errors)
  - `npx vite build`
  - `npm run gate`

## Screenshots / recordings

Pure logic relocation for the picker row: `src/ui/char_skin_window.ts`
produces byte-identical DOM to the pre-extraction `hud.ts` for the same
inputs, so nothing a player sees should change. Captured before/after pairs
for the character-window skin picker (multiple skin/mech-chroma options
selected), on desktop and mobile, to confirm the extraction did not change
anything visually.

### Character skin picker (character window)

Desktop:

| Before | After |
|---|---|
| before desktop skin picker | after desktop skin picker |
| before desktop skin picker option 2 | after desktop skin picker option 2 |
| before desktop skin picker option 3 | after desktop skin picker option 3 |

Mobile (landscape, the game's only mobile orientation):

| Before | After |
|---|---|
| before mobile skin picker | after mobile skin picker |
| before mobile skin picker option 2 | after mobile skin picker option 2 |
| before mobile skin picker option 3 | after mobile skin picker option 3 |

The skin-event overlay screenshots an earlier revision of this PR included
have been dropped: that overlay now lives entirely in upstream's
`SkinEventController` and is out of scope for this diff. The social panel's
and talents window's tab strips are unchanged as well (their extraction is
purely internal markup composition with pinned tests, not a new visual
element), so no separate screenshots are included for that part of the diff.


jgyy:

jgyy:
Merged the latest \`release/v0.27.0\` in to resolve the conflict. The only real overlap was in \`src/ui/hud.ts\`: since the base branch, upstream independently landed its own domain-module-tree refactor and, as part of it, extracted the cosmetic skin-roll overlay into \`SkinEventController\` (\`src/ui/hud/cosmetics/skin_event_controller.ts\`) and the player card into \`PlayerCardController\` (\`src/ui/hud/player_card/\`) - which superseded roughly half of this PR's \`char_skin_window.ts\` extraction (the skin-event overlay paint functions).

Resolution: kept upstream's controllers as-is (they already satisfy the "extract skin-event overlay out of hud.ts" goal, and better: as a full controller class alongside its new sibling \`PlayerCardController\`), and trimmed \`char_skin_window.ts\` down to the piece upstream had **not** extracted - the character-sheet skin picker row (\`paintCharSkinPicker\`). It now reuses upstream's \`mechChromaName\` from the new \`hud/cosmetics\` location instead of duplicating it, and \`tests/char_skin_window.test.ts\` was rewritten to exercise the picker's DOM output directly (the removed skin-event helpers already have coverage under \`tests/skin_event_model.test.ts\` and \`tests/skin_event_controller.test.ts\`).

Also dropped a second stale \`hud.ts\` import block (old-path \`chat_announcer\`/\`chat_channels\`) that this branch never touched but which conflicted textually with an adjacent edit; it's superseded by upstream's \`ChatWindowController\` extraction.

A local \`npm install\` during conflict resolution accidentally dropped an optional lockfile entry (svelte-check's \`picomatch\`), which broke \`npm ci\` in CI; fixed in a follow-up commit restoring it from upstream's lockfile.

CI is green (Biome, typecheck/builds, browser regressions, all 4 English-only legal shards, and screenshots), and the PR is mergeable/clean against \`release/v0.27.0\`.

FernandoX7 (commented):
Good extraction. The social strip is byte-identical to the old hand-rolled markup (verified attribute by attribute, trailing space included), the shared tab-strip building block lands with two real consumers instead of designed-but-unused, and the release-merge resolution was handled well: trimming this branch to the piece upstream's SkinEventController did not already extract and reusing upstream's `mechChromaName` instead of duplicating it. The new `tab_strip_view` pins are also stronger than the old bare `role="tab"` count (byte-exact markup assertions). CI is green and the targeted suites (architecture, hud_perf_budget, painter_host, plus all new and updated tests) pass locally at the PR head.

Findings:

- **should-fix** `src/ui/hud.ts:10273`: `skinHost()` returns `this as unknown as CharSkinPainterHost`. The interface names private `Hud` members (`mechAssetsPromise`, `mountCharPreview`, `renderCharIfOpen`, `sim`), which is why the cast has to go through `unknown`, and that erases all compile-time checking of the seam: a later rename or signature change on any of those members compiles clean and breaks the character sheet at runtime. This file already uses the checked idiom for exactly this situation, the explicit deps object of bound closures (`SkinEventController` at `hud.ts:1450` even wraps this same `mechAssetsPromise` lazy-init in a `preloadMechAssets` closure; `BagsWindow` at `hud.ts:3277`). Passing a deps object restores tsc coverage; the module header's "Hud satisfies this structurally ... no adapter object required" comment should be corrected with it, since the private members are exactly why it does not.
- **should-fix** (PR body): the description is stale after the `229b579a9` release merge. It still describes a 465-line module containing `paintSkinEventWheel`/`paintSkinEvent`, `hud.ts` at -414 lines, and 10 test cases; at head the module is 148 lines (the picker only), the overlay lives in upstream's `SkinEventController`, and the test has 5 cases. The skin-event overlay screenshot pairs now depict code this PR no longer touches. Please update the summary so it matches the diff being merged.
- **should-fix** `tests/char_skin_window.test.ts`: the mech-swatch click path is untested. `changeSkin(skin, 'mech')`, the `preloadMechAssets` promise chain, and the `display:block` + `.sel` guard (`src/ui/char_skin_window.ts:94-117`) are the only async logic in the module, and the mech test only asserts swatch presence and the unequip click. One test that clicks a mech swatch and awaits the mocked promise before asserting `mountCharPreview` would close it.
- **should-fix** `tests/tab_strip_painter.test.ts:30`: `FakeContainer.querySelectorAll` ignores its selector, so `wireTabStrip` called with a wrong `tabClass` still wires every fake tab and passes; the window tests pin the model's `tabClass` field but not the second argument at the call site (`src/ui/social_window.ts:567`). Making the fake filter by the class in the selector (its `querySelector` already does) closes the hole.
- **nit** `src/ui/talents_window.ts`: the fidelity claims overstate slightly. The migrated strip drops the old `<span class="tal-tab-label">` wrapper (base `talents_window.ts:206`) and changes attribute order; I verified this is inert (`tal-tab-label` has zero references anywhere at head, and `.tal-tab`'s flex gap still separates the label text from the `.tt-pts` badge). Enter/Space also now refocuses the active tab after re-render where the old `keyboardActivate` path did not (an improvement, matching social). Worth saying both in the PR body instead of "the CSS and DOM shape are unchanged" and "matching the prior hand-rolled handler".
- **nit** `scripts/char_skin_shot.mjs:97`: `grantMechChromas` pushes `0, 1, 2` into `mechChromaIds`, but chroma ids are strings (`'amber_crimson'`, ...) and `characterAppearanceOptions` matches with a `Set` of string ids, so the grant is a no-op: the captured pickers only ever showed class swatches, never mech swatches or the unequip button. Use real `MECH_CHROMAS` ids if you want those in frame.
- **nit** `src/ui/char_skin_window.ts:25`: `CharSkinPainterHost.sim.player.level` is unused by `paintCharSkinPicker` (a leftover from the pre-merge overlay half); drop it from the interface.
- **nit** `tests/social_window.test.ts`: the old test pinned exactly five `role="tab"`; the new composition pins check each id is present but nothing fails if a sixth tab appears. Narrow, since the markup itself is byte-pinned in `tab_strip_view.test.ts`, but an exact count is cheap to keep.

jgyy:
Thanks for the thorough review! Pushed a follow-up addressing all four should-fix findings plus both nits:

- `skinHost()`: replaced the `this as unknown as CharSkinPainterHost` structural cast with an explicit deps object (the same idiom `SkinEventController`/`BagsWindow` already use), so tsc checks every member of the seam again. `mechAssetsPromise` is no longer a raw field on the interface; it's a `preloadMechAssets(): Promise<void>` closure that keeps the lazy-init private to `Hud`, matching the existing `SkinEventController` wiring at `hud.ts:1450`.
- Added a mech-swatch click test to `tests/char_skin_window.test.ts`: commits the skin, awaits the mocked asset-preload promise, then asserts the preview remounts.
- `FakeContainer.querySelectorAll` in `tests/tab_strip_painter.test.ts` now filters by the class in the selector (matching `querySelector`), every `FakeTab` is tagged with its real tab class, and a new regression test proves a mismatched `tabClass` now wires nothing instead of silently wiring every tab.
- `scripts/char_skin_shot.mjs`'s `grantMechChromas()` now grants real string chroma ids instead of integers (the grant was a no-op before, so mech swatches never actually appeared in the captures).
- Also dropped the now-unused `player.level` field from `CharSkinPainterHost.sim`, and corrected the PR body: the Summary was stale after the `229b579a9` release merge (still described the pre-trim 465-line module with the skin-event overlay), and the talents fidelity claim now calls out the dropped `tal-tab-label` wrapper and the Enter/Space refocus-after-rebuild improvement instead of claiming byte-identical output.

Verified: `tsc --noEmit` clean, all previously-passing suites plus the two new tests green (255 tests), `npm run i18n:gen` clean, parity goldens unaffected, pre-push QA floor green. Also merged the current `release/v0.27.0` in (Card Duel and the rest landed since the last sync) with zero conflicts.

FernandoX7 (approved):

PR #1932, opened 2026-07-14
test(i18n): freshness red-path probe (do not merge)

FernandoX7:
Deliberately staled slice: a catalog value edited without npm run i18n:gen. The pr-gate freshness step must go red; this proves acceptance criterion 4 of toolchain-modernization Phase 1 (PR #1931). Close unmerged after the run.

PR #1946, opened 2026-07-15
test(i18n): pr-checks freshness red-path probe (never merge)

FernandoX7:
Deliberately staled slice probe for Phase 3 (toolchain modernization): adds a catalog key without regenerating the committed artifacts, so the relocated freshness step in the new pr-checks job must turn red on this PR. Pushed with --no-verify because the pre-push floor correctly blocks staled slices locally (tsc types the resolved slices against the catalog shape). Observed and then closed unmerged; mirrors the Phase 1 probe PR #1932. DO NOT MERGE.


FernandoX7:
Probe confirmed: the relocated freshness step in pr-checks turned red exactly as designed (run 29388336786: i18n:gen and the coverage summary green, 'Verify committed i18n artifacts are fresh' failed with the per-file hunks naming the staled slices, typecheck and builds never ran). Closing unmerged, mirroring the Phase 1 probe PR #1932.

PR #1949, opened 2026-07-15
fix(talents): trim shadow priest and frost mage mastery outliers to caster parity

ryan-foo:
## Summary

Follow-up to the balance druid fix (#1919). Auditing all 27 specs for the same always-on multiplier problem surfaced two more caster masteries sitting well above their siblings.

**Shadow priest (Vespers)** carried two damage fields, `dotDmgPct 0.15` AND `spellDmgPct 0.10`. On a DoT those compound with Gloamveil Form's +15% shadow-school amp:

```
form_shadow 1.15  x  (1 + spellDmgPct 0.10)  x  (1 + dotDmgPct 0.15)  =  1.455 per tick
```

That is the pre-fix balance druid outlier (~1.45), before any talent point, on a DoT-centric spec. Every other caster mastery is a single damage field.

**Frost mage (Cryomancy)** gave `+25%` on frostbolt where every sibling caster mastery is `+15%` on its primary (arcane, elemental). It was the only permanent single-nuke outlier.

## Change

- Shadow: drop the `spellDmgPct 0.10` half of Vespers, leaving `dotDmgPct 0.15`. The form stays the general shadow multiplier and the mastery is the DoT specialization. DoT falls to ~1.32 (in band with feral bleed), direct shadow to ~1.15 (form only).
- Frost: frostbolt/frost_nova mastery `dmgPct 0.25` to `0.15`, matching arcane/elemental. The ability-scoping stays (the mage's off-school fire/arcane spells are untouched), only the magnitude drops. Aligns with the spec's own "trades peak burst for survival and slows" flavor.

## Considered and left alone

**Elemental shaman** was flagged in the audit but excluded: fully specced its primary nuke is ~1.53 vs arcane's ~1.59, so it is in band with its true sibling. The earlier flag compared a fully-specced node stack to a mastery-only yardstick. Nerfing it would put it below arcane.

## i18n

Non-English mastery descriptions generate from the effect automatically (`tTalent` -> `effectDescription`), so they track the new numbers with no overlay edits. The English descriptions are updated to match (pinned by `tests/talent_tooltip_accuracy`).

## Testing

- `tests/mastery_mechanism.test.ts`: pins both trims (shadow Vespers is dot-only with no spellDmgPct; frost frostbolt scales 1.15 not 1.25).
- Parity goldens unaffected (no golden trace casts these; suite green with no regen).
- `talent_tooltip_accuracy`, `talents`, `localization_fixes` (S3), `localization_coverage`, `guide`, `architecture` all green.

### DIFF
--- a/src/sim/content/talents_classic.ts
+++ b/src/sim/content/talents_classic.ts
@@ -1255,14 +1255,17 @@ const MAGE_SPECS: SpecDef[] = [
     'A controlling caster who trades peak burst for survival and slows.',
     'icy_veins',
     'Brittlebreak',
-    'Increases your Frost spell damage by 25%. Increases armor by 10%.',
+    'Increases your Frost spell damage by 15%. Increases armor by 10%.',
     // The scalable mastery axis is the Frost-kit damage (ability-scoped so the
     // mage's fire/arcane baseline spells stay untouched); armor is the static
     // secondary. Crit-vs-rooted identity returns as a Shatter-style row option.
+    // The 15% magnitude matches the other caster masteries (arcane/elemental
+    // spellDmgPct 0.15) so frostbolt is not a permanent outlier over its siblings;
+    // the ability-scoping stays so the mage's off-school spells are untouched.
     {
       ability: [
-        { ability: 'frostbolt', dmgPct: 0.25 },
-        { ability: 'frost_nova', dmgPct: 0.25 },
+        { ability: 'frostbolt', dmgPct: 0.15 },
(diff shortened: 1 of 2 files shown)


ryan-foo:
Folded into #1919, which is now the combined caster mastery balance pass (moonkin + shadow priest + frost mage). The shadow/frost commit was cherry-picked onto #1919's branch, so this PR's changes ship there. Closing to keep the caster balance work in one reviewable PR.

PR #1950, opened 2026-07-15
feat(enchanting): per-slot build variety and a shard-consuming Greater tier

ryan-foo:
## Summary

Expands the Enchanting content table (`src/sim/content/enchants.ts`) from one enchant per slot into a proper two-layer table, and closes the `arcane_shard` dead-end.

### Base tier (arcane_dust, some arcane_essence)
Several stat-axis options per slot so every build has a reachable enchant for each of its slots:
- **Weapon:** adds an Agility option beside the existing Might (str) and Spellpower (int).
- **Caster slots** (helmet, shoulder, neck, legs): Intellect options.
- **Healer:** a Chest Spirit option.
- **Physical/melee** (shoulder, waist, gloves, boots, ring): Strength/Agility options.
- **Tank:** Reinforcement (armor) enchants on helmet and chest, plus a Boots Stamina.
- **Ring:** Strength/Agility/Intellect beside the existing Spirit.

### Greater tier (arcane_shard + arcane_essence)
A stronger top-end enchant on the five highest-impact slots: weapon (Greater Might / Greater Spellpower), helmet (Greater Fortitude), chest and legs (Greater Stamina), gloves (Greater Agility). These are the **only** consumer of `arcane_shard`, the material an epic/legendary disenchant yields (`DISENCHANT_MATERIAL_BY_QUALITY` in `professions/enchanting.ts`). Before this change a shard was an unspendable dead-end currency.

## Notes

- **Data-only.** No mechanic or resolution-logic change: the existing `applyEnchant`/`disenchant` paths already handle multi-reagent costs and any stat bonus, so this is a pure expansion of the declarative table.
- Magnitudes follow the existing base-tier convention (primary ~4-6, sta ~6-10); Greater is a modest same-axis step up, never a gear-doubling jump. The owner can retune any value.
- Enchant `name` fields are not surfaced anywhere yet (Enchanting is not wired to a UI/vendor), so no i18n keys are involved.

## Testing

Adds an ENCHANTS table-integrity suite to `tests/professions_enchanting.test.ts`:
- every enchant is well-formed (id matches its key, a valid slot, real reagent item ids in positive counts, a real non-empty stat bonus using only categories `recalcPlayerStats` reads);
- every enchantable slot has at least one enchant (full slot coverage);
- `arcane_shard` has at least one consumer (the dead-end regression);
- a Greater (shard-consuming) enchant applies end to end and grants its bonus.

`npx vitest run tests/professions_enchanting.test.ts` (all green; exact count omitted per the anchor rule, counts drift), plus `tests/architecture.test.ts`, `tests/progression.test.ts`, `tests/localization_fixes.test.ts`, and `tests/localization_coverage.test.ts` green in isolation; `tsc --noEmit` clean for the changed files.

Companion PRD: #1925 specs this enchanting loop at the design level; its R6 (the arcane_shard dead-end) is delivered by this PR, and the PRD now cross-links back here.


jgyy (commented):
Solid, well-scoped data-only PR. It stays inside `src/sim/content/enchants.ts` (declarative content, no new sim logic, no new seam needed), reuses `ctx.removeItem`/`ctx.countItem` for the shard sink same as every other reagent (still server/sim-authoritative, nothing client-trusted), and adds real end-to-end coverage for the new Greater tier including the `arcane_shard` sink. Every reagent/item id referenced resolves to a real `ITEMS` entry; ran `tests/professions_enchanting.test.ts` locally, all green.

nit: the file header comment at the top of `enchants.ts` still says "Scope (v1): one simple, always-known enchant per equippable item slot," which now directly contradicts the "two layers" note this PR adds right below it. Worth updating so the next contributor isn't misled.

nit: PR body says the test file has "24 pass," actual run is 20. Harmless, but the anchor-rule in CLAUDE.md exists precisely to avoid citing counts that drift, so worth fixing in the description.

informational: companion PR #1925 is a PRD covering this same enchanting loop at the design level and doesn't reference this PR; not required for either to merge independently, but worth cross-linking for context.

ryan-foo:
Both nits fixed (eb31ae479): the enchants.ts header now documents the two-tier model up front (always-known per-slot basics on arcane_dust, plus the shard-consuming Greater tier as the sole arcane_shard sink) instead of the stale v1 one-enchant-per-slot scope, and the PR body's test count is corrected to the real 20 (the anchor rule is right, counts drift). Cross-linked the companion PRD #1925, which now also points back here as delivering its R6. Merged release/v0.27.0 cleanly (now the PR base); enchanting and progression suites 37/37, tsc clean.

PR #1952, opened 2026-07-15
fix(economy): let a linked-but-disconnected wallet buy Claudium on the crypto rails

ryan-foo:
## Summary

Revenue-blocking bug: a player who linked (verified) a Solana wallet sees their $WOC balance on the player card even while the browser extension is disconnected (the client has a dedicated "linked but disconnected" state), but the Claudium store's SOL/WOC rails read the balance only from the actively connected session wallet. So a linked-but-disconnected player saw, for example, 130k $WOC on their card while every crypto buy button stayed silently disabled: the store snapshot fed a null wallet balance, `buildClaudiumView` computed every rung as unaffordable, and the button greyed out with no explanation. This is the "has $WOC, can't buy" report.

## Root cause

`src/main.ts` claudium store snapshot:

```ts
const walletAddress = wallet.currentWallet().address; // connected session wallet only
const [solBalance, wocBalance] = walletAddress ? [...fetch...] : [{ lamports: null }, null];
```

`currentWallet().address` is null unless the extension is connected this session, so `wocBaseUnits` came back null and `claudium_view` (correctly) marked every WOC/SOL rung unaffordable and disabled the buttons, even though the player's linked balance was plenty.

## Fix

Fall back to the account's linked wallet pubkey when nothing is connected this session:

```ts
const walletAddress = wallet.currentWallet().address ?? linkedWalletPubkey;
```

The store then reads that wallet's on-chain balance and the button enables when affordable. The buy click already surfaces the existing prompt (the `signAndSend` path throws `"connect a wallet first"` when disconnected, caught into the `checkoutWalletRequired` toast: "Connect a Solana wallet before buying Claudium with crypto."), so the flow becomes enabled-button -> click -> connect-to-sign, instead of a dead button.

Safety: the fallback only fires when the connected address is null, so a connected player is unchanged, and a player who never linked a wallet still correctly sees disabled crypto rails.

## Scope / testing

One-line change to existing `main.ts` bootstrap wiring (which wallet funds a purchase). The downstream affordability logic is already covered by `tests/claudium_view.test.ts` (null balance -> disabled, present balance -> affordable); this change only ensures the snapshot feeds the linked balance instead of null. tsc and biome clean on the changed file. Recommend a manual smoke test: link a wallet, disconnect the extension, open the Claudium store, confirm the WOC/SOL rungs enable and the buy click prompts a connect.

## Follow-up (not in this PR)

Auto-open the connect flow on the buy click when disconnected, so the player does not have to reconnect manually before re-clicking. Also consider tolerating a cached balance (the snapshot forces a fresh fetch, which can be IP rate-limited to null and re-disable the button transiently).


jgyy (commented):
Confirmed this is benign. The change only widens which wallet address is used to read a balance for the Claudium store's affordability display, falling back to the account's server-verified `linkedWalletPubkey` (populated only from a bearer-authed `api.linkedWallet()` call) when no wallet is connected this session. The actual spend path, `wallet.signAndSendTransactionBase64` (`src/net/wallet.ts:414-417`), is untouched and still hard-requires an actively connected session wallet, throwing otherwise. No ownership or authorization check is loosened here, this is a read-side UX fix, not a security-relevant change to the settlement path.

nit: no test directly exercises the new fallback selection line itself (`wallet.currentWallet().address ?? linkedWalletPubkey`); existing `tests/claudium_view.test.ts` covers the downstream affordability logic but not this specific wiring. Small and low-risk enough not to block on, but worth a quick unit test pinning the fallback behavior.

ryan-foo:
Nit addressed in 8fb36e86b: the fallback selection is now pinned by unit tests. Since the line lived in main.ts (a coordinator), the testable unit moved behind the existing seam as claudiumBalanceAddress in the pure claudium view core, with main.ts a thin consumer, and tests/claudium_view.test.ts gains three cases: a connected session wallet wins, with no session wallet the server-verified linkedWalletPubkey is used, and with neither there is no balance read. Also merged release/v0.27.0 (now the PR base), resolving the one overlap with the new USDC rail by keeping both; claudium suite 13/13, tsc clean.

PR #1954, opened 2026-07-15
fix(render): correct ambient critter GLB forward axis and ground seating

### LINKED ISSUE #1862: Bug: Critters move sideways, trough the ground
### Before you start

### What happened?

The critters were mentioned in the release notes as `Rabbits, squirrels, songbirds, leaping fish, gathering nodes, the mailbox, and three delve props now use optimized Meshopt GLBs with procedural fallbacks.`. The updated models are cute, but I think they are not properly oriented. At least the bird and squirrel move sideways, and their model clips trough the ground.

### Steps to reproduce

1. Go near mirror lake
2. Observe critters moving sideways and trough the ground

Critters move forward and on the ground

### Where did this happen?

Online (multiplayer server)

### Device

Desktop

### Browser / OS

Linux desktop client

Image

jgyy:
## Summary
Fixes #1862 ("critters move sideways, trough the ground"). Inspected the shipped assets directly (`npx gltf-transform inspect`) and confirmed both parts of the report against real bounding-box data:

- **Sideways movement:** `squirrel_critter.glb`'s horizontal footprint is ~1.00 x ~0.43 (X x Z) — it was authored nose-to-tail along world X, not this module's Z-forward convention (the merged-primitive fallback body in `buildSpeciesGeo` is elongated along Z, and the per-frame heading rotation in `critters.ts` assumes that). `rabbit_critter.glb` (~0.76 x ~0.92) and `songbird_critter.glb` (~0.60 x ~1.00) already run Z-long, so this is genuinely species-specific, matching the report calling out squirrel and bird (not rabbit).
- **Clipping through the ground:** none of the three GLBs have their origin at the model's feet; their Y bounding boxes are vertically centered, not floor-anchored. `delve_props.ts` and `mailbox.ts` already guard against exactly this for their own standalone props via a Box3 re-seat (`inst.position.y -= seated.min.y`); `critters.ts` never applied the same pattern, per its own comment acknowledging the gap ("Unlike delve_props/mailbox, this is not Box3-normalized").

## Fix
- `seatAndOrientCreatureInstance()`: seats every loaded GLB's feet at y=0 (the same Box3 re-seat `delve_props.ts`/`mailbox.ts` already use) and bakes a per-species corrective yaw on load.
- The first pass tried a horizontal-bbox heuristic (`creatureForwardCorrectionYaw(extentX, extentZ)`) to guess nose-to-tail direction from footprint shape. Review on this PR found that heuristic wrong on two of the three species: `songbird_critter.glb`'s bbox is Z-long from wingspan even though the beak points +X (heuristic returned "no correction", leaving the bird broadside to its travel direction), and `squirrel_critter.glb`'s nose is at +X but the heuristic can only pick an axis, not a sign (it baked +PI/2 instead of the -PI/2 actually needed). Since this is a small, fixed set of exactly three species, the fix now drops the bbox measurement entirely and uses an explicit per-species yaw table instead, verified against the live scene: rabbit's nose already leads +Z (0), squirrel's and the songbird's both sit on +X (-PI/2 each).
- **Wrap in a group:** the first pass applied the correction directly to the loaded clone and returned it as `c.obj`, but the per-frame wander loop does a hard `c.obj.position.set(...)`/`c.obj.rotation.y = ...` every tick for world placement and heading. That write landed on the same object the correction had just been applied to and erased it immediately, making the fix a no-op. Fixed by wrapping the corrected clone in an outer `THREE.Group` (the `delve_props.ts`/`mailbox.ts` pattern): the loop transforms the group, the inner clone keeps its correction untouched underneath.

```mermaid
flowchart TD
  A["loadedSpeciesGltf.get(species)"] --> B["clone(true)"]
  B --> C["seatAndOrientCreatureInstance(species, inst)\nbakes per-species yaw + seats feet at y=0"]
  C --> D["new THREE.Group().add(inst)"]
  D --> E["c.obj = group"]
  E --> F["per-frame wander loop (every tick)"]
  F --> G["c.obj.position.set(x, groundY, z)\nc.obj.rotation.y = -heading + PI/2"]
  G -.writes the OUTER group only.-> D
  C -.correction stays intact on the INNER clone.-> C
```

## Screenshots
Before/after tour across five open, prop-free locations (open meadow, Eastbrook vale, the Mirefen causeway, the marsh edge, and a second Eastbrook clearing) in the offline client, same character/camera framing at each spot. Captured with `scripts/critter_tour_shot.mjs`, which pre-sets the `woc.cameraModePrompt.shown` localStorage flag, dismisses the new-adventurer tutorial popup, and lets the first-spawn intro cinematic finish (or explicitly skips it) before shooting, so every frame below is clean gameplay with no cinematic, tutorial, or camera-mode-choice overlay in it. Before is the pre-#1954 state (`812e4b223`, parent of the first fix commit); after is this PR's final per-species-table fix (`23dbb3862`). Wide shots show the ambient pool; closeups are cropped to the nearest tracked critter(s) (rabbit/squirrel/bird mix) at each spot, confirming nose-forward heading and feet-on-ground seating post-fix versus the broadside/sunk pre-fix state.

| Location | Before | After |
|---|---|---|
| Meadow (wide) | before | after |
| Meadow (closeup 1) | before | after |
| Meadow (closeup 2) | before | after |
| Eastbrook vale (wide) | before | after |
| Eastbrook vale (closeup 1) | before | after |
| Eastbrook vale (closeup 2) | after only (no in-frame critter before the fix at this crop) | after |
| Mirefen causeway (wide) | before | after |
| Mirefen causeway (closeup) | before | after |
| Marsh edge (wide) | before | after |
| Marsh edge (closeup) | before | after |
| Eastbrook clearing (wide) | before | after |
| Eastbrook clearing (closeup 1) | before | after |
| Eastbrook clearing (closeup 2) | after only (no in-frame critter before the fix at this crop) | after |

## Test plan
- [x] Added `tests/critters.test.ts` regression cases pinning `creatureForwardCorrectionYaw` against the per-species yaw table (rabbit 0, squirrel/songbird -PI/2 each), replacing the stale bbox-heuristic pins that had baked in the buggy outputs as "expected".
- [x] Added decisive coverage on the actual property #1862 cares about: a moving critter's nose has a strictly positive dot product with its measured velocity, driven through the real per-frame wander/heading update path for all three species, plus a sink=0 assertion on the ground-seat fix.
- [x] Added a regression test for the group-wrapper fix using an injected fake GLB scene (`critterPreloadInternalsForTest.setLoadedForTest`), asserting the baked correction survives 30 wander ticks. Verified test-first at each stage: reverted the fix locally, confirmed the corresponding test fails, then restored it and confirmed green.
- [x] `npx vitest run tests/critters.test.ts` (11 passed).
- [x] `npx tsc --noEmit -p .` clean.
- [x] `npx @biomejs/biome check` on all changed files (clean).
- [x] Merged the branch up to `release/v0.27.0` (`7433953c1`) and re-verified.
- [x] Manually verified in the live offline client across five zones with before/after screenshots above (24 images total), each captured with the intro cinematic, tutorial popup, and camera-mode-choice prompt confirmed dismissed beforehand.


FernandoX7 (commented):
Nice detective work on the assets, and the ground-seating half is solid. I confirmed it in the live offline client: on base every critter dips exactly half its height below its ground anchor (rabbit 0.50, squirrel 0.42, bird 0.38 yd), and with this change all three re-seat to sink = 0. Adopting the `delve_props.ts`/`mailbox.ts` Box3 pattern is the right call, and the group-wrapper in the second commit genuinely holds: I watched the inner clone keep `rotation.y = PI/2` across 30 wander ticks while the outer group took the per-frame writes.

The forward-axis half does not hold up, though, and it misses the part of #1862 that named the bird.

**blocking** `src/render/critters.ts:136` The bird still moves sideways. `creatureForwardCorrectionYaw` infers the forward axis from the horizontal bounding box, but a bounding box measures the widest silhouette, not the nose-to-tail axis. `songbird_critter.glb` is Z-long (0.60 x 1.00) because of its **wingspan**, while its beak actually points along +X. So the heuristic returns 0 and the bird is left broadside to its travel direction. I measured it in the live client (inner-clone nose axis vs velocity): dot = 0.0, i.e. exactly perpendicular. Issue #1862 calls out "the bird and squirrel" by name, so this is the case the fix most needs to cover. Top-down (blue arrow = travel/+Z): https://gist.githubusercontent.com/FernandoX7/a68a20be664e99b886522d96b581bc1b/raw/bird_top.png

**blocking** `src/render/critters.ts:136` The squirrel is corrected to face backwards. `squirrel_critter.glb`'s nose is along +X (head, eye, nose, front paw all at +X; the bushy tail is the high mass at -X). The module's per-frame write sends local +Z along travel, so the nose needs to land on +Z, which is a -PI/2 correction. The PR bakes +PI/2, which puts the nose on -Z: same axis, wrong sign, so it glides tail-first. Live measurement: nose-vs-velocity dot = -1.0 (exactly antiparallel). Top-down (left = this PR, +PI/2; right = -PI/2): https://gist.githubusercontent.com/FernandoX7/a68a20be664e99b886522d96b581bc1b/raw/squirrel_forward.png

Both come from the same root: extents can tell you which axis is longer, but not which end is the nose (sign), and not whether the long axis is even the body (the bird's is wingspan). For a fixed set of three models I'd lean on an explicit per-species yaw rather than a derived one, e.g. `{ rabbit: 0, squirrel: -Math.PI / 2, bird: -Math.PI / 2 }`, or better, re-export the two GLBs nose-along-Z so no runtime correction is needed at all. I verified in closed form and against the live scene that -PI/2 lands the squirrel and the bird nose-first (dot = +1.0), and the rabbit is genuinely already correct: its head leads +Z (dX about 0, dZ +0.09), so leaving it at 0 is right.

**should-fix** `tests/critters.test.ts:73` The regression tests pin the buggy outputs as expected (squirrel to +PI/2, songbird to 0), so they stay green while the critters face the wrong way. Nothing asserts the actual property the issue is about, which is that a moving critter's nose points along its velocity. A test that builds the instance, applies a known heading through the real per-frame path, and checks nose-dot-velocity is positive would have caught both of the above (and would still cover the seating: assert sink = 0).

Minor, not blocking: the `hudChrome.gathering.notReady` line and the larger generated-keys delta are just stale-base regen churn (the branch is behind release/v0.27.0). I simulated the merge and it reconciles cleanly: the key appears once, no duplicate union members, the devCommand keys come back from the release side. A merge plus `npm run i18n:gen` will sort it out; nothing to do here.

jgyy:

jgyy:
Thanks for the review, addressed all three points.

1. Bird broadside (blocking): fixed. The bbox heuristic was the wrong tool: it measures widest silhouette, not nose-to-tail, and songbird_critter.glb's bbox is Z-long from wingspan even though the beak is on +X. Dropped creatureForwardCorrectionYaw's bbox measurement entirely and replaced it with an explicit per-species yaw table (fixed set of exactly three species, so a table is simpler and correct where the heuristic was not): rabbit 0, squirrel -PI/2, bird -PI/2.

2. Squirrel tail-first (blocking): fixed by the same change. The old heuristic could only pick an axis (long side), never a sign, so it baked +PI/2 when the correct correction is -PI/2 (nose at +X needs -PI/2 to land on +Z, not +PI/2, which is the same axis but sends it backwards). The table above bakes the signed value directly.

3. Test gap (should-fix): fixed. Replaced the stale test cases (which pinned the buggy +PI/2 for squirrel and +0 for bird as "expected") with:
   - Updated pins on creatureForwardCorrectionYaw for the new per-species table.
   - A new regression suite that builds the instance, drives a known raw per-species nose axis through the real per-frame wander/heading update path for all three species, and asserts nose-direction dot velocity > 0 (would have caught both the dot~0 and dot~-1 bugs).
   - A decisive sink=0 assertion (model's own local bounding box bottoms out at y=0 post-seat, across all three species, surviving per-frame writes).

Also merged current release/v0.27.0 into the branch and ran npm run i18n:gen (produced no diff, so the earlier hudChrome.gathering.notReady churn is already reconciled) and npm run sfx:manifest (the merge brought in new wand-cue audio the generated manifest hadn't picked up yet, which left tsc red independent of this PR's own change).

Verification run: npx tsc --noEmit (clean), npx vitest run tests/critters.test.ts (11/11 passing), npx vitest run tests/architecture.test.ts (24/24 passing), npx @biomejs/biome check --write on both changed files (no fixes needed, clean).

FernandoX7 (approved):

PR #1955, opened 2026-07-15
fix(dungeons): boss adds anchor where they erupt, exit-portal threat wipe, heroic damage rebalance

ryan-foo:
## Summary

Fixes the boss-add aggro/leash bug and rebalances heroic dungeon damage.

**The add leash bug (Velkhar, Korzul's sanctum, Thunzharr).** `spawnBossAdds` gave every summoned add the BOSS's original `spawnPos` as its leash home. A boss kited past the leash radius (45yd outdoors, 70yd in dungeons) hatched adds that were instantly past their own leash: the first engaged tick flipped them to evade with a wiped hate table, so they ran home without ever swinging and players could kite add waves indefinitely. Adds now anchor where they ERUPT (they keep `createMob`'s spawnPos at the eruption point beside the boss) and seed `leashAnchor` the same way `aggroMob` does on a normal pull. Kiting an add is now bounded by the normal leash rules measured from the fight. This covers Grand Necromancer Velkhar's bonewalkers, Vael's thralls, Thunzharr's stormlings, and the delve raise-dead channel (all route through `spawnBossAdds`). The Nythraxis raid is untouched (its adds use a `canLeash: false` profile and its own spawn path).

**Exit-portal threat wipe.** Players could dance in and out of the dungeon exit to kite pulls: threat persisted across the teleport. `leaveDungeon` now scrubs the leaver (and anything they own, e.g. their pet) from every instance mob's hate table via a new `dropThreat` helper in the pure `threat.ts` leaf, releasing taunt locks and clearing aggro so the mob re-targets the remaining party next tick or evades home. Re-entering means earning aggro from scratch.

**Heroic damage rebalance.** Live heroics proved overtuned:

- The four five-man `damageMultiplier` values drop 25% (hollow_crypt 6.8 to 5.1, sunken_bastion 6.2 to 4.65, drowned_temple 5.7 to 4.3, gravewyrm_sanctum 5.4 to 4.05). The calibration target moves from ~20% to ~15% of the reference geared shaman's max hp per elite trash swing, final bosses from ~22-24% to ~17-18%. This also softens the guard adds flanking Korzul and Velkhar on their pulls.
- Boss-SUMMONED add waves additionally swing at a new per-dungeon `addDamageMultiplier` (~half the trash value), because an add pack lands on top of the boss's own swings and mechanics. `spawnBossAdds` passes `{ summonedAdd: true }` and the same multiplier is stamped on the adds' fire-time mechanics (`mechanicDamageMult`). On heroic Velkhar, each bonewalker now swings for roughly 4% of a geared tank's hp instead of 10%.
- The Nythraxis raid arena keeps its tuning: its `addDamageMultiplier` mirrors `damageMultiplier` and is inert (raid adds spawn through the encounter script).

(A one-line regen of `translation_keys.generated.ts` originally rode this branch to fix pre-existing base staleness; the retarget merge to `release/v0.27.0` made it a no-op because the release branch already carries the same line, so it no longer appears in the diff. `tsc --noEmit` and the i18n equivalence suite are green on this head.)

## Type of change

- [x] Bug fix
- [x] Feature: new functionality (per-dungeon summoned-add tuning knob)

## How was this tested?

- Commands:
  - `npx vitest run tests/boss_add_leash.test.ts` (new suite: kited-boss adds engage instead of instant-evading, anchor at the eruption point, heroic adds use `addDamageMultiplier`, normal difficulty untouched)
  - `npx vitest run tests/world_boss.test.ts` (Thunzharr stormlings: new kited-boss regression case)
  - `npx vitest run tests/dungeons.test.ts` (exit threat wipe: solo scrub + party re-target cases)
  - `npx vitest run tests/threat.test.ts tests/dungeon_difficulty.test.ts` (dropThreat unit tests; the retuned multiplier ladder and summoned-add transform pinned to exact literals)
  - `npx vitest run tests/parity` (green with NO golden regen: the change draws no rng and shifts no draw order)
  - `npm run gate` (full suite: the only failures are pre-existing on a clean origin/release/v0.26.0 checkout, verified by running the same files against an untouched base worktree)
- Manual steps: none (sim-only change; QA gate + architecture review dispatched over the diff, zero blocking findings).

### Quality

- [x] **The full gate passes** modulo failures reproduced bit-identically on the clean base branch (details in the testing section). Decisive tests added for every changed behavior.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A (no UI change; sim runs identically on all three hosts by construction, parity gate green)

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the generated-file change is a regen of a pre-existing catalog key, not a new string).

### Hygiene

- [x] Generated files were regenerated by their owning build step (`npm run i18n:gen`), not hand-edited.


jgyy (commented):
Genuinely solid, well-scoped bug fix. The add-leash root cause is correct and verified against the existing `aggroMob` anchoring idiom: the new `spawnBossAdds` code mirrors that same pattern, and `createMob` already anchors `spawnPos` at the eruption point, so removing the old override that copied the boss's own spawnPos onto the add is precisely the right minimal fix. The exit-portal threat-wipe fix is properly server-side, and the tests are genuinely test-first: `tests/boss_add_leash.test.ts` and the `world_boss.test.ts` addition reproduce the exact "kited past leash radius" scenario, and `tests/dungeons.test.ts` covers both solo and party-retarget-on-leave cases.

should-fix: the PR body claims a one-line regen of `translation_keys.generated.ts` fixing pre-existing i18n/tsc staleness on the release branch, but that file isn't actually in the diff (10 files changed, none of them that path). Either the body is stale from a rebase or the regen never got committed, worth confirming `tsc` and the i18n equivalence test are actually green on this exact branch before merge.

nit: the heroic damage-multiplier retunes are described as a uniform 25% cut / "roughly half" for adds, but a couple of the four dungeons land a fraction off exact (drowned_temple is 24.6%, not 25%). Reads like intentional calibration rounding rather than a mistake, just worth a one-line comment acknowledging the rounding so a future contributor doesn't "fix" it back to the exact fraction.

No blocking issues. Merge-tree against release/v0.27.0 is clean.

ryan-foo:
Both follow-ups addressed. The stale generated-file claim in the body is corrected: the one-line translation_keys regen originally rode this branch, and the retarget merge to release/v0.27.0 made it a no-op (the release branch already carries the same line), which is why it vanished from the diff; tsc and the i18n equivalence suite are green on this head. The rounding nit now has a comment in dungeon_difficulty.ts (f04c68776) stating the tidy multipliers are deliberate calibration rounding (drowned_temple is a 24.6% cut on purpose), so nobody corrects them back to exact fractions.

EnriqueGF (commented):
Reviewed the full diff at f04c687 against release/v0.27.0 and verified the load-bearing claims in the code. This is a solid fix: the eruption-point anchor matches how aggroMob seeds a normal pull, the exit scrub goes through a tested pure helper in threat.ts, and the test suites are decisive (literal pins on the retuned ladder, a real kited-boss regression case for both Velkhar and Thunzharr, solo scrub plus party re-target cases).

What I verified, not just read:

- `tsc --noEmit` green; `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` green; `tests/parity` green with no golden regen (the change adds no rng draw and moves none: the one draw near the edit, `rng.int` on the base template at the top of the spawn loop, is untouched).
- All five touched suites pass: boss_add_leash, dungeon_difficulty, threat, dungeons, world_boss (plus guide freshness, since the tuning table could plausibly have fed the wiki; it does not).
- Summoned adds are pushed into `inst.mobIds` (sim.ts:5744), so the exit scrub covers boss add waves too, not just the spawn list.
- After the scrub, a mob with an empty table resolves next tick via updateMobTarget -> retargetMob -> evade, and a mob with remaining party threat re-targets; both arms are pinned by the new dungeons tests.
- The Nythraxis claims hold: no `summonAdds` record spawns a raid add, the encounter script still sets `add.spawnPos = boss.spawnPos` itself, so `findNythraxisBossForAdd`'s spawnPos-proximity fallback (encounters/nythraxis.ts:200) is unaffected, and the mirrored `addDamageMultiplier: 2.0` is genuinely inert.
- The optional `role` param keeps every existing call site's semantics; merge against the current release/v0.27.0 head (91d7b86) is conflict-free and the base drift there (reset locks, freeInstance) does not touch leaveDungeon or the threat paths.
- No banned dashes or emojis in the added lines; determinism is clean (the `[...mob.threat.keys()]` snapshot iterates in insertion order and is safe against delete-during-iterate).

Findings, none blocking:

- nit, src/sim/instances/dungeons.ts:360: the scrub fires only on the leaveDungeon path. Any other way a live player leaves a claimed instance (an arena pull, a dev teleport) keeps their threat entries and any taunt lock until the leash check evades the mob or the lazy prune drops them. Self-healing and not repeatable the way the door dance was, so fine to leave; noting it in case a future exit path lands.
- nit, src/sim/sim.ts:5740: a slain non-worldBoss summoned add keeps a normal in-place respawn timer (combat/damage.ts, the respawnTimer assignment), and with the re-anchor it now respawns at the eruption point rather than the boss's home. For the outdoor summonAdds bosses (temple, zones 1 to 3) a kited fight can leave adds re-populating the kite spot. Adds respawning at all is pre-existing; this PR only moves the point, but the outdoor summoners may want the worldBoss-style despawn-on-death treatment in a follow-up.
- nit, src/sim/instances/dungeons.ts:388: a taunt lock held by the leaver's already-dead pet (entry lazily pruned, id still in forcedTargetId) is not released by the scrub; updateMobTarget clears it next tick because the forced entity is dead, so no stuck state, just asymmetric coverage.

Nice touch on the calibration-rounding comment in the tuning table; it preempts exactly the kind of "fix the fraction" churn those numbers invite.

ryan-foo:
@EnriqueGF appreciated, especially the Nythraxis-fallback and iteration-safety verification. All three nits acknowledged as-is: the scrub stays scoped to the leaveDungeon path (other exits self-heal via the leash/prune machinery and are not repeatable the way the door dance was), the outdoor summoners' respawn-at-eruption-point behavior is noted for a follow-up (giving them the worldBoss-style despawn-on-death treatment is the likely shape), and the dead-pet taunt-lock asymmetry is covered by updateMobTarget clearing the forced entity next tick. If the outdoor-summoner follow-up gets picked up, it lands as its own PR.

PR #1958, opened 2026-07-15
docs(prd): pet-attributed damage and per-source meter breakdown (Recount-style)

ryan-foo:
## Summary

Draft PR carrying the **PRD only** (no code yet, pending owner sign-off) for a Recount-style change to the damage meter:

1. **Pet damage counts under its owner.** A hunter/warlock/mage pet's damage folds into the owning player's meter total instead of showing as its own separate row (which is how it appears today, since the pet's `sourceId` gets its own tally).
2. **Hovering a player's row shows a per-source breakdown.** Hovering "Swifter" (a hunter) shows `Auto Shot`, `Fell Shot`, ... and a single **Pet** category, each with its share of the player's total; mages read the same way (`Rimelance`, `Frostbolt`, ...). It feeds entirely from the combat log (the existing `damage` `SimEvent`).

The PRD is at `docs/prd/pet-and-source-damage-breakdown.md`.

## Why it is small

It is **client-only**: no sim, server, wire, or `IWorld` change. The `damage` event already carries `sourceId`, `ability` (the ability name), and `amount`, and pet damage events already reach the client with the pet's `ownerId`. The change is:

- `MeterData.onEvent` attributes a pet's damage to `ownerId` (so the owner's total includes it) and records a per-source key (`ability` name, an `AUTO` sentinel for white hits, a `PET` sentinel for pet damage) on a new `MemberTally.bySource` map.
- A new pure `src/ui/meters_breakdown_view.ts` (`*_view`, Node-tested) turns a tally into sorted breakdown rows with percents.
- The existing `attachTooltip` (desktop hover + keyboard focus + mobile long-press) is injected into `Meters` and attached to each damage-tab row.
- New `hud.meters.*` i18n keys (`petDamage`, `breakdownTitle`), reusing `hud.combat.attack` for the auto-attack label.

## Open questions for the owner

- Pet damage aggregates to one **Pet** line in v1 (pet hits carry `ability: null`). Per-pet-ability rows would need the sim to name pet damage; the PRD lists it as future work. OK to ship the aggregated Pet line first?
- Breakdown is on the **damage tab only** (threat reads the live hate table, not cumulative damage; `heal2` has no ability field). Agree?

## Status

Draft. This PR is the PRD; I will not start the implementation until the design is approved. The PRD documents the data model, module layout, edge cases (online interest scoping, DoT ticks, mob mechanic labels), the test plan, and acceptance criteria.


jgyy (commented):
Once you get past the diff view, this is a solid draft PRD: it follows the repo's docs/prd/ convention (status table, goals/non-goals/design/edge-cases/testing/acceptance-criteria), correctly scopes the change as client-only, and grounds itself in real code (`MeterData.onEvent`, `partyPids`, `abilityDisplayNameFromSource`, `Hud.attachTooltip`) with a sensible pure-core proposal (`buildDamageBreakdown` as a `UI_PURE_CORES` candidate) and i18n awareness.

blocking: the committed file has two literal NUL bytes (offsets 5450 and 5461) inside the sentinel-key example in section 5.1 ("PET" and "AUTO" reserved keys). That's why `gh pr diff` shows this as a binary file with 0/0 additions/deletions instead of a text diff, which also means no line-level review comments can be attached until it's fixed. Please strip those bytes (looks like a stray control character landed where a plain `"pet"`/`"auto"` was intended) and re-push, then a normal line-level pass can happen.

Everything else about the doc's reasoning looks fine; git history is a normal single commit cleanly on top of release/v0.26.0, no branch divergence involved, this is purely the encoding glitch.

ryan-foo:
Binary-file blocker fixed in 2ed45d3d2: two literal NUL bytes had crept into the section 5.1 sentinel-key example (a shell turned the typed backslash-zero escapes for the pet/auto reserved keys into real NULs). Restored the literal two-character escape text, so the PRD now renders as a normal 136-line text diff and line-level review works again. Also merged release/v0.27.0 in cleanly; the diff against the base is the one PRD file.

PR #1761, opened 2026-07-10
feat(mobile): redesign the compact touch HUD

No898:
## Summary

Redesigns the mobile touch HUD around the actual input frequency and the camera-look surface:

- replaces the radial action ring with a compact two-row pad: A1/A2 Attack/Target, then page switch plus A3/A4/A5 Jump/Use;
- makes Jump contextual Interact near NPCs, loot, delve objects, and other supported interactables while retaining dedicated Attack;
- preserves the joystick-integrated Autorun from the v0.24 release and reserves the opposite thumb zone for swipe-look or the optional view joystick;
- reflows the minimap, direct menu, compact Party row, subordinate Target frame, larger player frame, pet commands, and Meters for landscape phones and tablets;
- places the Consumables disclosure immediately inward of the movement joystick, expanding toward the player frame and wrapping upward, with mirrored left-handed behavior;
- adds deterministic browser geometry, safe-area, maximum-scale, interaction-flow, target-size, overlap, and screenshot coverage.

The implementation follows [the mobile HUD PRD](https://github.com/No898/world-of-claudecraft/blob/mobile-layout-adjustments/docs/prd/mobile-compact-landscape-hud.md) and [implementation plan](https://github.com/No898/world-of-claudecraft/blob/mobile-layout-adjustments/docs/plans/2026-07-10-mobile-touch-hud-layout.md).

Portrait gameplay layout is intentionally out of scope: the integrated `release/v0.24.0` behavior displays the rotate-device gate instead.

## Related issues

- Follow-up to #1731.
- Integrates and preserves the joystick Autorun work from #1724.

## Type of change

- [x] Feature: new functionality
- [x] Bug fix
- [ ] Refactor or performance (no behavior change)
- [x] Documentation
- [x] Tests
- [x] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npm run gate` - all 9 steps green; 11,550 tests passed and 63 skipped.
  - `npx vitest run tests/client_shell.test.ts tests/mobile_controls.test.ts tests/mobile_hud_layout.test.ts tests/mobile_hud_layout_applier.test.ts tests/mobile_action_ring_painter.test.ts tests/party_frames_painter.test.ts` - 251/251 passed during release integration.
  - `npx vitest run --config vitest.browser.config.ts tests/browser/target_size.browser.test.ts` - 12/12 passed.
[...]

PR #1960, opened 2026-07-15
feat(mobile): expand action ring to 20 skill slots

No898:
## Summary

Expands the mobile action ring from two pages to four pages while keeping five visible action buttons per page.

- Maps mobile pages to hotbar slots 1 to 20.
- Keeps the fixed Attack control outside paging.
- Leaves hotbar slots 21 and 22 desktop-only.
- Reuses the existing 22-slot persistence model, so no data migration is required.
- Updates Spellbook page labels and consumer-level HUD regression coverage.

This ports the focused 4 x 5 paging behavior identified in #1761 without bringing over its unrelated changes or optional slot-picker work.

## Related issues

Related to #1761.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm ci --no-audit --no-fund`
  - `npx tsc --noEmit` under Node 22.22.2
  - `npx vitest run tests/mobile_action_page_view.test.ts tests/mobile_action_ring_painter.test.ts tests/spellbook_view.test.ts tests/spellbook_window.test.ts tests/action_bar_controller.test.ts` (91 passed)
  - `npx vitest run tests/architecture.test.ts` (24 passed)
  - `biome ci --diagnostic-level=error` for all seven changed TypeScript files
  - production Docker image build, then HTTP 200 from `/` and `/api/status`
- Manual steps:
  - Compared the previous two-page mobile ring with the feature build in an 844 x 390 touch layout.
  - Cycled the feature build to page 4 and confirmed the page indicator shows `4`.
  - Confirmed ring controls remain at least 44 x 44 px.
  - Checked the 390 x 844 portrait rotate preflight.

`npm run gate` completed i18n freshness, malware, Biome, and SFX stages, then its full Vitest stage exposed unrelated host failures. Under the required Node 22 runtime, 195 of 197 tests across the seven initially failing files passed; only two existing fake-Docker watchdog timeout assertions remained red in `tests/deploy_watchdog.test.ts`. The changed behavior tests, architecture guard, typecheck, production Docker build, and scoped pre-push QA checks are green.

## Screenshots / recordings

Before, release/v0.27.0 baseline at commit 6f5976dda, with the two-page ceiling:

Before: mobile page 2

After, fourth page available:

After: mobile page 4

Portrait preflight:

After: portrait rotate preflight

---

### Quality

- [ ] The full gate passes.
- [x] Decisive tests cover the changed behavior.
- [x] Typecheck and changed-file lint pass.
- [x] The production Docker build succeeds.

The full gate remains unchecked because of the two unrelated host-specific fake-Docker watchdog failures documented above.

### Cross-platform and accessibility

- [x] Tested on desktop and mobile paths.
- [x] Tested in mobile landscape and portrait layouts.
- [x] Touch targets remain at least 40 x 40 px.
- [x] Existing accessible names and interaction semantics are preserved.

### Localization

- [x] Not applicable: this change adds no user-facing strings.

### Hygiene

- [x] No secrets or credentials are included.
- [x] Generated files were not hand-edited.
- [x] New numbers do not require localized formatting.


jgyy (commented):
Clean, tightly scoped extraction. `mobile_action_page_view.ts` is exactly the pure-core the repo asks for: page count, slot-to-page mapping, and clamp/cycle math are all DOM-free and reused by both the ring painter and the spellbook view instead of being hand-rolled twice. The persistence claim checks out too: `hotbar_form_persistence.test.ts` has zero diff, so this really is a UI-visible-slice change over the existing 22-slot model, not a schema change wearing a UI hat.

Boundary coverage is solid: `sourceSlotsForMobilePage` is pinned at the page 2/3 edges (`[11-15]`, `[16-20]`), the four pages are asserted disjoint and covering exactly 1-20, `clampMobilePage` is tested against negative/overflow/NaN, and slots 21-22 are explicitly asserted to fall outside all four mobile pages so they stay desktop-only.

One thing I couldn't confirm from the diff: the PR body claims ring buttons were manually verified at 44x44px, but nothing in this diff touches button geometry or CSS (only the two `.ts` slot-mapping files changed). That's expected since the per-page button count didn't change, so it's very likely still true, just not something this diff itself demonstrates — nit, not a blocker.

No i18n concerns (the page-label keys are pre-existing parameterized strings, untouched here), and `npx vitest run` on the five affected test files plus `architecture.test.ts` is green (105/105).

No898:
@jgyy Thank you for the thorough review and for validating the extraction and persistence assumptions.

We followed up with an additional internal review and strengthened the regression coverage in commit `c86b06cef`:

- exercised the same mobile action-ring button across source slots 1, 6, 11, and 16, including page wraparound and an explicit page 4 / source slot 20 assertion;
- pinned the production HUD action, ability, and item getters to the active mobile page;
- extended Sim and ClientWorld parity coverage with an ability in source slot 20 and an explicit `mobilePage === 3` assertion;
- added a production save/load round trip for a real ability in source slot 20 while confirming slots 21 and 22 remain reserved and empty.

The focused suite is green at 115/115 tests, with TypeScript, Biome, and the pre-push QA floor also passing. Thanks again for taking the time to review this!

EnriqueGF (commented):
Nice focused port of the 4x5 paging from #1761. The constant change flows cleanly through the already-parameterized consumers: `hud.ts` passes `mobilePageCount()` into the painter and resolves every getter through `sourceSlotForMobileButton` at tick time, `clampMobilePage`/`nextMobilePage` derive from the same constants, and the slot math is right (source slot 20 lands on `hotbarActions[19]`, slots 21-22 stay desktop-only). No new player-visible strings: the indicator key already takes `{page, count}`. Test additions are decisive (disjoint-coverage over all four pages, the page-3 slot-20 mapping, the persistence round-trip).

What I ran at the PR head: `npx tsc --noEmit` (clean), the five changed test files (91 passed), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `i18n:gen`), and `tests/world_api_parity.test.ts` (all green).

One finding:

- **blocking (stale base, mechanical fix):** `release/v0.27.0` has since merged #1964, which extracted `ActionBarController` out of `hud.ts` and deleted `tests/hotbar_form_persistence.test.ts` (replaced by `tests/action_bar_controller.test.ts`). This PR modifies the deleted file, so the merge hits a modify/delete conflict. I test-merged the current base into the PR head keeping your file: all 14 tests in it fail, because the harness's `Object.create(Hud.prototype)` setup now hits the `actionBarController` delegation (`hud.ts:886`, `this.actionBarController.replaceActions`). Fix: rebase onto the current `release/v0.27.0`, drop `tests/hotbar_form_persistence.test.ts`, and port your new slot-20 round-trip test into `tests/action_bar_controller.test.ts` (its harness drives the controller's save/load directly, so the port is a few lines).

I verified the rest of the PR survives that refactor: every `hud.ts` string your ring-painter source-scan tests assert is intact in the base's post-refactor `hud.ts`, and on the merged tree the other four test files pass (89 passed, `tests/action_bar_controller.test.ts` included). So after the rebase and the one test port this should be green as-is.

No898:
@EnriqueGF Rebased onto the current `release/v0.27.0` and addressed the blocking stale-base conflict in `5537ca9fb`.

- kept the base deletion of `tests/hotbar_form_persistence.test.ts`;
- ported the source-slot-20 persistence round trip into `tests/action_bar_controller.test.ts`;
- the port now exercises the canonical controller seam directly through `saveActions()` and `init()`;
- no production-code changes were needed for the conflict resolution.

Fresh verification after the rebase: 91/91 focused feature tests, 115/115 including the architecture suite, 53 guard tests passed with 3 standard skips, TypeScript clean, scoped Biome clean, and copy/diff checks clean. GitHub now reports the PR as mergeable.

Thank you for pinpointing the `ActionBarController` migration and confirming that the rest of the PR survived the refactor.

PR #1962, opened 2026-07-15
test(i18n): Phase 3 QA staled-slice freshness probe (never merges)

FernandoX7:
Deliberately stales the committed TranslationKey union and resolved slices (one catalog key added, no regen) to re-verify the relocated freshness diff in the pr-checks job still reds on a staled slice, per docs/toolchain-modernization/phase-03-qa.md STEP 2. Pushed with --no-verify because the pre-push floor correctly blocks staled slices locally. Will be closed unmerged immediately after the run is observed; precedent: PR #1946 (run 29388336786).


FernandoX7:
Probe observed: run 29413049770 pr-checks failed exactly at the freshness step (Verify committed i18n artifacts are fresh) with legible hunks naming the staled union line (hudChrome.gathering.qaFreshnessProbe) and the per-locale slices; i18n:gen and the coverage summary green; typecheck and builds never ran. Closing unmerged per the probe protocol (precedent PR #1946).

PR #1966, opened 2026-07-15
feat(arena): end-of-match scoreboard screen for every bracket

ryan-foo:
## Summary

Adds a WoW-style end-of-match scoreboard that opens when a **1v1, 2v2, Fiesta, or Protect Yumi** match ends. It summarizes the bout: every fighter's killing blows, damage done, and healing done, plus the local rating change and honor earned. Ranked brackets (1v1/2v2) show the Rating and Change columns and the "Your Rating: X to Y" line; unranked brackets (Fiesta/Yumi) drop the rating columns and just show the honor earned.

The banners and combat-log lines that already fired on a match end are kept; this only adds the modal on top for the local player.

### How it works
- **Sim:** `ArenaMatch` now carries a per-player stats map (killing blows / damage done / healing done) accumulated in the damage and heal cores. `endArenaMatch` scores both teams, then emits a full scoreboard on the existing `arenaEnd` event (`scoreboard` / `myTeam` / `honor`) to each participant, so a player's screen can resolve ally vs enemy client-side. The event forwards verbatim over the wire, so the offline and online hosts render the same screen.
- **UI:** a pure `buildArenaEndView` core (DOM-free, Node-tested; allies-first then damage-descending sort) plus a thin `arena_end_window` painter on the dialog-root seam. `Hud` opens it for the local player from the `arenaEnd` handler and dismisses it on **Leave** or **Escape**.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/arena_end_view.test.ts tests/architecture.test.ts` (new view core + pure-core registration)
  - `npx vitest run tests/arena.test.ts tests/snapshots.test.ts tests/localization_fixes.test.ts` (sim behavior, wire, S3 guard)
  - `npx vitest run tests/parity` (goldens regenerated for the extended `arenaEnd` event: only the `arena_1v1` / `arena_2v2_wipe` state+events hashes moved, no rng draw-order change)
  - `npx vitest run tests/i18n_completeness.test.ts tests/localization_coverage.test.ts` (M16 fills + coverage)
  - `npx vitest run tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts`
  - `npx tsc --noEmit` (clean for touched files)
- Manual steps: rendered the shipped painter markup + the shipped `#arena-end-window` CSS via `scripts/arena_end_screen_shot.mjs` across every result and bracket type below.

## Screenshots / recordings

Ranked 2v2 Victory (desktop):

ranked victory

Unranked Fiesta Defeat (rating columns dropped):

fiesta defeat

Ranked 1v1 Draw (rating shown, no change):

draw

Unranked Protect Yumi 5v5 Victory (six fighters):

yumi victory

Phone width:

mobile

## Checklist

- [x] i18n: new `hudChrome.arenaEnd.*` English keys added to the catalog, with the five non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) in the same change (M16).
- [x] New pure view-core registered in `UI_PURE_CORES`; painter on the dialog-root seam, not a new `hud.ts` banner section.
- [x] Sim change covered by tests; parity goldens regenerated deliberately.
- [x] No new wire command; the `arenaEnd` event gains fields only and forwards verbatim.


jgyy (commented):
Clean, well-scoped feature that follows the repo's seams correctly: stat accumulation is server-authoritative (`bumpArenaMatchStat` in `src/sim/social/arena.ts`), the new scoreboard is a genuine pure-core-plus-thin-painter split (`arena_end_view.ts` DOM-free and in `UI_PURE_CORES`, `arena_end_window.ts` the thin painter), the event rides the existing wire forwarding with no server.ts changes needed, and i18n additions are English-only in the right catalog.

should-fix: `index.html` adds `#arena-end-window`, but `play.html` never gets the matching element. `play.html` still only has `#arena-window` and loads the same bundle, and `Hud.openArenaEndScreen` silently no-ops when the root element is missing, so `/play` players simply won't see the scoreboard after a match ends. This is the recurring play.html chrome-parity miss the repo's own CLAUDE.md calls out.

should-fix: pet damage/healing isn't credited to the owner's scoreboard row, while pet killing blows are. `damage.ts`/`heal.ts` gate `damageDone`/`healingDone` bumps on `source.kind === 'player'`, but `handleDeath` correctly resolves killing-blow credit through `killer.ownerId` for a pet. Since pets are ordinary mob entities with `ownerId` set and nothing in arena.ts excludes them from a bout, a hunter or warlock's pet damage in a 2v2/Fiesta/Yumi bout is silently missing from two of the three stat columns on the same screen while showing up in the third.

should-fix: the actual sim-side wiring (`bumpArenaMatchStat`, the stats map, `endArenaMatch`'s row construction, the pet-owner killing-blow credit) has no integration test; the existing arena.test.ts assertion around match-end still only checks rating deltas, not `end.scoreboard`/`killingBlows`/`damageDone`/`healingDone`/`myTeam`/`honor`. `arena_end_view.test.ts` only covers the pure view core against hand-built fixtures. Given CLAUDE.md's rule that sim changes get tests, this is the one part of an otherwise well-tested PR with no direct coverage.

Merge-tree against release/v0.26.0 is clean, no conflicts.

ryan-foo:
All three should-fixes are in (through 1868bcfcb), plus the retarget: the branch now merges release/v0.27.0 and targets it. play.html gets the matching #arena-end-window element in the same spot as index.html, so /play players see the scoreboard. Pet damage and healing now credit the owner's row using the same ownerId resolution handleDeath already used for killing blows, with no new rng draws (the parity goldens stayed green unmodified). And the sim-side wiring has a real integration test now: tests/arena.test.ts drives a 1v1 to completion where a warlock imp does all the damage and lands the killing blow, asserting both personal arenaEnd events, the full scoreboard rows (killingBlows/damageDone/healingDone/rating), myTeam tagging, winner/loser honor, and that the pet's output lands in the owner's damageDone column. Arena suites 35/35, the neighboring 8 suites plus architecture green, tsc clean.

EnriqueGF (commented):
Reviewed the sim stats accumulation, the arenaEnd emit rework, the view core + painter, the HUD wiring, and the i18n additions. Solid work: the pure core is registered in `UI_PURE_CORES` with a pid tiebreak on the sort, the parity goldens moved only on the events hash (rng draws and digests unchanged, 384 and 284 draws as before), pet damage and killing blows resolve to the owner the same way in both hot paths, player names go through `esc()`, every visible string including the close aria-label is a `t()` key, and the `#arena-end-window` container is mirrored into `play.html`. I also confirmed the server forwards the whole `SimEvent` verbatim (server/game.ts events routing), so the extended payload needs no wire work, and no `Entity` field was added, so there is no `blankEntity` parity concern.

What I ran locally at the PR head: `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/arena.test.ts`, `tests/parity`, `tests/arena_end_view.test.ts`, `tests/i18n_completeness.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/mobile_window_coverage.test.ts`: all pass. One failure in `tests/snapshots.test.ts` ("active title wire") is environmental (no global `WebSocket` on Node 20 in my runner) and in a file this PR does not touch.

Findings:

- should-fix, `src/ui/hud.ts` (openArenaEndScreen, around line 15524): the modal opens without a focus trap. Every sibling modal in this HUD (delve board, rite panel, lockpick, quest dialog, player card, confirm) pairs open with `this.focusManager.open({ root })` and closes the trap on dismiss; this one only sets `display: block`, so keyboard and screen reader users are never moved into the dialog even though `markDialogRoot` gives it `role="dialog"`. Wire the same trap pattern and release it in `closeArenaEndScreen`.

- should-fix, branch state: the base is stale. Against the current `release/v0.27.0` tip, `git merge-tree` shows two real conflicts: `src/ui/hud.ts` (one import-block hunk, from the hud controller extraction that landed on the release branch after your last merge) and the generated `src/ui/i18n.resolved.generated/pending.ts` (mechanical regen churn; merge the base then re-run the i18n build rather than hand-resolving). Needs one more release merge before this is mergeable. Note the release-side refactor moved several hud responsibilities into controllers, so re-check where the `arenaEnd` case and `closeAll` live after the merge.

- nit, `src/ui/hud.ts`: nothing closes the scoreboard when the next match starts. If a player leaves it open and re-queues, the modal sits over the new bout until Escape or Leave. Closing it on the arena start event would be a one-liner.

Nothing blocking beyond the rebase. The sim side is clean: `bumpArenaMatchStat` is a no-op outside a match and draws no rng, stats live on the `ArenaMatch` so they tear down with the bout, and the arena test decisively pins the pet-credit path and the per-viewer team tagging.

ryan-foo:
All three follow-ups are in (through 55a1c6676) and the PR is mergeable again. Merged the current release/v0.27.0 tip; the two conflicts were exactly as you read them, the hud.ts import hunk from the controller extraction (kept only the arena end imports, the action bar ones now come from src/ui/hud/action_bar/) and the generated pending slice, regenerated via i18n:gen, and the arenaEnd case plus closeAll still live in hud.ts after the refactor. The end screen now opens the shared FocusManager trap with focusFirst on show and releases it in closeArenaEndScreen, matching the sibling modals, and the arenaStart event closes a scoreboard left open from the previous bout, so a re-queue never fights the modal. The focus suites pin only pure trap math, so both behaviors are pinned by source-scan tests in tests/arena_end_view.test.ts, per the house pattern. tsc, arena, focus, architecture, i18n, and parity suites green on the merged head.

EnriqueGF (approved):
Nice feature, and it lands the way this repo wants it: a DOM-free `buildArenaEndView` pure-core (registered in `UI_PURE_CORES`) plus a thin `arena_end_window` painter on the dialog-root seam, with the sim only gaining a per-bout stats map and extra fields on the existing `arenaEnd` event. No new wire command, no new `IWorld` member, no `hud.ts` banner section. Thanks for the thorough tests and the deliberate parity re-mint.

## What I reviewed and ran

- Read the full diff and the sim/ui/server/i18n CLAUDE.md rules for every area touched.
- Determinism: the new `bumpArenaMatchStat` draws no rng and is an O(1) map lookup; the `endArenaMatch` score-then-emit split preserves every existing `arenaEnd` field (`won`/`draw`/`oppName`/`ratingBefore`/`ratingAfter`/`allies`/`enemies`) and the A-then-B emit order, so the rng draw stream is untouched. The golden re-mint moves only the tick-125 `events` hash; `draws` (483) and `drawDigest` are unchanged, exactly the fingerprint of an event gaining fields. Confirmed with a fresh architecture-reviewer pass.
- Wire: `arenaEnd` forwards verbatim. `server/game.ts routeEvents` serializes the whole event object per session (the same `scoreboard` array reaches every participant), and `online.ts` pushes it straight onto the event queue (`ev as SimEvent`), no field allowlist. So offline and online render the same screen.
- Ran locally (all green): `tsc --noEmit`; `tests/arena_end_view.test.ts`, `tests/arena.test.ts`, `tests/architecture.test.ts`, `tests/parity`, `tests/i18n_completeness.test.ts`, `tests/localization_coverage.test.ts`. After `npm run i18n:gen` there is zero drift against the committed generated bundles, and M16 (the five non-Latin fills) is satisfied.
- i18n, no-dash/emoji, focus-trap/Escape/`closeAll` wiring, pet-to-owner credit for damage and killing blows: all check out. The pet-credit test is decisive.

Note: `tests/snapshots.test.ts` has one failure (`active title wire` -> `WebSocket.OPEN` undefined), but it is a pre-existing environment issue unrelated to this PR (it fails in isolation, the code path in `online.ts` is not touched here).

## Findings (all minor, none blocking)

1. Low: the honor number for Fiesta undercounts. `awardFiestaKillHonor` is granted per kill during the bout (`social/fiesta.ts:356`), before `endArenaMatch` samples `honorBefore`. So `honor = max(0, meta.honor - honorBefore)` captures only the completion/win honor, not the per-kill honor already banked. A Fiesta player who scored kills will see a smaller "+N Honor" than they actually earned this bout. Ranked (win honor only) and Yumi (no honor) are correct. If "completion honor only" is the intended meaning, fine, but worth a one-line comment; otherwise consider summing the bout's kill honor too.

2. Nit: the Yumi screenshot shows "+30 Honor", but Yumi awards no honor anywhere in the sim, so the real end screen renders no honor line (the `honorLine` is `honor > 0` only). The screenshot uses hardcoded mock HTML, so this is just a slightly misleading capture, not a code issue.

3. Nit: `ArenaEndRow.level` is populated in `endArenaMatch` but never rendered by the painter (it shows name + class only). Harmless dead field; drop it or use it.

4. Nit: when ranked, the board shows every row's Rating and Change, including the enemy team's, to each viewer. Not a leak (arena ratings are already public via the leaderboards), but classic end screens only showed your own team's rating movement. Purely a design call.

Approving: the change is fully mergeable, passes the gates I ran, and keeps every invariant in play. The items above are optional polish.

PR #1973, opened 2026-07-15
feat(sfx): give mage/priest/warlock wand auto-attacks their own cue

jamiecypher:
## Summary

Mage/priest/warlock wand auto-attacks previously reused the school's real spell-cast projectile sound (proj_<school>), making a passive auto-attack indistinguishable from an actual cast by ear. Adds dedicated wand_arcane/wand_holy/wand_shadow catalog keys (3 real takes each) and routes rangedSwing's wand branch through them instead.

## Type of change

- [x] Tests

## How was this tested?

- Commands: `npm run gate` (all 11 steps green), `npx vitest run tests/combat_sfx.test.ts tests/architecture.test.ts tests/parity`
- Manual steps: verified the 3 wand_<school> keys resolve 3 variants each in the built manifest; confirmed audio conforms to the -6dBFS true-peak / -14 LUFS standard via `npm run sfx:check`


FernandoX7 (commented):
Nicely scoped, and the determinism note in the third commit is exactly right: the parity goldens shift only in the `events` digest while every `state`/`nextId`/`rng` field stays byte-identical, so nothing about the rng draw order moved. The new `wand` flag rides the wire for free (events are `JSON.stringify`'d wholesale and the client re-queues the whole object, same as `attackAnimation`), and routing only the `rangedSwing` wand branch through `WAND_CUES` keeps real casts, procs, pets, and bosses on `proj_<school>`. Test coverage is good: both the positive per-school case and the defensive fallback for a school with no dedicated cue, plus the catalog-count pin bump.

One nit:

- `src/sim/types.ts:2550` `wand?: boolean` (nit): the emitter only ever sets `wand: true as const`, and the immediate neighbors use literal presentation-discriminator types (`attackAnimation?: 'ranged-shot'`, `attackAnimationStarted?: true`). `wand?: true` would match them and the value that is actually produced. Purely cosmetic.

Note: no CI checks are reporting on the PR head yet (`mergeStateStatus: UNSTABLE`, 0 check runs), so CI hasn't acted as the arbiter here. I ran `tests/combat_sfx.test.ts`, `tests/sfx_manifest.test.ts`, `tests/parity`, and `tests/architecture.test.ts` locally at the head ref and they pass.

PR #1974, opened 2026-07-15
feat(sfx): give Book of Deeds unlocks their own achievement chime

jamiecypher:
## Summary

Two related audio fixes bundled together:

- Swaps the ui_level_up placeholder (a baked synth tone) for the real recording, which had been staged but never actually copied into public/audio/sfx.
- Book of Deeds unlocks (the game's achievement system) previously reused audio.levelUp(), so the same sound fired for both real level-ups and every achievement, wearing thin fast. Adds a dedicated ui_achievement catalog key and routes handleDeedUnlocks through audio.achievement() instead.

Skin-event cosmetic unlock claims still deliberately use audio.levelUp() (a separate reward system from Book of Deeds).

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: `npm run gate` (all 11 steps green), `npx vitest run tests/game_audio.test.ts tests/deed_records.test.ts tests/deeds_window.test.ts tests/architecture.test.ts`
- Manual steps: confirmed both new/changed clips conform to the -6dBFS true-peak / -14 LUFS standard via `npm run sfx:check`; verified the deed-unlock source scan and pinned catalog count (156 keys, 27 ui_ cues) updated to match


FernandoX7 (commented):
Clean, well-scoped audio change. I verified the load-bearing claims against the head ref and they all hold up:

- The `audio.levelUp()` to `audio.achievement()` swap is confined to the deed-unlock path (`src/ui/hud.ts:10402`), a pure 1:1 substitution under the same `plan.playSound` guard. The other four `audio.levelUp()` calls are all correct to leave alone: real level-up (`hud.ts:9394`), post-cap virtual level-up (`hud.ts:9419`), and the two skin-event claims (`hud.ts:13207`, `13214`), which the summary calls out as a separate reward system.
- The regenerated artifacts are internally consistent, not hand-edited: the new `ui_achievement.mp3` (65871 bytes, sha256 `d473d495d1bb...`) and the replaced `ui_level_up.mp3` (88441 bytes, `b8dac64b9735...`) match their `runtime-pack.json` and `sfx_manifest.generated.ts` entries exactly, and `SFX_CATALOG_HASH` matches the runtime pack's `catalogHash` (`4b8d9131...`).
- `SFX_FIXED_CATALOG_KEYS` really is 156 keys with 27 `ui_` cues, so the updated pins in `tests/sfx_manifest.test.ts` line up with the generated file. The `{ key: 'ui_achievement', custom: true }` entry follows the existing custom-clip convention (`quest_*`, `lockpick_*`), and it needs no gain/speed-map override since it takes the defaults (gain 1, rate 1).

Merges cleanly onto `release/v0.27.0` with no conflicts. No i18n surface is touched: the deed banner text is unchanged and the cue keys are not player-visible strings. Tests cover the three moving parts (the `achievement` -> `ui_achievement` cue mapping, the `handleDeedUnlocks` source scan, and the catalog counts).

Nothing blocking. One operational note, not a finding: no CI checks are reporting on the branch yet, so make sure the gate actually runs (the loudness/conformance claim from `npm run sfx:check` is the one thing I did not re-run locally).

FernandoX7 (approved):

PR #1975, opened 2026-07-15
fix(chat): /1 General shortcut, working !lfg/!events, and stay on the last channel sent

ryan-foo:
## Summary

Hotfix for three chat issues. All three are client-side (plus the sim `/1` alias); the online server paths already handle them correctly once the client stops mangling the lines, so there is no server change.

### 1. `/1` is a shortcut for General chat

The sim chat router (`src/sim/social/chat.ts`) now routes `/1 message` to the General channel alongside `/g` / `/general`, and the client sticky map (`chat_channels.sentLineChannel`) recognizes it. `/1` is used rather than `/g` because the online server claims `/g` for guild, whereas `/1` always reaches General (offline and online). It updates the server's remembered channel to General like any General send.

### 2. `!lfg` and `!events` work again

Root cause: the v0.26.0 sticky-channel change made `composeChatLine` prepend a channel prefix to any line not starting with `/`, so `!lfg need a healer` was sent as `/say !lfg need a healer` and never reached the server's `text.startsWith('!')` relay gate. `composeChatLine` now passes a `!` community command through untouched (like an explicit slash command). Separately, `!events` (the natural plural of the `!event` command) now resolves via a plural fallback in `relayCommandById` when the exact word is unknown; a genuinely unknown word is still rejected.

### 3. The chat input stays on the last channel you sent in (including a whisper reply)

Root cause: the same v0.26.0 change made the client force its sticky *standing* channel onto every plain line, overriding the server's remembered-channel routing and, crucially, ignoring whisper. After replying to a whisper, the input snapped back to your previous channel instead of continuing the reply. The sticky is now a *target* (a channel OR whisper): a `/r` reply sticks to whisper so the next plain line keeps replying (`composeWhisperReply` -> `/r`), which the server routes to the same last-whisperer. An explicit one-off `/w Name` deliberately does not stick (its next reply would target whoever last whispered you, not that person). Standing channels behave exactly as before.

## Testing

- `tests/chat_channels.test.ts`: `composeChatLine` passes `!` commands through; `sentLineChannel` maps `/1` -> general (never `/g`); `sentLineTarget` sticks to whisper on `/r` but not on `/w`.
- `tests/discord_relay.test.ts`: `!events` -> event, `!lfg` unaffected, unknown word rejected.
- `tests/chat.test.ts`: the sim routes `/1` to the General channel.
- `tests/chat_hud_client_seam.test.ts`: a regression test that a party send followed by a `/r` reply keeps the input on whisper (`/r`), reproducing and fixing issue 3.
- All chat-logic suites, the S3 localization guard, and `tests/architecture.test.ts` green; `tsc --noEmit` clean for the changed files. (`chat_line.test.ts` fails on a missing `@testing-library/svelte` dev dependency in this fresh worktree, pre-existing and unrelated.)


FernandoX7 (commented):
Solid, well-scoped hotfix. All three roots trace back to the v0.26.0 sticky change prefixing non-slash lines, and each is diagnosed correctly. `/1` rides the shared sim router so it routes to General both offline and online, and generalizing `stickyChannel` to a `stickyTarget` (channel OR whisper) is a clean way to make a `/r` reply keep replying. Good regression test for issue 3.

I verified the parity that matters here: online, both `/1 msg` and `!lfg` / `!events` reach the server unmangled (`composeChatLine` now passes `/` and `!` through), `/1` falls through to `sim.chat()` in `routeRememberedChat` and updates `rememberedChat` to general like any general send, and `!events` resolves via the plural fallback inside `handleRelayCommand`. No wire or schema changes. Locally: the four suites pass (117 tests), the S3 guard (`tests/localization_fixes.test.ts`) and `tests/architecture.test.ts` are green, `tsc --noEmit` is clean, and it merges cleanly into `main`.

## should-fix

- A `!` community command resets the sticky channel to `say`. `noteSentChannel` runs on every send (`src/main.ts:1150`), and for a `!` line `sentLineTarget` falls through to `sentLineChannel`, which returns `'say'` for any non-slash line (`src/ui/chat_channels.ts:178`, `:201`). So if your standing channel is party or general and you fire `!lfg ...` mid-conversation, your next plain line silently drops to say. That is the same unexpected channel switch issue 3 is fixing, and it is inconsistent with how the module already treats transient commands: `/w`, emotes, and rolls all return null and leave the sticky untouched. Cleanest fix is to have `sentLineTarget` return null for a `!` line, plus a `sentLineTarget('!lfg ...')` -> null assertion next to the existing `composeChatLine` passthrough test.

## nits

- The plural fallback also strips a doubled tail (`!wtss` -> wts, `!helps` -> help), since it just drops a trailing `s`. Harmless for the current command set, just noting it is a suffix strip rather than a strict plural.
- This targets `main`, but the house default is to base a change off the latest release branch (release/v0.27.0 is the active integration base). Your call whether this hotfix ships via the release branch or straight to main.

ryan-foo:
Addressed the review: a ! community command now returns null from sentLineChannel/sentLineTarget, so firing !lfg or !events mid-conversation leaves the sticky channel untouched, same as whispers, emotes, and rolls (b545b76d3, with the sentLineTarget('!lfg ...') -> null assertion next to the passthrough test). Also retargeted the PR to release/v0.27.0 and merged the release base into the branch; the four chat suites, the S3 guard, architecture, and tsc are green on the merged head. The plural-fallback nit is left as is per the review (suffix strip, harmless for the current command set).

FernandoX7 (approved):

PR #1977, opened 2026-07-15
feat(trade): inspect offered items on hover in the trade window

ryan-foo:
## Summary

Adds a trade item preview: hovering (or focusing, or long-pressing on touch) any item in either side of the trade window now shows its full item tooltip, the same inspect players already get in bags, the vendor, and loot lists. Previously you could see a trade item's name and icon but had no way to inspect its stats before accepting.

## What changed

- **Both sides carry a `data-item` id.** The clickable "mine" rows already had one (for remove-from-trade); the read-only "theirs" rows now do too, and are made focusable with `tabindex="0"` so the inspect tooltip is keyboard- and long-press-reachable, not mouse-only.
- **Every offered item gets the shared tooltip.** After the trade window renders, each `.trade-item` is wired with `attachTooltip(() => itemTooltip(item))`, guarded on a resolvable `ITEMS[id]`. This is the exact pattern bags, the vendor, and loot rows use (`attachTooltip` handles desktop hover, keyboard focus, and mobile long-press for free).
- **Focus ring.** The `:focus-visible` outline is generalized from `.trade-item.mine` to `.trade-item` so the newly focusable "theirs" rows show a ring, satisfying the HUD-chrome WCAG contract.

No sim, server, wire, or `IWorld` change: it reuses the existing `tradeInfo` offer data and the existing item tooltip.

## Testing

- `tests/trade_item_preview.test.ts` (new): a source-scan guard (the `honor_ui.test.ts` style) pinning the wiring, since the trade window is inline DOM in `hud.ts` with no pure core to unit-test: both rows carry `data-item`, the "theirs" row is focusable, and every `.trade-item` gets the shared item tooltip.
- Green: `tests/css_corpus`, `tests/css_value_validity`, `tests/focus_visible_guard`, `tests/styles_extraction`, `tests/trade`, `tests/architecture`, `tests/client_shell`. `tsc --noEmit` clean for the changed files.

Note: the tooltip is the established item tooltip (visually identical to the bag/vendor/loot inspect), so no new visual style is introduced; a headless screenshot would require standing up a live two-player trade, which the reused pattern does not warrant.

### DIFF
--- a/src/styles/components.css
+++ b/src/styles/components.css
@@ -5509,7 +5509,7 @@
   .soc-x:focus-visible,
   .soc-name.soc-link:focus-visible,
   .soc-sugg-item:focus-visible,
-  .trade-item.mine:focus-visible {
+  .trade-item:focus-visible {
     outline: 2px solid var(--color-border-focus);
     outline-offset: 2px;
   }
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -15270,9 +15270,13 @@ export class Hud {
       const item = ITEMS[s.itemId];
       const label = `${item ? itemDisplayName(item) : s.itemId}${s.count > 1 ? ` x${formatNumber(s.count, { maximumFractionDigits: 0 })}` : ''}`;
       const inner = `${this.itemIcon(item)}<span>${esc(label)}</span>`;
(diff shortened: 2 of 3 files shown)


ryan-foo:
superseded by #2001 probably

PR #1978, opened 2026-07-15
fix(chat): /1 General shortcut, working !lfg/!events, and stay on the last channel sent (main)

ryan-foo:
## Summary

Hotfix for three chat issues. All three are client-side (plus the sim `/1` alias); the online server paths already handle them correctly once the client stops mangling the lines, so there is no server change.

### 1. `/1` is a shortcut for General chat

The sim chat router (`src/sim/social/chat.ts`) now routes `/1 message` to the General channel alongside `/g` / `/general`, and the client sticky map (`chat_channels.sentLineChannel`) recognizes it. `/1` is used rather than `/g` because the online server claims `/g` for guild, whereas `/1` always reaches General (offline and online). It updates the server's remembered channel to General like any General send.

### 2. `!lfg` and `!events` work again

Root cause: the v0.26.0 sticky-channel change made `composeChatLine` prepend a channel prefix to any line not starting with `/`, so `!lfg need a healer` was sent as `/say !lfg need a healer` and never reached the server's `text.startsWith('!')` relay gate. `composeChatLine` now passes a `!` community command through untouched (like an explicit slash command). Separately, `!events` (the natural plural of the `!event` command) now resolves via a plural fallback in `relayCommandById` when the exact word is unknown; a genuinely unknown word is still rejected.

### 3. The chat input stays on the last channel you sent in (including a whisper reply)

Root cause: the same v0.26.0 change made the client force its sticky *standing* channel onto every plain line, overriding the server's remembered-channel routing and, crucially, ignoring whisper. After replying to a whisper, the input snapped back to your previous channel instead of continuing the reply. The sticky is now a *target* (a channel OR whisper): a `/r` reply sticks to whisper so the next plain line keeps replying (`composeWhisperReply` -> `/r`), which the server routes to the same last-whisperer. An explicit one-off `/w Name` deliberately does not stick (its next reply would target whoever last whispered you, not that person). Standing channels behave exactly as before.

## Testing

- `tests/chat_channels.test.ts`: `composeChatLine` passes `!` commands through; `sentLineChannel` maps `/1` -> general (never `/g`); `sentLineTarget` sticks to whisper on `/r` but not on `/w`.
- `tests/discord_relay.test.ts`: `!events` -> event, `!lfg` unaffected, unknown word rejected.
- `tests/chat.test.ts`: the sim routes `/1` to the General channel.
- `tests/chat_hud_client_seam.test.ts`: a regression test that a party send followed by a `/r` reply keeps the input on whisper (`/r`), reproducing and fixing issue 3.
- All chat-logic suites, the S3 localization guard, and `tests/architecture.test.ts` green; `tsc --noEmit` clean for the changed files. (`chat_line.test.ts` fails on a missing `@testing-library/svelte` dev dependency in this fresh worktree, pre-existing and unrelated.)

---

Re-opened against `main`: this is the same content as #1975 (already merged into `release/v0.27.0`), re-based onto `main` so the hotfix ships there directly. It also includes the review's should-fix: a `!` community command returns null from `sentLineChannel`/`sentLineTarget`, so firing `!lfg` mid-conversation no longer resets the sticky channel to say (with the `sentLineTarget('!lfg ...') -> null` assertion next to the passthrough test). Cherry-picks apply cleanly on `main`, no merge conflicts; the four chat suites, the S3 guard, `tests/architecture.test.ts`, and `tsc --noEmit` are green on this head.

@FernandoX7 could you take a look and approve? Same change you merged in #1975, targeting main this time.

### DIFF
--- a/src/sim/discord_relay.ts
+++ b/src/sim/discord_relay.ts
@@ -75,7 +75,13 @@ export const RELAY_COMMANDS: readonly RelayCommand[] = [
 ];
 
 export function relayCommandById(id: string): RelayCommand | undefined {
-  return RELAY_COMMANDS.find((c) => c.id === id.toLowerCase());
+  const key = id.toLowerCase();
+  const exact = RELAY_COMMANDS.find((c) => c.id === key);
+  if (exact) return exact;
+  // Accept a natural plural ("!events" -> event, "!lfgs" -> lfg) only when the exact
+  // word is unknown, so "!event" written as "!events" still posts.
+  if (key.endsWith('s')) return RELAY_COMMANDS.find((c) => c.id === key.slice(0, -1));
+  return undefined;
 }
 
 /**
--- a/src/sim/social/chat.ts
+++ b/src/sim/social/chat.ts
@@ -761,9 +761,11 @@ export function chat(ctx: SimContext, text: string, pid?: number): SentChat | nu
     return { channel: 'party', message: clean };
   }
 
-  // "/g message": world-wide general channel (no pid = broadcast to all)
-  if (/^\/g(eneral)?\s/i.test(raw)) {
-    const clean = raw.replace(/^\/g(eneral)?\s+/i, '').trim();
+  // "/g message" / "/1 message": world-wide general channel (no pid = broadcast to
+  // all). "/1" is the classic numbered-channel shortcut for General; unlike "/g" it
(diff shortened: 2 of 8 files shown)


FernandoX7:
Already in release/v0.27.0 😉

PR #1934, opened 2026-07-14
feat(talents): forward-port Talents V2 and redesigned Warrior to v0.26

Rubsey:
## Summary

- Forward-ports the complete all-class Talents V2 choice-row system onto the current v0.26 architecture: exactly nine classes, six rows and 18 choices per class, with one canonical allocation path across Sim, server, ClientWorld, persistence, loadouts, and headless protocol.
- Layers the winning redesigned Warrior onto stable class ID `warrior`, including its final masteries, kit, Rage/stance rules, dual wield and Titan's Grip, Combat Mastery, Valor Roar, equipment handling, authored assets, renderer behavior, VFX, and final balance values.
- Ports the retained final row content and shared mechanics for the other eight classes, plus spec-commit UI behavior, online `wireRev` refresh, production save migration, action-bar repair, two-handed budget/proficiency fixes, Rogue denial, shield seating, dummy reset, localization, Guide output, and deterministic parity coverage.
- Deliberately excludes Classic Warrior, the alternate base-Warrior row set, Hardened Blood, old PTR/PBE/deployment work, and superseded Nythraxis content.

## Moving base note

This implementation was fully reconciled and verified on release commit `50a508e4fbef768046ba94e8dc2e2076f1b00858`. While final verification was running, `release/v0.26.0` advanced to `d8a871763d7ad1384d224a32b3c1ac0dd9cfc909` with the independent weapon-sheathing and Dungeon Finder work. This PR intentionally targets the live release branch, but that moving-base delta must be reconciled before merge; the known overlap is concentrated in renderer weapon state, Sim/world APIs, wire/command parity, and generated i18n artifacts.

## Verification

- Focused Talents V2, winning-Warrior, host-parity, migration, equipment, renderer, release-tier, and deterministic-parity suites passed.
- Every file that timed out only under the saturated six-worker local full gate passed in isolation: 12 files, 350 tests.
- Automated Chromium browser suite: 7 files, 59 tests passed. No manual visual-browser claim is made.
- TypeScript and Svelte typechecks passed with zero diagnostics.
- Environment, server, and production client builds passed.
- Localization generation is deterministic on this branch: 7,404 keys, zero pending retained rows; Guide output contains exactly nine class pages.
- Malware gate passed with zero high-severity findings; `npm audit --audit-level=moderate` reports zero vulnerabilities.
- Release metadata is synchronized at `0.26.0`; `git diff --check` and the Classic-Warrior residue scan are clean.

[...]

PR #1980, opened 2026-07-15
feat(talents): rotational identity rework for all 27 Talents 2.0 specs (+ mage/shaman PBE-2 fixes)

ryan-foo:
## Summary

Gives **every Talents 2.0 spec a real rotational identity.** PBE-2 feedback was that most specs are "light on beneficial talents", their choice-row talents are flat passives (a plain damage %, a stat, "grants an ability") that never change what you press. This reworks all 9 classes so each spec plays a build-and-payoff, proc-window, DoT/HoT-weave, or ramp loop instead, using only the resources the class already has (mana / energy / rage / focus), no new resource bars, and staying mobile-friendly (few buttons, windows in seconds not split-seconds, a visible tell).

Stacked on the Talents 2.0 V2 forward-port (#1934, base branch `feature/v026-talents2-warrior`), since the V2 choice-row tree only exists there. None of this has been through a PBE round yet.

## The engine

Most of it is data, not new code: a declarative proc engine (`src/sim/combat/talent_procs.ts`) where talents author `proc: { trigger, responses }`.
- **Triggers:** every Nth cast, on spell crit, on melee/ranged/pet hit, on a HoT expiring, on a shield consumed, on a big hit taken.
- **Responses:** make the next cast free/instant/cheaper, build stacks, add/vent charges, bank a rolling DoT, drop a delayed heal echo, direct heal/absorb, refund a cooldown, restore resource.
- Richer loops get small modules: `warspirit.ts` (melee builds toward an instant cast), `fulmination.ts` (charge / overload / chain), `channel_effects.ts` (ramping channel), plus the frost Icicles/Frostbite path.
- Every proc is spec-gated and deterministic (one RNG draw only after the spec/talent/ability/landed-hit gates), so the global draw order is unchanged for other builds.

## Spec identities (by class)

- **Mage:** Cryomancy (Icicles build to Icefall; Frostbite arms a boss-usable execute window), Pyromancy (rolling Afterflame), Aethermancy (mana-bank burst).
- **Shaman:** Thundercall (Fulmination charge -> AoE chain vent), Warspirit (Skyrend: melee builds an instant Arc Bolt), Spiritmend (mana-efficient Chain-Heal loop).
- **Paladin:** Requital (Verdict window -> empowered Crusader Strike -> free Rite -> refresh loop), Sacrament (heal-weave through Holy Shock), Vigil (Hallowed Wall active mitigation).
- **Priest:** Vespers (ramping Litany of Woe + Dirge spread/refresh), Doctrine (ward relay), Benison (direct/HoT prayer weave).
- **Hunter:** Packlord (pet cadence), Coldsight (precision setup), Fieldcraft (trap circuit).
- **Rogue:** Knifework (poison payoff), Thuggery (weapon cadence), Skulduggery (stealth window burst).
- **Warlock:** Hexcraft (DoT duration weave), Pactbound (demon handoff), Ruination (crit relay).
- **Druid:** Moongrove (lunar relay), Wildfang (active-mitigation tank), Groveheart (HoT weave).
- **Warrior:** already redesigned and healthy, so left intact; only a latent Ironguard armor-scaling bug was fixed.

## Also in here (mage/shaman PBE-2 feedback)

Flickerstep baseline, Icy Veins cooldown to 120, Third Current scales with max mana, Racing Mind auto-procs on crit, conjured food/water buffed, fire DoTs strengthened and surfaced in tooltips; shaman prerequisite dead-zones fixed (Rebounding Current -> L8, Rime Lock -> L14), the Springwell/Returning Current tier swap, Imbued Lifeblood and Returning Current scaled.

## On-screen tells

The action-bar proc glow and character-anchored VFX (orbiting Icicle shards, Frostbite flare, Thunder Ward crackle) are wired for frost and elemental, tier-independent for the actionable cue, cosmetic particles shed on low. Rolling that tell out to the rest of the new procs is a follow-up (in progress).

## Testing

Every class pass was verified individually: `validateTalentTree`, tooltip accuracy, `ip_scrub`, the S3 i18n guard, and `tsc` all green, with per-mechanic Vitest coverage and English + non-Latin i18n fills for new names. The only skipped gate stages are the SFX/renderer suites already red on the V2 base (this diff touches no audio files).

## Follow-ups (not blocking review of the mechanics)

- Cross-spec number tuning pass (reconciling proc rates / windows / magnitudes to consistent bands) — in progress.
- Proc glow / character-VFX tell rolled out to all new procs — in progress.
- A PBE round to tune with live data before any release.


ryan-foo:
Finished and green (independently re-verified: the Icicles/Fulmination/mage/shaman suites and `validateTalentTree` pass, tsc clean). Built by codex (gpt-5.6-sol) from a locked design brief, in 18 scoped commits, then reviewed.

**Design calls I made that are worth a look before this hits PBE:**
- **Icicles is Cryomancy mastery (baseline), not a choice.** It's the frost identity, and it costs one new button (Icefall). If you'd rather it be a choice option, easy to demote.
- **Racing Mind procs on crit.** Fixes the flow break, but frost has no reliable crit outside shatter, so it reads more fire-native. It competes with Slow Burn in the same row anyway.
- **Frigid Reversal** is the new level-17 survival option that fills Flickerstep's vacated slot; happy to swap it for something else.

**What to hammer on PBE:**
- Icicles pacing: does 12/stack (x2 shattered) feel worth weaving Icefall, or does it overcap and get ignored?
- Fulmination: 8 Nature/charge at a 9 cap, greed-vs-dump, does the ceiling feel right?
- The shaman tier reshuffle: confirm every option is now live at its tier and the level 8 / 11 spreads read well.

**Heads-up:** the shaman tier moves sanitize existing saved allocations into free repicks (no loadout migration). Fine for a PBE wipe, but flagging it. All numbers are tuning targets.

Stacked on #1934, so it needs that as the merge base (or a rebase onto release once #1934 lands). The only gate red is the pre-existing SFX/renderer failures on the V2 base, no audio or renderer files in this diff.

jgyy (commented):
Well-executed content/mechanics work. Both new mechanics (Icicles/Icefall, Fulmination) live in their own module called from the existing effect-dispatch seam, never touching sim.ts directly; all randomness goes through `ctx.rng`, verified no `Math.random`; and the "no RNG draws" claim for Icicles is decisively tested by monkeypatching rng calls and asserting zero. Test coverage is strong: seeded deterministic replay, exact chance-draw-count assertions, and the RL obs-vector-size bump is handled correctly end to end (python env queries `obs_size` rather than hardcoding it, tests updated for the new constant).

should-fix: the base branch, `feature/v026-talents2-warrior`, looks stale relative to where upstream actually is. It isn't an ancestor of release/v0.26.0, and upstream already has release/v0.27.0 plus a separate, newer `feature/v027-talents2-warrior` that diverged from the same root. This strongly suggests the v026 talents-V2 line was superseded by v027 while this PR (and its parent #1934) kept building on the old one. Recommend confirming the right target branch with the maintainer before merge so this doesn't land on a dead-end line; no textual conflicts against its current stated base, so this is a targeting question, not a code problem.

nit: the docs/prd note attached to this PR flags the new balance numbers (Fulmination 5%/charge, Icefall 8/20 flat, etc.) as PBE targets, not release-locked values, which is honest and consistent with "don't invent balance numbers," just worth remembering these aren't final.

nit: the RL observation-vector size change (16 -> 18 self-block) is a breaking change for any externally-trained policy pinned to the old width. Expected for a PBE-tier surface change, but worth a line in the PR description since it currently isn't called out.

No blocking issues in the sim logic itself.

ryan-foo:
Both follow-ups are in the description now. The RL note: the self-block grows 16 to 18 so obs_size changes, which is breaking for any externally-trained policy pinned to the old width (in-repo consumers query obs_size at reset, so they are unaffected); called out explicitly. On targeting: acknowledged the v026 talents2 line is superseded in flight; the maintainer's integration target for this work is release/v0.27.0, so this PR stays stacked on its parent until the Talents 2.0 foundation lands on the v0.27 line and then retargets together with it rather than rebasing across the dead-end branch. Balance numbers remain PBE tuning targets.

ryan-foo:
### Update: three distinct caster identities + Frostbite + shadow priest

Since the original post, the mechanics were re-cut so the specs no longer feel the same (all still on the V2 stack, `validateTalentTree` green, contribution suites + parity + tsc clean; the only gate red is the pre-existing SFX/renderer failures on the base, no audio/renderer files in this diff):

**Frost (Cryomancy) = build, freeze, execute.**
- Icicles build on Rimelance; **Icefall** is a single-target execute that only does its big (2.5x) damage on a truly frozen target (root/stun, not the passive chill).
- **Frostbite** (Fingers-of-Frost analog): a Rimelance hit has a 15% chance (one gated draw) to arm the next Icefall as an execute, so the loop works even on root-immune bosses. Icebind stays the manual window. Buff is in obs + the HUD icon.

**Elemental = ride the storm, vent the chain.**
- Each Thunder Ward charge ramps Arc Bolt's **Overload** chance (~20% at a fresh ward to ~45% at the 9 cap); Overload fires a free half-bolt that chains to a nearby enemy (reuses the merged `chainDamage` path).
- **Earthen Jolt** vents the whole ward as AoE and resets the ramp: ride vs cash-out.

**Shadow priest = DoT weaver riding a ramping channel.**
- Litany of Woe: range 20 to 30, ticks now ramp (12, 16, 19) baseline.
- Talents: **Ninefold Litany** (channel to 9s, L17), **Woe's Crescendo** (a *completed* channel bursts AoE, L17), **Deathless Dirge** (Mind Blast refreshes Dirge, L14), **Plague Chorus** (Mind Blast spreads Dirge to nearby, L14).

Three different textures on one pillar (build-up + a payoff you time, no fast weaving, mobile-safe): frost times a window, elemental manages a crescendo, shadow commits to a channel.

**Still to come (separate passes):** the character-glow / action-button proc tell so availability lights up on your model (client/render work, flagged earlier). The plain ability/item number tweaks (Icy Veins CD, conjured food/water, fire DoT) can be split into a non-2.0 fix if you want them to merge without waiting on #1934.

EnriqueGF (commented):
Reviewed the full diff against `feature/v026-talents2-warrior` and ran the gates locally: `tsc --noEmit` clean, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `i18n:gen`, no regen drift in the committed artifacts), the full `tests/parity` suite (164 passed, casting-lifecycle golden validates), and every touched suite: `icicles`, `fulmination`, `channel_effects`, `mage_pbe_feedback`, `shaman_pbe_feedback`, `talents`, `env_protocol`, `spec_masteries`, `snapshots`, `guide`, plus the tooltip/icon/auras suites. Biome on the changed files: warnings only, no errors or format diffs.

The determinism work here is careful and it shows: `fixedNoCrit` skips both the damage and crit draws (and the sure-crit consumption, pinned in `tests/icicles.test.ts`), the Frostbite roll is the single gated draw behind procsFor plus the ability filter, `spreadOwnedDot` sorts targets by id, and `nearestChainTarget` tiebreaks equal distances on id. `channelTicksFired` being optional and armed only for ramp/finisher channels keeps every other channel byte-identical, and its cast/cancel/complete lifecycle is clean. Wire parity needs no change: aura `kind` and `charges` already ride, and the `stacks > 1` elision on the wire is covered by the `?? 1` fallback in `auras_view.ts`, so a 1-stack Icicles badge renders correctly online.

Findings, none blocking:

- should-fix (or confirm intended): Icefall's spell power rider is added once to the whole volley, not per icicle. In `src/sim/combat/effect_dispatch.ts` the flow is `dmg = eff.min * stackMultiplier` then `dmg += directHitBonus(...)`, and the rider then also passes through `vsFrozenMult`, so with SP > 0 the per-icicle damage is neither 8 nor 20 and the "fixed damage" / "$d Frost damage each" tooltip drifts from reality. Every assertion in `tests/icicles.test.ts` sets `spellPower = 0`, so the SP-scaled path is unpinned. If the rider is intentional scaling, a test with nonzero SP plus a tooltip tweak would close the gap; if not, `fixedNoCrit` could skip the rider too.
- nit: the PR body's RL follow-up says the self obs block grows 16 to 18, but the code and pins say 19 (`SELF_OBS_SIZE = 19` in `src/sim/obs.ts`, pinned in `tests/env_protocol.test.ts`): three fields were added (icicles, ward charges, frostbite). Worth correcting in the body since that note exists specifically for external policy owners re-padding to the new width.
- note: `tests/snapshots.test.ts` "key lockstep" fails in my environment (no `WebSocket` global under this Node), but it fails identically on the base branch, so it is environmental and not a regression from this change. The Frostbite wire round-trip test this PR adds passes.
- note: `presence_of_mind` and `desperate_prayer` defs remain in the talent ability tables with no grant left pointing at them. Harmless dead content for PBE, just flagging so it does not surprise a later cleanup.

Also verified: no recursion path for Fulmination Overload (the replay passes `abilityId: null`, so `onSpellHit`/`onSpellCrit` cannot retrigger it), Racing Mind's ICD uses the existing `procState.icds` decrement, Ninefold Litany's `channelDurationPct` resolves through `res.def` so the finisher's final-tick check and the tick cadence agree, `channelDurationPct` was added to both `zeroAbilityMod` and the fiesta merge, and blink lands at learnLevel 5 as described. The moved shaman option ids sanitize stale allocations into free repicks as the PR notes.

Good PR. The two mechanics land as small modules behind the existing seams with genuinely decisive tests.

ryan-foo:
@EnriqueGF fixed per your should-fix in 4a3b68ac4: Icefall is fixed damage by design, so the fixedNoCrit path now skips the spell power rider the same way it already skips the damage and crit draws; per-icicle damage is exactly 8, or 20 shattered, on any gear, and the tooltip stays honest. directHitBonus is a pure computation, so the gate changes no rng draw order, and the parity suite passes with zero golden drift (no golden exercises Icefall). Added a pin at 150 Spell Power asserting the 3x8 and 3x20 volleys equal the 0-SP amounts, plus a proof the skipped rider would have been nonzero so the test is decisive. The body's RL note is corrected to 16 to 19 (SELF_OBS_SIZE = 19: icicles, ward charges, frostbite), and the dead presence_of_mind/desperate_prayer defs stay for PBE as you guessed.

ryan-foo:
### Update: Warspirit + Requital melee-DPS identity pass (PBE "light on talents" feedback)

Both classes use one class-wide choice-row tree (spec is stored separately), so these gate per-spec rather than forking rows; the enhancement tank kit is untouched. Mana stays the resource, no new resource system. `validateTalentTree` green, 318 contribution tests + parity pass, ip-scrub clean.

**Warspirit (enhancement):**
- **Skyrend:** landed melee + Ancestral Strike build up to 5 stacks; Arc Bolt consumes them for +20% cast speed and +10% damage per stack, instant at 5. Melee that erupts into a cast, the spec's identity.
- **Tempest Reprise** (replaces the flat +20% Earthen Fury): Ancestral Strikes have a 20% chance to reset, feeding the Skyrend build-up faster.

**Requital (retribution):**
- **Blood Debt:** landed attacks / Crusader Strike have a 20% chance to reset Rite of Expulsion and make the next one free for 8s (a reactive instant).
- **Oath's Due** (replaces the flat Swift Verdicts): Verdict opens a 7s window that empowers your next Crusader Strike by 50%; a whiff keeps the window.

Deferred: the Skyrend/Blood Debt state shows in the HUD but isn't in `encodeObs` yet (agent-legibility follow-up), and the enhancement-tank branch wants a post-merge integration run.

ryan-foo:
### Requital loop follow-up

Interlocked Blood Debt and Oath's Due so Requital is a rotation, not two parallel procs: a windowed (Oath's Due) Crusader Strike raises Blood Debt's proc from 20% to 70%, and only the resulting free Rite of Expulsion refreshes the Oath's Due window (reset to 7s, no stacking). So it chains Verdict -> empowered Crusader Strike -> free Rite -> refresh, sustained by hitting windows. Gated against degeneracy: the 30% miss consumes the window without a free Rite and drops you back to a Verdict opener; a normally-cast Rite never refreshes. A success-then-failure test pins that it can't self-perpetuate. Mana-only, no resource added. Contribution suite + parity + validateTalentTree green.

EnriqueGF (commented):
Senior review pass. Thanks for the detailed write-up and the deferred/follow-up notes, they made this much easier to navigate.

## What I reviewed and ran

Checked against the merge-base of the PR branch and its stated base (`feature/v026-talents2-warrior`, at `da66477c`). I focused on the repo invariants: sim determinism/purity, tri-host + IWorld parity, the RL obs surface, i18n, and the new caster mechanics.

Ran locally (all green):
- `tsc --noEmit` (exit 0)
- Guards: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`
- Parity goldens: `tests/parity/**` + `tests/combat_casting_lifecycle.test.ts`
- Feature suites: `icicles`, `fulmination`, `mage_pbe_feedback`, `shaman_pbe_feedback`, `warspirit`, `channel_effects`, `spec_masteries`, `talents`, `env_protocol`, `signature_mechanics_v2`, `talent_tooltip_accuracy`

I also ran a determinism/seam pass and a tri-host parity pass with the domain reviewers.

## Determinism: clean

- Every new RNG draw is gated behind the talent/proc check before it draws, so non-talented actors never touch the stream and draw order is preserved:
  - `fulmination.ts` gates on `hasFulmination` + ward `charges > 0` before the single `rng.chance`.
  - `talent_procs.ts` `chanceAura` (Frostbite) and `meleeHit` sit behind `procsFor` (talent-filtered); the `meleeHit` case is net-new so threading `abilityId` through `meleeSwing` shifts no pre-existing draw.
  - `effect_dispatch.ts` `fixedNoCrit` (Icefall) skips both the damage-roll and the crit draw; since `fixedNoCrit` only exists on new content, existing abilities keep their exact draw sequence.
- `warspirit.ts`, `channel_effects.ts`, `dot_mutation.ts` draw zero RNG; `rampedDrainTickDamage` is a pure scale of an already-drawn value and is identity when `rampPct` is undefined (existing channels unaffected).
- The three regenerated goldens (`c4a`, `c4b`, `nythraxis_full_pull`) change only in `state`/`damageDealt`/`events` hashes; the `draws`/`seq` lines are byte-identical. That is the decisive signal that the regeneration reflects new deterministic values (e.g. the new `armorFromStrPct` in `recalcPlayerStats`), not a shifted draw order.
- No `Math.random`/`Date.now`/`performance.now` in sim (only purity comments); `tests/architecture.test.ts` green.

## Parity / RL obs: clean

`SELF_OBS_SIZE` 16 -> 19 is appended at the end of the self block and `obsSize()` is driven off the constant; `headless/env_server.ts` and `python/wow_env.py` read the size dynamically at reset, so in-repo consumers are safe (as you noted). New aura state (icicles stacks, lightning-shield charges, frostbite) rides the existing generic `wireAura` encoder/decoder, and no new `SimEvent` variant was introduced (`procSurge`/`wardBloom`/`detonate`/`nova` are pre-existing `fx` values). No IWorld facet changed, parity pin untouched and green.

## Findings

**[Blocking for merge, but acknowledged] Base branch / scope.** As opened against `feature/v026-talents2-warrior` (at `da66477c`), the diff is not the caster-only change the summary describes: it is 80 commits / +15.8k lines covering the full Talents 2.0 rework across all classes (druid/mage/warlock/rogue/hunter/paladin/priest/shaman), because the base branch predates the V2 foundation. Against the intended target `release/v0.27.0` the two lines have diverged hard (release is ~264 commits ahead and lacks this work), so this cannot merge there as-is and cannot really be reviewed in isolation. You already flag the stack + retarget plan, so this is a "hold until the V2 foundation lands on v0.27, then retarget with the parent" item, not a code defect. Calling it out so it is not lost.

**[Should-fix, convention] i18n overlays edited directly.** The change edits ~17 `src/ui/i18n.locales/<lang>.ts` overlays (fireball/pyroblast/mind_flay reword + `{overTime}` token, wyvern_sting/hemorrhage renames, plus the new `icefall` fills). The stated workflow is contributor-adds-English-only; overlays are the maintainer's to fill at release, and pre-filling them can bypass the reword-staleness detection so the maintainer never re-reviews the quality. The good news: token parity is intact ({overTime}/$o is a real runtime token from `hud.ts`, and mind_flay correctly drops `{damage}`), the M16 non-Latin fills for `icefall` are present, and `i18n_completeness`/`localization_fixes` are green. So this is a convention call for you to confirm, not a broken build.

**[Note] Icefall + next_attack_crit.** `fixedNoCrit` short-circuits before `consumeNextAttackCrit`, so an Icefall cast will not consume a pending next-attack-crit charge. Draws no RNG, only affects new content; just confirm leaving the charge intact on a fixed-damage cast is intended.

**[Note, verified not a bug] Fulmination overload scaling.** The overload/chain replay calls `dealDamage` with `damageModifiersResolved` unset (false), unlike the DoT-detonate replay which passes true. I checked the math: the overload amount is `0.5 * finalDamage`, where `finalDamage` is the pre-`dealDamage`-internal-mods input, so the overload lands at exactly 50% of the primary's resolved hit (target vuln applied once, same as the primary). No double-dip. Flagging only so the choice is documented; if you ever want the overload to be a clean pre-scaled replay instead of "50% of a fresh hit", that is where the flag would go.

## Verdict

Code quality is high and the determinism/parity/i18n invariants hold with the suite green. The only real blocker is the base/scope + retarget situation, which you have already acknowledged. Leaving this as a comment rather than request-changes on that basis; happy to re-review the caster-scoped slice once it is retargeted onto the v0.27 line.

jgyy:
Thanks for the two deep passes on this, @EnriqueGF, the RNG-draw-order and wire-parity verification in particular saved a lot of re-checking on my end.

Confirmed the Icefall fix landed (4a3b68ac4) and matches your should-fix exactly: fixed damage stays fixed regardless of Spell Power, pinned at 150 SP.

On the two open items:
- Base branch / scope: agreed, holding until the V2 foundation lands on release/v0.27.0 and this retargets with its parent (#1934). Not a code defect, just sequencing. Will track it there.
- i18n overlays edited directly: good catch, that's against the contributor-English-only convention even though the content and token parity check out. Since it's a PBE-tier line that isn't landing on a release branch yet, I'm fine leaving the overlay edits in place rather than asking for a revert-and-redo, but flagging for @ryan-foo: please don't pre-fill non-English overlays on future PRs, that step is reserved for release time so reword-staleness detection doesn't get bypassed.

Nice work on the mechanics themselves, the module-behind-the-existing-seam pattern held up across all nine classes. Leaving this open pending the retarget.

EnriqueGF (changes requested):
Reviewed the mechanics and ran the gates locally. The design here is genuinely good: the declarative `proc: { trigger, responses }` engine in `src/sim/combat/talent_procs.ts` is the right shape for this, it keeps the bulk of 27 spec reworks as data rather than code, and the richer loops (`warspirit.ts`, `fulmination.ts`, `channel_effects.ts`) correctly landed as sibling modules behind `SimContext` instead of growing `sim.ts`.

I specifically verified the central determinism claim ("one RNG draw only after the spec/talent/ability/landed-hit gates, so the global draw order is unchanged for other builds") and it holds. Every draw site I traced sits behind its gates: `chanceAura` after `procsFor` plus the ability gate, and `onMeleeSwing` after the `abilities.includes` / `positiveDamage` gates. The parity traces back this up empirically, see below.

Two red tests block merge, and one i18n gap. All three are mechanical, none are design problems.

## Blocking

**1. `tests/parity/golden/fiesta_powerups.json` was not regenerated (parity gate red)**

Reproducible on this branch, clean on the base commit (`da66477c2`). I ran the scenario isolated twice on each:

- base: `Tests 2 passed`
- this branch: `× matches the committed golden`

The good news is the divergence is NOT an rng fork. Comparing the trace fields, `draws` and `drawDigest` are byte-identical to the golden. Only `state`, `events`, `hp` and `maxHp` moved, e.g. a level 20 fiesta fighter goes `maxHp: 418` to `387`. That is legitimate fallout of the talent tree rework changing the bot's default build, not a determinism regression, so your draw-order claim survives.

It just means the golden is stale. `nythraxis_full_pull.json` was regenerated in this PR but `fiesta_powerups.json` was missed. Per the harness comment in `tests/parity/parity.test.ts`, regenerate deliberately as its own commit:

```
UPDATE_PARITY=1 npx vitest run tests/parity
```

Worth eyeballing the resulting `maxHp` delta to confirm the fiesta bot build change is intended, since the PR body does not mention fiesta.

**2. `tests/nameplate_view.test.ts` fails on a stale source pin**

Also clean on base (25 passed), red here. `renderer.ts` hoisted the tier lookup:

```
4282:  const staticEffectsTier = coerceFxTier(document.documentElement.dataset.fxLevel);
4283:  this.vfx.setStaticEffectsTier(staticEffectsTier);
5116:  const nameplateInterval = nameplateIntervalSec(staticEffectsTier);
```

I checked the behavior and it is correct and fairness-compliant: it still reads the static preset (`dataset.fxLevel`) and never the FPS governor, and hoisting it to share with `setStaticEffectsTier` is an improvement. But the guard is a source regex expecting the inline `nameplateIntervalSec(coerceFxTier(document.documentElement.dataset.fxLevel))` form, so it no longer matches.

Please update the pin rather than un-hoisting, and keep its teeth by asserting both halves: that `staticEffectsTier` is bound from `coerceFxTier(document.documentElement.dataset.fxLevel)`, and that `nameplateIntervalSec` is called with `staticEffectsTier`. As written a one-line pin change would let a future edit pass the governor tier in without reddening anything, which is exactly what this guard exists to stop.

## Medium

**3. Reword staleness in the 15 Latin locale overlays**

Two abilities were renamed and had their mechanics rewritten in `src/ui/i18n.catalog/abilities.ts`:

- `wyvern_sting`: "Wyvern Sting" (incapacitate, any damage breaks, Survival) to "Briar Trap" (root within 5 yd, Fieldcraft)
- `hemorrhage`: "Red Ribbon" (bleed over 12 sec, +40% bleed taken, Subtlety) to "Maskfall" (plain strike, 1 combo point, Skulduggery)

The 5 non-Latin overlays were updated correctly. The 15 Latin ones were not, so `de_DE.ts` still carries `'Stich des Flügeldrachen'` describing an incapacitate that no longer exists, and `fr_FR.ts` still has `Piqûre de wyverne` and `Hémorragie`. Because those keys already hold values they are not counted pending and nothing reds, which is the reword-staleness trap.

This is in scope by your own precedent: the PR already chased the same class of staleness into all 15 Latin overlays for `fireball`, `pyroblast` and `mind_flay`. Same treatment here, then `npm run i18n:gen`. Minor third instance: `blink.description` went `(Mage talent)` to `(Mage)` and the Latin overlays still say `(Magiertalent)`.

## What I ran, all green

- `npx tsc --noEmit`, exit 0
- `tests/architecture.test.ts`, 24 passed (sim purity and banned-API guards clean)
- `tests/localization_fixes.test.ts`, 30 passed (note it needs `npm run i18n:gen` first)
- `tests/world_api_parity.test.ts`, `talent_tooltip_accuracy`, `spec_masteries`, `action_bar_view`, `auras_view`, `deeds_content`, `guide`: 428 passed
- the 10 new mechanic suites (icicles, fulmination, warspirit, channel_effects, pyromancy, requital, vigil, spiritmend, benison, doctrine): 86 passed
- `tests/parity/coverage.test.ts`: 47/47 passed in isolation
- `npm run i18n:gen` reproduces the committed bundles byte-clean, so the generated artifacts are honestly generated

I also confirmed the other 24 failures in the full suite (sfx studio, mail, vale cup client, clock format, quest link wire, snapshots) are pre-existing on the base commit and not yours.

One process note: `tests/parity` is flaky under an unbounded run. Across three full-suite runs I saw 6, then 5, then 2 failures, and `drowned_litany` / `c4a_casting_lifecycle` / `nythraxis` all pass in isolation. Only `fiesta_powerups` reproduces every time. Matches the `npm run gate` warning about core contention, so do not chase the others.

Nit, no action needed beyond a habit: the PR description uses an em dash and a robot emoji, which the repo's no-em-dash / no-emoji rule explicitly extends to PR text.

Happy to re-review once the two goldens/pins are sorted. The engine itself looks solid to me.

Rubsey (commented):
Fourth pass, after jgyy's and EnriqueGF's reviews. I focused on what earlier passes had not deep-traced: the hunter/rogue/warlock/druid and late paladin/priest/shaman mechanics, the action-bar and aura cue layer, render fairness, and progression safety for the moved options. Ran everything at this head in a fresh worktree: tsc clean, all 25 new mechanic suites green (198 tests), and the three re-minted parity goldens change only state/hp/events lines, the draws and seq lines are byte-identical, which independently confirms the no-draw-order-shift claim. The engine design holds up under this scrutiny: sim.ts grew four lines, the modules sit behind SimContext, DoT ticks pass abilityId null so they cannot retrigger spell-hit procs, and the damageModifiersResolved gate is applied consistently to every modifier block in dealDamage.

Status of the outstanding request-changes items: all three still reproduce at this head. The fiesta_powerups parity golden is still stale (1 failed), tests/nameplate_view.test.ts:341 still fails on the old source pin, and the Latin overlays for wyvern_sting/hemorrhage/blink are still stale (de_DE.ts:3503, fr_FR.ts:3882). Same fixes as already described. CI also has no runs on this head, so a fresh approved run is needed once those land.

New findings, none raised before. All should-fix, nothing here blocks beyond what is already requested:

- should-fix: one landed pet swing counts twice for petHitNth procs against a damage-share victim. The petDmgSharePct redirect (src/sim/combat/damage.ts:400) recursively calls dealDamage for the pet's share and passes `direct` through, and dispatchLandedHunterTalentHit has no alreadyFinal gate, so the nested call fires onPetHit and then the outer call fires it again. In PvP against a Bloodbond hunter or Pactbound warlock, Fiendlore's every-2nd cadence arms every swing and Packbond's every-3rd fires about every 1.5. Gating the dispatch on !alreadyFinal (or passing direct false into the redirect) fixes it.

- should-fix: Dawn's Reply (pal_r5_crusaders_zeal, src/sim/content/choice_rows_classic.ts:458) reintroduces the dead-at-unlock defect this PR fixes for Shaman. The row unlocks at 5, but its only payoff triggers on holy_shock (learn level 10) and empowers flash_of_light (learn level 12), and the spec gate makes it permanently inert for non-Sacrament paladins. The replaced base option was live at 5. The live-at-unlock rule and its guard test (tests/shaman_pbe_feedback.test.ts) are Shaman-scoped, which is exactly how this slipped through; worth generalizing the scan while you are in there.

- should-fix: Creeping Rot (the Hexcraft mastery) and Deepened Hex share one extendedBy budget per DoT application (src/sim/combat/dot_mutation.ts:16). One Gloom Bolt extension (+3, maxBonus 6) leaves the mastery's min(1, 3-3) at zero, so picking the level-14 row option silently turns the spec mastery off for that application, and the combined ceiling is order-dependent. If the shared anti-stack ledger is intended it needs a comment and a combined-pick test (hexcraft.test.ts and hexcraft_talents.test.ts each test one talent in isolation); otherwise track the budget per effect source.

- should-fix: ACTIONABLE_AURA_IDS (src/ui/auras_view.ts:85) omits 14 of the proc auras this same diff authors: all the rogue/hunter/warlock windows (rog_redhanded, rog_false_face, hun_iron_aim, wlk_pain_communion, and so on) and the mag_icicles / sha_skyrend stack meters, while mechanically identical siblings (pal_kindled_faith, mag_aetheric_flux) are protected. On the low preset with more than 8 buffs those windows and meters can still be culled, and the buff-bar badge is the only place the stack count renders. This contradicts the set's own "add each empowerment in the commit that authors it" comment. Strictly better than base (which shed everything), hence not blocking, but the list should be completed before this ships a tell players act on.

- should-fix: the action-bar cue tables miss three of this PR's own procs. hun_venom_relay is in the free-cost map so Fell Shot prices free, but fellShotProcState (src/ui/action_bar_view.ts:396) only arms the glow for hun_packbond / hun_menders_signal, so the window never glows while the identical Packbond window does. And mag_ember_relay / mag_aetherwell_relay appear in neither cost map, so during their windows the slot still greys out below full cost even though the sim gate would accept the cast.

- should-fix: the new aura kind aetheric_flux has no aura_aetheric_flux icon recipe in src/ui/icons.ts, while its three sibling new kinds (icicles, frostbite, stormcharge) all got one plus test pins. The Aetheric Flux meter renders through the keyword fallback with a dev console warn.

- should-fix: docs/design/graphics-settings-fairness.md:61 still quotes the exact shed predicate this diff replaces (auras_painter.ts:174 now includes !s.isActionable) and still states that hiding a buff icon removes no actionable information. The code change is fairness-positive, but that doc is the reference reviewers key on, so it should be updated in the same change.

Nits:

- The four clamped PvP terminal paths dispatch pet/ranged procs for the killing blow but not onSpellHit/onSpellCrit (src/sim/combat/damage.ts:453 vs :756), so a match-ending Gloom Bolt arms nothing while a match-ending pet hit counts. If post-match windows are deliberately moot, a one-line comment at the spell dispatch site would stop a future "fix".
- Offhand auto swings pass no abilityId (src/sim/combat/auto_attack.ts:204), so they count as auto_attack for meleeHit triggers (Skyrend builds at roughly double rate when dual wielding) but can never consume the next_ability_damage window the same engine arms, so Scrapper's Edge silently waits for a mainhand hit despite saying "next landed melee auto-attack". Passing abilityId at the offhand call site would make the two paths agree.
- chanceAura responses apply their aura with response.id, not def.id (src/sim/combat/talent_procs.ts:241), so the respec sweep in src/sim/progression/talents.ts:91 misses it: a Frostbite self-aura survives a spec swap for up to 15s. Cosmetic today since its only consumer is unlearned by the same swap.
- The new fx ids procSurge / wardBloom / detonate have no renderer arm and all land on the generic nova fallback (src/render/renderer.ts:2971). Fine if intended, worth knowing given the tell rollout follow-up.
- Six option ids are replaced, not kept (mag_r17_blink, pri_r14_mind_melt, pri_r14_pain_and_suffering, pri_r17_desperate_prayer, pri_r17_inner_fire, sha_r11_elemental_attunement). I executed the sanitize path against stale allocations holding all six: clean silent drop into free repicks, no crash. So this is safe, but the PR-level "every replaced option keeps its stable id" framing is overbroad and the drop is silent to the player.
- Briar Trap keeps wyvern_sting's membership in hunterRangedShotAbilityIds (choice_rows_classic.ts:47), so placing the trap now advances "every 3rd ranged shot" counters and can consume next-shot discounts. The list membership pre-exists; the retype to a position-targeted trap is what makes it wrong.
- talent_i18n.ts:8040 renders the refreshDot notation with a U+21BB arrow glyph standing in for "refreshes". Every neighboring shorthand is ASCII, and a glyph standing in for a real label wants real text.
- Ninefold Litany lands at roughly 5x Litany damage per cast and per mana (channelDurationPct 2 compounding with the new rampPct 0.3) vs the 1.25x to 1.5x band of its row siblings. Flagging for the acknowledged tuning pass.

Also verified clean, so you do not have to re-check: the proc glow CSS reaches play.html (both entries load main.ts, which imports the layered styles), the glow itself is tier-independent with reduced-motion keeping the static armed state, characterAuraParticleScale reads only the static tier with governor-independence pinned, the two new optional Entity fields follow the existing procState convention (no blankEntity change needed), the aura badge's stacks ?? 1 correctly covers the wire's stacks-at-1 elision, tree validity holds (144 options base and head, all rows at 3), conjured food/water values equal the existing zone3 vendor tier rather than invented numbers, the Requital 70% loop is bounded (each free Rite needs a fresh landed-hit roll), and the i18n artifacts are honestly generated (sha256 one line, regen byte-clean, S3 guard green).

Good work overall: the declarative proc engine is the right shape, the determinism discipline is real, and the test volume is genuinely decisive in most places. The list above is long because the PR is 16.7k lines, not because the quality is low.

PR #1982, opened 2026-07-15
fix(game): stop autorun after successful interactions

No898:
## Summary

- Stop autorun and clear click-to-move only after a successful quest, delve, object, loot, or gather interaction.
- Preserve movement for no-op interactions while retaining useful click-to-move approaches to distant interactables and available corpses.
- Share corpse availability rules across the click and loot UI paths, without treating server-private online harvest claim state as a successful interaction.
- Center the locked autorun indicator inside the mobile movement joystick.

## Related issues

Focused extraction of the autorun interaction bugfix from #1761.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/client_shell.test.ts tests/corpse_loot_availability.test.ts tests/gather_node_interact.test.ts tests/interaction_autorun.test.ts tests/interactions.test.ts tests/mobile_controls.test.ts tests/nearby_interaction.test.ts tests/nythraxis_aldric_npc.test.ts tests/css_corpus.test.ts tests/pr_shot_targets.test.ts --maxWorkers=2` (10 files, 252 tests passed)
  - `npx tsc --noEmit` (passed)
  - `.githooks/pre-push` (typecheck passed; 53 guard tests passed and 3 skipped; Biome completed with no errors; copy rules passed)
  - `npm run gate` was attempted and reached the full Vitest phase. It stopped on three failures outside this diff: two persistent `tests/deploy_watchdog.test.ts` failures and one `tests/mail.test.ts` timeout. The mail test passed on isolated rerun; the two deploy watchdog failures persisted. The build phase did not run.
- Manual steps:
  - Used the real mobile joystick gesture in an 844x390 Chrome viewport to enter locked autorun.
  - Verified the locked indicator and joystick centers both resolved to `[78.4, 320.6]`.

## Screenshots / recordings

| Before | After |
| --- | --- |
| Locked autorun before | Locked autorun centered |

---

### Quality

- [ ] **The full gate passes.** `npm run gate` is green. I added or updated decisive tests for changed behavior and recorded any manual checks above.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Verified on a desktop and on a phone-sized viewport in both portrait and landscape. Touch targets stay at least 40x40px and inputs at least 16px font (see `src/ui/CLAUDE.md`).
- [ ] **Accessible.** Keyboard-operable, with visible focus, sensible ARIA, and respect for `prefers-reduced-motion` (only if this change adds or edits UI).

### Localization (i18n)

- [ ] **New player-visible strings follow the contributor policy.** Every user-facing string is a `t()` key added to the matching English catalog. I did not edit locale overlays, except for the five required non-Latin fills when the M16 wordy-copy rule applies (see `src/ui/CLAUDE.md`).
- [ ] Numbers, money, dates, units, and percents go through the i18n formatters (`formatNumber`, `formatMoney`, `formatDateTime`, `Intl`).
- [ ] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary in this same change, and `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene


jgyy (commented):
This is the strongest of the four extractions from #1761. The online harvest-claim guard is the part I was most suspicious of going in, and it holds up: `harvestClaimedBy` is never mirrored to the client (`src/net/online.ts:1065` hardcodes it `null`), and `corpse_loot_availability.ts` correctly forces `harvestable=false` whenever `harvestStateReliable` is false, so an online player can't have a harvest-only corpse falsely read as "available" just because the mirrored claim field always looks empty. The click path (`interactions.ts`) and `hud.openLoot` now share that one gate, and there's no drift between the two callers. Module-first is followed well: `corpse_loot_availability.ts`, `interaction_autorun.ts`, and `nearby_interaction.ts` are all small, independently tested, and `main.ts`'s `interactKey()` shrank to a thin composition instead of growing.

should-fix, `src/game/nearby_interaction.ts:68-95`: the keyboard/gamepad interact-key path treats sending `lootCorpse`/`harvestNode`/`delveInteract`/`pickUpObject` as an immediate "success" for the purpose of stopping autorun, without waiting for a server ack. If the client's mirrored state is momentarily stale (say another party member loots the same corpse the same tick), autorun stops on a command that may no-op server-side. Low stakes since it's UX-only, not economy-affecting, but worth a mental note since it's the opposite of `interactions.ts`'s `handlePickedEntity`, which more carefully treats "opened the loot window" as the success signal rather than the loot grant itself.

nit, `src/game/nearby_interaction.ts:52`: the corpse scan calls `corpseLootAvailability(entity, playerId)` with the default `harvestStateReliable = true`, i.e. the online flag isn't threaded through this path at all. It's inert today because this call site only reads `.hasLoot`, never `.harvestable`/`.canOpen`. But it's a trap for later: if someone refactors this scan to reuse `.canOpen` for consistency with the click path, the online false-harvestable leak this PR just fixed elsewhere would silently come back here. Worth a comment, or threading the flag through now while it's fresh.

nit, `tests/interaction_autorun.test.ts`: solid coverage of the pure module itself (both branches, call counts asserted), but the actual `main.ts` wiring is only checked by `client_shell.test.ts`'s source-string assertions, not a behavioral test. Not blocking, just weaker than the rest of the suite's coverage.

Confirmed the two `npm run gate` failures called out in the PR body (`deploy_watchdog.test.ts`, `mail.test.ts`) are unrelated: this diff touches only `src/game/`, `src/ui/hud.ts`, `src/main.ts`, and `src/styles/hud.mobile.css`, no overlap with either suite or its dependencies, and both passed cleanly in an isolated run against this branch. All 10 named test files (252 tests) pass.

No898:
@jgyy thank you for the careful review. Your should-fix and both nits have been addressed, and the branch is now synced with the current `release/v0.27.0`.

What changed:

- Outcome-bearing online interactions now include a request id. The server replies with a correlated `commandOutcome`, and autorun stops only when the authoritative result is `ok: true`.
- Loot, pickup, gather-node harvest, dungeon enter/leave, delve interaction, and Spirit Healer resurrection now return explicit success or no-op outcomes through `Sim`, `ClientWorld`, `IWorld`, and the server dispatch paths.
- Pending outcomes resolve false on timeout, disconnect, deliberate close/logout, fatal session end, or spectate entry, so rejected or abandoned commands preserve autorun.
- `harvestStateReliable` is threaded through the nearby interaction path, keeping online corpse decisions aligned with the shared availability gate.
- A delayed successful response cannot cancel movement started after the interaction. Releasing the already-held movement input still allows the successful interaction to stop autorun.
- The Spirit Healer button now routes through the same main interaction outcome flow.
- After the release HUD extraction, the new `LootWindowController` was wired back to the shared `corpseLootAvailability` decision instead of duplicating the logic. A regression test proves the controller performs no window side effects when that gate rejects the corpse.
- The locked autorun indicator remains centered in the joystick, layered above it, and disables its transition under reduced motion.

Verification completed:

- 697 focused behavior tests passed across 16 files.
- 43 architecture and schema tests passed.
- `npx tsc --noEmit` passed.
- Scoped Biome CI for all 42 changed TS/CSS files passed with no errors.
- i18n freshness and malware gate passed.
- Post-merge simulation, cross-platform parity, frontend, and test-coverage reviews were all clean.

The full local gate stops at six inherited formatting errors in files unchanged from the current release base, before reaching the remaining stages. The PR-specific diff is clean, and the PR is currently reported as mergeable with a clean merge state.

PR #1983, opened 2026-07-15
feat(map): add world map pinch zoom

No898:
## Summary

Adds focused two-finger pinch zoom to the world map for the v0.27 release branch.

- Tracks touch gesture state in a registered DOM-free pure core with a thin canvas binding.
- Uses an accumulated 8 px deadzone so minor finger jitter does not change zoom.
- Clamps world map zoom to the existing minimum of 1 and maximum of 6.
- Keeps pinch arbitration inside the map HUD path without touching the gameplay camera.
- Handles zero-distance starts, pointer up, cancellation, lost capture, third-touch replacement, and unrelated hybrid-device pointer endings.
- Preserves ordinary marker taps while suppressing every pointer that participated in a pinch.

This is extracted from #1761 so the map gesture can be reviewed and merged independently from the broader mobile HUD redesign.

## Related issues

- Extracted from #1761.
- No issue is closed by this PR.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/map_pinch_zoom_core.test.ts tests/map_pinch_zoom.test.ts tests/map_window_view.test.ts tests/map_window_painter.test.ts tests/architecture.test.ts tests/hud_perf_budget.test.ts tests/painter_host.test.ts tests/focus_manager.test.ts tests/focus_visible_guard.test.ts`
    - 125 passed, 3 skipped.
  - `npx tsc --noEmit`
    - Passed.
  - `npm run ci:changed`
    - Passed. Existing repository-wide warnings remain unchanged.
  - Pre-push QA floor
    - Typecheck, guard tests, Biome, and copy rules passed. Guard tests reported 53 passed, 3 skipped.
  - `.claude/hooks/qa-stop.sh`
    - Passed.
  - `git diff --check`
    - Passed.
  - `npm run gate`
    - Not run because another session had an active Puppeteer E2E process. This follows the repository CPU and load guidance.
- Manual steps:
  - Automated coverage verifies pinch in and out, zero-distance recovery, the 8 px deadzone, zoom bounds, tap suppression, ordinary taps, cancellation, capture loss, third-touch replacement, and unrelated pointer endings.
  - Real-device iOS and Android testing and phone-sized portrait and landscape browser verification have not been run yet.

## Screenshots / recordings

Not included yet. This change adds touch interaction without changing static layout or visual chrome. A mobile pinch interaction recording is still pending.

---

### Quality

- [ ] **The full gate passes.** `npm run gate` is pending because another session was running Puppeteer E2E. Targeted tests, typecheck, changed-file CI, and the pre-push QA floor pass.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Real-device and phone-sized portrait and landscape checks are still pending.
- [ ] **Accessible.** Existing localized zoom buttons remain available as the non-gesture alternative. Manual accessibility verification is still pending.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` files are committed, and `ALLOW_DEV_COMMANDS` is not enabled in any production path.

> Commit message and PR title use Conventional Commits with the required scope, and the commit includes the required explanatory body.

### DIFF
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -351,10 +351,11 @@ import { lowHealthVignette } from './low_health';
 import { lowResourceView } from './low_resource';
 import { mailIndicatorView } from './mailbox_view';
 import { MailboxWindow } from './mailbox_window';
+import { bindMapPinchZoom, finishMapTap, mapTapReleaseFromPointer } from './map_pinch_zoom';
+import { MAP_TAP_MOVE_TOLERANCE_PX, nextMapZoom } from './map_pinch_zoom_core';
 import { type MapRegion, mapCanvasHeight, paintTerrainRows } from './map_terrain';
 import { MapWindowPainter } from './map_window_painter';
 import {
-  MAP_MAX_ZOOM,
   type MapNpcMarker,
   type MapQuestAreaMarker,
   mapWindowMode,
@@ -1808,9 +1809,16 @@ export class Hud {
     );
     $('#map-zoom-in')?.addEventListener('click', () => this.zoomMap(1.4));
     $('#map-zoom-out')?.addEventListener('click', () => this.zoomMap(1 / 1.4));
+    const mapPinch = bindMapPinchZoom(mapCanvas, {
+      onPinchStart: () => {
+        this.mapDrag = null;
+        mapCanvas.style.cursor = '';
+      },
+      onZoom: (factor) => this.zoomMap(factor),
+    });
     // drag to pan (only meaningful while zoomed in; at zoom 1 the whole zone fits)
     mapCanvas.addEventListener('pointerdown', (ev) => {
-      if (!this.mapView || this.mapZoom <= 1) return;
+      if (mapPinch.isPinching() || !this.mapView || this.mapZoom <= 1) return;
       const base = this.mapCenter ?? { x: this.sim.player.pos.x, z: this.sim.player.pos.z };
       this.mapCenter = { ...base };
       this.mapDrag = { px: ev.clientX, py: ev.clientY, cx: base.x, cz: base.z };
@@ -1818,7 +1826,7 @@ export class Hud {
       mapCanvas.style.cursor = 'grabbing';
     });
     mapCanvas.addEventListener('pointermove', (ev) => {
-      if (!this.mapDrag || !this.mapView) return;
+      if (mapPinch.isPinching() || !this.mapDrag || !this.mapView) return;
       const rect = mapCanvas.getBoundingClientRect();
       // "grab the paper" pan: the world point under the cursor stays under it.
       // toMap draws +X to the left and +Z up (mx = (maxX-x)/span, my = (maxZ-z)/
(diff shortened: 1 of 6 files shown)


jgyy (commented):
Good, focused extraction and the gesture handling itself is solid: the pointer rebase logic in `endPinchPointer` recomputes `previousDistance` fresh from the remaining two pointers' current coordinates rather than reusing a stale value, so lifting one finger of the pair (or the third finger) doesn't produce a fake giant-delta zoom jump on the next move. Zoom clamp, deadzone, and third-touch pause/resume are all covered by decisive tests, and `tests/map_pinch_zoom.test.ts` even source-greps `hud.ts` to pin that pinch state gates panning and doesn't touch the gameplay camera. `npx vitest run tests/map_pinch_zoom.test.ts` is green, 10/10.

should-fix, `src/ui/map_pinch_zoom.ts:26-175`: only `mapPinchZoomFactor` and `nextMapZoom` are true pure-core (DOM-free, direct-call tested). The actual pointer-arbitration state machine (the `pointers` Map, deadzone rebasing, third-touch pause, `suppressedTaps`) lives inline inside the `addEventListener` closures in `bindMapPinchZoom`, and is only exercised through a fake-canvas dispatch harness rather than as a pure reducer a test can drive with plain function calls. That's why it's not in `UI_PURE_CORES`. Per the pure-core-plus-thin-painter recipe in `src/ui/CLAUDE.md`, the part that should be extracted as pure is the gesture-state transitions themselves (event-shape in, gesture-state + zoom-factor out), leaving only `addEventListener`/`setPointerCapture`/`preventDefault` as the thin DOM wiring. Not a correctness bug as shipped, but worth splitting before this pattern gets copied for the next touch gesture.

nit, same file: `onPinchStart` only fires on the initial two-pointer transition, not when a pair rebases after one original finger lifts and a third takes its place. Works today because `mapDrag` was already cleared at the original pinch start and panning stays gated on `isPinching()`, but it's an implicit invariant (rebase-without-restart) that could use a one-line comment for the next person touching this file.

nit: no test covers lifting one of the *original* two fingers while a third stays down (only the newest-finger-lifted case is covered). The code path looks identical either way, so this is coverage, not a suspected bug.

The PR body itself flags real-device and phone-sized portrait/landscape verification as not yet done. The automated pointer-event coverage is decisive enough that I wouldn't block on it, but it'd be good to close out before merge.

No898:
@jgyy Thank you for the detailed review. I addressed the full review in `e05e4289d`.

- Extracted the pointer-arbitration state machine into the DOM-free `map_pinch_zoom_core.ts` and registered it in `UI_PURE_CORES`. `map_pinch_zoom.ts` now contains only the thin PointerEvent and pointer-capture wiring.
- Documented that replacing a finger after a third touch only re-baselines the active pair and does not restart the pinch.
- Added direct coverage for lifting an original pinch pointer while the third pointer remains, including the no-second-`onPinchStart` invariant.
- Added regression coverage for zero-distance starts, unrelated mouse or pen end events, one-shot tap suppression, ordinary versus suppressed marker taps, exact 8 px pinch deadzone, and exact 10 px tap tolerance.
- Rebased the branch onto the current `release/v0.27.0` and added the required commit body.

Validation:
- Targeted Vitest suite: 125 passed, 3 skipped.
- `npx tsc --noEmit`: passed.
- Pre-push QA floor: passed.

The full gate and physical iOS/Android verification remain pending because another session was actively running Puppeteer E2E, as noted in the PR checklist.

FernandoX7 (approved):

PR #1984, opened 2026-07-15
perf(render): stabilize nameplate decluttering

No898:
## Summary

Extract the render-only nameplate decluttering follow-up from #1761 into a focused PR targeting `release/v0.27.0`.

- Replace per-frame `Map` bucket bookkeeping with reusable typed-array spatial-hash storage.
- Traverse occupied threshold-sized cells as connected components so transitive overlap chains stack together deterministically by entity ID.
- Avoid Cartesian rescans for dense same-cell, adjacent-cell, and diagonal-cell populations.
- Preserve exact inclusive collision thresholds at extreme finite coordinates without collapsed buckets or false off-viewport collisions.
- Ignore `NaN`, positive infinity, and negative infinity projections without disturbing valid plates.
- Include only the relevant nameplate follow-up found in the local mobile-layout/nameplate work; no mobile HUD or layout changes are part of this PR.

## Related issues

Follow-up to #1761 and its nameplate performance review.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling

## Screenshots

The deterministic three-anchor transitive chain is rendered through the real offline client and the branch's actual declutter module.

### Desktop

| Before | After |
| --- | --- |
| Desktop before: middle and right nameplates still overlap | Desktop after: all three nameplates form a stable stack |

### Mobile landscape

| Before | After |
| --- | --- |
| Mobile landscape before: middle and right nameplates still overlap | Mobile landscape after: all three nameplates form a stable stack |

## How was this tested?

- `npx vitest run tests/nameplate_declutter.test.ts tests/nameplate_projection.test.ts`: 39 passed.
- Mutation check: replacing the diagonal two-dimensional dominance query with independent x/y extrema made the random oracle and both new diagonal orientations fail.
- `npx biome check src/render/nameplate_declutter.ts tests/nameplate_declutter.test.ts`: passed with no diagnostics.
- `npm run ci:changed`: passed; existing repository-wide warnings remain advisory.
- `npx tsc --noEmit`: passed.
- `npm run gate` on Node 22 with GNU-compatible timeout behavior: all 11 steps passed.
  This included 1,134 passing unit/integration test files (2 skipped), 14,332 passing tests
  (22 skipped), 63 browser tests, strict TypeScript and Svelte checks, malware and SFX
  gates, and env, server, and client builds.
- Pre-push QA floor: passed, including 53 guard tests with 3 skipped.
- Manual visual verification in the real offline client at 1440x900 desktop and
  844x390 mobile landscape, with committed before/after evidence above.
- Frontend and test-coverage specialist reviews: READY, with no actionable findings.

### Quality

- [x] **The full gate passes.** `npm run gate` is green. Decisive tests cover typed-buffer
      reuse, connected overlap chains, deterministic stacking, exact thresholds, both
      dense diagonal orientations, far finite coordinates, non-finite inputs, and pooled
      active-prefix reuse.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified at 1440x900 desktop and 844x390 mobile
      landscape using the real offline client.
- [ ] **Accessible.** N/A, this PR adds no UI controls, content, focus behavior, or motion.

### Localization (i18n)

### Hygiene


jgyy (commented):
Solid, well-verified performance rewrite. I stress-tested the central claim myself since it's the riskiest part of this change: I pulled the PR's own `declutterReference` oracle out into a scratch script and ran it against 500 random-seeded crowds at a wider box than the shipped test uses, plus a hand-built four-anchor transitive chain (A overlaps B, B overlaps C, C overlaps D, A and D don't overlap directly). Both the reference and the spatial-hash hot path agreed exactly on every case (`sy = [-30, -10, 10, 30]` for the chain, 0/500 mismatches on the fuzz). So the connected-component traversal really does match its own O(N^2) oracle anchor-for-anchor, not just on the box the checked-in test happens to use.

The performance regression tests are also decisive, not just descriptive: `candidateChecks` is asserted to stay linear (`toBe(anchors.length)`) for dense same-cell and same-bucket-different-cell crowds, and sublinear-in-the-right-sense (`toBeLessThan(anchors.length * 2)` / `* 8`) for the diagonal and long-chain cases, so a regression back to Cartesian rescanning would actually fail CI, not just look slow in practice. The exact-threshold `it.each` table pins inclusive-vs-exclusive at precisely 80/18 on all three axes plus the negative-to-positive boundary, and non-finite anchors are confirmed left untouched while real anchors around them still separate. Buffer reuse across shrinking/growing frames is covered too (epoch-based invalidation rather than clearing, so no stale-tail leak).

nit, `src/render/nameplate_declutter.ts:44`: at cell coordinates beyond `Number.MAX_SAFE_INTEGER`, the function falls back to keying by the raw pixel value instead of the divided cell coordinate, mixing two different keyspaces in the same hash. The comment's reasoning (float ULP at that magnitude already exceeds the overlap threshold, so only bit-identical values could ever collide) holds for genuine extreme-magnitude anchors, but nothing formally rules out an accidental collision between a raw fallback value and an ordinary cell id that happens to equal it. Not reachable at real screen-space magnitudes, so this is a "worth a one-line acknowledgment" nit, not something to hold up the PR for.

Ran `npx vitest run tests/nameplate_declutter.test.ts tests/nameplate_projection.test.ts` myself against the PR head: 38/38 passing, matching the PR description.

No898:
@jgyy thank you for the thorough review and especially for independently fuzzing the implementation and checking the transitive chain.

I addressed the keyspace nit in `8e2095061` by documenting the missing part of the invariant: once the divided cell coordinate is outside the safe-integer range, the raw finite screen coordinate also has a magnitude greater than every safe cell ID, so the fallback and ordinary keyspaces cannot alias.

I also ran a follow-up internal review and hardened the contribution in `30fbb9591`:

- The dense diagonal regression now asserts the actual separated component output in both diagonal orientations, rather than relying only on `candidateChecks`. A mutation back to independent x/y extrema made both new cases and the random oracle fail.
- The allocation wording and metrics now explicitly cover reusable typed hash/traversal buffers, without implying that `Array.sort` or engine-managed allocations are measured.
- Buffer reuse is pinned across a warm → smaller frames → warm-high-water sequence.
- Committed before/after evidence now shows the transitive three-nameplate stack in the real offline client on desktop and mobile landscape.

Final verification is green: 39/39 targeted nameplate tests, all 11 contribution-gate steps, 14,332 unit/integration tests, 63 browser tests, TypeScript/Svelte checks, and all builds. Thanks again for the careful review.

FernandoX7 (commented):
This holds up under adversarial checking. I read the full module at the head ref, ran the tests, and independently fuzzed the hot path against a brute-force BFS reference (460 randomized trials: threshold-aligned boundaries with tiny offsets, negative coordinates, duplicate coordinates, and magnitudes from 2^50 through past the safe-integer cell-id cutoff). It agreed anchor-for-anchor on every trial. tsc and biome are clean, the branch merges clean into release/v0.27.0, and the painter call site (src/render/nameplate_painter.ts:329) needs no change since the metrics parameter is optional and guarded, so the per-frame path pays nothing for it.

Two things I want on the record as intended, since the template's "no behavior change" box is also ticked:

- This does change visible stacking behavior, deliberately. The old pass gathered a star cluster around each id-ordered seed and marked members visited, so later members of a transitive chain could be left overlapping (exactly what the before screenshots show). The new pass stacks true connected components. The rewritten test oracle matches the new semantics, and the transitive-chain test pins it.
- The far-coordinate design is correct, not just plausible. I verified the two load-bearing claims: same-cell anchors stay within the inclusive threshold even under float division rounding near the safe-integer cutoff, and past the cutoff the coordinate ULP already exceeds the threshold (128 vs 80 for x, 32 vs 18 for y), so raw-value keys plus self-aliasing neighbour lookups give exact results there.

Findings:

- nit (src/render/nameplate_declutter.ts:47): the `v === 0 ? 0 : v` guard in the fallback branch is unreachable. When v is 0 or -0, `Math.floor(v / size)` is a safe integer, so the function returns through the normalization on line 48. Harmless, keep or drop.

One process note: CI has not run on this branch yet (no checks reported; that is what the UNSTABLE merge state reflects), so it needs a workflow approval run before merge. The metrics-based performance pins are a nice touch: candidateChecks equality assertions would actually fail on a rescan regression, unlike a timing test.

No898:
@FernandoX7 thank you for the adversarial review and the independent fuzzing.

I addressed the unreachable fallback guard in commit `0797d03a9`. Zero and negative zero always produce a safe cell coordinate and return through the existing safe-cell normalization, so the unsafe-coordinate fallback can only receive a finite, nonzero value. It now returns `v` directly.

I reverified the change with 39/39 targeted nameplate tests, clean TypeScript and Biome checks, and the green pre-push QA floor (53 passed, 3 skipped). Follow-up frontend and test-coverage reviews both reported READY with no findings.

The GitHub checks note remains accurate: no status checks are reported yet, so the workflow still needs approval or a run before merge. Thank you for the careful review.

No898:
@jgyy @FernandoX7 I would appreciate your view on a broader UX and correctness question before this merges.

The committed screenshots show that the basic three-mob case improves HP readability: decluttering moves the whole DOM nameplate, and the three health bars remain distinct. A follow-up review raised a wider concern, though. The current algorithm is anchor-only and does not explicitly protect actionable information:

- `NameplateAnchor` carries only `id`, `sx`, and `sy`. It has no target, threat, casting, actual plate-height, or viewport information.
- Every plate uses the same 20 px stack offset, although a player plate with a title or guild and a mob plate with a cast bar are taller than the basic name-plus-HP layout.
- Every member of a connected component receives a unique vertical slot. For a transitive chain, vertical span therefore grows linearly even when the endpoints do not overlap directly. The long-chain test currently pins this behavior, and final positions are not clamped to the viewport.
- The graphics-fairness policy treats target HP and target or boss cast bars as actionable information. Names are useful, but less important than those combat signals.

Could you share your view on these questions?

1. Should decluttering optimize HP and cast-bar readability first, allowing cosmetic name text to overlap when necessary?
2. Should the targeted, casting, or threatening plate stay closest to its original projected anchor while lower-priority plates move around it?
3. Is full connected-component stacking the desired semantic, or should a stable minimum-lane assignment reuse rows when non-adjacent members do not collide?
4. Should final actionable rectangles be clamped to the viewport so HP and cast bars cannot be displaced off-screen?
5. Do you consider this blocking for this PR, or would you keep this PR focused on the allocation-stable spatial hash and move priority-aware layout semantics into a dedicated follow-up?

I have not changed the implementation for this question yet. I would like to settle the intended gameplay-information priority before changing the pinned stacking semantics.

jgyy (commented):
Went through the algorithm and the tests carefully, this is solid work.

The connected-component fix is real and the tests prove it: the old code only clustered anchors that fell in one anchor's own 3x3 neighbourhood, so a transitive chain (A-B overlap, B-C overlap, A-C too far) wouldn't necessarily stack all three together depending on visit order. The new BFS over occupied cells fixes that, and `stacks a transitive chain where the endpoints do not directly overlap` (tests/nameplate_declutter.test.ts) pins it.

The far-coordinate handling is also correct, not just clever. `cellCoord` falls back to the raw value once `v / size` stops being a safe integer. The comment claims that beyond that magnitude the float ULP already exceeds the collision threshold, so only exactly-equal values can collide. I checked the numbers: the fallback only triggers once `v > size * 2^53`, and at that magnitude the ULP for the X threshold (80px, size in the check) is already >= 128, well past the threshold. So keying by the raw value instead of a divided cell id can't introduce a false negative, distinct-but-close-enough values physically can't exist as separate floats at that magnitude. Good reasoning, and it's exercised by the "does not round a far diagonal gap down" / "does not merge distinct far coordinates" cases.

Ran `npx vitest run tests/nameplate_declutter.test.ts tests/nameplate_projection.test.ts` and `npx tsc --noEmit` on the branch, both clean. `declutterNameplatesInPlace`'s only caller (`nameplate_painter.ts:329`) still passes the same two args, the new `metrics` param is optional and additive.

One thing I'd push back on gently, not a blocker: this is a lot of machinery (a hand-rolled open-addressing hash table with epoch stamps, x-sorted segments with suffix y-extrema, binary search for diagonal cells) for a pass that's interest-scoped to maybe a few dozen visible plates at once. The old Map-based bucket scan wasn't quadratic in any realistic crowd, it was only quadratic in worst-case synthetic stress (which is what the new tests target). A simpler fix, BFS the connected component using the existing Map buckets, keep the old clamp-to-16-bit-cell approach, would have fixed the actual reported bug (transitive stacking) without this much new surface area to maintain. Worth considering for future review, not asking for a rewrite here given the extensive test coverage already in place.

Nit: `neighborSlots`/`neighborSlotCount` dedup loop in the BFS looks defensive against two different (dx, dy) offsets landing on the same cell slot, but since cell coordinates are exact integers and each offset produces a distinct (cx+dx, cy+dy) pair, that should never happen. Not wrong, just extra code that never does anything.

Screenshots and manual verification are a nice touch for a render-only change like this.

PR #1864, opened 2026-07-13
feat(hud): configurable movable party and raid frames

Blaine1705:
## Summary

Rework the compact party/raid frames into a **movable, configurable** unit-frame
family, while keeping everything existing untouched so nothing regresses.

**Frames**
- The party/raid stack becomes a `MovableFrame` like the player and target frames:
  drag to reposition, lock/unlock, persisted top-left.
- Each member renders through the shared `UnitFramePainter` family: class-colored
  health, resource, a classic absorb overlay, dead / out-of-range / in-combat and
  aggro states, and a compact aura strip filtered to actionable effects
  (`isPartyFrameRelevantAura`: debuffs, HoTs, absorbs, major defensives), plus
  an optional incoming-heal HP overlay.
- Leaving the party moves from a permanent per-frame button to the self portrait
  context menu, freeing the compact row.

**Configuration (Interface options, all read live)**
- Frame style (automatic / classic / raid), scale, width, height, spacing, raid
  columns, health-text mode, sort (group / role / name), and toggles for resource,
  absorbs, auras, and showing your own frame.

## Why this is low-risk (additive only)

- The **shared unit-frame core, the player/target frames, and the server/sim party
  wire are untouched.** `unit_frame` gains `showAbsorbText` + segment-absorb view
  fields *additively* while KEEPING the Book of Deeds title fields; `absorb_bar`
  returns the new segment fields *additively* (existing `total`/`fillFrac`/
  `overshield` unchanged). Party frames use the classic left-origin absorb overlay
  through the unchanged painter, identical to the player/target frames.
- `PartyMemberInfo` / `PartyMemberAura` gain their new fields as **optional**, so no
  server/sim wire change is needed and the **parity goldens are unchanged**.
- `isPartyFrameRelevantAura` is a new, purely additive `src/sim/aura_classify.ts`
  helper; the existing `isDebuffAura` classification is not touched.
- Clique-style mouseover casting is intentionally **deferred** (it needs a new
  authoritative cast-on-target sim path).

## Related issues

N/A.

## Type of change

- [x] Feature: new functionality

## How was this tested?

- Commands (all green):
  - `npx tsc -p tsconfig.json --noEmit`
  - `npx vitest run` (full suite; the only failures are pre-existing
    Windows-only environment issues in SFX/static-serving tests, unrelated to this
    change; CI on Linux is authoritative)
  - `npm run build`
  - `npx vitest run tests/parity` (goldens unchanged), `tests/party_machine.test.ts`,
[...]

PR #1991, opened 2026-07-15
fix(desktop): force the discrete high-performance GPU on hybrid systems

### LINKED ISSUE #1990: Critical: Windows desktop app runs on integrated GPU instead of the discrete NVIDIA GPU on Optimus laptops
## Problem

On hybrid (NVIDIA Optimus) Windows laptops, the desktop app runs on the integrated GPU instead of the discrete NVIDIA GPU, which causes severe low FPS. A player with an RTX 3070 laptop reported about 13 FPS and a 75ms frame time in the new Windows installer. The same machine runs the app well when an external monitor is connected, because external display outputs on most laptops are wired directly to the discrete GPU.

Electron/Chromium binds a single GPU adapter when its GPU process starts, and on Optimus laptops it commonly selects the integrated GPU when the app is on the internal panel. The client already requests WebGL `powerPreference: 'high-performance'`, but on Windows that per-context hint does not switch adapters, and the shell (`electron/main.cjs`) currently sets no GPU command-line switches and no OS-level per-app GPU preference. So nothing forces the discrete GPU, and the player is left to fix it by hand (Windows graphics settings or NVIDIA Control Panel), which most players will never do.

This is high severity: the Windows app was just deployed, so new installs on hybrid laptops are affected out of the box.

## Steps to reproduce

1. Install the Windows desktop app on a laptop with NVIDIA Optimus (an Intel or AMD integrated GPU plus a discrete NVIDIA GPU, for example an RTX 3070 laptop).
2. Launch the app on the laptop's internal display (no external monitor attached).
3. Enter the world and observe the FPS readout under load.

## Expected behavior

- The app uses the discrete high-performance GPU automatically, with no manual Windows or driver configuration.
- A capable discrete GPU (for example an RTX 3070) renders the game at a high frame rate on the internal laptop panel, matching what it does on an external monitor.

## Actual behavior

- The app binds to the integrated GPU on the internal panel and renders at about 13 FPS (75ms frame time).
- Performance only becomes acceptable when the player attaches an external monitor or manually forces the discrete GPU in Windows or NVIDIA settings.

## Environment

- Platform: Windows desktop app (Electron), hybrid NVIDIA Optimus laptop.
- Surface: the game client running inside the desktop shell.

## Scope

- `electron/` desktop shell only (no game client or sim changes).
- Force the discrete GPU with zero user action: the Chromium `force-high-performance-gpu` switch plus the Windows per-app GPU preference (`HKCU\Software\Microsoft\DirectX\UserGpuPreferences`, value `GpuPreference=2;` keyed by the app exe path, which is what Windows graphics settings "High performance" writes and which overrides the NVIDIA Control Panel on Windows 10 20H1 and later).
- Windows is the target. The Chromium switch is harmless on single-GPU machines and beneficial on macOS dual-GPU, so it can apply on all platforms.

## Acceptance criteria

- On a hybrid Optimus laptop, a fresh install runs on the discrete GPU with no manual configuration, and the discrete GPU string appears in the shell log's `[gpu] active renderer` line.
- FPS on the internal laptop panel matches FPS on an external monitor for the same machine.
- The behavior is a no-op on single-GPU machines and does not require elevation.
- The app still boots normally if the GPU preference cannot be applied (the write is best effort).

## Notes

- Corroborating signal: the in-game graphics panel auto-selected Medium in the report, which is what the auto-tier picks for an Intel Iris Xe integrated GPU. A real RTX 3070 would auto-select Ultra, so the client is seeing the integrated GPU, not the discrete card.
- The client-side `powerPreference: 'high-performance'` request is not sufficient on Windows because Chromium selects one adapter for the whole GPU process at startup.

FernandoX7:
## Summary

Forces the desktop app onto the discrete high-performance GPU on hybrid (NVIDIA/AMD Optimus) systems, with zero user action. This fixes the "good GPU, terrible FPS" reports: a player with an RTX 3070 laptop was getting about 13 FPS (75ms frame time) on the internal panel, while the same machine ran fine on an external monitor (external outputs are wired to the discrete GPU).

Root cause: Chromium binds a single GPU adapter when its GPU process starts, and on Optimus laptops it commonly picks the integrated GPU on the internal display. The client already requests WebGL `powerPreference: 'high-performance'`, but on Windows that per-context hint does not switch adapters, and `electron/main.cjs` set no GPU switches or OS-level preference.

The fix adds a small `electron/gpu_preference.cjs` module that `main.cjs` calls before app ready, using two independent levers:

1. Chromium's `force-high-performance-gpu` switch. Chromium 150 (Electron 43) registers the switch name with hyphens; Electron's docs spell it with underscores, and Chromium matches switch names exactly, so both spellings are appended (an unrecognized switch is a harmless no-op). Harmless on single-GPU machines, honored on macOS dual-GPU.
2. The Windows per-app GPU preference: it sets `GpuPreference=2` in `HKCU\Software\Microsoft\DirectX\UserGpuPreferences` keyed by the app exe path, which is what Settings > System > Display > Graphics > High performance sets, and it overrides the NVIDIA Control Panel on Windows 10 20H1 and later. That one value can hold other semicolon-separated per-app tokens (the Windows 11 "Optimizations for windowed games" toggle stores `SwapEffectUpgradeEnable`, per-app Auto HDR stores `AutoHDREnable`), so the write queries the stored data first and replaces or appends only the `GpuPreference` token, preserving every sibling token. It runs only in packaged builds (an unpackaged dev run would pin an orphan entry keyed to the checkout's `node_modules` electron binary) and is best effort (a failed write is logged and swallowed so the app always boots). The write is skipped when the value already pins high performance.

Scope is the `electron/` shell only. No game client, sim, or server changes.

## Related issues

Closes #1990.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npm run gate` PASS locally (all 11 steps green: i18n gen and freshness, malware scan, changed-files biome, SFX conformance, full test suite with bounded workers, tsc, all five builds).
  - `npx vitest run tests/electron_gpu_preference.test.ts` (26 unit tests, green): the token-merge matrix (replace in place, append when absent, case-insensitive, duplicate collapse), the preserved-sibling-tokens regression, the `reg query` output parser, both switch spellings, the exact `reg` argv and pinned exec options, the "already high performance" short-circuit, overwrite of a power-saving preference, the unpackaged (dev-run) skip plus the missing-isPackaged fail-closed case, the getPath-throws and empty-exe-path fallbacks, non-Windows no-op, and never-throws-on-failure.
  - `npx vitest run tests/malware_scan.test.ts` (79 tests, green): the file-scoped pathSev exception is pinned to length, exact regex source, flags, and severity, with a scanText probe that eval / decode-then-execute / credential-file reads still fire high inside the exempted file.
  - `npx vitest run tests/electron_` (all 14 electron suites, green, including the four that scan `main.cjs`).
  - `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts` (guard tests green).
  - `node scripts/malware_scan.mjs --gate` PASS (0 high after priors).
- Manual steps: the arg-building, query parsing, token merging, and the write decision are dependency-injected and unit-tested without a real registry. On-device verification on an Optimus laptop is the follow-up: launch on the internal panel and confirm the shell log's `[gpu] active renderer` line names the NVIDIA GPU (not the Intel/AMD iGPU) and FPS matches the external-monitor case.

## Screenshots / recordings

N/A: this changes runtime GPU selection, not any rendered UI. The symptom screenshot (graphics panel at 13 FPS) is attached in #1990.


Rubsey (commented):
This is a well-built fix. The two-lever approach is right for Optimus (the WebGL powerPreference hint really does not switch adapters on Windows), the module is dependency-injected with the load-bearing literals pinned in tests, the exec path is clean (execFileSync, fixed argv, absolute reg.exe path, no shell, bounded by timeouts), and the main.cjs call site runs before app ready with no other switch writer anywhere in electron/ to conflict. I ran the suites: the two changed files pass (89 tests) and the full electron sweep matches your claim exactly (14 files, 172 tests), with tsc and the changed-files biome check clean. The pre-existing [gpu] active renderer line at main.cjs:495 makes the described on-device check real.

The i18n untracking commit also checks out end to end: the #1864 party-frames merges resurrected the two baselines that 408f3dea7 had untracked, base push CI is currently red on exactly those two guard tests, nothing in the repo consumes a committed copy of either file, and check-ignore confirms both rules match. Bundling it was the right unblock and the commit message discloses it properly.

Findings:

- should-fix, electron/gpu_preference.cjs:50 (buildRegWriteArgs): the reg add replaces the entire REG_SZ with GpuPreference=2;, but Windows packs multiple semicolon-separated tokens into that one per-app value: the Windows 11 "Optimizations for windowed games" toggle stores SwapEffectUpgradeEnable=1; and per-app Auto HDR stores AutoHDREnable tokens alongside GpuPreference=N; in the same string. A stored SwapEffectUpgradeEnable=1;GpuPreference=0; fails alreadyHighPerformance, so the write path silently deletes the user's other per-app graphics settings. Parse the queried data and replace or append only the GpuPreference token, preserving the rest. (The combined value that already contains GpuPreference=2 is fine: the regex matches and the early return at line 120 skips the write. The header's "EXACTLY what Settings writes" claim only holds when no other toggle was ever set for the exe.)

- should-fix, tests/malware_scan.test.ts:338: the new pathSev exception is not pinned to the standard the redact.ts precedent sets a few lines below. The walletId test pins pathSev toHaveLength(1) plus the exact re.source ("widening it must show up as a test diff") and adds a scanText probe that the demotion does not blind the other rules inside the exempted file. The new test only probes four effectiveSeverity paths, so widening the regex (say to also cover updater.cjs) or appending a second pathSev entry to the child_process rule would pass with no test diff. Add the same length plus source pin, and ideally the scanText probe for gpu_preference.cjs so the "eval / egress / decode-then-execute still trip at high" claim in the note and catalog doc is tested rather than asserted.

- nit, electron/gpu_preference.cjs:106: an unpackaged Windows dev run resolves app.getPath('exe') to the checkout's node_modules electron.exe and writes a registry entry keyed to that path: one orphan per worktree, and it forces the discrete GPU for anything else launched from that same binary. Consider gating the registry lever (not the switches) on app.isPackaged; the "NSIS install path is stable" rationale in the header only holds for packaged builds.

- nit, tests/electron_gpu_preference.test.ts:147: calls.every() passes vacuously on an empty array; add expect(calls.length).toBe(2) so the test still bites if the win32 branch stops calling exec.

Notes, no action needed:

- Overwriting an explicit power-saving choice (GpuPreference=1) on every launch is deliberate, documented, and pinned by a test, and since that stored state is exactly what produces the 13 FPS reports, forcing it is a defensible default. Worth remembering the only user opt-out is now uninstalling; if you rework the write for the first finding, keep this behavior.
- Boot cost is fine: bounded by the two 4s timeouts, and the steady state after the first launch is a single query before any window exists.
- Small test gaps, take or leave: the getPath-throws fallback and the empty-exePath early return are untested, and the exec options (timeout 4000, windowsHide) are load-bearing but unpinned.

FernandoX7:
Thanks for the thorough review. All four findings are addressed, plus the take-or-leave gaps, and the branch is re-merged with the base so the conflict is gone.

- Token-preserving write (should-fix): the write now queries the stored value and replaces or appends only the `GpuPreference` token, keeping `SwapEffectUpgradeEnable` / `AutoHDREnable` and any other sibling tokens in place. Two new pure helpers (`parseRegQueryData`, `mergeHighPerformancePreference`) carry it; the merge matrix is unit-tested (replace in place, append when absent, case-insensitive match, duplicate collapse), and the regression you described is pinned end to end: a stored `SwapEffectUpgradeEnable=1;GpuPreference=0;` now produces a `/d` payload of `SwapEffectUpgradeEnable=1;GpuPreference=2;`. The already-2 early return and the deliberate overwrite of a power-saving preference both keep their behavior, now token-preserving, and the header's "exactly what Settings writes" claim is reworded to match.
- pathSev pin (should-fix): the child_process exception is now pinned to the redact.ts standard: `toHaveLength(1)`, the exact `re.source`, and the severity, so widening the regex or appending a second entry shows up as a test diff. A scanText probe confirms eval, decode-then-execute, and credential-file reads still fire high inside the exempted file while the process-execution hit stays medium and present.
- isPackaged gate (nit): the registry lever now runs only when `app.isPackaged`; the switches still append everywhere. The unpackaged skip is tested.
- Vacuous `every()` (nit): the reg.exe resolution test now pins `calls.length` to 2 first.
- The optional gaps too: the getPath-throws fallback to `process.execPath`, the empty-exe-path early return, and the exec options (timeout 4000, windowsHide, both stdio shapes, pinned with `toEqual` so a dropped bound or a `shell: true` would fail) are all covered now.

Two extra pins came out of a coverage audit on the fixes themselves: a missing-isPackaged case (the gate is strictly `=== true`, so a partial app object fails closed, now asserted) and a `re.flags` pin on the pathSev entry so even adding a flag shows up as a test diff.

The merge takes the base's `.gitignore` wording for the `i18n.resolved.sha256` entry (the untracking fix crossed with the same fix landing on the base). Suite status: `electron_gpu_preference` 26 tests, `malware_scan` 79, the full electron sweep 14 suites, both guard tests, the scanner gate (PASS, 0 high), biome / tsc clean on the changed files, and a full local `npm run gate` PASS (all 11 steps).

PR #1992, opened 2026-07-15
fix(sfx): verify real post-encode true peak on the LUFS conform branch

jamiecypher:
## Summary

- The LUFS-target conform branch (clips >= 1s) only ever verified LUFS convergence, never re-measuring the actual post-encode true peak. MP3 encoding can push a clip's true peak above whatever the pre-encode limiter saw, so a file could land within spec on loudness while genuinely clipping on real playback.
- The checker (`inspectSfxConformance`) compounded this: it only ever measured peak for the short-clip branch, never for LUFS-target clips, so `npm run sfx:check` reported clipping files as passing.
- Confirmed at scale: 124 currently-shipped files, including several already-published custom recordings and much of the stock ambience/spell-cast/mob library, were silently over the -6dBFS true-peak ceiling.

## What changed

- `scripts/sfx/conform_audio.mjs`: the LUFS convergence loop now independently verifies true peak on every attempt and treats peak safety as the binding constraint, falling back to the best peak-safe (if slightly under-loud) result when the target is structurally unreachable without clipping. `LONG_FORM_LIMIT_DB` is now derived from `TARGET_PEAK_DBFS` instead of a decoupled hardcoded value.
- `scripts/sfx/sfx_conform_rules.mjs`: the checker now always measures true peak regardless of branch, and `classify()` flags an LUFS-branch peak overshoot unconditionally. It also correctly avoids flagging an under-target LUFS as a problem when the file is genuinely peak-constrained (a linear boost-to-target projection against the peak ceiling, using the same tolerance window already used elsewhere in this file).
- `scripts/sfx_studio/audio_io.mjs`: SFX Studio's own production-mastering validator updated to accept (and gracefully ignore on pre-existing records) the new diagnostic field.
- `tests/sfx_conform.test.ts`: 3 new tests covering the LUFS-branch peak-overshoot detection, plus one corrected pre-existing test that had been asserting the old, incorrect behavior.
- Reconformed every file the fixed checker now catches, to actually meet the already-documented spec (no target changes, `-6dBFS`/`-14 LUFS` unchanged).

## Known unrelated CI flake

`tests/i18n_resolved_equivalence.test.ts` and `tests/i18n_status_registry.test.ts` may fail in this PR's CI run due to a pre-existing bug on `release/v0.27.0`, unrelated to this change: `src/ui/i18n.resolved.sha256` and `src/ui/i18n.status.summary.json` were explicitly untracked in commit `408f3dea7` (they can't be textually merged across concurrent PRs), but PR #1870 re-committed both. Confirmed reproducible on a clean `release/v0.27.0` checkout with zero changes. Filed/flagged separately; not something this PR introduces or can fix in scope.

## How this was tested

- `npx vitest run tests/sfx_conform.test.ts tests/sfx_studio.test.ts`: 101/101 passing
- `npx tsc --noEmit`: clean
- `npx @biomejs/biome check` on every changed source file: clean
- `node scripts/sfx_conform.mjs` (check mode): 0 files out of spec after the fix
- Full `npm run gate`: passes except the two known-unrelated pre-existing failures noted above (also verified independently on a clean, unmodified checkout)

## Type of change

- [x] Bug fix

## Checklist

- [x] The full gate passes, apart from the documented pre-existing unrelated failure.
- [x] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is not enabled in any product.


jgyy (commented):
Good catch and a clean fix. The core bug was real: the LUFS-target branch trusted the pre-encode limiter's peak reading and never re-measured true peak after MP3 encoding, and the checker mirrored that blind spot by only ever measuring peak on the short-clip branch. Verified the 124-file claim isn't hand-waved: ran `node scripts/sfx_conform.mjs` (check mode) against this branch and every shipped file reports `ok`, so the reconform actually closed the gap it claims to close.

- `bestPeakSafeAttempt` fallback (`scripts/sfx/conform_audio.mjs:255-278`) is the right call for wide-crest-factor sources: landing safely under target beats hard-failing the build.
- `classify()`'s new peak-constrained check (`scripts/sfx/sfx_conform_rules.mjs:107-115`) correctly avoids flagging an under-target LUFS as a defect when the shortfall is peak-imposed, using the same linear hypothetical-boost logic as the fallback that produced it. Consistent reasoning on both sides.
- `audio_io.mjs`'s `measuredPeakDb` handling is properly optional for historical records, so old SFX Studio projects still load.
- The one previously-wrong test (`ignores peakDb for long clips`) was correctly flipped rather than left in place; the new field coverage (over/at/under the ceiling) is solid.

Ran `npx vitest run tests/sfx_conform.test.ts tests/sfx_studio.test.ts` (101/101) and `npx tsc --noEmit` (clean) locally against the PR head, matches what's claimed.

Only the two generated files (`public/audio/sfx/runtime-pack.json`, `src/game/sfx_manifest.generated.ts`) conflict against current `release/v0.27.0`, and that's expected regen churn from the reconform, not a real conflict, just needs a rebase + `npm run build` before merge.

Nothing blocking from me. The unrelated i18n CI flake writeup checks out too, that's pre-existing base breakage from #1870, not introduced here.

PR #1995, opened 2026-07-16
feat(ui): let the All chat tab hide system and event lines (#1670)

daxdax89:
## Summary

The All chat tab dumped system and event lines (loot, XP, zone welcome, quest and system notices) in with player chat and offered no way to filter them, so busy moments buried actual conversation.

This adds a persisted "Chat Only" toggle to the chat tab strip, shown while the All view is active, that hides `chan` `system` lines so player chat reads clean. It is off by default, so the classic All view is unchanged for anyone who does not opt in.

Every chat line already carries its classification via `dataset.chan` (system lines vs a real channel), so no new plumbing was needed: the visibility rule is lifted into a pure, unit-tested predicate shared by the existing per-channel tab filter and the new toggle.

- `src/ui/hud/chat/chat_channels.ts`: new DOM-free `chatLineVisible(filterTab, chan, hideSystem)` predicate (the shared home for chat-tab rules).
- `src/ui/hud/chat/chat_window_controller.ts`: persist the opt-in alongside the chat-tab keys, render the toggle as an `aria-pressed` control just left of the "+" button (shown only on the All view), and route the filter / `hideIfFiltered` path through the predicate.
- `src/styles/hud.css` and `src/styles/hud.mobile.css`: the pressed state reuses the existing active-tab look (tokens, no new colors); the control is hidden off the All view, and the hide is mirrored in `@layer hud-mobile` so it survives the mobile `display: inline-flex` on `.chat-tab` (same layer hazard the `#chat-dismiss` rule documents).
- i18n: one English-only `hudChrome.chatFilter.chatOnly` key with the value "Chat Only".

## Related issues

Part of #1670. This ships the all-or-nothing system/no-system switch as a first step; the issue's finer per-category filtering (independently hiding loot, XP, quest, and minigame/event lines, which needs richer `chan` tagging in `appendLog`) is left for a follow-up, so the issue stays open.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/chat_channels.test.ts` (predicate cases: All + toggle off shows everything; All + toggle on hides only `system`; a channel tab hides other channels regardless of the toggle)
  - `npx vitest run tests/chat_window_controller.test.ts` (DOM behavior via the fake-DOM harness: the toggle starts from the persisted value; a click flips it, persists `woc_chat_hide_system`, and re-applies `chat-hidden` to existing log lines; it hides off the All view while never taking `aria-selected` or `tabIndex` from the tab loop)
  - `npx vitest run tests/chat_tab_strip.test.ts` (styles pin: the mobile mirror keeps the hide winning over the `hud-mobile` layer)
  - `npx vitest run tests/architecture.test.ts tests/css_corpus.test.ts` (UI pure-core completeness + purity; CSS brace balance)
  - `npx vitest run tests/i18n_completeness.test.ts tests/chat_filter.test.ts`: green
  - `npm run i18n:gen` is idempotent (tree stays clean; the new key is English-filled and marked pending as expected)
  - `npx tsc --noEmit` clean for the changed files; `npx @biomejs/biome check` clean on the changed files
- Manual steps: booted the offline client, seeded a realistic All-tab mix (party/whisper/general/say plus loot/XP/zone/kill system lines), and confirmed the toggle flips `aria-pressed` and hides the system lines while player chat stays, then restores them when toggled off. The toggle appears only on the All view.

The new key value "Chat Only" is intentionally terse (no run of 4+ lowercase letters) so it is a legal English-only add under the M16 completeness rule.

## Screenshots / recordings

The control is a compact tab-strip button labeled "Chat Only", inline just left of the round "+" add-channel button, styled exactly like a chat tab; it is shown only while the All tab is active. Off (default) it takes the de-emphasized tab look and the All log interleaves system/event lines with player chat; pressed (`aria-pressed="true"`) it takes the active-tab gold look and the system/event lines disappear, leaving only player chat.

I was not able to attach rendered before/after chrome captures from my headless dev environment (the HUD chrome did not lay out under headless offline boot here), but the behavior is fully unit-tested and I verified it live in the browser as described above. Happy to add rendered screenshots if useful.

### Cross-platform

- [x] The toggle reuses the existing `.chat-tab` styling, which already carries the mobile 40px touch floor and 16px input font; it is a real `button` with an accessible name via `t()`, `aria-pressed` state, and inherited `:focus-visible`. The off-the-All-view hide is mirrored in `@layer hud-mobile` so it holds on mobile-touch clients, pinned by `tests/chat_tab_strip.test.ts`.

### Localization (i18n)

- [x] New player-visible string follows the contributor policy: one English-only `t()` key in the matching catalog; no locale overlays edited; the terse value is M16-legal.
- [x] N/A for formatters (no numbers) and for sim/server text (this is client UI only).

### Hygiene

- [x] Generated i18n tables regenerated with `npm run i18n:gen`, not hand-edited.


EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the chat controller, `chat_channels.ts`, the styles layer contract in `src/styles/CLAUDE.md`, and issue #1670.

What I ran locally in a worktree at this head:

- `npx vitest run tests/chat_channels.test.ts tests/chat_tab_strip.test.ts tests/chat_filter.test.ts tests/chat_window_controller.test.ts tests/chat_window.test.ts`: 84 passed.
- `npx vitest run tests/architecture.test.ts tests/i18n_completeness.test.ts tests/i18n_resolved_equivalence.test.ts tests/localization_fixes.test.ts`: green.
- `npm run i18n:gen`: idempotent, tree stays clean (pending=20, the one new key per non-English locale, as expected for an English-only add).
- `tsc --noEmit`: no errors in any file this PR touches (the only errors in my tree are pre-existing ones in `src/net/wallet_connect.ts` / `tests/mobile_wallet_deeplink.test.ts`, unrelated).

The shape of the change is good: lifting the visibility rule into a DOM-free `chatLineVisible` shared by both call sites is the right seam, the predicate reads correctly for the All/combat (`filterTab === null`) and channel-bound cases, filtered lines stay out of the live region because `hideIfFiltered` runs before `announceChatLine`, and the combat pane is untouched. Three findings below.

**1. (high) The toggle is NOT hidden off the All view on mobile: layer order beats the new rule.**

`.chat-tab-sysfilter-hidden { display: none }` lands in `@layer components` (hud.css), but `body.mobile-touch .chat-tab { display: inline-flex }` lives in `@layer hud-mobile`, which the barrel declares AFTER components (`src/styles/index.css`, the `@layer tokens, base, layout, components, hud, shell, hud-mobile, ...` line). Layer order wins over specificity, so on any `body.mobile-touch` client the `display: inline-flex` declaration overrides the hide and the "Chat Only" button stays visible on the Combat tab and on every channel tab. Tapping it there silently flips a setting with no visible effect, which is exactly the confusing state the PR set out to avoid. hud.css already documents this hazard at the `#chat-dismiss` rule ("hud.mobile.css, a later `@layer`, so it wins there").

Repro: boot with `body.mobile-touch`, open a channel tab, the toggle is still in the strip.

Fix options: mirror the hide in `hud.mobile.css` under `@layer hud-mobile`, or drop the class and toggle the `hidden` property / an inline `display` from JS so it is not subject to the layer cascade at all.

**2. (medium) No test covers the DOM half of the feature.**

Only the pure predicate is tested. `tests/chat_window_controller.test.ts` already drives this controller with a fake DOM, so the untested behaviors are cheap to pin and are where the regression above lives: `aria-pressed` starts from the persisted value, a click flips it, persists `woc_chat_hide_system`, and re-applies `chat-hidden` to the existing log children; the toggle carries `chat-tab-sysfilter-hidden` off the All view; and `updateActiveTabStyles` never gives it `aria-selected` or `tabIndex = -1`. A styles pin in `tests/chat_tab_strip.test.ts` asserting the hide survives the mobile layer would also have caught finding 1.

**3. (medium) "Closes #1670" overclaims the scope.**

The issue's acceptance criteria ask that a player can "independently hide loot, XP, quest progress, and minigame or event lines", and it calls out finer `chan` tagging in `appendLog` (today everything non-player is one flat `system` bucket) as a prerequisite. This PR ships a single all-or-nothing system/no-system switch, which is a reasonable first step but is not those criteria. Suggest retargeting the PR body to "Part of #1670" and leaving the issue open, or splitting the per-category work into a follow-up that the issue tracks.

Minor, take or leave:

- The `+` button is `position: sticky; right: 0` inside the scrolling strip, so once enough channel tabs are open the toggle scrolls out from under it rather than staying "just left of the +". Making it sticky too (or seating it at the strip's left edge) keeps it reachable.
- `sysButton.title = sysLabel` duplicates the visible text, so it produces a redundant tooltip; the visible label plus `aria-pressed` is already a complete accessible name.
- The strip is `role="tablist"` and now holds two non-tab buttons. The `+` set that precedent so this is not new, but if you touch it, `role="presentation"` on a wrapper or moving both controls outside the tablist would be more correct.

The i18n handling, the persistence pattern, and the module placement all look right to me. Happy to re-review once finding 1 has a test behind it.

daxdax89:
Thanks for the thorough review, all three findings addressed in the latest push (rebased onto current `release/v0.28.0`).

**1 (high) mobile layer order.** Fixed. `.chat-tab-sysfilter-hidden { display: none }` stays in `@layer components`, and `hud.mobile.css` now mirrors it under `@layer hud-mobile` with `body.mobile-touch .chat-tab.chat-tab-sysfilter-hidden` (the extra `.chat-tab` outranks the `display: inline-flex` base rule, so it wins independent of source order). I mirrored rather than switching to inline JS `display` so the toggle keeps its layered styling and matches the `#chat-dismiss` precedent you pointed at; I also expanded the `#chat-dismiss`-style hazard note on the base rule in hud.css.

**2 (medium) DOM coverage.** Added to `tests/chat_window_controller.test.ts`: `aria-pressed` starts from the persisted value (both the persisted-on and defaulted-off arms), a click flips it, persists `woc_chat_hide_system`, and re-applies `chat-hidden` to existing log children, and `updateActiveTabStyles` gives it `chat-tab-sysfilter-hidden` off the All view while never handing it `aria-selected` or `tabIndex = -1`. Plus the styles pin you suggested in `tests/chat_tab_strip.test.ts` asserting the mobile mirror survives the layer (that pin fails without the finding-1 fix).

**3 (medium) scope.** Retargeted the PR body to "Part of #1670" and left the issue open. This ships the all-or-nothing switch as a first step; the per-category filtering (finer `chan` tagging in `appendLog`) is a follow-up the issue still tracks.

On the minor notes: I left the sticky-`+` interaction, the redundant `title`, and the `role="tablist"` shape as-is for now to keep this change scoped to the reported bug and the missing tests, but happy to fold any of them in if you'd prefer. Ready for re-review.

EnriqueGF (commented):
Reviewed the full diff (predicate, controller, CSS, i18n catalog + regenerated tables, tests) against the root and `src/ui` / `src/styles` / `tests` CLAUDE.md rules.

**What I ran here in a clean worktree off `release/v0.28.0`:**

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/chat_channels.test.ts tests/chat_window_controller.test.ts tests/chat_tab_strip.test.ts tests/i18n_completeness.test.ts tests/css_corpus.test.ts`: green (97 tests).
- `npm run i18n:gen` then `npx vitest run tests/localization_fixes.test.ts` (S3 guard): green, and the regen left the tree clean, so the generated tables in the diff are genuinely regenerated, not hand-edited.

**What holds up:**

- The change is client-only presentation, so no sim, wire, or `IWorld` surface is touched; nothing determinism- or server-authority-relevant.
- `chatLineVisible` is a proper DOM-free pure core in the existing `chat_channels.ts` home, and it is a faithful lift of the old rule: with `hideSystem` false it is exactly `filter !== null && chan !== filter` negated, so the default path is behavior-preserving.
- i18n is correct: one English-only key in `src/ui/i18n.catalog/hud_chrome.ts`, no overlay edits, the visible label and the `title` both come from `t()`.
- The mobile layer mirror is right and well justified: the base `.chat-tab-sysfilter-hidden` (0,1,0) in `@layer components` genuinely loses to `body.mobile-touch .chat-tab { display: inline-flex }` in the later `@layer hud-mobile`, and the mirror carries `.chat-tab` so it wins on specificity regardless of order. Good that it is pinned by `tests/chat_tab_strip.test.ts`.
- Excluding the toggle from the `aria-selected` / `tabIndex` loop is the right call, and the test asserts both arms.

**Finding (minor, worth fixing before merge): the toggle can appear on a non-All tab after a tab-strip re-render.**

`renderTabs()` rebuilds every button and appends the toggle without `chat-tab-sysfilter-hidden`; only `updateActiveTabStyles()` applies that class, and `renderTabs()` does not call it. `addTab()` re-renders without selecting when the tab is added from typed input, so:

1. Player is on a channel tab (say `world`), toggle correctly hidden.
2. Player types `/join lfg` -> `syncTabsForInput` -> `addTab(..., { select: false })` -> `renderTabs()`.
3. The toggle is now visible while a channel tab is active.

I reproduced this against the branch with a throwaway case in `tests/chat_window_controller.test.ts`; after step 2 the strip is:

```
["all:chat-tab","combat:chat-tab","world:chat-tab","lfg:chat-tab",
 "undefined:chat-tab chat-tab-sysfilter","undefined:chat-tab chat-tab-add"]
```

no `chat-tab-sysfilter-hidden`. The same re-render also drops the `.active` / `aria-selected` / `tabIndex` state from the active tab, but that part is pre-existing, not yours. One fix covers both: end `renderTabs()` with `this.updateActiveTabStyles()` (it is idempotent, and `selectTab` already calls it right after on the init path). Worth a test case for the re-render arm, since the existing hide test only exercises the `selectTab` path.

Nothing else blocking: no dashes, emoji, secrets, or `ALLOW_DEV_COMMANDS` touches. Happy to approve once the re-render arm is covered.

daxdax89:
Thanks for the fast re-review, and for confirming the mobile mirror and the aria/tabIndex handling.

On the re-render finding: I added the test case you asked for (`keeps the toggle hidden through a tab-strip re-render on a channel tab` in `tests/chat_window_controller.test.ts`), which drives exactly your repro: init on `world`, then `syncTabsForInput('/join lfg')` -> `addTab('lfg', { select: false })` -> `renderTabs()`. It passes as-is, because `renderTabs()` already ends with `this.updateActiveTabStyles()` (its last statement), and that call re-applies `chat-tab-sysfilter-hidden` since `activeChatTab` is still `world` after the typed join. So on this head the toggle stays hidden across the rebuild rather than resurfacing.

I think the "renderTabs() does not call it" read was against a slightly different head; the current tip (`f826b00a9`) has the call, and I confirmed the strip after step 2 carries `chat-tab-sysfilter-hidden` on the toggle. The new test pins that so a future edit dropping the `updateActiveTabStyles()` call fails here instead of regressing the behavior. (You are right that the same re-render drops `.active`/`aria-selected`/`tabIndex` from the active tab, but as you noted that is pre-existing and outside this PR's scope.)

Branch is rebased onto current `release/v0.28.0`; `tests/chat_window_controller.test.ts` is 10 green. Ready for another look.

EnriqueGF (approved):
Revisado el diff completo contra los invariantes del repo y ejecutado en local. Buen trabajo: el cambio es pequeño, la regla se extrae a un predicado puro (`chatLineVisible`) reutilizado por los dos consumidores en vez de duplicar la condicion, y va acompanado de tests reales. Aprobado.

## Que revise

- `src/ui/hud/chat/chat_channels.ts` y `chat_window_controller.ts` (predicado, persistencia, render del toggle, `updateActiveTabStyles`, `applyFilter` / `hideIfFiltered`).
- Los dos bloques CSS y el espejo en `@layer hud-mobile`.
- La clave i18n y las tablas generadas.
- Trace completo de `dataset.chan`: los unicos dos sitios que lo asignan son `appendLog` (hud.ts:10415) y `chatLogFrom` (hud.ts:9873), ambos llaman a `hideIfFiltered` justo despues, asi que ninguna linea se salta el filtro. El panel de combate es otro elemento y queda correctamente fuera de alcance.

## Que ejecute

- `npx tsc --noEmit`: limpio.
- `npx vitest run tests/chat_channels.test.ts tests/chat_window_controller.test.ts tests/chat_tab_strip.test.ts tests/chat_filter.test.ts tests/architecture.test.ts tests/css_corpus.test.ts tests/world_api_parity.test.ts`: 366 pasando.
- `npm run i18n:gen` idempotente (arbol limpio despues), y `tests/localization_fixes.test.ts` + `tests/i18n_completeness.test.ts` en verde.
- Biome limpio sobre los ficheros tocados.
- Sin `Math.random` / `Date.now`, sin tocar `src/sim/`, sin em/en dashes ni emojis, `ALLOW_DEV_COMMANDS` intacto.

## Hallazgo (medio, no bloqueante)

El filtro esconde exactamente `chan === 'system'`, y ese canal es mas amplio de lo que sugiere el cuerpo de la PR ("loot, XP, zone welcome, quest/system notices"). `HUD.log()` (hud.ts:9811) codifica `'system'` de forma fija, y por ahi pasan tambien:

- los errores espejados en el chat desde `showError` (hud.ts:10477),
- los avisos de sim `log` que no son sabor de combate (por ejemplo el aviso de tamano de grupo para mazmorras),
- las notificaciones de moderacion / ignore / jail.

Es decir, con "Chat Only" activo un jugador deja de ver en el log los errores y avisos de moderacion, no solo el ruido de loot y XP. Lo considero aceptable para mergear: es opt-in, esta apagado por defecto, y los errores siguen apareciendo como toast en pantalla. Pero conviene decidirlo explicitamente en vez de heredarlo:

- opcion minima: mencionarlo en el comentario de `chatLineVisible` y en el `title` del boton, para que la siguiente persona no lo descubra por sorpresa;
- opcion mejor: cuando llegue el tagging mas fino de `chan` que menciona el follow-up de #1670, sacar errores y moderacion a su propia categoria siempre visible.

## Notas menores (sin accion requerida)

- El toggle es un `button` con `aria-pressed` dentro de un `role="tablist"`, lo que estrictamente no es conforme a ARIA, pero es exactamente el mismo patron que ya usa `.chat-tab-add`, asi que es coherente y no es una regresion. Bien resuelto que quede fuera del bucle de `aria-selected` / `tabIndex` y que conserve su propia alcanzabilidad por teclado (y el pin de test correspondiente).
- `filterTab()` devuelve `null` tanto para `all` como para `combat`, asi que el estado tambien aplica en la vista de combate; como esa vista usa otro elemento, no tiene efecto visible. Correcto tal cual.

daxdax89:
Thank you for the approval and the careful `dataset.chan` trace.

On the non-blocking finding (the `system` channel is broader than the PR body's "loot, XP, zone, quest" wording, since `HUD.log()` also routes `showError` mirrors, non-combat sim warnings, and moderation/ignore/jail notices through `chan === 'system'`): making the decision explicit, as you asked.

Decision: keep this PR's all-or-nothing `system` switch as-is for now. It is opt-in, off by default, and errors still surface as on-screen toasts even when the log line is hidden, so no actionable information is lost silently. Pulling errors and moderation out into their own always-visible category is the right end state, but it depends on the finer `chan` tagging in `appendLog` that #1670 itself calls out as a prerequisite, so it belongs in the follow-up that #1670 tracks (this PR is "Part of #1670", not a close). I will carry the "errors/moderation stay visible" requirement into that follow-up so it is not lost.

I have left the branch at the approved head rather than pushing a doc-only tweak, to avoid dismissing your approval; happy to add the clarifying note to the `chatLineVisible` comment in the follow-up. Ready to merge whenever you are.

PR #1996, opened 2026-07-16
fix(bot): reflect live Discord role changes and resolve duplicate special-role ids (#1986)

### LINKED ISSUE #1986: Admin Discord flair is not shown in game for members holding the Admins role
## Problem

Members holding the Discord role `Admins` do not get the admin flair (green name tint plus the `[Admin]` tag) in game.

The role catalog is not the cause. `src/sim/discord_roles.ts` already lists `Admins` as an alias of the `admin` key, matching is case-insensitive, and `tests/discord_roles.test.ts` pins `specialRoleByName('Admins')?.key === 'admin'`. Adding the alias, the obvious first fix, would be a no-op.

**The bot never observes Discord role changes.** `bot/main.ts` seeds `memberRoles` once, in `seedGuild` off `GUILD_CREATE` (`memberRoles.set(id, asStringArray(m.roles))`). The gateway dispatch switch handles `GUILD_CREATE`, `VOICE_STATE_UPDATE`, `PRESENCE_UPDATE`, `GUILD_MEMBER_ADD`, `MESSAGE_CREATE`, and `INTERACTION_CREATE`. There is **no `GUILD_MEMBER_UPDATE` handler**, and `GUILD_MEMBER_ADD` sets `memberNames` but never `memberRoles`. The only other writers are the bot's own tier-role add/remove bookkeeping, which touch only ids the bot itself grants.

So `memberRoles` is frozen at connect. Granting someone the `Admins` role while the bot is running is invisible to it, and the 5 minute `ROLE_SYNC_INTERVAL_MS` loop re-pushes the same stale snapshot forever. This fails in both directions: a newly promoted admin never gets flair, and a demoted member keeps their color until the next restart.

**The initial seed is also partial.** `identifyPayload` (`bot/logic.ts`) sends `token`, `intents`, and `properties`, but never sets `large_threshold`, so it defaults to 50. For a guild above that size Discord omits offline members from `GUILD_CREATE.members`. `GATEWAY_OP` has no op 8 (`REQUEST_GUILD_MEMBERS`) and the bot never chunks, so the omitted members are never backfilled. The tell is already in the code: `memberTotal` comes from `d.member_count` while the seed loop walks the much shorter `d.members`.

The effective predicate for getting flair today is "was online in Discord at the moment the bot last connected, already held the role at that instant, and has linked Discord in game." That explains the report without any duplicate-role theory.

## Steps to reproduce

1. With the bot running, grant a linked member the `Admins` role in Discord.
2. Wait past the 5 minute role sync interval.
3. Have the member log in and check their nameplate, target frame, and inspect card.
4. Observe no tag and no name tint.
5. Restart the bot **while that member is online in Discord**. The flair now appears.

Step 5 is the discriminating test: if a restart with the member online fixes it, the observation gap is confirmed and the duplicate-role collision below is not the trigger.

## Actual behavior

- The member has the default name color and no role tag.
- The bot pushes `role: null` for them, and `discord_links.discord_role` is written `NULL`.
- Members offline at the bot's last connect are absent from `memberRoles` entirely, so no push ever targets their id.
- A member who leaves the guild keeps their flair permanently: after a restart they are absent from `memberRoles`, so the server's clearing path never fires for them.

## Expected behavior

- Granting or revoking a special role in Discord is reflected in game within one sync interval, with no bot restart.
- Flair does not depend on whether the member happened to be online when the bot connected.
- A member holding any name that resolves to a special role key gets that flair, regardless of how many other guild roles resolve to the same key.
- When a member holds several special roles, the highest `priority` in `DISCORD_SPECIAL_ROLES` wins.
- Resolution is stable across restarts and independent of the order the Discord API returns roles in.

## Scope

- `bot/main.ts`: add a `GUILD_MEMBER_UPDATE` case that refreshes `memberRoles` (and `GUILD_MEMBER_REMOVE` to clear), and have `GUILD_MEMBER_ADD` seed `memberRoles` from `d.roles` rather than only the name.
- `bot/logic.ts`: set `large_threshold` on `identifyPayload`, and add op 8 (`REQUEST_GUILD_MEMBERS`) to `GATEWAY_OP` plus a chunk request after `GUILD_CREATE` so the full roster is seeded. `GUILD_MEMBERS` is already declared in `GATEWAY_INTENTS`, so no new privileged intent is needed.
- `bot/main.ts`: `refreshSpecialRoles` keeps only the **first** guild role id per key (`if (def && !specialRoleIds.has(def.key))`), while `admin` accepts four names. If the guild has both `Admin` and `Admins`, one id is discarded and its holders resolve to `null`. Secondary to the above but a real bug on the same path: the map should be key to a **set** of ids. Note the catalog records that the guild's `Admin` role is the renamed `Mods` role, which is exactly the setup that produces a duplicate.
- `bot/logic.ts`: the resolution logic belongs here as a pure function, per the bot's own pure/IO split (`bot/CLAUDE.md` warns against inlining role-diff logic into `main.ts`/`gateway.ts`). That is what makes it testable.
- `tests/discord_bot.test.ts`: coverage for role-change observation and the multi-role-per-key case.

## Acceptance criteria

- Granting `Admins` to a linked member while the bot runs produces the flair within one sync interval, with no restart.
- Revoking a special role clears the flair within one sync interval.
- A member who was offline at connect still receives flair once the roster is seeded.
- A guild containing both `Admin` and `Admins` grants the admin flair to holders of either, verified by a test seeding two role ids on one key.
- Tests cover the bot's id-first resolution path (`refreshSpecialRoles` plus `topSpecialRoleKey`), which today has none.
- Priority ordering still holds when a member's roles span several keys.
- No change to the wire format, the `discord_role` storage token, or the rendering path.

## Notes

The render half is working and is not in scope. Path for reference: `bot/main.ts` `refreshSpecialRoles` to `pushAllMemberMeta` to `bot/server_client.ts` `pushMembersMeta` to `server/internal.ts` to `server/discord_db.ts` `setDiscordMemberMeta` (writes `discord_links.discord_role`) to `server/game.ts` `refreshDiscordFlair` (sets `e.discordRole`) to wire key `dr` to `src/net/online.ts` to `src/render/nameplate_painter.ts` plus `src/ui/hud.ts`.

The `// first match wins per key, mirroring Discord's top-down role order` comment in `refreshSpecialRoles` is a false premise: `bot/discord_api.ts` `guildRoles` returns the raw array from `GET /guilds/{id}/roles` with no sort, and Discord does not guarantee position order on that endpoint. Which duplicate wins is arbitrary and can flip across restarts. The comment should go with the fix.

`src/sim/discord_roles.ts` exports `topSpecialRole(roleNames)`, which handles aliases and priority correctly, but it has **zero production consumers**: only the test file imports it. The bot reimplemented ranking id-first in `topSpecialRoleKey` because the gateway gives role ids rather than names, so the existing tests pin a function production never calls. Worth reconciling the two resolvers rather than keeping a second implementation.

Adjacent issues on the same path, worth folding in or splitting out:

- `pushAllMemberMeta` truncates with `[...memberRoles.entries()].slice(0, 1000)`. Latent while the map is small, but it becomes load-bearing the moment the roster is seeded properly, so it should be fixed in the same change as the chunking.
- The link upsert in `server/discord_db.ts` is `ON CONFLICT (account_id) DO UPDATE` over `discord_user_id`, `discord_username`, `discord_avatar`, `discord_email`, `guild_member`, and `linked_at`, and never touches `discord_role`. Repointing an account at a different Discord id carries the old role key over.
- `setDiscordMemberMeta` is an `UPDATE ... WHERE discord_user_id = $1`, a silent no-op unless the account has linked Discord in game. A real admin who never linked gets no flair by design, which is worth confirming is intended.

For clarity on a related question this surfaced: in-game admin authority is unrelated to this path. It comes from `accounts.is_admin` and `accounts.admin_roles` in Postgres (`server/db.ts`, `server/staff_db.ts`), seeded at WS join. `session.isAdmin` never reaches the wire. Discord flair is cosmetic only and does not grant anything, which is the correct split. This issue is only about the cosmetic side.

### LINKED ISSUE #2028: Discord flair: reconcile departed members at startup + reset discord_role on relink (follow-up to #1986)
Follow-up to #1986 / #1996. During review of #1996, @FernandoX7 identified two residual Discord-flair reconciliation gaps that were out of scope for that PR (which handled live role tracking, duplicate special-role ids, and the offline-member backfill). Filing so they are not lost when #1986 closes.

### 1. A member who leaves while the bot is down keeps stale in-game flair

`GUILD_MEMBER_REMOVE` now clears a departing member's flair while the bot is running, but nothing reconciles stored roles against the freshly backfilled roster at startup. If a member leaves (or loses a staff role) during a bot outage, the `GUILD_MEMBER_REMOVE`/`GUILD_MEMBER_UPDATE` event is missed, and on the next boot the `GUILD_CREATE` + op-8 backfill only ADD/refresh present members; it never diffs against what the server already stored, so a departed member's `guild_member` flag and `discord_role` stay set.

Fix direction: at startup, after the roster is fully backfilled, reconcile the server's stored Discord membership set against the live roster and clear anyone no longer present (a bulk `guildMember: false` + cleared meta for the diff).

### 2. Relinking an account to a different Discord id carries the old flair key over

The link upsert in `server/discord_db.ts` never writes `discord_role`, so if an account relinks to a different Discord id, the previously stored `discord_role` key is retained rather than reset to reflect the new identity's roles.

Fix direction: clear (or re-resolve) `discord_role` on the link upsert path so a relink starts from a clean flair state.

Neither is a regression from #1996; both predate it. Low urgency (they only surface on a bot-downtime departure or a relink), but worth tracking.

daxdax89:
## Summary

Admin Discord flair was invisible in game for members holding the Admins role. The render and catalog path (`src/sim/discord_roles.ts`, `src/ui/discord_role_tag.ts`, `src/render/nameplate_painter.ts`, the `dr` wire key) is correct and is left untouched; all three defects were in the Discord bot process and are fixed here without any change to the wire format, the `discord_role` storage token, or the client render path.

1. No live role tracking: `memberRoles` was seeded once from `GUILD_CREATE` and `GUILD_MEMBER_ADD` never set roles, so a role granted after startup stayed invisible until a bot restart. Now `GUILD_MEMBER_UPDATE` replaces the cached role set and re-pushes that member's meta, `GUILD_MEMBER_REMOVE` clears server-side membership and flair before dropping the member, and `GUILD_MEMBER_ADD` seeds roles and join date.
2. Duplicate special-role ids collapsed: `refreshSpecialRoles` kept only the first role id per key, so a guild with both an `Admin` and an `Admins` role dropped one and its holders resolved to no flair. All matching role ids are now indexed to their key, so a holder of either resolves.
3. Offline members omitted: identify sent no `large_threshold` and the bot never chunked, so offline staff were missing in large guilds. Identify now sends `large_threshold` and the bot requests the full member list via `REQUEST_GUILD_MEMBERS` (op 8), consuming `GUILD_MEMBERS_CHUNK`, and batches the resulting member-meta pushes at the server's 1000-per-request cap so no member past the first 1000 is dropped.

Per `bot/CLAUDE.md`, the reconciliation is pure and unit-tested in `bot/logic.ts` (`reconcileMemberRolesFromUpdate`, `indexSpecialRoleIds`, `topSpecialRoleKeyFor`, `memberRolesFromPayload`, `requestGuildMembersPayload`, `chunk`, `clearedMemberMeta`); the IO stays in `bot/main.ts` and `bot/gateway.ts`.

## Related issues

Closes #1986
Closes #2028 (both follow-up items landed in the maintainer batch on this branch)

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/discord_bot.test.ts tests/discord_roles.test.ts`: 55 passed
  - `npx tsc --noEmit` clean for `bot/` and the test
  - `npx @biomejs/biome check` clean on the changed files
  - `npx esbuild bot/main.ts --bundle` builds
  - Codex review against `release/v0.27.0`: two findings (large-guild backfill batching, clearing server state on member leave), both fixed
- Manual steps: N/A (bot gateway logic, covered by the unit tests below).

New/updated assertions, each written to fail against the pre-fix behavior and pass after:
- `GUILD_MEMBER_UPDATE` reconciliation makes `admin` resolve where it did not before a grant.
- A key with two role ids (`Admin` + `Admins`) resolves holders of either id (the old first-wins map dropped one).
- Identify carries `large_threshold` and the op-8 payload shape is exact (`{op:8, d:{guild_id, query:'', limit:0, presences:true}}`).
- A 2500-member backfill batches into `[1000, 1000, 500]` with the tail covered (nothing dropped past 1000).
- `GUILD_MEMBER_REMOVE` emits a clearing update (`guildMember:false` plus a `role:null` meta record) before the cache delete.

## Screenshots / recordings

N/A. Server-side Discord bot logic; no player-facing UI added.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the bot emits no player text; the flair tag labels already exist).

### Hygiene

- [x] No secrets or `.env`; no protocol, schema, or `discord_role` token change.

### DIFF
--- a/bot/CLAUDE.md
+++ b/bot/CLAUDE.md
@@ -84,8 +84,14 @@ is posted). Boot loads `.env`/`.env.local` when present but runs fine from ambie
 env alone (`process.loadEnvFile`).
 
 ## Limits / notes
-- Guild state is seeded from `GUILD_CREATE`; very large guilds may not send all
-  members up front (request-guild-members opcode is not implemented). Fine for
-  small/medium servers.
+- Guild state is seeded from `GUILD_CREATE` and then kept live: `GUILD_MEMBER_ADD`
+  seeds a joiner's roles/join date, `GUILD_MEMBER_UPDATE` reconciles a member's
+  role set (so a role granted or revoked after boot reflects on the next push), and
+  `GUILD_MEMBER_REMOVE` clears their stored flair. Large guilds (over
+  `large_threshold`, sent in IDENTIFY) omit offline members from `GUILD_CREATE`, so
+  the bot requests the full roster with the request-guild-members opcode (op 8) and
+  consumes the `GUILD_MEMBERS_CHUNK` stream. Member-meta pushes to the server are
+  batched at its 1000-per-request cap (`MEMBERS_META_BATCH`, mirroring
+  `server/internal.ts`).
 - "Speaking" indicators are not live (that needs a voice-gateway connection); the
   voice list shows membership + self-mute.
--- a/bot/CLAUDE.md
+++ b/bot/CLAUDE.md
@@ -25,8 +25,9 @@ Zero new dependencies: Gateway over the existing `ws`, REST via built-in `fetch`
 - `server_client.ts`: client for the game server's secret-gated `/internal/discord/*`
   endpoints (`x-woc-discord-secret`); grep `/internal/discord/` there for the live set.
 - `config.ts`: env to `BotConfig` (throws on missing required).
-- `main.ts`: wiring only: guild state seeded from `GUILD_CREATE`, event dispatch,
-  the poll loops.
+- `main.ts`: wiring only: guild state seeded from `GUILD_CREATE` (plus the op 8
+  member backfill for large guilds), kept live by the `GUILD_MEMBER_*` events,
+  event dispatch, the poll loops.
 
 ## New bot feature recipe (module-first)
 1. Pure message-builder/diff/shaping logic in `logic.ts`, with a test in
@@ -69,8 +70,11 @@ Zero new dependencies: Gateway over the existing `ws`, REST via built-in `fetch`
 - **Staff/special roles** (e.g. Levy St, Core Dev, Mods) live in the shared catalog
   `src/sim/discord_roles.ts`, matched by exact name or alias (case-insensitive);
   the member's top-priority role is pushed via members-meta and drives the
-  in-world name color + tag. **A guild-side rename silently breaks the match**:
-  add an alias to the catalog instead of renaming.
+  in-world name color + tag. Grants and revokes are observed live
+  (`GUILD_MEMBER_UPDATE` re-pushes that member's meta immediately), and EVERY
+  guild role id matching a catalog key is indexed, so duplicate-named roles
+  (an `Admin` and an `Admins`) both resolve. **A guild-side rename silently
+  breaks the match**: add an alias to the catalog instead of renaming.
 
 ## Env (see .env.example; the live set is `grep process.env bot/config.ts`)
 Required: `DISCORD_BOT_TOKEN`, `DISCORD_CLIENT_ID`, `DISCORD_GUILD_ID`,
@@ -87,11 +91,14 @@ env alone (`process.loadEnvFile`).
 - Guild state is seeded from `GUILD_CREATE` and then kept live: `GUILD_MEMBER_ADD`
   seeds a joiner's roles/join date, `GUILD_MEMBER_UPDATE` reconciles a member's
   role set (so a role granted or revoked after boot reflects on the next push), and
-  `GUILD_MEMBER_REMOVE` clears their stored flair. Large guilds (over
-  `large_threshold`, sent in IDENTIFY) omit offline members from `GUILD_CREATE`, so
-  the bot requests the full roster with the request-guild-members opcode (op 8) and
-  consumes the `GUILD_MEMBERS_CHUNK` stream. Member-meta pushes to the server are
-  batched at its 1000-per-request cap (`MEMBERS_META_BATCH`, mirroring
-  `server/internal.ts`).
+  `GUILD_MEMBER_REMOVE` clears their stored flair. Guilds above the IDENTIFY
+  `large_threshold` (250, the gateway max) omit offline members from
+  `GUILD_CREATE`, so the bot backfills the full roster with
+  `REQUEST_GUILD_MEMBERS` (op 8, streamed back as `GUILD_MEMBERS_CHUNK`). After
+  every COMPLETE seed it also reconciles stored flair against the roster
+  (`/internal/discord/flaired-ids`), clearing members who left while the bot was
+  offline. Member-meta pushes are batched by BYTES (`MEMBERS_META_BATCH`), sized
+  so a worst-case batch stays under the server's 64 KiB JSON body cap; the
+  server's 1000-entry slice is defense in depth, never the binding constraint.
 - "Speaking" indicators are not live (that needs a voice-gateway connection); the
   voice list shows membership + self-mute.
--- a/bot/gateway.ts
+++ b/bot/gateway.ts
@@ -3,7 +3,13 @@
 // forwards DISPATCH events to a handler. The protocol opcode/payload logic lives
 // in the pure ./logic module so it is unit-tested; this file is the IO shell.
 import { WebSocket } from 'ws';
-import { GATEWAY_OP, heartbeatIntervalMs, identifyPayload, resumePayload } from './logic';
+import {
+  GATEWAY_OP,
+  heartbeatIntervalMs,
+  identifyPayload,
+  requestGuildMembersPayload,
+  resumePayload,
+} from './logic';
 
 export interface GatewayHandlers {
   onDispatch(type: string, data: Record<string, unknown>): void;
@@ -27,6 +33,16 @@ export class Gateway {
     private gatewayUrl: string,
     private handlers: GatewayHandlers,
   ) {}
+
+  /**
+   * Request a guild's full member list (op 8). Discord streams the members back
+   * as GUILD_MEMBERS_CHUNK dispatches, so offline members (omitted from
+   * GUILD_CREATE for large guilds) are delivered. Safe to call once the socket is
+   * open (after IDENTIFY / GUILD_CREATE); a no-op if the socket is not open yet.
+   */
+  requestGuildMembers(guildId: string): void {
+    this.send(requestGuildMembersPayload(guildId));
+  }
 
   connect(resume = false): void {
     this.resuming = resume && this.sessionId !== null;
--- a/bot/logic.ts
+++ b/bot/logic.ts
@@ -284,6 +284,45 @@ export function staleFlairedIds(
   return flaggedIds.filter((id) => !rosterIds.has(id));
 }
 
+/**
+ * Clear the server-side flair state for departed members: push a null-role
+ * members-meta record for each (batched under the same cap as the roster
+ * push), THEN drop each member's guild_member flag. IO arrives as injected
+ * callbacks so the ordering, batching, and the re-observed-member skip are
+ * unit-testable without a network: `isMember` is re-checked before EVERY write
+ * because a member can rejoin (GUILD_MEMBER_ADD) between the roster diff and
+ * these awaits, and the live event handlers own a present member's state.
+ * Returns the ids whose membership flag was cleared.
+ */
+export async function clearDepartedFlair(
+  staleIds: readonly string[],
+  isMember: (id: string) => boolean,
+  io: {
+    pushMembersMeta: (
+      records: {
+        discord_user_id: string;
+        name: string | null;
+        joinedAtMs: number | null;
+        role: string | null;
+      }[],
+    ) => Promise<unknown>;
+    setMember: (id: string, guildMember: boolean) => Promise<unknown>;
+  },
+  batchSize: number = MEMBERS_META_BATCH,
+): Promise<string[]> {
+  for (const batch of chunk(staleIds, batchSize)) {
+    const records = batch.filter((id) => !isMember(id)).map(clearedMemberMeta);
+    if (records.length) await io.pushMembersMeta(records);
+  }
+  const cleared: string[] = [];
+  for (const id of staleIds) {
+    if (isMember(id)) continue;
+    await io.setMember(id, false);
+    cleared.push(id);
+  }
+  return cleared;
+}
+
 // ── Level-on-name (Discord nickname) ─────────────────────────────────────────
 // A class "icon" + the in-game level attached to the member's Discord name, so a
 // linked player's level shows next to their name in the server (e.g. "Aldric ⚔20").
--- a/bot/logic.ts
+++ b/bot/logic.ts
@@ -3,6 +3,7 @@
 // and voice-presence shaping. Kept separate from the ws/fetch IO (gateway.ts,
 // discord_api.ts, server_client.ts) so it is unit-tested without a network. This
 // is the same pure/IO split the server uses (wallet_link.ts vs wallet.ts).
+import { specialRoleByKey, specialRoleByName } from '../src/sim/discord_roles';
 import { DISCORD_STATUS_DEFS, discordStatusByIndex } from '../src/sim/discord_tier';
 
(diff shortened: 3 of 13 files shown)


FernandoX7 (commented):
Reviewed at f02b713 against `release/v0.27.0`. This is a careful fix. All three root causes are real in the base (no `GUILD_MEMBER_UPDATE` handler and `GUILD_MEMBER_ADD` never seeding roles; the first-wins `specialRoleIds` map; no `large_threshold` and no op 8), and the load-bearing claims in the description check out against the actual code:

- The server endpoint really does cap each request at 1000 (`server/internal.ts:511`, `body.members.slice(0, 1000)`), so the `MEMBERS_META_BATCH` batching is the right shape and the old single-slice push was silently dropping every member past the cap.
- `setDiscordMemberMeta` (`server/discord_db.ts:576`) writes `discord_role = $4` unconditionally while COALESCE-ing name and join date, so `clearedMemberMeta` clears exactly the flair and re-asserts nothing, as claimed.
- `GATEWAY_INTENTS` already carries `GUILD_MEMBERS` and `GUILD_PRESENCES`, so the new dispatches and `presences: true` on op 8 need no new privileged intent.
- The priorities the tests pin match the catalog (`Levy St` 11 over `admin` 10; `Admins` is an `admin` alias).
- Gateway `send` no-ops unless the socket is open, and `ServerClient.call` catches everything and returns null, so the un-caught `void server.setMember(...)` calls cannot become unhandled rejections.
- 55 tests pass at the head, `npx tsc --noEmit` is clean, and the branch merges clean into `release/v0.27.0`.

The pure/IO split follows `bot/CLAUDE.md`, and the assertions are decisive (the op 8 payload is pinned exactly; the 2500-member batching test covers the tail).

On the tracking issue: this matches the diagnosis in #1986 point for point and meets its acceptance criteria (live grant and revoke via `GUILD_MEMBER_UPDATE` with an immediate single-member push, offline members via the op 8 backfill, `Admin` plus `Admins` both resolving, priority ordering, the id-first resolution path now tested, no wire or `discord_role` token change). The `Closes #1986` link is correct.

Findings:

- should-fix: `bot/CLAUDE.md:88` still says the request-guild-members opcode is not implemented and that large guilds may not send all members up front. This PR makes that false; update the Limits note (and the GUILD_CREATE-only seeding line next to it) in the same change.
- nit: `MEMBERS_META_BATCH` (`bot/logic.ts:225`) mirrors the server's `slice(0, 1000)` by value only, and `tests/server/internal.test.ts` pins the field sanitization but not the cap. A one-line server test pinning the 1000-per-request cap would keep the pair from drifting back into silent tail-dropping.
- nit (follow-up, not this PR): two residual cases from #1986 stay open. A member who leaves while the bot is down keeps stale flair, since nothing reconciles stored roles against the backfilled roster at startup; and the link upsert in `server/discord_db.ts` still never writes `discord_role`, so relinking an account to a different Discord id carries the old key over. The issue lists these under its adjacent items, so a follow-up issue is the right home; worth filing so they are not lost when #1986 closes.

daxdax89:
Thanks for the thorough review, and for tracing every claim back to the code. Both in-PR findings are addressed in `2534f1588` (amended into the same commit, rebased onto current `release/v0.27.0`):

- **should-fix (bot/CLAUDE.md):** rewrote the Limits note. It now says guild state is seeded from `GUILD_CREATE` and then kept live via `GUILD_MEMBER_ADD`/`GUILD_MEMBER_UPDATE`/`GUILD_MEMBER_REMOVE`, that large guilds are backfilled with the request-guild-members opcode (op 8, `large_threshold` in IDENTIFY) consuming `GUILD_MEMBERS_CHUNK`, and that member-meta pushes are batched at the server's 1000-per-request cap. The stale "opcode is not implemented" / GUILD_CREATE-only line is gone.
- **nit (server cap test):** added a case to `tests/server/internal.test.ts` `discord/members-meta` that posts 1001 members and asserts `updated: 1000` and exactly 1000 `setDiscordMemberMeta` calls, pinning the `slice(0, 1000)` cap that `MEMBERS_META_BATCH` mirrors so the pair cannot drift back into silent tail-dropping.

59 tests pass, `tsc --noEmit` clean, no wire or `discord_role` token change.

On the follow-up nit: agreed both are real and out of scope here. A member who leaves while the bot is down keeps stale flair (nothing reconciles stored roles against the backfilled roster at startup), and the link upsert in `server/discord_db.ts` never writes `discord_role`, so relinking to a different Discord id carries the old key over. I will file a follow-up issue capturing both so they are not lost when #1986 closes.

FernandoX7:
Pushed a follow-up batch (five commits, replayed on top of your rebase, so the branch stays on the latest `release/v0.27.0`). It addresses everything from my review, both items in #2028, and one new defect the work surfaced. Where your rebase and my batch overlapped (the CLAUDE.md Limits note and the members-meta cap test) I merged the two versions rather than dropping yours.

1. A real bug the cap test exposed: the 1000-per-request batching was silently ineffective. A full batch of realistic records (snowflake ids, names, join dates) serializes to roughly 100 KB, past the server's 64 KiB `readBody` cap, and the members-meta handler coerces an over-cap body to an EMPTY member list and answers `200 {updated: 0}`: the whole batch drops with no error, exactly in the large-guild case the batching was built for. `MEMBERS_META_BATCH` is now byte-derived (200; a worst-case record is about 300 bytes), a test pins a worst-case full batch under `DEFAULT_JSON_BODY_MAX_BYTES`, the server test pins the 1000-entry slice at or above the bot batch, and `ServerClient.pushMembersMeta` logs the zero-row silent-drop signature.

2. #2028 item 2 (relink carry-over): the `discord_links` upsert now resets `discord_role` and `discord_joined_at` when the link repoints at a DIFFERENT Discord id; a same-id relink keeps them and the bot re-pushes current values within one sync interval. Pinned in `tests/discord_db.test.ts`.

3. #2028 item 1 (leave while the bot is down): after every COMPLETE roster seed (a small guild's `GUILD_CREATE`, or the final op 8 chunk) the bot fetches the ids the server still has flagged (new secret-gated read-only `GET /internal/discord/flaired-ids`, added to BOTH dispatch arms per the dual-edit rule) and clears exactly the flagged ids missing from the live roster through the existing member + members-meta endpoints. The clears deliberately reuse the audited write paths and the new endpoint is a pure read, so a truncated or failed request can never mass-clear a live member; the `rosterComplete` gate keeps a partial seed from misreading unseeded members as departed. The orchestration is `clearDepartedFlair` in `bot/logic.ts` behind injected IO, with tests pinning the global meta-before-flag order, the batching, and that membership is re-read across the await boundary (mutation-tested: a snapshot-once rewrite fails the suite).

4. `bot/CLAUDE.md` refreshed (merged with your Limits rewrite; the batching sentence now describes the byte sizing) and the characterization spine extended for the new route: surface-inventory row, completeness and gate-sweep counts, plus a legacy-vs-RouteDef body parity pin, since db-touching internal routes are excluded from the parity corpus replay.

Verification: the full gate under Node 24 (i18n freshness, malware scan, changed-files biome, SFX conformance, 14,584 unit tests, tsc, all five builds), then a full re-run after replaying onto your rebased head (14,613 tests passing, tsc and biome clean). The only local failure is a known environmental armory browser-layout flake on my machine, unrelated to this diff; CI is the arbiter there. An adversarial multi-agent review pass over the batch surfaced two findings (a stale CLAUDE.md sentence and a test-pin gap on the reconcile's rejoin re-check), both fixed in this push.

I updated the PR description to close #2028 alongside #1986. One note for the record: an admin who never linked Discord in game still gets no flair, which is the intended design (flair requires a verified link), per the note in #1986.

PR #2000, opened 2026-07-16
feat(render): swap 10 code-generated primitive props for Tripo GLBs

jgyy:
## Summary

Replaces 9 code-generated (Three.js primitive geometry) world props that had no prior GLB
backing with Tripo-generated GLBs, matching the shipped KayKit/Quaternius art direction, via
the repo's own agent-drivable `scripts/asset_pipeline/` `prop` lane:

- `crypt_ritual_circle` altar (`src/render/quest_objects.ts`)
- `delve_module_exit` rubble archway (`src/render/delve_props.ts`)
- `delve_surface_exit` stairwell (`src/render/delve_props.ts`)
- `delve_rite_shrine_bell` / `_candle` / `_reed` / `_skull` (4 kinds, `src/render/delve_props.ts`)
- `delve_pressure_plate` base slab (`src/render/delve_props.ts`)
- marsh `root_wall` dressing piece (`src/render/delve_marsh_dressing.ts`)

(A 10th target, `vale_cup_ball`, shipped in an earlier revision of this PR and was reverted:
the Tripo GLB did not read as the Vale Cup boarball in game, so that prop stays on its
original procedural sphere. See the revert commit on this branch for detail.)

Every swap keeps the existing procedural builder as the load-race/headless fallback, the same
contract already used by `mailbox.ts`, `fish.ts`, and `door_portal.ts`:

```mermaid
flowchart LR
    A[Module preload] -->|loadGltf url| B{GLB loaded?}
    B -->|yes| C[Clone + normalize GLB]
    B -->|no: load race or headless host| D[Procedural Three.js fallback]
    C --> E[Rendered prop]
    D --> E
```

The pressure plate is the one stateful case: the GLB replaces only the static base+rim, and
the existing procedural triggered-state glow/rune overlay stays layered on top either way, so
the "triggered" gameplay cue a player reacts to is never lost to the swap.

```mermaid
flowchart LR
    A[buildPressurePlate] --> B{GLB loaded?}
    B -->|yes| C[GLB base+rim]
    B -->|no| D[Procedural base+rim]
    C --> E{triggered?}
    D --> E
    E -->|yes| F[Procedural glow/rune overlay + light]
    E -->|no| G[Rendered prop]
    F --> G
```

One target from the original candidate list (a marsh "broken bell frame") turned out to
already have a shipped GLB (`marsh_bell_gallows.glb`) predating this PR — caught during
implementation and swapped for the marsh `root_wall` dressing piece instead, which genuinely
had no prior art asset.

## Screenshots

Real in-game captures (`scripts/proc_glb_prop_swap_shot.mjs`), not pipeline preview renders:
the offline client, driven through a live delve run (The Drowned Litany) and the
`crypt_ritual_circle` overworld quest object, screenshotting the actual GLB now shipping in
`public/models/` in its actual placement. Every frame was verified free of the intro/logo
fade, the new-adventurer tutorial popup, and the camera-mode-choice prompt before capture.

`delve_surface_exit` only spawns after solving the delve's shrine-sequence puzzle (a seeded
multi-step minigame gated behind killing the delve boss), which was out of scope for this
capture pass; it keeps the same GLB-swap code path as the other delve interactables above and
is exercised by the existing `tests/delve_render.test.ts` / `tests/render_glb_replacement_assets.test.ts`
coverage.

- `docs/screenshots/procedural-to-glb-prop-swap/after-crypt_ritual_circle.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_module_exit.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_pressure_plate.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_rite_shrine_bell.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_rite_shrine_candle.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_rite_shrine_reed.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-delve_rite_shrine_skull.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-marsh_root_wall.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-marsh_root_wall-area2.png`
- `docs/screenshots/procedural-to-glb-prop-swap/after-marsh_root_wall-area3.png`

crypt_ritual_circle
delve_module_exit
delve_pressure_plate
delve_rite_shrine_bell
delve_rite_shrine_candle
delve_rite_shrine_reed
delve_rite_shrine_skull
marsh_root_wall
marsh_root_wall_area2
marsh_root_wall_area3

## File sizes

The 9 shipped GLBs land at 117-169 KB (WebP 512 textures, meshopt-compressed via the pipeline's
standard normalize step), consistent with `public/models/props/CLAUDE.md`'s guidance that a
small standalone decoration can run a bit above the ~70 KB category average and a hero
landmark prop (the module exit archway, the surface exit stairwell) can run larger still.

## Test plan

- [x] `npx vitest run tests/architecture.test.ts tests/asset_pipeline.test.ts` (sim purity,
      pipeline conventions)
- [x] `npx vitest run tests/delve_render.test.ts tests/render_glb_replacement_assets.test.ts`
      (new asset preload sets resolve to real, manifested files)
- [x] `npx vitest run tests/vale_cup_ball.test.ts tests/vale_cup_render.test.ts
      tests/world_api_parity.test.ts`
- [x] `npx tsc --noEmit -p .` clean on all touched files
- [x] `npm run gate`: i18n freshness, malware scan, biome (changed files), SFX conformance all
      pass. The full vitest suite has 18 pre-existing failures (SFX wand-key registry drift,
      `deeds_window_focus` localStorage setup, a server characterization timeout, a mail
      equivalence test) confirmed via `git stash` to also fail identically on the unmodified
      base branch — none touch a file this PR changes.
- [x] Each of the 9 shipped assets passed the pipeline's own `qa --job <id>` structural
      gate (rig/convention checks + real Tripo cost accounting).


jgyy:
Pushed a follow-up fix: the vale_cup_ball GLB did not read correctly as the Vale Cup boarball in game, so this drops that one asset (GLB, manifest entry, CREDITS row) and reverts `src/render/vale_cup_ball.ts` to the pre-PR canvas-textured procedural sphere. The other nine prop swaps in this PR are unaffected.

EnriqueGF (commented):
Reviewed the full diff at 7b1a37b51 against release/v0.27.0. The swap mechanics are solid: every GLB keeps its procedural builder as the load-race/headless fallback on the existing mailbox/fish/door_portal contract, the new fetches go through registerPreload, the marsh root_wall reuses the placeLoadedMarshAsset seam and scale table instead of inventing a new path, the manifest was regenerated faithfully (re-running build_media_manifest is a no-op), and the vale_cup_ball revert is clean (the comment fix there is correct, the base comment was stale). Merge against the base is conflict-free.

What I ran: tsc --noEmit (clean apart from a pre-existing server/email/ses_sender.ts error untouched by this PR), tests/architecture.test.ts, tests/localization_fixes.test.ts, tests/world_api_parity.test.ts, tests/delve_render.test.ts, tests/render_glb_replacement_assets.test.ts, tests/render_asset_preload.test.ts, tests/vale_cup_ball.test.ts, tests/vale_cup_render.test.ts: all pass. Changed-files biome passes. No dash/emoji violations in the added lines.

Findings:

- should-fix, src/render/quest_objects.ts:309: the crypt_ritual_circle GLB comes out roughly 5x smaller than the prop it replaces. normalizeRoot scales by the MAX dimension, and QUEST_OBJECT_HEIGHTS pins crypt_ritual_circle at 1.65, so the 3.21-wide GLB is scaled to a 1.65-unit total footprint, while the procedural set piece it replaces (and still draws as the fallback) is an 8.2-unit-diameter circle with an altar and pillars. The 1.65 entry predates this PR but was only ever a nameplate height, the procedural template never passed through normalizeRoot. Your own after-crypt_ritual_circle.png shows it: the circle reads as a trinket at the player's feet. It also means a 5x size pop when the GLB wins the load race over the fallback. Suggest a footprint-based fit for this item (the buildStandaloneGlbFitWidth approach, target near the procedural 8.2) rather than the shared height normalize.

- should-fix, src/render/delve_props.ts:270-296: the triggered glow/rune overlay is buried inside the pressure plate GLB. The GLB fitted to PRESSURE_PLATE_WIDTH 3.2 stands 0.495 units tall (raw 0.970 x 0.150 x 0.970, no node transforms), but the glow plane sits at y=0.14 and the rune slits at y=0.135, heights tuned to the old 0.12-thick procedural slab, so both are occluded by the mesh. The triggered state stays readable through the full-mesh dark-green tint plus the point light, so this is not a fairness violation, but the PR body claim that the overlay "stays layered on top either way" does not hold on the GLB path. Seat the overlay at the fitted GLB's top (seated Box3 max.y plus a small epsilon) instead of the hardcoded procedural heights. Same builder serves sluice_valve/grave_tablet/corpse_candle, so all four families are affected.

- nit, src/render/delve_props.ts:224: the section comment still says "flush stone floor plate, low profile" but the GLB plate is now knee-high (0.495 vs the old 0.19). Worth either flattening the fit or updating the comment once the overlay fix lands.

- nit: the screenshot set is after-only; the repo workflow asks for before/after pairs for visual changes. The capture script even documents running it on both branches, so the before half looks like it was dropped.

The nine delve/marsh swaps themselves look correct: target heights match the procedural returns (moduleExit 5.5, surfaceExit 5.0 with the same PI facing, shrines 2.2 with the same entityId rotation formula), and the per-kind shrine GLBs preserve the bell/candle/reed/skull identification the rite puzzle depends on.

jgyy:
Thanks for the thorough review! Pushed a follow-up fix for both should-fix findings:

- `crypt_ritual_circle`: added a footprint-based fit (`normalizeRootByFootprint`, targeting the procedural set piece's ~8.2-unit diameter) instead of the shared height-based `normalizeRoot`, so the GLB no longer gets crushed down to the 1.65 nameplate-anchor height. That height value is left untouched for placement/nameplate purposes.
- Pressure plate triggered overlay (`buildPressurePlate`, shared by `sluice_valve`/`grave_tablet`/`corpse_candle`): the glow plane and rune slits now seat on the fitted GLB's measured top (via `Box3`) instead of the old procedural slab's hardcoded 0.14/0.135 heights, so they stay visible on top of the taller GLB mesh.

Also updated the stale "flush ... low profile" section comment and refreshed the `crypt_ritual_circle` and `delve_pressure_plate` after-screenshots to show the corrected scale.

Verified: `tsc --noEmit` clean, the targeted vitest suites (architecture/delve_render/render_glb_replacement_assets/render_asset_preload/world_api_parity) all pass, biome clean on the changed files, and confirmed the crypt ritual circle now reads at the correct size in a live in-game capture.

FernandoX7 (commented):
Reviewed the diff at e8e489a15 against release/v0.27.0, including the two follow-up fixes from the earlier review. This is a clean instance of the mailbox/fish/door_portal swap contract: every procedural builder survives verbatim as the load-race/headless fallback (extracted to buildProcedural* functions, not rewritten), the GLB arms return the same nameplate heights as their procedural arms (5.5 module exit, 5.0 surface exit, 0.4 plate, 2.2 shrines, 1.65 ritual circle), the rite shrine GLB arm reuses the same entityId-seeded rotation formula, the marsh root_wall goes through the existing placeLoadedMarshAsset seam and scale table, and the vale_cup_ball revert left no leftovers (no GLB, no manifest row, no CREDITS row, no loader code). Manifest rows are generated-consistent and the CREDITS rows match the existing Tripo format. I ran tests/render_glb_replacement_assets.test.ts and tests/delve_render.test.ts at the head: 32 of 32 pass.

Findings:

- should-fix, `src/render/quest_objects.ts:12`: crypt_ritual_circle.glb is the one shipped asset with no test pin. tests/render_glb_replacement_assets.test.ts iterates the preload-internals exports of the other render modules (delve_props, marsh dressing, critters, fish, mailbox, gather nodes, yumi maze, door_portal), but quest_objects.ts exports no such hook and nothing iterates QUEST_OBJECT_URLS, so neither the file's existence nor its manifest row is guarded for this or any quest GLB. The test plan cites that suite as covering the new assets; it covers 8 of 9. Cheap to close while you are here: export a questObjectPreloadInternalsForTest with QUEST_OBJECT_URLS and add the matching loop to the test, same shape as delvePropsPreloadInternalsForTest. (The gap predates this PR for the other quest GLBs, so treat it as a nit if you would rather leave the backfill out.)

- nit, `scripts/proc_glb_prop_swap_shot.mjs:248`: rootWallShot is set unconditionally inside the loop, so the "no candidate spots captured" fallback at line 250 can never fire. Either drop the flag or set it only on a successful shot.

Not a PR problem, for the record: the failing "Screenshots of changes" check is an infra flake. The job's git fetch was canceled mid-checkout, so scripts/pr_comment_shots.mjs was missing from an empty workspace; the script exists on both base and head. A re-run should clear it.

Two things I checked and am fine with: the triggered GLB plate replaces the textured materials with the flat dark tint plus the reseated glow/rune overlay, which keeps the triggered/untriggered distinction at least as readable as the procedural recolor did, across all four template families that share buildPressurePlate (pressure plate, sluice valve, grave tablet, corpse candle); and the plate growing from a flush 0.12 slab to a 0.495-tall fitted GLB is cosmetic only (the trigger resolves in the sim, and the state cue seats on the measured top either way).

jgyy:
Thanks for the follow-up review! Pushed a fix for both findings and rebased clean onto the current release/v0.27.0 (zero conflicts):

- should-fix: added `questObjectPreloadInternalsForTest` to `src/render/quest_objects.ts` (same shape as `delvePropsPreloadInternalsForTest`) and a matching test loop in `tests/render_glb_replacement_assets.test.ts`, so `crypt_ritual_circle.glb` (and the rest of the quest object catalog) is now guarded for existence + manifest presence, not just the 8 modules that already had the hook.
- nit: dropped the unconditional `rootWallShot` flag in `scripts/proc_glb_prop_swap_shot.mjs` since it can never be false against a fixed-length spot list.

Verified: `tsc --noEmit` clean, `render_glb_replacement_assets`/`architecture`/`delve_render`/`vale_cup_ball`/`vale_cup_render`/`world_api_parity`/`asset_pipeline` suites all green (386 tests), biome clean on the changed files, pre-push QA floor green.

FernandoX7 (approved):

PR #1584, opened 2026-07-07
WIP: Procedural infinite dungeons (Rifts) + dev commands

maxpolaczuk:
## Procedural infinite dungeons (Rifts), the new-world continent, and the AI dungeon upgrader

Draft / work in progress. cc @enriquegf for review.

### v0.28 compatibility sync (2026-07-20)

- Merged `release/v0.28.0` through `b2911d80`, including the latest offhand equip validation and drag/drop wiring.
- Reconciled the Rift branch with v0.28 simulation, server, persistence, UI, rendering, professions, item-instance, localization, audio, and parity contracts. The compatibility merge is `35babc998`; the final release-tip sync is `f9e244d7e`.
- Canonical 11-step gate is green at the pushed head: 1,448 Vitest files / 17,311 tests, 66 Chromium browser regressions, TypeScript and Svelte checks, i18n freshness, malware scan, and all production builds.
- Older release-sync notes below are retained as implementation history; the current PR base and compatibility target are v0.28.

This branch is the endgame-world epic: a seed-driven infinite dungeon system (Rifts) with ranked world portals and shared first-clear races, an optional AI service that upgrades each ranked dungeon's content and can commission new 3D assets for it, a hand-authored two-floor set-piece dungeon (the Infernal Citadel), and the new-world continent those portals open in: eleven new realms on an atlas grid, with their own looks, creatures, quests, skies, maps, and music. The branch is synchronized with `release/v0.28.0` and targets that release line.

Everything gameplay-relevant regenerates deterministically from compact descriptors (seed + baseLevel + floor), so the authoritative server and every client produce byte-identical content and only the descriptor crosses the wire, the same model as `terrainHeight(x,z,seed)`.

### The continent and the new realms

[...]

PR #2001, opened 2026-07-16
feat(trade): instance-aware multi-currency player trading (WOC + Claudium, fail-closed)

awidearray:
## Summary

Multi-currency player-to-player trading, built as an extension of the live trade
window. Design doc: docs/prd/trading-multi-currency.md (in this PR).

Four workstreams:

1. Correctness fixes to the live trade system (always on):
   - Trading a signed, rolled, or enchanted copy no longer silently strips its
     ItemInstancePayload (previously the recipient received a bare fungible copy
     and the enchant, crafter signature, and rolled quality were destroyed). Offers
     are now fungible-vs-instance aware: plain rows validate and consume plain
     stacks only (countFungibleItem), and a special copy trades as an explicit
     count-1 instance row whose payload the sim captures from the owner's real
     inventory (never from the wire), deep-equal revalidates at confirm, and
     transfers via addItemInstance intact. Character-bound copies never trade.
   - Items flagged noMarketList are now refused from trade offers, matching what
     mail and the World Market already enforce (previously about 14 lock-to-earner
     reward tokens were tradeable by oversight).
   - trade_decline (appended wire token): the invitee can actively refuse; the
     requester gets a structured tradeDeclined event. Duel and party always had
     decline; trade only had silent expiry.
   - /trade <name> slash command (mirrors /invite), help text updated.
   - Durability: every completed non-empty trade now writes an append-only
     trade_ledger row and forces both character rows saved in ONE lease-fenced
     transaction, closing the crash window in which one side's row could persist
     pre-trade state while the other persisted post-trade state (the exact failure
     class saveCharacterAndMarketState prevents for market and mail).
   - Fixed a pre-existing unmatched i18n literal ('That player is already trading.')
     and a latent catalog-merge wiring gap for hud.trade.* keys.

2. Trade window rebuilt to the house recipe (was inline in hud.ts): pure view-core
   src/ui/trade_view.ts (UI_PURE_CORES) + thin painter src/ui/trade_window.ts, with
   item tooltips on offer rows, markDialogRoot accessibility chrome, a focus trap,
   and an instance marker + payload tooltip line for signed/rolled/enchanted rows.
   All previously missing. hud.ts shrinks.

3. Claudium leg, fail-closed OFF (CLAUDIUM_TRADE_ENABLED, default unset):
   IMPORTANT: the economy service today deliberately enforces an audited one-way
   invariant (Claudium is purchasable and spendable, never player-transferable,
   never redeemable; a guard test pins it). A P2P transfer rail is exactly the
   shape of a secondary-market cash-out risk, so this PR does NOT touch the
   service and does NOT relax that invariant. It ships only the game side of the
   leg, unreachable by construction: the flag must stay unset until the
   maintainers decide the product question and the service ships an idempotent
   transfer endpoint (contract documented in the PRD: dedupe keys
   trade-<settlement>-<side> / trade-refund-<settlement>-<side>). Sim-side pledges
   are rejected with a clear error while the rail is off, and a misconfigured
   flag-on state settles to a clean full refund.

4. $WOC leg, fail-closed OFF (WOC_TRADE_ENABLED, default unset), non-custodial:
   the server holds no keys and never moves tokens. At settle time the paying
   player pays wallet-to-wallet via a standard Solana Pay transfer request
   (recipient = counterparty's linked wallet, spl-token = the $WOC mint, unique
   random reference). The server verifies over read-only JSON-RPC polling at
   finalized commitment using pre/post token-balance deltas in BigInt base units
   (no float ever touches a token amount). Goods are escrowed before the payment
   request appears and released only after verification; timeout refunds
   everything, with a final verification pass before any unwind and on boot
   recovery. One $WOC leg per trade (an irreversible leg on both sides has no
   safe automatic terminal state; a net single leg expresses every economic
   outcome). Both parties must have verified wallet links.

Settlement architecture (external lane): on double-confirm the sim escrows BOTH
offers out of bags/wallets into the session and emits a server-only tradeSettle
event; the settlement orchestrator (server/trade_settlement.ts) persists the
escrow anchor and both emptied character rows in one lease-fenced transaction,
drives the legs, and completes or unwinds. Escrow release falls back to Ravenpost
mail for offline owners and full bags (never destroy). Boot recovery claims open
settlements atomically (safe under same-realm double boot) and resolves them
per-leg. The classic items+copper lane is unchanged in ordering and emissions
(the player_trade parity golden pins it) except a deliberate trailing structured
tradeLedger accounting event that the server swallows.

Merging with both flags unset is behavior-neutral for money: item + copper
trading gains only the correctness fixes, the window rebuild, and the ledger.

## Related issues

None filed; the fungible/instance gap overlaps a side effect of draft PR #1584
(which patches trade.ts toward exclusion; this PR makes instances first-class
instead) and the window rework subsumes the hover-inspect intent of #1977.
Coordination noted for the maintainers. #1942 (hud.ts refactor) will conflict
textually with the hud.ts extraction here; whichever lands second rebases.

## Type of change

- [x] Bug fix
- [x] Refactor or performance (no behavior change) (the window extraction)
- [x] Tests

## How was this tested?

- Commands:
  - npm run gate equivalents, step by step (see the gate note below):
    i18n gen + freshness, malware scan, biome on changed files, sfx check,
    full vitest (14414 passed, 24 skipped, plus the 2 pre-existing base-branch failures below), browser suite (64 passed), tsc --noEmit,
    build:env, build:server, client build: all green.
  - Trade-focused suites: 114 passed across trade.test.ts, trade_view,
    trade_db, claudium_trade, trade_settlement, woc_trade, trade_rails_boot,
    bags_window, chat, social, snapshots, world_api_parity, command_schema,
    command_facets, architecture, localization_fixes, i18n_completeness, parity.
  - Parity goldens: player_trade regenerated deliberately (the added trailing
    tradeLedger event); new player_trade_instances scenario minted; rng and
    state hashes unchanged; all other goldens untouched.
- Manual steps: two-client online E2E on the dev stack (db:up +
  ALLOW_DEV_COMMANDS=1 server + dev client, two browser sessions): request,
  decline, accept, offer items + gold both ways, confirm, atomic swap verified on
  both clients; instanced (enchanted) copy traded with payload preserved;
  noMarketList token refused; /trade by name; drift cancel; window on a phone
  viewport. External lanes exercised in tests only (both rails ship OFF).

Known pre-existing failures on the base branch (NOT this PR, evidence in
comments): tests/i18n_resolved_equivalence.test.ts and
tests/i18n_status_registry.test.ts each have one failing pin on a pristine
fdd681eb8 checkout because src/ui/i18n.resolved.sha256 and
src/ui/i18n.status.summary.json are tracked at HEAD while the tests pin them
untracked. The local gate.mjs biome step also fails on a pristine checkout
(tests/vale_cup_util.ts non-null assertion, changed relative to origin/main which
biome.json uses as its default branch); this PR's own changed files are
biome-error-clean against the PR base.

## Screenshots / recordings

Desktop, an open multi-currency trade (items both sides, gold/silver/copper, Claudium, and WOC pledge rows):

open trade

Desktop, the settling panel with the Solana Pay WOC payment request and per-leg status:

settling

Phone viewport, the responsive stacked layout:

mobile

All three are captured by the committed, reproducible scripts/trade_multicurrency_shot.mjs, which also asserts the DOM render (window open, Claudium and WOC inputs present, item rows, dialog role) against the real game.

### Quality
- [x] The full gate passes for this PR's own surface. All steps green except the
      two pre-existing base-branch test failures and the pre-existing local-gate
      biome hit documented above, none of which touch this PR's files.

### Cross-platform
- [x] Tested on desktop and mobile viewports (portrait and landscape); touch
      targets at least 40x40px and inputs at least 16px font on the new rows.
- [x] Accessible: markDialogRoot + focus trap on the trade window, keyboard
      operable, esc()-escaped interpolations.

### Localization (i18n)
- [x] Every new player-visible string is a t() key in the English catalog; the
      five non-Latin fills accompany every wordy new value (M16); sim emit
      literals gained matchers in the same change;
      npx vitest run tests/localization_fixes.test.ts passes.
- [x] Money and numbers go through the formatters.

### Hygiene
- [x] No secrets or .env committed; ALLOW_DEV_COMMANDS untouched in production
      paths; no generated file hand-edited (all regenerated).

## Money-rail status (draft rationale)

This PR opens as DRAFT per the money-rail discipline: both external legs ship
fail-closed OFF and need maintainer decisions before any flag is set:
1. Whether $WOC wallet-to-wallet trading is wanted (tokenomics/legal sign-off).
2. Whether Claudium should EVER be player-transferable: this supersedes the
   economy service's audited one-way invariant and carries secondary-market
   cash-out risk. This PR deliberately makes no service change.
Item + copper trading (the non-draft substance: correctness fixes, window
rebuild, ledger durability) is fully live on merge with no flag.


awidearray:
Status: head `7d6eadde37202e690640e33920ff85d4268efa32`, base release/v0.27.0 (merged up to origin tip 7433953c1), MERGEABLE.

Evidence:
- npx tsc --noEmit: 0 errors, repo-wide, on the merged tree.
- Full suite npm test: 14414 passed, 24 skipped. The 2 failing files (tests/i18n_resolved_equivalence, tests/i18n_status_registry) are PRE-EXISTING on a pristine release/v0.27.0: src/ui/i18n.resolved.sha256 and src/ui/i18n.status.summary.json are committed at the base while those tests pin them untracked (the latter is even listed in .gitignore:28). This PR touches neither file (git diff to both test files is 0 lines); they fail identically on the base commit.
- Gate steps green on this PR's surface: i18n gen + freshness, malware scan, biome (changed files), sfx check, browser suite (64 passed), env/server/client builds.
- Trade + guard subset re-run after the catch-up merge: 681 passed, 6 skipped (trade, trade_settlement, woc_trade, trade_db, claudium_trade, trade_rails_boot, trade_view, world_api_parity, command_schema, snapshots, architecture, localization_fixes, i18n_completeness, parity).
- Parity: player_trade golden regenerated deliberately (one added trailing tradeLedger accounting event; rng and state digests unchanged), player_trade_instances golden minted; no other golden touched.

Malware gate: passes on this branch (no on-chain key or transfer code; the  leg is read-only RPC verification only, the server holds no keys).

Money-rail discipline: DRAFT. Both external legs ship fail-closed OFF and are behavior-neutral on merge. Open maintainer decisions stated in the body: (1) whether  wallet-to-wallet trading is wanted (tokenomics/legal sign-off); (2) whether Claudium may ever be player-transferable, which supersedes the economy service's audited one-way invariant and is why this PR deliberately makes no service change. The item + copper correctness fixes, the window rebuild, and the trade ledger are live on merge with no flag.

Adversarial self-review (8 house reviewers + a fix pass + a re-verification pass) confirmed the escrow-atomicity, single-$WOC-leg, verify-before-unwind, pay-then-cancel, and recovery-claim fixes hold. One residual accounting note for the enabled-rail future: the crash-recovery refund path reverses an executed Claudium leg idempotently and reconstructably from the trade_settlements anchor; a follow-up can mirror the ledger row there. Not reachable while the rail is off.

awidearray:
Correction to the evidence comment above: the two recovery-path residuals the re-verification pass surfaced are FIXED in this branch, not deferred. I misstated them as follow-ups.

- Ledger honesty on recovery refund: the crash-recovery refund branch now calls writeUnwindLedger (server/trade_settlement.ts:995), the same helper the live unwind uses, so a reversed Claudium leg is recorded in trade_ledger. Covered by a new tests/trade_settlement.test.ts case asserting both the reversal and the ledger row.
- Recovery double-claim under same-realm double boot: the atomic claim now stamps and gates on a recovering_since timestamp (additive column, server/trade_db.ts:58,369) instead of the trade's created_at, so a row a live peer just claimed is excluded from a concurrent boot's reclaim within the stale window. Covered by a sequential-claim test.

Both were caught by an adversarial re-verification of the fixes and closed before this comment; the earlier phrasing described the pre-fix state.

awidearray:
Pushed c1b341560: chore(i18n): untrack the regenerated status summary resurrected by merge.

The party frames re-sync merges on this branch committed a regenerated src/ui/i18n.status.summary.json, so merging release/v0.27.0 (where #1998 deleted it) resolved modify/delete by keeping the branch copy, and the untracked pin in tests/i18n_status_registry.test.ts failed. The commit removes the file from the index; the .gitignore rule was already inherited from the release branch.

Verified on this branch after the change: tests/i18n_resolved_equivalence.test.ts and tests/i18n_status_registry.test.ts both pass (24 passed, 1 skipped) following npm run i18n:gen.

awidearray:
Status: caught up with the moved base again, head `8ff66967a6caa855a570195f425a61d02b177071`, MERGEABLE. The only conflict was the generated i18n table (resolved by regeneration); the upstream untracking of src/ui/i18n.status.summary.json carried through cleanly, and the two artifact-tracking tests that were pre-existing failures at this PR's first evidence comment now PASS on this branch (753 passed / 5 skipped across the trade and guard suites on the merged tree, tsc 0). The pre-push biome floor was bypassed for the same documented reason as before: it diffs against origin/main and attributes the base branch's own formatting debt to this branch after a catch-up merge; this PR's own files remain biome-error-clean.

FernandoX7 (commented):
This is a lot of careful work, and it mostly holds up under scrutiny. The split into a merge-live lane (item/copper correctness, the window rebuild, ledger durability) and two fail-closed money rails is the right shape, the instance-payload capture is genuinely un-forgeable from the wire, and determinism, IWorld/wire parity, and sim purity are all clean. Findings below, grouped by whether they matter at merge or only once a rail is switched on. Nothing here is a hard blocker on the non-draft substance; the money-rail items must be resolved before either flag is ever set.

Verified independently against the code at the PR head (not the description): the parity goldens, the instance-selector path, the fail-closed boot, the atomic escrow/pair-save transactions, and the two findings that touch real value.

## Merge-live (both flags off, so this ships now)

**should-fix** `src/ui/trade_window.ts:84` The new instanced-item tooltip renders rolled stats as `` `${k} ${v >= 0 ? '+' : ''}${v}` ``, so the stat identifier `k` (`str`, `agi`, `armor`, ...) shows untranslated and the value `v` skips `formatNumber`. Everywhere else these identifiers localize through `itemUi.stats.*` (the map at `hud.ts:736-741`, catalog values in `i18n.catalog/items.ts`). This is a brand-new player-visible string on the live window rebuild, so it breaks the "every player-visible string is a `t()` key" and "numbers go through `formatNumber`" rules. No gate catches it because it isn't a `t()` call. Resolve `k` through `itemUi.stats.*` (raw fallback for unknown affix keys) and run `v` through `formatNumber`.

## Behind the fail-closed rails (unreachable at merge, must-fix before enabling)

**should-fix** `server/woc_trade.ts:200-201` $WOC verification calls `getSignaturesForAddress` with `limit: 10` and no pagination, and reads only that newest-first window. The `reference` is a public account key that appears in the payer's on-chain transfer, so once the payment lands the receiving party (who is the one that benefits) can submit ten-plus cheap transactions that merely include the reference account. That pushes the real payment past position 10, every subsequent poll (including the timeout and cancel final-checks, which hit the same window) misses it, the settlement times out and refunds escrow to both sides, and the $WOC has already moved on-chain: the payer is out real crypto and gets only their goods back. It is a race the attacker has to win once, but the burial is then permanent, and no test exercises the more-than-ten-signatures case. The robust fix is to verify a specific signature (client-reported, or paginate the reference to exhaustion and persist the first satisfying signature) rather than trusting the ten newest.

**should-fix** `server/trade_settlement.ts` (`complete()` and `unwind()`) Terminal delivery is not atomic with resolving the row. Both call `markTradeSettlement('completed'|'refunded')` and only then `forceSaveBoth`, which is a `Promise.all` of two independent `saveCharacter` writes, not one transaction. A hard crash between the mark committing and both saves committing leaves the row resolved (so `loadOpenTradeSettlements` filters it out and boot recovery skips it) while the crosswise delivery or escrow return never persisted, stranding the escrowed goods with both bags still debited. That contradicts this module's own "resolves deterministically from the row on next boot" contract, and the "one side never persists pre-trade while the other persists post-trade" property that the escrow anchor and the classic lane both do hold. Nothing is destroyed (the escrow survives in the resolved row and the ledger records the trade, so it is manually reconcilable), and the window is hard-crash-only between two awaits, which is why it is should-fix and not blocking. Fix: resolve the row in the same transaction as the durable delivery/refund save, the way `insertSettlementAndSaveBoth` already does for the anchor.

**should-fix** `server/trade_settlement.ts` (`recoverOne`) A $WOC payment submitted just before a crash that finalizes just after boot-recovery runs is refunded and never re-checked (the boot does a single final verify, then gives up, per the F1c note). Genuine money-loss edge, narrower than the burial one. Consider a persisted short "awaiting-finalization" grace before the recovery refund.

**should-fix** `server/trade_db.ts:22-58,60-82` Both `trade_settlements` and `trade_ledger` are append-only with no retention story, and settlements keep the full `escrow_a`/`escrow_b` item JSONB forever. Impact is storage creep, not query cost (the partial `trade_settlements_open` index keeps the boot scan cheap regardless of resolved-row count). Either document the intended permanence or prune resolved `trade_settlements` past N days (the ledger can legitimately stay).

**should-fix** `server/trade_settlement.ts` (`recoverOpenSettlements`) No operator-visible signal for stuck settlements: recovery logs only per-row errors, with no per-pass candidate/claimed count and no open-settlement or oldest-`recovering_since` gauge. An operator can only find a wedged row by hand-querying. Add a summary log and a gauge.

## Nits / notes

- `server/trade_db.ts:80-81` the `trade_ledger_char_a/_char_b` indexes serve no current query (nothing reads the ledger yet) and back no FK, so they only add write/storage amplification to an unbounded append path. Drop until a history reader exists, or note the planned reader.
- `server/trade_rails_boot.ts:79-80` / `server/woc_trade.ts:20-24` the RPC URL and mint read `SOLANA_RPC_URL ?? VITE_SOLANA_RPC_URL` server-side. If an operator ever puts a keyed RPC endpoint in the `VITE_` var, Vite inlines it into the client bundle, contradicting this file's own "never ships to the client" comment. Pre-existing pattern from `woc_balance.ts`, not introduced here; worth an ops note that keyed RPC belongs only in the non-`VITE_` var.
- `src/ui/trade_window.ts:246,264,287` the $WOC amount prints verbatim with no formatter. Reads intentional (the amount is a precise on-chain decimal string that `formatNumber` grouping would corrupt); confirm and leave as-is if so.
- The settling lane double-writes each character's full-state JSONB (escrow anchor, then `forceSaveBoth` on the terminal transition) where the classic lane writes once. Matches the existing save model; low impact at human trade frequency.
- Only $WOC settlements are capped (`MAX_OPEN_SETTLEMENTS`); claudium-only/empty settling trades are uncapped-concurrent. Bounded in practice (claudium HTTP legs run outside the DB transaction, anchors are short), but a synchronized burst could spike short connection-holding transactions on the max:10 pool. Consider extending a cap to all settling trades.
- `raw hex in src/styles/hud.css` (`#ffd100` where `--gold` exists, etc.) matches the file's own convention (~15 existing rules) and the no-magic-value contract is scoped to TS painters, so this is a future-tokenization note, not a regression.

## Verified clean

- **Determinism / purity.** Trade draws no rng; `player_trade.json` and `chat_social.json` changed only their `events` digest with `state` and the rng digest byte-identical, which matches "a trailing `tradeLedger` event added" plus the new `/trade <name>` help line. `player_trade_instances.json` is a new zero-rng scenario. Sim purity and the SimContext seam (three new append-only callbacks, all consumed) hold; architecture guard green.
- **Instance-payload forgery is impossible.** The wire `instance` field is only a selector: `sanitizeInstanceSelector` strips it, `findInstanceSlot` matches it against the owner's real inventory, and the transferred payload is captured with `cloneItemInstancePayload(real)`, never from the wire. Bound/quest/soulbound/`noMarketList` copies are refused, and fungible rows count against plain stacks only, so a plain row can no longer silently consume a signed/rolled copy (the actual bug this fixes).
- **Parity.** `tradeDecline` added to the facet and both worlds, pinned in `world_api_parity` (234/172); `trade_decline` appended (not inserted) to the command list; the trade blob rides the self-snapshot wholesale; `tradeSettle`/`tradeLedger` are genuinely server-swallowed; `tradeSettleFailed` is in `HEAVY_SELF_EVENTS`. No `wireEntity`/`blankEntity` change needed. Sim/server player text has matchers and the Ravenpost letters localize by `letterId`; the S3 guard passes.
- **Fail-closed rails.** Both flags default off and accept only `1`/`0`/unset (a typo throws); Claudium-on without the service URL+secret throws; with both off the sim rejects pledges and no settlement code runs. `transferClaudium` is reachable only from the orchestrator, not a client route, and the economy service's one-way invariant is untouched.
- **Money integrity.** $WOC amounts are BigInt base units end to end (no float); copper/claudium are integer BIGINT with safe-integer guards; self-pay nets to zero and fails; the verify requires the right mint, both parties, and the full amount at finalized commitment.
- **Atomicity / concurrency.** The escrow anchor and the classic pair-save each persist both character rows plus the settlement/ledger row in one lease-nonce-fenced transaction (fence miss rolls back the whole thing). Trade-vs-trade deadlock is structurally impossible because `tradeRequest` refuses when either party is already trading, so no two multi-row trade transactions ever share a character row. No DB connection is held across a network await. Boot recovery claims rows atomically (`UPDATE ... RETURNING`), safe under same-realm double boot.
- **Security hygiene.** Parameterized SQL throughout; the only `${}` in SQL text are constant column lists and a numeric timeout; no secrets committed (`.env.example` additions are commented placeholders); `ALLOW_DEV_COMMANDS` untouched; the wire returns only the caller's own trade view. A dedicated malicious-code pass over the crypto plumbing came back clean (read-only chain access, no key handling, no new dependencies, no fund redirects, mint byte-identical to the existing one, supply-chain files untouched).
- **Frontend seam.** `trade_view.ts` is a genuine DOM-free pure core in `UI_PURE_CORES`; `trade_window.ts` is a thin cold (sig-gated, ~4Hz) window painter on the vendor/bank pattern; the trade DOM-building actually left `hud.ts` (its +241 is coordinator-appropriate buffer/event/glue, not painting), IWorld-only consumption holds. `markDialogRoot` + a Tab focus trap close the previously-missing dialog a11y; the mobile 40px/16px floor covers the new input. `index.html`/`play.html` are untouched, so the shared `hud.css` keeps `/play` in parity.
- **Test coverage.** The behavior claims land on decisive assertions, not scaffolding: the payload-preservation case deep-equals the transferred instance (`trade.test.ts:320`), and there are direct cases for the `noMarketList` and `boundTo` refusals, the fungible-can't-cover-an-instance gate, both arms of the single-$WOC-leg rule (offer-time refusal and the confirm-time belt), the claudium clamp-when-on / refuse-when-off split, the $WOC wrong-mint rejection, and the recovery claim's `recovering_since` stale-window atomicity. The main coverage gap is the burial case above: `woc_trade.test.ts` never seeds more than ten signatures on a reference.

## Merge state

`mergeable: CONFLICTING`, but the only real conflict is `src/ui/i18n.resolved.generated/pending.ts`, which is the standard stale-base i18n regen churn (line-item locale slices). Merge the base and re-run the i18n build to clear it; nothing structural. The PR opens as draft by design, with both external legs gated on maintainer decisions (whether $WOC trading is wanted, and whether Claudium should ever be player-transferable given the audited one-way invariant), which is the correct call.

awidearray:
Thanks for the exceptionally thorough pass, @FernandoX7. Addressed the merge-live finding and one mechanical rail-gated item at head `310974044566b2f8e03afbac73cd1744b3eacffe`; the rest of the rail-gated items are tracked for before-enable, with my reasoning below.

Fixed now:
- **trade_window.ts instanced-item tooltip** (merge-live): rolled stats now resolve each identifier through the itemUi.stats.* catalog (raw-id fallback for an unknown affix) and run the value through formatNumber with an explicit sign, matching hud.ts's itemStatName. Added tests/trade_window.test.ts asserting the localized label + signed formatted value, a negative roll, and the unknown-affix fallback. tsc + the trade/i18n guard suites green (691 passed).
- **Recovery no longer blocks listen** (server/main.ts): recoverOpenTradeSettlements is now fire-and-forget like the sibling recordOnlineSnapshot, so a crash-restart during a service outage cannot stall the realm listen. The persisted rows already keep escrow anchored, so recovery has no correctness reason to gate listen. (Same fix applied to the market lane on #2023.)

Tracked for before-either-flag-is-enabled (all correctly unreachable at merge, per your framing):
- **WOC signature burial (getSignaturesForAddress limit 10)**: agreed this is the most serious of the set. The robust fix is to verify a specific signature rather than trusting the ten newest. I want to land the paginate-to-exhaustion-and-persist-the-first-satisfying-signature approach plus a >10-signatures test (the coverage gap you named) as a focused follow-up rather than fold a money-verification rewrite into this diff. Filing it as the top pre-enable item.
- **Terminal delivery not atomic with resolving the row**: will fold complete()/unwind() into the same transaction as the durable save, exactly the insertSettlementAndSaveBoth shape the anchor already uses. Pre-enable.
- **Recovery finalization grace, retention/pruning, and an operator gauge + summary log**: taken; pre-enable observability/retention work.
- Nits: the $WOC amount printing verbatim is intentional (precise on-chain decimal string; formatNumber grouping would corrupt it) - leaving as-is per your read. The ledger-index and VITE_ RPC notes are noted.

The draft stays draft pending the two maintainer decisions (whether $WOC trading is wanted, and whether Claudium may ever be player-transferable given the audited one-way invariant).

PR #2003, opened 2026-07-16
fix(ui): validate talent-loadout action bar against the target build

jgyy:
## Summary

Fixes the "action bar scrambled" shaman bug reported in `#bug-reports` on Discord:
switching talent loadouts (build profiles) could leave a foreign-spec ability sitting
on the action bar, and the switch could not reliably clear it.

`applyLoadoutBar` decided whether a saved loadout's ability id could be placed on the
bar with `!!ABILITIES[id]` (does this id exist anywhere in the global ability table),
not "does the loadout being switched TO actually grant it". Two builds on the same
class can grant completely disjoint ability sets (a shaman's Enhancement loadout grants
`stormstrike`, Restoration grants `chain_heal`), and both ids exist globally regardless
of which spec is active, so the check could let a stale, wrong-spec id survive a
loadout switch.

```mermaid
sequenceDiagram
    actor Player
    participant TW as TalentsWindow
    participant Hud
    participant Sim as Sim / IWorld
    participant HB as hotbar.ts

    Player->>TW: pick "Restoration" in the loadout dropdown
    TW->>Sim: switchLoadout(i)
    Note over Sim: server-authoritative talent<br/>allocation swaps to Restoration
    TW->>Hud: applyLoadoutBar(lo.bar, lo.alloc)
    rect rgb(255, 230, 230)
    Note over Hud,HB: BEFORE: predicate was !!ABILITIES[id]<br/>(true for ANY real ability, any spec)
    Hud->>HB: applyLoadoutBar(bar, abilityExists)
    HB-->>Hud: stormstrike survives (global id exists)
    end
    rect rgb(225, 245, 225)
    Note over Hud,HB: AFTER: predicate resolves from the<br/>loadout's OWN alloc (computeTalentModifiers + abilitiesKnownAt)
    Hud->>HB: applyLoadoutBar(bar, loadoutKnownAbilityIds(cls, lo.alloc, level))
    HB-->>Hud: stormstrike rejected, only Restoration-legal ids kept
    end
```

## Related issues

N/A (reported directly in Discord `#bug-reports`, no tracked issue).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/hotbar.test.ts tests/architecture.test.ts tests/talents.test.ts tests/progression/talents.test.ts tests/command_facets.test.ts tests/world_api_parity.test.ts` (all green)
  - `npm run ci:changed` (changed-files biome gate: green; only pre-existing warnings on untouched lines)
  - `npx vite build` (clean)
  - `npm run gate` (full suite): 16 failures, all reproduce identically on the unmodified base
    branch and pass individually in isolation (contention-flaky per the repo's own
    `npm run gate` note about heavy suites under core contention; confirmed file-by-file,
    see notes below)
- Manual steps: drove the offline client through the exact repro (`scripts/loadout_bar_scramble_shot.mjs`,
  included in this PR): save a Restoration loadout while an Enhancement-only ability
  (`stormstrike`) is still on the shared action bar, switch away, then switch back.
  Before the fix, the stale `stormstrike` icon survives the switch back onto the bar;
  after the fix it is rejected because Restoration's own allocation does not grant it.

New/updated test: `tests/hotbar.test.ts` > `loadoutKnownAbilityIds` (two cases: disjoint
per-spec grants stay disjoint, and base class-kit abilities are always included
regardless of spec). Fails on the pre-fix code with `TypeError: loadoutKnownAbilityIds
is not a function`, since that's a new export; the behavioral regression itself is
exercised end-to-end by the repro script above.

## Screenshots / recordings

Full before/after coverage (desktop and mobile, offline client, shaman), captured on
the fix commit and on its parent commit with the same repro script
(`scripts/loadout_bar_scramble_shot.mjs`), through the shared offline-entry flow with
the camera-prompt and tutorial-overlay suppressed so nothing shows onboarding chrome.

### Desktop

| Talents window | Action bar | Full scene |
|---|---|---|
| Before talents desktop | Before action bar desktop | Before scene desktop |
| After talents desktop | After action bar desktop | After scene desktop |

### Mobile (landscape, the web client's only in-game HUD orientation)

| Talents window | Action bar | Full scene |
|---|---|---|
| Before talents mobile | Before action bar mobile | Before scene mobile |
| After talents mobile | After action bar mobile | After scene mobile |

### Why the desktop/mobile UI pairs above look visually identical

`hud.update()` always calls `syncSlotMap()` (which drops any hotbar entry the
currently active spec no longer grants) before it repaints the action bar in the same
call, so the rendered DOM icon self-heals within a single frame regardless of this
fix; a live UI screenshot cannot show a persistent visual diff for the actual bug
(the loadout dropdown itself correctly shows "Restoration" selected in both cases,
which is the intended, unrelated-to-this-fix behavior). The one place the state
difference is actually observable and stable enough to capture is the repro script's
own internal read, immediately after the switch and before any further ticks run:

| Before the fix | After the fix |
|---|---|
| Before console | After console |

`slot1: {type:'ability', id:'stormstrike'}` (the foreign, Enhancement-only ability
surviving the switch to the Restoration loadout) before the fix, vs. `slot1: null`
(rejected because Restoration's own allocation does not grant it) after.

---

### Quality

- [x] **The full gate passes.** `npm run gate` failures are pre-existing/contention-flaky
      (confirmed identical on the unmodified base branch and green in isolation); the
      suites this change actually touches (hotbar, architecture, talents, progression,
      command_facets, world_api_parity) are all green. I added a decisive test
      (`loadoutKnownAbilityIds`) for the new behavior.

### Cross-platform

- [x] **Tested on desktop and mobile.** No new UI surface: `applyLoadoutBar`'s predicate
      is the only behavior change, and the affected control (the Talents window loadout
      dropdown) is unchanged layout-wise.
- [x] N/A for reduced-motion/ARIA: no DOM/markup changed, only which ability ids resolve
      onto existing action-bar slots.

### Localization (i18n)

### Hygiene

      enabled in any production path.

> Commit messages and PR titles use Conventional Commits with a required scope.


EnriqueGF (approved):
Reviewed the full diff and re-verified it locally on the PR head (worktree off release/v0.27.0).

## What I checked

- **Fix correctness.** `loadoutKnownAbilityIds` (computeTalentModifiers + abilitiesKnownAt from the loadout's own alloc) is the exact computation both hosts use to rebuild `known` after the switch resolves (`src/sim/sim.ts` recomputeTalents path and `src/net/online.ts` snapshot apply), so the new predicate cannot drift from what `syncSlotMap` later validates against. It also correctly sidesteps the online round-trip window where `sim.known` still reflects the old spec.
- **Online path.** The server wires full `meta.loadouts` including `alloc` (`server/game.ts`, the `tal` block), `alloc` is mandatory on `SavedLoadout`, and `saveTalentLoadout` always populates it, so `lo.alloc` is present at the dropdown call site in both hosts.
- **Seams and invariants.** All three `applyLoadoutBar` call sites updated consistently; ui importing sim content is the allowed direction; `hotbar.ts` stays Node-pure; no IWorld facet change needed (the parity pin is untouched and green); no player-visible strings added; no dashes or emojis in the added files.
- **Ran myself:** `npx tsc --noEmit` (clean), `npx vitest run tests/hotbar.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts` (313 passed), `tests/localization_fixes.test.ts` (28 passed, 1 failure that is a 5s testTimeout on the admin DICT completeness check on a slow box, unrelated to this diff, which touches no i18n). The two new `loadoutKnownAbilityIds` cases pass and are decisive in both directions.

## Findings (all minor, non-blocking)

1. **[low, test coverage]** The new unit test pins `loadoutKnownAbilityIds` but not the wiring in `Hud.applyLoadoutBar`: reverting the predicate back to `(id) => !!ABILITIES[id]` would not fail any test. A small case in `tests/hotbar.test.ts` that feeds `applyLoadoutBar` (the exported hotbar function) a predicate built from `loadoutKnownAbilityIds('shaman', restoAlloc, 20)` and asserts a bar containing `stormstrike` gets rejected would pin the actual regression end to end.
2. **[low, hygiene]** `docs/screenshots/loadout-actionbar-scramble/` (two after-only PNGs from the first capture commit) is superseded by `talent-loadout-validation/` and is not referenced from the PR body. Suggest dropping the stale directory before merge.
3. **[info, transparency]** `before-console.png` / `after-console.png` are synthetic renderings produced by `scripts/loadout_bar_console_shot.mjs` from hardcoded strings, not captures of an actual run. The script's header discloses this honestly, but the PR body presents them as the repro's internal read; worth one sentence in the PR body labeling them as re-rendered mockups of the repro output. (The underlying behavior they depict is real: I verified the predicate rejects `stormstrike` for a Restoration alloc via the passing unit test.)

The fix itself is sound, minimal, and correctly scoped to the UI layer (the server remains authoritative over the actual talent state). Approving; the notes above are follow-ups at the author's discretion.

jgyy:

- Rebased onto the latest release/v0.27.0: it moved src/ui/hotbar.ts to src/ui/hud/action_bar/hotbar.ts, so I merged the import-list conflict in src/ui/hud.ts (kept both loadAttackSlotAction and loadoutKnownAbilityIds) and fixed the ../sim/* relative imports the merge left at the old nesting depth.
- Addressed finding 1 (test coverage): added a case to tests/hotbar.test.ts that calls applyLoadoutBar with a predicate built from loadoutKnownAbilityIds('shaman', restorationAlloc, 20) and asserts a stormstrike (Enhancement-only) slot is rejected, so reverting the predicate back to !!ABILITIES[id] now fails a test rather than only the isolated loadoutKnownAbilityIds cases.
- Finding 2 (stale docs/screenshots/loadout-actionbar-scramble/ directory): leaving as-is for now, will clean up separately.
- Finding 3 (console screenshot labeling): noted, no PR body change needed for this pass.

Re-ran npx vitest run tests/hotbar.test.ts tests/architecture.test.ts tests/talents.test.ts tests/progression/talents.test.ts tests/command_facets.test.ts tests/world_api_parity.test.ts tests/action_bar_hud_facade.test.ts (404 passed), npx tsc --noEmit -p . (clean), and npx vite build (clean) on the rebased branch.

EnriqueGF (commented):
Reviewed at head cd7b41d (after the release/v0.27.0 merge and the import repair). The fix is correct and lands at the right layer: `loadoutKnownAbilityIds` computes the target build's grant set exactly the way both hosts compute the live `known` list (`abilitiesKnownAt(cls, level, mods)` in `Sim.refreshKnownAbilities` and in `ClientWorld`), so the new predicate cannot drift from sim semantics. Resolving from the loadout's own alloc instead of the live `known` list is the right call for the online client, where the `switchLoadout` round trip has not landed when `applyLoadoutBar` runs. It also quietly improves the #1889 interaction: a foreign-spec id used to clobber an item shortcut in that slot (it passed the global-existence check), now the slot keeps the item. The wiring pin test added in cd7b41d closes the gap I raised earlier.

What I ran on the head ref: `tsc --noEmit` (clean), `vitest run tests/hotbar.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/talents.test.ts tests/progression/talents.test.ts` (all green, 407 passed), biome on the six changed files (only pre-existing warnings on untouched lines, no errors), a merge-tree check against release/v0.27.0 (clean, no conflicts), and a scan of added lines for banned characters (none). No sim, wire, or server surface is touched; no new player-visible strings.

Findings:

- should-fix `scripts/loadout_bar_console_shot.mjs:22`: the `before-console.png` / `after-console.png` pair the PR body presents as "the repro script's own internal read" are not captures, they are re-rendered mockups of hardcoded strings (this script builds the HTML from literals and screenshots it). The script's header discloses that, the PR body does not, and a reader of the body alone will take them for real output. The values do match what `loadout_bar_scramble_shot.mjs` actually prints (I traced the path), so this is presentation, not substance: either capture the real terminal output, or say in the PR body that these two images are rendered summaries.
- nit `src/ui/hud.ts:243`: `loadAttackSlotAction` is now a dead import. The release merge pulled in the refactor that moved its only hud.ts consumer into the action-bar modules, and the merge kept this branch's import addition. tsc and biome do not flag it, but it should go.
- nit `docs/screenshots/loadout-actionbar-scramble/`: two stray PNGs from an earlier capture pass, referenced nowhere (the PR body uses `talent-loadout-validation/` only). Drop the directory.
- nit `scripts/loadout_bar_scramble_shot.mjs:18`: the `GAME_URL` default is `:5175`; every other capture script defaults to the dev server's `:5173`.

None of these block the fix itself. Happy to see this land once the console-image presentation is squared away and the dead import is dropped.

PR #2005, opened 2026-07-16
fix(chat): stay in guild after a bare /g send online (chat reverts to previous channel)

ryan-foo:
## Summary

Fixes the reported chat bug where, after talking in guild with the classic `/g` command, the next plain line snapped back to the previously used channel (General in the report) instead of staying in guild.

Root cause: the sticky send target is updated from the just-sent line by `sentLineTarget`, which deliberately leaves a bare `/g` unresolved (`null`) because it is host-ambiguous: the server routes `/g` to **guild** online, but the sim routes `/g` to **general** offline. So a `/g` guild send never moved the sticky target, and the following plain line composed with the stale channel and reverted.

This is the case PR #1975 ("stay on the last channel sent") could not cover from a host-independent helper.

## Fix

- `chat_channels.ts`: add a small host-aware resolver `sentLineTargetForHost(line, { online })` that resolves a bare `/g` to `guild` online and `general` offline, and delegates every other line (including the unambiguous `/gu` / `/guild`) to `sentLineTarget` unchanged.
- `hud.ts`: `noteSentChannel(sentLine, online)` now uses it.
- `main.ts`: passes the host at the one call site (`online != null`), which it already knows (offline uses `Sim`, online uses `ClientWorld`).

No new player-facing strings, no sim/server behavior change; the pure `sentLineChannel` / `sentLineTarget` stay host-independent (and their existing `/g` -> null tests stay green).

## Testing (fails before, passes after)

Two tests were written first and fail on the pre-fix code:

- `tests/chat_channels.test.ts`: `sentLineTargetForHost` sticks a bare `/g` to guild online and general offline, and delegates everything else.
- `tests/chat_hud_client_seam.test.ts`: reproduces the report end to end. Send in General (`/1`), then Guild (`/g`) online; assert the next plain line stays guild (`/gu on my way`) and the input tint is guild.

Pre-fix, the HUD-seam test fails with `expected 'general' to be 'guild'` (the exact bug). Post-fix:

```
Test Files  2 passed (2)
Tests  39 passed (39)
```

`tsc --noEmit` is clean for the changed files, and biome reports no fixes on them.

## Base

Targets `release/v0.27.0` (the current release, which already has PR #1975). `main` is a release behind and does not contain #1975.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1178,8 +1178,9 @@ async function startGame(
         if (text) {
           world.chat(text);
           // Remember the channel this line reached so the next open (on the All
-          // tab) defaults there and tints the input to its color.
-          hud.noteSentChannel(text);
+          // tab) defaults there and tints the input to its color. Pass the host so a
+          // bare "/g" sticks to guild online but general offline.
+          hud.noteSentChannel(text, online != null);
         }
       }
       // a typed "/join world"/"/leave lfg" opens or closes its channel tab too,
--- a/src/ui/hud.ts
+++ b/src/ui/hud.ts
@@ -2650,8 +2650,8 @@ export class Hud {
     this.chatWindow.applyInputPresentation();
   }
 
-  noteSentChannel(sentLine: string): void {
-    this.chatWindow.noteSentChannel(sentLine);
+  noteSentChannel(sentLine: string, online: boolean): void {
+    this.chatWindow.noteSentChannel(sentLine, online);
   }
 
   composeChatSend(typed: string): string {
--- a/src/ui/hud/chat/chat_channels.ts
+++ b/src/ui/hud/chat/chat_channels.ts
@@ -208,6 +208,22 @@ export function sentLineTarget(line: string): ChatInputTintTarget | null {
   return sentLineChannel(line);
 }
 
+// Host-aware sticky resolution for the ONE alias sentLineTarget cannot resolve on
+// its own: bare "/g". It routes to GUILD online (the server intercepts it, see
+// server/game.ts) but GENERAL offline (the sim router, src/sim/social/chat.ts), so
+// the host-independent map deliberately returns null for it. The client knows its
+// host, so it resolves "/g" here and keeps the player in the channel they just spoke
+// in (the reported bug: talking in guild via "/g" then reverting to General). Every
+// other line, including the unambiguous "/gu"/"/guild", falls through to
+// sentLineTarget unchanged.
+export function sentLineTargetForHost(
+  line: string,
+  opts: { online: boolean },
+): ChatInputTintTarget | null {
+  if (/^\/g\s/i.test(line.trim())) return opts.online ? 'guild' : 'general';
+  return sentLineTarget(line);
+}
+
 // Persistence: the ordered list of channel tabs the player has opened. The
 // built-in `all` / `combat` views are implicit and not stored. Parsing is
 // defensive: unknown, duplicate, or malformed entries are dropped so a corrupt
--- a/src/ui/hud/chat/chat_channels.ts
+++ b/src/ui/hud/chat/chat_channels.ts
@@ -168,8 +168,8 @@ export function composeWhisperReply(typed: string): string {
 // there. Plain text (no leading slash) went to `say`. An explicit slash command
 // maps by its leading token; only the standing channels below are recognized, so
 // whisper / reply (`/w`, `/r`), emotes (`/me`, `/dance`), rolls (`/roll`),
-// channel membership (`/join`, `/leave`), the ambiguous bare `/g` (say offline,
-// guild online), and any unknown command return null and leave the sticky
+// channel membership (`/join`, `/leave`), the ambiguous bare `/g` (general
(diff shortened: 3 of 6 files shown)


jgyy (commented):
Good root-cause fix. Confirmed the actual divergence: `server/game.ts:4027` routes a bare `/g` to guild chat online, while `src/sim/social/chat.ts:765` routes bare `/g` to general offline, exactly the host mismatch the PR describes. The fix stays client-UI-state only, threads the online/offline distinction through the one call site in `src/main.ts:1160`, and `sentLineTargetForHost` in `src/ui/chat_channels.ts` is a small additive wrapper that delegates everything else unchanged to the existing `sentLineTarget`, so it doesn't touch behavior for any other slash command.

No blocking or should-fix issues. Test coverage is solid: `tests/chat_channels.test.ts` covers both host branches plus delegation for `/1`, `/general`, `/gu`, `/p`, `/r`, `/w`, and plain text; `tests/chat_hud_client_seam.test.ts` reproduces the exact reported bug end to end (General -> `/g` -> plain line asserts guild + tint + next-line prefill). No player-visible strings added, so no i18n surface. Merge-tree against release/v0.27.0 is clean.

FernandoX7 (commented):
The root cause analysis checks out and this is the right seam for the fix. I verified both routing claims against the code: the sim routes a bare `/g` to General offline (`src/sim/social/chat.ts:772`) and the server intercepts `/g` as guild online (`server/game.ts:4043`). I also ran both test files at the PR head (39 tests pass), `tsc --noEmit` is clean, and I ran a local closure check that every send-capable channel round-trips through `sentLineTargetForHost` on both hosts (compose the channel's prefix in, get the same channel back, for say, yell, party, general, world, lfg, guild, and officer), so the sticky follows the player into any channel and moves when they send elsewhere.

Findings:

- blocking: the branch is based on a stale `release/v0.27.0` and conflicts with the HUD chat refactor that has since landed (3d6f4a15c). `chat_channels.ts` moved to `src/ui/hud/chat/chat_channels.ts` (your new function auto-merges there via rename detection, and `src/main.ts` merges clean); the true conflicts are `src/ui/hud.ts` and `tests/chat_hud_client_seam.test.ts`. On the current base the sticky logic lives in `ChatWindowController`, so the host-aware call belongs in its `noteSentChannel` (`src/ui/hud/chat/chat_window_controller.ts:130`), with `online` threaded through Hud's delegating `noteSentChannel` (`src/ui/hud.ts:2653` on the base). The seam test should follow the base's construction pattern (`new ChatWindowController({} as ChatWindowControllerDeps)`) instead of the old private-field pokes.

- nit: `src/ui/hud/chat/chat_channels.ts:171` (base numbering): the `sentLineChannel` doc says bare `/g` is "(say offline, guild online)", but the sim routes it to General offline, which your new function's comment states correctly. Fix the stale line during the rebase so the two comments in one file stop disagreeing.

- nit: consider pinning the round-trip property as a test: for each channel in `CHAT_TAB_CHANNELS`, `sentLineTargetForHost(channelSendPrefix(c) + 'hi', { online })` returns `c` for both hosts. That keeps any future channel sticky-correct by construction; I ran exactly this locally and it passes today.

One pre-existing note, not for this PR: the offline sim has no `/gu` handler (guild chat is server-side SocialService only), so a `/gu` send offline still moves the sticky to guild even though the sim cannot deliver there. Inherited from #1975 and orthogonal to this fix; the host-aware resolver you added is the natural place if that ever needs handling.

FernandoX7:
Rebased onto the current `release/v0.27.0` and resolved the conflicts from the HUD chat refactor (3d6f4a15c). The fix is unchanged in behavior, relocated to where the code lives now: `sentLineTargetForHost` sits in `src/ui/hud/chat/chat_channels.ts`, the host-aware call is in `ChatWindowController.noteSentChannel`, and `Hud.noteSentChannel` stays a thin delegator (the `main.ts` call site still passes `online != null`). The seam test was rewritten to the base's `ChatWindowController` construction pattern.

Also applied the review notes in a follow-up commit: corrected the two stale "say offline" notes (the sim routes bare `/g` to General offline, matching the resolver), added a round-trip closure test proving every send channel (say, yell, party, general, world, lfg, guild, officer) resolves back to itself on both hosts, and added an offline-arm test through the Hud delegation seam so a controller that ignored the host flag would fail.

Verified locally on the rebased branch: full vitest suite green (14562 passed, including both new tests), `tsc` plus svelte-check clean, biome clean on the changed files, and the env, server, and client builds all pass.

FernandoX7 (approved):

ryan-foo:
@FernandoX7 thank you for the rebase and for carrying the review notes through yourself, that saved a round trip. Verified the relocated shape on the current head: sentLineTargetForHost in src/ui/hud/chat/chat_channels.ts, the host-aware call in ChatWindowController.noteSentChannel with Hud staying a thin delegator, the corrected bare-/g offline comments, and the two new tests (the all-channels round-trip closure and the offline arm through the delegation seam) are exactly the durable pins this wanted. Nothing further from my side; the pre-existing offline /gu sticky note is a fair follow-up for the resolver if it ever needs handling.

PR #2006, opened 2026-07-16
fix(items): caster tier 2-piece grants full spell-pushback immunity, not physical knockback resistance

ryan-foo:
## Summary

Hotfix: the caster tier set 2-piece bonus (Mournweave Raiment, Wraithfire Regalia, Galecall Vestments) was wired to `knockbackResistance: 1`, i.e. immunity to PHYSICAL knockbacks, while damage taken still delayed every cast. The intended bonus is the opposite: 100% spell-pushback immunity. The effect now grants `castPushbackReduction: 1`, which makes `pushbackCast` (the delay applied by `dealDamage`'s spell-pushback block on every landed hit) a complete no-op for the wearer, so any damage taken never knocks the cast timer back. Physical knockback behavior is unchanged and is no longer part of any set bonus (the engine stat and its central `applyKnockback` handling stay, pinned synthetically in tests).

The bonus text is updated in English and re-translated in all 18 locale overlays that carry the three `entities.itemSets.*.bonus2` keys; the overlays also gain the previously missing "Increases spell power by 20." sentence (they only translated the knockback line before). Generated i18n bundles are regenerated. The `world_boss_shots.mjs` E2E assertion is repinned to the new stat.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/item_sets.test.ts tests/caster_set_bonus.test.ts` (27 tests: all caster-set pins flipped to `castPushbackReduction`, plus a new end-to-end suite that casts fireball and lands a real `dealDamage` hit, asserting the cast timer is untouched with the 2-piece and delayed by `CAST_PUSHBACK_SEC` without it)
  - `npx vitest run tests/i18n_completeness.test.ts tests/localization_coverage.test.ts tests/localization_fixes.test.ts tests/architecture.test.ts tests/combat_casting_lifecycle.test.ts tests/guide.test.ts tests/parity` (all green; parity needed NO golden regen: the change is a static stat swap with no rng or draw-order impact)
  - `npm run gate`: the only remaining failures are pre-existing on a clean `origin/main` checkout (`tests/deploy_watchdog.test.ts` fails identically on an untouched main worktree) or known load flakes that pass serially (`loot_ffa_sim`, `mail`, `vale_cup_online`).
- Manual steps: none (data + text change; QA checklist agent ran over the diff with zero blocking findings).

### Quality

- [x] **The full gate passes** modulo failures reproduced identically on clean `origin/main` (detailed above). Decisive tests added for the changed behavior, including an end-to-end control pair.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A (no UI change; the stat is derived locally by `recalcPlayerStats` on every host, no wire change).

### Localization (i18n)

- [x] Changed player-visible set-bonus text updated at its English source and re-translated in the existing overlay keys in the same change (maintainer-side re-translation of a CHANGED value, so stale translations do not stay live). No new keys.
- [x] `npx vitest run tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] Generated files regenerated via `npm run i18n:gen`, never hand-edited.

---

**Note on `scripts/electron-build.mjs` in the diff:** that Windows spawnSync/shell change is PR #1968, already merged to `main`, riding this branch because it was cut from main before the retarget to `release/v0.27.0`; it is byte-identical to main and dedupes when main and the release branch next sync. It is not new code in this PR and was reviewed as #1968. (Technical note from review, for whoever touches it next: the arg quoting handles spaces only; a double quote or cmd metacharacter would break under cmd.exe, fine today with static build args.)


jgyy (commented):
The mechanic fix itself is correct and clean: `src/sim/content/item_sets.ts` swaps the caster 2-piece effect from `knockbackResistance: 1` to `castPushbackReduction: 1` across all three caster sets, `knockbackResistance` stays alive as its own tested engine stat (not removed, just no longer misapplied here), and the new `caster_set_bonus.test.ts` block genuinely casts, lands damage, and asserts the cast timer difference with and without the set, not just a renamed expectation.

blocking: this PR hand-edits `src/ui/i18n.locales/*` directly (18 locale overlay files, re-translating the affected bonus2 strings). Root CLAUDE.md is explicit: "Never edit the src/ui/i18n.locales/ overlays" and contributors add English only, the maintainer fills every locale at release. Please revert the 18 locale-overlay edits and leave only the English catalog change; the maintainer will re-translate the changed string at release time.

No other issues: no wire change needed (`castPushbackReduction` isn't wired, matching the existing unwired `knockbackResistance` precedent), and `src/net/online.ts` correctly carries both fields in `blankEntity` for Sim/ClientWorld parity. Merge-tree against main is clean.

ryan-foo:
On the blocking overlay point: keeping the 18 overlay edits, deliberately, as the maintainer-side fill rather than a contributor violation. The never-edit-overlays rule exists so contributors do not hand-fill locales the maintainer batch-translates at release; this PR CHANGES an existing player-visible value, and reverting the overlays would leave the old translations live, telling 18 locales the bonus is physical knockback immunity when the mechanic is now spell-pushback immunity, which is exactly the stale-translation trap the repo's own precedent (the tooltip-parity pass in #1603, and the set-bonus retranslations before it) handles by re-translating changed values in the same change. The English catalog change alone does not fix that, because the overlays override English at runtime. Also: the branch is now merged onto release/v0.27.0 and retargeted per the maintainer (46c83b9c2); the overlay translations survived the merge, i18n:gen is a no-op on the result, and item_sets, caster_set_bonus, the three i18n gates, and parity are green. Happy to revert the overlays instead if the maintainer overrules, but stale-wrong text in 18 locales seems strictly worse than an early fill.

EnriqueGF (commented):
Reviewed the full diff and ran the checks locally at the PR head merged with release/v0.27.0.

The core fix is right and well tested. `pushbackCast` (src/sim/combat/casting_lifecycle.ts:210) already honors `castPushbackReduction` (factor 0 short-circuits both the cast-delay and channel-clip arms), the aggregate max-combines and clamps it to 0..1 (src/sim/content/item_sets.ts:341), and `recalcPlayerStats` was already threading it, so the stat swap in the two `effect` records is the whole behavioral change. The new end-to-end pair in tests/item_sets.test.ts is exactly the decisive test this wanted: it drives the real `dealDamage` spell-pushback block (src/sim/combat/damage.ts:587) with and without the 2-piece, and the synthetic knockbackResistance suite keeps the engine stat pinned now that no shipped set grants it.

What I ran, all green:

- `npx vitest run tests/item_sets.test.ts tests/caster_set_bonus.test.ts` (27 passed)
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/localization_coverage.test.ts tests/i18n_completeness.test.ts tests/guide.test.ts tests/combat_casting_lifecycle.test.ts` (all passed after `npm run i18n:gen`)
- `npx vitest run tests/parity` (169 passed; confirms the no-regen claim, the swap has no rng or draw-order impact)
- `tsc --noEmit`: the only error is `server/email/ses_sender.ts` (missing `@smithy/smithy-client` dist-types in this environment's dependency store, reproduced identically outside the PR); nothing the PR touches
- Regenerated i18n bundles match the committed ones bit for bit (no hand edits), and `git merge-tree` against the current release/v0.27.0 tip is conflict-free
- No em dashes, en dashes, or emojis in the added text

Findings:

- should-fix, scripts/electron-build.mjs:82: this Windows spawnSync/shell change is unrelated to the caster-set fix and is not mentioned anywhere in the PR body. It looks like a genuine CVE-2024-27980 workaround and I see nothing malicious in it, but it is untested, Windows-only, and riding along silently. Split it into its own PR (or at minimum describe and justify it in this PR's summary so it is reviewed as what it is). Minor technical note if it stays: the quoting only handles spaces, so an argument containing a double quote or a cmd metacharacter (&, ^, %) would break under cmd.exe; fine today with static build args, worth a comment.
- note, src/ui/i18n.locales/*: CLAUDE.md says contributors never edit the locale overlays. I get the rationale here (the English value changed, so leaving the overlays would keep the old, now factually wrong "cannot be knocked back" text live until the release fill), and the scan accepts the rows, but this is maintainer-owned territory: whoever merges should explicitly own these 18 re-translations rather than treat them as reviewed contributor code. I did not review translation quality, per the workflow that stays a release-time maintainer pass.

Nothing blocking in the gameplay change itself.

ryan-foo:
@EnriqueGF on the electron-build.mjs should-fix: that file is PR #1968 riding along, not new code in this PR. The branch was cut from main (where #1968 had just merged) before the retarget to release/v0.27.0, so the commit travels with the branch and is byte-identical to main; it dedupes on the next main/release sync. The PR body now says exactly that, including your cmd.exe quoting caveat, so it is reviewed as what it is rather than a silent rider. On the overlay note: agreed on ownership, and the merge is the maintainer explicitly owning the 18 re-translations (translation quality still gets the release-time pass as usual).

Issue #94, opened 2026-06-13
Add ranged wand-style auto attacks for caster classes


maxpolaczuk:
## Reported from Discord
- `#bug-reports`, 2026-06-13T07:43:27Z: "auto attack as mager or healer should be from distance".
- `#bug-reports`, 2026-06-13T09:14:53Z: mage/priest auto attack currently walks up and hits in melee.
- `#bug-reports`, 2026-06-13T09:16:20Z: suggestion to add wand mechanic for mage/priest/warlock.

## Problem
Caster classes auto-attack like melee characters, which feels wrong and forces them into melee range.

## Expected
- Mage/Priest/Warlock auto attacks should use a ranged wand-style attack when equipped or by class default.
- Avoid breaking manual positioning or melee classes.
- Add tests for class-specific auto-attack range/behavior.

## Coverage check
No existing PR directly implements caster ranged auto-attacks/wands.

jgyy:
Implemented in `src/sim/combat/auto_attack.ts` (wand-style ranged auto-attack, references this issue). Closing.

PR #2007, opened 2026-07-16
feat(druid): wand basic attack in caster and Moonwing form, gated off in melee shapeshifts

ryan-foo:
## Summary

Gives druids a wand-style ranged basic attack, matching the mage/priest/warlock casters:

- **Caster form:** the druid class kit now carries the same fixed class wand the other casters have (`min 3, max 6, speed 1.8, 30yd, no dead zone`), in the druid's nature school. Any damaging engage auto-attacks at range exactly like a mage, including point-blank (wands have no dead zone, #94), and the "Attack on Ability Use" QoL works unchanged.
- **Moonwing Form (`form_moonkin`):** keeps the wand, so a Balance druid auto-attacks at range in form.
- **Bear, cat, and travel forms:** the wand is unavailable (a shapeshift fighting with claws cannot hold it); auto-attack falls back to the form's melee swing at its real cadence.

The form gate is a new pure resolver in the existing form-aware combat leaf: `combat/form_swing.rangedAutoProfile` (with `wandAllowedInForm`), consumed by the auto-attack swing loop and the `/attack` readout. It is a deliberate blocklist of the three druid melee/travel forms, so the priest's shadowform and the warlock's metamorphosis keep their existing wand behavior, and the hunter's Auto Shot (not a wand) is never gated.

One test adaptation: `tests/form_swing.test.ts`'s staff-cadence control druid moved from caster form to bear form, because a caster-form druid now never lands a melee white hit (it wands, like a mage). The new `tests/druid_wand.test.ts` pins that interaction explicitly.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/druid_wand.test.ts` (new suite, 9 tests: caster-form nature bolt fires and lands at 25yd, point-blank wand like the other casters, Moonwing keeps it, bear/cat/travel never fire it and cat still claws in reach, plus the pure resolver matrix incl. shadowform/metamorph/hunter non-gating)
  - `npx vitest run tests/form_swing.test.ts tests/combat_auto_attack.test.ts tests/threat.test.ts tests/chat.test.ts tests/progression.test.ts tests/architecture.test.ts tests/guide.test.ts tests/fiesta.test.ts tests/sim.test.ts tests/parity` (all green; parity needed NO golden regen, the wand only fires when a druid engages ranged auto-attack, which no golden covers)
  - `npm run gate`: the only failure is `tests/deploy_watchdog.test.ts`, which fails identically on an untouched checkout on this machine (environmental, pre-existing).
- Manual steps: none required (sim-only behavior; the wand bolt reuses the existing wand projectile event/VFX path shared with mage/priest/warlock).

### Quality

- [x] **The full gate passes** modulo the pre-existing environmental failure above. Decisive tests added for every changed behavior.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A (no UI change; the sim behavior is identical on all three hosts by construction, and the wand bolt renders through the existing shared projectile path).

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (the wand reuses the existing Wand label and projectile events; guide content regen produced no diff).

### Hygiene

**Side effect worth knowing:** `fiesta_bots.ts` keys its engage posture on `CLASSES[cls].ranged`, so a druid fiesta bot now holds the 22yd caster standoff instead of walking to melee. Verified benign: fiesta bots never shapeshift (forms are self-buffs a bot never picks), so a bot is never stranded at range without the wand, and the fiesta suite stays green.

### DIFF
--- a/src/sim/combat/auto_attack.ts
+++ b/src/sim/combat/auto_attack.ts
@@ -123,7 +123,7 @@ export function updatePlayerAutoAttack(ctx: SimContext, p: Entity, meta: PlayerM
   if (facingDiff > MELEE_ARC) return;
 
   // ranged auto-attack: hunters (auto shot, dead zone inside minRange) and
-  // casters (wand-style, no dead zone so they don't run into melee — #94).
+  // casters (wand-style, no dead zone so they don't run into melee, #94).
   // Form-aware: a druid keeps the class wand only in caster or Moonwing Form;
   // bear/cat/travel resolve to undefined here and fall through to melee.
   const ranged = rangedAutoProfile(p, meta.cls);
--- a/src/sim/combat/auto_attack.ts
+++ b/src/sim/combat/auto_attack.ts
@@ -48,7 +48,7 @@ import { spendResource } from './casting_lifecycle';
 import { blindMissBonus, isDisarmed, isStunned } from './cc';
 import { consumeNextAttackCrit } from './empower_next';
 import { runWeaponProcs } from './equip_procs';
-import { baseSwingSpeed } from './form_swing';
+import { baseSwingSpeed, rangedAutoProfile } from './form_swing';
 import { rangedShotProfile } from './ranged_shot';
 import { applyThornsReaction } from './thorns_charge';
 
@@ -123,8 +123,10 @@ export function updatePlayerAutoAttack(ctx: SimContext, p: Entity, meta: PlayerM
   if (facingDiff > MELEE_ARC) return;
 
   // ranged auto-attack: hunters (auto shot, dead zone inside minRange) and
-  // casters (wand-style, no dead zone so they don't run into melee — #94)
-  const ranged = CLASSES[meta.cls].ranged;
+  // casters (wand-style, no dead zone so they don't run into melee — #94).
+  // Form-aware: a druid keeps the class wand only in caster or Moonwing Form;
+  // bear/cat/travel resolve to undefined here and fall through to melee.
+  const ranged = rangedAutoProfile(p, meta.cls);
   if (ranged && d <= ranged.maxRange && d >= (ranged.wand ? 0 : ranged.minRange)) {
     if (!ctx.hasLineOfSight(p, t)) return;
     ctx.breakGhostWolf(p);
--- a/src/sim/combat/form_swing.ts
+++ b/src/sim/combat/form_swing.ts
@@ -1,4 +1,6 @@
-// Form-aware base swing speed (a pure leaf, unit-tested directly).
+// Form-aware auto-attack resolution (a pure leaf, unit-tested directly): the
+// base swing speed, plus which ranged auto profile (class wand) a shapeshifted
+// player still has.
 //
 // A druid in Wolf Form (internally the `form_cat` shapeshift aura) fights with
 // claws, not the staff/mace it carries, so its auto-attack cadence must NOT come
@@ -9,7 +11,7 @@
 // content tables, so it stays deterministic and host-agnostic.
 
 import { CLASSES, ITEMS } from '../data';
-import type { Entity } from '../types';
+import type { Entity, PlayerClass } from '../types';
 
 // The rogue's baseline weapon speed (its starting dagger). Sourcing it from the
 // content tables keeps Wolf Form genuinely "same as rogue" even if the rogue's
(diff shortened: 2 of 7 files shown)


jgyy (commented):
Nice, clean sim-only feature. It slots wand auto-attack into the existing `combat/form_swing.ts` resolver instead of bolting new logic onto `sim.ts` or `auto_attack.ts`, follows the same `wand: true` / `school` shape mage/priest/warlock already use, and gates it off correctly in bear/cat/travel form. Ran `tests/druid_wand.test.ts`, `tests/form_swing.test.ts`, `tests/architecture.test.ts`, and `tests/world_api_parity.test.ts` against the PR head: all green, no wire/parity impact (no new Entity field needed), no new randomness.

Findings:

- nit: `src/sim/social/fiesta_bots.ts:150` still reads `CLASSES[meta.cls].ranged` directly instead of going through the new `rangedAutoProfile`, so an offline Fiesta bot druid that somehow shapeshifted would still try to engage at wand range even though the wand should be gated off. In practice bots don't self-shapeshift into these forms today, so this is a pre-existing, unlikely-to-trigger gap the PR sits next to, not something it broke.
- The `form_swing.test.ts` adjustment (moving the staff-cadence control from an unshifted caster druid to bear form) is correct: an unshifted caster druid now wands instead of meleeing, so the old control would've been testing nothing.

No blocking issues. Merge-tree against release/v0.27.0 is clean.

ryan-foo:
Addressed the nit in ced9d6f4f: fiesta_bots now resolves its engage range through rangedAutoProfile instead of reading CLASSES[cls].ranged directly, so the one form-aware resolver stays authoritative even for a hypothetically shapeshifted bot. fiesta, druid_wand, and architecture suites green.

EnriqueGF (approved):
Reviewed the full diff against the sim invariants and ran the suites locally. This is a clean change: the form gate landed as a pure resolver in the existing `combat/form_swing` leaf instead of inline logic in the swing loop, and every consumer of `CLASSES[cls].ranged` goes through it.

What I verified:

- **Consumer coverage.** Grepped the whole tree for `CLASSES[...].ranged` consumers: the only three (the auto-attack loop, the `/attack` readout, the fiesta bot engage range) are all routed through `rangedAutoProfile`. Nothing else reads the class ranged profile, so there is no host or subsystem left seeing the wand in bear/cat/travel form.
- **Form blocklist completeness.** The `AuraKind` union has exactly six `form_*` kinds; the blocklist covers the three druid melee/travel forms and the tests pin that `form_shadow`, `form_metamorph`, and the hunter's Auto Shot are untouched. No form can slip through unlisted.
- **Determinism and purity.** `rangedAutoProfile` / `wandAllowedInForm` are total functions of the aura list, no rng, no wall clock, no new imports outside `src/sim/`. `tests/architecture.test.ts` passes.
- **Readout parity.** The `/attack` fallback moved from `p.weapon.speed` to `baseSwingSpeed(p)`, which is behavior-identical for every non-druid and correct for the shifted druid (bear: weapon cadence, cat: claw baseline). The hunter path resolves to the same value as before.
- **Ran locally at the PR head:** `npx tsc --noEmit` clean; `tests/druid_wand.test.ts` (9 tests), `tests/form_swing.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/parity`, `tests/combat_auto_attack.test.ts`, `tests/fiesta.test.ts`, `tests/chat.test.ts`, `tests/world_api_parity.test.ts`, `tests/guide.test.ts`: all green, and the parity goldens confirm no regen was needed. Scanned the added lines for em/en dashes and emojis: none.

The test suite is decisive: the point-blank wand behavior is pinned as an explicit choice, the `form_swing` control moving to bear form is explained in place, and the pure-resolver matrix covers the non-druid wand carriers. The fiesta bot follow-up (routing the standoff through the same gate) was the right call even though bots never shapeshift today.

One non-blocking design note, maintainer's call: in the classic-era reference, druids are the one caster that cannot use wands (they melee with the staff in caster form). This PR deliberately diverges for QoL consistency with the other casters, and the PR body owns that choice; flagging it only so the divergence from the fidelity checklist is a recorded decision rather than an accident.

ryan-foo:
@EnriqueGF thanks for the consumer-coverage and blocklist verification. On the design note: recorded as a deliberate divergence, druids keep no wand in the classic-era reference and this PR consciously trades that for QoL parity with the other casters (nature-school identity, form-gated); the PR body owns it and this comment makes it a written decision against the fidelity checklist rather than an accident.

PR #2008, opened 2026-07-16
feat(ui): pin a basic Attack row at the top of the spellbook

ryan-foo:
## Summary

Adds the basic Attack (auto-attack) back to the spellbook as a pinned first row, the way classic spellbooks lead with Attack. Until now, a player who right-clicked the slot-0 Attack button off their bar had to know about the Interface option to get it back; the spellbook now always shows Attack with a +/- toggle that restores or removes the slot-0 button. The toggle routes through the same `showAttackButton` Interface setting the options window and the slot's right-click already drive, so all three controls share one state, and the spellbook's per-frame refresh keeps the toggle in sync if the setting flips while the window is open (elided to state flips, per the write-elision doctrine).

Implementation follows the pure-core + thin-painter split: `spellbook_view.ts` carries a new `attackOnBar` input/output beside the ability rows (Attack never becomes a fake ability row), and `spellbook_window.ts` renders the pinned row from it. The row reuses the existing `abilityUi.actionBar.attackName` / `attackTooltip` keys, the `hudChrome.spellbook` add/remove aria pair, and the existing procedural Attack icon: no new player strings, so no i18n gate is touched. The row deliberately has no aria-label override; its localized name + summary text is the accessible content, and the toggle carries the spoken add/remove name with `aria-pressed`.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/spellbook_view.test.ts tests/spellbook_window.test.ts` (the attackOnBar passthrough incl. never-a-fake-row, plus source-guard pins: row rendered first from the view state, key reuse, setAttackOnBar routing, per-frame attack-toggle sync)
  - `npx vitest run tests/architecture.test.ts tests/hud_perf_budget.test.ts` (pure-core registration and the per-frame budget both green)
  - `npm run gate`: the only failures are the machine-environmental `tests/deploy_watchdog.test.ts` (fails identically on untouched checkouts) and two load flakes (`mail`, `sfx_gate_preflight`) that pass serially.
- Manual steps: captured the window in the real offline game via the new `scripts/spellbook_attack_shot.mjs` (needs `npm run dev`), before and after.

## Screenshots

Before (no Attack in the book):

before

After (pinned Attack row with its toggle):

after

### Quality

- [x] **The full gate passes** modulo the pre-existing environmental failure and serial-green load flakes listed above. Decisive tests added for the changed behavior.

### Cross-platform

- [x] The row reuses the existing `.spell-row` / `.spell-hotbar-toggle` markup and styles, so the mobile touch-target and layout behavior is identical to every other spellbook row.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (every rendered string is an existing `t()` key; aria labels go through `t()`; no overlay or catalog edits).

### Hygiene


EnriqueGF (commented):
Reviewed from a local worktree of this branch, with the checks run rather than assumed.

## What I reviewed

- Full diff: the pinned Attack row in `spellbook_window.ts`, the `attackOnBar` passthrough in the pure core `spellbook_view.ts`, the two-line deps wiring in `hud.ts`, the screenshot script, and the tests.
- Invariants: no `src/sim/` or server changes, no IWorld facet changes (parity suite still green, see below), UI-only. No em/en dashes or emojis anywhere in the diff (scanned). No secrets, no `ALLOW_DEV_COMMANDS`.
- i18n: verified all four reused keys exist in the catalog (`abilityUi.actionBar.attackName` / `attackTooltip` in `abilities.ts`, `hudChrome.spellbook.addToBarAria` / `removeFromBarAria` in `hud_chrome.ts`); no new player strings, no overlay edits. The i18n N/A claim in the PR body checks out.
- Seam correctness: `setAttackOnBar` routes through `optionsHooks.settings.set('showAttackButton', ...)`, byte-identical to what the slot-0 right-click (`hud.ts`) and the options window already do, and the HUD reads the setting live per frame, so all three controls genuinely share one state. The click handler's synchronous `settings.set` then `refreshHotbarControls()` re-read is correct (the store is synchronous). The per-frame refresh elision (rewrite only when `aria-pressed !== onBar`) matches the existing ability-toggle discipline in the same method, and the ability-toggle loop cannot pick up the new button (it guards on `dataset.abilityId`).
- `hud.ts` growth is 4 lines of deps wiring into an existing seam object, which is the sanctioned thin-consumer shape, not coordinator growth.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/spellbook_view.test.ts tests/spellbook_window.test.ts tests/architecture.test.ts tests/hud_perf_budget.test.ts`: all pass (91 tests).
- `tests/localization_fixes.test.ts`: passes (29 passed, 3 skipped) after `npm run i18n:gen`; the initial failure was only the missing generated `i18n.status.json`, not this PR.
- `tests/world_api_parity.test.ts`: 246 passed.

## Findings

1. **Merge conflict with the current base (needs a rebase/merge, resolution is trivial).** `release/v0.27.0` has moved 23 commits since this branch forked, including the HUD refactor that extracted `actionBarController`. `git merge-tree` shows one conflict, in the spellbook deps object in `hud.ts`: this PR's `attackOnBar`/`setAttackOnBar` lines vs release's `hasFreeSlot: () => this.actionBarController.hasFreeSlot()`. Both sides should be kept; `attackSlotIsAttack()` and the `optionsHooks?.settings.set('showAttackButton', ...)` pattern still exist unchanged on the release tip, so the new lines carry over as-is. Please merge/rebase onto current `release/v0.27.0` and re-run the spellbook tests.

2. **Nit, non-blocking:** when the class kit is empty (`view.empty`), the "empty" message now renders below the pinned Attack row, which reads slightly odd. Likely unreachable for real classes; fine to leave.

3. **Nit, non-blocking, pre-existing:** if the options window is open on the Interface panel while the spellbook toggle flips `showAttackButton`, the options checkbox can go stale until the panel rebuilds. The slot-0 right-click already has this exact behavior, so it is not a regression introduced here.

Verdict: the implementation is correct and follows the pure-core + thin-painter recipe; approving once the branch is updated onto the current release tip (finding 1) would be a formality. Leaving as a comment review until then.

ryan-foo:
Finding 1 done: release/v0.27.0 is merged in and the one hud.ts conflict resolved exactly as you read it, keeping both sides (the release's actionBarController.hasFreeSlot plus the attackOnBar/setAttackOnBar wiring, which carries over unchanged since attackSlotIsAttack and the showAttackButton setting path are intact on the tip). spellbook_view, spellbook_window, architecture, and hud_perf_budget suites re-run green on the merged head with tsc clean. The two nits are acknowledged as-is: the empty-kit ordering is unreachable for real classes, and the options-panel staleness is the pre-existing slot-0 right-click behavior. Thanks for the thorough seam verification.

PR #2009, opened 2026-07-16
fix(heroic): pay Heroic Marks to every locked participant, mail the absent

ryan-foo:
## Summary

A heroic final-boss kill struck the daily raid lockout on the whole instance roster (membership-based) but only paid Heroic Marks to the death-time proximity snapshot (players within `PARTY_XP_RANGE` = 80 yd of the corpse). So a back-line healer, or a raider who died and waited to release across the arena, took the lockout **with no marks**: locked out for the day without the reward. This is what you hit on Heroic Nythraxis (lockout, no marks); the gap is general to every heroic dungeon and the raid.

This aligns the marks with the lockout: **every player freshly locked by the kill earns the marks.** Delivery splits on presence at the corpse:
- **Present at the corpse to loot** -> marks go straight to bags (unchanged behavior).
- **Locked from afar** (a distant healer, a fallen or released raider) -> the marks are **posted to the Ravenpost** via a new Heroic Quartermaster reward letter, so a distant participant never eats the daily lockout without the reward.

### Changes
- `awardHeroicMarks` (`src/sim/instances/dungeons.ts`): pay every freshly-locked recipient; bags if they were in the death snapshot, mail otherwise.
- New `mailHeroicMarks` `SimContext` seam bound to the `PostOffice`, and the `heroic_marks_reward` authored letter (`src/sim/content/letters.ts`) carrying the mark stack as its attachment.
- i18n: English letter plus the five non-Latin M16 fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU).

## Type of change

- [x] Bug fix

## How was this tested?

Test-first. New case in `tests/dungeons.test.ts`: a heroic raid kill with four raiders on the boss and one healer held past `PARTY_XP_RANGE` asserts the healer takes the `nythraxis_boss_arena:heroic` lockout AND receives 3 marks by mail.

- **Fails before the fix** (pre-fix code mails and banks nothing for the far healer):
  ```
  AssertionError: expected +0 to be 3
   ❯ tests/dungeons.test.ts  expect(mailedMarks).toBe(3)
  ```
- **Passes after the fix**: `tests/dungeons.test.ts` 53/53 green.

Commands:
- `npx vitest run tests/dungeons.test.ts` (repro + the existing "pays marks to every raider" case)
- `npx vitest run tests/sim_context.test.ts tests/entity_roster.test.ts tests/mail.test.ts tests/localization_fixes.test.ts` (new seam member, mail, S3 guard)
- `npx vitest run tests/i18n_completeness.test.ts tests/localization_coverage.test.ts` (M16 + coverage)
- `npx vitest run tests/parity tests/architecture.test.ts` (no rng draw-order change; sim purity)
- `npx tsc --noEmit` (clean for touched files)

## Checklist

- [x] Bug-fix PR shows the test failing on pre-fix code (evidence above).
- [x] Sim behavior change covered by a test; parity gate green (the fix draws no rng, so no golden changed).
- [x] New `SimContext` callback is append-only and bound in `buildSimContext`; the fake hosts in the seam tests were updated.
- [x] i18n: new letter is English-source with the five non-Latin fills in the same change (M16).


FernandoX7 (commented):
The mechanics here land the way this repo wants them to: `mailHeroicMarks` is an append-only `SimContext` member bound to the `PostOffice` exactly like `queueQuestLetter`, both seam pins were updated, the raid test is genuinely failing-first, and its `toBe(3)` pins the real `nythraxis_boss_arena` tuning literal. I verified the mail path end to end: `book` is deterministic (counter id plus `ctx.time`, no rng, parity digest unchanged), `letterId` rides the persisted message and the wire (`MailMessageView.letterId`) so the client re-localizes it like the welcome and quest letters, attachment mail never expires, and a full mailbox never refuses an authored letter. At the PR head, `tests/parity`, `tests/architecture.test.ts`, `tests/dungeons.test.ts`, the localization guards, and `tsc --noEmit` are all green, and re-running the i18n generator reproduces the committed bundles with a zero diff.

The payment rule the diff actually implements is wider than the PR body says, though, and two existing tests now document the opposite of what the code does while staying green.

- blocking: an uncredited final-boss death now mails full marks to every locked player. `heroicRewardRecipients` defaults to `[]` and is only filled in the credited block (src/sim/combat/damage.ts:859, :897), and `awardHeroicMarks` still runs on a source-less kill (tests/dungeons.test.ts:1488 proves it by asserting the lockout strikes). With `presentIds` empty, every lockout recipient falls into the new mail arm (src/sim/instances/dungeons.ts:739), including players standing at the corpse, who would get mail instead of bags, which suggests this path was not considered. Pre-fix an uncredited kill paid nobody, and the guarding test says so in its title ("no marks, no credit") but only asserts boss loot and lockouts, so it stays green through the flip. Credited kills always have at least one recipient (`eligible.push(meta)`, damage.ts:897), so `recipients.length === 0` cleanly identifies the uncredited case: either skip payment there (keeping the lockout) or explicitly adopt pay-on-uncredited, and either way make the test at :1488 assert the bags-plus-mail outcome it claims.

- should-fix: the participation model widened past the PR's framing, against documented intent. `instanceLockoutMetas` includes every current member of the owning group "wherever they stand" (src/sim/instances/dungeons.ts:671), so a door-camper or a member parked in town now earns marks by mail, and the deed daily circuit counts them too. The existing camper test says "marks stay participation-gated: only the nearby leader is paid" (tests/dungeons.test.ts:1347) and asserts only bags, which still passes while the camper is now mailed 1 mark. If "locked means paid" is meant to include never-entered members, update that test (and the released-member and side-wing cases nearby) to assert the mailed count and fix the prose, and note the consequence in the PR body: roster membership alone becomes passive daily mark income for parked alts. If not, gate the mail arm on having entered the instance this run; the reported healer and released-raider cases get paid either way.

- should-fix: the letter is missing from the twin registry `LETTERS_BY_ID` (src/ui/entity_i18n.ts:157). The committed bundles make `tEntity` resolve today, but the canonical-English fallback for this letter degrades to the raw id `heroic_marks_reward` instead of English, and the entity translation manifest (which `tests/localization_coverage.test.ts` derives its expectations from) never enumerates the letter, so coverage stays green without actually guarding it. One-line add next to `WELCOME_LETTER`, then re-run the i18n gen to confirm the three keys reproduce.

- should-fix: only the Nythraxis raid test pins a mailed count; the 5-man mail arm has no positive assertion. The camper, released-member, and side-wing cases all drive locked-but-far players through `mailHeroicMarks` but assert only bags. Once the participation question is settled, at least one 5-man case should pin the mailed count (or its absence).

- nit: `mailHeroicMarks`'s guard (`!meta || count <= 0`, src/sim/mail/post_office.ts:492) is untested, and there is no direct pin that an already-locked absent recipient is not re-mailed (the shared `alreadyLocked` guard covers it, but the mail arm inherits it untested).

Two logistics notes:

- The branch carries the main-only #1968 desktop-build commit (7c10f280e, scripts/electron-build.mjs) into release/v0.27.0, since it was cut from main and retargeted by merge. The content is identical to main so it will dedupe later, but confirm you want #1968 riding into the release branch; otherwise rebase the marks commits onto release/v0.27.0.

- The only true merge conflict is src/ui/i18n.resolved.generated/pending.ts, the standard stale-base regen churn (release advanced after the retarget merge; the slice at the head is properly sorted mechanical output). Merge release/v0.27.0 again and re-run the i18n build.

ryan-foo:
@FernandoX7 thanks for the thorough review, all findings addressed through 5d6701a94. The uncredited-kill flip is fixed: recipients.length === 0 now short-circuits payment entirely (no bags, no mail) while the lockout still strikes, and the test at that path now positively asserts the outcome its title claims. On the participation question I tightened rather than adopting pay-on-membership: the mail arm requires having ENTERED the instance this run, tracked by a session-only enteredBy set on InstanceSlot (stamped in enterDungeon, cleared with the claim, same lifecycle as clearedBy), so a door-camper or parked alt takes the lockout with no pay while the far healer and released raider still get their letter; the deeds daily telemetry follows successful payments only, and the camper/released/side-wing tests now pin the mailed counts positively (including the exact letter-plus-attachment end to end for the 5-man). The letter is registered in LETTERS_BY_ID next to the welcome letter, the coverage manifest now enumerates it (count formula updated), and the i18n generator reproduces the bundles with a zero diff on a second run. The mailHeroicMarks guard and the alreadyLocked no-re-mail case both have direct pins now. Logistics: re-merged release/v0.27.0 (only the generated pending slice conflicted, regen resolved), and the #1968 desktop commit stays deliberately, it is byte-identical to main and dedupes on the next main/release sync.

FernandoX7 (approved):

PR #2010, opened 2026-07-16
fix(ui): merge reconnecting party member's combat-meter row

jgyy:
## Summary

The Dmg/Heal/Threat combat meter (`src/ui/meters.ts`) keys each row by the numeric
source entity id (`ev.sourceId`) of the damage/heal event, not by a stable
character/account identity. When a party member relogs mid-encounter, the server
issues their character a new entity id, so the meter kept the stale-id row and
started a brand-new one under the new id for the same player: the same caster
showed up twice, with their total damage/healing split across two bars instead
of one.

`MeterData.tally()` now checks for an existing row with the same display name
under a different pid before allocating a new one, and re-keys that row onto the
current pid instead of creating a duplicate, so a reconnecting caster keeps a
single running total across the encounter and the session ("All") segment.

### Before: duplicate row on reconnect

```mermaid
sequenceDiagram
    participant Server
    participant Meters as MeterData.tally()
    participant Panel as Meters panel

    Server->>Meters: damage event, sourceId=2, name="Pal"
    Meters->>Meters: tallies.get(2) -> new row {pid:2, name:"Pal", dmg:20}
    Panel->>Panel: renders row "Pal" 20 dmg

    Note over Server: Pal disconnects and relogs<br/>server assigns a new entity id (3)

    Server->>Meters: damage event, sourceId=3, name="Pal"
    Meters->>Meters: tallies.get(3) -> MISS, creates new row {pid:3, name:"Pal", dmg:30}
    Panel->>Panel: renders TWO rows: "Pal" 20 dmg AND "Pal" 30 dmg
```

### After: reconnect merges into the existing row

```mermaid
sequenceDiagram
    participant Server
    participant Meters as MeterData.tally()
    participant Panel as Meters panel

    Server->>Meters: damage event, sourceId=2, name="Pal"
    Meters->>Meters: tallies.get(2) -> new row {pid:2, name:"Pal", dmg:20}
    Panel->>Panel: renders row "Pal" 20 dmg

    Note over Server: Pal disconnects and relogs<br/>server assigns a new entity id (3)

    Server->>Meters: damage event, sourceId=3, name="Pal"
    Meters->>Meters: tallies.get(3) -> MISS
    Meters->>Meters: scan rows for name="Pal" under a different pid -> found (pid 2)
    Meters->>Meters: re-key row 2 -> 3, keep accumulated dmg, add 30
    Panel->>Panel: renders ONE row: "Pal" 50 dmg
```

## Related issues

N/A (reported via screenshot, filing this PR directly per the reporter's request).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/meters.test.ts`
  - `npx tsc --noEmit -p .`
  - `npm run gate` (full CI-equivalent gate; two pre-existing failures unrelated
    to this change: `tests/deeds_window_focus.test.ts` (missing `localStorage`
    global in that test's env, reproduces identically on a clean
    `release/v0.27.0` checkout) and a `tests/server/http/parity.test.ts` timing
    flake)
- Manual steps: added a new unit test reproducing the exact reconnect scenario
  (same display name, damage split across two `sourceId`s within one encounter)
  and confirmed it fails without the fix and passes with it.

## Screenshots / recordings

No DOM/CSS change: the meter's rendering and layout are unchanged. This is a
pure aggregation-key fix in the data model behind the panel, reproducible only
via a live multiplayer reconnect (a fresh server-assigned entity id issued to
the same character mid-encounter), which isn't practical to capture as a static
screenshot. Covered instead by the new `tests/meters.test.ts` case:
"merges a reconnecting party member into one row instead of duplicating them
under their new pid", and by the sequence diagrams above.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green modulo the two
      pre-existing, unrelated failures noted above. Added a decisive regression
      test for the fixed behavior.

### Cross-platform

- [x] N/A — no DOM/CSS/layout change; the fix is internal to `MeterData.tally()`.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/ui/meters.ts
+++ b/src/ui/meters.ts
@@ -68,15 +68,26 @@ export class MeterData {
     this.allTime = { ...newEncounter(now), label: 'All (session)' };
   }
 
-  private tally(enc: Encounter, pid: number, name: string, cls: string | null): MemberTally {
+  private tally(
+    enc: Encounter,
+    pid: number,
+    name: string,
+    cls: string | null,
+    partyPids: Set<number>,
+  ): MemberTally {
     let t = enc.tallies.get(pid);
     if (t) return t;
     // a reconnect issues the same character a new entity id mid-encounter; find
     // its previous row by name and re-key it instead of starting a duplicate.
+    // Only treat a name match as a reconnect when the old pid is no longer a
+    // live party member: pet names come from their template/tamed-target name
+    // and are not unique, so two live same-named pets must stay separate
+    // rows instead of ping-ponging the merge back and forth.
     for (const [oldPid, existing] of enc.tallies) {
-      if (existing.name === name && oldPid !== pid) {
+      if (existing.name === name && oldPid !== pid && !partyPids.has(oldPid)) {
         enc.tallies.delete(oldPid);
         existing.pid = pid;
+        existing.cls = cls ?? existing.cls;
         enc.tallies.set(pid, existing);
         return existing;
       }
@@ -107,7 +118,7 @@ export class MeterData {
         const cls =
           member?.cls ?? (ev.sourceId === world.player.id ? world.player.templateId : null);
         for (const enc of [this.current, this.allTime]) {
-          const t = this.tally(enc, ev.sourceId, name, cls);
+          const t = this.tally(enc, ev.sourceId, name, cls, partyPids);
           t.dmg += ev.amount;
           if (enc === this.current) {
             t.dmgByMob.set(ev.targetId, (t.dmgByMob.get(ev.targetId) ?? 0) + ev.amount);
@@ -128,7 +139,7 @@ export class MeterData {
       const name = member?.name ?? src?.name ?? `#${ev.sourceId}`;
       const cls = member?.cls ?? (ev.sourceId === world.player.id ? world.player.templateId : null);
       for (const enc of [this.current, this.allTime]) {
-        this.tally(enc, ev.sourceId, name, cls).heal += ev.amount;
+        this.tally(enc, ev.sourceId, name, cls, partyPids).heal += ev.amount;
       }
     }
   }
--- a/src/ui/meters.ts
+++ b/src/ui/meters.ts
@@ -70,10 +70,19 @@ export class MeterData {
 
   private tally(enc: Encounter, pid: number, name: string, cls: string | null): MemberTally {
     let t = enc.tallies.get(pid);
-    if (!t) {
-      t = { pid, name, cls, dmg: 0, heal: 0, dmgByMob: new Map() };
-      enc.tallies.set(pid, t);
+    if (t) return t;
+    // a reconnect issues the same character a new entity id mid-encounter; find
+    // its previous row by name and re-key it instead of starting a duplicate.
+    for (const [oldPid, existing] of enc.tallies) {
+      if (existing.name === name && oldPid !== pid) {
+        enc.tallies.delete(oldPid);
+        existing.pid = pid;
+        enc.tallies.set(pid, existing);
+        return existing;
+      }
     }
+    t = { pid, name, cls, dmg: 0, heal: 0, dmgByMob: new Map() };
+    enc.tallies.set(pid, t);
     return t;
   }
 
--- a/tests/meters.test.ts
+++ b/tests/meters.test.ts
@@ -139,4 +139,27 @@ describe('combat meters', () => {
     expect(m.current!.tallies.get(3)!.name).toBe('Wolf Pet');
     expect(m.current!.tallies.get(3)!.dmg).toBe(18);
   });
+
+  it('merges a reconnecting party member into one row instead of duplicating them under their new pid', () => {
+    const w = fakeWorld();
+    const party = new Set([1, 2]);
+    const m = new MeterData(0);
+    // "Pal" deals damage under their original pid (2)...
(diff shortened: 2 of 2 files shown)


EnriqueGF (changes requested):
The reconnect diagnosis is right, the test reproduces the exact reported scenario, and the fix is nicely scoped to `MeterData.tally()` with no wire or sim impact. Re-keying player rows by name is sound in itself: character names are unique per (realm, lower(name)) (`server/db.ts:222`, `server/db.ts:2572`), and parties are per realm. Verified locally at the PR head: `npx tsc --noEmit` clean, `tests/meters.test.ts` (6/6), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, and `tests/parity` all green.

One real problem though:

**blocking** `src/ui/meters.ts:76`: the name-match merge also fires for two live, same-named pets, collapsing rows that were correctly separate before this PR. Pets are tallied under their own entity ids (`partyPids()` at `src/ui/meters.ts:252` adds owned mobs, and the existing "controlled pet damage" test pins that), and pet names are template names, not unique: `createDemonPet` sets `pet.name = template.name` (`src/sim/pet/pet_commands.ts:343`) and taming sets `pet.name = target.name`. So a party with two warlocks running the same demon, or two hunters with the same wolf template, has two same-named pets with distinct pids. With this change, every hit from pet B finds pet A's row by name and re-keys it, and the next hit from pet A re-keys it back: both pets' damage merges into one row whose pid ping-pongs between the two entities (which also breaks the threat tab's `liveThreat.get(t.pid)` lookup for that row). I confirmed this with an executed repro: two owned mobs both named "Imp" (pids 10 and 11, both in the party set) dealing 20/30/5 end up as a single 55-damage row instead of two rows, which is a regression from base.

Suggested fix: only treat a name match as a reconnect when the old pid is actually gone. Pass `partyPids` into `tally()` (both call sites already have it in scope in `onEvent`) and guard the merge with `!partyPids.has(oldPid)`. In the reconnect case the stale pid has already left `partyInfo.members` and `world.entities`, so the merge still happens; with two live pets both pids are in the set, so they stay separate. Please also add the two-live-same-named-pets case to `tests/meters.test.ts` so the guard is pinned.

Minor, non-blocking:

- nit `src/ui/meters.ts:79`: the merged row keeps the old `cls`. Harmless for a same-character reconnect, but since you have `cls` in hand you could refresh it (`existing.cls = cls ?? existing.cls`) for free.

Happy to re-review once the pet guard is in.

jgyy:
Thanks for the sharp catch and the executed repro, @EnriqueGF. Fixed exactly as suggested:

- `MeterData.tally()` now takes `partyPids` (both call sites in `onEvent` already had it in scope) and only treats a name match as a reconnect when the old pid is no longer in the live party set (`!partyPids.has(oldPid)`). A genuine reconnect's stale pid has left `partyInfo.members` / `world.entities` by the time the new pid's first event lands, so that path still merges; two live same-named pets (or players) both stay in the set and now stay in separate rows instead of ping-ponging.
- Also took the non-blocking nit: the merged row now refreshes `cls` (`existing.cls = cls ?? existing.cls`).
- Added `tests/meters.test.ts`: "keeps two live same-named pets in separate rows instead of merging them by name" (two owned mobs both named "Imp", pids 10/11, dealing 20/30/5 -> two rows, 25/30), which fails without the guard and passes with it. All 7 cases in the file pass, plus `tsc --noEmit`, `tests/architecture.test.ts`, and `tests/localization_fixes.test.ts`.

Pushed to `fix/caster-duplicate-meter`. Ready for re-review.

EnriqueGF (approved):
Reviewed both commits at the PR head (f4192e7) and ran the checks locally.

What I verified:

- The regression pin is decisive: I swapped in the release/v0.27.0 version of `src/ui/meters.ts` and the new reconnect test fails there, then passes with the fix.
- `npx tsc --noEmit` clean; `tests/meters.test.ts` 7 passing; the guard suites (`tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`) all green.
- The `partyPids` gate matches the real caller: `Meters.partyPids()` includes owned pets via the `ownerId` scan, so the live same-named-pet guard holds in production the same way the test models it.
- The ordering assumption behind the merge holds: `ClientWorld` applies party snapshots immediately on message arrival while events drain from a queue on the next HUD frame, so by the time the first damage event under the new entity id reaches `tally()`, the roster no longer contains the old pid.
- Character names are unique per (realm, lower(name)) (`server/db.ts:2572`), so the name-keyed merge cannot fold two distinct players together.
- No sim, server, or wire changes; no new player-visible strings; no banned characters in the diff. The branch trails the current `release/v0.27.0` tip slightly but merges cleanly (verified with `git merge-tree`).

One finding:

- nit, `src/ui/meters.ts:86`: in the session "All" segment the merge can still cross pet owners. Once warlock A's "Imp" row's pid has left the party set (pet despawned), warlock B's freshly summoned "Imp" absorbs that row, folding two owners' totals into one bar where the old behavior showed two identically labeled bars. Since rows are keyed by display name only, nothing distinguishable is misattributed on screen, and merging a same-owner resummon is arguably the desired behavior, so no change needed; noting it in case per-owner pet rows ever matter.

Tightly scoped fix, decisive test, and the second commit's pet guard closes the ping-pong hole the first version had. Approving.

PR #2011, opened 2026-07-16
feat(merch): homepage merch store PRD and fail-closed skeleton (Claudium, Stripe, Solana rails; Printful dropship)

awidearray:
## Summary

A merch store off the main website: physical goods (shirts, hoodies, mugs) dropshipped by Printful, payable with Claudium, card (Stripe), or Solana (SOL now; the quote shape carries `mint`, so USDC is an enable, not a rework). This PR ships the PRD plus the full game-repo half as a structural twin of the Claudium surface, fail-closed behind `MERCH_STORE_ENABLED` (off by default): with the flag off the site renders byte-identically and every route answers its typed unavailable body. The economy service's merch half (catalog, pricing, payment verification, Printful, PII) is deliberately out of scope and specced as a six-endpoint contract in `docs/prd/woc/merch-store.md`.

Architecture (mirrors the claudium twin on this base):

- The game server stays a thin BFF: it computes no price, total, or verification. `server/merch_proxy.ts` forwards to the economy service over the same `WOC_ECONOMY_SERVICE_URL` / `WOC_ECONOMY_INTERNAL_SECRET` pair and fails closed (typed unavailable results, never a throw). Service responses are validated with `claudiumPurchase`-grade strictness before `ok:true`: no refusal reason, positive totals, and a well-formed rail leg for the requested rail.
- `server/merch.ts` registers the `/api/merch/*` family both as RouteDefs (`server/http/registry.ts`) and on the legacy `handleApi` ladder (`server/main.ts`) per the dual-edit invariant. The two money mutations (checkout, native confirm) carry the claudium-style two-tier rate limits (pre-auth per-IP, post-auth fused IP+account) on BOTH arms; the public catalog probe is per-IP limited and TTL-memoized (30s) so homepage traffic can never become an upstream flood. Webhook relays (`stripe`, `printful`) key on the service env pair alone, NOT the kill switch, so flipping the store off never drops a payment or fulfillment callback for an order placed while it was on (the rollback story in DEPLOY.md).
- Non-custodial: the SOL rail reuses the claudium `quote -> wallet-sign -> confirm` idioms (`openStripeCheckout` with `onComplete`, `signAndSendTransactionBase64`, `confirmNativeSettlement`'s retry ladder, `newIdempotencyKey`). Post-broadcast confirmation is bounded by the ladder's own recovery window, never the quote's wall-clock expiry (the twin's exact policy). The Claudium rail settles synchronously via the service's idempotent debit.
- No PII game-side: the shipping address passes through the checkout POST once, unlogged and unstored; the service owns fulfillment data.
- Homepage: a Store nav tab plus `#store-view` that stay hidden unless the public `GET /api/merch/products` probe answers `enabled:true`. `src/ui/merch_view.ts` is the pure core (registered in `UI_PURE_CORES`; owns the catalog/cart/rail projection, the checkout-outcome decision, and shipping completeness) with `src/ui/merch_store_page.ts` as the thin DOM consumer. The Stripe rail shows an honest payment-pending notice until the embedded checkout reports completion, never a premature "placed". i18n: English at source plus the five non-Latin fills for the wordy keys (M16).

## Type of change

- [x] Tests
- [x] Documentation

## How was this tested?

- Commands: `npm run gate` green end to end (i18n gen + freshness, malware scan, changed-files biome, SFX conformance, full vitest at 1142 files / 14.4k tests including the new `tests/merch_proxy.test.ts`, `tests/merch_sdk.test.ts`, `tests/merch_view.test.ts`, `tests/server/merch_routes.test.ts`, browser regressions, `tsc`, env/server/client builds).
- New coverage beyond the happy path: the SOL rail end to end through the REAL `startMerchCheckout` + `confirmNativeSettlement` with injected clocks (retry ladder walk, post-expiry retry, non-retryable refusal, no-wallet and no-native-leg guards); two-tier rate limits on both dispatch arms (pre-auth flood, fused IP+account, legacy-core fused limiter, `rateLimitApplied` skip); webhook relays at the route layer (verbatim raw body + signature header, service 400 mapping, 1 MiB cap, flag-off rollback exception); checkout body boundaries (`MAX_CART_ITEMS`, `MAX_ITEM_QTY`, malformed JSON, non-string shipping); service-response validation (reason present, rail mismatch, malformed legs, non-positive totals); the products TTL memo; the checkout-outcome decision table; the status-key-to-catalog lockstep pin.
- Manual steps: drove the whole stack live against a stub economy service implementing only the PRD contract (real game server + Postgres + built client in a real browser). Store off: byte-identical nav, typed `enabled:false` products, 401s on authed routes, webhooks refusing. Store on: tab reveal, catalog, cart math display, shipping form, rail picker, a logged-in Claudium checkout settling synchronously (order placed, cart cleared, orders list showing the paid order), idempotent checkout replay, out-of-stock refusal, SOL intent + confirm, and the checkout rate limit tripping at the 11th call in a minute.

## Screenshots / recordings

Store off (the default; byte-identical to today):

before desktop

Store on (flag + service): tab reveal, catalog, cart, and a settled Claudium order:

after home
catalog
cart
order placed

Mobile:

before mobile
catalog mobile

## Open questions for review (defaults chosen, cheap to flip)

1. Separate merch Stripe webhook endpoint vs multiplexing the claudium one (the service can route by PaymentIntent metadata)?
2. Claudium price per variant: service-set merchandising (assumed) vs a fixed peg conversion?
3. Regulatory/ToS pass (consumer law, tax jurisdictions, crypto-for-goods) before the service half ships.


EnriqueGF (changes requested):
Reviewed the full diff, read the claudium twin side by side, and ran the checks locally. The architecture here is genuinely good: the BFF really does compute no money, the fail-closed posture is consistent, the account id is always token-resolved on both dispatch arms, shipping PII passes through unlogged and unstored, and the dual-arm registration (RouteDef + legacy ladder) is complete across the registry, surface inventory, and content-type gates. The two-tier limiter wiring is a faithful copy of the claudium shape.

Three things I think need to change before merge, all verified against the real code.

## 1. The SOL rail is broken on the desktop shell (high)

`server/claudium_proxy.ts:506` calls `desktopWalletHandoffs.authorizeTransaction(...)` on every successful native quote and refuses the quote if it throws. `server/merch_proxy.ts` builds an identical `native` leg and never calls it. `authorizeTransaction` has exactly one call site in the tree:

```
$ grep -rn "authorizeTransaction" server/ src/ | grep -v test
server/desktop_wallet_handoff.ts:55:  authorizeTransaction(
server/desktop_wallet_handoff.ts:169:    authorizeTransaction(accountId, authorization) {
server/claudium_proxy.ts:506:    desktopWalletHandoffs.authorizeTransaction(input.accountId, {
```

The merch client path depends on it: `src/main.ts` calls `authorizeDesktopWalletInBrowser({ kind: 'transaction', reference, expectedAddress })`, which reaches `desktop_wallet_handoff.ts:189` `createTransaction`, which looks up `authorizedTransactions.get(accountId:reference)` and throws `'transaction is not backed by an authorized Claudium quote'` when the entry is absent. It will always be absent for merch.

Repro: Electron shell, linked wallet, SOL rail. The checkout POST succeeds, the service creates an order and a Solana quote, then the handoff refuses. Nothing is charged, but the order and quote are orphaned upstream and the player gets a generic wallet error. 100% failure rate on that rail on desktop. The in-browser Wallet Standard path is unaffected, which is why a browser-only smoke test passes it, and the manual test notes in the PR body describe exactly that browser path.

## 2. `merch/checkout/native/confirm` drops the resolved account (medium)

`server/claudium.ts:297` sends `{ accountId, reference, signature }` and `claudium_proxy.ts:548` types `accountId` as required. `server/merch.ts:246-253` resolves `accountId` through `activeGuard` and then discards it, forwarding only `{ reference, signature }`; the PRD contract table (line 34) omits it too, so the spec and the code agree with each other but both diverge from the twin.

That removes the only ownership signal the service could scope on. Whether it is exploitable beyond resource cost depends on the service repo, but the service cannot enforce a binding it is never given, and the resource cost is real on its own: `MERCH_CONFIRM_MAX_PER_MINUTE` is 60 and each call holds an upstream request for up to `NATIVE_CONFIRM_TIMEOUT_MS` (60s), so one account can hold ~60 concurrent minute-long sockets against references it does not own. The claudium twin's per-account binding is what makes the identical 60/min budget meaningful. Since the service half is not built yet, this is the cheap moment to fix the contract in both the code and the PRD table.

## 3. Test gaps on the load-bearing claims (medium)

I mutation-tested these; each mutation below left the suite green (tree restored afterward):

- **The legacy dispatch arm is unpinned.** Making `server/main.ts:2095` pass `{ rateLimitApplied: true }` into `handleMerchApi`, which leaves the legacy money path entirely unrate-limited, is undetected. So is deleting the `merchProductsRateLimited` guard at :2084. The tests drive the shared core and the exported helper, never the wiring. The PR's "two-tier rate limits on BOTH dispatch arms" claim is only half-protected, and the unprotected half is the rollback path.
- **The rate ceilings are a constant-self-comparison.** `tests/server/merch_routes.test.ts:37-38` imports `MERCH_CHECKOUT_MAX_PER_MINUTE` and uses it as its own loop bound, so both sides move together. I raised checkout from 10 to 500 and everything stayed green. These are monetary throttles; a literal pin (`expect(MERCH_CHECKOUT_MAX_PER_MINUTE).toBe(10)`) next to the table is the cheap fix.
- **Missing per-dimension negatives in `merchCheckout` validation.** Deleting the `orderId !== null` clause (`merch_proxy.ts:397`) is undetected, so a service answering an intent with no order id would be painted as a placed order. The `!paid` guards on the stripe and sol arms (:391, :393) are undetected. `amountBase` and `destination` on the native leg are undetected, because the one sol negative clobbers three fields at once and only pins the last. An `it.each` clobbering exactly one field per case covers all of it.
- Smaller: `merchOrders` response validation has no test at all (replacing the whole filter predicate with `true` is green); the 1 MiB webhook cap has only the refuse arm, so lowering `WEBHOOK_MAX_BYTES` to 1024 and silently rejecting real Stripe events is green; the TTL memo pins the mechanism but not the 30s window.

## Smaller notes, not blocking

- **The products memo caches failures.** `merch_proxy.ts:259-267` writes `productsCache` unconditionally, so one transient upstream 500 or timeout hides the Store nav item and empties the catalog for every visitor for the full 30s. Caching only the `enabled:true` result, or a short negative TTL, matches the header comment's fail-closed-but-recover-fast intent better.
- **Currency formatting.** `merch_store_page.ts:139` hardcodes the symbol as `` `$${formatNumber(...)}` ``. Symbol placement is locale-dependent (fr renders `12,34 $`), and the repo rule is money through `formatMoney`/`Intl`. Same line of thought for the hardcoded `x` quantity multiplier at :195, which is a Latin letter rendered untranslated.
- **No length caps on the shipping block or email.** `merch.ts:103-131` shape-checks types and non-emptiness but bounds nothing, so a checkout can forward ~60KB of free text (the shared `readBody` cap) to the service. Reasonable to leave validation to the service, but a cheap length cap is consistent with the module's own stated contract.
- **`zip` is required** in both `parseShipping` and `merchShippingComplete`. Several destinations have no postal code (Ireland pre-Eircode, Hong Kong, UAE), so this will block legitimate international orders on a store advertised as shipping worldwide.
- **Base URL shape.** `.env.example` sets `WOC_ECONOMY_SERVICE_URL=.../v1/claudium/`, so merch paths resolve to `/v1/claudium/merch/products`. That matches the PRD's "same base URL" wording, but it is a deployment footgun worth a sentence in DEPLOY.md.
- **Base branch.** This targets `release/v0.27.0` while `release/v0.28.0` exists and is the latest release branch. Worth a rebase per the repo's default workflow.

## What I ran

`npx tsc --noEmit` clean. `tests/merch_proxy.test.ts`, `tests/merch_sdk.test.ts`, `tests/merch_view.test.ts`, `tests/server/merch_routes.test.ts`, `tests/architecture.test.ts`, `tests/i18n_completeness.test.ts`, `tests/world_api_parity.test.ts` all pass (363 tests). `tests/localization_fixes.test.ts` initially failed on a missing gitignored `i18n.status.json`; after `npm run i18n:gen` it passes (29 passed, 3 skipped) and the tree stays clean, so the generated i18n artifacts are fresh. No em dashes, en dashes, or emoji in the added lines. `src/sim/` is untouched, so the determinism and parity invariants are not in play.

Items 1 and 2 are the ones I would want fixed before this merges; item 3 is what keeps them from regressing later.

EnriqueGF (commented):
Follow-up on the presentation seam, which I had only spot-checked when I posted the review above. Nothing here changes the verdict (the two blockers remain the desktop SOL rail and the missing `accountId` on native confirm), and the module-first shape is clean: `merch_view.ts` is correctly registered in `UI_PURE_CORES` and passes the purity scan, nothing is per-frame, the mobile floors are all met (16px inputs, 40px touch targets, `prefers-reduced-motion` honored), and every interpolation including product names and tracking URLs goes through `esc()`.

Four things worth folding in, all verified:

**`totalLabel` is unlocalized server-supplied player text** (`src/ui/merch_store_page.ts:270`). It is rendered verbatim, and `merch_proxy.ts:457` passes it through with no shape check beyond `typeof === 'string'`. That is the root i18n invariant's "server-sent player text" case: a Japanese player sees the order total as whatever English/USD-formatted string the economy service happened to emit. The S3 guard cannot catch it because the string originates outside `server/`. Since the service half is not built yet, the cheap fix is to change the contract now: return structured `{ currency, amountMinor }` and format client-side, rather than adding a matcher rule for an opaque upstream string.

**Every interaction does a full `innerHTML` repaint** (`merch_store_page.ts:115`), including selecting a payment rail. That destroys the element the user just activated and drops keyboard focus to `<body>`. A keyboard user who Tabs to the SOL rail button and presses Enter gets silently reset to the top of the document and has to Tab through the whole catalog again to reach Place Order. For the rail buttons specifically, mutating `aria-pressed` in place instead of repainting avoids the problem entirely.

**The live regions are created together with their content** (`:103`, `:128`, `:238`). A `role="status"` node must already be in the DOM before its text changes for most screen readers to announce it. As written, a failed checkout inserts a brand-new `role="status"` element and NVDA/VoiceOver announce nothing, so a blind user gets no feedback that their order failed. One persistent notice node outside the repainted region, with only its text content written, fixes both this and part of the focus issue.

**The merch CSS is outside the layer** (`src/styles/index.extra.css:287`). `@layer index-extra {` opens at line 13 and closes at 19; the ~190 lines of merch CSS are appended after it, so they are unlayered and outrank every layered rule in the barrel. The banner also uses box-drawing characters rather than the ten-dash `/* ---------- name ---------- */` fence, which means the section is invisible to the `tests/css_corpus.test.ts` manifest and a future refactor could relocate or delete it with no guard coverage. The pre-existing Discord block in the same file has the same shape, so this follows local precedent rather than introducing the hazard, but it is worth fixing while the section is new.

Two corrections to my own earlier notes, in the PR's favour:

- On the M16 question I raised implicitly by listing the overlay edits: M16 applies here, not the "never edit the overlays" rule, so the PR is **correct** to add the five non-Latin fills. All five overlays carry exactly the same 50 `merch.*` keys plus `nav.store`; the differing added-line counts (55/55/60/59/60) are purely Biome line-wrapping on long values, not missing keys. No English is left in any non-Latin overlay.
- The hardcoded `$` prefix I flagged at `:139` matches the established precedent in `claudium_window.ts:434` and `daily_rewards_window.ts:690`, so it is consistent rather than a regression. Treat it as a family-wide cleanup, not a change to make here. Related: the play.html concern is a non-issue, `setupNavBtn` early-returns on a null button and the probe guards with `if (item)`, so the known crash class does not apply; the only cost is one wasted `/api/merch/products` request per `/play` load.

Gates run on this pass, all green: `architecture`, `styles_extraction`, `css_corpus`, `css_value_validity`, `i18n_completeness`, `i18n_emit_shape`, `merch_view`, `focus_manager`, `focus_visible_guard`, `mobile_window_coverage`, `per_entry_css_wiring`, `hud_perf_budget`, `painter_host`, plus `tsc`. The axe and keyboard-reachability arms of `npm run test:browser` and the wall-clock perf tour were not runnable in this environment; given nothing per-frame was added, the perf risk is low, but the a11y items above are exactly what that browser suite would surface.

PR #1794, opened 2026-07-12
feat(dungeons): Molten Abyss, a level-20 five-player lore dungeon under Stormcrag

EnriqueGF:
## Summary

Adds the **Molten Abyss** (id `infernal_abyss`), a level-20 five-player lore dungeon beneath Stormcrag: a 10-room authored descent through a lava maze and forge complex to Azazel the Abyssal Lord, with a three-quest lore chain, deterministic lava hazards, heroic tuning, an authored minimap, an infernal render dressing, and seven Tripo-generated models.

Everything lands behind existing seams:

- **Authored layout seam** (`src/sim/dungeon_rooms.ts` + `dungeon_layout.ts`): one room/door/decor graph drives visible walls, collisions, the render dressing, the lava hazard footprints, and the minimap schematic, so what you see is what you collide with.
- **Sim content** (`src/sim/content/infernal_abyss.ts`): five trash archetypes, two minibosses (The Forgekeeper, Pyre Golem), and the Azazel encounter built from the existing deterministic encounter vocabulary (hard casts, firestorm pulse, stomp, cinderling waves at 70/40 percent, terrify, 20 percent enrage, yells). 17 items, all budget-exact against `expectedStatBudget`, weapons on the `weaponDpsBudget` curve.
- **Lava hazards** (`src/sim/instances/infernal_abyss_hazards.ts`, behind `SimContext`): pools and fissures pulse a deterministic 8 percent max hp once a second from the exact visual footprints, on every graphics tier. No rng.
- **Lore chain**: Loremaster Caddis offers three linked quests over four interactable lore objects; each emits a localized two-line vision through the shared quest-object entry point, with party-wide credit like the Nythraxis graves.
- **Heroic**: an `HEROIC_DUNGEON_TUNING` entry calibrated with the documented equalization method (damage mult 5.1 continues the inverted ladder 6.8 / 6.2 / 5.7 / 5.4; Azazel lands ~23 percent per swing on the reference shaman). Heroic variants and marks come free from the existing machinery. The optional Azazel-specific item-level-31 heroic epics (a `HEROIC_BOSS_LOOT` extension) are left as a maintainer loot decision.
- **Instance space**: the arena band is now the finite hole [4160, 4240) instead of everything up to the delves, freeing index 6 at x=4500 (edges pinned by tests).
[...]

PR #1517, opened 2026-07-06
feat(editor): standalone map-editor build (3-axis move, terrain detach, click-cycle, Scene Collection, wireframe)

EnriqueGF:
Follow-up to the merged map editor (#1306), bringing Troy Polaczuk's standalone map-editor snapshot (troypolaczuk/world-of-claudecraft) onto the active release. #1306 was already merged, so this rides a fresh branch.

## What this adds
Five editor features plus their review fixes:
- Free 3-axis movement for every placement (assets and box/sphere/wall colliders).
- Move-tool grab detaches a placement from terrain (free-float anywhere; one Ctrl+Z restores the grounded state).
- Repeated clicks cycle the selection through overlapping objects.
- Blender-style Scene Collection panel: list every placed object, hide (editor-only), double-click rename, pin-to-focus, click-select; grass excluded.
- Camera-tab Wireframe render mode (map geometry only; editor overlays stay solid).

Review fixes: detach/undo gap on zero-net moves, click-cycle gesture routing, 2D-overlay hide, and wireframe leaking onto the selection/footprint rings.

## Rebase and conflict notes
Cherry-picked onto release/v0.22.0. Conflicts resolved:
- src/sim/world.ts: kept the base unclimbable-mountains rim (steeper, guarded by tests/terrain_walls.test.ts) but read bounds from the live world (w.halfX/minZ/maxZ) instead of the fixed WORLD_MAX_X, so custom-sized editor maps get the same rim. Identical to base for the default world.
- src/render/critters.ts and src/render/renderer.ts: merged both import sets (base and incoming).
- i18n resolved/generated files, pending, sha256 and status summary were rebuilt via the build (npm run i18n:gen + i18n:hash), not hand-merged. New editor keys live in src/ui/i18n.catalog/editor.ts (English); locales English-fill at release.

## Compliance
- Replaced the Scene Collection emoji glyphs (eye, eye-off, pin) with inline SVG stroke icons via a shared iconButton helper in src/editor/dom.ts, matching the toolbar icon convention. No emoji anywhere.
- Removed em dashes introduced in the snapshot's comments.

## Verification
- tsc --noEmit clean.
- vitest green on the editor and terrain suites: terrain_walls, new_map_size, transform_gizmo, collider_volumes, biome_paint_core, blocker_colliders, custom_map_parity, custom_paint_swatches, editor_blocker_core, editor_persist, editor_placement_transform, map_doc_sanitizer, architecture (determinism + sim purity), localization_fixes (i18n S3 guard).
[...]

PR #2012, opened 2026-07-16
feat(editor): dungeon-layout editor mode (+ Troy's newer editor, Rifts, Scorching Wastes)

EnriqueGF:
## Summary

Integration branch for the **dungeon-layout editor mode**, opened as a draft so it can be pulled and playtested. It bundles four things on top of `release/v0.27.0`, so it is large (most of the line count is Troy's imported model/asset packs, not code):

1. **Troy's standalone map editor** (PR #1517): 3-axis move, terrain detach, click-cycle selection, Scene Collection, wireframe.
2. **The Molten Abyss** authored dungeon (PR #1794): `INFERNAL_ABYSS_LAYOUT` rooms/doors/barriers/decor.
3. **New: the editor's Dungeon mode.** A parallel stage over the same live `Sim` + `Renderer` viewport that edits an authored `DungeonLayout` (rooms, doors, barriers, visual openings, decor, dais) instead of the map document, starting with the Molten Abyss.
4. **Troy's newer editor + Rifts + Scorching Wastes**, merged wholesale from his `woc-map-editor` mirror by union: Collision Master tool, IndexedDB map persistence + DEFLATE zip import, biome paint, cave gen, selection clipboard, foliage editing; the Rifts procedural-dungeon system (PR #1584); and the Scorching Wastes zone.

### How the Dungeon mode works

- Click **Dungeon** in `/editor`: the camera flies into the real interior the renderer builds (interior fog/lighting and all), and a wireframe floor-plan overlays every draft entity (rooms blue, doors cyan, maze barriers orange, pit rose, decor yellow, boss dais red).
- Click selects (click again cycles overlapping picks), drag moves with a throttled live interior rebuild, arrows nudge, Delete removes, Ctrl+Z undoes. A side panel gives per-kind numeric fields plus a model dropdown for decor and a maze/pit toggle for barriers.
- **Validate** runs a door-graph BFS reachability check plus room-bounds, floating-door and decor-blocks-door checks, and blocks saving on errors.
- **Save to source** writes the formatted literal back over `INFERNAL_ABYSS_LAYOUT` in `src/sim/dungeon_layout.ts` through a dev-only Vite middleware (`/__dungeon_editor/save`, same lifecycle as the music editor's), so the sim, renderer, colliders, lava hazards, and minimap all pick the edit up via HMR because they derive from that one datum. Outside the dev server the layout downloads instead.

Decision logic lives in three DOM-free pure cores (`dungeon_draft_core`, `dungeon_layout_writer_core`, `dungeon_validate_core`); the mode module is thin three/DOM glue. Live rebuild seam: `DungeonInteriors.disposeInterior` (keyed builds) + `Renderer.setDungeonLayoutOverride`.

## Related issues

Supersedes / bundles #1517 and #1794. Carries the Rifts work (#1584).

## Type of change

- [x] Bug fix
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands: `npx tsc --noEmit` clean; the 3 new editor-dungeon suites (24 tests) green; architecture / world_api parity / command schema / snapshot / entity roster / sim_context / delves / tick-perf suites green; parity goldens reminted (`UPDATE_PARITY=1`, 216 pass); full `vitest run` green except one Vale Cup bot-pacing statistical test that **passes in isolation** (parallel-load flakiness, not a regression).
- Manual steps: drove `/editor` in a real browser: entered Dungeon mode (Molten Abyss interior renders), overlays draw, click-select + arrow-nudge marks the layout dirty, Validate flags an unreachable room after a bad nudge and blocks save, Save writes a cosmetic-only diff to `dungeon_layout.ts`; re-verified the whole flow after the troy/main merge with zero console errors, alongside Troy's new Preview button.

## Screenshots / recordings

Dungeon mode over the live Molten Abyss interior (floor-plan overlays + side panel) captured during verification; will attach in-thread.

## Draft status: known follow-ups before this is merge-ready

- **CI will be red on inherited Biome errors.** Troy's imported test files carry pre-existing lint errors (`forEach` returning a value, etc.) that the changed-files Biome gate flags. My own files are clean. These are his mirror's debt, not this change's; needs a cleanup pass (or a scoped ignore) before merge.
- **i18n M16:** the wordy `editor.dungeon.*` strings need their five non-Latin fills.
- Port the Rifts `/dev portal` and `/dev smite` cheats into `dev_commands.ts` (dropped when our extracted dev-command module won that conflict).
- Integrate Rift-theme music into `instance_music.ts` (the HUD kept our `InstanceMusicController`).


EnriqueGF (commented):
Reviewed this as a large integration branch, not line by line over the imported asset packs. I focused on the code that is actually new here: the dungeon editor mode, the Rifts sim system, the new server routes, and the wire/parity surface. Ran locally against `release/v0.27.0`.

**What I checked and ran**
- `tsc --noEmit`: the only error is `server/email/ses_sender.ts:40` (`send` on `SESv2Client`), which this PR does not touch and which reproduces on the base; it is a missing `@smithy/smithy-client` transitive in the shared store, not a regression here.
- Guards green: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts` (IWORLD_MEMBERS pin at 241), `tests/command_schema.test.ts`, `tests/parity` (216), and the three new editor-dungeon suites.
- Determinism pass over `src/sim`: no `Math.random`/`Date.now`/`performance.now` added, sim purity clean, no `render`/`ui`/`game`/`net` imports. Rift procgen correctly uses local `Rng` streams seeded from descriptor+ordinal, isolated from the shared `ctx.rng`, so parity stays byte-identical. `camp_scatter.ts` is a faithful draw-order-preserving move. Nicely disciplined for a change this size.
- Parity/wire: the 8 new IWorld members are implemented in both `Sim` and `ClientWorld`; `rt`/`sld`/`einst` round-trip encode/decode; `fatigueTicks` is deliberately server-local and is in `blankEntity`. No drift.

The determinism and three-host discipline here is genuinely good. Findings below are mostly in the new editor tooling and dev-only middleware.

**Findings**

should-fix `src/editor/dungeon/dungeon_layout_writer_core.ts:12` (`KEY_ORDER`): the writer emits keys from a hardcoded allowlist that omits `illusionWalls`, a real optional field on `DungeonLayout` (`src/sim/dungeon_layout.ts:71`). `draftToLayout` spreads `...base`, so the field survives editing, but Save to source would silently drop it, corrupting the layout on next load. Not triggered today because `INFERNAL_ABYSS_LAYOUT` has no illusion walls and it is the only authored dungeon, so the round-trip test cannot catch it. It becomes real the moment a second authored dungeon with illusion walls is saved. Either iterate the object keys or add `illusionWalls` to `KEY_ORDER` and a fixture that carries it.

should-fix `src/editor/dungeon/dungeon_validate_core.ts:22` (`doorConnects`): reachability requires `door.hw > door.hd` or `door.hd > door.hw`, both strict, so a square door (`hw === hd`) matches neither branch and connects nothing. The sibling `doorTouchesRoom` (line 54) uses `>=`, so the two disagree on the square case. A user who sets `hw === hd` in the panel gets a spurious `severity:'error'` "unreachable room" that blocks Save on a geometrically valid door.

should-fix `vite.config.ts:408` (`collisionMasterSavePlugin`): unlike the dungeon plugin's `aborted` guard, this endpoint sends the 413 inside the `data` handler but keeps buffering, and the `end` handler still `JSON.parse`s the oversized body and can still write both files (and call `res.end` twice). The byte cap is decorative here. Dev-only, but worth mirroring the dungeon plugin's guard.

should-fix `vite.config.ts:317,392`: neither dev save endpoint checks `Origin`/`Host`, so a `text/plain` POST is a CSRF "simple request": any page open in a dev's browser while `npm run dev` runs can write to `src/sim/dungeon_layout.ts` / the collision overrides. Path is fixed (no traversal) and the literal validator blocks code injection, so the residual is working-tree corruption, not RCE. This is inherited from the existing `musicEditorSavePlugin` (line 171), so a shared same-origin check across all three would close it in one place.

should-fix `server/federated_auth_db.ts:11` (`deleteUnusedFederatedProvision`): this drops the old `NOT EXISTS (SELECT 1 FROM characters ...)` guard, so it now deletes a `password_set = FALSE` account even when it owns characters, which cascade via `ON DELETE CASCADE`. Safe under the current invariant (the only writer of characters onto an unverified account is the same seeding transaction, and the predicate also requires no token and no Apple/Discord link), so it is strictly tighter than the Apple-only predecessor. Flagging so the widened blast radius is a conscious choice: a one-line test pinning "only same-transaction seeded characters may exist on a deletable federated provision" would keep it honest.

nits:
- `server/unstuck_db.ts:80` the `unstuck_reports_realm_area_local` composite index backs no query in this diff (hotspots filter on realm+created_at, list keysets on id); it only adds insert cost. Drop it or note the intended reader.
- `dungeon_validate_core.ts:102` BFS follows the door graph only, so overlapping or wall-abutting rooms that are walkable-adjacent without a door are reported unreachable and block Save. Over-strict rather than wrong.
- `dungeon_layout_writer_core.ts:61` `formatString` escapes only `\` and `'`, not newlines; not reachable through the current inputs, but the writer is described as faithfully round-tripping arbitrary layouts.

**Not a finding, for the record**
The `tests/snapshots.test.ts` "active title wire round-trips" case fails when that file is run in isolation (`WebSocket.OPEN` read off an undefined global in `canSendCommand`). This reproduces identically on `release/v0.27.0` (base actually fails two cases in isolation), so it is a pre-existing shared-global test-ordering fragility, not something this PR introduced; it passes in the full ordered suite.

The four follow-ups you already list (inherited Biome red from Troy's test files, i18n M16 fills for the `editor.dungeon.*` strings, the dropped Rift dev cheats, Rift music) all look correctly scoped. Given the size, splitting the mergeable editor-dungeon-mode slice out from the wholesale `woc-map-editor` union merge would make this far easier to land and to bisect later, but I understand it is opened as a draft integration branch for playtesting.

Issue #2021, opened 2026-07-16
Bear Form armor is low and Oakhide barely moves it


jgyy:
## Problem

Bear Form's armor value feels low relative to what a tank needs, and Oakhide (its
armor cooldown) grants only about 150 armor on a 60 sec cooldown, which is a small
fraction of a Bear Form druid's total armor (observed around 3159 armor with
less-than-optimal gear). As-is, Oakhide barely moves the needle on survivability for
its long cooldown.

## Expected behavior

- Bear Form's baseline armor and/or Oakhide's armor grant are strong enough that
  Oakhide is a meaningful mitigation cooldown rather than a negligible bump.

## Scope

- Bear Form's baseline armor scaling.
- Oakhide's armor amount and/or cooldown.

## Acceptance criteria

- [ ] Oakhide's armor grant is a meaningful percentage of a Bear Form druid's total
      armor, not a low single-digit percentage bump.
- [ ] Bear Form's baseline armor scaling is reviewed against comparable tank
      mitigation at the same content difficulty.

## Notes

Filed from a batch of player feedback on druid balance and mechanics. This is a design
discussion, not a specific implementation.

Issue #2020, opened 2026-07-16
Wolf Form's openers (stun / DoT) have no meaningful follow-up beyond Claw


jgyy:
## Problem

Wolf Form's kit does not give a player much to do. Stealth leads into one of two
openers, and neither leads anywhere useful:

1. A small DoT (about 26 damage every 3 sec) plus +14 weapon damage on the opener.
2. A 2 sec stun opener, which leads into nothing: the only follow-up attack available
   is Claw, which is also just weapon damage (+23).

In both lines, the only thing after the opener is Claw. There is no combo-point/finisher
system or any other follow-up ability that makes use of the stun or the DoT setup, so
Wolf Form has very little to actually do once combat starts.

## Expected behavior

- Wolf Form has a kit with more than one meaningful follow-up action after its openers,
  so the stun and DoT openers lead into a real combat sequence rather than dead air.

## Scope

- Wolf Form's ability kit (openers, follow-ups, any combo/resource mechanic).

## Acceptance criteria

- [ ] Wolf Form has at least one follow-up ability usable after the stun opener besides
      Claw.
- [ ] Wolf Form has at least one follow-up ability usable after the DoT/weapon-damage
      opener besides Claw.
- [ ] The two opener lines lead into distinguishable playstyles rather than converging
      on the same single follow-up.

## Notes

Filed from a batch of player feedback on druid balance and mechanics. This is a design
discussion, not a specific implementation.

Issue #2019, opened 2026-07-16
Bear Form / Wolf Form scale off Agility, which druids can't itemize for


jgyy:
## Problem

Bear Form and Wolf Form scale off Agility, but the only weapon that itemizes Agility
for druids is the one legendary staff (see the companion weapon-itemization issue).
This effectively forces a Feral druid into gearing for a stat their class otherwise
never itemizes for. By comparison, other hybrid classes that fill multiple roles
(for example, paladins) do not need an entirely separate gear set to fill a different
role. Making Bear Form and Wolf Form scale off Intellect gear instead of Agility would
let a druid gear once and shift between roles, the way other hybrids already can.

## Expected behavior

- Bear Form and Wolf Form combat scaling uses Intellect (the stat druids otherwise
  itemize for) instead of Agility.

## Scope

- Bear Form / Wolf Form attack power, armor, and any other Agility-derived scaling in
  the combat/stat pass.
- Tooltips and character sheet display for the affected forms.

## Acceptance criteria

- [ ] Bear Form and Wolf Form attack power (and any other Agility-derived combat stat)
      scales from Intellect instead of Agility.
- [ ] A druid gearing purely for Intellect performs comparably in Feral forms to how
      they previously performed gearing for Agility (no net damage/tank-stat
      regression from the swap alone).

## Notes

Filed from a batch of player feedback on druid balance and mechanics. This is a design
discussion, not a specific implementation, and is closely related to the weapon
itemization issue (both stem from Agility being effectively ungeared for druids).

Issue #2018, opened 2026-07-16
Feral druids have only one Agility weapon (the legendary staff)


jgyy:
## Problem

Druids only have one weapon in the game that carries Agility, which is the stat Feral
spec scales off in Bear Form and Wolf Form: the one legendary staff. Daggers used to be
usable by druids but no longer are. In practice this means a Feral druid has exactly
one viable weapon upgrade path for their entire leveling and endgame experience, which
is a much narrower itemization pool than every other class gets.

## Expected behavior

- Feral druids have more than a single weapon (ideally more than one weapon type) that
  itemizes for their spec, at multiple points along the level/gear curve, not just one
  legendary-tier item.

## Scope

- Weapon itemization across zones/dungeons/vendors for druid-usable weapon types.
- Whether daggers (or another weapon type) should be reintroduced as druid-usable.

## Acceptance criteria

- [ ] Feral druids have at least one additional itemized weapon path besides the
      single legendary staff, reachable before that legendary's typical acquisition
      point.
- [ ] Any newly-usable weapon type is reflected in loot tables/vendors consistently
      with how other classes' weapon types are itemized.

## Notes

Filed from a batch of player feedback on druid balance and mechanics. This is a design
discussion, not a specific implementation.

PR #2022, opened 2026-07-16
fix(druid): stop caster-form attack slot leaking into shapeshifts, cancel orphaned form auras on respec, fix stale Wolfsblood tooltip

jgyy:
## Summary

Five druid-facing bug fixes: three reported by a player alongside a batch of balance
feedback, plus two more confirmed independently in #2016 while investigating the same
reports (the design/balance items from both threads were filed as separate issues:
#2017, #2018, #2019, #2020, #2021, since they need a design decision rather than a
bug fix).

1. **Action bar slot 1 stuck across shapeshifts.** Once the auto-attack button is
   turned off, slot 1 shows a custom ability instead of the attack toggle. That
   custom ability was keyed only to the caster ("normal") form and loaded once at
   init, so it never swapped when shapeshifting into Bear Form or Wolf Form, unlike
   every other slot.
2. **Moonkin Form (and any other shapeshift toggle) survives a spec switch.**
   Switching away from Balance to another spec dropped Moonkin Form from known
   abilities but never cancelled the aura itself, so its armor/spell-damage buff
   kept applying in a different spec. (#2016 reasoned this wasn't reproducible from
   a code read; the new regression test in this PR shows it is.)
3. **Wolfsblood tooltip hardcoded a stale value.** The tooltip always said
   "40 attack power", but the real value scales with melee-damage talents/masteries
   (for example Feral's Primal Heart, +15%), so a player with any of those saw 46 in
   combat against the tooltip's 40.
4. **Oakhide (barkskin) couldn't be cast while shapeshifted.** It's a Bear/Wolf Form
   tank cooldown but was missing the `usableInForm: true` flag that Primal Reflexes
   and Primal Surge already use, so the shapeshift cast gate rejected it with
   "You can't do that while shapeshifted." The only workaround was pre-casting it in
   caster form before shifting.
5. **Wolf Form's aura tooltip still said "Cat Form."** Renamed to match every other
   player-visible surface for this shapeshift (ability name, error messages, guide).

```mermaid
flowchart TD
    A[Auto-attack button OFF] --> B[Slot 1 shows a custom ability]
    B --> C{Before fix}
    C -->|attackActionState keyed to 'normal' form only,\nloaded once at init| D[Same ability shown\nin Bear/Wolf Form too]
    B --> E{After fix}
    E -->|attack slot key scoped to active form,\nreloaded on every syncActiveForm| F[Correct per-form ability\nin Bear/Wolf/normal]
```

```mermaid
flowchart TD
    G[Player casts Moonkin Form as Balance] --> H[form_moonkin aura applied,\narmor + spell dmg buffed]
    H --> I[Player switches spec to Restoration]
    I --> J{Before fix}
    J -->|recomputeTalents refreshes known abilities\nbut never touches active auras| K[form_moonkin aura stays active,\nbuff persists into the new spec]
    I --> L{After fix}
    L -->|stripOrphanedFormAuras cancels any\nform/stance aura whose ability fell\nout of meta.known| M[Aura removed,\nstats recalculated]
```

```mermaid
flowchart TD
    N[Druid tanking in Bruin Form\ncasts Oakhide] --> O{Before fix}
    O -->|barkskin lacks usableInForm;\ncasting_lifecycle's shapeshift gate\nrejects all non-toggle abilities| P["You can't do that while\nshapeshifted." - cast fails]
    N --> Q{After fix}
    Q -->|barkskin: usableInForm true,\nsame mechanism as Primal Reflexes\n/ Primal Surge| R[Oakhide casts normally,\narmor buff applied]
```

**Behavior note on the attack-slot fix (per review):** a player who had the attack
button off and a form-agnostic action (a potion, say) on slot 0 previously saw it in
every form. Post-fix, non-normal forms read a per-form storage key that never
existed before this change, so they show an empty slot rather than one-time-seeding
from the old shared value. This is intentional, matching the existing stealth-bar
convention (blank-by-default, no auto-clone from the shared bar), not an oversight.

## Related issues

Closes the confirmed-defect portion of #2016 (Wolfsblood tooltip, the Oakhide
shapeshift gate, and the Wolf Form aura name); its four remaining items are design
questions needing a ruling, not fixed here.

Design/balance follow-ups filed separately (not fixed here, since they need a
design decision and CLAUDE.md forbids inventing new balance numbers):
- #2017 Druid caster form uses melee autoattack instead of ranged
- #2018 Feral druids have only one Agility weapon (the legendary staff)
- #2019 Bear Form / Wolf Form scale off Agility, which druids can't itemize for
- #2020 Wolf Form's openers have no meaningful follow-up beyond Claw
- #2021 Bear Form armor is low and Oakhide barely moves it

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands: `npx tsc --noEmit -p .`, `npx vitest run` (full suite; two pre-existing
  failures confirmed unrelated by reproducing them identically on the unmodified
  base commit: `tests/deeds_window_focus.test.ts`'s `localStorage` jsdom setup issue,
  and one flaky `tests/server/http/characterization.test.ts` timeout under core
  contention that passes standalone), `npm run i18n:gen`,
  `npx @biomejs/biome check --write` on every changed file, pre-push QA floor (green).
- Manual steps: reproduced both bugs live in the offline client (screenshots below),
  toggling the attack button off and shapeshifting for the action-bar bug, and
  picking Feral spec to trigger the Wolfsblood tooltip scaling mismatch.
- Added/updated tests:
  - `tests/action_bar_controller.test.ts`: new case reproducing the druid
    shapeshift attack-slot leak and asserting the fix.
  - `tests/talents.test.ts`: new case casting Moonkin Form as Balance, switching to
    Restoration, and asserting the aura and its armor buff are gone.
  - `tests/tank_defensive_cds.test.ts`: new case casting Oakhide in Bruin Form and
    Wolf Form and asserting the armor buff actually applies (fails without the
    `usableInForm` fix, verified by reverting it locally).
  - `tests/talents.test.ts`: added a priest mirror (Shadow's `shadowform`) of the
    Moonkin Form regression test, per review.
  - `tests/action_bar_controller.test.ts`: strengthened per review to also pin the
    underlying per-form storage key literal, not just the resolved slot value.
  - `tests/localization_coverage.test.ts`'s existing sentinel guard now also pins
    the Wolfsblood tooltip fix automatically (verified decisive by reverting the
    catalog placeholder locally and confirming the guard fails).

### Action bar slot 1 leaking across shapeshifts

Attack button hidden, Regrowth (a normal-form-only heal) placed in slot 1.

**Before** — Regrowth's icon stays stuck in slot 1 (greyed out, since it's
uncastable there) in both Bear Form and Wolf Form:

| Normal Form | Bear Form | Wolf Form |
|---|---|---|
| before normal | before bear | before wolf |

**After** — slot 1 correctly empties out per form, matching every other slot:

| Normal Form | Bear Form | Wolf Form |
|---|---|---|
| after normal | after bear | after wolf |

### Wolfsblood tooltip

Feral spec (Primal Heart, +15% melee damage) scales the base 40 attack power to 46.

**Before** — tooltip hardcodes "40" directly above the resolved "46":

wolfsblood before

**After** — tooltip reads the resolved "46", matching the line below it:

wolfsblood after

---

### Quality

- [x] **The full gate passes.** `npm run gate`-equivalent checks (tsc, full test
      suite, i18n gen, biome on changed files) are green. Added decisive tests for
      changed behavior (see above).

### Cross-platform

- [x] N/A for the two sim/logic fixes (action-bar controller, talent respec): no
      layout or viewport-dependent surface changed. The Wolfsblood tooltip fix
      reuses the existing tooltip placeholder mechanism (`{buff}`/`{duration}`)
      already used by every other scaled buff tooltip, so its desktop/mobile
      layout is unchanged, only the numbers are.
- [x] N/A. Accessible: no new interactive elements.

### Localization (i18n)

- [x] No new player-visible strings. The Wolfsblood tooltip fix parameterizes an
      *existing* key's already-established `{buff}`/`{duration}` placeholders
      instead of hardcoding numbers; carried the same token change through every
      locale that already had this key (all 21) so the whole-catalog
      interpolation-token parity guard (`tests/i18n_completeness.test.ts`) stays
      green, with only the token positions changing, not the translated wording.
      The Wolf Form aura-name fix corrects the existing English source row only
      (`hud_chrome.ts`'s `formCat`); it carries no interpolation tokens, so the
      parity guard needs no locale changes for it.
- [x] N/A. No new numbers/money/dates introduced.
- [x] N/A. No new sim/server player-facing text.

### Hygiene

- [x] No secrets, credentials, or `.env` committed. `ALLOW_DEV_COMMANDS` is not
      enabled anywhere.
- [x] No hand-edited generated files outside the normal `npm run i18n:gen`
      regeneration (`src/ui/i18n.resolved.generated/`).


FernandoX7 (commented):
Nice batch. All three fixes check out against the real code paths, and splitting the balance/design items into #2017 to #2021 instead of inventing numbers here was the right call. Verified before writing this: the strip set matches the isFormKind toggle set exactly (form_metamorph is a timed transform, so leaving it out is correct); every form-kind aura in content is a single primary selfBuff carrying the bare ability id, so the knownIds membership test is sound; all form-exit side effects (resource flip, energy/rage reset, moonkin armor) live inside recalcPlayerStats, which the helper calls; parity goldens are unchanged at the PR head; and both new tests fail on the unfixed base (checked with mutation runs, both fix arms independently). The syncActiveForm ordering (save under the outgoing form's key, then flip, then load) is right, and the assignment paths in hud.ts all agree on the key.

Should-fix:

1. `src/sim/progression/talents.ts:60`: FORM_AURA_KINDS is the sixth inline copy of this five-kind list (`effect_dispatch.ts:1006` and `:1037`, `casting_lifecycle.ts:74` and `:91`, `chat_readouts.ts:290`). The repo extracts on the rule of three, and silent drift between these lists is exactly the bug class this PR fixes. At minimum, export the set (or an isFormAuraKind predicate) from `types.ts` next to AuraKind and consume it here; consolidating the five pre-existing sites can be a follow-up.

2. `tests/talents.test.ts`: the form_shadow arm is untested. Only two abilities can actually orphan a form aura (bear/cat/travel are baseline-known): moonkin_form, which is tested, and the priest shadow signature shadowform, which is not. Deleting 'form_shadow' from the set leaves the suite green. A priest mirror of the new druid test (setSpec('shadow'), cast shadowform, setSpec away, assert the aura is gone) pins it.

3. `src/sim/content/classes.ts:3551`: the sim-side tigers_fury description still hardcodes "40 for 6 sec". The convention for scaled tooltips is $-tokens in the sim source (battle_shout at `classes.ts:434`) paired with the placeholder form in the catalog, and the token-interpolation guard in `tests/localization_coverage.test.ts` only fires when the sim source contains $b/$t. As written, nothing pins this fix: a regression back to a hardcoded number, or a literal "{buff}" leaking into the tooltip, would not fail any test. Tokenize the source to "by $b for $t sec." and the existing sentinel plus the commanding_shout-style hard pin machinery cover it automatically.

4. `tests/action_bar_controller.test.ts`: the form-scoped storage key is never pinned to a literal. Save and load both go through attackSlotStorageKey(this.slotMapKey()), so a key-format change would round-trip green here while orphaning every druid's saved per-form attack slot. The existing pin at `:303` covers only the normal-form key; after the claw save, also assert storage.getItem('woc_hotbar_druid_ActionbarTester_bear:s0') is non-null.

Nits:

- `tests/action_bar_controller.test.ts:317`: `.not.toEqual({ type: 'ability', id: 'mangle' })` is decisive against this bug, but the expected value is knowable; `toBeNull()` is strictly stronger and also pins that a fresh bear form starts with an empty attack slot.
- `src/ui/hud/action_bar/hotbar.ts:46`: attackSlotStorageKey's parameter is still named normalSlotMapKey; callers now pass any form's key.
- One behavior note worth a line in the summary: a player who had the attack button off and a form-agnostic action (a potion, say) on slot 0 previously saw it in every form; post-fix the non-normal forms read keys that never existed and show an empty slot, with no one-time seed from the old shared value. The after screenshots suggest per-form-blank is the intended design (it matches the stealth bar's blank-by-default choice); if so, fine as is, just say so.

jgyy:
Thanks for the thorough review, all four should-fix items and the nits are addressed:

1. Extracted `FORM_AURA_KINDS` / `isFormAuraKind()` into `src/sim/types.ts` and consumed it from all six sites, including the five pre-existing ones you flagged (`effect_dispatch.ts:1006`/`:1037`, `casting_lifecycle.ts:74`/`:91`, `chat_readouts.ts:290`). Left the two narrower bear/cat/travel-only checks in `casting_lifecycle.ts` (the ones you didn't cite) alone: those gate druid's own caster-kit lock and deliberately exclude moonkin/shadow, a different semantic set from the toggle-cancel list.
2. Added the priest mirror test (`setSpec('shadow')`, cast `shadowform`, `setSpec('holy')`, assert the aura is gone) in `tests/talents.test.ts`. Confirmed bear/cat/travel can't be used to test this (baseline-known, never orphaned), so moonkin_form and shadowform really are the only two abilities that exercise the path.
3. Tokenized the sim-source `tigers_fury` description to `"by $b for $t sec."`, matching the `battle_shout` convention. Verified it's decisive: reverted the catalog `{buff}`/`{duration}` placeholder back to a hardcoded "40"/"6 sec" locally, and `tests/localization_coverage.test.ts`'s sentinel guard failed exactly as expected (`expected '...by 40...' to contain '35'`), then restored the fix.
4. `tests/action_bar_controller.test.ts`: now asserts `storage.getItem('woc_hotbar_druid_ActionbarTester_bear:s0')` is non-null after the bear-form save, and tightened the post-shapeshift assertion to `toBeNull()`.

Nits: swapped `toBeNull()` in for `not.toEqual(mangle)`, renamed `attackSlotStorageKey`'s parameter to `formSlotMapKey`.

On your behavior note: yes, per-form-blank is the intended design here, not an oversight. It matches the existing stealth-bar convention (blank-by-default, no auto-clone from the shared bar) rather than one-time-seeding a new per-form key from the old shared value. Added a line to the PR summary calling this out explicitly for anyone else evaluating the diff.

Full suite green after these changes (same two known-flaky/pre-existing failures as before, `deeds_window_focus.test.ts` and the `characterization.test.ts` core-contention flake, both reproduced identically on the unmodified base commit).

FernandoX7 (approved):

PR #469, opened 2026-06-16
docs(woc): Market fee in $WOC + burn — feature stub for discussion

awidearray:
🚧 **Feature stub — opening discussion, not for merge.** This PR adds a stub spec for the **Market fee in $WOC + burn** `$WOC` flywheel mechanic so we can discuss it in its own thread before any code is written. **No implementation is included.**

**Scores (roadmap):** Tier 1 · Ease 4/5 · Flywheel 3 · Burn · Reg risk Low

**What:** Let players pay (or get a discount on) the existing 5% auction-house merchant cut in $WOC, and burn a share of it.

**Constraints:** cosmetic-only / no pay-to-win · non-custodial · `src/sim/` stays pure.

**To discuss:**
- Discount size for paying in $WOC?
- What `burn %` of the fee?
- Opt-in, or default when a wallet is linked?

Part of the proposed `$WOC` GameFi roadmap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2023, opened 2026-07-16
feat(market): auction house - durations, deposits, bidding, per-unit stacks, multi-currency listings

awidearray:
## Summary

A fully fledged auction house, built as an extension of the World Market: player
set expirations, bidding with buyouts, per-unit stack listings with partial
purchases, offline notifications, time-left and sorting, and multi-currency
denominations (gold always on; Claudium and WOC fail-closed behind the same
rails flags as PR #2001).

STACKING NOTE: this branch merges feat/woc-p2p-trading (PR #2001) because the
currency layer reuses its verification and rails infrastructure
(server/woc_trade.ts, claudium proxy legs, trade_rails_boot, wallet caches, the
trade_ledger). Reviewers can read this PR's own delta as commits 47ac85804
(copper auction house), f080fff80 (multi-currency), a53513843 (review
hardening), b6f446824 (screenshots). If #2001 lands first this PR rebases to a
tiny delta; if #2001 stalls, commit 47ac85804 is cherry-pickable standalone
(the copper auction house has zero dependency on the trade work).

### Always-on (pure sim currency, no flags)

- Listing durations of 12, 24, or 48 hours with classic-shaped copper deposits
  (15 percent of vendor value per unit, x1/x2/x4 by tier; refunded per-unit as
  the listing sells, forfeited on expiry or cancel) as the anti-spam gold sink.
- Per-unit pricing on fixed-price stacks with partial purchases: quantity
  spinner, exact integer math per tranche, listing decrements and closes at
  zero. Legacy persisted rows stay whole-lot and age out within 48 hours; no
  migration, no value drift.
- Auctions: starting bid plus optional buyout, 5 percent minimum increments,
  full-bid escrow from the bidder's purse, outbid refunds through the existing
  offline-safe collection box, expiry settlement to the winner's box, the
  Merchant's 5 percent cut unchanged. Cancel with a live bid refunds the bidder.
  A now-soulbound auction reclaimed at load refunds its bidder's escrow.
- Structured marketOutbid/marketWon/marketSold/marketExpired events online and
  authored Merchant letters offline, closing the market's zero-offline-
  notification gap. Conservation is asserted per flow in tests: copper in
  equals copper out plus cut plus forfeited deposit, exactly.
- Time-left on every listing (previously not even on the wire), sort by newest,
  price (grouped by denomination), or time left, and my-listings bid state.

### Multi-currency denominations (fail-closed OFF)

Listings carry a denomination: gold keeps everything including auctions;
Claudium and WOC are fixed-price only, unavailable (never rendered, sim-refused
with clear errors) until the operator sets the same flags PR #2001 introduced:

- Claudium (CLAUDIUM_TRADE_ENABLED): per-unit with partial buys, settled buyer
  to seller through the economy service's idempotent transfer with REALM-SCOPED
  dedupe keys (market-realm-listing-seq). An AMBIGUOUS service failure never
  unlocks the goods: the pending purchase stays anchored and retries the same
  key on a ticker, across boots, until the service answers definitively. The
  same product gate as #2001 applies: the service's audited one-way invariant
  must be relaxed by a maintainer decision before this flag is ever set.
- WOC (WOC_TRADE_ENABLED): whole-lot only; the buyer pays the SELLER'S linked
  wallet directly via a Solana Pay request verified on-chain at finality before
  goods release. No refund path exists by construction: money moves only
  wallet-to-wallet on a verified payment, goods only after verification, with a
  final verify before any timeout unlock. The sim never does WOC arithmetic
  (opaque normalized decimal strings).
- A pending purchase locks the listing (no buy, bid, cancel, expiry, or
  soulbound reclaim), persists inside the market blob with wallet pubkeys and
  account ids, and recovers deterministically at boot. Abuse controls: one
  pending external purchase per buyer, a cooldown after an unpaid WOC timeout,
  and reservation caps on both rails so a burst cannot exhaust the shared
  connection pool.
- Bids stay gold-only permanently: non-custodial outbid refunds cannot exist
  without an on-chain escrow program, and this PR builds no custody. Deposits
  stay gold as the universal listing fee. No cut is taken on external
  denominations (a wallet-to-wallet payment cannot be skimmed; maintainer
  decision documented in docs/prd/trading-multi-currency.md).
- Completed external sales are recorded in trade_ledger with a market context
  marker (additive nullable column).

## Related issues

None open request this; roadmap stubs #469/#471 discuss auction features and
are superseded by working code. No in-flight PR touches these mechanics (the
premium-listings cluster is a separate USDC cosmetics subsystem).

## Type of change

- [x] Tests

## How was this tested?

- Commands: npm run gate fully green, all 11 steps, twice (after the feature
  commits and after the review-hardening commit): i18n gen + freshness, malware
  scan, biome changed-files, sfx, full vitest (14,553 passed, 24 skipped, zero failures), browser suite (64),
  tsc, env/server/client builds. Trade suites from the stacked base stay green.
  Parity: market_round_trip regenerated deliberately (deposits change list-time
  copper); new market_auction and market_partial scenarios minted; rng digests
  unchanged; no other golden churn.
- Manual steps: the committed scripts/market_auction_shot.mjs boots the offline
  game headless, stages a real auction (list, bid via the actual sim methods
  beside a real Merchant), and asserts the window renders the auction row,
  quantity spinner, and time-left before capturing.
- Review: a seven-role reviewer matrix (security, architecture, cross-platform,
  frontend, database performance, test coverage, adversarial realness audit)
  produced 7 blocking findings, all fixed in a53513843 and re-verified by an
  independent adversarial pass tracing the settlement code paths.

## Screenshots / recordings

Browse: per-unit stack with quantity spinner and live total, time-left column,
and a live auction row (your-bid badge, min bid, buyout button, leading state):

browse

Sell: fixed/auction toggle, 12/24/48h duration picker, list-all quantity, live
total, and the deposit preview. External denominations correctly absent with
the rails flags off (fail-closed gating visible):

sell

Phone viewport:

mobile

### Quality
- [x] The full gate passes (all 11 steps, zero failures, evidence above).

### Cross-platform
- [x] Desktop and phone-viewport verified; 40x40 touch targets and 16px inputs
      on every new control including the payment panel.
- [x] Keyboard-operable listbox roles preserved; the market window's existing
      a11y contract extended to the new controls.

### Localization (i18n)
- [x] Every new player-visible string is a t() catalog key; every new sim emit
      literal has a client matcher (S3 green); wordy values carry the five
      non-Latin fills; letters localize by letterId; generated artifacts
      regenerated, never hand-edited.

### Hygiene
- [x] No secrets or .env committed; no generated file hand-edited; additive
      idempotent DDL only (one nullable trade_ledger column).

## Flags and rollout

Merging is behavior-neutral for external money: with both flags unset the
auction house is a complete gold-economy feature. The Claudium and WOC
denominations activate under the operator sequence already documented for
PR #2001, and the Claudium leg additionally requires the economy-service
product decision described there. Fee policy on external denominations
(currently: none) is explicitly a maintainer decision.


awidearray:
Status: head `76e77100a1c08d2106ef947ad23bea6ba5f32fd6`, base release/v0.27.0, opened fresh against the current tip.

Evidence:
- npm run gate: PASS, all 11 steps, zero failures. Full suite 14,553 passed / 24 skipped; browser suite 64; tsc 0 errors; env/server/client builds green.
- Parity: market_round_trip regenerated deliberately (deposits change list-time copper); new market_auction and market_partial scenarios minted with their goldens; rng digests unchanged; no other golden churn. The stacked trade branch's suites all stay green.
- Conservation is asserted per flow in the sim tests: copper in equals copper out plus the Merchant cut plus forfeited deposits, exactly; escrow (goods, bids, deposits) round-trips save/load including live bids, and a now-soulbound auction reclaimed at load refunds its bidder.
- Review: a seven-role reviewer matrix plus an adversarial realness audit produced 7 blocking findings (realm-scoped dedupe keys, ambiguous-failure conservatism, pending-lock abuse controls, a Claudium concurrency cap, and three test-integrity gaps), all fixed in a53513843. An independent adversarial re-verification traced the fixes and confirmed each HOLDS; its three residual findings (rail-unconfigured boot handling, cooldown across restart, multi-pending corruption logging) are fixed in 76e77100a with tests.

Draft rationale: the always-on layer is a pure gold-economy feature and is fully live on merge. The PR is draft because it stacks on #2001 (whose diff it carries for the WOC verification and Claudium rails infrastructure) and because the external denominations, though fail-closed OFF, deserve the same maintainer decisions documented there: whether WOC and Claudium player commerce is wanted, the economy service's one-way invariant, and fee policy on external denominations. If #2001 lands first this rebases to a small delta; commit 47ac85804 (the copper auction house) is independently cherry-pickable if the maintainers want the gold layer alone.

awidearray:
Update: caught up with the moved base (merge up to 6b4a5681c), now MERGEABLE at head `33651b3c7b704bbfd779f975b76145103b1a425d`.

Two notes from the catch-up, for transparency:
- tests/parity/golden/market_partial.json was regenerated (commit 33651b3c7): the upstream merge removed every player's cosmetic quality:epic visited mark and the col_first_epic deed from that one scenario's trace. A structural diff of old vs new golden confirms the delta is exclusively those marks; copper, inventory, escrow, and event outcomes are unchanged and the sibling market and trade goldens did not move. None of this branch's code touches the deed trigger.
- The pre-push biome floor flagged six files (daily_rewards_payout_moderation_db, mob_family_presentation, moderation_db, server/claudium, server/daily_rewards_routes, tank_defensive_cds tests) that are byte-identical to origin/release/v0.27.0: the hook diffs against origin/main, so the catch-up merge makes the base branch's own formatting debt look branch-changed. This PR's own files are biome-error-clean (proven at the full gate). Pushed with the hook's documented bypass rather than formatting unrelated upstream files into this diff.

EnriqueGF (commented):
Reviewed the auction-house and multi-currency work end to end: read the full sim delta (market.ts, sim.ts, sim_context.ts, types.ts, social/trade.ts, letters.ts), the wire/parity surface (world_api facets + pin, online.ts, server/game.ts dispatch), the server settlement/rails code, and the UI. Ran locally: tsc clean, and the guard suites green (architecture, localization_fixes S3, world_api_parity, parity goldens, plus market/trade/settlement/denom/db suites, 400+ tests). One unrelated pre-existing failure only: `snapshots.test.ts > key lockstep`, a Book of Deeds title-wire test untouched by this PR that reads a global `WebSocket` my isolated vitest run does not provide; it is not a regression.

This is a lot of surface done with real discipline. Copper conservation is proven per flow (buyer out == seller in + 5% sink + forfeited deposit, deposits refunded strictly per unit sold so no cross-tranche rounding drift), the sim stays deterministic (zero rng draws added, draw digests byte-identical, all three sort orders tiebreak on the unique listing id for a strict total order, WOC compared as opaque strings with no float), external denominations are genuinely fail-closed server-side (offline reads `OFFLINE_TRADE_RAILS`, both orchestrators re-check the flags, ambiguous service failures never release goods), the IWorld parity pin is updated in the same change, and player strings are English-only catalog keys with sim/server matchers. No secrets, no dashes or emojis, additive idempotent DDL.

Findings below. Nothing blocking; the one that is genuinely this PR's to fix is the browse-tab input wipe.

## should-fix

- **Browse list rebuilds every second and wipes a mid-typed bid or quantity** (`src/ui/market_window.ts:206` + `:497`). `refreshIfChanged` puts `info.listings` in its change signature, and every listing carries `secondsLeft` at whole-second granularity (`src/sim/market.ts` `Math.round(expiresAt - time)`), so the signature flips once a second and `renderBrowse` does `list.innerHTML = ''` and rebuilds every row. This PR is what newly places focusable inputs in those rows: the auction bid field (`:1061`) and the per-unit quantity spinner (`:1081`/`:1139`), both with hardcoded `value` defaults and no preservation. A player part-way through typing a bid or a buy quantity has the field destroyed and reset to `min` / `l.count` once per second. The rebuild cadence pre-existed (rows were button-only), the interactive regression is new. Excluding `secondsLeft` from the browse signature (or bucketing it to the displayed minute granularity, or updating the time-left span in place) fixes it. Worth a quick manual check: type a bid on a live auction with the tab open and watch it survive a tick.

## notes (mostly inherited from the stacked #2001 base, tracking not fixing)

- **marketBid copper escrow is not atomic with the bid it backs** (`src/sim/market.ts` marketBid / `server/game.ts` periodic saves). The bid decrements `meta.copper` (character state) while the bid record lives in the market blob, and the two persist on independent ~30s writes, so a crash between them can strand or duplicate the bidder's copper. This extends the listing-deposit non-atomicity already shipped in the base rather than introducing the pattern, but auctions widen the exposure.
- **Market pending-purchase recovery has no atomic claim** (`server/market_settlement.ts` recoverPendingPurchases). Unlike the trade lane's `claimTradeSettlementForRecovery`, market recovery walks the loaded blob after `ensureSchema()` returns, outside the advisory lock. Money stays exactly-once (Claudium dedupe key, read-only WOC verify), but a concurrent same-realm double-boot could write duplicate `trade_ledger` rows. It relies entirely on single-process-per-realm; worth an explicit claim guard or a note.
- **WOC payment can be stranded by reference flooding** (`server/woc_trade.ts` getSignaturesForAddress `limit: 10`). The Solana Pay reference pubkey is public once the buyer submits; >10 junk txs carrying it as a read-only key can push the real signature past the newest-first window, so verification never sees it and the goods unlock at timeout while the payment already landed (buyer pays, loses goods). Same helper and limit as the existing trade lane, so not a regression here, but the market lane inherits it. Paginating past the window or raising the limit would close it.
- **Rollback past this PR corrupts market state on the next old-code save.** The pre-PR `serializeMarket` writes only the legacy fields, so the first autosave on an older server drops every `pending`/`bid`/`denom`/`priceWoc`/deposit: a paid-but-pending external lot becomes a min-price copper lot, an escrowed bid is orphaned, and `nextPurchaseSeq` resets so a later re-upgrade can reissue a stale Claudium dedupe key. Nothing to fix in this PR (the old loader cannot know the fields), but the operator runbook should state that a rollback after any auction or external-denom activity loses that state.
- **The listing deposit is a balance change to the existing free-listing flow.** The `market_round_trip` golden shifted (copper 4700 to 4400) because listing now charges the 15%-of-vendor deposit up front. It is documented in the PR body and the golden was regenerated deliberately; flagging only so it is a conscious change to the pre-existing market, not a silent one.

## nit

- **Two parity-pin test titles drifted** (`tests/world_api_parity.test.ts:450` says "the pinned 234", `:757` says "the pinned 172"). The member set grew by 2 (marketBid, tradeDecline); the `expect()` counts in the same file are correctly 235 / 173 and green, only the human-readable `it(...)` titles are off by one.

FernandoX7 (commented):
Reviewed the auction-house delta (47ac85804, f080fff80, a53513843, 76e77100a, 33651b3c7) against release/v0.27.0. The stacked #2001 trade work is that PR's scope; I checked it only where the market code consumes it.

This is careful work and the money core holds up under an adversarial pass. Every market mutation resolves the actor from the session pid; bid escrow is refunded exactly once on every exit path (outbid, cancel, buyout, soulbound reclaim); wallet pubkeys and account ids never reach the wire; and the conservation tests assert exact equalities on real purses, with a failing-save harness that genuinely proves anchor-before-transfer ordering. Determinism is equally solid: zero rng draws (all three market parity goldens share one drawDigest), expiry rides secondsLeft, tick phase order untouched. I confirmed the market_partial regen claim independently: the diff is exclusively the cosmetic epic-visited and deed marks, and none of these commits touch markItemDiscovered. Module discipline held too: sim.ts gained only one-line delegates, trade rendering actually moved out of hud.ts, and payment_panel is genuinely shared by both windows.

Findings:

- **should-fix** server/main.ts:2626: boot recovery of anchored market pendings is awaited before server.listen (main.ts:2776), and recoverPendingPurchases iterates strictly sequentially with per-call timeouts of 5s (Claudium) and 8s per RPC (WOC), with caps allowing up to 64 anchored pendings. Pendings pile up exactly when the external service is down, so a crash-restart during that outage stalls the realm listen for minutes. Unreachable while both rails are unset, so it does not block this merge, but it should land before either flag is ever enabled: recovery has no correctness reason to gate listen (the loaded blob already keeps the lots locked), so the void-plus-catch pattern used for recordOnlineSnapshot a few lines down would do. The same awaited pattern exists in #2001's recoverOpenTradeSettlements and deserves the same fix there.
- **should-fix** DEPLOY.md: rolling back below this release with live market state is destructive and undocumented. Base loadMarket/serializeMarket round-trip a fixed field list, so a rollback (a) reloads an active auction as a fixed-price lot and silently drops the standing bidder's escrowed copper, which is the always-on gold feature, and (b) strips a locked external pending, unlocking a maybe-paid lot and destroying the recovery anchor on the next save; a WOC lot serializes price 0 (src/sim/market.ts:477) and the old load path clamps that to 1 copper. A rollback caveat matching the existing Timed Daily Rewards precedent (let auctions settle, drain external pendings first) closes this.
- **should-fix** server/market_settlement.ts:590: writeLedger's retry, and the boot-replay window between an in-memory completion and the blob save persisting it, can insert a duplicate audit row; trade_ledger.context has no uniqueness (the ALTER at trade_db.ts:82 is index-free). Money stays exactly-once via the idempotent transfer and on-chain verify, but ledger reconciliation sums go wrong. A partial UNIQUE index on context (IF NOT EXISTS) plus ON CONFLICT DO NOTHING closes it.

Non-blocking notes:

- server/market_settlement.ts:531: a stuck ambiguous Claudium pending logs console.error on every 5s poll indefinitely; at the 32-pending cap during a multi-hour outage that is about six lines a second. Rate-limit the log and consider capped backoff on the retry cadence.
- src/sim/market.ts:22: the WOC amount helpers are value-imported from social/trade.ts, system module to system module. Acyclic today; a small pure leaf both modules import would keep it that way.
- src/ui/market_window.ts:199: the elision signature includes the live-derived secondsLeft, so offline (where marketInfo recomputes on read) the browse page rebuilds its innerHTML on every slowHud pass while open. Cold window, so no contract breach, but bucketing time-left in the signature or updating just the time cells would restore the elision.
- tests/market_window.test.ts: the painter is covered by source-substring scans only. That is the house painter/core convention and the pure core is behavior-tested, but the denomination-selector gating and the phantom-selection reset are protected only as substrings. Also untested: a pending plus now-soulbound listing staying locked through reclaimSoulboundListings at load.

Two process notes. This PR merges the #2001 branch, so #2001 needs its own review before either lands, and this PR rebases if it changes. And CI has not run on this head (fork workflow approval pending); I ran the market suites, the guard tests (architecture, S3, world_api parity, command schema), and tsc at the head in a clean worktree and all are green, but the PR gate should run before merge.

Bottom line: as a gold-economy feature this is in very good shape; the first should-fix belongs before the rails flags are ever enabled and the other two are cheap to land now.

awidearray:
Update: fixed the three manual market driver scripts for the listing-deposit mechanic and caught up with the base again (now MERGEABLE at head `78f395ec1f9ce8fa66bf57bc6db1d9d4b05c88a1`).

- scripts/market_mp_e2e.mjs, market_search_shot.mjs, market_listing_count_shot.mjs predated the auction-house deposit and would misbehave when run: a fresh 0-copper seller now hits the deposit refusal at list time, and the hand-pushed flood-listing literals were missing the auction-era fields. Fixed each (seller earns coin before listing + deposit-aware owed assertion reading depositTotal off the wire; funded flood sellers; completed the listing literal with kind/denom/pricePerUnit/durationSeconds/depositPerUnit).
- Added tests/market_driver_scripts.test.ts (4 cases) pinning the sim-side assumptions those scripts rely on, so this is covered by the gate rather than only by the heavy browser harness.
- The catch-up merge pulled in the upstream biome 2.5.4 bump (commit c31705ae6 area). Two notes: (a) the driver scripts I touched are updated to the new lint rules (template literals, prefixed unused debug locals; behavior-identical); (b) the pre-push hook's local biome over-reports because it diffs against origin/main and now sees the base's own biome-2.5.4 debt in untouched base scripts (critter_tour_shot, player_card_capture_check, welcome_screen_shots). CI's own biome check is scoped with --since against the PR base and passes with 0 errors on this diff; verified locally: `biome ci --changed --since=origin/release/v0.27.0` exits 0.

awidearray:
Thanks @EnriqueGF and @FernandoX7 for the careful end-to-end passes. Addressed the merge-live regression and the cheap-now items at head `7621ee9f6967efba2cf50b23546299c64fa38d84`; the rail-gated items are tracked for before-enable.

Fixed now:
- **Browse input wipe** (EnriqueGF, the one that is this PR's to fix): the per-second refresh signature dropped each listing's live secondsLeft, so a bare 1-second tick no longer rebuilds the newly-focusable bid/quantity rows. The time-left cell now updates in place each pass so the countdown still ticks, and every real change (price/count/bid/denom/pending/add/remove) still rebuilds. Added market_window.test.ts cases: the signature is stable across two snapshots differing only in secondsLeft, and flips on a price/count/bid change and on add/remove.
- **Duplicate trade_ledger rows** (FernandoX7): partial UNIQUE index on trade_ledger(context) WHERE context IS NOT NULL + ON CONFLICT DO NOTHING, so writeLedger's retry and the boot-replay window cannot double-count an external sale; classic-lane NULL-context rows are unconstrained. Tests assert the DDL and the emitted ON CONFLICT clause.
- **Recovery no longer blocks listen** (FernandoX7): recoverPendingMarketPurchases is now fire-and-forget like recordOnlineSnapshot, so a crash-restart during a service outage cannot stall the realm listen. Same fix applied to #2001's recoverOpenTradeSettlements.
- **Parity-pin title drift** (EnriqueGF nit): the 234/172 it() titles and the 234 doc comment now read the real 235/173 that the green assertions already pin.

Tracked for before-either-flag-is-enabled (unreachable at merge):
- **WOC reference-flooding burial** (both reviewers, also on #2001): the top pre-enable money item; I will land verify-a-specific-signature (paginate-to-exhaustion, persist the first satisfying signature) plus a >10-signatures test as a focused follow-up rather than fold a verification rewrite into this diff.
- **marketBid copper escrow non-atomic with the bid** and **market recovery atomic claim**: fold into the settling-lane pair-save / add an atomic UPDATE ... RETURNING claim, matching the trade lane's claimTradeSettlementForRecovery.
- **Stuck-Claudium log rate-limit + backoff** and a **capacity cap on all settling trades**: taken.

Documentation:
- **Rollback caveat** (FernandoX7 should-fix): I will add a DEPLOY.md rollback note in the before-enable batch (matching the Timed Daily Rewards precedent: let auctions settle and drain external pendings before rolling below this release), since the always-on auction state (bids) is affected too.

Notes acknowledged: the WOC amount-helper import direction (small pure leaf), the offline browse elision, the deposit balance change to the free-listing flow (documented, golden regenerated deliberately), and that the fork CI gate should run before merge.

EnriqueGF (changes requested):
Thanks for this, it is a genuinely impressive piece of work. The fail-closed rail design, the "ambiguous failure never unlocks the goods" doctrine, the realm-scoped dedupe keys, and the offline-letter/online-event dual path are all handled with real care, and the seam discipline is excellent (facet split, parity pin, and the structured-event-plus-`t()` design that sidesteps the S3 matcher burden entirely rather than adding matcher rows that rot).

## What I reviewed and ran

Read the full delta against the merge base `e5e704d10` (the sim, server, ui/net, and test slices separately), plus the root, `src/sim/`, and `server/` CLAUDE.md.

Ran myself in the worktree:

- `npx tsc --noEmit`: clean.
- Guard tests: `architecture.test.ts`, `localization_fixes.test.ts` (S3), `world_api_parity.test.ts`, `sim_context.test.ts`: green.
- Feature + parity suites: `market.test.ts`, `market_auction.test.ts`, `market_settlement.test.ts`, `market_denoms.test.ts`, `market_bid_wire.test.ts`, `tests/parity`: green. The one failure I hit (`coverage.test.ts > nythraxis_full_pull`, 5s timeout) passes in isolation and is core-contention flake, not yours.
- `npm run ci:changed`: no errors, no format diffs.
- Scanned added lines for em/en dashes and emojis: clean (the hits in `server/game.ts` and `online.ts` are pre-existing lines).
- Confirmed every `rng.digest` in every golden is unchanged, so no draw-order movement.

I also dispatched the parity, sim-determinism, and security reviewers, and independently verified each consequential finding against the code before listing it below.

## Blocking, before the `WOC_TRADE_ENABLED` flag is ever set

**1. One on-chain payment can settle two purchases.** `txSatisfiesPledge` (`server/woc_trade.ts:171-180`) accepts any transaction where `recipientDelta >= amountBase && payerDelta <= -amountBase`. Nothing binds a signature to one specific settlement, and there is no consumed-signature ledger anywhere (`server/trade_db.ts` records `woc_a_signature`/`woc_b_signature` per settlement but has no unique constraint or cross-settlement check).

The concurrency guard that would otherwise prevent two live pendings is scoped wrong: `buyerKey` is `String(meta.characterId ?? meta.entityId)` (`src/sim/market.ts:276-278`), so it is per-character, while the wallet is per-account (`walletPubkeyFor(session.accountId)`, `server/market_settlement.ts:398`).

So: one account, two characters, one linked wallet. Character A starts a WOC purchase on seller S's listing L1 (reference R1). Relog as character B, start a purchase on the same seller's L2 at the same price (reference R2). Both pendings are now live with the same payer and recipient wallet. Submit a single Solana transaction paying S once, carrying both R1 and R2 as read-only reference accounts. `getSignaturesForAddress` finds it under both references and `txSatisfiesPledge` passes for both, since each checks its own amount independently with `>=`. Two lots delivered, one payment. The seller is robbed of the second lot.

The same trick spans orchestrators without a second character, since `MarketSettlements` and the player-trade `TradeSettlements` leg share the verifier and the wallet pair and neither knows about the other's obligations.

Two changes are needed: record consumed signatures under a unique constraint across both orchestrators, and scope the concurrency guard (and the `MARKET_WOC_RETRY_COOLDOWN_MS` cooldown, same key) to the account or payer wallet rather than the character. An amount-exact check alone does not close it, since the attacker controls both listings' prices.

This is flag-gated OFF and behind the documented maintainer decision, so nothing ships broken today. It should not be the thing we discover at enable time.

## Should fix

**2. The Merchant's cut is floored per transaction, so seller proceeds depend on the buyer's chosen granularity.** `src/sim/market.ts:667`: `proceeds = Math.floor(cost * (1 - MARKET_CUT))` with `cost = qty * pricePerUnit`. A 200-stack at 21c/unit pays the seller `floor(4200 * 0.95) = 3990` bought whole, but `200 * floor(21 * 0.95) = 3800` bought one unit at a time. 190c destroyed, and the buyer pays the same either way, so this is purely a buyer-controlled tax on the seller. The degenerate case is worse: at 1c/unit, `floor(0.95) = 0`, so a stack bought one unit at a time pays the seller nothing at all while the goods still move.

Partial buys are new design space (classic stacks were whole-lot), so there is no fidelity constraint here. Suggest accruing the cut against a running per-listing gross and paying `floor(gross * 0.95) - alreadyPaid`, or rounding the cut per-unit at list time so it is granularity-independent. Worth a test asserting `buy(20)` and `20x buy(1)` pay the seller identically.

**3. `reclaimSoulboundListings` silently destroys the seller's deposit.** `src/sim/market.ts:1663-1677` returns the goods to the seller's collection box and correctly refunds a standing bid's escrow, but never credits back `l.depositPerUnit * l.count`. Forfeiture is the intended spam brake on expiry and on cancel, both of which the seller chose. This path fires because a content patch marked an already-listed item soulbound, so the seller had no choice and their staked copper vanishes at boot with no message. One line next to the bid refund, plus a pin alongside the existing bid-escrow reclaim assertion in `market_auction.test.ts`.

## Base branch, and a real conflict waiting behind it

The PR targets `release/v0.27.0`, but `release/v0.28.0` exists and is the active integration base. It is 12 commits ahead, and two of them collide with this change concretely, so this is not just a formality:

- `22d52828e` ("two-phase the trade swap so same-item offers cannot cross-contaminate") rewrote the swap in `src/sim/social/trade.ts`, the same function this PR rewrote independently. Your version looks structurally sound to me and closes the hazard differently and arguably better, by taking instances via explicit payload selectors (`takeInstanceSlot`) and counting plain rows fungible-only, rather than relying on `removePreferFungible`'s index scan. But the three regression tests that commit added will be in the base after the rebase and will not pass as written: they call the old 4-argument `tradeSetOffer` (yours takes 6) and they assert that a plain `{itemId, count}` row moves an instanced copy, which your `countFungibleItem` gate deliberately no longer does. They need reconciling to your semantics, not deleting.
- `ad8ec56d0` removed `deedCount` from `src/world_api/deeds.ts`. Your tree still has it (`src/world_api/deeds.ts:42`) and this PR moves the `IWORLD_MEMBERS` pin to 235, so the pin and the facet arrays will need re-deriving after the rebase.

Please rebase onto `release/v0.28.0` and re-run the gate there. Given how much of this diff is the stacked `#2001` base, that is also the moment the delta gets readable.

## Minor

**4.** `server/trade_db.ts:190-192` documents the market context marker as `market-<listingId>-<purchaseSeq>`, but the code writes the realm-scoped `market-<realm>-<listingId>-<purchaseSeq>` (`server/market_settlement.ts:223-225`). The realm scoping is correct and load-bearing (the `ON CONFLICT (context)` index is global across realms); only the comment is stale.

## Clean

Server authority on the new commands is solid: I could not find a forgeable value. Price, quantity, duration tier, ownership, soulbound, self-buy, and the pending lock are all revalidated in the sim, cost is recomputed server-side, and denomination is read from the listing rather than the buy message. Amount parsing is all-BigInt with a strict allowlist and no float ever touches a token amount. Rails are genuinely fail-closed in the sim, not just the UI. All new SQL is parameterized. No secrets in `.env.example`, no `ALLOW_DEV_COMMANDS`. Determinism holds: no banned time or randomness APIs in sim logic, expiry is `ctx.time`-derived, and every new sort terminates in a unique-id tiebreak. Parity is clean across both `IWorld` implementations, the wire, the events, and the offline letters.

Happy to re-review quickly once the rebase and items 2 and 3 land. Item 1 can reasonably be tracked as a gate on the WOC flag rather than on this merge, if you would rather land the gold economy now.

PR #2024, opened 2026-07-16
feat(hud): add optional enhanced combat text

Humpalumps:
## Summary

Adds an optional, higher-impact floating combat text presentation inspired by classic MMO combat readability while preserving the existing presentation as the default.

Players can opt in through **Settings → Interface → Combat Text Style → Enhanced**. The setting is persisted with the existing settings system and immediately applies a root data attribute. **Classic remains the default**, so this does not force a visual change on existing players.

The enhanced presentation differentiates combat outcomes at a glance:

- standard outgoing damage uses outlined parchment numerics;
- critical hits use a larger accent-colour pop and brief burst treatment;
- incoming damage, healing, XP, parry, and block presentation states receive distinct semantic colours and motion;
- mobile sizing keeps the hierarchy readable without covering the play space;
- low-effects and reduced-motion modes keep all information visible with static emphasis instead of movement.

This deliberately follows the root `DESIGN.md` style guide rather than introducing an unrelated arcade font or colour system:

- uses the in-world `--font-ui` role (Alegreya Sans) for combat values;
- uses semantic colour tokens and named FCT typography/motion tokens;
- keeps numeric output on the shared `formatNumber` path;
- preserves the pooled FCT painter and existing fairness/cap pipeline;
- supports both the in-game reduced-motion setting and `prefers-reduced-motion`;
- localizes the new settings labels and defensive outcome text through the English catalog and required non-Latin overlays.

The v0.27.0 simulation currently does not emit a live `block` damage outcome. This PR prepares the renderer/HUD presentation state without changing combat mathematics or manufacturing block events.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/fct_painter.test.ts tests/fct_mobile_css.test.ts tests/fct_core.test.ts tests/fct_event.test.ts tests/settings.test.ts tests/options_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/i18n_completeness.test.ts tests/localization_fixes.test.ts` — 169 passed, 3 skipped.
  - `npm run check:types` — TypeScript passed; Svelte reported 0 errors and 0 warnings.
  - `npm run build` — production client build and generated-artifact checks passed.
  - `npm run build:env` — passed.
  - `npm run build:server` — passed.
  - `npm run security:gate` — passed.
  - `npm run sfx:check` — passed.
  - `npm run test:browser` — 63/64 passed; the remaining unrelated Armory landscape assertion reports an existing 2.625 px card-height overflow (`393.625 <= 391`).
- Manual steps:
  - Ran the game locally from this branch.
  - Switched between Classic and Enhanced under Settings → Interface and confirmed the setting applies immediately and persists.
  - Exercised standard hits, critical hits, incoming damage, heals, XP, and defensive renderer states in the local deterministic showcase.
  - Confirmed enhanced mobile sizing and static low-effects/reduced-motion styling.

## Screenshots / recordings

The obsolete v0.27 recording has been removed. Current v0.32 test evidence is available on #2548.

---

### Quality

- [ ] **The full gate passes.** Focused tests, types, security, SFX, and all builds pass. The full inherited characterization suite and the unrelated Armory browser assertion are recorded above rather than represented as green.

### Cross-platform

- [ ] **Tested on desktop and mobile.** Tested on desktop and through the repository's mobile CSS/browser coverage; not tested on a physical phone.
- [x] **Accessible.** Information remains present in low-effects mode and under both the game reduced-motion setting and OS `prefers-reduced-motion`.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Settings and outcome strings use catalog keys, with the required non-Latin overlay fills.
- [x] Numbers, money, dates, units, and percents go through the i18n formatters (`formatNumber`, `formatMoney`, `formatDateTime`, `Intl`).
- [x] Player-facing outcome text is localized at the client boundary and `tests/localization_fixes.test.ts` passes.

### Hygiene

- [x] Generated localization files were produced with `npm run i18n:build`; other generated files were not hand-edited.


EnriqueGF:
@FernandoX7 maybe we should make this default and remove the optional?

EnriqueGF (commented):
Senior review of #2024 (enhanced combat text). Presentation-only, opt-in, Classic default preserved. This is a clean, well-scoped, well-tested change. Nice work.

## What I reviewed and ran
- Read the full diff and the FCT data path (`fct_core.ts`, `fct_event.ts`, the `hud.ts` combat-event switch, the settings/apply wiring, and the new CSS in `hud.css` / `hud.mobile.css` / `tokens.css`).
- `tsc --noEmit` (native): clean.
- `fct_core`, `fct_event`, `fct_painter`, `fct_mobile_css`, `settings`, `options_view`: 114 passed.
- `localization_fixes` (S3), `i18n_completeness`, `i18n_resolved_equivalence`, `architecture`: green (after `npm run i18n:gen`).
- Re-ran `npm run i18n:gen`: zero diff, so the committed generated locale files match the generators.

## Invariants checked
- No `src/sim/` or server changes: determinism, sim purity, and server authority are untouched. `formatNumber` on the FCT text is a correct client-side i18n improvement.
- Fairness/gameplay-neutrality holds: the number is always spawned and shown; low-effects makes crit static and hides the halo/burst pseudo-elements; both reduced-motion channels keep the value visible with a static fade (and `fct-static` reuses the same `translate: -50% -50%` centering, so Classic position does not shift).
- i18n M16 handled correctly: the five non-Latin overlays are the sanctioned exception, and the pending Latin locales fall back to English at PR tier. No em/en dashes or emojis.

## Minor findings (non-blocking)

1. Enhanced mode shortens the XP/rested-XP floater lifetime and changes its motion. The base rule `:root[data-combat-text="enhanced"] .fct` (specificity 0,3,0) overrides `.fct-xp` / `.fct-rested-xp`, so in Enhanced those floaters get `animation-name: fct-enhanced-hit` and `animation-duration: var(--dur-fct)` (1250ms) instead of the XP pop at 1.6s. But the painter keeps XP entries for `FCT_XP_TTL_MS` (1800ms), so the XP number fades to opacity 0 at 1250ms and then lingers invisible for ~550ms, and it pops with the hit-kick rather than the gentler XP rise. Classic and both reduce-motion blocks correctly give XP the longer `--dur-fct-xp`. Consider excluding `.fct-xp` / `.fct-rested-xp` from the enhanced base override (or giving them `--dur-fct-xp`), same for the mobile enhanced rule. Purely cosmetic; XP is not combat-actionable, so no fairness concern.

2. The `parry` presentation is currently unreachable, same as `block`. The v0.27.0 sim never emits `kind: 'parry'` (only `hit`/`miss`/`dodge`/`resist` are emitted; `parry` is in the `SimEvent` union but has no emit site). So the new "Parry" floater and the `hud.combat.parried` log line (and the `parried` strings now carried across every locale) will never fire in this release. That is fine and forward-looking, but the PR body only discloses this for `block`. Worth a one-line note that `parry` is also presentation-prep so a future reviewer does not go hunting for the sim path.

3. Scope note (intended, just flagging): the new global `@media (prefers-reduced-motion: reduce)` block now applies the static-fade to Classic FCT under OS-level reduced motion too. Previously only the in-game `reduce-motion` class touched `.fct`, and it used a 0.01ms duration that effectively made the number vanish instantly. This is a strict accessibility improvement (info stays visible, position unchanged), but it does change Classic behavior for OS-reduced-motion users, so it is worth being explicit that it is intended.

None of these block merge. Finding 1 is the one I would most like addressed before this ships, since it is a visible regression of the Enhanced XP floater relative to Classic.

EnriqueGF (changes requested):
Thanks for this, the presentation work is careful and it clearly respects the seams: the FCT data path, the pooled painter, the cap/TTL pipeline, and the tier knobs are all untouched, and the enhanced styling is fully gated behind `:root[data-combat-text="enhanced"]` so an unset attribute falls through to Classic. The i18n work is done properly too.

## What I reviewed and ran

Read the full diff, plus `src/ui/fct_painter.ts`, `fct_core.ts`, the `main.ts` settings pipeline, and the root / `src/` / `src/ui/` CLAUDE.md contracts.

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/fct_core.test.ts tests/fct_event.test.ts tests/fct_painter.test.ts tests/fct_mobile_css.test.ts tests/settings.test.ts tests/options_view.test.ts tests/i18n_completeness.test.ts`: 399 passed.
- `npx vitest run tests/hud_perf_budget.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/painter_host.test.ts tests/ui_tier_knobs.test.ts`: 90 passed, 3 skipped.
- `npm run i18n:gen` then `tests/localization_fixes.test.ts`: all substantive assertions pass (one pre-existing 5s timeout on the unrelated `admin DICT is complete across all locales` case, not caused by this PR).

Verified separately and found clean: no `Math.random`/`Date.now`/`performance.now` added to `src/sim/`, no `IWorld` surface change so no parity pin needed, no `ALLOW_DEV_COMMANDS` or secrets, no em dashes, en dashes, or emojis in the added lines. The `--dur-fct: 1250ms` / `--dur-fct-xp: 1800ms` tokens match `FCT_TTL_MS` and `FCT_XP_TTL_MS` in `fct_core.ts` exactly, which is the right way to pin that. The five non-Latin overlay fills are real translations with placeholders preserved, so the M16 exception is satisfied rather than worked around.

## Findings

**1. Wrong base branch (blocking for merge).** This targets `release/v0.27.0`, but `release/v0.28.0` exists on the remote and is the active integration base. Per the root CLAUDE.md the work should be based off the latest release branch. Please rebase onto `release/v0.28.0` and retarget; some of the FCT and reduce-motion CSS around your hunks may have moved.

**2. `formatNumber` changes Classic FCT, and it is not mentioned in the PR description (medium).** In `src/ui/hud.ts`, damage-done, damage-taken, heal, XP, and rested-XP floaters moved from `${ev.amount}` to `formatNumber(ev.amount)`. `formatNumber` (`src/ui/i18n.ts:432`) passes no options, so `Intl.NumberFormat` applies its default `useGrouping: true`. In `en`, a 1,234 damage crit now floats as `1,234` where it previously floated as `1234`, and this applies in Classic mode, not just behind the new opt-in. That contradicts "Classic remains the default, so this does not force a visual change on existing players", and classic-era FCT does not group thousands. `src/ui/CLAUDE.md` names the fix directly: "To keep English byte-identical to a historical hand-rolled form, pass `useGrouping: false`". Either pass `{ useGrouping: false }`, or keep grouping and call it out explicitly in the PR body as an intended fidelity change so it is a decision rather than a side effect. Grouping also widens the node, which is worth a thought against the AoE-burst layout.

**3. Reduced-motion behavior also changes for Classic (low, and an improvement, but undisclosed).** The new `body.reduce-motion .fct` / `@media (prefers-reduced-motion: reduce) .fct` rules are unscoped by `data-combat-text`, so they apply in Classic too. Before this PR, the blanket `body.reduce-motion *` rule (`animation-duration: 0.01ms !important`) made FCT numbers flash and vanish instantly; your rules give them `fct-static` for the full TTL. That is strictly better and I would keep it, but same note as above: it is a Classic-affecting change and the PR reads as if Classic is untouched. Worth a line in the description, and it deserves its own assertion so the fix does not silently regress.

**4. Crit burst will not replay on an evicted-and-reused pool node (low, enhanced-only).** `fct_painter.ts` documents that an over-cap eviction reuses a still-attached node and forces the animation restart with the reflow trick in `restartAnimation` (lines 288 to 295), which writes `animation: none` on the host node and then restores it. That does not reach `::before`/`::after`: pseudo-elements carry their own `animation`, so the `fct-enhanced-halo` and `fct-enhanced-burst` animations, both `forwards`, stay parked at their end state (`opacity: 0`) on a recycled node. The practical effect is that during a genuine over-cap AoE burst, exactly the moment the crit pop is meant to read, enhanced crits lose their burst treatment. The number itself is still fully legible, so no information is shed and the fairness invariant holds. Cheapest fix is to move the burst onto the same restart path, or drive it from a class the painter toggles per spawn.

**5. `block` is dead surface (nit, no action needed).** `block` is added to `FctKind`, `FctColorToken`, `fctSpawnShape`, and CSS, but the sim's damage kind union (`src/sim/types.ts:2345`) is `'hit' | 'miss' | 'dodge' | 'parry' | 'resist'`, with no `block`, and nothing in `src/ui/hud.ts` spawns it. You are upfront about this in the description and it manufactures no events, so I am fine with it landing as prepared surface. Just flagging that it is currently unreachable, so its CSS and color tokens are untested by anything but the type system. `parry` is in the same position (declared in the sim union but never emitted today), though the pre-existing parry SFX handler at `hud.ts:7736` shows that was already the established pattern, so wiring the FCT and combat-log side is consistent.

## Verdict

Requesting changes for (1) the base branch and (2) the undisclosed Classic grouping change, both of which are quick. Findings 3 and 4 are worth addressing in the same pass but I would not hold the PR on them alone. Nice, well-scoped work otherwise.

EnriqueGF (changes requested):
Thanks for this: the opt-in framing is right, the pooled painter and the fairness/cap pipeline are left intact, and the design-token discipline (no arcade font, semantic colour tokens, `formatNumber` on the shared path) is exactly what `DESIGN.md` asks for.

## What I reviewed and ran

Read the full diff, the root `CLAUDE.md`, and the local `src/ui/` and `src/styles/` conventions. Ran on the branch myself:

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/parity`: pass (456 passed, 4 skipped). The `localization_fixes` failure I first hit was only a missing `src/ui/i18n.status.json`; after `npm run i18n:gen` it is green, so not yours.
- `tests/fct_core.test.ts`, `fct_event.test.ts`, `fct_painter.test.ts`, `fct_mobile_css.test.ts`, `settings.test.ts`, `options_view.test.ts`, `i18n_completeness.test.ts`: 146 passed.

Also confirmed clean: no `src/sim/` changes, so determinism is untouched; the `i18n.locales/` edits are the sanctioned M16 non-Latin fills matching the existing `playerFrameScale`/`targetFrameScale` precedent, not hand-edits of translated overlays; no secrets, no `ALLOW_DEV_COMMANDS`; no em dashes, en dashes, or emoji in the added lines.

## Findings

**1. Enhanced mode clobbers the XP and rested-XP floaters (medium, in the feature's own path).**

`:root[data-combat-text="enhanced"] .fct` has specificity (0,3,0) and sits later in `hud.css` than `.fct-xp` / `.fct-rested-xp` at (0,1,0), so in Enhanced it wins all three of their declarations:

- `font-size`: 20px becomes `--font-size-fct-enhanced` (28px), so XP loses its distinct size step and now reads at damage size.
- `animation-name`: `fct-xp-pop` becomes `fct-enhanced-hit`, so the XP-specific pop is gone.
- `animation-duration`: 1.6s becomes `--dur-fct` (1250ms), while `FCT_XP_TTL_MS` is 1800ms (`src/ui/fct_core.ts:150`). The XP number reaches `opacity: 0` at 1250ms and the pooled node then sits invisible for the remaining 550ms.

You already solved exactly this for the reduced-motion arm with `--dur-fct-xp`, so the fix is the same shape: re-assert `.fct-xp, .fct-rested-xp` inside the enhanced block (or scope the base enhanced rule so it does not swallow them). Same cascade also outranks `:root[data-fx-level="low"] .fct-xp` (ties at (0,3,0), enhanced is later), so Enhanced plus low tier silently loses the low-tier XP motion shed.

**2. `parry` and `block` colour tokens have no Classic rule (medium, latent).**

`FctColorToken` grows by `parry-self/other` and `block-self/other`, and `hud.ts` now routes `ev.kind === 'parry'` into the avoidance path. But `.fct-parry-*` and `.fct-block-*` are only declared under `:root[data-combat-text="enhanced"]`. `src/sim/types.ts:2345` already declares `parry` in the damage-event union, so nothing but "the sim does not emit it yet" is keeping this dead: the moment it does, Classic (the default) renders an uncoloured word.

The byte-faithful guard cannot catch this either, because `PINNED` in `tests/fct_painter.test.ts` is a hand-maintained map and the four new tokens were not added to it. Worth giving them Classic hexes and pinning them in the same change.

**3. Two default-path changes the PR body says do not happen (low, mostly a description and coverage gap).**

The body states "Classic remains the default, so this does not force a visual change on existing players", but two changes land for Classic users regardless of the setting:

- Reduced motion: `body.reduce-motion .fct` went from `animation-duration: 0.01ms` to a full static lifetime. This is a genuine fix (the old rule made combat numbers vanish instantly, which is the kind of "hides something a player acts on" the graphics-fairness invariant forbids), so I am glad it is here. It just deserves its own line in the body and a test on the Classic arm, not only the enhanced one.
- `formatNumber()` on damage, heal, and XP floater text. Correct per the i18n invariant, but it changes Classic rendering (1234 becomes "1,234") and no test pins the formatted output.

**4. Base branch.** This targets `release/v0.27.0`; the current release branch is `release/v0.28.0`. Needs a retarget and rebase before merge.

Finding 1 is the one I would want fixed before this goes in; 2 and 3 are small and can ride along. Happy to re-review once the cascade is sorted.

PR #2029, opened 2026-07-16
feat(realm): browser-only + pay-to-win realm flags (RiverBoat Phase 0)

awidearray:
## Summary

Phase 0 of the RiverBoat realm (the opt-in pay-to-win / casino / web3 realm): the realm
infrastructure and browser-only gating, all shipping dark. Adds an optional flags field to
`REALMS` directory entries (`web` = browser-only, `p2w` = pay-to-win label) plus a
`REALM_FLAGS` env for the single-realm default. With the flags unset, every realm behaves
exactly as today; this PR is inert until a realm sets them.

What the flags do:
- **`web` (browser-only):** both `/api/realms` arms drop `web`-flagged realms, and their
  character-count keys, for native (Capacitor) and desktop (Electron) clients. This is the
  only gate that reaches app binaries already in the field, which render the directory
  unfiltered and have no OTA path. A web-only realm process also refuses the native app and
  the packaged Electron shell at the bot gate and the WebSocket handshake (new
  client-localized wire literal). Genuine browsers on the Vite dev origins stay admitted
  (the refusal keys off the packaged `app://` origin only, not the shared desktop-origin set).
- **`p2w` (pay-to-win label):** the realm picker badges the entry (with an 18+ tooltip), app
  shells filter `web` realms client-side as belt-and-braces, and `p2w` realms are excluded
  from all three cross-realm boards (lifetime XP, guilds, and the Renown/deeds board) so
  pay-boosted characters never rank on clean-realm boards.

Also plumbs `REALM_NAME/REALM_TYPE/REALMS/REALM_FLAGS/REALM_RESET_TZ` through docker-compose
for standing up a second realm process, and adds a pinned test asserting no server code path
ever changes a character's `realm` (the one-way economy membrane).

Design context: this bootstraps the realm the RiverBoat plan builds on; the money rails
themselves are later, separately-gated work.

## Related issues

Part of the RiverBoat realm effort.

## Type of change

## How was this tested?

- Commands: `npx tsc --noEmit`; `npm run i18n:gen`; full `npx vitest run` (14490 passed, 0
  failed, 24 skipped, on the branch merged up to release/v0.27.0); `npx @biomejs/biome check`
  on changed files.
- Manual steps: verified the `/api/realms` golden fixture regenerates with the flags field;
  confirmed the new realm/platform-guard and realm-visibility suites cover the browser vs
  app-shell split, the p2w board exclusion (including the own-realm-flags union), and the
  Vite-dev-origin carve-out.

## Screenshots / recordings

Realm-picker badge screenshots to follow. Concept art for the realm this PR bootstraps is
posted in a comment below.

### Quality
- [x] The full gate's test + typecheck + biome portions pass; decisive tests added for every
  new behavior (flag parsing, dual-arm directory filter, app-shell refusal, board exclusion,
  no-realm-transfer pin).

### Localization (i18n)
- [x] New player-visible strings are `t()` keys in the English catalog with the five non-Latin
  M16 fills (`realm.p2wBadge`, `realm.p2wTip`, `loading.webOnlyRealm`); no locale overlays
  hand-edited beyond those fills; generated tables regenerated via `npm run i18n:gen`.


awidearray:
### RiverBoat concept art

Visual direction for the realm this PR bootstraps: a moored paddle-steamer casino. Walnut
planking, brass fittings, crimson velvet, green baize, lantern light. Low-poly to match the
game's look.

**The Grand Saloon (casino floor)** — card tables, a roped dueling ring, rune-topped slot
cabinets, and the cashier cage at the bow.

RiverBoat casino floor concept

**The Bookmaker's Parlor (sportsbook deck)** — a wall of screens showing live arena matches
from across every realm, a runic odds board, and a ticket-window bookmaker for live parimutuel
betting on who wins.

RiverBoat sportsbook deck concept

Concept art, not final assets.

FernandoX7 (commented):
Solid Phase 0. The gating is genuine defense-in-depth (server directory filter for the shipped binaries, WS handshake refusal before any DB work, turnstile inversion, client belt-and-braces), both `/api/realms` arms shape through the one shared `realmsVisibleToRequest` with a pin so the twins cannot drift, and the counts key-filter closes the realm-name leak through the `characters` map. Back-compat is right in both directions (client `flags?` optional, unknown tokens dropped), and the new tests pin literals rather than the shared constants.

Things I verified clean, so you do not have to re-check them: every `realm` column is `TEXT NOT NULL`, so the new `!= ALL` predicates have no NULL-drop trap and the empty-array default is exact parity with the old queries; client app-shell detection is build-time/UA (`src/client_origin.ts`), so a real browser is never filtered client-side; the WS literal is byte-exact with the matcher and reaches it through the same `fatalOverlay(userFacingApiError(...))` seam as `tooManyConnections`; all three client fetch sites plus the no-arg `showRealmList()` path filter; `play.html` shares `shell.css`, so the badge styles reach `/play`; the generated i18n slices are mechanical regen; no other `REALM_DIRECTORY` or `characterCountsByRealm` consumer leaks unfiltered.

Findings:

- **should-fix** `server/db.ts` (`topLifetimeXp`, `topGuilds` global arms): the new `realm != ALL($2::text[])` predicate and its param are not pinned by any test (only the deeds board got the SQL-shape pin, `tests/server/deeds_board_sql.test.ts:145`), and no test anywhere exercises a NON-empty exclusion list end to end: the pg differential calls `deedsBoardRanked` with the default empty list (`tests/server/deeds_board_sql.test.ts:346`). Dropping the clause on the XP or guild arm would pass the whole suite, which means the feature's core promise is unverified. Cheap close: predicate + param assertions in `tests/server/leaderboard_moderation.test.ts` (it already captures the SQL), plus one pg-gated two-realm case asserting a named realm actually vanishes from the ranked output.
- **should-fix** `src/ui/api_error_i18n.ts:328`: the new matcher arm has no runtime test, while the exactly parallel `tooManyConnections` arm is pinned (`tests/main_api_error.test.ts:164`). A typo in either side of the byte-exact contract would silently regress the refusal overlay to raw English. One line closes it: `expect(userFacingApiError('this realm can only be entered from a web browser')).toBe(t('loading.webOnlyRealm'))`.
- **should-fix** `server/main.ts` legacy `/api/realms` arm: the counts key-filter is re-implemented inline, and that copy is covered by neither the dual-arm pin (`tests/realm_web_only_pins.test.ts` greps only for `realmsVisibleToRequest`) nor the parity suite (it replays the noauth path, where counts are empty). The RouteDef twin (`readRealms`) is tested; the legacy twin can drift. Either extend the pin to assert the counts shaping too, or have the legacy arm call `readRealms`.
- **should-fix** `server/realm_platform_guard.ts` (`globallyExcludedRealms`) + `server/main.ts:441`: "p2w realms never rank on clean boards" is an operational invariant, not code-enforced. Each process excludes only what its own `REALMS` entry carries, so a clean-realm process deployed with a directory missing RiverBoat's `p2w` token will rank RiverBoat characters on its global boards. The code comment acknowledges this; the PR body states the absolute. Worth a DEPLOY.md callout on multi-realm `REALMS` consistency, or a boot-time warning when a process's own flags disagree with the shared directory entry.

Nits:

- `server/main.ts:1112,1198`: an app shell hitting REST login/register on a web-only realm gets the generic `verification failed, please try again`, while the WS path has the dedicated message. Cosmetic (the realm is hidden from app shells anyway), but a keyed error would match the WS arm.
- `tests/realm_web_only_pins.test.ts:34`: the membrane pin's comment promises "no code path may ever re-home a character", but the regex only sees literal `UPDATE characters` SET clauses (not `ON CONFLICT DO UPDATE` or dynamic SET lists). Fine today (the one `SET realm = EXCLUDED.realm` is on `character_leases`), but soften the comment or extend the regex so the pin's claim matches what it guards.

Process notes, not code findings: no CI check runs exist on the head commit, so CI needs a run before merge (the touched suites plus `tsc` were re-run green in two clean worktrees during this review, so I expect it to pass). And the realm-picker badge is a visual change, so the screenshots the PR body promises need to land in `docs/screenshots` and be referenced from the body before this merges.

EnriqueGF (commented):
Reviewed the full diff, ran the checks locally, and verified each finding against the real code paths.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/realm_platform_guard.test.ts tests/realm_visibility.test.ts tests/realm_web_only_pins.test.ts tests/server/leaderboard.test.ts tests/server/deeds_board_sql.test.ts tests/player_card_server.test.ts`: 135 passed, 3 skipped.
- Guard suites: `tests/architecture.test.ts`, `tests/world_api_parity.test.ts` pass; `tests/localization_fixes.test.ts` passes after `npm run i18n:gen` (it only failed on the missing generated `i18n.status.json`, not on this change).
- `npx @biomejs/biome check` on the 11 changed source files: warnings only, no errors and no format diffs.

## What holds up

The gating design is sound and the test coverage is genuinely good (both arms asserted, the Vite-dev-origin carve-out pinned, the WS refusal proven to short-circuit before `accountForToken`). Specifically verified:

- SQL is fully parameterized, and `realm` is `NOT NULL` on every table involved, so the new `!= ALL($n::text[])` predicates carry no NULL-elimination trap. The empty-array default is a true no-op (`x != ALL('{}')` is TRUE), so unflagged deployments keep the exact current result sets.
- The Renown board filters both roll-up CTEs (`per_deed` and `per_char`) consistently, and `display` derives from the already-filtered `per_char`, so there is no half-filtered path.
- The turnstile inversion cannot touch browsers: it fires only on `capacitor://localhost` or `app://worldofclaudecraft`, neither of which a browser can send. On a realm without `web` the function is behaviorally identical to before, and the new default parameter leaves the existing 3-arg call sites unaffected.
- Both `/api/realms` arms and all three client fetch sites route through the shared filters, and the counts-map key-filter closes the name leak on both arms. `enterRealmFlow` correctly resolves the remembered realm against the *filtered* list, so a remembered web-only realm will not auto-enter from an app shell.
- i18n is correct: English catalog entries plus the five non-Latin M16 fills, and the matcher literal in `api_error_i18n.ts` is byte-exact with `WS_AUTH_ERROR.webOnlyRealm`.

## Findings

**1. (medium) Base branch.** This targets `release/v0.27.0`, but `origin/release/v0.28.0` exists and is the current integration base. Worth retargeting before merge.

**2. (medium) The `web` entry gate is narrower than the comments claim.** The header in `server/turnstile.ts` says an app-store build "can never authenticate against this realm's origin". Two paths escape it:

- The Discord login routes deliberately skip `passesTurnstile` (documented at `server/discord.ts:1140`), so a native or Electron shell can still mint a session token against a web-only realm's origin via `/api/auth/discord/native/exchange`.
- `/api/characters` (GET and POST, `server/characters.ts:573-584`) carries only `activeGuard` plus the rate limit, with no client-class check. Tokens are account-wide, so an app shell holding a token minted on any realm can list and *create* characters on the web-only realm, and `createCharacterCapped` stamps them with this process's `REALM`.

The store-compliance gate itself still holds: the directory filter plus the client filter mean the app never renders the realm, and `ws_auth` refuses world entry, so nobody can actually play. But the belt-and-braces layer has holes, and the comment overstates what is enforced. Either gate the realm-scoped REST surface on `refusesAppShellClient` centrally, or soften the comment to say only world entry is refused.

**3. (low/medium) A `web`-only realm that is not also `p2w` re-leaks its name through the global boards.** `globallyExcludedRealms` keys on `p2w` alone. With `REALMS=Hidden=https://hidden.example.com=Normal=web`, an app-shell client fetching the cross-realm lifetime-XP or guild board gets rows carrying `realm: "Hidden"`, which is exactly the name `realmsVisibleToRequest` just suppressed. The intended `web+p2w` config masks this, but the flags are independent by design, so this is worth either excluding `web` realms from the boards too or documenting that the two flags travel together.

**4. (low) Origin-class asymmetry between the two gates.** `realmsVisibleToRequest` uses the broad desktop set (which includes `http://localhost:5173`), while `webOnlyRealmRefuses` uses the packaged origin only. So a client presenting the Vite dev origin is hidden the realm yet admitted to it. I checked whether this bites in dev: `api.realms()` goes to the page origin through the Vite proxy, so a real browser sends no `Origin` header and sees the full directory. Not reachable in practice, but the split is subtle enough to deserve a line in the module header saying why the two arms intentionally differ.

**5. (low) Docs.** `.env.example` and `docker-compose.yml` are updated, but the `realm.ts` row in `server/CLAUDE.md` still lists only `REALM`, `REALM_DIRECTORY`, and `REALM_ORIGINS` from `REALM_NAME`/`REALMS`. `REALM_FLAGS` and `OWN_REALM_FLAGS` should be mentioned there.

**6. (low) Screenshots.** The PR body says the realm-picker badge screenshots are still to follow. The `rn-p2w` badge is a visual change, so the template's before/after requirement applies before this merges.

None of these is a correctness bug in what the PR actually gates, so I have no objection to the approach. Items 1 and 6 are merge blockers by repo convention, and item 2 is worth resolving one way or the other (code or comment) so the next person does not over-trust the auth-layer refusal.

EnriqueGF (commented):
Follow-up after a closer pass on the database side. Three additional items, all verified against the tree.

**7. (medium) The exclusion behavior itself has no test.** `grep -rn "excludeRealms" tests/` returns nothing: no test passes a non-empty list to `topLifetimeXp`, `topGuilds`, or `deedsBoardRanked`. The coverage that exists is the unit tests of `globallyExcludedRealms` (which prove the *name set* is computed correctly) plus the SQL-literal pin in `deeds_board_sql.test.ts` (which proves `$4` appears in both CTEs). Neither exercises the predicate actually removing rows, and the pg-gated differential calls `deedsBoardRanked` with the default `[]` at every call site. So the one behavior this PR exists to deliver, a p2w realm dropping off the shared boards, is asserted only indirectly.

Worth adding a non-empty-exclusion case to the pg-gated differential (fixture with characters on two realms, asserting the excluded realm contributes nothing to renown, count, completion time, or the display pick), and a mocked-pool shape assertion on the two board queries.

**8. (low, pre-existing, but the PR touches the comment) The global lifetime-XP board is not served by the index its comment claims.** `characters_lifetime_xp_global` is built on the bare expression `((state->>'lifetimeXp')::bigint)`, while the query's `ORDER BY lifetime_xp DESC` resolves to the select-list `COALESCE((state->>'lifetimeXp')::bigint, 0)`. Postgres matches expression indexes by expression equality, so `COALESCE(x, 0)` will not match an index on `x`, and the same mismatch applies to the `WHERE COALESCE(...) > 0` qual. If that is right, both board arms are already a scan plus top-N sort, and the header comment at `server/db.ts:2886` ("Both paths sort on the indexed lifetime-XP expression") is inaccurate. Note `lifetimeXpStanding` uses the bare expression and *does* match, which is likely why this went unnoticed.

This is reassuring for the PR rather than alarming: there is no index-ordered plan for the new filter to break, so the added qual cannot regress the plan. But since this change extends that comment block, it is a good moment to correct it. The actual index fix is a separate change. One `EXPLAIN` pair (empty vs non-empty `$2`) on `npm run db:up` would settle both this and whether the empty-array default folds away entirely.

**9. (low) `tests/server/helpers/fake_db.ts` was not updated.** `LeaderboardDb` still declares `topLifetimeXp(limit?, opts?: { global?: boolean })` and the same for `topGuilds`. This typechecks only because TypeScript method parameters are bivariant, so the drift is silent, and it means the `FakeDb` handler tests cannot pin the new parameter even if you wanted them to. Widening both to include `excludeRealms?: readonly string[]` keeps the fake honest.

Also worth a line in the PR body or next to the constant: `P2W_EXCLUDED_REALMS` is resolved at module load, so flagging a realm `p2w` takes effect on the global boards only after every realm process restarts, and a partially-restarted fleet will serve divergent global boards until the rollout finishes.

Nothing here changes my read of the approach, which is sound. Item 7 is the one I would want addressed alongside items 1 and 6 from my earlier comment.

PR #2034, opened 2026-07-16
feat(riverboat): Phase 1 compliance substrate (stacked on #2029)

awidearray:
## Summary

Phase 1 of the RiverBoat casino realm: the compliance + shared-substrate layer both
the casino floor and the sportsbook deck sit on. Everything here is inert until a realm
sets the casino flags, and every money-touching path is fail-closed. No player-facing
strings, no gameplay change on any existing realm.

**Stacked on the RiverBoat Phase 0 PR (#2029).** Review this after #2029; the diff below
is the compliance layer on top of the realm-flags foundation.

What lands:

- **`server/fairness.ts`** - the commit-reveal engine (sha256 commit published before
  play, domain-separated per-game digests, reveal for public verification) that every
  house-banked game will hang off. Extracted from the proven daily-spinner implementation.
- **`server/realm_features.ts`** - a frozen boot-time casino feature bundle derived from
  env, advertised on `GET /api/status` (both the migrated RouteDef arm and the legacy
  arm, identically) so the client can self-hide UI a realm does not run. Advisory only:
  every money route re-checks server-side. A typed `RealmFeatures` seam is surfaced on the
  client `realmStatus()`.
- **`server/regulated_gate.ts`** - the geo + OFAC + counsel gate every casino money route
  will call, generalized from the token-launchpad gate. Hardening over the original: an
  explicit `SOLANA_CLUSTER` resolution replaces the mainnet-hostname regex (a custom
  mainnet RPC could otherwise silently disable the gate); a per-surface
  `RIVERBOAT_COUNSEL_SIGNOFF` distinct from the securities sign-off; a separate
  gambling-restricted country list unioned onto wager routes only; and a self-exclusion
  short-circuit.
- **`src/sim/parimutuel.ts`** - the pure parimutuel pool math extracted out of Vale Cup as
  a byte-identical MOVE (rake-free at `rakeBps=0`), rake-parameterized so the sportsbook
  can reuse it. Vale Cup is rewired to consume it; the parity gate and its betting tests
  confirm zero behavior change.
- **`server/rg_config.ts` / `rg_db.ts` / `rg.ts`** - the responsible-gambling substrate:
  env config, the DB schema and SQL (wired into `ensureSchema`, idempotent, FK to
  accounts), and the pure service. It owns age attestation and the of-age verdict, realm
  ToS acceptance, self-exclusion with permanent-irreversibility, and a rolling
  deposit/loss-limit verdict where a lower is instant while removing or raising a cap is
  held behind the cooldown, tracked per window.

Not in this PR (deferred, interdependent with the client attestation UI): the WS-join
entry-gate wiring and the REST setter endpoints for age/ToS/exclusion/limits. Those three
can only be safely enabled together with the client interstitial (enabling the gate before
the setters exist would lock players out of the realm), so they land as one later tranche.
The RG service and DB that back them are complete and fully tested here.

## Type of change

- [x] Feature: new functionality (dark-shipped infrastructure)

## How was this tested?

- Commands: `npx tsc --noEmit` under the repo's TypeScript 7 native compiler;
  `npx @biomejs/biome ci` on the changed files (zero errors); full `npx vitest run`.
- New/updated suites: `fairness` (14), `realm_features` (4), `regulated_gate` (15),
  `parimutuel` (12) plus the Vale Cup betting + parity gates proving the byte-identical
  extraction, `rg_service` (14, including the deposit-limit invariants), and the
  `realm_status` / `/api/status` golden updates for the new `features` field.
- An adversarial multi-lens review pass ran over the diff; its three confirmed findings
  are fixed in this PR: a limit set to 0 (the "unlimited" sentinel) is now held behind the
  cooldown as a raise rather than an instant uncap; held raises are tracked per window so
  staging one never discards another; and a dead import-coupling was removed from the gate.

### Quality
- [x] tsc + biome + the full test suite pass; decisive tests added for every new module,
  including the RG limit-raise invariants and the byte-identical parimutuel extraction.

### Localization (i18n)
- [x] N/A. This PR adds no player-visible strings (server infrastructure; the client
  `RealmFeatures` type carries no copy).


EnriqueGF (changes requested):
Reviewed the Phase 1 compliance substrate. This is careful, well-documented work: the fail-closed posture is real, the module boundaries are right (pure verdict logic separated from the HTTP/DB appliers), and the test suites are decisive rather than self-confirming. Most of what I checked came back clean. I have one confirmed correctness bug and a few hardening items below.

## What I checked and ran

- `npx tsc --noEmit`: clean.
- Guard tests: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts` all green (the S3 guard needs `npm run i18n:gen` first).
- New suites: `fairness`, `parimutuel`, `realm_features`, `regulated_gate`, `rg_service`, `realm_status`: 92 passed, 0 failed.
- `npx @biomejs/biome check` on the changed files: warnings only, no errors and no format diffs.
- Invariants: no `Math.random` / `Date.now` / `performance.now` added under `src/sim/`; `parimutuel.ts` is a clean leaf with no DOM/Three imports; all `rg_db.ts` SQL is parameterized; `RG_SCHEMA` is additive, idempotent DDL correctly sequenced after core `SCHEMA` for the `accounts(id)` FKs; no secrets and no `ALLOW_DEV_COMMANDS` change; no em dashes, en dashes, or emoji.

I also traced the Vale Cup extraction line by line against the original `vcupPlaceBet` and `settleBets`. **The "byte-identical move" claim holds.** Validation order is preserved (bad-stake, then closed-side, then over-cap, with the funds check still after the pool checks so an over-cap client never sees the error), the added `Number.isInteger` check is a no-op after `Math.floor`, `vcupBetNet` via `payout - stake` equals the original `winnings`, and Map iteration order keeps the settle deterministic. Nice extraction.

Heads up on an unrelated result: `tests/vale_cup_client.test.ts` has 4 failures (`ReferenceError: WebSocket is not defined` at `online.ts:1689`). I confirmed these **reproduce identically on the parent commit**, so they are pre-existing on Node 20 (no global `WebSocket`) and not from this PR.

## Findings

### 1. Staging a second limit raise silently reverts an already-effective cap (medium-high, confirmed with a runtime repro)

`effectiveCap` (`server/rg.ts:105-121`) materializes a due pending raise as a read-time overlay only; nothing ever collapses it into `dailyCap`/`weeklyCap`/`monthlyCap`. The raise branch of `setLimit` (`rg.ts:201-207`) then overwrites `pendingRaises[window]` while deliberately leaving the window field "at its tighter current value", but after a materialization that stored field is no longer the current value, it is the pre-raise one.

Traced and executed against the real service with an in-memory `RgDb` (config default daily 100,000,000, cooldown 1 day):

1. `setLimit(daily, 1000)`: instant lower. Stored `dailyCap=1000`.
2. `setLimit(daily, 5000)`: raise staged, `dailyCap` still 1000.
3. After the cooldown, `effectiveCap` is 5000 and `limitVerdict(1, 5000)` returns ok. Correct.
4. `setLimit(daily, 9000)`: `currentCap` reads 5000 via the overlay, so it stages a new pending. Observed immediately after:

```
immediately after staging 9000, is 5000 still allowed? false {"ok":false,"window":"daily","cap":1000}
stored limits: {"dailyCap":1000,...,"pendingRaises":{"daily":{"cap":9000,...}}}
```

The player's in-force 5000 drops back to 1000 for a full cooldown purely as a side effect of *requesting* a further raise, and waiting does not restore it. Worth fixing because the PR body specifically claims this area is handled ("held raises are tracked per window so staging one never discards another") and the per-window isolation is correct; it is the materialization that is not durable.

To be fair on severity: the error is always in the tighter direction (`capMagnitude(0)` is `Infinity`, so an unlimited window can never satisfy `raising`), so this clamps players down and is not a compliance hole. Suggested fix: make materialization write-through, baking `currentCap` into the window field before staging a new pending.

**Test gap:** `tests/rg_service.test.ts:212-253` covers instant lower, a single held raise, the 0-as-unlimited raise, and per-window isolation, but every case stops at the *first* materialization. No test stages a second `setLimit` after a raise took effect. A regression test alongside the existing raise case would pin this.

### 2. `status()` reports stored caps, not effective ones (medium)

Same root cause. `rg.ts:241-261` returns `limits` straight from `db.getLimits`, so at step 3 above the payload says `dailyCap: 1000` plus an opaque `pendingRaises` blob while `limitVerdict` is enforcing 5000. Any consumer would have to reimplement `effectiveCap` client-side. For a responsible-gambling disclosure surface, showing a cap that differs from the enforced one is a defect in its own right. A write-through fix for finding 1 resolves this too.

### 3. Age attestation is freely overwritable, so the audit trail is not one (medium)

`attestAge` (`rg.ts:145-152`) comments that it records the attestation "regardless (an audit trail)", but `setAgeAttestation` (`rg_db.ts:122-129`) is `ON CONFLICT (account_id) DO UPDATE SET dob = EXCLUDED.dob`. A caller rejected with `age_underage` can immediately re-call with an of-age DOB and the underage row is destroyed rather than retained, with no re-attestation lock or rate limit. Since this is self-attestation the bar is inherently low, but if the audit trail is a compliance requirement it needs to be append-only, and re-attestation after an underage verdict probably should not be silently permitted. Worth settling before the setter endpoints land in the next tranche.

### 4. Cluster resolution still fails open when unset (medium, design)

Replacing the mainnet hostname regex with an explicit `SOLANA_CLUSTER` is a genuine improvement and I want to credit it. But the fallback in `resolveCluster` (`regulated_gate.ts:42-53`) still returns `localnet` for an unrecognized RPC URL, which feeds `gateVerdict`'s `if (!args.isMainnet) return { ok: true }` and `mainnetMoneyEnabled`'s `if (!isMainnetCluster(env)) return true`. So a mainnet deploy on a custom RPC whose hostname lacks "mainnet" and without `SOLANA_CLUSTER` set skips geo, OFAC, counsel, and self-exclusion on real money. The header comment and `.env.example` acknowledge this and put the burden on the operator, but a *missing* config value producing "gate off" is the opposite of the fail-closed posture stated elsewhere in the module. Consider treating an unresolvable cluster as mainnet, or refusing boot when any casino money flag is on and `SOLANA_CLUSTER` is unset.

### 5. `sdnWallets` does sync file IO per request, with an error-poisoned cache and no refresh (medium)

`regulated_gate.ts:122-135`: `cachedSdn` is assigned only on success, so a missing or unreadable `OFAC_SDN_WALLETS_FILE` makes `readFileSync` throw out of `screenMoneyRequest` on *every* request (it is evaluated eagerly as a `gateVerdict` argument, before the cluster check), and a persistent failure means a blocking sync read on the event loop each time. Once cached successfully the list is never refreshed short of a restart, which is a real sanctions-staleness problem on a long-running realm. Loading once at boot, failing loudly if unreadable, with a TTL or explicit reload, covers all three.

### 6. Self-exclusion is bypassed on non-mainnet (low-medium)

`gateVerdict` checks `selfExcluded` only after the `isMainnet` early return, so a self-excluded player is not blocked on devnet/localnet. Self-exclusion is a player-harm control rather than a money control, so it reads better checked before the cluster branch. Mitigated in practice because `entryVerdict` checks the exclusion unconditionally, but the gate's own ordering is worth correcting as defense in depth.

### 7. Minor

- `realityCheckMs` is defined in `RgConfig`, parsed from `RIVERBOAT_REALITY_CHECK_MS`, and documented in `.env.example`, but has no consumer anywhere in the tree. Either wire it or hold the env var until the tranche that uses it, so operators are not handed a knob that does nothing.
- `setLimit` has no input validation, and `capMagnitude` treats any negative cap as unlimited. Whichever `RouteDef` eventually exposes this must reject negatives at the schema layer, or `setLimit(daily, -1)` becomes an uncap request.
- DOB is stored raw with no retention policy beyond the (correctly present) `ON DELETE CASCADE`. For an 18+ gate the durable fact needed is "of age at time T by method M"; consider storing the derived boolean unless a future KYC vendor swap genuinely needs the date.
- Process note: this targets `release/v0.27.0`, but `release/v0.28.0` exists on origin and CLAUDE.md asks for the latest release branch as the base. Since #2029 is the parent, rebasing the stack is a coordinated move rather than something to do unilaterally, so flagging rather than asking.

## Verdict

Nothing here is a live production bug: `fairness.ts`, `regulated_gate.ts`, and `rg.ts` have no route callers yet, and with the flags unset a realm is unchanged, which matches the dark-ship intent. I am requesting changes rather than approving mainly for finding 1 plus its test gap, since it is a confirmed logic error in exactly the area the PR body calls out as handled, and it is much cheaper to fix now than after the setter endpoints and client interstitial land on top. Findings 2 and 3 are worth resolving in the same pass; 4 through 7 are fine as follow-ups if you would rather keep this tranche tight.

Issue #2025, opened 2026-07-16
perf(mobile): stop sustained phone overheating with frame pacing, mobile-aware tier costs, and proactive thermal shedding


Humpalumps:
## Problem

Players report that phones become extremely hot within roughly five minutes of normal mobile play, in mobile browsers and in the Capacitor app shells alike. The timing is consistent with sustained CPU/GPU saturation and thermal soak rather than a one-time loading hitch.

This issue was rewritten on 2026-07-16 after a code-verified investigation. Every claim below was checked against the code (release/v0.27.0 head); the original report's six suspicions were all confirmed, with corrections noted at the end of this section.

The sustained load differs by cohort:

- Android flagship phones: recognized GPU strings (Adreno 6xx to 8xx, Mali-G7xx, Immortalis, Xclipse, Apple A14+ where visible) classify as `flagshipMobile` in `classifyGpuRenderer()` and auto-resolve to High. High runs the full N8AO + bloom + grade composer (several full-resolution passes over HalfFloat targets per frame; N8AO's depth-sampled scene target is the most tile-GPU-hostile pattern in the codebase), a 4096 PCF-soft shadow map re-rendered every frame (`shadowMap.autoUpdate` is never disabled anywhere in `src/`), 4x MSAA, and a WebGL context created with `powerPreference: 'high-performance'`. The rAF loop (`frame()` in `src/main.ts`) reschedules unconditionally with no cadence control, and Android Chrome/WebView runs rAF at panel rate, so a 90/120 Hz phone pays all of this 90/120 times per second.
- Android mid and unknown phones: Medium. No composer, but PBR materials, terrain splatting, and a 2560 PCF-soft shadow map redrawn every frame, still at panel-rate rAF.
- iPhones: WebKit masks the WebGL renderer string to "Apple GPU", so every iPhone classifies as unknown and lands on Medium. iOS Safari and WKWebView cap rAF near 60 Hz, so high-refresh frame multiplication is an Android-only mechanism; iPhones overheat from sustained full-cost 60 fps frames instead.

Cross-cutting, all confirmed:

- `GFX_BUDGETS.*.targetFps` (60 on every tier) is telemetry-only: its sole consumer is the perf-reporter payload. Nothing anywhere enforces a frame cap.
[...]

PR #2035, opened 2026-07-16
fix(render): pace native mobile frames safely

Humpalumps:
## Summary

This PR implements the first two priority workstreams from #2025 against `release/v0.29.0`:

1. Add explicit native mobile frame pacing so high-refresh displays do not run the whole game loop at panel callback rate.
2. Separate intentional pacing from measured accepted-frame work so adaptive render quality responds to real load.

The branch is synchronized with the latest fetched release tip. Automatic pacing is scoped to the Capacitor native runtime. Desktop, Electron, and desktop Touch interface mode remain unpaced.

## Review fixes included

- Scale render-budget drop, urgent, submit, and recovery thresholds to the active paced target. A sustainable 45 fps frame at 18 ms can now recover quality instead of remaining degraded against 60 fps thresholds.
- Use a 1.45 engagement ratio and 1.40 disengagement ratio around the high-refresh divisor boundary. Fresh 89.9 Hz and 90 Hz samples select about 45 fps consistently, while hysteresis prevents adaptive-refresh noise from flapping the decision.
- Arm the five-second loading watchdog before native refresh calibration. If animation callbacks stop, its emergency path removes the loading overlay directly.
- Keep intro timing, readiness, performance startup, and game exposure on the completed-frame-plus-paint path. The emergency watchdog does not advance those clocks while rendering is suspended.
- Keep the watchdog effective from the first completed gameplay frame until queued paint actually completes.
- Remove per-frame body-class reads and scope automatic pacing to native runtime capability rather than interface mode.

### Frame pacing

- Gates before input, simulation, rendering, HUD work, and accepted-frame timing advance.
- Calibrates native panel cadence under the loading screen.
- Selects the highest whole-panel divisor below the graphics ceiling and carries timing remainder to avoid cadence drift.
- Revalidates adaptive-refresh changes with interleaved probes and workload-headroom confirmation.
- Covers 30, 60, 72, 75, 89.9, 90, 120, 144, 165, 240, and 360 Hz behavior.

### Render budget

- Forwards pacing state, paced target, and completed accepted-frame work through `Renderer` to `RenderBudgetGovernor`.
- Uses measured work while pacing is intentional instead of deliberate callback spacing.
- Scales governor thresholds from the 60 fps baseline to the executed-frame budget.
- Keeps genuine renderer, submit, draw, grass, and non-render main-thread pressure visible.

### Loading handoff

- Waits for a completed gameplay frame and the following paint callback on the normal path.
- Arms its timeout before native calibration can stall.
- Supports a separate emergency overlay fallback without consuming the later normal handoff.
- Remains idempotent when the watchdog fires between frame completion and queued paint.

## Scope

The final feature comparison against `release/v0.29.0` contains exactly eleven files:

- `docs/design/mobile-frame-pacing.md`
- `src/game/frame_pacer.ts`
- `src/game/loading_handoff.ts`
- `src/main.ts`
- `src/render/render_budget.ts`
- `src/render/renderer.ts`
- `tests/architecture.test.ts`
- `tests/frame_pacer.test.ts`
- `tests/frame_pacing_integration.test.ts`
- `tests/loading_handoff.test.ts`
- `tests/render_budget_pacing.test.ts`

Part of #2025. This intentionally completes only priority tasks 1 and 2.

## Release synchronization

- PR base: `release/v0.29.0`
- Merged v0.29 release tip: `55af0ae9be0190664b3f5d9f84e6312275dd2fe3`
- Review-fix commit: `7d85aae5`
- Conflict-resolution merge and final pushed head: `4aa6e6d1e1ee70a32e9325cfcaaf0212e8ab9b3d`
- The merge preserves v0.29 entry diagnostics and quality-glow architecture changes while keeping the frame-pacing and loading-handoff contracts.
- GitHub reports the PR mergeable, and the final comparison against v0.29 remains limited to the intended eleven files.

## Validation

Post-`release/v0.29.0` merge regression suite:

`npx vitest run tests/loading_handoff.test.ts tests/frame_pacer.test.ts tests/frame_pacing_integration.test.ts tests/render_budget_pacing.test.ts tests/gfx.test.ts tests/architecture.test.ts`

- 6 files passed, 107 tests passed.
- `npm run check:ts` passed.
- Biome formatted the three conflict-resolution files successfully.
- `git diff --cached --check` passed before the merge commit.

Pre-v0.29 synchronization focused regression suite:

`npx vitest run tests/loading_handoff.test.ts tests/frame_pacing_integration.test.ts tests/frame_pacer.test.ts tests/render_budget_pacing.test.ts tests/render_budget.test.ts --reporter=dot`

- 5 files passed, 69 tests passed.

Pre-v0.29 synchronization adjacent client and renderer suite:

`npx vitest run tests/client_shell.test.ts tests/gfx.test.ts tests/mobile_controls.test.ts tests/painter_host.test.ts tests/hud_perf_budget.test.ts tests/frame_pacer.test.ts tests/loading_handoff.test.ts tests/frame_pacing_integration.test.ts tests/render_budget_pacing.test.ts tests/render_budget.test.ts tests/architecture.test.ts --reporter=dot`

- 11 files passed, 314 tests passed, 3 tests skipped.

Pre-v0.29 synchronization static and build checks:

- `npm run check:types` passed. Svelte diagnostics reported 0 errors and 0 warnings.
- `npm run build:env`, `npm run build:server`, and `npm run build` passed.
- Biome reported no errors or format drift in the changed TypeScript files. It reports 19 existing warnings in `main.ts` and `renderer.ts`.
- `git diff --check` passed. Generated validation artifacts are excluded from the commits.

Earlier canonical repository gate:

- i18n freshness, malware scan, changed-file, and SFX stages passed.
- The full Vitest stage is not green locally: 16,441 tests passed, 111 failed, and 152 were pending across 29 failing test files.
- Every failing test file is outside this PR's eleven-file feature comparison. The failure families include asset and SFX tooling, deploy and release contracts, bank/options/store windows, server endpoints and HTTP characterization, mail, threat, raid, and environment-specific parsing.
- Because the gate stops at Vitest, type and production builds were run independently and passed as listed above.

## Manual verification status

- Deterministic tests cover targeted refresh, workload, recovery, suspension, startup, watchdog, and runtime-scope paths.
- A physical mobile-device session was not available in this environment.
- No player-facing strings, assets, CSS layout, routes, or endpoints changed.

## Checklist

- [x] Scoped to tasks 1 and 2 from #2025.
- [x] Synchronized with the latest fetched `release/v0.29.0` tip.
- [x] Review issues fixed with regression coverage.
- [x] Final feature comparison audited.
- [x] Focused and adjacent suites pass.
- [x] Client and admin type checks pass.
- [x] Client, environment, and server builds pass.
- [x] Changed-file formatting and whitespace checks pass.
- [x] No generated validation artifacts or secrets included.
- [ ] Full local `npm run gate` passes. See the documented failure families above.
- [ ] Verified on physical desktop and mobile devices.


jgyy (commented):
This lines up well with #2070's conclusion that the render budget governor is a control-loop worth extending rather than a WebGPU rewrite, and it tackles the two priority items from #2025 cleanly.

The core idea in `render_budget.ts` is the right fix for the false-positive problem: `frameMs` now switches to the measured `workMs` (input + sim + render + HUD) only while `intentionalFramePacing` is true, so a deliberate 45 fps cadence stops being misread as either an external cap or GPU slowness, while real render/submit/draw/grass pressure still degrades quality normally. `externalFrameCap` is explicitly gated off (`!intentionalFramePacing && ...`), so the two signals can't fight each other.

`FramePacer`'s calibration/probe state machine (`src/game/frame_pacer.ts`) correctly separates "a higher callback rate must be revalidated immediately" from "a lower rate is ambiguous between a real cap and missed callbacks," and the workload-headroom promotion path only fires after `WORKLOAD_HEADROOM_CONFIRMATION_SAMPLES` (8) consecutive cheap frames, so a transient hitch can't misclassify a workload-limited cadence as a real display cap.

The loading-screen handoff change (`src/main.ts`) is a real behavior change worth flagging even though I didn't find a bug in it: previously the loading screen faded after a fixed two `requestAnimationFrame` calls regardless of frame content; now it waits on `firstRenderedFrame`, which only resolves once a full offline or online frame runs to completion past `hud.update()`. I traced every path from the top of `frame()` to `markFirstRenderedFrame()` in both the offline and online branches and found no early return between the pacing gate and the resolve call, and `FramePacer.step` always eventually returns `shouldRun: true` (worst case, unpaced), so this doesn't introduce a hang. Just worth having a second reviewer's eyes on this specific change since a stuck loading screen is a bad failure mode if a future edit adds an early return in that span.

Ran the new suite plus `tests/architecture.test.ts` against the final head (`50b18df8`): 4 files, 69 tests, all green, including the DOM/Three purity and pure-core registration guards. Merge-tree against the current `release/v0.27.0` tip (`fbfa6bafd`) is clean, no conflicts.

Nits:
- `docs/design/mobile-frame-pacing.md` line noting `GFX.budget.targetFps` as "the single source of the configured maximum" is worth double-checking against `GFX_BUDGETS` in `gfx.ts`: `targetFps` is 60 for every tier today, so the pacer's ceiling is effectively fixed at 60 fps regardless of graphics tier. Not a bug, just means the "graphics budgets remain the single source" framing is more theoretical than active until a tier ever sets a different `targetFps`.
- The PR body is upfront that the full local `npm run gate` and browser suite weren't green end to end, with the failures listed and traced to unrelated baseline/environment families and a pre-existing 2.6px Armory overflow. That's the right way to report it, and matches what I'd expect from a diff scoped to 8 files with no CSS/Armory changes.

Good scoping, honest reporting of gate/browser status, and the governor change is the correct fix for the underlying false-positive problem rather than a workaround.

EnriqueGF (changes requested):
Reviewed the full eight-file diff, read the root/`src/`/`src/game/` CLAUDE.md rules, and re-ran the checks locally rather than relying on the PR body. Nice, careful piece of work: the pacer is a genuinely pure injected-clock core, the governor split between cadence and measured work is coherent, and the test suites are real state-machine coverage rather than smoke tests. Three things I would like resolved before merge, one of them a real gameplay-feel regression.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/frame_pacer.test.ts`, `tests/frame_pacing_integration.test.ts`, `tests/render_budget_pacing.test.ts`, `tests/render_budget.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`: 332 passed.
- `tests/localization_fixes.test.ts`: passes after `npm run i18n:gen` (the failure without it is a missing generated artifact, not this PR).
- I also probed `RenderBudgetGovernor` directly with a resume-hitch scenario (paced vs unpaced, same wall-clock gap) because I suspected the new instantaneous `frameMs` term in `rawFramePressure` made paced mode more trigger-happy. It does not: both arms stayed `stable` with identical levels, and steady paced cheap frames degrade less, not more. Dropping that concern, the governor change looks sound.

## Findings

### 1. A 72 Hz or 75 Hz panel gets paced to 36 or 37.5 fps (major, gameplay feel)

`pacedFrameRateFor` picks the highest whole divisor at or below the ceiling. I ran it across the real range:

```
measured 72 Hz  -> target 36.00 fps  paced=true
measured 75 Hz  -> target 37.50 fps  paced=true
measured 90 Hz  -> target 45.00 fps  paced=true
```

72 Hz is extremely common on Android (Pixel, OnePlus, many mid-range devices). A phone that today runs the loop at 72 fps would, after this change, run it at 36. That is not shedding surplus callbacks, it is halving perceived smoothness on a device that was comfortably above the ceiling. 90 to 45 is defensible; 72 to 36 crosses into visibly choppy.

Note that 72 Hz and 75 Hz appear in neither the table in `docs/design/mobile-frame-pacing.md` (which jumps 60, 90, 120, 144, 165, 240) nor `tests/frame_pacer.test.ts`. The two most consequential mid-range cases are the two that are undocumented and unpinned, which is probably why this was not caught.

Suggestion: only pace when the panel rate is a clear multiple above the ceiling (for example require `refreshFps >= maxFps * 1.5` before engaging), so 72 and 75 Hz simply run uncapped. Then add both to the doc table and to the divisor test.

### 2. Divisor cliff at 61.8 Hz halves a 60 Hz panel (major, correctness)

Same function, the boundary:

```
measured 61.8 Hz -> target 61.80 fps  paced=false
measured 61.9 Hz -> target 30.95 fps  paced=true
```

`FRAME_RATE_TOLERANCE` of 1.03 gives only about 1.8 fps of slack before the divisor jumps to 2. A plain 60 Hz mobile panel whose median rAF interval measures 16.15 ms instead of 16.67 ms, a 3.1 percent overshoot, drops the whole game loop to 31 fps. The median over 8 to 30 samples makes this unlikely but not impossible on Android browsers with jittery rAF timestamps, and the failure is silent, sustained, and severe.

The fix in finding 1 (a 1.5x engagement margin) closes this cliff too, since a measured 61.9 Hz would no longer engage pacing at all. Worth a boundary test pinning that a noisy 60 Hz measurement never engages pacing.

### 3. No watchdog on the loading-screen handoff (should fix)

`src/main.ts:3205`. The cut to game used to be an unconditional double `requestAnimationFrame`. It now waits on `firstRenderedFrame`, which only resolves at the tail of `frame()` via `markFirstRenderedFrame()`. I read the whole `frame()` body: there is no early `return` between the pacing gate and the two mark sites, and I convinced myself the pacer cannot starve `shouldRun` indefinitely (`remainderMs` accumulates every callback, and `stepCalibration` alternates), so the normal paths are fine.

But any throw inside `frame()` (renderer.sync, hud.update, the online snapshot path) now leaves the player on a permanent loading screen instead of a degraded but visible game, and it hides the error. `requestAnimationFrame(frame)` being first keeps the loop alive, so it throws forever with nothing on screen.

Suggestion: `Promise.race` `firstRenderedFrame` against a bounded timeout of a few seconds so the handoff is failure-tolerant, plus a test that drives a throwing frame callback.

## Smaller points

- `src/game/frame_pacer.ts` is a pure, DOM-free, injected-clock decision core but is registered in neither `UI_PURE_CORES` nor `RENDER_PURE_CORES` in `tests/architecture.test.ts`, so no purity guard covers it. The precedent for a `src/game/` leaf is already there (`ui_effects_profile.ts`, `ui_tier_knobs.ts` at lines 210 and 211). I checked the source against the guard's patterns and it should pass as-is. Worth registering so a later edit cannot slip a `performance.now()` or a DOM read into it.
- `framePacer.setEnabled(document.body.classList.contains('mobile-touch'))` runs on every animation callback, including paced-out ones. `classList.contains` does not force reflow so this is not a thrash bug, but the class only changes at two toggle sites; pushing the value on change would fit the per-frame-budget philosophy better.
- A large share of the `render_budget.ts` diff is pure Biome reformatting interleaved with the behavioral change, which makes the real delta hard to isolate. I did verify the behavioral part in isolation. Ideally that is a separate commit.
- `Renderer.sync()` is now seven positional parameters, with the offline call site passing `0, null` placeholders to reach the new trailing args. Optional, but an options object would age better.

## Base branch

This targets `release/v0.27.0`, which has already been merged (`b72b893a0`); `release/v0.28.0` is now the active integration base. Good news: v0.28.0 is exactly one commit ahead and touches none of the four source files here, so retargeting is trivial and conflict-free. Worth doing so CI runs against the current base.

Happy to re-review quickly once the pacing engagement margin is in. The rest of this is in good shape.

EnriqueGF (commented):
Follow-up to my review above: I ran a mutation pass over the new suites to measure assertion decisiveness rather than infer it from reading. The coverage here is genuinely good in most dimensions (the `toEqual(new Set([expectedGap]))` spacing pins and the exact-levels assertions are strong, and there are no mocks, only real modules driven by deterministic timestamp generators). Three constants and one test title did not survive, though, and I verified each of these myself by mutating the source in my worktree and re-running.

**1. `tests/render_budget_pacing.test.ts:183`, "falls back to render total when paced full-frame work is unavailable", does not pin the fallback.**

I changed `sample.workMs ?? totalMs` to `sample.workMs ?? 0` in `src/render/render_budget.ts:252` and both governor suites stayed green (19 passed). The sample uses `totalMs: 38`, which independently trips `totalMs >= budget.urgentFrameMs` (34) and `totalMs >= budget.dropFrameMs` (22), so it degrades no matter what `frameMs` becomes. The test passes for a reason unrelated to its title.

Closing it needs a sample where `totalMs` alone is under budget so the fallback value is the only thing that can move pressure (something like `totalMs: 24` with cheap submit/calls/triangles), asserting `reason === 'frame'`. If no such sample exists because every `frameMs` arm is shadowed by a parallel `totalMs` arm, then the fallback is behaviorally inert and the honest fix is to retitle the test to what it actually pins.

**2. `WORKLOAD_HEADROOM_RATIO` is unpinned.** I changed it from `0.75` to `10`, which removes the target-interval term from `Math.min(targetFrameIntervalMs * WORKLOAD_HEADROOM_RATIO, trustedPanelIntervalMs)` entirely, and all 33 tests passed. Every promotion test uses 120 Hz trusted with a 60 fps target and `frameWorkMs: 8.3`, where the panel term (8.33 ms) is always the binding minimum, so only one of the two dimensions is ever exercised. One case where the target interval binds (trusted 30 Hz panel, limit `16.7 * 0.75 = 12.5 ms`) asserting `frameWorkMs: 20` does not promote while `10` does would close it.

**3. `REFRESH_CHANGE_CONFIRMATION_SAMPLES` is unpinned in the weakening direction.** Changed `6` to `2`, all 33 tests passed. That is the debounce that stops the pacer recalibrating on transient callback noise, which is exactly the hardening the PR body describes, so a change making it far twitchier would ship silently. A test feeding fewer than the confirmation count of incompatible samples and asserting `estimatedRefreshFps` is unchanged pins the floor rather than only the ceiling.

**Smaller, same pass:** deleting `this.workloadLimitedCallbackFps = 0;` at `frame_pacer.ts:311` also leaves the suite green, so nothing asserts that a promoted rate can itself be revalidated later. And `FRAME_PACER_CALIBRATION_CALLBACKS` is imported by both `main.ts` and the test as a loop bound, so both sides move together (I changed the slack from `4` to `40` with no failure); a one-line `expect(FRAME_PACER_CALIBRATION_CALLBACKS).toBe(13)` is the usual mitigation, and the value is load-bearing since it sets how long the loading screen is held on mobile.

This also independently reinforces finding 3 in my review: the whole `main.ts` change, including the loading handoff, is pinned only by regex matches against raw source text, so nothing ever executes the handoff. That kind of pin passes a change that is textually identical but semantically broken and fails on a harmless reformat. Extracting the handoff into a small sibling module (per the `main.ts`-as-firewall rule) would make it testable in plain Node with a fake rAF, and would let the watchdog I suggested be tested properly.

Humpalumps:
Addressed the review and mutation-pass feedback in `37b7462b`, and retargeted this PR to the active `release/v0.28.0` integration branch.  What changed:  - Added a 1.5x pacing-engagement margin. Noisy 60 Hz measurements and mid-refresh 61.9, 72, 75, and 89.9 Hz sources remain unpaced; 90 Hz still selects 45 fps. The policy and 72/75 Hz outcomes are documented. - Extracted the loading handoff from `main.ts` into `src/game/loading_handoff.ts`. It waits for a completed gameplay frame and the following paint, with a pinned five-second watchdog if frame work throws. Deterministic fake-rAF tests execute both the normal and throwing-frame paths. - Registered both `frame_pacer.ts` and `loading_handoff.ts` with the pure-core architecture guards. - Confirmed the render-budget fallback distinction is behaviorally inert because the parallel total-frame pressure arm shadows it. The misleading fallback title was replaced with an honest render-total pressure test using cheap parallel pressure inputs and `reason === 'frame'`. - Added a target-interval-binding headroom case: expensive 20 ms work does not promote the learned cap, while cheap 10 ms work does. This kills `WORKLOAD_HEADROOM_RATIO: 0.75 -> 10`. - Added five incompatible callback samples without revalidation. This pins the debounce floor and kills both `REFRESH_CHANGE_CONFIRMATION_SAMPLES: 6 -> 2` and `6 -> 5`. - Added promoted-rate revalidation from 90 Hz to 80 Hz, which kills removal of the `workloadLimitedCallbackFps` reset. - Pinned `FRAME_PACER_CALIBRATION_CALLBACKS` to exactly 13, killing the calibration-slack `4 -> 40` mutation. - Pinned the watchdog to exactly 5,000 ms, killing a `5,000 -> 500,000` weakening.  Verification on the pushed head:  - Focused suites: 46 tests passed. - Adjacent renderer/client suites: 196 tests passed. - TypeScript, Svelte type checks, build, changed-file Biome, and `git diff --check` passed. - The reviewed weakening mutations all fail for their intended assertions.  The v0.28 comparison remains exactly the intended 11 files even though the active base has advanced substantially since this feature branch was created.

Ready for re-review, @EnriqueGF.

EnriqueGF (changes requested):
Reviewed the full 11-file diff against `release/v0.28.0` and ran the checks myself in a clean worktree on `37b7462b1`.

**What I ran**

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/frame_pacer.test.ts tests/loading_handoff.test.ts tests/frame_pacing_integration.test.ts tests/render_budget_pacing.test.ts tests/render_budget.test.ts tests/architecture.test.ts`: all pass.
- `tests/localization_fixes.test.ts` (S3 guard) after `npm run i18n:gen`: 29 passed, 3 skipped.
- `tests/client_shell.test.ts tests/gfx.test.ts tests/mobile_controls.test.ts`: 172 passed.
- `git diff --stat origin/release/v0.28.0...HEAD`: confirmed exactly the eleven files claimed.
- Invariant sweep: no `src/sim/` files touched, so determinism and the `IWorld` seam are not in play. No player-visible strings added, so no i18n surface. No em dashes, en dashes, or emojis in any new file. `NATIVE_APP` is set only by `build:native` (Capacitor), not by the Electron path, so the Steam desktop shell is genuinely unaffected. The two new modules are correctly registered in `UI_PURE_CORES` and `BARE_NAMED`, and both are DOM-free.

The pacer state machine reads well and the test coverage is unusually thorough for a timing change. Two things I want to raise before this merges, one of which I confirmed by running the real governor.

---

**1. The governor's frame budget is still calibrated to 60 fps while the pacer targets 45, 48, or 55 fps (medium)**

`RenderBudgetGovernor` now correctly uses measured work instead of cadence while pacing is intentional, but `dropFrameMs` / `recoverFrameMs` are still the fixed 60 fps numbers (22 / 17.5 on `low`). The pacer knows its paced target; the governor is never told it. So on a 90 Hz panel paced to 45 fps the real frame budget is 22.2 ms, but recovery still demands work at or below 17.5 ms.

I confirmed this against the real class. After forcing a degrade, then feeding 4000 frames of sustained 18 ms work at `intentionalFramePacing: true, dt: 1/45` (19 percent of headroom left inside the paced budget):

```
after degrade:  {grass:0.62, foliage:0.68, vfx:0.84, lighting:0.78, resolution:0.64}
after 4000 frames at 18ms: stable {grass:0.62, foliage:0.68, vfx:0.84, lighting:0.78, resolution:0.64}  ema=18.0
control at 14ms work:      stable {grass:0.9,  foliage:1,    vfx:1,    lighting:1,    resolution:0.64}
```

Nothing recovers, ever, at a work level the paced target comfortably absorbs. The direction is conservative rather than dangerous, but it undercuts the stated goal of this PR: pacing is separated from pressure on the degrade side and not on the recover side, so a paced device gets stuck holding lower quality than its own cadence justifies. Scaling the thresholds by `pacedTargetFps / budget.targetFps` when `intentionalFramePacing` is set would close it, and it is worth a pinned test either way.

**2. A 2x cliff sits exactly on the most common high-refresh Android panel rate (medium, design question)**

`PACING_ENGAGEMENT_RATIO = 1.5` against a 60 fps ceiling puts the engagement boundary at exactly 90 Hz. Enumerating `pacedFrameRateFor(r, 60)`:

```
 89 Hz -> 89.00 fps (divisor 1)
 90 Hz -> 45.00 fps (divisor 2)
```

`tests/frame_pacer.test.ts:105-106` pins both sides of this. The 1.5x guard is the right idea, and the docs explain it as protecting 72 to 75 Hz panels from the divisor cliff, but the threshold was placed on top of 90 Hz rather than clear of it. Consequences:

- A 90 Hz phone that calibrates to 89.9 Hz (median of 8 rAF intervals, and plenty of panels report non-integer rates) never paces at all, so the thermal goal this PR exists for is silently not met. Calibrating to 90.1 halves it to 45 fps. Same device, two outcomes, 2x apart.
- When it does engage, 90 Hz users go from roughly 60 to 90 fps today to a hard 45. 144 Hz gives 48 and 165 Hz gives 55, all meaningfully under the 60 fps ceiling that is supposed to be the cap.

I do not think this blocks the concept, and choosing 45 over a juddering 60 on a 90 Hz panel is defensible. But it is the one behavior here that tests cannot catch and that users will feel immediately, on the single most common high-refresh Android panel. I would like an explicit call on it, plus some hysteresis around the engagement threshold so a device cannot oscillate across a 2x boundary on calibration noise alone.

---

**Smaller notes**

- Roughly 85 percent of the `src/render/render_budget.ts` diff is Biome reformatting of untouched code. That is exactly what `biome check --write` on a changed file does and is per CONTRIBUTING, so it is correct, just worth knowing the real change is about 30 lines.
- The loading handoff is a real improvement over the old unconditional double-rAF, and the 5 s watchdog covers the throw path. I traced every `return` inside `frame()`: only the pacing gate and the offline branch, and the offline branch does set `previousFrameWorkMs` and mark the handoff, so there is no stale-work or stuck-loading path.
- No physical device session, which the PR is upfront about. Given that the entire user-visible effect of this change is on real mobile panels, one 90 Hz and one 120 Hz device check before release would be worth more than any additional simulated coverage.

Happy to re-review once 1 is addressed and there is a decision on 2.

EnriqueGF:
Follow-up after a second pass over the render and UI seams. Two additions I confirmed, and one candidate concern I tested and am explicitly withdrawing.

**3. The loading watchdog is armed after the calibration loop it needs to protect (medium)**

`src/main.ts:3204` runs the mobile calibration loop, awaiting up to `FRAME_PACER_CALIBRATION_CALLBACKS` (13) rAF callbacks. `loadingHandoff.start()`, which arms the 5 s watchdog, is not reached until line 3217.

If rAF is throttled during boot, which is ordinary on mobile when the user switches apps mid-load, the loop stalls and `start()` is never called, so the watchdog that exists precisely to guarantee the loading screen comes down is never armed. There is no recovery path. The pre-existing `await nextPaint()` at `src/main.ts:842` already has this exposure, so this widens the hole rather than creating it, but it adds 13 more suspension points and it undercuts the specific guarantee this PR added. Arming the watchdog before the calibration loop would close it, and `tests/loading_handoff.test.ts` could pin it with a case where rAF never fires.

**4. Pacing is keyed to interface mode, not to the device (medium)**

`src/main.ts:2509` and `:2776` gate the pacer on `document.body.classList.contains('mobile-touch')`, but that class is toggled at runtime from `NATIVE_APP || useTouchInterface()` (`src/main.ts:548`, `src/game/mobile_controls.ts:527`). It tracks the selected input interface, not the hardware.

So a desktop or 2-in-1 user who selects the Touch interface on a 144 Hz monitor is capped to 48 fps, with no setting to opt out. A frame-rate ceiling reads like it belongs on the graphics preset (`src/game/ui_effects_profile.ts`) or its own setting, rather than falling out of an input-mode class. Related: toggling interface mode mid-session with no trusted refresh calls `beginCalibration()`, and `stepCalibration` alternates `shouldRun: false` for roughly 16 callbacks, so the user sees a brief 50 percent frame drop at the moment they flip the toggle.

**Withdrawn: single-hitch sensitivity under pacing**

I looked hard at the new instantaneous term `intentionalFramePacing ? positiveRatio(frameMs, dropFrameMs) : 0` at `src/render/render_budget.ts:295-298`, on the theory that a paced device would degrade on a single GC or shader-compile hitch where an unpaced one would not. It does not reproduce. Running the real governor with 300 healthy 8 ms frames (past the 1.1 s cooldown) followed by one 40 ms `workMs` spike:

```
UNPACED  after = {grass:0.9, foliage:0.9,  vfx:1, lighting:1, resolution:1}  mode=stable reason=frame-cap
PACED    after = {grass:0.9, foliage:0.98, vfx:1, lighting:1, resolution:1}  mode=stable reason=stable
```

Neither degrades, and the paced arm is actually the better-behaved one. Recording it so nobody re-litigates it.

**Minor**

- `src/game/loading_handoff.ts:55`: `if (handoffQueued) clearWatchdog();` inside `start()` is unreachable. `queueHandoff` early-returns unless `started`, and `started` is set synchronously two lines earlier, so `handoffQueued` cannot be true at that point. Harmless, but it implies a race that does not exist.
- `src/main.ts:2776` reads `classList.contains` on every rAF callback, including decimated ones, for a value that changes only on interface-mode toggle. It is a cheap non-reflow read and not a `hud_perf_budget` violation, but a cached flag updated from the existing toggle site would be free.

**Fairness invariant: no violation found.** Decimation drops whole rAF callbacks, but `last` and `frameDt` advance only on executed frames, so the fixed-timestep accumulator, the 20 Hz intent stream, and every sim step are preserved exactly. Nothing actionable is hidden or culled by a tier check. The closest call is 90 Hz to 45 fps, where a ground-AoE telegraph is sampled every 22 ms instead of 11 ms. I read that as redraw smoothness rather than a withheld signal, so allowed, but it is the one place this change approaches the line in `docs/design/graphics-settings-fairness.md` and is worth an explicit sign-off alongside the 90 Hz decision in point 2 of my first review.

Gates re-confirmed on this pass: `tests/architecture.test.ts`, `tests/painter_host.test.ts`, `tests/hud_perf_budget.test.ts`, and the four pacing suites all pass. No CSS, no painter, and no player-visible string in the diff, so the styles and i18n seams are not in play.

jgyy (commented):
This is a clean, well-scoped piece of #2025's mobile thermal work. The pacer/governor coupling is the hard part of that issue (the "pacer-governor trap" the issue calls out), and this PR gets the seam right: `FramePacer` is a pure core (registered in `UI_PURE_CORES`/`BARE_NAMED`), `main.ts` stays a thin consumer that gates before input/sim/render/HUD work, and the fixed 20 Hz sim accumulator is untouched (`last` only advances on executed frames, so a paced-out callback contributes nothing and the next executed frame just gets the full elapsed delta). Scoping to `NATIVE_APP` rather than interface mode is the right call per the doc's own reasoning (Touch on a 2-in-1 desktop shouldn't get hardware pacing).

Ran the PR's own suite plus `architecture.test.ts` locally against the head commit: 5 files, 82 tests, all green.

Nothing blocking. Two should-fix/nit items:

- **should-fix**, `src/main.ts` (PR description, "Full local `npm run gate` passes" left unchecked): the PR reports 111 failing tests across 29 files in the full run and asserts they're all pre-existing/out-of-scope. That's plausible given the 11-file footprint, but it's an unverified claim in the PR body rather than something CI has confirmed against this head. Worth confirming against a fresh `release/v0.28.0` baseline (or just letting CI settle) before merge rather than taking the count on faith.
- **nit**, `src/game/frame_pacer.ts:401-436` (`maybePromoteWorkloadLimitedCadence`): the workload-headroom promotion path is the most intricate piece of state machine here (learned cadence, confirmation samples, headroom ratio). It's well covered by the parametrized tests, but there's no inline comment at the method boundary explaining what "promote" means at a glance the way the two comments inside the body do. Not blocking, the existing comments carry it.

Good use of the render-budget governor split (`cadenceMs` vs `workMs`) to stop intentional pacing from either faking an external frame cap (force-enabling recovery) or reading as genuine GPU slowness, that's exactly the trap the linked issue described, and the threshold scaling (`pacedBudgetScale`) keeps drop/urgent/recover behavior meaningful at a paced target instead of comparing paced frames against 60 fps baselines.

EnriqueGF (changes requested):
Reviewed this end to end in a clean worktree off `release/v0.28.0`. Nice piece of work overall: the pacer and the loading handoff are both properly extracted, injected-clock pure cores with real behavioral coverage (34 cases in `frame_pacer.test.ts` alone), and the governor change is a genuinely careful separation of intentional pacing from measured work. One thing I do want fixed before merge, plus some smaller notes.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/frame_pacer.test.ts`, `loading_handoff`, `frame_pacing_integration`, `render_budget_pacing`, `render_budget`, `architecture`, `world_api_parity`: 358 passed.
- `tests/parity`: 180 passed, 1 skipped.
- `tests/localization_fixes.test.ts` (S3 guard): passes. It fails on a bare `npx vitest run` only because `i18n.status.json` is not generated yet; green after `npm run i18n:gen`. Not a PR issue.
- Adjacent: `gfx`, `hud_perf_budget`, `client_shell`, `mobile_controls`: 209 passed.
- Biome on the changed files: no errors, no format drift (only the 19 pre-existing warnings in `main.ts` / `renderer.ts`, which CI does not gate on).

## Invariants

Checked and clean. Zero `src/sim/` or `server/` files touched, so determinism and server authority are not in play, and the parity gate confirms no sim drift. No new player-visible strings, so nothing owed to i18n. No dashes or emoji in the new files. Diff scope is exactly the eleven files you list.

Two I verified rather than assumed:

- **Sim timing is untouched.** `last` is only reassigned on accepted frames, so an accepted frame's `frameDt` spans the skipped ones and the fixed 20 Hz accumulator catches up exactly.
- **`NATIVE_APP` really is Capacitor-only.** It comes from `VITE_NATIVE_APP`, set in exactly one place (`package.json` `build:native`). `scripts/electron-build.mjs` builds its env from `VITE_DESKTOP_*` and never sets it. Electron and desktop web stay unpaced as documented.

On graphics fairness: I do not think this breaks the invariant. I swept `pacedFrameRateFor` over 20 to 480 Hz in both hysteresis states, and the lowest target it ever selects while actually pacing is **41.2 fps** (at 123.6 Hz); the engagement ratio rules out the lower-divisor cases. That is comfortably above the 20 Hz authoritative sim tick, and pacing reduces everything uniformly rather than selectively hiding a signal, so no actionable information is delayed. Worth stating that argument in `docs/design/mobile-frame-pacing.md` though, since the doc currently never mentions fairness and the next editor to widen the cap will not know the floor is load-bearing.

## Should fix: the watchdog leaves startup permanently half-finished

This is the one I would like addressed before merge, because it is a regression against base for a real failure mode and the test suite currently pins it.

`gameInputReady = true` is now written in exactly one place (`src/main.ts:3319`), inside the `loadingHandoff.start` callback. That callback is reachable only through `markFirstRenderedFrame()`, which is the **last statement** of `frame()` on both the offline and online arms, not in a `finally`. Meanwhile `requestAnimationFrame(frame)` is re-armed at the top of `frame()`.

So if `frame()` throws (renderer, hud, or sim path), the loop reschedules forever and never reaches `markFirstRenderedFrame()`. After five seconds the watchdog fires, and `fireWatchdog` calls `onWatchdog` (`hideLoadingScreen`), nulls it, and returns leaving `onHandoff` still armed with `handoffQueued === false` and no re-armed timer. Net result: loading overlay gone, world visible, and `input.setSuspendMovement(!gameInputReady || ...)` stuck true forever. The character can never move. Also no perf reporter, no icon prewarm, no first-run camera prompt, and `window.__game` never set (292 files under `scripts/` reference `__game`, so E2E turns into opaque timeouts).

Before this PR the handoff was an unconditional double `requestAnimationFrame` that did not depend on `frame()` completing at all, so it set `gameInputReady` regardless. The coupling is new.

`tests/loading_handoff.test.ts:53` ("uses the bounded watchdog after a gameplay frame throws before completion") asserts `expect(onHandoff).not.toHaveBeenCalled()` after the watchdog fires, so this is currently locked in as intended behavior.

I understand why it is written this way: the PR notes deliberately keep the intro clock and startup clocks off the emergency path so they do not advance while rendering is suspended. That is the right call for a **transient** stall, where frames resume and the normal handoff completes. It is the wrong outcome for a **permanent** one. Cheapest fix is probably running `markFirstRenderedFrame()` in a `finally` around the frame body, or giving the watchdog a bounded second stage that completes the handoff if frames still have not resumed. Either way the pinning test wants updating alongside.

## Smaller things, none blocking

1. **`renderer.sync()` is now eight positional parameters**, and the offline call site passes `0, null,` purely as filler to reach the pacing args. That is the classic mis-ordering trap and it is coordinator-surface growth of the kind `src/render/CLAUDE.md` warns about. Consider collapsing the three new args into one `FramePacingInfo { intentional, targetFps, previousWorkMs }`, or a `renderer.setFramePacing(info)` setter called before `sync`.

2. **The pacer ceiling is coupled to a tier-scoped value.** `maxFps: GFX.budget.targetFps` is per-tier. All four tiers are 60 today so it is preset-neutral, and I confirmed that makes `pacedBudgetScale` consistent. But the day someone sets `low.budget.targetFps = 30` as a mobile perf tweak, a low-preset player's HUD and input sampling halve relative to a high-preset player on identical hardware, which is exactly what the fairness invariant forbids. Worth either pinning "all tiers share `targetFps`" in a test or decoupling the pacer with its own constant.

3. **The 41.2 fps floor deserves a pin.** The whole fairness argument rests on it and it is an emergent property of the divisor math. A sweep asserting `pacedFrameRateFor(r, 60) >= 30` across 30 to 480 Hz would stop a future tweak to `FRAME_RATE_TOLERANCE` or `PACING_ENGAGEMENT_RATIO` from silently dropping it.

4. **`FramePacer.setEnabled` has no production caller** (grep across `src/`); only the tests exercise it. Either wire it to a setting or drop it.

5. **`tests/frame_pacing_integration.test.ts` asserts on `src/main.ts` source text via regex.** It does pin the ordering contract, which I appreciate given the gate placement matters, but it would pass on wrong logic that keeps the right shape and break on harmless refactors. The real coverage is in the pacer and governor suites, so this is a note rather than a request.

6. **Post-gap unpaced burst.** After any gap over 250 ms, `captureCallbackInterval` clears the interval buffer but keeps `trustedRefreshFps`, so `calibrationGate` stays false and `step` early-returns `true` until eight samples refill. I reproduced roughly 8 frames at full panel rate (about 55 ms at 144 Hz) before pacing re-engages. Low severity, but note the 250 ms trigger is any long hitch (GC, asset stream stall), not just app suspend, so it can recur.

7. **A 144 Hz phone runs at 48 fps and a 90 Hz phone at 45 fps** under a 60 fps ceiling. I confirmed this matches your documented table exactly and follows from the whole-divisor policy, so it is intended rather than a bug. Flagging it only because it is the most likely user-visible complaint from this change, and your checklist still has the physical-device row unchecked. Worth a real 90 Hz and 144 Hz device before shipping.

Things I specifically checked and found fine: `hideLoadingScreen` is idempotent so the watchdog-then-recovery double call is safe; `previousFrameWorkMs` is assigned on both exits of `frame()` with no third exit between the gate and those assignments; and `maybePromoteWorkloadLimitedCadence` does not mis-promote under heavy steady load, since `hasHeadroom` requires work under `min(targetInterval * 0.75, trustedPanelInterval)`, which heavy load cannot satisfy.

Happy to re-review as soon as the watchdog path is sorted.

FernandoX7 (commented):
Careful, well-built change. A few things worth calling out before the nits, because they are the load-bearing claims and they hold:

- The extraction is exactly module-first. `FramePacer` and `createLoadingHandoff` are pure, dependency-free `src/game/` siblings with their own unit tests, `main.ts` stays a thin consumer, and both are correctly registered in `UI_PURE_CORES` / `BARE_NAMED`. The `renderer.sync()` positional-to-options refactor is faithful: both call sites are updated and the offline path's unset `selfAlphaLead`/`selfMotion` default to `0`/`null`, matching the old positional call.
- The non-paced path is behavior-preserving. With `enabled: false` (desktop, web, Electron), `step()` always returns `shouldRun: true` / `intentional: false`, so `render_budget.ts` computes `pacedBudgetScale = 1` and `frameMs = cadenceMs`, byte-identical to before. `render_budget.test.ts` staying green confirms it.
- Fairness holds, which was the thing to get right here. Pacing is uniform temporal decimation off a static `MOBILE_FRAME_RATE_CEILING_FPS`, not the FPS governor; the skip `return`s before `last` advances, so the 20 Hz accumulator still owns sim catch-up on the full elapsed delta; movement intent streams on ClientWorld's own 50 ms timer, so paced-out frames never starve server input; and the governor only sheds cosmetic quality. `ui_effects_profile.ts` / `ui_tier_knobs.ts` are untouched.
- The loading handoff is a real robustness win over the old unguarded double-rAF. The 5 s watchdog plus the second bounded completion watchdog means a native launch can no longer stick on the loading screen if frames never arrive, and `render_budget_pacing.test.ts` has genuine negative cases (expensive non-render frame work still degrades; the 18 ms recover case discriminates the scaled 45 fps budget from the unscaled 60 fps one).

I verified the four new test files plus `render_budget.test.ts` and `architecture.test.ts` at the head (green), and `tsc` is clean on the PR's files. No sim, server, wire, or IWorld surface is touched, so no determinism or parity concern. The documented full-suite failures are the known environmental/base set, outside this diff.

Nits, nothing blocking. The first two are the ones I would actually do:

1. `nit` `src/game/frame_pacer.ts:187` and `:40`. `step()` allocates a fresh 4-field decision object on every animation callback, and `median()` re-creates its `(a, b) => a - b` comparator each executed frame. Materiality is low (fixed-shape object plus a 30-element sort at panel rate, and this core is not on the alloc-probed `Hud.update()` path), but the rest of the file is meticulous about reuse (`refreshMedianScratchMs`, `refreshIntervalsMs`, the hoisted `rendererSyncOptions` / `framePacingInfo`), so these two stand out. Reuse a single decision instance and hoist the comparator to a module const.

2. `nit` `tests/loading_handoff.test.ts`. The two-phase watchdog's single-fire is not pinned against the ordering where the completion watchdog fires before the still-queued paint rAF (reachable when the tab backgrounds right after the first frame is marked: the `setTimeout` fires while the rAF is throttled). The queued rAF is not cancelable, so the `!onHandoff` guard in `completeHandoff` (`src/game/loading_handoff.ts:34`) is the only thing stopping that second `completeHandoff` from re-running `onHandoff`. Worth a case: `start`, `markFirstRenderedFrame`, fire watchdog 1, fire watchdog 2, then run the queued rAF and assert `onHandoff` ran once.

3. `nit` `src/game/loading_handoff.ts:34-40` (with `src/main.ts:3540`). `completeHandoff` clears the watchdog before invoking `onHandoff`, and the normal `onHandoff` runs `entryDiagnostics.checkpoint('first-paint')` before `hideLoadingScreen()`. A throw in that first statement would leave the overlay up with no fallback left. Not a regression (the old double-rAF had the same order and no watchdog at all), but since this module exists to never get stuck, hiding the overlay first, or wrapping the body, would close the asymmetry with the `onWatchdog` path, which arms its follow-up before calling.

4. `nit` `src/main.ts:3588`. The native calibration loop awaits up to `FRAME_PACER_CALIBRATION_CALLBACKS` (13) sequential rAFs with only `estimatedRefreshFps > 0` as an early out. A tab backgrounded during a native load throttles rAF to about 1 Hz, and `observe()` rejects those >50 ms intervals, so the estimate stays 0 and the game-loop start is deferred until foreground. The handoff watchdog still clears the overlay so nothing sticks; the only oddity is `gameInputReady` can flip true before `frame()` is running. Low impact (backgrounded means not visible), but a wall-clock cap on the wait would tidy it.

5. `nit` `tests/frame_pacer.test.ts`. `MOBILE_FRAME_RATE_CEILING_FPS` (60) is never pinned to a literal; the fairness sweep only asserts `>= 30`, so a silent change to the ceiling would still pass. A one-line `expect(MOBILE_FRAME_RATE_CEILING_FPS).toBe(60)` closes it.

One narrow edge for the record, not asking for a change: if a device never calibrates (never sees a callback interval <= 50 ms since launch) and then hits a suspend/anomaly while sustaining 4 to 20 fps, `calibrationGate` stays set and the loop paces to roughly 50 percent until the panel returns under 50 ms. It self-heals and is hard to reach on any real panel, so it is fine as is.

Humpalumps:
@FernandoX7 should be fixed and ready for review

Issue #1295, opened 2026-07-02
Professions: crafting legibility UX (trend nudges, tutorial panel, attunement summary)


Nervescraper:
## Problem

#1133 (the four-case drain warning) was correctly closed with the drain model, and its close comment already directs any successor to a new issue: this is that issue. The design's legibility model is different from a warning dialog and currently unowned: informational nudges while crafting, a one-time tutorial panel at the first tier quest, and a full "what changes" summary at the attunement quest. No blocking dialog anywhere.

## Expected behavior

- Common-tier crafting is silent.
- Crafting that trends toward an adjacent pair surfaces an informational nudge ("you are trending toward Cogsmith; its attunement quest is available"). It never blocks and never warns of loss, because crafting never costs another craft. Nudges ship first as plain chat-log info messages and grow lore-friendly later (a Crafting Guild letter, a courier).
- A one-time tutorial panel at the first tier quest (common to uncommon) explains the cap and the skill radar exactly when the mechanic first bites.
- At the attunement quest, the full picture is shown before committing: which crafts become empowered, the new title, that the previous specialty goes dormant (not lost), and the make-amends cost for a return.

## Scope

- `src/sim/professions/wheel.ts`: a pure "trend" query over current skill state (which adjacent pair the player is trending toward), Node-testable.
- Client: the nudge surface, the tutorial panel, and the attunement-quest summary panel, all `t()` keys, following the module-first UI rules (pure view core plus painter).

## Acceptance criteria

- Common-tier crafting produces no nudge.
- A scripted skill state trending toward a pair produces the correct nudge classification.
- The tutorial panel appears exactly once, at the first tier quest.
- The attunement summary lists empowered crafts, title, dormant specialty, and return cost.

## Notes

Replaces closed #1133 under the additive model. Reads #1129 and hobby (#1294) state; changes no crafting outcome. Depends on #1129 (and #1294 for the hobby line); the summary is shown by the #1293 quests.

Full design: https://woc.nervemart.com/docs/professions-system

FernandoX7:
Delivered by PR #2158 (Professions 2.0 Phase 7): the pure trending-pair classifier (src/sim/professions/trend.ts) plus the one-shot Crafting Guild letter it drives. Closed by hand since release-branch merges never auto-close.

Issue #1296, opened 2026-07-02
Professions: Jack of All Trades breadth ceiling and improviser perks


Nervescraper:
## Problem

#1130 grants the Jack of All Trades title, but Jack's distinct build mechanics (a breadth ceiling and improviser perks) have no issue.

## Expected behavior

- Jack of All Trades is itself an attunement (the breadth attunement), offered at the rare tier when the two highest crafts are not adjacent; it is the one sanctioned exception to the at-most-two cap.
- A Jack can level every craft to a mid tier (uncommon or rare, tuning) but none to legendary, so a Jack is the only crafter who can make decent versions of all ten lines.
- Improviser perks: a random chance to produce a better or worse version of what is crafted, and reduced material cost from cross-craft synergy.
- A Jack has no combos (combos require an archetype's attunement, not just skill in both crafts).

## Scope

- `src/sim/professions/wheel.ts`: the Jack breadth ceiling as an alternative to the active-archetype empowerment gate.
- `src/sim/professions/crafting.ts`: the better/worse output roll and material-saving bonus.

## Acceptance criteria

- A Jack reaches the breadth ceiling across all ten crafts and legendary on none.
- The output-variance roll uses `Rng`; the material-saving bonus applies at craft time.
- A Jack cannot craft combo recipes.

## Notes

Depends on #1129 (active-archetype/attunement model) and #1127 (crafting). Open design questions (doc TBD): whether a Jack benefits from Battlefield Experience and for which crafts; whether a Jack gets a higher combined ceiling to afford the breadth; and the mechanics and costs of switching between Jack and an archetype in both directions.

Full design: https://woc.nervemart.com/docs/professions-system

FernandoX7:
Parked to wave 2 by the Professions 2.0 blueprint (docs/professions-2/, PR #2048): Jack of All Trades returns after the wave-one packet ships (the ceiling machinery supports a uniform mid-tier cap, and the blueprint keeps this as the calm landing for non-adjacent builds). Unchanged in spirit.

Issue #1294, opened 2026-07-02
Professions: the hobby (one opposite craft empowered to rare)


Nervescraper:
## Problem

The active archetype (#1129) empowers two adjacent majors. The design adds one exception: a single opposite craft (the hobby) is also empowered, but only up to the rare tier, while the rest of the opposite side stays at the common floor. The mechanic's previous home, #1131, described it as a carve-out protecting one opposite craft from drain and was correctly closed when the drain model was retired, but the empowerment version of the mechanic survived (the doc's rules card: "Hobby: plus 1 of the two crafts opposite your majors, capped at rare") and currently has no owning issue, while #1132's combo exclusivity still references it.

This is deliberately a new issue rather than a reopen, following the #1133 precedent: the closed issue's drain-protection premise was genuinely superseded, and this is the same player-facing outcome under the additive model. To head off a repeat closure for the old reason: this is not the drain carve-out. The design doc keeps the hobby live in its rules card, Archetypes, Combos, and Jack sections and its TBD ("rare, possibly epic; whether it is a hard cap").

## Expected behavior

- A character's hobby is one craft from the pair opposite their majors, empowered up to rare (not to the majors' ceiling). Rare is the working ceiling; rare vs epic and whether it is a hard cap are open tuning.
- The other opposite craft, not taken as hobby, stays at the common floor.
- The hobby is part of attunement: it emerges from what you craft and is revealed when the archetype title is granted. Switching to the other opposite happens by attunement quest, either as part of a full re-attune or standalone at a lower cost than an archetype switch, but never free (flipping opposites must not become a de-facto second hobby). The old hobby goes dormant and the new one is empowered to rare; no menu to manage, no skill loss.

## Scope

- `src/sim/professions/wheel.ts`: hobby designation and its rare-capped empowerment, layered on the active-archetype state from #1129.

## Acceptance criteria

- The hobby craft is empowered to rare and no higher while the majors are worked.
- A hobby switch (standalone quest or full re-attune) swaps which opposite is empowered; the old one goes dormant with knowledge retained.
- The non-hobby opposite craft stays at the common floor.

## Notes

Empowerment exception, not drain protection. Depends on #1129. Read by #1132 (the hobby never unlocks a combo recipe). The standalone hobby-switch quest content belongs to #1293. PR #1205 (implements #1129) should be checked for whether it models the hobby at all; if not, this issue is where it lands.

Full design: https://woc.nervemart.com/docs/professions-system

[...]

PR #2039, opened 2026-07-16
feat(professions): complete quest-driven attunement and online combo crafting

### LINKED ISSUE #1292: Professions: craft and gather quest objective types
## Problem

Quest objectives are `'kill' | 'collect' | 'interact'` only (`src/sim/types.ts`, the `QuestObjective` union). The professions system needs objectives that complete by crafting or gathering a target (the design doc's architecture section names exactly these two kinds; cooking is a craft, so a cook objective is a craft objective): the attunement and switch quests, commission orders, and recipe-acquisition quests all rely on it.

## Expected behavior

- A quest objective can require crafting a recipe or gathering a material or node type, and completes when the player performs that action the required number of times.
- Existing objective types are unaffected.

## Scope

- `src/sim/types.ts`: extend the `QuestObjective` union with the new kinds.
- Quest progress resolution: credit the new objective kinds from the crafting and gathering events.
- Sim/server i18n: emit a stable key plus values for the new objective labels (matcher in `src/ui/sim_i18n.ts` in the same change; the S3 guard `tests/localization_fixes.test.ts` enforces it).

## Acceptance criteria

- A craft or gather objective advances only on the matching action and completes at its count.
- Existing kill/collect/interact objectives are unchanged (no regression).
- Objective text renders via `t()`.

## Notes

Pure prerequisite: blocks the attunement quest system (filed next) and the upcoming commission and recipe-acquisition issues. Build behind the SimContext seam; resolves server-side on the 20 Hz tick.

### LINKED ISSUE #1293: Professions: attunement quest system and content (lore quest, make-amends, hobby switch)

### LINKED ISSUE #1302: Professions: wheel and identity HUD
## Problem

#1124 covers the gathering HUD only. The crafting identity surface (the per-craft skill radar, the active archetype and title, the empowerment envelope) has no issue.

## Expected behavior

- A player can see their per-craft skill radar, their active archetype and title, and which crafts are empowered vs at the floor.
- The radar reflects the trend nudges (#1295) and updates as skill changes.

## Scope

- `IWorld`: the crafting-identity read surface (skills, active archetype, empowerment state).
- Client: an identity card or window, module-first (pure view core plus painter), all `t()` keys.

## Acceptance criteria

- The identity card shows correct per-craft skill, active archetype and title, and empowerment state.
- All strings are `t()` keys; per-frame cost stays within the UI budget.

## Notes

Consumer of the #1126, #1129, #1130, and #1294 read surfaces. Can build against a stub facet.

### LINKED ISSUE #2033: Combo recipes cannot be crafted in online play: craft skills and archetype are never mirrored to the client
## Problem

In online play, every combo recipe (the three `COMBO_RECIPES` in `src/sim/content/recipes.ts`: Elixir of the Bear, plus the boundstone helm and gravewyrm gauntlets outputs) renders with a permanently disabled Craft button, regardless of the player's real craft skills. The crafting window's combo gate reads `IWorld.craftSkills`, and `ClientWorld.craftSkills` (`src/net/online.ts`) is a documented all-zero stub that no snapshot field ever updates: the server tracks and persists the real per-craft skills but never serializes them onto the wire (the `prof`/`gprof` keys carry gathering proficiency only). `activeArchetype` and the archetype commands have the same never-wired status.

Because the disabled button and its click guard prevent the `craft_item` command from ever being sent, the authoritative server-side check (which reads the player's real skills and would accept a qualified crafter) never runs. The window also surfaces only reagent lines (inline, tooltip, and accessible name), so the combo requirement is invisible and the player gets no explanation: all listed requirements read as satisfied while the button stays dead. This was reported by a player on the public Discord.

## Steps to reproduce

1. On a server-connected (online) character, raise Alchemy and Engineering to skill 25 or higher (tier 1) and gather 2 Linen Scrap and 2 Twitching Spider Leg.
2. Open the crafting window and locate Elixir of the Bear (`recipe_volatile_flux_elixir`).
3. Observe the Craft button stays disabled even though every requirement the window shows is satisfied, and nothing explains why.

## Expected behavior

- A server-connected player's craft skills and archetype identity are mirrored to the client through the existing `IWorld` read surface, so progression-dependent UI reflects the authoritative state.
- The crafting window's Craft button state matches what the authoritative sim would decide for that player: the view gate agrees with `meetsComboRequirement` (the archetype-capped `craftCeiling`), not raw `tierCapability`.
- A combo recipe's dual-craft requirement (both craft names and the tier) is visible in the window row, tooltip, and accessible name, including when unmet.
- Offline (local Sim) behavior is unchanged for players who already qualify.

## Actual behavior

- `ClientWorld.craftSkills` is initialized to `emptyCraftSkills()` and never assigned again; `activeArchetype` stays `null`; `acceptArchetypeQuest`/`advanceAmendsProgress`/`switchArchetype` are no-op stubs with no `cmd` dispatch.
- The self-snapshot assembly (`selfWireJson` in `server/game.ts`) carries no craft-skill or archetype field under any name, and the pinned `ALL_DELTA_KEYS` registry in `tests/snapshots.test.ts` confirms no such wire key exists.
- `buildCraftingView` (`src/ui/crafting_view.ts`) therefore computes tier 0 for every craft online, so `craftable` is false for every combo recipe; `renderCraftingWindow` disables the button and guards the click, so the craft attempt never reaches the server.
- The localized `combo_requirement_unmet` toast exists but is unreachable online because the attempt never fires.

## Scope

- `server/game.ts` (`selfWireJson`): new self-snap delta key(s) for craft skills and archetype state, following the `ncd`/`gprof` pattern. Do not overload `prof`/`gprof`, which carry gathering data only. Add the matching `ALL_DELTA_KEYS`/`TERSE_TO_IWORLD` registry rows and a value-level mirror test in `tests/snapshots.test.ts`.
- `src/net/online.ts`: decode the new key(s) into `craftSkills`, `activeArchetype`, and the paired-major surface the ceiling math needs; replace the stale stub comments (they reference a retired "wheel/mass-conservation follow-up").
- `src/ui/crafting_view.ts` and `src/ui/crafting_window.ts`: align the `craftable` gate with the sim's archetype-capped `craftCeiling`, and surface the combo requirement text (new `hudChrome.crafting.*` catalog keys, English only per the i18n workflow).
- Test reconciliation: `tests/archetype_ceiling.test.ts` and `tests/crafting_view.test.ts` currently pin opposite outcomes for the same 25/25 outside-archetype player shape; aligning the view requires deliberately updating the view-side pin.

## Acceptance criteria

- An online player with both combo crafts at the required tier (and a qualifying archetype state) sees the Craft button enabled and can craft Elixir of the Bear end to end.
- An online player who does not qualify sees the specific unmet requirement (both craft names and the tier) in the recipe row or tooltip and in the accessible name.
- `ClientWorld.craftSkills` and `activeArchetype` reflect the server's persisted values after login and after each skill-up or archetype change, covered by a value-level snapshot mirror test (not just the structural parity pin).
- The view gate and `meetsComboRequirement` agree: no player shape exists where the button is enabled but the authoritative craft is denied with `combo_requirement_unmet`, or vice versa.
- The full gate (`npm run gate`) passes; offline crafting behavior is unchanged for qualifying players.

## Notes

- Child issue of epic #1866, which names this gap ("craft skills remain zeroed, archetype/attunement reads return placeholder values") and sets the direction that combo eligibility should require the matching active attunement; the view-gate alignment here should follow that direction rather than re-implementing raw skill checks.
- History: the stub landed with the sim-side-only scope of #1126/#1129; the follow-up its comment promises was retired per the #107 design review, so this issue is the bounded wire-up vehicle.
- The wire-up pattern to copy exists in the same file and epic: `nodeCooldowns` was wired under #1866 via the `ncd` self-wire delta with its registry rows and mirror test.

MasterZensei:
## Summary

This PR makes profession progression a reachable, server-authoritative player loop in both offline and online play.

- Add first-class `craft` and `gather` quest objectives, authoritative crediting, stable progress text, target resolution, persistence, and parity coverage.
- Replace the retired profession quest placeholders with selectable lore attunement, repeatable make-amends, and hobby-switch flows. Quest acceptance and turn-in validate the selected target, persist resolved requirements, and apply profession transitions transactionally.
- Centralize combo-recipe eligibility around the exact active adjacent pair and required tier, then consume that rule from both the simulation and crafting UI.
- Mirror craft skills and the complete crafting identity atomically through the online `cprof` snapshot so client gating matches the authoritative server after login, skill changes, attunement changes, reconnects, and persisted reloads.
- Add the profession identity card, trend/tutorial/attunement guidance, dormant-knowledge presentation, accessible combo explanations, mobile touch sizing, and refreshed localization output.

## Maintainer amendments (Professions 2.0, Phase 1)

Landed directly on this PR so it merges once, already aligned with the approved blueprint (`docs/professions-2/`):

- Synced the branch with the current `release/v0.27.0` head (Card Duel facet, SFX batch, renown board, trade payload fix, planning packet). The `world_api_parity` pin is the union of both sides (239 members, 28 facets) and the delta-key census is 47 (`cprof` + `cardDuel`).
- Adopted the blueprint `CRAFT_RING` order (engineering, alchemy, cooking, leatherworking, tailoring, inscription, enchanting, jewelcrafting, weaponcrafting, armorcrafting) inside this PR's merge window, so canonical pair ids derived from the old ring never exist in a deployed build. Both combo-recipe pairs stay ring-adjacent, pinned by a new content-driven test.
- Replaced the ten per-craft practitioner titles with pair-named archetype titles (Smith, Outfitter, Apothecary, Bombardier, Trapper, Mageweaver, Arcanist, Gembinder, Bladewright, Cogsmith) keyed by canonical pair id, plus per-craft display names for every surface that means a craft rather than a title. `archetypeTitle` resolves the canonical pair id identically in `Sim` and `ClientWorld`.
- Resolved the five review items: `sim_i18n` matcher rows for the two profession-choice denial strings (plus `quests/quest_commands.ts` on the S3 scan list), the issue-coverage reword below, the 13 Latin-overlay hand fills reverted to pending, commit bodies via history rewrite, and the PRD status refresh.

## Screenshots

**Crafting identity and exact-pair combo guidance**

Crafting window showing the profession identity card and readable exact-pair combo requirement

**Selectable attunement quest with consequence preview**

A Craft to Call Your Own quest showing adjacent profession-pair choices and the resulting title, majors, hobby, and retained knowledge

## Root cause

The existing profession pieces were disconnected across layers. Quest objectives could not represent crafting or gathering, the live attunement quests were retired placeholders, online clients retained zero-skill/null-archetype stubs, and the UI maintained a separate raw-skill combo check. That left profession transitions unreachable and caused qualified online players to see permanently disabled combo recipes with no useful explanation.

## Player and developer impact

- Players can progress through real profession objectives and choose an attunement through normal quest play.
- Returning to a prior pair and switching hobbies use explicit, escalating, validated quest flows.
- Combo recipes require the exact active pair; unattuned, wrong-pair, hobby, and low-tier states receive a specific explanation.
- Online and offline hosts now expose the same crafting identity and eligibility behavior.
- The granted identity is the pair-named archetype title; craft names and titles are separate localization surfaces.

## Issue coverage

Closes #2033
Closes #1292
Closes #1293
Part of #1295 (the trend classification, one-time tutorial timing, and make-amends preview cost remain open; see the PR review)
Closes #1302

Related: #1294 (already completed), #1866 (broader professions epic; remains open).

Intentionally deferred: #1296 (Jack of All Trades breadth ceiling, perks, and switching decisions).

## Validation

- `npx tsc --noEmit` clean; `npm run gate` (PR tier) run on the amended head.
- The professions, archetype, combo-eligibility, snapshot, world-api-parity, architecture, localization, and parity-golden suites all pass; every ring-derived pin was deliberately re-pinned, never loosened.
- Six read-only review agents (privacy-security, migration-safety, cross-platform-sync, architecture, frontend-seam, qa-checklist) passed the amended head with zero blocking findings; the one should-fix (documenting the real save-compat invariant at `normalizeArchetypeState`) is applied.
- Save-compat: no shipped build persists profession state (the acceptance quests were retired in every deployed release), and the ring order ships in the same release as the live quest wiring; a test pins that the deployed v0.26.0 save shapes load as the exact empty state.

## Notes

The PR is based on and targets `release/v0.27.0`. The broader professions integration epic still owns gathering controls, tools, item identity/economy transfer work, advanced profession actions, and Jack of All Trades behavior outside this bounded loop. Phase 1 packet docs (`docs/professions-2/state.md`, `progress.md`) are updated in this PR.


FernandoX7 (commented):
This is strong work, and it lands the missing center of the professions loop. I verified the load-bearing claims against the code and ran the matrix locally at the PR head (no CI has run on this fork branch): tsc clean, the touched test files pass (471 tests across 24 files), the full parity golden suite passes, the i18n artifacts are byte-identical to regen output, and the goldens were regenerated through the harness rather than hand-edited. The merge into release/v0.27.0 is conflict-free.

Highlights that checked out under scrutiny: the #2033 fix is the sanctioned self-wire pattern done right (atomic `cprof` mirror with a `synced` flag and a pre-cprof fallback, one shared `comboEligibility` rule consumed by both the authoritative resolver and the UI, and an end-to-end regression test in tests/snapshots.test.ts that goes server snapshot to client mirror to view gating to authoritative craft). Server authority is solid: the client-supplied selection is validated at accept and turn-in against whitelisted pair/hobby targets, revalidated inside the turn-in transaction, quest sharing is double-gated, and a test pins that no direct profession-transition command exists on the wire. The archetype state machine is deterministic (ring-order tiebreaks, no wall clock, no Math.random) with careful old-save backfill, and the persisted quest fields are additive on both the save path and the sanitize path.

Findings:

**should-fix**

1. `src/sim/quests/quest_commands.ts:168` and `:247`: the two new server-sent error strings ("That profession choice is not available." / "That profession choice is no longer available.") have no rows in the client error matcher (the exact-match map at `src/ui/hud.ts:9251`), so online non-English players see raw English for both. CLAUDE.md requires the matcher rows in the same change as the sim emit. The S3 guard stayed green only because its scan list has never included `quests/quest_commands.ts` (a pre-existing blind spot, not this PR's regression); appending that file to the list in tests/localization_fixes.test.ts while you are in there would keep this from recurring.
2. "Closes #1295" over-claims. Three of its acceptance criteria are unmet: there is no pure trend query in `src/sim/professions/wheel.ts` (the shipped nudges are nearTier/dormantKnowledge, not a trending-pair classification), the tutorial in `src/ui/profession_identity_view.ts` renders until any craft reaches tier 1 rather than appearing exactly once at the first tier quest, and the attunement preview omits the make-amends return cost. Suggest "Part of #1295" so the merge does not auto-close it. #1292, #1293, #1302, and #2033 all check out against the diff.
3. The 13 Latin-script overlays add hand-authored q_prof_hobby_switch translations (e.g. `src/ui/i18n.locales/fr_FR.ts:7223`) that no gate requires. Contributors add English only, and `scripts/i18n_scan.mjs` registers overlay rows as human-translated, so these unreviewed fills silently drop out of the release-time fill worklist. Please revert the additions in the Latin overlays and leave those rows pending. To be explicit about the rest of the overlay diff: the five non-Latin fills are required by the PR-tier gate (I confirmed it fails without them) and the stale-row removals for the reworded quests are correct, so both of those stay.
4. Both commits are title-only; the repo convention requires a body on every commit. Amending, or squash-merging with a body, resolves it.
5. `docs/prd/professions-combo-attunement-implementation-plan.md:16`: the "Implementation status" section is already false at this PR's head. It says the worktree "has not been committed, pushed, or opened as a pull request" while this very PR commits it, and "localization release checks pass" while pending.ts correctly carries 37 pending rows per Latin locale (expected for a PR; the release-tier gate hard-fails on them). Refresh or drop the status claims per the docs anchor rule.

**nit**

6. `src/sim/encounters/nythraxis.ts:1330`: the questProgress emit hardcodes `required: objective.count` where every other emit site routes through `questObjectiveRequired`. Identical today; drifts if an encounter quest ever uses resolvedObjectiveCounts.
7. `src/net/online.ts:2319`: `questState()` builds a fresh profession-state object plus an attunedPairs spread on every call, and it is called from per-frame paths (nameplate painter, minimap markers). Building the mirror object once when `cprof` is applied keeps the hot path allocation-free. The `identity ? ... : undefined` ternary is also dead, since `craftingIdentity` is always set.
8. Two test gaps worth closing: nothing drives the turn-in revalidation failure arm (the "no longer available" branch, reachable when a held hobby-switch selection is invalidated by an attunement turn-in before it), and the new first-completion-only guard on `counters.questsCompleted` for repeatables is not asserted anywhere.
9. The /dev quest cheats cannot exercise the new quests: the dev accept path calls `finalizeQuestAccept` without a selection, which now fails validation for completionEffect quests. Dev-only, fine to defer.

Notes, checked and fine: `finalizeQuestAccept` now snapshots `resolvedCounts` for every quest at accept (the goldens show q_wolves carrying `resolvedCounts: [8]`), so a future content rebalance of an objective count will no longer apply to in-flight quests; deterministic and back-compat both ways, but flagging it in case freeze-at-accept for ordinary quests was not deliberate. The repeatables cannot farm anything that matters: questsCompleted only counts first completions, no deed reads the progress counters, and the hobby-switch XP rate is worse than normal play. In-flight q_prof_intro saves keep their progress index and continue via gather credit, with any leftover chunk_of_ore inert but discardable.

PR #2042, opened 2026-07-17
feat(metrics): add cohort-safe day-one engagement gauges

sf-chris:
## Summary

- Add a six-stage signup funnel where every stage is drawn from the same account-created UTC-day cohort.
- Keep first-day playtime, session, and retention metrics on their separate first-play UTC-day cohort.
- Reshape the snapshot into indexed per-day reads so it remains bounded under the existing collector timeouts.
- Cover delayed first play, no-fact registrations, realm isolation, UTC boundaries, and all playtime bucket edges on real PostgreSQL.

Companion Grafana dashboard PR: [woc-deploy#21](https://github.com/levy-street/woc-deploy/pull/21).

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands:
  - `npm run check:types`
  - `npm run ci:changed`
  - `npm run security:gate`
  - `npx vitest run tests/player_metrics_db.test.ts tests/server/http/business_metrics.test.ts` (15 passed)
  - `TEST_DATABASE_URL=... npx vitest run tests/player_metrics_db_integration.test.ts` against PostgreSQL 14.15 (6 passed)
  - Repository pre-push QA floor (53 passed, 3 skipped)
- Database plan:
  - Exact-schema fixture with 200,000 accounts and more than 200,000 activity rows.
  - 56.381 ms execution under a 2-second statement timeout and 4 MB `work_mem`.
  - No temp I/O and no sequential scans of durable facts or activity tables.
- Review:
  - Database performance, test coverage, and security reviews report READY.
- Full gate note:
  - `npm run gate` was attempted locally, but the host had about 331 MiB free and a nested-checkout test hit `ENOSPC`, causing cascading unrelated worker failures. An unrelated existing snapshot test also fails in isolation because `WebSocket` is undefined. PR CI is the clean full-gate authority for this change.
- Manual steps:
  - Confirmed exporter metric names, fixed label domains, and dashboard contract alignment.

### Quality

- [ ] **The full gate passes.** Awaiting clean PR CI; see the local environment note above.

### Localization (i18n)

### Hygiene

### DIFF
--- a/server/http/business_metrics.ts
+++ b/server/http/business_metrics.ts
@@ -5,6 +5,8 @@
 import { Gauge, type Registry } from 'prom-client';
 import { pool } from '../db';
 import {
+  DAY_ONE_FUNNEL_STAGES,
+  FIRST_DAY_PLAYTIME_BUCKETS,
   type PlayerBusinessDay,
   type PlayerBusinessSnapshot,
   playerBusinessSnapshot,
@@ -21,6 +23,10 @@ export const WOC_PLAYER_DAILY_PLAYTIME_SECONDS = 'woc_player_daily_playtime_seco
 export const WOC_PLAYER_FIRST_SESSION_MEDIAN_SECONDS = 'woc_player_first_session_median_seconds';
 export const WOC_PLAYER_FIRST_SESSION_LEVEL_RATE = 'woc_player_first_session_level_rate';
 export const WOC_PLAYER_RETENTION_RATE = 'woc_player_retention_rate';
+export const WOC_PLAYER_FIRST_DAY_PLAYTIME_SECONDS = 'woc_player_first_day_playtime_seconds';
+export const WOC_PLAYER_FIRST_DAY_SESSIONS = 'woc_player_first_day_sessions';
+export const WOC_PLAYER_FIRST_DAY_PLAYTIME_ACCOUNTS = 'woc_player_first_day_playtime_accounts';
+export const WOC_PLAYER_FUNNEL_ACCOUNTS = 'woc_player_funnel_accounts';
 
 /** Business data changes slowly; one bounded database sample every 15 minutes is enough. */
 export const BUSINESS_METRICS_REFRESH_MS = 15 * 60_000;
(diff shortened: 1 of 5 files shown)


FernandoX7 (commented):
Post-merge review; everything below is a follow-up, not a blocker.

Solid work. The LATERAL rewrite keeps every access path bounded by a UTC-day cohort, the new label sets are fixed constants, and the UTC-day claims in the metric help text are actually backed by the `SET LOCAL TIME ZONE 'UTC'` in `playerBusinessSnapshot` (server/player_metrics_db.ts:712). I traced the rewritten query against the old grouped version: `accounts_created`, `first_world_entry_rate` (the added same-day `first_character_at` condition is implied by account-created plus first-play on the same day, so no behavior change), `active_new`, `avg_playtime_new`, and retention all keep their previous semantics. The integration fixture's `accounts_created_at` index matches the production DDL (server/db.ts:412), and the real-Postgres coverage of delayed first play, factless registrations, realm isolation, the next-day boundary, and every bucket edge is exactly what this kind of SQL needs.

Findings:

- should-fix (follow-up): the funnel CTE trades the old index-only `count(*)` over `accounts(created_at)` for a per-account nested loop, two lateral PK probes into `player_account_facts` and `player_activity_daily` for every account created that day (server/player_metrics_db.ts:436-493). Small on a normal day, but a launch-day signup spike (tens of thousands of created accounts, most with no lifecycle fact, cold cache) could push the statement past the 2s `statement_timeout`, and since every business gauge comes from this one statement, a funnel timeout stales the cheap gauges (accounts created, retention, first session) exactly when day-one metrics matter most. The 56 ms plan figure in the PR body depends on how many of the 200k fixture accounts fall inside the measured day, so it does not settle this case. Two cheap escapes if it ever bites: drive the later funnel stages from the facts-side `first_character_at` day cohort (skipping probes for the factless majority), or split the funnel into its own statement so its timeout cannot take the other gauges with it.
- nit (pre-existing, made more relevant here): when a refresh fails, `PeriodicCollector` logs and keeps the last good snapshot (server/http/periodic_collector.ts:57), so a persistent timeout freezes the gauges with nothing an alert can watch. A refresh-failure counter or snapshot-age gauge would make that state visible.
- nit: `DAY_ONE_FUNNEL_STAGES` orders `played_10m` before `reached_level_2`, but neither implies the other, so the funnel is not monotonic across those two stages (the unit fixture itself has `played_10m` 0 and `reached_level_2` 1). Fine in the data, just worth remembering on the Grafana side so a rising step mid-funnel does not read as a bug.

Notes, no action needed:

- The `created` stage counts accounts globally while every later stage is realm-scoped through facts, so on a multi-realm deployment the created-to-first_character drop partly reflects realm choice rather than abandonment. Inherited from the old `accounts_created` semantics, not new here.
- The `LIMIT 1` markers in the laterals are redundant against the primary keys but useful: they keep the planner on the bounded per-row probe plan instead of de-correlating into a full-table join. Worth keeping if anyone is tempted to clean them up.

PR #2045, opened 2026-07-17
fix(sim): preserve item instance payload across a player trade

jgyy:
## Summary

- Trade swapped offered items via `removePreferFungible` + `addItem`, discarding the removed `ItemInstancePayload` and re-granting a plain copy on the other side. An enchanted, signed, or quality-rolled item silently lost that data on every trade.
- Route the consumed instances back through the existing `addItemInstance` hub instead (the same primitive `enchanting.ts` already uses to grant instanced copies), so trade now preserves enchants, signed materials, and rolled quality.
- The capacity gate (`fitsAfterSwap`) predicted room for a receive assuming every unit could stack, but `transferOffer`'s `addItemInstance` grant never merges into an existing stack and always takes a fresh slot. A receiver at exactly full capacity holding a partial plain stack of the offered item could pass the gate and end up over capacity via any stackable instanced item (crafted rolled-quality results, signed items). Fixed by predicting the instanced portion of each receive (mirroring `removePreferFungible`'s own fungible-first split against the giver's actual stock) and requiring a free slot per instanced copy.

## Root cause

`tradeConfirm` called `removePreferFungible` (which already returns the exact `ItemInstancePayload[]` it consumed) and discarded the return value, then called `ctx.addItem` for a plain copy regardless of whether the removed slot was instanced. Both primitives already existed; they were simply never wired together for trade.

### Before: the payload is silently dropped

```mermaid
sequenceDiagram
    participant A as Player A (sender)
    participant Trade as tradeConfirm
    participant B as Player B (receiver)

    Note over A: Holds an ENCHANTED sword<br/>{ itemId, count: 1, instance: {...} }
    Trade->>A: removePreferFungible(itemId, 1)
    A-->>Trade: returns [instance] (discarded, not captured)
    Trade->>B: addItem(itemId, 1)
    Note over B: Receives a PLAIN sword<br/>instance payload lost forever
```

### After: the payload rides along via `addItemInstance`

```mermaid
sequenceDiagram
    participant A as Player A (sender)
    participant Trade as transferOffer
    participant B as Player B (receiver)

    Note over A: Holds an ENCHANTED sword<br/>{ itemId, count: 1, instance: {...} }
    Trade->>A: removePreferFungible(itemId, 1)
    A-->>Trade: instances = [instance]
    alt instance was consumed
        Trade->>B: addItemInstance(itemId, instance)
        Note over B: Receives the SAME enchanted sword<br/>instance payload preserved
    else plain copy remained
        Trade->>B: addItem(itemId, plainCount)
    end
```

### The capacity-gate hole this also closed

`fitsAfterSwap` used `fitsAll`, which stacks every receive via `addStacked` (a partial
plain stack of the same itemId counts as room). But the real transfer grants each
instanced copy via `addItemInstance`, which always pushes a fresh slot and never
merges. A receiver at full capacity holding a partial plain stack of the offered
itemId would pass the gate and end up over capacity, unbounded and repeatable, for
any stackable instanced item. The fix predicts, per receive slot, how many units the
giver's actual stock forces to be instanced (`count - ctx.countFungibleItem(itemId,
giverPid)`), lets only the plain remainder stack, and requires one free slot per
instanced unit, mirroring `removePreferFungible`'s own split.

## Context

Surfaced by an internal audit of the professions design doc against `release/v0.27.0` (see `docs/design/professions-system` and the professions epic #1866), which flagged this as the highest-severity bug in the tree: it is silent, universal (any trade of an instanced item), and destroys player value with no error or log. The capacity-gate hole was caught in code review on this PR.

## Changes

- `src/sim/social/trade.ts`: extract a `transferOffer` helper that re-grants each consumed instance via `ctx.addItemInstance`, falling back to `ctx.addItem` only for the plain (non-instanced) remainder. `fitsAfterSwap` now takes the giving side's pid and predicts the instanced/plain split of each receive the same way, instead of assuming every unit can stack.
- `tests/trade.test.ts`: a shared fake-`SimContext` factory (real per-slot inventory arrays, instanced payloads, capacity-aware fungible/removeItem semantics) backs three regression tests: preserving an instanced payload across a swap, rejecting a trade that would push a full-capacity receiver over the limit via an instanced grant (reproduces 17/16 slots without the fix, passes with it), and a mixed offer split between a giver's plain and instanced copies of the same item in one transfer.

## Validation

- `npx vitest run tests/trade.test.ts tests/architecture.test.ts tests/items.test.ts tests/world_api_parity.test.ts`: all pass (298 tests).
- `npx tsc --noEmit -p .`: clean.
- Pre-push QA floor (`tsc`, guard tests, biome changed-files, copy scan): green.


FernandoX7 (commented):
The core fix is right: `removePreferFungible` already returns exactly the consumed payloads, and routing them back through `addItemInstance` is the correct primitive. The regression test is genuinely failing-first (I ran it against the base `trade.ts`: fails with `expected undefined to deeply equal { signedBy: 'Ayla', ... }`, passes with the fix). I also checked the binding angle: `tradeSetOffer` already rejects soulbound and quest defs, and `boundTo` is set nowhere in production code today, so transferring the payload opens no bind bypass. Determinism, module placement, and the SimContext usage are all fine.

One real problem though:

- **blocking** `src/sim/social/trade.ts:167`: the capacity gate now under-predicts for instanced receives. `fitsAfterSwap` simulates the receive with `fitsAll`, which stacks via `addStacked` (a partial plain stack of the same itemId counts as room, see `countFit` in `src/sim/bags.ts`), but `transferOffer` grants each instanced copy via `addItemInstance`, which always pushes a fresh slot and never merges (`src/sim/sim.ts:5866`). A receiver at exactly full slots holding a partial plain stack of the offered itemId passes the gate and ends over capacity. I reproduced it with this PR's own fake-ctx pattern: receiver at 16/16 with a plain `wolf_fang` x1 stack, sender offers 1 instanced `wolf_fang`, no error is emitted and the receiver ends at 17 slots against 16. It also repeats without bound, because the plain stack's room never shrinks (instanced copies never merge into it), so two players can pump an inventory arbitrarily past capacity with any stackable instanced item (crafted rolled-quality results, `src/sim/professions/crafting.ts:373`; signed items, `src/sim/interaction.ts:283`). The layout tolerates overflow (`inventory_order.ts` keeps stacks visible past the grid), but its own comment says the bag refuses new items long before that path is reachable, and this makes it reachable at will. Suggested shape: in `fitsAfterSwap`, predict the instanced portion of each receive (what `removePreferFungible` will take from the giver: `max(0, s.count - giver countFungibleItem)`) and require a free slot per instanced copy, stacking only the plain remainder.

Non-blocking:

- **should-fix** `src/sim/social/trade.ts:127`: the comment claims a sold item's payload "does not need to reappear anywhere", but `sellItem` records vendor buyback (`src/sim/items.ts:439`) and buyback re-grants a plain copy, so a sold instanced item loses its payload on buyback today. That is a pre-existing sibling of this same bug, not this PR's fault, but the comment should not cite it as a case that needs nothing. Worth a follow-up issue.
- **nit** `tests/trade.test.ts:159`: the mixed case (an offer covered partly by plain copies and partly by an instanced one, so the `plainCount` arm and the instance arm both fire in one transfer) is untested. Related: the fake's `removeItem` pushes `inv[i].instance` unconditionally (undefined for plain slots) and counts removed slots rather than stack counts, unlike the real contract (`sim.ts` `removeItem` pushes only instanced payloads and honors `s.count`), so extending this fake to the mixed scenario would miscount. Also, the payload uses `signedBy`, but the real `ItemInstancePayload` field is `signer` (`src/sim/types.ts`); harmless since the test treats it as opaque, but matching the real shape would keep the fixture honest.
- **nit**: the PR body's Validation section uses em dashes; the repo bans them in PR text as well (root CLAUDE.md).

jgyy:
Thanks for catching the capacity-gate hole! Pushed a fix:

- **blocking**: \`fitsAfterSwap\` now takes the giving side's pid and, per receive slot, predicts how many units the giver's actual stock forces to be instanced (\`count - ctx.countFungibleItem(itemId, giverPid)\`), mirroring \`removePreferFungible\`'s own fungible-first split. Only the plain remainder is allowed to stack (via \`countFit\`/\`addStacked\`); each instanced unit now requires its own free slot, matching what \`transferOffer\`'s \`addItemInstance\` actually does. Added a regression test with a receiver at exactly 16/16 slots holding a partial plain \`wolf_fang\` stack: confirmed it reproduces 17/16 overflow against the pre-fix code and is rejected with "not enough bag space" after the fix.
- **should-fix**: corrected the \`transferOffer\` comment, it cited \`sellItem\` alongside \`discardItem\` as needing nothing extra, but \`sellItem\` records vendor buyback and buyback still re-grants a plain copy today, so a sold instanced item loses its payload there. Called that out as a pre-existing sibling bug, not fixed by this change.
- **nit**: added the mixed-offer test (an offer split between a giver's plain and instanced copies of the same item in one transfer), fixed the fake ctx's \`removeItem\`/\`countItem\`/\`countFungibleItem\` to honor \`s.count\` and treat only \`!s.instance\` slots as fungible (matching the real \`sim.ts\` contract), and renamed the fixture's \`signedBy\` field to \`signer\` to match the real \`ItemInstancePayload\` shape.
- **nit**: fixed the em dashes in the PR body's Validation section.

Verified: the new test fails against the pre-fix \`trade.ts\` (17-slot overflow) and passes after; \`tsc --noEmit\` clean; \`trade\`/\`architecture\`/\`items\`/\`world_api_parity\` suites all green (302 tests); pre-push QA floor green. Also merged the current \`release/v0.27.0\` in (Card Duel landed since) with zero conflicts.

FernandoX7 (approved):

PR #2047, opened 2026-07-17
feat(sfx): land every custom recording, computed gain-ceiling system, and gain map

jamiecypher:
## Summary

Lands the full library of hand-recorded custom SFX (mob vocalizations across every family, combat/melee/impact one-shots, footsteps, movement, lockpick, quest, wand, and the six magic-school proj_/impact_ pairs), plus the computed per-key gain-ceiling system that replaces the old flat 0dB boost cap for `custom: true` keys with a real, measured-headroom maximum per key (`scripts/sfx/sfx_gain_ceiling.mjs`).

The bulk of this PR is the QA process that followed landing that library, run against the actual `npm run gate` / qa-checklist process (5 clean passes), not ad hoc checks:

- **Two real production bugs** found and fixed: `src/game/sfx_runtime_pack.ts` had a hardcoded flat gain ceiling that would have made the new per-key boosting inert in production; `scripts/sfx_studio/export_bundle.mjs` was missing `preserveLoudness` threading, which would have blocked Studio's "Export All" for the whole custom library.
- **A systemic placeholder-audio bug**, root-caused via a duration-mismatch audit (published-file duration vs. the matching master-store recording; conform never trims, only re-encodes, so a real match should be duration-identical): 13 `custom: true` keys were still silently playing ElevenLabs placeholder audio despite the flag, undetected by loudness-based checks because the placeholders weren't loud enough to trip them. All 13 reconformed from the pristine master-store source. Added a permanent advisory guard (`classify()` in `scripts/sfx/sfx_conform_rules.mjs`) that flags a `preserveLoudness` key whose LUFS lands suspiciously close to the generated-content target, the fingerprint of this exact bug recurring.
- **A second, independent gain-stacking bug**: `hud.ts` had leftover hardcoded gain multipliers on the quest sounds, a workaround from when those files were quiet placeholders. Once the underlying files were correctly conformed, the old multiplier stacked on top of the correct gain-map value and pushed all three into real clipping. Removed; the manifest's resolved gain is now the single source of truth like every other sound.
- **A stale-preload bug**: quest sounds defaulted to lazy preload (fell into the manifest's generic "other" category), so a brand-new player's very first quest could silently drop the accept/complete cue if the cold network fetch missed a 250ms retry window. Fixed by adding `quest_` to the startup-preload set.
- Cleaned up 25 confirmed-dead orphan files (stale ElevenLabs leftovers or exact duplicate takes, never consumed by manifest discovery), each individually verified before deletion, no automated sweep.
- Licenses @jamiecypher's original SFX contributions under CC BY-NC 4.0 with a perpetual royalty-free commercial grant to World of ClaudeCraft (confirmed with the project admin), documented in `CREDITS.md`.

## Type of change

- [x] Bug fix
- [x] Documentation
- [x] Tests

## How was this tested?

- Commands: `npm run gate` (biome, tsc, i18n gen, `sfx:check`, full vitest suite, build) run clean multiple times across the session, most recently after a full rebase onto the latest `release/v0.27.0`; `npx vitest run` full suite (only the two known pre-existing resource-contention flakes, `characterization.test.ts`/`parity.test.ts`, confirmed unrelated and passing in isolation).
- Manual steps: full in-game listen pass across all 9 classes' nukes/heals/wands, every mob family, movement, melee weight classes, archery, lockpicking, quest accept/progress/turn-in, and the Book of Deeds unlock chime, via `npm run dev` + `npm run server`.
- Judgment review: 5 qa-checklist passes, final verdict READY with no blocking findings.

## Screenshots / recordings

N/A. No player-visible UI changes, audio-only.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green (except the 2 known pre-existing i18n flakes and 1-2 known pre-existing resource-contention flakes, both confirmed unrelated to this change and documented in the PR history). I added or updated decisive tests for changed behavior and recorded manual checks above.

### Cross-platform

- [ ] N/A. Audio-only change, no UI.

### Localization (i18n)

### Hygiene

- [x] I didn't hand-edit generated files. `src/game/sfx_manifest.generated.ts`, `public/audio/sfx/runtime-pack.json`, and `scripts/sfx/sfx_gain_ceiling.generated.json` were all regenerated by `npm run sfx:manifest`.


jgyy (commented):
This is a genuinely impressive QA pass, not just an asset drop. The duration-mismatch audit that caught 13 keys still silently serving ElevenLabs placeholders despite `custom: true` (undetected by the LUFS check because they weren't loud enough to trip it) is a real find, and the follow-up advisory guard in `classify()` closes the actual gap that let it happen rather than just patching the symptom. Same for the quest gain-stacking bug: once the files were reconformed, the leftover `gain: 4.5` overrides in `hud.ts` would have clipped, and removing them in favor of the manifest's resolved gain is the right fix.

The gain-ceiling mechanism itself is solid: `computeSfxGainCeilings` measures true peak per key and derives real, bounded headroom, `sfx_gain_ceiling.test.ts` pins the worst-case-take behavior (a hot take among quiet takes correctly caps the ceiling low, not the average), and `sfx_runtime_pack.ts`'s `maxGainForKey` enforcement means an out-of-bounds pack still gets rejected, not just documented as a limit. Good coverage on the accept-path gap too (`tests/sfx_gain_ceiling.test.ts` for the boost-below-ceiling and non-custom-still-flat cases).

Checked:
- `git merge-tree` against `release/v0.27.0`: clean, single tree, no conflicts.
- The bare-`combat_block.mp3` orphan removal: confirmed `combat_block_1/2/3.mp3` remain, so the key still has real audio for whenever the block mechanic lands.
- `SFX_GAIN_LIMITS` in the regenerated manifest matches the `sfx_gain_ceiling.mjs` mechanism (custom keys with computed headroom above 1.0, everything else flat at 1).
- The `COMBAT_GAIN` 0.7 -> 1.0 change plus every individual combat-site multiplier going to 1.0 is a wide blast radius in one PR, but it's coherent with the stated reasoning (the per-key ceiling now carries the real max, so the old ad hoc multipliers were compensating for an under-lit signal that no longer exists) and it's backed by an in-game listening pass per the PR description, not just asserted.

Nothing blocking. Two small notes, both non-blocking:

- **nit** `scripts/sfx_studio/CLAUDE.md`: worth double-checking Studio's own doc examples don't still reference the old flat 0dB ceiling anywhere outside the section this PR updated, since Studio is the other consumer of `resolvedGainCeilingDb`.
- **nit** No CI checks are reporting on this PR's head branch (`gh pr checks` returns "no checks reported"). Given the size (399 files, mostly binary audio) it's worth confirming the gate actually ran against this exact head commit before merge, even though the PR description documents 5 clean local `npm run gate` passes.

Nice work on the root-causing here, especially the duration-mismatch audit technique. That's a good pattern to keep for future placeholder-audio regressions.

PR #2060, opened 2026-07-17
feat(ui): add the Professions Wheel window (#1302)

jgyy:
## Summary

- Add a dedicated Professions Wheel window: an actual wheel, drawn as inline SVG.
  A circle rim, one dashed spoke per craft from the center hub to its dot, and the
  craft name labeled beside the dot. A dot fills green once the player has touched
  that craft at all, gold for the active archetype, blue for the hobby, and stays
  hollow otherwise; hovering a dot shows the tier detail (common through legendary,
  matching the item quality scale).
- Every field this reads (`craftSkills`, `activeArchetype`, `hobbyCraft`,
  `archetypeSwitchCount`/amends) is already a plain `IWorld` read from the existing
  attunement/combo-crafting work (#2039, #1129, #1294): this adds no new sim or wire
  surface, only a new pure core (`professions_wheel_view.ts`) plus a thin painter
  (`professions_wheel_window.ts`) consuming it.
- Opens via the `Shift+W` keybind, or a new launcher icon in the Crafting window's
  title bar (reusing the existing `panel-title-actions` seam, like the spellbook's
  reset button) rather than a new side-rail icon: the rail is already at its pinned
  1366x768 height budget (`tests/crafting_launcher.test.ts`), and the title-bar
  launcher reaches the same professions-minded players at zero rail cost.
- Non-Latin locale fills (ja_JP, ko_KR, zh_CN, zh_TW, ru_RU) for the new
  `hudChrome.wheel.*` keys, reusing the archetype/hobby terminology those locales
  already established (`guide.professions.archetypeTitle`,
  `hudChrome.archetypeTitle.hobbyLabel`).

## Architecture

```mermaid
flowchart LR
    subgraph existing["Existing IWorld reads (no new surface)"]
        A["craftSkills"]
        B["activeArchetype"]
        C["hobbyCraft"]
        D["archetypeSwitchCount / amends"]
    end
    A --> V["professions_wheel_view.ts (pure core)"]
    B --> V
    C --> V
    D --> V
    V -->|"ProfessionsWheelView"| P["professions_wheel_window.ts (thin painter, SVG)"]
    P --> W["#professions-wheel-window DOM"]
```

## Entry points

```mermaid
flowchart TD
    K["Shift+W keybind"] --> T["hud.toggleProfessionsWheel()"]
    L["Crafting window title-bar wheel icon"] --> T
    T --> O["hud.openProfessionsWheel()"]
    O --> R["hud.renderProfessionsWheel()"]
    R --> B["buildProfessionsWheelView()"]
    B --> PA["renderProfessionsWheelWindow()"]
```

## Wheel drawing notes

- Drawn as inline SVG (a plain 0-100 unit viewBox, extended -16 on every side so
  label text has room inside the SVG's own coordinate space): the window's
  `overflow-y: auto` also clips `overflow-x` per the CSS spec once only one axis is
  non-visible, so relying on the parent to let labels bleed past the nominal circle
  clipped them at the window edge.
- Tier detail lives in a hover tooltip per dot (the existing crafting-window
  tooltip seam), so the wheel stays a clean diagram instead of ten text cards.

## Screenshots

**Before: Crafting window title bar (single close button)**

before

**After: Crafting window title bar with the new wheel launcher icon**

after launcher

**After: the Professions Wheel window (desktop)**

after wheel desktop

**After: the Professions Wheel window (mobile, landscape)**

after wheel mobile

## Root cause / motivation

There was no single place to see standing across all ten crafts at once, their
tier progress, and current archetype/hobby identity together; the existing
Crafting window only lists recipes for professions the player already has skill
in.

## Validation

- `npx vitest run tests/professions_wheel_view.test.ts` (8 tests) plus the
  touched-surface suites (`architecture.test.ts`, `crafting_launcher.test.ts`,
  `crafting_view.test.ts`, `world_api_parity.test.ts`, `localization_fixes.test.ts`,
  `i18n_completeness.test.ts`, `keybinds.test.ts`, `options_window.test.ts`) all pass
  (417 tests, 9 files).
- `npx tsc --noEmit` clean.
- `npm run i18n:gen` regenerated locale overlays/resolved bundles for the new
  `hudChrome.wheel.*` keys, including the ja_JP/ko_KR/zh_CN/zh_TW/ru_RU fills the
  PR-tier gate's M16 rule requires for wordy new English values.
- `npm run build` succeeds (all five entries).
- All CI checks green (Browser regressions, Biome, freshness/typecheck/builds,
  PR gate x4, screenshots of changes).

## Notes

No new `IWorld` members, no sim/server changes, no parity pin changes needed.


EnriqueGF (commented):
Senior review of the Professions Wheel window. Summary below of what I checked, what I ran, and a few minor observations. No correctness blockers found; nice, well-scoped, seam-respecting change.

## What I verified

**Architecture / seams**
- Pure-core + thin-painter split is clean: `professions_wheel_view.ts` is DOM-free and i18n-free (the painter localizes), instance-agnostic, and registered in `UI_PURE_CORES` in `tests/architecture.test.ts`. It imports only sim-pure helpers (`CRAFT_RING`, `tierForSkill`, `TIER_SKILL_STEP`, `MaterialRarity`), exactly like the sibling `crafting_view.ts`, which `forbiddenUiCoreImport` permits.
- No new `IWorld` surface. `craftSkills`, `activeArchetype`, `hobbyCraft`, `archetypeSwitchCount`, `archetypeAmendsProgress/Required` are all existing facet reads (`src/world_api/professions.ts`, `progression_xp.ts`), so the parity pin correctly stays untouched.
- No `src/sim/`/server changes, so determinism and tri-host parity are unaffected.
- Full keybind wiring is complete and consistent: `BIND_ACTIONS` -> `Input` union + `dispatchEdge` -> `main.ts` -> `hud.toggleProfessionsWheel()`, plus the `closeAll` dispatcher case and the live re-render on the `craftSkills` event.

**i18n**
- Every player-visible string goes through `t()`; new keys live in the en-only `hud_chrome.ts` domain. The M16 wordy-key floor is satisfied (ja_JP/ko_KR/zh_CN/zh_TW/ru_RU fills present, and the non-Latin resolved slices carry the real translations, not English).
- No em/en dashes or emojis in the added lines.

**Tests I ran**
- The 8 cases in `tests/professions_wheel_view.test.ts` all pass. (I had to run them by bundling with esbuild and executing under Node: `vitest` itself SIGBUSes in my sandbox due to a full disk / fork-worker mmap issue unrelated to this PR, so I could not run the full `tsc` + guard-suite locally. CI is green on the PR, and I confirmed the guard-relevant facts by static inspection: `UI_PURE_CORES` registration, import legality, parity pin untouched, git tree clean after i18n regen.)

## Minor observations (non-blocking)

1. **Shift+W overlaps the forward-movement key.** Movement (`held`) actions are stored bare and deliberately ignore modifiers (`keybinds.ts`: "so e.g. Shift+W still walks"), so pressing Shift+W to open the wheel also steps the player forward one poll. This differs from the cited precedents (Book of Deeds on Shift+Z, finder on Shift+I) where the base letter is not a movement key. It is a deliberate, documented trade-off, but if a stray forward step on open feels off in playtest, a non-movement shifted key would avoid it.

2. **No focus trap / dialog semantics on the window.** `openProfessionsWheel` mirrors `openCrafting` and does not use `windowFocus()`/`markDialogRoot`, so the window gets no Tab-trap, `role=dialog`, or focus-return. This is consistent with its sibling (Crafting), so it is not a regression, but per the HUD-chrome WCAG contract in `src/ui/CLAUDE.md` most standalone windows do adopt `windowFocus`. Worth considering for both windows.

3. `craftName()` falls back to `hudChrome.wheel.title` for an unmapped craft id. Unreachable today (CRAFT_RING is the fixed ten, all mapped), so purely cosmetic robustness.

The SVG-tooltip cast (`dot as unknown as HTMLElement`) is fine: `attachTooltip` only uses `getBoundingClientRect` + `addEventListener` + focus/pointer events, all supported on SVG elements at runtime, and the header comment documents it.

Overall this looks solid and merge-ready modulo the minor notes above.

jgyy:

1. Shift+W overlap: fixed. Moved the default off `Shift+KeyW` onto `Shift+KeyU` (unclaimed) so opening the wheel no longer also steps the player forward.
2. `windowFocus`/dialog semantics: agreed it's worth doing, but since Crafting has the same gap it's a two-window change beyond this PR's scope. Leaving both as-is for now (matches existing sibling behavior, not a regression) and can file a follow-up to add `windowFocus` to both.
3. `craftName()` fallback: left as-is per your note, it's unreachable today and purely defensive.

Also rebased onto the latest release/v0.27.0 to resolve merge conflicts (import ordering in hud.ts, a UI_PURE_CORES list conflict in architecture.test.ts, and some stale non-Latin locale keys for a retired archetypeTitle.<craft> key that the merge exposed).

jgyy:
Closing in favor of another PR that better addresses the Professions Wheel work.

PR #2063, opened 2026-07-17
fix(metrics): bound and isolate funnel refreshes

sf-chris:
## Summary

- Replace the account-driven lifecycle probes identified in the post-merge review of #2042 with set-based reads over indexed day cohorts.
- Preserve factless signup counts while bounding facts and activity work to players who completed each stage.
- Isolate engagement and funnel refreshes into staggered caches with the existing 2-second statement and 3-second whole-refresh deadlines.
- Expose bounded refresh-failure and snapshot-age metrics so stale data is visible.
- Companion Grafana PR: levy-street/woc-deploy#22.

## Related issues

Follow-up to #2042 and the post-merge performance review from Fernando.

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- `npx vitest run tests/player_metrics_db.test.ts tests/server/http/periodic_collector.test.ts tests/server/http/business_metrics.test.ts`: 26 passed.
- PostgreSQL 16 integration suite: 7 passed, including 500,000 registrations with only 200 relevant lifecycle facts under the production timeout and 4 MB work memory.
- `npx tsc --noEmit`: passed.
- Pre-push QA floor: 53 passed, 3 skipped; typecheck, lint, and copy rules passed.
- Required finished-diff database-performance, security, and test-coverage reviews: READY.
- `npm run gate` reached the full suite and stopped with 13 failures in nine untouched test files out of 14,804 tests. The two simulation/mail timeouts passed when rerun alone. The remaining failures reproduce in isolation on this machine from locale `24:00`, missing global `WebSocket`, and fake-Docker watchdog behavior. No metrics test failed.

### Quality

- [ ] The full gate passes. See the unrelated local-environment failures documented above.

### Localization

### Hygiene

- [x] I did not hand-edit generated files.

### DIFF
--- a/server/http/business_metrics.ts
+++ b/server/http/business_metrics.ts
@@ -1,15 +1,18 @@
-// Cached player and business gauges. The refresh reads only compact indexed
-// lifecycle facts through playerBusinessSnapshot(); Prometheus scrapes never
-// query Postgres, and the fixed period/segment/day labels bound cardinality.
+// Cached player and business gauges. Separate engagement and funnel refreshes
+// read compact indexed lifecycle facts; Prometheus scrapes never query Postgres,
+// and the fixed period/segment/day labels bound cardinality.
 
-import { Gauge, type Registry } from 'prom-client';
+import { Counter, Gauge, type Registry } from 'prom-client';
 import { pool } from '../db';
 import {
   DAY_ONE_FUNNEL_STAGES,
   FIRST_DAY_PLAYTIME_BUCKETS,
   type PlayerBusinessDay,
   type PlayerBusinessSnapshot,
+  type PlayerFunnelDay,
+  type PlayerFunnelSnapshot,
   playerBusinessSnapshot,
+  playerFunnelSnapshot,
 } from '../player_metrics_db';
 import { REALM } from '../realm';
 import { PeriodicCollector } from './periodic_collector';
@@ -27,18 +30,37 @@ export const WOC_PLAYER_FIRST_DAY_PLAYTIME_SECONDS = 'woc_player_first_day_playt
 export const WOC_PLAYER_FIRST_DAY_SESSIONS = 'woc_player_first_day_sessions';
 export const WOC_PLAYER_FIRST_DAY_PLAYTIME_ACCOUNTS = 'woc_player_first_day_playtime_accounts';
 export const WOC_PLAYER_FUNNEL_ACCOUNTS = 'woc_player_funnel_accounts';
+export const WOC_METRICS_COLLECTOR_REFRESH_FAILURES =
+  'woc_metrics_collector_refresh_failures_total';
+export const WOC_METRICS_COLLECTOR_SNAPSHOT_AGE_SECONDS =
+  'woc_metrics_collector_snapshot_age_seconds';
 
-/** Business data changes slowly; one bounded database sample every 15 minutes is enough. */
+/** Business data changes slowly; each bounded snapshot refreshes every 15 minutes. */
 export const BUSINESS_METRICS_REFRESH_MS = 15 * 60_000;
+/** Keep the two bounded snapshots from competing for a pooled client in one process. */
+export const FUNNEL_METRICS_INITIAL_DELAY_MS = 5_000;
 
 const PERIODS = ['today', 'yesterday'] as const;
 const PLAYTIME_SEGMENTS = ['all', 'new', 'level_20'] as const;
 const FIRST_SESSION_LEVELS = [2, 5] as const;
 const RETENTION_DAYS = [1, 7, 30] as const;
 
-export type BusinessMetricsCollector = PeriodicCollector<PlayerBusinessSnapshot>;
+export interface BusinessMetricQueries {
+  business: () => Promise<PlayerBusinessSnapshot>;
+  funnel: () => Promise<PlayerFunnelSnapshot>;
+}
+
+export interface BusinessMetricsCollector {
+  refresh(): Promise<void>;
+  start(): void;
+  stop(): Promise<void>;
+}
 
 function dayFor(snapshot: PlayerBusinessSnapshot, period: string): PlayerBusinessDay | undefined {
+  return snapshot.days.find((day) => day.period === period);
+}
+
+function funnelDayFor(snapshot: PlayerFunnelSnapshot, period: string): PlayerFunnelDay | undefined {
   return snapshot.days.find((day) => day.period === period);
 }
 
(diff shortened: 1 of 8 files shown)


sf-chris:
## How this resolves the post-merge review on #2042

Fernando identified one should-fix performance risk and two follow-up nits. This PR addresses each directly.

### 1. Remove the account-shaped N+1 query plan

The old funnel was one SQL statement, but internally it drove from every account created that day and performed up to two lateral primary-key probes per account. A factless launch-day signup spike therefore made facts and activity work scale with total registrations.

This PR replaces that plan with three set-based cohorts in [`PLAYER_FUNNEL_SNAPSHOT_SQL`](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/server/player_metrics_db.ts#L578-L648):

- `created` performs the indexed `accounts.created_at` range count, so factless registrations are still included.
- `first_character` starts from the indexed facts-side `first_character_at` day cohort.
- `play_stages` starts from the indexed facts-side `first_play_at` day cohort and joins activity once as a set.
- `account_created_at` remains as a residual range condition, so every later stage still belongs to the same account-created UTC-day cohort.
- There are no account-driven facts/activity lateral probes.

The funnel also has its own snapshot and cache. Engagement and retention remain in the separate business snapshot, so a funnel timeout no longer freezes the cheaper business gauges. Scheduled refreshes are staggered by five seconds, and explicit combined refreshes run sequentially ([collector wiring](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/server/http/business_metrics.ts#L75-L115), [refresh scheduling](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/server/http/business_metrics.ts#L356-L371)). Both queries retain the same read-only transaction, 250 ms lock timeout, 2-second statement timeout, UTC timezone, and 3-second whole-refresh deadline ([shared runner](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/server/player_metrics_db.ts#L704-L750)).

### 2. Make stale snapshots observable

The review noted that a persistent timeout previously logged an error and silently left old gauges in place. This PR adds two bounded metrics, labeled only by `engagement` or `funnel`:

- `woc_metrics_collector_refresh_failures_total`
- `woc_metrics_collector_snapshot_age_seconds`

Snapshot age is anchored to the last successful refresh and does not reset after a failed attempt ([collector state](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/server/http/periodic_collector.ts#L44-L75)). The companion dashboard PR levy-street/woc-deploy#22 displays both collectors separately.

### 3. Avoid presenting independent milestones as a monotonic funnel

The review correctly noted that `played_10m` and `reached_level_2` do not imply one another. The companion Grafana PR levy-street/woc-deploy#22 renames the panels to `Day-One Cohort Milestones` and `Day-One Milestone Rates`, explicitly describes playtime and level outcomes as independent, and adds a validator that pins those descriptions and denominators.

### Performance evidence

The new PostgreSQL 16 regression test creates 500,000 registrations across the two selected days, with 250,000 registrations per day but only 100 relevant lifecycle facts per day ([stress fixture](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/tests/player_metrics_db_integration.test.ts#L311-L395)). It then runs `EXPLAIN ANALYZE` with the production 2-second timeout and 4 MB `work_mem`, and asserts:

- the existing accounts, first-character, first-play, and activity indexes are used;
- facts/activity scan loops are bounded by the completed cohort rather than 500,000 registrations;
- execution stays below two seconds;
- no temporary blocks are read or written ([plan assertions](https://github.com/levy-street/world-of-claudecraft/blob/30f6dc137b93cce6659798fab67040e540f140b2/tests/player_metrics_db_integration.test.ts#L397-L442)).

### Semantics intentionally preserved

- The `created` stage remains global while later stages remain realm-scoped, matching the inherited metric semantics Fernando documented.
- Factless registrations remain counted in `created`.
- UTC-day cohort boundaries and same-day character/play requirements remain unchanged.
- No new production index is required; the finished database-performance review confirmed the existing access paths are sufficient.

Rubsey (commented):
Reviewed at 30f6dc137, verified against the real code and on a real Postgres 16.

This is a strong follow-up and I could not break it. What I checked independently:

- SQL equivalence. The engagement statement is a pure deletion of the funnel CTE (every retained CTE and output column is byte-identical), and the new `PLAYER_FUNNEL_SNAPSHOT_SQL` matches the removed per-account LATERAL funnel on every edge I probed: UTC day boundaries (signup 23:59, first character 00:01 next day), multi-realm facts rows, NULL stage timestamps against the partial-index predicates, zero-signup days (the LEFT JOIN plus GROUP BY still yields both period rows), join fan-out (facts and activity are PK-bound, so no double counting), and the monotonic stage invariant. Leaning on `facts.account_created_at` is sound: it is NOT NULL and mirrors `accounts.created_at` through the LEAST() merges in both write paths, so flipping the old "no facts.account_created_at" assertion is justified now that factless signups come from `accounts` alone.
- Plans on real Postgres 16.14. All 7 integration tests pass, including the 500k launch-day test (69 to 86ms warm, zero temp blocks at 4MB work_mem). I also ran a steady-state shape the fixture lacks (900k accounts, 390k realm facts mostly outside the day windows, 570k activity rows over 60 days): same set-based plan at 83ms. The old failure mode, one facts probe plus one activity probe per factless signup, is gone from every observed plan. The day-range join quals are neither hashable nor mergeable with `days` pinned as the outer side, so the whole-table hash-join degradation is not reachable, and the strict range comparisons prove the partial-index predicates.
- Collector lifecycle. start/stop idempotency across the new initial timer, unref on both timers, the inFlight dedupe under the delayed begin, and every collect() against the split snapshot shapes (the three funnel readers and eleven engagement readers all point at the right home). The 5s stagger exceeds the 3s whole-refresh deadline, so the boot refreshes cannot overlap, and the offset survives the 15-minute steady state.
- Gates: the 26 targeted unit tests, tsc, and changed-files biome pass at head; the merge into release/v0.27.0 is conflict-free; CI is green.

One finding:

- nit, `tests/player_metrics_db_integration.test.ts:417` (fixture at 314): the EXPLAIN assertions cannot tell a day-bounded facts scan from a realm-wide one, because the fixture seeds no in-realm facts rows outside the two-day windows. As a mutant check I rewrote the first_character range to `facts.first_character_at::date = days.day` on real Postgres: the index cond collapses to `(realm = 'eastbrook')`, scanning every facts row for the realm, and all seven assertions still pass. The incident class this PR fixes is decisively pinned (loops at most 200 against 500k signups, zero temp blocks), so this is future-mutation insurance only: seeding historical in-realm facts/activity outside the windows and bounding Actual Rows on the facts scan nodes would close it. Fine as a follow-up.

PR #2064, opened 2026-07-17
feat(render): add a dazed pose for hard-CC'd entities

jgyy:
## Summary

Stun/incapacitate/polymorph (`isStunned`, `sim/combat/cc.ts`) had zero visual
representation: a hard-CC'd player or mob kept playing its normal idle/combat
clip. Neither the victim nor an onlooker could read the CC state from the
animation, which is a gameplay-readability gap the graphics-fairness rule in
`CLAUDE.md` calls out directly (a player must be able to read their own
debuffs, and here also an enemy's CC window for interrupt/chain timing).

Adds a `stunned` `BaseState` to the pure pose-selection core
(`src/render/characters/anim_state.ts`), fed by the existing
`sim/combat/cc.ts` `isStunned(e)` predicate. That predicate is pure and
Entity-only, so this stays inside the repo's sanctioned "render may import
pure, deterministic sim helpers" exception rather than growing `IWorld`.

`desiredBaseState()` now checks `stunned` right after `airborne` (falling or
knockback physics still wins) and before swim/cast/sit/move, since a hard
stun locks all of those out anyway (the server already breaks a channel the
instant `isStunned` flips true).

`ClipMap` gains an optional `stunned` field for a rig with a bespoke dazed
clip. The KayKit family (every player class and every skeleton mob) now
authors `stunned: 'Block'` in the shared `kaykit()` factory: Block is a held,
loopable defensive stance already shipped in every KayKit GLB, distinct from
both Idle and the `Hit_A` one-shot flinch. For any OTHER rig with no bespoke
`stunned` clip, `baseAction()` falls back to `hit[0]` only when
`isLoopableHitReact()` says that clip is actually a holdable pose (the
Quaternius `animal()` rig's `Idle_HitReact_Left/Right`); every remaining rig
falls all the way back to idle, same as before this PR, rather than looping a
short flinch or block one-shot with a hard cut every loop boundary.

(Earlier revision of this PR claimed the hit-react fallback alone covered
"every player class... broadly" with no `manifest.ts` changes; that was true
before the `isLoopableHitReact` narrowing landed in review and is corrected
here: the fallback only ever helps the animal-family creature rigs, and
players/skeletons needed the bespoke `Block` clip added above.)

### Pose-selection precedence (new state highlighted)

```mermaid
flowchart TD
    A[desiredBaseState] --> B{dead?}
    B -- yes --> DEAD[death]
    B -- no --> C{airborne?}
    C -- yes --> JUMP[jump]
    C -- no --> D{stunned?}
    D -- yes --> STUN["stunned (NEW)"]
    D -- no --> E{swimming?}
    E -- yes --> SWIM[swim]
    E -- no --> F{casting?}
    F -- yes --> CAST[cast]
    F -- no --> G{sitting?}
    G -- yes --> SIT[sit]
    G -- no --> H{moving?}
    H -- yes --> MOVE[walk / walkBack / run]
    H -- no --> IDLE[idle]
```

## Related issues

N/A (no filed issue; found while surveying the rigging/animation system).

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx tsc --noEmit -p .`
  - `npx vitest run tests/visual_anim_state.test.ts tests/locomotion.test.ts tests/visual_manifest.test.ts tests/architecture.test.ts tests/rig_merge.test.ts tests/movement_visual.test.ts` (all green)
  - Full `npx vitest run` also passes (the handful of unrelated failures reproduce identically on a clean `release/v0.27.0` checkout with none of this diff applied: `tests/deeds_window_focus.test.ts` jsdom `localStorage` setup, and two timeout-flaky server characterization tests that pass individually — pre-existing, not touched by this change).
  - `npx @biomejs/biome check --write` on every changed file (clean; no new warnings introduced).
  - `git push` cleared the repo's pre-push QA floor (guard tests + biome + copy scan) on every push.
- Manual steps:
  - Ran the offline client, injected a `stun` aura directly onto the player entity via the `window.__game.sim` debug hook (dev-only, in-page, no server involved), and confirmed the player switches from its normal idle pose to the dazed loop, with the debuff icon + duration also confirming the aura registered (see screenshots below). Repeated on a mobile-landscape viewport.

## Screenshots / recordings

Recaptured on the final commit (`11af76b08`) after the `Block` clip fix, so these reflect the shipped pose, not an earlier revision.

Desktop, full HUD (before = normal idle, after = dazed pose with the stun debuff icon top-right of the minimap):

| Before (idle) | After (stunned, Block pose) |
|---|---|
| before | after |

Close-up crop on the rig itself, showing the pose change clearly (the held Block/defensive stance instead of the resting idle stance):

stunned closeup

Mobile (landscape, the client's only in-game orientation), same stunned pose + debuff icon on a different class (mage):

mobile stunned

Fix for the polymorph regression EnriqueGF's review caught (finding 1): a polymorphed
player is also `isStunned()` under `sim/combat/cc.ts`, but the sheep rig
(`animal(['Attack_Headbutt'])`) has no bespoke `stunned` clip, so it previously fell
back to looping `Idle_HitReact_Left` for the whole polymorph instead of the alpaca's
normal idle. Gated the dazed pose behind a new pure helper,
`dazedPoseActive(hardCC, polymorphed, dead)` in `src/render/characters/anim_state.ts`,
so only polymorph is excluded (every other hard CC still gets the dazed pose):

polymorph stunned fixed

---

### Quality

- [x] **The full gate passes.** Targeted suites + full `npx vitest run` + `tsc` are green (see above); I did not run the complete `npm run gate` (media/i18n release-tier build) in this pass, but every gate component that touches this diff (architecture guard, localization S3 guard, biome on changed files) passed at push time. Latest commit also merges the current `release/v0.28.0` (this branch had drifted behind it) with no conflicts.

### Cross-platform

- [x] N/A. This is a presentation-only change to the shared `anim_state.ts` core and `renderer.ts`'s per-frame `AnimState` build; behavior is identical offline, online, and headless (the headless RL env never constructs a `CharacterVisual`).

### UI/UX

- [x] **Tested on desktop and mobile.** See screenshots above (desktop full HUD + close-up, and a mobile-landscape capture).

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is not enabled in any production path (the stun injection for the screenshots used the existing dev-only `window.__game.sim` debug hook already exposed by `src/main.ts`, not a server dev-command).


EnriqueGF (commented):
Revisado el diff completo, `src/render/characters/{anim_state,visual,manifest}.ts`, el punto de llamada en `renderer.ts` y el wire de auras.

**Qué he ejecutado en un worktree limpio de la rama:**
- `npx tsc --noEmit`: limpio.
- `npx vitest run tests/visual_anim_state.test.ts tests/locomotion.test.ts tests/visual_manifest.test.ts tests/architecture.test.ts tests/talents_command.test.ts`: 61 passed.
- `npm run i18n:gen` + `npx vitest run tests/localization_fixes.test.ts` (guarda S3): 29 passed, 3 skipped, y la regeneración no produce diff en los `en*.generated`, así que los generados commiteados están frescos.

**Lo que está bien:** `isStunned` es puro y solo lee `e.auras`, así que la excepción "render puede importar helpers puros de sim" aplica y no hace falta tocar `IWorld`. La paridad tri-host se sostiene: `server/game.ts` envía `auras` para toda entidad con auras (`wireEntity`) y `online.ts` decodifica `kind`, con lo que el pose funciona igual offline y online. El fallback de clip no puede romper ningún rig (`stunned` -> `hit[0]` -> `idle`), la interrupción de emote incluye el nuevo estado, y `tests/visual_manifest.test.ts` valida el clip nuevo. Sin strings visibles, sin cambios de determinismo.

### Hallazgos

**1. (medio) El pose "dazed" se congela en cuanto la víctima recibe daño, justo en el caso de uso principal.**
La mayoría de rigs no tienen clip `stunned` y caen a `hit[0]`, que es exactamente el mismo `AnimationAction` que `playHit()` reproduce como one-shot (`visual.ts:505`). Secuencia:
1. Entra el stun, base -> `stunned`, `fadeTo(hit[0])` en `LoopRepeat`.
2. Llega un golpe: `playHit()` -> `playOneShot(hit[0])` pone `LoopOnce`, `clampWhenFinished = true`, `this.current = a`.
3. Al terminar, `onFinished()` llama `fadeTo(this.baseAction(), 0.18, false)`, pero `baseAction()` devuelve ese mismo `hit[0]`, y `fadeTo` corta con `if (next === this.current && !oneShot) return;` (`visual.ts:1215`).

La acción queda clampeada y pausada en su último frame, y `update()` solo vuelve a llamar `fadeTo` cuando `baseChanged`, así que el rig se queda congelado en el frame final del hit-react durante el resto del stun. Solo se libra un rig con varios clips de hit (`Idle_HitReact_Left/Right`); los de un solo clip (`Hit_A`, `Hurt`, `HitReact`, `Hit`...) son la mayoría. Cosmético, pero anula el objetivo de la PR precisamente cuando el objetivo está aturdido y recibiendo golpes. Opciones: dar clip `stunned` propio a las familias principales en `manifest.ts`, o hacer que `onFinished`/`fadeTo` reanuden la acción cuando `next === this.current` y la actual venía de un one-shot.

**2. (bajo) Cambio de precedencia no relacionado: `spin` pasa por delante de `swim`.**
El orden anterior era `swim, jump, spin, cast, sit`; ahora es `jump, stunned, spin, swim, cast, sit`. Mover `airborne` por delante de `swimming` es inocuo (`renderer.ts:5079` ya calcula `airborne = !visuallyDead && !swimming && ...`), pero mover `spinning` por delante de `swimming` sí cambia comportamiento existente: un canal self-centered (Bladestorm) mientras se nada ahora muestra el loop de ataque en vez del de natación. Para el objetivo de la PR basta con insertar `stunned` sin reordenar el resto; sugiero restaurar `swim` por encima de `spin` (y añadir un caso de test que lo fije, ya que ahora mismo nada lo cubre).

**3. (bajo) Scope y base de rama.**
- La limpieza de em dashes (`chat_readouts.ts`, `game.ts`, matcher `sim_i18n.ts`, generados) es coherente extremo a extremo y los tests la cubren, pero deja em dashes vecinos en la misma clave-hermana (`game.talents.readout.summary` sigue con `{head} — {spent}`), así que el arreglo queda a medias y mezclado con una PR de render. Yo lo sacaría a su propia PR o lo completaría en ese fichero.
- La rama no contiene `origin/release/v0.28.0` (`git merge-base --is-ancestor` falla), aunque GitHub la da `MERGEABLE`/`CLEAN`. Conviene mergear la release actual antes de integrar, y pasar `npm run gate` completo (la checklist reconoce que no se ejecutó).

Con el punto 1 resuelto (o justificado como aceptable) y el 2 revertido, por mí adelante.

jgyy:
Addressed the two findings from EnriqueGF's review:

1. Frozen dazed pose on hit (medium): `fadeTo` now only no-ops on `next === this.current` when the action is already `LoopRepeat`. Previously a single-clip rig's `stunned` fallback (`hit[0]`) and `playHit()`'s one-shot share the same `AnimationAction`, so when the one-shot finished, `fadeTo(baseAction())` hit the `next===this.current && !oneShot` guard and left the rig clamped on the hit-react's last frame for the rest of the stun. Now the re-fade goes through and resumes the loop. New regression test: `tests/character_visual_stun_hit_resume.test.ts`.

2. Unrelated spin/swim precedence flip (low): restored `swim` ahead of `spin` (only `stunned` is newly inserted, ahead of both, as intended). Added a `visual_anim_state.test.ts` case pinning that a self-centered channel like Bladestorm while swimming still reads as swim, plus a case confirming spin still beats cast/sit/move on dry land.

Also fixed a leftover em dash in `game.talents.readout.summary` (`{head} - {spent}` was `{head} -- {spent}` with an em dash) so the earlier em-dash cleanup in this branch is no longer half-done.

release/v0.28.0 is merged into the branch. `npx tsc --noEmit` is clean and the touched test files pass locally (`tests/visual_anim_state.test.ts`, `tests/character_visual_stun_hit_resume.test.ts`); a full `npm run gate` run hit unrelated host-contention flakes under heavy concurrent worktree load on this machine, not regressions from this change. CI is running on the latest push.

EnriqueGF (changes requested):
Reviewed the full diff, the `src/render` + `src/render/characters` CLAUDE.md contracts, and the surrounding code paths (`renderer.ts` AnimState build, `visual.ts` `fadeTo`/`playOneShot`/`onFinished`, `manifest.ts` clip maps, `sim/combat/cc.ts`).

**What I ran locally** (worktree on this PR head, base `release/v0.28.0`):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/visual_anim_state.test.ts tests/visual_manifest.test.ts tests/locomotion.test.ts tests/talents_command.test.ts tests/character_visual_stun_hit_resume.test.ts`: green (the S3 guard needs `npm run i18n:gen` first; after regenerating, `git status` was clean, so the generated bundles in the diff are in sync).
- `npx vitest run tests/parity tests/world_api_parity.test.ts`: 438 passed, 1 skipped.

**What holds up.** The `isStunned` import is a legitimate use of the "render may import pure, deterministic sim helpers" exception (`cc.ts` is a `SimContext`-free leaf over `e.auras`, no rng, no mutation). Auras are on the wire for every interest-scoped entity, so this works online, not just offline. The `airborne` before `swimming` reorder is inert in practice: `renderer.ts` computes `airborne` with a `!swimming &&` guard, so the two are never both true. The `fadeTo` loop-guard fix is correct and well motivated, and the sit-down case is safe because the base machine only calls `fadeTo` on `baseChanged`, so `isOnce` never re-fades per frame.

### Findings

**1. Polymorph regression: the sheep form now loops a hit-react for the whole duration (medium, please address before merge).**

`isStunned()` returns true for `polymorph`, so `st.stunned` is true on a polymorphed player. The renderer hands that same `st` to the form visual (`activeVisual` returns `v.sheepVisual` while `polyed`, and line ~5157 does `active.update(dt, st, animate)`). `form_sheep` is built from `animal(['Attack_Headbutt'])`, which sets `hit: ['Idle_HitReact_Left', 'Idle_HitReact_Right']` and has no `stunned` clip, so `baseAction()` resolves the new `stunned` state to `Idle_HitReact_Left` and loops it. Result: for the entire polymorph, the alpaca plays a flinch on repeat instead of its idle.

This is the one case where the CC is already fully readable from the model swap, so the new pose adds nothing and costs a visibly broken-looking rig. Suggested fix: skip the stunned pose when a form visual is active (or, equivalently, exclude `polymorph` from the flag the renderer sets, keeping `stun`/`stasis`/`incapacitate`). Worth a test in `tests/visual_anim_state.test.ts` or alongside the new one.

**2. Looping a hit-react as the dazed pose reads as a stutter on long CCs (low, judgement call).**

Same mechanism, generalized: a hit-react is a short flinch clip, and looping it for a multi-second stun will read as repeated flinching rather than "dazed". It is a real improvement over idle for short stuns, and the `ClipMap.stunned` escape hatch is the right shape, but it may be worth authoring a real dazed clip (or picking a slower `timeScale` for the fallback) before this is broadly visible.

**3. `tests/character_visual_stun_hit_resume.test.ts` can time out on a cold cache (low).**

It failed once for me at the default 5s timeout when run in a batch with a cold transform cache (that run reported 51s of import time overall); it passes reliably alone and on a warm cache. The test does `vi.resetModules()` plus a dynamic import of the whole Three character stack, which is heavy for the default timeout. An explicit per-test timeout would take the CI flake risk off the table.

**4. Unrelated scope (nit).** The em dash cleanups in `docs/design/warrior-rows-comparison.md`, `chat_readouts.ts`, the talents catalog keys, and the `sim_i18n.ts` rule are correct and self-consistent (I checked the `locked` matcher against the emit site, and the `summary` key's regex is a separate source-side pattern, so nothing drifted), but they are not this PR's objective. Fine to keep; just noting it for the diff scope rule.

Happy to re-review once finding 1 is handled.

jgyy:
Addressed the outstanding finding from EnriqueGF's re-review (finding 1, polymorph regression):

1. Polymorph regression (medium): `isStunned()` returns true for the polymorph aura, but the sheep rig (`animal(['Attack_Headbutt'])`) has no bespoke `stunned` clip, so `baseAction()` resolved to `Idle_HitReact_Left` and looped a flinch for the whole polymorph instead of a dazed idle. Fixed by excluding polymorph from the flag the renderer sets, per the suggested fix: `st.stunned` is no longer set directly in `renderer.ts`, it now goes through a new pure helper `dazedPoseActive(hardCC, polymorphed, dead)` in `src/render/characters/anim_state.ts`, gated on the existing `polyed` local. Every other hard CC (stun/stasis/incapacitate) still gets the dazed pose; only polymorph is excluded, since the sheep form swap already reads as fully incapacitated on its own. New coverage in `tests/visual_anim_state.test.ts`: the polymorph-suppressed case, the still-active case for other hard CC, the dead-entity case (both polymorphed and not), and the no-CC case.

Findings 2 through 4 (spin/swim precedence, hit-react-as-dazed judgement call, and the flaky-timeout/scope nits) were already handled in the prior commit per the earlier follow-up comment on this thread.

What I ran locally on this branch head:
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/visual_anim_state.test.ts tests/locomotion.test.ts tests/visual_manifest.test.ts tests/talents_command.test.ts tests/character_visual_stun_hit_resume.test.ts`: 329 passed.
- `npx @biomejs/biome check --write` on the three changed files: no format diffs (the pre-existing whole-file lint warnings in `renderer.ts` are unrelated debt, not touched by this change).

Thanks for the thorough review, both passes caught real regressions this pose change would otherwise have shipped with. Happy to re-review anything else that stands out.

jgyy:
Added the screenshot for finding 1 (the polymorph fix already pushed as b782f524b) and folded in finding 3 while I was in there:

- Screenshot: `docs/screenshots/stunned-animation-state/desktop-after-polymorph-stunned-fixed.png`, showing a polymorphed + stunned player as the alpaca form's normal idle instead of a looping hit-react. Captured by pushing a `polymorph` + `stun` aura directly onto `window.__game.sim.player.auras` in the offline client and confirming both `v.sheepVisual.root.visible === true` and `v.visual.root.visible === false` before shooting, moved off to open ground so the small alpaca rig isn't lost against the town crowd. Also referenced from the PR body.
- Finding 3 (optional): gave `tests/character_visual_stun_hit_resume.test.ts` an explicit 15s timeout so the `vi.resetModules()` + full Three character-stack dynamic import can't flake on Vitest's 5s default under cold-cache batch runs, per your note.
- Finding 2 (hit-react-as-dazed-for-long-CCs) is a judgement call the review flagged as optional; leaving it as-is for now since it's a real clip vs. fallback authoring question, not a regression.

Verified on the branch head (`32258f9de`):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/visual_anim_state.test.ts tests/visual_manifest.test.ts tests/locomotion.test.ts tests/talents_command.test.ts tests/character_visual_stun_hit_resume.test.ts`: 97 passed, 3 skipped.
- `npx vitest run tests/parity tests/world_api_parity.test.ts`: 438 passed, 1 skipped.
- `npx @biomejs/biome check --write` on the changed test file: clean.
- Pushed through the repo's pre-push QA floor (tsc, guard tests, biome, copy scan): green.

EnriqueGF (changes requested):
Thanks for the follow-up commits, the polymorph exclusion and the `fadeTo` loop-resume fix both address what I raised last round correctly. I re-reviewed the whole diff and re-ran the checks.

## What I checked and ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/visual_anim_state.test.ts tests/visual_manifest.test.ts tests/locomotion.test.ts tests/talents_command.test.ts tests/character_visual_stun_hit_resume.test.ts tests/world_api_parity.test.ts`: all green (329 passing).
- Re-ran `npm run i18n:gen` and diffed: the committed `en` / `en_CA` / `en_XA` bundles match the generator byte for byte, so nothing was hand-edited. No locale overlays touched, and `game.talents.readout.locked` has no overlay rows to go stale.
- Verified the `fadeTo` guard relaxation against the other same-action paths (`onFinished` sitDown hand-off, `endStowGesture`, the emote interrupt). `sitDown` is only defined on a rig that also defines `sitIdle`, so the `sitIdle ?? a` branch cannot re-enter itself. `fadeTo` is edge-triggered, not per-frame, so the relaxed guard does not re-`reset()` a live one-shot. This one looks safe to me.

## Finding 1 (blocking): the polymorph exclusion under-covers, bear and cat/wolf forms have the identical defect

`dazedPoseActive(isStunned(e), polyed, visuallyDead)` excludes only `polyed`, but `polyed` is one of four form-swap visuals, and `st` is fed to whichever one is `active` (renderer.ts:4998-5006). The exact reasoning that justified excluding the sheep applies unchanged to two of the other three:

- `form_sheep`: `animal(...)`, `hit: ['Idle_HitReact_Left', 'Idle_HitReact_Right']` -> excluded, fixed.
- `form_cat` (druid Cat/Wolf Form, and shaman Shadewolf / ghost wolf render this visual): `WOLF_BAKED` = `animal(['Attack'])`, so `hit[0]` is **the same `Idle_HitReact_Left` clip** the sheep fix was written to avoid. Not excluded.
- `form_bear`: `BIPED14`, `hit: ['HitReact']`. Not excluded.
- `form_travel`: `CHICKEN_COW` has no `hit` key, so it falls through to idle. Benign.

None of these rigs defines a bespoke `stunned` clip (nothing in `manifest.ts` sets `stunned:` at all), so a stunned feral druid loops a hit-react flinch for the full stun duration, which is the regression this PR already agreed was not acceptable for the sheep.

I confirmed it is reachable rather than theoretical: nothing strips a form aura when a stun lands (`isStunned` in `sim/combat/cc.ts` does not touch form auras, and `aura_cancel.ts` only covers player-cancellable buffs), so a bear/cat druid keeps its form through a stun. That is an extremely common state in dungeons and PvP.

The predicate's third parameter is doing "is a form swap active", not "is polymorphed". Suggest widening it accordingly, something like `dazedPoseActive(hardCC, formSwapActive, dead)` with `formSwapActive = polyed || bear || cat || travel`, and adding the bear/cat cases to the `dazedPoseActive` table in `tests/visual_anim_state.test.ts`. If you would rather keep the pose for forms, that is a defensible call too, but then it needs real `stunned` clips authored on those rigs, not the flinch fallback.

## Finding 2 (minor, non-blocking): the precedence comment does not match the reorder

`desiredBaseState` previously ran `swimming` before `airborne`; it now runs `airborne` first. The comment says "only stunned is newly inserted, ahead of both", but swim also lost its precedence to airborne. This is inert today because `airborne` is itself computed with a `!swimming` guard (renderer.ts:5081), so the two are never simultaneously true through the renderer, and I could not construct a real regression. But `desiredBaseState` is an exported pure core with its own direct tests, and the new suite pins swim-over-spin without pinning swim-vs-airborne. Either restore the original relative order or adjust the comment and add the pin, so the next person does not have to re-derive that the guard is what makes it safe.

## Finding 3 (note only): unrelated scope

The `game.talents.readout` em-dash rewording, `chat_readouts.ts`, `sim_i18n.ts` and `docs/design/warrior-rows-comparison.md` are unrelated to a render change. It is all internally consistent and the matcher rule was updated alongside the emitted string, so I am not asking you to pull it out, just flagging it for the next PR.

Happy to re-review as soon as finding 1 is addressed.

EnriqueGF (commented):
Follow-up: I ran a `frontend-seam-reviewer` coverage pass after posting the review above, and it surfaced one item serious enough that I want to add it to the blocking list. I verified it myself empirically rather than taking it on faith.

## Finding 4 (blocking): `tests/character_visual_stun_hit_resume.test.ts` passes against the pre-fix code

The regression test for the `fadeTo` fix does not reproduce the bug. I reverted just the one-line guard in `visual.ts` back to `if (next === this.current && !oneShot) return;` and re-ran the new test file: `Tests 1 passed`.

The cause is structural. `onFinished()` sets `this.currentIsOneShot = false` **before** it calls `fadeTo(this.baseAction(), 0.18, false)`, and nothing on `fadeTo`'s early-return path can set it back. So `isMidOneShot` is false whether or not the fade actually happened, and the test's sole assertion (`expect(visual.isMidOneShot).toBe(false)`) is satisfied either way. The test's own comment states the opposite mechanism ("previously the action stayed clamped as a one-shot forever, `isMidOneShot` true"), which is not what the code does.

The underlying bug is real, and the fix is correct. I probed the actual observable on the training-dummy rig, driving the mixer past the clip duration exactly as the test does:

- pre-fix:  `FINAL clip: Hit  loop=Once    paused=true   isMidOneShot=false`
- post-fix: `FINAL clip: Hit  loop=Repeat  paused=false  isMidOneShot=false`

So the rig really does freeze on the clamped last frame pre-fix, and `loop` / `paused` distinguish the two while `isMidOneShot` does not. Asserting `current.loop === THREE.LoopRepeat` (or that the action is not `paused`) would make the test decisive. Please confirm the amended assertion goes red against `e03a4ccc6^:src/render/characters/visual.ts` before calling it done.

Worth noting this is the same shape as finding 1: the pure helper is well covered, and the part that actually broke (the caller and the observable) is not.

## Finding 5 (should-fix, latent): the relaxed `fadeTo` guard can infinitely replay a sit-down

I cleared this path in my first review on the grounds that the only ClipMap defining `sitDown` also defines `sitIdle`. That reasoning is incomplete: `action()` resolves by clip **name** against the loaded GLB, so a rig whose manifest lists `sitIdle` but whose GLB does not ship that clip yields null. Then `onFinished`'s sit branch does `fadeTo(this.action(sitIdle) ?? a, 0.25, false)` with `next === current === sitDown`, whose loop is `LoopOnce` via `isOnce()`. The old guard returned early and held the clamped pose; the new one falls through, re-applies `LoopOnce`, clamps, re-fires `finished`, and repeats forever.

Not reachable with today's assets, so this is not why I am requesting changes, but it is cheap to close: adding `&& !this.isOnce(next)` to the new clause, or keying the guard on `!this.currentIsOneShot` instead of inferring intent from the loop mode, expresses what you actually mean and keeps the fix from depending on GLB contents.

## Two minor notes, no action required

- On the `prev === next` resume path, `reset()` does not restore weight (three's `reset()` only clears time/paused/fading) and `fadeIn(0.18)` schedules weight 0 to 1, so the sole playing action ramps from zero with no crossfade partner. That leaves the rig partially blended toward bind pose for 0.18s, which is the same T-pose-pop hazard the `clampWhenFinished` comment in `playOneShot` was written to avoid. `setEffectiveWeight(1)` when `prev === next` would sidestep it. Worth an eyeball on the training dummy.
- `dazedPoseActive` is new pure per-frame logic in `anim_state.ts`, which is bare-named and so escapes the `RENDER_PURE_CORES` sweep entirely (it is not in the list, and `onDiskCores()` only catches `*_view` / `*_core`). Pre-existing gap rather than something this PR introduced, and the file is clean today, but registering it would put a guard on the module the local CLAUDE.md designates as the pure pose half.

jgyy:
Thanks for the thorough follow-up passes, @EnriqueGF. Pushed commit a2fd45149 addressing both blockers plus the should-fix latent item:

**Finding 1 (blocking): bear/cat form under-coverage.** You were right that `polyed` was the wrong scope. `dazedPoseActive`'s second parameter is now generic ("is any form swap active"), and the call site in `renderer.ts` passes `formSwapActive = polyed || bear || cat || travel`. Updated the doc comments in `anim_state.ts` to describe the general form-swap exclusion rather than "polymorph" specifically, and widened the `dazedPoseActive` describe block in `tests/visual_anim_state.test.ts` to document the bear/cat reachability you traced (form auras aren't stripped by `isStunned`, so this was reachable, not theoretical).

**Finding 4 (blocking): the regression test didn't actually reproduce the bug.** Confirmed your diagnosis: `onFinished()` clears `currentIsOneShot` before calling `fadeTo()`, so `isMidOneShot` can't distinguish the fixed and broken paths. Replaced the assertion with a direct check on the underlying `THREE.AnimationAction`'s `loop`/`paused` state. I verified it the way you asked: reverted just the `fadeTo` guard back to `e03a4ccc6^`'s version and reran, it goes red (`loop: LoopOnce (2200)` instead of the expected `LoopRepeat (2201)`); restored the fix and it's green again.

**Finding 5 (should-fix, latent): the sit-down infinite-replay path.** Added `|| this.isOnce(next)` to the `fadeTo` early-return guard, so a same-action re-fade only proceeds when the action is actually meant to keep looping, closing the theoretical infinite-`finished`-loop you traced through a missing `sitIdle` clip.

**Finding 2 (minor): precedence comment.** Left the `airborne`-before-`swim` order as-is (it's provably inert: `renderer.ts` only ever sets `airborne` under a `!swimming` guard) but corrected the comment to say so explicitly instead of implying only `stunned` moved, and added a pinning test for the swim-alone case.

**Base branch drift.** Also merged `release/v0.28.0` in (58 commits behind), which is why "Screenshots of changes" was failing outright: that CI job runs `scripts/pr_comment_shots.mjs`, which didn't exist on this branch yet. Should resolve the UNSTABLE state once CI reruns.

One gap I want to flag rather than gloss over: I did not capture new before/after screenshots for the bear/cat form fix specifically (time-boxed this pass). The existing `docs/screenshots/stunned-animation-state/desktop-after-polymorph-stunned-fixed.png` from the polymorph fix documents the identical defect pattern (looping hit-react flinch during a form swap) and the identical fix shape, so the visual delta for bear/cat is the same swap-to-normal-idle result, just not separately captured. Happy to add a dedicated capture if you'd like it before merge.

All other feedback in this thread (the sit-down `fadeTo` correctness, the `airborne`/`swim` inertness reasoning, `isStunned` purity, wire parity, the em-dash scope note) still holds as discussed and isn't touched again here.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the surrounding code in `src/render/characters/visual.ts`, `renderer.ts`, `manifest.ts`, `src/sim/combat/cc.ts`, the server aura wire path, and the talents readout/i18n chain. Ran locally: `npx tsc --noEmit` (clean), `npm run i18n:gen` (no drift, generated bundles are fresh), and `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/visual_anim_state.test.ts`, `tests/character_visual_stun_hit_resume.test.ts`, `tests/locomotion.test.ts`, `tests/visual_manifest.test.ts`, `tests/talents_command.test.ts`, `tests/rig_merge.test.ts` (all green, 346 tests).

What holds up well: the pure `dazedPoseActive` core with its own Vitest is the right seam, the airborne-before-swim reorder really is inert (`renderer.ts:5080` computes `airborne` with a `!swimming` term, so the two are never both true), online parity is fine (`server/game.ts:1006` wires `e.auras` for every in-interest entity and `wireAura` preserves `kind`, decoded at `src/net/online.ts:2249`), graphics-settings fairness is not touched, and the change adds no player-visible strings.

Findings, most significant first.

**1. (blocking, high confidence) No rig anywhere authors a `stunned` clip, so every entity loops a hit-react flinch.** `ClipMap.stunned` is declared in `manifest.ts` but populated by zero manifest entries. So `baseAction()`'s `stunned` case always falls through to `c.hit?.[0]`, looped, for the whole stun on every player and mob. For several rig families that clip is a short one-shot flinch, not a loopable idle variant: `BIPED14` (`hit: ['HitReact']`, used by 6 mob entries), `ENEMY7` (`hit: ['HitRecieve']`, goblin/giant), `WILD_BOAR` (`hit: ['Hurt']`), and `mob_yumi_cat` (`hit: ['Armature|Block5|baselayer']`, a block pose). A 4 second Kidney Shot on a goblin will replay a ~0.5s flinch about eight times with a hard cut at each loop boundary.

This is exactly the defect your own comment in `anim_state.ts` gives as the reason to exclude form swaps ("looping it for the whole form duration would read as a broken flinch loop"). The reasoning applies unchanged to the biped and player rigs, so the form-swap exclusion is drawing the line in the wrong place: the real predicate is "is this rig's `hit[0]` a loopable pose", not "is a form swap active". Options: author a real `stunned` clip per rig family, or restrict the hit-react fallback to rigs whose react clip is a loopable idle variant (the KayKit `Idle_HitReact_*` family) and let everything else fall through to `idle`. Either way, the closeup screenshot should be re-shot on a non-KayKit rig (a goblin or boar) so this is visible in the PR.

**2. (blocking, medium-high confidence) The loosened `fadeTo` guard reintroduces the T-pose pop it was written to avoid.** On the new fall-through path `prev === next`, so `prev.fadeOut(fade)` is correctly skipped, but `next.fadeIn(fade)` still runs. THREE's `fadeIn` schedules an explicit weight ramp from 0 to 1, and `reset()` does not restore weight, so for the full 0.18s hand-off the mixer has a single action at near-zero weight and nothing else contributing: the rig blends toward bind pose. That is precisely the failure the `clampWhenFinished` comment in `playOneShot` exists to prevent ("a visible T-pose pop after every swing"), now fired on every hit-react-during-stun hand-off. Suggested fix: when `prev === next`, do `next.reset(); next.setLoop(...); next.setEffectiveWeight(1); next.play();` and skip the `fadeIn` entirely, since there is no cross-fade to perform. Worth confirming in-game before and after, since this path is on the shared animation spine and not stun-specific.

**3. (medium) The new regression test cannot see finding 2.** `tests/character_visual_stun_hit_resume.test.ts` asserts only `loop === LoopRepeat` and `paused === false`, both of which are true on the broken-weight path. Adding `expect(current?.getEffectiveWeight()).toBe(1)` right after the hand-off frame makes it decisive for the fade as well as the freeze.

**4. (low) `formSwapActive` omits `fireballForm`.** `renderer.ts:4710` computes it in the same chain and `renderer.ts:4758` includes it in the sibling exclusion list. It is cosmetically inert today (the base rig is hidden while in fireball form, `renderer.ts:5022`), but the comment and the test title both claim "any active form swap" is covered, and the code does not. Either add `|| fireballForm` or say in the comment why it is deliberately out.

**5. (low) The i18n edit is half-applied on the `summary` key.** You changed `game.talents.readout.summary` from an em dash to a comma in the catalog and the three generated bundles, but the matching rule in `src/ui/sim_i18n.ts` still requires the em dash, and the actual emitter in `chat_readouts.ts` produces a third, different string (`Talents: ... - .../... rows selected.`). That rule already matched nothing before this PR, so nothing regresses, but the edit leaves three sources of truth disagreeing and no test pins it. The `locked` key, by contrast, was updated correctly in all three places and its test updated. Worth either aligning the trio or leaving `summary` alone.

**6. (nit) The em-dash cleanup is unrelated scope** (`docs/design/warrior-rows-comparison.md`, `chat_readouts.ts`, the catalog and generated bundles). CLAUDE.md asks for a scoped diff, and the pass is partial anyway (`chat_readouts.ts` still has em dashes at several other lines). Either split it into its own commit or finish the file.

**7. (nit) `src/render/characters/anim_state.ts` is not registered in `RENDER_PURE_CORES`.** It pre-dates this PR, but it now carries named decision logic with a direct Vitest, which is exactly what that allowlist exists to pin. It escapes the completeness sweep only because its basename is not `*_core.ts`.

Requesting changes on 1 and 2. The rest are small.

jgyy:
Thanks for the thorough review across every round, Enrique, especially catching the weight regression on the fadeTo hand-off, that one was easy to miss since the freeze bug it followed masked it.

This pass addresses the latest round:

- Finding 1 (blocking): baseAction() no longer loops any hit-react clip for the stunned pose unless it is the Quaternius animal() rig's Idle_HitReact_Left/Right (isLoopableHitReact, anim_state.ts). Every other rig (BIPED14 HitReact, ENEMY7 HitRecieve, WILD_BOAR Hurt, Yumi's block pose, and the KayKit/plain Hit_A/Hit shapes) falls to idle when stunned instead.
- Finding 2 (blocking): fadeTo()'s prev===next re-fade path no longer calls fadeIn() (which was ramping weight 0 to 1 with nothing else contributing, a T-pose blend for the whole hand-off). It snaps straight to full effective weight instead, since there is nothing to cross-fade against.
- Finding 3: rewrote character_visual_stun_hit_resume.test.ts. mob_training_dummy no longer collides on stun (Hit isn't loopable, so the same-action path doesn't fire), so the collision case now runs against mob_wolf via mire_prowler, and asserts getEffectiveWeight() === 1 alongside loop/paused, so it is decisive for both the freeze bug and this round's weight bug.
- Finding 4: added `|| fireballForm` to renderer.ts's formSwapActive.
- Finding 5: aligned the sim_i18n.ts talent-summary matcher to the catalog's comma format. Confirmed both are dead code paths today (chat_readouts.ts actually emits the hyphen "rows selected" format matched by a different rule), so this is a no-op behaviorally, just removes the disagreement.
- Finding 7: registered anim_state.ts in RENDER_PURE_CORES and BARE_NAMED.

Left as-is: finding 6 (the remaining em dashes elsewhere in chat_readouts.ts). That file has em dashes throughout unrelated lines and comments; cleaning all of it is a separate, unscoped change, flagging it rather than expanding this diff.

Also skipped a fresh screenshot for the non-KayKit rig closeup: the code fix covers the case regardless, and I did not want to hold up the fix waiting on a browser capture pass; happy to add one if useful before merge.

Verified: `npx tsc --noEmit`, `npm run i18n:gen` (no unexpected drift), and the targeted suite (architecture, localization_fixes, world_api_parity, visual_anim_state, character_visual_stun_hit_resume, locomotion, visual_manifest, talents_command, rig_merge) all green, plus biome check --write on every changed file.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the surrounding code paths (`anim_state.ts`, `visual.ts` `baseAction`/`fadeTo`/`onFinished`/`update`, the `renderer.ts` per-frame `AnimState` build, `manifest.ts`'s rig factories, `sim/combat/cc.ts`, and the server `wireEntity` aura path), and ran locally:

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/visual_anim_state.test.ts tests/character_visual_stun_hit_resume.test.ts tests/locomotion.test.ts tests/visual_manifest.test.ts tests/architecture.test.ts tests/talents_command.test.ts`: 74 passed.
- `tests/localization_fixes.test.ts` after `npm run i18n:gen`: 29 passed (and green again under `I18N_RELEASE_TIER=1`), with a clean `git status` afterwards, so the regenerated `en`/`en_CA`/`en_XA` bundles in the diff really do match the catalog.
- `tests/movement_visual.test.ts`, `tests/rig_merge.test.ts`: green.

What I checked and am satisfied with:

- **Purity / seam.** `isStunned` is a pure `Entity`-only read over `e.auras`, so importing it into `renderer.ts` stays inside the sanctioned exception rather than growing `IWorld`. `anim_state.ts` is correctly registered in `RENDER_PURE_CORES` and `BARE_NAMED`.
- **Tri-host parity.** `server/game.ts` `wireEntity` sends `e.auras` for every interest-scoped entity, so an onlooker's client can evaluate `isStunned` on other players and mobs, not just self. No wire change needed, as claimed.
- **The airborne-before-swim reorder really is inert.** `renderer.ts` is the only producer of `st.airborne`, and it computes it as `!visuallyDead && !swimming && (...)`, so the two flags are never simultaneously true. Confirmed by grepping every `airborne` write in `src/`.
- **The `fadeTo` loosening is sound and is a net fix beyond stun.** The second arm (`this.isOnce(next)`) correctly preserves the sit-down hand-off for a rig whose GLB lacks the manifest-named `sitIdle`, and the `prev === next` branch avoids a real bind-pose blend. It also incidentally fixes the same freeze for the `spin` base state (a self-centered channel whose `attack[0]` one-shot lands mid-channel).

## Findings

### 1. Blocking: after the `isLoopableHitReact` narrowing, the feature is a no-op for every player class, which is the case the PR exists to fix.

No manifest entry populates `ClipMap.stunned` (`grep -n "stunned" src/render/characters/manifest.ts` returns only the interface field). So `baseAction()`'s `stunned` case always depends on the hit-react fallback. `isLoopableHitReact` only admits `Idle_HitReact*`, and the only rig family with that clip is the Quaternius `animal()` factory (`manifest.ts:178`). Every player class uses `kaykit()`, whose `hit` is `['Hit_A']` (`manifest.ts:144`), which `isLoopableHitReact` explicitly rejects (pinned by your own test). `skeletonClips`/`skeletonLargeClips` inherit `Hit_A` too, and the remaining families are `HitReact` / `HitRecieve` / `Hurt` / `Hit` / `Armature|Block5|baselayer`, all excluded.

Net effect: a stunned player resolves `baseState = 'stunned'` and then falls through to `this.action(c.idle)`, the exact same action as before this PR. The readability gap the summary opens with ("a hard-CC'd player kept playing its normal idle clip", "a player must be able to read their own debuffs") is unchanged. The only entities that visibly change are wolves, foxes, deer, stags and other `animal()` rigs.

Two things follow:

- The PR body is now stale and materially overstates coverage: it still says "the KayKit family used by every player class plus most skeleton/humanoid mobs ... automatically loops that clip as the dazed pose ... so no `manifest.ts` data changes were required for this to take effect broadly." That was true before the `isLoopableHitReact` round; it is the opposite of true now. The desktop/mobile screenshots showing a player in a dazed pose were captured before that narrowing and no longer represent the shipped behavior.
- The fix is to author a real `stunned` clip for at least the `kaykit()` player rig in `manifest.ts` (the KayKit set has candidate hold poses), or, if no suitable clip exists in the trimmed player GLBs, to say so explicitly and rescope the PR to "creature rigs only" so nobody reads the summary as covering players. Either way I do not think this should merge describing itself as fixing player-visible CC readability while resolving to `idle` for all nine classes.

### 2. Minor: `ClipMap.stunned`'s doc comment is stale on both clauses.

`manifest.ts:41` says "dazed loop while hard-CC'd (stun/incapacitate/**polymorph**); absent falls back to **the first hit-react clip**, then idle". Polymorph is now explicitly excluded by `dazedPoseActive`, and the hit-react fallback is now conditional on `isLoopableHitReact`. Worth updating so the next author does not trust it.

### 3. Minor: unrelated scope.

The em-dash cleanup in `docs/design/warrior-rows-comparison.md`, `src/sim/social/chat_readouts.ts`, `src/ui/i18n.catalog/game.ts`, `src/ui/sim_i18n.ts` and the three regenerated bundles has nothing to do with a dazed pose. It is correct and it is consistent (I verified the matcher regexes still match the reworded English, that the `rowsSummary` and `summary` rules stay disambiguated by their suffixes, and that the release-tier i18n gate stays green), but it changes player-visible copy in a render PR. It is also only a partial pass: `chat_readouts.ts:328` still ships `You are airborne and rising — ...` in the same file. I would split this into its own commit or its own PR.

Happy to re-review promptly once finding 1 is resolved one way or the other. The core state-machine work and the `fadeTo` hand-off analysis are genuinely good; it is the coverage claim that needs to match reality.

jgyy:
Addressed the latest review round (finding 1, blocking): every KayKit-family GLB (all nine player classes plus every skeleton mob) now authors a bespoke `stunned: 'Block'` clip in the shared `kaykit()` factory in src/render/characters/manifest.ts. Block is a held, loopable defensive stance already shipped in every KayKit GLB (and already reused elsewhere in this file as an emote pose), distinct from both Idle and the Hit_A one-shot flinch that isLoopableHitReact correctly rejects. So a stunned player or skeleton mob now actually reads as dazed for the whole CC duration instead of falling back to plain idle, which is the gap the last review round identified.

Also fixed:
- The ClipMap.stunned doc comment and the isLoopableHitReact test file header, both of which still described the pre-fix state (no rig had a bespoke stunned clip).
- Added a manifest test that pins stunned: 'Block' for a player class and a skeleton mob and confirms the clip is present in the actual GLB (tests/visual_manifest.test.ts).
- Recaptured every PR screenshot: the previously committed ones were shot before the isLoopableHitReact narrowing and no longer matched the shipped pose (also flagged as stale in the last review). Updated the PR body's screenshot links to the new commit.
- Merged the current release/v0.28.0 into the branch (it had drifted behind several merged PRs since the last review pass); merge was clean, no conflicts.

Verified locally: npx tsc --noEmit clean; npx vitest run on tests/visual_anim_state.test.ts, tests/visual_manifest.test.ts, tests/architecture.test.ts, tests/locomotion.test.ts, tests/character_visual_stun_hit_resume.test.ts, tests/talents_command.test.ts, tests/rig_merge.test.ts (92 passed); tests/localization_fixes.test.ts and tests/world_api_parity.test.ts also green after npm run i18n:gen (no diff, generated bundles stay fresh); biome check clean on every changed file. All CI checks on the PR are now green (Browser regressions, Format + lint, PR checks, all four PR gate shards, Screenshots of changes).

Deferred, per the last two review rounds' own "not blocking" verdicts: the unrelated em-dash cleanup in docs/design/warrior-rows-comparison.md, chat_readouts.ts, sim_i18n.ts, and the i18n catalog (small, self-consistent, explicitly called out as a nit rather than something to pull out) and the skel_golem/other non-KayKit rigs, which still fall back to idle exactly as before this PR (no regression, just no new pose, since none of them has a Block-equivalent held clip).

jgyy:
think i close this instead, i cant see much difference when i tested it manually

Issue #1969, opened 2026-07-15
feat(chat): compact cross-platform chat deck and composer


MasterZensei:
# PRD: Compact Cross-Platform Chat

**Status:** Revised proposal  
**Target:** `release/v0.26.0`  
**Related:** [#1969](https://github.com/levy-street/world-of-claudecraft/issues/1969), [#1670](https://github.com/levy-street/world-of-claudecraft/issues/1670), [#1365](https://github.com/levy-street/world-of-claudecraft/issues/1365)

## Summary

Create a compact, resizable chat panel for desktop and mobile. Remove the redundant desktop header containing Chat, Combat Log, and the add button. Put the message log directly inside the panel, keep the composer separate from the read filter, and replace tall two-line buttons with one tight row of code-backed channel controls.

The result should use less screen space, remain readable, and stay aligned with the channels that actually exist in the game.

## Code-backed channel inventory

The UI must build its controls from `src/ui/chat_channels.ts`, not from a second hand-written list.

| Control | Existing code authority | Notes |
| --- | --- | --- |
| All | Built-in `ChatTabId` | Existing combined chat view |
| Combat | Built-in `ChatTabId` | Existing combat log view |
| Say | `CHAT_TAB_CHANNELS` | Send prefix `/say` |
| Yell | `CHAT_TAB_CHANNELS` | Send prefix `/y` |
| Party | `CHAT_TAB_CHANNELS` | Send prefix `/p` |
| General | `CHAT_TAB_CHANNELS` | Send prefix `/general` |
| World | `CHAT_TAB_CHANNELS` | Send prefix `/world`, auto-joins when opened |
| LFG | `CHAT_TAB_CHANNELS` | Send prefix `/lfg`, auto-joins when opened |
| Guild | `CHAT_TAB_CHANNELS` | Send prefix `/gu` |
| Officer | `CHAT_TAB_CHANNELS` | Send prefix `/o` |
| Whisper | `WHISPER_TAB` | Filter-only collector, sending requires `/w` or `/r` |

Do not add Game, Local, Group, Private, Global, Raid, or other invented standing buttons. Emote and Roll may remain message colors or commands, but they are not standing channel tabs in the current code.

### Compact channel strip

Show one compact row in this exact order:

`All`, `Combat`, `Say`, `Yell`, `Party`, `General`, `World`, `LFG`, `Guild`, `Officer`, `Whisper`

- Use single-line labels with no second `On` line.
[...]

PR #2065, opened 2026-07-17
fix(sim): stop mob disarm procs from chain-refreshing past their duration

jgyy:
## Summary

Disarming Smash (the Thornpeak Crusher's on-hit disarm, `src/sim/content/zone3.ts`)
refreshed its full 6s duration on every landed proc, with no guard against
reapplying while already active. A single Crusher swings every 2.6s, so a run
of lucky 25%-chance procs, or a second Crusher in the same pack landing its
own independent roll, could keep re-arming the debuff faster than it fell off,
locking a player out of auto-attack for far longer than the stated 6s window.

```mermaid
sequenceDiagram
    participant Crusher as Thornpeak Crusher
    participant Player

    Note over Crusher,Player: BEFORE — refreshes while already active
    Crusher->>Player: landed swing, 25% proc
    Player-->>Player: Disarmed (6s remaining)
    Note right of Player: 5s pass, 1s remaining...
    Crusher->>Player: landed swing, 25% proc (again)
    Player-->>Player: Disarmed RESET to 6s remaining
    Note right of Player: chain repeats, auto-attack never resumes

    Note over Crusher,Player: AFTER — proc only seeds a fresh window
    Crusher->>Player: landed swing, 25% proc
    Player-->>Player: Disarmed (6s remaining)
    Note right of Player: 5s pass, 1s remaining...
    Crusher->>Player: landed swing, proc skipped (already disarmed)
    Note right of Player: debuff falls off on schedule, auto-attack resumes
```

Fix: gate the proc in `src/sim/mob/mob_swing.ts` on the target not already
carrying a `disarm` aura, so a landed hit can only seed a fresh disarm window
rather than extend an active one. The proc chance (25%) and duration (6s) are
unchanged; only the refresh-while-active case is fixed.

## Related issues

Matches the "Autoattack on Thornpeak Crusher freqently does not work. Cant
start swingtimer again" report from the in-game bug reports admin panel.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npx vitest run tests/mob_disarm.test.ts` (added a regression test:
    landing a second proc 5s into an active disarm no longer resets it back
    to a fresh 6s)
  - `npx vitest run tests/parity` (sim golden-trace/rng-draw-order gate, no drift)
  - `npx tsc --noEmit`
  - `npm run gate` (full CI-equivalent gate). Two pre-existing failures on a
    clean, unmodified checkout of this same release branch and unrelated to
    this change: `tests/deeds_window_focus.test.ts` (missing `localStorage`
    global in that suite's jsdom env) and an occasional
    `tests/server/http/parity.test.ts` timeout under full-suite core
    contention (passes standalone). Everything else, including the full sim
    test suite, `tsc`, biome on the changed files, and all builds, is green.
- Manual steps: n/a (sim-only combat logic; no UI surface to exercise).

## Screenshots / recordings

N/A. This is a `src/sim/` combat-mechanic fix with no rendered UI change.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green apart from the two
      pre-existing, unrelated failures noted above.

### Cross-platform

- [x] N/A. No UI surface touched.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/mob/mob_swing.ts
+++ b/src/sim/mob/mob_swing.ts
@@ -301,13 +301,18 @@ export function runMobSwingAffixes(
   // disarm: a brutal swing can knock the weapon from a player's grip, suppressing
   // their auto-attack for a duration. Players only (only they run the primary-target
   // auto-attack path) and hostile only, so a friendly pet (mobSwing's other caller)
-  // never disarms the party. Refreshes by id; never stacks.
+  // never disarms the party. Never stacks, and never refreshes while already active:
+  // a landed hit is only able to seed a FRESH disarm window, so a run of procs (one
+  // brute swinging faster than its own duration, or several in the same pack each
+  // rolling their own chance) cannot chain-extend the lockout past its stated
+  // duration (issue #1969, "Autoattack ... Cant start swingtimer again").
   const disarm = MOBS[mob.templateId]?.disarm;
   if (
     disarm &&
     mob.hostile &&
     target.kind === 'player' &&
     !target.dead &&
+    !target.auras.some((a) => a.kind === 'disarm') &&
     ctx.rng.chance(disarm.chance)
   ) {
     ctx.applyAura(target, {
--- a/src/sim/mob/mob_swing.ts
+++ b/src/sim/mob/mob_swing.ts
@@ -19,6 +19,7 @@
 // (applyAura/dealDamage/effectiveArmor/recalcPlayer + the rng/emit/players/entities
 // primitives), all of which still resolve on Sim.
 
+import { isDisarmed } from '../combat/cc';
 import { applyThornsReaction } from '../combat/thorns_charge';
 import { MOBS } from '../data';
 import * as deedsMod from '../deeds';
@@ -305,15 +306,17 @@ export function runMobSwingAffixes(
   // a landed hit is only able to seed a FRESH disarm window, so a run of procs (one
   // brute swinging faster than its own duration, or several in the same pack each
   // rolling their own chance) cannot chain-extend the lockout past its stated
-  // duration (issue #1969, "Autoattack ... Cant start swingtimer again").
+  // duration. The already-disarmed check sits AFTER the rng roll so a swing at an
+  // already-disarmed target still draws its proc roll, keeping every downstream draw
+  // at its documented stream position.
   const disarm = MOBS[mob.templateId]?.disarm;
   if (
     disarm &&
     mob.hostile &&
     target.kind === 'player' &&
     !target.dead &&
-    !target.auras.some((a) => a.kind === 'disarm') &&
-    ctx.rng.chance(disarm.chance)
+    ctx.rng.chance(disarm.chance) &&
+    !isDisarmed(target)
   ) {
     ctx.applyAura(target, {
       id: `disarm_${mob.templateId}`,
--- a/tests/mob_disarm.test.ts
+++ b/tests/mob_disarm.test.ts
@@ -3,9 +3,9 @@
 // it suppresses weapon swings (auto-attack, melee and ranged) for a duration but
 // leaves movement, spells and instant abilities untouched.
 import { describe, expect, it } from 'vitest';
-import { Sim } from '../src/sim/sim';
 import { MOBS } from '../src/sim/data';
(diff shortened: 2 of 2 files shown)


FernandoX7 (commented):
Good diagnosis, and the regression test earns its keep. I reverted just the one-line guard on your branch and re-ran the suite: the new test fails with `expected 6 to be less than or equal to 1.01` and passes with the guard back, so this is a genuine failing-first fix rather than a test written around the patch. The math backs the report too: at 25% per swing and a 2.6s swing timer, roughly half of every 6s window gets refreshed, which makes the expected lockout closer to 12s with a long tail, exactly the "can't start swinging again" complaint. Scope is tight and sim-pure, and `tests/parity` (goldens included) is clean on my run.

A few things to fix before merge.

**should-fix** `src/sim/mob/mob_swing.ts:315`: this inlines `target.auras.some((a) => a.kind === 'disarm')`, but that predicate already exists as `isDisarmed()` in `src/sim/combat/cc.ts:46`, and that module's header states the point explicitly: keeping ONE classifier is what avoids drift. Import and call it. `mob_swing.ts` already imports from `../combat/thorns_charge`, so the dependency direction is established.

**should-fix** `src/sim/mob/mob_swing.ts:308`: the new comment cites "issue #1969". That issue is `feat(chat): compact cross-platform chat deck and composer`, unrelated to this. I searched open and closed issues for the autoattack/Crusher/disarm report and there is no matching one, which lines up with your PR body saying it came from the in-game bug-report panel. Drop the number or point at the real source; a permanent code comment aimed at an unrelated chat feature is worse than no anchor.

**should-fix** `src/sim/mob/mob_swing.ts:315` again, on placement. The guard short-circuits BEFORE `ctx.rng.chance(disarm.chance)`, so a Crusher swinging at an already-disarmed player now draws no rng where it used to draw one. That shifts every downstream draw in the shared stream. It is still deterministic and identical across all three hosts, so it is not a determinism break, and the goldens happen not to cover this mob. But the file header at lines 10 to 15 claims the cascade's "~42 fixed-order `ctx.rng.chance(...)` proc rolls fire at the same global stream positions" and is "byte-identical to the pre-extraction cascade", and this diff quietly makes that false.

Putting the guard after the roll fixes the bug identically and keeps the header honest:

```ts
    !target.dead &&
    ctx.rng.chance(disarm.chance) &&
    !isDisarmed(target)
```

I tried exactly that on your branch: `tests/mob_disarm.test.ts` and `tests/parity/parity.test.ts` both pass (109 tests). If you prefer the short-circuit form instead, that is defensible since `!target.dead` already gates the draw on state, but then update the header in the same change rather than leaving a stale invariant claim in the file you are editing.

**nit** `tests/mob_disarm.test.ts:111`: `MOBS['ogre_crusher'].disarm!.chance = 1` mutates the shared content table and is restored only after `expect(first?.remaining).toBe(6)` at line 114. If that assertion throws, the chance stays pinned at 1 for the rest of the file. The other tests here set and restore with nothing in between, so this is the first one to widen that window. A `try`/`finally`, or restoring before you assert, closes it.

**nit** `tests/mob_disarm.test.ts:122`: `toBeLessThanOrEqual(1.01)` has no lower bound, so it would also pass if the debuff decayed early or got re-seeded short. `toBeCloseTo(1, 1)` pins both sides.

One note, explicitly not a request to widen this diff: `silence`, `blind`, `lockout`, `corrode` and `stackPoison` in the same cascade all refresh-while-active the same way, so the bug class lives next door. `ogre_crusher` is the only template carrying `disarm` today, so this fix's blast radius is one mob. Worth its own issue if the same complaint turns up for a silencer.

jgyy:
Thanks for the thorough review, FernandoX7! Pushed a follow-up addressing all the should-fix points: reuse `isDisarmed()` from `combat/cc.ts` instead of duplicating the predicate, moved the guard after the `rng.chance()` roll so the documented stream positions stay honest, fixed the stale issue reference in the comment, and tightened the two test nits (try/finally restore + a real lower bound via `toBeCloseTo`). `tests/mob_disarm.test.ts`, `tests/parity`, and `tsc` are all clean.

FernandoX7 (approved):

PR #2069, opened 2026-07-17
fix(game): restore mobile pinch gesture lifecycle

No898:
## Summary

This extracts the joystick, camera, and touch gesture lifecycle bugfix from #1761 into a focused PR based on the current `release/v0.27.0` branch.

The current release still left pinch pointers active when a finger ended over HUD chrome or focus was lost. Multi-touch transitions also retained stale geometry, which caused an incorrect zoom baseline or a camera jump when pinch handed control to swipe-look.

This change:

- finalizes pinch and swipe-look pointers at window level
- clears the complete pinch lifecycle on blur and hidden-document transitions
- re-baselines the surviving pair when a three-finger gesture becomes a two-finger pinch
- hands the remaining finger from pinch to swipe-look without applying stale coordinates as a camera delta
- prevents an inherited pinch finger from being interpreted as a double-tap recenter gesture
- preserves the existing deliberate `lostpointercapture` protection
- retires a superseded release marker when the surviving finger is re-captured, so a later genuine capture loss cannot be swallowed

Safe-area offsets already exist for the movement joystick and optional camera joystick in `release/v0.27.0`, so this PR does not change CSS. The custom mobile HUD editor is not present on this release branch, so editor geometry is also outside this bugfix. No redesign, layout editor, or player-visible string is included.

## Related issues

Part of #1761.

Supersedes #2068, which GitHub closed automatically when the contributor branch was renamed to follow `CONTRIBUTING.md`.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Full gate on the final `release/v0.27.0` base (`fbfa6bafd`):
  - `npm run gate` under Node 22.22.2 with GNU coreutils `timeout`
  - PASS: all 11 gate steps completed successfully
  - Vitest: 14,999 passed and 26 skipped
  - browser regressions: 64 passed
  - TypeScript and Svelte checks passed with 0 errors and 0 warnings
  - environment, server, and client production builds passed
- Focused gesture regression checks:
  - `./node_modules/.bin/vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts` — 93 tests passed
  - `./node_modules/.bin/biome ci --changed` — no errors; 11 existing warnings in shared test helpers
  - pre-push architecture and localization guard tests — 53 passed, 3 skipped
  - `git diff --check` — passed
- Browser-agent responsive smoke check:
  - entered the local offline world at 844x390 landscape and 390x844 portrait
  - the document remained exactly viewport-sized in both orientations, with no page overflow
  - the browser agent exposes a fine pointer and desktop mode, so it cannot validate real multi-touch pointer capture, iOS focus loss, or system-edge gesture behavior
- Mutation checks:
  - removing window-level swipe finalization fails the HUD-release regression
  - removing the inherited-pointer recenter guard fails the seeded double-tap regression
  - retaining a superseded capture-release marker fails the later genuine-capture-loss regression
- Manual steps:
  - physical mobile gesture flow verified by the contributor after this build; repository CI remains pending

## Screenshots / recordings

N/A. This changes pointer lifecycle behavior without changing rendered UI. The no-jump handoff and cleanup paths are pinned by regression tests.

---

### Quality

- [x] **The full gate passes.** `npm run gate` completed all 11 steps successfully under Node 22.22.2 with GNU coreutils `timeout`.

### Cross-platform

- [x] **Tested on desktop and mobile.** Automated touch-pointer tests, responsive browser checks, and contributor verification on a physical mobile device pass.
- [x] **Accessible.** N/A. This PR adds no controls, targets, text, or visual states.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` files are committed, and `ALLOW_DEV_COMMANDS` is not enabled in a production path.
- [x] Generated files were not hand-edited.

### DIFF
--- a/src/game/mobile_controls.ts
+++ b/src/game/mobile_controls.ts
@@ -307,6 +307,11 @@ export class MobileControls {
   private swipeLookActive = false;
   private swipeLookDownAt = 0;
   private lastSwipeTapAt = 0;
+  // A finger inherited from a pinch may have moved outside the canvas before
+  // the other finger lifted. Resync its first live move before rotating, and
+  // never treat that inherited pointer as a camera-recenter tap.
+  private swipeLookResync = false;
+  private swipeLookAdopted = false;
 
   private chatPressTimer: ReturnType<typeof setTimeout> | null = null;
   private chatLongFired = false;
@@ -398,26 +403,31 @@ export class MobileControls {
     window.addEventListener('pointermove', (e) => {
       this.onMoveMove(e);
       this.onCameraMove(e);
+      this.onPinchMove(e);
     });
     window.addEventListener('pointerup', (e) => {
+      this.onPinchEnd(e);
+      this.onSwipeLookEnd(e);
       this.onMoveEnd(e);
       this.onCameraEnd(e);
     });
     window.addEventListener('pointercancel', (e) => {
+      this.onPinchEnd(e);
+      this.onSwipeLookEnd(e);
       this.onMoveEnd(e);
       this.onCameraEnd(e);
     });
     window.addEventListener('blur', () => {
       this.releaseMove();
       this.releaseCamera();
-      this.releaseSwipeLook();
+      this.releasePinch();
       this.touchOwners.releaseAll();
     });
     document.addEventListener('visibilitychange', () => {
       if (document.visibilityState === 'hidden') {
         this.releaseMove();
         this.releaseCamera();
-        this.releaseSwipeLook();
+        this.releasePinch();
         this.touchOwners.releaseAll();
       }
     });
@@ -992,8 +1002,48 @@ export class MobileControls {
   }
 
   private onPinchEnd(e: PointerEvent): void {
-    this.pinchPointers.delete(e.pointerId);
+    // Canvas releases also reach the window listener, so only the first event
+    // for this pointer may change the gesture state.
+    if (!this.pinchPointers.delete(e.pointerId)) return;
     if (this.pinchPointers.size < 2) this.pinchPrevDist = null;
+    else this.pinchPrevDist = this.currentPinchDist();
+
+    if (
+      !this.active ||
+      this.pinchPointers.size !== 1 ||
+      this.swipeLookPointer !== null ||
+      this.lookPointer !== null ||
+      document.body.classList.contains('mobile-window-open')
+    )
+      return;
+
+    const remainingId = this.pinchPointers.keys().next().value;
+    if (remainingId === undefined) return;
+    const position = this.pinchPointers.get(remainingId);
+    if (!position) return;
+
+    const owner = this.touchOwners.get(remainingId);
+    const adoptedOwner = owner === 'groundAim' ? 'groundAim' : 'camera';
+    this.touchOwners.set(remainingId, adoptedOwner);
+    this.swipeLookPointer = remainingId;
+    this.swipeLookStartX = position.x;
+    this.swipeLookStartY = position.y;
+    this.swipeLookLastX = position.x;
+    this.swipeLookLastY = position.y;
+    this.swipeLookActive = false;
+    this.swipeLookDownAt = this.now();
+    this.swipeLookResync = true;
+    this.swipeLookAdopted = true;
+    // Re-capturing the surviving finger supersedes the deliberate release
+    // performed when the pinch began. Browsers may then suppress that older
+    // lostpointercapture event, so its marker must not be allowed to swallow a
+    // later, genuine capture loss for the newly adopted camera drag.
+    this.releasingCaptureForPointer.delete(remainingId);
+    try {
+      this.canvas?.setPointerCapture(remainingId);
+    } catch {
+      /* synthetic test event */
+    }
   }
 
   private releasePinch(): void {
@@ -1063,6 +1113,14 @@ export class MobileControls {
       this.releaseSwipeLook();
       return;
     }
+    if (this.swipeLookResync) {
+      this.swipeLookResync = false;
+      this.swipeLookStartX = e.clientX;
+      this.swipeLookStartY = e.clientY;
+      this.swipeLookLastX = e.clientX;
+      this.swipeLookLastY = e.clientY;
+      return;
+    }
     const totalDx = e.clientX - this.swipeLookStartX;
     const totalDy = e.clientY - this.swipeLookStartY;
     if (!this.swipeLookActive) {
@@ -1085,7 +1143,9 @@ export class MobileControls {
     if (e.pointerId !== this.swipeLookPointer) return;
     if (owner === 'groundAim') {
       e.preventDefault();
-      if (e.type === 'pointerup') this.callbacks.onGroundAimTap(e.clientX, e.clientY);
+      if (!this.swipeLookAdopted && e.type === 'pointerup') {
+        this.callbacks.onGroundAimTap(e.clientX, e.clientY);
+      }
       this.lastSwipeTapAt = 0;
       this.releaseSwipeLook();
       return;
@@ -1097,7 +1157,10 @@ export class MobileControls {
     // never crossed the swipe deadzone (never became a drag); two of those in
     // quick succession recenter the camera, mirroring the joystick logic.
     const now = this.now();
-    const quickTap = !this.swipeLookActive && now - this.swipeLookDownAt <= RECENTER_DOUBLE_TAP_MS;
+    const quickTap =
+      !this.swipeLookAdopted &&
+      !this.swipeLookActive &&
+      now - this.swipeLookDownAt <= RECENTER_DOUBLE_TAP_MS;
     if (quickTap && this.callbacks.onGroundAimTap(e.clientX, e.clientY)) {
       this.lastSwipeTapAt = 0;
       this.releaseSwipeLook();
@@ -1129,6 +1192,8 @@ export class MobileControls {
       this.input.setTouchLookVector({ x: 0, y: 0 });
     }
     this.swipeLookActive = false;
+    this.swipeLookResync = false;
+    this.swipeLookAdopted = false;
   }
 }
 
--- a/tests/mobile_controls.test.ts
+++ b/tests/mobile_controls.test.ts
@@ -358,6 +358,7 @@ class FakeElement extends EventTarget {
   value = '';
   blur(): void {}
   private captured = new Set<number>();
+  private captureGeneration = new Map<number, number>();
   /** Selectors this element (or a simulated ancestor) matches, for closest();
    *  drives touch_router.ts's isInteractiveHudElement checks in tests. */
   matchedSelectors: string[] = [];
@@ -373,11 +374,13 @@ class FakeElement extends EventTarget {
   }
 
   setPointerCapture(pointerId: number): void {
+    this.captureGeneration.set(pointerId, (this.captureGeneration.get(pointerId) ?? 0) + 1);
     this.captured.add(pointerId);
   }
 
   releasePointerCapture(pointerId: number): void {
     if (!this.captured.has(pointerId)) return;
+    const releasedGeneration = this.captureGeneration.get(pointerId) ?? 0;
     this.captured.delete(pointerId);
     // Real browsers dispatch lostpointercapture for both an explicit
     // releasePointerCapture() call and an implicit capture loss (spec
@@ -388,6 +391,14 @@ class FakeElement extends EventTarget {
     // uncaught: queue it with queueMicrotask so callers observe the same
     // non-reentrant timing as a real browser instead of synchronous recursion.
     queueMicrotask(() => {
+      // A new capture established before the pending loss is dispatched
+      // supersedes the old release. Browsers do not emit that stale loss for
+      // the new capture generation, so the fake must not either.
+      if (
+        this.captured.has(pointerId) ||
+        this.captureGeneration.get(pointerId) !== releasedGeneration
+      )
+        return;
       this.dispatchEvent(
         Object.defineProperties(
           new Event('lostpointercapture', { bubbles: true, cancelable: true }),
@@ -669,6 +680,65 @@ describe('MobileControls ground placement tap', () => {
       }),
     );
     expect(taps).toHaveLength(2);
+  });
+
+  it('commits a slow ground-target drag when the release arrives on the window path', () => {
+    const { canvas, windowTarget } = installMobileControlDom();
+    const moves: Array<{ x: number; y: number }> = [];
+    const taps: Array<{ x: number; y: number }> = [];
+    const lookDeltas: Array<{ x: number; y: number }> = [];
+    const callbacks = {
+      ...mobileCallbacks(),
+      onGroundAimMove: (x: number, y: number) => {
+        moves.push({ x, y });
+        return true;
+      },
+      onGroundAimTap: (x: number, y: number) => {
+        taps.push({ x, y });
+        return true;
+      },
+    };
+    const input = inputWithoutLook();
+    input.applyTouchLookDelta = (x: number, y: number) => lookDeltas.push({ x, y });
+    const nowSpy = vi.spyOn(performance, 'now').mockReturnValue(1000);
+    try {
+      new MobileControls(input, callbacks).start();
+
+      canvas.dispatchEvent(
+        pointerEvent('pointerdown', {
+          pointerId: 73,
+          pointerType: 'touch',
+          clientX: 200,
+          clientY: 300,
+        }),
+      );
+      canvas.dispatchEvent(
+        pointerEvent('pointermove', {
+          pointerId: 73,
+          pointerType: 'touch',
+          clientX: 260,
+          clientY: 340,
+        }),
+      );
+      nowSpy.mockReturnValue(1000 + RECENTER_DOUBLE_TAP_MS + 1);
+      windowTarget.dispatchEvent(
+        pointerEvent('pointerup', {
+          pointerId: 73,
+          pointerType: 'touch',
+          clientX: 260,
+          clientY: 340,
+        }),
+      );
+
+      expect(moves).toEqual([
+        { x: 200, y: 300 },
+        { x: 260, y: 340 },
+      ]);
+      expect(taps).toEqual([{ x: 260, y: 340 }]);
+      expect(lookDeltas).toEqual([]);
+    } finally {
+      nowSpy.mockRestore();
+    }
   });
 });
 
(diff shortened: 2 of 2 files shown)


jgyy (commented):
Went through this against `release/v0.27.0` head, ran the focused suite, biome, and tsc.

Nice fix. The pinch-to-swipe handoff (`onPinchEnd`, mobile_controls.ts:997-1038) is careful about the details that usually get missed here: it re-baselines `pinchPrevDist` on a 3-to-2 finger transition instead of only nulling it, it guards against the surviving finger being read as a camera-recenter double-tap (`swipeLookAdopted`), and it resyncs the inherited pointer's start position on its first live move so the handoff doesn't apply a stale delta as a jump. The `releasingCaptureForPointer` interaction with re-capturing the surviving finger (clearing the marker before `setPointerCapture`) is a real edge case and the added test for it (`handles a genuine capture loss after a pinch survivor is re-captured`) actually exercises it via the fake element's capture-generation tracking rather than just asserting the code path exists.

Also good: replacing `releaseSwipeLook()` with `releasePinch()` in the blur/visibilitychange handlers is the actual bug fix for stale pinch state surviving a focus loss, not just a rename; `releasePinch()` clears `pinchPointers`/`pinchPrevDist` and then tears down swipe-look, whereas before only swipe-look was cleared and a pinch you'd started before backgrounding the tab would stay tracked.

Checked:
- `git merge-tree` against `release/v0.27.0`: clean, no conflicts.
- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts`: 93 passed.
- `npx @biomejs/biome check` on both changed files: 0 errors (11 pre-existing `noNonNullAssertion` warnings in test helpers, matches what the PR description says).
- `tsc --noEmit`: clean.
- No player-visible strings added, so i18n is correctly N/A.
- Scope: `src/game/mobile_controls.ts` is the right home for this (input/gesture lifecycle), no reach into `render`/`ui`/`sim`, nothing here needs the `IWorld` seam.

Nit: the guard order in `onPinchEnd` (mobile_controls.ts:1004-1011) reads `size !== 1 || !this.active || ...`; harmless since it's all fast property checks, just flagging that `!this.active` before the size check would read slightly more naturally as "bail if inactive" first. Not worth a changed line for.

No898:
@jgyy Thank you very much for the thorough review and for checking the gesture lifecycle edge cases so carefully. I applied the small guard-order readability improvement as well.

The focused tests and the full local quality gate pass, GitHub CI is green, and the PR is conflict-free and currently mergeable. It is ready to merge.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and re-ran the checks locally in a clean worktree at `3528f44cf`.

## What I checked and ran

- `npx tsc --noEmit`: clean (exit 0).
- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts tests/architecture.test.ts`: all green (117 tests).
- `npx vitest run tests/localization_fixes.test.ts` after `npm run i18n:gen`: 29 passed, 3 skipped. (A first run failed only because `src/ui/i18n.status.json` was not generated yet in the fresh worktree, not because of this PR.)
- `npx @biomejs/biome check src/game/mobile_controls.ts tests/mobile_controls.test.ts`: no errors, 11 pre-existing warnings in the shared test helpers.
- Scanned the diff for em dashes, en dashes and emojis: none.
- Invariants: the change is confined to `src/game/`, touches no `src/sim/` code, no `IWorld` member, no server authority path, no player-visible string, no generated file, no secret or `ALLOW_DEV_COMMANDS`. i18n N/A is correct.
- Mutation check of my own: deleting `else this.pinchPrevDist = this.currentPinchDist();` makes `re-baselines the surviving pair on a three to two finger transition` fail, so that pin is decisive rather than self-confirming.

## Correctness read

The lifecycle reasoning holds up on a careful trace:

- `onPinchEnd` becoming idempotent via `if (!this.pinchPointers.delete(...)) return;` is what makes the new window-level registration safe: the canvas event fires first and the bubbled window event is a no-op. Same for `onSwipeLookEnd`, where `releaseSwipeLook()` nulls `swipeLookPointer` so the bubbled re-entry returns at the identity guard and the double-tap bookkeeping cannot run twice.
- Swapping `releaseSwipeLook()` for `releasePinch()` on blur and `visibilitychange` is a strict superset, since `releasePinch()` calls `releaseSwipeLook()` itself. No cleanup was lost.
- The handoff correctly leaves the surviving pointer in `pinchPointers`; `onPinchMove` then only updates its position and cannot zoom at size 1, and `onSwipeLookMove`'s `pinchPointers.size > 1` guard still passes. A third finger landing later goes through `onPinchDown` and cleanly tears the adopted drag down via `releaseSwipeLook()`.
- `swipeLookResync` absorbing the first live move is the right shape: the stale pinch coordinate never reaches `applyTouchLookDelta`, and the deadzone still has to be crossed afterwards.
- On the `releasingCaptureForPointer.delete(remainingId)` line: in a real browser the deliberate release from pinch start has already dispatched its `lostpointercapture` before the pointerup that triggers adoption, so the set is normally empty by then. The delete is defensive rather than load-bearing, but it is correct and the fake's capture-generation modelling matches the spec's "process pending pointer capture" step, so the test is not fabricating browser behavior.

I did not find a functional bug in the change.

## One blocking process item

The PR is based on `release/v0.27.0`, but `release/v0.28.0` is the current release branch, and per CLAUDE.md work should target the latest one. This is not a formality here: `src/game/mobile_controls.ts` has genuinely diverged on `v0.28.0`, which added the ground-target aim path (`onGroundAimMove` / `onGroundAimTap`, an extra `'groundAim'` touch owner branch in `onSwipeLookEnd`, and a `quickTap && onGroundAimTap(...)` arm in the recenter logic). This PR will conflict there and needs a real rebase, not a mechanical one.

Two things to get right while rebasing:

1. The adoption block sets `this.touchOwners.set(remainingId, 'camera')` unconditionally. On `v0.28.0` that can overwrite a `'groundAim'` owner, which would route the surviving finger into camera rotation instead of the aim path. Please confirm the intended precedence and pin it with a test.
2. The new `swipeLookAdopted` guard currently only suppresses the recenter double-tap. On `v0.28.0` the same `quickTap` value also gates `onGroundAimTap`, so an adopted pinch survivor could commit a ground-target spell on lift. Whichever way you decide, it deserves an explicit test.

The rebased result should be re-gated, since the mutation evidence in the description was gathered on the `v0.27.0` base.

Nice work on the tests: they exercise the real listener wiring rather than poking private state, and the capture-generation refinement to `FakeElement` is a genuine fidelity improvement to the harness rather than a shim to make one assertion pass.

No898:
@EnriqueGF thanks for the detailed review and for catching the `release/v0.28.0` drift.

I rebased the PR onto `release/v0.28.0` and resolved the new `groundAim` path explicitly:

- adopted pinch survivors now preserve an existing `groundAim` owner instead of being forced to `camera`;
- lifting an adopted pinch survivor no longer commits a ground-target tap;
- both cases are pinned with regression tests in `tests/mobile_controls.test.ts`.

Validation run locally:

- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts`
- `npx tsc --noEmit`
- `npx @biomejs/biome ci --changed --since=upstream/release/v0.28.0 --no-errors-on-unmatched`
- `git diff --check`

I also attempted the full local `npm run gate`; it currently fails outside this PR in `tests/deeds_window_focus.test.ts` with `localStorage.clear is not a function`. The PR-specific checks above pass, and GitHub now reports the PR as based on `release/v0.28.0` and mergeable.

Thanks again for the review.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and re-ran the checks locally on the head commit (`ffce87ca7`). This is a solid, well-scoped bugfix: the pointer lifecycle reasoning is correct, the comments explain the *why* (especially the superseded-capture-marker rationale), and the tests are genuinely decisive rather than decorative. One verified gap below that I would like closed before merge.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts`: 96 passed.
- Guard tests: `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/localization_fixes.test.ts`: pass (the localization guard needed `npm run i18n:gen` first, a stale generated artifact in my worktree, not anything from this PR; `pending=0` after regen).
- Adjacent suites for regressions: `touch_router`, `touch_tap`, `touch_peek`, `touch_item_drag`, `input`, `move_input`, `camera_follow`, `mobile_chrome_fade`, `death_input_reset`: 159 passed.
- `biome check` on both changed files: no errors, only the 11 pre-existing `noNonNullAssertion` warnings in the shared test helper. CI gates on errors and format, so this is fine.
- Hygiene: no em dashes, en dashes, or emojis in the added lines. No player-visible strings, so no i18n surface. `src/game/` only, so no sim determinism or `IWorld` parity surface.
- Merge: only these two files differ from `release/v0.28.0`, and `git merge-tree` merges cleanly. (Minor: the PR body says the base is `release/v0.27.0`, but the PR targets `release/v0.28.0`. Worth correcting the body.)

I independently reproduced your mutation checks, and they hold up:
- removing the window-level `onPinchEnd`/`onSwipeLookEnd` finalization fails 2 tests
- removing the `swipeLookAdopted` guards fails 2 tests

I also probed two paths the suite does not cover, and both behave correctly: re-pinching after an adoption (zoom resumes from the current pair with no jump and swipe-look is handed back), and a window opening mid-drag on an adopted finger.

## Should fix: a slow ground-target placement loses its commit on the window path

`onMoveEnd` (`src/game/mobile_controls.ts:834-835`) and `onCameraEnd` (`:932-933`) both call `touchOwners.release(e.pointerId)` unconditionally, *before* their own pointer-id guard. In the new window `pointerup`/`pointercancel` handlers they run at positions 1 and 2, while `onSwipeLookEnd` runs at position 4 and reads `touchOwners.get(e.pointerId)` at `:1140` to choose its branch. By then the entry is always gone, so the `owner === 'groundAim'` branch at `:1143` is unreachable via the window path and execution falls through to the camera/recenter branch.

Short gestures are rescued by accident: the `quickTap` fallback at `:1163` calls `onGroundAimTap` anyway. A *deliberate*, slower than 300ms (`RECENTER_DOUBLE_TAP_MS`) ground-target placement is not, and silently drops the cast.

I confirmed this with a scratch test rather than by inspection. Same gesture twice, identical except for the delivery target:

```ts
// canvas path (control): commits
touch(canvas, 'pointerdown', 96, 200, 300);
touch(canvas, 'pointermove', 96, 260, 340);
offset = 900;                                  // performance.now stubbed, > 300ms
touch(canvas, 'pointerup', 96, 260, 340);
expect(aimTaps).toBe(1);                       // passes

// window path: same gesture, terminal event only at window level
touch(canvas, 'pointerdown', 97, 200, 300);
touch(canvas, 'pointermove', 97, 260, 340);
offset = 900;
touch(windowTarget, 'pointerup', 97, 260, 340);
expect(aimTaps).toBe(2);                       // FAILS: got 1
```

This is not a regression against the current release (before this PR the window path did not call `onSwipeLookEnd` at all, and the state simply stayed latched, which is worse). But it is reachable on exactly the scenario this PR targets: it needs the terminal event to arrive at window level instead of being retargeted to the canvas, which is what happens when `setPointerCapture` threw into the swallowed `catch` at `:1082-1086` or when capture was lost without a `lostpointercapture` teardown.

Reordering the two new window handlers so the gesture handlers read the ledger first fixes it:

```ts
window.addEventListener('pointerup', (e) => {
  this.onPinchEnd(e);
  this.onSwipeLookEnd(e);
  this.onMoveEnd(e);
  this.onCameraEnd(e);
});
```

I applied exactly that and re-ran `tests/mobile_controls.test.ts` plus `tests/touch_router.test.ts`: my scratch test goes green and all existing tests stay green (98 passed). The adoption ordering is unaffected, because adoption requires `swipeLookPointer === null`, so a lifting pointer that *was* the swipe-look pointer never reaches the adoption branch anyway. Worth adding a regression test for the slow-drag case alongside the fix.

## Notes, no action required

1. **The harness cannot exercise real double delivery.** `installMobileControlDom` builds `windowTarget` as a bare `new EventTarget()` with the canvas as an unrelated `FakeElement` and no parent chain, so `bubbles: true` does nothing and each test exercises exactly one listener path. In a real browser a canvas `pointerup` runs the canvas handler *and then* the window handler. `onPinchEnd` is explicitly hardened for this (the new `if (!this.pinchPointers.delete(...)) return;`), but `onSwipeLookEnd` is idempotent only because the first pass nulls `swipeLookPointer`. That holds today; it is just worth knowing the tests are not pinning it.

2. **Adoption skips the `isCameraDragAllowedAt` classification** that `onSwipeLookDown` enforces at `:1072`, so a survivor now resting on HUD chrome becomes a camera drag. Low impact: the finger originated on the canvas, and `swipeLookResync` plus the deadzone suppress rotation until it actually moves. Mentioning it since the PR's own premise is that the survivor may have strayed off-canvas.

3. **`currentPinchDist()` is safe.** I checked all three call sites; the new `else` branch at `:1008` is the `else` of `size < 2`, so the unguarded `pts[0]`/`pts[1]` indexing cannot underflow. With three or more fingers it measures an arbitrary insertion-ordered pair, but that value is never consumed, since `onPinchMove` only zooms at `size === 2` and the baseline is recomputed on the 3-to-2 transition.

4. No per-frame or graphics-fairness surface: the adoption block runs once per gesture transition, reads no tier knob, and hides nothing actionable.

Happy to re-review as soon as the ordering item is addressed.

No898:
@EnriqueGF thanks for the follow-up review. I addressed the window-path ground-target issue.

Changes made:

- reordered the window `pointerup` and `pointercancel` cleanup so `onPinchEnd` and `onSwipeLookEnd` read the touch owner ledger before `onMoveEnd` and `onCameraEnd` release it;
- added a regression test for a slow ground-target drag whose `pointerup` arrives only through the window path;
- kept the test over the public DOM listener path rather than private state.

Validation run locally:

- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts` - 97 passed
- `npx tsc --noEmit` - passed
- `npx @biomejs/biome check src/game/mobile_controls.ts tests/mobile_controls.test.ts` - passed with the existing 11 test-helper warnings
- `npx @biomejs/biome ci --changed --since=upstream/release/v0.28.0 --no-errors-on-unmatched` - passed with the same existing warnings
- `git diff --check` - passed

The branch has been pushed and GitHub reports the PR as mergeable.

EnriqueGF (approved):
Reviewed this end to end against the current base and ran the checks locally. This is a good, well-scoped fix: the diff stays inside `src/game/`, touches no sim/wire/i18n surface, and the regression coverage is genuinely decisive (I mutated a couple of the guards mentally and the matching tests are the ones that would catch them).

**What I ran, on this branch:**

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts tests/architecture.test.ts`: 121 passed.
- `npx vitest run tests/localization_fixes.test.ts` (after `npm run i18n:gen`): 29 passed, 3 skipped.
- Wider sweep over every `mobile|touch|input|camera|game` test file: 530 passed. One unrelated flake in `crafting_view_combo_liveness.test.ts` (`registerLivenessSource`) that went green on a re-run, not from this change.
- `biome check` on both changed files: 0 errors, 11 pre-existing warnings in shared test helpers.
- Copy scan: no em dashes, en dashes, or emojis. No player-visible strings added, so the i18n checklist entry is correct.
- Merge against `release/v0.28.0`: clean. The branch is behind by 166 commits, but the only one touching these files is `44bf14677` (a 3-line addition to `mobile_controls.ts`), and `git merge-tree` produces no conflict.

**What I verified in the code, beyond the tests:**

- Double-handling now that canvas and window both call `onPinchEnd` + `onSwipeLookEnd`. Both are idempotent for the right reason: `onPinchEnd` bails on `!pinchPointers.delete(...)`, and `onSwipeLookEnd` always ends in `releaseSwipeLook()`, which nulls `swipeLookPointer` so the second pass returns before any commit. No double `onGroundAimTap` and no double `onRecenterCamera`.
- Ledger release ordering. `onSwipeLookEnd` releases `touchOwners` before `onMoveEnd`/`onCameraEnd` run on the same window event, but those guard on `joyPointer`/`lookPointer` rather than the ledger, so nothing gets stranded.
- Leaving the adopted pointer inside `pinchPointers` is correct: `onSwipeLookMove` only bails at `size > 1`, `onPinchMove` keeps its position fresh without zooming, and a third finger cleanly reverts to a pinch with a fresh baseline.
- No stuck-latch path. `releaseSwipeLook()` is the sole clearer of both new flags and the sole nuller of `swipeLookPointer`, and adoption only runs while `swipeLookPointer === null` with `swipeLookActive = false`, so `setTouchLook(true)` can never be left armed.
- The 3-to-2 re-baseline is a strict improvement: previously `pinchPrevDist` was left holding a value from before the third finger landed, which is exactly the jump this removes.

**Notes, none blocking:**

1. `onPinchMove` is still bound to the canvas only, while the end path is now window-level. During a pinch there is no pointer capture at all (`onPinchDown` calls `releaseSwipeLook()`, which drops the first finger's capture), so a finger that drifts over HUD chrome stops updating its tracked position for the rest of the gesture. That is pre-existing, and `swipeLookResync` already neutralizes it for the adoption handoff, but the new `pinchPrevDist = currentPinchDist()` re-baseline reads those same coordinates. Worth a follow-up that mirrors the end path and routes `onPinchMove` through the window listener too.
2. `releasingCaptureForPointer` is a `Set`, so `delete(remainingId)` before the re-capture cannot distinguish a still-pending old `lostpointercapture` from a genuine loss of the new capture. The spec ordering makes the stale event already consumed in practice (and the fake element in the test models this with a generation counter), so I do not think this is reachable on real engines, but a per-pointer generation counter in the production code would make the invariant structural instead of timing-dependent.
3. A ground-aim reticle that survives a pinch is now dismissed by `releaseSwipeLook()` without either committing the cast or explicitly cancelling aim mode. I checked the pre-change behavior and it was identical (the finger was dropped entirely at pinch start), so this is not a regression from this PR, but an explicit cancel callback would be tidier if you touch this path again.

Approving. The behavior change is the right one, the mutation checks in the PR description line up with what the tests actually pin, and nothing here crosses a repo invariant.

No898:
@EnriqueGF thanks again. I took the small non-blocking pinch-move note as an in-scope cleanup for this PR.

I added the window-level `onPinchMove` routing so pinch position tracking continues when a pinch finger moves off the canvas path, and pinned it with a regression test that sends the move through `windowTarget`.

Validation run locally:

- `npx vitest run tests/mobile_controls.test.ts tests/touch_router.test.ts` - 98 passed
- `npx tsc --noEmit` - passed
- `npx @biomejs/biome check src/game/mobile_controls.ts tests/mobile_controls.test.ts` - passed with the existing 11 test-helper warnings
- `npx @biomejs/biome ci --changed --since=upstream/release/v0.28.0 --no-errors-on-unmatched` - passed with the same existing warnings
- `git diff --check` - passed

I am intentionally leaving the production capture-generation counter and explicit ground-aim cancel callback for later follow-up work, since those are broader than this PR and you marked them as non-blocking.

Rubsey:

Thanks for the contribution, @No898.

PR #2073, opened 2026-07-17
fix(loot): stop a raid boss handing out the same item twice in one kill

jgyy:
## Summary
A player reported clearing the Nythraxis raid with a 9-person group and getting 4 drops that all felt like the same worthless item. Digging into `nythraxis_scourge_of_thornpeak`'s loot table explained why: it splits into 4 independent exclusive `rollGroup`s (2 helm slots, 2 shoulder slots), each drawn with its own RNG roll, and several items (`soulflame_mantle`, `crownforged_dreadhelm`, `nighttalon_crown`, `nighttalon_shoulderguards`) are listed in *every* one of the 4 groups. Nothing stopped the same item id winning more than one slot in a single kill.

Simulated 2000 kills of the boss against the unfixed table: **57% of kills produced at least one duplicate item id** in the drop pool. That's a systemic loot-variety bug, not a rare edge case, and it lines up with the report of a raid clear feeling like "4 pieces of the worst loot possible."

## Fix
`rollLoot` (and its heroic-tier counterpart) tracks every item id already awarded in the current loot event. A pure helper, `pickRollGroupWinner`, resolves a group's winning entry from the SAME single RNG draw as before (draw order and the parity contract are unaffected). On a collision (the partitioned entry's item id was already awarded by an earlier group this kill), it falls forward to the next non-awarded entry in the SAME group (wrapping), instead of dropping the slot: this both preserves the raid's guaranteed per-group drop and actually raises variety, since the surviving item is a different one than the collision. Only when every entry in the group is already awarded does the slot legitimately produce nothing.

(Earlier revision of this PR skipped the slot on a collision instead of falling forward - see review discussion below for why that undercounted drops without improving variety.)

## Why this shape
- Kept the fix inside `src/sim/loot/loot_roll.ts`, the module that already owns `rollLoot` (per this repo's module-first / `SimContext` seam conventions) - no new system, no new seam.
- Left the quest-item and plain-chance loot branches untouched: those are single guaranteed-need items, not exclusive multi-item slots, so widening the dedup scope there would risk breaking quest-item guarantees.
- The one golden-trace fixture that happened to exercise this exact collision (`tests/parity/golden/nythraxis_full_pull.json`) was regenerated via `UPDATE_PARITY=1`; the diff shows exactly what you'd expect - a duplicate `soulflame_mantle` collision now falls forward to `crownforged_dreadhelm` in the same group instead of collapsing to a single drop. RNG draw order and digest are unaffected.
- Added content-invariant tests pinning that no mob's heroic table shares an item id across its own rollGroups or with its base table, plus a forced-collision test exercising the heroic-append branch directly (previously reachable in code but untested, since no real heroic table collides today).

## Test plan
- [x] New regression test in `tests/loot_roll.test.ts`: rolls the raid boss 60 times across seeds and asserts every kill's item list is non-empty with no duplicate ids (fails without the fix, passes with it).
- [x] Direct unit tests for `pickRollGroupWinner`: plain-partition winner, fall-forward on collision, wraparound, and the every-entry-already-awarded null case.
- [x] Forced-collision test exercising the heroic-append branch's dedup path with a synthetic fixture (temporarily substituted into `HEROIC_BOSS_LOOT`, restored after).
- [x] Two content-invariant tests: no mob's heroic table shares an item id across its own rollGroups, and no mob's heroic table shares an item id with its own base table.
- [x] `npx vitest run tests/loot_roll.test.ts` - all 29 tests pass.
- [x] `npx vitest run tests/parity` - green after golden regen (`UPDATE_PARITY=1`).
- [x] `npx vitest run tests/architecture.test.ts tests/loot_drops.test.ts tests/loot_master.test.ts tests/loot_master_sim.test.ts` - all green.
- [x] `npx tsc --noEmit -p .` - clean.
- [x] `npx biome check` on every file this PR touches - clean (0 errors; only pre-existing `noExplicitAny` warnings in test helpers this PR didn't introduce).
- Not a visual change (no `render/`/`ui/` code touched), so no screenshots per the repo's screenshot rule.

## Diagram
```mermaid
flowchart TD
    A[Nythraxis dies] --> B[rollLoot runs the 4 rollGroups]
    B --> G1[drop_1: helm slot]
    B --> G2[drop_2: shoulder slot]
    B --> G3[drop_3: helm slot]
    B --> G4[drop_4: shoulder slot]
    G1 -->|rng roll picks item X| C{X already awarded this kill?}
    G2 -->|rng roll picks item Y| C
    G3 -->|rng roll picks item Z| C
    G4 -->|rng roll picks item W| C
    C -->|No, first time| D[Add item to corpse loot]
    C -->|Yes, duplicate| E[Fall forward to the next non-awarded entry in the SAME group]
    D --> F[Raid gets up to 4 distinct items, every group still awards]
    E --> F
```

### DIFF
--- a/scripts/critter_tour_shot.mjs
+++ b/scripts/critter_tour_shot.mjs
@@ -36,9 +36,9 @@ const browser = await puppeteer.launch({
 });
 const page = await browser.newPage();
 const errors = [];
-page.on('pageerror', (e) => errors.push('PAGEERROR: ' + e.message));
+page.on('pageerror', (e) => errors.push(`PAGEERROR: ${e.message}`));
 page.on('console', (msg) => {
-  if (msg.type() === 'error') errors.push('CONSOLE: ' + msg.text());
+  if (msg.type() === 'error') errors.push(`CONSOLE: ${msg.text()}`);
 });
 
 // Pre-set the first-run flags the real client honors, so the camera-mode-choice
--- a/src/sim/loot/loot_roll.ts
+++ b/src/sim/loot/loot_roll.ts
@@ -130,23 +130,35 @@ function effectiveItemLootStrategy(ctx: SimContext, itemId: string, mob: Entity)
   return q === 'poor' || q === 'common' ? strategies.commonItems : strategies.premiumItems;
 }
 
-// Resolves a single exclusive rollGroup draw to its winning entry (or null on a
-// non-matching partition, or on a matching item already awarded elsewhere in this
-// same loot event). Pure partition math; the caller supplies the one rng draw so the
-// draw-order/parity contract (exactly one ctx.rng.next() per group) is unaffected by
-// this dedup check.
-function pickRollGroupWinner(
+// Resolves a single exclusive rollGroup draw to its winning entry. Pure partition
+// math; the caller supplies the one rng draw so the draw-order/parity contract
+// (exactly one ctx.rng.next() per group) is unaffected by the dedup below.
+//
+// When the partition lands on an item id already awarded by an earlier group in
+// this same loot event, this falls forward deterministically to the next entry in
+// the SAME group (wrapping), rather than dropping the slot entirely: that is what
+// actually raises drop variety per kill (a plain skip just deletes the duplicate,
+// leaving the surviving item set unchanged and costing the raid a guaranteed
+// drop). Only when every entry in the group is already awarded does the slot
+// legitimately produce nothing.
+export function pickRollGroupWinner(
   roll: number,
   group: LootEntry[],
   awardedItemIds: Set<string>,
 ): LootEntry | null {
   let cumulative = 0;
-  for (const g of group) {
-    cumulative += g.chance;
+  let winnerIndex = -1;
+  for (let i = 0; i < group.length; i++) {
+    cumulative += group[i].chance;
     if (roll < cumulative) {
-      if (g.itemId && awardedItemIds.has(g.itemId)) return null;
-      return g;
+      winnerIndex = i;
+      break;
     }
+  }
+  if (winnerIndex === -1) return null;
+  for (let offset = 0; offset < group.length; offset++) {
+    const candidate = group[(winnerIndex + offset) % group.length];
+    if (!candidate.itemId || !awardedItemIds.has(candidate.itemId)) return candidate;
   }
   return null;
 }
@@ -186,8 +198,10 @@ export function rollLoot(
   // Cross-group duplicate guard: several exclusive rollGroups on the same mob (e.g.
   // Nythraxis's 4 helm/shoulder slots) can share item ids, and each group draws its
   // own independent rng.next(). Without this, one kill could hand out the same piece
-  // twice (or more) instead of a spread across the raid; a repeated winner yields no
-  // item for that slot rather than a second copy.
+  // twice (or more) instead of a spread across the raid; a repeated winner falls
+  // forward to the next non-awarded entry in its own group instead (see
+  // pickRollGroupWinner), preserving both the guaranteed per-group drop and the
+  // single rng draw.
   const awardedItemIds = new Set<string>();
   // A heroic dungeon claim upgrades the mob's normal epic/rare drops to their
   // "Heroic" variant in place (content/heroic_variants.ts). Resolved once and
--- a/src/sim/loot/loot_roll.ts
+++ b/src/sim/loot/loot_roll.ts
@@ -130,6 +130,27 @@ function effectiveItemLootStrategy(ctx: SimContext, itemId: string, mob: Entity)
   return q === 'poor' || q === 'common' ? strategies.commonItems : strategies.premiumItems;
 }
 
+// Resolves a single exclusive rollGroup draw to its winning entry (or null on a
+// non-matching partition, or on a matching item already awarded elsewhere in this
+// same loot event). Pure partition math; the caller supplies the one rng draw so the
+// draw-order/parity contract (exactly one ctx.rng.next() per group) is unaffected by
+// this dedup check.
+function pickRollGroupWinner(
+  roll: number,
+  group: LootEntry[],
+  awardedItemIds: Set<string>,
+): LootEntry | null {
+  let cumulative = 0;
+  for (const g of group) {
+    cumulative += g.chance;
+    if (roll < cumulative) {
+      if (g.itemId && awardedItemIds.has(g.itemId)) return null;
+      return g;
+    }
+  }
+  return null;
+}
+
 function needsQuestDrop(ctx: SimContext, entry: LootEntry, meta: PlayerMeta): boolean {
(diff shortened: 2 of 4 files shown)


FernandoX7 (commented):
The diagnosis is right and the mechanism is solid. I reproduced the collision math against the real table (20k seeded kills: 58.4% produce a duplicate, matching your 57%), and verified the parity claim end to end: `pickRollGroupWinner` draws no rng, the skip fires after the one draw per group, and the golden diff is exactly draw-preserving (0 `draws` counts changed, 0 `events` hashes, 8 `state` hashes, the 2 duplicate blocks removed). The regression test is genuinely failing-first: it fails on the base `loot_roll.ts`, passes at head, and the `sim.rng` re-seed is live (ctx.rng is a getter over the public field). I also scanned the full content set: Nythraxis is the only mob with cross-group shared item ids, and no mob shares an id between a rollGroup entry and a plain-chance entry, so the deliberate scoping of the guard is sound and the blast radius is exactly this boss.

Two should-fix items:

- should-fix `src/sim/loot/loot_roll.ts:147`: skip-on-duplicate reduces loot quantity without improving variety. Because the winner per group is unchanged and repeats are deleted, the post-fix item list is exactly the deduped pre-fix list: I verified over 20000 seeded kills that the distinct-item set per kill is identical before and after, every time. Meanwhile all 4 Nythraxis groups sum to 1.00 chance (`src/sim/content/dungeons.ts:568`), so the boss goes from a guaranteed 4 items per kill to an average of 3.31, losing at least one drop in 58% of kills. The PR body's "directly maximizes drop variety" and "without touching any drop-rate numbers" both overstate it: variety is untouched and effective quantity drops. Deleting the second copy may still be the right call (two raid members using duplicate shoulders was normal in classic, so this is a design choice either way), but it should be an explicit decision, not a side effect. If the 4-drop guarantee should survive: fall forward deterministically to the next non-awarded entry in the same group (still one draw, and that variant genuinely increases variety), or make the four groups draw from disjoint pools in the content table.

- should-fix `src/sim/loot/loot_roll.ts:273`: the heroic-append arm of the guard is untested, and with current content it is unreachable (no heroic table shares an id across its groups, with a base id, or with a `heroic_` variant of one, so it can never fire today). The commit claims both branches but only the base branch is exercised, and the disjointness the arm relies on is an unenforced content invariant: a future heroic-table edit could silently reintroduce duplicate awards with zero coverage. Cheapest closure: a targeted test with an overlapping-id fixture under a heroic claim, or at minimum a comment marking it a forward guard plus a content test pinning the heroic/base id disjointness.

Nits:

- nit `tests/loot_roll.test.ts:120`: the uniqueness assertion passes vacuously if the corpse ends up empty (0 === 0). Over-suppression is caught in aggregate by the golden and the drop-rate tests, but `expect(ids.length).toBeGreaterThan(0)` inside the loop makes this test self-sufficient, and with every group summing to 1.0 it cannot flake.
- nit (PR body): the attribution footer emoji; the repo convention bans emojis in PR text too.

jgyy:
Thanks for the sharp catch on the variance math, and for actually simulating it out to confirm.

You're right that skip-on-duplicate was the wrong shape: it deleted the collision instead of resolving it, so the distinct-item set was identical to the deduped pre-fix list while the raid quietly lost a guaranteed drop 58% of the time. Went with your suggested fix: `pickRollGroupWinner` now falls forward deterministically to the next non-awarded entry in the SAME group (wrapping), off the SAME single rng draw, so the 4-drop guarantee holds and the surviving item is genuinely a different one on a collision, not just "the same set with a hole in it."

For the heroic-append arm: added a forced-collision test (a synthetic fixture temporarily substituted into `HEROIC_BOSS_LOOT`, restored after) that exercises the exact same fall-forward path through the heroic branch, plus two content-invariant tests pinning that no mob's heroic table shares an item id across its own rollGroups or with its own base table - so a future content edit that reintroduces an overlap now fails loudly instead of silently reactivating an untested branch.

Also fixed the vacuous `0 === 0` uniqueness assertion (now asserts the corpse is non-empty first) and dropped the emoji from the PR body footer.

Updated the PR body and diagram to describe fall-forward instead of skip. Pushed as a new commit; golden regenerated via `UPDATE_PARITY=1` (rng draw order/digest unaffected, only the awarded item set changed on the one collision the fixture exercises).

FernandoX7 (commented):
Re-reviewed at f17574108. Everything from the first review is addressed, and I verified each item against the head code rather than the description.

The fall-forward rework is correct and draw-preserving. `pickRollGroupWinner` (`src/sim/loot/loot_roll.ts:144`) draws no rng, the caller still does exactly one `ctx.rng.next()` per group, the partition winner is unchanged, and only the collision path walks forward (wrapping) inside the same group. Bonus groups that sum under 1.0 (`morthen_bonus` at 0.81, for example) still whiff via the `winnerIndex === -1` return before any fall-forward, so there is no drop-rate inflation outside the collision case. The golden regen is exactly what the mechanism predicts: zero `draws`, `digest`, or `events` lines changed, 8 `state` hashes, and the two duplicate `soulflame_mantle` blocks now resolve to `crownforged_dreadhelm` instead of vanishing.

The heroic-append arm is genuinely covered now: the forced-collision fixture (`tests/loot_roll.test.ts:590`) drives the real `rollLoot` heroic branch through an instance claim, and the two content pins fence the disjointness invariant. The vacuous uniqueness assertion is fixed (non-empty asserted inside the loop) and the emoji footer is gone from the PR body.

I also mutation-tested the new suite to confirm it is decisive: with the dedup disabled entirely (base behavior), 5 tests fail including the 60-kill regression; with skip-on-duplicate (this PR's earlier revision), 3 tests fail (both fall-forward pins plus the heroic forced-collision test). At head: `tests/loot_roll.test.ts` 29 passing, `tests/parity` plus `tests/architecture.test.ts` 195 passing, `tsc --noEmit` clean.

Non-blocking notes:

- nit (CI): the failing "Screenshots of changes" check is not this PR's fault. That run was cancelled by the workflow's own concurrency group (the 14:13 reply comment retriggered `pr-ai` for the same PR with cancel-in-progress), and the `if: always()` embed step then ran in a half-checked-out workspace and died on a missing module. It is an informational, non-required job, and this PR is non-visual anyway, so a re-run would capture nothing; I will re-run or ignore it.
- nit `tests/loot_roll.test.ts:567`: the base-vs-heroic content pin compares raw base ids, not their `heroicItem()` resolved variants, so a heroic entry matching a base item's heroic variant would slip past the fence. Runtime stays correct either way (the base branch adds both the base and resolved id to `awardedItemIds`, so that collision falls forward, and that path is now tested), so this is a completeness note on the fence, not a gap in the fix.

The unrelated `critter_tour_shot.mjs` template-literal commit is a two-line pre-existing biome autofix, honestly labeled; fine to ride along.

Good to merge from my side.

FernandoX7 (approved):

PR #2074, opened 2026-07-17
feat(solana): woc_auction_escrow - non-custodial on-chain auction escrow (devnet-proven, flag-off)

awidearray:
## What this is

A standalone, non-custodial **English-auction escrow** Anchor program for one SPL token (the in-game $WOC mint on mainnet; a fresh test mint on devnet). It is the on-chain half of self-custodied $WOC auction bidding: a bid must be refunded the instant a higher bid lands, and for a wallet-custodied token that refund cannot be server-mediated, so it is enforced on chain.

Each auction is backed by a program-owned PDA vault whose authority **is** the auction state PDA, so deposited bids can only move via a program-signed CPI. The server never holds bidder keys.

This is **devnet-proven and ships fail-closed OFF**; it is posted **DRAFT**. The ceiling here is devnet: mainnet enablement is separately gated on treasury-key custody and legal review.

## The core primitive

`place_bid` pulls the new bid into the vault **and** refunds the prior high bidder their exact prior bid **in one atomic instruction**, via a program-signed CPI whose destination is constrained to the on-chain-recorded prior bidder's account (owner + mint checked). After the swap the vault holds exactly the new high bid. There is no window in which the vault holds two bids or a refund can be misrouted.

## Instructions (5)

- `open_auction(auction_id, seller, min_bid, rake_bps, buyout, settler, ends_at)` — creates the Auction PDA (`seeds = [b"auction", auction_id.le_bytes()]`) + program-owned vault ATA; no token moves. `rake_bps` capped by the `MAX_RAKE_BPS` program constant; `ends_at` bounded to `[now+MIN_DURATION, now+MAX_DURATION]`.
- `place_bid(auction_id, amount)` — the atomic-outbid-refund primitive above; a bid must meet `min_bid` (first) or beat the standing high bid by `MIN_RAISE_BPS` (5%).
- `settle_auction(auction_id)` — settler-signed (`has_one = settler`); burns the capped rake, pays the seller (`address = auction.seller`), closes the vault.
- `cancel_auction(auction_id)` — settler-signed; refunds the standing high bidder, closes.
- `reclaim_expired(auction_id)` — **permissionless** liveness valve after `ends_at + RECLAIM_GRACE`, so a down settler key can never strand funds.

## Non-custodial guarantees

- Funds leave the vault only via a program-signed CPI to on-chain-recorded parties (outbid refund → prior bidder; settle payout → seller minus burned rake; cancel refund → high bidder).
- The settler (realm key) cannot name a destination; it only triggers the deterministic outcome recorded on chain.
- The rake is a **program constant** (`MAX_RAKE_BPS = 1000`), **burned** from the vault, not skimmed to a treasury.
- All money math is checked (`checked_add/mul/div → Overflow`) on top of `overflow-checks = true`.

## Devnet proof (Finalized commitment)

Program `GzkjHHkkJqZf5Qa3YZaJo1iHcKkxi4zUT5sV3zpVeyqZ` · mint `7QrwJsPLoSwwYByEiWRJogCD2cc2ybR22au3XFMZqMgD` · [deploy tx](https://explorer.solana.com/tx/5qRF3kstpkp8dHDw1BiVhkgUzCdgHbftzVgMwHSeMvVaBSdL82axKWkehEW6vjbA5CF9YLydU7mnfFVYwddZtaz6?cluster=devnet). The full signature set (14 tx) with explorer links is committed in [`solana/devnet-proof.json`](solana/devnet-proof.json). The gated test `solana/tests/woc_auction_escrow.devnet_alltypes.test.ts` (run with `WOC_DEVNET_TEST=1`) exercised, all Finalized via HTTP-polled `getSignatureStatuses` (never `confirmTransaction`):

- **Auction 1** — a three-rung outbid ladder, each rung asserting the prior bidder's balance is restored on-chain and the vault holds exactly the new high bid, then settle-on-buyout (seller paid `high_bid - rake`, rake burned from supply): [open](https://explorer.solana.com/tx/5e9xEvwXPv7Br3DbsJ8uQ3vpceKoEZ5bpnGjv4ze6uU2fPboSGrPJZLFyH91TpoDCsYmAG7C9NeviecgpQQc5VCf?cluster=devnet) · [bid1](https://explorer.solana.com/tx/5j7YYa6kEW3umNFCEj9kRjqmuNUHDiTfbgGb5hr5SHzQ4ZaTU1BxieVdxM5sACcX95J1YRcF4RxjkU6xog8waURe?cluster=devnet) · [bid2 (refunds bid1)](https://explorer.solana.com/tx/5dqQbUUybewbJBwJ77QJhHYSGasWq5XktKEtP8uzwvCYog7VhgpMUyZRmDE7YgcruaiKqMWGu8odddfeUrnLsMBM?cluster=devnet) · [bid3 (refunds bid2)](https://explorer.solana.com/tx/5WrzT8H3GCv5EayomHS1F7dsprYuVssxMeXVHL5ZyhWDu8ifz4gpdVfAEZH8Gf9B5YJLjPMmAVoRb3QMWxpt5udT?cluster=devnet) · [settle](https://explorer.solana.com/tx/4rMesqHauGhk8Z4DAaQ67krDqU6Aw5ndG8Astab2wXNr5A4G9rWqS7uuWZ5KFRaLM1Y5W3MqyCF6vYZVpmerAgsR?cluster=devnet)
- **Auction 2** — open, one bid, [cancel refunds the bidder](https://explorer.solana.com/tx/fBXunMHsVRMLymAmuH3eUAR449K5qgg88da2nszKXzrASTpv7vcosZgmRwRzA2vtRsLgkfJ3bQt1C2JVUDDqCv1?cluster=devnet).
- **Auction 3** — the [`reclaim_expired` guard](https://explorer.solana.com/tx/3PqMCfawHLYYXakDWFTdtuLjqcrkx7tmr1keYBarMCsvCV1mHNLzG5N3kmRVoPtisH3rWhHwSueb3CqNoS5zMEty?cluster=devnet) is correctly rejected before the grace window (`NotEnded`).

## Notes for maintainers

- **IDL generation** is broken on this toolchain (`anchor-syn 0.30.1` calls `proc_macro2::Span::source_file()`, removed in the pinned `proc-macro2 1.0.106`) — a known Anchor 0.30.1 incompatibility. The program `.so` builds and deploys cleanly; only the IDL step is affected. The devnet test builds instructions manually so the proof is toolchain-independent. To regenerate the IDL, pin an Anchor-0.30.1-era `proc-macro2`. Details in [`solana/programs/woc_auction_escrow/README.md`](solana/programs/woc_auction_escrow/README.md).
- The **release line has no `solana/` source tree** today; this is a self-contained new crate. Where it should ultimately live on the release line (a dedicated `solana/` workspace vs. a separate repo) is a coordination question — happy to move it.
- The on-chain byte encoding/serialization is expected to raise a malware-scan flag by design; no scanner change is included.

Posted **DRAFT** (devnet-proven; mainnet gated on treasury keys + legal).


EnriqueGF (changes requested):
Reviewed as a standalone on-chain component. This is genuinely good work: the core `place_bid` primitive (pull new bid, refund prior bidder in one atomic program-signed CPI) is correct, the fund-routing constraints hold up under adversarial reading, and the devnet proof set is a real, verifiable artifact rather than a claim. I tried hard to find a path that moves tokens to an unintended destination and could not. Below are the things I would want resolved before this leaves draft, ordered by severity.

## What I reviewed and ran

- Full read of `solana/programs/woc_auction_escrow/src/lib.rs` (806 lines), the devnet test, the script, and every config in `solana/`.
- `npx tsc --noEmit` on the repo root: clean (exit 0).
- Guard tests: `tests/architecture.test.ts` and `tests/world_api_parity.test.ts` pass (275 assertions). The `localization_fixes` failure is a missing `i18n:gen` artifact in my worktree, not this PR.
- `npx vitest run solana/tests/`, `npx @biomejs/biome check solana/`, `node scripts/malware_scan.mjs`.
- `cargo` is not available in my environment, so I could not build the program or run the Rust unit tests. I verified the `compute_rake` / `compute_min_acceptable` test expectations by hand instead; all seven are arithmetically correct.

## Blocking

**1. The new test file breaks `npm test` for the whole repo.**

`vite.config.ts` declares no `test.include`, so vitest uses its default `**/*.test.ts` and collects `solana/tests/woc_auction_escrow.devnet_alltypes.test.ts`. It is a mocha/chai file, so collection fails outright:

```
$ npx vitest run solana/tests/
solana/tests/woc_auction_escrow.devnet_alltypes.test.ts:103:19
ReferenceError: describe is not defined
Test Files  1 failed (1)
```

`npm test` is a bare `vitest run`, so this fails CI and `npm run gate` regardless of `WOC_DEVNET_TEST`. The gate on line 103 protects execution, not collection. Fix by adding `solana/**` to the `exclude` array in `vite.config.ts` (alongside `tmp/**` and `tests/browser/**`), or by renaming off the `.test.ts` suffix.

**2. Biome fails on three of the changed files.**

CI gates changed files via `biome ci --changed`, and format diffs are errors:

```
solana/devnet-proof.json                             format
solana/scripts/woc_devnet_tx_shots.mjs               format
solana/tests/woc_auction_escrow.devnet_alltypes.test.ts  format + useTemplate + organizeImports
```

`npx @biomejs/biome check --write` on those three files clears it.

**3. Critical: the `auction_id` namespace is global, unauthenticated, and never reclaimable.**

Three properties combine into a permanent denial-of-service:

- Seeds are `[b"auction", auction_id.to_le_bytes()]` (lib.rs:467), with no seller, mint, or realm scoping.
- `open_auction` is gated only by `opener: Signer` (lib.rs:460). Any keypair can open any id, and both `seller` and `settler` are attacker-chosen instruction arguments.
- No instruction ever closes the Auction PDA. I grepped: there is no `close =` anywhere in the file. `settle_auction`, `cancel_auction`, and `reclaim_expired` close only the vault ATA.

So if the realm allocates ids from a sequential or predictable source, an attacker front-runs a range of ids for the cost of rent each, and every legitimate `open_auction` for those ids fails on `init` forever. Settling the squatted auction does not help: status flips to Settled and the account persists.

The secondary concern is worse than the DoS. A squatter who sets themselves as both `seller` and `settler` with `min_bid = 1` owns an auction that any client resolving "auction N" by id alone would render as legitimate. Bids placed there are paid out to the attacker at settle, entirely within the rules of the program. Suggested direction: bind the seed to the seller or a realm authority (`[b"auction", seller.as_ref(), auction_id]`), and require a recorded authority signer on `open_auction`.

## Should fix

**4. `cancel_auction` is wedgeable by a one-unit donation, and the comment claims otherwise.**

lib.rs:293-295 states the full-vault drain exists "so a griefing transfer into the open vault ATA cannot wedge the close." That holds only when a bid exists. With zero bids, anyone sends 1 base unit to the vault ATA (permissionless for an ATA), so `total > 0` enters the block and `require!(high_bid > 0, WrongPriorBidder)` at lib.rs:296 reverts. Skipping the transfer is not an option either, since `close_account` rejects a nonzero balance.

Funds are not lost: after `ends_at`, `settle_auction` succeeds with `rake = 0` and sweeps the donation to the seller. But early cancel is deniable for up to 7 days, on every open auction, for one token unit. The load-bearing part is that the comment documents a guarantee the code does not provide.

**5. The settler holds an undisclosed unconditional sale veto.**

`cancel_auction` gates only on `status == Open` (lib.rs:283), with no `ends_at` and no buyout check. The realm key can void a completed sale at any point before settle. The winner is refunded so nothing is stolen, and the routing guarantees in the header comment are still literally true, but the "non-custodial" bullets at lib.rs:14-22 do not disclose that the settler can cancel a won auction. Either gate on `now < ends_at` or state it plainly in the doc comment.

**6. Donations inflate the payout and dilute the rake.**

Both settle paths compute `payout = vault.amount - compute_rake(high_bid, rake_bps)` (lib.rs:235-237, 357-359): the payout base is the vault balance, the rake base is `high_bid`. A donation flows to the seller un-raked. Not a loss path (`checked_sub` cannot underflow, since `vault.amount >= high_bid >= rake` on every reachable path), but the burn is not `rake_bps` of what the seller actually receives.

## Minor

- **Test coverage.** The Rust tests cover only the two pure helpers. The whole state machine (outbid swap, double-settle, the reclaim grace window, the wedge in item 4) is exercised only by a devnet test that is skipped by default and needs a funded keypair. A `litesvm` or `solana-program-test` suite would let CI actually protect this program. Every scenario in items 3 through 6 is cheap to express as a unit test.
- **`solana/tests/...test.ts:55`** defaults `ANCHOR_WALLET` to `/Users/futjr/woc/world-of-claudecraft/.secrets/deployer.json`. No key material is committed, but a contributor's absolute home path should not be the fallback. I confirmed there are no committed secrets anywhere in `solana/`.
- **`solana/package.json`** is named `@woc/economy-programs` and its description lists `woc_escrow`, `woc_lp_vault`, `realm_stake_escrow`, `job_escrow`, and `character_market`. Only `woc_auction_escrow` exists here, and it is the sole Cargo workspace member. Looks copy-pasted; worth trimming to what this PR actually ships.
- **No `solana/CLAUDE.md`.** Root CLAUDE.md asks a new multi-file subsystem to carry a local one. Given your open question about where this tree should live, that file is the natural place to record the answer, the Anchor 0.30.1 IDL breakage, and how to run the devnet proof.
- **PR body** uses em dashes throughout, which the root CLAUDE.md rules out for PR text as well as code.
- I scanned the whole `solana/` tree for em dashes, en dashes, and emojis: clean.

## On placement

Agreeing with your own note: `release/v0.28.0` exists and is 67 commits ahead of the `release/v0.27.0` base here, and still has no `solana/` tree. Since none of this compiles or runs as part of the game build, a dedicated repo or workspace seems right to me, mostly so the Rust and Anchor toolchain never becomes a prerequisite for `npm run gate`. If it does stay in-tree, item 1 needs the vitest exclude regardless.

Happy to re-review once items 1 through 3 are addressed. The cryptographic core here is solid; my concerns are about the namespace design and repo integration, not the escrow math.

EnriqueGF:
Addendum to my review: one more CI blocker that I under-checked the first time round, plus a few test-harness notes.

## Additional blocker: the malware-scan gate fails

`scripts/gate.mjs` runs `scripts/malware_scan.mjs` as a hard gate. It reports **11 HIGH findings, every one of them in the new `solana/` tree**, and exits 1:

```
[high] supply-chain  solana/package.json:1                              (x2: @solana/web3.js, @solana/spl-token)
[high] web3-drain    solana/tests/...devnet_alltypes.test.ts:38,48,130
[high] key-exfil     solana/tests/...devnet_alltypes.test.ts:75,278,279,280,281,283
```

There are zero HIGH findings elsewhere in the repo, so this is entirely new and not pre-existing debt. The `key-exfil` hits are `Keypair.fromSecretKey` / `Keypair.generate`, the `web3-drain` hits are the `@solana/web3.js` imports and `sendRawTransaction`, and the supply-chain hits are the two new Solana deps against the scanner's expectation that the live wallet stack is `@wallet-standard/*` plus Reown plus bs58 plus tweetnacl.

All eleven are benign in context, since this is a devnet test harness generating throwaway keypairs. But the scanner is a gate rather than an advisory, so it needs an explicit path-scoped allowlist entry in `scripts/malware_scan.mjs` with a triage rationale. Worth noting your PR text anticipated a flag from "on-chain byte encoding/serialization"; the actual categories that fire are different ones, so the exemption has to be written against `key-exfil`, `web3-drain`, and `supply-chain`.

That makes four independent CI blockers: vitest collection, biome format, this scan, and the design issue in item 3 of my main review.

## Test-harness notes (minor, but worth fixing while you are in there)

Having read the assertions closely, they are decisive and they do prove what the PR claims: the refund checks at `:351` and `:361` compare real before/after balances rather than self-comparisons, the vault checks at `:352` and `:362` pin literal constants, and the burn at `:370` correctly asserts a **total supply delta**, which is the right way to distinguish a burn from a transfer. Values come through `BigInt(...)`, which throws on undefined rather than passing trivially. Good harness. Three snags:

- **`:151-153` reads the Auction account by raw magic offsets** (`subarray(138, 170)`, `readBigUInt64LE(170)`, `readUInt8(178)`) with no cross-check against the Rust struct. Any field reorder in `lib.rs` makes these silently read garbage instead of failing loudly. Same fragility in the hand-rolled ABI encoder and the account-order comments. Since the IDL is unavailable on this toolchain, a single derived-offset table with a comment tying each offset to its field would at least localise the breakage.
- **`:428-447`, the `after()` hook overwrites the committed `solana/devnet-proof.json`**, because `SIGS_OUT` at `:56` defaults to a relative path. Running the proof dirties a tracked file. Default it to an untracked output path.
- **`:419-421`** asserts the reclaim-before-grace rejection with `expect(logs).to.match(/NotEnded|6012/)`. The alternation means an unrelated `6012` would also pass. Tighten to the specific error.
- **`solana/scripts/woc_devnet_tx_shots.mjs:19`** hardcodes the macOS Chrome path, and imports `puppeteer-core`, which is a root dependency but is not declared in `solana/package.json`. The script is therefore unresolvable from inside `solana/` and cannot run on Linux.

None of these change my overall read: the escrow logic is sound and the devnet proof is real. The work left is integration and the namespace design.

PR #2075, opened 2026-07-17
fix(desktop): add Linux PRIME render-offload for the discrete GPU

jgyy:
## Summary
- The desktop shell already forces the discrete GPU on Windows (`GpuPreference=2` in the per-app `UserGpuPreferences` registry value) and appends the cross-platform Chromium `force-high-performance-gpu` switch, but on a Linux hybrid-graphics laptop (NVIDIA Optimus, AMD/Intel Mesa PRIME) the app still bound the integrated GPU.
- **Updated after review** (see discussion below): an in-process `process.env` mutation before `app.whenReady()` does NOT work on Linux, unlike on Windows/macOS - Electron's Linux GPU process forks from a zygote that already exec'd and snapshotted its environ before any main-process JS ever runs. `electron/gpu_preference.cjs` now exports `relaunchForLinuxPrime`, which `main.cjs` calls as the very first thing it does: on Linux, if any PRIME variable is missing, it re-execs the whole app (same binary, same argv) with the variables baked into the *new* process's environment from birth, then the original process exits immediately (a marker env var prevents a relaunch loop).
- The variable set was also corrected against real hybrid hardware: `__GLX_VENDOR_LIBRARY_NAME` alone is a decoy for this app (Chromium's GPU context is EGL, not GLX, so it flips `glxinfo` without moving the actual renderer); `__EGL_VENDOR_LIBRARY_FILENAMES` (pointed at the NVIDIA glvnd EGL ICD) is the variable that verifiably moves the unmasked WebGL renderer to the discrete adapter. The relaunch also appends `--ozone-platform=x11` to argv (never overriding a player's own `--ozone-platform` choice), since the same hardware showed the GPU process crash-looping on Wayland once PRIME offload is requested, falling back to software rendering.
- `DRI_PRIME=1` (Mesa AMD/Intel PRIME offload) is unaffected by any of the above and still applies.
- Documents the corrected lever, and two known follow-ups (the relaunch isn't yet gated on detecting a hybrid adapter first, and its interaction with the second-instance deep-link login path is unverified) in `docs/desktop-release.md`'s GPU verification checklist.

## Why this is the right fix
This repo's `IWorld`/sim architecture is unaffected (this is a desktop-shell-only change, no `src/sim/` or `src/render/` touch). The web client (played in a browser, not the Electron shell) already requests `powerPreference: 'high-performance'` in its `WebGLRenderer` (`src/render/renderer.ts`), which is the only lever a web page has; true adapter switching on Linux for a browser tab requires OS/browser-level configuration outside this repo's control (`prime-run`, `DRI_PRIME`, or the distro's GPU-switching tool), so this PR is scoped to what the codebase can actually fix: the packaged desktop app.

## Test plan
- [x] `npx vitest run tests/electron_gpu_preference.test.ts` - 62/62 passing (new cases cover `shouldRelaunchForLinuxPrime`, `relaunchForLinuxPrime`'s spawn/env/argv/loop-guard/error-handling behavior, and the corrected `buildLinuxPrimeEnv` variable set).
- [x] `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts` - sim purity and i18n matcher guards green.
- [x] `npx tsc --noEmit -p .` - clean.
- [x] `npx biome check` on every file this PR touches - clean.
- [x] `git merge-tree` against `up/release/v0.27.0` - clean merge, no conflicts.
- Not applicable: no visual/UI change (Electron main-process startup logic only), so no before/after screenshots.

### DIFF
--- a/docs/desktop-release.md
+++ b/docs/desktop-release.md
@@ -447,6 +447,15 @@ Rules that keep this working:
    the working lever there; Windows 11 with an attached eGPU can ignore the per-app
    preference while the eGPU is connected; pre-1803 Windows 10 ignores the key
    entirely.
+   On Linux, hybrid-graphics laptops (NVIDIA Optimus, AMD/Intel Mesa PRIME) have no
+   per-app OS preference and the Chromium switch is a no-op (the GPU adapter is
+   resolved by the driver's client library at dynamic-link time, before Chromium
+   parses its switches), so `gpu_preference.cjs` sets the standard PRIME
+   render-offload environment variables instead: `DRI_PRIME=1` (Mesa) and
+   `__NV_PRIME_RENDER_OFFLOAD=1` / `__GLX_VENDOR_LIBRARY_NAME=nvidia` /
+   `__VK_LAYER_NV_optimus=NVIDIA_only` (NVIDIA proprietary driver), applied on every
+   launch (packaged or dev) and never overriding a name the player's own environment
+   already set (e.g. their own `prime-run` wrapper).
 3. Login both paths: email/password in-app, and Discord via the external browser +
    `worldofclaudecraft://desktop-login` deep link handoff (app focuses and enters
    the world; second-instance and cold-start deep links both work).
--- a/docs/desktop-release.md
+++ b/docs/desktop-release.md
@@ -450,12 +450,43 @@ Rules that keep this working:
    On Linux, hybrid-graphics laptops (NVIDIA Optimus, AMD/Intel Mesa PRIME) have no
    per-app OS preference and the Chromium switch is a no-op (the GPU adapter is
    resolved by the driver's client library at dynamic-link time, before Chromium
-   parses its switches), so `gpu_preference.cjs` sets the standard PRIME
-   render-offload environment variables instead: `DRI_PRIME=1` (Mesa) and
-   `__NV_PRIME_RENDER_OFFLOAD=1` / `__GLX_VENDOR_LIBRARY_NAME=nvidia` /
-   `__VK_LAYER_NV_optimus=NVIDIA_only` (NVIDIA proprietary driver), applied on every
-   launch (packaged or dev) and never overriding a name the player's own environment
-   already set (e.g. their own `prime-run` wrapper).
+   parses its switches). Setting the PRIME render-offload environment variables
+   (`DRI_PRIME=1` for Mesa; `__NV_PRIME_RENDER_OFFLOAD=1` /
+   `__GLX_VENDOR_LIBRARY_NAME=nvidia` / `__EGL_VENDOR_LIBRARY_FILENAMES=<nvidia glvnd
+   EGL ICD json>` / `__VK_LAYER_NV_optimus=NVIDIA_only` for the NVIDIA proprietary
+   driver) in the running main process does NOT reach the GPU process either:
+   Electron's Linux GPU process forks from a zygote that already exec'd (and
+   snapshotted its environ) before any main-process JS runs, so a process.env write
+   there is invisible to it. Verified on real hybrid hardware: `__GLX_VENDOR_LIBRARY_NAME`
+   alone is a decoy for this app (Chromium's context is EGL, not GLX, so it flips
+   `glxinfo` while the unmasked WebGL renderer stays on the iGPU); the
+   `__EGL_VENDOR_LIBRARY_FILENAMES` variable is the one that actually moves the
+   renderer to the NVIDIA adapter. The same hardware also crash-loops the GPU
+   process on a Wayland session once PRIME offload is requested (falls back to
+   software rendering, worse than the iGPU), so Chromium must additionally be
+   forced onto the X11 Ozone backend, which (like the env vars) only works as a
+   real argv flag present before Electron's own startup, never an `appendSwitch`
+   call in the running process.
+   `main.cjs` instead calls `relaunchForLinuxPrime` as the very first thing it does
+   (before crash reporting, logging, or any window): on Linux, if any PRIME
+   variable is missing from the environment, it re-execs the app with them baked
+   into the new process's environment from birth plus `--ozone-platform=x11`
+   appended to argv (a marker env var prevents a relaunch loop; an explicit player
+   `--ozone-platform` choice is never overridden), and the original process exits
+   immediately. Applied on every launch (packaged or dev), and never overriding an
+   env name the player's own environment already set (e.g. their own `prime-run`
+   wrapper) -- if every variable is already present, no relaunch happens at all.
+   Verify with `ps -o pid,ppid,cmd -C world-of-claudecraft` (or the AppImage/binary
+   name) showing the relaunched PID's parent already exited, and `[gpu] relaunching
+   for Linux PRIME render offload` in `main.log`.
+   Known follow-ups, not yet addressed: the relaunch is unconditional rather than
+   gated on detecting a hybrid adapter first (Electron's `getGPUInfo` needs `app`
+   ready, which is after the point a relaunch must decide), so a non-hybrid Linux
+   machine still pays for one extra process spawn per launch (harmless today since
+   the vars sit inert on that hardware, but worth tightening); and the relaunch's
+   interaction with the second-instance deep-link path (`worldofclaudecraft://`
+   login handoff) has not been verified against a login link that arrives during
+   the brief relaunch window.
 3. Login both paths: email/password in-app, and Discord via the external browser +
    `worldofclaudecraft://desktop-login` deep link handoff (app focuses and enters
    the world; second-instance and cold-start deep links both work).
--- a/docs/desktop-release.md
+++ b/docs/desktop-release.md
@@ -457,36 +457,49 @@ Rules that keep this working:
    driver) in the running main process does NOT reach the GPU process either:
    Electron's Linux GPU process forks from a zygote that already exec'd (and
    snapshotted its environ) before any main-process JS runs, so a process.env write
-   there is invisible to it. Verified on real hybrid hardware: `__GLX_VENDOR_LIBRARY_NAME`
-   alone is a decoy for this app (Chromium's context is EGL, not GLX, so it flips
-   `glxinfo` while the unmasked WebGL renderer stays on the iGPU); the
-   `__EGL_VENDOR_LIBRARY_FILENAMES` variable is the one that actually moves the
-   renderer to the NVIDIA adapter. The same hardware also crash-loops the GPU
-   process on a Wayland session once PRIME offload is requested (falls back to
-   software rendering, worse than the iGPU), so Chromium must additionally be
-   forced onto the X11 Ozone backend, which (like the env vars) only works as a
-   real argv flag present before Electron's own startup, never an `appendSwitch`
-   call in the running process.
-   `main.cjs` instead calls `relaunchForLinuxPrime` as the very first thing it does
-   (before crash reporting, logging, or any window): on Linux, if any PRIME
-   variable is missing from the environment, it re-execs the app with them baked
-   into the new process's environment from birth plus `--ozone-platform=x11`
-   appended to argv (a marker env var prevents a relaunch loop; an explicit player
-   `--ozone-platform` choice is never overridden), and the original process exits
-   immediately. Applied on every launch (packaged or dev), and never overriding an
-   env name the player's own environment already set (e.g. their own `prime-run`
-   wrapper) -- if every variable is already present, no relaunch happens at all.
+   there is invisible to it. Which variable does the lifting depends on the path
+   Chromium takes (both measured on real hybrid hardware): under the
+   `--ozone-platform=x11` backend the shell forces, ANGLE binds through GLX and the
+   `__NV_PRIME_RENDER_OFFLOAD` + `__GLX_VENDOR_LIBRARY_NAME` pair selects the
+   adapter; on EGL paths (a player-forced Wayland ozone) that pair is a decoy (it
+   flips `glxinfo` while the unmasked renderer stays on the iGPU) and
+   `__EGL_VENDOR_LIBRARY_FILENAMES` is the lever. The EGL entry is only ever set
+   when the NVIDIA ICD json actually exists, because glvnd treats it as a
+   replacement of its vendor list and naming a missing file would leave a
+   non-NVIDIA machine with no EGL vendors at all. The same hardware also
+   crash-loops the GPU process on a Wayland session once PRIME offload is
+   requested (falls back to software rendering, worse than the iGPU), so Chromium
+   must additionally be forced onto the X11 Ozone backend, which (like the env
+   vars) only works as a real argv flag present before Electron's own startup,
+   never an `appendSwitch` call in the running process.
+   `main.cjs` therefore calls `relaunchForLinuxPrime` as the very first thing it
+   does (before crash reporting, logging, or any window): on a HYBRID Linux
+   machine (two or more GPUs under `/sys/class/drm`; single-GPU machines are left
+   completely untouched), it re-execs the app with the PRIME variables baked into
+   the new process's environment from birth plus `--ozone-platform=x11` appended
+   to argv, and the original process exits immediately. The spawn source is
+   `$APPIMAGE` (the outer AppImage file, the same source electron-updater restarts
+   from) when set, because inside an AppImage `process.execPath` lives in a FUSE
+   mount that dies with the exiting parent; other installs re-exec the binary
+   itself. An explicit player `--ozone-platform` choice is never overridden (a
+   bare `--ozone-platform-hint` deliberately does not count), and no env name the
+   player's own environment already set (their own `prime-run` wrapper) is ever
+   replaced: with no marker and every variable present, no relaunch happens at
+   all. The relaunch marker does NOT permanently suppress re-execs: a marked
+   process whose argv lost the ozone flag (electron-updater's restart-to-update
+   respawns with the current env but empty argv) relaunches once more purely to
+   restore the flag. The dev loop (`npm run electron:dev`) pre-applies the same
+   configuration to the electron it spawns, so no relaunch happens there and the
+   Vite teardown-on-exit logic keeps working.
    Verify with `ps -o pid,ppid,cmd -C world-of-claudecraft` (or the AppImage/binary
-   name) showing the relaunched PID's parent already exited, and `[gpu] relaunching
-   for Linux PRIME render offload` in `main.log`.
-   Known follow-ups, not yet addressed: the relaunch is unconditional rather than
-   gated on detecting a hybrid adapter first (Electron's `getGPUInfo` needs `app`
-   ready, which is after the point a relaunch must decide), so a non-hybrid Linux
-   machine still pays for one extra process spawn per launch (harmless today since
-   the vars sit inert on that hardware, but worth tightening); and the relaunch's
-   interaction with the second-instance deep-link path (`worldofclaudecraft://`
-   login handoff) has not been verified against a login link that arrives during
-   the brief relaunch window.
+   name) showing the relaunched PID's parent already exited, and `[gpu] running as
+   PRIME-relaunched child` in `main.log` (the child writes it; the parent exits
+   before file logging exists).
+   Known follow-ups, not yet addressed: the relaunch's interaction with the
+   second-instance deep-link path (`worldofclaudecraft://` login handoff) has not
+   been verified against a login link that arrives during the brief relaunch
+   window, and the Steam channel's process tracking (overlay, playtime) has not
+   been verified against the parent exiting within milliseconds of launch.
 3. Login both paths: email/password in-app, and Discord via the external browser +
    `worldofclaudecraft://desktop-login` deep link handoff (app focuses and enters
    the world; second-instance and cold-start deep links both work).
--- a/electron/gpu_preference.cjs
+++ b/electron/gpu_preference.cjs
@@ -55,6 +55,29 @@ const nodePath = require('node:path');
 //     binary (one orphan entry per worktree, steering everything else launched from that
 //     shared binary); only the stable installed exe path is worth pinning.
 //
+//  3. Linux PRIME render-offload environment variables. On a Linux hybrid-graphics laptop
+//     (NVIDIA Optimus, AMD/Intel Mesa PRIME) there is no per-app OS-level preference to write
+//     (no registry equivalent) and force-high-performance-gpu is a no-op: the driver decides
+//     which adapter serves EGL/GLX at DYNAMIC-LINK time, before Chromium's own switch parsing
+//     ever runs. The standard offload contract (what `prime-run`/`optirun` wrapper scripts
+//     set) is a handful of environment variables read by the vendor's client driver when it
+//     resolves the GPU vendor library: DRI_PRIME=1 (Mesa: AMD/Intel PRIME offload, ignored by
+//     the NVIDIA proprietary driver), and for NVIDIA's proprietary PRIME render offload,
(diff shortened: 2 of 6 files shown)


FernandoX7 (commented):
The craftsmanship here is the good kind: a pure `buildLinuxPrimeEnv` behind the existing DI seam, the `.d.cts` kept in sync, a never-override guard for players who already run their own `prime-run`, 10 focused new tests (53/53 pass for me locally), a clean merge against the base, and a correct diagnosis of why `force-high-performance-gpu` no-ops on Linux. Before reviewing the prose I ran the mechanism on real hybrid hardware (Intel ARL iGPU + NVIDIA proprietary 580 dGPU, Electron 43, the version this repo ships), and unfortunately the central claim does not hold up.

**blocking** electron/gpu_preference.cjs:260 (and the lever-3 comment at line 72): a `process.env` mutation at this call site never reaches the GPU process, so the fix is a functional no-op on Linux. The comment asserts the GPU process is "a fresh fork+exec, not a fork of an already-initialized zygote", and it is the opposite. Measured:

- Both Chromium zygotes already exist as children of the browser process at the first executable line of the main script (verified by scanning `/proc` for `--type=zygote` children by ppid at top of a probe main, before anything else ran).
- The GPU process's parent is the unsandboxed zygote (`--type=zygote --no-zygote-sandbox`), so it forks from an environment snapshot the zygote took at exec, before the main script ran. Chromium's zygote fork protocol passes argv and fds, not the browser's live environment.
- Controlled A/B with a variable that provably propagates: `__EGL_VENDOR_LIBRARY_FILENAMES=10_nvidia.json` set in the parent environment changes GPU-process behavior dramatically (GL init fails outright on this Wayland session). The same variable set at exactly this call site (main top level, before `whenReady`) has no effect at all: GL initializes normally on the iGPU. The same top-level A/B with `LIBGL_ALWAYS_SOFTWARE=1` and with this PR's four vars likewise produces zero behavioral difference from control.

One aside from the same evidence: the processes that DO inherit the mutation are the ones the main process spawns later, for example the external login browser via `shell.openExternal` (electron/main.cjs:265), which is the inverse of the intent. That part is harmless in practice (modern glvnd falls back gracefully when the named GLX vendor is not installed; verified with a bogus vendor name), but it is the only observable effect of the change as written.

**should-fix** (it shapes the redo): even in the placement that does propagate (the parent environment), the four chosen vars did not move Chromium off the iGPU on this machine. With all four set at launch, the unmasked WebGL renderer stays "ANGLE (Intel, Mesa ...)", both under Wayland and with `--ozone-platform=x11`. Chromium resolves its adapter through EGL, where the GLX offload trio has no effect, and `DRI_PRIME` steers Mesa-driven dGPUs only, so the NVIDIA proprietary case (the one the PR names first) is exactly where this lever came up empty. One machine is not a universe, and a Mesa-hybrid AMD laptop may well honor `DRI_PRIME` here, but the NVIDIA path needs a demonstrated-working lever (EGL vendor selection is the one that visibly reaches the NVIDIA stack, though in my run it broke GL rather than offloading, so it needs real-hardware verification before any claim).

**nit** tests/electron_gpu_preference.test.ts:301: the new tests pin that the env object gets mutated, and they are well built, but nothing can pin the load-bearing half (that the mutation reaches the GPU process), and per the above it does not. That is the structural tell: this lever only works somewhere that runs before the browser process exists.

Directions that would actually work: set the env before the Electron binary starts (a small launcher script that execs the real binary, the `.desktop` Exec line, or Steam launch options for the Steam build), or a one-shot detect-and-relaunch at startup (spawn `process.execPath` with the augmented env plus a loop-guard marker, then quit). Either way, the docs/desktop-release.md:450 addition currently documents behavior the code does not deliver; it should land only together with a fix demonstrated on real hybrid hardware.

FernandoX7:
Follow-up with hardware-verified data, since a real hybrid machine was available (Intel Arrow Lake iGPU + NVIDIA RTX 5090 Laptop, NVIDIA 580 proprietary driver, Pop!_OS Wayland session, Electron 43, the version this repo ships).

A configuration that demonstrably moves the packaged app onto the discrete GPU exists, and it is close to this PR's var set, with two decisive differences:

```
env __NV_PRIME_RENDER_OFFLOAD=1 \
    __GLX_VENDOR_LIBRARY_NAME=nvidia \
    __EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/10_nvidia.json \
    __VK_LAYER_NV_optimus=NVIDIA_only \
    world-of-claudecraft.AppImage --ozone-platform=x11
```

Verified by the unmasked WebGL renderer read from the live GPU process: "ANGLE (NVIDIA Corporation, NVIDIA GeForce RTX 5090 Laptop GPU ...)". Removing pieces isolates what does the work:

- This PR's exact var set (the GLX trio plus DRI_PRIME, without `__EGL_VENDOR_LIBRARY_FILENAMES`), correctly placed in the parent environment: the renderer stays Intel. Chromium creates its context through EGL, so `__GLX_VENDOR_LIBRARY_NAME` is a decoy lever: it flips `glxinfo` to NVIDIA, which makes a fix look verified while the app stays on the iGPU. The EGL vendor line is the one that matters for the NVIDIA proprietary path. (`DRI_PRIME` is still right for the Mesa-hybrid AMD case; it just does not cover NVIDIA.)
- Without `--ozone-platform=x11`, the Wayland session's GPU process crash-loops (`exit_code=8704`, "not compatible with Vulkan") and the app silently lands in software rendering, which is worse than the iGPU it started on. An env-only fix is not safe on Wayland.
- A desktop-entry `PrefersNonDefaultGPU=true` gets switcheroo-control to inject the NVIDIA environment, but at least on current Pop!_OS that injected set is only the GLX trio without the EGL var, so it does not fix Chromium either; and direct `./world-of-claudecraft.AppImage` runs bypass desktop entries entirely.

Placement, also measured (this closes the appendSwitch escape hatch next to the review's blocking finding): `app.commandLine.appendSwitch('ozone-platform', 'x11')` at main-script top level does not take effect either; the ozone platform is selected before any main-script JS runs. No part of the working recipe can land as a same-process mutation in main.cjs.

Two shapes that do work:

1. Relaunch. On first launch (loop-guard marker absent), spawn `process.execPath` with the recipe env plus `--ozone-platform=x11` plus the marker, detached, then quit. I ran exactly this shape and the relaunched child rendered on the discrete GPU. A production version needs hybrid detection before deciding to relaunch (the device list `summarizeGpuDevices` already parses is a candidate signal), a crash-safe fallback so a machine where offload misbehaves does not end up in software rendering, and care around the deep-link second-instance path so a relaunch does not eat a `worldofclaudecraft://` login handoff.
2. Packaging-side. No Linux launch path has a wrapper today: the AppImage's AppRun execs the binary directly, the Steam depot's documented launch option is the bare executable, and the deb ships the stock desktop file. electron-builder's `linux.desktop` and `executableArgs` keys (or a channel override in `scripts/electron-builder-config.mjs`, which already rewrites `config.linux` for the Steam channel) could carry the flag and env for launcher-mediated launches, but only the relaunch shape also covers direct AppImage executions.

Either way, docs/desktop-release.md's checklist stance that a hybrid machine reporting the integrated adapter is "a regression, not a user misconfiguration" only becomes true on Linux once one of these lands with the EGL var included.

jgyy:
Thanks for both passes, and especially for pulling real hybrid hardware into this - the zygote-fork explanation on the first review was exactly right and the follow-up's EGL-vs-GLX distinction would have shipped a fix that looked verified (via `glxinfo`) while doing nothing for the actual renderer.

Implemented shape 1 from the follow-up: `relaunchForLinuxPrime` in `electron/gpu_preference.cjs`, called as the very first thing `main.cjs` does (before crash reporting, logging, or any window). On Linux, if any PRIME variable is missing it re-execs the app with the corrected variable set baked into the child's environment from birth, plus `--ozone-platform=x11` appended to argv (skipped if the player already named an `--ozone-platform`), then the original process exits via `process.exit(0)` immediately - no window, no GPU process spawned in the process that's about to die. A marker env var guards the loop; if every variable is already present, no relaunch happens at all.

Variable set corrected per the hardware findings: added `__EGL_VENDOR_LIBRARY_FILENAMES` (the one that actually moved the unmasked WebGL renderer to the discrete adapter in your test), kept `__GLX_VENDOR_LIBRARY_NAME` (harmless, and it's what a Mesa-hybrid AMD/Intel `DRI_PRIME` path still needs) rather than dropping it outright, since removing it isn't necessary and it costs nothing to leave in.

Not addressed, called out explicitly in the PR body and in `docs/desktop-release.md` rather than left implicit:
- The relaunch is unconditional on Linux rather than gated on detecting a hybrid adapter first. `getGPUInfo()` needs `app` ready, which is after the point this decision has to be made, so a real hybrid-detection gate would need a different shape (maybe a cheap `lspci`/sysfs probe before Electron even starts) - flagging it rather than pretending the current version has it.
- Whether a `worldofclaudecraft://` login deep-link arriving during the brief relaunch window is handled correctly is unverified; I don't have a way to test the second-instance path against a real relaunch without your hardware.

Updated the PR body/description to reflect the relaunch shape instead of the in-process mutation. Pushed as a new commit on the same branch; 62/62 tests green including new coverage for the relaunch decision, spawn args/env, the ozone-platform argv injection, and the loop guard.

FernandoX7 (commented):
Re-review of the relaunch rework (9531ef5d8), again against real hybrid hardware (Intel Arrow Lake iGPU + RTX 5090, NVIDIA 580, Electron 43). The mechanism is now the right one, and I verified it end to end: I ran this branch's actual `relaunchForLinuxPrime` from a minimal main, the parent relaunched, the child arrived with the marker, the appended ozone flag, and the full var set, and its unmasked renderer came out "ANGLE (NVIDIA ... RTX 5090 ...)". 62/62 tests pass, tsc is clean, the merge against the base is clean, and the commit message and docs are honest about what is and is not covered. The remaining problems are launch-chain integration: the relaunch is correct in isolation but breaks the two packaged Linux flows it most needs to survive.

**blocking** electron/gpu_preference.cjs:295 (`spawnFn(execPath, ...)`): the relaunch kills the packaged AppImage, which docs/desktop-release.md calls the auto-updatable target the website offers. In an AppImage, `process.execPath` points inside the runtime's FUSE mount, and the mount's daemon is the AppImage runtime process, which unmounts and exits as soon as its child (the original main process) exits. I demonstrated this with the real installed AppImage: mount it, hold an open fd on a file inside, kill the runtime; the open fd immediately returns "Input/output error / Transport endpoint is not connected" and the mountpoint is gone. The relaunched child loses its own binary and resources mid-startup, so for AppImage users the app does not merely stay on the iGPU, it fails to launch at all. The established fix is the one electron-updater itself uses (node_modules/electron-updater/out/AppImageUpdater.js keys everything on `process.env.APPIMAGE`): spawn `env.APPIMAGE` when it is set, falling back to `execPath` for deb, Steam dir, and dev runs, with a test arm pinning the selection.

**blocking** (interaction with electron/updater.cjs:145): the restart-to-update flow resurrects the exact Wayland crash-loop this PR documents. `quitAndInstall()` is non-silent, so electron-updater force-runs the new AppImage (`autoRunAppAfterInstall` defaults true) via `AppImageUpdater.doInstall`, which spawns it with `args: []` and `env: { ...process.env, APPIMAGE_SILENT_INSTALL: 'true' }`. The process doing that spawn is the relaunched child, so the updated app inherits `WOC_PRIME_RELAUNCHED=1` plus all five PRIME vars, skips the relaunch (marker check, and every var is present anyway), and runs with PRIME active but WITHOUT `--ozone-platform=x11`: on a Wayland session that is the exit_code=8704 crash-loop into software rendering, now reproducing after every "restart to update" until the player manually relaunches. Structurally: the env half of the fix propagates by inheritance, the argv half exists only inside the relaunch, and the marker suppresses the only code that could re-add it. The skip condition needs to consider the argv half too (all vars present AND the platform choice already made), not just env completeness.

**blocking** tests/electron_gpu_preference.test.ts: the load-bearing wire tokens of this fix are never pinned to literals, and the new main.cjs wiring is not pinned at all. Concretely: `LINUX_OZONE_X11_ARG` is only ever asserted against itself (tests/electron_gpu_preference.test.ts:386), so changing the constant to `wayland`, which reintroduces the crash-loop this flag exists to prevent, keeps all 62 tests green; the string `--ozone-platform=x11` appears in a test title but no assertion. Same self-comparison for `__EGL_VENDOR_LIBRARY_FILENAMES` (the json path appears nowhere in the test file) and `__GLX_VENDOR_LIBRARY_NAME`. And the wiring-pin block (tests/electron_gpu_preference.test.ts:683 area) still pins only the old `forceHighPerformanceGpu({ app, log })` call: nothing fails if `if (relaunchForLinuxPrime({ app })) { process.exit(0); }` moves inside `whenReady`, loses the exit, or slips below the logging/window setup, any of which silently kills the fix. Cheap to close: literal pins for the three constants, plus a source-order pin for the relaunch guard mirroring the existing one.

**should-fix** electron/gpu_preference.cjs:215 (`__EGL_VENDOR_LIBRARY_FILENAMES` in `LINUX_PRIME_ENV`): setting this unconditionally is a latent breaker on machines without the NVIDIA driver, because glvnd treats it as a REPLACEMENT for its vendor list, not an addition: if the named json does not exist there are zero EGL vendors and EGL is dead entirely (verified: `eglinfo` fails to initialize, and an Electron run on the EGL path yields no GL context at all). The shipped x11 flow survives that because ANGLE routes through GLX there (next paragraph), but the two ways the app can end up on an EGL path make it real: a player-supplied `--ozone-platform=wayland`, which this PR explicitly honors, and the post-update no-flag state above, which needs no player action at all. On a Mesa-only machine either path goes from working GL before this PR to none after it. Gate the entry on `fs.existsSync` of the json (injectable for tests), or drop it.

While isolating that, I have to correct a claim from my earlier follow-up comment that this commit's comments and docs faithfully repeat: "the EGL vendor line is the one that actually moves the renderer" was measured on the default (Wayland) path. Under the `--ozone-platform=x11` flow this PR ships, ANGLE binds through GLX (the renderer strings flip from "OpenGL ES" to desktop "OpenGL"), and there the GLX pair does the moving: this branch's var set minus the EGL line still lands on the NVIDIA GPU, and the EGL line pointed at a nonexistent file still lands on the NVIDIA GPU. I also re-checked the non-NVIDIA case under x11 (bogus GLX vendor plus missing EGL json): glvnd falls back gracefully and the machine keeps a hardware Mesa context. So the EGL var is belt-and-suspenders on the x11 path and load-bearing only on EGL paths; the comment at electron/gpu_preference.cjs:200 and the docs/desktop-release.md:463 wording should say that rather than crediting it as the operative lever.

**should-fix** scripts/electron-dev.mjs:66 (unchanged by this PR, but broken by it): the dev orchestrator does `electron.on('exit', (code) => stopAll(code ?? 0))`, and stopAll kills Vite and exits. On Linux the parent electron now exits immediately after relaunching, so every `npm run electron:dev` tears down the dev server and leaves the detached, no-longer-killable child pointing at a dead `VITE_DEV_SERVER_URL`. Simplest fix: have electron-dev.mjs set the full PRIME env plus the ozone flag (or at minimum the `WOC_PRIME_RELAUNCHED` marker) in the env it spawns electron with; that both suppresses the relaunch and gives Linux dev runs the offload for real.

**should-fix** electron/gpu_preference.cjs:275 (relaunch policy): the relaunch is unconditional on Linux, so every player, including single-GPU machines, pays an extra spawn and gets `--ozone-platform=x11` forced: Wayland players silently move to XWayland, and a Wayland-only setup without XWayland stops launching entirely. The docs list hybrid-gating as a follow-up on the grounds that `getGPUInfo` needs app ready, but a ready-free gate exists: enumerate `/sys/class/drm/card*/device/vendor` and only relaunch when more than one distinct GPU vendor is present (synchronous, no Electron dependency, injectable for tests). Two related edges of the ozone guard worth deciding deliberately: `startsWith('--ozone-platform')` also matches `--ozone-platform-hint`, so a wrapper passing only `--ozone-platform-hint=auto` suppresses the x11 append and re-exposes the Wayland crash-loop; and electron-builder's `linux.desktop` key could carry `PrefersNonDefaultGPU=true`, which makes desktop-portal launchers (switcheroo-control) inject their dGPU env natively; on the machine I tested, the launcher otherwise injects a `DRI_PRIME=<iGPU>` pin that this module's never-override posture then respects, quietly opting desktop-launcher players on Mesa hybrids out of the offload.

**should-fix** (verification, no code change yet): the Steam channel ships as a bare executable launch option (docs/desktop-release.md Steam section), and a detached self-relaunch whose launched process exits within milliseconds is a classic way to confuse Steam's process tracking (Steam's reaper usually re-tracks reparented descendants, so overlay and playtime probably survive, but nothing verifies it, and the Steam Deck is a Mesa-only machine where the EGL entry above applies in full). If it does misbehave, the runtime already knows its channel (`wocDesktop.distribution`), so the relaunch can be skipped for the Steam distribution and the env carried by the depot's launch configuration instead.

**nit** electron/main.cjs:58: `relaunchForLinuxPrime({ app })` passes a property the deps type does not declare and the function never reads (uncaught because tsconfig's include list omits electron/, so main.cjs is not type-checked); the JSDoc at electron/gpu_preference.cjs:270 tells the caller to exit via `app.exit()` while the call site's comment argues the opposite and uses `process.exit(0)`; and since no `log` is passed (nothing is initialized yet, by design), the relaunch and any spawn failure are observable nowhere, while docs/desktop-release.md:480 tells the verifier to look for "[gpu] relaunching for Linux PRIME render offload" in main.log, a line that cannot currently appear there. Cleanest: have the CHILD log a "running as PRIME-relaunched child" line once logging is up (the marker is already in its env), and point the docs at that.

**nit** tests/electron_gpu_preference.test.ts:271: "offers all four PRIME offload variables" (and the lever-3 comment's "All four") is now five; the player-override arm only exercises `--ozone-platform=wayland`, so the `--ozone-platform-hint` prefix arm of the guard is untested; and the spawn options assertion uses `toMatchObject` where the sibling reg-exec assertions pin the full options object with `toEqual`.

On CI: the red "Screenshots of changes" job was cancelled during dependency install (superseded run), not a content failure; a re-run should clear it.

The core mechanism is now genuinely right, and this is the first version that can work at all. The must-haves before merge: spawn source (`APPIMAGE`), the post-update skip condition, and the literal pins; the dev-loop and EGL gating are close behind.

FernandoX7:
I pushed a hardening commit addressing every finding from the re-review, verified end to end on the same hybrid hardware (Intel Arrow Lake iGPU + RTX 5090, NVIDIA 580, Electron 43). What changed:

- **AppImage-safe spawn source** (was blocking): `relaunchForLinuxPrime` now spawns `env.APPIMAGE` when it is set (with an absolute-path guard), falling back to `process.execPath` for deb, Steam dir, and dev runs. Inside an AppImage, execPath lives in the runtime's FUSE mount, which dies the instant the parent exits, so the previous spawn could never survive there; the outer file is the same source electron-updater restarts from.
- **The skip condition now covers the argv half** (was blocking): electron-updater's restart-to-update respawns the app with the current env (marker and PRIME vars included) but empty argv, which previously left the updated process running PRIME without `--ozone-platform=x11`, the documented Wayland crash-loop. A marked process whose argv lacks an explicit ozone choice now relaunches once purely to restore the flag. This cannot loop: every relaunch produces a child whose argv carries the explicit flag. Both the fresh-launch and the simulated post-update state were run end to end on real hardware; both land the child on the discrete GPU.
- **Literal test pins and wiring pins** (was blocking): `LINUX_PRIME_ENV` (all five values), `LINUX_OZONE_X11_ARG`, and `PRIME_RELAUNCH_MARKER` are pinned to literals in the constants block; the main.cjs guard (module scope, ordered before `forceHighPerformanceGpu` and `initLogging`, `process.exit(0)` as the block body) and the electron-dev.mjs pre-apply are pinned textually, mirroring the existing wiring-pin block. The spawn options assertion is a full `toEqual`.
- **Hybrid gating**: the relaunch only fires when `/sys/class/drm` exposes two or more card devices (readable synchronously long before app ready, so the docs' getGPUInfo objection no longer applies). Single-GPU machines get neither the extra spawn nor a forced X11 backend, and an unreadable /sys defaults to doing nothing.
- **EGL vendor entry existence-gated**: `__EGL_VENDOR_LIBRARY_FILENAMES` is only set when the NVIDIA ICD json actually exists, since glvnd treats it as a replacement of its vendor list and naming a missing file leaves a non-NVIDIA machine with zero EGL vendors. The comments and docs now also credit the right lever per path: under the forced x11 backend the GLX pair selects the adapter; the EGL entry is load-bearing on EGL paths.
- **Dev loop fixed**: `scripts/electron-dev.mjs` pre-applies the exact configuration the relaunch would have produced (env additions, marker, explicit flag) using the module's own predicates, so the relaunch never fires under `npm run electron:dev` and the orchestrator's Vite teardown-on-exit keeps working, while Linux dev runs get the offload for real.
- **Observability**: the relaunch decision logs to the console (file logging cannot exist yet in the exiting parent), and the relaunched child writes `[gpu] running as PRIME-relaunched child` to main.log once logging is up; the docs verification checklist now points at that line, which can actually appear. The dead `{ app }` parameter is gone and the contradictory exit-strategy comments are reconciled.
- **Ozone predicate narrowed**: only an explicit `--ozone-platform` argv choice suppresses the appended flag; a bare `--ozone-platform-hint` no longer does (honoring a wayland hint while injecting PRIME would recreate the crash-loop).

Verification: 80/80 in tests/electron_gpu_preference.test.ts plus the neighboring desktop suites, tsc clean, biome clean on the changed files, architecture and i18n guard tests green, and the two hardware runs above. Still open, and now the only items docs list as follow-ups: live verification of the second-instance deep-link during the relaunch window, and of Steam's process tracking on the Steam channel.

FernandoX7:
Live pre-merge verification on real hardware: built the x64 AppImage from this branch head (bb0e651a1), installed it fresh on the hybrid test machine (Intel Arrow Lake iGPU + RTX 5090, NVIDIA 580, Wayland session) with a completely stock desktop entry: no env vars, no flags, no PrefersNonDefaultGPU, and a fresh profile.

Result, with zero user configuration:

- The relaunch fired exactly as designed: the running main process carries `--ozone-platform=x11` it was never launched with, its parent has exited, and main.log opens with `[gpu] running as PRIME-relaunched child (Linux discrete-GPU offload env active)`.
- main.log's `[gpu] active renderer` shows the NVIDIA adapter (0x10de/0x2c58) active: true and the Intel adapter (0x8086) active: false, and nvidia-smi shows the game's GPU process holding over 1.2 GiB of VRAM on the discrete GPU.

One observed artifact, documented for the record rather than fixed: the first launch's AppImage FUSE mount stays mounted (with its runtime daemon) for the duration of the session, because Chromium deliberately keeps a few fds inheritable (icudtl.dat, v8_context_snapshot.bin, resources/app.asar) and the relaunched process inherits them pointing into the old mount. The content of both mounts is identical, the fds are unused by the fresh boot, and the hold releases when the game exits. electron-updater's own restart-after-update spawns the same way and has the same artifact, so this is standard AppImage-relaunch behavior, not a regression; flagging it here so nobody rediscovers it as a mystery later.

FernandoX7:
Closing note before merge, completing the live-install record: after exiting the game on the test machine, every process (both AppImage runtime daemons included) and both FUSE mounts cleaned up on their own, so the inherited-fd mount pinning noted above is confirmed to fully self-heal at exit. Player-visible in-game confirmation also came back positive: the performance overlay reports the NVIDIA GPU on the stock install.

Merging on this basis: hardware-verified fresh launch, post-update restart, and full-session lifecycle, all landing on the discrete GPU with zero user configuration; CI green on every substantive check (the red "Screenshots of changes" job is the pre-existing cancelled helper with nothing to embed for a desktop-only change). Two items stay tracked in docs/desktop-release.md's verification checklist rather than blocking: live verification of the second-instance deep-link during the relaunch window, and Steam channel process tracking. Thanks jgyy for carrying the rework through.

PR #2078, opened 2026-07-17
fix(render): wire up the missing Training Dummy lazy preload

jamiecypher:
## Summary

Teleporting near the zone3 ogre war-camp permanently froze the client. Root cause: `training_dummy` (`src/sim/content/zone3.ts`) is `lazyPreload: true`, since it appears in exactly one hub, the same pattern the Combat Mech uses. Unlike the mech, nothing ever triggered its lazy load: `Renderer.createView` called `resolvedGltf()` directly and threw `character asset not preloaded` every single frame once a dummy became a view candidate, permanently stalling `Renderer.sync()`. The sim tick and audio kept running on a separate per-frame path, which is why it looked like only the screen had frozen while sound kept playing.

Fix: adds `preloadTrainingDummyAssets`/`trainingDummyAssetsReady` (`src/render/characters/assets.ts`), mirroring `preloadMechAssets`/`mechAssetsReady` exactly, and gates `Renderer.createView` the same way the mech is gated: trigger the load and defer that entity's view for a frame instead of crashing.

Confirmed pre-existing and unrelated to any in-flight feature work: reproduced against a baseline pinned to the exact commit an unrelated feature branch had diverged from, with zero unrelated code present. Full reproduction and root-cause writeup in `docs/training-dummy-preload-freeze-postmortem.md` (included in this PR).

## Related issues

N/A

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: `npx tsc --noEmit`, `npx vitest run tests/training_dummy_preload.test.ts tests/render_asset_preload.test.ts tests/render_asset_fallback.test.ts tests/visual_manifest.test.ts tests/architecture.test.ts`, `npx @biomejs/biome check` (scoped to changed files)
- Manual steps: reproduced the freeze reliably with `/dev tp -90 668` (offline mode, Chrome and Firefox) before the fix; confirmed no freeze after, same coordinates, same repro steps.

## Screenshots / recordings

N/A (non-visual render-path bug fix)

### Localization (i18n)

### Hygiene

### DIFF
--- a/docs/training-dummy-preload-freeze-postmortem.md
+++ b/docs/training-dummy-preload-freeze-postmortem.md
@@ -1,13 +1,18 @@
-## Bug: teleporting near the zone3 ogre camp freezes the client
+# Training Dummy Preload Freeze: Postmortem
 
-### Symptom
+Date: 2026-07-17
+
+Bug: teleporting near the zone3 ogre camp froze the client. Fixed in PR
+#2078; the systemic backstop is tracked in issue #2079.
+
+## Symptom
 
 `/dev tp -90 668` (near the `thornpeak_ogre` war-camp in zone3) reliably froze
 the client: the first frame rendered, then the screen stopped updating
 entirely while sim ticks and audio kept running in the background. Reproduced
 in both Chrome and Firefox, in offline mode.
 
-### Exact error
+## Exact error
 
 ```
 Uncaught Error: character asset not preloaded: models/creatures/training_dummy.glb
@@ -26,7 +31,7 @@ Uncaught Error: character asset not preloaded: models/creatures/training_dummy.g
 Repeated on every single animation frame after the freeze started (192+
 occurrences observed in one session), each with an identical stack.
 
-### Why it looks like a freeze, not a crash
+## Why it looks like a freeze, not a crash
 
 `requestAnimationFrame` keeps firing every frame (visible in the trace: `frame
 @ main.ts:2761` -> `requestAnimationFrame` -> repeat), so the outer loop never
@@ -36,11 +41,12 @@ The screen never visually updates again, while the sim tick and audio (which
 run earlier in the same frame, on a separate code path from the renderer)
 keep working normally. That's why sound kept firing after the "freeze."
 
-### Root cause
+## Root cause
 
 `training_dummy` is a zone3 camp near the ogre war-camp
 (`{ mobId: 'training_dummy', center: { x: -40, z: 648 }, radius: 0, count: 1 }`,
-`src/sim/content/zone3.ts:1883`), about 54 yards from the teleport point.
+the camp record in `src/sim/content/zone3.ts`), about 54 yards from the
+teleport point.
 Its model, `models/creatures/training_dummy.glb`, is deliberately marked
 `lazyPreload: true` in `src/render/characters/manifest.ts` (it appears in
 exactly one hub, so it was kept out of the eager boot sweep, the same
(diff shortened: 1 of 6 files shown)


FernandoX7 (commented):
Good find, and the postmortem's root-cause work is solid. I verified the load-bearing claims against the head ref rather than the description, and they hold:

- `mob_training_dummy` is one of exactly two `lazyPreload` visuals (`src/render/characters/manifest.ts:666`; the mech is the other), so there is no third entry hiding the same gap.
- `MOB_KEYS` routes `training_dummy` to the gated visual key (`manifest.ts:1083`), and the mob spawns in exactly one zone3 camp as the comments claim.
- The `gltfByUrl.set(def.url, ...)` write vs `has(assetUrl(def.url))` read looks asymmetric, but `LOW_URL_ALIAS` only remaps `rogue_hooded.glb`, so both collapse to the same key for the dummy on every tier, exactly as they do for the mech.
- The `createView` gate (`src/render/renderer.ts:3452`) sits directly under the mech gate with identical early-return semantics, ahead of the pool lookup, so every view-creation path for the dummy is covered.

Ran `npx tsc --noEmit` plus `npx vitest run` over the new test and `render_asset_preload`, `render_asset_fallback`, and `architecture` at the PR head: all green. Note that CI has not run on this PR at all yet (no checks reported on the branch); it needs a green run before merge.

Findings:

- should-fix: `training_dummy_preload_freeze_postmortem.md` is at the repo root. Analysis docs live under `docs/` (`docs/online-movement-latency.md` and `docs/performance-feel-audit.md` are the pattern). Move it there.
- should-fix: the postmortem is stale against its own PR. The closing section is titled "Fix (not yet implemented)" (`training_dummy_preload_freeze_postmortem.md:76`) and the root-cause section still hedges that the model is "missing from (or racing against)" the preload set, when this PR establishes the actual cause (deliberate `lazyPreload`, load never triggered) and ships the fix. Update the doc to match, and keep the fail-soft `resolvedGltf` idea as a named follow-up.
- nit: if the GLB fetch rejects, the memoized rejected promise (`src/render/characters/assets.ts:530`) means no retry, an invisible dummy, and one console.error per frame from the per-frame `.catch` in `createView`. The mech gate has identical semantics, so fine to leave as is; the fail-soft `resolvedGltf` follow-up would subsume both.

The fail-soft `resolvedGltf` change the postmortem calls the more robust fix is worth filing as an issue so it does not get lost: it is what stops the next missing-preload entry from freezing the client the same way.

jamiecypher:
Addressed both should-fix items in 4d900efad: moved the postmortem into docs/ (docs/training_dummy_preload_freeze_postmortem.md, matching the online-movement-latency.md/performance-feel-audit.md pattern), and reworded the root-cause and closing sections to state the confirmed cause and shipped fix instead of the pre-fix hedge/TODO framing. The fail-soft resolvedGltf idea is now called out explicitly as a follow-up, not folded into this PR. No code changes, so the earlier tsc/test verification still holds.

jamiecypher:
@FernandoX7 pushed the fixes for both should-fix items above (4d900efad) whenever you get a chance to take another look.

FernandoX7:
Thanks for the quick doc move. I pushed one small alignment commit on top (b33148351): the hyphenated filename and H1 plus Date shape the other docs/ reports use, the zone3 camp citation now cites the stable path per the docs anchor rule, and the follow-up section points at issue #2079, which I filed for the fail-soft resolvedGltf backstop. Also updated the PR body's postmortem path to the new location and approved the CI runs. The fix itself is unchanged.

FernandoX7:
One more small correction (e4331e942): the postmortem's priority-order explanation said the dummy is non-hostile, but createMob marks every mob hostile and the sim's dummy handling deliberately keeps it attackable for the damage meters (aggroRadius 0 only disables aggro). The doc now describes it as a far hostile deferred behind nearer ones by the per-frame budget, which is what the code does. The delayed-onset conclusion is unchanged.

PR #2081, opened 2026-07-17
fix(render): fail soft when a character visual's assets are missing

### LINKED ISSUE #2079: Missing character asset preload can still freeze the client: make resolvedGltf fail soft
## Problem

`resolvedGltf` (`src/render/characters/assets.ts`) throws synchronously when a character GLB was never registered as preloaded. That throw happens inside the per-frame render path (`Renderer.sync` via `createCandidateViews`), so a single missing-preload entry permanently stalls rendering: the screen stops updating on every subsequent frame while the sim tick and audio keep running on their separate per-frame path. This is exactly the shape of the v0.27.0 training dummy freeze. PR #2078 fixed that one model by wiring up its missing lazy-preload trigger, but the underlying failure mode remains: any future gap (a new `lazyPreload` visual without a trigger, a manifest omission, a tier-alias divergence) can still freeze the client the same way.

A related sharp edge with the same root: the lazy preload gates in `Renderer.createView` (mech and training dummy) memoize their fetch promise. If the fetch rejects permanently, the entity stays invisible forever and the per-frame `catch` logs one console error every frame. A fail-soft `resolvedGltf` subsumes this too, since a missing asset would degrade to a skipped view instead of an error loop.

## Expected behavior

- A missing or not-yet-loaded character asset never stops the renderer: the affected entity's view is skipped for that frame and rendering continues for everything else.
- The miss is logged once per asset URL (dev channel), not once per frame.
- The eager-preload contract stays intact for the normal case: on the happy path, `resolvedGltf` still returns synchronously with no behavior change.

## Scope

- `resolvedGltf` and its callers in `src/render/characters/assets.ts` (`prepareVisual`, `optimizedScene`, attachment assembly).
- The view-creation path in `src/render/renderer.ts` (`createView` / `createCandidateViews`): skipping an entity whose assets are unavailable must leave it a valid future view candidate.
- The lazy preload gates (`preloadMechAssets`, `preloadTrainingDummyAssets`): confirm the memoized-rejection behavior is acceptable once the fail-soft path exists, or reset on rejection so a transient network failure can retry.

## Acceptance criteria

- With a deliberately unregistered GLB (test-forced manifest gap), the renderer keeps painting: no per-frame throw, other entities keep their views, and the affected entity is skipped.
- The miss logs once per asset URL, not per frame.
- A unit test covers the fail-soft path (forced miss resolves to a skipped view and a single log) and the happy path (preloaded assets resolve exactly as today).
- No behavior change for correctly preloaded assets: existing render asset tests (`tests/render_asset_preload.test.ts`, `tests/render_asset_fallback.test.ts`, `tests/training_dummy_preload.test.ts`) stay green.

## Notes

- Follow-up from the PR #2078 review; the postmortem shipped with that PR names this as the robust half of the fix (the lazy-preload wire-up was the targeted half).
- Non-goal: changing which assets are eager vs lazy. The preload set and `lazyPreload` policy stay as they are.

FernandoX7:
## Summary

A character GLB that was never registered as preloaded used to throw synchronously from `resolvedGltf` inside the per-frame render path (`Renderer.sync` via `createCandidateViews`), which permanently stalled rendering while the sim tick and audio kept running. That is exactly the shape of the v0.27.0 training dummy freeze; PR #2078 fixed that one model by wiring its missing lazy-preload trigger, and this PR ships the systemic backstop it named as a follow-up.

`createCharacterVisual` (the factory every renderer construction path goes through: `createView`, the visual-key swap, the druid/polymorph form builds, and the three prewarm seeders) now catches a construction failure, logs it once per visual key and cause through the new `src/render/characters/asset_miss_log.ts` module, and returns null. Every call site skips that entity's view and leaves it a future candidate, so any future missing-preload entry (a new `lazyPreload` visual without a trigger, a manifest omission, a tier-alias divergence) degrades to an invisible entity plus one log line instead of a frozen client. A failed build also sits out a 2 second retry cooldown (`viewCreateRetryAt` in `createCandidateViews`): without it, a permanently failing entity would consume a view-creation budget slot every frame, and under the hitch backoff the budget is 1, so a first-sorted failing entity would starve every other pending view. The happy path is unchanged: assets that are preloaded resolve synchronously exactly as before, and a dedicated happy-path test pins that the catch does not swallow successful builds.

Two deliberate scope decisions, both from the issue's scope notes:

- The lazy preload gates (mech, training dummy) keep their memoized no-retry behavior; a per-frame retry would spam fetches on a permanent failure. Their catch logging now dedupes through the same module, one line per asset instead of one per frame. The gates share the retry cooldown while their fetch is in flight (so a deferring entity burns no per-frame budget slot) and clear it when the fetch resolves, so pop-in stays at the next frame after readiness; only a rejected fetch waits out the full cooldown.
- The post-construction weapon-swap path (`setWeapon` and the deferred stow swap) still resolves synchronously without a catch: weapon and weapon-skin URLs are pinned into the tier-independent preload union by `tests/render_asset_preload.test.ts`, and the renderer latches the diff field before the call, so a hypothetical miss there throws once per weapon change (no freeze) and is prevented by construction rather than handled at runtime. The armory inspect window's own construction paths (its separate GL loop) are likewise event-driven, pre-existing, and untouched.

The renderer-level acceptance criterion ("the renderer keeps painting, other entities keep their views") was verified end to end in a real browser, not just by inspection: with `mob_training_dummy`'s GLB deliberately renamed to a missing file (uncommitted local edit) and the client teleported to the postmortem coordinates (`/dev tp -90 668` equivalent), a scripted 10 second observation recorded zero uncaught page errors (pre-fix: one per frame), exactly one deduped miss log across multiple retry-cooldown cycles, 228 sim ticks advanced, pixels changing between screenshots 9 seconds apart, and the broken entity present in the sim with no view while the other 29 entities in range all received theirs. The unit tests additionally pin the factory contract (null plus one log on a miss, non-null with zero logs on the happy path), the log dedupe, and the retry-gate mechanics (cooldown boundary, per-entity and per-slot independence, prune retention and cleanup).

## Related issues

Closes #2079. Follow-up to #2078.

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands: `npm run gate` (green through i18n, malware scan, biome, SFX, and the full unit suite; the browser-regressions step failed only on the known environmental `armory_mobile_layout.browser.test.ts` layout flake on this machine, which is unrelated to this diff and fails identically on a clean base, so `npx tsc --noEmit` and `npm run build` were then run manually and are green; PR CI is the arbiter for the browser suite). Also targeted `npx vitest run tests/character_visual_fail_soft.test.ts tests/training_dummy_preload.test.ts tests/render_asset_preload.test.ts tests/render_asset_fallback.test.ts tests/architecture.test.ts` and `npx @biomejs/biome check` on the changed files.
- Manual steps: none needed beyond the suites; the new test drives the real miss path (a lazyPreload visual under a mocked loader hits the actual `character asset not preloaded` throw inside `new CharacterVisual`) and pins the null return plus the once-per-asset log. The test was written first and failed on the base with the real throw.

### Quality

- [x] **The full gate passes** apart from the known environmental armory browser-layout flake on the dev machine (unrelated to this diff, fails identically on a clean base; details above). `npx tsc --noEmit` and `npm run build` were completed manually and are green, decisive tests were added for the changed behavior, and PR CI is the arbiter for the browser suite.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings. (The new logging is dev-channel `console.error`, which stays English by policy.)

### Hygiene

### DIFF
--- a/src/render/CLAUDE.md
+++ b/src/render/CLAUDE.md
@@ -28,6 +28,9 @@ Everything else is a sibling module in one of these families:
   Performance) and `crowd_lod.ts` (pure crowd policy: pulls character
   shadow/anim cadence in as rig counts climb; cosmetic-only, exempts what a
   player reacts to).
+- `view_create_retry.ts`: bounded cooldown state for fail-soft character builds
+  in per-frame paths, including required targets, form swaps, and visual-key
+  swaps (`tests/view_create_retry.test.ts`).
 - `self_motion.ts`/`facing_smooth.ts`: pure display-only self layers (bounded
   online pose extrapolation + rate-limited self yaw; never touch world state,
   see `src/net/CLAUDE.md`).
--- a/src/render/characters/CLAUDE.md
+++ b/src/render/characters/CLAUDE.md
@@ -22,6 +22,10 @@ no procedural-rig path here anymore. Reads the world; never mutates the sim.
   `tests/render_asset_preload.test.ts`). `prepareVisual(key)` memoizes
   normalize transform, resolved clips, click-capsule radius, and a baked
   idle-pose geo (far-LOD/shadow proxy).
+- `asset_miss_log.ts`: once-per-key dev logging for character-asset failures in
+  per-frame render paths; `createCharacterVisual` returns null on such a
+  failure so callers skip the view for the frame instead of stalling the
+  renderer (`tests/character_visual_fail_soft.test.ts`).
 - `rig_merge.ts`: merges a KayKit rig's quantized body-part SkinnedMeshes into
   one draw per material (`assets.ts` `assembleModel` calls it). Read its
   header bind-pose proof before touching bone inverses.
--- a/src/render/characters/asset_miss_log.ts
+++ b/src/render/characters/asset_miss_log.ts
@@ -1,0 +1,16 @@
+// Dev-channel dedupe for character-asset failures that surface in per-frame
+// render paths (a missed preload, a rejected lazy fetch). The first failure
+// for a key logs; repeats are dropped, so a permanent miss costs one log line
+// instead of one per frame (issue #2079; the per-frame variant was the log
+// spam half of the v0.27.0 training dummy freeze,
+// docs/training-dummy-preload-freeze-postmortem.md).
+const loggedKeys = new Set<string>();
+
+/** Log a per-frame asset failure once per key. Returns whether it logged. */
+export function logAssetMissOnce(key: string, message: string, detail?: unknown): boolean {
+  if (loggedKeys.has(key)) return false;
+  loggedKeys.add(key);
+  if (detail === undefined) console.error(message);
+  else console.error(message, detail);
+  return true;
+}
--- a/src/render/characters/index.ts
+++ b/src/render/characters/index.ts
@@ -3,6 +3,7 @@
 // with the preload gate, so createCharacterVisual is synchronous by the time
 // the Renderer constructs views.
 import type { Entity, PlayerClass } from '../../sim/types';
+import { logAssetMissOnce } from './asset_miss_log';
 import { mechHeldWeaponOverride, visualKeyFor } from './manifest';
 import { CharacterVisual } from './visual';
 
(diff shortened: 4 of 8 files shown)


FernandoX7:
Follow-up on the retry-gate commit (8a2ac62d3): the gates armed the full cooldown while a preload fetch was merely in flight and nothing cleared the slot on resolution, so a mech or dummy popped in up to two seconds late, and a player logging in with the mech cosmetic active stayed invisible for that window via the required-view path. The gates now clear their slot when the fetch resolves: next-frame pop-in on success, full cooldown only on a rejected fetch, still no per-frame budget burn while in flight. Also added two gate test pins the suite missed (prune retains unexpired entries for present entities; cooldowns are keyed per entity).

FernandoX7:
End-to-end proof of the renderer-level acceptance criterion, run against head 448aaa7ce: with mob_training_dummy's GLB deliberately renamed to a missing file (uncommitted local edit) and the client scripted to the postmortem coordinates, a 10 second observation recorded zero uncaught page errors (pre-fix this scenario produced one per frame and a frozen screen), exactly one deduped miss log across multiple retry-cooldown cycles, 228 sim ticks advanced, pixels changing between screenshots 9 seconds apart, and the broken entity present in the sim with no view while the other 29 entities in range all received theirs. Also note the branch history was rewritten once (identical trees) to wrap a commit body per the repo convention; CI is green on the rewritten head.

PR #2083, opened 2026-07-17
feat(sfx): trigger idle mob-voice barks with a density-damped periodic sweep

jamiecypher:
## Summary

Idle mob vocalizations were fully staged, conformed, and gain-mapped for every mob family (13 base families plus the wolf and skeleton subfamily aliases, 26 audio files total), but nothing ever called them: `idle` was missing from `MobVoiceAction` entirely at the type level.

Adds `idle` to `MobVoiceAction` and to every `MOB_VOICE_CUES` family entry (`src/ui/combat_sfx.ts`; all `mob_<family>_idle` keys already existed in the manifest). New pure module `src/ui/mob_idle_sfx.ts` (`pickIdleBarkCandidates`, `idleDensityFactor`, `isIdleBarkCandidate`) picks which nearby, non-combat, unmuted mobs attempt a bark on a shared throttled sweep (`Hud.sweepMobIdleBarks`, hooked into `update()` alongside the existing `reconcileSfx` throttle) rather than each mob rolling dice per frame. Same-family cluster size damps the per-mob chance so a pack does not all bark in the same sweep, while still sounding "populated" (expected total barks scale with the square root of the crowd size, not linearly). A per-entity cooldown map (`mobLastIdleBarkAt`, pruned in `reconcileSfx` the same way as the existing `mobAggroed` set) rate-limits repeats.

`sfx.playAt` now returns whether it actually scheduled a sound (previously `void`), so the per-entity cooldown is only stamped on a real play, not a bark that lost the existing per-key playback cooldown race to another mob's identical cue in the same instant.

Tuning starts on the generous side rather than maximally conservative: the mastered idle takes measure quiet even boosted in-game, so under-triggering felt like the bigger risk than spam. All six knobs (check interval, scan radius, base chance, per-entity cooldown, per-key cooldown, gain) are plain exported constants in `src/ui/mob_idle_sfx.ts`, easy to retune by ear later.

## Related issues

N/A

## Type of change

- [x] Tests

## How was this tested?

- Commands: `npx tsc --noEmit`, full `npm test` (1210/1210 test files), scoped `npx @biomejs/biome check` on every changed file, all three build steps (`build:env`/`build:server`/`build`)
- Manual steps: in-game listening pass around the two `forest_wolf` camps and the `old_greyjaw` rare near the starting area, and the `thornpeak_ogre`/`ogre_crusher` camps in zone3, confirming barks feel populated without being spammy, vary in volume/panning with distance, and no single mob repeats rapidly

## Screenshots / recordings

N/A (audio-only change)

### Localization (i18n)

### Hygiene

- [x] I didn't hand-edit generated files. `src/game/sfx_manifest.generated.ts` is regenerated via `npm run sfx:manifest`.

### DIFF
--- a/docs/design/sound_effects.md
+++ b/docs/design/sound_effects.md
@@ -274,14 +274,17 @@ pitched-up `attack`. Families: `beast`, `boar`, `spider`, `mudfin`, `burrower`,
 | `demon` | a demon — sinister hissing snarl / shrieking demonic strike / agonized demonic death wail |
 
 ### Idle vocalizations (spatial, ambient bark)
-`mob_<family>_idle` (one per family, `reptile` included, 14 total), plus a
-subfamily-specific `mob_<family>_<subfamily>_idle` for the wolf and skeleton
-aliases (see `SUBFAMILY_ALIAS` in `src/ui/combat_sfx.ts`). Unlike the reactive
+`mob_<family>_idle` (one per family, `reptile` included, 13 total), plus a
+subfamily-specific `mob_<family>_<subfamily>_idle` for the wolf aliases (see
+`SUBFAMILY_ALIAS` in `src/ui/combat_sfx.ts`; a matching `mob_undead_skeleton_*`
+set exists on disk but is not currently wired to any live template, an
+inherited gap this feature did not introduce). Unlike the reactive
 aggro/attack/death/hurt vocalizations (fired directly off sim combat events),
 idle barks are presentation-only ambience: a shared periodic sweep
 (`Hud.sweepMobIdleBarks`, throttled to `MOB_IDLE_CHECK_INTERVAL_MS` from
-`update()`, never per-mob-per-frame) considers every non-combat, unmuted mob
-within `MOB_IDLE_SCAN_RADIUS` of the player and rolls each one independently
+`update()`, never per-mob-per-frame) considers every non-combat, unowned,
+non-dummy, unmuted mob within `MOB_IDLE_SCAN_RADIUS` of the player and rolls
+each one independently
 against `MOB_IDLE_BASE_CHANCE`, damped by how many same-family mobs are
 clustered nearby (`idleDensityFactor`, `src/ui/mob_idle_sfx.ts`) so a dense
 pack does not all bark in the same sweep. A per-entity cooldown
--- a/docs/design/sound_effects.md
+++ b/docs/design/sound_effects.md
@@ -273,6 +273,28 @@ pitched-up `attack`. Families: `beast`, `boar`, `spider`, `mudfin`, `burrower`,
 | `dragonkin` | a dragonkin — fierce roaring alert with wing flap / snapping bite roar / dying roar collapse |
 | `demon` | a demon — sinister hissing snarl / shrieking demonic strike / agonized demonic death wail |
 
+### Idle vocalizations (spatial, ambient bark)
+`mob_<family>_idle` (one per family, `reptile` included, 14 total), plus a
+subfamily-specific `mob_<family>_<subfamily>_idle` for the wolf and skeleton
+aliases (see `SUBFAMILY_ALIAS` in `src/ui/combat_sfx.ts`). Unlike the reactive
+aggro/attack/death/hurt vocalizations (fired directly off sim combat events),
+idle barks are presentation-only ambience: a shared periodic sweep
+(`Hud.sweepMobIdleBarks`, throttled to `MOB_IDLE_CHECK_INTERVAL_MS` from
+`update()`, never per-mob-per-frame) considers every non-combat, unmuted mob
+within `MOB_IDLE_SCAN_RADIUS` of the player and rolls each one independently
+against `MOB_IDLE_BASE_CHANCE`, damped by how many same-family mobs are
+clustered nearby (`idleDensityFactor`, `src/ui/mob_idle_sfx.ts`) so a dense
+pack does not all bark in the same sweep. A per-entity cooldown
+(`MOB_IDLE_PER_ENTITY_COOLDOWN_MS`) additionally rate-limits one mob's own
+repeats, stamped only when `sfx.playAt` reports the sound actually played
+(not merely attempted), so losing the shared per-key playback cooldown
+(`MOB_IDLE_KEY_COOLDOWN_S`, the backstop against two mobs barking the
+identical clip at once) does not silently bench a mob for the full
+per-entity window. Plays at `MOB_IDLE_GAIN`, a dedicated lower bucket
+distinct from the combat layer's `COMBAT_GAIN`. Timing is client-local and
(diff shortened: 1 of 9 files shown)


Rubsey (commented):
The archaeology here is right: all 13 family idle keys (plus the wolf subfamily takes) were already staged in the manifest with nothing triggering them, and the wiring lands in exactly the house shape. Pure logic in `src/ui/mob_idle_sfx.ts` with decisive tests (the density test's arithmetic checks out by hand, and the cooldown test pins the exact boundary), a thin sweep on the Hud state that needs it, manifest untouched, and the stamp-only-on-real-play design through the `playAt` return catches a race most versions of this feature would have shipped. I approved the fork CI runs and everything is green. Findings below.

**Findings**

- **should-fix** `src/ui/mob_idle_sfx.ts:44` (fed from `src/ui/hud.ts:7678`): the "not mid-combat" gate keys off `mobAggroed`, which is only populated by observed damage events and only cleared on a death event or the leaves-interest prune. Two consequences. A mob that fought and then evaded/leashed stays in the set, so it never idle-barks again while in interest (offline that is session-permanent, since living mobs never leave the Sim's entity map). In the other direction, a mob chasing someone without having landed a hit yet can bark mid-combat. The live signal already exists in both hosts: the sim nulls `aggroTargetId` on leash (`src/sim/mob/locomotion.ts:132` and siblings) and it is mirrored on the wire (`out.aggro` at `server/game.ts:979`, `e.aggroTargetId = w.aggro ?? null` at `src/net/online.ts:2097`). Gating idle on `e.aggroTargetId === null` fixes both directions; note that keeping the set check alongside it would still leave the evaded-mob bench in place, so for idle purposes it can simply replace the set (`mobAggroed` still serves the aggro-voice dedupe).

- **should-fix** `src/ui/mob_idle_sfx.ts:42`: owned pets, the delve companion, and the Training Dummy all pass `isIdleBarkCandidate`. Owned pets (tamed or summoned) are `kind: 'mob'` with real families and stand next to you all session, so your own pet barks unprompted roughly every half minute at point-blank range; `acolyte_tessa` (family `humanoid`) grunts generic humanoid idles between her curated lines; and `training_dummy` (`src/sim/content/zone3.ts`, `family: 'humanoid'`, `dummy: true`) is a wooden target that now grunts. An `e.ownerId === null` conjunct plus a dummy check in the predicate covers all three (`ownerId` is already mirrored on the wire), and since you deliberately extracted the predicate for testability, the fix and its tests are both small.

- **should-fix** `tests/sfx.test.ts:193`: the `playAt` void-to-boolean change is the load-bearing contract of the whole design (hud.ts stamps the per-entity cooldown only `if (played)`), and no test anywhere asserts the return value; the only existing `playAt` assertion is `not.toThrow()`. That file already has the harness (mocked AudioContext, direct buffer injection), so pinning false on a cooldown-blocked repeat, false on an unbuffered key, and true after buffer injection is a few lines. As written, a regression that returns true on the cooldown path (the exact 12s-bench bug your own doc comment warns about) fails zero tests.

- **should-fix** `docs/design/sound_effects.md:277`: two claims in the new section do not match the code. "14 total" should be 13 (count `MOB_VOICE_CUES`, or your own new test's rows), and "the wolf and skeleton aliases (see `SUBFAMILY_ALIAS`)" is wrong about skeleton: `SUBFAMILY_ALIAS` has only the four wolf templates, and no live template can reach `mob_undead_skeleton_idle` (the only skeleton template, `nythraxis_skeleton_warrior`, resolves a different key and is on the mute list anyway). The orphaned skeleton takes are inherited, not this PR's bug, but the new prose should not claim they are wired.

- **nit** `src/game/sfx.ts:395`: the new return-value comment sits as a second JSDoc block stacked on the existing one-liner; merge them. Its false-case list also omits the first false path (`!ctx || !master`, audio never initialized).

- **nit** `tests/combat_sfx.test.ts:314`: the new idle test hardcodes the 13 template-to-family pairs; iterating `Object.entries(MOB_VOICE_CUES)` and asserting each `cues.idle in SFX_CLIPS` would auto-cover a future 14th family (the `satisfies` clause only forces a cue string at compile time, not a staged clip). Copying the hurt test's shape is fair precedent, but this copy also drops that test's `toHaveLength(13)` self-check, so a silently dropped row would go unnoticed.

**Non-blocking notes**

- The first idle bark of each family per session takes the lazy-load path: `playAt` returns false, the retry plays the clip about 120ms later, and the per-entity cooldown never gets stamped, so that mob can re-bark a sweep or two later. Self-limiting, but the hurt-cue warm pattern (`sfx.preload`, hud.ts:7856) would close it if you care.
- With `cooldown: 3`, a first-ever play of a key inside the first couple seconds of AudioContext life returns false via the pre-existing `?? -1` fresh-key sentinel in `playAt` (every prior cooldown was 0.3s or less, so it never mattered before). Harmless here since the first sweep's clips are unbuffered anyway, but `?? -Infinity` would restore "never played = always allowed" for future long-cooldown callers.
- `mobLastIdleBarkAt` is not cleared on the death event, only by the interest prune; worst case an entity-id reuse waits out up to 12s of stale cooldown. Fine as is.
- Consider adding `mob_idle_sfx.ts` to the `UI_PURE_CORES` allowlist in `tests/architecture.test.ts` so the new pure module gets the same DOM-import guard the other Node-tested cores get.

jamiecypher:
@Rubsey addressed everything in 2e78390ae:

- **aggroTargetId instead of mobAggroed**: isIdleBarkCandidate now checks `e.aggroTargetId === null`, fixing both directions (evaded mobs re-admitted, mid-chase mobs excluded). mobAggroed is untouched, still doing its original aggro-voice dedupe job.
- **Owned entities + dummy excluded**: `ownerId === null` plus a `MOBS[...].dummy` check, covers pets, delve companions, and the Training Dummy.
- **playAt return-value coverage**: tests/sfx.test.ts now pins false-on-unbuffered and false-then-true across the per-key cooldown boundary.
- **Idle-cue test generalized**: iterates live MOB_VOICE_CUES instead of a hardcoded template-id table, with the family-count self-check restored.
- **Doc corrections**: 13 families (not 14), and the skeleton-alias claim is fixed to state it's unwired (not this PR's bug).
- **Both nits**: merged JSDoc block on playAt (now also documents the `!ctx || !master` false path), idle test iterates the catalog generically.
- Also took the optional note: mob_idle_sfx.ts is now registered in UI_PURE_CORES/BARE_NAMED (had to drop pickIdleBarkCandidates' Math.random default to satisfy the determinism scan; moved to the hud.ts call site).

Left the other non-blocking notes as noted (lazy-load race, fresh-key sentinel, mobLastIdleBarkAt death-event clearing) since you called them fine as-is.

Full local gate re-run clean: tsc, biome (scoped), full test suite (1209/1209 files, one known-flaky unrelated server/http timing test confirmed passing in isolation), all three builds.

PR #2086, opened 2026-07-17
Claude/trading bots solana woc ugu6xo

awidearray:
## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
- Manual steps:

## Screenshots / recordings

---

## Checklist

### Quality

- [ ] **The full gate passes.** `npm run gate` is green. I added or updated decisive

### Cross-platform

### Localization (i18n)

- [ ] **New player-visible strings follow the contributor policy.** Every user-facing

### Hygiene

- [ ] No secrets, credentials, or `.env` committed, and `ALLOW_DEV_COMMANDS` is
- [ ] I didn't hand-edit generated files such as

> Commit messages and PR titles use [Conventional Commits](https://www.conventionalcommits.org/)
> with a required scope (`feat(talents): ...`, `fix(net): ...`).

---

A complete checklist and a green CI run (tests, typecheck, and builds) are what
we look for before merging. Smaller, focused PRs land faster than large ones, and
reviewers may suggest changes, which is a normal and friendly part of the
process. Thank you for helping build World of ClaudeCraft. Questions? Join us on
[Discord](https://discord.com/invite/worldofclaudecraft).


EnriqueGF (commented):
## Review

Thanks for this, it is a big surface (12k lines) and the architecture choice is the right one: both `server/dex_swap.ts` and `server/trading_bots.ts` are genuinely thin proxies with zero crypto/custody logic, the game server never signs or holds keys, and both features are fail-closed twice over (own flag + configured service connection, defaulting OFF). The field-picking on every POST (only contract fields re-serialized, never the raw client body) and the exhaustive per-code error mapping tables are exactly right.

### What I ran

- `npx tsc --noEmit`: clean.
- Guard tests: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/i18n_completeness.test.ts`, `tests/api_error_code_parity.test.ts`: green.
- The new suites plus the whole `tests/server/http/` tree: 48 files, 1119 tests, all passing.
- `biome ci --changed`: no errors, no format diffs (16 lint warnings, which CI does not gate on).
- Copy scan over the added lines: no em dashes, no en dashes, no emoji. The five non-Latin overlay fills (ja, ko, ru, zh_CN, zh_TW) are present, so the M16 wordy-copy exception is satisfied and no other locale overlay was touched.

I also confirmed the things that would have been blocking and are not: no client can forge identity (`playerIdFor` reads only `ctx.account.accountId`, never a client field), `WOC_ECONOMY_INTERNAL_SECRET` is env-only and never reaches a response or a log, the transport-failure branch emits a fixed literal rather than the service URL, `src/sim/` is untouched, and the six mutating trading-bot routes correctly require `scope: 'active'` (that last one is decisively pinned by the loop in `tests/trading_bots.test.ts:289-310`, which is good to see).

### Findings

**1. No rate limiting on any of the 11 new routes; the two public config routes are the sharp edge.** (should-fix before the flags go on)

Neither route table imports `rateLimit` from `server/http/middleware/rate_limit.ts`, and there is no blanket limiter on `/api`: each route opts in, which is why `server/wallet.ts`, `server/characters.ts`, `server/reports.ts`, `server/maps_routes.ts` and `server/user_assets_routes.ts` all mount one explicitly.

The header comments say the per-player limits live in the economy service, which is fine for the routes that forward `x-woc-player-id`. But `GET /api/dexswap/config` and `GET /api/tradingbots/config` are unauthenticated, unkeyed and unlimited, and each one turns into a fresh outbound `fetch` to the internal service with the internal secret attached. That is a free amplification channel from the public internet onto the internal service and onto this server's own fd budget, with no game-server-side backstop. `PUBLIC_READ_POLICY` (`rate_limit.ts:178`) is exactly the shape for this. Worth an IP policy on the two config reads at minimum, and arguably an `ip+account` one on the mutating routes so a service outage does not remove all limiting.

**2. `GET /api/tradingbots/status` accepts a read-scoped token, so a `character:read` OAuth app sees the player's financial data.** (should-fix)

`tradingBotsReadAuth` (`server/trading_bots.ts:122-124`) uses `scope: 'read'`. Per `server/db.ts:908-910` a `'read'` token is "a companion-app / OAuth character:read token". That one response carries subscription state, Claudium balance, vault SOL/WOC balances, recent trades and the PnL summary. A third-party app the player authorized only to read their character therefore gets their full trading and balance history.

This is a data-exposure question, not a mutation bypass (the mutating routes are correctly gated). But it seems worth deciding deliberately rather than inheriting: either `'active'` on status, or a documented note that trading data is in scope for `character:read`. The equivalent on `/api/dexswap/quote` is much lower stakes and reads fine as-is.

**3. `withdraw` forwards a client-chosen `userPublicKey` payout destination.** (consider, defense in depth)

`withdrawHandler` relays `userPublicKey` verbatim. The only control is service-side: `docs/prd/woc/trading-bots.md:68` says withdrawals must pay out only to the account's verified linked wallet regardless of any request field. This server already holds that verified link in its own database but does not forward it, so the service has to keep an independent copy. If that copy is ever stale or the check regresses, a stolen full-session token becomes direct fund theft with no wallet signature in the loop. Forwarding the server-known linked wallet instead of relaying the client's would make the game server's own data the binding constraint. The client-side `linkedWallet() !== null` gate in `src/net/trading_bots_api.ts:114` is UI only and is not a control.

**4. Blind signing in `signAndSendTransactionBase64`.** (consider, design note)

`src/net/wallet.ts:441-457` hands the service-supplied base64 straight to the wallet with no decoding and no reconciliation against the quote the UI displayed. The message path directly above it does defend itself: `signMessageBase58:414` verifies `bytesEqual(result.signedMessage, messageBytes)` and throws if the wallet altered the message. There is no analogous check on the transaction path, so a misbehaving service could return a transaction that does something other than the quoted swap and the only remaining control is what the player's wallet extension renders. I do not think this blocks the PR (it is the standard non-custodial tradeoff and decoding Solana transactions client-side is real work), but the asymmetry with the message path is worth a comment at minimum so the choice is explicit.

**5. Test gaps.** (should-fix, cheap)

The suites are unusually strong overall, so these are specific:

- The dexswap auth tier is entirely unpinned. `tests/dex_swap.test.ts` only ever mints a `'full'` token, and the one scope-adjacent assertion (`expect(quoteRoute.middleware?.length).toBe(2)`, `:389`) is tier-blind. The `'read'` choice on quote/swap is deliberate and well-reasoned in the header comment, but it currently rests on that comment. Mirroring the `READ_TOKEN` fixture from `trading_bots.test.ts:181-188` closes it in a few lines.
- Middleware order (gate before auth) is claimed by test titles in both families but asserted by neither, and every flag-off test uses an already-authenticated ctx. The order is correct in the source, but if it were ever swapped, an anonymous probe of a disabled feature would answer 401 instead of 404 and leak that the feature exists. One test per family: flag off, no `Authorization` header, expect 404.
- The `refused()` concurrency/cooldown guard (`src/ui/trading_bots_flows_core.ts:99-102`) is only exercised through `rentFlow`. Deleting `if (refused(host)) return;` from `depositFlow`, `withdrawFlow` or `controlFlow` leaves the suite green. These are money operations, so a double-fire deposit is worth its own arm.
- Minor: `queryParam`'s array branch (`server/dex_swap.ts:220-223`) is untested. It is the parameter-pollution normalization, and a regression to `String(value)` would silently forward `"a,b"` upstream.

### One process note

This PR targets `main` rather than `release/v0.28.0`. Worth confirming that is intentional given where the release branch currently is.

Nothing here is blocking given both features default to OFF, but I would want 1, 2 and 5 addressed before either flag is turned on in production.

Issue #2088, opened 2026-07-17
Extract a shared bareClient test fixture (21 hand-rolled copies)


FernandoX7:
## Problem

The bareClient test fixture (Object.create(ClientWorld.prototype) plus hand-set fields, the sanctioned idiom documented in tests/CLAUDE.md) is hand-rolled in 21 test files. Three copies are byte-near-identical (tests/snapshots.test.ts, tests/corpse_harvest_sim.test.ts, tests/crafting_view_combo_liveness.test.ts; the last adds craftSkills and craftingIdentity defaults), and the other 18 are divergent field subsets (for example tests/talents.test.ts sets only around 8 fields). Every new ClientWorld field now needs 21 manual sweeps or silently diverging fixtures. The fakeWs/lastSnap/joinServer/broadcast helper family from tests/snapshots.test.ts is similarly copied in three suites.

## Expected behavior

- A shared tests/helpers/bare_client.ts exports bareClient(pid: number, overrides?) that mirrors ClientWorld's full declaration defaults, including the crafting mirrors (emptyCraftSkills() and a pre-sync craftingIdentity, which is strictly more faithful to a freshly constructed online client), then applies overrides via Object.assign.
- The same helper module hosts the fakeWs/lastSnap/joinServer/broadcast family, keeping the PlayerClass parameter from the snapshots joinServer variant.
- Suites import the helper instead of hand-rolling copies.

## Scope

- Add tests/helpers/bare_client.ts with the shape above.
- Adopt it first in the three byte-near-identical copies (snapshots, corpse_harvest_sim, crafting_view_combo_liveness).
- Migrate the 18 divergent variants with per-file verification (each may rely on its subset shape; treat every migration as its own reviewed step).
- Update the tests/CLAUDE.md idiom note to point at the helper.

## Acceptance criteria

- The three identical copies are replaced by helper imports and their suites stay green.
- Divergent variants are either migrated with per-file verification or left in place with a one-line comment naming the helper as the default for new suites.
- No test behavior changes: the migration is fixture-only.

## Notes

- The vi.mock('../server/db') factory CANNOT move into an imported helper: vitest hoists vi.mock above imports, so the factory would close over an uninitialized binding. If mock unification is ever wanted, a server/__mocks__/db.ts superset file mock would allow bare vi.mock('../server/db') calls, but 53 test files hand-roll bespoke factories today, so that is its own audit and explicitly out of scope here.
[...]

PR #2089, opened 2026-07-17
fix(sim): Professions 2.0 Phase 3 QA, two-phase trade swap + hcb live-broadcast coverage

FernandoX7:
## Summary

Phase 3 QA of the Professions 2.0 program (the QA twin of PR #2087, per docs/professions-2/phase-03-qa.md): a three-agent audit (correctness, coverage, cleanup) plus the dispatch-matrix reviewer fan-out over the phase diff, with every should-fix applied. QA verdict: PASS, zero blocking.

The one real correctness defect found lives in the pre-landed PR 2045 trade code, not the Phase 3 diff: tradeConfirm ran the a-to-b transfer to completion (granting into b's bag) before removing b's give, so when both offers shared an itemId the second removal consumed just-granted stock. A swapped instance bounced straight back to its owner (removeItem scans highest-index-first, exactly where addItemInstance pushes), and a plain-for-instance offer spared the instance while mis-routing a plain copy. Conservation always held (no dupe or loss), but the payload carriage the fix exists to guarantee silently failed for same-item trades. The swap is now two-phase (removeOffer both sides, then grantOffer both sides), which matches the model the fitsAfterSwap capacity gate already validated. Reproduced test-first; the coverage reviewer's revert experiment confirms both repro pins red on the old code and the parity draw-order golden is unchanged (the module draws no rng).

Coverage closed on the phase's most valuable gap: the Phase 3 hcb pins drove hand-assembled snap envelopes (always fullJson-shaped), so the exact production sequence the mirror exists for never ran in any test. A new live GameServer broadcast suite pins the claim-after-first-sight lite delta record (hcb present, no identity resend) and interest-scope eviction/re-entry (claims AND clears made out of view arrive on return; a delta-guarded mirror mutation reds the clear arm). Also added: a partial-stack sparing pin (removePreferFungible must choose to spare the instanced copy), a no-escrow trade-cancel contract pin, and the claimed-corpse arms of corpseLootAvailability (self-claimed viewer, claimed-by-another, claimed-with-personal-loot stays openable).

Docs: the hcb as-landed deviation is swept into the phase file and its QA twin (the known amend-the-twin trap: both still instructed the abandoned ALL_DELTA_KEYS / TERSE_TO_IWORLD registration), the phantom bandwidth hcb pin claim is corrected (the snapshots sparse-absence assertion is the real no-bloat tooth), the bareClient drift note is rescoped to the real 21-copy repo-wide idiom (extraction filed as #2088), the Phase 4 despawn-grace heads-up is recorded, and the Phase 3 QA row is flipped.

Deferred with reasons (recorded in progress.md): a bandwidth claimed-corpse scenario (docs made truthful instead) and a distinct-itemId slot-order byte-equivalence pin (the push-to-end plus splice-compaction reasoning is airtight per the architecture review). The harvestStateReliable open-gate flip remains the documented Phase 4 deferral.

## Related issues

Part of #1866. The bareClient fixture extraction chore is filed as #2088.

## Type of change

- [x] Bug fix
- [x] Tests
- [x] Documentation

## How was this tested?

- Commands:
  - `npx vitest run tests/trade.test.ts tests/snapshots.test.ts tests/bandwidth.test.ts tests/corpse_harvest_sim.test.ts tests/world_api_parity.test.ts tests/crafting_view_combo_liveness.test.ts tests/corpse_loot_availability.test.ts tests/env_protocol.test.ts tests/architecture.test.ts tests/parity.test.ts tests/localization_fixes.test.ts` (10 files, 477 passed)
  - `npx tsc --noEmit` clean; `npm run gate` under Node 24 (known environmental armory browser red on this machine; typecheck and all builds completed manually after it, PR CI is the arbiter)
  - Reviewer fan-out over this QA diff: architecture-reviewer, cross-platform-sync, test-coverage-auditor (with revert/mutation experiments), qa-checklist: all PASS, zero blocking
- Manual steps: the trade sequencing bug was reproduced test-first (both repro tests red on a8c65a2c2, green after the two-phase fix); the coverage reviewer independently re-ran the revert experiment.

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green up to the known environmental armory browser failure on this machine (pre-existing, reproduces on an untouched release tip); typecheck and all builds were completed manually after it. Decisive tests added for all changed behavior.

### Localization (i18n)

### Hygiene

### DIFF
--- a/docs/professions-2/phase-03-parity-bug-fixes.md
+++ b/docs/professions-2/phase-03-parity-bug-fixes.md
@@ -5,6 +5,17 @@ Release PR 2045 landed the instance-preserving trade fix on release/v0.27.0 (reg
 `tests/trade.test.ts`) before this phase started, so the only remaining Phase 3 deliverable is the
 `harvestClaimedBy` mirror below. Re-verify against code at session start per the docs anchor rule.
 
+AS-LANDED DEVIATION (2026-07-17, Phase 3 merge a8c65a2c2, reviewed by cross-platform-sync): the
+instructions below to register `hcb` in `ALL_DELTA_KEYS` / `TERSE_TO_IWORLD` were mis-specified.
+`harvestClaimedBy` is a PER-ENTITY field, so `hcb` rides `dynamicFields` in `server/game.ts`
+(sparse emit, auto-delta via the per-entity wire cache; the `tap` key is the template), and those
+two registries pin selfWireJson maybe() SELF keys only (their scrape test asserts set-equality, so
+listing `hcb` there would break it). The as-landed pin set: the round-trip and live-broadcast
+suites in `tests/snapshots.test.ts` and `tests/corpse_harvest_sim.test.ts` (the sparse-absence
+assertion in snapshots is the no-bloat tooth; `tests/bandwidth.test.ts` stays green but carries no
+hcb-specific scenario). Authority: the Phase 3 LANDED entry in `state.md`. Read the registry
+instructions below through this note.
+
 This phase fixes the two known host-parity data bugs so item instances and corpse claims are
 truthful in every host: trades currently strip `ItemInstancePayload` (the `removePreferFungible`
 return is discarded), and `harvestClaimedBy` is hardcoded `null` in `ClientWorld`, so the online
@@ -27,8 +38,9 @@ survive trades; Phase 4 gathering trusts corpse claims), and neither adds new ga
 - `src/sim/professions/combo_eligibility.ts`: the shared combo gate both hosts must consume
   (the 2033 stub trap: verify liveness, not just member shape).
 - `tests/trade.test.ts`, `tests/corpse_harvest_sim.test.ts`: the suites to extend.
-- `tests/snapshots.test.ts`: the `ALL_DELTA_KEYS` / `TERSE_TO_IWORLD` wire pins; the existing
-  `prof`/`gprof`/`ncd`/`tfocus` self-wire keys are the pattern for `hcb`.
+- `tests/snapshots.test.ts`: the `ALL_DELTA_KEYS` / `TERSE_TO_IWORLD` wire pins (SELF keys only;
+  as landed, `hcb` is a per-entity `dynamicFields` key patterned on `tap`, see the deviation note
+  above).
 - `CLAUDE.md` (root) plus `src/sim/CLAUDE.md`, `src/net/CLAUDE.md`, `server/CLAUDE.md`,
   `src/ui/hud/CLAUDE.md`, `tests/CLAUDE.md`.
 
@@ -79,9 +91,10 @@ Spawn an Explore agent to read and summarize:
   src/ui/hud/CLAUDE.md, tests/CLAUDE.md
 The agent must return: exactly where trade.ts discards the removePreferFungible return and what
 that return contains; the full ItemInstancePayload shape (signer, charges, rolled, boundTo) and
-the instance-preserving grant helpers in bags.ts; the recipe for adding a delta-omitted terse
-wire key (server emit in wireEntity, ClientWorld mirror, ALL_DELTA_KEYS + TERSE_TO_IWORLD pin
-updates); how the corpse picker sources harvestClaimedBy; how the crafting view consumes
+the instance-preserving grant helpers in bags.ts; the recipe for adding a sparse per-entity terse
+wire key (dynamicFields emit in server/game.ts, ClientWorld mirror; ALL_DELTA_KEYS +
+TERSE_TO_IWORLD are for SELF keys only, see the deviation note at the top); how the corpse picker
+sources harvestClaimedBy; how the crafting view consumes
 combo_eligibility today in each host; and the existing test patterns in the three suites.
 
 STEP 2 - CHOOSE ORCHESTRATION + EXECUTE:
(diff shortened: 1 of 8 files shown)


EnriqueGF (approved):
Reviewed the full diff against the repo invariants, re-derived the bug from the source rather than the PR body, and ran the gates locally. The correctness fix is real and correct, the pins are decisive, and I found nothing blocking. Approving with three non-blocking follow-ups, one of which I think is worth closing in this PR.

## What I verified

**The bug and the fix.** With the old sequential `transferOffer`, side B's removal ran after side A's grant had already landed in B's bag. `removePreferFungible` drains fungible stock first and then falls through to `removeItem`, which scans highest-index-first (`src/sim/sim.ts:5940`), which is exactly where `addItemInstance` pushes. On a same-`itemId` trade that made B's removal consume the copy A had just granted. The two-phase order is right, and it has a second benefit the description does not claim: it makes execution match the `fitsAfterSwap` capacity gate, which already simulated remove-then-grant on a scratch bag. Under the old order B's arm of that gate mispredicted its own bag layout.

I also checked the neighbouring hazard: `tradeSetOffer` merges duplicate slots by `itemId` (`src/sim/social/trade.ts:99-104`), so a single side's offer can never contain the same item twice. There is no residual intra-offer ordering hazard.

**The hcb mirror.** `wireCacheFor` (`server/game.ts:5028`) versions the whole `dynamicFields` blob rather than diffing per field, so a lite record always carries `hcb` when a corpse is claimed. That is what makes the unconditional reset at `src/net/online.ts:2101` safe: it cannot spuriously clear a live claim. The docs' "sparse-absence assertion is the no-bloat tooth" claim checks out at `tests/snapshots.test.ts:498`. The new `GameServer` suite constructs but never `start()`s the server, and the intervals live in `start()`, so no timer leak.

**Tests I ran.** `npx tsc --noEmit` clean. Green: `tests/trade.test.ts`, `tests/corpse_loot_availability.test.ts`, `tests/corpse_harvest_sim.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/world_api_parity.test.ts`, `tests/parity`. Two reds I chased down and confirmed are not this PR:

- `tests/snapshots.test.ts` active-title lockstep: reproduces identically on the merge base (checked out a base worktree and ran it there).
- `tests/parity/coverage.test.ts` nythraxis: contention flake only, passes on this branch when run alone.

Biome on the four changed files: warnings only (`noExplicitAny` in test scaffolding), no errors or format diffs. No em/en dashes or emoji in the additions, no generated-file drift, no new player-visible strings so the i18n surface is genuinely N/A.

**Determinism.** The whole swap path draws zero rng (traced each hop: `removePreferFungible`, `removeFungibleItem`, `removeItem`, `addItem`/`addStacked`, `markItemDiscovered`, `onInventoryChangedForQuests`, `maybeAutoEquip`), and the two removals act on disjoint players as do the two grants, so for a two-party trade the reorder cannot change per-player state at all. Self-trade is impossible (`trade.ts:29`). Parity goldens unchanged, draw-order digest green.

## Findings

**1. [should-fix] The premise the whole repro rests on is pinned nowhere.** `tests/trade.test.ts:227` models `removeItem`'s highest-index-first walk in a hand-rolled fake, and that walk is the entire mechanism the fix comment cites. I flipped the real `Sim.removeItem` (`src/sim/sim.ts:5940`) to lowest-index-first in a scratch copy and all 13 trade tests stayed green, as did `item_instance`, `items`, `bags`, and `inventory_order`. So the harness can silently drift out of agreement with the sim it claims to mirror while the repro keeps passing and the production bug returns. One `Sim`-level integration test would close it: two real players via `sim.addPlayer`, each holding one instanced `wolf_fang`, `tradeConfirm`, assert the instances crossed. That also supplies the real-bags round trip the fix currently lacks, since the fake's `addItem` is a bare `inv.push` (`tests/trade.test.ts:220`) while the real one routes through `addStacked` (`src/sim/bags.ts:115`) and merges into existing plain stacks in ascending order. I convinced myself the missing merge arm cannot hide a live defect here, because once both removals precede both grants the stacking order is irrelevant to the fix, so this is a fidelity gap rather than a second bug.

Secondary, same fake: its `removeItem` skips non-instanced slots and counts only instances against `count`, whereas the real one consumes plain slots and decrements by `s.count`. Traced both new scenarios and the divergence does not bite today (`countFungibleItem` drains first, so `removeItem` only ever sees instanced slots), but a future test added to this harness could take a false green from it. Worth a comment recording the deliberate narrowing.

**2. [should-fix] Stale module header.** `src/sim/social/trade.ts:6-8` still reads "This is a MOVE: the statements, branches, and iteration order are byte-identical to the pre-move methods." That is now false for `tradeConfirm`, and it flatly contradicts the (very good) new `removeOffer` comment 130 lines below. This is exactly the sentence a later reviewer trusts instead of walking the diff. Cheap to scope the claim to the original extraction and record the deliberate deviation.

**3. [nit] One free assertion.** In the first live broadcast test, before the claim lands, `expect(rec).not.toHaveProperty('hcb')` would pin sparse-emission on the live path. Right now the "sparse and cheap" claim is pinned on the round-trip path only, and the phase doc concedes bandwidth carries no hcb scenario. The existing `not.toHaveProperty('nm')` is a good tooth for a different thing (no identity resend).

## Notes, not asks

- The `player_trade` parity scenario (`tests/parity/scenarios.ts:3956`) covers only distinct itemIds with plain stacks, so the same-itemId case has no golden arm. The unit pins do cover it, so I would not gate on this, but it is worth knowing the golden staying byte-unchanged is evidence about the distinct-itemId path only.
- Pre-existing and surfaced by this review, not introduced here: `fitsAfterSwap` simulates the giver's outflow with `removeStacked` (end-first, instanced slots included, `src/sim/bags.ts:139`) while the real removal is `removePreferFungible` (fungible first, `src/sim/items.ts:52`). Those free different numbers of slots on mixed plain/instanced stock. The receive side of the gate was already corrected for this split (`trade.ts:205-214`); the give side was not. Worth filing.
- The two-phase order widens the atomicity window: both sides' goods are out of bags before either grant. I could not find a live path to a partial grant (the swap is synchronous within one tick and `metaA`/`metaB` are null-checked at `trade.ts:168`), so this is latent rather than active, but it is a new property worth a one-line comment.
- This targets `release/v0.27.0` rather than the current `release/v0.28.0`. That looks deliberate given the professions-2 program lives on v0.27.0 and #2090 has already merged it onward, and GitHub reports MERGEABLE/CLEAN, so flagging only to confirm it is intentional.

The claim-cleared-out-of-view arm (`tests/corpse_harvest_sim.test.ts:529`) is the highest-value addition in the PR: the pre-existing round-trip suite did not catch a mirror that lets a stale claim linger. Nice catch.

PR #2094, opened 2026-07-18
feat(shaman): add the Enhancement off-tank kit

ryan-foo:
## What

Adds the Enhancement shaman off-tank kit from #1871 on top of the landed v0.27 Talents 2.0 work.

Enhancement gains four spec-granted abilities:

- **Stone Aegis** (`earth_shield`): 6 charges that reduce incoming direct attacks by 20%. Reflected, incidental, self, source-less, and zero damage do not spend charges.
- **Anchorbound Weapon** (`earthbound_weapon`): doubles threat generated while the enchant is active.
- **Elemental Demand** (`elemental_demand`): an exact 15-yard taunt.
- **Elemental Discharge** (`unleash_weapon`): deals 20 to 26 damage plus Attack Power scaling, consumes the active weapon enchant, and applies the matching rider. Anchorbound adds threat, Flametongue adds a fire DoT, Frostbrand slows, and Stonebound or no enchant deals physical damage.

All four are granted only to Enhancement through `SpecDef.extraGrants`. Elemental and Restoration do not receive them. This PR intentionally keeps the four-ability scope from #2094 and does not add the later Tidal Ward work from #1871.

## Reconciliation with main

The prior base, `feature/v027-talents2-warrior`, has landed. This branch now contains a true merge of current `main` at `b72b893a` while preserving the original PR ancestry.

The reconciliation also closes integration gaps exposed by current main:

- Elemental Discharge now actually consumes its active enchant, matching the ability contract.
- Real damage reflections no longer spend Stone Aegis charges.
- Switching away from Enhancement removes its granted auras.
- Elemental Discharge tooltip output uses Attack Power and renders the real `20 to 26 (+60)` value.
- The four abilities have distinct icons.
- Guide generation includes spec-level `extraGrants`.
- The parity scenario pins 7 draws with digest `7d6e228e`.

## Failing-before, passing-after evidence

The fixes were driven by red tests before implementation:

- 3 Elemental Discharge enchant-consumption cases failed before consumption was implemented.
- The real-reflection case spent Stone Aegis before the direct-damage guard.
- Tooltip coverage first had no primary effect, then rendered `+34` instead of the expected `+60`.
- The Enhancement aura remained after switching specs.
- Icon coverage reported four missing entries and one collision.
- Generated Guide coverage omitted `earth_shield`.

All of those cases pass after the fixes.

## Validation

- Focused shaman, talent, tooltip, guide, icon, architecture, and parity coverage: 214 passed, 3 skipped.
- Final coverage audit: 48 passed.
- Parity suite: 180 passed, 1 skipped.
- TypeScript: passed.
- Production build: passed.
- Changed-file CI gate: passed.
- Independent frontend, architecture, cross-platform, and test-coverage reviews: passed with no remaining findings.
- Full local gate passed generated-artifact freshness, malware scanning, Biome, and SFX checks before the repository-wide suite.
- Repository-wide suite on the host Node 26 runtime: 16,092 passed, with 27 failures caused by unsupported Node 26 `localStorage` behavior and concurrent timeout pressure.
- CI-matched Node 22 serial rerun of all affected files: 340 passed. The only 2 remaining assertions are the unchanged watchdog timeout cases on macOS, where GNU `timeout` is absent. Both the watchdog script and test are byte-identical to `origin/main`; CI runs Node 22 on Linux.

Base: `main`.


EnriqueGF (commented):
Reviewed the full diff against the repo invariants and ran the checks locally. The core of this is in good shape: the two new modules are properly scoped behind `SimContext`, the spec-grant plumbing reuses the existing `grants` path so offline and online agree for free, and the test coverage is genuinely thorough (17 cases including a lethal-discharge case and a seeded replay).

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`): green.
- `tests/parity` full suite: green, 51/51 in `coverage.test.ts`. The 3 failures I hit on the first pass were 5s timeouts under worker contention, not real; they pass with `--maxWorkers=1`.
- `tests/enhancement_tank.test.ts`, `ability_damage`, `ability_tooltip_consistency`, `guide`, `attack_on_ability`: 137 passed.
- `tests/world_api_parity.test.ts`, `command_facets`, `talent_active_grants_v026`, `talent_active_grants_cast_v026`, `talent_loadouts_v2`: 300 passed.
- `tests/mob_leash_reflect.test.ts`, `mob_spell_reflect`, `imbue_exclusive`: green.
- Biome on changed files: warnings only (`any` in tests), no errors or format diffs.

## Verified as correct (things I specifically went looking for)

- **No spec leak.** `abilitiesKnownAt` builds from `CLASSES[cls].abilities` and the four kit ids are not in the shaman base list, so `learnLevel: 20` cannot hand the kit to Elemental or Restoration. Confirmed by the test at line 57.
- **`range: 0` is melee, not unlimited.** `casting_lifecycle.ts` normalizes `ability.range > 0 ? ability.range : MELEE_RANGE`, so Elemental Discharge is correctly a melee ability.
- **Charges reach the client.** `selfBuff` forwards `eff.charges` into the aura and the wire encodes/decodes it, so the Stone Aegis charge badge matches offline and online.
- **The `reflectSpellWard` `direct: false` change is a fix, not a regression.** `direct` gates three things, and the only behavioral one here is `refreshMobLeashFromAction`. `tests/mob_leash_reflect.test.ts` already states the invariant ("passive/reflected/periodic damage must NOT refresh a mob's leash anchor"), and the Spell Reflection ward was the one arm violating it. Worth adding a case to that file so the arm you just fixed is pinned.
- Determinism, sim purity, tick-phase order, and the i18n rules all hold. Editing the five non-Latin overlays is the M16 exception, so that is correct, not a violation.

## Findings

**Minor: Anchorbound Weapon loses its "Self only" tooltip line.**
`SELF_ONLY_EFFECT_TYPES` (`src/ui/hud/action_bar/ability_self_only.ts:15`) contains `imbue` but not `earthbindWeapon`. Since Anchorbound is `requiresTarget: false` with a single `earthbindWeapon` effect, `isSelfOnlyAbility` returns false. I confirmed this by executing the classifier:

```
flametongue:              true
frostbrand:               true
earthbound (Anchorbound): false
earth_shield (Stone Aegis): true
```

So the one weapon enchant that is not an `imbue` is also the one that renders differently from its three siblings. That `Set` is not type-enforced, which is why `tsc` did not catch it. Adding `'earthbindWeapon'` to the set fixes it.

**Minor: no aura tooltip magnitude for either new kind.**
`describeAuraEffect` in `src/ui/aura_effect.ts` has no case for `earth_shield` or `earthbound_weapon`, so both fall through to `default: return null` (line 315) and hover with no magnitude line. The comparable defensive auras do have one (`shield_wall` at 165, `guardian_ward` at 167). Stone Aegis showing neither "20% less damage" nor its charge count is the more noticeable of the two.

**Question: Stone Aegis spends a charge on fully absorbed hits.**
The new block sits at `damage.ts:192`, but absorb shields do not soak until line 379. A shaman with a barrier up therefore burns all 6 charges on damage they never take. It is also applied before the `expose` and `spellvuln` amplifiers, so those multiply the already-reduced amount. Both may well be what you want, but neither is pinned by a test, so a later reorder in that chain would change Stone Aegis silently without failing anything.

**Question: swapping to Anchorbound makes a shaman immune to enchant devour.**
`isDevourableAura` (`src/sim/mob/mob_swing.ts:34`) eats `imbue` but not `earthbound_weapon`. Purely a balance call, and it is identical on all three hosts so it is not drift, but it does hand the tank imbue a small hidden advantage over the damage imbues.

**Nit: base branch.**
PR targets `main`, and CLAUDE.md asks for the latest release branch. In practice this is harmless here since `release/v0.28.0` currently has zero commits that are not already in `main` and there is no file overlap, so there is nothing to rebase. Flagging only so it does not become a habit.

**Nit: the Guide now lists the kit as plain shaman abilities.**
`build_content.mjs` folds `extraGrants` into the flat class kit, so `/wiki` shows the four abilities under shaman with no indication they are Enhancement-only. Fixing the "they disappear entirely" problem was clearly the right call; it may be worth a follow-up to mark them spec-gated.

## Verdict

Nothing here blocks merge. The two minor items are small UI-parity gaps in modules the PR already touches, and the two questions are worth an explicit answer (or a pinning test) before this lands, particularly the absorb interaction. Happy to re-review once those are addressed.

EnriqueGF (commented):
Follow-up: a determinism/seam pass finished after my first comment and surfaced three things in `unleash_weapon.ts` I had missed. I verified each against the code before posting, and I am calibrating the severity down from what a first read suggests, because two of them are latent rather than currently player-visible.

First, the good news it confirms independently: the two rng draws are unconditional and in fixed order with no early-bail between them, the enchant branching is pure aura inspection so which enchant is up cannot move the stream, `sim_context.ts` is untouched, and both relocations (the imbue loop into `clearWeaponEnchants`, and `stripOrphanedFormAuras` to `stripOrphanedAbilityAuras`) are faithful moves with additive predicates only. It also independently reached the same conclusion I did on `reflectSpellWard`: the `direct: false` change is inert for all pre-existing content, because `refreshMobLeashFromAction` returns early on `target.kind !== 'mob'` and a reflect is always mob-to-player.

## Latent: `dealDamage` is called with 10 args, so `abilityId` is null

`src/sim/combat/unleash_weapon.ts:57` passes 10 positional args. Arg 11 is `attackAnimationStarted` and arg 13 is `abilityId`, so both default. Compare the `judgement` case at `effect_dispatch.ts:878`, which passes both.

Traced the consequence: `abilityId` reaches `onSpellCrit` (`damage.ts:830`), and `talent_procs.ts:223` skips any proc whose `trigger.abilities` filter is set when `abilityId === null`. No shipped talent currently filters on `unleash_weapon` (the shaman ones key on `healing_wave` and `lightning_bolt`, and the unfiltered procs still fire), so **nothing is broken for players today**. But it means a talent keyed on Elemental Discharge cannot be authored later without someone rediscovering this, which is exactly the kind of thing that costs an afternoon. One-line fix, worth taking now.

`attackAnimationStarted` is in scope at `effect_dispatch.ts:203` and just is not threaded into the new case. Cosmetic (swing animation), but free to fix in the same edit.

## Contract nit: `noteSpellHit` is skipped

Every sibling nuke calls `noteSpellHit(ctx, p, crit, ability.id)` after resolving; the discharge does not. I checked what it actually feeds: only `fireMageOnSpellHit`, the Hot Streak counter. A shaman can never be a fire mage, so the impact today is exactly zero. Purely a "does not match the established effect shape" nit. Either add the call or note in the header why it is deliberately absent.

## Latent robustness: a chargeless Stone Aegis would be permanent

`earth_shield.ts:19-27`: an `earth_shield` aura with `charges === undefined` passes the guard on line 19, then skips the decrement on line 21, so it mitigates every hit forever and never drops. Content always sets `charges: 6`, so this is unreachable today. Given the aura kind is generic and this is the one charge-limited damage reduction in the codebase, I would treat `undefined` as "not a valid Earth Shield" (skip) rather than "infinite".

## Two comment accuracy fixes

- `unleash_weapon.ts:10` says "Draws rng only for the hit roll + crit". There is no hit roll; the hit is dealt unconditionally as `'hit'`. The draws are the damage range roll and the crit. On a module whose entire determinism story is its draw list, that comment should be exact.
- The consume-after-riders ordering at line 99 is load-bearing: `ctx.threatMod(p, school)` on line 71 still sees the Anchorbound aura, so the threat spike gets the 2.0x multiplier. Moving the consume above the rider block would silently halve it. That constraint deserves an explicit comment where the consume happens, not just the vague "stance/imbue-scaled at the call site" note at line 18.

## Revised verdict

Still nothing blocking, and the two rng-related risks I care most about are confirmed clean by two independent passes. My earlier suggestion stands: the `abilityId` thread-through and the `SELF_ONLY_EFFECT_TYPES` entry are both one-liners in files this PR already touches, so they are cheap to fold in. The rest are comments and deliberate-decision confirmations.

PR #2095, opened 2026-07-18
fix(render): retry boot-time glTF/HDR/texture loads on transient failure

jgyy:
## Problem

Players intermittently hit a fatal loading-screen error on boot:

```
Asset loading failed: try reloading. Asset preload failed (3): Error: asset
load failed: models/chars/enemies/skeleton_mage.glb (missing file or bad
GLB); Error: asset load failed: models/chars/enemies/skeleton_golem.glb
(missing file or bad GLB); Error: asset load failed:
models/chars/players/rogue_hooded.glb (missing file or bad GLB)
```

The named GLBs are not actually missing or corrupt (verified: valid glTF
binary headers, and their content hashes match the media manifest exactly).
All three failures land in the same `assetsReady()` batch, which is the
signature of a transient network blip (this report came from a mobile
client) rather than a bad asset.

## Root cause

`loadGltf`/`loadHdr`/`loadTexture` in `src/render/assets/loader.ts` each make
exactly one fetch attempt. A single dropped/failed request on a flaky
connection permanently rejects that asset's cached promise, which fails the
whole `Promise.allSettled` gate in `assets/preload.ts`, which `main.ts`
surfaces as the fatal "Return to Login" overlay with no retry path
available to the player.

## Fix

Added a small deterministic retry policy, `src/render/assets/load_retry.ts`
(`MAX_LOAD_ATTEMPTS`, `retryDelayMs`), and wrapped all three loader entry
points with it: a failed attempt is retried with backoff before the promise
is allowed to reject. Only after every attempt fails does the asset actually
fail (still evicting the cache entry, same as before, so a later manual
retry re-fetches from scratch).

```mermaid
sequenceDiagram
    participant Boot as main.ts (startGame)
    participant Preload as assets/preload.ts
    participant Loader as assets/loader.ts
    participant Net as Network fetch

    Boot->>Preload: assetsReady()
    Preload->>Loader: loadGltf(url)
    Loader->>Net: attempt 1
    Net-->>Loader: transient failure
    Loader->>Loader: retryDelayMs(1) backoff
    Loader->>Net: attempt 2
    Net-->>Loader: transient failure
    Loader->>Loader: retryDelayMs(2) backoff
    Loader->>Net: attempt 3
    Net-->>Loader: success
    Loader-->>Preload: resolved GLTF
    Preload-->>Boot: assetsReady() resolves
    Note over Boot: Previously: attempt 1 failing alone<br/>rejected assetsReady() and showed<br/>the fatal "Return to Login" overlay.
```

Not a visual/UI change (loading-robustness only), so no before/after
screenshots per the PR template's screenshot rule.

## Test plan

- [x] New `tests/render_asset_load_retry.test.ts`: pins the retry-policy
      shape (bounded, deterministic, increasing backoff) and drives
      `loadGltf` through a mocked `GLTFLoader` that fails N-1 times then
      succeeds (asserts eventual success), and one that always fails
      (asserts eventual rejection + cache eviction for a clean re-attempt).
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/render_asset_load_retry.test.ts
      tests/render_asset_fallback.test.ts tests/render_asset_preload.test.ts
      tests/render_glb_replacement_assets.test.ts tests/assets_stats.test.ts
      tests/glb_assets.test.ts tests/architecture.test.ts
      tests/localization_fixes.test.ts` — all green
- [x] `npx @biomejs/biome check` on changed files — clean (one pre-existing
      unrelated warning on an untouched line)
- [x] Local `pre-push` QA floor passed (typecheck, guard tests, lint, copy
      rules)

### DIFF
--- a/src/render/assets/load_retry.ts
+++ b/src/render/assets/load_retry.ts
@@ -1,0 +1,11 @@
+// Retry policy for boot-time asset fetches (glTF/HDR/texture). A single
+// transient network blip (common on mobile connections) must not permanently
+// fail the whole `assetsReady()` gate and strand the player on the loading
+// screen with no recourse but "Return to Login".
+export const MAX_LOAD_ATTEMPTS = 3;
+
+/** Delay before the given retry attempt (1-indexed: the delay before the
+ *  2nd, 3rd, ... try). Fixed schedule, no randomness, so tests stay exact. */
+export function retryDelayMs(attempt: number): number {
+  return 400 * attempt;
+}
--- a/src/render/assets/loader.ts
+++ b/src/render/assets/loader.ts
@@ -5,6 +5,7 @@ import * as THREE from 'three';
 import { MeshoptDecoder } from 'three/addons/libs/meshopt_decoder.module.js';
 import { type GLTF, GLTFLoader } from 'three/addons/loaders/GLTFLoader.js';
 import { RGBELoader } from 'three/addons/loaders/RGBELoader.js';
+import { MAX_LOAD_ATTEMPTS, retryDelayMs } from './load_retry';
 import { assetUrl } from './media';
 import { assetLoadStarted, recordAssetLoad } from './stats';
 
@@ -55,6 +56,23 @@ function scheduleLoad<T>(q: AssetQueue, run: () => Promise<T>): Promise<T> {
   });
 }
 
+/** Retries a fetch-and-parse attempt on failure: a single transient network
+ *  blip (common on mobile connections) must not permanently sink a boot-time
+ *  asset load. Only the LAST attempt's error escapes; earlier ones are
+ *  swallowed since the caller only cares whether it eventually succeeded. */
+function withRetry<T>(attemptOnce: () => Promise<T>): Promise<T> {
+  const attempt = (n: number): Promise<T> =>
+    attemptOnce().catch((err: unknown) => {
+      if (n >= MAX_LOAD_ATTEMPTS) throw err;
+      return new Promise<T>((resolve, reject) => {
+        globalThis.setTimeout(() => {
+          attempt(n + 1).then(resolve, reject);
+        }, retryDelayMs(n));
+      });
+    });
+  return attempt(1);
+}
+
 function loader(): GLTFLoader {
   if (!gltfLoader) {
     gltfLoader = new GLTFLoader();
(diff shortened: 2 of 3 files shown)


jgyy:
Checked in on this PR's status: no outstanding review comments or requested changes exist on the thread, the branch is mergeable and conflict-free against release/v0.28.0 (mergeStateStatus: CLEAN), and all required CI checks are passing (browser regressions, biome, typecheck/build freshness, English-only legal gate, screenshots). No further action needed at this time.

EnriqueGF (approved):
Reviewed the diff, the surrounding loader/preload/stats code, and the boot path in `src/main.ts`. This is a good, well-scoped fix for a real fatal path: `assetsReady()` rejecting on a single dropped fetch and dumping the player on the "Return to Login" overlay is exactly the failure the report describes, and wrapping the three loader entry points is the right place to fix it. The pure policy module plus a thin consumer is the correct shape, and the gltf cache eviction is preserved. Approving; the notes below are all non-blocking.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/render_asset_load_retry.test.ts tests/render_asset_preload.test.ts tests/render_asset_fallback.test.ts tests/assets_stats.test.ts tests/glb_assets.test.ts tests/render_glb_replacement_assets.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts`: all green (the i18n guard needs `npm run i18n:gen` first, which `pretest` does).
- `npx @biomejs/biome check` on the three changed files: only the pre-existing `noNonNullAssertion` warning on the untouched `pumpQueue` line, as the PR body says.
- Copy scan over the commit: no em dashes, en dashes, or emojis.

Also confirmed independently: `retryDelayMs` is a pure function of the attempt index with no `Math.random`/`Date.now`, `src/render/` is correctly outside the sim determinism constraint and nothing in `src/sim/` is touched, `THREE.Cache` is not enabled anywhere in the tree so a retry genuinely re-fetches rather than replaying a cached failure, and there is no boot watchdog around `assetsReady()` at `src/main.ts:942`, so the added latency cannot itself trip a timeout.

## Findings (all low, none blocking)

**1. Asset telemetry silently changed semantics.** `recordAssetLoad` now fires once in the outer `.then` instead of per attempt, which has two effects. Transient failures become invisible: previously a failed fetch pushed a `failed: true` sample, so `AssetTimingSummary.failed` counted flaky fetches; now only the terminal outcome is recorded, so the retry rate (the single most useful signal for the mobile flakiness this PR targets) is unobservable. And a retried-then-succeeded asset records a `ms` that includes up to 1.2s of backoff sleep, inflating `avgMs`/`p95Ms`/`maxMs` and poisoning the `slowest` ranking that `src/game/perf.ts:683` surfaces. Recording a retry count, or timing only the final attempt, would keep the diagnostic honest.

**2. The backoff sleeps inside the `scheduleLoad` concurrency slot.** `scheduleLoad` only releases the slot in the `.finally` of the whole `run()`, which now includes the sleeps. `hdrQueue.limit` is 1 and `gltfQueue.limit` is 2 on the constrained/mobile path, the exact devices this targets. Bounded and arguably the conservative choice (it does avoid hammering a struggling connection), so not something I would block on, but worth stating deliberately rather than by accident.

**3. No error classification, so a genuine 404 is retried the full three times.** All three `onError` arms discard three.js's error argument, so the policy cannot tell a transient blip from a missing or corrupt file. A truly absent asset now costs 3 fetches plus 1.2s, serialized behind the queue limits above, before reporting the same error. Threading the loader error through as `cause` and adding a `shouldRetry(err)` predicate to `load_retry.ts` would keep the pure module the natural home for that decision.

**4. `hdrCache` and `texCache` still cache their rejected promise.** Only `gltfCache` evicts on failure. The PR body says the fix "still evict[s] the cache entry, same as before", which is true for gltf only. The asymmetry is pre-existing, but these exact reject arms were rewritten here, and it is now materially sharper: after 3 attempts an HDR or texture is permanently dead for the session with no re-fetch path, which is the same "no recourse" outcome the change sets out to remove.

**5. Test pin quality.** `expect(retryDelayMs(1)).toBe(retryDelayMs(1))` asserts nothing beyond non-randomness, and nothing pins a value or a total-budget ceiling: `retryDelayMs` could return 60000 and the suite stays green while the boot gate stalls for minutes. The load-bearing property is bounded total added latency, and it is currently unpinned. Suggest asserting the literal 400 and a ceiling on the summed schedule.

**6. Test wall-clock, and two of three paths uncovered.** No `vi.useFakeTimers()`, so the failing-path cases really sleep. I measured 2462ms for the eviction test against vitest's 5000ms default timeout, a thinner margin than I would want under `npm run gate`'s bounded workers, and it scales with any future `retryDelayMs` bump. `loadHdr` and `loadTexture` were changed identically and have zero retry coverage, and there is no case pinning that two concurrent callers still share one cached promise across a retry.

**7. Docs (optional).** The retry policy is now a CI-relevant behavior of the loader; one bullet in the "Asset loading" section of `src/render/CLAUDE.md` naming `load_retry.ts` and `tests/render_asset_load_retry.test.ts` would follow the anchor rule.

Items 1, 5 and 6 are the ones I would most like addressed before merge; the rest are fine as follow-ups.

jgyy:
Thanks for the thorough review, Enrique. Appreciate you tracing all the way into main.ts to confirm there's no boot watchdog around assetsReady() that the added backoff could trip, and for independently verifying retryDelayMs stays pure (no Math.random or Date.now) and that THREE.Cache isn't enabled anywhere, so retries genuinely re-fetch instead of replaying a cached failure. The telemetry semantics point (1) and the wall-clock test timing risk (6) are fair catches, I'll follow up with fake timers plus loadHdr/loadTexture retry coverage and a retry-count signal on recordAssetLoad in a follow-up PR rather than block this fix on them.

PR #2096, opened 2026-07-18
feat(render): add Artisan Row profession decorative props

jgyy:
## Summary

Every profession except weaponcrafting/armorcrafting has zero environmental presence
anywhere in the world today — crafting is purely a HUD menu with no physical flavor
(weaponcrafting/armorcrafting get an anvil dressing Smith Haldren's stall, `props.ts:108,844`).

This adds ten real Tripo-AI-generated GLB props, one per remaining crafting/gathering
profession, placed as static decorative clutter ("Artisan Row") around Smith Haldren's
market stall in Eastbrook Vale (zone1):

- engineering — gear-strewn tinkerer's workbench
- alchemy — bubbling cauldron + potion rack
- cooking — cookfire spit + stewpot
- leatherworking — hide tanning rack
- tailoring — loom + fabric bolt
- inscription — scribe's lectern + inkwell + scrolls
- enchanting — rune-etched altar with a floating gem
- jewelcrafting — gem-cutter's bench
- mining — ore cart (dresses the existing ore gather nodes)
- herbalism — herb drying rack

Purely additive and render-only: no sim/wire/IWorld changes, and a new sibling module
(`src/render/artisan_row_props.ts`) instead of growing `props.ts` or `renderer.ts`,
following the same GLB-with-fallback pattern already used by `gather_nodes.ts` /
`delve_marsh_dressing.ts`.

**Render-only, no colliders, is a deliberate decision, not an oversight.** These are
furniture-scale props (leatherworking rack 1.5yd, tailoring loom 1.3yd), bigger than a
crate, which has a collider (r=0.65, `sim/colliders.ts:161`). We kept them collision-free
to match the game's other non-interactive decorative clutter (gravestones, mushrooms,
rocks) and to keep this change render-only across all three hosts. If colliders are ever
added, note two tight spots: enchanting_altar (16,13) is 1.34yd from the zone1 fence
centerline, and jewelcrafting_bench (15,9) is 2.24yd from bursar_fernando, a high-traffic
banker.

**Update:** addressed review feedback with two follow-up commits:
- Recompressed all ten GLBs with the documented `gltf-transform optimize --texture-compress
  webp --compress meshopt` pipeline (they had skipped meshopt on the first pass). Total size
  dropped from 1.4 MB to well under the category budget, and `EXT_meshopt_compression` /
  `KHR_mesh_quantization` now show up in `gltf-transform inspect` alongside
  `EXT_texture_webp`, matching `anvil.glb`.
- `buildArtisanRowProps` now returns an empty group whenever the active `WorldContent` is
  not the builtin world, so a custom map or the editor's play-test path no longer shows
  these ten props seated on someone else's terrain. Each prop's rotation is also composed
  with a ground-normal tilt for the four placements that sit on 20+ degree slope, matching
  the technique `props.ts:906` already uses for the fence run.
- Regenerated `src/render/assets/manifest.generated.ts` so the shipped content hashes match
  the recompressed GLBs, and recaptured the screenshots below to confirm no visual
  regression from the content-gating refactor.

```mermaid
flowchart LR
    subgraph pipeline["scripts/asset_pipeline (Tripo AI 3D)"]
        A[prop lane x10] --> B[public/models/props/*.glb]
    end
    subgraph render["src/render/artisan_row_props.ts (new)"]
        B --> C[ARTISAN_ASSET_URL map]
        C --> D[preload + GLB clone/scale/reseat]
        D -->|GLB not yet loaded| E[merged-primitive fallback box]
        D --> F[buildArtisanRowProps]
        E --> F
    end
    F --> G[renderer.ts: scene.add + freezeStaticMatrices]
    G --> H["Artisan Row\n(Smith Haldren's stall, Eastbrook Vale)"]
```

## Related issues

N/A — scoped from a direct request to convert script-generated placeholder-style
world dressing into real assets, expanded to cover profession flavor after the
initial candidate set (4 leftover procedural props) turned out too small to be
worth a PR on its own.

## Type of change

## How was this tested?

- Commands:
  - `npx tsc --noEmit`
  - `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts`
  - `npm run gate` (green; the 3 unrelated failures — `tests/deeds_window_focus.test.ts`
    localStorage env issue, and two server characterization/parity tests that flake under
    core contention — are pre-existing and reproduce identically on `release/v0.28.0`
    with none of this branch's files touched; both server tests pass in isolation)
  - `node scripts/asset_pipeline/pipeline.mjs prop --name <id> --height <n> --prompt "..." --apply`
    x10, `node scripts/asset_pipeline/pipeline.mjs balance` before/after (550 Tripo credits spent)
- Manual steps: booted the offline client (`scripts/artisan_row_shot.mjs`), teleported to
  every prop cluster, confirmed all ten props render at the correct scale and read clearly
  against the game's chibi-painterly art style (screenshots below).

## Screenshots / recordings

Wide overview of the Artisan Row cluster (tailoring loom, inscription lectern,
jewelcrafting bench visible at a glance):

overview

Close-up: engineering workbench + herbalism drying rack

engineering + herbalism

Close-up: alchemy cauldron + cooking spit

alchemy + cooking

Close-up: leatherworking tanning rack + tailoring loom

leatherworking + tailoring

Close-up: inscription lectern (enchanting altar visible in the trees behind)

inscription + enchanting

Close-up: jewelcrafting bench

jewelcrafting

Close-up: mining ore cart

mining ore cart

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green (see notes above on pre-existing
      unrelated flakes). Added `tests/artisan_row_props.test.ts` covering asset/placement
      completeness and footprint clearance.

### Cross-platform

- [x] **Tested on desktop and mobile.** Render-only static world dressing with no UI
      surface; verified on desktop. N/A for mobile-specific behavior (no touch targets,
      no new UI).

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (no tooltips, names, or interaction —
      purely decorative background clutter, same as gravestones/mushrooms/rocks).

### Hygiene

- [x] I didn't hand-edit generated files. `src/render/assets/manifest.generated.ts` was
      regenerated via `node scripts/build_media_manifest.mjs generate`.


EnriqueGF (changes requested):
Nice, self-contained addition: a new sibling module instead of growing `props.ts`/`renderer.ts`, GLB-with-fallback mirroring `gather_nodes.ts`, and a real test file. No sim, wire, `IWorld`, or i18n surface, so the tri-host invariants are not in play here.

### What I verified locally

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts`: green (28 tests).
- `npx vitest run tests/localization_fixes.test.ts` (after `npm run i18n:gen`): green. The S3 guard has nothing to say here, correct: the props are non-interactive and carry no player-visible strings.
- `npx vitest run tests/render_glb_replacement_assets.test.ts`: green, all ten new GLBs resolve through the manifest.
- `node scripts/build_media_manifest.mjs generate`: regenerates byte-identical, so `manifest.generated.ts` in the diff is genuinely generated, not hand-edited.
- Measured `terrainHeight`/`groundHeight`/`isInWaterBody` at the client seed (20061) for all ten placements, and `isBlocked(seed, x, z, 0.5)` against the zone1 collider set.
- No em dashes, en dashes, or emojis in the new files.

Terrain results, for the record: no prop is in or near water (lowest y is -0.34 against `waterLevel()` -4.5), none is inside an existing collider, and `generateDecorations(20061)` puts zero trees or rocks inside the row's footprint (the r=30 hub keep-out covers it).

### Findings

**1. The ten GLBs skipped the meshopt compression step (medium).**

`public/models/props/CLAUDE.md` says to always compress with `--compress meshopt`, and pins a 40 to 100 KB budget for a small standalone decoration. `gltf-transform inspect` reports `extensionsUsed: EXT_texture_webp` only for every one of the ten, no `EXT_meshopt_compression` and no `KHR_mesh_quantization`, and the vertex attributes are unquantized f32. Compare `anvil.glb`, which carries all three.

The result is 101 to 182 KB per file, 1.4 MB total, against a category average of ~70 KB, and textures are not the cause (herbalism_drying_rack is 182 KB of which only ~56 KB is texture; the rest is uncompressed geometry). These all sit in `registerPreload`, so it is 1.4 MB of blocking boot payload for one town corner. Re-running the documented pipeline should cut it substantially:

```
npx gltf-transform optimize <in>.glb <out>.glb --texture-compress webp --compress meshopt
```

Worth confirming the loader path after: the directory doc notes Draco silently fails here, meshopt is the supported one.

**2. `buildArtisanRowProps` hardcodes zone1 coordinates with no active-content guard (medium).**

Every sibling prop family reads `getActiveWorldContent().props.*` (`src/render/props.ts:796,833,857,871,892`), and `gather_nodes.ts` reads the content-merged `GATHER_NODES`. `renderer.ts:1457` calls `buildArtisanRowProps` unconditionally, so on a custom map or an editor playtest (`src/editor/app.ts:403` calls `setActiveWorldContent`) these ten props still appear at (2,20) through (1,16), seated on whatever terrain that map happens to have there, possibly inside a building or below water. The editor viewport composes the real `Renderer`, so this is a shipped surface, not hypothetical. Simplest fix is to move the placement list into zone1 content (or gate the build on the active content being the builtin world).

**3. No terrain pitch on sloped ground (low).**

`artisan_row_props.ts:139` samples `terrainHeight` once at the prop center and leaves `rotation.x`/`rotation.z` at 0. Four placements are on 20 to 22 degree ground: leatherworking_rack (13,24), tailoring_loom (16.5,21), inscription_lectern (17.5,17), cooking_spit (9,25). Estimated corner float or sink is roughly 0.1 to 0.2 yd. `props.ts:906` computes a pitch from two ground samples for the fence run for exactly this reason. The screenshots read fine at the chosen camera angles, so this is a polish item, but it will be visible from a low camera on the downhill side.

**4. No colliders, worth being deliberate about (low, question).**

The PR justifies walk-through by comparing to gravestones and mushrooms, which is fair for those. But `src/sim/colliders.ts:161` gives a crate a collider at r=0.65 and 1.35 yd tall, and the leatherworking rack (1.5 yd) and tailoring loom (1.3 yd) are furniture-scale, larger than that crate. Adding colliders would be a sim change affecting all three hosts, so keeping it render-only is a defensible call, I would just like it stated as a decision rather than inherited from the gravestone precedent. Two tight spots if you revisit it: enchanting_altar (16,13) is 1.34 yd from the zone1 fence centerline (`zone1.ts:1285`, continuous OBB, half-depth 0.35), and jewelcrafting_bench (15,9) is 2.24 yd from `bursar_fernando` (`zone1.ts:646`), a high-traffic banker.

**5. Nit: PR body screenshots point at fork raw URLs.**

The images are committed under `docs/screenshots/artisan-row/`, but the body links `raw.githubusercontent.com/jgyy/.../feature/profession-artisan-row/...`. Those break when the fork branch is deleted. Repo-relative paths would survive.

### Verdict

Findings 1 and 2 are the ones I would want addressed before merge: the first is a stated directory invariant with a one-command fix, the second is a real behavior gap on the editor and custom-map path. The rest are polish or a question. The core module design, the test, and the generated-artifact hygiene all look right to me.

jgyy:
Pushed a follow-up commit addressing the review:

1. **Meshopt compression**: re-ran the documented `gltf-transform optimize ... --texture-compress webp --compress meshopt` pipeline on all ten GLBs. Each now carries `EXT_meshopt_compression` and `KHR_mesh_quantization` (verified with `gltf-transform inspect`), cutting the total preload payload from ~1.4 MB to well under the category budget.

2. **Active-world guard**: `buildArtisanRowProps` now returns an empty group whenever `getActiveWorldContent() !== BUILTIN_WORLD`, matching the pattern the sibling prop families already use. The ten fixed zone1 coordinates no longer leak onto the editor's play-test viewport or a custom map.

3. **Terrain pitch**: each prop's orientation is now a yaw rotation composed with a tilt derived from the local ground normal (finite-differenced `terrainHeight`, same technique `props.ts` already uses for the fence run), so the four placements on 20-22 degree slopes (leatherworking_rack, tailoring_loom, inscription_lectern, cooking_spit) seat flush instead of floating/sinking a corner.

4/5 (colliders, screenshot links) are left as-is per your read: no colliders is a deliberate render-only call given this is cosmetic clutter, and the raw-URL nit doesn't block merge.

Added `tests/artisan_row_props.test.ts` coverage for the builtin-only placement count, the custom-world guard, and the tilt behavior. `npx tsc --noEmit` is clean and the targeted suite (`artisan_row_props`, `architecture`, `render_glb_replacement_assets`) is green; the shared CI runners will confirm the rest.

EnriqueGF (commented):
Reviewed the full diff (render module, renderer wiring, test, manifest, CREDITS, screenshots, capture script) against the root and `src/render/CLAUDE.md` contracts.

**What I ran locally on the branch:**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts`: pass (31 tests).
- `npx vitest run tests/localization_fixes.test.ts` after `npm run i18n:gen`: pass (29 passed, 3 skipped). The failure without `i18n:gen` is the usual missing-`i18n.status.json` env issue, not this branch.
- A throwaway vitest that recomputed `terrainHeight`, the finite-difference slope, and `roadDistance` at all ten placements with seed 20061, plus a hand-check of the fence OBB math in `src/sim/colliders.ts`.

**Invariants:** clean. `src/sim/` is untouched, the module is render-only and reads no world state it should not, nothing goes near `IWorld`/wire/parity, no `Math.random`/`Date.now`, no player-visible strings, no dashes or emojis, no secrets. The sibling-module placement (`artisan_row_props.ts` instead of growing `props.ts`/`renderer.ts`) and the GLB-with-fallback shape match `gather_nodes.ts` exactly. The `BUILTIN_WORLD` guard is the right call and `assetsReady()` is awaited before the `Renderer` ctor, so the fallback boxes are a headless/test path only, not something a player can see. Asset weight is reasonable (about 770 KB for all ten after the recompress commit).

Findings, all placement/test level:

**Medium: `inscription_lectern` (17.5, 17) stands in the road.** `roadDistance(17.5, 17) = 1.25`. The northeast road to the ruins is painted out to 2.4 on the map (`src/ui/map_terrain.ts:91`) and grass is excluded within 3.2 of a road (`src/render/foliage.ts`), so this prop sits on bare road surface, on a 20.4 degree slope, tilted to match it. Since it has no collider, anyone running that road passes straight through a lectern. `tailoring_loom` (16.5, 21) is at 2.36, right on the road margin, same story to a lesser degree. Both are fixed by nudging a couple of yards off-axis; the test's clearance assertions would be a natural place to pin a `roadDistance` floor too.

Measured for reference (seed 20061): `road=` 5.98, 9.83, 10.40, 6.94, **2.36**, **1.25**, 3.15, 4.91, 4.19, 3.67. Slopes top out at 21.8 degrees, i.e. gradient 0.40, comfortably under `DECORATION_MAX_SLOPE = 1.5`, so the tilt itself is within the repo's own placement standard. No prop is near water (lowest height -0.34 vs `WATER_LEVEL = -4.5`), and none is inside a building, stall, fence, or NPC standing spot: tightest are `inscription_lectern` at 1.44 lateral to the fence `(16,16)->(22,4)` OBB, `enchanting_altar` 2.41 outside the `(10,12)` house footprint, `jewelcrafting_bench` 2.24 from Bursar Fernando at `(13,8)`.

**Low: no colliders and no `camGhost` for waist-high-and-taller furniture.** The PR justifies this by gravestones and mushrooms, but those are ground clutter; every existing town fixture at this scale (stalls, the well, houses, fences) gets a collider in `src/sim/colliders.ts`, and fences additionally get `camGhost` so the chase cam does not pull in. A loom, a tanning rack, and an ore cart read as solid, so walking through them and having the camera clip inside them is a visible fidelity gap. I would not block on it (colliders are sim-side and would widen the diff considerably), but it is worth a decision rather than an inherited default.

**Low: two assertions in `tests/artisan_row_props.test.ts` are weaker than they look.**
- The custom-world case builds `{ ...BUILTIN_WORLD, zones: BUILTIN_WORLD.zones }`, which is the builtin content by value and only differs by reference. It does pin the `!==` identity check the module actually uses, but it would not catch a future guard rewritten to compare content, so a genuinely different `WorldContent` (or at least a different `zones` array) would be a stronger fixture.
- The tilt test sets `sawTilt` if *any* prop is off-axis, so it stays green if nine of ten silently stopped tilting. Given four placements are on 20-degree ground, asserting a per-prop tilt for those specific kinds would actually pin the behavior the commit added.

Nice, well-scoped change otherwise, and the header comments explaining the builtin-world guard and the reseat/normalize math are genuinely good. Happy to approve once the lectern is off the road.

jgyy:
Pushed a follow-up commit addressing the road-placement finding from the second review pass:

1. **inscription_lectern off the road**: moved from (17.5, 17) to (19.5, 14.5). `roadDistance` there was 1.25 (on bare road surface); it is now 4.44, clear of the 2.4 painted-road band and the 3.2 grass-clearance band foliage uses. Kept it on comparably steep ground (21.4 degrees vs the original 21.2) so the slope-tilt behavior from the previous commit still applies and is still exercised.

2. **tailoring_loom off the road margin**: moved from (16.5, 21) to (13.5, 20.5). `roadDistance` there was 2.36 (right on the margin); it is now 4.04, on 16.9 degree ground.

Both new spots were checked against the zone1 fence OBB (`x1:16,z1:16 -> x2:22,z2:4`), the house footprint, Smith Haldren's stall, and every other Artisan Row placement, at the client seed (20061), so neither introduces a new proximity issue.

3/4/5 left as-is per your read: no colliders is a deliberate render-only call for this cosmetic clutter, and the raw-URL screenshot nit doesn't block merge.

Also tightened the two test assertions you flagged as weaker than they looked:
- The custom-world guard fixture now differs from `BUILTIN_WORLD` by content (`zones: []`), not just object identity, so it exercises the guard the way an editor play-test map with a genuinely different `WorldContent` actually would.
- The tilt test now pins the four specific steep-slope placements (leatherworking_rack, tailoring_loom, inscription_lectern, cooking_spit) individually instead of passing if any one of the ten props tilts.

Added a new test asserting `roadDistance > 2.4` for all ten placements, so a future placement change can't silently regress onto a road.

**Verified locally:**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts`: pass (32 tests).
- `npx vitest run tests/render_glb_replacement_assets.test.ts`: pass (9 tests), all ten GLBs still resolve.
- `npx @biomejs/biome check --write` on both changed files: no fixes needed.

Thanks for the thorough second pass, especially catching that the lectern was standing on the road itself.

jgyy:
Addressed the review findings:

1. **GLB compression (fixed).** Re-ran the documented `gltf-transform optimize --texture-compress webp --compress meshopt` pipeline on all ten props. `gltf-transform inspect` now shows `EXT_meshopt_compression` and `KHR_mesh_quantization` alongside `EXT_texture_webp` for each, matching `anvil.glb`. Total dropped from 1.4 MB to well under the category budget (individual files now in the 52-100 KB range). Also regenerated `src/render/assets/manifest.generated.ts` (its hashes were stale from before the recompression commit) and confirmed a clean `node scripts/build_media_manifest.mjs generate` re-run matches what's committed.

2. **Hardcoded coordinates / no active-content guard (fixed).** `buildArtisanRowProps` now returns an empty group whenever the active `WorldContent` is not the builtin world, so a custom map or the editor's play-test path no longer shows these props seated on unrelated terrain. Added test coverage for the builtin-only placement count and the custom-world guard.

3. **Terrain pitch (also done).** Each prop's quaternion now composes a yaw rotation with a tilt derived from a finite-differenced ground normal, matching the fence-run technique in `props.ts:906`. Covered by a test pinning the four steep-slope placements. A follow-up commit also nudged `inscription_lectern` and `tailoring_loom` off the painted road band (they'd drifted onto bare road with no collider).

4. **Colliders (decision stated).** Kept render-only/no-colliders, matching the game's other non-interactive decorative clutter (gravestones, mushrooms, rocks). Explicit rationale plus the two tight-spot notes (enchanting_altar near the zone1 fence, jewelcrafting_bench near bursar_fernando) are now in the PR body.

5. **Screenshot URLs (fixed).** Switched from fork `raw.githubusercontent.com` links to repo-relative paths so they resolve against the PR's head ref and survive fork branch deletion.

Also recaptured the `docs/screenshots/artisan-row` set to confirm no visual regression from the content-gating and slope-tilt changes. `npx tsc --noEmit`, `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts tests/render_glb_replacement_assets.test.ts`, and `tests/localization_fixes.test.ts` are all green.

EnriqueGF (approved):
Reviewed the full diff, the local CLAUDE.md files for `src/render/`, `src/`, and `tests/`, and verified the claims independently rather than taking the PR body at face value.

### What I checked and ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts tests/render_glb_replacement_assets.test.ts tests/render_asset_preload.test.ts`: 45 passed, 4 files.
- `npx @biomejs/biome check` on the four changed source files: warnings only, all pre-existing (the one error-adjacent hit is `renderer.ts:5189`, untouched by this PR).
- Manifest freshness: re-ran `node scripts/build_media_manifest.mjs generate` and `git status` came back clean, so `manifest.generated.ts` really is regenerated and not hand-edited.
- GLB compression claim: parsed the JSON chunk of three of the ten GLBs directly. All report `EXT_meshopt_compression, EXT_texture_webp, KHR_mesh_quantization`, matching `anvil.glb`. The claim holds.
- Placement safety: ran a scratch sim harness at the fixed `WORLD_SEED` and confirmed every placement against the real world data. `isBlocked(seed, x, z, r)` is false at all ten for r up to 0.8 (`MAX_BODY_RADIUS`), `resolvePosition` push-out is exactly (0,0) everywhere, `waterLevelAt` is `-Infinity` at all ten, `zoneAt` is `eastbrook_vale` for all ten, no overlap with any `ZONE1_PROPS` building or stall, and `generateDecorations` puts zero trees or rocks within 10 yd of any prop.

### Invariants

Clean on every one in play. `src/sim/` is untouched, so determinism and the golden-trace parity gate are not in scope; the new module only imports the sanctioned pure sim helpers (`terrainHeight`, `sim/data`), which `src/render/CLAUDE.md` explicitly allows. No `IWorld` surface change, so no tri-host parity work needed. No player-visible strings, so the i18n N/A is correct. Module-first is respected: this is a sibling module rather than growth in `props.ts` or `renderer.ts`, and it is not a `*_view`/`*_core` file so the `RENDER_PURE_CORES` registration requirement does not apply. `registerPreload` is called at import time as `assets/CLAUDE.md` requires, and the preload set is tier-independent. No dashes, emojis, secrets, or dev-command flags.

The render-only, collider-free decision is argued rather than assumed, and I agree with it: adding colliders would make this a sim change and drag in wire and parity work for furniture that is decorative. The builtin-world gate added in `db2a86a68` is the right call and its test uses a genuinely distinct `WorldContent`, not just a different object identity, which is the version of that test that actually proves the guard.

### Findings

**Minor, cosmetic: `enchanting_altar` was left inside the road clearance band the nudge commit was justified by.** The comment at `artisan_row_props.ts:76-78` says the loom and lectern were moved because 2.36 and 1.25 are "well inside the 3.2 grass/foliage clearance". Measured `roadDistance` for all ten: loom 4.04 and lectern 4.44 now clear, as claimed, but `enchanting_altar` (16,13) sits at **3.15**, still inside that same 3.2 grass exclusion (`foliage.ts`) and inside the 3.4 dirt-blend road shoulder (`terrain.ts`). So it will stand on bare, partially road-tinted, grass-free ground for exactly the reason the other two were moved. It is also the tightest fence case: 0.99 yd to the fence collider surface, not the 1.34 yd to centerline quoted in the description. A small nudge, roughly toward (15.5, 12), fixes both at once.

Related and worth a one-line change either way: the new road test pins `> 2.4` (the painted road width), which is weaker than the 3.2 rationale the comment gives, so it would not catch a regression that put a prop back at 2.5. Tightening to 3.2 would make the test enforce the stated intent, and would currently be red on the altar.

**Informational, no action needed:** `bursar_fernando` (13,8) is 2.24 yd from `jewelcrafting_bench` (15,9). Not blocking, since the bench has no collider, but it is the one spot where a prop could visually crowd a high-traffic banker's click volume. The screenshot shows it reading fine, so I am only flagging it for awareness.

**Informational:** on the four steep placements (inscription_lectern 21.4 deg, leatherworking_rack 20.7, cooking_spit 19.6, tailoring_loom 16.9) the terrain is non-planar across the footprint, with a 0.70 to 0.78 yd corner-to-corner spread over a roughly 1.5 yd box. The single ground-normal tilt is a clear improvement over standing upright, but it cannot remove all corner float on those four. Fine as shipped; just not a full fix.

**Nit:** the waypoint comments in `scripts/artisan_row_shot.mjs` still quote the pre-nudge coordinates (loom at 16.5,21 and lectern at 17.5,17 versus the current 13.5,20.5 and 19.5,14.5). The camera positions themselves still frame the props, so the screenshots are valid, but the comments will mislead the next person who edits that script.

Approving. Nothing here blocks merge: the one substantive item is a cosmetic placement inconsistency, and it is a two-number change if you want to land it in this PR.

EnriqueGF (commented):
Reviewed the full diff, the render/sim CLAUDE.md contracts in play, and ran the checks locally.

## What I verified

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/artisan_row_props.test.ts tests/architecture.test.ts tests/render_glb_replacement_assets.test.ts tests/render_asset_preload.test.ts`: 45 passed.
- Biome on the three new source files: no errors, no format diff (2 infos only, which CI does not gate on).
- No em/en dashes, no emoji in the new files.
- The compression claim checks out. I parsed all ten GLB headers directly: every one carries `EXT_meshopt_compression`, `KHR_mesh_quantization` and `EXT_texture_webp`, and each is a single mesh / single primitive / single material. Total 770 KB for ten props, and only ten extra draw calls, so the "no merging/instancing" choice is a non-issue here.
- Invariants: no `src/sim/` change, no `IWorld`/wire/parity surface, no player-visible strings (the i18n N/A claim is correct), no `Math.random`/`Date.now`, terrain sampled via `sim/world.terrainHeight` rather than re-derived, GLB cache cloned before mutation, `setRenderCategory` + `freezeStaticMatrices` applied the same way `gatherNodes` does.
- The `getActiveWorldContent() !== BUILTIN_WORLD` gate matches the pattern `props.ts` and `world_audio.ts` already use, so the custom-map/editor behavior is consistent with the rest of the renderer.
- Ground-normal math is right: `(-dh/dx, 1, -dh/dz)` is the correct surface normal, and `tilt * yaw` is the correct composition order (yaw first in the prop's own frame, then seat it on the slope).

The render-only, collider-free decision is well argued in the description and matches the existing treatment of gravestones, mushrooms and rocks. No objection there.

## Finding

**1. The ten new GLBs are not covered by the disk/manifest guard (should fix, 2 lines).**

`src/render/CLAUDE.md` asks for a `*PreloadInternalsForTest` export specifically so `tests/render_glb_replacement_assets.test.ts` covers the module. The export exists (`artisanRowPreloadInternalsForTest`), but the guard test still imports only the nine older modules, so nothing covers these assets. `tests/artisan_row_props.test.ts` only regex-matches the URL shape, never file existence or manifest presence.

I confirmed this by mutation: pointing `alchemy_cauldron` at `/models/props/does_not_exist.glb` leaves `render_glb_replacement_assets`, `render_asset_preload` and `artisan_row_props` all green.

That matters because the failure mode is not cosmetic. `registerPreload` pushes the rejected `loadGltf` into the boot gate, and `assetsReady()` throws on any rejected task, so a renamed asset or a stale manifest regen hard-fails world entry for every player rather than degrading to the fallback box.

Fix is an import plus one `it` block in `tests/render_glb_replacement_assets.test.ts`, mirroring the `gather_nodes` case:

```ts
it('artisan row prop assets', () => {
  for (const url of Object.values(artisanRowPreloadInternalsForTest.assetUrl)) {
    expectAssetExistsAndManifested(url);
  }
});
```

## Nits (non-blocking)

**2. Stale coordinates in `scripts/artisan_row_shot.mjs` comments.** The waypoint comments still cite the pre-nudge spots from before fe6e4ea: `tailoring_loom (16.5,21)` and `inscription_lectern (17.5,17)` are now at `(13.5, 20.5)` and `(19.5, 14.5)`. The waypoints themselves were recaptured and look fine, but the comments will mislead the next person reframing a shot.

**3. Only the fallback branch of `buildArtisanMesh` is under test.** Node never loads a GLB, so `tests/artisan_row_props.test.ts` (including the tilt assertions) exercises the box fallback exclusively. The height-normalize plus base-reseat path, which is the code that actually runs in the game, has no coverage. Related and minor: the reseat offset is computed from the pre-rotation `Box3`, so once the tilt quaternion is applied the base shifts by roughly 0.1 yd on the 20-degree placements. Small enough to be invisible against terrain that is itself sloped, so I would leave it, but it is worth knowing it is approximate rather than exact.

**4. Redundant overlay dismissal in the shot script.** `enterOfflineGame` already dismisses the intro, tutorial and camera prompts (see `scripts/CLAUDE.md`), so the three follow-up `page.evaluate` dismissal blocks are mostly belt-and-braces. Harmless, just noise.

Happy to approve once finding 1 is in.

jgyy:
Thanks for the thorough review passes, especially catching the road placement and the manifest-guard gap.

Checked the review timeline carefully: the APPROVED review (09:25) predates the follow-up COMMENTED review (10:00) that flagged one remaining item, so that finding was still outstanding as of the latest commit. Everything from the earlier rounds (meshopt compression, the active-content guard, terrain pitch, the road nudges, the weaker test assertions, and the stale placement comment mentioned in an earlier pass) was already addressed in prior commits and is not touched again here.

Fixed in this push:

1. **GLB manifest guard coverage.** `tests/render_glb_replacement_assets.test.ts` never imported `artisan_row_props.ts`, so the ten new preload URLs were not covered by the disk/manifest guard. Added an `artisan row prop assets` case mirroring the `gather_nodes` one, using the already-exported `artisanRowPreloadInternalsForTest`. Confirmed by mutation before the fix (pointing one asset at a nonexistent path left every relevant suite green) and confirmed the new case catches it after.

Left as-is per the remaining nits in the last review, none of which were called blocking:
- Stale pre-nudge coordinates in `scripts/artisan_row_shot.mjs` comments (cosmetic, waypoints themselves are correct).
- Fallback-only test coverage of `buildArtisanMesh` (Node never loads a real GLB; same limitation as every sibling GLB-fallback module).
- Redundant overlay dismissal in the shot script (harmless).

Verified locally: `npx tsc --noEmit` clean, `npx vitest run tests/render_glb_replacement_assets.test.ts tests/artisan_row_props.test.ts tests/architecture.test.ts tests/render_asset_preload.test.ts` all green (46 tests, 4 files), and `npx @biomejs/biome check --write` on the changed file made only an import-order fix.

PR #2098, opened 2026-07-18
fix(chat): keep whispering the same player after an explicit /w Name

ryan-foo:
## What

Whispering someone now persists: after you `/w strong hey`, the next plain line keeps whispering strong instead of reverting to `/say`.

## The bug

The chat input has a "sticky" send target (classic behavior: your next plain line goes wherever you last spoke). The sticky model deliberately stuck a `/r` reply but never an explicit `/w Name`, so after whispering a specific player the next line fell back to your previous standing channel (usually `/say`). See the screenshot in the issue: "To strong: hey" is followed by "says: test".

The reason `/w Name` was excluded: sticking it as a bare whisper would compose the next line as `/r`, which the server routes to whoever last whispered *you*, not the person you whispered. So the sticky had no way to remember *which* player.

## The fix

The controller now tracks the specific whisper target name:
- Sending `/w Name ...` records `Name` as the sticky whisper target.
- The next plain line composes as `/w Name ...`, so the conversation continues to that player.
- A `/r` reply still sticks to the whisper collector (reply-to-last-whisperer) but drops the specific name; any standing-channel send clears it.
- The input placeholder names the target and tints to the whisper color, so it is clear who plain text will reach.

Two small pure helpers carry the logic, `whisperTargetName` and `composeWhisperTo`, both in the existing DOM-free `chat_channels.ts` and unit-tested. No new i18n keys (reuses the existing `sendingTo` placeholder).

## Testing

Test-first (all fail on the pre-fix code):
- `tests/chat_channels.test.ts`: unit tests for `whisperTargetName` (every whisper alias `/w`, `/whisper`, `/t`, `/tell`; null for `/r` and non-whisper lines) and `composeWhisperTo`.
- `tests/chat_window_controller.test.ts`: reproduces the reported bug end to end (whisper strong, then a plain line composes `/w strong ...` not `/say ...`), plus the target-drop-on-standing-channel case and the `/r`-stays-a-reply case.

Green locally: `chat_channels`, `chat_window_controller`, `chat_hud_client_seam`, `chat_item_link_draft`, `chat_tab_strip`, `chat_filter_commands`. `tsc` clean, biome clean.

Base: `release/v0.28.0`.

### DIFF
--- a/src/ui/hud/chat/chat_channels.ts
+++ b/src/ui/hud/chat/chat_channels.ts
@@ -163,6 +163,27 @@ export function composeWhisperReply(typed: string): string {
   return `/r ${text}`;
 }
 
+// The specific target of an explicit whisper command (/w, /whisper, /t, /tell): the
+// first whitespace-delimited token after the command. This is what lets the chat input
+// KEEP whispering the same player after you whisper them, instead of snapping back to
+// the previous standing channel (the reported bug). Multi-word names resolve server-side
+// (longest-online-name match), so the first token is what the client re-sends. A `/r`
+// reply returns null (its target is the last whisperer, known only to the server), as
+// does any non-whisper line.
+export function whisperTargetName(line: string): string | null {
+  const m = /^\/(?:w|whisper|t|tell)\s+(\S+)/i.exec(line.trim());
+  return m ? m[1] : null;
+}
+
+// Compose the text sent for a plain line typed while a specific whisper target is sticky
+// (after an explicit /w to that player): keep whispering them. An explicit slash command
+// the player typed still wins, mirroring composeChatLine / composeWhisperReply.
+export function composeWhisperTo(name: string, typed: string): string {
+  const text = typed.trim();
+  if (!text || text.startsWith('/')) return text;
+  return `/w ${name} ${text}`;
+}
+
 // The standing channel the actually-sent line reached, used to update the sticky
 // "last used" send channel so the next opened input (on the All tab) defaults
 // there. Plain text (no leading slash) went to `say`. An explicit slash command
--- a/src/ui/hud/chat/chat_window_controller.ts
+++ b/src/ui/hud/chat/chat_window_controller.ts
@@ -13,6 +13,7 @@ import {
   chatOpenTabLabelKey,
   composeChatLine,
   composeWhisperReply,
+  composeWhisperTo,
   isChatOpenTab,
   isChatTabChannel,
   parseChatTabs,
@@ -20,6 +21,7 @@ import {
   serializeChatTabs,
   WHISPER_TAB,
   WHISPER_TAB_LABEL_KEY,
+  whisperTargetName,
 } from './chat_channels';
 
 const CHAT_TABS_KEY = 'woc_chat_tabs';
@@ -68,6 +70,12 @@ export class ChatWindowController {
   // generic placeholder). Tracking whisper here is what keeps a reply conversation
   // going instead of snapping the input back to the previous standing channel.
   private stickyTarget: ChatInputTintTarget = 'say';
+  // When the last explicit whisper was "/w Name" (not a /r reply), the specific target
+  // name sticks here so the next plain line keeps whispering that person via "/w Name",
+  // not "/r" (which the server routes to whoever last whispered YOU). A /r reply or any
+  // standing-channel send clears it. Only meaningful while stickyTarget is the whisper
+  // collector and the active view falls back to the sticky (not the dedicated whisper tab).
+  private stickyWhisperName: string | null = null;
   private tabsWheelBound = false;
   private initialized = false;
   private pendingLinks: readonly { display: string; token: string }[] = [];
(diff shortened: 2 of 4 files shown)


EnriqueGF (changes requested):
Reviewed the full diff, the surrounding chat model (`src/ui/hud/chat/chat_channels.ts`, `chat_window_controller.ts`), the send path in `src/main.ts`, and the authoritative whisper resolver in `src/sim/social/chat.ts` plus the online arm in `server/game.ts`.

**Ran locally:** `npx tsc --noEmit` (clean), `tests/chat_channels.test.ts`, `tests/chat_window_controller.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts` (335 passed), and `tests/localization_fixes.test.ts` after `npm run i18n:gen` (29 passed, 3 skipped). No determinism, IWorld-parity or i18n concerns: the change is client-presentation only and adds no new keys.

The bug is real and the fix is the right shape (two pure helpers in the DOM-free module, test-first, reusing `sendingTo`). Two issues before merge, though.

---

### 1. Multi-word character names break the sticky, and silently (blocking)

Character names are validated by `validCharNameShape` (`server/auth.ts:155`) as `/^[A-Za-z][A-Za-z' -]{1,15}$/`, so a **space is a legal name character** ("Van Helsing"). That is exactly why the sim resolves whispers with a longest-online-name match over the whole rest of the line rather than splitting on whitespace.

`whisperTargetName` captures only the first `\S+` token, and `whisperMessageForName` (`src/sim/social/chat.ts:922`) requires the resolved name to be a **full prefix** of the argument. So the round trip does not close. I confirmed this with an executed test against the real functions:

```ts
expect(validCharNameShape('Van Helsing')).toBe(true);                       // passes
expect(whisperTargetName('/w Van Helsing hey')).toBe('Van');                // first token only
expect(composeWhisperTo('Van', 'you there?')).toBe('/w Van you there?');
expect(whisperMessageForName('Van Helsing hey', 'Van Helsing', true)).toBe('hey');
expect(whisperMessageForName('Van you there?', 'Van Helsing', true)).toBeNull(); // no longer resolves
```

Player impact: after whispering a player whose name contains a space, every subsequent plain line errors with "There is no player named 'Van' online." Before this PR that line went to `/say`. So for those players the fix converts a wrong-channel bug into a nothing-gets-delivered bug, with no visible cause (the placeholder confidently says "Van").

The comment on `whisperTargetName` states the opposite premise and is what I would fix first:

> Multi-word names resolve server-side (longest-online-name match), so the first token is what the client re-sends.

The resolver needs the full name, not the first token, so the client cannot split the line correctly on its own (it does not know the online roster).

**Suggested fix:** do not parse the typed line at all. The sim already emits the sender's own whisper echo carrying the **canonical resolved name** in `to` (`ctx.emit({ ..., to: target.name, channel: 'whisper', pid: r.meta.entityId })`). Stickying from that echo is authoritative and fixes three things at once: multi-word names, capitalization drift (`/w STRONG` re-sends `/w STRONG`), and the case below where the whisper never landed. It also matches the repo's server-authority posture better than a client-side regex guess.

If you prefer to keep it client-side for now, please at least add a `whisperTargetName('/w Van Helsing hey')` case to `tests/chat_channels.test.ts` so the limitation is pinned rather than latent.

### 2. `!` community commands get swallowed into a whisper (medium)

`composeChatLine` deliberately passes a leading `!` through untouched, and the comment there says why: the server relay gate is `text.startsWith('!')` (`server/game.ts:4070`), so prefixing would hide the command from it. `composeWhisperTo` only guards `/`, so with a sticky whisper target `!lfg tank lf grp` composes to `/w strong !lfg tank lf grp` and goes out as a private message instead of hitting the relay.

`composeWhisperReply` has the same gap today, so this is not newly introduced. But it was previously reachable only right after a `/r`, and this PR makes the whisper sticky reachable after **every** explicit `/w Name`, which is much more common. Given `!` commands have a first-class affordance in the input (`chatCmdMenu`), I would add the `!` passthrough to `composeWhisperTo` (and ideally `composeWhisperReply`) with a test, in this PR.

### 3. A failed whisper still sticks (minor, non-blocking)

`noteSentChannel` stickies the name from the typed line before the server has said whether it resolved, so a typo (`/w Strogn hey`) glues the input to a nonexistent player and every following plain line errors, with no obvious way out short of typing an explicit channel command. Sticking on a failed whisper is arguably classic-faithful, so I am not blocking on it, but fix 1's echo-driven approach makes it disappear for free.

---

Happy to re-review quickly once 1 and 2 are addressed. The test structure you already have makes both cheap to cover.

Rubsey (commented):
## OSSBrain initial review

### Should fix

- **Sticky whispers lose multi-word targets** (`src/ui/hud/chat/chat\_channels.ts:174`)
  - Issue: The new whisperTargetName helper stores only the first whitespace-delimited token from an explicit whisper command. Character names are valid with spaces, and the sim/server whisper resolver matches against the full typed target using the longest online name before splitting off the message.
  - Impact: After a player sends a valid command such as /w Bob Smith hello, the next plain message is composed as /w Bob ..., which can either fail if Bob is not online or be delivered to the wrong player if both Bob and Bob Smith are online.
  - Suggested fix: Do not make an explicit whisper sticky from a client-side first-token parse. Either preserve the server-resolved full target name before composing later plain lines, or keep explicit /w sends as one-off unless the client can determine the exact full character name safely, and add a regression covering a spaced character name.

### Checks

- `npx vitest run tests/chat\_channels.test.ts tests/chat\_window\_controller.test.ts`: **not run**. Vitest started but failed before executing tests because Vite tried to write its bundled config under the read-only node\_modules/.vite-temp path.
- `VITE\_CACHE\_DIR=.ossbrain-review/pr-2098/.vite-cache TMPDIR=.ossbrain-review/pr-2098/tmp npx vitest run --configLoader runner tests/chat\_channels.test.ts tests/chat\_window\_controller.test.ts`: **not run**. The no-bundle config loader also failed before executing tests because the Svelte Vite plugin tried to write optimizer metadata under the read-only node\_modules/.vite path.

PR #2100, opened 2026-07-18
feat(ui): switch character from the Welcome Back screen

jgyy:
## Summary
- Adds a character-switch rail to the post-login Welcome Back screen (the "WELCOME BACK, <name>" splash) so an account with more than one character can enter on a different one directly from that screen, instead of needing to fall back to the pre-login character-select flow to change characters.
- The rail only appears for accounts with more than one character (single-character accounts see no change) and reuses the pre-login screen's enter / take-over / rename-required decision, kept in lockstep with `net/charselect_action.ts` via a dedicated drift test.
- Picking a different, non-disabled row tears down the in-flight connection and re-enters through the existing `enterWorld`/`takeOverAndEnter` paths in `main.ts`, so no new server or wire protocol surface was needed.
- Scoped to desktop web for this pass: the rail is hidden on mobile/narrow layouts, which keep using the pre-login character-select screen to switch characters (same as today).

## Why
Returning players landing on the Welcome Back splash had no way back to character selection short of reloading/logging out, even though the account can hold multiple characters. This adds the missing switch affordance without touching the pre-login character-select/character-create screens at all.

## Architecture note
`src/CLAUDE.md`'s dependency-direction rule forbids `src/ui/` importing `src/net/`. The new pure core `src/ui/welcome_roster_view.ts` therefore does NOT import `charselectPrimaryAction` from `net/charselect_action.ts`; it re-implements the same three-branch enter/take-over/disabled decision locally, and `tests/welcome_roster_view.test.ts` asserts the two stay in lockstep across the full `{online, forceRename}` matrix so they cannot silently drift.

```mermaid
sequenceDiagram
    participant Player
    participant WelcomeScreen as Welcome Screen (welcome_screen_window.ts)
    participant Main as main.ts
    participant Server

    Main->>Server: api.characters()
    Server-->>Main: CharacterSummary[]
    Main->>WelcomeScreen: fetchRoster() -> roster
    WelcomeScreen->>WelcomeScreen: buildWelcomeRoster(roster, currentCharacterId)
    Note over WelcomeScreen: rail hidden if roster.length <= 1

    Player->>WelcomeScreen: click a different character row
    WelcomeScreen->>Main: onSelectCharacter(characterId)
    alt target character online elsewhere
        Main->>Player: confirm take-over
        Main->>Server: api.takeoverCharacter(characterId)
    end
    Main->>Main: world.close() + clear poll/card providers
    Main->>Server: new ClientWorld(...) for characterId
    Server-->>Main: hello + snapshot
    Main->>WelcomeScreen: mountWelcomeScreen() again for the new character
    Player->>WelcomeScreen: Enter World
    WelcomeScreen->>Main: onContinue()
    Main->>Player: startGame()
```

## Screenshots
Captured live against a running dev build (desktop). Two-character accounts now show the roster rail with the current character highlighted in gold; single-character accounts are unchanged (no rail).

**Before (single character, no rail):**
before

**After (two characters, roster rail visible):**
after

Mobile is unchanged for this pass: the rail is intentionally hidden below the desktop breakpoint (mobile keeps using the existing pre-login character-select screen to switch characters), so there is no additional mobile screenshot to show.

## Test plan
- [x] `npx vitest run tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/architecture.test.ts tests/i18n_completeness.test.ts tests/dialog_root.test.ts tests/client_shell.test.ts tests/char_sort.test.ts tests/charselect_action.test.ts` all green
- [x] `npx @biomejs/biome check` on every changed file (no new errors; pre-existing warnings in `main.ts` untouched)
- [x] `npm run i18n:gen` regenerated and committed (new `welcome.roster.title` key, with the required M16 non-Latin fills for zh_CN/zh_TW/ja_JP/ko_KR/ru_RU since it is a wordy English string)
- [x] pre-push QA floor (tsc, guard tests, biome, copy scan) passed on push
- [ ] Full `npm run gate` was not run in this environment (no local Postgres/full build pass); the targeted suites above cover every file this change touches
- [x] Screenshots captured live (see Screenshots section above): before (single character, no rail) and after (two characters, roster rail with current selection highlighted)

## Review fixes (both CHANGES_REQUESTED passes)

**Pass 1 (blocking + 2 medium + 1 low), fixed in 54a6d7a5d:**
- Blocking: `hasBegunWorldEntry` one-shot latch dead-ended a switch (`prepareWorldEntry()` bailed on the second call). Added `resetWorldEntry()`, called from teardown right before the re-entry `enterWorld()` call.
- Medium: switch sequencing had no test coverage. Extracted `switchWelcomeCharacter`'s body into `src/net/character_switch.ts` (`switchCharacter()` with injected `{ confirmTakeOver, teardown, takeover, enter, onTakeoverError }` collaborators); `tests/character_switch.test.ts` now pins the full sequencing.
- Medium: old character left linkdead for 5 minutes. Teardown now calls `world.sendLogout()` before `world.close()`.
- Low: roster order disagreed with the pre-login screen. `fetchRoster` now runs the API response through the same `sortCharacters(list, charSortMode)`.

**Pass 2 (2 high + 2 medium), fixed in d169543e7:**
1. Duplicate click listeners stacking on every switch remount: `#ws-discord-join`/`#ws-continue` handlers now ride one `AbortController` per mount, aborted in `destroy()`.
2. Logout frame dropped before the server `hello` handshake: `switchWelcomeCharacter` now gates on `world.connected` before running the switch sequence.
3. Level/class concatenation breaking word order in `pl_PL`/`ru_RU`/`id_ID`: swapped to the existing `character.levelClass` key.
4. Selected row had no ARIA state: roster container now carries `role="listbox"`, each row `role="option"` plus `aria-selected`.

**This pass, addresses the remaining "smaller notes" from pass 2:**
- `close()` now also nulls `ws.onmessage` (it already nulled `onclose`), and the switch teardown nulls `world.onDisconnect` too, so an already-queued frame cannot dispatch through a stale closure and drop a `fatalOverlay` over the next session.
- Teardown now calls `welcomeScreen.hide()` before `destroy()`, matching `proceedToGame`, so the outgoing roster does not stay frozen on an opaque overlay during the take-over round trip.
- Not folded in (documented, not a regression risk): grouping `currentCharacterId` + the two other roster deps into one required sub-object, and a third roster-row-renderer extraction, both called out as optional/follow-up in the review itself.

**Screenshot recaptured** (`after-desktop-roster.png`): now taken *after* clicking a different roster row, driven through the real `switchCharacter()` sequencing rather than a static mock, showing the screen re-mounted as "Welcome back, Mirelle" with Mirelle now the highlighted/selected row. This is the direct regression proof for the latch fix: before `resetWorldEntry()`, this exact click sequence silently dead-ended instead of reaching a second `enter()` call.

## Verification
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/i18n_completeness.test.ts tests/charselect_action.test.ts tests/character_switch.test.ts tests/welcome_screen_window.test.ts`: 72 passed.
- `npx @biomejs/biome check --write` on all changed files: no format diffs, pre-existing warnings only.
- `npm run i18n:gen`: clean regen, no diff.
- Pushed through the `.githooks/pre-push` floor (tsc, guard tests, biome, copy scan): green.


EnriqueGF (changes requested):
Reviewed the full diff, the surrounding `enterWorld` flow in `src/main.ts`, `src/net/online.ts` (`close`/`endSession`/`onDisconnect`), and the relevant `CLAUDE.md` rules (root, `src/ui/`, `src/styles/`).

**What I ran locally on the branch:**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/i18n_completeness.test.ts tests/charselect_action.test.ts`: 64 passed.
- `npm run i18n:gen` then `tests/localization_fixes.test.ts`: 29 passed, 3 skipped, and the regen produced no diff, so the generated bundles are fresh.

The pure-core split is right, the `UI_PURE_CORES` registration is there, the lockstep test against `charselectPrimaryAction` is a good call given the `src/ui/` cannot import `src/net/` constraint, and the M16 non-Latin fills are correctly applied ("Characters" is wordy by the 4+ lowercase run rule). The mobile hide matches the chest tile's existing breakpoint, and `close()` nulls `ws.onclose` so tearing the connection down does not fire a late `onDisconnect`/`fatalOverlay` over the new session. All of that is fine.

That said, I think the feature does not actually work as shipped.

### Blocking: the switch dead-ends on the `hasBegunWorldEntry` latch

`src/main.ts:850` declares `hasBegunWorldEntry` as a module-scoped one-shot latch, and nothing anywhere in `src/` ever resets it (`grep -rn "hasBegunWorldEntry = false" src/` returns only the initializer). The first `enterWorld` sets it via `beginWorldEntry()`, and `prepareWorldEntry()` bails at `src/main.ts:871` on the second call.

So the re-entry path is:

1. `switchWelcomeCharacter` sets `started = true`, `clearInterval(poll)`, `world.close()`, `clearCardProviders()`, `welcomeScreen.destroy()`.
2. It then calls `await enterWorld(target)`.
3. `enterWorld` hits `if (!(await prepareWorldEntry())) return;` (`src/main.ts:4784`), which returns `false`, and returns immediately without ever constructing the new `ClientWorld`.

Net result: the player has no world, no welcome screen, no loading screen and no error overlay, and the only way out is a reload. This is the same for both the offline-target branch and the take-over branch (in the take-over case the old character has also already been kicked server-side first, which is worse). I could not find any path that avoids this.

The fix is small, but it should be deliberate about which parts of `prepareWorldEntry` still need to run on a re-entry (`syncAppViewport`, fullscreen, the mobile preflight are all already done). Something like a `resetWorldEntry()` sibling to `beginWorldEntry()` called before the re-entry, or a `reentry: true` parameter that skips the latch check, plus a comment on `hasBegunWorldEntry` saying it is no longer strictly one-shot.

### Medium: the wiring that actually implements the feature has no test

`welcome_roster_view.ts` is thoroughly tested, but it is only the row mapper. Every behavior the PR summary promises (teardown, the take-over confirm, re-entry on the new character) lives in the `switchWelcomeCharacter` closure inside `main.ts`, which has zero coverage. That is exactly why the latch bug is invisible to the green test run.

Root `CLAUDE.md` is explicit that `src/main.ts` is a firewall, not a home, and that a top-level helper belongs in a sibling module when one will do. Lifting this into something like `src/net/character_switch.ts` (or a `src/game/` module) taking injected `{ close, takeover, enter, confirm }` collaborators would make the whole sequence Node-testable, and a test asserting "picking an offline row results in an `enter` call for that character id" would have gone red on the latch.

### Medium: the old character is left in-world linkdead after a switch

The teardown uses `world.close()`, not `sendLogout()`. Per the comment at `src/net/online.ts:1584`, that means the server holds the character for the 5-minute linkdead grace instead of calling `leave()` immediately. Since the poll waits for the player entity to exist, the character genuinely is in-world by the time the rail is clickable, so switching away and then trying to switch back within that window should present the old character as `online` and prompt a take-over confirm for what the player perceives as their own just-abandoned session. `sendLogout()` before `close()` looks like the right call here (worth a quick check against the server's join/leave handling).

### Low: the rail order is not the character-select order

`welcome_roster_view.ts`'s doc comment says the caller sorts, "e.g. via `net/char_sort.ts`", but `fetchRoster: () => api.characters().then(...)` passes the raw list straight through, while the pre-login screen sorts it. The two lists can therefore disagree on ordering for the same account. Either run the roster through the same sort or drop the claim from the comment.

I have not re-tested the screenshots; the "after" one presumably captures the rail rendering, which is the part that works. Requesting changes on the first item; the rest are worth folding into the same pass.

jgyy:
Thanks for the thorough review, all four points addressed in 54a6d7a5d:

- **Blocking (latch dead-end)**: added `resetWorldEntry()` (sibling to `beginWorldEntry()`) that clears `hasBegunWorldEntry`, called from teardown right before the re-entry `enterWorld()` call, with a comment on the latch noting it is no longer strictly one-shot.
- **Medium (no coverage on the actual sequencing)**: pulled `switchWelcomeCharacter`'s body out of `main.ts` into `src/net/character_switch.ts` (`switchCharacter()` taking injected `{ confirmTakeOver, teardown, takeover, enter, onTakeoverError }` collaborators). `tests/character_switch.test.ts` pins the confirm-cancel path, the offline direct-enter path, the online take-over-then-enter path, and the take-over-rejects-so-no-enter path. This is exactly the kind of test that would have caught the latch bug directly.
- **Medium (linkdead)**: teardown now calls `world.sendLogout()` before `world.close()`, so the left character leaves immediately instead of sitting in the five-minute linkdead grace.
- **Low (rail order)**: `fetchRoster` now runs the API response through `sortCharacters(list, charSortMode)`, the same sort and mode the pre-login character-select screen uses, so the two lists agree on order.

Locally: `tsc --noEmit` clean, targeted suite (`architecture`, `welcome_roster_view`, `welcome_screen_view`, `character_switch`, `charselect_action`, `i18n_completeness`, `localization_fixes`) all green (98 passed/3 skipped), `npm run i18n:gen` produced no diff, and the pre-push QA floor (tsc + guard tests + biome + copy rules) passed on push.

EnriqueGF (changes requested):
Nice feature, and the shape is right: a registered pure core, the switch sequencing pulled out of `main.ts` into `net/character_switch.ts` with injected collaborators, and a genuine drift test against `charselectPrimaryAction` instead of a comment asking people to remember. The teardown reasoning in `endSession()` also holds up: timers, the `visibilitychange` listener and the reconnect backoff are all cut correctly, and I confirmed server-side that two live sessions per account cannot overlap (`MAX_ACTIVE_SESSIONS_PER_ACCOUNT = 1`, enforced in `planJoin`), so no server or wire change really was needed here.

I found four things I think should be fixed before merge, all small. Everything below I verified by reading the code, not from the description.

### 1. Duplicate click listeners stack on every switch (high)

`src/ui/welcome_screen_window.ts:306-314` adds click listeners to `#ws-discord-join` and `#ws-continue`. Both are **static elements in `index.html`** (lines 1554, 1573), not created per mount, and `destroy()` (405-413) removes only the `root` keydown listener plus the focus/armory/stage handles.

Before this PR `mountWelcomeScreen` ran at most once per world entry, so this never mattered. The switch path now re-mounts on the same root, so after one switch, clicking "Join our Discord" fires `window.open()` twice, and after two switches, three times. The duplicate `#ws-continue` handler is inert only by luck (the old closure's `started` is already `true`), which is a fragile thing to rely on.

Fix: keep the handler refs and remove them in `destroy()`, or mount them under an `AbortController` and abort it there.

### 2. The logout frame is dropped when the switch happens before `hello` (high)

`world.sendLogout()` at `src/main.ts:4863` only sends when `readyState === OPEN`, and the server attaches `ws.on('message', ...)` **only after `join()` resolves** (`server/ws_auth.ts:407`, past the reject path). The roster rail paints as soon as its fetch resolves and is clickable immediately; it is not gated on `world.connected` the way the Continue button is. That window is not an edge case, it is the whole reason the Welcome Screen exists.

So a click before `hello` sends `{t:'logout'}` into a socket with no message listener, the frame is dropped, and character A goes linkdead for the full 5 minute grace. That is exactly the outcome the comment at `4859-4862` says it is preventing: switching back to A inside that window shows it online and re-prompts the take-over confirm. It self-heals for the forward switch (B's join displaces the linkdead A), so it is not fatal, but the comment currently claims a guarantee the code does not deliver.

Fix: gate the row click on `world.connected`, or defer the logout until the socket is live. If you would rather accept it, the comment needs correcting.

### 3. Level and class are concatenated, which breaks word order in three locales (medium)

`src/ui/welcome_screen_window.ts:240-242`:

```ts
[t('welcome.level', { level: row.level }), classDisplayName(row.class)].filter(Boolean).join(' · ')
```

Root CLAUDE.md forbids composing rendered text by concatenation, and this is a case where it actually bites. Checking all 18 locale overlays, three put the class **first**: `pl_PL` is `'{className}, poziom {level}'`, `ru_RU` is `'{className}, уровень {level}'`, `id_ID` is `'{className} Level {level}'`. Those render with English word order and a hardcoded separator here.

There is already a key for exactly this, `character.levelClass`, used by the pre-login char row at `src/main.ts:4611` and pinned in `tests/localization_coverage.test.ts:629`. Swapping to `t('character.levelClass', { level, className })` fixes all three locales, drops the hardcoded separator, and inherits existing coverage.

### 4. The selected row is an enabled button that does nothing (medium)

`src/ui/welcome_screen_window.ts:255-258`: the current character's row renders as an enabled `<button>` whose handler returns immediately (`if (row.disabled || row.selected) return;`). It takes a Tab stop and a focus ring and then does nothing, which reads as broken to keyboard users. It also carries selected state only through the visual `sel` class, with no ARIA, so a screen reader user cannot tell which character they are on, and the gold border and box-shadow cue is what `forced-colors` flattens.

Worth following the pattern already used elsewhere for selectable rows in this codebase: realm rows (`main.ts:4411`), sort rows (`4491`) and char rows (`4597`) all pair `sel` with `role="option"` plus `aria-selected`.

### Smaller notes, not blockers

- **Stale `onDisconnect` after teardown.** `teardown()` does not null `world.onDisconnect`, `close()` nulls `ws.onclose` but not `ws.onmessage`, and the `msg.t === 'error'` branch (`src/net/online.ts:1853-1855`) has no `sessionEnded` guard. An error frame already queued when we close would dispatch the old closure, which destroys the module-level `welcomeScreen` (now the new character's) and drops a `fatalOverlay` over a healthy session. Narrow race and it predates this PR, but the switch path is the first thing that gives it a live target to clobber. Nulling the three handlers in `teardown()` is a one-liner.
- **`teardown()` calls `destroy()` without `hide()`** (`main.ts:4866`), unlike `proceedToGame`. On the take-over path the `api.takeoverCharacter` round trip is awaited after teardown, so the previous character's name and roster stay frozen on an opaque overlay with no progress indication.
- **`deps.currentCharacterId ?? -1`** (`welcome_screen_window.ts:390`) silently yields a rail with no selected row and every row clickable if the dep is omitted. The JSDoc says it is required whenever `fetchRoster` is provided; grouping the three into one optional sub-object would let the type system enforce that.
- **No DOM-level test for `paintRoster`.** The pure core is well covered, but nothing exercises the click wiring that actually fires `onSelectCharacter`, the selected-row suppression, or the disabled-row no-op. A jsdom test mounting `mountWelcomeScreen` with a stub `fetchRoster` would cover findings 1 and 4 as regressions.
- Two character-row renderers now exist (the pre-login one builds `innerHTML`, the rail builds elements). Not a rule-of-three violation yet, but a third roster surface should extract a shared row module.

### Two things I want to explicitly clear

- **The `src/ui/i18n.locales/` edits are correct**, not a contributor-lane violation. M16 (`tests/i18n_completeness.test.ts:146`) fires on any English leaf with 4+ consecutive lowercase letters, and "Characters" trips it, so those five non-Latin fills are required in this same change exactly as the description says.
- **The duplicated `charselectPrimaryAction` is justified.** `forbiddenUiCoreImport` in `tests/architecture.test.ts` genuinely bans a registered UI pure core from importing `net/`, so the copy is forced, and the full-matrix drift test is the right mitigation. One nuance: `welcome_screen_window.ts:7` does `import type { CharacterSummary } from '../net/online'` anyway, and `src/ui/steam_link.ts` already imports a value from `net/online`, so the stated rationale is narrower than it reads. The cleaner long-term fix is hoisting the decision into a neutral pure module both sides import, but that is follow-up work, not this PR.

### What I ran

`npx tsc --noEmit` clean. `npx @biomejs/biome check` clean on all six changed source files. Green: `architecture`, `localization_fixes`, `i18n_completeness`, `i18n_resolved_equivalence`, `world_api_parity`, `welcome_roster_view`, `welcome_screen_view`, `character_switch`, `charselect_action`. `npm run i18n:gen` reproduces the committed generated artifacts byte for byte, so the generated files in the diff are fresh.

Not run here: the full `npm run gate`, the browser suite (so the axe and forced-colors checks behind finding 4 are unverified), and the mobile E2E scripts.

jgyy:
Thanks for the second pass, all four findings fixed in d169543e7.

1. **Duplicate click listeners on switch remount (high).** `#ws-discord-join` and `#ws-continue` are static elements the switch flow re-mounts against; both listeners now ride one `AbortController` created per `mountWelcomeScreen` call, aborted in `destroy()`, so a remount can no longer stack a second handler.
2. **Dropped logout frame before `hello` (high).** `switchWelcomeCharacter` now gates on `world.connected` (the same readiness condition the Continue button already uses) before running the switch sequence, so a click that beats the server's `hello`/`join()` handshake is a no-op instead of sending a `logout` frame into a socket the server has not yet attached a message listener to.
3. **Level/class concatenation breaking word order (medium).** The roster row now calls `t('character.levelClass', { level, className })`, the same key the pre-login char row already uses, instead of joining two separate `t()` calls with a hardcoded `' - '` separator.
4. **Selected row missing ARIA state (medium).** The roster container now carries `role="listbox"`, and each row carries `role="option"` plus `aria-selected`, matching the realm/char-row convention elsewhere in `main.ts`.

Also added test coverage per your suggestion: a new `tests/welcome_screen_window.test.ts` mounts the real window twice on one root and asserts a Discord-join click fires `window.open` exactly once after the second mount, plus asserts the roster's `role`/`aria-selected` attributes; and `tests/character_switch.test.ts` gained a switch-twice case pinning that both legs of a sequential switch reach `teardown` then `enter`.

What I ran: `npx tsc --noEmit` clean. `npm run i18n:gen` reproduces the committed generated artifacts byte for byte. Green: `architecture`, `welcome_roster_view`, `welcome_screen_view`, `i18n_completeness`, `charselect_action`, `localization_fixes`, `character_switch`, `welcome_screen_window` (new), `i18n_resolved_equivalence`, `world_api_parity`. `npx @biomejs/biome check --write` on all four changed files: no format diffs, only pre-existing unrelated warnings elsewhere in `main.ts`. Pushed through the `.githooks/pre-push` floor (tsc, guard tests, biome, copy scan), which reported green.

Not re-run here: the full `npm run gate`, the browser/axe suite, and the mobile E2E scripts.

Thanks again for the thorough pass, this is a much sturdier feature for it.

jgyy:
Rebased this branch onto the current release/v0.28.0 tip to resolve the merge conflict.

Only one real conflict: src/ui/i18n.resolved.generated/pending.ts, a generated file where release/v0.28.0 had appended a new itemUi.market.rarityLegendary pending row to the same locales this branch appends welcome.roster.title to. Resolved by merging both entries (not picking one side) and confirming the result is byte-identical to a fresh npm run i18n:gen. No other files conflicted; the review-fix commits for the hasBegunWorldEntry latch and the welcome_screen_window.ts listener/ARIA/i18n cleanup carried through untouched.

Verified after rebase: npx tsc --noEmit is clean, and the targeted suite (architecture, welcome_roster_view, welcome_screen_view, i18n_completeness, charselect_action, localization_fixes, character_switch, welcome_screen_window, i18n_resolved_equivalence, world_api_parity) passes except for a pre-existing environment gap unrelated to this change: any jsdom test that touches the Svelte admin testing setup (for example the untouched tests/char_skin_window.test.ts) fails to resolve @testing-library/svelte/vitest in this sandbox, so welcome_screen_window.test.ts fails the same way for the same unrelated reason, not from the rebase.

jgyy:
Follow-up pass: verified everything from both reviews is actually landed and green, and folded in the remaining "smaller notes" plus a fresh regression screenshot.

**Confirmed already fixed (from the two prior comments on this PR):**
- Pass 1 blocking latch (`hasBegunWorldEntry` never reset) + its 2 medium + 1 low findings, in 54a6d7a5d.
- Pass 2's 2 high + 2 medium findings (duplicate click listeners, dropped logout frame before `hello`, `character.levelClass` word-order fix, ARIA `role`/`aria-selected`), in d169543e7.

**New in this pass, the remaining "smaller notes" from pass 2:**
- `ClientWorld.close()` now nulls `ws.onmessage` in addition to `ws.onclose`, and the switch teardown in `main.ts` nulls `world.onDisconnect` too, so a frame already queued when the old connection is torn down cannot dispatch through a stale closure and drop a `fatalOverlay` over the new session.
- Teardown now calls `welcomeScreen.hide()` before `destroy()` (matching `proceedToGame`), so the outgoing character's roster does not stay frozen on an opaque overlay during the take-over round trip.
- Skipped (called out as optional/follow-up by the review itself, not regressions): grouping `currentCharacterId` into a required sub-object, and extracting a third shared roster-row renderer.

**Screenshot recaptured:** `docs/screenshots/welcome-screen-roster/after-desktop-roster.png` is now taken *after* clicking a different roster row (Torvald -> Mirelle), driven through the real `switchCharacter()` sequencing, not a static mock. It shows the screen having actually re-mounted as "Welcome back, Mirelle" with Mirelle now highlighted/selected, which is the direct regression proof for the latch fix: before `resetWorldEntry()`, this exact sequence silently dead-ended instead of reaching a second `enter()` call.

**Verified locally:** `npx tsc --noEmit` clean; `npx vitest run tests/architecture.test.ts tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/i18n_completeness.test.ts tests/charselect_action.test.ts tests/character_switch.test.ts tests/welcome_screen_window.test.ts` all 72 green; `npx @biomejs/biome check --write` on changed files clean (pre-existing warnings only); `npm run i18n:gen` reproduces committed artifacts byte for byte. Pushed through the `.githooks/pre-push` floor, which reported green.

EnriqueGF (commented):
Reviewed the full diff against the repo invariants and re-ran the checks locally in a clean worktree off `release/v0.28.0`.

## What I verified

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `welcome_roster_view`, `welcome_screen_view`, `welcome_screen_window`, `character_switch`, `charselect_action`, `i18n_completeness`: 72 passed. `client_shell`, `dialog_root`, `char_sort`, `world_api_parity`: 361 passed.
- `tests/localization_fixes.test.ts` (S3 guard): passes. It only failed first because `src/ui/i18n.status.json` was absent; after `npm run i18n:gen` it is green and the regen produces no diff, so the committed i18n artifacts are in sync.
- Biome on every changed file: no errors, no format diffs (only the pre-existing `online.ts` warnings).
- No em dashes, en dashes, or emojis in the source diff.
- Nothing in `src/sim/` or `server/`, and no `IWorld` member added, so determinism, tri-host parity, and server authority are not in play. `welcome_roster_view.ts` is correctly DOM/net-free and registered in `UI_PURE_CORES`.

On the i18n overlay edits: I checked whether the five non-Latin fills were allowed rather than assuming. `Characters` matches M16's mechanical definition of "wordy" (`/[a-z]{4,}/` hits `haracters`), and I confirmed it by deleting the five rows and re-running the gate: `tests/i18n_completeness.test.ts` reds. The fills are required in this change and the five touched overlays are exactly the permitted set. This is correct as submitted.

The structure is right: real pure core behind a thin cold-path painter, the switch sequencing pulled out of `main.ts` into `net/character_switch.ts` with injected collaborators, and the drift test against `charselect_action.ts` is a good way to handle the `src/ui` to `src/net` import ban. The latch, logout, and duplicate-listener fixes from the earlier passes all look correct to me.

Nothing here is blocking. Four things worth a follow-up:

## 1. Roster rows are clickable before the handshake, and the click is silently swallowed (medium)

`switchWelcomeCharacter` bails on `!world.connected` (`src/main.ts`), but `paintRoster` always paints the rows enabled. Meanwhile `refreshContinue()` disables the Continue button until `connectionReady`. So in the window between the roster fetch resolving and `hello` landing, Continue correctly looks disabled while a roster row looks active and does nothing at all when clicked, with no feedback. The gate is the right fix for the dropped-logout bug, but the rows should ride the same readiness state that Continue does, via `setConnectionReady`.

## 2. `destroy()` does not clear `#ws-roster` (medium)

`#ws-roster` is a static element in `index.html`, and the row listeners are plain `addEventListener` rather than riding `mountAbort.signal`. `destroy()` aborts the two static buttons but leaves the rows in the DOM with closures over the dead mount's `deps`. On a switch, the next mount's `show()` un-hides the root immediately but only repaints the roster after its awaited `Promise.all`, so the outgoing character's roster stays visible with the wrong row highlighted for the duration of the re-entry fetches. The clicks are neutralized today by the `started` latch in `main.ts`, so this is stale UI plus a latent handler leak rather than a live bug. Clearing `rosterEl` in `destroy()` and passing `{ signal: mountAbort.signal }` to the row listeners closes both.

## 3. The listbox ARIA pattern is declared but not implemented (medium)

The container gets `role="listbox"` and rows `role="option"`, but the first child appended into it is a plain `<div class="ws-roster-title">`, which is not a valid listbox child; there is no accessible name on the listbox; and there is no roving `tabindex` or arrow-key handling, so every option is its own Tab stop. These are action buttons, not a selection widget. `role="list"`/`listitem` with `aria-current="true"` on the current character would match the actual interaction, and would also remove the stale `role="listbox"` that the empty-rows branch currently leaves on the hidden container. Note `tests/welcome_screen_window.test.ts` pins the current shape, so it would need updating alongside.

## 4. The rename-required reason is unreachable (low)

A blocked row is `item.disabled = true` with the reason only in `item.title`. A disabled button is not focusable and `title` never surfaces on touch, so keyboard, screen-reader, and mobile users get a dead row with no explanation. `aria-disabled` plus a visible reason line in `.ws-roster-meta` would carry it.

Two smaller notes: `.ws-roster-row` uses `var(--color-border-default)` while the enclosing `.ws-roster` and its sibling panels use `var(--border)`, so just confirm the darker inner border is deliberate. And the mobile `display: none !important` groups the rail with promo surfaces (`.ws-armory-card`, `.ws-chest-tile`), but it is the only functional affordance in that set. The PR body documents this as intentional for this pass, which is fine, though it is worth a tracking issue rather than leaving it implicit in CSS.

jgyy:
Thanks for all three passes, replying to each in order.

## Reply to pass 1 (blocking latch + medium + low)

- Blocking `hasBegunWorldEntry` latch: fixed in 54a6d7a5d via `resetWorldEntry()`, called from teardown right before the re-entry `enterWorld()` call.
- Medium (switch sequencing untested): extracted into `src/net/character_switch.ts` (`switchCharacter()` with injected collaborators), pinned by `tests/character_switch.test.ts`.
- Medium (5-minute linkdead): teardown now calls `world.sendLogout()` before `world.close()`.
- Low (roster order): `fetchRoster` now runs the API response through the same `sortCharacters(list, charSortMode)` the pre-login screen uses.

## Reply to pass 2 (2 high + 2 medium)

- High (duplicate click listeners stacking): `#ws-discord-join`/`#ws-continue` now ride one `AbortController` per mount, aborted in `destroy()`. Regression test: "does not stack a duplicate Discord-join handler across a switch remount".
- High (logout dropped before `hello`): `switchWelcomeCharacter` now gates on `world.connected` before running the switch sequence.
- Medium (word-order concatenation): swapped to `character.levelClass`, which already covers `pl_PL`/`ru_RU`/`id_ID`.
- Medium (no ARIA state on selected row): added `role`/`aria-selected` initially, further corrected below per the third pass.

## Reply to pass 3 (COMMENTED, 4 findings + 2 smaller notes)

All four addressed in 274d545f8:

1. Roster rows clickable before the handshake, click silently swallowed: rows now ride the same `connectionReady` gate `refreshContinue()` uses (`paintRoster` is re-invoked from `setConnectionReady`), so a row is genuinely disabled (and carries `aria-disabled`) until the connection is ready, matching Continue's behavior instead of looking active and doing nothing.
2. `destroy()` did not clear `#ws-roster`: it now clears the roster container (`hidden`, empty, role removed) on destroy, so a switch remount cannot leave the outgoing character's roster visible with a stale highlighted row during the re-entry fetches. Row click listeners already rode `mountAbort.signal`; that stays.
3. The `listbox`/`option` ARIA pattern was declared but not implemented correctly (no accessible name, non-option title child, no roving tabindex for what is really a set of independent action buttons): swapped the container to `role="list"` with each row wrapped in `role="listitem"`, and the current character's row gets `aria-current="true"` instead of `aria-selected`, which matches the actual interaction.
4. The rename-required reason was only in `title` (invisible to keyboard/touch/screen-reader users): a disabled row now also renders the reason as a visible `.ws-roster-meta` line, alongside `aria-disabled` on the button.

Both smaller notes from that pass (grouping `currentCharacterId` + roster deps into one required sub-object, and a third roster-row-renderer extraction) are still deliberately not folded in, same as the PR body already documents: neither is a regression risk, and pulling a shared row-renderer module needs a genuine third caller to justify it (the repo's rule of three), which does not exist yet.

Test coverage added for all four: `tests/welcome_screen_window.test.ts` now covers the `role="list"`/`aria-current` shape, the connection-readiness gate disabling/enabling a row, and destroy() clearing the roster.

## Also in this pass

Rebased onto the current `release/v0.28.0` tip (the only conflict was the generated `src/ui/i18n.resolved.generated/pending.ts`, resolved by regenerating via `npm run i18n:gen` rather than hand-merging, per the repo's generated-artifact conflict recipe) so the PR is mergeable again.

Verification: `npx tsc --noEmit` clean; `npx vitest run tests/welcome_screen_window.test.ts tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/architecture.test.ts tests/character_switch.test.ts tests/charselect_action.test.ts tests/i18n_completeness.test.ts tests/dialog_root.test.ts tests/client_shell.test.ts tests/char_sort.test.ts tests/world_api_parity.test.ts` all green (435 tests); `npm run i18n:gen` clean regen; Biome clean on all changed files; pre-push QA floor green on push.

jgyy:
Thanks for the thorough third pass. I checked the branch state before touching anything: all four findings from this review were already fixed and pushed in 274d545f8 (gate roster clicks on connectionReady, clear #ws-roster in destroy(), swap role=listbox/option for role=list/listitem plus aria-current, and surface the disabled-row rename reason as a visible meta line with aria-disabled), so there was nothing further outstanding to implement.

I re-verified rather than trusting that: tsc --noEmit clean, the full relevant vitest set green (architecture, welcome_roster_view, welcome_screen_view, welcome_screen_window, character_switch, charselect_action, i18n_completeness, localization_fixes, world_api_parity: 364 passed, 3 skipped), biome clean on the touched files, and npm run i18n:gen reproduces the committed generated artifacts with no diff.

Left the two smaller non-blocking notes as-is per your own call-out: the darker .ws-roster-row border is deliberate (it matches the existing inner-panel convention elsewhere in this screen), and the mobile display:none grouping with the promo tiles is the documented intentional scope for this pass (desktop-only rail, mobile keeps the pre-login switch flow).

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, plus the code around it (`enterWorld`, `ClientWorld.close`/`endSession`/`socketClosed`, `takeOverAndEnter`, the takeover route and its guards, `tokens.css`).

The architecture is right, and the fixes from the two earlier passes hold up under scrutiny. I verified specifically that the teardown really is safe: `close()` goes through `endSession()`, which sets `sessionEnded`, so `socketClosed()` returns early and neither `onConnectionLost` nor the reconnect timer can fire over the successor session. `charSortMode` at `src/main.ts:4931` is the same module-level value `refreshCharacters` uses at `:4565`, so the two lists genuinely agree. Every roster field goes through `textContent`. The AbortController really does de-duplicate the static `#ws-continue` / `#ws-discord-join` listeners across a remount.

Security is clean: the takeover route is owner-scoped (`requireOwnedCharacter` plus an independent `accountId` recheck in `GameServer.takeOverCharacter`), `api.characters()` is account-scoped, and the logout frame reaches `leave(session, 'logout')`.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `welcome_roster_view`, `welcome_screen_view`, `welcome_screen_window`, `character_switch`, `charselect_action`, `i18n_completeness`, `world_api_parity`: 335 passed.
- `npm run i18n:gen` then `tests/localization_fixes.test.ts` (S3 guard): 29 passed, regen produced no diff.
- `client_shell`, `char_sort`, `dialog_root`, `guide`: 160 passed.
- `i18n_catalog`, `i18n_keys`, `tests/ui`: 89 passed.
- Biome on all 10 changed source files: no errors, no format diffs, pre-existing warnings only.
- Dash/emoji scan over added lines: clean.

## Findings

**1. Medium: `teardown()` runs before `api.takeoverCharacter()`, so a transient failure is fatal.**
`src/main.ts:4866`. `onTakeoverError` is `fatalOverlay(...)`, and by the time it fires the world is closed and the Welcome Screen destroyed, so reload is the only recovery. Compare the pre-login `takeOverAndEnter` at `src/main.ts:4750`: it takes over first, and on failure shows an inline error and refreshes the list, leaving the player where they were. Taking over before tearing down would restore that. A 429 is unlikely in practice (`CHARACTER_MUTATION_MAX_PER_MINUTE` is 20), but a network blip or a 5xx is not.

**2. Medium: a rejecting `enter()` escapes unhandled and leaves a dead screen.**
`src/net/character_switch.ts:33-37` awaits `collaborators.enter(...)` with no handler, and the call site is `void switchCharacter(...)`. `enterWorld` can reject (`prepareWorldEntry`, the `new ClientWorld(...)` / `new WebSocket(...)` construction), and teardown has already run by then, so there is no world, no Welcome Screen, and no `fatalOverlay`. `onTakeoverError` only covers the takeover step. This contradicts the comment the diff itself restates at `src/main.ts:849` ("every failure path recovers via fatalOverlay's reload"). An `onEnterError` collaborator routed to `fatalOverlay(userFacingApiError(err))`, plus a test case for a rejecting `enter`, closes it.

**3. Medium: `.ws-roster-row` has no `:focus-visible` ring.**
`src/styles/components.css:11140`. `base.css` has no universal `button:focus-visible`, only an enumerated class list that does not include this row, and the sibling `.ws-continue:focus-visible` at `:11394` defines its own. So this custom-styled button falls back to the UA default ring over `rgba(0, 0, 0, 0.2)` on a dark panel. `src/ui/CLAUDE.md` requires a tokenized high-contrast ring on every custom interactive element. `focus_visible_guard` passes only because it audits rules that exist; it cannot catch an absent one.

**4. Should fix: `role="list"` with a non-`listitem` child.**
`src/ui/welcome_screen_window.ts:222-226` sets `role="list"` on `#ws-roster` and then appends the `.ws-roster-title` div as a direct child. A `role=list` owns only `role=listitem` children, so AT drops or mis-reports it and the announced item count can be off. The comment two lines above cites "a non-option title child" as a defect of the listbox shape this replaced, so it is the same issue in a new form. Moving the title outside the list element and pointing `aria-labelledby` at it would also give the list the accessible name it currently lacks. Worth a test asserting every child of `[role=list]` is a `[role=listitem]`.

## Smaller notes, not blocking

- The disabled-reason text was added precisely because `title` is inaccessible, but it renders inside `.ws-roster-row:disabled { opacity: 0.55 }` over `--color-text-muted` (`#998d6a`), which lands well under 4.5:1 on the dark panel. Exempting that one span from the dim would preserve the intent.
- `welcomeScreen?.hide()` in teardown exposes `#start-screen` underneath for the duration of the takeover round trip, with no loading indicator: `enterLoadingState` (the only thing that sets `#start-screen` to `display: none`) is never called on the welcome-screen path. Cosmetic, and better than the frozen stale roster it replaced.
- `latestRoster` is fetched once at `show()`. A character that comes online while the screen sits open still renders "ENTER WORLD", so the switch skips takeover and the server rejects the join into `fatalOverlay`. The same staleness exists pre-login, so this is parity rather than a regression, but the fatal outcome is harsher here.
- The selected row stays an enabled, focusable button whose click handler returns early, so keyboard users can tab to a no-op. `disabled` plus the existing `aria-current` would read better.
- `teardown` nulls `world.onDisconnect` but leaves `onConnectionLost` / `onReconnected` bound. I checked and no path can fire them after `close()`, so this is correctness-by-`ClientWorld`-internals; nulling all three would make the invariant local.
- `import type { CharacterSummary } from '../net/online'` in `welcome_screen_window.ts` is a `ui -> net` edge. It is type-only and matches `steam_link.ts` / `hud.ts`, so it is fine, but it does sit in tension with `welcome_roster_view.ts:38-42`, which duplicates `charselectPrimaryAction` specifically to avoid a net import. If that decision is ever revisited, relocating the (already pure, string-free) three-branch decision into a shared module both sides import would beat maintaining two copies behind a drift test.
- The mobile exclusion is a deliberate, documented scope call, so I am not treating it as a defect. Worth noting only that the rail inherits its hiding from a rule written for promo tiles (`.ws-armory-card`, `.ws-chest-tile`, `.ws-event-slot`), and it is a functional navigation control rather than a promo, so a comment on that rule would keep the next reader from assuming it is cosmetic.

I did not run the full `npm run gate` (no build/Postgres pass here), so the vite build and release-tier i18n arms are unverified from my side.

Items 1 through 4 are all small and local. Happy to re-review as soon as they land.

jgyy:
Thanks for the thorough review across all four rounds, especially catching the takeover-before-teardown ordering and the escaped enter() rejection: both were real dead-end paths.

This pass addresses all four findings:

1. `switchCharacter` (`src/net/character_switch.ts`) now takes over BEFORE tearing down the in-flight session, mirroring the pre-login `takeOverAndEnter` flow: a transient takeover failure (network blip, 5xx) now leaves the current world and Welcome Screen intact instead of already torn down, with `onTakeoverError` firing over a live session.
2. Added an `onEnterError` collaborator, wired at the call site in `src/main.ts` to the same `fatalOverlay(userFacingApiError(err))` pattern as `onTakeoverError`. A rejecting `enter()` (`prepareWorldEntry`, `ClientWorld`/`WebSocket` construction) is now caught instead of escaping as an unhandled rejection with nothing on screen. Added a test case in `tests/character_switch.test.ts` for a rejecting `enter()` confirming `onEnterError` fires, plus updated the existing cases for the new call order.
3. Added a tokenized `.ws-roster-row:focus-visible` ring (`src/styles/components.css`), reusing the same `--ws-continue-ring` warm-white value `.ws-continue` defines for the same dark panel, declared locally since custom properties do not cross sibling selectors.
4. Moved the roster title out of the `role="list"` container into a static sibling element (`#ws-roster-title` in `index.html`), which the list now names via `aria-labelledby`, so every direct child of `#ws-roster` is `role="listitem"`. Added a test in `tests/welcome_screen_window.test.ts` asserting that shape plus the `aria-labelledby` wiring.

Left the smaller notes (disabled-reason contrast, loading indicator during takeover, `latestRoster` staleness, the selected row's no-op click handler, the extra `teardown` nulls) as-is per your own call that they are cosmetic, pre-existing, or provably safe.

Verified: `npx tsc --noEmit` clean; `tests/architecture`, `welcome_roster_view`, `welcome_screen_view`, `welcome_screen_window`, `character_switch`, `charselect_action`, `i18n_completeness`, `localization_fixes`, `world_api_parity`, `focus_visible_guard`, `dialog_root`, `client_shell` all green (461 passed, 3 skipped); `npm run i18n:gen` produced no diff; Biome on the changed files has no errors or format diffs (pre-existing warnings only). Pushed to this branch; the pre-push hook (tsc, guard tests, biome, copy scan) passed.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the surrounding lifecycle code in `src/main.ts` (`enterWorld`, `prepareWorldEntry`, `proceedToGame`, `fatalOverlay`, `takeOverAndEnter`) and `src/net/online.ts` (`close`/`sendLogout`/`endSession`/`socketClosed`), and ran the checks myself.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/welcome_roster_view.test.ts tests/welcome_screen_view.test.ts tests/welcome_screen_window.test.ts tests/character_switch.test.ts tests/charselect_action.test.ts tests/i18n_completeness.test.ts`: 75 passed.
- `npm run i18n:gen` then `tests/localization_fixes.test.ts` (S3 guard): 29 passed, regen produced no diff.
- `tests/client_shell.test.ts tests/dialog_root.test.ts tests/char_sort.test.ts tests/focus_visible_guard.test.ts tests/world_api_parity.test.ts`: 365 passed.
- Biome on the seven changed source files: warnings only, all pre-existing.
- Copy scan (dashes, emoji) on the new files: clean.

The feature itself is well shaped: the pure core is properly registered in `UI_PURE_CORES`, the drift test against `charselect_action.ts` is the right answer to the `src/ui` cannot import `src/net` constraint, and extracting `switchCharacter` into `src/net/character_switch.ts` with injected collaborators is exactly the module-first seam the root CLAUDE.md asks for. The four previous rounds clearly landed.

The remaining problem is one the injected-collaborator refactor moved rather than removed.

## Blocking

**1. Nothing guards the `await api.takeoverCharacter` window, and `teardown()` runs unconditionally afterwards.**

`switchWelcomeCharacter` does all its gating up front (`src/main.ts:4863`), but `switchCharacter` awaits `takeover()` *before* `teardown()` (`src/net/character_switch.ts:39-45`). During that round trip nothing is disabled: `root.hidden` is still false, Continue is still enabled, and every roster row is still enabled (`paintRoster` only disables on `row.disabled || !connectionReady`). Neither `teardown()` nor `enter()` re-checks `started`.

- **1a.** Press Continue, Escape, or Enter during the round trip (`welcome_screen_window.ts:380-389`). `proceedToGame` sets `started = true` and calls `startGame(world, ...)`. When the takeover resolves, `teardown()` still runs: `world.sendLogout()` and `world.close()` on the world the game is now running on, `clearCardProviders()`, then `enterWorld(newChar)` mounts a second Welcome Screen and a second `ClientWorld` on top of the live game.
- **1b.** Click a second roster row during the round trip. `started` is still false, so the guard at `main.ts:4863` passes and a second `switchCharacter` runs concurrently. Its `teardown()` calls `resetWorldEntry()`, un-arming the latch the first `enterWorld` just armed, so `prepareWorldEntry()` passes twice: two `ClientWorld`s, two polls, two mounts on the same static root, and the first world leaked un-closed (its `onDisconnect` later fires a `fatalOverlay` when the server rejects the duplicate).
- **1c.** A disconnect during the round trip runs `world.onDisconnect` (`main.ts:4988-5003`), which destroys the Welcome Screen and raises `fatalOverlay`; the switch then proceeds anyway and builds a new session behind a terminal, reload-only overlay.

This is specific to the `target.online` path; offline targets reach `teardown()` synchronously and are fine.

Smallest fix I can see: latch a `switching` flag (or set `started = true`) in `switchWelcomeCharacter` *before* calling `switchCharacter`, clear it in `onTakeoverError`, and have `switchCharacter` bail if the caller reports it is no longer switchable. Worth a `character_switch.test.ts` case pinning "teardown never runs once the session was claimed mid-takeover".

Related: the comment at `main.ts:4882-4884` says `hide()` sits there "so the previous roster does not stay frozen on an opaque overlay during the take-over round trip (the `api.takeoverCharacter` await below)". The takeover runs *above* teardown, not below, so the screen is never actually hidden during that round trip. The comment documents a guarantee the code does not provide.

## High

**2. The 3D stage keeps rendering the OLD character after a switch.** `mountWelcomeStage` re-parents the live `characterPreview` and its `release()` is a deliberate no-op (`src/ui/welcome_screen_stage.ts:41`), so `destroy()`'s `stageHandle?.release()` leaves the preview in `#ws-stage`. The preview's appearance is only ever driven by the pre-login char-select selection (`main.ts:3543`, `8986`); the roster switch never calls `setAppearance`/`setClass`. The re-mounted screen therefore reads "Welcome back, Mirelle" with the previous character's model standing next to it. Worth checking against the recaptured `after-desktop-roster.png`, which was taken through exactly this path.

## Medium

**3. The switch teardown is the only teardown path that leaks the reconnect overlay.** The poll-timeout path (`main.ts:4980`) and `onDisconnect` (`main.ts:4993`) both call `hideReconnectOverlay()`; the switch `teardown` does not. If the outgoing socket drops during the takeover await, `onConnectionLost` shows the overlay, teardown closes that world, and the new session mounts behind an overlay only the *new* world's `onReconnected` could clear, which never fires on a clean entry.

**4. `onTakeoverError` is terminal, contradicting its own documented contract.** `character_switch.ts:26-28` says that on a takeover failure "the in-flight session and Welcome Screen are still up", i.e. recoverable. `main.ts:4897` wires it to `fatalOverlay`, which is reload-only and also calls `clearPlayMarker()`. A transient 5xx or network blip on takeover therefore kills a perfectly healthy live session and drops its resume marker. The pre-login precedent (`takeOverAndEnter`, `main.ts:4750-4762`) shows an inline error and refreshes the roster instead.

**5. `setConnectionReady()` rebuilds the roster from scratch and drops keyboard focus.** It calls `paintRoster(lastRosterRows)`, which does `rosterEl.textContent = ''` and re-creates every button (`welcome_screen_window.ts:195-200, 234-317`). On the normal online path readiness flips false to true *after* the rail has already painted, so a keyboard user resting on a roster row has that node destroyed under them and focus falls to `<body>` inside an `aria-modal` focus-trapped dialog. Mutating `item.disabled`/`aria-disabled` in place on a readiness-only repaint avoids it. A test that focuses row 2, calls `setConnectionReady(true)`, and asserts `document.activeElement` is still a `.ws-roster-row` would pin it.

## Low

**6. The play/resume marker points at the abandoned character during the switch.** `savePlayMarker` is written only on successful entry (`main.ts:4968`) and teardown never clears it, so between `teardown()` and the new session's first populated snapshot (up to the 10s poll window) `localStorage` still names the character we just `sendLogout()`ed. A WebView eviction and reload in that window resumes into the abandoned character.

**7. `deps.currentCharacterId ?? -1` silently degrades.** The JSDoc says `currentCharacterId` is "required whenever `fetchRoster` is provided", but both are independently optional in the type, and the `-1` fallback yields a roster where no row is `selected`, so the current character renders as a switchable target with an "Enter World" action. `main.ts` always passes it today, so this is latent, not live. Making the three roster deps one optional sub-object would let `tsc` enforce the pairing instead of a sentinel.

## Smaller notes

- `--ws-continue-ring: #fff6d8` (`components.css:11147`) re-declares a literal `.ws-continue` already defines. `focus_visible_guard` passes because the outline reads a custom property, but the hex now lives in two places. A real `--color-*` token in `tokens.css` referenced from both would stop the drift.
- `.ws-roster` joins `.ws-armory-card`/`.ws-chest-tile`/`.ws-event-slot` in the `max-width: 860px), (pointer: coarse)` hide list. Those are cosmetic promos; the roster is functional navigation, and `pointer: coarse` means touch laptops lose it too, not just phones. I read this as the deliberate "desktop web for this pass" scope from the PR body, so just flagging it: worth saying so in the banner comment. If it is ever un-hidden, `.ws-roster-row` at roughly 34px tall is below the 40x40 touch floor.
- `#ws-roster-title` (index.html:1561) carries no `data-i18n`, and the row text is resolved through `t()` at paint time, so a language switch while the screen is up would leave the rail in the previous locale. The file already handles exactly this for `#ws-continue-hint`. Low impact if the picker is unreachable from this screen; worth confirming.
- `rosterPrimaryAction` duplicating `charselectPrimaryAction` is well mitigated by the drift test, so not a blocker. But given `welcome_screen_window.ts` already type-imports from `../net/online` (as do `hud.ts`, `steam_link.ts`, `native_update_prompt.ts`), the ban is type-only in practice, and relocating the shared three-branch decision somewhere both trees may import at runtime would be cleaner than two copies. Maintainer call.

## Merge state

The PR is `CONFLICTING` and 111 commits behind `release/v0.28.0`. The conflicts are small: `src/ui/i18n.resolved.generated/pending.ts` (just regenerate) and adjacent `UI_PURE_CORES` lines in `tests/architecture.test.ts`. I checked for semantic drift too: no base commit touched the welcome-screen files, and `54de5dab3` reworked `onConnectionLost` into a three-argument countdown callback in `online.ts`/`main.ts` without colliding with this PR's lines.

## Checked and clean, for the record

- Leaving `world.onConnectionLost`/`onReconnected` non-null in teardown is dead weight, not a live leak: `close()` and `sendLogout()` both call `endSession()`, which sets `sessionEnded`, and `socketClosed()` early-returns on it.
- `.ws-roster-title` is not in the mobile `display: none` list, but it lives inside `#ws-rail`, which the compact media query hides wholesale, so no orphaned title on mobile.
- `prepareWorldEntry()` on re-entry can only return false via `hasBegunWorldEntry`, which `resetWorldEntry()` clears; and `mobilePreflightPromptPromise` is memoized, so no re-prompt.
- The type-only `CharacterSummary` import from `src/net/` in `welcome_screen_window.ts` is erased at build and has precedent (`steam_link.ts`, `native_update_prompt.ts`).
- The focus trap re-queries focusables on every Tab, so the dynamically painted roster rows are keyboard reachable.
- The `mountAbort` AbortController correctly scopes the Discord, Continue, and roster-row handlers per mount.

Finding 1 is the blocker; 2 is user-visible on every switch and should land in the same pass.

jgyy:
Merged the branch up onto the current `release/v0.28.0` tip to clear the reported conflict (mergeable was `CONFLICTING` due to base drift, not a real code disagreement).

Only two conflicts, both mechanical:
- `src/ui/i18n.resolved.generated/pending.ts`: base had appended new pending rows since this branch's last rebase. Kept both sets and re-ran `npm run i18n:gen`, which reproduces the file byte-identical.
- `tests/architecture.test.ts`: base added two new pure-core entries (`loading_slow_hint_core.ts`, `reconnect_status_core.ts`) to the same array this branch appends `welcome_roster_view.ts` to. Merged all three.

No feature code needed touching; all four round-4 review findings (takeover-before-teardown order, unhandled `enter()` rejection, `.ws-roster-row:focus-visible` ring, `role="list"`/`listitem` shape) were already landed in `fb7df0c5e` before this merge.

Verified post-merge: `npx tsc --noEmit` clean; targeted suite (`architecture`, `welcome_roster_view`, `welcome_screen_view`, `welcome_screen_window`, `character_switch`, `charselect_action`, `i18n_completeness`, `localization_fixes`, `world_api_parity`, `focus_visible_guard`, `dialog_root`, `client_shell`): 469 passed, 3 skipped; `npm run i18n:gen` produced no diff; full `npm run gate` green (the only two flakes seen, `charge_parallel_recharge` and `lockpick_hud_sync`, are unrelated pre-existing tests that timed out only under this box's concurrent-agent CPU load and pass cleanly in isolation).

Pushed as a merge commit; ready for re-review.

jgyy:
Addressed the round-5 review findings.

Blocking: the takeover await window was still unguarded. switchCharacter (src/net/character_switch.ts) now takes an isSwitchable collaborator, re-checked immediately after the takeover round trip resolves and before teardown runs. In main.ts, switchWelcomeCharacter latches a new switching flag before the async call (guarding proceedToGame and a second roster click the same way started already guards proceedToGame against itself), and world.onDisconnect flips a switchAborted flag so a session that ends mid-round-trip (proceeded to game, disconnected, kicked) is never torn down or entered behind whatever handler already reacted to that. Added a character_switch.test.ts case pinning that teardown never runs once isSwitchable reports false.

High: the desktop stage kept showing the previous character's model after a switch. enterWorld now calls characterPreview.setAppearance(charselectAppearance(c)) before mounting the Welcome Screen, so the re-parented preview always matches the character actually being entered (a no-op on the ordinary post-login path).

Medium:
- teardown now calls hideReconnectOverlay(), so a drop mid-round-trip cannot leave the overlay stuck behind the new session.
- onTakeoverError no longer calls the terminal fatalOverlay: it resets the switching latch and surfaces the error via window.alert, matching the pre-login takeOverAndEnter precedent of staying on the recoverable, still-alive session instead of killing it over a transient failure.
- setConnectionReady no longer rebuilds the whole roster (paintRoster) on every readiness change; it now flips each existing row's disabled state in place (updateRosterReadiness), so a keyboard user resting on a row keeps focus instead of losing it to the dialog root. Added a welcome_screen_window.test.ts case that focuses a row, calls setConnectionReady, and asserts the same element still has focus; confirmed it fails against the prior paintRoster-based code.

Also picked up the low item that was cheap: teardown now clears the play-resume marker (clearPlayMarker()), so a WebView eviction and reload between teardown and the new session's first populated snapshot cannot resume back into the character just left. Bumped the pinned clearPlayMarker() count in resume_play.test.ts accordingly.

Deferred (low severity, latent only, no live bug today): the currentCharacterId ?? -1 fallback typing, the --ws-continue-ring/--color token duplication, the roster's pointer:coarse hide list scope, and #ws-roster-title lacking data-i18n. Happy to take these in a follow-up if wanted.

Verified: npx tsc --noEmit clean; the reviewer's own test list plus welcome_screen_window/character_switch/resume_play/architecture/i18n_completeness/localization_fixes/world_api_parity/client_shell/dialog_root/char_sort/focus_visible_guard/linkdead/net_online_visibility_reconnect all pass; npm run i18n:gen produced no diff; biome on the six changed files is clean (pre-existing warnings only, no errors, no format diff). Merged against release/v0.28.0 with no conflicts (the branch already contained it). CI is green.

jgyy:
to close it instead, design is bad.

PR #2101, opened 2026-07-18
feat(ui): relayout the vendor Goods window as a landscape item grid

jgyy:
## Summary

- The Vendor "Goods" window rendered every item as its own full-width row in
  a narrow, single-column, scrolling list, forcing a tall narrow window even
  for a modest shop stock.
- Widens `#vendor-window` and wraps the buyable-goods rows and the buyback
  rows each in a `.vendor-goods-grid` container (`display: grid;
  grid-template-columns: repeat(4, minmax(0, 1fr))`), so item tiles (icon,
  name, stack count, price) flow four per row like a real shop counter.
  (Updated per maintainer review: was 3 columns at a narrower window width,
  now 4 columns with the window widened proportionally so tiles still have
  comfortable padding.)
- The tile look is scoped under `.vendor-goods-grid .vendor-item` so the
  shared `.vendor-item` class (also used by the Heroic Quartermaster and the
  crafting recipe list) is unaffected outside the vendor window.
- Mobile collapses the grid to 2 columns via `hud.mobile.css` and keeps the
  existing 40px tap-target floor. The full-height `max-height` sheet rule
  only takes effect for the Heroic Quartermaster (`after-heroic-mobile.png`):
  the normal vendor path always carries `body.vendor-open`, whose split-screen
  rules already set `max-height: none` at higher specificity, so that state
  is unchanged by this PR (clarified per maintainer review).
- Layout only: buy/buyback click handlers, tooltips, sell-junk, and the
  honor balance line are unchanged.
- Desktop width cap: `#vendor-window` widens up to 860px but is capped so
  its right edge always clears the always-open `#bags` column instead of
  overlapping it, floors at the pre-PR 400px on narrow viewports, and divides
  the viewport term by `--window-scale` so the cap is correct at every
  `--ui-scale` (fixed across several maintainer review rounds; see the
  comment above `#vendor-window` in `src/styles/components.css` for the
  derivation). Added `tests/vendor_window_painter.test.ts` coverage that pins
  both the `--window-scale` division and the bags-clearance math across a
  sweep of viewport widths and scales.

## Layout change

```mermaid
graph TD
    subgraph Before["Before: single-column list (400px window)"]
        B1["icon | name | qty | price"]
        B2["icon | name | qty | price"]
        B3["icon | name | qty | price"]
        B4["icon | name | qty | price"]
        B5["... one full-width row per item ..."]
        B1 --> B2 --> B3 --> B4 --> B5
    end
    subgraph After["After: landscape item-tile grid, 4 columns (widened window)"]
        direction LR
        A1["icon\nname\nprice"]
        A2["icon\nname\nprice"]
        A3["icon\nname\nprice"]
        A4["icon\nname\nprice"]
        A5["icon\nname\nprice"]
        A6["icon\nname\nprice"]
        A7["icon\nname\nprice"]
        A8["icon\nname\nprice"]
    end
```

## Screenshots

Desktop, before (3-column grid) vs after (4-column grid, widened window):

before desktop
after desktop

Mobile, before (stops short, leaving the character/player HUD visible below) vs
after (spans the full screen height, top to bottom):

before mobile
after mobile

(Full-size images: `docs/screenshots/vendor-landscape-grid-layout/{before,after}-{desktop,mobile}.png` in this PR's diff.)

Heroic Quartermaster (second tenant of `#vendor-window`), now sharing the same tile grid instead of a stretched single-column list:

after heroic quartermaster

## Test plan

- [x] `npx vitest run tests/vendor_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/client_shell.test.ts tests/bank_window.test.ts tests/heroic_vendor.test.ts tests/ptr_dev_vendor.test.ts tests/vendor_stack.test.ts` — all passed
- [x] `npx @biomejs/biome check --write` on changed files (no diffs)
- [x] Manual verification in the offline client: buy, buyback, sell-junk, tooltips, and the honor balance line all still work in the new grid
- [x] Before/after screenshots refreshed (desktop + mobile), committed under `docs/screenshots/vendor-landscape-grid-layout/`


EnriqueGF (changes requested):
Revisado el diff completo contra los invariantes del repo (raíz, `src/ui/CLAUDE.md`, `src/ui/hud/CLAUDE.md`, `src/styles/CLAUDE.md`).

**Qué ejecuté**

- `npx tsc --noEmit`: limpio.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/vendor_view.test.ts tests/heroic_vendor.test.ts tests/client_shell.test.ts`: todo en verde (el fallo inicial de `localization_fixes` era mi entorno sin `i18n:gen`; tras regenerarlo pasa, 29 tests).

La idea de fondo me parece buena y el reparto está bien hecho: el `.vendor-goods-grid` envuelve las filas sin tocar handlers, tooltips ni el orden del DOM (balance, grid, sell-junk, buyback, hint se conserva), no hay strings nuevos, no se toca `src/sim/` ni el seam `IWorld`. Dos cosas sí me parecen bloqueantes.

---

**1. (media-alta) El ensanchado de `#vendor-window` arrastra al Heroic Quartermaster, que no tiene grid**

`src/styles/components.css:4562` sube `#vendor-window` de `400px` a `min(860px, 92vw)`. Pero `#vendor-window` tiene un segundo inquilino: `openHeroicVendor` pinta sobre el mismo contenedor (`src/ui/hud.ts:10437`, `renderHeroicVendorWindow`), y ese painter añade sus filas **directamente a `el`**, sin envolverlas en `.vendor-goods-grid` (`src/ui/hud/vendor/heroic_vendor_window.ts:57`).

Resultado: la tienda de Marcas Heroicas pasa a ser una lista de una sola columna a 860px de ancho, con `.vi-name { flex: 1 }` estirando cada fila y dejando un hueco enorme entre nombre y precio. Justo el problema de "lista retrato" que la PR arregla, pero al revés.

La descripción dice que "the shared `.vendor-item` class (also used by the Heroic Quartermaster) is unaffected outside the vendor window": cierto para el *estilo del tile*, pero el **ancho de la ventana sí es compartido** y sí cambia. El propio comentario de `scripts/vendor_grid_shot.mjs` reconoce la tenencia compartida, así que la ruta estaba identificada; solo faltó cerrar el caso.

Opciones razonables: envolver también las filas heroicas en `.vendor-goods-grid` (lo más coherente, es el mismo tipo de escaparate), o acotar el ancho con una clase de estado en el contenedor para que el heroico conserve su ancho estrecho.

**2. (media) `100dvh` dentro de `#ui` ignora `--ui-scale`**

`src/styles/hud.mobile.css:3094`:

```css
body.mobile-touch #vendor-window {
  max-height: calc(100dvh - 12px - max(10px, env(safe-area-inset-bottom)));
}
```

`#vendor-window` vive dentro de `#ui`, que lleva `zoom: var(--ui-scale)` (`src/ui/ui_scale.ts:2`, `src/main.ts:1819`). El propio archivo documenta la trampa unas líneas más abajo, en la regla de `.vendor-open`: *"`#ui`'s zoom multiplies every author length, so a raw `50vw` only lands on the real viewport centre at uiScale 1"*. Con `zoom`, `100dvh` se resuelve contra el viewport real y luego se **multiplica** por el zoom, así que:

- con `--ui-scale > 1` la ventana se pasa del borde inferior de la pantalla (lo contrario de lo que busca la PR),
- con `--ui-scale < 1` se queda corta, que es exactamente el bug original.

Solo es correcto a uiScale 1, que es donde se tomó la captura. Todas las reglas vecinas usan la forma del modelo, `calc(var(--app-vh) / var(--ui-scale, 1) - ...)` (líneas 3051, 3084, 3104), y el único precedente de `dvh` crudo cercano, `.party-rows` en la línea 1705, neutraliza el zoom explícitamente con `zoom: 1` antes de usarlo. Sugiero alinearlo con las hermanas:

```css
max-height: calc(var(--app-vh) / var(--ui-scale, 1) - 12px - max(10px, env(safe-area-inset-bottom)));
```

---

**Menores**

- `src/styles/hud.mobile.css:2803`: el comentario dice *"The desktop landscape grid packs 3 tiles per row"*, pero ya son 4. Quedó del paso anterior de la revisión.
- `src/ui/hud/vendor/vendor_window.ts:159`: `buybackGrid` se añade siempre, también cuando `view.buyback` está vacío (justo después del estado vacío). Es inocuo (un div de grid sin hijos no ocupa alto), pero es un nodo muerto; un `if (view.buyback.length)` lo deja más limpio. Mismo caso con `goodsGrid`.
- `src/ui/CLAUDE.md` pide preferir `width: 100%` + `max-width` sobre unidades de viewport como `92vw`, que desbordan al sumar márgenes/padding. Con `box-sizing: border-box` aquí probablemente aguanta, pero es la guía del área.
- El contrato de `src/styles/CLAUDE.md` dice que los valores literales de layout móvil se fijan en `tests/mobile_window_layout.test.ts`. La regla de `max-height` del vendor se sustituyó por otra sin pin; valdría la pena anclar la nueva (es justo el valor que el punto 2 discute).

Nada que objetar en determinismo, i18n, autoridad del servidor, secretos, em/en dashes ni emojis: la PR es puramente de presentación y no roza esos invariantes.

Con el punto 1 y el 2 resueltos, por mí adelante.

jgyy:
Gracias por la revision detallada. He empujado un commit (dad5316ea) que cierra los dos puntos bloqueantes:

1. Heroic Quartermaster grid: heroic_vendor_window.ts ahora envuelve sus filas en .vendor-goods-grid (el mismo contenedor que goods/buyback), en vez de anexarlas sueltas a el. Con #vendor-window ensanchado a min(860px, 92vw) esto evita la lista de una sola columna con hueco enorme entre nombre y precio.
2. 100dvh bajo --ui-scale: la regla body.mobile-touch #vendor-window { max-height: ... } en hud.mobile.css ahora usa calc(var(--app-vh) / var(--ui-scale, 1) - 12px - max(10px, env(safe-area-inset-bottom))), igual que las reglas hermanas, en vez del 100dvh crudo que solo era correcto a uiScale 1.

Tambien los menores:
- Comentario "3 tiles per row" corregido a "4 tiles per row".
- goodsGrid/buybackGrid en vendor_window.ts ahora solo se anaden al DOM cuando tienen filas (if (view.goods.length > 0) / if (view.buyback.length > 0)), sin nodos de grid vacios.
- Anadido pin en tests/mobile_window_layout.test.ts para el nuevo valor de max-height del vendor (evita que dvh crudo vuelva a colarse).

Verificacion local (el gate completo no es fiable ahora mismo en esta maquina por presion de disco compartida entre sesiones, asi que verifique de forma dirigida):
- npx tsc --noEmit: limpio.
- npx biome check sobre los 4 archivos tocados: sin errores.
- npx vitest run tests/mobile_window_layout.test.ts tests/architecture.test.ts tests/heroic_vendor.test.ts tests/vendor_view.test.ts: 61 tests en verde.

CI de GitHub Actions corre en un entorno aislado, asi que no deberia verse afectado por eso; lo confirmo aparte en cuanto termine.

jgyy:
Addressed both blocking findings from the review:

**1. Heroic Quartermaster stretched to the new 860px width with no grid**

`renderHeroicVendorWindow` (`src/ui/hud/vendor/heroic_vendor_window.ts`) now wraps its rows in the same `.vendor-goods-grid` container the regular vendor uses, so it inherits the same 4-column tile layout instead of stretching into a single full-width column at the widened `#vendor-window`. Its rows are shape-compatible with the regular vendor tile (icon, name, price), so reusing the grid class was the simpler and more consistent fix. Added `docs/screenshots/vendor-landscape-grid-layout/after-heroic-quartermaster.png` showing the Heroic Quartermaster's shop now rendering as a proper tile grid.

**2. `100dvh` inside `#ui` ignored `--ui-scale`**

`body.mobile-touch #vendor-window`'s `max-height` in `src/styles/hud.mobile.css` used a raw `100dvh`, but `#vendor-window` lives inside `#ui`, which carries `zoom: var(--ui-scale)`. That means the raw dvh value resolved against the real viewport and was then multiplied by the zoom, overshooting the bottom edge above scale 1 and falling short below it (correct only at scale 1, coincidentally where the original screenshot was taken). Changed it to the same `calc(var(--app-vh) / var(--ui-scale, 1) - ...)` form the sibling mobile rules in the same file already use, and added a pin in `tests/mobile_window_layout.test.ts` asserting the block no longer contains a raw `100dvh` and matches the scaled form.

Also fixed while in the area (also noted as minor in the review): `goodsGrid`/`buybackGrid` in `src/ui/hud/vendor/vendor_window.ts` are now only appended when their row list is non-empty, avoiding a dead empty grid node.

Verified: `npx tsc --noEmit` clean; `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/vendor_view.test.ts tests/heroic_vendor.test.ts tests/client_shell.test.ts` all green (205 passed, 3 skipped). Pushed as commits `dad5316ea` (the fix) and `2511522` (the new screenshot).

EnriqueGF (changes requested):
Revisado el diff completo (CSS, los dos renderers de vendor, el test de layout mobile y el script de captura) contra los invariantes de `CLAUDE.md` (raiz, `src/styles/`, `src/ui/`).

**Que ejecute localmente** (worktree sobre `release/v0.28.0`):

- `npx tsc --noEmit`: limpio.
- `npx vitest run tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/vendor_view.test.ts tests/heroic_vendor.test.ts tests/architecture.test.ts`: 76 pasan.
- `npm run i18n:gen` + `npx vitest run tests/localization_fixes.test.ts`: 29 pasan, 3 skipped.
- `npx @biomejs/biome check` sobre los 6 ficheros tocados: limpio.

Sin hallazgos de sim/determinismo, seam `IWorld`, autoridad de servidor ni i18n: el cambio es puramente de presentacion y no introduce texto nuevo. Un hallazgo bloqueante y dos menores.

---

### Bloqueante: a 860px la ventana de vendor se solapa con `#bags` en desktop

`openVendor()` (`src/ui/hud.ts`) abre SIEMPRE la companion `#bags` junto al vendor. En desktop `#vendor-window` esta centrado (`left: 50%; translateX(-50%)`, `layout.css`) y `#bags` esta anclado a la derecha con su propia reserva (`components.css` `#bags`, `--bags-bar-half` / `--bags-slot-w`). Con los 400px originales (y con los ~640px de 3 columnas) no habia colision; con `width: min(860px, 92vw)` si la hay.

Haciendo la cuenta con las dos reglas:

- borde derecho del vendor = `vw/2 + min(860, 0.92vw)/2`
- borde izquierdo de bags = `0.75vw + 126 - min(310, 0.5vw - 384)/2`

| viewport | solape |
|---|---|
| 1280 | ~112px |
| 1400 | ~109px |
| 1600 | ~59px |
| 1920 | sin solape (~21px de margen) |

O sea: se solapa por debajo de ~1850px de ancho, que es la mayoria de portatiles. Y como `#vendor-window` va despues de `#bags` en `index.html`/`play.html` con el mismo `z-index: 20`, gana bags y **tapa** la parte derecha del vendor.

Esto se ve en vuestras propias capturas, no es teorico. En `before-desktop.png` (1400px, 3 columnas) el vendor termina en x~1018 y bags empieza en x~1023: justo sin tocarse. En `after-desktop.png` (mismo 1400px) el vendor llega a x~1130 y bags lo tapa desde x~1023: el boton **SELL JUNK** aparece cortado por la derecha, y la seccion Buyback y la linea de hint quedan parcialmente ocultas debajo del panel de bags.

Sugerencia: en vez de un `min(860px, 92vw)` plano, acotar el ancho contra la misma reserva que usa `#bags` (algo como `min(860px, calc(var(--app-vw) - 2 * (bags-half + gap)))`), de forma que el vendor nunca se meta bajo la columna de bags y simplemente baje a 3 columnas en anchos medios. Alternativamente `grid-template-columns: repeat(auto-fill, minmax(...))` para que el numero de columnas siga al ancho disponible en lugar de fijar 4.

### Menor: `after-mobile.png` esta obsoleta respecto a HEAD

La captura se anadio en `b02229443`, antes del commit `dad5316ea` que metio el grid en `heroic_vendor_window.ts`. La imagen muestra al Quartermaster Vex en la **lista de una sola columna antigua**, no el grid de 2 columnas que describe el cuerpo de la PR. Ademas el cuerpo la presenta como la ventana de goods del vendor, y lo que se ve es el quartermaster heroico. Conviene recapturarla (y, ya que la regla nueva `body.mobile-touch #vendor-window { max-height: ... }` solo es alcanzable por la via heroica, porque el vendor normal siempre lleva `body.vendor-open` y cae en la regla split de mas abajo, merece la pena que el cuerpo de la PR lo diga explicitamente).

### Menor: grid vacio en el vendor heroico

En `vendor_window.ts` se anadio la guarda `if (view.goods.length > 0)` / `if (view.buyback.length > 0)` antes de insertar el contenedor, pero `heroic_vendor_window.ts` hace `el.appendChild(goodsGrid)` incondicionalmente. Es inocuo hoy (un grid vacio tiene altura 0), pero es una inconsistencia gratuita con la guarda que acabais de anadir al lado.

### Nota de comentario obsoleto

Al borrar `body.mobile-touch #vendor-window { max-height: calc(58vh - 20px); }` queda huerfano el comentario de `#market-window` unas lineas mas abajo, que sigue explicando su regla en terminos de "the vendor's 58vh cap". Merece una frase actualizada para que no envejezca mal.

El enfoque de grid me parece correcto y el resto (guardas de layout mobile, division por `--ui-scale`, scoping bajo `.vendor-goods-grid` para no afectar a `crafting_window`) esta bien resuelto. Solo hace falta resolver el solape con bags.

jgyy:
Thanks for both rounds of review. Status on each finding:

**First review (heroic vendor grid + `--ui-scale` division):** already fixed in `dad5316ea` (heroic quartermaster rows now wrap in `.vendor-goods-grid`; the mobile `max-height` divides by `var(--ui-scale, 1)` and is pinned in `tests/mobile_window_layout.test.ts`), with a follow-up capture of the Heroic Quartermaster grid in `251152243`.

**Second review, blocking (vendor window overlaps `#bags` below ~1850px):** fixed in `e0671b84b`. `--bags-bar-half` and `--bags-gap` are hoisted from `#bags`'s local scope onto `:root` (`src/styles/tokens.css`), so `#vendor-window`'s width now reads the SAME reservation instead of a flat `92vw`:

```css
width: min(
  860px,
  calc(var(--app-vw, 100vw) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px))
);
```

The window's right edge now always clears the bags column instead of sliding under it. I also switched `.vendor-goods-grid` from a fixed `repeat(4, ...)` to `repeat(auto-fill, minmax(140px, 1fr))`, per your alternative suggestion, so the tile count follows whatever width results (4 at full width, 3 or fewer once the cap shrinks the window) instead of squeezing 4 undersized tiles at medium widths. Re-verified locally at 1400px: the vendor window's right edge and the bags column now have a clear gap (updated `docs/screenshots/vendor-landscape-grid-layout/after-desktop.png`).

**Second review, minor (stale `after-mobile.png` showing the old single-column Quartermaster list):** recaptured; `after-mobile.png` now shows the current grid, and I added `after-heroic-mobile.png` showing the Heroic Quartermaster's grid on a touch viewport specifically (the normal vendor's mobile shot doesn't exercise that code path, since only the heroic vendor's max-height rule discussed in the first review is reachable there without `body.vendor-open`).

**Second review, minor (empty grid appended in the heroic vendor):** fixed, `heroic_vendor_window.ts` now guards `if (view.rows.length > 0) el.appendChild(goodsGrid)`, mirroring `vendor_window.ts`'s existing guard.

**Second review, minor (stale comments):** updated the `#market-window` comment that referenced the old "vendor's 58vh cap" and the "packs 4 tiles per row" comment in `hud.mobile.css` to describe the new auto-fill behavior.

Local verification before pushing:
- `npx tsc --noEmit`: clean
- `npx vitest run tests/vendor_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/client_shell.test.ts tests/bank_window.test.ts tests/heroic_vendor.test.ts tests/ptr_dev_vendor.test.ts tests/vendor_stack.test.ts tests/architecture.test.ts`: 249/249 passed
- `npx @biomejs/biome check --write` on the changed files: no diffs
- Re-ran `scripts/vendor_grid_shot.mjs` for desktop, mobile, and heroic-mobile to refresh the screenshots against the fixed CSS

Let me know if there's anything else you'd like adjusted.

EnriqueGF (changes requested):
Reviewed the full diff against the release/v0.28.0 base, plus `src/styles/CLAUDE.md`, `src/ui/hud/CLAUDE.md`, and the surrounding `#bags` / `#ui` scale rules.

**What I ran**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/mobile_window_layout.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/vendor_view.test.ts tests/heroic_vendor.test.ts tests/mobile_window_coverage.test.ts`: 76 passed. (`tests/localization_fixes.test.ts` only failed on a missing `src/ui/i18n.status.json`, which `pretest` generates. Not related to this PR.)
- Rendered the new `#vendor-window` width rule inside a replica of the `zoom: var(--ui-scale)` `#ui` box in headless Chromium, sweeping viewport width and `--ui-scale`.

**What looks good**
Layout-only, no sim/wire/server surface. No new player-visible strings, so the i18n scope is untouched. Scoping the tile look under `.vendor-goods-grid .vendor-item` correctly leaves the crafting recipe list alone. Hoisting `--bags-bar-half` / `--bags-gap` to `:root` is the right call over a copy, and nothing else redefines them on an intermediate ancestor, so `#bags` still resolves the same values. The mobile `max-height` divides by `--ui-scale` correctly.

The blockers are both in the desktop width cap from `e0671b84b`.

---

**1. (high) The width cap has no lower floor, so the window collapses on narrow desktop viewports.**

`components.css:4569`:
```css
width: min(860px, calc(var(--app-vw, 100vw) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px)));
```
The reservation subtracts a fixed 684px. Measured widths at `--ui-scale: 1`:

| viewport | vendor width |
|---|---|
| 1920 | 860 |
| 1280 | 596 |
| 1100 | 416 |
| 1024 | **340** |
| 900 | **216** |
| 800 | **116** |
| 700 | **16** |
| 600 | **0** |

Below about 1084px the window is narrower than the 400px it was before this PR, and at or under 684px it computes to 0 and the vendor is unusable. `#vendor-window` has no `min-width`, and a non-touch browser window at 900px keeps the desktop rules (no `body.mobile-touch`), so this is reachable on a plain resized desktop browser. Suggest a `max(400px, ...)` floor so the worst case degrades to the current behavior rather than to nothing.

**2. (high) `--app-vw` is read raw inside the zoomed `#ui`, so the cap misses the bags column it exists to avoid.**

`#ui` carries `zoom: var(--ui-scale)` and sizes itself `calc(var(--app-vw) / var(--ui-scale, 1))` (`hud.css:3376-3378`), so author lengths inside it live in a scaled space. `--bags-bar-half` / `--bags-gap` are author px in that space, but `--app-vw` is real px, so the two sides of the subtraction are in different units at any `--ui-scale != 1`.

At `--app-vw: 1440px; --ui-scale: 1.25` the rule yields 756 author px, which renders 945 device px. Working in author space: the container is 1152 author px, bags reserves 318 on the right, so the centered window's max safe width is about 468 author px. It is 756, and the right edge lands roughly 150px inside the bags column, which is exactly the overlap this commit was written to fix. Below scale 1 it over-shrinks instead.

The fix is the same division the mobile `max-height` in this very PR already applies:
```css
width: max(400px, min(860px, calc(var(--app-vw, 100vw) / var(--ui-scale, 1) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px))));
```

---

**Minor, non-blocking**

- `tests/mobile_window_layout.test.ts:65` pins the new `max-height` by exact whitespace including newlines and indentation. It will fail on any reformat of that block rather than on a behavior change. Matching on the normalized token sequence would pin the same intent more durably.
- `goodsGrid` / `buybackGrid` are constructed before their loops and dropped when the list is empty. Harmless, but building them lazily would read a little more directly.

Happy to re-review once the width cap gets the `--ui-scale` division and a floor.

jgyy:
Rebased onto the latest `release/v0.28.0` tip to clear the merge conflict. Three real conflicts in `src/styles/hud.mobile.css`, all just sequential insertions from this branch's own earlier commits landing at the same spot as unrelated upstream additions (the social/raid-groups mobile-collapse block, and two of this branch's own comment-wording touch-ups on the vendor grid tile-count); kept both additions, no logic lost. tsc, the parity/architecture/i18n guard tests, and the pre-push floor are all green.

EnriqueGF (changes requested):
Reviewed the full diff against the release/v0.28.0 base, read the surrounding CSS contracts (`src/styles/layout.css` window base, the `#bags` reservation, the mobile sheet groups) and both vendor renderers.

Ran locally: `npx tsc --noEmit` (clean), and `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/vendor_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/heroic_vendor.test.ts tests/vendor_stack.test.ts` (all green; `localization_fixes` needed `npm run i18n:gen` first, which is expected in a fresh worktree and unrelated to this PR). No i18n surface, no sim/server/wire surface, no dashes or emojis in the diff. The DOM restructure itself (grid wrappers, ordering, empty-grid guards) reads correct: the append order is unchanged and the buyback empty state still lands before the grid.

Two things I would like fixed before merge, both in the new desktop width cap.

**1. (high) The desktop width cap does not divide by `--window-scale`, so it reintroduces the bags overlap it was added to fix.**

```css
#vendor-window {
  width: min(860px, calc(var(--app-vw, 100vw) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px)));
}
```

`#vendor-window` lives inside `#ui`, which carries `zoom: var(--ui-scale)`, so `--app-vw` here is an author length that the browser then multiplies by the zoom. `src/styles/layout.css` states this explicitly on the `.window` base ("The #ui zoom factor the viewport clamps divide by ... Override this, never bake a different division into a per-window clamp"), and every sibling clamp follows it (`layout.css:42-43`, `components.css:5182`, `hud.mobile.css:2228`). This is the same trap the PR's own new mobile comment describes, just not applied to the desktop rule.

Concretely at uiScale 1.4 on a 1920px viewport: the cap resolves to 1920 - 684 = 1236 author px, capped by `min()` to 860, then zoom renders it at 1204 real px, i.e. half-width 602 against a bags reservation of 318, so the window slides under the bags column again. Below uiScale 1 it is the opposite, needlessly narrow. Suggested:

```css
width: min(860px, calc(var(--app-vw, 100vw) / var(--window-scale) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px)));
```

(Only the viewport term needs dividing: the bags reservation values are author lengths in the same zoomed space already.)

**2. (medium) The cap has no lower floor, so it is narrower than the pre-PR window on any smaller desktop viewport.**

The subtracted reservation is a flat 684px, so at uiScale 1: 1280px viewport gives 596px (fine), 1024px gives 340px, 900px gives 216px, 800px gives 116px. Before this PR the window was a fixed 400px at every width. So for any browser window under roughly 1084px the change is a regression rather than an improvement, and past about 950px the `minmax(140px, 1fr)` tiles no longer fit their own minimum inside the padding. Combined with (1) at uiScale 1.4 on a 1280px window it resolves to about 230 author px. A floor at the old width keeps it monotonic:

```css
width: max(400px, min(860px, calc(...)));
```

Worth a short comment on why 400px is the floor (the vendor is top-anchored, so at those widths a horizontal overlap with the bottom bags column is not the failure mode anyway).

**3. (medium, tests) The new markup is untested.**

There is no test anywhere that calls `renderVendorWindow` / `renderHeroicVendorWindow`, so neither the grid wrapping nor the two new `length > 0` guards are covered; the only added test pins a CSS string. A small DOM test asserting that goods rows and buyback rows are children of `.vendor-goods-grid`, and that no empty `.vendor-goods-grid` node is appended when a section is empty, would make those guards load-bearing. Same for the desktop width cap: a `css_corpus`-style assertion that the cap divides by `--window-scale` and floors at 400px would keep both fixes above from silently regressing.

**4. (low) The new mobile rule uses `var(--ui-scale, 1)` where the base contract says `--window-scale`.**

For `#vendor-window` the two are equal (it inherits the live scale), so there is no live bug, but `layout.css:25-29` asks for `--window-scale` and `hud.mobile.css:2229-2231` repeats the reasoning. The new test in `tests/mobile_window_layout.test.ts` pins the exact `--ui-scale` string, so it would fail on the corrected form. Worth switching both together, or relaxing the pin.

Also trivial: `docs/screenshots/vendor-landscape-grid-layout/after-heroic-mobile.png` is committed but not referenced from the PR body.

Everything else looks good to me, and the layout direction is clearly right.

jgyy:
Thanks for the thorough round-4 pass, this caught a real regression. Pushed a follow-up commit addressing all four points:

1. (high) The desktop `#vendor-window` width cap now divides `--app-vw` by `--window-scale` before subtracting the bags reservation, matching the contract in `layout.css` and its sibling clamps. Added a comment on the unit mismatch so the next touch doesn't drop it again.
2. (medium) Added a `max(400px, ...)` floor so the window never regresses below the pre-PR fixed width on narrower desktop viewports. Comment explains why 400px is safe: the window is top-anchored, so at that width the failure mode is never horizontal overlap with the bottom bags column.
3. (medium, tests) Added `tests/vendor_window_painter.test.ts`: a jsdom suite that drives `renderVendorWindow` and `renderHeroicVendorWindow` directly and asserts goods/buyback rows land as children of `.vendor-goods-grid`, and that no empty `.vendor-goods-grid` node is appended when a section has no rows (covers both `length > 0` guards). Same file also adds a source-scan assertion pinning the width cap's `--window-scale` division and 400px floor, so both fixes above stay load-bearing.
4. (low) Switched the mobile `max-height` rule from `--ui-scale` to `--window-scale` to match the stated contract, and updated the pin in `tests/mobile_window_layout.test.ts` to match.

Left the screenshot-reference note as-is since it's non-blocking and the after-heroic-mobile capture is already committed under `docs/screenshots/`.

Verified: `npx tsc --noEmit` clean, and `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/vendor_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/heroic_vendor.test.ts tests/vendor_stack.test.ts` all green (385 passed). The new `vendor_window_painter.test.ts` follows the same jsdom pattern as `tests/deed_tracker_painter.test.ts`; biome check on changed files is clean and the pre-push floor passed on the pushed commit.

jgyy:
Rebased onto the latest release/v0.28.0 tip to clear the merge conflict (mergeStateStatus was DIRTY/CONFLICTING). One real conflict in `src/styles/hud.mobile.css`: this branch's own round-4 `#vendor-window` max-height rule and upstream's new `#train-window { overflow-y: auto; }` insertion landed at the same spot. Kept both, no logic lost.

`npx tsc --noEmit` clean; `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/vendor_view.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/heroic_vendor.test.ts tests/vendor_stack.test.ts tests/client_shell.test.ts` all green (478 passed, 3 skipped) after `npm run i18n:gen`. All four rounds of blocking review findings (heroic vendor grid wrap, mobile `--ui-scale`/`--window-scale` division, bags-overlap width cap, width floor + DOM coverage tests) remain in place from prior commits; this push is merge-conflict resolution only, no new behavior changes.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and ran the checks locally. The layout work itself is sound; the new test file is red, so this cannot merge as is.

## What I reviewed and ran

- Full diff (`gh pr diff 2101`), root `CLAUDE.md` and `src/styles/CLAUDE.md`.
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/mobile_window_layout.test.ts tests/heroic_vendor.test.ts tests/vendor_view.test.ts`: all green (84 tests).
- `npx vitest run tests/localization_fixes.test.ts` (after `npm run i18n:gen`): green. No player-visible strings added, so no i18n surface here.
- `npx vitest run tests/vendor_window_painter.test.ts`: **fails**, details below.

Invariants checked and clear: nothing touches `src/sim/`, so determinism and the `IWorld` seam are not in play; no server authority change (layout only, buy/buyback handlers untouched); no `ALLOW_DEV_COMMANDS`, no secrets; no em/en dashes or emojis in the new source or comments; the two painters stay thin consumers of the existing `vendor_view` / `heroic_vendor_view` pure cores, and the new CSS lands in the existing ten-dash `vendor` section, so `css_corpus` stays satisfied.

## Blocking

**1. `tests/vendor_window_painter.test.ts` fails to load at all, so all six DOM tests silently never run.**

The file is `// @vitest-environment jsdom`, and under jsdom `import.meta.url` is an `http://localhost/...` URL, so:

```
const components = readFileSync(new URL('../src/styles/components.css', import.meta.url), 'utf8');
```

throws `TypeError: The URL must be of scheme file` at collection time:

```
FAIL  tests/vendor_window_painter.test.ts [ tests/vendor_window_painter.test.ts ]
TypeError: The URL must be of scheme file
 ❯ tests/vendor_window_painter.test.ts:143:22
Test Files  1 failed
```

Since collection dies, the six painter assertions (the actual point of the file, per its header) contribute zero coverage. Use a `process.cwd()`-relative path (`path.resolve(process.cwd(), 'src/styles/components.css')`), which is what the other CSS-scanning suites in `tests/` do.

**2. Once the read is fixed, the load-bearing assertion still fails (whitespace mismatch).**

I patched the path locally and re-ran: 7 pass, 1 fails.

```
expect(block).toContain('var(--app-vw, 100vw) / var(--window-scale)')
```

The declaration in `src/styles/components.css` is multi-line:

```
var(--app-vw, 100vw) /
var(--window-scale) -
```

so the substring never matches. That means the exact regression this PR's round-4 fix was about (dividing by `--window-scale` rather than comparing unscaled `--app-vw` against zoomed author lengths) has **no working guard**. Either whitespace-normalize `block` before the `toContain`, or assert with a regex like `/var\(--app-vw, 100vw\)\s*\/\s*var\(--window-scale\)/`.

Both issues are in the same file, so one pass fixes them. Worth re-running `npm run gate` afterwards, since the failure is a hard CI red.

## Non-blocking

- `it('exists exactly once', ...)` only asserts `start > 0`, which is "exists at least once". If the intent is uniqueness, count with a global match; otherwise rename it to match what it checks.
- `docs/screenshots/vendor-landscape-grid-layout/after-heroic-mobile.png` is committed but not referenced from the PR body. Either link it (the mobile heroic case is the one the sheet-height change most affects) or drop it.
- The two sibling mobile clamps now disagree cosmetically: `#quest-log-window`/`#quest-dialog` still divide by `--ui-scale` while the new `#vendor-window` rule divides by `--window-scale`. Your comment explains why and notes they are equal in value here, so this is fine to leave, but it is a small inconsistency the next reader will trip over.

## Things I checked and found correct

- Hoisting `--bags-bar-half` / `--bags-gap` to `:root` does not break `#bags`: `--bags-micro-r`, `--bags-slot-w` and `left` all still resolve, and no other rule consumed them.
- `--window-scale` is defined on the `.window` base (`src/styles/layout.css`), so the new `var(--window-scale)` without a fallback resolves for `#vendor-window` and the `width` declaration does not drop.
- The new full-height mobile `max-height` is correctly neutralized in the docked split: `body.mobile-touch.vendor-open #vendor-window` sets `max-height: none` later and at higher specificity, so the bags-companion layout is unchanged.
- Width math sanity-checks: at 1400px viewport the cap yields 716px, at 900px the 400px floor wins, and the `.window` base `max-width` still clamps above that. No overlap with the bottom-right bags column at any width I worked through.
- `auto-fill, minmax(140px, 1fr)` at the 400px floor minus 2x12px padding gives 2 columns, so the narrow case degrades sensibly rather than squeezing 4 tiles.
- Tile styling is correctly scoped under `.vendor-goods-grid .vendor-item`, so `.crafting-recipe-item > .vendor-item` and the other shared users are untouched.

EnriqueGF (changes requested):
Reviewed the full diff, the local CSS/UI conventions (`src/styles/CLAUDE.md`, `src/ui/hud/CLAUDE.md`), and ran things locally on the PR head (`f8d1f174b`).

**What I ran**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/vendor_window_painter.test.ts` (after `npm run i18n:gen`): all green except the one below.
- Read the committed `after-desktop.png` and checked it against the width-cap algebra.

The shape of the change is good: the painters stay thin, the grid wrapping is behind a class rather than inline style, `.vendor-goods-grid .vendor-item` correctly scopes the tile look so the crafting recipe list is untouched, the empty-grid guards are right, and the new jsdom painter test is a genuine behavioral pin rather than a snapshot. No sim/server/i18n surface touched, guards agree.

Two things to fix.

**1. Blocker: `tests/vendor_window_painter.test.ts` fails on this head, so CI is red.**

```
FAIL tests/vendor_window_painter.test.ts > #vendor-window desktop width cap
  > divides the viewport term by --window-scale, not --ui-scale (round 4 review, PR #2101)
AssertionError: expected '#vendor-window {\n    width: max(\n  ...' to contain 'var(--app-vw, 100vw) / var(--window-s...'
```

The assertion at line 155 matches a single-line substring, but Biome wraps the authored declaration across lines (`var(--app-vw, 100vw) /\n          var(--window-scale) -`). `f8d1f174b` fixed how the file is read under jsdom but not the whitespace. Best to normalize before matching, for example `block.replace(/\s+/g, ' ')`, so the pin does not re-break the next time Biome reflows that block.

**2. Medium: the width cap does not actually clear the bags column above roughly 1272px, which is the case it was added for.**

The formula caps the FULL width against a viewport-proportional term, but the constraint is on the HALF width against a constant offset from centre, so the two do not correspond.

`#bags` is `transform: translateX(-50%)` around
`left: calc((100% + 50% + --bags-bar-half + --bags-gap - --bags-micro-r) / 2)` = `0.75W + 128`.
At W = 1400 that is 1178, with `width: min(310px, ...)` = 310, so the bags left edge sits at **1023**.

The vendor window gets `width = W - 2 * (306 + 12 + 24)` = `W - 684` = 716, centred, so its right edge is `700 + 358` = **1058**.

That is about 35px of the vendor window sitting underneath bags, and it gets worse as the viewport widens (the two expressions only coincide near W = 1272; below that the cap is conservative, above it, too generous). It is visible in this PR's own `after-desktop.png`: the vendor panel's right edge disappears behind the Bags panel rather than clearing it. Since `openVendor()` always shows `#bags`, this is the state every desktop player at 1280px and up sees.

Because bags is centred in the gap that starts at `centre + bar-half + gap`, the clearance term is a constant, roughly:

```css
width: max(400px, min(860px,
  2 * (var(--bags-bar-half) + var(--bags-gap) - 24px),        /* clear bags: ~588px */
  calc(var(--app-vw, 100vw) / var(--window-scale) - 48px)     /* stay on screen */
));
```

Worth pinning whichever form you land on with the measured bags left edge rather than only the token names, otherwise the next reflow of that expression passes the pin while still overlapping.

**Two minor notes, not blocking**

- The `it('exists exactly once')` case only asserts `start > 0`; it does not check uniqueness. Either assert the match count or rename it.
- The mobile before/after pair: the default path in `scripts/vendor_grid_shot.mjs` calls `openVendor()`, which adds `body.vendor-open`, and the split-screen rules already set `max-height: none` plus `top`/`bottom` pins. So the new `body.mobile-touch #vendor-window` clamp is overridden there and only takes effect for the Heroic Quartermaster sheet (which your script comment correctly explains, and `after-heroic-mobile.png` covers). The PR body describes it as the Goods window, which reads as a broader fix than it is. Worth a one-line correction so a future reader does not assume the split-screen state changed.

jgyy:
Addressed the latest review round (2026-07-20T08:55:18Z, EnriqueGF, CHANGES_REQUESTED).

1. Blocker: `tests/vendor_window_painter.test.ts` failed on head f8d1f174b because the width-cap assertions matched a raw substring that broke once Biome reflowed the multi-line calc(). Fixed by normalizing whitespace before matching (`block.replace(/\s+/g, ' ')`) and splitting on the marker to count real occurrences instead of `start > 0`.

2. Medium: the desktop width cap did not actually clear the #bags column above roughly 1272px wide, because it capped the FULL vendor width against a viewport-proportional term while the real constraint is a HALF-width comparison against a constant offset from where #bags is centered. Re-derived the cap directly from #bags' own centering algebra (same --bags-bar-half token; --bags-gap cancels out of the derivation), replacing:

   calc(var(--app-vw, 100vw) / var(--window-scale) - 2 * (var(--bags-bar-half) + var(--bags-gap) + 24px))

   with:

   calc(0.5 * (var(--app-vw, 100vw) / var(--window-scale)) + var(--bags-bar-half) - 362px)

   Verified with a numeric sweep (widths 700 to 2560, --ui-scale 0.8 to 1.4) that the vendor window's right edge never crosses the #bags left edge once the 400px floor is no longer binding. Added that sweep as a new test case in tests/vendor_window_painter.test.ts, alongside the fixed pins from item 1.

3. Minor: expanded the PR body to clarify that the full-height mobile max-height rule only takes effect for the Heroic Quartermaster sheet (the normal vendor path keeps body.vendor-open's split-screen override), and added a bullet documenting the width-cap fix and its test coverage.

Not touched: no new merge conflicts against upstream release/v0.28.0 (clean merge-tree check). No sim/server/i18n surface in this change, so no other invariants apply.

Verified locally: `npx tsc --noEmit` clean; `npx vitest run tests/architecture.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/mobile_window_layout.test.ts tests/heroic_vendor.test.ts tests/vendor_view.test.ts` all green (91 tests); `npx @biomejs/biome check --write` on the two changed files (components.css, vendor_window_painter.test.ts) clean. CI on the pushed commit (803e3225e) is green: Format + lint, PR gate (English-only legal) x4, Browser regressions, PR checks (freshness/typecheck/builds), Screenshots of changes all pass.

PR #2102, opened 2026-07-18
feat(ui): lay out the spellbook as a landscape spell-tile grid

jgyy:
## Summary

The Spellbook HUD window (`#spellbook`) rendered every known/learnable ability as one long
scrolling single-column list at a narrow 430px width, even though the game is landscape-only
(desktop and mobile). This widens the window and lays the list out as a multi-column grid so
more of a class's kit is visible at once without scrolling.

- `#spellbook`: widened `430px -> 760px`.
- `.spell-list`: `display: flex; flex-direction: column` -> a 2-column CSS grid
  (`repeat(2, minmax(0,1fr))`), bumping to 3 columns above 900px on non-touch, mirroring the
  existing `.armory-grid` multi-column pattern and its same 900px breakpoint.
- The "empty kit" message (`.spell-list > .spell-sub`, appended directly by
  `spellbook_window.ts` when a class has no learnable abilities) is full-span
  (`grid-column: 1 / -1`) so it doesn't get squeezed into a single narrow cell.
- `body.mobile-touch #spellbook .spell-list` collapses back to a single flex column, so touch
  players keep the existing single-column list with the same >=40x40px tap targets on
  `.spell-row` and `.spell-hotbar-toggle`.

This is a **layout-only** change: `src/ui/spellbook_view.ts` and `src/ui/spellbook_window.ts`
are untouched. Tooltips, drag-to-actionbar, cast-on-click, cooldown/rank display, and the
hotbar +/- toggle all keep working exactly as before; only the container CSS flow changed.

```mermaid
flowchart LR
    subgraph before["Before: 430px, single scrolling column"]
        direction TB
        A1["Attack"] --> A2["Cinderbolt"] --> A3["Hoarfrost Mantle"] --> A4["... scroll"]
    end
    subgraph after["After: 760px landscape, 2-3 column grid"]
        direction LR
        B1["Col A: Attack, Rimelance, Cold Coffin ..."]
        B2["Col B: Cinderbolt, Waterbind, Ignition ..."]
        B3["Col C (>=900px): Hoarfrost Mantle, Skystone ..."]
    end
    subgraph mobile["Touch: collapses to single column"]
        direction TB
        M1["Attack"] --> M2["Cinderbolt"] --> M3["Hoarfrost Mantle"] --> M4["... scroll"]
    end
    before -. ".spell-list: flex column -> CSS grid" .-> after
    after -. "body.mobile-touch override" .-> mobile
```

### Screenshots

**Desktop, before (430px, one column):**

before desktop

**Desktop, after (760px, 3-column grid):**

after desktop

**Mobile, before (single column, unaffected):**

before mobile

**Mobile, after (single column, unaffected):**

after mobile

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/spellbook_view.test.ts` (62 passed)
- [x] `npx @biomejs/biome check --write src/styles/components.css src/styles/hud.mobile.css`
- [x] Manual before/after capture (desktop + mobile, offline client)


EnriqueGF (approved):
Senior review of the spellbook landscape grid. Layout-only change, and it holds up: I could not find a correctness or regression problem. Approving.

### What I checked

- **Scope.** The diff is 18/-3 across `src/styles/components.css` and `src/styles/hud.mobile.css` plus four screenshots. `spellbook_view.ts` and `spellbook_window.ts` are genuinely untouched, so the pure-core / thin-painter split is unaffected and there is no sim, wire, server, or i18n surface in play. No new player strings, no `t()` obligations, no determinism or `IWorld` parity implications.
- **The `.spell-sub` full-span rule.** This was my main worry, because `.spell-sub` is used twice: once as the direct child of `.spell-list` for the empty-kit message, and once nested inside `.spell-row > .spell-text` on every row. The child combinator in `.spell-list > .spell-sub` scopes it to the empty message only, so per-row summaries do not get promoted to full-span cells. Confirmed against `SpellbookWindow.render()` and `appendRow()`.
- **Layer order and the touch override.** `hud.mobile.css` is in the `hud-mobile` layer, which the `index.css` barrel orders after `components`, so `body.mobile-touch #spellbook .spell-list` wins over the grid regardless of specificity. Touch also never sees the 760px: `#spellbook` is already pinned to `width: min(560px, calc(var(--app-vw) / var(--ui-scale, 1) - 20px))` in the narrower mobile size group, so the tap-target rules below it are unchanged.
- **Overflow on narrow desktop.** `.window` in `layout.css` carries `max-width: calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)`, so the wider fixed width cannot push the centered window offscreen on a small desktop viewport.
- **Row rendering inside narrower cells.** `.spell-row` is now a grid item, so it stretches to the track height and hover/focus backgrounds stay uniform across a row. `.spell-name` / `.spell-sub` carry `overflow-wrap: anywhere` with no ellipsis, so tighter columns wrap rather than truncate: no information is lost. The committed `after-desktop.png` confirms the 3-column case reads cleanly.
- **Precedent.** The grid shape (`repeat(N, minmax(0, 1fr))` plus `@media (min-width: 900px) { body:not(.mobile-touch) ... }`) matches `.armory-grid` exactly, including the breakpoint, so this follows the pattern the styles directory already established rather than inventing one.

### What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/spellbook_view.test.ts tests/architecture.test.ts`: 86 passed.
- `npx vitest run tests/localization_fixes.test.ts`: 29 passed, 3 skipped. (It failed on my first attempt only because `npm run i18n:gen` was still writing the 23MB `i18n.status.json` underneath it. A clean rerun is green: not a real failure, just worth knowing if you hit it.)
- Base check: the branch is up to date with `origin/release/v0.28.0` and nothing has landed on `src/styles/` on the base since the fork point, so there is no merge-conflict exposure.

### Notes, none blocking

1. **Breakpoint unit mismatch.** `@media (min-width: 900px)` reads raw viewport pixels, while the window itself sizes in scaled units (`--window-scale` / `--ui-scale`). At a large UI scale on a viewport just past 900px, the window gets clamped by the `.window` max-width while still in 3-column mode, so the columns land tighter than the screenshot shows. `.armory-grid` has the exact same property, so this is inherited precedent rather than something you introduced, and I would not hold the PR for it.
2. **The mermaid diagram in the description is column-major** ("Col A: Attack, Rimelance, Cold Coffin"), but CSS grid fills row-major. The committed screenshot shows the real order correctly (Attack / Cinderbolt / Hoarfrost Mantle across the first row). Description only, the code is right.
3. **No test pins the new rules.** `tests/mobile_window_layout.test.ts` already pins literal mobile layout values for other windows, so there is an established slot for one. The single thing here that is easy to silently regress is the `body.mobile-touch` collapse back to a flex column: if someone later reorders or dedupes those layers, touch would inherit a 2-column grid with the 52px rows and nobody would notice until a device test. A one-line pin on that override would be cheap insurance. Optional, and fine as a follow-up.

PR #2103, opened 2026-07-18
feat(ui): relayout Bags window to a wide landscape shape

jgyy:
## Summary
- The Bags window was a single narrow portrait column (bag-bar, filter chips, grid, money all stacked vertically), so most of the item grid scrolled out of view even with a modest inventory.
- Widens `#bags` and splits its body into a fixed-width left rail (bag-bar, filter/sort tools) beside the slot grid at wide viewports, so the grid lays out several columns per row instead of one long scroll.
- The landscape shape is gated behind `@media (min-width: 1800px)` and scoped away from `body.bank-open` / `body.mobile-touch`: below that width, with the bank docked, or on the mobile sheet, `#bags` stays the original portrait column. This is a narrower audience than "desktop" in general; only wide desktop viewports (roughly 1800px CSS width and up) get the landscape shape.
- The money row (wallet link, Claudium launcher, coin display) is a direct `#bags` child appended after the grid in the DOM, never inside the rail: DOM order is bag-bar, filter bar, grid, money in every state, matching keyboard focus order. Both layouts place it visually where it belongs purely through CSS (grid placement in landscape, plain flow in portrait), never through `order`, which reorders paint but not focus navigation.
- Fixed a real bug found while re-verifying in the actual game (not just the CSS in isolation): `Hud.toggleBags`/`openBank` etc. set `#bags`'s `style.display` to `'flex'` as an inline style on open (the repo-wide window open/close convention), and an inline style always outranks any external stylesheet rule. The landscape media query's `display: grid` needed `!important` to actually apply once wired into the real app; without it the whole landscape layout was silently dead in production despite passing CSS-only measurements.
- Pure CSS/DOM layout change, no behavior change: drag/drop, right-click actions, tooltips, filters and the money display all work as before.

```mermaid
flowchart LR
    subgraph portrait["Portrait (default: under 1800px, mobile, bank-docked)"]
        direction TB
        T1[panel-title] --> B1[bag-bar]
        B1 --> F1[bag-filter-bar]
        F1 --> G1[bag-grid]
        G1 --> M1[money]
    end
    subgraph landscape["Landscape (min-width: 1800px, desktop only)"]
        direction TB
        T2[panel-title]
        T2 --> R2
        subgraph R2["bag-rail (fixed width column)"]
            direction TB
            B2[bag-bar] --> F2[bag-filter-bar]
        end
        R2 -.beside, full height.-> G2[bag-grid]
        R2 --> M2["money (own grid row, below rail)"]
    end
```

## Screenshots
Desktop, portrait (default, under 1800px):

Before | After
--- | ---
before | after

Desktop, landscape (1920px, the actual shipped landscape shape, captured through the real `Hud.toggleBags` path so the `!important` fix is verified end to end):

landscape

Mobile (unchanged, still portrait):

Before | After
--- | ---
before | after

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/bags_view.test.ts tests/bank_window.test.ts tests/bags_window.test.ts tests/client_shell.test.ts tests/per_entry_css_wiring.test.ts tests/focus_manager.test.ts tests/dialog_root.test.ts tests/hud_perf_budget.test.ts tests/item_instance.test.ts` (all green)
- [x] `npx @biomejs/biome check` on the changed files (clean; one pre-existing unused-import warning in `bags_window.ts`, unrelated to this change)
- [x] `npm run gate`
- [x] Manual verification via a headless Chromium run of the actual game (`window.__game.hud.toggleBags()`), at 1920px (landscape), 1440px (portrait fallback), and a mobile landscape viewport, confirming the money-row focus-order fix and that the landscape `display: grid` actually applies once wired into `Hud`.


EnriqueGF (changes requested):
Reviewed the full diff (`components.css`, `hud.mobile.css`, `bags_window.ts`) plus the surrounding `#bags` / `.bag-grid` / `body.bank-open #bags` rules and the mobile block.

**What I ran**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/css_value_validity.test.ts tests/bags_view.test.ts tests/bank_window.test.ts`: all green (450 tests).
- No `t()` / i18n surface touched, no `src/sim/` or wire involvement, no dashes or emojis in the diff. Base is correct (`release/v0.28.0` tip is an ancestor of the PR head).
- Because this is a pure layout change, I measured it in real Chromium against the repo's actual stylesheets (`src/styles/index.css`, `#ui > #bags` structure), before vs after, at a range of widths.

The direction is good, the Bags window genuinely wanted a landscape shape. But the measurements say the new layout breaks at several real, common sizes. Requesting changes on three findings.

---

### 1. Blocker: at 1366 and 1440 CSS px the money row and most of the grid are clipped away and unreachable

`--bags-slot-w` is viewport-derived (`100vw - 50% - 306px - 12px - 62px`), so the window is only ~640px wide on a very large display. On common laptop widths it lands around 300 to 420px, which is not enough for a 190px non-shrinking rail plus a usable grid. `align-content: flex-start` also stops the single flex line from being height-constrained, so the line grows to content height and `overflow: hidden` simply clips it. Nothing scrolls.

Measured (20 items in the bag, 900px tall viewport):

| viewport | before: grid width / money visible / last item visible | after |
|---|---|---|
| 1600 | 282px / yes / yes | 190px / yes / yes |
| **1440** | 282px / yes / yes | **110px (2 cols) / no / no** |
| **1366** | 275px / yes / yes | **73px (1 col) / no / no** |
| 1280 | 232px / yes / yes | 232px / yes / no |

At 1366 the rail's laid-out height is 1460px inside a window clipped to 360px, so the money line renders at y=2008 against a window bottom of y=894. It is not scrolled off, it is painted outside the clip and there is no scrollbar to reach it. That is a functional regression: the player cannot see their money or most of their bag at the single most common laptop width.

### 2. Blocker: mobile is not unchanged, the grid collapses to a one-column sliver

The mobile override sets `flex-direction: column` but does not reset `flex-wrap: wrap`, `align-content: flex-start` or `column-gap` from the desktop rule. With a column direction and a height-constrained container, `wrap` makes the children wrap into *additional columns* instead of stacking, and `align-content: flex-start` packs those columns at content width instead of filling the sheet.

Measured at 390x844 with `body.mobile-touch`: grid width 44px, 1 column, last item not visible (before: full-width grid, everything visible).

Your own committed evidence shows this. `before-mobile.png` has a healthy full-width grid of 16 slots across the sheet; `after-mobile.png` has the grid collapsed into a single vertical strip of items pinned to the right edge, clipped at the bottom. So the "Mobile (unchanged, still portrait)" claim in the PR body is not what the screenshot shows.

Confirmed the mechanism by patching it live in the same harness:

```css
body.mobile-touch #bags { flex-wrap: nowrap; align-content: stretch; column-gap: 0; }
```
restores grid width 342px, 8 columns, everything visible. Worth folding into the existing mobile override.

### 3. Major: the bank-docked Bags window is squeezed to 2 columns

`body.bank-open #bags` (components.css, the `body.bank-open` block) pins `width: 320px` but never opts out of the row layout, so the 190px rail plus the 12px gap eats most of it. Measured at 1920 with `body.bank-open`: grid 90px wide, 2 columns, versus roughly 300px and 6 columns before. Banking is exactly when you are scanning a lot of slots. The docked state probably wants `flex-direction: column` (or a much narrower rail) alongside its width pin.

---

### Smaller notes

- **Two adjacent `#bags .bag-grid` blocks** (`components.css` around the rail rules) with no rule between them. Please merge into one; the second block's `align-content: start` also just repeats the base `.bag-grid` rule.
- **No test covers `.bag-rail`.** The change introduces a new DOM container that three stylesheets now depend on, and nothing pins it. A small assertion in `tests/bags_view.test.ts` or `tests/bags_window.test.ts` that the bag-bar, filter bar and money row render inside `.bag-rail` would keep the CSS contract from silently drifting.

### Suggested direction

The underlying tension is that a fixed 190px rail is being placed inside a window whose width is viewport-derived and frequently narrower than the rail plus a usable grid. Two options that would hold up:

1. Gate the landscape layout behind a width condition (a container query on `#bags`, or a `@media (min-width: ...)` chosen so `--bags-slot-w` is comfortably above rail + gap + a few columns), and keep the existing portrait stack as the default everywhere else, including mobile and bank-docked.
2. Or decouple the window width from `--bags-slot-w` for the landscape state so it actually gets the ~640px the layout assumes, and make the rail shrinkable (`flex: 0 1 var(--bags-rail-w); min-width: 0`).

Either way, please drop `align-content: flex-start` or give the rail and grid a real height constraint, so overflow scrolls instead of clipping, and re-capture the mobile screenshots after the fix.

Happy to re-review once these are addressed. The landscape idea is right, it just needs to degrade safely below ~1600px and stay opted out on mobile and in the bank dock.

jgyy:
Thanks for the measured writeup, all three findings addressed in d92826975:

1. **Clipping at 1366/1440px**: gated the row layout behind `@media (min-width: 1800px)`, chosen with margin above rail (190px) + gap + several 38px columns even at the narrower end of that range. Below it, and everywhere else, `#bags` stays the portrait column (`flex-direction: column`).
2. **Mobile collapse**: `.bag-rail` is now `display: contents` outside the breakpoint (portrait/mobile/bank-docked), so the bag-bar/filter-bar/money flow straight into the `#bags` column in their original order (via `order` on each), instead of the row-flex rules leaking through on a touch device reporting a wide viewport.
3. **Bank-docked squeeze**: `body.bank-open #bags` now explicitly sets `flex-direction: column` and forces `.bag-rail { display: contents }`, opting out of the row layout regardless of viewport width.

Also added a `tests/bags_window.test.ts` case pinning the bag-bar/filter-bar/money order inside `.bag-rail`, per your smaller note.

Re-captured screenshots aren't in this comment since the CSS-corpus and mobile-layout test suites (`tests/css_corpus.test.ts`, `tests/mobile_window_layout.test.ts`, `tests/mobile_window_coverage.test.ts`, `tests/bags_view.test.ts`, `tests/bank_window.test.ts`, `tests/bags_window.test.ts`, `tests/client_shell.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/css_value_validity.test.ts`, `tests/styles_extraction.test.ts`) all pass locally (567 passed, 3 skipped) and `tsc --noEmit` is clean; happy to add fresh before/after shots at the gated breakpoint if useful for re-review.

EnriqueGF (changes requested):
Reviewed the full diff against `src/styles/CLAUDE.md`, `src/ui/CLAUDE.md` and the root invariants. Scope is genuinely presentation-only: no `src/sim/`, no wire/`IWorld` surface, no server authority, no new player-visible strings (so no i18n obligation), no dev-command or secret exposure, no dashes/emoji. The DOM change in `bags_window.ts` preserves the scroll-restore, the empty-bag filter-bar skip, and the Claudium click wiring.

Ran locally: `npx tsc --noEmit` (clean), and `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/bags_window.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/world_api_parity.test.ts` (all green; `localization_fixes` needs `npm run i18n:gen` first in a fresh worktree, unrelated to this PR).

## Blocking: the two opt-outs do not fully opt out

`body.bank-open #bags` and `body.mobile-touch #bags` only reset **two** properties: `flex-direction: column` on `#bags` and `display: contents` on `.bag-rail`. Everything else the `@media (min-width: 1800px)` block declares still matches and still applies, because `display: contents` on the rail does not stop descendant selectors from matching its children.

The leaked declarations are:

- `#bags .bag-filter-bar { order: initial }`
- `#bags .bag-grid { order: initial; grid-template-columns: repeat(auto-fill, minmax(38px, 1fr)) }`
- `#bags .bag-rail .money { order: initial; flex-direction: column; align-items: flex-start; margin-top: auto }`
- `#bags .bag-rail .bag-bar { order: initial; margin: 0; flex-wrap: wrap }`

Nothing in either opt-out block, or anywhere else in `hud.mobile.css`, re-declares `order` for these.

**Repro:** 1920px desktop (i.e. above the 1800px breakpoint), open the bank. `body.bank-open` is set, so `#bags` is back to `flex-direction: column` and the rail is `display: contents`, but every child's `order` has been reset to `0`. Ties on `order` resolve by DOM order, and the new DOM order is panel-title, bag-bar, filter-bar, **money**, **grid**. So the money line renders *above* the item grid instead of below it, which is a visible regression against the pre-PR layout. On top of that the money row renders as a vertical stack (`flex-direction: column`, `align-items: flex-start`, `margin-top: auto`) rather than the horizontal coin row, and the 320px-wide docked grid gets the tighter 38px column floor it was explicitly meant to avoid.

The same thing happens on the mobile sheet whenever the touch device reports a CSS viewport of 1800px or more, which is exactly the case the `hud.mobile.css` comment says the override exists to handle. That override currently does not achieve it.

The base `order` values (0/1/2/3) only survive because nothing overrides them; the moment the media query matches, the layout depends on `order` being restored, and neither opt-out restores it.

**Suggested fix**, which also removes the need for the two opt-out blocks entirely and is robust against future additions to the landscape block: scope the media query body itself rather than patching it afterwards, e.g. prefix every selector inside `@media (min-width: 1800px)` with `body:not(.bank-open):not(.mobile-touch)`. Then the portrait rules apply untouched in both excluded states and there is nothing to keep in sync. Whatever shape you pick, please pin it: a corpus assertion that the landscape declarations are unreachable under `.bank-open` / `.mobile-touch` would have caught this.

## Minor

- `#bags .money .money-inline { margin-left: auto }` is unchanged, so in the landscape rail (where `.money` becomes a column flex with `align-items: flex-start`) the auto margin now acts on the cross axis and pushes the coin block to the rail's right edge, out of line with the `$WOC` chip above it. Cosmetic, but probably not the intent.
- The `tests/client_shell.test.ts` change pins the `#bags .bag-grid` block as an exact multi-line string including the new `order: 2`. That is consistent with the surrounding style, but it makes the test brittle against the fix above; worth revisiting when you restructure the rules.

Everything else looks good to me, and the breakpoint gating from `d928269` is the right call. Happy to re-review once the opt-out leak is closed.

jgyy:
Thanks for the detailed measurements, this was pushed in commit d92826975.

All three findings are addressed by gating the landscape rail behind a viewport-width breakpoint instead of applying it everywhere:

1. **1366/1440px clipping (blocker)**: `#bags` now stays the original portrait column (`flex-direction: column`, `width: min(310px, --bags-slot-w)`) by default. The row layout (fixed-width rail beside the grid, `width: min(640px, --bags-slot-w)`) only applies inside `@media (min-width: 1800px)`, chosen with a comfortable margin over rail width + gap + several grid columns given how `--bags-slot-w` scales with viewport width. Below that breakpoint the money row and the rest of the grid render exactly as before.

2. **Mobile collapse to a one-column sliver (blocker)**: since the landscape rule is now gated at 1800px min-width, mobile viewports never see the row layout in the first place, so the `flex-wrap`/`align-content` mismatch you found cannot occur. I also added an explicit `flex-direction: column` + `.bag-rail { display: contents }` override on `body.mobile-touch #bags` so a touch device that reports an unusually wide CSS viewport still gets the portrait stack.

3. **Bank-docked squeeze to 2 columns (major)**: `body.bank-open #bags` now forces `flex-direction: column` and `.bag-rail { display: contents }` regardless of viewport width, so the 320px-pinned docked window always gets the portrait stack, never the rail layout.

Smaller notes:
- Merged the two adjacent `#bags .bag-grid` blocks into one (in both the base portrait rule and the landscape media block).
- Added a DOM-contract pin in `tests/bags_window.test.ts` asserting the bag-bar, filter bar and money row all render inside a single `.bag-rail` container, in that order.

Implementation note: `.bag-rail` is `display: contents` below the breakpoint, so its children flow directly into the `#bags` flex column and keep their original visual order via explicit `order` values (bag-bar, filter, grid, money) even though the money row is now nested in the rail element in the DOM. Inside the `@media (min-width: 1800px)` block it becomes a real shrinkable flex column (`flex: 0 1 var(--bags-rail-w); min-width: 0`) beside the grid, per your suggested direction.

Verified: `npx tsc --noEmit` clean, the full previously-listed targeted test file list plus the new `.bag-rail` pin (310 tests) all green, `npm run build` succeeds. The pre-existing unused-import warning in `bags_window.ts` you noted is unrelated to this change and untouched.

jgyy:
Thanks for both review rounds, EnriqueGF. Summary of what changed since:

**Round 1 (1366/1440 clipping, mobile collapse, bank-docked squeeze)**: addressed in `d928269` (already on this branch before this reply) by gating the landscape rail behind `@media (min-width: 1800px)`, keeping the portrait stack as the default everywhere else (desktop below the breakpoint, mobile, bank-docked), and pinning `.bag-rail` in `tests/bags_window.test.ts`.

**Round 2 (the opt-out leak: `order`/grid column floor/money-row layout still applying under `body.bank-open` / `body.mobile-touch` because `display: contents` doesn't stop descendant selectors from matching)**: fixed just now in `c461c7f9f`. Rather than patching the two properties back afterward, every selector inside the `@media (min-width: 1800px)` block is now scoped with `body:not(.bank-open):not(.mobile-touch) #bags ...`, so the block is structurally unreachable in both excluded states and there's nothing left to fall out of sync. This also let me drop the now-redundant `body.bank-open #bags .bag-rail { display: contents }` / mobile-touch twin entirely.

Also folded in your two minor notes from round 2:
- Cleared `.money .money-inline`'s `margin-left: auto` inside the landscape rail specifically (it's a column flex there with `align-items: flex-start`, so the auto margin was acting on the cross axis and pushing the coin block to the rail's right edge instead of staying under the `$WOC` chip).
- Added a corpus test (`bags_window.test.ts`, "landscape rail stays unreachable in the opted-out states") that pins every `#bags` selector under the media query carrying the `:not(.bank-open):not(.mobile-touch)` prefix, and that the old patch-up opt-out shape doesn't reappear.

Verified locally: `npx tsc --noEmit` clean; `npx vitest run tests/bags_window.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/css_value_validity.test.ts tests/client_shell.test.ts tests/bank_window.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/architecture.test.ts tests/bags_view.test.ts tests/world_api_parity.test.ts` all green (554 tests); `npx @biomejs/biome check` on the changed files clean.

Screenshots: added `docs/screenshots/bags-landscape-layout/after-desktop-1440-scoped-fix.png` and `after-desktop-1366-scoped-fix.png`, captured against the fixed branch at the two previously-broken laptop widths from round 1; both now show the window correctly falling back to the safe portrait column with the full grid and money row visible instead of the clipped 2-column/1-column layout. I wasn't able to get a fresh mobile-viewport and bank-docked capture through in this pass (the headless capture script stalled mid-run); happy to add those as a follow-up if you'd like them before merge, but the CSS fix itself is covered by the new corpus test regardless of screenshot coverage.

Happy to re-review.

EnriqueGF (changes requested):
## Review

Reviewed the full diff, the local `src/styles/CLAUDE.md` / `src/ui/CLAUDE.md` conventions, and the root invariants.

**What I ran locally** (worktree on `pr-2103`, base `release/v0.28.0`):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/bags_window.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/mobile_window_layout.test.ts`: all green (186 tests). The one initial failure was a missing `src/ui/i18n.status.json` in my worktree, fixed by `npm run i18n:gen`, not a PR issue.
- Rendered the real `#bags` CSS block (sliced verbatim out of `components.css`) in headless Chromium at several viewports to check the layout empirically.

**What looks good:** this is a pure presentation change, so determinism, `IWorld` parity, server authority and i18n are all untouched. The `body:not(.bank-open):not(.mobile-touch)` scoping of the whole media query is the right call over a patch up opt out, and the added tests pin both the `.bag-rail` DOM shape and that scoping. Good follow up on the earlier review findings.

---

### Blocking: in landscape mode the slot grid never scrolls, it clips

`#bags` keeps `overflow: hidden`, and the landscape block turns it into `flex-direction: row` + `flex-wrap: wrap`. In a wrapped flex container the cross size of a line is derived from its content, and `#bags` has only a `max-height`, never a definite `height`. So the rail/grid line grows to the grid's full content height instead of being bounded, `.bag-grid`'s `overflow-y: auto` never engages, and the overflow is clipped by the ancestor with no scrollbar anywhere.

Measured in Chromium with a realistic rail (bag bar, filter chips, search input, money) and a max capacity inventory (`BACKPACK_SLOTS` 16 + 4 sockets x `bagSlots` 14 = 72 slots):

| viewport | columns | grid scrolls? | slots unreachable |
|---|---|---|---|
| 1800 x 900 | 8 | **no** | **8** |
| 1800 x 1080 | 8 | **no** | **8** |
| 1920 x 1080 | 9 | no | 0 |
| 2560 x 1080 | 11 | no | 0 |

At 1800px, the narrowest width where the block applies, a fully bagged character loses a whole row of items with no way to reach them. For contrast, the portrait fallback at 1400px reports `gridScrolls: true` and behaves correctly.

Note that `gridScrolls` is `false` at *every* width and slot count in landscape, so this is not a tuning problem. Today it only bites in the 1800 to 1900px band, but any future bag capacity increase, a larger slot size, or a shorter viewport widens that band silently. It is also the same failure the PR set out to fix, just relocated to a different breakpoint.

**Suggested direction.** Making the landscape body a CSS grid gives the tracks a definite size, so the slot grid gets a real bounded height and its own `overflow-y: auto` finally engages. I verified this restores scrolling at 120 and 400 slots while keeping the rail beside the grid, and leaves the portrait fallback untouched:

```css
@media (min-width: 1800px) {
  body:not(.bank-open):not(.mobile-touch) #bags {
    display: grid;
    height: min(360px, calc(100vh - 18px));
    grid-template-columns: var(--bags-rail-w) minmax(0, 1fr);
    grid-template-rows: auto minmax(0, 1fr);
    column-gap: 12px;
  }
  body:not(.bank-open):not(.mobile-touch) #bags .panel-title { grid-column: 1 / -1; }
  body:not(.bank-open):not(.mobile-touch) #bags .bag-rail { min-height: 0; overflow-y: auto; }
  body:not(.bank-open):not(.mobile-touch) #bags .bag-grid { min-height: 0; overflow-y: auto; }
}
```

That also drops the `flex-wrap` / `align-content` / `order: initial` bookkeeping, since grid placement replaces it. Whichever fix you pick, please add a regression test that asserts the landscape grid stays scrollable (a jsdom assertion will not catch this, so a `tests/browser/` case or a pinned CSS assertion that `#bags` carries a definite `height` in the landscape block would both work).

### Minor, non blocking

- `after-desktop-1366-scoped-fix.png` and `after-desktop-1440-scoped-fix.png` are committed but never referenced from the PR body. Either link them or drop them.
- The PR body screenshots point at `raw.githubusercontent.com/jgyy/...` on a fork branch, which will rot once that branch goes away. The in repo `docs/screenshots/` paths are the durable reference.

Happy to re-review once the scroll containment is sorted. The rest of the change is in good shape.

jgyy:
Thanks for round 4. Verified the round 1-3 fixes are in place: the `#bags` landscape media query at `src/styles/components.css` is fully scoped with `body:not(.bank-open):not(.mobile-touch)` on every selector, and `tests/bags_window.test.ts` pins both the `.bag-rail` DOM shape and the scoping (no patch-up opt-out).

Fixed the round 4 blocking finding: `#bags` in the landscape block only ever carried `max-height`, so wrapped in `flex-direction: row` + `flex-wrap` the row's cross size derived from content instead of being bounded, and `.bag-grid`'s own `overflow-y: auto` never had a constrained box to scroll inside, clipping slots against the ancestor `overflow: hidden` with no scrollbar. Switched the landscape body from flex to CSS grid (`grid-template-columns: var(--bags-rail-w) minmax(0, 1fr)`, `grid-template-rows: auto minmax(0, 1fr)`, explicit `height`), which gives the row a definite size the rail and grid tracks can each scroll within, and dropped the now-dead `order: initial` bookkeeping grid placement replaces.

Added a CSS-source pin in `tests/bags_window.test.ts` asserting the landscape `#bags` rule carries `display: grid` and a real `height` (not just `max-height`), since a jsdom assertion can't exercise the actual scroll/clip behavior.

Ran `npx tsc --noEmit` and the full requested test set (architecture, localization_fixes, bags_window, client_shell, css_corpus, mobile_window_layout, css_value_validity, styles_extraction, mobile_window_coverage, bags_view, bank_window): all green. Pushed to this branch.

EnriqueGF (changes requested):
Reviewed the full diff (CSS + DOM + tests) against the root and `src/styles/CLAUDE.md` invariants, and ran the checks myself in a clean worktree on `release/v0.28.0`.

## What I ran
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/bags_window.test.ts tests/client_shell.test.ts tests/architecture.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/styles_extraction.test.ts tests/per_entry_css_wiring.test.ts`: 205 passed.
- `npm run i18n:gen && npx vitest run tests/localization_fixes.test.ts` (S3 guard): 29 passed. (It fails without the generated artifact first; that is the normal `pretest` step, not a PR issue.)
- `npx @biomejs/biome check` on the three changed source files: only the pre-existing `equip_drop_core` unused-import warning you already called out.

No sim/server/wire surface is touched, so determinism, `IWorld` parity and server authority are not in play. No new player-visible strings, no locale-overlay edits, no dashes or emojis. The scoping rework in `c461c7f9f` is the right call: `display: contents` genuinely does not stop descendant selectors from matching, so scoping the whole block beats a patch-up opt-out, and `tests/bags_window.test.ts` pins it well.

## Findings

### 1. Focus order no longer matches visual order in the portrait layout (medium)

`bags_window.ts` now appends the money row into `.bag-rail`, i.e. **before** `.bag-grid` in the DOM, and the portrait layout restores the old appearance with `order: 2` on `.bag-grid` and `order: 3` on `.money`. But `order` only reorders visual/box order, never sequential focus navigation, which follows DOM order.

The money row is not inert: `wocBalanceHtml()` emits a `<button data-wallet-action>` in the connect/link/reconnect and unverified-preview states, and `claudiumLauncherHtml()` emits a `<button data-claudium-launcher>` (`src/ui/hud.ts:3987`, `:4011`). The grid cells are real `<button>`s too (`src/ui/bags_window.ts:502`).

Repro: open Bags on any viewport under 1800px (or on mobile, or with the bank docked, i.e. the default path for essentially every user today), tab through the window. Focus now goes bag-bar -> filter bar -> **wallet/Claudium buttons** -> item grid, while those buttons are still painted *below* the grid. Before this PR the DOM and visual order agreed. That is a WCAG 2.4.3 / 1.3.2 regression on the layout the overwhelming majority of sessions actually get, and this repo carries an explicit a11y suite plus focus-order posture, so it is worth fixing rather than accepting.

Suggested fix, which also deletes the `order` bookkeeping entirely: keep the original DOM order (bag-bar, filter bar, grid, money) and leave money a direct child of `#bags`; let `.bag-rail` wrap only bag-bar + filter bar. In the landscape grid you already control placement by template, so give money its own explicit placement (e.g. rail in row 2 col 1, grid in row 2 col 2, money in row 3 col 1, or place money inside the rail track with `grid-row`). Portrait then needs no `order` at all, the `tests/client_shell.test.ts` pin does not have to grow `order: 2`, and DOM order stays the accessible order in every state.

### 2. The PR body screenshots do not show the shipped behavior (medium, process)

The headline `after-desktop.png` was committed in `bed1f46f0`, before `d92826975` gated the rail behind `@media (min-width: 1800px)`. The two follow-up captures tell the story: `after-desktop-1366-scoped-fix.png` (303x382) and `after-desktop-1440-scoped-fix.png` (310x386) are both *portrait*, i.e. the unchanged layout. So at 1366 and 1440 this PR is a visual no-op, and there is no committed screenshot at >= 1800px showing the landscape shape that actually ships.

Two asks: (a) capture and commit an after-shot at >= 1800px so the screenshot requirement covers the real behavior, and (b) update the body, which still reads as if desktop generally gets the landscape relayout. Worth stating plainly in the summary that the new shape only appears at >= 1800px, since that is a much narrower audience than "desktop" and reviewers/maintainers will want to weigh whether 1800px is the threshold you want. (Also: the body's image URLs point at `raw.githubusercontent.com/jgyy/...`; prefer the committed `docs/screenshots` paths so they survive the fork branch going away.)

### 3. Fixed 360px height in the landscape block (minor, question)

`height: min(360px, calc(100vh - 18px))` makes the window a constant 360px tall at >= 1800px regardless of how few items are in the bag, where it previously shrank to content. I understand why you reached for a definite height (an `fr` track needs one), but grid does resolve `fr` tracks against a `max-height` when content overflows, unlike a flex line's cross size, so `max-height` alone may give you the bounded scroll box *and* keep content sizing for a near-empty bag. Worth a quick browser check; if `max-height` alone works, drop the `height` and the always-empty lower half goes away.

Also note the landscape `height` uses raw `100vh` while the bank-docked rule next door uses `calc(var(--app-vh, 100vh) * 0.72 / var(--window-scale) - 24px)`. Under a `--window-scale` above 1 the 360px cap is applied pre-scale, so the painted window can exceed its intended share of the viewport. Minor, and only at >= 1800px, but it is the kind of drift that is cheaper to align now.

## Verdict
Requesting changes, mainly for finding 1 (a real a11y regression on the default layout) and finding 2 (the screenshots and summary do not describe what ships). Finding 3 is a suggestion. The CSS scoping work itself is solid and the tests around it are good pins.

jgyy:
Confirmed the shard-2 failure (`tests/chronomancy.test.ts` > Temporal Barrier > shields self or ally..., expected 184 got 194) is a pre-existing, unrelated flake, not a regression from this PR: this PR only touches `src/styles/components.css`, `src/styles/hud.mobile.css`, `src/ui/bags_window.ts`, and the bags/client-shell tests. `tests/chronomancy.test.ts` passes cleanly (12/12) run in isolation against a fresh `release/v0.28.0` checkout, so this looks like cross-file test-isolation pollution in the full shard run rather than anything this diff introduced. Re-running the shard.

jgyy:
Follow-up on the shard-2 flake: ran the full test suite (unsharded, all 1326 files) against this branch locally. `tests/chronomancy.test.ts` passed with zero failures. The only failures in the full local run were unrelated timeouts (vale_cup_online, server characterization) caused by local resource contention from running the whole suite at once, not sharded the way CI does it. This confirms the shard-2 CI failure on `tests/chronomancy.test.ts` is an environment-timing flake unrelated to this PR's diff, not a real regression. Safe to re-run/ignore if it recurs.

jgyy:
Related: #2194

Rubsey:
@jgyy Can you fix the screenshots on this

jgyy:
the screenshot in the files change, but i close this one instead cause the landscape for bag still feel off will redesign.

PR #2104, opened 2026-07-18
fix(chat): keep whispering the same player after an explicit /w Name (hotfix)

ryan-foo:
## Hotfix: whisper channel does not persist

Targets `main` for immediate deployment (the same fix also lands for the release line in #2098).

## The bug

After whispering a player (`/w Poor test`), the next plain line reverted to `/say` instead of continuing the whisper. The chat input's sticky send-target only stuck a `/r` reply, never an explicit `/w Name`, because sticking a bare whisper would compose the next line as `/r`, which the server routes to whoever last whispered YOU, not the person you whispered. So there was no way to remember which player.

## The fix

The chat controller now tracks the specific whisper target name:
- `/w Name ...` records `Name` as the sticky whisper target; the next plain line composes as `/w Name ...`.
- A `/r` reply still sticks as reply-to-last-whisperer but drops the specific name; any standing-channel send clears it.
- The input placeholder names the target and tints to the whisper color.

Two pure helpers (`whisperTargetName`, `composeWhisperTo`) in the DOM-free `chat_channels.ts` carry the logic. No new i18n keys (reuses the existing `sendingTo` placeholder).

## Testing

Test-first (all fail on the pre-fix code):
- `tests/chat_channels.test.ts`: unit tests for both helpers.
- `tests/chat_window_controller.test.ts`: reproduces the reported bug end to end, plus target-drop-on-standing-channel and `/r`-stays-a-reply.

Green locally: both suites, `tsc`, biome. The chat files are byte-identical on `main` and `release/v0.28.0`, so this cherry-picks cleanly.

Base: `main`. Companion release-line PR: #2098.

### DIFF
--- a/src/ui/hud/chat/chat_channels.ts
+++ b/src/ui/hud/chat/chat_channels.ts
@@ -163,6 +163,27 @@ export function composeWhisperReply(typed: string): string {
   return `/r ${text}`;
 }
 
+// The specific target of an explicit whisper command (/w, /whisper, /t, /tell): the
+// first whitespace-delimited token after the command. This is what lets the chat input
+// KEEP whispering the same player after you whisper them, instead of snapping back to
+// the previous standing channel (the reported bug). Multi-word names resolve server-side
+// (longest-online-name match), so the first token is what the client re-sends. A `/r`
+// reply returns null (its target is the last whisperer, known only to the server), as
+// does any non-whisper line.
+export function whisperTargetName(line: string): string | null {
+  const m = /^\/(?:w|whisper|t|tell)\s+(\S+)/i.exec(line.trim());
+  return m ? m[1] : null;
+}
+
+// Compose the text sent for a plain line typed while a specific whisper target is sticky
+// (after an explicit /w to that player): keep whispering them. An explicit slash command
+// the player typed still wins, mirroring composeChatLine / composeWhisperReply.
+export function composeWhisperTo(name: string, typed: string): string {
+  const text = typed.trim();
+  if (!text || text.startsWith('/')) return text;
+  return `/w ${name} ${text}`;
+}
+
 // The standing channel the actually-sent line reached, used to update the sticky
 // "last used" send channel so the next opened input (on the All tab) defaults
 // there. Plain text (no leading slash) went to `say`. An explicit slash command
--- a/src/ui/hud/chat/chat_window_controller.ts
+++ b/src/ui/hud/chat/chat_window_controller.ts
@@ -13,6 +13,7 @@ import {
   chatOpenTabLabelKey,
   composeChatLine,
   composeWhisperReply,
+  composeWhisperTo,
   isChatOpenTab,
   isChatTabChannel,
   parseChatTabs,
@@ -20,6 +21,7 @@ import {
   serializeChatTabs,
   WHISPER_TAB,
   WHISPER_TAB_LABEL_KEY,
+  whisperTargetName,
 } from './chat_channels';
 
 const CHAT_TABS_KEY = 'woc_chat_tabs';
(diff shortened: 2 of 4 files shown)


ryan-foo:
@Rubsey could you take a look when you have a moment? This is a small chat hotfix targeting `main` for a fast deploy: whispers now persist (after `/w Name`, the next plain line keeps whispering that player instead of reverting to /say). Clean cherry-pick, tests pass on both bases, tsc and biome clean. Companion release-line PR is #2098. Thanks!

EnriqueGF (changes requested):
Reviewed the full diff, the chat command parsers it has to agree with (`src/sim/social/chat.ts` `resolveWhisperTarget` / `whisperMessageForName`, `server/game.ts:1063`), and the client call site (`src/main.ts` `noteSentChannel`).

Ran locally: `npx tsc --noEmit` (clean), `tests/chat_channels.test.ts`, `tests/chat_window_controller.test.ts`, `tests/architecture.test.ts` (all green), `tests/localization_fixes.test.ts` (green after `npm run i18n:gen`). `tests/chat.test.ts` has two failures (`/played`, `/playtime`) that are 5s timeouts unrelated to this diff. No i18n keys added, no new locale rows, no sim/server change, so the invariants in play here are just the seam and the UI purity ones, and those hold: the logic landed in the DOM-free `chat_channels.ts` with unit tests, which is the right shape.

The bug and the direction are right. Two things I would fix before merging.

**Blocking: multi-word character names lose their surname, and the message can reach the wrong player.**

`validCharNameShape` in `server/auth.ts:155` is `/^[A-Za-z][A-Za-z' -]{1,15}$/`, so a space is legal in a character name ("Van Cleef"), and the sim resolver is built for that: `resolveWhisperTarget` prefix-matches the whole rest of the line against every online name and takes the longest match. `whisperTargetName` only captures the first `\S+` token, so after `/w Van Cleef hey` the sticky target becomes `Van`, and the next plain line composes as `/w Van you there?`.

Verified with a scratch test against the real helpers:

```ts
whisperTargetName('/w Van Cleef hey')            // 'Van'
composeWhisperTo('Van', 'you there?')            // '/w Van you there?'
```

Two outcomes, both bad. If nobody is named "Van", the server answers "There is no player named 'Van' online." and the message is dropped, every line, until the player notices. If someone IS named "Van", `whisperMessageForName('Van you there?', 'Van')` matches and `Van Cleef` does not, so the whisper intended for one player is delivered to a different one. That is a private message going to the wrong recipient, which is worse than the bug being fixed.

**Also worth fixing: the sticky is set even when the whisper never landed.**

`noteSentChannel` runs on the composed text unconditionally (`src/main.ts:1197-1202`), with no knowledge of whether the whisper resolved. So `/w Nobodyy hi` (typo, offline player, or a self-whisper) now leaves the input tinted whisper-purple and captioned "Sending to Nobodyy", and every following plain line is silently rejected by the server. Before this change those lines at least went to `/say`.

**One suggestion that fixes both at once.** The client already has the authoritative, fully resolved recipient: the sender's own whisper echo carries `to: target.name` (emitted in `src/sim/social/chat.ts`, consumed in `src/ui/hud.ts` in the `case 'whisper'` arm that renders the "To {name}" line). Setting `stickyWhisperName` from that event instead of from the typed line gives you the complete multi-word name, the server's capitalization, and only sticks when the whisper actually reached someone. `whisperTargetName` would then only need to answer "was this a whisper", not "to whom".

Minor, non-blocking: `composeWhisperTo` does not exempt a leading `!`, so a community command typed while a whisper target is sticky becomes `/w Name !lfg ...` and skips the server relay gate that `composeChatLine` deliberately preserves. `composeWhisperReply` has the same gap today, so this is consistent rather than a regression, but it is now reachable from a lot more states.

Test coverage is otherwise good and genuinely test-first. Adding a two-word-name case to `tests/chat_channels.test.ts` would have caught the blocking one.

PR #2105, opened 2026-07-18
fix(sim): free the Abandoned Crypt door from its own mine-mound collider

jgyy:
## Summary

Players reported being stuck dead outside the Abandoned Crypt: releasing spirit worked, but walking back to the door to corpse-run did nothing ("cant even enter the crypt when dead now").

```
[16:17] Robinn has left the party.
[16:17] [Party] Newkali: i cant repsawn
[16:17] You can't do that while dead.
[16:18] [Party] Newkali: cant even enter the crypt when dead now
[16:18] [Party] Kyubi [Veteran]: me2
```

A released ghost can only ever re-enter its instance through the proximity walk-in trigger (`updateDoorTriggers`, 2.0yd radius): `interact()` unconditionally refuses a dead player (by design — see `tests/ghost_dead_gate.test.ts`), so the door-click path a living player uses never reaches `enterDungeon` for a ghost.

Root cause: the Abandoned Crypt entrance visually reuses the generic "mine entrance" prop (rock mound + timber portal), and shares its `(x, z)` exactly with the dungeon door's own trigger point. The mound's collider approximates the boulder pile as a single circle, offset 3.4yd behind the portal with a 5yd radius — which bleeds **1.6yd forward past the door itself**. Living players never noticed because clicking the door works from a few yards back and is never blocked while alive. A ghost, though, depends solely on getting within the door's 2.0yd walk trigger, and the mound made that physically impossible: any approach resolved to ~2.1yd from the door, just outside the trigger, so ghosts were stranded outside forever.

```mermaid
flowchart TD
    A[Player dies inside Abandoned Crypt] --> B[Releases spirit -> ghost]
    B --> C[Ghost walks back to the door]
    C --> D{interact / click door?}
    D -- "dead: always refused" --> E[No path in via click]
    C --> F{Within 2.0yd walk trigger?}
    F -- "mound collider bled 1.6yd\npast the door (BEFORE fix)" --> G[Push-out lands ~2.1yd away\nNEVER triggers]
    F -- "moundOffset widened to 8yd\n(AFTER fix)" --> H[Door tile itself walkable\nTrigger fires, ghost resurrects]
```

## Fix

Added an optional per-entry `moundOffset` to the mine prop shape (`src/sim/types.ts`) and set the Abandoned Crypt entry's to 8yd (`src/sim/content/zone3.ts`), clearing the door tile and its trigger radius. `src/sim/colliders.ts` falls back to the previous 3.4yd default when the field is absent, so every other mine entrance (Deeprock Burrows, the zone1 mine) is unchanged. Purely a collision-geometry fix: no rendered visual changes, so no before/after screenshots apply (the rock mound mesh itself is untouched).

## Test plan

- [x] New repro test at the door's exact world position through the full collider stack (`tests/fixes.test.ts` — "keeps the Abandoned Crypt door clear of the mine-mound collider"), fails without the fix.
- [x] New end-to-end test: a ghost corpse-runs into `nythraxis_crypt` via a collision-resolved approach point instead of a bare teleport (`tests/dungeons.test.ts` — "pulls a ghost back into the Abandoned Crypt from a realistic (collision-resolved) approach"), fails without the fix.
- [x] `npx vitest run tests/dungeons.test.ts tests/fixes.test.ts tests/door_portal.test.ts tests/map_dungeon_portals.test.ts tests/ghost_dead_gate.test.ts tests/nythraxis_raid.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts` — all green.
- [x] `npx tsc --noEmit` — clean.
- [x] `npx @biomejs/biome check` on changed files — no errors/format diffs (pre-existing whole-repo warnings untouched).
- [x] Full `npx vitest run` suite — green.

### DIFF
--- a/src/sim/colliders.ts
+++ b/src/sim/colliders.ts
@@ -126,7 +126,7 @@ function staticWorldColliders(seed: number): Collider[] {
 
   // mines: mound behind the timber portal
   for (const m of PROPS.mines) {
-    const mound = rotY(0, -3.4, m.rot);
+    const mound = rotY(0, -(m.moundOffset ?? 3.4), m.rot);
     const x = m.x + mound.x,
       z = m.z + mound.z;
     out.push({ type: 'circle', x, z, r: 5, cameraTopY: topY(seed, x, z, 5.2), camGhost: true });
--- a/src/sim/colliders.ts
+++ b/src/sim/colliders.ts
@@ -126,10 +126,11 @@ function staticWorldColliders(seed: number): Collider[] {
 
   // mines: mound behind the timber portal
   for (const m of PROPS.mines) {
+    const r = m.moundRadius ?? 5;
     const mound = rotY(0, -(m.moundOffset ?? 3.4), m.rot);
     const x = m.x + mound.x,
       z = m.z + mound.z;
-    out.push({ type: 'circle', x, z, r: 5, cameraTopY: topY(seed, x, z, 5.2), camGhost: true });
+    out.push({ type: 'circle', x, z, r, cameraTopY: topY(seed, x, z, r + 0.2), camGhost: true });
   }
 
   // Dock decks are raised walkable ground in world.ts; only the hut blocks.
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -3302,8 +3302,11 @@ export const ZONE3_PROPS: ZonePropsDef = {
     // trigger point, so the mound's collider needs a bigger backward offset
     // (moundOffset) than the generic mine default or it swallows the door
     // itself, stranding any ghost that can only walk-trigger it (issue: dead
-    // players unable to enter/corpse-run the crypt).
-    { x: -152, z: 610, rot: Math.PI / 2, moundOffset: 8 },
+    // players unable to enter/corpse-run the crypt). moundRadius is also
+    // shrunk from the generic default so the circle hugs this entry's own,
+    // smaller-footprint rock pile (src/render/props.ts) instead of bleeding
+    // onto open ground behind it and off the flanking boulders themselves.
+    { x: -152, z: 610, rot: Math.PI / 2, moundOffset: 4, moundRadius: 3.3 },
   ],
   docks: [],
   tents: [
--- a/src/sim/content/zone3.ts
+++ b/src/sim/content/zone3.ts
@@ -3298,7 +3298,12 @@ export const ZONE3_PROPS: ZonePropsDef = {
   ],
   mines: [
     { x: 88, z: 612, rot: -2.0 }, // Deeprock Burrows
-    { x: -152, z: 610, rot: Math.PI / 2 }, // Abandoned crypt entrance
+    // Abandoned crypt entrance: shares its (x, z) with the dungeon door's own
+    // trigger point, so the mound's collider needs a bigger backward offset
+    // (moundOffset) than the generic mine default or it swallows the door
+    // itself, stranding any ghost that can only walk-trigger it (issue: dead
+    // players unable to enter/corpse-run the crypt).
+    { x: -152, z: 610, rot: Math.PI / 2, moundOffset: 8 },
   ],
   docks: [],
   tents: [
--- a/src/sim/types.ts
+++ b/src/sim/types.ts
@@ -2342,12 +2342,14 @@ export interface ZonePropsDef {
   buildings: BuildingDef[];
   wells: { x: number; z: number; r: number }[];
   stalls: { x: number; z: number; rot: number; r: number; smithy?: true }[];
-  // moundOffset overrides the collider's default backward offset (colliders.ts)
-  // for the rock mound behind the timber portal, for a mine entry whose (x, z)
-  // doubles as a real interactable's trigger point (the Abandoned Crypt door):
-  // the default offset lets the mound's collision circle bleed into the
-  // approach side and swallow the point itself.
-  mines: { x: number; z: number; rot: number; moundOffset?: number }[];
+  // moundOffset/moundRadius override the collider's default backward offset
+  // and radius (colliders.ts) for the rock mound behind the timber portal,
+  // for a mine entry whose (x, z) doubles as a real interactable's trigger
+  // point (the Abandoned Crypt door): the defaults let the mound's collision
+  // circle bleed into the approach side and swallow the point itself. Keep
+  // both close to the entry's actual rendered mound extent (src/render/props.ts)
+  // so the collider does not drift onto open, visually clear ground.
+  mines: { x: number; z: number; rot: number; moundOffset?: number; moundRadius?: number }[];
   docks: {
     x: number;
     z: number;
--- a/src/sim/types.ts
+++ b/src/sim/types.ts
@@ -2342,7 +2342,12 @@ export interface ZonePropsDef {
   buildings: BuildingDef[];
   wells: { x: number; z: number; r: number }[];
   stalls: { x: number; z: number; rot: number; r: number; smithy?: true }[];
-  mines: { x: number; z: number; rot: number }[];
+  // moundOffset overrides the collider's default backward offset (colliders.ts)
+  // for the rock mound behind the timber portal, for a mine entry whose (x, z)
+  // doubles as a real interactable's trigger point (the Abandoned Crypt door):
+  // the default offset lets the mound's collision circle bleed into the
+  // approach side and swallow the point itself.
+  mines: { x: number; z: number; rot: number; moundOffset?: number }[];
   docks: {
     x: number;
     z: number;
--- a/tests/dungeons.test.ts
+++ b/tests/dungeons.test.ts
@@ -4,6 +4,7 @@
 // the party-shared instance, the claim -> free empty-reset, and the raid-lockout gate.
 
 import { describe, expect, it } from 'vitest';
+import { resolvePosition } from '../src/sim/colliders';
 import { HEROIC_DUNGEON_TUNING, HEROIC_MARK_ITEM_ID } from '../src/sim/content/dungeon_difficulty';
 import { HEROIC_BOSS_LOOT } from '../src/sim/content/heroic_loot';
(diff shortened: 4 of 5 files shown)


Rubsey (commented):
The root cause writeup checks out against the real code, and this fixes a genuine stranded-ghost bug. The mound circle (r 5, centered 3.4yd behind the portal, src/sim/colliders.ts:129) did bleed 1.6yd past the door, the walk-in trigger needs 2.0yd (src/sim/instances/dungeons.ts:41), and interact() refuses dead players (src/sim/interaction.ts:99), so a released spirit truly had no path back in. I re-verified the repro claim locally: both new tests fail with the base geometry and pass with the fix, and the full-collider-stack scan in tests/fixes.test.ts is a decisive assertion. Leaving every other mine on the 3.4 default is right; neither Deeprock (88, 612) nor the zone1 mine (-88, -68) sits near a dungeon door (doors are at (80, 90), (45, 515), (0, 858), (-152, 610)).

One finding:

- should-fix (cosmetic, not blocking): the collider and the visible rock pile now disagree at the crypt. The renderer's crypt-specific mound (src/render/props.ts:1119) still spans roughly local z 0 to -6, while the collision circle now covers local z -3 to -13. I probed isBlocked around the entrance to confirm both halves: a player can skirt the 2.0yd door trigger and stand inside the flanking boulders (world (-154.5, 607.5), inside the rock at local (2.65, -2.3): blocked before, open now), and there is now an invisible wall over about 7yd of open, previously walkable ground behind the pile (world x about -158 to -165 at z 610; nothing lives there, I checked zone3 mob camps, quest objects, and graveyards, so it is purely cosmetic). Consider matching the circle to the visible pile instead of pushing the full r 5 circle 8yd back: a per-entry offset plus radius (something like offset 4.5, r 3) clears the door tile and the trigger just as well while keeping collision under the big rocks and off the open ground. Also worth noting the colliders.ts header ("what you see is what you collide with", src/sim/colliders.ts:33) is now slightly untrue for this one entry.

One observation, no action needed: living players get a small behavior change too. The crypt door now walk-triggers like every other dungeon door, where the old collider accidentally made it click-only by keeping everyone outside 2.0yd. That is a consistency win.

EnriqueGF:
Thanks for chasing this one down :) the stranded-ghost diagnosis is correct and well evidenced, and the repro tests are genuinely decisive. My verdict is still changes requested, because the fix trades a gameplay bug for a collider/art desync that the diff does not close.

Verified first: base branch is `release/v0.28.0` and the branch already contains its tip, so the base gate passes. I ran `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/parity`, and `tests/fixes.test.ts tests/dungeons.test.ts tests/door_portal.test.ts tests/ghost_dead_gate.test.ts`: all green. I also reverted the fix locally and confirmed both new tests fail on the base geometry (`isBlocked` true at the door, and the resolved approach landing at 2.0999 yd), exactly as described.

## should-fix 1: the renderer never reads `moundOffset`, so mesh and collider now disagree

This is the same area Rubsey raised, but the root is more concrete than a cosmetic mismatch, and it is still open. `src/render/props.ts:1182-1183` hardcodes the old constant:

```ts
// mound circle behind the portal - same offset/radius as the collider
const mx = m.x - 3.4 * Math.sin(m.rot),
  mz = m.z - 3.4 * Math.cos(m.rot);
```

That comment is now false for this entry, and nothing in the diff touches it. `moundOffset` is consumed only by `src/sim/colliders.ts:129`, which also makes the header contract at `src/sim/colliders.ts:31-33` untrue ("the renderer builds its meshes from the same defs, so what you see is what you collide with"). This is the first field to break that guarantee.

I probed the real collider stack at seed 20061, `r=0.4`, along z=610, before and after:

```
before: blocked x -160 .. -151   (door at -152 blocked)
after:  blocked x -165 .. -155   (door at -152 open)
```

The circle was translated wholesale by 4.6yd rather than reshaped. Consequences, cross-checked against the authored rock anchors in `src/render/props.ts:1119-1140` (the `abandonedCrypt` branch spans local z about -0.04 to -6.25):

- roughly a 3yd deep band of drawn rock at the front of the pile is now walk-through, including the two forward boulders at local z -1.25 and -1.2
- roughly 6.7yd of blocking circle now sits over bare terrain behind the pile (world x about -158 to -165) with nothing rendered there
- the camera-occlusion footprint (`registerHideable`/`circleFootprint`, still at 3.4) and the collider's `camGhost`/`cameraTopY` circle are now 4.6yd apart, so prop fade and camera collision disagree for this entrance

Suggested shape: keep the door tile clear by shrinking or reshaping the mound collider so it still hugs the drawn rubble (a per-entry radius alongside the offset, which matches how `wells` and `stalls` already carry `r` in their defs), and have `props.ts` derive its anchor from `m.moundOffset ?? 3.4` in the same change so the single-source-of-truth claim stays true. That would also let the coordinate heuristic at `props.ts` (`m.x < -140 && m.z > 590 && m.z < 630`) go away.

## should-fix 2: only the override arm of `?? 3.4` is pinned

`src/sim/colliders.ts:129` has two arms and the tests only exercise the crypt entry. Nothing asserts that Deeprock Burrows `(88, 612)` and the zone1 mine `(-88, -68)` still get the 3.4 mound, so an edit to the fallback (or to `rotY`) regresses silently while the whole suite stays green. A single `isBlocked` pin on one default mine mound alongside the new crypt assertions would close it.

## should-fix 3: the new test has no upper bound

`tests/fixes.test.ts` asserts only that the door tile is reachable, which would pass for an arbitrarily large backward offset. Please add a companion assertion that the mound's forward rock anchors (local z about -1.2) are still blocked, so a future offset bump cannot silently re-open walk-through rock. Worth noting the current margin against a partial fix is one grid step (2.25 vs the 2.0 threshold).

## Notes, no action required

- The mine circles also feed spell line of sight via `sightBlockedAt` in `colliders.ts`, so moving the circle moves the LOS blocker too. I confirmed the impact is nil today (nearest camp `old_cragmaw` is about 78yd away, and no NPC, quest object, or graveyard is within 60yd), but it is worth accepting consciously.
- One reviewer pass saw a single non-reproducible failure of both new tests (16 later runs, including concurrent-load runs, were clean). The collider grid is cached per `(WorldContent, seed)` in a `WeakMap` and several suites call `setActiveWorldContent`, so if it ever reproduces it is a shared-active-world-content leak rather than a defect in this diff. One full `npm test` before merge would settle it.
- `tests/dungeons.test.ts`: the title says "realistic (collision-resolved) approach", but the body teleports onto the door tile after asserting `resolvePosition` is a no-op there. The protection is real, so this is naming only.
- Agreed that no screenshots are needed: no rendered geometry changes. The flip side is that the art is exactly what the collider no longer matches, which is what should-fix 1 asks you to reconcile.

Nice find on the underlying bug, and the walk-trigger consistency win for living players is a good side effect. Once the collider hugs the visible mound again and the two default-arm and upper-bound assertions land, this should be ready.

jgyy:
Thanks for the thorough re-verification, and good catch on the collider/visual mismatch.

Fixed in 40a3274: added an optional `moundRadius` alongside `moundOffset` on the mine-prop shape (`src/sim/types.ts`, `src/sim/colliders.ts`), defaulting to the existing 5yd so every other mine entrance is unchanged, and tightened the Abandoned Crypt entry to `moundOffset: 4, moundRadius: 3.3` (`src/sim/content/zone3.ts`). That circle now hugs the rendered pile (near edge ~0.7yd from the door, far edge ~7.3yd, versus the visible rocks' own ~0.1 to 6.15yd span) instead of the old 3yd-to-13yd sprawl, while still clearing the door's 2.0yd walk-in trigger with margin.

Added a regression test (`tests/fixes.test.ts` — "keeps the Abandoned Crypt mound collider matched to its visible rock pile") pinning both halves of your finding: the flanking boulder at world (-154.5, 607.5) stays blocked, and open ground well behind the pile (door.x - 9) is no longer swallowed. The existing door-clear test still passes unchanged.

Rubsey (commented):
Re-verified at 40a3274e2: the should-fix from my review is resolved. I re-ran the walkability probe around the entrance and the collider now hugs the rendered pile (roughly world x -159 to -153, z 607 to 613), the door tile stays clear, the point inside the flanking boulder at (-154.5, 607.5) is solid again, and the open ground behind the pile is walkable. Nice touch pinning both of those exact points in the new test, and keeping the other mines byte-identical via r + 0.2 for cameraTopY. tests/fixes.test.ts and tests/dungeons.test.ts pass locally at this head. Good to go from my side once CI finishes.

jgyy:
Closed the remaining should-fix items from the last pass:

- **should-fix 1 (renderer/collider drift):** `src/render/props.ts` was still hardcoding the old 3.4yd/5yd constants for the mound footprint circle even after the collider gained per-entry `moundOffset`/`moundRadius`. Now derives both from `m.moundOffset ?? 3.4` / `m.moundRadius ?? 5`, matching the sim-side `camGhost`/`cameraTopY` circle in `src/sim/colliders.ts`, which already read those fields correctly. Single source of truth restored; the stale comment is gone.
- **should-fix 2 (unpinned default arm):** added a test pinning the `?? 3.4` / `?? 5` fallback against Deeprock Burrows `(88, 612)` and the zone1 mine `(-88, -68)`: mound center blocked, well past the 5yd radius open. An edit to either default now fails loudly instead of only affecting the crypt's overridden entry.
- **should-fix 3 (no upper bound):** added an assertion to the door-clear test that the forward rock anchors nearest the portal (local `(1.75, -1.2)` r 1.1 and `(-1.7, -1.25)` r 1.15, per `src/render/props.ts`) stay blocked, so a future offset bump can't silently turn the visible rubble into walk-through air.

All in 40a3274 -> f28bb726a. Ran the full named regression set plus `tests/architecture.test.ts`/`tests/localization_fixes.test.ts` (green), `tsc --noEmit` (clean), and `biome check` on the changed files (no new errors/format diffs, only pre-existing whole-file warnings untouched).

PR #2106, opened 2026-07-18
feat(net): reconnect attempt/countdown and slow-connection loading hint

jgyy:
## Summary

Tested the client under network throttling (Chrome DevTools "Slow 3G" + 4x CPU throttle
against the production build) and found two real gaps in the network-resilience path, both
now fixed:

- **Reconnect overlay was a static string.** When the WS drops (exactly what packet loss /
  throttling triggers), `src/ui/reconnect_overlay.ts` showed a frozen "Reconnecting..."
  message for up to 15s per retry, up to 40 attempts (up to ~8 minutes total, see
  `src/net/online.ts`), with zero feedback on whether it was stuck or working. It now shows
  a live attempt count and a per-second retry countdown.
- **Loading screen gave no reason for a stall.** The existing progress bar + rotating tips
  (`src/main.ts`) look identical whether progress is moving or has stalled on a slow
  connection. A new "This is taking longer than usual. Check your internet connection." hint
  now appears once progress has been quiet for 6+ seconds, and clears the moment progress
  resumes.

Both changes are informational/cosmetic only (no gameplay effect, server stays authoritative)
and land as small, tested sibling modules per the repo's module-first convention.

## Review fixes

- **Blocking: `/play` broke on world entry.** `#ls-slow-hint` only existed in `index.html`;
  `play.html` lacked it, and the lookup used the unguarded `$` helper, so every call threw on
  `/play`, ultimately leaving `#loading-screen` permanently visible and movement/input
  permanently suspended. Fixed by adding the element to `play.html` and switching the lookup
  to a null-safe `document.querySelector('#ls-slow-hint')?.classList.toggle(...)`. Added a
  `tests/client_shell.test.ts` assertion that both entries carry the loading-screen element
  set, including `#ls-slow-hint`.
- **Design concern: the hint blamed the network for CPU-bound work.** `stopSlowConnectionWatch()`
  is now called as soon as `assetsReady()` resolves, so the watch only spans the genuinely
  network-bound asset-fetch phase, not the CPU-bound scene-construction phase after it.
- Non-blocking items folded in: moved `reconnect_status.ts` under `src/ui/` (the `net/`
  import direction fix), and the orphaned `loading.reconnecting` i18n key was addressed.

## Screenshots

The original before/after screenshots (`docs/screenshots/network-throttling-improvements/`)
already documented the feature correctly on `index.html` and did not need to change. Added
one new screenshot, `after-play-slow-hint.png`, showing the previously-broken `/play.html`
entry now displaying the hint without throwing, proving the fix.

## Type of change

- [x] Bug fix

## How was this tested?

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/i18n_completeness.test.ts tests/client_shell.test.ts tests/backoff.test.ts tests/net_online_visibility_reconnect.test.ts tests/linkdead.test.ts tests/reconnect_status.test.ts tests/loading_slow_hint.test.ts`: 432 passed.
- `npm run i18n:gen`: regen clean.
- `npx @biomejs/biome check --write` on changed files: clean.
- Full pre-push QA gate: green.
- Manual: confirmed `#ls-slow-hint` toggles without throwing on both `index.html` and `play.html`.


EnriqueGF (changes requested):
Thanks for this: the underlying problem is real and worth fixing. A frozen "Reconnecting..." across up to 40 backoff retries genuinely is indistinguishable from a hung client, and the pure-module-plus-test shape (`reconnect_status.ts`, `loading_slow_hint.ts`) is exactly the right structure. The `onConnectionLost` signature change is clean, and I confirmed it is a `ClientWorld` field rather than an `IWorld` member, so your "no parity pin needed" call is correct.

One blocking issue though, plus a design concern.

## What I reviewed and ran

- Full diff against `release/v0.28.0`.
- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/i18n_completeness.test.ts`, `tests/client_shell.test.ts`, `tests/backoff.test.ts`, `tests/net_online_visibility_reconnect.test.ts`, `tests/linkdead.test.ts`, plus both new test files: all green (385 + 52 passing).
- `npm run i18n:gen` then `tests/localization_fixes.test.ts` (S3 guard): green, and the regen leaves the tree clean.
- Dash/emoji scan of the diff: clean. No secrets, no `ALLOW_DEV_COMMANDS`. Screenshot sizes are in line with existing `docs/screenshots` directories.

## Blocking: `/play` breaks on world entry

`index.html` gained `<div id="ls-slow-hint">`, but **`play.html` did not**. Both entries load `src/main.ts`, and `src/CLAUDE.md` calls this out specifically:

> `index.html` AND `play.html` both load `src/main.ts`. Guard index-only DOM lookups with `?.`: `/play` lacks the marketing shell, and an unguarded lookup throws there (a bug class that has shipped before).

`setSlowConnectionHintVisible` uses the unguarded helper `$` (`src/main.ts:303`), which is `document.querySelector(sel) as T`: non-null in the types, `null` at runtime. On `/play` every call throws `TypeError`. Note the existing tip code right next door at `src/main.ts:837` uses a null-guarded `document.querySelector<HTMLElement>('#ls-tip')` for exactly this reason.

The consequences are not cosmetic. Tracing each call site:

1. The 1s interval throws every tick for the whole load (uncaught, interval keeps running).
2. `setLoadingProgress` throws on its last line. This one is swallowed, because `preload.ts:27` wraps the progress callback in `.catch(() => undefined)`.
3. **The damaging one:** `hideLoadingScreen` calls `stopSlowConnectionWatch()`, which throws *before* `loadingHideTimer = window.setTimeout(...)` is scheduled. So `.visible` is never removed. `#loading-screen.visible.fade` is `display:flex; opacity:0` on a `position:absolute; inset:0; z-index:150` element, so an invisible full-screen overlay is left permanently on top of the game, swallowing pointer input.
4. That throw happens inside the "cut to the game" rAF callback (`src/main.ts:3224`), which aborts the rest of that callback: `gameInputReady = true` (line 3229) never runs, so `input.setSuspendMovement(...)` at line 2825 keeps movement suspended forever, and the perf reporter never starts.

Net effect on `/play`: the player loads in and cannot move or click. I reproduced 3 and 4 in jsdom against the `play.html` loading-screen markup: `hideLoadingScreen()` throws, the fade timer is never scheduled, and `.visible` stays set.

Fix is small: add the element to `play.html` as well, and make the lookup null-safe (`document.querySelector('#ls-slow-hint')?.classList.toggle(...)`) so this cannot recur. Worth a `tests/client_shell.test.ts` assertion that both entries carry the loading-screen element set, since that is the guard that would have caught it.

## Design concern: the hint blames the network for CPU work

`lastLoadingProgressAt` has exactly two writers: `startSlowConnectionWatch` and `setLoadingProgress`. `setLoadingProgress` is only ever fed by `assetsReady` (`src/main.ts:973`). But the watch stays armed until `hideLoadingScreen`, and the whole phase *after* assets finish has no heartbeat at all: `mountGameUi()` plus `translatePage()`, the synchronous `Renderer` and `Hud` construction, `prewarmInitialScene()`, then two rAFs to the first frame.

That stretch is CPU-bound scene building, not network. On a low-end phone, or under exactly the 4x CPU throttle you tested with, it plausibly exceeds the 6s threshold, and the player gets "This is taking longer than usual. Check your internet connection." while their connection is fine. Misleading advice during the most common slow-entry case.

Simplest fix is to stop the watch once `assetsReady` resolves, since asset fetching is the only network-bound phase it can actually speak to. Alternatively give the "Entering the world" phase its own heartbeat, or reword the string so it does not assert a network cause.

## Non-blocking

- **Layer direction.** `src/ui/reconnect_overlay.ts` imports `../net/reconnect_status`. `src/CLAUDE.md` states `render/`, `ui/`, `game/` may import `IWorld` plus their own area, "**not** `net/`". The `forbiddenUiCoreImport` guard in `tests/architecture.test.ts:538` classifies exactly this as forbidden, but only scans registered `UI_PURE_CORES`, so this import slips past. Since the module is consumed solely by the overlay, moving it to `src/ui/` resolves the direction cleanly.
- **Pure-core naming.** Neither new module uses the `*_view` / `*_core` suffix, so both escape the completeness sweep and the purity/determinism scans by name. `src/ui/CLAUDE.md` flags that the convention name is what makes a forgotten registration fail the guard. Minor, and there is precedent (`loading_tips.ts` is the same), but `loading_slow_hint_core.ts` registered in `UI_PURE_CORES` would be tidier.
- **Orphaned key.** `loading.reconnecting` now has no `src/` reference; `loading.reconnectingAttempt` fully replaced it. It still carries translations in all 21 locales. Either keep it deliberately as the zero-data fallback or drop it.

Happy to re-review once the `/play` element and the null-guard are in. Everything else here is solid work.

EnriqueGF (commented):
Follow-up after running the repo's `frontend-seam-reviewer` and `cross-platform-sync` reviewers over the diff. Both confirmed no parity drift (`onConnectionLost` is correctly outside `IWorld`, all callers migrated, `tsc` clean, wire protocol and `SimEvent` surfaces untouched). Three additional items I verified in the source myself, one of which I would treat as blocking alongside the `/play` bug.

## Blocking-ish: `onConnectionLost` fires before the retry timer is armed

`src/net/online.ts:1562` calls the callback, but `this.reconnectTimer = window.setTimeout(...)` is only assigned at `:1565`. So if the callback throws, `reconnectTimer` is never set and `reconnectAttempts` has already been incremented: auto-reconnect is dead for the rest of the session, with no retry, no `onDisconnect`, and no fatal overlay. The player just sits on a reconnect overlay forever.

The ordering predates this PR, but the PR materially widens the throw surface. Previously the callback was `showReconnectOverlay()` guarded by an early `if (document.getElementById(OVERLAY_ID)) return;`, so a repeat drop did essentially nothing. It is now an unconditional DOM create/mutate plus a `window.setInterval` and a `t()` call on every drop.

The `t()` call is the part worth thinking about. `src/ui/CLAUDE.md` states `t()` "hard-fails a pending key on a release build". `loading.reconnectingAttempt` is filled for `en` plus the five M16 non-Latin locales, but the remaining locales are `pending` on this branch. The release-tier gate is what stops that reaching production, so this is not a live hole today, but it means a socket drop can now throw on a path where throwing kills reconnect permanently.

Cheap fix: move `this.onConnectionLost?.(...)` to after the `setTimeout` block. It only reads `delayMs` and `this.reconnectAttempts`, both already computed, so reordering costs nothing and removes the failure mode entirely.

## Stale countdown on the mobile-foreground fast retry

`src/net/online.ts:1499-1505`: when the tab is foregrounded with a reconnect already pending, the handler clears `reconnectTimer` and re-schedules it at `Math.random() * 1000`, but does not re-fire `onConnectionLost`. The overlay keeps counting toward the original `nextRetryAtMs`, so at higher attempt counts it can display "retrying in 40s" while the real retry fires in under a second.

That is the phone-unlock path, which is exactly the case this feature exists to reassure. Firing `onConnectionLost` with the new timestamp there would keep the display honest.

Related and smaller: once the countdown reaches zero the interval keeps repainting "0s" with a stale attempt number until the next drop, which partly reintroduces the "looks hung" symptom. A "retrying now..." key at zero would close it.

## Accessibility of the new hint

`#ls-slow-hint` appears and disappears purely via a `display:none` to `block` class toggle with no `role="status"` / `aria-live="polite"`. A screen-reader user gets no announcement of the one message that explains the stall. Given the WCAG 2.2 AA HUD-chrome contract in `src/ui/CLAUDE.md`, worth adding.

I would not block on the reconnect overlay's own lack of a role, since the pre-existing `fatal-overlay` has none either, so that is consistent precedent rather than a regression.

## Smaller notes

- `#ls-slow-hint` ships with empty inline text, unlike its sibling `#ls-status` which carries inline English (`Entering the world...`) as a fallback. It depends entirely on the async boot `translatePage()` at `src/main.ts:7367`. If that has not landed or throws, the hint reveals as an empty error-colored div, on precisely the slow-network path it targets. Matching the sibling pattern with inline English is the safer default.
- No test covers the overlay's new stateful half: nothing pins that a second `showReconnectOverlay` replaces rather than duplicates the interval, or that `hideReconnectOverlay` clears it. The extracted pure function is well tested; the part that owns the leak risk is not. A `tests/reconnect_overlay.test.ts` with fake timers over show/show/hide would cover it. I did verify by reading that every terminal path in `main.ts` (`:4931`, `:4944`, `:4960`) calls `hideReconnectOverlay` before `fatalOverlay`, so there is no dangling interval today.
- On number formatting: the reviewer flagged `attempt`/`maxAttempts`/`seconds` bypassing `formatNumber`. I do not think that needs changing. They are small unformatted counters where grouping would be wrong, and it matches the existing `loading.worldProgress` `{done}/{total}` precedent in the same namespace, which is what your PR description already argued.

Same conclusion as before: the feature is worth having, the structure is right. The `/play` element plus null-guard is the must-fix, and the callback reordering above is a two-line change I would take while you are in there.

jgyy:
Thanks for the thorough pass, both rounds. Pushed a follow-up commit that addresses everything raised:

**Blocking: `/play` breaks on world entry**
Added `#ls-slow-hint` to `play.html` (it only shipped on `index.html`) and switched `setSlowConnectionHintVisible` from the unguarded `$()` helper to a null-safe `document.querySelector(...)?.classList.toggle(...)`, matching the existing `#ls-tip` pattern next to it. Added a `tests/client_shell.test.ts` pin asserting both entries carry the full loading-screen element set (`#loading-screen`, `#ls-fill`, `#ls-status`, `#ls-tip`, `#ls-slow-hint` with its `role="status" aria-live="polite"` markup), so this class of drift fails the guard next time.

**Design concern: the hint blaming the network for CPU work**
`stopSlowConnectionWatch()` is now called as soon as `assetsReady` resolves in `main.ts`, since that is the only network-bound phase `lastLoadingProgressAt` can speak to. The CPU-bound scene-build stretch after it (`mountGameUi`, `prewarmInitialScene`, the two rAFs to first frame) no longer keeps the watch armed, so a slow device with a fine connection cannot trip the hint.

**Blocking-ish: `onConnectionLost` fires before the retry timer is armed**
Reordered in `src/net/online.ts`: `reconnectTimer` is now assigned before `onConnectionLost` fires, so a throw inside the callback (DOM mutation, `setInterval`, `t()`) can no longer leave `reconnectAttempts` incremented with no retry scheduled.

**Stale countdown on the mobile-foreground fast retry**
The `visibilitychange` fast-retry path now re-fires `onConnectionLost` with the freshly computed short delay, so the overlay stops showing a stale multi-second countdown when the real retry is about to fire in under a second. Also added the "related and smaller" fix: once the countdown reaches 0 the overlay swaps to a distinct `loading.reconnectingNow` message instead of repainting a stale "0s" with the old attempt number.

**Accessibility of the new hint**
`#ls-slow-hint` now carries `role="status" aria-live="polite"` on both entries, plus inline English fallback text matching its `#ls-status` sibling (so a `translatePage()` failure or race never reveals it as an empty error-colored div).

**Layer direction / pure-core naming**
Moved `reconnect_status.ts` to `src/ui/reconnect_status_core.ts` (it's consumed only by the overlay, so `ui/` is the right home; this also closes the `ui`-may-not-import-`net` violation the `forbiddenUiCoreImport` guard couldn't see because the file wasn't registered) and renamed `loading_slow_hint.ts` to `loading_slow_hint_core.ts`. Both are now registered in `UI_PURE_CORES`.

**Orphaned key**
Dropped `loading.reconnecting` from the catalog and every locale overlay (fully superseded by `reconnectingAttempt`/the new `reconnectingNow`).

**No test for the overlay's stateful half**
Added `tests/reconnect_overlay.test.ts` with fake timers: pins that a second `showReconnectOverlay` replaces the message/interval in place rather than duplicating either, that `hideReconnectOverlay` clears the interval, and that the display swaps to the "reconnecting now" message at the zero-countdown boundary.

Left as-is per your call: the `formatNumber` non-blocking note (agreed, unformatted counters are correct here), and the reconnect overlay's own lack of `role`/`aria-live` (consistent with the existing `fatal-overlay` precedent).

Verified: `npx tsc --noEmit` clean, the full targeted set (`reconnect_overlay`, `reconnect_status_core`, `loading_slow_hint_core`, `client_shell`, `architecture`, `localization_fixes`, `net_online_visibility_reconnect`, `backoff`, `linkdead`, `i18n_completeness`) all green, `i18n:gen` regenerated and staged, and the pre-push QA floor (tsc + guard tests + biome + copy scan) passed on push. The full `npm run gate` run in my sandbox surfaces `tests/deeds_window_focus.test.ts` (a `localStorage.clear()` TypeError, environment-only) and two `tests/server/http/{characterization,parity}.test.ts` timeouts under heavy parallel load; all three reproduce identically at the `release/v0.28.0` merge base with none of this branch's changes applied, and the two server tests pass cleanly when run in isolation, so they're pre-existing sandbox flakiness rather than anything introduced here.

Branch is rebased onto the current `release/v0.28.0` tip (merged cleanly, no conflicts).

jgyy:
Rebased this branch onto the current release/v0.28.0 tip to resolve the merge conflict.

Only one real conflict: the generated src/ui/i18n.resolved.generated/pending.ts (both branches added pending rows). Resolved by taking base branch's version then regenerating with npm run i18n:gen, which cleanly reintroduced this branch's own pending rows. No other files conflicted; all touched paths (src/net, src/ui/reconnect_overlay.ts, src/ui/loading_slow_hint.ts, etc.) were disjoint from what moved on release/v0.28.0.

Verified after rebase: npx tsc --noEmit clean, and tests/architecture.test.ts, tests/world_api_parity.test.ts, tests/localization_fixes.test.ts, tests/reconnect_status.test.ts, tests/loading_slow_hint.test.ts, tests/client_shell.test.ts all pass (401 passed, 3 skipped). Force-pushed the rebased branch; PR now reads mergeable.

jgyy:
Thanks for the thorough review. Both blocking items are fixed:

1. **\`/play\` breakage**: added \`#ls-slow-hint\` to \`play.html\` and switched the lookup to a null-safe optional-chain call, so it can never throw on either entry. Added a \`tests/client_shell.test.ts\` assertion pinning both entries carry the loading-screen element set, which would have caught this.
2. **CPU-vs-network heartbeat**: \`stopSlowConnectionWatch()\` now fires as soon as \`assetsReady()\` resolves, so the watch no longer spans the CPU-bound scene-construction phase after assets finish loading.

Also folded in the three non-blocking notes: moved \`reconnect_status.ts\` under \`src/ui/\` for the correct import direction, and addressed the orphaned \`loading.reconnecting\` i18n key.

Verified: \`tsc\` clean, the full named test suite green (432 passed), biome clean, i18n regen clean, full pre-push gate green.

The original screenshots already correctly documented the feature on \`index.html\` and didn't need to change; added one new screenshot (\`after-play-slow-hint.png\`) proving the \`/play.html\` fix specifically, since that was the actual regression.

EnriqueGF (commented):
Reviewed the full diff against `release/v0.28.0`, read the surrounding code paths in `src/main.ts`, `src/net/online.ts` and `src/ui/reconnect_overlay.ts`, and ran the gates locally. Nice change overall: the problem is real, both fixes are informational only, the sim is untouched so determinism is not in play, and the earlier `/play.html` regression is properly fixed and pinned. **I did not find a production bug.** Everything below is convention and test-confidence.

## What I ran

- `npx tsc --noEmit`: clean.
- Guards: `architecture`, `localization_fixes`, `world_api_parity`, `client_shell`: pass. (`localization_fixes` first failed only on a missing `i18n.status.json`; after `npm run i18n:gen` it passes and `git status` is empty, so the generated artifacts in the PR are fresh.)
- Feature + net: `reconnect_overlay`, `reconnect_status_core`, `loading_slow_hint_core`, `backoff`, `net_online_visibility_reconnect`, `reconnect_policy`, `i18n_completeness`, `linkdead`: all green.
- My own jsdom repro confirming the countdown really decrements across ticks and the interval is genuinely dead after `hideReconnectOverlay`. Both hold.
- Traced every arm/disarm path for the slow-hint watch. The `assetsFailed` branch reaches `fatalOverlay` to `hideLoadingScreen` to `stopSlowConnectionWatch` while the screen is still `visible`, so the early return there cannot strand the interval. No leak.
- Copy scan clean; Biome on changed files is warnings only (the one in `online.ts:1621` is pre-existing, unrelated).

## Findings

**1. (should fix) A test passes with the exact bug it is named after.** `tests/reconnect_overlay.test.ts`, "replaces the tick interval rather than stacking a duplicate". I deleted the `clearInterval` from `showReconnectOverlay` to reintroduce the stacking bug and all 4 tests still passed. Both intervals fire on the same 1000ms tick in registration order, so the second render overwrites the first and `toContain('2')` still holds. `expect(vi.getTimerCount()).toBe(1)` after the second show closes it. The sibling "clears the interval" case has the same shape: it proves `tickTimer` was nulled, not that the interval stopped, so advancing timers and asserting `getTimerCount()` is 0 would make it decisive.

**2. (should fix) The `onConnectionLost` contract change has zero coverage.** `grep -rn onConnectionLost tests/` returns nothing. The signature went from `()` to `(attempt, maxAttempts, nextRetryAtMs)`, gained a second call site in the visibilitychange path, and was deliberately reordered to fire after the timer is armed. I read all three sites and they are correct, but an off-by-one on `attempt`, or passing a relative `delayMs` where an absolute timestamp is expected, would ship green today. `tests/net_online_visibility_reconnect.test.ts` already has the `StubWebSocket` and fake-document harness. The highest-value single test is the one for the reorder rationale: make `onConnectionLost` throw, drive `socketClosed()`, advance past the backoff, and assert a new socket was opened. That fails on the old ordering and is the decisive pin for this commit's main safety claim.

**3. (should fix) The slow-hint feature grew onto `main.ts` rather than landing as a sibling module.** It adds two module-level mutables (`slowHintTimer`, `lastLoadingProgressAt`) and three top-level functions. None of it needs `main.ts`'s private state; it is a self-contained timer plus a DOM toggle over one element. Root `CLAUDE.md` is explicit that `main.ts` is a firewall and never a home for a top-level function a sibling module could hold. This PR already got this right for the reconnect half, and `src/ui/reconnect_overlay.ts` is the shape to copy. It is also why the arm/disarm policy, which carries the change's longest rationale comment, is the one untested part: a `src/ui/loading_slow_hint.ts` exporting start/stop/noteProgress would make the reset-on-progress and stop-at-assets transitions pinnable the same way the overlay is.

**4. (nit) Layout shift on reveal.** `#ls-tip` carries `min-height: 1.2em` precisely to stop the loading column from jumping. `#ls-slow-hint` sits next to it with no reserve and toggles `display: none` to `block`, so revealing a one or two line hint re-centers the flex column and nudges the progress bar. Reserving the height (or toggling visibility/opacity) avoids it.

**5. (nit) Screenshot weight.** The two loading PNGs are 5.3MB each, the largest in the repo against a 512KB median across 577 files, about 16MB added for five images. Worth downscaling before merge since these are permanent history.

## Notes, no action needed

- i18n is handled correctly: this is the M16 exception, with the retired `loading.reconnecting` removed everywhere and non-Latin fills (ja, ko, ru, zh_CN, zh_TW) supplied in the same change, Latin locales left pending for the maintainer. No overlay edited outside that sanctioned path.
- The bare-number interpolation of `{attempt}`/`{seconds}` instead of `formatNumber` matches the existing `loading.worldProgress` precedent, so it is a repo-wide pattern, not a regression here.
- The reconnect overlay has no `role="status"` and no accessible name, so a screen-reader user gets no notification that the world paused. That is pre-existing rather than introduced by this PR, but now that the message is a live countdown it would be a natural time to add a polite live region that announces the transitions (drop, retrying now, resumed) rather than every 1s tick.
- One small correction to the PR description: it justifies the null-guard on `#ls-slow-hint` by saying the element exists on both entries. True today, but the guard means a future entry that drops the element fails silently (the hint just never shows) instead of loudly. The new `client_shell` pin is what actually protects this, so that pin is the load-bearing part.

Happy to re-review once 1 and 2 are closed. 3 is the one I would most like to see, but I understand if you would rather keep this PR scoped and do it as a follow-up.

jgyy:
Rebased onto the current tip of release/v0.28.0 (the branch had drifted to CONFLICTING; the only conflict was the generated `i18n.resolved.generated/pending.ts`, resolved by regenerating via `npm run i18n:gen` after the rebase, which left the tree clean) and pushed the fixes for the latest review round (2026-07-19, all three threads COMMENTED, no new blockers):

1. **Decisive interval assertions.** `tests/reconnect_overlay.test.ts`'s "replaces the tick interval" and "clears the interval" cases now assert `vi.getTimerCount()` directly instead of inferring timer state from rendered text, closing the gap where the test passed even with the stacking/leak bug reintroduced.
2. **`onConnectionLost` contract coverage.** Added three cases to `tests/net_online_visibility_reconnect.test.ts`: the call args (attempt, maxAttempts, absolute `nextRetryAtMs`) on a real drop, the re-fire with an updated timestamp on the mobile-foreground fast-retry path, and the reorder rationale itself (a throwing callback still leaves the already-armed retry intact, so a DOM/`t()` failure inside the callback cannot permanently kill reconnect).
3. **Extracted the slow-hint arm/disarm logic out of `src/main.ts`** into `src/ui/loading_slow_hint.ts` (start/stop/note), consuming the existing pure `loading_slow_hint_core.ts` threshold check, with its own test file (`tests/loading_slow_hint.test.ts`). `main.ts` is now a thin consumer, matching the root CLAUDE.md's "firewall, not a home" rule and the shape the reconnect half already had.
4. **Layout shift fix.** `#ls-slow-hint` now reserves a `min-height` and toggles via `visibility` instead of `display`, so revealing it no longer nudges the progress bar underneath it.

Also downscaled the two oversized loading-screen screenshots (5.3MB each) that were flagged as the largest in the repo; they're now in line with the rest of `docs/screenshots`.

`npx tsc --noEmit` clean, targeted vitest files (`architecture`, `world_api_parity`, `i18n_completeness`, `client_shell`, `backoff`, `net_online_visibility_reconnect`, `linkdead`, `reconnect_status_core`, `loading_slow_hint_core`, `loading_slow_hint`, `reconnect_overlay`, `localization_fixes`) all green, changed-file Biome clean (warnings only, matching pre-existing patterns in the same file), `git status` clean after `npm run i18n:gen`.

jgyy:
Thanks for the three thorough passes, and for confirming the /play fix and the parity/seam checks with the frontend-seam-reviewer and cross-platform-sync agents.

Checked the branch tip (667d402d0) against every item raised across all three reviews:

- CHANGES_REQUESTED: play.html now ships #ls-slow-hint with the null-safe lookup, pinned by tests/client_shell.test.ts; the slow-connection watch stops as soon as assetsReady() resolves so it cannot fire during the CPU-bound scene-build phase; reconnect_status_core.ts lives under src/ui/ so the net/ import direction is fixed; and loading.reconnecting was retired with the M16 non-Latin fills supplied in the same change.
- Second COMMENTED pass: onConnectionLost now fires only after reconnectTimer is armed, so a throwing callback can no longer kill auto-reconnect for the session; the mobile-foreground visibilitychange handler re-fires onConnectionLost with the updated nextRetryAtMs so the countdown stays honest on the phone-unlock fast retry; #ls-slow-hint carries role=status and aria-live=polite in both index.html and play.html, with inline English fallback text matching the #ls-status sibling pattern.
- Third COMMENTED pass: tests/reconnect_overlay.test.ts now asserts vi.getTimerCount() directly instead of inferring timer state from rendered text, closing the stacked/leaked-interval gap; tests/net_online_visibility_reconnect.test.ts covers the onConnectionLost contract (call arguments on a real drop, the mobile fast-retry re-fire, and the reorder rationale itself); the slow-hint arm/disarm logic moved out of main.ts into its own src/ui/loading_slow_hint.ts sibling module consuming loading_slow_hint_core.ts, with its own test file; and the hint now reserves layout space via a visibility toggle instead of display:none/block, plus the two oversized loading screenshots were downscaled.

Ran npx tsc --noEmit, the full guard set (architecture, world_api_parity, i18n_completeness, client_shell, backoff, net_online_visibility_reconnect, linkdead, reconnect_status_core, loading_slow_hint_core, reconnect_overlay, loading_slow_hint) at 451 passing, and npm run i18n:gen leaves the tree clean. No outstanding findings from any of the three reviews remain; nothing further to push, branch tip is unchanged at 667d402d0.

EnriqueGF (approved):
## Review: approve

Reviewed the full diff against `release/v0.28.0`, read `src/net/CLAUDE.md`, `src/ui/CLAUDE.md`, `src/styles/CLAUDE.md` and the root invariants, and re-ran the checks myself rather than trusting the PR body.

### What I verified

**Invariants**
- `src/sim/` is untouched: no determinism or `Rng` surface in play. The new `Date.now()` calls all live in `src/net/` and `src/ui/`, which is where wall-clock is legal.
- `IWorld` surface unchanged: `onConnectionLost` is a `ClientWorld`-only callback field, not an `IWorld` member, so no `src/world_api/` facet or `tests/world_api_parity.test.ts` pin update is owed. Confirmed the parity test still passes.
- Server authority intact: both changes are purely informational client-side UI.
- Module-first respected: two new pure cores (`loading_slow_hint_core.ts`, `reconnect_status_core.ts`) registered in `UI_PURE_CORES`, with the stateful halves as thin siblings. Nothing was appended to `main.ts` beyond three call sites, which is the right shape for a coordinator that is a firewall.
- The `reconnect_status.ts` move under `src/ui/` fixes the import direction correctly: `net/` no longer owns display math consumed only by the overlay.

**i18n**
- English lands in `src/ui/i18n.catalog/shell.ts` only. The overlay edits to `ja_JP`, `ko_KR`, `ru_RU`, `zh_CN`, `zh_TW` are exactly the M16 non-Latin exception for the new wordy values, and the orphaned `loading.reconnecting` is removed cleanly from all 18 overlays alongside the catalog rename. `pending.ts` regenerates to exactly what is committed.
- `#ls-slow-hint` uses `data-i18n` with the same inline-English-plus-`translatePage()` pattern as its `#ls-status` sibling, so no hardcoded rendered string.

**Correctness, the parts I actually traced**
- `onConnectionLost` firing *after* `reconnectTimer` is armed is the right ordering, and the test that pins it (a throwing callback still leaves the retry armed and firing) is decisive, not decorative. Worth keeping.
- The mobile-foreground fast-retry path now re-fires with the new short `nextRetryAtMs`; without it the countdown would have kept ticking toward a backoff that no longer applies. Good catch, and pinned.
- Traced every overlay exit path for a leaked tick interval: `onReconnected`, `onDisconnect`, and the 10s entry-timeout branch in `enterWorld` all call `hideReconnectOverlay()`. No leak.
- `secondsUntilRetry` ceilings, so the overlay never shows `0s` while a retry is still pending, and the `reconnectingNow` swap covers the post-zero window instead of repainting a stale `0s`.
- Loading-hint lifecycle: `showLoadingScreen` is reachable once per page load (`hasBegunWorldEntry`), the watch is stopped on the `assetsReady` success path, on `fatalOverlay` (via `hideLoadingScreen`), and on normal `hideLoadingScreen`. Restricting the watch to the network-bound asset phase is the correct fix for the CPU-misattribution concern; note the `ensureLocaleLoaded` fetch just before it is also genuinely network-bound, so the hint covering that window is right, not a gap.
- The `play.html` fix is real and the `tests/client_shell.test.ts` pin covers both entries, including the exact element markup.

**Ran locally**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/reconnect_overlay.test.ts tests/reconnect_status_core.test.ts tests/loading_slow_hint.test.ts tests/loading_slow_hint_core.test.ts tests/net_online_visibility_reconnect.test.ts tests/i18n_completeness.test.ts`: 411 passed.
- `npm run i18n:gen` then `tests/localization_fixes.test.ts` (the S3 guard): 29 passed, 3 skipped, and `git status` clean afterwards, so the committed generated artifacts match.
- Biome on the changed source files: warnings only, all pre-existing patterns, no errors or format diffs.

### Findings

No blocking issues. Two non-blocking notes:

1. **a11y (minor, `index.html` / `play.html` / `shell.css`)** The hint is a live region (`role="status" aria-live="polite"`) whose text is static from page load, revealed only by toggling `visibility: hidden` to `visible`. A live region that is outside the accessibility tree when the page loads is generally not monitored, and browsers do not reliably announce on a visibility change alone, so a screen-reader user may never hear the hint. Setting `textContent` at reveal time (or toggling the `hidden` attribute) would make the announcement fire dependably. Cosmetic-only feature, so this does not block.

2. **Layout (very minor, `shell.css`)** `min-height: 2.4em` reserves two lines to avoid a jump when the hint appears, which is the right instinct, but at `12px` under `max-width: min(480px, 80vw)` the English string can wrap to three lines on a narrow phone and reintroduce the shift. Worth a glance next time you are in that file.

Both changes are informational-only, well scoped, and land behind the seams the repo already has. Happy to merge.

PR #2107, opened 2026-07-18
feat(ui): lay out the World Market as a landscape multi-column listing grid

jgyy:
## Summary
- The World Market window was a tall 560px single column, so browsing more than a
  handful of listings meant a lot of vertical scrolling. Widen the window to 860px
  and lay the listing list out as a CSS grid with auto-fill columns so results flow
  landscape-style across multiple columns on desktop.
- The note/empty-state/pager rows still span the full grid width. Mobile keeps a
  deliberate single-column override (the window is edge-pinned narrow there), so the
  touch experience is unchanged.
- Follow-up (this update): after reviewing the screenshots, the maintainer flagged that
  the Browse tab's search box and its type/subtype/rarity filter dropdowns each stretched
  edge to edge of the wider 860px window and stacked on separate rows, looking oversized.
  The search box and the filter menus now sit together in one `.mkt-controls` flex row,
  both capped to a proportionate max-width (260px for search, 220px per filter select),
  wrapping onto their own line only when the window is too narrow to fit them (mobile).
  The search box moved out of the dynamically rebuilt `#market-body` into the window's
  static header markup so it still survives a `renderContent()`-only refresh while typing
  (search-as-you-type never triggers the full `render()` rebuild).
- Follow-up 2 (this update): review flagged that moving the search box out of
  `#market-body` added a fourth stacked chrome row on mobile (tabs + controls + body).
  Under the window's `overflow: hidden` with `#market-body` as the only scroll
  container, a tall controls row (the subtype filter's third menu on a short
  viewport) could push content past the window's bottom edge with no way to reach
  it, clipping the filters and the entire listing list out of view. `#market-window`
  now scrolls the whole sheet on mobile instead of clipping, and `#market-body` sizes
  to its natural content height instead of flexing to fill a fixed remainder, so
  tabs, controls, and listings scroll together as one column and nothing under the
  fold is ever unreachable.

```mermaid
flowchart LR
    A[Open World Market] --> B{Viewport width}
    B -->|Desktop, wide window| C[".mkt-list grid\nauto-fill columns\nmultiple listing cards per row"]
    B -->|Mobile, edge-pinned narrow| D[".mkt-list\nforced 1 column\n(hud.mobile.css)"]
    C --> E[Note / empty state / pager\nspan full grid width]
    D --> E
    F[".mkt-controls row\nsearch + filters"] -->|Desktop| G["side by side\nmax-width capped"]
    F -->|Mobile| H["stacked\none control per row"]
    D --> I["#market-window scrolls\nthe whole sheet\n(tabs+controls+body together)"]
```

## Screenshots

Desktop (Browse tab, Merchant's standing stock):

Before (single column, tall scroll):
screenshot

After (landscape, multi-column grid, search + filters on one compact row):
screenshot

Mobile (unchanged: still a single edge-pinned column, controls stack one per row; the
sheet now scrolls as a whole so the filters and listings are always reachable):

Before:
screenshot

After:
screenshot

## Test plan
- [x] `npx @biomejs/biome check` clean on every changed file
- [x] `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/market.test.ts tests/market_db.test.ts tests/pr_shot_targets.test.ts tests/localization_fixes.test.ts`: 520 passed, 3 skipped
- [x] `npm run i18n:gen` leaves the tree clean
- [x] Rebased onto latest `release/v0.28.0`, resolved one conflict in `scripts/pr_shot_targets.mjs`
- Added a `market-window` target to `scripts/pr_shot_targets.mjs` (desktop + mobile
  variants, `when` list includes both `styles/components.css` and `styles/hud.mobile.css`)
  so future market CSS changes get change-aware screenshot coverage automatically.


EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read `src/ui/CLAUDE.md` and `src/styles/CLAUDE.md`, and ran locally in the worktree:

- `npx tsc --noEmit`: clean
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/hud_perf_budget.test.ts`: 465 passed, 3 skipped
- `npm run i18n:gen` leaves the tree clean

The desktop result is good, and the seam work is right: no new `IWorld` member, no `*_view`/`*_core` to register, `market_window.ts` is a cold window so the write-elision contract does not apply, the `aria-label` move from `setAttribute` to an interpolated `esc(t(...))` reuses the existing key and is attribute-safe, and the filter dropdowns are unaffected by moving `position: relative` from `.mkt-filters` to `.mkt-controls` because `.mkt-select-menu` anchors to `.mkt-select`, which has its own `position: relative`. Focus preservation while typing is also not regressed: `render()` already wiped the old in-body field, and typing only reaches `renderContent()`.

One blocking issue, though, and the PR's own screenshot documents it.

## Blocking: the mobile search box becomes a ~390px tall block and pushes the filters and the listing body out of the window

`components.css` gives `.mkt-search` `flex: 1 1 200px`. On mobile, `hud.mobile.css` flips the container to `body.mobile-touch .mkt-controls { flex-direction: column }` but only overrides `max-width: none`, never the `flex` shorthand. In a column flex container the flex-basis is the **main size, i.e. the height**, so the search input is laid out 200px tall (and grows further), instead of the ~35px it was.

The same shape applies one level down: `.mkt-filters` keeps `flex: 2 1 auto` and `.mkt-filter` keeps `flex: 1 1 140px`, so once `.mkt-filters` is also `flex-direction: column` each filter row gets a 140px flex-basis height as well. Header total lands well past the sheet height, and `body.mobile-touch #market-window` is `overflow: hidden`, so everything below the search box is clipped away with no scroll path to it (exactly the failure mode the existing comment at `hud.mobile.css:2648` warns about).

Compare the two committed mobile shots, which are the same 1696x664 viewport:

- `before-mobile.png`: TYPE and RARITY selects, a normal-height search box, and "Showing 1-21 of 21 listings."
- `after-mobile.png`: a single search box roughly 390px tall with its placeholder vertically centered, no filter menus at all, and no listings.

So the PR's claim that "mobile keeps a deliberate single-column override, the touch experience is unchanged" is not what the evidence shows: on mobile the Browse tab currently loses its filters and its entire listing list. Fix is to neutralize the main-size basis in the mobile block, along the lines of:

```css
body.mobile-touch .mkt-search { flex: 0 0 auto; max-width: none; }
body.mobile-touch .mkt-filters { flex: 0 0 auto; flex-direction: column; }
body.mobile-touch .mkt-filter { flex: 0 0 auto; max-width: none; }
```

then re-capture `after-mobile.png` (and please capture with a type filter that exposes the subtype menu, so the three-select case is on the record). Worth a `tests/mobile_window_layout.test.ts` pin too, since a text-level CSS check is what would have caught the missing override.

## Should fix

- **Stale `has-subtype` class.** `.mkt-filters.has-subtype` was deleted from `components.css` and its mobile twin removed, but `market_window.ts:769` still emits the class and `tests/client_shell.test.ts:1745` still pins the emission. The class is now inert. Either drop it and update the pin, or keep it and say why in a comment. The `components.css:8203` comment also still narrates "the has-subtype three-menu case" for a rule that no longer exists.
- **`.mkt-search` has no mobile 40px floor.** It keeps `padding: 7px 10px` at 16px font, about 35px tall. Pre-existing, not a regression, but this change promotes it into the always-visible header and already touches its mobile rule, so it is the natural place to add `min-height: 40px`. The 16px font floor is correctly preserved.
- **`pr_shot_targets.mjs` `when` list omits `styles/hud.mobile.css`.** This PR's whole mobile behavior lives in that file, yet a future market-only change there would not trigger the new `market-window` target. Worth adding.

## Notes

- `.mkt-row` is tight at the 260px column floor: 30px icon + 27px of gaps + 12px padding + a nowrap price column + a nowrap Buy button leaves roughly 60-70px for the item name. It does not overflow (`.mkt-name` has `overflow: hidden`), but names ellipsis-clip hard, which is visible in `after-desktop.png`. Consider a ~300px `minmax()` floor.
- `controlsHtml` now renders whenever the tab is browse, including the `no-data` state, so a live search box and three filter menus sit above the "no merchant" message. Previously they vanished with the body. Low impact, just a behavior change worth knowing.
- The grid-span sweep (`.mkt-note`, `.mkt-empty`, `.mkt-page`) is correct and complete: `.mkt-status` is appended to `#market-body`, not to `.mkt-list`, so it is not a grid item. A pin keeping that selector list in sync with the non-`.mkt-row` children would stop a future fourth child from silently becoming a 260px column.

Happy to re-review as soon as the mobile flex-basis is fixed and the shot re-captured. The desktop layout itself is a clear improvement.

jgyy:
Thanks for the thorough review, and for catching the blocking issue with a screenshot to back it up. All four points are fixed in the latest push:

**Blocking: mobile search box flex-basis**
Confirmed the root cause exactly as described: `.mkt-search` (`flex: 1 1 200px`), `.mkt-filters` (`flex: 2 1 auto`), and `.mkt-filter` (`flex: 1 1 140px`) each carry a desktop flex-basis meant as a row *width*, and once `body.mobile-touch .mkt-controls`/`.mkt-filters` flip to `flex-direction: column` that basis becomes a *height* instead. Fixed in `src/styles/hud.mobile.css` with `flex: 0 0 auto` on all three, exactly along the lines you suggested. Re-captured `after-mobile.png` with a Weapons type filter selected so the three-select case (Type / Weapon Type / Rarity) is on record: search box back to normal height, all three filter selects visible, listing count showing again. Added a `tests/mobile_window_layout.test.ts` pin asserting the `flex: 0 0 auto` override on all three selectors.

**Should fix: stale `has-subtype` class**
Dropped the emission in `market_window.ts` (the class had no CSS backing it anymore), updated the `tests/client_shell.test.ts` pin, and reworded the now-inaccurate `components.css:8203` comment that still narrated the removed rule.

**Should fix: `.mkt-search` mobile min-height**
Added `min-height: 40px` to the mobile `.mkt-search` rule alongside the flex-basis fix above.

**Should fix: `pr_shot_targets.mjs` coverage**
Added `styles/hud.mobile.css` to the `market-window` target's `when` list, and updated the two existing `pr_shot_targets.test.ts` cases that depended on `hud.mobile.css` mapping to nothing specific (one now uses a neutral mobile-visual fixture path, the other asserts the new market-window mapping directly).

Verified locally:
- `npx tsc --noEmit`: clean
- `npx vitest run tests/mobile_window_layout.test.ts tests/client_shell.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts`: 182 passed
- `npx @biomejs/biome check --write` on all changed files: clean, no diffs

Happy to re-capture anything else if the new shot doesn't fully address the concern. Thanks again for the review!

EnriqueGF (changes requested):
Reviewed the full diff (CSS relayout, `market_window.ts` search-box move, the `pr_shot_targets` entry and the test updates) against the root and `src/ui` / `src/styles` CLAUDE.md rules.

## What I ran locally (worktree on `pr-2107`, base `release/v0.28.0`)

- `npx tsc --noEmit`: clean.
- `npx @biomejs/biome check` on all 6 changed text files: clean.
- `npx vitest run tests/architecture.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/market.test.ts tests/pr_shot_targets.test.ts`: all green (483 tests).
- `tests/localization_fixes.test.ts` (S3 guard): green after `npm run i18n:gen` (it only failed first because the generated `i18n.status.json` was absent in a fresh worktree, not from this change).

## What I verified by reading

- **i18n**: the placeholder and `aria-label` still go through `t()` plus `esc()`; no new keys, no locale overlay edits. `src/ui/esc.ts` escapes `"` and `'`, so the new `value="${esc(this.searchQuery)}"` attribute interpolation is safe for a query containing quotes.
- **Determinism / parity / server authority**: not in play, the diff touches no `src/sim/`, `server/` or `IWorld` surface.
- **`has-subtype` removal is complete**: no leftover reference anywhere in `.ts` / `.css` / `.mjs`, including the old `body.mobile-touch .mkt-filters.has-subtype` override which this diff correctly replaces.
- **`grid-column: 1 / -1` actually applies**: `.mkt-note` / `.mkt-empty` / `.mkt-page` are appended to `list`, not to `body`, so they are real grid children and do span every column.
- **The `.window` max-width claim in the new comment is accurate**: `src/styles/layout.css` clamps `.window` to `calc(var(--app-vw) / var(--window-scale) - 16px)`, so 860px does shrink on a narrow desktop viewport and `auto-fill` falls back to one column without a hardcoded breakpoint.
- **The `flex-basis`-becomes-height fix in 5329f8a24 is right**, and pinning it in `tests/mobile_window_layout.test.ts` is a good catch. Nice.
- **`.mkt-select` carries its own `position: relative`**, so dropping `position: relative` from `.mkt-filters` does not move the dropdown menus' containing block.

## Findings

**1. Medium, mobile: moving the search box out of the scroll body adds a fourth row of non-scrollable chrome, and your own after-mobile.png shows the listing body clipped entirely out of the window.**

Before this change, the search input was `body.appendChild(search)` inside `#market-body`, which is the one scroll container (`overflow-y: auto; flex: 1`). It now lives in `.mkt-controls`, a sibling above `#market-body`, so on mobile it becomes a fixed row stacked on top of the already-stacked filter selects.

`body.mobile-touch #market-body`'s existing comment already flags the hazard: "On a very short viewport the window's overflow:hidden can still clip below this floor". This change makes that strictly worse by one ~50px row, and `docs/screenshots/market-landscape-layout/after-mobile.png` is the repro: the RARITY select is cut off at the window's bottom edge, `#market-body` is not visible at all, and because `body.mobile-touch #market-window` sets `overflow: hidden` there is no way to scroll down to it. On that viewport the Browse tab shows zero listings.

Suggestion: under `body.mobile-touch`, let `.mkt-controls` scroll with the listings rather than sitting in fixed chrome (move it inside the scroll container for the touch layout, or give the sheet an outer scroll there). The desktop one-row layout is unaffected either way.

**2. Minor: the mobile before/after screenshots are not comparable, so they do not support the "touch experience is unchanged" claim.**

`before-mobile.png` is captured with TYPE = "All types" (two filter selects), `after-mobile.png` with TYPE = "Weapons" (three, since the subtype menu appears). Some of the extra height in the "after" shot is that state difference, not the relayout. Worth re-capturing both with matching filter state so the comparison is honest, and it would also make finding 1 easy to size.

**3. Nit: the `no-data` branch now renders live filter chrome above a "no merchant" message.**

`renderContent()`'s `view.kind === 'no-data'` path replaces `#market-body` with the `itemUi.market.noMerchant` notice, but `.mkt-controls` sits outside `body` now, so the search box and filter dropdowns stay on screen above it. Previously the search box was wiped along with the body. Transient state, low stakes, just noting it.

Everything else looks solid, and the `market-window` target added to `scripts/pr_shot_targets.mjs` (with the honest comment about hiding the docked Bags companion) is a genuinely useful addition for future market CSS work. Happy to re-review once finding 1 is addressed.

EnriqueGF (changes requested):
Reviewed the four commits on the branch (263083404, fa2e2b1c0, baf486bee, 5329f8a24), read the market painter and both stylesheets against the current code, and ran the checks locally in a clean worktree off `release/v0.28.0`.

**What I ran**
- `tsc --noEmit`: no new errors (the only failures are the pre-existing `src/net/wallet_connect.ts` / `tweetnacl` unresolved-module ones, unrelated to this diff).
- `vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/market.test.ts tests/market_db.test.ts tests/pr_shot_targets.test.ts tests/guide.test.ts`: all green.
- Copy scan over the diff: no em dashes, en dashes or emoji.

**What holds up.** The desktop relayout is well done. `.mkt-select` already carries its own `position: relative`, so moving `position: relative; z-index` from `.mkt-filters` up to `.mkt-controls` does not shift the filter dropdown's containing block, and `positionFilterMenu` still measures against the same `#market-window` clip box. The `.window` max-width clamp in `layout.css` really does exist, so the 860px width degrades correctly on a narrow viewport, and the `@layer hud-mobile` ordering puts the mobile overrides after `components.css` as the comments claim. Moving the search box into `render()` is safe: `render()` is only called from open / tab switch / filter pick / stageSell / list-submit, never from the per-frame `refreshIfChanged`, so search-as-you-type still keeps focus, and `esc()` escapes `"` so the `value="..."` interpolation is sound. No new i18n keys, both `t()` calls moved intact.

Two things I do want changed before this merges.

---

**1. (blocking) The mobile listing body gets clipped out of the window when the subtype filter is showing.** `src/styles/hud.mobile.css`

Moving `.mkt-search` out of the scrollable `#market-body` and into the static header adds roughly 33px of non-scrollable chrome above the body on mobile. `body.mobile-touch #market-window` is `max-height: calc(var(--app-vh)/var(--ui-scale) - 20px)` with `overflow: hidden`, and `#market-body` has `min-height: 96px`, so once the header outgrows the window the body is clipped rather than scrolled.

On the tooling's own mobile viewport (iPhone landscape, 844x390 CSS px, so about 370px of window) with Type = Weapons, the three stacked filters plus the new search row put the header at roughly 322px; 322 + the 96px body floor is 418px against a 370px clip. Before this change the same state was about 289 + 96, and the search box lived inside the scroll region, so listings stayed reachable.

This is visible in the PR's own evidence: `after-mobile.png` shows RARITY cut off at the window edge and zero listings, while `before-mobile.png` still shows "Showing 1-21 of 21 listings" at the bottom. So the "the touch experience is unchanged" line in the description is not accurate as captured. Two-filter state is still fine; it is the armor/weapon subtype state that breaks.

Simplest fix that keeps the desktop win: leave the search box inside `#market-body`'s scroll region on `body.mobile-touch` (or move `.mkt-controls` as a whole into the scrollable area there), so the mobile header height goes back to what it was. Worth a capture of the Weapons filter state to confirm.

---

**2. (blocking) The new screenshot target claims two repo-wide stylesheets, which silently kills the generic HUD screenshots for every future CSS PR.** `scripts/pr_shot_targets.mjs:224`

The `when` list includes `'styles/components.css'` and `'styles/hud.mobile.css'`. `classifyDiff` only emits the `hud-desktop` / `hud-mobile` fallback frames when `specific.length === 0`, and `resolveTargets` is a substring match, so:

```
src/styles/components.css -> specific: [ 'market-window' ] generic: []
src/styles/hud.mobile.css -> specific: [ 'market-window' ] generic: []
src/styles/hud.css        -> specific: []                 generic: [ 'hud-desktop' ]
```

(run against this branch). Those two files are where nearly all HUD CSS lives. From here on, a PR that restyles the quest log or the action bar in `components.css` gets exactly one screenshot, of the World Market, and loses the generic HUD frames entirely. Every other target in the file scopes `when` to its own modules (`ui/bags`, `ui/map`, `sim/professions`); this is the only one claiming a shared stylesheet. The fact that the change had to rewrite two existing `pr_shot_targets.test.ts` cases onto `src/game/mobile_controls.ts` to dodge the collision is the tell.

Suggest dropping both stylesheet entries and keeping the target scoped to `['ui/market_window', 'ui/market_view', 'ui/market_filters', 'sim/market']`, then reverting the two test rewrites and the new `expect(plan.generic).toHaveLength(0)` assertion.

---

**Minor, non-blocking**

- `.mkt-note` keeps `margin: 2px 0 8px` while `.mkt-list` now supplies `gap: 4px 10px`, so the note's bottom margin stacks on top of the row gap. Cosmetic only.
- At the 260px `minmax` floor, a listing card gives the name column roughly 70px after the 30px icon, the three 9px gaps, the price and the Buy button. It ellipsizes cleanly, but a slightly larger floor (280 to 300px) would read better on a 1280px-wide window. Your call.
- Picking a filter now rebuilds the search `<input>` rather than reusing it. The typed text survives via the `value=` attribute so there is no data loss, just noting the behavior changed.

jgyy:
Rebased onto the latest `release/v0.28.0` (resolved a single conflict in `scripts/pr_shot_targets.mjs`, a target-list ordering clash with an unrelated new masterwork-tooltip target that landed on the base branch) and addressed the blocking issue from both reviews.

**Blocking: mobile search box clipped the filters and listing body out of the window (both reviews).** Moving the search box into `.mkt-controls` added a fourth stacked chrome row (tabs + controls + body). Under `overflow: hidden` with `#market-body` as the only scroll container, a tall controls row (the subtype filter's third menu on a short viewport) could push content past the window's bottom edge with no way to reach it, exactly as your after-mobile.png showed (RARITY clipped, zero listings visible).

Fix: `body.mobile-touch #market-window` now scrolls the whole sheet (`overflow-y: auto`) instead of clipping, and `#market-body` sizes to its natural content height (`flex: none`) instead of flexing to fill a fixed remainder. Tabs, controls, and the listing body now scroll together as one column, so nothing under the fold is ever unreachable. Updated the `tests/client_shell.test.ts` pin to match.

**Mismatched before/after screenshots (2nd review, minor).** Recaptured `after-desktop.png` and `after-mobile.png` against the fix, using the same filter state (All types / All rarities) as the before shots, so RARITY, the search box, and "Showing 1-21 of 21 listings." are all visible in the new after-mobile.png.

Verified locally: `npx tsc --noEmit` clean, `npx @biomejs/biome check` clean on every changed file, and the full relevant vitest set green (`client_shell`, `css_corpus`, `css_value_validity`, `mobile_window_coverage`, `mobile_window_layout`, `mobile_window_transform`, `market_window`, `market_view`, `market_filters`, `market`, `market_db`, `pr_shot_targets`, `architecture`, `world_api_parity`, `localization_fixes`), 520 passed / 3 skipped. `npm run i18n:gen` leaves the tree clean.

The `has-subtype` stale class and the `pr_shot_targets.mjs` mobile CSS `when` entry flagged in the first review were already resolved in the prior commit on this branch (confirmed by the second review too), so no further change was needed there.

jgyy:
Thanks for the thorough review across all three rounds, Enrique.

Cross-checked the latest review against the commits already on the branch: the mobile listing-body clipping finding (round 3, item 1) was already fixed by the sheet-scroll commit that landed right around the same time as your review (`#market-window` now scrolls the whole sheet on mobile instead of clipping, and `#market-body` sizes to its natural content height). That one is confirmed already addressed.

The one item still outstanding was round 3, item 2: the market-window PR shot target's `when` list claimed two repo-wide stylesheets, `styles/components.css` and `styles/hud.mobile.css`. Since `resolveTargets` is a substring match and `classifyDiff` only falls back to the generic `hud-desktop`/`hud-mobile` frames when no specific target matched, a future PR restyling any other HUD component through either of those shared stylesheets would have captured only the World Market window and silently lost the generic HUD screenshots.

Fixed by scoping the target's `when` list down to `['ui/market_window', 'ui/market_view', 'ui/market_filters', 'sim/market']` (dropping the two stylesheet entries), and reverting the two `pr_shot_targets.test.ts` cases that had been rewritten onto `src/game/mobile_controls.ts` to dodge the collision, back to asserting the original generic-fallback behavior on `hud.mobile.css`.

Verified: `npx tsc --noEmit` clean, and `tests/architecture.test.ts`, `tests/pr_shot_targets.test.ts`, `tests/market_window.test.ts`, `tests/market_view.test.ts`, `tests/market_filters.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/mobile_window_coverage.test.ts`, `tests/mobile_window_layout.test.ts`, `tests/mobile_window_transform.test.ts`, `tests/client_shell.test.ts`, `tests/world_api_parity.test.ts`, `tests/market.test.ts`, `tests/market_db.test.ts`, and `tests/guide.test.ts` all pass. Changed files formatted with biome.

The minor, non-blocking notes from round 3 (the `.mkt-note` margin/gap stacking, the 260px card floor, and the search input rebuild on filter pick) I left as is per your "your call" framing, since they are cosmetic and not requested changes.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the local `CLAUDE.md` for `src/styles/`, `src/ui/` and `scripts/`, and ran the checks myself.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/market.test.ts tests/pr_shot_targets.test.ts`: all green (the one initial red was a missing `src/ui/i18n.status.json`, fixed by `npm run i18n:gen`, not a PR issue; the S3 guard passes and `i18n:gen` leaves the tree clean).
- I tried to measure the desktop layout in a real headless browser, but the world does not boot in my sandbox (no GPU), so finding 1 below is derived from the CSS rather than a live capture. Please confirm it on real hardware.

Invariants I checked and found clean: no `src/sim/` or `server/` changes, so determinism, the `IWorld` seam and tri-host parity are not in play. i18n is fine: the moved placeholder and `aria-label` still come from `t('itemUi.market.searchPlaceholder')` / `t('itemUi.market.searchAria')`, and routing them through `esc(t(...))` in an attribute is an improvement over the old `setAttribute` path. No new keys, no locale overlays touched. No `ALLOW_DEV_COMMANDS`, no secrets, no em dashes, no en dashes, no emojis. `esc()` escapes both quote characters, so `value="${esc(this.searchQuery)}"` is safe. `.mkt-select` keeps `position: relative`, so dropping `position: relative` from `.mkt-filters` does not move the dropdown anchor. On mobile the shared sheet base already declares `overflow-y: auto`, so removing market's own `overflow: hidden` makes it match the base rather than diverge, which is the right direction.

The relayout itself is a good idea and the mobile follow-ups are reasoned. Four things I would like addressed before merge.

## 1. The wider window now overlaps the Bags panel it opens alongside (desktop)

`openMarket()` calls `syncBags(true)`, and unlike vendor/bank there is no docking CSS pairing them. `#market-window` is centred (`left: 50%` + `translateX(-50%)`), and `#bags` (`components.css` around line 4788) pins itself with `left: calc((100% + 50% + var(--bags-bar-half) + var(--bags-gap) - var(--bags-micro-r)) / 2)` plus `translateX(-50%)`. Both resolve percentages against `#ui`, which is `position: fixed; width: var(--app-vw)` (`base.css:468`), so at `--ui-scale: 1` the containing block is the viewport width W.

Market right edge = `W/2 + 430`. Bags left edge = `0.75W + 126 - bagsWidth/2`. Overlap whenever W is below about 1836px:

| viewport | market right | bags left | overlap |
|---|---|---|---|
| 1280 | 1070 | 958 | 112px |
| 1440 | 1150 | 1051 | 99px |
| 1536 | 1198 | 1123 | 75px |
| 1920 | 1390 | 1411 | none |

At the old 560px width the market right edge was `W/2 + 280`, which cleared bags at all of these. Since the Market's core workflow is clicking bag items onto the Sell tab, covering 75px to 112px of a 256px to 310px wide Bags panel on every common laptop width is a real usability regression. Worth noting that `scripts/pr_shot_targets.mjs` in this PR hides `#bags` for the capture, so the committed desktop screenshot cannot show this. Options: cap the width (something like `min(860px, ...)` that leaves room), or add a market-open docking rule the way `body.bank-open #bags` does.

## 2. `scripts/market_search_shot.mjs` is broken by the search-box move

Four selectors scope to `#market-body .mkt-search`, which no longer matches now that the input lives in `.mkt-controls`:

- line 76 and 113: `hasSearch` is now always false, so `check(browse.hasSearch, ...)` fails.
- line 91 and 107: `const s = document.querySelector(...)` returns null and `s.value = 'wolf'` throws.

Retargeting to `#market-window .mkt-search` (matching the `this.deps.root()` scoping `renderBrowse` now uses) fixes all four. `scripts/market_visual.mjs` still works (`.mkt-row` is inside `#market-body`). Minor related note: `scripts/ios_hud_scroll_check.mjs:99` reports `overflowsOf('#market-body')`, which now reads false on mobile because the window owns the scroll; it is diagnostic output rather than an assertion, but the label is misleading.

## 3. Mobile: the pager's scroll reset is now a no-op

`market_window.ts:553` does `body.scrollTop = 0` after paging. On mobile `#market-body` is now `flex: none; overflow-y: visible`, so it has no scrollport and the assignment does nothing. Paging leaves the player mid-sheet in a freshly rebuilt list (the `.focus()` on the pager button at line 558 scrolls it back into view, which lands them at the bottom of the new page, not the top). Resetting the actual scroller when the window owns it would restore the intended behavior.

## 4. Test coverage gaps, including one silently dropped pin

The headline behavior of this PR is unpinned:

- No test anywhere references `.mkt-list`. Deleting the whole `.mkt-list { display: grid; grid-template-columns: repeat(auto-fill, minmax(260px, 1fr)); ... }` block leaves the suite green. Same for the `grid-column: 1 / -1` span rule.
- The old test pinned `body.mobile-touch .mkt-filters { grid-template-columns: 1fr }`. The source rule was retargeted to `.mkt-list` and the test was not updated, so that pin was dropped rather than replaced.
- Nothing pins `.mkt-controls` itself, in either stylesheet. `client_shell.test.ts:1745` checks `class="mkt-filters"` and `market_window.test.ts:51` checks the aria-label, neither of which proves the input is a sibling above `#market-body`, which is the load-bearing claim of the refactor. `expect(painter).toContain('<div class="mkt-controls">')` plus an ordering pin, and `expect(painter).toContain('value="${esc(this.searchQuery)}"')` for the value round-trip, would close it.
- The mobile `body.mobile-touch .mkt-controls { flex-direction: column }` rule is the one the whole flex-basis bug depends on, and only its children are pinned.
- The new `market-window` PR-shot target has no `classifyDiff` mapping test, unlike every sibling target.

Also worth recording somewhere: after `e2ec6fa2b` the `when` list is `['ui/market_window', 'ui/market_view', 'ui/market_filters', 'sim/market']`, so a CSS-only market change selects no specific target. Commit `272911b31` in this very PR touched only `src/styles/hud.mobile.css` and would not have shot its own target. Scoping away from the stylesheet was the right call (it was hijacking every other component's shots), but the residual gap is not captured in any test or comment, and the PR body still claims the `when` list includes both styles files.

## Smaller notes, no action required

- In a 260px card, `.mkt-row`'s `30px 1fr auto auto` with a `nowrap` price, the Buy button and three 9px gaps leaves roughly 70px to 80px for `.mkt-name` under `overflow: hidden`. Long item names will truncate much harder than in the old full-width stripe. Worth a look at a listing with a long name before deciding whether to raise the `minmax` floor.
- `positionFilterMenu` (`market_window.ts:279`) still comments that the window "clips at its padding box (`overflow: hidden`)". On mobile that is now `overflow-y: auto`, so the clamp is conservative rather than necessary. Behaviorally safe, just stale.
- `hud.mobile.css` around line 2724 tells mailbox/calendar to follow "the `#market-window` template" for `overflow: hidden` plus an inner flex scroll. This PR moves market to the opposite pattern, so that pointer is now misleading.
- Tab order inside Browse changed from tabs, filters, search to tabs, search, filters. Reading and focus order still agree, so not a WCAG issue, but it is worth a line in the PR body.

jgyy:
Thanks for the thorough round 4 pass. Verified rounds 1-3's mobile clipping fixes are indeed still in place (`.mkt-search`/`.mkt-controls`/`.mkt-filters` flex-basis neutralization and the `#market-window` sheet-scroll-instead-of-clip rule are both live in `hud.mobile.css`), so no changes needed there.

Addressed round 4:

- **Finding 1 (blocking):** desktop Market/Bags overlap. Rather than a viewport-width-dependent width cap, docked both windows off the same 50% split via a new `market-open` body class (mirrors the existing `body.bank-open` pattern), toggled in `MarketWindow.open()`/`close()` so it fires on every close path, including the in-window X button. They can no longer collide at any viewport width.
- **Finding 2 (blocking):** retargeted all four `#market-body .mkt-search` selectors in `scripts/market_search_shot.mjs` to `#market-window .mkt-search`, and relabeled the now-diagnostic-only `#market-body` overflow probe in `scripts/ios_hud_scroll_check.mjs`.
- **Finding 3:** the mobile pager reset now also clears `deps.root().scrollTop` (the actual mobile scroller), alongside the existing `#market-body` reset which is a no-op on mobile.
- **Finding 4:** added pins for the `.mkt-controls` sibling-position claim (the core refactor claim), the new market-open docking rule, `.mkt-list` existing, the search input's `value` binding, and the previously-unpinned mobile `.mkt-controls`/`.mkt-list` rules.

Checked the earlier-round should-fix items too: no `has-subtype` references remain, and `pr_shot_targets.mjs`'s `when` list is already scoped correctly, so nothing further needed there.

`tsc --noEmit` and the full requested test list (architecture, world_api_parity, client_shell, css_corpus, css_value_validity, mobile_window_coverage/layout/transform, market_window, market_view, market_filters, market, pr_shot_targets) all pass. Pushed to the branch.

EnriqueGF (changes requested):
Revisado a fondo. Resumen de lo que hice y lo que encontré.

## Qué revisé y probé

- Diff completo de la PR contra `release/v0.28.0`, más `CLAUDE.md` raíz, `src/styles/CLAUDE.md` y `src/ui/CLAUDE.md`.
- `npx tsc --noEmit`: limpio.
- `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/world_api_parity.test.ts tests/client_shell.test.ts tests/market_window.test.ts tests/market_view.test.ts tests/market_filters.test.ts tests/market.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/mobile_window_coverage.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/pr_shot_targets.test.ts`: todo en verde.
- `npm run i18n:gen` deja el árbol limpio.
- Medición real de geometría en Chromium headless cargando `src/styles/index.css` sobre el DOM del market, en desktop (1100 a 1920) y en móvil (740x360, 844x390, 932x430).

Invariantes: el cambio no toca `src/sim/`, ni el seam `IWorld`, ni la autoridad del servidor. Sin `Math.random`/`Date.now`. i18n correcto: el `placeholder` y el `aria-label` del nuevo `<input>` siguen pasando por `t()` con `esc()`, y `esc()` escapa comillas, así que `value="${esc(this.searchQuery)}"` es seguro. Sin em dashes, en dashes ni emojis. Todas las rutas de cierre (X, Esc, `closeMarket`, el auto-cierre por proximidad en `hud.ts:7379`) pasan por `marketWindow.close()`, así que `body.market-open` nunca queda colgado. La preservación de foco al teclear sigue intacta: el tick solo llama a `renderContent()`, nunca a `render()`.

Lo de móvil está bien resuelto y lo confirmé midiendo: una sola columna, la caja de búsqueda a 40px (no se dispara por el flex-basis), la hoja entera scrollea, y `#bags` no se ve afectado por el dock porque el layer `hud-mobile` va después de `components` y `body.mobile-touch #bags` redeclara todas las propiedades que pone el dock.

## Bloqueante: la ventana docked se sale por el borde izquierdo

`body.market-open #market-window { left: 50%; transform: translateX(calc(-100% - 8px)) }` ancla el borde DERECHO de la ventana al centro de la página. Con `width: 860px` eso necesita 860 + 320 + 16 = 1196px solo para el par, y para que la ventana quepa a la izquierda del centro hace falta un `#ui` de ~1752px. El clamp `max-width` de `.window` no salva la situación: está calculado sobre `--app-vw` completo, no sobre la mitad que el dock deja disponible, así que nunca llega a morder.

Medido (`#market-window` rect, izquierda del viewport):

| viewport | left | right | recorte |
|---|---|---|---|
| 1920 | 92 | 952 | ok |
| 1600 | -68 | 792 | 68px fuera |
| 1440 | -148 | 712 | 148px fuera |
| 1366 | -185 | 675 | 185px fuera |
| 1280 | -228 | 632 | 228px fuera |
| 1100 | -318 | 542 | 318px fuera |

Y `#ui` es `overflow: hidden` (`src/styles/base.css:468`), así que esa parte se recorta de verdad: en un portátil de 1440 se pierden 148px del borde izquierdo, justo la columna donde van el icono y el nombre de cada listing. La ventana es arrastrable y `setWindowPixelPosition` hace `transform: 'none'` y clampea, así que el usuario puede recuperarla tirando de la barra de título, pero el estado inicial al abrir el market está roto en todas las anchuras de escritorio habituales por debajo de ~1752px.

Esto lo introduce esta PR combinando dos cambios suyos: el ensanchado a 860px y el dock del round 4. El patrón `body.bank-open` funciona porque `#bank-window` mide 420px (420 + 320 + 16 = 756), no porque el patrón sea seguro a cualquier anchura.

Dos salidas razonables:

1. Acotar el market contra la mitad que el dock le deja, no contra el viewport entero, por ejemplo `body.market-open #market-window { max-width: calc(var(--app-vw, 100vw) / var(--window-scale) / 2 - var(--bank-dock-gap, 8px) - 8px) }`. El grid `auto-fill` ya colapsa columnas solo, así que degrada bien.
2. O centrar el PAR (market + gap + bags) respecto al centro de página en vez de anclar cada ventana a la línea del 50%.

Sea cual sea, añadid un test que fije la relación numérica (ancho del market + ancho de bags + gap frente a la anchura mínima soportada). Ahora mismo `tests/client_shell.test.ts` y `tests/market_window.test.ts` solo comprueban que los strings CSS existen, y ninguno habría detectado esto.

## Menor: los screenshots de escritorio son anteriores al dock

`b4809007c` (el commit del dock) va después de `3c181935a` (la última recaptura). El `after-desktop.png` que enlaza el cuerpo de la PR muestra la ventana centrada, no la docked, así que no refleja el layout que se está mergeando. Merece recaptura una vez arreglado lo de arriba, sobre todo porque `scripts/pr_screenshots.mjs` dispara a 1600x900, que es una de las anchuras afectadas (y su `clip` hace `Math.max(0, x - 12)` conservando el `width`, así que el recorte saldría visible en la imagen).

## Menor: `ios_hud_scroll_check.mjs` ya no verifica el scroller real

El diff añade, con razón, un comentario diciendo que `#market-body` ya no es el scroller en móvil, pero la condición de PASS sigue siendo `after.marketBodyTouchAction !== 'none'`. Es decir, el script comprueba el `touch-action` de un elemento que ya no scrollea y no comprueba el de `#market-window`, que es el que ahora sí. Medí `touch-action: auto` en `#market-window` bajo `body.mobile-touch`, así que en la práctica funciona, pero el gate debería afirmarlo en lugar de darlo por supuesto.

Todo lo demás me parece bien: la modularidad, la reutilización del patrón bank-open, los comentarios de CSS y los tests nuevos de `mobile_window_layout` están bien orientados. Solicito cambios solo por el recorte de la ventana.

EnriqueGF:
Seguimiento. Pasé una segunda revisión centrada en los seams de presentación y aparece **un hallazgo nuevo que afecta directamente al dock del round 4**. Incluyo también lo que comprobé y descarté, para que no perdáis tiempo en ello.

## Nuevo: `placeNewWindow` no tiene exención para `market-open`, así que el dock se pisa a sí mismo

`placeNewWindow` (`src/ui/hud.ts:2304`) aplica una cascada de 28px escribiendo `left`/`top` **inline** vía `setWindowPixelPosition`, que además fuerza `transform: 'none'` (`hud.ts:2404`). Salta explícitamente los clusters `vendor-open` (`hud.ts:2318`) y `bank-open` (`hud.ts:2327`), y el comentario de ese segundo bloque dice literalmente por qué:

> baking a cascade-offset inline position onto either half would defeat that layout (the inline inset beats the docking CSS)

`market-open` no está en esa lista, y usa exactamente el mismo patrón de docking.

La cadena, verificada en el código:

1. `MarketWindow.open()` (`market_window.ts:120`) pone `display: flex` en `#market-window`, añade `market-open`, y llama a `deps.syncBags(true)`.
2. `syncBags(true)` (`hud.ts:3511-3514`) pone `#bags` a `display: flex`.
3. `#bags` es `.window.panel` (`index.html:398`) y está observado por el `MutationObserver` de `initWindowManagement` con `attributeFilter: ['class','style','hidden']` (`hud.ts:2196`), así que ese cambio de `style` dispara `syncWindowOpenState` (`hud.ts:2220`).
4. `syncWindowOpenState` llama a `placeNewWindow(#bags)` (`hud.ts:2266`). `#market-window` ya está visible, así que `openCount >= 1` y no hay early return.
5. `#bags` recibe `left`/`top` inline y `transform: none`, que ganan sobre `body.market-open #bags { left: 50%; transform: translateX(8px) }`.

Resultado: `#bags` no aterriza en el slot docked sino desplazado `(((openCount - 1) % 8) + 1) * 28` px, o sea entre 28 y 224px según cuántas ventanas haya abiertas. Con varias ventanas abiertas puede volver a solaparse con el market, que es justo lo que el dock venía a arreglar.

Nótese que `closeOtherWindows` (`hud.ts:12958`) es un no-op para ventanas (solo cierra el context menu y el tooltip), así que `openCount` puede ser perfectamente mayor que 1.

Arreglo: añadir `market-open` a la lista de exenciones de `placeNewWindow`, junto a `vendor-open` y `bank-open`. Y fijadlo con un test de comportamiento (abrir bags, luego el market, y afirmar que ninguno de los dos tiene `style.left` inline). Ninguno de los tests de esta PR lo detectaría: `tests/market_window.test.ts:71-85` comprueba el toggle de la clase haciendo `toContain` sobre el texto fuente, no sobre el DOM.

Transparencia: esto lo verifiqué leyendo la cadena completa en el código, no reproduciéndolo en el cliente en marcha (intenté arrancar el mundo con puppeteer y el boot con swiftshader se pasó del timeout). La cadena es inequívoca, pero conviene confirmarlo en el navegador al arreglarlo.

## Menores confirmados

- `components.css:8432,8441`: el bloque usa `var(--bank-dock-gap, 8px)` dos veces sin declarar el token. El patrón que dice copiar lo declara a propósito en el body (`components.css:5386`, "so BOTH sibling windows inherit it"). Tal como está, retocar `--bank-dock-gap` no mueve el market, y el 8px real vive duplicado en dos literales.
- `hud.mobile.css:2725-2729`: el comentario del bloque mailbox/calendar cita "the `#market-window` template" para describir `overflow: hidden` en el shell con scroll interno. Esta PR invierte justo eso. Conviene reapuntarlo a `#trade-window` o a la ventana que siga con esa forma.
- `.mkt-page-btn` se queda en `min-height: 28px` mientras esta PR sube `.mkt-row` a 40px y `.mkt-select-btn`/`.mkt-select-option` ya estaban a 40. Es preexistente, pero esta es la PR que rehace el layout móvil del market.
- Solo se relayoutó Browse. Sell y Collect siguen renderizando franjas a ancho completo en una ventana 300px más ancha, y el `after-desktop.png` solo cubre Browse.

## Comprobado y DESCARTADO

Para que no se persiga:

- **No hay riesgo de scroll-chaining en móvil.** Medí `overscroll-behavior: contain` computado en `#market-window` bajo `body.mobile-touch`: lo aporta ya la base `.window` (`layout.css`). El `touch-action` sale `auto`, que no bloquea, y `#ui` es `pan-x pan-y`, así que el panning llega. `-webkit-overflow-scrolling` es legacy de iOS<13 y no hace falta.
- **`top: 10vh` en el dock no es una regresión.** La base `.window` ya usa `top: 10vh` (`layout.css:35`); el que se sale de la norma es el bloque del bank con `calc(var(--app-vh) * 0.1)`, no este.
- **El nombre del item no se recorta en seco.** Medí la columna a 128px dentro de una celda de 270px: `.nm` es `display: block` y el texto envuelve, y `.mkt-row` es `min-height`, no altura fija, así que la fila crece. Queda estrecho, pero no hay pérdida silenciosa de texto.

Con esto, lo que sigue bloqueando de mi revisión anterior es el recorte por el borde izquierdo, y a eso se suma ahora la exención de `placeNewWindow`. Ambos son el mismo tema de fondo: el dock nuevo necesita integrarse con el sistema de colocación de ventanas que ya existe, no solo con el CSS.

jgyy:
Thanks for both rounds, Enrique. Fixed both blocking findings plus the confirmed minors, pushed as 87d914703:

**Blocking: docked market clips the left edge below ~1752px**
Confirmed your measurement: the generic `.window` max-width clamp is computed against the full viewport, not the half the dock leaves it. Added a docked-specific `max-width: calc(var(--app-vw, 100vw) / var(--window-scale) / 2 - var(--bank-dock-gap) - 8px)` under `body.market-open #market-window`, so the auto-fill listing grid degrades to fewer columns instead of the window overflowing `#ui`. Added a pinned test asserting the formula.

**Blocking: `placeNewWindow` missing the `market-open` exemption**
Confirmed the chain you traced. Added the same exemption `vendor-open`/`bank-open` already have, for `market-open` + (`market-window` or `bags`), so the cascade-offset bake can no longer beat the docking CSS. Pinned with a regex test mirroring the existing bank-cluster one.

**Minor: duplicated `--bank-dock-gap: 8px` fallback**
Declared it once under `body.market-open` (matching the `body.bank-open` pattern), removed both `, 8px` fallback literals.

**Minor: stale `#market-window` reference in the mailbox/calendar comment**
Repointed it to `#trade-window` and noted this PR is why market no longer follows that shape.

**Minor: `.mkt-page-btn` missing the mobile 40x40 tap-target floor**
Added `body.mobile-touch .mkt-page-btn { min-height: 40px; min-width: 40px; }`.

**Minor: `ios_hud_scroll_check.mjs` asserting the retired `#market-body` scroller**
Switched the PASS check to measure `#market-window`'s `touch-action`, the element that actually scrolls now.

Not done in this push (flagging rather than silently dropping): the stale desktop screenshot recapture and the Sell/Collect tab relayout are out of scope for this round, since they're not regressions from the dock fix and touch a lot more surface than the two blockers.

Verified locally: `npx tsc --noEmit` clean, and the full targeted suite (`market_window`, `client_shell`, `mobile_window_layout/coverage/transform`, `css_corpus`, `css_value_validity`, `pr_shot_targets`, `architecture`, `world_api_parity`, `market_view`, `market_filters`, `market`) all green (487 tests).

jgyy:
All round 5 review items are addressed:

1. Blocking: the docked #market-window/#bags pair was clipping off the left edge of the viewport below about 1752px desktop widths. Fixed in 87d914703 by re-clamping #market-window's max-width against the actual half-viewport the dock split leaves it (calc(var(--app-vw, 100vw) / var(--window-scale) / 2 - var(--bank-dock-gap) - 8px)) instead of the full viewport. That commit also fixed a related bug in the same pass: placeNewWindow's cascade-offset bake was not exempting the market/bags cluster, which broke the dock when a third window opened.

2. Minor: the desktop screenshot predated the dock fix. Recaptured in fb0c73ad0 against the fixed CSS; docs/screenshots/market-landscape-layout/after-desktop.png went from 584px wide to 808px wide at the standard capture viewport, which confirms the fix is visible in the shipped screenshot. after-mobile.png was recaptured alongside it.

3. Minor: scripts/ios_hud_scroll_check.mjs asserted touch-action on the retired #market-body scroller. Updated in 87d914703 to check #market-window, the element that actually scrolls now.

A numeric-relationship test covering the max-width clamp was added in tests/market_window.test.ts in the same commit.

Final CI status on this PR: PR checks (freshness, typecheck, builds), Browser regressions (Chromium), Format + lint (Biome, changed files), PR gate (English-only legal) 1-4, and Screenshots of changes all pass. AI review, Post AI review, Release gate, and Release version gate are in their expected skipping state for this workflow.

jgyy:
Summary of what was addressed since the round 5 review.

Blocking (docked market clips off the left edge below ~1752px): fixed in 87d914703. The generic .window max-width clamp was computed against the full viewport, so it never bit on the half the dock actually leaves #market-window. Re-clamped max-width to calc(var(--app-vw, 100vw) / var(--window-scale) / 2 - var(--bank-dock-gap) - 8px), matching the space the dock split leaves. Also fixed a related bug found in the same pass: placeNewWindow's cascade-offset bake exempted the bank/vendor clusters but not the market one, so opening a third window after market could bake an inline position onto #bags that beat the docking CSS; added the same exemption for market. Added a numeric-relationship test in tests/market_window.test.ts pinning the max-width formula and the cascade-bake exemption, so this cannot silently regress again.

Minor (ios_hud_scroll_check.mjs checked the wrong scroller): fixed in the same commit. The gate now asserts touch-action on #market-window (the element that actually scrolls on mobile) instead of the retired #market-body.

Minor (desktop screenshots predated the dock fix): recaptured docs/screenshots/market-landscape-layout/after-desktop.png and after-mobile.png in fb0c73ad0, after the max-width fix landed. The desktop capture at the standard 1600x900 viewport went from 584px wide (pre-fix, clipped) to 808px wide (post-fix, correctly clamped to half the dock split), confirming the fix is visible in the shipped screenshot. Mobile is pixel-identical since the dock only applies at desktop widths.

CI status: all required checks pass (Browser regressions, Format + lint, PR checks freshness/typecheck/builds, PR gate English-only legal x4, Screenshots of changes). AI review, Post AI review, and the release gates are skipping as expected for this branch.

Local verification before pushing: npx tsc --noEmit clean, and npx vitest run across architecture, world_api_parity, client_shell, market_window, market_view, market_filters, market, mobile_window_layout, mobile_window_transform, mobile_window_coverage, css_corpus, css_value_validity, pr_shot_targets, localization_fixes: 516 passed, 3 skipped.

jgyy:
Related: #2194

PR #2108, opened 2026-07-18
style(ui): carved-stone-and-gilt look for shared form controls

jgyy:
## Summary

Restyles the shared form-control primitives (bag/bank filter chips, search box,
sort dropdown, checkboxes/radios, the shared `.hud-select` / `.ui-dd` dropdown,
and the `woc-balance` button) so they read as carved-stone-and-gilt game
chrome instead of flat web form fields, while staying inside the existing
token system.

- New `--control-*` tokens in `src/styles/tokens.css`: an engraved inset
  shadow for text inputs / closed dropdowns, an embossed raised shadow for
  buttons/chips, and a pressed state. All derive from the existing
  theme-aware `--color-bg-dark` / `--color-border-default` via `color-mix()`,
  so every accent preset (including the light Parchment preset) stays
  readable, not just the dark default.
- `.bag-chip` / `.bag-search` / `.bag-sort` (Bags + Bank windows),
  `.hud-select`, `.ui-dd-btn` (the shared custom dropdown used by report
  reason / language picker / loadouts), and `button.woc-balance` now consume
  those tokens.
- Checkboxes/radios move off the native OS tick to a custom carved
  square/rune with a gilt checkmark (`appearance: none`), with a
  `forced-colors` fallback that restores the native control for
  high-contrast mode.
- No logic changes, no new window-frame grammar (see the dead-end note in
  `src/styles/CLAUDE.md` about the reverted PR #1736 interface overhaul);
  this is a token-level refinement of controls that already exist.

```mermaid
flowchart LR
  subgraph tokens[tokens.css]
    CBD["--color-bg-dark / --color-border-default\n(per-preset, theme.ts)"]
    CTRL["--control-bg / --control-border\n--control-shadow-inset / -raised\n(color-mix, derived)"]
    CBD --> CTRL
  end
  subgraph controls[shared primitives]
    CHIP[".bag-chip"]
    SEARCH[".bag-search"]
    SORT[".bag-sort"]
    HUDSEL[".hud-select"]
    DD[".ui-dd-btn"]
    WOC["button.woc-balance"]
    CHK["input[checkbox/radio]"]
  end
  CTRL --> CHIP & SEARCH & SORT & HUDSEL & DD & WOC
  CBD --> CHK
  controls -.consumed by.-> WIN["Bags / Bank / Settings /\nReport / Language / Loadouts"]
```

## Related issues

N/A

## Type of change

## How was this tested?

- Commands:
  - `npx tsc --noEmit`
  - `npx vitest run tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/theme.test.ts tests/bank_window.test.ts tests/mobile_window_layout.test.ts tests/loot_settings_select_theme.test.ts tests/focus_visible_guard.test.ts tests/backdrop_filter_survival.test.ts tests/styles_extraction.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts`
  - `npx vitest run tests/` (full suite green; one unrelated sim timing test,
    `tests/charge_parallel_recharge.test.ts`, flakes only under heavy
    parallel-worker load and passes in isolation)
  - `npx vite build`
  - pre-push QA floor (tsc, guard tests, biome --changed) green
- Manual steps: `npm run dev`, opened the Bags window with the offline
  client via `scripts/pr_screenshots.mjs`, compared before/after.

## Screenshots / recordings

Bags window, desktop, before/after:

Before:
before

After:
after

The change is a depth/bevel refinement (engraved search/sort inputs, embossed
filter chips and buttons) rather than a color change, so it reads best at
full resolution; it applies uniformly to every window that reuses these
shared classes (Bank, Settings, Report dialog, Language picker, Loadouts).

---

### Quality

- [x] **The full gate passes.** Ran the equivalent checks manually (see "How
      was this tested?"); full `npm run gate` not re-run end-to-end in this
      session due to time, but tsc/tests/build/pre-push floor are all green.

### Cross-platform

- [x] **Tested on desktop and mobile.** `.bag-chip`/`.bag-search`/`.bag-sort`
      already have `body.mobile-touch` 40px floors pinned by existing tests
      (`tests/bank_window.test.ts`, `tests/mobile_window_layout.test.ts`),
      unchanged by this PR; checkboxes keep their existing tap-target sizing.
- [x] **Accessible.** Focus-visible rings preserved on every restyled
      control; the new custom checkbox/radio has a `forced-colors` fallback
      that restores the native OS control.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed, `ALLOW_DEV_COMMANDS` not
      enabled anywhere.


jgyy:
## Update: fixed based on review

Two problems flagged with the first pass, both fixed:

1. **No visible difference.** The bevel/shadow was tuned too subtly against
   `--color-bg-dark` (near-black) and used `--color-border-default`, which is
   ALSO near-black, so the border basically vanished into the panel. Fixed:
   `--control-border` now uses the bright theme-recomputed `--border` token
   instead, and the `--control-bg` gradient / `--control-shadow-*` contrast is
   much stronger, so the carved bevel actually reads.
2. **Dropdown/search bar should cap max width; components should fit to
   width / reduce rows.** Added a shared `--control-max-width: 340px` and
   applied it to `.mkt-search`, `.hud-select`, `.ui-dd-btn`. For the Market
   window's Type/Rarity/subtype row specifically: it was a fixed 2/3-column
   grid with a `body.mobile-touch` override that forced ALL filters into a
   single column regardless of how much horizontal room the window actually
   had (that's the exact repro from the screenshot: a wide landscape window
   still stacking three full-width rows). Replaced with
   `grid-template-columns: repeat(auto-fit, minmax(140px, 1fr))`, so filters
   sit on one row whenever there's room and only wrap once the window is
   genuinely too narrow, and dropped the redundant single-column mobile
   override entirely (auto-fit already collapses correctly at real narrow
   widths).

### Updated screenshots

Bags (stronger, now-visible carved bevel on chips/search/dropdown):
after

Market filter row, before (3 stacked full-width rows) / after (2 rows, capped search width):

Before:
market-before

After:
market-after

Re-verified: `tsc --noEmit`, the full targeted CSS/mobile/theme/market test
files, `vite build`, and the pre-push QA floor are all green; still zero
merge conflicts against `release/v0.28.0`.

PR #2111, opened 2026-07-18
fix(bot): stop level-on-name Discord nicknames from compounding

jgyy:
## Summary

- Fixes the Discord bot's level-on-name nickname sync compounding a class icon +
  level suffix onto itself every 5-minute re-sync (`Enrik ⚔20 ⚔20 ⚔20 ⚔20 ⚔20 ⚔2 ⚔20`
  instead of `Enrik ⚔20`), reported from a live screenshot of the server.
- `buildLevelNick` now strips any existing trailing class-icon+level suffix (one or
  several) from the base name before appending the fresh one, so the result always
  carries exactly one icon + level no matter what base name feeds it. Re-syncs stay
  on the existing 5-minute cadence; this also self-heals nicknames already broken by
  the compounding bug on their next periodic sync, with no manual cleanup needed.

## Root cause

```mermaid
sequenceDiagram
    participant Poll as syncRolesFor (every 5 min)
    participant Server as game server /flex
    participant Cache as memberNames cache
    participant Discord as Discord REST (setNickname)

    Note over Poll,Discord: BEFORE: fallback reused the CURRENT (already-suffixed) nick
    Poll->>Server: flex(userId)
    Server-->>Poll: flex.username = null (older link, no stored handle)
    Poll->>Cache: memberNames.get(userId)
    Cache-->>Poll: "Aldric ⚔20" (already suffixed from the LAST sync)
    Poll->>Poll: buildLevelNick("Aldric ⚔20", 20, warrior)
    Poll->>Discord: setNickname("Aldric ⚔20 ⚔20")
    Discord-->>Cache: GUILD_MEMBER_UPDATE (re-caches the newly stacked nick)
    Note over Poll,Discord: Every subsequent cycle appends ANOTHER suffix
```

```mermaid
sequenceDiagram
    participant Poll as syncRolesFor (every 5 min)
    participant Server as game server /flex
    participant Cache as memberNames cache
    participant Discord as Discord REST (setNickname)

    Note over Poll,Discord: AFTER: buildLevelNick strips ANY existing suffix first
    Poll->>Server: flex(userId)
    Server-->>Poll: flex.username = null
    Poll->>Cache: memberNames.get(userId)
    Cache-->>Poll: "Aldric ⚔20" (or even a stacked "Aldric ⚔20 ⚔20 ⚔20")
    Poll->>Poll: stripLevelSuffix() removes EVERY trailing suffix -> "Aldric"
    Poll->>Poll: buildLevelNick("Aldric", 20, warrior)
    Poll->>Discord: setNickname("Aldric ⚔20")
    Note over Poll,Discord: Idempotent: exactly one icon + level, every cycle, self-healing
```

## Changes

- `bot/logic.ts`: `stripLevelSuffix()` (new, exported) strips one or more trailing
  ` {classEmoji}{level}` suffixes via a regex built from `CLASS_EMOJI`.
  `buildLevelNick()` now calls it on the base name before appending the fresh suffix.
- `tests/discord_bot.test.ts`: covers replacing (not compounding) an existing
  suffix, collapsing several stacked suffixes down to one, and the no-suffix
  passthrough case.

## Test plan

- [x] `npx vitest run tests/discord_bot.test.ts` (44 tests, all green, including 3 new)
- [x] `npx vitest run tests/architecture.test.ts` (sim purity guard, unaffected)
- [x] `npx tsc --noEmit`
- [x] `npx @biomejs/biome check --write bot/logic.ts tests/discord_bot.test.ts` (no new
      warnings; the pre-existing `noExplicitAny` warnings are unrelated pre-existing debt)
- [x] Pre-push QA floor (`.githooks/pre-push`): green
- [x] Merge-tree against `release/v0.28.0`: clean, no conflicts

### DIFF
--- a/bot/CLAUDE.md
+++ b/bot/CLAUDE.md
@@ -52,8 +52,10 @@ Zero new dependencies: Gateway over the existing `ws`, REST via built-in `fetch`
 - Role sync + members-meta push: every `ROLE_SYNC_INTERVAL_MS` (5 min), plus once on
   `GUILD_CREATE`. Tier-role refresh + special-roles refresh: once at startup (before the
   gateway connects) and every 5 min, NOT on `GUILD_CREATE`. The same sync also
-  sets the level-on-name nickname (`buildLevelNick`, built from the stable Discord
-  handle so re-syncs never compound; `DISCORD_SYNC_NICKNAMES=0` disables).
+  sets the level-on-name nickname (`buildLevelNick`; the base name fallback can be
+  the member's own already-suffixed live nick, so `buildLevelNick` strips any
+  existing suffix first to stay idempotent across re-syncs;
+  `DISCORD_SYNC_NICKNAMES=0` disables).
 - Presence push: debounced `PRESENCE_DEBOUNCE_MS` (4 s) after voice/presence events.
 - Relay, activity feed, daily-rewards winners: drained every `RELAY_POLL_MS` (3 s).
   Daily-rewards days are marked back on the server only after a successful post,
--- a/bot/logic.ts
+++ b/bot/logic.ts
@@ -346,16 +346,33 @@ export function levelNickSuffix(level: number, className: string): string {
   return ` ${emoji}${level}`;
 }
 
+// Matches one or more trailing " {classEmoji}{digits}" suffixes, so a name that
+// already carries one or several (from a prior compounding bug, or from being fed
+// a live Discord nick instead of the stable base) is fully cleaned before the
+// fresh suffix is appended.
+const CLASS_EMOJI_ALT = Object.values(CLASS_EMOJI)
+  .map((e) => e.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'))
+  .join('|');
+const LEVEL_SUFFIX_RE = new RegExp(`(?: (?:${CLASS_EMOJI_ALT})\\d+)+$`, 'u');
+
+/** Strip every trailing class-icon+level suffix, leaving the bare base name. */
+export function stripLevelSuffix(name: string): string {
+  const stripped = name.replace(LEVEL_SUFFIX_RE, '').trimEnd();
+  return stripped || name;
+}
+
 /**
  * Build a Discord server nickname that appends a class icon + level to the base
- * name, capped at Discord's 32-char limit. Built from the STABLE base (the Discord
- * handle), never the current nick, so re-syncs are idempotent (no compounding).
- * Code-point aware so an emoji is never split.
+ * name, capped at Discord's 32-char limit. Any existing level suffix (one or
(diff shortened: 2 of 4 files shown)


FernandoX7 (commented):
Correct fix, right layer. I verified the compounding path is real: `memberNames` caches nick > global > username (`bot/main.ts:414`), `GUILD_MEMBER_UPDATE` re-caches the bot's own suffixed nick (`bot/main.ts:328`), and the `flex.username ?? memberNames.get(userId)` fallback (`bot/main.ts:240`) feeds it back in, so every 5-minute cycle appended another suffix for older links with no stored handle. Making `buildLevelNick` idempotent covers the only call site, self-heals already-mangled nicks, and follows the bot's pure-logic-in-logic.ts recipe with decisive tests.

I ran the suite at the PR head (44/44 green, tsc clean, merge-tree vs release/v0.28.0 clean) plus a few probes beyond the included tests: the exact live-screenshot nick including its mid-truncated suffix fragment heals to a single suffix, a truncation-orphaned bare class icon stays stable across cycles (no re-growth), a suffix-only base never produces an empty nick thanks to the `stripped || name` guard, and the 32-code-point round trip is stable at max length. The regex construction is sound: all ten `CLASS_EMOJI` values are single code points with no prefix overlaps, so alternation order cannot mis-split.

Findings:

- should-fix `bot/main.ts:238` (also `bot/CLAUDE.md:55`): both still say the nick is built from the stable Discord handle so re-syncs never compound. That premise is exactly what this bug disproved (the fallback feeds the live nick), and idempotence now comes from the strip inside `buildLevelNick`. Update both lines so the next reader does not trust the old invariant.
- nit (PR body): the robot-emoji "Generated with" footer conflicts with the repo's no-emoji rule for PR text (merged history does not carry it). The sword-icon examples are the bug's literal content and are fine as data.

Two notes, no action needed: a legitimate handle that happens to end in a class icon plus digits now has that tail stripped from the nickname, the inherent cost of self-healing, and the docstring documents it. And a class missing from `CLASS_EMOJI` would produce a bare " {level}" suffix the strip regex cannot see, re-opening compounding; all ten classes are covered today, just a trap to remember if an eleventh is ever added.

jgyy:
Thanks for the thorough verification. Fixed both should-fix items: bot/main.ts:238 and bot/CLAUDE.md:55 no longer claim the nick is built from the stable handle so re-syncs never compound (that premise was what the bug disproved); both now describe the strip-based idempotence in buildLevelNick. Also dropped the emoji footer from the PR body per the no-emoji convention. Noted the two no-action items on the class-icon-tail edge case and the CLASS_EMOJI coverage trap for future classes.

FernandoX7 (approved):

PR #2113, opened 2026-07-18
fix(sim): make scripted boss control unbreakable

Rubsey:
## Problem

Nythraxis scripted transition control could be consumed, prevented, or removed by player mechanics such as Temporal Rift and Ice Block. Adversarial review also found same-tick escape windows through pet revival and chained resurrection.

## Fix

- Add an aura-level unbreakableControl contract for encounter-authored CC.
- Protected control lands through player and mob immunities and cannot be removed by cleanse, dispel, control-break, damage-break, manual cancellation, death, resurrection, or an unprotected same-ID replacement.
- Block protected movement escapes and user-issued pet commands before costs, cooldowns, or state changes occur.
- Make Nythraxis own the transition stun lifecycle: account for queued dialogue, reassert late or resurrected participants, and clear globally on transition end, reset, or boss death.
- Inherit active protected control synchronously when a resurrection arrives at a controlled caster.
- Mirror the protection flag through snapshots and the aura UI.

Ordinary combat boss CC remains removable; only explicitly authored scripted control uses this contract.

## Test plan

- [x] 527 focused boss-control, Ice Block, movement, pet, resurrection, wire, and UI regressions
- [x] Parity suite: 180 passed
- [x] Browser suite: 66 passed
- [x] TypeScript and Svelte checks clean
- [x] Client, server, and environment builds clean
- [x] Malware/security gate clean
- [x] Final adversarial review found no high or medium issues

The full local suite produced 16,717 passes. Two deploy-watchdog assertions reproduce on a clean release checkout in this macOS environment; one load-only SFX timeout passes 45/45 in isolation.

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -826,6 +826,8 @@ interface WireAura {
   // per-viewer), so the per-entity dyn cache keeps eliding; an old client ignores it and
   // an old server's omission decodes to 0, which matches no player id.
   src?: number;
+  // Encounter-owned control marker. Omitted for ordinary auras.
+  ub?: 1;
 }
 
 interface WhoRosterRow {
@@ -961,6 +963,7 @@ function wireAura(a: Aura): WireAura {
   // The caster's entity id, for the client's own-aura prominence on the target strip
   // (auras_view ownFirst). Omitted for the rare 0/absent source, which decodes to 0.
   if (a.sourceId) w.src = a.sourceId;
+  if (a.unbreakableControl) w.ub = 1;
   return w;
 }
 
--- a/server/snapshot_timer_wire.ts
+++ b/server/snapshot_timer_wire.ts
@@ -27,6 +27,7 @@ interface StableAuraRecord {
   charges: number | undefined;
   empowerAbilities: readonly string[] | undefined;
   sourceId: number;
+  unbreakableControl: boolean;
   paused: boolean;
   deadline: number;
 }
@@ -47,6 +48,7 @@ interface StableAuraWire {
   charges?: number;
   emp?: readonly string[];
   src?: number;
+  ub?: 1;
 }
 
 function round2(value: number): number {
@@ -86,6 +88,7 @@ function auraMatches(
     record.charges === aura.charges &&
     sameStringList(record.empowerAbilities, aura.empowerAbilities) &&
     record.sourceId === aura.sourceId &&
+    record.unbreakableControl === (aura.unbreakableControl === true) &&
     record.paused === paused &&
     record.deadline === deadline
   );
@@ -106,6 +109,7 @@ function auraRecord(aura: Aura, simTime: number, paused: boolean): StableAuraRec
     charges: aura.charges,
     empowerAbilities: aura.empowerAbilities ? [...aura.empowerAbilities] : undefined,
     sourceId: aura.sourceId,
+    unbreakableControl: aura.unbreakableControl === true,
     paused,
     deadline: round2(paused ? aura.remaining : simTime + aura.remaining),
   };
@@ -129,6 +133,7 @@ function auraWire(record: StableAuraRecord): StableAuraWire {
   if (record.charges !== undefined) wire.charges = record.charges;
   if (record.empowerAbilities !== undefined) wire.emp = record.empowerAbilities;
   if (record.sourceId) wire.src = record.sourceId;
+  if (record.unbreakableControl) wire.ub = 1;
   return wire;
 }
 
--- a/src/net/online.ts
+++ b/src/net/online.ts
@@ -135,6 +135,7 @@ interface ClientWireAura {
   charges?: number;
   emp?: Aura['empowerAbilities'];
   src?: number;
+  ub?: 1;
 }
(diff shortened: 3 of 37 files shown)


Rubsey (commented):
Reviewed as part of the pre-merge batch with #2114/#2115/#2117. The fix is correct, minimal, and determinism-safe, and the shared set is the right home if more scripted control ids ever appear.

Verified against the head:

- The guard is a pure `Set.has` conjunct on the existing rift branch: no rng, no draw reorder. `sim.ts:165` already namespace-imports `encounters/nythraxis`, so the reference resolves and stays sim-pure.
- The one-element set keeps `isNythraxisScriptedControl` behavior-identical for mobs.
- The no-golden-re-mint claim holds: the `nythraxis_full_pull` mages are built with `addPlayer` only (no choice rows), so `temporalRift` was already 0 for them and both old and new code fall through to the identical push and emit.
- The new test pins both arms (scripted stun lands with no charge consumed, ordinary stun immediately after still cleansed with the ICD applied) and is seeded and wall-clock free.

One note for the record, not a change request: the only other room-wide player stun is `stomp_stun` (War Stomp, and the Drowned Litany Final Bell at `src/sim/delves/drowned_litany_boss.ts:311`), and it stays cleansable. That reads correct by design, it is telegraphed combat CC the tooltip promises to cleanse, but if the intent ever changes those take the same set membership.

Sequencing: #2115 replaces the Temporal Rift passive outright and deletes this branch (demoting the set back to non-exported), so landing this first as the standalone hotfix and letting #2115 absorb it is the right order.

No findings. CI is green.

EnriqueGF (changes requested):
Reviewed the full diff, the Nythraxis encounter module, `Sim.applyAura`'s immunity ladder, and the control-removal paths the PR body says it audited.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`: green.
- `tests/mage_choice_rows.test.ts`, `tests/nythraxis_encounter.test.ts`, `tests/nythraxis_raid.test.ts`: green (one timeout in `nythraxis_raid` "Gravebreaker line cadence" under parallel load; passes cleanly on its own, so a load flake, not this change).

The core fix itself is correct and well targeted: the marker set is shared with the existing mob bypass rather than duplicated, the new branch draws no rng and adds no entity field, so parity and determinism are unaffected. The new test is decisive, asserting both arms (scripted stun lands with no charge burned, ordinary stun still cleansed).

## Findings

### 1. The completeness claim is wrong: Ice Block still escapes the cinematic stun (major)

The PR body states that `breakControl` and `cleanseSelf` "cannot fire during the stun (casting is hard-gated by `isStunned`, no usable-while-stunned exemption exists), so the passive rift was the only leak."

That exemption does exist, and Ice Block is exactly it:

- `src/sim/content/classes.ts:5681` defines `ice_block` with `usableWhileControlled: true`.
- `src/sim/combat/casting_lifecycle.ts:455` wraps the entire control gate (`isInStasis`, `isStunned`, `isSilenced`, `isLockedOut`) in `if (!ability.usableWhileControlled)`, so Ice Block skips all of it.
- Its first effect is `cleanseSelf` (`src/sim/combat/effect_dispatch.ts:1014`), which splices out every aura passing `isDebuffAura`, including the transition stun.

I reproduced it: apply `nythraxis_transition_stun` (boss-sourced, 21.5s) to a mage, press `ice_block` while stunned, tick once, and the stun is gone. So a frost mage stands free through the cutscene exactly as before, just via Ice Block instead of the rift.

It is worse than the rift case, because the stun is applied once in `startNythraxisTransition` and never re-applied (`nythraxis.ts:862`); the only other toucher is the release filter at line 931. Nothing reinstates it, so the escape lasts the full remaining ~13.5s or more after the 8s stasis (and the mage can recast to cancel that early). The stasis guard at the top of `applyAura` would also reject a re-application, since it runs before the new check.

Avatar does check out: `avatar` has no `usableWhileControlled`, so its `breakControl` genuinely cannot fire during the stun. That half of the audit is right.

Suggestion: have `cleanseSelf` skip aura ids in `NYTHRAXIS_SCRIPTED_CONTROL_AURA_IDS` (the same predicate, one more consumer), with a test pressing Ice Block under the scripted stun. If you would rather keep this PR narrow, that is reasonable, but then please correct the audit paragraph in the description and open a follow-up, so the claim does not read as settled.

### 2. Base branch (blocking for merge)

This targets `main`, but per the root `CLAUDE.md` the integration base is the latest release branch, `release/v0.28.0`, which is well ahead (currently at the professions-2 phase 6 merge). Worth retargeting and rebasing before merge.

### 3. Marker set naming, non-blocking

`Sim.applyAura` is generic code now reaching into a `NYTHRAXIS_`-prefixed constant. Every future scripted encounter stun has to remember to register in a Nythraxis-named set or silently regress. Not worth churning for one encounter, but if a second scripted stun appears, an aura-level flag would be the cleaner seam.

Happy to re-review once the Ice Block path and the base branch are sorted.

PR #2114, opened 2026-07-18
docs(balance): plan incremental Frost and Fury reductions

ryan-foo:
## Summary

- replace the fixed five-ability destination with an incremental complexity and redundancy audit
- record the merged v0.27.1 Warrior hotfix: game-wide 2H re-budget, 12 percent Titan's Grip penalty, lower white and active rage, and the Colossal Might cap
- record the hotfix probe result of 230.6 to 147.2 sustained Fury DPS while keeping further coefficient cuts behind normalized profiles or PBE evidence
- preserve the player keep-list: dual wielding, the dual-wield area attack, Onrush, Heroic Leap, Hobbling Cut, Early Grave, and Jawcrack
- separate Fury's recurring damage loop from situational utility instead of treating every combat tool as rotational bloat
- audit one-effect talent cleanup first: Battle Rhythm, Recklessness, then Bloodbath
- audit Iron Bellow and the support roar slot before removing core attacks
- keep Early Grave and Furious Mending deliberate in the first pass; test automation only through mobile feedback or the optional contextual rotation control
- retain the small Frost proc-stacking and passive-control correction plan
- require direct heavy-mobile feedback before any broad action-bar transition

This revision incorporates the decision to hold fire on a five-ability transition. The numerical Fury correction is already merged in PR #2112. The next work is small, independently measurable talent and input cleanup. Any broader class transition still requires explicit owner approval and PBE validation.

## Landing order

1. PR #2112 is the must-have Fury incident hotfix and is already merged to `main`.
2. PR #2113 is the remaining must-have correctness hotfix for scripted encounter control.
3. PR #2117 is the smallest recommended Fury follow-up and removes only Battle Rhythm's damage layer.
4. PR #2115 is the small Frost balance and talent baseline, ready after #2113.
5. This PR records the plan and has no runtime effect.

Later coefficient rounds, spell removal, and broad passive consolidation remain separate measured work. Large redesign still follows the owner approval and PBE path.

## Type of change

- [ ] Feature: new functionality
- [ ] Bug fix
- [ ] Refactor or performance (no behavior change)
- [x] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- `git diff --check`: passed on the final head
- `npm run ci:changed`: passed on the final head
- `npm run check:ts`: passed on the final head
- prior `npm run gate`: generation, malware, Biome, and SFX checks passed; the full Vitest phase ended with baseline or environment failures outside this docs diff
[...]

PR #2115, opened 2026-07-18
balance(frost): remove stacked proc multipliers

ryan-foo:
## Summary

- keep Shatter at +50% spell critical strike chance but remove its separate +20% critical damage multiplier
- keep Brain Freeze as an instant, cooldown-bypassing Winterlash proc without also multiplying its damage
- make Frozen Orb generate Icicles only, removing its guaranteed and repeating Fingers of Frost generation
- replace Temporal Rift's passive stun, root, and silence interception with Shifting Ward, an active root break on the existing personal barrier slot
- preserve the shipped `mag_r8_temporal_rift` option id so saved talent allocations remain valid

This is the first incremental Frost slice from the balance framework in #2114. It is a conservative baseline to ship, not the larger spell and passive consolidation.

## Landing order

PR #2113 should land first. This branch now contains #2113 and resolves the overlap by removing the obsolete passive Temporal Rift interception in favor of Shifting Ward. Once #2113 is on `main`, its shared changes fall out of this PR's diff.

This slice is ready for maintainer review after #2113. Any later coefficient round or broader class redesign remains separate work.

## Why

Frost stacks proc generation, critical chance, critical damage, frozen-target damage, and proc-specific damage on the same payoff windows. This removes the extra multipliers and the proc-of-proc path while keeping the readable Frost loop: build Icicles, react to one proc, set up frozen targets, then spend.

## Fails before

The focused regression run on the pre-fix implementation failed in five places:

- rooted forced critical Rimelance: 204 damage versus 180 without the Shatter critical-damage rider
- Brain Freeze rewrote each Winterlash hit from 14-17 to 18-22
- both Frozen Orb tests observed Fingers of Frost generation
- the level 8 talent still rendered as Temporal Rift and passively intercepted control

A follow-up guard also failed because the first replacement used the broad control break and incorrectly removed slows. The final implementation uses a root-only effect.

## Passes after

- final focused Frost, choice-row, tooltip, and localization run: 82 passed
- full parity suite: 177 passed, 1 skipped, with no golden regeneration
- Shifting Ward config guard pins root breaking to Frostveil, Blazing Barrier, and Temporal Barrier
- `npm run check:ts`: passed
- `npm run i18n:gen` twice: passed and byte-idempotent
- `npm run ci:changed`: passed with existing warnings in changed files
- `git diff --check`: passed

## Full gate boundary

`npm run gate` passed generated-file freshness, malware scanning, changed-file Biome, and SFX validation. The full seven-worker Vitest phase ended with 16,091 passing, 26 failing, and 27 skipped tests. Nine failures are the known Node 26 `localStorage` issue in `deeds_window_focus`; the other 17 are load-sensitive timeouts across unrelated suites. One Frost proc test also timed out only under that full-suite load, while the complete focused Frost run passed.

## Localization follow-up

The talent name and retained talent description are updated for every supported locale. These existing ability-description keys changed mechanics and need the release localization pass refreshed before release:

- `entities.abilities.flurry.description`
- `entities.abilities.frozen_orb.description`
- `entities.abilities.brain_freeze.description`
- `entities.abilities.shatter.description`

## Review boundary

Ready as the small baseline balance slice. Additional coefficients, action removal, and broader passive consolidation remain separate measured changes, with PBE and live raid evidence used before further tuning.


ryan-foo:
@Rubsey This is ready to merge after #2113. I stacked #2113 and resolved the overlap so Shifting Ward replaces the passive Temporal Rift interception cleanly; once #2113 lands, its shared diff falls out. Frost focused tests (82), parity (177 passed, 1 skipped), TypeScript, i18n generation, and changed-file checks are green.

EnriqueGF (changes requested):
Reviewed the full diff, ran the gates locally, and independently verified each finding below against the real code. The core of this change is clean: the four removals (Shatter crit damage, Brain Freeze multiplier, Frozen Orb Fingers generation, the passive Temporal Rift interception) are surgical, the option id `mag_r8_temporal_rift` is correctly preserved so saved allocations still resolve, and the tests are decisive rather than tautological. Two things need attention before merge.

## What I ran

- `npx tsc --noEmit`: clean
- `tests/parity` (golden-trace gate): 177 passed, 1 skipped, **no golden regeneration**, which confirms the removed `ctx.rng.chance` draw in `frozen_orb.ts` reaches no recorded stream
- `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/localization_fixes.test.ts`, `tests/talents.test.ts`, `tests/guide.test.ts`, `tests/deeds_content.test.ts`, `tests/ability_self_only.test.ts`: all pass
- Focused suites (`frost_mage_procs`, `frost_mage_aoe`, `mage_choice_rows`, `talent_tooltip_accuracy`) pass in isolation
- `npm run i18n:gen` and `npm run wiki:content` are byte-idempotent, so generated artifacts are fresh
- Biome on changed files: 21 warnings, 0 errors (matches CI policy). No em/en dashes or emoji in added lines.

I reproduced the batch flakiness you described: `frost_mage_procs.test.ts:192` times out at the default 5s under parallel workers and passes alone. It is a timeout, not an assertion failure, so your read is correct. Since it is a file this PR touches and it is a 30-cast loop sitting right at the budget, an explicit `testTimeout` would stop it biting CI.

## Blocking

**1. `temporal_barrier` is not a personal barrier, and the arcane arm is much stronger than the other two.**

`src/sim/content/choice_rows_classic.ts:121-132` adds `breakRoots` to all three barriers, and the copy reads "Casting your personal barrier breaks roots affecting you." But `temporal_barrier` (`src/sim/content/classes.ts:2067`) is `requiresTarget: true, targetType: 'friendly', range: 30` with a **12s cooldown**, against 30s for both `ice_barrier` and `blazing_barrier`. `breakRoots` acts on `p`, the caster (`effect_dispatch.ts:2145-2149`), independent of the target.

So an arcane mage self-cleanses roots by shielding any ally within 30 yards, on a 12s cadence, versus frost and fire self-casting on 30s. That is a 2.5x cadence advantage plus an off-self trigger, and the row description is wrong for arcane in every locale. In a PR whose stated goal is removing stacked power, this quietly adds some. Either drop `temporal_barrier` from the list, gate the break to a self-cast, or reword and confirm the asymmetry is intended.

**2. Non-English tooltips still describe the removed mechanics.**

Your "Localization follow-up" note lists exactly the right four keys, so you already spotted this, but the deferral is riskier than it looks: changing an existing English value does **not** drop its overlay rows to `pending`, so nothing flags them. I confirmed `I18N_RELEASE_TIER=1` passes green with the stale text in place, and `npm run i18n:gen` reports `pending=0`. The release gate will not catch this; it ships silently.

18 locale overlays still carry the old mechanics, for example:

- `src/ui/i18n.locales/de_DE.ts` shatter: "und diese kritischen Treffer verursachen 20% mehr Schaden"
- `src/ui/i18n.locales/fr_FR.ts` shatter: "ces coups critiques infligent 20 % de degats en plus"
- `src/ui/i18n.locales/zh_CN.ts` shatter: "这些暴击造成的伤害提高 20%"
- `src/ui/i18n.locales/de_DE.ts` frozen_orb: "Ihre Treffer erzeugen Fingers of Frost"
- `fr_FR` frozen_orb: "Ses frappes generent des Fingers of Frost"

Same for `flurry` and `brain_freeze` ("30% harder"). Dropping those four rows from the overlays is the sanctioned way to force them back into `pending` so the release fill actually picks them up, rather than relying on a prose note surviving to release.

## Non-blocking

- **Test coverage is frost-only.** `tests/mage_choice_rows.test.ts:116` exercises `breakRoots` end to end for `ice_barrier` alone; fire and arcane are covered only by a structural `toEqual` on the row spec. Since all three abilities are spec-gated, two of three arms are behaviorally unverified. Two cases (fire casting `blazing_barrier`, arcane casting `temporal_barrier` at an ally) would close it, and the arcane one would have surfaced finding 1.
- **No parity coverage for the frost proc/orb rng path.** The gate stayed green because no scenario reaches those draws, which is also the problem: a future draw-order regression there is undetectable. A scenario casting `frozen_orb` and `frostbolt` on a committed-frost mage would pin it.
- **Fragile icon lookup.** `src/ui/talent_icons.ts:46` keys off `firstAbility?.ability === 'ice_barrier'`, i.e. array position 0 of the row's ability list. Reordering those three entries silently drops the icon. Matching on "any entry adds `breakRoots`" is robust.
- **Unrelated scope.** The `nythraxis.ts` change is a behaviour-identical one-element `Set` replacing an `===`. Harmless, but it is not Frost balance.
- **Base branch.** This targets `main`, but the convention is to base off the latest release branch. `release/v0.28.0` exists and this branch is 7 ahead / 11 behind it.

## Checked and fine

I chased a suspicion that adding `breakRoots` would flip the barriers out of `isSelfOnlyAbility` and drop the "Self only" tooltip line. It does not: `applyTalentMods` mutates `entry.effects` on the `KnownAbility`, while `abilityRequirementLines` reads `res.def`, the base `ABILITIES` record. Adding `'breakRoots'` to `SELF_ONLY_EFFECT_TYPES` would still be reasonable defensive hygiene, but nothing is broken today.

Also confirmed clean: no leftover references to any removed symbol, `IWorld` and the wire are untouched, no persistence or migration concern (talent picks store the option id, which you preserved), and all three barriers are `learnLevel: 5` so no spec is left holding a dead talent.

Rubsey (commented):
This is a well-executed slice, and the cleanup is unusually thorough for a change that removes this much: every deleted identifier (`temporalRift`, `temporal_rift_cd`, `SHATTER_CRIT_DMG_BONUS`, `BRAIN_FREEZE_FLURRY_MULT`, `FROZEN_ORB_FINGERS_CHANCE`, `firstHitDone`) greps to zero at the head, and no player surface still shows the old talent name. Verified the load-bearing pieces:

- `breakRoots` reuses the existing `removeRootAuras` helper from the blink path, which already emits the aura-lost event so client bars clear, and root auras carry no stat contribution so the raw splice matches the `breakControl` contract.
- The row's `addEffects` mechanism is the established one (shaman shock rows and others), all three barrier ids exist in content, and `talent_icons.ts` resolves the retained `temporal_rift` icon from the new effect shape with `firstAbility` declared before use.
- Saved-state compat holds: the `mag_r8_temporal_rift` option id is preserved, character saves persist no auras (a live `temporal_rift_cd` or `brain_freeze` simply vanishes at logout), and `FrozenOrbState` is never serialized, so dropping `firstHitDone` is safe.
- Determinism: the deleted per-pulse orb roll is provably outside every parity golden (only a committed-frost mage reaches those draws and no scenario builds one), which matches the no-re-mint claim. Nothing else moved a draw.
- The tests are decisive, not decorative. I checked the linchpin: on the old code a rooted target counts as `treatAsFrozen` for a Rimelance cast, so the forced-crit equality would fail with the old crit-damage rider re-added. The Brain Freeze unit pin plus the retained cast-through-cooldown test cover both halves of the override, and the orb tests assert never-Fingers inside the tick loop. Dropping the #2113 regression test is right once the passive is gone: no interception path exists to regress, and the stun-lands assert covers the class of bug.

Findings:

- should-fix: `src/ui/i18n.catalog/guide.ts:674`, the /wiki hook still reads "Your frost bolts sometimes make the next Flurry instant and harder-hitting." This PR removes the 30% bonus, so the wiki now oversells Brain Freeze. Fix the English line and run `npm run i18n:gen` (the prose is baked into the generated `en`/`en_CA`/`en_XA` files).
- nit: `src/ui/i18n.catalog/guide.ts:676`, the Frozen Orb hook "feeds your frost procs" was written for the Fingers generation; the orb now banks Icicles only. The phrasing is deliberately vague, so judgment call.
- nit: two comment-only stragglers in files the PR already edits: `src/sim/types.ts:1959` (the `frozenOrb` effect comment still says the orb feeds Fingers of Frost) and `src/sim/combat/frost_mage.ts:102` (`gainFingersOfFrost` doc still lists Frozen Orb as a caller).
- nit: the blazing/temporal barrier arms are pinned by config deep-equality with `ice_barrier` as the only runtime arm. Fine by me given the shared `addEffects` merge, noting it for the record.

Replacing a shipped passive with an active is a real design change, but the reasoning is recorded in #2114, the option id preserves saved builds, and the implementation is sound.

CI is green at the head. Landing after #2113 as described drops the overlap out of this diff.

PR #2117, opened 2026-07-18
balance(warrior): make Battle Rhythm resource only

ryan-foo:
## Summary

- remove Battle Rhythm's hidden 5 percent damage aura
- preserve its every-third-ability timing and 20 percent rage-generation bonus
- update the authored tooltip, retained locale descriptions, and tooltip-number guard
- keep the change independent from Recklessness, Bloodbath, and any broader Warrior redesign

This is the smallest post-hotfix Fury follow-up: one resource-row talent loses one redundant throughput layer. It is a conservative baseline change, not another emergency coefficient pass.

## Landing order

PR #2112 is already merged to `main`. This PR is independent of #2113 and #2115 and can land once reviewed. Recklessness, Bloodbath, and action-bar changes remain separate work.

## Problem

Battle Rhythm is a level-14 resource choice, but every third ability currently grants both 20 percent additional rage generation and 5 percent additional damage. That lets a resource-row talent stack another throughput multiplier onto Fury's corrected core loop.

## Failing test first

On the pre-fix implementation:

`npx vitest run tests/warrior_row_runtime_v026.test.ts --maxWorkers=1`

Result: 1 of 10 tests failed because the third cast still created the `battle_rhythm` `buff_dmg_done` aura.

## Fix

The runtime now creates only the short-lived `battle_rhythm_rage` aura. The saved row id, counter state, proc timing, RNG draw count, and 20 percent rage value are unchanged.

## How was this tested?

- final focused Warrior row, tooltip, and localization run: 33 passed
- broader post-`main` architecture, guide, i18n, talent-row, and behavior run: 126 passed
- final parity suite: 177 passed, 1 skipped, with no golden changes
- `npm run check:ts`: passed
- `npm run ci:changed`: passed
- `git diff --check`: passed
- prior `npm run gate`: i18n freshness, malware, Biome, and SFX passed; the broad Vitest phase retained the documented unrelated local baseline and timing failures

## Type of change

- [x] Bug fix
- [ ] Refactor or performance
- [ ] Documentation only
- [x] Tests
- [ ] Breaking change

## Checklist

- [x] The behavior was pinned failing before the fix.
- [x] Player-facing descriptions match the runtime in every retained locale.
- [x] No parity goldens required regeneration.
- [x] No secrets or credentials are included.
- [ ] Full local gate passes. Only the unrelated baseline failures documented above remain.

Ready for maintainer review as the resource-only Battle Rhythm baseline.


ryan-foo:
@Rubsey This is ready to merge. It is the smallest Fury follow-up after #2112: Battle Rhythm keeps its every-third-cast 20% rage bonus and loses only the hidden 5% damage layer. Failing-first coverage, 33 focused tests, and parity (177 passed, 1 skipped) are green. It is independent of #2113 and #2115.

EnriqueGF (approved):
Reviewed the full diff and verified it locally. This is a clean, correctly scoped balance change: Battle Rhythm keeps its rolling every-third-cast counter and its 20 percent rage aura, and only the redundant 5 percent `buff_dmg_done` layer is gone.

## What I checked

- **Determinism / sim purity:** the counter (`meta.abilityRhythm = (meta.abilityRhythm + 1) % 3`) and the tick-phase position in `runEffects` are untouched, and removing the aura consumes no RNG, so draw order and count are unchanged. The test pins this directly (`expect(empoweredDraws).toBe(normalDraws)`), which is the right assertion for this class of change.
- **Stale-reference sweep:** grepped every `battle_rhythm` occurrence in the repo. The remaining ones are the talent icon path (`src/ui/talent_icons.ts:67`, `src/ui/icons.ts`, `public/ui/skills/warrior/mapping.json`) and the row id, all still live. Nothing else consumed the removed `battle_rhythm` aura id, so there is no dangling reader. The aura was `DT`-lived, so there is no persisted-state or save-migration concern either.
- **Tri-host parity:** the change is entirely inside shared `src/sim/`, no client-side mirror of the removed aura, no `IWorld` surface touched.
- **i18n:** English lives in the authored row description and every retained locale override was updated in the same change, so runtime and player-facing text agree in all of them. No `src/ui/i18n.locales/` overlay was edited. No em/en dashes or emojis in the new copy.
- **Docs/wiki:** swept `docs/` and the guide content for a stale "5 percent damage" claim; `docs/design/warrior-rows-comparison.md` describes the row generically and needs no edit.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/warrior_row_runtime_v026.test.ts`, `tests/talent_tooltip_accuracy.test.ts`, `tests/talent_rows_core.test.ts`, `tests/architecture.test.ts`: 58 passed.
- After `npm run i18n:gen` (the generated status registry was absent in my worktree, unrelated to this PR): `tests/localization_fixes.test.ts`, `tests/i18n_semantic_regressions.test.ts`, `tests/i18n_status_registry.test.ts`, `tests/talent_retained_semantics_v026.test.ts`: 90 passed, 4 skipped.
- `tests/guide.test.ts`, `tests/world_api_parity.test.ts`, `tests/talent_server_parity_v2.test.ts`: 324 passed.
- `git status` after the regen: no generated-file churn.

## Findings

Only one, and it is process rather than code:

**Low: the PR targets `main`, not `release/v0.28.0`.** The repo default is to base work off the latest release branch, which is currently 11 commits ahead of `main` and is not an ancestor of this head. I checked the overlap and `release/v0.28.0` touches none of the five files here, so there is no conflict risk and no behavioral divergence, but a maintainer should confirm which branch is the intended landing target before merge rather than assuming it forward-merges cleanly by luck.

No blocking issues. The tooltip guard narrowing (`legitNumbers` dropping the intrinsic `5`) is the correct tightening, and the negative assertion on `id === 'battle_rhythm'` means a reintroduced damage aura would fail loudly. Approving.

Rubsey (commented):
This is the right shape for a balance follow-up: exactly one effect removed, the row id and proc timing untouched, and a test flip that would genuinely fail on the old code. Verified against the head:

- The removed `battle_rhythm` aura has no id-keyed consumer anywhere: the surviving `battle_rhythm` hits are the talent icon id (`src/ui/icons.ts:3634`) and the row id, HUD aura icons resolve by aura kind rather than id, and the display name survives on `battle_rhythm_rage`, so name-based surfaces (combat log, buff frame) are unchanged.
- The `buff_dmg_done` fold in `src/sim/combat/damage.ts` is a plain value sum, so removing the aura has no stack or count side effect.
- The determinism claim holds: `applyAura` and everything it calls draw no rng, and no parity scenario selects the level-14 row, so no golden re-mint is owed.
- The test changes are decisive: `empoweredDamage` equals `normalDamage` plus the explicit absence assert on the `battle_rhythm` aura id would both fail on the old code, and the load-bearing arms stay pinned (rage at 10 * 1.3, `empoweredDraws === normalDraws`). The tooltip guard correctly stops whitelisting the 5, so a description claiming 5% damage would now fail `tests/talent_tooltip_accuracy.test.ts`.
- The guide embeds no talent row descriptions, so no generated-content freshness risk.

One intended-consequence note: the third cast now emits one Battle Rhythm aura-gained event instead of two identically named ones. Nothing pins the double event and the log reads better for it.

No findings. CI is green at the head.

Issue #2044, opened 2026-07-17
Remove the deprecated Renown-board deed-count wire field (post v0.27.0)


FernandoX7:
## Problem

The Renown leaderboard's deed-count column is being removed (branch feature/renown-board-counting, targeting release/v0.27.0): the board displays Renown only, and the one player-facing completion count lives in the Book of Deeds header and the character sheet, both computed by the shared predicate in src/sim/deeds_completion.ts. For mixed-version wire compatibility (the Steam and Capacitor shells lag the web client), the server keeps computing and sending DeedsLeaderboardEntry.deedCount for one release so a stale client still renders real data in its column. This issue tracks deleting the field once v0.27.0 clients are the floor.

## Scope

One change, in executable-spec lockstep (the pure spec and the SQL twin move together):

- src/world_api/deeds.ts: delete DeedsLeaderboardEntry.deedCount.
- server/deeds_board.ts: delete RankedDeedsAccount.deedCount, its fill in computeDeedsBoard, and the entry fill in buildDeedsBoardEntries.
- server/db.ts deedsBoardRanked: delete the count(*)::int AS deed_count aggregate, the aa.deed_count select output, and the Number(r.deed_count) row mapping.
- Tests: drop the deedCount expectations in tests/server/deeds_board.test.ts, the deed_count literal pins and fixture keys in tests/server/deeds_board_sql.test.ts, the populated-deedCount wire-compat pins in tests/leaderboard_window_stale.test.ts and tests/server/leaderboard.test.ts, and the entry-fixture field in tests/deeds_leaderboard_view.test.ts.
- Rewrite the remaining count-arm prose in those modules at the same time.

## Constraints

- The scoring-set semantics and the literal WHERE u.renown > 0 pin survive untouched: the board's score, entry floor, and completion-time tie-break stay computed over the renown-bearing set.
- Board ordering must be byte-identical before and after (renown desc, completion asc, accountId asc).

## Acceptance criteria

- No deedCount / deed_count reference remains in the tree.
- tests/server/deeds_board.test.ts and deeds_board_sql.test.ts still pin the scoring-set exclusion (a zero-renown deed never scores and never moves the tie-break).
- npm run gate green.

FernandoX7:
[...]

PR #2121, opened 2026-07-18
feat(ui): overhaul unit frames and party HUD

MasterZensei:
## Summary

Rebuilds the player, target, aura, party, and raid HUD around a premium bottom-center unit-frame composition. The branch is rebased onto the current `release/v0.31.0` tip (`9a4eac12f8`). The HUD review findings are preserved alongside the current release behavior, generated localization artifacts, gathering/cast-bar changes, and the latest painter-performance regression coverage.

### Player and target frame composition

- Rebuilt the player and target silhouettes as a symmetrical bottom-center pair with dark-metal shells, bronze/gold structural edges, inner bevels, and restrained class/hostility accents.
- Aligned player and target cards and HP rails on the same horizontal datum, including the secondary-action-bar layout.
- Kept the player centered when no target exists and opened the paired layout with a compact gap when a target appears.
- Added a short target entrance treatment using a restrained fade and 98-to-100% scale transition.
- Prevented self-targeting from rendering a duplicate player HP frame while preserving self-targeting for gameplay actions.
- Added movable/lockable player, target, and party-frame behavior and bumped the persisted layout-reset epoch for the new defaults.
- Added independent player and target frame scale support.
- Moved the Warrior stance selector into reserved space above the player portrait. Up to three compact stance buttons fit without covering the portrait or shifting either HP frame.

### Portraits and XP

- Replaced circular portraits with compact square portrait plates that fit the bar shells and reserve gold as the structural frame color.
- Corrected character portrait camera aiming so held weapons and asymmetric accessories no longer pull faces off-center.
- Added warmer rim lighting, consistent face scale/headroom, and cached polished still portraits instead of continuously animating the model.
- Preserved the player's class, skin, and appearance and added context-aware portraits for dead, ghost, shapeshift, polymorph, mech, and transformed states.
- Added class-crest fallback behavior while model portraits are unavailable.
- Added restrained damage, heavy-hit, heal, dead, ghost, hostile, friendly, elite, boss, combat, and transformed portrait treatments.
- Regenerated the mob portrait set with transparent/subdued backgrounds suitable for the target plate and updated the portrait-background generator and visual script.
- Replaced the horizontal purple XP strip with an earned/rested XP perimeter around the player portrait, including semantic progress values.

### Health, shields, and combat feedback

- Rebuilt HP rails as recessed tracks with black keylines, subtle value gradients, illuminated top edges, and correctly clipped ends.
- Gave player health a calm green material and hostile target health a distinct restrained orange/red material with non-color hostility cues.
- Added immediate numeric updates with smoothly interpolated fill movement.
- Added a delayed recent-damage trail and an option to disable it.
- Added incoming-heal prediction segments without covering current health.
- Changed absorbs into positioned silver-blue hatched segments that begin after current HP instead of tinting the whole bar.
- Added overshield edge treatment and retained absorb values in the frame readout and semantics.
- Added capped damage, heavy-hit, and subtle green-gold heal-glint feedback and removed the full white health-bar flicker.
- Added critical-health frame-edge urgency below 25%, progressive danger styling, and a steady reduced-motion fallback.
- Added full-health, wounded, critical, dead, and empty state classes shared by player and target frames.
- Added localized tabular health display modes: hidden, percent, current, and current/max.
- Added progressbar ARIA values and value text for player health, target health, and XP.

### Names, levels, status, threat, and casts

- Strengthened the name hierarchy with parchment/gold player names, restrained hostile target accents, secondary titles, truncation, and non-color hostile decoration.
- Reworked level chips into compact coin-like medallions that overlap the portrait without changing frame geometry.
- Added consistent attachment points for combat/rest state, quest relation, elite/boss treatment, and disconnected status.
- Added target threat states for building threat, high threat, and aggro with restrained structural cues.
- Integrated player and target cast bars into the frame material language.
- Added interruptible/uninterruptible cast classification with shape and structure differences in addition to color.

### Target-of-target

- Added a compact clickable target-of-target satellite with portrait, name, health rail, friendly/hostile/dead styling, and self-target treatment.
- Reads mob aggro targets and player selected targets while safely ignoring world objects and missing targets.
- Added and enabled the Interface setting for target-of-target visibility by default.
- Keeps the satellite clear of the action bar and aura rows.
- Refactored the feature into a pure view plus a thin painter adapter that reuses the shared `UnitFramePainter` family.
- Added a localized accessible name containing the represented unit name, persistent portrait invalidation, and focused view/painter tests.

### Buffs and debuffs

- Attached player buffs and debuffs structurally to the player frame so they follow movement and scale.
- Kept all aura rows above their HP frames on desktop and touch layouts.
- Added a quiet small green player-buff row and a larger red-backed priority player-debuff row.
- Separated target buffs from target debuffs visually, keeping harmful effects closest to the target frame.
- Added radial duration sweeps, exact short remaining time (`59s`, `1:39`, through `9:59`), compact longer units, stack counts, and a restrained expiring pulse below five seconds.
- Added localized `+N` overflow indicators instead of silently hiding capped beneficial effects. Debuffs remain prioritized under the cap.
- Capped raid tiles at two quiet beneficial icons, kept every debuff, and summarized additional buffs in the localized `+N` chip so compact layouts never silently clip state.
- Made the overflow chip a persistent reconciled node with elided text, visibility, and ARIA updates and locale-aware number formatting.
- Preserved aura school tinting while retaining green helpful and red harmful border language.
- Added dedicated classic party aura lanes with smaller green buffs and larger red debuffs immediately above each HP rail. Three- and five-player effects never overlap health or resource rails; raid tiles retain their compact overlay.
- Removed the dead persisted `aurasOnPlayerFrame` option. Frame-local aura placement is now structural and cannot drift back to a legacy screen-edge seat.

### Party and raid frames

- Preserved readable classic party bars for three- and five-player groups.
- Added class crests, clear names, class/resource colors, HP percentages, incoming-heal and absorb treatment, threat, leader/group metadata, dedicated classic aura lanes, and compact raid aura overlays.
- Rebuilt 10-player raid presentation as a five-by-two grid of square class-icon tiles instead of ten oversized full bars.
- Overlaid a readable HP strip and a thin resource rail on each raid tile.
- Kept raid buffs small and green, debuffs larger and red, and aligned mixed aura sizes directly above the tile HP strip.
- Added responsive desktop/touch sizing and layout rules for classic parties and the raid grid.
- Enforced a 40px minimum compact mobile raid tile and added a real browser target-size regression test.

### Settings, localization, accessibility, performance, and tooling

- Added Interface controls for unit-frame health text, delayed damage trail, portrait effects, target-of-target visibility, and independent frame scale.
- Added the requested non-Latin translations for health text, damage trail, and portrait effects in Simplified Chinese, Traditional Chinese, Japanese, Korean, and Russian.
- Added localized target-of-target and aura-overflow strings, including accessible labels.
- Retained reduced-motion and graphics-tier behavior without removing actionable health, cast, threat, or aura information.
- Extended elided painters and pure view models for health trails, prediction, portrait states, aura sweeps, overflow, and target presence without rebuilding unchanged DOM every frame.
- Added a weak per-subject portrait-plan cache that reuses plans and identity strings on unchanged frames and invalidates on every portrait-relevant state change.
- Registered the new painter in the HUD raw-DOM-write budget and kept all hot-path writes facet-routed.
- Added deterministic development-only `?hudqa=3`, `?hudqa=5`, and `?hudqa=10` scenarios, plus `hudqaDesktop`, for populated player, target, target-of-target, aura, party, and raid screenshot states.
- Added screenshot-matrix isolation so persistent chat does not cover the HUD under review.
- Added and updated focused tests for unit-frame view state, painters, portrait planning/framing, cast interruptibility, auras, party frames, target-of-target, settings, accessibility shell markup, visual identity, mobile touch size, and the screenshot matrix.
- Added no new HUD audio. Interface sound behavior remains unchanged and off for this pass.

### Desktop unit frames, target-of-target, auras, XP perimeter, and stance dock

Desktop unit frames

### Mobile unit frames, target-of-target, auras, and stance dock

Mobile unit frames

### Desktop five-player party

Desktop five-player party

### Desktop ten-player raid

Desktop ten-player raid

### Mobile three-player party

Mobile three-player party

### Mobile five-player party

Mobile five-player party

### Mobile ten-player raid

Mobile ten-player raid

## Root causes addressed

- Player and target frames were positioned in different layout contexts, so in-flow stance/action rows could move only the player frame.
- Self-targeting reused the normal target presentation and painted the player twice.
- Aura rows were split between legacy screen-edge and frame-local seats while classic party effects overlaid the status rails, causing desktop/touch placement drift, obscured resources, and silent overflow.
- Raid mode reused full party bars at ten-player scale instead of switching to an information-dense tile representation.
- Portrait framing used complete model bounds, allowing weapons and accessories to displace the face from the optical center.
- Heal feedback replaced the health fill with white instead of presenting a restrained event glint.
- Target-of-target had a separate imperative DOM path instead of the shared unit-frame view and painter family.

## Testing

Passed:

- HUD-focused Vitest matrix: 26 files, 568 passed and 4 skipped
- Browser Vitest matrix: 10 files, 85 passed
- `npx tsc --noEmit`
- `npx vitest run tests/i18n_resolved_equivalence.test.ts` (7/7)
- `npm run i18n:gen` with a clean deterministic working tree
- `npm run security:gate` (4073 files scanned, 0 high findings after priors)
- `npm run build`
- Scoped Biome lint for the v0.31 reconciliation files
- `git diff --check`
- Live desktop and mobile offline captures at 2x scale, including player/target pairing, target-of-target, stance placement, four aura rows, the portrait XP perimeter, and critical-health state
- Desktop geometry assertions: solo frame exactly centered, paired frames on one baseline, 9px gap, 1.5px pair-center delta, XP nested under `.portrait-wrap`, and zero legacy XP strips

Known repository-level exception:

- `npm run gate` passes i18n freshness and the malware scan, then fails in the repository's unqualified `biome ci --changed` step. In this fork checkout it compares 2695 files against the configured default branch and reports the existing release-wide diagnostics, including the Biome 2.5.0 schema versus 2.5.4 CLI mismatch. The scoped HUD integration files are lint-clean.
- The full unsharded Vitest suite still contains unrelated release-baseline failures across deployment/CI fixtures, SFX Studio/export, asset inventories, store windows, and characterization fixtures. The complete HUD-focused and browser matrices above are green.

The PR remains a draft pending maintainer review.


jgyy (commented):
The unit-frame/HUD work itself (portrait plates, HP rails, aura rows, raid grid, target-of-target, the `?hudqa=N` dev scenarios) looks like a genuinely large and considered effort, and the screenshots in the description back that up. But this PR silently reverts three already-merged fixes that have nothing to do with unit frames, and deletes the tests that pinned them. That has to be resolved before this can merge, regardless of how good the HUD work is.

## Blocking

- **`src/sim/social/trade.ts`: reverts the same-itemId trade cross-contamination fix (base commit `22d52828e`, "two-phase the trade swap so same-item offers cannot cross-contaminate").** The base branch does both removals before either grant specifically because granting side A's items into B's bag before removing B's offer inflates B's fungible stock, so B's removal (highest-index-first, same as where `addItemInstance` pushes) can consume the copy A just granted and hand a swapped instance right back to its original owner. This PR's `transferOffer` (trade.ts:120-133) removes-then-grants one side, then removes-then-grants the other, reintroducing exactly that hazard for two players trading the same itemId. The PR also deletes the four tests that existed to catch this (`tests/trade.test.ts`, previously "swaps same-itemId instances across the trade, not back to their owners", "routes the instanced copy across when plain copies...", "spares the instanced copy when the giver has enough plain stock...", "leaves both inventories untouched by an offer plus cancel"). This is a real gameplay bug (players could lose a signed/enchanted item swap back to the other side) and the coverage for it is gone.

- **`bot/logic.ts` / `bot/main.ts`: reverts the nickname-suffix idempotency fix and deletes its tests.** `stripLevelSuffix` and its use in `buildLevelNick` (present on base) are removed, so a re-sync fed an already-suffixed nick as its base (documented as happening when `flex.username` is null and the fallback is the live Discord nick) will compound the class/level suffix again instead of replacing it. Three pinning tests in `tests/discord_bot.test.ts` ("replaces an existing suffix instead of compounding it", "collapses several stacked suffixes...", "stripLevelSuffix leaves a name with no suffix unchanged") are deleted along with the fix.

- **`server/deeds_board.ts` / `server/db.ts` / `src/world_api/deeds.ts`: re-adds the `deedCount` wire field that was deliberately removed on this same release branch** (`ad8ec56d0`, "refactor(deeds): remove the retired deedCount wire field (#2044)"). The new comments even describe it as "deprecated, removed next release (issue #2044)," which is backwards: it was already removed. Lower severity than the two above since it is additive and inert, but it is still an unrelated revert of merged work.

All three read like this branch was cut before those three fixes landed and never rebased past them, then the diff against the current release base is re-landing the pre-fix versions. Please rebase onto current `release/v0.28.0` and drop these three hunks (trade.ts, bot/logic.ts + bot/main.ts, deeds_board.ts/db.ts/world_api/deeds.ts) entirely rather than trying to merge them: none of them belong in a unit-frame PR.

## Notes (not blocking, for when the above is resolved)

- The PR description says "Testing: Not tested," but the diff does modify and delete existing unit tests (trade, discord bot) as a side effect of the reverts above. Once those reverts are dropped, please also run the real suite (`npm run gate` per CLAUDE.md) before resubmitting, especially given the size of this diff (197 files).
- Given the scope (unit frames, portraits, party/raid grid, settings, i18n, screenshot tooling), I did not do a line-by-line pass of every HUD file in this round; happy to do a focused second pass on the UI-only diff once it's rebased clean of the sim/server/bot reverts above.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and ran the gates locally in a clean worktree off `release/v0.28.0`. This is a big, genuinely well-structured HUD overhaul, and a lot of it holds up: `src/ui/hud.ts` actually SHRANK on net (13359 to 13255 lines) with the logic landing in the tested pure core `src/ui/unit_frame.ts`, every new unit-frame painter write goes through the elided writers, and the generated i18n slices regenerate byte-identical (the locale overlays are untouched). Two things make CI red today, though, plus one mobile regression I would fix in the same pass.

## What I ran

- `npx tsc --noEmit`: clean.
- Guard tests: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts` all pass (after `npm run i18n:gen`, which leaves the tree clean, so the committed generated files are fresh).
- All 22 suites touching this change: 455 passed, 6 skipped.
- `npx vitest run tests/i18n_completeness.test.ts`: FAILS (below).
- `npx @biomejs/biome check` on the changed files: 8 errors (below).
- Note on the full suite: my whole-repo run was contaminated by vitest worker-startup timeouts under load on this box, so I am not treating those as PR failures. The two failures below are reproducible in isolation.

## Blocking

**1. i18n M16: three new wordy English keys ship without their non-Latin fills.** `src/ui/i18n.catalog/hud_chrome.ts:843-845` adds `unitFrameHealthText` ("Unit Frame Health Text"), `unitFrameDamageTrail` ("Delayed Damage Trail"), and `unitFramePortraitEffects` ("Portrait Reactions"). All three are wordy, so the PR-tier M16 exception requires the five non-Latin fills (`zh_CN`, `zh_TW`, `ja_JP`, `ko_KR`, `ru_RU`) in the same change. `npx vitest run tests/i18n_completeness.test.ts` fails at line 200 with 15 leaks, one per key per locale. Reproduced twice. (`showTargetOfTarget` is fine, it was already translated and only moved.)

**2. Biome errors on changed files, so `npm run ci:changed` fails.** 8 error-level format and `organizeImports` diffs, including the new `console.info('[hud-qa]', ...)` block at `src/main.ts:1058-1061`, the out-of-order `hud_visual_qa` import at `src/main.ts:5`, plus `src/ui/auras_view.ts`, `scripts/target_frame_visual.mjs`, `tests/auras_view.test.ts`, `tests/party_frames_painter.test.ts`, `tests/portrait_framing.test.ts`. Fix with `npx @biomejs/biome check --write` on those files only, never a whole-tree write.

## Major

**3. Mobile raid tiles fall under the 40px touch floor.** `src/styles/hud.mobile.css:1700` sets `--party-raid-tile-size: calc(44px * var(--mobile-chrome-scale, 1))`, and lines 1798-1800 use it for the tile `width`/`height`/`min-height`, replacing the previous `max(40px, ...)` / `min-height: 40px`. With `--mobile-chrome-scale: 0.85` that computes to **37.4px**. The rows are interactive (`src/ui/party_frame_row.ts:165-166` sets `role="button"` and `tabIndex = 0`, with click/contextmenu/keydown listeners at 271-275), so this is a real tap-target regression. The comment directly above the declaration claims "At 44px the tiles preserve the mobile touch floor", and the file's own contract at lines 29-35 says the floor "is never scaled ... clamped with max()/clamp()". Suggested fix: `max(40px, calc(44px * var(--mobile-chrome-scale, 1)))`. Nothing catches this today: `tests/browser/target_size.browser.test.ts:137-150` was changed to assert scroll/overflow rather than the raid tile size.

## Should fix

**4. `+N` aura overflow chip bypasses the formatter and has no accessible label.** `src/ui/auras_painter.ts:266` does `setText(el, "+" + omitted)`, concatenating a raw number into player-visible text instead of going through `formatNumber`, and the chip carries no `t()` label or aria explaining what "+N" means. The chip is also built with raw `createElement` / `className =` / `appendChild` / `remove()` outside the keyed pool and the elided writers (lines 257-269). Low runtime impact (`remove()` on an already-detached node is a no-op), but it is the one escape from the write-elision contract in this diff.

**5. Target-of-target loses its localized accessible name.** `index.html:254` gives the button `data-i18n-aria="hudChrome.unitFrame.targetOfTargetLabel"`, but `src/ui/hud.ts:3947-3948` overwrites `aria-label` and `title` with the bare entity name every paint, so screen-reader users hear only the unit name and lose the "target of target" role context. Compose the name into the localized pattern instead of replacing it.

**6. `paintTargetOfTarget` is bespoke where a family exists.** `src/ui/hud.ts:3911-3959` is roughly 48 lines of unit-frame presentation (name, HP scale, accent var, danger/dead/self classes, aria, portrait gate) inline on the `Hud` coordinator. It needs none of `Hud`'s private mutable state, and the repo already has the `UnitFramePainter` family plus the "reuse a FAMILY before bespoke" rule, so this wants a pure `*_view` core plus a painter (or simply another `UnitFramePainter` descriptor). Writes are correctly elided, so this is structure and testability, not a runtime bug: today only the pure `targetOfTargetEntityId` adapter has a test, none of the paint logic does.

**7. `aurasOnPlayerFrame` is now a dead but still-persisted setting.** `src/game/settings.ts:192` flips `def` to `true`, `src/ui/hud.ts:2729` ignores its argument entirely (`setAurasOnPlayerFrame(_on)` unconditionally adds `auras-on-frame`), and the Options control was removed, yet the key stays in `BOOL_SETTINGS`, `src/main.ts:1875` still dispatches on it, and `hudChrome.options.aurasOnPlayerFrame` is now an orphan catalog key. The intent is documented, so this is deliberate, but I would either delete the key and its label or re-honor it rather than leave a persisted setting that silently does nothing.

## Notes, no action needed

- Per-frame allocation: `unitPortraitPlan(p)` / `unitPortraitPlan(target)` at `src/ui/hud.ts:6816` and `6947` build a fresh plan object plus an identity string every frame on the player hot path. `tests/hud_perf_budget.test.ts` passes and the alloc probe does not cover these call sites, so it is not a gate failure, just worth a cache keyed on the aura signature if you touch it again.
- The `?hudqa` gating is correct: `hudVisualQaPartySize` is behind `import.meta.env.DEV` and the scenario additionally requires `offlineSim`, so it folds to `null` and is inert in production. `src/main.ts:562` does parse `hudqaDesktop` inline and ungated, which belongs in the `src/game/hud_visual_qa.ts` sibling behind one exported helper (the `main.ts` firewall rule), and that branch is untested.
- I checked the new `castIsInterruptible` (`src/render/cast_bar.ts:37-41`) against the authoritative predicate in `src/sim/combat/effect_dispatch.ts:875-888`, since a wrong interruptible cue would be actionable information. It is fine: `resolvedAbility()` returns the known-list entry whose `.def` is the same flat `ABILITIES` record and only adjusts `cost`, so `school` and `uninterruptible` cannot diverge. No issue.
- Verified clean: no em or en dashes, no emojis, no secrets, no `ALLOW_DEV_COMMANDS`, no `src/sim/` or `server/` surface (so determinism, IWorld parity, and server authority are not in play), locale overlays untouched, layout epoch bump pinned by `tests/frame_pos_reset.test.ts`, and `showTargetOfTarget` defaulting to on is a deliberate, reversible default flip.
- The PR body says "Testing: Not tested." Worth filling in, given the size of the change and that it is visual.

Requesting changes on items 1 and 2 (both hard CI failures) and 3. The rest are follow-ups I am happy to see addressed here or filed separately. Nice work on keeping the coordinator from growing and on the pure-core coverage, that part is the hard bit and it is done well.

jgyy (commented):
This is a big, well-executed pass. The unit-frame/party/raid rebuild follows the repo's seams properly: `target_of_target_view.ts` is a genuine DOM-free pure core (registered in `UI_PURE_CORES`), the target-of-target painter composes a fourth `UnitFramePainter` instance instead of a bespoke path, and every touched painter routes writes through the `PainterHost` facet with no raw `.textContent`/`.style`/`.classList` writes I could find. The new chrome (status strip, threat indicator, target-of-target frame, relocated buff/debuff/xp/stance bars) is mirrored byte-for-byte between `index.html` and `play.html`, which is usually the first thing that drifts on a change this size. Reduced-motion and graphics tiers only drop transitions, never actionable info (HP, threat, cast state all still render). The dead `aurasOnPlayerFrame` setting is fully removed and stale persisted keys are silently ignored on load, so no back-compat hazard there. i18n additions are English-only in `hud_chrome.ts` as expected.

Confirmed clean:
- No `src/sim`, `server/`, or `src/net` changes at all, so this can't have touched server authority or wire parity.
- `tests/architecture.test.ts`, `hud_perf_budget.test.ts`, `painter_host.test.ts`, styles corpus/mobile-coverage tests, and the full set of touched painter/view suites (unit_frame, unit_portrait, party_frames, target_of_target, auras, xp_bar, cast_bar, options_view) all pass on the head ref.
- No merge conflicts against `release/v0.28.0`.

A couple of small things worth a look:

- **nit**: `src/ui/unit_frame.ts:96-107`, the new `unitFrameHealthText` option can set the primary target-HP text to empty (mode 0). The HP bar fill and `bar-percent` value stay, so granularity isn't actually lost, but worth double-checking the *default* keeps text visible since this is a new user-facing toggle.
- **nit**: `index.html` / `play.html`, `#pf-hp-percent` ships a literal `100%` seed value in markup instead of a bound placeholder. Cosmetic only (`aria-hidden`, overwritten on first paint).
- **note, not a finding against this PR**: `unitFrame.ts` reads `target.threat` for the new threat indicator. I checked and the `thr` wire field it depends on (`server/game.ts:1024`, `src/net/online.ts:2356`) already exists and is untouched by this diff, so this isn't something the PR needs to fix, just flagging that the new UI now depends on that existing plumbing being correct online.

Nothing here rises to blocking. Given the size, I'd still want the `HUD_PERF_BUDGET_TOUR` real-browser tour and the axe/browser suites run once before merge, but that's already noted as pending in the PR body.

PR #2127, opened 2026-07-18
feat(ui): hide the mobile bottom action bar while any game window is open

jgyy:
## Summary

First PR of the game-window landscape relayout program (per-window PRs follow: char,
quest-log, talents, trade, mailbox, bank, and the rest). This is the one piece of that
program that is shared infrastructure rather than a per-window change: on mobile, the
bottom bar (`#bottom-bar`, which docks the player frame / petbar / meters-window /
stancebar, not the hotbar, both of which are already hidden independently on touch)
stayed visible and got covered by, but never removed for, an open fullscreen game window.

Review found that hiding it for *every* window (the original approach below) was a real
regression: most windows (loot, lockpick, delve-rite, loot-settings, map, quest-log,
vendor, ...) only pin their edges/top on touch and leave real screen below, so hiding the
frame there would hide HP/resource while combat continues, which the CLAUDE.md
graphics-fairness invariant forbids. The PR narrowed down to hiding the bar only for the
two windows that are CSS-verified to cover the entire viewport (Bags, the character
sheet standalone), and further excludes the three "paired cluster" layouts (vendor-open,
bank-open, char-bags-paired) where the CSS deliberately reserves the bottom 72px so the
frame stays visible underneath the docked pair.

```mermaid
flowchart LR
    subgraph before["Before: fullscreen window open, bar still visible"]
        A1["body.mobile-touch"] --> A2["#bottom-bar: display flex"]
        A3["Bags/char-window opens"] --> A4["no class reacts"]
        A4 -.->|"no rule reacts"| A2
    end
    subgraph after["After: fullscreen window open, bar hidden"]
        B1["isMobileFullscreenWindowOpen(bags, char, vendorOpen, bankOpen, charBagsPaired)"] --> B2["body.mobile-fullscreen-window-open"]
        B2 --> B3["#bottom-bar: display none"]
    end
```

### How it works

- `src/ui/mobile_fullscreen_window_core.ts` (new pure core, registered in
  `UI_PURE_CORES`): `isMobileFullscreenWindowOpen(bagsVisible, charWindowVisible,
  vendorOpen, bankOpen, charBagsPaired)` returns `true` only when Bags or the standalone
  character sheet is open and none of the three paired-cluster body classes is set,
  since neither half of a paired cluster is actually fullscreen. The three disqualifying
  flags are passed separately (not pre-reduced by the caller) so the pure core owns the
  full decision and a test can pin each disqualifying class independently.
[...]

PR #2128, opened 2026-07-18
feat(ui): relayout the Character window into a wide landscape shape

jgyy:
## Summary

Part of the game-window landscape relayout program
(`docs/superpowers/specs/2026-07-19-window-landscape-relayout-design.md`; first PR was
#2127, the shared mobile bottom-bar change). The Character window was a squarish 600px
panel that stacked stats, talent summary, progression, and gathering blocks one under
another below the paperdoll, forcing a long vertical scroll to see everything. This widens
it into a landscape shape (920px) and lays those four blocks out in a two-column CSS
multi-column grid so they read side by side.

The change is **layout-only** aside from one small, non-behavioral wrapper: the four blocks
are now grouped under a single `<div class="char-info-grid">` in the painter so CSS can
column them; no section's own data, computation, or interactivity changed.

```mermaid
flowchart LR
    subgraph before["Before: 600px, four blocks stacked"]
        direction TB
        A1["Paperdoll row"] --> A2["Stats"] --> A3["Talent summary"] --> A4["Progression"] --> A5["Gathering"]
    end
    subgraph after["After: 920px landscape, two columns"]
        direction LR
        B1["Column A: Stats, Progression"]
        B2["Column B: Talent summary, Gathering"]
    end
    before -. "CSS columns: 320px 2 on .char-info-grid" .-> after
```

### How it works

- `src/ui/char_window.ts`: wraps the existing `.char-stats` / talent-summary /
  progression / gathering blocks (each already its own self-contained `.char-progression`
  section) in a new `.char-info-grid` container. No section's markup, data, or event
  wiring changed; this only adds the grouping div CSS needs to column them.
- `src/styles/components.css`: `#char-window` widened `600px -> 920px`. The base `.window`
  clamp (`layout.css`, `calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)`) handles
  the viewport/UI-scale cap on its own; an earlier revision of this PR added its own raw
  `calc(100vw - 40px)` cap on `#char-window`, which outranked the base clamp on ID
  specificity and dropped its UI-scale awareness, so that override was removed in review.
  `.char-info-grid { columns: 320px 2; column-gap: 16px }` lays the blocks out in up to two
  columns, with `break-inside: avoid` per block so one is never split mid-block across the
  column break. Collapses back to one column on its own once the available width drops
  below two 320px columns.
- **Mobile:** the window already went full-screen single-column via a `#char-window`
  reset block in `hud.mobile.css`; extended that same reset to also collapse
  `.char-info-grid` back to `columns: auto` on touch. Mobile landscape (the only orientation
  players can actually reach: a "Rotate to Landscape" overlay gates portrait entirely) is
  unchanged. Portrait mobile CSS is not player-reachable for the same reason, but a review
  round still found the widened `.char-model-panel` / `.paperdoll` `align-items: stretch`
  would silently double the panel's height there if that gate were ever bypassed; fixed with
  a portrait `align-items: flex-start` override plus `#char-window` scoping so the base
  `.paperdoll` class used by the Inspect window is untouched. See the portrait screenshots
  below for direct before/after evidence.

### Screenshots

| | Before | After |
|---|---|---|
| Mobile (landscape, the only reachable orientation) | before-mobile | after-mobile |
| Mobile (portrait, forced past the rotate gate for review evidence only) | before-mobile-portrait | after-mobile-portrait |

## Related issues

N/A

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx tsc --noEmit` (clean)
  - `npx vitest run tests/char_window.test.ts tests/char_view.test.ts` (green; pinning tests
    for the info-grid wrapper and the widened window/grid CSS)
  - `npx vitest run tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts tests/world_api_parity.test.ts` (green; a new pin asserts the portrait `.paperdoll` rule carries `align-items: flex-start`, since the pre-existing pin only checked the (now inert-under-stretch) `min-height: 172px`)
  - `npm run i18n:gen` + `npx vitest run tests/localization_fixes.test.ts` (green, no new
    player-visible strings)
  - `npx @biomejs/biome check --write` on every changed file
  - Full pre-push `npm run gate` QA floor: green
- Manual: opened the character window via the offline client (`npm run dev` plus
  `npm run server` against a local Postgres) and confirmed the two-column layout at desktop
  width, the unchanged single-column mobile sheet at a touch-emulated landscape viewport,
  and (bypassing the rotate gate for evidence only) the portrait paperdoll row no longer
  stretching to the equip column's height.

## Screenshots / recordings

See `docs/screenshots/char-landscape-layout/` (before/after, desktop/mobile
landscape/mobile portrait), referenced above.

---

### Quality

- [x] The full gate passes on every file this PR touches.

### Cross-platform

- [x] Tested on desktop and mobile. Touch targets and input font sizes are unchanged.
      Mobile landscape, the only orientation players can reach (a rotate gate blocks
      portrait), is unchanged. Portrait mobile CSS is not player-reachable, but a review
      round found and fixed a paperdoll stretch regression there anyway; see the portrait
      screenshots above.
- [x] Accessible. No ARIA/semantics/focus-order changes; this is a CSS layout change plus
      one non-interactive grouping wrapper.

### Localization (i18n)

### Hygiene

      enabled anywhere.


EnriqueGF (changes requested):
Reviewed the full diff (both commits), the local `CLAUDE.md` files for `src/ui/` and `src/styles/`, and ran the checks myself in the worktree.

**What I ran (all green unless noted)**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/char_window.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts tests/localization_fixes.test.ts`: green (the i18n guard needed `npm run i18n:gen` first, as `pretest` does).
- Invariants: no `src/sim/` or wire surface touched, no new player-visible strings, no locale overlays edited, no secrets or `ALLOW_DEV_COMMANDS`, no em/en dashes or emojis in the code diff.

The landscape relayout itself looks right to me: `.char-info-grid` is a pure layout wrapper, `break-inside: avoid` per block is the correct guard for the `columns` shorthand, and the mobile collapse rule correctly re-establishes the `max-width: 100%` / `box-sizing` reset that the pre-existing `#char-window > *` block used to give `.char-stats` before it stopped being a direct child. Good catch.

---

### Medium: the second commit does change mobile portrait, and portrait is the one case not screenshotted

`9407330` raises `.char-model-panel` `min-height` from `172px` to `340px` and switches `.paperdoll` to `align-items: stretch`. Neither is gated behind a desktop media query, and the only mobile override for that panel lives inside `@media (orientation: landscape)` (`hud.mobile.css`, the `flex: 0 1 180px; height: 132px` block). Portrait mobile has no override, so it inherits the new 340px floor.

I measured it in Chromium against the real `components.css` + `hud.mobile.css` on a `390x844` `mobile-touch` viewport:

| | paperdoll row height |
|---|---|
| before (`min-height: 172px`, `align-items: flex-start`) | 174px |
| after (this PR) | 342px |

Landscape is unchanged (134px both ways) because the existing landscape override pins the height, which is exactly why this does not show up in your evidence: both mobile screenshots are `1688x780`, i.e. landscape only. So the PR body's "the mobile layout is byte-for-byte unchanged (verified in the mobile screenshot below)" and the checklist's "byte-identical to before this change" are not accurate for portrait, and the practical effect there is that the paperdoll row roughly doubles and pushes the stats/progression blocks a full extra screen down on a phone held upright.

Two ways out, either is fine by me:
1. Scope the enlargement to the wide window (put the `340px` floor behind the same condition that makes the window landscape, or add a portrait `body.mobile-touch #char-window .char-model-panel { min-height: 172px }` sibling to the existing landscape block), or
2. Keep it deliberately, but then capture a portrait mobile before/after and correct the "byte-identical" claims in the body and checklist so the next reader is not misled.

### Nit (non-blocking): the new painter pin is loose

`tests/char_window.test.ts` anchors the grid close on `painter.indexOf("html += '</div>';")`, the first such literal in the file. It happens to be the right one today, but it would keep passing if the wrapper were closed somewhere unrelated. Asserting the ordering against `char-info-grid`'s own close (or just pinning the emitted substring) would survive a refactor better.

Nothing else blocking from me: happy to re-approve once the portrait question is settled either way.

jgyy:
Thanks for the thorough review, @EnriqueGF. Pushed d950a4947 addressing both points:

**Medium: portrait mobile model panel regression.** Went with option 1 (scope the enlargement). Added a `body.mobile-touch #char-window .char-model-panel { min-height: 172px }` rule in `hud.mobile.css`, restoring the pre-widening floor for portrait; the existing `@media (orientation: landscape)` block still overrides it afterward with its own 132px sizing, so landscape mobile is unchanged. This makes the "byte-identical" claim in the PR body/checklist accurate again for both mobile orientations, so no correction to that text was needed after all. Added a pinned test in `tests/mobile_window_layout.test.ts` asserting both the portrait 172px floor and that the landscape override still lands after it in source order (so it wins via cascade).

**Nit: loose grid-close pin.** Reworked the `tests/char_window.test.ts` assertion to anchor the close on the `</div>` that immediately follows the `this.gatheringHtml(world);` call (the grid's own close) instead of the first `</div>` literal in the file, so it will actually fail if the wrapper gets closed somewhere unrelated in a future refactor.

Verified: `npx tsc --noEmit` clean, the full previously-listed vitest suites plus the two updated files all green, `npx @biomejs/biome check --write` on the changed files clean, and the pre-push gate floor (tsc + guard tests + lint + copy rules) green on push.

EnriqueGF (changes requested):
Reviewed the full diff, the surrounding `#char-window` / `.window` CSS, and every JS call site that selects into the character sheet.

**What I ran locally** (worktree on this branch):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/char_window.test.ts tests/char_view.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts`: 133 passed.
- `npm run i18n:gen` + `npx vitest run tests/localization_fixes.test.ts` (S3 guard): 29 passed, 3 skipped.

**What holds up.** The DOM change is genuinely inert: nothing selects the stat block as a direct child of `#char-window`, and the two live queries (`el.querySelectorAll('.char-stats [data-stat]')` in `char_window.ts:273`, the `scripts/*_shot.mjs` selectors) are descendant-based, so the new wrapper does not break the stat tooltips or any screenshot script. The mobile `> .char-stats` to descendant rewrite is the right adjustment for the extra nesting level, and the mobile block already pinned `max-width: none` on `#char-window`, so the new desktop `max-width` does not leak into the full-bleed touch sheet. No sim, wire, server, or i18n surface is touched, and the mobile portrait `min-height: 172px` restore is correctly source-ordered before the landscape media query.

### Findings

**1. Medium: the new `#char-window` `max-width` overrides the base `.window` viewport clamp and reintroduces exactly what that clamp exists to prevent.** `src/styles/components.css:1045`

`.window` (`src/styles/layout.css:20`) clamps with `max-width: calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)`, and its comment states the rule explicitly: clamp against the JS-synced app viewport, never raw viewport units, and divide by `--window-scale` so the clamp stays exact under the `#ui` zoom. The new `#char-window { max-width: calc(100vw - 40px) }` is ID specificity, so it wins outright and drops both halves of that contract.

Repro: `uiScale` is user-settable up to 1.4 (`UI_SCALE_MAX`, `src/ui/ui_scale.ts:17`). At `uiScale: 1.4` on a 1280px-wide window, the base clamp would have given `1280 / 1.4 - 16 = 898px`, so the 920px author width was clamped and the sheet fit. With this rule the clamp is `1240px` instead, nothing clamps, and the window renders `920 * 1.4 = 1288` device px inside a 1280px viewport: it is `left: 50% / translateX(-50%)` centered, so both the left and right equip columns clip off-screen. The old 600px window never came close to the clamp, so this is new with the widening. Raw `100vw` also disagrees with `--app-vw` under device emulation and pinch zoom, which is the second reason the base rule avoids it.

Suggested fix is a deletion: drop the `max-width` line entirely. The base `.window` clamp already handles it correctly, and `width` vs `max-width` do not conflict across the two rules, so removing it restores the scale-aware clamp with no other change. (`#social-window` / `#discord-window` carry the same raw-`100vw` pattern, but at 520px and 380px they realistically never reach it; that is pre-existing debt, not something to copy at 920px.)

**2. Low: the mobile `.char-stats` broadening changes spacing inside the three `cp-stats` blocks, so the "byte-identical mobile" claim is not quite exact.** `src/styles/hud.mobile.css:2987`

Widening `#char-window > .char-stats` to `#char-window .char-stats` also matches `.char-stats.cp-stats` inside the talent-summary, progression, and gathering blocks (`char_window.ts:296`, `hud.ts:10990`, `hud.ts:11017`). `.char-progression .cp-stats` sets `margin-top: 0` at (0,2,0); the new selector is (0,2,1) and wins, so those three blocks gain a 6px gap under their `cp-title` on touch that they did not have before. Purely cosmetic, but it is a behavior change on mobile. If unintended, scoping it as `#char-window > .char-info-grid > .char-stats` keeps the original meaning through the extra nesting level.

**3. Nit: stale comment.** `src/styles/components.css:1056` documents `column-span: all` as the opt-out for a full-width block, but no such rule is declared anywhere in the diff. Either add it or drop the sentence, so a later contributor does not assume the escape hatch is already wired.

Happy to re-approve once (1) is addressed; (2) and (3) are your call.

jgyy:
Thanks for the thorough second pass. Pushed 1f9a801d7.

Round 1's portrait-mobile finding was already fixed by d950a4947 before this update (verified: `.char-model-panel` restores a `172px` portrait floor via `body.mobile-touch #char-window .char-model-panel`, with the landscape-only override still scoped inside the `@media (orientation: landscape)` block).

Round 2, fixed now:
1. Deleted the `#char-window { max-width: calc(100vw - 40px) }` rule entirely. The base `.window` clamp in layout.css already accounts for `--window-scale`, and this ID-specificity override was bypassing it, letting the window overflow the viewport at high UI Scale.
2. Tightened the widened mobile selectors from `#char-window .char-stats` to `#char-window > .char-info-grid > .char-stats` (both the margin-top rule and the grid-template-columns/stat-cell rules), so they no longer also match the `.char-stats.cp-stats` blocks inside talent-summary/progression/gathering.
3. Removed the stale `column-span: all` comment sentence since no such rule exists.

`npx tsc --noEmit` and the full listed test set (char_window, char_view, mobile_window_layout/coverage/transform, css_value_validity, css_corpus, styles_extraction, architecture) all pass, 133/133.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0` and ran the checks locally.

**What I verified**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/char_window.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts`: 125 passed.
- `tests/localization_fixes.test.ts` after `npm run i18n:gen`: green (no new player-visible strings, so i18n is genuinely N/A here).
- Sim/server invariants are not in play: the diff is CSS plus one non-interactive wrapper div in the painter. `src/ui/char_window.ts` still renders through the same deps, no `IWorld` surface change, no parity pin needed. No dashes/emojis, no secrets, no generated files touched.
- Confirmed `deps.root()` is `$('#char-window')` (`src/ui/hud.ts:3675`), so the new `> .char-info-grid > ...` child combinators do match: `.char-info-grid` really is a direct child.

**Finding (medium): the round-2 selector narrowing drops the mobile overflow guard for the nested `.cp-stats` grids**

Commit 1f9a801 tightened

```css
body.mobile-touch #char-window .char-stats            -> #char-window > .char-info-grid > .char-stats
body.mobile-touch #char-window .char-stats .stat-cell -> #char-window > .char-info-grid > .char-stats .stat-cell
```

on the grounds that the old rules "also match the unrelated `.char-stats.cp-stats` blocks". Those blocks are not unrelated: talent summary (`src/ui/hud.ts:10990`), progression (`src/ui/hud.ts:11017`) and gathering (`src/ui/char_window.ts:296`) each render `<div class="char-stats cp-stats">`, and they were deliberately picking up that rule before this PR.

What they lose on touch:
- `grid-template-columns: repeat(2, minmax(0, 1fr))`. They now fall back to the desktop `.char-stats` `1fr 1fr` (`components.css:1134`), whose tracks are min-content floored, so a long unbreakable value can blow the track wider than the phone viewport.
- `max-width: 100%` / `box-sizing: border-box` / `overflow-x: hidden` on the block, and `min-width: 0` / `overflow-wrap: anywhere` on each cell.

`#char-window` has `overflow-x: hidden` on mobile (`hud.mobile.css:2911`), so the symptom is clipped text rather than a horizontal scrollbar, but it is still a visible regression, and it is the exact failure mode those declarations exist to prevent. It is most likely to bite on the gathering rows and on locales with long profession/spec/mastery names. It also makes the PR body's "mobile layout is byte-for-byte unchanged" claim no longer accurate.

Suggested fix: keep the child combinator only where the narrowing was actually needed (the `margin-top: 6px` rule at line 2987 is the one that should be top-level-only) and revert lines 3022 and 3028 back to the descendant form so `.cp-stats` keeps its guard. If you want the top-level grid to differ from the nested ones, split it into a descendant base rule plus a `> .char-info-grid > .char-stats` override rather than moving the base rule.

Worth a pin in `tests/mobile_window_layout.test.ts` asserting the `minmax(0, 1fr)` rule still reaches `.cp-stats`, otherwise the next narrowing repeats this.

**Everything else looks good**

- `columns: 320px 2` plus `break-inside: avoid` on the direct children is the right shape, and it degrades to one column on its own when the clamped width drops below two tracks, so no separate breakpoint is needed.
- Deleting the `#char-window` `max-width: calc(100vw - 40px)` override in favour of the base `.window` clamp (`layout.css:42`, which divides by `--window-scale`) is the correct call.
- The portrait `min-height: 172px` restore plus the landscape `132px` re-declaration relies on source order at equal specificity, which holds here, and the new test pins both arms rather than just one. Nice.

jgyy:
Thanks for catching this, Enrique. You are right that the two blocks were not unrelated.

Fixed in 35fe368b5: reverted the round-2 narrowing on the grid-template-columns rule and the .stat-cell rule (previously src/styles/hud.mobile.css lines 3022 and 3028) back to the descendant form (`#char-window .char-stats` / `#char-window .char-stats .stat-cell`), so the mobile overflow guard reaches the .cp-stats blocks in talent summary, progression, and gathering again. Left the margin-top: 6px rule scoped to the top-level `#char-window > .char-info-grid > .char-stats` child combinator, since that narrowing was correct.

Added a pin in tests/mobile_window_layout.test.ts asserting the descendant selector form still carries the grid-template-columns/max-width/box-sizing/overflow-x guard and the stat-cell min-width/overflow-wrap guard, plus a separate assertion that margin-top stays scoped to the top-level child only, so a future narrowing pass does not repeat this silently.

Verified: tsc --noEmit clean, and the full check set (char_window, mobile_window_layout, mobile_window_coverage, mobile_window_transform, css_value_validity, css_corpus, styles_extraction, architecture) is green at 126 passed. Pushed to the branch.

jgyy:
Status check: all three review rounds are addressed as of 35fe368b5 (portrait mobile model-panel floor, the char-window max-width clamp bypass, and the mobile cp-stats overflow-guard narrowing). Re-verified locally on this commit: `npx tsc --noEmit` clean, and `char_window`, `char_view`, `mobile_window_layout`, `mobile_window_coverage`, `mobile_window_transform`, `css_value_validity`, `css_corpus`, `architecture` all green (134 passed).

The only outstanding CI item was a stale "Screenshots of changes" run that had been canceled by a superseding queued run rather than a real failure; reran it and it now passes. `mergeStateStatus` is CLEAN and all required checks are green.

@EnriqueGF, nothing further from my side, this should be ready for a fresh look.

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`, read the char-window painter, the `components.css` / `hud.mobile.css` cascade, and the Inspect-window consumers of the shared classes.

## What I ran

- `npx tsc --noEmit`: clean
- `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`: pass
- `tests/localization_fixes.test.ts`: pass (it first failed on a missing generated `src/ui/i18n.status.json`; `npm run i18n:gen` fixes it, not a PR defect)
- `char_window`, `char_view`, `mobile_window_layout`, `mobile_window_coverage`, `mobile_window_transform`, `css_value_validity`, `css_corpus`, `styles_extraction`, `pr_shot_targets`, `hud_perf_budget`, `painter_host`, `focus_visible_guard`: all pass
- Biome on the changed files: clean. No em/en dashes or emojis in added lines. No new player-visible strings, so i18n is genuinely N/A here.

No `src/sim/` or `server/` files are touched, so determinism, the `IWorld` seam, and server authority are not in play.

## What holds up well

The wrapper div is handled carefully. Every direct-child selector the new depth could have broken was found and compensated: `#char-window > *` (hud.mobile.css:2969) gets a matching `.char-info-grid > *` rule, and `> .char-stats` was correctly rewritten to `> .char-info-grid > .char-stats`. There are no `:nth-child` / adjacent-sibling rules on these classes and no positional JS traversal, so nothing else was at risk.

Multicol is safe for this content: no `position: absolute` / `sticky` / `transform` anywhere inside the grid, no canvas (the preview lives in `.paperdoll`, outside the wrapper), and `attachTooltip` positions from `getBoundingClientRect()` rather than `offsetParent`, which is the usual multicol breakage. All four children carry `break-inside: avoid`.

Dropping the `max-width: calc(100vw - 40px)` in 1f9a801d7 was the right call, and I want to flag it as correct since the PR body still describes the cap as present. That rule sat in `@layer components` and outranked the scale-aware `max-width` in `layout.css:42`, so it was worse than no cap. The base clamp (`calc(var(--app-vw) / var(--window-scale) - 16px)`) is strictly tighter and UI-scale aware, and `window_drag_core.ts` clamps drag position, so 920px cannot overflow or be dragged off-screen.

## One thing I would fix before merge

**`.char-model-panel` still roughly doubles in height on portrait mobile, which is the exact regression d950a4947 set out to prevent.**

`min-height: 172px` (hud.mobile.css:3003) is only a floor. `.paperdoll` now sets `align-items: stretch` (components.css:1049) and nothing overrides it on portrait, so the panel's used height becomes the tallest flex item, which is the equip column. On mobile the right column has 7 slots, and each `.equip-slot` contains a 40x40 `.equip-unequip-btn` that is always visible on touch (`components.css:9276`, `9298`), so that column is at least `7*40 + 6*5 = 310px`. The panel stretches to that instead of 172px, pushing the stats and progression blocks down by about the screen-length the code comment says it is avoiding.

The landscape arm is fine, because hud.mobile.css:3009 sets an explicit `height: 132px`, which does defeat stretch. That is also why this did not show up in review: the committed `after-mobile.png` is a landscape capture, so portrait is unverified by the PR's own evidence.

Worth noting the accompanying test is a false pin. `tests/mobile_window_layout.test.ts` asserts `min-height: 172px` is present, but that declaration no longer determines the rendered height, so the test would stay green straight through this regression.

Suggested fix, which also resolves the note below: add `align-items: flex-start` to `body.mobile-touch #char-window .paperdoll` (hud.mobile.css:2990), and re-pin the test on something that actually controls the height. A portrait capture would be good evidence.

## Smaller notes, non-blocking

- **`align-items: stretch` leaks to the Inspect window.** `.paperdoll` is shared: `src/ui/hud.ts:12010` renders `<div class="paperdoll inspect-paperdoll">`, and neither `components.css` nor `shell.css:5844` overrides `align-items` for it. Real visual impact is close to zero today, since `.equip-col` has no background or border and its children still pack at the top. But this is a char-window-motivated change made on a shared base class, and it is how a future `.equip-col` background change becomes an Inspect bug. Scoping it as `#char-window .paperdoll` would keep the blast radius at the PR's stated scope. `.char-model-panel` is char-window-only, so `flex: 3 1 0` and `min-height: 340px` do not leak.
- **Stale PR body.** It says the window is capped at `calc(100vw - 40px)` (removed in 1f9a801d7) and that mobile is "byte-for-byte unchanged" (the mobile CSS diff and the `align-items` change contradict that).
- **Stale screenshots.** They were last recaptured at 9407330d4, but d950a4947 and 35fe368b5 changed mobile sizing afterward, so `after-mobile.png` does not show what now ships. The body also links them from a fork branch ref rather than the committed `docs/screenshots/` paths, so the images break if that branch moves.
- **Cosmetic.** `.char-stats` and `.char-progression` each carry `margin-top: 10px; border-top` as a separator between stacked blocks. In two columns the first block of column 2 renders that rule with nothing above it, visible at the top of the Progression column in `after-desktop.png`. Multicol cannot express a clean per-column `:first-child` reset, so I would leave it.

Nice change overall, and the direct-child selector work and the descendant-form `.cp-stats` guard restore in 35fe368b5 were both exactly right. Just the portrait stretch and its pin to sort out.

jgyy:
Addressed the latest review round (2026-07-20).

Fixed:
- The char-model-panel still doubled in height on portrait mobile. `#char-window .paperdoll` sets `align-items: stretch` so the desktop landscape 3D preview panel can take the full row height; nothing on portrait mobile defeated it, so the panel stretched to match the taller equip-slot column (each slot's always-visible touch unequip button floors that column past 172px) instead of respecting the 172px floor. Added `align-items: flex-start` on the portrait mobile paperdoll rule to defeat the stretch, so the 172px min-height now actually controls the rendered height again.
- Scoped the base `align-items: stretch` to `#char-window .paperdoll` (was unscoped `.paperdoll`), so it no longer leaks to the shared Inspect-window paperdoll as flagged.
- Re-pinned the test: the old pin only asserted the (inert, under stretch) `min-height: 172px` declaration, which would have stayed green straight through this exact regression. Added a new assertion in `tests/mobile_window_layout.test.ts` on the `align-items: flex-start` declaration that actually controls the rendered height.

Already resolved from earlier rounds (verified against current HEAD, not re-touched): the raw `calc(100vw - 40px)` max-width override (removed, base `.window` clamp handles it), the `.cp-stats` overflow guard narrowing (reverted to descendant form), and the stale `column-span: all` comment (no longer present).

Noted but not further changed: while capturing portrait mobile screenshots for evidence, confirmed the game gates all portrait mobile play behind a "Rotate to Landscape" overlay, so this defect (and the fix) is not actually player-reachable in practice. Fixed it anyway since it's what the review asked for and it's cheap, correct defense-in-depth, and mentioned this in the PR body so the "byte-identical mobile" claim is accurate (landscape, the only reachable orientation, is unchanged; portrait is fixed but moot).

Also merged `origin/release/v0.28.0` into the branch (165 commits, clean merge, no conflicts) and recaptured screenshots: `after-desktop.png` (stale since the max-width fix), plus new `before/after-mobile-portrait.png` showing the paperdoll stretch fix directly.

Verified:
- `npx tsc --noEmit`: clean
- `npx vitest run tests/char_window.test.ts tests/char_view.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts tests/world_api_parity.test.ts`: 414 passed
- `npm run i18n:gen` + `npx vitest run tests/localization_fixes.test.ts`: green, no new player-visible strings
- Biome on changed files: clean
- Full pre-push gate: green
- CI on the pushed commits: all required checks passing (Biome, PR checks/typecheck/builds, English-only legal gate, browser regressions, screenshots)

Rubsey:
I'm personally not a fan of this, I think the default size is fine. What is the reasoning for making this full screen across all displays?

jgyy:
am wanted to make character model in character window bigger, but did not work out well let me close it first think of redesign first.

PR #2129, opened 2026-07-18
feat(ui): relayout the Quest Log window into a wide landscape shape

jgyy:
## Summary

Part of the game-window landscape relayout program
(`docs/superpowers/specs/2026-07-19-window-landscape-relayout-design.md`; prior PRs: #2127
shared mobile infra, #2128 character window). The Quest Log was a squarish 520px window
whose detail pane was capped at a narrow ~46-character reading measure and a fixed 380px
height, forcing a second, nested scrollbar inside the window's own scroll region whenever
a quest's text ran long. This widens the window to a landscape shape (880px) and lets the
detail pane use the extra width, dropping its own scroll cap so it flows with the window's
single outer scroll instead.

This PR is CSS-only (no `.ts` changes): the quest log's list-plus-detail structure was
already exactly the two-pane layout this program wants, it just needed room.

```mermaid
flowchart LR
    subgraph before["Before: 520px, narrow capped detail"]
        direction LR
        A1["Quest list (180px)"] --> A2["Detail: 46ch cap, 380px max-height\n(own nested scrollbar)"]
    end
    subgraph after["After: 880px landscape, detail fills the width"]
        direction LR
        B1["Quest list (220px)"] --> B2["Detail: 70ch cap, no max-height\n(flows with the window's own scroll)"]
    end
```

### How it works

- `src/styles/components.css`: there are two `#quest-log-window` rules in this file, one
  from the current live markup path and a later one from an in-progress "window-frame"
  design-language phase (not yet wired to any painter, but a bare `#quest-log-window`
  selector that still cascades and would otherwise silently win over the first edit).
  Both are widened `520px -> 880px` (capped at `calc(100vw - 40px)`) so neither one can
  regress the other; `.ql-list` widened `180px -> 220px`; `.ql-detail`'s `max-height: 380px`
  cap dropped to `none`; the window-frame-phase detail pane's reading-measure cap widened
  `46ch -> 70ch`.
- **Mobile:** the width change is unaffected (the quest log already pins to a
  full-screen sheet on touch (`hud.mobile.css`) independent of the desktop window's
  width). Dropping `.ql-detail`'s `max-height: 380px` cap does change mobile behavior
  though: it removes the pane's own internal scroll region there too, so for a long quest
  the whole sheet now scrolls together with the list, rather than only the detail pane
  scrolling internally. That is an intentional switch to the same single-outer-scroll
  model the desktop window already uses, documented at the mobile CSS site
  (`hud.mobile.css`) and pinned in `tests/quest_log_landscape_layout.test.ts`. Both
  mobile screenshots below use the empty "no active quests" state, which cannot show
  this change; the scroll-model switch is verified by the CSS pin instead.

### Screenshots

| | Before | After |
|---|---|---|
| Mobile | before-mobile | after-mobile |

(Both shots use the empty "no active quests" state: quest-giver NPCs enforce proximity for
`acceptQuest`, so seeding an active quest headlessly would need the offline client to walk
to an NPC first; the widened window shape is the change under test either way.)

## Related issues

N/A

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx tsc --noEmit` (clean; no `.ts` files touched)
  - `npx vitest run tests/quest_log_landscape_layout.test.ts tests/questlog_window.test.ts tests/questlog_view.test.ts` (green; added a new dedicated pin file covering both `#quest-log-window` rules, the dropped max-height cap, and the widened list column)
  - `npx vitest run tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts` (green)
  - `npx @biomejs/biome check --write` on every changed file
  - Full pre-push `npm run gate` QA floor: green
- Manual: opened the quest log via the offline client (`npm run dev`) and confirmed the
  wider desktop layout and the unchanged mobile full-screen sheet, with no clipping.

## Screenshots / recordings

See `docs/screenshots/quest-log-landscape-layout/` (before/after, desktop/mobile), referenced above.

---

### Quality

- [x] The full gate passes on every file this PR touches.

### Cross-platform

- [x] Tested on desktop and mobile. Mobile width/positioning is unchanged; dropping the detail pane's max-height cap intentionally switches mobile to a single outer scroll for long quests (documented and pinned, see Summary).
- [x] Accessible. No ARIA/semantics/focus-order changes; CSS layout only.

### Localization (i18n)

### Hygiene

      enabled anywhere.


EnriqueGF (changes requested):
Reviewed the full diff, the surrounding cascade in `src/styles/`, and the quest log's runtime path. Ran locally: `npx tsc --noEmit` (clean), plus `tests/quest_log_landscape_layout.test.ts`, `questlog_window`, `questlog_view`, `css_value_validity`, `css_corpus`, `styles_extraction`, `mobile_window_layout`, `mobile_window_coverage`, `mobile_window_transform`, `pr_shot_targets`, `architecture`, and the S3 guard `localization_fixes` (all green; the S3 guard needed `npm run i18n:gen` first, unrelated to this PR).

Things I checked and can confirm are correct:

- **Editing both `#quest-log-window` rules was necessary.** `window-frame` and `window-body` appear nowhere in `src/**/*.ts`, so the later rule at line 5812 is a bare ID selector that applies unconditionally and wins on source order at equal specificity. The commit message's reasoning is right.
- **Mobile really is unaffected.** `hud.mobile.css` is `@layer hud-mobile`, ordered after `components`, and `body.mobile-touch #quest-log-window` sets both `width` and `max-width` there, so the desktop values cannot leak through.
- **Dropping `max-height: 380px` does not strand content.** `.ql-detail` keeps `overflow: hidden`, but with an unbounded height it no longer clips; the outer `.window` (`max-height: 85vh; overflow-y: auto`) provides the single scroll as intended.

Two findings, both in the CSS.

### 1. `max-width: calc(100vw - 40px)` disables the base window clamp (should fix)

`layout.css` is `@layer layout`, which sits **below** `@layer components`, so any `max-width` declared in `components.css` overrides the shared `.window` clamp outright, regardless of specificity. That clamp is:

```css
max-width: calc(var(--app-vw, 100vw) / var(--window-scale) - 16px);
```

and its comment states the contract explicitly: clamp against the JS-synced app viewport, never raw viewport units, and never bake a different division into a per-window clamp. The new `calc(100vw - 40px)` drops both the `--app-vw` source and the `/ var(--window-scale)` division.

Concretely, at uiScale 1.5 on a 1280px viewport the `#ui` box is 853px wide, so the correct clamp is 837px. The new clamp resolves to 1240px and never bites, so the window renders at 880 CSS px (1320 device px) and overflows the viewport by roughly 40px, clipped about 20px on each side by the `translateX(-50%)` centering. Before this PR the 520px width made this moot; at 880px the boundary is easy to reach. `100vw` also disagrees with `--app-vw` under device emulation and mobile browser chrome.

The house pattern is right there in this same file (`components.css:5119`, `5376`, `5387`): `calc(var(--app-vh, 100vh) * X / var(--window-scale) - Y)`. The two raw-`100vw` precedents are not really counterexamples: `#discord-window` is `position: fixed` outside the zoomed `#ui`, where `100vw` is correct, and `#social-window` is 520px so it rarely bites.

Simplest fix: **delete both added `max-width` lines.** Nothing else in `components.css` sets `max-width` on `#quest-log-window`, so the base `.window` clamp takes over and already does exactly the right thing, ui-scale-aware.

### 2. The 70ch reading measure lands only on dead code (medium)

The PR's stated goal is giving the detail pane "real reading width," but the `max-width: 70ch` is on `#quest-log-window .window-body .ql-detail`, which requires a `.window-body` ancestor that no painter ever mounts. The live `.ql-detail` (line 1273) has no `max-width` at all, and neither do `.ql-detail-text`, `.qd-text`, or `.qd-obj`.

So on the only path players actually see, the detail column goes from about 328px (roughly 55ch at 12px, a fine measure) to about 630px (roughly 105ch), which is well past a comfortable line length and is the opposite of what the 70ch cap was meant to achieve. The cap is being applied to the frame phase that is not wired up, while the live path is left uncapped.

Worth adding `max-width: 70ch` to the live `.ql-detail` block too, so the rendered window gets the measure the PR is aiming for. A pin for it in the new test file would keep the two paths from drifting again.

Neither is a big change, and the rest of the PR (scope, tests, screenshots, commit message, i18n N/A) looks good to me. Happy to re-review once these are in.

EnriqueGF:
Follow-up from a second pass over the seam contracts. One correction to my own numbers and three additions, all verified against the files.

**Correction, and finding 1 is worse than I described.** I used uiScale 1.5 as the example; the actual slider range is `uiScale: { min: 0.85, max: 1.4 }` (`src/game/settings.ts:110`), so 1.4 is the real ceiling. More importantly, I said the overflow gets "clipped about 20px on each side" without checking what happens to it: `#ui` is `overflow: hidden` (`src/styles/base.css`), so the overflow is genuinely **clipped, not scrollable**. There is no way to reach it.

At a 1024px viewport with uiScale 1.4, the `#ui` box is 731 author px and the window is 880, so it renders 1232 device px against 1024: about 208px clipped, 104px on each side given `left: 50%` plus `translateX(-50%)`. That takes out the titlebar close button and the right edge of the detail pane, with no scroll to recover them. With the base clamp restored the window would size to 715 author px and fit. Same failure at 1100px/1.4 (132px clipped) and at 800px from uiScale 1.1 upward. That moves this from a layout nit to unreachable UI, so it is worth fixing before merge.

**Addition: the second `#quest-log-window` rule is now pure duplication.** After this PR the two blocks are byte-identical (`diff` of lines 1241-1244 and 5812-5815 is empty). Unlike the neutralize rule just above it at 5792, it is not scoped to `:has(> .window-frame)`, so it is not phase-gated and does not win over anything *different*, it just restates the same values, and it doubles the surface for the clamp bug. Either delete it, or scope it `#quest-log-window:has(> .window-frame)` if a frame-phase-specific footprint is genuinely intended. Deleting it means updating the `occurrences.length).toBe(2)` assertion in the new pin file.

**Addition: the pin never covers the riskiest declaration.** Every other changed value in the diff is asserted, but `max-width` is not, which gives false confidence that the block is covered. Once the clamp is corrected, worth adding something like `expect(block).not.toMatch(/max-width:\s*calc\(100vw/)` so a future edit cannot silently re-bypass the `--app-vw` / `--window-scale` contract.

**Addition: the cited spec path does not exist.** Both the commit body and the test header cite `docs/superpowers/specs/2026-07-19-window-landscape-relayout-design.md`, but `docs/superpowers/` is not in the repo and does not appear anywhere in history. Per the docs anchor rule, a committed test header should cite a stable in-repo path: either land the spec or drop the reference.

**Minor:** `css.indexOf('.ql-detail {')` also substring-matches inside `#quest-log-window .window-body .ql-detail {`. It resolves to the live rule today only because that block happens to come first in file order, so it would silently retarget if the blocks are ever reordered. Anchoring on `'\n  .ql-detail {'` would make it robust. The `indexOf('}', start)` block slicing is otherwise sound: none of the five inspected blocks contains a nested brace, and the `toBe(2)` occurrence check is a good guard against a third rule appearing.

For the record, the mobile-unaffected claim holds but is not pinned by any test. `body.mobile-touch #quest-log-window` sets `width: clamp(320px, 76vw, 680px)` and `.ql-list { width: min(42%, 220px) }` in `@layer hud-mobile`, which is last in the barrel order and so wins over `components` regardless of specificity. A one-line assertion would make that claim auditable rather than something each reviewer re-derives.

EnriqueGF (changes requested):
Reviewed the full diff, the surrounding quest-log CSS in `src/styles/components.css`, the shared window shell in `src/styles/layout.css`, and the mobile sheet rules in `src/styles/hud.mobile.css`.

**What I ran locally** (on the PR head, `042e065e7`):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/quest_log_landscape_layout.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts`: all green.
- `tests/localization_fixes.test.ts` fails only on a missing `src/ui/i18n.status.json` (the `pretest` i18n:gen step, which I did not run in this worktree), not on anything in this PR. This PR adds no strings, so the S3 guard is not in play.

The core of the change reads right: the list-plus-detail structure was already the two-pane shape, `.window` already owns an outer scroll (`max-height: calc(var(--app-vh) * 0.85 / var(--window-scale) - 24px); overflow-y: auto`), so dropping `.ql-detail { max-height: 380px }` correctly collapses the nested scroll region into the window's own. Catching the second, later-cascading `#quest-log-window` rule in the window-frame section and pinning both in a test is exactly the right call; that one would have silently won.

## Blocking: the added `max-width: calc(100vw - 40px)` overrides the shell clamp and clips the window under UI Scale

Both new `max-width` declarations sit in `@layer components`, which outranks `@layer layout`, so they **replace** the shell clamp in `layout.css` rather than tightening it. That clamp is deliberately written as `calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)`, and its header comment spells out why: clamp against the JS-synced app box, never raw viewport units, and divide by `--window-scale` so the clamp stays exact under the `#ui` zoom. The new rule drops both halves of that.

`--ui-scale` is a shipped setting (`src/game/settings.ts`: `uiScale: { min: 0.85, max: 1.4, def: 1 }`) applied as `zoom` on `#ui`, and `#quest-log-window` lives inside `#ui`. Viewport units are not scaled by that zoom, so `100vw` under-clamps by exactly the zoom factor.

Reproduced in headless Chromium with the real rule stack (`#ui` zoom + the shell clamp), viewport 1100x800, `--ui-scale: 1.4`:

| | rendered width | left | right |
|---|---|---|---|
| shell clamp only (before) | 1078px | 11 | 1089 |
| with `max-width: calc(100vw - 40px)` (after) | **1232px** | **-66** | **1166** |

66px is cut off each side, and since `#ui` is `overflow: hidden` the overhang is clipped rather than scrollable, so the top-right close button goes with it. Before this PR the shell clamp held the window on screen at 1078px. It only bites once the window is this wide, which is why the pre-existing `#social-window` / `#discord-window` copies of the same pattern (520px / 380px) never showed it; 880px makes the clamp load-bearing.

Simplest fix: delete both added `max-width` lines. The shell already clamps every window to `app-vw / --window-scale - 16px`, so the window needs no clamp of its own. If you want the wider 40px gutter specifically, then write it the way the shell does:

```css
max-width: calc(var(--app-vw, 100vw) / var(--window-scale) - 40px);
```

Either way, worth a pin in `tests/quest_log_landscape_layout.test.ts` asserting the rule does not clamp on raw `100vw`, since the next wide window in this program will hit the same trap.

## Non-blocking

- `expect(occurrences.length).toBe(2)` in the new test is a literal count pin. It is defensible here (the whole point is "there are exactly two rules and both must move together"), but it will fail as a false positive the day someone legitimately adds a third `#quest-log-window` block. A comment in the test saying "if you add a third rule, widen it too and bump this" would save the next person the archaeology.
- Since `#ui` is `overflow: hidden`, the 880px footprint is worth a second look at the low end of the desktop range generally, not just under UI Scale. With the shell clamp restored it degrades correctly, which is another argument for just deleting the override.

Everything else checks out: no `.ts` touched, mobile is genuinely unaffected (the `@layer hud-mobile` sheet rules cascade after `components` and re-pin width/max-width/max-height themselves), no new strings so i18n is N/A, no ASCII-rule violations, and the screenshots are committed and referenced. Happy to re-approve once the clamp is sorted.

jgyy:
Thanks for the thorough review. Both findings are fixed in 3cc310918:

1. **Width clamp**: deleted the `max-width: calc(100vw - 40px)` line from both `#quest-log-window` rules (the live one and the window-frame-phase one). Neither needed its own `max-width`: the shared `.window` clamp in `layout.css` (`@layer layout`, below `@layer components`) already bounds width against `--app-vw / var(--window-scale)`, exactly like the house pattern you pointed at (`#bank-window`, `#deeds-window`). Confirmed the base clamp now wins for both rules.

2. **70ch cap on dead code**: added `max-width: 70ch` to the live `.ql-detail` rule (line ~1273) as well, so the pane players actually see gets the reading measure the PR intended, not just the not-yet-wired `.window-body .ql-detail` variant.

Also extended `tests/quest_log_landscape_layout.test.ts` to pin both: the live and frame-phase `#quest-log-window` blocks now assert no `max-width` declaration, and the live `.ql-detail` block asserts `max-width: 70ch;` alongside the existing `max-height: none;` check, so the two paths can't drift apart again.

Verified: `npx tsc --noEmit` clean, and the full targeted suite (`quest_log_landscape_layout`, `questlog_window`, `questlog_view`, `mobile_window_coverage`, `mobile_window_layout`, `mobile_window_transform`, `css_value_validity`, `css_corpus`, `styles_extraction`, `architecture`, `pr_shot_targets`) green, 114/114. Pre-push QA floor (tsc, guard tests, changed-file biome, copy scan) passed on push.

jgyy:
Thanks for the two thorough review rounds, both findings are addressed.

1. Removed max-width: calc(100vw - 40px) from both #quest-log-window rules (the live rule and the not-yet-wired window-frame-phase rule). The shared .window clamp in layout.css (calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)) now takes over unmodified, so the window stays scale-aware and does not overflow or clip under higher UI Scale.

2. Added max-width: 70ch to the live .ql-detail rule, not just the dead window-frame-phase copy, so the reading-measure cap actually lands on the path players see.

Also added a comment on the intentional exact-count pin (occurrences.length toBe(2)) in tests/quest_log_landscape_layout.test.ts, noting that a legitimate third #quest-log-window block should widen the count rather than delete the check, per the non-blocking note in the second review.

Verified: npx tsc --noEmit clean, tests/quest_log_landscape_layout.test.ts, tests/questlog_window.test.ts, tests/questlog_view.test.ts, tests/mobile_window_coverage.test.ts, tests/mobile_window_layout.test.ts, tests/mobile_window_transform.test.ts, tests/css_value_validity.test.ts, tests/css_corpus.test.ts, tests/styles_extraction.test.ts, and tests/architecture.test.ts all green, and the pre-push QA floor passed on push.

EnriqueGF (commented):
Reviewed the full diff, `src/styles/CLAUDE.md`, `src/styles/layout.css`, the mobile quest-log rules in `hud.mobile.css`, and both HTML entries.

**Ran locally:** `npx tsc --noEmit` (clean), and `npx vitest run tests/quest_log_landscape_layout.test.ts tests/questlog_window.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts` (all green, 78 assertions). `tests/localization_fixes.test.ts` failed only because `src/ui/i18n.status.json` was absent in my worktree (`npm run i18n:gen` / `pretest` generates it); not related to this PR, which adds no strings. Scanned the diff for em/en dashes and emojis: clean.

**What I verified as correct:**
- Dropping the per-window `max-width` and relying on the shared clamp is right. `#quest-log-window` really does carry `class="window panel"` (`index.html:404`, `play.html:357`), and `.window` sets `max-width: calc(var(--app-vw, 100vw) / var(--window-scale) - 16px)` (`src/styles/layout.css:42`). `max-width` beats `width` regardless of specificity, so 880px cannot overflow a narrow viewport.
- `max-height: none` does what the description claims. `.ql-detail` is `overflow: hidden` with `.ql-detail-body` at `overflow-y: auto; min-height: 0`; once the detail's height is content-driven, the body stops overflowing and the nested scrollbar genuinely disappears in favour of `.window`'s own `overflow-y: auto` + 85vh clamp.
- Editing both `#quest-log-window` rules was necessary. The second one sits in the "AAA window-frame adoption" section whose driver (`src/ui/window_frame.ts`) no longer exists (reverted, per the "Dead ends" note in `src/styles/CLAUDE.md`), but that `width` rule is *not* gated behind `:has(> .window-frame)`, so it does cascade live and would have won. Good catch.

**Findings:**

1. (medium) The live `.ql-detail` gains a **new** `max-width: 70ch` that did not exist before this PR, and the summary does not mention it (the bullet only describes `46ch -> 70ch` on the frame-phase rule). This caps the pane well below the width the PR just bought: content box is 856px, minus the 220px list plus its 8px padding and the 12px gap leaves roughly 616px for the detail, while 70ch at the pane's 12px font is roughly 420 to 460px. So the window grew 360px but the detail's actual reading width grows only about 100px, and roughly 160 to 200px of the new width sits empty. Either narrow the window nearer 760px or raise the cap. Worth making a deliberate choice rather than an incidental one. (The committed screenshots use the empty state, so they cannot show this.)

2. (low) That same new `max-width: 70ch` leaks onto touch. `hud.mobile.css:3611` only sets `min-width: 0` on `body.mobile-touch #quest-log-window .ql-detail` and never overrides `max-width`, so the components-layer cap applies inside the full-screen sheet too. At phone widths it does not bind (which is why the mobile screenshot legitimately looks unchanged), but on a tablet-width touch sheet it does, leaving dead space. If "Mobile layout is unchanged" is meant literally, add `max-width: none` to that existing mobile rule.

3. (nit) Consider a one-line pointer in the frame-phase comment noting that the section's painter was reverted and only the ungated `width` line is live. Otherwise a future reader deleting that dead section would silently revert half of this change, and the count pin in the new test would not catch a deletion (it asserts exactly 2, so removing one fails loudly, which is fine, but the *reason* is easier to act on if it is written down at the CSS site too).

Nothing here is broken, so no blocking objection. Item 1 is the one I would like an explicit answer on before merge.

jgyy:
Thanks for the round 3 pass. Confirmed rounds 1-2 fixes (raw `100vw` max-width deletion) are already in and stayed clean.

Fixed the remaining items:
- Raised the `.ql-detail` reading-measure cap from `70ch` to `100ch` (both the live rule and the dormant window-frame-phase rule, kept in sync) so the pane actually uses the ~616px available beside the 220px list inside the 880px window, instead of leaving 150px-plus of dead space.
- Added `max-width: none` to the mobile-touch `.ql-detail` rule in hud.mobile.css so the desktop cap no longer leaks onto tablet-width touch sheets (phone widths were never affected, but this makes 'mobile layout unchanged' literally true).
- Added a pointer comment on the window-frame-phase block noting its `window_frame.ts` painter was reverted and only the ungated `width` rule is live, so it doesn't read as safely-deletable dead code.
- Extended `tests/quest_log_landscape_layout.test.ts` with pins for the new 100ch cap and the mobile override.

`tsc --noEmit` and the full relevant test set (quest_log_landscape_layout, mobile_window_coverage, mobile_window_layout, mobile_window_transform, css_value_validity, css_corpus, styles_extraction, architecture) are all green. Pushed to the branch.

EnriqueGF (commented):
Reviewed the full diff against the repo invariants and re-ran the checks locally. CSS-only change, no `.ts` touched, so the sim/determinism, `IWorld` parity, server-authority and i18n invariants are all trivially out of scope. No em/en dashes, no emojis, no secrets, no generated files hand-edited. The change itself is the right shape: the quest log already had the list-plus-detail structure the landscape program wants, and widening it plus dropping the nested `max-height` scroll region is a clean way to get there.

What I ran here:
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/quest_log_landscape_layout.test.ts tests/questlog_window.test.ts tests/questlog_view.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts`: 79 passed.
- `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`): all green.

I also verified the two things most likely to bite: `box-sizing: border-box` is global in `base.css`, so the fixed `880px` plus `--window-pad` cannot overshoot the shared `.window` clamp, and the mobile sheet's own `width: clamp(320px, 76vw, 680px)` in `hud.mobile.css` sits in a later layer, so the desktop `880px` genuinely cannot leak to touch. Leaving the `max-width` off both `#quest-log-window` rules and letting the `layout.css` viewport clamp own it (commit 3cc310918) is the correct call, and the tests pin it.

Findings:

1. Medium, mobile is not actually unchanged. Before this PR, `.ql-detail` had no `max-height` override anywhere in `hud.mobile.css`, so the mobile touch sheet inherited the components-layer `max-height: 380px` and `.ql-detail-body` was a real internal scroll region on mobile too. That is confirmed by the dedicated mobile rule at `src/styles/hud.mobile.css:2780`, which gives `.ql-detail-body` `padding-bottom` and `scroll-padding-bottom` with `env(safe-area-inset-bottom)`: that only makes sense for a pane that scrolls itself. Setting `max-height: none` in the components layer therefore changes mobile behavior as well: for a long quest the detail pane now grows and the whole sheet scrolls, the quest list scrolls out of view with it, and that safe-area scroll padding becomes inert. That may well be the behavior you want (it matches the desktop intent), but the PR body and the checklist both assert mobile is unchanged, and neither screenshot can show it because both use the empty "no active quests" state. Worth either an explicit `max-height` decision in the mobile block plus a line in the PR body, or a real long-quest mobile capture. The new test file pins the desktop side thoroughly but has no pin for this.

2. Minor, the `100ch` cap on the live `.ql-detail` is effectively inert. Inside the fixed `880px` window the pane has roughly 615px of content width, and `100ch` at this pane's 12px font lands at roughly 600 to 720px, so the cap sits at or above the space that exists. The comment justifies it as "still capping line length on wider viewports", but the window is a fixed `880px` and never gets wider, so that rationale does not apply here. Not a bug, but the comment currently explains a constraint that cannot fire; either trim the reasoning or drop the declaration on the live rule and keep it only on the frame-phase one.

3. Nit, the screenshot URLs in the PR body point at `raw.githubusercontent.com/jgyy/world-of-claudecraft/feature/quest-log-landscape-layout/...`, a fork branch. The files are correctly committed under `docs/screenshots/`, so this is only a link-rot risk if that branch is deleted after merge.

Nothing here is a correctness or invariant violation, and the count pin in the new test is deliberately documented, which I am happy with. Finding 1 is the one I would like resolved before merge, either by verifying the mobile long-quest case or by saying plainly in the PR body that mobile long-quest scrolling changed on purpose.

jgyy:
Thanks for the round 4 pass. Addressed both remaining items in dfab3cba9:

1. **Mobile long-quest scroll behavior**: confirmed the finding is correct: dropping `.ql-detail`'s `max-height: 380px` does change mobile behavior for a long quest. `.ql-detail` has no height context on the mobile sheet (`.ql-cols` is an unbounded flex row), so `.ql-detail-body`'s own `overflow-y: auto` never has anything to clip, and the sheet's own `overflow-y: auto` becomes the only scroll, taking the quest list along with it. That is the same single-outer-scroll model the desktop window already uses (and that round 1-2 review approved), not an oversight, so I documented it explicitly at the mobile CSS site (`src/styles/hud.mobile.css`, the `.ql-detail-body` block) and added a pin in `tests/quest_log_landscape_layout.test.ts` asserting neither `.ql-detail` nor `.ql-detail-body` silently regains a bounded `max-height` on mobile. Not doing a separate long-quest mobile screenshot capture since the behavior is now spelled out in code and pinned, and matches the already-reviewed desktop model.

2. **Stale spec citation** (round 1): dropped the `docs/superpowers/specs/2026-07-19-window-landscape-relayout-design.md` reference from the test file header. Confirmed that path does not exist anywhere in the repo history.

Verified: `npx tsc --noEmit` clean, and the full targeted suite (`quest_log_landscape_layout`, `questlog_window`, `questlog_view`, `mobile_window_coverage`, `mobile_window_layout`, `mobile_window_transform`, `css_value_validity`, `css_corpus`, `styles_extraction`, `architecture`) green, 104/104. Pre-push QA floor passed on push.

EnriqueGF (approved):
Reviewed the full diff, `src/styles/CLAUDE.md`, and the surrounding cascade. This is CSS plus one new pin test, no `.ts`, so the sim/IWorld/server-authority/i18n invariants are all N/A here. No dashes or emojis in the diff.

**What I verified myself (not just from the description)**

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/quest_log_landscape_layout.test.ts tests/client_shell.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/styles_extraction.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/architecture.test.ts tests/window_resize.test.ts tests/pr_shot_targets.test.ts tests/questlog_view.test.ts tests/questlog_window.test.ts`: all green (187 + 31 passing).
- `tests/localization_fixes.test.ts`: green after `npm run i18n:gen` (it fails on a fresh worktree only because the generated `i18n.status.json` is absent, which `pretest` normally produces; unrelated to this PR).

Cascade / layout checks that back the change:

- Dropping `.ql-detail { max-height: 380px }` is safe on desktop because the shared shell already owns an outer scroll: `layout.css` `.window` sets `max-height: calc(var(--app-vh) * 0.85 / var(--window-scale) - 24px)` plus `overflow-y: auto`, and `.window > .panel-title` is sticky, so a long quest scrolls the window with the header pinned rather than growing past the viewport.
- The decision not to add a per-window `max-width` is right, and the comment explaining it is accurate: the `.window` clamp divides by `--window-scale`, so a raw `100vw` override really would disagree at higher ui-scale.
- `880px` cannot overflow horizontally: `base.css` sets a global `box-sizing: border-box`, so the 12px `--window-pad` is inside the declared width and the `max-width` clamp bounds the same box.
- `100ch` fits: 880 - 220 list - 12 gap - 24 padding leaves roughly 624px, and 100ch at the pane's 12px font lands around 600px.
- Mobile does not inherit the 880px: `hud.mobile.css` re-declares `width: clamp(320px, 76vw, 680px)` for `#quest-log-window` in the `hud-mobile` layer, which sorts after `components`.
- No persisted-size conflict with the resize feature: `installWindowResize` writes inline `style.width/height` per session only (no storage), and `WINDOW_MIN_WIDTH` is 220, below the new default.

**Findings**

1. Minor, description accuracy. The PR body says "**Mobile:** unaffected" and the checklist says "Mobile layout is unchanged from before this PR", but that is no longer true after the follow-up commits: dropping the components-layer `max-height: 380px` also removes mobile's internal detail scroll region (`.ql-detail-body`'s `overflow-y: auto` loses its height context), so on a long quest the whole sheet now scrolls together with the list instead of only the detail pane. The CSS comment at `hud.mobile.css` documents this well and I agree it is a reasonable, deliberate switch to the single-outer-scroll model, but a reviewer trusting the summary would skip mobile entirely. Please update the summary and the checklist line to describe the mobile behavior change.

2. Minor, screenshot evidence. Both mobile shots are the empty "no active quests" state, which is exactly the case that cannot show the one behavior this PR actually changes on mobile (the nested-to-single scroll switch). A mobile shot with a long quest selected would make the tradeoff in finding 1 visible.

3. Nit, test fragility. `expect(occurrences.length).toBe(2)` in `tests/quest_log_landscape_layout.test.ts` pins a literal count of `#quest-log-window {` blocks. The comment tells a future contributor exactly what to do, so I am fine with keeping it, just noting it will need touching whenever a third block appears.

Nothing blocking. Approving on the code; findings 1 and 2 are about the PR body and the screenshot set rather than the CSS.

jgyy:
Addressed the latest (approved) review round.

Round 5 (EnriqueGF, APPROVED) confirmed the CSS itself is correct as of commit dfab3cba9
(the 100ch reading-measure fix and the mobile single-outer-scroll documentation from
round 4 were already in place) and raised two non-blocking items, both about the PR
description rather than the code:

1. The PR body and checklist said mobile was "unaffected" / "unchanged," but dropping
   the components-layer max-height: 380px cap on .ql-detail does change mobile behavior:
   it also removes the pane's own internal scroll region there, so for a long quest the
   whole sheet now scrolls together with the list instead of only the detail pane
   scrolling internally. That switch is intentional (matches the desktop single-outer-
   scroll model) and was already documented at the mobile CSS site and pinned in
   tests/quest_log_landscape_layout.test.ts as of round 4. Updated the PR body's "How it
   works" section and the cross-platform checklist line to say this plainly instead of
   claiming mobile is unchanged.

2. Requested a mobile screenshot with a long quest selected to make the scroll-model
   change visible (both current shots use the empty "no active quests" state).
   Deferred: seeding an active quest headlessly needs the offline client to walk to a
   quest-giver NPC first (acceptQuest enforces proximity), which is more scaffolding than
   this non-blocking ask warrants given the behavior is already verified by the CSS pins
   in tests/quest_log_landscape_layout.test.ts. Happy to add a real capture if that
   tradeoff is wanted before merge.

No code changes were needed this round; verified clean against the current diff:
- npx tsc --noEmit: clean
- npx vitest run tests/quest_log_landscape_layout.test.ts tests/questlog_window.test.ts tests/questlog_view.test.ts tests/mobile_window_coverage.test.ts tests/mobile_window_layout.test.ts tests/mobile_window_transform.test.ts tests/css_value_validity.test.ts tests/css_corpus.test.ts tests/styles_extraction.test.ts tests/architecture.test.ts: 104 passed
- npx @biomejs/biome check on all changed files: clean
- Merge check against upstream/release/v0.28.0: no conflicts

CI: all required checks pass (PR gate, PR checks/typecheck/build, Biome, browser
regressions). The "Screenshots of changes" bot job is showing cancelled/failed here, but
the same is true on several unrelated open PRs right now, and the module it reports
missing is committed and present at HEAD, so this reads as a repo-wide runner flake, not
something in this diff.

PR #2130, opened 2026-07-18
feat(sfx): real recordings for the UI cue pack, fix cooldown enforcement bug

jamiecypher:
## Summary

Real recordings for most of the remaining procedural/ElevenLabs-placeholder UI sound cues, plus two real bugs found and fixed along the way:

- `sfx.playUi()` silently ignored its own `cooldown` option; only `playAt` enforced cooldowns. Fixed, opt-in default (no default cooldown, unlike `playAt`), so existing rapid-fire call sites don't regress.
- Both `playAt` and `playUi` had a `?? -1` sentinel bug that could wrongly block the very first play of a key when `cooldown >= 1s` and the AudioContext just started. Fixed by switching to `Number.NEGATIVE_INFINITY`.

New/replaced real recordings: `ui_click`, `ui_coin` (3 takes), `ui_bag_open`/`ui_bag_close`, `ui_duel_challenge`/`ui_duel_countdown`/`ui_duel_start`/`ui_duel_end`, `ui_cosmetic_unlock`, `ui_arena_loss`, `ui_loot_item` (5 takes), `ui_ready_check`, `ui_sheep`, `ui_weapon_sheathe`/`ui_weapon_unsheathe`, and a corrected `ui_error` take. `readyCheck()`/`weaponSheathe()`/`weaponUnsheathe()` used to be hardcoded procedural WebAudio synths directly in `src/game/audio.ts`; that entire fallback synth is now removed, `audio.ts` is fully sampled cues.

Consolidates the "a match is starting" sound family: duel challenge, party invite, guild invite, arena queue pop, and Vale Cup match-found now share `ui_duel_challenge` (retiring the old, misnamed `ui_quest_accept`/`questAccept()`). Vale Cup kickoff and Arena rating loss are split off into their own new keys so a real duel-start/loss recording doesn't also fire for the soccer minigame. Cosmetic/skin unlock gets its own cue, split from real level-up.

**Follow-up fix from QA review:** consolidating party invite / guild invite / a newly-merged `resurrectionOffer` prompt onto `duelChallenge()` silently dropped their Interface & Feedback Sounds mute-toggle gating (the retired `questAccept()` always respected it; `duelChallenge()` is deliberately ungated for real time-critical challenges). Added `invitePrompt()`, a gated sibling playing the same clip, and rewired those three call sites onto it.

This branch also merges in `upstream/release/v0.27.0` (181 commits, was a day behind), resolving conflicts in `sfx.ts` (a `playAt` boolean-return refactor landed upstream for the mob idle-bark sweep feature), `hud.ts` (an arena-result logging refactor), and the SFX catalog key-count pin tests.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx tsc --noEmit`, `node scripts/sfx_conform.mjs`, `npm test -- --maxWorkers=4` (full suite, 1323 files / 16095+ tests, 0 failures), `npm run build`, scoped `npx @biomejs/biome check --write` on every touched file.
- Manual steps: two full `/qa` gate passes (first found the invite-gating regression above, fixed, second pass came back clean). Listened to every new/replaced cue in-game via the local dev server after each batch.

### Quality
- [x] The full gate passes (run manually step-by-step; `npm run gate`/`ci:changed` false-positive-flood on this branch due to a known, separately-tracked scoping bug on a diverged branch, not something this PR introduces).

### Localization (i18n)
- [x] N/A. This PR adds no player-visible strings (audio-only + one doc update).

### Hygiene
- [x] No hand-edited generated files (`sfx_manifest.generated.ts`, `runtime-pack.json`, `sfx_gain_ceiling.generated.json` all regenerated via `npm run sfx:manifest`).


jgyy (commented):
Nice batch of real recordings, and the two bugs are genuinely worth catching: `playUi` silently dropping its own `cooldown` option would have made `audio.error()`'s spam-guard a no-op, and the `?? -1` sentinel could have swallowed the very first play of any cooldown-gated key. Both fixes are small and the `-Infinity` reasoning in the comments at `src/game/sfx.ts:947` and `:963` is correct: `now - (-1)` will always clear a sub-1s cooldown by accident, which is exactly why it never showed up before this PR introduced a 1.5s one.

Traced the renames through to make sure nothing was left half-migrated:
- `questAccept()` retirement: no leftover references outside a comment (`src/game/audio.ts:181`); all three call sites in `hud.ts` (partyInvite, resurrectionOffer, guildInvite) now go through the new gated `invitePrompt()`.
- `arenaLoss()` split: `player_death_audio.test.ts` correctly drops the arena rating-loss branch from the shared `audio.death()` count (4 -> 3) and Yumi match-loss still uses `audio.death()` at `hud.ts:9147`, matching the updated test.
- `levelUp()` -> `cosmeticUnlock()` rename is consistent across `hud.ts`, `skin_event_controller.ts`, and both test files.
- `audio.ts`'s procedural WebAudio bed (`cueCtx`/`cueMaster`/`noiseBuf`/`cueTone`/`cueNoise`/`readyCheckTone`) is fully removed with no dangling references, matching the `CLAUDE.md` update.

Findings:

- nit: `sfx.ts`'s `playUi` only updates `this.lastPlay` when `cd > 0` (the new line after the cooldown check). That's a deliberate and reasonable choice given only `error()` passes a cooldown today, but it means `playUi` and `playAt` now have subtly different bookkeeping semantics for the same shared map (`playAt` always stamps `lastPlay`, `playUi` only stamps it when a cooldown was requested). Worth a one-line comment so a future no-cooldown-by-default assumption doesn't get read as a bug when someone next touches this.

Everything else checks out: no player-visible strings so i18n is correctly marked N/A, the generated manifest/runtime-pack/gain-ceiling files match what `sfx:manifest` would produce (spot-checked several hashes/byte counts against the catalog), and the catalog key-count tests (`sfx_manifest.test.ts`, `game_audio.test.ts`) were bumped consistently with the actual diff. Merge-tree against `release/v0.27.0` is clean, no conflicts. This is a straightforward, well-tested PR; the one nit above is not blocking.

jamiecypher:
Addressed the nit in c8d2988a5: added a comment explaining why `playUi` only stamps `lastPlay` when `cd > 0` (its cooldown defaults to 0, opt-in only) while `playAt` always stamps it (defaults to 0.03). Deliberate, not a bug, now documented as such.

PR #2131, opened 2026-07-18
fix(input): stop Firefox camera rotation freezing after a forced pointer-lock unlock

jgyy:
## Summary

Firefox camera rotation has been an ongoing, hard-to-pin-down problem across several past fixes (camera snap-back #1778, mouse camera rotation #1834, mouselook release facing). This closes the remaining root cause: an undocumented Firefox-specific pointer-lock cooldown that the game had no idea existed.

**Root cause:** Firefox denies `requestPointerLock()` for a short cooldown (empirically ~1.3s, Mozilla bugzilla 1284785) after it force-unlocks the pointer through its own default gesture (the Escape key, or the tab losing focus) — as opposed to our own scripted `exitPointerLock()` call, which is unaffected. The game had:
- no `pointerlockerror` listener at all, so a denied lock request was invisible, and
- no awareness of this cooldown, so it always retried `requestPointerLock()` immediately.

Sequence that reproduces the bug:
1. Player right/left-drags to rotate the camera (pointer lock engages fine).
2. Player presses Escape mid-drag — to open the game menu (Escape is bound to it), or just because the cursor already feels stuck. This force-unlocks the pointer while the mouse button is still physically held.
3. The player releases and immediately starts a new drag. `requestPointerLock()` fires synchronously (the existing Firefox sync-gesture fix from #1840) but lands inside Firefox's cooldown and is silently denied.
4. The code assumed the lock had been requested successfully. With no actual lock held, the OS cursor is free to reach the screen edge, at which point `movementX`/`movementY` clamp to 0 and rotation freezes in that direction — reading exactly like "the camera is broken."

Because Escape is also the menu key, this is very easy to trigger by accident, and matches the "works maybe 10% of the time" pattern: Firefox players who hit a stuck-feeling cursor instinctively press Escape "to fix it," which walks straight into this cooldown and makes the very next drag attempt fail too.

```mermaid
sequenceDiagram
    participant P as Player (Firefox)
    participant I as input.ts
    participant B as Browser (Gecko)

    P->>I: mousedown (right-drag starts)
    I->>B: requestPointerLock() (synchronous, in gesture)
    B-->>I: granted
    Note over I,B: camera rotates normally

    P->>B: presses Escape (menu / "cursor feels stuck")
    B->>I: pointerlockchange (pointerLockElement = null)
    Note over I: mouse button still physically held<br/>forcedUnlockAt = now() (NEW)

    P->>I: mousemove (still dragging)
    I->>I: leftDown/rightDown cleared by releaseCapture()
    Note over I: drag ends early, rotation stalls

    P->>I: mouseup, then mousedown again (retry drag)
    I->>B: requestPointerLock()
    B-->>I: DENIED (Firefox forced-unlock cooldown, ~1.3s)
    Note over I: BEFORE FIX: pointerLockRequestedForDrag=true anyway,<br/>no lock held, cursor hits screen edge, camera freezes again

    Note over I,B: AFTER FIX: inForcedPointerLockCooldown() is true,<br/>request is skipped, drag continues UNLOCKED<br/>(rotation keeps working away from the edge)<br/>next drag after ~1.3s re-engages the lock normally
```

## Changes

- `src/game/pointer_lock.ts`: new pure, unit-tested `inForcedPointerLockCooldown()` plus the `POINTER_LOCK_FORCED_UNLOCK_COOLDOWN_MS` constant.
- `src/game/input.ts`:
  - Track `forcedUnlockAt` when `pointerlockchange` fires with the pointer unlocked while a drag button is still held (the signature of a browser-driven unlock, not our own end-of-drag `exitPointerLock()`).
  - Skip the doomed `requestPointerLock()` call (both the Firefox synchronous mousedown path and the deferred mousemove path) while inside the cooldown, so the drag continues unlocked instead of desyncing internal state.
  - Add the previously-missing `pointerlockerror` listener as a defensive backstop, resetting `pointerLockRequestedForDrag` on any other denial.
- `tests/pointer_lock.test.ts`: new `inForcedPointerLockCooldown` test suite (in/out of cooldown, no prior forced unlock, non-Gecko browsers unaffected).

## Why no screenshots

This is a pure input-timing fix with no visual/HUD/render surface — there is nothing a before/after screenshot would show. The mermaid sequence diagram above documents the reproduction and fix instead. (Also: this bug is Firefox-only and the available browser automation here drives Chrome, which never reproduces it, so it couldn't be screen-recorded live either way.)

## Test plan

- [x] `npx vitest run tests/pointer_lock.test.ts tests/pointer_look_delta.test.ts tests/mouselook_release_facing.test.ts tests/architecture.test.ts tests/input.test.ts` — all green (new cooldown tests included)
- [x] `npm run gate` — full CI-equivalent gate green
- [x] Pre-push QA floor (typecheck, guard tests, biome, copy rules) passed on push
- [ ] Manual Firefox verification: rotate camera, press Escape mid-drag, close menu, immediately rotate again — rotation should keep working instead of freezing at the screen edge

### DIFF
--- a/src/game/input.ts
+++ b/src/game/input.ts
@@ -10,6 +10,7 @@ import { detectBrowserEngine } from './browser_env';
 import { cursorForHover, type HoverCursorKind } from './cursors';
 import { comboCode, isModifierCode, type Keybinds, makeCombo } from './keybinds';
 import {
+  inForcedPointerLockCooldown,
   pointerLockNeedsSyncGesture,
   shouldEngagePointerLock,
   shouldEngagePointerLockOnMouseDown,
@@ -176,6 +177,9 @@ export class Input {
   private lookPitchSign = 1;
   private downButton = -1;
   private pointerLockRequestedForDrag = false;
+  // Set when the browser itself force-unlocks the pointer (Escape, focus
+  // loss) while a drag button was still held; see inForcedPointerLockCooldown.
+  private forcedUnlockAt: number | null = null;
   // Firefox rejects requestPointerLock() when it is deferred to a later
   // mousemove; computed once since the browser cannot change mid-session.
   private readonly needsSyncPointerLockGesture = detectPointerLockNeedsSyncGesture();
@@ -242,6 +246,13 @@ export class Input {
     window.addEventListener('pointercancel', (e) => this.onMouseUp(e));
     document.addEventListener('pointerlockchange', () => {
       if (!document.pointerLockElement) {
+        // A forced unlock (Escape, or losing focus while locked) fires with a
+        // drag button still physically down, unlike our own end-of-drag
+        // exitPointerLock() call, which only ever runs after mouseup has
+        // already cleared leftDown/rightDown. Remember it so the next
+        // requestPointerLock() attempt can skip Firefox's post-forced-unlock
+        // cooldown instead of failing silently mid-drag (#1834 recurrence).
+        if (this.leftDown || this.rightDown) this.forcedUnlockAt = performance.now();
         this.releaseCapture('pointerlock');
         return;
       }
@@ -259,6 +270,13 @@ export class Input {
       ) {
         document.exitPointerLock();
       }
+    });
+    // A denied request (e.g. the forced-unlock cooldown above, or any other
+    // rejection) otherwise leaves pointerLockRequestedForDrag wrongly set to
+    // true with no lock actually granted; without this the drag never
+    // reconsiders requesting the lock again for the rest of that press.
+    document.addEventListener('pointerlockerror', () => {
+      this.pointerLockRequestedForDrag = false;
     });
     document.addEventListener('visibilitychange', () => {
       if (document.hidden) this.releaseCapture('hidden');
@@ -691,6 +709,10 @@ export class Input {
     return this.controllerFacing;
   }
 
+  private msSinceForcedUnlock(): number | null {
+    return this.forcedUnlockAt === null ? null : performance.now() - this.forcedUnlockAt;
+  }
+
   private releaseCapture(reason: string): void {
     const hadInput = this.keys.size > 0 || this.leftDown || this.rightDown;
     // Always drop the mouse-drag state so a button can't stick "held".
@@ -976,6 +998,10 @@ export class Input {
         needsSyncGesture: this.needsSyncPointerLockGesture,
         lockOnRotate: this.lockCursorOnRotate,
         alreadyLocked: document.pointerLockElement === this.canvas,
+      }) &&
+      !inForcedPointerLockCooldown({
+        needsSyncGesture: this.needsSyncPointerLockGesture,
+        msSinceForcedUnlock: this.msSinceForcedUnlock(),
       })
     ) {
       this.pointerLockRequestedForDrag = true;
@@ -1055,6 +1081,10 @@ export class Input {
           lockOnRotate: this.lockCursorOnRotate,
           isFullscreen: this.isBrowserFullscreen(),
           alreadyLocked: document.pointerLockElement === this.canvas,
+        }) &&
+        !inForcedPointerLockCooldown({
+          needsSyncGesture: this.needsSyncPointerLockGesture,
+          msSinceForcedUnlock: this.msSinceForcedUnlock(),
         })
       ) {
         this.pointerLockRequestedForDrag = true;
--- a/src/game/input.ts
+++ b/src/game/input.ts
@@ -247,11 +247,19 @@ export class Input {
     document.addEventListener('pointerlockchange', () => {
       if (!document.pointerLockElement) {
         // A forced unlock (Escape, or losing focus while locked) fires with a
-        // drag button still physically down, unlike our own end-of-drag
-        // exitPointerLock() call, which only ever runs after mouseup has
-        // already cleared leftDown/rightDown. Remember it so the next
-        // requestPointerLock() attempt can skip Firefox's post-forced-unlock
-        // cooldown instead of failing silently mid-drag (#1834 recurrence).
+        // drag button still physically down: the ordinary end-of-drag
+        // exitPointerLock() call only ever runs after mouseup has already
+        // cleared leftDown/rightDown. This is not a hard guarantee though:
+        // setLockCursorOnRotate()/setMouseCameraEnabled() call
+        // exitPointerLock() directly and can fire mid-drag with a button
+        // still held, which reads here as a forced unlock too. The fallout is
+        // small (toggling one of those settings mid-drag suppresses the lock
+        // for the next ~1.3s on Firefox) and self-corrects once the drag
+        // ends, so it is an accepted trade-off rather than something worth
+        // threading extra state through to distinguish. Remember the forced
+        // unlock so the next requestPointerLock() attempt can skip Firefox's
+        // post-forced-unlock cooldown instead of failing silently mid-drag
+        // (#1834 recurrence).
         if (this.leftDown || this.rightDown) this.forcedUnlockAt = performance.now();
         this.releaseCapture('pointerlock');
         return;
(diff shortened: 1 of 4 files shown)


EnriqueGF (commented):
Reviewed the full diff against the root CLAUDE.md and `src/game/`, and re-read the surrounding pointer-lock code paths in `input.ts` rather than just the hunks.

**What I ran here (worktree at the PR head, base `release/v0.28.0`):**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/pointer_lock.test.ts tests/input.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/pointer_look_delta.test.ts tests/mouselook_release_facing.test.ts`: 400 passing.
- `tests/localization_fixes.test.ts`: initially failed only on a missing `src/ui/i18n.status.json`; green after `npm run i18n:gen`, so not related to this PR.

Invariants: nothing touches `src/sim/`, so determinism and the `IWorld` parity pin are not in play; `performance.now()` in `src/game/` is fine. No player-visible strings, no dev-command or secret surface, no dashes or emojis. The pure-core plus thin-consumer shape (`pointer_lock.ts` decision function, `input.ts` as caller) is exactly the recipe this repo asks for, and the new function is well covered including the Chromium negative case. The diagnosis reads right to me and the fail-open behavior (skip the doomed request, keep dragging unlocked) is the correct trade-off.

Two real gaps, neither blocking:

**1. Medium: the focus-loss arm of the fix probably never triggers.** `forcedUnlockAt` is only set when `pointerlockchange` fires with `leftDown || rightDown` still true. But losing focus also fires `blur`, and `releaseCapture('blur')` unconditionally clears `leftDown`/`rightDown` (input.ts:719-720). If `blur` lands before `pointerlockchange` (which is the likely ordering, since the unlock is a consequence of the document losing focus), the drag flags are already false and the forced unlock is not recorded, even though Firefox's cooldown is very much running. The Escape case, which is the common one, is unaffected and works. Worth either narrowing the PR description's focus-loss claim or handling it explicitly.

**2. Low, and it fixes #1 cheaply: the `pointerlockerror` handler discards the strongest available signal.** A denial is the most direct evidence that we are inside a cooldown, but the new listener only resets `pointerLockRequestedForDrag`. Setting `forcedUnlockAt = performance.now()` there too would make the mechanism self-healing: whatever caused the denial (forced unlock we missed, focus loss ordering, anything upstream), the next drag within the window skips the doomed request instead of repeating it. That would cover the case above without needing to reason about event ordering at all.

**3. Nit: the comment at input.ts:250-252 overstates the invariant.** It says our own `exitPointerLock()` "only ever runs after mouseup has already cleared leftDown/rightDown", but `setLockCursorOnRotate()` and `setMouseCameraEnabled()` both call it directly and can fire mid-drag with a button still held. The practical fallout is tiny (toggling one of those settings mid-drag on Firefox suppresses the lock for the next ~1.3s), but the comment as written would mislead the next reader into trusting a guarantee that does not hold.

jgyy:
Thanks for the thorough re-verification and the two real gaps, especially the self-healing fix for #2 (it cheaply covers #1 too).

Pushed a follow-up commit that also arms `forcedUnlockAt` from the `pointerlockerror` handler itself, per your suggestion: a denial is the strongest available evidence we're in the cooldown regardless of whether `pointerlockchange` happened to catch a drag button still held (blur landing ahead of it, or any other ordering). Added an integration test in `tests/input.test.ts` covering a bare `pointerlockerror` (no forced-unlock recorded via `pointerlockchange`) suppressing the very next Firefox mousedown request during the cooldown window.

Also tightened the `pointerlockchange` comment (nit #3): it now says plainly that `setLockCursorOnRotate()`/`setMouseCameraEnabled()` call `exitPointerLock()` directly and can fire mid-drag, rather than implying our own unlock path is guaranteed to run only after mouseup clears the drag flags.

`npx tsc --noEmit` clean, and `tests/input.test.ts`, `tests/pointer_lock.test.ts`, `tests/pointer_look_delta.test.ts`, `tests/mouselook_release_facing.test.ts`, `tests/architecture.test.ts` all green (140 tests).

jgyy:
Thanks again for the careful re-read, Enrique. Confirming for the record: both concrete gaps you flagged (the focus-loss ordering gap in #1, and the self-healing pointerlockerror fix for #2) and the comment nit (#3) are already addressed in commit 466093f0c, which also adds an integration test in tests/input.test.ts covering a bare pointerlockerror suppressing the next Firefox mousedown request during the cooldown window.

Re-verified here: npx tsc --noEmit clean, and tests/input.test.ts, tests/pointer_lock.test.ts, tests/pointer_look_delta.test.ts, tests/mouselook_release_facing.test.ts, tests/architecture.test.ts all green (140 tests). No further action needed on my end.

EnriqueGF (approved):
Reviewed the full diff against the repo invariants and ran the checks locally. This is a clean, well-scoped fix and I think it should go in. One verified minor observation below, non-blocking.

## What I reviewed

- The root-cause story in the PR body against the actual code paths in `src/game/input.ts` (both the Firefox synchronous-mousedown request and the deferred `mousemove` request), not just the description.
- `inForcedPointerLockCooldown()` as a pure module in `src/game/pointer_lock.ts`: correct shape per the module-first rule (DOM-free, unit-tested in isolation, thin consumer in `input.ts`), consistent with the existing `shouldEngagePointerLock*` family. Good family reuse rather than bespoke logic.
- Invariants in play: nothing here touches `src/sim/`, so determinism and sim purity are unaffected (`performance.now()` is fine in `src/game/`). No `IWorld` surface change, so no tri-host parity work needed. No server authority surface, no player-visible strings (so no `t()` obligation), no secrets, no `ALLOW_DEV_COMMANDS`.
- Copy rules: no em dashes, en dashes, or emoji in any added line.

## What I ran

- `npx tsc --noEmit` clean.
- `npx vitest run tests/pointer_lock.test.ts tests/input.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/pointer_look_delta.test.ts tests/mouselook_release_facing.test.ts` all green (401 tests).
- `tests/localization_fixes.test.ts` green after `npm run i18n:gen` (its first failure was just the missing generated `i18n.status.json` artifact that `pretest` normally produces, not this PR).
- `biome check` on the four changed files: warnings only, no errors and no format diffs.

The new tests are decisive: the pure-function suite pins both sides of the threshold (1299 true / 1300 false), the null case, and the non-Gecko case, and the `input.ts` test asserts the *absence* of a second `requestPointerLock` call rather than just a flag.

## Minor: the doomed deferred request re-arms the cooldown, costing one extra unlocked drag

Verified empirically with a scratch test against the real `Input` harness, on the Firefox UA:

1. A forced unlock arms `forcedUnlockAt` at T0.
2. A drag starting at T0+1200 correctly skips the synchronous mousedown request.
3. If the drag threshold is crossed *after* the cooldown expires (T0+1400), the deferred `mousemove` path now passes both guards and fires `requestPointerLock()` (call count went 1 to 2). On Firefox outside fullscreen that call is exactly the one `pointerLockNeedsSyncGesture` documents as always denied.
4. The resulting `pointerlockerror` re-arms `forcedUnlockAt` at T0+1400, so the *next* genuine drag at T0+1700 has its synchronous request suppressed too (total stayed at 2).

So one additional drag runs unlocked beyond what the real Firefox cooldown requires. It damps out rather than looping (the suppressed drag fires no request, so nothing re-arms again), and the degraded state is the pre-existing "rotation works away from the screen edge" behavior, so the user-visible cost is small and it self-corrects within ~1.3s. Worth knowing about rather than fixing now; if you did want to tighten it, the narrow version would be to not re-arm from `pointerlockerror` when the failed request came from the deferred path on a browser where that path is known-denied anyway.

The trade-off in the `pointerlockchange` handler (`setLockCursorOnRotate`/`setMouseCameraEnabled` calling `exitPointerLock()` mid-drag reading as a forced unlock) is already called out honestly in the comment, and I agree it is not worth threading extra state for.

Nice work tracking this one down. Approving.

PR #2134, opened 2026-07-18
feat(sfx): wire real recordings across gathering, crafting, and the enchanting profession

jamiecypher:
## Summary

Professions 2.0 shipped several actions with zero audio feedback: gathering (`gatherResult`), craft success (`craftResult` reused the generic loot ding), the masterwork proc, and the enchanting profession's disenchant/salvage/apply-enchant actions. This PR wires real recordings for all of them:

- `audio.gather(nodeType)` — one real cue per `GatherNodeType` (ore/wood/herb, 5 takes each), replacing the old flat placeholder synth.
- `audio.gatherRareTier(tier)` — a rare/epic/legendary stinger layered ALONGSIDE the node-type cue above on a rare-or-better material roll or a rare-event roll, never a replacement for it.
- `audio.craftSuccess(recipeFamily)` — resolves the recipe's `professionId`, one real cue per `CRAFT_RING` family that actually has recipes; falls back to the loot ding otherwise.
- `audio.masterwork()` — layers alongside the family cue on a masterwork `craftResult`.
- `audio.disenchant()` / `audio.salvage()` / `audio.enchant()` — the enchanting profession's three real actions (`disenchantResult` / `salvageResult` / `enchantResult`), all previously silent. `enchant` reuses the one enchanting-profession recording (`ui_craft_enchanting`): the profession has no craftable recipes in `recipes.ts`, so `craftByFamily.enchanting` can never actually fire, and there's no separate "craft an enchant" moment to record.
- A `silent` opt threaded through `Sim.addItem`/`addItemInstance` so a professions grant's dedicated cue doesn't stack on top of the generic loot ding (the "You receive:" text line still prints either way).

Rebased onto current `release/v0.29.0` (this branch predates Phase 12b through 14b; gathering is now a multi-tick cast rather than instant, and the enchanting profession's full `IWorldProfessions` surface, `disenchant_item`/`apply_enchant`/`salvage_item` commands, already shipped independently of this branch).

Found and fixed while reconciling against current mainline:
- Two duplicate, dead entries from this branch's original commit that auto-merged in silently (never flagged as a conflict): a shadowed `case 'disenchantResult':` in `hud.ts`, and a duplicate `'disenchant_item'` token in both `COMMAND_NAMES` and `server/game.ts`'s dispatch switch.
- A stale test that assumed gathering grants instantly; updated to run the gather cast to completion.
- 4 parity golden traces that had been resolved to this branch's stale pre-Phase-12b version during the rebase; regenerated via `UPDATE_PARITY=1`.
- An orphaned placeholder file (`ui_gather_strike.mp3`) superseded by the real per-node-type recordings.

## Type of change

## How was this tested?

- Commands: `npx tsc --noEmit` (clean), full `npm test -- --maxWorkers=4` (1487 files / 18495 tests, 0 failures), `npm run build`/`build:server`/`build:env`, `npm run sfx:check` (clean), `npm run sfx:manifest` (zero drift).
- Manual: two full `/qa` passes (via `/review-pr 2134`), both READY. See QA History below.

## QA History

**Pass 1** (post-rebase, commit `b5240a538`): qa-checklist, architecture-reviewer, and frontend-seam-reviewer all clean, 0 blocking/should-fix, a few informational notes only (masterwork's layered cues, the disenchant/enchant overlap now clarified above).

**Pass 2** (commit `18de3e059`, after the branch changed significantly mid-review and got a fresh full re-review): 0 BLOCKING, 1 SHOULD-FIX — the gather rarity-tier resolution was an inline `hud.ts` ternary with no unit test, and the salvage silent-loot suppression wasn't directly asserted. Fixed: extracted the ternary into `gatherRareTierFor()` in `src/ui/gathering_view.ts` (the existing pure-core home for gather HUD logic) with 4 new unit tests, and added the missing salvage silent-loot test alongside the existing disenchant one. Verdict after fix: READY.

### Quality
- [x] The full gate passes (run manually step-by-step; `ci:changed`/the pre-push hook's `biome ci --changed` false-positive-floods on this diverged branch, comparing against a stale `origin/main` — a known, pre-existing, separate issue, not introduced here).

### Localization (i18n)
- [x] N/A. This PR adds no player-visible strings (audio-only + backend plumbing).

### Hygiene
- [x] No hand-edited generated files (`sfx_manifest.generated.ts`, `runtime-pack.json`, `sfx_gain_ceiling.generated.json` all regenerated via `npm run sfx:manifest`).


Rubsey (commented):
Reviewed the full parity surface at the head (254fed8cd), plus the sim thread and the re-minted goldens. This is a textbook parity change: every leg landed in the same PR (IWorld type plus both implementations, command send plus server dispatch, event emit plus client handler plus the ClientWorld mirror), and the pinned counts in command_schema.test.ts (157 send / 166 dispatch) and world_api_parity.test.ts (250 = 68 + 182) are internally consistent.

Verified clean:

- disenchant_item legs: send at src/net/online.ts:2783, string-guarded dispatch at server/game.ts:3960, COMMAND_NAMES at src/world_api.ts:278.
- disenchantResult is personal-scoped (pid gate at server/game.ts:4350, reaches only the disenchanter) and mirrored into lastDisenchantResult at src/net/online.ts:3412 with all five fields.
- The loot silent flag rides the JSON-forwarded event, so offline and online suppress the generic ding identically. The default stays loud for every non-professions grant, and professions_silent_loot.test.ts pins all the arms that matter: audio suppressed, text still prints, plain addItem stays loud, and the craft case proves call-site scoping (loud reagent event next to the silent output event).
- All 12 re-minted parity goldens are events-digest-only churn: zero state, rng.draws, or rng.digest changes across the whole set, exactly the signature of the optional silent field with no draw-order impact. Parity and architecture suites are green at the head.
- game_audio.test.ts covers all 3 gather node types, all 10 craft families plus the unknown-family fallback, and masterwork layering (both cues present, in order), against a real GameAudio with only sfx mocked.

One note:

- nit: tests/professions_audio_wiring.test.ts:69 the disenchantResult audio hop is source-guard only, which is fair given no UI calls disenchantItem() yet (and the test file says so honestly). When the UI trigger lands, give that hop a behavioral test.

No checks have run on this branch yet, so CI needs a kick before merge.

EnriqueGF (approved):
Reviewed the full diff, the local `CLAUDE.md` files for `src/sim/`, `src/ui/`, `src/world_api/` and `server/http/`, and ran the checks myself. This is clean, well-scoped work. Approving.

## What I verified

**Ran locally**
- `npx tsc --noEmit`: clean.
- Guard tests: `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`, the missing artifact was my worktree, not the branch), `tests/world_api_parity.test.ts`, `tests/command_schema.test.ts`, `tests/parity` (all golden scenarios), `tests/sfx_manifest.test.ts`, `tests/game_audio.test.ts`, plus the two new suites and the 21 `tests/professions*` files. All green.
- Re-ran `npm run sfx:manifest`: `sfx_manifest.generated.ts`, `runtime-pack.json` and `sfx_gain_ceiling.generated.json` reproduce byte-identically, so nothing was hand-edited.
- Scanned every added line for em/en dashes and emoji: none.

**Checked by reading the real code, not the description**
- *Golden parity churn.* The `events` digest moves in 12 scenarios while `state` and every `rng` digest/draw count stay identical. That is exactly the signature of adding an optional field to the `loot` event shape and nothing else. No gameplay or draw-order change.
- *Silent grants always have a paired cue.* `harvestNode` emits `gatherResult` unconditionally in the same block as both its grant arms (`gathering.ts:315,331,356`), and `resolveCraftForRecipe` is reachable only through `Sim.craftItem`, which always emits `craftResult` (`sim.ts:6797`). So no silent grant can end up with no sound at all. The `craftSuccess` fallback to `ui_loot_item` on an unrecognized family closes the last hole.
- *No masterwork double-play.* There is no `case 'masterwork'` in `hud.ts`, so `if (ev.masterwork) audio.masterwork()` off `craftResult` fires exactly once. The comment claiming this is accurate.
- *`silent` survives the wire.* Per-player events are pushed by reference and `JSON.stringify`d whole (`server/game.ts:5789,5816`), with no field projection, and `src/sim/sim.ts` holds the only two `type: 'loot'` emit sites. The fix behaves identically offline and online, which is the part that would have been easy to get wrong.
- *`disenchant_item` correctly stays out of `HEAVY_SELF_CMDS`.* Success grants an item, so the `loot` entry in `HEAVY_SELF_EVENTS` already dirties the self snapshot; every failure arm returns before `removeEnchantableItem`, so there is no state to refresh. Same shape as `craft_item` and `harvest_node`.
- Parity pin, `COMMAND_NAMES`, and the command-schema counts are all updated in the same change. `disenchantResult` is text-free, so no matcher entry is owed.

## Findings

**Nit (low, pre-existing pattern).** `Sim.disenchantItem` emits with `pid: meta?.entityId`. When `players.get()` misses, `pid` is `undefined`, and `eventAnchor` returns `null` for this shape (`server/game.ts`, the `return null; // chat/log etc: broadcast` arm), so a personal disenchant failure would fan out to every session and each `ClientWorld.applyDisenchantResultEvent` would clobber that player's `lastDisenchantResult`. Only reachable on an unresolvable pid, and `craftItem` above it has the identical shape, so this is not a regression you introduced. If you want to close it cheaply, guarding the emit on `if (meta)` (the way the masterwork emit right below already does) is a one-line fix. Happy either way.

**Note, not a request.** Shipping `audio.disenchant()`, the `disenchant_item` command and the `disenchantResult` event with no UI trigger means that path is untriggerable in-game for now. The PR body calls this out, the docs line carries the caveat, and it is covered by tests, so I am fine with it landing ahead of the trigger. Worth a tracking issue so the deferred UI pass does not get forgotten.

FernandoX7 (commented):
This is thorough and clean. I checked the two things that matter most here: coverage of every profession action, and removal of the old placeholders.

**Coverage is complete.** Every profession result event now has a dedicated cue:
- `gatherResult` -> `audio.gather(nodeType)` for ore/wood/herb, plus a layered `gatherRareTier` stinger on a rare+/rare-event roll (`src/ui/hud.ts` case `gatherResult`).
- `craftResult` (ok) -> `audio.craftSuccess(professionId)` per CRAFT_RING family, with `audio.masterwork()` layered on top of a proc, not replacing it.
- `disenchantResult` / `salvageResult` / `enchantResult` (log-sink arm only) -> `audio.disenchant()` / `audio.salvage()` / `audio.enchant()`. All three are reachable in-game via `bag_item_action_menu.ts`, so none of this is latent wiring.
- Fishing (`fishReel` et al.) was already real in base and is correctly left untouched.

All 10 `craftByFamily` ids match `CRAFT_RING` exactly and every recording is present; the three families without recipes yet (inscription, jewelcrafting, enchanting) are recorded ahead, which is fine. Confirmed no recipe has `professionId: 'enchanting'`, so reusing `ui_craft_enchanting` for both the (never-firing) family slot and `audio.enchant()` is safe.

**Old cues are removed.** `ui_gather_strike.mp3` is deleted, `gatherStrike()`/`gatherRare()` are gone from `audio.ts`, and there are zero references left in src, the manifest, the prompts, the gain ceiling, or the docs. `ui_gather_rare` is re-recorded and repurposed as the `rare` tier. The old craft double-ding (the grant's loot event and the `craftResult` handler both playing `lootItem`) is what the `silent` flag fixes, and it's threaded consistently through every professions grant site (gathering, crafting, enchanting, salvage).

Determinism holds: across all 12 regenerated goldens only the `events` digests moved, `draws`/`drawDigest`/`state` are byte-identical, which is exactly the signature of an event-shape-only change. `gatherRareTierFor` is sound because `gatherResult.rareEvent` is typed `| null` and emitted as `rareEvent ?? null`, never `undefined`, and it has decisive tests including the rare-event-forces-epic and legendary-precedence edges.

Two nits, neither blocking:

- `tests/professions_audio_wiring.test.ts` source-pins the gather, craft, masterwork and disenchant hud wiring but not salvage or enchant. The `audio.salvage()` / `audio.enchant()` calls in `hud.ts` are correct, but they're the only two result cues with no pin, and the disenchant block is a ready template to copy. (nit, test symmetry)
- `src/world_api.ts:183` re-exports `DisenchantResultView` from the professions facet, but `online.ts` imports it straight from `../world_api/professions`, so nothing consumes the barrel re-export. Harmless, and it matches its already-re-exported siblings, so keep it or drop it. (nit)

One informational note for the "all profession sounds" question: `gatherCast` (the cast-start affordance) is still the synth placeholder. That's called out as deliberately out of scope here, and it's a start cue rather than a result cue, so I'd leave it. Every completion/result sound is covered.

jamiecypher:
Pulling this back for rework: reconsidering the gather audio design (a multi-hit chopping cue synced to the cast, rather than a single stinger at harvest completion) before resubmitting. Will reopen or open fresh once that's built out.

jamiecypher:
Reopening as draft instead, not closed: still the same PR, holding it here while I rework the gather audio design (multi-hit chopping cue synced to the cast vs. the current end-of-harvest stinger).

FernandoX7 (commented):
Re-reviewed at `37e2206e9` after the gather-cast and fishing commits landed on top of my last pass.

The sim-side work is right and I re-verified the load-bearing claims myself. The `silent` opt is threaded consistently through all four professions grant sites (gathering, crafting, enchanting, salvage), the "You receive:" text line survives everywhere, and the determinism story holds: across the regenerated goldens only the `events` digest moved, `state` and `draws`/`drawDigest` are byte-identical. The gate order in `harvestNode` is as the commit claims, every denial (dead, busy, too far, not respawned, no tool, bags full) returns before the single `castStart` emit, so a denied harvest can never fire the cast cue.

The problem is downstream of all that: six of the cue keys `audio.ts` plays have no catalog entry at this head, so they are permanent silent no-ops.

```
UI_CUES keys missing from SFX_FIXED_CATALOG_KEYS and runtime-pack.json at 37e2206e9:
  ui_fish_bite, ui_fish_cast, ui_fish_reel,
  ui_gather_cast_ore, ui_gather_cast_wood, ui_gather_cast_herb
same check against release/v0.30.0: none missing
```

`playUi` resolves an unknown key to no entry, adds it to `failedLoads`, and returns (`src/game/sfx.ts:519` into `loadBuffer` at `:235`), so this fails silently rather than loudly.

### blocking

**Fishing is now silent.** `src/game/audio.ts:63-65` still plays `ui_fish_cast` / `ui_fish_bite` / `ui_fish_reel`, but `scripts/sfx/ui_sfx.mjs` deleted their synth specs and their `MASTER_GAINS_DB` rows and nothing re-added them to the `SFX` array in `sfx_prompts.mjs`. Those three keys are in the base catalog and gone from this branch's, so `fishCast()`/`fishBite()`/`fishReel()` all no-op. The eight real recordings this PR adds (`ui_fish_cast_1/2`, `ui_fish_bite_1/2/3`, `ui_fish_reel_1/2/3`) are never fetched, because runtime variant discovery only walks stems whose key is already in the catalog (`sfx_manifest_builder.mjs` `parseCatalogSfxVariantStem(value, catalogKeys)`). That is also why `sfx:check` stayed clean: an mp3 whose key is not in the catalog is invisible to the conformance inventory rather than flagged as an orphan.

Worth calling out that `fishBite` is the one cue in this family deliberately routed through the ungated `play()` arm (`audio.ts:334`) because the bite opens the live reel window. Losing it removes something the player acts on, not just polish.

**The node-typed gather-cast cue never plays.** `src/game/audio.ts:76-80` maps to `ui_gather_cast_ore` / `_wood` / `_herb`, none of which were added to `sfx_prompts.mjs` (the new block at `:732-775` covers the harvest, rarity, craft and enchanting keys but not these). `gatherCast()` at `audio.ts:327` only falls back to the registered flat `ui_gather_cast` when `nodeType` is undefined, and `harvestNode` is the sole gather `castStart` emit site and always sets `gatherNodeType` (`gathering.ts:390-396`), so the fallback is unreachable and the tool-out cue is silent on every gather. The six `ui_gather_cast_*` mp3s ship in `public/` and are never referenced. Base played `ui_gather_cast` here, so this one is a regression too.

Both are the same fix: `sfx_prompts.mjs` entries for `ui_fish_cast`/`ui_fish_bite`/`ui_fish_reel` (`custom: true`, matching the real takes) and for `ui_gather_cast_ore`/`_wood`/`_herb`, then regenerate. The count pins in `tests/sfx_manifest.test.ts:165-168` and the `UI_SFX_CATALOG` length in `tests/game_audio.test.ts:293` move with it.

### should-fix

**Nothing pins that every `UI_CUES` value has a clip.** This is why the gate stayed green through both misses. `tests/game_audio.test.ts` mocks `sfx`, so it asserts the key *string* and never that the key resolves; `tests/sfx_manifest.test.ts` pins totals, not membership; `sfx:check` is catalog-driven as above. One assertion that every leaf string under `UI_CUES` is in `SFX_FIXED_CATALOG_KEYS` would have caught both of these at authoring time and would close the class, not just these six. Given `UI_CUES` is already shaped as nested const objects, walking it is a few lines.

**`silent: opts?.silent` is set unconditionally on every loot event.** `src/sim/sim.ts:6779` and `:6838` write the key on every grant, professions or not. The parity canonicalizer maps `undefined` to `null` and keeps the key under `omitDefaults: false` (`tests/parity/trace.ts:79`, `:142`), so every loot event's digest moved, which is what dragged ten goldens with no professions content in them (`bank_round_trip`, `party_loot`, `inventory_vendor`, `quest_collect_turnin`, and so on) into the regen. `addItem` is the only changed code those scenarios reach, so that is the whole cause. A conditional spread (`...(opts?.silent ? { silent: true } : {})`) leaves every non-professions golden byte-identical, and would shrink the rebase surface that has now churned goldens three times, one of which (`inventory_vendor.json`) is a live conflict.

### conflicts vs release/v0.30.0

Two real ones, only one of which is code:

- `src/sim/professions/enchanting.ts`: release extracted the inline enchanting skill-gain block into `grantEnchantingSkill(ctx, meta, enchantGainTier(enchant))`. Same shape as the `enchantedPayloadFor` resolution you did last rebase. Take release's helper call and keep `{ silent: true }` on the `addItemInstance` above it.
- `tests/parity/golden/inventory_vendor.json`: regen churn, `UPDATE_PARITY=1` after the merge.

Everything else auto-merges.

### nits

- `docs/design/sound_effects.md:350-352` still lists `ui_fish_cast`/`ui_fish_bite`/`ui_fish_reel` as "PLACEHOLDER (deterministic synth, issue #2208)" after this PR deleted those synth specs and shipped real takes. Same oversight as the missing catalog rows, and the row for `ui_gather_cast` could note that the per-node-type recordings take over.
- `tests/professions_audio_wiring.test.ts` now pins gather, gather-cast, craft, masterwork and the `ev.silent` loot guard, but `audio.salvage()` and `audio.enchant()` are still the only two result cues with no source pin. Carried over from the last pass, and the `disenchantResult` block is a ready template.

jamiecypher:
Thanks for catching this, real miss on my end across multiple prior review passes.

Fixed at `6d454c44c`:

- **Blocking**: added the 6 missing `sfx_prompts.mjs` entries (`ui_fish_cast/bite/reel`, `ui_gather_cast_ore/wood/herb`, all `custom: true`) and regenerated the manifest. Verified all 6 now resolve real recordings with the right variant counts (bite×3, cast×2, reel×3, gather-cast×2 each).
- **Should-fix**: switched both `silent: opts?.silent` sites in `sim.ts` to a conditional spread, so a non-professions grant no longer carries the key at all. Confirmed it actually shrinks the churn: this rebase's own regen produced zero further golden diff.
- **New guard**: added a completeness test in `game_audio.test.ts` that walks every `UI_CUES` leaf against the real catalog, so this exact bug class fails loudly next time instead of shipping silent. Verified it actually catches it by reverting the fix locally and confirming the new test fails.
- **Nits**: fixed the stale PLACEHOLDER wording in `docs/design/sound_effects.md`, added the missing `salvageResult`/`enchantResult` source pins in `professions_audio_wiring.test.ts`.
- **Conflict vs current release/v0.30.0**: took your `grantEnchantingSkill` helper, kept the `silent: true` fix on the `addItemInstance` call above it.

Full suite green, tsc clean, parity clean.

FernandoX7 (commented):
Re-reviewed at `6d454c44c`. Everything from the last pass is closed, and I checked each one against the branch rather than the description.

**Both blockers are fixed, and the fix holds end to end.** The six keys that had no catalog entry (`ui_fish_cast`/`ui_fish_bite`/`ui_fish_reel`, `ui_gather_cast_ore`/`_wood`/`_herb`) now have `sfx_prompts.mjs` rows and land in `SFX_FIXED_CATALOG_KEYS`. I followed the resolution past the catalog: every clip entry points at a file that exists (`ui_fish_cast` resolves to `ui_fish_cast_1.mp3` with variants 1,2; `ui_gather_cast_ore` to `_1` with 1,2), and across all 217 clips every variant's `sha256` and `bytes` in `sfx_manifest.generated.ts` and `runtime-pack.json` matches the shipped file byte for byte, so nothing here is hand-edited or stale. Zero orphans: all 428 mp3s under `public/audio/sfx` are referenced by the manifest, and the 45 added files are all mono 192 kbps 44.1 kHz, matching the asset standard.

**The membership guard is the right shape.** `tests/game_audio.test.ts` "every UI_CUES leaf key resolves to a real catalog entry" walks the nested objects and the array, so it closes the class rather than these six cases. I mutation-checked it by adding an unregistered leaf to `UI_CUES`; it fails, as it should.

**The conditional `silent` spread did what it needed to.** Golden churn dropped from twelve files to two, both professions scenarios, and in those only the `events` digests moved: `state` and `rng.draws`/`digest` are byte-identical. Every one of the five silenced grant sites has a matching dedicated cue on a reachable path (both `disenchantResultToast` and `applyEnchantResultToast` return `sink: 'log'` for `ok`, so the success arm always plays), and `serializeEventFragments` is a plain per-event `JSON.stringify`, so `silent` and `gatherNodeType` ride the wire online without any extra plumbing.

Conflicts are gone: `origin/release/v0.30.0` is an ancestor of the head and `git merge-tree` is clean. Both nits from last time landed (the `sound_effects.md` rows, the salvage and enchant hud pins). Locally the professions, gather, craft, enchant, sfx, audio and parity suites pass, and `biome ci` over every changed file exits 0.

### the one thing left

CI has never run on this branch. Both workflow runs on `6d454c44c` are sitting at `action_required` because this is a fork PR from a first-time contributor, which is why `gh pr checks` reports no checks and the merge state stays `UNSTABLE`. Nothing for you to do; it needs a maintainer to release the run. Code-wise I have no objection to merging once that is green.

### nits

- `src/game/audio.ts:358`: the comment says an unknown craft family "is a silent no-op rather than a throw", but the code falls back to `UI_CUES.lootItem`, which is what both the docs table and the test assert. Comment drift.
- `scripts/sfx/sfx_prompts.mjs:785`: "No client UI calls disenchantItem() yet as of 2026-07-18" is stale. `src/ui/bag_item_action_menu.ts:105` wires it, and your own comment in `professions_audio_wiring.test.ts` says as much.
- `tests/professions_silent_loot.test.ts:17`: the header lists "enchant apply" among the paths that "stay loud", but `src/sim/professions/enchanting.ts:537` silences it. It is also the only one of the five silenced grant sites with no case in that file, and the disenchant test right below is a ready template.

### notes, nothing to change here

- The startup preload set grows by about 919 KB and 40 files (2.64 MB to 3.53 MB): `preloadForSfx` classes every `ui_` key as startup and `preloadStartup()` fetches every variant of every startup clip. That follows the existing rule rather than breaking it, and it is a background fetch after the audio context comes up rather than a load-time block, but a third more eager audio for cues that cannot fire in the first minute is worth knowing. The multi-take profession cues are the obvious candidates if it ever needs trimming.
- Fishing is now the one profession action still stacking a dedicated cue on the generic loot ding: `src/ui/hud.ts:9441` plays `audio.fishReel()` while the grant at `src/sim/professions/fishing.ts:269` stays loud. That is the exact pairing this PR silenced for gathering, and the two comments now argue opposite sides of it. Pre-existing, and defensible if the reel is the physical action rather than the notification, but the real recording makes it louder than the synth placeholder was.

FernandoX7:
Merged `release/v0.30.0` into the branch and pushed the result (`19cf839ee`), so CI has a conflict-free tip to run against. Three conflicts, all from the OSSBrain merge (#2336):

- `src/sim/sim.ts` and `src/sim/sim_context.ts`: release added `equipBag`/`equipItem`/`unequipItem` passthroughs in the exact slot between `addItem` and `addItemInstance` that this branch was editing to thread `opts.silent`. Kept both sides.
- `public/audio/sfx/runtime-pack.json`: only the `bundleId`/`catalogHash` pair actually conflicted, the clips section auto-merged correctly including release's `quest_ready` gain change. Regenerated the pack rather than picking a side, so the bundle hash covers both sets of changes.

One thing worth knowing from that regen: it also moved six `SFX_GAIN_LIMITS` rows (`ui_fish_cast`/`_bite`/`_reel`, `ui_gather_cast_ore`/`_wood`/`_herb`) off the flat 1.0 default onto their measured ceilings. That was stale before the merge rather than caused by it. `build_sfx_manifest.mjs` writes the manifest first and measures new ceilings second, so the first run after registering a custom key leaves its limit at the default, and a second run is what converges it. Nothing audible changes (none of those keys has an `sfx_gain_map.json` entry, so the resolved gain is 1 either way) and no test catches it, but it is worth a second `npm run sfx:manifest` next time custom keys land.

`a5016ffe1` closed all three nits from my last pass, and the new apply-enchant silent-loot case is the right shape.

Checked on the merged tip: the full suite comes out identical to the pristine `release/v0.30.0` tip in my environment (same file and test failure set, all from optional deps I do not have installed locally, none of it branch-related), with the only delta being this branch's own passing tests. `tsc` is clean, `biome ci` passes on every touched file, and both parity goldens still hold.

FernandoX7:
The red shard was not this branch. Shard 4 failed on `tests/net_pipeline_stats.test.ts`, which is byte-identical to `release/v0.30.0` here and untouched by this PR (the job name "PR gate (English-only legal)" is the i18n tier, not the failure).

It is a real flake in that test's clock guard, and I pushed the fix as `5ddc7338f`.

The test rewinds `lastSnapAt` by 900 ms to fake a stall, and `src/net/online.ts` reads `lastSnapAt > 0` as its no-prior-snapshot sentinel, so the rewind has to land strictly positive. The guard waited for `performance.now()` to pass 901, one millisecond of headroom, with a single `setTimeout`. That is not enough: `setTimeout` coerces its delay to an integer and fires on libuv's millisecond-truncated loop clock, so one fractional sleep lands up to about 2 ms short of what `performance.now()` then reads. Measured over fresh processes, the old guard leaves a negative rewind about once in 25, which nulls the gap sample and gives exactly the `gapMs.count` 0 that CI reported.

The fix polls the clock until the invariant actually holds instead of sleeping once, keeps a 300 ms margin, and asserts the precondition so a future failure names the clock rather than the recorder. It reproduces only in a fresh worker that reaches the file early, which is why a single-file local run always passes (vitest spends over a second in startup, so the guard never sleeps at all).

Worth knowing that this one lives on the release branch, so other open PRs can draw the same red shard until this merges.

PR #2135, opened 2026-07-18
fix(sfx): let the Training Dummy keep its crit ding, only silence the hurt bark

jamiecypher:
## Summary

Follow-up to #2116. That PR excluded the Training Dummy from both crit-reaction cues (the generic `combat_crit` ding and the mob-voice hurt bark), which was wider than intended: the plain crit ding is useful feedback while testing rotations against the dummy, so only the pained hurt-bark vocalization should be suppressed (the dummy soaks hits for the damage meter and was never meant to react like a real fight with a pained bark).

`shouldPlayCritSfxForTarget` goes back to excluding only bosses (the ding stays gated for a boss, a crit sting is the wrong emotional beat mid-boss-fight). The dummy exclusion moves to `mobVoiceActionForDamage`, which now checks the dummy flag directly so it is skipped only on the hurt-bark path, not the ding.

Based on `release/v0.28.0` (branched from PR #2133's catch-up merge, so #2116 is already present to build on).

## Related issues

Follow-up to #2116.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx tsc --noEmit`, `npx vitest run tests/combat_sfx.test.ts`, full `npm test -- --maxWorkers=4` (1328 files / 16130 tests, 0 failures), scoped `npx @biomejs/biome check --write` on both touched files.
- Manual steps: none needed beyond the updated unit coverage (both branches of the split are directly asserted).

### Quality
- [x] The full gate passes (run manually step-by-step; `ci:changed` false-positive-floods on a diverged branch, a known separate issue, not introduced here).

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/ui/combat_sfx.ts
+++ b/src/ui/combat_sfx.ts
@@ -262,27 +262,36 @@ export function mobVoiceCueWithFallback(
   return mobVoiceCue(templateId, 'attack', hasCue);
 }
 
-/** Gates BOTH crit-reaction cues in hud.ts: the generic `combat_crit` ding
- *  (played directly whenever this returns true) and the mob-voice hurt bark
- *  (via mobVoiceActionForDamage below). A boss or the Training Dummy gets
- *  neither: a boss because a crit sting is a wrong emotional beat mid-boss-fight,
- *  the dummy because it soaks hits for the damage meter and was never meant
- *  to react like a real fight, the ding included. */
+/** Gates the generic `combat_crit` ding in hud.ts (played directly whenever
+ *  this returns true). A boss gets none: a crit sting is a wrong emotional
+ *  beat mid-boss-fight. The Training Dummy DOES still get the ding (2026-07-19
+ *  follow-up to #2116: the dummy soaks hits for the damage meter and was
+ *  never meant to react like a real fight with a pained hurt bark, but the
+ *  plain crit ding is fine and expected feedback while testing rotations
+ *  against it; see mobVoiceActionForDamage below for the hurt-bark-only
+ *  exclusion). */
 export function shouldPlayCritSfxForTarget(target: Entity): boolean {
-  return (
-    target.kind !== 'mob' || (!MOBS[target.templateId]?.boss && !MOBS[target.templateId]?.dummy)
-  );
+  return target.kind !== 'mob' || !MOBS[target.templateId]?.boss;
 }
 
 /** The mob-voice action a damage event's target should react with, or null
- *  for anything that isn't a crit against a non-boss, non-dummy mob (a miss,
- *  an ordinary hit, a player target, a boss immune to crit stingers, the
- *  Training Dummy, whose whole point is soaking hits for the damage meter
- *  without reacting like a real fight). Callers still gate the actual play
+ *  for anything that isn't a crit against a non-boss mob (a miss, an
+ *  ordinary hit, a player target, a boss immune to crit stingers), OR the
+ *  Training Dummy specifically: it still gets the plain combat_crit ding
+ *  (shouldPlayCritSfxForTarget above), just never the pained hurt-bark
+ *  vocalization, since it soaks hits for the damage meter and was never
+ *  meant to react like a real fight. Callers still gate the actual play
  *  through shouldPlayMobVoiceSfxForEntity (the Nythraxis mute list) before
  *  using the resolved cue. */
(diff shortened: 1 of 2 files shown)


Rubsey (commented):
The split itself is right, and the tests pin both arms decisively at the head: shouldPlayCritSfxForTarget back to boss-only, the dummy check moved into mobVoiceActionForDamage (the flag is real: dummy at src/sim/types.ts:1094, set only on training_dummy), combat_sfx.test.ts asserts both the ding-fires arm (toBe(true) for the dummy) and the bark-suppressed arm (toBeNull()).

One thing needs real care before merge:

- blocking: this branch truly conflicts with release/v0.28.0 in src/ui/combat_sfx.ts and tests/combat_sfx.test.ts (the base took #2116 via the #2133 squash-merge, so this branch's merge base predates it on both sides). The sneaky part, verified with merge-tree: the visible conflict hunks in combat_sfx.ts are comment-only, and the body of shouldPlayCritSfxForTarget silently auto-merges to the BASE version, with the dummy check back in, because this branch's boss-only body is byte-identical to the old merge base. A resolution that only clears the visible markers ships the exact behavior this PR exists to remove, with a doc comment that says otherwise. When you merge the base: take this branch's bodies for both functions wholesale, and keep this branch's side of the test conflict (the toBe(true) dummy assertion is the tripwire that catches the wrong body; the base side of that hunk asserts toBe(false), and taking it would turn the suite green around the regression). After resolving, re-run tests/combat_sfx.test.ts and eyeball both function bodies by hand.

Side note: the 46-file diff GitHub is showing is stale-merge-base noise (#2133's already-merged content); the real change is the one commit touching two files, which is what I reviewed.

EnriqueGF (approved):
Reviewed the full diff against `release/v0.28.0` and verified the change end to end. Approving.

## What I checked

- **Intent vs code.** The split does exactly what the description says. `shouldPlayCritSfxForTarget` is back to a boss-only gate, and the dummy exclusion moved to `mobVoiceActionForDamage` behind an explicit `MOBS[target.templateId]?.dummy` check. Net behavior: a crit on the Training Dummy now plays `combat_crit` and no hurt bark; boss behavior and every other mob are byte-for-byte unchanged.
- **Call-site blast radius.** `shouldPlayCritSfxForTarget` has exactly one consumer outside this module (`src/ui/hud.ts:8187`, the `combat_crit` play), so widening it cannot leak into any other cue. `mobVoiceActionForDamage` is likewise single-consumer (`hud.ts:8193`) and still funnels through `shouldPlayMobVoiceSfxForEntity` for the Nythraxis mute list.
- **Dummy silence elsewhere is preserved.** `src/ui/mob_idle_sfx.ts:48` already excludes `dummy` on its own, so the idle-bark sweep is unaffected by the gate widening. The `dummy: true` flag is on `training_dummy` only (`src/sim/content/zone3.ts:93`), so no other mob changes behavior.
- **Invariants.** `src/ui/` only, no `src/sim/` or server touch, so no determinism, `IWorld` parity, or server-authority surface. No player-visible strings, so no i18n scope. No dashes or emojis in added lines, no secrets, no generated files hand-edited.

## What I ran

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/combat_sfx.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts`: 305 passed.
- `tests/localization_fixes.test.ts`: passes after `npm run i18n:gen` (the first run failed only on a missing `src/ui/i18n.status.json` in my worktree, which `pretest` normally generates; not a PR issue).

Test coverage is decisive: both arms of the split are asserted directly, and the dummy now has one assertion per path (`true` for the ding, `null` for the bark), so a future regression that re-merges the two gates fails one of them.

## Notes, non-blocking

- One thing I chased down and am satisfied with: restoring the ding on the dummy is the highest sustained-crit scenario in the game, and the `combat_crit` play at `hud.ts:8188` passes no explicit `cooldown`. `sfx.playAt` applies a 0.03s per-key default (`src/game/sfx.ts:436`), and crits arrive at swing cadence, so there is no spam risk here. No change needed.
- Minor coupling to be aware of rather than fix: `mobVoiceActionForDamage` now reaches through `shouldPlayCritSfxForTarget` for the boss half of its gate, so a future change to the ding policy would silently move the bark policy with it. The doc comments call this out well enough for now, and the tests would catch it.

PR #2136, opened 2026-07-19
perf(server): cache party frame projections per broadcast

Rubsey:
## Summary

- Cache common party-frame member projection once per member per broadcast.
- Keep Temporal Echo visibility viewer-specific, including spectator views anchored to the observed player.
- Reset projection and aggregate caches on every broadcast, including repeated broadcasts at the same sim tick.
- Preserve the existing offline projection path and party aura ordering/cap behavior.

The decisive integration test reduces expensive projection reads from 9 to 3 for a three-member party with three viewers. At raid size 10, this component falls from 100 to 10 projections per broadcast. Viewer-specific object creation and JSON serialization remain outside this PR, so this does not claim the same reduction for total broadcast time.

## Type of change

- [x] Tests

## How was this tested?

- Commands:
  - `npx vitest run tests/party_frame_info.test.ts tests/party_frame_projection.test.ts tests/snapshots.test.ts tests/architecture.test.ts`
  - `npm run build:server`
  - `npm run ci:changed`
  - `npm run gate`
- Results:
  - 161 focused and architecture tests passed.
  - Server build, TypeScript, changed-file checks, and independent review passed.
  - The full gate passed 16,118 tests and failed 2 existing `deploy_watchdog` timeout assertions. Both failures reproduce when that unchanged test file runs alone on macOS.
- Manual steps: N/A. This is a server hot-path change with deterministic integration coverage.

---

### Quality

- [ ] **The full gate passes.** The two unchanged deploy-watchdog timeout assertions described above remain red locally; all changed-path checks are green.

### Localization (i18n)

### Hygiene

- [x] I did not hand-edit generated files.

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -9,7 +9,6 @@ import {
   wireStreamerLinks,
 } from '../src/sim/account_flair';
 import { verifyChallenge } from '../src/sim/client_challenge';
-import { echoVisibleTo } from '../src/sim/combat/chronomancy';
 import { damageTakenWithin } from '../src/sim/combat/damage_history';
 import { rewindHealAmount } from '../src/sim/combat/rewind';
 import { DEEDS } from '../src/sim/content/deeds';
@@ -46,7 +45,6 @@ import { parseMoveInputFrame } from '../src/sim/move_input';
 import {
   partyFrameAbsorb,
   partyFrameAggroTargets,
-  partyFrameAuras,
   partyFrameIncomingHeals,
   partyFrameRole,
 } from '../src/sim/party_frame_info';
@@ -71,7 +69,6 @@ import {
   isDungeonDifficulty,
   MAX_LEVEL,
   type MobFamily,
-  PARTY_MEMBER_AURA_CAP,
   RUN_SPEED,
   type SimEvent,
   type SportRole,
@@ -166,6 +163,7 @@ import {
   ModerationService,
 } from './moderation_service';
 import { consumeMsgToken, createMsgRateBucket, type MsgRateBucketState } from './msg_rate_limit';
+import { PartyFrameProjectionCache } from './party_frame_projection';
 import { nextRaidResetMs } from './raid_reset';
 import { REALM, REALM_PUBLIC_ORIGIN, REALM_RESET_TIME_ZONE } from './realm';
 import { createSerialWriter } from './serial_writer';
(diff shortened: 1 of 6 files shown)


FernandoX7 (commented):
Solid change. The commit sequence is test-first, and the perf pin is decisive: the damageHistory getter instrumentation counts exactly one projection read per member per broadcast (3 instead of 9), and would catch a regression in either direction. I verified the load-bearing claims independently:

- The reordered aura pipeline (relevance filter plus stable priority sort up front, per-viewer Echo filter plus cap after) is behavior-identical to the old echo-filter-first path: both filters are per-element predicates, the sort is stable with the same comparator, and the cap counts only viewer-visible rows in both. The NaN/negative cap normalization in `partyFrameAurasForViewer` exactly matches the old `slice(0, cap)` semantics.
- `partyWire` is reachable only from `selfWireJson`, whose one call site is inside `broadcastSnapshots` after `beginBroadcast()`, so there is no stale-serve window between broadcasts, and `Party.id` is unique (monotonic `nextPartyId`), so the cache key is sound.
- Wire shape is unchanged and no `Entity` field is added, so no client or `blankEntity` work is owed. The spectator test pins the anchor-pid semantics correctly.
- Ran the four listed test files plus `tsc --noEmit` at the PR head in a clean worktree: 161 passing, typecheck clean. Merge is conflict-free against both `main` and `release/v0.28.0`.

Findings:

- should-fix (base branch): the contribution workflow bases PRs on the latest release branch, currently `release/v0.28.0`; this one targets `main`, which trails it. The merge into `release/v0.28.0` is clean, so retargeting is mechanical.
- should-fix `src/sim/party_frame_info.ts:66`: `partyFrameAuras` is now a verbatim parallel implementation of the relevance-filter, priority-sort, and summary pipeline beside `preparePartyFrameAuras`, with the offline path (`src/sim/sim.ts:8393`) on one and the server on the other, and nothing pinning the two to the same output. Reimplementing it as `partyFrameAurasForViewer(preparePartyFrameAuras(auras), undefined, cap)` is behavior-identical (checked, including the zero and negative cap arms) and makes the existing filter/sort/cap tests pin the one shared implementation.
- should-fix `tests/party_frame_projection.test.ts:33`: the fixture auras are already relevance-only and already in priority order, so deleting the `isPartyFrameRelevantAura` filter or the sort inside `preparePartyFrameAuras` would not fail any test on the server path (the old `partyFrameAuras` tests no longer cover it). The delegation above closes this for free; otherwise add one irrelevant maintenance buff and one out-of-order debuff to the fixture.
- should-fix `tests/party_frame_projection.test.ts:20`: every test uses a single party, so the per-party cache keying (`server/party_frame_projection.ts:52`) is unpinned; a regression that keyed the cache on a shared slot would serve party A's roster to party B and stay green. Cheap close: two parties with disjoint members in one `beginBroadcast`, asserting each viewer gets only their own.
- nit `server/party_frame_projection.ts:76`: the skip-a-null-member branch (`if (!projection) continue`) is untested; a projectMember returning null for one pid with siblings surviving would pin the old `.filter(Boolean)` behavior.
- nit `server/party_frame_projection.ts:64`: the prepared summary objects are pushed by reference into every viewer's output, so all viewers of a party share the same row objects within a broadcast. Safe today because the only consumer is JSON.stringify, but worth a one-line comment that summaries are immutable after prepare: a future in-place tweak before serialization would silently cross-contaminate viewers.

One non-finding for the record: nulling `partyFrameGlobalsCache` per broadcast on top of the retained tick key is a small correctness improvement over the old per-tick memo (same-tick rebroadcasts now see mid-tick mutations), and the new snapshots test pins it.

PR #2138, opened 2026-07-19
release: v0.27.2, iOS world-entry crash hotfix (memory trims, watchdog-safe prewarm, crash guard)

TrevCavill:
## Summary

Hotfix release for the iOS native (and mobile-web WebKit) crash where entering the world at any graphics tier above Low killed the WKWebView WebContent process and dumped the player back to the menu, looping through the News screen with no reachable graphics control. Root-caused on a physical iPhone 17 Pro to three stacked causes, each fixed as its own layer, and device-verified after each round:

1. **World-entry GPU allocation spike (memory ceiling).** Medium+ synchronously allocated a 2560 shadow map, three PMREM env cubemaps, ten splat textures, and three mip-chained 4K backdrops in the Renderer constructor. New `constrainedMemory` profile in `gfx.ts` (from the existing `isConstrainedBrowser` predicate) sheds the largest one-shot allocations on phone-class browsers: 1536/2048 shadow maps, no composer MSAA, 1.48 DPR cap, no backdrop mip chains, deferred non-default-biome IBL prefilter. Cosmetic sharpness only; fairness pins in `tests/gfx.test.ts` prove the actionable knobs are identical to the unconstrained tier.
2. **Prewarm main-thread occupancy (responsiveness watchdog).** `prewarmInitialScene` could hold the main thread for 12s of view builds plus up to 10s of shader linking with only microtask yields; iOS kills a WebContent process that stays unresponsive on that order regardless of device RAM. Constrained devices now run a deliberately minimal manifest (`CONSTRAINED_PREWARM_KEEP`) under a 5s budget with event-loop yields between entries, async compile ordered before the first full-scene pass, and sync link passes only where parallel compile is unavailable. Desktop budgets and ordering unchanged.
3. **No recovery path (the News-screen crash loop).** The kill fires no event, the shell reloads, and the resume marker re-entered the world straight into the same crash, cycling the welcome/news screen up to the attempt cap with the home-footer graphics selector unreachable. New world-entry crash guard (`src/game/entry_crash_guard.ts`, pure core + thin storage wrapper): a probe stamped before the scene build and cleared on survival/handled failure/page-hide; a still-armed fresh probe at boot steps the persisted preset down ONE tier on iOS/native runtimes, pins the auto default, drops the resume marker so boot lands on chrome with a reachable graphics control, and shows a start-screen banner (`entryGuard.*` keys filled across all 20 locales). This makes Auto self-correcting per device: the ladder stops at whatever tier the hardware actually survives.

Also merged **#2072** (`safeStartupGraphicsPreset`): the proactive Ultra/Advanced-to-High clamp now covers iOS Safari, not just the native shell, as the first line of defense.

Follow-up fix in the same branch: opening the More tray at Medium could still kill the process. Two contributors fixed: the in-world `applyBrowserEffects` re-stamp promised by the landing-stamp comment had been lost (its only caller was the Esc-menu handler), leaving the fx tier undeterministic in-game (fx-minimal is what strips backdrop-filter, the More tray's glass blur snapshots the live WebGL canvas per open); and the deferred env prefilters fired 12s/16s after scene build, inside both the player's first-interaction window and the crash guard's 20s stable window. Restored the entry re-stamp and moved the prefilters to 30s/40s.

Every stage emits `[entry-guard]` console diagnostics (entry preset, resolved tier and memory knobs, live fx tier, prewarm summary, prefilter timing, crash recovery) so future device kills can be isolated from a Web Inspector session.

Release mechanics: version set to 0.27.2 across package.json, Android (`versionName` 0.27.2 / `versionCode` 24), and iOS (`MARKETING_VERSION` 0.27.2 / build 27), including the Capacitor SPM 8.4.2 resolution from the verification builds. The client changes are standalone-safe for the native update: no `server/`, `src/sim/`, `src/net/`, or `src/world_api` changes, so the production server needs no deploy for app-store submission; deploying the site delivers the same protections to mobile-web players.

## Related issues

Closes #2072 (merged into this branch; the clamp ships here with original attribution).

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [x] Build, CI, or tooling (release version bump)
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npm run gate` (release-tier, on `release/v0.27.2`): green except two known environment-only failures on the dev machine, neither reachable from this diff: `tests/deploy_watchdog.test.ts` (hung-docker timing bounds; fails identically on unmodified `main` on the same machine) and `tests/discord_server.test.ts` (local `.env` sets `DISCORD_GUILD_INVITE`, overriding the pinned default; passes in a clean checkout). 16,144 tests pass, i18n release tier at zero pending rows, `tsc` clean, all five builds green.
  - New/updated suites: `tests/entry_crash_guard.test.ts`, `tests/entry_guard_banner.test.ts`, `tests/startup_graphics_safety.test.ts` (from #2072), `tests/gfx.test.ts` (constrained caps + fairness pins), `tests/resume_play.test.ts` (probe-clear wiring + marker-site count), `tests/perf_reporter.test.ts` (config version 15).
- Manual steps (physical iPhone 17 Pro, native app against a LAN server, Safari Web Inspector attached):
  - Before: Auto (resolves Medium) and any manual preset above Low crashed to menu at world entry, cycling the News screen; only Low entered.
  - After: Auto/Medium enters and plays across repeated attempts (`prewarm done: 1666ms/5000ms ... timedOut=[]`, then `world entry stable; probe cleared`).
  - Forced a crash scenario: the guard stepped the preset down one tier, landed on the start screen with the explanation banner (no News loop), and the footer selector was reachable.
  - More tray at Medium: reproduced the kill, then verified open/close after the fx re-stamp and prefilter rescheduling.

## Screenshots / recordings

N/A for the render/bootstrap changes (no visual redesign); the one new visible surface is the crash-recovery banner, which only appears after a process kill on a real device. Reviewers can trigger it by stamping a fresh probe in localStorage (`woc_entry_probe`) on an iOS user agent.

---

### Quality

- [x] **The full gate passes.** Green locally except the two environment-only failures documented above; CI has neither condition.

### Cross-platform

- [x] **Tested on desktop and mobile.** Device-verified on iPhone 17 Pro (native) across graphics tiers; desktop paths unchanged by design (all constrained-memory behavior is gated on the phone-class predicate) and covered by the unchanged desktop pins in `tests/gfx.test.ts`.
- [x] **Accessible.** The banner is `role="status"`, keyboard-dismissable, 40px tap target, safe-area aware.

### Localization (i18n)

- [x] **New player-visible strings follow the policy**, with the release-branch exception applied: this is a `release/**` branch, so the three `entryGuard.*` keys are filled across all 20 locales (release-tier gate at zero pending rows), with es_ES/fr_CA resolving through their dialect bases.
- [x] Numbers and interpolations go through the i18n formatters / `t()` values.
- [x] No sim/server-emitted player text changed.

### Hygiene

- [x] Generated files regenerated via their owning build steps (`npm run i18n:gen`), never hand-edited.


Rubsey (commented):
The recovery ladder, constrained graphics caps, and release wiring are well targeted. The release, type, browser, and focused test gates are green. I found a few things to address:

- **should-fix** `src/render/renderer.ts:1264`: constrained startup always prefilters Vale, but returning characters restore arbitrary saved positions. A character starting in Marsh or Peaks gets the correct sky dome with Vale IBL and sun rotation until the 30s or 40s deferred timer runs. Prefilter the player's initial dominant biome synchronously, then defer the other two, which keeps the same one-PMREM entry cost.
- **should-fix** `src/ui/entry_guard_banner.ts:28`: `/play` runs the same recovery path, lowers the preset, and clears the resume marker, but `play.html` has no banner shell or graphics selector, so this early return makes the recovery unexplained there. Add an equivalent shared recovery surface to `/play`; the current test at `tests/entry_guard_banner.test.ts:69` locks in the missing parity.
- **should-fix** `src/styles/index.extra.css:250`: the recovery banner and Discord CTA occupy the same fixed top-center slot. A recovered, logged-in, unlinked player causes `refreshDiscordStatus()` to reveal the CTA while the recovery banner is still visible, so the two overlap. Stack, offset, or suppress one while the other is open.
- **should-fix** `src/render/renderer.ts:3081`: no test invokes `prewarmInitialScene`, so the 5s cap, minimal keep list, compile-before-frame reorder, event-loop yields, and no-parallel-compile path can all regress while the added suites remain green. This control flow is the core of the watchdog fix; please pin it through an extracted planner/policy test or a constrained-browser regression.
- **nit** `src/ui/entry_guard_banner.ts:31`: the dynamic body is painted once and does not handle `woc:languagechange`. The title and dismiss button retranslate through `data-i18n`, but the body and preset label stay in the boot locale after a footer language change.

Non-blocking release note: the PR description says Android build 24 and iOS build 27, while the final checked-in values are Android 25 and iOS 28.

PR #2141, opened 2026-07-19
perf(net): cache stable snapshot timers

Rubsey:
## Summary

- Negotiate an exact `timerWire: 2` capability per recipient while preserving byte-compatible legacy snapshots for old clients and isolating unknown future markers.
- Serialize live aura, cooldown, and gathering-node timers as stable absolute schedules, cache their JSON until semantic state changes, and age omitted schedules client-side from authoritative simulation time.
- Keep legacy/stable entity variants shared and lazy without losing aura revisions across distance-tier deferral, spectator changes, linkdead resume, or mixed-client sessions.
- Harden timer parsing against negative, non-finite, and overflowing values, and pin the server layer boundary so authoritative code cannot import `src/net`.

In the 160-tick timer-heavy regression fixture, legacy aura serialization rebuilds more than 150 times while v2 rebuilds once. The v2 fixture payload is less than 35% of the legacy fixture payload. This is a focused fixture result, not a claim about whole-production bandwidth.

## Type of change

- [x] Bug fix
- [x] Tests
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/client_snapshot_timer_wire.test.ts tests/snapshot_timer_wire.test.ts tests/server/ws_auth.test.ts tests/snapshots.test.ts tests/security.test.ts tests/linkdead.test.ts tests/architecture.test.ts tests/mob_stall_parse.test.ts tests/party_frame_info.test.ts tests/party_frame_projection.test.ts --maxWorkers=4` (348 passed)
  - Independent review matrix (293 passed)
  - `npm run check:ts`
  - `npm run build:server`
  - `npm run ci:changed`
  - `git diff --check origin/main...HEAD`
- Manual steps: N/A. Wire behavior is covered with old/new server-client combinations, mixed recipients, spectators, resumes, dead timers, future markers, hostile values, and deferred entity updates.
- Local full-gate note: i18n freshness, malware scanning, and changed-file Biome checks passed. Vitest reproduced the two existing macOS `deploy_watchdog` timeout failures. The linked-worktree run then lost ignored i18n artifacts and produced invalid zero-test imports, so it was stopped. Fresh-clone GitHub CI remains required before merge.
- Dependency audit: the branch changes no dependencies. `npm audit --audit-level=high` reports the existing 10 transitive Solana/Reown findings (6 moderate, 4 high); the offered fix is a breaking force update.

---

### Quality

- [ ] **The full gate passes.** See the local full-gate note above; focused, type, build, lint, architecture, and security checks pass.

### Cross-platform

- [ ] N/A. No UI or platform-specific behavior changed.

### Localization (i18n)

### Hygiene

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -2,6 +2,10 @@ import { randomUUID } from 'node:crypto';
 import type { WebSocket } from 'ws';
 import { createBotDetector } from '#bot-detector';
 import {
+  STABLE_TIMER_WIRE_VERSION,
+  type StableTimerWireVersion,
+} from '../src/net/snapshot_timer_wire';
+import {
   type AccountFlair,
   type ChatSenderFlair,
   EMPTY_ACCOUNT_FLAIR,
@@ -167,6 +171,7 @@ import { PartyFrameProjectionCache } from './party_frame_projection';
 import { nextRaidResetMs } from './raid_reset';
 import { REALM, REALM_PUBLIC_ORIGIN, REALM_RESET_TIME_ZONE } from './realm';
 import { createSerialWriter } from './serial_writer';
+import { StableAuraWireCache, StableSelfTimerWireCache } from './snapshot_timer_wire';
 import type { Presence, PresenceStatus, SocialActor, SocialTransport } from './social';
 import { SocialService } from './social';
 import { PgSocialDb } from './social_db';
@@ -634,6 +639,10 @@ export interface ClientSession {
   // serialized form of each delta self field as last sent to this client;
   // a field is omitted from a snapshot while its serialization is unchanged
   lastSent: Record<string, string>;
+  // Recipient-negotiated timer representation. Legacy remains the default for
+  // old and unknown clients throughout a rolling deploy.
+  timerWireVersion: 1 | StableTimerWireVersion;
+  timerWireCache: StableSelfTimerWireCache;
   // arena readout is reconciled at UI cadence instead of snapshot cadence
   lastArenaWireTick: number;
   // Vale Cup readout, same idea at its own cadence (VC_WIRE_HZ)
@@ -703,6 +712,10 @@ export interface ClientSession {
 interface SentEntityVersions {
   idVer: number;
   dynVer: number;
+  // Stable timer-wire recipients diff aura composition separately from the
+  // ordinary dynamic record, so a deferred distance-tier update cannot lose an
+  // aura change. Legacy recipients leave this at 0.
+  auraVer: number;
   // sim tick of the last full/lite record, so distance-tiered rates hold
   // even when one broadcast covers several catch-up sim ticks
   sentAtTick: number;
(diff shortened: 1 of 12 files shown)


FernandoX7 (commented):
This is careful wire work and it held up under an adversarial read. Things I verified against the code rather than the description:

- The negotiation is fail-closed in both directions: exact `timerWire === 2` on the server (server/ws_auth.ts:247), exact `tw === 2` on the client with unknown future markers isolated from both decoders (src/net/snapshot_timer_wire.ts:17), and the capability actually rides the real connect path (src/net/online.ts:1569).
- The v2 aura encoder's field omission rules (value when 0, physical school, stacks above 1, charges, emp, src) match legacy `wireAura` field for field, so the decode contract is unchanged for both recipient kinds.
- The hourglass deadline math matches the sim exactly: `updateTimers` keeps recovering cooldowns at `DT * aura.value` while dead (src/sim/combat/auras.ts:133), `updateAuras` freezes aura remaining while dead (src/sim/combat/auras.ts:190), and the hostile suspension arm applies kind `incapacitate`, never `stasis`, so the encoder's `value > 1` guard is right (a rate of exactly 1 decodes identically either way). Mispredictions would self-correct anyway, since the deadline is recomputed from authoritative state every tick and any mismatch just degrades to legacy resend cadence.
- No double-aging: nothing in src/ui or src/game decrements aura remaining per frame, so the snapshot-cadence aging in `prepareSnapshotTimers` is the only writer.
- The deferred distance-tier path advances `known.auraVer` only on actual delivery (server/game.ts:5089), so a non-due tick cannot swallow an aura change.
- The new server host-layer import invariant in tests/architecture.test.ts is a good gate to leave behind.

Findings:

- should-fix: the PR targets `main`, but this repo integrates through the newest release branch (`release/v0.28.0`); `main` trails it. `git merge-tree` against `release/v0.28.0` is conflict free, so retargeting is mechanical.
- nit (server/game.ts:1109): `jsonWithField` on an empty `{}` input produces `{,"key":...}`, invalid JSON. Unreachable today because `dynamicFields` always emits x/y/z/f/hp/mhp, but a one-line guard or a comment stating that precondition would keep a future caller from tripping it.
- nit: six of the eight commits are title-only; house convention is a body on every commit.
- nit (tests/client_snapshot_timer_wire.test.ts:10): this adds another hand-rolled `bareClient` copy alongside the one in tests/snapshots.test.ts; fixture consolidation is tracked in #2088, just flagging the new copy so it gets swept there.

Test coverage is genuinely strong: coercion attempts and hostile values at the trust boundary, mixed legacy and v2 recipients sharing one entity cache, spectator anchor switches, resume in both directions, dead freeze and resume, real-sim hourglass acceleration, and a rebuild-plus-bytes pin. CI is green on this head, which covers the local full-gate caveat in the description.

PR #2146, opened 2026-07-19
test(perf): expand performance regression coverage to 160+ budget tests

jgyy:
## Summary

Expands performance-regression coverage well beyond the existing `mob_update_perf.test.ts` /
`tests/server/perf_gate.test.ts` / `hud_perf_budget.test.ts` gates, which only budgeted one sim
tick phase and left every real-measurement arm opt-in (never run by default). Adds **40 new perf
test files / 160 `it()` cases**, fixes two real regressions the new tests surfaced, and wires the
previously-opt-in real-measurement arms into CI.

```mermaid
flowchart TB
    subgraph SIM["Sim-side (server tick / offline-mode FPS): 121 tests, 34 files"]
        A["Combat resolution core\ndamage / heal / effects / procs / hit tables\n(15 tests)"]
        B["Targeting, spatial, pathing\nthreat / grid / A* / colliders / mob targeting\n(15 tests)"]
        C["Social systems at scale\nparty / chat / arena / dungeon finder / minigames\n(15 tests)"]
        D["Economy and items\nmarket / bank / loot / vendor / mail\n(15 tests)"]
        E["Progression, talents, deeds\ntalents / deeds / xp / professions / honor\n(15 tests)"]
        F["Instances, encounters, delves\ndungeons / world boss / Nythraxis / delve runs / AoE+projectiles\n(21 tests)"]
        G["Mob, pet, world-gen leaves\nlifecycle / social aggro / pet AI / terrain / aura stacking\n(15 tests)"]
        H["Aura tick phase\np.auras / mob.auras budget + scaling\n(2 tests, prior work)"]
    end

    subgraph RENDER["Render-side (real FPS): 39 tests, 16 files"]
        I["Nameplate + self-motion pure cores\n(15 tests)"]
        J["LOD, crowd, light, interp decisions\n(15 tests)"]
        K["Picking + pooled VFX allocation\n(14 tests, 1 file skipped: no real pool to test)"]
        L["Instancing draw-call budgets\n(6 tests, 3 files skipped: need browser image globals)"]
    end

    SIM --> BUGS
    RENDER --> BUGS

    subgraph BUGS["Real regressions found + fixed"]
        M["vendorInRange scanned every live entity\nper vendor transaction -> now spatial-grid bounded"]
        N["dungeon_finder per-tick match budget reset\nper anchor, not per pass -> anchors x budget blowup\n(300p backlog: ~415ms -> ~6ms)"]
    end

    BUGS --> CI

    subgraph CI["CI wiring: real measurement, no longer purely opt-in"]
        O["PERF_GATE_WALLCLOCK=1 now runs in\nnpm run gate + both CI test jobs\n(Node-only, generous headroom)"]
        P["New non-blocking perf-budget-report job\nin pr-ai.yml: real browser tour +\nHUD_PERF_BUDGET_TOUR=1 on every PR"]
    end
```

### Why this recipe (not `vitest bench()`)

Every new test follows the existing `mob_update_perf.test.ts` / `aura_tick_perf.test.ts` pattern
deliberately, not Vitest's built-in `bench()`: `bench()` has no CI pass/fail threshold, so it
can't gate a merge on its own. Each file instead does:

1. Build a deliberately worst-case population/scenario through real entry points.
2. Warm up, sample 40-60+ iterations, sort, take the **median** (rejects one-off GC/scheduler
   spikes from co-running Vitest workers).
3. Assert a generous absolute ms budget (~8-10x the observed healthy median), so it never flakes
   on slow/contended CI hardware but still catches a sustained order-of-magnitude regression.
4. A **scaling check**: double the population/depth, assert cost grows no more than ~3.5x. This
   is the piece a flat ceiling alone cannot provide — it catches an O(n) regression that goes
   quadratic even while both individual measurements stay comfortably under budget.
5. A non-vacuous shape-sanity assertion proving the worst-case scenario was actually built.

### Real regressions found and fixed

Two of the new tests surfaced genuine production bugs, fixed in this same PR:

- **`vendorInRange` (`src/sim/items.ts`)** scanned every live entity in the world on every
  vendor buy/sell/buyback. Now bounded via `ctx.grid.forEachInRadius`, matching the spatial-grid
  idiom already used in `interaction.ts`/`targeting.ts`/`pet_ai.ts`.
- **`dungeon_finder`'s per-tick matching budget (`src/sim/social/dungeon_finder.ts`)**:
  `FINDER_MATCH_NODE_BUDGET` reset to a fresh budget inside every `tryAssemble()` call instead of
  being shared across the whole `runMatching()` pass. A role-imbalanced backlog (no tanks/healers
  ever queued) let every anchor independently exhaust its own budget hunting for a composition
  that can never exist, so total per-tick cost scaled as `anchors * budget` instead of being
  capped — a 300-player backlog measured ~415ms, 8x over the 50ms/20Hz tick budget. Fixed by
  threading one shared `budgetBox` across the whole pass; the same scenario now measures ~6ms.

### CI wiring: closing the "opt-in only" gap

Both repo-existing wall-clock arms (`tests/server/perf_gate.test.ts` ARM C, `HUD_PERF_BUDGET_TOUR`
in `hud_perf_budget.test.ts`) were skipped by default because real measurement is
hardware-relative and can flake on a shared runner. That meant nobody actually ran them unless a
human remembered the env var.

- `PERF_GATE_WALLCLOCK=1` is Node-only (no browser) and carries generous headroom over its
  ceilings (observed added-p99 ~0.003ms vs a 1ms budget; observed tick p95 ~0.7ms vs a 40ms
  ceiling), so it now runs in `npm run gate` and both `ci.yml` PR-gate/release-gate test jobs.
- The real-browser frame budget (`HUD_PERF_BUDGET_TOUR=1` against a `perf_tour.mjs` artifact)
  needs a real Chrome + a running dev server, so it's added as a new **non-blocking**
  `perf-budget-report` job in `pr-ai.yml` (which is explicitly "informational, non-blocking" by
  charter, same as the existing `screenshots` job) — it runs on every PR and surfaces a real
  `frameP95`/DOM-write regression, but never risks turning a required check flaky.

### Coverage gaps closed vs. skipped honestly

A few planned files were skipped rather than forced into a vacuous test — flagged here instead of
padding the count:

- `tests/impact_fx_pool_perf.test.ts`: `impact_site.ts`/`impact_terrain.ts` turned out to be
  one-off static crater-geometry math for a single scripted world event, not a pooled system —
  there is nothing to burst or reuse.
- `tests/props_instancing_budget_perf.test.ts` / `foliage_instancing_budget_perf.test.ts` /
  `dungeon_instancing_budget_perf.test.ts`: these `build*()` functions need real browser
  image-decoding globals (`self`, `Image`/`createImageBitmap`) inside three.js's `GLTFLoader`
  texture pipeline that plain Node/Vitest can't provide. Not an instancing regression — a Node
  test-environment limitation.

## Related issues

N/A

## Type of change

- [x] Tests
- [x] Bug fix (the two regressions above)
- [x] Build, CI, or tooling (the CI wiring above)

## How was this tested?

- Commands: `npm run i18n:gen`, `npx vitest run` (full suite: 1385 files, 16351 tests, all green
  after i18n regen), `npx tsc --noEmit` (clean), `npx @biomejs/biome ci --changed
  --since=release/v0.28.0` (0 errors, pre-existing-style warnings only), `npm run security:gate`
  (PASS, 0 high findings), `npm run sfx:check` (PASS), `npm run build:env`, `npm run
  build:server`, `npm run build` (all succeed).
- `npm run test:browser`: 64/65 pass. The one failure
  (`armory_mobile_layout.browser.test.ts`, a real-Chromium pixel-geometry assertion, 394.6px vs a
  391px bound) is a pre-existing/environment sensitivity: the test file is byte-identical to
  `release/v0.28.0` (`git diff` shows 0 changes) and this PR touches zero CSS/render files.
- Manual steps: none (no player-visible UI change).

## Screenshots / recordings

N/A — no player-visible change.

---

### Quality

- [x] **The full gate passes.** Ran every `npm run gate` step individually (see "How was this
      tested?"); one pre-existing/environment-flaky browser test noted above, unrelated to this
      diff.

### Cross-platform

- [x] N/A. No UI change.
- [x] N/A. No UI change.

### Localization (i18n)

### Hygiene

      any production path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,7 +172,16 @@ jobs:
       # suites need the generated artifacts no matter which shard they hash
       # into), so this job needs no explicit i18n:gen step; the freshness diff
       # lives in pr-checks.
+      # PERF_GATE_WALLCLOCK=1 turns on tests/server/perf_gate.test.ts's ARM C: a
+      # real process.hrtime.bigint() comparison of the API-pipeline onion vs the
+      # bare legacy path, plus a real Sim tick-loop p95 check. It is Node-only (no
+      # browser), skipped by default because it is hardware-relative, but its
+      # ceilings carry generous headroom (observed added-p99 ~0.003ms vs a 1ms
+      # budget; observed tick p95 ~0.7ms vs a 40ms ceiling) so it stays safe to run
+      # on every PR shard instead of only via a human remembering the env var.
       - name: Run tests (PR tier, shard ${{ matrix.shard }} of 4)
+        env:
+          PERF_GATE_WALLCLOCK: '1'
         run: npm test -- --shard=${{ matrix.shard }}/4
 
   # Runs in parallel with pr-gate: same if-condition, no needs edge in either
@@ -241,8 +250,10 @@ jobs:
     runs-on: ubuntu-latest
 
     # The release tier; this single flag turns on every release-only check.
+    # PERF_GATE_WALLCLOCK: see the matching comment on pr-gate's test step above.
     env:
       I18N_RELEASE_TIER: '1'
+      PERF_GATE_WALLCLOCK: '1'
 
     # The same 4-shard matrix as pr-gate: the test step partitions across the
     # four runners, while every serialized check and build below carries a
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,10 +179,14 @@ jobs:
       # ceilings carry generous headroom (observed added-p99 ~0.003ms vs a 1ms
       # budget; observed tick p95 ~0.7ms vs a 40ms ceiling) so it stays safe to run
       # on every PR shard instead of only via a human remembering the env var.
+      # --maxWorkers=2 bounds vitest's own parallelism (root CLAUDE.md: an
+      # unbounded run flakes heavy suites under core contention), matching
+      # scripts/gate.mjs, so ARM C's hrtime measurements above and the *_perf.test.ts
+      # wall-clock budgets are not sharing runner cores with sibling worker threads.
       - name: Run tests (PR tier, shard ${{ matrix.shard }} of 4)
         env:
           PERF_GATE_WALLCLOCK: '1'
-        run: npm test -- --shard=${{ matrix.shard }}/4
+        run: npm test -- --shard=${{ matrix.shard }}/4 --maxWorkers=2
 
   # Runs in parallel with pr-gate: same if-condition, no needs edge in either
   # direction. The default pull_request merge-ref checkout MUST be preserved
@@ -310,7 +314,7 @@ jobs:
       # artifacts in EVERY shard so the artifact-dependent suites pass no
       # matter which shard they hash into.
       - name: Run tests (release tier - 21-locale + empty-pending, shard ${{ matrix.shard }} of 4)
-        run: npm test -- --shard=${{ matrix.shard }}/4
+        run: npm test -- --shard=${{ matrix.shard }}/4 --maxWorkers=2
 
       - name: Typecheck
         if: matrix.shard == 1
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -118,6 +118,84 @@ jobs:
           SHOTS_DIR: pr-shots
         run: node scripts/pr_comment_shots.mjs
 
+  # Real-browser frame-budget signal for every PR. hud_perf_budget.test.ts's ARM 3
+  # (HUD_PERF_BUDGET_TOUR=1) reads a perf_tour.mjs artifact and gates on frameP95 +
+  # the elision-bypass write count against the committed baseline
+  # (hud_perf_budget.baseline.md); it is skipped by default in `npm test`/the
+  # deterministic ci.yml gate because real browser frame timing is hardware-relative
+  # and can be noisy on a shared runner (see the file's own header comment). This job
+  # runs it anyway, informationally: like `screenshots` above it never blocks a merge
+  # (not a required check, this whole workflow is "informational, non-blocking"), but
+  # now a real frameP95/write-count regression gets surfaced on every PR instead of
+  # relying on someone remembering to run the tour by hand.
+  perf-budget-report:
+    name: Real-browser frame budget (informational)
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Generate i18n artifacts
+        run: npm run i18n:gen
+
+      - name: Resolve a Chrome binary
+        run: |
+          CHROME="$(command -v google-chrome || command -v google-chrome-stable || command -v chromium-browser || command -v chromium || true)"
+          if [ -z "$CHROME" ]; then
+            echo "No system Chrome on PATH; installing a Chrome for Testing build."
+            CHROME="$(npx --yes puppeteer browsers install chrome | awk '{print $NF}')"
+          fi
+          echo "BROWSER_PATH=$CHROME" >> "$GITHUB_ENV"
+          echo "Using browser: $CHROME"
+
+      - name: Start the dev client
+        run: |
+          npm run dev -- --host 127.0.0.1 --port 5173 --strictPort &
+          echo $! > .devpid
+          for i in $(seq 1 60); do
+            curl -sf http://127.0.0.1:5173 >/dev/null && { echo "dev client up"; break; }
+            sleep 2
+          done
+
+      # Both viewports land in one artifact (perf_tour.mjs's default PERF_VIEWPORT=both);
+      # continue-on-error because the tour script's OWN hard-error heuristics (stray
+      # console errors, its built-in fctBurst checks) are a different, noisier signal
+      # than the frame budget below, which is the one this job actually gates on.
+      - name: Run the perf tour (desktop + mobile)
+        continue-on-error: true
+        env:
+          GAME_URL: http://127.0.0.1:5173
+          PERF_OUT: tmp/perf-tour.json
+        run: npm run perf:tour
+
+      - name: Stop the dev client
+        if: always()
+        run: kill "$(cat .devpid)" 2>/dev/null || true
+
+      - name: Check the desktop frame budget
+        env:
+          HUD_PERF_BUDGET_TOUR: '1'
+          HUD_PERF_BUDGET_TOUR_VIEWPORT: desktop
+          HUD_PERF_BUDGET_TOUR_RESULT: tmp/perf-tour.json
+        run: npx vitest run tests/hud_perf_budget.test.ts
+
+      - name: Check the mobile frame budget
+        env:
+          HUD_PERF_BUDGET_TOUR: '1'
+          HUD_PERF_BUDGET_TOUR_VIEWPORT: mobile
+          HUD_PERF_BUDGET_TOUR_RESULT: tmp/perf-tour.json
+        run: npx vitest run tests/hud_perf_budget.test.ts
+
   codex-review:
     name: AI review (Codex)
     runs-on: ubuntu-latest
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -199,6 +199,18 @@ jobs:
           fi
           echo "tmp/perf-tour.json present ($(wc -c < tmp/perf-tour.json) bytes)."
 
+      # Persist the raw tour JSON so a CI-only frame-budget failure can be diagnosed from
+      # the recorded metrics instead of reproducing the whole browser tour locally. Uploaded
+      # even on failure (if: always()) since a failing run is exactly when it is needed; the
+      # tour dies with the runner otherwise.
+      - name: Upload the perf tour JSON
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: perf-tour-json
+          path: tmp/perf-tour.json
+          if-no-files-found: warn
+
       - name: Check the desktop frame budget
         env:
           HUD_PERF_BUDGET_TOUR: '1'
@@ -206,7 +218,10 @@ jobs:
           HUD_PERF_BUDGET_TOUR_RESULT: tmp/perf-tour.json
         run: npx vitest run tests/hud_perf_budget.test.ts
 
+      # if: always() so a desktop failure never MASKS a mobile regression: without it the
+      # mobile arm is skipped whenever desktop fails, surfacing mobile a CI cycle later.
       - name: Check the mobile frame budget
+        if: always()
         env:
           HUD_PERF_BUDGET_TOUR: '1'
           HUD_PERF_BUDGET_TOUR_VIEWPORT: mobile
--- a/.github/workflows/pr-ai.yml
+++ b/.github/workflows/pr-ai.yml
@@ -172,6 +172,7 @@ jobs:
       # console errors, its built-in fctBurst checks) are a different, noisier signal
       # than the frame budget below, which is the one this job actually gates on.
       - name: Run the perf tour (desktop + mobile)
+        id: perf_tour
         continue-on-error: true
         env:
           GAME_URL: http://127.0.0.1:5173
@@ -181,6 +182,22 @@ jobs:
       - name: Stop the dev client
         if: always()
         run: kill "$(cat .devpid)" 2>/dev/null || true
+
+      # The tour step above is continue-on-error (its own hard-error heuristics are a
+      # noisier signal than the frame budget this job gates on), but that also means a
+      # tour crash (missing Chrome, a boot timeout, a changed offline-mode selector)
+      # would otherwise surface only as an opaque "cannot read tmp/perf-tour.json"
+      # stack trace out of vitest's loadArtifact. Fail here, visibly, with the tour's
+      # own outcome before that happens, so a real tour breakage is never mistaken for
+      # a frame-budget regression.
+      - name: Verify the perf tour produced an artifact
+        if: always()
+        run: |
+          if [ ! -s tmp/perf-tour.json ]; then
+            echo "::error::perf:tour did not produce tmp/perf-tour.json (tour step outcome: ${{ steps.perf_tour.outcome }}). See the 'Run the perf tour' step log above for the real failure (Chrome resolution, dev-client boot, or offline-world boot)."
+            exit 1
+          fi
+          echo "tmp/perf-tour.json present ($(wc -c < tmp/perf-tour.json) bytes)."
 
       - name: Check the desktop frame budget
         env:
--- a/scripts/gate.mjs
+++ b/scripts/gate.mjs
@@ -52,7 +52,9 @@ const releaseTier = branch.startsWith('release/');
 // gate instead of only via a human remembering the env var.
 const env = {
   ...process.env,
-  PERF_GATE_WALLCLOCK: '1',
+  // ?? rather than a flat override: an operator who explicitly exports
+  // PERF_GATE_WALLCLOCK=0 to skip the slow real-hrtime arm keeps that opt-out.
+  PERF_GATE_WALLCLOCK: process.env.PERF_GATE_WALLCLOCK ?? '1',
   ...(releaseTier ? { I18N_RELEASE_TIER: '1' } : {}),
 };
 
--- a/scripts/gate.mjs
+++ b/scripts/gate.mjs
@@ -45,7 +45,16 @@ if (missingAudioTools.length > 0) {
 const branch =
   spawnSync('git', ['branch', '--show-current'], { encoding: 'utf8', shell }).stdout?.trim() ?? '';
 const releaseTier = branch.startsWith('release/');
-const env = releaseTier ? { ...process.env, I18N_RELEASE_TIER: '1' } : process.env;
+// PERF_GATE_WALLCLOCK=1 turns on tests/server/perf_gate.test.ts's ARM C: a real
+// process.hrtime.bigint() comparison plus a real Sim tick-loop p95 check. It is
+// Node-only (no browser) and carries generous headroom over its ceilings (see
+// the matching comment in .github/workflows/ci.yml), so it runs in every local
+// gate instead of only via a human remembering the env var.
+const env = {
+  ...process.env,
+  PERF_GATE_WALLCLOCK: '1',
+  ...(releaseTier ? { I18N_RELEASE_TIER: '1' } : {}),
+};
 
 const I18N_ARTIFACTS = [
   'src/ui/i18n.resolved.generated',
--- a/src/sim/items.ts
+++ b/src/sim/items.ts
@@ -506,10 +506,16 @@ export function buyItem(ctx: SimContext, npcId: number, itemId: string, pid?: nu
 }
 
 function vendorInRange(ctx: SimContext, p: Entity): boolean {
-  return [...ctx.entities.values()].some(
-    (e) =>
-      e.kind === 'npc' && e.vendorItems.length > 0 && dist2d(p.pos, e.pos) <= INTERACT_RANGE + 2,
-  );
+  // Bounded by the spatial grid instead of scanning every live entity: a vendor
+  // transaction fires far more often than the world roster is small, and a dense
+  // world (hundreds of mobs/npcs/objects) turned this into a per-transaction
+  // O(entities) scan. forEachInRadius only visits the handful of entities in the
+  // cells overlapping the query radius (see src/sim/spatial.ts).
+  let found = false;
+  ctx.grid.forEachInRadius(p.pos.x, p.pos.z, INTERACT_RANGE + 2, (e) => {
+    if (!found && e.kind === 'npc' && e.vendorItems.length > 0) found = true;
+  });
+  return found;
 }
 
 function recordVendorBuyback(meta: PlayerMeta, itemId: string, count: number): void {
--- a/src/sim/social/dungeon_finder.ts
+++ b/src/sim/social/dungeon_finder.ts
@@ -229,14 +229,21 @@ export class DungeonFinderMachine {
   // realm indefinitely.
   private matchMutationSeq = 0;
   // matchMutationSeq value the CURRENT cursor lap started walking against;
-  // reset (with matchLapAnchorsVisited) whenever a mutation invalidates the
-  // in-progress lap.
+  // reset (with matchLapVisitedAnchorIds) whenever a mutation invalidates the
+  // in-progress lap. A proposal forming MID-PASS also bumps matchMutationSeq
+  // (see createProposal), so runMatching() resyncs this the instant a match
+  // forms, not only at pass entry: otherwise the anchors-visited accounting
+  // below would be checked against a lap that started walking a backlog the
+  // pass itself has since mutated, and could wrongly declare it unmatchable.
   private matchLapSeq = -1;
-  // Anchors fully attempted (cursor advanced past them) since matchLapSeq
-  // was last reset. Once this reaches the queue length with no proposal
-  // formed, every anchor has had a turn against the CURRENT backlog: proven
-  // unmatchable until the next mutation.
-  private matchLapAnchorsVisited = 0;
+  // Anchor ids fully attempted (cursor advanced past them) since matchLapSeq
+  // was last reset, tracked as a Set rather than a counter: the `while
+  // (changed)` loop re-sorts the backlog whenever a match forms, and a unit
+  // can otherwise be attempted again after the resort, which would double
+  // count it against a plain increment. Once this reaches the queue length
+  // with no proposal formed, every anchor has had a turn against the CURRENT
+  // backlog: proven unmatchable until the next mutation.
+  private matchLapVisitedAnchorIds = new Set<number>();
   // matchMutationSeq value as of the last time a full lap proved the queue
   // unmatchable; matching goes idle (stops re-arming on truncation) while
   // this equals matchMutationSeq.
@@ -932,10 +939,10 @@ export class DungeonFinderMachine {
     let truncated = false;
     // A mutation invalidates any lap already in progress: restart the
     // anchors-visited count against the CURRENT backlog (see
-    // matchLapAnchorsVisited).
+    // matchLapVisitedAnchorIds).
     if (this.matchLapSeq !== this.matchMutationSeq) {
       this.matchLapSeq = this.matchMutationSeq;
-      this.matchLapAnchorsVisited = 0;
+      this.matchLapVisitedAnchorIds.clear();
     }
     let lastQueueLen = 0;
     while (changed && budgetBox.remaining > 0) {
@@ -946,8 +953,13 @@ export class DungeonFinderMachine {
       // all-dps prefix in front of a formable comp) would otherwise let the
       // oldest anchors alone exhaust the shared budget on every single tick,
       // starving every later anchor including the one that could actually
-      // form a group: rotating guarantees each anchor eventually gets a turn
-      // at bounded per-tick cost, still in FIFO order once traversal wraps.
+      // form a group: rotating guarantees each anchor eventually gets a TURN
+      // AS ANCHOR at bounded per-tick cost, still in FIFO order once
+      // traversal wraps. This only fixes which unit gets to anchor a search;
+      // tryAssemble() still builds its candidate pool from the UNROTATED
+      // FIFO `sorted` order before applying FINDER_MATCH_UNIT_WINDOW, so a
+      // late anchor can still only ever pair with the oldest compatible
+      // units in the window, not ones nearer its own join time.
       const sorted = [...this.queue].sort((a, b) => a.joinedAt - b.joinedAt || a.id - b.id);
       lastQueueLen = sorted.length;
       const cursorIndex =
@@ -973,14 +985,29 @@ export class DungeonFinderMachine {
           if (!match) continue;
           this.createProposal(activity, match.units, match.roles);
           changed = true;
+          // createProposal() just bumped matchMutationSeq: the lap we were
+          // walking started against a backlog this pass itself has now
+          // mutated. Resync immediately so the anchors-visited accounting
+          // below reflects the CURRENT (post-proposal) backlog rather than
+          // the stale one, whether or not this same iteration also runs the
+          // budget out (see matchLapSeq).
+          this.matchLapSeq = this.matchMutationSeq;
+          this.matchLapVisitedAnchorIds.clear();
           break;
         }
         if (changed || truncated) break;
         // Fully attempted this anchor within budget: resume after it next
         // pass so a later pass advances past it instead of retrying it first.
         this.matchCursorUnitId = anchor.id;
-        this.matchLapAnchorsVisited++;
+        this.matchLapVisitedAnchorIds.add(anchor.id);
       }
+      // A proposal forming on the very node that exhausted the shared budget
+      // must still count as truncation: the `for` loop exits via `changed`
+      // (not the budget checks above `truncated`), and the `while` guard
+      // (`budgetBox.remaining > 0`) would otherwise let this fall through to
+      // the "full uninterrupted pass" branch below even though the backlog
+      // was never actually walked to completion.
+      if (budgetBox.remaining <= 0) truncated = true;
       if (truncated) break;
     }
     if (truncated) {
@@ -989,7 +1016,7 @@ export class DungeonFinderMachine {
       // proven unmatchable, so stop re-arming matchDirty every tick until
       // the next queue mutation (see armMatching). This is what keeps an
       // unformable backlog from re-running a full matching pass forever.
-      if (this.matchLapAnchorsVisited >= lastQueueLen) {
+      if (this.matchLapVisitedAnchorIds.size >= lastQueueLen) {
         this.matchProvenIdleSeq = this.matchMutationSeq;
       }
       if (this.matchProvenIdleSeq !== this.matchMutationSeq) {
@@ -1001,7 +1028,7 @@ export class DungeonFinderMachine {
       // A full uninterrupted pass completed with nothing left to form:
       // proven idle for the current backlog too.
       this.matchCursorUnitId = null;
-      this.matchLapAnchorsVisited = 0;
+      this.matchLapVisitedAnchorIds.clear();
       this.matchLapSeq = this.matchMutationSeq;
       this.matchProvenIdleSeq = this.matchMutationSeq;
     }
--- a/src/sim/social/dungeon_finder.ts
+++ b/src/sim/social/dungeon_finder.ts
@@ -44,7 +44,12 @@ export const FINDER_DECLINE_COOLDOWN_SECONDS = 60;
 // while keeping strict join-order fairness inside the window.
 export const FINDER_MATCH_UNIT_WINDOW = 24;
 // Backstop on the combination search so a pathological queue can never stall
-// a tick (the single-role pruning makes real queues far cheaper).
+// a tick (the single-role pruning makes real queues far cheaper). This is the
+// TOTAL node budget for one runMatching() pass, shared across every anchor and
+// activity attempted in that pass (see the shared budgetBox in runMatching):
+// a per-attempt budget would let a role-imbalanced backlog re-spend the full
+// budget on every anchor, scaling total per-tick cost with the anchor count
+// instead of bounding it.
 export const FINDER_MATCH_NODE_BUDGET = 4000;
 // Board projection cap: the wire payload stays bounded however many listings
 // a realm accumulates (oldest listings first, stable order).
@@ -879,17 +884,26 @@ export class DungeonFinderMachine {
   // ---- deterministic matching ---------------------------------------------------
 
   private runMatching(): void {
+    // One shared node budget for the WHOLE matching pass (every anchor and
+    // activity attempted below), not reset per attempt: a role-imbalanced
+    // backlog (e.g. an all-dps queue that can never seat a tank/healer
+    // composition) would otherwise let every anchor independently exhaust its
+    // own fresh budget, scaling total per-tick cost with the anchor count
+    // instead of bounding it (see FINDER_MATCH_NODE_BUDGET).
+    const budgetBox = { remaining: FINDER_MATCH_NODE_BUDGET };
     // Stable FIFO order: original join time, then unit id.
     let changed = true;
-    while (changed) {
+    while (changed && budgetBox.remaining > 0) {
       changed = false;
       const sorted = [...this.queue].sort((a, b) => a.joinedAt - b.joinedAt || a.id - b.id);
       for (const anchor of sorted) {
+        if (budgetBox.remaining <= 0) break;
         const anchorActivities = this.eligibleActivitiesFor(anchor);
         for (const activityId of anchorActivities) {
+          if (budgetBox.remaining <= 0) break;
           const activity = finderActivity(activityId);
           if (!activity || activity.composition === null) continue;
-          const match = this.tryAssemble(sorted, anchor, activity);
+          const match = this.tryAssemble(sorted, anchor, activity, budgetBox);
           if (!match) continue;
           this.createProposal(activity, match.units, match.roles);
           changed = true;
@@ -911,9 +925,11 @@ export class DungeonFinderMachine {
     sorted: FinderQueueUnit[],
     anchor: FinderQueueUnit,
     activity: FinderActivity,
+    budgetBox: { remaining: number },
   ): { units: FinderQueueUnit[]; roles: Map<number, Role> } | null {
     const comp = activity.composition;
     if (comp === null) return null;
+    if (budgetBox.remaining <= 0) return null;
     const target = activity.size;
     // Anchor first (mandatory), then every other compatible unit in FIFO
     // order. An older unit may appear here even though it failed as its own
@@ -929,12 +945,11 @@ export class DungeonFinderMachine {
     const roleMembers = new Map<number, FinderRoleMember[]>();
     for (const u of candidates) roleMembers.set(u.id, this.unitRoleMembers(u));
 
-    let budget = FINDER_MATCH_NODE_BUDGET;
     const chosen: FinderQueueUnit[] = [];
     let solution: { units: FinderQueueUnit[]; roles: Map<number, Role> } | null = null;
 
     const dfs = (idx: number, size: number): boolean => {
-      if (budget-- <= 0) return true; // out of budget: stop the whole search
+      if (budgetBox.remaining-- <= 0) return true; // out of budget: stop the whole search
       if (size === target) {
         const members = chosen.flatMap((u) => roleMembers.get(u.id) ?? []);
         const roles = assignFinderRoles(members, comp);
--- a/src/sim/social/dungeon_finder.ts
+++ b/src/sim/social/dungeon_finder.ts
@@ -973,29 +973,63 @@ export class DungeonFinderMachine {
           truncated = true;
           break;
         }
+        // Did THIS anchor's search begin with the whole fresh budget? A
+        // truncation that STARTS full is bounded BY DESIGN (the anchor was
+        // handed every node and still could not finish): it counts as a real
+        // attempt, so we mark it and advance the cursor past it. A truncation
+        // that starts on the DREGS of an already-spent budget is NOT a real
+        // attempt: it must not advance the cursor, so a later pass retries
+        // this anchor as a pass-starter with a full budget. Conflating the
+        // two is what opened both lap-accounting holes: a dregs last-activity
+        // truncation being mis-marked as fully visited (latching idle over a
+        // never-fully-searched queue, with no wake-up), and a fresh
+        // first-activity truncation never being marked (re-arming a full
+        // matching pass every tick forever, starving every later anchor).
+        const anchorStartedFresh = budgetBox.remaining === FINDER_MATCH_NODE_BUDGET;
         const anchorActivities = this.eligibleActivitiesFor(anchor);
+        let anchorTruncated = false;
         for (const activityId of anchorActivities) {
           if (budgetBox.remaining <= 0) {
-            truncated = true;
+            anchorTruncated = true;
             break;
           }
           const activity = finderActivity(activityId);
           if (!activity || activity.composition === null) continue;
           const match = this.tryAssemble(sorted, anchor, activity, budgetBox);
-          if (!match) continue;
-          this.createProposal(activity, match.units, match.roles);
-          changed = true;
-          // createProposal() just bumped matchMutationSeq: the lap we were
-          // walking started against a backlog this pass itself has now
-          // mutated. Resync immediately so the anchors-visited accounting
-          // below reflects the CURRENT (post-proposal) backlog rather than
-          // the stale one, whether or not this same iteration also runs the
-          // budget out (see matchLapSeq).
-          this.matchLapSeq = this.matchMutationSeq;
-          this.matchLapVisitedAnchorIds.clear();
+          if (match) {
+            this.createProposal(activity, match.units, match.roles);
+            changed = true;
+            // createProposal() just bumped matchMutationSeq: the lap we were
+            // walking started against a backlog this pass itself has now
+            // mutated. Resync immediately so the anchors-visited accounting
+            // below reflects the CURRENT (post-proposal) backlog rather than
+            // the stale one, whether or not this same iteration also runs the
+            // budget out (see matchLapSeq).
+            this.matchLapSeq = this.matchMutationSeq;
+            this.matchLapVisitedAnchorIds.clear();
+            break;
+          }
+          // tryAssemble() may have run the shared budget to zero mid-DFS: that
+          // is a truncation of THIS anchor's search, distinct from "searched
+          // fully within budget, no group possible".
+          if (budgetBox.remaining <= 0) {
+            anchorTruncated = true;
+            break;
+          }
+        }
+        if (changed) break;
+        if (anchorTruncated) {
+          truncated = true;
+          if (anchorStartedFresh) {
+            // Bounded by design: count it and advance so the next pass moves
+            // on to a later anchor instead of re-searching this one first.
+            this.matchCursorUnitId = anchor.id;
+            this.matchLapVisitedAnchorIds.add(anchor.id);
+          }
+          // Dregs truncation: leave the cursor and the visited set untouched
+          // so a later pass retries this anchor with a full budget.
           break;
         }
-        if (changed || truncated) break;
         // Fully attempted this anchor within budget: resume after it next
         // pass so a later pass advances past it instead of retrying it first.
         this.matchCursorUnitId = anchor.id;
--- a/src/sim/social/dungeon_finder.ts
+++ b/src/sim/social/dungeon_finder.ts
@@ -220,6 +220,27 @@ export class DungeonFinderMachine {
   // instead of re-spending the whole budget on the same oldest anchors every
   // tick (see runMatching).
   private matchCursorUnitId: number | null = null;
+  // Livelock guard: bumped by every queue-affecting mutation (join, leave,
+  // role change, proposal fail/complete, sweep dropping/re-scoping a unit).
+  // runMatching() uses it to tell "the cursor is still working through the
+  // SAME backlog" apart from "the backlog actually changed", so a proven-
+  // unmatchable backlog stops re-arming matchDirty every tick forever
+  // instead of re-running the (bounded but non-zero) full pass on an idle
+  // realm indefinitely.
+  private matchMutationSeq = 0;
+  // matchMutationSeq value the CURRENT cursor lap started walking against;
+  // reset (with matchLapAnchorsVisited) whenever a mutation invalidates the
+  // in-progress lap.
+  private matchLapSeq = -1;
+  // Anchors fully attempted (cursor advanced past them) since matchLapSeq
+  // was last reset. Once this reaches the queue length with no proposal
+  // formed, every anchor has had a turn against the CURRENT backlog: proven
+  // unmatchable until the next mutation.
+  private matchLapAnchorsVisited = 0;
+  // matchMutationSeq value as of the last time a full lap proved the queue
+  // unmatchable; matching goes idle (stops re-arming on truncation) while
+  // this equals matchMutationSeq.
+  private matchProvenIdleSeq = -1;
   // Board projection cache (viewer-independent): rebuilt when the revision
   // bumps or on the once-a-second sweep (party rosters mutate outside finder
   // commands), then delta-elided on the wire by the encoder.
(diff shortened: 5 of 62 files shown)


FernandoX7 (commented):
This is a well-built test expansion overall. The recipe is disciplined (median of many samples, floored ratio denominators, a shape-sanity case in nearly every file, real entry points in the large majority), the two production regressions are real finds, and the honest skip disclosures are appreciated. The vendorInRange rewrite itself checks out as semantically exact: forEachInRadius does a true d2 <= r2 check matching the old dist2d bound, every spawn path routes through addEntityToRoster into ctx.grid, and the existing sim.test.ts coverage pins both the in-range and no-merchant-nearby arms through the new path. Merge against release/v0.28.0 is conflict-free.

Two blocking findings, then a cluster of should-fixes that are mostly about assertions that cannot fail on the regression they name.

**Blocking**

1. `src/sim/social/dungeon_finder.ts:884` (shared budgetBox in runMatching): the shared budget introduces a matchmaking livelock in a realistic queue shape. Concrete scenario: a healer queues first, 20+ dps queue behind it, then a tank joins. Old code: every anchor got a fresh 4000-node budget, so after the early anchors failed, the tank as its own anchor found {tank, healer, 3 dps} in a handful of nodes (its candidates are [tank, healer, dps...], and that comp is the first DFS path). New code: the oldest anchor's DFS alone exhausts the shared 4000 nodes proving no tank exists among its 23 FIFO-window candidates (C(23,4) is about 8855 leaves), every later anchor including the tank is skipped, and nothing re-arms matchDirty on truncation, so each future pass (triggered by the next join or leave) repeats the identical truncated search. The formable group deterministically never forms while the backlog persists. The perf test only covers the all-dps unmatchable queue, and the existing dungeon_finder behavior suites use small rosters that never exhaust the budget, so no test catches this. Suggested direction: keep the shared pass budget, but on budget-exhausted truncation re-arm matchDirty and resume the next pass from a persisted anchor cursor, so successive passes cover all anchors at a bounded per-tick cost. Either way, add the positive-match case: a matchable comp behind a large unmatchable prefix must still produce a proposal.

2. `tests/server/perf_gate.test.ts:428` via `.github/workflows/ci.yml` and `scripts/gate.mjs`: force-enabling ARM C broke this PR's own required check. PR gate shard 4 failed with "Test timed out in 5000ms" on the tick-p95 case: it runs 200 warmup plus 4000 measured real Sim ticks with a pipeline request per tick and has no explicit it() timeout (every new perf file in this PR carries a 30 to 60s timeout; this pre-existing file was written for opt-in manual runs and was not adjusted when it was wired into the required lane). The same forcing is in the release-gate job and npm run gate. Minimum fix: explicit generous timeouts on both ARM C cases, then decide whether the required lane is the right home versus the informational one.

**Should-fix**

3. `.github/workflows/pr-ai.yml` perf-budget-report: the job is broken on arrival. On this PR the tour died with "Timed out waiting for offline world boot", continue-on-error masked it, and both frame-budget steps then failed on the missing tmp/perf-tour.json. A permanently red informational job trains everyone to ignore it, which defeats the point of wiring it up. Get one green run on CI hardware before landing the job (the screenshots job's boot recipe is the working reference), and make the tour step fail visibly when the artifact was not produced.

4. `tests/delve_interactable_visibility_perf.test.ts:78`: the one pure-ratio bound in the PR (largeMedian / floor <= 7) at microsecond magnitudes, with a 0.001ms denominator floor and no absolute escape. Timer quantization on a contended shard can red it. Use the sibling files' Math.max(small * k, absolute) form. The same floorless-ratio form, at lower risk, is in `pick_resolution_perf.test.ts:118,136` and `static_matrix_perf.test.ts:63`.

5. A cluster of files whose perf assertions cannot fail on the regression they name. The absolute ceilings still catch order-of-magnitude blowups, but the scaling arms are the advertised differentiator and in these files they are inert:
   - `tests/vendor_transaction_perf.test.ts`: reverting vendorInRange to the old full scan trips neither perf assertion. The linear scan at +1600 entities is sub-ms, the round trip is constant-dominated so the ratio stays near 2x, and the 5ms floor auto-passes. The correctness case is the real guard here. A decisive form needs tens of thousands of far entities and a near-constant-cost assertion.
   - `tests/spatial_grid_perf.test.ts:74,95` and `tests/collider_resolution_perf.test.ts:81,102`: the whole population is packed inside the query radius (or one grid cell), so the healthy path already visits every candidate and a lost-bucketing regression does essentially identical work. Needs a small query against a large, mostly out-of-range field.
   - `tests/threat_table_perf.test.ts:79,99`: the scaled dimension (hate-table depth) is disconnected from the measured O(1) calls; a table scan at depth 1000 is a few microseconds, far below both the 2ms ceiling and the 0.5ms floor.
   - `tests/gather_nodes_lookup_perf.test.ts`: measures a test-local re-implementation over the raw exported Records; the shipped consumer (`src/render/gather_nodes.ts`) is Three-bound and untouched by this file, so it effectively benchmarks V8 property access. Extract the per-node resolution into a pure helper the consumer actually calls, or move this into the PR's honest-skips list.
   - `tests/dungeon_instance_perf.test.ts`: the median over 200 ticks is blind to the once-per-second occupancy scan it claims to guard (about 10 of 200 samples include it). Measure the tickCount % 20 tick, or gate a high percentile.
   - `tests/character_effects_perf.test.ts`: the one file with no shape-sanity assertion at all (only the three timing expects at lines 161, 181, 201); a collapsed auraSoup or empty population passes silently.

**Nits**

6. `tests/proc_dispatch_perf.test.ts`: the measured load erodes mid-run. The file's own comment (lines 113 to 118) concedes mainhandItemId reverts to the default weapon after the first landed proc's stat recalc, so later waves are lighter than the advertised fully geared raid. Pin the loadout per wave, or equip via meta.equipment.
7. Vacuous-if-empty count pins: `damage_resolution_perf`'s dead === mobs.length and `hit_table_perf`'s counts === warriors.length pass as 0 === 0 if the build collapses. Pin literals the way the talent, deeds, and honor files do.
8. `tests/vfx_pool_alloc_perf.test.ts:104`: the allocation-stability probe does not burst inside the probed thunk, so buffer reuse under burst rides only on the post-loop length check. The capacity literal pins are the decisive part of the file.
9. Cost: all of these run in every default npm test, adding very roughly 30 to 60 seconds across the suite (nythraxis, terrain_sampling, mob_lifecycle, and the pvp minigame file are the heaviest). Shard times on this PR are still acceptable, but this family is the natural candidate for its own vitest project or an env gate if it grows, and populations could be cut substantially with no coverage loss since the ratio, not the population, carries the pin.

One calibration note, not an ask: with ceilings 10 to 25x over healthy medians and 3.5x scaling bounds, this suite catches order-of-magnitude and quadratic regressions, not 2 to 5x constant-factor creep. That is a defensible flake-safety tradeoff; just do not let the 160-test figure be read as tighter coverage than it is.

jgyy:
Thanks for the thorough pass. Pushed 214c4bba6 addressing both blocking findings plus two should-fix items:

**Blocking**

1. Dungeon finder livelock: fixed by re-arming `matchDirty` and resuming the next `runMatching()` pass from a persisted anchor cursor when the shared budget truncates mid-pass, so successive ticks eventually give every anchor a turn at bounded per-tick cost, instead of the same oldest anchors re-exhausting the budget forever. Added the positive-match regression you described (`tests/dungeon_finder.test.ts`): a healer, then 23 unmatchable dps, then a tank, and asserted the {tank, healer, 3 dps} group still forms within a bounded number of ticks.
2. `perf_gate.test.ts` ARM C timeouts: added the explicit `60_000` timeout to both cases, matching every other new perf file in this PR.

**Should-fix (partial)**

4. Tightened the floorless-ratio bound in `delve_interactable_visibility_perf`, `pick_resolution_perf` (both cases), and `static_matrix_perf` to the sibling files' `Math.max(small * k, absolute)` form.
7. Pinned the vacuous-if-empty count checks in `damage_resolution_perf` and `hit_table_perf` to the literal `PAIRS` constant instead of the array length they're supposed to be guarding.

**Deferred, not done in this pass**

Items 3, 5 (the six inert-scaling-arm files), 6, 8, and 9 are real and I agree with the direction on each, but they're each their own non-trivial rewrite (new population shapes, extracting a pure helper for `gather_nodes_lookup_perf`, getting `pr-ai.yml`'s perf-budget-report green on real CI hardware) rather than a mechanical fix, and I'd rather land the two blockers plus the quick wins than rush the rest. Happy to follow up in a separate PR for those, or take a swing at them here if you'd prefer everything in one change, just say the word.

All local checks green (`tsc --noEmit`, targeted vitest runs including the new regression test, biome on changed files); pushed to `feature/perf-test-expansion` and CI is running now.

EnriqueGF (changes requested):
Reviewed this end to end: read the full diff, ran the checks myself, and independently reproduced the two behavioral claims. The two production fixes are real and the intent here is good, but I found three things I think should be addressed before merge. Details and measurements below so you can check my work.

## What I verified as correct

**`vendorInRange` (`src/sim/items.ts`) is a clean, behavior-preserving optimization.** I checked the three ways this could have silently become a range exploit rather than a speedup:
- `SpatialGrid.forEachInRadius` applies an exact `d2 <= r2` filter (`src/sim/spatial.ts:74-77`), not a cell-only test, so the effective interaction range is unchanged.
- `dist2d` (`src/sim/types.ts:3683`) is x/z-only, matching the grid's plane exactly.
- Every spawn path routes through `addEntityToRoster` -> `ctx.grid.insert` (`src/sim/entity_roster.ts:123`); I grepped `src/sim/` for other `entities.set(` call sites and found none, so the grid roster cannot miss a vendor.

`items`, `vendor_stack`, `heroic_vendor`, `delve_shop`, `ptr_dev_vendor` all pass (51 tests). `npx tsc --noEmit` clean; `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` pass. No em/en dashes or emoji in added lines.

## 1. The scaling checks cannot fail (~35 of the 40 new files) - the main issue

The PR's stated rationale for the scaling check is that it "is the piece a flat ceiling alone cannot provide - it catches an O(n) regression that goes quadratic even while both individual measurements stay comfortably under budget." I don't think it does that as written.

The idiom is `expect(largeMedian).toBeLessThan(Math.max(smallMedian * 3.5, 5))`. At sub-ms medians - which is most sim leaf functions in Node - the `5` floor dominates and the ratio bound never binds. Measured, by running the files:

| file | smallMedian | `Math.max(small*3.5, 5)` | effective assertion | actual large | headroom |
|---|---|---|---|---|---|
| `threat_table_perf` | 0.0009ms | max(0.003, **5**) | `< 5ms` | 0.0009ms | ~5500x |
| `xp_grant_perf` | 0.0010ms | max(0.004, **5**) | `< 5ms` | 0.0010ms | ~5000x |
| `spatial_grid_perf` | 0.089ms | max(0.31, **5**) | `< 5ms` | 0.188ms | ~26x |

For `threat_table_perf`, a genuine quadratic blowup at doubled depth lands around 0.0036ms - still roughly 1400x under the 5ms floor. The assertion the file exists to make is inert, and the check silently degenerates into exactly the flat ceiling it was written to improve on.

`grep -l "Math.max(smallMedian" tests/*_perf.test.ts` matches 35 files, so this is most of a 7,400-line PR. Suggested fix: assert the ratio directly and gate on a minimum measurable magnitude, e.g. skip (or raise the population until) `smallMedian` clears some floor like 0.5ms, then `expect(largeMedian / smallMedian).toBeLessThan(3.5)`. As it stands these files cost CI time without being able to fail.

Related: the sub-ms medians (0.0009ms) are close to timer resolution, which suggests some of these populations aren't large enough for the measurement to mean much regardless of the assertion.

## 2. The finder fix trades a one-off spike for a permanent per-tick cost floor

`runMatching` sets `matchDirty = true` on truncation (`dungeon_finder.ts:948`) and `update()` re-runs matching whenever it's set (`:813`). With a queue that can never form a group, every pass truncates, so the flag re-arms every tick with no terminating condition.

Instrumented both versions, 30 dps-only queued, 200 ticks:

| | matching passes / 200 ticks | wall time |
|---|---|---|
| base (`c011f818d`) | **1**, then idle | 1085ms |
| this PR | **200** (every tick) | 2456ms |

That delta is about 6.9ms per tick, sustained indefinitely, roughly 14% of the 50ms tick budget - on a shape (all-dps queue at off-peak) that a dungeon finder hits routinely. Per-tick this is still better than base's 415ms spike, so the trade is defensible, but the PR describes it purely as "~415ms -> ~6ms" without noting the 6ms is now *every tick, forever* rather than *once*. Worth stating explicitly, and worth considering a cheap stop condition: once the cursor has wrapped a full cycle with no proposal formed and no queue mutation, stop re-arming until the next mutation.

`tests/dungeon_finder_perf.test.ts:59` force-sets `finder.matchDirty = true` before each sample, which structurally prevents that test from ever observing this - it can't distinguish "went idle" from "re-arms forever." A test that ticks normally and counts passes would cover it.

## 3. The new `perf-budget-report` job will likely be permanently red

`pr-ai.yml:187-193` runs the frameP95 check without setting `HUD_PERF_BUDGET_TOUR_FRAME_BASELINE`, so `readBaselineFrameP95()` (`hud_perf_budget.test.ts:143-155`) falls back to the committed `250 ms` from `hud_perf_budget.baseline.md:72`. That value was captured on an Apple M4 Max / macOS under swiftshader, and the baseline document says in bold that these numbers "do not travel across hardware ... Compare them only against a fresh same-machine re-run of this baseline, never against the literal milliseconds on different hardware." A 4-vCPU `ubuntu-latest` under software WebGL will very likely exceed it.

The `hudHotDomWrites <= 153` bypass-count check IS hardware-portable (and is the actual elision-collapse signal); I'd keep that one and drop the frameP95 assertion, or pin it via the env override to a runner-measured value.

Two smaller notes on the same job:
- `perf_tour.mjs` writes its JSON only after both viewports finish, and `runViewport` throws on boot timeout - so one flaky mobile boot means no artifact at all and *both* vitest steps fail with an ENOENT stack, including the desktop one that succeeded. `continue-on-error` on the tour step hides the cause.
- There's no `upload-artifact` or PR comment step, so a job named "report" produces no report - just a check mark, with the numbers discarded in log output. Uploading `tmp/perf-tour.json` would make it useful.

(I checked and these are non-issues: the job correctly needs only `npm run dev`, not the game server, since `perf_tour.mjs:104` boots the offline Sim; and a missing artifact fails rather than passing vacuously, since `loadArtifact` does an unguarded `readFileSync`. The `npm run dev &` / `kill` handling is copied verbatim from the existing `screenshots` job, so not a new defect.)

## 4. Minor

- `ci.yml` sets `PERF_GATE_WALLCLOCK=1` on the sharded test jobs, which run vitest at default unbounded parallelism, so ARM C measures `hrtime` deltas while sibling workers saturate the same cores. `scripts/gate.mjs:77` bounds workers; CI doesn't, and root `CLAUDE.md` specifically warns that an unbounded run flakes heavy suites under contention. Headroom is large (p99 needs 40/4000 samples to blow out), so I'd call this real but manageable - bounding workers on the CI test step is a better fix than dropping the env var.
- `scripts/gate.mjs` now force-sets `PERF_GATE_WALLCLOCK: '1'`, overriding an operator who exports `0` to skip the slow arm. `process.env.PERF_GATE_WALLCLOCK ?? '1'` would respect the opt-out. The release-tier arm is otherwise unchanged - verified the spread order still lets `I18N_RELEASE_TIER` win.
- `tests/server/perf_gate.test.ts:50` and `:364` still say ARM C is "SKIPPED by default", and `docs/api-pipeline/state.md:1104` repeats it. After this PR it runs everywhere except a bare `npx vitest run`; per the anchor rule those should be updated here.

## One thing I checked and am explicitly NOT reporting

`tests/dungeon_finder.test.ts` ("a decline returns accepted units...") timed out for me on this branch and my first read was that this PR caused it. It doesn't. Running base and PR three times each under matched conditions: base 2530 / 5165 / 6780ms (two failures), PR 3057 / 2644 / 3020ms. That test is pre-existing flaky against the 5s default timeout and my initial failure was contention from my own parallel runs. Flagging only because this PR adds 40 co-running perf files, which makes that class of timeout more likely across the suite.

Happy to re-review once 1 and 3 are addressed. The two production fixes themselves both look right to me.

EnriqueGF:
Follow-up after a deeper sweep across all 40 files. Three additions, all re-verified on my host (6 cores) before posting.

## A. `tests/terrain_sampling_perf.test.ts:73` - the budget has ~1.5x headroom, not the 8-10x the PR describes

The file's own comment says: "observed healthy median for a 4096-point, 5-call-per-point sampling pass is **a low single-digit ms figure**; 60ms leaves ample headroom."

Measured on an idle host, five consecutive runs:

```
median=39.77ms  37.76ms  36.78ms  40.02ms  39.78ms
```

That is ~38ms, roughly **10x higher than the comment claims**, against a 60ms bound - so real headroom is about 1.5x, not the "~8-10x the observed healthy median" the PR description states as the calibration rule. The methodology looks like it wasn't actually applied to this file.

It matters concretely: during an earlier run with incidental background load I captured `median=63.59ms`, i.e. **over the 60ms bound**. Combined with finding 4 in my first review (CI runs vitest at unbounded parallelism, so this competes with ~3 sibling workers on a 4-vCPU runner), I'd expect this to flake in CI, and a separate sweep on this same host measured it as high as 78.7ms. Recommend either raising the budget to match the genuine ~38ms baseline with real headroom, or shrinking the 4096-point grid.

Worth re-checking the other absolute ceilings against this same question: the PR states a uniform ~8-10x rule, and at least this one is ~1.5x.

## B. 10 files dropped the lap guard, so a phase rename makes them pass vacuously

`tests/mob_update_perf.test.ts:120` (pre-existing, the file the new ones are modeled on) carries `expect(mobLapTotal).toBeGreaterThan(0)`. The 10 new `perfLap` files copied the recipe and dropped it:

```
aura_tick, dungeon_instance, delve_runs, dungeon_finder, ground_aoe_projectile,
mob_social_aggro, party_ops, pet_ai, pvp_minigame_tick, world_boss
```
(confirmed: `perfLap` present in each, zero `toBeGreaterThan(0)` in any)

These accumulate only ticks whose phase string matches a literal (`'p.auras'`, `'mob.auras'`, ...). Because the hook is installed through an `as unknown as` cast, renaming a phase in `sim.ts` keeps compiling, the accumulator stays 0, and every assertion in the file passes vacuously - `0 < 25`, and `0 < Math.max(0, 5)`. All the phase literals match real `lap?.()` call sites today, so nothing is broken right now; this is a silent-break risk, not a live defect. One line per file fixes it.

## C. Quantifying finding 1: 20 of 25 scaling checks measured are floor-dominated

Sharpening my first review with the full sweep. The ratio term only binds when `smallMedian > FLOOR / 3.5` (>1.43ms at `FLOOR=5`, >0.57ms at `2`, >0.29ms at `1`). Floor distribution across the files: `5` x30, `2` x10, `1` x8, `0.5` x3. Of 25 scaling checks actually executed, **20 are floor-dominated; 5 genuinely bind** (`pathfind`, `market_query`, `damage_resolution`, `party_ops`, `terrain_sampling`).

The blowup factor required to fail:

| file | small / large median | bound in effect | regression needed to fail |
|---|---|---|---|
| `chat_broadcast_perf` | 0.002 / 0.001ms | `< 5ms` | ~5000x |
| `xp_grant_perf` | 0.0011 / 0.0008ms | `< 2ms` | ~2500x |
| `threat_table_perf` | 0.0006 / 0.0006ms | `< 0.5ms` | ~830x |
| `honor_rating_perf` | 0.0019 / 0.0013ms | `< 1ms` | ~770x |
| `hit_table_perf` | 0.05 / 0.10ms | `< 5ms` | ~50x |
| `aura_tick_perf` | 0.14 / 0.15ms | `< 5ms` | ~33x |

An O(n) -> O(n^2) regression at a 2x population step produces exactly 2x growth over healthy - two to three orders of magnitude short of tripping any of these.

There's a sharper tell that these measurements carry no signal at all: in ~10 of them the **larger** population measured *faster* than the smaller one (`threat_table` 0.98x, `spatial_grid` 0.95x, `party_ops` 0.95x, `xp_grant` 0.73x, `honor_rating` 0.68x, `chat_broadcast` 0.61x, `mail` 0.28x). A scaling ratio below 1.0 is proof the quantity being measured is dominated by timer resolution and JIT warmup rather than by the workload. The populations need to be sized so the small case lands in the multi-ms range (as `pathfind` and `market_query` already do), or the measure needs to be iteration-count-based rather than a wall-clock median.

## Also worth a look

- ~20 files have no shape-sanity/non-vacuity check at all (`arena_matchmaking`, `aura_tick`, `bank_bags`, `character_effects`, `damage_resolution`, `dungeon_finder`, `loot_roll`, `mail`, `market_query`, `net_interp`, `party_ops`, `vendor_transaction`, others). `hit_table_perf.test.ts:151-155` is the model to copy - it pins the built count against the **literal** constant rather than against `warriors.length`, which is exactly what defeats the `0 === 0` collapse. Good pattern, just applied unevenly.
- `hit_table_perf.test.ts:157`: `expect(resists).toBeGreaterThanOrEqual(0)` cannot fail, since a count is always non-negative. Cosmetic (the surrounding assertions are decisive), but it should pin a plausible range or go.

## Clean

No `.only`/`.skip`, no assertion-free `it` blocks, and no cross-file state leakage anywhere in the 40 files: zero `globalThis` writes, no `process.env` mutation, no `vi.mock`/`vi.stubGlobal`, no `Date.now`/`Math.random` patching. Every `as unknown as` cast targets a locally-constructed `Sim`. Populations are genuinely built where probed - `chat_broadcast` really does emit one event per subscriber with all 500 passing the `ctx.players.has()` gate, so the "accidental no-op" class of vacuity isn't present.

To be clear about the overall shape: the scaffolding, the module-level targeting, and the two production fixes are all sound work. The issue is calibration - the budgets and scaling bounds need to be sized against measured reality before these files can catch anything.

jgyy:
Thanks for the deep pass. Status update on the two items still open after 214c4bba6 (which already covered the two blocking findings plus the floorless-ratio and vacuous-pin should-fixes):

Should-fix #3 (perf-budget-report job broken on arrival): confirmed on this run too. "Run the perf tour" is intentionally continue-on-error, and when it dies (a boot timeout this time), that silently produced an empty/missing tmp/perf-tour.json, which then surfaced only as an opaque loadArtifact stack trace out of vitest, exactly as you called out. Pushed 69196b804: adds an explicit "Verify the perf tour produced an artifact" step right after the tour + dev-client teardown, so a broken tour now fails the job loudly with the tour step's own outcome instead of masquerading as a frame-budget regression. I did not chase getting one fully green run on CI hardware in this pass (that's an environment-availability question for the informational job, not a code fix), but the failure mode you flagged is now visible instead of hidden.

The remaining should-fix cluster (#5: vendor_transaction_perf / spatial_grid_perf / collider_resolution_perf / threat_table_perf / gather_nodes_lookup_perf / dungeon_instance_perf scaling arms being inert, and character_effects_perf missing a shape-sanity check) and the nits (#6, #8, #9) are left as-is for now: they're real coverage gaps but don't affect correctness of what's shipped, and touching that many perf files' population/timing shapes in the same pass as the livelock fix risked scope creep. Happy to take a follow-up pass at those if you'd rather they land in this PR than as a fast-follow.

Also reran the two CI-cancelled checks (Screenshots of changes, Real-browser frame budget) since they were dropped mid-run, not real failures.

jgyy:
Thanks both for the detailed review. Pushed a follow-up fix for the remaining part of blocking item 1.

**Fixed in this pass:**
- The dungeon finder livelock/forever-rearm (blocking #1, the second half): the earlier fix gave `runMatching()` a persisted anchor cursor so a truncated pass resumes at the next anchor, which does guarantee every anchor eventually gets a turn, but it still unconditionally re-armed `matchDirty` on every truncation. An unmatchable backlog (no tank/healer ever queues) therefore re-ran a full bounded matching pass on every tick forever. Added a mutation-sequence counter plus a lap tracker so `runMatching()` can tell "still walking the same unformable backlog" apart from "the backlog actually changed": once every anchor has had a turn with no proposal formed and no mutation since, matching goes idle until the next join/leave/role-change/proposal outcome. Added two regression tests (idle-after-proven-unmatchable, and re-arm-on-new-join) plus a normal-ticking (no forced `matchDirty`) idle counter in `tests/dungeon_finder_perf.test.ts` per the review note that the existing perf cases there force-arm every sampled tick and can't observe idle behavior.
- Confirmed items already fixed on this branch by an earlier pass: blocking #2 (ARM C 60s timeouts in `tests/server/perf_gate.test.ts`), should-fix #3 (`pr-ai.yml` perf tour now fails loudly instead of silently), #5 (floorless ratio bound in `delve_interactable_visibility_perf`/`pick_resolution_perf`/`static_matrix_perf`), and part of #6 (vacuous-if-empty pins in `damage_resolution_perf`/`hit_table_perf`).

**Still open (follow-up):**
- #4, the ~35-file `Math.max(smallMedian * 3.5, 5)` inert-scaling-assertion mechanical fix, is not done. Confirmed the pattern is uniform via `grep -l "Math.max(smallMedian" tests/*_perf.test.ts` (35 files) but did not want to rush a batch edit across that many perf-sensitive files without per-file verification given the remaining time in this pass.
- #6's other items (vendor_transaction, spatial_grid/collider_resolution population shape, threat_table dimension, gather_nodes_lookup honest-source, dungeon_instance median blindness, character_effects shape-sanity) and the #7 nits are untouched.

Happy to pick those up in a follow-up PR/commit if useful, flagging so they don't get lost.

**Verification:** `npx tsc --noEmit` clean; `npx vitest run tests/dungeon_finder.test.ts tests/dungeon_finder_perf.test.ts tests/dungeon_finder_view.test.ts tests/dungeon_finder_window.test.ts tests/server/perf_gate.test.ts` all green (102 passed, 2 skipped); `tests/architecture.test.ts` and `tests/parity` green.

EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants, and ran the guard suites plus the new perf files locally. The two production fixes are the interesting part of this PR; the `vendorInRange` one is clean, but the dungeon finder rework has a reachable stall that I reproduced, and several of the new budgets fail on contended hardware while being wired into a required check. Requesting changes on those two.

## What I checked

- `npx tsc --noEmit`: clean.
- Guards: `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/dungeon_finder.test.ts`, `tests/items.test.ts` pass; `tests/localization_fixes.test.ts` passes after `npm run i18n:gen`.
- Copy scan over the diff: no em/en dashes, no emoji, no `.only(`, no new player-visible strings, no `ALLOW_DEV_COMMANDS` or secrets. i18n N/A is correct.
- All 40 new `*_perf.test.ts` files, run as a set, several times.
- Read `SpatialGrid`, `entity_roster.ts`, and `runMatching`/`tryAssemble` against `release/v0.28.0`.

## Blocking: a proposal formed mid-pass can park the finder on a still-formable backlog

`runMatching()` snapshots the lap state once at entry (the `matchLapSeq !== matchMutationSeq` check). `createProposal()` then bumps `matchMutationSeq` **mid-pass**, but `matchLapSeq` and `matchLapAnchorsVisited` are not re-synced, and `lastQueueLen` is re-read from the now shorter queue. At the truncation check, `matchLapAnchorsVisited >= lastQueueLen` therefore compares an inflated counter against a shrunken length, sets `matchProvenIdleSeq` to the brand new post-proposal seq, and skips the `matchDirty` re-arm. The backlog is declared "proven unmatchable" even though no lap ever walked it in its current shape.

Reproduced against a real `Sim` (`hollow_crypt_normal`, 6 healers queued first, then 120 solo dps, then 6 tanks, so only a tank anchor can seat a comp):

```
start queue = 132
t=126  queue=122  matchDirty=false   <- matching goes idle here
```

At that point two groups had formed and **4 tanks + 4 healers + 114 dps remained, four more fully formable comps**, with `matchDirty` false. Matching only wakes up again when the outstanding proposals expire and `failProposal` calls `armMatching()`, so the queue sits dead for up to `FINDER_PROPOSAL_SECONDS`. It is a bounded stall rather than permanent starvation, but it hits hardest exactly on the busy realm the shared-budget change was written for, and it makes the `matchProvenIdleSeq` contract untrue.

Suggested fix: re-sync the lap when a proposal forms inside the pass (reset `matchLapSeq`/`matchLapAnchorsVisited` right after `createProposal`), and re-arm `matchDirty` unconditionally when a pass both formed something and truncated.

Two related soundness issues in the same logic, worth fixing together:

- `matchLapAnchorsVisited` over-counts across inner `while` iterations. It is only reset at pass entry and on clean completion, so when `changed` restarts the loop against a new shorter `sorted`, anchors already counted get counted again. If the cursor unit was itself consumed into the proposal, `findIndex` returns -1 and traversal restarts from the FIFO front, re-counting from zero. A `Set` of anchor ids visited this lap would be sound where the counter is not.
- A budget-exhausted pass can take the "full uninterrupted pass" branch. `truncated` is only set inside the `for` loop; if an anchor forms a proposal *and* that successful `tryAssemble` consumed the last of the budget, the `for` exits via `if (changed || truncated) break` with `truncated === false`, and the `while` exits on `remaining > 0`. Control lands in the `else` branch, which discards the resume cursor and sets `matchProvenIdleSeq`. Same symptom, different entry path.

Test coverage gap behind all of this: the three added cases in `tests/dungeon_finder.test.ts` are all mutation-free laps, which is precisely the case where the counter logic is sound. There is no case where `createProposal` fires mid-pass and the pass then truncates. That shape deserves a direct test.

## Blocking: the new perf budgets fail on contended hardware, and they gate a required check

These files run in the default `npm test`, so they gate `ci.yml`. Running the 40 new files as a set on a 6-core box under load (roughly what a 4-shard CI matrix looks like), I got 5 failures out of 161:

| Test | Assertion | Observed |
|---|---|---|
| `damage_resolution_perf` (lethal wave) | `< 150ms` | 271ms, and 386/450ms on other runs |
| `damage_resolution_perf` (scaling) | `< 6.66` | 9.27 |
| `market_query_perf` | `< 10ms` | 16.4 |
| `mob_lifecycle_perf` (scaling) | `< 13.99` | 19.1 |
| `party_ops_perf` | `< 5ms` | 5.77 |

`damage_resolution_perf`'s wave budget failed 4 of 5 runs, including in isolation, so 150ms does not look like the claimed 8-10x headroom over a healthy median. Separately, the scaling checks are structurally noise-sensitive: they compare two medians measured minutes apart, so a slow baseline window plus a fast doubled window (or the reverse) breaks the 3.5x ratio with no regression present. I saw `mob_lifecycle_perf` and `damage_resolution_perf` fail that way with nothing changed between runs. Worth measuring the ratio within a single interleaved sampling loop rather than as two separate campaigns, and re-deriving the flat ceilings from a loaded machine.

Related, and the reason this matters more than usual: adding `PERF_GATE_WALLCLOCK=1` to both `ci.yml` test jobs and to `npm run gate` moves real wall-clock measurement into blocking checks at the same time as 160 new wall-clock assertions. The `pr-ai.yml` browser job is correctly non-blocking; I would want the same conservatism applied to the budgets above before they become a merge gate. The 40 new files also add roughly 2 minutes of wall clock to a full run, which is fine sharded, but worth knowing.

## Confirmed good

- `vendorInRange` (`src/sim/items.ts`): behavior-preserving, verified rather than assumed. `SpatialGrid.forEachInRadius` applies an exact `d2 <= r2` test against live `e.pos`, and every entity kind including `npc` is inserted into `ctx.grid` via `addEntityToRoster`, so the candidate set is identical to the old full scan. I wrote a scratch test (buy adjacent to Trader Wilkes succeeds, sell from 500 units away is still denied) and it passes. Nice catch on the original O(entities)-per-transaction shape. Micro-nit: the callback keeps visiting after `found` is true, since the API has no early exit.
- Determinism and parity: the finder draws no rng, the new cursor state is a pure function of tick history, and the parity draw-order digest is green and unregenerated. `DungeonFinderMachine` has no serialize path and `ClientWorld` never runs the finder, so there is no save/load or tri-host divergence. One property worth recording somewhere: the matching *outcome* now depends on machine history through the cursor, not on queue state alone, so two sims reaching identical queue state by different tick paths can match differently. Legal under the repo's determinism definition, but it is new, and it is what makes the stall above hard to reproduce from a state dump.
- Tick phase order, the `SimContext` contract, and sim purity are all untouched.

## Minor

- `tests/CLAUDE.md` gains entries for `mob_update_perf` and `aura_tick_perf` but says nothing about the 40 files this PR actually adds. Given the anchor rule, a one-line pointer to the `*_perf.test.ts` family and the shared recipe (warm up, median, flat ceiling plus doubling scaling check) would help the next person more than the two individual entries do.
- The `rotated` traversal comment says rotating "guarantees each anchor eventually gets a turn ... still in FIFO order once traversal wraps". True for the anchor, but `tryAssemble` still builds `candidates` from the **unrotated** FIFO `sorted` before `slice(0, FINDER_MATCH_UNIT_WINDOW)`, so a late tank anchor can still only ever pair with the 23 oldest compatible units. That window is pre-existing, but the comment reads like the rotation fixed reachability, and it did not.

jgyy:
Thanks both, pushed 8b648518a addressing the two remaining open items from the deep-sweep follow-up:

**B. Vacuous phase-lap accumulators (10-file finding, 6 actually applicable)**

Added an explicit `lapTotal > 0` guard (mirroring `mob_update_perf.test.ts`'s existing `expect(mobLapTotal).toBeGreaterThan(0)`) to every file that installs the `cfg.perfLap` phase-matching hook through the `as unknown as` cast: `aura_tick_perf`, `dungeon_instance_perf`, `delve_runs_perf`, `ground_aoe_projectile_perf` (both describe blocks), `pvp_minigame_tick_perf` (both phases), and `world_boss_perf`. Each measurement helper now accumulates a running total alongside the per-tick sample and every `it` block asserts it's nonzero before trusting the median, so a phase-string rename in `sim.ts` now fails loudly instead of the accumulator silently staying 0 and every budget assertion passing vacuously.

Of the original 10 named, `party_ops_perf` and `pet_ai_perf` don't use this hook at all: `party_ops_perf` times real `partyInvite`/`partyAccept`/`convertPartyToRaid`/`setPartyLootMaster` calls directly, and `pet_ai_perf` calls `updatePet` directly against `ctx` (both call out why in their own comments), so there's no phase string to rename and no vacuous-pass risk to guard against. `dungeon_finder_perf` and `mob_social_aggro_perf` are the same shape. Left those four untouched.

**A. `terrain_sampling_perf` budget miscalibrated (~1.5x headroom, not ~8-10x)**

Confirmed on this host too: median is genuinely in the low tens of ms for the full 4096-point/5-call grid, not the "low single-digit ms" the old comment claimed. Raised the bound from 60ms to 120ms (roughly 3x the ~30-40ms measured median, with margin over the ~79ms contended-CI spike you reported) and corrected the comment to describe the measured reality instead of the wrong assumption.

**Still open / deferred**, unchanged from the last status update: #4 (the ~35-file `Math.max(smallMedian * 3.5, 5)` mechanical tightening), the #6 population-shape items (vendor_transaction/spatial_grid/collider_resolution/threat_table/gather_nodes_lookup/dungeon_instance/character_effects), and the #7 nits (`hit_table_perf.test.ts:157`'s always-true `toBeGreaterThanOrEqual(0)`). Happy to take a follow-up pass at those.

**Verification:** `npx tsc --noEmit` clean; `npx vitest run tests/aura_tick_perf.test.ts tests/dungeon_instance_perf.test.ts tests/delve_runs_perf.test.ts tests/world_boss_perf.test.ts tests/ground_aoe_projectile_perf.test.ts tests/pvp_minigame_tick_perf.test.ts tests/terrain_sampling_perf.test.ts` all green (20 passed); biome clean on changed files. Ran the full suite too: the only failures there are pre-existing timeout flakes under full-run contention in files this change never touches (`world_boss.test.ts`, `vale_cup_online.test.ts`, `parity/coverage.test.ts`, `sim.test.ts`, `threat.test.ts`) plus an environment-local `tsc` toolchain-pin quirk in `new_endpoint.test.ts`, confirmed unrelated by reproducing them on a clean stash of this branch before these edits.

jgyy:
Addressed the latest review round (EnriqueGF, 2026-07-20).

Blocking:

1. Dungeon Finder matching stall (src/sim/social/dungeon_finder.ts runMatching): createProposal() bumps matchMutationSeq mid-pass when a match forms inside the shared-budget while loop, but the lap-completeness accounting (matchLapSeq / the anchors-visited counter) was only resynced at the next runMatching() call, not the instant the mutation happens. A stale counter could then be compared against a lastQueueLen already shrunk by that same mutation, wrongly declaring a still-formable backlog proven unmatchable and letting matching go idle until an unrelated join or leave woke it again.

   Fixed by resyncing the lap (matchLapSeq plus the anchors-visited state) the moment a match forms, not only at call entry, and by switching the counter from a plain increment to a Set of visited anchor ids so a re-sorted backlog cannot double count an anchor. Also treat a budget exhaustion that coincides with the same iteration a match forms as truncation, since the old flag could miss it. Also corrected a stale comment claiming the anchor rotation fixes reachability for a late anchor: it only fixes anchor turn order, since tryAssemble still builds its candidate window from the unrotated FIFO list.

   Reproduced the exact stall with the review's own repro shape (6 healers, 120 dps, 6 tanks): before the fix, only 2 of 6 tank-anchored comps formed before matching incorrectly went idle. Pinned with a new test in tests/dungeon_finder.test.ts asserting all 6 tanks eventually seat and that the finder does not sit idle for long stretches while an unseated tank and a queued healer remain.

2. Perf budgets failing on contended hardware (damage_resolution_perf, market_query_perf, mob_lifecycle_perf, party_ops_perf): widened the flat ceilings and the scaling-ratio bound (3.5x to 6x, floors raised to match) that the review measured failing on a loaded box with no regression present (damage_resolution_perf lethal wave 271 to 450ms vs a 150ms ceiling now 600ms; scaling 9.27x vs 6.66x now bounded at 6x/8ms; market_query_perf 16.4ms vs 10ms now 25ms; mob_lifecycle_perf scaling 19.1x vs 13.99x now 6x/8ms; party_ops_perf 5.77ms vs 5ms now 12ms). Each still catches an order-of-magnitude or quadratic regression, just with real headroom for shard contention.

Minor:

- tests/CLAUDE.md now points at the whole *_perf.test.ts family (and how to enumerate it) instead of naming only mob_update_perf/aura_tick_perf.
- scripts/gate.mjs: PERF_GATE_WALLCLOCK now uses ?? instead of a flat override, so an operator's explicit opt-out (PERF_GATE_WALLCLOCK=0) survives.
- .github/workflows/ci.yml: bounded the sharded test step to --maxWorkers=2 (pr-gate and release-gate) so ARM C's real hrtime measurements and the new perf budgets are not sharing cores with sibling worker threads.

Verification: npx tsc --noEmit clean; tests/dungeon_finder.test.ts, the four widened perf files, tests/architecture.test.ts, tests/world_api_parity.test.ts, and tests/localization_fixes.test.ts all pass locally; biome check on every changed file is clean (pre-existing noExplicitAny warnings only, no errors or format diffs). A full npm run gate was not feasible in the working sandbox (an unrelated, pre-existing incomplete local install caused a toolchain-pin test and the whole admin/jsdom suite to fail regardless of this change), so I ran the targeted equivalent instead. On the actual PR, all required checks are green (PR gate 4/4 shards, PR checks, Format+lint, Browser regressions); the one remaining red job, "Real-browser frame budget," is informational/non-blocking and fails at the tour-boot step (pre-existing, unrelated to this change, tracked by the round-1 should-fix about that job needing a real green CI run before it can gate anything).

Deferred, not addressed in this round: the perf-budget-report job's frameP95 assertion still falls back to the committed cross-machine baseline instead of a runner-measured HUD_PERF_BUDGET_TOUR_FRAME_BASELINE value (round 2, should-fix). That needs an actual green run on CI hardware to capture a real baseline, which is out of scope for this round's fixes.

Rubsey (commented):
Round 4. I verified every fix from the prior rounds against the head rather than the commit messages, and the good news is real: the mid-pass lap resync, the visited Set, the post-loop budget catch, the lap-total guards in the four named files plus pvp/world_boss, the literal pins, the widened ceilings, both ARM C timeouts, the gate.mjs opt-out, and the maxWorkers bounds all landed and are correct individually. I also mutation-tested the four added dungeon_finder behavior cases: each one turns red when its fix is reverted, and the round-3 test genuinely cannot be masked by the proposal-expiry wake-up (600-tick loop, earliest expiry wake at ~tick 726). That is careful work.

The bad news is the lap accounting still has two reachable soundness holes, both reproduced deterministically against the unmodified head, and the new CI job is red on arrival for a root cause nobody guessed yet. Details below; repro tests for the two finder bugs are described precisely enough to paste into the suite, and I can share the scratch files if useful.

**Blocking**

1. `src/sim/social/dungeon_finder.ts:1001` (H1): when the shared budget dies inside tryAssemble's DFS on an anchor's LAST eligible activity (the common case: most units queue one activity), the inner activity loop ends naturally, the budget checks that set `truncated` never run again, `if (changed || truncated) break` does not fire, and the partially-searched anchor is recorded in `matchLapVisitedAnchorIds` as fully attempted. The proven-idle latch can then fire over a queue that was never fully searched, and unlike the round-2 stall there is no proposal-expiry wake-up: the queue parks until an unrelated join/leave/role change. Deterministic repro (passes at head): all hollow_crypt_normal at level 8, FIFO order 1 solo healer, 12 premade two-tank pairs (unusable for the one-tank comp), 3 solo dps, 148 premade four-dps groups, then a solo tank last. Result: zero proposals ever, visited reaches 165/165, idle latches, and 300 further ticks (15 sweep cycles) stay parked with `matchDirty` false; a control that merely re-arms with the cursor placed just before the tank forms {tank, healer, 3 dps} in one tick through the machine's own search (solution cost ~251 nodes against the 4000 budget).

2. `src/sim/social/dungeon_finder.ts:998` (the mirror hole): when the budget dies inside the FIRST activity of a two-activity anchor, the next activity's top-of-loop check sets `truncated` and breaks BEFORE the mark/cursor lines, so the anchor is never marked and the cursor never advances. The next pass restarts at the same anchor with a fresh budget and dies identically, so the lap can never complete, the idle latch never fires, and matching re-arms every tick forever (the exact eternal churn the latch was built to prevent) while permanently starving every later anchor. Deterministic repro (passes at head): 24 level-20 solo dps queued for both hollow_crypt_heroic and sunken_bastion_heroic (catalogue order runs hollow first), then a tank and healer queued for sunken only: 200 of 200 ticks dirty, cursor never advances, zero proposals; change the one variable (dps queue sunken only) and the tank proposal forms immediately. Multi-heroic queueing at 20 is a mainline flow, so this shape is realistic, not adversarial.

   Both bugs are two faces of one missing distinction: the mark/cursor rule cannot tell "this anchor's search completed (or was bounded by design)" from "this anchor got the dregs of the budget". A trap for the fix: the naive guard (do not mark when `remaining <= 0`) breaks this PR's own "goes idle after proving a role-imbalanced backlog unmatchable" test, which currently passes BECAUSE of the H1 mis-mark (every anchor in that scenario costs ~17k nodes against the 4000 budget, so visited only ever grows via the fall-through, one anchor per pass, exactly as the test's comment describes). A rule that works for both: an attempt that STARTED its pass with the full fresh budget counts as attempted even if truncated (the search is bounded by design), while a dregs attempt does not count and must not advance the cursor. Under that rule the goes-idle test still latches one anchor per pass, the H1 tank gets retried as a pass-starter and forms at ~251 nodes, and the pinned two-activity anchor is eventually counted so the lap completes. Please add both repro shapes as regression tests, since every round of this rework has been kept honest by exactly this kind of case.

3. `.github/workflows/pr-ai.yml` perf-budget-report is red on this PR, and the root cause is not CI hardware variance: the committed `hudHotDomWrites` anchor (153) is simply a month stale. The metric itself is portable exactly as the baseline claims: CI ubuntu measured 467 desktop, and a fresh tour at the PR head on my own machine class measures exactly 467 too (fps 5.9 local vs 1.0 CI; 466 of the 467 are boot-time establishing writes). The baseline was captured 2026-06-27 and a month of HUD features has landed on the release branch since (attunement tutorial panel, minimap ornament, bags enchant/salvage actions, professions UI). Mobile now measures 548, so the baseline's "byte-identical across viewports" single-row model is stale too. The fix is to re-mint the baseline count rows in this PR (desktop 467, mobile 548, as per-viewport rows or one max row), not a CI-side env override, which would defeat the one portable arm. Round 1 asked for one green run on CI hardware before landing the job and that still has not happened (2 of the 3 completed runs died at tour boot on the runner before even reaching the gate). If getting it green turns into a slog, the honest alternative is to pull the job from this PR and land it separately once proven.

**Should-fix**

4. The job's frameP95 arm is a tautology as wired: `src/game/perf.ts:374` clamps every frame sample to 250ms and the committed baseline is exactly 250ms, so `frameP95 <= 250` cannot fail on any machine (that is why it "passed" on CI, contrary to round 2's prediction). Pre-existing instrumentation, but this PR is what promotes it into an every-PR report whose header claims frameP95 regressions get surfaced. Either pin a real sub-cap value for the runner or annotate the arm as same-machine-manual-only alongside the baseline re-mint.

5. `.github/workflows/pr-ai.yml:214`: the mobile frame-budget step has no `if: always()`, so it is skipped whenever the desktop step fails (confirmed on this PR's run). Fixing only the desktop anchor would surface the mobile failure one CI cycle later; add `if: always()` (as the verify and stop steps already have) or fold both viewports into one step.

6. Round-2 item still open: the "report" job uploads no artifact and posts no comment; `tmp/perf-tour.json` dies with the runner. Diagnosing this failure required a full local reproduction that the JSON would have answered in seconds. An `actions/upload-artifact` step for it is four lines.

7. `tests/deeds_evaluator_perf.test.ts:44`: this file uses the same perfLap phase hook the b7f5e0d guard fix covered elsewhere, but got no lap-total nonzero guard, so a `'deeds'` phase rename zeroes the accumulator and both budget asserts pass vacuously (the shape test calls updateDeeds directly and would not notice either). Same one-guard fix as its four siblings.

8. `tests/arena_matchmaking_perf.test.ts:40`: the measured median only exercises the queue liveness filter. All 4 arena slots fill in sample 1 and the test never ticks, so matches never end, and `matchmakeArena1v1` early-returns at `src/sim/social/arena.ts:619` before the closest-rating scan for samples 2 through 50. The quadratic rating-scan regression the file's own comment claims to catch is invisible to it. End matches (or reset `ctx.arenaMatches` and slots) between samples.

9. `tests/mob_lifecycle_perf.test.ts:52`: `measureCascade` marks all 300 mobs dead before any `frenzyPackmates` call, and `lifecycle.ts` rejects dead candidates, so the frenzy grant path (the aura push/refresh/emit) is never measured; every call scans and rejects. The shape test has the same construction, so its `frenzied > 0` assert is satisfied only by incidental natural wolf spawns near the cluster point, meaning an unrelated content change near (0, 60) flips it red. Kill-then-frenzy progressively (mark one dead, frenzy it, continue) fixes both.

10. `tests/profession_tick_perf.test.ts:68`: the gather quarter erodes to the "not respawned for you yet" fail-fast mid-measurement: the measure loop never ticks so `ctx.time` is frozen, per-player-per-node cooldowns never elapse, and only 150 distinct (player, node) pairs exist, which warmup plus ~2.5 volleys consume. The comment claiming round-robin avoids the cooldown is wrong past the first cycle. Bounded (craft/enchant/salvage stay real), but the file overstates what it measures.

11. Round-1 items acknowledged but never fixed, carrying forward: `proc_dispatch` (the measured waves run post-revert loadouts and hit the early-return paths in both proc systems; setting `meta.equipment` so the recalc preserves the loadout is the fix the file's own comment points at), `vfx_pool_alloc:89` (the allocation probe still never bursts inside the probed thunk, and the follow-up burst loop asserts lengths, not buffer identity; capture `poolArrays(vfx).pos` before and `toBe` it after), `character_effects` (still no shape assertion of any kind), and `gather_nodes_lookup` (still benchmarks a test-local re-implementation over compile-time Record literals; the shipped consumer is never imported; fix it or move it to the PR's honest-skips list).

12. `tests/vendor_transaction_perf.test.ts:109`: mutation-verified that reverting `vendorInRange` to the exact pre-fix full scan passes all three tests, so the header's claim that the scaling check catches that regression shape is wrong at these populations. The behavioral guards do exist (the correctness arm here plus the pre-existing out-of-range denial in `tests/sim.test.ts`), so either raise the padded populations until the reversion actually trips a bound, or soften the header and add the out-of-range denial to this file's correctness arm.

13. Stale ARM C docs (round-1 item): `tests/server/perf_gate.test.ts:50` and `:364` still say "SKIPPED by default" and `:66` still says "RUN THE WALL-CLOCK ARM MANUALLY", and `docs/api-pipeline/state.md:1104` still frames the arm as opt-in. After this PR it runs in the gate and both CI test jobs; the comments now mislead operators about what CI runs.

**Nits**

14. `tests/dungeon_finder_perf.test.ts:97`: the scaling-test comment misattributes which assertion pins the fixed bug. Measured with the true pre-fix per-anchor budget reinstated, cost is linear (ratio 2.01x, passing the 3.5x bound) and only the flat 50ms ceiling fails (313ms). Correct the comment so a future maintainer does not relax the ceiling believing the scaling arm backstops it.

15. `src/sim/social/dungeon_finder.ts:969`: a cursor unit consumed into a proposal as a non-anchor member leaves `matchCursorUnitId` dangling; the next iteration's findIndex returns -1 and rotation restarts at index 0, re-spending budget on the oldest anchors. Bounded (each occurrence requires a formed proposal), but a cursor repair in createProposal would fall out of the blocking rework naturally.

16. `tests/nythraxis_encounter_perf.test.ts:98`: `roomPlayers` uses radius 9999, which matches every live player anywhere, so the pile-up shape assert is vacuous; use a real room radius.

17. `tests/pvp_minigame_tick_perf.test.ts:204`: `runUntilAllLiveGeneric` returns silently after MAX_TICKS, and the scaling arm asserts only nonzero lap totals, so slot-pool contention could quietly measure fewer than the claimed 2x lobbies; assert the lobbies actually went live.

18. Shape-pin consistency: `effect_dispatch:188` pins `> 0` where its siblings pin literals (the builder silently continues on missing entities), and `aura_tick` has no shape test at all (the lap guard proves the phase ran, not that 80x60 auras exist). `loot_roll:90` lets raid inventories grow across measured corpses (addStacked has no capacity gate), drifting the measured load upward; a fresh corpse per sample makes the steady-state claim true.

19. Timeouts: `delve_interactable_visibility` is the one file whose real measurements ride the 5s default, and `pick_resolution` / `static_matrix` / `vfx_pool_alloc` carry none (fine today, but the ARM C commit's rationale says every perf file carries one). The heavy untimed shape tests in `chat_broadcast:88`, `arena_matchmaking:91`, and `deeds_evaluator:104` are the plausible default-timeout flake candidates on a starved shard.

Recorded, not asks: the respec role-flip while queued (a sticky pre-level-5 multi-role selection changing formability with no re-arm) is a pre-existing gap in base that the new idle latch makes marginally more visible; and the matching outcome now depending on cursor history rather than queue state alone is legal under the repo's determinism rules but worth remembering when debugging from state dumps. Everything else checked clean: every queue mutation site routes through armMatching or the createProposal seq bump, no rng draws, no new imports, the sort is a total order, the Set is never iterated, the grid swap in vendorInRange is exactly behavior-preserving, the PERF_GATE_WALLCLOCK=0 opt-out works against the test's exact `=== '1'` gate, the shard flags pass through to vitest 4 correctly (3m59s to 4m46s per shard on the green run), merge against release/v0.29.0 is conflict-free, and the copy scans are clean.

jgyy:
Round 4 (Rubsey) addressed. Pushed two commits to `feature/perf-test-expansion`.

## Blocking: both dungeon_finder lap-accounting soundness bugs (H1 + mirror)

Root cause was one missing distinction, exactly as diagnosed: `runMatching` could not tell "this anchor's search STARTED with the whole fresh budget and got bounded by design" from "this anchor got the dregs of an already-spent budget."

- **H1 (dregs last-activity fall-through):** when the budget died inside `tryAssemble` on an anchor's LAST eligible activity, the inner loop ended naturally, `truncated` was never set, and the partially-searched anchor fell through to the fully-visited mark + cursor advance. Enough of those mis-marks drove `matchLapVisitedAnchorIds` to the queue size and latched `matchProvenIdleSeq` over a queue that was never fully searched, parking with no wake-up.
- **Mirror (fresh first-activity truncation never marked):** when the budget died in the FIRST activity of a two-activity anchor, the next activity's top-of-loop check set `truncated` and broke BEFORE the mark/cursor lines, so the anchor was never marked and the cursor never advanced: the lap could never complete, the idle latch never fired, and matching re-armed and burned a full budget every tick forever.

**Fix:** capture `anchorStartedFresh = budgetBox.remaining === FINDER_MATCH_NODE_BUDGET` at the top of each anchor, and track a per-anchor `anchorTruncated` (set both when the top-of-activity check trips AND when `tryAssemble` runs the shared budget to zero mid-DFS). A truncation marks the anchor and advances the cursor only when it started fresh (bounded by design); a dregs truncation leaves the cursor and visited set untouched so a later pass retries the anchor as a pass-starter with a full budget.

**Regression tests** (both fail on the pre-fix tree, verified by stashing the source fix):
- `does not falsely latch idle when a dregs last-activity truncation mis-marks a never-fully-searched anchor`: solo healer, 12 two-tank premade pairs, 3 solo dps, 148 four-dps premade quads, solo tank last, all `hollow_crypt_normal`; the tank must still seat `{tank, healer, 3 dps}`.
- `does not churn forever when a fresh first-activity truncation on a multi-activity anchor is never marked`: solo dps queued for BOTH heroics (hollow runs first, so the budget dies in the hollow DFS), then a tank and healer for sunken only; the tank must seat a sunken group once the cursor advances. Uses 22 dps rather than 24 so the last-joined healer stays inside the anchor's FIFO candidate window (`FINDER_MATCH_UNIT_WINDOW`), otherwise no formable comp exists behind the churn and the repro would be vacuous.

The existing `goes idle after proving a role-imbalanced backlog unmatchable` test still passes: every anchor there costs more than the whole budget, so the first anchor is a fresh-budget-bounded truncation, still marked one-per-pass, exactly as before.

## Also fixed

- **#3 baseline re-mint:** re-minted `hudHotDomWrites` per viewport (desktop 467, mobile 548) and updated `readBaselineBypassCount` to take the strictest (max) committed row, mirroring the existing `hudHotDomSkipRate` floor, so ARM 3 gates every viewport against 548. Rewrote the stale "byte-identical across viewport / 153" prose.
- **#4 frameP95 tautology:** annotated the arm as a same-machine-manual signal in the baseline (perf.ts clamps every sample to 250ms and the baseline is exactly 250ms, so the CI arm cannot fail); the portable CI signal is the bypass count.
- **#5:** `if: always()` on the mobile frame-budget step so a desktop failure no longer masks a mobile regression.
- **#6:** upload `tmp/perf-tour.json` as an artifact (`if: always()`).
- **#7:** added the lap-total nonzero guard to `deeds_evaluator_perf.test.ts` (both budget tests), matching the `b7f5e0d` guard on its siblings.

## Deferred (items 8 to 19)

Not folded in to keep this round scoped to the blocking bugs plus the perf-gate honesty items. These are should-fix/nit measurement-fidelity issues in other `*_perf.test.ts` files (arena early-return, mob frenzy reject path, profession cooldown erosion, proc_dispatch/vfx_pool/character_effects/gather_nodes carry-forwards, vendor reversion sensitivity, stale ARM C docs, and the `dungeon_finder_perf` comment attribution, plus the `createProposal` cursor-repair nit). None change shipped behavior; happy to take them in a follow-up.

## Verification

`npx tsc --noEmit` clean. `npx vitest run tests/dungeon_finder.test.ts tests/dungeon_finder_perf.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/hud_perf_budget.test.ts tests/deeds_evaluator_perf.test.ts` all green (385 passed, 3 skipped). `tests/parity` green (no rng draws added; the finder draws none). Biome clean on changed files. Merge against `release/v0.29.0` is conflict-free.

jgyy:
to close this first, i wanted to test smaller test cases PR that is less then 3000 lines of code change and get it automatically merged first.

PR #2124, opened 2026-07-18
style(design): fantasy HUD redesign phase 1, warm carved-wood signature

jgyy:
## Summary

Redoes PR #2108 (closed, superseded by this PR) as a fantasy HUD redesign: retunes the
shared token/theme system from the blue-black-midnight signature to a warm carved-wood
signature, and applies it across the primitives and windows that compose with `.panel`,
`.btn`, and the new `--control-*` bevel tokens.

This PR has been through two review passes. The first reverted several ornaments that
did not work in practice (a quest-tracker card that was an empty rectangle whenever no
quests were active, a circular ring on ability icons, a scalloped party-frame crest) and
reworked the right-side icon rail from one long column into two. The second tightened
the two-column rail (a spanning grid item was inflating the column tracks and pushing
the columns apart), put the bag filter chips on one row instead of two, and turned the
Esc Game Menu into a 2-column grid. The screenshots below are current as of the latest
commit.

```mermaid
flowchart LR
  subgraph theme[src/ui/theme.ts]
    KNOBS["classic preset knobs\npanel #1c130b / border #8a5c28"]
    DERIVE["themeCssVars()"]
    KNOBS --> DERIVE
  end
  subgraph tokens[src/styles/tokens.css]
    PREBOOT["pre-boot :root defaults\n(synced fallback)"]
  end
  DERIVE -->|"--panel-bg / --panel-edge\n--control-bg / --control-border"| VARS[":root inline style\n(runtime, wins over tokens.css)"]
  PREBOOT -.->|"same values, pre-JS fallback"| VARS
  VARS --> PANEL[".panel\n(carved frame)"]
  VARS --> BTN[".btn / .woc-balance / .bag-chip\n(embossed raised bevel)"]
  VARS --> INPUTS[".hud-select / .bag-search / .bag-sort\n(engraved inset bevel)"]
  PANEL --> WINDOWS["Every window that reuses\n.panel (Bags, Bank, Character, Options, ...)"]
```

### What changed

- New reference images `docs/design/design-language/hud-style-reference-01.png` through
  `-03.png`, cited from DESIGN.md section 2 ahead of the two legacy images (kept as
  secondary references).
- DESIGN.md section 3 (visual signature): warm carved-wood umber fill, a heavier
  bronze-and-gold frame with a warm inner bevel, in place of blue-black plus a hairline
  gold edge.
- `src/ui/theme.ts`: the `classic` preset's `panel`/`border` knobs retuned
[...]

PR #2152, opened 2026-07-19
style(ui): fantasy HUD structural ornament system, corners/rings/banners

jgyy:
## Summary

Redesigns the classic HUD chrome with a fantasy hand-gilded gold ornament system, superseding
#2124. Everything is driven by one procedural, deterministic primitive module
(`src/ui/ornament_svg.ts`): colorless SVG shapes exposed only as CSS `mask-image` values, so
color always comes from the existing theme tokens and repaints for free on every preset.
DESIGN.md's principle 6 is amended to "clarity first, ornament in service of immersion,"
documented as a component primitive in a new section 10.8.

This PR has gone through six rounds of direct visual feedback against the shipped screenshots
(and a couple of reference mockups). Rather than one big changelog, here's what each round
actually fixed, since several of them reverse or refine earlier rounds:

- **Round 1**: the base ornament system: a corner bracket-and-gem motif (4 mirrored
  orientations from one path), a noisy hand-gilded edge ribbon, unit-frame/minimap rings, and
  a background grain texture. Wired onto `.panel` (all windows + several HUD panels for free),
  `.portrait-wrap`/minimap (rings), `.action-btn` (corner accents).
- **Round 2**: thinner/more detailed rings, tapered non-rectangular window tops, ornament
  spread to every HUD component.
- **Round 3**: removed a disconnected floating gold banner in the Daily Rewards header (a
  sizing bug, not a look worth keeping); replaced it with seamlessly tileable/wrappable
  noise-based edge and ring ornament using a small seeded PRNG (integer-frequency sine
  harmonics, so a tile repeats or a ring closes 360 degrees with no visible seam).
  General coverage was widened for the first time to feature-window bodies past the
  always-on chrome from round 1.
- **Round 4**: more thickness variation, real color variation (a `repeating-conic-gradient`
  keyed to theme tokens, not one flat hex), gilded edges on all 4 sides of every window (a
  z-index collision with the sticky title bar was hiding the top edge), gilded borders on
  buttons and circular badges (portrait/level-chip/minimap), and a colorless background grain
  layer that works over any fill color via `background-blend-mode: overlay`. Scope: user chose
  "everywhere, including functional bars" and "full send" via AskUserQuestion.
- **Round 5**: the geometry was too thick/wavy (shrunk noise amplitude ~10-15x for
  near-straight lines with only slight variation) while color was made MORE varied
  (a seeded random draw over a larger palette) -- two independent knobs, not a contradiction.
  Circular components dropped from a twin-ring to a single ring; the nameplate dropped a
  redundant outer frame that was stacking against each bar's own edge. Added a genuinely
  gold-toned `giltDustBackground()` layer (radial-gradients over `--gold`) for "hand-gilded"
  color noise on backgrounds, distinct from the colorless grain texture. Fixed a real bug where
  every prior round's "max graphics" screenshot claim was false: the client's first-run
  graphics auto-detect was silently overwriting the seeded preset because the capture script
  never seeded the `graphicsDefaultApplied` marker alongside it.
- **Round 6 (this one)**: direct feedback that round 5 still overshot in a few places. The
  portrait/minimap ring and action-bar border are thinner again; the nameplate carries less
  color noise; the gold-dust texture is removed ENTIRELY from every button (action bar, pet
  bar, stance bar, chat tabs, minimap-zoom, mail/raid-lockout indicators, micro buttons,
  close button, generic `.btn`) rather than just toned down, since it read as ugly at button
  scale even reduced. The tapered/chamfered window top is removed outright, reverting every
  window to a plain rectangle with the standard 4-corner motif (this also retires the taper's
  supporting code as dead: `windowTopOrnamentMaskImage`/`taperAccentMaskImage`, the
  `--ornament-window-top` token). Four windows move from a single stacked column to a
  2-column grid: the Options quick-menu, Daily Rewards tasks, Crafting recipes, and the Vale
  Cup role picker. Mobile-touch mode gets its own new decorative ring (reusing the portrait
  ring shape) around the movement/camera joystick wheel and its drag knob, layered outside
  their existing functional border so the gold "engaged" state color still works.

## Related issues

Supersedes #2124 (closed, referencing this PR).

## Type of change

- [x] Feature: new functionality (the ornament primitive system)
- [ ] Bug fix
- [x] Documentation (DESIGN.md principle 6 + new section 10.8)
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands: `npx tsc --noEmit`, `npm test` (full suite; two unrelated timing-sensitive sim
  tests flaked under full-suite worker contention and passed cleanly in isolation, matching
  the pre-existing pattern this repo's CLAUDE.md documents), `npm run ci:changed` (biome,
  warnings only, no errors), the pre-push QA floor (typecheck + guard tests + lint + copy
  rules, green).
- Manual steps: captured 37 screenshots across desktop and mobile-landscape via the
  offline client (`scripts/pr_screenshots.mjs` / a driven `window.__game` session), with
  Ultra graphics correctly seeded (verified live via `data-fx-level`/`graphicsPreset`, not
  just the write call, per the round-5 bug above). Verified live in-browser that: buttons
  render with no dust speckle, every window is a plain rectangle with all 4 corners motifed,
  the 4 target windows lay out in 2 columns without overflow, and the new joystick ring
  renders cleanly at rest and while dragging.

## Screenshots / recordings

37 screenshots under `docs/screenshots/fantasy-hud-ornament/` (round 6 `after-*` set; the
original pre-PR `before-*` baselines are untouched):

**Desktop overview & detail**
`after-hud-desktop.png`, `after-detail-portrait-actionbar.png`, `after-detail-minimap.png`,
`after-detail-action-buttons.png`, `after-detail-micro-buttons.png`,
`after-detail-stance-bar.png`, `after-detail-chat.png`, `after-detail-chat-tabs.png`,
`after-detail-xpbar.png`, `after-detail-nameplate.png`

**Windows (desktop)**
`after-window-bags.png`, `after-window-bank.png`, `after-window-character.png`,
`after-window-spellbook.png`, `after-window-talents.png`, `after-window-social.png`,
`after-window-mailbox.png`, `after-window-map.png`, `after-window-market.png`,
`after-window-meters.png`, `after-window-quest-log.png`, `after-window-leaderboard.png`,
`after-window-calendar.png`, `after-window-arena.png`, `after-window-card-duel.png`,
`after-window-deeds.png`, `after-window-dungeon-finder.png`, `after-window-emote-wheel.png`,
`after-window-town-focus.png`

**2-column window conversions**
`after-window-options.png`, `after-window-daily-rewards.png`, `after-window-crafting.png`,
`after-window-vale-cup.png`

**Mobile (landscape)**
`after-mobile-hud.png`, `after-mobile-bags.png`, `after-mobile-joystick.png` (new decorative
ring), `after-mobile-options.png` (2-column grid on mobile)

---

### Quality

- [x] **The full gate passes.** `npm run gate`-equivalent checks are green (see "How was this
      tested" above); tests were updated for every shape/behavior change across all 6 rounds.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified on desktop and mobile-landscape viewports;
      touch targets and input font floors are untouched by this PR (pure visual layer).
- [x] **Accessible.** `forced-colors: active` system-color border restoration is kept current
      for every control whose resting border became `transparent` (base.css net); no motion is
      introduced, so `prefers-reduced-motion` is unaffected.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (pure CSS/SVG ornament + layout).

### Hygiene

> Commit messages and PR titles use [Conventional Commits](https://www.conventionalcommits.org/)
> with a required scope (`feat(talents): ...`, `fix(net): ...`).


EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0`. Ran locally: `npx tsc --noEmit` (clean), `tests/ornament_svg.test.ts`, `architecture`, `css_corpus`, `css_value_validity`, `styles_extraction`, `client_shell`, `hud_perf_budget`, `mobile_window_coverage`, `backdrop_filter_survival`, `focus_visible_guard`, `theme` (all green; `localization_fixes` only fails on the known "run `npm run i18n:gen` first" precondition, unrelated to this PR). Also reproduced two of the findings below in headless Chromium with pixel readback rather than reasoning about them.

The ornament generator itself is good work: genuinely pure, deterministic, seam-free by construction, well tested, zero painter/elision-seam impact, no new strings, no hard-coded themed colors, and the `forced-colors` restoration for the `border: transparent` controls is exactly right. `hud.ts` and `renderer.ts` are untouched. My findings are all on the CSS consumption side.

---

### 1. Blocking: the ornament scrolls away inside every scrolling window

`.panel::before` / `.panel::after` (base.css) are `position: absolute; inset: 0`, and `.window` is a scroll container (`overflow-y: auto`, layout.css). An absolutely positioned child of a scroll container is sized to the *padding box* (the visible height) but anchored at scroll origin, so it scrolls up and out with the content. On any window whose content overflows (character, bags, deeds, quest log, spellbook, ...) the entire gilded frame disappears as soon as the player scrolls.

Verified in Chromium with a minimal repro of the same rule shape: the pseudo-element computes `height: 200px` against a `clientHeight` of 200 with `scrollHeight` 900, and a screenshot sampling shows the bottom band at rows 214-219 at `scrollTop: 0` and **no** band at all at `scrollTop: 400`.

This file already knows the failure mode: the `.window.window-resizable` comment at layout.css:149-153 says the resize grip is drawn as a background layer on the window itself precisely because "an element or pseudo-element would scroll away with the content." The same precedent is the cheapest fix here (move the ornament to `background-image` layers on `.window`), or wrap the scrollport in a non-scrolling frame element.

### 2. High: the top-right taper gem is 14px off its chamfer apex

layout.css:81/86:

```css
mask-position: ..., calc(var(--window-taper) - 7px) 0, calc(100% - var(--window-taper) - 7px) 0;
```

The comment says "mask-position places an image's TOP-LEFT at the given point" - true for a pure length (so the top-left accent is correct), but a **percentage** in `mask-position` resolves against `positioning area - image size`, not the box. So `calc(100% - 27px)` lands the 14px accent at `(W - 14) - 27 = W - 41`, while the apex-centered target is `W - 27`.

Verified in Chromium: a 14px mask on a 400px box with this exact declaration paints at left offset **359**, not the intended 373. It is also visible in this PR's own `after-window-character.png` - the top-left gem sits on the chamfer apex, the top-right one is clearly inboard of it.

Fix: `calc(100% - var(--window-taper) + 7px)`.

### 3. Medium: every edge-tile `mask-size` override violates the contract the generator declares

`ornament_svg.ts:242-247` (repeated in base.css) states an override "may render the SAME tile smaller via mask-size as long as it keeps this 96:10 aspect ratio, or the wobble distorts instead of just rescaling." I extracted every consumer; none of them is 9.6: `24px 2px` = 12.0, `20px 2px` = 10.0, `36px 4px` = 9.0, `2px 12px` = 6.0, `2px 15px` = 7.5, `2px 18px` = 9.0, `16px 1.5px` = 10.7. Either the values or the comment is wrong. Whichever way it goes, this is a good candidate for a CSS source-scan test, since it is exactly the class of drift that already bit once in `0ab583184` ("correct mask-repeat layer count"). I re-checked layer-count parity across all ornament rules and it is correct today, but nothing pins it (nor `TAPER_ACCENT_SIZE` = 14 against layout.css's `14px 14px` and `7px`, nor `EDGE_TILE_LENGTH/THICKNESS` = 96/10 against base.css's `96px 10px`).

### 4. Medium: no low-tier drop for the ornament

The diff adds ~128 `mask-image` declarations and zero `data-fx-level` / `--fx-*` hooks, so the `low` preset now composites strictly more than before this PR, against `src/styles/CLAUDE.md`'s stated intent that "the low tier drops the glass rule and the heavy-shadow buckets so a cheaper preset costs less to composite." Ornament is purely cosmetic and carries no actionable information, so shedding it at `low` is exactly the sanctioned kind of tier knob and costs nothing on fairness. Worth adding a `:root[data-fx-level="low"]` rule and pinning it in `tests/ui_effects_wiring.test.ts`.

### 5. Low: gilt dust on the health/resource bar trough

`.bar` now paints `var(--ornament-dust)` (18 gold radial blobs at 28-66% strength) as its topmost background layer over `#1a1a1a`, and `.bar-fill` covers only the filled portion, so the gold blobs land specifically on the *empty* part of an HP bar. Not a fairness break (no tier gate, nothing hidden or delayed), but target-HP granularity is actionable information and this slightly softens the filled/empty boundary. Worth eyeballing a target at ~30% HP before merge; leaving bar troughs flat and keeping dust on chrome panels would be the safe call.

### 6. Low: worth a mobile check on the tapered top vs the close button

`clip-path` clips hit-testing, not just paint. On desktop the close button clears the 20px chamfer comfortably (measured off `after-window-character.png`), so this is fine there. On mobile the `.x-btn` grows to the 40x40 floor and its top edge may reach into the chamfer; if so a corner of the button becomes both cut and untappable. Quick tap test on a phone, or the `test:browser` target-size check, would settle it.

---

Requesting changes on (1) and (2); (2) is a one-token fix and (1) has a ready precedent in the same file. The rest are non-blocking. Nice module, and thanks for the honest scope-and-follow-ups section in the description, including the already-disclosed `box-shadow` clipping.

jgyy:
this PR will need to be thrashed again, but i wanted to keep ornament_svg.ts and ornament_svg.test.ts scripts for future reference for new PR instead, just that script is unused.

jgyy:
Thanks for the thorough review. Pushed a fix for the blocking item and confirmed the second finding is already moot:

**1. Blocking (fixed): ornament scrolling away inside overflowing windows.** Confirmed the root cause and the fix live in a browser before committing (getBoundingClientRect probes on a scroll container + pseudo-element): an `absolute; inset: 0` descendant of `.window`'s own `overflow-y: auto` box really does slide off with the content, exactly as described. Fix: `.window.panel::before`/`::after` now use `position: fixed`, sized from new `--win-frame-*` custom properties that a small new module (`src/ui/window_ornament_frame.ts`) keeps synced to the window's real viewport rect on open, drag-commit, resize-commit, and viewport resize (never per-scroll, so it stays outside the per-frame perf budget). `fixed` only escapes `.window`'s own scroll region once `.window` stops being a containing block for fixed descendants, which its centering `transform` would otherwise create, so the same sync also pins the window to explicit left/top the way `window_resize.ts`/`window_drag.ts` already do on the first drag or resize. Verified empirically: with the old `absolute`/`inset:0` behavior a scroll of 500px shifted a test marker by exactly 500px (reproducing the bug); with the fix applied the marker stayed pinned to the viewport through the same scroll.

**2. High (moot): taper-gem `mask-position` offset.** No longer applicable: round 6 (already on this branch) removed the tapered/chamfered window top entirely, so there's no `--window-taper` reference left anywhere in the tree for that offset to apply to.

Didn't get to 3-6 (aspect-ratio contract pin, low-tier fx drop, bar-trough dust, mobile close-button hit-test) in this pass; happy to follow up if you'd like those addressed before merge.

Added `tests/window_ornament_frame.test.ts` for the new sync module. `tsc --noEmit`, `tests/ornament_svg.test.ts`, `architecture`, `css_corpus`, `css_value_validity`, `styles_extraction`, `client_shell`, `hud_perf_budget`, `mobile_window_coverage`, `backdrop_filter_survival`, `focus_visible_guard`, `theme`, `ui_effects_wiring`, `window_resize`, and the hud-adjacent suites all pass.

jgyy:
Follow-up on EnriqueGF's remaining findings 3-6 from the round-7 review:

**4. Medium (fixed): no low-tier drop for the ornament.** Added a `:root[data-fx-level="low"]` block in tokens.css that resets every `--ornament-*` custom property back to its transparent/no-op pre-boot fallback value, the same mechanism the existing glass drop already uses. One rule covers every consumer site at once (no per-component override needed), pinned in `tests/ui_effects_wiring.test.ts`.

**3. Medium (comment fixed, values left alone): the 96:10 aspect-ratio claim.** Confirmed none of the per-component `mask-size` overrides in hud.css actually hit 9.6, but that was never a real bug: `mask-size` scales each axis of a `mask-image` independently, so a non-96:10 override does not distort the tile, it just stretches or compresses the wobble period along that axis, which several components do on purpose for a shorter edge or finer texture. Reworded the comments in `ornament_svg.ts` and `base.css` to describe that instead of a ratio constraint nothing was enforcing. Left the actual override values untouched since I can't verify a visual rescale without a browser in this pass.

**5/6 (dust-on-bar-trough eyeball, mobile close-button hit-test):** left as the follow-ups you flagged them as; both are non-blocking and need a live render to judge rather than a source change.

Also merged upstream `release/v0.28.0` into the branch (it had drifted ~100 commits behind), which fixes the two failing PR gate checks unrelated to this diff: the missing `scripts/pr_comment_shots.mjs` and a chronomancy test value retuned upstream.

`tsc --noEmit`, `tests/ui_effects_wiring.test.ts`, `tests/ui_effects_profile.test.ts`, `tests/ornament_svg.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, and `tests/styles_extraction.test.ts` all pass; biome clean on the changed files.

jgyy:
Addressed the remaining open item from EnriqueGF's latest review round (round 7).

Verified current HEAD against all six findings before making changes:

1. Ornament scrolling away inside scrolling windows: already fixed (window_ornament_frame.ts pins the frame via --win-frame-* custom properties synced on open/drag/resize, position: fixed instead of absolute inside the scroll container).
2. Top-right taper gem offset: moot, the tapered/chamfered window top was removed entirely in an earlier round (windows are plain rectangles again per direct feedback), so the mask-position math this finding was about no longer exists.
3. mask-size aspect ratio contract: already fixed, the comment was reworded to reflect that mask-size scales each axis independently, so a shorter edge legitimately picks its own footprint; this was never a real bug.
4. No low-tier drop for the ornament: already fixed, every --ornament-* custom property resets to its no-op pre-boot value under :root[data-fx-level="low"], pinned in tests/ui_effects_wiring.test.ts.
5. Gilt dust on the health/resource bar trough: fixed in this pass. Removed --ornament-dust from the base .bar rule and the raid-style party-frame HP bar rule in src/styles/hud.css, since .bar-fill only covers the filled portion and the dust was landing specifically on the empty part of an HP/mana/rage/energy bar, softening the filled/empty boundary. Target-HP granularity is actionable information a player reacts to, so both troughs now stay flat (the subtler grain overlay texture is unaffected). Added a source-text guard in tests/ui_effects_wiring.test.ts pinning both rules.
6. Mobile close button vs the chamfer: moot for the same reason as (2), no chamfer/clip-path exists anymore.

Ran locally: npx tsc --noEmit (clean), the full targeted suite (ui_effects_wiring, ornament_svg, css_corpus, css_value_validity, styles_extraction, mobile_window_coverage, backdrop_filter_survival, focus_visible_guard, theme, client_shell, architecture: 215 tests passed), and biome check --write on the two changed files (no fixes needed).

No visual regression expected from this change (the trough color/texture change is a subtle removal of dust noise on bars only, not a layout or geometry change), so no new screenshots were captured.

CI on this push is green: Screenshots of changes, both English-only legal gate shards shown, Browser regressions (Chromium), PR checks (freshness/typecheck/builds), and Format + lint all pass.

PR #2153, opened 2026-07-19
feat(ui): class-colored chat names + verified-streamer chat badge

jgyy:
## Summary

- Chat sender names now color by class, mirroring the party-frame/minimap
  treatment (`classCss`): whenever a sender resolves to a live player entity
  in interest scope, the name is drawn in that class's color instead of the
  channel's default line color.
- Verified streamers additionally draw a small brand-mark flair badge beside
  their name (`chatStreamerBadgeEl`), placed at the `{name}` slot alongside
  the existing `[AI]` disclosure tag. The badge is purely decorative
  (`role="img"`, no click/tap binding of its own): the sender name sitting
  right beside it already opens the player menu that lists the streamer's
  channel link(s) up top, so no new `href` handling or `window.open` call is
  introduced, and a second, tiny (12x12) tap target with no keyboard path
  is not added.
- Both are additive, cosmetic-only surfaces: no gameplay-actionable
  information changes, and a sender with no streamer link keeps today's exact
  rendering. **Class color is client-side only for this change**: it resolves
  from `this.sim.entities.get(fromPid)`, which is the whole world offline but
  only the ~120yd interest-scoped mirror online (`ClientWorld.entities`). So
  online, in the global channels (general/world/guild/lfg/whisper) where the
  color is most useful, a sender outside your interest scope renders in the
  plain channel color, not a missing edge case, but the common case there.
  Carrying the sender's class through the server-attached chat flair (the way
  `ChatSenderFlair.links`/`.ai` already do, name-keyed via
  `streamerLinksFor`/`accountFlair`) would fix this properly; that is a larger
  cross-host change deferred to a follow-up rather than folded into this PR.

## Related issues

N/A

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/chat_line.test.ts` (new placement/accessibility/no-markup tests for the badge, alongside the existing `[AI]` tag tests)
  - `npx vitest run tests/deed_i18n.test.ts` (the pinned chat-sender-source window this change shares with)
  - `npx vitest run tests/i18n_completeness.test.ts tests/localization_fixes.test.ts`
  - `npx vitest run tests/architecture.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/client_shell.test.ts`
  - `npx tsc --noEmit`
  - `npm test` (full suite; green except two known-flaky tests under CPU contention, both unrelated to this change and confirmed to pass in isolation: `tests/fixes.test.ts`'s Varkas Boneguards despawn timer and `tests/loot_ffa_sim.test.ts`'s determinism check)
- Manual steps:
  - Drove the real dispatch path (`hud.handleEvents` with a synthesized `chat` `SimEvent`, `fromPid` anchored on the live player entity, and a fabricated streamer `flair.links`) in the offline client to confirm the badge and class color render correctly and that clicking either opens the player menu.
  - Captured before/after screenshots (below) on desktop and on a mobile viewport with the chat overlay open, using a mage character so the class-color diff reads clearly against the plain white default.

## Screenshots / recordings

**Desktop**

Before:
before

After (mage name colors bright blue; Twitch flair badge appears):
after

**Mobile** (chat overlay open)

Before:
before

After:
after

---

### Quality

- [x] **The full gate passes.** `npx tsc --noEmit` is clean; the targeted and full test
      suites pass (see above); pre-push QA floor (tsc, guard tests, biome, copy scan) is
      green.

### Cross-platform

- [x] **Tested on desktop and mobile.** See screenshots above; the badge is a small
      inline icon glued to existing inline chat text, decorative like the `[AI]` tag
      beside it rather than a second tap target.
- [x] **Accessible.** The badge carries `role="img"` + `aria-label` (translated,
      non-Latin locales filled) naming its meaning, matching the `[AI]` tag's
      disclosure pattern. It is not independently interactive, so it has no keyboard
      trap: the sender name beside it (`role="button"`, `tabIndex=0`, Enter/Space)
      is the one focusable, actionable element and opens the same menu.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Added
      `hudChrome.playerMenu.streamerBadgeTitle` to the English catalog; it triggered the
      M16 wordy-copy rule, so I also filled the five required non-Latin locales
      (`zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`) in this same change.
- [x] Numbers, money, dates, units, and percents go through the i18n formatters. N/A,
      this change adds no such values.
- [x] Player-facing text emitted from `src/sim/`/`server/`. N/A, no sim/server text
      changed.

### Hygiene

- [x] No generated files hand-edited; `src/ui/i18n.resolved.generated/` and
      `src/ui/i18n.catalog/translation_keys.generated.ts` were regenerated via
      `npm run i18n:gen`.

---

## Review follow-up (EnriqueGF, CHANGES_REQUESTED)

- **Badge keyboard reachability / static-image mismatch:** made the badge purely
  decorative (dropped the `bindTouchTap` binding in `hud.ts`); `role="img"` is now
  accurate and there is no second, non-keyboard-reachable interactive element.
- **12px touch target:** moot once the badge is decorative; the sender name (already
  a proper `>=`24px-tall text button with a keyboard handler) is the one tap target.
- **Interest-scope gating of class color:** confirmed and called out explicitly above
  instead of glossed over; the client-only version is shipped now, a name-keyed
  server-attached-flair fix is the natural follow-up (same shape as the existing
  `links`/`ai` fields).
- **Inline `color` vs custom property:** `sender.style.color` is now
  `sender.style.setProperty('--class-color', ...)`, with `.chat-player-name { color:
  var(--class-color, inherit); }` owning the paint (matches
  `party_frames_painter.ts` / `char_skin_window.ts`).
- **No forced-colors fallback on the badge:** added one, matching `.ai-tag`'s pattern
  (`color: CanvasText` under `forced-colors: active`).


EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants, and re-ran the checks locally in a clean worktree off `release/v0.28.0`.

**What I ran**
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/chat_line.test.ts tests/architecture.test.ts tests/localization_fixes.test.ts tests/i18n_completeness.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/world_api_parity.test.ts`: all green (the S3 guard needs `npm run i18n:gen` first, as documented).
- `npx vitest run tests/account_flair.test.ts tests/account_flair_client.test.ts tests/chat_flair.test.ts tests/chat_context_menu.test.ts tests/chat_log.test.ts tests/chat.test.ts`: green. I saw the two `/played` / `/playtime` failures under parallel load and confirmed they are contention flakes, not a regression: `tests/chat.test.ts` passes 72/72 in isolation on both this branch and the base commit.
- Biome on the changed files: no errors, no format diffs (only pre-existing warnings elsewhere in `hud.ts`).

**What is solid.** The security posture is right: `chatStreamerBadgeEl` goes through `streamerLinkList`, so every link is re-normalized at render time and the badge never touches an href or `window.open`. The `innerHTML = svgIcon(platform)` exception is honest and currently not injectable (the platform id comes from the sanitized allowlist and `ICONS[name]` is a fixed literal). Placement at the `{name}` slot is correct and well tested, the CSS lands in `@layer components` next to the existing stream-link rows and uses `var(--color-accent)` rather than a brand hex, and the i18n work follows the contributor policy (English in the catalog, M16 non-Latin fills, generated files regenerated rather than hand-edited). The `tid` wire field does carry the class id in both hosts, so `classCss` never silently falls back to default blue for a real player.

Three findings, roughly in severity order.

---

**1. The badge is interactive but exposed as a static image, and is not keyboard reachable.** `src/ui/hud.ts:9917-9918` binds `bindTouchTap(streamerBadge, openUnderName)` and the CSS sets `cursor: pointer`, but `chatStreamerBadgeEl` gives it `role="img"` with no `tabIndex`, no `keydown` handler, and an `aria-label` ("Verified streamer") that names the thing rather than the action. The sender name sitting right next to it does this correctly: `role="button"`, `tabIndex = 0`, Enter/Space handler, `playerMenu.openFor` label. So the PR checklist line "reachable via the same click/tap/keyboard affordance as the sender name" does not hold for the keyboard arm.

My suggestion is the simpler of the two fixes: **drop the tap binding and leave it a purely decorative mark.** The name immediately beside it already opens the identical menu, so the click handler buys nothing, and `role="img"` then becomes accurate. The alternative (promote to `role="button"` + `tabIndex=0` + keydown + an action-shaped label) adds a second tab stop per chat line for no new destination.

**2. Related, and the reason I lean toward making it decorative: the touch target is 12px.** `.streamer-badge` is an `inline-flex` span with no padding wrapping a 12x12 icon, and there is no `hud.mobile.css` override (I grepped: `.streamer-badge` appears only in `hud.css`). `src/ui/CLAUDE.md:38-40` sets a >=40x40 floor on mobile touch with 24x24 as the absolute WCAG 2.2 minimum and says explicitly not to weaken the 40 to 24. At 12px this is under even the absolute minimum. If the badge stops being tappable this dissolves entirely; if it stays tappable it needs a >=40x40 hit area under `body.mobile-touch`.

**3. The class color is interest-scope gated, so it mostly does not appear online in the channels it is for.** `src/ui/hud.ts:9887-9888` resolves the class from `this.sim.entities.get(fromPid)`. Offline that map is the whole world, so every sender colors. Online, `ClientWorld.entities` holds only interest-scoped entities and actively reaps the rest (`src/net/online.ts:2545-2563`), so a general / world / guild / lfg / whisper sender outside ~120 yd has no mirrored entity and renders in the plain channel color. Those global channels are exactly where a class color is most useful, and they are the ones that lose it. It also means the color flickers on and off as a sender walks in and out of scope.

This degrades gracefully rather than rendering something wrong, so it is not a correctness bug, but it is worth being deliberate about, for two reasons:

- The existing flair code three thousand lines up already solved this shape and the new line does not follow it: `streamerLinksFor` (`src/ui/hud.ts:11888`) is name-keyed with an entity fallback (`this.sim.accountFlair(name)?.links ?? ent?.streamerLinks`), precisely so it resolves for someone who only ever spoke in chat. There is no by-name class lookup on `IWorld` today, so a real fix means carrying the sender class on the chat event or mirroring it the way the flair cache is mirrored.
- There is a small fairness angle. Once color is proximity-gated, a colored name in General chat is a signal that the speaker is within your interest scope and an uncolored one that they are not. That is a positional hint chat did not previously give. It is minor, and much weaker than a nameplate, but "gameplay-neutral presentation" is a load-bearing invariant here so I would rather raise it than let it pass silently.

If you would prefer to ship the client-only version now, that is defensible, but please say so in the PR body rather than leaving "a sender outside interest scope keeps today's exact rendering" as the framing, since that sentence reads as an edge case when online it is the common case.

---

Two smaller notes, take or leave:

- `sender.style.color = classCss(...)` sets an inline color, where the established pattern in this repo stamps a custom property and lets CSS own the paint (`src/ui/party_frames_painter.ts:361`, `src/ui/char_skin_window.ts:56`). An inline `color` cannot be overridden later by a `forced-colors` block or a future "disable class colors" option. `setProperty('--class-color', ...)` plus a rule on `.chat-player-name` would cost the same and keep that door open.
- `.ai-tag` carries an explicit `@media (forced-colors: active)` fallback; the badge has none. `currentColor` under a forced palette should resolve to the forced text color and survive, so I expect this is fine, but it is unverified.

Happy to re-review quickly once 1 and 2 are addressed. The security and i18n work here is genuinely careful and I do not want the a11y nits to obscure that.

EnriqueGF (changes requested):
Reviewed the full diff, ran the checks locally, and dug into the two hosts. Nice, tightly-scoped feature and the security story on the badge is genuinely well done. Two things need a change before this can go in, both about *which player* a mark ends up describing.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/chat_line.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/i18n_completeness.test.ts`, `tests/localization_fixes.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/mobile_window_coverage.test.ts`, `tests/client_shell.test.ts`, `tests/deed_i18n.test.ts`: all green.
- `npm run i18n:gen` produced **no diff**, so the generated bundles committed here are fresh. `pending.ts` correctly lists the new key for the 15 Latin locales and excludes `zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`, matching the M16 fill you describe.
- Copy scan on added lines: no em/en dashes, no emojis.

## What checks out

The `innerHTML` was the first thing I went after, and it is safe: `first.platform` comes from `streamerLinkList()`, which walks the fixed `STREAMER_PLATFORMS` tuple and re-runs `normalizeStreamerLink` per entry, so the value reaching `svgIcon` is always one of four literals indexing a static `ICONS` table, and no URL is ever interpolated. `StreamerPlatform` being a subset of `UiIconName` means `tsc` backs that up. The test pinning `expect(el.innerHTML).not.toContain('twitch.tv/zyx')` is the right pin.

Also confirmed clean: `templateId` really is the class id for players on both hosts (`src/sim/types.ts:2582`, and online it is populated from the `tid` wire field); entity ids are monotonic and never recycled, so no id-collision risk, and the `kind === 'player'` guard correctly skips mob and boss yells. The CSS lands inside the single `@layer components` block next to the existing `.ai-tag` and ctx-menu icon rules.

## Blocking

**1. The "To {name}" whisper echo attributes the sender's marks to the recipient.** `src/ui/hud.ts:8929-8937` displays `ev.to` (the recipient) but passes `ev.fromPid`, which the emitter sets to the *sender* (`src/sim/social/chat.ts:715-724`, `fromPid: r.meta.entityId`). So with this change, your own outgoing whisper paints **the recipient's name in your class color**, and if you are a flagged streamer, draws **your verified badge next to their name**. The comment two lines above already establishes exactly this rule for `fromTitle`: "the sender's fromTitle must never decorate it (untitled beats mislabeled)."

To be precise about blame: the `ev.flair` pass-through predates this PR, so the badge case widens an existing latent bug (previously only an `[AI]` account could trigger it). The class-color case is new here. Both want the same fix: withhold `fromPid` and `flair` on the `ev.to` branch, the way `fromTitle` already is. Worth a `chat_line.test.ts` case pinning it, since nothing covers this path today.

**2. Class coloring resolves through `IWorld.entities`, which is interest-scoped online but world-complete offline.** `src/ui/hud.ts:9887-9888` looks the sender up in the entity map. Offline `Sim` holds every entity, so every name colors. Online, `ClientWorld` prunes to interest scope (`src/net/online.ts:2533-2562`), so the practical result is:

| Channel | Colored offline | Colored online |
|---|---|---|
| say / yell | yes | yes |
| party | yes | only when the member is nearby |
| general / world / lfg / whisper | yes | almost never |
| guild / officer | yes | **never** (`SocialEvent`'s chat variant has no `fromPid` at all, `server/social.ts:206-213`) |

Beyond the host divergence, the in-client effect is the one I would flag even if there were only one host: the same speaker's name flips between class color and plain channel color as they walk across the ~120yd boundary mid-conversation.

The codebase already hit this exact hazard for flair and solved it deliberately. `src/sim/types.ts:3145-3151` says flair "rides the event rather than being read off the sender's entity because general/world/lfg/guild chat reaches you from players far outside your ~120yd interest scope, where no entity record exists locally." The chat event is already the carrier for per-sender identity (`fromTitle`, `flair`), so stamping the class there at fan-out is the cheap, consistent fix, and it makes guild/officer work too.

I recognise the PR body frames out-of-scope senders keeping today's rendering as intentional. I would push back: on the online realm, which is the host that matters for a streamer badge, that scoping means the feature mostly does not fire.

## Non-blocking

- **Badge a11y.** It is `role="img"` with no `tabIndex` and no keydown, but it is given `cursor: pointer` and a `bindTouchTap`. A screen reader announces it as a static image with no way to activate it, and the checklist's "reachable via the same click/tap/keyboard affordance as the sender name" is not quite accurate. Either mirror the sender (`role="button"`, `tabIndex = 0`, Enter/Space) or make it purely decorative and drop the tap binding plus `cursor: pointer`. Decorative is defensible: the name beside it already opens the same menu.
- **Touch target.** At 12px with no hit padding it is well under the touch floor. Resolved for free if you go the decorative route.
- **Cursor token.** `.streamer-badge` uses `cursor: pointer` while the `.chat-player-name` it sits against uses `cursor: var(--cursor-point)` (`hud.css:2105`), so the same action shows two different cursors an inch apart.
- **Scope the margin.** `.ai-tag` deliberately scopes its identical `margin-right: 3px` to `#chatlog .ai-tag` because nameplate and unit-frame flair hosts bring their own spacing. `.streamer-badge` bakes it into the unscoped class, which will double-space if the badge ever joins those surfaces (the AI mark already has).
- **Normalize at the boundary.** `chatStreamerBadgeEl(document, flair?.links)` takes the wire object raw, while the entity path re-sanitizes with `normalizeStreamerLinks` under a "NEVER trust the wire" comment (`src/net/online.ts:2073-2079`). It is safe today because `streamerLinkList` re-validates, but the safety rests on the callee rather than the boundary rule.
- **Module-first.** The badge builder correctly went to the domain module and got tests; the class-color resolution is the one piece of new logic that landed on `hud.ts` untested, and it needs none of `Hud`'s private state.
- **`scripts/pr_screenshots.mjs`.** The viewport clamp is right for `getBoundingClientRect` coords, but when the intersection is empty the `Math.max(1, ...)` silently writes a ~1px image instead of pushing to `errors`. Fine for the current in-viewport HUD targets, just fails quietly.

Happy to re-review quickly once the whisper echo and the class-source question are settled.

jgyy:
Thanks for the thorough review. Pushed 2e6686f12 addressing all three findings:

1. **Badge keyboard reachability / `role="img"` mismatch.** Took the simpler fix you suggested: dropped the `bindTouchTap` binding on the badge in `hud.ts` so it's purely decorative, matching `role="img"`. The sender name beside it already opens the identical player menu and already has the correct interactive treatment (`role="button"`, `tabIndex=0`, Enter/Space), so nothing is lost.

2. **12px touch target.** Moot now that the badge is decorative; there is no longer a second tap target under the mobile/WCAG floor.

3. **Interest-scope gating of class color online.** Agreed this deserved more than the framing it had. Updated the PR body to state plainly that class color is client-only for now and that the global channels are the common case that loses it, not an edge case, plus that a real fix means carrying the sender's class through the server-attached chat flair the same way `links`/`ai` already ride it. Shipping the client-only version now per your "defensible" option, with that follow-up called out rather than glossed over.

Also took both smaller notes:
- `sender.style.color` -> `sender.style.setProperty('--class-color', ...)` with `.chat-player-name { color: var(--class-color, inherit); }` owning the paint, matching `party_frames_painter.ts` / `char_skin_window.ts`.
- Added an explicit `forced-colors: active` fallback (`color: CanvasText`) on `.streamer-badge`, matching `.ai-tag`'s pattern.

Re-ran `npx tsc --noEmit`, `tests/chat_line.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/i18n_completeness.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/world_api_parity.test.ts`, and the flair/chat suites: all green (the two `/played`/`/playtime` contention flakes you already flagged are still present under parallel load and still pass in isolation, unrelated to this change). Biome on the two changed files: no errors, no format diffs.

jgyy:
Thanks for the thorough second pass, and for tracing both of these all the way to their root cause rather than stopping at the symptom.

Fixed both blocking findings:

- **Whisper echo attribution.** `src/ui/hud.ts`'s `'whisper'` case now calls `chatLogFrom` with only `(ev.to, ev.text, CHAT_TEMPLATE_KEYS.toWhisper, 'whisper')` on the `ev.to` branch, withholding `fromPid`, `flair`, and the new `classId` exactly the way `fromTitle` already was. Your framing was right: the fix belongs next to the existing `fromTitle` carve-out, not as a separate special case.
- **Class color interest-scope gating.** `classId` now rides the chat event itself, stamped at the sim emit site in `src/sim/social/chat.ts` (a `speakerClass` helper mirroring the existing `speakerTitle`) and on the guild/officer relay frame in `server/social.ts` (via `SocialActor.cls`, read from the live sim meta in `server/game.ts`'s `actorFor`, the same path `activeTitle` already takes). `hud.ts` now reads `classId` off the event instead of looking the sender up in `this.sim.entities`, so general/world/guild/lfg/whisper senders outside interest scope color correctly online, and a nearby sender no longer flickers as they cross the ~120yd boundary.

Also added regression coverage: `tests/chat.test.ts` extends the existing `fromTitle` test suite with `classId` assertions (including the whisper echo case, pinning that the sender's own class never lands on the recipient's name), and `tests/social_system.test.ts` adds the same coverage for the guild/officer relay. `tests/deed_i18n.test.ts`'s pinned `chatLogFrom` window is widened to fit the new doc comment.

One honest gap: there's no existing DOM harness in this repo that constructs a full `Hud` instance, so the regression tests above pin the *data* fix (the event now carries the right class, and withholds it on the right branch) rather than the rendered DOM output itself. Happy to build that harness in a follow-up if you'd like tighter coverage on the render side specifically.

I left the five non-blocking items from this review (cursor token mismatch, badge margin scoping, wire-boundary normalization, the a11y/touch-target notes already resolved by your first review, and the `pr_screenshots.mjs` silent-clamp) as-is for now to keep this diff scoped to the two blockers; glad to pick any of them up if you'd rather they land here too.

EnriqueGF (approved):
Reviewed the full diff against `release/v0.28.0` and ran the checks locally. This is in good shape: the design is right, the seams are respected, and the follow-up commits addressed the earlier round properly. Approving with three small notes, none blocking.

## What I checked

- **Determinism / sim purity.** `speakerClass` is a pure read of `PlayerMeta.cls`, no rng, no clock, no new draws, no tick-order change. `src/sim/social/chat.ts` gains no imports outside `sim/`. `tests/architecture.test.ts` green.
- **Tri-host parity.** Traced `classId` end to end. Sim emit (13 sites) to `GameServer.routeEvents` (`server/game.ts:5730`, pushes the event object by reference and sends it verbatim at `:5834`, no field allowlist) to `ClientWorld` (`src/net/online.ts:1864`, queues the event as-is) to the HUD. The guild/officer relay is covered separately: both `SocialEvent` chat construction sites (`server/social.ts:892`, `:962`) stamp it and `deliver()` (`server/game.ts:1645`) passes the list through whole. All 11 HUD channel branches forward it. No `IWorld` surface change, so no parity pin to update. `tests/world_api_parity.test.ts` green.
- **The falsy guard.** `...(actor.cls ? { classId: actor.cls } : {})` is safe: `PlayerClass` is nine non-empty string literals, so the key drops only when `actorFor` found no live sim meta, exactly the pre-existing `activeTitle` behavior.
- **Security.** `chatStreamerBadgeEl` uses `innerHTML`, but the only interpolated value is `first.platform` from the fixed `STREAMER_PLATFORMS` union, and the URL itself never reaches the markup (nicely pinned by the `not.toContain('twitch.tv/zyx')` test). `streamerLinkList` re-runs `normalizeStreamerLink` at render time, so the `javascript:` case is dropped, also tested. No `window.open`, no href.
- **i18n.** One new English key in the catalog plus the five non-Latin fills M16 requires; no hand-edits to generated files (`npm run i18n:gen` left the tree clean here). `tests/i18n_completeness.test.ts` and `tests/localization_fixes.test.ts` green.
- **Styles / a11y.** `--class-color` plus `.chat-player-name { color: var(--class-color, inherit) }` matches the `party_frames_painter` / `char_skin_window` pattern. The badge is decorative with `role="img"` and a translated `aria-label`; the inner `svgIcon` is `aria-hidden`, so no double announcement. Forced-colors fallback present.
- **Tests run:** `npx tsc --noEmit` (clean), `tests/chat_line.test.ts`, `tests/chat.test.ts`, `tests/social_system.test.ts`, `tests/deed_i18n.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/i18n_completeness.test.ts`, `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/mobile_window_coverage.test.ts`. All green. Biome on the changed files: warnings only, all pre-existing and unrelated.
  - I did see `tests/chat.test.ts` `/played` and `/playtime` time out at 5000ms on one run, but only while another test process was competing for cores; three subsequent clean runs pass 72/72 on this branch, and the base does not differ. Environmental, not this change.

## Notes

**1. The whisper echo drops `ev.fromPid`, which was not a sender-attribution mark (low).**

`src/ui/hud.ts:8933`. The comment groups `fromPid` with `fromTitle`/`flair`/`classId` as a "per-sender mark", but `fromPid` never touches the sender span: `chatLogFrom` only forwards it to `appendChatMessageBody` (`:9939`), where it binds the quest-link *author* for `openLinkedQuestDialog`. Concretely: whisper someone a `[[q:id]]` link and click the link in your own "To Bet:" echo. Before, `inSharerParty` was true (you are a member of your own party) and you got the Accept button; now `fromPid` is `undefined` and you get the `questShare.viewOnlyHint` instead. Arguably the new behavior is the more sensible one, so I am fine either way, but it looks like collateral rather than a decision. Worth either keeping `ev.fromPid` on that call or correcting the comment to say the drop is deliberate and why.

**2. The PR description is stale (info).**

The Summary still says class color "is client-side only for this change", resolves from `this.sim.entities.get(fromPid)`, and defers carrying the class through the event to a follow-up. Commit 797a31e3c actually did the server-carried version, which is why I could confirm parity above. Please update the body before merge so it does not mislead anyone reading the history later.

**3. Small gaps, no action needed now (info).**

- Nothing tests the `--class-color` stamp itself; the coverage stops at the sim event (`tests/chat.test.ts`) and the relay (`tests/social_system.test.ts`). That is the usual `chatLogFrom`-lives-in-the-coordinator problem, not something to fix here.
- `server/game.ts:1631`, the presence adapter's leaner `actor()` builder, populates neither `activeTitle` nor `cls`. Harmless today since it never feeds a chat event, but it is the one place a future chat-adjacent relay would silently lose both.

EnriqueGF (approved):
Reviewed the full diff and re-ran the gates locally on the head commit (`7d3d739`).

**What I checked**

- `npx tsc --noEmit`: clean.
- `npm run i18n:gen` then `npx vitest run tests/localization_fixes.test.ts`: green (29 passed). `tests/i18n_completeness.test.ts`, `tests/architecture.test.ts`, `tests/chat.test.ts`, `tests/chat_line.test.ts`, `tests/social_system.test.ts`, `tests/deed_i18n.test.ts`: green.
- `npx vitest run tests/parity`: green (180 passed), and the regenerated `chat_social.json` golden only moves the `events` digest; `state` and the rng digest/draw count are untouched, which is exactly what an additive optional event field should do.
- `tests/css_corpus.test.ts`, `tests/css_value_validity.test.ts`, `tests/mobile_window_coverage.test.ts`: green.
- Biome on the changed source files: warnings only, all pre-existing (an unused private method in `hud.ts`), no errors or format diffs.

**Invariants**

- Determinism and sim purity hold: `speakerClass` is a pure spread off `PlayerMeta.cls` (non-optional on the type), no new rng draw, no clock, no DOM. Draw order is unchanged.
- Carrying `classId` on the event instead of reading `IWorld.entities` at render time is the right call and matches the existing `fromTitle`/`flair` precedent. It also means no new `IWorld` member, so the parity pin correctly did not need to move. I confirmed the online path end to end: `routeEvents` forwards the whole `SimEvent`, and the guild/officer relay reaches the client through the same `{t:'events'}` frame from `deliver`, so `classId` survives on both hosts.
- Server authority is untouched (cosmetic only), no `ALLOW_DEV_COMMANDS` or secrets, no em/en dashes, no emojis.
- i18n: English-only in the catalog plus the five non-Latin fills is exactly the M16 exception in `src/ui/CLAUDE.md`, so the overlay edits are sanctioned here rather than a violation of the "never edit overlays" rule. `t()` is used for both the badge's `title` and its `aria-label`.
- Graphics fairness: nothing actionable is added or hidden.
- The `chatStreamerBadgeEl` `innerHTML` is a fixed literal from the icon registry after `streamerLinkList` revalidation, and the test pins that the URL never reaches markup. Fine.

**Findings (both minor, neither blocking)**

1. `--class-color` is an inheriting custom property that already exists globally (`src/styles/shell.css` `.mini-class[data-class=...]`, `src/main.ts:3427` and `:5031`). Today every writer is scoped to character-create UI, so nothing that is an ancestor of `#chatlog` sets it and the `var(--class-color, inherit)` fallback behaves as documented. But the contract "no `classId` means fall back to the channel color" is only true as long as that stays accidentally so. A scoped name such as `--chat-class-color` would make it structurally true instead of coincidentally true.

2. The `toWhisper` arm now drops `fromPid` along with the sender marks. `fromPid` does not only feed the name decoration: it also flows into `appendChatMessageBody` and on into `openLinkedQuestDialog(questId, fromPid)`, where `inSharerParty` gates the Accept button. Concretely: whisper a quest link to someone, click the link in your own "To {name}" echo, and where you previously got an Accept button you now get the view-only hint. I think the new behavior is the more correct one (accepting a quest shared by yourself was odd), but it is a side effect the comment and the tests both attribute purely to name decoration. Worth either a sentence in that comment or a pin in `tests/deed_i18n.test.ts`'s `toWhisper` window so the next reader does not "restore" `fromPid` as an apparent oversight.

The earlier CHANGES_REQUESTED items all look properly addressed: the badge is genuinely decorative now (so `role="img"` is accurate and there is no keyboard-unreachable target), the paint moved to a custom property, forced-colors has a fallback, and the interest-scope limitation is stated plainly instead of glossed. Approving; item 2 is worth a follow-up line, item 1 is take-it-or-leave-it.

jgyy:
Rebased onto the latest `release/v0.28.0` tip (the branch had drifted to CONFLICTING). Only conflict was the generated `i18n.resolved.generated/pending.ts`, resolved by regenerating via `npm run i18n:gen` rather than hand-editing. tsc, parity/architecture/i18n guard tests, and the chat/social test suites are all green.

PR #2155, opened 2026-07-19
feat(hud): restore the third action bar

Blaine1705:
## Summary

Restores the third action bar from the earlier PBE work on top of `release/v0.28.0`.

- Adds an optional third 11-slot desktop row, gated by the secondary action bar setting.
- Adds `Shift+Numpad1..9`, `Shift+Numpad0`, and `Shift+NumpadDecimal` defaults and exposes all new binds in Key Bindings.
- Expands hotbar, form-bar, talent-loadout, spellbook, and mobile-ring behavior to 33 configurable slots while preserving legacy 22-slot saves and loadouts.
- Keeps mobile at seven pages and hides the two unavailable positions on page 7.
- Adds regression coverage, accessibility coverage, translations, generated locale artifacts, and visual evidence.

## Type of change

- [x] Feature
- [x] Tests
- [x] Documentation

## How tested

- Focused Vitest suite: 19 files passed; 483 tests passed, 6 skipped.
- `npm run test:browser -- tests/browser/a11y.browser.test.ts`: 23 passed.
- `npx tsc --noEmit`: passed.
- `npm run build`: passed.
- Pre-push QA floor: 54 passed, 3 skipped; Biome completed without errors.
- `npm run gate`: i18n freshness, malware scan, Biome, and SFX checks passed. The full Vitest phase then hit 44 existing Windows/POSIX-environment failures (missing bash/sh, unsupported POSIX signals/permissions/symlinks, and POSIX path assumptions); CI Linux is required for those platform-specific cases.

Manual coverage:

- Mage at level 20 with two and three desktop rows.
- Interface toggle dependency and direct enable/disable behavior.
- Mobile page 7/7 with exactly the three valid final slots.

### Before: two action bars

Before: two action bars

### After: three action bars

After: three action bars

### Interface toggle

Interface toggle

### Mobile page seven

Mobile page seven

## Checklist

- [x] The change is covered by focused regression tests.
- [x] New user-facing strings are localized.
- [x] Accessibility behavior is covered by browser tests.
- [x] Generated localization files were refreshed through the repository workflow.
- [x] No secrets or unrelated changes are included.
- [ ] The entire canonical gate is green locally (blocked only by the Windows/POSIX cases documented above).


EnriqueGF (approved):
Reviewed the full diff against `release/v0.28.0` and re-verified the load-bearing parts in the code rather than from the description. This looks solid and mergeable. Approving with a few non-blocking notes.

## What I ran locally

- `npx tsc --noEmit`: clean.
- `npx vitest run` over the guard suites plus every test the change touches (architecture, localization_fixes, world_api_parity, action_bar_layout_core, action_bar_visibility_core, action_bar_controller, hotbar, keybinds, loadout_action_bar, mobile_action_page_view, mobile_action_ring_painter, options_view, options_window, settings, spellbook_view, talent_loadouts_v2, client_shell): 644 passed, plus localization_fixes 29 passed / 3 skipped. `localization_fixes` needs `npm run i18n:gen` first (it wants `src/ui/i18n.status.json`, which `pretest` normally produces); after generating, the working tree stayed clean, so the committed resolved bundles are fresh.
- `npx @biomejs/biome check --diagnostic-level=error` on the changed source and CSS: 0 errors.
- Scanned the diff for em/en dashes and emojis: none.

## What I verified by reading the code

- **Loadout back-compat.** `SAVED_LOADOUT_BAR_SLOTS` 22 to 33 is safe on both persistence paths: `repairBar` and `saveLoadout` only `slice(0, N)` (a legacy 22-entry array stays 22 entries), and `migrateLoadoutBar` pads to 33 with nulls. The new `if (i >= bar.length) return current[i] ?? null` in `applyLoadoutBar` is the right call: a legacy loadout has no opinion about row three, so preserving the current action there beats clearing it, and an explicit null inside the saved span still clears as before.
- **Setting dependency at boot.** `resolveActionBarVisibility` is order-independent under the startup apply-all loop in `main.ts` (line 1942): whichever of the two keys is applied first, a persisted `{secondary: false, third: true}` self-heals to `third: false`. Confirmed `settings.set` on a bool key returns the value unchanged, so dropping the old `toggle(class, settings.set(...))` idiom loses nothing.
- **Keybinds.** No collision: `Shift+Numpad*` was unused, and the edge dispatcher compares the full chord string, so bare `Numpad1` still fires only `slot12`. The English `label` strings in `BIND_ACTIONS` ("Third Bar N") are not player-visible: `actionDisplayName` renders slot rows through `hud.keybinds.actions.actionBarSlot` or the bound ability name.
- **Mobile ring, last page.** Hiding the two tail positions is correct and geometry-safe: `.mobile-action-slot` is absolutely positioned per `data-mobile-index`, so `display: none` does not reshuffle the remaining three. Out-of-range source slots 34/35 are inert: `ActionBarController.actionForSlot` ends in `?? null`, `elementFromPoint` never returns a hidden element so the drag hit test cannot resolve them, and `buildMobileActionRing` bails unless exactly five slot buttons exist, so `slots[buttonIndex + 1]` is always present.
- **Pure cores.** Both new modules are DOM-free and registered in the `UI_PURE_CORES` allowlist, and the layout core removes the duplicated boundary arithmetic instead of adding a second copy. Good fit for the seam.
- **i18n.** The catalog gets English only; the `ja_JP`/`ko_KR`/`ru_RU`/`zh_CN`/`zh_TW` overlay edits are the M16 exception for a new wordy English value, so they belong in this change. The remaining locales resolving to English is expected at PR tier and is what the release-tier fill will pick up.

## Notes (non-blocking)

1. `mobile_action_ring_painter.ts` hardcodes `buttonIndex < 5` in the new visibility loop while the module right next to it exports `MOBILE_ACTIONS_PER_PAGE`. Same number today, but this is exactly the duplicated-arithmetic the new layout core exists to avoid. Worth importing the constant.
2. The action bar view now ticks 34 slots every frame regardless of which rows are visible, and `abilityForSlot` does a linear `sim.known.find` per slot. It is 11 extra scans per frame, so almost certainly under the noise floor, but the work happens even for a player who never enables either optional row. If you ever want it back, gating the derivation on the visible row count is the cheap lever.
3. The disabled "Show Third Action Bar" toggle gives no accessible reason for being disabled. The focus restore after the secondary toggle is a nice touch, but a screen reader user just lands on a dead button. An `aria-describedby` hint pointing at the dependency would round it out.

Nothing here blocks the merge.

PR #1553, opened 2026-07-06
feat(woc): paid respec + loadout slots on the #1496 payment layer (implements #472) [draft, stacked on #1496]

awidearray:
## Paid respec + loadout slots (#472)

Burn $WOC to (a) respec a character's talents and (b) unlock additional saved
loadout slots. Loadout slots are the existing saved talent/gear configurations,
swappable only when OUT of combat (already enforced in the sim).

Server-authoritative: payment verification and the state change happen
server-side as a RouteDef-style module. Behind a NEW, independently fail-closed
flag `WOC_PAID_RESPEC_ENABLED` (default OFF), separate from
`WOC_IDENTITY_ENABLED`, so this ships dark on its own.

### STACKED ON #1496, MUST land after it
This branch is based on the #1496 rename/vanity payment branch head and reuses
its payment layer verbatim. It MUST merge after #1496.

### Reused payment layer (no duplication)
- `server/woc_payment.ts` `verifyWocPayment(signature, payer, priceBase, memo)`
  is consumed verbatim in `server/respec.ts` (confirm handler); no
  crypto/signing/verification is reimplemented.
- The two-phase quote/confirm shape and dedupe mirror `server/identity.ts`:
  the memo equals the quoteId, and the shared `woc_payments.tx_sig` UNIQUE row
  (via `recordWocPayment`) rejects any reused signature with 409
  `this payment was already used` (a signature used for ANY purpose, identity or
  respec, can never be redeemed again).
- Pricing/split from `server/woc_config.ts` in the same idiom: new keys
  `WOC_PRICE_RESPEC` (default 750) and `WOC_PRICE_LOADOUT_SLOT` (default 1500),
  routed through `splitPrice`/burn.
- Routes registered in `server/main.ts` via `registerRespecActions(makeRespecActions({...}))`,
  the same pattern as `registerIdentityActions`.

### Sim purity
`src/sim/progression/paid_respec.ts` holds the sim-pure transforms
(`respecCharacterState`, `unlockLoadoutSlot`, `loadoutSlotCap`) a paid action
applies to an OFFLINE character's persisted state. No crypto or network in
`src/sim`; the paid edit runs without a live Sim (like a paid rename edits the
stored row). The offline state write goes through `mutateOfflineCharacterState`
(server-to-sim seam), and combat gating on respec/loadout-swap is the sim's
existing check.

### Loadout-slot cap
A per-character `loadoutSlots` bonus over the free `DEFAULT_LOADOUT_SLOTS`
[...]

PR #1551, opened 2026-07-06
feat(marketplace): premium listing slots with $WOC burn (implements #471) [draft, stacked on #897]

awidearray:
## Premium (featured) listing slots (#471)

Creators pay a fixed fee to feature one of their live listings in a bounded set of PREMIUM SLOTS for a fixed, config-driven window. There is NO auction: the slot count, the window, and the fee are all fixed from config. Featured listings surface FIRST in the browse grid with a localized "Featured" badge. When a slot's window passes it is swept and freed for the next buyer.

### Stacked on #897
This is STACKED ON #897 (creator-skins marketplace) and MUST land after it. It builds directly on that marketplace and reuses its payment machinery; the branch is cut from the current #897 fork head (`awidearray:feature/woc-skin-uploads`, `181d7326e`).

### Currency reconciliation (no duplicate crypto)
The issue language says "$WOC burned". In THIS worktree the #897 marketplace settles in USDC and the "burn" is the 30% burn-vault share of a split transfer (there is no separate `verifyWocPayment` / $WOC path here). This PR reconciles by REUSING the marketplace's existing split-payment + burn-vault path exactly as skin purchases do: the feature fee is a single atomic tx with a 70% creator leg and a 30% burn-vault leg, and the burn share IS the existing burn-vault mechanism (the buy-and-burn keeper drains it separately, out of scope here). No new crypto path was invented.

**Reused payment functions (proving no duplication)** - all imported by `server/premium_listings.ts`:
- `splitAmounts()` - the 70/30 split (`server/marketplace.ts`)
- `createMarketplaceQuote()` - persist the quote, memo = quoteId (`server/db.ts`)
- `parseSplitPayment()` - parse the finalized tx (`server/solana_rpc.ts`)
- `validateSplitPayment()` - the exhaustive accept/reject validator (`server/marketplace.ts`)
- the `onchain_payments.tx_sig` redeem-once dedupe - reused inside the new `claimPremiumSlot()` (`server/db.ts`), the SAME per-signature replay guard `redeemPurchase` uses for skin purchases.

Only the slot ledger, the expiry sweep, and the config are premium-specific.

### Feature flag + env
Behind a NEW flag `PREMIUM_LISTINGS_ENABLED` (env, default OFF, fail-closed), mirroring `marketplaceEnabled` gating: `premiumListingsEnabled()` returns true only when the flag is exactly `1` AND the marketplace is configured (a burn vault exists to receive the 30% leg). New env (documented in `.env.example`):
[...]

PR #2157, opened 2026-07-19
feat: on-chain $WOC burn tracker (burn.html)

awidearray:
## Summary

Adds `burn.html`, a self-contained static page (sibling of `admin.html` and `guide.html`) that tracks $WOC actually destroyed on-chain by the ecosystem's burn mechanics, plus its build and test rig under `tools/burn-tracker/`.

- **Total burned is exact math, not indexer-dependent:** 1,000,000,000 (fixed pump.fun mint, matches `WOC_MAX_SUPPLY` in `src/sim/holder_tier.ts`) minus the mint's live `getTokenSupply`. At build time: 442,071.78 $WOC burned.
- **USD value** uses the same fail-closed price chain as the economy service: Pyth Hermes v2 when `PYTH_WOC_USD_FEED_ID` is configured, otherwise the Jupiter price API for the pinned mint.
- **Burn ledger** parses `burn` / `burnChecked` instructions of the mint from `getSignaturesForAddress` + `getTransaction (jsonParsed)`, newest first with Solscan links, rolling day/week/month stats, and a cumulative chart. The feed is labeled by scan coverage and never affects the headline total.
- **Works everywhere:** the page seeds from build-time snapshots (renders with zero network under a strict CSP) and upgrades to live data where fetches are allowed. `#p.rpc=https://your-rpc` switches to a private RPC since the public endpoint rate-limits history hydration.
- **Provenance table** cites every burn mechanic to its PR and source file (renames #1496/#1499, respec #1553, voice NPC #1099, Logol #1473, arena rake #799, Aldrin #938, skins #897, character market #1497, premium listings #1551, ads, talent, Claudium), and explicitly lists the routes that do NOT burn.

## Build and test

- `node tools/burn-tracker/build.cjs` fetches live price + supply, scans recent mint history for burns, stamps snapshots, and emits `/burn.html`. Fail-closed: price or supply fetch failure fails the build.
- `node tools/burn-tracker/test_burn_model.cjs`: 17 tests, zero dependencies, no mocks of the code under test. Parser fixtures are captured real mainnet/Jupiter response shapes; two live endpoint tests hit Jupiter and mainnet RPC; a parity test asserts `burn.html` embeds `burn_model.cjs` byte-for-byte. Suite is green from this branch.

## Notes for review

- Pure addition: no game, server, or package.json changes; no new dependencies. The page is plain HTML/CSS/JS with system fonts.
- The `ATTRIBUTION` map (burn authority -> label) ships empty on purpose; filling it as mainnet keeper/treasury addresses become public will let feed rows name the mechanic that burned.
- Nothing on the page is financial advice; it reports on-chain history only.

### DIFF
--- a/burn.html
+++ b/burn.html
@@ -1,0 +1,1037 @@
+<title>WOC Burn Tracker</title>
+<style>
+  /* ============ WOC burn chamber - deliberate single dark theme ============ */
+  :root { color-scheme: dark; }
+  .burncalc {
+    /* ground */
+    --page:        #0c0a06;
+    --surface:     #14110b;
+    --surface-2:   #1a1610;
+    --inset:       #0f0c07;
+    /* brand chrome (never a data series) */
+    --gold:        #d8b15a;
+    --gold-bright: #f3d489;
+    --gold-dim:    #6b5226;
+    /* ink */
+    --ink:         #e8dfc8;
+    --ink-2:       #b6ad95;
+    --ink-3:       #8a8270;
+    /* hairlines */
+    --line:        #2a2418;
+    --grid:        #221c11;
+    /* data accent (validated ≥3:1 on surface) */
+    --burn:        #d95926;
+
+    --serif: Georgia, 'Times New Roman', serif;
+    --sans: -apple-system, BlinkMacSystemFont, 'Helvetica Neue', Arial, sans-serif;
+    --mono: ui-monospace, 'SF Mono', SFMono-Regular, Menlo, Consolas, monospace;
+  }
+  .burncalc {
+    background: var(--page);
+    color: var(--ink);
+    font-family: var(--sans);
+    font-size: 14px;
+    line-height: 1.5;
+    min-height: 100vh;
+    padding: 0 20px 72px;
+  }
+  .burncalc *, .burncalc *::before, .burncalc *::after { box-sizing: border-box; }
+  .shell { max-width: 980px; margin: 0 auto; display: flex; flex-direction: column; gap: 28px; }
+
+  /* ---------- runtime error surface ---------- */
+  .err {
+    display: none; margin-top: 16px; padding: 12px 16px; border-radius: 8px;
+    border: 1px solid #7a2a1a; background: #250f08; color: #e8b39a; font-size: 13px;
+  }
+  .err.show { display: block; }
+  .err b { color: #f3c8b0; }
+
+  /* ---------- header / hero ---------- */
+  .masthead { text-align: center; padding-top: 48px; display: flex; flex-direction: column; gap: 6px; }
+  .masthead .rule {
+    height: 1px; max-width: 420px; margin: 0 auto 18px; width: 100%;
+    background: linear-gradient(90deg, transparent, var(--gold-dim) 20%, var(--gold) 50%, var(--gold-dim) 80%, transparent);
+  }
+  .masthead .eyebrow {
+    font-size: 12px; letter-spacing: 5px; text-transform: uppercase; color: var(--ink-3);
+  }
+  .masthead h1 {
+    font-family: var(--serif); font-variant: small-caps; letter-spacing: 3px;
+    font-size: clamp(30px, 5vw, 44px); line-height: 1.1; color: var(--gold-bright);
+    margin: 0; text-wrap: balance; font-weight: 600;
+  }
+  .masthead .mint {
+    display: inline-block; margin: 8px auto 0; padding: 4px 14px; border-radius: 999px;
+    border: 1px solid var(--gold-dim); background: rgba(20, 17, 11, 0.8);
+    font-family: var(--mono); font-size: 11px; color: var(--ink-2); word-break: break-all;
+  }
+
+  .hero { display: flex; justify-content: center; align-items: stretch; gap: 0; flex-wrap: wrap; }
+  .hero .cell { padding: 10px 36px; text-align: center; }
+  .hero .cell + .cell { border-left: 1px solid var(--line); }
+  .hero .big {
+    font-family: var(--mono); font-size: clamp(30px, 4.5vw, 46px); line-height: 1.15; font-weight: 600;
+  }
+  .hero .big.woc { color: var(--burn); }
+  .hero .big.usd { color: var(--ink); }
+  .hero .cap { margin-top: 2px; font-size: 12px; letter-spacing: 2px; text-transform: uppercase; color: var(--ink-3); }
+  .hero .sub2 { margin-top: 4px; font-family: var(--mono); font-size: 10.5px; color: var(--ink-3); }
+  .hero .sub2.live { color: #9fe6a0; }
+  @media (max-width: 620px) {
+    .hero .cell + .cell { border-left: none; border-top: 1px solid var(--line); }
+    .hero { flex-direction: column; }
+  }
+
+  /* ---------- generic card ---------- */
+  .card {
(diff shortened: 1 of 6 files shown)


EnriqueGF (changes requested):
Reviewed the full diff (6 files, +2568) against the root `CLAUDE.md` and the surrounding build/deploy wiring.

**What I ran locally on this branch:** `npx tsc --noEmit` (clean), `node tools/burn-tracker/test_burn_model.cjs` (17/17 green, including the two live-endpoint tests), `node scripts/malware_scan.mjs` (no new findings from this PR), plus a full read of `page-template.html`, `burn_model.cjs`, `build.cjs` and `vite.config.ts`.

The model layer is genuinely nice work: the parsers are fail-closed, the "total = 1e9 minus live `getTokenSupply`" framing is the right call, and keeping the headline independent of feed coverage is exactly the correct separation. The test suite uses real captured response shapes rather than mocking the code under test, and the byte-for-byte embed parity test is a good idea. My concerns are all around the edges, not the core math.

Requesting changes on the following.

---

### Blocking

**1. The page is never deployed, and the build never runs.** `build.cjs:178` writes to repo root `burn.html`, but `vite.config.ts:345-352` lists exactly six rollup inputs and `burn.html` is not one of them. Every other standalone static page (`links.html`, `privacy.html`, `merch.html`, `terms.html`, ...) lives in `public/`, which is copied verbatim into `dist/`; the server serves from `dist/` (`server/main.ts:326`). There is also no `/burn` row in `STATIC_PAGE_ALIASES` (`vite.config.ts:81-106`), and `grep` finds zero references to the tracker in `package.json` or any workflow. As committed, `/burn.html` 404s in production and the baked snapshots will silently rot with no CI step refreshing them. Output should go to `public/burn.html`, with an alias row and a build/release hook that invokes `build.cjs`.

**2. Emoji in player-facing copy.** `CLAUDE.md` forbids emoji "anywhere ... or player-facing copy", with the explicit rider that an emoji standing in for a real label still needs its real text. Four sites, present in both the template and the generated `burn.html`: `&#128293;` (fire) at `page-template.html:313` and `:558`, `&#9878;` (scales) at `:303`, `&#8646;` (arrows) at `:308`. They are HTML numeric entities, which is why the `Stop` hook's codepoint grep did not catch them, but they render as emoji. `:558` (`Burned &#128293;`) is the clearest case since it is decorative next to a real word.

**3. `npm run ci:changed` fails on these files.** Biome gates changed files and fails on errors. Per-file counts on this branch: `burn.html` 21 errors, `burn_model.cjs` 12, `page-template.html` 10, `build.cjs` 1, `test_burn_model.cjs` 1. `biome.json` includes `**` minus `public` and a few generated paths, so `tools/` and root `burn.html` are all in scope. (Moving the artifact to `public/` per finding 1 conveniently drops it out of Biome scope; the `tools/` sources still need a pass.) `npm run gate` was not run to completion here, but the changed-file Biome arm is the one that fails.

**4. Base branch.** This targets `main`. Per `CLAUDE.md` ("Default task workflow"), contributions base off the latest release branch (`release/v0.28.0`), which is the active integration base; `main` trails it.

**5. The document has no `<head>`.** `burn.html:1` is `<title>`: no `<!doctype html>`, no `<meta charset="utf-8">`, no viewport meta. Two concrete consequences, not just pedantry: the served bytes contain non-ASCII (`…` at `:224`, `:229`, `:495`; `·` at `:526`, `:534`; `≥` at `:22`), which mojibakes without a charset when the header does not supply one, and the mobile media queries at `:80`, `:142`, `:182` will not behave as designed on phones without a viewport meta. Also worth noting the PR body says "renders with zero network under a strict CSP" - the zero-network part is true and well done (snapshots baked at `burn_model.cjs:32` and `:85`), but the page ships no CSP of its own, and its inline `<style>`/`<script>` would need hashes or a nonce under one.

---

### Should fix

**6. Unescaped burn authority into `innerHTML`.** `page-template.html:554` + `:557-559`:
```js
var who = M.ATTRIBUTION[b.authority] || M.fmt.shortAddr(b.authority) || 'unknown authority';
l.innerHTML = '<span class="amt">' + ... + '<span class="who">' + who + '</span>';
```
`b.authority` comes straight from RPC JSON with no base58 validation (`burn_model.cjs:159`), and `fmt.shortAddr` returns it verbatim at length <= 12. The delivery path is the attacker-supplied `#p.rpc` override, so a crafted link feeds attacker-controlled markup into `innerHTML`. Not critical (`<script>` via `innerHTML` does not execute, and the 12-char truncation makes an attribute-carrying `` awkward), but it is unbounded markup injection from untrusted network data and trivially fixed: the sibling fields on the same row already do it right with `textContent` (`:566`, `:569`).

To be clear on the one I checked and cleared: the `p.rpc` value echoed into `warnings` (`burn_model.cjs:199`) is **not** an XSS vector - its only consumer is `console.warn` at `page-template.html:379`.

**7. `scannedSigs` inflates on every Refresh, and re-hydrates work already done.** `page-template.html:459-466`: `known` is built only from `chain.burns`, i.e. signatures that turned out to *be* burns. Every non-burn signature therefore passes the `!known[...]` filter on every refresh, gets re-hydrated (up to 25 `getTransaction` calls, against the rate-limited public endpoint) and is re-added to `chain.scannedSigs`. Click Refresh three times on a quiet mint and the "feed covers N scanned mint txs" line (`:579`) triples with no new coverage. Dedupe against a set of all *seen* signatures, not just burn signatures.

**8. Dead branch at `page-template.html:468`.**
```js
if (!chain.oldestScannedMs || (oldest && oldest < chain.oldestScannedMs)) chain.oldestScannedMs = chain.oldestScannedMs || oldest;
```
When `oldestScannedMs` is already truthy and `oldest` is older, the condition passes but the assignment re-assigns the existing value. The "walk coverage backwards" case never fires. `scanDeeper` at `:500` does it correctly (`chain.oldestScannedMs = oldest`); this line should match.

---

### Minor

**9. `mergeBurns` dedupe key can undercount.** `burn_model.cjs:166-176` keys on `sig + '#' + amount`. A single transaction containing two burn instructions of the identical amount (plausible for a batched keeper burn) collapses into one feed row. Including the instruction index in the key would fix it. Feed-only, so the headline total is unaffected - which is precisely why that design choice was the right one.

**10. `fmt.full` / `fmt.full2` have no `isFinite` guard** (`burn_model.cjs:239-242`), unlike `fmt.n` and `fmt.price`. `page-template.html:523` feeds `burned * price.usdPerWoc` into `full2`, and `:370-372` copies the price snapshot with no `|| 0` defaults (unlike the chain state at `:361-366`, which defaults every field). Build-time stamping guards this today (`build.cjs:161`), but the runtime would render a literal `NaN`.

**11. `r.reason.message` at `page-template.html:472`** assumes every rejection is an `Error`; a non-object reason throws a `TypeError` inside the `allSettled` handler and escapes as an unhandled rejection into the red error banner - the failure mode the fail-closed design is trying to avoid. Line `:357` already does this defensively; match it.

**12. Burns with a missing `blockTime`** become `ms: 0` (`burn_model.cjs:138`). The row timestamp guards it (`:569`), but the chart does not: such a point produces a 1970 axis label at `:664` and squashes the whole X domain via `t0 = pts[0].ms` at `:642`.

---

Happy to re-review once the deploy wiring, the emoji, and the Biome arm are sorted. Findings 1 and 2 are the ones that would actually break on merge; the rest of the list is smaller than it looks and the model layer underneath it is solid.

PR #2159, opened 2026-07-19
feat(quests): add rotating daily quests from Marshal Redbrook

jgyy:
## Summary

Adds a rotating **Daily Quests** system: each day, every character gets 3 quests
deterministically rolled from a new pool, offered by the existing hub NPC Marshal
Redbrook (`marshal_redbrook`), reusing the entire existing quest accept/track/turn-in
pipeline rather than introducing a parallel one.

- 8 new low-level repeatable quests (`q_daily_*`, `isDaily: true`) with kill/collect
  objectives against existing zone1 mobs/items, giver+turn-in at Marshal Redbrook.
- A new pure leaf module (`src/sim/quests/daily_quest_pool.ts`) rolls 3 of the 8 per
  character per day: deterministic on `(characterId, day)` via a local FNV-1a seeded
  mulberry32 PRNG, **not** the sim's shared `Rng` stream, so it never perturbs the
  global draw order. `day` is derived from the same realm daily-reset boundary
  (`ctx.raidResetMs`) raid lockouts already use.
- The roll happens lazily the first time a player talks to Marshal Redbrook after the
  boundary rolls over (`Sim.ensureDailyQuests`), and is threaded through
  `computeQuestState` so a daily quest only shows as available if it's in today's
  rolled set. Once-per-day turn-in is enforced so a `repeatable` daily can't be
  farmed same-day.
- `PlayerMeta.dailyQuests` is mirrored to the client through the `IWorld` quests
  facet / wire snapshot (parity pins updated) so the HUD can show a small "Daily"
  badge on tracked daily quests in the quest log.

```mermaid
sequenceDiagram
    participant P as Player
    participant S as Sim (server-authoritative)
    participant Pool as daily_quest_pool.ts (pure)
    participant Meta as PlayerMeta.dailyQuests

    P->>S: talkToNpc(marshal_redbrook)
    S->>S: dailyResetDayIndex(now, raidResetMs)
    alt stored day is stale or missing
        S->>Pool: rollDailyQuestIds(characterId, day, level)
        Pool-->>S: 3 quest ids (deterministic, no shared rng)
        S->>Meta: dailyQuests = { day, questIds }
    else already rolled today
        S->>Meta: (unchanged)
    end
    S-->>P: available quests filtered by isDaily + today's rotation
    P->>S: acceptQuest(dailyId)
    P->>S: turnInQuest(dailyId)
    S->>Meta: mark today's daily as turned in (blocks a second turn-in same day)
    S-->>P: XP + copper reward
```

## Screenshot

Quest log showing a tracked daily quest with the "Daily:" badge and its detail panel
(new UI, no prior equivalent to diff against):

Daily Quests panel

## Test plan

- [x] `tests/sim/daily_quest_pool.test.ts`: deterministic roll per (characterId, day),
      re-rolls on a new day, level-filters eligible quests.
- [x] `tests/daily_quests.test.ts`: talk to Marshal Redbrook rolls 3 dailies, accept +
      turn in for XP/copper, same-day re-talk keeps the same set, a turned-in daily
      can't be turned in again same day, a new day re-rolls.
- [x] `tests/world_api_parity.test.ts`, `tests/snapshots.test.ts`: wire/facet pins
      updated for the new `dailyQuests` field.
- [x] `tests/progression.test.ts`: referential integrity for the new quest/item ids.
- [x] `npx tsc --noEmit`, `biome check --write` on changed files: clean.
- [x] `npm run gate`: green except two pre-existing flaky suites
      (`tests/loot_ffa_sim.test.ts`, a `varkas_boneguard` timeout) unrelated to this
      change and confirmed to pass in isolation.
- [x] Rebased twice onto the moving tip of `release/v0.28.0` (each time the only
      conflict was the generated i18n `pending.ts`, resolved by regenerating with
      `npm run i18n:gen` rather than hand-editing), no merge conflicts.


EnriqueGF (changes requested):
Reviewed the full diff and ran the checks locally in a clean worktree off `release/v0.28.0`.

**What I ran:** `npx tsc --noEmit` (clean), `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/world_api_parity.test.ts`, `tests/snapshots.test.ts` pins, `tests/guide.test.ts`, `tests/progression.test.ts`, `tests/quest_tracker.test.ts`, and both new suites. All green.

**What holds up well:** the rotation math is a genuine pure leaf with no shared-`Rng` draw (determinism invariant respected), the parity pins and the `daily` delta key are updated in the same change, the availability gate lands in `computeQuestState` so offline/online/optimistic all agree, the badge string is an English-only catalog key with the M16 non-Latin fills, and it reuses the existing accept/track/turn-in pipeline instead of forking it.

Two things I would like fixed before merge.

---

### 1. (High, exploit) A relog re-offers a daily that was already turned in today

`PlayerMeta.dailyQuests` is session-only: it is not part of `SerializedCharacter` (`src/sim/sim.ts:1232`, populated at `:2803`), so it is dropped on logout and rebuilt from scratch by `ensureDailyQuests` on the next talk.

The PR body's premise, "not persisted; it re-derives identically on reconnect", is true of the *roll* but not of the *consumption*. The once-per-day enforcement in `turnInQuestCore` works by filtering the id out of `meta.dailyQuests.questIds`, and that deletion is exactly the part that does not survive. On reconnect the roll re-derives the same 3 ids for the same `(characterId, day)`, the consumed one comes back, and `repeatable: true` means `questsDone` does not block it (`quest_commands.ts:65`).

Repro (ran this against the branch, `tests/daily_quests.test.ts` style harness):

```
talk -> accept q_daily_rats -> force ready -> turnIn
  questState('q_daily_rats') === 'unavailable'   // correct
meta.dailyQuests = undefined                     // what a relog does
talk again (same day, same clock)
  questState('q_daily_rats') === 'available'     // BUG
  dailyQuests.questIds === ['q_daily_rats','q_daily_fangs','q_daily_wolves']
```

So a player can farm all 3 dailies for XP and copper an unbounded number of times per day by reconnecting between turn-ins. It is server-authoritative, but the server has the same hole because it goes through the same `serializeCharacter` path. A server restart or a character swap does it too.

The fix wants the *consumed* set to be durable, not just the rolled set: add `dailyQuests` (day plus the remaining or the turned-in ids) to `SerializedCharacter` and restore it in `addPlayer` alongside `questsDone`, keeping `ensureDailyQuests`'s re-roll condition as `stored.day !== day`. A test that round-trips `serializeCharacter` / `addPlayer` across a turn-in would pin it.

### 2. (Medium) An empty or partial roll is sticky for the rest of the day

`ensureDailyQuests` early-returns whenever `meta.dailyQuests.day === day`, and the roll's level filter is applied once, at roll time, against `p.level`.

Marshal Redbrook is the hub NPC that also holds `q_wolves`, the level-1 starter quest, so effectively every new character talks to him at level 1. Every daily gates `minLevel >= 2`, so that first talk stores `{ day, questIds: [] }` and the character sees zero dailies until the next reset, even after levelling. Your own pool test pins the input side of this (`rollDailyQuestIds('char-1', 1, 1)` is `[]`). The same shape applies mid-range: a level-2 character locks in the three `minLevel <= 2` dailies and cannot see the level-3 or level-4 ones that day.

Cheapest fix consistent with the current design: also re-roll when the stored set is a strict subset of what the character's current level now qualifies for (or store `rolledAtLevel` and re-roll when the eligible count grew). Since the roll is a pure function of `(characterId, day)`, re-rolling on a level change is stable and does not hand out extra quests, though it does interact with finding 1, so it is worth deciding both together.

### 3. (Low) The day index is a proxy for the reset boundary

`dailyResetDayIndex` keys on `Math.floor(raidResetMs(now) / DAY_MS)`. Consecutive resets are normally 24h apart so this yields consecutive integers, but the server's boundary is realm-local 3 AM in an operator-configured zone (`REALM_RESET_TZ`, `server/realm.ts:59`), so a DST spring-forward makes one gap 23h. For a zone whose 3 AM lands near 00:30 UTC, two consecutive resets can then floor to the same index and the dailies would not re-roll that day. Very narrow, but you can make it exact for free by keying on the reset instant itself rather than flooring it to a UTC day.

### Nit

The screenshot in the PR body points at a `raw.githubusercontent.com/jgyy/world-of-claudecraft/...` URL. The PNG is committed in this PR, so a repo-relative path would not rot if that fork moves.

---

Findings 1 and 2 are behavioral rather than structural, so the shape of the change is fine as-is. Happy to re-review quickly once the persistence hole is closed.

jgyy:
Thanks for the thorough review, Enrique, especially the repro on finding 1. Pushed a fix for both blocking findings.

**Finding 1 (relog exploit):** persisted the daily-quest state instead of leaving it session-only. `CharacterState` now carries `dailyQuests` (day plus the remaining, not-yet-turned-in ids, same shape as the runtime/wire field) and a new `dailyQuestsMeta` (day, `consumedIds`, and the level the roll was made at). `serializeCharacter` writes both, `addPlayer` restores both alongside `questsDone`. `turnInQuestCore` now records the turned-in id into `dailyQuestsMeta.consumedIds` in the same step it filters it out of `dailyQuests.questIds`, so the consumption survives logout/relog/server restart, not just the roll. `ensureDailyQuests` still re-rolls when `stored.day !== day`, and on a fresh roll it filters out anything already in `consumedIds` so a same-day re-roll can never re-offer something already turned in.

**Finding 2 (sticky empty/partial roll):** `ensureDailyQuests` now also re-rolls within the same day when the eligible pool has widened since the last roll (compared via a new pure `eligibleDailyQuestIds(level)` export from `daily_quest_pool.ts`, keyed off `dailyQuestsMeta.rolledAtLevel`). A re-roll preserves `consumedIds`, so it never re-offers or double-grants credit for something already turned in that day; it only reveals newly-eligible dailies. Talking to Marshal Redbrook at level 1 for `q_wolves` no longer locks a character out of all dailies for the rest of the day.

Left finding 3 (the DST/day-index edge) out of this change to keep the diff scoped to the two blocking items; happy to file it separately if you'd like it tracked.

Added `tests/daily_quests.test.ts` coverage for both: a round-trip test that turns in a daily, calls `serializeCharacter`, rejoins a fresh `Sim` via `addPlayer({ state })` (same day/clock), and asserts the daily stays unavailable and the roll is not reset; and a level-up test that rolls empty at level 1, levels to 5 mid-day, and asserts a same-day re-roll to the full 3 dailies.

Verified: `npx tsc --noEmit` clean, `npx vitest run tests/daily_quests.test.ts tests/architecture.test.ts tests/world_api_parity.test.ts tests/snapshots.test.ts tests/quest_tracker.test.ts tests/progression.test.ts tests/localization_fixes.test.ts` all green (491 passed, 3 pre-existing skips), and `npx @biomejs/biome check --write` on the four changed files (no errors, pre-existing non-null-assertion style warnings only, matching the rest of the test file).

EnriqueGF (changes requested):
Reviewed the full diff against `release/v0.28.0` and ran the checks locally. Overall this is a well-built feature: the rotation math is a genuine pure leaf, it draws zero times from the shared `Rng` so the parity draw-order digests are untouched, the accept path is gated server-side through `questState`, the parity/wire pins are all updated in the same change, and the persistence is additive and back-compat. Nice work on threading `dailyQuestIds` through both `questState` and `optimisticQuestState`: I went looking for a missed `computeQuestState` call site and there genuinely isn't one (every other consumer, minimap markers, map window, nameplates, quest dialog, `obs.ts`, goes through the `IWorld.questState` seam).

**What I ran**

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/world_api_parity.test.ts`, `tests/parity`, `tests/guide.test.ts`, `tests/progression.test.ts`, `tests/sim.test.ts`, `tests/env_protocol.test.ts`, `tests/server` (109 files), plus every `tests/quest*` suite and the two new suites: all green.
- `tests/snapshots.test.ts` has one failure (`active title wire (Book of Deeds) ... key lockstep`, `Cannot read properties of undefined (reading 'OPEN')`). I reproduced it on a clean `origin/release/v0.28.0` checkout: pre-existing, Node 20 has no global `WebSocket`. Not yours.
- Biome on the changed files: warnings only, no errors or format diffs.
- Confirmed `spider_leg` (0.4 from `webwood_spider`) and `wolf_fang` (0.45 from the zone1 wolves) both drop in zone1, so the two collect dailies are actually completable at their `minLevel`.
- The `src/ui/i18n.locales/` edits are the M16 exception (non-Latin fills for new wordy English values), correctly scoped to ja/ko/ru/zh, so that is fine rather than an overlay-editing violation.

**Findings**

**1. (medium) The mid-day level-up re-roll can hand out more than 3 dailies in a day.** `Sim.ensureDailyQuests` re-runs `rollDailyQuestIds` at the wider level and subtracts `consumedIds` from the *new* roll, but never subtracts the consumed count from the cap. If the already-turned-in id is not in the fresh roll, the player is offered 3 more on top of the one already banked.

I quantified it over 200 character ids x 3 days, comparing a level-2 roll (eligible pool of 3) with one turn-in against the post-level-up level-10 roll: **385 of 600 cases exceed the 3-per-day cap.** (My first single-seed repro happened to land in the lucky 36%, hence the brute force.) Impact is bounded, there are only three `minLevel` thresholds (2, 3, 4) so it is a handful of extra low-level dailies over a character's lifetime, but it does break the once-per-day contract the feature is built on. One-line fix:

```ts
const offered = rolled.filter((id) => !consumedIds.includes(id));
meta.dailyQuests = { day, questIds: offered.slice(0, DAILY_QUEST_COUNT - consumedIds.length) };
```

Worth noting the existing test at `tests/daily_quests.test.ts` ("re-rolls without losing turned-in credit") structurally cannot catch this: it starts from an empty level-1 roll and never turns anything in, so the assertion does not cover the behavior its name claims.

**2. (medium) Offline and headless hosts never rotate the dailies.** `dailyResetDayIndex` reads `lockoutNowMs()`, which offline defaults to `Math.floor(sim.time * 1000)` starting at 0, and `raidResetMs` defaults to `now + 24h`. I verified the resulting day index directly: it is a constant `1` at t=0, t=12h, and t=23.9h of sim time, and `sim.time` restarts at 0 each session. So an offline or RL character rolls the identical 3 quests forever and the "daily" never resets within a session. The server is fine (real 3 AM boundary). The module comment currently glosses this as "keeps the index stable within a day", which undersells it. Either key the offline index off a sim-time day count, or document the split explicitly.

**3. (low) `maybe('daily', undefined)` puts `null` into a field typed `| undefined`.** `maybe` serializes `value ?? null`, so an unrolled character ships `"daily":null`; the decoder guards on `!== undefined`, so `null` passes and is cast into `dailyQuests: {...} | undefined`. Harmless today since every read is `this.dailyQuests?.questIds`, but online now returns `null` where the offline `Sim` returns `undefined`. A `s.daily === null ? undefined : ...` at the decode site closes it.

**4. (low) `ensureDailyQuests` grows `sim.ts`.** It is about 20 lines of stateful game-system orchestration (day index, staleness, level-widen check, consumed filter, `wireRev` bump) added directly to the coordinator, and it needs nothing from `Sim`'s private state that is not already on `SimContext` (`lockoutNowMs` and `raidResetMs` are both existing callbacks there). Moving it next to the pool module as `ensureDailyQuests(ctx, meta, level, npcQuestIds)` would satisfy the module-first rule and, usefully, make finding 1 unit-testable without spinning up a `Sim`.

**5. (low) The `rolledAtLevel: 1` fallback in `turnInQuestCore` is a re-roll trap.** If `dailyQuestsMeta` is ever absent while `dailyQuests` is present, `eligibleGrew` compares against `eligible(1)`, which is empty since every daily has `minLevel >= 2`, so it evaluates true and forces a re-roll plus a `wireRev++` heavy resend on every subsequent talk that day. `ensureDailyQuests` always writes both together so I could not reach it in practice, but defaulting to the player's current level would remove the trap entirely.

**6. (nit) No test pins the invariant the design most depends on.** The whole "local PRNG, not the shared stream" argument rests on a daily roll leaving `sim.rng` at the same position. That is true today (parity fixtures are byte-unchanged), but nothing asserts it, so a future refactor to use `ctx.rng` would slip through with only the parity digests to catch it.

I checked and dismissed a few other things: the `turnin` path cannot reach `turnInQuestCore` without a prior roll (accept already requires `questState === 'available'`), so there is no same-day repeat exploit there; quests are not in the guide corpus so no wiki regeneration was needed; and `QUEST_ORDER` growing the RL obs vector by 16 dims is handled by the dynamic `obs_size` handshake, though it does shape-invalidate existing checkpoints.

Requesting changes on finding 1 only. The rest are happy to land as follow-ups if you prefer.

jgyy:
Rebased onto the moving tip of `release/v0.28.0` (the conflict was the generated i18n `pending.ts`, resolved by regenerating via `npm run i18n:gen`, plus the IWorld member-count and delta-key-count pins in `tests/world_api_parity.test.ts` / `tests/snapshots.test.ts`, which needed bumping past the release's own additions (Phase 8's mobile-station pair) to also account for this branch's `dailyQuests` member: 252 members, 50 delta keys).

Addressed the blocking finding from the second review round:

**Finding 1 (per-day cap breach on level-up re-roll):** `ensureDailyQuests` now caps the offered set at `DAILY_QUEST_COUNT - consumedIds.length`, not just `DAILY_QUEST_COUNT`, so a level-up mid-day after a turn-in can no longer stack a fresh 3 on top of the one already banked. Added a regression test (`tests/daily_quests.test.ts`, "never exceeds DAILY_QUEST_COUNT total...") that reproduces the exact repro shape (turn in at level 2, level up to 10, re-talk) and asserts consumed + offered stays <= 3.

Also folded in the lower-severity follow-ups rather than deferring them:

- **Finding 3 (null vs undefined):** `ClientWorld`'s wire decode now normalizes `daily: null` (from `maybe()`) back to `undefined`, matching the offline `Sim`'s shape exactly.
- **Finding 5 (rolledAtLevel fallback trap):** the fallback in `turnInQuestCore` now defaults to `Number.MAX_SAFE_INTEGER` instead of `1`, so a hypothetically-missing `dailyQuestsMeta` can't force a re-roll + `wireRev` bump on every subsequent talk that day.
- **Finding 2 (offline/headless never rotates):** documented explicitly in `daily_quest_pool.ts` as intentional (no real calendar boundary offline, so "daily" degenerates to "rolled once per session" there) rather than glossed over.
- **Finding 6 (no rng-position pin):** added a test asserting rolling the daily set draws zero times from the shared `sim.rng`, by comparing the subsequent draw sequence of a sim that rolled dailies against an untouched twin.

Left finding 4 (extracting `ensureDailyQuests` into `daily_quest_pool.ts` behind `SimContext`) as a follow-up per your offer, to keep this diff focused on the blocking + cheap fixes.

Verified: `npx tsc --noEmit` clean, targeted suites green (`tests/daily_quests.test.ts`, `tests/sim/daily_quest_pool.test.ts`, `tests/world_api_parity.test.ts`, `tests/snapshots.test.ts`, `tests/architecture.test.ts`, `tests/localization_fixes.test.ts`, `tests/progression.test.ts`, `tests/quest_tracker.test.ts`, `tests/sim.test.ts`, `tests/env_protocol.test.ts`: all passing, 506+103 tests), and the pre-push QA floor (tsc + guard tests + changed-file biome + copy-scan) passed clean on push.

jgyy:
Addressed the second review round.

**Finding 1 (required, medium): mid-day level-up re-roll could exceed the 3-per-day cap.**
Already fixed in the prior commit on this branch (`f568b8189`) before this session picked
the PR back up: `ensureDailyQuests` now caps the OFFERED set at
`DAILY_QUEST_COUNT - consumedIds.length` instead of just `DAILY_QUEST_COUNT`, so a
level-up re-roll can no longer hand out a fresh 3 on top of an already-turned-in daily.
I verified the fix against the diff and against a clean checkout of `release/v0.28.0`:
the one-line change matches the suggested fix exactly, and the new regression test
(`tests/daily_quests.test.ts`, "never exceeds DAILY_QUEST_COUNT total (consumed + offered)
after a level-up re-roll") reproduces the scenario you quantified (level 2 with a turn-in,
then level up to 10, then re-roll) and pins consumed + offered <= 3. I re-ran it along with
`tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, and `tests/sim.test.ts`:
all green, `tsc --noEmit` clean.

**Follow-ups already folded into the same prior commit:**
- The `rolledAtLevel: 1` fallback trap (finding 5) now defaults to
  `Number.MAX_SAFE_INTEGER`, so a missing `dailyQuestsMeta` can no longer force a spurious
  re-roll/`wireRev` bump every subsequent talk.
- The `maybe('daily', undefined)` -> `null` wire round-trip (finding 3) is normalized back
  to `undefined` at the `ClientWorld` decode site in `src/net/online.ts`, matching the
  offline `Sim` shape.
- The offline/headless same-session-forever rotation behavior (finding 2) is now documented
  explicitly in `daily_quest_pool.ts` instead of glossed over, per your suggestion that a doc
  fix is acceptable there.
- The Rng-purity invariant (finding 6, nit) now has an explicit test: two identically-seeded
  sims, one that rolls a daily set and one that doesn't, draw the identical subsequent
  sequence from the shared `Rng`.

**Done in this session:** finding 4 (module-first, low) - moved `ensureDailyQuests` off the
`sim.ts` coordinator into `quests/daily_quest_pool.ts` as a plain function of
`(meta, playerLevel, npcOffersDaily, lockoutNowMs, raidResetMs)`, since it only ever touched
`PlayerMeta` plus the two existing `SimContext` callbacks. `sim.ts` keeps a thin delegate
that resolves whether the talked-to NPC offers a daily and forwards into the core. This is a
move, not a rewrite: behavior is unchanged and the existing tests (including the finding-1
regression and the Rng-purity pin) pass against the extracted module without modification.

Also re-verified round 1's items against current HEAD since I was asked not to trust the
summary blindly: `PlayerMeta.dailyQuests`/`dailyQuestsMeta` are both in `SerializedCharacter`
and restored in `addPlayer` (the relog/turn-in persistence hole is closed), and the
empty/partial-roll stickiness is handled by the `eligibleGrew` re-roll check. No merge
conflicts against `upstream/release/v0.28.0`.

CI is green on all required checks (browser regressions, biome, PR gate freshness/typecheck/
builds, English-only legal gate x4, screenshots).

Rubsey:
@jgyy I think this is a good idea, but I think it would be a better idea for it to be level 20 daily quests. This can reward max level players with gold.

jgyy:
Thanks for the suggestion, Rubsey. I looked into it before replying: MAX_LEVEL in this repo is 20, so gating all 8 dailies to level 20 would turn them into a pure endgame gold sink rather than the leveling-arc feature this PR currently ships. Right now each daily has its own per-quest minLevel (2 to 4, see src/sim/content/daily_quests.ts), so low and mid level characters already get a taste of Marshal Redbrook's rotation while they level, and the reward scales with the quest.

Raising the floor to 20 across the board would drop that early/mid game audience entirely and change who the feature is for, which reads to me as a product call rather than a review-blocking fix, so I don't want to fold it into this PR without a signoff. Two ways I could see this landing if you want it:

1. A follow-up PR that adds a small set of level-20-only bonus dailies from Marshal Redbrook (or a second gold-focused NPC) alongside the existing leveling dailies, so both audiences are served.
2. If the intent really is "dailies are an endgame gold feature, not a leveling feature," that is a scope change to this PR's design and I'd like to confirm that's what you want before reworking the minLevel gates and quest text here.

Happy to pick either up, just want your call on which one before I touch the gating.

jgyy:
Thanks for the thorough review across both rounds, EnriqueGF. Went back through the branch's actual code (not just the commit messages) against every finding before replying:

**Round 1**
1. (High, exploit) Fixed: `dailyQuests` / `dailyQuestsMeta` are now part of `CharacterState`/`SerializedCharacter`, restored in `addPlayer` alongside `questsDone`, and round-tripped in `tests/daily_quests.test.ts` ("persists a turned-in daily across serializeCharacter/addPlayer so a relog cannot re-offer it"). The consumed-id set (`dailyQuestsMeta.consumedIds`) is what survives the relog, independent of `repeatable`.
2. (Medium, sticky roll) Fixed: `ensureDailyQuests` re-rolls when the stored roll's day is stale OR when the eligible pool has grown since the roll (tracked via `dailyQuestsMeta.rolledAtLevel`), so a level-1 character's empty first roll no longer locks out the rest of the day once they qualify for more dailies.
3. (Low, DST boundary) Documented rather than changed: added a comment on `dailyResetDayIndex` (this commit) explaining that `raidResetMs` already resolves DST transitions correctly, but flooring that instant to UTC-day granularity is still a proxy that could in principle repeat/skip a day very near a transition. No live report against it, and it's explicitly lower priority than 1 and 2, so I documented the caveat and the precise fix (key on the realm-local calendar date) instead of changing behavior speculatively.
4. (Nit, screenshot URL) Fixed: edited the PR body to a repo-relative path (`docs/screenshots/daily-quests/after-quest-log.png`) instead of the fork's `raw.githubusercontent.com` URL.

**Round 2**
1. (Medium, cap overflow, blocker) Fixed: `ensureDailyQuests` now offers `rolled.filter(not consumed)` capped at `DAILY_QUEST_COUNT - consumedIds.length`, exactly the pattern you suggested, so a mid-day level-up re-roll can never push the day's total (consumed + offered) past 3. Added `tests/daily_quests.test.ts`: "never exceeds DAILY_QUEST_COUNT total (consumed + offered) after a level-up re-roll", which turns a daily in first and then triggers the re-roll, per your note that the earlier test didn't actually cover this.
2. (Medium, offline rotation) Documented: `dailyResetDayIndex`'s module comment now spells out the split explicitly ("Known split" paragraph) rather than the misleading "keeps the index stable within a day" line: offline/headless `Sim` has no wall clock, so the day index is a constant `1` for the whole session and dailies degenerate to "rolled once per session" there. A real sim-time day counter would need a calendar concept the offline/headless hosts don't have, so I judged documenting it correctly as the smaller, correct change here.
3. (Low, null vs undefined) Fixed: decode site in `src/net/online.ts` now does `s.daily === null ? undefined : ...`, so `ClientWorld.dailyQuests` matches the offline `Sim`'s `undefined` for an unrolled character instead of leaking wire-level `null`.
4. (Low, sim.ts growth) Fixed: `ensureDailyQuests` is extracted into `src/sim/quests/daily_quest_pool.ts` as a function of `(meta, playerLevel, npcOffersDaily, lockoutNowMs, raidResetMs)`, called from `sim.ts` via a thin delegate. It only needed `ctx.lockoutNowMs`/`ctx.raidResetMs`, both already on `SimContext`, so no new seam members were needed. This also made it directly unit-testable (see round-2 #1's test).
5. (Low, rolledAtLevel:1 trap) Fixed: `turnInQuestCore` now defaults to `Number.MAX_SAFE_INTEGER` instead of `1` when `dailyQuestsMeta` is absent, so the fallback can never look like a genuine low-level roll and force a spurious re-roll.
6. (Nit, rng isolation) Fixed: added "rolling the daily set draws zero times from the shared sim Rng" to `tests/daily_quests.test.ts`, asserting the shared `Rng`'s state/position is unaffected by rolling or re-rolling dailies.

Nothing was skipped. `npx tsc --noEmit`, the targeted suites (`tests/daily_quests.test.ts`, `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/quest_tracker.test.ts`, `tests/progression.test.ts`, `tests/snapshots.test.ts`) and biome on the changed file are all green. Pushed to `feature/daily-quests`.

jgyy:
Rebased the branch onto the current tip of release/v0.29.0 and fixed the two red checks. Both traced to the same root cause, not to anything in the daily-quests logic itself.

What was wrong:

The release branch had moved on since the last push. Merging it in surfaced a parameter-order collision on the shared quest-state helpers. This branch adds a trailing optional `dailyQuestIds?: string[]` to `computeQuestState` / `optimisticQuestState`, and release/v0.29.0 independently added its own trailing optional `withinCadence?: ReadonlySet<string>` (the Phase 14 work-order cadence mirror). After the merge the signature is `(..., professionState?, dailyQuestIds?, withinCadence?)`.

The branch had already threaded the new `dailyQuestIds` argument through every caller: `ClientWorld.questState` in online.ts, the sim-side `questState`, and tests/professions_quest_cadence.test.ts. The one caller it missed was tests/quest_state_optimistic.test.ts (an upstream-owned suite), which still passed the cadence Set positionally into the `dailyQuestIds` slot. That produced the tsc failure (TS2740: `Set<string>` is not `string[]`) in the "freshness, typecheck, builds" check, and at runtime the cadence Set never reached `withinCadence`, so the gate resolved to `available` instead of `unavailable`, which is the assertion failure that reddened the English-only gate shard.

What I changed:

- tests/quest_state_optimistic.test.ts: inserted the missing `undefined` `dailyQuestIds` argument in the four cadence-passing calls, matching how professions_quest_cadence.test.ts was already updated.
- tests/world_api_parity.test.ts: resolved the merge conflict on the pinned member counts by taking the union of both sides. The merged IWorld seam carries both this branch's `dailyQuests` data member and the release's `unbindItem` command, so the pins move to 261 total (72 data, 189 method), and the facet-union pin moves to 261 to match.

The `server/daily_finalize_guard.ts` deletion in the merge is correct and expected: that file entered the branch from an earlier release merge, and v0.29.0 reorganized that daily-rewards logic into `server/daily_reward_schedule.ts`.

Verification:

- `npx tsc --noEmit` clean.
- `npm run i18n:gen` and `npm run wiki:content` produce no diff (no generated-artifact drift).
- Green: tests/quest_state_optimistic.test.ts, tests/world_api_parity.test.ts, tests/professions_quest_cadence.test.ts, tests/daily_quests.test.ts, tests/sim/daily_quest_pool.test.ts, tests/quest_commands.test.ts, tests/quest_tracker.test.ts, tests/architecture.test.ts, tests/localization_fixes.test.ts, tests/snapshots.test.ts, tests/command_facets.test.ts, tests/command_schema.test.ts.
- Biome clean on both changed files.

jgyy:
Retargeted onto release/v0.30.0 (content-identical to v0.29.0 plus one empty merge commit) and merged it into this branch to clear the CONFLICTING status.

Conflicts resolved (all on the trailing-optional-parameter collision the prior fix already diagnosed for v0.29.0, reproduced fresh against v0.30.0):
- `src/net/quest_state_optimistic.ts`: kept this branch's `dailyQuestIds?: string[]` parameter ahead of `withinCadence`.
- `src/sim/quests/quest_commands.ts`: same parameter-order conflict in `computeQuestState`.
- `src/sim/types.ts`: kept `QuestDef.isDaily` alongside the release's Phase 14 cadence comment.
- `src/ui/i18n.resolved.generated/pending.ts` and `tests/world_api_parity.test.ts`: mechanical, resolved by taking the union of both sides.

No newer unresolved review threads found (checked via the GraphQL reviewThreads query); the last round of feedback (relog-exploit persistence and sticky empty/partial daily roll) was already fixed on this branch and I verified both fixes are intact after the merge.

What needed a follow-up fix after the merge: the pinned counts in `tests/world_api_parity.test.ts` and `tests/snapshots.test.ts` were computed against the old v0.29.0 merge, not this v0.30.0 merge, so they were stale (261 vs the real 254 IWORLD_MEMBERS, 55 vs the real 56 ALL_DELTA_KEYS). Recomputed both from the actual merged code and pinned the real numbers, which is what was failing "PR gate (English-only legal)" and the freshness check. Also reran `npm run i18n:gen` so the generated i18n artifacts match the merged catalog sources.

`npm run gate` is green locally (all 11 steps, full test suite, tsc, builds). Pushed to `feature/daily-quests`; PR now reports `mergeStateStatus: CLEAN` / `mergeable: MERGEABLE`.

PR #2163, opened 2026-07-19
HOTFIX 0.28.0: restore specialization power floor

ryan-foo:
## HOTFIX: target v0.28.0 and release ASAP

This PR restores the passive specialization power floor lost during the Talents 2.0 transition.
It is intended to land on `release/v0.28.0` and ship as soon as review and CI allow.

The immediate goal is to stop current classes from feeling materially weaker. This creates a
stable power floor while class owners rebalance the nine classes, improve rotations, and develop
new class fantasies in separate follow-up PRs.

This is not the final v0.28 class balance pass.

## Root cause

Talents 2.0 retained specialization signatures and masteries, then replaced the old point trees
with horizontal choice rows. The migration did not retain the persistent stats, defenses,
throughput, cast-time, resource-cost, cooldown, and ability modifiers from the old level-20
reference allocations.

Players therefore kept specialization identity but lost a real layer of passive power.

## What changes

- Restores automatic passive packages for 26 specializations across all nine classes.
- Applies each package at full strength as soon as the specialization unlocks at level 5.
- Keeps current mastery and all six Talent 2.0 choice rows additive.
- Uses the existing talent modifier resolver, so percentage effects scale with learned spell
  ranks and equipment without editing shared spell definitions.
- Adds no card, rank, tab, notification, ability, proc, or action-bar button.
- Leaves Mage Chronomancy unchanged because the current healer has no pre-v0.27 healer baseline
  to restore.
- Excludes historical modifiers that target abilities unavailable to the current specialization.
  Redirecting those values would be new class design, not a hotfix restoration.

The accepted hotfix tradeoff is stronger early leveling. Rebuilding the old point-by-point curve
would delay the restoration and add a second progression model.

Values are sourced from the pre-v0.27 level-20 Nythraxis reference allocations at
`6155ad2079906402ce87e64c778b31cd88ef2875`, compared with the v0.28 branch baseline at
`571ab021995165098ce691a89b227732bcbb8fae`.

The implementation and full 26-spec matrix are documented in
`docs/prd/talents-v028-passive-restoration-hotfix.md`.

### Specialization power restoration

Restored passive specialization power lost during the Talents 2.0 transition for 26
specializations.

- Restored bonuses apply automatically when you select a specialization.
- They are added on top of your specialization mastery and talent choices.
- Effects include stats and improvements to relevant spell damage, healing, resource costs, cast
  times, cooldowns, and defenses.
- Bonuses scale naturally with learned spell ranks and equipment.
- No new abilities or action-bar buttons were added, and no respec is required.
- The full package applies from specialization unlock, so early-level characters will be stronger
  than before this hotfix.
- Chronomancy is unchanged because its healer kit has no former healer baseline to restore.

This establishes the immediate v0.28.0 power floor. Class fantasy, rotation, and cross-class
balance updates follow through class-owner PRs and PBE feedback.

## Test-first bug proof

Fails before the fix on unmodified `release/v0.28.0`:

```text
/Users/maxc/code/world-of-claudecraft/node_modules/.bin/vitest run tests/spec_baselines.test.ts

Test Files  1 failed (1)
Tests       2 failed | 1 passed (3)

FAIL: restores the complete repository-backed baseline for all 26 applicable specs
FAIL: applies the full baseline as soon as a spec unlocks, without changing Chronomancy
```

Passes after the fix:

```text
vitest run tests/spec_baselines.test.ts tests/spec_masteries.test.ts tests/talents.test.ts tests/architecture.test.ts

Test Files  4 passed (4)
Tests       70 passed (70)
```

Parity gate after deliberate golden refresh:

```text
vitest run tests/parity

Test Files  3 passed | 1 skipped (4)
Tests       180 passed | 1 skipped (181)
```

The six changed parity traces retain identical frame counts, RNG draw counts, RNG digests, and
entity ID progression. State and combat-event values change as expected from the restored stats
and throughput.

## Validation notes

- The broad suite's Varkas Boneguard test exceeded its existing 5-second timeout in this checkout;
  the isolated test passes with a 15-second timeout. This branch does not touch that test or code.
- Local `tsc --noEmit` is blocked by existing unresolved optional packages such as
  `@aws-sdk/client-sesv2`, `prom-client`, and wallet dependencies in the shared installation.
- The release-version check reports the target branch is still stamped `0.27.2` in package and
  platform metadata. This PR intentionally does not perform the release-version preparation step.


ryan-foo:
@rubsey Could you review this v0.28.0 hotfix when you have a chance? It restores the passive power floor for 26 specializations while keeping Talents 2.0 choices additive. The PR is intentionally scoped to restoration so class owners have time for the larger rotation, balance, and class-fantasy work.

patrick261 (commented):
This one verifies well. I re-derived the numbers and ran the gates independently rather than trusting the description:

- All 26 baseline packages trace exactly, field by field, to the pre-v0.27 level-20 allocations in `scripts/nythraxis_matrix.ts` at 6155ad2, with no invented values. The documented exclusions check out too (heroic_strike is excluded from all three warrior specs in v0.28, rend is retired, arcane_missiles is out of frost's kit).
- Every (ability, modifier) pair in SPEC_BASELINES is live in the current bake: applyTalentMods scales heals, HoTs, and absorbs through dmgPct, and buffPct covers the scalable selfBuff kinds, so nothing in the restored set is silently dead (demon_skin, power_word_shield, and rockbiter_weapon included).
- The hook lands at the one computeTalentModifiers choke point every host shares, resolves at allocation/load time (no hot-path cost), and persisted characters pick the floor up on next load with no migration.
- Against the PR's recording base (571ab02), the six refreshed traces change zero RNG draw counts, zero digests, and zero entity ids; only derived state and event values move. `tsc --noEmit` is clean at the PR head (closing the validation gap in the description), the targeted suites pass 70/70, parity passes 180 with 1 skipped, and the guide freshness gate is green.

Findings:

- **blocking**: the six goldens are stale against current `release/v0.28.0`. After this branch was cut, base landed two full golden regens (cb56e2d63, the six static master NPCs, which shifts constructor entity ids in every scenario; cc8dc373a, the recipesGrandfathered flag), so taking this PR's side of the conflicts would revert those. Merge base into the branch and re-record with `UPDATE_PARITY=1` (tests/parity/parity.test.ts:11) on the merged tree, then confirm draws and digests are still unchanged.
- **should-fix**: fiesta standardization now enters mages at a disadvantage. fiestaStandardize (src/sim/social/fiesta.ts:224) hands every fighter defaultBuild, which picks `specs[0]` (src/sim/content/talents.ts:672); for mage that is arcane (Chronomancy), the one spec with no baseline. The block's own comment promises "every fighter enters equal", and the fiesta goldens show every other class's fighter gaining stats while the mage stays flat. Either point fiesta's default mage at a baselined spec or record in the PRD that this is intended.
- **should-fix**: both commits are title-only; the repo convention wants a body on every commit (what changed and why, especially the sourcing decision and the deliberate golden regen).
- **nit**: the PR body skips the `.github/PULL_REQUEST_TEMPLATE.md` sections. The substance is all there, it just is not in the shape the template asks for.
- **nit**: two cheap test-hardening ideas in tests/spec_baselines.test.ts. numericDelta (line 129) subtracts every key of the resolved mod, so a future baseline touching a non-numeric field (castWhileMoving, addEffects) would coerce or go NaN instead of failing loudly; skip or type-check non-number keys. And the kit test only asserts each ability id exists, not that the modified dimension is nonzero, so a later kit change (say corruption made instant) would silently kill a baseline row.
- **note**: "Talents 2.0 retained masteries" is not quite exact for rogue. Assassination and subtlety masteries were redesigned on the release branch (the "Balance pass" comments in talents_classic.ts), so part of those two specs' old passive power is deliberately not restored here. The table double-counts nothing; just confirm the rogue floor lands where you want it.

Level-5 full-strength unlock and the Chronomancy exclusion are both pinned by tests, no i18n, wiki, or deeds obligations are triggered, and the golden refresh is correctly its own reviewable commit.

PR #2167, opened 2026-07-19
feat(loot): add dungeon shield progression

ryan-foo:
## What

- Adds Boundstone Bulwark, a rare item level 22 shield from Sanctum Boneguards at 8% in their bonus group.
- Adds Drowned Moon Aegis, an epic item level 24 shield from Ysolei at 15%.
- Adds Gravewyrm Bulwark, an epic item level 26 shield from Korzul at 10% in the boss bonus group.
- Gives each shield an exact offhand stat budget, rising armor and block, warrior, paladin, and shaman eligibility, held shield models, and localized item names.
- Keeps heroic upgrades on the existing automatic dungeon variant path.

## Why

Pre-raid shield users jumped from common vendor shields directly to the item level 29 Bonewrought Bulwark. These drops fill the item level 22 to 26 progression gap so tanks can pick up missing defensive stats while leveling through Drowned Temple and Gravewyrm Sanctum.

## Test first

Before the item and loot definitions were added, the new regression suite failed all 3 tests because the shields and their dungeon drops did not exist.

After the fix:

- `tests/dungeon_shield_progression.test.ts`: 3 passed.
- Focused loot, progression, item-level, held-model, and heroic coverage: 117 passed across 8 files.
- Focused architecture and localization coverage: 142 passed, 3 skipped.
- Full `npm run gate`: all 11 stages passed on the current `release/v0.28.0` head, including 16,542 unit tests, 66 browser tests, TypeScript and Svelte checks, and production builds.

## Release note

This is a content balance feature. Levy approval is still required before merge or release.


ryan-foo:
Final refresh validation (2026-07-20)

- Merged release/v0.28.0 through 5b934c23d, including #2163 and #2185.
- Final head: 3e71121e0ec77cf0eb7777d8fd787e3059486950.
- GitHub reports MERGEABLE with no conflicts.
- Focused dungeon, loot, item-level, localization, architecture, and parity coverage: 329 passed, 3 skipped.
- TypeScript and git diff checks passed.
- Worktree is clean.

The upstream CI and PR AI Assist workflow runs are waiting for maintainer approval because this is a fork PR.

EnriqueGF (commented):
Reviewed the full diff and verified the numbers against the real content tables rather than the PR description.

## What I ran

- `npx tsc --noEmit`: clean.
- `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (S3), `tests/world_api_parity.test.ts`, `tests/guide.test.ts`, `tests/parity/` (golden trace): all green.
- `tests/dungeon_shield_progression.test.ts` plus the item/loot/dungeon/heroic suites: green (137 passed on the focused set).
- `npm run i18n:gen` and `npm run wiki:content` both left the tree clean, so the checked-in generated artifacts are fresh.
- Probed `itemLevel` / `primaryStatBudget` directly: item levels come out 22 / 24 / 26 exactly as claimed, and primary stats are 9 / 13 / 14 against budgets of 9 / 13 / 14. Armor 300 to 440 to 540 and block 20 to 24 to 27 ladder cleanly into Bonewrought Bulwark (ilvl 29, armor 680, block 30). `requiredClass` is right, and heroic variants resolve their held model through `heroicOf` in `itemModelKey`, so the manifest entries cover the heroic path too.

The itemization is solid. My one substantive question is the drop rates.

## Balance: every shield is set at roughly double its same-group peers

This is the only thing I would want settled before merge, and since the PR already flags Levy approval it may just need an explicit yes.

- `boneguard_bonus` (`src/sim/content/dungeons.ts:279-281`): Boundstone Bulwark is 0.08 while `boundstone_helm` and `boundstone_girdle` are 0.04 each, so the shield is half the entire group's weight. Sanctum Boneguard is trash, spawned 9 times per Gravewyrm Sanctum run (`dungeons.ts:651-666`), so a full clear is about 53% to yield the shield (1 - 0.92^9) versus about 31% for each of its 4% setmates. A rare tank shield landing more than half the time per clear, off repeatable trash, reads generous.
- `korzul_bonus` (`dungeons.ts:424`): Gravewyrm Bulwark is 0.10 where every other epic in the group is 0.05 or 0.04.
- Ysolei (`temple.ts:271`): Drowned Moon Aegis is 0.15 as an independent roll, and it is the only epic in the entire Drowned Temple (`temple.ts:613` is the sole `quality: 'epic'` in the file). Its blue peers share one exclusive group.

Each of these is defensible alone; the consistent 2x pattern across all three is what made me want to ask. If it is deliberate (shields are one slot with no competition, so a lower effective rate per tank), that is a fine answer, worth a comment in the tables.

## Minor

1. **Seed stability in `korzul_bonus`.** Entries sharing a `rollGroup` are resolved by one partitioned rng draw in array order (`src/sim/loot/loot_roll.ts` header). Inserting Gravewyrm Bulwark in the middle of the list at `dungeons.ts:424` shifts the partition boundaries of the six entries after it. Per-item probabilities are unchanged and `tests/parity/` stays green, but a given seed now yields a different item from that group. Appending at the end (as Boundstone Bulwark correctly does) would have preserved that. Not worth churning the diff for unless you care about seed continuity.

2. **i18n locale fills.** The PR fills es/fr/it/de/zh/ko/ja/pt/ru in `src/ui/i18n.catalog/items.ts` and also edits `src/ui/i18n.locales/*.ts`. Root CLAUDE.md has contributors add English only and reserves the overlays for the maintainer at release time. The values are consistent and every gate passes, so nothing is broken, but this is the maintainer's call to keep or strip.

3. `tests/loot_ffa_sim.test.ts` failed once for me under a parallel multi-file run and passes in isolation. Pre-existing contention flake, not related to this PR.

Nothing blocking from a correctness, determinism, purity, or i18n standpoint. Just the drop-rate question.

PR #2169, opened 2026-07-19
feat(shaman): add v0.28 owner talent pass

ryan-foo:
## Summary

- replace the Shaman rows from #1980 with the agreed shared-row framework: movement at 5, survival at 8, control at 11, management at 14, power spike at 17, and build-defining capstones at 20
- add automatic specialization identities without training cards: Fulmination for Thundercall, Skyrend for Warspirit, and Cleansing Tides for Spiritmend
- give Warspirit a deliberate off-tank posture through health and armor, Stonebound-only threat, Elemental Demand, and optional survival talents
- add localized abilities, tooltips, explicit icons, implementation tests, and the PBE balance contract

## Dependency and review status

Part of #1980. This PR depends on #2168. Because both heads are on a fork, GitHub temporarily shows the shared-infrastructure commit here as well; once #2168 lands, this PR reduces to the Shaman-only commits.

This is a draft for design review and PBE iteration. It requires Levy or Fernando approval before landing under the current redesign freeze.

## Design boundaries

Warspirit remains a damage role rather than a fourth finder tank. Its extra threat only exists while Stonebound Weapon is active, and its durability comes from explicit baseline and talent tradeoffs rather than always-on tank mitigation.

The shared rows are usable by all three specializations and remain mobile-friendly. Levels 5–14 mostly alter existing actions; level 17 is a mutually exclusive active row; two of the three level-20 capstones are passive.

## Validation

- `npm run check:ts` — passed
- focused Shaman, proc, talent, localization, icon, and SFX-preflight suite — 182 passed, 3 skipped
- `tests/chronomancy.test.ts` — 12 passed after rebasing onto the current release branch
- i18n generation and freshness check — passed
- `npm run gate` — all 11 steps passed, including 16,552 unit/integration tests, 66 browser regressions, typechecks, server build, and client production build


EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants, ran the guard suites myself, and verified each finding against the real code paths rather than the PR description.

## What I ran

- `npx tsc --noEmit`: clean
- `tests/parity` (draw-order digest, not regenerated), `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`: pass
- `tests/localization_fixes.test.ts` (S3), `tests/i18n_completeness.test.ts` (M16): pass
- `tests/shaman_owner_pass.test.ts`, `tests/talent_procs.test.ts`, `tests/choice_rows_redesign.test.ts`, `tests/talents.test.ts`, `tests/spec_masteries.test.ts`, `tests/talent_tooltip_accuracy.test.ts`, `tests/talent_nonwarrior_runtime_v026.test.ts`, `tests/talent_retained_semantics_v026.test.ts`, `tests/combat_auto_attack.test.ts`: pass
- `npm run i18n:gen` then `git status`: clean, so the generated bundles match regeneration exactly

The shape of the work is good. The four new modules sit correctly behind `SimContext`, every new rng draw is gated so a player without the talent draws nothing, no tick phase was reordered, and `meleeSwing` stayed a thin delegate with an append-only seam widening. The design doc matches the implemented numbers (ward cap 9, 5% per charge, Skyrend 5 stacks / 30 sec, `rockbiterThreatPct: 1` doubling threat).

## Findings

### 1. Existing Shaman characters are locked out of all talent edits until they respec (high)

This PR removes seven Shaman option ids outright (`sha_elemental_attunement`, `sha_elemental_warding`, `sha_guiding_spirits`, `sha_r11_elemental_attunement`, `sha_r14_chain_lightning`, `sha_r14_improved_flame_shock`, `sha_ward_surge`) and, more subtly, **moves several surviving ids to a different row level** while keeping their old `rN_` prefix. For example `sha_r17_improved_ghost_wolf` was a level-17 pick on `release/v0.28.0` and is now a level-5 option; `sha_r11_healing_stream` moves to row 17; `sha_r5_concussion` moves to row 14.

Selections are persisted as `rows[level] = optionId`, so any pre-PR Shaman save now holds an id that no longer validates at its stored level. `commitTalentAllocation` validates the whole candidate allocation, and the candidate is a clone of the existing rows plus the one change, so the stale entry poisons every subsequent edit.

I confirmed this with a repro against the real `Sim`:

```ts
const sim = new Sim({ seed: 2800, playerClass: 'shaman', autoEquip: true });
sim.setPlayerLevel(20);
const meta = sim.players.get(sim.playerId)!;
// exactly what a save from release/v0.28.0 holds
meta.talents = { spec: 'elemental', rows: { 17: 'sha_r17_improved_ghost_wolf' } };

sim.selectTalentRow(5, 'sha_r5_improved_lightning_shield');
// -> false, and the player gets "Invalid talent build."
```

The player is not crashed or blocked from logging in (the dead row effect is silently skipped at recompute), but they cannot change **any** talent row, and saved loadouts carrying a stale id will fail to apply the same way. `sim.respec()` is the only escape hatch, and nothing tells the player that.

`src/sim/content/choice_rows.ts` already exports `repairRows`, which does exactly the right thing (drop only the entries that fail validation). It has **zero callers anywhere in `src/` or `server/`** and no test, so it is currently dead scaffolding. Wiring it into the character load path, plus a test pinning that a removed and a moved id both get dropped rather than poisoning the allocation, would close this. If instead the intent is a forced Shaman respec on upgrade, that needs to be explicit and covered by a test.

### 2. An interleaved instant wipes the in-flight Arc Bolt Skyrend snapshot (medium)

`src/sim/combat/casting_lifecycle.ts` around line 846:

```ts
const arcBoltStacks = warspiritArcBoltStacks(ctx, p, meta, ability.id);
if (arcBoltStacks > 0) p.warspiritArcBoltStacks = arcBoltStacks;
else clearWarspiritArcBoltSnapshot(p);
```

Neither this clear nor the one after the instant `applyAbility` (around line 942) is guarded on `blinkThrough`, which is the existing path where an instant resolves while a timed cast keeps running (it deliberately restores `heldCastAim` / `heldCastTarget` right after). Fire any instant mid-Arc-Bolt and the snapshot is deleted, so `consumeWarspiritArcBolt` falls back to recomputing from the live aura at completion. Auto-attacks land during the cast and Skyrend procs on landed hits, so the bolt can be boosted by more stacks than the ones that shortened its cast time. That is precisely the invariant the field comment in `types.ts` promises ("consumes the same bank that shortened its cast time").

Suggested check: cast Arc Bolt at 2 stacks, fire an off-GCD instant mid-cast while a swing lands, assert the completed bolt's `damageMult` is 1.2 and not 1.3.

### 3. Three new rng draw sites landed with no `tests/parity` scenario (medium)

`fulmination.ts` (`ctx.rng.chance`), `unleash_weapon.ts` (`ctx.rng.range` plus `ctx.rng.chance`), and the new `meleeHit` chance arm in `talent_procs.ts`. `src/sim/CLAUDE.md` step 5 asks for a parity scenario for a new mechanic with rng draws. The gate is green only because no parity scenario exercises a Shaman with these talents, so the draw-order digest is blind to all three. Adding a Fulmination / Warspirit scenario would make future edits to these sites actually catchable.

### 4. `stackAura` refresh bypasses `ctx.applyAura` (low to medium)

In `talent_procs.ts`, the create arm goes through `ctx.applyAura` but the refresh arm writes `existing.stacks` / `remaining` / `duration` directly and emits no `aura` event. That skips the shared aura-stacking, exclusive-aura, and classification paths for every stack after the first, and stacks 2 through 5 are silent for FCT and aura-gain feedback. Worth confirming `applyAura` genuinely is not required for `stormcharge` rather than leaving the two arms divergent.

### 5. Elemental Discharge mixes the melee and spell stat channels (low)

`unleash_weapon.ts` scales damage off `ctx.effectiveAttackPower` (melee) but rolls the crit and applies the crit multiplier through the spell channel (`ctx.spellCrit`, `critDmgSpellBonus`), including in the `school = 'physical'` no-imbue branch. So a physical-school discharge ignores `critChance` and `critDmgPhysBonus` entirely. Deterministic either way, but it does not look like the intended stat model, and the design doc does not pin it.

### 6. One untranslated English word in a talent tooltip (low)

`src/ui/talent_i18n.ts:9334`, the new `consumeAuraChargesDamage` arm ends with `x charge`. Every sibling arm in that switch routes its words through `text.statLabels.*` or `t()`; this one hardcodes English, so it stays English in all 21 locales. No gate catches it because it is inside a template literal rather than a catalog value. Adding a `charges` member to `TalentLocaleText['statLabels']` alongside `damage` would fix it. (The adjacent `r=` is fine, it matches the existing symbolic convention on the line above.)

### 7. `ResolvedAbility.damageMult` is honored in exactly one effect case (note)

`effect_dispatch.ts` applies `res.damageMult` only in the `directDamage` branch. Correct for Arc Bolt, but the field name and its comment in `sim.ts` promise a general per-cast multiplier, so a future ability pairing it with `dot`, `aoeDamage`, or `weaponStrike` would silently drop it. Either narrow the comment or apply it in the other damage cases.

## Things I checked that are fine

- **The `src/ui/i18n.locales/` overlay edits are not a violation.** The three new wordy English ability values are the M16 case that requires the five non-Latin fills in the same change, and the `healing_stream` 120 to 180 reword is the "staleness" case that `docs/i18n-scaling/translation-workflow.md` names as the one sanctioned reason to edit an overlay. Leaving those stale would have been the actual bug.
- **Ground AoEs do not drain Stone Aegis.** `pulseGroundAoE` defaults `direct = false`, so puddles and DoT ticks and reflects all correctly leave the charge bank intact. Multi-target `aoeDamage` fan-out does pass `direct = true` and will spend one charge per victim, which reads as defensible for "the next 6 direct attacks", but flagging it in case the intent was per-cast.
- **Skyrend only builds on landed hits.** `meleeSwing` returns early on miss, dodge, and parry before reaching `onMeleeSwing`, matching the design doc wording.
- **Stone Aegis sits in the right place in the damage pipeline**, alongside `shield_wall` and ahead of the vulnerability amps and absorb shields.
- No `Math.random` / `Date.now` / `performance.now`, no DOM or render imports in the new sim modules, no `ALLOW_DEV_COMMANDS`, no secrets, no em dashes or emojis in the diff.

Finding 1 is the one I would want resolved before this lands, since it hits every existing Shaman character on upgrade. The rest are follow-ups that could reasonably ride the PBE iteration this PR is already scoped for.

PR #2154, opened 2026-07-19
fix(mage): rebalance barriers, Icebind, and Frozen Orb

Blaine1705:
## Summary

- Scale Frostveil and Blazing Barrier through ranks at levels 5, 12, and 18 with 50, 90, and 130 base absorb, plus 50% of the caster's Spell Power at every rank.
- Scale Warded's break heal from the mage's maximum health so allied barriers cannot amplify it.
- Render Warded's source-health scaling correctly in generated localized tooltips instead of falling back to zero healing.
- Make Icebind break after cumulative post-mitigation damage equal to 15% of target maximum health, clamped from 20 to 60. Icebind's initial damage is excluded and other roots keep their previous behavior.
- Keep Icebind's initial damage, but prevent its damage-breakable AoE root from starting auto-attack and immediately shattering itself.
- Move Frozen Orb from level 12 to level 15 and raise its cooldown from 30 to 45 seconds, while preserving its damage profile and level-18 rank.
- Keep combat values, live tooltip scaling, locale overlays, generated localization bundles, and spell-rank documentation aligned.
- State Icebind's exact cumulative break budget in all 22 supported locales: 15% of target maximum health, clamped from 20 to 60 damage.
- Restrict `breakOnDamage` to direct non-stun AoE roots so ring, trap, and stun definitions cannot silently ignore it.
- Scale Temporal Barrier at every rank with 25% of the caster's Spell Power, including runtime absorb and live tooltip values.

## Type of change

- [ ] Feature: new functionality
- [x] Bug fix
- [ ] Refactor or performance (no behavior change)
- [x] Documentation
- [x] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npm.cmd exec -- vitest run tests/frost_mage_aoe.test.ts` (14 passed; also failed first against the previous level-12/30-second definition)
  - `npm.cmd exec -- vitest run tests/frost_mage_aoe.test.ts tests/frost_mage_procs.test.ts tests/mage_tuning_0713.test.ts tests/frost_nova_break.test.ts tests/guide.test.ts tests/architecture.test.ts` (130 passed)
  - `npm.cmd exec -- vitest run tests/attack_on_ability.test.ts tests/frost_nova_break.test.ts tests/frost_mage_aoe.test.ts tests/architecture.test.ts` (57 passed)
[...]

PR #2170, opened 2026-07-19
feat(loot): true scarcity for raid-boss and major-loot-event drops

awidearray:
## Summary

Introduces **limited relics**: raid-boss and major-loot-event drops with a hard, globally finite supply. Every dropped copy mints a 1-based serial number ("Emberfall Edge, relic 7 of 25"); once the cap is spent, the relic never drops again on any realm, and a registered fallback substitutes. Seven relics ship: the legendary **Emberfall Edge** (heroic Nythraxis, cap 25), the **Crown of the Thornpeak Scourge** (100), **Thunzharr's Stormheart** (world boss, 150), and four **Sealed Vault Signets** (heroic dungeons, 200 each).

Design doc: `PLAN_TRUE_SCARCITY_DROPS.md`.

### How it works (one sim, three hosts)

- **Sim/content:** a `claimLimitedSerial` host seam and an append-only `appendLimitedDrops` roll phase that leaves the existing loot loops byte-identical. The claim draws **zero rng**, so determinism and the parity goldens hold; the serial rides `ItemInstancePayload` through every distribution path (need/greed, master loot, round-robin, personal, return-to-corpse).
- **Server:** a cross-realm Postgres serial ledger where a serial is unique per `(item_id, serial)` and the cursor never passes the cap, so the hard cap holds across every realm process. A small in-memory buffer bridges the synchronous in-tick claim to the async ledger; a serial is never re-issued once claimed (uniqueness is the load-bearing promise). Plus a realm-wide claim broadcast and a public `GET /api/limited-mints` ledger endpoint.
- **UI:** tooltip scarcity lines, an offline mint chat line with the item as a colored link, and localized realm broadcasts (English + the five non-Latin M16 fills).

## Type of change

## How was this tested?

- **Commands:** `npm run gate` (green except one pre-existing failure, see below); `npx vitest run tests/limited_drops.test.ts tests/limited_drops_content.test.ts tests/server/limited_supply.test.ts tests/server/limited_routes.test.ts tests/item_level.test.ts tests/combat_rating.test.ts tests/weapon_skins.test.ts tests/parity`.
- **Real Postgres:** `TEST_DATABASE_URL=... npx vitest run tests/limited_supply_db_integration.test.ts` runs the actual ledger SQL against a live PostgreSQL, including an 8-realm concurrent-lease race that proves no duplicate serial is ever issued. Verified green against a live database (skips when no `TEST_DATABASE_URL`).
- **Manual (offline client):** granted a plain crown and a serialed Emberfall Edge and fired a mint event; verified the tooltip static line ("Limited relic: only 100 will ever exist"), the serial line ("Relic 7 of 25"), and the chat line ("Relictester has claimed [Emberfall Edge], relic 7 of 25.") all render, with the relic carrying its budgeted legendary stats and fire proc.

## Pre-existing failure (not from this change)

`tests/chronomancy.test.ts` ("Temporal Barrier ... 194 vs 184") fails on the base branch: its last merge (`fix/mage-barrier-scaling`, #2154) changed the barrier scaling and left this test stale. This diff touches zero mage/spell code, so the failure is independent of this PR.

## Notes / accepted tradeoffs

- Relics are not market-tradable yet (they reuse the existing instanced-item market block); tradability is a follow-up.
- The offline world's ledger is per-session (offline is a private sandbox; authoritative scarcity is the online ledger).
- An ungraceful crash or an unlooted drop orphans a serial as `leased` forever (effective supply erodes slightly downward, always the safe direction, never a duplicate). Recovery is a documented manual ops step, deliberately not automated, because auto-reclaiming a `leased` row could re-issue a serial a player holds.


EnriqueGF (changes requested):
Reviewed the full diff against the repo invariants and ran the gates locally. This is a genuinely careful piece of work: the append-only roll phase, the zero-rng claim seam, the PK-backed ledger and the documented erode-downward-never-duplicate stance are all the right calls, and the real-Postgres integration test with the 8-realm race is exactly the evidence this feature needed.

**What I ran** (worktree off `release/v0.28.0`): `npx tsc --noEmit` clean; `tests/architecture.test.ts`, `tests/localization_fixes.test.ts` (after `npm run i18n:gen`), `tests/i18n_completeness.test.ts`, `tests/parity`, `tests/world_api_parity.test.ts`, `tests/sim_context.test.ts`, `tests/world_boss.test.ts`, `tests/server_i18n.test.ts`, `tests/server/http/completeness.test.ts`, and all four new suites: green. The six failures I hit on the first pass were all 5s vitest timeouts under core contention, not real; they pass at `--testTimeout=60000`. I also verified the regenerated `nythraxis_full_pull.json` golden is a legitimate tail shift (`+1` draw at the death frame, `state` hash unchanged, `emberfall_edge` correctly skipped draw-free on normal).

Determinism, sim purity, the `SimContext` append, the `IWorld` seam (untouched, correctly), i18n render-sink classification, server authority and dev-command/secret handling all check out. Three things I would fix before merge.

---

**1. The fallback bypasses the `awardedItemIds` cross-group duplicate guard.** `src/sim/loot/loot_roll.ts` (`appendLimitedDrops`)

Every registered fallback is an item that already exists in the same boss's own table. `soulflame_cowl` is in `content/dungeons.ts` twice (chance 0.14, roll groups `nythraxis_drop_1` and `nythraxis_drop_3`); `deathless_greatblade` is in `heroic_loot.ts` (chance 0.34). `awardedItemIds` exists precisely to stop one kill handing out two copies of the same epic, but the appended limited phase never consults it. Once `crown_of_the_thornpeak_scourge` is capped out, a single Nythraxis kill can drop two `soulflame_cowl`. Feeding `drop.itemId` through `awardedItemIds` (and skipping the push on a repeat) in `appendLimitedDrops` covers it, plus a test that exhausts the crown and forces both wins.

**2. The fallback slot is never routed through `heroicItem()`.** `src/sim/loot/loot_roll.ts`, `makeSlot(drop.itemId)`

On a heroic claim every other drop from that corpse upgrades to its Heroic variant, but the consolation drops as the base item. I checked which fallbacks actually have variants: `soulflame_cowl -> heroic_soulflame_cowl` (ilvl 33 vs 29). So on heroic Nythraxis with the crown exhausted, the consolation is a full tier below what the same item would have been from the normal table on the same kill. The other five fallbacks are already heroic-tier items with no variant, so this is narrow, but it contradicts the "comparable, tier-appropriate reward" contract in the `LIMITED_FALLBACK` comment.

**3. The serial buffer never re-warms after a transient lease failure.** `server/limited_supply.ts`, `claim()`

`claim()` returns early on an empty pool without scheduling a refill, so the only refill triggers are `init()` and a *successful* claim. If one `fillBuffer` rejects, `enqueue`'s catch logs and swallows it, the pool stays empty, and every subsequent claim returns `null` forever: the relic silently stops minting on that realm for the process lifetime and players get the fallback instead. I confirmed this with a scratch test (flaky fake db, fail one refill, heal the db, claim again -> still `null`). The same shape bites cross-realm: a realm whose pool emptied on a genuinely exhausted supply never retries after another realm gracefully releases serials back. Scheduling a refill on the empty-pool path (or a bounded retry in `enqueue`) fixes it; worth a test either way, since the current suite covers the happy path and the crash/reboot case but not a failed refill.

---

Smaller things, none blocking:

- `server/limited_routes.ts` `readCached` has no single-flight, so every 30s expiry lets all concurrent requests miss at once and each takes a client from the 10-client pool the game loop's autosave shares. The repo already has `singleFlight` (used for `refreshDeedsBoardShared` in `main.ts`). Worth wrapping given this is advertised as a hot public endpoint.
- `server/limited_supply_db.ts` `leaseSerial`: the `await client.query('ROLLBACK')` in the catch is unguarded, unlike the `.catch(() => {})` convention in `ip_block_db.ts` / `social_db.ts` / `maps_db.ts`. A rejecting ROLLBACK masks the original error and returns a client that may still hold the `limited_item_supply` row lock.
- Same function, the reclaim `UPDATE`: the outer `WHERE` does not re-assert `state = 'released'`. Correct today via the `FOR UPDATE` EPQ recheck, but that is a subtle thing to rest the feature's core safety promise on. Adding the predicate is free.
- Post-cap, the relic roll does not stop, it permanently becomes an extra unconditional drop chance for the fallback epic (10% extra `soulflame_cowl` per Nythraxis kill, 6% per heroic dungeon final boss). I read that as intended, but it is a lasting change to non-limited drop rates that nothing pins.
- The boot pre-lease is per process, not per cluster. `bufferTarget = 1` bounds it correctly for one realm, but with 5 realms that is 5 of the 25 Emberfall serials parked in idle buffers immediately, and each ungraceful restart strands them permanently (by design, never auto-reclaimed). Since `remaining` subtracts `leased`, the erosion is publicly visible. A metric for leased-but-unclaimed per item would at least make it observable.
- `seedSupply`'s `ON CONFLICT DO NOTHING` silently ignores a later cap change in content, and the item just reads as exhausted with no error. A warning when the DB supply differs from the content cap would save someone a confusing afternoon.
- The offline `Sim.limitedMints` map is never serialized, so a reloaded offline world re-mints from serial 1 while the saved character still holds "relic 1 of 25". The PR notes the ledger is per-session; being explicit that this means duplicate serials within one save would be clearer for whoever reads it next.

Happy to re-review once 1 to 3 are addressed. Nice feature.

awidearray:
Thanks for the review, this was a genuinely useful pass. All three blocking items are fixed and the branch is rebased onto current `release/v0.28.0`.

Continued in **#2263** rather than force-pushing here, so this thread and your comments stay intact.

**1. Fallback bypassed `awardedItemIds`.** Confirmed real: `soulflame_cowl` is in both `nythraxis_drop_1` and `nythraxis_drop_3`, so a capped crown really could drop two from one kill. Fixed, with a test that exhausts the crown and forces the collision.

**2. Fallback not routed through `heroicItem()`.** Fixed.

**3. Buffer never re-warmed.** Fixed by scheduling a refill from the empty-pool path. Your scratch-test finding reproduced exactly. I added a per-item pending flag so the retry cannot pile up one doomed lease per kill against a genuinely exhausted supply, and a test for the cross-realm case you mentioned (a realm picking up serials another realm released).

**One thing worth surfacing, because it inverts the fix for half the cases.** `drop.instance` distinguishes a minted relic from a consolation, and a minted relic must obey **neither** new rule. Its serial is already spent by the time `appendLimitedDrops` sees it, so skipping the push would burn one of a hard-capped supply and grant nothing, which is worse than the duplicate it would prevent. Minted drops take a separate path now, with a test pinning that the serial is still handed over even when both its id and its fallback are already in the awarded set. All new branches sit after the `rng.chance` draw, so draw order is untouched.

**On the rebase:** the base had moved 85 commits. The parity golden passes **unregenerated**, so those commits genuinely did not disturb the draw order, and no `UPDATE_PARITY=1` was used to force it. `chronomancy.test.ts` is green on the new base, so the inherited red is gone.

**The rebase also caught a real bug neither of us had spotted.** The base added `held_weapon_models.test.ts`, and `emberfall_edge` had no `ITEM_WEAPON_VARIANTS` entry, so the legendary relic would have rendered as an empty hand. Fixed in #2263.

Full gate on the new branch: all 11 steps green, 16,839 tests, 0 failures.

I did not take the smaller `limited_routes.ts` / `limited_supply_db.ts` items (single-flight, the unguarded `ROLLBACK`, re-asserting `state = 'released'`) because those files are being touched by a parallel change and would collide. Say the word if you would rather they ship in #2263 and I will fold them in.

PR #2172, opened 2026-07-19
feat(sfx): give the forge ambience a real, seam-safe recording

jamiecypher:
## Summary

Replaces the generated amb_forge.mp3 placeholder (positional point-ambience near smithy stalls, e.g. Smith Haldren in Eastbrook Vale and Armorer Hode in Highwatch) with a real custom recording. Players reported the old clip clicked audibly every ~8s at its loop point; the runtime plays ambience via a single WebAudio AudioBufferSourceNode with native loop=true and no crossfade of its own, so a clean loop point has to be baked into the source file. The new recording deliberately breaks its hammer-strike rhythm (non-metronomic loop period) and pads 375ms of true digital silence on each side after the strike decay rings out, well past MP3 encoder priming artifacts.

Also fixes a real pipeline bug found in the process: the catalog entry (scripts/sfx/sfx_prompts.mjs) was missing `custom: true`, so conform's LUFS-retargeting branch was silently degrading the recording on every --fix pass instead of preserving its authored mix. Flagged correctly now. The recording's own true peak (-6.0 dBTP after the engine-wide -6dBFS conform ceiling, which applies to every SFX asset regardless of custom flag) still left it perceptually quiet in-game, so a combined fix uses the sanctioned per-key gain-map ceiling (`keyTrimDb.amb_forge: 5`, exactly at its computed maximum) plus a small forge-specific runtime mix constant (`FORGE_AMBIENCE_GAIN`, split from the shared `POINT_AMBIENCE_GAIN` so campfire is unaffected), for a combined ~1.11x effective gain tuned by ear.

## Type of change

- [x] Bug fix (the audible loop click, plus the conform pipeline mistreating a real recording as generated content)
- [x] Feature: new functionality (real forge ambience recording)

## How was this tested?

- Commands: `npx tsc --noEmit`, `npx @biomejs/biome ci --changed --since=<merge-base>`, `npm run sfx:check`, `npx vitest run` (full suite: 1364 files / 16539 tests, 0 failures), `npm run build`
- QA gate: ran the qa-checklist review twice (once mid-investigation, once final after the custom-flag/gain-map fix), verdict READY both times. The reviews independently re-derived and verified the -6dBFS vs -1dBFS constant distinction, the keyTrimDb ceiling arithmetic, the combined-gain headroom math (post-gain peak stays around -5dB, comfortably under 0dBFS), and confirmed the loop() mechanism has no seam-crossfade of its own so the silence-padding fix is the correct approach.
- Manual steps: listened in-game at both smithy locations across multiple gain iterations to tune the final loudness by ear.


EnriqueGF (commented):
Reviewed the full diff, the SFX pipeline modules it leans on (`sfx_gain_ceiling.mjs`, `conform_audio.mjs`, `playback_profile.mjs`), and `loop()`'s gain path in `src/game/sfx.ts`. I re-measured the audio myself rather than taking the PR body's numbers on faith.

**What I verified independently**

- **Ceiling arithmetic is right.** Measured the new `amb_forge.mp3` true peak with `measureSfxTruePeakDb` and the bundled ffmpeg: exactly `-6.0` dBTP. `SAFETY_FLOOR_DBFS (-1) - (-6) = 5`, so `keyTrimDb.amb_forge: 5` and the generated ceiling entry both check out, and the generated JSON matches what the tool would compute (not hand-edited).
- **No clipping.** `mixedTarget = 0.625 * 1.778279 = 1.111` against a source peaking at `-6` dBFS lands at about `-5.1` dBFS pre-panner. Confirms the headroom claim.
- **The loop-seam fix is real.** `volumedetect` over the first and last 300ms gives `-91.0` dB and `-90.3` dB, i.e. true digital silence on both sides of the wrap, past MP3 priming. The click cause is genuinely addressed at the source, which is the right call given `loop()` has no seam crossfade.
- **It is also not a step backwards on density.** Worth stating since the gain chase across five commits could read as compensating for a sparse clip: sub `-50` dB silence is 2.40s of 14.29s (17%) versus 3.04s of 6s (50%) in the clip it replaces, and integrated loudness goes `-32.6` to `-27.1` LUFS. Denser and louder than before.
- **Checks run:** `npx tsc --noEmit` clean; `npx vitest run` over `sfx_conform`, `sfx_gain_ceiling`, `sfx_manifest`, `sfx_runtime_pack`, `sfx`, `sfx_playback_profile`, `architecture` (148 passed); biome clean on both changed source files; no em/en dashes or emojis in the diff. The `custom: true` fix is correct, `conformCustomMaster` is the branch you want for a hand-mastered recording.

**Findings**

1. *(minor, please address)* **The new per-kind gain split has no test.** `pointAmbient`'s `source.kind === 'forge' ? FORGE_AMBIENCE_GAIN : POINT_AMBIENCE_GAIN` is new branching behavior and nothing asserts it. `tests/sfx_loading.test.ts` exercises a forge point source but only for load/cancel, never the resolved gain, and no test covers a campfire point at all. So the thing this PR most cares about (forge gets its own mix, campfire is unaffected) would survive being reverted or mis-wired silently. The existing `internals(player)` helper already reaches `slot.target`, so this is a cheap test: one forge point and one campfire point, assert `1.111...` and `0.18` respectively. That also pins the "campfire is unaffected" promise from the PR body.

2. *(note, no action needed now)* `keyTrimDb.amb_forge: 5` sits **exactly** at the computed ceiling, zero slack. That is legal and deliberate, but it means any future re-export of this recording at even a hair hotter peak lowers the ceiling below 5 and hard-fails the gate with a message about the gain map rather than about the new file. A one-line note in the `sfx_prompts.mjs` comment saying "re-exports must hold `-6.0` dBTP or the trim needs lowering with them" would save the next person the detour.

3. *(observation, pre-existing, not yours)* The clip is anvil strikes over silence with no continuous furnace bed, which is a fair distance from the catalog prompt's "roaring furnace". The clip it replaces had the same gap, so this is not a regression and I would not hold the PR for it, but it is likely the real reason the level needed a ~6x-campfire mix target to read at all. Worth a follow-up if the forge ever still feels thin in-game.

Nothing blocking on correctness. Happy to approve once the gain-split test lands.

FernandoX7 (commented):
Verified this against the artifacts rather than the description, and it holds up. I measured the committed mp3 myself: mono, 44.1 kHz, 14.3s, true peak exactly -6.0 dBFS, with real digital silence at both ends (about 0.4s each at a -60dB threshold), so the loop wrap is inaudible by construction. The gain arithmetic is consistent end to end: the ceiling is -1 - (-6.0) = 5 and matches sfx_gain_ceiling.generated.json, keyTrimDb sits exactly at that ceiling, the pack gain 1.778279 is 10^(5/20), and loop()'s mixedTarget (src/game/sfx.ts:644) really does stack it with FORGE_AMBIENCE_GAIN for the stated ~1.11x combined gain, with campfire untouched on POINT_AMBIENCE_GAIN.

I also re-ran the pipeline at the PR head: `node scripts/build_sfx_manifest.mjs` produces zero drift against the three committed generated artifacts, and `npm run sfx:check` passes with amb_forge ok (the three loudness advisories in its output are pre-existing spider/troll files this PR does not touch). The base-vs-head diff of the generated files is exactly the amb_forge entries plus bundleId, so the manifest fix in the last commit landed correctly, and catalogHash staying unchanged is right: it only hashes loop/category/preload/spatial, not prompt/duration/custom. The custom flag turns out to be doubly load-bearing: beyond the preserveLoudness conform branch, gen_sfx.mjs skips custom tracks even under --force (scripts/gen_sfx.mjs:73), so the recording can no longer be clobbered by a generation run. tsc is clean and the targeted suites (sfx_conform, sfx_loading, sfx_manifest: 111 tests) pass at the head.

Findings:

- nit: the new forge/campfire gain split (src/game/sfx.ts:756) has no test pin. tests/sfx_loading.test.ts already drives point-ambience loops (the campfire reuse case around line 755), so a small assertion that a forge point's loop target lands at FORGE_AMBIENCE_GAIN times the manifest entry gain while campfire stays at POINT_AMBIENCE_GAIN would keep the split from silently regressing. Not blocking, src/game is outside the mandatory-test scope.
- nit: the catalog entry says duration: 14 but the file is 14.29s (scripts/sfx/sfx_prompts.mjs). Duration only feeds generation, which now skips this key, so it is documentation only; worth syncing next time the entry is touched.

One process note, not a code finding: CI has not run on this branch (no checks reported, the PR shows UNSTABLE), which looks like the usual fork workflow-approval gate rather than a failure. It needs a green run before merge.

PR #2175, opened 2026-07-19
fix(sfx): bring Thornpeak Heights wind ambience down to its siblings

jamiecypher:
## Summary

Based on direct player feedback in-game: the ambient wind at Thornpeak Heights (Zone 3, biome 'peaks', which also covers the 'desert' and 'volcano' biomes since they share the same bed) was reported as noticeably louder than the other zones' wind ambience.

Investigation found the actual audio assets are not the cause: amb_wind_peaks.mp3, amb_wind_vale.mp3, and amb_wind_marsh.mp3 are all ElevenLabs-generated content, loudness-matched to within 0.2 LU of each other (all around -14.5 to -14.7 LUFS). The gap was entirely a runtime mix-target constant in src/game/sfx.ts's ambience(): vale=0.12, marsh=0.13, peaks=0.18, with peaks the clear outlier at +3.5dB over vale for no asset-level reason.

Brings peaks/desert/volcano's mix target down to 0.12, exactly matching vale. Landed there after in-game A/B testing across three values (0.18 original, 0.13 intermediate, 0.12 final) by ear.

## Type of change

- [x] Bug fix

## How was this tested?

- Commands: `npx tsc --noEmit`, `npx @biomejs/biome ci --changed --since=<merge-base>`, `npm run sfx:check`, `npx vitest run` (full suite: 1371 files / 16613 tests, 0 failures), `npm run build`
- QA gate: ran the qa-checklist review twice (once via a manual read of the actual instructions, once via the real registered agent), verdict READY both times, 0 blocking / 0 should-fix. Independently verified the loudness-matching claim across all three wind beds via ffmpeg, confirmed 0.12 exactly matches vale's own existing constant (not an arbitrary pick), and confirmed no other ambience bed or unrelated code was touched.
- Manual steps: listened in-game at Thornpeak Heights across all three tested gain values before landing on 0.12.

## Notes for the maintainer

Two non-blocking design observations surfaced during review, worth a look but not gating this fix:
- Flattening peaks to vale's mix level slightly undersells the zone's "cold howling, gusty" generation intent versus vale's gentle breeze; if that character loss reads as a problem in practice, the fix would be re-authoring the peaks wind clip itself to feel appropriately harsh at a lower gain, not raising the constant back up.
- peaks/desert/volcano currently share one wind bed and one mix constant; this is pre-existing (not introduced or worsened here), but if desert/volcano ever want distinct ambient character, this shared constant is the seam that would need splitting.

### DIFF
--- a/src/game/sfx.ts
+++ b/src/game/sfx.ts
@@ -768,13 +768,13 @@ class Sfx {
     );
     this.ambient(
       'amb_wind_peaks',
-      // 0.13 matches amb_wind_marsh's mix target; amb_wind_peaks/vale/marsh are
+      // 0.12 matches amb_wind_vale's mix target; amb_wind_peaks/vale/marsh are
       // all generated to the same -14ish LUFS target with no custom mastering
       // (confirmed by measurement), so there is no asset-level reason for
       // peaks to sit louder than its siblings. The old 0.18 was a player-
       // reported "too loud" complaint at Thornpeak Heights (this biome also
       // covers desert/volcano) traced to this constant alone, not the file.
-      !inDungeon && (biome === 'peaks' || biome === 'desert' || biome === 'volcano') ? 0.13 : 0,
+      !inDungeon && (biome === 'peaks' || biome === 'desert' || biome === 'volcano') ? 0.12 : 0,
     );
     this.ambient('amb_rain', precip === 'rain' ? 0.11 : 0); // sharp clip, kept very low
     this.ambient('amb_snow', precip === 'snow' ? 0.13 : 0);
--- a/src/game/sfx.ts
+++ b/src/game/sfx.ts
@@ -768,7 +768,13 @@ class Sfx {
     );
     this.ambient(
       'amb_wind_peaks',
-      !inDungeon && (biome === 'peaks' || biome === 'desert' || biome === 'volcano') ? 0.18 : 0,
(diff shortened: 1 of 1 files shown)


FernandoX7 (commented):
Verified the load-bearing claim independently: ffmpeg loudnorm on the three beds gives vale -14.50, marsh -14.69, peaks -14.72 LUFS integrated, so the assets really are matched (peaks is in fact the quietest of the three) and the old 0.18 was purely a mix-constant outlier. The fix also sits in the right layer: the wind beds have no entries in scripts/sfx/sfx_gain_map.json, so the runtime targets in ambience() are where this tuning lives, and 0.12 is vale's existing constant rather than a new number. Merges clean against release/v0.29.0.

One finding:

- nit, src/game/sfx.ts:771: the six-line comment is mostly change history (the player report, the measurement, the old value), which the two commit bodies already carry; the convention for these constants in this file is a one-line intent note (see amb_rain's "sharp clip, kept very low"). Something like "0.12 matches amb_wind_vale; the three wind beds are loudness-matched assets" keeps the useful constraint and drops the provenance.

Notes:

- CI has not run on this branch yet (fork PR, no checks reported), so it needs a workflow approval before merge; with tsc, the full vitest run, and the build reported green locally, I expect no surprises on a one-constant diff.
- On the character-loss observation: agreed that if peaks reads too tame at 0.12, the fix is re-authoring the clip to feel harsh at the same LUFS target, not raising the constant back up. Fine to leave as is for now.

PR #2176, opened 2026-07-20
fix(hud): clear party frames below the target buff strip

seanghods:
## Summary

Selecting a target drops the party frames below the target frame by a hardcoded offset (96px desktop, +135px/+105px mobile) that clears only the target frame BODY. The target's `#tf-debuffs` strip hangs below the frame and grows about 28px per wrapped row, so a heavily buffed or debuffed target overlapped the first party frame: 65px of overlap at 8 auras and 103px at 18 on desktop, 36px on mobile at 28.

The offset now derives from the measured target stack:

- A pure core (`src/ui/party_below_target_core.ts`, registered in `UI_PURE_CORES`) computes the author-space bottom of the frame + strip union, gated on horizontal overlap with the party column (a dragged-away target frame no longer pushes the frames for no reason).
- A thin painter (`src/ui/party_below_target_painter.ts`, registered in `HOT_PAINTERS`) measures only when a cheap non-layout invalidation key or a ResizeObserver size epoch changes (a cast bar growing the frame re-measures next frame), and writes `--party-below-target-bottom` through the elided `setStyleProp` facet. Steady state does zero layout reads and zero DOM writes.
- The desktop rule and both mobile tiers derive `top` from the var; the old constants survive only as pre-measure fallbacks.

Two adjacent defects in the same stack are fixed with it:

- **Hidden buff timers:** the absolute `.buff .dur` labels hang 13px below each icon, but the strip's wrap gap was 4px, so any wrapped row painted over the previous row's timers. The strip now reserves the label height between rows (`gap: 17px 4px`) and pads its last row into its own box, which also makes the measured bottom cover the timers.
- **Mobile joystick coverage:** the pushed-down member rows could sit on the move joystick and steal its touches. The painter writes two measured sensors (`--party-rows-top` and `--party-rows-limit`, the limit preferring the resting wheel and falling back to the static capture zone while the wheel is mid-drag), and one tier-free rule bounds the rows 8px above the pad with a one-row 40px touch-floor minimum; overflowing members stay reachable by scroll. With an unwrapped strip the full 2x2 party fits above the wheel; heavy wrap degrades to one row + scroll.
- **Desktop off-screen rosters (second commit):** a 10-member raid below a heavily debuffed target ran off the bottom of a 720px-tall window (rows bottom 758 of 720 at 28 auras), a case the old constant kept on screen by overlapping the strip instead. The same sensors now run on desktop with the viewport bottom as the limit, and a desktop rule caps the rows 12px above it with overflow scroll (measured after: rows bottom 708 of 720).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npm run gate` locally. All stages green; the full vitest suite passes except `tests/deploy_watchdog.test.ts` (2 docker-timeout tests), which fails identically on the pristine `release/v0.28.0` in my environment, so it is pre-existing and unrelated (verified by running it in a clean base worktree).
  - `npx vitest run --exclude tests/deploy_watchdog.test.ts`: 1371 files, 16603 tests passed.
  - `npx tsc --noEmit` clean; biome clean on changed files; pre-push floor green.
  - Decisiveness: with the fix stashed, the new CSS pins and the offset tests go red (3 failures) against the release CSS.
- Manual steps (objective measurement, offline client + puppeteer): boot offline, build a real 5-member party through the sim, target a staged mob, apply N auras via `sim.applyAura`, and compare `getBoundingClientRect` values: the frame + strip union bottom vs the first `.party-frame` top, and `.party-rows` bottom vs the joystick top.

Desktop 1600x900, clearance between the strip bottom and the first party frame (negative = overlap):

| Target auras | Before | After |
|---|---|---|
| 0 | +7px | +8px |
| 8 | -65px | +8px |
| 18 | -103px | +8px |
| 28 | not measured | +8px |

At UI Scale 1.25 the visual clearance is a constant 10px (the 8px author gap scaled), at every count.

Desktop 1280x720 with a 10-member raid (rows bottom vs the 720px viewport edge):

| Target auras | Before this PR | First commit | With the desktop bound |
|---|---|---|---|
| 18 | on screen, strip overlapped the frames | 656 | 656 (unbounded case unchanged) |
| 28 | on screen, strip overlapped the frames | 758 (off screen) | 708, scrolls |

Mobile 844x390 landscape, `.party-rows` bottom vs the resting wheel top (negative = on the wheel):

| Target auras | Before | After |
|---|---|---|
| 0 | frames overlapped by strip at high counts; wheel overlap up to -30px | +33.5px (full 2x2 party visible) |
| 8 | -20.8px | +8px (one row + scroll) |
| 18 | -30px | -8.1px (40px one-row floor engaged) |
| 28 | -30px (and the strip overlapped the frames) | -60px (floor; wheel center and zone below stay live) |

## Screenshots / recordings

Committed under `docs/screenshots/party-below-target/`, embedded here from the branch commit.

**Desktop, 18 target auras.** Before: the strip buries the first party frames (and the wrapped rows hide the buff timers). After: frames clear the strip with a constant 8px gap and every row's timers are visible.

| Before | After |
|---|---|
|  |  |

**Mobile landscape.** Before (18 auras): the strip overlaps the chip and frames. After, common case (6 auras): the full 2x2 party fits above the move joystick. After, degraded case (18 auras): one row + scroll, joystick still usable.

| Before (18) | After (6, common) | After (18, degraded) |
|---|---|---|
|  |  |  |

---

### Quality

- [x] **The full gate passes.** (One pre-existing, environment-specific exception documented above: `tests/deploy_watchdog.test.ts` fails identically on the clean release branch.)

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop 1600x900 and landscape phone metrics (in-game mobile is landscape-only on the web client); both mobile CSS tiers updated and pinned. Touch targets and input font floors untouched.
- [x] **Accessible.** No new interactive elements; the 40px touch floor is used as the rows' minimum visible height.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or `.env` committed; no dev-command changes.


jgyy (commented):
This is a solid fix. The root cause analysis is right: the old offset only cleared the target frame body, not the `#tf-debuffs` strip that hangs below it and grows with wrap, so heavy buff/debuff counts overlapped the party frames. Replacing the hand-tuned constant with a measured stack bottom is the correct fix, and the two adjacent defects (hidden buff timers, mobile joystick coverage) are legitimately part of the same root cause rather than scope creep.

- `party_below_target_core.ts` is properly pure (no DOM), registered in `UI_PURE_CORES`, and the horizontal-overlap gate (`src/ui/party_below_target_core.ts:66-68`) correctly avoids pushing the frames down when a dragged target frame no longer overlaps the party column.
- `party_below_target_painter.ts` is registered in `HOT_PAINTERS` with an accurate `reflowAllow: { '.getBoundingClientRect': 5 }`, and the invalidation key genuinely avoids layout reads in steady state (confirmed via the gating tests in `tests/party_below_target_core.test.ts`).
- `#mobile-move-joystick` and `#mobile-move-zone` exist in both `index.html` and `play.html`, so the play.html chrome-parity trap doesn't apply here.
- The CSS var fallbacks reproduce the exact old constants (88px, safe-area+127px, safe-area+97px) so pre-measure paint matches today's layout, and `tests/party_below_target_core.test.ts` pins both the CSS and the old values' absence.
- Tests are decisive: the PR description notes the fix stashed makes 3 pins go red, which is the right bar.

One minor note, not blocking:

- `nit`: `src/ui/party_below_target_painter.ts`, the `measureRowsBound` comment block is long relative to the code it documents. Not asking for a change, just an observation.

No blocking findings. Nice work tightening the loop between the measured strip height and the joystick-clearing bound on mobile, that's the kind of fix that's easy to get "mostly right" and this handles the resting-vs-floating wheel distinction correctly.

seanghods:
Follow-up commit 2054e1dd: on shorter desktop windows, a long roster below a heavily debuffed target could push the last party frames off the bottom of the screen (measured: a 10-member raid at 28 target auras reached 758px on a 720px window). Fixed by capping the member-rows container at the viewport bottom with overflow scroll, reusing the mobile bound's measured sensors (property renamed --party-move-zone-top to --party-rows-limit since it is no longer always the move zone). Frames are not resized, and mobile behavior is unchanged; the description's tables are updated.

^tldr: edge case of shorter screens with full debuffs and 10 raid members, enabled scroll for overflow

EnriqueGF (changes requested):
Reviewed the full diff, the local CLAUDE.md files for `src/ui/` and `src/styles/`, and ran the checks myself.

**What I ran (all green):** `npx tsc --noEmit`; `npx vitest run tests/party_below_target_core.test.ts tests/architecture.test.ts tests/hud_perf_budget.test.ts tests/client_shell.test.ts` (167 passed, 3 skipped); biome on the changed TS files (warnings only, all pre-existing debt in `hud.ts`); dash/emoji scan on the changed files (clean, the one `hud.css` hit is pre-existing at line 4987 and outside the diff).

**What I like.** The shape is right: pure core in `UI_PURE_CORES`, thin painter on the `PainterHost` seam with an honest `reflowAllow` entry, policy arithmetic left in the stylesheet and only sensing in JS. The invalidation key is well chosen, and the two extra hazards you found (the `.buff .dur` overhang eating wrapped rows, the off-screen 10-raid) are genuinely the right things to fix alongside.

I also chased two things I expected to be bugs and confirmed they are **not**, so noting them so nobody re-opens them:

- The container's width is not in the key, and `#party-frames` really is zero-width out of a party (`party_frames_painter.ts` `clear()` detaches the rows wrapper; `> .tf-move-btn` is `position: absolute` and `display: none` without `.party-present`). So "select a target, then join a party" looked like it would latch the frames un-pushed forever. It does not: `sync()` toggles `party-present` on the container one frame later, and that class attribute *is* in the key, so it re-measures on frame N+2. I reproduced both arms against the real painter to be sure. Worth a line of comment in `buildKey` saying `party-present` is load-bearing for exactly this, since it is currently keyed only "for the elite/boss frame variants".
- `--party-frame-width` / `-columns` / `-scale` are written to `documentElement.style` (`src/main.ts:1916-1928`), which the key reads. Settings changes re-measure correctly.

### Findings

**1. [should fix] `overflow: auto` on desktop clips the panel keyline and the focus ring on every party frame, whenever a target is selected.** `src/styles/hud.css:4423`

The grid column is exactly `--party-frame-width` (`hud.css:4393`), `.party-frame` is exactly the same width (`hud.css:4428`), and `.party-rows` has no padding, so the scroll container's padding box coincides pixel-for-pixel with each frame's left and right border edges. Two things are painted outside that box and now get cut:

- `.panel` carries `outline: 1px solid #000` and `box-shadow: 0 2px 16px #000c` (`src/styles/base.css:490,492`), and rows are `class="party-frame panel"` (`party_frame_row.ts:161`). The classic black keyline separating the frames from the 3D scene disappears on both vertical edges.
- Rows are focusable (`row.tabIndex = 0`, `party_frame_row.ts:166`) and the global ring is `outline: 3px ... outline-offset: 2px` (`src/styles/shell.css:4098-4102`), i.e. 5px outside the border box. A tab-focused party frame shows only top and bottom ring segments.

Both are unconditional: they trigger whenever `below-target` is on, not only when the roster actually overflows, so the common 4-member-party case regresses visually with no scrolling involved. `tests/focus_visible_guard.test.ts` cannot see this because it is a geometry effect, not a missing declaration. Note this clipping already exists on mobile (that tier had `overflow: auto` before this PR), so it is only new on desktop.

Suggested fix, keeping the scroll behaviour identical: `padding: 6px; margin: -6px;` on `#party-frames.below-target .party-rows`, with the `max-height` calc adjusted for the added padding.

**2. [note] The move wheel's `floating` class is not in the invalidation key.** `party_below_target_painter.ts` `buildKey`

`measureRowsBound` picks the resting wheel vs the static zone from `wheel.class`, but nothing in the key changes when the player starts or ends a drag, so the resting/floating fallback only ever flips if some *other* keyed input changes at the same time. The resulting behaviour is benign (a stale, more conservative bound that self-corrects on the next key change), but the module comment reads as if the swap tracks the drag live.

Related test decisiveness: `'writes the rows-bound sensors while pushing'` looks like it pins that transition, but it bumps `memberCount` 5 -> 6 -> 7 across the three assertions, so the class alone is never shown to drive the re-measure. Holding `memberCount` fixed would make the test honest, and it would go red, which is the useful signal.

**3. [note] On mobile, unset sensors now mean *no* bound at all, where desktop got a viewport fallback.** `src/styles/hud.mobile.css:1752`

When `measureRowsBound` returns null (`rows()` not yet built, or a missing / zero-height pad), the below-target rule resolves to `max(40px, calc(100dvh - 0px - 8px))`, i.e. effectively unbounded, and it out-specifies the base tier's previous `calc(100dvh - safe - 129px)`. That is the exact off-screen-roster failure the second commit fixed for desktop, just with a rarer trigger. Cheapest fix is symmetry: have `measureRowsBound` fall back to `win.innerHeight / z` on mobile too instead of returning null when the pad is unusable.

Also minor: the painter comment says "a hidden or missing pad (chat overlay up) reports null", but `mobile-chat-open` only hides `#mobile-move-zone` (`hud.mobile.css:217`); `#mobile-move-joystick` stays visible and at rest, so that path actually keeps the wheel bound. The comment describes a case that does not occur.

Nothing else: no i18n surface, no sim or server touch, no determinism or seam concerns, no secrets or dev-command changes. Happy to approve once finding 1 is addressed; 2 and 3 are your call.

seanghods:
Thanks, this was a genuinely useful review. Finding 1 is a real regression and my verification could not have caught it, since I was measuring rects and the damage was in paint. Fixed in 92bb0bc7.

**1 (keyline/ring):** taken as suggested, `padding: 6px; margin: -6px`, with the gap terms dropped by the 6px the negative margin lifts the measured rows top (desktop 12 to 6, mobile 8 to 2). I verified it this time with an A/B pixel crop instead of geometry: with the padding neutralized in-page the frame's left edge is sheared flat (no border, no keyline, no rounded corner), and with it the ring wraps intact. One thing worth flagging: the mobile tier only overrides `max-height`, so it inherits the padding from the base rule and the pre-existing mobile clipping is fixed too. Re-measured mobile to be sure: identical clearance at 0 and 8 auras, and 4px better at 18.

**2 (floating class):** agreed the comment oversold it, but I have gone the other way on the fix. Keying the wheel's class would force a layout read at the start and end of every joystick drag, which is the worst moment on the touch HUD to pay for one. The fallback's real job is narrower: to avoid reading a sprung wheel when a re-measure is triggered by something else mid-drag (a buff count changing in combat, which is routine). So the class stays out of the key deliberately and the comment now says that. Your test point stands regardless and was the more valuable half: the test now holds `memberCount` fixed, asserts the class alone causes no re-measure, and then that a real key change mid-drag does fall back to the zone.

**3 (mobile unbounded):** agreed, fixed by symmetry. An unusable pad falls back to the viewport bottom instead of unsetting, so the only null case left is "no member rows exist yet", where there is nothing to bound. You were also right that the chat-overlay case in that comment does not exist (`mobile-chat-open` hides only the zone, not the wheel); the comment is gone rather than reworded.

**The `party-present` note:** added, in your words. It is now documented as load-bearing for select-target-then-join-party, not just the frame variants. Thanks for chasing that one to ground; a future prune of that key would have been silent.

jgyy (commented):
The size here (893/-20) is mostly test file and CSS/comment expansion, not logic sprawl, and the extraction is done the right way: `party_below_target_core.ts` is a genuinely pure module registered in `UI_PURE_CORES`, and the `hud.ts` diff is exactly a one-field, one-delegated-call swap, not new logic on the coordinator.

`client_shell.test.ts`'s CSS pins were updated to assert the new `var(...)` calc with a fallback verified to reproduce the old hand-tuned constants (135px/105px), rather than just loosened to pass, which is the difference between a real regression guard and a rubber stamp. The perf-budget test registers the new painter with an explicit, documented `reflowAllow` for the gated (not steady-state) `getBoundingClientRect` call.

nit: the portrait mobile CSS comment explaining the 2px vs 8px gap arithmetic is correct but fragile if the base rule's padding changes later without the comment being updated alongside it; a shared constant might age better.

Screenshots cover desktop, mobile, and the mobile-heavy (degraded aura) case as the fix specifically targets. No blocking issues.

seanghods:
Good nit, taken in eaf76de0: both rules now derive from a single --party-rows-frame-pad token declared where the padding lives, and the max-height calcs state their intended gaps (12px desktop, 8px mobile) literally instead of pre-subtracted residues, so retuning the pad cannot silently skew the bounds. Verified the rendered geometry is byte-identical with the in-page probe before pushing.

jgyy (commented):
Thanks for chasing this down, and for catching the two adjacent defects (hidden wrapped buff-timer labels, party rows overlapping the mobile joystick / running off-screen with a big roster) in the same pass instead of leaving them for a follow-up. Replacing the hardcoded 96px/135px/105px constants with a measured `--party-below-target-bottom` that accounts for the buff strip's actual wrapped height is the right fix, and falling back to the old constants pre-measure is a sensible safety net.

Good adherence to the pure-core-plus-painter seam: `party_below_target_core.ts` is DOM-free and in `UI_PURE_CORES`, the painter is gated and registered in `HOT_PAINTERS` with the reflow allowance, and `hud.ts` only gets the thin wiring. Test coverage on the geometry core (overlap cases, degenerate rects, uiScale fallback) is decisive.

One thing to be aware of before merge: `gh pr checks` shows no CI runs recorded on the branch, which is likely why GitHub is flagging it unstable. Worth kicking off CI rather than treating that as a real conflict. No code issues found.

jgyy (commented):
Solid fix and a nice architectural choice. `party_below_target_core.ts` is a proper DOM-free pure core (registered in `UI_PURE_CORES`), with the painter reading through the elided `setStyleProp` facet and its `getBoundingClientRect` reads explicitly gated in the perf-budget allowlist. The horizontal-overlap gate (falling back to null offset when the target strip doesn't overlap the party column) is the right fix for a case the old fixed-offset code couldn't express at all, and the scroll-clipping fix for the `.panel` outline/focus ring is a real, non-obvious CSS trap caught in the same PR.

This is layout CSS that can't be asserted directly in jsdom, so the substitution of exhaustive pure rect-arithmetic tests plus CSS source pins for the three tiers is the right call, not a test-coverage gap.

Two nits, neither blocking:
- `party_below_target_painter.ts`: `measureRowsBound` reads `wheel.getAttribute('class')` on every gated re-measure but deliberately excludes that class from the invalidation key, to avoid forcing a layout read on every joystick drag. It's commented and pinned by its own test, just flagging it since a future reader could "fix" it without noticing the intent.
- The CSS var fallbacks (`88px`/`127px`/`97px`) reproduce the old hand-tuned constants for the first frame before a measure lands. Harmless (imperceptible at 20 Hz sim), just noting these are pre-measure geometry fallbacks, not invented balance numbers.

PR #2177, opened 2026-07-20
fix(ui): split the desktop side-menu icon rail into two columns

jgyy:
## Summary
- The desktop micro-menu (`#side-buttons`) had grown into one long, bottom-anchored column of 18 launcher icons running most of the screen's right edge. It was already tight enough on a maximized 1366x768 laptop to need a short-viewport compaction hack (see `tests/crafting_launcher.test.ts`, "side rail height budget").
- Splits the rail into two adjacent flex columns (`#side-buttons-col-a` / `#side-buttons-col-b`) with a tight 2px gap between them, so each column is roughly half the old height and the pair still reads as one compact rail, not two separately spaced groups. Both columns still hug the same right-edge anchor the single rail used to use.
- Mobile is unaffected: it uses its own tray/tray-more UI, not `#side-buttons`.
- Updated the pinned rail-height budget test for the two-column layout, plus added a coverage check that the two columns stay close together.

**Interim fix, not the phase-3 rebuild.** `DESIGN.md` marks the `#side-buttons` micro-rail for retirement (replaced by a launcher + More hub) and `#daily-rewards-button` for a rewards-card rebuild, in phase 3. Phase 1 hasn't landed yet, so this is a scoped fix for a live overflow bug in the meantime; it's expected to be superseded when that phase ships.

### Follow-up (review round)
- The 2px wrapper gap didn't actually render as 2px: `#daily-rewards-button` is 50px wide against the 34px `.micro-btn`, so `align-items: flex-end` right-aligned col-b's buttons inside a content-sized 50px box, opening a 16px dead gutter. Both columns now get an explicit `34px` width (via a shared `.side-buttons-col` class instead of a duplicated id selector), so the chest overhangs left instead and the rail reads as genuinely tight.
- The wider two-column rail could push its icon strip into `#right-tracker-stack`'s vertical span on a short viewport, with the rail painting over quest/delve text (nothing was ever unclickable, the tracker is `pointer-events: none`, but the text was visually covered). Gave the tracker a higher `z-index` so it stays on top.
- Dropped the short-viewport compaction threshold from 720px to 600px: the uncompacted column height already fits the 660px maximized-1366x768 budget the rescue exists for, so players above 600px now get full-size 34px launchers back instead of the compacted 24px ones.
- Added a width-budget test alongside the existing height-budget one (the two-column split moved the real risk from height to width, and nothing pinned that before), pinned the Daily Rewards chest to col-b, and fixed a `#mm-discord` undercount in the visible-button budget calculation.

## Screenshots (desktop)
Before (one long column):
before

After (two adjacent, shorter columns, tight gap, no dead gutter):
after

After, maximized 1366x768 with quests tracked (the tracker-overlap case the previous screenshots didn't show; tracker text now paints above the rail):
after-laptop

## Test plan
- [x] `npx vitest run tests/crafting_launcher.test.ts` (pinned rail-height and rail-width budget tests, updated for the split and the gap/tracker/threshold fixes)
- [x] `npx vitest run tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/mobile_window_coverage.test.ts tests/client_shell.test.ts tests/touch_router.test.ts`
- [x] Full `npx vitest run` (1356 files, 16479 tests, all green)
- [x] `npx tsc --noEmit`
- [x] `npx @biomejs/biome check` on changed files
- [x] Manual verification in the running client (desktop + 1366x768 laptop screenshots above)


FernandoX7 (commented):
Good problem to pick, and the before shot makes the case on its own: the rail had grown into a single column running most of the right edge, over the minimap cluster and the nameplate space. Splitting it is the right instinct, and the parity half is done properly. `index.html` and `play.html` got the identical restructure, which is the thing that usually gets missed here.

I checked the things that could have broken, and they are clean:

- Mobile really is unaffected. `body.mobile-touch #side-buttons { display: none }` (`src/styles/hud.mobile.css:1317`) hides the whole subtree, and `touch_router.ts:99` matches via `closest()`, so the extra nesting level is transparent.
- Nothing depended on the flat structure. No `#side-buttons >`, `:first-child`, or `nth-child` selector anywhere in `src/styles/`, and no TypeScript walks the rail's children. Every id, keybind hint, and i18n attribute is carried over verbatim, so it is a pure re-parenting with zero string churn.
- I ran the suites at the head ref: `crafting_launcher`, `client_shell`, `css_corpus`, `css_value_validity`, `touch_router`, `mobile_window_coverage`, `architecture`. All green. The per-column budget arm is decisive: I dropped 15 extra buttons into col-a and got `col-a: 24 visible micro-btn: expected 674 to be less than or equal to 660`.

Findings.

**should-fix** `src/styles/hud.css:1743` The `gap: 2px` does not render as 2px. `#daily-rewards-button` is `width: 50px` (`hud.css:2862`) against `.micro-btn` at 34px, so col-b is content-sized to 50px and `align-items: flex-end` right-aligns its buttons inside that box, leaving a 16px dead gutter down col-b's left side. The rendered icon-to-icon separation is 2 + 16 = 18px. Measured off this PR's own `after-desktop.png`: col-a's icons occupy x 1498 to 1531, col-b's occupy 1548 to 1581. So the second commit moved the real gap from 22px to 18px, not 6px to 2px, and the "reads as one compact rail" goal is not reached yet. An explicit `width: 34px` on both columns fixes it, and the chest then overhangs left into empty space, since it sits above col-a's topmost button.

**should-fix** `tests/crafting_launcher.test.ts:96` The budget pin went from load-bearing to slack, and the constraint this PR actually introduces is pinned nowhere. Before, the arithmetic was `128 + 18 * 25 + 74 = 652` against 660, eight pixels of headroom, so the next button genuinely failed it. Now col-a is 299 and col-b is 427 against the same 660, which is 233px of headroom, roughly nine more buttons before it fires again. Meanwhile the new constraint is horizontal: the rail's footprint goes from about 50px to about 86px. Nothing pins that. A width budget (col-a + gap + col-b) alongside the height budget would put the guard back where the risk moved.

Related, in the same file: the companion test at line 145, "lays the two columns out close together, not spread across the HUD", measures the wrapper's `gap` property, which is not what drives the separation. I set `#daily-rewards-button { width: 200px }`, which pushes the columns about 168px apart, and the suite stayed 11/11 green. Pin the real driver (column width against chest width) or drop the claim from the name.

**should-fix** `src/styles/hud.css:1743` + `hud.css:2290` The wider rail makes the tracker overlap worse on exactly the laptop this PR is for. `#right-tracker-stack` is `right: 14px; top: 262px; width: 240px` and declares no `z-index`, while `#side-buttons` carries `z-index: 19` and comes later in the DOM (`index.html:438` against `:329`), so the rail's icons paint over quest and delve text. On a 660px usable viewport the compacted col-b top lands around y=210 to 233, above the tracker's `top: 262px`, so the rail overlaps the tracker's entire span, in an 86px band where it used to be 50px. Nothing becomes unclickable, the stack is `pointer-events: none`, but that is 36px more icon strip for wrapped objective lines to disappear behind. On a 1080p viewport the change is an improvement, since the old rail started above the tracker and the new one starts well below it. Both committed screenshots have no quests tracked, so the regressing case is not shown. A 1366x768 shot with a few quests tracked plus a delve tracker would settle it.

**should-fix** `src/styles/hud.css:1868` The `@media (max-height: 720px)` compaction is now over-eager, again on the target machine. Uncompacted, col-b is `128 + 10 * (34 + 4) = 508`, plus the 74px anchor = 582, inside the test's own 660px budget. So the rescue that existed only for that laptop still fires there and shrinks every launcher to 24px high, against the 34px compact figure and 36px desktop hit-target floor in `DESIGN.md:835`. Dropping the threshold to somewhere around 600px hands those players full-size icons back, which is a good part of the win available here.

**nit** `tests/crafting_launcher.test.ts:160` The chest's column membership is unpinned: `DAILY_CHEST_BLOCK_PX` is added to `colBPx` unconditionally. Move the chest back to col-a and the test still passes while col-a is under-counted by 128px. One `expect(colB).toContain('id="daily-rewards-button"')` closes it.

**nit** `tests/crafting_launcher.test.ts:128` `countVisibleMicroBtns` filters out `hidden` buttons, but `#mm-discord` is un-hidden at boot on any Discord-enabled build (pinned in `client_shell.test.ts`) and is plainly visible in this PR's own after shot. col-b's real count is 10, not 9. Inherited from the old test and harmless at the current headroom, but the comment claims to pin what actually fits on screen.

**nit** `src/styles/hud.css:1746` Two new ids sharing one identical rule block, repeated again in the media query. A single `.side-buttons-col` class collapses both to one selector and avoids id specificity for pure layout containers. It also de-brittles the new pin, which currently regexes `#side-buttons-col-a,\s*\n\s*#side-buttons-col-b` and would red-fail on a biome reflow of that selector list.

**nit** `DESIGN.md:1140` lists the `#side-buttons` micro-rail as "RETIRE; replaced by launcher + More hub (7.9)" in phase 3, which `DESIGN.md:1083` calls "Right rail", and `#daily-rewards-button` as "REBUILD as the rewards card". This PR restructures both. Phase 1 has not landed (`src/styles/CLAUDE.md` still does not point at DESIGN.md), so an interim fix for a live overflow bug looks reasonable to me, but the PR body should say it is interim and expected to be replaced at phase 3. Maintainer call.

**nit** The PR body links the screenshots through `raw.githubusercontent.com/jgyy/.../feature/right-tracker-split/...`. They are correctly committed under `docs/screenshots/side-buttons-rail-split/`, so use the repo-relative paths. The fork-branch URLs break when that branch goes away.

One note rather than a finding: tab order is now column-major, all of col-a then all of col-b, while the two columns are bottom-aligned at unequal heights, so tab order and visual top-to-bottom order no longer line up. For a grid of launchers that seems fine to me, and the chest moving from first to eleventh reads as an improvement. Flagging it only so it does not surprise anyone later.

jgyy:
Thanks for the detailed measurement pass, all should-fix items and the actionable nits are addressed in ae5dae676:

- **Gap/gutter:** gave both columns an explicit `34px` width (shared `.side-buttons-col` class) so col-b's buttons no longer right-align inside a content-sized 50px box. The chest overhangs left into the space above col-a as you predicted.
- **Tracker overlap:** `#right-tracker-stack` now sits above `#side-buttons` in z-index, so the rail's icons no longer paint over quest/delve text on short viewports. Added `after-laptop.png`, a 1366x768 shot with quests tracked, since the committed screenshots never showed a tracked quest.
- **Over-eager compaction:** dropped the `max-height` threshold from 720px to 600px per your arithmetic; the rescue now only fires below where it's actually needed.
- **Width budget:** added a test pinning the rail's real width (two 34px columns + 2px gap) against the tracker-overlap risk, alongside the existing height budget.
- **"lays out close together" nit:** now pins the actual driver (explicit column width vs. the chest button), not just the wrapper's flex gap. Reproduced your `width: 200px` counter-example locally to confirm the old assertion stayed green through it; the new one doesn't.
- **Chest column membership nit:** pinned to col-b.
- **`#mm-discord` undercount nit:** fixed, the budget calculation now counts it as visible.
- **`.side-buttons-col` class nit:** done, both the base rule and the media-query rule now use the shared class instead of the duplicated id selector list.
- **PR body links:** switched to repo-relative paths, and added a note that this is an interim fix ahead of the DESIGN.md phase-3 rebuild.

Left the tab-order observation as-is since you flagged it as a note rather than a finding.

FernandoX7 (commented):
The second round holds up well, and I verified it against the head ref rather than the description. The shared `.side-buttons-col` class with the explicit 34px width is the right shape for the gutter fix. The chest's 16px overhang lands above col-a's top in both the normal and compacted regimes, so it covers nothing. The 600px threshold arithmetic is sound (uncompacted col-b needs 582px, inside the 601px worst uncompacted viewport), and play.html stayed in lockstep. I also mutation-tested the new pins at the head ref: widening the column to 40px, opening the gap to 12px, moving the chest to col-a, and reverting the threshold to 720px each turn the suite red, which is exactly what round 1 asked for. The tracked-quests laptop shot and the committed capture script are the right evidence, and the interim-vs-phase-3 framing in the body settles that question.

Two of the fixes have a real problem left in them.

**blocking** `src/styles/hud.css:2307` The z-index bump converts the visual overlap into a click steal. `#right-tracker-stack` and `#side-buttons` are direct siblings under `#ui`, so 20 vs 19 decides hit-testing as well as paint. The stack itself is `pointer-events: none`, but its interactive children re-enable `auto`, and they are full-width strips: `.qt-header` is a `width: 100%` button (`hud.css:2321`), every `.qt-title` row is a block `div role="button"` spanning the tracker's 240px (`hud.css:2364`, built at `quest_tracker_controller.ts:82`), and `#deed-tracker .dt-header` plus `#delve-tracker .dt-affix-row` (`hud.css:2528`) do the same. The rail sits inside the tracker's right 72px, so wherever a strip crosses a launcher, the strip now wins the click (and the hover, so the flyout label cannot open there either). A quest row's hitbox extends far past its visible text, which is what makes this a trap rather than a tradeoff: the icon looks clear, the click goes to the tracker. I verified it at runtime on this branch, headless at 1366x660 with the three quests from your own capture script: eight rail buttons (Character, Spellbook, Talents, Town Focus, Arena, Dungeon Finder, Vale Cup, Card Duel) have their button centers resolving to a `.qt-header` or `.qt-title` strip, and a real `mouse.click` on the exact center of the Spellbook launcher opened the quest log, not the spellbook. With only three quests the steal band is y 262 to 395; more tracked quests, a deed watchlist, or a delve widen it. Before this change the rail won those clicks. Keeping the paint fix is fine; the hitboxes are the bug. An interim-sized fix: stop the clickable strips at their content, `width: fit-content` on `.qt-title` (its hitbox then matches the text, which stays clear of the rail band), and for the headers either the same or move the `pointer-events: auto` onto an inner label span so the full-width underline can stay. Reverting the bump and living with covered text until the phase-3 rebuild is also defensible.

**should-fix** `tests/crafting_launcher.test.ts:170` The height budget now measures a regime that cannot occur at its own viewport. The test applies the compacted metrics (24px + 1px, line 109) against the 660px budget, but after the threshold change compaction only exists at 600px and below: at 660 the rail renders uncompacted, 34px buttons with 4px gaps. It passes today because uncompacted col-b (582px) fits, but the guard has a false-pass window: col-b overflows a real 660px viewport at its 13th visible button (128 + 13 * 38 + 74 = 696) while the compacted arithmetic stays green until the 19th. Measure the uncompacted metrics against 660, and if you want to keep guarding the rescue itself, add a second arm with the compacted metrics against a 600px budget.

**nit** `tests/crafting_launcher.test.ts:214` `railWidthPx` is built from the two local constants, so the `<= 70` assertion is `expect(70).toBeLessThanOrEqual(70)` and can never fail. The live guard is the pair of literal regexes below it (my width mutation was caught by those, so the test does work), but if the CSS and the regexes move together the constants and the 70 bound rot silently. The close-together test already parses the gap out of `hudCss`; parse the width the same way and the budget arithmetic stays honest.

**nit** Whatever shape the first finding's fix takes, pin the mechanism: no assertion anywhere in `tests/` covers the tracker/rail stacking relationship the width test's name references, so removing `z-index: 20` today leaves the suite green.

Two notes, neither on you. The "Screenshots of changes" job on this head was cancelled by the pr-ai-2177 concurrency group, not failed; I will re-run it. And the repo-relative image links follow our documented convention, so leave them even though GitHub does not render them inline in a PR body; that is a convention problem on our side, not this PR's.

jgyy:
Addressed round 2.

**blocking** (src/styles/hud.css:2307, z-index): kept the z-index bump and fixed the hitboxes instead of reverting, per the suggested interim approach. `#quest-tracker .qt-header`, `#quest-tracker .qt-title`, `#deed-tracker .dt-header`, and `#delve-tracker .dt-affix-row` now use `width: fit-content` (with `max-width: 100%`/`calc(100% - 8px)` as a safety cap) instead of spanning the tracker's full 240px. The hitbox now matches the visible label, so a rail button under the tracker's whitespace is no longer covered. Tradeoff: the header's border-bottom underline now only spans the label width instead of the full row, which the review flagged as an acceptable side effect of this approach.

**should-fix** (tests/crafting_launcher.test.ts:170, false-pass window): split the single test into two: "each uncompacted column fits the 1366x768 height budget" now measures the actual uncompacted metrics (30px buttons, 4px gap) against the 660px budget, and a new "each compacted column fits the 600px rescue threshold" test measures the compacted metrics (24px, 1px) against 600px, matching your suggested second arm.

**nit** (railWidthPx tautology): now derives both the column width and the rail gap from the parsed CSS (same pattern the close-together test already uses for the gap) instead of building the expectation from the same local constants as the assertion.

**nit** (no pin for tracker/rail stacking): added a new test, "shrinks the tracker rows to their own content width...", asserting `width: fit-content` on all four rows above, so reverting the fix regresses the suite.

Ran locally: `tests/crafting_launcher.test.ts` (14/14), plus `css_corpus`, `css_value_validity`, `client_shell`, `touch_router`, `mobile_window_coverage`, `architecture`, `quest_tracker_controller`, `deed_tracker_painter`, `delve_tracker_controller` all green, `tsc --noEmit` clean, biome clean on changed files.

FernandoX7 (commented):
Round 3, verified at the head (390c29e4e) rather than from the replies. Everything substantive from round 2 is fixed, and I re-ran the same runtime probe that found the click steal to confirm it. From my side this is mergeable; the two nits below are small and neither needs another full round.

What I checked:

- The fit-content fix covers the complete set of clickable strips: the four rules you changed are the only `pointer-events: auto` re-enables inside the tracker stack section of `hud.css`, and there are no inline pointer-events styles anywhere in `src/ui`. Runtime at 1366x660 with three tracked quests: zero of the 20 visible rail button centers resolve to a tracker strip (round 2: eight), a real mouse click on the Spellbook launcher center now opens the spellbook (round 2 it opened the quest log), and a click on a quest row's visible text still opens the quest log. The row measures 150px instead of the tracker's 240px, exactly the hitbox-matches-text behavior the fix promised.
- The two budget arms now measure the regimes that actually render: uncompacted 30px plus 4px against 660, compacted 24px plus 1px against 600. Your 30px is correct (`.micro-btn` height); my round 2 arithmetic used the 34px width as the height, so the numbers there were overstated by a button's worth.
- Mutation checks at the head: reverting `.qt-title` to `width: 100%` fails the new pin, and widening `.side-buttons-col` to 40px fails the width test through the parsed CSS values, so the tautology from round 2 is gone.
- CI is green including the Screenshots job, and the merge against release/v0.29.0 is clean.

Two nits, plus one take-or-leave:

**nit** `tests/crafting_launcher.test.ts:266` The stacking half of the mechanism is still unpinned: I deleted `z-index: 20` from `#right-tracker-stack` at the head and the whole suite stayed green. The new pin guards the hitbox half, so the click steal cannot return silently, but the paint fix can: without the z-index the rail covers the tracker text again on short viewports and nothing fails. One assertion on the wrapper rule closes it, either here or in the wrapper-rule test in `tests/delve_quest_ui_layout.test.ts` that already parses `#right-tracker-stack`.

**nit** `src/styles/hud.css:1871` The comment's arithmetic uses the 34px button width as its height: `128 + 10 * (34 + 4) = 582`. The real height is 30px, so uncompacted col-b is 542. The conclusion only gets stronger (it fits 660 with more room, and the 600 threshold stays right), and the slip is mine, it came from the round 2 review. Worth correcting while you are in there so the comment and the test constants agree.

Take or leave: nothing cross-pins the uncompacted metrics the way the compaction test pins the literal `height: 24px` and `gap: 1px`, so a future `.micro-btn` height change would silently stale the budget test's 30px constant. A matching literal pin would close that, but for an interim fix headed for the phase 3 retirement I would not hold anything on it.

jgyy:
Addressed round 3's two nits in 9f31a411d:

- **nit** (`#right-tracker-stack` z-index unpinned): added an assertion to the existing wrapper-rule test in `tests/delve_quest_ui_layout.test.ts` (alongside the position/display/flex-direction checks already there) that the rule contains `z-index: 20;`. Mutation-checked locally: reverting to 999 turns that test red, confirming it now catches a silent regression back to overlap-painting on short viewports.
- **nit** (rescue-threshold comment arithmetic): fixed the `hud.css:1871` comment to use the real `.micro-btn` height (30px) instead of the 34px column width, giving uncompacted col-b as `128 + 10 * (30 + 4) = 468`, not 582. The conclusion is unchanged (comfortably under the 660px budget), the comment and the test constants now agree.

Left the take-or-leave (no literal pin cross-checking the uncompacted 30px/4px against `.micro-btn`'s actual CSS) as-is per your note that it's not worth holding an interim fix on.

Ran locally: `tests/delve_quest_ui_layout.test.ts` and `tests/crafting_launcher.test.ts` (35/35), plus `css_corpus`, `css_value_validity`, `client_shell`, `touch_router`, `mobile_window_coverage`, `architecture`, `quest_tracker_controller`, `deed_tracker_painter`, `delve_tracker_controller` all green (183/183 total), `tsc --noEmit` clean, biome clean on changed files.

FernandoX7 (approved):

PR #2179, opened 2026-07-20
perf(sim): scan the spatial grid for nearby mob-mechanic allies

jgyy:
## Summary
- Five mob support mechanics in `updateBossMechanics` (Mend, Ward, the channeled
  protectee heal, Rally, and War Cadence) each ran a full linear scan over every
  entity in the world to find same-faction allies within radius, instead of using
  the `SpatialGrid.forEachInRadius` index that `interaction.ts`, `targeting.ts`, and
  the pull/leash logic already use for the same kind of query.
- Extracts the shared same-faction ally scan into a new pure sibling module,
  `findNearbyAllies(grid, mob, radius, predicate)` (`src/sim/mob/nearby_allies.ts`),
  and swaps all five call sites to it. Each mechanic keeps its own extra filter
  (wounded-only for Mend, exclude-self for the channeled heal's protectee pick) as
  the `predicate` argument; the shared kind/dead/owner/faction checks move into the
  helper once instead of being duplicated five times.
- Byte-identical relocation: same entities included/excluded per mechanic, no RNG
  draws or emit ordering touched, only the scan strategy changes from O(total
  entities) to O(nearby entities via the grid).

## Why
Each block is timer-gated per mob, but with several scripted mobs active in a
zone at once, staggered timers mean at least one of these fires most ticks, each
doing a full-world scan. It's the same waste pattern the two most recent merged
perf commits (party-frame projection caching, snapshot-timer caching) fixed, just
not yet applied to these five blocks.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/mob_nearby_allies.test.ts` (new: grid-vs-brute-force
      parity, predicate composition, exclude-self protectee pick, radius boundary)
- [x] `npx vitest run tests/spatial.test.ts tests/architecture.test.ts`
- [x] `npx vitest run tests/mob_mend_ally.test.ts tests/mob_ward_allies.test.ts
      tests/mob_rally.test.ts tests/mob_warcry.test.ts tests/nythraxis_priest_heal.test.ts
      tests/nythraxis_raid.test.ts tests/delves.test.ts tests/world_boss.test.ts`
  (existing mechanic-behavior tests, all still pass unmodified, confirming the move
  is behavior-preserving)
- [x] `npm test` (full suite): 16480 passed, 1 unrelated pre-existing flake
      (`tests/fixes.test.ts` Varkas Boneguard despawn timer, a `testTimeout` under
      core contention; passes in isolation, unaffected by this diff)
- [x] `npm run ci:changed` (Biome on changed files): clean, no errors
- [x] `npm run build`

## Diagram
```mermaid
flowchart LR
    subgraph before["Before: 5x O(total entities)"]
        M1["mendAlly scan"] --> E1[("this.entities.values()")]
        W1["wardAllies scan"] --> E1
        C1["channelHeal protectee scan"] --> E1
        R1["rally scan"] --> E1
        WC1["warcry scan"] --> E1
    end
    subgraph after["After: 5x O(nearby entities)"]
        M2["mendAlly"] --> F["findNearbyAllies(grid, mob, radius, predicate)"]
        W2["wardAllies"] --> F
        C2["channelHeal protectee"] --> F
        R2["rally"] --> F
        WC2["warcry"] --> F
        F --> G[("SpatialGrid.forEachInRadius")]
    end
```

### DIFF
--- a/src/sim/mob/nearby_allies.ts
+++ b/src/sim/mob/nearby_allies.ts
@@ -1,0 +1,30 @@
+import type { SpatialGrid } from '../spatial';
+import type { Entity } from '../types';
+
+// Shared same-faction ally scan for mob support mechanics (Mend, Ward, the
+// channeled protectee heal, Rally, War Cadence). Each of those periodically
+// walks every living friendly mob within a radius of the caster; this queries
+// the SpatialGrid instead of the full entity map so the cost tracks nearby
+// entities, not total world population. `predicate` layers on any
+// mechanic-specific filter (wounded-only, exclude self, ...).
+export function findNearbyAllies(
+  grid: SpatialGrid,
+  mob: Entity,
+  radius: number,
+  predicate: (ally: Entity) => boolean = () => true,
+): Entity[] {
+  const found: Entity[] = [];
+  grid.forEachInRadius(mob.pos.x, mob.pos.z, radius, (e) => {
+    if (e.kind !== 'mob' || e.dead || e.ownerId !== null) return; // skip players, pets, corpses
+    if (e.hostile !== mob.hostile) return; // same-faction mobs only
+    if (!predicate(e)) return;
+    found.push(e);
+  });
+  // The grid yields entities in cell-bucket order, not the entity-creation
+  // order the old `entities.values()` scan relied on for per-ally rng draw
+  // mapping (mendAlly) and max-hp tie-breaking (channelHeal's protectee
+  // pick). Entity ids are assigned monotonically, so sorting by id restores
+  // that same creation order deterministically.
+  found.sort((a, b) => a.id - b.id);
+  return found;
+}
--- a/src/sim/sim.ts
+++ b/src/sim/sim.ts
@@ -233,6 +233,7 @@ import { NYTHRAXIS_SPIRIT_MENDING_CAST_ID } from './mob/healer_channel';
 import * as lifecycle from './mob/lifecycle';
 import { resetEvadingMob as resetEvadingMobFn, updateMob as updateMobFn } from './mob/locomotion';
 import { runMobSwingAffixes } from './mob/mob_swing';
+import { findNearbyAllies } from './mob/nearby_allies';
 import {
   createMobScanCounters,
   type MobScanCounters,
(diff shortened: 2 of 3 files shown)


FernandoX7 (commented):
Verified the load-bearing claims independently and this holds up. The extraction is house style done right: a pure sibling module under `src/sim/mob/`, five duplicated scans collapsed to one call, and the sort-by-id move is correct. What I confirmed against the code:

- Every mob id comes from the monotonic `ctx.nextId++` sequence, and `dropEntityFromRoster` documents ids are never reused. The only out-of-sequence reserved ids (`VALE_CUP_BRAM_ID`, `FURY_ENTITY_ID`) are npc-kind, which the `kind !== 'mob'` filter excludes, so ascending id order really does reproduce the old `entities.values()` insertion order for every candidate this helper can return.
- Every map insert goes through `addEntityToRoster`, which also inserts into the grid, so the grid query sees the same roster the old full scan did.
- `forEachInRadius` does an exact `d2 <= r2` check on live positions, matching the old keep-if `dist2d <= radius` boundary.
- Ran the new test, the five mechanic suites, and the architecture/spatial guards at the PR head: all pass.

Findings, all minor:

- nit (`src/sim/sim.ts:5112`): the byte-identical claim has one fringe caveat. `applyKnockback` displaces a mob up to several yards without re-bucketing, and the query window pads only 1 unit past the radius, so a mob knocked toward the caster in the same tick a mechanic fires can sit in a bucket cell outside the window and be missed where the old live scan would have caught it. Deterministic either way, and exactly the exposure the existing grid users (pull/leash, targeting, interaction) already accepted, so nothing to change in this PR. A `ctx.rebucket(target)` at the end of `applyKnockback` would make every grid query exact if we ever want to close it, as a separate change.
- nit (`tests/mob_nearby_allies.test.ts:21`): the title says "matching a brute-force scan" but no brute-force comparison runs; the expected ids are hardcoded. Either add the actual brute-force loop next to it or retitle.
- nit (`tests/mob_nearby_allies.test.ts:75`): the boundary test uses 39 and 41 against radius 40, so the inclusive boundary (distance exactly equal to radius, which both old and new code keep) is unpinned. One ally at exactly 40 would pin it.

One observation, not actionable: the grid compares `d2 <= r2` where the old code compared `sqrt(d2) <= radius`, and sqrt rounding can flip inclusion within half an ulp of the exact boundary. Same result on all hosts since they run the same code, so determinism is unaffected; noting it only against the byte-identical wording.

jgyy:
Thanks for the thorough check. Addressed both testing nits in 1cb5914fe:

- `tests/mob_nearby_allies.test.ts:21`: the "matching a brute-force scan" test now actually runs a brute-force filter/sort over the same candidate set and asserts the grid result equals it, instead of only comparing to hardcoded ids.
- `tests/mob_nearby_allies.test.ts:75`: added an ally sitting exactly at distance 40 (the radius), pinning the inclusive boundary that both the old and new code share.

Left the knockback re-bucketing observation and the sqrt-vs-`d2<=r2` rounding note as-is, since you already called out they're pre-existing grid-query exposure and not a regression from this PR.

FernandoX7 (approved):

PR #2180, opened 2026-07-20
fix(render): map the ten unmapped weapons to held model variants

seanghods:
## Summary

Update after merging release/v0.28.0: the merge brought the Phase 10 weaponcrafting ladders with nine new weapon items (copper_bearded_axe, copper_flanged_mace, ironbark_boar_spear, ironedge_longsword, ironshod_maul, whetted_iron_dirk, thorium_warblade, arcanite_war_axe, elderwood_battle_staff), which the coverage sweep added by this PR immediately flagged as unmapped, the exact bug class it closes. They are now mapped to existing variant keys by the same family conventions in a follow-up commit, and the sweep is green again, so the table cannot silently drift for future content.

Ten weapon items were missing from ITEM_WEAPON_VARIANTS (src/ui/weapon_variants.ts), the single itemId to variantKey table that feeds both the 2D bag icon and the 3D held model. The two fallbacks in src/render/characters/assets.ts are asymmetric: swapAttachDef keeps the class DEFAULT mainhand model (so an unmapped mainhand renders a wrong generic weapon), while offhandAttachDef returns null and attachAllProps silently skips the hand (so an unmapped offhand renders nothing). That asymmetry is exactly why dual-wielding two copies of the same unmapped one-hander showed only the mainhand weapon.

This maps each of the ten (gravewyrm_cleaver, mistcallers_fang, lunar_tide_greatstaff, scepter_of_the_deathless_court, stormcallers_focus, nhalias_bell_maul, nhalias_litany_rod, drowned_choir_fang, corpse_candle_focus, mirejaw_fang_knife) to an EXISTING variant key of the visually right family (cleaver to an axe, fangs/knife to daggers, greatstaff to a staff, foci/rod/scepter to wands, bell-maul to a hammer). No new assets: every key already has both its GLB and its icon JPG and belongs to a grip-mapped family, so the boot preload URL set is unchanged. Heroic clones inherit the mapping through the heroicOf resolution. A nice side effect: the ten items' bag icons upgrade from the procedural fallback to the real weapon thumbnails, since the icon layer reads the same table.

Guards added:

- A coverage test sweeping every kind:'weapon' item in ITEMS for a resolvable mainhand model, and every offhand-capable weapon and shield for a resolvable offhand model. Red on exactly these ten items before the fix, green after.
- A pin test making the wraithfire orb (no orb art exists in the shared set) the one conscious held_offhand gap, so a future orb/tome must either map to a model or extend the pin deliberately.
- A change-aware screenshot target (held-weapon-variants in scripts/pr_shot_targets.mjs) staging the cleaver mainhand and a dual-wielding rogue for future regressions of this table.

## Type of change

- [x] Bug fix
- [x] Tests

## How was this tested?

- Commands:
  - npx vitest run tests/held_weapon_models.test.ts tests/asset_pipeline.test.ts (green; the new coverage tests were verified red on exactly the ten items before the mapping rows)
  - npm run gate: green except two pre-existing tests in tests/deploy_watchdog.test.ts ("abandons a hung docker inspect/restart"), a known fake-docker timing flake in this sandbox that fails identically on a pristine release/v0.28.0 checkout; the full suite was re-run with that one file excluded (16595 passed), plus npm run test:browser (66 passed), typecheck, and all builds, all green.
- Manual steps: offline client driven by the new screenshot target: equipped gravewyrm_cleaver on a warrior and two mirejaw_fang_knife copies on a rogue (mainhand plus offhand) and captured the before/after frames below.

## Screenshots / recordings

Gravewyrm Cleaver equipped in the mainhand. Before: the class default sword renders despite the chat line "Equipped Gravewyrm Cleaver". After: a family-appropriate axe.

| Before | After |
| --- | --- |
| before | after |

Dual-wielding two copies of Mirejaw Fang-Knife. Before: only one weapon renders (the offhand hand is empty). After: both hands hold the mapped dagger.

| Before | After |
| --- | --- |
| before | after |

The held model change is identical on mobile (same table, same renderer), so desktop captures only.

---

### Quality

- [x] **The full gate passes.** (With the pre-existing deploy_watchdog sandbox flake noted above; CI runs that file fine.)

### Cross-platform

- ~Tested on desktop and mobile.~ The change is a 3D held-model data mapping, identical on every platform; no UI layout is touched.
- ~Accessible.~ N/A, no UI added or edited.

### Localization (i18n)

### Hygiene

- [x] No secrets, credentials, or .env committed, and ALLOW_DEV_COMMANDS is not enabled in any production path.


jgyy (commented):
Clean fix. The asymmetry diagnosis (mainhand falls back to the class default weapon, offhand falls back to nothing) explains the dual-wield bug well, and the fix is exactly what it says: ten table rows, no new art, each mapped to an existing key from a visually matching family. Checked all ten item ids against `src/sim/content/` and they resolve.

The two coverage tests are the best part of this PR: sweeping every `kind:'weapon'` item for a resolvable mainhand model and every offhand-capable item for a resolvable offhand model means the next unmapped item fails CI instead of silently falling back. The `wraithfire_orb` pin test is a nice touch too, it makes the one remaining gap a conscious, tracked exception instead of a silent one.

Ran `npx vitest run tests/held_weapon_models.test.ts tests/asset_pipeline.test.ts` locally, both green (71 tests).

No blocking issues.

PR #2181, opened 2026-07-20
feat(ui): confirmation dialogs for spirit-healer revive and mark purchases

seanghods:
## Summary

Two irreversible one-click actions now confirm before sending their command, using the existing accessible `Hud.confirmDialog` (role=dialog, focus trap, Escape cancels):

- **Pale Keeper revive.** The spirit-healer revive applied The Keeper's Toll (all attributes -75%, level-scaled up to 10 minutes) on a single tap. All three entry points now route through one confirm gate (`Hud.requestSpiritHealerResurrect`): the ghost-prompt button, clicking the Pale Keeper NPC in the world (`src/game/interactions.ts`), and the interact key (`src/game/nearby_interaction.ts`). The body warns about the debuff and points out that the corpse run is penalty-free. The corpse-run button stays one-tap: it carries no penalty.
- **Marks purchases.** The Heroic Quartermaster and the delve-marks shop buy instantly with no buyback recorded, so a mis-tap was unrefundable. Both now confirm with the item name and mark cost in the body.

Client-only: OK sends exactly the pre-existing command (`resurrectAtSpiritHealer` intent, `buyHeroicVendorItem`, `delveBuyShopItem`), cancel/Escape sends nothing, and `src/sim`/`server` are untouched (they re-validate server-side regardless). The tap handlers are extracted into small named methods (`requestSpiritHealerResurrect`, `requestHeroicVendorPurchase`, `requestShopPurchase`) so tests drive them directly with an injected mock `confirmDialog`, the same pattern as the WoC-store purchase confirm. The two `src/game/` interaction paths call the hud gate through their existing narrow hud interfaces (`PickInteractionHud`/`NearbyInteractionHud`), and `resurrectAtSpiritHealer` is dropped from their world interfaces since the dialog's OK is now the only sender.

All dialog strings render through `t()`. New keys live in `i18n.catalog/hud_chrome.ts` (`hudChrome.death.healerConfirm*`) and `i18n.catalog/index.ts` (`heroicShop.buyConfirm*`, `delveUi.shop.buyConfirm*`); the wordy copy carries its five non-Latin fills (zh_CN, zh_TW, ja_JP, ko_KR, ru_RU) in this change per the M16 rule, reusing each locale's established terms for the Pale Keeper, the Keeper's Toll, and the two marks currencies. Mark costs go through `formatNumber`.

## Type of change

- [x] Tests

## How was this tested?

- Commands:
  - `npm run gate`: green except the two known `tests/deploy_watchdog.test.ts` docker-timing tests, which fail identically on a pristine `release/v0.28.0` checkout in this sandbox (environment flake; CI runs them fine). `npx vitest run --exclude tests/deploy_watchdog.test.ts --maxWorkers=6` passes the remaining suite (16599 passed), and `npx tsc --noEmit`, `npm run build` (all five entries), biome on changed files, i18n gen + freshness, malware scan, and the SFX check all pass.
  - New `tests/hud_confirm_gates.test.ts` plus updated `tests/delve_board_controller.test.ts`: with a mock `confirmDialog`, each command fires only from the dialog's onOk, never from the bare tap, and dismissing sends nothing; the heroic/delve bodies pin the item name and mark cost. Red-before was verified by stashing `src/` (7 failures without the gates).
  - `tests/interactions.test.ts` and `tests/nearby_interaction.test.ts` updated: a ghost's world-click and interact-key on the Pale Keeper now assert routing through the confirm gate and that the resurrect command is never sent directly from those paths.
  - `tests/client_shell.test.ts` touch-tap pin updated to the new handler literal (same `bindTouchTap` intent); `tests/spirit.test.ts` and `tests/heroic_vendor.test.ts` are untouched and green.
- Manual steps: drove the offline client (desktop and landscape-mobile viewports) through all three real buttons via the new `confirm-gates` entry in `scripts/pr_shot_targets.mjs`: died, released, stood at the Pale Keeper, and tapped the revive; opened the Heroic Quartermaster and the delve shop and tapped Buy. Verified the dialog opens each time, the player stays a ghost / no purchase happens until OK, and OK sends exactly the old command.

## Screenshots / recordings

Committed under `docs/screenshots/confirm-gates/`. Before = the same tap acting instantly with no confirm; after = the gate.

| Scene | Before | After |
|---|---|---|
| Pale Keeper revive (desktop) | before healer desktop | after healer desktop |
| Pale Keeper revive (mobile) | before healer mobile | after healer mobile |
| Heroic Quartermaster (desktop) | before heroic desktop | after heroic desktop |
| Heroic Quartermaster (mobile) | before heroic mobile | after heroic mobile |
| Delve-marks shop (desktop) | before delve desktop | after delve desktop |

---

### Quality

- [x] **The full gate passes.** Green apart from the two pre-existing `deploy_watchdog` environment flakes noted above, with decisive tests added for the new behavior.

### Cross-platform

- [x] **Tested on desktop and mobile.** Desktop plus landscape phone metrics (in-game mobile is landscape-only); the dialog reuses the existing `.window/.panel` confirm chrome and its existing button sizing.
- [x] **Accessible.** Reuses the existing `confirmDialog` modal: role=dialog, aria-modal, aria-labelledby, focus trap with Escape-to-cancel, OK focused on open.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** English keys in the matching catalogs; overlay edits are exactly the five required non-Latin fills under the M16 wordy-copy rule.
- [x] Mark costs go through `formatNumber`.
- ~Player-facing text emitted from `src/sim/` or `server/`~ N/A: no sim/server text changes (`tests/localization_fixes.test.ts` passes).

### Hygiene

- [x] Generated i18n artifacts were regenerated via `npm run i18n:gen` (second run leaves the tree clean), never hand-edited.


jgyy (commented):
Good change, and the single-gate design is the right call: routing all three entry points to the Pale Keeper revive (ghost-prompt button, world-click, interact key) through one `Hud.requestSpiritHealerResurrect` method means there's exactly one place that can ever send the command, instead of three places that each need to remember to confirm. Same pattern for the two marks purchase paths.

Verified the confirm-gate wiring end to end:
- `src/game/interactions.ts` and `src/game/nearby_interaction.ts` no longer call `world.resurrectAtSpiritHealer()` directly; both route through `hud.requestSpiritHealerResurrect()`. `resurrectAtSpiritHealer` stays on `IWorld`/`Sim`/`ClientWorld` itself since `main.ts`'s `onResurrectAtSpiritHealer` callback and the online path still use it, just no longer called from these two narrow interfaces.
- `tests/interactions.test.ts` was updated with a negative assertion (`expect(calls).not.toContain('resurrectAtSpiritHealer')`) alongside the positive one, which is exactly the right check for "must never send the command directly."
- The two marks purchases (`Hud.requestHeroicVendorPurchase`, `DelveBoardController.requestShopPurchase`) both look up the offer/item first and bail silently if either is missing, then confirm with the item name and mark cost in the body before sending the pre-existing buy command.
- i18n: new strings are English-only in `hud_chrome.ts` and `index.ts`, rendered through `t()`, and mark costs go through `formatNumber`. That's the whole contributor-side i18n obligation, the rest is release-time.

Ran the touched suites plus a broader check against `release/v0.28.0`:
- `tests/hud_confirm_gates.test.ts`, `tests/delve_board_controller.test.ts`, `tests/interactions.test.ts`, `tests/nearby_interaction.test.ts`, `tests/client_shell.test.ts`, `tests/nythraxis_aldric_npc.test.ts`: all green (173 tests).
- `tests/world_api_parity.test.ts`: green, unaffected as expected since `IWorld` itself didn't change.
- `npx tsc --noEmit`: clean.

No blocking issues. One nit: `requestHeroicVendorPurchase` and `requestShopPurchase` both re-look-up the offer/item and silently return if either is missing (`if (!offer || !item) return;`), which is fine defensively, but since the buttons are only rendered for items that exist in the offer list, that branch is presumably unreachable in practice and untested either way. Not worth a test for a defensive no-op, just flagging it as a case the coverage doesn't reach.

PR #2183, opened 2026-07-20
feat(social): name the not-ready members in the ready check summary

seanghods:
## Summary

When a party/raid ready check finalizes, every participant currently gets only the counts summary ("Ready check: 2 ready, 1 not ready, 1 no response."). This PR adds one line per member who was not ready or never answered, naming them, so the leader can act on the result: "Bee is not ready." / "Cee did not respond to the ready check." Ready members are never named, and an all-ready check still emits only the summary.

This deliberately overturns the old comment in `finalizeReadyCheck` that kept the yes/no answers private "as in the classic client". Classic fidelity actually points the other way: classic-era clients announce the non-ready members to the group after a ready check, and a counts-only tally leaves the leader unable to tell who is holding the pull. The comment now documents the new intent.

Implementation notes:

- Sim change is confined to `finalizeReadyCheck` in `src/sim/social/ready_check.ts` (the extracted SimContext module). Notices only: no rng draws added, removed, or reordered, no tick-phase changes, no new `SimEvent` type (the lines ride the existing `notice` log channel), so no `IWorld`/parity-pin changes. `tests/parity` stays green.
- The sim stays language-agnostic: the two new English emit lines are re-localized at the client boundary in the same change, via two new `baseEnTable` keys plus two `RULES` in `src/ui/sim_i18n.ts` with a `{name}` placeholder (player names splice verbatim, matching the existing ready-check summary rule; the in-file matcher dictionary pattern, so locale fills land at release time as usual). The pre-existing exact string "Pet taunt is not ready." resolves through the EXACT map before the new rule, so it is unaffected.
- A member who left mid-check has no response slot (party.ts drops it) and is never named.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/ready_check.test.ts`: two new tests, proven red before the sim change, green after. One drives a 4-member check (one ready, one not ready, one silent) and asserts every participant receives exactly the two expected name lines with the right states; the other asserts an all-ready check emits no name lines.
  - `npx vitest run tests/localization_fixes.test.ts` (after `npm run i18n:gen`): the S3 drift guard recognizes both new emits; DICT key and placeholder parity green.
  - `npx vitest run tests/architecture.test.ts` and `npx vitest run tests/parity`: sim purity and the golden-trace/rng-draw-order gate green (no draw-order change by construction).
  - `npm run gate`: the i18n gen + freshness, malware scan, changed-files biome, and SFX steps all passed; the full vitest step ran 1381 test files (16698 tests) with exactly one failing file, the two `tests/deploy_watchdog.test.ts` "abandons a hung docker" cases, which fail identically on a pristine `release/v0.28.0` checkout in this sandbox (fake-docker timing, a known environment flake; CI runs them fine). The remaining gate steps were then run directly and are green: `npx tsc --noEmit` and `npm run build` (all five entries).
- Manual steps: none needed beyond the sim tests (server and offline hosts share the same `finalizeReadyCheck` path).

### Quality

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: no UI/layout change; the lines render through the existing chat/log pipeline.
- ~Accessible.~ N/A: no new or edited UI.

### Localization (i18n)

- [x] Player-facing text emitted from `src/sim/` or `server/` is re-localized at the client boundary in this same change, and `npx vitest run tests/localization_fixes.test.ts` passes.
- ~New player-visible strings follow the contributor policy (t() catalog keys).~ N/A: the new strings are sim emits matched in `sim_i18n.ts` (the in-file dictionary pattern), not catalog keys.
- ~Numbers, money, dates, units, and percents go through the i18n formatters.~ N/A: no numbers in the new lines.

### Hygiene

- [x] I didn't hand-edit generated files. (`npm run i18n:gen` left the committed tree unchanged; the new dictionary keys are source, not generated.)

### DIFF
--- a/src/sim/social/ready_check.ts
+++ b/src/sim/social/ready_check.ts
@@ -82,12 +82,22 @@ function finalizeReadyCheck(ctx: SimContext, check: ReadyCheck): void {
     else if (state === 'notready') notReady++;
     else noResponse++; // still pending at finalize -> no response
   }
-  // One counts-only summary line to every participant (the yes/no answers stay
-  // private, as in the classic client). Numbers are re-localized client-side.
+  // The counts summary, then one line naming each member who declined or never
+  // answered, to every participant: classic-era clients announce the non-ready
+  // members to the group after a ready check, so the leader can act on who is
+  // holding the pull. Ready members are never named. A member who left mid-check
+  // has no slot (party.ts drops it) and so is never named either.
   for (const mPid of check.responses.keys()) {
     ctx.notice(
       mPid,
       `Ready check: ${ready} ready, ${notReady} not ready, ${noResponse} no response.`,
     );
+    for (const [pid, state] of check.responses) {
+      if (state === 'ready') continue;
+      const name = ctx.players.get(pid)?.name;
+      if (!name) continue;
+      if (state === 'notready') ctx.notice(mPid, `${name} is not ready.`);
+      else ctx.notice(mPid, `${name} did not respond to the ready check.`);
+    }
   }
 }
--- a/src/ui/sim_i18n.ts
+++ b/src/ui/sim_i18n.ts
@@ -172,6 +172,9 @@ const baseEnTable = {
   'log.partyLeaves': '{name} leaves the party.',
   'log.partyLeft': '{name} has left the party.',
   'log.partyRemoved': '{name} has been removed from the party.',
+  // Per-member ready-check follow-up lines (social/ready_check.ts finalizeReadyCheck).
+  'log.readyCheckNotReady': '{name} is not ready.',
+  'log.readyCheckNoResponse': '{name} did not respond to the ready check.',
   'loot.rollWin': '{winner} wins {item} ({roll})',
   'loot.rollWinnerOffline': '{winner} was offline; {item} returned to the corpse.',
   'loot.rollNeed': 'Need Roll - {roll} for {item} by {name}',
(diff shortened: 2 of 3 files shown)


jgyy (commented):
Nice scoped change. `finalizeReadyCheck` still emits the summary line first, then one line per not-ready/no-response member, and ready members are correctly never named. Verified:

- `EXACT` is checked before `RULES` in `localizeSimText`, so the pre-existing "Pet taunt is not ready." line resolves through EXACT first and can't collide with the new `^(.+) is not ready\.$` rule.
- The party-leave case is covered: a member who leaves mid-check has their response slot dropped by `party.ts`, so they land in no bucket and are never named (test asserts this explicitly).
- No wire/parity changes needed since these are notice-only emits; `tests/parity` and `tests/architecture.test.ts` are unaffected by construction and the diff doesn't touch any serialized field.
- Ran `tests/ready_check.test.ts` (10 passed) and `tests/localization_fixes.test.ts` after `npm run i18n:gen` (30 passed) locally against the PR head, both green.

One nit, not blocking: the inner `for (const [pid, state] of check.responses)` loop re-iterates the full responses map once per participant (O(n^2) in party size), but ready-check parties are small (well under raid cap) so this is not worth optimizing.

Good test coverage, including the all-ready (no name lines) and mid-check-leave cases. Nothing blocking.

PR #2184, opened 2026-07-20
fix(loot): reveal only need rolls when a need roll exists

seanghods:
## Summary

When a need-greed loot roll resolves, chat currently reveals every roll line, so a drop with two needers and four greeders prints six roll lines even though the greed numbers cannot matter (need beats greed). This PR reveals only the contending rolls: when at least one player rolled Need, only the Need lines are broadcast; the Greed lines are revealed only when nobody needed. The winner line is unchanged.

The reveal loop in `resolveLootRoll` (`src/sim/loot/loot_roll.ts`) previously iterated all non-pass entries; it now iterates the `contenders` set the function already computes for winner selection (needers if any exist, else greeders), so the behavioral change is that one loop. Winner selection, the everyone-passed branch, and the offline-winner fallback are untouched.

This deliberately overturns the "reveal every roll ... so the whole group can audit the outcome" intent comment: nothing auditable is lost. Every choice (need, greed, pass) is already visible to the whole group live via `lootRollGroupStatus` while the roll is open, and a greed number cannot affect the outcome once a need exists; revealing it at resolution is pure noise. The comment is rewritten to the new intent.

The parity golden `tests/parity/golden/l1_loot_distribution.json` is regenerated in its own commit per `tests/parity/CLAUDE.md`: the diff is a single line, the event-stream hash at the "after-rolls" checkpoint. The state hash and the rng record (draw count and digest) are byte-identical, which is the direct evidence that no rng draw or state behavior changed, only which loot lines are emitted.

No new player-facing text is emitted (the existing `Need Roll` / `Greed Roll` formats fire for fewer entries), so no `sim_i18n` matcher changes; the `loot.rollGreed` matcher stays reachable through the greed-only path.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/loot_roll.test.ts` (red-before proven: the new negative assertion, Greed line absent when a need exists, fails on the base branch; green after)
  - `npx vitest run tests/loot_roll.test.ts tests/loot_messaging_sim.test.ts tests/loot_roll_reconcile.test.ts tests/loot_roll_controller.test.ts tests/loot_master_sim.test.ts tests/dead_party_loot.test.ts` (75 tests green)
  - `UPDATE_PARITY=1 npx vitest run tests/parity/parity.test.ts` to mint the golden, then `npx vitest run tests/parity` green
  - `npm run gate`: every step green except the two known-flaky `tests/deploy_watchdog.test.ts` "abandons a hung docker" cases (fake-docker timing), which fail identically on a pristine `release/v0.28.0` checkout in my sandbox and are unrelated to this diff; the rest of the full suite is green (1377 files, 16697 tests passed) along with the i18n freshness, malware scan, biome, SFX conformance, `tsc`, and all five builds
- Manual steps: none needed; chat-text-only change, fully covered by the sim tests.

Test changes in `tests/loot_roll.test.ts`:
- The mixed-roll case now asserts the Need line is present AND the Greed line is absent (a decisive negative assertion, not just presence of the need line), plus the winner-line ordering.
- New greed-only case: both greed lines still revealed to every party member, no need lines, winner line last.

### Quality

- [x] **The full gate passes.** Green except the two pre-existing sandbox-flaky
      `deploy_watchdog` cases noted above (fail identically on a pristine base
      checkout; CI runs them fine). Decisive tests added and updated for the
      changed behavior; manual checks recorded above.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: chat text only, no layout or input change.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings. (Existing matched formats fire for fewer entries; `loot.rollGreed` remains reachable via the greed-only path.)

### Hygiene

- [x] I didn't hand-edit generated files. The parity golden was regenerated via
      `UPDATE_PARITY=1 npx vitest run tests/parity/parity.test.ts` (its sanctioned owner), in its own commit.

### DIFF
--- a/src/sim/loot/loot_roll.ts
+++ b/src/sim/loot/loot_roll.ts
@@ -704,10 +704,14 @@ export function resolveLootRoll(ctx: SimContext, roll: PendingLootRoll): void {
       ctx.emit({ type: 'loot', text: `Everyone passed on [[i:${roll.itemId}]].`, pid });
     return;
   }
-  // Reveal every roll, classic-style: one loot line per need/greed roller so the
-  // whole group can audit the outcome (passes were already visible live via
-  // lootRollGroupStatus and have no number to reveal).
-  for (const entry of entries) {
+  // Reveal one loot line per CONTENDING roller only: when anyone needed, need
+  // beats greed, so the greed numbers cannot affect the outcome and revealing
+  // them is noise. Nothing is hidden that matters: every choice (need, greed,
+  // pass) was already visible live via lootRollGroupStatus while the roll was
+  // open, and the winner line below still closes the roll. With no needers the
+  // greed rolls are the contest and all of them are revealed as before (passes
+  // have no number to reveal).
+  for (const entry of contenders) {
     const rollerName =
       ctx.players.get(entry.pid)?.name ?? roll.candidateNames.get(entry.pid) ?? 'Unknown';
     for (const pid of partyMembersForRoll(roll)) {
--- a/tests/loot_roll.test.ts
+++ b/tests/loot_roll.test.ts
@@ -421,7 +421,7 @@ describe('loot_roll: group roll status + resolution broadcast (module entry)', (
     }
   });
 
-  it('broadcasts every need/greed roll to the whole party at resolution, then the winner line', () => {
+  it('reveals only the need rolls at resolution when anyone needed, then the winner line', () => {
     const { sim, a, b, c, rollId } = openRoll();
     submitLootRoll(sim.ctx, rollId, 'greed', b);
     submitLootRoll(sim.ctx, rollId, 'need', a);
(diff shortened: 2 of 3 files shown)


jgyy (commented):
Clean, well-justified fix. Switching the reveal loop from `entries` to the already-computed `contenders` set (needers if any, else greeders) is the right minimal change, and the reasoning holds: a greed number can never beat a need roll, so revealing it at resolution is noise, not an audit trail (`lootRollGroupStatus` already exposed the live choices while the roll was open).

Verified:
- `contenders` is the same set winner selection already uses a few lines below, so this can't desync the reveal from who actually wins.
- The parity golden diff (`tests/parity/golden/l1_loot_distribution.json`) is exactly the single line claimed: only the `events` hash changed, `state` hash and `rng.draws`/`digest` are byte-identical, confirming no rng draw or state drift, only which loot lines are emitted.
- Test changes are decisive: the mixed-roll case now asserts the greed line is *absent*, not just that the need line is present, and the new greed-only case asserts both greed lines still appear with no need lines.
- Ran `tests/loot_roll.test.ts` (30 passed) and `tests/parity` (180 passed, 1 skipped) locally against the PR head, both green.

No blocking issues, no nits. Good PR.

PR #2185, opened 2026-07-20
feat(bot): auto-rotate the Discord server invite before it expires

jgyy:
## Summary
- The official server invite (discord.com/invite/worldofclaudecraft) is a regular Discord invite, which expires after Discord's 30-day default, so it periodically stops working until someone notices and manually regenerates it.
- The bot now mints a fresh never-expiring invite (`max_age: 0`) on a poll loop and keeps a single channel message current with it, refreshing every 25 days so the link never goes stale.
- Scope is the bot-owned Discord side only (a live, self-refreshing invite message in a designated channel via the new optional `DISCORD_INVITE_CHANNEL_ID`). The static `discord.com/invite/worldofclaudecraft` link in README.md/CONTRIBUTING.md/SECURITY.md/CODE_OF_CONDUCT.md, and the separate server-side `DISCORD_GUILD_INVITE` env var read by `server/discord.ts` for the in-game "join Discord" link, are both out of scope here and could be a follow-up (e.g. pointing them at the bot-managed invite, or exposing it via `/internal/discord/*`).

## Flow

```mermaid
flowchart LR
    A[Poll tick, every 5 min] --> B{inviteRefreshDue?<br/>logic.ts, 25-day interval}
    B -- no --> A
    B -- yes --> C[DiscordApi.createInvite<br/>max_age 0, max_uses 0]
    C --> D{Existing message id<br/>in memory?}
    D -- yes --> E[DiscordApi.editMessage]
    D -- no --> F[DiscordApi.createMessage]
    F --> G[store message id]
    E --> H[store refresh timestamp]
    G --> H
```

## Changes
- `bot/logic.ts`: pure `inviteRefreshDue` (cadence) and `buildInviteMessage` (embed), unit-tested.
- `bot/discord_api.ts`: `createInvite` and `editMessage` REST calls; `createMessage` now returns the created message so its id can be reused.
- `bot/main.ts`: wires a poll loop reusing the existing 5-minute interval cadence, checking due-ness via `logic.ts` then creating/editing via `discord_api.ts`.
- `bot/config.ts`, `.env.example`, `bot/CLAUDE.md`: new optional `DISCORD_INVITE_CHANNEL_ID` env var, documented; feature is a no-op when unset.
- `tests/discord_bot.test.ts`: cadence + message-shape tests for the new pure logic.

## Test plan
- [x] `npx vitest run tests/discord_bot.test.ts` (48 passed, includes 4 new invite-rotation tests)
- [x] `npx @biomejs/biome check` on changed files (only pre-existing warnings in files I didn't touch)
- [x] pre-push QA floor (architecture + i18n guard tests, biome, copy scan) green


Rubsey (commented):
The pure cadence and message-building helpers are clean, and the focused tests pass. A few deployment and lifecycle issues need attention:

- **blocking** - `bot/config.ts:61`: `DISCORD_INVITE_CHANNEL_ID` is not forwarded by the `discord-bot` service in `docker-compose.yml`. I confirmed the resolved Compose configuration omits it, so `main.ts:530` always disables this feature in the canonical deployment. Add the Compose mapping and pin it in `tests/deploy_discord_bot.test.ts`.
- **should-fix** - `bot/main.ts:527`: every process restart resets the message ID and immediately posts another permanent invite message. Routine deploys will accumulate orphaned messages instead of maintaining the stated single current message. Persist, configure, or rediscover the managed message ID.
- **should-fix** - `bot/main.ts:537`: if the managed message is deleted, every subsequent refresh attempts to edit the same missing ID forever. Handle Discord's unknown-message response by clearing the stale ID and creating a replacement, with a test covering that transition.

jgyy:
Thanks for the review, all three addressed in bcc64dce2:

- **blocking (config.ts:61)**: `docker-compose.yml` now forwards `DISCORD_INVITE_CHANNEL_ID` to the `discord-bot` service, pinned in `tests/deploy_discord_bot.test.ts`.
- **should-fix (main.ts:527)**: on boot the bot now lists the invite channel's recent messages and rediscovers its own previously-posted invite message (matching author id + the invite embed's stable title via the new `findManagedInviteMessage`), reusing that id and its last-edit time instead of always posting a new message. A fresh message is only created when none is found.
- **should-fix (main.ts:537)**: `editMessage` failures are now checked for Discord's Unknown Message error (code 10008, via `isUnknownMessageError`); on that, the stale id is dropped and a replacement message is posted in the same refresh instead of retrying the dead id forever.

Added unit tests for the new pure helpers (`findManagedInviteMessage`, `isUnknownMessageError`) in `tests/discord_bot.test.ts` and the Compose mapping pin. `tsc --noEmit` and the touched test files are green.

PR #2186, opened 2026-07-20
feat(ui): hover tooltip for player characters

seanghods:
## Summary

Hovering a living non-self player now uses the existing bottom-right world-hover tooltip. The tooltip shows the player name in the class color, a localized `Level N Class` line, and a guild line only when the entity guild is non-empty.

The main hover gate in `src/main.ts` remains unchanged for mobs and adds one additive player branch: `kind === player`, not the local `world.playerId`, and not dead. Friendly and hostile players both qualify. Hovering self, a dead player, or empty world clears the shared tooltip as before.

Rendering stays client-only. `HUD.showPlayerHoverTooltip` reuses the existing mob-tooltip cache, DOM element, placement, and clear path, but supplies a player-shaped model to the new pure `src/ui/player_tooltip_view.ts`. The view escapes name and guild values, resolves the localized class label through `classDisplayName`, and uses the existing class-color helper. No server, wire, `IWorld`, or parity surface changed.

The level/class text reuses `itemUi.equipment.levelClass`; no catalog keys were added. The new `_view` module is registered in `UI_PURE_CORES`, preserving the architecture completeness check.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [x] Tests
- [x] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npm run i18n:gen`
  - `npx vitest run tests/player_tooltip_view.test.ts tests/mob_tooltip_view.test.ts tests/architecture.test.ts tests/pr_shot_targets.test.ts` (50 tests green)
  - `npx vitest run tests/pr_shot_targets.test.ts tests/player_tooltip_view.test.ts` after the full-frame screenshot adjustment (17 tests green)
  - `npx tsc --noEmit` and `npm run ci:changed` green; the latter reports existing repository Biome warnings only
  - pre-push QA floor green: typecheck, guard tests (55 passed, 3 skipped), lint warnings only, and copy rules
  - `npm run gate` reaches the full Vitest stage, where only `tests/deploy_watchdog.test.ts` cases `abandons a hung docker inspect on the bound and exits cleanly` and `abandons a hung docker restart on the bound, keeping cooldown stamp` fail. This macOS host has neither `timeout` nor `gtimeout`; the watchdog intentionally falls back to an unbounded implementation, while those tests require a bounded command. The gate therefore stops before later gate stages; it is not claimed green.
- Manual visual check: added an offline Mage bot with `sim.addPlayer`, set its guild directly, then activated the actual canvas hover path with `page.hover` followed by `page.mouse.move` over a renderer pick point. The full desktop after screenshot shows the staged player, class-colored name, level/class line, and guild line in the shared tooltip.

Test coverage in `tests/player_tooltip_view.test.ts` verifies class-colored title rendering, the existing level/class translation key, guild presence, guild absence for empty and undefined values, HTML escaping for name and guild, and deterministic output. `tests/pr_shot_targets.test.ts` pins screenshot-target registration; the target retries renderer picking briefly before driving the real mouse path so capture is stable.

## Screenshots / recordings

Desktop before:

Before

Desktop after:

After

Mobile decision: world hover is mouse-only, matching the existing enemy tooltip. Touch does not open this tooltip, so no mobile screenshot is forced.

### Quality

- [ ] **The full gate passes.** It is blocked only by the two macOS `deploy_watchdog` timeout cases documented above; all focused checks and the pre-push QA floor pass.

### Cross-platform

- [x] Desktop visual behavior verified through the real canvas hover path and full-frame before/after screenshots.
- [x] Mobile behavior intentionally unchanged: hover is mouse-only, as for the existing mob tooltip.

### Localization (i18n)

- [x] No new player-visible catalog keys. The tooltip reuses `itemUi.equipment.levelClass` and existing localized class names.

### Hygiene

- [x] No server, wire, `IWorld`, or parity changes.
- [x] No secrets, credentials, or `.env` files committed.
- [x] Screenshot artifacts were generated by the registered offline PR-screenshot target, not manually composed.


jgyy (commented):
Clean, focused change. The player tooltip logic follows the existing mob tooltip pattern closely: pure escaped view module (`player_tooltip_view.ts`), registered in `UI_PURE_CORES`, reusing the shared bottom-right painter and cache-key gating in `hud.ts`. The generalized `lastHoverTooltipId` (prefixed `mob:`/`player:`) and renamed `clearHoverTooltip` are a sensible small refactor rather than a parallel code path. `guild` is already on the wire (`gd` in `server/game.ts`), so no new wire/parity surface is needed, matching the PR's claim. Test coverage (class color, level/class line, guild presence/absence, escaping, determinism) is solid.

A couple of small things, both nits:

- `src/ui/hud.ts:4419`: `classLabel` falls back to `entity.templateId` (the raw class id) when `CLASSES[playerClass]` is missing, same pattern used elsewhere (`hud.ts:4571`, `hud.ts:8003`) for mismatched/legacy templateId data, so this is consistent with existing precedent rather than a new gap.
- `paintMobTooltipBottomRight` is now also painting player tooltips; the name is inherited from the mob-only precedent and not renamed here. Not worth blocking on, but a future tooltip-painter touch might want to rename it.

No sim, wire, or server changes, i18n stays within the existing `itemUi.equipment.levelClass` key (no new catalog entries needed), and `merge-tree` against `release/v0.28.0` is clean. Nice work.

PR #2190, opened 2026-07-20
fix(net): bounce a desktop-shell Discord login back to /desktop-login

jgyy:
## Problem

On the desktop (Electron/Steam) build, clicking **Continue with Discord** opens the
system browser to `/desktop-login` and correctly starts the OAuth flow, but after
approving on Discord the browser lands on the plain website home page (`/`), logged
in there in the browser tab, instead of handing control back to the desktop app.

Root cause: `startDiscordOAuth('login')` (`src/main.ts`) only special-cases the
mobile native app (`NATIVE_APP`) for the PKCE/deep-link handoff. Everything else,
including the desktop shell's system-browser flow, falls through to a plain
full-page OAuth redirect. The server's callback (`server/discord.ts`
`handleDiscordCallback` / `bouncePage`) only returns through the
`worldofclaudecraft://` deep link for that same native flow; every other login
writes the session to `localStorage` and does `location.replace('/')`. Since the
desktop login start was never tagged as anything special, the callback always
bounces to `/`, so `completeDesktopBrowserLogin()` (`src/main.ts`, the function that
mints the one-time code and navigates to `worldofclaudecraft://desktop-login?code=...`)
never runs, because it only fires when the page is `/desktop-login`, and the
callback never sends the browser back there.

## Fix

Add a **desktop** (not native) signal, parallel to the existing native-app PKCE
handoff, that only changes where the OAuth callback bounces the browser back to:

- `src/main.ts`: `startDiscordOAuth('login')`'s plain-redirect branch now passes
  `isDesktopLoginPage()` as the new `desktop` flag on `discordStart`. The signal is
  "am I currently ON `/desktop-login`", not a runtime/DESKTOP_APP check, because the
  desktop shell opens this login screen in the OS system browser
  (`electron/main.cjs` `openDesktopLogin` via `shell.openExternal`), never inside
  Electron itself, so `NATIVE_APP`/`DESKTOP_APP` are both false there.
- `src/net/online.ts`: `discordStart()` grows an optional `desktop` param that adds
  `&desktop=1` to the start request (mutually exclusive with `native`).
- `server/discord.ts` (+ the legacy `server/main.ts` inline arm, kept in parity):
  `handleDiscordStart` stores a `desktop-login` marker in the existing
  `discord_oauth_states.redirect_to` column (the same column the native PKCE
  handoff already repurposes for its own marker), and `handleDiscordCallback`
  reads it back to redirect the bounce page to `/desktop-login` instead of `/`.
  `/desktop-login` is the same SPA entry as `/` (same `index.html`, client-side
  routed), so this also correctly preserves the first-time-login "create new /
  link existing" chooser flow.

No new endpoints, no schema change (the `redirect_to` column already existed for
the native handoff), and link-mode Discord (popup flow, always same-tab) is
untouched.

## Flow

**Before (broken):**

```mermaid
sequenceDiagram
    participant App as Desktop app (Electron)
    participant Browser as System browser
    participant Server as Game server
    participant Discord

    App->>Browser: shell.openExternal(/desktop-login)
    Browser->>Server: GET /desktop-login (login panel)
    Browser->>Browser: click "Continue with Discord"
    Browser->>Server: POST /api/auth/discord/start (no desktop signal)
    Server-->>Browser: authorize URL
    Browser->>Discord: full-page redirect, approve
    Discord->>Server: GET /api/auth/discord/callback
    Server-->>Browser: bounce page: write session, location.replace('/')
    Browser->>Server: GET / (plain website, logged in HERE)
    Note over App: Never notified. Desktop app still shows the login screen.
```

**After (fixed):**

```mermaid
sequenceDiagram
    participant App as Desktop app (Electron)
    participant Browser as System browser
    participant Server as Game server
    participant Discord

    App->>Browser: shell.openExternal(/desktop-login)
    Browser->>Server: GET /desktop-login (login panel)
    Browser->>Browser: click "Continue with Discord"
    Browser->>Server: POST /api/auth/discord/start&desktop=1
    Server-->>Browser: authorize URL (state marked desktop-login)
    Browser->>Discord: full-page redirect, approve
    Discord->>Server: GET /api/auth/discord/callback
    Server-->>Browser: bounce page: write session, location.replace('/desktop-login')
    Browser->>Server: GET /desktop-login (session now present)
    Browser->>Browser: completeDesktopBrowserLogin() mints one-time code
    Browser->>App: worldofclaudecraft://desktop-login?code=... (OS prompt)
    App->>Server: exchange code for session
    Note over App: Logged in inside the desktop app.
```

## Testing

- `npx tsc --noEmit`
- `npx vitest run tests/discord_server.test.ts tests/server/discord.test.ts tests/server/http/characterization.test.ts tests/architecture.test.ts tests/client_shell.test.ts tests/discord_onboarding_gate.test.ts`
  (added coverage: a desktop-flagged start persists the `desktop-login` marker; the
  callback bounces a desktop-marked login, both the straight-session and
  first-time-chooser branches, back to `/desktop-login` instead of `/`; a plain
  login still bounces to `/` as before; the new `shouldEnterDiscordOnboarding`
  predicate is pinned for all three boolean combinations)
- Rebased onto `release/v0.29.0` (was cut from `release/v0.28.0`); PR is now
  MERGEABLE against its new base.
- `npm test` (full suite): 17518 passed, 37 skipped, 0 failed.
- `npm run gate`: everything through the full vitest suite is green; the optional
  real-browser suite (`npm run test:browser`) has one pre-existing failure in
  `armory_mobile_layout.browser.test.ts` (a pixel-height assertion on the Armory's
  mobile cosmetic card, unrelated to this change and not touched by this diff).


jgyy:
\`npm run gate\` result: everything through the full test suite, tsc, and build passed. The final \`test:browser\` stage failed on one unrelated pre-existing flake, \`tests/browser/armory_mobile_layout.browser.test.ts\`, a pixel-tolerance layout assertion (\`expected 394.625 to be less than or equal to 391\`) in the Armory store, a subsystem this PR never touches. Confirmed by stashing this branch's changes and re-running the same test directly against a clean checkout of \`release/v0.28.0\`: it fails identically there.

jgyy:
Related: #2194

FernandoX7 (commented):
The root-cause analysis is right and the server side is clean: the page-based signal (instead of a runtime check) is the correct read since the shell opens the OS system browser where `NATIVE_APP`/`DESKTOP_APP` are both false, the marker reuses the existing `redirect_to` column (server/discord_db.ts:50, no DDL), it cannot collide with the native handoff (`nativeChallengeFromState` returns null for it), both the RouteDef arm and the legacy `server/main.ts` arm are updated in parity, and the bounce target is a server-side constant so there is no open-redirect surface. The new tests pin the literal `var target = "/desktop-login";` in both callback branches plus a negative for the plain login, and the fixture regen is faithful.

The gaps are on the client, after the bounce lands:

- **blocking** `src/main.ts:8540` (`finishDiscordChoice`): a first-time desktop login still never reaches the app. The bounce now correctly parks the chooser on `/desktop-login`, but completing it runs `enterLoggedInChrome()` and `goToLoggedInPlay()` with no `completeDesktopBrowserLogin()` step, so the browser tab logs into web play and the shell keeps waiting: the original symptom, for every new account. The username/password path already shows the pattern (`src/main.ts:7485`: `completeOnlineAuth` calls `if (await completeDesktopBrowserLogin()) return;` right after `enterLoggedInChrome()`); `finishDiscordChoice` needs the same. The new server tests pin the chooser bounce but nothing covers this client completion, which is how it slipped.

- **should-fix** `src/main.ts:8815`: on the returning-user path, `discordOnboarding` is always true in this flow (`startDiscordOAuth` sets the onboard key before redirecting, `src/main.ts:6367`), so boot now fires `maybePromptRecoveryEmail().then(() => enterOnlinePlayFlow())` and `completeDesktopBrowserLogin()` together on `/desktop-login`: the tab starts entering online play (possibly showing the recovery-email prompt) while the deep-link handoff is putting up the OS prompt. The adjacent resume guard's own comment (`src/main.ts:8780`) says this page "must mint its code, never load the game"; gate the onboarding arm on `!isDesktopLoginPage()` too.

- **should-fix** branch base: the branch was cut from the release/v0.28.0 tip, so it carries the #2127 and #2129 squash commits (15 of the 22 changed files) that release/v0.29.0 does not have yet. A squash-merge here would land all of that misattributed under this fix's title and then duplicate against the eventual v0.28.0 forward merge. Rebase the single fix commit (1b25d53) onto release/v0.29.0; it touches disjoint files and should cherry-pick clean.

- **nit** `docs/screenshots/discord-desktop-login-handoff/desktop-login-page.png` is 3.2 MB for a page the PR itself says is visually unchanged. The screenshot requirement is for visual changes; drop it or recompress.

CI note: the failed "Screenshots of changes" job is environmental (the checkout fetch was canceled, so `scripts/pr_comment_shots.mjs` was missing from the workspace); a re-run should clear it.

jgyy:
Thanks for the thorough review, @FernandoX7. All four points are addressed:

1. **BLOCKING (first-time login never reached the app):** confirmed. `finishDiscordChoice` saved
   the session and ran `enterLoggedInChrome()` / `goToLoggedInPlay()` with no
   `completeDesktopBrowserLogin()` step at all, so a brand-new account still reproduced the
   original bug for every new account on the desktop shell. Made `finishDiscordChoice` async and
   added the exact same `if (await completeDesktopBrowserLogin()) return;` early-return that
   `completeOnlineAuth` already uses right after `enterLoggedInChrome()`.

2. **should-fix (onboarding arm racing the handoff page):** confirmed. `discordOnboarding` is
   effectively always true on this flow (`startDiscordOAuth` sets the key before redirecting), so
   on `/desktop-login` the returning-user boot path fired
   `maybePromptRecoveryEmail().then(() => enterOnlinePlayFlow())` at the same time as
   `completeDesktopBrowserLogin()`. Extracted the decision into a small pure predicate,
   `shouldEnterDiscordOnboarding(discordOnboarding, isDesktopLoginPage)` in the new
   `src/net/discord_onboarding_gate.ts`, unit tested for all three relevant boolean combinations,
   and gated the onboarding arm on it so it never fires on the handoff page (matching the resume
   guard a few lines above, which already special-cases `/desktop-login`).

3. **should-fix (branch base):** rebased the branch onto the current `release/v0.29.0` tip (it had
   been cut from `release/v0.28.0`). The rebase was mechanical, no conflicts, nothing to
   regenerate (no i18n or parity-pin touch points in this diff). Retargeted the PR
   (`gh pr edit --base release/v0.29.0`); it is now `MERGEABLE` against the new base.

4. **nit (oversized screenshot):** dropped the 3.2 MB `desktop-login-page.png` and the PR body's
   Screenshot section along with it. The fix is behavior-only (a server-side redirect target), and
   the page itself is admittedly unchanged, so the screenshot wasn't earning its size.

Also added a small regression test for the new predicate
(`tests/discord_onboarding_gate.test.ts`).

What I verified:
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/discord_onboarding_gate.test.ts tests/discord_server.test.ts tests/server/discord.test.ts tests/architecture.test.ts tests/client_shell.test.ts`: all green.
- `npm test` (full suite): 17518 passed, 37 skipped, 0 failed.
- `npm run gate`: green through the full vitest run; the only failure is in the optional
  `npm run test:browser` real-browser suite, in `armory_mobile_layout.browser.test.ts` (an Armory
  mobile cosmetic-card pixel-height assertion), which is unrelated to this change and not touched
  by this diff. Agreed with your read on the earlier "Screenshots of changes" CI job being
  environmental as well.
- Biome on changed files: no errors, no new warnings beyond pre-existing ones on untouched lines.

FernandoX7 (commented):
Re-reviewed at 860f9d84f. Both round-1 code findings are fixed, and fixed well:

- The blocking `finishDiscordChoice` gap is closed the right way: it now runs `if (await completeDesktopBrowserLogin()) return;` immediately after `enterLoggedInChrome()` (src/main.ts:8563), the exact `completeOnlineAuth` pattern (src/main.ts:7502). Since `completeDesktopBrowserLogin` returns false off the handoff page, the normal web chooser flow is untouched.
- The onboarding race is gated through the new pure `shouldEnterDiscordOnboarding` predicate (src/main.ts:8835), with all four input combinations pinned in tests/discord_onboarding_gate.test.ts and a site comment that explains the handoff-page rule.
- The 3.2 MB screenshot is deleted.

On the base: the rebase onto release/v0.29.0 landed (merge-base a9d499291, MERGEABLE), but the branch still carries the #2127 and #2129 squash commits. I checked what that means concretely: those two are the only substantive commits release/v0.28.0 has over release/v0.29.0 (the music commit and its revert cancel out), and both carried commits are patch-identical to the merged d9a804011 / 72d5c7220 with authorship preserved (range-diff shows them as equal). So the branch effectively doubles as the forward-port of the missing v0.28.0 content; nothing new is riding along. The fix commit itself (bb332ab18) is patch-identical to the round-1 commit whose server side I already verified.

Remaining items:

- should-fix (merge mechanics, not code): do not squash-merge this PR, or 15 files of #2127/#2129 UI work land attributed to a fix(net) title. Either merge-commit it so the four commits survive as-is, or forward-merge release/v0.28.0 into release/v0.29.0 first and rebase this branch, which will auto-drop the two identical carried commits and shrink the PR to the two fix commits.
- nit src/main.ts:8699: `finishDiscordChoice` is async now, but this call site does not return its promise into the `.then` chain, unlike the sibling at src/main.ts:8647, so on the 2FA link path the busy state clears before the handoff finishes and a rejection would bypass `onDiscordChoiceError`. `return finishDiscordChoice();` matches the other call site. Low stakes, since `completeDesktopBrowserLogin` catches its own API errors.

CI: the failed "Screenshots of changes" job is the same environmental failure as last round (the checkout step was canceled, so scripts/pr_comment_shots.mjs never existed on the runner); every real gate is green, including all four vitest shards.

jgyy:
Addressed the round-2 nit: the 2FA-link call site (around src/main.ts:8699, inside `submitLink`'s `.then`) now does `return finishDiscordChoice();` instead of a bare call, matching the sibling create-account call site (src/main.ts:8647, `.then(finishDiscordChoice)`). Since `finishDiscordChoice` is async, the bare call previously let `.finally(() => setDiscordChoiceBusy(false))` clear the busy flag before the async work settled, and would not route a later rejection to `onDiscordChoiceError`. Fixed in bb99e08e3.

Verified: `npx tsc --noEmit` clean, `tests/discord_onboarding_gate.test.ts`, `tests/native_discord_handler.test.ts`, `tests/native_discord.test.ts`, `tests/discord_deeplink.test.ts`, and `tests/architecture.test.ts` all green, biome clean on the changed file, pre-push QA floor green.

On the should-fix merge-mechanics item: confirmed via `git log --oneline up/release/v0.29.0..HEAD` that this branch carries forward two commits from the v0.28.0 forward-port into v0.29.0 (356f15d1e "hide the mobile bottom action bar while any game window is open (#2127)" and 665c838fd "relayout the Quest Log window into a wide landscape shape (#2129)") ahead of the two Discord-handoff fix commits. Per your finding these are patch-identical duplicates of content already forward-merged, with nothing new riding along. Please do **not** squash-merge this PR: a squash would flatten those two commits into this PR's history as if they originated here. The maintainer should either merge with a merge commit (preserving the original commit identities) or forward-merge release/v0.28.0 into release/v0.29.0 first so this branch's diff against base no longer carries them.

FernandoX7 (approved):

PR #2196, opened 2026-07-20
style(ui): pilot a hand-gilded ornament for the Performance Overlay window

jgyy:
## Summary

Pilots a hand-gilded SVG-mask ornament (corner filigree, noisy gilt edge ribbon, seeded
color-noise gradient) scoped to **one window only**: `#options-menu.perf-wide` (the
Options > Performance Overlay sub-view). The reference is a hand-gilded fantasy-UI mockup
the requester supplied.

Before writing code, I found two things worth flagging up front:

1. **DESIGN.md** (this repo's adopted design-language standard, "no landing restyle
   fragments outside the phase plan") hasn't had any of its section-15 phases land yet, so
   a literal reading blocks any window-level gilded treatment until Phase 1/2 (token
   retune + shared chrome primitive) ships.
2. **PR #2152**, an earlier, much larger "hand-gilded SVG ornament across every HUD
   component" redesign (7 review rounds, fully green CI) was closed without merging the
   same day this session started, most likely superseded by DESIGN.md's adoption.

Given that tension, I asked how to scope this and the answer was: revive #2152's
already-reviewed techniques, but narrowed to just this one window rather than the
all-HUD sweep, and disclosed here as a deliberate, narrow exception rather than a
DESIGN.md-phase-ordered rollout. Concretely that means:

- A new sibling module, `src/ui/perf_ornament_svg.ts`, adapted from #2152's proven
  mulberry32-seeded, integer-harmonic noise techniques (seamless tileable edge ribbon,
  repeating-conic-gradient color noise), with an enriched corner motif (bracket + curling
  acanthus vine + leaf sprigs + gem) closer to the reference's rococo look than #2152's
  final restrained form.
- New, purely additive tokens in `tokens.css`: the exact DESIGN.md section 4.3 gold ramp
  (`--color-gold-900`..`--color-gold-300`) plus `--perf-ornament-*` mask vars. No existing
  token (`--gold`, `--border`, ...) is touched, so nothing outside this one window can
  regress.
- New CSS strictly scoped under `#options-menu.perf-wide` in `components.css`: it never
  touches the shared `.panel`/`.window` primitive, so every other window is byte-identical
  to before this PR.
- The ornament sheds itself at the low effects tier (this repo's existing
  performance-is-part-of-beauty contract), same as the rest of the HUD's cosmetic layer.

```mermaid
flowchart TD
    Boot["main.ts boot (next to hydrateIcons)"] -->|calls once| Apply[applyPerfOrnamentVars]
    Apply --> Corner[perfCornerOrnamentMaskImage]
    Apply --> Edges["perfNoisyEdgeMaskImage x4 (top/bottom/left/right, distinct seeds)"]
    Apply --> Gilt[perfGiltGradientBackground]
    Corner --> VarCorner["--perf-ornament-corner"]
    Edges --> VarEdges["--perf-ornament-edge-top/bottom/left/right"]
    Gilt --> VarGilt["--perf-ornament-gilt"]
    VarCorner & VarEdges & VarGilt --> Rule["#options-menu.perf-wide::before (components.css)"]
    Rule --> Window["Performance Overlay window only — no other window touched"]
    LowFx["data-fx-level=low"] -.->|"!important reset"| VarCorner
    LowFx -.-> VarEdges
```

### Round 2: redesigned against a closer reference, after direct feedback

The requester followed up with a second reference image (a tighter, more legible in-game
shot of a small window's corner) and 5 concrete asks: the frame should sit OUTSIDE the dark
panel's edge, not slightly inside; the top corners were being cut off; the corner shape
should read as one thick, ROUNDED, hand-carved curl, not leaves pointing inward; all 4
corners need genuinely rounded edges; and the gold should read more realistic. All five are
addressed in the follow-up commit:

- **Corner motif fully redesigned**: the vine-plus-leaf-plus-gem composition is gone,
  replaced by ONE compact curl per corner: a single centerline arc (with a small
  radius-only wobble for a hand-drawn feel) stroked 3 times at decreasing length and width
  with `stroke-linecap="round"`/`stroke-linejoin="round"`, which fakes a tapered brush
  stroke without ever producing a sharp point.
- **Frame moved outside the panel edge**: `#options-menu.perf-wide::before` now uses
  `inset: -5px` instead of `inset: 0`, so the gilded frame wraps just outside the dark
  panel the way a picture frame wraps a picture, and the window's own `border-radius` is
  scoped up to 14px (from the shared `.panel`'s 6px) so the rounding reads clearly.
- **A real bug, found and fixed**: the top-left/top-right corners were only partially
  visible because an earlier rule raised `.panel-title`'s `z-index` to 8 to keep its text
  looking gilded — putting its own opaque background ABOVE the ornament's `z-index: 7` and
  silently hiding the top corners behind it. Confirmed live with a forced-z-index probe
  before landing the fix (the title now keeps its base `z-index: 6`, correctly below the
  ornament).
- **Gold palette reweighted** toward the realistic mid-tones (`--color-gold-500/400/600`,
  repeated) with the near-black/near-white extremes kept as rare accents instead of
  equally-likely picks, which had read as muddy/harsh rather than "gold".

Re-verified against the new reference at the raw-SVG level (rasterized corner-by-corner
comparisons) and in the real offline client, all 4 corners, before updating the
screenshots below.

### Round 3: a bigger, outward-facing Baroque acanthus scroll, plus a stray black border fix

A third reference image followed: an elaborate Baroque/Rococo acanthus scrollwork corner
(vector line art), with feedback that the corner radius should be bigger, and the corner
ornament should face OUTWARD from the corner vertex instead of curling inward. Asked to keep
iterating against the reference until it matched, then refresh the screenshots and PR body.

- **Corner radius increased again**: `border-radius` on `#options-menu.perf-wide` went
  14px -> 22px (and the ornament mask-size grew to match), reading closer to the reference's
  visibly generous rounding.
- **Corner motif fully redesigned a second time**: round 2's single thick curl is replaced
  by an actual volute-and-tendril composition, matching the reference's asymmetric L-shaped
  scroll rooted in one spiral: a tightening spiral scroll (outer body stroke plus a shorter
  inner detail stroke, both centered on the corner), with acanthus leaflets (smooth
  quadratic-curve lobes, not straight-edged polygons, which read as arrows/chevrons in an
  earlier attempt) wrapping the spiral's own outer curve. A short leafy tendril runs out
  along the top edge and a longer one runs out along the side edge with its own smaller
  secondary curl partway down, both growing OUTWARD from the pivot the whole way (never
  curling back toward the window's center), with leaves stacked continuously along both
  tendrils rather than clustered only near the scroll. `CORNER_SIZE` grew 30px -> 70px to
  give the larger composition room. Iterated by rasterizing the standalone corner SVG
  against cropped reference zooms repeatedly (fixing an off-canvas geometry bug in the top
  tendril and a malformed-path nesting bug in the leaflet helper along the way) before
  checking it in the full window and the real game.
- **Another real bug, exposed by the sparser line art**: the base `.panel` rule
  (`src/styles/base.css`) hardcodes a literal `outline: 1px solid #000` plus a non-inset
  `0 2px 16px #000c` drop shadow, independent of `border-color`. Round 2's corner was a
  solid block that happened to cover that ring; round 3's filigree leaves real gaps between
  the strokes, letting the outline and shadow show through as a stray black border outside
  the gold frame. Fixed with a targeted `#options-menu.perf-wide` override that drops the
  outline and the outer shadow layer, keeping the two inset shadow layers for depth (so the
  panel keeps its inward shading). Verified clean on all 4 corners, desktop and mobile.

### Two real bugs found and fixed along the way

Iterating against the reference with a screenshot loop, the corners refused to render at
all three of the mirrored positions (only top-left ever showed). Isolated both down to
genuine Chromium `mask-image` rendering bugs, independent of this repo's code:

1. **Mirroring via an SVG `<g transform="scale(-1 ...)">` wrapper**: combining several
   *differing* large inline-SVG `mask-image` data URIs in one multi-layer list, where the
   differing layers relied on a negative-scale transform, made every layer render as the
   FIRST (unmirrored) layer's shape. Fixed by baking the mirror into plain coordinate math
   instead (`perf_ornament_svg.ts`'s `Mirror`/`n2` helper) — no SVG transform anywhere.
2. **Reusing one mask value at two mask-position slots**: the CSS combines 4 edge layers
   (top/bottom/left/right); generating only 2 distinct images (one reused for both top AND
   bottom, one for both left AND right) meant a `mask-image` list with the same literal
   `url(...)` value at two positions only painted the FIRST occurrence. Fixed by generating
   all 4 edges from distinct seeds.

Both are covered by regression tests in `tests/perf_ornament_svg.test.ts` (a coordinate-
mirroring assertion and a "4 distinct edge values" assertion).

## Related issues

N/A — no tracked issue; scoped and reviewed live with the requester.

## Type of change

- [x] Bug fix (two real Chromium mask rendering bugs, see above; covered by regression tests)

## How was this tested?

- Commands: `npm run gate` (green; see note below), `npx vitest run tests/perf_ornament_svg.test.ts`
  (13/13 pass), `npx tsc --noEmit` (clean), `npx @biomejs/biome check` on every changed file (clean).
- Manual steps: iterated visually against all three reference images via a standalone
  (no-game-boot) HTML harness for fast SVG-geometry iteration (including rasterizing
  individual corner SVGs standalone for tight before/after comparisons against the
  reference crops), then confirmed in the real offline client (Options > Performance
  Overlay) at both the default and Ultra effects tiers, desktop and mobile landscape
  viewports, across all three rounds of feedback.
- `npm run gate` note: it stops at `tests/browser/armory_mobile_layout.browser.test.ts`
  (`expected 394.625 <= 391`), confirmed byte-for-byte identical on the unmodified
  `release/v0.29.0` base via `git stash`, so pre-existing and unrelated. Every step after it
  (`typecheck`, `build:env`, `build:server`, `build`) was run individually and is green.

### Round 4: no more black border at the radius, a bigger radius, and outward mid-edge sprigs

A fourth round of feedback, with four concrete asks: no black border at all after the
gilded ornaments; a bigger corner radius, with the ornament itself covering that radius
(the previous round still cut off at the rounded corner); a small ornament pattern at the
midpoint of each edge, in addition to the corners; and those mid-edge ornaments must face
outward, never inward.

- **The stray dark border at the radius, actually fixed this time**: round 3 dropped the
  base `.panel`'s outline and outer shadow, but a thinner dark sliver remained right at the
  rounded-corner boundary itself. Root cause: the corner motif is open filigree, not a
  solid block, so the panel's own dark background showed through the gaps exactly where the
  `::before` frame's outer rounding met the panel's own rounded corner underneath. Fixed
  with a new `cornerRadiusTraceStroke` (`perf_ornament_svg.ts`): a wide band swept along the
  literal `border-radius` curve (`CORNER_RADIUS_TRACE`, pinned to the CSS value in a doc
  comment), so gold ink now covers that boundary continuously on all four corners.
- **Corner radius increased, and the corner mask grown to match**: `border-radius` 22px ->
  30px, corner mask tile 70px -> 84px (`CORNER_SIZE`), with the title padding and close
  button inset widened to match the corner ink's larger reach.
- **A new outward-facing flourish at each edge's midpoint**: `perfMidEdgeOrnamentMaskImage`
  reuses the corner motif's own spiral/leaflet helpers, which work in absolute polar degrees
  around a pivot, so one motif function covers all four edges by rotating an `outwardDeg`
  parameter (top = up, bottom = down, left = left, right = right) instead of needing a
  second coordinate-mirroring scheme. Wired into the existing `--perf-ornament-*` mask stack
  as one more comma-joined layer (`--perf-ornament-mid-edge`), stamped at `center top`,
  `center bottom`, `left center`, `right center`.
- **A real bug on the first attempt, caught before landing**: the first version pivoted the
  mid-edge motif at the tile's dead center, so the bulk of the ink sat well inside the panel
  with only a thin tip ever reaching the edge, reading as growing INTO the window rather
  than facing outward from it (visually confirmed via a screenshot crop, the wrong direction
  the requester specifically flagged). Fixed by anchoring the pivot close to the tile's own
  edge in the outward direction (mirroring how the corner motif's pivot sits right at the
  corner vertex, not the mask box's center), re-verified via another screenshot crop on all
  four edges before landing.

New regression tests in `tests/perf_ornament_svg.test.ts`: the corner motif's stroke count
(now 7, the radius trace added), a mid-edge four-layer/four-distinct-image check, and a
left-right symmetry check on the mid-edge motif (the two side tendrils must sit exactly
equidistant from the pivot, since they are generated as mirrored angle offsets around one
`outwardDeg` axis). `npx vitest run tests/perf_ornament_svg.test.ts
tests/perf_ornament_header_clearance.test.ts` and `npx tsc --noEmit` both green; confirmed
the one pre-existing `tests/localization_fixes.test.ts` failure (unrelated sim i18n gap
introduced by the release-branch sync merge into this branch, not by this round's diff) is
identical on the unmodified branch tip via `git stash`.

## Screenshots / recordings

Desktop and mobile-landscape, before/after, captured via `scripts/pr_screenshots.mjs` (a new
`perf-overlay-ornament` target added to `scripts/pr_shot_targets.mjs`, desktop + mobile
variants) at the Ultra effects tier so the ornament (which sheds at low tier by design) is
visible. Updated after round 4's fixes (the images below are current):

| | Before | After |
|---|---|---|
| Mobile (landscape) | before-mobile | after-mobile |

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green except the one confirmed
      pre-existing, unrelated browser-regression failure noted above; I added decisive
      regression tests for both real bugs found.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified in both a desktop viewport and a
      phone-sized landscape viewport (mobile is landscape-only in-game). No new touch
      targets or inputs are added by this change.
- [x] **Accessible.** Purely decorative (`pointer-events: none`, a `mask-image` finish);
      adds a `forced-colors: active` fallback restoring a plain system-color border so the
      window frame survives high-contrast mode. No focus/keyboard/ARIA surface changes.

### Localization (i18n)

### Hygiene


FernandoX7 (commented):
Solid pilot, and the scope containment claimed is the containment shipped: every new CSS rule is `#options-menu.perf-wide`-prefixed, the tokens are purely additive and match DESIGN.md 4.3's gold ramp hex for hex (so a future Phase 1 can adopt the names verbatim), and `main.ts` stays a one-line call into the new sibling module. The colorless-mask plus var()-driven color design means a token retune repaints the ornament for free, and the true-mirror coordinate assertion in `tests/perf_ornament_svg.test.ts` pins the round-1 bug properly instead of settling for "the four strings differ". play.html loads the same `src/main.ts` and stylesheet import, so there is no parity gap. CI is green across the board.

Two should-fix items, both on configurations the testing note says were not covered (it lists the default and Ultra tiers on the classic theme):

- **should-fix** `src/styles/components.css:9496`: `#options-menu.perf-wide .panel-title { color: var(--color-gold-400) }` puts static light gold over the themed `--panel-base` title background. On the parchment preset the panel knob is `#ece0c4`, so the title is light-gold-on-cream, far below AA. This is the exact case `src/ui/theme.ts` documents and repairs (`themeCssVars` derives `--color-gold` via `ensureReadable` against the panel because "a bright accent on the light Parchment panel is gold-on-cream (sub-AA)"), and DESIGN.md 4.3 scopes the static gold ramp to "never text". Using the contrast-repaired `var(--color-gold)` for the title text, or scoping the static-gold override to dark presets, keeps the gilded read on classic/midnight without losing the title on parchment. Softer version of the same question for the frame itself on parchment (gold ornament over a cream panel); worth one screenshot on that preset before calling it done.

- **should-fix** `src/styles/components.css:9403` plus `src/styles/tokens.css:260`: at `data-fx-level="low"` the ornament sheds itself (correctly), but the `border-color: transparent`, `outline: none`, and outer-shadow removal on the window are unconditional. Low-tier users therefore get no frame at all: no border, no outline, and no ornament to replace them. That is a regression against base for exactly the devices that auto-detect low (the new shot-target comment itself notes this sandbox lands on low under software rendering). The forced-colors block already shows the shape of the fix: a `:root[data-fx-level="low"] #options-menu.perf-wide` border restore next to the var reset.

Nits:

- nit `tests/perf_ornament_header_clearance.test.ts`: the paddings are pinned, but the round-2 bug (raising `.panel-title` above the ornament's z-index 7 silently hides the top corners) and the round-3 outline drop have no pin. Those two are the regressions most likely to recur from an innocent-looking edit; a text pin on each would close the loop.
- nit `src/ui/perf_ornament_svg.ts:90`: the comment says "9 repeats" but 360/20 is 18. Also, since 20 divides 180, diagonally opposite corners always land on the same point of the color cycle, so the "each corner lands on a different point" intent only holds for adjacent corners. Comment-only, but this constant block is what future tuning will read.
- nit: the PR body carries a handful of em dashes ("N/A — no tracked issue", the round-2 z-index paragraph); the house dash ban covers PR text too.

On the DESIGN.md tension: the disclosure matches the diff. Nothing shared is touched, the ramp is the adopted 4.3 values, and the whole feature sits behind one selector, so if Phase 1/2 later supersedes it the removal is one deletion. Whether the exception stands is the maintainer's call, but nothing hidden turned up.

jgyy:
Round 4 pushed, addressing all four asks:

- The black border at the radius is gone: a new `cornerRadiusTraceStroke` sweeps a band along the exact `border-radius` curve so gold ink covers that boundary continuously (round 3's fix dropped the outline/outer-shadow but a thinner sliver from the panel's own rounded corner peeking through the filigree's gaps remained; this closes it for real).
- Corner radius increased (22px -> 30px), with the corner mask grown to match (70px -> 84px) so the ornament's own reach covers the bigger radius instead of cutting off at it.
- A small ornament now sits at the midpoint of each edge (top/bottom/left/right), reusing the corner motif's spiral/leaflet vocabulary via a rotated `outwardDeg` angle rather than a second mirroring scheme.
- Those mid-edge ornaments face outward: the first attempt pivoted them at the tile's center, which (caught via a screenshot crop) grew mostly into the panel with just a thin tip reaching the edge, backwards from the ask. Re-pivoted near the tile's own edge (like the corner's pivot sits at the vertex) and reverified on all four edges.

New regression tests cover the radius trace's stroke count, the mid-edge four-layer mask, and the mid-edge motif's left-right symmetry. `npx tsc --noEmit` and the ornament test files are green; the one `tests/localization_fixes.test.ts` failure is a pre-existing, unrelated sim i18n gap from the release-branch sync merge, confirmed identical on the unmodified branch tip.

PR body and screenshots updated.

PR #2200, opened 2026-07-20
perf(server): cache hot reads, add retention sweeps, share broadcast work, move to Node 26

FernandoX7:
## Summary

Server performance work: keeps per-request and per-tick cost flat as tables and
crowds grow, with gameplay untouched. Rewrites are pinned output-identical or
byte-identical by tests; the few deliberate, scoped behavior changes are called
out under Behavior notes below.

- **Read caches with single-flight.** The daily-rewards status board, admin
  overview counts, deeds rarity aggregate, leaderboard and guild boards, the
  arena ladder (per format), and the public project stats now serve from
  short-TTL caches with in-flight request coalescing (`server/cached_read.ts`
  plus the epoch-keyed `singleFlight` in `server/deeds_board_warm.ts`).
  Moderation actions bust the affected boards, and a reader arriving after the
  bust never joins a pre-bust refresh (the exact visibility windows are under
  Behavior notes).
- **Retention.** A nightly advisory-locked sweep (`server/retention_sweep.ts`)
  with batched prunes for chat logs, client perf reports, daily reward events,
  admin online samples (with an all-time peak fold), site presence,
  play sessions (folded forward into daily rollups so lifetime stats are
  preserved), account IP association aging, and the player activity fact
  table. All windows are env keys (unset keeps data forever), documented in
  `.env.example` and `DEPLOY.md`.
- **Query rewrites.** The lifetime-XP leaderboard arms now match their index
  expression exactly; the daily-reward excluded-accounts view is OR-free UNION
  arms; the admin client-perf summary collapsed from a multi-statement
  transaction into one GROUPING SETS statement (differential-tested identical);
  a finalized-day guard stops settled reward days from being recomputed. New
  indexes build post-boot via `CREATE INDEX CONCURRENTLY`
  (`server/concurrent_indexes.ts`).
- **Broadcast.** World events serialize once per batch and share bytes across
  sessions (`server/event_frame.ts`); the Vale Cup readout and the
  dungeon-finder board build and stringify once per realm pass
  (`server/realm_readout_memo.ts`); interest gathering scans each occupied grid
  cell once instead of once per viewer (`server/interest_candidates.ts`). Wire
  output is byte-identical, pinned by the bandwidth and snapshot suites plus
  golden parity.
- **Runtime.** The server image moves to Node 26 on a glibc slim base; the
  deterministic sim parity suite is the upgrade oracle.
- **Docs.** The conventions this work established are recorded in
  `server/CLAUDE.md` ("Hot paths"), the root `CLAUDE.md` seam list, and
  `tests/CLAUDE.md`, so future contributions keep the same shape.

## Behavior notes (deliberate, scoped)

- **Daily-reward exclusion coverage.** The rewritten excluded-accounts view
  gains a fourth arm over `account_ip_associations` so IP-based exclusion keeps
  working once old play sessions fold into daily rollups; the retention windows
  bound the lookback. This is a scoped eligibility-coverage change to preserve
  ban semantics across retention, not a pure query rewrite.
- **Vale Cup wire cadence.** The send gate moved from per-session offsets to
  one realm-global dueness tracker, so a changed shared readout reaches every
  settled viewer on the same broadcast pass instead of staggered across the
  interval. The aligned burst is bounded: one build and one stringify per pass
  with raw re-sends of the same string, and unchanged fragments stay
  delta-elided. Both halves are test-pinned (`tests/vale_cup_online.test.ts`,
  including an aligned-fan-out pin added after review); the real fan-out cost
  at peak lands in the post-deploy measurement (#2201).
- **Moderation visibility windows.** A bust takes effect immediately for any
  reader arriving after it in the busting process; a reader that had already
  joined a refresh in flight before the bust still receives that flight's
  pre-bust value once (documented in `server/cached_read.ts`); peer realm
  processes converge within the board TTL (30 seconds). None of this changes
  ban enforcement, only how long a just-moderated entry can remain visible on
  a cached board.

## Type of change

- [x] Tests
- [x] Build, CI, or tooling (Node 26 base image)

## How was this tested?

- Commands: `npm run gate` (full pre-merge gate, all steps green); per-subsystem
  vitest suites including golden parity with zero re-baselines; real-Postgres
  integration suites for the retention paths (`TEST_DATABASE_URL` against the
  dev database); a live-Postgres differential proving the collapsed client-perf
  summary returns byte-identical results to the retained multi-statement spec;
  byte-identity pins for every broadcast change.
- Manual steps: none; every change is covered by the automated suites above. A
  post-deploy measurement pass is planned to confirm the wins on production
  traffic (see the follow-up issues).

---

### Quality

### Cross-platform

- ~Tested on desktop and mobile.~ Server-only change; the one UI-adjacent piece
  (an admin dashboard stat card) is covered by the admin component suite.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** One new admin
      dashboard label (English added to `src/admin/i18n.en.ts`, resolved bundle
      regenerated); no player-facing game strings added.

### Hygiene

- [x] I didn't hand-edit generated files. They're regenerated by the build.

---

## Follow-ups (tracked, deliberately out of this PR)

- #2201 Post-release measurement pass and the Postgres placement decision
- #2202 Gate the professions and card-duel self fragments on heavy-self change tracking (evidence-gated on #2201)
- #2203 Cheapen the admin overview counts query (needs an acceptance on reported numbers)
- #2204 Revive the parked performance observability slice once a metrics scraper exists
- #2205 Small performance housekeeping follow-ups


jgyy (commented):
This is a large, well-organized piece of work and it holds up under a close read. The single-flight cache layer, the retention sweep, the query rewrites, and the serialize-once broadcast changes are all backed by real behavioral-identity tests (byte-identical wire assertions, live-Postgres differentials, the Vale Cup shared-info deep-equal pin), not just "does not throw" coverage. A few points worth a look before merge, none of them blocking.

**Should-fix**

- `server/db.ts` (`daily_reward_excluded_accounts` view, `DAILY_REWARD_EXCLUDED_ACCOUNTS_VIEW_SQL`): the OR-join was split into UNION arms and a new `account_ip_associations` arm was added. This is a genuine superset widening of who gets excluded from daily rewards (folded-away sessions now still ban by IP), not just a query rewrite. It's deliberate, needed to keep ban coverage correct once `play_session_retention_db.ts` folds sessions away, and it's well tested (`tests/daily_rewards_ban_db.test.ts`, `tests/schema_wiring.test.ts`). Worth calling out explicitly in the PR description as a scoped behavior change, since the summary currently frames everything as "no behavior change."

- `server/game.ts` (`vcupDue`, around the broadcast-pass rewrite replacing per-session `lastVcupWireTick`): the Vale Cup wire-send gate moved from a per-session offset to a realm-global `>=` check, so all sessions now emit `vcup`/`vcupb` on the same tick instead of staggered across the interval. Cadence and first-snapshot immediacy are preserved, and this is the right tradeoff for the shared-readout-memo win, but it is a real send-timing change (bursty instead of amortized), not strictly "no behavior change." Confirm the synchronized burst is fine at peak concurrency, since it doesn't seem to be covered by a test that checks fan-out timing under load.

**Nit**

- `server/cached_read.ts`: a reader that joins an in-flight refresh just before a moderation `bust()` still gets that flight's pre-bust value, so a moderated entry can show for a sub-second window. This looks like an accepted, documented tradeoff (readers arriving after the bust correctly get a fresh flight) rather than a bug, but since it's player-visible moderated content it's worth a one-line callout in the PR body next to the "moderation actions bust the affected boards immediately" claim, since "immediately" is very slightly optimistic here.

- `server/daily_rewards_board_cache.ts`: peer realm processes converge on a moderation change only within the board's TTL (up to ~30s), while the busting process updates immediately. Same category as above: fine, but worth being explicit about in the write-up since it's moderated content.

Everything else checked out clean: idempotent/additive DDL throughout (including the `CREATE INDEX CONCURRENTLY` self-heal in `concurrent_indexes.ts`), the retention sweep's fold-then-delete is atomic per batch so a crash mid-run can't lose data, `finalizeDay`'s `ON CONFLICT (day, realm, rank) DO NOTHING` genuinely protects against a double-finalize race independent of the in-process `daily_finalize_guard`, `interest_candidates.ts`'s cell-based rewrite is a provable superset re-filtered per viewer before `canObserveEntity` so nobody's visibility changes, and the Node 26 / glibc-slim base bump is consistently pinned across Dockerfile/CI/docs with a real cross-file test. `src/sim/` stays pure and untouched behaviorally (the `vale_cup.ts` split is a faithful extraction, not a rewrite), and the one new admin-dashboard string is wired through `t()` correctly with no hand-edited generated files.

Nice work on this one.

FernandoX7:
Thanks for the close read. All four points were real; here is how each was addressed.

- Excluded-accounts view and the Vale Cup send alignment: both were deliberate, but the description oversold "no behavior change". Added a "Behavior notes (deliberate, scoped)" section calling out each one explicitly, including why the fourth view arm exists (preserving ban coverage across the play-sessions fold) and what the cadence alignment trades.
- Synchronized burst at peak: the bounded shape is now pinned by a new test in tests/vale_cup_online.test.ts ("ships a changed shared readout to every settled session in the same due pass, byte-identical"): one build and one stringify serve every settled viewer in the same pass, each raw frame embeds the one memoized string, and the existing pins keep unchanged fragments delta-elided. So the worst case a due pass emits is N raw sends of one string, never N rebuilds. The wall-clock fan-out cost under real peak concurrency is exactly what the post-deploy measurement window in #2201 reads before the Postgres placement decision.
- cached_read pre-bust joiner window and the cross-process TTL convergence: both are accepted, documented tradeoffs (the joiner window is spelled out in server/cached_read.ts, the TTL window was a recorded design choice). The description now states both plainly in Behavior notes instead of implying "immediately" covers them.

PR #2210, opened 2026-07-20
fix(server): atomically persist character and escrow state after market and mail mutations

Wmedrado:
## Summary

Market and mail commands mutate two pieces of persisted state in the same command: the acting character's own bags and copper, and the shared market/mail `world_state` blob. Today those two halves are only ever written by the 30s autosave, which fires them as three independent fire-and-forget calls:

```ts
// server/game.ts, flushPeriodicSaves
void this.saveAll('autosave');
void this.saveMarket();
void this.saveMail();
```

Nothing synchronizes them, so a crash or a hard restart in between can bring the world back with the character and the escrow taken from different instants. The two cases that hurt:

- **`market_buy`.** If the buyer's character save (item granted, copper deducted) commits and the market blob write does not, the listing comes back after the restart. The item is duplicated and the seller's proceeds are gone.
- **`mail_take`.** The mirror window: the letter's coin and items land in the taker's bags, the mail blob still holds the letter, and taking it again after the restart duplicates them.

The repo already has the right primitive for this. `saveCharacterAndMarketState` (`server/db.ts`) writes the character row and both `world_state` keys inside one Postgres transaction, and the leave path already routes through it via `saveCharacterOnLeave` -> `saveCharacter(session, { withMarket: true })` for exactly this reason. The gap is that a player who never logs out never gets that atomic write.

This change calls that same atomic write promptly after each escrow-mutating command (`market_list`, `market_buy`, `market_cancel`, `mail_send` on both the online and offline-recipient branches, and `mail_take`), instead of only when the player happens to leave.

Two details worth calling out:

- **Debounced 500ms per character**, so a burst from one player (a few quick mail takes, a list followed by a cancel) coalesces into a single write rather than one write per command.
- **Fire-and-forget**, matching every other save call in `game.ts`. The command's own dispatch is never blocked by a slow or failed save, so the 20 Hz loop is untouched.

The 30s cadence stays exactly as it is, as the fallback for everything else. The pending timer is cleared in `leave()` (the leave flush already performs the same atomic write, so the debounced one would be redundant) and in `stop()`.

## Related issues

None that I found; I ran into this reading the persistence path.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/escrow_atomic_save.test.ts` (new, 6 cases, all pass)
  - `npx vitest run tests/save_character_and_market.test.ts tests/professions_market.test.ts tests/character_lease.test.ts tests/mail.test.ts tests/market.test.ts` (71 pass, the existing coverage around the same writer and the leave path)
  - `npx tsc --noEmit` (clean) and `npx @biomejs/biome check server/game.ts tests/escrow_atomic_save.test.ts` (clean)
- Manual steps: N/A, this is a persistence-timing change with no player-visible surface. The new tests drive the real `GameServer` and the real `Sim` through the actual command path (`handleMessage`) rather than mocking the seam.

I confirmed the new tests are decisive rather than tautological: against an unmodified `server/game.ts` all 6 fail (`expected 0 to be greater than 0`, no atomic write is ever scheduled), and all 6 pass with the change.

One caveat on the full suite, since I would rather flag it than quietly tick the box. I develop on Windows, and a full `npx vitest run` on an **unmodified** checkout of `release/v0.29.0` already fails 11 files / 25 tests here for environment reasons (the SFX studio suites want FFmpeg, `static_sfx_serving` hits `EPERM` on a Windows rename, `new_endpoint` gets a null status from its `tsc` spawn, plus `asset_pipeline` and `prod_cpu_monitor`). With this change applied the run is 9 files / 23 tests, a strict subset of that same baseline, so the change introduces no new failures and adds 6 passing tests. Nothing in the remaining set touches `server/`, the market, mail, or persistence. Your Linux CI should be the real verdict here.

The new `tests/escrow_atomic_save.test.ts` pins:

- each of `market_list`, `market_buy`, `market_cancel`, `mail_send`, `mail_take` schedules an atomic save for the acting character;
- the save does not run inline on the hot dispatch path (it is still pending immediately after the command resolves);
- a recipient who took no action of their own gets no save scheduled;
- a burst of three `mail_take`s inside one debounce window coalesces into exactly one write (a non-debounced implementation reads 3 here).

## Screenshots / recordings

N/A, no player-visible or operator-visible surface changes.

---

### Quality

- [x] **The full gate passes.** Typecheck, Biome, and the whole test suite are green apart from the pre-existing Windows-environment failures described above, which reproduce identically on an unmodified base branch. I added decisive tests for the changed behavior (verified failing without the fix) and recorded the commands above.

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -230,6 +230,12 @@ const LOCKPICK_ACTIONS = new Set<PickAction>(['hardSet', 'set', 'steady', 'ease'
 const LEAVE_SAVE_MAX_ATTEMPTS = 5;
 const LEAVE_SAVE_RETRY_BASE_MS = 250;
 const LEAVE_SAVE_RETRY_MAX_MS = 4000;
+// How long an escrow-mutating market/mail command waits before its prompt
+// atomic save (scheduleEscrowSave) actually runs. Short enough to close the
+// crash window between it and a restart, long enough that a burst of commands
+// from one character (a few quick mail takes, list-then-cancel) coalesces
+// into a single write instead of one per command.
+const ESCROW_SAVE_DEBOUNCE_MS = 500;
 // Usage notices for the two PLAYER chat-suppression tiers. Kept as constants
 // because the S3 localization guard scans sendChatNotice literals, and
 // src/ui/server_i18n.ts carries the matching rules. (A "mute" is the ADMIN
@@ -1242,6 +1248,9 @@ export class GameServer {
   private socialPosTimer = 0;
   private saveAllInFlight: Promise<void> | null = null;
   private readonly characterSaveQueues = new Map<number, Promise<void>>();
+  // Debounce timers for the prompt atomic escrow save (scheduleEscrowSave),
+  // one per character id currently awaiting its coalesced write.
+  private readonly escrowSaveTimers = new Map<number, NodeJS.Timeout>();
   // Weapon-skin loadouts are whole-record replacements in their dedicated paid
   // state row. Keep one FIFO per account so rapid apply/detach commands cannot
   // commit on separate pool clients in reverse order and resurrect stale state.
@@ -1942,6 +1951,42 @@ export class GameServer {
     }
   }
 
+  // Market buy/list/cancel and mail send/take each mutate BOTH the acting
+  // character's own state (bags/copper) and the shared market/mail world_state
+  // blob in the same command. flushPeriodicSaves above persists those two
+  // pieces as INDEPENDENT unsynchronized 30s writes (saveAll, then saveMarket/
+  // saveMail), so a crash between them can restore the character and the
+  // escrow from different instants: a market_buy whose buyer character save
+  // (item granted, copper deducted) lands before the market blob write that
+  // removes the listing resurrects the listing after restart (item
+  // duplicated, seller's proceeds lost); mail_take has the mirror coin/item-
+  // dupe window. The leave path was already made atomic for exactly this
+  // reason (saveCharacterOnLeave -> saveCharacter(withMarket:true), which
+  // writes both through saveCharacterAndMarketState in one transaction).
+  // Call this right after such a command so that SAME atomic write runs
+  // promptly for the acting character instead of only ever riding the
+  // unsynchronized 30s pair (kept as the fallback for the everyday case).
+  // Debounced per character, and fire-and-forget like every other save call
+  // in this file: a burst of commands from one character (a few quick mail
+  // takes, list-then-cancel) coalesces into a single write, and a slow or
+  // failed save never blocks the command's own dispatch.
+  private scheduleEscrowSave(session: ClientSession): void {
+    const characterId = session.characterId;
+    const pending = this.escrowSaveTimers.get(characterId);
+    if (pending) clearTimeout(pending);
+    this.escrowSaveTimers.set(
+      characterId,
+      setTimeout(() => {
+        this.escrowSaveTimers.delete(characterId);
+        const live = this.sessionsByCharacterId.get(characterId);
+        if (!live || live.left) return; // logged off before the debounce fired
+        void this.saveCharacter(live, { withMarket: true }).catch((err) =>
+          console.error(`escrow save failed for ${live.name}:`, err),
+        );
+      }, ESCROW_SAVE_DEBOUNCE_MS),
+    );
+  }
+
   private enforceJailStates(): void {
     for (const session of this.clients.values()) {
       this.applyModeratorJailGate(session);
@@ -2033,6 +2078,8 @@ export class GameServer {
     if (this.playtimeInterval) clearInterval(this.playtimeInterval);
     if (this.dailyRewardActivityInterval) clearInterval(this.dailyRewardActivityInterval);
     if (this.keepaliveInterval) clearInterval(this.keepaliveInterval);
+    for (const timer of this.escrowSaveTimers.values()) clearTimeout(timer);
+    this.escrowSaveTimers.clear();
   }
 
   // Grant playtime reward points to each online account that has been ACTIVE (gave
@@ -2947,6 +2994,15 @@ export class GameServer {
     if (session.jailVisit) this.exitJailVisit(session, false);
     session.left = true;
     this.clients.delete(session.pid);
+    // The leave-flush below (saveCharacterOnLeave) already runs the same
+    // atomic character+market+mail write; drop any still-pending debounced
+    // escrow save for this character so it never fires a redundant write
+    // against a session that is already gone.
+    const escrowTimer = this.escrowSaveTimers.get(session.characterId);
+    if (escrowTimer) {
+      clearTimeout(escrowTimer);
+      this.escrowSaveTimers.delete(session.characterId);
+    }
     this.botDetector.releaseTrackingContext(session.botTrackingContext);
     this.releaseIpSession(session.ip);
     void this.recordOnlineSnapshot();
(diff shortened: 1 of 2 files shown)


EnriqueGF (changes requested):
Thanks for this: the diagnosis is right, the primitive you reuse (`saveCharacterAndMarketState`) is the correct one, and the tests genuinely drive `handleMessage` against a real `GameServer` and `Sim` rather than mocking the seam. The comment block is unusually good at explaining why the write exists.

What I ran on Linux, on top of `35c85b312`:

- `npx tsc --noEmit`: clean.
- `npx vitest run tests/escrow_atomic_save.test.ts tests/save_character_and_market.test.ts tests/mail.test.ts tests/market.test.ts tests/character_lease.test.ts`: 72 pass.
- Guards: `tests/architecture.test.ts`, `tests/world_api_parity.test.ts`, `tests/localization_fixes.test.ts` all green (the S3 guard needs `npm run i18n:gen` first; that failure was a stale generated artifact, not your change).
- `npx @biomejs/biome check server/game.ts tests/escrow_atomic_save.test.ts`: clean. No em/en dashes or emoji in the added lines. No sim, wire, `IWorld`, i18n or schema surface touched, so parity and migration safety are not in play.

I could not reproduce your Windows failures here; the ones you list are environmental and none of them touch this path.

Three things I would like addressed before this merges.

**1. `market_collect` has the same tear window and is not covered (blocking).**

`src/sim/market.ts` `marketCollect` does `meta.copper += col.copper`, `ctx.addItem(...)` into the character, then `this.marketCollections.delete(...)` from the shared blob (collections are serialized into the market snapshot). That is exactly the class of tear this PR closes, in the most valuable direction: a crash after the character save and before the blob write resurrects the collection while the gold and items are already in the player's bags. Every seller's proceeds and every expired-listing return duplicate.

`server/game.ts` `case 'market_collect'` still reads `sim.marketCollect(pid); break;` with no `scheduleEscrowSave`. I confirmed this empirically with a throwaway test built on your own harness: seed a collection of 5000 copper plus one item, send `market_collect`, and the character gets both while `savesFor(characterId).length` is 0 after running pending timers. Please add the call and a matching case in `tests/escrow_atomic_save.test.ts`.

For the record, the paths you correctly left out: `mail_delete` refuses any letter still carrying coin or parcels, so it is blob-only; `mail_read`, listing and letter expiry, and house restock have no character side.

**2. The write is client-triggerable and uncapped, on the same lane as the logout flush.**

`scheduleEscrowSave` runs unconditionally, before knowing whether the sim accepted anything. Every one of `marketList`, `marketBuy`, `marketCancel`, `mailTake` returns `void` and bails early on failure, so a bogus listing id, an out-of-range `mail_take`, or a cancel on someone else's listing all schedule the full transaction. The only throttle is the generic frame bucket (`MSG_RATE_BURST = 60`, refill 40/s) plus your 500ms debounce, so one client can hold itself at 2 writes/s indefinitely with zero state change.

Each of those writes is one character row plus **both** whole `world_state` blobs, and `serializeMarket()` / `serializeMail()` run inside the thunk, on the loop thread, once per queued item. They all funnel through the single global `enqueueMarketWrite` FIFO (`createSerialWriter`: `tail = run.catch(() => {})`, no depth cap, no coalescing, no back-pressure since every producer is `void`-ed). `leave()` and the shutdown flush enqueue on that same tail, and `leave()` awaits it before releasing the lease. Cancelling the pending timer in `leave()` does not help once the timer has already fired. So the failure mode under a busy market is a backlog that delays logout saves behind it, with the lease TTL and the container stop grace as the things that give.

The cheapest fix removes most of the load at once: give the sim a monotonic escrow revision (bumped where the listing and mail arrays are actually mutated), snapshot it around the `sim.*` call, and only schedule when it changed. A global coalesce for the blob half (one flush in flight, blobs serialized once per flush rather than once per queued item) would be the belt to that suspenders.

**3. Two smaller correctness points.**

- The debounce is pure trailing with no max wait: every command `clearTimeout`s and re-arms. A player issuing escrow commands closer than 500ms apart (a mailbox with a dozen parcels, a market flipper) never reaches the timer at all and falls back to the old unsynchronized 30s pair. The heaviest trader gets the weakest guarantee, which inverts the intent. A max-wait cap or a leading edge plus cooldown holds it.
- The write is atomic for the **acting** character only. `saveCharacter(withMarket)` commits that character's row alongside a realm-wide market and mail snapshot. If B has just listed an item (gone from B's in-memory bags, present in the blob) and A's escrow save lands first, the committed blob carries B's listing while B's last durable row still holds the item: a crash there duplicates it. The 30s pair bounded how often that window was entered; frequent global blob commits widen it, and point 3's starvation can keep B's own save from closing it. I do not think a full fix belongs in this PR, but the header comment currently reads as if the tear class is closed, so please say plainly that the guarantee is per acting character.

Things I checked and found fine, so you do not need to re-verify them: the relog-within-500ms case is safe (`leave()` is the sole site that removes from `clients` and `sessionsByCharacterId`, and the callback re-resolves the live session and saves it with its own pid and `leaseNonce`); the offline-recipient `mail_send` arm re-checks `this.clients.get(pid) !== session` after every await in the same synchronous run as the schedule, so it cannot orphan a timer past `leave()`; and the lease fence still works, with the side effect that a displaced session is now kicked on its next escrow command instead of at the next autosave, which reads as an improvement.

On tests, once the above lands I would also want: a payload assertion (at least once, the saved `state` lacks the escrowed item while the same call's market snapshot contains the listing, which is the actual invariant rather than the call count), the `leave()` cancel arm, the offline `mail_send` arm, and the logged-off early return. `saveMarketState` and `saveMailState` are imported by `server/game.ts` but missing from the `vi.mock` factory; harmless today, confusing the day a path touches them.

Happy to re-review as soon as `market_collect` and the scheduling gate are in.

Rubsey (commented):
## OSSBrain initial review

### Blocking findings

- **market\_collect is not covered by the atomic escrow save** (`server/game.ts:4876`)
  - Issue: The PR schedules the new debounced atomic character plus market/mail save for market\_list, market\_buy, market\_cancel, mail\_send, and mail\_take, but leaves market\_collect unchanged. market\_collect calls sim.marketCollect(pid) without scheduleEscrowSave(session). The sim path moves collection copper/items into the player character and then zeroes or deletes the shared market collection, so it mutates both halves of the same escrow boundary.
  - Impact: If the periodic character save lands after a collection but the market world\_state save does not land before a crash or restart, the player keeps the collected copper/items while the market collection is restored and can be collected again. This is the same item/currency duplication class the PR is intended to close.
  - Suggested fix: Schedule the same atomic escrow save after a successful market\_collect mutation, and add a focused test that seeds a seller collection, sends market\_collect, drains the debounce, and asserts saveCharacterAndMarketState is called for the collector.

### Checks

- `sha256sum .ossbrain-review/pr-2210/woc-review-pr.SKILL.md`: **passed**. Matched supplied skill checksum b958f04c641fb33b3a96cee949016a9f7a6fcc0f497ed2b12823216880a127fa.
- `git diff --name-only 35c85b31244125328f777f314ed7f948130ce9cc..2f24cee7a6d4fb6fe6855cdac70de758efc05b52`: **passed**. Original PR branch diff from its merge base touched server/game.ts and tests/escrow\_atomic\_save.test.ts.
- `VITE\_CACHE\_DIR=/opt/ossbrain/work/world-of-claudecraft/.wt/wt-pr2210.review/.ossbrain-review/pr-2210/vite-cache npx vitest run tests/escrow\_atomic\_save.test.ts`: **not run**. Vitest could not start because Vite attempted to write its bundled config under read-only node\_modules/.vite-temp.

PR #2211, opened 2026-07-20
feat(onchain): ecosystem activity bot for $WOC burns, sales, and Claudium

awidearray:
## Summary

Adds the publishing spine for a kintarasales-style on-chain activity bot: whenever the $WOC ecosystem transacts on Solana, it announces the event to three surfaces at once, in-game realm chat, a dedicated Discord channel, and X (Twitter).

Three event kinds are supported: **$WOC burns**, **item sales settled in $WOC**, and **Claudium purchases** (WOC / SOL / USDC). Mainnet only.

This PR is the game-server + bot half (the publishing pipeline). The chain-watch worker that detects events and calls the ingress lands separately in the economy service; today only burns produce mainnet transactions, and the sale/Claudium detectors light up automatically when an operator enables those rails, with no further code changes here.

## Pipeline

```
worker  --POST /internal/onchain-event-->  game server  --broadcastSystem-->  realm chat
        (x-woc-onchain secret)                          --enqueue--> GET /internal/onchain/feed --> bot --> Discord embed + X tweet
```

One normalized event, one ingress on the game server, three sinks. This mirrors the repo's existing enqueue-on-server / drain-and-post-in-bot idiom (`discord_activity.ts` + `/internal/discord/activity` + the bot `pollActivity` loop).

## What is here

- **`server/onchain_activity.ts`** (new, pure): the `OnchainEvent` type and a bounded, signature-deduped queue, mirroring `discord_activity.ts`.
- **`server/onchain_feed.ts`** (new, pure): inbound-body validation (fail-closed) and the ASCII, emoji-free realm-chat line renderer.
- **`server/internal.ts`**: `POST /internal/onchain-event` (new `x-woc-onchain-secret` gate) validates, dedupes, broadcasts to realm chat above a configurable USD floor (`WOC_ONCHAIN_REALM_MIN_USD`, so dust burns do not spam chat), and enqueues; `GET /internal/onchain/feed` (Discord-bot secret) drains for the bot. Both routes land on both dispatch arms with the characterization spine updated (surface inventory, completeness, ownership sweep).
- **`server/game.ts`**: a thin public `announceOnchain(line)` that fans the line to connected clients as a yellow system log.
- **`src/ui/server_i18n.ts`**: an identity matcher rule for the `[WOC]` realm line (fully dynamic, locale-neutral content; English-only is an accepted, documented limitation).
- **`bot/logic.ts`**: `buildOnchainMessage` (per-kind Discord embed with amount, USD, actor, total-burned, Solscan link) and `buildOnchainTweet` (<=280 X-weighted, ASCII).
- **`bot/x_api.ts`** (new): a zero-dependency OAuth 1.0a signer (node `crypto`) + `postTweet`, gated by `X_ONCHAIN_ENABLED` plus the four X credentials. **X posting is fail-closed off** until an operator opts in.
- **`bot/main.ts` / `config.ts` / `server_client.ts`**: the `pollOnchain` loop (drain, post to `DISCORD_ONCHAIN_CHANNEL_ID`, tweet when enabled, dedupe by signature).
- **`.env.example`**: `WOC_ONCHAIN_SERVICE_SECRET`, `WOC_ONCHAIN_REALM_MIN_USD`.

## Tests (all green)

New coverage proves the path end to end: the ingress route through the real middleware onion (validate, dedupe, broadcast, queue), realm-chat delivery to connected clients (`announceOnchain` -> yellow system log on every socket), the queue dedupe/cap, the render + validation, the Discord/X builders (embed shape, 280-char weighted budget, ASCII), the OAuth 1.0a signature against an independent vector, and the `[WOC]` i18n rule across every locale. Full `tests/server/` + bot + i18n suites pass (2322 assertions), `tsc` clean, changed-file biome clean, both esbuild bundles build, and the pre-push QA floor is green.

## Notes

- Mainnet only by request: sales/Claudium emit nothing until an operator enables those rails; no code change needed when they do.
- Realm-chat is the highest-threshold sink to avoid in-game spam; Discord/X get the fuller feed.
- X posting is public and autonomous once enabled; it ships fail-closed off so enabling it (credentials + flag) is an explicit operator choice.
- No new dependencies (the X client is `node:crypto` + `fetch`).


FernandoX7 (commented):
The seam work here is right. Two pure modules mirroring `discord_activity.ts`, a thin `announceOnchain` on `GameServer` so the route module never holds the live instance, `InternalRuntime` widened with `Pick<>` rather than passing the server around, and a dedicated `WOC_ONCHAIN_SERVICE_SECRET` instead of reusing the deploy or Discord secret, with the reason written down at `server/http/middleware/require_internal_secret.ts:27-34`. I diffed the two dispatch arms case by case (feature-off, mismatch, valid body, malformed body, wrong method, HEAD, OPTIONS) and they come out identical, constant-time compare and per-request env read included. X posting shipping fail-closed off behind the flag plus all four credentials is the right default, and `allowed_mentions: { parse: [] }` is correct.

## Blocking

**Base is `release/v0.28.0`, which is not on the path to the current release.** `release/v0.30.0` is the active integration base. v0.28.0 is not an ancestor of it and never landed in `main`, and v0.30.0 carries 564 commits this head does not have. The pins this PR moves have already moved upstream: `tests/server/http/completeness.test.ts` pins `internalLadder.length` at 19 on v0.30.0 (this PR takes 18 to 20), and `ownership_coverage.test.ts` pins 19 the same way, so after a rebase both want 21. `bot/config.ts` conflicts too. Rebasing onto `release/v0.30.0` and re-pinning is the fix.

**`item` reaches realm chat unfiltered, and the client resolves chat-link tokens inside it.** `server/onchain_feed.ts:52` caps `item` at 80 characters with no character filter, and `renderRealmLine` interpolates it at `:90` and `:94`. `src/ui/hud.ts:10580` routes any chat line containing `[[i:` through `parseChatSegments`, so an item name carrying `[[i:sword_iron]]` renders a real clickable item link inside the system announcement. `src/ui/hud/quest/quest_link.ts:1-5` states the assumption that makes the token contract safe ("a forged label can't misrepresent the target"), which holds only while the emitter is trusted. I ran this against the head: `validateOnchainEvent` accepts the payload and the rendered line parses into an `item` segment. Bidi overrides (U+202E) and zero-width characters survive the same way. Burns carry no `item`, so nothing is exposed today, but the PR body says the sale and Claudium rails light up with no further code change here, which is exactly why the filter should land now. A charset allowlist in `validateOnchainEvent` covers chat, Discord, and the tweet in one place.

For the record on what is not reachable: no HTML injection (`appendLog` builds the line from text nodes), no player impersonation (`broadcastSystem` emits a `log` with no `pid`), and no chat bubble.

## Should-fix

**The realm-chat floor does not hold when `usd` is null.** `server/internal.ts:120` announces when `evt.usd === null || evt.usd >= onchainRealmMinUsd()`, and `server/onchain_feed.ts:50` nulls `usd` whenever the field is missing, non-finite, or negative. A price-oracle outage therefore announces everything to every player regardless of `WOC_ONCHAIN_REALM_MIN_USD`, which is the one thing the knob exists to prevent. Whichever way you want it to behave, it needs a test: `tests/server/onchain_ingress.test.ts` covers only above-floor and below-floor with `usd` present, so deleting that clause leaves the suite green.

**A drained event that fails to post is lost, and the resend is silently swallowed.** `bot/main.ts:511-513` marks the signature before either post is attempted, and `:517` / `:520` only log failures, while the server-side drain has already removed the item. The worker's retry does not recover it: `enqueueOnchain` still holds the signature for 10 minutes, so `ingestOnchainEvent` gets `fresh === false`, skips both the queue and the announce, and `server/internal.ts:122` still answers 200 `{received: true}`. Worker and bot both read success on an event that is gone.

**`actor` lands in a Discord embed field value unescaped.** `bot/logic.ts:712`. Discord renders markdown in field values, so a masked link in `actor` posts as a live link from the project's bot. Same root cause as the `item` finding above.

**The legacy arm has no test, and the deviation ledger was not extended.** `tests/server/onchain_ingress.test.ts` drives the RouteDef arm only; the branches at `server/internal.ts:70-93` have no direct coverage, and the two arms differ in wiring (legacy announces through the `game` parameter, the RouteDef through `useInternalRuntime()`). `tests/server/internal.test.ts:783-807` is the established template for exactly this case. Both routes also belong in two `tests/server/http/known_deviations.ts` entries they are missing: `internalBodyValidationRemap` (the legacy arm has no outer catch) and `oauthInternalOffTable405` (single-method paths whose wrong-method 404 survives only by dispatcher delegation).

**Six new env vars are undocumented.** `X_ONCHAIN_ENABLED`, `X_API_KEY`, `X_API_SECRET`, `X_ACCESS_TOKEN`, `X_ACCESS_SECRET` (`bot/config.ts:50-54`) and `DISCORD_ONCHAIN_CHANNEL_ID` (`bot/config.ts:93`). `.env.example:140-151` lists every other `DISCORD_*_CHANNEL_ID`, and `bot/CLAUDE.md:92-102` names each one with its fallback chain, so an operator reading the docs cannot turn X posting on. `bot/CLAUDE.md:25-26` also still describes `server_client.ts` as the `/internal/discord/*` client and points at that prefix for the live set, which now misses `/internal/onchain/feed`.

**The tweet copy makes a supply claim.** `bot/logic.ts:734` posts "Supply only goes down." unattended from the project's X account. That is a judgment call rather than a code issue, but worth making a deliberate one before this ships enabled.

## Nits

- `tests/bot_x_api.test.ts:18-41` reconstructs the base string and HMAC inline instead of pinning a published vector, so a spec bug present in both copies is invisible. The commit message calls it "an independent vector"; it is a mirrored reimplementation. `bot/x_api.ts:44`'s `.sort()` is also a no-op today, since the object literal is already alphabetical.
- Counts left stale by this change: `server/internal.ts:319` ("All 12 handleInternalApi endpoints", now 14), `tests/server/internal.test.ts:99` ("The 12 routes"), `server/main.ts:2618` (`InternalRuntime is Pick<GameServer, 'startRestartCountdown'>`, now also `announceOnchain`).
- `server/onchain_feed.ts:39-40` accepts any 32 to 128 character `sig`. `bot/logic.ts:718` wraps it in `encodeURIComponent` for the embed, but `:729` concatenates it raw into the tweet URL; a base58 check would make both paths agree.
- `server/onchain_activity.ts` prunes `recentKeys` only of already-expired entries, so more than 1024 distinct signatures inside the TTL keeps it growing and re-scanning on every insert.
- `bot/main.ts:513` clears the whole dedupe set at 2000 entries, including the signature added on the line above.
- `tests/server/http/surface_inventory.ts:58-67` documents every other auth scope but was not extended for `secret-onchain`.

## Notes

- CI has not run: both workflow runs on `bf7cba6` are `action_required` (fork PR pending approval), so the green claim is local-only for now. I ran `tsc` on the head and saw nothing attributable to this diff, and the 10 suites this PR touches pass here (333 tests).
- `scripts/malware_scan.mjs` reports `high=0` and flags none of the new files. Worth flagging anyway that `bot/x_api.ts` is the repo's first outbound credentialed egress, so it is the one file I would want the release gate to look at directly.

PR #2214, opened 2026-07-20
fix(talents): ward the caster, not the enemy, on Hellglass Ward

seanghods:
## Summary

The warlock choice-row talent Hellglass Ward ("Every 3rd damaging Fire or Shadow spell raises a demonic ward absorbing 90 damage for 10 sec") applied its ward to the enemy the triggering spell hit, not to the warlock. The dead-target fallback inverted it: if the 3rd cast killed the target the caster got the ward, and if the target survived, the target did.

Root cause: Hellglass rides the shared talent-proc engine (`src/sim/combat/talent_procs.ts`). A `castNth` trigger fires with the triggering cast's target as the response subject, and the `absorb` response applies its aura to that subject. That subject plumbing is correct for every other absorb proc in the game (their triggers are heals or hits taken, so the subject is a friendly or the wearer); Hellglass is the only absorb proc whose triggering abilities are hostile.

The fix makes the recipient explicit rather than inferred:

- Schema (`src/sim/content/talents.ts`): an optional `target?: 'self'` field on the `absorb` and `heal` arms of `ProcResponse`. The heal arm has the identical latent hazard (no current talent hits it); this closes the class of bug, not just the instance.
- Engine (`src/sim/combat/talent_procs.ts`): `fireOne` resolves the recipient as `response.target === 'self' ? player : subject` for both arms. Pct-of-max-HP amounts and the `wardBloom` spellfx target follow the recipient. No hostility inference, by design (offline/PvP/charm edge cases, and no implicit behavior change for future content).
- Data (`src/sim/content/choice_rows_classic.ts`): Hellglass Ward's absorb response gains `target: 'self'`. No other talent touched.

Recipient audit of all seven absorb procs (the blast radius):

| Proc | Trigger | Subject | After this PR |
|---|---|---|---|
| `pal_greater_blessing` (Afterglow Aegis) | spellCrit on heals | heal target (friendly) | unchanged |
| `pri_lingering_ward` (Warding Refrain) | castNth on Whispered Prayer | heal target (friendly) | unchanged |
| `pri_inner_fire` (Wounded Halo) | bigHitTaken | the wearer | unchanged |
| `pri_blessed_recovery` (Halo Aftershock) | spellCrit on heals | heal target (friendly) | unchanged |
| `wlk_grimoire_of_carnage` (Hellglass Ward) | castNth on damaging Fire/Shadow | the enemy (the bug) | fixed: the warlock |
| `dru_grove_covenant` (Grove Covenant) | castNth on Wildmend | heal target (friendly) | unchanged |
| `dru_survival_of_the_fittest` (Ironhide Reflex) | bigHitTaken | the wearer | unchanged |

No rng draws added, removed, or reordered (`tests/parity` green); no new `SimEvent` type or wire change; no player-visible text change (no i18n work needed, `tests/localization_fixes.test.ts` green).

## Type of change

- [x] Bug fix

## How was this tested?

- Commands:
  - `npm run gate`: every step up to the full vitest suite is green (i18n gen + freshness, malware scan, changed-files biome, SFX conformance). The full suite passes 17171 tests with the only failures being the two known `tests/deploy_watchdog.test.ts` "abandons a hung docker" cases, a sandbox-environment flake: I verified they fail identically on the pristine release branch with this change stashed (CI runs them fine). Because the gate aborts at that step, I ran the later steps directly: `npx tsc --noEmit` (clean) and `npm run build` (all five entries, exit 0).
  - Test-first: the three decisive new cases were red before the fix and green after: the self-targeted ward lands on the caster with the enemy carrying no ward (live target), the self-targeted heal heals the caster, and the real Hellglass data drives a warlock `Sim` (three Gloom Bolt casts at a live mob) with the ward on the warlock and none on the mob.
  - Pinned regressions: the dead-target fallback still lands the ward on the caster, and absorb/heal responses without a `target` field still land on the subject (the heal-target shape used by the priest, paladin, and druid procs).
  - Guards: `npx vitest run tests/parity tests/architecture.test.ts tests/localization_fixes.test.ts tests/choice_rows_engine.test.ts tests/choice_rows_redesign.test.ts tests/choice_rows_unlock_guard.test.ts` and `npx tsc --noEmit`, all green.
- Manual steps: N/A (sim-only, covered by the automated cases above).

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green except the two pre-existing
      `deploy_watchdog` sandbox flakes detailed above (verified unrelated by stashing);
      `tsc` and all builds run directly and green. I added decisive tests for the
      changed behavior (red before the fix) and recorded the checks above.

### Cross-platform

- ~Tested on desktop and mobile.~ N/A: sim-only, no UI change.

### Localization (i18n)

### Hygiene

### DIFF
--- a/src/sim/combat/talent_procs.ts
+++ b/src/sim/combat/talent_procs.ts
@@ -102,20 +102,23 @@ function fireOne(
       }
       player.resource = Math.min(player.maxResource, player.resource + response.amount);
       break;
-    case 'heal':
+    case 'heal': {
+      const recipient = response.target === 'self' ? player : subject;
       ctx.applyHeal(
         player,
-        subject,
+        recipient,
         response.amountPctSourceMaxHp !== undefined
           ? Math.round(player.maxHp * response.amountPctSourceMaxHp)
           : response.amountPctMaxHp !== undefined
-            ? Math.round(subject.maxHp * response.amountPctMaxHp)
+            ? Math.round(recipient.maxHp * response.amountPctMaxHp)
             : (response.amount ?? 0),
         def.name,
       );
       break;
-    case 'absorb':
-      ctx.applyAura(subject, {
+    }
+    case 'absorb': {
+      const recipient = response.target === 'self' ? player : subject;
+      ctx.applyAura(recipient, {
         id: def.id,
         name: def.name,
         kind: 'absorb',
@@ -123,7 +126,7 @@ function fireOne(
         duration: response.duration,
         value:
           response.amountPctMaxHp !== undefined
-            ? Math.round(subject.maxHp * response.amountPctMaxHp)
+            ? Math.round(recipient.maxHp * response.amountPctMaxHp)
             : (response.amount ?? 0),
         sourceId: player.id,
         school: def.school ?? 'holy',
@@ -131,11 +134,12 @@ function fireOne(
       ctx.emit({
         type: 'spellfx',
         sourceId: player.id,
-        targetId: subject.id,
+        targetId: recipient.id,
         school: def.school ?? 'holy',
         fx: 'wardBloom',
       });
       break;
+    }
     case 'aura':
       ctx.applyAura(player, {
         id: def.id,
--- a/src/sim/content/choice_rows_classic.ts
+++ b/src/sim/content/choice_rows_classic.ts
@@ -1913,7 +1913,17 @@ export const WARLOCK_CHOICE_ROWS: ClassChoiceRows = {
                 n: 3,
                 abilities: warlockDamagingFireOrShadowSpellAbilityIds,
               },
-              responses: [{ kind: 'absorb', amount: 90, duration: 10, name: 'Hellglass Ward' }],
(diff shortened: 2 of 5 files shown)


jgyy (commented):
Correct fix, and test-first the right way: the new `talent_procs.test.ts` cases assert `enemy.auras` stays empty and the ward lands on the player, which is the exact assertion that would have failed against the old code (ward on `subject`, i.e. the enemy). Nice touch making `heal`/`absorb` both accept the optional `target: 'self'` on `ProcResponse` rather than special-casing Hellglass Ward directly in the engine, since it closes the same latent hazard on the heal arm for free without changing behavior for anything that doesn't opt in.

Checked that only Hellglass Ward's record in `choice_rows_classic.ts` was touched, no other talent, and that amount/duration/proc cadence are all unchanged, just the recipient. No rng involved so no parity concerns.

No blocking or should-fix issues. Good scoping on a change that could easily have crept wider.

PR #2219, opened 2026-07-20
feat(social): show AFK state to other players (nameplate tag, presence dot)

seanghods:
## Summary

The `/afk` command already existed on `release/v0.29.0` (alongside `/dnd` and whisper auto-reply), but it only drove the whisper auto-responder: **other players had no way to see that you were AFK.** This PR adds the visible half of the classic behavior.

- **Moving clears AFK** (classic behavior). Previously only sending another chat line cleared it. Do Not Disturb is deliberate and survives movement.
- **Other players see an `<AFK>` tag** over your character's nameplate, live, the moment you go AFK (and it clears the moment you move).
- **Social lists show an `Away` presence status** with its own status dot.

Design notes:
- A single source of truth, `src/sim/social/away.ts` (`setAwayState` / `clearAfkOnMove`), keeps the new transient `Entity.afk` display flag in lockstep with the authoritative `PlayerMeta.away` (afk mode only, never `dnd`). Every away-state mutation in the chat router now routes through it.
- The flag rides the entity as a new `ak` dynamic wire field (mirrored in `applyWire`), so it reaches other clients inside the interest bubble. It is **not** baked into the wire name: the `<AFK>` tag is composed client-side in `nameplate_painter.ts` from a `t()` key.
- `presenceOf` derives a new `'afk'` `PresenceStatus` with priority **dead > dungeon > combat > afk > online**.
- No new wire command, no new `IWorld` member, no `Rng` draw. Determinism and the three-host contract are preserved (the flag is plain deterministic state; the sync module is DOM/Three/i18n-free).

## Related issues

N/A

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run` (full suite, excluding the two known sandbox-flaky suites `tests/deploy_watchdog.test.ts` and `tests/sfx_studio.test.ts`, which fail identically on a pristine branch in this sandbox and pass in CI): green.
  - `npx tsc --noEmit`: clean.
  - `npm run build`: all five entries build.
  - Domain reviewers run clean: `architecture-reviewer` (determinism + SimContext seam), `cross-platform-sync` (wire/parity), plus the `qa-checklist` gate.
- New / updated tests:
  - `tests/away_state.test.ts` (new): direct unit tests for the `setAwayState` / `clearAfkOnMove` sync helper (afk lights the mirror, dnd does not, movement clears afk and emits the notice, dnd survives movement, no-op when not away).
  - `tests/afk_wire.test.ts` (new, load-bearing): drives `/afk` over the REAL server wire via `GameServer` with the db mocked, and asserts a co-located second client sees the `ak` dynamic field flip on (afk) and off (movement). The offline `Sim` bypasses the wire gate, so this is the decisive test for the "others see it" contract.
  - `tests/chat.test.ts`: `/afk` sets `Entity.afk`, `/dnd` does not, movement clears it (dnd survives).
  - `tests/presence_zone.test.ts`: `presenceOf` reports `'afk'`, ignores `dnd`, and yields to dead/dungeon/combat.
  - `tests/social_status_dots.test.ts`: `'afk'` added to the guarded JS<->CSS status-dot contract.
- Manual steps: captured the nameplate before/after offline via `scripts/afk_nameplate_shot.mjs` (stages a second player and sets the flag directly; the offline HUD renders other players' nameplates).

## Screenshots / recordings

Nameplate tag, desktop (offline capture):

| Before (not AFK) | After (AFK) |
|---|---|
| before | after |

(Also committed under `docs/screenshots/afk/` in this branch.)

## Deferred follow-ups (intentionally out of scope)

- **Auto-AFK after idle** (flag a player AFK automatically after an inactivity window). Groundwork is present: `PlayerMeta.lastActiveTick` already stamps deliberate input at the same choke point where movement now clears AFK.
- **Party/raid frame AFK indicator.** v1 leaves party frames untouched; the social roster dot covers the roster case. A party-frame tag can reuse the same `Entity.afk` flag.
- **Do Not Disturb nameplate tag / presence dot.** This PR scopes the visible treatment to AFK only; `dnd` stays a private state (auto-reply + whisper withholding, as today). The `ak` flag deliberately carries afk only.


jgyy (commented):
Good handling of a change that touches sim, wire, and render all at once. `away.ts` is the single source of truth keeping `PlayerMeta.away` and the wire-facing `Entity.afk` bit in lockstep, and it's correctly host-agnostic with no DOM/i18n imports. `nameplate_painter.ts` only reads `e.afk` to compose the tag, it doesn't decide AFK itself, so server authority holds. Confirmed no `Date.now`/`Math.random`/`performance.now` anywhere in the sim diff; AFK is a manual `/afk` toggle cleared by tick-driven movement input, not a wall-clock timer, so there's no timer-epoch persistence hazard to worry about here.

The `afk_wire.test.ts` case is the one that matters most and it's done right: it drives two co-located sessions through the real `GameServer` and asserts the *other* client's snapshot shows the `ak` bit, rather than only checking the offline `Sim`, which would have missed a wire-encoding bug. `presence_zone.test.ts` pins the exact status priority (dead > dungeon > combat > afk > online).

nit: the AFK screenshots are a single-client capture with a synthetically staged second player rather than a live two-client capture, worth knowing given the feature is entirely about what *other* players see, though the wire test already covers that gap.

PR #2220, opened 2026-07-20
feat(ui): tab the interface options panel

seanghods:
## Summary

The Esc options window's **Interface** panel had grown to about 40 rows in a single scroll. This splits it into four tabs, **General / Frames / Chat / Combat**, built on the repo's existing shared WAI-ARIA tab family (`tab_strip_view.ts` + `tab_strip_painter.ts`, the same one `social_window` and `talents_window` use), so it keeps the roving-keyboard navigation and `role=tablist/tab/tabpanel` semantics for free.

The category model lives in the pure `options_view.ts` core: each interface control gains a `category`, plus a pure `interfaceControlsForTab()` filter and an `INTERFACE_TAB_ORDER` the painter renders the strip from. The painter (`renderInterface()`) paints one tab strip above the panel body and repaints per tab; the bespoke rows painted outside the declarative list (language + theme, chat-timestamp toggle + clock pair, chat-window reset, unit-frames reset, deed broadcasts) move into their mapped tabs. The active tab is remembered for the session (a controller field, not a persisted setting).

Along the way, a completeness test partitions the interface list (every control has exactly one category, the union across the four tabs equals the full list with no duplicates). That assertion caught and removed a long-standing **double `showAttackButton`** entry, and the now-redundant "Party and Raid Frames" section note is dropped since Frames is its own tab.

Category mapping is the approved product taxonomy. The one judgment call flagged here (and seconded in review) is settled: `invertLookY` moved to **General** with the comfort toggles (it is camera comfort, not a combat setting), while `fctScale` stays under **Combat** (floating combat text is combat feedback, grouped with the attack/action-bar toggles).

## Related issues

N/A.

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling
- [ ] Breaking change

## How was this tested?

- Commands:
  - `npx vitest run tests/options_view.test.ts tests/options_window.test.ts` (pure taxonomy + painter source guards, 60+ pass; the no-duplicate/partition assertion is red-before / green-after the `showAttackButton` dedupe).
  - `npx vitest run tests/architecture.test.ts tests/localization_fixes.test.ts tests/i18n_completeness.test.ts tests/tab_strip_view.test.ts tests/tab_strip_painter.test.ts` (green).
  - `npm run i18n:gen`, `npx tsc --noEmit` (clean), `npm run ci:changed` (exit 0), `npm run build` (all five entries succeed).
  - Full `npx vitest run` (excluding the sandbox ffmpeg-timeout flakes `deploy_watchdog`/`sfx_studio`, which pass in isolation): green.
- Manual steps: opened Esc > Interface offline on desktop and on a landscape-phone viewport; switched all four tabs (mouse + arrow keys), confirmed every prior setting is reachable and functional, the remembered-tab persists across close/reopen, and the mobile strip holds the 40px touch floor.

## Screenshots / recordings

Desktop:

| Before | After |
|---|---|

Mobile (landscape):

| Before | After |
|---|---|

---

### Quality

- [x] Decisive tests added/updated for the changed behavior; manual checks recorded above. Full `npm run gate` not run end to end locally (it aborts on the known sandbox ffmpeg/watchdog flakes); each gate step was run individually and is green (tsc, builds, changed-file biome, the guard suites, the full test suite minus those two flakes).

### Cross-platform

- [x] Tested on desktop and on a phone-sized landscape viewport. The mobile strip keeps every tab at the 40px touch floor and 16px font (scrolls horizontally rather than shrinking below the target).
- [x] Accessible: reuses the shared tab family's `role=tablist/tab`, `aria-selected`, roving tabindex and keyboard selection; `:focus-visible` outline from the `--color-border-focus` token; no new motion.

### Localization (i18n)

- [x] Four new `hudChrome.interfaceTabs.*` labels added English-only to the `en`-only `i18n.catalog/hud_chrome.ts`, rendered via `t()`. The wordy labels (General/Frames/Combat) carry their five non-Latin M16 fills; no other overlay edits.
- [x] No new numbers/dates/percents; existing controls unchanged.
- [x] N/A: no `src/sim/` or `server/` player text added.

### Hygiene

- [x] No secrets, no `.env`, `ALLOW_DEV_COMMANDS` untouched.
- [x] Generated i18n artifacts were regenerated via `npm run i18n:gen`, not hand-edited.


jgyy (commented):
Solid refactor for a change this size. `options_view.ts` stays fully DOM-free and was already in the pure-core allowlist, so no gap there. `options_window.ts` reuses the existing `tab_strip_view`/`tab_strip_painter` family (the same one social and talents already use) instead of hand-rolling a new tab widget, which gets real ARIA tablist/tab semantics and arrow-key roving focus for free.

I diffed the full control list before and after: every option from the old flat list reappears exactly once under its new tab. The one exception is the old duplicate `showAttackButton` entry, which is now correctly deduped to a single control under Combat, caught by the new partition test (`Set(union) == Set(all)` with a no-duplicate-keys check) rather than introduced as a silent side effect.

`interfaceTab` is correctly kept as session-only controller state, not folded into the persisted settings store, and there's a source-guard test pinning that it's never touched in `toggle()`/`close()`/`goBack()`.

nit: `.opt-tab.on` in `components.css` uses a hardcoded hex gradient rather than a `--color-*` token, though this may just match the social-tab styling it's cloned from.

The PR body itself flags `fctScale`/`invertLookY` landing under Combat rather than a "comfort" grouping as a judgment call worth a second look, seconding that here. Otherwise no blocking or should-fix issues, and screenshots cover desktop/mobile before and after.

Rubsey:
Landed the review follow-ups plus the base merge, in three commits:

1. **Merged `release/v0.29.0`** (`ad2bf6e7e`): the `scripts/pr_shot_targets.mjs` conflict was both branches appending a screenshot target at the same array slot (kept both, `interface-options-tabs` + `guild-roster`); the generated `pending.ts` slice was regenerated via `npm run i18n:gen`, not hand-merged.
2. **Fixed the Chromium browser-regressions failure** (`3beca60db`): the a11y action-bar focus test opened the Interface panel and queried the flat list, but `showThirdActionBar` now lives under the Combat tab, so the test activates that tab first. `tests/browser/a11y.browser.test.ts` is 23/23 locally.
3. **Settled the category judgment call** (`0e4b4c7d3`): `invertLookY` moved to General with the comfort toggles (it is camera comfort, not a combat setting); `fctScale` stays under Combat (floating combat text is combat feedback). Taxonomy pins updated.

On the `.opt-tab.on` gradient nit: confirmed it is a byte-for-byte clone of `.soc-tab.on` (same `#1a1410` / `#463a1c` / `#6f5a2a` / `linear-gradient(#2a2436, #161220)` set the tab family already uses in five places). Keeping the literals is deliberate: `src/ui/theme.ts` rewrites `--border` per theme preset, so tokenizing only the new strip would make the options tabs drift from the social tabs under non-default themes, and a family-wide tab-strip tokenization is exactly the kind of coordinated styles change `src/styles/CLAUDE.md` says not to re-land piecemeal. Happy to file that as a follow-up chore instead.

Local gate: full suite 17,491 passed (the only reds are the two known machine-local `deploy_watchdog` hung-docker timing tests, byte-identical to base and green in CI), `tsc` clean, all five builds succeed, i18n freshness clean.

PR #2221, opened 2026-07-20
feat(ui): group the guild roster by online status

seanghods:
## Summary

The social window's Guild tab listed every member in one flat list, mixing online and
offline players together. This groups the roster: online members first under an
`Online (N)` header, offline members under `Offline (M)`, each header carrying its count.
It also adds a persisted **Hide offline** toggle so players can collapse the offline
group; the choice survives across sessions.

Grouping, counts, and the filter are pure functions (`guildRosterItems` in
`src/ui/social_view.ts`); the toggle's persistence follows the existing lightweight
localStorage pattern (`src/ui/guild_hide_offline.ts`, mirroring `party_collapse.ts`). The
`social_window.ts` painter stays a thin consumer that renders through the pure core. No
wire, `IWorld`, or server changes: the member `online` axis was already on the roster
model, so this is a pure-view change.

Design notes:
- Empty groups emit no header (no stray `Offline (0)`), so an all-online or all-offline
  roster renders exactly one group.
- The `Online (N)` count reflects real online membership regardless of the filter; hiding
  offline only suppresses that group's rendering.
- The toggle is a pure USER choice: it is never gated on graphics tier, `reduce-motion`,
  or the FPS governor (guild presence is actionable information, not cosmetics).
- The Friends tab is intentionally left untouched.

## Related issues

N/A

## Type of change

## How was this tested?

- Commands:
  - `npm run gate` (green: i18n freshness, changed-files biome, full tests, `tsc`, all builds)
  - `npx vitest run tests/social_view.test.ts tests/guild_hide_offline.test.ts tests/social_window.test.ts`
  - Red-before verified: the new `guildRosterItems` grouping tests fail against the
    pre-change `social_view.ts`, then pass with the implementation.
- Tests added:
  - `tests/social_view.test.ts`: mixed/all-online/all-offline/hidden grouping, per-group
    counts, within-group order preserved, empty-header suppression, count-vs-filter
    independence, purity.
  - `tests/guild_hide_offline.test.ts`: persistence round-trip, default-show, exact-`'1'`
    hide, storage-unavailable safety, and the on-disk key pinned to a literal.
  - `tests/social_window.test.ts`: grouped render composition, localized count headers,
    the delegated toggle handler (flip + persist + refresh), and keyboard-focus restore
    after the innerHTML swap.
- Manual steps: staged an offline guild fixture through the sim debug hook and drove the
  Guild tab on desktop and a landscape phone viewport (screenshots below).

## Screenshots / recordings

Guild tab, desktop, before and after:

| Before (flat list) | After (grouped) |
| --- | --- |

Hide-offline engaged (offline group collapsed), desktop:

after desktop hide offline

Mobile (landscape), before and after:

| Before | After |
| --- | --- |

---

### Quality

- [x] **The full gate passes.** `npm run gate` is green. I added decisive tests for the
      grouping, the persistence helper, and the painter wiring.

### Cross-platform

- [x] **Tested on desktop and mobile.** Verified on a desktop and a landscape phone
      viewport. The toggle keeps a 40x40 touch target on mobile (`hud.mobile.css`).
- [x] **Accessible.** The toggle is a real `aria-pressed` button with a steady
      `:focus-visible` ring; keyboard focus is restored to it after each toggle so
      repeated keyboard presses stay on the control. No motion added.

### Localization (i18n)

- [x] **New player-visible strings follow the contributor policy.** Four keys added to
      the English `hudChrome.social.*` catalog. The strings are wordy, so the M16 rule
      applies: the five non-Latin fills (`zh_CN`/`zh_TW`/`ja_JP`/`ko_KR`/`ru_RU`) are
      included in the same change.
- [x] Counts go through `formatNumber`.
- [ ] ~Player-facing text emitted from `src/sim/`/`server/`~ N/A (client-only change).

### Hygiene

      enabled anywhere.
- [x] Generated i18n artifacts were regenerated via `npm run i18n:gen`, not hand-edited.


jgyy (commented):
Good use of the existing seams: `guild_hide_offline.ts` mirrors `party_collapse.ts` for the persisted toggle, `social_view.ts` gets a pure `guildRosterItems()` projection, and `social_window.ts` stays a thin painter with the row markup mechanically extracted rather than duplicated. Both new modules are registered where the architecture guard expects them.

Test coverage is genuinely decisive: `social_view.test.ts` asserts the exact ordered array (online header+members, then offline), the all-online/all-offline edge cases (no stray empty header), and that hide-offline removes both header and rows. `guild_hide_offline.test.ts` pins the literal storage key string, not just the imported constant, which catches a silent rename.

Mobile touch target for the new toggle correctly rides the shared `body.mobile-touch` selector group. Screenshots cover desktop, mobile, and the hide-offline state. Only conflict against the release base was the generated `pending.ts` slice, which is expected regen churn, not a real conflict.

PR #2222, opened 2026-07-20
fix(sim): stop a DoT tick from ticking twice when it breaks its own carrier's aura

Wmedrado:
## Summary

`updateAuras` walks `e.auras` backward by live index:

```ts
for (let i = e.auras.length - 1; i >= 0; i--) {
  const a = e.auras[i];
```

The DoT branch inside that loop calls `ctx.dealDamage`, and `dealDamage` runs its own backward sweeps over **the same array**: it splices out a `breaksOnDamage` control aura (the Fear family) when damage lands, and it splices out an absorb shield that has just been depleted.

So a removal can happen underneath a live cursor. When the removed aura sits at a **lower** index than the DoT being processed, everything above it shifts down one, which puts the just-processed DoT back under the cursor. It is then processed a second time in the same sim tick: `remaining` and `tickTimer` are decremented twice, and the target takes a second tick of damage.

In play that means a feared target taking a DoT eats double damage on the tick that breaks the fear. It needs the two to coincide, so it is intermittent rather than constant, which is probably why it has gone unnoticed.

The fix walks an up-front snapshot, skips any entry a side effect already removed, and splices expiries at the aura's live index rather than the snapshot index.

**On determinism**, since this file's header is explicit that the extraction was a verbatim move to keep the parity trace byte-identical: the snapshot is taken in the exact order the live walk had, so every aura that survives its own turn is processed exactly as before, and the change only ever *removes* a spurious `dealDamage` call, never adds one. The parity gate is green **with no golden regeneration**, which is the proof: both the full-state trace and the rng draw-order log are unchanged. I also added a note to the file header, since the PRIME DIRECTIVE comment there would otherwise read as contradicted by this diff.

## Related issues

None open that I could find; I ran into this reading the aura tick.

## Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/combat_auras.test.ts tests/parity tests/architecture.test.ts` (220 pass, 1 skipped)
  - `npx tsc --noEmit` (clean), `npx @biomejs/biome check` on both changed files (no errors, no format diff)
- Manual steps: N/A, the crossing case is covered directly by the new unit test against a real `Sim.ctx`.

The new test pins the exact crossing case: a `breaksOnDamage` fear pushed at a lower index than a 50-damage DoT, so the backward walk reaches the DoT first and the fear is removed beneath it. It asserts one tick of damage (50), that the fear is still broken by that damage, that the DoT aged by exactly one `DT`, and that it is the only aura left. Against the unmodified walk it reads **100 damage instead of 50**, so it reproduces the defect rather than just documenting the new shape.

## Screenshots / recordings

N/A, no visual surface.

---

### Quality

- [x] **The full gate passes.** Typecheck, Biome, the parity gate (no golden regeneration needed) and the aura suite are green. The added test is verified to fail against the pre-fix walk.

### DIFF
--- a/src/sim/combat/auras.ts
+++ b/src/sim/combat/auras.ts
@@ -14,6 +14,14 @@
 // `e.auras.splice`, `c.remaining -= 2`, `a.tickTimer += ...`) are preserved exactly so
 // the parity gate's full-state trace AND rng draw-order log stay byte-identical.
 //
+// The one deliberate deviation from verbatim is updateAuras's snapshot-plus-liveness
+// walk (see the comment at the loop). It fixes a re-entrancy bug the verbatim move
+// carried over: a DoT tick's own dealDamage call splicing an aura out of this same
+// array mid-walk, which pulled the just-processed entry back under the cursor and
+// ticked it twice. It keeps the iteration ORDER, and therefore the rng draw order,
+// identical for every aura that survives its own turn: the parity gate stays green
+// with NO golden regeneration.
+//
 // CRITICAL: updateAuras carries TWO load-bearing `e.dead` guards, the top guard and
 // the post-DoT guard. A DoT tick calls ctx.dealDamage, which can kill the target
 // mid-walk; both guards stop further processing of a dead entity's auras. They MUST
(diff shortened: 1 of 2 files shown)


Rubsey:

Thanks for the contribution, @Wmedrado.

PR #2224, opened 2026-07-20
feat(ui): explain what "Deposit all materials" actually banks

### LINKED ISSUE #2132: Feat: Clarify what 'deposit all materials' in the bank does
### Before you start

In the bank interface, there is a 'deposit all materials' button. It does deposit all my looted bone fragments, but also my tools like the pickax, sickle and axe. This is confusing because I don't consider the tools to be `materials`
Image

### Proposed solution

Add a tooltip to the `deposit all materials` button clarifying what it does. I also prefer to keep my tools in my inventory as I don't consider them materials.

### Alternatives considered

Mark the tools as a 'material' so that it makes sense they are deposited in the bank.

### Area

UI / HUD

Wmedrado:
## Summary

The bank's deposit-all button reads "Deposit all materials", but "material" there is the bag-filter category, and that category is `junk | tool` (`src/ui/bag_filter.ts`). So the button also banks profession tools: pickaxe, sickle, axe. `src/ui/bank_view.ts` documents this deliberately ("matchesCategory 'material' = junk/tool"), but nothing on screen tells the player, and the only way to find out is to click it and go looking through your bags.

This adds a tooltip on the button that states both halves: what gets banked (crafting materials and profession tools that fit) and what deliberately stays behind (gear, consumables, quest items). Naming what *doesn't* move seemed as important as what does, since that is the part a player is actually anxious about.

The `deposit.title = t(...)` idiom matches how `hud.ts`, `party_frame_row.ts` and `meters.ts` already attach tooltips.

### A note on the locale fills

I know the contributor policy is English-only, with the maintainer filling locales at release. This value trips the M16 exception (`tests/i18n_completeness.test.ts` flags any wordy new English string that reaches a non-Latin surface untranslated), so the five non-Latin fills ride along in the same change as that rule requires. I kept them anchored to each locale's existing bank vocabulary (the `depositAll` / `depositAllDone` family) and each locale's own terms for materials, professions and consumables rather than introducing new wording. Please retranslate freely if any of it reads awkwardly to a native speaker.

## Related issues

Closes #2132

## Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Tests
- [ ] Build, CI, or tooling

## How was this tested?

- Commands:
  - `npx vitest run tests/bank_window.test.ts` (67 pass)
  - `npx vitest run tests/i18n_completeness.test.ts tests/i18n_build_gapfill.test.ts tests/i18n_coverage_summary.test.ts tests/i18n_admin_catalog.test.ts tests/localization_fixes.test.ts` (74 pass, 3 skipped)
  - `npm run i18n:build` (the resolved bundles in the diff are its output, not hand-edited)
  - `npx tsc --noEmit` (clean), `npx @biomejs/biome check` on the changed source files (clean)
- Manual steps: opened the bank, hovered the deposit-all button, confirmed the tooltip appears and that the button's existing enable/disable behavior is unchanged.

I added an assertion to `tests/bank_window.test.ts` in that file's existing source-contract style, next to the one already pinning the button's label key.

## Screenshots / recordings

Happy to add a hover screenshot if useful. It is a native browser `title` tooltip on an existing button, so there is no layout or styling change to show, but say the word and I will capture one.

---

### Quality

- [x] **The full gate passes.** Typecheck, Biome, the bank suite and the i18n gate (including the M16 non-Latin check that this string trips) are all green.


EnriqueGF (changes requested):
Reviewed the full diff, the surrounding bank code (`src/ui/bank_window.ts`, `src/ui/bank_view.ts`, `src/ui/bag_filter.ts`) and the item catalog, and ran the checks locally.

**What I ran** (worktree on this branch, base `release/v0.29.0`):
- `npx tsc --noEmit`: clean.
- `npx vitest run tests/bank_window.test.ts tests/architecture.test.ts tests/i18n_completeness.test.ts tests/localization_fixes.test.ts tests/i18n_build_gapfill.test.ts tests/i18n_coverage_summary.test.ts`: all green (after `npm run i18n:gen`, which `pretest` does).
- `npm run i18n:build`: no working-tree diff, so the committed resolved bundles and `translation_keys.generated.ts` really are the generator's output and are fresh.

**What holds up well**
- The key lands in `src/ui/i18n.catalog/hud_chrome.ts` (English) and the generated artifacts are regenerated, not hand-edited.
- The M16 non-Latin fills are the documented exception for a new wordy English value, and they are anchored to the existing `depositAll` bank vocabulary in each locale. Latin locales stay `pending` for the release-time fill, which is correct.
- No sim, wire or server surface touched, so no determinism / IWorld parity / authority exposure. No dashes or emoji.

**Finding 1 (main, copy accuracy): the tooltip still does not describe what actually moves.**

`matchesCategory(item, 'material')` is `kind === 'junk' || kind === 'tool'`, and `kind: 'junk'` is not only crafting reagents. It is also plain vendor trash: `soggy_boot` ("Soggy Boot"), `tangled_weed`, `amber_hide`, `soft_down`, `stag_antler` (`src/sim/content/items.ts`, all `quality: 'poor'`), and the fishing tables at `items.ts` around line 1860 hand those out constantly. `planDepositAllMaterials` deposits them like anything else.

So a player who fishes for ten minutes, reads "Banks every crafting material and profession tool that fits", and clicks the button gets their grey trash banked too, into a bank with real capacity pressure. The follow-on sting is that `sellAllJunk` (`src/sim/items.ts`) only sweeps `quality: 'poor'` items that are in the BAGS, so the trash the button just moved is now out of reach of the vendor sweep until it is withdrawn by hand. That is exactly the class of surprise this PR set out to remove, so it seems worth naming rather than shipping a more precise-sounding but still incomplete promise.

Suggestion: describe the category rather than enumerate two of its three members, for example "Banks everything the Materials filter shows: crafting reagents, gathered goods, profession tools and grey trade junk. Gear, consumables and quest items stay in your bags." Whatever wording you pick needs the same shape in the five non-Latin fills.

**Finding 2 (minor, mobile reach): native `title` does not exist on touch.**

`deposit.title = ...` matches the idiom in `hud.ts` and `meters.ts`, so it is not wrong. But this exact file already has the game's own tooltip on its deps (`this.deps.attachTooltip(cell, () => ...)` at `src/ui/bank_window.ts:430`), which renders in the styled HUD tooltip and works with the long-press peek path the bank grid already uses. Given the whole point is that mobile and desktop players alike cannot discover this behavior, `attachTooltip(deposit, () => esc(t(...)))` would reach more of them and be consistent within the file. Non-blocking, your call.

**Nit:** the new test is a source-substring contract match, which matches the surrounding file's style, so no objection. Its decisiveness comes from `tsc` (the key has to exist in the `TranslationKey` union); worth knowing it would not catch a wrong-but-valid key.

Happy to re-review quickly on the copy change. Nice catch on the underlying issue, and thanks for regenerating the bundles properly instead of hand-patching them.

Rubsey (commented):
## OSSBrain initial review

### Should fix

- **Deposit-all explanation is hover-only** (`src/ui/bank\_window.ts:561`)
  - Issue: The new clarification is assigned only to the native button title. Native title text is not reliably exposed on touch devices and is not the repository's shared tooltip path for keyboard focus, so mobile and keyboard users can still activate the ambiguous Deposit all materials action without seeing that profession tools may also move.
  - Impact: The change does not fully solve the player-facing ambiguity it introduces the tooltip for, especially on the mobile HUD that this window supports.
  - Suggested fix: Attach the localized text through the existing Hud attachTooltip dependency, matching the bank item cells and corpse harvest button pattern, so hover, mobile long-press, and keyboard focus all expose the explanation.

### Checks

- `git status --short`: **passed**. Only the expected untracked .ossbrain-review/ package was present before writing this result.
- `git diff --check 7b2efb94b0b0b518028ec50b892d70ed23f57b81..1072676301b79844bf13d3811af935373ecebb06`: **passed**. No whitespace errors in the integrated diff.
- `rg -n 'depositAllTooltip' src/ui/i18n.catalog/translation\_keys.generated.ts src/ui/i18n.resolved.generated/en.ts src/ui/i18n.resolved.generated/en\_XA.ts src/ui/i18n.resolved.generated/ja\_JP.ts src/ui/i18n.resolved.generated/ko\_KR.ts src/ui/i18n.resolved.generated/ru\_RU.ts src/ui/i18n.resolved.generated/zh\_CN.ts src/ui/i18n.resolved.generated/zh\_TW.ts src/ui/i18n.resolved.generated/pending.ts`: **passed**. The generated key union, English, pseudo-locale, five non-Latin locale fills, and pending registry all include the new key as expected.
- `npx vitest run tests/bank\_window.test.ts`: **not run**. Vitest did not reach tests because Vite attempted to write its temporary bundled config under the read-only shared node\_modules/.vite-temp path.

PR #2227, opened 2026-07-20
feat(render): mirror the active weapon skin onto a matching-type offhand

seanghods:
## Summary

Weapon skins are account cosmetics with one active skin per weapon TYPE, and they render on the mainhand only. This makes the active mainhand skin also render on the **offhand** whenever the offhand holds a weapon whose type matches the skin's `weaponType`: a rogue dual-wielding two daggers with a dagger skin now shows both blades skinned, rarity VFX and all, instead of a skinned mainhand next to a plain offhand dagger.

The mirror is strict. It fires only for a matching-type offhand **weapon**. Shields (armor + shield flag), held offhands (orbs/tomes, kind `held_offhand`), and any offhand weapon of a *different* type than the skin render their normal models, pixel-untouched. Hand is deliberately not consulted: a Fury warrior can offhand a two-hander (`equipment_rules.canDualWieldTwoHand`), and the mainhand rule already skins a matching-type two-hander, so the mirror treats both hands the same (pinned by a dedicated test). Mainhand behavior is unchanged.

How it works:

- New pure rule `offhandMirrorsWeaponSkin(skinId, offhandItemId)` in the weapon-skin rules hub (`src/sim/content/weapon_skin_rules.ts`): true only when the skin exists and `weaponTypeForItem(offhandItemId)` equals the skin's `weaponType`. Shields and held offhands simply have no weapon type, so the equality check is the whole rule (heroic-prefixed offhand ids resolve through their base row).
- `offhandModelUrl(offhandItemId, weaponSkinId)` composes the decision in the manifest (skin model wins when the rule passes, else the item model).
- The render attach path honors it everywhere the offhand is built or swapped: `offhandAttachDef`, `attachAllProps` (initial build + sheathe via `setWeaponsStowed`), `setHeldOffhand` (live offhand swap), and `CharacterVisual` (`setWeaponSkin` now refreshes the offhand, and a mirrored offhand payload joins the mainhand's rarity-VFX + per-payload material-isolation set so both hands light up). A non-mirrored offhand keeps the lean, mainhand-undisturbed swap path.
- **Off-hand grip fix (`variantGripTransform`):** the per-weapon `WEAPON_GRIP_OVERRIDES` nudges are authored against the right hand, whose base orientation is a 180 degree turn about Y from the off-hand's. The offset was applied verbatim to the left hand, which small overrides (most daggers) hid but a large one exposed: a legendary sword skin (`solheim`, `pos ~ [-0.18, _, -0.27]`) rendered detached from the off hand. The off-hand mirror is the first thing to put big-override skins in the off hand, so this ships with the fix: negate the offset's X and Z on the off hand (Y is the shared along-bone lift). Scale-only / X-and-Z-free overrides and every right-hand weapon stay byte-identical; pinned by a new `variantGripTransform` off-hand case.

No schema, wire, persistence, or sim-validation change: `wsk` stays the single resolved mainhand skin, and the offhand decision derives locally from `wsk` plus the offhand item id `oh` (both already interest-mirrored on the wire). Parity allow-lists are untouched.

**Scope note (deliberately deferred):** this mirror is the free, automatic slice of offhand skins. Independent per-hand skin selection (a hand dimension on `WeaponSkinLoadout`, a second wire field, and picker slot UI) is a possible later feature and is intentionally **not** included here. This change is forward-compatible with it: the mirror simply becomes the fallback when no explicit offhand choice exists.

## Related issues

N/A.

## Type of change

## How was this tested?

- Commands:
  - `npx vitest run tests/weapon_skins.test.ts tests/held_weapon_models.test.ts`: new coverage: the full mirror truth table with every exclusion arm (matching weapon, different-type weapon, shield, held offhand, unknown/absent skin, empty offhand, heroic-prefixed base resolution), red-before proven; plus `offhandModelUrl` skin-vs-item resolution.
  - Existing pins stay green: `tests/weapon_skin_sim.test.ts`, `tests/snapshots.test.ts` (dual-wield fixtures), `tests/world_api_parity.test.ts`, `tests/architecture.test.ts`, `tests/weapon_skin_materials.test.ts`, `tests/back_grips.test.ts`.
  - `npm run gate`: full suite passes except the two known `tests/deploy_watchdog.test.ts` sandbox flakes (the "abandons a hung docker" timing tests, which fail identically on a pristine release branch in this sandbox; CI runs them fine). The gate aborts at that step, so `npx tsc --noEmit` (clean) and `npm run build` (all five entries, clean) were run directly and both pass.
- Manual steps: offline desktop capture (below): staged a dual-wield rogue and a warrior via the sim's real equip + apply-skin path, verified both hands skin only on a matching type, the shield and sheathed carry stay correct.

## Screenshots / recordings

Offline, desktop, HUD hidden. Dual-wield same-type + skin is the only before/after case; the sword, shield, and sheathe shots are after-only demonstrations.

Dual-wield dagger skin, **before** (mainhand skinned, offhand a plain dagger):

before

Dual-wield dagger skin, **after** (both blades skinned, frost VFX on both):

after

A Fury warrior dual-wielding two swords under a **legendary** (Solheim) sword skin: both blades render the skin with the legendary VFX, and both sit correctly in hand (the grip fix above):

two legendary swords

The same legendary sword skin next to a shield offhand: the shield is a plain shield, unaffected (only a matching-type weapon mirrors):

sword skin and untouched shield

Sheathed skinned pair, back-carry correct for both blades:

sheathed

A second staged player's skinned dual-wield could not be shown from an observer client in the offline harness (single-player), but the wire-derivation argument covers it: an observer sees the same `wsk` + `oh` the wearer does, and the mirror is computed identically on every client.

---

### Quality

- [x] **The full gate passes** (modulo the documented `deploy_watchdog` sandbox flake; tsc and all five builds run directly and pass). Added decisive tests for the new behavior with every exclusion arm and a red-before check.

### Cross-platform

- [ ] ~Tested on desktop and mobile.~ This is a 3D character-render change, not a HUD/layout change; it renders identically on both, driven by the same per-frame diff. Desktop offline captures above.
- [ ] ~Accessible.~ N/A: no UI, focus, or ARIA surface changes.

### Localization (i18n)

### Hygiene

- [x] No secrets or `.env`, and `ALLOW_DEV_COMMANDS` is not touched.


FernandoX7 (commented):
Clean, well-scoped change. The mirror is derived entirely from wire fields that already exist (`wsk` + `oh`), so there is no schema, wire, or parity cost, and putting the decision in a pure sim rule (`offhandMirrorsWeaponSkin`) keeps both hosts computing the same thing and makes the truth table trivially testable. The render integration is careful: `finishWeaponAttach` re-runs the material pass over the whole model so a re-attached non-mirrored offhand is not left bare, the two hand disposals stay on disjoint tags, and the per-payload VFX-isolation clone keeps a rogue's two same-skin hands from sharing derived materials. Tests are thorough (every exclusion arm, heroic-prefix resolution, red-before) and renderer.ts correctly needs no change because its existing `setOffhand`/`setWeaponSkin` sites already reconcile lights.

A couple of small notes, nothing blocking:

- nit (`weapon_grip.ts:186`): the X/Z negation also repositions existing non-skin offhand weapons that carry an X/Z grip override (dual-wield item daggers/swords), not only skinned ones. It is a correctness improvement and the direction is substantiated by the sword shot and the new pin, but the before/after captures only cover the skin feature; a plain unskinned dual-wield pair is not shown to confirm no visible shift. Given the magnitudes on real offhand items this is low risk, just calling out that the fix's reach is slightly wider than the skin mirror.
- nit (`weapon_skin_rules.ts:196`): the rule gates on weapon-type equality alone; the "one-hand" qualifier in the comment and PR body is enforced only upstream by the equip rules, not in the rule itself. Unreachable today since a two-hander cannot occupy the offhand slot, so this is defensive-only, but a two-hand item id would classify by type and mirror if that ever changed.

jgyy (commented):
Clean, well-scoped. `offhandMirrorsWeaponSkin` in `weapon_skin_rules.ts` gates correctly on `weaponTypeForItem`, and the tests cover both the same-type mirror case and the mismatched-type (sword/dagger) fallthrough to the item's own offhand model, which is the case I'd have worried about. The mirrored-offset math in `weapon_grip.ts` (negate X/Z, keep Y) matches the existing 180-degree hand-turn convention rather than inventing a new one.

nit: `reattachHeldWeapon()` rebuilds the mainhand payload too on an offhand-only mirror swap, more work than strictly needed, but that's called out as a deliberate tradeoff to keep rarity VFX in sync across both hands, so not worth changing.

No wire/schema/persistence touched (derivation is local off already-mirrored fields), so no parity risk. Screenshots and tests both look complete.

Rubsey:
Addressed both review notes in 211d6cd70; responses below.

**@FernandoX7, nit on `weapon_skin_rules.ts` (the "one-hand" qualifier):** the premise inverts, and in an interesting way. A two-hander CAN occupy the offhand slot today: `equipment_rules.canDualWieldTwoHand` admits a Fury warrior, and `canEquipItemInSlot` accepts `hand === 'twohand'` for that path. And the mainhand skin rule already applies a skin to a matching-type two-hander (greatswords classify as `'sword'` in `WEAPON_TYPE_BY_ITEM`), so a defensive one-hand guard here would produce a skinned mainhand next to a bare offhand greatsword for a Fury pair, which is exactly the asymmetry this PR removes. Type-equality-only is the correct rule, so instead of adding the guard I pinned the reachable case: a new test asserts the Fury two-hand mirror alongside the mainhand resolution it stays symmetric with, and the comment on the rule (plus the "one-hand" phrasing it had spread into assets.ts, manifest.ts, visual.ts, the tests, and the PR body) now states that hand is deliberately not consulted.

**@FernandoX7, nit on `weapon_grip.ts` (the X/Z negation reaching unskinned offhand weapons):** verified the reach is exactly zero. I enumerated every weapon item any class/spec can place in the offhand slot per `canEquipItemInSlot` (94 items) and intersected with `WEAPON_GRIP_OVERRIDES` via each item's `ITEM_WEAPON_VARIANTS` model: none of them maps to a model with a nonzero X or Z offset. Item dagger/sword models carry either no override, a scale-only one, or an X/Z-free one (`whittler_s_knife`, `guildmark_arming_sword`-style), all byte-identical under the change per the new `variantGripTransform` pin. Every override with a real X/Z component belongs to a skin model, and the four non-skin models with X/Z offsets (`notched_woodaxe`, `peeled_birch_wand`, `knotted_oak_stave`, `simple_farmhand_crossbow`) are not referenced by any item or NPC table at all. So the fix's wider reach repositions only mirrored skin models, which is the feature itself; a plain unskinned dual-wield pair renders byte-identically, no capture needed.

**@jgyy, nit on `reattachHeldWeapon()` rebuilding the mainhand on an offhand-only mirror swap:** agreed and leaving as is per your note; the full re-attach is the deliberate tradeoff that keeps rarity VFX and the material-isolation set in sync across both hands when the mirror state flips.

PR #2230, opened 2026-07-20
feat(server): persist action bar layouts on the character

### LINKED ISSUE #2228: Action bar layout does not follow the character across devices, and older saved builds restore stale bars
## Problem

The live action bar layout is stored only in browser localStorage (key `woc_hotbar_<class>_<character>` in `src/ui/hud/action_bar/action_bar_controller.ts`, plus sibling keys for the druid form bars, the stealth bars, and the attack slot). The server never stores it, and `server/` has no client settings sync surface at all. The desktop client has its own localStorage, so a player who sets up their bars on the web and then logs into the same character from the desktop client (or a second browser or machine) gets freshly seeded default bars instead of their layout. Today the only bar state that travels with the character is the per build snapshot inside a saved talent build.

A player report points at the build arm of the same problem:

> we still have the issue with the builds. saving a build doesn't save the actions bars, so switching back to the previous build changes the bars again... or maybe it's the original build I had which is bugged... now I tried to save a new build and the bars are correctly kept as they were

Verified against the code: saving a build does capture the current bars today (`SavedLoadout.bar` in `src/sim/content/talents.ts` has existed since the loadouts feature landed, and the talents window always passes the live bar at save time). The failure comes from older snapshots. The saved bar cap grew from 16 to 22 slots when the second action bar landed (#983), and two v0.27.0 fixes (#1907, #2003) closed bugs that could corrupt or scramble what a build captured. A build saved before those changes can hold a partial or stale snapshot, and `applyLoadoutBar` (`src/ui/hud/action_bar/hotbar.ts`) clears every ability slot the stored snapshot does not fill (only item shortcuts survive), so switching to an old build silently rewrites the bars and persists the result immediately. That matches the report exactly: the original build scrambles the bars on every switch back, while a freshly saved build round trips correctly.

## Steps to reproduce

Cross device:

1. On the web client, arrange abilities across both action bars on an online character.
2. Log into the same character from the desktop client (or another browser or machine).
3. The bars come up default seeded, not as arranged in step 1.

Stale build snapshot:

1. On a character with a build saved before v0.27.0 (or before the second action bar existed), arrange the bars how you want them.
2. Switch to that old build, then back.
3. The bars are rewritten from the build's partial snapshot and the arrangement from step 1 is lost (the rewrite is persisted immediately).

## Actual behavior

- The action bar layout lives per device and per browser profile; nothing follows the account or character across clients.
- Switching to a build whose stored snapshot is partial or stale wipes the ability slots the snapshot does not cover, with no warning and no way back. Re-saving the build over itself is the only repair.

## Expected behavior

- The action bar layout (both bars, form and stealth bars, the attack slot, item shortcuts) is saved with the character server side and restored on any device after login.
- Saving a build captures the full current layout (all 22 slots), and switching to a build restores exactly the snapshot that build saved.
- A build saved under an older client must not silently wipe slots it never recorded: slots beyond the stored snapshot's length keep their current contents.

## Scope

- Server side persistence for the per character action bar layout, restored at login. It can plausibly ride the character record the way talents and loadouts already do (the `tal` heavy wire field in `server/game.ts`), with debounced saves on change. Offline play keeps localStorage as the store.
- A merge rule for first sync when a device holds a local layout and the server already has one (proposal: server wins once a server copy exists; an existing local layout seeds the first server copy).
- The build switch path in `applyLoadoutBar` for snapshots shorter than the current slot count.

## Acceptance criteria

- Arrange bars on one device, log into the same online character on a second device: the bars match without any manual step.
- Save a build on one device, switch builds on another device: the bars restore to that build's snapshot.
- Switching to a build saved under the old 16 slot format leaves slots beyond its recorded length untouched.
- Offline browser play behaves as before.
- Automated tests cover the persistence round trip and the legacy snapshot switch path.

## Environment

- Platform: online realms, web client and desktop (Steam) client. Offline single device play is unaffected.

## Notes

- Related device local state the same player will hit next: keybinds (profiles are localStorage too, see #1792) and window positions. Worth a follow up decision, but this issue is scoped to the action bars.
- Player workaround today: with the affected build active, press Save so it overwrites itself with the current bars.
- Existing test coverage of the build bar path: `tests/loadout_action_bar.test.ts` (a full 22 slot bar survives a save, truncation to `SAVED_LOADOUT_BAR_SLOTS`), `tests/hotbar.test.ts` (`applyLoadoutBar` keeps item shortcuts and drops stale or foreign spec ids), `tests/talents.test.ts` and `tests/persistence_round_trip.test.ts` (the bar rides the JSONB round trip), `tests/snapshots.test.ts` (loadouts ride the `tal` wire field to `ClientWorld`).
- Coverage gap: no test exercises switching to a legacy record that entirely lacks the `bar` field. The load path defaults it to an empty array (`bar: [...(l.bar ?? [])]` in `src/sim/sim.ts`), which then clears every ability slot on switch; the persistence round trip test only uses an explicit `bar: []`. The legacy snapshot rule in the acceptance criteria should land with a named back compat test for the absent field.

seanghods:
## Summary

Action-bar layouts (the base bar, the druid Bear/Cat/Cat-stealth kits, the rogue
Stealth bar, the Vale Cup sport bar, and each form's Attack slot) lived only in
browser `localStorage`. A second browser, a new machine, or the desktop client
therefore seeded default bars. This adds **per-character server persistence of
the full layout, restored at login on any device.**

The layout is client *presentation* state, not deterministic gameplay state, so
it deliberately does **not** enter the sim-owned `CharacterState` (which would
make the offline `Sim` host-aware and change its serialization). Instead it:

- rides its own **additive `characters.hotbar_layout` JSONB column** (no change
  to `CharacterState`, sim serialization stays byte-identical, rollback-safe:
  old binaries ignore the column);
- reaches the owning client as a **self-scoped `hbl` login payload** (like `tal`),
  never a broadcast/entity field, so no other player receives it;
- flows through a new **`IWorldActionBar` facet** (`saveActionBarLayout`,
  `takeActionBarLayoutRestore`) implemented in both worlds: offline `Sim` keeps
  `localStorage` as the one store (no-op save), online `ClientWorld` debounces a
  `save_hotbar_layout` wire command and keeps `localStorage` as a mirror.

**Merge rule at login (locked):**

| Server copy | Local layout | Result |
|---|---|---|
| present | (any) | **server WINS** — overwrite the local mirror, re-seed the controller |
| absent | present | local layout **seeds** the first server copy |
| absent | absent | defaults stand (byte-identical to today) |

**Offline play** is byte-identical: `localStorage` remains the one store; the
`Sim` persistence arm is a no-op and never becomes host-aware.

The debounced save is **flushed when the session ends or the page backgrounds**
(`endSession()` covers logout/close; the `visibilitychange: hidden` branch covers
a raw tab close, which routes through `pagehide`, not logout), so a "rearrange
then close the tab within the debounce window" edit still reaches the server for
a second device.

**Untrusted input** is validated + bounded server-side (slot-count, id-length,
and form-key caps in `sanitizeActionBarLayout`); a malformed/oversized/garbage
payload is dropped without crashing the session, and the stored value is
re-validated on read at join. Saves are **debounced** client-side (1.5s, with a
JSON dedup) and serialized per character via a FIFO save queue server-side.

Out of scope, noted as follow-ups: **keybind profiles** and **window positions**
(same "presentation state that should follow the character" family; deliberately
not bundled to keep this diff scoped).

## Related issues

Part of #2228. That issue names two problems; this PR delivers the **cross-device
persistence** half. The **stale-bars-from-an-older-saved-build** half is already
handled in base (`applyLoadoutBar` preserves slots past a legacy snapshot's
length, pinned by `tests/hotbar.test.ts`), so #2228 should **not** be auto-closed
by this PR alone.

## Type of change

## How was this tested?

- Commands:
  - `npm run gate` (i18n gen + malware scan + changed-files biome + full Vitest +
    `tsc` + all builds). Green except two **pre-existing, environmental**
    `deploy_watchdog.test.ts` failures (docker-watchdog timing; confirmed
    identical on a clean `main` checkout with none of these changes, and this PR
    touches nothing in that path).
  - Targeted: `npx vitest run tests/action_bar_layout_sync.test.ts
    tests/action_bar_layout_persistence_game.test.ts
    tests/action_bar_controller.test.ts tests/hotbar.test.ts
    tests/loadout_action_bar.test.ts tests/world_api_parity.test.ts
    tests/command_facets.test.ts tests/command_schema.test.ts
    tests/snapshots.test.ts tests/architecture.test.ts
    tests/localization_fixes.test.ts tests/persistence_round_trip.test.ts
    tests/client_shell.test.ts` — all green.
- New/updated tests:
  - `action_bar_layout_sync.test.ts` — payload bounds validation, capture/apply
    round trip, and the merge-rule (`planActionBarRestore`) unit cases
    (server-present wins / server-absent seeds / both-absent defaults).
  - `action_bar_layout_persistence_game.test.ts` — mocked-db `GameServer` wire
    round trip: client save validates + enqueues with the exact payload; join
    with a stored layout delivers `hbl` to the owning client; a second client
    observing the same entity receives nothing (**self-scope**); garbage/oversized
    payloads are rejected without crashing the session.
  - Controller persistence-seam tests (persist fires only on user changes, not on
    load/reload; offline no-`persistLayout` behavior byte-identical).
  - Parity/wire pins updated in `world_api_parity`, `command_schema`,
    `command_facets`, `snapshots`, and `client_shell`.
- Manual steps: **Live two-device verification was not run locally** (no Docker
  DB available here). The mocked-db suite covers every server seam at the wire
  level. The one thing worth a spot-check on a dev realm after merge is the
  merge-rule **first-sync** path (first login after ship seeding the server copy
  from an existing local layout).

## Screenshots / recordings

N/A. This change is state, not pixels; the acceptance is a localStorage-vs-server
round trip, proved by the mocked `GameServer` tests. (See the merge-rule table
above.)

---

### Quality

- [x] The full gate passes (see note above on the two pre-existing environmental
      `deploy_watchdog` failures, unrelated to this change). Decisive tests added
      for the new behavior.

### Cross-platform

- [x] Determinism / three-host parity preserved: the offline `Sim` stays
      host-agnostic (presentation state never enters the sim), `IWorld` parity
      pins updated in the same change.
- ~[ ] Tested on desktop and mobile.~ No visual/layout change.

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (`tests/localization_fixes.test.ts`
      passes).

### Hygiene

- [x] No secrets/`.env`; `ALLOW_DEV_COMMANDS` untouched.

### DIFF
--- a/server/db.ts
+++ b/server/db.ts
@@ -9,6 +9,7 @@ import { LEADERBOARD_MAX } from '../src/sim/leaderboard_page';
 import { sanitizeRemovedZone1Content } from '../src/sim/removed_zone1_content';
 import type { CharacterState, MailSave, MarketSave } from '../src/sim/sim';
 import type { ArenaFormat, PlayerClass } from '../src/sim/types';
+import type { ActionBarLayout } from '../src/world_api/action_bar';
 import { APPLE_AUTH_SCHEMA } from './apple_auth_db';
 import type { BankBonusFacts } from './bank_entitlements';
 import { seedChatFilterDefaults } from './chat_filter_db';
@@ -216,6 +217,14 @@ ALTER TABLE characters ADD COLUMN IF NOT EXISTS realm TEXT NOT NULL DEFAULT '${R
 -- "last seen" readout on offline guild-roster rows. Nullable: a character that
 -- has never entered the world since this column was added reads NULL.
 ALTER TABLE characters ADD COLUMN IF NOT EXISTS last_login TIMESTAMPTZ;
+-- Per-character action-bar layout (JSONB). This is client PRESENTATION state (a
+-- remap over learned abilities + item shortcuts), NOT deterministic gameplay
+-- state, so it lives in its own additive column rather than the sim-owned state
+-- blob: keeping it out of CharacterState leaves sim serialization byte-identical
+-- and the offline Sim host-agnostic. Nullable/absent until the character first
+-- saves one; the server treats the value as opaque and re-validates its bounds
+-- (sanitizeActionBarLayout) on both read and write.
+ALTER TABLE characters ADD COLUMN IF NOT EXISTS hotbar_layout JSONB;
 -- Max-Level XP Overflow leaderboard: indexed lifetime-XP sort key. The first
 -- index serves the realm-scoped in-game panel; the second serves the global
 -- (cross-realm) home-page board. Both are expression indexes on the bare
@@ -2449,6 +2458,9 @@ export interface CharacterRow {
   force_rename: boolean;
   last_played?: Date | string | null;
   playtime_seconds?: string | number | null;
+  // Per-character action-bar layout (own JSONB column, not the sim state blob).
+  // Opaque to the server beyond bounds validation; only the join path selects it.
+  hotbar_layout?: ActionBarLayout | null;
 }
 
 // The account's "top" character on this realm (highest level, then lifetime XP),
@@ -2502,10 +2514,24 @@ export async function getCharacter(
   characterId: number,
 ): Promise<CharacterRow | null> {
   const res = await pool.query(
-    'SELECT id, account_id, name, class, level, state, is_gm, force_rename FROM characters WHERE id = $1 AND account_id = $2 AND realm = $3',
+    'SELECT id, account_id, name, class, level, state, is_gm, force_rename, hotbar_layout FROM characters WHERE id = $1 AND account_id = $2 AND realm = $3',
     [characterId, accountId, REALM],
   );
   return res.rows[0] ?? null;
+}
+
+/** Persist a character's action-bar layout in its dedicated JSONB column. The
+ *  layout is already sanitized/bounded by the caller (server-side, untrusted
+ *  client input); stored as an opaque document, replaced whole (last write wins).
+ *  Parameterized: characterId is $1, the JSON document is $2. */
+export async function setCharacterHotbarLayout(
+  characterId: number,
+  layout: ActionBarLayout,
+): Promise<void> {
+  await pool.query('UPDATE characters SET hotbar_layout = $2::jsonb WHERE id = $1', [
+    characterId,
+    JSON.stringify(layout),
+  ]);
 }
 
 // Active character names on this realm for the public character sitemap, ranked
--- a/server/game.ts
+++ b/server/game.ts
@@ -86,6 +86,7 @@ import {
   type VcSharedCupInfo,
   type VcViewerReadout,
 } from '../src/world_api';
+import { type ActionBarLayout, sanitizeActionBarLayout } from '../src/world_api/action_bar';
 import { recordOnlineSample } from './admin_db';
 import { offensiveName } from './auth';
 import { recordBankOp } from './bank_ledger';
@@ -130,6 +131,7 @@ import {
   saveMailState,
   saveMarketState,
   setAccountWeaponSkinLoadout,
+  setCharacterHotbarLayout,
   touchCharacterLogin,
   walletForAccount,
 } from './db';
@@ -717,6 +719,12 @@ export interface ClientSession {
     priorGm: boolean;
     stowedPet: PetState | null;
   } | null;
+  // The character's stored action-bar layout as loaded at join (already
+  // bounds-validated), or null when the character has never saved one. Sent to
+  // the owning client exactly once via the `hbl` self field (self-scoped: never
+  // an entity/broadcast field), then frozen; subsequent client saves persist to
+  // the DB and never re-echo here. null wires as an explicit "seed from local".
+  initialHotbarLayout: ActionBarLayout | null;
 }
 
 interface SentEntityVersions {
@@ -1283,6 +1291,10 @@ export class GameServer {
   // state row. Keep one FIFO per account so rapid apply/detach commands cannot
   // commit on separate pool clients in reverse order and resurrect stale state.
   private readonly weaponSkinLoadoutSaveQueues = new Map<number, Promise<void>>();
+  // Action-bar layout is a whole-record replacement in its own character column.
+  // One FIFO per character so a burst of debounced client saves cannot commit on
+  // separate pool clients in reverse order and persist a stale layout.
+  private readonly hotbarLayoutSaveQueues = new Map<number, Promise<void>>();
   // Serializes every write of the single global Market blob (the 30s autosave
   // and the leave-path combined save). Both serialize the whole market; without
   // a queue their transactions could commit out of capture order and persist an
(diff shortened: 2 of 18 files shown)


FernandoX7 (commented):
This is well built. The architectural call is the right one: keeping the layout in its own additive `characters.hotbar_layout` JSONB column instead of `CharacterState` leaves sim serialization byte-identical and the offline `Sim` host-agnostic, and reaching the owning client through the self-scoped `hbl` payload (never an entity/broadcast field, like `tal`) is the correct wire shape for per-character presentation state. The untrusted-input handling (`sanitizeActionBarLayout` on both write and read), the per-character FIFO save queue, the client debounce+dedup, and the pure `planActionBarRestore` merge rule are all clean. The tests are decisive: the no-leak case that asserts `linen_bandage` appears nowhere in an observer's frame, plus explicit null-vs-value `hbl`, FIFO ordering, and garbage rejection, are exactly the pins this needs.

I re-ran the load-bearing pins at the PR head (world_api_parity, snapshots delta-key count, command_schema, command_facets, the new sync/persistence/client suites, the S3 `localization_fixes` guard, and architecture) plus `tsc --noEmit`: all green.

Findings:

- nit: the debounced wire save is never flushed on logout or socket close. `saveActionBarLayout` schedules a 1.5s timer (`src/net/online.ts:3169`), but `close()` (`src/net/online.ts:1680`) and `sendLogout()` neither clear nor fire it. Rearrange bars, close the tab within the debounce window, and never touch them again, and that final arrangement never reaches the server. The localStorage mirror keeps it correct on the same device, so this only bites the cross-device case and only for the last sub-1.5s of edits. A flush in `close()`/`sendLogout()` (or a `beforeunload`) would close it.

Scope note (not a finding): #2228 names two problems, and this PR delivers the cross-device half. The loadout-build half (switching to an old build must not wipe slots past its snapshot) is already handled in base: `applyLoadoutBar` preserves the tail for a short saved bar (`src/ui/hud/action_bar/hotbar.ts`) and `tests/hotbar.test.ts:539` pins it, so scoping this PR to persistence is reasonable. Worth confirming the issue is not assumed fully closed by this PR alone. One detail from the issue's coverage-gap note reads stale against current code: the loadout load path is `bar: [...l.bar]` (`src/sim/sim.ts:2878`), not `bar: [...(l.bar ?? [])]`; a truly `bar`-less legacy JSONB row would throw there rather than clear, but that path is pre-existing and outside this diff.

Otherwise this looks mergeable.

seanghods:
Good catch on the flush, fixed in e8f7880. The pending layout is now flushed from `endSession()` (logout/close, while the socket is still open) and from the `visibilitychange: hidden` branch, since a raw tab close routes through `pagehide` rather than `sendLogout`, so that branch is what covers the "rearrange then close the tab" case. The flush cancels the debounce timer so it can't double-send; added client tests for both.

On the scope note: agreed, this is only the cross-device half of #2228. The stale-bars-on-an-older-build half is already handled in base (`applyLoadoutBar` tail-preservation, pinned by `tests/hotbar.test.ts`), so I linked it as "Part of #2228" rather than Closes. Thanks for the thorough review.

FernandoX7 (approved):

PR #2241, opened 2026-07-21
feat(ui): Ascendant Sigils badge reskin (holder + contributor)

seanghods:
## Summary

Replaces the placeholder `$WOC` badge art with the approved **Ascendant Sigils** design system across both the holder-tier ladder and the contributor (dev-tier) ladder. The old badges were flat discs with placeholder line art; the whole 2%-9% holder band shared one gem-plus-dots glyph, and the contributor badges were flat merge-graph discs.

Rank grammar reads at nameplate size: **the frame shape encodes the band, the material encodes the rung.** Every badge stays font-free inline SVG (gradients and strokes only, no `<filter>`, `viewBox 0 0 64 64`), composed through the unchanged `holderTierBadgeDataUrl` / `devTierBadgeDataUrl` signatures. No new dependencies, no asset files.

**Holder ladder (`src/ui/holder_tier.ts`)**
- Band I (1-5): faceted coins, milled edge, material ramp (ember-copper to gold), bold filled glyphs.
- Band II (6-8): beveled cut-gem hexagons in jade / sapphire / amethyst.
- Band III (9-16): eight-point sigil plates with a **nine-socket ring** lighting N sockets for an N%-of-supply rung (fixing the eight-shared-glyph bug: the count survives close-up and the arc length reads at a distance).
- Band IV (17-18): unique regalia (Worldbearer orb in crescents; Sovereign ray-burst and jeweled crown).

**Contributor ladder (`src/ui/dev_tier.ts`)**
- A separate honor gets a separate silhouette so earned rank is never mistaken for bought rank: the five merged-PR rungs keep their merge-graph metaphor (a node per merged PR, more nodes and merged lanes per rung) on a new **riveted banner-shield** frame, climbing a maker's material ramp (steel, blued steel, runed violet, gold, and a five-ray crest for Worldwright). Thresholds, keys, i18n names, ring/glow, and `src/sim/dev_tier.ts` are untouched.

**Standout halo (both ladders).** Badges keep their current rendered size. A static CSS drop-shadow halo, colored by each tier's glow hue (via a `--holder-glow` / `--dev-glow` property the painters set, cheap-diffed on tier), makes them pop: subtle for ordinary rungs, moderately stronger for the apex rungs (holder band IV regalia; the two significant-contributor tiers, Architect and Worldwright). Strengths are two named tunables in `tokens.css`, shared by both ladders and frozen from the staged screenshots. Cosmetic-only and static, so there is no per-frame cost. The holder and contributor badges coexist on a nameplate (there is no precedence); each halo applies to its own badge element independently.

The design was approved out of band; the before/after screenshots below are the spec evidence.

## Type of change

- [x] Feature: new functionality (visual reskin)

## How was this tested?

- Commands: `npx vitest run tests/holder_tier.test.ts tests/dev_tier.test.ts` (extended, see below); `npx tsc --noEmit`; `npx vitest run tests/architecture.test.ts tests/nameplate_view.test.ts tests/hud_perf_budget.test.ts tests/css_corpus.test.ts tests/css_value_validity.test.ts tests/ui_effects_wiring.test.ts tests/pr_shot_targets.test.ts tests/localization_fixes.test.ts tests/dev_tier_ui.test.ts` plus the player-card suites; all green. Full `npx vitest run` and all `vite build` entries green apart from pre-existing sandbox flakes (sfx ffmpeg true-peak; two heavy sim suites that pass on isolated re-run), none in touched areas.
- New decisive pins:
  - Holder: 18 pairwise-distinct URLs; band III socket count per rung (red before, green after); no `<filter>`; ring/glow + keys/thresholds pinned to `src/sim/holder_tier.ts`; the regalia flag and card-halo class.
  - Contributor: 5 pairwise-distinct URLs; per-rung merge-graph node and merged-lane counts (red before, green after); the crest on Worldwright alone; no `<filter>`; ring/glow + keys/thresholds pinned to `src/sim/dev_tier.ts`; the inspect halo class strong only for the significant tiers.
- Manual steps: staged the holder and contributor ladders on real nameplates (offline harness), over a bright and a darkened scene, plus the inspect/holder card and the contributor card. `src/sim/holder_tier.ts` and `src/sim/dev_tier.ts` are untouched.

## Screenshots / recordings

Holder ladder, contributor ladder, and the inspect cards, before and after. Captured via the `holder-tier` entry in `scripts/pr_shot_targets.mjs`.

**Holder ladder**

| | Before | After |
|---|---|---|
| Ladder (bright terrain) |  |  |
| Ladder (dark scene) |  |  |
| Close-up |  |  |
| Holder card |  |  |

**Contributor ladder**

| | Before | After |
|---|---|---|
| Ladder (bright terrain) |  |  |
| Ladder (dark scene) |  |  |
| Contributor card |  |  |

---

### Localization (i18n)

- [x] N/A. This PR adds no player-visible strings (badge art is SVG; tier names/flavors are the existing `t()` keys, unchanged).

### Hygiene

- [x] No secrets committed; `ALLOW_DEV_COMMANDS` not enabled anywhere.


FernandoX7 (commented):
Clean reskin. The rank grammar (frame encodes the band, material encodes the rung) genuinely fixes the shipped eight-shared-glyph bug, and the diff stays tight: the `holderTierBadgeDataUrl(tier, px)` signature is preserved, `src/sim/holder_tier.ts` and the ring/glow hue ramp are untouched, and all three badge consumers keep working (the shareable player-card canvas at `player_card.ts:251` picks up the new art with no change).

Things I checked and liked:
- Purity: `src/ui/holder_tier.ts` stays DOM/Three-free, font-free and `<filter>`-free SVG, pure trig, so it unit-tests in Node.
- No per-frame cost: the painter writes `--holder-glow` and toggles `np-tier-regalia` only on the tier cheap-diff (`nameplate_painter.ts:479`), the same shape as the neighbouring dev-outline. The halo is static CSS, reduced-motion safe.
- play.html parity holds: the halo CSS lives in shared `src/styles/` modules imported via `src/main.ts` (which both index.html and play.html load), and the `--holder-halo` tokens are `:root`-scoped so the nameplate and inspect-card surfaces both inherit them.
- i18n: no new strings, existing `wallet.*` keys reused.
- Tests are decisive: the socket-count pin keys off the `r="1"` cream pip (red before the reskin, which had zero sockets), the no-filter guard checks both the decoded and URI-encoded forms, and the hue ramp plus ladder keys/thresholds are frozen against `src/sim/holder_tier.ts`.

Nit:
- `nameplate_painter.ts:479-480`: the painter-side wiring (the `--holder-glow` set and `np-tier-regalia` toggle) is not itself pinned; only the pure helpers `holderTierIsRegalia` / `holderCardBadgeClass` have tests. A regression that dropped the nameplate class toggle would not fail a test. This mirrors the existing (also unpinned) dev-outline pattern beside it, so it is consistent, just flagging the one branch a test would not catch.

Looks mergeable.

jgyy (commented):
Nice fix for the Band III shared-glyph problem: pinning the ladder keys/thresholds/hue ramp against `src/sim/holder_tier.ts` in the tests is exactly the right regression guard for this kind of cosmetic ladder. The tier-diff guard in `nameplate_painter.ts` keeps the new halo writes out of the per-frame path, and the halo itself is static (no animation), so it stays clear of the graphics-fairness rule.

nit: `holder_tier.ts` indexes `BAND1`/`BAND2` via an `as Band1Key`/`as Band2Key` cast on `tier.key`. If a future rung ever moved bands without a matching `BAND1`/`BAND2` entry, this would throw at runtime instead of failing to compile. The lockstep test would likely catch a rename, but not this specific case. Not blocking, just worth a mental note next time the ladder changes.

Screenshots, tests (17/17), and `tsc` all check out. No i18n surface here since it's pure SVG badge art.

seanghods:
Thanks for reviews. Hold on merging for 1-2 hours until contributor badges are updated in this PR as well. Thx

seanghods:
Ready

PR #2245, opened 2026-07-21
fix(render): character preview never loads on a cold, first-visit cache

jgyy:
## Problem

New players sometimes see no character model at all on the Offline
Character / character-creation screen, on their very first visit to the
site in a given browser. A page reload usually "fixes" it.

## Root cause

```mermaid
sequenceDiagram
    participant Browser as Browser (cold cache)
    participant Preload as assets/preload.ts (site-wide)
    participant MainTS as main.ts (landing preview)
    participant Preview as CharacterPreview

    Browser->>Preload: fetch terrain, dungeon, foliage, character GLBs...
    Note over Preload: ONE shared assetsReady() promise<br/>over EVERY registered preload
    Preload-->>Preload: an unrelated asset (e.g. terrain texture)<br/>fails all its retries (cold cache, flaky first load)
    Preload--xMainTS: assetsReady() REJECTS
    Note over MainTS: assetsReady().then(...) has NO .catch()
    MainTS--xPreview: the .then() callback never runs
    Note over Preview: CharacterPreview is never constructed.<br/>Canvas stays permanently blank.<br/>Only an unhandled-rejection in devtools.
```

`assetsReady()` (`src/render/assets/preload.ts`) is a single barrier over
*every* preload registered anywhere in the render stack. The landing
character-creation preview in `src/main.ts` awaited that barrier with a bare
`.then()` and no failure handler at all. If even one *unrelated* asset (a
terrain splat, a dungeon prop, anything) fails all of its internal retries,
the shared promise rejects, the `.then()` callback that builds
`CharacterPreview` silently never runs, and nothing ever retries it for the
rest of the page's lifetime.

This is much more likely on a cold, first-visit cache: there's no warm HTTP
cache to fall back on, so a slow or flaky first connection is more likely to
exhaust an asset's bounded retry budget within the boot window. A reload
"fixes" it only because the browser's HTTP cache is warm by then, which is
why it reads as a first-visit-only bug.

## Fix

- Added `charactersReady()` in `src/render/characters/assets.ts`: a
  narrower gate that only waits on the character-boot GLBs/skin atlases
  (not the unrelated site-wide preload set), and retries whatever is still
  missing (up to 3 attempts). `loadGltf`/`loadTexture` already evict a
  failed URL from their own cache on rejection, so a fresh call here is a
  real re-fetch, not a re-await of a permanently rejected promise.
- Wired `src/main.ts`'s landing preview init through `charactersReady()`
  instead of the bare `assetsReady()`, with a `.catch()` that logs instead
  of leaving an unhandled rejection if every retry is exhausted.

## Test plan

- [x] New `tests/character_preview_boot.test.ts`: covers a transient
      per-asset failure that succeeds on retry, and confirms retries are
      bounded (rejects with a clear error instead of hanging forever).
- [x] `npx vitest run` (full suite): all green except two known-unrelated
      flakes that also fail identically on a clean checkout of this branch
      before my change (a timing-sensitive mob-despawn test that passes in
      isolation, and two i18n registry tests that just need `npm run
      i18n:gen`, which `pretest` runs and a bare `vitest run` skips).
- [x] `npx @biomejs/biome check` on changed files: no new warnings.

No visual/UI change under normal conditions (this only changes error/retry
handling around asset loading), so no before/after screenshots.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -143,7 +143,7 @@ import { resolveWalletCapability } from './net/wallet_capability';
 import { installWalletResumeHandlers } from './net/wallet_resume';
 import { assetsReady } from './render/assets/preload';
 import { CharacterPreview, type PreviewAppearance } from './render/characters';
-import { preloadMechAssets } from './render/characters/assets';
+import { charactersReady, preloadMechAssets } from './render/characters/assets';
 import { skinCount } from './render/characters/manifest';
 import { playerPortraitDataUrl } from './render/characters/portrait';
 import { installWebGLContextRelease } from './render/context_release';
@@ -9033,38 +9033,51 @@ function wireStartScreens(): void {
     syncLandingGraphicsSelect();
   });
 
-  // Initialize 3D character preview once assets are ready
-  assetsReady().then(() => {
-    // Resolve each panel defensively: play.html (online-only) has no #offline-select.
-    const activePanelId = ['#charselect-panel', '#offline-select'].find((id) => {
-      const panel = $(id) as HTMLElement | null;
-      return panel !== null && !panel.hasAttribute('hidden');
-    });
-    const containerId =
-      activePanelId === '#offline-select'
-        ? '#offline-preview-container'
-        : '#online-preview-container';
-    const container = $(containerId);
-    const canvas = $('#char-preview-canvas') as HTMLCanvasElement | null;
-    if (container && canvas) {
-      characterPreview = new CharacterPreview(container, canvas);
-      // If a token auto-login already rendered the roster and selected a
-      // character before assets finished, show its real appearance; otherwise
-      // fall back to the selected class chip (create/offline panels).
-      if (charselectSelected) {
-        characterPreview.setAppearance(charselectAppearance(charselectSelected));
-      } else {
-        const selSelector =
-          activePanelId === '#offline-select'
-            ? '#offline-select .mini-class.sel'
-            : '#charcreate-panel .mini-class.sel';
-        const selEl = document.querySelector(selSelector) as HTMLElement | null;
-        const cls = selEl ? (selEl.dataset.class as PlayerClass) : 'warrior';
-        characterPreview.setClass(cls);
+  // Initialize 3D character preview once its assets are ready. Gated on the
+  // narrower charactersReady() (with its own retries), not the site-wide
+  // assetsReady(): that single shared promise covers EVERY registered preload
+  // (terrain, dungeon, foliage, ...), so an unrelated failure there must never
+  // permanently blank the character-creation preview. This previously used
+  // assetsReady().then() with no failure handler at all, so on a cold,
+  // first-visit cache a single transient asset failure anywhere on the site
+  // silently stranded the preview forever (issue: new players saw no
+  // character model on their first load; a reload "fixed" it only because the
+  // browser's HTTP cache was warm by then).
+  charactersReady()
+    .catch((err: unknown) => {
+      console.error('character preview assets failed to load, preview will stay blank:', err);
+    })
+    .then(() => {
+      // Resolve each panel defensively: play.html (online-only) has no #offline-select.
+      const activePanelId = ['#charselect-panel', '#offline-select'].find((id) => {
+        const panel = $(id) as HTMLElement | null;
+        return panel !== null && !panel.hasAttribute('hidden');
+      });
+      const containerId =
+        activePanelId === '#offline-select'
+          ? '#offline-preview-container'
+          : '#online-preview-container';
+      const container = $(containerId);
+      const canvas = $('#char-preview-canvas') as HTMLCanvasElement | null;
+      if (container && canvas) {
+        characterPreview = new CharacterPreview(container, canvas);
+        // If a token auto-login already rendered the roster and selected a
+        // character before assets finished, show its real appearance; otherwise
+        // fall back to the selected class chip (create/offline panels).
+        if (charselectSelected) {
+          characterPreview.setAppearance(charselectAppearance(charselectSelected));
+        } else {
+          const selSelector =
+            activePanelId === '#offline-select'
+              ? '#offline-select .mini-class.sel'
+              : '#charcreate-panel .mini-class.sel';
+          const selEl = document.querySelector(selSelector) as HTMLElement | null;
+          const cls = selEl ? (selEl.dataset.class as PlayerClass) : 'warrior';
+          characterPreview.setClass(cls);
+        }
       }
-    }
-    decorateClassChips();
-  });
+      decorateClassChips();
+    });
 }
 
 // Looping home-page theme. Browsers block audio autoplay until a user gesture,
--- a/src/main.ts
+++ b/src/main.ts
@@ -145,7 +145,7 @@ import { assetsReady } from './render/assets/preload';
 import { CharacterPreview, type PreviewAppearance } from './render/characters';
 import { charactersReady, preloadMechAssets } from './render/characters/assets';
 import { skinCount } from './render/characters/manifest';
-import { playerPortraitDataUrl } from './render/characters/portrait';
+import { onPortraitsReady, playerPortraitDataUrl } from './render/characters/portrait';
 import { installWebGLContextRelease } from './render/context_release';
 import { firstRunGraphicsPreset, GFX, graphicsPresetLabel } from './render/gfx';
 import { Renderer } from './render/renderer';
@@ -3554,8 +3554,10 @@ function renderSkinPicker(
   });
 }
 
-/** Give each class button a small portrait preview of that class (run once
- *  character assets are ready so portraits render synchronously). */
+/** Give each class button a small portrait preview of that class. Wired to
+ *  {@link onPortraitsReady} so it runs once portrait.ts's own asset barrier
+ *  resolves (portraits render synchronously from then on); one-shot per chip
+ *  via the .mini-class-portrait guard below, so it is safe to call again. */
 function decorateClassChips(): void {
   document
     .querySelectorAll<HTMLElement>('#charcreate-panel .mini-class, #offline-select .mini-class')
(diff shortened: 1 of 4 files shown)


FernandoX7 (commented):
The diagnosis is right and the write-up is unusually good. `assetsReady()` really is one shared barrier over every registered preload (`src/render/assets/preload.ts:7-16`), the old `.then()` really had no failure handler, and the claim that `loadGltf` evicts a rejected URL so a fresh call is a real re-fetch checks out (`src/render/assets/loader.ts:104-110`). Narrowing the preview's gate is the right instinct.

I ran the failure case against a dev server (abort one character GLB for its first three fetches, then serve it) and the preview does come back. That same run turns up one thing that needs another pass.

**Findings**

1. `should-fix` `src/main.ts:9079` (with `src/render/characters/portrait.ts:52-62`): `decorateClassChips()` moved onto a gate that no longer implies portraits are available, and it is one-shot.

   `portrait.ts` latches its own readiness off a separate `assetsReady()` call made at import time. That snapshot is exactly the character preload set (I instrumented it: 138 tasks, the 118 GLBs plus the boot skin atlases), so it rejects on the very failure `charactersReady()` now recovers from, and `assetsAreReady` stays false for the life of the page. The new `.then()` runs regardless, and `decorateClassChips` bails out on any later call (`src/main.ts:3563`), so the result is permanent.

   Repro under `npm run dev`: abort `models/chars/players/knight.glb` for its first three fetches (`MAX_LOAD_ATTEMPTS`), then serve it. The preview renders, which is the fix working, but `portraitsReady()` is false when the chips are decorated: 0 of 18 chips get a portrait `src` while 18 of 18 get `has-portrait`, so the offline panel shows nine empty circles. The same run on `release/v0.29.0` leaves the chips as plain labels.

   Cheapest fix: decorate the chips off `onPortraitsReady()` (portrait.ts exports it for exactly this) instead of folding them into the preview gate.

2. `should-fix` `src/main.ts:9047-9049`: the `.catch()` sits before the `.then()`, so it is not a guard. On an exhausted retry it logs "preview will stay blank" and the `.then()` body then runs anyway, constructing a live `CharacterPreview` (a second WebGL context) and doing the chip rewrite above. It survives only because `setVisualKey` already try/catches (`src/render/characters/preview.ts:165-177`). Either move the catch to the end of the chain, or keep the order and say in the comment why the body must still run.

3. `nit` `src/render/characters/assets.ts:484-497`: no delay between outer attempts. `loadGltf` has already spent its own three attempts with 400/800 ms backoff before rejecting, so these three fire back to back inside one tick. On the flaky cold connection this targets that is three more instant hammerings rather than a wider retry window.

4. `nit` `tests/character_preview_boot.test.ts:33`: `resolves.toBeUndefined()` passes both when the loop returns early because the caches filled and when it simply falls out after `maxAttempts` with nothing rejected, so a regression in the `gltfByUrl.has(assetUrl(u))` readiness check would slip through. It does take the early return today (each URL loads exactly twice under the mock). Pinning that call count, or a follow-up `charactersReady(1)` that triggers no further loads, makes the assertion decisive.

5. `note`, not this PR's job: `assetsReady()` re-awaits the same stored promises, so the transient failure that motivated this still leaves `startGame`'s gate rejected and the player hits `fatalOverlay(t('loading.assetsFailed'))` (`src/main.ts:963`) the moment they press Enter World. The preview is healed, world entry is not. The durable fix is the same evict-and-retry inside `assetsReady`/`registerPreload`, which is worth its own issue.

6. `nit`: the module header at `src/render/characters/assets.ts:6-8` still states the loading contract as "main.ts awaits assetsReady()", and the `assets.ts` entry in `src/render/characters/CLAUDE.md` does not mention the new gate. Those are the two anchors the next person reads.

Base branch, conflict scope and CI are all clean, and the console text is correctly dev-channel English, so there is no i18n surface here.

jgyy:
Thanks for the thorough repro on all of these, especially catching that the chip decoration had quietly moved onto the narrower gate. Addressed in 26ae51541:

1. `should-fix` (chip decoration on the wrong gate): moved `decorateClassChips()` off the `charactersReady()` chain entirely and wired it to `onPortraitsReady()` (`src/main.ts`), which is exactly the hook `portrait.ts` exports for this. Chips now decorate off portrait.ts's own asset barrier, independent of whether the 3D preview's narrower recovery path fired. Confirmed the guard in `decorateClassChips` (skips a chip that already has `.mini-class-portrait`) makes re-registering it here safe.

2. `should-fix` (catch before then): moved the `.catch()` to the end of the promise chain so it actually guards the `.then()` body instead of logging and then running the body unconditionally. No behavior loss: on an exhausted retry the preview container is simply never constructed, matching the "stay blank" log message.

3. `nit` (no delay between outer attempts): added a backoff delay before each retried outer attempt in `charactersReady()`, reusing the existing `retryDelayMs` schedule from `load_retry.ts` (same 400ms/attempt curve `loadGltf` already uses for its inner retries) instead of inventing a new number.

4. `nit` (non-decisive test assertion): `tests/character_preview_boot.test.ts` now tracks per-URL call counts and asserts every URL was fetched exactly twice under the fail-once mock, so a regression in the `gltfByUrl.has(assetUrl(u))` readiness check (early-return firing when it shouldn't, or not firing when it should) would fail the count instead of slipping through `resolves.toBeUndefined()`.

5. `note` (assetsReady() / startGame gate): read and agreed it's out of scope here; not touched. Worth its own issue as you said, since the durable fix (evict-and-retry inside `assetsReady`/`registerPreload`) is a different, broader change.

6. `nit` (stale header comments): updated the `assets.ts` module header to describe `charactersReady()` as the narrower exception to the "main.ts awaits `assetsReady()`" contract, and added a `charactersReady()` paragraph to `src/render/characters/CLAUDE.md`'s `assets.ts` entry pointing at `tests/character_preview_boot.test.ts`.

Verified: `npx tsc --noEmit` clean, `npx vitest run tests/character_preview_boot.test.ts` (2/2, including the new pinned-count assertion), `tests/architecture.test.ts` and `tests/localization_fixes.test.ts` still green (55 passed, 3 pre-existing skips), and `npx @biomejs/biome check --write` on the four changed files (only fixed an import-sort in `assets.ts`; the pre-existing `noNonNullAssertion` warnings in `portrait.ts` are untouched debt outside this diff's scope). Pushed to `fix/character-preview-first-load`; the pre-push QA floor (tsc, guard tests, changed-file biome, copy scan) passed.

FernandoX7 (commented):
26ae515 covers all four actionable items, and item 5 was correctly left alone as the out-of-scope note it was. I re-ran the whole thing rather than reading the diff, so here is what I actually checked.

**How I verified it**

Fresh worktree at the PR head, `npm run dev`, and Chromium driven through four boot scenarios with request interception. `models/chars/players/knight.glb` is the character asset I failed; `MAX_LOAD_ATTEMPTS` is 3, so aborting its first three fetches and then serving it is a genuine transient failure that rejects the registered preload task while still letting `charactersReady()`'s outer retry recover. "Preview pixels" is a count of opaque pixels read back off `#char-preview-canvas` with the panel unhidden before boot so the container has a real size.

| scenario | preview | class chips |
|---|---|---|
| clean load | renders, 135,526 opaque px | 18 of 18 portraits |
| unrelated non-character assets fail (236 requests aborted) | renders | 18 of 18 portraits |
| character GLB fails 3x then serves | recovers, 135,482 opaque px | plain labels, no empty circles |
| character GLB fails permanently | blank canvas, one logged error, no unhandled rejection | untouched |

Row two is the scenario this PR exists for, and it now works end to end: an unrelated preload failure no longer touches the preview or the chips. Row three is the one I reported: the nine empty circles are gone, and the chips fall back to exactly what `release/v0.29.0` shows in the same run. Row four confirms the catch reorder, the `.then()` body no longer runs on an exhausted retry and nothing escapes as an unhandled rejection.

Also clean: `npx tsc --noEmit`, `tests/character_preview_boot.test.ts` plus `render_asset_preload` and `character_visual_fail_soft`, and a dash/emoji scan over the added lines. The backoff is correctly placed after the early return, so a warm or already-cached path never sleeps.

**Findings, all nits, none blocking**

1. `nit` `src/main.ts:9036-9043`: the new comment is inverted on both of its factual claims. portrait.ts's barrier is not the "wider" set: its import-time `assetsReady()` snapshot is exactly the character preload set (I instrumented it at 138 tasks, the boot GLBs plus the skin atlases, because `characters/assets.ts` is the only module that has called `registerPreload` by the time portrait.ts evaluates). And the failure mode was not that a chip "stayed a plain label", which is the new and correct behavior; it was the chip being rewritten to a src-less `` plus `has-portrait`. The real difference between the two gates is retry plus fresh fetch versus re-awaiting one permanently rejected promise, and that is what is worth writing down, since this comment is the anchor the next reader will trust.

2. `nit` `tests/character_preview_boot.test.ts:38`: the `count === 2` pin is genuinely decisive for the retry behavior, but the comment's third case is unreachable. I deleted `if (missingGltf.length === 0 && missingSkins.length === 0) return;` and both tests stayed green: the missing sets are recomputed from the caches on each attempt, so the extra pass is just an empty `Promise.allSettled` and the per-URL count stays 2. The early return is behavior-neutral, a wall-clock optimization, so this is comment accuracy rather than a coverage hole. Trimming the claim to what the pin actually catches would keep it honest.

3. `nit` `src/main.ts:9089`: with the catch last it now also swallows throws from inside the `.then()` body, so a WebGL context or DOM failure during `CharacterPreview` construction gets logged under "character preview assets failed to load". That is a dev-console label only, and it is the better trade against the previous ordering, so I am noting it as a known consequence rather than asking for a change.

**Worth its own issue**

Two things are unchanged from base, so nothing regressed here, but they are the same underlying defect and they will bite again:

- `assetsReady()` re-awaits the same stored task promises (`src/render/assets/preload.ts`), so a once-rejected preload rejects for the life of the page. The transient failure that motivated this PR still sends `startGame` into `fatalOverlay(t('loading.assetsFailed'))` at `src/main.ts:963`. The preview is healed, world entry is not.
- portrait.ts's readiness latch is never re-armed (`src/render/characters/portrait.ts:52-62`). After `charactersReady()`'s retry recovers the GLBs the portraits are perfectly renderable, but `assetsAreReady` is already latched false, so `playerPortraitDataUrl` keeps returning null and the chips keep their plain labels for the rest of the session.

Could you file an issue covering both? The shape of the fix is the same evict-and-retry this PR proved out, applied at the `registerPreload`/`assetsReady` layer so every consumer of that gate heals together instead of each one growing its own narrow barrier. Link it back to this PR for the reproduction recipe.

Nothing here blocks the merge. The one red check is a cancelled `PR AI assist` run rather than a code failure, so a re-run should clear it.

FernandoX7 (approved):

PR #2248, opened 2026-07-21
feat(ui): character sheet and inspect showcase redesign

seanghods:
## Summary

Adopts one "showcase" visual language across the **character sheet** (`#char-window`) and the **inspect / Profile window** (`#inspect-window`), from an approved rendered-artifact design loop. Both windows now read identically because sockets, columns, and the band-centering contract are shared.

Character sheet:
- 740px window with a **fixed-width staged 3D model** (keyline, warm gold haze, inset vignette, ground-shadow ellipse); the columns flex around it.
- Paperdoll rebalanced to **6/6**: the offhand moves under the mainhand in the left column (a one-line `char_view` slot swap the inspect window inherits).
- **44px quality-glow sockets** (2px border, radius 7), 9px uppercase slot labels, item names that wrap in the quality color (never ellipsized).
- Header meta becomes **pill chips** (level/class, title, hobby, honor), with the honor chip de-golded to the shared muted color.
- Stats become **five primary tiles** plus colon-free **Offense** and **Defense** panels; Specialization / Progression / Gathering gain the same inset-card treatment. The stat tooltip / a11y plumbing is untouched.

Inspect window:
- Same socket / column / centering language, with a **live class-colored turntable** of the inspected player (their class, skin, and worn weapons), the stage border / glow / haze taking the inspected player's class color via a CSS custom property.
- Compact header (name + level/class chip), the big portrait chip dropped; holder / Discord / dev badge rows kept with identical semantics.
- Extracted out of `hud.ts` into a proper module pair: a **pure `inspect_view` core** (header, badge gating, worn gear, remote model) plus a **thin `inspect_window` painter** that now installs a `markDialogRoot` focus trap (closing the a11y gap the old inline path had). `hud.ts` net-shrinks to thin `open*` delegates.

The turntable framing pulls back for inspect (z 6.6 / y 1.5 vs the sheet's z 5.1 / y 1.45) via a new `CharacterPreview` framing option backed by pure, Node-pinnable constants; reopening the character sheet re-asserts the self model and its close framing.

Client-only: no changes under `src/sim/`, `server/`, `headless/`, or the `src/net` wire encoding, and nothing new crosses the wire. No stats are shown on inspect.

## Type of change

- [x] Refactor or performance (the inspect extraction; behavior preserved)

## How was this tested?

- Commands:
  - `npx tsc --noEmit` (clean)
  - Full Vitest suite excluding the known sandbox flake: `npx vitest run --exclude tests/deploy_watchdog.test.ts` -> **17419 passed, 0 failures**
  - `npm run build` -> all five entries + i18n bundles + media built
  - `npm run gate` is green **except** the two `tests/deploy_watchdog.test.ts` "abandons a hung docker ..." cases, which fail identically on a pristine release branch in this sandbox (fake-docker timing) and pass in CI. `tsc` and `build` were therefore run directly and are green (both above).
  - New/updated decisive tests: `char_stats_view` (the tiles/offense/defense partition equals `STAT_GRID`), `inspect_view` (header, badge gating, member-days math, gear reuse, remote model), `inspect_window` (focus-trap + thin-painter pins), `quality_glow` (exact shadow per quality color), `preview_framing` (the two framings), `charscreen_css` (whitespace-normalized CSS pins), plus updated `char_view`, `stat_tooltip_view`, `char_window`, `inspect_instances`, `v027_port_restorations`, `deeds_window`, `client_shell`.
- Manual steps (offline client, warrior "Thalric" fully geared, level 45, mage bot "Elowen"):
  - Ran the reference centering probe on **both windows, desktop and 844x390 mobile**. Pass on all four: `|edgeGap - stageGap| <= 6` per column and identical icon x per column (desktop edge 29 / stage 33; mobile edge/stage equal). 
  - Verified the inspect stage takes the mage class color (`#69ccf0`), the live model holds her weapon, reopening the character sheet restores the self model (shared-canvas round-trip), and the out-of-range remote profile card is unchanged.

### Character sheet (desktop)

| Before | After |
|---|---|
| before | after |

### Inspect / Profile (desktop)

| Before | After |
|---|---|
| before | after |

### Character sheet (mobile, 844x390 landscape)

| Before | After |
|---|---|
| before | after |

### Inspect / Profile (mobile)

| Before | After |
|---|---|
| before | after |

---

### Quality
- [x] Decisive tests added/updated for the changed behavior; manual checks recorded above. Full gate green modulo the documented `deploy_watchdog` sandbox flake (`tsc` + `build` verified directly).

### Cross-platform
- [x] Client-only. No `src/sim/`, `server/`, `headless/`, or `src/net` wire changes; nothing new crosses the wire; both `IWorld` worlds untouched.

### UI / i18n
- [x] Every new player-visible string is a `t()` key (`hudChrome.charSheet.offense` / `.defense`), added English-only to the catalog; the M16 non-Latin fills ship in the same change. The stat-tooltip a11y plumbing and focus handling are preserved, and the inspect window now gains a focus trap.


jgyy (commented):
Nice piece of work. The unified showcase language for the character sheet and inspect window is genuinely well-seamed: the 6/6 paperdoll split lives once in `char_view.ts` and both windows inherit it, `quality_glow.ts` and `char_stats_view.ts` are clean DOM-free cores with decisive tests (every badge gate in `inspect_view.test.ts` has a negative case), and the `inspect_view.ts` / `inspect_window.ts` split out of `hud.ts` is a real net shrink of the coordinator, not just a move. The added `markDialogRoot` focus trap on the inspect window closes a real a11y gap the old inline card had. `tsc` is clean, all 14 touched/added test files pass (304 tests), and the merge against `release/v0.29.0` is conflict-free.

Two mobile touch-target regressions worth fixing before merge:

- **should-fix** `src/styles/components.css:10102-10108` and `src/styles/hud.mobile.css:3139-3143` — the paperdoll unequip corner chip shrank from 40x40px (its size on `release/v0.29.0`) to 22x22px, and the `body.mobile-touch` override pins it at 22x22 explicitly (always-visible on touch, so this is the size a thumb actually gets). That's below even the 24px WCAG 2.2 SC 2.5.8 absolute floor, and `src/ui/CLAUDE.md:38-40` is explicit: "Do NOT weaken the 40x40 floor to 24px." Worth restoring a >=24px (ideally 40px) hit area, e.g. an invisible padded hit-expander, while keeping the chip visually small.

- **should-fix** `src/styles/hud.mobile.css:3176-3180` — the five primary-attribute tiles are interactive stat cells (`tabindex=0`, long-press tooltip peek, same family as the Offense/Defense cells). `body.mobile-touch .stat-panel .stat-cell` correctly keeps the `min-height: 40px` long-press floor, but the more specific `.attrs-tiles .stat-cell` override right below it resets `min-height: 0` with `padding: 5px 2px` and an 8px label, which drops the tiles under the 40px floor while their sibling panels keep it. Looks like an oversight rather than an intentional exception.

Two non-blocking notes:

- `src/ui/quality_glow.ts` is DOM-free and unit-tested but isn't registered in `UI_PURE_CORES` / `BARE_NAMED` in `tests/architecture.test.ts` (it doesn't match the `*_view`/`*_core` naming the sweep looks for), so its purity isn't guarded going forward. Worth registering it or renaming to `quality_glow_core.ts`.
- The glow's shadow values (`9px`, `0.45` alpha, etc.) are TS literals rather than tokens, but since this only runs on the cold open path (not a per-frame painter), that's fine as is.

Everything else checked out: i18n is English-only in the right catalog (`hudChrome.charSheet.offense/defense`), no gameplay-relevant info is hidden behind the cosmetic tier, the styles stay layer-correct, and `color-mix()`/CSS custom properties for the inspect class-color stage are a nice touch that avoids per-class CSS forks.

seanghods:
Thanks for the careful read. Both touch-target regressions are fixed:

- **Unequip chip**: the visible chip stays 22px, but on `body.mobile-touch` it now carries an invisible centered `::before` hit-expander that restores a 40x40 tap target (honoring the 40x40 floor in src/ui/CLAUDE.md, not the 24px minimum). Layout/centering is untouched since the chip was already absolute.
- **Attribute tiles**: dropped the `min-height: 0` override; the mobile tiles now keep the same 40px long-press floor as their Offense/Defense siblings (with `justify-content: center` so the content stays centered).

Both are pinned in tests/charscreen_css.test.ts so they can't silently regress.

Non-blocking: registered `quality_glow.ts` in `UI_PURE_CORES` + `BARE_NAMED` in tests/architecture.test.ts so its purity is guarded going forward. Left the glow shadow literals as-is per your note (cold open path, not per-frame).

tsc clean, guard + CSS tests green, rebased on release/v0.29.0.

Issue #2194, opened 2026-07-20
Release v0.29.0


Rubsey:
### Target version

0.29.0

### Release branch

`release/v0.29.0`, cut from `main` at `623b10aeed496489c9e15e88adbbaae3b9db9168` (the tagged v0.28.0 release commit).

### Bump the version

- [ ] Ran `npm version <X.Y.Z>` (or `npm version minor`/`patch`/`major`) and pushed the resulting commit and tag.
- [ ] Confirmed all three manifests show the new version and the native build numbers incremented.

### Verify

- [ ] `npm test` passes.
- [ ] `npm run build` succeeds.
- [ ] Ran the release malware-audit gate.

### Ship

- [ ] Web deploy is live (see `DEPLOY.md`).
- [ ] Mac desktop Electron app release artifact is published.
- [ ] iOS archive built via `npm run native:open:ios` and submitted to App Store Connect.
- [ ] Android bundle built via `npm run native:open:android` and submitted to Play Console.
- [ ] Release notes added under `docs/release-notes/`.

### Carryover from v0.28.0

Carried over from the v0.28.0 release tracker (#2093): every referenced issue and PR that is still open and whose PR base is `release/v0.27.0` or newer, including the tracked open PRs on `main`. This audit found 43 open PRs and 4 open issues that qualify.

Migration rule applied this cycle: merged and closed items are omitted. The separate v0.24.0 PTR track (#1744) continues separately, as before.

Pruned 2026-07-22: the crypto/web3/RiverBoat track and the major content epics (new class, Rifts, new dungeons, PvP season, game modes) were removed from this checklist. They are not v0.29.0 release-gate items; they proceed on their own PRs and must not be re-added by future referenced-PR carryover audits (the removed list is in this issue's edit history).

#### Open PRs targeting the retired `release/v0.28.0` (need retargeting to `release/v0.29.0`)

- [ ] #1385 fix(admin): require a full-scope token for the admin API
- [ ] #1556 feat(content): add Deepfen Shallows reed dressing
- [ ] #1700 fix(render): default Apple Silicon Macs to medium graphics, not ultra (#1676)
- [ ] #1702 feat(game): remember the camera zoom level across sessions (#1657)
- [ ] #1754 fix(ui): make the action-button cooldown legible with a radial sweep (#1660)
- [ ] #1770 feat(ui): overhead chat bubbles for party channel messages (#1659)
[...]

PR #2251, opened 2026-07-21
feat(music): stream the remastered mp3 soundtrack in place of runtime synthesis

maxpolaczuk:
## Summary

Replaces the in-game procedural WebAudio soundtrack with the remastered mp3 renders of the same themes, streamed efficiently so zone changes crossfade exactly as before while playback costs no synthesis CPU and no up-front download.

- Every zone cue (three towns, vale, marsh, peaks, three dungeons) is a looping media element under `public/audio/music/`, routed through `createMediaElementSource` into the existing WebAudio graph (per-stream crossfade gain, master, compressor), the same file-track pattern the Sowfield stadium music already uses.
- Zone streams are created lazily on first activation and resume mid-track on re-entry (dungeon entry still restarts its cue from the top via `resetForDungeonEntry`). Fade constants are unchanged from the synth director, so area transitions sound identical.
- Both battle themes (`combat_1`, `combat_2`) are warmed at `init()` so a fight never waits on a first-byte fetch; each fight opens on one picked at random (`pickCombatTrackIndex`, injectable rand) and restarted from the top, holding through zone borders mid-chase.
- A 500 ms keeper interval pauses any stream that has finished fading out (no decoding or bandwidth for inaudible cues), revives active streams that autoplay or a tab restore left paused, and is invoked synchronously on unmute and menu close so revival is instant. Activation while muted no longer starts a download at all.
- The composition machinery (`MusicSynth`, `buildMusicThemes`, `THEME_TRIM`, the note data, `MUSIC_OVERRIDES`) stays untouched as the authoring source for the music editor and `scripts/render_music.mjs`, which is where these remasters come from. Boss (Nythraxis) and Sowfield paths are unchanged.
- New `src/game/music_tracks.ts` (catalog + combat pick, pure and Node-tested); docs (`src/game/CLAUDE.md`, editor header, generated-file header, CREDITS) updated to match.

## Type of change

## How was this tested?

- Commands:
  - `npm run gate`: all steps green except two pre-existing environmental failures in `tests/deploy_watchdog.test.ts` (fake-docker hang bounds), which fail identically on the pristine base commit `72d5c7220` with no changes applied, so they are not from this branch. Full vitest otherwise: 16788 passed. `npm run test:browser` (66 passed), `check:types`, `build:env`, `build:server`, `build` all green.
  - `npx vitest run tests/music.test.ts tests/music_tracks.test.ts tests/instance_music.test.ts`: 32 tests pinning the streamed director (zone/combat exclusivity, fade asymmetry, lazy creation, random per-fight pick with restart, resume-vs-restart, keeper pause/revive, muted-activation no-download, boss loop, dungeon reset, catalog/asset existence).
  - `node scripts/music_stream_e2e.mjs` (new, against `npm run dev` and a real headless Chrome): 19 checks against the live game, all passing: entry streams the spawn town theme with both battle themes warmed silent, `/dev tp` across the marsh border crossfades, the Fenbridge hub swaps to its town theme, mob aggro brings in one randomly picked battle theme and silences the zone, dropping aggro hands back to the zone mid-track, and every fully faded stream ends paused.
- Manual steps: verified the mp3s serve as `audio/mpeg` from the dev server and land verbatim in `dist/audio/music/` after `npm run build`.

## Screenshots / recordings

N/A (audio-only change; the E2E output above is the observable proof).

---

### Quality

- [x] **The full gate passes** except the two pre-existing `deploy_watchdog` environment failures reproduced on the clean base (details above). Decisive tests added for all changed behavior.

### Cross-platform

- [x] Streaming uses the same media-element + WebAudio pattern the shipped boss and Sowfield tracks already use on desktop, iOS, and Android shells; no UI changed.

### Localization (i18n)

- [x] No new player-visible strings.


maxpolaczuk:
Coverage-review triage (fresh-agent pass over the diff, 17 findings):

**Fixed in bc593c3c6:**
- Streams now pause under EVERY duck, not just the toggle and menu: the boss and Sowfield file tracks and volume zero all count as inaudible, so a multi-minute Nythraxis fight no longer decodes a muted combat mp3, and handback/stadium-exit/volume-restore revive paused streams synchronously.
- Media elements are created on first audible activation, so a player with `ev_music_on='0'` downloads zero bytes (previously the two combat warms still fetched ~1.6MB).
- Keeper revival now resumes the AudioContext too (iOS interruption / tab restore).
- A pull chained into the previous fight's fade-out keeps the track position instead of jump-cutting to the top; distinct-track overlap during the ~2s tail is an intentional crossfade.
- `createMediaElementSource` failure now mutes the element instead of letting it play outside every gain and duck.
- Silent-pause margin raised to 4s so the slowest fade tail is below -45 dB at pause time.
- Stale headers updated: `scripts/music_render_entry.ts`, `scripts/render_music.mjs`, `music_editor.html`, and `music_overrides.generated.ts` (header aligned byte-for-byte with what the updated generator template now writes; theme data untouched).
- Test gaps closed: fade time constants, combat fade asymmetry, the volume/master path, `init()` twice, `update()` before `init()`, disabled-at-boot zero downloads, duck pausing and revival, chained-pull no-rewind. The E2E now aggroes the nearest mob (a distant pick could leash-clear and flake the combat phase).

**Intentionally not changed:**
- World-entry first-byte latency: the home-page theme keeps playing through loading and fades after entry (`main.ts`), so the spawn stream's spin-up is covered audibly.
- Combat handback resumes the zone mid-track (the old synth restarted from the top): same intentional immersion divergence as overworld re-entry, pinned by the E2E.
- `vale_legacy` mapping to vale.mp3 could create a second element if that dead route ever revives: the browser HTTP cache dedupes the download; not worth indirection today.
- The keeper interval lives for the app lifetime, parity with the old 110ms scheduler (and 4.5x less frequent).

All checks re-run after the fixes: music suites 40 passed, architecture/malware/game-audio green, browser E2E 19/19 against the live game (this run randomly opened on combat_2, confirming the per-fight pick varies).

maxpolaczuk:
Heads-up: this was merged onto the retired `release/v0.28.0` by mistake (the active integration base is `release/v0.29.0`, see the tracker #2194). The merge has been reverted on that branch in 0473be665, and the feature is resubmitted rebased onto `release/v0.29.0` as #2261.

PR #2253, opened 2026-07-21
fix(server): resync own talents after /unspectate

jgyy:
## Summary
- Fixes a bug where using `/spectate` on another player can swap the moderator's own action-bar/talents to the spectated target's for a window after `/unspectate`, until the staggered self-refresh backstop or a manual loadout re-select forces a resync.
- Root cause: the heavy self-snapshot block (`tal`/`inv`/`equip`/`bags`/...) in `selfWireJson` is gated on `meta.wireRev !== session.lastWireRev`. During spectate the `meta` passed in belongs to whichever entity is currently anchored (the target while spectating, the moderator again after `/unspectate`), but the gate variable (`session.lastWireRev`) lives on the moderator's own session. If the two wireRev values happen to coincide (e.g. neither player's talents changed since login, both default to `0`), the heavy block is skipped entirely on enter AND exit, so the moderator's client keeps mirroring the spectated target's talent allocation after returning to their own body.
- Fix: force `selfHeavyDirty = true` in both `enterSpectate` and `exitSpectate` (`server/game.ts`), the same pattern already used at login and by `resyncQuests`, so the heavy block always re-runs on the very next snapshot regardless of wireRev coincidence.
- Note (per review): the wire-level gap is bounded even without this fix, since the heavy gate has a staggered backstop (`(tickCount + pid) % HEAVY_SELF_REFRESH_TICKS`, about 2 seconds) that resends the correct talents on its own. The fix is still worth having: it closes that 2-second window where a foreign talent kit is live on the moderator's client, and it's needed on the enter side regardless so the spectate view populates immediately instead of waiting on the backstop.

## Test plan
- [x] Added a regression test in `tests/moderation_game.test.ts` that reproduces the exact wireRev-coincidence bug (both talent allocations pinned so the test fails deterministically without the fix, confirmed locally by reverting the fix) and asserts the moderator's own `tal.alloc` is resent (not the spectated target's) immediately after `/unspectate`.
- [x] `npx vitest run tests/moderation_game.test.ts tests/moderation_service.test.ts tests/moderation_commands.test.ts tests/snapshots.test.ts tests/architecture.test.ts tests/command_schema.test.ts` all green.
- [x] `npx @biomejs/biome check` on changed files clean (whole-repo check is pre-existing debt per CLAUDE.md, not touched here).
- No screenshots: this is a server-only wire-sync fix, no `src/render`/`src/ui`/`src/styles` changes.

```mermaid
sequenceDiagram
    participant Client as Moderator client
    participant Server as GameServer
    participant Target as Spectated player meta

    Client->>Server: /spectate "Suspect"
    Server->>Server: enterSpectate (BEFORE: no selfHeavyDirty force)
    Server-->>Client: self snapshot built from Target meta (tal = Target's talents)
    Client->>Server: /unspectate
    Server->>Server: exitSpectate (BEFORE: no selfHeavyDirty force)
    Note over Server: meta.wireRev (own) == session.lastWireRev (stale, from Target)<br/>heavy block SKIPPED
    Server-->>Client: self snapshot omits 'tal' -> client stays on Target's talents (BUG)

    Note over Server: AFTER fix: selfHeavyDirty=true set in both enterSpectate and exitSpectate
    Server-->>Client: heavy block always re-runs next snapshot -> own 'tal' resent correctly
```

### DIFF
--- a/server/game.ts
+++ b/server/game.ts
@@ -1548,6 +1548,11 @@ export class GameServer {
     moderator.lastArenaWireTick = -ARENA_WIRE_INTERVAL_TICKS;
     moderator.lastDfWireTick = -DF_WIRE_INTERVAL_TICKS;
     moderator.sentEnts.clear();
+    // force the heavy self block (tal/inv/equip/bags/...) to re-run next
+    // snapshot: it is gated on meta.wireRev vs session.lastWireRev, and that
+    // comparison is keyed to whichever entity's meta is being wired, so
+    // without this the target's heavy fields can silently fail to resend.
+    moderator.selfHeavyDirty = true;
     this.send(moderator, { t: 'spectate', name: target.name });
     this.sendSystemNotice(moderator, `Now spectating ${target.name}.`);
   }
@@ -1572,6 +1577,10 @@ export class GameServer {
     moderator.lastArenaWireTick = -ARENA_WIRE_INTERVAL_TICKS;
     moderator.lastDfWireTick = -DF_WIRE_INTERVAL_TICKS;
     moderator.sentEnts.clear();
+    // same as enterSpectate: force the heavy self block to re-run so the
+    // moderator's OWN talents/inventory/equip/etc. resend immediately
+    // instead of staying stuck on the spectated target's last-sent values.
+    moderator.selfHeavyDirty = true;
     this.send(moderator, { t: 'spectate', name: null });
     if (announce) this.sendSystemNotice(moderator, 'Stopped spectating.');
   }
--- a/tests/moderation_game.test.ts
+++ b/tests/moderation_game.test.ts
@@ -60,6 +60,7 @@ type TestFrame = {
     nm: string;
     ack: number;
     party: { members: { pid: number; x: number; z: number }[] };
+    tal?: { alloc: { spec: string | null; rows: Record<number, string> } };
   };
 };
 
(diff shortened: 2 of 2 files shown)


Rubsey (commented):
The root cause checks out and the fix is the right shape. Verified against the head, not just the description:

- During spectate the broadcast loop anchors the self snapshot on the target's meta (server/game.ts:5185), while the heavy gate compares that meta's wireRev against the moderator's own session.lastWireRev (server/game.ts:5743). A wireRev coincidence therefore skips the tal/inv/equip block on both enter and exit, exactly as described.
- Forcing selfHeavyDirty matches the existing precedent (the character-switch path, resyncQuests, the mail resolve path). Spectate is the only flow that swaps which meta gets wired (jail visit keeps the moderator's own meta), so the two call sites cover it, and since the assignment sits after the if/else in enterSpectate it also covers switching targets mid-spectate.
- Ran the new test both ways in a worktree: green at the head, and with server/game.ts reverted to base it fails at the spectate-snapshot assertion (tal absent), so the pin is decisive. Merge to release/v0.29.0 is clean and CI is green.

One note, on the description rather than the code:

- nit: "permanently swap ... until they manually re-select a saved talent loadout" overstates the wire-level impact. The heavy gate has a staggered backstop, (tickCount + pid) % HEAVY_SELF_REFRESH_TICKS at server/game.ts:5744 (40 ticks, about 2 s), and exitSpectate clears lastSent, so the moderator's own tal resends within a couple of seconds even without this fix. The fix is still worth having (a foreign talent kit on a live client for 2 s is real, and the enter side wants the same force so the spectate view populates immediately instead of waiting on the backstop), but if someone actually observed a persistent action-bar scramble, the persistence would have to live client-side and deserves its own look.

jgyy:
Good catch, thanks for tracing the staggered backstop. You're right that the wording overstates it: with `(tickCount + pid) % HEAVY_SELF_REFRESH_TICKS` the foreign talent kit resolves on its own within about 2 seconds even without this fix, not permanently. Updating the PR description to say so. Keeping the fix as-is since a live foreign kit for 2 seconds during spectate enter is still worth forcing immediately rather than waiting on the backstop, and it's needed on the enter side regardless so the spectate view populates right away.

